PR #3123, opened 2020-01-13
test: refactor ssh related functions in test helper classes

skshetry:

### DIFF
--- a/tests/remotes.py
+++ b/tests/remotes.py
@@ -66,23 +66,6 @@ def _should_test_gcp():
     return True
 
 
-def _should_test_ssh():
-    do_test = env2bool("DVC_TEST_SSH", undefined=None)
-    if do_test is not None:
-        return do_test
-
-    # FIXME: enable on windows
-    if os.name == "nt":
-        return False
-
-    try:
-        check_output(["ssh", "-o", "BatchMode=yes", "127.0.0.1", "ls"])
-    except (CalledProcessError, IOError):
-        return False
-
-    return True
-
-
 def _should_test_hdfs():
     if platform.system() != "Linux":
         return False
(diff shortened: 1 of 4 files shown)


tests/remotes.py
```diff
@@ -114,31 +97,6 @@ def get_local_url():
     return get_local_storagepath()
 
 
-def get_ssh_url():
-    return "ssh://{}@127.0.0.1:22{}".format(
-        getpass.getuser(), get_local_storagepath()
-    )
```
skshetry: I did not find any usage of this, probably, because we are moving away from real server to mocked ssh server? Anyway, I didn't want to change this atm.

tests/remotes.py
```diff
@@ -269,8 +227,53 @@ def get_url():
...
+            # NOTE: On Windows get_local_storagepath() will return an
+            # ntpath that looks something like `C:\some\path`, which is not
+            # compatible with SFTP paths [1], so we need to convert it to
+            # a proper posixpath.
+            # To do that, we should construct a posixpath that would be
+            # relative to the server's root.
+            # In our case our ssh server is running with `c:/` as a root,
+            # and our URL format requires absolute paths, so the
+            # resulting path would look like `/some/path`.
+            #
```
skshetry: I splitted the comment here as it was getting more than 80chars.

tests/remotes.py
efiop: @skshetry I think it is still used in tests/func/test_api.py . We are close to getting rid of it, that is true 🙂 Let's not tackle it in this PR though.

skshetry: Yep, it is still used in https://github.com/iterative/dvc/blob/028d30e22d1e1088b0e7b68df063380f76f2a818/tests/func/test_api.py#L26

I only grepped for the `get_ssh_url`/`SSH.get_url` and didn't find any. 🤦‍♂️

Issue #2793, opened 2019-11-14
Remove back ticks (and quotes) from command output? How to highlight?


jorgeorpinel:
In #2644 @efiop and I decided it would be good to standardize the use of back ticks ` and single quotes in command outputs. We decided the former would be used for file names, commands, arguments, etc. and the latter for dynamic output (stuff that varies in the message). Let's review this convention per 
 https://github.com/iterative/dvc/pull/2765#pullrequestreview-316671192. @shcheklein's comments:

> is it a regular practice to use backticks in command line? It seems to me that we should not use markup if the client (command line) does not know how to render it. CLI UI should use plain english with common regular rules. For example it should not include jso or XML or html or Markdown... Probably it's a good idea to highlight path somehow. I would take look at how other popular command line tools do this.

jorgeorpinel:
> highlight path somehow ... how other popular command line tools do this.

I think mostly with terminal text coloring e.g. using `colorama` in Python (related: #2784)

So yeah, firstly, separating between back ticks and single quotes was probably not that useful since to the user, which part of the text is dynamic is probably completely irrelevant. So the convention I would recommend is to either only use `'` which I doubt can cause any trouble in terminal output (we need to check) or to not use either type of quote, but rather a special color that's not used for anything else (again, see #2784). What do you think?

efiop:
backtics are not only markup, they are also an old-ish way of "evaluate" for your commands. E.g. in sh you can use backtics or $() to wrap your command.

efiop:
You can look through things like cpython, backtics are quite widely used in command output.

jorgeorpinel:
@shcheklein what do you think?

@efiop OK good points.  But so you don't think colors are better than back ticks?
And what do you think about stopping using single quotes per:

> probably not that useful since, to the user, which part of the output string is dynamic is probably completely irrelevant...

?

shcheklein:
@jorgeorpinel I would say it's a bit unusual for me, but I don't have a strong opinion - so I'm fine with either for now. We can get back to it later when we start doing UI/UX more systematically.

efiop:
@jorgeorpinel Colors are good too, but they won't show up in the logs when you redirect them to the file. Marking things with quotes and ticks is not only for the user, but also for the developer, so I can visually inspect the log and understand what is a dynamic part and what is not on first sight, instead of having to look into code right away.

jorgeorpinel:
[...]

PR #3130, opened 2020-01-13
api: better code documentation

### LINKED ISSUE #3092: api: inline documentation (docstrings)
I don't think we use docstrings much since the code is meant to be self-documenting for developers.

However, for `dvc.api` (currently a single file: https://github.com/iterative/dvc/blob/master/dvc/api.py) I think we should have significant code documentation (multi-line file and function [docstrings](https://www.python.org/dev/peps/pep-0257/#id17), other comments).

Let's find a good docstring format we want to use and maintain e.g. [Sphinx](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html)?

- [ ] Should probably also work on iterative/dvcx/issues/1 along with this one.

jorgeorpinel:
Close #3092

### DIFF
--- a/dvc/api.py
+++ b/dvc/api.py
@@ -58,6 +58,7 @@ def get_url(path, repo=None, rev=None, remote=None):
     """
     Returns the full URL to the data artifact specified by its `path` in a
     `repo`.
+    NOTE: There is no guarantee that the file actually exists in that location.
     """
     try:
         with _make_repo(repo, rev=rev) as _repo:
--- a/dvc/api.py
+++ b/dvc/api.py
@@ -45,7 +45,10 @@ class SummonError(DvcException):
 
 
 def get_url(path, repo=None, rev=None, remote=None):
-    """Returns an url of a resource specified by path in repo"""
+    """
+    Returns the full URL to the data artifact specified by its `path` in a
+    `repo`.
+    """
     with _make_repo(repo, rev=rev) as _repo:
         abspath = os.path.join(_repo.root_dir, path)
         out, = _repo.find_outs_by_path(abspath)
@@ -54,7 +57,7 @@ def get_url(path, repo=None, rev=None, remote=None):
 
 
 def open(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
-    """Opens a specified resource as a file descriptor"""
+    """Context manager to open a file artifact as a file object."""
     args = (path,)
     kwargs = {
         "repo": repo,
@@ -73,7 +76,7 @@ class _OpenContextManager(GCM):
 
     def __getattr__(self, name):
         raise AttributeError(
-            "dvc.api.open() should be used in a with statement"
+            "dvc.api.open() should be used in a with statement."
         )
 
 
@@ -87,7 +90,7 @@ def _open(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
 
 
 def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
-    """Read a specified resource into string"""
+    """Returns the contents of a file artifact."""
     with open(
         path, repo=repo, rev=rev, remote=remote, mode=mode, encoding=encoding
     ) as fd:
@@ -97,7 +100,9 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
 @contextmanager
 def _make_repo(repo_url, rev=None):
     if not repo_url or urlparse(repo_url).scheme == "":
-        assert rev is None, "Custom revision is not supported for local repo"
+        assert (
+            rev is None
+        ), "Git revisions are not supported for local DVC projects."
         yield Repo(repo_url)
     else:
         with external_repo(url=repo_url, rev=rev) as repo:
@@ -105,7 +110,7 @@ def _make_repo(repo_url, rev=None):
 
 
 def summon(name, repo=None, rev=None, summon_file="dvcsummon.yaml", args=None):
-    """Instantiate an object described in the summon file."""
+    """Instantiate an object described in the `summon_file`."""
     with prepare_summon(
         name, repo=repo, rev=rev, summon_file=summon_file
     ) as desc:


dvc/api.py
```diff
@@ -45,7 +45,10 @@ class SummonError(DvcException):
 
 
 def get_url(path, repo=None, rev=None, remote=None):
-    """Returns an url of a resource specified by path in repo"""
+    """
+    Returns the full URL to the data artifact specified by its `path` in a
+    `repo`.
```
pared: I would suggest noting here that it does not guarantee that URL exists.
Related #3155

Suor: I don't like backticks, we don't use them anywhere else. Python stdlib doesn't use them either.

shcheklein: any suggestions? is there a standard way of doing markup that is understood by the tools that generate a view for this (like F1 in PyCharm, quick help in VS Code, docs generators, etc)

btw, we do use them as far as I can tell internally, also, Markdown markup or some markup is being used in other libraries for external APIs

Python stdlib:

```
Return whether an object is an instance of a class or of a subclass thereof.
    
    A tuple, as in ``isinstance(x, (A, B, ...))``, may be given as the target to
    check against. This is equivalent to ``isinstance(x, A) or isinstance(x, B)
    or ...`` etc.
```

double backtics to highlight the code (and probably make it linkable).

Am I missing something here?

mroutis: @shcheklein , there are plenty _docstring_ formats: https://stackoverflow.com/questions/3898572/what-is-the-standard-python-docstring-format 

From [PEP 287](https://www.python.org/dev/peps/pep-0287/):
> Text enclosed in single backquotes is recognized as "interpreted text", whose interpretation is application-dependent. In the context of a Python docstring, the default interpretation of interpreted text is as Python identifiers. The text will be marked up with a hyperlink connected to the documentation for the identifier given

Not sure if any particular IDE takes advantages of such feature.

> we don't use them anywhere else

@Suor, we actually do :sweat_smile:  but it is quite inconsistent: `grep -R '`.*`' dvc`

--- 

I'd say that there's no need to block a PR for this. I created a separate issue to discuss the formatting across the code base: https://github.com/iterative/dvc/issues/3176

dvc/api.py
```diff
@@ -97,15 +100,17 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
 @contextmanager
 def _make_repo(repo_url, rev=None):
     if not repo_url or urlparse(repo_url).scheme == "":
-        assert rev is None, "Custom revision is not supported for local repo"
+        assert (
+            rev is None
+        ), "Git revisions are not supported for local DVC projects."
```
mroutis: Why not "local repositories"? Not sure I understand what's a _local DVC project_.

dvc/api.py
shcheklein: > Not sure if any particular IDE takes advantages of such feature.

yep, all of those I've used at least.

jorgeorpinel: > it does not guarantee that URL exists
OK, noted in 2b818e20 @pared.

dvc/api.py
jorgeorpinel: A DVC project isn't always a repo. It can be initialized with `--no-scm` and you can provide the local path to that project in the file system to the `repo` argument (misleading name BTW) of API functions.

jorgeorpinel:
Appreciate it, @mroutis and everyone! My idea was that this PR (which is a draft BTW) would just help start the conversation about docstrings... And it has! I guess I'll make it a regular PR, work on getting it merge, and open a separate issue to further improve on the docstrings with whatever we decide in https://github.com/iterative/dvc/pull/3130#pullrequestreview-343835585 above.

dvc/api.py
jorgeorpinel: > we don't use them anywhere else

@Suor are there docstrings anywhere else? In any case, the API is user-facing so we probably don't need to compare it with the rest of the code in terms of comments/docstrings. And we do use back ticks in all of the `dvc` command output referring to files, commands, cmd options, etc. See #2793

jorgeorpinel: UPDATE: OK, I see Ramón extracted the current use of docstrings and backquotes in the rest of the core repo to #3176. Thanks!

jorgeorpinel: > Text enclosed in single backquotes is recognized as "interpreted text"... In the context of a Python docstring, the default interpretation of interpreted text is as Python identifiers. The text will be marked up with a hyperlink connected to the documentation for the identifier given

This is what I was going for here. So keep it for now guys? I don't initially have a strong opinion.

>  there a standard way of doing markup that is understood by the tools that generate a view for this (like F1 in PyCharm, quick help in VS Code, docs generators, etc)

Yeah great question @shcheklein! I'll look it up.

jorgeorpinel: > I'll look it up.

Ok so for doc generators, as probably all of you know, [Sphinx](http://www.sphinx-doc.org/en/master/) is by far the most popular, with [pdoc](https://pdoc3.github.io/pdoc) a far 2nd place AFAICS.

Sphinx doesn't support Markdown. actually, but rather[reStructuredText](https://docutils.sourceforge.io/docs/user/rst/quickref.html). BTW, backquotes in reST are indeed used for interpreted text, among other uses.

**pydoc** is interesting because it supports md and [other interesting formats](https://pdoc3.github.io/pdoc/doc/pdoc/#supported-docstring-formats) worth checking out actually (from Numpy and Google).

> A good general guide about docstring options is in https://docs.python-guide.org/writing/documentation/

> An interesting VSCode extension to help write [docstrings](https://www.python.org/dev/peps/pep-0257/) in all of the above formats is https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring

Still need to research how IDE cod completion works...

jorgeorpinel: > Still need to research how IDE code completion works...

VSCode [intellisense](https://code.visualstudio.com/docs/editor/intellisense#_intellisense-features) doesn't interpret any format in the docstrings it seems. It mostly works by building and inspecting the Python code by itself.

image

image

And from what I can see [here](https://www.jetbrains.com/help/pycharm/auto-completing-code.html) PyCharm works similarly.

So for IDEs, easy to read docstrings e.g. [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) or [Google style](http://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings) may be best.

jorgeorpinel: p.s. numpydoc uses single backticks to enclose variables. Google doesn't seem to use them at all though.

jorgeorpinel:
> open a separate issue to further improve on the docstrings with whatever we decide

Actually, I guess we can just use #3176 for this? I'll add some notes. Although again, API code doc should probably be held to a higher standard than regular ones 🙂

dvc/api.py
jorgeorpinel: Extracted docstring format discussion fully to https://github.com/iterative/dvc/issues/3176#issuecomment-575927160

jorgeorpinel: So to resolve this one @Suor, as most available formats use backquotes except Google, are you able to approve this for the time being? We will revisit this once #3176 is resolved anyway 🙂

PR #3132, opened 2020-01-14
remote: error out when modifying a non-existing remote

### LINKED ISSUE #3131: remote: allow modifying only existing entries
**DVC version**: `0.81.0`

Using `dvc remote modify` on a non-existing entry creates an invalid remote configuration, since it creates a remote without a URL.

```bash
dvc init --no-scm
dvc remote modify missing <option> <value>
```

Ideally, raising an error should prevent the user from introducing an invalid remote entry.

[Discord discussion](https://discordapp.com/channels/485586884165107732/485596304961962003/666411899499315221)

mroutis:
Close #3131

### DIFF
--- a/dvc/remote/config.py
+++ b/dvc/remote/config.py
@@ -3,8 +3,7 @@ import os
 import posixpath
 from urllib.parse import urlparse
 
-from dvc.config import Config
-from dvc.config import ConfigError
+from dvc.config import Config, ConfigError
 from dvc.utils import relpath
 
 
@@ -134,6 +133,8 @@ class RemoteConfig(object):
                 break
 
     def modify(self, name, option, value, level=None):
+        self.get_settings(name)
+
         self.config.set(
             Config.SECTION_REMOTE_FMT.format(name), option, value, level=level
         )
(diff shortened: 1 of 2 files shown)


dvc/remote/config.py
```diff
@@ -134,6 +133,8 @@ def remove(self, name, level=None):
                 break
 
     def modify(self, name, option, value, level=None):
+        self.get_settings(name)
```
mroutis: Using `get_settings` raises the following error when a remote is not found:

```python
raise ConfigError("unable to find remote section '{name}''")
```

tests/func/test_remote.py
```diff
@@ -31,10 +31,10 @@ def test(self):
 
         self.assertEqual(main(["remote", "list"]), 0)
 
-        self.assertEqual(main(["remote", "remove", remotes[0]]), 0)
         self.assertEqual(
             main(["remote", "modify", remotes[0], "option", "value"]), 0
         )
+        self.assertEqual(main(["remote", "remove", remotes[0]]), 0)
```
mroutis: Updated the this test because it was trying to modify a setting after removing it :sweat_smile:

dvc/remote/config.py
efiop: @mroutis Let's add a comment there though, please.

PR #3133, opened 2020-01-14
remote: config checksum_jobs

### LINKED ISSUE #2920: core.checksum_jobs doesn't work for remotes
https://github.com/iterative/dvc/issues/2891#issuecomment-562812250

JIoJIaJIu:
Close #2920
## Additional details
I would like to suggest to define and use such terms for avoiding ambiguity                                                              
- **config** - global repo config                                                                                                                        
- **settings** - dict of **config** section                                                                                                            
Hence refactor **RemoteBase** from                                                                                                                     
```
    def __init__(self, repo, config):  ->  def __init__(self, repo, settings):
```
with all changes as `config -> settings`
                                                                                                                                                
Then get some setting as:                                                                                                                
1. Check first into component **settings**                                                                                                             
2. Check in **config** if requires                                                                                                                     
3. Return default 
<hr/>

### DIFF
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -86,11 +86,7 @@ class RemoteBASE(object):
         self.repo = repo
 
         self._check_requires(config)
-
-        core = config.get(Config.SECTION_CORE, {})
-        self.checksum_jobs = core.get(
-            Config.SECTION_CORE_CHECKSUM_JOBS, self.CHECKSUM_JOBS
-        )
+        self.checksum_jobs = self._get_checksum_jobs(config)
 
         self.protected = False
         self.no_traverse = config.get(Config.SECTION_REMOTE_NO_TRAVERSE, True)
@@ -141,6 +137,19 @@ class RemoteBASE(object):
             "using https://github.com/iterative/dvc/issues. Thank you!"
         ).format(url, missing, " ".join(missing), self.scheme)
         raise RemoteMissingDepsError(msg)
+
+    def _get_checksum_jobs(self, config):
+        checksum_jobs = config.get(Config.SECTION_REMOTE_CHECKSUM_JOBS)
+        if checksum_jobs:
+            return checksum_jobs
+
+        if self.repo:
+            core = self.repo.config.config.get(Config.SECTION_CORE, {})
+            return core.get(
+                Config.SECTION_CORE_CHECKSUM_JOBS, self.CHECKSUM_JOBS
+            )
+
+        return self.CHECKSUM_JOBS
 
     def __repr__(self):
         return "{class_name}: '{path_info}'".format(
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -186,6 +186,7 @@ class Config(object):  # pylint: disable=too-many-instance-attributes
     SECTION_GDRIVE_CLIENT_SECRET = "gdrive_client_secret"
     SECTION_GDRIVE_USER_CREDENTIALS_FILE = "gdrive_user_credentials_file"
 
+    SECTION_REMOTE_CHECKSUM_JOBS = "checksum_jobs"
     SECTION_REMOTE_REGEX = r'^\s*remote\s*"(?P<name>.*)"\s*$'
     SECTION_REMOTE_FMT = 'remote "{}"'
     SECTION_REMOTE_URL = "url"
@@ -214,6 +215,7 @@ class Config(object):  # pylint: disable=too-many-instance-attributes
         SECTION_GCP_PROJECTNAME: str,
         SECTION_CACHE_TYPE: supported_cache_type,
         Optional(SECTION_CACHE_PROTECTED, default=False): Bool,
+        SECTION_REMOTE_CHECKSUM_JOBS: All(Coerce(int), Range(1)),
         SECTION_REMOTE_USER: str,
         SECTION_REMOTE_PORT: Coerce(int),
         SECTION_REMOTE_KEY_FILE: str,
--- a/tests/unit/remote/test_gdrive.py
+++ b/tests/unit/remote/test_gdrive.py
@@ -1,6 +1,7 @@
 import pytest
 import os
 
+from dvc.config import Config
 from dvc.remote.gdrive import (
     RemoteGDrive,
     GDriveAccessTokenRefreshError,
@@ -14,6 +15,7 @@ USER_CREDS_MISSED_KEY_ERROR = "{}"
 
 class Repo(object):
     tmp_dir = ""
+    config = Config()
 
 
 class TestRemoteGDrive(object):
--- a/tests/unit/remote/test_remote.py
+++ b/tests/unit/remote/test_remote.py
@@ -1,0 +1,41 @@
+from dvc.remote import Remote
+
+
+def set_config_opts(dvc, commands):
+    list(map(lambda args: dvc.config.set(*args), commands))
+
+
+def test_remote_with_checksum_jobs(dvc):
+    set_config_opts(
+        dvc,
+        [
+            ('remote "with_checksum_jobs"', "url", "s3://bucket/name"),
+            ('remote "with_checksum_jobs"', "checksum_jobs", 100),
+            ("core", "checksum_jobs", 200),
+        ],
+    )
+
+    remote = Remote(dvc, name="with_checksum_jobs")
+    assert remote.checksum_jobs == 100
+
+
+def test_remote_without_checksum_jobs(dvc):
+    set_config_opts(
+        dvc,
+        [
+            ('remote "without_checksum_jobs"', "url", "s3://bucket/name"),
+            ("core", "checksum_jobs", "200"),
+        ],
+    )
+
+    remote = Remote(dvc, name="without_checksum_jobs")
+    assert remote.checksum_jobs == 200
+
+
+def test_remote_without_checksum_jobs_default(dvc):
+    set_config_opts(
+        dvc, [('remote "without_checksum_jobs"', "url", "s3://bucket/name")]
+    )
+
+    remote = Remote(dvc, name="without_checksum_jobs")
+    assert remote.checksum_jobs == remote.CHECKSUM_JOBS


tests/func/test_remote.py
```diff
@@ -170,6 +170,43 @@ def test(self):
         self.assertEqual(ret, 0)
 
 
+class TestRemoteChecksumJobs(TestDvc):
```
mroutis: @JIoJIaJIu , sorry for the confusion and extra work, but we are in the middle of a migration from `unittest` to `pytest`.
Could you rewrite your tests to `pytest`? Happy to help with further questions about this)

dvc/remote/base.py
```diff
@@ -142,6 +138,21 @@ def _check_requires(self, config):
         ).format(url, missing, " ".join(missing), self.scheme)
         raise RemoteMissingDepsError(msg)
 
+    def _get_checksum_jobs(self, repo, config):
```
mroutis: Are we going to have both, _core_ and _remote_ `checksum_jobs`?  Why not going `remote`-only?

dvc/remote/base.py
```diff
@@ -142,6 +138,21 @@ def _check_requires(self, config):
...
+        checksum_jobs = config.get(Config.SECTION_REMOTE_CHECKSUM_JOBS)
+        if not checksum_jobs:
+            if repo:
```
mroutis: Since you are catching `AttributeError`, you could remove the `if/else`. If there's no `repo` it is going to do `None.config` and fallback to `self.CHECKSUM_JOBS`.

mroutis:
Thanks for the PR, @JIoJIaJIu, left some comments :)

Is your suggestion about using `dvc.config.settings` instead of `dvc.config.config`?

dvc/remote/base.py
JIoJIaJIu: I've implemented fallback logic as was described in [the comment](https://github.com/iterative/dvc/issues/2920#issuecomment-564128058)

JIoJIaJIu:
> Thanks for the PR, @JIoJIaJIu, left some comments :)
> 
> Is your suggestion about renaming `dvc.config.config` to `dvc.config.settings`?

Hello @mroutis 
No, I meant the part when some component is initialized with partial of the `dvc.config`
Here for example it has names as a **settings** - https://github.com/iterative/dvc/blob/fe635a5040c9d593cd1bdef3fa31f4df7af85259/dvc/remote/__init__.py#L42
But further in **RemoteBASE** it has naming as `config` that introduces ambiguity - https://github.com/iterative/dvc/blob/fe635a5040c9d593cd1bdef3fa31f4df7af85259/dvc/remote/base.py#L85

I suggest just to define such terms as
```
# whole config
...
[core]  # specific section - core
# everything under section are settings
checksum_jobs = 100

['remote "aws"'] # specific section - remote aws
# everything under section are settings
url = s3://bucket/name 
...
```

And change https://github.com/iterative/dvc/blob/fe635a5040c9d593cd1bdef3fa31f4df7af85259/dvc/remote/base.py#L85 to
```
def __init__(self, repo, settings):
```
So it means that component (**Remote**) can be initialized with settings which have priority1 and use config if needs as fallback rather then use terms 2 configs where the one is a child of the second
What do you think?

tests/func/test_remote.py
JIoJIaJIu: > Could you rewrite your tests to pytest?

Have I got right your request?
https://github.com/JIoJIaJIu/dvc/commit/42c34d7269c24187a09bbebe0469647862a75fdd#

dvc/remote/base.py
```diff
@@ -86,11 +86,7 @@ def __init__(self, repo, config):
         self.repo = repo
 
         self._check_requires(config)
-
-        core = config.get(Config.SECTION_CORE, {})
-        self.checksum_jobs = core.get(
-            Config.SECTION_CORE_CHECKSUM_JOBS, self.CHECKSUM_JOBS
-        )
+        self.checksum_jobs = self._get_checksum_jobs(repo, config)
```
efiop: No need to pass `repo`, you have `self.repo` set already.

dvc/remote/base.py
```diff
@@ -142,6 +138,18 @@ def _check_requires(self, config):
...
 
+    def _get_checksum_jobs(self, repo, config):
+        checksum_jobs = config.get(Config.SECTION_REMOTE_CHECKSUM_JOBS)
+        if not checksum_jobs:
+            try:
+                core = repo.config.config.get(Config.SECTION_CORE)
+                checksum_jobs = core.get(
+                    Config.SECTION_CORE_CHECKSUM_JOBS, self.CHECKSUM_JOBS
+                )
+            except (AttributeError, ConfigError):
```
efiop: Could you please clarify what `AttributeError` is for?

tests/func/test_remote.py
```diff
@@ -170,6 +170,41 @@ def test(self):
...
+    assert ret == 0
+    ret = main(["config", "core.checksum_jobs", "200"])
+    assert ret == 0
+
+    dvc.config.load()
+    remote = Remote(dvc, name="without_checksum_jobs")
+    assert remote.checksum_jobs == 200
+
+
+def test_with_checksum_default(dvc):
```
efiop: We could do these tests as unit ones in tests/unit/remote/test_base.py, which should be much simpler and lightweight.

dvc/remote/base.py
efiop: @JIoJIaJIu Got it, thanks! Actually, thinking about it, if `checksum_jobs` didn't work anyway, we could consider just getting rid of it, as it doesn't make much sense in many cases. E.g. if core.checksum_jobs is set for 200, then it makes sense for local remote or s3, but will break everything for ssh remote as we will run out of connections. But that will break the backward compatibility, so not worth bothering with right now. So your current solution is correct, let's stick with it for now 👍

efiop:
@JIoJIaJIu Looks good!

Good point about `settings`! Could you please create an issue for that?

dvc/remote/base.py
JIoJIaJIu: The reason why I decided to pass **repo** as argument for avoiding initialization dependency - otherwise needs to care that `self.repo` is initialized above. But I'm not sure that it's a good way

efiop: Sorry, not sure I follow. Could you elaborate/rephrase, please?

dvc/remote/base.py
JIoJIaJIu: there are places in **tests** that pass `None` or custom `Repo()` to `RemoteBASE` so such way I am checking that case.
https://github.com/JIoJIaJIu/dvc/blob/42c34d7269c24187a09bbebe0469647862a75fdd/tests/unit/remote/test_gdrive.py#L27

efiop: Why not `if self.repo` then? Attribute error might be too generic, as it will catch something that might happen deeper. Your first implementation had both `if self.repo` as well as `except AttributeError`, right? Let's keep only `if self.repo` then.

tests/func/test_remote.py
efiop: Just to clarify, this is totally mergable, but could be improved. So your call.

dvc/remote/base.py
JIoJIaJIu: I meant that needs first init `self.repo` like
```
def __init__(...):
   self.repo = ...
   self.checksum_jobs = ...
```
introduce complexity that `self.checksum_jobs` should be initialized after `self.repo`, so in terms of refactoring the order can change and it breaks the constructor

efiop: Well, but it will break in an obvious way with an attribute error, as `self.repo` won't be set. I would understand your logic if you would've made _get_checksum_jobs a static method or a helper function, but you've made it a full-blown method so using `self.repo` there would be totally fine. But current implementation works too. I guess another way to make it obvious would be to use `_get_checksum_jobs(repo.config if repo else {}, config)` 🙂But, again, these are non-vital details and neat-picking, so your call 🙂 Current implementation for this is mergeable.

dvc/remote/base.py
JIoJIaJIu: > Why not if self.repo then

Cause here there is a repo but not config, https://github.com/JIoJIaJIu/dvc/blob/42c34d7269c24187a09bbebe0469647862a75fdd/tests/unit/remote/test_gdrive.py#L27
I didn't dare to touch the test 
so I need to check `self.repo` and `self.repo` can be without config. I can change it to
```
self.repo.get('config', {})
```
Should I?

tests/func/test_remote.py
JIoJIaJIu: I sounds as a call :) so I will rework it

dvc/remote/base.py
efiop: That specific test will still have repo.config though, because of https://github.com/iterative/dvc/blob/master/dvc/repo/__init__.py#L82 . Maybe some other test is not mocking it properly, then it should be fixed, unless it is unreasonably hard, of course 🙂

dvc/remote/base.py
mroutis: @efiop , so we are staying with both `core` and `remote` for backward compatibility issues.

dvc/remote/base.py
mroutis: @JIoJIaJIu , I thought I had left a comment about `AttributeError`, but I couldn't find it.
Do you remember if I posted it? Am I going cuckoo :see_no_evil: ?

tests/func/test_remote.py
```diff
@@ -170,6 +170,41 @@ def test(self):
         self.assertEqual(ret, 0)
 
 
+def test_with_checksum_jobs(dvc):
+    ret = main(["remote", "add", "with_checksum_jobs", "s3://bucket/name"])
+    assert ret == 0
```
mroutis: I would in-line this as `assert 0 == main(...)`

tests/func/test_remote.py
mroutis: @JIoJIaJIu, you did) Thanks!

dvc/remote/base.py
efiop: @mroutis Yep, at least for now.

dvc/remote/base.py
efiop: @mroutis https://github.com/iterative/dvc/pull/3133#discussion_r366151735

JIoJIaJIu: For avoiding checking `config` attribute existence I extended `Repo` class in `GDrive` tests
https://github.com/JIoJIaJIu/dvc/commit/aec7781c7cc9f7d0fa4b61ce7d70c60c4f4f15ba#diff-b780ac32bee8290565619b80d7051c85R18
Could you please double check that it doesn't break anything?

JIoJIaJIu: @mroutis [I've resolved](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/commenting-on-a-pull-request#resolving-conversations) it marking that the changes were applied. Is it ok or better to keep all threads open?

dvc/remote/base.py
```diff
@@ -142,6 +138,22 @@ def _check_requires(self, config):
...
+        if checksum_jobs:
+            return checksum_jobs
+
+        if self.repo:
+            try:
+                core = self.repo.config.config.get(Config.SECTION_CORE, {})
+                return core.get(
+                    Config.SECTION_CORE_CHECKSUM_JOBS, self.CHECKSUM_JOBS
+                )
+            except ConfigError:
```
efiop: Hm, what would raise ConfigError though? Looks like you are accessing the dics directly.

dvc/remote/base.py
mroutis: @JIoJIaJIu, it's fine to resolve them, I do the same) Just couldn't find it this time :sweat_smile:

efiop:
@JIoJIaJIu Oops, looks like we forgot about docs. Could you please either create an issue in dvc.org or, even better, submit a PR to dvc.org?

Issue #3084, opened 2020-01-08
perf: copy cache operations very slow on network drive


rxxg:
Execute the following script on a Windows network drive
```
mkdir repo
cd repo
git init --quiet
dvc init -q
dvc config cache.type copy
measure-command { fsutil file createnew data 134217728 }
# Probably less than 1 second
measure-command { dvc add data }
# Around 30 seconds on our setup
```
Digging into the problem a little bit it seems that the cache copy operations end up in `shutil.copyfileobj` which reads the file to be copied into memory in 16kb chunks before writing out again.  Unless the network is very local this is always going to be a performance killer.

The situation might be better with Python 3.8 (#3033), but it would be good to ease the pain until DVC supports this version, and even then some users will not be able to upgrade straightaway.

efiop:
Hi @rxxg !

How did you reach the conclusion about blocksize in `shutil.copyfileobj`? Also, could you please elaborate on what will happen in 3.8 with it? Sorry, not aware of those upcoming changes 🙂 

Also, it is important to note that dvc is not only copying the file, but also is computing an md5 for it to be able to facilitate it in its cache. Could you try running md5sum(not sure what is the alternative command for windows) and time it to see how much time it takes compared to `dvc add`? And also timing `cp data data_copy` would be interesting to see as well, just so we could understand better the situation.

rxxg:
Hi @efiop,
I just looked at the [3.7 source code for shutil.copyfile and copyfileobj](https://github.com/python/cpython/blob/3.7/Lib/shutil.py) - by default the chunks are 16*1024 bytes.
There were a number of patches in this area for 3.8, for example [issue 33671](https://bugs.python.org/issue33671), [issue 36103](https://bugs.python.org/issue36103) and [issue 36610](https://bugs.python.org/issue36610) which mean that the [current code](https://github.com/python/cpython/blob/3.8/Lib/shutil.py) is much more complete.

A few small projects exist to add or complement this functionality - e.g. [pyfastcopy](https://pypi.org/project/pyfastcopy/) and [speedcopy](https://github.com/antirotor/speedcopy).  I have started playing with `speedcopy` since I am interested in Windows network drive performance but it is early code and doesn't seem to work with relative paths so it will need patching to work for DVC.

The ideal for me would be to have full server-side copy - although I am not sure whether my file servers actually support this!

rxxg:
Here are the detailed perf figures for my SSD and the network drive
```
[...]

PR #3095, opened 2020-01-09
Perf improvement for Windows share

rxxg:
* [X] ❗ Have you followed the guidelines in the [Contributing to DVC](https://dvc.org/doc/user-guide/contributing/core) list?

* [X] 📖 Check this box if this PR **does not** require [documentation](https://dvc.org/doc) updates, or if it does **and** you have created a separate PR in [dvc.org](https://github.com/iterative/dvc.org) with such updates (or at least opened an issue about it in that repo). Please link below to your PR (or issue) in the [dvc.org](https://github.com/iterative/dvc.org) repo.

* [X] ❌ Have you checked DeepSource, CodeClimate, and other sanity checks below? We consider their findings recommendatory and don't expect everything to be addressed. Please review them carefully and fix those that actually improve code or fix bugs.

Thank you for the contribution - we'll try to review it as soon as possible. 🙏

Improves performance for Windows network shares (#3084).
Be aware, it uses a [patched version of the external library speedcopy](https://github.com/rxxg/speedcopy/tree/5), so I don't expect you will want to pull as-is.

Also - my local tests (Windows machine) showed 17 failures, but no new ones from the base case.

Issue #3033, opened 2020-01-01
support python 3.8


mroutis:
It is officially released, but we are waiting for [pyarrow](https://pypi.org/project/pyarrow/#files) to release their wheels ([issue](https://issues.apache.org/jira/projects/ARROW/issues/ARROW-6955?filter=allopenissues)).

There's still some other projects that are not supporting Python 3.8 -- https://pyreadiness.org/3.8/

efiop:
Related https://github.com/iterative/dvc/pull/2912

PR #3135, opened 2020-01-14
Perf improvement for file copies

rxxg:

### DIFF
--- a/dvc/system.py
+++ b/dvc/system.py
@@ -1,7 +1,7 @@
 import errno
 import logging
 import os
-import speedcopy
+import platform
 
 from dvc.compat import fspath
 from dvc.exceptions import DvcException
@@ -18,7 +18,20 @@ class System(object):
     @staticmethod
     def copy(src, dest):
         src, dest = fspath(src), fspath(dest)
-        return speedcopy.copyfile(src, dest)
+        system = platform.system()
+        if system == "Windows":
+            import speedcopy
+
+            return speedcopy.copyfile(src, dest)
+        else:
+            import shutil
+            import sys
+
+            if sys.version_info < (3, 8):
+                # Importing the module monkey-patches shutil.copyfile
+                import pyfastcopy  # noqa: F401
+
+            return shutil.copyfile(src, dest)
 
     @staticmethod
     def hardlink(source, link_name):
@@ -90,8 +103,6 @@ class System(object):
 
     @staticmethod
     def reflink(source, link_name):
-        import platform
-
         source, link_name = fspath(source), fspath(link_name)
 
         system = platform.system()
--- a/dvc/system.py
+++ b/dvc/system.py
@@ -1,7 +1,7 @@
 import errno
 import logging
 import os
-import shutil
+import speedcopy
 
 from dvc.compat import fspath
 from dvc.exceptions import DvcException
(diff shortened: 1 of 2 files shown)


rxxg:
Fixes #3084.  Replaces #3095.  Should be revisited and perf characteristics checked on various systems when Python 3.8 compatibility arrives (#3033).

Probably should not merge until upstream project ([speedcopy](https://github.com/antirotor/speedcopy)) produces official release.

setup.py
```diff
@@ -75,6 +75,8 @@ def run(self):
     "win-unicode-console>=0.5; sys_platform == 'win32'",
     "pywin32>=225; sys_platform == 'win32'",
     "networkx>=2.1,<2.4",
+    "speedcopy @ git+https://github.com/rxxg/speedcopy@5",
```
efiop: As @rxxg noted, need to wait for official release here, as otherwise it will break our pypi package, as it doesn't support direct gh dependencies 🙁

rxxg: Version 2.0.1 of speedcopy has been released and is available on pypi

dvc/system.py
```diff
@@ -18,7 +18,20 @@ def is_unix():
     @staticmethod
     def copy(src, dest):
         src, dest = fspath(src), fspath(dest)
-        return shutil.copyfile(src, dest)
+        system = platform.system()
+        if system == "Windows":
+            import speedcopy
+
+            return speedcopy.copyfile(src, dest)
+        else:
```
efiop: This `else` is not needed since there is a `return` above.

setup.py
```diff
@@ -75,6 +75,8 @@ def run(self):
     "win-unicode-console>=0.5; sys_platform == 'win32'",
     "pywin32>=225; sys_platform == 'win32'",
     "networkx>=2.1,<2.4",
+    "speedcopy>=2.0.1",
+    "pyfastcopy>=1.0.3",
```
efiop: Both packages are not available in conda, so this PR will break our conda package 🙁Would you be willing to submit and maintain those packages in conda? Another simpler option is to make these packages optional, so that our System.copy can work with or without those packages. So pypi would be faster than conda package.

efiop: It is not a dealbreaker, as we could help out with those packages too, we have quite some experience with conda-forge already. Just asking. 🙂

Suor: Also, these deps should be conditional, i.e. none are needed in Python 3.8/Linux

dvc/system.py
```diff
@@ -18,7 +18,20 @@ def is_unix():
     @staticmethod
     def copy(src, dest):
         src, dest = fspath(src), fspath(dest)
-        return shutil.copyfile(src, dest)
+        system = platform.system()
+        if system == "Windows":
```
Suor: Maybe do that at "compile time":
```python
if platform.system() == "Windows":
    def copy(src, dest):
        # ...
elif:
    # ... 
```

setup.py
rxxg: Hi @efiop, I won't have enough time to maintain conda packages as well, I will let you guys make the call as to whether that is worthwhile - not forgetting that this code will potentially be deleted under Python 3.8.

I have updated the patch to fall back to standard code if packages are missing.

efiop: @rxxg No worries. Thanks for the PR! We'll try to create and maintain those conda packages ourselves.

Issue #2878, opened 2019-12-01
Refactor test/remotes.py by merging inside remote classes


skshetry:
#2853 was just merged, which refactored different helper functions into `test/remotes.py`. There are different remote classes (such as S3, Azure, etc) and functions such as `get_aws_url()`/`get_gcp_url`. 

It'd be good to merge them inside their respective remote classes. :)

skshetry:
I have started working on it. :)

skshetry:
I haven't been able to continue working on this. I'll continue working on this after Christmas. Sorry for not being quick on this. :slightly_smiling_face:

efiop:
@skshetry No worries 🙂 Thanks for the heads up!

PR #3136, opened 2020-01-14
test: use helper remotes in TestCases

skshetry:
This is what I had in mind when refactoring #2878. This is still a work in progress at the moment though.

### TODO
- [ ] Run all 9 tests to see if I messed up anything. :)

---

### DIFF
--- a/tests/func/test_data_cloud.py
+++ b/tests/func/test_data_cloud.py
@@ -77,17 +77,19 @@ class TestDataCloudBase(TestDvc):
     def _get_cloud_class(self):
         return None
 
-    def _should_test(self):
+    @staticmethod
+    def should_test():
         return False
 
-    def _get_url(self):
-        return ""
+    @staticmethod
+    def get_url():
+        return NotImplementedError
 
     def _get_keyfile(self):
         return None
 
     def _ensure_should_run(self):
-        if not self._should_test():
+        if not self.should_test():
             raise SkipTest(
                 "Test {} is disabled".format(self.__class__.__name__)
             )
@@ -95,7 +97,7 @@ class TestDataCloudBase(TestDvc):
     def _setup_cloud(self):
         self._ensure_should_run()
 
-        repo = self._get_url()
+        repo = self.get_url()
         keyfile = self._get_keyfile()
 
         config = copy.deepcopy(TEST_CONFIG)
@@ -188,25 +190,16 @@ class TestDataCloudBase(TestDvc):
         self._test_cloud()
 
 
-class TestRemoteS3(TestDataCloudBase):
-    def _should_test(self):
-        return S3.should_test()
-
-    def _get_url(self):
-        return S3.get_url()
-
+class TestRemoteS3(S3, TestDataCloudBase):
     def _get_cloud_class(self):
         return RemoteS3
 
 
-class TestRemoteGDrive(TestDataCloudBase):
-    def _should_test(self):
-        return GDrive.should_test()
-
+class TestRemoteGDrive(GDrive, TestDataCloudBase):
     def _setup_cloud(self):
         self._ensure_should_run()
 
-        repo = self._get_url()
+        repo = self.get_url()
 
         config = copy.deepcopy(TEST_CONFIG)
         config[TEST_SECTION][Config.SECTION_REMOTE_URL] = repo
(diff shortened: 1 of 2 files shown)


tests/func/test_data_cloud.py
```diff
@@ -77,12 +77,20 @@ class TestDataCloudBase(TestDvc):
...
+    @staticmethod
+    def get_url():
         return ""
 
+    def _should_test(self):
+        return self.should_test.__get__(None, object)()
+
+    def _get_url(self):
+        return self.get_url.__get__(None, object)()
+
```
efiop: This looks a bit weird. Maybe we could simply use `self.should_test` and `get_url` directly?

tests/func/test_data_cloud.py
```diff
@@ -247,41 +240,23 @@ def _setup_cloud(self):
...
 
-class TestRemoteLOCAL(TestDataCloudBase):
-    def _should_test(self):
-        return Local.should_test()
-
+class TestRemoteLOCAL(Local, TestDataCloudBase):
     def _get_url(self):
-        self.dname = Local.get_url()
+        self.dname = super()._get_url()
         return self.dname
```
efiop: Oh, this one doesn't match directly with LOCAL.get_url 🙁 I guess the line that uses `dname` down below is a pretty useless test, but just not to dive into it, we could simply wrap `super().get_url` here.

tests/remotes.py
```diff
@@ -42,9 +42,11 @@
 )
 TEST_GDRIVE_CLIENT_SECRET = "2fy_HyzSwkxkGzEken7hThXb"
 
+always_test = staticmethod(lambda: True)
```
efiop: Neat!

tests/func/test_data_cloud.py
skshetry: `should_test()` and `get_url()` are `staticmethod`, so, we cannot do `self.should_test()`, and I wanted to use it (i.e. `S3`, `GCP` etc. helper classes) as a mixin on inherited classes downward. I'm still looking into other options.

skshetry: Ah, it worked. Don't know what I changed. I was getting `requires 0 arguments but given 1` when doing `self.should_test()`. :confused:

skshetry: Found it. Before, `should_test` was a lambda function, but, as I've changed it to `staticmethod`, it works. :sweat_smile:

tests/func/test_data_cloud.py
```diff
@@ -73,29 +73,31 @@ def test(self):
...
-class TestDataCloudBase(TestDvc):
+class DataCloudBaseTestCase(TestDvc):
     def _get_cloud_class(self):
         return None
 
-    def _should_test(self):
-        return False
+    @staticmethod
+    def should_test():
+        raise NotImplementedError
```
efiop: Looks like some of our tests depend on this returning False by default. You could do the same here for now. The patch looks great overall!

skshetry: This *base* class has `test` method, that's why it always runs (was running before as well but would get skipped). Other tests are not failing. It's better I just fix this somehow. :)

efiop: @skshetry Your call :) But you've definitely tapped into the part of our tests that require reimplementation the most 😄 So just manage your effort and don't spend too much time, as they will be refactored later for sure. Your PR is still great though, as it makes our old tests use new helpers, which it will continue using even when refactored later.

tests/func/test_data_cloud.py
```diff
@@ -73,29 +73,31 @@ def test(self):
...
-    def _should_test(self):
+    @staticmethod
+    def should_test():
         return False
 
-    def _get_url(self):
-        return ""
+    @staticmethod
+    def get_url():
+        return NotImplementedError
```
skshetry: I still want to keep this, so, if `get_url` is not defined in inherited classes, we will know. :)

tests/func/test_data_cloud.py
skshetry: Done as @efiop proposed.

tests/func/test_data_cloud.py
efiop: @skshetry makes sense! 👍

Issue #3087, opened 2020-01-08
import/get directory with git files and dvc outputs


JPFrancoia:
DVC version: 0.80.0
DVC installed with Brew, on Mac OS

I'm trying to import/fetch external data.
I just did a:

```
dvc import https://github.com/my_company/my_repo data
```

which gives me a `data` folder and a `data.dvc` files in my current folder. Now, I'd like to pull the files contained in the `data` folder. The files are stored in a S3 bucket.

```
dvc pull -R data.dvc
```

But I immediately get this error:

```
ERROR: unexpected error - 'rev_lock'
```

Am I using the right workflow? How do I solve this issue?

Cheers

efiop:
Hi @JPFrancoia !

Could you show log for `dvc pull -R data.dvc -v`, please? And `dvc version` log too, please.

JPFrancoia:
Hi @efiop , 

Here is the log for the pull:

```
DEBUG: PRAGMA user_version;
DEBUG: fetched: [(3,)]
DEBUG: CREATE TABLE IF NOT EXISTS state (inode INTEGER PRIMARY KEY, mtime TEXT NOT NULL, size TEXT NOT NULL, md5 TEXT NOT NULL, timestamp TEXT NOT NULL)
DEBUG: CREATE TABLE IF NOT EXISTS state_info (count INTEGER)
DEBUG: CREATE TABLE IF NOT EXISTS link_state (path TEXT PRIMARY KEY, inode INTEGER NOT NULL, mtime TEXT NOT NULL)
DEBUG: INSERT OR IGNORE INTO state_info (count) SELECT 0 WHERE NOT EXISTS (SELECT * FROM state_info)
DEBUG: PRAGMA user_version = 3;
DEBUG: SELECT count from state_info WHERE rowid=?
DEBUG: fetched: [(12202,)]
DEBUG: UPDATE state_info SET count = ? WHERE rowid = ?
ERROR: unexpected error - 'rev_lock'
------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/Cellar/dvc/0.80.0/libexec/lib/python3.7/site-packages/dvc/main.py", line 48, in main
    ret = cmd.run()
  File "/usr/local/Cellar/dvc/0.80.0/libexec/lib/python3.7/site-packages/dvc/command/data_sync.py", line 30, in run
    recursive=self.args.recursive,
[...]

Issue #68, opened 2017-05-08
feature request: add option for multiple data directories


ophiry:
currently (as much as I understand) you can configure only one directory that is tracked in dvc
it would be useful to be able to specify more then one directory (for repositories that contain more than one project)

dmpetrov:
@ophiry I thought about that but don't have clear requirements so far. Could you please give more details?

A few possible solutions:
1) Why are two directories not enough? For example: `data/raw` and `data/db_dump`?
2) Would you like to use these two directories as completely separate projects\repositories?
Let say I work on `data/db_dump` and you share `data/db_dump` with me and you have an additional repo `data/raw`.

What is your scenario?

ophiry:
The scenario is adding the data files to a repository with multiple projects
proj1
proj2
...

I prefer to have the data files of each project as a sub folder of the project folder (which is the current standard for all other project files, like configurations, small data files, etc.)
proj1/data
proj2/data
...

PR #3150, opened 2020-01-14
status: implement support for imported files

### LINKED ISSUE #2959: status: implement support for import-ed files
See https://github.com/iterative/dvc/pull/2889#discussion_r357950906 for context

fabiosantoscode:
This fix only works when the remote is located in your machine. For remote DVC projects, it doesn't work and I'm looking into why.

Also I'm having a lot of trouble writing a unit test for this. The CLI works fine but when I call `dvc.status(with_deps=True)` in the test it reports that nothing changed.

Fixes #2959

------

:point_up: `dvc lock` docs imply that locked files are impervious to `dvc status`.

:point_up: I changed some code which already had DeepSource antipatterns and it re-detected them.

### DIFF
--- a/dvc/dependency/repo.py
+++ b/dvc/dependency/repo.py
@@ -115,7 +115,6 @@ class DependencyREPO(DependencyLOCAL):
     def status(self):
         current_fileinfo = self._get_fileinfo(updated=False)
         updated_fileinfo = self._get_fileinfo(updated=True)
-        print("CURRENT", current_fileinfo, "UPDATED", updated_fileinfo)
 
         if current_fileinfo["checksum"] and updated_fileinfo["checksum"]:
             has_changed = (
--- a/dvc/dependency/repo.py
+++ b/dvc/dependency/repo.py
@@ -1,6 +1,7 @@
 import copy
 import os
 from contextlib import contextmanager
+import filecmp
 
 from funcy import merge
 
@@ -10,6 +11,7 @@ from dvc.external_repo import external_repo
 from dvc.exceptions import NotDvcRepoError
 from dvc.exceptions import OutputNotFoundError
 from dvc.exceptions import PathMissingError
+from dvc.exceptions import NoOutputInExternalRepoError
 from dvc.utils.fs import fs_copy
 
 
@@ -46,14 +48,29 @@ class DependencyREPO(DependencyLOCAL):
             yield repo
 
     def status(self):
-        with self._make_repo() as repo:
-            current = repo.find_out_by_relpath(self.def_path).info
+        try:
+            with self._make_repo() as repo:
+                current = repo.find_out_by_relpath(self.def_path).info
 
-        with self._make_repo(rev_lock=None) as repo:
-            updated = repo.find_out_by_relpath(self.def_path).info
+            with self._make_repo(rev_lock=None) as repo:
+                updated = repo.find_out_by_relpath(self.def_path).info
 
-        if current != updated:
-            return {str(self): "update available"}
+            if current != updated:
+                return {str(self): "update available"}
+
+        except NoOutputInExternalRepoError:
+            url = self.def_repo["url"]
+            old_clone = cached_clone(url, rev=self.def_repo.get("rev_lock"))
+            new_clone = cached_clone(url)
+            old_file = os.path.join(old_clone, self.def_path)
+            new_file = os.path.join(new_clone, self.def_path)
+
+            if not os.path.exists(old_file) or not os.path.exists(new_file):
+                raise
+
+            files_differ = not filecmp.cmp(old_file, new_file)
+            if files_differ:
+                return {str(self): "update available"}
 
         return {}
 
--- a/dvc/dependency/repo.py
+++ b/dvc/dependency/repo.py
@@ -11,7 +11,6 @@ from dvc.external_repo import external_repo
 from dvc.exceptions import NotDvcRepoError
 from dvc.exceptions import OutputNotFoundError
 from dvc.exceptions import PathMissingError
-from dvc.exceptions import NoOutputInExternalRepoError
 from dvc.utils.fs import fs_copy
 
 
@@ -48,29 +47,23 @@ class DependencyREPO(DependencyLOCAL):
             yield repo
 
     def status(self):
-        try:
-            with self._make_repo() as repo:
+        with self._make_repo() as repo, self._make_repo(
+            rev_lock=None
+        ) as updated_repo:
+            try:
                 current = repo.find_out_by_relpath(self.def_path).info
+                updated = updated_repo.find_out_by_relpath(self.def_path).info
 
-            with self._make_repo(rev_lock=None) as repo:
-                updated = repo.find_out_by_relpath(self.def_path).info
+                has_changed = current != updated
+            except OutputNotFoundError:
+                current_path = os.path.join(repo.root_dir, self.def_path)
+                updated_path = os.path.join(
+                    updated_repo.root_dir, self.def_path
+                )
+                has_changed = not filecmp.cmp(current_path, updated_path)
 
-            if current != updated:
-                return {str(self): "update available"}
-
-        except NoOutputInExternalRepoError:
-            url = self.def_repo["url"]
-            old_clone = cached_clone(url, rev=self.def_repo.get("rev_lock"))
-            new_clone = cached_clone(url)
-            old_file = os.path.join(old_clone, self.def_path)
-            new_file = os.path.join(new_clone, self.def_path)
-
-            if not os.path.exists(old_file) or not os.path.exists(new_file):
-                raise
-
-            files_differ = not filecmp.cmp(old_file, new_file)
-            if files_differ:
-                return {str(self): "update available"}
+        if has_changed:
+            return {str(self): "update available"}
 
         return {}
 
--- a/dvc/dependency/repo.py
+++ b/dvc/dependency/repo.py
@@ -53,35 +53,37 @@ class DependencyREPO(DependencyLOCAL):
             PathInfo(os.path.join(repo.root_dir, self.def_path))
         )
 
-    def _has_changed(self):
-        with self._make_repo() as repo, self._make_repo(
-            rev_lock=None
-        ) as updated_repo:
-            try:
-                current = repo.find_out_by_relpath(self.def_path).info
-                updated = updated_repo.find_out_by_relpath(self.def_path).info
+    # fileinfo is a dictionary containing "checksum" and "path"
+    def _get_fileinfo(self, updated=False):
+        rev_lock = None
+        if not updated:
+            rev_lock = self.def_repo.get(self.PARAM_REV_LOCK)
 
-                return current != updated
-            except OutputNotFoundError:
-                # Need to load the state before calculating checksums
-                repo.state.load()
-                updated_repo.state.load()
+        try:
+            with self._make_repo(rev_lock=rev_lock) as repo:
+                path = os.path.join(repo.root_dir, self.def_path)
+                try:
+                    checksum = repo.find_out_by_relpath(self.def_path).info[
+                        "md5"
+                    ]
+                except OutputNotFoundError:
+                    repo.state.load()
+                    checksum = repo.cache.local.get_checksum(PathInfo(path))
 
-                return self._get_checksum_in_repo(
-                    repo
-                ) != self._get_checksum_in_repo(updated_repo)
+                assert checksum
 
-    def _has_changed_non_dvc(self):
-        url = self.def_repo[self.PARAM_URL]
-        rev = self.def_repo.get(self.PARAM_REV)
-        rev_lock = self.def_repo.get(self.PARAM_REV_LOCK, rev)
+                return {"checksum": checksum, "path": path, "repo": repo}
 
-        current_repo = cached_clone(url, rev=rev_lock or rev)
-        updated_repo = cached_clone(url, rev=rev)
+        except NotDvcRepoError:
+            repo_path = cached_clone(
+                self.def_repo[self.PARAM_URL],
+                rev=rev_lock or self.def_repo.get(self.PARAM_REV),
+            )
+            path = os.path.join(repo_path, self.def_path)
+            return {"checksum": None, "path": path, "repo": None}
 
-        current_path = os.path.join(current_repo, self.def_path)
-        updated_path = os.path.join(updated_repo, self.def_path)
-
+    @staticmethod
+    def _paths_changed(current_path, updated_path):
         if not os.path.exists(current_path):
             raise FileNotFoundError(current_path)
 
@@ -102,11 +104,29 @@ class DependencyREPO(DependencyLOCAL):
 
         return not filecmp.cmp(current_path, updated_path, shallow=False)
 
+    def _checkout_if_needed(self, fileinfo):
+        if os.path.exists(fileinfo["path"]):
+            return
+        if "repo" not in fileinfo:
+            return
+
+        fileinfo["repo"].checkout([fileinfo["path"] + ".dvc"], recursive=True)
+
     def status(self):
-        try:
-            has_changed = self._has_changed()
-        except NotDvcRepoError:
-            has_changed = self._has_changed_non_dvc()
+        current_fileinfo = self._get_fileinfo(updated=False)
+        updated_fileinfo = self._get_fileinfo(updated=True)
+        print("CURRENT", current_fileinfo, "UPDATED", updated_fileinfo)
+
+        if current_fileinfo["checksum"] and updated_fileinfo["checksum"]:
+            has_changed = (
+                current_fileinfo["checksum"] != updated_fileinfo["checksum"]
+            )
+        else:
+            self._checkout_if_needed(current_fileinfo)
+            self._checkout_if_needed(updated_fileinfo)
+            has_changed = DependencyREPO._paths_changed(
+                current_fileinfo["path"], updated_fileinfo["path"]
+            )
 
         if has_changed:
             return {str(self): "update available"}
--- a/dvc/dependency/repo.py
+++ b/dvc/dependency/repo.py
@@ -1,7 +1,6 @@
 import copy
 import os
 from contextlib import contextmanager
-import filecmp
 
 from funcy import merge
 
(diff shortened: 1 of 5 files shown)


dvc/dependency/repo.py
```diff
@@ -46,13 +47,22 @@ def _make_repo(self, **overrides):
...
+                current = repo.find_out_by_relpath(self.def_path).info
+                updated = updated_repo.find_out_by_relpath(self.def_path).info
+
+                has_changed = current != updated
+            except OutputNotFoundError:
+                current_path = os.path.join(repo.root_dir, self.def_path)
+                updated_path = os.path.join(
+                    updated_repo.root_dir, self.def_path
+                )
+                has_changed = not filecmp.cmp(current_path, updated_path)
```
efiop: There is a gotcha in filecmp.cmp:
```suggestion
                has_changed = not filecmp.cmp(current_path, updated_path, shallow=False)
```
without it, it will only check sizes, I think.

Btw, these paths might be dirs and IIRC cmp() doesn't work with dirs. Maybe let's use md5 that we use for cached outputs? See `RemoteLOCAL.get_checksum()`, which you could invoke through `self.repo.cache.local.get_checksum()`.

dvc/dependency/repo.py
```diff
@@ -46,13 +47,22 @@ def _make_repo(self, **overrides):
...
-        if current != updated:
+        with self._make_repo() as repo, self._make_repo(
+            rev_lock=None
+        ) as updated_repo:
+            try:
+                current = repo.find_out_by_relpath(self.def_path).info
+                updated = updated_repo.find_out_by_relpath(self.def_path).info
+
+                has_changed = current != updated
+            except OutputNotFoundError:
```
efiop: Hm, there is also an interesting case, when a file/dir was git-tracked before but then became dvc-tracked or vice versa. But the thingy with get_checksum should make those comparable, I think.

dvc/repo/status.py
```diff
@@ -10,28 +10,32 @@
...
 
-def _local_status(self, targets=None, with_deps=False):
+def _joint_status(stages):
     status = {}
 
-    if targets:
-        stages = cat(self.collect(t, with_deps=with_deps) for t in targets)
-    else:
-        stages = self.collect(None, with_deps=with_deps)
-
```
efiop: I guess you've split these because of deep source, right? Nothing wrong with that, just asking 🙂

fabiosantoscode: That's exactly it :)

dvc/dependency/repo.py
fabiosantoscode: Thanks for pointing out where I can find a checksum function :) I already suspected filecmp wasn't good for this.

tests/func/test_update.py
```diff
@@ -19,8 +19,13 @@ def test_update_import(tmp_dir, dvc, erepo_dir):
...
     clean_repos()
 
-    assert dvc.status([stage.path]) == {}
+    status, = dvc.status([stage.path])["version.dvc"]
+    changed_dep, = list(status["changed deps"].items())
+    assert changed_dep[0].startswith("version ")
+    assert changed_dep[1] == "update available"
+
     dvc.update(stage.path)
+
```
efiop: This one only tests cached files, but we actually have a bug described in https://github.com/iterative/dvc/issues/3087#issuecomment-574267505 that will make the test break for git files 🙁I've submitted https://github.com/iterative/dvc/pull/3151 , that will unblock you.

efiop: @fabiosantoscode Could you please add a cross test though? The one where file gets updated from being git tracked to dvc tracked.

fabiosantoscode: So that's why rev_lock was showing up inconsistently on my test files :)

efiop: @fabiosantoscode Merged and unblocked. Thank you for your patience 🙂

fabiosantoscode: Super cool @efiop :) 

So this cross-test.

I think you mean I need to add a test that imports from a DVC project a file that's not under DVC (but is under git), then track the file under DVC and do an update?

Or is it an import before `dvc init` and an update after?

Maybe both? :)

efiop: @fabiosantoscode Good question. What I meant is:
1) import git-tracked `file` from erepo
2) make `file` dvc-tracked in erepo, create a commit
3) run `dvc update`

So your first description, looks like.

> Or is it an import before dvc init and an update after?

But that is indistinguishable from the previous description from dvc's point of view 🙂Unless I'm missing something. But the scenario is more believable than mine, for sure.

fabiosantoscode: @efiop after I got the first one passing, I tried the second one just to see if it passed.

It failed, and revealed that this PR doesn't allow `status` for imported files from non-git repositories.

So will need to do more stuff :) but it's getting late. Will do some more of this tomorrow!

fabiosantoscode:
@efiop now that was quite fun :)

I got `dvc status` for non-DVC remote imports working. The cross-cases are now working as well, since the files are treated individually. In the case that both have an md5 we can use, that md5 is used. If they don't, it's a matter of doing a deep copy. If one of the versions is tracked by DVC but not the other, a `dvc checkout` is performed beforehand. Hopefully they're both small! If they're not, I think this is enough of a corner case to not matter.

efiop:
Thank you for your patience 🙂 

> The case where a file turns into a directory (or vise-versa) is treated with an assert. I'm not sure if we're just supposed to say "hey, this changed" or raise a specific exception in this case.

@fabiosantoscode Hm, but why is that an issue though? 🤔

dvc/dependency/repo.py
```diff
@@ -46,14 +48,86 @@ def _make_repo(self, **overrides):
...
+        try:
+            with self._make_repo(rev_lock=rev_lock) as repo:
+                path = os.path.join(repo.root_dir, self.def_path)
+                try:
+                    checksum = repo.find_out_by_relpath(self.def_path).info[
+                        "md5"
+                    ]
+                except OutputNotFoundError:
+                    repo.state.load()
+                    checksum = repo.cache.local.get_checksum(PathInfo(path))
```
efiop: @fabiosantoscode Please don't `load` it like that without closing after. We are always treating it as:
`with repo.state`, you should do the same here.

dvc/dependency/repo.py
```diff
@@ -46,14 +48,86 @@ def _make_repo(self, **overrides):
...
+        if "repo" not in fileinfo:
+            return
+
+        fileinfo["repo"].checkout([fileinfo["path"] + ".dvc"], recursive=True)
+
     def status(self):
-        with self._make_repo() as repo:
-            current = repo.find_out_by_relpath(self.def_path).info
+        current_fileinfo = self._get_fileinfo(updated=False)
+        updated_fileinfo = self._get_fileinfo(updated=True)
```
efiop: Why `file` though? It could be a directory, right?

dvc/dependency/repo.py
```diff
@@ -46,14 +48,86 @@ def _make_repo(self, **overrides):
...
-        with self._make_repo(rev_lock=None) as repo:
-            updated = repo.find_out_by_relpath(self.def_path).info
+        if current_fileinfo["checksum"] and updated_fileinfo["checksum"]:
+            has_changed = (
+                current_fileinfo["checksum"] != updated_fileinfo["checksum"]
+            )
+        else:
+            self._checkout_if_needed(current_fileinfo)
+            self._checkout_if_needed(updated_fileinfo)
+            has_changed = DependencyREPO._paths_changed(
```
efiop: Why not
```suggestion
            has_changed = self._paths_changed(
```
?

dvc/dependency/repo.py
```diff
@@ -46,14 +48,86 @@ def _make_repo(self, **overrides):
...
+            return {"checksum": None, "path": path, "repo": None}
+
+    @staticmethod
+    def _paths_changed(current_path, updated_path):
+        if not os.path.exists(current_path):
+            raise FileNotFoundError(current_path)
+
+        if not os.path.exists(updated_path):
+            raise FileNotFoundError(updated_path)
+
```
efiop: Isn't `path` the same though? Except for location of a cloned repo, sure, but this is strange that you treat them as being completely different. Also, why not check existance and raise in _getfileinfo?

Also note that you can't really use existing paths when you take them out of `with _make_repo` contest. as it might be deleted on `__exit__`. It is not deleted now, but it is not nice to spread the context so much across your code.

dvc/dependency/repo.py
```diff
@@ -46,14 +48,86 @@ def _make_repo(self, **overrides):
...
+        if is_dir:
+            comparison = filecmp.dircmp(current_path, updated_path)
+            return not (
+                comparison.left_only
+                or comparison.right_only
+                or comparison.diff_files
+            )
+
+        return not filecmp.cmp(current_path, updated_path, shallow=False)
```
efiop: Not sure why are you using filecmp again, when you have `md5`s already, that you can directly compare.

dvc/dependency/repo.py
```diff
@@ -46,14 +48,86 @@ def _make_repo(self, **overrides):
...
+
+                return {"checksum": checksum, "path": path, "repo": repo}
+
+        except NotDvcRepoError:
+            repo_path = cached_clone(
+                self.def_repo[self.PARAM_URL],
+                rev=rev_lock or self.def_repo.get(self.PARAM_REV),
+            )
+            path = os.path.join(repo_path, self.def_path)
+            return {"checksum": None, "path": path, "repo": None}
```
efiop: You can toatlly `get_checksum()` a git file/dir, just use `self.repo.cache.local.get_checksum()`  instead of `repo.cache.local.get_checksum()`.

dvc/dependency/repo.py
```diff
@@ -46,14 +48,86 @@ def _make_repo(self, **overrides):
...
+        return not filecmp.cmp(current_path, updated_path, shallow=False)
+
+    def _checkout_if_needed(self, fileinfo):
+        if os.path.exists(fileinfo["path"]):
+            return
+        if "repo" not in fileinfo:
+            return
+
+        fileinfo["repo"].checkout([fileinfo["path"] + ".dvc"], recursive=True)
+
```
efiop: You don't need to checkout anything. In fact, usually you don't even have the data locally to checkout with, because you didn't update yet.

tests/func/test_update.py
```diff
@@ -51,6 +110,48 @@ def test_update_import(tmp_dir, dvc, erepo_dir, cached):
...
+        shutil.rmtree(".dvc")
+        erepo_dir.scm_gen("file", "first version")
+        erepo_dir.scm.add(["file"])
+        erepo_dir.scm.commit("first version")
+        old_rev = erepo_dir.scm.get_rev()
+
+    stage = dvc.imp(fspath(erepo_dir), "file", "file")
+
+    with erepo_dir.chdir():
+        Repo.init()
```
efiop: Hm, not sure why this is needed. erepo is already a git repo. Am I missing something?

dvc/stage.py
```diff
@@ -1006,10 +1006,12 @@ def _status(entries):
         return ret
 
     @rwlocked(read=["deps", "outs"])
-    def status(self):
+    def status(self, for_status_command=False):
```
efiop: Hm, the naming seems a bit cumbersome to me. How about `check_updates`?

dvc/dependency/repo.py
fabiosantoscode: I wanted to be explicit that this is a static method. But if that's the convention I'm happy to follow.

dvc/dependency/repo.py
fabiosantoscode: My reasoning for checking both is that the remote repository might not have it any more in the updated version. I checked the "past" one as a sanity check.

fabiosantoscode: Regarding the context, that one's going to be complicated to make readable :) I might have to open the contexts out in the `status` method so that they're still open and guaranteed to not be deleted.

dvc/dependency/repo.py
fabiosantoscode: @efiop if the target repository is not DVC we don't have an MD5. I would generate it but the functions that generate the MD5 currently require a working DVC repo to run.

fabiosantoscode: This is also for the case where `dvc import` is used on a non-DVC repository and then that repository us upgraded to DVC.

tests/func/test_update.py
fabiosantoscode: `Repo` here is DVC. I'm enabling DVC on the git repo. In the first part of the test I delete `.dvc` to test the case where an external repo gets `dvc init`'ed after an import.

dvc/stage.py
fabiosantoscode: Works for me!

dvc/dependency/repo.py
fabiosantoscode: However, this function may be called for this repo as well as with the "other repo" which has just been checked out and might not be DVC enabled.

efiop: @fabiosantoscode It doesn't has to be dvc enabled, we are using our current repo's facilities to compute checksum for a file/dir on filesystem, no matter if it is in git or dvc repo itself.

fabiosantoscode: Ooooooh, nice :) I'll do just that!

fabiosantoscode:
@efiop thanks a lot for your comments! I wouldn't have thought that it was possible to generate checksums without a DVC repo. The code seemed to require a repo.

This changes a lot, and the resulting code is way easier to understand.

dvc/dependency/repo.py
```diff
@@ -46,14 +47,37 @@ def _make_repo(self, **overrides):
         with external_repo(**merge(self.def_repo, overrides)) as repo:
             yield repo
 
-    def status(self):
-        with self._make_repo() as repo:
-            current = repo.find_out_by_relpath(self.def_path).info
+    # fileinfo is a dictionary containing "checksum" and "path"
```
efiop: No longer relevant, seems like?

dvc/dependency/repo.py
```diff
@@ -46,14 +47,37 @@ def _make_repo(self, **overrides):
...
-            updated = repo.find_out_by_relpath(self.def_path).info
+        try:
+            with self._make_repo(rev_lock=rev_lock) as repo:
+                path = os.path.join(repo.root_dir, self.def_path)
+
+                try:
+                    return repo.find_out_by_relpath(self.def_path).info["md5"]
+                except OutputNotFoundError:
+                    with repo.state:
+                        return self.repo.cache.local.get_checksum(
```
efiop: No need to take `repo.state` if you are using `self.repo` anyways. Also, maybe you could fallthrough here, same as in your first `get` patch?

dvc/dependency/repo.py
fabiosantoscode: True

dvc/dependency/repo.py
fabiosantoscode: Yes, and yes :)

fabiosantoscode:
@efiop done

dvc/dependency/repo.py
```diff
@@ -46,14 +47,33 @@ def _make_repo(self, **overrides):
...
 
-        with self._make_repo(rev_lock=None) as repo:
-            updated = repo.find_out_by_relpath(self.def_path).info
+        try:
+            with self._make_repo(rev_lock=rev_lock) as repo:
+                try:
+                    return repo.find_out_by_relpath(self.def_path).info["md5"]
+                except OutputNotFoundError:
+                    # Fall through to after exception handler
+                    path = os.path.join(repo.root_dir, self.def_path)
```
efiop: You have the same line in #68 ,why duplicating it?

efiop: Oh, I see you have a second except logic here.

fabiosantoscode: I'm wondering if it's good to use `repo` outside of the `_make_repo`.

dvc/dependency/repo.py
```diff
@@ -46,14 +47,33 @@ def _make_repo(self, **overrides):
...
+
+        except NotDvcRepoError:
+            repo_path = cached_clone(
+                self.def_repo[self.PARAM_URL],
+                rev=rev_lock or self.def_repo.get(self.PARAM_REV),
+            )
+            path = os.path.join(repo_path, self.def_path)
+
```
efiop: @fabiosantoscode Please take a look here https://github.com/iterative/dvc/blob/master/dvc/repo/get.py#L66 . You can do pretty much the same and won't have to duplicate logic or catch raw OutputNotFoundError

dvc/dependency/repo.py
efiop: @fabiosantoscode But that is not true, you are using `cached_clone` down below, so you are using it and not the `_make_repo`s part.

efiop: Though I can see the issue here, that is a gotcha for me 🙂

efiop:
@fabiosantoscode almost done here, just one minor comment left. Thank you for your patience 🙂

fabiosantoscode:
@efiop done, it looks way simpler now.

dvc/dependency/repo.py
```diff
@@ -46,14 +48,31 @@ def _make_repo(self, **overrides):
...
-        with self._make_repo(rev_lock=None) as repo:
-            updated = repo.find_out_by_relpath(self.def_path).info
+        try:
+            with self._make_repo(rev_lock=rev_lock) as repo:
+                return repo.find_out_by_relpath(self.def_path).info["md5"]
+        except (NotDvcRepoError, NoOutputInExternalRepoError):
+            # Fall through and clone
+            pass
+
+        repo_path = cached_clone(
```
efiop: Luckily we do have `git + dvc` clones refactor on our todo list 🙂Not a part of this PR or anything, it is just us messing up earlier.  😅

dvc/dependency/repo.py
```diff
@@ -46,14 +48,31 @@ def _make_repo(self, **overrides):
...
+            self.def_repo[self.PARAM_URL],
+            rev=rev_lock or self.def_repo.get(self.PARAM_REV),
+        )
+        path = PathInfo(os.path.join(repo_path, self.def_path))
+
+        return self.repo.cache.local.get_checksum(path)
```
Suor: We really shouldn't do this. If path happens to be a dir then it will add that dir listing to `self.repo` cache.

efiop: @Suor It is not an issue really. Those are tiny files and will get cleaned up on gc.

PR #3156, opened 2020-01-15
get: implement --show-url to display only url/path to remote

### LINKED ISSUE #2994: get: show url to cache
In some cases, users need a direct link to a data file in the cloud. It might be just a matter of convenience or when DVC is used from automation tools like CD4ML scenarios. We need to provide a way to get a direct link.

I suggest creating a new command `dvc resolve`. Other options can be also be considered. For example, it might be a part of `dvc status` (when it start supporting data files as argument).

```
$ dvc resolve file.txt df.csv
file.txt s3://dvc-temp/dataset1/2d10f00a05f1fe70eaa3c42aa9f44b95
df.csv s3://dvc-temp/dataset1/99370a5386cac00c93c9fdc836076a7d
```

It should support revisions (checksum, tag, branch or HEAD^^):
```
$ dvc resolve file.txt df.csv --rev ada9d973
file.txt s3://dvc-temp/dataset1/d30abac6950583e6a5f1dcd31cb1043b
df.csv gs://project42/cache/2290babda371e52eeca2a2065a358783
```

Note, the commands with `revision` options should resolve old remotes as well (see `s3://` and `gs://` above).

skshetry:
Closes #2994. 

From the product requirements, it deviates on the following issue:
1. For now, it does not support showing URLs for multiple paths.
   So, we can only do `dvc get <url> <path> --show-url` for a single path, so as 
   to make it compatible with `dvc get`.
2. It also does not support specifying remote at the moment, so as to not introduce new flags.
3. It does not work with the file inside of a directory (i.e. `dir/file`). Limitation of `dvc.api.get_url()`. Requires adding directory granularity support.
	```sh
	➜ dvc get . dir/file --show-url
	ERROR: failed to show url - unable to find DVC-file with output 'dir/file'
	```
4. How to show path/URL to the directory? Currently, it only shows `.dir` file.
	```sh
	➜ dvc get . dir --show-url
    ../tmp.fKEyzBMWge/12/11325135bf45fe5f67efa975110a57.dir
	```
   Same is the case with `dvc get` (cc @jorgeorpinel for docs regarding `dvc.api.get_url()`).
   
5. It does not work with non-dvc repos (i.e. git only repos). So, the following won't work:
   ```sh
   ➜ dvc get https://github.com/schacon/cowsay install.sh --show-url 
   ERROR: failed to show url - URL 'https://github.com/schacon/cowsay' is not a dvc repository.
   ```
6. Custom revision does not work for the local repo.
   Not fixing at the moment. Will create a separate issue for this.
	```sh
	➜ dvc get . foo --rev=HEAD@{5.days.ago} # works
	➜ dvc get . foo --rev=HEAD@{5.days.ago} --show-url
	ERROR: unexpected error - Custom revision is not supported for local repo
   ```

### Example Usage
```sh
➜ dvc get . foo --show-url
../tmp.uJVvn5ZaLY/d3/b07384d113edec49eaa6238ad5ff00
➜ dvc get https://github.com/iterative/dataset-registry.git get-started/data.xml --rev=HEAD@{5.days.ago} --show-url
https://remote.dvc.org/dataset-registry/a3/04afb96060aad90176268345e10355
```
EDIT: Made changes to format from `<path> <url>` previously to just `<url>`. 

---

#### Edgecases
- [x] https://github.com/iterative/dvc/pull/3156#discussion_r366942056

### DIFF
--- a/dvc/api.py
+++ b/dvc/api.py
@@ -10,7 +10,7 @@ import ruamel.yaml
 from voluptuous import Schema, Required, Invalid
 
 from dvc.repo import Repo
-from dvc.exceptions import DvcException
+from dvc.exceptions import DvcException, NotDvcRepoError
 from dvc.external_repo import external_repo
 
 
@@ -43,13 +43,27 @@ class SummonError(DvcException):
     pass
 
 
+class UrlNotDvcRepoError(DvcException):
+    """Thrown if given url is not a DVC repository.
+
+    Args:
+        url (str): url to the repository.
+    """
+
+    def __init__(self, url):
+        super().__init__("URL '{}' is not a dvc repository.".format(url))
+
+
 def get_url(path, repo=None, rev=None, remote=None):
     """Returns an url of a resource specified by path in repo"""
-    with _make_repo(repo, rev=rev) as _repo:
-        abspath = os.path.join(_repo.root_dir, path)
-        out, = _repo.find_outs_by_path(abspath)
-        remote_obj = _repo.cloud.get_remote(remote)
-        return str(remote_obj.checksum_to_path_info(out.checksum))
+    try:
+        with _make_repo(repo, rev=rev) as _repo:
+            abspath = os.path.join(_repo.root_dir, path)
+            out, = _repo.find_outs_by_path(abspath)
+            remote_obj = _repo.cloud.get_remote(remote)
+            return str(remote_obj.checksum_to_path_info(out.checksum))
+    except NotDvcRepoError:
+        raise UrlNotDvcRepoError(repo)
 
 
 def open(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
--- a/dvc/command/get.py
+++ b/dvc/command/get.py
@@ -3,21 +3,32 @@ import logging
 
 from .base import append_doc_link
 from .base import CmdBaseNoRepo
-from dvc.exceptions import DvcException
+from dvc.exceptions import DvcException, NotDvcRepoError
 
 
 logger = logging.getLogger(__name__)
 
 
 class CmdGet(CmdBaseNoRepo):
-    def _run_get_url(self):
+    def _show_url(self):
         from dvc.api import get_url
 
         try:
             url = get_url(
                 self.args.path, repo=self.args.url, rev=self.args.rev
             )
-            logger.info("{} {}".format(self.args.path, url))
+            logger.info(url)
+        except NotDvcRepoError:
+            # NOTE: there's no way to get direct links to files inside
+            # git repository.
+            logger.exception(
+                "Could not show url for '{}' from '{}'. ".format(
+                    self.args.path, self.args.url
+                )
+                + "Only DVC repository is supported currently.",
+                exc_info=None,
+            )
+            return 1
         except DvcException:
             logger.exception(
                 "failed to show url for '{}' from '{}'".format(
@@ -25,13 +36,16 @@ class CmdGet(CmdBaseNoRepo):
                 )
             )
             return 1
+
         return 0
 
     def run(self):
-
         if self.args.show_url:
-            return self._run_get_url()
+            return self._show_url()
 
+        return self._get_file_from_repo()
+
+    def _get_file_from_repo(self):
         from dvc.repo import Repo
 
         try:
--- a/dvc/command/get.py
+++ b/dvc/command/get.py
@@ -3,7 +3,7 @@ import logging
 
 from .base import append_doc_link
 from .base import CmdBaseNoRepo
-from dvc.exceptions import DvcException, NotDvcRepoError
+from dvc.exceptions import DvcException
 
 
 logger = logging.getLogger(__name__)
@@ -18,23 +18,8 @@ class CmdGet(CmdBaseNoRepo):
                 self.args.path, repo=self.args.url, rev=self.args.rev
             )
             logger.info(url)
-        except NotDvcRepoError:
-            # NOTE: there's no way to get direct links to files inside
-            # git repository.
-            logger.exception(
-                "Could not show url for '{}' from '{}'. ".format(
-                    self.args.path, self.args.url
-                )
-                + "Only DVC repository is supported currently.",
-                exc_info=None,
-            )
-            return 1
         except DvcException:
-            logger.exception(
-                "failed to show url for '{}' from '{}'".format(
-                    self.args.path, self.args.url
-                )
-            )
+            logger.exception("failed to show url")
             return 1
 
         return 0
--- a/dvc/command/get.py
+++ b/dvc/command/get.py
@@ -10,7 +10,28 @@ logger = logging.getLogger(__name__)
 
 
 class CmdGet(CmdBaseNoRepo):
+    def _run_get_url(self):
+        from dvc.api import get_url
+
+        try:
+            url = get_url(
+                self.args.path, repo=self.args.url, rev=self.args.rev
+            )
+            logger.info("{} {}".format(self.args.path, url))
+        except DvcException:
+            logger.exception(
+                "failed to show url for '{}' from '{}'".format(
+                    self.args.path, self.args.url
+                )
+            )
+            return 1
+        return 0
+
     def run(self):
+
+        if self.args.show_url:
+            return self._run_get_url()
+
         from dvc.repo import Repo
 
         try:
@@ -54,5 +75,10 @@ def add_parser(subparsers, parent_parser):
     )
     get_parser.add_argument(
         "--rev", nargs="?", help="Git revision (e.g. branch, tag, SHA)"
+    )
+    get_parser.add_argument(
+        "--show-url",
+        action="store_true",
+        help="Returns path/url to the location in remote for specified path",
     )
     get_parser.set_defaults(func=CmdGet)
--- a/tests/func/test_api.py
+++ b/tests/func/test_api.py
@@ -57,7 +57,8 @@ def test_get_url_git_only_repo_throws_exception(tmp_dir, scm):
     with pytest.raises(UrlNotDvcRepoError) as exc_info:
         api.get_url("foo", fspath(tmp_dir))
 
-    assert fspath(tmp_dir) in str(exc_info)
+    # On windows, `exc_info` has path escaped, eg: `C:\\\\Users\\\\travis`
+    assert fspath(tmp_dir) in str(exc_info).replace("\\\\", "\\")
 
 
 @pytest.mark.parametrize("remote_url", all_remote_params, indirect=True)
--- a/tests/func/test_api.py
+++ b/tests/func/test_api.py
@@ -6,7 +6,7 @@ import ruamel.yaml
 import pytest
 
 from dvc import api
-from dvc.api import SummonError
+from dvc.api import SummonError, UrlNotDvcRepoError
 from dvc.compat import fspath
 from dvc.exceptions import FileMissingError
 from dvc.main import main
@@ -49,6 +49,15 @@ def test_get_url_external(erepo_dir, remote_url):
     repo_url = "file://{}".format(erepo_dir)
     expected_url = URLInfo(remote_url) / "ac/bd18db4cc2f85cedef654fccc4a4d8"
     assert api.get_url("foo", repo=repo_url) == expected_url
+
+
+def test_get_url_git_only_repo_throws_exception(tmp_dir, scm):
+    tmp_dir.scm_gen({"foo": "foo"}, commit="initial")
+
+    with pytest.raises(UrlNotDvcRepoError) as exc_info:
+        api.get_url("foo", fspath(tmp_dir))
+
+    assert fspath(tmp_dir) in str(exc_info)
 
 
 @pytest.mark.parametrize("remote_url", all_remote_params, indirect=True)
--- a/tests/func/test_get.py
+++ b/tests/func/test_get.py
@@ -203,7 +203,7 @@ def test_get_url_not_existing(tmp_dir, erepo_dir, caplog):
             main(["get", fspath(erepo_dir), "not-existing-file", "--show-url"])
             == 1
         )
-        assert "failed to show url for 'not-existing-file'" in caplog.text
+        assert "failed to show url" in caplog.text
 
 
 def test_get_url_git_only_repo(tmp_dir, scm, caplog):
@@ -211,4 +211,4 @@ def test_get_url_git_only_repo(tmp_dir, scm, caplog):
 
     with caplog.at_level(logging.ERROR):
         assert main(["get", fspath(tmp_dir), "foo", "--show-url"]) == 1
-        assert "Only DVC repository is supported currently" in caplog.text
+        assert "failed to show url" in caplog.text
--- a/tests/func/test_get.py
+++ b/tests/func/test_get.py
@@ -187,13 +187,28 @@ def test_get_from_non_dvc_master(tmp_dir, erepo_dir, caplog):
     assert (tmp_dir / dst).read_text() == "some_contents"
 
 
-def test_get_url(tmp_dir, erepo_dir):
+def test_get_url_positive(tmp_dir, erepo_dir, caplog):
     with erepo_dir.chdir():
         erepo_dir.dvc_gen("foo", "foo")
-        assert main(["get", ".", "foo", "--show-url"]) == 0
+
+    caplog.clear()
+    with caplog.at_level(logging.ERROR, logger="dvc"):
+        assert main(["get", fspath(erepo_dir), "foo", "--show-url"]) == 0
+        assert caplog.text == ""
 
 
-def test_get_url_not_existing(tmp_dir, erepo_dir):
-    with erepo_dir.chdir():
-        erepo_dir.dvc_gen("foo", "foo")
-        assert main(["get", ".", "food", "--show-url"]) == 1
+def test_get_url_not_existing(tmp_dir, erepo_dir, caplog):
+    with caplog.at_level(logging.ERROR, logger="dvc"):
+        assert (
+            main(["get", fspath(erepo_dir), "not-existing-file", "--show-url"])
+            == 1
+        )
+        assert "failed to show url for 'not-existing-file'" in caplog.text
+
+
+def test_get_url_git_only_repo(tmp_dir, scm, caplog):
+    tmp_dir.scm_gen({"foo": "foo"}, commit="initial")
+
+    with caplog.at_level(logging.ERROR):
+        assert main(["get", fspath(tmp_dir), "foo", "--show-url"]) == 1
+        assert "Only DVC repository is supported currently" in caplog.text
--- a/tests/func/test_get.py
+++ b/tests/func/test_get.py
@@ -5,6 +5,7 @@ import pytest
 
 from dvc.cache import Cache
 from dvc.config import Config
+from dvc.main import main
 from dvc.repo.get import GetDVCFileError, PathMissingError
 from dvc.repo import Repo
 from dvc.system import System
@@ -184,3 +185,15 @@ def test_get_from_non_dvc_master(tmp_dir, erepo_dir, caplog):
 
     assert caplog.text == ""
     assert (tmp_dir / dst).read_text() == "some_contents"
+
+
+def test_get_url(tmp_dir, erepo_dir):
+    with erepo_dir.chdir():
+        erepo_dir.dvc_gen("foo", "foo")
+        assert main(["get", ".", "foo", "--show-url"]) == 0
+
+
+def test_get_url_not_existing(tmp_dir, erepo_dir):
+    with erepo_dir.chdir():
+        erepo_dir.dvc_gen("foo", "foo")
+        assert main(["get", ".", "food", "--show-url"]) == 1
--- a/tests/unit/command/test_get.py
+++ b/tests/unit/command/test_get.py
@@ -23,10 +23,9 @@ def test_get_url(mocker, caplog):
     assert cli_args.func == CmdGet
 
     cmd = cli_args.func(cli_args)
-    m = mocker.patch("dvc.api.get_url")
-    m.return_value = "url"
+    m = mocker.patch("dvc.api.get_url", return_value="resource_url")
 
     assert cmd.run() == 0
-    assert "url" in caplog.text
+    assert "resource_url" in caplog.text
 
     m.assert_called_once_with("src", repo="repo_url", rev="version")
--- a/tests/unit/command/test_get.py
+++ b/tests/unit/command/test_get.py
@@ -24,6 +24,7 @@ def test_get_url(mocker):
 
     cmd = cli_args.func(cli_args)
     m = mocker.patch("dvc.api.get_url")
+    m.return_value = "url"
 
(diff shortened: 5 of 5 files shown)


skshetry:
Some subtle difference that I noticed with `--show-url` flags vis-a-vis without flags:

1. `get` has path as a positional argument. I don't want to change it to have `nargs=*`. Is it okay to have `get <url> <single_path> --show-url`?

2. `dvc get . foo --rev=HEAD~` works, but, `dvc get . foo --rev=HEAD~ --show-url` does not work because of assertion in `_make_repo()`.

So, this is currently failing:
```python

def test_get_url_with_rev(tmp_dir, erepo_dir):
    with erepo_dir.chdir():
        erepo_dir.dvc_gen("foo", "foo")
        assert main(["get", ".", "foo", "--rev", "HEAD", "--show-url"]) == 0

```

dvc/command/get.py
```diff
@@ -55,4 +76,5 @@ def add_parser(subparsers, parent_parser):
     get_parser.add_argument(
         "--rev", nargs="?", help="Git revision (e.g. branch, tag, SHA)"
     )
+    get_parser.add_argument("--show-url", action="store_true")
```
efiop: Need to add `help` as well.

skshetry:
oops, I accidentally amended merge commit. Fixing.

dvc/command/get.py
```diff
@@ -10,7 +10,28 @@
...
+        from dvc.api import get_url
+
+        try:
+            url = get_url(
+                self.args.path, repo=self.args.url, rev=self.args.rev
+            )
+            logger.info("{} {}".format(self.args.path, url))
+        except DvcException:
+            logger.exception(
+                "failed to get url for '{}' from '{}'".format(
```
efiop: Maybe
```suggestion
                "failed to show url for '{}' from '{}'".format(
```
so it is more in-check with `--show-url`?

dvc/command/get.py
```diff
@@ -10,7 +10,28 @@
 
 
 class CmdGet(CmdBaseNoRepo):
+    def _run_get_url(self):
```
efiop: I guess good question here is whether or not `--show-url` should also become a part of `repo.get`. Hm... For now, we can leave it as is, since api has get_url anyways. 👍

skshetry: Yes, I wanted to refactor `get_url` and make that part of the internals, but, I left it for now.

efiop: Though, we could consider splitting it into `_get_url` and `_show_url` here, so that we could test them separately without invoking `main` each time. Kinda like we do in `pipeline show`.

dvc/command/get.py
```diff
@@ -10,7 +10,28 @@
...
+        from dvc.api import get_url
+
+        try:
+            url = get_url(
+                self.args.path, repo=self.args.url, rev=self.args.rev
+            )
```
efiop: Btw, another discrepancy is that `get_url` doesn't work with git files 🙁 Need to at least show a proper error message for now. We could consider showing a url to a raw file though, e.g. kinda like raw.github.com/... , but i'm not sure if that is natively supported by git somehow, especially if it is not github. 🤔

shcheklein:
@mroutis I think the short version `--url` will be even more confusing. `get --url` - feels like we provide an URL to get from or something. The whole thing is not ideal, but I still don't feel we should add a global command just for this. Also, like @jorgeorpinel we can move this to the future `dvc list` - it probably makes even more sense than `dvc get`.

mroutis:
@shcheklein , indeed! let's roll with `--show-url` and then move it to `list`

jorgeorpinel:
> Is `get --show-url | wget` the same as get?

Only for HTTP remotes.

mroutis:
> Only for HTTP remotes.

So, if I use it for S3 it won't give me a usable URL?

efiop:
@mroutis It will, but not usable with wget 🙂 I think

```
dvc get .. --show-url | dvc get-url
```

would work though 😄 Checkmate, wget!

skshetry:
> ```
> dvc get .. --show-url | dvc get-url
> ```
@efiop, do we support reading from stdin? Couldn't find any, and above command does not work, because we don't read from stdin (and, I need to change format of the output).

Workaround for now is following:
```sh
dvc get .. --show-url | xargs dvc get-url
```

> Is `get --show-url | wget` the same as `get`?
The following also does not seem to work unless we use `xargs` or `wget -i -`.

But, I have to change the output from `<path> <url>` to just `<url>`, as it works for only one path at the time.

efiop:
Good point @skshetry ! So `wget` behaves the same, let's keep the same behavior in `get-url` for now as well 🙂

skshetry:
> Good point @skshetry ! So `wget` behaves the same, let's keep the same behavior in `get-url` for now as well

You mean, introducing `-` arg? Or, leave as it is (like, plain `wget`)?

efiop:
@skshetry I meant just leaving it as is for now, indeed 🙂

tests/func/test_get.py
```diff
@@ -179,3 +180,15 @@ def test_get_from_non_dvc_master(tmp_dir, erepo_dir, caplog):
...
+
+def test_get_url(tmp_dir, erepo_dir):
+    with erepo_dir.chdir():
+        erepo_dir.dvc_gen("foo", "foo")
+        assert main(["get", ".", "foo", "--show-url"]) == 0
+
+
+def test_get_url_not_existing(tmp_dir, erepo_dir):
+    with erepo_dir.chdir():
+        erepo_dir.dvc_gen("foo", "foo")
```
pared: no need to generate anything if we want it to fail

tests/func/test_get.py
```diff
@@ -179,3 +180,15 @@ def test_get_from_non_dvc_master(tmp_dir, erepo_dir, caplog):
...
+        assert main(["get", ".", "food", "--show-url"]) == 1
```
pared: to pinpoint that its invalid command, you can go with `non-existing-file` instead of `food`

tests/func/test_get.py
```diff
@@ -179,3 +180,15 @@ def test_get_from_non_dvc_master(tmp_dir, erepo_dir, caplog):
 
     assert caplog.text == ""
     assert (tmp_dir / dst).read_text() == "some_contents"
+
+
+def test_get_url(tmp_dir, erepo_dir):
+    with erepo_dir.chdir():
+        erepo_dir.dvc_gen("foo", "foo")
+        assert main(["get", ".", "foo", "--show-url"]) == 0
```
pared: If we use `erepo` I would recommend making assert outside of `chdir` and replacing "." with fspath(erepo_dir). If we want to test in the same dir, lets just use `dvc` fixture :)

pared: and by making assert outside, I just mean removing the intdentation

tests/func/test_get.py
pared: Also, as mentioned above, lets make this assert inside tmp_dir. That way we don't have to use `erepo_dir.chdir()` at all.

tests/func/test_get.py
pared: Consider using caplog fixture and checking whether expected log message(s) is(are) in place

skshetry: thanks, that's better. will change.

dvc/command/get.py
skshetry: Handled by catching exception for `NotDvcRepoError`. For now, we won't support git-only repo files. We could theoretically generate a URL for GitHub for convenience, but, I'd leave it as an enhancement for now.

dvc/command/get.py
skshetry: I refactored it a bit to make it like `pipeline show`, but, I didn't see any tests where we directly make use of `Cmd*` commands (we do check for isinstance in few places though). `pipeline show` tests also calls `main`.

dvc/command/get.py
```diff
@@ -3,14 +3,49 @@
...
+            logger.info(url)
+        except NotDvcRepoError:
+            # NOTE: there's no way to get direct links to files inside
+            # git repository.
+            logger.exception(
+                "Could not show url for '{}' from '{}'. ".format(
+                    self.args.path, self.args.url
+                )
+                + "Only DVC repository is supported currently.",
```
efiop: let's include it into the string above, fees a bit odd to `+` after format 🙂

dvc/command/get.py
```diff
@@ -3,14 +3,49 @@
...
+        try:
+            url = get_url(
+                self.args.path, repo=self.args.url, rev=self.args.rev
+            )
+            logger.info(url)
+        except NotDvcRepoError:
+            # NOTE: there's no way to get direct links to files inside
+            # git repository.
+            logger.exception(
+                "Could not show url for '{}' from '{}'. ".format(
```
efiop: Feels weird that here we start with a capital letter and down below with a lowercase. Need to stick to one or another.

dvc/command/get.py
```diff
@@ -3,14 +3,49 @@
...
 class CmdGet(CmdBaseNoRepo):
+    def _show_url(self):
+        from dvc.api import get_url
+
+        try:
+            url = get_url(
+                self.args.path, repo=self.args.url, rev=self.args.rev
+            )
+            logger.info(url)
+        except NotDvcRepoError:
```
efiop: Hm, maybe it would make sense to move processing of this to `get_url` itself and raise an appropriate exception from there?

dvc/command/get.py
skshetry: Good catch. Let me see, what do we do usually. :slightly_smiling_face:

tests/func/test_get.py
```diff
@@ -179,3 +180,31 @@ def test_get_from_non_dvc_master(tmp_dir, erepo_dir, caplog):
...
+    with caplog.at_level(logging.ERROR, logger="dvc"):
+        assert (
+            main(["get", fspath(erepo_dir), "not-existing-file", "--show-url"])
+            == 1
+        )
+        assert "failed to show url for 'not-existing-file'" in caplog.text
+
+
+def test_get_url_git_only_repo(tmp_dir, scm, caplog):
+    with tmp_dir.chdir():
```
efiop: You are already in tmp_dir though 🙂 No need to chdir.

tests/func/test_get.py
```diff
@@ -179,3 +180,31 @@ def test_get_from_non_dvc_master(tmp_dir, erepo_dir, caplog):
 
     assert caplog.text == ""
     assert (tmp_dir / dst).read_text() == "some_contents"
+
+
+def test_get_url_positive(tmp_dir, erepo_dir, caplog):
+    with erepo_dir.chdir():
```
efiop: Same, no need to chdir.

tests/func/test_get.py
```diff
@@ -179,3 +180,31 @@ def test_get_from_non_dvc_master(tmp_dir, erepo_dir, caplog):
...
+        erepo_dir.dvc_gen("foo", "foo")
+
+    caplog.clear()
+    with caplog.at_level(logging.WARNING, logger="dvc"):
```
efiop: Hm, why are we testing warnings specifically?

skshetry: I wanted to check that there's no log above `WARNING`, but, I think it's more confusing this way, and I'm probably overdoing here. I'll change it to `ERROR`.

dvc/command/get.py
efiop: @skshetry If we raise proper exc from `get_url` as suggested in https://github.com/iterative/dvc/pull/3156#discussion_r367378039 , we will have only one exc processing snippet, so it will be consistent automatically 🙂

dvc/api.py
```diff
@@ -43,13 +43,27 @@ class SummonError(DvcException):
...
-        remote_obj = _repo.cloud.get_remote(remote)
-        return str(remote_obj.checksum_to_path_info(out.checksum))
+    try:
+        with _make_repo(repo, rev=rev) as _repo:
+            abspath = os.path.join(_repo.root_dir, path)
+            out, = _repo.find_outs_by_path(abspath)
+            remote_obj = _repo.cloud.get_remote(remote)
+            return str(remote_obj.checksum_to_path_info(out.checksum))
+    except NotDvcRepoError:
+        raise UrlNotDvcRepoError(repo)
```
skshetry: I'll raise this same error in `_make_repo` in a separate PR. Just keeping it here for now.

tests/func/test_api.py
```diff
@@ -51,6 +51,15 @@ def test_get_url_external(erepo_dir, remote_url):
     assert api.get_url("foo", repo=repo_url) == expected_url
 
 
+def test_get_url_git_only_repo_throws_exception(tmp_dir, scm):
+    tmp_dir.scm_gen({"foo": "foo"}, commit="initial")
+
+    with pytest.raises(UrlNotDvcRepoError) as exc_info:
+        api.get_url("foo", fspath(tmp_dir))
+
+    assert fspath(tmp_dir) in str(exc_info)
```
skshetry: :confused: Need to unescape somehow.
```sh
>       assert fspath(tmp_dir) in str(exc_info)
E       assert 'C:\\Users\\travis\\AppData\\Local\\Temp\\pytest-of-travis\\pytest-0\\popen-gw2\\test_get_url_git_only_repo_thr0' in '<ExceptionInfo UrlNotDvcRepoError("URL \'C:\\\\Users\\\\travis\\\\AppData\\\\Local\\\\Temp\\\\pytest-of-travis\\\\pytest-0\\\\popen-gw2\\\\test_get_url_git_only_repo_thr0\' is not a dvc repository.") tblen=2>'
```

skshetry: `str(exc_info).replace("\\", "\")`. :see_no_evil:

tests/func/test_get.py
```diff
@@ -184,3 +185,30 @@ def test_get_from_non_dvc_master(tmp_dir, erepo_dir, caplog):
...
     assert caplog.text == ""
     assert (tmp_dir / dst).read_text() == "some_contents"
+
+
+def test_get_url_positive(tmp_dir, erepo_dir, caplog):
+    with erepo_dir.chdir():
+        erepo_dir.dvc_gen("foo", "foo")
+
+    caplog.clear()
+    with caplog.at_level(logging.ERROR, logger="dvc"):
```
pared: I think that in this test, you should set caplog level to `INFO`, and verify that proper message is displayed 
(In your case it will be `relpath` to `erepo` + ".dvc/cache" + "checksum_part"))
where checksum part is checksum[:2]/checksum[2:] (thats how we store files in cache).

This is important because it is testing the core of what this change is all about - informing the user what is the resource path. I would also consider displaying an absolute path, but we can leave this discussion for some other time.

skshetry: Honestly, I did not want to delve into how `dvc.api.get_url()` works previously, and verify it's correctness. I just wanted it to be delegated to `get_url()` and be its concern. I wanted this to be checked in unit tests, but, I quite missed to check the `caplog` there.

Would you still like to have a functional test for this? Or, are you okay with changes to the unit test? What do you think?

pared: @skshetry you are right about delegating work to `get_url`. I would be ok with patching `get_api` in unit test and verifying that proper value has been logged. As I see its already done, so lets just add caplog there and verify that proper value is displayed.

skshetry: Fixed by 02a73a0

tests/unit/command/test_get.py
```diff
@@ -14,3 +14,19 @@ def test_get(mocker):
...
+
+
+def test_get_url(mocker, caplog):
+    cli_args = parse_args(
+        ["get", "repo_url", "src", "--rev", "version", "--show-url"]
+    )
+    assert cli_args.func == CmdGet
+
+    cmd = cli_args.func(cli_args)
+    m = mocker.patch("dvc.api.get_url")
```
pared:
```suggestion
    m = mocker.patch("dvc.api.get_url", return_value="resource_url")
```

tests/unit/command/test_get.py
```diff
@@ -14,3 +14,19 @@ def test_get(mocker):
...
+    m.return_value = "url"
```
pared:
```suggestion
```

tests/unit/command/test_get.py
```diff
@@ -14,3 +14,19 @@ def test_get(mocker):
...
+
+    assert cmd.run() == 0
+    assert "url" in caplog.text
```
pared:
```suggestion
     assert "resource_url" in caplog.text
```

dvc/command/get.py
```diff
@@ -55,4 +75,9 @@ def add_parser(subparsers, parent_parser):
     get_parser.add_argument(
         "--rev", nargs="?", help="Git revision (e.g. branch, tag, SHA)"
     )
+    get_parser.add_argument(
+        "--show-url",
+        action="store_true",
+        help="Returns path/url to the location in remote for specified path",
```
efiop: Could use the description from https://github.com/iterative/dvc/pull/3130/files .

efiop:
@skshetry Also, need to not forget about docs 🙂

dvc/command/get.py
efiop: Though, looks like it is still up for discussion, so i'm sure we will modify it while working on the docs. Let's keep it as is for now 👍

jorgeorpinel: Ah yeah I would change this description, its confusing. Sorry, didn't notice before merge. Maybe
"Returns URL to the storage location of the target data".

dvc/api.py
```diff
@@ -43,13 +43,27 @@ class SummonError(DvcException):
...
 
+class UrlNotDvcRepoError(DvcException):
+    """Thrown if given url is not a DVC repository.
+
+    Args:
+        url (str): url to the repository.
+    """
+
+    def __init__(self, url):
+        super().__init__("URL '{}' is not a dvc repository.".format(url))
```
jorgeorpinel: Capitalize DVC.

dvc/api.py
```diff
@@ -43,13 +43,27 @@ class SummonError(DvcException):
     pass
 
 
+class UrlNotDvcRepoError(DvcException):
+    """Thrown if given url is not a DVC repository.
+
+    Args:
+        url (str): url to the repository.
```
jorgeorpinel: I don't see a `url` arg.
p.s. I'd capitalize "URL" in the explanation side.

dvc/command/get.py
```diff
@@ -10,7 +10,27 @@
...
+    def _show_url(self):
+        from dvc.api import get_url
+
+        try:
+            url = get_url(
+                self.args.path, repo=self.args.url, rev=self.args.rev
+            )
+            logger.info(url)
+        except DvcException:
+            logger.exception("failed to show url")
```
jorgeorpinel: Capitalization: "Failed to show URL"
Maybe end in period `.` ? (This depends on whether it's printed as part of a larger message – may also apply to capitalizing first letter `f`.

dvc/command/get.py
jorgeorpinel: Maybe "Returns URL to the storage location of the target data."
See https://github.com/iterative/dvc.org/pull/936#discussion_r368746593

tests/func/test_api.py
```diff
@@ -51,6 +51,16 @@ def test_get_url_external(erepo_dir, remote_url):
     assert api.get_url("foo", repo=repo_url) == expected_url
 
 
+def test_get_url_git_only_repo_throws_exception(tmp_dir, scm):
+    tmp_dir.scm_gen({"foo": "foo"}, commit="initial")
+
+    with pytest.raises(UrlNotDvcRepoError) as exc_info:
+        api.get_url("foo", fspath(tmp_dir))
+
+    # On windows, `exc_info` has path escaped, eg: `C:\\\\Users\\\\travis`
```
jorgeorpinel: "Windows", "e.g." :P – grammar not too important in comments, of course. Just saying

dvc/command/get.py
skshetry: @jorgeorpinel, the lowercase seems to be used everywhere in the codebase. And, because this is a `logger.exception`, our formatter handles adding extra information from raised exception besides this particular logging message. Notice the messages on following examples:
```sh
➜ dvc get https://github.com/schacon/cowsay install.sh --show-url 
ERROR: failed to show url - URL 'https://github.com/schacon/cowsay' is not a dvc repository.
➜ dvc get . dir/file --show-url
ERROR: failed to show url - unable to find DVC-file with output 'dir/file'
```

dvc/api.py
skshetry: It's in `__init__(self, url)`, and is following Google-style docstrings. In Google-style, `__init__()` method can be both documented in class-level or inside `__init__` methods itself. I'll make a quick PR regarding the *URL* though.

dvc/command/get.py
shcheklein: `failed to show url - URL` it looks strange though :) First it proves that `the lowercase seems to be used everywhere` is not true (and it should not be, typically, acronyms and initialisms are written in all capital letters to distinguish them from ordinary words. - NASA, URL, DVC, etc). Second, you have a repetition here. Repetition is probably fine if reasons vary and you can't actually control them.

skshetry: @shcheklein, it's not something that I add, but, I'm utilizing the fact that something will be there. So, it's more like "(what failed) - (the reason)".

skshetry: And, yes, regarding lowercase, I just looked into it and it's not everywhere for other cases, but for `exceptions`, lowercase is used *everywhere*.

dvc/api.py
jorgeorpinel: > It's in __init__(self, url)

🤦‍♂ sry didn't notice it's a class haha

> following Google-style docstrings

Great! Did you read my proposal to use that in https://github.com/iterative/dvc/issues/3176#issuecomment-576539623? Or is this a happy coincidence?

Don't worry about `url`, I'll just address all my own feedback here in a quick PR. Working on it now.

dvc/command/get.py
jorgeorpinel: > because this is a logger.exception, our formatter handles adding extra information

Thanks, very useful to know this.

> lowercase seems to be used everywhere is not true...

➕ 1️⃣  No need to use bad grammar on purpose because of existing bad grammar.

> >failed to show url - URL it looks strange though :) ...
>
> it's more like "(what failed) - (the reason)".

So then `the reason` part in this case should start directly with `http://...` – but that's not part of this PR...

jorgeorpinel:
UPDATE: I'm addressing my own feedback in #3220.

dvc/command/get.py
jorgeorpinel: > the reason part in this case should start directly with http://... – but that's not part of this PR 

OK I changed this in https://github.com/iterative/dvc/pull/3220/commits/ec0c39a28ec8a26b76843a715449d54d810fcbb3#diff-3b2745249f7139953ab4b45b382e69e9 but this whole conversation kind of opened a can of worms and that PR grew quite a bit 😅  Hoping core team doesn't flip out.

> but for exceptions, lowercase is used everywhere

Thanks for the tip. I may review some of these too 😋

jorgeorpinel: p.s. that PR = iterative/dvc/pull/3220

PR #3165, opened 2020-01-15
setup: remove py2 backports

mroutis:


shcheklein:
just a reminder - don't forget to update conda, apt (?), etc ...

mroutis:
@shcheklein , as far as I know, both `conda` and `debian` uses the source code to build from python, no need to tweak anything specific.

efiop:
@mroutis Both have separate list of requirements, so we do need to update them(e.g. https://github.com/conda-forge/dvc-feedstock/blob/master/recipe/meta.yaml#L29). Only configparser matters though, as mock is just a test requirement. Debian WIP package is having hard times now, as we make a lot of deps changes, so let's not bother with it for now, we will need to revisit it when we actually release it anyways. No need to send a patch for conda too, I will have to update it soon too.

efiop:
@mroutis Tests are failing for some reason though...

mroutis:
@efiop , I'll leave `mock` for now, since I can't tell why it is not working as intended.

mroutis:
would you mind re-reviewing the PR, @efiop ?

PR #3166, opened 2020-01-16
api: identify local repos with os.path.exists

### LINKED ISSUE #3111: dvc.api: doesn't work with repo with SSH URL
```console
$ dvc version
DVC version: 0.80.0
Python version: 3.7.6
Binary: False
Package: brew
```
This script
```py
import csv
import dvc.api

with dvc.api.open(
  "sea_ice.csv",
  repo="git@github.com:iterative/df_sea_ice_no_header.git"
) as fd:
  reader = csv.reader(fd)
  for row in reader:
    print(row[0])
```
Throws
```
  File "tests/open-test.py", line 6, in <module>
    repo="git@github.com:iterative/df_sea_ice_no_header.git"
  File "/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "/Users/drbidibombom/DVC-repos/api-tests/.env/lib/python3.7/site-packages/dvc/api.py", line 78, in _open
  File "/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "/Users/drbidibombom/DVC-repos/api-tests/.env/lib/python3.7/site-packages/dvc/api.py", line 98, in _make_repo
    yield Repo(repo_url)
  File "/Users/drbidibombom/DVC-repos/api-tests/.env/lib/python3.7/site-packages/dvc/repo/__init__.py", line 77, in __init__
    root_dir = self.find_root(root_dir)
  File "/Users/drbidibombom/DVC-repos/api-tests/.env/lib/python3.7/site-packages/dvc/repo/__init__.py", line 145, in find_root
    raise NotDvcRepoError(root)
dvc.exceptions.NotDvcRepoError: you are not inside of a dvc repository (checked up to mount point '/')
```

It looks like it's trying to use my cwd `.` as the repo (and I'm not working from a DVC repo).

mroutis:
Close #3111

### DIFF
--- a/dvc/api.py
+++ b/dvc/api.py
@@ -2,7 +2,6 @@ from builtins import open as builtin_open
 import importlib
 import os
 import sys
-from urllib.parse import urlparse
 from contextlib import contextmanager, _GeneratorContextManager as GCM
 import threading
 
@@ -96,7 +95,7 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
 
 @contextmanager
 def _make_repo(repo_url, rev=None):
-    if not repo_url or urlparse(repo_url).scheme == "":
+    if not repo_url or os.path.exists(repo_url):
         assert rev is None, "Custom revision is not supported for local repo"
         yield Repo(repo_url)
     else:


dvc/api.py
```diff
@@ -96,7 +95,7 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
 
 @contextmanager
 def _make_repo(repo_url, rev=None):
-    if not repo_url or urlparse(repo_url).scheme == "":
+    if not repo_url or os.path.exists(repo_url):
```
mroutis: I'm using a solution suggested by @efiop , which identifies local repositories by checking if they are in the file system. There's no way a URL (e.g. `https://`, `git@github`) could be missed like that.

I'm not aware of any other edge case.

mroutis: Probably, `file://` but no one uses it :sweat_smile:

efiop: Yeah, we don't support `file://` anywhere else, so no need to worry about it for now. Neither any user asked for it ever 🙂

skshetry:
It'd be good to have a regression test so that ssh URLs don't get broken later on. :slightly_smiling_face:

mroutis:
sure, @skshetry , didn't know how to test it properly, to be honest :sweat_smile: 

Any ideas? Do you want to send the pull request? :sweat_smile:

efiop:
@skshetry The `exists` works pretty good there, as we don't mess with urls at all. Whatever we have - we just pass it to git to deal with it. You mean something like a mock test that would ensure that ssh-like git url would be passed to _make_repo instead of only to Repo()?

PR #3169, opened 2020-01-16
pull: treat HTTP redirects without Location header as error

### LINKED ISSUE #2963: `dvc pull` with wrong S3 remote failed but user wasn't informed
System
----------

My `dvc` version is 0.77.3, I installed & upgraded `dvc` through `pip3` and I'm using Fedora 31.

Problem
-----------

I followed the 'Getting Started' guide and ran the following commands

```
$ dvc init
$ git add .dvc/* && git commit -m 'Initialized dvc'
$ dvc remote add -d s3remote https://s3.amazonaws.com/<bucket-name>
$ dvc add <some-file>
$ git add .gitignore <some-file>.dvc && git commit -m 'Added <some-file'
$ dvc push -r s3remote
```
The last command finished with `Everything is up to date.` which is weird since the S3 bucket was empty when checking it. I went on to delete `<some-file>` and run `dvc pull -r s3remote` which would restore `<some-file>` without throwing any errors. This was surprising since it did not get clear where `dvc` is storing my file backups and why my S3 bucket was still empty without `dvc` communicating any errors.

Only when I removed the `.dvc/cache` folder and ran `dvc pull -r s3remote` it would complain with the following:

```
ERROR: failed to download 'https://s3.amazonaws.com/<bucket-name>/31/69f7ce4ebb503afca037d35b7eb3a9' to '.dvc/cache/31/69f7ce4ebb503afca037d35b7eb3a9' - '301 Moved Permanently'

ERROR: failed to download 'https://s3.amazonaws.com/<bucket-name>/a3/04afb96060aad90176268345e10355' to '.dvc/cache/a3/04afb96060aad90176268345e10355' - '301 Moved Permanently'

ERROR: failed to pull data from the cloud - 2 files failed to download


```

The fix
---------

On my end, it was an easy fix. After consulting the documentation in more detail, I realized that I set up my S3 remote incorrectly using the `https://s3.amazonaws.com/<bucket-name>` URL to the bucket whereas I should have used the `s3://<bucket-name>` URL. Hence, running

```
$ dvc remote modify s3remote url s3://<bucket-name>
```

did the job.

It still remains weird why I was never warned about the remote being incorrectly set up and why `dvc pull -r s3remote` worked without any problems even though there were no files in the S3 bucket (it seems that it restored them from local cache as a fall-back) but it should have given me a headsup!

fabiosantoscode:
The reason for this problem is a 301 without a Location header. In our HTTP remote we can treat this as an error, since we don't know where to actually find the data.

Fixes #2963

-------

### DIFF
--- a/dvc/remote/http.py
+++ b/dvc/remote/http.py
@@ -103,7 +103,13 @@ class RemoteHTTP(RemoteBASE):
         try:
             res = self._session.request(method, url, **kwargs)
 
-            if kwargs["allow_redirects"] and res.status_code in (301, 302):
+            redirect_no_location = (
+                kwargs["allow_redirects"]
+                and res.status_code in (301, 302)
+                and "location" not in res.headers
+            )
+
+            if redirect_no_location:
                 # AWS s3 doesn't like to add a location header to its redirects
                 # from https://s3.amazonaws.com/<bucket name>/* type URLs.
                 # This should be treated as an error
--- a/dvc/remote/http.py
+++ b/dvc/remote/http.py
@@ -103,7 +103,7 @@ class RemoteHTTP(RemoteBASE):
         try:
             res = self._session.request(method, url, **kwargs)
 
-            if kwargs["allow_redirects"] and res.is_redirect:
+            if kwargs["allow_redirects"] and res.status_code in (301, 302):
                 # AWS s3 doesn't like to add a location header to its redirects
                 # from https://s3.amazonaws.com/<bucket name>/* type URLs.
                 # This should be treated as an error
--- a/dvc/remote/http.py
+++ b/dvc/remote/http.py
@@ -109,6 +109,8 @@ class RemoteHTTP(RemoteBASE):
                 # This should be treated as an error
                 raise requests.exceptions.RequestException
 
+            return res
+
         except requests.exceptions.RequestException:
             raise DvcException("could not perform a {} request".format(method))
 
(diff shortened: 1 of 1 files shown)


dvc/remote/http.py
```diff
@@ -101,7 +101,16 @@ def _request(self, method, url, **kwargs):
...
 
         try:
-            return self._session.request(method, url, **kwargs)
+            res = self._session.request(method, url, **kwargs)
+
+            if kwargs["allow_redirects"] and res.status_code in (301, 302):
+                # AWS s3 doesn't like to add a location header to its redirects
+                # from https://s3.amazonaws.com/<bucket name>/* type URLs.
+                # This should be treated as an error
+                raise requests.exceptions.RequestException
```
Suor: What if this is not s3 and location header is present?

fabiosantoscode: Then `requests` would've followed the redirect.

fabiosantoscode: Maybe better to be explicit and check that a location header is not present?

Suor: Yes, that would be better. At least the comment will follow the code, now it looks like a bug in implementation. So adjusting comment would be good too.

fabiosantoscode: Makes sense

efiop:
@fabiosantoscode For the record, your fix doesn't really address #2963 , because that issue is about `push`, which won't work anyways because we don't have support for `upload`ing through HTTP remote. Plus in that issue another concern is that `dvc push` when doesn't need to push anything, doesn't even try to access the remote as a sanity check.

Your PR is still useful, but I don't have any way of reproducing the issue that it is solving. But looks alright - so merging. Thanks! 🙂

PR #3170, opened 2020-01-16
utils: resolve_output: normpath before extracting basename

### LINKED ISSUE #3105: dvc get can can delete the current directory
First off, I saw it coming, had my stuff backed up etc, so no harm done. 
I was trying to test out dvc get to pull models from a different project and the cache must not have been present in the remote:

`models/mlr/production` is defined as the output of a dvc file. 

```
$ dvc get <internal_repo> models/mlr/production/ --rev <name of branch>
name: models/mlr/production, md5: ebbb5c351028a6e5485dfdec639fbb57.dir
Missing cache for directory '../../../tmp/tmp5jtl3fccdvc-erepo/models/mlr/production'. Cache for files inside will be lost. Would you like to continue? Use '-f' to force. [y/n] y            
name: models/mlr/production, md5: ebbb5c351028a6e5485dfdec639fbb57.dir
WARNING: Cache 'ebbb5c351028a6e5485dfdec639fbb57.dir' not found. File '.' won't be created.                                                                                                   
file '.' is going to be removed. Are you sure you want to proceed? [y/n] y                                                                                                                    
ERROR: unexpected error - [Errno 2] No such file or directory
```

Once I saw the `File '.' won't be created.  file '.' is going to be removed. ` I knew what was going to happen, but was surprised that dvc lets you do that. (It wiped out my whole git repo ;)) 

DVC version 0.80.0,
Platform WSL
method of installation: pip

pared:

Fixes #3105

### DIFF
--- a/dvc/utils/__init__.py
+++ b/dvc/utils/__init__.py
@@ -336,7 +336,7 @@ def env2bool(var, undefined=False):
 def resolve_output(inp, out):
     from urllib.parse import urlparse
 
-    name = os.path.basename(urlparse(inp).path)
+    name = os.path.basename(os.path.normpath(urlparse(inp).path))
     if not out:
         return name
     if os.path.isdir(out):
--- a/tests/unit/utils/test_utils.py
+++ b/tests/unit/utils/test_utils.py
@@ -4,7 +4,7 @@ import os
 import pytest
 
 from dvc.path_info import PathInfo
-from dvc.utils import file_md5
+from dvc.utils import file_md5, resolve_output
 from dvc.utils import fix_env
 from dvc.utils import relpath
 from dvc.utils import to_chunks
@@ -104,3 +104,28 @@ def test_relpath():
     path_info = PathInfo(path)
 
     assert relpath(path) == relpath(path_info)
+
+
+@pytest.mark.parametrize(
+    "inp,out,is_dir,expected",
+    [
+        ["target", None, False, "target"],
+        ["target", "dir", True, os.path.join("dir", "target")],
+        ["target", "file_target", False, "file_target"],
+        [
+            "target",
+            os.path.join("dir", "subdir"),
+            True,
+            os.path.join("dir", "subdir", "target"),
+        ],
+        ["dir/", None, False, "dir"],
+        ["dir", None, False, "dir"],
+        ["dir", "other_dir", False, "other_dir"],
+        ["dir", "other_dir", True, os.path.join("other_dir", "dir")],
+    ],
+)
+def test_resolve_output(inp, out, is_dir, expected, mocker):
+    with mocker.patch("os.path.isdir", return_value=is_dir):
+        result = resolve_output(inp, out)
+
+    assert result == expected


tests/unit/utils/test_utils.py
```diff
@@ -1,10 +1,11 @@
 import re
 import os
 
+import mock
```
skshetry:
```suggestion
from unittest import mock
```

skshetry: Are we moving from `mock` to `unittest.mock`? cc @efiop 

But, it'd be good to not introduce here?

efiop: @pared Or even better, just use `mocker` fixture, that would be the most correct way to go about it.

efiop: @skshetry Even better, we are moving to pytest, so should simply use mocker fixture. Great catch!

tests/unit/utils/test_utils.py
```diff
@@ -104,3 +105,28 @@ def test_relpath():
...
+            True,
+            os.path.join("dir", "subdir", "target"),
+        ],
+        ["dir/", None, False, "dir"],
+        ["dir", None, False, "dir"],
+        ["dir", "other_dir", False, "other_dir"],
+        ["dir", "other_dir", True, os.path.join("other_dir", "dir")],
+    ],
+)
+def test_resolve_output(inp, out, is_dir, expected):
```
skshetry: Personally, it's too hard for me to parse this matrix. I'd like to have it flattened-out somehow, eg:
```python
with mock.patch("os.path.isdir", return_value=True):
	assert resolve_output("target", "dir") == os.path.join("target", "dir")
    # ... more tests

with mock.patch("os.path.isdir", return_value=False):
	assert resolve_output("target", None) == "target"
    # ... more tests
```

You don't have to change anything, I'm just stating my preference. I'm fine with this as well.

dvc/utils/__init__.py
```diff
@@ -336,7 +336,7 @@ def env2bool(var, undefined=False):
 def resolve_output(inp, out):
     from urllib.parse import urlparse
 
-    name = os.path.basename(urlparse(inp).path)
+    name = os.path.basename(os.path.normpath(urlparse(inp).path))
```
mroutis: :thinking: hard to know that it was originally wrong (before `normpath`). Wondering if there's a better way to handle this cases without thinking too much about _correctness_.

mroutis: good catch, @pared :+1:

tests/unit/utils/test_utils.py
pared: I agree that it might be easier to read through "flat" version, but this matrix (in my opinion) makes it easier to compare particular cases. Also, If we have "flat" version, during test execution, in case of singular fail. we will only know that this test failed, while in case of parameterization we will know exactly which argument set has failed.

pared: I would also say that many `with` statements one under another kind of obscures what the test is about.

pared:
> why you didn't add the test that you wrote on the issue?

@mroutis when I noted that the problem is with resolving the output, it came to my mind that this particular bug is not about `get`, but only about resolving the output. Removing functionality worked properly, so I thought there is no need to make a functional test.

Issue #3124, opened 2020-01-13
import/get/api: external repo and cached_clone are problematic


Suor:
There are several issues with `external_repo()`, `cached_clone()` and code using them - get, import and api stuff:

- there are 3 parallel code paths doing very close things, we add the same features for them, fix the same bugs, essentially do triple work
- `external_repo()` and `cached_clone()` provide two entry points to conceptually the same thing, calling code need to make dup calls, i.e. `cached_clone()` and then `external_repo()` if that happens to be a dvc repo. 
- `_external_repo()` and `cached_clone()` share cache, which leads to `_external_repo()` not making proper initialization under some circumstances
- tests are insufficent: 
    - we don't properly test urls, these are either dirs or rarely `file://`,
    - we don't test caching behavior
- `cached_clone()` should really be a context manager or be integrated back into `external_repo()`
- some bugs are tricky to fix under all these circumstances, i.e. #3062

So we need a common abstraction extending `external_repo()` to work with non-dvc repos/branches, which might be reused across get, import and api. We then need to refactor the latter to use that new abstraction.

PR #3172, opened 2020-01-16
update: handle imported files from non-DVC git repositories

### LINKED ISSUE #2976: update: Handle Git-tracked import-ed files
After https://github.com/iterative/dvc/pull/2889, `dvc import` can also import files that are tracked by Git but not DVC. When the user tries to `dvc update` them, the operation fails. See also https://github.com/iterative/dvc/issues/2959

fabiosantoscode:
This makes `dvc update` update files from git repositories. git-tracked imported files were actually already handled.

Fixes #2976

-----

### DIFF
--- a/tests/dir_helpers.py
+++ b/tests/dir_helpers.py
@@ -61,7 +61,7 @@ __all__ = [
     "repo_template",
     "run_copy",
     "erepo_dir",
-    "git_dir",
+    "external_git_dir",
 ]
 REPO_TEMPLATE = {
     "foo": "foo",
@@ -296,10 +296,10 @@ def erepo_dir(tmp_path_factory, monkeypatch):
 
 
 @pytest.fixture
-def git_dir(tmp_path_factory, monkeypatch):
+def external_git_dir(tmp_path_factory, monkeypatch):
     from dvc.scm.git import Git
 
-    path = TmpDir(fspath_py35(tmp_path_factory.mktemp("git_dir")))
+    path = TmpDir(fspath_py35(tmp_path_factory.mktemp("external_git_dir")))
 
     # Create git repo
     with path.chdir():
--- a/tests/dir_helpers.py
+++ b/tests/dir_helpers.py
@@ -307,6 +307,7 @@ def git_dir(tmp_path_factory, monkeypatch):
 
     try:
         path.scm = Git(fspath(path))
+        path.scm.commit("initial commit to create the master branch")
         yield path
     finally:
         path.scm.close()
--- a/tests/dir_helpers.py
+++ b/tests/dir_helpers.py
@@ -54,7 +54,15 @@ from dvc.utils.fs import makedirs
 from dvc.compat import fspath, fspath_py35
 
 
-__all__ = ["tmp_dir", "scm", "dvc", "repo_template", "run_copy", "erepo_dir"]
+__all__ = [
+    "tmp_dir",
+    "scm",
+    "dvc",
+    "repo_template",
+    "run_copy",
+    "erepo_dir",
+    "git_dir",
+]
 REPO_TEMPLATE = {
     "foo": "foo",
     "bar": "bar",
@@ -285,3 +293,20 @@ def erepo_dir(tmp_path_factory, monkeypatch):
         path.dvc.close()
 
     return path
+
+
+@pytest.fixture
+def git_dir(tmp_path_factory, monkeypatch):
+    from dvc.scm.git import Git
+
+    path = TmpDir(fspath_py35(tmp_path_factory.mktemp("git_dir")))
+
+    # Create git repo
+    with path.chdir():
+        _git_init()
+
+    try:
+        path.scm = Git(fspath(path))
+        yield path
+    finally:
+        path.scm.close()
--- a/tests/func/test_status.py
+++ b/tests/func/test_status.py
@@ -61,23 +61,25 @@ def test_status_non_dvc_repo_import(tmp_dir, dvc, erepo_dir):
     }
 
 
-def test_status_before_and_after_dvc_init(tmp_dir, dvc, git_dir):
-    with git_dir.chdir():
-        git_dir.scm_gen("file", "first version", commit="first version")
-        old_rev = git_dir.scm.get_rev()
+def test_status_before_and_after_dvc_init(tmp_dir, dvc, external_git_dir):
+    with external_git_dir.chdir():
+        external_git_dir.scm_gen(
+            "file", "first version", commit="first version"
+        )
+        old_rev = external_git_dir.scm.get_rev()
 
-    dvc.imp(fspath(git_dir), "file", "file")
+    dvc.imp(fspath(external_git_dir), "file", "file")
 
     assert dvc.status(["file.dvc"]) == {}
 
-    with git_dir.chdir():
-        git_dir.dvc = Repo.init()
-        git_dir.scm.repo.index.remove(["file"])
+    with external_git_dir.chdir():
+        external_git_dir.dvc = Repo.init()
+        external_git_dir.scm.repo.index.remove(["file"])
         os.remove("file")
-        git_dir.dvc_gen("file", "second version")
-        git_dir.scm.add([".dvc", "file.dvc"])
-        git_dir.scm.commit("version with dvc")
-        new_rev = git_dir.scm.get_rev()
+        external_git_dir.dvc_gen("file", "second version")
+        external_git_dir.scm.add([".dvc", "file.dvc"])
+        external_git_dir.scm.commit("version with dvc")
+        new_rev = external_git_dir.scm.get_rev()
 
     assert old_rev != new_rev
 
@@ -89,6 +91,6 @@ def test_status_before_and_after_dvc_init(tmp_dir, dvc, git_dir):
 
     assert status == {
         "changed deps": {
-            "file ({})".format(fspath(git_dir)): "update available"
+            "file ({})".format(fspath(external_git_dir)): "update available"
         }
     }
--- a/tests/func/test_status.py
+++ b/tests/func/test_status.py
@@ -61,27 +61,23 @@ def test_status_non_dvc_repo_import(tmp_dir, dvc, erepo_dir):
     }
 
 
-def test_status_before_and_after_dvc_init(tmp_dir, dvc, erepo_dir):
-    with erepo_dir.chdir():
-        erepo_dir.scm.repo.index.remove([".dvc"], r=True)
-        shutil.rmtree(".dvc")
-        erepo_dir.scm_gen("file", "first version")
-        erepo_dir.scm.add(["file"])
-        erepo_dir.scm.commit("first version")
-        old_rev = erepo_dir.scm.get_rev()
+def test_status_before_and_after_dvc_init(tmp_dir, dvc, git_dir):
+    with git_dir.chdir():
+        git_dir.scm_gen("file", "first version", commit="first version")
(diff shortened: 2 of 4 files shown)


tests/func/test_update.py
```diff
@@ -172,3 +172,35 @@ def test_update_import_url(tmp_dir, dvc, tmp_path_factory):
 
     assert dst.is_file()
     assert dst.read_text() == "updated file content"
+
+
+def test_update_git_tracked(tmp_dir, dvc, erepo_dir):
+    with erepo_dir.chdir():
+        erepo_dir.scm.repo.index.remove([".dvc"], r=True)
```
mroutis: what's the need of removing `.dvc` from the index itself? :thinking:

tests/func/test_update.py
```diff
@@ -172,3 +172,35 @@ def test_update_import_url(tmp_dir, dvc, tmp_path_factory):
...
+        shutil.rmtree(".dvc")
+        erepo_dir.scm_gen("file", "first version")
```
mroutis: you could do:
```suggestion
        erepo_dir.scm_gen("file", "first version", commit="first version")
```

tests/func/test_update.py
```diff
@@ -172,3 +172,35 @@ def test_update_import_url(tmp_dir, dvc, tmp_path_factory):
...
+        erepo_dir.scm.add(["file"])
+        erepo_dir.scm.commit("first version")
+
+    stage = dvc.imp(fspath(erepo_dir), "file", "file")
+
+    # Just to make sure it doesn't crash
+    dvc.update(stage.path)
```
mroutis: why would this crash?

skshetry: probably defensive testing, to check that there's no error even if the file is unchanged?

tests/func/test_update.py
```diff
@@ -172,3 +172,35 @@ def test_update_import_url(tmp_dir, dvc, tmp_path_factory):
...
+
+    assert (tmp_dir / "file").read_text() == "first version"
+
+    with erepo_dir.chdir():
+        erepo_dir.scm.repo.index.remove(["file"])
+        os.remove("file")
```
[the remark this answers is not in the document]
mroutis: Thanks for explaining this, @skshetry)

skshetry:
The following is failing for me for some reasons:
```sh
#! /usr/bin/env sh

set -ex

repo=$(mktemp -d)
upstream_repo=$(mktemp -d)

cd $repo
git init && dvc init

cd $upstream_repo
git init && dvc init
echo "foo" >> foo
git add foo
git commit -m "first version"

cd $repo
dvc import $upstream_repo foo

cd $upstream_repo
echo "foo" >> foo
git commit -am "second version"

cd $repo
dvc update foo.dvc
dvc update foo.dvc
```
with following error:
```
ERROR: failed to update 'foo.dvc'. - config file error: Section 'remote "auto-generated-upstream"' already exists. Use `-f|--force` to overwrite section with new value.
```

If i do not `dvc init` in the upstream repo, then, it works perfectly fine.
cc @mroutis

mroutis:
> https://github.com/iterative/dvc/pull/3172#issuecomment-575461388

@skshetry , interesting! I guess that's another error, since it happens on DVC initialized repositories.
I guess the error comes while trying to _checkout_ the file.

To be honest, not sure what's going on here, @skshetry, I would need to take a look :)

efiop:
@skshetry That is a nice catch! Might be some issues with our caching mechanism, where it tries to auto-create a remote twice here https://github.com/iterative/dvc/blob/master/dvc/external_repo.py#L91 .  Haven't looked deeply into it yet though.

skshetry:
@efiop, how about this patch? Should `_external_repo()` complain about this?

```
diff --git a/dvc/external_repo.py b/dvc/external_repo.py
index 9ff2f2a4..dd4c30ce 100644
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -92,6 +92,7 @@ def _external_repo(url=None, rev=None, cache_dir=None):
                         "auto-generated-upstream",
                         original_repo.cache.local.cache_dir,
                         default=True,
+                        force=True,
                         level=Config.LEVEL_LOCAL,
                     )
                 finally:
```

efiop:
@skshetry Yep, that would work 🙂 But that might be us covering up a bug, will need to look why that happened in the first place.

tests/func/test_update.py
fabiosantoscode: The purpose of this issue is to check that we can `dvc import` and then `dvc update` from a git repository. So I'm purging DVC and committing that. My reasoning is that when the repo is cloned so that it can be imported or updated, its `.dvc` directory being in the index makes a difference.

tests/func/test_update.py
fabiosantoscode: That's right. The purpose here is to increase coverage a bit.

tests/func/test_update.py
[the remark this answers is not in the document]
fabiosantoscode: Oh, I didn't know. I'll try that out

fabiosantoscode:
@efiop want me to have a look in scope of this PR?

fabiosantoscode:
@skshetry nice catch!

I had a quick look and if I change `dvc.external_repo._default_remote_set` to always return True this doesn't happen. Probably means that this function is not finding the default remote, and then it fails to add a default remote.

Maybe we can try to add the remote and catch the resulting error instead of doing an `if` and checking for the default remote existing? It's probably more pythonic and less prone to race conditions.

tests/func/test_update.py
fabiosantoscode: I can do that but would still need a commit to remove `.dvc` in a way that can be cloned as a regular git repo

fabiosantoscode:
@mroutis I simplified the test with most of your example, but still need to remove the .dvc directory since this test is for pure-git repositories. DVC repositories are already tested in the same file.

Maybe we could have a new fixture `erepo_raw_git_dir`?

skshetry:
@fabiosantoscode, I guess, the fix is out of the scope of this PR (and, not introduced by the changes).

tests/func/test_update.py
skshetry: Can you also simplify this?
```suggestion
        erepo_dir.scm_gen("file", "first version", commit="first version")
```

fabiosantoscode:
@skshetry makes sense. I'll stop investigating for now.

For the record, if you add this line to the top of DependencyRepo.status() it also fixes the problem:

`from dvc.external_repo import clean_repos; clean_repos()`

skshetry:
> Maybe we could have a new fixture `erepo_raw_git_dir`?

I think, it makes sense to create a *git*-only repo, as we are supporting more and more non-DVC good repositories.
But, for now, as it's only a single test, I guess you could do the following:
```python
from tests.dir_helpers import TmpDir
path = TmpDir()

with path.chdir():
    path.scm_gen({"file": "file"}, "first version", commit="first version")
```

this is done here, you could take it as an example:
https://github.com/iterative/dvc/blob/049367e4751238363ee110d6c867d5b74de67009/tests/func/test_ignore.py#L136-L139

fabiosantoscode:
@skshetry this doesn't work. TmpDir doesn't get an scm by default, the only one that does is `tmp_dir`. I'll add a new fixture because it's easier, and there's at least one other test where it can be used.

skshetry:
> @skshetry this doesn't work. TmpDir doesn't get an scm by default, the only one that does is `tmp_dir`. I'll add a new fixture because it's easier, and there's at least one other test where it can be used.
@fabiosantoscode, sorry for misleading. Didn't really try. `

tests/func/test_update.py
fabiosantoscode: Using new `git_dir` dir helper now.

tests/func/test_update.py
fabiosantoscode: Now done through `git_dir` fixture

tests/func/test_update.py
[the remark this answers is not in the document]
fabiosantoscode: Done!

fabiosantoscode:
@skshetry @mroutis should be a lot cleaner now with the `git_dir` fixture yielding a simple git repository.

DeepSource is complaining about me changing the `.scm` property of the TmpDir, does it like java-style setters?

skshetry:
> DeepSource is complaining about me changing the `.scm` property of the TmpDir, does it like java-style setters?
@fabiosantoscode, looked through it, it's fine. Deepsource complains old stuff from the file that we have changed. We do the same `scm` injection on other fixtures.

tests/dir_helpers.py
```diff
@@ -285,3 +293,21 @@ def erepo_dir(tmp_path_factory, monkeypatch):
         path.dvc.close()
 
     return path
+
+
+@pytest.fixture
+def git_dir(tmp_path_factory, monkeypatch):
```
pared: I would suggest renaming it so that we know its "external git dir".
Also, if we are already initializing git repository here, how about refactoring `erepo_dir` so that it uses this fixture? If you consider my point valid, `path` generation should have logic basing on `request.fixturenames` (e.g. https://github.com/iterative/dvc/blob/482473aa69fd97ebde8bdd2ae23b0c703158f765/tests/dir_helpers.py#L223)

fabiosantoscode: I don't think it's very helpful to make this one react to other things. For example, the test `test_update_git_tracked` which I've added, needs DVC on the tmp_dir, but not on the external git dir. This would make this fixture useless for that test, since it would have DVC anyway.

We could refactor erepo_dir to use this fixture, but that would mean two commits, and the tests are pretty slow as it is right now.

tests/func/test_status.py
```diff
@@ -61,27 +61,25 @@ def test_status_non_dvc_repo_import(tmp_dir, dvc, erepo_dir):
...
-        erepo_dir.scm.repo.index.remove(["file"])
+    with external_git_dir.chdir():
+        external_git_dir.dvc = Repo.init()
+        external_git_dir.scm.repo.index.remove(["file"])
         os.remove("file")
-        erepo_dir.dvc_gen("file", "second version")
-        erepo_dir.scm.add([".dvc", "file.dvc"])
-        erepo_dir.scm.commit("version with dvc")
-        new_rev = erepo_dir.scm.get_rev()
+        external_git_dir.dvc_gen("file", "second version")
```
mroutis: Adding the `commit` kwarg is equivalent to `add` + `commit`.
```suggestion
        external_git_dir.dvc_gen("file", "second version", commit="version with dvc")
```

tests/dir_helpers.py
```diff
@@ -285,3 +293,21 @@ def erepo_dir(tmp_path_factory, monkeypatch):
         path.dvc.close()
 
     return path
+
+
+@pytest.fixture
+def external_git_dir(tmp_path_factory, monkeypatch):
```
Suor: Making a new fixture for a single test is an overkill, doing some copy-paste inside is even worse. You may simply use `erepo_dir` and remove `.dvc` dir there.

Suor:
I suggest stopping on this until https://github.com/iterative/dvc/issues/3124 is resolved, at least `external_repo()` and `cached_clone()` are somehow unified, that one will either solve this issue automatically or make fixing it a no-brainer. 

Trying to fix it using `cached_clone()` is a way to multiply a mess we have, i.e. counterproductive. So I suggest closing this PR.

efiop:
Closing in favor of https://github.com/iterative/dvc/pull/3190 . Sorry for wasted effort, @fabiosantoscode !

Issue #3177, opened 2020-01-17
Add parallelism options to speed up dvc add 


Ykid:
One of the project I did contains directories of more than 1m files. When I do `dvc add dir1`, it takes quite a while to run and I saw that dvc only utilize one cpu to do so, it would be nice if there's parallelism options that can utilize multiple cores of the machine.

Ykid:
current workaround: use multiple terminals to do `dvc add`

efiop:
Hi @Ykid !

> current workaround: use multiple terminals to do dvc add

Not sure how that is going to help to parallelize single `dvc add dir`. Or maybe you meant `dvc add dir1 dir2 ...`?

mroutis:
> Not sure how that is going to help to parallelize single dvc add dir. Or maybe you meant dvc add dir1 dir2 ...?

@efiop , it is `dvc add dir_1 dir_2 dir_n` (clear it out on the Discord conversation, forgot to update it here)

efiop:
@Ykid Also, what dvc was showing when you were looking at CPU utilization? We do run things like checksum computation in parallel https://github.com/iterative/dvc/blob/master/dvc/remote/base.py#L178 .

mroutis:
@efiop , from the discord convo again, @Ykid mentioned 1 CPU

efiop:
Discord: https://discordapp.com/channels/485586884165107732/563406153334128681/667622134100787210

Ykid:
@efiop this is what I got

```
ps aux | grep dvc
> some-user    1980 99.2  3.9 3483664 2634612 pts/14 Rl+ 13:57  56:44 python dvc add dir1
```

efiop:
@Ykid Yes, but what was it doing at that time? Meaining what it was showing to stdout? Was it saying that it is computing checksums or maybe that it is linking?

Ykid:
should be copying stuff

efiop:
For the record: decided to try out hardlinks or symlinks as a workaround.

efiop:
For the record: linking is still single-threaded in dvc, so need to look into parallelizing it as well.

Suor:
@efiop might a part or a follow up after checkout refactor

arrowkato:
I agree with Ykid.

As with dvc fetch and other commands, I hope that `dvc add` is added the --jobs option which allow it to execute multi- thread processing.
[...]

Issue #1843, opened 2019-04-07
logger: use logging format interpolation


ghost:
E.g.

```
logger.debug("Saving {} to {}.".format(path_info, to_info))
```
should be
```
logger.debug("Saving '%s' to '%s'.", path_info, to_info)
```

~~As suggested by @ei-grad in this PR: https://github.com/iterative/dvc/pull/1822~~

~~Reference: https://docs.python.org/3/howto/logging.html#optimization~~

~~It needs to support braces `{}` formatting, you can use a log adapter for Python 2.7:~~
<s>```python
import logging

class Message(object):
    def __init__(self, fmt, args):
        self.fmt = fmt
        self.args = args

    def __str__(self):
        return self.fmt.format(*self.args)

class StyleAdapter(logging.LoggerAdapter):
    def __init__(self, logger, extra=None):
        super(StyleAdapter, self).__init__(logger, extra or {})

    def log(self, level, msg, *args, **kwargs):
        if self.isEnabledFor(level):
            msg, kwargs = self.process(msg, kwargs)
            self.logger._log(level, Message(msg, args), (), **kwargs)
```</s>

shcheklein:
Consider introducing actual laziness - https://stackoverflow.com/questions/4148790/lazy-logger-message-string-evaluation . I'm not sure though it can be done w/o introducing some ugly wrapper.

mroutis:
Closed by #3178

efiop:
@mroutis Not closed, as there are plenty of placed that still `.format()` strings. Hence why I didn't include "Closing" into that PR. 🙂  Reopening.

PR #3178, opened 2020-01-17
logger: remote: use lazy formatting

efiop:
Part of #1843 , related to https://github.com/iterative/dvc/issues/3177

As our previous investigations showed, path stringification and stuff
like relpath are taking a very large chunk of time when working with
giant directories. This patch removes `format()`s and uses lazy
formatting provided by logger instead, so stuff like path_info is not
stringified until actually needed (e.g. on `-v`).

Benchmark results coming soon

### DIFF
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -24,7 +24,7 @@ from dvc.path_info import PathInfo, URLInfo
 from dvc.progress import Tqdm
 from dvc.remote.slow_link_detection import slow_link_guard
 from dvc.state import StateNoop
-from dvc.utils import relpath, tmp_fname
+from dvc.utils import tmp_fname
 from dvc.utils.fs import move, makedirs
 from dvc.utils.http import open_url
 
@@ -275,8 +275,10 @@ class RemoteBASE(object):
             raise DirCacheError(checksum) from exc
 
         if not isinstance(d, list):
-            msg = "dir cache file format error '{}' [skipping the file]"
-            logger.error(msg.format(relpath(path_info)))
+            logger.error(
+                "dir cache file format error '{}' [skipping the file]",
+                path_info,
+            )
             return []
 
         for info in d:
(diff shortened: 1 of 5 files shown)


efiop:
Benchmark:
```
#!/bin/bash

set -e
set -x

rm -rf myrepo
mkdir myrepo
cd myrepo

git init 
dvc init

mkdir dir
for i in {1..100000}; do
        echo $i > dir/$i
done

python -mcProfile -s cumtime -m dvc add dir &> log.log
```

~242 sec with this patch and ~282 without it. E.g. `path_info.py:49(__str__)` went from 53 sec to 24sec. Pretty nice. Need to look into `state` too.

efiop:
For the record: used incorrect formatting, fixing right now...

dvc/remote/base.py
```diff
@@ -275,8 +275,10 @@ def load_dir_cache(self, checksum):
             raise DirCacheError(checksum) from exc
 
         if not isinstance(d, list):
-            msg = "dir cache file format error '{}' [skipping the file]"
-            logger.error(msg.format(relpath(path_info)))
+            logger.error(
+                "dir cache file format error '{}' [skipping the file]",
+                path_info,
+            )
```
Suor: Still use msg maybe?
```python
msg = "dir cache file format error '{}' [skipping the file]"
logger.error(msg, path_info)
```

dvc/remote/base.py
```diff
@@ -342,36 +344,35 @@ def changed(self, path_info, checksum_info):
...
         if checksum != actual:
             logger.debug(
-                "checksum '{}'(actual '{}') for '{}' has changed.".format(
-                    checksum, actual, path_info
-                )
+                "checksum '{}'(actual '{}') for '{}' has changed.",
+                checksum,
+                actual,
+                path_info,
             )
```
Suor: Same as above:
```python
msg = "checksum '{}'(actual '{}') for '{}' has changed."
logger.debug(msg, checksum, actual, path_info)
```

dvc/rwlock.py
```diff
@@ -9,7 +9,7 @@
 
 from .exceptions import DvcException
 from .lock import LockError
-from .utils.fs import relpath
+from .utils import relpath
```
Suor: Are we trying to move this?

efiop:
> Did you try to bench it?

@Suor https://github.com/iterative/dvc/pull/3178#issuecomment-575550466

dvc/remote/base.py
efiop: No point in a new var, it is an old code from time when we weren't sure on formatting. We are inlining it these days, so it makes sense to inline here as well.

dvc/rwlock.py
efiop: @Suor It is a faulty import from an incorrect place that broke once I've removed no longer used `relpath` re-import from fs.py.

PR #3186, opened 2020-01-17
s3: copy large objects without chunking

### LINKED ISSUE #3174: remote s3 caching uses multipart copy which may cause an ETag mismatch
DVC version: 0.81.3, ubuntu 18.04 LTS, pip install, python 3.7.4

When specifying a remote object on s3 as a cached output, an ETag mismatch error is sometimes raised during the caching stage. E.g., observe the following debug output.

```bash
DEBUG: Removing s3://XXXXX/artifacts/test/train/models/model.01.ckpt.data-00000-of-00002
DEBUG: Created 'copy': s3://XXXXX/s3cache/ee/28f5ee86fc9aaca1aca65e64abde58 -> s3://XXXXX/artifacts/test/train/models/model.01.ckpt.data-00000-of-00002
DEBUG: cache 's3://XXXXX/s3cache/82/ed1912264aec805954939a0c84b5ff' expected '82ed1912264aec805954939a0c84b5ff' actual 'None'
DEBUG: SELECT count from state_info WHERE rowid=?
DEBUG: fetched: [(429,)]
DEBUG: UPDATE state_info SET count = ? WHERE rowid = ?
ERROR: failed to run command - ETag mismatch detected when copying file to cache! (expected: '82ed1912264aec805954939a0c84b5ff', actual: '284b1e7130c3689baad589f9de093810-11')
```

In this example,
`s3://XXXXX/artifacts/test/train/models/model.01.ckpt.data-00000-of-00002` is 81.4 MB. When `dvc` attempts to cache the file, it produces a multipart copy with a mismatched ETAG.

I believe this is because `s3.copy` https://github.com/iterative/dvc/blob/482473aa69fd97ebde8bdd2ae23b0c703158f765/dvc/remote/s3.py#L172 uses a default chunk size and chunk threshold of 8MB.

I've patched the problem locally for myself with the following lines:
```python
import boto3
s3.copy(
    source, to_info.bucket, to_info.path, ExtraArgs=extra_args,
    Config=boto3.s3.transfer.TransferConfig(
        multipart_threshold=1024**3
   )
)
```
There are probably more elegant methods to avoid a multipart copy.  For instance, https://stackoverflow.com/a/38058798 suggests using `put_object`.

woodshop:
When executing `boto3.s3.Client.Copy`, large files are chunked into
multiple parts. The default chunk threshold and size is 8MB. This
caused an issue when caching an object larger than 8M, where the
original object had not been uploaded in multiple parts. The original
object and the cached object would have different ETags, causing an
exception. This fixes the issue by dynamically setting the chunk
threshold to be one byte larger than the size of the original object
therefore avoiding multipart copies.

Fixes #3174

* [:white_check_mark:] ❗ Have you followed the guidelines in the [Contributing to DVC](https://dvc.org/doc/user-guide/contributing/core) list?

* [:white_check_mark:] 📖 Check this box if this PR **does not** require [documentation](https://dvc.org/doc) updates, or if it does **and** you have created a separate PR in [dvc.org](https://github.com/iterative/dvc.org) with such updates (or at least opened an issue about it in that repo). Please link below to your PR (or issue) in the [dvc.org](https://github.com/iterative/dvc.org) repo.

* [:white_check_mark:] ❌ Have you checked DeepSource, CodeClimate, and other sanity checks below? We consider their findings recommendatory and don't expect everything to be addressed. Please review them carefully and fix those that actually improve code or fix bugs.

### DIFF
--- a/dvc/remote/s3.py
+++ b/dvc/remote/s3.py
@@ -3,6 +3,7 @@
 import logging
 import os
 import threading
+from boto3.s3.transfer import TransferConfig
 
 from funcy import cached_property, wrap_prop
 
@@ -169,7 +170,13 @@ class RemoteS3(RemoteBASE):
             )
         else:
             source = {"Bucket": from_info.bucket, "Key": from_info.path}
-            s3.copy(source, to_info.bucket, to_info.path, ExtraArgs=extra_args)
+            s3.copy(
+                source,
+                to_info.bucket,
+                to_info.path,
+                ExtraArgs=extra_args,
+                Config=TransferConfig(multipart_threshold=size + 1)
+            )
 
         cached_etag = cls.get_etag(s3, to_info.bucket, to_info.path)
         if etag != cached_etag:


efiop:
Thank you @woodshop ! 🙏

dvc/remote/s3.py
```diff
@@ -3,6 +3,7 @@
 import logging
 import os
 import threading
+from boto3.s3.transfer import TransferConfig
```
skshetry: Shouldn't this import be inside `_copy()`?

shcheklein: good catch @skshetry ! I think it's better to move inside indeed. Mind creating a PR for this?

mroutis: @skshetry , got you) https://github.com/iterative/dvc/pull/3195

Good catch, indeed!

efiop: Totally missed that one! Thank you @skshetry !

PR #3189, opened 2020-01-18
test: drop test that should never existed

Suor:
- it tests for external_repo() behavior indirectly
- fragile text comparison

Once we support git files we will write a proper test. If we don't we
should test `external_repo()` throws approproate error in some unit test
instead or leave it as is.


Suor:
Things have changed in #3190, test should be adjusted.

PR #3190, opened 2020-01-18
external: refactor external_repo() and its users

### LINKED ISSUE #2824: Refactor handling external repo errors in get command
Catching DVC repo exceptions and rethrowing them in `get` ([here](https://github.com/iterative/dvc/blob/f563892760018a3e2ccb807558df99043d097252/dvc/repo/get.py#L59) and [here](https://github.com/iterative/dvc/blob/f563892760018a3e2ccb807558df99043d097252/dvc/repo/get.py#L61)) should probably be handled in [external repo](https://github.com/iterative/dvc/blob/f563892760018a3e2ccb807558df99043d097252/dvc/external_repo.py#L28). Also, it seems to me that now there is no situation where OutputNotFound will not be converted to `NoOutputInExternalRepoError`.

### LINKED ISSUE #2976: update: Handle Git-tracked import-ed files

### LINKED ISSUE #3124: import/get/api: external repo and cached_clone are problematic

### LINKED ISSUE #3179: api/get: local repo with custom rev is not supported
```python
import dvc.api

dvc.api.get_url("file", repo=".", rev="HEAD")
```
throws following error:
```
AssertionError: Custom revision is not supported for local repo
```

Similarly, issue exists with `dvc get --show-url` (introduced by #3156):
```
➜ dvc get . foo --rev=HEAD@{5.days.ago} --show-url
ERROR: unexpected error - Custom revision is not supported for local repo
```

Suor:
Some things are fixed along the way:
- no cache shared between incompatible funcs
- some incosistent exceptions are now consistent
- import updates for git files now works both for dvc and git repos
- "auto-generated-upstream already exists" fixed
- api.open()/read() works with git repos now
- api funcs allow specifying `rev` for local repos now

Fixes #3124, fixes #2976, fixes #3179, and fixes #2824. Closes #3172.

### DIFF
--- a/dvc/api.py
+++ b/dvc/api.py
@@ -109,7 +109,7 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
 
 @contextmanager
 def _make_repo(repo_url, rev=None):
-    if not repo_url or os.path.exists(repo_url):
+    if rev is None and (not repo_url or os.path.exists(repo_url)):
         yield Repo(repo_url)
     else:
         with external_repo(url=repo_url, rev=rev) as repo:
--- a/dvc/api.py
+++ b/dvc/api.py
@@ -43,7 +43,7 @@ class SummonError(DvcException):
     pass
 
 
-class UrlNotDvcRepoError(DvcException):
+class UrlNotDvcRepoError(NotDvcRepoError):
     """Thrown if given url is not a DVC repository.
 
     Args:
@@ -60,14 +60,11 @@ def get_url(path, repo=None, rev=None, remote=None):
     `repo`.
     NOTE: There is no guarantee that the file actually exists in that location.
     """
-    try:
-        with _make_repo(repo, rev=rev) as _repo:
-            abspath = os.path.join(_repo.root_dir, path)
-            out, = _repo.find_outs_by_path(abspath)
-            remote_obj = _repo.cloud.get_remote(remote)
-            return str(remote_obj.checksum_to_path_info(out.checksum))
-    except NotDvcRepoError:
-        raise UrlNotDvcRepoError(repo)
+    with _make_repo(repo, rev=rev) as _repo:
+        _require_dvc(_repo)
+        out = _repo.find_out_by_relpath(path)
+        remote_obj = _repo.cloud.get_remote(remote)
+        return str(remote_obj.checksum_to_path_info(out.checksum))
 
 
 def open(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
@@ -96,9 +93,8 @@ class _OpenContextManager(GCM):
 
 def _open(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
     with _make_repo(repo, rev=rev) as _repo:
-        abspath = os.path.join(_repo.root_dir, path)
-        with _repo.open(
-            abspath, remote=remote, mode=mode, encoding=encoding
+        with _repo.open_by_relpath(
+            path, remote=remote, mode=mode, encoding=encoding
         ) as fd:
             yield fd
 
@@ -114,9 +110,6 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
 @contextmanager
 def _make_repo(repo_url, rev=None):
     if not repo_url or os.path.exists(repo_url):
-        assert (
-            rev is None
-        ), "Git revisions are not supported for local DVC projects."
         yield Repo(repo_url)
     else:
         with external_repo(url=repo_url, rev=rev) as repo:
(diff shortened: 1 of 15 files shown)


efiop:
@Suor Please rebase.

dvc/api.py
```diff
@@ -56,14 +56,11 @@ def __init__(self, url):
...
-    try:
-        with _make_repo(repo, rev=rev) as _repo:
-            abspath = os.path.join(_repo.root_dir, path)
-            out, = _repo.find_outs_by_path(abspath)
-            remote_obj = _repo.cloud.get_remote(remote)
-            return str(remote_obj.checksum_to_path_info(out.checksum))
-    except NotDvcRepoError:
-        raise UrlNotDvcRepoError(repo)
+    with _make_repo(repo, rev=rev) as _repo:
+        _require_dvc(_repo)
```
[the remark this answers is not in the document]
Suor: It doesn't throw it anymore because we support git only repos transparently now. No need to call `external_repo()` and `cached_clone()` separately.

Suor:
@efiop rebased. Moved no remote/cache present fix to new code, left a TODO there because issues tend to get lost. I will fix it in a subsequent PR, doesn't look small enough to fit here.

efiop:
@Suor TODOs get lost more, please remove it and create an issue.

efiop:
@Suor there is some problem with an author of your commits, what is going on?

Suor:
We have lots of dead issues and no lost TODOs )

Suor:
rebased

dvc/api.py
```diff
@@ -113,10 +109,7 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
...
 @contextmanager
 def _make_repo(repo_url, rev=None):
-    if not repo_url or os.path.exists(repo_url):
-        assert (
-            rev is None
-        ), "Git revisions are not supported for local DVC projects."
+    if rev is None and (not repo_url or os.path.exists(repo_url)):
         yield Repo(repo_url)
     else:
         with external_repo(url=repo_url, rev=rev) as repo:
```
jorgeorpinel: Why is `external_repo()` used if there's a `rev`? What if I want to access a file in the current DVC repo but from a previous commit? I'm trying to test this but I'm getting `AttributeError: module 'dvc' has no attribute 'api'` when running:

```py
import dvc

print(dvc.api.read('data'))
```
Using DVC 0.87.0 installed with `pip` in a py3 virtualenv.

@Suor @skshetry @shcheklein

jorgeorpinel: UPDATE: Nevermind about the Python error. I just had to `import dvc.api`

skshetry: @jorgeorpinel, it should just clone the current repo and checkout to given `rev` and access that file.

jorgeorpinel: Testing this, I see it works locally anyway. But is it using `external_repo()` to clone the current repo in a temporary folder and then use that instead of just the current Git repo's tree?

jorgeorpinel: p.s. my test script:

```py
import dvc.api

print(dvc.api.read('data', rev='HEAD^'))
```
`data` should be a DVC-tracked file with 2 different consecutive versions (in `HEAD^` and in `HEAD`).

jorgeorpinel: My point is I think `Repo()` also accepts `rev`, just wondering why it's not passed, and whether that would be a better implementation.

skshetry: `Repo` does not accept `rev`. Repo is all about current workspace, so, you cannot just checkout the user's workspace and modify/read. So, even though it's still is a `Repo` underneath (if it's a DVC repo), it's cloned, checked out to a `rev`, put into a `tmp` cache directory and operated inside of it.

jorgeorpinel: Got it. I was just confused by the name `external_repo` then. I guess that's the only way to get a previous rev from the current repo unfortunately.

PR #3193, opened 2020-01-19
config: verify before saving a modification

### LINKED ISSUE #3192: config: check schema before applying modifications
**Version:**

```
DVC version: 0.81.3+7f5e4a
Python version: 3.7.5
Binary: False
Package: None
```

**Reproduce:**

```
dvc remote add test https://example.com
dvc remote modify test random_key
dvc version
```

will produce: `ERROR: configuration error - config file error: extra keys not allowed`

Probably we should do a check before we set the value.

mroutis:
Close #3192

### DIFF
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -385,7 +385,7 @@ class Config(object):  # pylint: disable=too-many-instance-attributes
         if not self.should_validate:
             return
 
-        d = self.validate()
+        d = self.validate(self.config)
         self.config = configobj.ConfigObj(d, write_empty_values=True)
 
     def save(self, config=None):
@@ -423,14 +423,9 @@ class Config(object):  # pylint: disable=too-many-instance-attributes
                 raise
         config.write()
 
-    def validate(self, config=None):
-        ret = copy.deepcopy(self.config)
-
-        if config:
-            ret.merge(config)
-
+    def validate(self, config):
         try:
-            return self.COMPILED_SCHEMA(ret.dict())
+            return self.COMPILED_SCHEMA(config.dict())
         except Invalid as exc:
             raise ConfigError(str(exc)) from exc
 
@@ -499,7 +494,10 @@ class Config(object):  # pylint: disable=too-many-instance-attributes
 
         config[section][opt] = value
 
-        self.validate(config)
+        result = copy.deepcopy(self.config)
+        result.merge(config)
+        self.validate(result)
+
         self.save(config)
 
     def get(self, section, opt=None, level=None):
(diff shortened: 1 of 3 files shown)


dvc/config.py
```diff
@@ -491,6 +495,8 @@ def set(self, section, opt, value, level=None, force=True):
             )
 
         config[section][opt] = value
+
+        self.validate(config)
```
efiop: Ok, but now if you have an already broken config, you won't be able to fix it with `dvc config`, because you will be hit with this error. Plus merging configs will be broken too (e.g. url is in config.local but some params are in config, which will raise ConfigError now if you try to modify `config`, because there is no URL there).

mroutis: @efiop , `dvc config --unset` or `dvc remote modify --unset` won't work if you entered an invalid key.

```console
$ dvc init --no-scm
$ dvc remote add example https://example.com
$ dvc remote modify example invalid value
$ dvc version
ERROR: configuration error - extra keys not allowed
$ dvc remote modify --unset example invalid
ERROR: configuration error - extra keys not allowed
```

Here's the resulting config file.
```ini
['remote "example"']
url = https://example.com
invalid = None
```

Although, It might be a problem with `--unset`.

---

> Plus merging configs will be broken too.

You are totally right. Fixing it on the next commit.

efiop: @mroutis Yes, it won't work with your PR, that is what I'm saying. Before your PR it did work, so if your colleague messed something up, you could unset stuff (e.g. started using some new config option from a newer dvc release). Maybe that is an ok approach, not sure.

tests/func/test_remote.py
```diff
@@ -257,3 +257,18 @@ def test_modify_missing_remote(dvc):
...
+
+    with pytest.raises(ConfigError, match=r"extra keys not allowed"):
+        remote_config.modify("example", "random_key", "value")
+
+
+def test_merging_two_levels(tmp_dir, dvc):
+    assert 0 == main(["remote", "add", "test", "https://example.com"])
+    assert 0 == main(["remote", "modify", "--local", "test", "password", "1"])
+    assert "example.com" in (tmp_dir / ".dvc" / "config").read_text()
+    assert "password" in (tmp_dir / ".dvc" / "config.local").read_text()
```
efiop: Your change is about `dvc config` in general, not `dvc remote` specifically. So might be worth testing there. Also, there is a somewhat decent API to config, so no need to use it through CLI, just check for specific exceptions instead. This is a note about tests, but the more imporant issue is with the design of `dvc config` overall.

dvc/config.py
mroutis: @efiop , what do you mean? The example that I shared with `--unset` doesn't work on master.

tests/func/test_remote.py
mroutis: @efiop , moved the tests to config.

efiop:
Thanks!

PR #3194, opened 2020-01-19
remote: verify repository exists when level requires it

### LINKED ISSUE #2309: config: `dvc config` doesn't throw an error when used outside of a dvc repository without --system or --global
```
DVC version: 0.52.1+1e9f1d
Python version: 3.7.3
Platform: Linux-5.2.1-arch1-1-ARCH-x86_64-with-arch
Binary: False
Filesystem type (workspace): ('xfs', '/dev/mapper/vg-root')
```
---

```bash
rm -rf .dvc
dvc remote modify something whatever yeah
```

#### Expected
```
ERROR: you are not inside of a dvc repository (checked up to mount point '/')
```

### LINKED ISSUE #3191: remote add: does not fail in non-DVC repo
*Version:*

```
DVC version: 0.81.3+7f5e4a
Python version: 3.7.5
Binary: False
Package: None
```

*Reproduce:*

```
mkdir project
cd project
dvc remote add storage gdrive://root/storage
```

should probably fail.

next command fails:

```
dvc remote modify storage gdrive_client_id test
ERROR: configuration error - config file error: unable to find remote section 'storage'
```

message should be different - no config file, or repo is not DVC, etc.

mroutis:
Close #2309

### DIFF
--- a/dvc/command/remote.py
+++ b/dvc/command/remote.py
@@ -1,7 +1,6 @@
 import argparse
 import logging
 
-from dvc.repo import Repo
 from dvc.command.base import append_doc_link
 from dvc.command.base import fix_subparsers
 from dvc.command.config import CmdConfig
@@ -14,12 +13,6 @@ class CmdRemoteConfig(CmdConfig):
     def __init__(self, args):
         super().__init__(args)
         self.remote_config = RemoteConfig(self.config)
-
-        # Verify repository existance when level requires to modify its config
-        repository_levels = [self.config.LEVEL_REPO, self.config.LEVEL_LOCAL]
-
-        if self.args.level in repository_levels:
-            Repo.find_dvc_dir()
 
 
 class CmdRemoteAdd(CmdRemoteConfig):
--- a/dvc/command/remote.py
+++ b/dvc/command/remote.py
@@ -1,6 +1,7 @@
 import argparse
 import logging
 
+from dvc.repo import Repo
 from dvc.command.base import append_doc_link
 from dvc.command.base import fix_subparsers
 from dvc.command.config import CmdConfig
@@ -13,6 +14,12 @@ class CmdRemoteConfig(CmdConfig):
     def __init__(self, args):
         super().__init__(args)
         self.remote_config = RemoteConfig(self.config)
+
+        # Verify repository existance when level requires to modify its config
+        repository_levels = [self.config.LEVEL_REPO, self.config.LEVEL_LOCAL]
+
+        if self.args.level in repository_levels:
+            Repo.find_dvc_dir()
 
 
 class CmdRemoteAdd(CmdRemoteConfig):
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -250,7 +250,7 @@ class Config(object):  # pylint: disable=too-many-instance-attributes
     }
     COMPILED_SCHEMA = Schema(SCHEMA)
 
-    def __init__(self, dvc_dir=None, validate=True):
+    def __init__(self, dvc_dir=None, validate=True, verify_repo=False):
         self.dvc_dir = dvc_dir
         self.validate = validate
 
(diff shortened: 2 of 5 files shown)


dvc/command/remote.py
```diff
@@ -14,6 +15,12 @@ def __init__(self, args):
         super().__init__(args)
         self.remote_config = RemoteConfig(self.config)
 
+        # Verify repository existance when level requires to modify its config
+        repository_levels = [self.config.LEVEL_REPO, self.config.LEVEL_LOCAL]
```
mroutis: Created a variable because _in-linening_ it exceeded 80 chars.

tests/func/test_remote.py
```diff
@@ -257,3 +257,22 @@ def test_modify_missing_remote(dvc):
...
+        ["remote", "remove", "example"],
+        ["remote", "remove", "--local", "example"],
+        ["remote", "list"],
+        ["remote", "list", "--local"],
+        ["remote", "default", "example"],
+        ["remote", "default", "--local", "example"],
+    ],
+)
+def test_commands_outside_dvc_repository(tmp_dir, cmd):
+    assert 253 == main(cmd)
```
mroutis: `253` is returned when `NotDvcRepoError` is raised.

dvc/config.py
```diff
@@ -250,7 +250,7 @@ class Config(object):  # pylint: disable=too-many-instance-attributes
     }
     COMPILED_SCHEMA = Schema(SCHEMA)
 
-    def __init__(self, dvc_dir=None, validate=True):
+    def __init__(self, dvc_dir=None, validate=True, verify_repo=False):
```
mroutis: Added a new flag because I didn't want to mess up with any previous logic dealing with having an empty `dvc_dir`.

efiop: @mroutis This looks like a hack rather than a proper solution 🙂

tests/func/test_remote.py
```diff
@@ -257,3 +257,22 @@ def test_modify_missing_remote(dvc):
 
     with pytest.raises(ConfigError, match=r"unable to find remote section"):
         remote_config.modify("myremote", "gdrive_client_id", "xxx")
+
+
+@pytest.mark.parametrize(
```
efiop: Solution is for config in general, but tests are for `remote` specifically, not nice 🙂Also not clear why are you testing CLI again and not API?

dvc/config.py
mroutis: @efiop, could you explain? Not sure how to proceed with such observation 😅

tests/func/test_remote.py
mroutis: My bad, that's a left over. Those tests are not relevant since they were replaced for the ones in `test_config.py`

mroutis: :see_no_evil: I got confused with the other pull-request:

https://github.com/iterative/dvc/pull/3193/files#diff-dcfb923229ccfbba150bdd77494f46eeR96-R97

mroutis: I prefer testing the CLI in this case, it is more complete, but if you want, I use the API.

efiop: @mroutis CLI is a dumb wrapper for API. We always test API, so let's do the same here. If you want to check that CLI returns proper codes on corresponding exceptions, just add a unit test to `tests/unit/command/`.

dvc/config.py
efiop: You are hacking your way through the bug instead of solving the bug. This `verify_repo` shouldn't be an option at all. Config should always raise if you try to modify in-repo config when there isn't one.

efiop:
Ok, looks like there is no good quick and neat way to solve it, it will look like a hack in any case. What we need here is a redesign. :slightly_frowning_face:

efiop:
Closing for now, this will require a config redesign first.

PR #3195, opened 2020-01-19
s3: import boto3 inside _copy instead of module level

mroutis:
Kudos to @skshetry for bringing this out during a review: https://github.com/iterative/dvc/pull/3186#discussion_r368209924

This is important, since someone can use DVC without having `boto3` installed.


efiop:
@skshetry @mroutis Thank you guys! Totally missed that one!

PR #3197, opened 2020-01-20
test: refactor tmp_dir/erepo_dir

Suor:
Provides `make_tmp_dir` fixture. A base for simplifying tests where we now remove `.dvc` in erepo. 

Cleaning up that code will conflict with #3190, so waiting for that.

### DIFF
--- a/tests/dir_helpers.py
+++ b/tests/dir_helpers.py
@@ -48,13 +48,13 @@ import pathlib
 from contextlib import contextmanager
 
 import pytest
-from funcy.py3 import lmap, retry
+from funcy import lmap, retry
 
 from dvc.utils.fs import makedirs
 from dvc.compat import fspath, fspath_py35
 
 
-__all__ = ["tmp_dir", "scm", "dvc", "run_copy", "erepo_dir"]
+__all__ = ["make_tmp_dir", "tmp_dir", "scm", "dvc", "run_copy", "erepo_dir"]
 
 
 class TmpDir(pathlib.Path):
@@ -72,6 +72,28 @@ class TmpDir(pathlib.Path):
     # Not needed in Python 3.6+
     def __fspath__(self):
         return str(self)
+
+    def init(self, *, scm=False, dvc=False):
+        from dvc.repo import Repo
+        from dvc.scm.git import Git
+
+        assert not scm or not hasattr(self, "scm")
+        assert not dvc or not hasattr(self, "dvc")
+
+        str_path = fspath(self)
+
+        if scm:
+            _git_init(str_path)
+        if dvc:
+            self.dvc = Repo.init(str_path, no_scm=True)
+        if scm:
+            self.scm = self.dvc.scm if hasattr(self, "dvc") else Git(str_path)
+        if dvc and hasattr(self, "scm"):
+            self.scm.commit("init dvc")
+
+    def close(self):
+        if hasattr(self, "scm"):
+            self.scm.close()
 
     def _require(self, name):
         if not hasattr(self, name):
@@ -151,10 +173,6 @@ class TmpDir(pathlib.Path):
         finally:
             self.scm.checkout(old)
 
-    # Introspection methods
-    def list(self):
-        return [p.name for p in self.iterdir()]
-
 
 def _coerce_filenames(filenames):
     if isinstance(filenames, (str, bytes, pathlib.PurePath)):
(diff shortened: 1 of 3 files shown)


tests/dir_helpers.py
```diff
@@ -81,6 +73,28 @@ def __new__(cls, *args, **kwargs):
     def __fspath__(self):
         return str(self)
 
+    def init(self, *, scm=False, dvc=False):
```
skshetry: How about having two methods: `TmpDir::dvc_init()` and `TmpDir::git_init` that only injects `dvc` and `scm` to the instance?

tests/dir_helpers.py
```diff
@@ -178,32 +188,36 @@ class PosixTmpDir(TmpDir, pathlib.PurePosixPath):
     pass
 
 
+@pytest.fixture(scope="session")
+def make_tmp_dir(tmp_path_factory, request):
```
skshetry: How about having a `raw_dir` that does no magic at all (with above suggestions https://github.com/iterative/dvc/pull/3197/files#r368376497), but only gives you an empty directory?

skshetry: Sorry, looks like this was the intention of the PR. We can have a separate `raw_dir` as well then. :slightly_smiling_face:

tests/dir_helpers.py
Suor: I tried, these two are interleaved too much, `.scm` construction depends on `.dvc` presense for example. Separate init methods make it less readable. You can look at current convoluted `scm` and `dvc` implementation.

tests/dir_helpers.py
Suor: There is a `tmp_path` pytest standard fixture.

Suor: This won't work when we need several of those. This is like `tmp_path` and `tmp_path_factory` from pytest.

tests/dir_helpers.py
skshetry: I might be missing something, but for, `TmpDir::dvc_init()`, wouldn't following work:
```python
self.dvc = Repo(path, no_scm=not hasattr(self, "scm"))
```
And, similarly, on `git_init()`: `self.scm = Git(str_path)`?

I just wanted it to be a bit flexible, but, without any magic, so that it'd be easier to just create a temp_dir and make anything out of it (such as git repo, dvc repo, or a normal directory).

Suor: Only I don't want to instantiate `Git` twice. This caused issues under Windows as far as I remember. So you can't assign `scm` in `init_scm()`, you need to postpone that. It ended up like this:
```python
class TmpDir(...):
    def __new__(...):
        # ...
        self.dvc = self.scm = None
        # ...

    @cached_property
    def scm(self):
        from dvc.scm.git import Git

        return self.dvc.scm if self.dvc else Git(fspath(self))

    def init_scm(self):
        assert not self.scm

        _git_init(fspath(self))
        del self.scm  # Might be dvc.scm or a separate Git instance so we delay

    def init_dvc(self):
        from dvc.repo import Repo
        from dvc.scm import NoSCM

        assert not self.dvc

        self.dvc = Repo.init(fspath(self), no_scm=True)
        if not isinstance(self.dvc.scm, NoSCM):
            self.scm = self.dvc.scm
            self.scm.commit("init dvc")
```

As I said they are too interleaved.

skshetry: Got it, thanks for the explanation.

tests/dir_helpers.py
```diff
@@ -212,45 +226,19 @@ def _git_init():
...
-
-@pytest.fixture
-def run_copy(tmp_dir, dvc, request):
+def run_copy(tmp_dir, dvc):
     tmp_dir.gen(
         "copy.py",
         "import sys, shutil\nshutil.copyfile(sys.argv[1], sys.argv[2])",
     )
 
     # Do we need this?
```
efiop: It is a leftover from your previous PRs that slipped-by, but is this still relevant or were you able to figure out? 🙂

Suor: This is a part of this PR

efiop: @Suor `    # Do we need this?` is a leftover that has been there before. Do we need it anymore?

Suor: Ah, not sure. But this question is independent from this PR )

Suor: Removing it doesn't break any tests though, it might if we will `run_copy` in context of some branches.

Suor: Removed it. Let's hope we won't need it)

PR #3200, opened 2020-01-20
fetch: trust remote checksums

### LINKED ISSUE #2676: option to have dvc pull trust the remote and not calculate its own MD5 hashes
The dvc pull command takes the time (which can be significant it seems) to calculate its own MD5 hashes. It would be nice if there were an option to tell it to trust the remote and skip this so that it would complete faster.  Or it could trust the remote by default and make you opt-in to recalculating the hashes if that is what you really wanted. It seems unlikely that it would be a bad breaking change to flip the default behavior here.  Or, maybe, we just don't need dvc pull to _ever_ calculate MD5 hashes. Thoughts?

pared:

Fixes #2676


dvc/data_cloud.py
```diff
@@ -73,13 +75,24 @@ def pull(self, cache, jobs=None, remote=None, show_checksums=False):
...
 
+        if not verify:
+            self._save_pulled_checksums(cache)
+
+        return downloaded_items_num
+
+    def _save_pulled_checksums(self, cache):
+        for checksum in cache["local"].keys():
+            cache_file = self.repo.cache.local.checksum_to_path_info(checksum)
+            if self.repo.cache.local.exists(cache_file):
```
efiop: We've talked about this being alright, as corrupted cache will be deleted prior to downloading (on `status`) and new corrupted files will stay in a temp file, so we are fine doing this, but let's put a comment to clarify that for anyone lurking around here in the future.

dvc/command/data_sync.py
```diff
@@ -145,6 +147,12 @@ def add_parser(subparsers, _parent_parser):
         default=False,
         help="Pull cache for subdirectories of the specified directory.",
     )
+    pull_parser.add_argument(
+        "--verify",
+        action="store_true",
+        default=False,
+        help="Verify downloaded files checksums.",
+    )
```
efiop: For the record: better to only provide per-remote config option for now instead of CLI.

tests/func/test_remote.py
```diff
@@ -257,3 +258,33 @@ def test_modify_missing_remote(dvc):
...
+        erepo_dir.dvc_gen(
+            {"dir": {"subfile": "file2 content"}}, commit="add " "dir"
+        )
+
+    Git.clone(fspath(erepo_dir), fspath(tmp_dir))
+
+    from dvc.repo import Repo
+
+    dvc = Repo(fspath(tmp_dir))
+
```
efiop: Not sure I understand why this is needed. COuld you elaborate, please?

pared: Well, to perform `pull` operation on "clear" repo I can either:
- create repo,
- create remote,
- create elements, add them and push
- remove local cache, remove local files
- perform pull
or:
- create elements in erepo
- clone it
- pull
 Second option seems faster

efiop: >     Git.clone(fspath(erepo_dir), fspath(tmp_dir))

will probably clone `erepo_dir` into `tmp_dir / "some_dir"`, or am i missing something?

efiop: Ah, yeah, it will clone to `tmp_dir` directly. Ok, need to give it a fresh look, so far I'm confused.

pared: Not quite, we wouldn't be able to initialize `Repo` with actual repo being in `some_dir` that would be subdir of `tmp_dir`. Also, double-checked `.dvc` is inside `tmp_dir`.

efiop: @pared Let's do it the other way around, please. I have to give it a very deep thought to figure out what is going on in this test, because it implies that erepo has remote already setup (which we will remove soon, as we have an auto-remote logic already).

dvc/data_cloud.py
```diff
@@ -73,13 +73,26 @@ def pull(self, cache, jobs=None, remote=None, show_checksums=False):
...
+                # We can safely save here, as existing corrupted files will be
+                # removed upon status, while files corrupted during download
+                # will not be copied from tmp_file (see download
```
efiop:
```suggestion
                # will not be moved from tmp_file (see download
```

Issue #2325, opened 2019-07-26
Reconsider gc implementation


pared:
As pointed out in discussion in #1691, we should reconsider `gc` implementation.
Currently, if called without any options, dvc will collect current branch dependencies and outputs checksums, and remove everything besides it. We can easily clear history of changes with this command. `gc` should be safer with default options. Straightforward implementation could get all outputs for all revisions in git repo and remove everything that is not on list.

As pointed out by @Suor, this approach might be slow for repository with long history.

Suor:
This is more about UI defaults then implementation, which will surely be affected too.

hhoeflin:
yes - very much agree. I am new to dvc, but the gc implementation is already making me nervous.

efiop:
Hi @hhoeflin ! Sorry for the delay.

> Should "metrics" files be treated differently than the rest? I agree that gc should easily collect old model files etc, but keeping the development of performance over time around would be nice.

We usually recommend keeping metrics files in the git (e.g. `dvc run -M`) for that precise reason plus if you'd lose access to your cache at some point, you would still be able to use git to see what the metric was like.

> I am not familiar in detail with the implementation, but the existence of the "--projects" parameter points to an interesting problem. What should happen to files that aren't referenced anywhere in the current git repository?

In the current state of things, those files are going to be removed, that is why `gc` needs `--projects` if you are using the same cache or remote for a number of projects. The perk of doing that is that you get inter-project deduplication of cache files, which might be handy :slightly_smiling_face: In other cases, we usually recommend keeping separate caches/remotes for your projects.

> a) are known where they are referenced

Currently dvc cache/remote doesn't have any notion of which project it came from, it would require some additional complications to make it aware of that.

> b) are cleared for deletion (e.g. by being in a certain branch or by having never versions for the same file in the a branch)

It is clear that we need different strategies for gc. Here are a few related tickets about other strategies: https://github.com/iterative/dvc/issues/678 https://github.com/iterative/dvc/issues/377 https://github.com/iterative/dvc/issues/155 https://github.com/iterative/dvc/issues/855
And we clearly need to adopt some reasonable default approach by gathering a number of strategies :slightly_smiling_face:

shcheklein:
An excellent overview and answer @efiop, thank you :)

[...]

Issue #1691, opened 2019-03-06
support push/pull/metrics/gc, etc across different commits


andrethrill:
Currently `dvc metrics show` can show metric values across different branches (`-a`) and different tags (`-T`).
Can you consider supporting showing different metric values across different commits in the same branch?

---

The background of this is (simplified example): say I'm currently training a model, where I'm changing a certain parameter, `param1` (for instance, number of trees in a forest). The way I probably would like to work is to find a first value for `param1`, commit the current state, continue changing `param1` and continue committing the successive states that I consider worth saving. At some point I would like to look back and identify the setup that gave me the best results.

The way DVC currently works forces me to create a new branch/tag for each trial I want to keep track of, and this seems a bit overwhelming.

Depending on how different the experiments I'm running are and their level of granularity I could decide how to keep track of them (new commits VS new branches/tags).

Notes:
- The example above is overly simplified and there are better ways of tuning specific models parameters. But this gets more complicated if I'm changing more stuff (model hyperparameters, data processing, features to use, etc).
- If dvc were to support what I'm proposing here, an extra argument would probably be required to limit how many commits DVC would look back at. Otherwise it would show all the metric values since the beginning of the repo history, which can be unhelpful and messy.

efiop:
@andrethrill Would you like to compare two specific commits or just the dynamics of your metrics changing across a range of commit? The latter one is probably more suitable for a graphical tool, like tensorboard or something.

andrethrill:
Hi @efiop !

I'm aware of TensorBoard but that's not exactly what I was talking about.

I would like to have a way of running a few consecutive different experiments and see their metrics. Just like `dvc metrics show -a` currently does but without needing to create different branches. DVC seems like a good fit for this because I could checkout the experiment that gave me the best results and have everything version controlled (model, data, etc).

> Or are you looking for a CLI way of doing that, using different filters (e.g. find max metric across N commits)?

If that were to be supported it would be great of course. But for what I'm talking about, just looking at the output in the same form as in `dvc metrics show -a` would be enough.

efiop:
@andrethrill Ah, so something like `dvc metrics show HEAD~10` to show 10 last commits on the current branch?

andrethrill:
Exactly @efiop ! And/or some other nice variations of it:  `dvc metrics show HEAD~{commitHash}` show metrics since `commitHash` on the current branch.

efiop:
[...]

PR #3207, opened 2020-01-21
[WIP] gc: change option to use single remove flags

skshetry:
Related to #2325.


dvc/command/gc.py
```diff
@@ -5,23 +5,46 @@
...
+            supported_filters = {"commits", "branches", "tags"}
+            removal_types = set(self.args.remove.split(","))
+
+            exclude_scm_filters = supported_filters - removal_types
+            unknown_options = removal_types - supported_filters
+            if unknown_options:
+                message = (
+                    "unknown option(s) specified for remove: {}. "
+                    "Expected combinations of: {}."
+                )
```
efiop: Maybe this could be handled by argparse choices?

dvc/command/gc.py
```diff
@@ -39,9 +62,9 @@ def run(self):
             return 1
 
         self.repo.gc(
-            all_branches=self.args.all_branches,
-            all_tags=self.args.all_tags,
-            all_commits=self.args.all_commits,
+            all_branches="branches" in exclude_scm_filters,
+            all_tags="tags" in exclude_scm_filters,
+            all_commits="commits" in exclude_scm_filters,
```
efiop: Sorry, we didn't clarify it initially, but we need to adjust API as well, to have that same default behavior as CLI

dvc/command/gc.py
```diff
@@ -5,23 +5,46 @@
...
 
 logger = logging.getLogger(__name__)
 
 
 class CmdGC(CmdBase):
     def run(self):
-        msg = "This will remove all cache except items used in "
+        has_scm = not isinstance(self.repo.scm, NoSCM)
+        exclude_scm_filters = {"commits"} if has_scm else set()
+
```
efiop: Oh, good one!

dvc/repo/gc.py
```diff
@@ -41,12 +42,13 @@ def gc(
 
         used = NamedCache()
         for repo in all_repos + [self]:
+            has_scm = not isinstance(repo.scm, NoSCM)
             used.update(
                 repo.used_cache(
-                    all_branches=all_branches,
+                    all_branches=not all_branches and has_scm,
```
skshetry: I decided not to change the arguments name, as it makes sense to be more aggressive by setting the arguments here. and looks to be good name. The other idea I had was `scm_exclude=None` where we could pass *Collections* (set/tuple/list/set/dict, etc).

dvc/command/gc.py
```diff
@@ -94,6 +83,12 @@ def add_parser(subparsers, parent_parser):
     gc_parser.add_argument(
         "-r", "--remote", help="Remote storage to collect garbage in."
     )
+    gc_parser.add_argument(
```
skshetry: I decided to go on with choices here instead of manual labor.

dvc/command/gc.py
```diff
@@ -94,6 +83,12 @@ def add_parser(subparsers, parent_parser):
...
+        "--remove",
+        nargs="*",
+        choices=supported_removal_filters,
+        help="Available choices for remove: %(choices)s",
```
efiop: I guess this `help` should describe different choices, not just list them.

dvc/command/gc.py
```diff
@@ -39,9 +48,9 @@ def run(self):
             return 1
 
         self.repo.gc(
-            all_branches=self.args.all_branches,
-            all_tags=self.args.all_tags,
-            all_commits=self.args.all_commits,
+            all_branches="branches" in self.args.remove,
+            all_tags="tags" in self.args.remove,
+            all_commits="commits" in self.args.remove,
```
efiop: Let's pass `remove` to the `repo.gc` and let it deal with it. We need to keep API and CLI symmetric here, no reason to change one and keep the other.

dvc/repo/gc.py
efiop: @skshetry I don't think you need to pass `has_scm`, brancher is able to deal with no-scm case by itself.

skshetry: That does not seem to be the case. The `brancher` uses `SCM::resolve_rev()` which is not available on `NoSCM`.
```sh
$ dvc init --no-scm
$ python -c '
from dvc.repo import Repo
repo = Repo(".")
print(list(repo.brancher(all_branches=True, all_commits=True, all_tags=True)))
'
  File "<string>", line 4, in <module>
  File "/home/saugat/repos/iterative/dvc/dvc/repo/brancher.py", line 45, in brancher
AttributeError: 'NoSCM' object has no attribute 'resolve_rev'
```

dvc/command/gc.py
skshetry: Agree about the symmetricity. But, I find it better to use args rather than a collection of *filters*, kinda like buttons which you can turn on or off to be more aggressive/effective. 

I don't have strong opinions though.

dvc/command/gc.py
skshetry: Okay, I'll update this when the PR is ready. Just making sure I haven't broken anything.

dvc/repo/gc.py
efiop: @skshetry Oh, that is a very good catch! But the issue is that `scm.resolve_rev` gets resolved even for empty revs, hence why this problem raises. It needs to be fixed there, there is no need for `has_scm`, which is more of a workaround than a solution.

skshetry: I was confused with the usecase of `brancher`. From it's docstring:
> Generator that iterates over specified revisions.

So, I thought, it should not even work for `NoSCM`.

dvc/command/gc.py
efiop: @skshetry well, but why not? :) Think about them as being used in future to remove specific references. E.g. `remove=mytag`, which you will have to pass as `remove=["mytag"]` to API anyways, so why bother with additional flags for `repo.gc`, if you can `repo.gc(remove=["commits", "tags", "branches"])` right away.

skshetry: Yes, it makes sense from *future*, as it allows a lot of flexibility. I was just saying, `all_commits=True, all_branches=True` are better interface than `remove={"commits", "branches"}`, unless there are too many arguments to pass (not really flexible).

I was looking from user interface-side. I'll fix.

dvc/repo/gc.py
efiop: @skshetry Got it. Yeah, we use NoSCM as a corner case, that doesn't have any branches, any commits, any tags, but does have the workspace 🙂 So most of the methods just don't return anything, which makes `brancher` work fine in that corner case of "no branches or revisions" 🙂

dvc/command/gc.py
Suor: `all_commits=True`, ... is unsafe, this is why we switch to requiring user explicitly state what he/she wants to remove. If we think of `Repo` methods as somewhat potentially public then we should mirror cli interface in `Repo.gc()` interface. It might be translated there to `all_commits` and friends.

Suor: I see three downsides to this, on top of @efiops point:
- it's very confusing that `all_branches` in `Repo.gc()` is somewhat but not entirely reverse to `all_branches` in `Repo.used_cache()/brancher()`
- three params instead of one combining them, I was looking into reducing the number of params in our functions and combining them is one of the ways
- we try to keep `Repo` methods API close to cli

skshetry: I assumed this was only related to CLI at first. I only realized very late that `all_commits=False` is not `--remove=commits`, but `--remove=commits,branches,tags`. Would it be good to make adjustments to the `Repo.brancher()` to exclude HEAD revisions from branches and  tags if `all_commits=True`  unless `all_tags` and/or `all_branches` is specified?

It'd be easier, I guess there, but, I'd like to handle it on `Repo.gc()` itself.

efiop: @skshetry I would probably consider adding `exclude` argument to the brancher, that you could then pass something like `exclude=(scm.list_branches() + scm.list_tags())`

efiop: @skshetry I would consider introducing an `exclude` flag for brancher and using it something like this: `brancher(all_commits=True, exclude=[scm.list_branches()])` to exclude branch heads. Other approach might be to collect used cache with `all_commit`, then with `all_branches` and substitute them 🙂

dvc/repo/gc.py
```diff
@@ -41,15 +42,27 @@ def gc(
...
+
             used.update(
                 repo.used_cache(
-                    all_branches=all_branches,
+                    all_branches="branches" not in remove,
                     with_deps=with_deps,
-                    all_tags=all_tags,
-                    all_commits=all_commits,
+                    all_tags="tags" not in remove,
+                    all_commits="commits" not in remove,
```
efiop: Need to not forget to remove `all_tags`, `all_branches` and `all_commits` from arguments for repo.gc

dvc/command/gc.py
Suor: @efiop I don't like changing brancher. The logic in both brancher and `.used_cache()` should stay the same. There is no reason to translate cli/api to internal code.

Suor: @efiop additionally all that new logic you propose is more complex than existing one and does not solve anything that is required by this PR/issue.

dvc/command/gc.py
```diff
@@ -9,19 +9,28 @@
...
+
+        self.args.remove = self.args.remove or []
+
+        if "commits" not in self.args.remove:
             msg += " and all git commits"
-        elif self.args.all_branches and self.args.all_tags:
+
+        if (
+            "branches" not in self.args.remove
+            and "tags" not in self.args.remove
```
Suor: Maybe make `self.args.remove` a set:
```python
if not {"branches", "tags"} & self.args.remove:
    # ...
```
```

File: dvc/repo/gc.py [hunk 10]
```diff
@@ -41,15 +40,27 @@ def gc(
 
         used = NamedCache()
         for repo in all_repos + [self]:
+            exclude_revs = (
+                repo.scm.list_branches()
```
Suor: Let's not spread scm logic all over codebase. `brancher` is not only used in gc, it is also used in push/pull and metrics. This PR meant to be very simple and local - we change how we accept things from user and translate it to internal calls, which are not changed.

dvc/command/gc.py
skshetry: Please don't nitpick at the moment. This PR does not work. It's a WIP for a reason.

dvc/command/gc.py
efiop: @Suor Let's be constructive here. What do you suggest?

Suor: @efiop wrote that below, but maybe to make it more clear:
- leave `.brancher()`, `.used_cache()` and any non-cli non Repo api-methods code unchanged.
- accept new options in `dvc gc`, raise some `DeprecatedError` on the old ones
- pass that to `Repo.gc()` as is
- translate `remove={...}` to `all_branches` and friends that `.brancher()` and `.used_cache()` understand in `Repo.gc()` implementation.

This way only `CmdGC` and `Repo.gc()` would be changed.

efiop:
For the record: @skshetry left for a holiday, so I'm taking over here 🙂

dvc/command/gc.py
efiop: > accept new options in dvc gc, raise some DeprecatedError on the old ones

No need, we are breaking the default behavior anyway, preserving old options would be hard and confusing. We need to drop them right away, without any DeprecationError's.

I think you are missing something here. There is no direct translation of current options into the new ones, so your proposition is impossible to implement without changing some of that. For example, `dvc gc --remove=branches` needs to collect all used cache for all references except for all branches, so it would translate to something like `git rev-list  --all --skip $ALL_BRANCHES`, which requires adjustments for `used_cache` and `brancher`.

dvc/scm/git/__init__.py
```diff
@@ -244,8 +244,18 @@ def list_branches(self):
     def list_tags(self):
         return [t.name for t in self.repo.tags]
 
-    def list_all_commits(self):
-        return [c.hexsha for c in self.repo.iter_commits("--all")]
```
efiop: This was a misuse of `iter_commits`, where passing `--all` to it worked accidentally because it was passing it as `rev` to `rev-list`. New implementation uses `rev-list` directly, which should be significantly faster (no benches yet).

efiop: Also, previous implementation was silently keeping `refs/remotes` and `refs/stash` as well, which might or might not be what we want, considering that we don't currently have any means of deleting cache for those.

dvc/command/gc.py
```diff
@@ -39,9 +40,9 @@ def run(self):
             return 1
 
         self.repo.gc(
-            all_branches=self.args.all_branches,
-            all_tags=self.args.all_tags,
-            all_commits=self.args.all_commits,
+            remove_all_tags=self.args.remove_all_tags,
```
efiop: Opted for explicit flags to avoid confusion such as previous `--remove=tags,commits`(where it is not clear what `commits` are and why you are not able to pass specific tag or branch to it) and also to simplify completion.

tests/func/test_gc.py
```diff
@@ -92,16 +92,26 @@ def test(self):
 
         self._check_cache(4)
 
-        self.dvc.gc(all_tags=True, all_branches=True)
+        self.dvc.gc()
```
efiop: Will look into converting to pytest and possibly adding new tests soon.

dvc/command/gc.py
```diff
@@ -12,17 +12,18 @@
...
+            "This will remove all cache except items used in the working tree"
+            "{tags}{history}{branches}{history}"
+        ).format(
+            tags="" if self.args.remove_all_tags else "and all git tags",
+            branches=""
+            if self.args.remove_all_branches
+            else "and all git branches",
+            history=""
+            if self.args.remove_all_history
+            else "and their history",
```
Suor: This makes the phrase weird - many ands. Maybe just list things instead:
```
This will remove all cache not referenced from:
- the working tree
- all branch heads
...
```
? Will make it more glanceable, the changing items will be always in the same place, any user already familiar with gc, will be simply skipping the entering phrase without loose of any meaning.

dvc/command/gc.py
```diff
@@ -65,34 +66,32 @@ def add_parser(subparsers, parent_parser):
...
         default=False,
-        help="Collect garbage in remote repository.",
+        help="Remove cache for all git branches.",
     )
     gc_parser.add_argument(
-        "-r", "--remote", help="Remote storage to collect garbage in."
+        "--remove-all-history",
+        action="store_true",
+        default=False,
+        help="Remove cache for all history of all branches and tags.",
```
Suor: As far as I recall, your intention was to remove history by default. Why did you change your mind?

dvc/repo/brancher.py
```diff
@@ -33,7 +33,10 @@ def brancher(  # noqa: E302
     yield "working tree"
 
     if all_commits:
-        revs = scm.list_all_commits()
+        revs = scm.list_all_commits(
+            exclude_all_tags=not all_tags,
+            exclude_all_branches=not all_branches,
+        )
```
Suor: This makes `.brancher` and `.used_cache()` interfaces lying ones, `all_commits` no longer means all commits.

Suor: This also makes `dvc gc --remove-all-branches --remove-all-tags` approaching noop for many scenarios, but not quite noop, a confusing situation. Suppose you make things iteratively and last commits only about some polish like README adjustments.

dvc/command/gc.py
efiop: @Suor To keep it as safe as possible for now :slightly_smiling_face: And, well, I implemented a proper approach by excluding refs when not deleting full history.

dvc/repo/brancher.py
efiop: Yeah, the naming `brancher` is at fault here. It should really have been something like `iter_commits()` instead.

But `--remove-all-branches --remove-all-tags` is not a noop, it will exclude tags and branches from the refs for which it iterates through.  Hence why I've mentioned an issue with `refs/remotes` and `refs/stash` etc. We should either exclude those by default or add an option to get rid of those. For now this is a "safe" approach, that can be revisited iteratively.

Suor: So, it's even more confusing. You need a whole paragraph of text to explain how it works, but still in the many scenarios it's just noop.

dvc/scm/git/__init__.py
Suor: So it was previously simply "keep anything, which is referenced from git". Now it's more complicated )

dvc/repo/brancher.py
efiop: @Suor Well, that is because of `refs/remotes` and `refs/stash`. Will need to handle them somehow. To me seems like a good idea to just drop them by default and only keep refs/heads and refs/tags. What do you think?

Suor:
Looking at this I see it becoming more and more weird. I hate to crush the party, but we should consider closing it with an original ticket.

So far I see only one advantage to whatever we have now - it's safer for no upstream or `dvc gc -r/-c` case. I see numerous disadvantages though:
- it's hard to explain/understand what does it really do
- it's hard to use it - long options, that you need to use in most cases, options doesn't reflect whatever people intend to do in many cases
- asymmetry with `push/pull`

What we should care instead probably:
- not loosing data sources, everything else might be recalculated
- not loosing things not pushed to upstream
- maybe separating local/cloud gc better

shcheklein:
Some background, the way I remember/see this. The whole things (interestingly enough) originates from the two discussions:

1. Being able to show metrics across different commits https://github.com/iterative/dvc/issues/1691:

> Can you consider supporting showing different metric values across different commits in the same branch?

(The discussion went into questions around symmetry, and I still believe symmetry is important if we can preserve it)

2. And the second one (and few other users reported this) - https://github.com/iterative/dvc/issues/1691#issuecomment-513412918 

To be precise, they were concerned about the default behavior. But as I discussed with them providing an option to keep some N last commits should be enough. Though some folks were complaining about the default behavior besides ODS guys.

--------------------------------

I agree that changing the default behavior creates a mess, but it also looks like it not exactly necessary to solve the problems pointed by the users. For example, we can do something like:

- do not allow to run `dvc gc` at all, only with `-f` or any flag that explicitly expands the scope of files to keep (tags, commits, branches)
- along with `--all-commits` introduce `-n` to keep some reasonable number of last commits. (and later add symmetrical option to all other commands which accept `-T`/`-a` - metrics, push, pull) - to finally get out of this rabbit hole and solve one of the actual problems.
- minor but nice: add a shortcut for the `-all-commits` - `-C` or something?

This way it can be safe enough, no need to change names/behavior, we still reuse the `all-commits` logic. It's symmetrical with push/pull, etc.

Suor:
@shcheklein my impression now is that people express their concerns with e.g. non-safe defaults but they don't really care about it. I mean they don't do any followups or even +1s on them, which means this is not bugging them in the long run. We are the only ones still pushing it.

I agree that those changes might be beneficial just not that necessary, thus I was talking about saving the effort.

efiop:
Ok, so we are lacking a complete solution here, which was expected, as this change was only meant to make **the first iteration** on our path to good gc. Something safe that we could jump to and then chip away from it with new options. But now the discussion is again about the broader topic, which will again result in it just hanging out there for a few more months more.

Current gc suffers from the same issues solved in this new version, but has old cryptic options that won't allow us to do something like `--all-commits --all-tags` with the intention of only getting rid of branches. And in the new one, you could do that with `--remove-all-branches`, that will delete branches and their dangling history automatically(e.g. say you have `mybranch` that has `mytag` midway, then all commits from `mytag` up to `mybranch` HEAD will get deleted by that flag).

I know that with that example I'm making something out to apply to current option list, but the reason I do that is that in the next step you could easily apply that to potential `--remove mybranch` or `--remove my.dvc`, unlike the old one, where syntax is focused on what to keep and not what to remove. From discussions on ODS and Discord, people always ask how to **remove** something specific, and the new `gc` is allowing for that expansion of functionality in the future, unlike old `gc`. Plus you could always add `--force` flag there, if someone absolutely needs to get rid of cache for one specific dvc file and he really doesn't care if it is used somewhere else in the history - again, unlike with old `gc`.

That being said, it is clear that we also need to improve user awareness, which should be solved by some `stats` command/flag/behaviour, that would tell you how many data sources you have, how many branches, what is their unique space usage, how much you could free up if you delete cache for them, whether or not cache is already on the cloud and so on and so forth. That needs a separate discussion, but even now it seems pretty clear that whatever we come up with will be mostly focused on **remove** and not on **keep** aspect, as you will need to tell `gc` what you no longer need.

So I think we should bite the bullet, switch to the safest behaviour and start iterating on this with a list of options that we pretty much already agree on like preserving data sources, freeing up local cache if there is stuff on the cloud already, separating cloud and local gc better, introducing history depth option etc. If there will be errors, so be it, we could always improve or switch behaviour, at least we will get more experience and will know better what we want and what we need, which will fuel next iterations. Otherwise, we will be "saving the effort" indefinitely and won't move anywhere, not gaining any new feedback/info and so being stuck in the discussion forever flying in our own fantasies.

The history of the current gc ticket is a clear illustration that we are stuck, we need to stop talking the talk and finally walk the walk, to iterate and collect new feedback.

Suor:
@efiop I don't agree that most people want to remove something specific. People usually want to simply free up some space, which `dvc gc` now does. If we switch to `--remove-*` flags then people will ask how they can remove everything but X, i.e. keep X, so that switch doesn't really solve anything but safety, which people don't really care about.

Bottom lines:
- we are putting effort into this, while nobody cares that much.
- current iteration is one step forward, two step backwards.

I understand that there is [an invested effort/code/time](https://en.wikipedia.org/wiki/Escalation_of_commitment) here, I was pushing this for a long time too, but now I think it's not worth it.

efiop:
@Suor I have never ever heard any user ask how to only keep something and delete the rest in all of these years, have you? Maybe @shcheklein or @dmpetrov have a different experience.

shcheklein:
@efiop I think there is a natural bias - `gc` options are made right now with a "keep" philosophy in mind, so users asks "how to remove" something. If we change it - we will start getting questions "how to keep" instead.

The best way to answer here is to try to understand what actually users want in their workflow. Probably again, not one can come up with a 100% correct answer, unfortunately. I think both approaches would be valid in different cases.

My 2cs here - keep semantics is more user friendly if you can put it that way. I care about certain things from my domain (experiments, models, datasets and whatnot) to stay on my machine because I still iterate with them. And I don't care about anything else.

Remove feels more advanced - I need to understand now what exactly do I want to remove (either do a negation from what I want to keep) or recall what takes most of the space and go remove it.

I definitely recognize the lack of a "pinpoint" remove operation - few users have been asking - how to un-add something, for example. But it feels it should be solved separately?

efiop:
@shcheklein How I imagine this. I'm a DS and I have created a new branch `mybranch`, that I've added some new things to, produced some intermediate files. Then I realise that I no longer need it and I would like to get rid of that data. How do I do that with the current "keep" terminology? Do I list all the other stuff that I have that I might need? But that would not be trivial, as no one keeps a full list of tags/branches in their mind. With the one that I suggest user would be able to `git branch -D branch` and `dvc gc`(no options), that will automatically get rid of the unused garbage, while not affecting the rest. Or it might potentially be `dvc gc --remove mybranch` in the future, that would eliminate the need for `git branch -D`. Again, I see no good way to implement that in the current `keep` approach.

People seem to rely on remotes as a backup and just push stuff as they go and nuke the whole local cache whenever it is taking too much space, as they don't understand current `dvc gc`. Yes, remotes are usually limitless compared to the local cache capasity, but the time will come that you will need to clean it up as well, at which point you will have to dive into `dvc gc` options and new `dvc gc` would be a good no-brainer no-bad-consequences command to run.

shcheklein:
> Again, I see no good way to implement that in the current keep approach.

how about `dvc gc --all-commits` in this case? it will keep all the stuff you have, except the removed branch? Granted, this operation gonna be slower.

> as they don't understand current `dvc gc`

so, we change it, right? we provide all-commits, last N commits, branches, tags, etc. Next step would be to provide filters on tags and branches - `--filter *-experiments` - and again this is symmetrical and needed for all commands. The same as you need -n for metrics, pull, push, etc. 

would love to better explorer the "they don't understand" part, for now I tend to agree that there are not that much evidence for this.

saying all that, I understand that there is not one-fits-all solution in this case - some might prefer keep, some remove semantics. In some case it would easier to write in one, in other cases in another. But, I'm not sure we have strong enough evidence that "remove" is superior to the extent to jump into breaking the symmetry, default behavior and redoing all options.

May be we should try to come with more examples to start feeling this better? (but again, it can be an argument to slightly change it for now, not changing the whole semantics of it?)

efiop:
> how about dvc gc --all-commits in this case? it will keep all the stuff you have, except the removed branch? Granted, this operation gonna be slower.

If you delete the branch prior to that - yes, but it is not a "no-brainer" command. And it doesn't extend to the aforementioned removal of particular dvc-file.

> so, we change it, right? we provide all-commits, last N commits, branches, tags, etc. Next step would be to provide filters on tags and branches - --filter *-experiments - and again this is symmetrical and needed for all commands. The same as you need -n for metrics, pull, push, etc.

No, it doesn't have to be symmetrical with metrics/pull/push, really, as they have different purposes. `dvc gc` that assumes `--all-commits` is good, while `dvc push/pull` that assumes `--all-commits` is bad.

> would love to better explorer the "they don't understand" part, for now I tend to agree that there are not that much evidence for this.

Wasn't it the previously constantly discussed argument about current `dvc gc`? That backwards logic of `keep` was not intuitive?

> saying all that, I understand that there is not one-fits-all solution in this case - some might prefer keep, some remove semantics. In some case it would easier to write in one, in other cases in another. But, I'm not sure we have strong enough evidence that "remove" is superior to the extent to jump into breaking the symmetry, default behavior and redoing all options.

Well, it has been a couple of years already, and the feedback we've received was about gc being dangerous and people asking how to delete something in particular, at least in my experience. I think we need to iterate based on what we have and not freeze this ticket again for another couple of years.

> May be we should try to come with more examples to start feeling this better? (but again, it can be an argument to slightly change it for now, not changing the whole semantics of it?)

Well, we could forbid `dvc gc` with no arguments as you've suggested, but then we will need to introduce a compatible option for the current behaviour, which will be `dvc gc --workspace` in current `keep` terminology or something like that. Is that really better to introduce another bump in the road for the user? I think `dvc gc` should be usable without arguments and provide a no-brainer low-consequence behaviour.

dmpetrov:
Interesting discussion... and it is more than one year long 😄 A lot of good points for each of the options.

> we are lacking a complete solution here, which was expected, as this change was only meant to make **the first iteration**

My major concern with GC - I have a strong feeling that we are still lacking a big picture and don't fully understand the scenarios. The only thing that we can say we confidence - the current GC does not solve users' problems.

Without the big pictures, it is quite dangerous to move forward with the current approach because it is quite opinionated and is based on the assumption that we can implement all possible scenarios (with probably `-n X` option which still looks like a workaround, not a solution).

> From discussions on ODS and Discord, people always ask how to **remove** something specific

It might be more efficient and much safer to implement a very basic, low-level GC-option for removing specific cache items. I can imagine a command `dvc gc data.csv` (with an optional `--rev abcd123`) which removes data for a specific cache item with surgical precision.

With this specific GC-commands, users will be well-equipped for implementing their own GC-scenarios. So, we can identify common GC-scenarios and implement them later as high-level options like `--remove-all-<something>`.

I'd suggest reverting the current change and implementing the low-level GC-option.

dmpetrov:
Also, I quite don’t understand why we care about a safe approach so much in the context of GC. The core propose of GC is remove data, it is not safe. I think it was a mistake (my mistake btw…. 😄)  to call this command GC. `dvc clean` or `dvc cache clean` might be a better name for it. We can consider renaming - any ideas are welcome!

Today `dvc gc` is safer than it looks like. It asks the question all the time in prompt mode: “Are you sure you want to proceed? [y/n]”.

PS:
We might need more command in the future to manage the cache: 
- how many versions of this dataset do we keep in local cache? ... in data-remote?
- how much space this file takes among the whole history?
- how much data repo imports from outside?
So, introducing a `dvc cache` umbrella with `dvc cache clean/gc` option can be a good move.

Suor:
> Granted, this operation gonna be slower.

Not really it would be the same speed. You can't really delete whatever is referenced from some branch because anything referenced from there might also be referenced from somewhere else. So the implementation is always the same:
- collect all the checksum, which should be kept
- walk through cache, removing anything not in the set above
This is why a proper unadd is such a complicated operation.

> That backwards logic of keep was not intuitive?

The issue always was that it is not safe, not that it's non-intuitive. At least this is how I understood that all the way.


> The only thing that we can say with confidence - the current GC does not solve users' problems.

I doubt this. It looks like most users are ok with it, some people worry occasionally about non-safe defaults, but they don't follow up on their worries, which makes me think the issue is theoritical and we are the only ones really pushing it.

dmpetrov:
It looks like different people have a different understanding and expectations of GC command. Some people (@efiop and I'm) trying to clean the cache and cloud/remote to solve user issues with space. While others (@Suor and @shcheklein) are trying to preserve more "keep" semantic which makes sense for GC name.

To my mind, the actual problem we need to focus on is cleaning space (local and cloud) and we should give up with the GC name. Let's introduce a new subcommand to `dvc cache` (today it has only `dvc cache dir`).

With this command, we can:
1. keep the current GC (or the safe version) - `dvc cache gc` 
2. introduce pinpoint remove - `dvc cache remove users.csv file.pkl`
3. introduce "remove" strategies - `dvc cache remove-all-branches` etc
4. introduce "keep" strategies - `dvc cache clean -n 10`
5. *in the future we might need additional cache commands like `history`/`log`/`du` and others.

For cleaning up space in remote an option `dvc cache -c` can be used or even a separate command `dvc cloud`.

efiop:
@dmpetrov Thanks for the feedback! I really like your `dvc cache gc/clean/remove`(depending on which name(s) we decide to go with) idea. Let me give it a thought and come up with a proposal based on it. So far it is clear that current `dvc gc` is here to stay for now and might be obsoleted after `dvc cache ...` cmds introduction, so clearly no need to adjust it right now and this PR will be closed soon. I'll send a bugfix for `--all-commits` separately so it is not lost and could be quickly merged.

Suor:
From my 1-1 with @efiop:

The remove specific branch or file commands won't care if same checksum is referenced anywhere else in the history, they will assume a user knows what he is doing.

efiop:
Closing this fo

PR #3208, opened 2020-01-21
tests: replace spy() with pytest-mock spy

### LINKED ISSUE #3198: Remove spy method
In some places, we use `tests/utils/spy` method to mock spied method so that we can observe how particular actions interact with the target method/function. Current implementation:
```
def spy(method_to_decorate):
    mock = MagicMock()

    def wrapper(self, *args, **kwargs):
        mock(*args, **kwargs)
        return method_to_decorate(self, *args, **kwargs)

    wrapper.mock = mock
    return wrapper
```

assumes that we are spying on some class method (by having `self` in wrapper args). This is not always true, and it leads to hiding first argument if we will use it to spy on function (eg `file_md5`) which results in trimmed call args for functions. 

Example:
```
def test_spy(tmp_dir):
    tmp_dir.gen("file", "file content")

    from tests.utils import spy
    import dvc
    checksum_spy = spy(dvc.utils.file_md5)
    checksum_spy("file")
    assert checksum_spy.mock.call_args_list[0].args == ("file",)
```
should pass, while it does not because args tuple is empty.

[EDIT]
As @mroutis  noted, `pytest`'s mocker can `spy`, so it would be good to remove current spy usages and replace them with `mocker.spy`.

fabiosantoscode:
After I'd started, @pared pointed out that it would be best to convert the tests to plain functions as opposed to classes, to better utilise the `mocker` fixture. I ended up having to create a bunch of wrappers just to use this fixture. Let me know if you'd like me to go ahead and convert all of the classes that were using spy into plain pytest function tests.

Fixes #3198

-------

### DIFF
--- a/tests/func/test_add.py
+++ b/tests/func/test_add.py
@@ -28,7 +28,6 @@ from dvc.utils.fs import path_isin
 from dvc.utils.stage import load_stage_file
 from tests.basic_env import TestDvc
 from tests.utils import get_gitignore_content
-from tests.utils import spy
 
 
 def test_add(tmp_dir, dvc):
@@ -245,9 +244,15 @@ class TestDoubleAddUnchanged(TestDvc):
         self.assertEqual(ret, 0)
 
 
-class TestShouldUpdateStateEntryForFileAfterAdd(TestDvc):
+class TestDvcWithMocker(TestDvc):
+    @pytest.fixture(autouse=True)
+    def use_mocker(self, mocker):
+        self.mocker = mocker
+
+
+class TestShouldUpdateStateEntryForFileAfterAdd(TestDvcWithMocker):
     def test(self):
-        file_md5_counter = spy(dvc.remote.local.file_md5)
+        file_md5_counter = self.mocker.spy(dvc.remote.local, "file_md5")
         with patch.object(dvc.remote.local, "file_md5", file_md5_counter):
             ret = main(["config", "cache.type", "copy"])
             self.assertEqual(ret, 0)
@@ -274,9 +279,9 @@ class TestShouldUpdateStateEntryForFileAfterAdd(TestDvc):
             self.assertEqual(file_md5_counter.mock.call_count, 1)
 
 
-class TestShouldUpdateStateEntryForDirectoryAfterAdd(TestDvc):
+class TestShouldUpdateStateEntryForDirectoryAfterAdd(TestDvcWithMocker):
     def test(self):
-        file_md5_counter = spy(dvc.remote.local.file_md5)
+        file_md5_counter = self.mocker.spy(dvc.remote.local, "file_md5")
         with patch.object(dvc.remote.local, "file_md5", file_md5_counter):
 
             ret = main(["config", "cache.type", "copy"])
@@ -320,11 +325,13 @@ class TestAddCommit(TestDvc):
         self.assertEqual(len(os.listdir(self.dvc.cache.local.cache_dir)), 1)
 
 
-class TestShouldCollectDirCacheOnlyOnce(TestDvc):
+class TestShouldCollectDirCacheOnlyOnce(TestDvcWithMocker):
     def test(self):
         from dvc.remote.local import RemoteLOCAL
 
-        get_dir_checksum_counter = spy(RemoteLOCAL.get_dir_checksum)
+        get_dir_checksum_counter = self.mocker.spy(
+            RemoteLOCAL, "get_dir_checksum"
+        )
         with patch.object(
             RemoteLOCAL, "get_dir_checksum", get_dir_checksum_counter
         ):
@@ -477,15 +484,15 @@ def test_should_cleanup_after_failed_add(tmp_dir, scm, dvc, repo_template):
     assert "/bar" not in gitignore_content
 
 
-class TestShouldNotTrackGitInternalFiles(TestDvc):
+class TestShouldNotTrackGitInternalFiles(TestDvcWithMocker):
     def test(self):
-        stage_creator_spy = spy(dvc.repo.add._create_stages)
+        stage_creator_spy = self.mocker.spy(dvc.repo.add, "_create_stages")
 
         with patch.object(dvc.repo.add, "_create_stages", stage_creator_spy):
             ret = main(["add", "-R", self.dvc.root_dir])
             self.assertEqual(0, ret)
 
-        created_stages_filenames = stage_creator_spy.mock.call_args[0][0]
+        created_stages_filenames = stage_creator_spy.mock.call_args[0][1]
         for fname in created_stages_filenames:
             self.assertNotIn(".git", fname)
 
@@ -572,7 +579,7 @@ def test_readding_dir_should_not_unprotect_all(tmp_dir, dvc, mocker):
     dvc.add("dir")
     tmp_dir.gen("dir/new_file", "new_file_content")
 
-    unprotect_spy = spy(RemoteLOCAL.unprotect)
+    unprotect_spy = mocker.spy(RemoteLOCAL, "unprotect")
     mocker.patch.object(RemoteLOCAL, "unprotect", unprotect_spy)
     dvc.add("dir")
 
@@ -587,7 +594,7 @@ def test_should_not_checkout_when_adding_cached_copy(tmp_dir, dvc, mocker):
 
     shutil.copy("bar", "foo")
 
-    copy_spy = spy(dvc.cache.local.copy)
+    copy_spy = mocker.spy(dvc.cache.local, "copy")
     mocker.patch.object(dvc.cache.local, "copy", copy_spy)
 
     dvc.add("foo")
--- a/tests/func/test_add.py
+++ b/tests/func/test_add.py
@@ -29,6 +29,8 @@ from dvc.utils.stage import load_stage_file
 from tests.basic_env import TestDvc
 from tests.utils import get_gitignore_content
 
+dvc_module = dvc  # to disambiguate between the module and the fixture
+
 
 def test_add(tmp_dir, dvc):
     stage, = tmp_dir.dvc_gen({"foo": "foo"})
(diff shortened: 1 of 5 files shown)


tests/func/test_add.py
```diff
@@ -245,9 +244,15 @@ def test_dir(self):
...
+    @pytest.fixture(autouse=True)
+    def use_mocker(self, mocker):
+        self.mocker = mocker
+
+
+class TestShouldUpdateStateEntryForFileAfterAdd(TestDvcWithMocker):
     def test(self):
-        file_md5_counter = spy(dvc.remote.local.file_md5)
+        file_md5_counter = self.mocker.spy(dvc.remote.local, "file_md5")
         with patch.object(dvc.remote.local, "file_md5", file_md5_counter):
```
pared: With mocker spy, there is no need to patch, it is automatically done by mocker, so with statement can be removed

tests/func/test_add.py
```diff
@@ -274,9 +279,9 @@ def test(self):
             self.assertEqual(file_md5_counter.mock.call_count, 1)
 
 
-class TestShouldUpdateStateEntryForDirectoryAfterAdd(TestDvc):
+class TestShouldUpdateStateEntryForDirectoryAfterAdd(TestDvcWithMocker):
     def test(self):
-        file_md5_counter = spy(dvc.remote.local.file_md5)
+        file_md5_counter = self.mocker.spy(dvc.remote.local, "file_md5")
         with patch.object(dvc.remote.local, "file_md5", file_md5_counter):
```
pared: As above

tests/func/test_add.py
```diff
@@ -320,11 +325,13 @@ def test(self):
...
-class TestShouldCollectDirCacheOnlyOnce(TestDvc):
+class TestShouldCollectDirCacheOnlyOnce(TestDvcWithMocker):
     def test(self):
         from dvc.remote.local import RemoteLOCAL
 
-        get_dir_checksum_counter = spy(RemoteLOCAL.get_dir_checksum)
+        get_dir_checksum_counter = self.mocker.spy(
+            RemoteLOCAL, "get_dir_checksum"
+        )
         with patch.object(
```
pared: As above

tests/func/test_add.py
```diff
@@ -477,15 +484,15 @@ def test_should_cleanup_after_failed_add(tmp_dir, scm, dvc, repo_template):
     assert "/bar" not in gitignore_content
 
 
-class TestShouldNotTrackGitInternalFiles(TestDvc):
+class TestShouldNotTrackGitInternalFiles(TestDvcWithMocker):
     def test(self):
-        stage_creator_spy = spy(dvc.repo.add._create_stages)
+        stage_creator_spy = self.mocker.spy(dvc.repo.add, "_create_stages")
 
         with patch.object(dvc.repo.add, "_create_stages", stage_creator_spy):
```
pared: As above

tests/func/test_add.py
```diff
@@ -572,7 +579,7 @@ def test_readding_dir_should_not_unprotect_all(tmp_dir, dvc, mocker):
     dvc.add("dir")
     tmp_dir.gen("dir/new_file", "new_file_content")
 
-    unprotect_spy = spy(RemoteLOCAL.unprotect)
+    unprotect_spy = mocker.spy(RemoteLOCAL, "unprotect")
     mocker.patch.object(RemoteLOCAL, "unprotect", unprotect_spy)
```
pared: As above

tests/func/test_add.py
```diff
@@ -587,7 +594,7 @@ def test_should_not_checkout_when_adding_cached_copy(tmp_dir, dvc, mocker):
 
     shutil.copy("bar", "foo")
 
-    copy_spy = spy(dvc.cache.local.copy)
+    copy_spy = mocker.spy(dvc.cache.local, "copy")
     mocker.patch.object(dvc.cache.local, "copy", copy_spy)
```
pared: as above

tests/func/test_data_cloud.py
```diff
@@ -586,8 +585,14 @@ def test(self):
...
+    @pytest.fixture(autouse=True)
+    def use_mocker(self, mocker):
+        self.mocker = mocker
+
     def test(self):
-        test_get_file_checksum = spy(RemoteLOCAL.get_file_checksum)
+        test_get_file_checksum = self.mocker.spy(
+            RemoteLOCAL, "get_file_checksum"
+        )
         with patch.object(
```
pared: as above

tests/func/test_import_url.py
```diff
@@ -50,7 +53,9 @@ def setUp(self):
             fobj.write("content")
 
     def test(self):
-        remove_outs_call_counter = spy(dvc.stage.Stage.remove_outs)
+        remove_outs_call_counter = self.mocker.spy(
+            dvc.stage.Stage, "remove_outs"
+        )
         with patch.object(
```
pared: as above

tests/unit/utils/test_fs.py
```diff
@@ -72,7 +75,9 @@ def test_should_return_false_on_no_more_dirs_below_path(
 
     @patch.object(System, "is_symlink", return_value=False)
     def test_should_call_recursive_on_no_condition_matched(self, _):
-        contains_symlink_spy = spy(contains_symlink_up_to)
+        contains_symlink_spy = self.mocker.spy(
+            dvc.utils.fs, "contains_symlink_up_to"
+        )
         with patch.object(
```
pared: as above

tests/func/test_add.py
fabiosantoscode: Great to know, thanks!

fabiosantoscode:
Removed the useless patch.object(), thanks @pared

pared:
@fabiosantoscode, moving to mocker's spy seems good. But to be honest I don't like that we are mixning `pytest` and `UnitTest`, I think that those tests (only those that use `spy`) should be converted, it is something that we will have to do anyway at some point. How does it sound?

fabiosantoscode:
Sounds good @pared let me get to that.

fabiosantoscode:
@pared there :snowman:

tests/func/test_add.py
```diff
@@ -28,7 +28,8 @@
 from dvc.utils.stage import load_stage_file
 from tests.basic_env import TestDvc
 from tests.utils import get_gitignore_content
-from tests.utils import spy
+
+dvc_module = dvc  # to disambiguate between the module and the fixture
```
pared: maybe `import dvc as dvc_module`?

tests/func/test_add.py
```diff
@@ -245,66 +246,72 @@ def test_dir(self):
         self.assertEqual(ret, 0)
 
 
-class TestShouldUpdateStateEntryForFileAfterAdd(TestDvc):
-    def test(self):
-        file_md5_counter = spy(dvc.remote.local.file_md5)
-        with patch.object(dvc.remote.local, "file_md5", file_md5_counter):
-            ret = main(["config", "cache.type", "copy"])
-            self.assertEqual(ret, 0)
+class TestDvcWithMocker(TestDvc):
```
pared: Seems like it can be removed now :)

tests/func/test_add.py
```diff
@@ -245,66 +246,72 @@ def test_dir(self):
...
+    ret = main(["config", "cache.type", "copy"])
+    assert ret == 0
 
-            os.rename(self.FOO, self.FOO + ".back")
-            ret = main(["checkout"])
-            self.assertEqual(ret, 0)
-            self.assertEqual(file_md5_counter.mock.call_count, 1)
+    ret = main(["add", "foo"])
+    assert ret == 0
+    assert file_md5_counter.mock.call_count == 1
```
pared: can be simplified to `file_md5_counter.call_count`, this applies to each use case of `spy`

tests/func/test_data_cloud.py
```diff
@@ -585,24 +583,21 @@ def test(self):
...
-            ret = main(["push"])
-            self.assertEqual(ret, 0)
-            ret = main(["run", "-d", self.FOO, "echo foo"])
-            self.assertEqual(ret, 0)
-        self.assertEqual(test_get_file_checksum.mock.call_count, 1)
+def test_checksum_recalculation(mocker, dvc, tmp_dir):
+    tmp_dir.gen({"foo": "foo"})
+    test_get_file_checksum = mocker.spy(RemoteLOCAL, "get_file_checksum")
+    url = Local.get_url()
+    ret = main(["remote", "add", "-d", TEST_REMOTE, url])
```
pared: No need to use `TEST_REMOTE`, you can use some string name, like "upstream"

tests/func/test_import_url.py
```diff
@@ -42,6 +43,10 @@ def test(self):
 
 
 class TestShouldRemoveOutsBeforeImport(TestDvc):
```
pared: This can also be refactored.
prepare external resource using tmp_path_factory fixture (eg `tmp_path_factory.mktemp("external_source") / "file"`

tests/func/test_import_url.py
```diff
@@ -50,16 +55,26 @@ def setUp(self):
...
+            dvc.stage.Stage, "remove_outs"
+        )
+        ret = main(["import-url", self.external_source])
+        self.assertEqual(0, ret)
 
         self.assertEqual(1, remove_outs_call_counter.mock.call_count)
 
 
+def test_should_remove_outs_before_import(mocker, dvc, tmp_path_factory):
+    import_dir = TmpDir(tmp_path_factory.mktemp("import"))
```
pared: We have `erepo_dir` fixture for such things:
`with erepo_dir.chdir():
    erepo_dir.dvc_gen("foo": "foo content")`

tests/unit/utils/test_fs.py
```diff
@@ -72,18 +75,17 @@ def test_should_return_false_on_no_more_dirs_below_path(
 
     @patch.object(System, "is_symlink", return_value=False)
     def test_should_call_recursive_on_no_condition_matched(self, _):
```
pared: This test can also be refactored. Just extract it to standalone test, not part of `TestContainsLink` class.

tests/func/test_import_url.py
pared: Also, this test does not pass for py3.5, need to wrap `tmp_path_factory.mktemp` in `fspath`.

tests/func/test_data_cloud.py
efiop: @pared I think it is out of scope for this change 🙂

tests/func/test_add.py
efiop: @pared It is a direct conversion that is explicit, we could leave it as is with `== 1`.

tests/func/test_import_url.py
fabiosantoscode: I copied this before refactoring, then forgot to remove the original :)

tests/func/test_import_url.py
fabiosantoscode: Makes sense.

tests/unit/utils/test_fs.py
fabiosantoscode: Done!

tests/func/test_add.py
fabiosantoscode: Should I change it to not have `.mock` then? It does seem to do the same thing.

fabiosantoscode:
I've fixed most comments @pared and @efiop except this one: https://github.com/iterative/dvc/pull/3208#discussion_r369589870 which I'm not clear on what to do.

tests/func/test_add.py
pared: It does, I don't think its crucial, let's leave it as is.

PR #3211, opened 2020-01-22
brancher: do not get tree if there's no revs

skshetry:
From https://github.com/iterative/dvc/pull/3207#discussion_r369496309

When if it's a dvc only repo, brancher was failing as `NoSCM` does not have `resolve_rev`.
```sh
$ dvc init --no-scm
$ python -c '
from dvc.repo import Repo
repo = Repo(".")
print(list(repo.brancher(all_branches=True, all_commits=True, all_tags=True)))
'
  File "<string>", line 4, in <module>
  File "/home/saugat/repos/iterative/dvc/dvc/repo/brancher.py", line 45, in brancher
AttributeError: 'NoSCM' object has no attribute 'resolve_rev'
```

### DIFF
--- a/dvc/repo/brancher.py
+++ b/dvc/repo/brancher.py
@@ -42,8 +42,9 @@ def brancher(  # noqa: E302
             revs.extend(scm.list_tags())
 
     try:
-        for sha, names in group_by(scm.resolve_rev, revs).items():
-            self.tree = scm.get_tree(sha)
-            yield ", ".join(names)
+        if revs:
+            for sha, names in group_by(scm.resolve_rev, revs).items():
+                self.tree = scm.get_tree(sha)
+                yield ", ".join(names)
     finally:
         self.tree = saved_tree


dvc/repo/brancher.py
```diff
@@ -42,8 +42,9 @@ def brancher(  # noqa: E302
             revs.extend(scm.list_tags())
 
     try:
-        for sha, names in group_by(scm.resolve_rev, revs).items():
-            self.tree = scm.get_tree(sha)
-            yield ", ".join(names)
+        if revs:
```
mroutis: this should be caught by:

```python
    if not any([revs, all_branches, all_tags, all_commits]):
        yield ""
        return
```

:thinking:

mroutis: ohhh, but I saw that you were doing `all_*=True` :slightly_smiling_face:

mroutis: Wait, @skshetry , why don't you solved it by doing `if not scm: return` ?

mroutis: As far as I understand, `brancher` doesn't have anything to do when `NoSCM` is used.

So, it would be:

```python
    scm = self.scm

    if not any([revs, all_branches, all_tags, all_commits]) or not scm:
        yield ""
        return
```

I guess that, since the end result is the same, it doesn't matter

skshetry: `scm` is always there. If it's a non-git repo, it will be `NoSCM`.
Honestly, I didn't really feel confident around `brancher()` (this is an uncovered corner). I really just added here because of the `finally` statement here that resets the `self.tree`.

mroutis: totally, @skshetry :+1:

mroutis: However, using `brancher` with `revs` on a `NoSCM` will raise the same error.

It depends on if we want to ignore the error and `yield ""` silently or actually raise an error that brancher doesn't work with `NoSCM`.

```python
    if (
        not any([revs, all_branches, all_tags, all_commits])
        or type(self.scm) is not Git
    ):
        yield ""
        return
```

efiop: @mroutis Brancher _does_ work with NoSCM, it only shows the working tree.

skshetry: yes, if someone pass the `revs`, they must be doing something *clever* and I'll assume they no their stuff. I'd like the consumers of that particular API take responsibility rather than doing a noop.

mroutis: @efiop:

```console
$ dvc init --no-scm
$ python -c '
from dvc.repo import Repo
repo = Repo(".")
print(list(repo.brancher(revs=["HEAD", "HEAD~1"])))
'
  File "<string>", line 4, in <module>
  File "/home/mroutis/src/dvc/dvc/repo/brancher.py", line 46, in brancher
AttributeError: 'NoSCM' object has no attribute 'resolve_rev'
```

efiop: @mroutis There are no revs in NoSCM though, so your script is intentionally breaking it in a way that is not used in dvc.

efiop: Which doesn't make your point wrong, I'm just saying.

mroutis: @efiop , I don't get why `revs=["something"]` is _intentionally breaking it_ but no `all_commits=True`.

efiop: @mroutis because it expects to receive revs that are valid in the scm that we are operating in and there are no valid revs in NoSCM. And `all_commits=True` asks scm for list of all commits, which are `[]` for NoSCM. I agree that brancher could be improved to handle those cases more smoothly, but this fix does get the job done in this case.

skshetry: As far as my understanding goes, the `NoSCM` case and dirty tree should not even be the concern of the `brancher()`.

skshetry: Looks like someone has been here and changed things, and forgot to update the docstrings.

Issue #2559, opened 2019-09-30
errors: add hints to solve the errors, if possible


ghost:
We can take as an example the way Elm handles errors in the compiler: https://elm-lang.org/news/compiler-errors-for-humans

DVC already does this in some extent but I'm sure there's room for improvement on this one.

Also, thinking about introducing another method for logger: `hint`, that will output something like:

```
ERROR: this stuff failed
HINT: you could try X or Y prevent this from happening
```

(`HINT` could be colored in purple)

jmhsi:
Ivan was helping me on discord and it seems relevant:

My error:
ERROR: failed to run command - file/directory 'data/csvs/raw_zipped_csvs' is specified as an output in more than onestage: run/01_download_LC_csvs.dvc
    data/csvs/raw_zipped_csvs.dvc

Potential Hints:
HINT: try running `dvc pipeline show --ascii data/csvs/raw_zipped_csvs.dvc`
HINT: remove 'data/csvs/raw_zipped_csvs' as an output from "run/01_download_LC_csvs.dvc" or
    "data/csvs/raw_zipped_csvs.dvc" with `dvc remove`

efiop:
First step: improve ^ error to something like

```
ERROR: failed to run command - file/directory 'data/csvs/raw_zipped_csvs' is already taken by `data/csvs/raw_zipped_csvs.dvc`, consider using a different output name.
```

efiop:
Closing as too generic. Output message was improved, we are trying to fix things as we go.

PR #3213, opened 2020-01-22
push: improve 'no remote specified' hint

### LINKED ISSUE #3121: push: improve 'no remote specified hint
> More context around https://discordapp.com/channels/485586884165107732/563406153334128681/666010834702696450

```console
$ dvc version
DVC version: 0.80.0
Python version: 3.7.6
Binary: False
Package: brew

$ dvc init --no-scm
...
$ dvc push
ERROR: failed to push data to the cloud - config file error: no remote specified. Setup default remote with
    dvc config core.remote <name>
or use:
    dvc push -r <name>



```
- ✅  `config file error: no remote specified`: OK, this is the problem.
- ❌  `Setup default remote with... or use...`: Oops, actually I haven't even setup a remote. Maybe the hint should be different in this case?
- ❓  There 3 new lines between the hint and `Having any troubles? ...`

fabiosantoscode:
This tries to clear up the error message of NoRemoteError in case the user doesn't have any remotes, and provides instructions on how to create a remote, or a default remote. It also clears the extra newline at the end.

Fixes #3121

----

### DIFF
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -24,31 +24,7 @@ class ConfigError(DvcException):
 
 
 class NoRemoteError(ConfigError):
-    def __init__(self, command, *, config=None):
-        has_any_remote = False
-        if config:
-            has_any_remote = bool(
-                config.list_options(
-                    Config.SECTION_REMOTE_REGEX, Config.SECTION_REMOTE_URL
-                )
-            )
-
-        if has_any_remote:
-            msg = (
-                "no remote specified. Setup default remote with\n"
-                "    dvc remote default <remote name>\n"
-                "or use:\n"
-                "    dvc {} -r <remote name>".format(command)
-            )
-        else:
-            msg = (
-                "no remote specified. Create a remote with\n"
-                "    dvc remote add <remote name> <remote url>\n"
-                "To create a default remote use the -d flag:\n"
-                "    dvc remote add -d <remote name> <remote url>"
-            )
-
-        super().__init__(msg)
+    pass
 
 
 def supported_cache_type(types):
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -24,7 +24,13 @@ class ConfigError(DvcException):
 
 
 class NoRemoteError(ConfigError):
-    def __init__(self, command, *, has_any_remote=True):
+    def __init__(self, command, *, config=None):
+        has_any_remote = False
+        if config:
+            has_any_remote = config.list_options(
+                Config.SECTION_REMOTE_REGEX, Config.SECTION_REMOTE_URL
+            )
+
         if has_any_remote:
             msg = (
                 "no remote specified. Setup default remote with\n"
@@ -37,7 +43,7 @@ class NoRemoteError(ConfigError):
                 "no remote specified. Create a remote with\n"
                 "    dvc remote add <remote name> <remote url>\n"
                 "To create a default remote use the -d flag:\n"
-                "    dvc remote add -d <remote name> <remote url>\n"
+                "    dvc remote add -d <remote name> <remote url>"
             )
 
         super().__init__(msg)
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -24,13 +24,21 @@ class ConfigError(DvcException):
 
 
 class NoRemoteError(ConfigError):
-    def __init__(self, command):
-        msg = (
-            "no remote specified. Setup default remote with\n"
-            "    dvc remote default <name>\n"
-            "or use:\n"
-            "    dvc {} -r <name>\n".format(command)
-        )
+    def __init__(self, command, *, has_any_remote=True):
+        if has_any_remote:
+            msg = (
+                "no remote specified. Setup default remote with\n"
+                "    dvc remote default <remote name>\n"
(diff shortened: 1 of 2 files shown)


dvc/data_cloud.py
```diff
@@ -39,7 +39,9 @@ def get_remote(self, remote=None, command="<command>"):
         if remote:
             return self._init_remote(remote)
 
-        raise NoRemoteError(command)
+        has_any_remote = any(key.startswith("remote ") for key in self._config)
```
Suor: This feels as a part of the logic inside exception itself. So everything should be either inside or outside.

Also, `key.startswith("remote ")` is a bit fishy, we check for regex in other places.

fabiosantoscode: Got it, changed that to use `list_options` as I've found elsewhere. Remote URLs are required so finding any means that there's a valid remote that could be used.

dvc/config.py
```diff
@@ -24,13 +24,29 @@ def __init__(self, msg):
...
 
 class NoRemoteError(ConfigError):
-    def __init__(self, command):
-        msg = (
-            "no remote specified. Setup default remote with\n"
-            "    dvc remote default <name>\n"
-            "or use:\n"
-            "    dvc {} -r <name>\n".format(command)
-        )
+    def __init__(self, command, *, config=None):
```
pared: what is '*' for?

fabiosantoscode: It's to make sure `config` can only be used as a keyword argument.

fabiosantoscode: The reasoning for it is to prevent the exception from being initialized with some context that might not be a Config.

pared: So it is to make sure that we will never use it as a keyword argument? Isn't it kind of protecting the code from the developer?

fabiosantoscode: Exactly.

Suor:
Approved it.

A general consideration regarding messages like this - we've discussed adding more or less uniform HINTs in such situations, which will take into account context better, e.g. remote(s) name, having several remotes, command/options/api call. How this should be implemented and how far should we go is still TBD in #2559.

Suor:
Thx, @fabiosantoscode.

dvc/config.py
pared: Sorry, yeah, I meant that its to make sure that we always use it as a keyword argument.

pared: @fabiosantoscode but I still do not understand why would we need that here, what makes `config` so special that it needs to be always used as keyword argument?

dvc/config.py
```diff
@@ -24,13 +24,29 @@ def __init__(self, msg):
...
+                "or use:\n"
+                "    dvc {} -r <remote name>".format(command)
+            )
+        else:
+            msg = (
+                "no remote specified. Create a remote with\n"
+                "    dvc remote add <remote name> <remote url>\n"
+                "To create a default remote use the -d flag:\n"
+                "    dvc remote add -d <remote name> <remote url>"
+            )
```
efiop: Having such a complex logic in the exception is self is rather odd. Why not process config in the code itself and then raise NoRemoteError with the pre-crafted message?

dvc/config.py
```diff
@@ -24,13 +24,29 @@ def __init__(self, msg):
...
+            msg = (
+                "no remote specified. Setup default remote with\n"
+                "    dvc remote default <remote name>\n"
+                "or use:\n"
+                "    dvc {} -r <remote name>".format(command)
+            )
+        else:
+            msg = (
+                "no remote specified. Create a remote with\n"
+                "    dvc remote add <remote name> <remote url>\n"
```
efiop: This part is no relevan without `-d`, because if someone would've provided the remove, we would go `if remote` branch of execution (see below where you raise). So the hint here should assume `-d`.

dvc/config.py
fabiosantoscode: Nothing makes it special really. I just thought somebody might make a mistake and pass something else as an argument here. But I'll change it back.

dvc/config.py
fabiosantoscode: That's right!

dvc/config.py
fabiosantoscode: Works for me. The message was being generated in `__init__` so I kept it there.

fabiosantoscode:
@efiop I've tried to address your comments with a new commit. I've removed the `dvc remote add` without the `-d` and moved the message formatting out of the exception class.

dvc/config.py
pared: @fabiosantoscode this is a good point, and we actually have seen few times when wrong arguments are passed due to ordering, but I guess that there is no point in introducing that in this single place if we do differently everywhere else. We just need to be careful when using `kwargs` as `args`.

Suor: I am +1 on using keyword only args in situations like this.

pared: @Suor  So maybe we should enforce this rule throughout the code?

efiop: @pared @Suor I think we have a ticket for doing that in api.py already. Just need to be consistent, we weren't doing that earlier because py2 didn't support it, IIRC.

PR #3217, opened 2020-01-22
pipeline: show: outs: eliminate extra edges in DAG

### LINKED ISSUE #1744: pipeline: show: outs: improper DAG
See https://discordapp.com/channels/485586884165107732/485586884165107734/557205322641702912 . We should connect outputs to dependencies here https://github.com/iterative/dvc/blob/0.32.1/dvc/command/pipeline.py#L70 instead of connecting all outputs to all outputs of a dependant stage.

fabiosantoscode:
This fixes extra nodes edges appearing in `pipeline show --outs`, by using networkx to traverse the graph instead of listing all edges.

I spent a lot of time trying to figure out how to test this, so I thought I'd either leave it untested or take suggestions from the PR review.

Fixes #1744

----

### DIFF
--- a/dvc/command/pipeline.py
+++ b/dvc/command/pipeline.py
@@ -39,34 +39,37 @@ class CmdPipelineShow(CmdBase):
         target_stage = Stage.load(self.repo, target)
         G = get_pipeline(self.repo.pipelines, target_stage)
 
-        nodes = []
-        for stage in networkx.preorder_nodes(G, target_stage):
+        nodes = set()
+        for stage in networkx.dfs_preorder_nodes(G, target_stage):
             if commands:
                 if stage.cmd is None:
                     continue
-                nodes.append(stage.cmd)
+                nodes.add(stage.cmd)
             elif outs:
                 for out in stage.outs:
-                    nodes.append(str(out))
+                    nodes.add(str(out))
                 for dep in stage.deps:
-                    nodes.append(str(dep))
+                    nodes.add(str(dep))
             else:
-                nodes.append(stage.relpath)
+                nodes.add(stage.relpath)
 
         edges = []
-        for from_stage, to_stage in networkx.dfs_edges(G, target_stage):
-            if commands:
-                if to_stage.cmd is None:
-                    continue
-                edges.append((from_stage.cmd, to_stage.cmd))
-            elif outs:
-                for from_dep in from_stage.deps:
-                    for from_out in from_stage.outs:
-                        edges.append((str(from_out), str(from_dep)))
-            else:
-                edges.append((from_stage.relpath, to_stage.relpath))
 
-        return nodes, edges, networkx.is_tree(G)
+        if outs and not commands:
+            for stage in networkx.dfs_preorder_nodes(G, target_stage):
+                for dep in stage.deps:
+                    for out in stage.outs:
+                        edges.append((str(out), str(dep)))
+        else:
+            for from_stage, to_stage in networkx.dfs_edges(G, target_stage):
+                if commands:
+                    if to_stage.cmd is None:
+                        continue
+                    edges.append((from_stage.cmd, to_stage.cmd))
+                else:
+                    edges.append((from_stage.relpath, to_stage.relpath))
+
+        return list(nodes), edges, networkx.is_tree(G)
 
     def _show_ascii(self, target, commands, outs):
         from dvc.dagascii import draw
--- a/dvc/command/pipeline.py
+++ b/dvc/command/pipeline.py
@@ -31,7 +31,7 @@ class CmdPipelineShow(CmdBase):
             else:
                 logger.info(stage.path_in_repo)
 
-    def __build_graph(self, target, commands, outs):
+    def _build_graph(self, target, commands, outs):
         import networkx
         from dvc.stage import Stage
         from dvc.repo.graph import get_pipeline
@@ -74,7 +74,7 @@ class CmdPipelineShow(CmdBase):
     def _show_ascii(self, target, commands, outs):
         from dvc.dagascii import draw
 
-        nodes, edges, _ = self.__build_graph(target, commands, outs)
+        nodes, edges, _ = self._build_graph(target, commands, outs)
 
         if not nodes:
             return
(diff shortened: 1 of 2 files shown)


fabiosantoscode:
@Suor I initially looked at the issue description on Discord and tried to reproduce the same issue. This fixes that diamond problem, where A generates B and C, and B doesn't go on to be a dependency of the ultimate target but is still in the graph.

I'll try to reproduce the more general issue and see what fixes it.

fabiosantoscode:
@Suor there. It's getting pretty nested and I could use itertools.product() to make those two for loops just one, but it's probably unnecessarily complicated.

fabiosantoscode:
Wait, that implementation is incorrect.

fabiosantoscode:
This should do it.

dvc/command/pipeline.py
```diff
@@ -36,31 +36,33 @@ def __build_graph(self, target, commands, outs):
...
             if commands:
                 if to_stage.cmd is None:
                     continue
                 edges.append((from_stage.cmd, to_stage.cmd))
             elif outs:
-                for from_out in from_stage.outs:
-                    for to_out in to_stage.outs:
-                        edges.append((str(from_out), str(to_out)))
+                for from_dep in from_stage.deps:
+                    for from_out in from_stage.outs:
```
Suor: So we are not using `to_stage` here anymore, looks like a bug.

fabiosantoscode: This seems to be complete. I think it's because we're iterating over the deps (as opposed to the outs) and therefore we have an idea of where the dep came from.

Another way of doing this would be to iterate using preorder_nodes only if the `--outs` option was enabled.

Suor: But, code looks strange. If we show outs/deps then stages are edges, not pairs of stages. Why we iterate pairs then? Won't we miss something in the end?

Let's make some test, which will make it more obvious.

fabiosantoscode: I've changed it so that when we're showing outs, we're iterating over stages (which are the vertices/nodes in the graph).

I've also added a test.

tests/func/test_pipeline.py
```diff
@@ -98,6 +99,29 @@ def test_dot_commands(self):
...
+
+    dvc.add("base")
+    dvc.run(deps=["base"], outs=["derived1"], cmd="echo derived1 > derived1")
+    dvc.run(deps=["base"], outs=["derived2"], cmd="echo derived2 > derived2")
+    final_stage = dvc.run(
+        deps=["derived1"], outs=["final"], cmd="echo final > final"
+    )
+
+    args = ["pipeline", "show", "--outs", "--dot", final_stage.path]
+    with tmp_dir.chdir():
```
efiop: You are already in tmp_dir, so no need to chdir. Or am I missing something here?

fabiosantoscode: That is correct! I didn't know about that. Updating.

tests/func/test_pipeline.py
```diff
@@ -98,6 +99,28 @@ def test_dot_commands(self):
...
+    )
+
+    args = ["pipeline", "show", "--outs", "--dot", final_stage.path]
+    command = CmdPipelineShow(args)
+    # Need to test __build_graph directly
+    nodes, edges, is_tree = command._CmdPipelineShow__build_graph(
+        final_stage.path, commands=False, outs=True
+    )
+
+    assert set(nodes) == set(["final", "derived1", "base"])
```
pared: no need for converting list to set, `set(["final", "derived1", "base"])` == `{"final", "derived1", "base"}`

tests/func/test_pipeline.py
```diff
@@ -98,6 +99,28 @@ def test_dot_commands(self):
...
+    tmp_dir.dvc_gen({"base": "base"})
+
+    dvc.add("base")
+    dvc.run(deps=["base"], outs=["derived1"], cmd="echo derived1 > derived1")
+    dvc.run(deps=["base"], outs=["derived2"], cmd="echo derived2 > derived2")
+    final_stage = dvc.run(
+        deps=["derived1"], outs=["final"], cmd="echo final > final"
+    )
+
+    args = ["pipeline", "show", "--outs", "--dot", final_stage.path]
```
pared: If we are calling `__build_graph` anyway, why would we provide those args? You can simply initialize command with empty list. :)

tests/func/test_pipeline.py
```diff
@@ -98,6 +99,28 @@ def test_dot_commands(self):
         self.assertEqual(ret, 0)
 
 
+def test_disconnected_stage(tmp_dir, dvc, caplog):
```
pared: caplog seems to be unused

efiop:
> Another thing, not necessarily in this change: @efiop, do you think that we should keep the name of the method __build_graph

Yeah, that hurts my eyes too, it was introduced a few years back by one of the contributors and we didn't touch that since. We should change it, but not necessarily in this PR:slightly_smiling_face:

pared:
@efiop that is very tactful of you, to say "one of our contributors" and not "you did that, remember?"
@fabiosantoscode If it's not a problem, I would suggest refactoring the name `__build_graph` -> `_build_graph`. If you mind, it can stay as is.

efiop:
> @efiop that is very tactful of you, to say "one of our contributors" and not "you did that, remember?"

Wait, I didn't mean that, though. I don't think it was you, was it? :eyes:

dvc/command/pipeline.py
```diff
@@ -36,35 +36,40 @@ def __build_graph(self, target, commands, outs):
...
-                edges.append((from_stage.cmd, to_stage.cmd))
-            elif outs:
-                for from_out in from_stage.outs:
-                    for to_out in to_stage.outs:
-                        edges.append((str(from_out), str(to_out)))
-            else:
-                edges.append((from_stage.relpath, to_stage.relpath))
 
-        return nodes, edges, networkx.is_tree(G)
+        if outs and not commands:
```
Suor: Aren't this two mutually exclusive?

fabiosantoscode:
I'll remove the `__` for a regular `_`. I agree that it's a bit uncalled for.

dvc/command/pipeline.py
fabiosantoscode: Yes they seem to be!

fabiosantoscode:
I believe I've addressed all of the feedback. Additionally, I found a `__write_dot` method and turned it into `_write_dot`.

PR #3220, opened 2020-01-23
get: copy edits aroud --show-url, "dvc", et al.

jorgeorpinel:
> Per https://github.com/iterative/dvc/pull/3156#pullrequestreview-345562608

Docs PR already exists: iterative/dvc.org/pull/926

### DIFF
--- a/dvc/command/get.py
+++ b/dvc/command/get.py
@@ -19,7 +19,7 @@ class CmdGet(CmdBaseNoRepo):
             )
             logger.info(url)
         except DvcException:
-            logger.exception("failed to show url")
+            logger.exception("Failed to show URL")
             return 1
 
         return 0
@@ -78,6 +78,6 @@ def add_parser(subparsers, parent_parser):
     get_parser.add_argument(
         "--show-url",
         action="store_true",
-        help="Returns path/url to the location in remote for specified path",
+        help="Returns URL to the storage location of the target data.",
     )
     get_parser.set_defaults(func=CmdGet)
--- a/dvc/command/get.py
+++ b/dvc/command/get.py
@@ -19,7 +19,7 @@ class CmdGet(CmdBaseNoRepo):
             )
             logger.info(url)
         except DvcException:
-            logger.exception("Failed to show URL")
+            logger.exception("failed to show URL")
             return 1
 
         return 0
--- a/dvc/command/get.py
+++ b/dvc/command/get.py
@@ -78,7 +78,7 @@ def add_parser(subparsers, parent_parser):
     get_parser.add_argument(
         "--show-url",
         action="store_true",
-        help="Just print the storage location (URL) the target data would be "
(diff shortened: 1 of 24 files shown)


dvc/command/get.py
```diff
@@ -78,6 +78,7 @@ def add_parser(subparsers, parent_parser):
     get_parser.add_argument(
         "--show-url",
         action="store_true",
-        help="Returns path/url to the location in remote for specified path",
+        help="Just print the storage location (URL) the target data would be "
```
skshetry: **Just** feels weird in help text. How about:
```suggestion
        help="Prints the storage location (URL) the target data would be "
```

jorgeorpinel: "Just" means that the command won't do anything else i.e. it won't download the target.

shcheklein: can we use `Only` instead ...or `print ... and exit`? I was confused with Just either

jorgeorpinel:
The last commit came out of https://github.com/iterative/dvc/pull/3156#pullrequestreview-345562608 but I didn't expect how many instances of the term "dvc" (lower case) in strings there are (`grep /".*dvc/`). I reviewed through the dvc/command/ dir so far. Depending on reaction to this PR, may go over the rest of the code base when I have a chance.

dvc/command/install.py
```diff
@@ -13,7 +13,7 @@ def run(self):
         try:
             self.repo.install()
         except Exception:
-            logger.exception("failed to install dvc hooks")
+            logger.exception("Failed to install DVC Git hooks")
```
shcheklein: most likely exceptions don't start with an upper case for a reason - this message just a small part of the whole thing when it's printed

dvc/command/get.py
jorgeorpinel: OK using "print and exit" but I think it's just a confusing option in general. Hopefully it will be more intuitive when it's moved to `dvc list`.

jorgeorpinel: Done in 6364c9ee.

shcheklein: We can be even more explicit - `Skip downloading, print ... and exit`

dvc/command/install.py
jorgeorpinel: It comes after `ERROR:` (at least per https://github.com/iterative/dvc/pull/3156#discussion_r368929870). From what I understand, a full sentence is capitalized after `:`

Of course this is logger output so grammar isn't very important. As you guys prefer, I can revert all these, lmk pls.

jorgeorpinel: I guess [it's optional](https://www.grammarly.com/blog/capitalization-after-colons/) anyway. Reverted all these in 69c1aa5b.

shcheklein: I would avoid touching this one for now, too many questions ...

PR #3222, opened 2020-01-23
api: fallback to external_repo if not a DVC repository

### LINKED ISSUE #3221: api: does not work with git only local repository
```sh
➜ cd "$(mktemp -d)" 
➜ git init
Initialized empty Git repository in /tmp/tmp.TGV6SVXYsz/.git/

➜ echo "foo" > foo
➜ git add foo
➜ git -c commit.gpgsign=false commit -m "first"
[master (root-commit) 136fd94] first
 1 file changed, 1 insertion(+)
 create mode 100644 foo

➜ python -c '
➜ import dvc.api
➜ with dvc.api.open(path="foo", repo=".") as fd:
➜     print(fd.read())
➜ '
  File "<string>", line 3, in <module>
  File "/home/saugat/.pyenv/versions/3.6.6/lib/python3.6/contextlib.py", line 81, in __enter__
    return next(self.gen)
  File "/home/saugat/repos/iterative/dvc/dvc/api.py", line 95, in _open
  File "/home/saugat/.pyenv/versions/3.6.6/lib/python3.6/contextlib.py", line 81, in __enter__
    return next(self.gen)
  File "/home/saugat/repos/iterative/dvc/dvc/api.py", line 113, in _make_repo
    yield Repo(repo_url)
  File "/home/saugat/repos/iterative/dvc/dvc/repo/__init__.py", line 76, in __init__
    root_dir = self.find_root(root_dir)
  File "/home/saugat/repos/iterative/dvc/dvc/repo/__init__.py", line 144, in find_root
    raise NotDvcRepoError(root)
dvc.exceptions.NotDvcRepoError: you are not inside of a dvc repository (checked up to mount point '/tmp')
```

This is because of the recent changes in `_make_repo` that instead of falling back to external repo on error, it assumes the local path to be a DVC repository.
https://github.com/iterative/dvc/blob/e9caa44b554daf7f0f21dc2bb9bb0df028369fb1/dvc/api.py#L112-L113

skshetry:
Fixes #3221. Now, it will just fallback to the `external_repo` if `Repo()` could not be instantiated.
Another idea would have been to check for the existence of `.dvc` directory, but, I went with this, as it feels correct and *robust*.

### DIFF
--- a/dvc/api.py
+++ b/dvc/api.py
@@ -108,8 +108,9 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
 
 
 @contextmanager
-def _make_repo(repo_url, rev=None):
-    if rev is None and (not repo_url or os.path.exists(repo_url)):
+def _make_repo(repo_url=None, rev=None):
+    repo_url = repo_url or os.getcwd()
+    if rev is None and os.path.exists(repo_url):
         try:
             yield Repo(repo_url)
             return
--- a/dvc/api.py
+++ b/dvc/api.py
@@ -10,7 +10,7 @@ import ruamel.yaml
 from voluptuous import Schema, Required, Invalid
 
 from dvc.repo import Repo
-from dvc.exceptions import DvcException
+from dvc.exceptions import DvcException, NotDvcRepoError
 from dvc.external_repo import external_repo
 
 
@@ -110,10 +110,13 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
 @contextmanager
 def _make_repo(repo_url, rev=None):
     if rev is None and (not repo_url or os.path.exists(repo_url)):
-        yield Repo(repo_url)
-    else:
-        with external_repo(url=repo_url, rev=rev) as repo:
-            yield repo
+        try:
+            yield Repo(repo_url)
+            return
+        except NotDvcRepoError:
+            pass  # fallthrough to external_repo
+    with external_repo(url=repo_url, rev=rev) as repo:
+        yield repo
 
 
 def summon(name, repo=None, rev=None, summon_file="dvcsummon.yaml", args=None):
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -128,10 +128,10 @@ class ExternalGitRepo:
             raise PathMissingError(path, self.url)
 
     @contextmanager
-    def open_by_relpath(self, path, mode="r", encoding=None):
+    def open_by_relpath(self, path, mode="r", encoding=None, **kwargs):
         try:
             abs_path = os.path.join(self.root_dir, path)
-            with open(abs_path, mode, encoding) as fd:
+            with open(abs_path, mode, encoding=encoding) as fd:
                 yield fd
         except FileNotFoundError:
             raise PathMissingError(path, self.url)
--- a/tests/func/test_api.py
+++ b/tests/func/test_api.py
@@ -8,7 +8,7 @@ import pytest
 from dvc import api
 from dvc.api import SummonError, UrlNotDvcRepoError
 from dvc.compat import fspath
-from dvc.exceptions import FileMissingError, NotDvcRepoError
+from dvc.exceptions import FileMissingError
 from dvc.main import main
 from dvc.path_info import URLInfo
 from dvc.remote.config import RemoteConfig
@@ -54,7 +54,7 @@ def test_get_url_external(erepo_dir, remote_url):
 def test_get_url_requires_dvc(tmp_dir, scm):
     tmp_dir.scm_gen({"foo": "foo"}, commit="initial")
 
-    with pytest.raises(NotDvcRepoError, match="not inside of a DVC repo"):
+    with pytest.raises(UrlNotDvcRepoError, match="not a DVC repository"):
         api.get_url("foo", repo=fspath(tmp_dir))
 
     with pytest.raises(UrlNotDvcRepoError):


dvc/external_repo.py
```diff
@@ -128,10 +128,10 @@ def pull_to(self, path, to_info):
             raise PathMissingError(path, self.url)
 
     @contextmanager
-    def open_by_relpath(self, path, mode="r", encoding=None):
+    def open_by_relpath(self, path, mode="r", encoding=None, **kwargs):
```
skshetry: open_by_relpath on the external repo also takes more arguments such as `remote`, etc which does not make sense here.

dvc/external_repo.py
```diff
@@ -128,10 +128,10 @@ def pull_to(self, path, to_info):
...
         try:
             abs_path = os.path.join(self.root_dir, path)
-            with open(abs_path, mode, encoding) as fd:
+            with open(abs_path, mode, encoding=encoding) as fd:
```
skshetry: Third argument is `buffering`.

efiop:
@skshetry Please check the tests, they've failed.

dvc/api.py
```diff
@@ -240,4 +243,6 @@ def _import_string(import_name):
 
 def _require_dvc(repo):
     if not isinstance(repo, Repo):
+        if os.path.exists(repo.url):
+            raise NotDvcRepoError(repo.url)
```
skshetry: The assumptions in tests were correct, only needed to add this.

skshetry: Is this correct (the given path is a git only repo)? @efiop 
```sh
➜ dvc get /tmp/tmp.Fn43F3mvVy foo --show-url   
ERROR: failed to show url - you are not inside of a dvc repository (checked up to mount point '/tmp')
```

skshetry: We probably need a different exception, one without `you are not inside of`.

skshetry: Okay, I'll make this change on a different PR. I will use the same exception, but somehow throw different message.

efiop: @skshetry Could you please elaborate a bit why this `if` is needed? Also, why doesn't it raise `UrlNotDvcRepoError`?

skshetry: I did not want to change the test. I felt it was right to raise NotDvcRepoError rather than UrlNotDvcRepoError, as expected in the test. It just felt more correct to me. :)

efiop: @skshetry which test though? :) Fixing code for the sake of test seems suspicious a bit, hence why I'm wondering.

skshetry: This is the test:
https://github.com/iterative/dvc/blob/ad6e58a45729c945e3e69bd9c949f8b27d55377e/tests/func/test_api.py#L54

skshetry: The only reason I used this is better wording:
`URL {} is not a DVC repository` vs worse `you are not inside of a repository`.

For local erepo, latter made more sense to me.

efiop: @skshetry But for local repo it also doesn't make sense, as you might be `dvc get`ing from outside of the local repo, so that message becomes confusing.  `UrlNotDvcRepoError` is more appropriate there.

skshetry: Yes, the message is worse now then before. Just the `URLNotDvcRepoError` is imo better for url cases, and `NotDvcRepoError` for local cases (with external repo cases).

I was thinking of using `NotDvcRepoError` and changing message if it's for external repo in a different PR. But, I'll just rollback, as the change is confusing.

dvc/api.py
```diff
@@ -110,10 +110,13 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
...
-        yield Repo(repo_url)
-    else:
-        with external_repo(url=repo_url, rev=rev) as repo:
-            yield repo
+        try:
+            yield Repo(repo_url)
+            return
+        except NotDvcRepoError:
+            pass  # fallthrough to external_repo
+    with external_repo(url=repo_url, rev=rev) as repo:
```
Suor: This will fails if `repo_url` is `None`, which means cur dir really.

dvc/api.py
Suor: If function is called `_require_dvc()` it should require dvc. 

@skshetry local repo is not always the one you are in, so it still should be `UrlNotDvcRepoError`.

dvc/external_repo.py
Suor: If it works it should stay as is. Let's not add ignored arguments, this will lead to silent errors.

skshetry: It does not as it also gets unexpected arguments from `Repo.open_by_relpath`:
https://github.com/iterative/dvc/blob/49039968f74282bdab3aa05b4582e5d392fc0ccc/dvc/repo/__init__.py#L430

skshetry: We could though add `remote`, but, I prefer this.

Suor: I don't understand. This line is never called from Repo.open_by_relpath().

dvc/api.py
efiop: @Suor Is there a test for that? I don't remember us discussing `None` here at all.

Suor: In all api funcs, `repo` defaults to `None`, which means current dir, same for `Repo()`.

skshetry: Fixed by cdd9d0d78488daf159e1d0847a421d4929f44d61

dvc/api.py
```diff
@@ -108,12 +108,16 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
 
 
 @contextmanager
-def _make_repo(repo_url, rev=None):
+def _make_repo(repo_url=None, rev=None):
+    repo_url = repo_url or os.getcwd()
```
efiop:
```suggestion
    repo_url = repo_url or "."
```

dvc/external_repo.py
skshetry: It's called by `dvc.api.open()` here with following args:
https://github.com/iterative/dvc/blob/49039968f74282bdab3aa05b4582e5d392fc0ccc/dvc/api.py#L70-L79

And, successively, `_open` calls this line:
https://github.com/iterative/dvc/blob/49039968f74282bdab3aa05b4582e5d392fc0ccc/dvc/api.py#L94-L100

Suor: Ok, then we should add only `remote` param not `**kwargs` and `assert remote is None`. This way things won't be silently ignored.

efiop: @Suor assert is not needed, as the point of that method is to provide the same interface for git and dvc repos, so it should be silently ignored there. Or am I missing something?

PR #3225, opened 2020-01-23
remote: verify cache exists on dir checksum

### LINKED ISSUE #2647: remote: getting a directory checksum requires uploading a file to an external cache
**Version**: 0.62.1

**Description**: The current approach to calculate a **directory checksum** is to collect its contents' checksum and then "upload" it and get the checksum provided by the remote:

https://github.com/iterative/dvc/blob/a091b67341b00fea8ef43906f59c88cb38f384ff/dvc/remote/base.py#L246-L256

For example, if it is an S3 remote, DVC will **create a file** containing an "index" of the directory **with its checksums** on the S3 cache **and then will try to get the ETag of such file**.

This **is an easy way to check if the directory has changed**. Since the "directory index" is already there, **you are a `head` operation away** from knowing if the directory changed or not. Otherwise, you'll need to recompute the file on the spot.

Right now **we don't require setting up a cache for using external dependencies, but as mentioned above, this is needed for directories.**

Let's discuss if there's another way around uploading files or just settle with that implementation and modify the code and docs accordingly.

**Related**: #1654

pared:

Fixes #2647 

[asciicast](https://asciinema.org/a/295538)

### DIFF
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -1,6 +1,6 @@
 """Exceptions raised by the dvc."""
 
-from dvc.utils import relpath
+from dvc.utils import relpath, format_link
 
 
 class DvcException(Exception):
@@ -310,3 +310,18 @@ class PathMissingError(DvcException):
             " neither as an output nor a git-handled file."
         )
         super().__init__(msg.format(path, repo))
+
+
+class RemoteCacheRequiredError(DvcException):
+    def __init__(self, path_info):
+        super().__init__(
+            (
+                "Current operation was unsuccessful because '{}' requires "
+                "existing cache on '{}' remote. See {} for information on how "
+                "to set up remote cache."
+            ).format(
+                path_info,
+                path_info.scheme,
+                format_link("https://man.dvc.org/config#cache"),
+            )
+        )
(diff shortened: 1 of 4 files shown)


dvc/remote/base.py
```diff
@@ -233,6 +241,9 @@ def _collect_dir(self, path_info):
         return sorted(result, key=itemgetter(self.PARAM_RELPATH))
 
     def get_dir_checksum(self, path_info):
+        if not self.cache:
```
efiop: Would it make sense to unify it somehow with https://github.com/iterative/dvc/blob/master/dvc/output/base.py#L102 ?

pared: @efiop Do you mean, using same error in both places, or using `self.REMOTE.scheme` to point user in right direction?

efiop: @pared Well, maybe both? :)

dvc/exceptions.py
```diff
@@ -310,3 +310,11 @@ def __init__(self, path, repo):
             " neither as an output nor a git-handled file."
         )
         super().__init__(msg.format(path, repo))
+
+
+class RemoteCacheRequiredError(DvcException):
+    def __init__(self, scheme):
+        super().__init__(
+            "Remote cache for '{}' remote is required to perform this "
+            "operation".format(scheme)
```
efiop: Could we add something helpful or add a link to docs?

efiop: Also "this operation" is not very clear. Maybe we could include path too to be more precise about what path has caused this exception?

Issue #2982, opened 2019-12-19
diff: clean up output for changed files


dmpetrov:
Context:
```
$ dvc -V
0.77.3+e03136
# all txt files are data files
$ ls
newfile.txt     small.txt       ttt.txt
newfile.txt.dvc small.txt.dvc   ttt.txt.dvc
```

```
$ echo qwert >> small.txt
$ dvc add small.txt
$ git add small.txt.dvc
$ git commit -m hello
$ dvc diff HEAD^
dvc diff from ada9d97 to 1a22314 .  # <<-- "diff ada9d97..1a22314" might be enough

diff for 'small.txt'
-small.txt with md5 5f73f210b6202aa07279cdff6776b64f
+small.txt with md5 b4301ef665d0343e028fcd5821654edc

added file with size 6 Bytes .  # <<-- File was NOT added, it was modified. 6 bytes is diff, not size.

diff for 'ttt.txt'
-ttt.txt with md5 92a47ab38589702d290647ce24386181
+ttt.txt with md5 92a47ab38589702d290647ce24386181

file size was not changed .  # <<-- what is a reason to output this then? Also, is it about the previous diff or the next one - hard to read?

diff for 'newfile.txt'
-newfile.txt with md5 99370a5386cac00c93c9fdc836076a7d
+newfile.txt with md5 99370a5386cac00c93c9fdc836076a7d

file size was not changed  # <<-- what is a reason to output it then?
```

Also, it would be great to simplify this output for parsing. 

Possible "clean" output:
```
$ dvc diff HEAD^
diff ada9d97..1a22314
Changed 'small.txt' +6 bytes
md5 5f73f210b6202aa07279cdff6776b64f b4301ef665d0343e028fcd5821654edc
New 'fakefile.txt' +13728494 bytes
md5 92a47ab38589702d290647ce24386181 99370a5386cac00c93c9fdc836076a7d
Removed 'fakefile_2222.txt' +8326436 bytes
md5 e0313660cfc07a10417543b8e4d08bea9 f4daf3bacc9a9494df7d641f572e92b66
```

Actions:
- [ ] Not show not-changed files (at least if it was not specified explicitly).
- [ ] Clearly distinguish added files and modified files - with the sizes.
- [ ] Improve output for reading. ~~not empty lines within the context of a single file (preferably not empty lines at all).~~ See https://github.com/iterative/dvc/issues/2982#issuecomment-568715481.
- [ ] Simplify the format for parsing. See `--porcelain`

[...]

Issue #2999, opened 2019-12-23
diff for not git-commited files


dmpetrov:
It feels like `diff` can do a better job on not committed changes.

```
$ dd if=/dev/random of=small.txt bs=1024000000 count=1 # modify the file
$ dvc add small.txt
$ dvc diff -t small.txt HEAD          # IT SUPPOSE TO BE DIFFERENT FROM HEAD
dvc diff from 2dae4a2 to 2dae4a2   # IT IS NOT :(  

$ dvc diff -t small.txt HEAD^
dvc diff from 8a1aa99 to 2dae4a2 .   # IT IS NOT DIFFERENT EVEN WITH OLDER VERSION:(

diff for 'small.txt'
-small.txt with md5 b4301ef665d0343e028fcd5821654edc
+small.txt with md5 b4301ef665d0343e028fcd5821654edc

file size was not changed

$ git add small.txt.dvc
$ git commit -m change
$ dvc diff -t small.txt HEAD^ . # NOW YOU CAN SEE THE DIFFERENCE
dvc diff from 2dae4a2 to 12ed32b

diff for 'small.txt'
-small.txt with md5 b4301ef665d0343e028fcd5821654edc
+small.txt with md5 4be464ee7907d9d5e558133907bb3510

added file with size 1.0 GB
```

jorgeorpinel:
Just to clarify, this is about uncommitted changes in Git+DVC repos, right? Not about `dvc diff` working for non-Git DVC projects.

If so, it was probably addressed by #3229 

```console
# Install pre-release bc new `dvc diff` isn't released yet as of rn:
$ virtualenv -p python3 .venv
...
$ source .venv/bin/activate
$ pip install git+https://github.com/iterative/dvc
...

$ git init
$ echo ".venv" >> .git/info/exclude
$ dvc init
$ git commit -am "New DVC repo"

$ echo "data" > small.txt
$ .venv/bin/dvc diff  # DEFAULTS TO HEAD
Added:                                                                  
    small.txt

summary: added (1), deleted (0), modified (0)
dvc diff -t small.txt HEAD^  # THIS WON'T WORK HERE BECAUSE THERE'S NO HEAD^ BUT OTHERWISE IT WOULD, I THINK
ERROR: unexpected error - Cmd('git') failed due to: exit code(128)      
  cmdline: git rev-parse --short HEAD^
[...]

Issue #2998, opened 2019-12-23
diff with no cache


dmpetrov:
```
$ dvc -V
0.77.3+e03136
```

The major job of `dvc diff` is to show if there is a file difference and show checksums. This can be done without actual file content. So, no cache is needed. But it looks like the current implementation requires the file content.

It is okay not to show the size diff if the cache is not presented. But the format needs to be changed - `size is ?` not the best way to show the absence of the sizes.

With cache:
```
$ dvc diff -t small.txt 9340a9bb6d49b073b6
dvc diff from 9340a9b to 2dae4a2

diff for 'small.txt'
-small.txt with md5 5f73f210b6202aa07279cdff6776b64f
+small.txt with md5 b4301ef665d0343e028fcd5821654edc

added file with size 6 Bytes
```

Without cache:
```
$ dvc diff -t small.txt 9340a9bb6d49b073b6
dvc diff from 9340a9b to 2dae4a2

diff for 'small.txt'
-small.txt with md5 5f73f210b6202aa07279cdff6776b64f

size is ?
```

jorgeorpinel:
> ...to show if there is a file difference and show checksums
> It is okay not to show the size diff if the cache is not presented. But the format needs to be changed
> ...it looks like the current implementation requires the file content.

It seems none of this applied for new revision of `dvc diff`, after #3229 (I tested). Should we close this?

efiop:
@jorgeorpinel Thanks for the heads up! It wasn't auto-closed because of the formatting. @mroutis Please note that if you close tickets in bulk, you need to add `fix` prefix to every one of those (thought there might be a better way).

mroutis:
Oh! I see, will properly check it next time.

jorgeorpinel:
Oh I see what happened!

![image](https://user-images.githubusercontent.com/1477535/73521341-60d1b680-43cb-11ea-9e99-ab8f94a9b404.png)

Cool, glad it's figured out now 🙂

dmpetrov:
See #3392

dmpetrov:
It seems like it is impossible to get a proper diff with no cache for data directories 😞

Some workaround is needed. Something like pulling data-dir caches when calculating `dvc diff`.

efiop:
Closing in favor of https://github.com/iterative/dvc/issues/3392

Issue #2372, opened 2019-08-06
Possibly an incorrect value in a test in diff.DIFF_IDENT


GildedHonour:
A test at https://github.com/iterative/dvc/blob/master/tests/func/test_diff.py#L207 passes correctly. However, it seems that there's a bug in it:

    # [............]
    def setUp(self):
        # [............]

        self.a_ref = self.git.git.rev_parse(self.git.head.commit, short=True)
        self.old_checksum = _get_checksum(self.dvc, self.DATA_DIR)
        self.new_file_content = "new_test_file_bigger_content_123456789"
        self.diff_len = len(self.new_file_content)
        self.create(self.new_file, self.new_file_content)
        self.dvc.add(self.DATA_DIR)
        self.git.index.add([self.DATA_DIR + ".dvc"])
        self.git.index.commit(message="modify file in the data dir")
        self.new_checksum = _get_checksum(self.dvc, self.DATA_DIR)
        self.b_ref = self.git.git.rev_parse(self.git.head.commit, short=True)

    def test(self):
        result = self.dvc.diff(self.a_ref, target=self.DATA_DIR)
        test_dct = {
            diff.DIFF_A_REF: self.git.git.rev_parse(self.a_ref, short=True),
            diff.DIFF_B_REF: self.git.git.rev_parse(self.b_ref, short=True),
            diff.DIFF_LIST: [
                {
                    diff.DIFF_IDENT: 2,  # <------
                    diff.DIFF_CHANGE: 1,
                    diff.DIFF_DEL: 0,
                    diff.DIFF_MOVE: 0,
                    diff.DIFF_NEW: 0,
                    # [.............]

                }
            ],
        }
        self.assertEqual(test_dct, result)

And

    class TestDiffDir(TestDvcGit):
        def setUp(self):
            super(TestDiffDir, self).setUp()

            self.dvc.add(self.DATA_DIR)
            self.git.index.add([self.DATA_DIR + ".dvc"])
            self.git.index.commit("adds data_dir")
            self.a_ref = self.git.git.rev_parse(
                self.dvc.scm.git.head.commit, short=True
            )
            self.old_checksum = _get_checksum(self.dvc, self.DATA_DIR)
            self.new_file = os.path.join(self.DATA_SUB_DIR, diff.DIFF_NEW_FILE)
            self.create(self.new_file, self.new_file)
            self.dvc.add(self.DATA_DIR)
            self.git.index.add([self.DATA_DIR + ".dvc"])
[...]

PR #3229, opened 2020-01-24
diff: reimplement interface and tests from scratch

mroutis:
Close: #2982, #2998, #2999, #1893, #2372
Related: #770

* Shows the difference between HEAD and the current workspace by default
* Don't show files that stayed the same between both trees.
* Remove `--target` option.
* Checksums are only displayed when `--checksum` option is provided.
* Format diff for machine readability when `--show-json` option is given.
* Changes in output:
  * Remove file size information
  * Recurse directory outputs and display its cached information.
  * Group diff by state (i.e. "Added", "Deleted", "Modified")
* Allow running `dvc diff` when there's no cache available.
* Update completion scripts
* Update documentation (https://github.com/iterative/dvc.org/pull/953)

Questions:
  * Should we introduce a `--recursive` option to explicitly recurse directory outputs?
  * Should we support `renamed` files?


dvc/command/diff.py
```diff
@@ -147,15 +31,6 @@ def add_parser(subparsers, parent_parser):
...
-    diff_parser.add_argument(
-        "-t",
-        "--target",
-        help=(
-            "Source path to a data file or directory. Default None. "
-            "If not specified, compares all files and directories "
-            "that are under DVC control in the current working space."
-        ),
-    )
     diff_parser.add_argument(
```
efiop: This one is no longer a required one, it should be optional, same as in `dvc metrics diff`.

dvc/repo/diff.py
```diff
@@ -1,246 +1,56 @@
...
-    for key, value in b_entries.items():
-        if key not in a_entries:
-            new_count += 1
-            diff_size += os.path.getsize(self.cache.local.get(b_entries[key]))
-    return new_count, diff_size
-
+def diffable_from_output(output):
+    try:
+        size = os.path.getsize(fspath(output.path_info))
+    except OSError:
```
efiop: You need to ignore ENOENT (or the corersponding exception, not OSERrror in general)

tests/func/test_diff.py
```diff
@@ -1,267 +1,86 @@
...
-from dvc.scm.base import FileNotInCommitError
-from tests.basic_env import TestDvcGit
 
+def test_no_scm(tmp_dir, dvc):
+    tmp_dir.dvc_gen("file", "text")
 
-class TestDiff(TestDvcGit):
-    def setUp(self):
-        super().setUp()
+    pytest.skip("TODO: define behavior, should it fail?")
```
efiop: So this is a WIP PR, right?

efiop:
@mroutis What is the point of these tests if they are all commented out and TODos?

dvc/repo/brancher.py
```diff
@@ -41,6 +41,8 @@ def brancher(  # noqa: E302
         if all_tags:
             revs.extend(scm.list_tags())
 
+    revs = filter(None, revs)
```
efiop: Why is this one needed in this PR?

efiop:
> As for the CLI, I was thinking on using the following as output:

I think the ticket was saying that we shouldn't show checksums by default.

dvc/repo/brancher.py
mroutis: To prevent passing `None` arguments (e.g. `revs=["HEAD", None]`) in the following statement:

```python
for branch in self.brancher(revs=[a_ref, b_ref]):
```

That line would need to exist either on `diff` or `brancher` and I decided moving it to `brancher`, since the reader would have more context to understand what would happen if you pass `None` inside the `revs`, otherwise I found myself explaining why I was filtering `None` on `diff` and it looked out of place.

dvc/command/diff.py
```diff
@@ -1,135 +1,19 @@
...
-        logger.info(msg)
-        return msg
-
     def run(self):
         try:
-            msg = self.repo.diff(
-                self.args.a_ref, target=self.args.target, b_ref=self.args.b_ref
-            )
-            self._show(msg)
+            self.repo.diff(self.args.a_ref, self.args.b_ref, self.args.target)
```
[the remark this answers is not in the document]
mroutis: Totally, will add them when we've agreed on the structure)

dvc/command/diff.py
```diff
@@ -1,135 +1,19 @@
...
         except DvcException:
-            msg = "failed to get 'diff {}'"
-            args = " ".join(
-                compact([self.args.target, self.args.a_ref, self.args.b_ref])
-            )
-            msg = msg.format(args)
-            logger.exception(msg)
+            logger.exception("failed to get 'diff {}'")
```
efiop: Incomplete message I suppose?
```
            logger.exception("failed to get 'diff {}'")
```

dvc/repo/diff.py
```diff
@@ -1,246 +1,56 @@
...
-        ]
-    elif target in diff_outs:
-        result[DIFF_LIST] = [_diff_royal(self, target, diff_outs[target])]
-    else:
-        msg = "Have not found file/directory '{}' in the commits"
-        raise FileNotInCommitError(msg.format(target))
-    return result
+    return {
+        "old": [entry._asdict() for entry in old],
+        "new": [entry._asdict() for entry in new],
```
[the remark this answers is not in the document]
mroutis: `Diffable` is because I'm using `set()` operations on them and dictionaries are not _hashable_.

tests/func/test_diff.py
mroutis: It is, but it would be great to know how should `dvc diff` should handle `NoSCM` cases.

tests/func/test_diff.py
```diff
@@ -1,267 +1,191 @@
...
+
+    assert dvc.diff("HEAD~2", "HEAD~1") == {
+        "added": [],
+        "deleted": [],
+        "modified": [
+            {"filename": "file", "checksum": {"old": HEAD_2, "new": HEAD_1}}
+        ],
+    }
+
+    pytest.skip('TODO: test dvc.diff("missing")')
```
mroutis: oh, I need to implement this error: `diff: unknown revision 'missing'.`

tests/func/test_diff.py
```diff
@@ -1,267 +1,191 @@
...
+
+    (tmp_dir / "dir" / "2").unlink()
+    dvc.add("dir")
+    scm.add("dir.dvc")
+    scm.commit("delete a file")
+
+    assert dvc.diff(":/init", ":/directory") == {
+        "added": [
+            {
+                "filename": "dir/",
```
mroutis: I'll change this...

dvc/repo/diff.py
```diff
@@ -1,246 +1,70 @@
...
-        self, a_entries, b_entries
-    )
-    result[DIFF_SIZE] = diff_size
-    result[DIFF_DEL], diff_size = _deleted_files(self, a_entries, b_entries)
-    result[DIFF_SIZE] += diff_size
-    result[DIFF_NEW], diff_size = _new_files(self, a_entries, b_entries)
-    result[DIFF_SIZE] += diff_size
-    return result
 
+def diffables_from_output(output):
```
mroutis: I will add an explanation about why is it doing what is doing :sweat_smile:

mroutis:
@efiop , @Suor , false alarm: sorry for requiring the review, I still need to implement error checking :see_no_evil:

tests/func/test_diff.py
```diff
@@ -1,267 +1,258 @@
...
-    def test(self):
-        ret = main(
-            [
-                "diff",
-                self.test_dct[diff.DIFF_A_REF],
-                self.test_dct[diff.DIFF_B_REF],
-            ]
-        )
-        self.assertEqual(ret, 0)
+def test_no_cache_entry(tmp_dir, scm, dvc):
```
mroutis: Not sure if I'm testing this one correctly.

dvc/scm/git/__init__.py
```diff
@@ -363,7 +363,12 @@ def get_rev(self):
         return self.repo.git.rev_parse("HEAD")
 
     def resolve_rev(self, rev):
-        return self.repo.git.rev_parse(rev)
+        from git.exc import GitCommandError
+
+        try:
+            return self.repo.git.rev_parse(rev)
+        except GitCommandError as exc:
+            raise RevError(url=self.root_dir, rev=rev) from exc
```
mroutis: It would be awesome if we could tweak `RevError` in this PR to pass a custom message instead of `url, rev`, this will make the error message less verbose, since there's no need for a `url`

mroutis: Not sure if this is out of scope, @efiop ?

dvc/repo/diff.py
```diff
@@ -1,246 +1,97 @@
...
+    added = new - old
+    deleted = old - new
+    delta = old ^ new
+
+    result = {
+        "added": [entry._asdict() for entry in sorted(added)],
+        "deleted": [entry._asdict() for entry in sorted(deleted)],
+        "modified": [],
+    }
+
```
mroutis: I think it is better to sort the result on `CmdDiff.format`.

dvc/repo/diff.py
```diff
@@ -1,246 +1,97 @@
...
+    old = outs[a_ref]
+    new = outs[b_ref or "working tree"]
+
+    added = new - old
+    deleted = old - new
+    delta = old ^ new
+
+    result = {
+        "added": [entry._asdict() for entry in sorted(added)],
```
mroutis: Converting the entry to dict to keep a consistent output. However, I could return Diffable entries

dvc/command/diff.py
```diff
@@ -15,121 +13,87 @@
...
+            "deleted": colorama.Fore.RED,
+        }
+
+        groups = []
+
+        for state, entries in diff.items():
+            if not entries:
+                continue
+
+            entries = [tuple(entry.values()) for entry in entries]
```
mroutis: converting them to tuples so I can unpack  them later as (filename, checksum)

tests/func/test_diff.py
```diff
@@ -1,267 +1,274 @@
...
+    tmp_dir.dvc_gen({"dir": {"1": "1", "2": "2"}}, commit="add a directory")
+    tmp_dir.dvc_gen({"dir": {"3": "3"}}, commit="add a file")
+    tmp_dir.dvc_gen({"dir": {"2": "two"}}, commit="modify a file")
+
+    (tmp_dir / "dir" / "2").unlink()
+    dvc.add("dir")
+    scm.add("dir.dvc")
+    scm.commit("delete a file")
+
+    assert dvc.diff(":/init", ":/directory") == {
```
efiop: Hm, what is ":/init" and ":/directory"? Esp confused about ":/"

tests/func/test_diff.py
```diff
@@ -1,267 +1,274 @@
...
+            }
         )
-        msg3 = "+{0} with md5 {1}".format(
-            self.test_dct[diff.DIFF_LIST][0][diff.DIFF_TARGET],
-            self.test_dct[diff.DIFF_LIST][0][diff.DIFF_NEW_CHECKSUM],
+        in capsys.readouterr().out
+    )
+
+
+def test_cli(tmp_dir, scm, dvc, mocker, capsys):
```
efiop: Let's move this to `tests/unit/command/test_diff.py`, same as we do in other similar commands.

efiop: Btw, we could test `_format` specifically, just move it into an out-of-class method. The CmdDiff could be tested same way we do it in `tests/unit/command/test_get.py`.

dvc/command/diff.py
```diff
@@ -157,14 +135,15 @@ def add_parser(subparsers, parent_parser):
         ),
     )
     diff_parser.add_argument(
-        "a_ref", help="Git reference from which diff calculates"
+        "--json",
```
efiop: How about `--show-json`, same as in `dvc metrics diff`?

dvc/command/diff.py
```diff
@@ -157,14 +135,15 @@ def add_parser(subparsers, parent_parser):
...
+        help=("Format the output into a JSON"),
```
efiop:
```suggestion
        help="Format the output into a JSON",
```

dvc/command/diff.py
```diff
@@ -157,14 +135,15 @@ def add_parser(subparsers, parent_parser):
...
     )
     diff_parser.add_argument(
-        "b_ref",
-        help=(
-            "Git reference until which diff calculates, if omitted "
-            "diff shows the difference between current HEAD and a_ref"
-        ),
-        nargs="?",
+        "--checksums",
+        help=("Display checksums for each entry"),
```
efiop:
```suggestion
        help="Display checksums for each entry",
```

dvc/command/diff.py
```diff
@@ -15,121 +13,87 @@
...
+            diff = self.repo.diff(
+                self.args.a_ref, self.args.b_ref, target=self.args.target
             )
-            self._show(msg)
+
+            if not any(diff.values()):
+                return 0
+
+            if self.args.json:
+                print(json.dumps(diff))
```
efiop: We use `logger.info` in other places (e.g. `pipeline show`, `metrics diff`, etc), so let's keep it consistent for now. Later we should move to `print` everywhere at once.

tests/func/test_diff.py
mroutis: Those are git references, it searches by commit message:

https://git-scm.com/docs/git-rev-parse#Documentation/git-rev-parse.txt-emlttextgtemegemfixnastybugem

dvc/command/diff.py
mroutis: It is implicit that you are going to _show_ an output, the option is a format modifier, that's why I only named it `--json`.

However, I'm OK with changing it to `--show-json`.

What do you think, @efiop ?

tests/func/test_diff.py
mroutis: I will add a comment on the test.

mroutis: I use them because I think they are more descriptive than the "HEAD~n" syntax

tests/func/test_diff.py
mroutis: neat, @efiop :ok_hand:

dvc/scm/git/__init__.py
efiop: @mroutis Could you please elaborate? But yeah, maybe let's move it into a separate ticket.

tests/func/test_diff.py
efiop: Wow, I didn't know that! Neat!

dvc/command/diff.py
efiop: @mroutis In `dvc metrics diff` my logic was that I didn't want it to be confused with `-t json` and stuff like that. `--show-json` is more explicit, and, well, let's keep it consistent between the commands. So I would use `--show-json` for now.

mroutis: All righty)

mroutis:
@dmpetrov, speaking of https://github.com/iterative/dvc/pull/3229#pullrequestreview-348892487

**Test Case 1 and 2**. I will change the `RevParse` implementation :sweat_smile:  (cc: @efiop  - https://github.com/iterative/dvc/pull/3229#discussion_r371404231)

**Test Case 3.** What you are asking is already provided by running `dvc status`, implementing this on diff would be possible but I prefer to keep the discussion on a separate issue.

**Test Case 4.** I would like to see a more complete version of `--target` that matches the filename by prefixes instead of an exact match with the output path (and maybe rename it to `--prefix`).

**Test Case 5**. Git always shows the full diff, no matter where you are, changing the current implementation to make it  work with the current directory instead of the `root_dir` will also change that compatibility. Also `dvc status` shows the full status, no matter the directory, requiring more changes.

mroutis:
@dmpetrov , **Test Case 6** is already addressed.

**Test Case 1 and 2** is could be solved with the following patch:
```diff
diff --git a/dvc/scm/base.py b/dvc/scm/base.py
index e35c7b7d..bbad771a 100644
--- a/dvc/scm/base.py
+++ b/dvc/scm/base.py
@@ -27,10 +27,7 @@ class CloneError(SCMError):
 
 
 class RevError(SCMError):
-    def __init__(self, url, rev):
-        super().__init__(
-            "failed to access revision '{}' for repo '{}'".format(rev, url)
-        )
+    pass
 
 
 class Base(object):
diff --git a/dvc/scm/git/__init__.py b/dvc/scm/git/__init__.py
index 8efb2dc5..75392a32 100644
--- a/dvc/scm/git/__init__.py
+++ b/dvc/scm/git/__init__.py
@@ -94,7 +94,10 @@ class Git(Base):
             try:
                 repo.checkout(rev)
             except git.exc.GitCommandError as exc:
-                raise RevError(url, rev) from exc
+                raise RevError(
+                    "failed to access revision '{}' for repo '{}'"
+                    .format(rev, url)
+                ) from exc
 
         return repo
 
@@ -367,8 +370,8 @@ class Git(Base):
 
         try:
             return self.repo.git.rev_parse(rev)
-        except GitCommandError as exc:
-            raise RevError(url=self.root_dir, rev=rev) from exc
+        except GitCommandError:
+            raise RevError("unknown Git revision '{}'".format(rev))
 
     def close(self):
         self.repo.close()
```

Would be great if I can include it on this PR (cc: @efiop )

Please create issues to discuss the rest of the Test cases, @dmpetrov :slightly_smiling_face:

dmpetrov:
> Please create issues to discuss the rest of the Test cases, @dmpetrov 🙂

@mroutis So, 6 is solved, 1 and  2 will be in the patch (from you or @efiop ), 3 - a separate issue will be created as we discussed (I can create it once the PR closed), 5 - it's might be ok to keep as is to be in parity with `git diff/git status/dvc status`. Only 4 left - `--target`.

If you are not planning to make a solution now and rename it to `--prefix`, I'd suggest removing this or at least hide this option. To be honest, I don't see good use cases for this option.

mroutis:
@dmpetrov , all righty, I'll remove `--target`

mroutis:
@dmpetrov , 1 and 2 is already on this PR, I've removed the `--target`.

What would be the next steps, @dmpetrov , do you think this PR is ready to be closed?

dvc/command/diff.py
```diff
@@ -15,121 +13,84 @@
...
-        elif dct.get(diff.DIFF_OLD_FILE):
-            msg = "deleted file with size {}".format(
-                humanize.naturalsize(abs(dct[diff.DIFF_SIZE]))
-            )
-        else:
-            msg = "file was modified, file size {}".format(
-                cls._print_size(dct[diff.DIFF_SIZE])
-            )
-        return msg
+            for state, entries in diff.items()
```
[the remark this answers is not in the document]
mroutis: that's why the py3.5 test suite was failing! :see_no_evil:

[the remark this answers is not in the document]
mroutis: now iterating with an explicit order

dvc/repo/diff.py
```diff
@@ -1,246 +1,94 @@
...
+            diffable
+            for stage in self.stages
+            for out in stage.outs
+            for diffable in _diffables_from_output(out)
+        )
+
+    old = outs[a_ref]
+    new = outs[b_ref or "working tree"]
+
+    added = new - old
```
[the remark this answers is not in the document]
mroutis: this was mind-bending :ok_hand:  thanks for pointing it out, @Suor !
just added it)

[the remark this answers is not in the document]
mroutis: however, instead of `repo.iter_outs`, I defined a method to return the dictionary with paths and checksums, might refactor that one later

efiop:
@mroutis please check the tests, looks like there are some dict ordering issues.

mroutis:
@efiop , doing it right now

jorgeorpinel:
What's the fastest way to see if this is released or get notified when it is? I don't think it's included in `0.82.6`, for example. Thanks

PR #3234, opened 2020-01-24
erepo: make it better

Suor:
- code moving during git pull won't affect any erepos
- source tag move will work as expected
- source cache dir setting will never break havoc
- cache is reused between all clones from same source

### DIFF
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -41,14 +41,20 @@ def external_repo(url, rev=None):
         raise PathMissingError(exc.path, url)
     finally:
         repo.close()
+        _remove(path)
+
+
+CLONES = {}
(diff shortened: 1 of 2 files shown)


dvc/external_repo.py
```diff
@@ -191,19 +204,29 @@ def _git_checkout(repo_path, rev):
...
-    git = git.Repo(repo_path)
     try:
-        msg = git.git.pull()
+        git.Repo(repo_path).commit(rev)
+        return True
+    except BadName:
+        return False
+
+
+def _git_pull(repo_path, rev):
```
efiop: DS reports unused argument here.

Suor: Fixed it.

Issue #2719, opened 2019-11-03
api: high level dvc.api import/export meets data catalog


Suor:
<p align="right"><i>
- what do you do for a living?<br>
- oh, different things, import, export, family business <br> </i>
(any Hollywood smuggler and now a dvc dev)
</p>

This feature was first proposed by @dmpetrov and was discussed several times here and there. Recent conversations/debates with @dmpetrov and @shcheklein resulted into some more or less defined approach to it, which I will lay out here.

**The need:** we want things to be easier reused across repos, so we want some higher level api than just `get()/read()/open()`, presumably some call, which will return a ready to use dataframe, an instantiated model, etc.

## The design

So here goes the design so far. The exporting side will need to list available things and write some instantiating glue code. Like:
```yaml
# dvc-artifacts.yml
artifacts:
    - name: face_rec_model:
      description: Tensorflow face recognition model 
      call: export.make_face_rec
      deps: ["data/face_rec_model.dat", "data/extra.dat"]
```
```python
# export.py
import tflearn

def make_face_rec(face_rec_model_data, extra_data):
    # ... construct a model
    return model
```
Then the importting side might simply do:
```python
import dvc.api

model = dvc.api.summon("face_rec_model", "https://path/to/repo", rev=...)
model.predict(...)
```
Summon is a good (to my liking) and short name that refers to summoning a genie or a tesla :)

This will provide a flexible base to export/summon anything. Some questions arise about requrements, repeatetive glue code, Python/R interoperability and easing things for a publisher. I will go through this.

## Requirements

Instantiating a model or a dataframe will require some libraries installed like pandas or PyTorch, or even their particular versions. In basic scenario we might ignore that - in most cases a user will know what is this about an will have it installed already. The thing is we can't provide it seamlessly since installing python libs inside `dvc.api.summon()` call will be surprising and might break some things. So deps management will be separate anyway. The possible way:

```yaml
artifacts:
    - name: some_dataframe
      ...
      reqs: 
        - pandas>=1.2.3,<2
        - otherlib>=1.2
```

```console
$ dvc summon-reqs <repo-url> <name> 
pandas>=1.2.3,<2
otherlib>=1.2
$ dvc summon-reqs <repo-url> <name> | xargs pip install
```
 

## Glue code

There are many common scenarios like export csv-file as dataframe, which will produce repetitive glue code. This could be handled by providing common export functions in our `dvcx` (for dvc export) library: 

```yaml
artifacts:
    - name: some_dataframe
[...]

PR #3238, opened 2020-01-26
Summon - next iteration

dmpetrov:
An iteration on summoning (see https://github.com/iterative/dvcx/pull/4):
1. Some minor Summon file format changes and default file name change to `Summon.yaml`.
2. Separating reading objects and pulling data. It is needed for flexibility and especially for data objects with several files like a compound CSV file (see in #2719).
3. Summon object update `update_dobj()`. This is not completed but it does not break current read-only logic.
4. Moving part of the logic to `SummonDesc` class.

### DIFF
--- a/dvc/api.py
+++ b/dvc/api.py
@@ -13,6 +13,7 @@ from dvc.repo import Repo
 from dvc.exceptions import DvcException, NotDvcRepoError
 from dvc.external_repo import external_repo
 
+DEF_SUMMON = "Summon.yaml"
 
 SUMMON_FILE_SCHEMA = Schema(
     {
@@ -40,6 +41,10 @@ SUMMON_PYTHON_SCHEMA = Schema(
 
 
 class SummonError(DvcException):
+    pass
+
+
+class SummonErrorNoObjectFound(SummonError):
     pass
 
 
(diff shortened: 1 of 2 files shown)


dvc/api.py
```diff
@@ -13,12 +13,15 @@
 from dvc.exceptions import DvcException, NotDvcRepoError
 from dvc.external_repo import external_repo
 
+DEF_SUMMON = "Summon.yaml"
```
shcheklein: minor: make it lower case? :) the only example is Makefile? and obviously, not the best example ...

dmpetrov: Agree. But then we need to consider renaming default repro goal - `Dvcfile`. Or accept two default options `Dvcfile` and `dvcfile`.

Suor: @dmpetrov it's `Dockerfile`, but `docker-compose.yml`. Basically everything is lowercase if it doesn't resemble Makefiles. Also, why not `dvcsummon.yaml`?

efiop: I agree with @Suor , plus we have `.dvcignore` already as an example of dvc-specific file :slightly_smiling_face:

dvc/api.py
```diff
@@ -13,12 +13,15 @@
...
+DOBJ_SECTION = "d-objects"
```
Suor: What does it stand for? Also, we was trying to go away from string constants) This makes it more readable.

dvc/api.py
```diff
@@ -120,94 +127,134 @@ def _make_repo(repo_url=None, rev=None):
...
 
         return objects[0]
 
-    except FileNotFoundError as exc:
-        raise SummonError("Summon file not found") from exc
-    except ruamel.yaml.YAMLError as exc:
-        raise SummonError("Failed to parse summon file") from exc
-    except Invalid as exc:
-        raise SummonError(str(exc)) from exc
+    def update_dobj(self, new_dobj, overwrite=False):
```
Suor: "update" verb presumes override, imho. Should this be `.set_dobject()`?

dvc/api.py
```diff
@@ -120,94 +127,134 @@ def _make_repo(repo_url=None, rev=None):
...
+        try:
+            name = new_dobj["name"]
+            dobj = self.get_dobject(name)
+
+            if overwrite:
+                idx = self.content[DOBJ_SECTION].index(dobj)
+                self.content[DOBJ_SECTION][idx] = new_dobj
```
Suor: Maybe we should use dict with name as key instead of a list.

dvc/api.py
```diff
@@ -120,94 +127,134 @@ def _make_repo(repo_url=None, rev=None):
...
+        desc.pull(dobj)
         _args = {**summon_dict.get("args", {}), **(args or {})}
         return _invoke_method(summon_dict["call"], _args, desc.repo.root_dir)
 
 
-@contextmanager
-def prepare_summon(name, repo=None, rev=None, summon_file="dvcsummon.yaml"):
-    """Does a couple of things every summon needs as a prerequisite:
-    clones the repo, parses the summon file and pulls the deps.
+class SummonDesc(object):
```
Suor: So it's now `SummonFile`. Why do you want to separate it anyway? Will we use single file to work with several objects?

dvc/api.py
dmpetrov: `dvc-objects` or `data-objects`. I'm making it `dvc-objects`.

dvc/api.py
dmpetrov: Right. Changing the `overwrite=False` to `True`.

dvc/api.py
dmpetrov: Changed to `SummonFile`.

Sure, in the resnet example we had many resnet models in a single file. `pretrained-models.pytorch` project contains 50+ models. We probably don't what to have 50 summon files (the code is localized in few files only for all the models).

dvc/api.py
dmpetrov: Let's keep `dvcsummon.yaml`

dvc/api.py
dmpetrov: Gush... great idea.

dmpetrov:
@Suor should we close this one?

PR #3239, opened 2020-01-26
Set corpora and driveId params values based on type of remote disk - shared or user's.

MaxRis:
PR is not yet ready: will require update of PyDrive2 along with this PR.

https://github.com/iterative/PyDrive2/pull/6 should be merged first.

### DIFF
--- a/dvc/remote/gdrive.py
+++ b/dvc/remote/gdrive.py
@@ -104,13 +104,15 @@ class RemoteGDrive(RemoteBASE):
                 ),
             )
         )
+        self.remote_drive_id = None
 
     @gdrive_retry
     def gdrive_upload_file(
         self, args, no_progress_bar=True, from_file="", progress_name=""
     ):
+        parent = {"id": args["parent_id"]}
         item = self.drive.CreateFile(
-            {"title": args["title"], "parents": [{"id": args["parent_id"]}]}
+            {"title": args["title"], "parents": [parent]}
         )
 
         with open(from_file, "rb") as fobj:
@@ -133,7 +135,9 @@ class RemoteGDrive(RemoteBASE):
     def gdrive_download_file(
         self, file_id, to_file, progress_name, no_progress_bar
     ):
-        gdrive_file = self.drive.CreateFile({"id": file_id})
+        param = {"id": file_id}
+        # it does not create a file on the remote
+        gdrive_file = self.drive.CreateFile(param)
         bar_format = (
             "Donwloading {desc:{ncols_desc}.{ncols_desc}}... "
             + Tqdm.format_sizeof(int(gdrive_file["fileSize"]), "B", 1024)
@@ -144,7 +148,12 @@ class RemoteGDrive(RemoteBASE):
             gdrive_file.GetContentFile(to_file)
 
     def gdrive_list_item(self, query):
-        file_list = self.drive.ListFile({"q": query, "maxResults": 1000})
+        param = {"q": query, "maxResults": 1000, "corpora": self.corpora}
+
+        if self.remote_drive_id:
+            param["driveId"] = self.remote_drive_id
+
+        file_list = self.drive.ListFile(param)
 
         # Isolate and decorate fetching of remote drive items in pages
         get_list = gdrive_retry(lambda: next(file_list, None))
@@ -240,21 +249,22 @@ class RemoteGDrive(RemoteBASE):
 
             self._gdrive = GoogleDrive(gauth)
 
+            if self.bucket != "root" and self.bucket != "appDataFolder":
+                self.remote_drive_id = self.get_remote_drive_id(self.bucket)
+            self.corpora = "drive" if self.remote_drive_id else "default"
             self.remote_root_id = self.get_remote_id(
                 self.path_info, create=True
             )
+
             self._cached_dirs, self._cached_ids = self.cache_root_dirs()
 
         return self._gdrive
 
     @gdrive_retry
     def create_remote_dir(self, parent_id, title):
+        parent = {"id": parent_id}
         item = self.drive.CreateFile(
-            {
-                "title": title,
-                "parents": [{"id": parent_id}],
-                "mimeType": FOLDER_MIME_TYPE,
-            }
+            {"title": title, "parents": [parent], "mimeType": FOLDER_MIME_TYPE}
         )
         item.Upload()
         return item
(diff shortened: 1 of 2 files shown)


dvc/remote/gdrive.py
```diff
@@ -109,8 +109,11 @@ def init_drive(self):
     def gdrive_upload_file(
         self, args, no_progress_bar=True, from_file="", progress_name=""
     ):
+        param = {"supportsAllDrives": True}
```
shcheklein: I think we can keep `support*Drivers` family of params inside? They are deprecated anyway and will be enabled by default soon? Also were they present even before we forked the PyDrive?

dvc/remote/gdrive.py
```diff
@@ -272,10 +282,16 @@ def get_remote_item(self, name, parents_ids):
 
         query += " and trashed=false and title='{}'".format(name)
 
+        param = {
+            "q": query,
+            "maxResults": 1,
```
shcheklein: why do we limit it here? I would put an assert if it returns >1 items in the result

dvc/remote/gdrive.py
MaxRis: In original PyDrive from pip some support of related params was added ( deprecated `supportsTeamDrives` was used ), but some bugs were fixed in PyDrive's master and not released to pip.

> I think we can keep support*Drivers family of params inside? 
Sounds good.

dvc/remote/gdrive.py
MaxRis: it's related to the issue regarding duplicated folders with the same title.

MaxRis: it is optimization to request and process only first value from the results set, but in case of duplicated folders issue it might return more than 1 result

shcheklein: Makes sense! Resolving

MaxRis:
@shcheklein 

- `support*Drivers` params are moved back into PyDrive2
- Shared `driveId` is obtained from the remote root directory properties (turned out that GDrive provides property `driveId` for items at shared drives only).
- If `driveId` is not empty corpora is set to `drive` otherwise `default` (should be analog to `user` in GDrive API v3). Since `allDrives` corpora is not used, hopefully, `incompleteSearch` will not occur.

shcheklein:
@MaxRis do we still need to update PyDrive to remove corpora/corpus from it?

also, just to double check - I hope that all these old `support*Drivers` (there are two sets of those btw, and we use older versions) do not contradict with `corpora`.

shcheklein:
@MaxRis tests are failing on Travis

MaxRis:
@shcheklein yes, https://github.com/iterative/PyDrive2/pull/6 should be merged into PyDrive2. Then we need to release new version of PyDrive2 and update this PR to pin newer version to ensure that Travis tests are passing.

MaxRis:
> 
> 
> @MaxRis do we still need to update PyDrive to remove corpora/corpus from it?
> 
> also, just to double check - I hope that all these old `support*Drivers` (there are two sets of those btw, and we use older versions) do not contradict with `corpora`.

right, https://github.com/iterative/PyDrive2/pull/6 does exactly above - removes `corpus` and updates older version of params to newer.

dvc/remote/gdrive.py
```diff
@@ -272,12 +282,27 @@ def get_remote_item(self, name, parents_ids):
...
-        item_list = self.drive.ListFile(
-            {"q": query, "maxResults": 1}
-        ).GetList()
+        item_list = self.drive.ListFile(param).GetList()
         return next(iter(item_list), None)
 
+    @gdrive_retry
+    def get_remote_drive_id(self, remote_id):
+        param = {"id": remote_id}
+        item = self.drive.CreateFile(param)
```
shcheklein: can we use something less confusing - GetFile? Something that would not create file if it's missing for some reason (wrong remote location)? I would really try to avoid side effects here

shcheklein:
@MaxRis 

> right, iterative/PyDrive2#6 does exactly above - removes corpus and updates older version of params to newer.

good, I've merged it. We will need to release and update the version in the setup.py and conda.

shcheklein:
@MaxRis 1.4.2 is released, let's update deps please (and don't forget to update conda when we ready to merge)

dvc/remote/gdrive.py
```diff
@@ -104,13 +104,16 @@ def init_drive(self):
                 ),
             )
         )
+        self.corpora = "allDrives"
+        self.remote_drive_id = ""
```
shcheklein: would None do here?

dvc/remote/gdrive.py
```diff
@@ -104,13 +104,16 @@ def init_drive(self):
                 ),
             )
         )
+        self.corpora = "allDrives"
```
shcheklein: do we need this default value? if not it would be safer to keep them None, to make sure that we don't hit `allDrive` after some changes in the future.

dvc/remote/gdrive.py
MaxRis: `PyDrive.file.CreateFile` doesn't create remote file unless `Upload` is called (not called in this case). We might need to rename that PyDrive's API to avoid such confusion.

shcheklein: yeah, I remember that it's not the first time I ask :) could you small comment around please?

dvc/remote/gdrive.py
MaxRis: yep, we need `allDrives` initially, since we don't know initially where the root directory is located - in shared or user's drive "collection".

dvc/remote/gdrive.py
shcheklein: could you also update that comment a bit to be more explicit - like "Because of ... it's possible and allowed for remote to have two files with the same name">

dvc/remote/gdrive.py
shcheklein: do we hit any APIs that require `corpora` though before we able to figure out this?

MaxRis: well, it looks like it might occur. I will look into possibility to put the assert into PyDrive2 to catch this.

shcheklein: not into PyDrive, probably 🤔  ... I meant that I still don't quite understand why do we need this default ... it's very close to the beginning where we get the actual value for it - default or drive, right? so why do we need this default? is it needed for the period we are still not sure is a shared drive or not? do we hit any APIs that require this value?

MaxRis: @shcheklein sorry, I meant, that, yes, currently ListFile API is called. But it is possible to avoid this, will update the code.

dvc/remote/gdrive.py
```diff
@@ -272,12 +283,28 @@ def get_remote_item(self, name, parents_ids):
...
-        item_list = self.drive.ListFile(
-            {"q": query, "maxResults": 1}
-        ).GetList()
+        item_list = self.drive.ListFile(param).GetList()
         return next(iter(item_list), None)
 
+    @gdrive_retry
+    def get_remote_drive_id(self, remote_id):
+        param = {"id": remote_id}
+        # drive.CreateFile creates only local GoogleDriveFile object
```
shcheklein: ...
`# it does not create a file on the remote`

dvc/remote/gdrive.py
shcheklein: gotcha, yep, let's avoid this if possible and remove the default value

MaxRis:
@shcheklein usage of `allDrives` is removed, but it seems I have found the `pull` issues if the URL consists only from single folder ID, like `gdrive://18HB_cxbipUlERdQK-HNT8EbGiopAMjjY`
Going to check why it happens

efiop:
@MaxRis Is this still WIP? :slightly_smiling_face:

PR #3244, opened 2020-01-27
diff/metics diff: update argument descriptions, et al.

jorgeorpinel:
to match https://github.com/iterative/dvc.org/pull/933


dvc/command/diff.py
```diff
@@ -157,13 +157,13 @@ def add_parser(subparsers, parent_parser):
...
+        "a_ref", help="Git reference from which the diff begins"
     )
     diff_parser.add_argument(
         "b_ref",
         help=(
-            "Git reference until which diff calculates, if omitted "
-            "diff shows the difference between current HEAD and a_ref"
+            "Git reference until which the diff ends. "
+            "If omitted, `HEAD` (latest commit) is used."
         ),
```
jorgeorpinel: @iterative/engineering should we call these "commits" instead of "references"? See https://git-scm.com/docs/git-diff for example. Note that a `ref` in Git is something else: https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefrefaref – a bit more technical.

Note that per https://github.com/iterative/dvc.org/pull/933#pullrequestreview-348855380, in docs we're going to start using the term "revision" (link to https://git-scm.com/docs/revisions) which according to [its glossary entry](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefrevisionarevision) is a "synonym of commit (the noun)" so you we could just say "Git revision" here instead, so we use that term consistently everywhere. (never "reference").

Thoughts?

jorgeorpinel: I went for this criteria in iterative/dvc.org/pull/933: Use "version" for data or DVC-tracked files, "revision" (and that link) in the context of pure Git/ source code, and "reference" sparingly when actually talking about Git references (to revisions/commits). Updating this repo accordingly...

p.s. We're also reviewing the use of "commit hash" and "SHA" terms...

setup.py
```diff
@@ -133,7 +133,7 @@ def run(self):
     description="Git for data scientists - manage your code and data together",
     long_description=open("README.rst", "r").read(),
     author="Dmitry Petrov",
-    author_email="dmitry@dataversioncontrol.com",
+    author_email="dmitry@iterative.ai",
```
jorgeorpinel: @dmpetrov please note this change. Is it correct?

dvc/command/diff.py
jorgeorpinel: UPDATE: Honestly, I think it would be nicer to only use one single term in the code base. In this case "revision" seems the most appropriate, so these params could be `a_rev` and `b_rev`. I would vote for "revision" because we already use it extensively for the `dvc get` and `dvc import` commands, in the form of the `--rev` option as well as `rev` and `rev_lock` fields of import stage DVC-files.

> Another term that may need review is "checksum", per https://github.com/iterative/dvc.org/pull/933#pullrequestreview-350538870.

shcheklein: that's the best way to explain it https://git-scm.com/docs/git-diff - ` Show changes between commits, commit and working tree, etc` .. just plain commits! I would vote to go with just commits everywhere. It would simplify the whole thing - understanding param names, docs, etc. Of course we can put a link to the ways of specifying commits.

shcheklein: if not commits, let's at least use revisions indeed. To at least have consistency with `dvc get`\`dvc import` and at least use the correct term.

mroutis: @jorgeorpinel , @shcheklein I recently opened an issue about exactly this :sweat_smile:  #3245 

I would prefer to use either revisions or commits. Revisions is more accurate but _commits_ are more user friendly, I guess both newbies and advanced users understand what would be talking about.

mroutis:
@jorgeorpinel , looks like you've change the old _diff_ implementation, I will include your changes in my PR.

dvc/command/diff.py
jorgeorpinel: Commit is accurate. If you examine https://git-scm.com/docs/gitglossary#def_revision, "revision" = "Synonym for commit (the noun)." 🙂 

> I recently opened an issue about exactly this 😅 #3245

Wow! OK, moving this discussion over there and resolving here.

p.s. please someone review this PR when you can 😬

setup.py
shcheklein: yep, it's good. resolving this.

shcheklein: may be it's even better to put dmitry@dvc.org

jorgeorpinel: OK. Updated in a bunch of places in 039f196d.

jorgeorpinel:
> looks like you've change the old diff implementation

@mroutis I changed just some of the help output stings, but I'm merging with master now and (mostly) preserved your text. Please check again 🙂

dvc/remote/base.py
```diff
@@ -191,7 +191,7 @@ def _calculate_checksums(self, file_infos):
                 tasks,
                 total=len(file_infos),
                 unit="md5",
-                desc="Computing hashes (only done once)",
+                desc="Computing file checksum(s) (only done once)",
```
jorgeorpinel: Should probably revert per https://github.com/iterative/dvc.org/pull/933#pullrequestreview-350538870

dvc/command/diff.py
```diff
@@ -134,17 +133,17 @@ def add_parser(subparsers, parent_parser):
...
         nargs="?",
         default="HEAD",
     )
     diff_parser.add_argument(
         "b_ref",
         help=(
-            "Git reference to the new version that you want to compare."
-            " (defaults to the working tree)"
+            "Git reference to the newer version to compare"
+            " (defaults to the current workspace including changes)"
```
mroutis: what do you mean by _including changes_? :thinking:

jorgeorpinel: Just trying to be extra clear that the workspace includes any uncommitted local changes. I guess its confusing, we could just leave "current workspace" which implies the rest.

mroutis: @jorgeorpinel , yep, it was confusing to me, tbh; I appreciate the rest of the changes, tho :+1:

dvc/remote/base.py
jorgeorpinel: OK I reverted this and applied "hash" terminology as much as possible to the whole file, as well as a few others in 5019fe14. If this commit is acceptable, I can update all the other instances (there's less than 70, in 19 files.
Or I can extract further changes on this to a separate issue/PR. Cc @mroutis @efiop

dvc/command/diff.py
jorgeorpinel: Simplified in 6984546a.

dvc/command/commit.py
```diff
@@ -47,7 +47,7 @@ def add_parser(subparsers, parent_parser):
         "--force",
         action="store_true",
         default=False,
-        help="Commit even if checksums for dependencies/outputs changed.",
+        help="Commit even if hash value for dependencies/outputs changed.",
```
mroutis: @jorgeorpinel , missed the conversation when we decided that _hash value_ was better than _checksums_ :sweat_smile:

dvc/command/diff.py
```diff
@@ -134,17 +133,17 @@ def add_parser(subparsers, parent_parser):
     diff_parser.add_argument(
         "a_ref",
         help=(
-            "Git reference to the old version that you want to compare"
-            " (defaults to HEAD)"
+            "Git reference to the older version to compare "
+            "(defaults to `HEAD`)"
```
mroutis: I'd say that there's no need for ` since it is not a command nor a function

dvc/command/commit.py
shcheklein: @jorgeorpinel let's keep it for DVC itself for now. It should be a separate discussion/PR to fix it.

@mroutis technically, checksum is not a right term for us. md5 (sha1, etc) is hash (that can be used to check for integrity - that's what checksums are for). But the most important here that we rely on it being a good _hash_. We need uniqueness. That's why for example HDFS checksums do not work for us for external outputs/inputs.

mroutis: @shcheklein , I agree that _checksum_ is not the best term, indeed) just caught for surprise.

mroutis: it would be great to have an issue for this, tho

dvc/command/diff.py
jorgeorpinel: Good point. Updated

dvc/command/commit.py
jorgeorpinel: > missed the conversation when we decided that hash value was better than checksums

Yeah sorry it happened somewhere in iterative/dvc.org/pull/933 but the discussion is now resolved/hidden since it's such a long PR. Ivan brought up that cheksums have the intention to check the validity of files and that's not the main reason we use them in DVC... Technically the MD5 string is a hash or "hash value" (more general concept).

There's no issue but there's a (draft) PR for this in iterative/dvc.org/pull/962 for now.

shcheklein: I think there is no reason do this ad-hoc in the DVC core repo. We should create a separate issue and deal with it "at-once".

jorgeorpinel: Yes, that's what iterative/dvc.org/pull/962 is for 🙂 but are you saying I should revert the few ones I changed here?

shcheklein: yes, that's the idea to keep PRs focused as much as possible. Make it longer to review and merge when there are many things (especially something that requires discussions) happening at once. I would revert this for now.

dvc/command/diff.py
```diff
@@ -134,17 +133,17 @@ def add_parser(subparsers, parent_parser):
     diff_parser.add_argument(
         "a_ref",
```
efiop: We need to rename `ref` -> `rev` in args too. Can do after this PR is merged.

dvc/version.py
```diff
@@ -11,7 +11,7 @@
 
 
 def _generate_version(base_version):
-    """Generate a version with information about the git repository"""
+    """Generate a version with information about the Git repository."""
```
efiop: To be honest, not a big fan of such changes in the code. It only pollutes the history and doesn't provide any value.

dvc/remote/base.py
```diff
@@ -962,7 +962,7 @@ def checkout(
         skip = False
         if not checksum:
             logger.warning(
-                "No checksum info found for '{}'. " "It won't be created.",
+                "No file hash info found for '{}'. " "It won't be created.",
```
efiop: Again, we call it checksum everywhere else, it is weird that we do this change in this PR.

dvc/command/diff.py
```diff
@@ -156,7 +155,7 @@ def add_parser(subparsers, parent_parser):
     )
     diff_parser.add_argument(
         "--checksums",
```
efiop: Will need to rename `checksum` -> `hash` in the whole code base. Need to do that in a sync way to not introduce more confusion for developers and users.

efiop:
Alright, I'm not happy about the mix of things here, but let's merge and move on. Don't want to waste time discussing this either.

dvc/version.py
jorgeorpinel: I'm used to making copy edits on docs all the time so this is kind of automatic for me but whatever policies you tell me for the core repo I'll abide to the best of my ability. Personally I don't understand the problem of history pollution but I guess it complicates your release management? Anyway, I'll follow your lead on these things of course.

dvc/remote/base.py
jorgeorpinel: Sorry just noticed this and it's been merged with the change at this point. I would've rolled these back if I had the capacity at the time. In any case they have all been changed in a separate PR on top of this one now. Resolving 🙂

jorgeorpinel:
> Please try to not dilute the focus of your PRs

Understood, sorry about that!

dvc/command/diff.py
jorgeorpinel: For the record, this question has been extracted to #3311

PR #3246, opened 2020-01-28
introduce `dvc list` command

### LINKED ISSUE #2509: new command to list data artifacts in a DVC project
Especially useful for "browsing" external DVC projects on Git hosting before using `dvc get` or `dvc import`. Looking at the Git repo doesn't show the artifacts because they're only referenced in DVC-files (which can be found anywhere), not tracked by Git.

Perhaps `dvc list` or `dvc artifacts`? (Or/and both `dvc get list` and `dvc import list`)

> As mentioned in https://github.com/iterative/dvc.org/pull/611#discussion_r324998285 and other discussions.

---

UPDATE: Proposed spec (from https://github.com/iterative/dvc/issues/2509#issuecomment-533019513):

```
usage: dvc list [-h] [-q | -v] [--recursive [LEVEL]] [--rev REV | --versions]
                url [target [target ...]]

positional arguments:
  url         URL of Git repository with DVC project to download from.
  target      Paths to DVC-files or directories within the repository to list outputs
              for.
```

UPDATE: Don't forget to update docs AND tab completion scripts when this is implemented.

JIoJIaJIu:
Provides `dvc list` command
```
usage: dvc list [-h] [-q | -v] [-R] [--rev [REV]] url [targets [targets ...]]

List DVC outputs in the repo

positional arguments:
  url              Supported urls:
...
  targets          Paths to directories within the repository to list outputs for.

optional arguments:
  -h, --help       show this help message and exit
  -q, --quiet      Be quiet.
  -v, --verbose    Be verbose.
  -R, --recursive  Recursively list files under directory targets.
  --rev [REV]      Git revision (e.g. branch, tag, SHA
```
~Uses [sparse checkout](http://schacon.github.io/git/git-read-tree.html#_sparse_checkout) for remote repos, but missed opportunity to list artifacts in repo by SHA commit (only by refs). Will be fixed in next commits
As an alternative I looked to the [partial-clone](https://git-scm.com/docs/partial-clone) but have chosen sparse checkout as the simplest solution, open for suggestions
**targets** params filter artifacts/outputs and not dvc-file locations~

Close #2509
<hr/>

* [x] Sufficient test coverage is provided

* [x] Support SHA commits as rev

* [x] There are shell completion scripts


dvc/repo/list_outs.py
```diff
@@ -0,0 +1,63 @@
...
+            matched = set()
+            for out in outs:
+                if func(out, path_info, is_dir):
+                    matched.add(out)
+                    matched_outs.append(out)
+            outs.difference_update(matched)
+    else:
+        matched_outs = outs
+
+    logger.info("\n".join(map(str, matched_outs)))
```
efiop: Let's separate the listing and representing parts. API should list and return something that will then be processed and rendered properly in CLI.

dvc/cli.py
```diff
@@ -60,6 +61,7 @@
     metrics,
     install,
     root,
+    listCmd,
```
efiop: A bit weird that it is `listCmd` here, but `list_outs.py` in `dvc/repo/`. Let's use something in particular. Thinking about `ls` again now, but not sure :slightly_smiling_face:

dvc/external_repo.py
```diff
@@ -187,6 +187,35 @@ def _cached_clone(url, rev):
...
+        with open(sc_path, "w") as sc_file:
+            sc_file.write("**/*.dvc\n")
+    git_repo.create_remote("origin", url)
+    try:
+        ref = rev or "master"  # TODO
+        git_repo.remote().pull(ref, **{"depth": 1})
+    except git.exc.GitCommandError as e:
+        logger.error("Unable to pull repo {} with ref {}", url, ref)
+        raise e
+    git_repo.close()
```
efiop: Probably should be in a `finally` block :slightly_smiling_face: Need to be careful with such things, as they might cause lots of troubles, especially on Windows.

dvc/external_repo.py
```diff
@@ -187,6 +187,35 @@ def _cached_clone(url, rev):
...
+    repo_path = tempfile.mkdtemp("dvc-sparse-erepo")
+    git_repo = git.Repo.init(repo_path)
+    with git_repo.config_writer() as cw:
+        cw.set("core", "sparseCheckout", True)
+        sc_path = os.path.join(git_repo.git_dir, "info", "sparse-checkout")
+        with open(sc_path, "w") as sc_file:
+            sc_file.write("**/*.dvc\n")
+    git_repo.create_remote("origin", url)
+    try:
+        ref = rev or "master"  # TODO
```
efiop: `master` is not a universal default branch, it might be anything.

dvc/repo/list_outs.py
```diff
@@ -0,0 +1,63 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@staticmethod
+def list_outs(url, targets=[], rev=None, recursive=None):
+    if os.path.exists(url):
+        path = os.path.abspath(url)
+        return _list_outs_local_repo(path, targets, recursive)
+    else:
```
efiop: This `else` is not necessary.

dvc/repo/list_outs.py
```diff
@@ -0,0 +1,63 @@
...
+
+    outs = set()
+    for stage in repo.stages:
+        outs = outs.union(stage.outs)
+
+    matched_outs = []
+    if len(targets):
+
+        def func(out, path_info, is_dir):
+            if out.scheme == "local" and out.path_info.__eq__(path_info):
```
efiop: Hm, why do we need to use `__eq__` directly instead of `==`?

tests/unit/command/test_list.py
```diff
@@ -0,0 +1,14 @@
+from dvc.cli import parse_args
+from dvc.command.list import CmdList
+
+
+def test_list(mocker):
+    cli_args = parse_args(["list", "repo_url"])
```
efiop: Need to test other flags as well.

tests/unit/command/test_list.py
```diff
@@ -0,0 +1,14 @@
+from dvc.cli import parse_args
+from dvc.command.list import CmdList
+
+
+def test_list(mocker):
```
efiop: Also need a functional test for `Repo.ls` API. Please be sure to add it to `tests/func/test_ls.py`

dvc/repo/list_outs.py
```diff
@@ -0,0 +1,63 @@
...
+                    return True
+                elif out.path_info.isin(path_info, 1):
+                    return True
+
+            return False
+
+        for path in targets:
+            abs_path = os.path.abspath(path)
+            path_info = PathInfo(abs_path)
+            is_dir = os.path.isdir(abs_path)
```
efiop: You could `dvc add dir/`, so it might not exist here(because there was no `dvc checkout` in that cloned repo yet). We need to handle that case as well.

dvc/repo/list_outs.py
```diff
@@ -0,0 +1,63 @@
...
+
+def _list_outs_local_repo(path, targets=[], recursive=None):
+    from dvc.repo import Repo
+    from dvc.path_info import PathInfo
+
+    repo = Repo(path)
+
+    outs = set()
+    for stage in repo.stages:
+        outs = outs.union(stage.outs)
```
efiop: Looks like we don't currently support git files. We should list them too along with dvc-outputs.

jorgeorpinel:
Hi. Will, `--show-url` from `dvc get` be moved to this new command in this introductory PR, or will that be left for a future review? Thanks!

dvc/cli.py
JIoJIaJIu: Yes, that just happened
I'm going to rename `listCmd` to `list_outs`

Maybe should I use it as `command.list`? But it becomes inconsistency

> could use dvc ls and use ls

+1

JIoJIaJIu:
Hi @jorgeorpinel 
Do you expect something like
```
dvc list  --show-url https://github.com/iterative/dataset-registry
tutorial/ver/new-labels.zip     https://remote.dvc.org/dataset-registry/2e/aa473159443e75e6fb7b29e56c0787
tutorial/nlp/pipeline.zip       https://remote.dvc.org/dataset-registry/1d/2070ee188fc5e4d94ad920e6cc82aa
get-started/data.xml            https://remote.dvc.org/dataset-registry/a3/04afb96060aad90176268345e1035
...
```
If it needs I can add `--show-url` to list command in this PR

dvc/repo/list_outs.py
JIoJIaJIu: So you expect that `dvc list` list regular files tracked by git?
```
dvc list git@github.com:repo
.gitignore
data/data.xml
```
If I got right then I think it's not a right way cause
1. It's not clear user usage (motives) of the command. Right now we are listing artifacts for getting them, for working with regular files there are `git` commands
2. Needs to differentiate somehow dvc artifacts from regular files, and add complexity for pipelining commands
    ```
    git list git@github.com:repo
    .gitignore
    outs:data/data.xml
    ...
    git list git@github.com:repo | grep outs: | awk ..
    ```
3. It makes command slower then it is, especially on huge repos

dvc/repo/list_outs.py
JIoJIaJIu: Could you please provide me more info about the case, I cannot understand it correctly
What steps should I proceed for reproducing behavior
There are mine:
```
$ mkdir add_dir
$ dvc add add_dir/                                                             
WARNING: 'add_dir' is empty.                                                                                                                           
....                                                                                                                                                   
                                                                                                                                                       
        git add add_dir.dvc .gitignore
$ python -m dvc list  .
auc.metric
data/data.xml
data/features
add_dir
model.pkl
data/prepared
```

efiop: @JIoJIaJIu Your example is almost right, except that after `dvc add add_dir` you could delete `add_dir`(but not add_dir.dvc) and `dvc list .` should still work and show `add_dir`. For example say you have this git repo on github, and you do `dvc list https://github.com/me/myproj`, then `add_dir` won't exist in git, so your `os.path.isdir()` will fail, because we don't `dvc checkout` in the cloned repo(it should stay that way), so you can't rely on os.path when dealing with dvc-tracked dirs. Let me know if that makes sense.

dvc/repo/list_outs.py
efiop: 1. `dvc import` and `dvc get` work with git files just fine, without any additional flags, so it makes sense to make `dvc list` behave that same way. :slightly_smiling_face: 

2. Differentiation is not required, as `dvc get/import` work the same with git-tracked and dvc-tracked files. :slightly_smiling_face: 

3. Because of us having to list git files too?

efiop:
@JIoJIaJIu @jorgeorpinel `--show-url` is out of scope for this PR. For now there is no agreement on whether it should be moved to `dvc list`.

dvc/cli.py
efiop: Or we might use `ls` for `ls.py` and API method, but keep `dvc list` in CLI, if there are any objections about `dvc list`. Same as we have `dvc import` but `imp()` and `imp.py`.

dvc/repo/list_outs.py
JIoJIaJIu: > `dvc import` and `dvc get` work with git files just

ah, right, now it makes sense

> Differentiation is not required

Ok, maybe it can be shown in future with `dvc list -l repo`

> Because of us having to list git files too?

Yes, mostly time I've spent is trying to speedup the process of listing (skip history, not necessary files, etc). Fetching whole repo before listing doesn't look good for me. But I can retrieve `dvc-files` only and list all other files through `git ls-tree` without fetching them
I think in that case it can be fast enough

JIoJIaJIu: **Proposal**
Maybe needs to support filter option for `list` as `--outputs` for showing outputs only?
```
dvc list --outputs git@github.com:repo
```

efiop: @JIoJIaJIu I would probably not focus on that kind of optimization for now and would use already available mechanisms instead. We have some clone caching present with `git pull`s on top, which is quite fast and we might simply consider making that cache persist over multiple `dvc` invocations instead of messing around with sparse clone/checkout. That being said, I haven't looked into the sparse clone implementation deep enough yet, maybe it would be reasonable to keep, not sure.

efiop:
@JIoJIaJIu Please rebase.

dvc/repo/ls.py
```diff
@@ -0,0 +1,99 @@
+import os
+
+
+@staticmethod
+def ls(url, targets=[], rev=None, recursive=None):
+    if os.path.exists(url):
```
efiop:
```suggestion
    if os.path.exists(url) and not rev:
```
so that revs work with local repos.

dvc/repo/ls.py
```diff
@@ -0,0 +1,99 @@
...
+    files = []
+    if not len(targets):
+        targets.append(".")
+
+    def prettify(prefix, filename, is_root):
+        if not is_root:
+            filename = os.path.join(prefix, filename)
+
+        if os.path.isdir(filename):
+            return f"{filename}/"
```
efiop: fstrings are only supported in 3.6, while we need to support py3.5 as well.

dvc/repo/ls.py
```diff
@@ -0,0 +1,99 @@
...
+    if os.path.exists(url):
+        return _ls_local_repo(url, targets, recursive)
+
+    return _ls_git_repo(url, targets, rev, recursive)
+
+
+def _ls_files_local_repo(path, targets=[], recursive=None):
+    from dvc.repo import Repo
+
+    repo = Repo(path)
```
efiop: Let's not forget to close it :slightly_smiling_face:

dvc/repo/ls.py
```diff
@@ -0,0 +1,99 @@
...
+    files = []
+    if not len(targets):
+        targets.append(".")
+
+    def prettify(prefix, filename, is_root):
+        if not is_root:
+            filename = os.path.join(prefix, filename)
+
+        if os.path.isdir(filename):
```
efiop: Using `os.path` works here, but using it is not correct, strictly speaking :slightly_smiling_face:  as you are using `repo.tree` which might not exist in the real world. `walk` provides you with files and dirs separately, so there is really no need to call any `isdir` here.

dvc/repo/ls.py
```diff
@@ -0,0 +1,99 @@
...
+
+    for target in targets:
+        target_path = os.path.join(path, target)
+        if not os.path.exists(target_path):
+            continue
+
+        for dirpath, dirnames, filenames in repo.tree.walk(target_path):
+            prefix = os.path.relpath(dirpath, path)
+            is_root = os.path.abspath(dirpath) == os.path.abspath(path)
+            files.extend(map(lambda f: prettify(prefix, f, is_root), dirnames))
```
efiop: is_root and prefix are local vars for this inside-loop, so maybe consier moving prettify inside, so you don't have to pass it these and won't have to wrap it with `lambda f`.

dvc/repo/ls.py
```diff
@@ -0,0 +1,99 @@
...
+            files.extend(
+                map(lambda f: prettify(prefix, f, is_root), filenames)
+            )
+            if not recursive:
+                dirnames.clear()
```
efiop: No need to modify `dirnames`
```suggestion
                break
```

dvc/path_info.py
```diff
@@ -67,14 +67,20 @@ def fspath(self):
     def relpath(self, other):
         return self.__class__(relpath(self, other))
 
-    def isin(self, other):
+    def isin(self, other, depth=-1):
```
efiop:
```suggestion
    def isin(self, other, depth=None):
```

dvc/path_info.py
```diff
@@ -67,14 +67,20 @@ def fspath(self):
...
             return False
         # Use cached casefolded parts to compare paths
         n = len(other._cparts)
-        return len(self._cparts) > n and self._cparts[:n] == other._cparts
+        if depth != -1:
+            return (
+                len(self._cparts) - n == depth
+                and self._cparts[:n] == other._cparts
+            )
+        else:
```
efiop: No need in this `else`

dvc/repo/ls.py
```diff
@@ -0,0 +1,99 @@
...
+        outs = outs.union(stage.outs)
+
+    def in_target(out, target_path_info, is_dir):
+        if out.scheme == "local" and out.path_info == target_path_info:
+            return True
+
+        if is_dir:
+            if recursive and out.path_info.isin(target_path_info):
+                return True
+            elif out.path_info.isin(target_path_info, 1):
```
efiop:
```suggestion
            if out.path_info.isin(target_path_info, 1):
```

efiop:
Btw, it should work with git-only repos as well, since `get/import` do too.
```
(3.7.0-dvc) ➜  dvc git:(dvc_list) dvc list https://github.com/iterative/dvc.org                   
ERROR: failed to list 'https://github.com/iterative/dvc.org' - you are not inside of a DVC reposit
ory (checked up to mount point '/')                                                               
                                                                                                  
                                                                                                  
```

dvc/repo/ls.py
JIoJIaJIu: @efiop what is the reason of this check? I think it's better to ignore  `rev` implicitly for the local path or throw an error  for invocation with local path with `rev`

dvc/repo/ls.py
JIoJIaJIu: The idea that I didn't want to create function in every iteration of the loop and make `prettify` more independent

dvc/repo/ls.py
JIoJIaJIu: I  wanted to split out `prettify` from loop logic like the function accepts  the filename and prettify it
I can create 2 functions as
```
def prettify_file(...)

def prettify_dir(...)

File: dvc/repo/ls.py [hunk 20]
```diff
@@ -0,0 +1,109 @@
...
+
+
+def _ls_local_repo(path, targets=[], recursive=None):
+    path = os.path.abspath(path)
+    files = _ls_files_local_repo(path, targets, recursive)
+    outs = _ls_outs_local_repo(path, targets, recursive)
+    result = files + list(map(str, outs))
+    result = list(set(result))
+    result.sort()
+    return result
```
Suor:
```python
return sorted(set(result))
```
Also, why set? Do we expect them to be duplicate?

dvc/repo/ls.py
```diff
@@ -0,0 +1,109 @@
...
+                    map(lambda f: prettify_file(prefix, f, is_root), filenames)
+                )
+                if not recursive:
+                    break
+
+    except NotDvcRepoError:
+        pass
+    finally:
+        repo.close()
+    return files
```
Suor: Overall this looks broken:
```bash
$ tree .
├── a
│   ├── 1.txt
│   └── 2.txt
└── b
    ├── 1.txt
    └── 3.txt

$ dvc list .
1.txt
2.txt
3.txt
```
The output format should be different for cases of single and multiple targets.

dvc/repo/ls.py
```diff
@@ -0,0 +1,109 @@
...
+    result = list(set(result))
+    result.sort()
+    return result
+
+
+def _ls_git_repo(url, targets=[], rev="master", recursive=None):
+    from dvc.external_repo import external_repo
+
+    with external_repo(url, rev) as repo:
+        targets = list(map(lambda t: os.path.join(repo.root_dir, t), targets))
```
Suor: We should not need to convert targets.

Suor:
On single/multiple targets: we don't have any agreement of how this should look for multiple targets. We opted for implementing a flat list, but that only works for single one. See how `ls` works with multiple targets, it's quite different.

dvc/repo/ls.py
```diff
@@ -0,0 +1,109 @@
...
+
+    return _ls_git_repo(url, targets, rev, recursive)
+
+
+def _ls_files_local_repo(path, targets=[], recursive=None):
+    from dvc.repo import Repo
+
+    files = []
+    try:
+        repo = Repo(path)
```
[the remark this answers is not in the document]
JIoJIaJIu: Good point, @Suor , thank you! In fact it's one of the approach, at the first iteration I used `git ls-tree`, I am going to refactor this

dvc/repo/ls.py
```diff
@@ -0,0 +1,109 @@
...
+        if out.scheme == "local" and out.path_info == target_path_info:
+            return True
+
+        if is_dir:
+            if recursive and out.path_info.isin(target_path_info):
+                return True
+            if out.path_info.isin(target_path_info, 1):
+                return True
+
+        return False
```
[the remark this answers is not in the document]
JIoJIaJIu: Agreed, one thing I wanted to clarify
```
dvc list repo target_dir
```
should return like ?
```
target_dir
```
I don't see the use case of such behavior for the user

I think that if user pointes the folder in the target it expect to see the content of it like (without `-R` flag)
```
target_dir/file1
target_dir/subfolder
```
and when he points `-R` flag it means to check subfolder
```

target_dir/file1
target_dir/subfolder
target_dir/subfolder/file2
```

dvc/repo/ls.py
JIoJIaJIu: Let me check, I remember that faced with duplicated rows, but maybe it was an issue in the logic if outs cannot be duplicated

dvc/repo/ls.py
JIoJIaJIu: Can you please provide steps for reproduce?
```
(dvc) jiojiajiu@pon4ik:~/workspace/dvc/list (dvc_list)$ tree .
.
├── a
│   ├── 1
│   └── 2
└── b
    ├── 1
    └── 2

2 directories, 4 files
(dvc) jiojiajiu@pon4ik:~/workspace/dvc/list (dvc_list)$ python -m dvc list .
a
b
(dvc) jiojiajiu@pon4ik:~/workspace/dvc/list (dvc_list)$ python -m dvc list . -R
a
a/1
a/2
b
b/1
b/2
```

JIoJIaJIu:
The last commit contains only changes requested by @efiop , for your notes @Suor I will provide in the next one

>  single/multiple targets: we don't have any agreement of how this should look for multiple targets

I think about targets as a filter that applied on the output, so multiple targets just means multiple filter, similar to
```
dvc list . | grep -E "a|b"
```
but that cares about folder structure
```
(dvc) jiojiajiu@pon4ik:~/workspace/dvc/list (dvc_list)$ tree
.
├── a
│   ├── 1
│   └── 2
├── b
│   ├── 1
│   ├── 2
│   └── a
│       ├── 1
│       └── 2
└── c
    ├── 1
    └── 2

4 directories, 8 files

(dvc) jiojiajiu@pon4ik:~/workspace/dvc/list (dvc_list)$ dvc list . -R | grep -E "a|c"
a
a/1
a/2
b/a
b/a/1
b/a/2
c
c/1
c/2

(dvc) jiojiajiu@pon4ik:~/workspace/dvc/list (dvc_list)$ python -m dvc list . a c -R
a/1
a/2
c/1
c/2
```

`ls` works like differently but IMHO it doesn't make sense to make it the same for `dvc list`, may be just with some option like `-h|--human`
```
(dvc) jiojiajiu@pon4ik:~/workspace/dvc/list (dvc_list)$ ls a b
a:
1  2

b:
1  2  a
(dvc) jiojiajiu@pon4ik:~/workspace/dvc/list (dvc_list)$ ls -l a b
a:
total 0
-rw-rw-r-- 1 jiojiajiu jiojiajiu 0 Feb  2 14:34 1
-rw-rw-r-- 1 jiojiajiu jiojiajiu 0 Feb  2 14:34 2

b:
total 4
-rw-rw-r-- 1 jiojiajiu jiojiajiu    0 Feb  2 14:34 1
-rw-rw-r-- 1 jiojiajiu jiojiajiu    0 Feb  2 14:34 2
drwxrwxr-x 2 jiojiajiu jiojiajiu 4096 Feb  2 14:47 a
```

What do you think?

tests/func/test_ls.py
```diff
@@ -0,0 +1,105 @@
...
+}
+
+DVC_STRUCTURE = {"dvc_file1": "content", "dir1/dvc_file2": "content"}
+
+
+def test_ls_repo(tmp_dir, dvc, scm):
+    tmp_dir.scm_gen(FS_STRUCTURE, commit="init")
+    tmp_dir.dvc_gen(DVC_STRUCTURE, commit="dvc")
+
+    files = Repo.ls(tmp_dir)
```
efiop: We could make ls work with `Path` objects, but it is not worth it and breaks the symmetry with the rest of the commands. So just make sure to covert properly in your tests (they are failing on 3.5 for this reason)

```suggestion
    files = Repo.ls(fspath(tmp_dir))
```

tests/func/test_ls.py
```diff
@@ -0,0 +1,105 @@
...
+def test_ls_repo_recursive(tmp_dir, dvc, scm):
+    tmp_dir.scm_gen(FS_STRUCTURE, commit="init")
+    tmp_dir.dvc_gen(DVC_STRUCTURE, commit="dvc")
+
+    files = Repo.ls(tmp_dir, recursive=True)
+    assert set(files) == set(
+        [
+            "file1",
+            "dir1",
+            "dir1/file2",
```
efiop: This is the reason why windows tests are failing
```suggestion
            os.path.join("dir1", "file2"),
```
and so on down below

dvc/repo/ls.py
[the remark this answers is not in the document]
efiop: This is a really good question. But to answer it, I would look into behavior of existing commands. For example `ls target_dir` and `aws s3 ls s3://bucket/target_dir` will show just `target_dir`. So I would stick to that. :slightly_smiling_face:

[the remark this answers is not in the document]
shcheklein: regarding `-R` - please, let's not use it. Let's use a regular ls-like, aws-ls like `-r`, `--recusrsive`. Along with some flag to filter only outputs.

Also, good question if we want and how do we want to show DVC-files along with the data they reference to.

Also, @JIoJIaJIu could you please share https://asciinema.org/ with different outputs to make it easier to review it? Thanks.

shcheklein:
Please check my comment regarding `-R`. Also, we need an option to show "outputs" only (not sure I like the name in this case, but the idea is to filter only files under DVC-control).

I also think it's not that important to support multiple targets at this point (target is a strange name in this case, btw).

Also, do we show DVC-files at all? Not sure, but it might be useful to show which output comes from what DVC-file. At least we should provide `-a` show all? My experience is that people even do alias to make `ls -la` a default behavior. So, it's a good question - should we optimize for machine readable format or somewhat human readable. cc @iterative/engineering

dvc/repo/ls.py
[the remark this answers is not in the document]
JIoJIaJIu: `ls` shows content of the folder by default, but supports `-d` option for listing dirs rather then content
`aws s3 ls s3://bucket/target_dir` also lists dir and with slash at the end it shows content
`aws s3 ls s3://bucket/target_dir/`
Based on discord communication we decided to implement showing `taget_dir` by default as `ls -d` does.
It allows user to check does the folder exists in the repo or not and make the api more simple
```
$ dvc list url target_dir
target_dir
```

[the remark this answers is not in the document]
JIoJIaJIu: Also I would like to add about it

> This duplicates Repo.find_outs_by_path(..., strict=False)

That cause I've implemented multiple targets support I didn't want to traverse all stages and outs on every target request. So it is a kind of optimization

JIoJIaJIu:
@shcheklein, sorry, I missed your first comment, but now I've read it

What should we do when `-a` option is provided?  
I think either need to support option `--outputs-only` or shows outputs by default and with `-a` add files

<hr/>

About `-l` option:
1. What information for files needs to output? if takes default `--format verbose` it becomes difficult to align/output common files with outputs cause the have different columns
2. Maybe implement `-l` option in the separated PR/issue and define first output colums? 

I see the output like
for `files` -  `{file_type}{mask} {size} {date} {filename}`  
for `outputs` - `{file_type}{mask} {pipeline} {filename}`
```
dvc list -l .
-rw-rw-r-- 1142 Jan 13 08:50 pipeline.dvc
-rw-rw-r-- pipeline.dvc      data/data.xml # what to do here if the outputs already on the fs?
```

dvc/repo/ls.py
JIoJIaJIu: I've got how it can happen - the output can  be fetched/generated and exists on the fs. Hence it will be returned from `_ls_files` and from `_ls_outs` also which leads to duplicates in the output

dvc/repo/ls.py
```diff
@@ -0,0 +1,109 @@
...
+    from dvc.repo import Repo
+
+    files = []
+    try:
+        repo = Repo(path)
+        if not len(targets):
+            targets.append(".")
+
+        def prettify_file(prefix, filename, is_root):
+            if not is_root:
```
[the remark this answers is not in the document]
JIoJIaJIu: @Suor is it ok for windows?

shcheklein:
@JIoJIaJIu we agreed in the ticket to have all files by default, --recursive and --outputs.

> What should we do when `-a` option is provided?

no need for this, it was just to show the idea - simple output (that can be consumed by grep) vs something more useful by default. We need to make that decision.

> Maybe implement -l option in the separated PR/issue and define first output colums?

Yep, it's fine to create a separate PR for this if we decide not do this by default.

-------------------

What about DVC-files? Do we show or skip them?

dvc/repo/ls.py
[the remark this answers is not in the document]
efiop: Yes, cwd is "." on windows too (os.curdir)

dvc/repo/ls.py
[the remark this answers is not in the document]
efiop: @shcheklein Let's use `-R`, same as in other commands. Otherwise it will overlap with `-r|--remote`.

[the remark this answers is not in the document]
efiop: > That cause I've implemented multiple targets support I didn't want to traverse all stages and outs on every target request. So it is a kind of optimization

Let's not optimize it that way then :slightly_smiling_face: Let's only accept one target and use `find_outs_by_path`.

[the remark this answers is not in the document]
shcheklein: ok, yes, I see that ls has `-R` indeed. Disregard my comment on this @JIoJIaJIu!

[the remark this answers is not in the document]
efiop: @JIoJIaJIu hm, though `dvc list url .` is a corner case of that approach :slightly_smiling_face: A natural behavior for `.` is to treat it as `./`. So maybe I'm wrong there and we should consider it more carefully. What do you think? CC @Suor @shcheklein

dvc/repo/ls.py
Suor: This was multiple target, not recursive case. We are not showing target itself, so files in different targets are mixed, anyway this won't be an issue if we don't support multiple targets.

Suor:
> What about DVC-files? Do we show or skip them?

They are just regular git files, not handled in any specific way as I expect.

I understand the default behavior of `dvc list <url> <path>` to be then same as:
```bash
git clone <url> repo
cd repo
dvc pull
ls <path>
```
Are we on the same page here?

dvc/repo/ls.py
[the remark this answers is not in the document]
JIoJIaJIu: @efiop 
I think that `dvc list dir` always should show content of the dir like
```
$ dvc list dir
file1
subdir
```
and apply behavior of `ls -d` only with `target` for checking does it exist or not
```
$ dvc list dir subdir
subdir
```

JIoJIaJIu:
> What about DVC-files? Do we show or skip them?

> They are just regular git files, not handled in any specific way as I expect.

Yes, they are just regular files that are listed by `_ls_files`. In next iterations that can be colored or marked somehow, but at this stage it's just a file

> I understand the default behavior of dvc list <url> <path> to be then same as

Yes, with skipped `.dvc`, `.git`, `.hg` (dvcignore)

dvc/repo/ls.py
```diff
@@ -0,0 +1,108 @@
...
+        segments = [parts.pop(0)]
+        if not recursive:
+            return segments
+
+        while parts:
+            segments.append(os.path.join(segments[-1], parts.pop(0)))
+        return segments
+
+    files = set()
+    for out in outs:
```
efiop: Since we no longer care about multiple targets, could we now use find_outs_by_path/relpath here?

dvc/repo/ls.py
```diff
@@ -0,0 +1,108 @@
...
+    files = (
+        _ls_files_local_repo(path, target, recursive)
+        if not outputs_only
+        else []
+    )
+    outs = _ls_outs_local_repo(path, target, recursive)
+    result = files + outs
+    result = list(set(result))
+    result.sort()
+    return result
```
efiop: For example, this could be done something like:
```
def _ls_local_repo(repo, target=None, recursive=None, outputs_only=False):
    result = set()                                                        
    if not outs_only:                                                     
        result.update(_ls_files_local_repo(repo, target, recursive))      
    if isinstance(repo, Repo):                                            
        result.update(_ls_outs_local_repo(repo, target, recursive))       
    result = list(result).sort()                                          
    return result                                                         
```

if you abstract away local url handling one level above, so you won't need to re-open Repo one level below.

dvc/repo/ls.py
```diff
@@ -0,0 +1,108 @@
...
+            if not recursive:
+                return files
+
+        for dirpath, dirnames, filenames in tree.walk(target_path):
+            prefix = os.path.relpath(dirpath, path)
+            files.extend(map(lambda f: prettify(prefix, f), dirnames))
+            files.extend(map(lambda f: prettify(prefix, f), filenames))
+            if not recursive:
+                break
+    except (NotADirectoryError, FileNotFoundError):
```
efiop: So if you run list with non-existing target
```
$ dvc list https://github.com/iterative/dvc.org sdfsdf                                                                                      
```
it won't show anything (except for a newline) and will finish successfully. It doesn't seem right to me.

dvc/repo/ls.py
```diff
@@ -0,0 +1,108 @@
...
+    files = set()
+    for out in outs:
+        if out.path_info.isin_or_eq(path_info):
+            if not recursive and target:
+                files.add(target)
+                break
+
+            files.update(prettify(out))
+
+    repo.close()
```
efiop: Here is where you are forgetting to put it into the `finally` block or something :) Hopefully the refactoring that I've mentioned above will handle this.

dvc/command/ls.py
```diff
@@ -0,0 +1,64 @@
...
+        "C:\\\\path\\to\\file\n"
+        "C:\\\\path\\to\\directory\n"
+        "https://github.com/path/to/repo\n"
+        "git@github.com:path/to/repo.git\n",
+    )
+    list_parser.add_argument(
+        "-R", "--recursive", action="store_true", help="Recursively list files"
+    )
+    list_parser.add_argument(
+        "--outputs-only", action="store_true", help="Show only dvc outputs."
```
efiop: We have a similar option called `--outs-only` in `dvc remove`, which is admitedly not our most popular command, but I would consider keeping it consistent. CC @jorgeorpinel

dvc/repo/ls.py
JIoJIaJIu: > it won't show anything (except for a newline) and will finish successfully. It doesn't seem right to me

What do you expect to see, an error? I think empty output is correct behavior in such case cause target is applied as a filter, so passing wrong/doesn't existed target means [] files output

I will add test for the case

dvc/repo/ls.py
JIoJIaJIu: I haven't got you point, unfortunately
I close repo at the end of the work and use `try`/`except` above trying to open repo

dvc/repo/ls.py
JIoJIaJIu: The point here that `_ls_files_local_repo` can work with pure filesystem. I think I need to rename it

dvc/repo/ls.py
efiop: If any error happens between `Repo()` and this `close()` you will leak resources. Better to try to organise your code in such a way to not leak anything even if you get some error that you don't expect. I think it will be solved by abstracting above, so that it is similar to `with external_repo` that closes your repo instance automatically when you are exiting the context, with error or without.

dvc/repo/ls.py
efiop: Well, for now it is not really a filter but rather a recursive list, right? It accepts something that should exist (e.g. directory) and lists everything under it. So this specific behaviour is similar to `$ ls non-existing` which raises an error and further away from `aws s3 ls` which always treats your argument as a filter.

dvc/repo/ls.py
JIoJIaJIu: I don't like `repo.find_outs_by_path/relpath` cause:

## Case#1
I request `dvc list repo`  

`repo.find_outs_by_path` return noting on trying to get content of the repo without `recursive=True`, what requires always use it with `recursive` and filter then the output depends on the depth
```
>>> os.getcwd()
'/home/jiojiajiu/workspace/dvc/test-project'
>>> repo = Repo(os.getcwd())
>>> repo.find_outs_by_path('.')
  File "<stdin>", line 1, in <module>
  File "/home/jiojiajiu/workspace/dvc/dvc/repo/__init__.py", line 416, in find_outs_by_path
    raise OutputNotFoundError(path, self)
dvc.exceptions.OutputNotFoundError: Unable to find DVC-file with output '.'
>>> repo.find_outs_by_path('.', recursive=True)
[OutputLOCAL: 'dup/1', OutputLOCAL: '1', OutputLOCAL: 'add_dir', OutputLOCAL: 'model.pkl', OutputLOCAL: 'data/prepared', OutputLOCAL: 'test_dir', OutputLOCAL: 'auc.metric', OutputLOCAL: 'data/features', OutputLOCAL: '2', OutputLOCAL: 'data.xml']
```
but
```
(dvc-get-started) jiojiajiu@pon4ik:~/workspace/dvc/test-project$ dvc list . --outputs-only
1
2
add_dir
auc.metric
data
dup
model.pkl
test_dir
```

efiop: Might be missing something, but isn't that what you currently do? You filter-out all of the outputs and then do some segments magic to get only first part of the path.

dvc/command/ls.py
jorgeorpinel: Agree. I vote for `--outs-only`.

jorgeorpinel: @JIoJIaJIu please also capitalize DVC in the `help` string.
Oh and please end the previous option's `help` in period `.` as well.
Thanks

jorgeorpinel:
> I understand the default behavior of dvc list <url> <path> to be then same as:
> 
> ```console
> git clone <url> repo
> cd repo
> dvc pull
> ls <path>
> ```

But do we actually have to clone the repo and `dvc pull` internally for `dvc list`? Or can it be done in a less heavy way?

JIoJIaJIu:
Hello @jorgeorpinel 
> But do we actually have to clone the repo and dvc pull internally for dvc list?

At this point we clone repo with whole history and then list files and outputs (based on dvc-files). To make it less heavy is possible but I think the first needs to determine the bottleneck case and bounds (size of the repo, command time)

> dvc pull internally

At this point artifacts are listed based on `dvc-file` and aren't fetched, if I got you right

dvc/repo/ls.py
JIoJIaJIu: > Well, for now it is not really a filter but rather a recursive list, right? It accepts something that should exist (e.g. directory) and lists everything under it

In fact no, generally based on the https://github.com/iterative/dvc/pull/3246#discussion_r373874769 `dvc list url target` checks does the target exists or not. Only if we provide `-R` flag it becomes recursive and shows content

> this specific behaviour is similar to $ ls non-existing which raises an error

I think this behavior is covered with `dvc list not_existed` which throws an error (I think the error message should be improved)
```
(dvc) jiojiajiu@pon4ik:~$ dvc list not_existed
ERROR: failed to list 'not_existed' - Failed to clone repo 'not_existed' to '/tmp/tmp7pu3j_76dvc-clone': Cmd('git') failed due to: exit code(128)
  cmdline: git clone --no-single-branch -v not_existed /tmp/tmp7pu3j_76dvc-clone
  stderr: 'fatal: repository 'not_existed' does not exist
```

but with `dvc list repo target` target works as a filter
```
dvc list repo not_exitsted

```
and returns nothing if there is no anything that matches

I think cause we always show the target if it matches, pointing empty output means that nothing was found with such target
What do you think @efiop

dvc/repo/ls.py
JIoJIaJIu: I think you are right @efiop, my bad

dvc/command/ls.py
JIoJIaJIu: Sure, @jorgeorpinel, thanks for pointing me out

dvc/repo/ls.py
```diff
@@ -0,0 +1,107 @@
...
+def _ls_local_dir(path, target=None, recursive=None, outputs_only=False):
+    from dvc.repo import Repo
+
+    result = set()
+    if not outputs_only:
+        result.update(_ls_files(path, target, recursive))
+    try:
+        repo = Repo(path)
+        result.update(_ls_outs_local_repo(repo, target, recursive))
+        repo.close()
```
efiop: You are still not handling the closure correctly. This should be in the `finally` block, so if any error happens it will be closed.

dvc/repo/ls.py
efiop: @JIoJIaJIu I think we should raise an error after all. Additional reason is that `dvc get/import` in that case would raise an error.

dvc/repo/ls.py
```diff
@@ -0,0 +1,107 @@
...
+def ls(url, target=None, rev=None, recursive=None, outputs_only=False):
+    if os.path.exists(url):
+        if rev:
+            raise DvcException("Local repo doesn't support revisions")
+        return _ls_local_dir(url, target, recursive, outputs_only)
+
+    try:
+        return _ls_remote_git_repo(url, target, rev, recursive, outputs_only)
+    except CloneError:
+        raise DvcException("There is no such url {}".format(url))
```
efiop: CloneError doesn't mean that there is no such url, it only means that we were not able to clone it (e.g. because of access issues). Even if we decide to wrap it, it is not suitable to do it here, it should be in `external_repo` or somewhere there.

dvc/repo/ls.py
```diff
@@ -0,0 +1,107 @@
...
+    for out in outs:
+        if not recursive and target and len(outs):
+            files.add(target)
+            break
+        files.update(prettify(out))
+
+    return list(files)
+
+
+def _ls_local_dir(path, target=None, recursive=None, outputs_only=False):
```
efiop:
```suggestion
def _ls_local_repo(path, target=None, recursive=None, outputs_only=False):
```
? It is about repo right, not about any "dir"?

dvc/repo/ls.py
```diff
@@ -0,0 +1,107 @@
...
+    from dvc.repo import Repo
+
+    result = set()
+    if not outputs_only:
+        result.update(_ls_files(path, target, recursive))
+    try:
+        repo = Repo(path)
```
efiop: `external_repo` provides you with `repo` already, this re-opening feels excessive and out-of-place here. Maybe we could do better?

dvc/repo/ls.py
```diff
@@ -0,0 +1,107 @@
...
+
+from dvc.exceptions import DvcException, NotDvcRepoError, OutputNotFoundError
+from dvc.scm.base import CloneError
+
+
+@staticmethod
+def ls(url, target=None, rev=None, recursive=None, outputs_only=False):
+    if os.path.exists(url):
+        if rev:
+            raise DvcException("Local repo doesn't support revisions")
```
efiop: There is an inconsistency with `dvc get` and `dvc import` here, as they both just `external_repo` even local urls, while `ls`treats it more like `get_url` from `api`. I do see the point in that, but it feels like we should be consistent here and since get/import are out of focus for this PR, we should make `ls` behave the same way. CC @shcheklein @Suor

dvc/repo/ls.py
```diff
@@ -0,0 +1,107 @@
...
+            if not recursive:
+                break
+    except (NotADirectoryError, FileNotFoundError):
+        pass
+
+    return files
+
+
+def _ls_outs_local_repo(repo, target=None, recursive=None):
+    def prettify(out):
```
efiop: Naming is misleading here. It doesn't prettify but rather extracts first part of the path in non-recursive case.

dvc/repo/ls.py
JIoJIaJIu: @efiop this method is used not only from `_ls_remote_git_repo`

dvc/repo/ls.py
JIoJIaJIu: Not only, it outs the list of strings for the pointed output
In case of non-recursive call the response contains the topest dir and in recursive - list of segments to the out. So it takes out and returs it's representation list
I've named it such way cause we have similar function in `_ls_files`, and purpose of it was prettify out somehow. I think it can be renamed to `get_segments`

dvc/repo/ls.py
JIoJIaJIu: `_ls_local_dir` can traverse simple file path, not repo
https://github.com/iterative/dvc/pull/3246#discussion_r373760365

dvc/repo/ls.py
JIoJIaJIu: It's an attempt to raise an error on non existed repo https://github.com/iterative/dvc/pull/3246#discussion_r375929735 but you are right. One problem that needs to extend errors for getting more info. The reason I've caught and raise the error here cause I would like to point that `url` that was requested in `dvc list url` doesn't exist. So I think it's right place for it, but I need to get more info from `dvc.scm`

efiop: Not sure I follow how that relates to that comment. In that comment I was talking about path within the repo that doesn't exist, not the url of the repo that doesn't exist.

dvc/repo/ls.py
efiop: @JIoJIaJIu So `dvc list some/random/directory` would work? I don't see that being discussed in the comment you've linked. And, once again, it is not how `dvc get/import` work.

dvc/repo/ls.py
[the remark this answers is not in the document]
efiop: @Suor For the record: `get` and `import` don't care about dvcignores when working with git-files, so I don't think we should use clean tree there.

dvc/repo/ls.py
```diff
@@ -0,0 +1,90 @@
...
+    with external_repo(url, rev) as repo:
+        result = set()
+        if isinstance(repo, Repo):
+            result.update(_ls_outs_repo(repo, target, recursive))
+        if not outputs_only:
+            try:
+                result.update(_ls_files_repo(repo, target, recursive))
+            except DvcException as exc:
+                if not len(result):
+                    raise exc
```
efiop: Empty output from `_ls_outs_repo` doesn't mean that this is an error, as it swallows OutputNotFoundError so in your code you are not distinguishing between "no output found" and "output directory is empty" cases. Please take a look at `dvc get`, or more specifically at `ExternalRepo.pull_to`, it has a lot of similar logic that you can apply here.

dvc/repo/ls.py
```diff
@@ -0,0 +1,90 @@
...
+
+    files = []
+    target_path = (
+        os.path.join(repo.root_dir, target) if target else repo.root_dir
+    )
+    tree = CleanTree(WorkingTree(target_path))
+
+    try:
+        if not os.path.exists(target_path):
+            raise FileNotFoundError("{} doesn't exist".format(target_path))
```
efiop: This error message is meaningless, you are catching it right away down below and re-raising it. You could raise what you want right away.

dvc/repo/ls.py
```diff
@@ -0,0 +1,90 @@
...
+
+        for dirpath, dirnames, filenames in tree.walk(target_path):
+            prefix = os.path.relpath(dirpath, repo.root_dir)
+            files.extend(map(lambda f: prettify(prefix, f), dirnames))
+            files.extend(map(lambda f: prettify(prefix, f), filenames))
+            if not recursive:
+                break
+    except FileNotFoundError:
+        raise DvcException("target {} doesn't exist".format(target))
+    except NotADirectoryError:
```
efiop: This one only applies to `tree.walk`, right? And `FileNotFoundError` re-raise is not needed as described in https://github.com/iterative/dvc/pull/3246/files#r376130721 , so you at least you should narrow the scope of this try &except NotADirectoryError to `tree.walk`, so this is less confusing.

dvc/repo/ls.py
efiop: Oh, though I didn't notice that if the dir is empty we are still returning ["empty_dir"], so this does work. :+1:

dvc/repo/ls.py
```diff
@@ -0,0 +1,90 @@
...
+        pass
+
+    return files
+
+
+def _ls_outs_repo(repo, target=None, recursive=None):
+    def get_path_segments(out):
+        relpath = out.path_info.relpath(repo.root_dir)
```
efiop: Hm, repo.root_dir part doesn't look correct. E.g.
```
dvc list . scripts/innosetup -R
scripts
scripts/innosetup
scripts/innosetup/.gitignore
...
```
where `scripts` and `scripts/innosetup` seem like they don't belong there. Or was it intended?

jorgeorpinel:
> At this point we clone repo with whole history and then list files and outputs (based on dvc-files).

OK that's probably fast enough, as long as we're not actually doing `dvc pull`, just reading the DVC-files.

dvc/command/ls.py
```diff
@@ -0,0 +1,67 @@
...
+        "--outs-only", action="store_true", help="Show only DVC outputs."
+    )
+    list_parser.add_argument(
+        "--rev", nargs="?", help="Git revision (e.g. branch, tag, SHA)."
+    )
+    list_parser.add_argument(
+        "target",
+        nargs="?",
+        help="Path to directory within the repository to list outputs for.",
+    )
```
jorgeorpinel: Minor: these 2 help ouputs don't need a period `.` by the end (not full sentences) 🙂 

p.s. thanks for changing the shell autocomplete scripts! Try to double check the text is the same there as well.

dvc/repo/ls.py
JIoJIaJIu: with `external_repo` `dvc list` now works similar to `dvc get`

dvc/repo/ls.py
JIoJIaJIu: At this point `dvc list some/random/directory` throws an error
```
$ dvc list some/random/directory
ERROR: failed to list 'some/random/directory' - Failed to clone repo 'some/random/directory' to '/tmp/tmp65hwp5m5dvc-clone': Cmd('git') failed due to: exit code(128)
  cmdline: git clone --no-single-branch -v some/random/directory /tmp/tmp65hwp5m5dvc-clone
  stderr: 'fatal: repository 'some/random/directory' does not exist

```

dvc/repo/ls.py
```diff
@@ -0,0 +1,73 @@
...
+    with external_repo(url, rev) as repo:
+        result = []
+        if isinstance(repo, Repo):
+            result.extend(_ls_outs_repo(repo, target, recursive))
+
+        if not outputs_only:
+            result.extend(_ls_files_repo(repo, target, recursive))
+
+        if target and not len(result):
+            raise DvcException("target {} doesn't exist".format(target))
```
efiop: Maybe PathMissingError as in `pull_to`?

dvc/repo/ls.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+@staticmethod
+def ls(url, target=None, rev=None, recursive=None, outputs_only=False):
+    from dvc.external_repo import external_repo
+    from dvc.repo import Repo
+    from dvc.utils import relpath
+
+    with external_repo(url, rev) as repo:
+        result = []
```
efiop: Why not `set()` right away? You are converting it later anyway

efiop:
@JIoJIaJIu There is some bug:
```
(3.7.0-dvc) ➜  dvc git:(dvc_list) ✗ dvc list https://github.com/iterative/dvc scripts 
scripts                                                                               
(3.7.0-dvc) ➜  dvc git:(dvc_list) ✗ dvc list https://github.com/iterative/dvc scripts/
scripts                                                                               
```

efiop:
@JIoJIaJIu Shouldn't it show files and dirs in `scripts` even without `--recursive`? I think it should, same as `ls scripts`

dvc/repo/ls.py
```diff
@@ -0,0 +1,73 @@
...
+        return []
+
+    if target and not recursive:
+        return [PathInfo(target_path)]
+
+    try:
+        for dirpath, dirnames, filenames in tree.walk(target_path):
+            files.extend(map(lambda f: PathInfo(dirpath, f), filenames))
+            if not recursive:
+                files.extend(map(lambda d: PathInfo(dirpath, d), dirnames))
```
efiop: This is probably the cause of that bug. This line should be outside of `if`, right?

JIoJIaJIu:
> @JIoJIaJIu There is some bug:
> 
> ```
> (3.7.0-dvc) ➜  dvc git:(dvc_list) ✗ dvc list https://github.com/iterative/dvc scripts 
> scripts                                                                               
> (3.7.0-dvc) ➜  dvc git:(dvc_list) ✗ dvc list https://github.com/iterative/dvc scripts/
> scripts                                                                               
> ```

It's not a bug but a feature))
Let me summarize request

> When `target` contains slash at the end (`/` - nix and `\` - for windows) needs to output content of the target

dvc/repo/ls.py
JIoJIaJIu: Cause I am everywhere uses list (as outs from `_ls_outs_repo`  and `_ls_files_repo`) and expect to return sorted output. I'm using `set` only for skipping possible duplicates, but not as a general data-structure

dvc/command/ls.py
```diff
@@ -0,0 +1,67 @@
...
+
+        try:
+            nodes = Repo.ls(
+                self.args.url,
+                self.args.target,
+                rev=self.args.rev,
+                recursive=self.args.recursive,
+                outputs_only=self.args.outs_only,
+            )
+            logger.info("\n".join(nodes))
```
skshetry: If there are no nodes, `logger.info` will add a newline on the output.

tests/func/test_ls.py
```diff
@@ -0,0 +1,153 @@
...
+    m = mocker.patch("dvc.scm.Git.clone")
+
+    def clone_repo(url, to_path, *args):
+        assert url == "git://github.com:repo"
+        copy_tree(fspath(tmp_dir), to_path)
+        return Git(fspath(tmp_dir))
+
+    m.side_effect = clone_repo
+
+    files = Repo.ls("git://github.com:repo")
```
skshetry: You can use `erepo_dir` fixture to create an external dvcrepo and `git_dir` for external git-only repository. If you want to test with url, you can use `file://` based urls as well. No need to mock.

dvc/repo/ls.py
```diff
@@ -0,0 +1,73 @@
...
+from dvc.exceptions import DvcException, OutputNotFoundError
+
+
+@staticmethod
+def ls(url, target=None, rev=None, recursive=None, outputs_only=False):
+    from dvc.external_repo import external_repo
+    from dvc.repo import Repo
+    from dvc.utils import relpath
+
+    with external_repo(url, rev) as repo:
```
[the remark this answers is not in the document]
efiop: @skshetry That is expected, same as in `dvc get/import`.

dvc/repo/ls.py
```diff
@@ -0,0 +1,73 @@
...
+def _ls_files_repo(repo, target=None, recursive=None):
+    from dvc.scm.tree import WorkingTree
+    from dvc.ignore import CleanTree
+    from dvc.path_info import PathInfo
+
+    files = []
+    target_path = (
+        os.path.join(repo.root_dir, target) if target else repo.root_dir
+    )
+    tree = CleanTree(WorkingTree(target_path))
```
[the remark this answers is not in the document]
efiop: @skshetry `repo` might be `Git` repo.

efiop:
> I'd like to have a more few tests for external repo (git-only + dvc) with possible combinations such as rev, etc. The tests you have added are good, but mostly covers the local dvc repo cases only. slightly_smiling_face

@skshetry  Good point! This patch only uses `external_repo`, same as other commands like get/import, so creating more tests for it is out-of-focus for this PR, we can create a ticket and get back to it later though.

dvc/repo/ls.py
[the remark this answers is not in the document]
skshetry: I don't think consistency argument holds, as `get/import`'s responsibility is working only with external repository, but `list` **should** work with both external repo and local repo.

But, I agree that we can improve this later on and not that important atm :slightly_smiling_face:

dvc/repo/ls.py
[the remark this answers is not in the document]
skshetry: Thanks, @efiop

tests/func/test_ls.py
efiop: Not sure if `file://` works these days. In any case, the logic still passes this to `external_repo`, so we could use local paths here with the same effect. I know that previously we've had a different implementation here for which the url in tests mattered, but it is no longer like that, so we could simplify and use `erepo_dir` as @skshetry suggested :slightly_smiling_face:

skshetry: @Suor always adds `file://` for external repo in combination with `erepo_dir` with`# Use URL to prevent any fishy optimizations` kind-of comments. 

https://github.com/iterative/dvc/blob/7821fa6866a492c0675248028039257d292a09ec/tests/func/test_external_repo.py#L48-L50

JIoJIaJIu:
I would like to pay attention on the command behaviour when `target` requested with `--outs-only` flag

`LICENSE` target exists into the repo (simple file)
```
$ dvc list https://github.com/iterative/dvc LICENSE
LICENSE
```
But when we request `--outs-only` the command returns nothing and doesn't throw an error
```
$ dvc list https://github.com/iterative/dvc LICENSE --outs-only

```
```
$ dvc list https://github.com/iterative/dvc NOT_EXISTED_LICENSE
ERROR: failed to list 'https://github.com/iterative/dvc' - The path 'NOT_EXISTED_LICENSE' does not exist in the target repository 'ExternalRepo: '/tmp/tmpms1fw854dvc-erepo'' neither as an output nor a git-handled file.

```

efiop:
> I would like to pay attention on the command behaviour when target requested with --outs-only flag

@JIoJIaJIu Do you think we should raise an error there? :slightly_smiling_face:

JIoJIaJIu:
> > I would like to pay attention on the command behaviour when target requested with --outs-only flag
> 
> @JIoJIaJIu Do you think we should raise an error there? slightly_smiling_face

@efiop I see that we shouldn't. Idea is that target exists into the repo, but this target doesn't contain `dvc artificats` (and isn't it by itself), for example
```
dvc list url my_folder_which_exists --outs-only
```
The behavior with throwing an error as `not existed target` can confuse user rather then empty output, which shows how many(0) `dvc artifacts` exist into the target

efiop:
@JIoJIaJIu But `--outs-only` implies that you are only looking for outputs, so it erroring out if it doesn't find any outputs makes sense.  `--outs-only` is only limiting the scope, the behaviour should stay the same as with full scope.

I think it is better to illustrate it with an example. Imagine that you are writing a shell script (with `set -e` that will make it error-out on first error) and want to use `dvc list` output to `dvc get` stuff. You would do something like:
```
dvc list URL path | xargs -0 dvc get URL "{}"
```
and if there is no `path` in `URL`, then `dvc ls` would error-out. Now you want to only get outputs. You would do:
```
dvc list URL path --outs-only | xargs -0 dvc get URL "{}"
```
but now your `dvc list` won't error-out, but just will print nothing, so your script will keep going. Which means that `dvc list` behaviour is inconsistent in how it handles errors.

dvc/repo/ls.py
```diff
@@ -0,0 +1,85 @@
...
+        if isinstance(repo, Repo):
+            result.extend(_ls_outs_repo(repo, target, recursive))
+        files = _ls_files_repo(repo, target, recursive)
+
+        if target and not len(result) and not len(files):
+            raise PathMissingError(target, repo)
+
+        if not outs_only:
+            result.extend(files)
+        target_path_info = _get_target_path_info(repo, target)
```
efiop: I'm pretty sure you can move this up and just pass target_path_info instead of `target` to other methods. Currently you are calling this helper three times for no reason :slightly_smiling_face:

JIoJIaJIu: > Currently you are calling this helper three times for no reason

The idea was to keep function interface lightweight and simple - `_ls_outs` and `_ls_files` requires only `repo` which is transparent and make the functions more simple to use. With `target_path_info` param I don't like much how functions signature looks

JIoJIaJIu:
@efiop I've applied the changes. I think it's a philosophy question and a few ways are possible.
If you think it's right way - let's do it, but let me summarize my thoughts bellow, maybe you find them relevant

## TLDR

As you know the idea of throwing an error on not existed target is not close to me in general
I think `get` throws an error on not existed `target` cause the result of the command depends on the existence, but it's not the same for `list` command
If we list not existed url (aka repo) - the error should be thrown, but when we list not existed target - shouldn't. I thought about the url as a **source** and about the target as a **filter**.  
After conversation we got that **source = url + target**

> and if there is no path in URL, then dvc ls would error-out

we've chosen such behavior as a requirement, so the error in the script shows that **the requested source is invalid** 

> but now your dvc list won't error-out, but just will print nothing, so your script will keep going. Which means that dvc list behaviour is inconsistent in how it handles errors

here, the **requested source is valid**, but there are **no outs** there - what is the different case. So if our script should get 1+ outputs on list it should be solved separately and
```
dvc list URL path --outs-only | xargs -0 dvc get URL "{}"
```
shouldn't throw if there are no outs on path but path exists

efiop:
@JIoJIaJIu Thanks for the explanation! I don't think it justifies the behaviour inconsistency from my example though :slightly_smiling_face: 

I would understand your approach if we would've stuck to filter-like approach in the base case, where we would not error-out and would simply print nothing, but I think it contradicts itself in its current form as it mismatches the approaches and results in an inconsistency between those two cases.

I feel like the misunderstanding here is in the way we understand the ordering of operations here. You seem to be thinking about `--outs-only` as a filter that is applied ON TOP of the result from the default `dvc ls`, while I think about it as a filter that applies to the set of files that is being fed to `dvc ls`. So in your view `dvc ls --outs-only` is `filter(filter(ALL_FILES, path), is_dvc_out)` while it is `filter(filter(ALL_FILES, is_dvc_out), path)` in my view, so I can clearly see that the behaviour should be the same as in default case.

But let's consider your view. We've established that in default case `filter(ALL_FILES, path)` raises an error if path doesn't exist, right? If so and we generalize it, we get that `filter(SET, criteria)` should raise an error if criteria is not met, right? If so, your view can be rewritten as `filter(FILTERED_SET, is_dvc_out)`, so if there are no dvc outs in the FILTERED_SET, it means that criteria is not met and we need to raise an error, so your approach contradicts itself in its current form. :slightly_smiling_face:

dvc/repo/ls.py
```diff
@@ -0,0 +1,84 @@
...
+        result = []
+        if isinstance(repo, Repo):
+            result.extend(_ls_outs_repo(repo, target_path_info, recursive))
+
+        if not outs_only:
+            result.extend(_ls_files_repo(target_path_info, recursive))
+
+        if target and not result:
+            raise PathMissingError(target, repo)
```
efiop: For the record: discussing improving this error in `--outs-only` case in PMs.

dvc/command/ls.py
```diff
@@ -0,0 +1,68 @@
...
+    list_parser.add_argument(
+        "--outs-only", action="store_true", help="Show only DVC outputs."
+    )
+    list_parser.add_argument(
+        "--rev", nargs="?", help="Git revision (e.g. branch, tag, SHA)"
+    )
+    list_parser.add_argument(
+        "target",
+        nargs="?",
+        help="Path to directory within the repository to list outputs for",
```
jorgeorpinel: If `target` can only be a directory path, why not call it `path` or `dir`?

dvc/exceptions.py
```diff
@@ -294,12 +294,23 @@ def __init__(self, code, reason):
 
 
 class PathMissingError(DvcException):
-    def __init__(self, path, repo):
+    default_msg = (
+        "The path '{}' does not exist in the target repository '{}'"
+        " neither as an output nor a git-handled file."
```
jorgeorpinel: "...as a DVC output nor as a Git-tracked file."

dvc/exceptions.py
```diff
@@ -294,12 +294,23 @@ def __init__(self, code, reason):
...
+    )
+    default_msg_output_only = (
+        "The path '{}' does not exist in the target repository '{}'"
+        " as an output."
```
jorgeorpinel: as an DVC output

Issue #2901, opened 2019-12-05
init: rename/alias --no-scm and --subdir?


dashohoxha:
UPDATE: Skip to https://github.com/iterative/dvc/issues/2901#issuecomment-596887165

Since Git is the only SCM supported and there are no plans (yet) to support any other SCM.

shcheklein:
We can make an alias and deprecate, hide --no-scm

jorgeorpinel:
Context: https://github.com/iterative/dvc.org/issues/540#issuecomment-562326532

jorgeorpinel:
We could even automate detecting whether there's a Git repo underneath and not offer this option, merging this issue with #2472.

jorgeorpinel:
Also, maybe --subdir should be --sub-git? (Context: https://github.com/iterative/dvc.org/pull/1022#issuecomment-592673004)

shcheklein:
I really don't the idea of `--sub-git` - super confusing. I understood the motivation for the `--subdir` (to make the intention extra clear and explicit), but I still prefer it to be DVC-centric, going from DVC use cases when we describe it. Meaning that Git should be part of the technical explanation, but not an initial user-facing interface.

The use case for me - is being able to create isolated DVC sub-projects. Mostly due to monorepos. So, starting explanations from Git or naming an option this leads this into a wrong direction.

jorgeorpinel:
Good points, I agree with a lot of that now that I thing about it.

> The use case for me - is being able to create isolated DVC sub-projects

Except it kind of is about Git originally, here's the original issue: **Support for multiple .dvc roots in a single git repo** iterative/dvc/issues/2349 . You can have a parent DVC project started with --subdir. "subdir" refers to a subdirectory of a Git repo, which can also be a DVC repo but that's not the key aspect.

Another interface I just thought of would be to kind of join these options into a single `--git` (or `--scm` if we want to avoid the work Git in our UI) that accepts 3 values: `root` (default), `subdir`, and `no`. To me right now that would make the most sense semantically.

```console
$ mkdir repo && cd repo
$ git init
$ dvc init --scm=root  # same as plain  dvc init

$ mkdir repo && cd repo
$ git init
$ mkdir dir && cd dir
$ dvc init --scm=subdir

$ mkdir project && cd project
$ dvc init --scm=no
```

Cc @efiop @pared thoughts?

shcheklein:
@jorgeorpinel 

looks like you don't need one of those (at least `--scm=root`)? 

[...]

Issue #2706, opened 2019-10-31
Allow restriction `dvc pull` to targets


Titousensei:
We have a big mono repo where all the projects are eventually merged. We plan to put the dvc files related to a model in the same directory, and we will have many such directories. In that case, doing `dvc pull` will start pulling all the files cached in remote storage (potentially terabytes of data), right? If a user does `dvc pull <target>` she will only load the cached file needed by this target, which is what we really want the user to do.

It would be nice to have a way to prevent the `dvc pull` without a target, like a config option `pull.disable_global`.

shcheklein:
Context: https://discordapp.com/channels/485586884165107732/563406153334128681/639582276388388870

Suor:
I think letting user interrupt the process is ok solution unless there is a significant delay before it's obvious that lots of things will be pulled now.

dmpetrov:
It is quite dangerous to change a command behavior (syntax of the command in this case) by a config option. This  should be done only in exceptional cases.

For this particular case, managing multiple DVC projects inside a single mono-repo might be a better solution with additional benefit. See #2349. The implementation is coming soon #3257.

I’m closing the issue in favor of #2349. Please provide more details if I missed something and multiple dvc repro won’t solve the issue.

Suor:
@dmpetrov #2349 looks way more dangerous and way more heavy.

dmpetrov:
@Suor it is more general and it should not be heavy. Otherwise, we will end up with many ways of doing the same thing.

Issue #2472, opened 2019-09-06
Consider throwing exception/logging warning on `dvc init --no-scm` in git repo


pared:
Currently DVC will silently ignore `--no-scm` on `dvc init` in `git` repository and use `git` as SCM.

That is not a big problem, but can make tests inconsistent:

`def test1(dvc_repo, git)`

and

`def test2(git, dvc_repo)`

will behave differently, as in second case `dvc_repo.scm` is `Git` instance and in first: `NoSCM`

jorgeorpinel:
I ➕1 this.

If a user accidentally `dvc init --no-scm`s in a Git repo, they may end up tracking data both with Git and DVC.

jorgeorpinel:
We could even automate detecting whether there's a Git repo underneath and not offer this option, merging this issue with #2901.

efiop:
This is no longer the case, as `no-scm` is now explicit, so sequential dvc runs won't use git automatically.

@jorgeorpinel `dvc init --no-scm` doesn't look like an accidental thing, it is a very explicit command.

jorgeorpinel:
What do you think about automating Git detection so the `--no-scm` option is not even needed @pared, @efiop, @shcheklein? Is it worth opening its own issue? Thanks

efiop:
@jorgeorpinel So if no git is detected we will assume no-scm? If so, that is error-prone.

jorgeorpinel:
Yes, that's my suggestion. I'm trying to think of how that would be error prone but I'm not seeing it:

If the directory is Git-tracked, why would you use `dvc init --no-scm`? That would be a mistake (as we discussed in https://github.com/iterative/dvc/issues/2472#issuecomment-590184536 and https://github.com/iterative/dvc/issues/2472#issuecomment-591239038 above) so actually not having the explicit option seems like its LESS error-prone from this PoV.

If it's not tracked by Git, you HAVE to use `--no-scm`, or `dvc init` fails.

efiop:
@jorgeorpinel Because it is an explicit action and you might want to `--no-scm` in a git repo if you don't want dvc to mess around with your git. We've used it previously for some debugging purpuses too. Yes, it is a very specific advanced feature that is not popular, but it has its own explicit flag that is unlikely to be used accidentally. Let's just leave it as is, it doesn't harm anything. :slightly_smiling_face:

jorgeorpinel:
[...]

PR #3257, opened 2020-01-30
SCM: allow multiple DVC repos inside single SCM repo

### LINKED ISSUE #2349: Support for multiple .dvc roots in a single git repo
**This request comes from a large company that has used DVC in the past but moved away mostly due to this issue.**

Some companies (.e.g [Google](https://cacm.acm.org/magazines/2016/7/204032-why-google-stores-billions-of-lines-of-code-in-a-single-repository/fulltext)) store all of their different projects' and teams' code in a single big git repo.

In this scenario, each project has its own subdirectory in the repo, and they are expected to only make changes to that subdirectory, unless they are contributing code to another project in the organization.

Unfortunately, DVC can only create its `.dvc` folder in the git repo root. This is problematic for a couple of reasons:

1. The data science team in question might want to work with DVC, but they are blocked from doing so, either just politically or even with automatically enforced authorizations
2. DVC can break backwards compatibility, and different projects in the monorepo might use different versions of DVC

I could probably think of a couple other reasons why this might be a problem.

It seems to me that requiring the `.dvc` folder be in the root folder is pretty arbitrary, and giving the option to put it in other places in the tree would open the way for wider adoption.

pared:

Fixes #2349

### DIFF
--- a/dvc/command/init.py
+++ b/dvc/command/init.py
@@ -15,7 +15,10 @@ class CmdInit(CmdBaseNoRepo):
 
         try:
             self.repo = Repo.init(
-                ".", no_scm=self.args.no_scm, force=self.args.force
+                ".",
+                no_scm=self.args.no_scm,
+                force=self.args.force,
+                subdir=self.args.subdir,
             )
             self.config = self.repo.config
         except InitError:
@@ -54,6 +57,15 @@ def add_parser(subparsers, parent_parser):
         help=(
             "Overwrite existing '.dvc/' directory. "
             "This operation removes local cache."
+        ),
+    )
+    init_parser.add_argument(
+        "--subdir",
+        action="store_true",
+        default=False,
+        help=(
+            "Necessary for running this command inside a subdirectory of a "
+            "parent SCM repository."
         ),
     )
     init_parser.set_defaults(func=CmdInit)
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -115,6 +115,7 @@ SCHEMA = {
         Optional("interactive", default=False): Bool,
         Optional("analytics", default=True): Bool,
         Optional("hardlink_lock", default=False): Bool,
+        Optional("no_scm", default=False): Bool,
     },
     "cache": {
         "local": str,
--- a/dvc/ignore.py
+++ b/dvc/ignore.py
@@ -71,10 +71,25 @@ class DvcIgnoreDirs(DvcIgnore):
         return self.basenames == other.basenames
 
 
+class DvcIgnoreRepo(DvcIgnore):
+    def __call__(self, root, dirs, files):
+        def is_dvc_repo(directory):
+            from dvc.repo import Repo
+
+            return os.path.isdir(os.path.join(directory, Repo.DVC_DIR))
+
+        dirs = [d for d in dirs if not is_dvc_repo(d)]
+
+        return dirs, files
+
+
 class DvcIgnoreFilter(object):
     def __init__(self, tree):
         self.tree = tree
-        self.ignores = {DvcIgnoreDirs([".git", ".hg", ".dvc"])}
+        self.ignores = {
+            DvcIgnoreDirs([".git", ".hg", ".dvc"]),
+            DvcIgnoreRepo(),
+        }
         for root, dirs, files in self.tree.walk(self.tree.tree_root):
             self._update(root)
             dirs[:], files[:] = self(root, dirs, files)
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -78,7 +78,8 @@ class Repo(object):
 
         self.config = Config(self.dvc_dir)
 
-        self.scm = SCM(self.root_dir)
+        no_scm = self.config["core"].get("no_scm", False)
+        self.scm = SCM(self.root_dir, no_scm=no_scm)
 
         self.tree = WorkingTree(self.root_dir)
 
@@ -146,10 +147,10 @@ class Repo(object):
         return os.path.join(root_dir, cls.DVC_DIR)
 
     @staticmethod
-    def init(root_dir=os.curdir, no_scm=False, force=False):
+    def init(root_dir=os.curdir, no_scm=False, force=False, subdir=False):
         from dvc.repo.init import init
 
-        init(root_dir=root_dir, no_scm=no_scm, force=force)
+        init(root_dir=root_dir, no_scm=no_scm, force=force, subdir=subdir)
         return Repo(root_dir)
 
     def unprotect(self, target):
--- a/dvc/repo/init.py
+++ b/dvc/repo/init.py
@@ -7,8 +7,8 @@ from dvc import analytics
 from dvc.config import Config
 from dvc.exceptions import InitError
 from dvc.repo import Repo
-from dvc.scm import NoSCM
 from dvc.scm import SCM
+from dvc.scm.base import SCMError
 from dvc.utils import boxify
 from dvc.utils import relpath
 from dvc.utils.fs import remove
@@ -44,7 +44,7 @@ def _welcome_message():
     logger.info(msg)
 
 
-def init(root_dir=os.curdir, no_scm=False, force=False):
+def init(root_dir=os.curdir, no_scm=False, force=False, subdir=False):
     """
     Creates an empty repo on the given directory -- basically a
     `.dvc` directory with subdirectories for configuration and cache.
@@ -63,15 +63,23 @@ def init(root_dir=os.curdir, no_scm=False, force=False):
     Raises:
         KeyError: Raises an exception.
     """
+
+    if no_scm and subdir:
+        raise InitError(
+            "Cannot initialize repo with `--no-scm` and `--subdir`"
+        )
+
     root_dir = os.path.realpath(root_dir)
     dvc_dir = os.path.join(root_dir, Repo.DVC_DIR)
-    scm = SCM(root_dir)
-    if isinstance(scm, NoSCM) and not no_scm:
+
+    try:
+        scm = SCM(root_dir, search_parent_directories=subdir, no_scm=no_scm)
+    except SCMError:
         raise InitError(
-            "{repo} is not tracked by any supported scm tool (e.g. git). "
-            "Use `--no-scm` if you don't want to use any scm.".format(
-                repo=root_dir
-            )
+            "{repo} is not tracked by any supported SCM tool (e.g. Git). "
+            "Use `--no-scm` if you don't want to use any SCM or "
+            "`--subdir` if initializing inside a subdirectory of a parent SCM "
+            "repository.".format(repo=root_dir)
         )
 
     if os.path.isdir(dvc_dir):
@@ -87,6 +95,11 @@ def init(root_dir=os.curdir, no_scm=False, force=False):
     os.mkdir(dvc_dir)
 
     config = Config.init(dvc_dir)
+
+    if no_scm:
+        with config.edit() as conf:
+            conf["core"]["no_scm"] = True
+
     proj = Repo(root_dir)
 
     scm.add([config.files["repo"]])
--- a/dvc/scm/__init__.py
+++ b/dvc/scm/__init__.py
@@ -10,18 +10,23 @@ class NoSCM(Base):
     pass
 
 
-def SCM(root_dir):  # pylint: disable=invalid-name
+def SCM(
+    root_dir, search_parent_directories=True, no_scm=False
+):  # pylint: disable=invalid-name
     """Returns SCM instance that corresponds to a repo at the specified
     path.
 
     Args:
         root_dir (str): path to a root directory of the repo.
-        repo (dvc.repo.Repo): DVC repo instance that root_dir belongs to.
+        search_parent_directories (bool): whether to look for repo root in
+        parent directories.
+        no_scm (bool): return NoSCM if True.
 
     Returns:
         dvc.scm.base.Base: SCM instance.
     """
-    if Git.is_repo(root_dir) or Git.is_submodule(root_dir):
-        return Git(root_dir)
 
-    return NoSCM(root_dir)
+    if no_scm:
+        return NoSCM(root_dir)
+
+    return Git(root_dir, search_parent_directories=search_parent_directories)
--- a/tests/func/test_init.py
+++ b/tests/func/test_init.py
@@ -1,5 +1,8 @@
+import logging
 import os
 
+from dvc.compat import fspath
+from dvc.config import Config
 from dvc.exceptions import InitError
 from dvc.main import main
 from dvc.repo import Repo as DvcRepo
@@ -49,21 +52,20 @@ class TestInitNoSCMFail(TestDir):
         self.assertNotEqual(ret, 0)
 
 
-class TestInitNoSCM(TestDir):
-    def _test_init(self):
-        self.assertTrue(os.path.exists(DvcRepo.DVC_DIR))
-        self.assertTrue(os.path.isdir(DvcRepo.DVC_DIR))
+def test_init_no_scm_api(tmp_dir):
+    repo = DvcRepo.init(no_scm=True)
 
-    def test_api(self):
-        DvcRepo.init(no_scm=True)
+    assert (tmp_dir / DvcRepo.DVC_DIR).is_dir()
+    assert repo.config["core"]["no_scm"]
 
-        self._test_init()
 
-    def test_cli(self):
-        ret = main(["init", "--no-scm"])
-        self.assertEqual(ret, 0)
+def test_init_no_scm_cli(tmp_dir):
+    ret = main(["init", "--no-scm"])
+    assert ret == 0
 
-        self._test_init()
+    dvc_path = tmp_dir / DvcRepo.DVC_DIR
+    assert dvc_path.is_dir()
+    assert Config(fspath(dvc_path))["core"]["no_scm"]
 
 
 def test_init_quiet_should_not_display_welcome_screen(tmp_dir, scm, caplog):
@@ -71,3 +73,32 @@ def test_init_quiet_should_not_display_welcome_screen(tmp_dir, scm, caplog):
 
     assert 0 == ret
     assert "" == caplog.text
+
+
+def test_allow_init_dvc_subdir(tmp_dir, scm, monkeypatch):
+    tmp_dir.gen({"subdir": {}})
+
+    with monkeypatch.context() as m:
+        m.chdir("subdir")
+        assert main(["init", "--subdir"]) == 0
+
+    repo = DvcRepo("subdir")
+    assert repo.root_dir == fspath(tmp_dir / "subdir")
+    assert repo.scm.root_dir == fspath(tmp_dir)
+
+
+def test_subdir_init_no_option(tmp_dir, scm, monkeypatch, caplog):
+    tmp_dir.gen({"subdir": {}})
+
+    caplog.clear()
+    with monkeypatch.context() as m:
+        m.chdir("subdir")
+        with caplog.at_level(logging.ERROR, logger="dvc"):
+            assert main(["init"]) == 1
+
+    assert (
+        "{} is not tracked by any supported SCM tool (e.g. Git). "
+        "Use `--no-scm` if you don't want to use any SCM or "
+        "`--subdir` if initializing inside a subdirectory of a parent SCM "
+        "repository.".format(fspath(tmp_dir / "subdir"))
+    ) in caplog.text
--- a/tests/func/test_stage.py
+++ b/tests/func/test_stage.py
@@ -5,6 +5,7 @@ import tempfile
 from dvc.main import main
 from dvc.output.local import OutputLOCAL
 from dvc.remote.local import RemoteLOCAL
+from dvc.repo import Repo
 from dvc.stage import Stage
 from dvc.stage import StageFileFormatError
 from dvc.utils.stage import dump_stage_file
@@ -179,3 +180,19 @@ def test_meta_is_preserved(tmp_dir, dvc):
 
     new_data = load_stage_file(stage.path)
     assert new_data["meta"] == data["meta"]
+
+
+def test_parent_repo_collect_stages(tmp_dir, scm, dvc):
+    tmp_dir.gen({"subdir": {}})
+    subrepo_dir = tmp_dir / "subdir"
+
+    with subrepo_dir.chdir():
+        subrepo = Repo.init(subdir=True)
+        subrepo_dir.gen("subrepo_file", "subrepo file content")
+        subrepo.add("subrepo_file")
+
+    stages = dvc.collect(None)
+    subrepo_stages = subrepo.collect(None)
+
+    assert stages == []
+    assert subrepo_stages != []
--- a/dvc/scm/base.py
+++ b/dvc/scm/base.py
@@ -28,7 +28,11 @@ class Base(object):
     """Base class for source control management driver implementations."""
 
     def __init__(self, root_dir=os.curdir):
-        self.root_dir = os.path.realpath(root_dir)
+        self._root_dir = os.path.realpath(root_dir)
+
+    @property
+    def root_dir(self):
+        return self._root_dir
 
     def __repr__(self):
         return "{class_name}: '{directory}'".format(
--- a/dvc/scm/git/__init__.py
+++ b/dvc/scm/git/__init__.py
@@ -22,7 +22,7 @@ class Git(Base):
     GITIGNORE = ".gitignore"
     GIT_DIR = ".git"
 
-    def __init__(self, root_dir=os.curdir):
+    def __init__(self, root_dir=os.curdir, search_parent_directories=True):
         """Git class constructor.
         Requires `Repo` class from `git` module (from gitpython package).
         """
@@ -32,10 +32,12 @@ class Git(Base):
         from git.exc import InvalidGitRepositoryError
 
         try:
-            self.repo = git.Repo(self.root_dir)
+            self.repo = git.Repo(
+                root_dir, search_parent_directories=search_parent_directories
+            )
         except InvalidGitRepositoryError:
             msg = "{} is not a git repository"
-            raise SCMError(msg.format(self.root_dir))
+            raise SCMError(msg.format(root_dir))
 
         # NOTE: fixing LD_LIBRARY_PATH for binary built by PyInstaller.
         # http://pyinstaller.readthedocs.io/en/stable/runtime-information.html
@@ -45,6 +47,10 @@ class Git(Base):
 
         self.ignored_paths = []
         self.files_to_track = set()
+
+    @property
+    def root_dir(self):
+        return self.repo.working_tree_dir
 
     @staticmethod
     def clone(url, to_path, rev=None):
@@ -94,14 +100,6 @@ class Git(Base):
         import git
 
         return rev and git.Repo.re_hexsha_shortened.search(rev)
-
-    @staticmethod
-    def is_repo(root_dir):
-        return os.path.isdir(Git._get_git_dir(root_dir))
-
-    @staticmethod
-    def is_submodule(root_dir):
-        return os.path.isfile(Git._get_git_dir(root_dir))
 
     @staticmethod
     def _get_git_dir(root_dir):
--- a/tests/dir_helpers.py
+++ b/tests/dir_helpers.py
@@ -99,7 +99,9 @@ class TmpDir(pathlib.Path):
         if scm:
             _git_init(str_path)
         if dvc:
-            self.dvc = Repo.init(str_path, no_scm=True)
+            self.dvc = Repo.init(
+                str_path, no_scm=not scm and not hasattr(self, "scm")
+            )
         if scm:
             self.scm = self.dvc.scm if hasattr(self, "dvc") else Git(str_path)
         if dvc and hasattr(self, "scm"):
--- a/tests/func/test_ignore.py
+++ b/tests/func/test_ignore.py
@@ -4,7 +4,12 @@ import shutil
 import pytest
 
 from dvc.exceptions import DvcIgnoreInCollectedDirError
-from dvc.ignore import DvcIgnore, DvcIgnoreDirs, DvcIgnorePatterns
+from dvc.ignore import (
+    DvcIgnore,
+    DvcIgnoreDirs,
+    DvcIgnorePatterns,
+    DvcIgnoreRepo,
+)
 from dvc.scm.tree import WorkingTree
 from dvc.utils import relpath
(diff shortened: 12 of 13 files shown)


dvc/config.py
```diff
@@ -109,6 +109,7 @@ class Config(object):  # pylint: disable=too-many-instance-attributes
     SECTION_CORE_ANALYTICS = "analytics"
     SECTION_CORE_CHECKSUM_JOBS = "checksum_jobs"
     SECTION_CORE_HARDLINK_LOCK = "hardlink_lock"
+    SECTION_CORE_MULTIPLE_ROOTS = "allow_multiple_roots"
```
efiop: So this one is set in a subrepo, right? If so the naming is confusing. Maybe `subrepo = true`?

efiop: One of the reasons why I'm asking about this particular thingy is because it might be linked to an imperfect understanding of what we are doing here in general :slightly_smiling_face:

pared:
@dmpetrov I noticed that in #2706 you mentioned that managing subrepos from "master" repo might be a viable use case. In this issue, I introduced `DvcIgnore` that would ignore subrepos when working from master repo. Do you think we should allow `dvc pull` and other "data sync" commands to work for subrepos?

dvc/command/init.py
```diff
@@ -56,4 +59,13 @@ def add_parser(subparsers, parent_parser):
             "This operation removes local cache."
         ),
     )
+    init_parser.add_argument(
+        "--sub",
```
efiop: How about
```suggestion
        "--subrepo",
```
then to match config?

dvc/repo/__init__.py
```diff
@@ -80,7 +80,16 @@ def __init__(self, root_dir=None):
...
+        if self.config.config.get(Config.SECTION_CORE, {}).get(
+            Config.SECTION_CORE_SUBREPO, False
+        ):
+            from dvc.repo.init import find_scm_dir
+
+            scm_root_dir = find_scm_dir(self.root_dir)
+        else:
+            scm_root_dir = self.root_dir
+
+        self.scm = SCM(scm_root_dir)
```
efiop: Maybe we could pass `subrepo` or some flag to SCM and let it find the scm root? Feels like you are doing scm-related logic here for no good reason.

dvc/repo/init.py
```diff
@@ -44,7 +44,17 @@ def _welcome_message():
...
+def find_scm_dir(path):
+    from git import Repo
+    from git.exc import InvalidGitRepositoryError
+
+    try:
+        return Repo(path, search_parent_directories=True).working_tree_dir
+    except InvalidGitRepositoryError:
+        raise InitError("Could not find Git repository in parent dirs")
+
+
```
efiop: It is really weird that you put scm-related logic in repo/init.py. Also, you call it `find_scm_dir`, but you are only looking for git repo. Maybe we could move this logic to scm/ ?

dvc/repo/init.py
```diff
@@ -63,9 +73,19 @@ def init(root_dir=os.curdir, no_scm=False, force=False):
...
     root_dir = os.path.realpath(root_dir)
+
+    if not subrepo:
+        scm_root_dir = root_dir
+    else:
+        scm_root_dir = find_scm_dir(root_dir)
+
     dvc_dir = os.path.join(root_dir, Repo.DVC_DIR)
-    scm = SCM(root_dir)
+    scm = SCM(scm_root_dir)
```
efiop: See, you are spreading the context here again, I think it could be moved nicely to scm/, which will also avoid code duplication that you have here and in `Repo.__init__`

tests/func/test_add.py
```diff
@@ -641,3 +641,32 @@ def test_escape_gitignore_entries(tmp_dir, scm, dvc):
...
+
+def test_parent_repo_pull(tmp_dir, scm, dvc, subrepo):
+    with subrepo.chdir():
+        subrepo.dvc_gen("file", "file content")
+        subrepo.dvc.push()
+        (subrepo / "file").unlink()
+
+    dvc.pull()
+
+    assert not (subrepo / "file").exists()
```
efiop: This test doesn't really belong to test_add.py, right? Maybe replace it with a `Repo` test that makes sure that subrepo's dvc-files are not collected. That should be a much more local way of testing it.

dvc/command/init.py
pared: I agree, though I would stay with `subrepo` for user convenience

efiop: @jorgeorpinel Would really appreciate your thoughts :slightly_smiling_face:

dvc/ignore.py
```diff
@@ -71,10 +71,25 @@ def __eq__(self, other):
         return self.basenames == other.basenames
 
 
+class DvcIgnoreRepo(DvcIgnore):
```
efiop: Would be interesting to see perf comparison with and without this change when you have giant directories in your repo.

dvc/scm/__init__.py
```diff
@@ -10,7 +10,8 @@ class NoSCM(Base):
     pass
 
 
-def SCM(root_dir):  # pylint: disable=invalid-name
+def SCM(root_dir, search_parent_directories=False):  # pylint:
+    # disable=invalid-name
```
efiop: Looks like `black` decided to split it or was it you? :slightly_smiling_face: Not sure if this works after it is split.

dvc/command/init.py
jorgeorpinel: Thanks for asking! I would avoid "repo" if it doesn't imply a Git repo. We use the term "DVC project" more generally. `--subproject`? Or copy Git (`submodule`/`subtree`)

Why do you need to use an option though? Can't DVC recognize it's already inside a DVC project? 🙂

skshetry:
`get` and `import` won't work for the dvc subrepo, right? (as it's Git-repo based, not dvc-repo based, see: #3277).

dvc/scm/__init__.py
pared: it was black, sorry

dvc/command/init.py
efiop: worried about `submodule` being confused with git submodules :slightly_frowning_face: `--subproject` seems alright, though not ideal too... Maybe there is a way to emphasize that this is option is to just tell dvc to use parent dir git...

> Why do you need to use an option though? Can't DVC recognize it's already inside a DVC project? slightly_smiling_face

Sure it can, but it can't understand that it needs to grab a git from an outside directory, this needs to be explicit.

This might be a point where `--no-scm` comes to bite us in the bum. If we would've made it explcit like `dvc config core.scm false`, then we could totally automatically use parent git here. Though that might also be error-prone, as some unsuspecting user might run it in one of his subdirs accidentally...

jorgeorpinel: OK. Not sure I understood completely, but anyway: if this option only applies to Git enabled DVC projects then `--subrepo` is a good name.

dvc/ignore.py
efiop: @pared ^^

dvc/config.py
```diff
@@ -109,6 +109,7 @@ class Config(object):  # pylint: disable=too-many-instance-attributes
     SECTION_CORE_ANALYTICS = "analytics"
     SECTION_CORE_CHECKSUM_JOBS = "checksum_jobs"
     SECTION_CORE_HARDLINK_LOCK = "hardlink_lock"
+    SECTION_CORE_SUBREPO = "subrepo"
```
dmpetrov: Do we really need this "signal" in the config? @pared I'd appreciate if you could explain?

dvc/command/init.py
```diff
@@ -56,4 +59,13 @@ def add_parser(subparsers, parent_parser):
             "This operation removes local cache."
         ),
     )
+    init_parser.add_argument(
+        "--subrepo",
+        action="store_true",
+        default=False,
+        help=(
+            "Use this flag if created repository is below Git repository"
+            "in tree."
```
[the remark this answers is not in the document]
pared: removed this option

dvc/config.py
pared: I needed it at the beginning, but thanks to @efiop review it got obsolete, removing

dvc/command/init.py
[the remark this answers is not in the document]
dmpetrov: hm... The option still might be needed. I think about subrepo as a not a usual scenario that should not be supported by default.

I'd prefer this behavior:
```
$ cd myrepo/
$ mkdir new_repo
$ dvc init
ERROR: failed to initiate DVC - /Users/dmitry/src/myrepo/new_repo is not tracked by any supported scm tool (e.g. git). Use `--no-scm` if you don't want to use any scm or `--subrepo` if sub-repository is needed.  # <-- see the last part of the message
$ dvc init --subrepo # <-- creates subrepo
```

dmpetrov:
> Do you think we should allow `dvc pull` and other "data sync" commands to work for subrepos?

@pared No, commands should not go to nested repos?

PS: we can even think about not supporting nested repos if it simplifies implementation or docs.

dvc/ignore.py
pared: Little summary of performed tests:
created directory with 10 directories, each having 10 directories and so on.
The tree had 6 such levels so alltogether at the bottom there were 1M dirs.
Adding new Ignore filter increased stages gathering time from `7m 58s` to `8m 15s` which is roughly 3%.  It is not incosiderable, but I would say that in case of sucha big project user will probably want to leverage the `.dvcignore` functionality to reduce this time by omitting different dirs.

dvc/command/init.py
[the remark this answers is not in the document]
pared: should be behaving as described now, added test for that case

dvc/repo/__init__.py
```diff
@@ -80,7 +80,7 @@ def __init__(self, root_dir=None):
 
         self.config = Config(self.dvc_dir)
 
-        self.scm = SCM(self.root_dir)
+        self.scm = SCM(root_dir, search_parent_directories=True)
```
efiop: Hm, but what if I want to have no-scm repo? I've mentioned that if we get rid of subrepo config option, we might need it for no-scm repo, to clearly mark that we shouldn't try to look for scm's in parent dirs. Or I am missing something?

efiop: It is a pretty synthetic use case, but still is possible. Just wondering if we should switch to a new behavior and if we need to support no-scm in git host repo.

pared: Ok, yes I did not noticed that.
So it seems that we have 2 options here:
1. Make config option for subrepo which will allow searching parent directories - no problems besides cluttering the config with additional option
2. Make no-scm config option for subrepo - that will break backwards compatibility with no-scm repos created before this change

pared:
@dmpetrov ok, so in the end some kind of flag will have to be introduced to config (take note of this discussion: https://github.com/iterative/dvc/pull/3257#discussion_r378126875)

I think we should go with the second option I mentioned there, which is introducing `no-scm` config option when `--no-scm` is used. That will break backward compatibility for existing `no-scm` repos
but the fix will be relatively easy (writing no-scm=True in config). Basing on user activity on discord and in issues, I would not expect too much of such cases.
Having subrepo as default behavior makes more sense to me, since this subject is recurring.

efiop:
@pared Also something to note about no-scm, is that it will require changing our tests, because base `dvc` fixture doesn't have git in it by default, it is `init`-ed with `no_scm=True` and then when git is applied on top.

dvc/repo/__init__.py
dmpetrov: @pared It is fine to keep this conf option. When I commented on this I didn't consider `--no-scm` case. Also, as a 3rd option, you can prohibit using `--subrepo` in not Git case. It ups to you.

pared: `no-scm` will be stored as config option and read here, by default we will look for Git in parent dirs

tests/func/test_init.py
```diff
@@ -1,8 +1,11 @@
+import logging
 import os
 
+from dvc.compat import fspath
+from dvc.config import Config
 from dvc.exceptions import InitError
 from dvc.main import main
-from dvc.repo import Repo as DvcRepo
+from dvc.repo import Repo as DvcRepo, Repo
```
efiop:
```suggestion
from dvc.repo import Repo as DvcRepo
```
? Kudos DS for catching this one :slightly_smiling_face:

tests/func/test_stage.py
```diff
@@ -178,3 +178,14 @@ def test_meta_is_preserved(tmp_dir, dvc):
 
     new_data = load_stage_file(stage.path)
     assert new_data["meta"] == data["meta"]
+
+
+def test_parent_repo_collect_stages(tmp_dir, scm, dvc, subrepo_dir):
```
efiop: Looks like this is the only case where we are using `subrepo_dir` fixture,right?  So maybe let's not mess with current fixtures then, and just create what we need here?

tests/func/test_init.py
pared: should have paid more attention to DS, sorry

dvc/command/init.py
```diff
@@ -56,4 +59,13 @@ def add_parser(subparsers, parent_parser):
             "This operation removes local cache."
         ),
     )
+    init_parser.add_argument(
+        "--subrepo",
+        action="store_true",
+        default=False,
+        help=(
+            "Use this flag if created repository is below Git repository"
+            " in tree."
```
efiop: Please move the space above, for consistency with other messages.

efiop: @jorgeorpinel Could you please take a look at this help message? :pray:

dvc/command/init.py
```diff
@@ -56,4 +59,13 @@ def add_parser(subparsers, parent_parser):
             "This operation removes local cache."
         ),
     )
+    init_parser.add_argument(
+        "--subrepo",
```
jorgeorpinel: I read #2349 again and I'm no longer sure this is an appropriate name. Are we talking about running `dvc init` in subdirectories of a Git repo, which are NOT Git subrepos themselves (no `git init` is run in them) ?

If that's the case and they are not Git subrepos (just subdirs), then `--subrepo` seems quite confusing. Maybe `--subdir`, `--subproject`, or `--subdvc`.

dvc/command/init.py
jorgeorpinel: Continued in https://github.com/iterative/dvc/pull/3257. Resolving here.

dvc/command/init.py
jorgeorpinel: Maybe `Necessary for running this command in a subdirectory of an existing DVC repository.`

jorgeorpinel: Maybe `Necessary for running this command inside a subdirectory of a parent DVC repository.`

Will check other strings... Thanks

dvc/repo/init.py
```diff
@@ -63,15 +63,21 @@ def init(root_dir=os.curdir, no_scm=False, force=False):
     Raises:
         KeyError: Raises an exception.
     """
+
+    if no_scm and subrepo:
+        raise InitError(
+            "Cannot initialize repo with `--no-scm` and `--subrepo`"
```
jorgeorpinel:
```suggestion
            "Cannot use `--subrepo` when initializing a `--no-scm`  DVC project."
```

dvc/repo/init.py
```diff
@@ -63,15 +63,21 @@ def init(root_dir=os.curdir, no_scm=False, force=False):
...
+
+    scm = SCM(root_dir, search_parent_directories=subrepo)
     if isinstance(scm, NoSCM) and not no_scm:
         raise InitError(
             "{repo} is not tracked by any supported scm tool (e.g. git). "
-            "Use `--no-scm` if you don't want to use any scm.".format(
-                repo=root_dir
-            )
+            "Use `--no-scm` if you don't want to use any scm or "
+            "`--subrepo` if sub-repository is needed.".format(repo=root_dir)
```
jorgeorpinel: Something more descriptive like:
```
"'{repo}' is not tracked by Git."
"Use `--no-scm` if you don't want to use Git"
" or `--subrepo` if initializing inside a subdirectory of a parent DVC repository."
```
> Obvi rewarp

tests/func/test_init.py
```diff
@@ -49,25 +52,54 @@ def test_cli(self):
...
+    caplog.clear()
+    with monkeypatch.context() as m:
+        m.chdir("subdir")
+        with caplog.at_level(logging.ERROR, logger="dvc"):
+            assert main(["init"]) == 1
+
+    assert (
+        "{} is not tracked by any supported scm tool (e.g. git). "
+        "Use `--no-scm` if you don't want to use any scm "
+        "or `--subrepo` if sub-repository is needed.".format(
```
jorgeorpinel: (Same)

dvc/repo/init.py
jorgeorpinel: Note: I thought we decided previously to just talk about Git in user output since we only support that, even when code is `scm` but if I'm wrong, just use "SCM" or "a SCM (e.g. Git)", etc. instead of "Git". Please use proper capitalization. Thanks

dvc/command/init.py
jorgeorpinel: I would go for `--subdir` cc @efiop @Suor

efiop: @jorgeorpinel `--subdir` sounds good to me too! @pared What do you think?

pared: I like it, ill refactor.

dvc/scm/__init__.py
```diff
@@ -21,7 +23,12 @@ def SCM(root_dir):  # pylint: disable=invalid-name
...
-    if Git.is_repo(root_dir) or Git.is_submodule(root_dir):
-        return Git(root_dir)
 
-    return NoSCM(root_dir)
+    try:
+        scm = Git(
+            root_dir, search_parent_directories=search_parent_directories
+        )
+    except SCMError:
+        scm = NoSCM(root_dir)
```
efiop: If we explicitly have `no-scm = True` in our config, we probably shouldn't try to search for git, as that might defeat the purpose :)

efiop: So maybe we might use NoSCM directly in `Repo.__init__` if `no-scm` instead of going through `SCM` wrapper?

efiop: Seems like we can `search_parent_directories=True` by default (in Git class itself) and replace it with `no_scm` flag here. Might be missing something, but right now I see that it is only used for that purpose but in an obscure way, so I suppose these are the leftovers from the previous implementation, that you didn't reconsider.

pared: I see one problem here: what about `Repo.init` method? If we will rely on `no_scm` there, then there is no point of having `--subdir` flag.
I believe it should stay as is because in `Repo.init` we have 3 situations to handle (and thats why we need 2 flags):
1. no-scm=True, subdir=False -> succesfull init
2. no-scm=False, subdir=False + NoSCM -> raise error
3. no-scm=False, subdir=True -> search parent dirs

efiop: @pared Agreed on that. So `search_parent_directories=False` should be used in `repo.init()` if `no_scm=False` and `subdir=False`, right? In all other cases it should be `search_parent_directories=True`, right?

dvc/config.py
```diff
@@ -108,6 +108,7 @@ class RelPath(str):
         Optional("interactive", default=False): Bool,
         Optional("analytics", default=True): Bool,
         Optional("hardlink_lock", default=False): Bool,
+        Optional("no-scm", default=False): Bool,
```
efiop:
```suggestion
        Optional("no_scm", default=False): Bool,
```
to keep it consistent with the rest.

dvc/repo/init.py
```diff
@@ -63,15 +63,22 @@ def init(root_dir=os.curdir, no_scm=False, force=False):
     Raises:
         KeyError: Raises an exception.
     """
+
+    if no_scm and subdir:
+        raise InitError(
+            "Cannot initialize repo with `--no-scm` and `--subdir`"
+        )
```
jorgeorpinel: OK I'm still a bit confused about the limitations here. Why can't a subdir project be no-scm?

What if the parent is no-scm? And i that's not allowed, where is the check? Thanks!

pared: @jorgeorpinel 
In case of initializing `--subdir`, we will make Git looks for git root inside parent directories,
so if we have this flag, and there is no parent scm, git will throw `InvalidRepositoryError`, which in result will make us initialize NoSCM, and since there is not `--no-scm` flag, it will raise InitError. 
sample script, working with current version of change:

```
#!/bin/bash

rm -rf repo
mkdir repo

pushd repo
mkdir subrepo
pushd subrepo

dvc init --subdir
```

dvc/scm/__init__.py
pared: @efiop that is correct

pared: @efiop, now logic related to choosing SCM is in `Repo.__init__`. Now I am not sure whether we need `SCM` method at all. I believe we could refactor `init` command to not use it. Do you think its still the scope of this task?

efiop:
@pared Looks good! How about docs though? Could you please create a doc PR before we merge this?

dvc/repo/__init__.py
```diff
@@ -78,7 +79,11 @@ def __init__(self, root_dir=None):
 
         self.config = Config(self.dvc_dir)
 
-        self.scm = SCM(self.root_dir)
+        no_scm = self.config["core"].get("no_scm", False)
+        if no_scm:
+            self.scm = NoSCM()
```
efiop:
```suggestion
            self.scm = NoSCM(self.root_dir)
```

dvc/repo/__init__.py
```diff
@@ -78,7 +79,11 @@ def __init__(self, root_dir=None):
...
+        else:
+            self.scm = SCM(root_dir)
```
efiop: So if there is no `no-scm=True` and no git, `SCM` will fallback to `NoSCM` silently, which doesnt look right.

dvc/repo/__init__.py
```diff
@@ -78,7 +78,8 @@ def __init__(self, root_dir=None):
 
         self.config = Config(self.dvc_dir)
 
-        self.scm = SCM(self.root_dir)
+        no_scm = self.config["core"].get("no_scm", False)
+        self.scm = SCM(root_dir, not no_scm, no_scm)
```
efiop: Please don't lose `self.root_dir`, it is not the same as `root_dir`.
```suggestion
        self.scm = SCM(self.root_dir, not no_scm, no_scm)
```

efiop: Also, do we need to pass `not no_scm, no_scm`?

pared: Fixed, that was unnecessary.

dvc/repo/__init__.py
pared: Fixed, SCM is handling this now. If no_scm is True and there is no git, we will get an exception.

dvc/repo/init.py
efiop: @pared Sorry, but I didn't get your explanation. Since `no-scm` is explicit now, `--no-scm --subdir` should be a totally valid case. Unless I'm missing something.

pared: @efiop well my explanation was valid when @jorgeorpinel asked the question. After our discussion on how to handle SCM, and changing the logic, it is technically possible to do that. Still, I would leave this check as using both flags at the same time indicates a misunderstanding of how they work.

pared:
https://github.com/iterative/dvc.org/pull/1022

dvc/ignore.py
```diff
@@ -71,10 +71,25 @@ def __eq__(self, other):
...
+
+        return dirs, files
+
+
 class DvcIgnoreFilter(object):
     def __init__(self, tree):
         self.tree = tree
-        self.ignores = {DvcIgnoreDirs([".git", ".hg", ".dvc"])}
+        self.ignores = {
+            DvcIgnoreDirs([".git", ".hg", ".dvc"]),
```
jorgeorpinel: .hg ? Do we support Mercurial somehow now??? Or is this legacy stuff? Thanks

skshetry: We don't want to traverse through `.git` or `.hg` folders. People might use `hg` for code versioning and use `--no-scm` with DVC.

jorgeorpinel: Hm... Why only do this for .hg? There's many many possible files or dirs the user may want to gitignore and they can just add it to .gitignore/ or .git/info/exclude

dvc/repo/init.py
jorgeorpinel: Yeah I guess it doesn't make sense to have a --subdir DVC project that does not put outputs into .gitignore if the parent DVC repo does.

Although... We do allow `dvc init --no-scm` in Git-tracked directories, so the enforced restriction is kind of inconsistent.

Same discussion as https://github.com/iterative/dvc/issues/2472#issuecomment-590184536 I guess...

jorgeorpinel: But I do agree with this restriction. I just think it should also be applied to `dvc init --no-scm` in Git-tracked dirs. See https://github.com/iterative/dvc/issues/2472#issuecomment-590184536

dvc/ignore.py
efiop: @jorgeorpinel add `.hg` to `.gitignore`? Maybe I'm missing something.

An alternative would be to add `.hg` to `.dvcignore` file, yes, but since we know about `.hg`, we are skipping it automatically.

jorgeorpinel: > add .hg to .gitignore?

Haha sorry I wasn't thinking right 😋 

> to add .hg to .dvcignore

Yeah, this! Since we don't support Mercurial in general why treat it specially? I'd let the user handle it, like he could want dvcignore any other files or directories for a bunch of reasons.

skshetry: > Since we don't support Mercurial in general why treat it specially?

Why traverse through such directories at all? It'll only make dvc slower. It's a reasonable default.

jorgeorpinel: > such directories 

Exactly. There can be tons of special/hidden dirs you don't want DVC to traverse (.idea/ .project/ .env/ .ds_store/ venv/), and you can just add them to .gitignore or .dvcignore. Why are we secretly integrating with Mercurial if we don't actually support it? What if someone wants to version a .hg dir? (edge case I know)

It just seems very arbitrary to me, but OK maybe I'm missing something...

shcheklein: My 2cs :) Not important thing, but I also don't understand why do we ignore some `.hg` dir which does not mean anything to us. It is some magic behavior not related to the product features in any way. The good example would be - does Git ignore .hg by default?

skshetry: > does Git ignore .hg by default?

No, it doesn't. And, why should it? It's not meant to work with other version controls. whereas, dvc needs to work with `git` and any other version controls out there. Other stuffs like `__pycache__`, `venv` can be assumed to be ignored in `.gitignore` or `.dvcignore`.
`.hg` is really a special case.

Do people really use `.hg` folders for their work? Unless they do (which they shouldn't), benefits outweigh the drawbacks.

shcheklein: @skshetry 

>  and any other version controls out there

that would be true if DVC worked with hg, right? It does not right now, and we don't have any plans do this.

> Do people really use .hg folders for their work? Unless they do (which they shouldn't), benefits outweigh the drawbacks.

it's a slippery slope. This logic applies easily to all other aforementioned cases like DSStore, etc. That's why there is a mechanism (and no list of defaults in Git for this).

Another example (if you don't like git itself) is `git-lfs`. DVC with its data management layer is kind-a the same extension to Git as git-lfs. I doubt that git-lfs, or git-annex introduce a list of defaults on top.

shcheklein: I just don't really see any difference between __pycache__, venv and .hg, really. I would say even including __pycache__ would even more useful by default.

.git and .dvc feel like the only special cases for DVC right now.

skshetry: > git-lfs

It only works on top of git. Not alongside `git` as DVC does. So, it makes sense there.

DVC has `--no-scm`, so they could be using any version control out there or none. The other one except of `git` that's popular is `hg`, so, I think it makes sense to ignore by default.

And, how many people are using or are aware of `.dvcignore`?

shcheklein: > And, how many people are using or are aware of .dvcignore?

yep, this applies to all other example mentioned above we should ignore then? A lot of them make even more value.

> ... that's popular is hg

perforce and svn have a bigger market share? I doubt that mercurial has more than a few %% (which is sad).

--------------

Interesting discussion. I realized that there might be indeed a stronger motivation to include more defaults (`.hg` is still strange and not the most painful example) into DVC ignore list. The difference between them is that for DVC it **affects performance silently** (imagine analyzing `.node_modules`), for Git it affects `git status` and **you immediately see that** this is wrong, and start looking for `.gitingnore`, etc. You won't have a motivation do this with DVC - I mean motivation to start looking for `.dvcignore` unless you need to actually ignore some files doing `dvc add`?

Does it make sense to include all files that are in `.gitignore` into `.dvcignore` ? (may be we already do this?) Does it makes sense to actually provide a bigger list of defaults?

jorgeorpinel: > DVC has --no-scm, so they could be using any version control out there

I don't think this is the case @skshetry. It's just kind of a legacy option name AFAIK. Refer to #2901. 

> Do people really use .hg folders for their work? Unless they do (which they shouldn't)

You'd be surprised. Why shouldn't they? Will the directory name police pul them over? 😛 

> I realized that there might be indeed a stronger motivation to include more defaults

@shcheklein I don't even see this. DVC ignores .git/ automatically because it integrates with Git. I don't see a reason to treat any other file/dir name specially. Your questions about .gitignore vs. .dvcignore are interesting though – I think there may be some issues already about that.

---

In summary, I vote to remove the 4 characters we're talking about, but I realize this is probably a moot point since no one even uses Mercurial anymore haha.

skshetry: @jorgeorpinel, I think something good thing to do is in middle. If people have a certain directories such as `__pycache__` and `venv` or similar kinds of directories, it'd be good to make them aware that `.dvcignore` exists, at least a warning that traversing will be slow.

jorgeorpinel: Agree about improving /dvcignore awareness. (I think the warning may exist BTW?)

For docs, I opened iterative/dvc.org/issues/1033.

shcheklein: > at least a warning that traversing will be slow.

it's a really good idea. Please create an issue for this in the DVC core repo.

I still don't see a reason to include `.hg` though. To my mind if we follow this logic it should be done properly like you described for example.

skshetry: Created #3436

skshetry: Resolving this for now.

PR #3260, opened 2020-01-30
diff: format summary to make it more human readable

mroutis:


efiop:
Oops, might've jumped the gun on this one, sorry. But just a heads up: if this is alright, we also need to adjust the docs accordingly.

PR #3263, opened 2020-01-30
lint:beautysh pre-commit hook

### LINKED ISSUE #2834: add shfmt hooks to format shell scripts
Our shell scripts are all over the place, it would be nice to set a specific coding style and enforce it, same as we do with black for python.  https://github.com/mvdan/sh#shfmt has some hooks for pre-commit too, and restyled.io has that too.

### LINKED ISSUE #2969: tidy scripts
- whitespace formatting (related #2834 automate formatting)
- consistent use of `print_error()`/`echo`/`echo -e`/`echo >&2`
- possibly abstract common functionality into one place
- related #2956

casperdcl:
- [x] whitespace formatting (fixes #2834 automate formatting)
- [ ] ~~consistent use of `print_error()`/`echo`/`echo -e`/`echo >&2`~~
- [ ] ~~possibly abstract common functionality into one place~~
- [x] add pre-commit hook
  + [x] ensure CI tests work
- [x] ~~discuss & document `go>=1.13` requirement~~ use python formatter!
- fixes #2969

if reviewing online, make sure to [ignore whitespace to see actual changes quicker][ws]

[image][ws]

[ws]: https://github.com/iterative/dvc/pull/3263/files?diff=unified&w=1


.pre-commit-config.yaml
```diff
@@ -9,3 +9,8 @@ repos:
     hooks:
     - id: flake8
       language_version: python3
+-   repo: https://github.com/casperdcl/sh
+    rev: master
+    hooks:
+    - id: shfmt
+      language: golang
```
shcheklein: why is it golang though? :)

casperdcl: requirement by `shfmt` (suggested from https://github.com/iterative/dvc/issues/2834).

Will have to find something different if we don't want `golang`...

.travis.yml
```diff
@@ -24,14 +24,21 @@ stages:
...
     before_install:
     install:
     script: ./scripts/ci/check_patch.sh
+  - name: "Script formatting"
+    os: linux
+    language: go
+    go: 1.13
+    before_install:
+    install:
+    script: ./scripts/ci/check_scripts.sh
```
efiop: No need, just add that script to `scripts/ci/check_patch.sh`, where we check README and stuff like that anyways.

efiop:
>  consistent use of print_error()/echo/echo -e/echo >&2
> possibly abstract common functionality into one place

Let's not bother with those for now, it has a very low priority.

.travis.yml
casperdcl: Can't because it requires `go: 1.13` while the other scripts need a decent `python: 3.7`. Travis doesn't support having both in one job. I could get around this by manually imstalling both in one job (maybe) but not sure if it's worth the effort

.pre-commit-config.yaml
casperdcl: btw on this point. Maybe don't have this as a pre-commit hook so we don't force devs to install `go`? We can keep it as a CI-only thing...

.travis.yml
casperdcl: to be explicit, I already tried: https://github.com/iterative/dvc/compare/7a9512d02..0f13f4726 (that diff fixed travis)

scripts/ci/check_patch.sh
```diff
@@ -3,10 +3,13 @@
...
+pip install Pygments collective.checkdocs pre-commit
 
 # stop the build if there are any readme formatting errors
 python setup.py checkdocs
 
 # stop the build if there are any formatting errors
-pre-commit run --all-files
+err=0
+pre-commit run --all-files black || err=1
+pre-commit run --all-files flake8 || err=1
```
casperdcl: I don't really like this as it's not future-proof (if more pre-commit hooks are added in `.pre-commit-config.yaml` in future, they'll have to be manually added here too. Ideally we need something like `pre-commit run --all-files --exclude-hooks shfmt`

casperdcl:
Oh also just found https://github.com/lovesegfault/beautysh which is python (yay) but does say

> in tests with large Linux system Bash scripts, its error-free score was ~99%

which sounds scary. It's ambiguous whether it means it'll *break* scripts, or just break formatting.

.pre-commit-config.yaml
casperdcl: nvm swapped to using `beautysh` which is python

.travis.yml
casperdcl: > nvm swapped to using `beautysh` which is python :tada:

scripts/ci/check_patch.sh
casperdcl: outdated

casperdcl:
> which sounds scary. It's ambiguous whether it means it'll _break_ scripts, or just break formatting.

Got confused with one script:

```bash
[[ condition1 \
  condition2 ]]
```

where it fixed indentation in such a way that it kept complaining about its own formatting.

I manually fixed this in a way that shouldn't be a problem.

efiop:
@casperdcl That is pretty scary. Would rather not rely on something so unreliable, as it might cause big problems for an unsuspecting contributor.

casperdcl:
well idk... it didn't break anything and also is used in https://github.com/Glavin001/atom-beautify. Seems like it'll solve a lot more problems than it may introduce. I know it hasn't broken anything so far because the diff between `shfmt` and `beautysh` is 4e87bf9e0e270. I don't like `shfmt`'s `go` dependency any more than @shcheklein :)

efiop:
@casperdcl To be honest, still quite worried about it, but I am willing to give it a try for now. If anything goes wrong - we'll just get rid of it. Thanks!

PR #3264, opened 2020-01-30
utils: more robust text file detection on checksum

### LINKED ISSUE #3261: Pull extrememly slow on ~400GB of data with hot DVC cache
OS: Docker image based off of `tensorflow/tensorflow:latest-gpu-py3`, with Python 3.7.5 as the system Python.

Initial setup:
```
# python --version
Python 3.7.5
# python -m venv .profiling_venv
# source .profiling_venv/bin/activate
# pip install dvc yappi
...
# dvc --version
0.82.6
```

DVC cache configuration:
```
# dvc config cache.dir
ERROR: configuration error - config file error: section 'cache' doesn't exist
# dvc config --local cache.dir
ERROR: configuration error - config file error: section 'cache' doesn't exist
# dvc config --global cache.dir
ERROR: configuration error - config file error: section 'cache' doesn't exist
# dvc config --system cache.dir
/ssd/.../dvc/cache
# cat /etc/xdg/dvc/config
[cache]
dir = /ssd/.../dvc/cache
protected = true
type = "hardlink,symlink"
```
Please note that the DVC cache is hot. In other words, most if not all files for `dvc fetch` are present and up-to-date at `/ssd/...`.

Make a fresh clone and profile `dvc pull`:
```
# git clone ssh://... repo
...
# cd repo
# yappi -f callgrind -o dvc_pull.prof -s $(which dvc) pull
...
```
**This `dvc pull`, uninstrumented, usually takes 40+ minutes with a hot DVC cache.**

Count the number of DVC-tracked files (symlinks, see the above config) and the total size of the repo:
```
# find . -type l | wc -l
29003
# du -shL .
403G    .
```

Looking at the `dvc_pull.prof` (in KCachegrind) suggests that the bottleneck is the checksum process. The `file_md5` and `dos2unix` functions in `utils/__init__.py` appear particularly costly.

Is this a known issue? Would the primary authors of DVC entertain a more performant version of `file_md5` (perhaps written in C/C++ and without TQDM integration)?

[dvc_pull_prof.zip](https://github.com/iterative/dvc/files/4135825/dvc_pull_prof.zip)

Screenshot_2020-01-30_14-02-20

### LINKED ISSUE #3364: Misclassification of binary files as text files
In the case where binary files have a large text header, the current
checksum routine will treat said files as text files and normalize
line-endings before performing the checksum. Not only is it dangerous to
manipulate binary files like this, it also doubles the runtime of the
checksum routine, as every block of data must be read twice.

As noted in #3264, at the minimum, DVC should probably match
Git's text file detection routine, which interrogates the first 8 kilobytes
(and doesn't do heuristics on ratio of printable characters,
as DVC currently does).

kevlar1818:
In the case where binary files have a large text header, the current
checksum routine will treat said files as text files and normalize
line-endings before performing the checksum. Not only is it dangerous to
manipulate binary files like this, it also doubles the runtime of the
checksum routine, as every block of data must be read twice. This patch
makes the process for detecting text files more robust by increasing the
number of bytes interrogated by DVC used to classify the file.

**Note:** In my team's case, this will invalidate our DVC cache and remote, as the checksums for most of our corpus were incorrectly calculated. We can, of course, fix this by re-`dvc add`ing all of our files, but I want to make sure you (the maintainers) are aware of the ramifications for others.

This doesn't outright fix #3261, but it was a result of that issue.

Fixes #3364 

### DIFF
--- a/dvc/utils/__init__.py
+++ b/dvc/utils/__init__.py
@@ -32,13 +32,13 @@ def dos2unix(data):
 def file_md5(fname):
     """ get the (md5 hexdigest, md5 digest) of a file """
     from dvc.progress import Tqdm
-    from dvc.istextfile import istextfile
+    from dvc.istextfile import istext
 
     fname = fspath_py35(fname)
 
     if os.path.exists(fname):
         hash_md5 = hashlib.md5()
-        binary = not istextfile(fname)
+        is_file_binary = None
         size = os.path.getsize(fname)
         no_progress_bar = True
         if size >= LARGE_FILE_SIZE:
@@ -62,7 +62,10 @@ def file_md5(fname):
                     if not data:
                         break
 
-                    if binary:
+                    if is_file_binary is None:
+                        is_file_binary = not istext(data)
+
+                    if is_file_binary:
                         chunk = data
                     else:
                         chunk = dos2unix(data)
--- a/dvc/istextfile.py
+++ b/dvc/istextfile.py
@@ -7,15 +7,11 @@
 TEXT_CHARS = bytes(range(32, 127)) + b"\n\r\t\f\b"
 
 
-def istextfile(fname, blocksize=512):
-    """ Uses heuristics to guess whether the given file is text or binary,
-        by reading a single block of bytes from the file.
-        If more than 30% of the chars in the block are non-text, or there
-        are NUL ('\x00') bytes in the block, assume this is a binary file.
+def istext(block):
+    """ Uses heuristics to guess whether the given block of bytes is text or
+    binary. If more than 30% of the chars in the block are non-text, or there
+    are NUL ('\x00') bytes in the block, assume this is a binary file.
     """
-    with open(fname, "rb") as fobj:
-        block = fobj.read(blocksize)
-
     if not block:
         # An empty file is considered a valid text file
         return True
@@ -28,3 +24,11 @@ def istextfile(fname, blocksize=512):
     # occurrences of TEXT_CHARS from the block
     nontext = block.translate(None, TEXT_CHARS)
     return float(len(nontext)) / len(block) <= 0.30
+
+
+def istextfile(fname, blocksize=2048):
+    """ Uses heuristics on the first 'blocksize' bytes in a file to guess
+    whether the given file is text or binary.
+    """
+    with open(fname, "rb") as fobj:
+        return istext(fobj.read(blocksize))


kevlar1818:
@mroutis I agree about removing the line-ending handling. That seems like something that the user should be responsible for. Heck, you could make a `dos2unix.dvc` DVC stage :smile:

kevlar1818:
Just to reiterate: Merging this PR and cutting a build may result in users seeing a lot `WARNING: corrupted cache file` in dvc, as a subset of files that were once being treated as text files will now be treated as binary files. Please consider these ramifications before merging. I'm happy to discuss more :+1:

mroutis:
@kevlar1818 , would making dos2unix optional works for you and your team? This one would be less aggro to the rest of the users by keeping compatibility with previous versions)

efiop:
> @kevlar1818 , would making dos2unix optional works for you and your team? This one would be less aggro to the rest of the users by keeping compatibility with previous versions)

It is a bad idea, as your remotes don't know anything about your config, and people use 1 remote for multiple projects, which might result in some terrible situations with cache corruption.

dvc/utils/__init__.py
```diff
@@ -62,7 +62,10 @@ def file_md5(fname):
                     if not data:
                         break
 
-                    if binary:
+                    if is_file_binary is None:
+                        is_file_binary = not istext(data)
```
efiop: This basically kiils the heuristics and starts checking the whole file contents. I suppose we could extend the heuristics to "check first N bytes + check last N bytes" to cover your particular case, but there is no real need for it, as the main intention for this heuristics is providing md5 compatibility for git-tracked files between *nix and windows, where git uses CRLF by default even if you didn't have them before in your file. So we actually need to have a heuristic that is an exact match to the one that git uses. :slightly_smiling_face:

That being said, we sure could look into optimizing this by not opening the file again or optimizing dos2unix() etc.

kevlar1818: FWIW: This code still only checks the first N bytes. Where N was 512 bytes before, now it's the first MB.

kevlar1818: Looks like Git may use [8K](https://git.kernel.org/pub/scm/git/git.git/tree/xdiff-interface.c?id=HEAD#n187)
See also: https://stackoverflow.com/questions/6119956/how-to-determine-if-git-handles-a-file-as-binary-or-as-text

kevlar1818:
What's the consensus on this PR? I think it makes sense to raise the number of bytes interrogated to determine if a file is text or binary, and I think looking over the first chunk read prior to checksum calculation (in this case 1MB) is reasonable. I would also support simply changing the text/binary detector to match Git (8K, only looking for null bytes -- no printable char checking).

efiop:
@kevlar1818 Waiting for prof results from you in https://github.com/iterative/dvc/issues/3261 , as you've promised :slightly_smiling_face: Need to figure out the causes first, this PR is not mergeable as is right now, we need really strong reasons for it.

efiop:
Closing since we've found the source in the issue.

PR #3269, opened 2020-01-31
gdrive: support Google Service accounts

MaxRis:
https://github.com/iterative/PyDrive2/pull/7 should be merged first

Introduces `gdrive_service_account_email`, `gdrive_service_account_user_email` and `gdrive_service_account_p12_file_path` DVC config params to support authorization with Google Service accounts.

Introduced config params are independent from `client_id` and `client_secret` config params

Update:

https://github.com/iterative/PyDrive2/pull/11 should be merged first

Introduces  `gdrive_user_service_account` DVC config param disabled by default.

### DIFF
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -172,9 +172,13 @@ SCHEMA = {
                     **REMOTE_COMMON,
                 },
                 "gdrive": {
+                    "gdrive_use_service_account": Bool,
                     "gdrive_client_id": str,
                     "gdrive_client_secret": str,
                     "gdrive_user_credentials_file": str,
+                    "gdrive_service_account_email": str,
+                    "gdrive_service_account_user_email": str,
+                    "gdrive_service_account_p12_file_path": str,
                     **REMOTE_COMMON,
                 },
                 "http": {**HTTP_COMMON, **REMOTE_COMMON},
--- a/dvc/remote/gdrive.py
+++ b/dvc/remote/gdrive.py
@@ -76,7 +76,7 @@ class RemoteGDrive(RemoteBASE):
     DEFAULT_NO_TRAVERSE = False
     DEFAULT_VERIFY = True
 
-    GDRIVE_USER_CREDENTIALS_DATA = "GDRIVE_USER_CREDENTIALS_DATA"
+    GDRIVE_CREDENTIALS_DATA = "GDRIVE_CREDENTIALS_DATA"
     DEFAULT_USER_CREDENTIALS_FILE = "gdrive-user-credentials.json"
 
     def __init__(self, repo, config):
@@ -93,17 +93,22 @@ class RemoteGDrive(RemoteBASE):
             )
 
         self._bucket = self.path_info.bucket
+        self._use_service_account = config.get("gdrive_use_service_account")
+        self._service_account_email = config.get(
+            "gdrive_service_account_email"
+        )
+        self._service_account_user_email = config.get(
+            "gdrive_service_account_user_email"
+        )
+        self._service_account_p12_file_path = config.get(
+            "gdrive_service_account_p12_file_path"
+        )
         self._client_id = config.get("gdrive_client_id")
         self._client_secret = config.get("gdrive_client_secret")
-        if not self._client_id or not self._client_secret:
-            raise DvcException(
-                "Please specify Google Drive's client id and "
-                "secret in DVC config. Learn more at "
-                "{}.".format(format_link("https://man.dvc.org/remote/add"))
-            )
+        self._validate_config()
         self._gdrive_user_credentials_path = (
             tmp_fname(os.path.join(self.repo.tmp_dir, ""))
-            if os.getenv(RemoteGDrive.GDRIVE_USER_CREDENTIALS_DATA)
+            if os.getenv(RemoteGDrive.GDRIVE_CREDENTIALS_DATA)
             else config.get(
                 "gdrive_user_credentials_file",
                 os.path.join(
@@ -114,6 +119,33 @@ class RemoteGDrive(RemoteBASE):
 
         self._list_params = None
 
+    def _validate_config(self):
+        # Validate Service Account configuration
+        if self._use_service_account and (
+            not self._service_account_email
+            or not self._service_account_p12_file_path
+        ):
+            raise DvcException(
+                "To use service account please specify {}, {} and "
+                "{} in DVC config. Learn more at "
+                "{}.".format(
+                    "gdrive_service_account_email",
+                    "gdrive_service_account_p12_file_path",
+                    "gdrive_service_account_user_email (optional)",
+                    format_link("https://man.dvc.org/remote/modify"),
+                )
+            )
+
+        # Validate OAuth 2.0 Client ID configuration
+        if not self._use_service_account and (
+            not self._client_id or not self._client_secret
+        ):
+            raise DvcException(
+                "Please specify Google Drive's client id and "
+                "secret in DVC config. Learn more at "
+                "{}.".format(format_link("https://man.dvc.org/remote/modify"))
+            )
+
     @wrap_prop(threading.RLock())
     @cached_property
     def drive(self):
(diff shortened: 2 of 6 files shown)


efiop:
For the record: PyDrive2 1.4.4 has been released.

efiop:
@MaxRis Any progress on this?

MaxRis:
@efiop I believe, it should be evaluated/tested by @shcheklein 
Also it is turned out that refresh token is not generated for Service accounts. Since PyDrive supports only old p12 keys to do Service account auth, Travis setup will require additional changes (bundling of p12 cert with non-default password; maybe limiting GDrive API scope only to drive.appdata).

shcheklein:
@MaxRis @efiop we add a lot of code, but I'm not convinced it solves the testing. At least for now. It does not solve the limits issue, unless I'm missing something? What else am I missing?

MaxRis:
@shcheklein yes, it seems it doesn't help to solve limits issues, since Service Account impersonate some another user in domain and the limits are the same.
These changes might be considered as "New functionality" to support login with Service Account. While it is not ready to be automated on Travis, it works perfectly locally.

MaxRis:
> These changes might be considered as "New functionality" to support login with Service Account. While it is not ready to be automated on Travis, it works perfectly locally.

Locally user specified path to p12 key and PyDrive automatically retrieves new access_token for Service Account after old one is expired. The difference to the auth for normal Google account is that in addition to an access_token a refresh_token is obtained, and PyDrive automatically retrieves a new access_token by using a refresh_token value.

shcheklein:
@MaxRis not sure I understood the last part about refresh tokens and differences, could you please elaborate a bit?

> These changes might be considered as "New functionality" to support login with Service Account. While it is not ready to be automated on Travis, it works perfectly locally.

what would the use case for this? automation? like in CI/CD or something?

> While it is not ready to be automated on Travis, it works perfectly locally.

what are we missing here?

MaxRis:
@shcheklein 

> what would the use case for this? automation? like in CI/CD or something?

Yes, use case is to run tests on Travis with Service Account.

> what are we missing here?

p12 key with custom password to do auth with Service Account from Iterative's G Suite and secure way (probably, one another env var) to pass custom password into PyDrive.ServiceAuth()

shcheklein:
@MaxRis do service accounts have their own GDrive space?

could you please elaborate on that difference in the token refresh mechanism, please? I still don't get it.

MaxRis:
@shcheklein 

> @MaxRis do service accounts have their own GDrive space?

No, Service account is a "resource" which has permissions to access other resources.

This PR was tested with following flow https://developers.google.com/identity/protocols/OAuth2ServiceAccount#delegatingauthority . As I understand, Service account receives access to any domain's data on behalf of any domain's user, but the access scope for this Service account can be limited to list of desired access scopes (for example, only `https://www.googleapis.com/auth/drive` ).

> could you please elaborate on that difference in the token refresh mechanism, please? I still don't get it.

For usual Google drive user generated `gdrive-user-credentials.json` contains both access_token and refresh_token. Just content of `gdrive-user-credentials.json` is enough to have unlimited in time access to the user's Google Drive.

For Service account generated `gdrive-user-credentials.json` contains only access_token which is valid only for 1 hour. After access_token is expired, it is supposed that p12 key file is available around to ask Google API to generate new access_token.

btw, [in that Google's doc](https://cloud.google.com/iam/docs/understanding-service-accounts) they warn:

```
Always discourage developers from checking keys into the source code or leaving them in the Downloads directory of their workstation.
```

shcheklein:
@MaxRis Ok, makes sense. I checked the page and it looks like there is an option to not involve user into this 2LO vs 3LO. Don't you think for automation and testing purposes it would be easier to use 2LO from that doc? I don't see a need to impersonate someone through a service account, at least it does not look superior to the regular service account. Unless I'm missing something again.

dvc/remote/gdrive.py
```diff
@@ -84,16 +84,51 @@ def __init__(self, repo, config):
...
+                Config.SECTION_GDRIVE_CLIENT_ID, None
+            )
+            self.client_secret = self.config.get(
+                Config.SECTION_GDRIVE_CLIENT_SECRET, None
+            )
+            if not self.client_id or not self.client_secret:
+                raise DvcException(
+                    "Please specify Google Drive's client id and "
+                    "secret in DVC's config. Learn more at "
+                    "{}.".format(format_link("https://man.dvc.org/remote/add"))
```
shcheklein: link to remote modify is better?

dvc/remote/gdrive.py
```diff
@@ -84,16 +84,51 @@ def __init__(self, repo, config):
...
+                )
+        else:
             raise DvcException(
-                "Please specify Google Drive's client id and "
-                "secret in DVC's config. Learn more at "
-                "{}.".format(format_link("https://man.dvc.org/remote/add"))
+                "Please specify {},{} and {} "
```
shcheklein: space missing

dvc/remote/gdrive.py
```diff
@@ -84,16 +84,51 @@ def __init__(self, repo, config):
...
+                "in DVC's config. Learn more at "
+                "{}.".format(
+                    Config.SECTION_GDRIVE_SERVICE_ACCOUNT_EMAIL,
+                    Config.SECTION_GDRIVE_SERVICE_ACCOUNT_USER_EMAIL,
+                    Config.SECTION_GDRIVE_SERVICE_ACCOUNT_P12_FILE_PATH,
+                    format_link("https://man.dvc.org/remote/add"),
```
shcheklein: same as above

dvc/remote/gdrive.py
```diff
@@ -84,16 +84,51 @@ def __init__(self, repo, config):
...
+        )
+        self.service_account_p12_file_path = self.config.get(
+            Config.SECTION_GDRIVE_SERVICE_ACCOUNT_P12_FILE_PATH, None
+        )
+        if (
+            self.service_account_email
+            and self.service_account_user_email
+            and self.service_account_p12_file_path
+        ):
+            self.use_service_account = True
```
shcheklein: let's make it an explicit option? It'll simplify all this logic. Also users will be able to switch between different credentials if needed.

dvc/config.py
```diff
@@ -177,6 +177,13 @@ class Config(object):  # pylint: disable=too-many-instance-attributes
     SECTION_GDRIVE_CLIENT_ID = "gdrive_client_id"
     SECTION_GDRIVE_CLIENT_SECRET = "gdrive_client_secret"
     SECTION_GDRIVE_USER_CREDENTIALS_FILE = "gdrive_user_credentials_file"
+    SECTION_GDRIVE_SERVICE_ACCOUNT_EMAIL = "gdrive_service_account_email"
+    SECTION_GDRIVE_SERVICE_ACCOUNT_USER_EMAIL = (
```
shcheklein: see my question re impersonating ... do we actually need this?

MaxRis:
> @MaxRis Ok, makes sense. I checked the page and it looks like there is an option to not involve user into this 2LO vs 3LO. Don't you think for automation and testing purposes it would be easier to use 2LO from that doc? I don't see a need to impersonate someone through a service account, at least it does not look superior to the regular service account. Unless I'm missing something again.

Good point!
So far, I was not able to find any evidences that in case of 2LO a Service account can access Google Drive storage (Google Drive storage, seems, should belong to the Google Project, but not to specific user, but I was not able to find the technical way to achieve that so far)

Another even more important thing is that PyDrive expects all 3 following params to be available https://github.com/iterative/PyDrive2/blob/master/pydrive2/auth.py#L146
And if, for example, `client_user_email` (which is used to impersonate real user) is not provided, https://github.com/iterative/PyDrive2/blob/master/pydrive2/auth.py#L416 throws "Insufficient service config in settings". I afraid that PyDrive for the moment doesn't support 2LO since providing of `client_user_email` always expected.

dvc/config.py
MaxRis: `SERVICE_ACCOUNT_USER_EMAIL` will become optional parameter

dvc/remote/gdrive.py
MaxRis: Have added DVC config option `gdrive_user_service_account`

MaxRis:
Once https://github.com/iterative/PyDrive2/pull/11 is merged, it should be sufficient to set secured `GDRIVE_USER_CREDENTIALS_DATA` env var in Travis settings for service account and specify service account email only.

shcheklein:
> @shcheklein I've rebased the branch to master already,
> Since config doesn't have global constants anymore, I have defined `self.GDRIVE_SERVICE_ACCOUNT_EMAIL` locally to reference it twice from inside `RemoteGDrive`. Is it a restriction now to use constants to reference some key from config?

yeah, if you put them in the schema, no need to define upper case constants anymore, just use strings, check similar cases in the gdrive.py

dvc/remote/gdrive.py
```diff
@@ -76,6 +76,10 @@ class RemoteGDrive(RemoteBASE):
     DEFAULT_NO_TRAVERSE = False
     DEFAULT_VERIFY = True
 
+    GDRIVE_SERVICE_ACCOUNT_EMAIL = "gdrive_service_account_email"
+    GDRIVE_SERVICE_ACCOUNT_P12_FILE_PATH = "gdrive_service_account_user_email"
+    GDRIVE_SERVICE_ACCOUNT_USER_EMAIL = "gdrive_service_account_p12_file_path"
```
shcheklein: please, see my comment - no need to do separate constants

dvc/config.py
```diff
@@ -172,9 +172,13 @@ class RelPath(str):
                     **REMOTE_COMMON,
                 },
                 "gdrive": {
+                    "gdrive_use_service_account": Bool,
                     "gdrive_client_id": str,
                     "gdrive_client_secret": str,
                     "gdrive_user_credentials_file": str,
+                    "gdrive_service_account_email": str,
+                    "gdrive_service_account_user_email": str,
+                    "gdrive_service_account_p12_file_path": str,
```
shcheklein: 👍

dvc/remote/gdrive.py
```diff
@@ -114,6 +123,33 @@ def __init__(self, repo, config):
 
         self._list_params = None
 
+    def _validate_config(self):
```
shcheklein: 👍

MaxRis:
@shcheklein this PR is not yet ready, need to setup Service Account and set Travis env var.

MaxRis:
@shcheklein should be ready to be merged

shcheklein:
@MaxRis travis failed, PTAL.

MaxRis:
@shcheklein thanks! Have fixed value of Travis env var and GDrive tests are passing now. There are still random failures of S3 tests (happen on all DVC Travis builds).

efiop:
S3 issues fixed, restarted the tests for this PR. Let's see.

shcheklein:
@efiop I think should be good to merge! :)

Issue #3164, opened 2020-01-15
improve message for rwlock error


efiop:
https://discordapp.com/channels/485586884165107732/563406153334128681/667133061380571146

Current message:
```
ERROR: failed to reproduce 'device_download.dvc': 'data/device_modifier.parquet' is busy, it is being read by:
  (PID 28324): dvc repro device_train.dvc

even though I removed .dvc/lock file
```

shcheklein:
consider adding the second troubleshooting section now that we have a good mechanism for this :)

mroutis:
It is not clear how to improve it :sweat_smile: 
@efiop , any specific suggestions?

shcheklein:
@mroutis I suppose in any way that would prevent us answering the same question on Discord again and again, right? :)

efiop:
@mroutis yep, at least a hint how to delete it. Same as we do with `.dvc/lock`

PR #3270, opened 2020-01-31
rwlock: include actionable in error message

mroutis:
Fix: #3164

### DIFF
--- a/dvc/rwlock.py
+++ b/dvc/rwlock.py
@@ -5,7 +5,6 @@ from collections import defaultdict
 from contextlib import contextmanager
 
 from voluptuous import Schema, Invalid
-from funcy.py3 import lmap, lfilter, lmapcat
 
 from .exceptions import DvcException
 from .lock import LockError
@@ -61,44 +60,25 @@ def _infos_to_str(infos):
     )
 
 
-def _check_no_writers(lock, info, path_infos):
-    for path_info in path_infos:
-        blocking_urls = lfilter(path_info.overlaps, lock["write"])
-        if not blocking_urls:
-            continue
+def _check_blockers(lock, info, *, mode, waiters):
+    for path_info in waiters:
+        blockers = [
+            info
+            for path, infos in lock[mode].items()
+            if path_info.overlaps(path)
+            if info not in (infos if type(infos) is list else [infos])
+        ]
 
-        writers = lmap(lock["write"].get, blocking_urls)
-        writers = lfilter(lambda i: info != i, writers)
-        if not writers:
-            continue
-
-        raise LockError(
-            "'{path}' is busy, it is being written to by:\n{process}"
-            "\n\n"
-            "If there's no process with such PID, you can manually remove"
-            "'.dvc/tmp/rwlock' and try again.".format(
-                path=str(path_info), process=_infos_to_str(writers)
-            )
-        )
-
-
-def _check_no_readers(lock, info, path_infos):
-    for path_info in path_infos:
-        blocking_urls = lfilter(path_info.overlaps, lock["read"])
-        if not blocking_urls:
-            continue
-
-        readers = lmapcat(lock["read"].get, blocking_urls)
-        readers = lfilter(lambda i: info != i, readers)
-        if not readers:
+        if not blockers:
             continue
 
         raise LockError(
-            "'{path}' is busy, it is being read by:\n{process}"
-            "\n\n"
-            "If there's no process with such PID, you can manually remove"
+            "'{path}' is busy, it is being blocked by:\n"
+            "{blockers}\n"
+            "\n"
+            "If there are no processes with such PIDs, you can manually remove"
             "'.dvc/tmp/rwlock' and try again.".format(
-                path=str(path_info), process=_infos_to_str(readers)
+                path=str(path_info), blockers=_infos_to_str(blockers)
             )
         )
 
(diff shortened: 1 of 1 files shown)


dvc/rwlock.py
```diff
@@ -73,8 +73,11 @@ def _check_no_writers(lock, info, path_infos):
             continue
 
         raise LockError(
-            "'{}' is busy, it is being written to by:\n{}".format(
-                str(path_info), _infos_to_str(writers)
+            "'{path}' is busy, it is being written to by:\n{process}"
+            "\n"
+            "If that is not the case, manually remove '.dvc/tmp/rwlock'"
```
shcheklein: do we still have a global lock file at all? or it is a single file? 

I would make a section in the troubleshooting guide that describes this case and put a link from this error, along with the case when `rwlock` is corrupted for some reason. 

"... if that is not the case" is not very descriptive.

mroutis: It is a global lock file where we keep track of the processes in a JSON.

shcheklein: I mean, is it the only file we use now? we used to have a global `lock` file as well. (and this one technically even being a single file is not a global lock - it introduced granular locks the way I understand this).

mroutis: @shcheklein , yes, single file, granular lock.

mroutis: @shcheklein , so, introducing an entry to the troubleshooting guide would make it for you?

What about: ""If that is not the case" -> If the process doesn't exist anymore"

shcheklein: so, we don't use `lock` file anymore, right? only `rwlock` now and this error happens only if rwlock for a specific path is busy?

`If the process doesn't exist anymore` looks better!

I would say if it's the only reason for LockError now, no `lock` files anymore, we don't need any any docs indeed.

mroutis: We do use `.dvc/lock` as a fallback :slightly_smiling_face: 

That's right, this error happens when a specific path is busy.

shcheklein: what is the `fallback` logic here? 

is it the same exception thrown when `.dvc/lock` exists?

mroutis: My bad, @shcheklein , it is not a fallback but a configurable option.
So, there's a `make_lock` method that we use to create locks and you can choose between rwlock and hardlink-lock.

But is either one or the other.

We have the following message for the hardlink based lock:
```python
    "cannot perform the command because another DVC process seems to be "
    "running on this project. If that is not the case, manually remove "
    "`.dvc/lock` and try again."
```

shcheklein: got it, thanks @mroutis !

dvc/rwlock.py
```diff
@@ -91,8 +94,11 @@ def _check_no_readers(lock, info, path_infos):
             continue
 
         raise LockError(
-            "'{}' is busy, it is being read by:\n{}".format(
-                str(path_info), _infos_to_str(readers)
+            "'{path}' is busy, it is being read by:\n{process}"
```
shcheklein: I think at this point it's better to this message compose logic to a function or string template. To avoid code duplication

mroutis:
I will handle the duplication in another pull request, as agreed on planning.
I think both methods can be abstracted.

mroutis:
@efiop , I already had the patch for it, since this PR isn't closed yet, I will upload it here.

dvc/rwlock.py
```diff
@@ -61,38 +60,25 @@ def _infos_to_str(infos):
...
-        readers = lmapcat(lock["read"].get, blocking_urls)
-        readers = lfilter(lambda i: info != i, readers)
-        if not readers:
-            continue
-
-        raise LockError(
-            "'{}' is busy, it is being read by:\n{}".format(
-                str(path_info), _infos_to_str(readers)
+            "'{path}' is busy, it is being blocked by:\n"
+            "{blockers}\n"
```
shcheklein: is it an array now? why? it should be always single blocker, right? let's put an assert and print the first element?

mroutis: @shcheklein , it has been always like that, it is more clear now :slightly_smiling_face: 

This implementation of RWLock allows multiple readers to a single file

shcheklein: ok, I see. it's quite strange by itself, but not related to this PR 👍

mroutis: what's strange, @shcheklein ? :slightly_smiling_face:

shcheklein: that blockers is an array ... unless I'm missing something there should 0 or 1 _blocker_

mroutis: @shcheklein , blockers contain process information, there could be many processes reading the same file, those processes are blockers for a writer.

hope that brings some clarity

shcheklein: my bad, you are right ... multiple readers situation is possible .. ideally we should take this into account in the message.

shcheklein: I mean put process into plural, ID into IDs ... etc ... probably minor nuance, but can be confusing a bit for those who read code

mroutis: done, @shcheklein )

shcheklein: thanks @mroutis !

dvc/rwlock.py
```diff
@@ -61,38 +60,25 @@ def _infos_to_str(infos):
...
-        if not blocking_urls:
-            continue
+def _check_blockers(lock, info, *, mode, waiters):
+    for path_info in waiters:
+        blockers = [
+            info
+            for path, infos in lock[mode].items()
+            if path_info.overlaps(path)
+            if info not in (infos if type(infos) is list else [infos])
+        ]
```
Suor: This logic is broken now. We create a list of `info` passed as param repeated 0 to several times, while in previous logic we were listing infos we've got from lock.

The right logic I guess is:
```python
blockers = [
    blocker
    for path, infos in lock[mode].items()
    if path_info.overlaps(path)
    for blocker in (infos if isinstance(infos, list) else [infos])
    if blocker != info
]
```

Suor: Fixed in #3285.

PR #3283, opened 2020-02-05
rwlock: require info keys to be present in schema

mroutis:
Previous implementation allowed incomplete info:

```python
SCHEMA({
    "write": {
        "file_1": {"pid": 1, "cmd": "exec"},
        "file_2": {"pid": 2},
        "file_3": {"cmd": "..."},
        "file_4": {},
    },
    "read": {
        "file_1": [{"pid": 1}],
    }
})
```

### DIFF
--- a/dvc/rwlock.py
+++ b/dvc/rwlock.py
@@ -4,15 +4,20 @@ import json
 from collections import defaultdict
 from contextlib import contextmanager
 
-from voluptuous import Schema, Invalid
+from voluptuous import Schema, Invalid, Required, Optional
 
 from .exceptions import DvcException
 from .lock import LockError
 from .utils import relpath
 
-INFO_SCHEMA = {"pid": int, "cmd": str}
+INFO_SCHEMA = {Required("pid"): int, Required("cmd"): str}
 
-SCHEMA = Schema({"write": {str: INFO_SCHEMA}, "read": {str: [INFO_SCHEMA]}})
+SCHEMA = Schema(
+    {
+        Optional("write", default={}): {str: INFO_SCHEMA},
+        Optional("read", default={}): {str: [INFO_SCHEMA]},
+    }
+)
 
 
 class RWLockFileCorruptedError(DvcException):
@@ -35,15 +40,13 @@ def _edit_rwlock(lock_dir):
     path = os.path.join(lock_dir, "rwlock")
     try:
         with open(path, "r") as fobj:
-            lock = json.load(fobj)
-        lock = SCHEMA(lock)
+            lock = SCHEMA(json.load(fobj))
     except FileNotFoundError:
         lock = SCHEMA({})
     except json.JSONDecodeError as exc:
         raise RWLockFileCorruptedError(path) from exc
     except Invalid as exc:
         raise RWLockFileFormatError(path) from exc
-    lock = defaultdict(dict, lock)
     lock["read"] = defaultdict(list, lock["read"])
     lock["write"] = defaultdict(dict, lock["write"])
     yield lock


dvc/rwlock.py
```diff
@@ -4,16 +4,20 @@
...
 from .exceptions import DvcException
 from .lock import LockError
 from .utils import relpath
 
-INFO_SCHEMA = {"pid": int, "cmd": str}
-
-SCHEMA = Schema({"write": {str: INFO_SCHEMA}, "read": {str: [INFO_SCHEMA]}})
+SCHEMA = Schema(
+    {
+        Optional("write", default={}): {Extra: {"pid": int, "cmd": str}},
```
efiop: IIRC, it is `Optional` by default. Or am I missing something?

efiop: Not sure I understand why `Extra` is needed though. Could you elaborate?

Extra is used to define specific format of extra data that could be present without raising an exception. But I think you just need `{Required("pid"): int, Required("cmd"): str}` here

mroutis: @efiop , we have the `required=True` option, so it requires every key by default.

Without `Extra` it will require at least one object inside the `write` or `read`.


```python
>>> SCHEMA = Schema(
...     {
...         Optional("write", default={}): {str: {"pid": int, "cmd": str}},
...         Optional("read", default={}): {Extra: [{"pid": int, "cmd": str}]},
...     },
...     required=True,
... )

>>> SCHEMA({"write": {}, "read": {}})

MultipleInvalid: required key not provided @ data['write'][<class 'str'>]
```

mroutis: However, you are right, if we remove the `required=True` we only need to require the `INFO_SCHEMA`. I will make the change since it would be easier to review.

dvc/rwlock.py
```diff
@@ -4,16 +4,21 @@
...
-from voluptuous import Schema, Invalid
+from voluptuous import Schema, Invalid, Required
 from funcy.py3 import lmap, lfilter, lmapcat
 
 from .exceptions import DvcException
 from .lock import LockError
 from .utils import relpath
 
-INFO_SCHEMA = {"pid": int, "cmd": str}
+INFO_SCHEMA = Schema({Required("pid"): int, Required("cmd"): str})
```
efiop: No need to wrap with `Schema`, as it is already present in `SCHEMA` down below.
```suggestion
INFO_SCHEMA = {Required("pid"): int, Required("cmd"): str}
```

mroutis: Indeed :+1:

dvc/rwlock.py
```diff
@@ -4,15 +4,20 @@
...
 from .utils import relpath
 
-INFO_SCHEMA = {"pid": int, "cmd": str}
+INFO_SCHEMA = {Required("pid"): int, Required("cmd"): str}
 
-SCHEMA = Schema({"write": {str: INFO_SCHEMA}, "read": {str: [INFO_SCHEMA]}})
+SCHEMA = Schema(
+    {
+        Required("write", default={}): {str: INFO_SCHEMA},
+        Required("read", default={}): {str: [INFO_SCHEMA]},
```
efiop: So these used to be `Optional`(default) but now they are `Required`. Is this intended?

Sure, de-facto we are always setting even empty dicts for these down below, so it won't break anything in dvc itself. But on the other hand, these are not required by our reader and it is not required in terms of format, so I don't really see a point in making them `Required` now. The problem you've described in the PR description is fully fixed by `INFO_SCHEMA` adjustment you've done above. If you want to get rid of `defaultdict` then let's `Optional(..., default={})` here instead.

mroutis: @efiop , I think `Required` and `Optional` are the same when a `default` is provided, is just semantically different.

I'll make the change, since it makes more sense in that regard.

PR #3286, opened 2020-02-07
erepo: cache all read only external repos by hexsha

### LINKED ISSUE #3280: Speed up dvc status for large projects

```
dvc version

DVC version: 0.82.8
Python version: 3.7.6
Platform: Linux-5.0.0-37-generic-x86_64-with-debian-buster-sid
Binary: False
Package: pip
Cache: reflink - not supported, hardlink - not supported, symlink - supported
Filesystem type (cache directory): ('ext4', '/dev/sda1')
Filesystem type (workspace): ('ext4', '/dev/nvme0n1p2')
```

expect `dvc status` can return in a shorter time, currently it takes 30s.

a few stats
```
du -shL my-project # 107G
```

```
dvc status -v
```
[log.txt](https://github.com/iterative/dvc/files/4153107/log.txt)

Suor:
So we have 3 things cached now separately:
- clean clones, to not ask for creds repatedly
- cache dirs, also shared between erepos with same origin
- checked out clones if they are read only, addressed by (url, hexsha)

Several additions to `Git` along the way:
- Git.is_sha() static method
- .pull() and .push() work with multiple returned records correctly
- .get_rev() and .resolve_rev() work faster
- .resolve_rev() looks for remote branches
- .has_rev()

Fixes #3280.

### DIFF
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -15,15 +15,16 @@ from dvc.exceptions import OutputNotFoundError, NoOutputInExternalRepoError
 from dvc.exceptions import FileMissingError, PathMissingError
 from dvc.remote import RemoteConfig
 from dvc.utils.fs import remove, fs_copy
-from dvc.scm import SCM
+from dvc.scm.git import Git
 
 
 logger = logging.getLogger(__name__)
 
 
 @contextmanager
-def external_repo(url, rev=None):
-    path = _cached_clone(url, rev)
+def external_repo(url, rev=None, for_write=False):
+    logger.debug("Creating external repo {}@{}", url, rev)
+    path = _cached_clone(url, rev, for_write=for_write)
     try:
         repo = ExternalRepo(path, url)
     except NotDvcRepoError:
@@ -41,7 +42,8 @@ def external_repo(url, rev=None):
         raise PathMissingError(exc.path, url)
     finally:
         repo.close()
-        _remove(path)
+        if for_write:
+            _remove(path)
 
 
 CLONES = {}
@@ -129,7 +131,7 @@ class ExternalGitRepo:
 
     @cached_property
     def scm(self):
-        return SCM(self.root_dir)
+        return Git(self.root_dir)
 
     def close(self):
         if "scm" in self.__dict__:
@@ -158,23 +160,21 @@ class ExternalGitRepo:
             raise PathMissingError(path, self.url)
 
 
-@wrap_with(threading.Lock())
-def _cached_clone(url, rev):
+def _cached_clone(url, rev, for_write=False):
     """Clone an external git repo to a temporary directory.
 
     Returns the path to a local temporary directory with the specified
-    revision checked out.
+    revision checked out. If for_write is set prevents reusing this dir via
+    cache.
     """
-    # Get or create a clean clone
-    clone_path = CLONES.get(url)
-    if clone_path:
-        # Do not pull for known shas, branches and tags might move
-        if not _is_known_sha(clone_path, rev):
-            _git_pull(clone_path)
-    else:
-        clone_path = tempfile.mkdtemp("dvc-clone")
-        _git_clone(url, clone_path)
-        CLONES[url] = clone_path
+    if not for_write and Git.is_sha(rev) and (url, rev) in CLONES:
+        return CLONES[url, rev]
+
+    clone_path = _clone_default_branch(url, rev)
+    rev_sha = Git(clone_path).resolve_rev(rev or "HEAD")
+
+    if not for_write and (url, rev_sha) in CLONES:
+        return CLONES[url, rev_sha]
 
     # Copy to a new dir to keep the clone clean
     repo_path = tempfile.mkdtemp("dvc-erepo")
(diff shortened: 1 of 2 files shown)


dvc/external_repo.py
```diff
@@ -15,15 +15,16 @@
...
 
 
 logger = logging.getLogger(__name__)
 
 
 @contextmanager
-def external_repo(url, rev=None):
-    path = _cached_clone(url, rev)
+def external_repo(url, rev=None, for_write=False):
+    logger.info("Creating external repo {}@{}", url, rev)
```
efiop: I don't think it should be at info level. Also it would say `Creating external repo url@None`, which is confusing and ugly :slightly_smiling_face:

dvc/external_repo.py
```diff
@@ -184,49 +184,43 @@ def _cached_clone(url, rev):
...
 
 
-def _git_clone(url, path):
-    from dvc.scm.git import Git
-
-    git = Git.clone(url, path)
-    git.close()
-
+@wrap_with(threading.Lock())
+def _clone_master(url, rev):
```
efiop: Name is misleading. It is not cloning master but, i guess, a default branch.

dvc/scm/git/__init__.py
```diff
@@ -324,15 +327,30 @@ def get_tree(self, rev):
...
+        from git.exc import BadName, GitCommandError
+        from contextlib import suppress
+
+        names = [rev] + ([] if Git.is_sha(rev) else ["origin/" + rev])
+        for name in names:
+            with suppress(BadName, GitCommandError):
+                try:
+                    # Try python implementation of rev-parse first, it's faster
+                    return self.repo.rev_parse(name).hexsha
+                except NotImplementedError:
```
efiop: Hm, when does it happen though?

dvc/external_repo.py
Suor: Yeah, I meant this to be a debug one.

dvc/scm/git/__init__.py
Suor: For text search, see `test_diff.test_directories`. There are also dates in `@{...}` format.

Suor:
Everything should be addressed now.

dvc/scm/git/__init__.py
```diff
@@ -324,15 +327,30 @@ def get_tree(self, rev):
...
-        return self.repo.git.rev_parse("HEAD")
+        return self.repo.rev_parse("HEAD").hexsha
 
     def resolve_rev(self, rev):
-        from git.exc import GitCommandError
-
+        from git.exc import BadName, GitCommandError
+        from contextlib import suppress
+
+        names = [rev] + ([] if Git.is_sha(rev) else ["origin/" + rev])
```
skshetry: Is it safe to assume `origin` here? This might not always be the case (except for external_repo, I guess).

>

skshetry: On second thought, this is fine because, if the reference does not exist, `BadName` is thrown.

But, we could check if remote exist via `git.repo.remote().exists()` or get a list of remotes  (`git.repo.remotes`) and check for rev.

Suor: Changed this to more advanced logic to be as close as possible to how `git checkout` behaves.

dvc/external_repo.py
```diff
@@ -158,23 +160,21 @@ def open_by_relpath(self, path, mode="r", encoding=None, **kwargs):
             raise PathMissingError(path, self.url)
 
 
-@wrap_with(threading.Lock())
-def _cached_clone(url, rev):
+def _cached_clone(url, rev, for_write=False):
```
skshetry: Is this `for_write` being used anywhere? Future-proofing?

skshetry:
Oops, forgot to mention. Can we have at least a test that no cloning is done for the same rev? Maybe, just an assert that the cloned directory is same when rev is same?

```python
with external_repo("/tmp/repo", "master") as repo:
    print(repo.root_dir)
with external_repo("/tmp/repo", "master") as repo:
    print(repo.root_dir)
```

dvc/external_repo.py
Suor: It's needed for write scenarios, which are done in dvcx.

Suor:
> Oops, forgot to mention. Can we have at least a test that no cloning is done for the same rev? Maybe, just an assert that the cloned directory is same when rev is same?

Not sure we should test that, this is an implementation detail.

dvc/scm/git/__init__.py
```diff
@@ -324,15 +327,45 @@ def get_tree(self, rev):
...
+            return sha
+
+        # Try all the remotes and if it resolves unambiguously then take it
+        if not Git.is_sha(rev):
+            shas = {
+                _resolve_rev("{}/{}".format(remote.name, rev))
+                for remote in self.repo.remotes
+            } - {None}
+            if len(shas) > 1:
+                raise RevError("ambiguous Git revision '{}'".format(rev))
```
efiop: IIRC git choses the ref-like rev if it can, even in ambigous case. It does print a warning though. Not sure if it matters that much here.

skshetry:
> Not sure we should test that, this is an implementation detail.

I'd argue, it's not an *implementation* detail but a *behavior* that is reasonable and expected, a subtle but important distinction because if I refactor this, I'd like to keep this *behavior* intact (without a need to change code).

How'd you prevent this from regressing?

efiop:
@Suor https://github.com/iterative/dvc/issues/3280#issuecomment-585540941 So we should've added tests, right? :wink:

casperdcl:
:D just posted https://github.com/iterative/dvc/issues/3280#issuecomment-585736296 before reading this

Issue #2865, opened 2019-11-28
Google Drive support further enhancements


maxhora:
This ticket is to keep track of required further improvements for Google Drive `remote` implementation after https://github.com/iterative/dvc/pull/2551 will be merged into `master`:

- [x] Processing of auth exceptions and printing more meaningful error message on failure. https://github.com/iterative/dvc/pull/2551#discussion_r348182771

- [x] Implement `gc` command support for `gdrive` ( `def remove(self, path_info)` overloading )

- [x] Validate resolved remote file object to have expected title ( `def resolve_remote_item_from_path(self, path_parts, create):`  )  

- [x] Simplify a few things using new and shiny `@wrap_prop` and `filter_errors` param of `@retry` from funcy 1.14 :) 

- [x] Make name -> id deterministic by choosing minimum id https://github.com/iterative/dvc/pull/2551#discussion_r350424816

- [x] Reimplement caching to have 2 entry points .path_info_to_ids() and .id_to_path_info(). The idea is to remove cache dictionaries completely, but it will require introduction of helper method which accepts parent_id, title and returns the actual id ( this method introduces 1 -> 1 relation between input and resulting remote id ). https://github.com/iterative/dvc/pull/2551#discussion_r350213128

- [x] Simplify `resolve_remote_item_from_path` https://github.com/iterative/dvc/pull/2551#discussion_r351940507

- [x] Protect `create_remote_dir` with lock. https://github.com/iterative/dvc/pull/2551#discussion_r350428630

- [x] Organize methods, params in ordered and unified way. https://github.com/iterative/dvc/pull/2551#discussion_r350152633

- [x] Enhance retrieving of HTTP error from GDrive API exceptions https://github.com/MaxRis/dvc/commit/783d8b610db9da2bdcee1817cd90366b435e81eb#r36184990

- [x] Create Iterative Google Drive account and Project to share client id and secret with final DVC users. Highest possible API usage limits quotas should be requested. Probably, it will be needed to have separate Google Project for CI. https://github.com/iterative/dvc/pull/2551#discussion_r351936393

- [x] Stop creating a path to the DVC remote root if it does not exist. Since Google Drive allows multiple folders with the same name (at least on in My Drive and one in Shared With Me) in case a path like `gdrive://root/storage` is used to access, collaborators see a `storage` empty folder after the first `dvc pull` attempt`. Instead we should just throw path does not exist.
[...]

PR #3294, opened 2020-02-10
GDrive improvements and code cleanup

shcheklein:
It's a major rewrite now. It was needed since it was very hard an unproductive fixing certain issues w/o addressing some underlying problems with code.

* Performance of remote id resolving improved. It does not hit API now for "remote path". Affects `pull`  - one API call per file less!
* Performance of the first run should be improved. We now cache path->id as we create them (case when we push a lot first time we create a remote).
* Number of retries increased. The previous was not enough with default GDrive limits.
* Does not create remote path anymore (see the `Stop creating a path to the DVC remote root if it does not exist` checkbox) in https://github.com/iterative/dvc/issues/2865
* Code cleanup.

-----------------


shcheklein:
@efiop I suggest we review codeclimate and deepsource settings a bit and make the more useful/less noisy for us. E.g. number of arguments per function - if we think that 4 is to aggressive let's increase it a bit?

PR #3295, opened 2020-02-10
repo: raise exception if root_dir is not a directory path

skshetry:
When Repo is instantiated with invalid urls such as 'http://github.com/iterative/dvc.git',
it does os.path.realpath which is not intelligent enough and just concatenates with current
working directory, which is then traversed outward. It might happen that the outward folder
is a dvc directory which is not really what we want.

So, a check for root to be a directory is added. Otherwise, NotDvcRepoError is thrown with
a custom message.

**TLDR**: `Repo("https://github.com")` and `Repo("not existing directory")` should throw `NotDvcRepoError` even if dvc repo exists outward.

### DIFF
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -112,18 +112,7 @@ class MoveNotDataSourceError(DvcException):
 
 
 class NotDvcRepoError(DvcException):
-    """Thrown if a directory is not a DVC repo.
-
-    Args:
-        root (str): path to the directory.
-    """
-
-    def __init__(self, root, msg=None):
-        msg = msg or (
-            "you are not inside of a DVC repository "
-            "(checked up to mount point '{}')"
-        )
-        super().__init__(msg.format(root))
+    """Thrown if a directory is not a DVC repo"""
 
 
 class DvcParserError(DvcException):
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -118,8 +118,8 @@ class NotDvcRepoError(DvcException):
         root (str): path to the directory.
     """
 
-    def __init__(self, root):
-        msg = (
+    def __init__(self, root, msg=None):
+        msg = msg or (
             "you are not inside of a DVC repository "
             "(checked up to mount point '{}')"
         )
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -129,10 +129,10 @@ class Repo(object):
 
     @classmethod
     def find_root(cls, root=None):
-        root_dir = os.path.abspath(os.path.realpath(root or os.getcwd()))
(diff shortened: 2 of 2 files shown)


dvc/exceptions.py
```diff
@@ -118,8 +118,8 @@ class NotDvcRepoError(DvcException):
         root (str): path to the directory.
     """
 
-    def __init__(self, root):
-        msg = (
+    def __init__(self, root, msg=None):
+        msg = msg or (
```
efiop: While we are at it, let's not include default message in the exception itself but rather provide the message when we raise. I think the only place where this is raised in `find_root`, so shouldn't be too hard to adjust, unless there are too many tests that depend on it.

dvc/repo/__init__.py
```diff
@@ -129,19 +129,19 @@ def __repr__(self):
...
     @classmethod
     def find_root(cls, root=None):
-        if root is None:
-            root = os.getcwd()
-        else:
-            root = os.path.abspath(os.path.realpath(root))
+        root_dir = os.path.abspath(os.path.realpath(root or os.getcwd()))
+
+        if not os.path.isdir(root_dir):
+            raise NotDvcRepoError(root, msg="folder {} does not exist")
```
efiop: We use `directory` terminology in code.

dvc/repo/__init__.py
```diff
@@ -129,19 +129,24 @@ def __repr__(self):
 
     @classmethod
     def find_root(cls, root=None):
-        if root is None:
-            root = os.getcwd()
-        else:
-            root = os.path.abspath(os.path.realpath(root))
+        root_dir = os.path.abspath(os.path.realpath(root or os.getcwd()))
```
pared: To avoid cwd -> relpath -> abspath in case of root == None:
we could `root_dir= os.path.abspath(os.path.relpath(root)) if root else os.getcwd()`

dvc/repo/__init__.py
```diff
@@ -129,19 +129,24 @@ def __repr__(self):
...
+
+        if not os.path.isdir(root_dir):
+            raise NotDvcRepoError("directory {} does not exist".format(root))
```
efiop: We try to enclose generated parts, so it is easier to read logs
```suggestion
            raise NotDvcRepoError("directory '{}' does not exist".format(root))
```

dvc/repo/__init__.py
skshetry: Just like what was before? I'll just rollback.

pared: as @skshetry noted, abspath is not necessary at all here

dvc/repo/__init__.py
skshetry: Updated the message.

PR #3296, opened 2020-02-10
add: meaningful message upon adding overlapping paths

### LINKED ISSUE #3218: dvc add a file from data directory
From a discussion with users: https://opendatascience.slack.com/archives/CGGLZJ119/p1579704007005600 (you need access to this community)

```
$ mkdir datadir
$ cp ~/whatever/* datadir/
$ dvc add datadir/
WARNING: Output 'datadir' of 'datadir.dvc' changed because it is 'modified'

          git add datadir.dvc
$ cp ~/Downloads/newfile.csv datadir/jan2020.csv
$ dvc add datadir/jan2020.csv
ERROR: Paths for outs:                  # <-- error is terrible btw (not related to this issue).
'datadir'('datadir.dvc')
'datadir/file4'('datadir/file4.dvc')
overlap. To avoid unpredictable behaviour, rerun command with non overlapping outs paths.
```

The last command fails because the file is inside a data dir and you suppose to update (`dvc add`) the entire dir. However, a user intuition says (for some users) to add a single file.

Ideally, this should work:
```
$ dvc add datadir/jan2020.csv
'jan2020' was added to dir 'datadir' and 'datadir.dvc' changed because it is 'modified'
100% Add|██████████████████████████████████|1.00/1.00 [00:00<00:00,  2.76file/s]

	git add datadir.dvc
```

pared:

Fixes #3218

### DIFF
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -188,17 +188,10 @@ class RecursiveAddingWhileUsingFilename(DvcException):
 
 
 class OverlappingOutputPathsError(DvcException):
-    def __init__(self, out_1, out_2):
-        super().__init__(
-            "Paths for outs:\n'{}'('{}')\n'{}'('{}')\noverlap. To avoid "
-            "unpredictable behaviour, rerun command with non overlapping outs "
-            "paths.".format(
-                str(out_1),
-                out_1.stage.relpath,
-                str(out_2),
-                out_2.stage.relpath,
-            )
-        )
+    def __init__(self, parent, overlapping_out, message):
+        self.parent = parent
+        self.overlapping_out = overlapping_out
+        super().__init__(message)
 
 
 class CheckoutErrorSuggestGit(DvcException):
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -334,7 +334,17 @@ class Repo(object):
             for out in stage.outs:
                 for p in out.path_info.parents:
                     if p in outs:
-                        raise OverlappingOutputPathsError(outs[p], out)
+                        msg = (
+                            "Paths for outs:\n'{}'('{}')\n'{}'('{}')\n"
+                            "overlap. To avoid unpredictable behaviour, "
+                            "rerun command with non overlapping outs paths."
+                        ).format(
+                            str(outs[p]),
+                            outs[p].stage.relpath,
+                            str(out),
+                            out.stage.relpath,
+                        )
+                        raise OverlappingOutputPathsError(outs[p], out, msg)
 
         for stage in stages:
             stage_path_info = PathInfo(stage.path)
--- a/dvc/repo/add.py
+++ b/dvc/repo/add.py
@@ -4,7 +4,10 @@ import os
 import colorama
 
 from . import locked
-from ..exceptions import RecursiveAddingWhileUsingFilename
+from ..exceptions import (
+    RecursiveAddingWhileUsingFilename,
+    OverlappingOutputPathsError,
+)
 from ..output.base import OutputDoesNotExistError
 from ..progress import Tqdm
 from ..repo.scm_context import scm_context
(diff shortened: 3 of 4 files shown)


pared:
Before:
[asciicast](https://asciinema.org/a/299758)

After:
[asciicast](https://asciinema.org/a/299760)

dvc/exceptions.py
```diff
@@ -326,3 +328,16 @@ def __init__(self, path_info):
...
+    def __init__(self, parent, overlapping_out):
+        super().__init__(
+            "Cannot add '{out}', because it is overlapping with other DVC "
+            "tracked output: '{parent}'.\nTo include '{out}' in '{parent}', "
+            "run 'dvc commit {parent_stage}'".format(
+                out=overlapping_out,
+                parent=parent,
+                parent_stage=parent.stage.relpath,
+            )
+        )
```
efiop: Let's do it a bit differently. Let's only keep `OverlappingOutputPathsError`, but raise it with specific message instead of hardcoding it into the exception class itself. That way we won't spawn new entities and it will be nicely catchable from outside if needed :slightly_smiling_face:

pared: Yeah, makes sense

dvc/repo/add.py
```diff
@@ -49,7 +52,21 @@ def add(repo, targets, recursive=False, no_commit=False, fname=None):
 
             stages = _create_stages(repo, sub_targets, fname, pbar=pbar)
 
-            repo.check_modified_graph(stages)
+            try:
+                repo.check_modified_graph(stages)
+            except OverlappingOutputPathsError as e:
+                msg = (
+                    "Cannot add '{out}', because it is overlapping with other "
+                    "DVC tracked output: '{parent}'.\nTo include '{out}' in "
```
efiop: It is better to move a line down if you have `\n`. E.g.
```
"Cannot add '{out}', because it is overlapping with other "
"DVC tracked output: '{parent}'.\n"
"To include '{out}' in ..."
```
much easier to comprehend that way. Or was it black who did that?

dvc/repo/add.py
```diff
@@ -49,7 +52,21 @@ def add(repo, targets, recursive=False, no_commit=False, fname=None):
 
             stages = _create_stages(repo, sub_targets, fname, pbar=pbar)
 
-            repo.check_modified_graph(stages)
+            try:
+                repo.check_modified_graph(stages)
+            except OverlappingOutputPathsError as e:
```
efiop:
```suggestion
            except OverlappingOutputPathsError as exc:
```
DS and others complain about such naming.

tests/func/test_add.py
```diff
@@ -641,3 +641,17 @@ def test_escape_gitignore_entries(tmp_dir, scm, dvc):
...
+def test_add_from_data_dir(tmp_dir, scm, dvc, caplog):
+    tmp_dir.dvc_gen({"dir": {"file1": "file1 content"}})
+
+    tmp_dir.gen({"dir": {"file2": "file2 content"}})
+
+    with pytest.raises(OverlappingOutputPathsError) as e:
+        dvc.add(os.path.join("dir", "file2"))
+    assert str(e.value) == (
+        "Cannot add 'dir/file2', because it is overlapping with other DVC "
+        "tracked output: 'dir'.\nTo include 'dir/file2' in 'dir', run "
```
efiop: Also please create a new line after `\n` here.

dvc/repo/add.py
pared: nah, black gets in the way only when lines do not match, I probably did that to save some lines :P

efiop:
@pared Also check the tests, please.

dvc/repo/__init__.py
```diff
@@ -334,7 +334,17 @@ def _collect_graph(self, stages=None):
             for out in stage.outs:
                 for p in out.path_info.parents:
                     if p in outs:
-                        raise OverlappingOutputPathsError(outs[p], out)
+                        msg = (
+                            "Paths for outs:\n'{}'('{}')\n'{}'('{}')\noverlap."
+                            " To avoid unpredictable behaviour, "
```
efiop: Please use consistent formatting. You skip trailing space in the first line but add it in the second one. IIRC prefered way to have `str ` and not ` str`

jorgeorpinel: Agree about consistency. Idk what's preferred though, I used to put trailing space last but later I noticed most of you use trailing first so rn there's a mix throughout docs. Maybe trailing space first is best because it makes it extra obvious that it's a continuation string? Up to you, I just rec to be consistent (available to review all existing strings once you decide, please lmk).

shcheklein:
minor: we use sometimes outs sometimes outputs/output in this PR - would be probably good to use one of those consistently (probably output/outputs to avoid slang-slang)

PR #3297, opened 2020-02-10
checkout: fix output message

### LINKED ISSUE #2979: checkout/add: change warning messages to regular/info if behavior is expected
```
$ dvc -V
0.75.0
```

```
$ dvc checkout
WARNING: data 'newfile.txt' exists. Removing before checkout.
WARNING: data 'ttt.txt' exists. Removing before checkout.
```

EDITED 12/18/19: The message is actually wrong! The files were not removed - they were replaced by a new version. It has to be fixed as well.

```
$ echo hello >> newfile.txt
$ dvc add newfile.txt
WARNING: Output 'newfile.txt' of 'newfile.txt.dvc' changed because it is 'modified'
100% Add|███████████████████████████████████████████|1.00/1.00 [00:00<00:00,  2.45file/s]

	git add newfile.txt.dvc
```

Based on my understanding warnings are needed when you need to alert a user. It is not needed for routine operations with expected results. In both of these cases, the priority needs to be reduced to regular/info.

PS: Are there any other warnings like these?

skshetry:

Fixes #2979.

I tried adding a better message, but could not think any better message. Suggestions will be appreciated.

I also did not like to have `logger.info` in `dvc.state`, so it needed a refactor. :slightly_smiling_face:

### DIFF
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -878,8 +878,8 @@ class RemoteBASE(object):
         """The file is changed we need to checkout a new copy"""
         cache_info = self.checksum_to_path_info(checksum)
         if self.exists(path_info):
-            logger.warning(
-                "data '{}' exists. Removing before checkout.", path_info
+            logger.info(
+                "The file '{}' already exists. It will be replaced.", path_info
             )
             self.safe_remove(path_info, force=force)
 
--- a/dvc/repo/checkout.py
+++ b/dvc/repo/checkout.py
@@ -15,7 +15,14 @@ def _cleanup_unused_links(repo):
         for out in stage.outs
         if out.scheme == "local"
     ]
-    return repo.state.remove_unused_links(used)
+
+    unused = repo.state.get_unused_links(used)
+    for link in unused:
+        logger.info(
+            "Removing '{}' as it already exists in the current worktree.", link
+        )
+    repo.state.remove_unused_links(unused)
+    return bool(unused)
 
 
 def get_all_files_numbers(pairs):
--- a/dvc/repo/checkout.py
+++ b/dvc/repo/checkout.py
@@ -21,7 +21,7 @@ def _cleanup_unused_links(repo):
         logger.info(
             "Removing '{}' as it already exists in the current worktree.", link
         )
-    repo.state.remove_unused_links(unused)
+    repo.state.remove_links(unused)
     return bool(unused)
 
 
--- a/dvc/repo/checkout.py
+++ b/dvc/repo/checkout.py
@@ -15,7 +15,7 @@ def _cleanup_unused_links(repo):
         for out in stage.outs
         if out.scheme == "local"
     ]
-    repo.state.remove_unused_links(used)
+    return repo.state.remove_unused_links(used)
 
 
 def get_all_files_numbers(pairs):
@@ -34,9 +34,10 @@ def _checkout(
 ):
     from dvc.stage import StageFileDoesNotExistError, StageFileBadNameError
 
+    cleaned = False
     if not targets:
         targets = [None]
-        _cleanup_unused_links(self)
+        cleaned = _cleanup_unused_links(self)
 
     pairs = set()
     for target in targets:
@@ -52,7 +53,7 @@ def _checkout(
             raise CheckoutErrorSuggestGit(target) from exc
 
     total = get_all_files_numbers(pairs)
-    if total == 0:
+    if total == 0 and not cleaned:
         logger.info("Nothing to do")
     failed = []
     with Tqdm(
--- a/dvc/state.py
+++ b/dvc/state.py
@@ -467,7 +467,10 @@ class State(object):  # pylint: disable=too-many-instance-attributes
             actual_mtime, _ = get_mtime_and_size(path, self.repo.tree)
 
             if inode == actual_inode and mtime == actual_mtime:
-                logger.debug("Removing '{}' as unused link.", path)
+                logger.info(
+                    "Removing '{}' as it is unused in the current worktree.",
+                    relpath,
+                )
                 remove(path)
                 unused.append(relpath)
 
@@ -478,6 +481,8 @@ class State(object):  # pylint: disable=too-many-instance-attributes
                 self.LINK_STATE_TABLE, ",".join(["?"] * len(chunk_unused))
             )
             self._execute(cmd, tuple(chunk_unused))
+
+        return bool(unused)
 
 
 def _connect_sqlite(filename, options):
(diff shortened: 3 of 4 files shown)


pared:
> I also did not like to have logger.info in dvc.state, so it needed a refactor.

@skshetry but is there any particular reason why `logging.info` should not be in `state`?

tests/func/test_state.py
```diff
@@ -78,7 +79,24 @@ def test_remove_unused_links(tmp_dir, dvc):
...
-        dvc.state.remove_unused_links([])
+        dvc.state.remove_unused_links(["foo", "bar"])
 
         result = dvc.state._execute(cmd_count_links).fetchone()[0]
         assert result == 0
+
+
+def test_get_unused_links(tmp_dir, dvc):
+    assert len(tmp_dir.dvc_gen("foo", "foo_content")) == 1
+    assert len(tmp_dir.dvc_gen("bar", "bar_content")) == 1
```
pared: I don't think we have to assert that `dir_helpers` work properly.
Also, we can achieve that using `tmp_dir.dvc_gen({"foo":"foo content", "bar": "bar content"})`

skshetry: I just copy-pasted from above. :stuck_out_tongue: Fixing ...

skshetry:
> @skshetry but is there any particular reason why `logging.info` should not be in `state`?

First of all, I tried to maintain consistency: `state` does not have `logger.info` anywhere. Therefore, I did not try to introduce one, because I wanted to separate concerns (ui vs state/model) wherever possible, and it seemed like a good idea here for doing just that. :slightly_smiling_face:

dvc/state.py
```diff
@@ -467,9 +467,18 @@ def remove_unused_links(self, used):
...
 
             if inode == actual_inode and mtime == actual_mtime:
-                logger.debug("Removing '{}' as unused link.", path)
-                remove(path)
-                unused.append(relpath)
+                unused.append(relative_path)
+
+        return unused
+
+    def remove_unused_links(self, unused):
```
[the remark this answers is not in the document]
skshetry: Fixed, thanks.

tests/func/test_state.py
skshetry: Updated. :slightly_smiling_face:

dvc/repo/checkout.py
```diff
@@ -15,7 +15,14 @@ def _cleanup_unused_links(repo):
         for out in stage.outs
         if out.scheme == "local"
     ]
-    repo.state.remove_unused_links(used)
+
+    unused = repo.state.get_unused_links(used)
+    for link in unused:
+        logger.info(
+            "Removing '{}' as it already exists in the current worktree.", link
```
shcheklein: cc @jorgeorpinel - please review the messages in this PR

skshetry: I couldn't think of succinct messages. :cry:

dvc/remote/base.py
```diff
@@ -878,8 +878,8 @@ def _checkout_file(
         """The file is changed we need to checkout a new copy"""
         cache_info = self.checksum_to_path_info(checksum)
         if self.exists(path_info):
-            logger.warning(
-                "data '{}' exists. Removing before checkout.", path_info
+            logger.info(
+                "The file '{}' already exists. It will be replaced.", path_info
             )
```
efiop: Not sure about converting it to info. Maybe let's remove it for good? Or make it `debug`?

dvc/repo/checkout.py
efiop: The message is not very useful for `info`, IMHO. How about we  move it to `debug`?

efiop: Also, it is pretty weird to split old `remove_unused_links` just to print a message here. Or am I missing something?

dvc/state.py
```diff
@@ -443,36 +443,37 @@ def save_link(self, path_info):
...
-                remove(path)
-                unused.append(relpath)
+            if (inode, mtime) == (actual_inode, actual_mtime):
+                unused.append(relative_path)
+
+        return unused
+
+    def remove_links(self, links):
+        for link in links:
+            remove(os.path.join(self.root_dir, link))
```
efiop: A more natural refactor would split old `remove_unused_links` into `get_unused_links` and `remove_link(s)`(as a reverse `save_link`). So we would `get_unused_links`, remove them in `checkout` and then `state.remove_links()` what we've deleted. That way the context would be properly spread across these parts of code.

dvc/remote/base.py
skshetry: This was the result of the discussion from #2979, no?
Also: https://github.com/iterative/dvc/issues/2979#issuecomment-567301148:

dvc/repo/checkout.py
skshetry: > Also, it is pretty weird to split old `remove_unused_links` just to print a message here. Or am I missing something?

yes: https://github.com/iterative/dvc/pull/3297#issuecomment-584204079

dvc/remote/base.py
efiop: @skshetry Good point, that needs to be clarified. Left a comment there.

dvc/repo/checkout.py
efiop: > First of all, I tried to maintain consistency: state does not have logger.info anywhere. Therefore, I did not try to introduce one, because I wanted to separate concerns (ui vs state/model) wherever possible, and it seemed like a good idea here for doing just that. slightly_smiling_face

So you moved it just to not use `logger.info`? That is a pretty weak argument for doing that, tbh :slightly_smiling_face: Please see https://github.com/iterative/dvc/pull/3297#discussion_r377620466 , I think that it should be like that or we should keep `remove_unused_links` in one piece. Let me know what you think.

skshetry: I don't consider separation of UI and business logic to be a weak argument. But probably as it's a cli app and already done this way on most of the places, maybe it is.

>  I think that it should be like that or we should keep remove_unused_links in one piece. Let me know what you think.

I am fine, either way. Previously, I did not want to separate those too far off like you suggested below and tried to just get the stuff done. :see_no_evil:
It does not really need to be separated if `logger.info` call is kept on `dvc.state` as it's only used in a single place.

skshetry:
Closing this PR in favor of #3326

dvc/repo/checkout.py
jorgeorpinel: > cc @jorgeorpinel - please review the messages in this PR

I see #3326 replaced this and only has a debug string change

PR #3298, opened 2020-02-10
dvc: refactor config

Suor:
Advantages:
- read like a dict
- modify as a dict for testing or manipulation purposes
- wrap modifications into context manager to save them:
    ```python
    with config.edit("local") as conf:
        conf["core"]["remote"] = "new-default"

Design decisions:
- dropped string constants
- remote sections are parsed into a simple subdict upon load
- remote config schema is now defined by url scheme
- relative paths are handled transparently upon load/save
- user config and remote manipulations are moved to commands
- same dev manipulations are done by simply modifying a config dict


efiop:
@Suor Looks amazing! Please check the tests though, looks like slight hickup with schemas for some remotes.

Suor:
@efiop I see two issues: 
- setting `type` for hdfs remote fails, which makes sense, tests need to be adjusted
- windows paths not being recognized sometimes by validation

I will fix these tomorrow.

efiop:
@Suor Please rebase.

Suor:
@efiop rebased

dvc/remote/s3.py
```diff
@@ -26,35 +26,33 @@ class RemoteS3(RemoteBASE):
...
         else:
             self.list_objects_api = "list_objects_v2"
 
-        self.use_ssl = config.get(Config.SECTION_AWS_USE_SSL, True)
+        self.use_ssl = config.get("use_ssl", True)
 
         self.extra_args = {}
 
-        self.sse = config.get(Config.SECTION_AWS_SSE, "")
+        self.sse = config.get("sse")
```
pared: Shouldn't here be `config.get("sse","")`? And if not, why do we do `config.get("acl","")`?

dvc/remote/local.py
```diff
@@ -46,22 +36,16 @@ class RemoteLOCAL(RemoteBASE):
...
         self._file_mode, self._dir_mode = self.SHARED_MODE_MAP[shared]
 
         if self.protected:
             # cache files are set to be read-only for everyone
             self._file_mode = stat.S_IREAD | stat.S_IRGRP | stat.S_IROTH
 
-        cache_dir = config.get(Config.SECTION_REMOTE_URL)
-
-        if cache_dir is not None and not os.path.isabs(cache_dir):
-            cwd = config[Config.PRIVATE_CWD]
```
efiop: @Suor You might've lost `_cwd` logic, which was resolving relative paths based on config location. Or am I missing something here?

shcheklein: I thought that we have test for this 🤔

dvc/remote/s3.py
Suor: Doesn't matter really until it's boolean false it won't be used anyway.

dvc/remote/local.py
Suor: This is not needed anymore:

> relative paths are handled transparently upon load/save

efiop: @Suor Could you please refer me to some part of the code?

dvc/remote/s3.py
Suor: Removed `""` for consistency.

dvc/remote/local.py
efiop: Nevermind, I see it now.

efiop:
@Suor Please check the DS and CC reports, there are some good points there.

dvc/config.py
```diff
@@ -300,250 +261,134 @@ def init(dvc_dir):
...
+        return conf
 
-    def load(self):
-        """Loads config from all the config files.
+    @staticmethod
+    def _load_paths(conf, filename):
+        def resolve(path):
+            if os.path.isabs(path) or re.match(r"\w+://", path):
+                return path
+            return RelPath(os.path.join(cwd, path))
```
efiop: cwd is defined after this helper in which it is used. This is error-prone.

Suor: Why?

efiop: This is minor. Nothing functionally wrong about it because of the same namespace, but causes a mental hickup.

Suor:
@efiop I actually looked them through a couple of days ago) Most of them not introduced by this PR, from the rest:
- `ByUrl` complexity [here](https://codeclimate.com/github/iterative/dvc/pull/3298) - splitting it is not a good idea
- not calling `dict.__init__()` [here](https://deepsource.io/gh/iterative/dvc/run/cadeccf3-9db0-4ab4-8882-658d7061de90/python/PYL-W0231/) - `Config` has very different constructor signature.
- `self` in `.validate()` [here](https://deepsource.io/gh/iterative/dvc/run/cadeccf3-9db0-4ab4-8882-658d7061de90/python/PYL-R0201/) and `if/elif` [here](https://deepsource.io/gh/iterative/dvc/run/cadeccf3-9db0-4ab4-8882-658d7061de90/python/PYL-R1705/) just nitpicks, can't cause any issues
- `get_dir()` [here](https://deepsource.io/gh/iterative/dvc/run/cadeccf3-9db0-4ab4-8882-658d7061de90/python/PYL-R1710/) protected by an assert.

While some stylistic changes can be done, that will bring no value. Diving into DS was a waste of time again.

dvc/config.py
Suor: Defining subfunction in the middle causes mental hiccup too)

Suor: I can move it if that worries you.

efiop:
@Suor 

> ByUrl complexity here - splitting it is not a good idea


:+1: 

> not calling dict.__init__() here - Config has very different constructor signature.

:+1: 

> self in .validate() here and if/elif here just nitpicks, can't cause any issues

These are worth fixing, it affects the code style, we avoid such things in other places.

> get_dir() here protected by an assert.

:+1: 

> While some stylistic changes can be done, that will bring no value. Diving into DS was a waste of time again.

Style is not a waste of time for DVC, please fix those few issues left.

efiop:
Thanks @Suor ! :pray:

PR #3299, opened 2020-02-10
diff: use `rev` instead of `ref`

### LINKED ISSUE #3245: term: replace (Git) "reference" for "revision" or "commit"
Revisions are slightly more accurate - https://git-scm.com/docs/revisions

> A revision parameter '<rev>' typically, but not necessarily, names a commit object.

References are the ones living in `.git/refs`: https://git-scm.com/book/en/v2/Git-Internals-Git-References

We use the term _references_ in `metrics` and `diff` commands, but both commands accept revisions, since `repo.brancher` accepts `revs`(not `refs`).

This change would also require updating the docs.

efiop:
Fixes #3245

### DIFF
--- a/dvc/repo/diff.py
+++ b/dvc/repo/diff.py
@@ -6,7 +6,7 @@ from dvc.scm.git import Git
 
 
 @locked
-def diff(self, a_ref="HEAD", b_ref=None):
+def diff(self, a_rev="HEAD", b_rev=None):
     """
     By default, it compares the working tree with the last commit's tree.
 
@@ -47,8 +47,8 @@ def diff(self, a_ref="HEAD", b_ref=None):
         return result
 
     working_tree = self.tree
-    a_tree = self.scm.get_tree(a_ref)
-    b_tree = self.scm.get_tree(b_ref) if b_ref else working_tree
+    a_tree = self.scm.get_tree(a_rev)
+    b_tree = self.scm.get_tree(b_rev) if b_rev else working_tree
 
     try:
         self.tree = a_tree
--- a/dvc/command/diff.py
+++ b/dvc/command/diff.py
@@ -95,7 +95,7 @@ class CmdDiff(CmdBase):
 
     def run(self):
         try:
-            diff = self.repo.diff(self.args.a_ref, self.args.b_ref)
+            diff = self.repo.diff(self.args.a_rev, self.args.b_rev)
 
             if not any(diff.values()):
                 return 0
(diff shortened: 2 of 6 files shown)


dvc/repo/diff.py
```diff
@@ -6,7 +6,7 @@
 
 
 @locked
-def diff(self, a_ref="HEAD", b_ref=None):
+def diff(self, a_rev="HEAD", b_rev=None):
     """
     By default, it compares the working tree with the last commit's tree.
```
jorgeorpinel: working tree -> current workspace in docstrings too? 🙂

shcheklein:
@johntharian I think we discussed this already and agreed on `--rev`? Otherwise it will be way bigger change (will involve APIs, get/import, update, etc, etc). I don't see a strong reason do this.

jorgeorpinel:
OK.

Unrelated: wrong handle there though haha. Why does GH suggest users randomly instead of the ones we work with the most?

dvc/repo/diff.py
efiop: @jorgeorpinel Would rather not do that, as we use `WorkingTree` in the code a lot and that would break the symmetry.

jorgeorpinel: I totally understand. But we have asymmetry in terminology between core and docs e.g. this case (rev vs commit) working tree vs. workpsace, project vs repo... I'm not sure how big of a deal it is but we may need to address this at some point.

efiop: @jorgeorpinel I agree, hence why we are changing docs and CLI/API. But changing code and comments is far from being as useful, so it can wait for now.

efiop:
@jorgeorpinel Btw, is there a WIP PR that fixes the naming in the docs? If not, I could send a PR to fix stuff related to this PR.

jorgeorpinel:
Thanks again @efiop and sorry to have noticed this late: but I think some instances are missing to fully close https://github.com/iterative/dvc/issues/3245#issuecomment-584414556, if that was the intention of this PR.

jorgeorpinel:
p.s. I can submit a separate PR if you'd like.

shcheklein:
the intention was to change `ref` to `rev`, not to get rid of them in favor of `commit`. Do we have `ref` somewhere (or `reference`) still?

shcheklein:
I see we also changed some help messages. Yes, then we should change them all indeed.

jorgeorpinel:
> is there a WIP PR that fixes the naming in the docs?

Nope, good catch @efiop, this needs a corresponding PR in docs indeed (unchecked the 2nd checkbox). Thanks!

> the intention was to change ref to rev

Right, that's what the title suggests @shcheklein but it was also marked as fixing _term: replace (Git) "reference" and "revision" for "commit"_ #3245 so I wasn't sure 🙂

Anyway, I reopenned that issue and indicated the missing instancs – if we're doing that now maybe the corresponding docs update can also include those BTW, Ruslan. Again, lmk if you need help, it looks like a small thing I could do myself.

jorgeorpinel:
> > is there a WIP PR that fixes the naming in the docs? If not, I could send a PR to fix stuff related to this PR.
>
> this needs a corresponding PR in docs indeed (unchecked the 2nd checkbox)

Hey @efiop don't worry about it, I'll do this myself in iterative/dvc.org/pull/933

efiop:
@jorgeorpinel Sorry, was about to submit a PR for that. Will clarify this before merging next time. Thank you!

PR #3301, opened 2020-02-10
repro: fix bug in _get_active_graph

efiop:
We are currently removing all of the deps stages from the DAG even if
those deps are referenced by some parallel branches of execution.
The new approach is to remove edges and then gradually remove nodes if
there are no more `in` edges left.

Discord context:
https://discordapp.com/channels/485586884165107732/485596304961962003/676464313942409265


efiop:
For the record: Waiting for confirmation from the user.

mdekstrand:
Original bug reporter here - I have tested this change in my repository, and it now works as expected.

PR #3309, opened 2020-02-12
get/import: retrieve files inside directory outs

### LINKED ISSUE #2458: dvc: provide granularity for commands that could target specific tracked files
> This can be seen as revisiting feature request #1026

UPDATE: Please scroll down to https://github.com/iterative/dvc/issues/2458#issuecomment-566649020 for most recent, summarized requirement.
Here is the original context also (still relevant):

---

There's different scenarios in which _being able to manipulate files granularly independently of how they were committed/pushed to DVC_ could be useful. The problem with using `dvc add -R` now is that it can generate lots of `.dvc` files, but what if a directory could be added without `-R` (producing a single DVC-file) and yet other commands (lock, update, get, etc) could be applied to individual files inside the added directory tree?

**Example** (from https://github.com/iterative/dataset-registry/commit/7476a858f6200864b5755863c729bff41d0fb045)

Project 1:
```console
$ tree
.
└── tutorial
    └── nlp
        ├── Posts.xml.zip
        └── pipeline.zip
$ dvc add tutorial
...
$ dvc push
...
```

Project 2:
```console
$ dvc import {project-1-url} tutorial/nlp/pipeline.zip
...
$ tree
.
├── tutorial
│   └── nlp
│       └── pipeline.zip
└── tutorial.dvc
```
> Not sure about where the `.dvc` would have to be placed in this example though.

And also this is how Git works, I believe. Files are tracked individually (in fact it doesn't even recognize empty dirs).

mroutis:
Close #2458

### DIFF
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -7,13 +7,15 @@ import threading
 
 from funcy import retry, suppress, wrap_with, cached_property
 
+from dvc.path_info import PathInfo
 from dvc.compat import fspath
 from dvc.repo import Repo
 from dvc.config import NoRemoteError, NotDvcRepoError
 from dvc.exceptions import NoRemoteInExternalRepoError
 from dvc.exceptions import OutputNotFoundError, NoOutputInExternalRepoError
 from dvc.exceptions import FileMissingError, PathMissingError
-from dvc.utils.fs import remove, fs_copy
+from dvc.utils.fs import remove, fs_copy, move
+from dvc.utils import tmp_fname
 from dvc.scm.git import Git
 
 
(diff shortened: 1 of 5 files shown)


tests/func/test_import.py
```diff
@@ -102,6 +102,42 @@ def test_import_dir(tmp_dir, scm, dvc, erepo_dir):
     }
 
 
+def test_get_file_from_dir(tmp_dir, scm, dvc, erepo_dir):
```
pared: `test_import_file_from_dir`?

dvc/external_repo.py
```diff
@@ -68,32 +70,49 @@ def __init__(self, root_dir, url):
...
+        checkout it into `to_info`.
+
+        It works with files tracked by Git and DVC, and also local files
+        outside the repository.
+        """
+        out = None
+        path_info = PathInfo(self.root_dir) / path
+
+        with suppress(OutputNotFoundError):
+            (out,) = self.find_outs_by_path(path_info, strict=False)
```
pared:
```suggestion
            (out,) = self.find_outs_by_path(fspath(path_info), strict=False)
```
3.5 will fail with PathInfo

mroutis: thanks, @pared :+1:

jorgeorpinel:
I  think this is missing a docs PR @efiop @pared

jorgeorpinel:
UPDATE: Created iterative/dvc.org/issues/1011 but feel free to submit an initial PR if anyone has the capacity.

PR #993, opened 2018-08-10
Added hall-of-fame to README

sergey48k:
Hey guys, this is how Hall-of-fame will look like on your repo.  I spoke to @dmpetrov about this PR before making it.

Note: it currently uses my GitHub user (sergey48k) for GitHub API. This takes ~10-20 requests off my limit. Not a biggie, but in case you are wondering why my username is in your source code, it's the reason why. I could also switch to any of you, but the only qualified DVC contributor is @efiop because only he has Sourcerer account linked with a GitHub account.

Also, note. If you accept this pull request, I will become a 'new' contributor to DVC, so my face will likely be on one of the badges for a few days. If you don't want it (and I won't be offended, don't worry), don't accept this request, but just copy the changes, and make a new pull request.

I hope you like it.
Best,
Sergey

PR #3312, opened 2020-02-12
imp_url: make default stage fname accompany import target

### LINKED ISSUE #3304: import: behave like `add`, `move`
> More context around https://discordapp.com/channels/485586884165107732/485596304961962003/676398069469806603

Currently `dvc import` has a `-o` option to specify where to place the downloaded data, but the DVC-file itself gets created in the CWD. It would be more intuitive to have the DVC-file in the same `--out` location, similar to how `dvc add` behaves.

This is especially strange when using `dvc move` on an import stage, as it moves both the data and the DVC-file into the `dst` location.

May apply to `dvc import-url` as well, haven't checked.

pared:

------
Fixes #3304

### DIFF
--- a/dvc/stage.py
+++ b/dvc/stage.py
@@ -502,12 +502,11 @@ class Stage(object):
         return True
 
     @staticmethod
-    def create(repo, **kwargs):
+    def create(repo, accompany_outs=False, **kwargs):
 
         wdir = kwargs.get("wdir", None)
         cwd = kwargs.get("cwd", None)
         fname = kwargs.get("fname", None)
-        add = kwargs.get("add", False)
 
         # Backward compatibility for `cwd` option
         if wdir is None and cwd is not None:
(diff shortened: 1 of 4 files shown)


dvc/stage.py
```diff
@@ -502,12 +502,11 @@ def is_cached(self):
         return True
 
     @staticmethod
-    def create(repo, **kwargs):
+    def create(repo, accompany_outs=False, **kwargs):
```
pared: I was the one to introduce the `kwargs` instead of list of arguments, but it seems to me that I overdone that, arguments required in this method should be present in arguments list and if necessary, passed down. `kwargs` occludes what is necessary for this method.

jorgeorpinel:
Thanks!

> Check this box if this PR does not require documentation updates

Were you able to check this doesn't require doc updates? Maybe in the `dvc import` examples? If you find any you can just open/find a corresponding issue in iterative/dvc.org

jorgeorpinel:
UPDATE: Not in examples, but I think the description of `-o` in https://dvc.org/doc/command-reference/import#options needs a small update. Unchecked box

jorgeorpinel:
OK never mind, I included this in #993 but please check the docs in future PRs before checking the box. Thanks!

PR #3315, opened 2020-02-12
git: quote filenames at remind_to_track

### LINKED ISSUE #3303: `dvc add` print not quote-wrapped file-paths if they contain spaces
dvc 0.82.8
installed via yum
centos 7

Some of our, khm, windows-users put files in project with spaces in names.
In this case `dvc add`, when prints hint-command `git add blah-blah`, not wrap path with quotes so copy-paste this hint-command not working ofc.

mroutis:
Fix #3303

### DIFF
--- a/tests/unit/scm/test_scm.py
+++ b/tests/unit/scm/test_scm.py
@@ -37,3 +37,9 @@ class TestScmContext(TestCase):
 
         self.assertEqual(0, self.scm_mock.reset_ignores.call_count)
         self.assertEqual(0, self.scm_mock.remind_to_track.call_count)
+
+
+def test_remind_to_track(scm, caplog):
+    scm.files_to_track = ["fname with spaces.txt", "тест", "foo"]
+    scm.remind_to_track()
+    assert "git add 'fname with spaces.txt' 'тест' foo" in caplog.text
--- a/dvc/scm/git/__init__.py
+++ b/dvc/scm/git/__init__.py
@@ -3,6 +3,7 @@
 import logging
 import os
 import yaml
+import shlex
 
 from funcy import cached_property
 from pathspec.patterns import GitWildMatchPattern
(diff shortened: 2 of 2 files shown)


tests/unit/scm/test_scm.py
```diff
@@ -37,3 +37,9 @@ def test_should_throw_and_cleanup(self):
 
         self.assertEqual(0, self.scm_mock.reset_ignores.call_count)
         self.assertEqual(0, self.scm_mock.remind_to_track.call_count)
+
+
+def test_remind_to_track(scm, caplog):
+    scm.files_to_track = ["fname with spaces.txt", "тест", "foo"]
+    scm.remind_to_track()
+    assert "git add 'fname with spaces.txt' 'тест' foo" in caplog.text
```
efiop: I remember windows cmdline being weird about single-quotes and having to use double-quotes. Also IIRC windows cmd always does completion with double-quotes. Searching around (e.g. https://ss64.com/nt/syntax-esc.html) I also see double quotes used and not single-quotes. Let's switch to double-quotes too, it will be compatible with both win and *nix.

casperdcl: well *nix will shell-expand things in double quotes:

```bash
$ export foo=bar
$ echo spam > \$foo
$ cat "$foo"  # double quotes :(
cat: bar: No such file or directory
```

casperdcl: Also @efiop this PR uses `shlex.quote` which avoids this

efiop: @casperdcl Yep, there is also `mslex` which properly quotes things for windows, but it doesn't have a conda package. I'm wondering if it is easier to just make dvc automatically `git add` things at this point. Easier for both us on implementation and for users in terms of UX.

casperdcl: yes not sure if I've mentioned this somewhere before but simply adding `--git` to any `dvc` command to auto-exec the obvious would be nice

efiop:
Merging as is, we'll see about mslex in the future, for now this will cover 90% (or more) of use cases.

Issue #3060, opened 2020-01-05
UI progress


casperdcl:
- [ ] add progress to other uses of `stage.dump`
- [ ] profile `repo.check_modified_graph()`, `stage.save()`, `stage.dump()` etc to find out what other functions are slow, perhaps add progress for them too
  + md5sums only take about 10% of the total time. There must be other more time consuming functions
- [ ] make sure that we show HINT message to enable cache optimization properly when copy is taking too long already
- continues #2699, #2846

casperdcl:
> make sure that we show HINT message to enable cache optimization properly when copy is taking too long already

@shcheklein is this done already? I seem to recall it may have been mentioned in a different PR.

shcheklein:
@casperdcl so, I'm not sure to be honest. It's super important and should be done in a way that is visible when certain thresholds are passed. I don't remember what is current policy even. I would revisit them.

efiop:
Current policy is per-link, so it only shows a warning if one individual link takes longer than 10 seconds. It doesn't account for like, 2 links of 5 seconds or anything like that. Feels like we could simply lower the threshold to improve the experience, but need to make `protected = True` by default first.

shcheklein:
First thing I would say for this ticket is make sure that it is visible clearly with all these changes to progress bars and whatnot.

PR #3316, opened 2020-02-12
ui: make progress bars respect `--quiet`

### LINKED ISSUE #3314: dvc push -q is not completely quiet, shows file transfer tqdm progress bars
Hey guys, love what you've done with DVC.

Had a quick bug that's causing me a little issue. When I use 'dvc push -q' I'm still seeing tqdm progress bars. Wouldn't be a huge issue, but I'm probably pushing 100K 250kb files. This is a local remote, so the transfer speeds are quick. I know in some of my other scripts where I use tqdm, if the iteration time is very small, the tqdm overhead of writing to std::out actually starts to contribute to performance.

dvc version: 0.83.0
os: Windows 10

image

casperdcl:
- [x] change to `no_progress_bar=None` default instead of `False`
   + this respects automatic `logger.getEffectiveLevel()` logic
- [x] fix a couple of other places missing automatic logic
- [x] perhaps add test
- [x] perhaps treat `disable=False` the same as `disable=None` instead
- fixes #3314
- related #3060

### DIFF
--- a/dvc/lock.py
+++ b/dvc/lock.py
@@ -44,7 +44,7 @@ class Lock(object):
         try:
             with Tqdm(
                 bar_format="{desc}",
-                disable=True if not self._friendly else None,
+                disable=not self._friendly,
                 desc=(
                     "If DVC froze, see `hardlink_lock` in {}".format(
                         format_link("man.dvc.org/config#core")
--- a/dvc/lock.py
+++ b/dvc/lock.py
@@ -44,7 +44,7 @@ class Lock(object):
         try:
             with Tqdm(
                 bar_format="{desc}",
-                disable=not self._friendly,
+                disable=True if not self._friendly else None,
                 desc=(
                     "If DVC froze, see `hardlink_lock` in {}".format(
                         format_link("man.dvc.org/config#core")
--- a/dvc/remote/azure.py
+++ b/dvc/remote/azure.py
@@ -108,7 +108,7 @@ class RemoteAZURE(RemoteBASE):
         return self._list_paths(self.path_info.bucket, self.path_info.path)
 
     def _upload(
-        self, from_file, to_info, name=None, no_progress_bar=None, **_kwargs
+        self, from_file, to_info, name=None, no_progress_bar=False, **_kwargs
     ):
         with Tqdm(desc=name, disable=no_progress_bar, bytes=True) as pbar:
             self.blob_service.create_blob_from_path(
@@ -119,7 +119,7 @@ class RemoteAZURE(RemoteBASE):
             )
 
     def _download(
-        self, from_info, to_file, name=None, no_progress_bar=None, **_kwargs
+        self, from_info, to_file, name=None, no_progress_bar=False, **_kwargs
     ):
         with Tqdm(desc=name, disable=no_progress_bar, bytes=True) as pbar:
             self.blob_service.get_blob_to_path(
--- a/dvc/remote/azure.py
+++ b/dvc/remote/azure.py
@@ -108,7 +108,7 @@ class RemoteAZURE(RemoteBASE):
         return self._list_paths(self.path_info.bucket, self.path_info.path)
 
     def _upload(
-        self, from_file, to_info, name=None, no_progress_bar=False, **_kwargs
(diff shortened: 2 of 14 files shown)


dvc/lock.py
```diff
@@ -44,7 +44,7 @@ def _do_lock(self):
         try:
             with Tqdm(
                 bar_format="{desc}",
-                disable=not self._friendly,
+                disable=True if not self._friendly else None,
```
casperdcl: @efiop hope this change is compatible with what you intended by this line

efiop: @casperdcl btw, are there any reasons why Tqdm requires `None` and `False` is not enough?

efiop: A bit weird though (all of these True if ... else None) and really easy to mess up in the future. Maybe `Tqdm` from `progress.py` should ignore `disable` and always disable pbar when we are in quiet mode?

casperdcl: Well we could treat `disable=False` as we currently treat `disable=None`. That would require fewer changes but would be counterintuitive to me.

casperdcl: Also there are only a few cases where we need to override `None` to `True`...

casperdcl: I will say we should probably drop `no_progress_bar` in a future PR

efiop: @casperdcl Not sure I follow. I does seem to me that if we are in quiet mode, `disable=False` or `disable=None` should not print the progress bar, because loglevel logic has a higher priority over `disable` option.

casperdcl: Yes. Ideally, loglevel logic *should* have always controlled verbosity, including for progress bars.

This has been the case with `disable=None`.

However there are 2 exceptions. In both cases we want to disable even if `log.level == INFO`. This requires `disable=True if condition else None`.

If in future we ever come up with a reason to use `disable=False` then we can think about how to support it at that point. I'm even tempted to (for now) include an `assert disable is not False` statement.

efiop: @casperdcl Again, might be missing something, but why not replace:
```
 if disable is None:                               
     disable = logger.getEffectiveLevel() > level  
```
with
```
 if not disable:
     disable = logger.getEffectiveLevel() > level  
```
So if it is disabled then it is 100% disabled, but if it is not disabled then we will determine based on loglevel whether or not we actually need to enable it.

casperdcl: > Well we could treat `disable=False` as we currently treat `disable=None`. That would require fewer changes but would be counterintuitive to me.

casperdcl: exactly what I was suggesting but it seems a little bad programming practice to treat what is essentially an explicit `disable=False` as if it was `disable=None` (i.e. auto)

efiop: @casperdcl Sounds alright to me because loglevel has higher priority over `disable=False`.

efiop: @casperdcl Also, we need a test for this PR.

casperdcl: right will also add test

casperdcl:
any reason why `no_progress_bar=True` in these three places?

- `remote.gdrive.RemoteGDrive.gdrive_upload_file()`
- `remote.base.RemoteBASE._get_dir_info_checksum():self.cache.upload()`
- `remote.gs.RemoteGDrive.RemoteGS._upload()` and `_download()`

casperdcl:
@efiop I think this is good to merge. I don't think it's worth reverting the `no_progress_bar=None` changes as I intend removing most of them in a future PR (see #3332).

dvc/remote/azure.py
```diff
@@ -108,7 +108,7 @@ def list_cache_paths(self):
         return self._list_paths(self.path_info.bucket, self.path_info.path)
 
     def _upload(
-        self, from_file, to_info, name=None, no_progress_bar=False, **_kwargs
+        self, from_file, to_info, name=None, no_progress_bar=None, **_kwargs
```
efiop: Hm, do we need to adjust these though for this PR?

casperdcl: no we don't; just not sure it's worth dealing with now. Argh probably easier to just revert rather than debate. nvm.

PR #3335, opened 2020-02-14
ui: fix missing progress & tidy

### LINKED ISSUE #3332: missing progress
any reason why `no_progress_bar=True` in these three places?

- `remote.gdrive.RemoteGDrive.gdrive_upload_file()`
  + from e7227898e4e10a1 by @MaxRis
- `remote.base.RemoteBASE._get_dir_info_checksum():self.cache.upload()`
  + from 00cfb04fea04694 by @efiop
- `remote.gs.RemoteGDrive.RemoteGS._upload()` and `_download()`
  + from 28b99dc4c78a7bc by, er, @casperdcl

(related: #3060)

casperdcl:
make `no_progress_bar=False` rather than `True`

- [x] `remote.gdrive.RemoteGDrive.gdrive_upload_file()`
- [x] `remote.base.RemoteBASE._get_dir_info_checksum():self.cache.upload()`
- [x] `remote.gs.RemoteGDrive.RemoteGS._upload()` and `_download()`

----

- fixes #3332
- replaces/closes #3334
- depends on #3316

### DIFF
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -234,7 +234,7 @@ class RemoteBASE(object):
 
         from_info = PathInfo(tmp)
         to_info = self.cache.path_info / tmp_fname("")
-        self.cache.upload(from_info, to_info, no_progress_bar=False)
+        self.cache.upload(from_info, to_info, no_progress_bar=True)
 
         checksum = self.get_file_checksum(to_info) + self.CHECKSUM_DIR_SUFFIX
         return checksum, to_info
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -234,7 +234,7 @@ class RemoteBASE(object):
 
         from_info = PathInfo(tmp)
         to_info = self.cache.path_info / tmp_fname("")
-        self.cache.upload(from_info, to_info, no_progress_bar=True)
+        self.cache.upload(from_info, to_info, no_progress_bar=False)
 
         checksum = self.get_file_checksum(to_info) + self.CHECKSUM_DIR_SUFFIX
         return checksum, to_info
--- a/dvc/remote/gdrive.py
+++ b/dvc/remote/gdrive.py
@@ -246,7 +246,7 @@ class RemoteGDrive(RemoteBASE):
         self,
         parent_id,
         title,
-        no_progress_bar=True,
+        no_progress_bar=False,
         from_file="",
         progress_name="",
     ):
--- a/dvc/remote/gs.py
+++ b/dvc/remote/gs.py
@@ -51,7 +51,7 @@ def _upload_to_bucket(
     to_info,
     chunk_size=None,
     name=None,
-    no_progress_bar=True,
+    no_progress_bar=False,
 ):
     blob = bucket.blob(to_info.path, chunk_size=chunk_size)
     with io.open(from_file, mode="rb") as fobj:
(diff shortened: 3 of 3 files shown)


dvc/remote/base.py
```diff
@@ -256,7 +256,7 @@ def _get_dir_info_checksum(self, dir_info):
 
         from_info = PathInfo(tmp)
         to_info = self.cache.path_info / tmp_fname("")
-        self.cache.upload(from_info, to_info, no_progress_bar=True)
+        self.cache.upload(from_info, to_info, no_progress_bar=False)
```
casperdcl: @efiop wanted to ask if this change is OK

P.S.: ignore the other changes in this PR for now (they'll disappear after #3316 is merged).

casperdcl: just to add to this - if this change is OK then there's a lot of nice tidying/refactoring possible since `upload()` would never require a `no_progress_bar` argument

efiop: @casperdcl Btw, how does the pbar look for this operation? I'm worried about it having some random looking paths in it.

dvc/remote/gdrive.py
```diff
@@ -246,7 +246,7 @@ def gdrive_upload_file(
         self,
         parent_id,
         title,
-        no_progress_bar=True,
+        no_progress_bar=False,
```
efiop: I suppose these were here (and in gs down below) because they didn't support pbars earlier and we forgot to adjust them here. Nice catch!

dvc/remote/base.py
casperdcl: this would occur on `dvc add <dir>`. It gets proper file names and is cleared away upon completion. Padding it with some persistent logging messages (and padding `fs.copyfile` with some logging & `sleep(5)`) gives this:

[asciicast](https://asciinema.org/a/a9dhhofKVQbElIJhwr78T4WgZ)

efiop: @casperdcl Yeah, not sure about that. That random name is not very helpful. How about we put some desc instead? There is a `name` param for that `upload` method, so should be pretty simple. Though, not sure how to describe this in a user-friendly way. Btw, did this particular part of code cause any issues in your research? It works with tiny files and usually uploading them is very quick.

As to the name here, maybe something like "Computing effective hash for directory '{}'" would be good enough.

casperdcl: Hmm. I was interested in adding this just because it would then mean we can drop `no_progress_bar` from **all** `upload()` functions. I don't think it's really useful to have a progress bar for this `cache.upload()` function though so maybe we should leave it for now.

efiop: @casperdcl So it doesn't make dvc hang? We need to do research when doing such things.

casperdcl: No I haven't found any hanging. In any case there's a master progress bar above this line which increments when `cache.upload()` completes.

It's possible if in future we add a "don't display progress bar until 4 seconds have elapsed" https://github.com/tqdm/tqdm/issues/704 then it may be useful to enable progress for `cache.upload()` for very large files.

efiop: @casperdcl Got it. So the plan is to only refactor the defaults in those remotes, right?

casperdcl: If we allow `no_progress_bar=False` then we can refactor away. But perhaps for now we should keep `no_progress_bar=True` and not refactor.

efiop: @casperdcl Sorry, looks like I'm not following. So gdrive/gs had `no_progress_bar=True`, which contradicts the other remotes. So we will fix those, right? Or are we going to do something else as well? :slightly_smiling_face: Just trying to understand the plan for this PR.

casperdcl: I was (and have always been) only talking about this specific `cache.upload()` line. The PR is now fine to merge.

`gdrive`/`gs` things elsewhere in this PR are no-brainers I think.

casperdcl: Just to clarify further - if we change the `cache.upload()` line to say `no_progress_bar=False`, then it would mean every single instance of `upload()` in the code base would only ever have `no_progress_bar=False` and thus we can drop that argument from all `upload()` functions.

Issue #3247, opened 2020-01-28
dvc push for http remote storage


harold1505:
I wanted to upload and download data files over http. But, i came to know dvc only support `dvc fetch` and `dvc pull` for http remotes. Why is it so? How to make `dvc push` work for http remote? New to dvc, please help.

> ⚠️ HTTP remotes only support downloads operations

efiop:
Hi @harold1505 !

Are you using http to access AWS s3 or something else? Or do you want to use pure http remote?

harold1505:
It is neither AWS S3 nor a popular platform. It is a platform with Rest API in which files can be retrieved and restored via http. So I guess, I want a pure http remote.

efiop:
@harold1505 Got it :slightly_smiling_face: So the reason why it is not supported is simple: we just didn't get any requests for supporting it until now. :slightly_smiling_face: To implement it, we will need to implement `_upload` method in `dvc/remote/http.py`. There might be some edge cases that we need to handle carefully, but overall the implementation should be pretty straightforward. If you feel like it, maybe even consider taking a shot and contributing a PR for it, we will be happy to help with anything we can. :slightly_smiling_face:

ayushais:
It would be really nice if this feature is available.

xkortex:
Seconded! I think there is some nuance in HTTP posts larger than 2 GB, but there is the Content-Range header, which allows you to split up the transfer into chunks, and I think a few other techniques. But it would be great to be able to write a POST endpoint to roll your own DVC remote backend.

pmrowla:
As discussed with Ivan, I'll be working on implementing this

pmrowla:
Just to clarify, the desired behavior for this is to have basic push support via making HTTP PUT (or some standardized POST format) requests to `http://<remote_url>/<filename>`, and not push via some other HTTP based protocol like WebDAV (see also #1153)?

@harold1505 can you give some details on what exactly the platform you are using expects in terms of PUT/POST requests for file uploads?

The main reason I'm asking is that as far as I understand it, git's HTTP remote is read-only (like dvc) for "dumb" web servers. HTTP remote write/push is only supported for web servers that can talk git's ["smart" protocol](https://www.git-scm.com/docs/http-protocol#_smart_clients) or WebDAV.

efiop:
[...]

PR #3343, opened 2020-02-17
remotes.http: Support dvc push for http remotes

pmrowla:

  iterative/dvc.org/pull/1006

WIP PR for #3247 

* Adds support for POST uploads to http remotes
* Adds support for HTTP Basic and Digest authentication methods (for both upload and download operations)
    * `basic_auth` and `digest_auth` boolean options added to the http(s) remote configurations (digest takes precedence if both are set to true)
    * When no auth method is specified (or both are set to false), existing behavior is used (no auth method, assumes all HTTP endpoints are public for both upload and download)
    * Username/password combinations can be configured in the same way as ssh remotes
    * Password can be set in config or prompted on command line via `ask_password` config option

Still TODO:

* [x] Add functional/remote tests
* [x] Docs PR

### DIFF
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -103,7 +103,7 @@ LOCAL_COMMON = {
 }
 HTTP_COMMON = {
     "auth": All(Lower, Choices("basic", "digest", "custom")),
-    "custom_header": str,
+    "custom_auth_header": str,
     "user": str,
     "password": str,
     "ask_password": Bool,
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -101,6 +101,13 @@ LOCAL_COMMON = {
     "shared": All(Lower, Choices("group")),
     Optional("slow_link_warning", default=True): Bool,
 }
+HTTP_COMMON = {
+    "basic_auth": Bool,
+    "digest_auth": Bool,
+    "user": str,
+    "password": str,
+    "ask_password": Bool,
+}
 SCHEMA = {
     "core": {
         "remote": Lower,
@@ -169,8 +176,8 @@ SCHEMA = {
                     "gdrive_user_credentials_file": str,
                     **REMOTE_COMMON,
                 },
-                "http": REMOTE_COMMON,
-                "https": REMOTE_COMMON,
+                "http": {**HTTP_COMMON, **REMOTE_COMMON},
+                "https": {**HTTP_COMMON, **REMOTE_COMMON},
                 "remote": {str: object},  # Any of the above options are valid
             }
         )
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -102,8 +102,8 @@ LOCAL_COMMON = {
     Optional("slow_link_warning", default=True): Bool,
 }
 HTTP_COMMON = {
-    "basic_auth": Bool,
-    "digest_auth": Bool,
+    "auth": All(Lower, Choices("basic", "digest", "custom")),
+    "custom_header": str,
     "user": str,
     "password": str,
     "ask_password": Bool,
--- a/dvc/remote/http.py
+++ b/dvc/remote/http.py
@@ -90,7 +90,7 @@ class RemoteHTTP(RemoteBASE):
                         yield chunk
 
             response = self._request("POST", to_info.url, data=chunks())
-            if response.status_code != 200:
+            if response.status_code not in (200, 201):
                 raise HTTPError(response.status_code, response.reason)
 
     def exists(self, path_info):
@@ -131,9 +131,9 @@ class RemoteHTTP(RemoteBASE):
                 self.password = ask_password(host, user)
             if self.auth == "basic":
                 return HTTPBasicAuth(path_info.user, self.password)
-            elif self.auth == "digest":
+            if self.auth == "digest":
                 return HTTPDigestAuth(path_info.user, self.password)
-            elif self.auth == "custom" and self.custom_auth_header:
+            if self.auth == "custom" and self.custom_auth_header:
                 self.headers.update({self.custom_auth_header: self.password})
         return None
 
--- a/dvc/remote/http.py
+++ b/dvc/remote/http.py
@@ -49,10 +49,11 @@ class RemoteHTTP(RemoteBASE):
                 "files. Use: `dvc remote modify <name> no_traverse true`"
             )
 
-        self.basic_auth = config.get("basic_auth", False)
-        self.digest_auth = config.get("digest_auth", False)
+        self.auth = config.get("auth", None)
+        self.custom_header = config.get("custom_header", None)
         self.password = config.get("password", None)
         self.ask_password = config.get("ask_password", False)
+        self.headers = {}
 
     def _download(self, from_info, to_file, name=None, no_progress_bar=False):
         response = self._request("GET", from_info.url, stream=True)
@@ -124,13 +125,16 @@ class RemoteHTTP(RemoteBASE):
         if path_info is None:
             path_info = self.path_info
 
-        if self.basic_auth or self.digest_auth:
+        if self.auth:
             if self.ask_password and self.password is None:
                 host, user = path_info.host, path_info.user
                 self.password = ask_password(host, user)
-            if self.digest_auth:
+            if self.auth == "basic":
+                return HTTPBasicAuth(path_info.user, self.password)
+            elif self.auth == "digest":
                 return HTTPDigestAuth(path_info.user, self.password)
-            return HTTPBasicAuth(path_info.user, self.password)
+            elif self.auth == "custom" and self.custom_header:
+                self.headers.update({self.custom_header: self.password})
         return None
 
     @wrap_prop(threading.Lock())
@@ -160,7 +164,11 @@ class RemoteHTTP(RemoteBASE):
 
         try:
             res = self._session.request(
-                method, url, auth=self.auth_method(), **kwargs,
+                method,
+                url,
+                auth=self.auth_method(),
+                headers=self.headers,
+                **kwargs,
             )
 
             redirect_no_location = (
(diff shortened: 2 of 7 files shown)


pmrowla:
I have been testing locally against a simple flask app with `/public` `/basic` `/digest` endpoints for each possible auth method: https://gist.github.com/pmrowla/6ed79cc0c4d20e26ab8e86d25570c983

dvc/config.py
```diff
@@ -101,6 +101,13 @@ class RelPath(str):
     "shared": All(Lower, Choices("group")),
     Optional("slow_link_warning", default=True): Bool,
 }
+HTTP_COMMON = {
+    "basic_auth": Bool,
+    "digest_auth": Bool,
```
efiop: @shcheklein Noticed that we might want to generalise this to support commonly-used custom headers. So how about we adjust this to:
```
"auth": All(Lower, Choices("basic", "digest", "custom")),   # with potential to expand it to oauth, kerberos, etc
"custom_header": str,
"user": str,
"password": str,
"ask_password": Bool,
}
```
and then for `custom` set something like `r.headers[self.custom_header] = self.password`. What do you think @pmrowla ?

pmrowla: Sounds like a reasonable solution to me

pmrowla: One question, should we just separate the ability to provide custom headers from the auth entirely?

i.e. only allow `Choices("basic", "digest")` for `auth`, and have a separate `custom_headers` config option where users can supply a list of `"Header: Value"` strings that will be added to any requests for that remote?

shcheklein: I think I'm fine with either since both options solve the problem. In general it's a choice of "being flexible but and bit lower level" vs "less flexible and more explicit".

A few things to consider:

- it means `auth` type will be become optional, right?
- also how would the `dvc remote modify` command look like? can we come with an example to see if it's not too bloated and more or less easy to write?
- is it fine with us that the whole option will have to go into the local config vs an ability to split header name and password into two configs and manage separately (I can't come up with a case where someone would need this to be honest)

what do you think @pmrowla ?

pmrowla: My initial line of thinking was that in .dvc/config it would look something like
```
['remote "http-custom"']
    url = http://localhost:5000/
    custom_header = My-Header1: value1
    custom_header = My-Header2: value2
    ...
```
And then in the http remote, we would parse all entries for `custom_header`. 

But thinking about it some more, I don't think the benefits of supporting this would be worth the added complexity of making the CLI tools handle options that can be specified multiple times (especially once you get into how you would unset specific values from the CLI).

I'll just go ahead and proceed with what @efiop originally suggested

shcheklein: @pmrowla sounds good! It might make sense to rename the `custom_header` into `custom_auth_header`? to make it more explicit again, and keep us an option of adding that list you mentioned.

dvc/remote/http.py
```diff
@@ -74,6 +119,24 @@ def get_file_checksum(self, path_info):
...
+            if self.ask_password and self.password is None:
+                host, user = path_info.host, path_info.user
+                self.password = ask_password(host, user)
+            if self.auth == "basic":
+                return HTTPBasicAuth(path_info.user, self.password)
+            elif self.auth == "digest":
+                return HTTPDigestAuth(path_info.user, self.password)
+            elif self.auth == "custom" and self.custom_auth_header:
+                self.headers.update({self.custom_auth_header: self.password})
+        return None
```
efiop: Really minor:
```
if not self.auth:
    return None

...
```

dvc/remote/http.py
```diff
@@ -74,6 +119,24 @@ def get_file_checksum(self, path_info):
...
+        if path_info is None:
+            path_info = self.path_info
+
+        if self.auth:
+            if self.ask_password and self.password is None:
+                host, user = path_info.host, path_info.user
+                self.password = ask_password(host, user)
+            if self.auth == "basic":
+                return HTTPBasicAuth(path_info.user, self.password)
+            elif self.auth == "digest":
```
efiop:
```suggestion
            if self.auth == "digest":
```

dvc/remote/http.py
```diff
@@ -74,6 +119,24 @@ def get_file_checksum(self, path_info):
...
+                return HTTPDigestAuth(path_info.user, self.password)
+            elif self.auth == "custom" and self.custom_auth_header:
```
efiop:
```suggestion
            if self.auth == "custom" and self.custom_auth_header:
```

tests/utils/httpd.py
```diff
@@ -35,6 +36,35 @@ class ContentMD5Handler(TestRequestHandler):
...
+        try:
+            os.makedirs(os.path.dirname(path), exist_ok=True)
+            with open(path, "wb") as fd:
+                if chunked:
+                    for chunk in self._chunks():
+                        fd.write(chunk)
+                else:
+                    size = int(self.headers.get("Content-Length", 0))
+                    fd.write(self.rfile.read(size))
+        except (IOError, OSError) as e:
```
skshetry: *Nitpicky*:
We don't need to support Python2.7, so, I don't think we need both here as `IOError` is an alias for `OSError` on Python3.
```suggestion
        except OSError as e:
```

dvc/remote/http.py
```diff
@@ -48,6 +71,28 @@ def _download(self, from_info, to_file, name=None, no_progress_bar=False):
...
+                with open(from_file, "rb") as fd:
+                    while True:
+                        chunk = fd.read(self.CHUNK_SIZE)
+                        if not chunk:
+                            break
+                        pbar.update(len(chunk))
+                        yield chunk
+
+            response = self._request("POST", to_info.url, data=chunks())
+            if response.status_code != 200:
```
skshetry: Response can be `201 Created` as well.

skshetry:
```suggestion
            if response.status_code in (200, 201):
```

efiop:
For the record: test failures are unrelated and will be fixed in https://github.com/iterative/dvc/pull/3368

tests/func/test_data_cloud.py
```diff
@@ -290,6 +291,20 @@ def _get_cloud_class(self):
         return RemoteHDFS
 
 
+@pytest.mark.usefixtures("http_server")
+class TestRemoteHTTP(HTTP, TestDataCloudBase):
```
skshetry: It'd be so easy to also test with `TestDataCloudCLIBase` if we had splitted into smaller tests (`gc` will fail atm).

tests/func/test_data_cloud.py
```diff
@@ -290,6 +291,20 @@ def _get_cloud_class(self):
...
+    @pytest.fixture(autouse=True)
+    def setup_method_fixture(self, request, http_server):
+        self.http_server = http_server
+        self.method_name = request.function.__name__
+
+    def get_url(self):
```
pared: I think DS has a valid point here, the original method is static, shouldn't we make all of them non-static now, that it is required by this particular change?

tests/func/test_data_cloud.py
efiop: @skshetry CLI tests are redundant here, we need to get rid of them at some point. Not worth creating it for http. `gc` fails because there is no `all` method implemented in http, it wasn't part of this ticket, unfortunately :slightly_frowning_face:

tests/func/test_data_cloud.py
pmrowla: For what it's worth, I wrote it like this because it was done the same way in `TestRemoteSSHMocked` (which overrides static `SSHMocked.get_url()` from `tests/remotes.py`).

pared: Well, it was only a question, its not severe, because any potential problems should be detected on the build stage. For me, it can stay as is.

dvc/remote/http.py
```diff
@@ -100,7 +163,13 @@ def _request(self, method, url, **kwargs):
         kwargs.setdefault("timeout", self.REQUEST_TIMEOUT)
 
         try:
-            res = self._session.request(method, url, **kwargs)
+            res = self._session.request(
```
skshetry: @pmrowla, I'm unable to make it work with Digest auth (I'm using the script that you provided on the gist).

skshetry: But, when I change this to
```suggestion
res = requests.Session().request(
```
it works. Something's wrong in our cached session perhaps?

skshetry: I think I figured it out. When I try to push, we send a `HEAD` request, after which the script sets a cookie. And, then all `POST` requests fails. So, I first tried clearing cookies with `self._session.cookies.clear()` and it worked.

skshetry: Probably, we need to set auth when on creating sessions?

pmrowla: I'm wondering if this is a side effect of how my test app in the gist was configured?

By default, flask uses client-side cookies for session data, so flask-httpauth does as well. But for a web server running digest auth to be properly secured, it needs to be handled server-side: https://flask-httpauth.readthedocs.io/en/latest/#security-concerns-with-digest-authentication

If I modify the test app to actually use server-side sessions (https://gist.github.com/pmrowla/0615f162d1308cab4f429b6efafe276a) the existing remote code works without needing to clear any cached cookies

If I set `session.auth` in the http remote at the time we first create the session instead of setting it per request, I still see the same issue making requests against the original test app. So I'm not sure if it's just that `requests.auth.HTTPDigestAuth` won't work properly when talking to an improperly configured flask apps?

skshetry: > If I set session.auth in the http remote at the time we first create the session instead of setting it per request, I still see the same issue making requests against the original test app.

@pmrowla, I tried that too but, does not work. It's unclear what's the best thing to do here. 

If it's probably only `flask-httpauth`, let's ignore for now then.

PR #3345, opened 2020-02-17
dont disable existing loggers when dvc is set up

### LINKED ISSUE #3344: dvc disables logging of other packages
DVC version is 0.80.0 and I am running Linux. The package was installed from AUR

To reproduce the issue run the following few lines of code
```
import logging

LOGGER = logging.getLogger(__name__)

print(LOGGER.disabled)

import dvc
print(LOGGER.disabled)
```

which will output
```
False
True
```

The reason for this is the usage of `logging.config.dictConfig` [here](https://github.com/iterative/dvc/blob/19edc1ab1d2020711c43f8bb9ad58930c74802f6/dvc/logger.py#L147) without adding `"disable_existing_loggers": False` to the dict

ttekampe:

very small change that fixes #3344

### DIFF
--- a/dvc/logger.py
+++ b/dvc/logger.py
@@ -200,6 +200,6 @@ def setup(level=logging.INFO):
                     ],
                 },
             },
-            "disable_existing_loggers": False
+            "disable_existing_loggers": False,
         }
     )
--- a/dvc/logger.py
+++ b/dvc/logger.py
@@ -141,6 +141,13 @@ def _stack_trace(exc_info):
     )
 
 
+def disable_other_loggers():
+    root = logging.root
+    for (logger_name, logger) in root.manager.loggerDict.items():
+        if logger_name != "dvc" and not logger_name.startswith("dvc."):
+            logger.disabled = True
+
+
 def setup(level=logging.INFO):
     colorama.init()
 
@@ -177,22 +184,6 @@ def setup(level=logging.INFO):
             "loggers": {
                 "dvc": {
                     "level": level,
-                    "handlers": [
-                        "console_info",
-                        "console_debug",
-                        "console_errors",
-                    ],
-                },
-                "paramiko": {
-                    "level": logging.CRITICAL,
-                    "handlers": [
-                        "console_info",
-                        "console_debug",
-                        "console_errors",
-                    ],
-                },
-                "flufl.lock": {
-                    "level": logging.CRITICAL,
                     "handlers": [
                         "console_info",
                         "console_debug",
--- a/dvc/logger.py
+++ b/dvc/logger.py
@@ -200,5 +200,6 @@ def setup(level=logging.INFO):
                     ],
                 },
             },
+            "disable_existing_loggers": False
         }
     )
--- a/tests/dir_helpers.py
+++ b/tests/dir_helpers.py
@@ -45,12 +45,12 @@ from global repo template to creating everything inplace, which:
 
 import os
 import pathlib
-import logging
 from contextlib import contextmanager
 
 import pytest
 from funcy import lmap, retry
 
+from dvc.logger import disable_other_loggers
 from dvc.utils.fs import makedirs
 from dvc.compat import fspath, fspath_py35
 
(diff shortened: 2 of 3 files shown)


efiop:
Hi @ttekampe !

Thanks for the PR! Looks like 1 test on windows is failing, due to some `caplog` misuse on our side. We'll take a look soon.

ttekampe:
Hi,

thanks a lot for the prompt response and also for maintaining this nice package!

dvc/logger.py
```diff
@@ -200,5 +200,6 @@ def setup(level=logging.INFO):
                     ],
                 },
             },
+            "disable_existing_loggers": False,
```
skshetry: How noisy is `gitpython` (or, similar dependencies)  on logging, @efiop? We might need to explicitly disable those for good, otherwise, it might mess with cli commands(?).

We need to reset loglevel of `moto` to `logging.CRITICAL`: https://github.com/spulec/moto/issues/2535, or just disable other loggers during test.

efiop: @skshetry They are indeed fairly noisy. Maybe let's `disable_existing_loggers` in CLI and not touch anything in API? We do a similar thingy with some warnings in `dvc/__init__.py`, but that is rather incorrect. We should disable logs and warnings in `dvc/main.py::main()` most likely. What do you think?

skshetry: Yeah, handling in `main.py` is better. The question is, should we be explicitly disable other loggers there or disable every loggers except of DVC in the `main.py`?

efiop: @skshetry I would just disable everything except DVC, to avoid chasing every weird dependency logger out there.

skshetry: Just a note regarding adjustment of tests: If we disable all the loggers in `main.py`, tests where we do not work with `main()` such as `Repo.push`, etc will have all other loggers enabled. So, we might need to disable some loggers on the tests as well (such as `moto` which is making a test fail at the moment).

JIoJIaJIu:
Thank you for your issue and PR, @ttekampe ! :raised_hands: 
I've applied the changes which were requested by @efiop and @skshetry, let wait  what they say about it

dvc/logger.py
```diff
@@ -141,6 +141,26 @@ def _stack_trace(exc_info):
     )
 
 
+def disable_other_loggers():
+    allowed_loggers = [
+        "dvc",
+        "paramiko",
+        "flufl.lock",
```
efiop: Why are we allowing these though? Paramiko is very noisy.

JIoJIaJIu: I took the list from declared [loggers](https://github.com/iterative/dvc/blob/a9bc65ee1f0446de766db59ad1b149de064c5360/dvc/logger.py#L177)
I got that pointing specific log level for `paramiko` and `flufl.lock` can be done for hiding a lot of messages from them, but wasn't sure
Anyway the output of them will be handled by specific local loggers as was previous

efiop: But if we disable them here, then we won't need to tweak them in logger.py, right? If so, let's just disable them here, as it was the original intention.

tests/dir_helpers.py
```diff
@@ -66,9 +66,17 @@
...
-git_logger.setLevel(logging.CRITICAL)
+disabled_loggers = [
+    # see https://github.com/iterative/dvc/issues/3167
+    "git",
+    # see https://github.com/iterative/dvc/pull/3345/files#r380276443
+    # https://github.com/spulec/moto/issues/2535
+    "moto",
+]
+
+for logger_name in disabled_loggers:
```
efiop: Btw, could we call `disable_other_loggers` here too? We are not using any 3rd party loggers in our tests anyway, so seems like we could disable them here as well.

tests/dir_helpers.py
```diff
@@ -67,8 +67,7 @@
 
 
 # see https://github.com/iterative/dvc/issues/3167
-git_logger = logging.getLogger("git")
-git_logger.setLevel(logging.CRITICAL)
+disable_other_loggers()
```
skshetry: It'd be great if we could move this to conftests. Not super important at the moment, but it does not belong here.

PR #3351, opened 2020-02-17
list: colorize output

### LINKED ISSUE #3347: colorize output for command `dvc list`
Needs to colorize a `dvc list` output when `isatty`  based on [**LS_COLORS** env](https://github.com/coreutils/coreutils/blob/61c49d369369773b45beeac659fa32b6f481e8be/src/ls.c#L2679)
Provide fallback if **LS_COLORS** isn't set

Relates https://github.com/iterative/dvc/issues/2509

JIoJIaJIu:
Close #3347


setup.py
```diff
@@ -49,6 +49,7 @@ def run(self):
     "colorama>=0.3.9",
     "configobj>=5.0.6",
     "gitpython>3",
+    "dircolors>=0.0.4",
```
efiop: Could you please check if there is a conda package for this?

dvc/utils/ls_colors.py
```diff
@@ -0,0 +1,32 @@
...
+        if not self.loaded:
+            return text
+
+        if fs_node.is_out:
+            return self._format_code(text, "dvc")
+
+        try:
+            statbuf = os.stat(fspath(fs_node.path_info))
+        except OSError as e:
+            return "%s [Error stat-ing: %s]" % (fs_node.path_info, e.strerror)
```
efiop: We don't use `%`-formatting, please use `.format` instead, as in the rest of the code base.

dvc/utils/ls_colors.py
```diff
@@ -0,0 +1,32 @@
+import os
```
efiop: Feels like this code won't ever be used by anyone else. So maybe let's create `dvc/repo/ls` directory and place this and colorize in there instead to keep it organized?

dvc/repo/ls.py
```diff
@@ -44,34 +72,39 @@ def _ls_files_repo(target_path_info, recursive=None):
...
             if not recursive:
-                files.extend(map(lambda d: PathInfo(dirpath, d), dirnames))
+                files.extend(
+                    map(lambda d: FSNode(PathInfo(dirpath, d)), dirnames)
+                )
                 break
     except NotADirectoryError:
         if os.path.isfile(fspath(target_path_info)):
-            return [target_path_info]
+            return [FSNode(target_path_info)]
```
efiop: Maybe let's wrap `files` in one place instead of wrapping every line in `FSNode`?

dvc/command/ls.py
```diff
@@ -1,28 +1,45 @@
...
                 self.args.target,
                 rev=self.args.rev,
                 recursive=self.args.recursive,
                 outs_only=self.args.outs_only,
+                with_color=with_color,
             )
-            if nodes:
-                logger.info("\n".join(nodes))
+            if fs_nodes:
+                log("\n".join(fs_nodes))
```
efiop: Not sure I understand why we choose `log` instead of continuing dumping into `logger.info` as we do in all other places?

dvc/utils/ls_colors.py
jorgeorpinel: Maybe something more descriptive (in plain language) like
```suggestion
            return "{} - Error getting file status for {}".format(fs_node.path_info, e.strerror)
```

dvc/command/ls.py
JIoJIaJIu: I need to check - the related logger's handler is `tty`, so handlers are depends on logger and log level, so in `self.get_log_fn` I got that info as
```
(with_color, log) = self.get_log_fn(logging.INFO)
```
 I can make it as
```
def hasatty_handler(logger, loglevel):
   ...

with_color = hasatty_handler(logger, logging.INFO)
...
logger.info(...)
```
but it splits two dependent sections and introduce place for an error

JIoJIaJIu: Now I've got possible problem if we have different handlers for the same log level: for example file and stdout. In that case the output will contain color codes
What do you think, @efiop, maybe better to support `color` with additional option (`-C/--color`)?

efiop: @JIoJIaJIu Ah, got it. Good point! If you look into other occurrences of `isatty` in our code, you'll see that we usually do some pretty bald assumptions and would generally just check sys.stdout.isatty() here and would just run with it. Even though, strictly speaking, that is not 100% correct, but for the sake of consistency let's do the same here. We have had lots of discussions about stopping abusing `logger` for things that `print` should handle, but it is better to do those in one sitting later.

dvc/repo/ls.py
```diff
@@ -21,20 +30,39 @@ def ls(url, target=None, rev=None, recursive=None, outs_only=False):
...
+        @wraps(f)
+        def wrapper(fs_node):
+            if not ls_colors:
+                return f(fs_node)
+            return ls_colors.format(fs_node, f(fs_node))
+
+        return wrapper
+
+    return decorator
+
```
efiop: It is quite weird that we have color knowledge in the API part. How about we return list of structures (FSNode i guess) that know if they are outs of some dvc file, or if they are dirs and then will colorize it as we wish in CLI part (dvc/command/ls.py)?

dvc/utils/ls_colors.py
efiop: Better: see https://github.com/iterative/dvc/pull/3351/files#r380398252

dvc/utils/ls_colors.py
efiop: Also not sure why we return such thing at all. You don't expect any errors here, right? So you should probably just raise.

dvc/utils/ls_colors.py
```diff
@@ -0,0 +1,32 @@
...
+    def format(self, fs_node, text=None):
+        text = text or fspath(fs_node.path_info)
+        if not self.loaded:
+            return text
+
+        if fs_node.is_out:
+            return self._format_code(text, "dvc")
+
+        try:
+            statbuf = os.stat(fspath(fs_node.path_info))
```
efiop: You could simply check that while doing `list` and include stat or stat-like structure in FSNode. That way your `is_out` attribute would be generalized with the rest of the stuff as well :slightly_smiling_face: Though I might be missing something here.

efiop:
@JIoJIaJIu Btw, we've been thinking about `--show-json` for `dvc ls` to make it useful in automated scenarios. How about we make `repo.ls` return a dict/list of dicts that would contain an info dict for each path? That info dict would tell if that path is an output, directory, etc. That info then could be processed and colorized properly in `dvc/command/ls.py`. And we'll get a cheap `dvc ls ... --show-json` in the future :slightly_smiling_face:

dvc/repo/ls.py
```diff
@@ -21,20 +30,39 @@ def ls(url, target=None, rev=None, recursive=None, outs_only=False):
...
 
-        def prettify(path_info):
-            if path_info == target_path_info:
-                return path_info.name
-            return relpath(path_info, target_path_info)
+        @colorize(with_color)
+        def prettify(fs_node):
+            if fs_node.path_info == target_path_info:
+                return fs_node.path_info.name
+            return relpath(fs_node.path_info, target_path_info)
```
Suor: I would say move the whole prettify thing not only colorization to command. Make `Repo.ls` return some structured info. I.e.:
```python
[{'path': path_info, 'is_out': True, 'is_dir': False}, ...]
```

dvc/command/ls.py
```diff
@@ -1,28 +1,45 @@
...
+            with_color = (
+                isinstance(handler, LoggerHandler) and handler.stream.isatty()
+            )
+
+        return (with_color, partial(logger.log, log_level))
+
     def run(self):
         from dvc.repo import Repo
 
+        (with_color, log) = self.get_log_fn()
```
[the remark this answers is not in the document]
efiop: @Suor No one asked for it, so I don't think we should add it here. We have some TTY var for other places already.

setup.py
JIoJIaJIu: Thanks for pointing me out, @efiop. I haven't found such package in conda and going to deploy it there. Meanwhile you can take a look on my source code changes

dvc/command/ls/ls_colors.py
```diff
@@ -0,0 +1,28 @@
...
+
+class LsColors(dircolors.Dircolors):
+    def __init__(self, load=True):
+        super().__init__(load)
+
+        if "dvc" not in self._codes:
+            self._codes["dvc"] = "01;31"  # Red
+
+        if ".dvc" not in self._extensions:
+            self._extensions[".dvc"] = "01;33"  # Yellow
```
efiop: These might overlap with the theme and so they will lose their value. Maybe let's color them as regular files/dirs? For marking dvc outputs we might figure some other way later.

JIoJIaJIu: I've disabled coloring `.dvc` files and dvc-outputs by default

setup.py
JIoJIaJIu: I decided not to use library

dvc/repo/ls.py
```diff
@@ -1,34 +1,63 @@
...
+        recursive (bool, optional)
+        outs_only (bool, optional): Shows only DVC-artifacts
+
+    Returns:
+        list of `fs_node`
+
+    Notes:
+        `fs_node` is a dictionary with structure
+        {
+            "path_info": PathInfo,
```
efiop: Maybe let's just return string `path` instead? Users of API shouldn't really care about our PathInfos, esp since PathInfos are not even pathlib's Path strictly speaking :slightly_smiling_face:

dvc/repo/ls.py
```diff
@@ -82,3 +113,22 @@ def _get_target_path_info(repo, target=None):
     if not target:
         return PathInfo(repo.root_dir)
     return PathInfo(repo.root_dir, target)
+
+
+def get_fs_node(path_info, isout=False):
```
efiop:
```suggestion
def _get_fs_node(path_info, isout=False):
```
since it is really a private method.

dvc/repo/ls.py
```diff
@@ -82,3 +113,22 @@ def _get_target_path_info(repo, target=None):
...
+    from dvc.compat import fspath
+
+    try:
+        isdir = os.path.isdir(fspath(path_info))
```
efiop: So for now outputs that are directories will be displayed as regular files(not dirs), right? Since they don't exist in the filesystem. Just clarifying.

I guess we could tell if an output is a directory by looking at it's hash, which will end with `.dir` suffix. So we could use `isdir = out.is_dir_checksum() if out.checksum else False` for outs.

dvc/repo/ls.py
```diff
@@ -1,34 +1,63 @@
...
-            result.extend(_ls_outs_repo(repo, target_path_info, recursive))
+            fs_nodes.extend(_ls_outs_repo(repo, target_path_info, recursive))
 
         if not outs_only:
-            result.extend(_ls_files_repo(target_path_info, recursive))
+            fs_nodes.extend(_ls_files_repo(target_path_info, recursive))
 
-        if target and not result:
+        if target and not fs_nodes:
             raise PathMissingError(target, repo, output_only=outs_only)
```
Suor: If path exists but is an empty dir then we shouldn't throw.

dvc/command/ls/__init__.py
```diff
@@ -14,15 +17,21 @@ def run(self):
...
                 recursive=self.args.recursive,
                 outs_only=self.args.outs_only,
             )
-            if nodes:
-                logger.info("\n".join(nodes))
+            if fs_nodes:
+
+                @colorize(with_color=sys.stdout.isatty())
+                def prettify(fs_node):
+                    return str(fs_node["path_info"])
```
Suor: Single use decorator on a throw away function looks weird. Why not simply combine the two into single function?

dvc/repo/ls.py
```diff
@@ -65,15 +94,17 @@ def _ls_outs_repo(repo, target_path_info, recursive=None):
...
         relpath = path_info.relpath(target_path_info)
         if relpath.parts:
-            return PathInfo(target_path_info, relpath.parts[0])
-        return path_info
+            isout = len(relpath.parts) == 1
+            return (PathInfo(target_path_info, relpath.parts[0]), isout)
+        return (path_info, True)
 
-    return list({get_top_part(out.path_info) for out in outs})
+    path_infos = {get_top_part(out.path_info) for out in outs}
```
Suor: `path_infos` and `get_top_part()` names are now confusing. Not expecting to deal with pairs.

dvc/repo/ls.py
```diff
@@ -65,15 +94,17 @@ def _ls_outs_repo(repo, target_path_info, recursive=None):
...
+    return map(lambda args: get_fs_node(*args), path_infos)
```
Suor: `map(lambda ...)` is kind of awkward, also less efficient than a list comprehension.

dvc/repo/ls.py
Suor: Not being able to use `out.isdir()` looks confusing, but this is how it is on non-dvc-checked-out working dir. I wonder if we have some bugs related to this.

tests/func/test_ls.py
```diff
@@ -23,9 +24,11 @@
 
 
 def match_files(files, expected_files):
-    assert set(files) == set(
-        map(lambda args: os.path.join(*args), expected_files)
-    )
+    left = [(str(f["path_info"]), f["isout"]) for f in files]
+    right = [
+        (os.path.join(*args), is_out) for (args, is_out) in expected_files
+    ]
```
Suor: You may use set comprehensions to not make intermediate lists.

dvc/command/ls/ls_colors.py
```diff
@@ -0,0 +1,61 @@
...
+        if fs_node.get("isdir", False):
+            return self._format(text, code="di")
+
+        if fs_node.get("isexec", False):
+            return self._format(text, code="ex")
+
+        _, ext = os.path.splitext(text)
+        if ext:
+            return self._format(text, ext=ext)
+        return text
```
Suor: Do we color outs?

tests/func/test_ls.py
```diff
@@ -57,16 +60,48 @@ def test_ls_repo(tmp_dir, dvc, scm):
...
+
+    os.environ["LS_COLORS"] = "rs=0:di=01;34:*.xml=01;31:*.dvc=01;33:"
+    cli_args = parse_args(["list", fspath(tmp_dir)])
+    cmd = cli_args.func(cli_args)
+    logger = logging.getLogger("dvc.command.ls")
+    m = mocker.patch.object(logger, "info")
+    m_stdout = mocker.patch.object(
+        sys.stdout, "isatty"
+    )  # needs to enable coloring
+    m_stdout.side_effect = lambda: True
```
Suor: We use `caplog` to test output everywhere else. Can you please make it uniform?

tests/unit/command/ls/test_ls_colors.py
```diff
@@ -0,0 +1,128 @@
...
+    ls_colors = LsColors()
+    ls_colors.load(LsColors.default)
+    output = ls_colors.format(
+        {
+            "path_info": PathInfo("file"),
+            "isexec": False,
+            "isdir": False,
+            "isout": True,
+        }
+    )
```
Suor: May use a template function to make tests smaller/more glancable:
```python
assert _colored("file", "d") == "file"
```
The second argument being a string containing any subset of `"edo"`.

Suor:
Many small things there, mostly cosmetics, a couple look like bugs.

dvc/repo/ls.py
JIoJIaJIu: we traverse only `git+dvc` repos, so cause it's not able to commit empty dir - there is no such case
There is one bug that I'm going to create issue for - show content for `out` directory

dvc/repo/ls.py
JIoJIaJIu: `get_top_part` returns basedir from the path like `dir1/dir2/file` -> `dir1`. So there are a lot of duplicates
I'm going to find another names

dvc/command/ls/ls_colors.py
efiop: We don't color them for now to avoid overlapping with themes. We'll find a better way to mark outs later.

JIoJIaJIu: I think @Suor mentions a general opportunity, @efiop 
I removed completely the logic with matching `dvc` outputs, what I need to recover I think (without forcing color)

dvc/command/ls/__init__.py
JIoJIaJIu: I would like to keep it as a decorator
1. output changes really often (and with including in future additional info will change also), splitting it by different functions make it more readable
2. I need to initialize `LsColors` object before invocation and decorator helps me to do it the most simple way

dvc/command/ls/__init__.py
```diff
@@ -14,15 +17,21 @@ def run(self):
...
-            if nodes:
-                logger.info("\n".join(nodes))
+            if fs_paths:
+
+                @colorize(with_color=sys.stdout.isatty())
+                def prettify(fs_path):
+                    return fs_path["path"]
+
+                fs_paths = map(prettify, fs_paths)
+                logger.info("\n".join(fs_paths))
```
efiop: I agree with @Suor , this is a bit too much. You do a map of prettify with decorator. Too complex. You could unwrap it into something much more feasible like:
```
+                                                                                
+def _colorize(entries):                                                         
+    from .ls_colors import LsColors                                             
+                                                                                
+    ls_colors = LsColors()                                                      
+    if ls_colors:                                                               
+        ls_colors.load_from_environ()                                           
+                                                                                
+                                                                                
+    return [ls_colors.format(fs_path, fs_path["path"]) for fs_path in entries]  
...
+                if sys.stdout.isatty():                                         
+                    entries = _colorize(fs_paths)                               
+                else:                                                           
+                    entries = [fs_path["path"] for fs_path in fs_paths]         
+                logger.info("\n".join(entries))                                 
```
Just from the top of my head, it could clearly be improved, but even as is it avoids the previous complexity.

dvc/command/ls/__init__.py
```diff
@@ -14,15 +17,21 @@ def run(self):
         from dvc.repo import Repo
 
         try:
-            nodes = Repo.ls(
+            fs_paths = Repo.ls(
```
efiop: `fs_paths` is a confusing name, even for me. Maybe there is a better one?

dvc/command/ls/__init__.py
efiop: Also
```
+    ls_colors = LsColors()                                                      
+    if ls_colors:                                                               
+        ls_colors.load_from_environ()    
```
doesn't look necessary, it could load automatically instead, no need for an explicit action here, as we don't even use it anywhere else.

JIoJIaJIu: > You could unwrap it into something much more feasible

I wasn't able to see advantages of the approach you suggest, rather then it's more imperative (is it good?), but it has a list of disadvantages
Could you please explain what advantages your approach has?

Let me summarize my motives:
### Lose coupling 
`ls` prints the output. It does based on some logic which can change, so the main idea is organize the code that easier adapts for changes
```python
def dynamic_function_for_map():
    ...
map(dynamic_function_for_map, data)
```
cause `dynamic_function_for_map` exists any changes are easier to apply. And they are not hardcoded in the parent function
Imagine if we decide no to output `fs_path["path"]` but some custom info as `drwxr-xr-x folder`, or add alignment
In your case there are a few places where needed to apply changes, but with design patterns it's much simpler
```python
@colorize(with_color=sys.stdout.isatty())
def prettify(fs_path, colorize):
  return "{} {}".format(fs_path["mask"] colorize(fs_path["path"]))
```

If you don't like decorator here I see the option to make it as
```python
def prettify(fs_path):
  return colorize(fs_path, fs_path["path"])
```
what is really close to what I did, one problem that I need to initialize `ls_colors` before
```python
ls_colors = LsColors() if sys.stdout.isatty() else None
def prettify(fs_path):
  if ls_colors:
      ls_colors.format(...)
  ...
```
the disadvantage here that `ls_colors` value appears in the method where such information is not needed. With decorator all internal stuff is wrapped in the corresponding function

## Single output
It mostly relates to the previous topic. With your approach you generates the output completely in the different logic branch: 1 - from colorize and 2 - from the main function. What introduces more places for errors and increase cost for applying changes

## Readability
I see the function like that
```
data <- give me data
create dyn(*opts) <- create dynamic prettifier based on options
map(dyn, data) <- return prettified data
```
what is structured, all information relates to the topic and internals are not exposed
vs
```
data <- give me data
if some_opt: <- shouldn't be there
  custom_convert(data)
else:
  [... data] 
```
If I refactor the code you suggest I can see the option as
```
data <- give me data
convert(data, *opts) <- return prettified data based on opts
```
but `convert` function such way becomes not so beautiful and flexible as simple function concatenation which are applied on the item from the list

JIoJIaJIu:
```
ls_colors = LsColors() if istty else None
if ls_colors: 
    ls_colors.load_from_environ() 
```
> doesn't look necessary, it could load automatically instead

It is mostly about design and was introduced specially, my thoughts:
1. `LsColors` formats the text based on option
2. `LsColors` does it based on the dictionary it has, which can be setup (or not) different ways. The case that it's setup from `environ` is the general, but `LsColors` logic is not depends on it competelly
3. If you need to initialize it in 1 row it can be done as
    ```python
    LsColors(os.environ.get("LS_COLORS"))
    ```
4. It allows to extend dictionary after without affecting `env` variable

dvc/command/ls/__init__.py
JIoJIaJIu: Agreed! :+1:  
I like `files`, but it doesn't describe the output well, IMHO

### TLDR
That's the reason why I didn't want to return object/dictionary from the method. The dict is specific
```
{
    "path": str,
    "isexec": bool,
    "isout": bool,
    "isdir": bool,
}
```
and uses only in one place. Hence we introduce additional structure that should have name, but it all names don't describe accurate the object.
I came from the main question - what this dict describes? I see the path where some `node`  (where `node` is file, dir or out) exists/can exist related repo with type of the node
Cause I am already using [fs_node](https://github.com/iterative/dvc/blob/bf4fe8aa3996cf4e8b65a0c1ac7bdfac7b0a3d14/dvc/repo/ls.py#L135) in `Repo.ls` with `path_info` (what means we really have such file on the path except out) during conversion to simple `str` - I needed to differentiate them somehow: https://github.com/iterative/dvc/blob/bf4fe8aa3996cf4e8b65a0c1ac7bdfac7b0a3d14/dvc/repo/ls.py#L64
So cause I use such name in `Repo.ls` I decided to keep it here

dvc/command/ls/__init__.py
```diff
@@ -1,28 +1,43 @@
...
 from dvc.command.base import append_doc_link
 from dvc.command.base import CmdBaseNoRepo
+from dvc.command.ls.ls_colors import LsColors
 from dvc.exceptions import DvcException
 
 
 logger = logging.getLogger(__name__)
 
 
+def prettify(entries, with_color=False):
```
efiop:
```suggestion
def _prettify(entries, with_color=False):
```

dvc/command/ls/ls_colors.py
```diff
@@ -0,0 +1,60 @@
...
+        self._codes = {}
+        if lscolors:
+            self.load(lscolors)
+        else:
+            self.load(os.environ.get("LS_COLORS", LsColors.default))
+
+    def clear(self):
+        self._codes.clear()
+        self._extensions.clear()
+
```
efiop: Looks like this is not needed, since LsColors is pretty much single-use object.

tests/func/test_ls.py
```diff
@@ -57,16 +57,44 @@ def test_ls_repo(tmp_dir, dvc, scm):
...
 
 
+def test_ls_repo_with_color(tmp_dir, dvc, scm, mocker, caplog):
+    import logging
+    from dvc.cli import parse_args
+
+    tmp_dir.scm_gen(FS_STRUCTURE, commit="init")
+    tmp_dir.dvc_gen(DVC_STRUCTURE, commit="dvc")
+
+    os.environ["LS_COLORS"] = "rs=0:di=01;34:*.xml=01;31:*.dvc=01;33:"
```
efiop: This is bad, you need to use mocker to temporarily set env. Otherwise you are affecting all the other tests.

tests/unit/command/ls/test_ls_colors.py
```diff
@@ -0,0 +1,60 @@
...
+            "isout": "o" in spec,
+        }
+        return ls_colors.format(fs_path)
+
+    return _colorize
+
+
+def test_ls_colors_out_file():
+    ls_colors = LsColors()
+    ls_colors.load(LsColors.default)
```
efiop: Why not just
```suggestion
    ls_colors = LsColors(LsColors.default)
```
?

tests/unit/command/ls/test_ls_colors.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def test_ls_colors_out_ext():
+    ls_colors = LsColors(LsColors.default + ":*.xml=01;33")
+    assert colorize(ls_colors)("file.xml", "o") == "\x1b[01;33mfile.xml\x1b[0m"
+
+
+def test_ls_colors_file():
+    ls_colors = LsColors()
+    ls_colors.load(LsColors.default)
```
efiop:
```suggestion
    ls_colors = LsColors(LsColors.default)
```

dvc/command/ls/ls_colors.py
```diff
@@ -0,0 +1,54 @@
...
+        self._extensions = {}
+        self._codes = {}
+        if lscolors:
+            self.load(lscolors)
+        else:
+            self.load(os.environ.get("LS_COLORS", LsColors.default))
+
+    def load(self, lscolors):
+        if not lscolors:
+            return
```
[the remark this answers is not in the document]
JIoJIaJIu: > Do you think that converting load -> _load would help us get rid of double-check?

I think it make sense, I wanted to provide API for loading custom dict, but after iterative changes it isn't used anywhere

> instead of return, go for self.load

Agreed, it looks much more nicer! Thank you!

dvc/repo/ls.py
```diff
@@ -82,3 +130,26 @@ def _get_target_path_info(repo, target=None):
     if not target:
         return PathInfo(repo.root_dir)
     return PathInfo(repo.root_dir, target)
+
+
+def _get_fs_node(path_info, out=None):
```
Suor: This one has two completely parallel code paths, may be split into two functions.

dvc/repo/ls.py
```diff
@@ -1,79 +1,127 @@
...
         return []
 
     files = []
-    tree = CleanTree(WorkingTree(target_path_info))
+    tree = CleanTree(WorkingTree(path_info))
     try:
-        for dirpath, dirnames, filenames in tree.walk(target_path_info):
-            files.extend(map(lambda f: PathInfo(dirpath, f), filenames))
+        for dirpath, dirnames, filenames in tree.walk(path_info):
+            files.extend([PathInfo(dirpath, f) for f in filenames])
```
Suor:
```suggestion
            files.extend(PathInfo(dirpath, f) for f in filenames)
```
No need to construct a list, you may pass a generator expression to `.extend()`. Same below.

dvc/repo/ls.py
```diff
@@ -1,79 +1,127 @@
...
     from dvc.path_info import PathInfo
     from dvc.scm.tree import WorkingTree
 
-    if not os.path.exists(fspath(target_path_info)):
+    if not os.path.exists(fspath(path_info)):
         return []
 
     files = []
-    tree = CleanTree(WorkingTree(target_path_info))
+    tree = CleanTree(WorkingTree(path_info))
```
Suor: This will not read `.dvcignore` files from repo root to `path_info`, should simply use `repo.tree` here.

dvc/repo/ls.py
```diff
@@ -82,3 +130,26 @@ def _get_target_path_info(repo, target=None):
     if not target:
         return PathInfo(repo.root_dir)
     return PathInfo(repo.root_dir, target)
+
+
+def _get_fs_node(path_info, out=None):
+    from dvc.compat import fspath
+
+    if out:
+        isdir = out.is_dir_checksum if out.checksum else False
```
Suor: If there is not checksum it might still be dir, I would write a note here about this edge case. Now the code looks quirky for no apparent reason.

tests/unit/command/ls/test_ls_colors.py
```diff
@@ -0,0 +1,58 @@
...
+            "isout": "o" in spec,
+        }
+        return ls_colors.format(fs_path)
+
+    return _colorize
+
+
+def test_ls_colors_out_file():
+    ls_colors = LsColors(LsColors.default)
+    assert colorize(ls_colors)("file", "o") == "file"
```
Suor: May use the same notation in `Repo.ls()`, also may use "/" as sep and then just str replace it, drop all those parentheses. Up to you.

This will also help you test isdir and isexec parts seamlessly.

dvc/repo/ls.py
```diff
@@ -1,79 +1,127 @@
...
+            )
+            return {
+                "path": path,
+                "isout": fs_node.get("isout", False),
+                "isdir": fs_node.get("isdir", False),
+                "isexec": fs_node.get("isexec", False),
+            }
 
+        entries = [get_entry(fs_node) for fs_node in fs_nodes]
+        entries.sort(key=lambda f: f["path"])
```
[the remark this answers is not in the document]
JIoJIaJIu: I use in-place sort so it is fast enough (even faster I assume), also it looks more readable - every line contains one instruction

dvc/repo/ls.py
```diff
@@ -1,79 +1,127 @@
...
+    url, target=None, rev=None, recursive=None, outs_only=False,
+):
+    """Methods for getting files and outputs for the repo.
+
+    Args:
+        url (str): The repo url
+        target (str, optional)
+        rev (str, optional)
+        recursive (bool, optional)
+        outs_only (bool, optional): Shows only DVC-artifacts
```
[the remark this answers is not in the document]
JIoJIaJIu: :+1: 

JFYI I provided the pydoc cause wanted to describe `Returns` section (really specific return dict).`Args` I provided just for consistency with minimal information

dvc/repo/ls.py
JIoJIaJIu: As I got if `out.checksum` is missed than `out.is_dir_checksum` fails
Do you have thoughts how can it be improved?

tests/unit/command/ls/test_ls_colors.py
JIoJIaJIu: > May use the same notation in Repo.ls(), also may use "/" as sep and then just str replace it, drop all those parentheses. Up to you.

I didn't get your point, could you please extend it? Or maybe an example of your idea?

dvc/command/ls/ls_colors.py
```diff
@@ -0,0 +1,51 @@
...
+class LsColors(object):
+    default = "rs=0:di=01;34:ex=01;32"
+
+    def __init__(self, lscolors=None):
+        self._extensions = {}
+        self._codes = {}
+        if lscolors:
+            self._load(lscolors)
+        else:
+            self._load(os.environ.get("LS_COLORS", LsColors.default))
```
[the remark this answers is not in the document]
JIoJIaJIu: I was surprised that `LS_COLORS= ls -al` works as you described. From my point of view the behavior with disabling coloring by pointing empty var is more correct
What do you think  cc @iterative/engineering 

Interesting that `LS_COLORS=1 ls -al` disables coloring and default colors for dir/links/etc are not used

dvc/repo/ls.py
JIoJIaJIu: I didn't get your point, could you elaborate it more, please?
Here `path_info` is `repo.root_dir` + `path`

dvc/command/ls/ls_colors.py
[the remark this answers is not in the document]
skshetry: On one hand, giving different meanings to an environment variable unset or empty is wrong. But, we also want to imitate  `ls`. Not a strong preference, but, I'd go with however `ls` works. :slightly_smiling_face:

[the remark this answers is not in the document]
Suor: In my understanding empty string should work the same as env var unset. If we have internal default than that is used naturally.

dvc/repo/ls.py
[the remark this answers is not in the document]
Suor: You can really write "returns a dict like ...", we are not set for google standard or anything. They it looks with types and optional and no desc is just making question I would say:
- ok, it's a string, but what does it means?
- ok, it's optional what does it defaults to?

dvc/repo/ls.py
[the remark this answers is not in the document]
Suor: I benched it, `sorted()` is faster. The rest is more about taste, I prefer to avoid mutable ops, so I might be biased )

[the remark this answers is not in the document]
Suor: Just resolve it if you wish.

dvc/repo/ls.py
Suor: If you have a structure:

```bash
.dvc/
.dvcignore
data/
    a
    b
    c
```

And `.dvcignore` contains `c`, then your code will list all of the files in `data`, while `c` should be skipped. The reason is that `.dvcignore` file won't be read because it is outside of your scope. If you use `repo.tree` it will work correctly, also no need to use internals like `CleanTree/WorkingTree`.

dvc/repo/ls.py
Suor: Yeah, it fails, we don't really no whether that is a dir. We may just say about this limitation in a comment.

tests/unit/command/ls/test_ls_colors.py
Suor: You can use:

```python
    match_files(
        files,
        [     
            ("some-dir/statistics/data.csv", "o"),
            ("some-dir/subdir", "d"),
            ("some-dir/subdir", "d"),
            ("derived", "do"),
        ]
    )
```

I.e. the same DSL for testing `Repo.ls()`, should make it more readable.

dvc/repo/ls.py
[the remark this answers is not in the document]
JIoJIaJIu: I've done [a small research](https://gist.github.com/JIoJIaJIu/19c954140a4cd9c6908f096c76ca495a) with **10^6** shuffled list
```bash
$ python 1.py 
         1000004 function calls in 0.822 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.822    0.822 1.py:11(comp_and_sort)
        1    0.284    0.284    0.413    0.413 1.py:12(<listcomp>)
  1000000    0.128    0.000    0.128    0.000 1.py:8(get_entry)
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
        1    0.409    0.409    0.409    0.409 {method 'sort' of 'list' objects}
```
```bash
$ python 1.py 2
         1000003 function calls in 0.820 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.820    0.820 1.py:17(map_and_sorted)
  1000000    0.128    0.000    0.128    0.000 1.py:8(get_entry)
        1    0.692    0.692    0.820    0.820 {built-in method builtins.sorted}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
```

They are the same (if we construct the list), what is logical cause in my approach I create list first(`<listcomp>`) and then sort it(`sort`), and in yours it's done in one place (`builtins.sorted`) 

About immutability - it makes sense and I think it would be better to use `sorted` then, thanks :+1:

dvc/repo/ls.py
JIoJIaJIu: > we don't really know whether that is a dir

In what case the out can be dir but doesn't have `checksum`?

dvc/repo/ls.py
JIoJIaJIu: For me it's not clear from requirements that `c` should be skipped. Could you please raise a bug, cause it goes out the scope of the PR? @Suor

dvc/command/ls/ls_colors.py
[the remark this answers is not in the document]
JIoJIaJIu: Ok, I'm going to imitate `ls` behavior as @Suor suggested

tests/unit/command/ls/test_ls_colors.py
JIoJIaJIu: Yes, it looks really nice, but let me keep as it is at cause:
1. Only one method in `tests/func/test_ls` really tests coloring, all others don't output colors (`isatty = False`)
2. The improvement requires a lot of efforts and doesn't bring real value at this point
2. I think it has sense when we add more types (`ln`, `broken ln`) and decide to cover coloring in `func/test_ls` also

dvc/repo/ls.py
JIoJIaJIu: I would like to keep as it is cause don't want produce additional functions with the same meaning. Also it will bloat the logic - needs dynamically choose right function, now it works inside one function

tests/unit/command/ls/test_ls_colors.py
Suor: Up to you, this has nothing to do with coloring BTW.

dvc/repo/ls.py
Suor: Yes.

dvc/repo/ls.py
Suor: If you have `.dvcignore` file inside path info you are listing then it will be read and applied if `.dvcignore` file is in dvc repo somewhere closer to the root than `path_info` then it is not read. This is inconsistent, a bug in this particular PR, not out of scope.

dvc/repo/ls.py
Suor: Ok let's leave it as is.

dvc/repo/ls.py
Suor: You can put different way, `CleanTree` should be instantiated from repo root, doing it for any subdir is an error, which causes a bug above.

Suor: Other thing is that `CleanTree` as a very internal one, now only `Repo` knows about it, any code outside that should not know about that to limit complexity, that was one of the ideas behind `CleanTree`. We use it in a couple of unit tests, but that's it.

JIoJIaJIu: > This is inconsistent, a bug in this particular PR, not out of scope.

Current PR relates colorizing output, I've create the ticket https://github.com/iterative/dvc/issues/3431 with the behavior you described

dvc/repo/ls.py
JIoJIaJIu: @Suor it was an open question - could you please provide the case when it can happen?

dvc/command/ls/__init__.py
```diff
@@ -1,28 +1,43 @@
...
+        ls_colors = LsColors()
+        fmt = ls_colors.format
+    else:
+
+        def fmt(entry):
+            return entry["path"]
+
+    return [fmt(entry) for entry in entries]
+
+
```
skshetry:
```suggestion
def _prettify(entries, with_color=False):
    def fmt(entry):
        return entry["path"]
    if with_color:
        ls_colors = LsColors()
        fmt = ls_colors.format
    return [fmt(entry) for entry in entries]
```

Or, maybe, even use a lambda?

tests/unit/command/ls/test_ls_colors.py
```diff
@@ -0,0 +1,58 @@
...
+    def _colorize(f, spec=""):
+        fs_path = {
+            "path": f,
+            "isexec": "e" in spec,
+            "isdir": "d" in spec,
+            "isout": "o" in spec,
+        }
+        return ls_colors.format(fs_path)
+
+    return _colorize
```
skshetry: It'd be one less redirection if we just did following on all of the tests:
```py
ls_colors.format({"path": f, "isexec": True})
```

JIoJIaJIu: Thank you @skshetry for the point!
The changes came from the comment https://github.com/iterative/dvc/pull/3351#discussion_r382584512 in terms of making tests more readable

skshetry: Are other keys mandatory? If they aren't, I don't really think there's any need for redirection.

If they are, you can just:
```py
shape = {
  "isexec": False,
  "isdir": False,
  "isout": False,
}
ls_colors.format({"path": f, **shape, "isexec": True})
```

But, I'm fine with it as well. But, one less redirection would be great.

dvc/repo/ls.py
Suor: When a user edits a `.dvc` file and deletes the output checksum there. This is not as weird as it might sound, people may want to do that to recalc the checksum from the actual dir they have.

dvc/repo/ls.py
Suor: Looked at the current repo code and you are right, this bug was there before. Ok solve it separately.

efiop:
@JIoJIaJIu @skshetry So what about the changes/questions above?

tests/unit/command/ls/test_ls_colors.py
JIoJIaJIu: > Are other keys mandatory?

Nope, there are not. In fact before [the comment](https://github.com/iterative/dvc/pull/3351#discussion_r382584512) my tests [were](https://github.com/iterative/dvc/commit/d4b0194c3b3d0caaa4236e1431a15f39a82e83b8#diff-02572f52f28eecd20a866fbcd2629201R24) as you described

I don't see the problem with additional step here (redirection you mean). I think it makes sense not to add additional functionality related to tests only, but let it keep here cause it's thin enough

dvc/command/ls/__init__.py
JIoJIaJIu: I've remembered why I did it as it is
with lambda there is an [E371](https://www.flake8rules.com/rules/E731) style error
```python
def _prettify(entries, with_color=False):
    if with_color:
        ls_colors = LsColors()
        fmt = ls_colors.format
    else:
        fmt = lambda entry: entry["path"]

    return [fmt(entry) for entry in entries]
```
```
dvc/command/ls/__init__.py:17:5: E731 do not assign a lambda expression, use a def
```
Unfortunately with you suggestion I am getting [F811](https://www.flake8rules.com/rules/F811.html)
```python
def _prettify(entries, with_color=False):
     def fmt(entry):
        return entry["path"]

    if with_color:
        ls_colors = LsColors()
        fmt = ls_colors.format
    return [fmt(entry) for entry in entries]
```
```
dvc/command/ls/__init__.py:20:9: F811 redefinition of unused 'fmt' from line 15
```

JIoJIaJIu:
> @JIoJIaJIu @skshetry So what about the changes/questions above?

@efiop I've provided my thoughts, I think we can keep the PR in the last shape if you don't mind

dvc/command/ls/__init__.py
skshetry: @JIoJIaJIu, It's fine to ignore redefinition warnings (it's what I suggested). But, I'm fine as-is, it's only a suggestion.

skshetry:
@efiop, I have already approved. It was just a few suggestions. I'm fine as-is.

dvc/repo/ls.py
JIoJIaJIu: Ok, I can imagine that user don't see folder when has been working on dvc file and doesn't understand what's going on, but it's really specific case. I can handle it in the https://github.com/iterative/dvc/issues/3394. What do you think @efiop?

Issue #3176, opened 2020-01-17
docstring format for api, possibly across code base?


ghost:
> Extracted from https://github.com/iterative/dvc/pull/3130#pullrequestreview-343835585

We currently don't have a _desired style_ to write docstrings.

**Should we come up with one**? If so, **which one**?

bhavaniravi:
Most of the standards suggest `"""` for docstrings. I think we should go with the same. 

https://stackoverflow.com/questions/3898572/what-is-the-standard-python-docstring-format

jorgeorpinel:
Thanks @bhavaniravi! I think thus much we can all agree on 🙂  The larger question is what text format to use INSIDE docstrings. Ref: https://stackoverflow.com/questions/3898572/what-is-the-standard-python-docstring-format

Suor:
I don't think we should spend time formalizing this, we have very limited amount of user facing functions really.

shcheklein:
@Suor I think you brought this topic, right? :) Are we good to close it (discussion around formatting) then and let Jorge decide this, for example?

jorgeorpinel:
I agree it's only important at the moment to have good full code doc for the API functions and I'm happy to just pick one format but would prefer to get some feedback from the core engineering team. In fact I wanted to summarize the options here (extracted from https://github.com/iterative/dvc/pull/3130#discussion_r368192562 +):

> is there a standard way of doing markup that is understood by the tools that generate a view for this (like F1 in PyCharm, quick help in VS Code, docs generators, etc) ?
 
Among doc generators, [Sphinx](http://www.sphinx-doc.org/en/master/) is by far the most popular, with [pdoc](https://pdoc3.github.io/pdoc) a far 2nd place, it seems. Sphinx uses [reStructuredText](https://docutils.sourceforge.io/docs/user/rst/quickref.html) format. **pydoc** is interesting because it supports md and [other interesting formats](https://pdoc3.github.io/pdoc/doc/pdoc/#supported-docstring-formats) worth checking out actually (from Numpy and Google). All of these formats use back quotes/ticks for interpreted text and other uses except Google's.

> A good general guide about docstring options is in https://docs.python-guide.org/writing/documentation/

[...]

PR #3352, opened 2020-02-18
api: docstring updates

jorgeorpinel:
Matches iterative/dvc.org/pull/908

### DIFF
--- a/dvc/api.py
+++ b/dvc/api.py
@@ -84,6 +84,5 @@ def _make_repo(repo_url=None, rev=None):
 
 
 def _require_dvc(repo):
-    """Raises UrlNotDvcRepoError if repo is not a DVC repository."""
     if not isinstance(repo, Repo):
         raise UrlNotDvcRepoError(repo.url)
--- a/dvc/api.py
+++ b/dvc/api.py
@@ -18,10 +18,10 @@ class UrlNotDvcRepoError(DvcException):
 
 
 def get_url(path, repo=None, rev=None, remote=None):
-    """
-    Returns the full URL to the data artifact specified by its `path` in a
-    `repo`.
-    NOTE: There is no guarantee that the file actually exists in that location.
+    """Returns URL to the storage location of a data artifact tracked
+    by DVC, specified by its path in a repo.
+
+    NOTE: There's no guarantee that the file actually exists in that location.
     """
     with _make_repo(repo, rev=rev) as _repo:
         _require_dvc(_repo)
@@ -31,7 +31,7 @@ def get_url(path, repo=None, rev=None, remote=None):
 
 
 def open(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
-    """Context manager to open a file artifact as a file object."""
+    """Context manager to open a tracked file as a file object."""
     args = (path,)
     kwargs = {
         "repo": repo,
@@ -63,7 +63,7 @@ def _open(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
 
 
 def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
-    """Returns the contents of a file artifact."""
+    """Returns the contents of a tracked file."""
     with open(
         path, repo=repo, rev=rev, remote=remote, mode=mode, encoding=encoding
     ) as fd:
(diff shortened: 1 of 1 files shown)


dvc/api.py
```diff
@@ -84,5 +84,6 @@ def _make_repo(repo_url=None, rev=None):
 
 
 def _require_dvc(repo):
+    """Raises UrlNotDvcRepoError if repo is not a DVC repository."""
```
efiop: Do we really need this? It is a private helper consisting of 2 lines :)

jorgeorpinel: Maybe not but per some recent chat with @shcheklein I thought we want to always mention exceptions in docstrings for user-facing modules at least (its possible an IDE might show this docstring to a user). I can easily remove it if I'm confused.

jorgeorpinel: Or mention it in the docstring of the corresponding public functions that use this private one?

efiop: @jorgeorpinel So there is a proper format for docstrings to mark that something raises something. The info about UrlNotDvcRepoError should be in public methods. E.g.
```
   Raises:
        UrlNotDvcRepoError: If repo is not a DVC repository.
```
I don't think any IDE will be able to figure out what some method rises based on private methods docstrings that it uses :slightly_smiling_face:

efiop: I would remove this one for now.

shcheklein: yes, the idea was too mention in the public method description exceptions it raises that important to know about, especially if they could be caught in some regular cases to do a fallback or something. Definitely agree with Ruslan no use to put docstrings into private helpers. At least not for the end users. Sometimes it might be helpful for developers.

efiop:
```suggestion
```

jorgeorpinel: Thanks for finishing this PR for me @efiop!

> there is a proper format for docstrings to mark that something raises something... E.g. in google docstrings

Confused 😕 I thought we decided NOT to use any fancy formats? See https://github.com/iterative/dvc/issues/3176#issuecomment-579872748

> yes, the idea was too mention in the public method description

OK @shcheklein so should we add this one to `get_url`? I can send a follow-up PR for that. Just not sure what (lack of) docstring format to use...
 https://github.com/iterative/dvc/blob/f673144b19c596cd66caa8f144cb511c8f3f4257/dvc/api.py#L20-L27

shcheklein: @jorgeorpinel could we check how stdlib deals with it in case they mention important exception?

jorgeorpinel: Checking `.../3.7/lib/python3.7/os.py` (`os` module), I see that it just casually mentions exceptions are raised (not even with exact types) in docstrings:

```py
def makedirs(name, mode=0o777, exist_ok=False):
    """makedirs(name [, mode=0o777][, exist_ok=False])

    ... If the target directory already
    exists, raise an OSError if exist_ok is False. Otherwise no exception is
    raised.
```
```py
def walk(top, topdown=True, onerror=None, followlinks=False):
    """Directory tree generator.

    ... It can
    report the error to continue with the walk, or raise the exception
    to abort the walk.
```
```py
def _fspath(path):
    """Return the path representation of a path-like object.

    ... If the
    path representation is not str or bytes, TypeError is raised. If the
    provided path is not str, bytes, or os.PathLike, TypeError is raised.
```

Note that not all raised exceptions raised in public functions are documented in docstrings, for example `get_exec_path()` can raise a `ValueError` (directly) but this is not mentioned in its docstring (maybe because it's a pretty generic/obvious exception). `popen()` can also raise `TypeError` and `ValueError` exceptions (not mentioned in docstring), same in `fdopen()` – probably because they're near the top of each fn so it's obvious/self-documenting.

shcheklein: so, sound like a pretty reasonable approach in this case, especially considering that we are taking stdlib as an inspiration for the format in general. What do you think?

efiop: > Note that not all raised exceptions raised in public functions are documented in docstrings (and _fspath is not even public – seems arbitrary)

This doesn't apply to the case we've described here, as that docstring was pointless, it was commenting on 2 lines of obvious code with an obvious comment :) 

> Confused confused I thought we decided NOT to use any fancy formats? See #3176 (comment)

Sorry, that discussion went by me. Left a comment there. But my comment in this PR was simply stating that obvious docstrings in private methods are not that useful as is and are not even useful in auto-parsed scenarios.

PR #3360, opened 2020-02-18
Fix bash completion

### LINKED ISSUE #1923: Autocomplete for file names doesn't work (partially) [bash]
Hello!
When I use autocompletion script from here `https://github.com/iterative/dvc.org/blob/master/static/docs/user-guide/autocomplete.md` it doesn't complete file and directory names from current directory for the command `dvc repro ...`.
But if I add some flag, for example `dvc repro --force ...` then it works fine.
What should be changed in bash autocomplete script to allow autocompletion for `dvc repro ...` without any flags?

casperdcl:
- [x] fix bash completion of files in cwd
  + [x] filter non-DVC-files where appropriate
- [x] fix potential name clashed
- [x] ~~replace `$(compgen -f)` with `$(compgen -f -X \!*?.dvc)`?~~
  + this would mean commands which expect any arbitrary file paths will NOT accept DVC-file paths
- fixes #1923

Caveats:

- still won't allow completion of subdir paths. That would require a major enough overhaul that it's probably for another PR.


casperdcl:
btw just a quick check - is there any instance where a command should autocomplete with BOTH DVC-files as well as non-DVC-files?

efiop:
> btw just a quick check - is there any instance where a command should autocomplete with BOTH DVC-files as well as non-DVC-files?

@casperdcl `pull/push/fetch/status/checkout` :slightly_smiling_face: They all support specifying particular output instead of dvc-file.

casperdcl:
Great this is mergeable then.

PR #3363, opened 2020-02-19
gc: make it safer by only activating by use of certain flags

### LINKED ISSUE #2325: Reconsider gc implementation

skshetry:
Previous plain `gc` is now not supported. The same behavior can be trigger by
`gc -w` or `gc --workspace`. When `--all-{tags, commits, branches}` are
specified, it will work as if `-w` is also specified (same as before).

https://github.com/iterative/dvc.org/issues/1013

Related to #2325

### DIFF
--- a/dvc/command/gc.py
+++ b/dvc/command/gc.py
@@ -12,6 +12,16 @@ logger = logging.getLogger(__name__)
 
 class CmdGC(CmdBase):
     def run(self):
+        from dvc.repo.gc import _raise_error_if_all_disabled
+
+        _raise_error_if_all_disabled(
+            all_branches=self.args.all_branches,
+            all_tags=self.args.all_tags,
+            all_commits=self.args.all_commits,
+            workspace=self.args.workspace,
+            cloud=self.args.cloud,
+        )
+
         msg = "This will remove all cache except items used in "
 
         msg += "the working tree"
@@ -47,6 +57,7 @@ class CmdGC(CmdBase):
             force=self.args.force,
             jobs=self.args.jobs,
             repos=self.args.repos,
+            workspace=self.args.workspace,
         )
         return 0
 
@@ -63,6 +74,13 @@ def add_parser(subparsers, parent_parser):
         description=append_doc_link(GC_DESCRIPTION, "gc"),
         help=GC_HELP,
         formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+    gc_parser.add_argument(
+        "-w",
+        "--workspace",
+        action="store_true",
+        default=False,
+        help="Keep data files used in the current workspace.",
     )
     gc_parser.add_argument(
         "-a",
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -7,6 +7,10 @@ class DvcException(Exception):
     """Base class for all dvc exceptions."""
 
 
+class InvalidArgumentError(ValueError, DvcException):
+    """Thrown if arguments are invalid."""
+
+
 class OutputDuplicationError(DvcException):
     """Thrown if a file/directory is specified as an output in more than one
     stage.
--- a/dvc/repo/gc.py
+++ b/dvc/repo/gc.py
@@ -2,6 +2,7 @@ import logging
 
 from . import locked
 from dvc.cache import NamedCache
+from dvc.exceptions import InvalidArgumentError
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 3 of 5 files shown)


dvc/repo/gc.py
```diff
@@ -13,6 +13,15 @@ def _do_gc(typ, func, clist):
         logger.info("No unused '{}' cache to remove.".format(typ))
 
 
+def _raise_error_if_all_disabled(**kwargs):
+    if not any(kwargs.values()):
+        raise ValueError(
+            (
+                "Invalid Arguments. " "Either of ({}) " "needs to be enabled."
```
Suor: What's with this separate string literals? This also needs to be `DvcException` to not be displayed as an unexpected error.

dvc/command/gc.py
```diff
@@ -47,6 +56,7 @@ def run(self):
             force=self.args.force,
             jobs=self.args.jobs,
             repos=self.args.repos,
+            workspace=self.args.workspace,
```
Suor: A general thought, not about this PR, - we should stop automatically making command line options into boolean kwargs, e.g. using `set()` would be more appropriate here.

dvc/repo/gc.py
```diff
@@ -25,7 +34,19 @@ def gc(
     force=False,
     jobs=None,
     repos=None,
+    workspace=False,
 ):
+
+    # require `workspace` to be true to come into effect.
+    # assume `workspace` to be enabled if any of `all_tags`, `all_commits`,
+    # or `all_branches` are enabled.
+    _raise_error_if_all_disabled(
```
skshetry: I think, we should make `-w` mandatory (as it'll be symmetrical with `-c`).

tests/func/test_gc.py
```diff
@@ -218,5 +219,47 @@ def test_gc_no_unpacked_dir(tmp_dir, dvc):
...
+    # assume `-w` enabled if `-a`/`-T`/`--all-commits` is specified
+    assert not any(
+        [
+            main(["gc", "-vf", "-w"]),
+            main(["gc", "-vf", "-w"]),
+            main(["gc", "-vf", "-w", "-a"]),
+            main(["gc", "-vf", "-T"]),
+            main(["gc", "-vf", "-a"]),
+            main(["gc", "-vf", "--all-commits"]),
+            main(["gc", "-vf", "-a", "-T"]),
```
Suor: A fail message here will probably be not helpful. Maybe use `for` loop instead.

tests/func/test_gc.py
```diff
@@ -218,5 +219,47 @@ def test_gc_no_unpacked_dir(tmp_dir, dvc):
...
+
+def test_gc_without_workspace(tmp_dir, dvc, caplog):
+    with caplog.at_level(logging.WARNING, logger="dvc"):
+        assert main(["gc", "-vf"]) == 255
+
+    assert "Invalid Arguments" in caplog.text
+
+
+def test_gc_with_possible_args_positive(tmp_dir, dvc):
+    # assume `-w` enabled if `-a`/`-T`/`--all-commits` is specified
```
Suor: How this comment relates to the test?

dvc/repo/gc.py
skshetry: dammit black. I'll fix it.

dvc/repo/gc.py
skshetry: Looking into it again, I feel it's dangerous to do just `-c` to also delete from the current workspace. I'll propose changing it, so that `-c` will only delete from the remote and `-w` only from the local cache. And, combined both, they can have the previous behavior.

Thoughts?

Suor: I would say we keep the current behavior in this PR. How `-c` should behave should be discussed separately.

skshetry: Kept the same behavior for now.

dvc/exceptions.py
```diff
@@ -7,6 +7,10 @@ class DvcException(Exception):
     """Base class for all dvc exceptions."""
 
 
+class InvalidArgumentError(ValueError, DvcException):
+    """Thrown if arguments are invalid."""
```
efiop: Nice one! Should we use it in otherplaces like `dvc/repo/__init__.py` and `dvc/repo/reproduce.py` too?

skshetry: I'd opt for it on a different PR. :slightly_smiling_face:

efiop: @skshetry Sure, let's not forget about it though :slightly_smiling_face:

efiop:
@skshetry Let's not forget about docs, please :)

skshetry:
@efiop, I have created an issue in https://github.com/iterative/dvc.org/issues/1013

PR #3365, opened 2020-02-19
Stage: fix used_cache warning on no checksum

pared:

### DIFF
--- a/dvc/output/base.py
+++ b/dvc/output/base.py
@@ -409,12 +409,23 @@ class OutputBase(object):
             cache.external[dep.repo_pair].add(dep.def_path)
             return cache
 
-        if not self.info:
-            logger.warning(
-                "Output '{}'({}) is missing version info. Cache for it will "
-                "not be collected. Use `dvc repro` to get your pipeline up to "
-                "date.".format(self, self.stage)
+        if not self.checksum:
+            msg = (
+                "Output '{}'({}) is missing version info. "
+                "Cache for it will not be collected. "
+                "Use `dvc repro` to get your pipeline up to date.".format(
+                    self, self.stage
+                )
             )
+            if self.exists:
+                msg += (
+                    "\n"
+                    "You can also use `dvc commit {stage}` to associate "
+                    "existing '{out}' with '{stage}'.".format(
+                        out=self, stage=self.stage.relpath
+                    )
+                )
+            logger.warning(msg)
             return NamedCache()
 
         ret = NamedCache.make(self.scheme, self.checksum, str(self))
--- a/tests/func/test_stage.py
+++ b/tests/func/test_stage.py
@@ -1,6 +1,7 @@
 import os
 import tempfile
 
+
 from dvc.main import main
 from dvc.output.local import OutputLOCAL
 from dvc.remote.local import RemoteLOCAL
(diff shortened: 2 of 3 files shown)


dvc/output/base.py
```diff
@@ -149,7 +149,11 @@ def cache_path(self):
 
     @property
     def checksum(self):
-        return self.info.get(self.remote.PARAM_CHECKSUM)
+        # TODO it will always call get, need to write it somewhere
+        checksum = self.info.get(self.remote.PARAM_CHECKSUM)
+        if not checksum:
+            checksum = self.remote.get_checksum(self.path_info)
```
efiop: This property should only return what is already computed, it shouldn't start computing it all over again.

See dvc/outpt/base.py:get_used_cache, where the issue rose. We check `if not self.info` there, while we should probably check `if not self.checksum`. Push shouldn't push uncommited changes, as it doesn't even have the cache for them in .dvc/cache.

dvc/output/base.py
```diff
@@ -409,12 +409,23 @@ def get_used_cache(self, **kwargs):
...
+                "Cache for it will not be collected. "
+                "Use `dvc repro` to get your pipeline up to date.".format(
+                    self, self.stage
+                )
             )
+            if self.exists:
+                msg += (
+                    "\n"
+                    "You can also use `dvc commit {stage}` to associate "
+                    "existing '{out}' with `{stage}`.".format(
```
efiop:
```suggestion
                    "existing '{out}' with '{stage}'.".format(
```

dvc/output/base.py
```diff
@@ -409,12 +409,23 @@ def get_used_cache(self, **kwargs):
...
+                        out=self, stage=self.stage.relpath
```
efiop:
```suggestion
                        out=self, stage=self.stage
```
same as above :slightly_smiling_face:

tests/func/test_stage.py
```diff
@@ -178,3 +181,29 @@ def test_meta_is_preserved(tmp_dir, dvc):
...
+    with caplog.at_level(logging.WARNING, logger="dvc"):
+        dvc.push()
+
+    assert first(caplog.messages) == (
+        "Output 'file'(Stage: 'file.dvc') is missing version info. "
+        "Cache for it will not be collected. "
+        "Use `dvc repro` to get your pipeline up to date.\n"
+        "You can also use `dvc commit file.dvc` to associate existing 'file' "
+        "with `file.dvc`."
+    )
```
efiop: Maybe let's narrow it down to get_used_cache unit test for output?

dvc/output/base.py
pared: AFAIK it will result in `dvc commit 'Stage file.dvc'` ;)

tests/func/test_stage.py
pared: makes sense

dvc/output/base.py
efiop: @pared Oh, but then the upper warning is broken too, right?

pared: I wouldn't say so: the upper message is pointing user to which stage particular output is associated to, while the lower one suggests command that can fix your situation. So in the upper one, its not crucial for `stage` to be printed just as relpath.

Issue #2985, opened 2019-12-19
tab completion scripts need more love


jorgeorpinel:
Seems to me it's about time we review and update tab completion scripts again (like in #1732)? There are some new and removed commands, options, etc, I think. See also #2983. Some other examples:

- `gc --all-commits` has been hidden from help output (#2681).
- `dvc run --remove-outs` is now deprecated.

And maybe we could add a checkbox to .github/PULL_REQUEST_TEMPLATE.md similar to the one for updating docs but for tab completion scripts?

mroutis:
Indeed, @jorgeorpinel

shcheklein:
+1 on adding a checkbox. Updating these scripts should be responsibility of one who is creating a PR.

It reminds me updating requirements though, and checking licenses, etc. We should consider creating a document for the core team - like extended instruction "before-merge-sequence".

efiop:
@shcheklein Licences and stuff like this is tiresome and will only make the contribution process harder, plus no one really reads that stuff anyway once it is past 2-3 checkboxes. Licences and completion can be automated by proper scripts, so I would rather take a look at that instead.

shcheklein:
@efiop yes, I didn't mean to put them as checkboxes. I suggest we keep them to 3 as we have now. More like an internal pre-merge check-list. If it can be made a part of the automation - that asks you certain questions before you able to merge - that would be really nice.

efiop:
@shcheklein Developers don't read those boxes either 🙂 For Licences and completion it would probably be easier to automate the checks or even generation (for bash and zsh). Need to take a look at some point.

shcheklein:
@efiop it's not for developers either - it's mostly for maintainers. Before you push the merge button make sure to check that list.

efiop:
@shcheklein I meant for core developers/maintainers. No one likes checklists.

shcheklein:
@efiop totally, that's why this "unlove" will be a constant reminder to automate, right? :) the worst scenario here is to have nothing and fix things afterwards. Also that check list is a requirement list for the automation anyway.

jorgeorpinel:
I like checklists. But anyway, since we want to keep this one 3 boxes long, should we at least add a note about trying to keep tab completion up-to-date in the core contrib guide?

efiop:
@shcheklein @jorgeorpinel Sure, let it be at least checkboxes for now.

benjaminal:
Claro como funciona

mroutis:
[...]

PR #3374, opened 2020-02-20
completion: bash: properly recurse targets

### LINKED ISSUE #1923: Autocomplete for file names doesn't work (partially) [bash]

### LINKED ISSUE #3361: completion: bash: handle subdirs
- fully fixes #1923
- continues #3360

casperdcl:
- [x] properly recurse commands & subcommands path completion
  + [x] only autocomplete DVC-files where appropriate
- [x] add inline documentation for maintenance
- fixes #3361
- fully fixes #1923
- continues #3360
- related #2985

## Notes

- `-o nospace` is required (as documented in the code) to avoid an unreasonable rewrite of everything.
- the `snap` package automatically benefits from this PR :)
- I'm tempted to write a test for future use, and perhaps even think of a way to automate based on `argparse` but that way likely lieth a massive independent project moste arcane.
  + have manually tested all current permutations though

I haven't seen any code examples take this approach and don't know why. It looks elegant, maintainable, and efficient to me.

### DIFF
--- a/scripts/completion/dvc.bash
+++ b/scripts/completion/dvc.bash
@@ -16,14 +16,14 @@ _dvc_add_COMPGEN=_dvc_compgen_files
 _dvc_cache='dir'
 _dvc_cache_dir=' --global --system --local -u --unset'
 _dvc_checkout='-d --with-deps -R --recursive -f --force --relink'
-_dvc_checkout_COMPGEN=_dvc_compgen_DVCfiles
+_dvc_checkout_COMPGEN=_dvc_compgen_DVCFiles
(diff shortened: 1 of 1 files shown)


casperdcl:
I haven't seen any code examples take this approach and don't know why. It looks elegant, maintainable, and efficient to me.

PR #3376, opened 2020-02-21
[WIP] gdrive: a pair of cleanup refactors

Suor:


Suor:
@efiop do we skip GDrive tests on travis?

skshetry:
@Suor, it only runs on `master` AFAIK.

efiop:
@skshetry is correct, but we also run on upstream branches too. So you could create a PR from an upstream branch instead.

dvc/remote/gdrive.py
```diff
@@ -41,30 +42,18 @@ def __init__(self, path):
...
-        retry_codes = ["403", "500", "502", "503", "504"]
-        if any(
-            "HttpError {}".format(code) in str(exception)
-            for code in retry_codes
-        ):
-            raise GDriveRetriableError("Google API request failed")
-        raise
-    return result
-
+    retry_re = re.compile(r"HttpError (403|500|502|503|504)")
```
shcheklein: great!, but we'll need to rebuilt it on the PyDrive2 side to get rid of this fragile stuff

dvc/remote/gdrive.py
```diff
@@ -41,30 +42,18 @@ def __init__(self, path):
...
 
-gdrive_retry = compose(
     # 15 tries, start at 0.5s, multiply by golden ratio, cap at 20s
-    retry(
-        15, GDriveRetriableError, timeout=lambda a: min(0.5 * 1.618 ** a, 20)
-    ),
-    _wrap_pydrive_retriable,
-)
+    return retry(
```
shcheklein: is there a simple way to write a debug message on a retry, btw?

dvc/remote/gdrive.py
```diff
@@ -362,11 +319,7 @@ def _get_remote_drive_id(self, remote_id):
     def _get_cached_remote_ids(self, path):
         if not path:
             return [self._bucket]
-        if self._cache_initialized:
```
shcheklein: this logic is important, we don't want to initialize cache if it's not initialized yet

shcheklein: cache property init makes a call to self._get_remote_id itself to get root id. I'm not sure who would it behave with these changes

dvc/remote/gdrive.py
```diff
@@ -126,120 +115,90 @@ def __init__(self, repo, config):
...
-        self._remote_root_id = self._get_remote_id(self.path_info)
-        for dir1 in self.gdrive_list_item(
-            "'{}' in parents and trashed=false".format(self._remote_root_id)
+    @wrap_prop(threading.RLock())
+    @cached_property
+    def cache(self):
+        cache = {"dirs": defaultdict(list), "ids": {}}
+
+        cache["root_id"] = self._get_remote_id(self.path_info)
+        self._cache_path(self.path_info.path, cache["root_id"], cache)
```
shcheklein: I believe we did't cache root id before in the cache itself, it might affect list all below - before it was returning only actual files in cache, now also directories under the root?

dvc/remote/gdrive.py
```diff
@@ -126,120 +115,90 @@ def __init__(self, repo, config):
...
+                self._gdrive_user_credentials_path, "w"
+            ) as credentials_file:
+                credentials_file.write(
+                    os.getenv(RemoteGDrive.GDRIVE_USER_CREDENTIALS_DATA)
+                )
 
-        if not self._gdrive:
-            from pydrive2.auth import GoogleAuth
-            from pydrive2.drive import GoogleDrive
+        GoogleAuth.DEFAULT_SETTINGS["client_config_backend"] = "settings"
```
shcheklein: not directly related - would really love this to go away from the class ... is there a good way? or at least assign the whole dict at once to DEFAULT_SETTINGS? This way it is done now is quite ugly and bloated

dvc/remote/gdrive.py
Suor: You can wrap a func, which you retry into [`@log_errors()`](https://funcy.readthedocs.io/en/stable/debug.html#log_errors):
```python
# ...
return retry(...)(log_errors(logger.error, stack=False)(func))
```
This will show errors, but there is no non-hacky way to show retries.

dvc/remote/gdrive.py
Suor: Yeah, it is. One can use `GoogleAuth.DEFAULT_SETTINGS.update()`, but the right way would be to make `GoogleAuth` accept settings as a dict not only file.

dvc/remote/gdrive.py
Suor: We cached path -> id in a way, made it separately. Plus extra code on cache read side. We did not cache id -> path though. This might affect listing cache, you are right.

Suor:
Closed in favor if #3382

Issue #2756, opened 2019-11-08
remote with relpath breaks get


pared:
```
#!/bin/bash

rm -rf repo remote_repo storage git_repo
mkdir repo remote_repo storage git_repo

pushd remote_repo
git init >> /dev/null && dvc init -q

set -x
set -e
echo data >> data
# dvc remote add -d str ../storage
dvc remote add -d str /full/path/to/storage
dvc add data
git add .dvc/config data.dvc .gitignore
git commit -m "add data"
dvc push
rm data
rm -rf .dvc/cache

popd
pushd repo

git init >> /dev/null && dvc init -q
dvc get ../remote_repo data
```
Will work just fine, but when we will use relpath to point to storage (as in commented line) 
we will get warnings that cache files do not exist and `data` will not be created.

It is happening, because, in `external_repo` we clone url to `/tmp` and therefore, relative path to cache cannot be resolved properly.

pared:
The solution could be to make `ExternalRepo` acknowledge local remotes with a relative path and make it resolve the paths appropriately.

tizoc:
Hello, I will take a look into this /cc @efiop

efiop:
@tizoc Sounds great! Please let us know if you'll have any questions :)

tizoc:
Will do!

So far what I have found out is that (unlike git) dvc is making non-absolute paths in the config [relative to the `.dvc` directory](https://github.com/iterative/dvc/blob/master/dvc/config.py#L307-L314), not the root of the repository:

```
$ dvc remote add -d str ../storage
Setting 'str' as a default remote.
$ cat .dvc/config
[core]
    remote = str
['remote "str"']
    url = ../../storage
```

git:

```
$ git remote add str ../storage
$ cat .git/config 
[remote "str"]
        url = ../storage
        fetch = +refs/heads/*:refs/remotes/str/*
```

If I manually change the path to be relative to the root of the repository the `dvc get ../remote_repo data` command works.

So, before going forward, my question is: Is this (paths relative to the `.dvc` directory instead of the root) the expected behavior?

efiop:
@tizoc We actually transform it relative to the config file location, which by default is `.dvc/config`. Git does indeed treat it literally, i don't think it will work after that though, unless you are in repo root.

[...]

PR #3378, opened 2020-02-21
erepo: fix relative url on default remote (#2756)

tizoc:

### Pending

- [x] figure out how to write a functional test for this

### DIFF
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -126,35 +126,30 @@ class ExternalRepo(Repo):
     def _fix_upstream(self):
         if os.path.isdir(self.url):
             upstream_name = self.config["core"].get("remote")
-            if upstream_name:
-                self._fix_local_remote(upstream_name)
-            else:
-                self._add_upstream()
+            src_repo = Repo(self.url)
+            try:
+                if upstream_name:
+                    self._fix_local_remote(src_repo, upstream_name)
+                else:
+                    self._add_upstream(src_repo)
+            finally:
+                src_repo.close()
 
-    def _fix_local_remote(self, remote_name):
+    def _fix_local_remote(self, src_repo, remote_name):
         # Keep the relative path for a local remote relative
         # to the original repo.
-        src_repo = Repo(self.url)
-        try:
-            new_remote = self.config["remote"][remote_name]
-            old_remote = src_repo.config["remote"][remote_name]
-            # NOTE: it is assumed that only local+relative urls
-            # will be different from their old version. Revise this
-            # test if that stops being true.
-            if new_remote["url"] != old_remote["url"]:
-                new_remote["url"] = old_remote["url"]
-        finally:
-            src_repo.close()
+        new_remote = self.config["remote"][remote_name]
+        old_remote = src_repo.config["remote"][remote_name]
+        # NOTE: it is assumed that only local+relative urls
+        # will be different from their old version. Revise this
+        # test if that stops being true.
+        if new_remote["url"] != old_remote["url"]:
+            new_remote["url"] = old_remote["url"]
 
-    def _add_upstream(self):
+    def _add_upstream(self, src_repo):
         # Fill the empty upstream entry with a new remote pointing to the
         # original repo's cache location.
-        src_repo = Repo(self.url)
-        try:
-            cache_dir = src_repo.cache.local.cache_dir
-        finally:
-            src_repo.close()
-
+        cache_dir = src_repo.cache.local.cache_dir
         self.config["remote"]["auto-generated-upstream"] = {"url": cache_dir}
         self.config["core"]["remote"] = "auto-generated-upstream"
 
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -124,20 +124,20 @@ class ExternalRepo(Repo):
         self.cache.local.cache_dir = cache_dir
 
     def _fix_upstream(self):
-        if os.path.isdir(self.url):
-            upstream_name = self.config["core"].get("remote")
-            src_repo = Repo(self.url)
-            try:
-                if upstream_name:
-                    self._fix_local_remote(src_repo, upstream_name)
-                else:
-                    self._add_upstream(src_repo)
-            finally:
-                src_repo.close()
+        if not os.path.isdir(self.url):
+            return
+
+        remote_name = self.config["core"].get("remote")
+        src_repo = Repo(self.url)
+        try:
+            if remote_name:
+                self._fix_local_remote(src_repo, remote_name)
+            else:
+                self._add_upstream(src_repo)
+        finally:
+            src_repo.close()
 
     def _fix_local_remote(self, src_repo, remote_name):
-        # Keep the relative path for a local remote relative
-        # to the original repo.
         new_remote = self.config["remote"][remote_name]
         old_remote = src_repo.config["remote"][remote_name]
         if new_remote["url"] != old_remote["url"]:
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -140,9 +140,6 @@ class ExternalRepo(Repo):
         # to the original repo.
         new_remote = self.config["remote"][remote_name]
         old_remote = src_repo.config["remote"][remote_name]
-        # NOTE: it is assumed that only local+relative urls
-        # will be different from their old version. Revise this
-        # test if that stops being true.
         if new_remote["url"] != old_remote["url"]:
             new_remote["url"] = old_remote["url"]
 
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -138,6 +138,9 @@ class ExternalRepo(Repo):
             src_repo.close()
 
     def _fix_local_remote(self, src_repo, remote_name):
+        # If a remote URL is relative to the source repo,
+        # it will have changed upon config load and made
+        # relative to this new repo. Restore the old one here.
         new_remote = self.config["remote"][remote_name]
         old_remote = src_repo.config["remote"][remote_name]
         if new_remote["url"] != old_remote["url"]:
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -124,11 +124,23 @@ class ExternalRepo(Repo):
         self.cache.local.cache_dir = cache_dir
 
     def _set_upstream(self):
-        # check if the URL is local and no default remote is present
-        # add default remote pointing to the original repo's cache location
+        # check if the URL is local and:
+        # - if no default remote is present, then add default remote
+        #   pointing to the original repo's cache location
+        # - if a default remote is present, make the URL
+        #   relative to the original repo
         if os.path.isdir(self.url):
-            if not self.config["core"].get("remote"):
-                src_repo = Repo(self.url)
+            src_repo = Repo(self.url)
+            remote_name = self.config["core"].get("remote")
+            if remote_name:
+                remote_cfg = self.config["remote"][remote_name]
+                old_remote_url = src_repo.config["remote"][remote_name]["url"]
+                if remote_cfg["url"] != old_remote_url:
+                    remote_cfg["url"] = os.path.join(
+                        src_repo.root_dir, old_remote_url
+                    )
+                src_repo.close()
+            else:
                 try:
                     cache_dir = src_repo.cache.local.cache_dir
                 finally:
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -66,6 +66,7 @@ class ExternalRepo(Repo):
         super().__init__(root_dir)
         self.url = url
         self._set_cache_dir()
+        self._set_relative_upstream()
         self._set_upstream()
 
     def pull_to(self, path, to_info):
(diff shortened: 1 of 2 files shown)


tizoc:
@pared @efiop added a test, I'm not entirely sure about it.

What I tried to do there is to reproduce the script in the description of the issue, but maybe that is too much and it would be better to just check that the path is absolute and that it exists? not sure. I will wait for your input on that.

tizoc:
Note that this change only modifies the default remote, should all remotes with relative urls be taken care of instead?

dvc/external_repo.py
```diff
@@ -66,6 +66,7 @@ def __init__(self, root_dir, url):
         super().__init__(root_dir)
         self.url = url
         self._set_cache_dir()
+        self._set_relative_upstream()
         self._set_upstream()
```
efiop: Naming is confusing now. The logic you've introduced adjusts the remote if it exists, and `_set_upstream` sets cache dir as a remote if no default remote is set. So they should be in separate branches of execution.

dvc/external_repo.py
```diff
@@ -123,6 +124,22 @@ def _set_cache_dir(self):
...
+            remote_name = self.config["core"].get("remote")
+            if remote_name:
+                src_repo = Repo(self.url)
+                remote_cfg = self.config["remote"][remote_name]
+                old_remote_url = src_repo.config["remote"][remote_name]["url"]
+                if remote_cfg["url"] != old_remote_url:
+                    remote_cfg["url"] = os.path.join(
+                        src_repo.root_dir, old_remote_url
+                    )
+                src_repo.close()
```
efiop: You might be leaking repo resources if any error happens above. Better to do the same thing as in `_set_upstream` down below and put `repo.close()` into `finally` block.

dvc/external_repo.py
```diff
@@ -123,6 +124,22 @@ def _set_cache_dir(self):
...
+    def _set_relative_upstream(self):
+        # check if the URL is local and if a default relative-path
+        # remote is present, if so, make the URL relative to the
+        # original repo
+        if os.path.isdir(self.url):
+            remote_name = self.config["core"].get("remote")
+            if remote_name:
+                src_repo = Repo(self.url)
+                remote_cfg = self.config["remote"][remote_name]
+                old_remote_url = src_repo.config["remote"][remote_name]["url"]
```
efiop: remote url might be a non-local one, in which case you shouldn't modify it.

dvc/external_repo.py
```diff
@@ -123,6 +124,22 @@ def _set_cache_dir(self):
...
+                if remote_cfg["url"] != old_remote_url:
+                    remote_cfg["url"] = os.path.join(
+                        src_repo.root_dir, old_remote_url
```
efiop: `old_remote_url` is an abs path anyway, so there is no need to `os.path.join`.

dvc/external_repo.py
Suor: I would say we can just assign with no check:
```python
self.config["remote"][name]["url"] = src_repo.config["remote"][name]["url"]
```
Since url should be the same anyway. Name method smth like `._fix_local_upstream()`, comment on what's the intention if that is not enough.

tizoc: Should an url for a remote in the erepo ever be different from the url of the matching remote in the original repository?
Because if not, then an equality check should be enough -- assuming we want to avoid unnecessary mutations to the config, otherwise it can be assigned directly as you said.

dvc/external_repo.py
tizoc: Thank you, I didn't notice that relative urls came already prefixed.

dvc/external_repo.py
tizoc: Something like:

```python
new_remote = self.config["remote"][remote_name]
old_remote = src_repo.config["remote"][remote_name]
if new_remote["url"] != old_remote["url"]:
    new_remote["url"] = old_remote["url"]
```

An additional check, in case an url in the new repo could be different from an url in the old repo is to check that `self.root_dir` is a prefix of the new url (which also ensures that it is local and relative).

efiop: Great point @Suor !

> Should an url for a remote in the erepo ever be different from the url of the matching remote in the original repository?

Yes, when `_set_upstream` is applied, because it sets it to the original cache location. But those are in different branches of execution, so it doesn't matter and we can simply assign here.

tizoc:
Thanks for the feedback everyone. Just pushed a new version based on your comments.

Branching has been moved to it's own function (`_fix_upstream`, wraps the other two), which then calls `_add_upstream` or `_fix_local_remote` (not "upstream" because it can be any remote passed as argument) depending on what is needed.

If other local relative remotes happen to need handling, `_fix_local_remote` can be used as-is.

dvc/external_repo.py
```diff
@@ -123,21 +123,40 @@ def _set_cache_dir(self):
...
+            src_repo.close()
+
+    def _add_upstream(self):
+        # Fill the empty upstream entry with a new remote pointing to the
+        # original repo's cache location.
+        src_repo = Repo(self.url)
+        try:
+            cache_dir = src_repo.cache.local.cache_dir
+        finally:
+            src_repo.close()
```
efiop: Let's move
```
src_repo = Repo(self.url)
try:
     ...
finally:
    src_repo.close()
```
to _fix_upstream, it will deduplicate some code.

tizoc: I thought about that, but I was unsure because of how `_set_upstream` still sets an upstream even if obtaining the cache dir fails. But since in that case it will set the url to `None` I guess it doesn't matter?

dvc/external_repo.py
```diff
@@ -123,21 +123,35 @@ def _set_cache_dir(self):
...
+                src_repo.close()
+
+    def _fix_local_remote(self, src_repo, remote_name):
+        # Keep the relative path for a local remote relative
+        # to the original repo.
+        new_remote = self.config["remote"][remote_name]
+        old_remote = src_repo.config["remote"][remote_name]
+        # NOTE: it is assumed that only local+relative urls
+        # will be different from their old version. Revise this
+        # test if that stops being true.
```
pared: I would remove that. There already is an explanation of what this method is doing at the beginning. 
Comment `Revise...` will not make anyone revise that, that is the purpose of test :)

tizoc:
@pared thank you! just made the changes you requested.

dvc/external_repo.py
```diff
@@ -123,21 +123,32 @@ def _set_cache_dir(self):
...
+            try:
+                if upstream_name:
+                    self._fix_local_remote(src_repo, upstream_name)
+                else:
+                    self._add_upstream(src_repo)
+            finally:
+                src_repo.close()
+
+    def _fix_local_remote(self, src_repo, remote_name):
+        # Keep the relative path for a local remote relative
```
JIoJIaJIu: I think the comment is not relevant:
1. Cause the paths are absolute as `Config.RelPath`
2. There are no any conversions of the paths

dvc/external_repo.py
```diff
@@ -123,21 +123,32 @@ def _set_cache_dir(self):
...
-                    src_repo.close()
-
-                self.config["remote"]["auto-generated-upstream"] = {
-                    "url": cache_dir
-                }
-                self.config["core"]["remote"] = "auto-generated-upstream"
+            upstream_name = self.config["core"].get("remote")
+            src_repo = Repo(self.url)
+            try:
+                if upstream_name:
```
JIoJIaJIu:
```python
try:
   self._add_upstream(src_repo)
...
```

and move the logic of resolving upstream into the `_add_upstream` method. Cause here the upstream is added, what type of it and how - can be handled inside the method
If you want to use conditions here then methods can be named differently, for example as
```python
try:
    if remote_name:
      self._add_upstream_from_remote(..) # not sure in relevance of `add_upstream`
   else:
      self._add_upstream_from_repo(...)
```

dvc/external_repo.py
tizoc: Ok, I guess the wording is not correct in that comment. The path is absolute, but it was built from a relative URL, and thats why it is being changed with the path from the source repo.

What do you think of making the test a check to see if the path is relative instead?

```python
if isinstance(new_remote["url"], RelPath):
    ...
```

This makes it explicit that what we care about there is paths with urls relative to the repository, with no more clarification required.

dvc/external_repo.py
```diff
@@ -123,21 +123,32 @@ def _set_cache_dir(self):
 
         self.cache.local.cache_dir = cache_dir
 
-    def _set_upstream(self):
-        # check if the URL is local and no default remote is present
-        # add default remote pointing to the original repo's cache location
+    def _fix_upstream(self):
```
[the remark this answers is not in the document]
tizoc: Thanks, thats better.

dvc/external_repo.py
```diff
@@ -123,21 +123,32 @@ def _set_cache_dir(self):
...
-                try:
-                    cache_dir = src_repo.cache.local.cache_dir
-                finally:
-                    src_repo.close()
-
-                self.config["remote"]["auto-generated-upstream"] = {
-                    "url": cache_dir
-                }
-                self.config["core"]["remote"] = "auto-generated-upstream"
+            upstream_name = self.config["core"].get("remote")
```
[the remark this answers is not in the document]
tizoc: Yes, but this one isn't any remote, right? what is being checked here is the default remote, not the other remotes (if any). Anyway, both names work for me, let me know what you think.

dvc/external_repo.py
tizoc: I'm thinking that maybe the two separate flows are not required? Consider my question here https://github.com/iterative/dvc/pull/3378#issuecomment-589820202

If instead of going one way or another depending on if there is a default remote set, I go through every remote and replace every RelPath with the one from the source repo, and then I check if there is a default repo set up (and handle that accordingly) there will be a single branch of execution.

Something like this:

```python
    def _fix_remotes(self):
        if not os.path.isdir(self.url):
            return

        src_repo = Repo(self.url)
        try:
            self._fix_source_relative_remotes(src_repo)
            if "remote" not in self.config["core"]:
                self._set_upstream(src_repo)
        finally:
            src_repo.close()

    def _fix_source_relative_remotes(self, src_repo):
        for remote_name, new_remote in self.config["remote"].items():
            if isinstance(new_remote["url"], RelPath):
                old_remote = src_repo.config["remote"][remote_name]
                new_remote["url"] = old_remote["url"]
```

/cc @efiop

tizoc: Maybe rename `_set_upstream` to `_set_upstream_to_repo` (also, the comment there is probably not needed anymore, it is describing what the three lines of code in that function already express clearly).

dvc/external_repo.py
efiop: @tizoc maybe `if not os.path.isabs(new_url)` then? It would work with strings and pathlib objects. Just worried about this being easilly broken with some hand-written dict config in tests or elsewhere.

dvc/external_repo.py
efiop: @tizoc I think two branches of execution are necessary, since `_set_upstream` is only needed when there is no default remote and `_fix_upstream` is only needed when there is a default remote but it is a local one with relpath url.

@JIoJIaJIu So you mean moving
```
                if upstream_name:
                    self._fix_local_remote(src_repo, upstream_name)
                else:
                    self._add_upstream(src_repo)
```

into a new method? Well, we could, but I don't see much value in doing that, tbh. I would be fine with the current approach as well.

tizoc: @efiop my reasoning is, if relative urls are imported from the source repo, we want all of them to be kept relative to the source repo, right? not just the one that happens to be the default. If that is the case, that is independent from if there is a default remote set or not.

dvc/external_repo.py
tizoc: Ok, I had not thought about that. But you are right, `if not os.path.isabs(new_url)` would work the same and without those possible breakages. The benefit of checking for `RelPath` instances was only to better express intent.

tizoc:
I just pushed the requested changes.

I have not included the change I proposed in https://github.com/iterative/dvc/pull/3378#discussion_r383879624

If you think doing that would be an improvement I will commit that too (but using `os.path.isabs` instead of checking for `RelPath` as per https://github.com/iterative/dvc/pull/3378#discussion_r384232317 )

dvc/external_repo.py
efiop: @tizoc But we don't really care about other remotes in this case. We only care about default remote, as it is the only thing that we are actually going to use, at least for now.

dvc/external_repo.py
```diff
@@ -123,21 +123,32 @@ def _set_cache_dir(self):
...
+                self._add_upstream(src_repo)
+        finally:
+            src_repo.close()
+
+    def _fix_local_remote(self, src_repo, remote_name):
+        new_remote = self.config["remote"][remote_name]
+        old_remote = src_repo.config["remote"][remote_name]
+        if new_remote["url"] != old_remote["url"]:
+            new_remote["url"] = old_remote["url"]
+
```
efiop: Would at least add a comment about what is going on here, though.

tizoc: Done. Btw, should I squash all commits into a single one?

efiop: As you wish. But we'll squash automatically anyways :slightly_smiling_face: Though if you think this is a final patch, you need to remove WIP prefix from the header.

tizoc: Ah cool! no need then. "WIP" removed.

efiop:
Thanks @tizoc ! :pray:

PR #3380, opened 2020-02-21
docs: update badges

### LINKED ISSUE #3233: readme: update badges
- travis badge seems to be hardcoded as green/ok (vis https://github.com/iterative/dvc.org/issues/918#issuecomment-577345881)
- add other install methods (vis https://github.com/iterative/dvc.org/issues/904#issue-546378276)
- ensure consistent style (flat, square, round)

casperdcl:
- [x] add `|Choco|`
- [x] add `|Brew|`
- [x] fix style for `|Conda|`
- [x] add `|PyPI|`
- [x] add `|Packages|`
- [x] add a Chocolatey section
- [x] reorder install options sections based on
  + number of users
  + number of potential users <-- I'd recommend this
  + number of desired users
- fixes #3233
- related iterative/dvc.org#904

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -8,7 +8,9 @@
 • `Tutorial <https://dvc.org/doc/get-started>`_
 • `Mailing List <https://sweedom.us10.list-manage.com/subscribe/post?u=a08bf93caae4063c4e6a351f6&id=24c0ecc49a>`_
 
-|Release| |CI| |Maintainability| |Coverage| |Donate| |Conda| |Snap| |DOI|
+|Release| |CI| |Maintainability| |Coverage| |Donate| |DOI|
+
+|Snap| |Choco| |Brew| |Conda| |PyPI| |Packages|
 
 |
 
@@ -92,6 +94,8 @@ Full instructions are `available here <https://dvc.org/doc/get-started/install>`
 pip (PyPI)
 ----------
 
+|PyPI|
+
 .. code-block:: bash
 
    pip install dvc
@@ -109,6 +113,8 @@ To install the development version, run:
 
 Homebrew
 --------
+
+|Brew|
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -8,7 +8,9 @@
 • `Tutorial <https://dvc.org/doc/get-started>`_
 • `Mailing List <https://sweedom.us10.list-manage.com/subscribe/post?u=a08bf93caae4063c4e6a351f6&id=24c0ecc49a>`_
 
-|Release| |CI| |Maintainability| |Coverage| |Donate| |Conda| |Snap| |DOI|
+|Release| |CI| |Maintainability| |Coverage| |Donate| |DOI|
+
+|Snap| |Choco| |Brew| |Conda| |PyPI| |Packages|
```
skshetry: Do we really need this number of badges? Maybe, we could just put a same page link to `Installation` section?

efiop: @skshetry Good point. So far the idea was that it is conveinient to see all of those badges, as you are intuitively looking for them whenever you are checking out a new project. Hiding them will make it less visible, so I'm not sure if it is a good thing. It does get a bit crowded, you are right.

casperdcl: Indeed, though if we want to drop things arguably `|Release| |CI| |Maintainability| |Coverage|` are least relevant to end users.

efiop: Ok, looking at readme now, I do see that it is way too crowded with installation badges :(

> Indeed, though if we want to drop things arguably |Release| |CI| |Maintainability| |Coverage| are least relevant to end users.

That is not true, maintainability/coverage/ci all matters, as it is a sign of a well-maintained project. Those should not be removed.

casperdcl: Doesn't look too crowded to me:

image

Showing we support different installation options means more users likely to bother reading on an actually install. Maintainability/coverage/ci (in terms of users, not devs) are either completely ignored or assumed to be perfect - in both cases they don't necessarily need to be at the top of the readme if we want to drop some badges.

skshetry: I feel that 19 links at the top is too many. It makes it hard to scan through when the objective of having top-links and badges is to be quick and easy.

Issue #2445, opened 2019-08-27
ui: dvc startup became slow


Suor:
```bash
$ time dvc --version
0.57.0+e3f6c7

real	0m2.186s
user	0m1.834s
sys	0m0.288s
```

efiop:
Might be imports again. Need to check cprofile report, it will make it clear.

Suor:
So it's imports:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
4903/2532    0.004    0.000    0.471    0.000 <frozen importlib._bootstrap>:1009(_handle_fromlist)
     1802    0.003    0.000    0.004    0.000 <frozen importlib._bootstrap>:103(release)
     1330    0.001    0.000    0.001    0.000 <frozen importlib._bootstrap>:143(__init__)
     1330    0.001    0.000    0.018    0.000 <frozen importlib._bootstrap>:147(__enter__)
     1330    0.001    0.000    0.004    0.000 <frozen importlib._bootstrap>:151(__exit__)
     1802    0.012    0.000    0.015    0.000 <frozen importlib._bootstrap>:157(_get_module_lock)
     1317    0.002    0.000    0.002    0.000 <frozen importlib._bootstrap>:176(cb)
      472    0.001    0.000    0.003    0.000 <frozen importlib._bootstrap>:194(_lock_unlock_module)
   1597/4    0.001    0.000    1.011    0.253 <frozen importlib._bootstrap>:211(_call_with_frames_removed)
    17493    0.004    0.000    0.004    0.000 <frozen importlib._bootstrap>:222(_verbose_message)
       23    0.000    0.000    0.000    0.000 <frozen importlib._bootstrap>:232(_requires_builtin_wrapper)
      169    0.000    0.000    0.000    0.000 <frozen importlib._bootstrap>:269(_module_repr)
     1242    0.001    0.000    0.001    0.000 <frozen importlib._bootstrap>:307(__init__)
     1242    0.001    0.000    0.001    0.000 <frozen importlib._bootstrap>:311(__enter__)
     1242    0.002    0.000    0.005    0.000 <frozen importlib._bootstrap>:318(__exit__)
     4954    0.001    0.000    0.001    0.000 <frozen importlib._bootstrap>:321(<genexpr>)
     1190    0.001    0.000    0.001    0.000 <frozen importlib._bootstrap>:35(_new_module)
[...]

PR #3400, opened 2020-02-25
test: check that no remote backend modules are loaded by default

Suor:
We had this missed out repeatedly over time, so I added a check that we don't import any remote related deps on initial load.

Examples of related issues are #3399 and #2445.

### DIFF
--- a/tests/unit/test_imports.py
+++ b/tests/unit/test_imports.py
@@ -1,0 +1,19 @@
+import subprocess
+
+
+def test_no_remote_imports():
+    remote_modules = {
+        "boto3",
+        "botocore",
+        "google.cloud.storage",
+        "azure.storage.blob",
+        "oss2",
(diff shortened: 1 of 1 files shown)


Suor:
BTW, the test should fail until #3399 lands.

tests/unit/test_imports.py
```diff
@@ -0,0 +1,18 @@
+import subprocess
+
+
+def test_no_remote_imports():
+    remote_modules = {
+        "boto3",
```
skshetry: Can you please also add `botocore` to the list?

shcheklein:
@Suor could you please rebase/merge?

Suor:
Rebased

Suor:
Fixed Python 3.5

Suor:
Should be ready to merge.

PR #3401, opened 2020-02-25
checkout: show summary by default and introduce --show-changes flag

### LINKED ISSUE #2979: checkout/add: change warning messages to regular/info if behavior is expected

skshetry:

https://github.com/iterative/dvc.org/pull/1054

Resolves #2979.

**TODO**
- [x] Add few tests.
- [x] `CheckoutError`: go with the consensus. 
- [x] Make changes to flags if required. (make default?)
- [ ] Handle `--relink` and `modified` case.
- [ ] Show `--relink` successfully message. (Separate PR)
- [ ] Show how many files were downloaded/fetched. (Separate PR)
- [ ] Remove warnings on checkout failure
- [ ] Fix pull/fetch outputs
- [ ] Autocompletion scripts and help messages

### DIFF
--- a/dvc/command/checkout.py
+++ b/dvc/command/checkout.py
@@ -82,10 +82,10 @@ class CmdCheckout(CmdBase):
             exc = _exc
             stats = exc.stats
 
-        if self.args.show_changes:
-            log_changes(stats)
-        else:
+        if self.args.summary:
             log_summary(stats)
+        else:
+            log_changes(stats)
 
         if exc:
             raise exc
@@ -103,10 +103,10 @@ def add_parser(subparsers, parent_parser):
         formatter_class=argparse.RawDescriptionHelpFormatter,
     )
     checkout_parser.add_argument(
-        "--show-changes",
+        "--summary",
         action="store_true",
         default=False,
-        help="Show list of changes",
+        help="Show summary of the changes",
     )
     checkout_parser.add_argument(
         "-d",
--- a/dvc/command/checkout.py
+++ b/dvc/command/checkout.py
@@ -10,8 +10,8 @@ from dvc.exceptions import CheckoutError
 logger = logging.getLogger(__name__)
 
 
-def _human_join(words=None):
-    words = list(words or [])
+def _human_join(words):
+    words = list(words)
     if not words:
         return ""
 
@@ -25,21 +25,13 @@ def _human_join(words=None):
 
 
 def log_summary(stats):
-    message = [
-        ("added", "{added} added"),
-        ("deleted", "{deleted} deleted"),
-        ("modified", "{modified} modified"),
-    ]
-    summary = {
-        stat: len(stats[stat]) for stat, num in message if stats.get(stat)
-    }
-    if not summary:
-        return
-
-    template = _human_join(
-        fragment for stat, fragment in message if stat in summary
+    states = ("added", "deleted", "modified")
+    summary = (
+        "{} {}".format(len(stats[state]), state)
+        for state in states
+        if stats.get(state)
     )
-    logger.info(template.format_map(summary))
+    logger.info(_human_join(summary))
 
 
 def log_changes(stats):
@@ -59,7 +51,7 @@ def log_changes(stats):
             logger.info(
                 "{color}{state}{nc}{spacing}{entry}".format(
                     color=color,
-                    state=state[0].capitalize(),
+                    state=state[0].upper(),
                     nc=colorama.Fore.RESET,
                     spacing="\t",
                     entry=entry,
--- a/dvc/command/checkout.py
+++ b/dvc/command/checkout.py
@@ -98,7 +98,7 @@ def add_parser(subparsers, parent_parser):
         "--summary",
         action="store_true",
         default=False,
-        help="Show summary of the changes",
+        help="Show summary of the changes.",
     )
     checkout_parser.add_argument(
         "-d",
--- a/dvc/command/checkout.py
+++ b/dvc/command/checkout.py
@@ -1,18 +1,94 @@
 import argparse
+import logging
+
+import colorama
 
 from dvc.command.base import append_doc_link
 from dvc.command.base import CmdBase
+from dvc.exceptions import CheckoutError
+
+logger = logging.getLogger(__name__)
+
+
+def _human_join(words=None):
+    words = list(words or [])
+    if not words:
+        return ""
+
+    return (
+        "{before} and {after}".format(
+            before=", ".join(words[:-1]), after=words[-1],
+        )
+        if len(words) > 1
+        else words[0]
+    )
+
+
+def log_summary(stats):
+    message = [
+        ("added", "{added} added"),
+        ("deleted", "{deleted} deleted"),
+        ("modified", "{modified} modified"),
+    ]
+    summary = {
+        stat: len(stats[stat]) for stat, num in message if stats.get(stat)
+    }
+    if not summary:
+        return
+
+    template = _human_join(
+        fragment for stat, fragment in message if stat in summary
+    )
+    logger.info(template.format_map(summary))
+
+
+def log_changes(stats):
+    colors = [
+        ("modified", colorama.Fore.YELLOW,),
+        ("added", colorama.Fore.GREEN),
+        ("deleted", colorama.Fore.RED,),
+    ]
+
+    for state, color in colors:
+        entries = stats.get(state)
+
+        if not entries:
+            continue
+
+        for entry in entries:
+            logger.info(
+                "{color}{state}{nc}{spacing}{entry}".format(
+                    color=color,
+                    state=state[0].capitalize(),
+                    nc=colorama.Fore.RESET,
+                    spacing="\t",
+                    entry=entry,
+                )
+            )
 
 
 class CmdCheckout(CmdBase):
     def run(self):
-        self.repo.checkout(
-            targets=self.args.targets,
-            with_deps=self.args.with_deps,
-            force=self.args.force,
-            relink=self.args.relink,
-            recursive=self.args.recursive,
-        )
+        stats, exc = None, None
+        try:
+            stats = self.repo.checkout(
+                targets=self.args.targets,
+                with_deps=self.args.with_deps,
+                force=self.args.force,
+                relink=self.args.relink,
+                recursive=self.args.recursive,
+            )
+        except CheckoutError as _exc:
+            exc = _exc
+            stats = exc.stats
+
+        if self.args.show_changes:
+            log_changes(stats)
+        else:
+            log_summary(stats)
+
+        if exc:
+            raise exc
         return 0
 
 
@@ -25,6 +101,12 @@ def add_parser(subparsers, parent_parser):
         description=append_doc_link(CHECKOUT_HELP, "checkout"),
         help=CHECKOUT_HELP,
         formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+    checkout_parser.add_argument(
+        "--show-changes",
+        action="store_true",
+        default=False,
+        help="Show list of changes",
     )
     checkout_parser.add_argument(
         "-d",
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -261,7 +261,7 @@ class CheckoutError(DvcException):
         self.stats = stats
         targets = [str(t) for t in target_infos]
         m = (
-            "Checkout failed for following targets:\n {}\nDid you "
+            "Checkout failed for following targets:\n{}\nDid you "
             "forget to fetch?".format("\n".join(targets))
         )
         super().__init__(m)
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -256,7 +256,9 @@ class UploadError(DvcException):
 
 
 class CheckoutError(DvcException):
-    def __init__(self, target_infos):
+    def __init__(self, target_infos, stats=None):
+        self.target_infos = target_infos
+        self.stats = stats
         targets = [str(t) for t in target_infos]
         m = (
             "Checkout failed for following targets:\n {}\nDid you "
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -106,17 +106,13 @@ class ExternalRepo(Repo):
             if out.changed_cache(filter_info=src):
                 self.cloud.pull(out.get_used_cache(filter_info=src))
 
-            failed = False
             try:
                 out.checkout(filter_info=src)
             except CheckoutError:
-                failed = True
+                raise FileNotFoundError
 
             move(src, dest)
             remove(tmp)
-
-            if failed:
-                raise FileNotFoundError
 
     @wrap_with(threading.Lock())
     def _set_cache_dir(self):
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -11,7 +11,7 @@ from dvc.path_info import PathInfo
 from dvc.compat import fspath
 from dvc.repo import Repo
 from dvc.config import NoRemoteError, NotDvcRepoError
-from dvc.exceptions import NoRemoteInExternalRepoError
+from dvc.exceptions import NoRemoteInExternalRepoError, CheckoutError
 from dvc.exceptions import OutputNotFoundError, NoOutputInExternalRepoError
 from dvc.exceptions import FileMissingError, PathMissingError
 from dvc.utils.fs import remove, fs_copy, move
@@ -106,7 +106,11 @@ class ExternalRepo(Repo):
             if out.changed_cache(filter_info=src):
                 self.cloud.pull(out.get_used_cache(filter_info=src))
 
-            failed = out.checkout(filter_info=src)
+            failed = False
+            try:
+                out.checkout(filter_info=src)
+            except CheckoutError:
+                failed = True
 
             move(src, dest)
             remove(tmp)
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -868,7 +868,6 @@ class RemoteBASE(object):
         if progress_callback:
             progress_callback(str(path_info))
 
-        # relink is not modified
         return added, modified and not relink
 
     def makedirs(self, path_info):
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -14,6 +14,7 @@ from shortuuid import uuid
 
 import dvc.prompt as prompt
 from dvc.exceptions import (
+    CheckoutError,
     DvcException,
     ConfirmRemoveError,
     DvcIgnoreInCollectedDirError,
@@ -851,19 +852,24 @@ class RemoteBASE(object):
         self.remove(path_info)
 
     def _checkout_file(
-        self, path_info, checksum, force, progress_callback=None
+        self, path_info, checksum, force, progress_callback=None, relink=False
     ):
         """The file is changed we need to checkout a new copy"""
+        added, modified = True, False
         cache_info = self.checksum_to_path_info(checksum)
         if self.exists(path_info):
             logger.debug("data '{}' will be replaced.", path_info)
             self.safe_remove(path_info, force=force)
+            added, modified = False, True
 
         self.link(cache_info, path_info)
         self.state.save_link(path_info)
         self.state.save(path_info, checksum)
         if progress_callback:
             progress_callback(str(path_info))
+
+        # relink is not modified
+        return added, modified and not relink
 
     def makedirs(self, path_info):
         """Optional: Implement only if the remote needs to create
@@ -879,9 +885,11 @@ class RemoteBASE(object):
         relink=False,
         filter_info=None,
     ):
+        added, modified = False, False
         # Create dir separately so that dir is created
         # even if there are no files in it
         if not self.exists(path_info):
+            added = True
             self.makedirs(path_info)
 
         dir_info = self.get_dir_cache(checksum)
@@ -899,16 +907,23 @@ class RemoteBASE(object):
 
             entry_checksum_info = {self.PARAM_CHECKSUM: entry_checksum}
             if relink or self.changed(entry_info, entry_checksum_info):
+                modified = True
                 self.safe_remove(entry_info, force=force)
                 self.link(entry_cache_info, entry_info)
                 self.state.save(entry_info, entry_checksum)
             if progress_callback:
                 progress_callback(str(entry_info))
 
-        self._remove_redundant_files(path_info, dir_info, force)
+        modified = (
+            self._remove_redundant_files(path_info, dir_info, force)
+            or modified
+        )
 
         self.state.save_link(path_info)
         self.state.save(path_info, checksum)
+
+        # relink is not modified, assume it as nochange
+        return added, not added and modified and not relink
 
     def _remove_redundant_files(self, path_info, dir_info, force):
         existing_files = set(self.walk_files(path_info))
@@ -916,9 +931,11 @@ class RemoteBASE(object):
         needed_files = {
             path_info / entry[self.PARAM_RELPATH] for entry in dir_info
         }
-
-        for path in existing_files - needed_files:
+        redundant_files = existing_files - needed_files
+        for path in redundant_files:
             self.safe_remove(path, force)
+
+        return bool(redundant_files)
 
     def checkout(
         self,
@@ -966,12 +983,14 @@ class RemoteBASE(object):
                         self.path_info, checksum, filter_info
                     ),
                 )
-            return failed
+            if failed:
+                raise CheckoutError([failed])
+            return
 
         logger.debug("Checking out '{}' with cache '{}'.", path_info, checksum)
 
-        self._checkout(
-            path_info, checksum, force, progress_callback, relink, filter_info
+        return self._checkout(
+            path_info, checksum, force, progress_callback, relink, filter_info,
         )
 
     def _checkout(
@@ -985,8 +1004,9 @@ class RemoteBASE(object):
     ):
         if not self.is_dir_checksum(checksum):
             return self._checkout_file(
-                path_info, checksum, force, progress_callback=progress_callback
+                path_info, checksum, force, progress_callback, relink
             )
+
         return self._checkout_dir(
             path_info, checksum, force, progress_callback, relink, filter_info
         )
--- a/dvc/repo/checkout.py
+++ b/dvc/repo/checkout.py
@@ -83,9 +83,9 @@ def _checkout(
                 filter_info=filter_info,
             )
             for data in ["failed", "added", "modified"]:
-                stats[data] += [
+                stats[data].extend(
                     _fspath_dir(path, self.root_dir) for path in result[data]
-                ]
+                )
 
     if stats.get("failed"):
         raise CheckoutError(stats["failed"], stats)
--- a/dvc/repo/checkout.py
+++ b/dvc/repo/checkout.py
@@ -1,21 +1,31 @@
 import logging
+import os
 
+from dvc.compat import fspath
 from dvc.exceptions import CheckoutError
 from dvc.exceptions import CheckoutErrorSuggestGit
 from dvc.progress import Tqdm
-
+from dvc.utils import relpath
 
 logger = logging.getLogger(__name__)
 
 
-def _cleanup_unused_links(repo):
+def _get_unused_links(repo):
     used = [
         out.fspath
         for stage in repo.stages
         for out in stage.outs
         if out.scheme == "local"
     ]
-    repo.state.remove_unused_links(used)
+    return repo.state.get_unused_links(used)
+
+
+def _fspath_dir(path, root):
+    if not os.path.exists(str(path)):
+        return str(path)
+
+    path = relpath(fspath(path), root)
+    return os.path.join(path, "") if os.path.isdir(path) else path
 
 
 def get_all_files_numbers(pairs):
@@ -34,9 +44,19 @@ def _checkout(
 ):
     from dvc.stage import StageFileDoesNotExistError, StageFileBadNameError
 
+    unused = []
+    stats = {
+        "added": [],
+        "deleted": [],
+        "modified": [],
+        "failed": [],
+    }
     if not targets:
         targets = [None]
-        _cleanup_unused_links(self)
+        unused = _get_unused_links(self)
+
+    stats["deleted"] = [_fspath_dir(u, self.root_dir) for u in unused]
+    self.state.remove_links(unused)
 
     pairs = set()
     for target in targets:
@@ -52,20 +72,23 @@ def _checkout(
             raise CheckoutErrorSuggestGit(target) from exc
 
     total = get_all_files_numbers(pairs)
-    if total == 0:
-        logger.info("Nothing to do")
-    failed = []
     with Tqdm(
         total=total, unit="file", desc="Checkout", disable=total == 0
     ) as pbar:
         for stage, filter_info in pairs:
-            failed.extend(
-                stage.checkout(
-                    force=force,
-                    progress_callback=pbar.update_desc,
-                    relink=relink,
-                    filter_info=filter_info,
-                )
+            result = stage.checkout(
+                force=force,
+                progress_callback=pbar.update_desc,
+                relink=relink,
+                filter_info=filter_info,
             )
-    if failed:
-        raise CheckoutError(failed)
+            for data in ["failed", "added", "modified"]:
+                stats[data] += [
+                    _fspath_dir(path, self.root_dir) for path in result[data]
+                ]
+
+    if stats.get("failed"):
+        raise CheckoutError(stats["failed"], stats)
+
+    del stats["failed"]
+    return stats
--- a/dvc/stage.py
+++ b/dvc/stage.py
@@ -15,7 +15,7 @@ from voluptuous import Any, Schema, MultipleInvalid
 import dvc.dependency as dependency
 import dvc.output as output
 import dvc.prompt as prompt
-from dvc.exceptions import DvcException
+from dvc.exceptions import CheckoutError, DvcException
 from dvc.utils import dict_md5
 from dvc.utils import fix_env
 from dvc.utils import relpath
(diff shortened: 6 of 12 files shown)


dvc/remote/base.py
```diff
@@ -966,12 +978,14 @@ def checkout(
                         self.path_info, checksum, filter_info
                     ),
                 )
-            return failed
+            if failed:
+                raise CheckoutError([failed])
```
skshetry: A lot of tests will fail because of this and I'm really not sure if this is a good thing or bad. But, for now, it really makes the control flow easier.

skshetry: Tests seem to have passed. :see_no_evil:

dvc/stage.py
```diff
@@ -982,18 +982,24 @@ def checkout(
...
+                    progress_callback=progress_callback,
+                    relink=relink,
+                    filter_info=filter_info,
+                )
+                if is_modified:
+                    checkouts["modified"].append(out.path_info)
+                elif is_modified is False:
+                    checkouts["added"].append(out.path_info)
+            except CheckoutError as exc:
+                checkouts["failed"].append(*exc.target_infos)
```
Suor: Should this be `.extend()`?

dvc/state.py
```diff
@@ -453,20 +453,22 @@ def remove_unused_links(self, used):
...
+            if (inode, mtime) == (actual_inode, actual_mtime):
                 logger.debug("Removing '{}' as unused link.", path)
-                remove(path)
                 unused.append(relpath)
 
+        return unused
+
+    def remove_links(self, unused):
+        for path in unused:
+            remove(os.path.join(self.root_dir, path))
```
Suor: There was no join in original code, so this might be an error. BTW, paths are rarely rel to dvc root, they are either absolute or relative to cwd.

tests/func/test_state.py
```diff
@@ -69,16 +70,33 @@ def test_get_state_record_for_inode(get_inode_mock, tmp_dir, dvc):
...
 
         result = dvc.state._execute(cmd_count_links).fetchone()[0]
         assert result == 0
+
+
+def test_get_unused_links(tmp_dir, dvc):
+    tmp_dir.dvc_gen({"foo": "foo_content", "bar": "bar_content"})
+
+    with dvc.state:
+        links = [os.path.join(dvc.root_dir, link) for link in ["foo", "bar"]]
```
Suor: Why using absolute paths?

dvc/command/checkout.py
```diff
@@ -26,6 +85,18 @@ def add_parser(subparsers, parent_parser):
...
+        action="store_true",
+        default=False,
+        help="Show stats of changes",
+    )
+    checkout_parser.add_argument(
+        "--summary",
+        action="store_true",
+        default=False,
+        help="Show summary of changes",
+    )
```
efiop: Not sure if we need these as options, seems like we could show both by-default.

skshetry:
@efiop @Suor, the plan was to just show stats/summary. More verbose checkout changes did not seem too hard with existing stats/summary implementation so I quickly prototyped on it with `--show-changes`. I've asked @dmpetrov for feedback.

dvc/command/checkout.py
skshetry: Good point! We could make it default, but, I'll wait for product feedback.

dvc/remote/base.py
```diff
@@ -899,27 +906,35 @@ def _checkout_dir(
...
             if relink or self.changed(entry_info, entry_checksum_info):
+                _modified = True
                 self.safe_remove(entry_info, force=force)
                 self.link(entry_cache_info, entry_info)
                 self.state.save(entry_info, entry_checksum)
             if progress_callback:
                 progress_callback(str(entry_info))
 
-        self._remove_redundant_files(path_info, dir_info, force)
+        _modified = _modified or self._remove_redundant_files(
```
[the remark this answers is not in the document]
skshetry: Nice catch! Thanks.

dvc/stage.py
```diff
@@ -982,18 +982,24 @@ def checkout(
...
+                is_modified = out.checkout(
+                    force=force,
+                    tag=self.tag,
+                    progress_callback=progress_callback,
+                    relink=relink,
+                    filter_info=filter_info,
+                )
+                if is_modified:
+                    checkouts["modified"].append(out.path_info)
+                elif is_modified is False:
```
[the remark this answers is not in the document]
skshetry: For now, *added* is *False*, *modified* is *True*, and *no change* is *None*. :see_no_evil:

dvc/state.py
skshetry: I splitted the function as-is. Original code used `path` which was absolute path. But, we might not need it.

dvc/repo/checkout.py
```diff
@@ -59,13 +63,17 @@ def _checkout(
...
-                    relink=relink,
-                    filter_info=filter_info,
-                )
+            result = stage.checkout(
+                force=force,
+                progress_callback=pbar.update_desc,
+                relink=relink,
+                filter_info=filter_info,
             )
+            failed.extend(result["failed"])
```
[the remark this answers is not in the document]
skshetry: Missed it! Thanks.

dvc/remote/base.py
```diff
@@ -899,27 +906,35 @@ def _checkout_dir(
...
+            path_info, dir_info, force
+        )
 
         self.state.save_link(path_info)
         self.state.save(path_info, checksum)
 
+        if _added or _modified:
+            return not _added and _modified
+
```
pared: Shouldn't we return something when `_added=False` and `_modified=False`?

skshetry: I understand it's confusing, and I'll probably refactor it to return some kind of enum status. I'll leave it as it is for now, till I know more requirements (or, if this suffice). Thanks for the review!

dvc/command/checkout.py
```diff
@@ -1,18 +1,91 @@
...
+import colorama
 
 from dvc.command.base import append_doc_link
 from dvc.command.base import CmdBase
+from dvc.exceptions import CheckoutError
+
+logger = logging.getLogger(__name__)
+
+
+def _humanize_wordseries(words=None):
```
shcheklein: move it to the utils? there should be some libraries that do different "humanization" stuff ... not sure if we want to utilize them.

dvc/command/checkout.py
```diff
@@ -1,18 +1,91 @@
...
+    if with_failure:
+        template.append("{failed} failed")
+
+    summary = {stat: len(num) for stat, num in stats.items()}
+    logger.info(_humanize_wordseries(template).format_map(summary))
+
+
+def _log_changes(stats, with_failure=False):
+    colors = [
+        ("added", colorama.Fore.GREEN),
```
shcheklein: does it makes to utilize LS_COLORS as well? may be check how does git do this with diffs?

dvc/command/checkout.py
shcheklein: Would love to see them by default (summary at least)! 

may be --show-json to being able to script this, but it's another PR probably

shcheklein: If summary is default, then we can name option -l (--long) or something like this to show all of the output

dvc/remote/base.py
```diff
@@ -899,27 +906,36 @@ def _checkout_dir(
...
+            self._remove_redundant_files(path_info, dir_info, force)
+            or modified
+        )
 
         self.state.save_link(path_info)
         self.state.save(path_info, checksum)
 
+        if added or modified:
+            return not added and modified
+
```
JIoJIaJIu: This section is not really clear for me and it can return `None` and `bool`
Maybe just 
```python
return modified and not added
```

dvc/stage.py
```diff
@@ -982,18 +982,24 @@ def checkout(
...
-                tag=self.tag,
-                progress_callback=progress_callback,
-                relink=relink,
-                filter_info=filter_info,
-            )
-            if failed:
-                failed_checkouts.append(failed)
-        return failed_checkouts
+            try:
+                is_modified = out.checkout(
```
JIoJIaJIu: What do you think to apply one naming, either `is_modified` or `modified everywhere?

dvc/command/checkout.py
JIoJIaJIu: I've checked git, and it works based on `diff.color`/`color.diff` config option
[diff.c](https://github.com/git/git/blob/bfdd66e72fffd18235757bedbf355fd4176d6019/diff.c#L57)
[diff.h](https://github.com/git/git/blob/b02fd2accad4d48078671adf38fe5b5976d77304/diff.h#L389)
[diff.color.](https://github.com/git/git/blob/bfdd66e72fffd18235757bedbf355fd4176d6019/diff.c#L435)
```
$ git config --local color.diff.new yellow
$ git log -p
```
https://stackoverflow.com/a/26941235/484582
<hr/>

but gnu diffutils with [this commit](http://git.savannah.gnu.org/cgit/diffutils.git/commit/?id=04f6d57177fb6c9e7bbcf716c521bc323936a9a5) started to use theme with the same notation as ls does
But it isn't propagated from the env variable
```
defaults to @samp{rs=0:hd=1:ad=32:de=31:ln=36} for red deleted lines,
green added lines, cyan line numbers, bold header
```
```
$ echo 1 > file1
$ echo 2 > file2
$ diff file1 file2 --color --palette="ad=01;35:de=01;36"
...
```

dvc/command/checkout.py
skshetry: I'd like to keep it near to where it's used for now. If we later need it elsewhere, we can do so at that time.

dvc/command/checkout.py
```diff
@@ -1,18 +1,77 @@
...
+                recursive=self.args.recursive,
+            )
+        except CheckoutError as _exc:
+            exc = _exc
+
+        stats = stats if not exc else exc.stats
+        if self.args.show_changes:
+            _log_changes(stats, exc is not None)
+        if self.args.summary:
+            _log_summary(stats, exc is not None)
```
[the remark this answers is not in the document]
skshetry: We do checkout other files even if some have failed, so, I see *failed* as one state of a checkout too. So, if we make `checkout` show stats by default, it'd be helpful to provide information what we already have.

dvc/command/checkout.py
shcheklein: 👍 up to you, absolutely minor thing. 

Let me share my intuition behind this, though :)

- **Readability** - any boilerplate code that makes file longer and absolutely not relevant/generic makes it harder to see the "bigger picture" (tools like CodeClimate complain about file being too long, about too many methods, etc for a good reason);
- **Reusability** - the knowledge about this generic "text" util is only in your head. Low chances that someone will be able to find it when it's needed again. Means some probability of a duplicate later.
- **Testing** - you can unit test this simple things, would be strange to write a unit test for a private method though.

dvc/command/checkout.py
skshetry: The current output is very similar to `dvc diff`.

> does it makes to utilize LS_COLORS as well?

you mean, colorizing pathnames?

Another possible output format to use is of `git checkout`:
```sh
$ git checkout branch7
M	.gitignore
A	data2/data
Switched to branch 'branch7'
```

shcheklein: I meant more of a reusing some colorization/output style of Git (diff first came to my mind with red for removed, green for added). Idea to mimic `git checkout` is also good.

Instead of hardcoding color codes utilize some environment (in a way Gut did his reserach)

skshetry: We could color it after #3351 is merged. It'd be better to go with however `ls` works (and, not have too many color formats).

`git checkout` style is easily greppable, whereas `dvc diff` style output is not. 
```diff
$ dvc checkout
+ M data
! D foo
+ A models
- F something
# ignore the colors and symbols
```

We could make the pathnames follow `LS_COLORS`, whereas keep *{M,A,D,F}* as {yellow, green, orange, and red}* (or, something even better). What do you think, @shcheklein?

skshetry: But, I assume `dvc diff` can be easily understood.

dvc/command/checkout.py
skshetry: Like we discussed below around `git checkout`, it is never silent. It'll display every changes if there's any unless silenced by `-q`.

Okay, the api already gives us a dict of checkouts. So, it should be easy to implement. But, let's keep it in another PR.

dvc/command/checkout.py
shcheklein: I like it! Both idea to use git checkout format and LS_COLORS.

dvc/remote/base.py
skshetry: @JIoJIaJIu, thanks for the review. This is something that I will fix it later on (but, before, I'd like to find what's better in terms of product POV).

If you have any suggestions regarding how we could do this better, please do tell. :slightly_smiling_face:

dvc/command/checkout.py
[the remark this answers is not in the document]
skshetry: Summary about fail is no longer thrown. But, other details are still shown even if there's a failure.

dvc/command/checkout.py
skshetry: This is now the default (if there's no changes, there's no output though).

dvc/command/checkout.py
skshetry: Let's discuss on `LS_COLORS` separately. I have made the output similar to `git checkout`.

dvc/stage.py
skshetry: Thanks for the suggestion. I have fixed it.

dvc/repo/checkout.py
```diff
@@ -8,14 +8,14 @@
...
-def _cleanup_unused_links(repo):
+def _get_unused_links(repo):
     used = [
         out.fspath
         for stage in repo.stages
         for out in stage.outs
         if out.scheme == "local"
     ]
-    repo.state.remove_unused_links(used)
+    return repo.state.get_unused_links(used)
```
skshetry: TODO:
- [ ] `get_unused_links` returns directory name without `/` at the end

dvc/command/checkout.py
Suor: It's not like humanizing, it's joining and not word series but some items in a list, so the name might be better, e.g. `human_join()` or smth.

Suor: On keeping it here - I am with @skshetry here, move it later if that is used elsewhere. Spreading code is the big source of errors. Also impairs readability.

dvc/repo/checkout.py
Suor: Because it is not for presentation. I don't like where it's headed all internal code should not care about presentation, that will complicate things a lot.

skshetry: Yes, I'm not suggesting to change internal functions at all. Just the `_get_unused_links` function inside the `repo/checkout.py`. I'd love to handle it in `command`, but as we need to return dict (as we might need to support `--show-json` later), I feel this is the place to do. 

But, I agree with you. I am not liking this at all, sprinkling logic all over the place to return changes. If you have any better suggestion, please do tell.

dvc/stage.py
```diff
@@ -982,22 +983,30 @@ def checkout(
...
-                    checkouts["added"].append(out.path_info)
+                added, modified = result or (None, None)
+                if added:
+                    checkouts["added"].append(_fspath_dir(out.path_info))
+                elif modified:
+                    checkouts["modified"].append(_fspath_dir(out.path_info))
             except CheckoutError as exc:
-                checkouts["failed"].extend(exc.target_infos)
+                checkouts["failed"].extend(
+                    [target for target in _fspath_dir(exc.target_infos)]
```
JIoJIaJIu: is it misprint?
```python
[_fspath_dir(target) for target in exc.target_infos]
```

skshetry: Nice catch! Thanks. I should write more tests. :)

dvc/stage.py
```diff
@@ -982,22 +983,30 @@ def checkout(
         relink=False,
         filter_info=None,
     ):
+        def _fspath_dir(path):
+            """Return string with "/" at the end if it's a dir."""
+            p = os.path.join(path, "") if os.path.isdir(path) else fspath(path)
+            return relpath(p, self.repo.root_dir)
```
JIoJIaJIu: What do you think about it? (if I got the function purpose right)
```python
rel = relpath(p, self.repo.root_dir)
return "{}{}".format(rel, os.path.sep) if os.path.isdir(path) else rel
```

dvc/repo/checkout.py
skshetry: @Suor, we could return stats without stringifying filepaths (i.e. as PathInfo), and only add the `os.sep` on the deleted files beforehand (on `Repo._checkout()` itself). This way, presentation logic would be on the `cli` side. However, this will divide the logic anyway.

tests/func/test_state.py
skshetry: `get_unused_links` already expect it to be a abspath. So, I'd not like to change it here (at least, not in this PR).

shcheklein:
@skshetry would be super helpful if you show some examples (just here in the comments) - how would outputs looks like in different settings, and summary of changes that were introduced.

shcheklein:
Will be writing questions here:

1. If checkout fails it does not show any summary of changes, even though it deletes files, may be even adds? We should first show changes, then fail.

shcheklein:
@skshetry please see comments above.

skshetry:
> 1. If checkout fails it does not show any summary of changes, even though it deletes files, may be even adds? We should first show changes, then fail.

That should already happen (infact it's tested). Does that not work for you or, you meant something like:
```
1 added, 2 deleted and 3 failed
```

Why I did not do this is because, I find the error message fairly good:
```
$ dvc checkout
WARNING: Cache '8d1069328fcefdcc8a0969744b3ab7ab' not found. File 'foo' won't be created.
1 added and 1 deleted
ERROR: unexpected error - Checkout failed for following targets:
 foo
Did you forget to fetch?
```

and,
```
$ dvc checkout --show-changes
WARNING: Cache '8d1069328fcefdcc8a0969744b3ab7ab' not found. File 'foo' won't be created.
A	data/
D	bar
ERROR: unexpected error - Checkout failed for following targets:
 foo
Did you forget to fetch?
```

> 2. I don't see any reason to introduce an option - let's make it a default behavior. It looks like it is exactly what git does, --show-changes looks great! If we need to, we can create an option to show a summary instead.

I was worried that it'll be very noisy during checkout. So, I just went with showing just the summary by default. What do you think?

> 3. I'm getting:
> 
> ```
> Nothing to do
> ...
> ```

Looks like I forgot to handle this in a different PR. I'll just remove this message. Checkout will be silent if nothing changes.

> 
> 1. Do we show a summary of changes inside directory or just `M  dir/path`? Would be really nice to see at least a summary but it can be a separate PR.

Let's do that in a separate PR. But, like above, this might get very noisy. Should we introduce a flag for these, or only when a stage file for a directory is checked out?

> 2. It affects `dvc pull`:
> 
> ```
> (.env) [ivan@ivan ~/Projects/example-get-started]$ dvc pull
> Nothing to do
> Everything is up to date.
> ```
> 
> 1. Since `checkout` is the last step of the `dvc pull` we should show the same summary (+ pull specific summary that can be a separate PR)

Working on it. Introduce `--show-changes` flag here as well?

> 2. `--relink` is shown as modified - it okay, but can be better. Modified can be confusing here.

Thanks for mentioning this. I have fixed this.

> 3. test carefully with `-R`, --with-deps`, etc - with `-R` I was getting some confusing results.

Yeah, will check. Thanks for reminding.
I'll add a few screenshots here as well after I fixed your suggestions.

cc @shcheklein

shcheklein:
@skshetry thanks!

> That should already happen (infact it's tested). Does that not work for you or, you meant something like:

I will check again, but most likely you are right. Most likely I just missed this message. Both warnings are way more colorful and distract attention - you just don't see the summary.

 > Does that not work for you or, you meant something like

Number of failed would be good, but mostly I meant that I didn't see the message at all and thought that we missed it.

Do we need that warning? Also, if we switch the mode to the `--show-changed` one, it can eliminate the problem (partially). 

>  Introduce --show-changes flag here as well?

as I mentioned, I don't see the need for `--show-changes`. I would make it a default mode. I think we are overcomplicating this a bit. `--show-changes` works perfectly and behave like Git. The same for `pull`.

> Thanks for mentioning this. I have fixed this.

yes, we should show `--relink` is a different way, otherwise we have M for both - relinked and modified files - not sure this is a good idea.

skshetry:
> Looks good @skshetry!
> 2 things:
> Main question:
> 
> 1. `fetch` - don't we want to show a summary of how much files have been downloaded?
>

I'll prefer to keep those on different PR. Let's keep the scope of this to checkout for now (pull only shows info about checkout). I'll create a separate issue later on.

> Side comment:
> 2. Regarding @shcheklein comments on dir contents:
> I think we should create issue for that. We are currently treating all outs as single entities during checkout, so check out for 1M files directory and single file out will show `2 changed`. I think the next step could be to introduce something like `SyncResult` object, that will take care of gathering information on what has been done during `fetch/checkout/pull` operations. That way we would not have to return tuples of booleans from sync methods.

I really did not like introducing a *global* object. Instead of booleans, we could use some kind of `Status` enum object that is collected upwards and then summarized/processed.

skshetry:
> Do we need that warning? Also, if we switch the mode to the `--show-changed` one, it can eliminate the problem (partially).

What do you think of checkout being noisy? What's your view on this, @shcheklein?

> yes, we should show `--relink` is a different way, otherwise we have M for both - relinked and modified files - not sure this is a good idea.

Okay, I won't extend the scope of this PR. For now, `--relink` is no-change (so, no outputs unless anything other also got modified on the workspace). 

We can just have `relinked successfully` message on the end if `--relink` is used. I think, this should be okay (unless `--show-changes` is specified, where no information is show for relinked).

shcheklein:
@skshetry I don't see any good reason to show that warning (along with the `Checkout failed for following targets:` message). Also, I don't see any good reason to have a separate `--show-changes` - I would make it default - it imitates git behavior and provides really good and clean output.

> For now, --relink is no-change 

agreed, this is good! the most important is not show all of the as modified.

> We can just have relinked successfully message on the end

agreed again! should be enough.

skshetry:
> I don't see any good reason to show that warning (along with the `Checkout failed for following targets:` message).

So, you meant something like:
```sh
$ dvc checkout --show-changes
A	data/
D	bar
F     foo
$ echo $?
1
```

> Also, I don't see any good reason to have a separate `--show-changes` - I would make it default - > it imitates git behavior and provides really good and clean output.

So, how'd this work? Should we need summary to show by default if we also show changes by default? This will make outputs ungreppable, so, we made them work separately.

shcheklein:
> So, you meant something like:

I'm not sure about F to be honest. If we consider this an error - let's show that last error message. But I would remove WARNING that happen before output.

> This will make outputs ungreppable, so, we made them work separately.

@skshetry probably, no summary then - do we actually need it?

(btw, my comparison with git checkout is wrong, but it does not show summary as well, so it's probably better to not compare it at all)

skshetry:
I'd love to hear other's opinion on this as well. cc @iterative/engineering 

I think, showing every change on the checkout is too verbose (maybe I'm optimizing for large changes by default) and would only like to show summary:
 ```sh
dvc checkout
A	data/
D	bar
```

vs

```sh
dvc checkout
1 added and 1 deleted
```

pared:
@skshetry @shcheklein I would say that writing changes are not too verbose until we are staying with the current approach to summary (which is calculating them on `outs`). If we will start to aggregate, for example, the whole contents of the directory, that might be too much.

My reasoning for that is that the number of outs is limited indirectly by user, who "marks" outs with `dvc` commands.

skshetry:
Another thing in favor of showing complete changes by default is if user has hooks for `git checkout`. This will have pretty similar output formats.

casperdcl:
I prefer `+-` over `AD`, and `?` or `E` over `F`

dvc/repo/checkout.py
```diff
@@ -1,21 +1,31 @@
...
+
+
+def _fspath_dir(path, root):
+    if not os.path.exists(path):
+        return str(path)
+
+    path = relpath(fspath(path), root)
```
pared: Looking at tests for `3.5` path should be string here, which is not always the case.
If we have `path` in arguments, I think we should always put path here and not PathInfo object. Now it is mixed.

tests/func/test_checkout.py
```diff
@@ -513,5 +520,230 @@ def test_checkout_relink_protected(tmp_dir, dvc, link):
...
+    remote.remove(file_path)
+    stats = dvc.checkout(force=True)
+    assert stats == {**empty_checkout, "added": [str(file_path)]}
+    assert remote.exists(file_path)
+
+    remote.s3.put_object(
+        Bucket=remote.path_info.bucket, Key=file_path.path, Body="foo\nfoo"
+    )
+    stats = dvc.checkout(force=True)
+    assert stats == {**empty_checkout, "modified": [str(file_path)]}
```
skshetry: I only added this tests after your review, @pared.

skshetry:
@shcheklein, I had a discussion with Dmitry. The objective of the PR was to:
> To my mind, we should NOT make checkout silent.

I feel this PR does more than that already. So, I'd like to move forward with the PR as-is for now and keep the discussions ongoing. What do you say? We can reach a conclusion later on and could be handled on separate PRs.

tests/func/test_checkout.py
pared: Thanks for pointing, makes review easier :)

shcheklein:
@skshetry travis is failing?

> I feel this PR does more than that already. 

I agree, that's why I was surprised to see two additional optional initially :)

If we keep `--show-changes` (up to you, but I would still reconsider this - remove or make it default - I just don't see much of a reason in this option, and it was not part of the requirement), please don't forget to update:

- autocomplete scripts - bash and zsh
- docs
- ask @jorgeorpinel to review the help message

Also, let me do the final pass - I'll try to run and check the PR today.

dvc/exceptions.py
```diff
@@ -256,7 +256,9 @@ def __init__(self, amount):
 
 
 class CheckoutError(DvcException):
-    def __init__(self, target_infos):
+    def __init__(self, target_infos, stats=None):
+        self.target_infos = target_infos
+        self.stats = stats
         targets = [str(t) for t in target_infos]
         m = (
             "Checkout failed for following targets:\n {}\nDid you "
```
shcheklein: sorry, really minor request (feel free to nor change this), came out as I was reviewing:

```
ERROR: unexpected error - Checkout failed for following targets:
 data/prepared
data/features
data/data.xml
```

there is an extra space that is not needed before `data/prepared`. I think the change is `\n {}\n` -> `\n{}\n`

shcheklein:
@skshetry going back to the requirements in the #2979 (you mention in the description) , I think it does not resolve them for the failed case:

```
WARNING: Cache '6836f797f3924fb46fcfd6b9f6aa6416.dir' not found. File 'data/prepared' won't be created.
WARNING: Cache 'a304afb96060aad90176268345e10355' not found. File 'data/data.xml' won't be created.
WARNING: Cache '42c7025fc0edeb174069280d17add2d4.dir' not found. File 'data/features' won't be created.
1 added
ERROR: unexpected error - Checkout failed for following targets:
 data/prepared
data/data.xml
data/features
Did you forget to fetch?

```

we still have excessive WARNINGS . We can create a separate PR for this or keep the ticket open. I think a summary is enough:

```
1 added, 3 failed

ERROR: unexpected error - Checkout failed for following targets:
 data/prepared
data/data.xml
data/features
Did you forget to fetch?
```

Again, happy to just keep that ticket open.

shcheklein:
It still affects `dvc pull` in some unexpected (?) ways:

```
(.env) [ivan@ivan ~/Projects/example-get-started]$ dvc pull
2 added
Everything is up to date.
```

shcheklein:
(Also, going through the example get started I have more and more reasons for the `--show-changes` option by default. One more reason - we mix directories and files when we count added, modified, etc. E.g. when it prints `1 added` but actually it's a directory with two preprocessed files it make it confusing if `dvc checkout model.dvc` prints the same. Just a thought to consider.)

skshetry:
> it's a directory with two preprocessed files it make it confusing if `dvc checkout model.dvc` prints the same. Just a thought to consider.)

I'm also more inclined to make this the default, even more so with making output formats of `git` and `dvc` checkouts similar (helpful with `post-checkout`). So, to not make anymore changes, I'll swap the defaults, and introduce `--summary` flag.

Later, we could decide if we really need `--summary`, and then either remove or even rename those as `--shortstat`/ `--short`/`--stat`/`--numstat`, etc.

> It still affects `dvc pull` in some unexpected (?) ways:

Looks like on second example, `pull` did fetch something and did checkout (checkout looks to be working fine), but `pull` did not report anything (same in-line with previous behavior that it didn't print anything if it fetched something). I'll fix `fetch` and `pull` behavior on separate PRs.

> we still have excessive WARNINGS . We can create a separate PR for this or keep the ticket open. I think a summary is enough:

It'd be better to just reduce the level of loggings here to `DEBUG`. Again, I'll handle it in a separate PRs (added to TODO).

Thanks Ivan for checking out!

Suor:
@skshetry I don't think you can simply make those warnings debug, this will affect more things than just checkout, e.g. `Stage.run()` does not check for those checkouts to succeed.

dvc/external_repo.py
```diff
@@ -106,7 +106,11 @@ def _pull_cached(self, out, path_info, dest):
             if out.changed_cache(filter_info=src):
                 self.cloud.pull(out.get_used_cache(filter_info=src))
 
-            failed = out.checkout(filter_info=src)
+            failed = False
+            try:
+                out.checkout(filter_info=src)
+            except CheckoutError:
+                failed = True
```
Suor: Looks like there is an error here introduced earlier, one shouldn't move/remove if there is already an error. Simply raise `FileNotFoundError` here instead of using `failed` flag.

dvc/remote/base.py
```diff
@@ -851,20 +852,25 @@ def safe_remove(self, path_info, force=False):
...
             self.safe_remove(path_info, force=force)
+            added, modified = False, True
 
         self.link(cache_info, path_info)
         self.state.save_link(path_info)
         self.state.save(path_info, checksum)
         if progress_callback:
             progress_callback(str(path_info))
 
+        # relink is not modified
```
Suor: You don't really know that relink is not modified, so this is untrue. I see that you are going to handle relink later, but still there should not be any lying comments in code.

dvc/repo/checkout.py
```diff
@@ -52,20 +72,23 @@ def _checkout(
...
+            result = stage.checkout(
+                force=force,
+                progress_callback=pbar.update_desc,
+                relink=relink,
+                filter_info=filter_info,
             )
-    if failed:
-        raise CheckoutError(failed)
+            for data in ["failed", "added", "modified"]:
+                stats[data] += [
```
Suor:
```suggestion
                stats[data].extend(
```
No need to create an intermediate list nor rewrite a source.

dvc/repo/checkout.py
```diff
@@ -34,9 +44,19 @@ def _checkout(
...
+        "modified": [],
+        "failed": [],
+    }
     if not targets:
         targets = [None]
-        _cleanup_unused_links(self)
+        unused = _get_unused_links(self)
+
+    stats["deleted"] = [_fspath_dir(u, self.root_dir) for u in unused]
+    self.state.remove_links(unused)
```
Suor: Why is this not under if? Looks fragile this way, we shouldn't call `remove_links()` at all if a particular target is specified.

skshetry: It's empty if there's no argument.

skshetry: Also, I need to check if something is a dir before deleting.

Suor: There is also an argument, it might be an empty sequence, but it's there.

Suor: Checkout in dvc:
- checks out particular thing(s) if that things (target) is specified
- if no target is specified it checks out everything and removes anything previously checked out and having no references now.

I.e. if there is target we don't try to change the rest of the workspace, so `.remove_links()` should not be called. The logic of your code is different.

skshetry:
@shcheklein, I have changed the defaults. There's a `--summary` flag for showing summary instead of detaliled changelog. Other stuffs will be fixed in a separate PR.

dvc/repo/checkout.py
skshetry: If targets are specified, `unused` will be empty and hence, nothing will get removed. If your concern is regarding calling `.remove_links()`, I don't think there's any reason to be defensive.

shcheklein:
sorry, wanted to approve initially, missed the button )

dvc/command/checkout.py
```diff
@@ -26,6 +94,12 @@ def add_parser(subparsers, parent_parser):
         help=CHECKOUT_HELP,
         formatter_class=argparse.RawDescriptionHelpFormatter,
     )
+    checkout_parser.add_argument(
+        "--summary",
+        action="store_true",
+        default=False,
+        help="Show summary of the changes",
```
jorgeorpinel:
```suggestion
        help="Show summary of the changes.",
```

skshetry:
@shcheklein, @efiop, I have created an issue https://github.com/iterative/dvc/issues/3483 for handling autocompletion scripts.

PR #3403, opened 2020-02-25
erepo: set up more logging

Suor:
Added more logging to erepo code, combined with timestamps that we now have this might be used to investigate issues like #3280.

### DIFF
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -194,6 +194,7 @@ def _cached_clone(url, rev, for_write=False):
 
     # Copy to a new dir to keep the clone clean
     repo_path = tempfile.mkdtemp("dvc-erepo")
+    logger.debug("erepo: making a copy of {} clone", url)
     copy_tree(clone_path, repo_path)
 
     # Check out the specified revision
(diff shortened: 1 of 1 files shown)


dvc/external_repo.py
```diff
@@ -219,8 +220,10 @@ def _clone_default_branch(url, rev):
             git = Git(clone_path)
             # Do not pull for known shas, branches and tags might move
             if not Git.is_sha(rev) or not git.has_rev(rev):
+                logger.debug("erepo: git pull {}", url, rev)
```
skshetry:
```suggestion
                logger.debug("erepo: git pull {} {}", url, rev)
```

skshetry: I think, we need two braces here:
```suggestion
                logger.debug("erepo: git pull {} {}", url, rev)
```

dvc/external_repo.py
```diff
@@ -219,8 +220,10 @@ def _clone_default_branch(url, rev):
...
                 git.pull()
         else:
+            logger.debug("erepo: git clone {} to a temporary dir", url, rev)
```
efiop: Missing braces here too :)

dvc/external_repo.py
Suor: We don't need `rev` here.

dvc/external_repo.py
Suor: Same.

Suor:
Removed extra `rev`.

PR #3406, opened 2020-02-25
Use pre-commit for Git hooks

### LINKED ISSUE #3271: install: support pre-commit util
So far looks like we only need `pre-commit` to add support for `post-checkout` hook. Authors are ok with implementing it https://github.com/pre-commit/pre-commit/issues/1120 and the implementation is pretty straightforward. Might consider contributing that to provide the best experience for the users (as our dvc hooks get into conflict with out pre-commit hooks already).

andrewhare:
Fixes #3271.

**NOTE:** This is blocked by a pending `2.2.0` release of [`pre-commit`](https://github.com/pre-commit/pre-commit) to support `post-checkout` hooks.

This introduces a new flag to the `dvc install` command:

```
$ dvc install --use-pre-commit-tool
```

Running this command will do several things:

* Verify that `pre-commit` is installed on the system
* Write the existing DVC git hooks bash files to `.dvc/tmp/hooks` instead of to `.git/hooks`
* Build a `pre-commit` configuration that calls the hook files in `.dvc/tmp/hooks`.
* Check the current directory for an existing `.pre-commit-config.yaml`, if one is found, merge the configuration into it, otherwise dump the configuration to a new `.pre-commit-config.yaml`.


efiop:
@andrewhare Looks really good! Thank you so much for the PR!

For the record: We've discussed during the meeting, that it might be worth creating those hook files somewhere (e.g. in .dvc/ or .dvc/tmp/) and triggering those from pre-commit config, instead of writing shell wrapper directly into them.

andrewhare:
@efiop Thanks for the feedback! 

I like the idea of writing the hook files to disk (possibly reusing/refactoring `_install_hook` to write to the correct directory) and calling them there.

dvc/scm/git/__init__.py
```diff
@@ -287,9 +302,32 @@ def install(self):
...
+        self._install_hook("pre-commit", [], "status", hook_path_fn)
+        self._install_hook("pre-push", [], "push", hook_path_fn)
+
+        if use_pre_commit_tool:
+            self._integrate_pre_commit_tool()
+
+    def _integrate_pre_commit_tool(self):
+        if not which("pre-commit"):
+            raise DvcException("pre-commit is not installed")
+
```
efiop: I wonder if we really need to run `pre-commit` for the user, maybe config is enough + a hint `now run \`pre-commit install\` to install git hooks`?

dvc/scm/git/__init__.py
```diff
@@ -287,9 +302,32 @@ def install(self):
...
+            self._pre_commit_tool_hook_path("pre-commit"),
+            self._pre_commit_tool_hook_path("push"),
+            self._pre_commit_tool_hook_path("post-checkout"),
         )
-        self._install_hook("pre-commit", [], "status")
-        self._install_hook("pre-push", [], "push")
+
+        conf_yaml = os.path.join(self.root_dir, ".pre-commit-config.yaml")
+        if not os.path.isfile(conf_yaml):
+            check_call(["pre-commit", "install"])
```
efiop: This one is needed in case user has ` .pre-commit-hooks.yaml`, but no config? Or am I missing something?

tests/unit/scm/test_pre_commit_tool.py
```diff
@@ -0,0 +1,31 @@
...
+        existing_conf = {"foo": [1, 2, 3]}
+        merged_conf = merge_pre_commit_tool_confs(existing_conf, self.conf)
+        self.assertEqual(self.conf, merged_conf)
+
+    def test_merge_pre_commit_tool_confs(self):
+        existing_conf = {"repos": [{}]}
+        merged_conf = merge_pre_commit_tool_confs(existing_conf, self.conf)
+        # Merging the new conf in should append the new repo to the end of
+        # the existing repos array on the existing conf.
+        self.assertEqual(self.conf["repos"][0], merged_conf["repos"][1])
```
pared: Is there some particular reason to use unittest? We re moving towards using pytest, I think it would be good to make it test also pytest-based.

efiop:
@pared Did you install the latest pre-commit version? As mentioned in the header, it requires 2.2.0

efiop:
For the record: taking over this PR.

efiop:
Currently you need to `dvc install --use-pre-commit-tool` on each clone, as `pre-commit` won't have .dvc/tmp/hook/* scripts to run without it. Looks like it would be better to run hooks directly through dvc. Looking into it.

.pre-commit-config.yaml
```diff
@@ -1,17 +1,35 @@
...
+    language_version: python3
+    stages:
+    - push
+  - always_run: true
+    id: dvc-post-checkout
+    language_version: python3
+    stages:
+    - post-checkout
+  repo: https://github.com/andrewhare/dvc
+  rev: WIP-pre-commit-tool
```
efiop: Will be fixed after merge.

efiop:
Ok, merging to adjust on master. So far works good for me, but it is tied to github repo, so need to try out on our master.

shcheklein:
@efiop do we need adjust any docs because of this change?

efiop:
@shcheklein Yes, hence "Docs will come separately after."

dvc/command/install.py
```diff
@@ -27,4 +27,11 @@ def add_parser(subparsers, parent_parser):
         help=INSTALL_HELP,
         formatter_class=argparse.RawDescriptionHelpFormatter,
     )
+    install_parser.add_argument(
+        "--use-pre-commit-tool",
```
skshetry: It'd be good to have `--use-pre-commit` or just `--pre-commit`. Tool is kind of redundant here.

efiop: @skshetry Im worried about it being confused with `pre-commit` hooks.

Issue #3353, opened 2020-02-18
TypeError: 'Stage' object is not iterable


vernt:
I got an error running a command. This happened when I'd just cloned my repo, before running `dvc pull`.

```
$ dvc repro --verbose mystage.dvc                                                                                                                                                                                               
2020-02-18 04:32:06,196 DEBUG: PRAGMA user_version;                                                                                                                                                                                                                                                    
2020-02-18 04:32:06,196 DEBUG: fetched: [(3,)]                                                                                                                                                                                                                                                         
2020-02-18 04:32:06,196 DEBUG: CREATE TABLE IF NOT EXISTS state (inode INTEGER PRIMARY KEY, mtime TEXT NOT NULL, size TEXT NOT NULL, md5 TEXT NOT NULL, timestamp TEXT NOT NULL)                                                                                                                       
2020-02-18 04:32:06,197 DEBUG: CREATE TABLE IF NOT EXISTS state_info (count INTEGER)                                                                                                                                                                                                                   
2020-02-18 04:32:06,197 DEBUG: CREATE TABLE IF NOT EXISTS link_state (path TEXT PRIMARY KEY, inode INTEGER NOT NULL, mtime TEXT NOT NULL)                                                                                                                                                              
2020-02-18 04:32:06,197 DEBUG: INSERT OR IGNORE INTO state_info (count) SELECT 0 WHERE NOT EXISTS (SELECT * FROM state_info)                                                                                                                                                                           
2020-02-18 04:32:06,197 DEBUG: PRAGMA user_version = 3;                                                                                                                                                                                                                                                
2020-02-18 04:32:06,465 DEBUG: SELECT count from state_info WHERE rowid=?                                                                                                                                                                                                                              
2020-02-18 04:32:06,465 DEBUG: fetched: [(0,)]                                                                                                                                                                                                                                                         
2020-02-18 04:32:06,465 DEBUG: UPDATE state_info SET count = ? WHERE rowid = ?                                                                                                                                                                                                                         
2020-02-18 04:32:06,696 ERROR: unexpected error - nbunch is not a node or a sequence of nodes.                                                                                                                                                                                                         
------------------------------------------------------------                                                                                                                                                                                                                                           
Traceback (most recent call last):
  File "/home/user/.cache/pypoetry/virtualenvs/proj-ngDuziVt-py3.6/lib/python3.6/site-packages/networkx/classes/graph.py", line 1908, in bunch_iter
    for n in nlist:
[...]

PR #3421, opened 2020-02-28
pipelines: fix edges gathering

### LINKED ISSUE #3410: dvc pipeline show --ascii - visual bug since version 0.82.9
DVC version: 0.82.9 and following
reproduced on several platforms, like:
4.15.0-55-generic #60-Ubuntu
macOS High Sierra 10.13.6
macOS Catalina 10.15.2

Hi,

the command "dvc pipeline show --ascii" seems to have a bug since version 0.89.9, which does not display the dependencies properly, when there is a branch in a Pipeline. 

I created a simple pipeline, which looks like this:

**Pipeline with Version 0.82.8**
Pipeline with Version 0.82.8

If i use the command "dvc pipeline show --ascii validation.dvc" with version 0.82.8, the whole pipeline gets displayed properly.
If i upgrade my Version to e.g. 0.83.0, the connection between "val.dvc" and the previous Stage is missing. If i use the command on val.dvc itself, the stages "train.dvc" and "validation.dvc" are missing (see screenshots).

**Pipeline with Version 0.83.0**
Pipeline with Version 0.83.0

**Pipeline with Version 0.83.0, when using val.dvc in the command**
Pipeline with Version 0.83.0, when using val.dvc in the command

pared:

Fixes #3410

### DIFF
--- a/dvc/command/pipeline.py
+++ b/dvc/command/pipeline.py
@@ -61,7 +61,7 @@ class CmdPipelineShow(CmdBase):
                     for out in stage.outs:
                         edges.append((str(out), str(dep)))
         else:
-            for from_stage, to_stage in networkx.dfs_edges(G, target_stage):
+            for from_stage, to_stage in networkx.edge_dfs(G, target_stage):
                 if commands:
                     if to_stage.cmd is None:
                         continue
--- a/tests/func/test_pipeline.py
+++ b/tests/func/test_pipeline.py
@@ -237,3 +237,29 @@ class TestDvcRepoPipeline(TestDvc):
 
         pipelines = self.dvc.pipelines
         self.assertEqual(len(pipelines), 0)
+
+
+def test_split_pipeline(tmp_dir, dvc):
+    tmp_dir.dvc_gen("data", "source file content")
+    dvc.run(
+        deps=["data"],
+        outs=["data_train", "data_valid"],
+        cmd="echo train >> data_train && echo valid >> data_valid",
+    )
+    stage = dvc.run(
+        deps=["data_train", "data_valid"],
+        outs=["result"],
+        cmd="echo result >> result",
+    )
+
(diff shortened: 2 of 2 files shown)


dvc/command/pipeline.py
```diff
@@ -61,7 +61,7 @@ def _build_graph(self, target, commands, outs):
                     for out in stage.outs:
                         edges.append((str(out), str(dep)))
         else:
-            for from_stage, to_stage in networkx.dfs_edges(G, target_stage):
+            for from_stage, to_stage in networkx.edge_dfs(G, target_stage):
```
efiop: This is not a complaint or anything like that, but just wanted to say that this is actually a really good case where a PR/patch description would help understand the difference :) Just like a short note about "we are currently using dfs_edges, but it is not what we need because ... , so we need to use edge_dfs". Since you dove into researching this, your summary would have a great value :slightly_smiling_face:

efiop:
@pared Great catch! :pray:  Let's add a test (probably unit test for `_build_graph`, if that makes sense to you) for it, so we are covered from regressions in the future.

dvc/command/pipeline.py
pared: Thats a valid point, Ill do that, it will probably be the best to include that in original issue, right?

pared:
@efiop I will prepare test
Summary of bug reason can be found in the original issue.

dvc/command/pipeline.py
skshetry: @pared, could this be related with this: https://github.com/iterative/dvc/issues/3353?

pared: @skshetry I don't think so. This issue was purely about `dvc pipelines`, and in #3353 throw happens upon `reproduce`. I would need to dive deeper into #3353 to say more.

PR #3426, opened 2020-03-02
api: docstring updates (2)

### LINKED ISSUE #3092: api: inline documentation (docstrings)

jorgeorpinel:
Continues #3352
to match iterative/dvc.org/pull/908

Closes #3092

### DIFF
--- a/dvc/api.py
+++ b/dvc/api.py
@@ -84,6 +84,5 @@ def _make_repo(repo_url=None, rev=None):
 
 
 def _require_dvc(repo):
-    """Raises UrlNotDvcRepoError if repo is not a DVC repository."""
     if not isinstance(repo, Repo):
         raise UrlNotDvcRepoError(repo.url)
--- a/dvc/api.py
+++ b/dvc/api.py
@@ -7,21 +7,21 @@ from dvc.external_repo import external_repo
 
 
 class UrlNotDvcRepoError(DvcException):
-    """Thrown if given URL is not a DVC repository.
-
-    Args:
-        url (str): URL to the repository
-    """
+    """Thrown if the given URL is not a DVC repository."""
 
     def __init__(self, url):
         super().__init__("'{}' is not a DVC repository.".format(url))
 
 
 def get_url(path, repo=None, rev=None, remote=None):
-    """Returns URL to the storage location of a data artifact tracked
-    by DVC, specified by its path in a repo.
+    """
+    Returns the URL to the storage location of a data file or directory tracked
+    in a DVC repo. Its formed by reading the remote configuration, and the
+    DVC-file where the given path is an output. For Git repos, HEAD is used
+    unless a rev argument is supplied.
 
-    NOTE: There's no guarantee that the file actually exists in that location.
+    NOTE: This function does not check for the actual existence of the file or
+    directory in the remote storage.
     """
     with _make_repo(repo, rev=rev) as _repo:
         _require_dvc(_repo)
@@ -31,7 +31,11 @@ def get_url(path, repo=None, rev=None, remote=None):
 
 
 def open(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
-    """Context manager to open a tracked file as a file object."""
+    """
+    Open file or model (path) tracked in a repo (by DVC or Git), and and return
+    a file object. For Git repos, HEAD is used unless a rev argument is
+    supplied. The default remote is tried unless a remote argument is supplied.
+    """
     args = (path,)
     kwargs = {
         "repo": repo,
@@ -63,7 +67,7 @@ def _open(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
 
 
 def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
-    """Returns the contents of a tracked file."""
+    """Returns the contents of a tracked file (by DVC or Git). Wraps open()"""
     with open(
         path, repo=repo, rev=rev, remote=remote, mode=mode, encoding=encoding
     ) as fd:
--- a/dvc/api.py
+++ b/dvc/api.py
@@ -18,10 +18,10 @@ class UrlNotDvcRepoError(DvcException):
 
 
 def get_url(path, repo=None, rev=None, remote=None):
-    """
-    Returns the full URL to the data artifact specified by its `path` in a
-    `repo`.
-    NOTE: There is no guarantee that the file actually exists in that location.
+    """Returns URL to the storage location of a data artifact tracked
+    by DVC, specified by its path in a repo.
+
+    NOTE: There's no guarantee that the file actually exists in that location.
     """
     with _make_repo(repo, rev=rev) as _repo:
         _require_dvc(_repo)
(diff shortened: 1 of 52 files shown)


dvc/api.py
```diff
@@ -63,7 +70,7 @@ def _open(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
 
 
 def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
-    """Returns the contents of a tracked file."""
+    """Returns the contents of a tracked file (by DVC or Git). Wraps open()"""
```
skshetry: `Wraps open()` is not important for consumers here.

dvc/api.py
```diff
@@ -81,8 +88,3 @@ def _make_repo(repo_url=None, rev=None):
             pass  # fallthrough to external_repo
     with external_repo(url=repo_url, rev=rev) as repo:
         yield repo
-
-
-def _require_dvc(repo):
-    if not isinstance(repo, Repo):
-        raise UrlNotDvcRepoError(repo.url)
```
skshetry: Any reason why this was shifted?

dvc/api.py
```diff
@@ -7,31 +7,38 @@
...
+            raise UrlNotDvcRepoError(_repo.url)
         out = _repo.find_out_by_relpath(path)
         remote_obj = _repo.cloud.get_remote(remote)
         return str(remote_obj.checksum_to_path_info(out.checksum))
 
 
 def open(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
-    """Context manager to open a tracked file as a file object."""
+    """
+    Open file or model (path) tracked in a repo (by DVC or Git), and and return
```
Suor: The first sentence is very confusing and misleading. Also hard to read because all of the parentheses. 

> Open file or model (path)

How can I open a model or a path?

> a repo (by DVC or Git)

does it mean "dvc or git repo"?

> return a file object

It doesn't return a file object, it returns a context manager. So this is misleading, which will lead to people trying to use it wrong and frustrating. Maybe the only easy way to explain how it works is a small example.

dvc/api.py
```diff
@@ -7,31 +7,38 @@
...
         super().__init__("'{}' is not a DVC repository.".format(url))
 
 
 def get_url(path, repo=None, rev=None, remote=None):
-    """Returns URL to the storage location of a data artifact tracked
-    by DVC, specified by its path in a repo.
+    """
+    Returns the URL to the storage location of a data file or directory tracked
+    in a DVC repo. Its formed by reading the remote configuration, and the
+    DVC-file where the given path is an output. For Git repos, HEAD is used
```
Suor: This thing of how it's formed blurs things more than explains unless you already very familiar with dvc. I would just say it gives you an url to a copy of file in a remote cache.

dvc/api.py
shcheklein: > It doesn't return a file object, it returns a context manager.

this is my bad. I just didn't expect this behavior from the function and didn't believe when I read it in the initial PR for docs. It's a separate question why did we make it this way, but we need to find a way to write it now, just "returns a context manager" even worse than "returns a files object" (taken from stdlib's open, quite excepted to behave this way here I would say. 

@Suor any suggestions?

dvc/api.py
jorgeorpinel: OK, then I need to copy paste the rest of the docstring from open() – about defaults, etc.

dvc/api.py
jorgeorpinel: The function was only used once.

jorgeorpinel: Is there a suggestion?

dvc/api.py
Suor: We should probably drop `mode` though, it doesn't make sense for read. Set 
```python
mode = "rb" if encoding is None else "r"
```

dvc/api.py
Suor: My suggestion above: show a quick example.

Suor: The reason we made it a context manager is the control of resources though, which is not working very well since we added caching. We might reconsider.

dvc/api.py
jorgeorpinel: OK, updated in f29e5ecb.

> We should probably drop mode

You mean the actual fn parameter @Suor? OK, up to you (separate PR I assume) but it's not mentioned in the docstring anyway 🙂

jorgeorpinel: p.s. Resolving this one.

dvc/api.py
jorgeorpinel: > Open file or model (path)

I'm trying to use the name of the parameters in the description, since we don't have structured sections such as `Params:` or `@param path str blah`. I agree the parentheses here were unnecessary: will use "file or model in the supplied path" instead.

p.s. I still think using backquotes would make including parameter names easier in these descriptions e.g. just "``file or model in `path` ``" or "``tried unless `remote` is supplied.``" – so its obvious the literals are arg names, not just words in a longer sentence (rn there's some ambiguity).

> tracked in a repo (by DVC or Git)

The stuff in parentheses refers more to "tracked" than to to "repo". Should we change to "tracked (by DVC or Git) in a repo"? I don't like how the parentheses interrupt the reading there. Thinking... ⏳

jorgeorpinel: > tracked in a repo (by DVC or Git)

OK I'm changing this parenthesis to "(both DVC and plain Git repos are supported)". Better?

jorgeorpinel: > It doesn't return a file object, it returns a context manager... Maybe the only easy way to explain how it works is a small example.

OK how about I say "and produces a file object (see example below)" and add the example? Done now, please check!

dvc/api.py
jorgeorpinel: True. I just removed that one sentence now (and completed the default arg value info, see 7f083979).

dvc/api.py
jorgeorpinel: I changed "produces" for "generates" now since I see yielding is involved here.

Or should I use "yields" directly, implying it's a generator? (The context manager returned is implemented [as a generator](https://book.pythontips.com/en/latest/context_managers.html#implementing-a-context-manager-as-a-generator), right?)

> See also the latest updates in iterative/dvc.org/pull/908, matching all this.

Suor: `yield ` is an implementation detail, there are other ways to make a context manager, so we don't have a generator here. Still hard to read, I would simply say:

```
Opens a file at specified path in a given repo. 
Should be used as a context manager:

    with dvc.api.open(
        "path/from/repo/root/file.ext",
        repo="https://example.com/url/to/repo",
    ) as fd:
        # ... Handle file object fd

Tries to stream file contents without downloading it. ...
Work for both dvc managed files and git handled ones, 
including ones in non-dvc git repos.
```

jorgeorpinel: OK I'm using something like that now, please check.

> I'm not sure I like that the only mention of the file object generated is in the example code comment (`Handle file object fd`) but OK, sounds like open() may change in the near future anyway (to behave more like builtin open)

PR #3428, opened 2020-03-02
add: do not verify hardlink if file is empty

### LINKED ISSUE #3390: add: empty files add broken when cache mode is hardlinks
## DVC Version

```
DVC version: 0.86.5+f67314.mod
Python version: 3.7.6
Binary: False
Package: None
```

## Reproduce

```
dvc config cache.type hardlink
touch foo
dvc add foo
```

outputs:

```
ERROR: no possible cache types left to try out.
```

skshetry:
Fixes #3390

### DIFF
--- a/dvc/remote/local.py
+++ b/dvc/remote/local.py
@@ -92,6 +92,12 @@ class RemoteLOCAL(RemoteBASE):
 
         return not self.changed_cache(current_md5)
 
+    def _verify_link(self, path_info, link_type):
+        if link_type == "hardlink" and self.getsize(path_info) == 0:
+            return
+
+        super()._verify_link(path_info, link_type)
+
     def is_empty(self, path_info):
         path = path_info.fspath
 
--- a/tests/func/test_add.py
+++ b/tests/func/test_add.py
@@ -675,13 +675,10 @@ def test_add_empty_files(tmp_dir, dvc, link):
     assert os.path.exists(stages[0].outs[0].cache_path)
 
 
-@mock.patch(
-    "dvc.remote.local.RemoteLOCAL.is_hardlink",
-    side_effect=RemoteLOCAL.is_hardlink,
-)
-def test_add_optimization_for_hardlink_on_empty_files(m, tmp_dir, dvc):
+def test_add_optimization_for_hardlink_on_empty_files(tmp_dir, dvc, mocker):
     dvc.cache.local.cache_types = ["hardlink"]
     tmp_dir.gen({"foo": "", "bar": "", "lorem": "lorem", "ipsum": "ipsum"})
+    m = mocker.spy(RemoteLOCAL, "is_hardlink")
     stages = dvc.add(["foo", "bar", "lorem", "ipsum"])
 
     m.assert_called_once()
--- a/tests/func/test_add.py
+++ b/tests/func/test_add.py
@@ -7,7 +7,7 @@ import time
 
 import colorama
 import pytest
-from mock import patch
+from mock import patch, call
 
 import dvc as dvc_module
 from dvc.cache import Cache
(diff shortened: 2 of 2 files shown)


dvc/remote/local.py
```diff
@@ -188,6 +188,8 @@ def hardlink(self, from_info, to_info):
                     src=str(from_info), dest=str(to_info)
                 )
             )
+            # skip verifying that hardlink is created
+            self.cache_type_confirmed = True
```
efiop: What will happen if we have 2 files that we need to create. First one is an empty one and the second one is a non-empty one and if hardlinks are not actually supported? Will link detection logic be able to recover from it even if it sees `            self.cache_type_confirmed = True` after empty file creation?

skshetry: Looks like it does not. Thanks for pointing that out.

skshetry: Looks like the `_verify_link` only verified the first time and later set `self.cache_type_confirmed = True`, so, no checks seem to have been done afterwards. Same as this current PR behavior which needs to be fixed.

https://github.com/iterative/dvc/blob/05cc023164766fbbec91642d7d8af513c8d4079b/dvc/remote/base.py#L377-L386

dvc/remote/local.py
```diff
@@ -188,7 +188,7 @@ def hardlink(self, from_info, to_info):
                     src=str(from_info), dest=str(to_info)
                 )
             )
-            return
+            return True
 
         System.hardlink(from_info, to_info)
```
pared: What will be returned here? Shouldn't we explicitly return False when we don't want to skip? 

Also, if there is a problem with link verification, maybe we should handle this particular use case in `_verify_link`?
When link type is hardlink and file is empty, don't set `cache_type_confirmed` but also don't raise Exception, because this particular case is special in our situation? What do you think?

skshetry: Too lazy to change this on every kind of links(they are in SSH as well). I also didn't want to spread the information of this optimization all around the codebase.

skshetry: If it'd be easier, I'd return `True` in case of explicit need of verification because `hardlink` returning True in optimization kind of looks awkward.

pared: @skshetry writing empty file looks awkward in our `remote/local.hardlink` :)
Also, if that is a special situation, why not override `_verfiy_link` in local?
I would think something like:

```
    def _verify_link(self, path_info, link_type):
        if self.cache_type_confirmed:
            return

        if link_type is "hardlink" and self.getsize(path_info) == 0:
            return
        super()._verify_link(path_info, link_type)
```

It's not pretty, but at least we keep it in one place. 
Having said so, it's opinionated, so I won't be pushing for this approach.

skshetry:
We (@efiop and I) discussed this on 1o1. A better way to fix this would be to try to create a temporary file to check if the **System** supports `{ref,hard,sym}`link. Though, it'll need some refactor, so, I'll create an issue for working on this later.

dvc/remote/local.py
skshetry: @JIoJIaJIu @pared, still, `verify_link` is not verifying anything (I mean, it shouldn't even reach here). @pared, thanks for the suggestion, it's not as convoluted as my PR, so, I'll go with what you suggested. I'll create a separate issue to fix this *hack* later.

dvc/remote/local.py
```diff
@@ -92,6 +92,12 @@ def already_cached(self, path_info):
 
         return not self.changed_cache(current_md5)
 
+    def _verify_link(self, path_info, link_type):
+        if link_type == "hardlink" and self.getsize(path_info) == 0:
+            return
+
```
efiop: Not great that it will be checking the size for each link it creates, might get expensive.

skshetry: We did that anyway on the `hardlink()` though.

efiop: @skshetry Maybe there is some nicer way to do this? Like making `hardlink`(and other link class methods) verify themselves? This is an honest question, I don't know myself either :slightly_smiling_face:

efiop: @skshetry Could keep it as is and create a ticket for it to reconsider later. Just asking.

skshetry: I like this suggestion https://github.com/iterative/dvc/pull/3428#issuecomment-596889084 more. Running verification on `hardlink` itself for only once (by setting `self.cache_type_confirmed`) is also hackish.

I'd say, we revisit this on next sprint and fix it?

efiop: @skshetry Ok, please create an issue and add it to the next sprint.

efiop:
@skshetry Check your tests, travis failed.

tests/func/test_add.py
```diff
@@ -662,3 +662,39 @@ def test_not_raises_on_re_add(tmp_dir, dvc):
...
+
+    assert (tmp_dir / file).exists()
+    assert (tmp_dir / (file + Stage.STAGE_FILE_SUFFIX)).exists()
+    assert os.path.exists(stages[0].outs[0].cache_path)
+
+
+@mock.patch(
+    "dvc.remote.local.RemoteLOCAL.is_hardlink",
+    side_effect=RemoteLOCAL.is_hardlink,
+)
```
pared: You can use `mocker.spy` for that. Sample usage:
`test_should_remove_outs_before_import`.

skshetry: Thanks @pared. Adjusted. :fire:

dvc/remote/local.py
pared: We could just check for `cache_type_verified` at the beggining again. I know that is duplication, but don't see any obvious workaround.

tests/func/test_add.py
```diff
@@ -662,3 +662,36 @@ def test_not_raises_on_re_add(tmp_dir, dvc):
...
+    assert os.path.exists(stages[0].outs[0].cache_path)
+
+
+def test_add_optimization_for_hardlink_on_empty_files(tmp_dir, dvc, mocker):
+    dvc.cache.local.cache_types = ["hardlink"]
+    tmp_dir.gen({"foo": "", "bar": "", "lorem": "lorem", "ipsum": "ipsum"})
+    m = mocker.spy(RemoteLOCAL, "is_hardlink")
+    stages = dvc.add(["foo", "bar", "lorem", "ipsum"])
+
+    m.assert_called_once()
```
pared: I think spy object does not have this method, you can do `assert m.call_count == 1`

skshetry: Seems like it does: https://github.com/pytest-dev/pytest-mock/#spy

efiop:
@skshetry Please check the tests.

tests/func/test_add.py
skshetry: @pared, looks like you are right. It fails on Python3.5 for some reasons (on other, it works fine). using what you suggested.

skshetry: `assert_called_once` was only added on Python3.6.  https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once 🤦

PR #3430, opened 2020-03-02
exceptions: use semantic exceptions

skshetry:
From https://github.com/iterative/dvc/pull/3363/files#r382117306

### DIFF
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -323,3 +323,7 @@ class RemoteCacheRequiredError(DvcException):
                 format_link("https://man.dvc.org/config#cache"),
             )
         )
+
+
+class IsADirectoryError(DvcException):
+    """Raised when a file operation is requested on a directory."""
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -11,6 +11,7 @@ from funcy import cached_property, cat
 from dvc.config import Config
 from dvc.exceptions import (
     FileMissingError,
+    IsADirectoryError,
     NotDvcRepoError,
     OutputNotFoundError,
 )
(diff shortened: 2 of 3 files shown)


dvc/exceptions.py
```diff
@@ -323,3 +323,7 @@ def __init__(self, path_info):
                 format_link("https://man.dvc.org/config#cache"),
             )
         )
+
+
+class IsADirectoryError(DvcException):
+    """Raised when a file operation is requested on a directory."""
```
skshetry: This is the same error as [IsADirectoryError](https://docs.python.org/3/library/exceptions.html#IsADirectoryError). But, it does not inherit from `OSError` for obvious reasons. I did not feel good to raise `InvalidArgumentError` because the argument was not invalid in the strict sense, it's property was.

I can remove this if we don't need this though. Too many exceptions, I know. :slightly_smiling_face:

Suor:
I would say our biggest issue with exceptions that we have too many error classes :). This might be useful for testing purposes, but only complicates the rest of the code. Adding to that is our `Repo` object is still private unlike `dvc.api`, so wouldn't spent much effort into polishing things like this for now. 

On this particular change. Why we don't simply raise a built in `IsADirectoryError` error here though? We wrap everything, i.e. all expected errors into `DvcException` to be presented to a user via cli nicely, but this method is called from api. This is why it was a simple `ValueError`.

We may do both of cause, I mean subclass from both `DvcException` and `IsADirectoryError`.

dvc/exceptions.py
skshetry: @Suor, regarding the comments below, `IsADirectoryError` is an `OSError`, and since this is not a system-related error/IO failure and we are only using it internally, I really didn't feel comfortable to have both `DvcException` and `OSError` in the inheritance chain. I only wanted to maintain the semantics of it.

Suor: It depends on how you look at it. Anyway, what is the purpose of this PR?

JIoJIaJIu: > this is not a system-related error/IO failure

As I see by usage it 's I/O error - we cannot read the dir. What is the purpose of creating custom `IsADirectoryError`? I think if we need custom exception it should has another abstraction level depends on the needs

skshetry: > Anyway, what is the purpose of this PR?

To make every exception we throw to be `DvcException`.

PR #2707, opened 2019-11-01
perf: refactor URLInfo to not use ParseResult inside

Suor:
This permites URLInfo construction from parts without building/reparsing
string, which makes it around 20% faster. Code is also simpler and a bit
shorter now.

Other improvements:
- added URLInfo.replace(path=...)
- URLInfo._path now stringifies properly
- protected against passing path not starting with /
- do not allow query, params, fragments and passwords in urls (they were
silently ignored previously)

* [x] Have you followed the guidelines in our
      [Contributing document](https://dvc.org/doc/user-guide/contributing/core)?

* [x] Does your PR affect documented changes or does it add new functionality
      that should be documented? If yes, have you created a PR for
      [dvc.org](https://github.com/iterative/dvc.org) documenting it or at
      least opened an issue for it? If so, please add a link to it.

-----

PR #3432, opened 2020-03-02
import-url: allow queries in URL

### LINKED ISSUE #3424: import-url: does not work for URL that contains a query string
```
(.env) [ivan@ivan ~/Projects/test-gs]$ dvc version
DVC version: 0.86.5+1f4fbf.mod
Python version: 3.7.6
Binary: False
Package: None
```

`dvc import-url -v https://www.dropbox.com/?test data`

gets you:

```
2020-02-29 21:01:49,849 ERROR: unexpected error
  File "/Users/ivan/Projects/dvc/dvc/main.py", line 49, in main
    ret = cmd.run()
  File "/Users/ivan/Projects/dvc/dvc/command/imp_url.py", line 16, in run
    self.args.url, out=self.args.out, fname=self.args.file
  File "/Users/ivan/Projects/dvc/dvc/repo/__init__.py", line 28, in wrapper
  File "/Users/ivan/Projects/dvc/dvc/repo/scm_context.py", line 4, in run
    result = method(repo, *args, **kw)
  File "/Users/ivan/Projects/dvc/dvc/repo/imp_url.py", line 20, in imp_url
    accompany_outs=True,
  File "/Users/ivan/Projects/dvc/dvc/stage.py", line 534, in create
    stage, kwargs.get("deps", []), erepo=kwargs.get("erepo", None)
  File "/Users/ivan/Projects/dvc/dvc/dependency/__init__.py", line 77, in loads_from
  File "/Users/ivan/Projects/dvc/dvc/dependency/__init__.py", line 61, in _get
    return d(stage, p, info)
  File "/Users/ivan/Projects/dvc/dvc/output/base.py", line 101, in __init__
  File "/Users/ivan/Projects/dvc/dvc/output/base.py", line 109, in _parse_path
    return self.REMOTE.path_cls(path)
  File "/Users/ivan/Projects/dvc/dvc/path_info.py", line 115, in __init__
    assert not p.query and not p.params and not p.fragment
AssertionError
```

Context: https://discordapp.com/channels/485586884165107732/485596304961962003/683485273085968424

casperdcl:
Looks like disallowing URLs containing queries happened in b83564d (#2707), i.e. `dvc>0.66.1`... Not sure why. Seems fine to me to remove the `assert not p.query` restriction.

- [x] fix `URLInfo` for http URLs (via  subclass `HTTPURLInfo(URLInfo)`)
  + [x] add support for params
  + [x] add support for queries
    * fixes #3424 `dvc import-url -v https://www.dropbox.com/?test data`
  + [x] add support for fragments
- [x] add tests

### DIFF
--- a/dvc/path_info.py
+++ b/dvc/path_info.py
@@ -112,7 +112,7 @@ class URLInfo(_BasePath):
 
     def __init__(self, url):
         p = urlparse(url)
-        assert not p.params and not p.fragment
+        assert not p.query and not p.params and not p.fragment
         assert p.password is None
 
         self.fill_parts(p.scheme, p.hostname, p.username, p.port, p.path)
--- a/dvc/path_info.py
+++ b/dvc/path_info.py
@@ -241,3 +241,47 @@ class CloudURLInfo(URLInfo):
     @property
     def path(self):
         return self._spath.lstrip("/")
+
+
+class HTTPURLInfo(URLInfo):
+    def __init__(self, url):
+        p = urlparse(url)
+        stripped = p._replace(params=None, query=None, fragment=None)
+        super().__init__(stripped.geturl())
+        self.params = p.params
+        self.query = p.query
+        self.fragment = p.fragment
+
+    @classmethod
+    def from_parts(
+        cls,
+        scheme=None,
+        host=None,
+        user=None,
+        port=None,
+        path="",
+        netloc=None,
+        params=None,
+        query=None,
+        fragment=None,
+    ):
+        assert bool(host) ^ bool(netloc)
+
+        if netloc is not None:
+            return cls(
+                "{}://{}{}{}{}{}".format(
+                    scheme,
+                    netloc,
+                    path,
+                    (";" + params) if params else "",
+                    ("?" + query) if query else "",
+                    ("#" + fragment) if fragment else "",
+                )
+            )
+
+        obj = cls.__new__(cls)
+        obj.fill_parts(scheme, host, user, port, path)
+        obj.params = params
+        obj.query = query
+        obj.fragment = fragment
+        return obj
--- a/dvc/path_info.py
+++ b/dvc/path_info.py
@@ -285,3 +285,32 @@ class HTTPURLInfo(URLInfo):
         obj.query = query
         obj.fragment = fragment
         return obj
+
+    @property
+    def _extra_parts(self):
+        return (self.params, self.query, self.fragment)
+
+    @property
+    def parts(self):
+        return self._base_parts + self._path.parts + self._extra_parts
+
+    @cached_property
+    def url(self):
+        return "{}://{}{}{}{}{}".format(
+            self.scheme,
+            self.netloc,
+            self._spath,
+            (";" + self.params) if self.params else "",
+            ("?" + self.query) if self.query else "",
+            ("#" + self.fragment) if self.fragment else "",
+        )
+
+    def __eq__(self, other):
+        if isinstance(other, (str, bytes)):
+            other = self.__class__(other)
+        return (
+            self.__class__ == other.__class__
+            and self._base_parts == other._base_parts
+            and self._path == other._path
+            and self._extra_parts == other._extra_parts
+        )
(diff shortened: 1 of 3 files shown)


Suor:
This basically breaks `URLInfo`. Many of its methods stop to work:
`url`, `__str__`, `__eq__`, `__hash__`, ... So info will be lost and they will start to behave weird when comparing or when using as dict keys etc.

casperdcl:
ok so this is still a bug then. We need to support queries because some sites need them for downloadable links (I seem to recall dropbox for example returns a html webpage if you leave out `?dl=1` - not sure if this is still the case but you get the point)

dvc/path_info.py
```diff
@@ -241,3 +241,79 @@ class CloudURLInfo(URLInfo):
...
+                    ("?" + query) if query else "",
+                    ("#" + fragment) if fragment else "",
+                )
+            )
+
+        obj = cls.__new__(cls)
+        obj.fill_parts(scheme, host, user, port, path)
+        obj.params = params
+        obj.query = query
+        obj.fragment = fragment
```
casperdcl: I didn't bother overriding `fill_parts` as it's not used publicly anywhere else

dvc/path_info.py
```diff
@@ -241,3 +241,79 @@ class CloudURLInfo(URLInfo):
...
+            other = self.__class__(other)
+        return (
+            self.__class__ == other.__class__
+            and self._base_parts == other._base_parts
+            and self._path == other._path
+            and self._extra_parts == other._extra_parts
+        )
+
+    def __hash__(self):
+        return hash(self.parts)
```
casperdcl: annoyingly not automatically inherited

casperdcl:
@shcheklein / @jorgeorpinel / @Suor ready for review

dvc/path_info.py
```diff
@@ -241,3 +241,79 @@ class CloudURLInfo(URLInfo):
     @property
     def path(self):
         return self._spath.lstrip("/")
+
+
+class HTTPURLInfo(URLInfo):
+    def __init__(self, url):
+        p = urlparse(url)
+        stripped = p._replace(params=None, query=None, fragment=None)
+        super().__init__(stripped.geturl())
```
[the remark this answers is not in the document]
casperdcl: More future-proof to use `super()` to use the parent's logic. Otherwise we'd need

```python
p = urlparse(url)
assert p.password is None
self.fill_parts(p.scheme, p.hostname, p.username, p.port, p.path)
```

and not call `super()`, which is allowed but not great practice.

dvc/path_info.py
casperdcl: Those use `from_parts` (which would this overridden method). The point is nothing else explicitly uses `fill_parts` directly so there's no need to override `fill_parts` for now.

dvc/path_info.py
casperdcl: That works for non-`@staticmethod`s? Hmm. Is there any documentation for this?

dvc/path_info.py
Suor: You keep saying don't need to override, but you override. Not sure I understand what you are up to here.

dvc/path_info.py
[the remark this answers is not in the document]
Suor: First, we can ignore all of this as long as it works. Some tech debt, but might be resolved later.

Needing to parse, restringify and parse is a kludge.

dvc/path_info.py
Suor: This is what they suggest in Python docs.

dvc/path_info.py
casperdcl: I mean technically the correct way to do things would be:

```python
def fill_parts(self, scheme, host, user, port, path, params, query, fragment):
    super().fill_parts(self, scheme, host, user, port, path)
    self.params = params
    self.query = query
    self.fragment = fragment
```

but this isn't required right now as `fill_parts` is really a private method not used anywhere else.

dvc/path_info.py
casperdcl: Ah somehow missed it in https://docs.python.org/3/reference/datamodel.html#object.__hash__

> If a class that overrides `__eq__()` needs to retain the implementation of `__hash__()` from a parent class, the interpreter must be told this explicitly by setting `__hash__ = <ParentClass>.__hash__`.

Issue #3452, opened 2020-03-06
progress: inconsistent nested format when no main bar


JIoJIaJIu:
## Overview
When storage is local (copy_file) on `dvc pull` first pbar is displayed as normal one ([self.pos = 0](https://github.com/iterative/dvc/blob/f05e6e0529ae147e82e3c392951f03c01b6b0c64/dvc/progress.py#L96)) and all others are nested

## Description
When we do remote.download for all scheme [except local](https://github.com/iterative/dvc/blob/f05e6e0529ae147e82e3c392951f03c01b6b0c64/dvc/remote/base.py#L585) there are [main bar](https://github.com/iterative/dvc/blob/f05e6e0529ae147e82e3c392951f03c01b6b0c64/dvc/remote/base.py#L619)
So only first pbar is normal, and all others are nested
```
$ dvc push grand_files
 51% grand_files/1/file-1|█████████████████████████████████████████████                                           |164M/320M [00:14<00:07,    21.1MB/s]
 38%|███▊      |grand_files/1/another-file-1                                                                       122M/320M [00:14<00:18,    11.3MB/s]
 66%|██████▋   |grand_files/1/second-file-1                                                                        212M/320M [00:14<00:08,    13.0MB/s]
 35%|███▌      |grand_files/1/third-file
```

but expects that all pbars are as 1-pbar

Relates other commands as well:
```
$ dvc import ../repo grand_files
Importing 'grand_files (../repo)' -> 'grand_files'                      
 56% grand_files.9vFFJ8yBMA9QLmwq5PL7xj.tmp/1/second-file-1|██████████████████████████████                        |178M/320M [00:03<00:06,    23.8MB/s]
 54%|█████▍    |grand_files.9vFFJ8yBMA9QLmwq5PL7xj.tmp/1/another-file-1                                            174M/320M [00:03<00:06,    23.1MB/s]
 57%|█████▋    |grand_files.9vFFJ8yBMA9QLmwq5PL7xj.tmp/1/file-1                                                    181M/320M [00:03<00:06,    23.0MB/s]
 54%|█████▍    |grand_files.9vFFJ8yBMA9QLmwq5PL7xj.tmp/1/third-file-1     
```
```
$ dvc pull -j3 grand_files
 78% ../../../../../../../tmp/tmp3zld67wsdvc-erepo/grand_files/1/another-file-1|██████████████████████████▍       |249M/320M [00:04<00:03,    21.7MB/s]
[...]

PR #3453, opened 2020-03-06
[WIP] pull: display a lot of jobs

### LINKED ISSUE #2905: pull: going completely crazy on large dataset
**Version:**

```
DVC version: 0.73.0
Python version: 3.6.8
Platform: Linux-4.15.0-1056-aws-x86_64-with-Ubuntu-18.04-bionic
Binary: False
Package: None
Filesystem type (cache directory): ('xfs', '/dev/nvme1n1')
Filesystem type (workspace): ('xfs', '/dev/nvme1n1')
```

**Reproduce:**

```
git clone git@github.com:iterative/dataset-registry-private.git
cd dataset-registry-private
dvc pull -j 100 ILSVRC.dvc
```

**Output:**

(sorry for the video in this format, was the fastest way to capture it while it was running)

https://www.dropbox.com/s/m821sgc1flk770o/dvc-pull-going-crazy.mov?dl=0

JIoJIaJIu:
## Disclaimer
I would like to change the implementation completely but shared for collaboration

## How to try
There is [a script](https://gist.github.com/JIoJIaJIu/05f3b72982373771da88f313bb5cad9f) for simulating behavior
[Before](https://asciinema.org/a/0JYsWtiSuOkQ1xVGkmNpVBKng)
[After](https://asciinema.org/a/j36cX6XYPTLBbduKs9KZJBUOY)

## TLDR
I tried an approach with limiting amount of visible pbars but would like to change the solution to the single bar that contains information about all files.
The idea was show only **N** pbars, and when one is finished - add new from the background. It allows to avoid flickering, but it has problems:
1. Pbars hierarchy (main/nested). At this moment main pbar is defined as pbar with [0 position](https://github.com/iterative/dvc/blob/f05e6e0529ae147e82e3c392951f03c01b6b0c64/dvc/progress.py#L96), but the order is not guaranteed. Can be addressed by more smart main bar definition, relates https://github.com/iterative/dvc/issues/3452
2. When files are small - there is flickering
3. the solution interacts with a lot of internal `Tqdm` stuff 

Close #2905

<hr/>

* [ ] ❗ Have you followed the guidelines in the [Contributing to DVC](https://dvc.org/doc/user-guide/contributing/core) list?

### DIFF
--- a/dvc/progress.py
+++ b/dvc/progress.py
@@ -11,8 +11,27 @@ from dvc.utils import env2bool
 logger = logging.getLogger(__name__)
 tqdm.set_lock(RLock())
 
+MAX_LINES = 20
 
-class Tqdm(tqdm):
+
+class Meta(type):
+    def __call__(cls, *args, **kwargs):
+        if not hasattr(cls, "_hidden_instances"):
+            cls._hidden_instances = set()
+        instance = cls.__new__(cls, *args, **kwargs)
+
+        if (
+            "disable" not in kwargs
+            and len(getattr(cls, "_instances", [])) > MAX_LINES
+        ):
+            kwargs["hidden"] = True
+            cls._hidden_instances.add(instance)
+
+        instance.__init__(*args, **kwargs)
+        return instance
+
+
+class Tqdm(tqdm, metaclass=Meta):
     """
     maximum-compatibility tqdm-based progressbars
     """
@@ -36,6 +55,20 @@ class Tqdm(tqdm):
         unit="B", unit_scale=True, unit_divisor=1024, miniters=1
     )
 
+    @classmethod
+    def _get_free_pos(cls, instance=None):
+        if getattr(instance, "hidden", False):
+            return None
+
+        positions = set(
+            abs(inst.pos)
+            for inst in cls._instances
+            if inst is not instance
+            and hasattr(inst, "pos")
+            and not inst.hidden
+        )
+        return min(set(range(len(positions) + 1)).difference(positions))
+
     def __init__(
         self,
         iterable=None,
@@ -47,6 +80,7 @@ class Tqdm(tqdm):
         bytes=False,  # pylint: disable=W0622
         file=None,
         total=None,
+        hidden=False,
         **kwargs
     ):
         """
@@ -59,6 +93,7 @@ class Tqdm(tqdm):
             will be determined by logging level.
             May be overridden to `True` due to non-TTY status.
             Skip override by specifying env var `DVC_IGNORE_ISATTY`.
+        hidden  : hide the bar
         kwargs  : anything accepted by `tqdm.tqdm()`
         """
         kwargs = kwargs.copy()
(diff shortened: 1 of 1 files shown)


shcheklein:
It's definitely a good first step. The main intention is too show that something is happening and probably it's fine to limit the number of `pbars` if we can somehow show the total progress as well and make sure that it's clear from the UI that number of pbars is artificially limited.

> I tried an approach with limiting amount of visible pbars but would like to change the solution to the single bar that contains information about all files.

how would it look like? do we show names, or only number of files?

can we take a look at how other tools like `lftp`, `rclone` show this? just to see if we like some ideas?

JIoJIaJIu:
@skshetry I have plans to change implementation, but would like to get your opinion about the current one. Maybe things change I will keep/improve this one

casperdcl:
I should ask - are there any opetations where a user would ever benefit from having more than about 5 threads? Surely I/O is a bottleneck for anything more. Maybe we can just set an upper limit to number of jobs?

JIoJIaJIu:
> I see what you're trying to do here, but it's not really a full fix (it's just the temporary patch I was thinking of in [#2905 (comment)](https://github.com/iterative/dvc/issues/2905#issuecomment-562516143)).
> 
> There's a more meta comment which maybe needs to be addressed first: [#2905 (comment)](https://github.com/iterative/dvc/issues/2905#issuecomment-599503500)

Hi @casperdcl , you mentioned that it's not a full fix, but how do you see it should work at the latest stage?
I see that `MAX_LINES` should be calculated dynamically based on terminal rows, what else?   
Maybe there should be completely different approach you see

>  from having more than about 5 threads?

As I got for network operations user gets benefit with multiple open connections, it would be great if you can share some info about it @efiop. Another point that it's just an interface - we provide option to point `jobs` parameter, that can be custom and is not limited to cpu count

casperdcl:
@JIoJIaJIu there are a few small bugs I can see with this PR but it's not really worth getting into if we can avoid it altogether by say setting `jobs = min(10, args.jobs)`.

Apart from that (in a general use case/if we decide there should be no upper bound on threads) there probably should be a PR on the upstream https://github.com/tqdm/tqdm repo...

Anyway for DVC purposes, the best way to go about this is to collect info about all tasks first (e.g. total number of bytes) and have them all report back to just one total bar if there are more than say 10 tasks.

@shcheklein/@efiop lemme know if actually you would prefer some per-task info printed at least after each task completes. If you do, this may be fairly easy to solve using something similar to this PR.

shcheklein:
@casperdcl there are definitely cases where 10 or even 20 jobs is not enough. Since some workloads are IO bound it might be beneficial to have 100 jobs, so I doubt that it's a good solution to limit to some upper bound.

Could we check some other tools that do parallel processing? sync tools, package managers, etc .. to see some ideas and how people solve this problem.

casperdcl:
Well I thought originally nobody wanted this approach (https://github.com/iterative/dvc/issues/2905#issuecomment-562520191) but it looks like a temporary patch is indeed best for now until jobs are combined into a single bar in other PR(s).

Rather than commenting a lot on this PR I'll open a new one.

casperdcl:
@JIoJIaJIu feel free to take over #3500

efiop:
Closing in favour or #3500

Issue #3457, opened 2020-03-08
GoogleDrive Remote


pford221:
Hi,

I'm having a difficult time setting up a GoogleDrive remote properly. Here's the situation:

### Specs:
**OS:** 
macOS Catalina 10.15.2

**Package versions**
dvc==0.87.0
PyDrive==1.3.1
PyDrive2==1.4.6

### Issue:
Set google drive remote as:
```dvc remote add -d gdpreds gdrive://<folderID>``` 
where folderID is the ID of the folder on google drive

Passed in `gdrive_client_id` and `gdrive_client_secret` and set `no_traverse` to `true` in `.dvc/config`

Ran `dvc add dvc_preds/preds_2020_03_08.csv` to version control a particular csv file in the `dvc_preds` folder of my local drive.

First time I ran `dvc_push` I went through the authentication process and the user credentials were stored in `.dvc/tmp/gdrive-user-credentials.json`

And I see something like this and it hangs for a couple minutes:
` 0% Querying cache in gdrive://<folderID>|`

Until I finally get this 403 HTTPError.
```
ERROR: unexpected error - <HttpError 403 when requesting https://www.googleapis.com/drive/v2/files?q=%28%271O6jLWKwS2pxhOOCFUlUd61fD42PfFuUy%27+in+parents%29+and+trashed%3Dfalse+and+title%3D%27bf%27&maxResults=1&corpora=drive&supportsAllDrives=true&includeItemsFromAllDrives=true&alt=json returned "The driveId parameter must be specified if and only if corpora is set to drive.">
```
I have tried to set the environment variable for `GOOGLE_AAPLICATION_CREDENTIAL` as specified [here in 3005](https://github.com/iterative/dvc/issues/3005). 

Thanks in advance for any help.

shcheklein:
@pford221 was able to reproduce this. Will try to fix and release a new version asap.

Meanwhile, you could use this notation:

`gdrive://root/<path-to-your-directory>`

it should be working fine.

shcheklein:
@pford221 https://github.com/iterative/dvc/pull/3458 should fix this. We'll try to release a fix in a day or two. Thanks for reporting this.

PR #3458, opened 2020-03-09
gdrive: share drive vs just folder id when we set corpora param

shcheklein:
Fixes https://github.com/iterative/dvc/issues/3457

----------------


MaxRis:
it is not directly related to this PR, but can we pin PyDrive2 to the specific version to avoid accidental incompatibility? `gdrive = ["pydrive2>=1.4.5"]`

shcheklein:
@MaxRis @efiop 

> but can we pin PyDrive2 to the specific version to avoid accidental incompatibility?

ideally we should be managing PyDrive2 versioning in a way that does not break compatibility. In the DVC dependencies we might "pin" a major version?

Issue #2671, opened 2019-10-25
Optimize stage collection for large repos


danfischetti:
I posted this issue #2666 yesterday which was closed but the patch did not solve my underlying issue. The DAG is constructed using the stages from the `repo.collect_stages()` method. This method is what is very slow for me with a large repo. So I would like to ask again if you could make the collection of every stage before any operation optional so it doesn't take hours to `dvc add` a single file with no dependencies.

efiop:
@danfischetti Could you please post a cProfile log from your specific use case so we can see more clearly what is slowing you down? :slightly_smiling_face:

efiop:
@danfischetti Also, did the change from yesterday make any difference for you?

danfischetti:
I just updated to 0.65.0 and I don't see any difference.

```
In [7]: cProfile.run("repo.collect_stages()")                                                                                                                                                     [389/1983]
^C         329107531 function calls (317924667 primitive calls) in 113.090 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    33595    0.019    0.000    0.048    0.000 <frozen importlib._bootstrap>:997(_handle_fromlist)
        1    0.000    0.000  113.095  113.095 <string>:1(<module>)
    99792    0.029    0.000    0.051    0.000 <string>:12(__new__)
     5764    0.022    0.000    0.980    0.000 __init__.py:100(loadd_from)
    17293    0.004    0.000    0.004    0.000 __init__.py:110(tree)
215138/5764    0.225    0.000    0.447    0.000 __init__.py:319(convert_to_unicode)
     3389    0.008    0.000    0.217    0.000 __init__.py:350(dvc_walk)
     3388    0.007    0.000   49.910    0.015 __init__.py:371(_filter_out_dirs)
     3444    3.412    0.001   49.902    0.014 __init__.py:373(filter_dirs)
        1    0.042    0.042  113.095  113.095 __init__.py:382(collect_stages)
     5764    0.015    0.000    0.316    0.000 __init__.py:453(relpath)
     8151    0.045    0.000    1.234    0.000 __init__.py:52(_get)
     5764    0.024    0.000    0.856    0.000 __init__.py:60(_get)
     5764    0.014    0.000    1.305    0.000 __init__.py:71(loadd_from)
    89676    0.101    0.000    0.221    0.000 _collections_abc.py:879(__iter__)
[...]

PR #3459, opened 2020-03-09
dvc: optimize all target specific commands to not build graph

Suor:
Should help with #2671 for commands like `push/pull/checkout` when particular stage file is specified directly.

### DIFF
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -179,27 +179,33 @@ class Repo(object):
         import networkx as nx
         from dvc.stage import Stage
 
-        G = graph or self.graph
-
         if not target:
-            return list(G)
+            return list(graph) if graph else self.stages
 
         target = os.path.abspath(target)
 
         if recursive and os.path.isdir(target):
-            stages = nx.dfs_postorder_nodes(G)
+            stages = nx.dfs_postorder_nodes(graph or self.graph)
             return [stage for stage in stages if path_isin(stage.path, target)]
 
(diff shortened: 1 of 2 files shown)


dvc/repo/__init__.py
```diff
@@ -179,28 +179,34 @@ def collect(self, target, with_deps=False, recursive=False, graph=None):
         import networkx as nx
         from dvc.stage import Stage
 
-        G = graph or self.graph
-
         if not target:
-            return list(G)
+            return list(graph) if graph else self.stages
```
skshetry:
```suggestion
            return list(graph or self.stages)
```

tests/unit/repo/test_repo.py
```diff
@@ -66,3 +67,18 @@ def test_locked(mocker):
...
+
+
+def test_collect_optimization(tmp_dir, dvc, mocker):
+    (stage,) = tmp_dir.dvc_gen("foo", "foo text")
+
+    # Forget cached stages and graph and error out on collection
+    dvc._reset()
+    mocker.patch(
+        "dvc.repo.Repo.stages",
+        property(raiser(Exception("Should not collect"))),
```
pared: +1

efiop:
@Suor Please address review comments.

dvc/repo/__init__.py
Suor: No need to make a shallow copy off stages.

Suor:
Should be good to go

PR #3468, opened 2020-03-10
[WIP] cmd: diff: don't load dir cache for diff

### LINKED ISSUE #3392: diff: does not behave properly when cache is not available
## Version

```
DVC version: 0.86.5+f67314.mod
Python version: 3.7.6
Binary: False
Package: None
```

## Reproduce

```
git clone git@github.com:iterative/example-get-started.git
cd example-get-started
dvc pull
dvc diff baseline-experiment bigrams-experiment
```

## Output

```
Added:
    data/features/test.pkl
    data/features/train.pkl

Modified:
    auc.metric
    data/features/
    model.pkl

files summary: 2 added, 0 deleted, 2 modified
```

which is wrong. Files `data/features/test.pkl` and `data/features/train.pkl` actually exist in both experiments. The only difference is that one of them cached, one is not.

pared:

Fixes #3392

### DIFF
--- a/dvc/repo/diff.py
+++ b/dvc/repo/diff.py
@@ -22,29 +22,25 @@ def diff(self, a_rev="HEAD", b_rev=None):
         A dictionary of checksums addressed by relpaths collected from
         the current tree outputs.
 
-        Unpack directories to include their entries
-
         To help distinguish between a directory and a file output,
         the former one will come with a trailing slash in the path:
 
             directory: "data/"
             file:      "data"
         """
-        result = {}
 
-        for stage in self.stages:
-            for output in stage.outs:
-                if not output.is_dir_checksum:
-                    result.update({str(output): output.checksum})
-                    continue
+        def _to_path(output):
+            return (
+                str(output)
+                if not output.is_dir_checksum
+                else os.path.join(str(output), "")
+            )
 
-                result.update({os.path.join(str(output), ""): output.checksum})
-
-                for entry in output.dir_cache:
-                    path = str(output.path_info / entry["relpath"])
-                    result.update({path: entry["md5"]})
-
-        return result
+        return {
+            _to_path(output): output.checksum
+            for stage in self.stages
+            for output in stage.outs
+        }
 
     working_tree = self.tree
     a_tree = self.scm.get_tree(a_rev)
--- a/tests/func/test_diff.py
+++ b/tests/func/test_diff.py
@@ -3,6 +3,8 @@ import os
 import pytest
 import shutil
 
+from funcy import first
+
 from dvc.compat import fspath
 from dvc.exceptions import DvcException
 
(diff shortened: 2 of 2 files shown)


efiop:
Sorry for a late question, but does this mean that we lose support for dirs from now on? The issue was that when one version has .dir cache file and the other one doesn't, then the output is weird, right?

pared:
@efiop Yes, in a way. After this change, `dir` out will have the same behavior as `non-dir` one:
it will be displayed as `added/deleted/modified` as a whole.

pared:
After discussing with @efiop, its probably the best to apply a hybrid approach:
when the cache is available, list differences, if not, treat dir as if they were not dirs.

shcheklein:
@pared @efiop I have some doubts about this - hybrid approach means that users will see different results in different situations, and will be happening silently for them. 

Or am I missing something about this hybrid approach? Can we highlight those dirs somehow, for example?

cc @dmpetrov .

dmpetrov:
> @pared @efiop I have some doubts about this - hybrid approach means that users will see different results in different situations, and will be happening silently for them.

Totally agree with @shcheklein It will lead to unstable diffs.

I'd suggest ignoring all dirs content (option 3 from #3392). This option has additional advantages:
1. No need in the cache and no need in data-remote credentials which is important for some CI/CD scenarios.
2. Easy to implement.

efiop:
Ok, I thought that dir cache support was desirable, but if it is not, then indeed let's just drop it. Thanks for clarifying. :pray: 

@pared Sorry for this back and forth, please revert the hybrid changes.

shcheklein:
@efiop my 2cs - it's desirable, but it should be done along with support for `dvc list` and other commands that require it (what about `gc` for example?).

efiop:
> it's desirable, but it should be done along with support for dvc list and other commands that require it 

In that case, I'm not sure why can't we start with diff here, since it is on the surface already. And then proceed with `dvc list` and others.

> what about gc for example?

Gc will attempt to pull .dir to analyze it, unlike diff/list/etc.

shcheklein:
> In that case, I'm not sure why can't we start with diff here, since it is on the surface already

I guess there is some misalignment on what do we think can be considered as a start? There were a few different suggestions that @pared brought - from very simple (never do for dirs) to the actual implementation (try and pull dir file). I personally fine to start with either. The only one I had concerns about is the hybrid one - when we silently behave in a different way depending on cache availability.

dmpetrov:
> I'm not sure why can't we start with diff here

First, because here a hybrid approach is proposed which does not give a reliable result.
Second, diff without cache is a required feature #2998 (closed due to this issue). It would be great to have asap because ci/cd needs it.

A good strategy might be - implement quickly the simple strategy (and make sure it does not require cache) and then prioritize the full solution with data dir check.

efiop:
Ok, thanks for clarifying. Sure, let's start with the simplest one first to unblock ci/cd. :+1:

PR #3471, opened 2020-03-11
dvc: update cmd output strings to match docs

jorgeorpinel:
to match iterative/dvc.org/pull/1014
and now iterative/dvc.org/pull/1040 too


dvc/command/metrics.py
```diff
@@ -245,7 +244,7 @@ def add_parser(subparsers, parent_parser):
     metrics_add_parser.add_argument("path", help="Path to a metric file.")
     metrics_add_parser.set_defaults(func=CmdMetricsAdd)
 
-    METRICS_MODIFY_HELP = "Modify metric file options."
+    METRICS_MODIFY_HELP = "Modify metric file values."
```
shcheklein: it should be options.

`modifying values` sounds like we actually modify metric values.

it is about setting up metric file settings (type, path, etc), not modifying metric values

jorgeorpinel: I see. I'll update this... ⏳  but still need the PR for other strings.

jorgeorpinel: What about "default formatting"? See latest version here and in iterative/dvc.org/pull/1014

shcheklein: `default formatting options`  is still better? )

jorgeorpinel: I'm trying to avoid "options" here since there are also command options and config options and it gets confusing some times in the explanations.

shcheklein: Yep, I got the idea. I think it's almost impossible to avoid unfortunately in general (even if it works in this case). At least I was not able to find a good solution - may be even primarily I never thought that it's a really big deal - usually it's clear from the context. And can add more words when it's not clear from the context I think.

dvc/command/ls.py
```diff
@@ -39,14 +39,12 @@ def add_parser(subparsers, parent_parser):
...
-        "C:\\\\path\\to\\file\n"
-        "C:\\\\path\\to\\directory\n"
-        "https://github.com/path/to/repo\n"
-        "git@github.com:path/to/repo.git\n",
+        "url", help="Location of DVC or Git repository to download from"
+    )
+    list_parser.add_argument(
+        "target",
+        nargs="?",
+        help="Path to directory within the repository to list outputs for",
```
jorgeorpinel: @JIoJIaJIu I moved this up bc it seems more important than the following args (options). It won't change the command behavior or help output right?

dvc/external_repo.py
```diff
@@ -182,6 +182,7 @@ def pull_to(self, path, to_info):
 
     @contextmanager
     def open_by_relpath(self, path, mode="r", encoding=None, **kwargs):
+        """Opens a specified resource as a file descriptor"""
```
jorgeorpinel: Note: Added this docstring since external_repo seems like an important module. Kind of like the Repo class. They're very similar from what I can see.

dvc/command/metrics.py
jorgeorpinel: I think I managed to avoid the term everywhere actually 🙂 assuming we let this one through. Yes it's not always a big deal but in this case and others I changed I thought it was confusing since all the "option" concepts are used in the same doc or even section of the doc sometimes.

dvc/command/data_sync.py
```diff
@@ -98,7 +98,7 @@ def add_parser(subparsers, _parent_parser):
     from dvc.command.status import CmdDataStatus
 
     # Pull
-    PULL_HELP = "Pull data files from a DVC remote storage."
+    PULL_HELP = "Download tracked files or directories from remote storage."
```
efiop: Not sure `pull` and `fetch` help messages are different enough now. But it is definitely not worse.

dvc/command/ls/__init__.py
```diff
@@ -54,14 +54,12 @@ def add_parser(subparsers, parent_parser):
         formatter_class=argparse.RawTextHelpFormatter,
     )
     list_parser.add_argument(
-        "url",
-        help="Supported urls:\n"
-        "/path/to/file\n"
```
efiop: Not sure why are we getting rid of these useful examples.

dvc/command/metrics.py
```diff
@@ -159,7 +159,7 @@ def run(self):
 
 
 def add_parser(subparsers, parent_parser):
-    METRICS_HELP = "Commands to add, manage, collect and display metrics."
+    METRICS_HELP = "Commands to add, manage, collect, and display metrics."
```
efiop: Are you sure comma is needed here? `and` is not really joining two independent clauses.

dvc/command/metrics.py
```diff
@@ -261,7 +261,7 @@ def add_parser(subparsers, parent_parser):
     metrics_modify_parser.add_argument("path", help="Path to a metric file.")
     metrics_modify_parser.set_defaults(func=CmdMetricsModify)
 
-    METRICS_REMOVE_HELP = "Remove files's metric tag."
+    METRICS_REMOVE_HELP = "Stop tracking a metric file."
```
efiop: This might be confusing. We are not stopping tracking the file but rather un-marking it as a metric.

dvc/command/pipeline.py
```diff
@@ -236,7 +236,7 @@ def add_parser(subparsers, parent_parser):
     )
     pipeline_show_parser.set_defaults(func=CmdPipelineShow)
 
-    PIPELINE_LIST_HELP = "List pipelines."
+    PIPELINE_LIST_HELP = "List connected groups of stages (pipelines)."
```
efiop: Why are we making this clarification here, but not in other places? `pipeline` is self descriptive enough, I don't think it needs a "connected groups of stages"

dvc/command/remote.py
```diff
@@ -116,7 +116,7 @@ def add_parser(subparsers, parent_parser):
 
     fix_subparsers(remote_subparsers)
 
-    REMOTE_ADD_HELP = "Add remote."
+    REMOTE_ADD_HELP = "Add a new data remote."
```
efiop: Are we sure we need `data` here? Won't it be confusing in regards to `models` and such?

dvc/command/remove.py
```diff
@@ -42,7 +42,7 @@ def run(self):
 
 
 def add_parser(subparsers, parent_parser):
-    REMOVE_HELP = "Remove DVC-file outputs."
+    REMOVE_HELP = "Remove data files or directories tracked by DVC."
```
efiop: This is not only about data, but about stages/models/etc too.

dvc/command/repro.py
```diff
@@ -57,7 +57,8 @@ def run(self):
 
 
 def add_parser(subparsers, parent_parser):
-    REPRO_HELP = "Check for changes and reproduce stages and dependencies."
+    REPRO_HELP = "Reproduce complete or partial pipelines"
+    " by executing their stages."
```
efiop: maybe `running`?

dvc/command/root.py
```diff
@@ -17,7 +17,7 @@ def run(self):
 
 
 def add_parser(subparsers, parent_parser):
-    ROOT_HELP = "Relative path to the repository's directory."
+    ROOT_HELP = "Return the relative path to the DVC project."
```
efiop:
```suggestion
    ROOT_HELP = "Return the relative path to the DVC project's root."
```

dvc/external_repo.py
efiop: We don't return file descriptors (low-level concept), but rather file objects. I see that you got confused by `fd` in the code, it should really be `fobj`, of course.

efiop:
Sorry for the delay.

dvc/command/ls.py
JIoJIaJIu: Hello @jorgeorpinel , sorry for the late reply
Yes, it's ok to move up

dvc/external_repo.py
jorgeorpinel: Right, sorry just a typo. `fd` is pretty a common var name for `with open()` so my bad. Updated!

dvc/command/data_sync.py
jorgeorpinel: Good point. I'm just copying from https://dvc.org/doc/command-reference/fetch and https://dvc.org/doc/command-reference/pull here. This whole PR is about matching docs which I think is desirable but if not, please lmk!

dvc/command/ls/__init__.py
jorgeorpinel: Yeah for consistency with get and import and so the output here is shorter. We could go the opposite way and list all the examples in all these commands if that's preferable. Or we could add a link to the cmd ref in each one. Please lmk, happy to change it.

dvc/command/metrics.py
jorgeorpinel: There's never absolute correctness with these things, but yes, in general everyone recommends to add a comma before "and/or" in the last element of a list. See https://lifehacker.com/always-put-a-comma-before-and-in-a-list-no-matter-wh-1793376109

dvc/command/metrics.py
jorgeorpinel: True. I'm updating this one and `metrics add`, as well as the cmd refs. See f262f483 and iterative/dvc.org/pull/1072. Thanks!

dvc/command/pipeline.py
jorgeorpinel: Good question. Again, this is to match docs, https://dvc.org/doc/command-reference/pipeline/list specifically.

> but not in other places

The only other place really is pipeline show which reads "Show stages in a pipeline." This one also mentions both terms "stage" and "pipeline" so it's similarly clear, I think.

dvc/command/remote.py
jorgeorpinel: Sorry to repeat the same answer haha but just matching docs here (https://dvc.org/doc/command-reference/remote/add).

For consistency, we use the term "data remote" in remote subcommands cmd refs (not in other docs though). This is kind of arbitrary, I agree on that. Would you prefer a different standard? Happy to update everywhere.

dvc/command/remove.py
jorgeorpinel: Hmmm OK changed to "Remove DVC-tracked files or directories." both here and in the cmd ref now. Better? The part about removing entire stage files is explained in the cmd ref desc.

dvc/command/repro.py
jorgeorpinel: Since we have ` dvc run` command, we use the term "execute" a lot in the repro cmd ref (https://dvc.org/doc/command-reference/repro).

dvc/command/root.py
jorgeorpinel: Hmmm I thought it's obvious but I guess you're right, since you may be using `dvc root` from inside the project (probably even the main/only use case!) Updated, thanks.

jorgeorpinel:
Np

efiop:
Ok, merging for now. Will try to keep an eye on the docs PRs.

PR #3472, opened 2020-03-11
dvc: use protected mode by default

### LINKED ISSUE #2041: config: should `cache.protected` be hidden and automatic?
Could/Should the `cache.protected` mode be hidden to the user, enabled automatically for `cache.type` `hardlink` or `symlink`, and disabled otherwise?

Just wondering what's the value of having the user manually configure something that is always required for those cache link types.

### LINKED ISSUE #3261: Pull extrememly slow on ~400GB of data with hot DVC cache

efiop:
Fixes #3261
Fixes #2041

https://github.com/iterative/dvc.org/pull/1058

### DIFF
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -97,7 +97,7 @@ REMOTE_COMMON = {
 }
 LOCAL_COMMON = {
     "type": supported_cache_type,
-    Optional("protected", default=False): Bool,
+    Optional("protected", default=False): Bool,  # obsoleted
     "shared": All(Lower, Choices("group")),
     Optional("slow_link_warning", default=True): Bool,
 }
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -83,6 +83,9 @@ class RemoteBASE(object):
     DEFAULT_NO_TRAVERSE = True
     DEFAULT_VERIFY = False
 
+    CACHE_MODE = None
+    SHARED_MODE_MAP = {None: (None, None), "group": (None, None)}
+
     state = StateNoop()
 
     def __init__(self, repo, config):
@@ -90,12 +93,14 @@ class RemoteBASE(object):
 
         self._check_requires(config)
 
+        shared = config.get("shared")
+        self._file_mode, self._dir_mode = self.SHARED_MODE_MAP[shared]
+
         self.checksum_jobs = (
             config.get("checksum_jobs")
             or (self.repo and self.repo.config["core"].get("checksum_jobs"))
             or self.CHECKSUM_JOBS
         )
-        self.protected = False
         self.no_traverse = config.get("no_traverse", self.DEFAULT_NO_TRAVERSE)
         self.verify = config.get("verify", self.DEFAULT_VERIFY)
         self._dir_info = {}
@@ -221,7 +226,7 @@ class RemoteBASE(object):
         new_info = self.cache.checksum_to_path_info(checksum)
         if self.cache.changed_cache_file(checksum):
             self.cache.makedirs(new_info.parent)
-            self.cache.move(tmp_info, new_info)
+            self.cache.move(tmp_info, new_info, mode=self.CACHE_MODE)
 
         self.state.save(path_info, checksum)
         self.state.save(new_info, checksum)
@@ -409,15 +414,8 @@ class RemoteBASE(object):
 
         link_method(from_info, to_info)
 
-        if self.protected:
-            self.protect(to_info)
-
         logger.debug(
-            "Created %s'%s': %s -> %s",
-            "protected " if self.protected else "",
-            self.cache_types[0],
-            from_info,
-            to_info,
+            "Created '%s': %s -> %s", self.cache_types[0], from_info, to_info,
         )
 
     def _save_file(self, path_info, checksum, save_link=True):
@@ -425,14 +423,11 @@ class RemoteBASE(object):
 
         cache_info = self.checksum_to_path_info(checksum)
         if self.changed_cache(checksum):
-            self.move(path_info, cache_info)
+            self.move(path_info, cache_info, mode=self.CACHE_MODE)
             self.link(cache_info, path_info)
         elif self.iscopy(path_info) and self._cache_is_copy(path_info):
             # Default relink procedure involves unneeded copy
-            if self.protected:
-                self.protect(path_info)
-            else:
-                self.unprotect(path_info)
+            self.unprotect(path_info)
         else:
             self.remove(path_info)
             self.link(cache_info, path_info)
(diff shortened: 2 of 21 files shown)


dvc/remote/base.py
```diff
@@ -82,19 +82,24 @@ class RemoteBASE(object):
     DEFAULT_NO_TRAVERSE = True
     DEFAULT_VERIFY = False
 
+    CACHE_MODE = None
```
efiop: Generalising it because SSH could use the same optimization a bit later.

dvc/remote/local.py
```diff
@@ -32,19 +32,11 @@ class RemoteLOCAL(RemoteBASE):
 
     DEFAULT_CACHE_TYPES = ["reflink", "copy"]
 
+    CACHE_MODE = 0o444
```
efiop: Continuing using octal notation since it is much more readable than stat constants. Need to double check the windows though, since it uses some weird perms.

dvc/remote/local.py
```diff
@@ -32,19 +32,11 @@ class RemoteLOCAL(RemoteBASE):
...
     SHARED_MODE_MAP = {None: (0o644, 0o755), "group": (0o664, 0o775)}
 
     def __init__(self, repo, config):
         super().__init__(repo, config)
-        self.protected = config.get("protected", False)
-
-        shared = config.get("shared")
-        self._file_mode, self._dir_mode = self.SHARED_MODE_MAP[shared]
-
-        if self.protected:
```
efiop: Previously we kinda mixed up cache and link protection, the new patch puts the emphasis on cache protection, which is what we really want and should be much less confusing.

dvc/remote/local.py
```diff
@@ -136,23 +128,25 @@ def remove(self, path_info):
...
 
-        if self.isfile(from_info):
-            mode = self._file_mode
-        else:
-            mode = self._dir_mode
+        if mode is None:
+            if self.isfile(from_info):
+                mode = self._file_mode
+            else:
+                mode = self._dir_mode
```
efiop: Not the nicest thing, but it was there from a "shared" PR, need to check if I can possibly make it more pretty by passing mode explicitly where `move` is used.

dvc/remote/local.py
```diff
@@ -196,9 +190,11 @@ def hardlink(self, from_info, to_info):
     def is_hardlink(path_info):
         return System.is_hardlink(path_info)
 
-    @staticmethod
-    def reflink(from_info, to_info):
-        System.reflink(from_info, to_info)
+    def reflink(self, from_info, to_info):
+        tmp_info = to_info.parent / tmp_fname(to_info.name)
+        System.reflink(from_info, tmp_info)
```
efiop: Yes, you can change mode on reflink without affecting the source. This is because reflinks have their own inodes, so from fs perspective they are separate files that have their own perms.

dvc/config.py
```diff
@@ -97,7 +97,7 @@ class RelPath(str):
 }
 LOCAL_COMMON = {
     "type": supported_cache_type,
-    Optional("protected", default=False): Bool,
+    Optional("protected", default=False): Bool,  # obsoleted
```
efiop: So this one is effectively defunct from now on. Just keeping it in the SCHEMA to not break on older configs.

efiop:
Windows is failing, will check it out tomorrow.

dvc/remote/local.py
efiop: Windows is fine.

dvc/remote/local.py
```diff
@@ -142,23 +134,25 @@ def remove(self, path_info):
...
+            else:
+                mode = self._dir_mode
 
         move(from_info, to_info, mode=mode)
 
     def copy(self, from_info, to_info):
         tmp_info = to_info.parent / tmp_fname(to_info.name)
         try:
             System.copy(from_info, tmp_info)
+            os.chmod(fspath_py35(tmp_info), self._file_mode)
```
skshetry: Is `copy` only used on files? The implementation looked very generic to me.

dvc/remote/local.py
skshetry: @efiop, Can you mention this on a code itself as a comment?

dvc/config.py
skshetry: This does not respect explicit `protected = False`, does it?

dvc/remote/local.py
skshetry: Also, can this not be done without link to `temp`

dvc/config.py
efiop: @skshetry It doesn't, as there is no clear use case for it.

dvc/remote/local.py
efiop: > @efiop, Can you mention this on a code itself as a comment?

Sure! Done.

> Also, can this not be done without link to temp?

Temp file is needed to make it atomic, so the target file appears already with correct permissions, so there is no chance of something trying to use it while it is in read-only mode.

skshetry: Got it, thanks.

shcheklein:
Please, update docs if needed after this change.

PR #3478, opened 2020-03-12
gdrive: config option to move remote file to trash only

### LINKED ISSUE #3477: gdrive: remote config value to Trash remote files instead of permanently deleting
At the moment `dvc gc` always triggers permanently deletion of files on remote. In case of shared drives the user might have insufficient permissions to delete files permanently; thus, we're looking to introduce `Trash` only new config option.

maxhora:
Fixes #3477

### DIFF
--- a/dvc/remote/gdrive.py
+++ b/dvc/remote/gdrive.py
@@ -39,12 +39,14 @@ class GDriveMissedCredentialKeyError(DvcException):
 
 
 def gdrive_retry(func):
-    retry_re = re.compile(r"HttpError (403|500|502|503|504)")
-
     def should_retry(exc):
         from pydrive2.files import ApiRequestError
 
-        return isinstance(exc, ApiRequestError) and retry_re.search(str(exc))
+        if not isinstance(exc, ApiRequestError):
+            return False
+
+        retry_codes = [403, 500, 502, 503, 504]
+        return exc.error.get("code", 0) in retry_codes
 
     # 15 tries, start at 0.5s, multiply by golden ratio, cap at 20s
     return retry(
@@ -52,6 +54,19 @@ def gdrive_retry(func):
         timeout=lambda a: min(0.5 * 1.618 ** a, 20),
         filter_errors=should_retry,
     )(func)
+
+
+def _location(exc):
+    from pydrive2.files import ApiRequestError
+
+    assert isinstance(exc, ApiRequestError)
+
+    # https://cloud.google.com/storage/docs/json_api/v1/status-codes#errorformat
+    return (
+        exc.error["errors"][0].get("location", "")
+        if exc.error.get("errors", [])
+        else ""
+    )
 
 
 class GDriveURLInfo(CloudURLInfo):
@@ -95,6 +110,7 @@ class RemoteGDrive(RemoteBASE):
             )
 
         self._bucket = self.path_info.bucket
+        self._trash_only = config.get("gdrive_trash_only")
         self._use_service_account = config.get("gdrive_use_service_account")
         self._service_account_email = config.get(
             "gdrive_service_account_email"
@@ -118,8 +134,6 @@ class RemoteGDrive(RemoteBASE):
                 ),
             )
         )
-
-        self._list_params = None
 
     def _validate_config(self):
         # Validate Service Account configuration
@@ -326,10 +340,35 @@ class RemoteGDrive(RemoteBASE):
 
     @gdrive_retry
     def _delete_remote_file(self, remote_id):
+        from pydrive2.files import ApiRequestError
+
         param = {"id": remote_id}
         # it does not create a file on the remote
         item = self.drive.CreateFile(param)
-        item.Delete()
+
+        try:
+            item.Trash() if self._trash_only else item.Delete()
+        except ApiRequestError as exc:
+            http_error_code = exc.error.get("code", 0)
+            if (
+                http_error_code == 403
+                and self.list_params["corpora"] == "drive"
+                and _location(exc) == "file.permissions"
+            ):
+                raise DvcException(
+                    "Insufficient permissions to {}. You should have {} "
+                    "access level for the used shared drive. More details "
+                    "at {}.".format(
+                        "move the file into Trash"
+                        if self._trash_only
+                        else "permanently delete the file",
+                        "Manager or Content Manager"
+                        if self._trash_only
+                        else "Manager",
+                        "https://support.google.com/a/answer/7337554",
+                    )
+                ) from exc
+            raise
 
     @gdrive_retry
     def _get_remote_item(self, name, parents_ids):
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,7 @@ install_requires = [
 # Extra dependencies for remote integrations
 
 gs = ["google-cloud-storage==1.19.0"]
-gdrive = ["pydrive2>=1.4.6"]
+gdrive = ["pydrive2>=1.4.8"]
 s3 = ["boto3>=1.9.201"]
 azure = ["azure-storage-blob==2.1.0"]
 oss = ["oss2==2.6.1"]
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -179,6 +179,7 @@ SCHEMA = {
                     "gdrive_service_account_email": str,
                     "gdrive_service_account_user_email": str,
                     "gdrive_service_account_p12_file_path": str,
+                    Optional("gdrive_trash_only", default=False): Bool,
                     **REMOTE_COMMON,
                 },
                 "http": {**HTTP_COMMON, **REMOTE_COMMON},


dvc/remote/gdrive.py
```diff
@@ -325,10 +323,37 @@ def _create_remote_dir(self, parent_id, title):
...
+
+        try:
+            item.Trash() if self._gdrive_trash_only else item.Delete()
+        except ApiRequestError as exc:
+            if self.list_params["corpora"] != "drive" or not exc.args:
+                raise
+            http_error = exc.args[0]
+            if (
+                http_error.resp.status == 403
+                and "Insufficient permissions for this file".encode("utf8")
```
shcheklein: I thought that we agreed on changing this on the PyDrive2 end? To fix the issue with retries, and avoid using string parsing.

MaxRis: `403` code seems is not an unique identifier to detect insufficient permissions error; it is required to check additionally http_error.content which has JSON string in bytes format with `message` key equal to "Insufficient permissions for this file".
Since we discussed that it doesn't make sense to introduce separate exception class in PyDrive2 along with ApiRequestError, it seems a simple alternative way just to handle this on DVC side by alanyzing 2 unique params (used string search might be replaced into more complex converting from bytes array to a map and checking key value).
Please let me know if I'm still missing here a better way to handle this, thanks!

shcheklein: as we discussed, Google Drive returns `reason` for every error it sends in the JSON structure - https://developers.google.com/drive/api/v3/handle-errors . We should be parsing this JSON, extracting status code, reason, all other relevant information and include it into ApiRequestError.

If the JSON you mentioned is exactly the one described in the doc and we already have it, we can parse it on our end.

dvc/remote/gdrive.py
```diff
@@ -41,10 +41,43 @@ def __init__(self, path):
...
+        return True
+
     def should_retry(exc):
         from pydrive2.files import ApiRequestError
 
-        return isinstance(exc, ApiRequestError) and retry_re.search(str(exc))
+        if not isinstance(exc, ApiRequestError):
+            return False
+
+        return retry_error(exc) and retry_re.search(str(exc))
```
shcheklein: can we completely get rid of `retry_re.search(str(exc))`?

dvc/remote/gdrive.py
```diff
@@ -41,10 +41,43 @@ def __init__(self, path):
 def gdrive_retry(func):
     retry_re = re.compile(r"HttpError (403|500|502|503|504)")
 
+    def retry_error(exc):
+        import json
+
+        if RemoteGDrive.corpora != "drive" or not exc.args:
```
shcheklein: are you sure we need this check? If we are getting rid of regexp then we need to get JSON, check HTTP code and make sure that it's not `file.permissions` + `forbidden`

dvc/remote/gdrive.py
```diff
@@ -79,6 +112,12 @@ class RemoteGDrive(RemoteBASE):
     GDRIVE_CREDENTIALS_DATA = "GDRIVE_CREDENTIALS_DATA"
     DEFAULT_USER_CREDENTIALS_FILE = "gdrive-user-credentials.json"
 
+    # Indicates usage of shared or default remote drive
+    corpora = "default"
```
shcheklein: looks like a hack, and we don't actually need this I believe

dvc/remote/gdrive.py
```diff
@@ -41,10 +41,43 @@ def __init__(self, path):
...
+            return True
+        http_error = exc.args[0]
+        content = json.loads(http_error.content)
+        if content and "error" not in content:
+            return True
+        err = content["error"]
+        if (
+            err["code"] == 403
+            and len(err["errors"])
+            and err["errors"][0].get("location", "") == "file.permissions"
```
shcheklein: it's better to here from what we consider _retriable_ exceptions - can we identify what error codes, locations, reasons it sends for limits, for example?

dvc/remote/gdrive.py
```diff
@@ -41,10 +41,43 @@ def __init__(self, path):
...
+        ):
+            raise DvcException(
```
shcheklein: we should catch the error in GC implementation method and convert it there, not in retry logic

dvc/remote/gdrive.py
MaxRis: Done, it checks numeric code now

dvc/remote/gdrive.py
MaxRis: Since the response with permissions error we are trying to identify is not documented, I'm not sure if GDrive API has some others undocumented cases. `forbidden` reason and `403` error code seems can be generated in others cases.

dvc/remote/gdrive.py
MaxRis: Removed, thanks

dvc/remote/gdrive.py
MaxRis: will try to check available API docs

dvc/remote/gdrive.py
MaxRis: changed, thanks

dvc/remote/gdrive.py
```diff
@@ -38,13 +38,34 @@ def __init__(self, path):
...
+    from pydrive2.files import ApiRequestError
+
+    assert isinstance(exc, ApiRequestError)
+
+    if hasattr(exc, "errors"):
+        return
+
+    exc.errors = None
 
+    if exc.args:
```
shcheklein: is it always HttpError with JSON? should we check the type?

dvc/remote/gdrive.py
```diff
@@ -38,13 +38,34 @@ def __init__(self, path):
         )
 
 
-def gdrive_retry(func):
-    retry_re = re.compile(r"HttpError (403|500|502|503|504)")
+def append_error_codes(exc):
```
shcheklein: append_error_codes -> append_error_details? (it's not about code only)

dvc/remote/gdrive.py
```diff
@@ -38,13 +38,34 @@ def __init__(self, path):
...
+    import json
+    from pydrive2.files import ApiRequestError
+
+    assert isinstance(exc, ApiRequestError)
+
+    if hasattr(exc, "errors"):
+        return
+
+    exc.errors = None
```
shcheklein: make it empty? so that we don't have to check for None below

dvc/remote/gdrive.py
```diff
@@ -38,13 +38,34 @@ def __init__(self, path):
...
     def should_retry(exc):
         from pydrive2.files import ApiRequestError
 
-        return isinstance(exc, ApiRequestError) and retry_re.search(str(exc))
+        if not isinstance(exc, ApiRequestError):
+            return False
+
+        append_error_codes(exc)
+        retry_codes = [403, 500, 502, 503, 504]
+        return exc.errors["code"] in retry_codes
```
shcheklein: exc.errors can be None, let's make it empty list and use `get("code", None)` ?

dvc/remote/gdrive.py
```diff
@@ -324,10 +344,38 @@ def _create_remote_dir(self, parent_id, title):
...
         param = {"id": remote_id}
         # it does not create a file on the remote
         item = self.drive.CreateFile(param)
-        item.Delete()
+
+        try:
+            item.Trash() if self._trash_only else item.Delete()
+        except ApiRequestError as exc:
+            append_error_codes(exc)
+            if not exc.errors and self.list_params["corpora"] != "drive":
```
shcheklein: can replace `not exc.errors` with `exc.errors.get("code", None)`

dvc/remote/gdrive.py
```diff
@@ -38,13 +38,34 @@ def __init__(self, path):
...
+    if hasattr(exc, "errors"):
+        return
+
+    exc.errors = None
 
+    if exc.args:
+        http_error = exc.args[0]
+        content = json.loads(http_error.content)
+        if content and "error" in content:
+            exc.errors = content["error"]
```
shcheklein: why do we call it `errors` is it's a single error?

dvc/remote/gdrive.py
shcheklein: empty hash

dvc/remote/gdrive.py
MaxRis: HttpError is part of `from apiclient import errors`, and seems we will need to add new `apiclient` deps to DVC.

MaxRis: Added there https://github.com/iterative/dvc/pull/3478/commits/91862981811c4813fbada625903abb3eac5fb915

dvc/remote/gdrive.py
MaxRis: Makes sense, thanks!

setup.py
```diff
@@ -84,7 +84,7 @@ def run(self):
 # Extra dependencies for remote integrations
 
 gs = ["google-cloud-storage==1.19.0"]
-gdrive = ["pydrive2>=1.4.6"]
+gdrive = ["pydrive2>=1.4.6", "google-api-python-client >= 1.7.12"]
```
shcheklein: we don't probably need this? pydrive installs the client anyway, HttpError should be available

dvc/remote/gdrive.py
shcheklein: We should be able to use it since pydrive depends on it, or am I missing something?

shcheklein: it might be a not exactly correct way of doing this though, it's better to be explicit indeed .... that's why I wanted to handle all of this on the pydrive end to keep all these details local to pydrive ... let's migrate this append_error_details to it? right into the ApiRequestError

setup.py
shcheklein: see my comment above, disregard this one

dvc/remote/gdrive.py
```diff
@@ -38,13 +38,37 @@ def __init__(self, path):
...
+    if hasattr(exc, "error"):
+        return
+
+    exc.error = {}
 
+    if exc.args:
+        from apiclient import errors
+
+        http_error = exc.args[0]
+        assert isinstance(http_error, errors.HttpError)
```
shcheklein: is it always HttpError? should we just ignore, keep it empty and raise to the user unknown exception

dvc/remote/gdrive.py
MaxRis: PyDrive2 PR for this https://github.com/iterative/PyDrive2/pull/18

dvc/remote/gdrive.py
MaxRis: HttpError currently is always passed to ApiRequestError. As discussed above, HttpError's json parsing is moved to PyDrive2 https://github.com/iterative/PyDrive2/pull/18

dvc/remote/gdrive.py
shcheklein: so, we don't need the whole append_error_details right? you will the code here?

MaxRis: yes, `append_error_details` will be removed completely (already tested locally)

MaxRis:
@shcheklein PR is updated, but depends on not yet merged changes in PyDrive2 (DVC pydrive2 deps version still should be increased in this PR)

setup.py
shcheklein: we don't need this dependency anymore I hope

dvc/remote/gdrive.py
```diff
@@ -324,10 +325,40 @@ def _create_remote_dir(self, parent_id, title):
...
         param = {"id": remote_id}
         # it does not create a file on the remote
         item = self.drive.CreateFile(param)
-        item.Delete()
+
+        try:
+            item.Trash() if self._trash_only else item.Delete()
+        except ApiRequestError as exc:
+            if (
+                not exc.error.get("code", None)
```
shcheklein: feels like these two conditions are too complicated, especially this one with two negations - really hard to read and understand - could we simplify it all?

dvc/remote/gdrive.py
```diff
@@ -324,10 +325,40 @@ def _create_remote_dir(self, parent_id, title):
...
+                and self.list_params["corpora"] != "drive"
+            ):
+                raise
+            if (
+                exc.error.get("code", None) == 403
```
shcheklein: I wonder if there is a case with it returning None?

dvc/remote/gdrive.py
```diff
@@ -324,10 +325,40 @@ def _create_remote_dir(self, parent_id, title):
...
+                and len(exc.error.get("errors", []))
+                and exc.error["errors"][0].get("location", "")
```
shcheklein: do you know if there is a case of multiple errors?

dvc/remote/gdrive.py
MaxRis: It seems it can't have 'None'. Finally, exactly this check is remove and upper condition is simplified to:
```
            http_error_code = exc.error.get("code", 0)
            if (
                http_error_code != 403
                or self.list_params["corpora"] != "drive"
            ):
                raise
```

dvc/remote/gdrive.py
MaxRis: can't say for sure, haven't seen more than one. But just discovered that `content` on which we do `json.loads` might be `list` sometimes according to https://github.com/googleapis/google-api-python-client/blob/master/googleapiclient/errors.py#L58

shcheklein:
@MaxRis  CI fails?

dvc/remote/gdrive.py
```diff
@@ -39,12 +39,14 @@ def __init__(self, path):
...
-
     def should_retry(exc):
         from pydrive2.files import ApiRequestError
 
-        return isinstance(exc, ApiRequestError) and retry_re.search(str(exc))
+        if not isinstance(exc, ApiRequestError):
+            return False
+
+        retry_codes = [403, 500, 502, 503, 504]
+        return exc.error.get("code", None) in retry_codes
```
shcheklein: change it here to the same as below?

dvc/remote/gdrive.py
```diff
@@ -324,10 +325,40 @@ def _create_remote_dir(self, parent_id, title):
...
+
         param = {"id": remote_id}
         # it does not create a file on the remote
         item = self.drive.CreateFile(param)
-        item.Delete()
+
+        try:
+            item.Trash() if self._trash_only else item.Delete()
+        except ApiRequestError as exc:
+            http_error_code = exc.error.get("code", 0)
```
shcheklein: so, if code is always set, let's just use `error["code"]`?

dvc/remote/gdrive.py
```diff
@@ -324,10 +325,40 @@ def _create_remote_dir(self, parent_id, title):
...
+            if (
+                http_error_code != 403
```
shcheklein: merge this into `if` below

dvc/remote/gdrive.py
```diff
@@ -324,10 +325,40 @@ def _create_remote_dir(self, parent_id, title):
...
+                or self.list_params["corpora"] != "drive"
+            ):
+                raise
+            if (
+                len(exc.error.get("errors", []))
```
shcheklein: make a small helper that extracts `location` - should be one simple line here

dvc/remote/gdrive.py
MaxRis: unfortunately, `code` key might be not available; only `exc.error` at the moment is initialized with empty dict `{}` in case if `code` is not available in original HttpError.

shcheklein: kk, let's at least use None or 0 consistently here and above

dvc/remote/gdrive.py
MaxRis: Done in https://github.com/iterative/dvc/pull/3478/commits/9ebc1000211bfe0a83bb76b4b1ac5a8715c67efc

dvc/remote/gdrive.py
```diff
@@ -54,6 +56,18 @@ def should_retry(exc):
     )(func)
 
 
+def location(exc):
```
shcheklein: `def _extract_error_details`

I would also extract `location` and `reason` and check both of them?

shcheklein: actually let's just do def `_location` - I think it's fine ... just put a comment what it does with a link to error strucutre

dvc/remote/gdrive.py
```diff
@@ -324,10 +337,35 @@ def _create_remote_dir(self, parent_id, title):
...
+
+        try:
+            item.Trash() if self._trash_only else item.Delete()
+        except ApiRequestError as exc:
+            http_error_code = exc.error.get("code", 0)
+            if (
+                http_error_code == 403
+                and self.list_params["corpora"] == "drive"
+                and location(exc) == "file.permissions"
+            ):
```
shcheklein: k, finally we are getting there - simple and human readable! thanks

shcheklein:
also, check this https://deepsource.io/gh/iterative/dvc/run/b292b25e-6435-478e-9713-39fb124b636c/python/PYL-C1801

dvc/remote/gdrive.py
```diff
@@ -324,10 +337,35 @@ def _create_remote_dir(self, parent_id, title):
...
+                    "Insufficient permissions to {}. You should have {} "
+                    "access level for the used shared drive. More details "
+                    "at {}".format(
+                        "move the file into Trash"
+                        if self._trash_only
+                        else "permanently delete the file",
+                        "Manager or Content Manager"
+                        if self._trash_only
+                        else "Manager",
+                        "https://support.google.com/a/answer/7337554?hl=en",
```
shcheklein: remove hl=en ?

dvc/remote/gdrive.py
```diff
@@ -324,10 +337,35 @@ def _create_remote_dir(self, parent_id, title):
...
+            http_error_code = exc.error.get("code", 0)
+            if (
+                http_error_code == 403
+                and self.list_params["corpora"] == "drive"
+                and location(exc) == "file.permissions"
+            ):
+                raise DvcException(
+                    "Insufficient permissions to {}. You should have {} "
+                    "access level for the used shared drive. More details "
+                    "at {}".format(
```
shcheklein: put a period at the end

MaxRis: Sounds good; Initially, haven't put period because it looked ugly to me (please see the attached screenshot).
2020-03-25_15-35-52-period

shcheklein:
@MaxRis Travis fails?

shcheklein:
@efiop good to merge when/if tests pass

PR #3490, opened 2020-03-17
dvc: add Repo(skip_checks) flag

### LINKED ISSUE #2671: Optimize stage collection for large repos

Suor:
Will skip modified graph checks on add, run and import. The aim is to
avoid collecting stages and building graph on these commands.

The skip_checks flag is aimed to be private, but let people use dvc on
big repos without monkey patching or writing yaml files themselves.

Closes #2671. For the time being.

### DIFF
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -174,15 +174,18 @@ class Repo(object):
 
     def check_modified_graph(self, new_stages):
         """Generate graph including the new stage to check for errors"""
-        # Building graph might be costly for ones with many stages,
-        # we provide this non-documented hack to skip this for add and run.
-        # Can be used as:
+        # Building graph might be costly for the ones with many DVC-files,
+        # so we provide this undocumented hack to skip it. See [1] for
+        # more details. The hack can be used as:
         #
         #     repo = Repo(...)
         #     repo._skip_graph_checks = True
         #     repo.add(...)
         #
-        # A user should care about not duplicating outs and not adding cycles.
+        # A user should care about not duplicating outs and not adding cycles,
+        # otherwise DVC might have an undefined behaviour.
+        #
+        # [1] https://github.com/iterative/dvc/issues/2671
         if not getattr(self, "_skip_graph_checks", False):
             self._collect_graph(self.stages + new_stages)
 
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -174,6 +174,15 @@ class Repo(object):
 
     def check_modified_graph(self, new_stages):
         """Generate graph including the new stage to check for errors"""
+        # Building graph might be costly for ones with many stages,
+        # we provide this non-documented hack to skip this for add and run.
+        # Can be used as:
+        #
+        #     repo = Repo(...)
+        #     repo._skip_graph_checks = True
+        #     repo.add(...)
+        #
+        # A user should care about not duplicating outs and not adding cycles.
         if not getattr(self, "_skip_graph_checks", False):
             self._collect_graph(self.stages + new_stages)
 
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -174,7 +174,8 @@ class Repo(object):
 
     def check_modified_graph(self, new_stages):
         """Generate graph including the new stage to check for errors"""
-        self._collect_graph(self.stages + new_stages)
+        if not getattr(self, "_skip_graph_checks", False):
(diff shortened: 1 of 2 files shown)


dvc/repo/__init__.py
```diff
@@ -60,7 +60,7 @@ class Repo(object):
     from dvc.repo.get_url import get_url
     from dvc.repo.update import update
 
-    def __init__(self, root_dir=None):
+    def __init__(self, root_dir=None, skip_checks=False):
```
efiop: I'm really worried about this ending very badly. We should at least hide it by making it `_skip_checks` and putting a comment/docstring that people should not use it at all, unless they really know what they are doing. We need to clarify the reasoning behind it.

skshetry: How about having a internal API to disable?
```python
dvc = Repo()
dvc._skip_checks()
```

Suor: We don't guarantee any backward compatibility here anyway.

Suor: I am fine with other ways though.

efiop: @skshetry Great idea! That way it is well hidden and is not hurting our eyes in the public interface.

@Suor Could you please adjust your PR to do that? E.g.
```
dvc = Repo()
dvc._skip_checks = True
```
would probably be the best. Or maybe `_disable_dag_checks`, but naming is up to you.

Suor: Changed to `repo._skip_graph_checks = True`.

dvc/repo/__init__.py
```diff
@@ -174,7 +174,8 @@ def _ignore(self):
 
     def check_modified_graph(self, new_stages):
         """Generate graph including the new stage to check for errors"""
-        self._collect_graph(self.stages + new_stages)
+        if not getattr(self, "_skip_graph_checks", False):
+            self._collect_graph(self.stages + new_stages)
```
efiop: Maybe set it in to `False` in `__init__` along with a nice comment explaining what it is for? Also, we need a test here.

Suor: Prefer not touching constructor, keep it in single place. Added an explanation comment and a test.

tests/unit/repo/test_repo.py
```diff
@@ -82,3 +82,18 @@ def test_collect_optimization(tmp_dir, dvc, mocker):
     # Should read stage directly instead of collecting the whole graph
     dvc.collect(stage.path)
     dvc.collect_granular(stage.path)
+
+
+def test_skip_graph_checks(tmp_dir, dvc, mocker, run_copy):
+    (stage,) = tmp_dir.dvc_gen("foo", "foo text")
+
+    # Error out on graph collection and raise a kip flag
```
skshetry: `kip`?

Suor: Fixed

Suor:
Should be good to go.

tests/unit/repo/test_repo.py
```diff
@@ -82,3 +82,18 @@ def test_collect_optimization(tmp_dir, dvc, mocker):
...
+
+
+def test_skip_graph_checks(tmp_dir, dvc, mocker, run_copy):
+    (stage,) = tmp_dir.dvc_gen("foo", "foo text")
+
+    # Error out on graph collection and raise a skip graph checks flag
+    mocker.patch(
+        "dvc.repo.Repo._collect_graph",
+        side_effect=Exception("Should not collect"),
+    )
```
efiop: It would be better to just assert that this wasn't called rather than putting this exception here, as we might have some logic catching this in our code (yes, we don't do that now and catching Exception is a bad thing, but still), which would result in a faulty test. Plus it would be more explicit to use `assert not m.called` rather than this unnecessary hack.

tests/unit/repo/test_repo.py
```diff
@@ -82,3 +82,18 @@ def test_collect_optimization(tmp_dir, dvc, mocker):
...
+    dvc._skip_graph_checks = True
+
+    tmp_dir.gen("foo", "foo text")
```
efiop: Also you have `dvc_gen` above. Not sure if it is intended.

PR #3493, opened 2020-03-17
gc: do not work without specifier when using -c flag

### LINKED ISSUE #3489: gc: `-c` works without scope specifier (-w,a,T etc)
As the title says, `dvc gc -c` works without `-w`/`-T`/`-a` specifiers (i.e. by default, it assumes `-w`). 
It'd be good to only get in action when those specifiers are mentioned for `dvc gc -c`.

### Refs:
[Discord discussion](https://discordapp.com/channels/485586884165107732/565699007037571084/689350993796006072)
#3363 made it mandatory to require use of flags.

skshetry:

Fix #3489.

### DIFF
--- a/dvc/repo/gc.py
+++ b/dvc/repo/gc.py
@@ -46,7 +46,6 @@ def gc(
         all_tags=all_tags,
         all_commits=all_commits,
         all_branches=all_branches,
-        cloud=cloud,
     )
 
     from contextlib import ExitStack
--- a/dvc/command/gc.py
+++ b/dvc/command/gc.py
@@ -19,7 +19,6 @@ class CmdGC(CmdBase):
             all_tags=self.args.all_tags,
             all_commits=self.args.all_commits,
             workspace=self.args.workspace,
-            cloud=self.args.cloud,
         )
 
         msg = "This will remove all cache except items used in "
--- a/tests/func/test_data_cloud.py
+++ b/tests/func/test_data_cloud.py
@@ -367,7 +367,7 @@ class TestDataCloudCLIBase(TestDvc):
         self.assertTrue(os.path.isfile(cache_dir))
 
         # NOTE: check if remote gc works correctly on directories
-        self.main(["gc", "-c", "-f"] + args)
+        self.main(["gc", "-cw", "-f"] + args)
         shutil.move(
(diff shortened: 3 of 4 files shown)


dvc/repo/gc.py
```diff
@@ -46,7 +46,6 @@ def gc(
         all_tags=all_tags,
         all_commits=all_commits,
         all_branches=all_branches,
-        cloud=cloud,
```
efiop: Not sure why we are removing it here, while we still have `if cloud` down below. Am i missing something?

skshetry: Previously, if we just did `-c` or `cloud=True`, it'd not raise a warning and by default assume `--workspace`. Now, even if `-c` is specified, unless `workspace/all_tags/all_commits/all_branches` are also specified, it will raise warning.

efiop: @skshetry Looks like there is some misunderstanding here. I simply don't see `cloud` defined anywhere in this method, while it is used down below in `if cloud`.

skshetry: So, `_raise_error_if_all_disabled` would throw following error if all of the arguments were false. 
```sh
ERROR: unexpected error - Invalid Arguments. Either of (all_branches, all_tags, all_commits, workspace) needs to be enabled.
```

So, when we do, `dvc gc -c`, `cloud` is `True`, and hence would not have thrown any error, and simply worked as if `-cw` was specified (as default is `workspace`).

But, now,I have removed `cloud` from the arguments. Even if `cloud` is set to `True`, unless any of the other arguments are `True`, it will raise an error. So: `dvc gc -c` will now throw the same error.

skshetry: Are you confusing it with passing to the `repo.gc()` function?

efiop: Ok, there was a misunderstanding on my side. I thought you are removing `cloud` from method's(gc) arguments, while you are simply removing it from arguments for `    _raise_error_if_all_disabled(	` . :ok_hand:

skshetry: Ahh, I see were the confusion comes from:
Screenshot from 2020-03-17 18-23-51

PR #3500, opened 2020-03-18
progress: limit maximum running bars

### LINKED ISSUE #2905: pull: going completely crazy on large dataset

casperdcl:
- [x] ~~introduce `Tqdm.MAX_BARS = 20` for parallel display~~
  + [x] ~~add ` ... (more hidden) ...` message for overflow bars~~
- [x] ~~auto-detect screen height? Should be done in upstream https://github.com/tqdm/tqdm/issues/918~~
- [x] use [`tqdm>=4.44.0`](https://tqdm.github.io/releases/#v4440-2020-03-28) which auto-detects screen height with a fallback of `20`
- TODO (future): combine sub bars into a single high-level bar
- fixes #2905
- Replaces/closes #3453
- related #3452

issue & fix:

[](https://asciinema.org/a/wpM1zYFg9TfH53uiNPMsKjiUt?speed=2)

(from #3453 (cc @JIoJIaJIu): [script for reproducing issue & fix](https://gist.github.com/JIoJIaJIu/05f3b72982373771da88f313bb5cad9f))


dvc/progress.py
```diff
@@ -35,6 +35,7 @@ class Tqdm(tqdm):
     BYTES_DEFAULTS = dict(
         unit="B", unit_scale=True, unit_divisor=1024, miniters=1
     )
+    MAX_BARS = 20  # maximum parallel bars to display
```
casperdcl: ideally auto-detect; see https://github.com/tqdm/tqdm/blob/master/tqdm/utils.py#L263-L330 for inspiration

efiop:
@casperdcl Could you show before/after capture, please?

efiop:
@casperdcl Also, wouldn't it make sense to patch tqdm itself to tidy up unused bars as it goes?

casperdcl:
> Could you show before/after capture, please?

The script and asciinema recordings from @JIoJIaJIu (see description at the top here) still apply. I can re-generate them but they'd look identical.

> Also, wouldn't it make sense to patch tqdm itself to tidy up unused bars as it goes?

This isn't about tidying up unused bars. Its about hiding running bars.

Specifically:

- completed `leave=True` bars have never been an issue - they just print normally
- running `disable=False` bars could scroll off screen if there are more than SCREEN_HEIGHT of them. This PR dynamically hides them.

By "dynamically" I mean

- completed `leave=True` bars would still print normally only after completion
- when a displaying bar completes it may be replaced by one which was hidden

casperdcl:
> The script and asciinema recordings from @JIoJIaJIu (see description at the top here) still apply. I can re-generate them but they'd look identical.


Actually not true. Updated:

[](https://asciinema.org/a/y5GY0r2kixuwPVMwkWAher81z?speed=4)

efiop:
@casperdcl By `unused bars` i meant blank lines, that make current pbars look like they are flickering in a mess. I like how new approach looks like :+1: But maybe this should be done in tqdm itself, after all? Seems like a feature everyone could benefit from. To be honest, I'm not too deep into the implementation details here, so I'm simply judging by the looks.

JIoJIaJIu:
> Actually not true. Updated

But you launched with `--tqdm` flag that means you use library tqdm version. 

I tried and noticed that it works! :tada: And it look nicer from the code side
What interesting when I was working on the issue it was my first approach - usage `display` and  `clean` methods, but it didn't help, but now I cannot reproduce it and everything works

One disadvantage - that it flicks too much. I think cause all pbars going up (`pos -=1`) on removing pbar. It would be better if we can just put hidden bar in the place of the closed one

casperdcl:
> But you launched with `--tqdm` flag that means you use library tqdm version.

Yes but at the end of the animation it's run again without the flag

casperdcl:
> It would be better if we can just put hidden bar in the place of the closed one

Interesting idea. I prefer moving up logically as newer bars at the bottom will be less complete and higher bars will be more complete - e.g. assuming similar file sizes:

```
bar10: ####
bar12: #######
bar13: ###########
bar14: #################
```

Rather than more randomly ordered:

```
bar10: ####
bar14: #################
bar12: #######
bar13: ###########
```

Maybe for a different issue/PR?

casperdcl:
Upstream issue for fixing this properly: https://github.com/tqdm/tqdm/issues/918

casperdcl:
updated with message:

[](https://asciinema.org/a/wpM1zYFg9TfH53uiNPMsKjiUt?speed=2)

Think this is mergeable now as-is.

efiop:
For the record: had a discussion about it last Sat and decided to wait for this functionality to become available in the upstream tqdm.

casperdcl:
... aaand new version!

PR #3501, opened 2020-03-18
remote: Optimize traverse/no_traverse behavior

### LINKED ISSUE #3488: remote: improve `traverse`/`no_traverse` behavior
Currently for `push`/`pull` operations, we either fetch the full list of all files in the remote in order to determine what local files already exist in the remote cache, or we individually check whether or not each local file exists in the remote cache (`no_traverse` method).

As explained in #2044, the first approach works best for cases where the local cache is large, and the remote cache is small, and the `no_traverse` approach works best when the local cache is small and the remote is large.

By default, we currently use the `no_traverse` method by default for all remotes except `gdrive` (discounting remotes with overridden `cache_exists()` methods like `ssh` and `local`). `no_traverse` can be configured via an option to enable it for `gdrive`.

Rather than relying on the user to explicitly specify the `no_traverse` option, it is possible for us take some steps towards determining which method to use during a `push`/`pull` operation.

Since files in the remote cache will be evenly distributed (according to md5 hash), we can fetch a subset of files from the remote cache (rather than the full cache all at once), and use the size of that subset to estimate the size of the entire remote cache. For example, if we only fetch cache entries under a single parent directory (a.k.a. the first byte of the md5 hash for those entries), the total number of files on the remote would be roughly `256 * len(subset_results)`.

From there, we can determine whether or not to use the `no_traverse` method, by comparing the local file count to the estimated remote file count.

pmrowla:

- Estimate remote file count by fetching a single parent cache dir, then
  determine whether or not to use no_traverse method for checking
  remainder of cache entries in `cache_exists`
- Thread requests when traversing full remote file lists (by fetching
  one parent cache dir per thread)

Will close #3488.

TODO:
- [x] Azure
- [x] S3
- [x] GDrive
- [x] Google CS
- [x] HDFS
- [x] OSS
- [x] Tests
- [ ] Docs


pmrowla:
Currently only tested on S3

`dvc push` runtime comparisons for S3 remote with 1M cache total entries, and smaller local repos:
```
remote-100k-files git:master ✶ ✩  py:dvc ❯ /usr/local/bin/dvc --version
0.88.0
remote-100k-files git:master ✶ ✩  py:dvc ❯ time /usr/local/bin/dvc push
 Everything is up to date.
/usr/local/bin/dvc push  421.31s user 25.58s system 37% cpu 19:58.64 total
remote-100k-files git:master ✶ ✩  py:dvc ❯ time dvc push
Everything is up to date.
dvc push  225.73s user 14.12s system 98% cpu 4:02.43 total
```

When testing with 1k and 10k local files, `no_traverse` is always faster

pmrowla:
Should be able to start running gdrive benchmarks tomorrow, currently waiting on a push to populate a test remote to finish overnight

pmrowla:
## Benchmarks

`0.88.0` from homebrew vs `git:3488` branch
All times measured with

```
time dvc push
```

for an up-to-date repo (so no data needs to be pushed, and `cache_exists` is the main runtime factor)

### S3

| dvc | `no_traverse` | local files | remote files | runtime |
| --- | --- | --- | --- | --- |
| 0.88.0 | true | 1k | 1k | `10.07s user 1.47s system 61% cpu 18.733 total` |
| 0.88.0 | false | 1k | 1k | `1.79s user 0.67s system 34% cpu 7.096 total` |
| git | n/a | 1k | 1k | `1.73s user 0.57s system 44% cpu 5.181 total`
| 0.88.0 | true | 10k | 100k | `44.44s user 3.28s system 39% cpu 2:01.46 total` |
| 0.88.0 | false | 10k | 100k | `21.48s user 1.42s system 18% cpu 2:07.25 total` |
| git | n/a | 10k | 100k | `24.64s user 1.71s system 91% cpu 28.717 total`
| 0.88.0 | true | 100k | 1M | `421.31s user 25.58s system 37% cpu 19:58.64 total` |
| git:master | false | 100k | 1M | `205.87s user 7.83s system 29% cpu 11:53.41 total` |
| git | n/a | 100k | 1M | `225.73s user 14.12s system 98% cpu 4:02.43 total` |


### GDrive

| dvc | `no_traverse` | local files | remote files | runtime |
| --- | --- | --- | --- | --- |
| 0.88.0 | false | 10k | 500k | >30min (killed before finished running) |
| git | n/a | 10k | 500k | `78.13s user 4.94s system 57% cpu 2:23.25 total` |
| git | n/a | 100k | 500k | `61.82s user 6.66s system 43% cpu 2:36.76 total` |

### HDFS

Tested w/local single node hadoop instance inside virtualbox, dvc `0.87.0` from deb

| dvc | `no_traverse` | local files | remote files | runtime |
| --- | --- | --- | --- | --- |
| 0.87.0 | true | 10k | 100k | `real 0m18.388s user 0m20.496s sys 0m5.096s` |
| git | n/a | 10k | 100k | `real 0m17.542s user 0m20.316s sys 0m2.292s` |
| 0.87.0 | true | 100k | 100k | `real 2m34.512s user 2m9.272s sys 0m32.828s` |
| git | n/a | 100k | 100k | `real 0m29.858s user 0m31.596 sys 0m3.096` |

dvc/remote/base.py
```diff
@@ -820,9 +825,102 @@ def cache_exists(self, checksums, jobs=None, name=None):
...
+                len(checksums)
+            )
+        )
+
+        traverse_prefixes = ["{:02x}".format(i) for i in range(1, 256)]
+        if self.TRAVERSE_PREFIX_LEN > 2:
+            traverse_prefixes += [
+                "{0:0{1}x}".format(i, self.TRAVERSE_PREFIX_LEN)
+                for i in range(1, pow(16, self.TRAVERSE_PREFIX_LEN - 2))
+            ]
```
pared: Shouldn't here be range(0,256)?
What if `TRAVERSE_PREFIX_LEN` == 2?
In that case, it seems to me that we will not gather remote cache content starting with `00`
Am I right?

dvc/remote/azure.py
```diff
@@ -103,8 +104,12 @@ def _list_paths(self, bucket, prefix):
 
             next_marker = blobs.next_marker
 
-    def list_cache_paths(self):
-        return self._list_paths(self.path_info.bucket, self.path_info.path)
+    def list_cache_paths(self, prefix=None):
+        if prefix:
+            prefix = "/".join([self.path_info.path, prefix[:2], prefix[2:]])
```
efiop:
```suggestion
            prefix = posixpath.join([self.path_info.path, prefix[:2], prefix[2:]])
```

dvc/remote/azure.py
```diff
@@ -26,6 +26,7 @@ class RemoteAZURE(RemoteBASE):
         r"(?P<connection_string>.+)?)?)$"
     )
     REQUIRES = {"azure-storage-blob": "azure.storage.blob"}
+    DEFAULT_NO_TRAVERSE = False
```
efiop: Do we still need this flag per-remote? Looks like we have a default automatic detection now. Should we also keep an ability to explicitly set no_traverse in config?

dvc/remote/base.py
```diff
@@ -820,9 +825,102 @@ def cache_exists(self, checksums, jobs=None, name=None):
...
+        else:
+            remote_size = traverse_dirs * len(remote_checksums)
+        logger.debug("Estimated remote size: {} files".format(remote_size))
+
+        traverse_pages = remote_size / self.LIST_OBJECT_PAGE_SIZE
+        # For sufficiently large remotes, traverse must be weighted to account
+        # for performance overhead from large lists/sets.
+        # From testing with S3, for remotes with 1M+ files, no_traverse is
+        # faster until len(checksums) is at least 10k~100k
+        if remote_size > 500000:
```
efiop: Let's use a constant here as well. E.g. `self.THRESHOLD_REMOTE_SIZE` or smth like that. Nitpick, but should be nicer and more flexible :wink:

dvc/remote/base.py
```diff
@@ -820,9 +825,102 @@ def cache_exists(self, checksums, jobs=None, name=None):
         Returns:
             A list with checksums that were found in the remote
         """
-        if not self.no_traverse:
-            return list(set(checksums) & set(self.all()))
 
+        if self.no_traverse:
+            return self._cache_exists_no_traverse(checksums, jobs, name)
```
efiop: Continuing question from https://github.com/iterative/dvc/pull/3501/files#r396638412 . Should we make autodetection default behaviour (no_traverse not set in config) with an ability to explicitly set no_traverse to true/false? Currently it seems like you only do auto-detection when `no_traverse` is `false`, is there a reason for that?

dvc/remote/hdfs.py
```diff
@@ -152,16 +154,26 @@ def open(self, path_info, mode="r", encoding=None):
...
             raise
 
-    def list_cache_paths(self):
+    def list_cache_paths(self, prefix=None):
         if not self.exists(self.path_info):
             return
 
-        dirs = deque([self.path_info.path])
+        if prefix:
+            root = "/".join([self.path_info.path, prefix[:2]])
```
efiop: Nitpick:
```suggestion
            root = posixpath.join([self.path_info.path, prefix[:2]])
```

dvc/remote/local.py
```diff
@@ -56,7 +56,7 @@ def cache_dir(self, value):
     def supported(cls, config):
         return True
 
-    def list_cache_paths(self):
+    def list_cache_paths(self, prefix=None):
         assert self.path_info is not None
```
efiop:
```suggestion
        assert prefix is None
        assert self.path_info is not None
```
just to make sure that no one will try to use it.

dvc/remote/ssh/__init__.py
```diff
@@ -249,7 +249,7 @@ def open(self, path_info, mode="r", encoding=None):
             else:
                 yield io.TextIOWrapper(fd, encoding=encoding)
 
-    def list_cache_paths(self):
+    def list_cache_paths(self, prefix=None):
```
efiop:
```suggestion
    def list_cache_paths(self, prefix=None):
        assert prefix is None
```

dvc/remote/base.py
pmrowla: The cache content starting with `00` or `000` (depending on prefix length) is fetched in the earlier step when we do the cache size estimation and gets passed into this function in `remote_checksums`

dvc/remote/azure.py
pmrowla: I was thinking that if we leave the `no_traverse` config option, users who are sure that `no_traverse = true` will always be faster for their particular use case can set it to avoid the auto-detection step entirely.

dvc/remote/base.py
pmrowla: This is for the same reason as the other question, assuming that we keep the option so users can force the `no_traverse = true` behavior, then we can skip the autodetection in that case.

Even if the option is explicitly set to `false` by a user, I think it makes sense to still do the autodetection, since in this case the cache content used to estimate the remote size will always be fetched regardless of whether or not we are doing the autodetection.

dvc/remote/base.py
pmrowla: So here we just need to fetch `01`-`ff` regardless of prefix length, plus `001-00f` when prefix length is 3

dvc/remote/azure.py
pmrowla: fixed in f17a0ad

dvc/remote/base.py
pmrowla: fixed in f17a0ad

dvc/remote/hdfs.py
pmrowla: fixed in f17a0ad

dvc/remote/local.py
pmrowla: fixed in f17a0ad

dvc/remote/ssh/__init__.py
pmrowla: fixed in f17a0ad

dvc/remote/base.py
shcheklein: should we consider making a new option if we want to keep the explicit behavior support? It will "reset" existing repos to the desirable auto detection, make this option name "better" (`no_traverse` is very confusing).

dvc/remote/base.py
```diff
@@ -804,7 +808,9 @@ def cache_exists(self, checksums, jobs=None, name=None):
 
         - Traverse: Get a list of all the files in the remote
             (traversing the cache directory) and compare it with
-            the given checksums.
+            the given checksums. Cache parent directories will
```
shcheklein: should we update the whole comment that we have this autodection now? also, would be great to have a wiki or put a comment with some logic behind it - it's core and it's not very simple of others to understand

dvc/remote/base.py
```diff
@@ -820,9 +826,102 @@ def cache_exists(self, checksums, jobs=None, name=None):
         Returns:
             A list with checksums that were found in the remote
         """
-        if not self.no_traverse:
-            return list(set(checksums) & set(self.all()))
 
+        if self.no_traverse:
+            return self._cache_exists_no_traverse(checksums, jobs, name)
+
+        # Fetch one parent cache dir for estimating size of entire remote cache
```
shcheklein: comment is a bit misleading ... am I right that we actually fetch not a "parent" dir for some remotes?

dvc/remote/base.py
```diff
@@ -820,9 +826,102 @@ def cache_exists(self, checksums, jobs=None, name=None):
...
+        prefix = "0" * self.TRAVERSE_PREFIX_LEN
+        traverse_dirs = pow(16, self.TRAVERSE_PREFIX_LEN)
+        remote_checksums = set(
+            map(self.path_to_checksum, self.list_cache_paths(prefix=prefix))
+        )
+        if not remote_checksums:
+            remote_size = traverse_dirs
+        else:
+            remote_size = traverse_dirs * len(remote_checksums)
+        logger.debug("Estimated remote size: {} files".format(remote_size))
```
shcheklein: I really like how loggers are used here 👍 we should be doing this more everywhere ... I miss this stuff a lot in DVC - it make most of our logs "unusable" and we have to ask people questions - a lot of questions

dvc/remote/base.py
```diff
@@ -820,9 +826,102 @@ def cache_exists(self, checksums, jobs=None, name=None):
...
+            return self._cache_exists_no_traverse(checksums, jobs, name)
+
+        # Fetch one parent cache dir for estimating size of entire remote cache
+        checksums = frozenset(checksums)
+        prefix = "0" * self.TRAVERSE_PREFIX_LEN
+        traverse_dirs = pow(16, self.TRAVERSE_PREFIX_LEN)
+        remote_checksums = set(
+            map(self.path_to_checksum, self.list_cache_paths(prefix=prefix))
+        )
+        if not remote_checksums:
```
shcheklein: minor: always better to use positive `if`

dvc/remote/base.py
```diff
@@ -820,9 +826,102 @@ def cache_exists(self, checksums, jobs=None, name=None):
...
-        if not self.no_traverse:
-            return list(set(checksums) & set(self.all()))
 
+        if self.no_traverse:
+            return self._cache_exists_no_traverse(checksums, jobs, name)
+
+        # Fetch one parent cache dir for estimating size of entire remote cache
+        checksums = frozenset(checksums)
+        prefix = "0" * self.TRAVERSE_PREFIX_LEN
+        traverse_dirs = pow(16, self.TRAVERSE_PREFIX_LEN)
```
shcheklein: could you briefly describe what's the logic here - why do we translate this into remote size if prefix is empty

dvc/remote/base.py
```diff
@@ -820,9 +826,102 @@ def cache_exists(self, checksums, jobs=None, name=None):
...
+        # for performance overhead from large lists/sets.
+        # From testing with S3, for remotes with 1M+ files, no_traverse is
+        # faster until len(checksums) is at least 10k~100k
+        if remote_size > self.TRAVERSE_THRESHOLD_SIZE:
+            traverse_weight = traverse_pages * self.TRAVERSE_WEIGHT_MULTIPLIER
+        else:
+            traverse_weight = traverse_pages
+        if len(checksums) < traverse_weight:
+            logger.debug(
+                "Large remote, using no_traverse for remaining checksums"
```
shcheklein: put len(checksums) and traverse_weight into the log?

dvc/remote/base.py
pmrowla: I wrote this with the assumption that `TRAVERSE_PREFIX_LEN` would never be zero - the remotes which don't use the prefix at all (ssh, local) have their own overridden `cache_exists()` methods. But it does make sense to account for an empty prefix here as well.

shcheklein: sorry, my question was confusing: `prefix empty == size of the object with that prefix in the if below` ... I was asking mostly about the `if` below when we already checked the size of the remote at the prefix.

pmrowla: If the fetched prefix is empty, we still need to estimate that the remote size is `< 256` or `< 4096` files (depending on prefix length) rather than `0`. Later on we use that size to estimate how many pages (`size / 1000` for most remotes) the full list of remote objects would return, and we want a non-zero page count.

shcheklein: I think the name confused me a bit ... since it's not so much about `dirs` in general case ... `total_prefixes`?

shcheklein:
suggestion, btw - does it make sense to short-cut on a certain number of checksums? like it's clear that if it's one we can always to a singe exist request?

dvc/remote/base.py
pmrowla: updated the comment to explain how the autodetection works in 
65d7e22

If we keep `no_traverse` option (or rename/add a new one) this should also go into the dvc.org docs since `no_traverse` was not documented at all in the first place

dvc/remote/base.py
pmrowla: clarified comment in 65d7e22 since we're really fetching everything under prefix and not parent dirs

dvc/remote/base.py
pmrowla: fixed in 65d7e22

dvc/remote/base.py
pmrowla: renamed var to `total_prefixes` in 65d7e22

pmrowla:
> suggestion, btw - does it make sense to short-cut on a certain number of checksums? like it's clear that if it's one we can always to a singe exist request?

I think it makes sense to short-cut for small numbers of checksums, but I'm not sure where to set the cutoff since I don't have a great idea of what the typical remote use case looks like (which affects how many requests the autodetection/size estimation takes)

dvc/remote/base.py
shcheklein: 👍

dvc/remote/base.py
shcheklein: thanks 🙏

dvc/remote/base.py
pared: Sorry, right, forgot that you used `00` for estimation of remote size.

dvc/remote/base.py
```diff
@@ -817,12 +823,120 @@ def cache_exists(self, checksums, jobs=None, name=None):
...
+            unit="dir",
+        ) as pbar:
+
+            def list_with_update(prefix):
+                ret = map(
+                    self.path_to_checksum,
+                    list(self.list_cache_paths(prefix=prefix)),
+                )
+                pbar.update_desc(
+                    "Querying cache in {}/{}".format(self.path_info, prefix)
```
efiop:
```suggestion
                    "Querying cache in '{}'".format(self.path_info / prefix)
```

dvc/remote/base.py
```diff
@@ -817,12 +823,120 @@ def cache_exists(self, checksums, jobs=None, name=None):
...
+        # for performance overhead from large lists/sets.
+        # From testing with S3, for remotes with 1M+ files, no_traverse is
+        # faster until len(checksums) is at least 10k~100k
+        if remote_size > self.TRAVERSE_THRESHOLD_SIZE:
+            traverse_weight = traverse_pages * self.TRAVERSE_WEIGHT_MULTIPLIER
+        else:
+            traverse_weight = traverse_pages
+        if len(checksums) < traverse_weight:
+            logger.debug(
+                "Large remote ({} checksums < {} traverse weight), "
```
efiop: Some minor things: we try to keep generated parts of the messages in single quotes, so that it is easier to visually see them in logs.
```suggestion
                "Large remote ('{}' checksums < '{}' traverse weight), "
```

dvc/remote/base.py
efiop: @pmrowla Forcing `no_traverse=true` doesn't seem like something actual users might use, as usefulness will depend on the number of hashes you need to check. So sounds like that might only be useful for debugging purposes, but in that case forcing `false` also makes sense. But I'm not sure if debugging purposes make sense here at all, to be honest. Maybe we could simply get rid of this config option (just make it obsolete and defunct)? Or do you see any use for it from user or dev perspective?

dvc/remote/base.py
```diff
@@ -817,12 +823,120 @@ def cache_exists(self, checksums, jobs=None, name=None):
...
+                "using no_traverse for remaining checksums".format(
```
efiop: Since `no_traverse` is confusing, do you have any better name for it? For in-code purposes, not even talking about renaming the config option necessarily.

efiop:
@pmrowla Looks great! The only question to figure out is https://github.com/iterative/dvc/pull/3501#discussion_r397082374 , the rest of my above comments are just tiny nitpicks.

dvc/remote/base.py
```diff
@@ -817,12 +823,120 @@ def cache_exists(self, checksums, jobs=None, name=None):
...
+
+        traverse_prefixes = ["{:02x}".format(i) for i in range(1, 256)]
+        if self.TRAVERSE_PREFIX_LEN > 2:
+            traverse_prefixes += [
+                "{0:0{1}x}".format(i, self.TRAVERSE_PREFIX_LEN)
+                for i in range(1, pow(16, self.TRAVERSE_PREFIX_LEN - 2))
+            ]
+        with Tqdm(
+            desc="Querying "
+            + ("cache in " + name if name else "remote cache"),
```
skshetry: Nitpicky: use format-strings?

skshetry: Looks like it's same on no-traverse.

dvc/remote/base.py
pmrowla: In that case I think it would make the most sense to just drop the config option entirely

efiop: @pmrowla :+1: Let's do that then in this PR. See `protected` config option as an example of obsoleting.

dvc/remote/base.py
pmrowla: We could just go with `list objects method` and `object exists method` instead of `traverse method` and `no_traverse method`

dvc/remote/base.py
pmrowla: `no_traverse` config option obsoleted in 88ea4dd

dvc/remote/base.py
pmrowla: fixed in 6093868

dvc/remote/base.py
pmrowla: got rid of references to `no_traverse` in 6093868

comments and code now use `traverse` (same as before) and `object exists`

dvc/remote/base.py
pmrowla: fixed in 6093868

dvc/remote/base.py
```diff
@@ -817,12 +822,120 @@ def cache_exists(self, checksums, jobs=None, name=None):
...
+        if len(checksums) == 1 or not self.CAN_TRAVERSE:
+            return self._cache_object_exists(checksums, jobs, name)
+
+        # Fetch cache entries beginning with "00..." prefix for estimating the
+        # size of entire remote cache
+        checksums = frozenset(checksums)
+        prefix = "0" * self.TRAVERSE_PREFIX_LEN
+        total_prefixes = pow(16, self.TRAVERSE_PREFIX_LEN)
+        remote_checksums = set(
+            map(self.path_to_checksum, self.list_cache_paths(prefix=prefix))
```
shcheklein: one more comment (after watching the demo) - I think what we miss here is also some threshold on how "deep" we want to go with the prefix. Let's imagine we have 256M files and we'd like to push a single file. We will make 1000 requests, right? While what we can do is to see the first 1000 and make some conclusions right away.

efiop: For the record: we've discussed this and it will be implemented in a separate PR.

shcheklein: 👍

efiop:
Thank you @pmrowla ! :pray: SSH is a tricky beast here, because it is way too sensitive to the number of connections that it can handle. Let's see how it goes in the future.

PR #3509, opened 2020-03-19
import: fix output message on not-existing output directory

### LINKED ISSUE #3338: Incorrect error message when importing to non-existent folder
## Setup
dvc 0.84.0
Mac, installation via conda

## Problem
When importing a file from a git repo into a subdirectory in the current repo, dvc fails with an incorrect error message if the subdirectory does not exist. 

To demonstrate this problem, run the following:

```
## Setup a git repo to import from
mkdir source_repo
cd source_repo
git init
echo Some code > code.txt
git add .
git commit -m "Save some code to this demo repo"

## setup a DVC repo to import from the previous
cd -
mkdir import_repo
cd import_repo
git init
dvc init
git add .
git commit -m "This is a dvc repo that will import some code"

echo --------------------------
echo The next command will fail with an incorrect error message
echo --------------------------
read -rsp $'Press any key to continue...\n' -n1 key

## Attempt an import which will fail
# This command will fail:
dvc import -o some_dir/code.txt ../source_repo code.txt

mkdir some_dir

echo --------------------------
echo Now we have made the output directory, it will succeed
echo --------------------------
read -rsp $'Press any key to continue...\n' -n1 key

dvc import -o some_dir/code.txt ../source_repo code.txt
```

## Expected behavior
Error message stating that the output directory does not exist

## Actual behaviour
This error message:
```
ERROR: failed to import 'code.txt' from '../source_repo'. - The path 'code.txt' does not exist in the target repository '../source_repo' neither as an output nor a git-handled file.
```
Note that `code.txt` does exist and is committed to git. DVC successfully imports it if the subdirectory `some_dir` exists.

skshetry:

Fixes #3338

b07a982 fixes exception message, i.e. removes `stage working or file path` confusion. It's either of those two now.

### DIFF
--- a/dvc/stage.py
+++ b/dvc/stage.py
@@ -75,21 +75,15 @@ class StageFileBadNameError(DvcException):
 
 
 class StagePathOutsideError(DvcException):
-    def __init__(self, path):
-        msg = "stage working or file path '{}' is outside of DVC repo"
-        super().__init__(msg.format(path))
+    pass
 
 
 class StagePathNotFoundError(DvcException):
-    def __init__(self, path):
-        msg = "stage working or file path '{}' does not exist"
-        super().__init__(msg.format(path))
+    pass
 
 
 class StagePathNotDirectoryError(DvcException):
-    def __init__(self, path):
-        msg = "stage working or file path '{}' is not directory"
-        super().__init__(msg.format(path))
+    pass
 
 
 class StageCommitError(DvcException):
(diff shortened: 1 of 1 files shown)


dvc/repo/imp_url.py
```diff
@@ -10,6 +14,15 @@ def imp_url(self, url, out=None, fname=None, erepo=None, locked=True):
...
     out = resolve_output(url, out)
 
+    out_dir = os.path.dirname(out or "")
+    # try creating an output directory if it does not already exist
+    if (
+        out_dir
+        and path_isin(os.path.abspath(out_dir), self.root_dir)
+        and not os.path.exists(out_dir)
+    ):
+        makedirs(out_dir, exist_ok=True)
```
efiop: @skshetry But not creating a dir is expected behavior. Same as when you cp/mv something, it doesn't pre-create things for you. In the issue we've discussed that this is expected behaviour, just need to look into the error message about stage path not existing

skshetry: @efiop, to be more clear, the output should have been:
`file <file> does not exist`, right? There should not be any mention of  `stage working` at all, right?

skshetry: maybe,`Stage.create(wdir=os.curdir, ...)` is what I'm looking for then.

efiop: @skshetry IIRC, we've discussed this ticket as the one that likely just needs better error messages. `stage working or file path` clearly is broken.

dvc/stage.py
```diff
@@ -557,7 +560,7 @@ def create(repo, accompany_outs=False, **kwargs):
         else:
             path = os.path.abspath(fname)
 
-        Stage._check_stage_path(repo, wdir)
+        Stage._check_stage_path(repo, wdir, is_wdir=kwargs.get("wdir"))
```
skshetry: This will only throw error for `dvc.run()` cases where `wdir` is explicitly set.

PR #3518, opened 2020-03-20
[WIP] remote: verify link support with a temp file

skshetry:

####  TODO
- [ ] Manually check with SSH and local storage if I have broken anything.
- [ ] See if hardlink/symlink is really getting verified.
- [ ] See if I haven't broken s3 for the very least (as I have changed `_cache_is_copy`).
- [ ] Try to cover the change.

### DIFF
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -376,25 +376,29 @@ class RemoteBASE(object):
         logger.debug("'%s' hasn't changed.", path_info)
         return False
 
-    def link(self, from_info, to_info):
-        self._link(from_info, to_info, self.cache_types)
+    def link(self, from_info, to_info, link_types=None):
+        """
+         Pass link_types if link needs to be of different type
+        than specified in user's config or default cache types
+        """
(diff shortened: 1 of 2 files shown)


skshetry:
I was stalling on this PR, because, I was not really sure if I wanted to move ahead with this (i.e. trying to create a temp file before to check for support).

But, looks like it's not that bad (the problem is that due to the optimizations on hardlink, we might not really know if something is verified before we add 100s of empty files).

skshetry:
I closed this because this does not really work on all times. This used temporary file to see if the links are available which does not really work for files across the filesystems.

We could use the file from workspace, but, it does not solve the problem completely (we still have edgecases, see https://github.com/iterative/dvc/issues/3567).
Or, we could try each one out every time, but it'll hit the performance.

We (@efiop and I) discussed that it would be better to handle during save and checkout, i.e trying to find suitable link before saving/checking out (for directory, only check for the one file and assume it's checked for others, whereas for file, try each one out).

We might have a problem with `_cache_is_copy()`, but we can check what the cache type was before hand as well.

PR #3520, opened 2020-03-23
pull/checkout: fix empty log output if no state gets changed

skshetry:

### DIFF
--- a/dvc/command/checkout.py
+++ b/dvc/command/checkout.py
@@ -31,7 +31,7 @@ def log_summary(stats):
         for state in states
         if stats.get(state)
     )
-    logger.info(_human_join(summary))
+    logger.info(_human_join(summary) or "No changes.")
 
 
 def log_changes(stats):
--- a/tests/unit/command/test_checkout.py
+++ b/tests/unit/command/test_checkout.py
@@ -47,7 +47,7 @@ def test_log_summary(caplog):
     del stats["modified"]
     _assert_output(stats, "3 added")
 
-    _assert_output({}, "")
+    _assert_output({}, "No changes.")
 
 
 def test_log_changes(caplog):


dvc/command/checkout.py
```diff
@@ -26,12 +26,13 @@ def _human_join(words):
...
-    summary = (
+    summary = [
         "{} {}".format(len(stats[state]), state)
         for state in states
         if stats.get(state)
-    )
-    logger.info(_human_join(summary))
+    ]
+    if summary:
+        logger.info(_human_join(summary))
```
efiop: Maybe it is worth putting some message that will say that there was nothing to do?

skshetry: Makes sense. Thanks.

tests/unit/command/test_checkout.py
```diff
@@ -47,7 +47,7 @@ def _assert_output(stats, expected_text):
     del stats["modified"]
     _assert_output(stats, "3 added")
 
-    _assert_output({}, "")
```
skshetry: This was a :snake::oil_drum: assertion. :slightly_frowning_face:

skshetry:
We should fix `pull` in this next sprint. The output will be very confusing since `fetch` won't print a thing if fetched something.

#### When nothing was fetched and not checked out:
```
$ dvc pull
No changes.
Everything is up to date.
```

#### When something was fetched, but not checked out.
```
$ dvc pull
No changes.
```

#### When something was fetched + checked out:
```
$ dvc pull
3 added
```

#### When something was not fetched, but checkout was done:
```
$ dvc pull
3 added
Everything is up to date.
```

efiop:
@skshetry Please create a ticket and add it to the next sprint.

efiop:
@skshetry Also, what about docs?

PR #3524, opened 2020-03-24
s3/gs: do not try to make directory on top level path

### LINKED ISSUE #3523: Failure when using S3 remote cache
Trying to run the command

`dvc run -d s3://mybucket/data.txt -o s3://mybucket/data_copy.txt aws s3 cp s3://mybucket/data.txt s3://mybucket/data_copy.txt`

Using AWS S3 remote cache.

It fails with the error message `ERROR: unexpected error - Parameter validation failed: Invalid length for parameter Key, value: 0, valid range: 1-inf`

Attached is the log with `-v` turned on.

[log.txt](https://github.com/iterative/dvc.org/files/4370652/log.txt)

and out of `pip freeze`

[pip_freeze.txt](https://github.com/iterative/dvc/files/4370722/pip_freeze.txt)

skshetry:

This resolves #3523. But, we should try not to make directory path objects if we really don't need to.

### DIFF
--- a/tests/func/test_s3.py
+++ b/tests/func/test_s3.py
@@ -2,6 +2,7 @@ from functools import wraps
 
 import boto3
 import moto.s3.models as s3model
+import pytest
 from moto import mock_s3
 
 from dvc.remote.s3 import RemoteS3
@@ -43,6 +44,30 @@ def test_copy_singlepart_preserve_etag():
     s3.put_object(Bucket=from_info.bucket, Key=from_info.path, Body="data")
 
     RemoteS3._copy(s3, from_info, to_info, {})
+
+
+@mock_s3
+@pytest.mark.parametrize(
+    "base_info",
+    [RemoteS3.path_cls("s3://bucket/"), RemoteS3.path_cls("s3://bucket/ns/")],
+)
+def test_link_created_on_non_nested_path(base_info, tmp_dir, dvc, scm):
+    remote = RemoteS3(dvc, {"url": str(base_info.parent)})
+    remote.s3.create_bucket(Bucket=base_info.bucket)
+    remote.s3.put_object(
+        Bucket=base_info.bucket, Key=(base_info / "from").path, Body="data"
+    )
+    remote.link(base_info / "from", base_info / "to")
+
+    assert remote.exists(base_info / "from")
+    assert remote.exists(base_info / "to")
+
+
+@mock_s3
+def test_makedirs_doesnot_try_on_top_level_paths(tmp_dir, dvc, scm):
+    base_info = RemoteS3.path_cls("s3://bucket/")
+    remote = RemoteS3(dvc, {"url": str(base_info)})
+    remote.makedirs(base_info)
 
 
 def _upload_multipart(s3, Bucket, Key):
--- a/tests/unit/remote/test_remote.py
+++ b/tests/unit/remote/test_remote.py
@@ -33,8 +33,10 @@ def test_remote_without_checksum_jobs_default(dvc):
 def test_makedirs_not_create_for_top_level_path(remote_cls, mocker):
     url = "{.scheme}://bucket/".format(remote_cls)
     remote = remote_cls(None, {"url": url})
-    exc = Exception("should not try to create dir on top-level path")
-    client = mocker.PropertyMock(side_effect=exc)
+    mocked_client = mocker.PropertyMock()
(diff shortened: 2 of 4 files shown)


tests/func/test_s3.py
```diff
@@ -45,6 +46,30 @@ def test_copy_singlepart_preserve_etag():
...
+    assert remote.exists(base_info / "to")
+
+
+@mock_s3
+def test_makedirs_doesnot_try_on_top_level_paths(tmp_dir, dvc, scm):
+    base_info = RemoteS3.path_cls("s3://bucket/")
+    remote = RemoteS3(dvc, {"url": str(base_info)})
+    remote.makedirs(base_info)
+
+
```
efiop: Not testing the gs though :) And also maybe it is not worth the full blown func test (though don't want to waste it too). Maybe let's just add unit tests for RemoteS3/GS.makedirs methods that ensure that it doesn't try to create paths with empty `path`(pretty much same as `mkdir -p /` does nothing successfully)?

skshetry: Yeah, the test is not really related to the issue here.. But, as we are already here, I thought, we should at least cover our public apis (and, it's mocked s3 anyway).

skshetry: I'll add `s3` and `gs` related unittest for `makedirs`.

tests/unit/remote/test_remote.py
```diff
@@ -25,3 +27,14 @@ def test_remote_without_checksum_jobs_default(dvc):
...
+
+
+@pytest.mark.parametrize("remote_cls", [RemoteGS, RemoteS3])
+def test_makedirs_not_create_for_top_level_path(remote_cls, mocker):
+    url = "{.scheme}://bucket/".format(remote_cls)
+    remote = remote_cls(None, {"url": url})
+    exc = Exception("should not try to create dir on top-level path")
+    client = mocker.PropertyMock(side_effect=exc)
+    # we use remote clients with same name as scheme to interact with remote
+    setattr(remote_cls, remote_cls.scheme, client)
```
pared: In one of his PR's @Suor proposed approach utilizing `funcy`:
`mocker.patch.object(remote_cls, remote_cls.scheme, property(raiser(Exception("should not try to create dir on top-level path"))))` .

We might want to consider this approach, but its a matter of taste, so don't consider it as a change request.

tests/unit/remote/test_remote.py
```diff
@@ -25,3 +27,14 @@ def test_remote_without_checksum_jobs_default(dvc):
 
     remote = Remote(dvc, name="without_checksum_jobs")
     assert remote.checksum_jobs == remote.CHECKSUM_JOBS
+
+
+@pytest.mark.parametrize("remote_cls", [RemoteGS, RemoteS3])
+def test_makedirs_not_create_for_top_level_path(remote_cls, mocker):
+    url = "{.scheme}://bucket/".format(remote_cls)
+    remote = remote_cls(None, {"url": url})
+    exc = Exception("should not try to create dir on top-level path")
```
pared: How about newline here? Would separate client exception setup from "others".

tests/unit/remote/test_remote.py
efiop: Why not just mock and `assert not mock.called`? Is there a reason we try to use setattr hack instead of simply using regular mocker facilities?

skshetry: Should be good now. :)

tests/unit/remote/test_remote.py
skshetry: Added (maybe too much). :)

tests/unit/remote/test_remote.py
```diff
@@ -25,3 +27,16 @@ def test_remote_without_checksum_jobs_default(dvc):
...
+@pytest.mark.parametrize("remote_cls", [RemoteGS, RemoteS3])
+def test_makedirs_not_create_for_top_level_path(remote_cls, mocker):
+    url = "{.scheme}://bucket/".format(remote_cls)
+    remote = remote_cls(None, {"url": url})
+    mocked_client = mocker.PropertyMock()
+    # we use remote clients with same name as scheme to interact with remote
+    mocker.patch.object(remote_cls, remote.scheme, mocked_client)
+
+    remote.makedirs(remote.path_info)
+
```
skshetry:
```suggestion
```

PR #3532, opened 2020-03-25
remote: use progress bar when paginating

### LINKED ISSUE #3526: remote: use progress bar when paginating
Followup for #3501

pmrowla:

* During size estimation, status is displayed via `BAR_FMT_NOTOTAL` progress bar, as current # of estimated total remote size
* During threaded traverse, status is displayed via a single progress bar, as # of objects fetched out of the estimated total remote size

Will close #3526

### DIFF
--- a/dvc/remote/azure.py
+++ b/dvc/remote/azure.py
@@ -97,10 +97,9 @@ class RemoteAZURE(RemoteBASE):
                 bucket, prefix=prefix, marker=next_marker
             )
 
-            if progress_callback:
-                progress_callback(len(blobs))
-
             for blob in blobs:
+                if progress_callback:
+                    progress_callback()
                 yield blob.name
 
             if not blobs.next_marker:
--- a/dvc/remote/azure.py
+++ b/dvc/remote/azure.py
@@ -29,6 +29,7 @@ class RemoteAZURE(RemoteBASE):
     REQUIRES = {"azure-storage-blob": "azure.storage.blob"}
     PARAM_CHECKSUM = "etag"
     COPY_POLL_SECONDS = 5
+    LIST_OBJECT_PAGE_SIZE = 5000
 
     def __init__(self, repo, config):
         super().__init__(repo, config)
@@ -88,13 +89,16 @@ class RemoteAZURE(RemoteBASE):
         logger.debug("Removing {}".format(path_info))
         self.blob_service.delete_blob(path_info.bucket, path_info.path)
 
-    def _list_paths(self, bucket, prefix):
+    def _list_paths(self, bucket, prefix, progress_callback=None):
         blob_service = self.blob_service
         next_marker = None
         while True:
             blobs = blob_service.list_blobs(
                 bucket, prefix=prefix, marker=next_marker
             )
+
+            if progress_callback:
+                progress_callback(len(blobs))
 
             for blob in blobs:
                 yield blob.name
@@ -104,14 +108,16 @@ class RemoteAZURE(RemoteBASE):
 
             next_marker = blobs.next_marker
 
-    def list_cache_paths(self, prefix=None):
+    def list_cache_paths(self, prefix=None, progress_callback=None):
         if prefix:
             prefix = posixpath.join(
                 self.path_info.path, prefix[:2], prefix[2:]
             )
         else:
             prefix = self.path_info.path
-        return self._list_paths(self.path_info.bucket, prefix)
+        return self._list_paths(
+            self.path_info.bucket, prefix, progress_callback
+        )
 
     def _upload(
         self, from_file, to_info, name=None, no_progress_bar=False, **_kwargs
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -850,15 +850,14 @@ class RemoteBASE(object):
         with Tqdm(
             desc="Estimating size of "
             + ("cache in '{}'".format(name) if name else "remote cache"),
-            bar_format=Tqdm.BAR_FMT_NOTOTAL,
-            unit="objects",
+            unit="file",
         ) as pbar:
             remote_checksums = set(
                 map(
                     self.path_to_checksum,
                     self.list_cache_paths(
                         prefix=prefix,
-                        progress_callback=lambda n: pbar.update(
+                        progress_callback=lambda n=1: pbar.update(
                             n * total_prefixes
                         ),
                     ),
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -691,7 +691,7 @@ class RemoteBASE(object):
     def checksum_to_path_info(self, checksum):
         return self.path_info / checksum[0:2] / checksum[2:]
 
-    def list_cache_paths(self, prefix=None):
+    def list_cache_paths(self, prefix=None, progress_callback=None):
         raise NotImplementedError
 
     def all(self):
@@ -892,11 +892,11 @@ class RemoteBASE(object):
             return list(checksums & set(self.all()))
 
         return self._cache_exists_traverse(
-            checksums, remote_checksums, jobs, name
+            checksums, remote_checksums, remote_size, jobs, name
         )
 
     def _cache_exists_traverse(
-        self, checksums, remote_checksums, jobs=None, name=None
+        self, checksums, remote_checksums, remote_size, jobs=None, name=None
     ):
         logger.debug(
             "Querying {} checksums via threaded traverse".format(
@@ -912,20 +912,21 @@ class RemoteBASE(object):
             ]
         with Tqdm(
             desc="Querying "
-            + ("cache in " + name if name else "remote cache"),
-            total=len(traverse_prefixes),
-            unit="dir",
+            + ("cache in '{}'".format(name) if name else "remote cache"),
+            total=remote_size,
+            initial=len(remote_checksums),
+            unit="objects",
         ) as pbar:
 
             def list_with_update(prefix):
-                ret = map(
+                return map(
                     self.path_to_checksum,
-                    list(self.list_cache_paths(prefix=prefix)),
+                    list(
+                        self.list_cache_paths(
+                            prefix=prefix, progress_callback=pbar.update
+                        )
+                    ),
                 )
-                pbar.update_desc(
-                    "Querying cache in '{}'".format(self.path_info / prefix)
-                )
-                return ret
 
             with ThreadPoolExecutor(max_workers=jobs or self.JOBS) as executor:
                 in_remote = executor.map(list_with_update, traverse_prefixes,)
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -852,17 +852,15 @@ class RemoteBASE(object):
             + ("cache in '{}'".format(name) if name else "remote cache"),
             unit="file",
         ) as pbar:
-            remote_checksums = set(
-                map(
-                    self.path_to_checksum,
-                    self.list_cache_paths(
-                        prefix=prefix,
-                        progress_callback=lambda n=1: pbar.update(
-                            n * total_prefixes
-                        ),
-                    ),
-                )
+
+            def update(n=1):
+                pbar.update(n * total_prefixes)
+
+            paths = self.list_cache_paths(
+                prefix=prefix, progress_callback=update
             )
+            remote_checksums = set(map(self.path_to_checksum, paths))
+
         if remote_checksums:
             remote_size = total_prefixes * len(remote_checksums)
         else:
@@ -932,14 +930,10 @@ class RemoteBASE(object):
         ) as pbar:
 
             def list_with_update(prefix):
-                return map(
-                    self.path_to_checksum,
-                    list(
-                        self.list_cache_paths(
-                            prefix=prefix, progress_callback=pbar.update
-                        )
-                    ),
+                paths = self.list_cache_paths(
+                    prefix=prefix, progress_callback=pbar.update
                 )
+                return map(self.path_to_checksum, list(paths))
 
             with ThreadPoolExecutor(max_workers=jobs or self.JOBS) as executor:
                 in_remote = executor.map(list_with_update, traverse_prefixes,)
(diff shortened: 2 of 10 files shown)


dvc/remote/oss.py
```diff
@@ -90,20 +91,28 @@ def remove(self, path_info):
...
+        # oss2.ObjectIterator lacks any convenient way to iterate over pages
+        # instead of the flattened list of blobs
+        count = 0
+        iterator = oss2.ObjectIterator(self.oss_service, prefix=prefix)
+        for blob in iterator:
             yield blob.key
+            if progress_callback:
+                count = (count + 1) % iterator.max_keys
+                if count == 0:
+                    progress_callback(iterator.max_keys)
```
pmrowla: this could be cleaner if we subclassed `oss2.ObjectIterator` ourselves so that we could do something like
```py
for page in OSSPageIterator():
    progress_callback(len(page.blobs))
    for blob in page.blobs:
        yield blob.key
```

but I was not sure if making `oss2` a top level import would be ok, since it's currently only imported for `RemoteOSS._list_paths()` and `RemoteOSS.oss_service()`

see: https://github.com/aliyun/aliyun-oss-python-sdk/blob/0ad0a0aa0bf31c94943f8c2b96d6458d01805254/oss2/iterators.py#L84

pmrowla:
for gdrive remote w/~300k-400k files:
[asciicast](https://asciinema.org/a/ziLpoXPsuCjjwFA73D7quxfg6)

shcheklein:
minor: NO_TOTAL progress bar still looks very similar to the regular one, I first thought there is a bug - wondering if this format could be improved to make it more explicit? cc @casperdcl

shcheklein:
awesome stuff, btw, @pmrowla ... thanks for taking care of UI here!

dvc/remote/s3.py
```diff
@@ -202,16 +204,25 @@ def _list_objects(self, path_info, max_items=None, prefix=None):
             kwargs["Prefix"] = posixpath.join(path_info.path, prefix[:2])
         paginator = self.s3.get_paginator(self.list_objects_api)
         for page in paginator.paginate(**kwargs):
-            yield from page.get("Contents", ())
+            contents = page.get("Contents", ())
+            if progress_callback:
+                progress_callback(len(contents))
+            yield from contents
```
casperdcl:
```suggestion
            if progress_callback:
                for item in contents:
                    progress_callback()
                    yield item
            else:
	            yield from contents
```

pmrowla: I was under the impression that it would be more efficient to update the progress bar once per page rather than per item in each page, especially given that the bar is being updated by multiple threads?

dvc/remote/base.py
```diff
@@ -912,20 +926,21 @@ def _cache_exists_traverse(
             ]
         with Tqdm(
             desc="Querying "
-            + ("cache in " + name if name else "remote cache"),
-            total=len(traverse_prefixes),
-            unit="dir",
+            + ("cache in '{}'".format(name) if name else "remote cache"),
+            total=remote_size,
+            initial=len(remote_checksums),
```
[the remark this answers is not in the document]
pmrowla: Some checksums are fetched/processed during the size estimation step, and then are passed into this function as `remote_checksums`. So we are only processing the remaining `remote_size - len(remote_checksums)` checksums

dvc/remote/s3.py
casperdcl: no, all the efficiency logic (skipping display etc) is handled by the bar itself.

dvc/remote/base.py
```diff
@@ -691,7 +691,7 @@ def path_to_checksum(self, path):
     def checksum_to_path_info(self, checksum):
         return self.path_info / checksum[0:2] / checksum[2:]
 
-    def list_cache_paths(self, prefix=None):
+    def list_cache_paths(self, prefix=None, progress_callback=None):
```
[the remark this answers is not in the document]
pmrowla: `progress_callback` is unused in ssh and local remotes, but the parameter was added to both so that overridden `list_cache_paths()` in ssh/local still match the abstract method defined in `RemoteBASE.list_cache_paths()`

dvc/remote/base.py
```diff
@@ -847,9 +847,23 @@ def cache_exists(self, checksums, jobs=None, name=None):
...
+            unit="objects",
+        ) as pbar:
+            remote_checksums = set(
+                map(
+                    self.path_to_checksum,
+                    self.list_cache_paths(
+                        prefix=prefix,
+                        progress_callback=lambda n: pbar.update(
+                            n * total_prefixes
+                        ),
```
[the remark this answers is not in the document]
pmrowla: If I use `unit_scale=total_prefixes` it scales the value of `n` properly after each `update()`, but this also causes the pbar to stop displaying the condensed metric prefixed values. Meaning the bar shows the total as `100000` rather than `1.0M` (even when I explicitly set `unit_divisor=1000`) - https://asciinema.org/a/XeDlDCqZoQyuDxgPsBvjtQQHV?t=12

Is there a way to make the units scale by `total_prefixes` (4096 in this example) but still keep the behavior to condense the displayed number by thousand/million/etc

dvc/remote/base.py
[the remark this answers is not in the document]
casperdcl: Did those `initial` things get a progressbar earlier? If yes, then this is fine. If not (because they were super quick) then we should probably instead just use:
```suggestion
            total=remote_size - len(remote_checksums),
```

dvc/remote/base.py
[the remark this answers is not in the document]
casperdcl: yes but presumably it should be used now that it's added?

dvc/remote/azure.py
```diff
@@ -29,6 +29,7 @@ class RemoteAZURE(RemoteBASE):
     REQUIRES = {"azure-storage-blob": "azure.storage.blob"}
     PARAM_CHECKSUM = "etag"
     COPY_POLL_SECONDS = 5
+    LIST_OBJECT_PAGE_SIZE = 5000
```
casperdcl: unused?

dvc/remote/hdfs.py
```diff
@@ -166,10 +166,13 @@ def list_cache_paths(self, prefix=None):
...
             while dirs:
                 try:
-                    for entry in hdfs.ls(dirs.pop(), detail=True):
+                    entries = hdfs.ls(dirs.pop(), detail=True)
+                    for entry in entries:
                         if entry["kind"] == "directory":
                             dirs.append(urlparse(entry["name"]).path)
                         elif entry["kind"] == "file":
+                            if progress_callback:
+                                progress_callback.update()
```
casperdcl:
```suggestion
                                progress_callback()
```

dvc/remote/oss.py
```diff
@@ -38,6 +38,7 @@ class RemoteOSS(RemoteBASE):
     REQUIRES = {"oss2": "oss2"}
     PARAM_CHECKSUM = "etag"
     COPY_POLL_SECONDS = 5
+    LIST_OBJECT_PAGE_SIZE = 100
```
casperdcl: unused?

tests/unit/remote/test_base.py
```diff
@@ -8,6 +8,16 @@
 from dvc.remote.base import RemoteMissingDepsError
 
 
+class _CallableOrNone(object):
+    """Helper for testing if object is callable() or None."""
+
+    def __eq__(self, other):
+        return True if other is None else callable(other)
```
casperdcl:
```suggestion
        return other is None or callable(other)
```

dvc/remote/base.py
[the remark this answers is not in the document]
casperdcl: hmm no, as of now `unit_divisor` is ignored unless `unit_scale=True`.

dvc/remote/azure.py
pmrowla: It gets used in `RemoteBASE`, it's overridden here since azure uses a page size that is different than the default (1000)

dvc/remote/base.py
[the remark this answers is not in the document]
pmrowla: For the size estimation step, we fetch `initial` items, to get the estimated total (len(initial) * 4096`) and display the status of that estimation with the no-total progressbar. And then here we are displaying the status of fetching the remaining items out of the estimated total

dvc/remote/azure.py
casperdcl: ah my bad; I didn't notice the indentation - thought it was a top-level constant rather than a class member.

dvc/remote/oss.py
casperdcl: class member not global constant

dvc/remote/base.py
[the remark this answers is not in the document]
pmrowla: both local and ssh have their own overridden `cache_exists()` implementation and neither one uses `list_cache_paths()` at all during `cache_exists()`. The only time `list_cache_paths()` would get called for either local or ssh remotes would be for `RemoteBASE.all()` during a `dvc gc` call.

`all()` does not currently display any progressbars at all, but now that I'm thinking about this, it probably should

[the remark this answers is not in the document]
casperdcl: indeed; though I'm not sure in principle about having a function exists which takes a non-null `progress_callback` and ignores it. This implies a bar will be displaying hanging at 0% on screen and then eventually suddenly disappear.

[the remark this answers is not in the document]
pmrowla: `list_cache_paths()` for local and ssh have been updated to use `progress_bar` (if it is set) to keep them consistent with the other remotes. A separate issue (#3543) has been filed for updating the `dvc gc`/`RemoteBASE.all()` behavior and will be addressed in a follow-up PR.

pmrowla:
[asciicast](https://asciinema.org/a/Kd7oTIsotxvgX0i9YRgOfXgXC?t=6:30)

The estimation progress bar still looks similar to the remote cache query progress bar, but given that the estimation status is just calculating the initial and total values for the cache query progress bar, it seems ok to me.

I'm not sure if it would make much of a difference to the user whether the estimation and main cache query appears to be a single operation or as two explicitly distinct operations.

shcheklein:
> The estimation progress bar still looks similar to the remote cache query progress bar

do you mean that it looks like an actual progress bar but in this case we don't show any progress?

shcheklein:
from what I see from code we do show actual progress now, right?

so, what is your concern, Peter? they are similar, and it looks totally fine to me ... may be I'm missing something.

pmrowla:
> from what I see from code we do show actual progress now, right?
> 
> so, what is your concern, Peter? they are similar, and it looks totally fine to me ... may be I'm missing something.

My concern was only that it still hasn't really changed since what you noted for the original video.

I think what you are noting as "showing actual progress now" is more a result of the performance differences between s3 and gdrive. For reference, the current code w/gdrive looks like:
[asciicast](https://asciinema.org/a/xYll5BAmA8874IEFZCTuVd0y8)

shcheklein:
yes, I see, it doesn't indeed show the progress itself - it's impossible since we don't know what the size is, right? we'll be able to do this though when we implement the short cutting in your other PR? at least we'll have an upper bound of how many page we want to go through

is there some other quick style we can apply - like spinner or something?

pmrowla:
Yes, after the short-cutting PR we would have an upper-bound for the `push`/`pull` cache size estimation step and could show progress that way. However for `dvc gc`/`RemoteBASE.all()` (#3543) which always needs to fetch all of the remote cache entries, we would be back to having no upper-bound on the size estimation step

shcheklein:
@pmrowla make sense, yep. 

@casperdcl is there a spinner-like mode for TQDM? :)

It looks like it's a bit more general problem that we should probably solve in a separate PR? looks like we just need a UX component for such cases, most likely we'll have more applications for it. @pmrowla @efiop what do you think? 

I believe there are lot of project we can even use (though we wan't be able to show units in a nice way?) - https://github.com/pavdmyt/yaspin. That's why it would be great to have a mode like this in TQDM itself.

dvc/remote/base.py
```diff
@@ -847,9 +847,22 @@ def cache_exists(self, checksums, jobs=None, name=None):
...
+        ) as pbar:
+            remote_checksums = set(
+                map(
+                    self.path_to_checksum,
+                    self.list_cache_paths(
+                        prefix=prefix,
+                        progress_callback=lambda n=1: pbar.update(
+                            n * total_prefixes
+                        ),
+                    ),
```
efiop: Would extract this into a separate var and use it instead. There are too many levels of brackets in this set(map(... line :slightly_smiling_face: Same down below.

dvc/remote/base.py
```diff
@@ -912,20 +925,21 @@ def _cache_exists_traverse(
...
                     self.path_to_checksum,
-                    list(self.list_cache_paths(prefix=prefix)),
-                )
-                pbar.update_desc(
-                    "Querying cache in '{}'".format(self.path_info / prefix)
+                    list(
+                        self.list_cache_paths(
+                            prefix=prefix, progress_callback=pbar.update
+                        )
+                    ),
```
efiop: let's move `list` into a separate variable and then `map(self.path_to_checksum, paths)`. Just to make it easier to read :slightly_smiling_face:

dvc/remote/base.py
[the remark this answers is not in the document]
casperdcl: btw just to clarify - every time I've said "unused?" it's more a reminder to address at some point (could be in a future PR) rather than an expectation to address in this PR.

casperdcl:
@shcheklein there's no spinner mode but I think the "no total" progress stats are much better than a spinner. I guess you'd like both the stats and a spinner? The updating stats function as a spinner to me :)

shcheklein:
@casperdcl yep, I agree that it's definitely not critical! ... like I mentioned, the only reason for my concern was that since they are so similar and happen one after another, you kind-a expect to see progress happening (visual part of it - where it's emptiness now) - but when you don't see it you start asking yourself a question - is still at 0% 😱 :)

Maybe just let it align the stats close to the text? To make it more explicit that visual part is not expected.

casperdcl:
:man_shrugging: no strong opinions here

PR #3534, opened 2020-03-25
diff: analyse working tree when dirty

### LINKED ISSUE #3386: diff: compare dirty vs committed workspace by default
By default `dvc diff` compares workspace to HEAD (or to a single commit if only one arg is provided). This is the one of the most important use cases. I would like to see the "uncommitted" (in a DVC-sense - when we update md5) changes to data files and directories:

```
git clone https://github.com/iterative/example-get-started
cd example-get-started
dvc pull
rm -f data/data.xml
touch data/features/foo
dvc diff
```

The last command produces an empty result. Not very informative to compare my "current" experiment with the baseline.

UPDATE:

Another interesting point that `dvc status` does exactly this - it analyzes actual changes.

pared:

Fixes #3386

### DIFF
--- a/tests/func/test_diff.py
+++ b/tests/func/test_diff.py
@@ -159,21 +159,51 @@ def test_directories(tmp_dir, scm, dvc):
 
 
 def test_diff_no_cache(tmp_dir, scm, dvc):
-    (stage,) = tmp_dir.dvc_gen(
-        {"dir": {"file": "file content"}}, commit="first"
-    )
+    tmp_dir.dvc_gen({"dir": {"file": "file content"}}, commit="first")
     scm.tag("v1")
+
     tmp_dir.dvc_gen(
         {"dir": {"file": "modified file content"}}, commit="second"
     )
+    scm.tag("v2")
 
-    remove(first(stage.outs).cache_path)
-    remove("dir")
+    remove(dvc.cache.local.cache_dir)
 
     # invalidate_dir_info to force cache loading
     dvc.cache.local._dir_info = {}
 
-    diff = dvc.diff("v1")
+    diff = dvc.diff("v1", "v2")
     assert diff["added"] == []
     assert diff["deleted"] == []
     assert first(diff["modified"])["path"] == os.path.join("dir", "")
+
+
+def test_diff_dirty(tmp_dir, scm, dvc):
+    tmp_dir.dvc_gen(
+        {"file": "file_content", "dir": {"dir_file1": "dir file content"}},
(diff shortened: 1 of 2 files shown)


tests/func/test_diff.py
```diff
@@ -163,17 +163,49 @@ def test_diff_no_cache(tmp_dir, scm, dvc):
...
 
-    remove(first(stage.outs).cache_path)
-    remove("dir")
+    remove(dvc.cache.local.cache_dir)
 
     # invalidate_dir_info to force cache loading
     dvc.cache.local._dir_info = {}
 
-    diff = dvc.diff("v1")
+    diff = dvc.diff("v1", "v2")
```
pared: The change in this test is in order to actually test `diff` when there is no dir cache, and not intermix it with dirty repo `diff`.

tests/func/test_diff.py
```diff
@@ -163,17 +163,49 @@ def test_diff_no_cache(tmp_dir, scm, dvc):
...
+    tmp_dir.dvc_gen("new_file", "new_file_content")
+
+    result = dvc.diff()
+
+    assert result == {
+        "added": [
+            {"hash": "86d049de17c76ac44cdcac146042ec9b", "path": "new_file"}
+        ],
+        "deleted": [
+            {"hash": "7f0b6bb0b7e951b7fd2b2a4a326297e1", "path": "file"}
```
pared: I have a problem with `deleted`, because user will not be able to commit stage file when there is no output. But I presumed that its better to just log `deleted` instead some ambiguous warning saying "Output is deleted but stage file still exists, so you will not be able to commit current state of `file`".

pmrowla: `deleted` seems ok to me, since it's consistent with what `dvc status` reports for this case
```
file.dvc:
        changed outs:
                deleted:            file
```

jorgeorpinel:
Hi! I don't completely understand this change but I'm pretty sure it requires some changes to the docs, right?

efiop:
@jorgeorpinel
> It defaults to comparing the current workspace and the last commit (HEAD), if arguments a_rev and b_rev are not specified.
is still relevant, we're just using actual workspace state here and not only what is recorded in the dvc-file.

This is an interesting topic whether or not we should stick to what is recorded in the dvc files or what is actually lying there (e.g. for cached files for any commit those would match, as cached files are determined by their hash in respective dvc-file, but for non-cached or git files/dirs it might be different), but in this case with dirty workspace we now clearly do the right thing.

PR #3537, opened 2020-03-26
remote: short-circuit remote size estimation for large remotes

### LINKED ISSUE #3530: remote: short-circuit remote size estimation for large remotes
Follow up to #3501 

Original comment:

> one more comment (after watching the demo) - I think what we miss here is also some threshold on how "deep" we want to go with the prefix. Let's imagine we have 256M files and we'd like to push a single file. We will make 1000 requests, right? While what we can do is to see the first 1000 and make some conclusions right away.

Based on the number of local checksums passed into `cache_exists()`, we already know how large of a remote it would require for us to prefer testing if each checksum individually exists in the remote. During the remote size estimation step (when we fetch a entries under a single prefix), we can stop fetching at the number of pages which would put the size of the remote over the threshold (i.e. we only care that `remote_size` is greater than the threshold, the total size is irrelevant).

pmrowla:

Will close #3530

### DIFF
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -845,30 +845,12 @@ class RemoteBASE(object):
         if len(checksums) == 1 or not self.CAN_TRAVERSE:
             return self._cache_object_exists(checksums, jobs, name)
 
-        # Fetch cache entries beginning with "00..." prefix for estimating the
-        # size of entire remote cache
         checksums = frozenset(checksums)
-        prefix = "0" * self.TRAVERSE_PREFIX_LEN
-        total_prefixes = pow(16, self.TRAVERSE_PREFIX_LEN)
-        with Tqdm(
-            desc="Estimating size of "
-            + ("cache in '{}'".format(name) if name else "remote cache"),
-            unit="file",
-        ) as pbar:
 
-            def update(n=1):
-                pbar.update(n * total_prefixes)
-
-            paths = self.list_cache_paths(
-                prefix=prefix, progress_callback=update
-            )
-            remote_checksums = set(map(self.path_to_checksum, paths))
-
-        if remote_checksums:
-            remote_size = total_prefixes * len(remote_checksums)
-        else:
-            remote_size = total_prefixes
-        logger.debug("Estimated remote size: {} files".format(remote_size))
+        # Max remote size allowed for us to use traverse method
+        remote_size, remote_checksums = self._estimate_cache_size(
+            checksums, name=name
+        )
 
         traverse_pages = remote_size / self.LIST_OBJECT_PAGE_SIZE
         # For sufficiently large remotes, traverse must be weighted to account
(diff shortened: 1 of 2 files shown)


pmrowla:
Blocked by #3532 
This PR will need rebase after 3532 as they both touch the same code (but I think it makes sense to keep the UI and size estimation logic changes in separate PRs)

The estimation behavior changes also won't be obvious without the size estimation pbar from the UI PR

pmrowla:
Short-circuit point (which depends on the # of local checksums) is now used as the upper bound for the `Estimating size of ...` progress bar. We would stop estimating at a size of 10M files in this example:

[asciicast](https://asciinema.org/a/RRRatNxdkebhq2vj7N57Ar86z?t=7)

dvc/remote/base.py
```diff
@@ -909,6 +891,65 @@ def update(n=1):
...
+            * self.LIST_OBJECT_PAGE_SIZE,
+        )
+
+    def _estimate_cache_size(self, checksums, short_circuit=True, name=None):
+        """Estimate remote cache size based on number of entries beginning with
+        "00..." prefix.
+        """
+        prefix = "0" * self.TRAVERSE_PREFIX_LEN
+        total_prefixes = pow(16, self.TRAVERSE_PREFIX_LEN)
+        if short_circuit:
```
efiop: Don't see this flag passed anywhere. Or am I missing something?

pmrowla: It's not passed anywhere at the moment, but will be needed for the `dvc gc` progress bar where we have to fetch the full list of entries

dvc/remote/base.py
```diff
@@ -909,6 +891,65 @@ def update(n=1):
...
+            max_remote_size = self._max_estimation_size(checksums)
```
shcheklein: minor: name is a bit confusing - probably should be `max_threshold` or something ... it's not remote size, right?

dvc/remote/base.py
```diff
@@ -909,6 +891,65 @@ def update(n=1):
...
+        else:
+            max_remote_size = None
+
+        with Tqdm(
```
shcheklein: My 2cs:

* let's check please GDrive (other remotes?) that it is indeed lazy (does not prefetch all pages for the prefix)
* short circuitry itself is great, but I would probably keep UI as it was before ... it became even more confusing - when I see 1% - 2% -> done ... let's imagine we will have it counting to 20% and I users would be scared to see it since they would expect to wait until 100%? 
* code wise, not this PR - size of this file got out of hands already .... not sure what is the best practice in Python (are there?), how people support large files in a good shape? or split into helpers, mixins, etc? cc @efiop @Suor

pmrowla: @shcheklein It's lazy for gdrive. The only remote I wasn't sure about is hdfs. For hdfs we `yield` each item returned by `hdfs.ls(<prefix>)`, but I don't think there's much else we can do in that case?

regarding the UI, I think the ideal thing would be to just use the current running estimated size without the total (so the way it was before this change) plus a spinner, as discussed in the other PR

shcheklein: @pmrowla agreed on both! (spinner itself can wait until we have something from tqdm for example, or can do minor expriments - align stats to the left or whatnot).

thanks for looking into this!

PR #3539, opened 2020-03-26
gdrive: use default client and secret id. Debug message on retry of f…

maxhora:
…ailed GDrive API call.

Related to #2865 

- Default client_id and client_secret values in DVC config ( from DVC's GCP project ).
- Validate pair of `client_id` and `client_secret` (both non-default should be specified by user).
- Debug message on retry of failed GDrive API call.

### DIFF
--- a/dvc/remote/gdrive.py
+++ b/dvc/remote/gdrive.py
@@ -46,7 +46,10 @@ def gdrive_retry(func):
             return False
 
         retry_codes = [403, 500, 502, 503, 504]
-        return exc.error.get("code", 0) in retry_codes
+        result = exc.error.get("code", 0) in retry_codes
+        if result:
+            logger.debug("Retry GDrive API call failed with {}.".format(exc))
+        return result
 
     # 15 tries, start at 0.5s, multiply by golden ratio, cap at 20s
     return retry(
@@ -95,6 +98,9 @@ class RemoteGDrive(RemoteBASE):
 
     GDRIVE_CREDENTIALS_DATA = "GDRIVE_CREDENTIALS_DATA"
     DEFAULT_USER_CREDENTIALS_FILE = "gdrive-user-credentials.json"
+
+    DEFAULT_GDRIVE_CLIENT_ID = "710796635688-iivsgbgsb6uv1fap6635dhvuei09o66c.apps.googleusercontent.com"  # noqa: E501
+    DEFAULT_GDRIVE_CLIENT_SECRET = "a1Fz59uTpVNeG_VGuSKDLJXv"
 
     def __init__(self, repo, config):
         super().__init__(repo, config)
@@ -153,14 +159,15 @@ class RemoteGDrive(RemoteBASE):
             )
 
         # Validate OAuth 2.0 Client ID configuration
-        if not self._use_service_account and (
-            not self._client_id or not self._client_secret
-        ):
-            raise DvcException(
-                "Please specify Google Drive's client id and "
-                "secret in DVC config. Learn more at "
-                "{}.".format(format_link("https://man.dvc.org/remote/modify"))
-            )
+        if not self._use_service_account:
+            if bool(self._client_id) != bool(self._client_secret):
+                raise DvcException(
+                    "Please specify Google Drive's client id and secret in "
+                    "DVC config or omit both to use defaults. Learn more at "
+                    "{}.".format(
+                        format_link("https://man.dvc.org/remote/modify")
+                    )
+                )
 
     @wrap_prop(threading.RLock())
     @cached_property
@@ -186,8 +193,9 @@ class RemoteGDrive(RemoteBASE):
             }
         else:
             GoogleAuth.DEFAULT_SETTINGS["client_config"] = {
-                "client_id": self._client_id,
-                "client_secret": self._client_secret,
+                "client_id": self._client_id or self.DEFAULT_GDRIVE_CLIENT_ID,
+                "client_secret": self._client_secret
+                or self.DEFAULT_GDRIVE_CLIENT_SECRET,
                 "auth_uri": "https://accounts.google.com/o/oauth2/auth",
                 "token_uri": "https://oauth2.googleapis.com/token",
                 "revoke_uri": "https://oauth2.googleapis.com/revoke",


dvc/remote/gdrive.py
```diff
@@ -46,7 +46,12 @@ def should_retry(exc):
             return False
 
         retry_codes = [403, 500, 502, 503, 504]
-        return exc.error.get("code", 0) in retry_codes
+        result = exc.error.get("code", 0) in retry_codes
+        if result:
+            logger.debug(
+                "Retry GDrive API call failed with {}".format(str(exc))
```
shcheklein: not sure you need `str( )` here (cc @efiop - please TAL)

dvc/remote/gdrive.py
```diff
@@ -153,14 +179,28 @@ def _validate_config(self):
...
-        if not self._use_service_account and (
-            not self._client_id or not self._client_secret
-        ):
-            raise DvcException(
-                "Please specify Google Drive's client id and "
-                "secret in DVC config. Learn more at "
-                "{}.".format(format_link("https://man.dvc.org/remote/modify"))
-            )
+        if not self._use_service_account:
+            if not self._client_id or not self._client_secret:
```
shcheklein: the both are set now, right? remove this check?

dvc/remote/gdrive.py
maxhora: right, seems, is not needed. Have seen such usage elsewhere in DVC.

dvc/remote/gdrive.py
```diff
@@ -69,6 +74,20 @@ def _location(exc):
...
+def _validate_default_credentials(equal_check, diff_check):
+    if equal_check[0] == equal_check[1] and diff_check[0] != diff_check[1]:
+        raise DvcException(
+            "Default value for client {0} is used, "
+            "but client {1} in DVC config is different from default. "
+            "Remove client {1} or set correct client {0} in DVC config. "
+            "Learn more at {2}.".format(
+                equal_check[2],
+                diff_check[2],
+                format_link("https://man.dvc.org/remote/modify"),
```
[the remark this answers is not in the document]
maxhora: this is not relevant anymore (code was completely removed)

dvc/remote/gdrive.py
maxhora: check is changed to "they both must be set only if one them is set"

dvc/remote/gdrive.py
```diff
@@ -121,8 +127,12 @@ def __init__(self, repo, config):
         self._service_account_p12_file_path = config.get(
             "gdrive_service_account_p12_file_path"
         )
-        self._client_id = config.get("gdrive_client_id")
-        self._client_secret = config.get("gdrive_client_secret")
+        self._client_id = config.get(
+            "gdrive_client_id", self.DEFAULT_GDRIVE_CLIENT_ID
```
shcheklein: need to update here to avoid default value

dvc/remote/gdrive.py
```diff
@@ -153,14 +163,17 @@ def _validate_config(self):
...
-        if not self._use_service_account and (
-            not self._client_id or not self._client_secret
-        ):
-            raise DvcException(
-                "Please specify Google Drive's client id and "
-                "secret in DVC config. Learn more at "
-                "{}.".format(format_link("https://man.dvc.org/remote/modify"))
-            )
+        if not self._use_service_account:
+            if (self._client_id and not self._client_secret) or (
```
shcheklein: `bool(one) != bool(second)` to simplify this check

dvc/remote/gdrive.py
```diff
@@ -186,8 +199,9 @@ def drive(self):
             }
         else:
             GoogleAuth.DEFAULT_SETTINGS["client_config"] = {
-                "client_id": self._client_id,
-                "client_secret": self._client_secret,
+                "client_id": self._client_id or self.DEFAULT_GDRIVE_CLIENT_ID,
+                "client_secret": self._client_secret
+                or self.DEFAULT_GDRIVE_CLIENT_SECRET,
```
shcheklein: auto formatted this way?

dvc/remote/gdrive.py
maxhora: yep, that was left accidentally

dvc/remote/gdrive.py
```diff
@@ -46,7 +46,10 @@ def should_retry(exc):
             return False
 
         retry_codes = [403, 500, 502, 503, 504]
-        return exc.error.get("code", 0) in retry_codes
+        result = exc.error.get("code", 0) in retry_codes
+        if result:
+            logger.debug("Retry GDrive API call failed with {}.".format(exc))
```
shcheklein: can we put a counter here (Retrying 2 time, API call failed with ... format(exc)`?

dvc/remote/gdrive.py
maxhora: yes, looks strange

dvc/remote/gdrive.py
maxhora: retry attempt seems is internal variable of funcy.retry decorator.
we can use some global counter variable, but seems it is hard to reset such counter in case if retry attempt was successful; Also not sure if there Python will make a copy of global variable per worker thread, or all threads will reference the same counter and increase incorrectly.

shcheklein: yes, not need for the global one ... let's keep it this way for now ... @Suor - it's a feature request for the future (or might be you know some better way to get this information - more context on what we are retrying and how many times)

dvc/remote/gdrive.py
```diff
@@ -153,14 +159,15 @@ def _validate_config(self):
...
-        ):
-            raise DvcException(
-                "Please specify Google Drive's client id and "
-                "secret in DVC config. Learn more at "
-                "{}.".format(format_link("https://man.dvc.org/remote/modify"))
-            )
+        if not self._use_service_account:
+            if bool(self._client_id) != bool(self._client_secret):
+                raise DvcException(
+                    "Please specify Google Drive's client id and "
```
shcheklein: ... or omit both to use defaults

shcheklein:
@efiop good to merge if tests pass

maxhora:
Travis passed

shcheklein:
@efiop could we please merge and release? :) 🙏

PR #3555, opened 2020-03-31
pull: update help output (-R option)

jorgeorpinel:
"pull from subdirectories" seems misleading. The pull happens from a remote.

### DIFF
--- a/dvc/command/data_sync.py
+++ b/dvc/command/data_sync.py
@@ -190,7 +190,7 @@ def add_parser(subparsers, _parent_parser):
         "--recursive",
         action="store_true",
         default=False,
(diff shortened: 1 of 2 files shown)


dvc/command/data_sync.py
```diff
@@ -190,7 +190,7 @@ def add_parser(subparsers, _parent_parser):
         "--recursive",
         action="store_true",
         default=False,
-        help="Push cache from subdirectories of specified directory.",
+        help="Push cache for subdirectories of specified directory.",
```
efiop: I guess it is still not clear that it will be pushed for dvc files inside of that directory, but it is still better.

efiop:
Thank you! :pray:

PR #3557, opened 2020-03-31
remote: skip non-cache looking files when estimating remote size

efiop:
`dvc push` and other commands will fail if they find a file that doesn't look like a cache file in the remote. The issue is that we've started using `list_cache_paths` for remote size estimation, which doesn't skip non-cache looking files. We should use `all()` instead.

Discord context: https://discordapp.com/channels/485586884165107732/485596304961962003/694454999492591657

### DIFF
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -687,7 +687,7 @@ class RemoteBASE(object):
         parts = self.path_cls(path).parts[-2:]
 
         if not (len(parts) == 2 and parts[0] and len(parts[0]) == 2):
-            raise ValueError("Bad cache file path")
+            raise ValueError("Bad cache file path '{}'".format(path))
 
         return "".join(parts)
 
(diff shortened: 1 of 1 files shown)


dvc/remote/base.py
```diff
@@ -697,17 +697,18 @@ def checksum_to_path_info(self, checksum):
     def list_cache_paths(self, prefix=None, progress_callback=None):
```
efiop: This is pretty much a `walk_files` method, but it is not supported in every remote yet, so we are using a legacy implementation for now.

efiop:
Oops, I think I forgot to change the places that expect cache paths but get hashes. Looking into it.

dvc/remote/base.py
efiop: Using `walk_files` would've been much more intuitive, esp since `walk` presumes that it is an iterator, unlike `list_cache_paths` which could be anything.

dvc/remote/base.py
```diff
@@ -936,13 +935,13 @@ def update(n=1):
                 pbar.update(n * total_prefixes)
 
             if max_remote_size:
-                paths = self._cache_paths_with_max(
+                checksums = self._all_with_limit(
```
efiop: Hm, maybe simply `islice()`?

Issue #3558, opened 2020-03-31
Bad cache file path using Google Cloud Storage remote repo


manuelramirezv:
The problem occurs in the google cloud shell with a python installation via pip in a virtual enviroment.
"dvc version" output:

DVC version: 0.91.2+d4b503                                                                                                                                                                               
Python version: 3.7.3                                                                                                                                                                                    
Platform: Linux-4.19.87+-x86_64-with-debian-9.12                                                                                                                                                         
Binary: False                                                                                                                                                                                            
Package: None
Supported remotes: gs, http, https
Cache: reflink - not supported, hardlink - supported, symlink - supported
Repo: dvc, git

**dvc pull output:**

```
2020-03-30 08:31:49,261 DEBUG: UPDATE state_info SET count = ? WHERE rowid = ?
    ret = cmd.run()
  File "/home/manuel_ramirez/env-dvc/lib/python3.7/site-packages/dvc/command/data_sync.py", line 53, in run
    recursive=self.args.recursive,
  File "/home/manuel_ramirez/env-dvc/lib/python3.7/site-packages/dvc/repo/__init__.py", line 28, in wrapper
    ret = f(repo, *args, **kwargs)
  File "/home/manuel_ramirez/env-dvc/lib/python3.7/site-packages/dvc/repo/push.py", line 25, in push
    return self.cloud.push(used, jobs, remote=remote)
  File "/home/manuel_ramirez/env-dvc/lib/python3.7/site-packages/dvc/data_cloud.py", line 66, in push
    show_checksums=show_checksums,
  File "/home/manuel_ramirez/env-dvc/lib/python3.7/site-packages/dvc/remote/local.py", line 374, in push
    download=False,
  File "/home/manuel_ramirez/env-dvc/lib/python3.7/site-packages/dvc/remote/local.py", line 347, in _process
    download=download,
  File "/home/manuel_ramirez/env-dvc/lib/python3.7/site-packages/dvc/remote/local.py", line 268, in status
    md5s, jobs=jobs, name=str(remote.path_info)
  File "/home/manuel_ramirez/env-dvc/lib/python3.7/site-packages/dvc/remote/base.py", line 851, in cache_exists
    map(self.path_to_checksum, self.list_cache_paths(prefix=prefix))
  File "/home/manuel_ramirez/env-dvc/lib/python3.7/site-packages/dvc/remote/base.py", line 687, in path_to_checksum
    raise ValueError("Bad cache file path")
ValueError: Bad cache file path

[...]

Issue #2329, opened 2019-07-27
ui / logger: rethink verbosity


mroutis:
Here's an old idea that I wanted to experiment with but never made time for it :sweat_smile: , now feel like sharing and spark some discussion  :)

The output of some commands _narrate_ what is happening underneath, this is awesome for someone curious that wants to take a peek behind the curtain, but might result tiring such verbosity when you already know (or just don't care) what the command did. Let's take `dvc add` for example:

```bash
$ dvc add hello

Saving 'hello' to '.dvc/cache/b1/946ac92492d2347c6235b4d2611184'.
Saving information to 'hello.dvc'.
```

My suggestion would be to create different verbose levels, `-vvv` (INFO, DEBUG, DATABASE?) and make the output of the first level serve the purpose to know what's going on:

```bash
==> Adding 'hello' file...
  -> computing checksum: b1946ac92492d2347c6235b4d261118
  -> moving file to cache '.dvc/cache/b1/946ac92492d2347c6235b4d261118'
  -> linking '.dvc/cache/b1/946ac92492d2347c6235b4d261118' to 'hello'
  -> generate stage file 'hello.dvc'
```

There are different levels to group operations:
```
:: H1
==> H2
  -> Content
  ---> SQL queries in case that we need those? Maybe information about inodes?

>> prompt
```

Creating some flexibility on how to organize the output so it make sense
```
:: Add (command / operation)
==> foo (file / target)
  -> Stuff about adding foo
==> bar
  -> Stuff about adding bar
```

Arrows might be a little bit confusing but if you add some color they look really great:
(the idea is taken from the way `pacman`/`yay` organizes its output).

Notice the *bold* on the headers that represent operations occurring during an update:
![2019-07-27-13:52:39](https://user-images.githubusercontent.com/7363250/61998446-d142f800-b075-11e9-8770-c9229faa6136.png)

Also notice that they are not masking the outputs of the download, probably was a `curl`)
![2019-07-27-13:57:40](https://user-images.githubusercontent.com/7363250/61998487-76f66700-b076-11e9-88af-c4f767a0cbab.png)

We can keep the color code for warnings/errors:
![2019-07-27-14:01:27](https://user-images.githubusercontent.com/7363250/61998514-03a12500-b077-11e9-8ffb-a58267d6aa78.png).

This can integrate very well with progress bar.

### Possible implementation
- Keep track of the indentation level on the Logger itself
[...]

PR #3559, opened 2020-03-31
remote: use progress bar for remote cache query status during `dvc gc`

### LINKED ISSUE #3543: remote: use progress bar for remote cache query status during `dvc gc`
As noted in #3532 
> both local and ssh have their own overridden cache_exists() implementation and neither one uses list_cache_paths() at all during cache_exists(). The only time list_cache_paths() would get called for either local or ssh remotes would be for RemoteBASE.all() during a dvc gc call.
>
> all() does not currently display any progressbars at all, but now that I'm thinking about this, it probably should

`RemoteBASE.all()` can use the same remote size estimation behavior as `cache_exists()` so that `dvc gc` displays actual progress/status when iterating over the full list of remote checksums. This will make `gc` behavior consistent with `push`/`pull`

pmrowla:

- `dvc gc -c` now uses the same behavior as `push`/`pull` when fetching checksums from a remote cache
- Progress bar is displayed using the estimated cache size as the total, and checksums are fetched in parallel by prefix
- This also reverts the estimation progress bar total changes as discussed in #3537 - we now just display the running estimated size without using the max limit as total 

Will close #3543

### DIFF
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -722,6 +722,10 @@ class RemoteBASE(object):
                 name if name else "remote cache"
             )
         )
+
+        if not self.CAN_TRAVERSE:
+            return self.cache_checksums()
+
         remote_size, remote_checksums = self._estimate_cache_size(name=name)
         return self._cache_checksums_traverse(
             remote_size, remote_checksums, jobs, name
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -697,12 +697,13 @@ class RemoteBASE(object):
     def list_cache_paths(self, prefix=None, progress_callback=None):
         raise NotImplementedError
 
-    def all(self, *args, **kwargs):
-        # NOTE: The list might be way too big(e.g. 100M entries, md5 for each
-        # is 32 bytes, so ~3200Mb list) and we don't really need all of it at
-        # the same time, so it makes sense to use a generator to gradually
-        # iterate over it, without keeping all of it in memory.
-        for path in self.list_cache_paths(*args, **kwargs):
+    def cache_checksums(self, prefix=None, progress_callback=None):
+        """Iterate over remote cache checksums.
+
+        If `prefix` is specified, only checksums which begin with `prefix`
+        will be returned.
+        """
+        for path in self.list_cache_paths(prefix, progress_callback):
             try:
                 yield self.path_to_checksum(path)
             except ValueError:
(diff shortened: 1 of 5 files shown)


pmrowla:
[asciicast](https://asciinema.org/a/Yjsm1vemX6bUTEEZ6NqE8O73j?t=2:45)

`dvc gc` is still very quiet until it gets to the actual cache query step - it still looks like it's hanging for the first 2:50 in the example

tests/unit/remote/test_base.py
```diff
@@ -94,45 +91,56 @@ def test_cache_exists(path_to_checksum, object_exists, traverse):
...
+
+def test_cache_checksums():
+    remote = RemoteBASE(None, {})
+    remote.path_info = PathInfo("foo")
+
+    with mock.patch.object(
+        remote, "list_cache_paths", return_value=["12/3456", "bar"]
+    ):
+        checksums = list(remote.cache_checksums())
+        assert checksums == ["123456"]
```
pmrowla: this test is for #3558

dvc/remote/base.py
```diff
@@ -697,27 +697,45 @@ def checksum_to_path_info(self, checksum):
...
+        If `prefix` is specified, only checksums which begin with `prefix`
+        will be returned.
+        """
+        for path in self.list_cache_paths(prefix, progress_callback):
             try:
                 yield self.path_to_checksum(path)
             except ValueError:
                 logger.debug(
                     "'%s' doesn't look like a cache file, skipping", path
                 )
```
pmrowla: This is makes push/pull/gc very verbose now to the point that I'm not sure it's worth logging?
```
...
2020-03-31 18:49:52,911 DEBUG: '/Users/pmrowla/git/remote-1m-files/.dvc/cache/c5/d84d5879b6266fe8624802e99ff211.dir.unpacked/000360c8.txt' doesn't look like a cache file, skipping
2020-03-31 18:49:52,935 DEBUG: '/Users/pmrowla/git/remote-1m-files/.dvc/cache/c5/d84d5879b6266fe8624802e99ff211.dir.unpacked/00001f77.txt' doesn't look like a cache file, skipping
2020-03-31 18:49:53,122 DEBUG: '/Users/pmrowla/git/remote-1m-files/.dvc/cache/c5/d84d5879b6266fe8624802e99ff211.dir.unpacked/000569a4.txt' doesn't look like a cache file, skipping
...
```

Prior to #3557 this exception handler was just a `pass` in `RemoteBASE.all()`

tests/unit/remote/test_base.py
efiop: @pmrowla Thank you so much! :pray:

dvc/remote/base.py
efiop: @pmrowla It is verbose in `-v` verbose mode, but at least we know that it does skip it. Maybe it is time to support `-vv` and hide this, "assuming" and state logs under it. Will need to look into it, added a comment to https://github.com/iterative/dvc/issues/2329 .

Issue #3409, opened 2020-02-26
Metrics: separating scalar and continuous/graphs metrics


dmpetrov:
## Problem

There are two types of metrics in ML projects: **scalar metrics** or just a number (like AUC) and **continuous metrics** or a sequence of numbers (like ROC curve - an array of numbers).

Today we support csv and json files as metrics without specifying what metrics are needed. In fact, DVC commands support only scalars: `dvc metrics diff`.

## Solution

It is important to support both, understand the semantics of the metrics and provide more value.

1. For **scalar metrics**, DVC should provide (and it does provide it now) `dvc metrics diff` functionality. We just need clearly file format in the documentation to make sure only scalar metrics are supported. Json/Ini are both good formats for scalar metrics.
2. For **continuous metrics**, DVC should provide plots/graph visualization functionality. Like `dvc viz show roc.json` and `dvc viz diff roc.json HEAD HEAD^^`. Both of the commands should generate graphs. The second command - two graphs (not diff).

## Types

Types might improve visualization a lot. There many possible types of graphs for `dvc viz`. It would be great to support a few common ones like regular plot, confusion matrix.

For scalar types, it is also important to specify a type. A scalar can be the result of the minimization of some function or maximization. If the user can specify this information it might help to show proper color: red if max meterics was decreased and green if increased.

We should not make the metrics file format complicated. So, it might be a better solution to specify metrics format in a separate file (like `Dvcfile` or special metrics file). 

```
- scalars_types:
      AUC: max
      error_1: min
- metrics:
      roc.json: plot
      error.csv: plot
      mx.json: confusion_matrix
```

## Open questions

1. Should we include `dvc viz` in the core dvc package? Problem - most likely it will bring dependencies to some graphics libraries which can cause installation issues in OS without graphics (like default EC2 instances). What options do we have?
2. How we should specify the format of the graphs. Do we need any customization? For example, mlflow supports only one type of graph and cannot show something more advance (like a confusion matrix). If we choose Vega-graph as the format we can support graph customization.
3. The metrics can be related to experiment parameters (see #3393) because we are introducing new DVC file types and can use a single DVC file to preserve some information about metrics or metrics types.

## Actions

- [ ] Modify DVC documentation - only scalar metrics are supported in the current format and DVC commands.
- [x] Implement continuous metrics with `dvc viz` command.
- [ ] Add support of types.

efiop:
[...]

PR #3569, opened 2020-04-01
[WIP] plot metrics: initital

pared:

Related to #3409
As of today: only json metrics structured as vega-compliant data accepted (basically list of `{"x":, "y":}`), due to CORS related issues.

Vega has capabilities of loading the data (and transforming) by itself, but that would require us to set up local server sharing theese files, which seems like overkill, at least in initial version of visualization. So for now we are stuck with injecting display code directly into page code.


dvc/repo/init.py
```diff
@@ -100,6 +101,8 @@ def init(root_dir=os.curdir, no_scm=False, force=False, subdir=False):
         with config.edit() as conf:
             conf["core"]["no_scm"] = True
 
+    init_plot_templates(dvc_dir)
```
efiop, line 104: What if this is a pre-existing repo and there is no plot template?

dvc/command/plot.py
```diff
@@ -0,0 +1,31 @@
...
+
+from dvc.command.base import append_doc_link, CmdBase
+from dvc.utils import format_link
+
+logger = logging.getLogger(__name__)
+
+
+class CmdPlot(CmdBase):
+    def run(self):
+        self.repo.plot(self.args.targets)
```
efiop, line 12: Nitpick: It could raise an exception from somewhere, right? Let's wrap it into try&except, same as we do in other places.

dvc/plot.py
```diff
@@ -0,0 +1,74 @@
...
+            os.path.join(self.plot_templates_dir, self.TEMPLATE_NAME), "r"
+        ) as fd:
+            vega_spec = json.load(fd)
+
+        vega_spec.update(update_dict)
+        return vega_spec
+
+
+def init_plot_templates(dvc_dir):
+
```
efiop, line 72: probably a leftover
```suggestion
```

dvc/plot.py
```diff
@@ -0,0 +1,74 @@
...
+    templates = [DefaultTemplate]
+    [t(dvc_dir).dump() for t in templates]
```
efiop, line 74: is there a reason to create this list?

dvc/repo/plot.py
```diff
@@ -0,0 +1,66 @@
...
+def plot(repo, targets, plot_path="plot.html", typ="json"):
+    divs = []
+    for target in targets:
+        data = _load(repo.tree, target)
+        vega_plot_json = DefaultTemplate(repo.dvc_dir).fill(data)
+        divs.append(DivTemplate.prepare_div(vega_plot_json))
+    PageTemplate.save(divs, plot_path)
+    logger.info(
+        "Your can see your plot by opening '{}' in your "
+        "browser!".format(plot_path)
```
efiop, line 65: oh, we could probably `format_link` it as well, for extra style points :slightly_smiling_face:

dvc/repo/plot.py
```diff
@@ -0,0 +1,66 @@
...
+
+def _load(tree, target, revision="current workspace"):
+    with tree.open(target, "r") as fd:
+        data = json.load(fd)
+        for d in data:
+            d["revision"] = revision
+    return data
+
+
+def plot(repo, targets, plot_path="plot.html", typ="json"):
```
efiop, line 56: Should it be `@locked` btw?

dvc/repo/plot.py
```diff
@@ -0,0 +1,66 @@
+import json
+import logging
+
+from dvc.plot import DefaultTemplate
+
+logger = logging.getLogger(__name__)
+
+
+class PageTemplate:
```
efiop, line 9: Seems like class is not really needed here. You are using it as a function anyways. Or am I missing something?

dvc/repo/plot.py
```diff
@@ -0,0 +1,66 @@
...
+</html>"""
+
+    @staticmethod
+    def save(divs, path):
+        page = PageTemplate.HTML.format(divs="\n".join(divs))
+        with open(path, "w") as fd:
+            fd.write(page)
+
+
+class DivTemplate:
```
efiop, line 29: Same here, doesn't seem like you need a class.

dvc/repo/plot.py
pared: definitely!

dvc/repo/plot.py
```diff
@@ -0,0 +1,66 @@
...
+        from shortuuid import uuid
+
+        id = uuid()
+        return DivTemplate.HTML.format(
+            id=str(id),
+            vega_json=json.dumps(vega_dict, indent=4, separators=(",", ": ")),
+        )
+
+
+def _load(tree, target, revision="current workspace"):
```
efiop, line 48: Doesn't look like `revision` is used. You probably plan on using brancher here?

pared: yup

dvc/repo/plot.py
efiop, line 56: Setting `plot_path` to a string is a bad style, use `None` and then use defaults in the code itself.

dvc/repo/plot.py
```diff
@@ -0,0 +1,66 @@
...
+
+def plot(repo, targets, plot_path="plot.html", typ="json"):
+    divs = []
+    for target in targets:
+        data = _load(repo.tree, target)
+        vega_plot_json = DefaultTemplate(repo.dvc_dir).fill(data)
+        divs.append(DivTemplate.prepare_div(vega_plot_json))
+    PageTemplate.save(divs, plot_path)
+    logger.info(
+        "Your can see your plot by opening '{}' in your "
```
efiop, line 64: Currently plot_path is relative by default, which won't be clickable. You need to use something like `file:/abs/path` instead, probably.

dvc/repo/plot.py
```diff
@@ -0,0 +1,66 @@
...
+
+        id = uuid()
+        return DivTemplate.HTML.format(
+            id=str(id),
+            vega_json=json.dumps(vega_dict, indent=4, separators=(",", ": ")),
+        )
+
+
+def _load(tree, target, revision="current workspace"):
+    with tree.open(target, "r") as fd:
```
efiop, line 49: Really minor: let's use `fobj` instead of `fd`. We have `fd`s sprinkled around the code, but it is misleading.

efiop:
@pared This looks great! :pray: Trying it out, I also see that the default resolution is quite crusty, is there a setting for that in the template? Or is it something on my end?

pared:
plot-sample-2020-04-01-150629

pared:
@efiop 
take a look at gif provided by me, you can directly edit the template, to meet your needs, basically what you need to know is how to structure Vega json, and adjust template accordingly.

pared:
@efiop 
> Could it show vector graphics by any chance?
Yes, take note that in the top right corner, there is a button with saving options, you can obtain SVG there.

pmrowla:
> @efiop
> 
> > Could it show vector graphics by any chance?
> 
> Yes, take note that in the top right corner, there is a button with saving options, you can obtain SVG there.

I think we may want to use vega's [responsive sizing](https://vega.github.io/vega-lite/docs/size.html#specifying-responsive-width-and-height) and just make our default template scale automatically to the user's browser window by setting a percentage-based height/width `style` attribute on the vega `div`. (I did notice that using `100%` for the div width/height makes some of the vega UI stuff run off of the screen, so use something smaller than that)

One other thing I noticed is that when you export to SVG, it generates a graphic with a fixed width/height in pixels so that the exported graphic doesn't actually scale. This always seemed to be the case regardless of whether or not the vega json was using default/fixed/responsive sizing. This can also be fixed by manually editing the SVG to use a percentage-based height/width but I'm wondering if there's a way to make vega do that automatically?

pared:
>I'm wondering if there's a way to make vega do that automatically?

@pmrowla 
Thanks for pointing that out.
I'll take note of that. Maybe I missed some settings in template specification. Also, that might be related to the fact that I am using vega-lite currently.

pared:
closing in favor of #3577

Issue #3572, opened 2020-04-01
Support yaml as metrics file


dmpetrov:
Yaml is one of the most used file format. We need to support it for metrics.
More details: https://github.com/iterative/dvc/issues/3409#issuecomment-605444643

ToDo:
- [ ] update docs

dmpetrov:
Ideally, as part of this issue a few other steps should be implemented from the link above:
[2] Add support for YAML since it is an often-used format - #3393 (comment)
[3] With two simple json/yaml formats, do we really need raw format? Let's get rid of raw metrics format.
[4] Do we really need xpath in dvc-files? I see a clear use case in metrics show as a filter but not in other commands. Let's consider removing XPath (from metrics files).
[5] Metrics type was mostly used for distinguishing xpath algorithms between hierarchical json and flat csv/tsv/hcsv/htsv. It looks like it is not needed if only json/yaml left. Let's remove metrics type.

efiop:
@dmpetrov So maybe let's get rid of *sv support, drop explicit XPath and type, support only yaml (could leave json, but is there really a point?) and pretty much do the same thing that we do with params? E.g.
```
dvc run -m all.AUC ... # will assume metrics.yaml, similar to how we assume `params.yaml`
dvc run -m custom_name.yam:all.AUC ...
```
?

dmpetrov:
@efiop agree re the format support (but I'd keep json). The good news is - both yaml and json are hierarchical and we can use the same patterns for filters (like path).

Granular metrics addressation (like `-m all.AUC`) seems a bit too complicated scenario. It has one advantage - multiple stages can append metrics to the same metrics file. But this metrics appending is not easy to implement. We can think about this feature when we have more demand for such scenarios.

jorgeorpinel:
>  update docs

@dmpetrov will you be sending a draft for this or should I include in my regular updates? Seems likt it can be a pretty quick update, just add "YAML" after "JSON" in metrics subcommand references.

jorgeorpinel:
Unless you want to include some special examples.

jorgeorpinel:
I meant for docs

efiop:
@jorgeorpinel Dmitry promised to send a PR :pray: There are some other changes in the way metrics work, so it won't be as easy as adding `YAML` to the docs.

shcheklein:
@jorgeorpinel let's create a ticket!

jorgeorpinel:
OK: iterative/dvc.org/issues/1175

PR #3573, opened 2020-04-01
dvc: introduce `dvc params diff`

efiop:

@dmpetrov Volunteered to provide docs for this with nice examples :pray: 


dvc/command/params.py
```diff
@@ -0,0 +1,117 @@
...
+from dvc.command.base import append_doc_link
+from dvc.command.base import CmdBase
+from dvc.command.base import fix_subparsers
+from dvc.exceptions import DvcException
+
+
+logger = logging.getLogger(__name__)
+
+
+def _show_diff(diff):
```
efiop, line 13: Don't like code duplication with what we have in metrics.py, will take a look.

dvc/command/params.py
```diff
@@ -0,0 +1,117 @@
...
+    return table.draw()
+
+
+class CmdParamsDiff(CmdBase):
+    def run(self):
+        try:
+            diff = self.repo.params.diff(
+                a_rev=self.args.a_rev,
+                b_rev=self.args.b_rev,
+                targets=self.args.targets,
```
efiop, line 39: Support for `targets` is not implemented yet.

dvc/remote/local.py
```diff
@@ -76,10 +76,9 @@ def get(self, md5):
 
         return self.checksum_to_path_info(md5).url
 
-    @staticmethod
-    def exists(path_info):
+    def exists(self, path_info):
         assert path_info.scheme == "local"
-        return os.path.lexists(fspath_py35(path_info))
+        return self.repo.tree.exists(fspath_py35(path_info))
```
efiop, line 81: So currently we have a mixup of direct `os.path` and `self.repo.tree` interfaces in our RemoteLOCAL, which is not ideal. We really should work on trees, but for now this is the only thing that was blocking `params diff`, so I've adjusted it. Luckily these are very easy to find with simple tests.

dvc/repo/params/diff.py
```diff
@@ -0,0 +1,17 @@
...
+        return params.get(rev or "", {})
+    except NoParamsError:
+        return {}
+
+
+def diff(repo, *args, a_rev=None, b_rev=None, **kwargs):
+    old = _get_params(repo, *args, **kwargs, rev=(a_rev or "HEAD"))
+    new = _get_params(repo, *args, **kwargs, rev=b_rev)
+
+    return dvc.utils.diff.diff(old, new)
```
efiop, line 17: This will return a dict with `diff` in it, same as with metrics. Not sure if it is okay to leave it as is and if that functionality will come in handy later. Will take a look.

dvc/repo/params/show.py
```diff
@@ -0,0 +1,92 @@
...
+        if not params:
+            continue
+
+        res[str(dep.path_info)] = params
+
+    return res
+
+
+@locked
+def show(
```
efiop, line 44: All of this code is very similar to metrics show, so might look into deduping it.

dvc/command/params.py
```diff
@@ -0,0 +1,117 @@
...
+                recursive=self.args.recursive,
+            )
+
+            if self.args.show_json:
```
efiop, line 43: No unit test present yet.

efiop:
Issues:
```
angles:          
    - 2          
    - 3          
    - 100        
    - 3.2        
    - 36.2       
    - 3.2        
df3: [1, 3]      
train:           
    lr: 0.0042   
    layers: 10   
    dropout: 0.21
processing:      
    th: 0.98     
    bow: 7000    
```

diff for angles will be shown as `angles.N` where N is the number of the element that changed in the list. Need to treat them as lists.

And if you change train.lr to `0.0042`, diff will show it as `0.004`. These affect metrics diff too.

Suor:
@efiop when do you expect this to be merged&

efiop:
@Suor Need to implement support for targets (will do today) and get an approval from @dmpetrov .

dvc/repo/params/show.py
efiop: On the other hand there will clearly be some refactoring in https://github.com/iterative/dvc/issues/3572 , so I will just leave it as it is and will unify in the followup PR.

efiop:
Ok, got aproval from Dmitry. Merging.

PR #3584, opened 2020-04-02
dvc: implement multi-stage dvcfile

### LINKED ISSUE #3606: dvc metrics add: should it dvc add untracked files?
I propose `dvc metrics add`, when used with an untracked file, automatically creates the orphan DVC-stage for it (calling `dvc add` under the hood) instead of

```console
λ dvc metrics add un-tracked.file
ERROR: failed to add metric file 'metric' - Unable to find DVC-file with output 'metric'
```

skshetry:

This PR adds `-n` flag (hidden at the moment), and will create a multi-stage dvcfile.
Eg:
```sh
$ dvc run -n stage1 -d foo -o bar "cat foo foo > bar"
```

For addressing the given stage:
```sh
$ dvc repro Dvcfile:stage1
$ dvc pipeline show --ascii :stage1  # assumes Dvcfile automatically
```

Multistage file format:
```yaml
stages:
  stage1:
    cmd: cat foo foo > bar
    deps:
    - foo
    outs_no_cache:
    - bar
   outs:
   wdir:
   ...
```

### Lockfile format
```json
{
  "stage1": {
    "cmd": "cat foo foo > bar",
    "deps": {
      "foo": "d3b07384d113edec49eaa6238ad5ff00"
    },
    "outs": {
      "bar": "5fb7ba7e8447a836e774b66155f5776a"
    }
  },
  "stage2": {
    "cmd": "cat lorem lorem > ipsum",
    "deps": {
      "lorem": "f737a087bca81f69a6048ec744c73e41"
    },
    "outs": {
      "ipsum": "bfb52667a9d6a89c7796332adb1da3cd"
    }
  }
}
```

---
Closes #3606


dvc/stage.py
```diff
@@ -1066,3 +1068,138 @@ def get_used_cache(self, *args, **kwargs):
...
+
+    @name.setter
+    def name(self, name):
+        self._name = name
+
+    def dump(self):
+        self.dvcfile.dump(self)
+
+    @staticmethod
+    def create(repo, accompany_outs=False, **kwargs):
```
skshetry, line 1111: It's still in the *make it work* stage, so lot's of copy-pasta.

dvc/repo/run.py
```diff
@@ -5,18 +5,26 @@
...
-    self.check_modified_graph([stage])
+    # TODO: check if the stage with given name already exists, don't allow that
+    self.check_modified_graph([stage], self.pipeline_stages)
+    self.pipeline_stages.append(stage)
 
     if not no_exec:
         stage.run(no_commit=kwargs.get("no_commit", False))
 
+    if stage_cls == PipelineStage:
+        stage.dvcfile.dump_multistages(stage, stage.path)
```
efiop, line 26: This is really strange. Why not
```
dvcfile.dump()
```
? Just let the dvcfile know what type it is, no reason for us to do such hacks.

skshetry: Yup, but we don't really need to dump to dvcfile at all (except `dvc run` obviously). This PR needs a lot of adjustments after #3619.

dvc/stage/__init__.py
```diff
@@ -732,3 +733,77 @@ def get_used_cache(self, *args, **kwargs):
...
+    def _changed(self):
+        return self._changed_deps() or self._changed_outs()
+
+    def reload(self):
+        from dvc.dvcfile import Dvcfile
+
+        return Dvcfile(self.repo, self.path).load_one(self.name)
+
+    @classmethod
+    def create(cls, repo, path, **kwargs):
```
skshetry, line 763: - [ ] Reorganize this

dvc/schema.py
```diff
@@ -0,0 +1,41 @@
...
+SINGLE_PIPELINE_STAGE_SCHEMA = {
+    str: {
+        StageParams.PARAM_CMD: str,
+        Optional(StageParams.PARAM_WDIR): str,
+        Optional(StageParams.PARAM_DEPS): [str],
+        Optional(StageParams.PARAM_OUTS): [str],
+        Optional(StageParams.PARAM_LOCKED): bool,
+        Optional(StageParams.PARAM_META): object,
+        Optional(StageParams.PARAM_ALWAYS_CHANGED): bool,
+        Optional("persist"): [str],
```
skshetry, line 34: I totally forgot about `persist`. 😂

skshetry: It'd be great if we don't even dump `cache: True, persist: False`, etc. on outs for output pointer file.

efiop: @skshetry Totally agree with that. We also don't need to dump default values. That is also something that we've been talking about for a long time, just didn't get to it :slightly_smiling_face: Maybe leave it for later PR, if it doesn't matter for your PR.

dvc/command/pipeline.py
```diff
@@ -8,15 +8,26 @@
 logger = logging.getLogger(__name__)
 
 
+def _stage_repr(stage):
```
efiop, line 11: Why not just bake it into stage objects? :slightly_smiling_face:

dvc/command/run.py
```diff
@@ -96,6 +97,7 @@ def add_parser(subparsers, parent_parser):
         default=[],
         help="Declare dependencies for reproducible cmd.",
     )
+    run_parser.add_argument("-n", "--name", help="Specify name of the stage.")
```
efiop, line 100: Let's make it hidden for now, until the 1.0 release.

dvc/dvcfile.py
```diff
@@ -29,6 +30,29 @@
...
+        stage.name: {
+            key: value
+            for key, value in {
+                stage.PARAM_CMD: stage.cmd,
+                stage.PARAM_WDIR: stage.resolve_wdir(),
+                stage.PARAM_DEPS: [d.def_path for d in stage.deps],
+                stage.PARAM_OUTS: [o.def_path for o in stage.outs],
+                stage.PARAM_LOCKED: stage.locked,
+                stage.PARAM_ALWAYS_CHANGED: stage.always_changed,
+                "persist": [o.def_path for o in stage.outs if o.persist],
```
efiop, line 49: We used to have a PARAM const for it too, why has it changed?

dvc/dvcfile.py
```diff
@@ -29,6 +30,29 @@
...
+            }.items()
+            if value
```
efiop, line 51: This will probably keep default wdir value, which is not really desirable.

dvc/lockfile.py
```diff
@@ -0,0 +1,86 @@
...
+if TYPE_CHECKING:
+    from dvc.repo import Repo
+    from dvc.stage import Stage
+
+
+class LockfileCorruptedError(DvcException):
+    message = "Lockfile is corrupted."
+
+    def __init__(self):
+        super().__init__(self.message)
```
efiop, line 19: Pretty weird organisation that doesn't really match the rest.

dvc/repo/__init__.py
```diff
@@ -189,14 +189,44 @@ def check_modified_graph(self, new_stages):
...
+            path = os.path.abspath(path)
+
+            if recursive and os.path.isdir(path):
+                stages = nx.dfs_postorder_nodes(graph or self.pipeline_graph)
+                return [
+                    stage for stage in stages if path_isin(stage.path, path)
+                ]
+            return [dvcfile.load_one(name)]
+
+        from dvc.stage import PipelineStage
```
efiop, line 214: Let's keep the imports orginised, no need to throw them around all over. Also, need to preserve order: system imports, 3rd party libs, our imports.

dvc/repo/run.py
```diff
@@ -2,11 +2,12 @@
...
 from .scm_context import scm_context
+from ..exceptions import DvcException
+from ..utils import relpath
 
 
-@locked
-@scm_context
-def run(self, fname=None, no_exec=False, **kwargs):
+def _legacy_run(self, fname=None, no_exec=False, **kwargs):
+    # XXX: make tests pass
```
efiop, line 10: I guess they pass already? :slightly_smiling_face:

dvc/repo/run.py
```diff
@@ -40,3 +41,55 @@ def run(self, fname=None, no_exec=False, **kwargs):
...
+    if dvcfile.exists() and not dvcfile.is_multi_stage():
+        raise DvcException(
+            "'{}' is a single-stage dvcfile. Please use "
+            "`-f <different-filename>` and try again.`.".format(
+                relpath(dvcfile.path)
+            )
+        )
+
+    # XXX: If the user has only output files, but not the pipeline files,
+    #  then the user should thyself delete and run.
```
efiop, line 80: You've mentioned that you wanted to get rid of this.

dvc/stage/__init__.py
```diff
@@ -32,7 +32,7 @@
 logger = logging.getLogger(__name__)
 
 
-class Stage(schema.StageParams):
+class Stage(params.StageParams):
```
efiop, line 35: This is also a pretty weird organisation :slightly_smiling_face: Though I understand that you did it to hack around quicker. We'll need to revisit all of this after and clean this up.

dvc/dvcfile.py
skshetry: Not really. `resolve_wdir` will return `None` if it's `os.curdir`.

dvc/command/pipeline.py
skshetry: Will come to this when I fix UI issues. The reason why I haven't fixed this is because, I want to make `__repr__` to be `Stage: <abspath>:<name>` and `__str__` to be `Stage: <relpath>:<name>`.  It feels like this should be `get_address` or `get_address_name` or something. Anyway, I'll revisit it later on.

skshetry:
@efiop, this should be good to merge. Please take a look (I have fixed all except 1).

efiop:
@skshetry Thanks! :pray: Let's run with it. Merging for now, will play around with the codebase some more while adding build-cache support.

tall-josh:
Looking forward to having a go at this! I've been hacking together a script that parses a `params.yaml` to form richer `dvc run` commands. (ie: adding `-o`, `-m` flags etc). This looks like it will do what I need, but in a more complete way.

tall-josh:
I just realized multi-stage doesn't seem to support `--params` deps. Is this on the radar? Or, am I missing something?

skshetry:
Hi, @tall-josh. We are still iterating on this issue, and `params` was left out from the first iteration, because we are still discussing on few key things like:
1.  We create a separate stage files for the outputs. Do users care about this separation or is just a clutter we should avoid?
2. Do we need lockfile or should that be similar to current `*.dvc` files that can support multiple stages?
3. Should that multistage Dvcfile or lockfile be human readable/editable? Should that be multi-line json or yaml? *et cetera*

Looks like you dive into this, maybe you have some inputs?

tall-josh:
Thanks @skshetry. My hope was to use the multi-stage files as a way to declaratively put together a pipeline. I'm not sure if this was the intention, but that goal in mind:

  1. Decluttering isn't my driving force, its making things declarative. Personally I create a separate `dvc` directory to contain the clutter generated by `*.dvc` files.
  2. I like the idea of the lock file being separate as it would keep the multi-stage file neater. Intuitively I feel like there should be one lock file per multi-stage file. I feel like that's easier for new comers to build a mental model around what's happening, but I don't have a strong opinion either way really.
  3. I would prefer multi-line yaml. Again, it just helps make sense of what's going on behind the scenes. 

We have another discussion on #3633 about referencing variables inside dvc (single and multi-stage)  files that may be helpful too. These two features would go a long way to my declarative dvc dream.

Issue #3496, opened 2020-03-17
update: cache git repos


casperdcl:
```
dvc import https://some/git/repo/ some_file
dvc update  # should not re-clone, should only pull into existing cache
```

Suor:
The thing is cache is not persisted between `dvc` runs, if we make it persist then that won't reclone only make `git pull` in `dvc update`.

casperdcl:
yes; this is about making it persistent & pulling rather than re-cloning.

jorgeorpinel:
What about a repo cache at the user level? Could be a system config var so you can disable it, like analytics.

> Context: #4203

casperdcl:
in light of #4246 being merged going to downgrade priority here...

johnyaku:
Persistent clones (as per #10511) are different from shallow clones (as per #4246).
Both speed up cloning (or potentially avoid it) but only persistent clones can allow us to work with imported data without internet connectivity, which is necessary for us on a HPC where most queues have no connectivity.

Persistent clones would also allow us to separate cloning (which requires connectivity) from other dvc operations (which don't). This would allow us to do the former in an environment (queue) with connectivity and the latter in environments without.

dberenbaum:
@johnyaku Have you considered keeping a clone on a shared space of the HPC so you can import from there instead of from the internet? Even if dvc had some support for caching clones, it would likely still need to check the internet to fetch updates from those clones. If you have your own clone of the repo, you can fully control when to update it and everyone can share that single repo copy (dvc will not make a new clone of a local repo).

johnyaku:
@dberenbaum I've been thinking along the same lines. We could maintain (and periodically update) repos on the local filesystem and specify the path to those repos instead of GitHub URLs. This would solve the no-internet access problem.

But we also want to maintain portability between platforms. (We work on two different HPCs, plus GCP.) So a URL that is accessible from any platform would be better from a portability perspective.

I can have calls to GitHub loopback to localhost in `~/.ssh/config` but then I'd need to change those settings back in order to update the local mirrors, which is potentially tedious (and error prone if there are other processes access `~/.ssh/config` at the same time). If I could maintain two separate configs then I might be able to get it to work but AFAIK the location of the SSH config is not configurable.

Happy to explore workarounds like this, or maybe dvc could keep the clones that is making already?

PR #3585, opened 2020-04-02
get/import: --rev BRANCH: shallow clone

### LINKED ISSUE #3473: 'dvc get' is cloning the full bitbucket repo when only one branch is needed
DVC version : 0.82.2

When using `dvc get` to get data from a large bitbucket repository (with full command  `dvc get --rev branchname  https://.../gitrepo  data/data.csv`), this results in cloning the full git repository, even though only the one specified branch was needed. It would be preferable to be able to only clone the one specified branch.

casperdcl:
`dvc <get|import> REPO --rev BRANCH` should clone with `--single-branch --branch BRANCH` rather than cloning the full repo.

We could also add `--depth 1` to avoid getting the full commit history.

Alternatively we could do `--branch BRANCH --no-single-branch --depth 1` to get all heads without history.

- [x] pass `--rev BRANCH` to `git clone --branch BRANCH`
  - [x] add `--depth=1`
    + [ ] expose `--depth` option?
- [x] fallback to full clone so that this doesn't break with other types of `--rev`
- [ ] depends on #3496
  + [ ] can fallback on `--unshallow` once we have a proper global cache
- fixes #3473

### DIFF
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -248,7 +248,7 @@ def _clone_default_branch(url, rev):
         else:
             logger.debug("erepo: git clone %s to a temporary dir", url)
             clone_path = tempfile.mkdtemp("dvc-clone")
-            git = Git.clone(url, clone_path)
+            git = Git.clone(url, clone_path, rev=rev)
             CLONES[url] = clone_path
     finally:
         if git:
--- a/dvc/scm/git/__init__.py
+++ b/dvc/scm/git/__init__.py
@@ -1,5 +1,6 @@
 """Manages Git."""
 
+from functools import partial
 import logging
 import os
 import yaml
@@ -71,18 +72,26 @@ class Git(Base):
             # [1] https://github.com/gitpython-developers/GitPython/issues/924
             env[ld_key] = ""
 
+        clone_from = partial(
+            git.Repo.clone_from,
+            url,
+            to_path,
+            env=env,  # needed before we can fix it in __init__
+            no_single_branch=True,
+        )
+
         try:
-            tmp_repo = git.Repo.clone_from(
-                url,
-                to_path,
-                env=env,  # needed before we can fix it in __init__
-                branch=rev,
-                no_single_branch=True,
-                depth=1,
-            )
-            tmp_repo.close()
+            tmp_repo = clone_from(branch=rev, depth=1)
         except git.exc.GitCommandError as exc:
-            raise CloneError(url, to_path) from exc
+            if f"Remote branch {rev} not found" not in str(exc):
+                raise CloneError(url, to_path) from exc
+            try:
+                # not a branch - need to do full clone
+                tmp_repo = clone_from()
+            except git.exc.GitCommandError as exc:
+                raise CloneError(url, to_path) from exc
+        else:
+            tmp_repo.close()
 
         # NOTE: using our wrapper to make sure that env is fixed in __init__
         repo = Git(to_path)
--- a/dvc/scm/git/__init__.py
+++ b/dvc/scm/git/__init__.py
@@ -90,8 +90,8 @@ class Git(Base):
                 tmp_repo = clone_from()
             except git.exc.GitCommandError as exc:
                 raise CloneError(url, to_path) from exc
-        else:
-            tmp_repo.close()
+
+        tmp_repo.close()
 
         # NOTE: using our wrapper to make sure that env is fixed in __init__
         repo = Git(to_path)
--- a/dvc/scm/git/__init__.py
+++ b/dvc/scm/git/__init__.py
@@ -76,7 +76,9 @@ class Git(Base):
                 url,
                 to_path,
                 env=env,  # needed before we can fix it in __init__
-                no_single_branch=True,
+                branch=rev,
+                single_branch=True,
+                depth=1,
             )
             tmp_repo.close()
         except git.exc.GitCommandError as exc:
--- a/dvc/scm/git/__init__.py
+++ b/dvc/scm/git/__init__.py
@@ -77,7 +77,7 @@ class Git(Base):
                 to_path,
                 env=env,  # needed before we can fix it in __init__
                 branch=rev,
-                single_branch=True,
+                no_single_branch=True,
                 depth=1,
             )
             tmp_repo.close()
--- a/dvc/scm/git/__init__.py
+++ b/dvc/scm/git/__init__.py
@@ -88,7 +88,7 @@ class Git(Base):
                 logger.debug("full clone")
                 tmp_repo = clone_from()
         except git.exc.GitCommandError as exc:
-            if not rev or f"Remote branch {rev} not found" not in str(exc):
+            if not rev or ("Remote branch %s not found" % rev) not in str(exc):
                 raise CloneError(url, to_path) from exc
             try:
                 logger.debug("not a branch - performing full clone")
--- a/dvc/scm/git/__init__.py
+++ b/dvc/scm/git/__init__.py
@@ -81,12 +81,17 @@ class Git(Base):
         )
 
         try:
-            tmp_repo = clone_from(branch=rev, depth=1)
+            if rev:
+                logger.debug("attempting shallow clone of branch %s", rev)
+                tmp_repo = clone_from(branch=rev, depth=1)
+            else:
+                logger.debug("full clone")
+                tmp_repo = clone_from()
         except git.exc.GitCommandError as exc:
-            if f"Remote branch {rev} not found" not in str(exc):
+            if not rev or f"Remote branch {rev} not found" not in str(exc):
                 raise CloneError(url, to_path) from exc
             try:
-                # not a branch - need to do full clone
+                logger.debug("not a branch - performing full clone")
                 tmp_repo = clone_from()
             except git.exc.GitCommandError as exc:
                 raise CloneError(url, to_path) from exc


casperdcl:
It's probably best to keep `--no-single-branch` so that other HEADs are tracked in case checking out something else is required.

It looks like `--rev` has always been passed to `git clone --branch`. This means that we've never supported non-branch revs...

```bash
dvc ((28d16f7...))$ dvc import https://github.com/iterative/dvc \
  README.rst --rev 28d16f7

ERROR: failed to import 'README.rst' from 'https://github.com/iterative/dvc'. - unknown Git revision '28d16f7eeeedd'
```

So I think it's best to have `--no-single-branch --depth=1`. This shouldn't break backward-compatibility. Supporting commit `rev`s should be for a different issue.

shcheklein:
@casperdcl there is a cache of external repos that we support (check `CLONES` in the `external_repo.py`). I think with this change we'll have to modify how this cache behave. It's not about modifying the clone operation only, unfortunately.

shcheklein:
@Suor can give more context on this.

casperdcl:
The way this is implemented (with the fallback on a full clone) I doubt anything should break (the tests are all passing). I'd need a unit test or at least a demo scenario where this would break.

Within a cached session, a shallow clone of a branch would have to be followed by a checkout of a SHA. That would break things. When would that ever realistically happen though?

casperdcl:
recap: untested behaviour is broken; need to sort out session (or even global #3496) repo caches first.

dvc/external_repo.py
```diff
@@ -248,7 +248,7 @@ def _clone_default_branch(url, rev):
         else:
             logger.debug("erepo: git clone %s to a temporary dir", url)
             clone_path = tempfile.mkdtemp("dvc-clone")
-            git = Git.clone(url, clone_path)
+            git = Git.clone(url, clone_path, rev=rev)
```
Suor, line 251: This breaks clones cache below. As it presumes it has default branch.

dvc/scm/git/__init__.py
```diff
@@ -71,16 +72,31 @@ def clone(url, to_path, rev=None):
...
-                env=env,  # needed before we can fix it in __init__
-                no_single_branch=True,
-            )
-            tmp_repo.close()
+            if rev:
+                logger.debug("attempting shallow clone of branch %s", rev)
+                tmp_repo = clone_from(branch=rev, depth=1)
+            else:
+                logger.debug("full clone")
+                tmp_repo = clone_from()
```
Suor, line 89: This is very dangerous code, which will cause us grief. We shouldn't hide that we are sometimes make shallow code and sometimes not from outside. Outside may rely on clone being full as it currently does in `external_repo.py`.

Suor:
Some explanations on how external repo cache works now:

- we cache default branch clone, which is presumed to be full clone. This clone is updated with `git pull` each time we try to access anything but not a known sha (branch, tag, unknown sha)
- we also cache checked out repos, these are copies of the default branch repo, so they are full too. They are cached by `(url, sha)` pair, so will never be updated. Full clone is not really needed.
- we have write mode, in which copied and checked out repo is presumed to be modified, we do the same as above, but don't cache it, the copy is removed in the end 
- caches are in-process, this means that separate `dvc` commands, either sequential or parallel, won't use shared cache. This may cause issues for people having big repos.
- we only clone each url once per process run. This was the requirement because `git clone` may ask for a password and we don't want people to be asked for the same password repeatedly. `git pull`, which we use to update an existing clone doesn't cause password prompt. Note, passwords are usually asked for https urls, git/ssh urls rely on keys.

Switching from full to shallow clones will cause some issues:
- we can't simply clone for each pair of `(url, rev)` and cache it because people will be asked password repeatedly
- if we make a shallow copy of default branch then we can't check out other branch or tag or sha as we do now
- we may, however, try to `git fetch` whatever we want to check out to, make a copy, then checkout. There are some subconsiderations:
    - need to check that `git fetch` won't ask password
    - do not need to fetch a known sha (same optimization we do know for `git pull`)
    - it doesn't matter whether we already know branch or tag, we still need `git fetch` since they might have moved

Making cache work between runs will also cause us some issues:
- several processes may perform the same clone simultaneously, have no idea whether git handles this anyhow
- several processes might be updating a default clone simultaneously. Git might already handle this, but need to check
- one process may be making a copy of default clone, while other is updating it via `git pull` or `git fetch`. If we switch from simple copying to `git clone local/path` this might be already solved by git, but this needs to be checked
- if git doesn't handle any of the above by its lock then we will need to protect the corresponding op with our lock

Bringing external repo cache outside a process is a prereq for #3496. This is not required to fix #3473 though, which is about shallow clones. So the parts above are independent.

dvc/scm/git/__init__.py
casperdcl: we're going to have to fix "outside" (i.e. `external_repo.py`) first

casperdcl:
@Suor (CC @shcheklein @efiop)

I think we're all on the same page; there's only one thing I disagree with (right at the end).

> * we cache default branch clone, which is presumed to be full clone.

This needs to be changed to remove this presumption (i.e. change to fetch missing things)

> This clone is updated with `git pull` each time we try to access anything but not a known sha (branch, tag, unknown sha)

Would need to `fetch --all` and potentially `--unshallow` if still not found.

> * we also cache checked out repos, these are copies of the default branch repo, so they are full too. They are cached by `(url, sha)` pair, so will never be updated. Full clone is not really needed.

You mean we have two copies of each repo? Why? Sounds like my response above should apply to this "working copy" repo instead.

> Switching from full to shallow clones will cause some issues:
> 
> * we can't simply clone for each pair of `(url, rev)` and cache it because people will be asked password repeatedly

I don't understand this at all. We'd clone (or pull if already in persistent cache) once for each `url`, and checkout for each `rev`.

> * if we make a shallow copy of default branch then we can't check out other branch or tag or sha as we do now
> * we may, however, try to `git fetch` whatever we want to check out to, make a copy, then checkout.

Currently we still really have to fetch/pull (e.g. to make sure when checking out a different branch that the branch is up-to-date). Maybe we've skipped that by assuming the repo was fully cloned in the same session and unlikely to have been updated since the last job within the session. Moving to a persistent cache, though, we naturally cannot assume checkout will work. I propose to fallback to `fetch --all` and then `--unshallow` when trying to checkout.

> * need to check that `git fetch` won't ask password

why? What's wrong with prompting if the user uses https without auth tokens? Git itself would prompt for every single operation in such scenarios. Git also has its own mechanism for a credential cache so we don't need to implement our own.

> * several processes may perform the same clone simultaneously, have no idea whether git handles this anyhow

the first process to create the target (cached) dir will succeed; the rest will fail without doing anything since the dir already exists. If the filesystem isn't atomic then I presume a process could needlessly perform an extra clone which gets immediately deleted

> * several processes might be updating a default clone simultaneously. Git might already handle this, but need to check

Yes, it should.

> * one process may be making a copy of default clone, while other is updating it via `git pull` or `git fetch`. If we switch from simple copying to `git clone local/path` this might be already solved by git, but this needs to be checked

Ah. starting to understand - so currently we create a full clone mirror and then copy from that mirror? Yeah that needs to change to:

1. not assume the mirror is "full" (fallback to `fetch --all` and then `--unshallow`). Ideally also make the mirror "bare".
2. clone - rather than copy - from the mirror to a working copy (ideally both shallow and sparse).

> * if git doesn't handle any of the above by its lock then we will need to protect the corresponding op with our lock

Yes; fun times.

> Bringing external repo cache outside a process is a prereq for #3496.

Some misunderstanding here. I think it *is* #3496, not a separate prereq.

> This is not required to fix #3473 though, which is about shallow clones. So the parts above are independent.

Fully disagree with this. As discussed above, my proposal would involve necessarily persisting repos first; then removing assumptions about their fullness/unshallowness, and then as an afterthought considering sparsity of working copies.

Quick recap of related issues (think we should create a new "super" issue or a mini-project to keep track of them?)

- #3496 persistent repo cache
- #3585 <- #3473 shallow clone
- #3448 sparse checkout

I think they have to be dealt with in this specific order.

efiop:
For the record: agreed to freeze this for now, I'll come back to this after I utilize DvcTree in external_repo stuff. Should give a fresh perspective on this.

efiop:
@pmrowla Could you take a look at this, please? It would be great to hear your opinion about it.

pmrowla:
For erepos, in most cases we now generate a `GitTree` for the needed revision (using `resolve_rev()`) instead of doing the `git checkout` into the clone directory. (dvcx `for_write=True` being the exception where we still need a full clone + checkout)

So other than the dvcx use case, we don't really need full clones at all now. I think we should be able to use a single shallow clone (`git clone --depth=1 --no-single-branch`) and `git fetch --depth=1` for being able to resolve additional specific revisions that we need. For branches we will need to be able to fetch/fast-forward to the correct revision so that HEAD points to the right place, but that is still doable with shallow clones.

I'm not sure how shallow clones + fetch work regarding the potential password prompt on fetch issue though.

pmrowla:
> You mean we have two copies of each repo? Why? Sounds like my response above should apply to this "working copy" repo instead.

Just to clarify, this is no longer the case after the tree related erepo changes. We still have the single main full clone, but rather than separate copies for checkouts of other revisions, we now create GitTree's directly from the one initial clone (with dvcx `for_write` as the exception).

I don't have the bandwidth to finish this right now, but regarding shallow clones:

* We can clone with `--no-single-branch --depth=1 --branch=<branch>` if the user is running `dvc get/import --rev=<branch>` where `<branch>` is an actual branch or a tag.
This should probably be handled by exposing `depth` and `branch` as kwargs in:
https://github.com/iterative/dvc/blob/c1b04b88fac76ed49b9591288f885e38a9cfe48c/dvc/scm/git/__init__.py#L96
and then passing the appropriate params when we create the erepo clone:
https://github.com/iterative/dvc/blob/c1b04b88fac76ed49b9591288f885e38a9cfe48c/dvc/external_repo.py#L272

* If <branch> is a revision sha, we still need to do the existing full clone behavior
* We also need to keep the existing standalone full clone for dvcx `for_write`
* For shallow clones, if we need to access a different rev, we will need to do `fetch --unshallow` where we currently do
https://github.com/iterative/dvc/blob/c1b04b88fac76ed49b9591288f885e38a9cfe48c/dvc/external_repo.py#L284-L286

Other than that, I don't think shallow clones should affect any of the tree related erepo behavior. Also, as far as I can tell from testing on my end, using shallow clones should not affect the current git password related behavior either.

pmrowla:
closing in favor of #4246.

PR #3591, opened 2020-04-04
CLI: redirect warnings to logger

shcheklein:
The `captureWarnings()` function can be used to integrate logging with the warnings module (that's what GDrive dependency libraries use to write this ugly message below).

We first use this function, then logger is silenced but since it has `py.` prefix.

Eventually, we should probably stop using loggers for the user interface and all WARNINGS should not be silenced, but be visible when I do `-v`? But we need to discuss this, may be we can introduce `-vvv` to show third party logger messages eventually.

- [x] Hide /Users/ivan/Projects/test-gdrive/.env/lib/python3.7/site-packages/oauth2client/_helpers.py:255: UserWarning: Cannot access /Users/ivan/Projects/test-gdrive/.dvc/tmp/gdrive-user-credentials.json: No such file or directory warnings.warn(_MISSING_FILE_MESSAGE.format(filename)) on the first auth. 

- [x] minor: unused var assignment

--------------------

### DIFF
--- a/dvc/logger.py
+++ b/dvc/logger.py
@@ -143,6 +143,7 @@ def _stack_trace(exc_info):
 
 
 def disable_other_loggers():
+    logging.captureWarnings(True)
     root = logging.root
     for (logger_name, logger) in root.manager.loggerDict.items():
         if logger_name != "dvc" and not logger_name.startswith("dvc."):
--- a/dvc/__init__.py
+++ b/dvc/__init__.py
@@ -3,17 +3,8 @@ DVC
 ----
 Make your data science projects reproducible and shareable.
 """
-import warnings
-
 import dvc.logger
 from dvc.version import __version__  # noqa: F401
 
 
 dvc.logger.setup()
-
-# Ignore numpy's runtime warnings: https://github.com/numpy/numpy/pull/432.
-# We don't directly import numpy, but our dependency networkx does, causing
-# these warnings in some environments. Luckily these warnings are benign and
-# we can simply ignore them so that they don't show up when you are using dvc.
-warnings.filterwarnings("ignore", message="numpy.dtype size changed")
-warnings.filterwarnings("ignore", message="numpy.ufunc size changed")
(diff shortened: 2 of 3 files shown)


shcheklein:
@jorgeorpinel if this merged we can remove `...` in the code snippets we have in your PR for the GDrive docs.

dvc/logger.py
```diff
@@ -143,6 +143,7 @@ def _stack_trace(exc_info):
 
 
 def disable_other_loggers():
+    logging.captureWarnings(True)
```
efiop, line 146: Oh, good to know there is such a feature! I now wonder If we still need https://github.com/iterative/dvc/blob/master/dvc/__init__.py#L18 .

efiop: But also, these will still appear in regular output, because all warning do (and they should). So not sure how useful this change is.

efiop: Ok, so it will redirect them to `py.warnings` logger, so we won't print it in CLI, so it means that indeed that ^ line where we handle warnings from numpy is no longer needed.

shcheklein:
thanks @efiop !

jorgeorpinel:
> @jorgeorpinel if this merged we can remove ... in the code snippets we have in your PR for the GDrive docs.

I only have those for `dvc push` but I'm still getting a long ugly message in there, using DVC 0.92.0:

```console
λ dvc push -r jopgd
  0% Querying cache in gdrive://13zVX0PgDW9LatVxibT3ckVErs6lkTqdd|                                                    |0/1 [00:00<?,     ?file/s]C:\Program Files (x86)\Data Version Control\oauth2client\_helpers.py:255: UserWarning: Cannot access C:\Users\poj12\DVC-repos\test\.dvc\tmp\gdrive-user-credentials.json: No such file or directory
Go to the following link in your browser:

    https://accounts.google.com/o/oauth2/auth?...
```

jorgeorpinel:
In any case the `0% Querying cache ...` part will remain right? So still need `...` in the examples.

shcheklein:
@jorgeorpinel the new version is not released yet.

I would not include progress bar.

jorgeorpinel:
OK, replaced ... for blank lines now 👍

PR #3595, opened 2020-04-05
scm: git: clone: add progress

casperdcl:
- [x] add progress for `git clone`
- [x] ~~add progress for `git fetch`/`pull`?~~ to be addressed later
- [x] append to rather than remove initial `desc`
  + [ ] (perhaps even in other places in a different issue)?

[asciicast](https://asciinema.org/a/l8NoFwah4hVobyT7j6MTbfAJ5)

### DIFF
--- a/dvc/scm/git/__init__.py
+++ b/dvc/scm/git/__init__.py
@@ -107,7 +107,7 @@ class Git(Base):
             env[ld_key] = ""
 
         try:
-            with TqdmGit(desc="Cloning") as pbar:
+            with TqdmGit(desc="Cloning", unit="obj") as pbar:
                 tmp_repo = git.Repo.clone_from(
                     url,
                     to_path,
--- a/dvc/scm/git/__init__.py
+++ b/dvc/scm/git/__init__.py
@@ -21,9 +21,36 @@ logger = logging.getLogger(__name__)
 
 class TqdmGit(Tqdm):
     def update_git(self, op_code, cur_count, max_count=None, message=""):
+        op_code = self.code2desc(op_code)
+        if op_code:
+            message = op_code + " " + message
         if message:
             self.set_description_str(message, refresh=False)
         self.update_to(cur_count, max_count)
+
+    @staticmethod
+    def code2desc(op_code):
+        from git import RootUpdateProgress as OP
+
+        ops = dict(
+            (
+                (OP.COUNTING, "Counting"),
+                (OP.COMPRESSING, "Compressing"),
+                (OP.WRITING, "Writing"),
+                (OP.RECEIVING, "Receiving"),
+                (OP.RESOLVING, "Resolving"),
+                (OP.FINDING_SOURCES, "Finding sources"),
+                (OP.CHECKING_OUT, "Checking out"),
+                (OP.CLONE, "Cloning"),
+                (OP.FETCH, "Fetching"),
+                (OP.UPDWKTREE, "Updating working tree"),
+                (OP.REMOVE, "Removing"),
+                (OP.PATHCHANGE, "Changing path"),
+                (OP.URLCHANGE, "Changing URL"),
+                (OP.BRANCHCHANGE, "Changing branch"),
+            )
+        )
+        return ops.get(op_code & OP.OP_MASK, "")
 
 
 class Git(Base):
--- a/dvc/scm/git/__init__.py
+++ b/dvc/scm/git/__init__.py
@@ -23,9 +23,9 @@ class TqdmGit(Tqdm):
     def update_git(self, op_code, cur_count, max_count=None, message=""):
         op_code = self.code2desc(op_code)
         if op_code:
-            message = op_code + " " + message
+            self.set_description_str(op_code, refresh=False)
         if message:
-            self.set_description_str(message, refresh=False)
+            self.set_postfix_str(message, refresh=False)
         self.update_to(cur_count, max_count)
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


shcheklein:
Is strange that first I'm getting:

`Cloning                                                                                                                                                                            0.00 [00:00,       ?it/s]`

After a while:

`900.00 KiB | 1.69 MiB/s                                                                                                                                                           4.73k [00:02,    14.0it/s]`

Is it expected? 

Also I don't see the progress bar itself? Is it because git python does not report it?

dvc/scm/git/__init__.py
```diff
@@ -18,6 +19,13 @@
 logger = logging.getLogger(__name__)
 
 
+class TqdmGit(Tqdm):
+    def update_git(self, op_code, cur_count, max_count=None, message=""):
```
efiop, line 23: Hm, don't we need an `op_code`? It tells us which clone stage we are on.

shcheklein:
@casperdcl the recording:

https://asciinema.org/a/TZXq1fD3SIi3VB6bGUguscA5Z

dvc/scm/git/__init__.py
casperdcl: atm I'm not parsing it. Could either:

1. ignore it
2. prefix current `message` with a formatted version of `op_code`

casperdcl:
Thanks @shcheklein again for the recording; yes it's expected:

> Is strange that first I'm getting:
> 
> `Cloning 0.00 [00:00, ?it/s]`
> 
> After a while:
> 
> `900.00 KiB | 1.69 MiB/s 4.73k [00:02, 14.0it/s]`
> 
> Is it expected?

`900.00 KiB | 1.69 MiB/s` is actually the message string provided by git to the `progress_callback`. Yes, this is weird. And we can't rely on it being formatted in this way (esp. for other git operations).

> Also I don't see the progress bar itself? Is it because git python does not report it?

Exactly, it doesn't report a `total` (specifically, it's passing in the default `max_count=None` to the callback), which is unfortunate.

dvc/scm/git/__init__.py
```diff
@@ -72,12 +80,14 @@ def clone(url, to_path, rev=None):
             env[ld_key] = ""
 
         try:
-            tmp_repo = git.Repo.clone_from(
-                url,
-                to_path,
-                env=env,  # needed before we can fix it in __init__
-                no_single_branch=True,
-            )
+            with TqdmGit(desc="Cloning") as pbar:
```
casperdcl, line 83: possibly do:
```suggestion
            with TqdmGit(desc="Cloning", unit="object") as pbar:
```

dvc/scm/git/__init__.py
casperdcl: reference: https://gitpython.readthedocs.io/en/stable/reference.html?highlight=RemoteProgress#git.remote.RemoteProgress.update


- op_code -
    Integer allowing to be compared against Operation IDs and stage IDs.

    Stage IDs are BEGIN and END. BEGIN will only be set once for each Operation ID as well as END. It may be that BEGIN and END are set at once in case only one progress message was emitted due to the speed of the operation. Between BEGIN and END, none of these flags will be set

    Operation IDs are all held within the OP_MASK. Only one Operation ID will be active per call.

-    cur_count – Current absolute count of items

-    max_count – The maximum count of items we expect. It may be None in case there is no maximum number of items or if it is (yet) unknown.

-    message – In case of the ‘WRITING’ operation, it contains the amount of bytes transferred. It may possibly be used for other purposes as well.

shcheklein:
@casperdcl but why do we replace `Cloning` with the stats message (900.00 KiB | 1.69 MiB/s)? And stats itself -4.73k [00:02, 14.0it/s] - do you get what does it mean?

casperdcl:
> @casperdcl but why do we replace `Cloning` with the stats message (900.00 KiB | 1.69 MiB/s)?

Seems like a good idea to display the message provided to the callback.

> And stats itself -4.73k [00:02, 14.0it/s] - do you get what does it mean?

It's from https://github.com/iterative/dvc/pull/3595#discussion_r403765189: number of "items." I believe in this case it's what git would call "objects" if you ran `git pull` manually:

```
$ git pull
...
Receiving objects: 100% (576/576), 585.47 KiB | 432.00 KiB/s, done.
...
```

shcheklein:
can we show all the stats to the right? the same as we do with main progress bar for Pull you have done? (no matter what library thinks). Or at least always prepend "Cloning "?

efiop:
@casperdcl What's the status of this PR?

casperdcl:
just updated the description - still awaiting a little tidy from me.

dvc/scm/git/__init__.py
casperdcl: opted for "obj"

dvc/scm/git/__init__.py
casperdcl: decided to parse `op_code` & update description accordingly, and use `message` in the postfix instead.

casperdcl:
@efiop ready to merge now I think. @shcheklein see the updated recording:

[asciicast](https://asciinema.org/a/l8NoFwah4hVobyT7j6MTbfAJ5)

Note that atm the py git api provides `total=None` to the progress callback so there's no way to guess a total hence there's no progress bar, just stats.

shcheklein:
LGTM! thanks @casperdcl

PR #3598, opened 2020-04-06
config: forbid setting core.no_scm outside repo

### LINKED ISSUE #3427: config: no_scm should be forbidden for --global and --system
Otherwise it will break all of your repos. Kudos @shcheklein

mroutis:
Raise a `ConfigError` when setting `core.no_scm` on a global or system level.

Close #3427

### DIFF
--- a/dvc/command/config.py
+++ b/dvc/command/config.py
@@ -48,6 +48,13 @@ class CmdConfig(CmdBaseNoRepo):
             msg = "option {} doesn't exist"
             raise ConfigError(msg.format(self.args.name))
 
+        if self.args.name == "core.no_scm" and self.args.level in [
+            "global",
+            "system",
+        ]:
+            msg = "{} not allowed on a {} level"
+            raise ConfigError(msg.format(self.args.name, self.args.level))
+
 
 parent_config_parser = argparse.ArgumentParser(add_help=False)
 parent_config_parser.add_argument(
--- a/tests/func/test_config.py
+++ b/tests/func/test_config.py
@@ -106,3 +106,10 @@ def test_merging_two_levels(dvc):
 def test_config_loads_without_error_for_non_dvc_repo(tmp_dir):
     # regression testing for https://github.com/iterative/dvc/issues/3328
     Config(validate=True)
+
+
+def test_no_scm(dvc):
+    assert 251 == main(["config", "--global", "core.no_scm", "true"])
+    assert 251 == main(["config", "--system", "core.no_scm", "true"])
+    assert 0 == main(["config", "--local", "core.no_scm", "true"])
+    assert 0 == main(["config", "core.no_scm", "true"])


mroutis:
@shcheklein , roger that. :slightly_smiling_face: 
what would be the way to go? I didn't find any similar implementation.

The current `validation` process happens after merging the configuration files:
https://github.com/iterative/dvc/blob/c1a32ff1bfb0852c575e101668da42229ab810fd/dvc/config.py#L275-L287

We could validate each configuration file separately, something like:

```diff
diff --git a/dvc/config.py b/dvc/config.py
index 6e1ccd7f..eba2768a 100644
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -281,10 +281,12 @@ class Config(dict):
         conf = {}
         for level in self.LEVELS:
             if level in self.files:
-                _merge(conf, self.load_one(level))
+                level_conf = self.load_one(level)
 
-        if validate:
-            conf = self.validate(conf)
+                if validate:
+                    level_conf = self.validate(level_conf, level)
+
+                conf = _merge(conf, level_conf)
```

What do you think?

shcheklein:
@mroutis it's a strong assumption. What if we have a section with two required fields and they are defined separately now in two configs. After your change it will start failing, right?

Also, question is how we define this rule on the SCHEMA level?

karajan1001:
In my opion, only those config in lower levels can affect the validation process .Maybe something like

```
         conf = {}
         for level in self.LEVELS:
             if level in self.files:
                _merge(conf, self.load_one(level))
                if validate:
                   conf = self.validate(conf)
```
I had go simliar problem in [pr3628](https://github.com/iterative/dvc/pull/3628)

efiop:
Closing as stale :slightly_frowning_face:

ribitskiyb:
@shcheklein, regarding 
> I think we should be doing this check on the schema level.

Schema and _levels_ (`system`, `global`, etc.) address different config content-related concerns -- validation and collection, respectively. Making the schema aware of the existence of _levels_ seems like an unnecessary complexification of it.
What if instead to introduce level-specific checks as a separate "concept" in the codebase?

shcheklein:
> Making the schema aware of the existence of levels seems like an unnecessary complexification of it.

It might be, @ribitskiyb . I was not trying to dictate any specific solution, to be honest. Rather convey the problem and brainstorm together here.

> What if instead to introduce level-specific checks as a separate "concept" in the codebase?

it depends, I guess. How would it look like? How would the access interface look like?

ribitskiyb:
@shcheklein how about simply 

1. adding a dict of level name to level-specific constraints (not sure yet if it can also be done with `voluptuous`)

1. checking the contents of level config just before merging it
`dvc/config.py`
image

shcheklein:
@ribitskiyb sounds good to me! Should if be part of the `load_one` logic?

ribitskiyb:
@shcheklein yea would be reasonable to put the check itself inside `load_one`

```
@@ -249,6 +249,9 @@ class Config(dict):
 
     # In the order they shadow each other
     LEVELS = ("system", "global", "repo", "local")
+    PER_LEVEL_PROHIBITED_OPTIONS = make_per_level_prohibited_options_config(
+        {"core.no_scm": {"system", "global"}}
+    )
 
     CONFIG = "config"
     CONFIG_LOCAL = "config.local"
@@ -368,6 +371,17 @@ class Config(dict):
         for key in COMPILED_SCHEMA.schema:
             conf.setdefault(key, {})
 
+        # Check if violates level-specific constraints
+        forbidden_options = self.PER_LEVEL_PROHIBITED_OPTIONS.get(level, ())
+        for jsonpath in forbidden_options:
+            if not jsonpath.find(conf):
+                continue
+            raise ConfigError(
+                "Setting {option} is forbidden on the {level} level".format(
+                    option=str(jsonpath), level=level,
+                )
+            )
+
         return conf
```

Issue #3568, opened 2020-04-01
Add summary to `dvc status -c [target]` 


AljoSt:
In order to estimate how long a `pull` is going to take, it would be good if `dvc status -c [target]` would supply a summary in the form of `Total: N files, X GB`

AljoSt:
had a brief look into this, but I'm not quite sure where to start. I guess we would have a method [here](https://github.com/iterative/dvc/blob/master/dvc/remote/local.py#L286) that adds the filesize for each object, and then summarized and printed [here](https://github.com/iterative/dvc/blob/master/dvc/command/status.py#L19)? Not sure though, how to get the file size

shcheklein:
@AljoSt this a really great feature request. You pointed the right places to do changes. The biggest question is how do we get file sizes. We don't have that functionality right now in the remote class, only `exists` method or `list all` that do check if files exists remotely or returns all files. They both do not provide any additional info currently. So, we would probably need to "teach" them to return more info. 

I suppose, we can start with one of the remotes to try. What do you think, @efiop ?

@AljoSt what remote storage you are mostly interested in?

AljoSt:
@shcheklein I'm currently using S3

shcheklein:
@AljoSt so, in case of S3 (check the `s3.py` file) we use `s3.head_object` and `list_objects`. I think both of them return sizes so we can extend (might be needed to rename `exists` or make a second function) them to returns file sizes.

Suggested TODO (@efiop please review :) ):

- [ ] Check that `s3.head_object` returns actual object size (even if it's a multi-part upload).
- [ ] Check that `s3.list_object` and `s3.list_object_v2` return sizes. Again, need to be sure that it properly returns it in case of multipart.
- [ ] Create a function like `get_file_size` that returns size of the object it if exists.
- [ ] Modify 
- [ ] Use `get_file_size` in the status check in the remote base
- [ ] Modify `cache_exists` (and helpers it calls) in the remote base to deal with sizes.
- [ ] Modify status itself to collect and print sizes.

If I'm not missing anything for this hight level picture we can start with checks and move forward, I'll try to provide more details and help discussing this.

h3ndrk:
@shcheklein We are using an SSH remote and would like to see the sizes in advance. I've not enough knowledge of the code but on a high level for SSH it might be possible to determine the file sizes on the remote in the cache?

shcheklein:
[...]

Issue #2147, opened 2019-06-18
perf: remote indexes


Suor:
As we discussed making dvc fast should be high priority as poor performance can draw people away easily. The big part of todays slowness is working with remotes, which almost always includes collecting file statuses, which could be slow for bigger remotes. All this leads to some form of indexes.

However, our remotes don't provide a luxury of atomic group writes nor reads nor read-modify-write operations. We still can use the following strategy:
- make an index dir on remote, say `index`,
- write a list of files (with names, checksums, mtimes and/or file sizes) to an index file in that dir say `1.idx`
- later when some client needs to update that, e.g. after pushing some new files it:
    - reads current index, 
    - updates it,
    - writes new list to `2.<uuid>.idx`, 
    - removes `1.idx`.
This way in a case of a race we will have several index files.
- if a client needs to read an index it downloads all index files and combines them.

Since we will not only have adds, but also deletes we will need smart combine procudure like in [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type).

File format to be discussed, simple JSON or gzipped JSON with a list of files may do the job though. 

What do you guys think? @shcheklein @dmpetrov @efiop @pared @mroutis

shcheklein:
First, the ideas we discussed before (just to document them and they might spark more experimentation or help focusing on a smaller problem):

1. Use hierarchical structure of the DVC cache as an index. Right now we split into up to 256 directories since we use only first two letters of a checksum, but we can (and probably should anyway to keep directories smaller) introduce at least couple more layers. Thus we will have 2^24 potential subdirectories. To get status we can run multiple parallel `ls`.

2. Use an index/flag file for the directory. The biggest performance bottleneck in our case is happening when we have large directories. Locally, we have done some optimizations (at least we hope they will work) to get complexity of certain operations to O(1). (Namely, we are reconstructing using hardlinks the full "shadow" copy of the directory inside the cache dir. It makes certain operations behave as if would have worked with a single file.) 
Can we do the same trick with remote? Push that json file we have for a directory to the remote or create a special flag-file that which would mean that all files exist? We will potentially need to use some data structures like @Suor mentioned.

Let me know if something above requires some clarifications :)

shcheklein:
Back to the idea of using lock-free self-healing structures. It's hard (for me) to think this all through at once. This stuff can be tricky. Can we first discuss and decide some "data remote model" - what guarantees do we want to provide, what things are okay to happen, what of them are not, etc:

[...]

PR #3600, opened 2020-04-06
[WIP] remote: locally index list of checksums available on cloud remotes

pmrowla:

Related to #2147.

### New behavior

* Checksums available on the remote are indexed (cached) locally in `.dvc/tmp/index/<filename>.idx` where `filename` will be the SHA256 digest of the remote URL.
* On any dvc command which queries remote status (`status -c`/`push`/`pull`/`fetch`/`gc -c`) index will be updated to include any new checksums which are found on the remote (or are pushed successfully to the remote)
* `--drop-index` can be used to delete the index for the remote (and force re-indexing)

### Implementation details

* `remote.cache_exists()`
    1. Validate our index by querying for .dir checksums on the remote. If a .dir checksum is in our index, but not the remote (i.e. someone else has run `gc -c`), clear the entire index.
    2. Look for checksums in the index - If all of the checksums being queried are already in the local index, the remote will not be queried at all.
    3. Any checksums which are not in the local index will be queried on the remote as usual, and the local index will be updated to include any new remote cache entries. (So doing a full list/traverse of the remote will is the same as (re)indexing the entire remote)
* `pull`/`fetch`: If any index related mismatch/error occurs (i.e. trying to download a file that is in our index but not on the remote), clear the entire index.
* `gc -c`: Queries and re-indexes the entire remote, regardless of what was in the index before running `gc -c`.
* Index file will be read once at the start of relevant dvc commands, and written to once at most (only if the index was modified during the dvc command)
* `RemoteIndex` class/module can be easily extended/modified in the future to support any other protocols which support the usual `dump()`/`load()` interface (pickle, json, msgpack, etc)

### Todo
* [ ] Figure out an actual storage solution (currently written in flat binary file format)
* [x] add `--drop-index` CLI option to `push/pull/status` so user can force re-indexing
* [x] Use `.dir` checksum existence on remote to validate/invalidate index
* [x] `push`: push .dir checksums last
* [x] `gc`: remove .dir checksums first
* [x] unit tests for index module
* [ ] functional tests for new behavior
* [ ] docs

### Other ideas:

* Only index .dir checksums to keep local size to a minimum? This was discussed in #2147, if we modify `push` behavior for `.dir` checksums to push the directory file contents first, and the actual `.dir` checksum last, we can treat the dir checksum as an index on its own - if the .dir checksum is on exists on the remote, it follows that the directory contents is also on the remote
* Index file size as well as checksums? (also related to #3568)

### DIFF
--- a/dvc/command/data_sync.py
+++ b/dvc/command/data_sync.py
@@ -30,6 +30,7 @@ class CmdDataPull(CmdDataBase):
                 with_deps=self.args.with_deps,
                 force=self.args.force,
                 recursive=self.args.recursive,
+                drop_index=self.args.drop_index,
             )
             log_summary(stats)
         except (CheckoutError, DvcException) as exc:
@@ -53,6 +54,7 @@ class CmdDataPush(CmdDataBase):
                 all_commits=self.args.all_commits,
                 with_deps=self.args.with_deps,
                 recursive=self.args.recursive,
+                drop_index=self.args.drop_index,
             )
         except DvcException:
             logger.exception("failed to push data to the cloud")
@@ -158,6 +160,12 @@ def add_parser(subparsers, _parent_parser):
         default=False,
         help="Pull cache for subdirectories of the specified directory.",
     )
+    pull_parser.add_argument(
+        "--drop-index",
+        action="store_true",
+        default=False,
+        help="Drop local index for the specified remote cache.",
+    )
     pull_parser.set_defaults(func=CmdDataPull)
 
     # Push
@@ -206,6 +214,12 @@ def add_parser(subparsers, _parent_parser):
         action="store_true",
         default=False,
         help="Push cache for subdirectories of specified directory.",
+    )
+    push_parser.add_argument(
+        "--drop-index",
+        action="store_true",
+        default=False,
+        help="Drop local index for the specified remote cache.",
     )
     push_parser.set_defaults(func=CmdDataPush)
 
@@ -323,5 +337,11 @@ def add_parser(subparsers, _parent_parser):
         action="store_true",
         default=False,
         help="Show status for all dependencies of the specified target.",
+    )
+    status_parser.add_argument(
+        "--drop-index",
+        action="store_true",
+        default=False,
+        help="Drop local index for the specified remote cache.",
     )
     status_parser.set_defaults(func=CmdDataStatus)
--- a/dvc/command/data_sync.py
+++ b/dvc/command/data_sync.py
@@ -164,7 +164,7 @@ def add_parser(subparsers, _parent_parser):
         "--drop-index",
         action="store_true",
         default=False,
-        help="Drop local index for the specified remote cache.",
+        help="Drop local index for the specified remote.",
     )
     pull_parser.set_defaults(func=CmdDataPull)
 
--- a/dvc/command/data_sync.py
+++ b/dvc/command/data_sync.py
@@ -219,7 +219,7 @@ def add_parser(subparsers, _parent_parser):
         "--drop-index",
         action="store_true",
         default=False,
-        help="Drop local index for the specified remote cache.",
+        help="Drop local index for the remote.",
     )
     push_parser.set_defaults(func=CmdDataPush)
 
@@ -342,6 +342,6 @@ def add_parser(subparsers, _parent_parser):
         "--drop-index",
         action="store_true",
         default=False,
-        help="Drop local index for the specified remote cache.",
+        help="Drop local index for the remote.",
     )
     status_parser.set_defaults(func=CmdDataStatus)
--- a/dvc/data_cloud.py
+++ b/dvc/data_cloud.py
@@ -66,6 +66,7 @@ class DataCloud(object):
                 By default remote from core.remote config option is used.
             show_checksums (bool): show checksums instead of file names in
                 information messages.
+            drop_index (bool): clear local index for the remote
         """
         return self.repo.cache.local.push(
             caches,
@@ -93,6 +94,7 @@ class DataCloud(object):
                 By default remote from core.remote config option is used.
             show_checksums (bool): show checksums instead of file names in
                 information messages.
+            drop_index (bool): clear local index for the remote
         """
         remote = self.get_remote(remote, "pull")
         downloaded_items_num = self.repo.cache.local.pull(
@@ -143,6 +145,7 @@ class DataCloud(object):
                 is used.
             show_checksums (bool): show checksums instead of file names in
                 information messages.
+            drop_index (bool): clear local index for the remote
         """
         remote = self.get_remote(remote, "status")
         return self.repo.cache.local.status(
--- a/dvc/data_cloud.py
+++ b/dvc/data_cloud.py
@@ -48,7 +48,14 @@ class DataCloud(object):
     def _init_remote(self, remote):
         return Remote(self.repo, name=remote)
 
-    def push(self, caches, jobs=None, remote=None, show_checksums=False):
+    def push(
+        self,
+        caches,
+        jobs=None,
+        remote=None,
+        show_checksums=False,
+        drop_index=False,
+    ):
         """Push data items in a cloud-agnostic way.
 
         Args:
@@ -65,9 +72,17 @@ class DataCloud(object):
             jobs=jobs,
             remote=self.get_remote(remote, "push"),
             show_checksums=show_checksums,
+            drop_index=drop_index,
         )
 
-    def pull(self, caches, jobs=None, remote=None, show_checksums=False):
+    def pull(
+        self,
+        caches,
+        jobs=None,
+        remote=None,
+        show_checksums=False,
+        drop_index=False,
+    ):
         """Pull data items in a cloud-agnostic way.
 
         Args:
@@ -81,7 +96,11 @@ class DataCloud(object):
         """
         remote = self.get_remote(remote, "pull")
         downloaded_items_num = self.repo.cache.local.pull(
-            caches, jobs=jobs, remote=remote, show_checksums=show_checksums
+            caches,
+            jobs=jobs,
+            remote=remote,
+            show_checksums=show_checksums,
+            drop_index=drop_index,
         )
 
         if not remote.verify:
@@ -105,7 +124,14 @@ class DataCloud(object):
                     # (see `RemoteBASE.download()`)
                     self.repo.state.save(cache_file, checksum)
 
-    def status(self, caches, jobs=None, remote=None, show_checksums=False):
+    def status(
+        self,
+        caches,
+        jobs=None,
+        remote=None,
+        show_checksums=False,
+        drop_index=False,
+    ):
         """Check status of data items in a cloud-agnostic way.
 
         Args:
@@ -120,5 +146,9 @@ class DataCloud(object):
         """
         remote = self.get_remote(remote, "status")
         return self.repo.cache.local.status(
-            caches, jobs=jobs, remote=remote, show_checksums=show_checksums
+            caches,
+            jobs=jobs,
+            remote=remote,
+            show_checksums=show_checksums,
+            drop_index=drop_index,
         )
--- a/dvc/data_cloud.py
+++ b/dvc/data_cloud.py
@@ -48,11 +48,12 @@ class DataCloud(object):
     def _init_remote(self, remote):
         return Remote(self.repo, name=remote)
 
-    def push(self, cache, jobs=None, remote=None, show_checksums=False):
+    def push(self, caches, jobs=None, remote=None, show_checksums=False):
         """Push data items in a cloud-agnostic way.
 
         Args:
-            cache (NamedCache): named checksums to push to the cloud.
+            caches (list): list of (dir_cache, file_cache) tuples containing
+                named checksums to push to the cloud.
             jobs (int): number of jobs that can be running simultaneously.
             remote (dvc.remote.base.RemoteBASE): optional remote to push to.
                 By default remote from core.remote config option is used.
@@ -60,17 +61,18 @@ class DataCloud(object):
                 information messages.
         """
         return self.repo.cache.local.push(
-            cache,
+            caches,
             jobs=jobs,
             remote=self.get_remote(remote, "push"),
             show_checksums=show_checksums,
         )
 
-    def pull(self, cache, jobs=None, remote=None, show_checksums=False):
+    def pull(self, caches, jobs=None, remote=None, show_checksums=False):
         """Pull data items in a cloud-agnostic way.
 
         Args:
-            cache (NamedCache): named checksums to pull from the cloud.
+            caches (list): list of (dir_cache, file_cache) tuples containing
+                named checksums to pull from the cloud.
             jobs (int): number of jobs that can be running simultaneously.
             remote (dvc.remote.base.RemoteBASE): optional remote to pull from.
                 By default remote from core.remote config option is used.
@@ -79,28 +81,36 @@ class DataCloud(object):
         """
         remote = self.get_remote(remote, "pull")
         downloaded_items_num = self.repo.cache.local.pull(
-            cache, jobs=jobs, remote=remote, show_checksums=show_checksums
+            caches, jobs=jobs, remote=remote, show_checksums=show_checksums
         )
 
         if not remote.verify:
-            self._save_pulled_checksums(cache)
+            self._save_pulled_checksums(caches)
 
         return downloaded_items_num
 
     def _save_pulled_checksums(self, cache):
-        for checksum in cache["local"].keys():
-            cache_file = self.repo.cache.local.checksum_to_path_info(checksum)
-            if self.repo.cache.local.exists(cache_file):
-                # We can safely save here, as existing corrupted files will be
-                # removed upon status, while files corrupted during download
-                # will not be moved from tmp_file (see `RemoteBASE.download()`)
-                self.repo.state.save(cache_file, checksum)
+        for dir_cache, file_cache in cache:
+            checksums = set(file_cache["local"].keys())
+            if dir_cache is not None:
+                checksums.update(dir_cache["local"].keys())
+            for checksum in checksums:
+                cache_file = self.repo.cache.local.checksum_to_path_info(
+                    checksum
+                )
+                if self.repo.cache.local.exists(cache_file):
+                    # We can safely save here, as existing corrupted files will
+                    # be removed upon status, while files corrupted during
+                    # download will not be moved from tmp_file
+                    # (see `RemoteBASE.download()`)
+                    self.repo.state.save(cache_file, checksum)
 
-    def status(self, cache, jobs=None, remote=None, show_checksums=False):
+    def status(self, caches, jobs=None, remote=None, show_checksums=False):
         """Check status of data items in a cloud-agnostic way.
 
         Args:
-            cache (NamedCache): named checksums to check status for.
+            caches (list): list of (dir_cache, file_cache) tuples containg
+                named checksums to check status for.
             jobs (int): number of jobs that can be running simultaneously.
             remote (dvc.remote.base.RemoteBASE): optional remote to compare
                 cache to. By default remote from core.remote config option
@@ -110,5 +120,5 @@ class DataCloud(object):
         """
         remote = self.get_remote(remote, "status")
         return self.repo.cache.local.status(
-            cache, jobs=jobs, remote=remote, show_checksums=show_checksums
+            caches, jobs=jobs, remote=remote, show_checksums=show_checksums
         )
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -11,6 +11,8 @@ from functools import partial
 from multiprocessing import cpu_count
 from operator import itemgetter
 
+from funcy import split
+
 from shortuuid import uuid
 
 import dvc.prompt as prompt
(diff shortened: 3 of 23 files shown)


pmrowla:
Regarding the actual index storage format, I looked into using [bcolz](https://github.com/Blosc/bcolz) which was suggested in the remote indexes discussion issue, but it's only available as a source distribution when installing w/pip (no binary wheels for any platform), and trying to build fails on my machine (osx catalina). There are conda binaries available though.

tests/unit/remote/test_index.py
```diff
@@ -0,0 +1,36 @@
+import pickle
+import os.path
+
+from dvc.remote.index import RemoteIndex
+from tests.basic_env import TestDvc
+
+
+class TestRemoteIndex(TestDvc):
```
efiop, line 8: Let's use pytest-style tests. TestDvc is obsoleted and will be removed soon, we are fully migrating to pytest.

dvc/repo/__init__.py
```diff
@@ -85,6 +85,8 @@ def __init__(self, root_dir=None):
 
         self.tmp_dir = os.path.join(self.dvc_dir, "tmp")
         makedirs(self.tmp_dir, exist_ok=True)
+        self.index_dir = os.path.join(self.dvc_dir, "index")
+        makedirs(self.index_dir, exist_ok=True)
```
efiop, line 89: Forgetting to gitignore it :) But actually, maybe let's use `.dvc/tmp/index` instead?

dvc/remote/index.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @property
+    def checksums(self):
+        return self._checksums
+
+    def load(self):
+        """(Re)load this index from disk."""
+        if self.path and self.path.exists():
+            try:
+                with open(self.path, "rb") as fd:
```
efiop, line 38:
```suggestion
                with open(self.path, "rb") as fobj:
```
let's use `fobj`, since `fd` is misleading. We have a mix all over the code, but `fobj` is the preferred one for the new code.

dvc/remote/index.py
```diff
@@ -0,0 +1,78 @@
...
+    """Class for locally indexing remote checksums.
+
+    Args:
+        repo: repo for this remote
+    """
+
+    INDEX_SUFFIX = ".idx"
+
+    def __init__(self, repo, name):
+        if repo and hasattr(repo, "index_dir") and name:
```
efiop, line 18: I guess some tests were failing?

dvc/remote/base.py
```diff
@@ -111,6 +113,13 @@ def __init__(self, repo, config):
         self.cache_types = config.get("type") or copy(self.DEFAULT_CACHE_TYPES)
         self.cache_type_confirmed = False
 
+        url = config.get("url")
+        if url:
+            index_name = hashlib.md5(url.encode("utf-8")).hexdigest()
```
efiop, line 118: Let's use sha256. It doesn't really matter, but we've been getting people lurking through our code and saying that the fact that dvc uses md5s doesn't help it to get accepted in the company for security reasons. So we might as well use sha256 to look better here :) Sure, we have a lot of other places where we need to get rid of md5, but it is a good start :slightly_smiling_face:

dvc/remote/index.py
```diff
@@ -0,0 +1,78 @@
...
+                logger.error(
+                    "Failed to load remote index from '{}'".format(self.path)
+                )
+
+    def save(self):
+        """Save this index to disk."""
+        if self.path:
+            try:
+                with open(self.path, "wb") as fd:
+                    pickle.dump(self._checksums, fd)
```
efiop, line 50: What are the compatibility guarantees for pickle, btw? Like with python versions and backward/forward compatibility.

dvc/remote/index.py
pmrowla: Yeah, there are some remote specific unit/func tests that pass `None` or mocked objects in place of a valid repo

dvc/remote/base.py
```diff
@@ -745,8 +754,11 @@ def gc(self, named_cache, jobs=None):
             path_info = self.checksum_to_path_info(checksum)
             if self.is_dir_checksum(checksum):
                 self._remove_unpacked_dir(checksum)
+            self.index.remove(checksum)
```
efiop, line 757: Wonder how fast this is, esp when we are running this for each hash. Maybe worth just dropping index and reindexing the next time?

dvc/remote/index.py
```diff
@@ -0,0 +1,78 @@
...
+            except IOError:
+                logger.error(
+                    "Failed to save remote index to '{}'".format(self.path)
+                )
```
efiop, line 54: Looks like we might want to ignore something more specific.

dvc/remote/index.py
efiop: @pmrowla Ideally it would be great to fix those tests to mock it properly, but I'm fine with this approach for now. IIRC we have a similar piece of code elsewhere (probably not even one). :+1:

dvc/remote/index.py
pmrowla: By default, any current python (up through 3.8) writes a protocol version that is compatible with >=3.4. There is a new protocol version that's only readable in 3.8, but it's not the default.

I'm also just using pickle as a WIP placeholder right now, I don't think it's actually the right solution here

shcheklein:
One question: what happens if someone else runs `dvc gc`? do we check for this somehow. (I remember we had some ideas about directories at least - using `.dir` file as a flag).

pmrowla:
> One question: what happens if someone else runs `dvc gc`? do we check for this somehow. (I remember we had some ideas about directories at least - using `.dir` file as a flag).

Right now we will just clear the entire local index the first time we try to pull something from the remote that isn't actually there. Taking advantage of the `.dir` files in some way is planned though

shcheklein:
So, what about `push` (and e.g. `status`) in this case? Will it skip files that it thinks (according to the local index cache) that already exists remotely?

pmrowla:
As is right now, yes it will skip pushing files that are in the local index (and we would not see that someone else's `gc -c` operation had removed them from the remote).

One solution to this would be to at least check for `.dir` files on the remote no matter what. But in the event that there are enough `.dir` checksums to in our query to make us do the full list/traverse of the remote, we wouldn't actually be improving anything in terms of performance.

pmrowla:
One thought I had was that if `.dir` checksums were stored separately from file checksums in our cache (so that `.dir`'s were under their own prefix), we would be able to fetch them without having to potentially query the entire remote cache, but that would be a pretty significant design change

pmrowla:
Discussed this with @efiop , potentially breaking backwards-compatibility and reorganizing our cache structure is something for future discussions

In the meantime we can add a `--drop-index` (or similarly named) CLI option to allow users to force re-indexing. So in the event that the user knows some other team member has run `gc -c`, the user can run `status -c/push/pull --drop-index` and force us to re-index the remote cache.

@shcheklein thoughts?

shcheklein:
It feels to me that for directories it's totally fine to use `exists`? It's quite unusual edge-case to have 1M directories for example. It also means 1M DVC-files?

I would not worry about performance of checking existences of `.dir` files.

I would worry for us not saving some data.

shcheklein:
> In the meantime we can add a --drop-index (or similarly named) CLI option to allow users to force re-indexing. So in the event that the user knows some other team member has run gc -c, the user can run status -c/push/pull --drop-index and force us to re-index the remote cache.

It all feels vary too risky to me. We are playing with data here and need to be conservative.

pmrowla:
CI expected to fail pending merge of #3604

dvc/remote/base.py
```diff
@@ -731,22 +740,32 @@ def all(self, jobs=None, name=None):
...
+                used.update(file_cache[self.scheme])
 
         removed = False
-        for checksum in self.all(jobs, str(self.path_info)):
+        # checksums must be sorted to ensure we always remove .dir files first
+        for checksum in sorted(
+            self.all(jobs, str(self.path_info)),
+            key=self.is_dir_checksum,
+            reverse=True,
+        ):
```
pmrowla, line 758: This is another place we would benefit from having separated .dir and file checksums in our cache structure

dvc/command/data_sync.py
```diff
@@ -158,6 +160,12 @@ def add_parser(subparsers, _parent_parser):
         default=False,
         help="Pull cache for subdirectories of the specified directory.",
     )
+    pull_parser.add_argument(
+        "--drop-index",
+        action="store_true",
+        default=False,
+        help="Drop local index for the specified remote cache.",
```
shcheklein, line 167: `remote cache` -> `remote` (we try to avoid using `cache` with remotes, just to make it less confusing with `.dvc/cache`) ?

dvc/repo/pull.py
```diff
@@ -28,6 +29,7 @@ def pull(
         all_commits=all_commits,
         with_deps=with_deps,
         recursive=recursive,
+        drop_index=drop_index,
```
shcheklein, line 32: should we start using `**kwargs` @pmrowla @efiop ? it seems we have to change a lot of method just to pass one arguments, and it's not very scalable - not ideal to have functions with 9 args?

dvc/remote/index.py
```diff
@@ -0,0 +1,84 @@
...
+    def checksums(self):
+        return self._checksums
+
+    def load(self):
+        """(Re)load this index from disk."""
+        if self.path and os.path.isfile(self.path):
+            try:
+                with open(self.path, "rb") as fobj:
+                    self._checksums = pickle.load(fobj)
+            except PermissionError:
```
shcheklein, line 44: why do we catch only PermissionErrors? not IO in general?

dvc/remote/local.py
```diff
@@ -348,48 +428,121 @@ def _process(
...
+            pushed_checksums = list(
+                map(self.path_to_checksum, dir_plans[0] + file_plans[0])
+            )
+            logger.debug(
+                "Adding {} pushed checksums to index".format(
+                    len(pushed_checksums)
+                )
+            )
+            remote.index.update(pushed_checksums)
+            remote.index.save()
```
shcheklein, line 495: make sure that it is thread/multi-process safe (we do allow multiple DVC commands in parallel) ... check other this new index functions if they are thread safe?

dvc/remote/local.py
```diff
@@ -348,48 +428,121 @@ def _process(
...
+    def _dir_upload(self, func, futures, from_info, to_info, name):
+        for future in as_completed(futures):
+            if future.result():
+                # do not upload this .dir file if any file in this
+                # directory failed to upload
+                logger.debug(
+                    "failed to upload full contents of '{}', "
+                    "aborting .dir file upload".format(name)
+                )
+                logger.error(
```
shcheklein, line 508: include the previous logger here?

pmrowla: The error will already have been caught and logged in the exception handler for the failed upload thread. I'm not sure if it's worth storing the exception info so it can be referenced across threads and logged here?

dvc/remote/index.py
pmrowla: there was an earlier review comment to catch more specific exceptions, I'll fix this to also catch general IOError as well

dvc/repo/pull.py
pmrowla: I think using `**kwargs` would make sense, but we'll want to make sure the expected params are documented properly. I think a lot of the top level function definitions for push/pull/gc/etc are currently missing docstrings.

dvc/remote/index.py
shcheklein: I wonder if we should at least show some message (let's not use WARNING though, please - they look ugly) in some cases? To avoid situation when some logic is broken or not enough permissions (can it happen at all?) and we silently degrade performance significantly.

dvc/remote/local.py
shcheklein: I think my question was mostly about the duplication logger.debug + logger.error - do we really need both?

pmrowla: Ah, my reasoning for separating it was that ".dir file was not uploaded" is relevant to the user, but the specific reason why in this case seemed less relevant outside of of a verbose/debugging output context

shcheklein: kk, up to you, I would personally try something like:

`upload {} to {}, one or more files failed` ?

it does not hurt to show some reason, will help us debug faster, simplifies code (best code == no code :) )

dvc/remote/index.py
pmrowla: I think the only time IO errors would actually happen here is in the event we can't read/write to the `.dvc/tmp` directory at all, whether it's due to permissions, lack of disk space, or some other reason (which I think would cause errors elsewhere in dvc).

dvc/repo/pull.py
pmrowla: created separate issue for this: #3612

dvc/remote/index.py
pmrowla: Updated this so that on an IO error we explicitly state that the remote will be re-indexed, and the message includes the original exception/error message.

It's still logged as an error though, should this be an info message since it's not fatal?

dvc/command/data_sync.py
```diff
@@ -207,6 +215,12 @@ def add_parser(subparsers, _parent_parser):
         default=False,
         help="Push cache for subdirectories of specified directory.",
     )
+    push_parser.add_argument(
+        "--drop-index",
+        action="store_true",
+        default=False,
+        help="Drop local index for the specified remote cache.",
```
shcheklein, line 222: same here - let's not use `cache` .. @jorgeorpinel could please review messages, btw?

dvc/command/data_sync.py
```diff
@@ -324,4 +338,10 @@ def add_parser(subparsers, _parent_parser):
         default=False,
         help="Show status for all dependencies of the specified target.",
     )
+    status_parser.add_argument(
+        "--drop-index",
+        action="store_true",
+        default=False,
+        help="Drop local index for the specified remote cache.",
```
shcheklein, line 345: .... and here

pmrowla:
Pushed WIP changes using the following index file format:
```
# Index format (v1) is the following:
#
# (all integer types are little-endian)
#
# Header
# ------
#   protocol_version (32-bit uint): index file protocol version
#   dir_checksum_count (64-bit uint): number of .dir checksums in data section
#   file_checksum_count (64-bit uint): number of file checksums in data section
#   compressed_len (64-bit uint): length of data section
#   crc - 32-bit CRC32 checksum of uncompressed data
#
# Data (zlib compressed)
# ----------------------
#   array of <dir_checksum_count> 128-bit MD5 .dir checksums
#   array of <file_checksum_count> 128-bit MD5 file checksums
```
There's currently a bug in loading files w/large checksum counts that causes one of the unit tests to fail.

For reading and writing an index with 1M checksums, python timeit reports:
```
dump(): 0.78s
load(): 0.60s
```

When testing with a random distribution of 1M checksums, zlib does not actually compress very well (and the final compressed file is usually still close to 16MB), so the read/write performance can be improved by just dropping the compression entirely.

dvc/command/data_sync.py
jorgeorpinel: Agree. We prefer "remote storage" or just "remote" as of now.

dvc/command/data_sync.py
```diff
@@ -158,6 +160,12 @@ def add_parser(subparsers, _parent_parser):
         default=False,
         help="Pull cache for subdirectories of the specified directory.",
     )
+    pull_parser.add_argument(
+        "--drop-index",
+        action="store_true",
+        default=False,
+        help="Drop local index for the specified remote.",
```
jorgeorpinel, line 167: Not sure about "specified remote" since sync commands don't always receive a -r arg. Maybe just remove the word "specified".

Also, "drop index" may not mean anything to a casual user, should this be a little more descriptive?

jorgeorpinel: Same in the other 2 sync commands.

dvc/data_cloud.py
```diff
@@ -48,29 +48,46 @@ def get_remote(self, remote=None, command="<command>"):
...
         return Remote(self.repo, name=remote)
 
-    def push(self, cache, jobs=None, remote=None, show_checksums=False):
+    def push(
+        self,
+        caches,
+        jobs=None,
+        remote=None,
+        show_checksums=False,
+        drop_index=False,
```
jorgeorpinel, line 57: Add `drop_index` to docstring Args?
Same in other functions/methods.

dvc/data_cloud.py
```diff
@@ -79,28 +96,47 @@ def pull(self, cache, jobs=None, remote=None, show_checksums=False):
...
+            for checksum in checksums:
+                cache_file = self.repo.cache.local.checksum_to_path_info(
+                    checksum
+                )
+                if self.repo.cache.local.exists(cache_file):
+                    # We can safely save here, as existing corrupted files will
+                    # be removed upon status, while files corrupted during
+                    # download will not be moved from tmp_file
+                    # (see `RemoteBASE.download()`)
+                    self.repo.state.save(cache_file, checksum)
```
jorgeorpinel, line 125: Comment getting long 😋

dvc/remote/index.py
```diff
@@ -0,0 +1,257 @@
...
+        protocol = DEFAULT_PROTOCOL
+    if protocol not in SUPPORTED_PROTOCOLS:
+        raise DvcException(
+            "unsupported remote index protocol version: {}".format(protocol)
+        )
+    return protocol
+
+
+def dump(dir_checksums, file_checksums, fobj, protocol=None):
+    """Write specified checksums to the open file object ``fobj``."""
```
jorgeorpinel, line 49: Question: what are double back quotes signaling here? Just curious

dvc/remote/index.py
```diff
@@ -0,0 +1,257 @@
...
+_header_v1 = struct.Struct("<IQQ")
+_compress_crc_v1 = struct.Struct("<QI")
+
+
+def _verify_protocol(protocol=None):
+    if protocol is None:
+        protocol = DEFAULT_PROTOCOL
+    if protocol not in SUPPORTED_PROTOCOLS:
+        raise DvcException(
+            "unsupported remote index protocol version: {}".format(protocol)
```
jorgeorpinel, line 43:
```suggestion
            "unsupported remote index protocol version: '{}'".format(protocol)
```
AFAIR we use `'` around most dynamic output.
There's a few more instances of this.

dvc/remote/index.py
```diff
@@ -0,0 +1,257 @@
...
+
+class RemoteIndex(object):
+    """Class for locally indexing remote checksums.
+
+    Args:
+        repo: repo for this remote index.
+        name: name for this index. If name is provided, this index will be
+            loaded from and saved to ``.dvc/tmp/index/{name}.idx``.
+            If name is not provided (i.e. for local remotes), this index will
+            be kept in memory but not saved to disk.
```
jorgeorpinel, line 148: Should this go in a constructor docstring instead?

dvc/remote/index.py
pmrowla: Old habit from writing sphinx/rst docstrings, I'll clean them up

shcheklein:
@pmrowla looks great, but a little bit complicated :)

could you update the description, please to get the up to date list of things that we do with this PR?

dvc/remote/index.py
pmrowla: afaik standard python conventions allow constructor args to be documented in either `__init__` or in the class declaration like here. I put them here because that's how it was done in some other dvc classes (see `dvc/cache.py::Cache`)

dvc/command/data_sync.py
pmrowla: How about `--clear-index` or `--reset-index`?

pmrowla:
> @pmrowla looks great, but a little bit complicated :)
> 
> could you update the description, please to get the up to date list of things that we do with this PR?

@shcheklein it's been updated.

I removed the compression stuff from the index file format since it's not needed, so the index file read/write code is simplified now.

dvc/remote/base.py
```diff
@@ -731,22 +742,31 @@ def all(self, jobs=None, name=None):
...
             if checksum in used:
                 continue
             path_info = self.checksum_to_path_info(checksum)
             if self.is_dir_checksum(checksum):
                 self._remove_unpacked_dir(checksum)
             self.remove(path_info)
             removed = True
+        if removed:
+            self.index.invalidate()
```
pmrowla, line 769: Thinking about this more, this is unnecessary.

`gc -c` will always fetch the full remote listing via `all()` (regardless of what is in our index). So we can treat `gc -c` as a full re-indexing of the remote. After `gc -c` our index should contain the full remote listing minus the unused checksums which were removed.

pmrowla:
After discussion with @efiop, we decided that a flat format (whether binary or json/msgpack/etc) won't be ideal for dealing with large file/checksum counts. And we will also ideally want random access support for cases like running push/pull with a small number of files.

A portable database format like sqlite may work for our needs. Will do some testing to see what kind of runtime performance and index file size we would get with sqlite.

dvc/repo/fetch.py
```diff
@@ -82,18 +106,20 @@ def _fetch_external(self, repo_url, repo_rev, files, jobs):
             if is_dvc_repo:
                 repo.cache.local.cache_dir = self.cache.local.cache_dir
                 with repo.state:
-                    cache = NamedCache()
+                    used_cache = []
```
efiop, line 109: Hm, not sure I understand why this is needed.

pmrowla:
This is going to be split into multiple PRs, the .dir checksum push/pull/gc behavior changes will come first

Issue #1871, opened 2019-04-10
store whole DAG in one DVC-file


Casyfill:
I understand the merits of having multiple `.dvc` files for complex processes,
but it would be just great to have the option to store the whole DAG in one Dvcfile!

I feel it might help the overall readability of the structure

mroutis:
I might be overestimating this, but it looks to me that we need to do a huge refactor to support this, or at least not treating `Dvcfile` as a stage and create a separate class to deal with the logic of parsing such file to come up with a pipeline.

I can see the analogy with `make` on this one, but from my perspective, `dvc` is a CLI tool to read and write (stage) files, whereas `make` tool only reads from (make) files.

I do appreciate the _aesthetics_ of having everything on a single file, tho.

Casyfill:
I see.
as a last resort, perhaps it is at least possible to add paths to the previous steps as comments in the file?

shcheklein:
@Casyfill I think we will release a new version very soon (next week? @efiop can confirm) that will be preserving comments in the DVC files. I'm not sure we should be generating these comments on the DVC side though - it's hard to keep them up to date. What editor are using? May be it's a good feature to support on the editor level - navigate up/down the pipeline cc @prihoda

shcheklein:
Just for the context - we got a request on the ods.ai #tool_dvc channel to support this:

(translation is mine)

```
Hi! Is there a way (feature/third-party tool/script/...) to describe the whole dvc-pipeline as a single file (as it is happening in build tools - make, bazel, maven, etc) instead of having a file per stage
```

salotz:
> I can see the analogy with make on this one, but from my perspective, dvc is a CLI tool to write, read and execute (stage) files, whereas make is reduced to read and control de execution of (make) files.

The single-file approach could be built upon the CLI approach as an extra optional layer. I suggested something like [tup](http://gittup.org/tup/) as another syntax option to make, which I happen to like. It inverts the dependency graph to be a flow graph and they look like unix pipes. It also can be faster due to the way it checks for updates.

> I do appreciate the aesthetics of having everything on a single file, tho.

It is more than just about aesthetics (although great aesthetics in a build tool would be nice). Its a complex problem and [waterbed theory](https://en.wikipedia.org/wiki/Waterbed_theory) is real.
[...]

Issue #1234, opened 2018-10-17
repro: use build cache for deterministic stages


BartekRoszak:
In my experiment I run a few different preprocessing steps which create a different CSV file, then I am modeling this data and also checking different parameters.
When I want to run the same experiment on 4 different machines (`dvc` is connected to the same remote cache).
Running every type of preprocessing will be done on every machine which takes a lot of time and could be omitted by running `dvc pull` before `dvc repro` and `dvc push` after it. 
It could work with one command like `dvc repro --remote`

efiop:
Hi @AratorField !

Thank you for sharing your scenario! It would be indeed a very useful feature! I think that `dvc checkout` would've benefited from it as well. How about instead something like a `autofetch` option for dvc config, that will tell dvc that it should fetch cache before performing operations such as repro and checkout? Would that be suitable for you?

Thanks,
Ruslan

dmpetrov:
@efiop the config option will change API. It won't be clear what exactly `dvc repro` does from code. An option `dvc repro --fetch` might be a good option instead of global config param.

AratorField:
I am running `dvc repro` in a loop so something like `-f -p / --fetch --push` will be ok for me. 
Setting in config `autofetch` and `autopush` is ok too.

AratorField:
I am doing it like you said when I run an experiment on a server where I run most of my experiments. But then I am reproducing manually some of the best experiments locally.

efiop:
Ah, I see. So it looks something like:

```
git pull
dvc pull
dvc repro
git commit ...
git push
dvc push
```

Right? If so, I think a more suitable approach would be to add git hooks that will call `dvc pull` after `git pull` and `dvc push` after `git push`, similar to what we currently have with `dvc checkout`, where you can call `dvc install` to install a post-checkout hook for you so `dvc checkout` is automatically called after `git checkout`. Would that suit you?

AratorField:
Oh... So it works totally differently than I thought. Now I know why my algorithm is doing preprocessing every time.

Let me explain you a bit more.
I am building about 100 models with different parameters (standard hyperparameter optimization). 
Before data modeling there is preprocessing, so in `dvc repro` has two steps, preprocessing and modeling.
I am working with text and in preprocessing I have script which can do two things:
1. convert all to lowercase
2. remove punctuation

So there could be 4 different types of preprocessing:
- None
- convert all to lowercase
- remove punctuation
- convert all to lowercase and remove punctuation

[...]

PR #3603, opened 2020-04-07
dvc: introduce local stage cache

efiop:
This patch introduces `.dvc/cache/stages` that is used to store previous
runs and their results, which could then be reused later when we stumble
upon the same command with the same deps and outs.

Format of build cache entries is single-line json, which is readable by
humans and might also be used for lock files discussed in #1871.

Related to #1871
Local part of #1234

### DIFF
--- a/dvc/repo/reproduce.py
+++ b/dvc/repo/reproduce.py
@@ -97,12 +97,7 @@ def reproduce(
 
 
 def _reproduce_stages(
-    G,
-    stages,
-    downstream=False,
-    ignore_build_cache=False,
-    single_item=False,
-    **kwargs
+    G, stages, downstream=False, single_item=False, **kwargs
 ):
     r"""Derive the evaluation of the given node for the given graph.
 
@@ -172,7 +167,7 @@ def _reproduce_stages(
         try:
             ret = _reproduce_stage(stage, **kwargs)
 
-            if len(ret) != 0 and ignore_build_cache:
+            if len(ret) != 0 and kwargs.get("ignore_build_cache", False):
                 # NOTE: we are walking our pipeline from the top to the
                 # bottom. If one stage is changed, it will be reproduced,
                 # which tells us that we should force reproducing all of
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -72,6 +72,7 @@ class Repo(object):
         from dvc.repo.params import Params
         from dvc.scm.tree import WorkingTree
         from dvc.utils.fs import makedirs
+        from dvc.stage.cache import StageCache
 
         root_dir = self.find_root(root_dir)
 
@@ -103,6 +104,8 @@ class Repo(object):
 
         self.cache = Cache(self)
         self.cloud = DataCloud(self)
+
+        self.stage_cache = StageCache(self.cache.local.cache_dir)
 
         self.metrics = Metrics(self)
         self.params = Params(self)
--- a/dvc/repo/run.py
+++ b/dvc/repo/run.py
@@ -68,6 +68,9 @@ def run(self, fname=None, no_exec=False, **kwargs):
         raise OutputDuplicationError(exc.output, set(exc.stages) - {stage})
 
     if not no_exec:
-        stage.run(no_commit=kwargs.get("no_commit", False))
+        stage.run(
+            no_commit=kwargs.get("no_commit", False),
+            ignore_build_cache=kwargs.get("ignore_build_cache", False),
+        )
     dvcfile.dump(stage, update_pipeline=True)
     return stage
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -491,6 +491,8 @@ class Stage(params.StageParams):
 
         self.md5 = self._compute_md5()
 
+        self.repo.stage_cache.save(self)
+
     @staticmethod
     def _changed_entries(entries):
         return [
@@ -617,7 +619,9 @@ class Stage(params.StageParams):
             raise StageCmdFailedError(self)
 
     @rwlocked(read=["deps"], write=["outs"])
-    def run(self, dry=False, no_commit=False, force=False):
+    def run(
+        self, dry=False, no_commit=False, force=False, ignore_build_cache=False
+    ):
         if (self.cmd or self.is_import) and not self.locked and not dry:
             self.remove_outs(ignore_remove=False, force=False)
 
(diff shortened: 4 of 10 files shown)


dvc/stage.py
```diff
@@ -1043,13 +1050,9 @@ def status(self, check_updates=False):
         return {}
 
     def _already_cached(self):
-        return (
-            not self.changed_md5()
```
efiop, line 1047: This was not really needed, as all we care about is cmd its deps (and their hashes) and outs. Things like `always_run` are handled separately.

dvc/stage_cache.py
```diff
@@ -0,0 +1,72 @@
...
+    for out in stage.outs:
+        if not out.def_path:
+            return None
+
+        string += _sha256(out.def_path)
+
+    return _sha256(string)
+
+
+def _get_cache(stage):
```
efiop, line 33: @skshetry Would it make sense to use the same format for the lockfile, btw?

tests/func/test_import.py
```diff
@@ -286,7 +286,7 @@ def test_import_non_existing(erepo_dir, tmp_dir, dvc):
 
     # https://github.com/iterative/dvc/pull/2837#discussion_r352123053
     with pytest.raises(PathMissingError):
-        tmp_dir.dvc.imp(fspath(erepo_dir), "/root/", "root")
```
efiop, line 289: Not sure how this has worked before, since `/root` does exist on linux'es. Might be missing something...

dvc/stage_cache.py
```diff
@@ -0,0 +1,72 @@
...
+        json.dump(_get_cache(stage), fobj)
+
+
+def restore(stage):
+    path = _get_cache_path(stage)
+    if not path or not os.path.exists(path):
+        return
+
+    with open(path, "r") as fobj:
+        cache = json.load(fobj)
```
efiop, line 68: Need to not forget to add a schema too, just to be safe...

efiop: Also hash validation would also be desirable.

dvc/stage_cache.py
```diff
@@ -0,0 +1,72 @@
+import os
+import json
+import hashlib
+
+from dvc.utils.fs import makedirs
+
+
+def _sha256(string):
```
efiop, line 8: Decided to go with sha256 over md5 just to avoid people getting scared by non-cryptographic md5 :slightly_smiling_face:

dvc/stage_cache.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def _get_cache_path(stage):
+    sha = _get_hash(stage)
+    if not sha:
+        return None
+
+    cache_dir = os.path.join(stage.repo.cache.local.cache_dir, "stages",)
+
+    return os.path.join(cache_dir, sha[:2], sha)
```
efiop, line 48: if sha is 123456, then path would be `12/123456` and not `12/3456` as it is in our regular data cache. the reason is do it more like git does, though I don't see any practical advantage right now. Might revert to `12/3456`, just to be consistent.

dvc/stage_cache.py
efiop: Also, need to think if any other fields really matter here for reproducibility. So far it seems like the top-level md5 that we current have in our dvcfiles is really only needed for cmd, as nothing else (except for `always_changed` really matters, unless someone is explicitly trying to do something fishy).

dvc/repo/reproduce.py
```diff
@@ -95,14 +95,7 @@ def reproduce(
...
 
-def _reproduce_stages(
-    G,
-    stage,
-    downstream=False,
-    ignore_build_cache=False,
-    single_item=False,
-    **kwargs
-):
+def _reproduce_stages(G, stage, downstream=False, single_item=False, **kwargs):
```
skshetry, line 98: :+1:

dvc/stage.py
```diff
@@ -794,6 +794,8 @@ def save(self):
 
         self.md5 = self._compute_md5()
 
+        stage_cache.save(self)
```
skshetry, line 797: Maybe, we could make it explicit here that it's only run for pipeline stages?
```suggestion
        if not self.is_data_source:
             stage_cache.save(self)
```

dvc/stage_cache.py
```diff
@@ -0,0 +1,75 @@
...
+    string = _sha256(stage.cmd)
+    for dep in stage.deps:
+        if not dep.def_path or not dep.get_checksum():
+            return None
+
+        string += _sha256(dep.def_path)
+        string += _sha256(dep.get_checksum())
+
+    for out in stage.outs:
+        if not out.def_path or out.persist:
```
skshetry, line 25: Why `out.persist` is None here?

dvc/stage_cache.py
skshetry: +1 for being consistent with our cache.

dvc/stage_cache.py
```diff
@@ -0,0 +1,75 @@
...
+
+    with open(path, "r") as fobj:
+        cache = json.load(fobj)
+
+    outs = {out.def_path: out for out in stage.outs}
+    for def_path, checksum in cache["outs"].items():
+        outs[def_path].checksum = checksum
+
+    for dep in stage.deps:
+        dep.save()
```
skshetry, line 75: Umm, why the `restore` is saving deps?

dvc/stage_cache.py
efiop: @skshetry If there is a persistent output, we can't really cache the stage. We have the same logic in our old build cache in Stage.create().

dvc/stage.py
efiop: Also need to account for imports and others. But yeah, good point. In stage cache itself, we ignore caching if there is no `cmd`, so we should do the same here.

dvc/stage_cache.py
efiop: @skshetry In case you don't have md5s for your deps (e.g. this is your first run). But yeah, could be made nicer. The main reason is the `checkout` part in `Stage.run`, which expects these deps to be already saved in order to checkout the outputs. Will take a look.

dvc/stage_cache.py
charlesbaynham: What about `wdir`?

efiop: @charlesbaynham `wdir` doesn't seem to matter, as cmd and relpaths we have in dvc-file stay the same.

charlesbaynham: Maybe I'm missing the point, but say cmd is

    cat "../data/file.txt" > "../out.txt"

Even if `file.txt` and `out.txt` are unchanged and `cmd` is unchanged too, I'd get a different result by changing `wdir` (i.e. I'd get an error since the file wouldn't be found anymore).

efiop: @charlesbaynham dvcfiles contain paths that are relative to wdir, so for your command `dep` would be `../data/file.txt` and out would be `../out.txt`. And no matter where you move the wdir, if dep is at the same relative path, then output will be at the same relative path.

charlesbaynham: Ah I see what you're saying, sorry for the bother

PR #3619, opened 2020-04-10
refactor: extract stage's relationship with Dvcfile out from Stage

skshetry:

This is just a refactor, which extracts `Stage`'s relationship with actual Dvcfile (with regards to parsing/dumping, validating schema, etc) to a separate abstraction (`Dvcfile`).
Now, `Stage` is just an in-memory representation that can work without in-relation to Dvcfile.

Related to #1871


dvc/stages/__init__.py
```diff
@@ -726,33 +548,7 @@ def dumpd(self):
...
-            # have it. We simply copy it over.
-            if "meta" in saved_state:
-                state["meta"] = saved_state["meta"]
-            apply_diff(state, saved_state)
-            state = saved_state
-
-        dump_stage_file(fname, state)
-
-        self.repo.scm.track_file(relpath(fname))
+        self.dvcfile.dump_single_stage(self)
```
efiop, line 551: Discussed with @skshetry that this is a bit backwards and it makes sense to get rid of `dump` method in Stage and just use Dvcfile.dump, to simplify next steps.

dvc/dvcfile.py
```diff
@@ -0,0 +1,152 @@
+import os
+import re
+import logging
+
+from voluptuous import MultipleInvalid
+
+from dvc import dependency as dependency, output as output
+from dvc.stages import schema
```
efiop, line 8: let's use `stage` instead of `stages` to match what we have in other parts of the code. I.e. `dvc/remote` `dvc/dependency` etc

dvc/stages/__init__.py
```diff
@@ -171,30 +84,109 @@ def wrapper(stage, *args, **kwargs):
     return wrapper
 
 
-class Stage(object):
-    STAGE_FILE = "Dvcfile"
-    STAGE_FILE_SUFFIX = ".dvc"
+def create_stage(repo, accompany_outs=False, **kwargs):
```
skshetry, line 87: - [ ] Create stage into dvcfile.

efiop, line 87: discussed with @skshetry that it is better to create `create_dvcfile` in `dvcfile.py` instead

dvc/repo/update.py
```diff
@@ -3,11 +3,11 @@
...
 def update(self, target, rev=None):
-    from dvc.stage import Stage
+    from ..dvcfile import Dvcfile
 
-    stage = Stage.load(self, target)
+    stage = Dvcfile(self, target).load()
     stage.update(rev)
 
-    stage.dump()
+    stage.dvcfile.dump(stage)
```
efiop, line 11: Feels like it is still backwards. In your upcoming PR dvcfile will take the central place, so why not do it justice and use properly:
```
dvcfile = Dvcfile(self, target)
dvcfile.stage.update()
dvcfile.dump()
```

We've discussed `stage.dvcfile.dump()` more as a hack for old tests to avoid messing around with them too much, but the core code should handle this properly right away.

skshetry: For the multistage, it makes more sense to have `dvcfile.dump(stage)`. At the moment, Dvcfile only have a single stage looks like a special case to me (and, hence I am generalizing).

efiop: I agree, but it is a good opportunity to get ready for that. Even for single-stage, the logic is that Dvcfile writes and reads stuff, not the stage, so we need to use it properly right away. `stage.dvcfile.dump(stage)` is extremely ugly for the core code.

efiop: @skshetry Or something like
```
dvcfile = Dvcfile(self, target)
stage = dvcfile.load()
stage.update()
dvcfile.dump(stage)
```
in that case. Just want to get the logic straight right away.

skshetry: > @skshetry Or something like

Yup, I went with something like that.

skshetry: But, writing `dvcfile = Dvcfile(self, target)` is still cumbersome.

skshetry:
Travis is offline for maintenance at the moment. /cc @efiop

dvc/repo/reproduce.py
```diff
@@ -21,7 +22,9 @@ def _reproduce_stage(stage, **kwargs):
         return []
 
     if not kwargs.get("dry", False):
-        stage.dump()
+        print("stage: ", stage.path)
```
efiop, line 25: Why is this print needed?

dvc/repo/reproduce.py
```diff
@@ -58,7 +61,7 @@ def reproduce(
     all_pipelines=False,
     **kwargs
 ):
-    from dvc.stage import Stage
+    from ..dvcfile import Dvcfile
```
efiop, line 64: Already imported above. Better to keep these import dynamic, same as they were before.

dvc/repo/reproduce.py
skshetry: Ooops, was debugging a failed test :smile: 
```suggestion
```

dvc/repo/run.py
```diff
@@ -5,9 +5,9 @@
 @locked
 @scm_context
 def run(self, no_exec=False, **kwargs):
-    from dvc.stage import Stage
+    from dvc.dvcfile import Dvcfile
 
-    stage = Stage.create(self, **kwargs)
+    stage = Dvcfile.create_stage(self, **kwargs)
```
efiop, line 10: Again, this feels weird (esp because of `Dvcfile(self, stage.path).dump(stage)` down below). The multistage PR still looks strange https://github.com/iterative/dvc/pull/3584/files#r407061889 . We could definitely do better.

dvc/dvcfile.py
```diff
@@ -0,0 +1,255 @@
...
+    def validate(d, fname=None):
+        from dvc.stage.schema import SINGLE_STAGE_SCHEMA
+
+        try:
+            SINGLE_STAGE_SCHEMA(d)
+        except MultipleInvalid as exc:
+            raise StageFileFormatError(fname, exc)
+
+    @classmethod
+    def create_stage(cls, repo, accompany_outs=False, **kwargs):
```
efiop, line 163: Pretty much the whole logic doesn't belong to dvcfile itself, but belongs to the Stage class. Seems like the only thing that `Dvcfile` should care about here is the filename.

efiop: E.g. it should care about `if os.path.exist` at the bottom. So, for example, it could be `Dvcfile.create()` method, that would in turn call `Stage.create()` with all that meat and then would check for already existing dvcfile with a path like that.

skshetry: It needs to be somewhere in between of those two. `Dvcfile` should care about filename, if it's a multistage dvc, if stage already exists with the given name, etc.
Rest can be taken care by `Stage`, but it's not really linear at the time. :(

efiop: @skshetry Yes, that was my point. Right now `create_stage` does too much with the Stage. It should pass to `Stage.create()` and deal with dvcfile file/existing dvcfile/etc.

efiop: Took a closer look. Looks like `Dvcfile.create` should handle things like filename/wdir validation. E.g. things like `_stage_name` and `_check_stage_path` don't belong to `Stage` anymore, they should be handled in `Dvcfile`.

dvc/command/run.py
```diff
@@ -47,7 +47,6 @@ def run(self):
                 no_exec=self.args.no_exec,
                 overwrite=overwrite,
                 ignore_build_cache=self.args.ignore_build_cache,
-                remove_outs=self.args.remove_outs,
```
skshetry, line 50: Removed this deprecated flag.

dvc/dvcfile.py
```diff
@@ -0,0 +1,174 @@
...
+
+    @classmethod
+    def is_valid_filename(cls, path):
+        return (
+            path.endswith(DVC_FILE_SUFFIX)
+            or os.path.basename(path) == DVC_FILE
+        )
+
+    @classmethod
+    def is_stage_file(cls, path):
```
skshetry, line 48: This kind of naming will be changing on upcoming PRs. For now, stage_file and dvcfile are essentially the same thing and should work.

dvc/dvcfile.py
```diff
@@ -0,0 +1,174 @@
...
+        return os.path.isfile(path) and cls.is_valid_filename(path)
+
+    @classmethod
+    def check_dvc_filename(cls, path):
+        if not cls.is_valid_filename(path):
+            raise StageFileBadNameError(
```
skshetry, line 54: Again, exceptions will probably change wherever this does not make sense in future PRs.

dvc/dvcfile.py
skshetry: Did a bit of refactor. Most of the things stays on `Stage::create()`. Only overwriting stuffs have been pulled to `Dvcfile`.

efiop:
We might've tried to bite more than we could chew here. Even though the arch is far from ideal yet, this PR does a lot of good refactorings, so let's merge it and move on, we'll introduce better interfaces this week.

PR #3626, opened 2020-04-12
Specify cmd arg names (metavars), et al.

### LINKED ISSUE #3608: Consistent subcommand order
Help output of commands with subcommands have a seemingly random order. I propose this (matches docs repo nav sidebar per iterative/dvc.org/pull/1110):

metrics: add, modify, show, diff, remove
pipeline: list, show
remote: add, default, modify, list, remove

I can probably send the PR too 🙂 lmk

### LINKED ISSUE #3625: dvc: meaningful option arguments
To match changes in documentation, the argument literals printed in -h help output of DVC commands should use th `metavar` param of `argparse`'s `add_argument()` to change the default argument name e.g.
```console
λ dvc add -h
usage: dvc add [-h] [-q | -v] [-R] [--no-commit] [-f <filename>]
               targets [targets ...]
```
instead of `[-f FILE]`

See the https://dvc.org/doc/command-reference to find all the updated  arg names in use in docs.

jorgeorpinel:
- Closes #3625
- Closes #3608


jorgeorpinel:
How come the test coverage is `Showing 43 files with coverage changes found.` when I only changes a single file? See https://codecov.io/gh/iterative/dvc/pull/3626/changes

shcheklein:
@jorgeorpinel I think codecov integration is not functional. Also, in this repo we I think the preferred way is to use your own fork for PRs :)

jorgeorpinel:
@efiop do you prefer a fork PR? I have no problem migrating it. Thanks

dvc/command/add.py
```diff
@@ -53,6 +53,7 @@ def add_parser(subparsers, parent_parser):
         "-f",
         "--file",
         help="Specify name of the DVC-file this command will generate.",
+        metavar="<filename>",
```
efiop, line 56: I'm not sure how this is better than `-f FILE`. `<>` look out of place here, and I don't really see the value, sorry.

efiop: But ok, if you want to, sure, lets do it this way. Please remove "Draft" flag.

shcheklein: I think, the value can be better seen here - https://dvc.org/doc/command-reference/run#run

check with what `dvc run` outputs.

jorgeorpinel: Ruslan, 2 benefits (besides matching the docs which we already changed 😅):

- The existing param names aren't descriptive. FILE isn't the best example of this but e.g. `-o OUTS` why repeat the option name when the param name can hint to what type of argument is accepted.
- We didn't like how upper case looked in the docs. Looks like we're yelling haha. We checked Git and noticed they use `<>` e.g. https://git-scm.com/docs/git-commit

dvc/command/metrics.py
```diff
@@ -194,13 +194,14 @@ def add_parser(subparsers, parent_parser):
         "-t",
         "--type",
         help=(
-            "Type of metrics (json/tsv/htsv/csv/hcsv). "
+            "Type of metrics (json/yaml). "
```
jorgeorpinel, line 197: Note I changed these too per #3409 and #3572.

PR #3632, opened 2020-04-13
remote: use .dir checksum existence to infer file contents existence

pmrowla:

`push`/`gc` behavior changes from #3600 

* `status`: (also affects `push`/`pull`) when checking .dir checksum exists on the remote, if the .dir checksum exists, presume that checksums for all files contained in that directory also exist on the remote
* `push`: push .dir checksums after file checksums. If any file in the directory fails to be uploaded, do not upload the .dir file
* `gc`: remove .dir checksums before file checksums

### DIFF
--- a/dvc/cache.py
+++ b/dvc/cache.py
@@ -71,9 +71,39 @@ class Cache(object):
     azure = _make_remote_property("azure")
 
 
+class NamedCacheItem(object):
+    def __init__(self):
+        self.names = set()
+        self.children = defaultdict(NamedCacheItem)
+
+    def __eq__(self, other):
+        return self.names == other.names and self.children == other.children
+
+    def child_keys(self):
+        for key, child in self.children.items():
+            yield key
+            yield from child.child_keys()
+
+    def child_names(self):
+        for key, child in self.children.items():
+            yield key, child.names
+            yield from child.child_names()
+
+    def add(self, checksum, item):
+        self.children[checksum].update(item)
+
+    def update(self, item, suffix=""):
+        if suffix:
+            self.names.update(n + suffix for n in item.names)
+        else:
+            self.names.update(item.names)
+        for checksum, item in item.children.items():
+            self.children[checksum].update(item)
+
+
 class NamedCache(object):
     def __init__(self):
-        self._items = defaultdict(lambda: defaultdict(set))
+        self._items = defaultdict(lambda: defaultdict(NamedCacheItem))
         self.external = defaultdict(set)
 
     @classmethod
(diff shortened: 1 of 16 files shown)


dvc/repo/fetch.py
```diff
@@ -82,18 +101,20 @@ def _fetch_external(self, repo_url, repo_rev, files, jobs):
             if is_dvc_repo:
                 repo.cache.local.cache_dir = self.cache.local.cache_dir
                 with repo.state:
-                    cache = NamedCache()
+                    used_cache = []
```
pmrowla, line 104: From the old PR:
> Hm, not sure I understand why this is needed.

`Output.get_used_cache()` now returns a tuple of `(used_dir_cache, used_file_cache)`, so we collect a list of these tuples instead of merging all of the used dir/file cache entries into a single `NamedCache`

dvc/output/base.py
```diff
@@ -429,16 +436,14 @@ def get_used_cache(self, **kwargs):
                     )
                 )
             logger.warning(msg)
-            return NamedCache()
+            return None, NamedCache()
```
efiop, line 439: Maybe it would make sense to push `dir` into NamedCache? So that NamedCache could contain both dirs and files(e.g. `cache["local"].dirs)? Just an idea, not sure if it would be nicer looking in the end.

pmrowla: NamedCache supports nesting cache items (for `.dir` checksums) after 02aaf42, the end result is a bit cleaner now

efiop:
@pmrowla Please check the tests, travis is failing.

efiop:
@pmrowla Please check the tests, travis is failing.

pmrowla:
@efiop test issue is resolved

efiop:
@pmrowla Check DS please too, e.g. this https://deepsource.io/gh/iterative/dvc/run/b2b2c666-0af4-4fb4-aed2-0210d2946dfc/python/PYL-R1704/ doesn't look good.

dvc/cache.py
```diff
@@ -71,9 +71,39 @@ def __init__(self, repo):
     azure = _make_remote_property("azure")
 
 
+class NamedCacheItem(object):
```
skshetry, line 74:
```suggestion
class NamedCacheItem:
```

dvc/cache.py
```diff
@@ -86,19 +116,51 @@ def __getitem__(self, key):
...
 
     def add(self, scheme, checksum, name):
-        self._items[scheme][checksum].add(name)
+        """Add a mapped name for the specified checksum."""
+        self._items[scheme][checksum].names.add(name)
+
+    def add_child_cache(self, checksum, cache, suffix=""):
+        """Add/update child cache for the specified checksum."""
+        for scheme, src in cache._items.items():
+            dst = self._items[scheme][checksum].children
```
skshetry, line 125: Do we need to check if `checksum` exists?

pmrowla: Each nested level is a defaultdict, so if `checksum` isn't already in our cache, it will be populated with the correct empty containers when we reference it here

skshetry: Oops, sorry. Forgot about that. :(

dvc/remote/local.py
```diff
@@ -255,37 +257,102 @@ def status(
...
+        dir_status, file_status, _ = self._status(
+            named_cache,
+            remote,
+            jobs=jobs,
+            show_checksums=show_checksums,
+            download=download,
+        )
+        return dict(dir_status, **file_status)
+
+    def _status(
```
shcheklein, line 270: a bit late to the party :) excellent PR overall. will mention a few things that I noticed reviewing that will might help later somehow.

this function definitely wants some split I would say... extract presentation logic out for example, and the that checks dir and excludes it? or may be some other refactoring ... but it's too long and complicated now

overall "internal-client" functions should be easy to read, even if it feels that extraction does not make much sense (e.g. helpers that are used only in a single place) think how will developers read this. It's easier to read story-like main function and go into details if needed

PR #3634, opened 2020-04-15
remote: locally index list of checksums available on cloud remotes

pmrowla:

Index changes from #3600.
Related to #2147.

Docs PR: iterative/dvc.org#1163

### New behavior

* Checksums available on the remote are indexed (cached) locally in a sqlite3 database. The db file will be stored as `.dvc/tmp/index/<filename>.idx` where `filename` is the SHA256 digest of the remote URL.
    * Only .dir checksums, and checksums for files contained inside those directories will be indexed
    * Standalone files will not be indexed
* On any dvc command which queries remote status (`status -c`/`push`/`pull`/`fetch`) index will be updated to include any new .dir's which are found on the remote (or are pushed successfully to the remote)

### Implementation details

* `remote.status/remote.cache_exists`
    1. Validate our index by querying for .dir checksums on the remote. If a .dir checksum is in our index, but not the remote (i.e. someone else has run `gc -c`), clear the entire index.
    2. If a .dir checksum exists on the remote, trust that directory file contents also exists on the remote (unchanged from current behavior)
    3. Update index with any new directories found on the remote (i.e. dirs that someone else has `push`'d since we last queried remote status)
    4. Query remote for standalone files (unchanged from current behavior)
* `pull`/`fetch`: If any index related mismatch/error occurs (i.e. trying to download a file that is in our index but not on the remote), clear the entire index.
* `push`: After push operation, successfully uploaded .dir checksums and file contents checksums for the directory contents are added to the index. Files uploaded during a partially failed push will not be indexed
* `gc -c`: Re-index will be required after a `gc -c` operation
* Changes to the index database are only committed once, after all operations on the remote have finished.

### Todo
* [x] use `.dir` checksum existence on remote to validate/invalidate index
* [x] unit tests for index module
* [x] functional tests for new behavior
* [x] docs
* [x] benchmarks
* [x] changes from [4/17 discussion](https://github.com/iterative/dvc/pull/3634#issuecomment-615196431)


dvc/remote/index.py
```diff
@@ -0,0 +1,228 @@
...
+        repo: repo for this remote index.
+        name: name for this index. Index db will be loaded from and saved to
+            ``.dvc/tmp/index/{name}.idx``.
+        dir_suffix: suffix used for naming directory checksums
+    """
+
+    INDEX_SUFFIX = ".idx"
+    VERSION = 1
+    INDEX_TABLE = "remote_index"
+    INDEX_TABLE_LAYOUT = "checksum TEXT PRIMARY KEY, " "dir INTEGER NOT NULL"
```
pmrowla, line 64: f we are using a database, would it be better to have a single .dvc/tmp/index db, with appropriate tables/columns for handling multiple remotes, vs having multiple single-table .dvc/tmp/index/... dbs?

I'm not sure how common it is for users to have multiple remotes configured, but in the case where you have multiple large remotes, as it is right now we would be duplicating data across multiple (potentially large) index db files.

dvc/remote/local.py
```diff
@@ -354,6 +375,29 @@ def make_names(checksum, names):
...
+                "clearing remote index".format(", ".join(missing_dirs))
+            )
+            remote.index.clear()
+        else:
+            # If .dir checksum exists on the remote, assume indexed
+            # directory contents still exists on the remote
+            for dir_checksum in dir_exists:
+                yield from remote.index.intersection(
+                    named_cache.child_keys(self.scheme, dir_checksum)
+                )
```
pmrowla, line 396: Safer to assume that only indexed dir contents exists here, rather than assuming that the full dir contents exists.

pmrowla:
Some benchmarks from my machine (dual-core i7-5557U@3.10GHz, osx catalina):

dataset-registry-private (2M imagenet files) w/empty local cache (no pulled/fetched data)

latest, no index
```
time dvc status -c
865.30s user 102.08s system 97% cpu 16:29.62 total
```

latest, with index
```
time dvc status -c
497.33s user 89.30s system 92% cpu 10:34.19 total
```

0.91.0 default (pre-optimizations)
```
```

0.91.0 no-traverse = False
```
time dvc status -c
802.44s user 87.22s system 42% cpu 35:04.45 total
```

rclone
```
time rclone sync --dry-run dvc-private:dvc-private/remote/dataset-registry-private/ .dvc/cache/
141.96s user 21.53s system 64% cpu 4:11.82 total
```

The traverse improvements from 0.91.0 to now should be more drastic on machines with more cores.
Also, it would be ideal if we had a convenient way to only benchmark the cloud/remote steps to see how we compare to tools like rclone

efiop:
@pmrowla And what about a case when we add 1 new file to the dataset? I think it should also shine there, compared to what we had before index.

dvc/remote/index.py
efiop: @pmrowla If we are talking about storing unpacked *.dir cache files, then there doesn't seem to be a point to store them for each remote, since no matter the remote if it has *.dir present we trust that it also has all cache files from within it. So maybe it is a premature optimization for later general indexes for whole remotes.

pmrowla:
Had discussion with @efiop, a couple of changes needed before this can be merged.

1. Keep behavior from current master regarding trusting .dir files on the remote
2. For push, only update index when entire dir has been uploaded successfully

pmrowla:
Previous benchmarks post has been updated with current master and PR branch numbers for `status -c`. Whenever there is no pre-existing index db (or when `--clear-index` is used to force reindexing), adding the initial 2M checksums to the db does add ~2 minutes of overhead, but after the index exists, the overhead for checking the existing index is negligible.

Running with cProfile, pathinfo handling code is a large bottleneck for us when checking cloud status, particularly in `dvc/stage/base.py::_collect_used_dir_cache()`

from `python -m cProfile -o status.prof -m dvc status -c` with an existing index db:

Screen Shot 2020-04-20 at 3 07 59 PM

Screen Shot 2020-04-20 at 3 18 02 PM

Screen Shot 2020-04-20 at 3 16 06 PM

zip w/cprofile output:
[cprofile.zip](https://github.com/iterative/dvc/files/4501522/cprofile.zip)

pmrowla:
Benchmarks for `status -c` after adding 1 file to an existing pushed (and indexed) dir with 100k files.

S3 remote, 100k total files in remote cache

latest PR branch
```
time dvc status -c
13.58s user 5.91s system 82% cpu 23.671 total
```

latest master
```
time dvc status -c
41.61s user 8.32s system 84% cpu 59.136 total
```

dvc/remote/base.py
```diff
@@ -872,10 +899,18 @@ def cache_exists(self, checksums, jobs=None, name=None):
         # cache_exists() (see ssh, local)
         assert self.TRAVERSE_PREFIX_LEN >= 2
 
-        if len(checksums) == 1 or not self.CAN_TRAVERSE:
-            return self._cache_object_exists(checksums, jobs, name)
+        checksums = set(checksums)
+        indexed_checksums = set(self.index.intersection(checksums))
+        checksums -= indexed_checksums
+        logger.debug(
+            "Matched {} indexed checksums".format(len(indexed_checksums))
```
efiop, line 906: minor: please try to wrap dynamic part of the message in single-quotes, it makes it much easier to visually parse. And to be consistent too :slightly_smiling_face: 
```suggestion
            "Matched '{}' indexed checksums".format(len(indexed_checksums))
```

shcheklein:
@pmrowla amazing stuff 👏 - ~7 times faster than rclone. 

I'm curious where do we spend 13seconds though in this scenario 🤔

shcheklein:
@pmrowla is it ` pathinfo handling code is a large bottleneck for us when checking cloud status, particularly in dvc/stage/base.py::_collect_used_dir_cache()`?

shcheklein:
Also, out of curiosity - do we actually need `--drop-index` option? A bit worried that we provide it to every data management related command.

dvc/command/data_sync.py
```diff
@@ -162,6 +164,12 @@ def add_parser(subparsers, _parent_parser):
         default=False,
         help="Pull cache for subdirectories of the specified directory.",
     )
+    pull_parser.add_argument(
+        "--clear-index",
+        action="store_true",
+        default=False,
+        help="Clear local index for the specified remote.",
```
efiop, line 171: Desc doesn't match the other ones. Doesn't look like it was intended.

efiop: Also, if we add it for `pull`, i guess we also have to add it for `fetch`. Hm...

efiop:
> Also, out of curiosity - do we actually need --drop-index option? A bit worried that we provide it to every data management related command.

Thinking about it, looks like current --clear-index simply drops the index for specific remote. So maybe it would be more logical do create `dvc remote clear-index`(maybe there is a better name) instead. There doesn't seem to be a good use case from the user point of view, but it is really helpful for us when debugging or providing the support. That being said, we could always ask user to `rm -rf .dvc/tmp/index` instead. So looks like it is indeed not strictly necessary but handy.

pmrowla:
> @pmrowla amazing stuff 👏 - ~7 times faster than rclone.
> 
> I'm curious where do we spend 13seconds though in this scenario 🤔

@shcheklein in the "add 1 new file" scenario, running with cProfile shows that ~1/3 of the execution time is spent in the same pathinfo handling code that causes the bottleneck in the 2M file example.

There is no clear bottleneck for the remaining time, which is split between checking whether or not files in the local cache have been modified, and checking if the previously indexed .dir checksum and newly added checksums exist on the remote.

Regarding the `--clear-index` option, from a user standpoint I think it can be removed. Originally it was intended to provide a bit more coverage for the "team member has run `gc -c`" scenario, but since we are always verifying that indexed .dir checksums still exist on the remote, that is no longer necessary. As @efiop it is useful for us with regard to debugging, but deleting the index dir accomplishes the same thing.

pmrowla:
`--clear-index` has been removed and the associated docs PR is closed

dvc/remote/base.py
```diff
@@ -70,11 +72,24 @@ def __init__(self, checksum):
         )
 
 
+def index_locked(f):
```
efiop, line 75: Do we really need this lock, btw? index is per-repo, so as long as repo is locked - index is locked too. And I don't see us unlocking `.dvc/lock` for status/gc/pull/pull (at least right now).

dvc/remote/index.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def replace(self, dir_checksums, file_checksums):
+        """Replace the contents of this index with the specified checksums.
+
+        Changes to the index will not committed until dump() is called.
+        """
+        self.clear()
+        self.update(dir_checksums, file_checksums)
+
+    def replace_all(self, checksums):
```
efiop, line 207: Might be missing something, but I don't  see it being used at all.

dvc/remote/base.py
pmrowla: The sqlite3 python lib isn't thread safe, or at least it's not safe to share cursor objects across threads, so it's more to make sure the index is only being accessed from the main thread

pmrowla: In terms of what's in this PR the lock isn't needed, but since there are parts of `cache_exists` which are multithreaded, it seems safer to prevent anyone from accidentally misusing the `remote.index` object if/when `status` or `cache_exists` behavior is updated in the future

efiop: @pmrowla Got it, makes sense to me. Thanks!

Issue #3635, opened 2020-04-15
remote: loading dirs is exteremely slow because of PathInfo


efiop:
This https://github.com/iterative/dvc/blob/0.93.0/dvc/remote/base.py#L290 takes ~50 seconds compared to ~0.4 for simple `replace()` that we used to have on our imagenet dataset.

efiop:
Actually, same with dumping. PathInfo is extremely slow when we need lots of it.

efiop:
Another confirmation https://github.com/iterative/dvc/pull/3634#issuecomment-616335337

pmrowla:
The issue is that in `OutputBase._collect_used_dir_cache()`, we add `str(path_info)` to our used_cache dict, rather than just adding `path_info`.

https://github.com/iterative/dvc/blob/52369bdf47d2fcafed8ca3d1f9b0deaecc99a236/dvc/output/base.py#L393-L397

`PathInfo.__str__()` is overridden to always compute the relative path from our current working dir, rather than than returning a string fspath.

https://github.com/iterative/dvc/blob/52369bdf47d2fcafed8ca3d1f9b0deaecc99a236/dvc/path_info.py#L45-L47

For *nix systems, `relpath()` just wraps `os.path.relpath`, but in the event that we have a directory with millions of files, we will end up making millions of `os.path.relpath`/`os.path.abspath`/`os.getcwd` calls here, which is very slow (see cpython [posixpath.py](https://github.com/python/cpython/blob/9e51aab37e9af6fa0fe406fd56184a0aded28e11/Lib/posixpath.py#L449)).

pmrowla:
Since we don't actually need the relative path until we are printing/logging status for modified files, one easy optimization is that we can just put `path_info` into `used_cache`, and delay calling `str(path_info)` until we actually want the relative path string. For the case where we only have a few modified files, we will only compute `relpath` for those modified files rather than for the entire directory.

But this does not cover the case where we have to potentially print/log millions of files. To get better behavior in that case, we can implement our own versions of `posixpath.abspath`/`posixpath.relpath` that would take advantage of the fact that things like `os.getcwd()` should not change between `abspath` and `relpath` calls for our use case.

efiop:
@pmrowla Very interesting idea about relpath/abspath re-implementation! But I'm a bit unsure if it is really worth creating such a general hack for it, when we could do a local hack something like so:
```
        path = str(self.path_info)
        filter_path = str(filter_info) if filter_info else None
[...]

PR #3637, opened 2020-04-15
remote: don't use PathInfo.from_posix

efiop:
Using it to load dirs takes around 50 sec compared to 0.4 with simple
`replace()`. Path objects are great, but not when you need lots of them.

Related to #3635

### DIFF
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -21,7 +21,7 @@ from dvc.exceptions import (
     RemoteCacheRequiredError,
 )
 from dvc.ignore import DvcIgnore
-from dvc.path_info import PathInfo, URLInfo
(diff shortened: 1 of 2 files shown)


dvc/remote/base.py
```diff
@@ -285,10 +286,13 @@ def load_dir_cache(self, checksum):
             )
             return []
 
-        for info in d:
-            # NOTE: here is a BUG, see comment to .as_posix() below
-            relative_path = PathInfo.from_posix(info[self.PARAM_RELPATH])
-            info[self.PARAM_RELPATH] = relative_path.fspath
+        if os.name == "nt":
+            # only need to convert it for Windows
```
efiop, line 290: Actually this is wrong, this code is not only for local paths. Looking into it...

Issue #3644, opened 2020-04-17
repro -P seems to check status of dependancies too often


charlesbaynham:
DVC 0.91.0, Windows 10, pip

Description
========

Running `dvc repro -P` on a large graph seems to check all the early dependences once for each output, even if it has already been checked. 

In the following repo:

                            step1.dvc
                              |
             ----------------------------------------
             |            |            |            |           
     step2_A.dvc  step2_B.dvc  step2_C.dvc  step2_D.dvc

Where each of the step 2 stages depends on steps 1, I think that calling `dvc repro -P` results in 4x checks of the outputs of step1. 

I conclude this because, in my repo, I have several hundred stages in "step2" which all depend on a "step1" which is an import. When I run `dvc repro -P`, I get the

    WARNING: DVC-file 'xxx.dvc' is locked. Its dependencies are not going to be reproduced.

message several hundred times.

charlesbaynham:
In Discord, @efiop suggested problem might be in 
https://github.com/iterative/dvc/blob/11e56dbdf921c6eacf603bcdcc2e7a46ac7a3b88/dvc/repo/reproduce.py#L78-L90

efiop:
Fixed by https://github.com/iterative/dvc/pull/3645 . Kudos @charlesbaynham :1st_place_medal:

PR #3645, opened 2020-04-17
Fix repro duplicating effort while checking steps #3644

charlesbaynham:


charlesbaynham:
To close #3644

efiop:
Hi @charlesbaynham ! Thanks a lot for the PR! Please make sure to install our pre-commit hooks as per contribution guide. They will check for styling and some lint issues.

charlesbaynham:
> 
> 
> Hi @charlesbaynham ! Thanks a lot for the PR! Please make sure to install our pre-commit hooks as per contribution guide. They will check for styling and some lint issues.

Don't worry, I will before I resolve the WIP!

charlesbaynham:
Ok dokey, that's ready for review @efiop

dvc/repo/reproduce.py
```diff
@@ -95,16 +95,21 @@ def reproduce(
         )
 
     ret = []
+    checked_stages = set()
     for target in targets:
-        stages = _reproduce_stages(active_graph, target, **kwargs)
+        stages, these_checked_stages = _reproduce_stages(
```
efiop, line 100: What if we simply pass all targets to `_reproduce_stages` and do the duplication something like
```
pipeline = stages                                                               
if not single_item:                                                             
    for stage in stages:                                                        
        if downstream:                                                                    
            nodes = nx.dfs_preorder_nodes(G.copy().reverse(copy=False), stage)  
        else:                                                                   
            nodes = nx.dfs_postorder_nodes(G, stage)                            
        new = [node for node in nodes if node not in pipeline]                  
        pipeline.extend(new)                                                    
```

efiop:
So sorry for the delay! Please see a comment above

dvc/repo/reproduce.py
charlesbaynham: Yep, It can definitely be de-duped in `_reproduce_stages`, I'll change that. I think there's a bug in the code above: the repro `target` stages will always be executed first, before their dependencies. I.e. if I specify D in 

                               E
                              / \
                             D   F
                            / \   \
                           B   C   G
                            \ /
                             A

Then my execution order would be [D, A, B, C]

charlesbaynham: By the way, is it a design decision to avoid type hints? I'd usually use them but I can't see them elsewhere in the codebase

efiop: > By the way, is it a design decision to avoid type hints? I'd usually use them but I can't see them elsewhere in the codebase

Not at all, we've just dropped py2 support in Jan and simply didn't use type hints widely yet. We do have some in stage-related code, IIRC. Please feel free to use them :slightly_smiling_face:

efiop: > Then my execution order would be [D, A, B, C]

Good catch! Sorry, that was my smooth brain moment :slightly_smiling_face:

PR #3647, opened 2020-04-18
remote: add support for WebDAV

### LINKED ISSUE #1153: WebDav: new remote type
WebDav is a quite common protocol for file sharing.

It is especially convenient because box.com provides 10Gb free and supports WebDav.

shizacat:
Fixes #1153

### DIFF
--- a/dvc/path_info.py
+++ b/dvc/path_info.py
@@ -104,7 +104,14 @@ class _URLPathParents(object):
 
 
 class URLInfo(_BasePath):
-    DEFAULT_PORTS = {"http": 80, "https": 443, "ssh": 22, "hdfs": 0}
+    DEFAULT_PORTS = {
+        "http": 80,
+        "https": 443,
+        "ssh": 22,
+        "hdfs": 0,
+        "webdav": 80,
+        "webdavs": 443,
+    }
 
     def __init__(self, url):
         p = urlparse(url)
--- a/dvc/path_info.py
+++ b/dvc/path_info.py
@@ -335,17 +335,3 @@ class WebdavURLInfo(HTTPURLInfo):
             ("?" + self.query) if self.query else "",
             ("#" + self.fragment) if self.fragment else "",
         )
-
-    def get_collections(self) -> list:
-        def pcol(path):
-            return "{}://{}{}".format(
-                self.scheme.replace("webdav", "http"), self.netloc, path,
-            )
-
-        p = self.path.split("/")[1:-1]
-        if not p:
-            return []
-        r = []
-        for i in range(len(p)):
-            r.append(pcol("/{}/".format("/".join(p[: i + 1]))))
-        return r
--- a/dvc/path_info.py
+++ b/dvc/path_info.py
@@ -312,3 +312,33 @@ class HTTPURLInfo(URLInfo):
             and self._path == other._path
             and self._extra_parts == other._extra_parts
         )
+
+
+class WebdavURLInfo(HTTPURLInfo):
+    def __init__(self, url):
+        super().__init__(url)
+
+    @cached_property
+    def url(self):
+        return "{}://{}{}{}{}{}".format(
+            self.scheme.replace("webdav", "http"),
+            self.netloc,
+            self._spath,
+            (";" + self.params) if self.params else "",
+            ("?" + self.query) if self.query else "",
+            ("#" + self.fragment) if self.fragment else "",
+        )
+
+    def get_collections(self) -> list:
+        def pcol(path):
+            return "{}://{}{}".format(
+                self.scheme.replace("webdav", "http"), self.netloc, path,
+            )
+
+        p = self.path.split("/")[1:-1]
+        if not p:
+            return []
+        r = []
+        for i in range(len(p)):
+            r.append(pcol("/{}/".format("/".join(p[: i + 1]))))
+        return r
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -552,8 +552,8 @@ class RemoteBASE(object):
         if not hasattr(self, "_upload"):
             raise RemoteActionNotImplemented("upload", self.scheme)
 
-        if to_info.scheme != self.scheme:
-            raise NotImplementedError
+        # if to_info.scheme != self.scheme:
+        #     raise NotImplementedError
 
         if from_info.scheme != "local":
             raise NotImplementedError
@@ -588,8 +588,8 @@ class RemoteBASE(object):
         if not hasattr(self, "_download"):
             raise RemoteActionNotImplemented("download", self.scheme)
 
-        if from_info.scheme != self.scheme:
-            raise NotImplementedError
+        # if from_info.scheme != self.scheme:
+        #     raise NotImplementedError
 
         if to_info.scheme == self.scheme != "local":
             self.copy(from_info, to_info)
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -552,8 +552,8 @@ class RemoteBASE(object):
         if not hasattr(self, "_upload"):
             raise RemoteActionNotImplemented("upload", self.scheme)
 
-        # if to_info.scheme != self.scheme:
-        #     raise NotImplementedError
+        if to_info.scheme != self.scheme:
+            raise NotImplementedError
 
         if from_info.scheme != "local":
             raise NotImplementedError
@@ -588,8 +588,8 @@ class RemoteBASE(object):
         if not hasattr(self, "_download"):
             raise RemoteActionNotImplemented("download", self.scheme)
 
-        # if from_info.scheme != self.scheme:
-        #     raise NotImplementedError
+        if from_info.scheme != self.scheme:
+            raise NotImplementedError
 
         if to_info.scheme == self.scheme != "local":
             self.copy(from_info, to_info)
--- a/dvc/remote/webdav.py
+++ b/dvc/remote/webdav.py
@@ -10,6 +10,7 @@ from dvc.path_info import WebdavURLInfo
 class RemoteWEBDAV(RemoteHTTP):
     scheme = Schemes.WEBDAV
     path_cls = WebdavURLInfo
+    REQUEST_TIMEOUT = 20
 
     def _upload(self, from_file, to_info, name=None, no_progress_bar=False):
         def chunks():
@@ -45,6 +46,8 @@ class RemoteWEBDAV(RemoteHTTP):
         for idx in range(from_idx, len(url_cols)):
             response = self._request("MKCOL", url_cols[idx])
             if response.status_code not in (200, 201):
+                if bool(self._request("HEAD", url_cols[idx])):
+                    continue
                 raise HTTPError(response.status_code, response.reason)
 
     def gc(self):
--- a/dvc/remote/webdav.py
+++ b/dvc/remote/webdav.py
@@ -1,7 +1,5 @@
 import os.path
 
-from funcy import cached_property
-
 from .http import RemoteHTTP
 from dvc.scheme import Schemes
 from dvc.progress import Tqdm
--- a/dvc/remote/webdav.py
+++ b/dvc/remote/webdav.py
@@ -50,8 +50,13 @@ class RemoteWEBDAV(RemoteHTTP):
                     continue
                 raise HTTPError(response.status_code, response.reason)
 
+    def remove(self, path_info):
+        response = self._request("DELETE", path_info.url)
+        if response.status_code not in (200, 201, 204):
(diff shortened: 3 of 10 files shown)


dvc/remote/webdav.py
```diff
@@ -0,0 +1,54 @@
...
+                    desc=to_info.url if name is None else name,
+                    disable=no_progress_bar,
+                ) as fd_wrapped:
+                    while True:
+                        chunk = fd_wrapped.read(self.CHUNK_SIZE)
+                        if not chunk:
+                            break
+                        yield chunk
+
+        response = self._request("PUT", to_info.url, data=chunks())
```
skshetry, line 52: I suppose you need an `MKCOL` here, right? AFAIK, webdav does not allow urls like this (eg: `/folder/file` without creating folder first. Correct me if I'm wrong.

skshetry, line 52: I suppose you need an `MKCOL` here, right? AFAIK, webdav does not allow urls like this (eg: `/folder/file` without creating folder first. Correct me if I'm wrong.

dvc/remote/base.py
```diff
@@ -552,8 +552,8 @@ def upload(self, from_info, to_info, name=None, no_progress_bar=False):
         if not hasattr(self, "_upload"):
             raise RemoteActionNotImplemented("upload", self.scheme)
 
-        if to_info.scheme != self.scheme:
-            raise NotImplementedError
+        # if to_info.scheme != self.scheme:
+        #     raise NotImplementedError
```
efiop, line 556: I suppose this PR is WIP? In that case, please add `[WIP]` prefix :slightly_smiling_face: These lines shouldn't be removed in the final PR, unless there is a really good reason for it.

dvc/remote/webdav.py
```diff
@@ -0,0 +1,54 @@
...
+class RemoteWEBDAV(RemoteHTTP):
+    scheme = Schemes.WEBDAV
+
+    def __init__(self, repo, config):
+        super().__init__(repo, config)
+
+        url = config.get("url")
+        if url:
+            self.path_info = self.path_cls(url)
+            self.path_info.scheme = self.path_info.scheme.replace(
```
efiop, line 19: So far looks like this simply copies HTTP implementation. Are there any differences?

dvc/remote/webdav.py
shizacat: Hi, I have tested on nginx with option ("_create_full_put_path  on;_") and work fine. But you right. Add calls to create directories.

dvc/remote/webdav.py
shizacat: Here I added only to change scheme. I was not sure if there would be problems in the future if I did it after. 
May be:
```

        if self.path_info is not None:
            self.path_info.scheme = self.path_info.scheme.replace(
                "webdav", "http")
```

dvc/remote/base.py
shizacat: If I understand correctly, to determine which 'remote' to use - the scheme is used. For this, I added the 'webdav' scheme, but it works on http. But the base is checking for compliance with the scheme - and removed it.

dvc/remote/webdav.py
efiop: @shizacat How does a typical webdav URL look like? Does it use `webdav` as a schema? If there are no differences, maybe let's just document that if one wants to use webdav he could access it as a HTTP(s) remote?

shizacat: He does not have his own separate scheme. This is a normal HTTP request.
It differs from the remote.http implementation in the following:
- To upload, you need another method (PUT, instead of POST). As a solution - add an option when configuring remote.http
- The creation of interstitial directories on the server may be required. Depends on the server settings, may not be critical if specified as a requirement. The same nginx allows this.

efiop: Is there any particular way of detecting if PUT or POST is needed? Maybe some specific error is raised when you use one or another and you could fallback to another one?

I'm not familiar with webdav, so I'm just wondering if we should make it work as a generic HTTP(s) or if we should create a special remote for it. So far not 100% clear to me.

shizacat: There is no 100% way to determine the webdav server. There is a request through the OPTIONS method, but is not specified in the standard. I think it's better to separate them.
In recent commits, I made a separate _WebdavURLInfo_ object, and it already returns the correct url, there is no need to edit _base.py_.

efiop: @shizacat Got it, thanks for clarifying! :pray: Looks better now. Btw, what about tests? Are there any options?

shizacat:
@efiop Add tests

tests/unit/test_path_info.py
```diff
@@ -89,3 +90,35 @@ def test_https_url_info_str():
 def test_path_info_as_posix(mocker, path, as_posix, osname):
     mocker.patch("os.name", osname)
     assert PathInfo(path).as_posix() == as_posix
+
+
+@pytest.mark.parametrize("cls", [WebdavURLInfo])
```
efiop, line 95: Looks like we don't need parametrization here and below.

shizacat: Fixed

efiop:
@shizacat Btw, looks like your git email is not added to your github email list. Please go to https://github.com/settings/emails and make sure that your git email (you can see it in git log, not posting it myself so it is not picked up by bots) is added there. That way all of your contributions will be taken into account by github slightly_smiling_face

dvc/path_info.py
```diff
@@ -312,3 +312,33 @@ def __eq__(self, other):
...
+        return "{}://{}{}{}{}{}".format(
+            self.scheme.replace("webdav", "http"),
+            self.netloc,
+            self._spath,
+            (";" + self.params) if self.params else "",
+            ("?" + self.query) if self.query else "",
+            ("#" + self.fragment) if self.fragment else "",
+        )
+
+    def get_collections(self) -> list:
```
efiop, line 332: Hm, I wonder how is this different from `parents`?

efiop:
@shizacat Could you also submit docs for this change, please? (see https://dvc.org/doc/command-reference/remote for `remote add/modify`). We are not very familiar with WebDav ourselves, so it would be rather hard for us to come up with the docs without your help. Surely we will help out with it too. I guess one of the interesting questions is does this work with box.com and maybe some other popular services that support Webdav? If so, which ones?

shizacat:
@skshetry You are right. I forgot to add in commit.

shizacat:
@shizacat Checked with yandex.ru (webdav), work, but due to threads, paths can be created in parallel. Added fix for this. About docs, I will look

dvc/path_info.py
```diff
@@ -312,3 +312,33 @@ def __eq__(self, other):
...
+
+
+class WebdavURLInfo(HTTPURLInfo):
+    def __init__(self, url):
+        super().__init__(url)
+
+    @cached_property
+    def url(self):
+        return "{}://{}{}{}{}{}".format(
+            self.scheme.replace("webdav", "http"),
```
shcheklein, line 324: what about https? webdavs?

dvc/remote/webdavs.py
```diff
@@ -0,0 +1,15 @@
+from .webdav import RemoteWEBDAV
+from dvc.scheme import Schemes
+
+
+class RemoteWEBDAVS(RemoteWEBDAV):
+    scheme = Schemes.WEBDAVS
+
+    def gc(self):
+        raise NotImplementedError
```
shcheklein, line 9: do we need to repeat these stubs?

pmrowla: `gc` needs to be explicitly disabled for HTTP(S) remotes since there is no way to implement `remote.remove()` for HTTP. But WebDAV does have a `DELETE` method, so should we implement `remote.remove()` for WebDAV (and by extension support `gc`)?

dvc/remote/webdavs.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def list_cache_paths(self, prefix=None, progress_callback=None):
```
pmrowla, line 11: This can be implemented using the WebDAV `PROPFIND` method, but may not be worth it for now unless we want to support `gc` for WebDAV remotes.

dvc/remote/webdavs.py
skshetry: @pmrowla, this works for most of the usecases already. So, I think, it's better to just not implement till any user asks for it.

dvc/path_info.py
shizacat: If replace in "webdav**S**" webdav on http will be "http**S**"

shizacat:
Docs - https://github.com/iterative/dvc.org/pull/1187

dvc/remote/webdavs.py
shizacat: @shcheklein This code is repeating because '**DeepSource**' swears.

shcheklein: 👍 right. Resolving this.

dvc/path_info.py
shcheklein: yep, 🤦 sorry :)

Ok, a few concerns still:

1. Do we need to adjust DEFAULT_PORTS in the URLInfo base class?
2. Redefining `url` like this means loosing information. E.g. when we write in logs we will be getting http instead. In general feels like not a good idea creating a discrepancy between what we have inside class and what we return here.

Feels like we should be doing it on the Remote class level, or there should be some separate method - access url or something?

shizacat: 1. Fixed

It seems to me to do a separate method is over head. It will be necessary to rewrite the entire code from _HTTP_ with one fix, and at the same time, there will still be the possibility use in logs the wrong _URL_.

shcheklein: Yes, it was merely a random suggestion. The whole approach with overriding `url` feels like a hack and can hit us somewhere down the road. There should be a better way of doing this - e.g. RemoteWebDav class should understand how to deal with those URLs (translate them if needed and pass to HTTPRemote), etc.

dvc/path_info.py
efiop: @shizacat ^ ?

dvc/path_info.py
efiop: Actually, makes me think if we should tackle it the other way around by simply adding some flag that will tell HTTP(s) remote to use webdav to access this remote that would otherwise have a normal HTTP(s) URL. Something like
```
dvc remote add myremote https://example.com/path
```
and then
```
dvc remote modify myremote webdav true
```
or
```
dvc remote modify myremote type webdav
```
or something else. Not sure if this makes sense. It kinda goes against current conventions, but also I'm not sure if webdav is unique enough for a separate remote, maybe it is better to handle it as HTTP with some special tweaks. :thinking:

dvc/path_info.py
shizacat: This me need only one thing - replace scheme

dvc/path_info.py
shizacat: Hack - not sure. The same **DEFAULT_PORTS** also change the state of the object without notice. The same S3 is HTTP.
Leaving url unchanged - then the purpose of the **_BasePath** class is not very clear. It seemed to me that he should return the ready-made address format for working with him.

As for - as an extension of http. The more I work with him, the more he is everyday. The same gc. I don’t know how it works, but if only the remote method is needed for it, it’s easy to add and now a big difference from HTTP is already obtained.

shcheklein: >  The same DEFAULT_PORTS also change the state of the object without notice

Default ports logic is part of the protocols ... e.g. we want to detect that `example.com:80` is the same as `example.com`. In your case you create an object that provides an inconsistent interface - some methods (and some logic internally) might still rely on `webdav`, while url has `http`. It can be very confusing - you ask and object to print its name - it prints Duck but then behave like Dog.

> The same S3 is HTTP

not sure I got this point, could you clarify, please?

> As for - as an extension of http. The more I work with him, the more he is everyday. The same gc. I don’t know how it works, but if only the remote method is needed for it, it’s easy to add and now a big difference from HTTP is already obtained.

Again, not sure I understood this. Btw, we can jump on Discord - dvc.org/chat and discuss this with me and Ruslan (ivan and ruslan there). Please, feel free to ping me/us.

efiop:
For the record: discussed privately that we should look around to see if there are any good libraries to handle webdav for us.

efiop:
Hi @shizacat ! Need any help? :slightly_smiling_face:

shizacat:
@efiop Sorry, May holidays ...

shizacat:
It need implemented through a separate library (for work with webdav). I will try this to do in the new PR.

PR #3649, opened 2020-04-18
tests: test_directories

nik123:
Fixed test_directories test in tests/func/test_diff.py


efiop:
Thank you @nik123 ! :pray:

PR #3650, opened 2020-04-19
metrics diff error message in --no-scm projects

### LINKED ISSUE #3605: metrics diff: cryptic error in --no-scm projects [qa]
```console
λ dvc version
DVC version: 0.92.0
Python version: 3.7.5
Binary: True
Package: exe

λ dvc metrics diff
ERROR: unexpected error - 'NoSCM' object has no attribute 'resolve_rev'
```

Run in a --no-scm DVC project. Compare to `dvc diff`:

```console
λ dvc diff
ERROR: failed to get diff - only supported for Git repositories
```

karajan1001:

I am not sure if there is any other scm tools besides git going to be supported. 
If so, the error message"only supported for Git repositories" is not proper one,
 

fix #3605

### DIFF
--- a/dvc/repo/brancher.py
+++ b/dvc/repo/brancher.py
@@ -1,7 +1,6 @@
 from funcy import group_by
 
 from dvc.scm.tree import WorkingTree
-from dvc.scm.base import NOSCMError
 
 
 def brancher(  # noqa: E302
@@ -47,7 +46,5 @@ def brancher(  # noqa: E302
             for sha, names in group_by(scm.resolve_rev, revs).items():
                 self.tree = scm.get_tree(sha)
                 yield ", ".join(names)
-    except AttributeError:
-        raise NOSCMError()
     finally:
         self.tree = saved_tree
--- a/dvc/repo/brancher.py
+++ b/dvc/repo/brancher.py
@@ -1,7 +1,7 @@
 from funcy import group_by
 
 from dvc.scm.tree import WorkingTree
-from dvc.exceptions import DvcException
+from dvc.scm.base import NOSCMError
 
 
 def brancher(  # noqa: E302
@@ -48,6 +48,6 @@ def brancher(  # noqa: E302
                 self.tree = scm.get_tree(sha)
                 yield ", ".join(names)
     except AttributeError:
-        raise DvcException("only supported for Git repositories")
+        raise NOSCMError()
     finally:
         self.tree = saved_tree
--- a/dvc/repo/brancher.py
+++ b/dvc/repo/brancher.py
@@ -1,6 +1,7 @@
 from funcy import group_by
 
 from dvc.scm.tree import WorkingTree
+from dvc.exceptions import DvcException
 
 
 def brancher(  # noqa: E302
@@ -46,5 +47,7 @@ def brancher(  # noqa: E302
             for sha, names in group_by(scm.resolve_rev, revs).items():
                 self.tree = scm.get_tree(sha)
                 yield ", ".join(names)
+    except AttributeError:
+        raise DvcException("only supported for Git repositories")
     finally:
         self.tree = saved_tree
--- a/dvc/scm/__init__.py
+++ b/dvc/scm/__init__.py
@@ -1,7 +1,6 @@
 """Manages source control systems (e.g. Git)."""
 
-from dvc.scm.base import Base
-from dvc.scm.base import NoSCMError
+from dvc.scm.base import Base, NoSCMError
 from dvc.scm.git import Git
 
 
--- a/dvc/scm/__init__.py
+++ b/dvc/scm/__init__.py
@@ -1,13 +1,15 @@
 """Manages source control systems (e.g. Git)."""
 
 from dvc.scm.base import Base
+from dvc.scm.base import NoSCMError
 from dvc.scm.git import Git
 
 
 # Syntactic sugar to signal that this is an actual implementation for a DVC
 # project under no SCM control.
 class NoSCM(Base):
-    pass
+    def __getattr__(self, name):
+        raise NoSCMError
 
 
 def SCM(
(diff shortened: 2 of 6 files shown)


karajan1001:
Updated, I found `dvc metrics dfif` raise this exception after I initialized a git repository.
This was because the dvc configuration do not updated.
And only with cmd `dvc config core.no_scm False` , it finally be solved.

dvc/scm/base.py
```diff
@@ -24,6 +24,16 @@ class RevError(SCMError):
     pass
 
 
+class NOSCMError(SCMError):
+    def __init__(self):
+        msg = (
+            "only supported for Git repositories. If you had\n already "
+            "initialized a git repo, this may be caused by dvc configuration. "
+            "Use\n `dvc config core.no_scm False` to update it."
```
jorgeorpinel, line 32: Shouldn't this begin with capital O? Depends if it's printed by itself (in that case yes) or if it's a continuation of a previously printed message (in that case no). Other capitalizations:
```suggestion
            "only supported for Git repositories. If you had\n already "
            "initialized a Git repo, this may be caused by DVC configuration. "
            "Use\n `dvc config core.no_scm False` to update it."
```

jorgeorpinel: > If you had already initialized a Git repo, this may be caused by DVC configuration...

I'm not sure I understand this explanation though. I think we mean something like ``If you're seeing this error in a Git repo, try updating the DVC configuration with `dvc config core.no_scm False`.`` ?

karajan1001: I think so, my pool english representation.

> > If you had already initialized a Git repo, this may be caused by DVC configuration...
> 
> I'm not sure I understand this explanation though. I think we mean something like `` If you're seeing this error in a Git repo, try updating the DVC configuration with `dvc config core.no_scm False`. `` ?

jorgeorpinel: Ah sorry I didn't realize you were an outside contributor! Don't worry about it we can polish it 🙂 this is a great change. Thanks a lot!

karajan1001: Thank you too.

efiop:
@karajan1001 Btw, looks like your git email is not added to your github email list. Please go to https://github.com/settings/emails and make sure that your git email (you can see it in `git log`, not posting it myself so it is not picked up by bots) is added there. That way all of your contributions will be taken into account by github :slightly_smiling_face:

dvc/scm/base.py
```diff
@@ -24,6 +24,16 @@ class RevError(SCMError):
     pass
 
 
+class NOSCMError(SCMError):
+    def __init__(self):
+        msg = (
+            "Only supported for Git repositories. If you're\n"
+            "seeing this error in a Git repo, try updating the DVC "
+            "configuration with\n`dvc config core.no_scm False`."
```
efiop, line 32: Doesn't seem like we need newlines here. Also slight adjustment for `False`
```suggestion
            "Only supported for Git repositories. If you're "
            "seeing this error in a Git repo, try updating the DVC "
            "configuration with `dvc config core.no_scm false`."
```

dvc/scm/base.py
```diff
@@ -24,6 +24,16 @@ class RevError(SCMError):
     pass
 
 
+class NOSCMError(SCMError):
```
efiop, line 27:
```suggestion
class NoSCMError(SCMError):
```

tests/func/test_metrics.py
```diff
@@ -1005,3 +1005,13 @@ def test_metrics_diff_deleted_metric(tmp_dir, scm, dvc):
...
     }
+
+
+def test_metrics_without_scm(tmp_dir, dvc):
+    metrics = {"acc": 0.97, "recall": 0.95}
+    metrics_name = "metrics.json"
+    tmp_dir.gen({metrics_name: json.dumps(metrics)})
+    dvc.add(metrics_name)
+    dvc.metrics.add(metrics_name)
+    with pytest.raises(DvcException, match=r"Only supported for Git repos"):
```
efiop, line 1016: I guess we could
```suggestion
    with pytest.raises(NoSCMError):
```
?

dvc/repo/brancher.py
```diff
@@ -46,5 +47,7 @@ def brancher(  # noqa: E302
             for sha, names in group_by(scm.resolve_rev, revs).items():
                 self.tree = scm.get_tree(sha)
                 yield ", ".join(names)
+    except AttributeError:
+        raise NOSCMError()
```
efiop, line 51: Let's use a more explicit check instead:
```
if isinstance(scm, NoSCM):
    raise NoSCMError
```

efiop:
Sorry for the delay! Please take a look at a few comments above.

jorgeorpinel:
Resolved a few items related to the output string.

karajan1001:
> Resolved a few items related to the output string.

Thank you, I will solve other things after work.

dvc/scm/__init__.py
```diff
@@ -1,13 +1,14 @@
...
-from dvc.scm.base import Base
+from dvc.scm.base import Base, NoSCMError
 from dvc.scm.git import Git
 
 
 # Syntactic sugar to signal that this is an actual implementation for a DVC
 # project under no SCM control.
 class NoSCM(Base):
-    pass
+    def __getattr__(self, name):
```
efiop, line 10: A bit worried about this being too radical of a change, as it might be used somewhere. But no tests failed and I don't see any obvious issues right now, so let's try and see. :+1:

efiop:

dvc/scm/__init__.py
karajan1001: This built in method could only be invoked if the attribute wasn't found the usual ways. It overrides all the exceptions like `'NoSCM' object has no attribute 'resolve_rev'` to A `NoSCMError`.

PR #3658, opened 2020-04-20
metrics: rework metrics implementation

### LINKED ISSUE #3572: Support yaml as metrics file

efiop:
Dropping support for all types except json and yaml. No more xpath
support.

Fixes #3572

@dmpetrov agreed to provide docs PR for this. :pray: 


efiop:
Need to use custom loader for yaml for metrics and params to load float as Decimal. That way it will be much nicer looking and we won't have to bother with precision.

dvc/command/metrics.py
```diff
@@ -224,19 +181,6 @@ def add_parser(subparsers, parent_parser):
...
-        help=(
-            "Type of metrics (json/yaml). "
-            "It can be detected by the file extension automatically. "
-            "Unsupported types will be treated as raw."
-        ),
-        metavar="<type>",
-    )
-    metrics_show_parser.add_argument(
-        "-x", "--xpath", help="json/yaml path.", metavar="<path>",
-    )
```
shcheklein, line 239: I thought we wanted to keep for metrics show (may be diff as well)?

shcheklein:
Do we flatten the table in metrics show as we do for metrics diff now?

dvc/command/metrics.py
efiop: Not in the same shape. As a filter - maybe, but later.

efiop:
> Do we flatten the table in metrics show as we do for metrics diff now?

Not like in `metrics diff` but somewhat similar. `dvc metrics show` didn't get a lot of attention because of the upcoming `show` for experiments, so we agreed to just not touch it much for now.

PR #3663, opened 2020-04-22
output: use strings instead of PathInfo for performance reasons

### LINKED ISSUE #3635: remote: loading dirs is exteremely slow because of PathInfo

pmrowla:

Partial fix for #3635.

### DIFF
--- a/dvc/output/base.py
+++ b/dvc/output/base.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from urllib.parse import urlparse
 from copy import copy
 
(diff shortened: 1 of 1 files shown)


pmrowla:
The optimization in `OutputBase._collect_used_dir_cache` improves the `_cloud_status` runtime from `938s` to `337s` in cProfile, for an S3 remote with 2M files.

There are still several places we can improve in `RemoteBASE` and `RemoteLOCAL`, since `remote.checksum_to_path_info` is slow when dealing with large numbers of files. The issue here would be that it looks like it may end up being a fairly substantial refactoring, since optimizing where pathinfo is currently used in `RemoteLOCAL.status` ends up affecting several other places in the upload/download pipeline which currently expect pathinfos. In testing, I was able to get `_cloud_status` runtime down to `221s` (but with broken push/pull).

@efiop, should I continue with this or do we want to just leave it at only the used_dir_cache change for now?

efiop:
@pmrowla Your call. I would be perfectly happy to merge this first though. Astonishing results!

efiop:
@pmrowla About checksum_to_path_info. We used to have `checksum_to_path` not that long ago, so maybe we need to go back and just convert to PathInfo where we really need it. Or we could have both checksum_to_path and checksum_to_path_info and use where appropriate, not sure.

pmrowla:
@efiop in that case, once I fix whatever is failing CI in windows I'll mark this PR as reviewable/mergeable. We can leave the original issue open and I'll keep looking into optimizing checksum_to_path_info

PR #3665, opened 2020-04-22
Convert all paths in config to posix-style

### LINKED ISSUE #3664: Config: Cache dir is inconsistant on Unix and Windows
dvc version = 0.93.0
Platform = Ubuntu and Windows 10

On Windows, calling `dvc cache dir ..\xyz` results in an entry in `.dvc/config` like

    [cache]
        dir = ..\..\xyz

On Linux it results in 

    [cache]
        dir = ../../xyz

This inconsistency results in noise in the git history when you work sometimes on Windows, sometimes on Linux. 

Manually editing the file to a Posix-style path on Windows works fine.

charlesbaynham:

Closes #3664


dvc/config.py
```diff
@@ -320,8 +330,10 @@ def rel(path):
                 return path
 
             if isinstance(path, RelPath) or not os.path.isabs(path):
-                return relpath(path, conf_dir)
-            return path
+                path = relpath(path, conf_dir)
+
+            posix_path = str(PurePosixPath(Path(path)))
```
efiop, line 335: So looks like this will also work with windows abs paths, nice! Could you please add a unit test or something for this helper? Just to make sure it behaves as expected. 

Btw, maybe our `PathInfo.as_posix` from `dvc/path_info.py` might come in handy.

charlesbaynham: Ah nice, I've changed to `PathInfo` for consistancy

charlesbaynham:
Unit test added and functional tests updates

tests/unit/command/test_cache.py
```diff
@@ -0,0 +1,40 @@
...
+
+from dvc.config import Config
+
+
+@pytest.mark.skipif(os.name == "nt", reason="Linux only")
+def test_path_replacement():
+
+    config_filename = "./config/file"
+
+    # Extract the "rel()" function from Config._save_paths
```
efiop, line 13: This is hacky, you could've just moved `rel` helper outside and tested it separately :slightly_smiling_face:

tests/unit/command/test_cache.py
```diff
@@ -0,0 +1,40 @@
+import os
+import pytest
+from mock import MagicMock
+
+from dvc.config import Config
+
+
+@pytest.mark.skipif(os.name == "nt", reason="Linux only")
+def test_path_replacement():
```
efiop, line 9: Let's move these tests to `tests/unit/test_config.py` please.

tests/func/test_remote.py
```diff
@@ -47,7 +48,7 @@ def test_relative_path(self):
         # dir path written to config should be just one level above.
         rel = os.path.join("..", dname)
         config = configobj.ConfigObj(self.dvc.config.files["repo"])
-        self.assertEqual(config['remote "mylocal"']["url"], rel)
+        self.assertEqual(Path(config['remote "mylocal"']["url"]), Path(rel))
```
efiop, line 51: Let's just
```
import posixpath
...
rel = posixpath.join("..", dname)
...
self.assertEqual(config['remote "mylocal"']["url"], rel)
```
just to make it more explicit :slightly_smiling_face:

efiop: Same in the tests above

efiop:
@charlesbaynham Looks great! :tada:  Just a few minor finishing touches left.

tests/unit/command/test_cache.py
charlesbaynham: It is quite. I thought about moving it out, but it needs to have `conf_dir` embedded into it. I haven't dug into `Schema` but I assume I'm not allowed to change the signature of `func`. 

Then again, I could use a partial which would be rather less hacky. I'll do that

efiop: @charlesbaynham Oh, missed that one. Another way to test this would be to use func tests for `config`. Instead of unit testing this particular helper. I probably should've suggested it initially.

tests/func/test_remote.py
charlesbaynham: How about 

```
def test_relative_path(self):
        dname_parts = ["..", "path", "to", "dir"]
        dname = posixpath.join(*dname_parts)
        ret = main(["remote", "add", "mylocal", dname])
        self.assertEqual(ret, 0)

        # NOTE: we are in the repo's root and config is in .dvc/, so
        # dir path written to config should be just one level above.
        rel = poxispath.join("..", *dname_parts)
        config = configobj.ConfigObj(self.dvc.config.files["repo"])
        self.assertEqual(Path(config['remote "mylocal"']["url"]), Path(rel))
```

instead? That way we still test passing in a Windows path and getting out a posix one.

tests/unit/command/test_cache.py
efiop: Ok, one more thing that I didn't notice initially: this change will also affect other paths in the config. Trying to think if that could be bad in some application...

tests/unit/command/test_cache.py
charlesbaynham: Sure!

tests/func/test_remote.py
efiop: Oops, sorry, didn't notice that `dname` is joined above. So I suppose you really meant:
```
        dname_parts = ["..", "path", "to", "dir"]
        dname = os.path.join(*dname_parts)
        ...
        rel = poxispath.join("..", *dname_parts) 
```
right? This actually qualifies as a func test that makes the unit tests we have below redundant :slightly_smiling_face: Maybe consider just keeping these instead.

charlesbaynham: Oops, you're right of course. Re the other tests, do you mean the others in this PR? If so, I think it's worth keeping them since they test relative, absolute and remote paths on both windows and linux, whereas this test is only for a relative path. Or did you mean pre-existing ones in this file?

efiop:
Thank you @charlesbaynham ! :pray:

charlesbaynham:
Very elegant tests, I learnt something :)

PR #3672, opened 2020-04-24
remote: use string paths over PathInfo for performance reasons

### LINKED ISSUE #3635: remote: loading dirs is exteremely slow because of PathInfo

pmrowla:

Should close #3635.

* LocalRemote now uses string paths instead of PathInfo in places where we expect to be dealing with a large number of paths (`status` and `LocalRemote.cache_exists`)
* `_process`/`upload`/`download` pipeline has been left alone for now - in the event that we are pushing or pulling enough files that using PathInfo's becomes noticeably slow, the expectation is already going to be that the operation will be slow, given that we will be uploading/downloading a lot of data.

### DIFF
--- a/dvc/remote/local.py
+++ b/dvc/remote/local.py
@@ -336,11 +336,11 @@ class LocalRemote(BaseRemote):
                     )
                 )
         return self._make_status(
-            named_cache, remote, show_checksums, local_exists, remote_exists
+            named_cache, show_checksums, local_exists, remote_exists
         )
 
     def _make_status(
-        self, named_cache, remote, show_checksums, local_exists, remote_exists
+        self, named_cache, show_checksums, local_exists, remote_exists
     ):
         def make_names(checksum, names):
             return {"name": checksum if show_checksums else " ".join(names)}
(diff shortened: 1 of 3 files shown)


pmrowla:
`_cloud_status` for 2M files is down to 159s from 938s in #3634

Screen Shot 2020-04-24 at 11 25 37 AM

dvc/remote/local.py
```diff
@@ -67,6 +67,13 @@ def cache_dir(self, value):
     def supported(cls, config):
         return True
 
+    @cached_property
+    def cache_path(self):
+        return os.path.abspath(self.cache_dir)
```
pmrowla, line 72: If path is not abspath or relpath from current working dir, posix `lstat()` syscall runtime doubles. `Pathinfo.__str__` uses relpath from cwd, but since cwd may change during runtime (like in some of our erepo tests), we can't cache `relpath(self.cache_dir)` and we don't want to compute it each time, so we just use abspath.

pmrowla:
@efiop solution is more generalized now. `remote.checksum_to_path` has been re-added to return string path on remotes which can benefit from it performance-wise (currently only local). On other remotes it just points to `checksum_to_path_info`

PR #3675, opened 2020-04-24
remote.ssh: suppress paramiko logging

### LINKED ISSUE #3482: Secsh channel 10 open FAILED: open failed: Connect failed

version 0.88.0
installed via pip on Linux and Mac.
Syncing to a Ubuntu host via ssh.

Everything seems to be working but I get a cryptic warning message every time I do anything. For example:

dvc push
  0% Querying cache in ssh://bror@zotac.bror.us/media/sda2/dvc/first_day| |0/42 Secsh channel 10 open FAILED: open failed: Connect failed
Secsh channel 10 open FAILED: open failed: Connect failed
Secsh channel 10 open FAILED: open failed: Connect failed
  2% /media/sda2/dvc/first_day/4e/4b31f0c5784a2e185d88a3120cac19| |1/42 [00:02<0Secsh channel 10 open FAILED: open failed: Connect failed
Everything is up to date.                          

This is probably an edge case due to my setup but I'm not sure how to quiet the message or resolve the issue.

pmrowla:

Will close #3482.

paramiko is loaded lazily, so the paramiko logger is not suppressed in `dvc/logger.py::disable_other_loggers()`. This leads to paramiko emitting expected error messages when we hit the max SFTP channel limit, even though it is not an error condition for DVC. The logger should just be suppressed when we load paramiko, since any actual SSH/SFTP errors will be caught and logged as DvcExceptions.

### DIFF
--- a/dvc/remote/ssh/connection.py
+++ b/dvc/remote/ssh/connection.py
@@ -9,6 +9,8 @@ from funcy import cached_property
 
 try:
     import paramiko
+
+    logging.getLogger("paramiko.transport").disabled = True
 except ImportError:
     paramiko = None
 
--- a/dvc/remote/ssh/connection.py
+++ b/dvc/remote/ssh/connection.py
@@ -9,8 +9,6 @@ from funcy import cached_property
 
 try:
     import paramiko
-
-    logging.getLogger("paramiko.transport").disabled = True
 except ImportError:
     paramiko = None
 
@@ -43,6 +41,14 @@ class SSHConnection:
         self.timeout = kwargs.get("timeout", 1800)
 
         self._ssh = paramiko.SSHClient()
+
+        # Explicitly disable paramiko logger. Due to how paramiko dynamically
+        # loads loggers, it is not disabled by DVC disable_other_loggers().
+        # See https://github.com/iterative/dvc/issues/3482
+        self._ssh.set_log_channel("dvc.paramiko")
+        logging.getLogger("dvc.paramiko").disabled = True
+        logging.getLogger("dvc.paramiko.sftp").disabled = True
+
         self._ssh.load_system_host_keys()
         self._ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
 


dvc/remote/ssh/connection.py
```diff
@@ -9,6 +9,8 @@
 
 try:
     import paramiko
+
+    logging.getLogger("paramiko.transport").disabled = True
```
pmrowla, line 13: Alternatively, we could explicitly suppress this only for `open_max_sftp_channels()`. But in the event that any SSH/SFTP errors occur, they would be caught and raised/logged as DvcExceptions, so we would just end up with duplicate DVC + paramiko error messages.

skshetry: This only disables `paramiko.transport` logger, right? Should we disable all of the paramiko loggers, or is this enough?

pmrowla: I'm not sure exactly. The [SFTPClient](https://github.com/paramiko/paramiko/blob/dccc563c3a581a14ceaa261b5fe0d08539f84b87/paramiko/sftp_client.py#L90) class used in DVC uses the `paramiko.transport` logger. Looking through their code, we could potentially also suppress `paramiko`, `paramiko.transport.sftp`, `paramiko.sftp`, but only using `paramiko.transport` suppresses the specific error message we want to hide.

It's interesting to note that calling `disable_other_loggers()` here still does not suppress the error message. I guess that unless we explicitly reference a name, the loggers aren't actually registered until we instantiate some paramiko client object?

efiop: @pmrowla Maybe it is because of `paramiko` being loaded dynamically, not sure.

skshetry: >  I guess that unless we explicitly reference a name, the loggers aren't actually registered until we instantiate some paramiko client object?

@pmrowla, @efiop, both are correct. When the `SFTPClient` is instantiated, [it creates a logger](https://github.com/paramiko/paramiko/blob/dccc563c3a581a14ceaa261b5fe0d08539f84b87/paramiko/sftp_client.py#L125-L127). The default is `paramiko.transport` + `.sftp` but anyone can set it via `transport.set_log_channel()` it seems.

And, you can see in the [utils](https://github.com/paramiko/paramiko/blob/dccc563c3a581a14ceaa261b5fe0d08539f84b87/paramiko/util.py#L270-L273), you can see that it instantiates logger from the name.

efiop: for the record: agreed that we need to put a comment explaining this.

skshetry: Another approach can be setting logger with custom name that we can control and disable it explicitly:
```
self._ssh = paramiko.SSHClient()
self._ssh.set_log_channel("<some_name>)
```

The advantage of this is it won't affect users that are using `paramiko` in the python scripts + dvc `Repo` API (rare combination though).

pmrowla: Updated to use `self._ssh.set_log_channel("dvc.paramiko")`, and then explicitly disable that plus `dvc.paramiko.sftp`.

dvc/remote/ssh/connection.py
```diff
@@ -41,6 +41,14 @@ def __init__(self, host, *args, **kwargs):
         self.timeout = kwargs.get("timeout", 1800)
 
         self._ssh = paramiko.SSHClient()
+
+        # Explicitly disable paramiko logger. Due to how paramiko dynamically
+        # loads loggers, it is not disabled by DVC disable_other_loggers().
+        # See https://github.com/iterative/dvc/issues/3482
+        self._ssh.set_log_channel("dvc.paramiko")
+        logging.getLogger("dvc.paramiko").disabled = True
+        logging.getLogger("dvc.paramiko.sftp").disabled = True
```
skshetry, line 50: Can we just add a `NullHandler` to the parent (i.e. `dvc.paramiko`)? I think that makes all child loggers a no-op?

But, I'm fine as-is, it was just a question (I'm not really sure :smile:).

efiop: @pmrowla What do you think?

pmrowla: Testing it out, it works if I add a `NullHandler()` to the root `paramiko` logger, but not if we are using a custom `dvc.paramiko` log channel.

So this works
```
logging.getLogger("paramiko").addHandler(logging.NullHandler())
```

but for some reason this does not
```
self._ssh.set_log_channel("dvc.paramiko")
logging.getLogger("dvc.paramiko").addHandler(logging.NullHandler())
```

If we use `NullHandler` with the root `paramiko` logger, it also works with the statement back at the top of the file after the initial `import paramiko`, but I don't really have a preference either way.
@efiop

skshetry: @pmrowla, what about the following?
```
logging.getLogger("dvc.paramiko").addHandler(logging.NullHandler())
self._ssh.set_log_channel("dvc.paramiko")
```

pmrowla: @skshetry yeah I also tried that and it does not work either

skshetry: Okay, thanks for trying out. I'll resolve this.

PR #3676, opened 2020-04-24
dvc: implement multistage dvcfile

skshetry:

https://github.com/iterative/dvc.org/issues/1173

#### TODO
- [x] `pull/push/gc/fetch` collect outs test
- [x] `checkout` tests
- [x] `remove` tests
- [x]  `destroy` for new lockfiles and pipeline files.
- [x] `ls/get/import` test
- [x] `status` tests 
- [x] `commit` tests 
- [x] `lock/unlock` tests
- [x] `run` tests
- [x] `repro` tests
- [x] `pipelines` tests 
- [x] Check DS and CC

#### On Future Iterations
- [ ] `params` (In a separate PR)
- [ ] UX for targets (In a separate PR)
- [ ] Should not allow tracking `lockfile` (will do after it's fixed what the default filename will be, also it already avoids tracking `pipelines.lock` file if inside a directory. But, `dvc add pipelines.lock` and `dvc run -o pipelines.lock` still work.)


dvc/schema.py
```diff
@@ -16,10 +16,11 @@
     StageParams.PARAM_ALWAYS_CHANGED: bool,
 }
 
+DATA_SCHEMA = {Required("md5"): output.CHECKSUM_SCHEMA, Required("path"): str}
```
efiop, line 19: `md5` is not the only checksum we have. It might also be `etag` for stuff like external s3 deps/outputs and so on.

skshetry: Missed it. Thanks.

skshetry: It's fixed by a6a7bc1.

dvc/utils/__init__.py
```diff
@@ -377,14 +379,28 @@ def format_link(link):
...
+        match.group("tag"),
+    )
     if not path:
-        logger.warning("Assuming file to be '%s'", default)
-    return path or default, name
+        path = default or PIPELINE_FILE
+        logger.warning("Assuming file to be '%s'", path)
+
+    if os.path.basename(path) == PIPELINE_LOCK:
+        raise Exception(
```
efiop, line 402: Should be DvcException.

dvc/repo/remove.py
```diff
@@ -1,14 +1,22 @@
...
+    dvcfile = Dvcfile(self, path)
+    stages = list(dvcfile.stages.filter(name).values())
+    for stage in stages:
         stage.remove_outs(force=True)
-    else:
-        stage.remove(force=True)
 
-    return stage
+    if not outs_only:
+        dvcfile.remove()
```
efiop, line 20: if we will remove one particular stage, the whole dvcfile will be removed? This seems odd.

dvc/repo/reproduce.py
```diff
@@ -73,15 +73,15 @@ def reproduce(
     if not interactive:
         kwargs["interactive"] = self.config["core"].get("interactive", False)
 
-    active_graph = _get_active_graph(self.pipeline_graph)
+    active_graph = _get_active_graph(self.graph)
     active_pipelines = get_pipelines(active_graph)
 
-    path, name = parse_target(target)
+    path, name, tag = parse_target(target)
```
efiop, line 79: btw might be a good time to throw out tags. But could be done separately too.

dvc/repo/run.py
```diff
@@ -33,30 +38,34 @@ def _get_file_path(**kwargs):
...
+    from dvc.dvcfile import Dvcfile, PIPELINE_FILE
 
-    stage_cls, path = PipelineStage, fname or DVC_FILE
-    if not kwargs.get("name"):
+    stage_cls, path = PipelineStage, PIPELINE_FILE
+    stage_name = kwargs.get("name")
+    if not stage_name:
         kwargs.pop("name", None)
-        stage_cls, path = Stage, fname or _get_file_path(**kwargs)
+        stage_cls, path = Stage, fname or _get_file_path(kwargs)
```
efiop, line 47: what's the point of doing it in a single line? Looks confusing :)

dvc/repo/reproduce.py
skshetry: I wanted to throw this away, but kept it for now. This should be my next PR, just want to get rid of it.

dvc/repo/run.py
```diff
@@ -33,30 +38,34 @@ def _get_file_path(**kwargs):
...
+    if dvcfile.exists():
+        if stage_name and stage_name in dvcfile.stages:
+            raise DuplicateStageName(stage_name, dvcfile)
+        if stage_cls != PipelineStage:
+            dvcfile.remove_with_prompt(force=kwargs.get("overwrite", True))
+
+    try:
+        self.check_modified_graph([stage])
+    except OutputDuplicationError as exc:
+        raise OutputDuplicationError(exc.output, set(exc.stages) - {stage})
```
efiop, line 66:
```suggestion
        raise OutputDuplicationError(exc.output, set(exc.stages) - {stage}) from exc
```
maybe?

efiop: Also, not sure I understand why we need to remove stage from exc.stages. We didn't do that before , did we?

skshetry: The stage that we are creating is also shown in the error message:
```sh
ERROR: file/directory 'a2.txt' is specified as an output in more than one stage:
        a2.txt.dvc
	pipelines.yaml:copy-a2-txt
```

dvc/schema.py
```diff
@@ -16,10 +17,11 @@
     StageParams.PARAM_ALWAYS_CHANGED: bool,
 }
 
+DATA_SCHEMA = {**CHECKSUMS_SCHEMA, Required("path"): str}
 LOCK_FILE_STAGE_SCHEMA = {
     Required(StageParams.PARAM_CMD): str,
-    Required(StageParams.PARAM_DEPS): {str: output.CHECKSUM_SCHEMA},
-    Required(StageParams.PARAM_OUTS): {str: output.CHECKSUM_SCHEMA},
+    StageParams.PARAM_DEPS: [DATA_SCHEMA],
```
efiop, line 23: I suppose this doesn't support `dvc import`, right? For it it needs `repo: ...` key

skshetry: Yup, import stages are not supported.

dvc/serialize.py
```diff
@@ -39,34 +38,25 @@ def to_dvcfile(stage: "PipelineStage"):
...
-                stage.name,
-                OrderedDict(
-                    [("cmd", stage.cmd), ("deps", deps,), ("outs", outs)]
-                ),
-            )
-        ]
-    )
+    res = {"cmd": stage.cmd}
+    deps = [
+        {"path": dep.def_path, dep.checksum_type: dep.get_checksum()}
```
efiop, line 47: This is not as simple for `dvc import`. Maybe let's implement some method in each dep/out class that would return what it thinks it needs to return? Kinda like we did previously.

dvc/stage/__init__.py
```diff
@@ -147,9 +147,22 @@ def dvcfile(self, dvcfile):
 
     def __repr__(self):
         return "Stage: '{path}'".format(
+            path=self.path_in_repo if self.path else "No path"
+        )
+
+    def __str__(self):
```
efiop, line 153: So previously str would return relpath, right? Why do we need `stage: ` here? Esp when we have `__repr__` above

efiop: Ah, i see, because of path... Tbh, a bit confusing with repr, str and addressing...

dvc/stage/__init__.py
```diff
@@ -215,9 +225,9 @@ def _changed_deps(self):
             status = dep.status()
             if status:
                 logger.debug(
-                    "Dependency '{dep}' of '{stage}' changed because it is "
+                    "Dependency '{dep}' of {stage} changed because it is "
```
efiop, line 228: "of stage: 'mystage'" looks very odd, especially that `:`

dvc/repo/remove.py
skshetry: It's handled inside `dvcfile.remove()` here. For single stage file, it's perfectly okay to delete it, but multistage file will deny removing the file unless forced. In short, pipeline file won't get deleted, single stage will be.

dvc/stage/__init__.py
```diff
@@ -229,20 +239,20 @@ def _changed_outs(self):
...
+                        out=out, stage=self, status=status[str(out)]
                     )
                 )
                 return True
 
         return False
 
-    def _changed_md5(self):
-        if self.changed_md5():
+    def stage_changed(self, warn=False):
```
efiop, line 251: Naming is confusing. Not sure how `Stage.stage_changed` and `Stage.changed` are different by the naming.

dvc/stage/__init__.py
```diff
@@ -229,20 +239,20 @@ def _changed_outs(self):
...
+        changed = self.md5 != self._compute_md5()
+        if changed and warn:
             logger.warning("DVC-file '{}' changed.".format(self.relpath))
```
efiop, line 254: Could hide to `debug` for good. It is noisy :slightly_smiling_face:

dvc/stage/exceptions.py
```diff
@@ -92,3 +92,39 @@ def __init__(self, missing_files):
...
+                name, file.relpath
+            )
+        )
+
+
+class StageNameUnspecified(DvcException):
+    def __init__(self, file):
+        super().__init__(
+            "Stage name not provided. "
+            "\nPlease specify the name as: `{0}:stage_name`".format(
```
efiop, line 110: No need for a space before \n, but i don't think \n is needed neither.
```suggestion
            "Stage name not provided."
            "Please specify the name as: `{0}:stage_name`".format(
```

dvc/stage/loader.py
```diff
@@ -91,10 +103,14 @@ def __contains__(self, name):
...
 class SingleStageLoader(collections.abc.Mapping):
-    def __init__(self, dvcfile, stage_data, stage_text=None):
+    def __init__(self, dvcfile, stage_data, stage_text=None, tag=None):
         self.dvcfile = dvcfile
         self.stage_data = stage_data or {}
         self.stage_text = stage_text
+        self.tag = tag
+
+    def filter(self, item=None):
+        return self
```
efiop, line 113:
```suggestion
        assert item is None
        return self
```
?

dvc/stage/__init__.py
skshetry: ^ For lack of better naming that works on both kind of stages.

skshetry: Any suggestions?

dvc/schema.py
skshetry: Did not we discuss to keep it single stage for now? And, to discuss those later on?

dvc/stage/loader.py
skshetry: It's a bit permissible for users to do `stage.dvc:stage_name`.

dvc/stage/__init__.py
skshetry: I do like it though:
```
Dependency 'dep' of stage: 'stage.dvc' changed because it is
```

I don't think stage needed to be uppercased.

dvc/serialize.py
skshetry: Does this need to support `import`? Or, am I missing something?

dvc/stage/__init__.py
skshetry: Another approach would be `__repr__` returning `stage: relpath:stage_name` and `__str__` returning presentational string `relpath:stage_name`.

But, my intention was to keep `__repr__` to be abspath from repo, but above approach can also be done.

dvc/repo/run.py
efiop: @skshetry That looks alright, or am I missing something?

dvc/schema.py
efiop: @skshetry Ah, yeah, that's a good point! My bad.

dvc/serialize.py
efiop: @skshetry No, it doesn't. It was my bad.

dvc/stage/__init__.py
efiop: yeah, but it doesn't need `:` too, right?

dvc/repo/run.py
skshetry: The thing is `a2.txt.dvc` does not exist at this point. So, I find it to be a bit misleading.

dvc/stage/__init__.py
skshetry: Yes, it does not need to there. I'll remove it.

skshetry: It looks kinda odd in `str(stage)`, but, I can just reserve it for presentational purpose. Removing ...

dvc/stage/__init__.py
efiop: None so far. Might get some ideas later.

PR #3686, opened 2020-04-26
gdrive: fix multi-remote workflow, cont. cleanup

shcheklein:
Address more checkboxes from the #2865.

**Specifically:**

- [X]  We should store one credentials file per remote. Solution: behave optimistically, do not require a config option or complicate default naming (difficult names that depend on remote name may lead to discrepancies in simple scenarios like renaming or changing settings of a single GDrive remote). Better detect discrepancy, cases of access failure due to wrong credentials being used.
- [X] Fix how path for the `gdrive_user_credentials_file` is handled - it should be resolved against config location if it is relative.
- [X] Cleanup and simplify code in GDrive auth and other error handlers
- [X] Minor cleanup and typos fixed in a few other cases
- [x] Add tests

Docs PR: https://github.com/iterative/dvc.org/pull/1192

----------

* [X] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here. If the CLI API is changed, I have updated [tab completion scripts](https://github.com/iterative/dvc/tree/master/scripts/completion).

### DIFF
--- a/dvc/remote/gdrive.py
+++ b/dvc/remote/gdrive.py
@@ -183,9 +183,8 @@ class GDriveRemote(BaseRemote):
     def _credentials_location(self):
         if os.getenv(GDriveRemote.GDRIVE_CREDENTIALS_DATA):
             return GDriveRemote.GDRIVE_CREDENTIALS_DATA
-        else:
-            if os.path.exists(self._gdrive_user_credentials_path):
-                return self._gdrive_user_credentials_path
+        if os.path.exists(self._gdrive_user_credentials_path):
+            return self._gdrive_user_credentials_path
         return None
 
     # Detects discrepancy in DVC config and cached credentials file.
--- a/dvc/remote/gdrive.py
+++ b/dvc/remote/gdrive.py
@@ -340,10 +340,10 @@ class GDriveRemote(BaseRemote):
             if error_code == 404:
                 raise DvcException(
                     "'{}' for '{}':\n\n"
-                    "1. Check that directory exists and you can access it.\n"
+                    "1. Check that the directory exists and you can access it.\n"
                     "2. Make sure that credentials in '{}'\n"
-                    "   are the right ones to use with this remote storage, "
-                    "e.g. use 'gdrive_user_credentials_file' config\n"
+                    "   are correct for this remote e.g. "
+                    "use the `gdrive_user_credentials_file` config\n"
                     "   option if you use multiple GDrive remotes with "
                     "different email accounts.\n\nDetails".format(
                         item_id, self.path_info, self._credentials_location
--- a/dvc/remote/gdrive.py
+++ b/dvc/remote/gdrive.py
@@ -21,21 +21,27 @@ FOLDER_MIME_TYPE = "application/vnd.google-apps.folder"
 
 
 class GDrivePathNotFound(DvcException):
-    def __init__(self, path_info):
-        super().__init__("Google Drive path '{}' not found.".format(path_info))
-
-
-class GDriveAccessTokenRefreshError(DvcException):
-    def __init__(self):
-        super().__init__("Google Drive access token refreshment is failed.")
-
-
-class GDriveMissedCredentialKeyError(DvcException):
-    def __init__(self, path):
+    def __init__(self, path_info, hint):
+        hint = "" if hint is None else " {}".format(hint)
         super().__init__(
-            "Google Drive user credentials file '{}' "
-            "misses value for key.".format(path)
+            "GDrive path '{}' not found.{}".format(path_info, hint)
         )
+
+
+class GDriveAuthError(DvcException):
+    def __init__(self, cred_location):
+
+        if cred_location:
+            message = (
+                "GDrive remote auth failed with credentials in '{}'.\n"
+                "Backup first, remove of fix it, and run the command.\n"
+                "It should do auth again and refresh the credentials.\n\n"
+                "Details:".format(cred_location)
+            )
+        else:
+            message = "Failed to authenticate GDrive remote"
+
+        super().__init__(message)
 
 
 def _extract(exc, field):
@@ -115,8 +121,7 @@ class GDriveRemote(BaseRemote):
 
         if not self.path_info.bucket:
             raise DvcException(
-                "Empty Google Drive URL '{}'. Learn more at "
-                "{}.".format(
+                "Empty GDrive URL '{}'. Learn more at {}".format(
                     config["url"],
                     format_link("https://man.dvc.org/remote/add"),
                 )
@@ -155,13 +160,11 @@ class GDriveRemote(BaseRemote):
             or not self._service_account_p12_file_path
         ):
             raise DvcException(
-                "To use service account please specify {}, {} and "
-                "{} in DVC config. Learn more at "
-                "{}.".format(
-                    "gdrive_service_account_email",
-                    "gdrive_service_account_p12_file_path",
-                    "gdrive_service_account_user_email (optional)",
-                    format_link("https://man.dvc.org/remote/modify"),
+                "To use service account set 'gdrive_service_account_email',\n"
+                "'gdrive_service_account_p12_file_path', and optionally "
+                "'gdrive_service_account_user_email'\nin DVC config. "
+                "Learn more at {}".format(
+                    format_link("https://man.dvc.org/remote/modify")
                 )
             )
 
@@ -169,9 +172,42 @@ class GDriveRemote(BaseRemote):
         if not self._use_service_account:
             if bool(self._client_id) != bool(self._client_secret):
                 raise DvcException(
-                    "Please specify Google Drive's client id and secret in "
-                    "DVC config or omit both to use defaults. Learn more at "
-                    "{}.".format(
+                    "Please specify GDrive's client id and secret in "
+                    "DVC config or omit both to use defaults.\n"
+                    "Learn more at {}".format(
+                        format_link("https://man.dvc.org/remote/modify")
+                    )
+                )
+
+    @cached_property
+    def _credentials_location(self):
+        if os.getenv(GDriveRemote.GDRIVE_CREDENTIALS_DATA):
+            return GDriveRemote.GDRIVE_CREDENTIALS_DATA
+        else:
+            if os.path.exists(self._gdrive_user_credentials_path):
+                return self._gdrive_user_credentials_path
+        return None
+
+    # Detects discrepancy in DVC config and cached credentials file.
+    # Usually happens when a second remote is added and it is using
+    # the same credentials default file. Or when someones decides to change
+    # DVC config client id or secret but forgets to remove the cached
+    # credentials file.
+    @staticmethod
+    def _validate_credentials(auth, settings):
+        if not os.getenv(GDriveRemote.GDRIVE_CREDENTIALS_DATA):
+            if (
+                settings["client_config"]["client_id"]
+                != auth.credentials.client_id
+                or settings["client_config"]["client_secret"]
+                != auth.credentials.client_secret
+            ):
+                logger.warning(
+                    "Client secret and id from the DVC config do not match "
+                    "actual ones used\nto access the remote storage. Do you "
+                    "use multiple GDrive remotes and forgot to\nset "
+                    "'gdrive_user_credentials_file' for one or more of them? "
+                    "Learn more at\n{}.\n".format(
                         format_link("https://man.dvc.org/remote/modify")
                     )
                 )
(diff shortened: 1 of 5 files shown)


dvc/remote/gdrive.py
```diff
@@ -21,21 +21,27 @@
 
 
 class GDrivePathNotFound(DvcException):
-    def __init__(self, path_info):
-        super().__init__("Google Drive path '{}' not found.".format(path_info))
+    def __init__(self, path_info, hint):
+        hint = "" if hint is None else " {}".format(hint)
+        super().__init__(
+            "GDrive path '{}' not found.{}".format(path_info, hint)
```
jorgeorpinel, line 27:
```suggestion
            "GDrive path '{}' not found. {}".format(path_info, hint)
```
or `\n`?

dvc/remote/gdrive.py
```diff
@@ -21,21 +21,27 @@
...
-    def __init__(self):
-        super().__init__("Google Drive access token refreshment is failed.")
+class GDriveAuthError(DvcException):
+    def __init__(self, cred_location):
 
+        if cred_location:
+            message = (
+                "GDrive remote auth failed with credentials in '{}'.\n"
+                "Backup first, remove of fix it, and run the command.\n"
+                "It should do auth again and refresh the credentials.\n\n"
```
jorgeorpinel, line 38:
```suggestion
                "Remove of fix it, and run the command again.\n\n"
```

dvc/remote/gdrive.py
```diff
@@ -155,51 +160,89 @@ def _validate_config(self):
...
-                "{} in DVC config. Learn more at "
-                "{}.".format(
-                    "gdrive_service_account_email",
-                    "gdrive_service_account_p12_file_path",
-                    "gdrive_service_account_user_email (optional)",
-                    format_link("https://man.dvc.org/remote/modify"),
+                "To use service account set 'gdrive_service_account_email',\n"
+                "'gdrive_service_account_p12_file_path', and optionally "
+                "'gdrive_service_account_user_email'\nin DVC config. "
+                "Learn more at {}".format(
```
jorgeorpinel, line 166:
```suggestion
                "To use service account, set `gdrive_service_account_email`,\n"
                "`gdrive_service_account_p12_file_path`, and optionally "
                "`gdrive_service_account_user_email`\nin DVC config. "
                "Learn more at {}".format(
```

dvc/remote/gdrive.py
```diff
@@ -155,51 +160,89 @@ def _validate_config(self):
...
 
         # Validate OAuth 2.0 Client ID configuration
         if not self._use_service_account:
             if bool(self._client_id) != bool(self._client_secret):
                 raise DvcException(
-                    "Please specify Google Drive's client id and secret in "
-                    "DVC config or omit both to use defaults. Learn more at "
-                    "{}.".format(
+                    "Please specify GDrive's client id and secret in "
+                    "DVC config or omit both to use defaults.\n"
```
jorgeorpinel, line 176:
```suggestion
                    "Please specify GDrive's client ID and secret in "
                    "DVC config or omit both to use the defaults.\n"
```

dvc/remote/gdrive.py
```diff
@@ -155,51 +160,89 @@ def _validate_config(self):
...
+                != auth.credentials.client_id
+                or settings["client_config"]["client_secret"]
+                != auth.credentials.client_secret
+            ):
+                logger.warning(
+                    "Client secret and id from the DVC config do not match "
+                    "actual ones used\nto access the remote storage. Do you "
+                    "use multiple GDrive remotes and forgot to\nset "
+                    "'gdrive_user_credentials_file' for one or more of them? "
+                    "Learn more at\n{}.\n".format(
```
jorgeorpinel, line 209:
```suggestion
                    "Client ID and secret configured do not match the "
                    "actual ones used\nto access the remote. Do you "
                    "use multiple GDrive remotes and forgot to\nset "
                    "`gdrive_user_credentials_file` for one or more of them? "
                    "Learn more at\n{}.\n".format(
```
> Needs rewrap

But I'm not sure I understand this part: `... do not match the actual ones used access the remote.`

dvc/remote/gdrive.py
```diff
@@ -272,18 +315,42 @@ def _list_params(self):
...
+            error_code = exc.error.get("code", 0)
+            if error_code == 404:
+                raise DvcException(
+                    "'{}' for '{}':\n\n"
+                    "1. Check that directory exists and you can access it.\n"
+                    "2. Make sure that credentials in '{}'\n"
+                    "   are the right ones to use with this remote storage, "
+                    "e.g. use 'gdrive_user_credentials_file' config\n"
+                    "   option if you use multiple GDrive remotes with "
+                    "different email accounts.\n\nDetails".format(
```
jorgeorpinel, line 348:
```suggestion
                    "1. Check that the directory exists and you can access it.\n"
                    "2. Make sure that credentials in '{}'\n"
                    "   are correct for this remote e.g. "
                    "use the `gdrive_user_credentials_file` config\n"
                    "   option if you use multiple GDrive remotes with "
                    "different email accounts.\n\nDetails".format(
```
> Needs rewrap

But do we really need the last part `e.g. use the ...` ? Feels a little long.

dvc/remote/gdrive.py
jorgeorpinel: But actually, not sure `\n`s are needed, I remember @efiop mentioning somewhere that they're not? (This would apply to all the other comments.)

dvc/remote/gdrive.py
shcheklein: if hint is absent (None) we don't want to see an extra space, see a line above please

dvc/remote/gdrive.py
shcheklein: again ... again (repetition with the next sentence)

`Backup ..` - I think it's better to include it. What is your concern here?

dvc/remote/gdrive.py
shcheklein: Yes, this is arguable decision. There is one place in this PR (where hint is particularly long and has 1. 2. ,etc) - in this case I would keep it for sure ... in this specific case - I would keep them as well, just makes it easier to read, but agree that it puts some additional burden to maintain that properly + breaks dev logs. I made this decision nonetheless - users come first, developers after that :)

dvc/remote/gdrive.py
jorgeorpinel: Just trying to simplify. Shouldn't they decide whether to back it up? Why back it up if it's not working?
As for "again", I don't see much problem in the repetition but actually I would just remove the last sentence altogether, it may be obvious that you'll go through the same process when you run the command again.

Up to you, I don't have string opinions here.

dvc/remote/gdrive.py
jorgeorpinel: OK but I would still use back quotes in all these. Single quotes are used to signal dynamic output in this repo from what I remember. Meaning output that comes from the `.format()` clause, especially if it's a variable.

dvc/remote/gdrive.py
jorgeorpinel: We use upper case "ID" in the GDrive guide, https://dvc.org/doc/user-guide/setup-google-drive-remote:

image

image

dvc/remote/gdrive.py
jorgeorpinel: Same insisting suggestion about upper case ID, back quotes, and also I don't know if "from the DVC config" is a phrase that's easy to understand, but not sure how else to say it. Just "configured" may not be specific enough.

dvc/remote/gdrive.py
jorgeorpinel: "Check that **the** directory"
"are the right ones to use with" -> "are correct for" seems more direct and simple? 🙂 
Also, I see "remote storage" as a concept, when specifying "a remote" or "a DVC remote" I try not to use the word "storage".
Back quotes

shcheklein: @jorgeorpinel I already applied almost all your suggestions before resolving :) please, refresh the page?

dvc/remote/gdrive.py
shcheklein: Resolving, I think it should be fine. And it's def better to ask to backup - I don't want to take any responsibility for someone deleting some credentials by mistake.

dvc/remote/gdrive.py
jorgeorpinel: Oh. Weird. It didn't refresh earlier, weird. OK, thanks! The first sentence is still missing a word though: "Check **that** the directory" 🙂

dvc/remote/gdrive.py
```diff
@@ -21,21 +21,27 @@
...
-class GDriveAccessTokenRefreshError(DvcException):
-    def __init__(self):
-        super().__init__("Google Drive access token refreshment is failed.")
+class GDriveAuthError(DvcException):
+    def __init__(self, cred_location):
 
+        if cred_location:
+            message = (
+                "GDrive remote auth failed with credentials in '{}'.\n"
+                "Backup first, remove of fix them, and run the command.\n"
```
jorgeorpinel, line 37: What command? I'm not sure it's clear/obvious.

dvc/remote/gdrive.py
```diff
@@ -248,7 +291,11 @@ def _ids_cache(self):
         cache = {
             "dirs": defaultdict(list),
             "ids": {},
-            "root_id": self._get_item_id(self.path_info, use_cache=False),
+            "root_id": self._get_item_id(
+                self.path_info,
+                use_cache=False,
+                hint="Check remote directory exists and you can access it.",
```
jorgeorpinel, line 297: "remote location", "directory in remote URL: {}" ?

dvc/remote/gdrive.py
shcheklein: it can work w/o that (I googled both options) - shorter version allows better code (not need to break the line one more time)

jorgeorpinel: "Check the directory" works by itself, but it's a slightly different meaning, to continue the phrase I believe "that" is needed (i.e. check **that** the dir something") if we're being strict. Otherwise "Confirm the dir exists" would sound better to me.

Not a big deal obviously but... user before developer 😬

shcheklein: Changed to confirm!

dvc/remote/gdrive.py
shcheklein: changed to `run DVC again` to make it extra clear

dvc/remote/gdrive.py
shcheklein: updated a bit

dvc/remote/gdrive.py
jorgeorpinel: 👍

dvc/remote/gdrive.py
```diff
@@ -155,51 +160,93 @@ def _validate_config(self):
...
+                    "DVC config or omit both to use the defaults.\n"
+                    "Learn more at {}".format(
+                        format_link("https://man.dvc.org/remote/modify")
+                    )
+                )
+
+    # Helper to determine where will it read credentials from.
+    # Mostly useful for tests, exception messages, etc
+    # Returns either env variable name if it's set or actual path to the
+    # credentials file
```
efiop, line 185: Should this be a docstring?

tests/func/remote/test_gdrive.py
```diff
@@ -0,0 +1,56 @@
...
+            [
+                "remote",
+                "modify",
+                remote_name,
+                "gdrive_user_credentials_file",
+                credentials,
+            ]
+        )
+        == 0
+    )
```
efiop, line 39: Let's use api instead of CLI. Just makes it so much cleaner. See `dvc/command/remote.py`

tests/func/remote/test_gdrive.py
```diff
@@ -0,0 +1,56 @@
...
+
+    # We need to load repo again to test updates to the config
+    str_path = fspath(tmp_dir)
+    repo = Repo(str_path)
+
```
efiop, line 44: Probably won't be needed if you use API as requested above.

tests/func/remote/test_gdrive.py
shcheklein: Was thinking since it's functional it's better to test "real" scenario ... API is _almost_ the same, but not exactly.

tests/func/remote/test_gdrive.py
shcheklein: I'll check, but most likely we'll need it anyway ... we change remote config and need initialize Repo again

we can for sure just initialize remote directly ... again, will be most like a unit test then

dvc/remote/gdrive.py
shcheklein: yep, can change to docstrings

tests/func/remote/test_gdrive.py
efiop: functional doesn't mean CLI. We try to use API in functional tests and just simply unit-test CLI, because CLI doesn't have much logic in it. In this PR you don't really change `dvc remote` CLI, but rather config API, so there isn't much sense in using CLI here.

shcheklein: do you mean something like:

```py
with self.config.edit() as conf:
     conf["remote"][name] = {url ... gdrive_...}
```

I had this initially :)

really, really don't like this idea. It is not functional to me. I don't see an actual Repo API to do these operations.

efiop: @shcheklein Then your test is out-of-place here. And maybe it is actually. You intent on testing that config properly resolves your paths, but in the test you touch a whole lot of things. Consider removing this test and writing a unit test for config object to make sure that it properly resolves `gdrive_user_credentials_file`. That would be much nicer.

shcheklein: I'm testing this scenario:

```
dvc add remote gdrive
dvc add remote modify gdrive_user_credentials_file relative_path
# path is saved properly
dcv <any command that uses gdrive_user_credentials_file> <--- make sure that it reads it properly
```

this is exactly _functional_ flow behavior from the end-user perspective. That's how `gdrive_user_credentials_file` is documented more or less.

efiop: Yeah, but the changes you've introduced only seem to affect config. There is no real need to run it through the whole CLI stack. But could keep it as is, better than no test. Your call.

shcheklein: 👍

Issue #3570, opened 2020-04-01
test: test_pull_git_imports is flaky


efiop:
https://travis-ci.com/github/iterative/dvc/jobs/309609531#L2604

PR #3689, opened 2020-04-28
repo: use reverse post-order DFS in repro --downstream

### LINKED ISSUE #3602: repro: downstream: reproducing in wrong order
DVC version : 0.92.0 (PyPI) + MacOS

It seems `dvc repro --downstream` (could also be the `-f` option) sometimes executes DAGs in the wrong order.

Steps to reproduction:
```sh
dvc run -o 1 -o 2 -f step1.dvc 'touch 1 | touch 2'
dvc run -d 1 -o 3 -f step2.dvc 'touch 3'
dvc run -d 3 -d 2 -o 4 -f step3.dvc 'touch 4'
dvc repro --downstream -f step1.dvc
```

Last bit does : 
```
Running command:                                                        
	touch 1 | touch 2

Running command:
	touch 4

Running command:
	touch 3
```
Even though the DAG looks like this :
```
             +-----------+        
             | step1.dvc |        
             +-----------+        
            ***          ***      
           *                *     
         **                  ***  
+-----------+                   * 
| step2.dvc |                ***  
+-----------+               *     
            ***          ***      
               *        *         
                **    **          
             +-----------+        
             | step3.dvc |        
             +-----------+   

```
In this case step2 (`touch 3`) should be executed before step3 (`touch 4`)

pmrowla:

Will close #3602.

from original issue:
 > Looks like the issue is that for `--downstream`, we are reversing the DAG, and then traversing the reversed graph w/DFS pre-ordered search. In the example case, using pre-ordered search provides no guarantee that `step3` is only visited after `step2`.
> ...
> I think what we should be doing is traversing the reversed graph w/reverse post-ordered search, to ensure that dependencies are run in the correct order.

### DIFF
--- a/tests/func/test_repro.py
+++ b/tests/func/test_repro.py
@@ -1586,7 +1586,7 @@ def _hide_md5(text):
     return re.sub(r"\b[a-f0-9]{32}\b", "<md5>", text)
 
 
-def test_repro_downstream(dvc):
+def test_downstream(dvc):
     # The dependency graph should look like this:
     #
     #       E
--- a/tests/func/test_repro.py
+++ b/tests/func/test_repro.py
@@ -1586,40 +1586,49 @@ def _hide_md5(text):
     return re.sub(r"\b[a-f0-9]{32}\b", "<md5>", text)
 
 
-class TestReproDownstream(TestDvc):
-    def test(self):
-        # The dependency graph should look like this:
-        #
-        #       E
-        #      / \
-        #     D   F
-        #    / \   \
-        #   B   C   G
-        #    \ /
-        #     A
-        #
-        assert main(["run", "-o", "A", "echo A>A"]) == 0
-        assert main(["run", "-d", "A", "-o", "B", "echo B>B"]) == 0
-        assert main(["run", "-d", "A", "-o", "C", "echo C>C"]) == 0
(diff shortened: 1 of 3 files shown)


tests/func/test_repro.py
```diff
@@ -1581,40 +1581,49 @@ def _hide_md5(text):
...
-        #    /
-        #   B
-        #
-        evaluation = self.dvc.reproduce("B.dvc", downstream=True, force=True)
-
-        assert len(evaluation) == 3
-        assert evaluation[0].relpath == "B.dvc"
-        assert evaluation[1].relpath == "D.dvc"
-        assert evaluation[2].relpath == "E.dvc"
+def test_repro_downstream(dvc):
```
skshetry, line 1584: @pmrowla, the same tests are duplicated on another place:

https://github.com/iterative/dvc/blob/fefbcc0b4ccd5a93ab96665c714a93b9fbdd1fa0/tests/func/test_repro_multistage.py#L175

Important tests were duplicated because this single-stage file format might die soon.

pmrowla: multistage test updated in 811e343

pmrowla:
test_pull_git_imports is occasionally failing on travis, but unrelated to this issue (see #3570)

Issue #2301, opened 2019-07-20
dvc: finalize `dvc tag` feature


efiop:
`dvc tag` feature was implemented awhile ago, but is hidden from CLI since then. Need to give it a good thought in current dvc model (especially with `dvc import` and `dvc get`) and see what needs to be done(if anything) for it next.

Related #1766

shcheklein:
To give everyone more context.

The idea was to give users a way to assign custom tags to files that are under DVC control:

```
dvc tag add v1.0 model.pkl
```

This command would update the DVC-file `model.pkl` output is defined in to save the tag name, + current checksum of the output.

**Why don't we reuse `git tags` for that?** They are not granular enough (the apply to the commit as a whole, which includes multiple datasets, models, code, etc). They are git-specific. They will create additional mess on Github/Gitlab for projects. They will create additional mess with `-T` option across our commands.

**Why scenarios we have in mind those tags can be useful?** In general, it's all about being able to granularly address an artifact and a certain version of that artifact. Top use cases that come to my mind are:

1. `dvc checkout dataset@v1.0` or something similar, to get a dataset of a certain version. It's instead of doing something like `git checkout -- data.dvc && dvc checkout data.dvc`, etc.
2. Use with `dvc import`, `dvc get` to provide a data artifact specific way to address it.
3. See the list of "releases" for the data artifact. It's not easy to achieve with Git tags.

Anything else I'm missing there? 

@villasv @sotte @PeterFogh @prihoda any thoughts on this guys?

It does not look like a critical feature to be honest to me, but can be really nice to have to organize data/models in a proper way. On the other had, it complicates the tool by introducing a new abstraction.

PeterFogh:
Hi @shcheklein and @efiop. In my team, we are currently working on putting one of our models (developed using a DVC pipeline) into production in Azure, using Azure blob storage and Azure functions. Version tagging of model and dataset files is an essential part of the management of production systems, however, there are many ways to reach the same effect in our use case.

Currently, we have a separate DVC stage which uploads the model files to Azure blob storage, and "tags" (i.e. names it) with the current git commit hash. And, as we do not use DVC to upload the model to Azure, I cannot see how a `dvc tag` command can help our current implementation.

**My opinion**:
We will probably not use a `dvc tag` command, mainly because of our setup of running and storing all DVC files on a remote Dask scheduler/workers, hence we neither use `dvc import` or `dvc get`.
I agree with @shcheklein, 

[...]

Issue #3693, opened 2020-04-28
cleanups: todo after implementation of pipeline file


skshetry:
Tracking things to cleanup after #3676 is merged, in addition to suggestions in the PR:
- [x] Remove `tags` (will make life simpler) #3699
- [x] Split `Repo.collect()` for pipelines usage and stage gathering:
      Data related commands don't care about orderings of pipeline stages, whereas for pipelines,
      things are more complex than `collect()` can handle.
      This will also remove duplicated warnings currently on `reproduce`.
     **Workaround on #3699 for now**
- [ ] Why is there schema/serialization/deserialization in 3 places? Yeah, true.
- [ ] Parallel logics of loading/transforming/dumping of `Output`, `Dependency` and `Stages` for single stage and pipeline stages (Requires further investigation)
- [x] `StageLoader` is too dynamic.
- [x] Messed with `OutputDuplicationsError`: Make sure it's safe to remove current stage from error message.
- [x] (Feature Request) Some way to remove stage from both pipelines.lock and pipelines.yaml? (#3881)
- [x] Renaming utils/stage.py and functions inside it. (#3923)
- [x] Make `dump()` ordered (Only for Python==3.5).
- [x] `StageParams` and `OutputParams`

Of course, we need lots of tests and UI issues fixes, but this meta issue is only regarding code cleanups.

jorgeorpinel:
I find the long name of options `--metrics` and `--plots` misleading, since they only accept one metrics/plot file. What about just `--metric` and `--plot`?

efiop:
@jorgeorpinel Those are just consistent with the rest of the options: `--outs/--outs-no-cache/etc`. We might want to introduce an alias or support commas like we do in `--params`. But for now could leave as is.

skshetry:
Most of the works is done. Next stage could be fixing the tests for the `repro`. (De)Serialization can be evaluated later, if we can do anything better. `StageParams` - the constants are something that we'll have to live with it.

Marking this as done for now.

PR #3699, opened 2020-04-29
tag: getting rid of it

skshetry:

Removing unofficial `tags`. Fixes https://github.com/iterative/dvc/issues/2301
Part of #3693


dvc/repo/__init__.py
```diff
@@ -433,7 +432,9 @@ def _collect_stages(self):
         for root, dirs, files in self.tree.walk(self.root_dir):
             for file_name in filter(is_valid_filename, files):
                 path = os.path.join(root, file_name)
-                stages.extend(list(Dvcfile(self, path).stages.values()))
+                stage_loader = Dvcfile(self, path).stages
+                with stage_loader.log_level(at=logging.DEBUG):
+                    stages.extend(stage_loader.values())
```
skshetry, line 437: @efiop, one thing that changed here related to `pipeline file` is this ^.

When we collect all stages and there's no lockfile, we repeatedly get `No lock Entry` and is very noisy. But, again,this is very useful when loading single stage. So, I'm just keeping it to debug when collecting stages and set to WARNING on other times.

dvc/utils/__init__.py
```diff
@@ -384,24 +382,23 @@ def parse_target(target, default=None):
...
-    path, name, tag = (
+    path, name = (
         match.group("path"),
         match.group("name"),
-        match.group("tag"),
     )
     if not path:
         path = default or PIPELINE_FILE
-        logger.warning("Assuming file to be '%s'", path)
+        logger.debug("Assuming file to be '%s'", path)
```
skshetry, line 397: And, also reduced this to `debug`.

dvc/stage/exceptions.py
```diff
@@ -124,5 +124,5 @@ class InvalidStageName(DvcException):
     def __init__(self,):
         super().__init__(
             "Stage name cannot contain invalid characters: "
-            "'\\', '/', '@' and ':'."
+            "'\\', '/', and ':'."
```
skshetry, line 127: Created https://github.com/iterative/dvc/issues/3700 to discuss this separately. This will allow `@` for now in the stage name.

efiop: Let's not allow @ for now, it is a useful character that will def come in handy at some point.

PR #3702, opened 2020-04-29
setup: relax python-dateutil pip version constraint to include v2.8.2.  #3701

### LINKED ISSUE #3701: Relax pip dependency versions constraints for python-dateutil
Description of the problem:
ERROR: dvc 0.93.0 has requirement python-dateutil **<2.8.1** ,>=2.1, but you'll have python-dateutil 2.8.1 which is incompatible.

It'd be good to remove or relax upper constraints on python-dateutil, to include 2.8.1 and 2.8.2.  Is there a good reason why these are not included, or something is broken after 2.8.1 ?

dchichkov:
[WIP]. Attempt to relax upper constraints on the version of python-dateutil, to include 2.8.1 and 2.8.2. 

The original constraint was:
    "python-dateutil<2.8.1,>=2.1",  # Consolidates azure-blob-storage and boto3

Requesting to do a CI run, see if new version constraint still works.

Fixes #3701

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,6 @@ class build_py(_build_py):
 
 
 install_requires = [
-    "python-dateutil<=2.8.2,>=2.1",  # Consolidates azure-blob-storage and boto3
     "ply>=3.9",  # See https://github.com/pyinstaller/pyinstaller/issues/1945
     "colorama>=0.3.9",
     "configobj>=5.0.6",
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ class build_py(_build_py):
 
 
 install_requires = [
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -47,7 +47,7 @@ def run(self):
 
 
 install_requires = [
-    "python-dateutil<2.8.1,>=2.1",  # Consolidates azure-blob-storage and boto3
+    "python-dateutil<=2.8.2,>=2.1",  # Consolidates azure-blob-storage and boto3
```
efiop, line 50:
```suggestion
```

efiop:
Thanks @dchichkov ! :pray:

PR #3711, opened 2020-04-30
dump: deterministic lockfile dump

### LINKED ISSUE #3696: lockfile: order of content changes on a repro
Eg: if we change order of `outs` in the pipeline file, and run `repro`, it's smart to detect that nothing was changed. But, if the outs or deps has really changed in the workspace and the stage is run, it will generate the lockfile which generates the content in the order of contents of pipeline file. It's not big issue, but it can be annoying to look at the diff.

But, leaving it as-is for now. Will come to this as soon as big tasks are complete.

skshetry:
The dump is no longer deterministic/dependent on the pipeline file,
but is sorted based on file names in outs, deps or params.
Also, the params inside each files are also sorted based on name.
However, the objects inside params are not sorted deterministically
as I think it's too much to sort that, and is not easy (considering
the types of objects it might hold, eg: lists, objects, etc).

This will also provide ordered dumps for Python3.5

Fixes #3696

### DIFF
--- a/dvc/serialize.py
+++ b/dvc/serialize.py
@@ -1,3 +1,6 @@
+from collections import OrderedDict
+from functools import partial
+from operator import attrgetter
 from typing import TYPE_CHECKING
 
 from funcy import rpartial, lsplit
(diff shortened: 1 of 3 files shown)


dvc/serialize.py
```diff
@@ -26,7 +32,7 @@ def _get_outs(stage: "PipelineStage"):
             bucket_key += ["no_cache"]
         key = "_".join(bucket_key)
         outs_bucket[key] = outs_bucket.get(key, []) + [o.def_path]
-    return outs_bucket
+    return [(key, outs_bucket[key]) for key in outs_bucket.keys()]
```
skshetry, line 35: Notice that we do not sort for `_get_outs()` at all. This will likely make dump for pipeline file non-deterministic. Not really
focusing it for now, as we will be rarely dumping to the pipeline file.

PR #3716, opened 2020-05-01
[WIP] repo: use unified RepoTree for erepos

### LINKED ISSUE #3611: external_repo doesn't  respect parent cache and progress bar
Somewhere during the external_repo refactor we've lost the logic that was using parent's cache dir to store cache. Currently, we are using tempdir (yes, in /tmp) to pull data too, which is really bad, as it might not have enough space available. Plus, we don't respect link type, trying to always copy stuff, which is also really bad. And on top of that we've lost our progress bar for when we are checking out the data after downloading during `import/get`.

pmrowla:

PR implements the following:

* `dvc.repo.tree.RepoTree` finished - RepoTree can be used to view a DvcTree and separate WorkingTree or GitTree as a single unified tree instance
    * `walk()`/`walk_files()` will walk and merge both trees
    * `copytree(src, dest)` will copy the contents of the tree (starting from `src`) into the dest pathname
* utils: Add equivalent versions of existing copytree/copyfile/file_md5/etc functions that work with trees and file objects instead of fs paths
* remote: Add support for trees/file objects
    * `remote.save_tree()`, `remote.save_obj()` can be used to save trees or file objects directly to local cache
* erepo: Use RepoTree for external repos
    * cloned external repos do not use `git checkout` unless `for_write=True`
    * GitTree is used to read files directly from git index rather than working with working copy files
    * only one clone is kept per erepo (rather than duplicate clones needed for every git revision that has been checked out)
* `dvc ls`: Now uses RepoTree instead of DvcTree + separate git file handling
* `dvc get`: Now uses ExternalRepo.get_external() for handling all dvc + git files
* `dvc fetch`: Now uses ExternalRepo.fetch_external() for handling all dvc + git files
* RepoDependency updated to use new erepo behavior
* `RecursiveImportError` now raised when trying to `import`/`get` a git directory which contains dvcfiles (#3087). Support for `import -R` will be implemented in a future PR.

TODO:
* [x] Cleanup state handling
* [x] Fix download issue that breaks dvcx (minor dvcx patch needed after this PR, issue has been filed in the dvcx repo)

Fixes #3611
Partial fix for #3087

### DIFF
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -166,17 +166,7 @@ class BaseExternalRepo:
             )
             failed += 1
         elif local_cache.changed_cache(info[local_cache.PARAM_CHECKSUM]):
-            if self.repo_tree.isfile(path_info):
-                with self.repo_tree.open(
-                    path_info, mode="rb", encoding=None
-                ) as fobj:
-                    local_cache.save_obj(fobj, info)
-            else:
-                for filename in self.repo_tree.walk_files(path_info):
-                    with self.repo_tree.open(
-                        filename, mode="rb", encoding=None
-                    ) as fobj:
-                        local_cache.save_obj(fobj, info)
+            local_cache.save_tree(self.repo_tree, path_info, info)
             logger.debug("fetched '{}' from '{}' repo".format(name, self.url))
             downloaded += 1
         return downloaded, failed
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -239,7 +239,7 @@ class ExternalRepo(Repo, BaseExternalRepo):
 
     def fetch_external(self, files, cache, **kwargs):
         self.cache.local.cache_dir = cache.cache_dir
-        self._fetch_external(files, self.cache.local, **kwargs)
+        return self._fetch_external(files, self.cache.local, **kwargs)
 
 
 class ExternalGitRepo(BaseExternalRepo):
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -5,7 +5,7 @@ import threading
 from contextlib import contextmanager
 from distutils.dir_util import copy_tree
 
-from funcy import cached_property, retry, suppress, wrap_with
+from funcy import cached_property, retry, wrap_with
 
 from dvc.config import NoRemoteError, NotDvcRepoError
 from dvc.exceptions import (
@@ -18,9 +18,10 @@ from dvc.exceptions import (
 )
 from dvc.path_info import PathInfo
 from dvc.repo import Repo
+from dvc.repo.tree import RepoTree
 from dvc.scm.git import Git
 from dvc.utils import tmp_fname
-from dvc.utils.fs import fs_copy, move, remove
+from dvc.utils.fs import makedirs, move, remove
 
 logger = logging.getLogger(__name__)
 
@@ -32,7 +33,9 @@ def external_repo(url, rev=None, for_write=False):
     try:
         repo = ExternalRepo(path, url)
     except NotDvcRepoError:
-        repo = ExternalGitRepo(path, url)
+        if not rev:
+            rev = "HEAD"
+        repo = ExternalGitRepo(path, url, rev)
 
     try:
         yield repo
@@ -64,12 +67,19 @@ def clean_repos():
         _remove(path)
 
 
-class ExternalRepo(Repo):
-    def __init__(self, root_dir, url):
-        super().__init__(root_dir)
-        self.url = url
-        self._set_cache_dir()
-        self._fix_upstream()
+class BaseExternalRepo:
+    def _pull_cached(self, src, dest):
+        return False
+
+    def _pull_file(self, tree, src, dest):
+        if tree.isdvc(src) and self._pull_cached(src, dest):
+            return
+        tree.copyfile(src, dest)
+
+    def _pull_dir(self, tree, src, dest):
+        if tree.isdvc(src) and self._pull_cached(src, dest):
+            return
+        makedirs(dest)
 
     def pull_to(self, path, to_info):
         """
@@ -79,43 +89,69 @@ class ExternalRepo(Repo):
         It works with files tracked by Git and DVC, and also local files
         outside the repository.
         """
-        out = None
+        logger.debug("pull_to root {} path {}".format(self.root_dir, path))
         path_info = PathInfo(self.root_dir) / path
+        logger.debug("pull_to {} {}".format(path_info, to_info))
+        tree = RepoTree(self)
 
-        with suppress(OutputNotFoundError):
-            (out,) = self.find_outs_by_path(path_info, strict=False)
-
-        try:
-            if out and out.use_cache:
-                self._pull_cached(out, path_info, to_info)
-                return
-
-            # Check if it is handled by Git (it can't have an absolute path)
-            if os.path.isabs(path):
-                raise FileNotFoundError
-
-            fs_copy(path_info, to_info)
-        except FileNotFoundError:
+        if not tree.exists(path_info):
             raise PathMissingError(path, self.url)
 
-    def _pull_cached(self, out, path_info, dest):
-        with self.state:
-            tmp = PathInfo(tmp_fname(dest))
-            src = tmp / path_info.relative_to(out.path_info)
+        if tree.isfile(path_info):
+            logger.debug("tree.isfile")
+            return self._pull_file(tree, path_info, to_info)
 
-            out.path_info = tmp
+        logger.debug("not tree.isfile")
+        self._pull_dir(tree, path_info, to_info)
 
-            # Only pull unless all needed cache is present
-            if out.changed_cache(filter_info=src):
-                self.cloud.pull(out.get_used_cache(filter_info=src))
+        # if path was DVC-only dir, there will be nothing left to recurse into
+        # after _pull_cached, and tree.isdir() will now be false
+        if tree.isdir(path_info):
+            for root, dirs, files in tree.walk(path_info):
+                logger.debug("walk {} {} {}".format(root, dirs, files))
+                root_path = PathInfo(root)
+                rel_root = path_info.relative_to(root_path)
+                for dirname in dirs:
+                    logger.debug("dir {} {}".format(root, dirs))
+                    dir_path = root_path / dirname
+                    dest_path = to_info / rel_root / dirname
+                    self._pull_dir(tree, dir_path, dest_path)
+                for filename in files:
+                    logger.debug("file {} {}".format(root, filename))
+                    file_path = root_path / filename
+                    dest_path = to_info / rel_root / filename
+                    self._pull_file(tree, file_path, dest_path)
 
-            try:
-                out.checkout(filter_info=src)
-            except CheckoutError:
-                raise FileNotFoundError
 
-            move(src, dest)
-            remove(tmp)
+class ExternalRepo(Repo, BaseExternalRepo):
+    def __init__(self, root_dir, url):
+        super().__init__(root_dir)
+        self.url = url
+        self._set_cache_dir()
+        self._fix_upstream()
+
+    def _pull_cached(self, path_info, dest):
+        (out,) = self.find_outs_by_path(path_info, strict=False)
+        if out and out.use_cache:
+            with self.state:
+                tmp = PathInfo(tmp_fname(dest))
+                src = tmp / path_info.relative_to(out.path_info)
+
+                out.path_info = tmp
+
+                # Only pull unless all needed cache is present
+                if out.changed_cache(filter_info=src):
+                    self.cloud.pull(out.get_used_cache(filter_info=src))
+
+                try:
+                    out.checkout(filter_info=src)
+                except CheckoutError:
+                    raise FileNotFoundError
+
+                move(src, dest)
+                remove(tmp)
+            return True
+        return False
 
     @wrap_with(threading.Lock())
     def _set_cache_dir(self):
@@ -180,16 +216,6 @@ class ExternalGitRepo:
 
     def find_out_by_relpath(self, path):
         raise OutputNotFoundError(path, self)
-
-    def pull_to(self, path, to_info):
-        try:
-            # Git handled files can't have absolute path
-            if os.path.isabs(path):
-                raise FileNotFoundError
-
-            fs_copy(os.path.join(self.root_dir, path), to_info)
-        except FileNotFoundError:
-            raise PathMissingError(path, self.url)
 
     @contextmanager
     def open_by_relpath(self, path, mode="r", encoding=None, **kwargs):
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -21,7 +21,6 @@ from dvc.path_info import PathInfo
 from dvc.repo import Repo
 from dvc.repo.tree import RepoTree
 from dvc.scm.git import Git
-from dvc.scm.tree import WorkingTree
 from dvc.utils import tmp_fname
 from dvc.utils.fs import makedirs, move, remove
 
@@ -77,15 +76,6 @@ class BaseExternalRepo:
     @cached_property
     def repo_tree(self):
         return RepoTree(self)
-
-    # temporary hack to make cache use WorkingTree and not GitTree, because
-    # cache dir doesn't exist in the latter.
-    @contextmanager
-    def fix_cache_tree(self):
-        saved_tree = self.tree
-        self.tree = WorkingTree(self.root_dir)
-        yield
-        self.tree = saved_tree
 
     def get_external(self, path, to_info, jobs=None):
         """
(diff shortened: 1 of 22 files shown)


dvc/external_repo.py
```diff
@@ -1,35 +1,43 @@
...
+from dvc.scm.tree import WorkingTree
 
 
 logger = logging.getLogger(__name__)
 
 
 @contextmanager
-def external_repo(url, rev=None, for_write=False):
+def external_repo(
+    url, rev=None, for_write=False, cache_dir=None, cache_types=None
```
Suor, line 29: I purposefully removed cache args from here. This way all the intermediate code needs to know about them, which makes things more complex overall.

Why do you change it back? Can we set it later?

dvc/external_repo.py
```diff
@@ -220,10 +341,6 @@ def _cached_clone(url, rev, for_write=False):
     logger.debug("erepo: making a copy of %s clone", url)
     copy_tree(clone_path, repo_path)
 
-    # Check out the specified revision
-    if rev is not None:
-        _git_checkout(repo_path, rev)
```
Suor, line 225: If we don't checkout then structure of CLONES cache should be simplified.

Suor: Another thing is this breaks current `for_write=True` use, which breaks how `dvcx` operates.

dvc/repo/tree.py
```diff
@@ -127,15 +134,86 @@ def isexec(self, path):
...
+            return
+
+        # walk and merge both trees, ensure that dvcfiles are ignored (handled
+        # as DVC outs, not as git versioned files)
+        repo_root, repo_dirs, repo_files = first(
+            self.repo.tree.walk(top, topdown=topdown)
+        )
+        dvc_root, dvc_dirs, dvc_files = first(
+            self.dvctree.walk(top, topdown=topdown)
+        )
```
Suor, line 185: Is this efficient enough?

dvc/external_repo.py
pmrowla: It seemed to me that rather than having
```
    with external_repo(...) as repo:
            if hasattr(repo, "cache"):
                repo.cache.local.cache_dir = self.repo.cache.local.cache_dir
```
everywhere we use erepos, we should just handle that in `external_repo()`, since it depends on whether or not we return an `ExternalRepo` or `GitExternalRepo`.

With the previous behavior, for `ExternalRepo` we always call `_set_cache_dir()` which sets `cache_dir` to a cached temp directory. And then later we would sometimes change `cache_dir` to some another existing cache directory.

It seemed to me that it makes more sense to just have `_set_cache_dir()` set it to the directory we want right away, rather than setting it to a temp dir that we don't actually need, and then changing it again later.

dvc/repo/tree.py
pmrowla: This is more of a placeholder while I'm getting the rest of the erepo behavior to work, but eventually, `RepoTree.walk` should only `repo.tree.walk` and `dvctree.walk` once, and do the proper merging/recursion with the initial set of iterators returned from each walk function

dvc/external_repo.py
```diff
@@ -61,68 +69,186 @@ def clean_repos():
...
+    # temporary hack to make cache use WorkingTree and not GitTree, because
+    # cache dir doesn't exist in the latter.
+    @contextmanager
+    def fix_cache_tree(self):
+        saved_tree = self.tree
+        self.tree = WorkingTree(self.root_dir)
+        yield
+        self.tree = saved_tree
+
+    def get_external(self, path, to_info, jobs=None):
```
pmrowla, line 86: this function is still mostly just old `pull_to` but what it should eventually do is more like

1. fetch needed outs from `path` into cache (`self.fetch_external([path])`)
2. `self.repo_tree.walk(path)`
3. use `repo_tree.open()` to copy (from fobj's) files from either git index or cache to `to_info`

dvc/external_repo.py
pmrowla: After looking into this some more, I get why it was structured this way now, and my changes have been reverted

dvc/external_repo.py
pmrowla: @Suor this has been partially reverted, so that when `for_write=True` we will do git checkout (in a new repo kept separate from CLONES cache), and use WorkingTree like before. Will that work for dvcx?

Suor: It should be, you may run dvcx tests to be sure.

dvc/ignore.py
```diff
@@ -139,7 +139,5 @@ def walk(self, top, topdown=True):
 
             yield root, dirs, files
 
-    def walk_files(self, top):
```
pmrowla, line 142: This function is moved into dvc.scm.tree.BaseTree

dvc/remote/base.py
```diff
@@ -259,8 +264,9 @@ def get_dir_checksum(self, path_info):
             self.cache.makedirs(new_info.parent)
             self.cache.move(tmp_info, new_info, mode=self.CACHE_MODE)
 
-        self.state.save(path_info, checksum)
-        self.state.save(new_info, checksum)
+        if is_working_tree(self.repo.tree):
+            self.state.save(path_info, checksum)
+        self.state.save(new_info, checksum, tree=self.cache.tree)
```
pmrowla, line 269: Now that `repo.tree` may be a GitTree, RemoteLocal (`self.cache` here) has it's own WorkingTree instance. When working with GitTree's, `repo.tree` must be used to look for paths in the git-index, and `RemoteLocal.tree` must be used to look for local cache paths. `state.save` also has to know which tree to use - it uses `repo.tree` by default, but when the `repo.tree` is a GitTree, `state.save` must be passed the local cache tree instead

dvc/remote/base.py
```diff
@@ -559,6 +565,41 @@ def _save(self, path_info, checksum, save_link=True):
...
+            self._save_tree_file(tree, entry_info, entry_checksum)
+
+        self.state.save(cache_info, checksum, tree=self.cache.tree)
+        if is_working_tree(tree):
+            self.state.save(path_info, checksum)
+
+    def _save_tree_file(self, tree, path_info, checksum):
+        with tree.open(path_info, mode="rb", encoding=None) as fobj:
+            self._save_obj(fobj, checksum)
+            return
```
pmrowla, line 596:
```suggestion
        with tree.open(path_info, mode="rb", encoding=None) as fobj:
            self._save_obj(fobj, checksum)
        if is_working_tree(tree):
            self.state.save(path_info, checksum)
```

tests/func/test_get.py
```diff
@@ -156,17 +156,13 @@ def test_get_to_dir(tmp_dir, erepo_dir, dname):
     assert (tmp_dir / dname / "file").read_text() == "contents"
 
 
-def test_get_from_non_dvc_master(tmp_dir, git_dir, caplog):
+def test_get_from_non_dvc_master(tmp_dir, git_dir):
```
pmrowla, line 159: changes to this test should get reverted after the logging bug is fixed

pmrowla:
There's a still bug that needs to be fixed, where git logger messages are not being suppressed and cause the caplog based tests in plot to fail, but other than that this can be reviewed now.

(but do not merge as long as it's still labeled WIP)

pmrowla:
From discussion w/@efiop: need to clean up state changes with regard to local cache (state should just be a no-op when the parent repo tree is not a working tree)

dvc/external_repo.py
```diff
@@ -65,66 +68,206 @@ def clean_repos():
...
-        super().__init__(root_dir)
-        self.url = url
-        self._set_cache_dir()
-        self._fix_upstream()
+class BaseExternalRepo:
+    _tree_rev = None
+    _local_cache = None
+
+    @contextmanager
+    def use_cache(self, cache):
```
pmrowla, line 76: This context manager lets us use an existing local cache for downloading instead of the default tmpdir erepo cache

dvc/external_repo.py
```diff
@@ -65,66 +68,206 @@ def clean_repos():
...
+        self._local_cache = cache
+        if hasattr(self, "cache"):
+            save_cache = self.cache.local
+            self.cache.local = cache
+        # make cache aware of our repo tree
+        with cache.erepo_tree(self.tree):
+            yield
+        if hasattr(self, "cache"):
+            self.cache.local = save_cache
+        self._local_cache = None
```
pmrowla, line 87: This context manager lets us use an existing local cache instead of the default tmpdir erepo cache. Before we would just edit `erepo.cache.local.cache_dir` to make downloads go where we wanted them to, but we were not using other cache settings like the proper link types.

combined w/the `Remote.erepo_tree` context manager, we can now use the correct local cache instance, and also make sure that local cache instance is aware of the erepo git tree

dvc/external_repo.py
```diff
@@ -65,66 +68,206 @@ def clean_repos():
...
+
+    def _fetch_info(self, path_info):
+        if self.repo_tree.isdvc(path_info):
+            (out,) = self.find_outs_by_path(path_info, strict=False)
+            filter_info = path_info.relative_to(out.path_info)
+        else:
+            filter_info = None
+        return path_info, filter_info
+
+    def get_external(self, path, to_info, recursive=False, **kwargs):
```
pmrowla, line 152: Future PR will be needed to add `import/get -R` support for #3087

dvc/external_repo.py
```diff
@@ -65,66 +68,206 @@ def clean_repos():
...
-            fs_copy(fspath(path_info), fspath(to_info))
-        except FileNotFoundError:
-            raise PathMissingError(path, self.url)
+        self._fetch_external(fetch_infos, **kwargs)
 
-    def _pull_cached(self, out, path_info, dest):
         with self.state:
-            tmp = PathInfo(tmp_fname(dest))
-            src = tmp / path_info.relative_to(out.path_info)
+            self.repo_tree.copytree(path_info, to_info)
```
pmrowla, line 174: `get_external` is now essentially just a `fetch` and then `checkout` (checkout is done for dvc outs in `RepoTree.copytree`)

dvc/remote/local.py
```diff
@@ -49,10 +58,24 @@ def __init__(self, repo, config):
...
         self._dir_info = {}
+        # repo.tree can be WorkingTree or GitTree, but GitTree does not contain
+        # cache directory
+        if self.repo:
+            if is_working_tree(self.repo.tree):
+                self._tree = self.repo.tree
+            else:
+                tree = WorkingTree(self.repo.root_dir)
+                if isinstance(self.repo.tree, CleanTree):
+                    tree = CleanTree(tree)
```
pared, line 69: I am not sure I follow here.
We get to this `else` statement if `repo.tree` is not `working_tree`, this happens when we are playing around with brancher.
So why do we need information whether tree on some branch is CleanTree?

pared:
Also, it seems like many py3.5 tests are failing, because we pass `PathInfo` objects to os methods that do not accept them. Probably need to insert some `fspath`'s in methods utilizing `tree`'s

dvc/remote/local.py
pmrowla: `repo.tree` will be `CleanTree(GitTree())` for external dvc repos, so I think my original logic here was that the new `WorkingTree` should also be wrapped with `CleanTree`.

But looking at it again, you're right. We only need `tree = WorkingTree(...)` to be able to see the .dvc/cache directory in this case, but it doesn't matter whether or not we wrap it with CleanTree.

pmrowla:
> Also, it seems like many py3.5 tests are failing, because we pass `PathInfo` objects to os methods that do not accept them. Probably need to insert some `fspath`'s in methods utilizing `tree`'s

I think these should all be fixed now? Travis is passing now after 8172d96

dvc/remote/local.py
pmrowla: fixed in d010f7f

dvc/utils/__init__.py
```diff
@@ -57,21 +73,37 @@ def file_md5(fname):
...
-                        chunk = data
-                    else:
-                        chunk = dos2unix(data)
+    return (None, None)
 
-                    hash_md5.update(chunk)
-                    pbar.update(len(data))
 
-        return (hash_md5.hexdigest(), hash_md5.digest())
+def tree_md5(tree, fname):
```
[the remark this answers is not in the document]
pmrowla: fixed in 40fddca

dvc/external_repo.py
```diff
@@ -65,66 +68,211 @@ def clean_repos():
...
+        files = [(PathInfo(self.root_dir) / name, None) for name in files]
+        return self._fetch_external(files, **kwargs)
+
+    def _fetch_external(self, fetch_infos, **kwargs):
+        downloaded, failed = 0, 0
 
+        with self.state:
+            for path_info, filter_info in fetch_infos:
+                if self.repo_tree.isdvc(path_info):
+                    (out,) = self.find_outs_by_path(path_info, strict=False)
```
efiop, line 195: Was kinda hoping we no longer need `find_outs_by_path` and could use tree instead.

dvc/external_repo.py
```diff
@@ -65,66 +68,211 @@ def clean_repos():
...
+        """Copy git tracked file into specified cache."""
+        downloaded, failed = 0, 0
+        if hasattr(self, "cache"):
+            local_cache = self.cache.local
+        elif self._local_cache:
+            local_cache = self._local_cache
+        else:
+            return downloaded, failed
+
+        info = local_cache.save_info(path_info)
```
efiop, line 233: Looks like we still depend on checked-out git repo, right? Because path_info needs to point to an existing git file. I though we could use `tree` directly to copy obj directly from index to our cache.

dvc/external_repo.py
```diff
@@ -165,11 +313,37 @@ def _add_upstream(self, src_repo):
         self.config["remote"]["auto-generated-upstream"] = {"url": cache_dir}
         self.config["core"]["remote"] = "auto-generated-upstream"
 
+    def get_checksum(self, path):
```
efiop, line 316: Hm, maybe it would make sense to make this part of the tree? Kinda like mtime, size, checksum(or rather hash) also sounds related. Kinda like some fs provide checksums.

dvc/external_repo.py
pmrowla: `path_info` points to a path that only exists in the git tree, so `remote.save_info` will return a checksum computed from the git-index obj. Then we do
```
elif local_cache(changed_cache...):
    local_cache.save_tree(...)
```
which copy's the obj directly from git index into our cache (if it hasn't already been copied/fetched)

dvc/external_repo.py
pmrowla: so something like `RepoTree.outs_for_path(...)`? and if the path isn't an out just raise OutputNotFoundError?

dvc/external_repo.py
pmrowla: hmm, I'm not sure about making this a part of the tree, since get_checksum is really a local cache operation and not a tree operation?

I added this function to ExternalRepo before the context managers for using the correct cache & trees, but now that we have the contextmanagers to use the right trees, I think we can just drop this function from ExternalRepo entirely. Originally this was meant to replace a `self.repo.cache.local.get_checksum(...)` call in `RepoDependency._get_checksum`, since getting checksums for git objects depends on the erepo git tree that the DependencyRepo cache is unaware of.  But now we can just use
```
cache = self.repo.cache.local
with cache.erepo_tree(erepo.tree):
    cache.get_checksum(...)
```
directly in DependencyRepo.

efiop: @pmrowla Hm, I thought we would make it accept obj so that we can compute the checksum that we need directly from git index, without us having to checkout stuff.

dvc/external_repo.py
efiop: Oh, got it, I didn't catch that initially.

dvc/external_repo.py
pmrowla: @efiop we do compute the checksum we need directly from git index (without `git checkout`) though

in LocalRemote we eventually call:
```py
    def get_file_checksum(self, path_info):
        if self.tree.exists(path_info):
            return file_md5(path_info)[0]
        if self._erepo_tree and self._erepo_tree.exists(path_info):
            return file_md5(path_info, self._erepo_tree)[0]
        return file_md5(path_info, self.repo.tree)[0]
```
and in this case, the only place `path_info` will exist is `_erepo_tree`, which is the git-index only GitTree. So `file_md5` will open the fobj and read directly from the GitTree.

dvc/external_repo.py
efiop: @pmrowla No, I actually thought about something native that would possibly abstract it away so we don't have to care about it being an output in this code at all, having tree handle all of that for us. Something like `self.cache.local.save_obj(tree.open(path))`. Obviously we'll need to somehow handle fetching inside of the tree (or rather somehow ask it what cache it needs us to download for specific path), so I'm not completely sure if this is doable in a nice way.

efiop:
For the record: discussed with @pmrowla that we could probably look into unifying the workflow even more by clearly separating `fetch` operation for the dependency and `checkout` for the output, no matter if it is dvc-tracked or git tracked.

pmrowla:
closing this, it will be reworked and split into smaller PRs

Issue #3408, opened 2020-02-25
api: support streaming from Google Drive remotes


jorgeorpinel:
> Extracted from https://github.com/iterative/dvc.org/pull/908#discussion_r382941301

GDrive is the only remote type that doesn't support file streaming. Should we work on adding it?

- [ ] and update docs (e.g. api ref)

jorgeorpinel:
p.s. maybe could be included in #2865 and closed here.

shcheklein:
@jorgeorpinel yes, let's add a check box to that list and keep/remove this ticket - I don't have a strong preference

casperdcl:
depends on https://github.com/iterative/PyDrive2/pull/39

jorgeorpinel:
Both PRs are merged now, does it mean maybe this is now supported? If so, to QA it. Thanks

casperdcl:
well it means #3910 is (maybe) no longer blocked so it's getting close to being supported :)

PR #3722, opened 2020-05-02
gdrive: download: stream & add progress

casperdcl:
- [x] ~~patch underlying `httplib2` for gdrive progress~~
- [x] ~~Also consider using something other than `httplib2`/`pydrive2` (e.g. use `http.client` directly or `requests` which both support iteration) to~~ avoid loading the whole file into RAM
  - [x] depends on https://github.com/iterative/PyDrive2/pull/30 (`pydrive>=1.4.11`)
    - > Let there be no more in-memory downloads. Let the files stream until the hard drives overfloweth. Let RAM be ignored, and let us answer the call to progress.
    - [x] use new `ApiRequestError.GetField()`
    - [x] use new `GetContentFile(callback=Tqdm.update_to)`
  - [x] depends on https://github.com/conda-forge/pydrive2-feedstock/pull/10
- [x] test
- part of #2865
  + related #3408

### DIFF
--- a/dvc/remote/gdrive.py
+++ b/dvc/remote/gdrive.py
@@ -396,14 +396,34 @@ class GDriveRemote(BaseRemote):
         param = {"id": item_id}
         # it does not create a file on the remote
         gdrive_file = self._drive.CreateFile(param)
-        bar_format = (
-            "Downloading {desc:{ncols_desc}.{ncols_desc}}... "
-            + Tqdm.format_sizeof(int(gdrive_file["fileSize"]), "B", 1024)
-        )
-        with Tqdm(
-            bar_format=bar_format, desc=progress_desc, disable=no_progress_bar
-        ):
-            gdrive_file.GetContentFile(to_file)  # TODO: actually use pbar
+
+        import httplib2
+
+        OrigClass = httplib2.HTTPConnectionWithTimeout.response_class
+
+        class Custom(OrigClass):
+            def _readall_chunked(self):
+                assert self.chunked != "UNKNOWN"
+                value = []
+                with Tqdm(
+                    total=int(gdrive_file["fileSize"]),
+                    desc=progress_desc,
+                    disable=no_progress_bar,
+                    bytes=True,
+                ) as pbar:
+                    while True:
+                        chunk_left = self._get_chunk_left()
+                        if chunk_left is None:
+                            break
+                        chunk = self._safe_read(chunk_left)
+                        value.append(chunk)
+                        pbar.update(len(chunk))
+                        self.chunk_left = 0
+                return b"".join(value)
+
+        httplib2.HTTPConnectionWithTimeout.response_class = Custom
+        gdrive_file.GetContentFile(to_file)
+        httplib2.HTTPConnectionWithTimeout.response_class = OrigClass
 
     @_gdrive_retry
     def _gdrive_delete_file(self, item_id):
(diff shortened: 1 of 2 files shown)


dvc/remote/gdrive.py
```diff
@@ -396,14 +396,34 @@ def _gdrive_download_file(
...
+                            break
+                        chunk = self._safe_read(chunk_left)
+                        value.append(chunk)
+                        pbar.update(len(chunk))
+                        self.chunk_left = 0
+                return b"".join(value)
+
+        httplib2.HTTPConnectionWithTimeout.response_class = Custom
+        gdrive_file.GetContentFile(to_file)
+        httplib2.HTTPConnectionWithTimeout.response_class = OrigClass
```
efiop, line 426: This needs to be in `finally` :)

casperdcl: P.S. not going to be relevant shortly when I push a proper fix which doesn't try to temporarily save whole files in RAM.

casperdcl:
P.S. @shcheklein I don't think we need to bother about `GetContentFile(remove_bom=True)` since we never `Upload({"convert": True})`.

shcheklein:
@casperdcl yep, that's why I was not that worried about bom in the first place (e.g. performance).

casperdcl:
Also #2865 says

> - and update docs

for the file streaming bit... What docs?

shcheklein:
@casperdcl yeah, this one doesn't solve the API streaming. It is the next level of avoiding writing stuff to the disk - `dcv.api.open` and `dvc get`. E.g. for S3 they call `get_signed_url` internally, instead of `_download`.

dvc/remote/gdrive.py
```diff
@@ -396,14 +385,14 @@ def _gdrive_download_file(
...
-        bar_format = (
-            "Downloading {desc:{ncols_desc}.{ncols_desc}}... "
-            + Tqdm.format_sizeof(int(gdrive_file["fileSize"]), "B", 1024)
-        )
+
         with Tqdm(
-            bar_format=bar_format, desc=progress_desc, disable=no_progress_bar
-        ):
-            gdrive_file.GetContentFile(to_file)
+            total=int(gdrive_file["fileSize"]),
```
shcheklein, line 390: does it make a meta call? can we avoid that? API calls are very expensive.

casperdcl:
> @casperdcl yeah, this one doesn't solve the API streaming. It is the next level of avoiding writing stuff to the disk - `dcv.api.open` and `dvc get`. E.g. for S3 they call `get_signed_url` internally, instead of `_download`.

ah I see

dvc/remote/gdrive.py
casperdcl: good spot. This actually isn't needed because the callback will eventually set/update the `total`.

shcheklein:
Patches like this:

Screen Shot 2020-05-05 at 3 31 49 PM

+ less (?) API calls + UI improved + download faster ...

just warm my heart :) Less code we have better I sleep. Thanks 🙏

casperdcl:
yes just pushed a final quick update (because of the deferred `total`) to ensure consistent formatting so a tiny bit more loc :)

casperdcl:
also please don't merge until https://github.com/conda-forge/pydrive2-feedstock/pull/10 (which doesn't exist yet)

casperdcl:
ok good to go

efiop:
Thank you @casperdcl ! :pray:

PR #3728, opened 2020-05-03
[WIP] dvc.yaml: rename stages -> run

efiop:
`Run` is a more natural terminology for data scientists than `stages` so
it makes sense to rename it so it is more in-sync with it as well as our
commands (e.g. `dvc run`).

### DIFF
--- a/dvc/dvcfile.py
+++ b/dvc/dvcfile.py
@@ -197,13 +197,13 @@ class PipelineFile(FileMixin):
         else:
             open(self.path, "w+").close()
 
-        data["stages"] = data.get("stages", {})
+        data["pipeline"] = data.get("pipeline", {})
         stage_data = serialize.to_pipeline_file(stage)
-        if data["stages"].get(stage.name):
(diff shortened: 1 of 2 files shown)


dmpetrov:
It is great idea to get rid of "stages", @efiop!
As Ivan said this might be confusion between a stage definition and a stage execution/instance. It might be better to use the term `pipeline` or `stages`. `Pipeline` is the best term I think.

efiop:
@dmpetrov `pipeline` or `pipelines`? :slightly_smiling_face:

efiop:
Changed the PR to use `pipeline`, but I no longer like it myself. `run` made more sense to me. But at the same time I don't have an extremely strong opinion about it, as long as it makes sense to our target audience.

skshetry:
@efiop, `pipeline` is not technically correct term to use here. Same with `run`. I think, `stages` is better than `pipeline(s)` or `run`, unless we have good alternatives.

efiop:
Ok, agreed not to rename it. `stages` is the best one.

PR #3730, opened 2020-05-04
ui: simplify/shorten error messages

### LINKED ISSUE #3706: run: ui issue for `run`
1. Eg:
```sh
$ dvc run -n train ...
ERROR: failed to run command - Stage with name 'train' already exists in 'pipelines.yaml'.
```
There's no need for `failed to run command` because the commands were never run and is a bit unnecessary.

2. 
```sh
$ dvc run -n train1 -d users.csv -p tr_conf.yaml:epochs,log_file,dropout -M summary.json -m logs.csv -o logs python train.py
ERROR: failed to run command - file/directory 'summary.json' is specified as an output in more than one stage:
	pipelines.yaml:train
This is not allowed. Consider using a different output name.
```

Suggestion:
```
ERROR: output 'summary.json' already exist on stage 'pipelines.yaml:train'.
```

3. 
```sh
$ dvc repro pipelines.yaml
ERROR: you've introduced a cycle in your pipeline that involves the following stages:
	- pipelines.yaml:train
	- pipelines.yaml:process
```

Suggestion:
```sh
ERROR: output 'IN.TSV' introduces a cycle that involves stages: train, process
```

### LINKED ISSUE #3718: ERROR: unexpected error - are any errors expected by the user?
The following example is to drive my point home, but, yeah, the `unexpected error` is unnecessary.
```sh
$ cd  `mktemp -d`
$ mkdir .dvc
$ dvc status
ERROR: unexpected error - expected str, bytes or os.PathLike object, not NoneType

```

skshetry:

closes #3706 and closes #3718

### DIFF
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -1,10 +1,15 @@
 """Exceptions raised by the dvc."""
+from funcy import first
 
 from dvc.utils import relpath, format_link
 
 
 class DvcException(Exception):
     """Base class for all dvc exceptions."""
+
+    def __init__(self, msg, *args):
+        assert msg
+        super().__init__(msg, *args)
 
 
 class InvalidArgumentError(ValueError, DvcException):
@@ -23,13 +28,15 @@ class OutputDuplicationError(DvcException):
     def __init__(self, output, stages):
         assert isinstance(output, str)
         assert all(hasattr(stage, "relpath") for stage in stages)
-        msg = (
-            "file/directory '{}' is specified as an output in more than one "
-            "stage: \n{}\n"
-            "This is not allowed. Consider using a different output name."
-        ).format(
-            output, "\n".join("\t{}".format(s.addressing) for s in stages)
-        )
+        if len(stages) == 1:
+            msg = "output '{}' is already specified on {}.".format(
+                output, first(stages)
+            )
+        else:
+            msg = "output '{}' is already specified on stages:\n{}".format(
+                output,
+                "\n".join("\t- {}".format(s.addressing) for s in stages),
+            )
         super().__init__(msg)
         self.stages = stages
         self.output = output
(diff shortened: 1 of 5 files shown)


dvc/main.py
```diff
@@ -58,6 +58,9 @@ def main(argv=None):
         ret = 253
     except DvcParserError:
         ret = 254
+    except DvcException:
+        ret = 255
+        logger.exception("")
```
skshetry, line 63: DvcException won't have `unexpected error` prepended. Other exceptions and assertion errors will still have that prefixed, which is a good compromise, I'd say.

dvc/stage/exceptions.py
```diff
@@ -2,8 +2,12 @@
 
 
 class StageCmdFailedError(DvcException):
-    def __init__(self, stage):
-        msg = "{} cmd '{}' failed".format(stage, stage.cmd)
+    def __init__(self, stage, status=None):
+        period = "."
```
efiop, line 6: What's the point? :slightly_smiling_face:

dvc/exceptions.py
```diff
@@ -23,13 +28,15 @@ class OutputDuplicationError(DvcException):
...
         assert all(hasattr(stage, "relpath") for stage in stages)
-        msg = (
-            "file/directory '{}' is specified as an output in more than one "
-            "stage: \n{}\n"
-            "This is not allowed. Consider using a different output name."
-        ).format(
-            output, "\n".join("\t{}".format(s.addressing) for s in stages)
-        )
+        if len(stages) == 1:
+            msg = "output '{}' is already specified on {}.".format(
```
efiop, line 32: Maybe `in` instead of `on`?

efiop: Also stage name should be wrapped in `'` to mark that it is generated. Same In messages below.

skshetry: we have `stage: 'stage name'` as `__str__`. So, no need to wrap here. :smile:

PR #3731, opened 2020-05-04
Make comparisons in is_cached independent of order

skshetry:
On is_cached, we used to compare two dicts: one of current stage in memory
created by `run` and one that's already written to the file.
As `outs` and `dicts` are lists (which is generate by `dumpd`), the
comparison was dependent on the order of outs and deps.

So, the `run` before and now, can be in different order and would fail the comparisons.

Also, given that we can compare without order, we can make pipeline files sorted by outs and deps.

Plus, in the post-pipeline-file world, we don't even need `is_cached`. We could just throw
`DuplicateStage` or `OutputDuplicationError`. It's convenient of course, to be able to `dvc run`
same command many times but, I'm not sure if it's worth the hassle of having `is_cached` for just that.

### DIFF
--- a/dvc/serialize.py
+++ b/dvc/serialize.py
@@ -23,7 +23,7 @@ sort_by_path = partial(sorted, key=attrgetter("def_path"))
 
 def _get_outs(stage: "PipelineStage"):
     outs_bucket = {}
-    for o in stage.outs:
+    for o in sort_by_path(stage.outs):
         bucket_key = ["metrics"] if o.metric else ["outs"]
 
         if not o.metric and o.persist:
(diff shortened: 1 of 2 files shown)


skshetry:
@efiop, we already have a test for pipeline stage that should cover this (`tests.func.test_run_multistage.test_multi_stage_run_cached`.

The external behavior that needs to be tested is following: 
```sh
dvc run  -d foo0 -d foo1 -o bar0 -o bar1 "cp foo0 bar0 && cp foo1 bar1"
dvc run  -d foo1 -d foo0 -o bar0 -o bar1 "cp foo0 bar0 && cp foo1 bar1"
```

Previous behavior considered these to be different, and would ask for overwriting the file `bar0.dvc` it would output. Now, it'll be considered as cached.

PR #3733, opened 2020-05-04
plot: completion for zsh

pared:


dvc/command/plot.py
```diff
@@ -96,7 +96,7 @@ def _revisions(self):
 
 def add_parser(subparsers, parent_parser):
     PLOT_HELP = (
-        "Generating plots for continuous metrics stored in structured files "
```
pared, line 99: Since `dvc plot` can be used to generate plot for non-continuous metrics too (like confusion matrix) It seems to me that this word should not be included in description of the plot command, because it can confuse users.

PR #3741, opened 2020-05-04
Snap version

casperdcl:
- [x] fix `snapcraft` on travis CI missing version tags

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -106,6 +106,7 @@ jobs:
     - sudo /snap/bin/lxd.migrate -yes
     - sudo /snap/bin/lxd waitready
     - sudo /snap/bin/lxd init --auto
+    - git fetch --tags
     script:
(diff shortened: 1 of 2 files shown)


casperdcl:
@efiop looks like the snap builds started working again miraculously but probably safest to still merge this PR in case it happens again.

.travis.yml
```diff
@@ -106,6 +106,7 @@ jobs:
     - sudo /snap/bin/lxd.migrate -yes
     - sudo /snap/bin/lxd waitready
     - sudo /snap/bin/lxd init --auto
+    - git fetch --tags
```
efiop, line 109: snap was failing on master and PR builds. Not sure how this is related.

casperdcl: I think travis was failing in some cases to pull tags, so `git describe` was failing, meaning `snapcraft` couldn't get a version.

Issue #3416, opened 2020-02-27
Add parallelism to checksum calculation for pulling multiple large files


Ykid:
>  when checking out, we do it sequentially, stage after stage, that is why checksums calculations are sequential

It would be better if multiple stages (which might correspond to large files) can perform checksum calculation at the same time.

Ykid:
@pared

efiop:
@pared Do you remember the context? Aren't we trusting pulled checksums these days? Or is it about unpacked dir case?

pared:
@efiop @Ykid Sorry for the delay,
Here is the start of our conversation: 
https://discordapp.com/channels/485586884165107732/563406153334128681/682124179885260804

@Ykid has a repository where there are few big files (A, B, C), each having its own stage (A.dvc, B.dvc, C.dvc). @Ykid noticed that checksum calculations on `dvc pull` take a lot of time.

So, I see 2 potential problems:
1. Files were on NFS, so reading files and feeding it to hash function might be taking a lot of time.
2. We do parallelize checksum computations, but on `out` level. What I mean by that, is thath when checking out [here](https://github.com/iterative/dvc/blob/abc4e3c2d0c8d09a2f04db88e6bf0b331ca166bc/dvc/repo/checkout.py#L61) we do it stage by stage, so we will have: 
1. checkout `A.dvc`
2. checkout `B.dvc`
and so on.
And as checksum calculation happens down the road of `stage.checkout` our checksum calculations will be sequentiall (because checkouts are sequential).

pared:
@Ykid 
What version of `dvc` are you using?

Ykid:
@pared 
```
DVC version: 0.86.5
Python version: 3.7.6
Platform: Linux-5.0.0-37-generic-x86_64-with-debian-buster-sid
Binary: False
Package: pip
Cache: reflink - not supported, hardlink - supported, symlink - supported
Filesystem type (cache directory): ('ext4', '/dev/nvme0n1p2')
Filesystem type (workspace): ('ext4', '/dev/nvme0n1p2')
```

efiop:
@pared Not sure I follow, what hashes are calculated on checkout? For the files we've downloaded? But we should've trusted those and then when checking out we are trusting the cache hashes, right?

pared:
@efiop Thats right, when pulling we should trust the checksums by default.
@Ykid did you configuration changed in any way since we talked on discord?
It used to be like that:
```
[cache]
    type = symlink
    protected = true
[...]

PR #3748, opened 2020-05-06
tests: abstract remote setup to fixture

### LINKED ISSUE #2888: tests: erepo upstream should not be its cache
Setting default remote as [repository cache](https://github.com/iterative/dvc/blob/f9d4ef8abd5427b0ec75f7f3c1211abd8528aa99/tests/conftest.py#L152) inside `erepo` is kinda hacky. This is not a real use case, and some test relay on the `erepo` property (that is automatically adding cached files to remote because cache and remote are the same thing). `erepo` remote should not be its cache. 

This issue will probably be obsolete after #2599. In that case, the code responsible for setting remote in `erepo` should be completely removed.

Related #2852

pared:

Fixes #2888, related to #3416 

The problem:
In #3416 I could reproduce user problems in script, but it seemed that I am unable to write a test for it. 
The problem turned out to be related to the fact that `erepo_dir` has been setting its default remote to be its cache. Due to that, I have been indirectly using cache "read only mode" optimization, while using remote. Because it is the second time i stumble upon this issue (#2888) I decided to fix it.
This issue introduces "setup_remote" fixture, which is intended to be default way to set up the remote for particular `Repo` instance.

### DIFF
--- a/tests/dir_helpers.py
+++ b/tests/dir_helpers.py
@@ -59,10 +59,10 @@ __all__ = [
     "tmp_dir",
     "scm",
     "dvc",
-    "local_remote",
     "run_copy",
     "erepo_dir",
     "git_dir",
+    "setup_remote",
 ]
 
 
@@ -257,17 +257,6 @@ def _git_init(path):
 
 
 @pytest.fixture
-def local_remote(request, tmp_dir, dvc, make_tmp_dir):
-    path = make_tmp_dir("local-remote")
-    with dvc.config.edit() as conf:
-        conf["remote"]["upstream"] = {"url": fspath(path)}
-        conf["core"]["remote"] = "upstream"
-    if "scm" in request.fixturenames:
-        tmp_dir.scm_add([dvc.config.files["repo"]], commit="add remote")
-    return path
-
-
-@pytest.fixture
 def run_copy(tmp_dir, dvc):
     tmp_dir.gen(
         "copy.py",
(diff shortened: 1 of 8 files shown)


tests/dir_helpers.py
```diff
@@ -200,6 +200,20 @@ def branch(self, name, new=False):
         finally:
             self.scm.checkout(old)
 
+    @contextmanager
+    def local_remote(self):
```
skshetry, line 204: @pared, this as contextmanager doesnot really make sense. It'd be better to have a explicit `push` on tests as well.

Also, we have a `local_remote` fixture already below.

pared: fixing it RN

pared: @skshetry, a new iteration of this issue has been pushed, I was able to remove most of the remote setups, though I am not sure whether this context manager should exist. What do you think?

skshetry: @pared, I'd love to have explicit push.
What about something like this:
```python
def test(tmp_dir, dvc, setup_remote):
# setup_remote fixture provides a function with which we can setup remote.
setup_remote(dvc, url='', name='', default=True)
# if url is not provided, create a tmp_dir and set that as a local remote
```

skshetry: It feels like the commands in `dvc remote` should really be part of `Repo()` api.

pared: @skshetry I introduced `setup_remote`

tests/dir_helpers.py
```diff
@@ -287,21 +276,30 @@ def run_copy(src, dst, **run_kwargs):
...
     return path
+
+
+@pytest.fixture
+def setup_remote(make_tmp_dir):
+    def create(repo, url="", name="", default=True):
+        if not url:
+            url = fspath(make_tmp_dir("local_remote"))
+        if not name:
+            name = "upstream"
```
skshetry, line 295:
```suggestion
    def create(repo, url=None, name="upstream", default=True):
        if not url:
            url = fspath(make_tmp_dir("local_remote"))
```

tests/dir_helpers.py
```diff
@@ -287,21 +276,30 @@ def run_copy(src, dst, **run_kwargs):
...
+        with repo.config.edit() as conf:
+            conf["remote"][name] = {"url": url}
+            if default:
+                conf["core"]["remote"] = name
+
+        repo.scm.add(repo.config.files["repo"])
+        repo.scm.commit("add repo")
```
skshetry, line 302:
```suggestion
        repo.scm.commit("add remote")
```

Maybe add a remote name as well in the commit message if you like.

PR #3756, opened 2020-05-07
[WIP] Copy file mode local push

### LINKED ISSUE #3416: Add parallelism to checksum calculation for pulling multiple large files

pared:

Merge only after #3748 
Related #3416


efiop:
Discussed privately that this is unwanted behaviour. The original issue should be solved in a different way.

@pared I suppose we could close this PR?

PR #3759, opened 2020-05-07
plot: update some output strings

jorgeorpinel:

### DIFF
--- a/dvc/command/plot.py
+++ b/dvc/command/plot.py
@@ -166,7 +166,7 @@ def add_parser(subparsers, parent_parser):
     plot_show_parser.add_argument("--xlab", default=None, help="X axis title.")
     plot_show_parser.add_argument("--ylab", default=None, help="Y axis title.")
     plot_show_parser.add_argument(
-        "datafile", nargs="?", default=None, help="Metrics file to visualize.",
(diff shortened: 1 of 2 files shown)


jorgeorpinel:
Only Py36 on Linux build failed (https://github.com/iterative/dvc/pull/3759/checks) Seems like a glitch.

dvc/command/plot.py
```diff
@@ -148,7 +148,7 @@ def add_parser(subparsers, parent_parser):
         "--stdout",
         action="store_true",
         default=False,
-        help="Print plot specification to stdout.",
+        help="Print plot location to stdout.",
```
pared, line 151: So this option lets user print the plot as a text to stdout. So, plot by default logs plot location to stdout. When using `--stdout` option, it is the plot file content that will be displayed on stdout.

jorgeorpinel: Oh I see. Got it wrong then, it seems. Let me FTFY ⏳

jorgeorpinel:
Fixed your comment, @pared. I also made a few more changes to match the now published version in https://dvc.org/doc/command-reference/plot/diff as well as a few more small upcoming changes in iterative/dvc.org/pull/1283. PLAT

jorgeorpinel:
I mean PTAL 😆

PR #3760, opened 2020-05-07
update: --recursive flag

### LINKED ISSUE #3511: update: add `-R, --recursive` flag
`dvc update -R/--recursive` should recursively find & update all DVC-files which are updateable (and be incompatible with `--rev`).

Allows replacing:
```
dvc update <long list of files> && dvc repro -P
```
with:
```
dvc update -R && dvc repro -P
```

nik123:
Fixes #3511 

> * [ ] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here. If the CLI API is changed, I have updated [tab completion scripts](https://github.com/iterative/dvc/tree/master/scripts/completion).

See https://github.com/iterative/dvc.org/issues/1259

### DIFF
--- a/dvc/repo/update.py
+++ b/dvc/repo/update.py
@@ -3,10 +3,13 @@ from . import locked
 
 
 @locked
-def update(self, targets=None, rev=None, recursive=False):
+def update(self, targets, rev=None, recursive=False):
     if not targets:
         stages = self.collect(targets, recursive=recursive)
     else:
+        if isinstance(targets, str):
+            targets = [targets]
+
         stages = set()
         for target in targets:
             stages.update(self.collect(target, recursive=recursive))
--- a/dvc/repo/update.py
+++ b/dvc/repo/update.py
@@ -5,14 +5,14 @@ from . import locked
 @locked
 def update(self, targets, rev=None, recursive=False):
     if not targets:
-        stages = self.collect(targets, recursive=recursive)
-    else:
-        if isinstance(targets, str):
-            targets = [targets]
+        targets = [None]
 
-        stages = set()
-        for target in targets:
-            stages.update(self.collect(target, recursive=recursive))
+    if isinstance(targets, str):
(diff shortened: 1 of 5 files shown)


dvc/repo/update.py
```diff
@@ -1,14 +1,20 @@
+from ..dvcfile import Dvcfile
 from . import locked
 
 
 @locked
-def update(self, target, rev=None):
-    from ..dvcfile import Dvcfile
+def update(self, targets=None, rev=None, recursive=False):
+    if not targets:
+        stages = self.collect(targets, recursive=recursive)
```
efiop, line 8: We've explicitly avoided supporting `update` without targets to avoid possible issues with user trying to update multiple files with a revision that is not actually relevant for all of them (imagine 2 imports from 2 projects and `--rev v5` not existing in one of them). So let's not introduce that for now, let's keep `targets` a required argument.

dvc/repo/update.py
```diff
@@ -1,14 +1,20 @@
...
+    else:
+        stages = set()
+        for target in targets:
```
efiop, line 11: Could also
```
if isinstance(targets, str):
    targets = [targets]     
```
just as a convenience (we have that in `dvc add` too), since it is guaranteed someone will try to misuse it :slightly_smiling_face:

nik123: By the way are the any other commands which should contain `isinstance(targets, str)` check? `repo.used_cache`, `repo.status`, `repo._checkout` seem good candidates.

efiop: @nik123 Probably not `used_cache` but the rest could use it, sure. But let's not bother with it in this PR, it is out of scope :slightly_smiling_face:

efiop:
Thanks @nik123 ! :pray:

PR #3768, opened 2020-05-08
push/pull: properly collect run cache

### LINKED ISSUE #1234: repro: use build cache for deterministic stages

efiop:
Ideally run cache should be a part of NamedCache and transferred as a regular cache through push/pull, but that requires a refactor, so for now we keep doing that manually to unblock CICD.

Fixes #1234

### DIFF
--- a/dvc/repo/fetch.py
+++ b/dvc/repo/fetch.py
@@ -34,6 +34,9 @@ def _fetch(
         config.NoRemoteError: thrown when downloading only local files and no
             remote is configured
     """
+
+    used_run_cache = self.stage_cache.pull(remote) if run_cache else []
+
     used = self.used_cache(
         targets,
         all_branches=all_branches,
@@ -44,6 +47,7 @@ def _fetch(
         remote=remote,
         jobs=jobs,
         recursive=recursive,
+        used_run_cache=used_run_cache,
     )
 
     downloaded = 0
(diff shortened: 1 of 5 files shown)


dvc/stage/cache.py
```diff
@@ -118,9 +120,36 @@ def _transfer(func, from_remote, to_remote):
...
+        from dvc.cache import NamedCache
+        from dvc.stage import create_stage, PipelineStage
+
+        cache = NamedCache()
+
+        for key, value in used_run_cache:
+            entry = self._load_cache(key, value)
+            if not entry:
+                continue
+            stage = create_stage(
```
efiop, line 145: Yep, this is hacky.

dvc/repo/fetch.py
```diff
@@ -34,6 +34,9 @@ def _fetch(
         config.NoRemoteError: thrown when downloading only local files and no
             remote is configured
     """
+
+    used_run_cache = self.stage_cache.pull(remote) if run_cache else []
```
efiop, line 38: Shame NamedCache relies on checksums so heavily(or rather push/pull itself), otherwise we could've fed stage_cache to it directly. But for now, this works fine.

PR #3778, opened 2020-05-11
run-cache: support uncached files

efiop:
Currently we only restore `--outs`, but `--outs-no-cache` are also
useful, especially when they are metrics files
(`-M|--metrics-no-cache`). So this PR makes run-cache save uncached
files to our cache and also tries to checkout them back during the
`restore` phase.

### DIFF
--- a/dvc/stage/cache.py
+++ b/dvc/stage/cache.py
@@ -1,5 +1,6 @@
 import logging
 import os
+from contextlib import contextmanager
 
 import yaml
 from funcy import first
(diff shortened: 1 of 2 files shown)


dvc/stage/cache.py
```diff
@@ -76,6 +77,45 @@ def _load(self, stage):
...
+        # stage and to workaround `commit/checkout` not working for uncached
+        # outputs.
+        cached_stage = self._create_stage(cache)
+
+        outs_no_cache = [
+            out.def_path for out in stage.outs if not out.use_cache
+        ]
+
+        # NOTE: using copy link to make it look like a git-tracked file
+        with self._cache_type_copy():
```
efiop, line 114: A bit weird that a method with such name does this implicitly. Will revisit in the morning on a fresh head.

dvc/stage/cache.py
```diff
@@ -76,6 +77,45 @@ def _load(self, stage):
 
         return None
 
+    def _create_stage(self, cache):
+        from dvc.stage import create_stage, PipelineStage
+
+        stage = create_stage(
```
efiop, line 83: Still hacky :slightly_smiling_face:

dvc/stage/cache.py
```diff
@@ -134,22 +183,13 @@ def pull(self, remote):
 
     def get_used_cache(self, used_run_cache, *args, **kwargs):
         from dvc.cache import NamedCache
-        from dvc.stage import create_stage, PipelineStage
```
skshetry, line 137: Cannot `Stage` take this responsibility of getting the `cache` entry?

efiop: Stage class is pretty overblown as it is, didn't want to put additional functionality onto it, especially since run-cache seems to work fine separately.

dvc/stage/cache.py
```diff
@@ -84,7 +124,13 @@ def save(self, stage):
         cache = to_single_stage_lockfile(stage)
         cache_value = _get_cache_hash(cache)
 
-        if self._load_cache(cache_key, cache_value):
+        existing_cache = self._load_cache(cache_key, cache_value)
+        cache = existing_cache or cache
+
+        for out in self._uncached_outs(stage, cache):
+            out.commit()
```
skshetry, line 131: `outs_no_cache` is more or less `do-not-push`, right? "outputs are not tracked by DVC" is no longer the case.

skshetry: Also, if the stage cache can be shared, but `outs_no_cache` are not being shared at all. Does this make sense at all?

efiop: Yes, we forcefully cache and share `outs_no_cache` if you use build cache. E.g. if you have a git-tracked metrics for your stage, run-cache will replace it to what it should be.

skshetry: Got it. I was confused that when I did `dvc push`, non-cached outputs was not on remote. I had forgotten that it requires `--run-cache` flag.

PR #3781, opened 2020-05-11
import: check erepo is None before modifying url - fixes #3780

### LINKED ISSUE #3780: `dvc import -o` fails  when source is location within same project
env: dvc v0.94, pip, Ubuntu

When I run the following dvc command:

    dvc import -o my/target/loc/ . some/long/path/to/dir

I get this error message: 

```
ERROR: failed to import 'some/long/path/to/dir' from '.'. - The path '../../../some/long/path/to/dir' does not exist in the target repository '.' neither as a DVC output nor\
 as a Git-tracked file
```
I tracked the issue down to [this line](https://github.com/iterative/dvc/blob/ce946d11609597f3b1c6c05e61f2701ca8f43f5c/dvc/repo/imp_url.py#L21), where a relative prefix is incorrectly being appended to the `url`.

```python
# NOTE: when user is importing something from within his own repository
if os.path.exists(url) and path_isin(os.path.abspath(url), self.root_dir):
    url = relpath(url, wdir)
```

I'll try to put together a PR

ned2:
Fixes #3780

### DIFF
--- a/dvc/repo/imp_url.py
+++ b/dvc/repo/imp_url.py
@@ -17,8 +17,12 @@ def imp_url(self, url, out=None, fname=None, erepo=None, locked=True):
     out = resolve_output(url, out)
     path, wdir, out = resolve_paths(self, out)
 
-    # NOTE: when user is importing something from within his own repository
-    if os.path.exists(url) and path_isin(os.path.abspath(url), self.root_dir):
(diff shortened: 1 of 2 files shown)


efiop:
Thank you @ned2 ! :tada:

PR #3784, opened 2020-05-12
diff: handle repo with no commits

### LINKED ISSUE #3398: diff: no need for error when there's no HEAD (new Git repo)
```console
$ dvc version
DVC version: 0.83.0
Python version: 3.7.6
Platform: Darwin-19.3.0-x86_64-i386-64bit
Binary: False
Package: brew

$ git init
$ dvc init
$ echo foo > data
$ dvc diff
ERROR: failed to get diff - unknown Git revision 'HEAD'                 

```

^ It could simply not print anything, I think.

> p.s. also not sure about the need for that `WARNING` in `dvc version`.

efiop:
`git diff` doesn't throw errors, so we could do the same.

Fixes #3398

### DIFF
--- a/dvc/repo/diff.py
+++ b/dvc/repo/diff.py
@@ -52,6 +52,9 @@ def diff(self, a_rev="HEAD", b_rev=None):
             if _exists(output)
         }
 
+    if self.scm.no_commits:
+        return {}
+
     working_tree = self.tree
     a_tree = self.scm.get_tree(a_rev)
     b_tree = self.scm.get_tree(b_rev) if b_rev else working_tree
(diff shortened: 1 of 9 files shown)


efiop:
@jorgeorpinel I'm not able to reproduce. Are you sure you've installed it correctly?

efiop:
merging for now. I think you just didn't install correctly.

jorgeorpinel:
Sounds good. You're probably right. I'll double check with the merged version ASAP.

jorgeorpinel:
Checked this finally. It's fixed in 1.0.1 but not in 0.94, which I know is normal. Just confirming. Thanks again

Issue #3777, opened 2020-05-10
1.0 (not only) testing - comments, finding, issues, etc


shcheklein:
Some notes on the playground, how to reproduce, and what I'm testing for.

Repo to reproduce: [here](https://github.com/shcheklein/example-get-started).

- I'm running testing on a new version of the `example-get-started` that I built with this updated gen [script](https://github.com/iterative/example-repos-dev/blob/1.0-migration/example-get-started/generate.sh). 

- The repo itself (that script from above generates) can be downloaded [here](https://github.com/shcheklein/example-get-started). Should be enough to reproduce issues below described below. And can be useful for everyone as an easy entry point to try DVC 1.0 on a bit more sophisticated project.

- This list below might include things not related 100% to the latest changes we introduced in 1.0, but all of them are potentially relevant. E.g., we change pipelines interface and should try scenarios around.

## `dvc repro` behavior and UI/UX

- [x] ~~I think we should assume `dvc.yaml` prefix by default. Instead of running a stage like this `dvc repro dvc.yaml:prepare`, I should be able to run it like this (if `dvc.yaml` is found in the current dir? or if it is a single one in the project?)~~: 

```
$ dvc repro prepare
```

#3842

- [x] Related to the previous one. We should probably accept an output as a target. It can solve some problems with autocomplete and improve experience. Related #3743 (especially last comments). (Extracted to #3875)

```
$ dvc repro model.pkl
```

- [x] ~~Repro excessive warning on the default target file name. Either make it info (in general I would say `WARNING` is a rare thing that we need to show), or don't show again. Instead make a message about the stage that we execute more meaningful (see below)~~.  #3822

```
$ dvc repro
WARNING: assuming default target 'dvc.yaml'.
```

- [x] ~~Import stage reusing lock mechanism leads to ugly warnings on every repro~~ (PR: #3823):

```
$ dvc repro
WARNING: stage: 'data/data.xml.dvc' is locked. Its dependencies are not going to be reproduced.
```

- [x] ~~When I run `dvc repro` I don't really see the stage name being reproduced - we can probably include it w/o within the "Running command" message. It can solve the problem of some WARNING (see above).~~ #3834

- [x] ~~When you change something and it hits cache it feels that we need to explain a bit on what's going on? show summary of changes to the workspace?~~ (Partially fixed by #3834)

## `dvc repro` options

- [x] <s> `--dry` does not work at all. </s> https://github.com/iterative/dvc/pull/3799

```
[...]

PR #3788, opened 2020-05-12
run/repro: rename "build cache" -> "run cache"

efiop:
Instead of `--ignore-build-cache` we now have `--no-run-cache` for run/repro which disables our actual run-cache and `--force-downstream` for `repro` that better describes the old behaviour.

Part of #3777

https://github.com/iterative/dvc.org/pull/1285


dvc/command/repro.py
```diff
@@ -136,7 +136,7 @@ def add_parser(subparsers, parent_parser):
         help="Reproduce all stages in the specified directory.",
     )
     repro_parser.add_argument(
-        "--ignore-build-cache",
+        "--ignore-run-cache",
         action="store_true",
         default=False,
         help="Reproduce all descendants of a changed stage even if their "
```
shcheklein, line 142: does this description stay?

shcheklein:
update zsh/bash?

update help if needed?

dvc/command/repro.py
```diff
@@ -136,7 +137,16 @@ def add_parser(subparsers, parent_parser):
         help="Reproduce all stages in the specified directory.",
     )
     repro_parser.add_argument(
-        "--ignore-build-cache",
+        "--no-run-cache",
+        action="store_true",
+        default=False,
+        help=(
+            "Run changed stage even if it has been already ran with the same "
```
efiop, line 144: This message almost directly copying the existing one in `run`, but it could clearly be improved. In my mind I would explain it through the run-cache concept, but we don't have it explained separately anywhere. Maybe we should consider that and base our other explanations on that.

If you have suggestions for this one right now, would appreciate them :slightly_smiling_face:

dvc/command/repro.py
efiop: Yes, but for `--force-downstream`. The new name reflects the functionality a lot better.

shcheklein: Yep 👍

dvc/command/repro.py
shcheklein: yeah, it sounds a bit strange "Run _stage_" while I'm actually running the whole pipeline with dvc repro usually or at least in some cases.

shcheklein:
Look good overall (from the naming perspective, was not reviewing specific impl details). Just a minor comment.

dvc/command/run.py
```diff
@@ -165,7 +165,7 @@ def add_parser(subparsers, parent_parser):
         help="Overwrite existing DVC-file without asking for confirmation.",
     )
     run_parser.add_argument(
-        "--ignore-build-cache",
+        "--no-run-cache",
         action="store_true",
         default=False,
         help="Run this stage even if it has been already ran with the same "
```
jorgeorpinel, line 171:
```suggestion
        help="Execute the command even if this stage has already been run "
        "with the same dependencies, outputs, etc.",
```
> Needs wrapping into the string below.

dvc/command/repro.py
jorgeorpinel:
```suggestion
            "Execute stage commands even if they have already been run "
            "with the same dependencies, outputs, etc.",
```

> I would explain it through the run-cache concept, but we don't have it explained separately anywhere. Maybe we should consider that

Yes but lets leave that for a separate update? We should be getting to explaining the run cache in docs soon, by this weekend I think.

jorgeorpinel: > I would explain it through the run-cache concept, but we don't have it explained separately anywhere. Maybe we should consider that

Yes but lets leave that for a separate update? We should be getting to explaining the run cache in docs soon, by this weekend I think.

PR #3792, opened 2020-05-13
Drop Python 3.5 support

### LINKED ISSUE #3765: Drop python 3.5 support?
Advantages of Python 3.6+ codebase:
1. f-strings (on exceptions mostly, but yay!)
2. insertion-ordered dict (in 3.6, it's an implementation detail, but who's gonna change that?)
3. Supports `Pathlike` objects which will help us get rid of some helpers.
4. Typings, which we are planning to start using. (with support for variable annotations)
5. We will be supporting 3 versions (right now, it's 4)
6. Per download stats, [3.8 has overtaken 3.5](https://pypistats.org/packages/dvc) and number is quite low.

Regarding 5, Numpy and most of the data-science tools follow [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html), by which they are on course to drop Python3.6 next month. Conda has already dropped support for Python3.5. 

And, with the [release of python 3.9](https://www.python.org/dev/peps/pep-0602/), there will be a new version every year (with 5 years of support). Do we need to support all actively supported Python versions?

New schedule of Python releases starting with 3.9 where new version is released annually

/cc @iterative/engineering

pmrowla:
Will close #3765. 
Docs PR iterative/dvc.org/pull/1287

Tasks:
* [x] Drop 3.5 from PyPI supported version list
* [x] Drop 3.5 CI builds
* [x] Remove `dvc.compat.fspath`/`dvc.compat.fspath_py35` in favor of `os.fspath`
* [x] Remove unneeded `fspath(path_info)` wrappers inside `os`/`os.path` function calls

Things not addressed in this PR:
* `os.path...(path_info)` calls will not be replaced with `PurePath` equivalents (i.e. `os.path.exists(path_info)` still used rather than `path_info.exists()`)


pared:
I searched a bit on `dict` thing:
https://docs.python.org/3.6/whatsnew/3.6.html#other-language-changes

```
The order-preserving aspect of this new implementation is considered an implementation detail and should not be relied upon
```
So maybe we should leave the OrderedDict in the end.

efiop:
image

PR #3793, opened 2020-05-13
stage: moving things around, refactor

skshetry:

This PR:
* Extracts some functions from `Stage` to `utils.py`.
* Change name of functions: makes name of the most of the functions to not have `_` prefix 
  (at least those that are used outside of the `Stage` class).
* `run` and `import` related functions are on `run.py` and `imports.py` respectively.
* merge some redundant functions (eg: `changed() and _changed()`).

This keeps the bugs and most of the codebases intact, just a small refactor that moves things around.

**TIP:** Look at each commits, that way it might be easy.

Apart from a few methods (eg: `status`, `changed`, etc.), I have splitted the methods from stage on
the basis of "if it modifies outs/deps/cmd". If it does, it stays, otherwise it's extracted to be a helper method.

### DIFF
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -218,6 +218,9 @@ class Stage(params.StageParams):
         if self.always_changed:
             return True
 
+        return self._changed_deps()
+
+    def _changed_deps(self):
         for dep in self.deps:
             status = dep.status()
             if status:
@@ -228,7 +231,6 @@ class Stage(params.StageParams):
                     )
                 )
                 return True
-
         return False
 
     def changed_outs(self):
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -5,16 +5,14 @@ import signal
 import string
 import subprocess
 import threading
-from itertools import chain, product
+from itertools import product
 
 from funcy import project
 
 import dvc.dependency as dependency
-import dvc.output as output
 import dvc.prompt as prompt
 from dvc.exceptions import CheckoutError, DvcException
 from dvc.utils import dict_md5, fix_env, relpath
-from dvc.utils.fs import path_isin
 
 from . import params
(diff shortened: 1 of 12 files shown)


skshetry:
A few more refactors for `run_stage`/`cmd_run`/`checkout`/`check_can_commit` will come in a separate PR as this one's already too big.

dvc/stage/__init__.py
```diff
@@ -791,12 +530,10 @@ def get_used_cache(self, *args, **kwargs):
 
 
 class PipelineStage(Stage):
-    def __init__(self, *args, name=None, meta=None, **kwargs):
+    def __init__(self, *args, name=None, **kwargs):
         super().__init__(*args, **kwargs)
         self.name = name
         self.cmd_changed = False
-        # This is how the Stage will discover any discrepancies
-        self.meta = meta or {}
```
skshetry, line 799: Note that this `meta` was never used before. The idea during the implementation was to provide metadata to the stage
on the discrepancies between lockfile and pipeline file (eg: someone added an out in pipeline file manually). But, since
we are moving away to having less warnings and messages, this _future-proofing_ is not needed.

efiop: @skshetry So `meta` is no longer supported in `dvc.yaml`?

skshetry: No, `meta` is still supported (this is about metadata regarding discrepancies that was supposed to be tracked between lockfile and dvc.yaml file.

Eg: when someone manually added 'foo' and removed 'bar' from dvc.yaml, but didn't change lockfile, we could provide message to the user.
```python
meta = {'added': ['foo'], 'removed': ['bar']}
# 'bar' was removed on stage: 'generate-foo'
# lockfile needs adjustments
```

But, I never implemented it, and from where we are moving with warning messages, they don't make much sense now.

PR #3800, opened 2020-05-14
refactor: simplify checkouts and run, get rid of check_can_commit

skshetry:

### DIFF
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -383,12 +383,17 @@ class Stage(params.StageParams):
             if entry.checksum and entry.changed_checksum()
         ]
 
+    def _changed_stage_entry(self):
+        return "'md5' of {} changed.".format(self)
+
     def changed_entries(self):
         changed_deps = self._changed_entries(self.deps)
         changed_outs = self._changed_entries(self.outs)
-        ret = []
-        self._status_stage(ret)
-        return changed_deps, changed_outs, next(iter(ret), None)
+        return (
+            changed_deps,
+            changed_outs,
+            self._changed_stage_entry() if self.changed_stage() else None,
+        )
 
     @rwlocked(write=["outs"])
     def commit(self):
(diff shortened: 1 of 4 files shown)


dvc/stage/__init__.py
```diff
@@ -568,5 +554,8 @@ def _status_stage(self, ret):
 
     def changed_stage(self, warn=False):
```
efiop, line 555: But this should not only check for cmd but also if some deps/outs were added or deleted, no?

skshetry: Yes, the naming is a bit confusing.

skshetry: The naming is this way because `PipelineStage` need to check for `cmd` and this `Stage` checks for `md5`.

dvc/stage/run.py
```diff
@@ -76,32 +78,34 @@ def cmd_run(stage, *args, **kwargs):
...
+def run_stage(stage, dry=False, force=False, run_cache=False):
+    if not dry:
+        stage_cached = not force and _is_cached(stage)
+        if stage_cached or _try_with_run_cache(stage, force, run_cache):
+            logger.info("Stage is cached, skipping.")
+            stage.checkout()
+            return
+
+    logger.info("Running command:\n\t%s", stage.cmd)
+    if not dry:
```
efiop, line 110: Neat!

dvc/stage/run.py
```diff
@@ -76,32 +78,34 @@ def cmd_run(stage, *args, **kwargs):
...
-    if stage_cached:
-        logger.info("Stage is cached, skipping.")
-        stage.checkout()
-    else:
-        logger.info("Running command:\n\t{}".format(stage.cmd))
+
+
+def _try_with_run_cache(stage, force, run_cache):
+    stage_cache = stage.repo.stage_cache
+    stage.save_deps()
```
efiop, line 91: Actually we can move this to `run_stage`, since it is not only needed for the run-cache to get a key, but also for regular `run` to save deps early, verifying that they exist.

skshetry: `already_cached()` won't find out if something changed that way.

dvc/stage/__init__.py
```diff
@@ -250,7 +251,7 @@ def changed_outs(self):
...
         changed = self.md5 != self.compute_md5()
         if changed and warn:
-            logger.debug("DVC-file '{}' changed.".format(self.relpath))
+            logger.debug(self._changed_stage_entry())
         return changed
```
skshetry, line 255: As we have this on debug, we can make it as `md5 of stage: foo.dvc changed.`.

PR #3804, opened 2020-05-14
dvc.yaml: make persist/cache bool flags in outs/metrics

efiop:
More explicit CLI-like options `metrics/persist_no_cache` are still supported,
but we use flags when generating dvc.yaml.

Part of #3409, prerequisite for `plot`.

### DIFF
--- a/dvc/schema.py
+++ b/dvc/schema.py
@@ -2,7 +2,8 @@ from voluptuous import Any, Optional, Required, Schema
 
 from dvc import dependency, output
 from dvc.output import CHECKSUMS_SCHEMA
-from dvc.stage.params import OutputParams, StageParams
+from dvc.output.base import BaseOutput
+from dvc.stage.params import StageParams
 
 STAGES = "stages"
 SINGLE_STAGE_SCHEMA = {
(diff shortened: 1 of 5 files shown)


dvc/stage/params.py
```diff
@@ -12,11 +9,9 @@ class StageParams:
...
 
-
-class OutputParams(Enum):
-    PERSIST = "outs_persist"
-    PERSIST_NO_CACHE = "outs_persist_no_cache"
-    METRICS_NO_CACHE = "metrics_no_cache"
-    METRICS = "metrics"
-    NO_CACHE = "outs_no_cache"
-    OUTS = "outs"
+    PARAM_OUTS_PERSIST = "outs_persist"
```
efiop, line 12: Btw, might be worth renaming it to `persist` instead of `outs_persist`. Looks a lot nicer that way.

dvc/schema.py
```diff
@@ -34,7 +35,24 @@
...
+                str,
+                {
+                    str: {
+                        Optional(BaseOutput.PARAM_CACHE): bool,
+                        Optional(BaseOutput.PARAM_PERSIST): bool,
+                    }
+                },
+            )
+        ],
+        Optional(StageParams.PARAM_OUTS_NO_CACHE): [str],
```
efiop, line 49: Being a bit restrictive here and not allowing additional flags (e.g. persist=True) for such fields.

dvc/schema.py
```diff
@@ -34,7 +35,24 @@
...
+        Optional(StageParams.PARAM_METRICS): [
+            Any(str, {str: {Optional(BaseOutput.PARAM_CACHE): bool}})
```
efiop, line 51: Not supporting persist for metrics because it is the way we currently generate it during the serialization. Plus we don't really have a CLI option for it. But seems like `persist`ent metrics does make sense, especially when we talk about plots that might accumulate data from run to run. But let's wait for someone to ask for it.

Note that in old dvc-files that(persistent metrics) was supported if user set `metric: True, persist=True` by hand.

dvc/serialize.py
```diff
@@ -24,14 +24,17 @@
...
     outs_bucket = {}
     for o in sort_by_path(stage.outs):
-        bucket_key = ["metrics"] if o.metric else ["outs"]
+        key = "metrics" if o.metric else "outs"
 
-        if not o.metric and o.persist:
-            bucket_key += ["persist"]
+        extra = {}
+        if o.persist:
+            extra["persist"] = True
```
efiop, line 31: Just matching existing style in this method, so using plain strings. Maybe we should discuss strings vs constants one more time, to see how people feel about it these days.

dvc/stage/params.py
shcheklein: may be in this case we can make it a flag within the `outs` item?

efiop: @shcheklein It is that already :) But we also keep `plain` format where it mimics CLI options. So I wonder if we should make it `persist` and `--persist` in `dvc run`.

shcheklein: yep, sounds good to me! never liked these JAVA-style - `--outs-enterprisy-long-name-include-everything`  :)

efiop:
Aaand I wondered why there is no `cache: True` even though I thought that we've merged it :D Closing in favor of https://github.com/iterative/dvc/pull/3785

PR #3805, opened 2020-05-15
pyupgrade: 3.6+

skshetry:

#### Script to reproduce
```sh
$ pip install autoflake https://github.com/skshetry/pyupgrade/archive/revert-mock-rewrite.zip
$ git ls-files {dvc,tests}"/*.py" | parallel 'pyupgrade --py36-plus {}; black {}; isort {};'
$ autoflake --in-place dvc/remote/gs.py  # remove `io` from imports in `dvc/remote/gs.py`
$ flake8 dvc tests
```
Note: I have reverted auto-mock rewrite on the above custom `pyupgrade`. `pyupgrade` converts
`mock` to `unittest.mock`, but `mock` has backports from `3.8` that makes it backward incompatible.

This PR:
- removes `object` inheritance from classes
- uses f-string instead of format strings
- remove redundant `mode="r"` from open
- convert `io.open` to `open`. `open()` is an alias for `io.open`.
- s/IOError/OSError
- remove `io` imports from `dvc/remote/gs.py`.
- remove encoding in top-level of `.py` files
- `super(Class, self)` -> `super()`

### DIFF
--- a/dvc/analytics.py
+++ b/dvc/analytics.py
@@ -72,7 +72,7 @@ def send(path):
     url = "https://analytics.dvc.org"
     headers = {"content-type": "application/json"}
 
-    with open(path, "r") as fobj:
+    with open(path) as fobj:
         report = json.load(fobj)
 
     report.update(_runtime_info())
@@ -158,7 +158,7 @@ def _find_or_create_user_id():
     try:
         with Lock(lockfile):
             try:
-                with open(fname, "r") as fobj:
+                with open(fname) as fobj:
                     user_id = json.load(fobj)["user_id"]
 
             except (FileNotFoundError, ValueError, KeyError):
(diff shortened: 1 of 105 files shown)


tests/unit/test_analytics.py
```diff
@@ -70,7 +70,7 @@ def test_send(mock_post, tmp_path):
 
     analytics.send(str(report_file))
     assert mock_post.called
-    assert mock_post.call_args.args[0] == url
+    assert mock_post.call_args[0][0] == url
```
skshetry, line 73: We were using `mock` before which backports all features from `3.8`. `call.args` is only available on 3.8. 

On this change, it's not necessary as we can do some via `[0][0]`.

skshetry:
Let's see if the test passes. I rollbacked `tests` changes and only modifying `dvc/**/*.py` files.

`pyupgrade` autowrites `mock` to `unittest.mock`.

PR #3806, opened 2020-05-15
logger: add trace, count --verbose/--quiet

### LINKED ISSUE #2332: logger: disable state logging by default
* Use different levels of verbosity as suggested by this comment: https://github.com/iterative/dvc/issues/2332#issuecomment-515740204

* Add a `TRACE` level to the logging module

* Change `logger.debug(x)` -> `logger.log(TRACE, x)` 

Related: https://github.com/iterative/dvc/issues/2329

casperdcl:
- [x] count `-q`/`-v` CLI flags to determine verbosity
- [x] add `trace`/`TRACE` logging level (more verbose than `debug`)
- [x] move some (e.g. database/state cmd) `debug` logging to `trace`
- [ ] add tests
  - [x] fix/update tests
- fixes #2332
- related (maybe fixes?) #2329

> [...] if you are convinced that you need custom levels, great care should be exercised when doing this, and it is possibly *a very bad idea to define custom levels if you are developing a library* [...] if multiple library authors all define their own custom levels [...]

-- https://docs.python.org/3/howto/logging.html#custom-levels

> We already do `disable_other_loggers()` so muhuhu hahaha.

-- @casperdcl

```bash
# before
dvc (logging-counts)$ dvc status -v | wc -l
34
# after moving state commands to trace (`-vv`)
dvc (logging-counts)$ python -m dvc status -v | wc -l
20
```

### DIFF
--- a/dvc/logger.py
+++ b/dvc/logger.py
@@ -7,7 +7,6 @@ import colorama
 
 from dvc.progress import Tqdm
 
-TRACE = 5  # logging level
 FOOTER = (
     "\n{yellow}Having any troubles?{nc}"
     " Hit us up at {blue}https://dvc.org/support{nc},"
@@ -17,6 +16,33 @@ FOOTER = (
     nc=colorama.Fore.RESET,
     yellow=colorama.Fore.YELLOW,
 )
+
+
+def addLoggingLevel(levelName, levelNum, methodName=None):
+    """
+    Adds a new logging level to the `logging` module and the
+    currently configured logging class.
+
+    Based on https://stackoverflow.com/questions/2183233
+    """
+    if methodName is None:
+        methodName = levelName.lower()
+
+    assert not hasattr(logging, levelName)
+    assert not hasattr(logging, methodName)
+    assert not hasattr(logging.getLoggerClass(), methodName)
+
+    def logForLevel(self, message, *args, **kwargs):
+        if self.isEnabledFor(levelNum):
+            self._log(levelNum, message, args, **kwargs)
+
+    def logToRoot(message, *args, **kwargs):
+        logging.log(levelNum, message, *args, **kwargs)
+
+    logging.addLevelName(levelNum, levelName)
+    setattr(logging, levelName, levelNum)
+    setattr(logging.getLoggerClass(), methodName, logForLevel)
+    setattr(logging, methodName, logToRoot)
 
 
 class LoggingException(Exception):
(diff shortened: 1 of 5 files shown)


dvc/main.py
```diff
@@ -38,11 +37,16 @@ def main(argv=None):
...
+        logger.setLevel(
+            {
+                -2: logging.FATAL,
+                -1: logging.ERROR,
+                0: logging.INFO,
+                1: logging.DEBUG,
+                2: logging.TRACE,
+            }[max(-2, min(args.verbose - args.quiet, 2))]
+        )
+        logger.trace(args)
```
casperdcl, line 49: this pretty much summarises this PR

dvc/main.py
```diff
@@ -38,11 +37,16 @@ def main(argv=None):
...
+        logger.setLevel(
+            {
+                -2: logging.CRITICAL,
+                -1: logging.ERROR,
+                0: logging.INFO,
+                1: logging.DEBUG,
+                2: logging.TRACE,
+            }[max(-2, min(args.verbose - args.quiet, 2))]
+        )
+        logger.trace(args)
```
casperdcl, line 49: this pretty much summarises this PR.

dvc/logger.py
```diff
@@ -47,6 +74,7 @@ class ColorFormatter(logging.Formatter):
     """
 
     color_code = {
+        "TRACE": colorama.Fore.GREEN,
         "DEBUG": colorama.Fore.BLUE,
         "WARNING": colorama.Fore.YELLOW,
         "ERROR": colorama.Fore.RED,
```
casperdcl, line 80: not particularly fussed about colours - feel free to suggest others

dvc/logger.py
```diff
@@ -175,6 +208,13 @@ def setup(level=logging.INFO):
                     "stream": "ext://sys.stdout",
                     "filters": ["exclude_info"],
                 },
+                "console_trace": {
+                    "class": "dvc.logger.LoggerHandler",
+                    "level": "TRACE",
+                    "formatter": "color",
+                    "stream": "ext://sys.stdout",
+                    "filters": ["exclude_info"],
```
casperdcl, line 216: does this need an `exclude_debug`?

casperdcl: yes, yes it does. fixed.

skshetry:
> We already do disable_other_loggers() so muhuhu hahaha.

We only disable loggers on cli commands.

casperdcl:
We could still merge this and if/when a user takes issue with this we can look into work-arounds. Really using DVC API + another package which adds a logging level other than the fairly common `TRACE=5` is probably very unlikely.

PR #3807, opened 2020-05-17
plots: add plot markers to DVC files

efiop:


dvc/repo/plots/show.py
```diff
@@ -0,0 +1,205 @@
...
+        for rev in repo.brancher(revs=revs):
+            plots.update(_collect_plots(repo))
+
+    if not plots:
+        datafile, plot = _show(repo, datafile=None, revs=revs, **kwargs)
+        return {datafile: plot}
+
+    ret = {}
+    for plot in plots:
+        ret[plot] = _show(repo, datafile=plot, revs=revs, **kwargs)[1]
```
efiop, line 203: need to catch errors and plot whatever we can.

dvc/stage/params.py
```diff
@@ -11,12 +11,16 @@ class StageParams:
     PARAM_META = "meta"
     PARAM_ALWAYS_CHANGED = "always_changed"
     PARAM_PARAMS = "params"
+    PARAM_METRICS = "metrics"
+    PARAM_PLOTS = "plots"
 
 
 class OutputParams(Enum):
```
skshetry, line 18: Still needed?

dvc/serialize.py
```diff
@@ -92,7 +91,20 @@ def to_pipeline_file(stage: "PipelineStage"):
...
+            ),
+        ),
+        (
+            stage.PARAM_METRICS,
+            _get_outs([out for out in stage.outs if out.metric]),
+        ),
+        (
+            stage.PARAM_PLOTS,
+            _get_outs([out for out in stage.outs if out.plot]),
+        ),
```
skshetry, line 107: Let's do this in a single loop in `_get_outs()`. What do you say?

dvc/stage/params.py
efiop: OutputParams? Not really, I just didn't touch them in this PR for now.

dvc/command/run.py
```diff
@@ -140,6 +144,20 @@ def add_parser(subparsers, parent_parser):
...
+        action="append",
+        default=[],
+        help="Declare output plot file.",
+        metavar="<path>",
+    )
+    run_parser.add_argument(
+        "--plots-no-cache",
+        action="append",
+        default=[],
+        help="Declare output plot file (do not put into DVC cache).",
```
efiop, line 158: Just copying help from the options above. Clearly this is not great.

dvc/serialize.py
```diff
@@ -35,10 +36,10 @@ def _get_out(out):
     res = OrderedDict()
     if not out.use_cache:
         res[PARAM_CACHE] = False
-    if out.metric:
-        res[PARAM_METRIC] = True
-    elif out.persist:
```
efiop, line 40: These are not really mutually exclusive, but our current schema for dvc.yaml doesn't support it. Plus our old dvcfiles were fine with it, so maybe worth just allowing it. Will revisit this line in this PR.

tests/func/plots/test_plots.py
```diff
@@ -27,10 +27,14 @@ def _remove_whitespace(value):
     return value.replace(" ", "").replace("\n", "")
 
 
-def _run_with_metric(tmp_dir, metric_filename, commit=None, tag=None):
-    tmp_dir.dvc.run(metrics_no_cache=[metric_filename], single_stage=True)
```
efiop, line 31: Will need to revisit this helper at some point. It was (and is) a bit hackish.

Suor:
Do we plan to add default metric in a plot there? Current "strategy" of selecting a last key doesn't look well.

efiop:
@Suor Could you elaborate?

Suor:
@efiop plot data file contains several plots, i.e. values of several metrics over something. [Here](https://github.com/iterative/johnnie-metrics) it's `plots.csv` with several columns for example. When `plot diff` shows the plot it selects, which column to use by selecting last column. This is not a good heuristic most time.

I think we should have a way to set default metric to build plot from, probably withing a dvc file.

pared:
@Suor 
We discussed with @efiop that for starters, we should let the user point in stage whether the given metric is plot or not. You are right that the particular parameters should be included, but we should probably do that as next "stage" of this issue.

Suor:
Adding to the above, there is a `-s`, `--select` option to do that on command line. That one accepts either field name or jsonpath.

Suor:
@pared knowing default metric to select from plot data would be very valuable to the corresponding viewer feature. Now there is no interface to change anything, so default matters, it will matter even when some interface will be added. Will also improve cli experience.

pared:
@Suor to not extend this PR scope, maybe we should create p0 for that?

Suor:
@pared I have no preference on where it should go. Let @efiop decide whether he wants to bundle or separate it.

Suor:
Another issue I see is that default metric (`y` axis) is thing across commits, but the `dvc` is stored within commit. So where do we get that? From master, some of the commits involved?

dvc/command/plots.py
```diff
@@ -44,10 +43,9 @@ def run(self):
             else:
                 fields = set(self.args.select.split(","))
         try:
-            plot_string = self.repo.plot(
+            plot_string = self._func(
```
pared, line 46: It is no longer a string, right?

dvc/output/local.py
```diff
@@ -71,17 +71,18 @@ def dumpd(self):
         return ret
 
     def verify_metric(self):
-        if not self.metric:
+        if not self.metric or self.plot:
```
skshetry, line 74: Is this correct? Or, should it be `not self.plot`?

dvc/serialize.py
```diff
@@ -35,10 +36,10 @@ def _get_out(out):
...
+    if out.persist:
         res[PARAM_PERSIST] = True
+    if out.plot and isinstance(out.plot, dict):
+        res.update(out.plot)
```
skshetry, line 42: Can you please point me where this is being set as dict? I couldn't find it. :confused:

dvc/output/local.py
efiop: It is correct. Checks for plots don't differ from metric.

dvc/serialize.py
efiop: @skshetry It is actually not set anywhere programmatically (we might introduce `dvc plots modify` later, but for now it should be set by-hand).

efiop:
@Suor Great questions! For now going only with `templates`, but we'll definitely introduce additional options for it (like labels and stuff) later.

efiop:
Ok, merging for now. Will have a followup PR.

Issue #3678, opened 2020-04-24
add/run: improve "already tracked" message


efiop:
As noted by @ammarasmro

https://github.com/iterative/dvc/issues/3561#issuecomment-619119527

our `ERROR: output 'models' is already tracked by SCM (e.g. Git)` could be improved.

Looks like we could recommend
```
git rm -r --cached path
git commit -m "stop tracking path"
```
as a solution. If the file/dir was accidentally added to git and is really big, then we could also recommend https://rtyley.github.io/bfg-repo-cleaner/ or something like that to remove it from the history.

efiop:
Let's not recommend bfg for now, as it is too specific.

PR #3808, opened 2020-05-17
(OutputAlreadyTrackedError): Add git suggestions to stop tracking files

PuneethaPai:
Fixes: #3678

When file has been already tracked under git, OutputAlreadyTrackedError gets raised.
This commits add git command suggestions to help user to stop tracking file from git,
so in future commits they can use DVC for the file.

### DIFF
--- a/dvc/output/base.py
+++ b/dvc/output/base.py
@@ -31,7 +31,11 @@ class OutputIsNotFileOrDirError(DvcException):
 
 class OutputAlreadyTrackedError(DvcException):
     def __init__(self, path):
-        msg = f"output '{path}' is already tracked by SCM (e.g. Git)"
+        msg = f""" output '{path}' is already tracked by SCM (e.g. Git).
+    You can remove it from git, then add to DVC.
+        To stop tracking from git:
+            git rm -r --cached '{path}'
+            git commit -m "stop tracking {path}" """
         super().__init__(msg)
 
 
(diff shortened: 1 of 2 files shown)


efiop:
@pared Let's not ecommend bfg for now, as it is too specific. Merging and closing the original issue.

Thanks! @PuneethaPai :pray:

dvc/output/base.py
```diff
@@ -31,7 +31,11 @@ def __init__(self, path):
 
 class OutputAlreadyTrackedError(DvcException):
     def __init__(self, path):
-        msg = f"output '{path}' is already tracked by SCM (e.g. Git)"
+        msg = f""" output '{path}' is already tracked by SCM (e.g. Git).
+    You can remove it from git, then add to DVC.
+        To stop tracking from git:
```
shcheklein, line 36: sorry that it's late a bit, but it should be probably`git` -> `Git` like it is in the previous line

efiop: Fixed in https://github.com/iterative/dvc/commit/7672e324d221198a307b5b7963ff103bad2b92ab . Thanks @shcheklein ! :pray:

shcheklein:
@PuneethaPai also, it looks there is something funny happening with your Git settings - Github does not attribute commits properly to you (probably some discrepancy in emails - Github and local Git?)

PuneethaPai:
Thanks @shcheklein  and @efiop . I just realized my local email was configured with different id than the one with github login. Thanks for pointing this out. Will keep in mind in future.
PS: Big fan of this project. 🥰

efiop:
@PuneethaPai You can still add it in https://github.com/settings/emails , so github will automatically associate your existing commits with your profile.

PR #3810, opened 2020-05-18
repo: Support streaming and pulling files on `RepoTree/DvcTree.open()`

pmrowla:

* `RepoTree` now supports all `BaseTree` methods
* Adds support for fetching or streaming DVC outs directly from remote on `RepoTree`/`DvcTree.open()`.
    * If an out is already in local cache, it will always be opened from cache.
    * For outs which are not in local cache, if `DvcTree` was initialized with `fetch=True` or `stream=True`, the out will be fetched or streamed, otherwise an error will be raised
* `Repo.open_by_relpath()` now uses `RepoTree.open()`
* `RepoTree`/`DvcTree.walk()` now support walking dir contents for dir outs
    * If tree was initialized with `fetch=True` or `stream=True`, tree will pull dir cache for dir outs on `walk` when needed
    * Otherwise, dir contents will not be walked (same as existing behavior)

Related to #3811.

### DIFF
--- a/dvc/repo/tree.py
+++ b/dvc/repo/tree.py
@@ -192,7 +192,8 @@ class RepoTree(BaseTree):
                     path, mode=mode, encoding=encoding, **kwargs
                 )
             except FileNotFoundError:
-                pass
+                if self.isdvc(path):
+                    raise
         return self.repo.tree.open(path, mode=mode, encoding=encoding)
 
     def exists(self, path):
--- a/dvc/repo/tree.py
+++ b/dvc/repo/tree.py
@@ -1,14 +1,34 @@
 import errno
+import logging
 import os
 
 from dvc.exceptions import OutputNotFoundError
 from dvc.path_info import PathInfo
+from dvc.remote.base import RemoteActionNotImplemented
+from dvc.repo import Repo
 from dvc.scm.tree import BaseTree, WorkingTree
+
+logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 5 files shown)


dvc/repo/tree.py
```diff
@@ -111,6 +146,15 @@ def walk(self, top, topdown=True):
         for out in outs:
             trie[out.path_info.parts] = out
 
+            if out.is_dir_checksum and (self.fetch or self.stream):
+                # will pull dir cache if needed
```
efiop, line 150: So it doesn't pull the whole dir (i mean with the files contents), yet, right? That will be introduced later? Just making sure I follow :slightly_smiling_face:

pmrowla: Yeah, that will come w/the next PR (for saving a tree to cache). For now I just wanted to get `walk()` working properly in this PR.

dvc/repo/tree.py
```diff
@@ -111,6 +146,15 @@ def walk(self, top, topdown=True):
...
+                with self.repo.state:
+                    cache = out.collect_used_dir_cache()
```
efiop, line 152: Might be slow to collect. You only need `out.dir_cache`, so maybe it is worth creating extracting some method to fetch .dir for it into an `out` method. But I'm ok with keeping it as is for now.

efiop:
Ok, looks great. We are on the right track. Merging.

Issue #3811, opened 2020-05-18
repo: abstract erepo dvc/git handling into RepoTree


pmrowla:
Issue for keeping track of what needs to be done for the tree abstraction

main tasks:

* [x] Support fetch/stream on `open()` (#3810)
* [x] Support saving from a tree directly to cache via `walk()`+`open()` (#3825)
* [x] Rework erepo to use new tree functionality (#3841)

other tasks:

* [x] `dvc list` should use `RepoTree` (#3841)

issues that should be fixed by this: #3611, #3087

efiop:
Looks like we are done here. Feel free to reopen if i've missed anything.

PR #3819, opened 2020-05-19
utils: support use of (GitTree) tree file objects in utils

pmrowla:

Miscellaneous improvements to GitTree and utils that will be needed for cache/repo tree changes.

Related to #3811.

### DIFF
--- a/dvc/utils/fs.py
+++ b/dvc/utils/fs.py
@@ -9,7 +9,6 @@ import nanotime
 from shortuuid import uuid
 
 from dvc.exceptions import DvcException
-from dvc.scm.tree import is_working_tree
 from dvc.system import System
 from dvc.utils import dict_md5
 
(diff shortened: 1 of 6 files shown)


dvc/utils/fs.py
```diff
@@ -202,6 +198,12 @@ def copyfile(src, dest, no_progress_bar=False, name=None):
                     fdest_wrapped.write(buf)
 
 
+def copy_fobj_to_file(fsrc, dest):
```
pmrowla, line 201: I think we will probably want progress bar support for this once we are saving files from git tree to local cache, but I think it may make more sense to have the pbar instance in the higher level `save()` method.

So at that point we can add a pbar update callback parameter to this method (and do buffered read/write instead of using `shutil.copyfileobj` as needed).

PR #3825, opened 2020-05-19
remote: support saving RepoTree objects directly to cache

### LINKED ISSUE #3541: api: support granularity (files inside dirs)
Ubuntu 16.04. Virtualenv with python 3.5.2. Pip install. DVC version 0.90.2

I am unable  to retrieve a file with api.read, however what DVC is tracking is not the file itself, but the folder the file is in. The file can be retrieved sucessfully with dvc get.

This is the code snippet:

```
import dvc.api

if name == "main":

   modelpkl = dvc.api.read('relative/path/in/python/project/model.pb',
                           repo='ssh://git@repo.com/myproject.git',
                           remote='https://my-bucket.s3-us-west-2.amazonaws.com/folder_in_bucket_where_DVC_data_is, 
                           mode='rb')
```

This doesn't work either:

`dvc list ssh://git@repo.com/myproject.git relative/path/in/python/project/model.pb`

`relative/path/in/python/project/model.pb does not exist in the target repository neither as a DVC output nor as a Git-tracked file`

But this does:

`dvc get ssh://git@repo.com/myproject.git relative/path/in/python/project/model.pb`

Discord chat: https://discordapp.com/channels/485586884165107732/485596304961962003/692766090081730651

pmrowla:

Related to #3811.

added support for:

* Granular support for `RepoTree`/`DvcTree.open()` (Will close #3541)
    * Files inside DVC directory outs can be opened by any function using `RepoTree.open()` (`Repo.open_by_relpath()`, `api.open()`)
    * If file is not cached, it will only be opened if tree was initialized with `fetch=True` or `stream=True`
* RepoTree paths can be saved directly to remote (i.e. local cache)

ex:
```py
tree = RepoTree(..., fetch=True)  # stream or fetch must be set to save DVC outs properly

cache = repo.cache.local
cache.save(path_in_tree, None, tree=tree)
```

For saving tree paths, rather than requiring the call to `cache.save_info(...)` to get a destination checksum, for tree paths the checksum will be calculated during `save()`. The reason for this is that for directories, it makes more sense to collect dir cache information as we walk the tree (and save files from the tree), rather than walking the tree once to collect dir cache information, and then walking it a second time to save files.

### DIFF
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -458,27 +458,33 @@ class BaseRemote:
             "Created '%s': %s -> %s", self.cache_types[0], from_info, to_info,
         )
 
-    def _save_file(self, path_info, checksum, save_link=True):
+    def _save_file(self, path_info, checksum, save_link=True, tree=None):
         assert checksum
 
         cache_info = self.checksum_to_path_info(checksum)
-        if self.changed_cache(checksum):
-            self.move(path_info, cache_info, mode=self.CACHE_MODE)
-            self.link(cache_info, path_info)
-        elif self.iscopy(path_info) and self._cache_is_copy(path_info):
-            # Default relink procedure involves unneeded copy
-            self.unprotect(path_info)
+        if tree:
+            if self.changed_cache(checksum):
+                with tree.open(path_info, mode="rb") as fobj:
+                    self.copy_fobj(fobj, cache_info)
         else:
-            self.remove(path_info)
-            self.link(cache_info, path_info)
+            if self.changed_cache(checksum):
+                self.move(path_info, cache_info, mode=self.CACHE_MODE)
+                self.link(cache_info, path_info)
+            elif self.iscopy(path_info) and self._cache_is_copy(path_info):
+                # Default relink procedure involves unneeded copy
+                self.unprotect(path_info)
+            else:
+                self.remove(path_info)
+                self.link(cache_info, path_info)
 
-        if save_link:
-            self.state.save_link(path_info)
+            if save_link:
+                self.state.save_link(path_info)
 
         # we need to update path and cache, since in case of reflink,
         # or copy cache type moving original file results in updates on
         # next executed command, which causes md5 recalculation
-        self.state.save(path_info, checksum)
+        if not tree or is_working_tree(tree):
+            self.state.save(path_info, checksum)
         self.state.save(cache_info, checksum)
 
     def _cache_is_copy(self, path_info):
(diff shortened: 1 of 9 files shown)


dvc/remote/base.py
```diff
@@ -497,22 +508,43 @@ def _cache_is_copy(self, path_info):
...
+        # return the resulting dir checksum
+        dir_info = []
+        for fname in tree.walk_files(path_info):
+            checksum = tree.get_file_checksum(fname)
+            file_info = {
+                self.PARAM_CHECKSUM: checksum,
+                self.PARAM_RELPATH: fname.relative_to(path_info).as_posix(),
+            }
+            self._save_file(fname, checksum, tree=tree)
+            dir_info.append(file_info)
```
pmrowla, line 543: The actual behavior for this `walk` depends on how `tree` was initialized.

Git files will always be saved directly from the tree in `_save_file()` via `tree.open()`
For DVC outs:
* If `fetch=False` and `stream=False`, `walk` will not recurse into DVC out directories
* If `fetch=True`, DVC outs will be fetched when `walk` recurses into the out directory
    * `_save_file` will skip the `tree.open()`->copy for fetched outs (since they will already exist in cache)
* If `stream=True`, DVC outs will be saved by streaming in `_save_file()` via `tree.open()`

Basically, for erepo's and import/get, we probably always want the behavior from `fetch=True`.

tests/unit/repo/test_repo_tree.py
```diff
@@ -90,30 +106,31 @@ def test_walk(tmp_dir, dvc):
...
-        os.path.join("dir", "subdir2", "foo2.dvc"),
-        os.path.join("dir", "foo"),
-        os.path.join("dir", "bar"),
+        "dir/subdir1",
+        "dir/subdir2",
+        "dir/subdir1/foo1",
+        "dir/subdir1/bar1",
+        "dir/subdir2/foo2",
+        "dir/foo",
+        "dir/bar",
```
pared, line 115: I think it fails on windows

pmrowla: should be resolved now

dvc/remote/base.py
```diff
@@ -497,22 +508,43 @@ def _cache_is_copy(self, path_info):
...
         return self.cache_types[0] == "copy"
 
-    def _save_dir(self, path_info, checksum, save_link=True):
-        cache_info = self.checksum_to_path_info(checksum)
-        dir_info = self.get_dir_cache(checksum)
+    def _save_dir(self, path_info, checksum, save_link=True, tree=None):
+        if tree:
+            checksum = self._save_tree(path_info, tree)
+        else:
+            dir_info = self.get_dir_cache(checksum)
```
efiop, line 515: We've discussed this before, but just to clarify: I suppose we need this because we are not ready to make `_collect_dir` use the tree, just yet, right? Because it uses `self.walk_files`, which raises another question of abstracting tree-like Remote methods somehow (maybe in RemoteTree's or something)?

dvc/remote/base.py
```diff
@@ -541,22 +573,36 @@ def walk_files(self, path_info):
...
-    def _save(self, path_info, checksum, save_link=True):
-        to_info = self.checksum_to_path_info(checksum)
-        logger.debug("Saving '%s' to '%s'.", path_info, to_info)
-        if self.isdir(path_info):
-            self._save_dir(path_info, checksum, save_link)
+        if tree:
+            isdir = tree.isdir
+            save_link = False
+        else:
+            isdir = self.isdir
```
efiop, line 600: I like how RemoteTree(or something like that) becomes more and more apparent :smile:

And that the state should probably belong to the tree. So here it would be a noop for git tree.

Not saying we should implement all of that right now, just noticing the things that we've discussed earlier :slightly_smiling_face:

dvc/remote/base.py
```diff
@@ -452,27 +457,33 @@ def _do_link(self, from_info, to_info, link_method):
...
-            self.link(cache_info, path_info)
+            if self.changed_cache(checksum):
+                self.move(path_info, cache_info, mode=self.CACHE_MODE)
+                self.link(cache_info, path_info)
+            elif self.iscopy(path_info) and self._cache_is_copy(path_info):
+                # Default relink procedure involves unneeded copy
+                self.unprotect(path_info)
+            else:
+                self.remove(path_info)
+                self.link(cache_info, path_info)
```
efiop, line 477: Could also think about making `save()` only save without checking out(these `link`s are effectively `checkout`). This `if&else` makes it clear that this method tries to do too much. But at the same time `move` is a very nice optimization for big files that allows us to make it happen instantly if we are within the same fs.

I would be fine with keeping it as is for now, but let's at least keep this in mind.

dvc/remote/base.py
pmrowla: So we can add tree support in `_collect_dir`, but it made the workflow look a bit strange to me, since `_collect_dir` is used in `get_dir_checksum`. If you go off of the current behavior for `fetch`ing external git files, we would do something along the lines of:
```
# get a dir checksum for the git dir via `get_dir_checksum()`/`_collect_dir()`/`tree.walk_files()`
dir_cache_info = cache.save_info(path, tree=tree)
# walk the tree again and save git file objects from tree
cache.save(path, dir_cache_info, tree=tree)
```

But walking the tree twice like this seems unnecessary for this use case, since we can save objects from the tree and collect the dir cache info at the same time during a single walk. I wasn't sure if saving objects from tree during the walk would be desired behavior for `_collect_dir` (because saving objects in a `get_dir_checksum` call definitely seems like non-desired behavior).

Thinking about it now, what we could do is add the `tree` param to `_collect_dir`, and if it's called with a tree then we just save the objects from tree during `walk_files`, and use `_collect_dir` here in place of `_save_tree`. And when `_collect_dir` is used w/o a tree param, it doesn't save any objects during `walk_files` and behaves the same way as before for `get_dir_checksum` calls.

PR #3835, opened 2020-05-20
[WIP] log: use bright/dim colors for ui

skshetry:

On top of #3834.

Screenshot from 2020-05-20 19-25-30

### DIFF
--- a/dvc/dvcfile.py
+++ b/dvc/dvcfile.py
@@ -15,7 +15,7 @@ from dvc.stage.exceptions import (
     StageFileIsNotDvcFileError,
 )
 from dvc.stage.loader import SingleStageLoader, StageLoader
-from dvc.utils import relpath
+from dvc.utils import relpath, styled
 from dvc.utils.collections import apply_diff
 from dvc.utils.stage import (
     dump_stage_file,
@@ -194,7 +194,9 @@ class PipelineFile(FileMixin):
             with open(self.path) as fd:
                 data = parse_stage_for_update(fd.read(), self.path)
         else:
-            logger.info("'%s' does not exist, creating…", self.relpath)
+            logger.info(
+                "%s does not exist, creating…", styled(self.relpath, "bold")
+            )
             open(self.path, "w+").close()
 
         data["stages"] = data.get("stages", {})
@@ -206,7 +208,9 @@ class PipelineFile(FileMixin):
             data["stages"].update(stage_data)
 
         logger.info(
-            "Adding stage '%s' to '%s'…", stage.name, self.relpath,
+            "Adding stage '%s' to '%s'…",
+            styled(stage.name, "bold"),
+            styled(self.relpath, "bold"),
         )
         dump_stage_file(self.path, data)
         self.repo.scm.track_file(relpath(self.path))
@@ -252,7 +256,7 @@ class Lockfile(FileMixin):
         stage_data = serialize.to_lockfile(stage)
         if not self.exists():
             modified = True
-            logger.info("Generating lock file…")
+            logger.info(styled("Generating lock file…", "dim"))
             data = stage_data
             open(self.path, "w+").close()
         else:
@@ -262,7 +266,7 @@ class Lockfile(FileMixin):
                 stage.name, {}
             )
             if modified:
-                logger.info("Updating lock file…")
+                logger.info(styled("Updating lock file…", "dim"))
             data.update(stage_data)
         dump_stage_file(self.path, data)
         if modified:
--- a/dvc/dvcfile.py
+++ b/dvc/dvcfile.py
@@ -181,9 +181,9 @@ class PipelineFile(FileMixin):
 
         assert isinstance(stage, PipelineStage)
         check_dvc_filename(self.path)
-        self._dump_lockfile(stage)
         if update_pipeline and not stage.is_data_source:
             self._dump_pipeline_file(stage)
+        self._dump_lockfile(stage)
 
     def _dump_lockfile(self, stage):
         self._lockfile.dump(stage)
@@ -194,6 +194,7 @@ class PipelineFile(FileMixin):
             with open(self.path) as fd:
                 data = parse_stage_for_update(fd.read(), self.path)
         else:
+            logger.info("'%s' does not exist, creating…", self.relpath)
             open(self.path, "w+").close()
 
         data["stages"] = data.get("stages", {})
@@ -204,6 +205,9 @@ class PipelineFile(FileMixin):
         else:
             data["stages"].update(stage_data)
 
+        logger.info(
+            "Adding stage '%s' to '%s'…", stage.name, self.relpath,
+        )
         dump_stage_file(self.path, data)
         self.repo.scm.track_file(relpath(self.path))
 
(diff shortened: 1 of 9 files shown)


dvc/utils/__init__.py
```diff
@@ -270,6 +270,17 @@ def colorize(message, color=None):
     )
 
 
+def styled(message, style=None):
```
shcheklein, line 273: @skshetry this awesome and long waited from my end ... the only question - how well it supported by older/SSH terminals, Windows, etc - we should do some fallback to regular colors? or will this logic handle this automatically?

(also, really wanted to improve the way we print the box around telemetry, updater messages - there was some concern with older terminals as well that prevented us from using nice symbols - may be we can detect it also)

shcheklein: (I see that we only change brightness here, also as a good next step we can start using normal modern colors instead of RED, BLUE - etc that are even hard to read sometimes - again the biggest question is compatibility and fallbacks)

skshetry: It should work on Windows as it uses colorama, but not sure of others.

skshetry: I was using colors before, like I showed yesterday. But, I moved away from it, because if other log appears in between, mix of colored and non-colored output looks weird. Plus, the colors that work on both dark/light terminals need to be choosen (most probably, red/green/yellow).

shcheklein: > But, I moved away from it, because if other log appears in between, mix of colored and non-colored output looks weird.

do you mean when we use `-v`?

> I was using colors before, like I showed yesterday. 

Sorry, I was confusing. Your are right about using brightness for these logs. Everything else I mentioned is not directly related to this PR. So, please feel free to disregard :)

I was saying about using "pastel" terminal colors instead of regular RED, BLUE - e.g. https://github.com/tartley/colorama/blob/master/colorama/ansi.py#L60 - for all other logs. Same as `nmp` or `heroku cli` do - I think they can detect if terminal supports modern colors or not.

Also using some nice looking symbols for the box instead of a regular dash.

skshetry: > do you mean when we use -v?

No. I mean, as we use INFO for ui, we can either support only one color (colored or not) or manually change all to whatever color we like. But, any `logger.info` call will not be colored, so we might end up with mixed output. To have a colored output, we'd constantly fighting with it. 

> Same as nmp or heroku cli do

I did gave a quick look to https://github.com/chalk/supports-color, looks like it's possible. But, do we really need True colors (i.e. 256 colors are enough for us? At least for a start).

shcheklein: 256 is totally enough. (again, not really related to this PR, I would suggest a ticket for this)

> No. I mean, as we use INFO for ui, we can either support only one color (colored or not) or manually change all to whatever color we like. But, any logger.info call will not be colored, so we might end up with mixed output. To have a colored output, we'd constantly fighting with it.

not sure I completely follow ... but looks like another reason to not use logger for UI.

dvc/dvcfile.py
```diff
@@ -194,6 +194,9 @@ def _dump_pipeline_file(self, stage):
             with open(self.path) as fd:
                 data = parse_stage_for_update(fd.read(), self.path)
         else:
+            logger.info(
+                "%s does not exist, creating…", styled(self.relpath, "bold")
```
jorgeorpinel, line 198: Minor: Do we use `…` (ellipsis character) a lot? I'm guessing it's mostly `...` (3 period chars) in other places, if any.

jorgeorpinel: Also
```suggestion
                "'%s' does not exist, creating…", styled(self.relpath, "bold")
```

dvc/dvcfile.py
```diff
@@ -204,6 +207,11 @@ def _dump_pipeline_file(self, stage):
         else:
             data["stages"].update(stage_data)
 
+        logger.info(
+            "Adding stage '%s' to '%s'…",
```
jorgeorpinel, line 211: I see `…` is used consistently in this file 👍 just wondering about other files.

dvc/stage/__init__.py
```diff
@@ -209,10 +209,11 @@ def changed_deps(self):
             return False
 
         if self.is_callback:
-            logger.warning(
-                '{stage} is a "callback" stage '
+            logger.debug(
+                '%s is a "callback" stage '
```
jorgeorpinel, line 213:
```suggestion
                ''%s' is a "callback" stage '
```

dvc/stage/__init__.py
```diff
@@ -259,14 +260,12 @@ def changed(self):
...
-            self.changed_stage(warn=True)
+            self.changed_stage()
             or self.changed_deps()
             or self.changed_outs()
         )
         if is_changed:
-            logger.info("%s changed.", self)
-        else:
-            logger.info("%s didn't change.", self)
+            logger.debug("%s changed.", self)
```
jorgeorpinel, line 268:
```suggestion
            logger.debug("'%s' changed.", self)
```

dvc/stage/run.py
```diff
@@ -79,11 +79,16 @@ def cmd_run(stage, *args, **kwargs):
...
 def _is_cached(stage):
-    return (
+    cached = (
         not stage.is_callback
         and not stage.always_changed
         and stage.already_cached()
     )
+    if cached:
+        logger.info(
+            "Stage %s is cached, skipping…", styled(stage.addressing, "bold"),
```
jorgeorpinel, line 89:
```suggestion
            "Stage '%s' is cached, skipping…", styled(stage.addressing, "bold"),
```

dvc/stage/run.py
```diff
@@ -93,7 +98,13 @@ def restored_from_cache(stage):
         return False
     # restore stage from build cache
     stage_cache.restore(stage)
-    return stage.outs_cached()
+    restored = stage.outs_cached()
+    if restored:
+        logger.info(
+            "Restored stage %s from run-cache, skipping…",
```
jorgeorpinel, line 104:
```suggestion
            "Restored stage '%s' from run-cache, skipping…",
```
Is run-cache (hyphenated) the official term we will be using consistently?

dvc/stage/run.py
```diff
@@ -102,10 +113,17 @@ def run_stage(stage, dry=False, force=False, run_cache=False):
...
-            logger.info("Stage is cached, skipping.")
             stage.checkout()
             return
 
-    logger.info("Running command:\n\t%s", stage.cmd)
+    callback_str = (
+        "{} ".format(styled("callback", "bold")) if stage.is_callback else ""
+    )
+    logger.info(
+        "Running %s" "stage %s with command:",
```
jorgeorpinel, line 123:
```suggestion
        "Running '%s'" "stage '%s' with command:",
```

dvc/stage/run.py
```diff
@@ -102,10 +113,17 @@ def run_stage(stage, dry=False, force=False, run_cache=False):
...
+        callback_str,
+        styled(stage.addressing, "bold"),
+    )
+    logger.info("\t%s", stage.cmd)
```
jorgeorpinel, line 127: Maybe wrap in `'` also. Not sure

skshetry:
Will rework on it after https://github.com/iterative/dvc/pull/3850. Closing for now.

PR #3840, opened 2020-05-21
plots: remove --select CLI option

efiop:

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here. If the CLI API is changed, I have updated [tab completion scripts](https://github.com/iterative/dvc/tree/master/scripts/completion)."

iterative/dvc.org#1255


dvc/command/plots.py
```diff
@@ -32,21 +32,12 @@ def _func(self, *args, **kwargs):
...
-        if self.args.select:
-            if self.args.select.startswith("$"):
-                jsonpath = self.args.select
-            else:
-                fields = set(self.args.select.split(","))
         try:
             plots = self._func(
                 targets=self.args.targets,
                 template=self.args.template,
-                fields=fields,
```
efiop, line 46: Not removing the API part because it is covered by the tests and we will definitely end up using it in the nearest future. Just removing it from CLI so one tries to use it in 1.0.

dvc/command/plots.py
```diff
@@ -128,12 +119,6 @@ def add_parser(subparsers, parent_parser):
     plots_show_parser.add_argument(
         "-o", "--out", default=None, help="Destination path to save plots to.",
     )
-    plots_show_parser.add_argument(
-        "-s",
-        "--select",
-        default=None,
-        help="Choose which field(s) or JSONPath to include in the plots.",
-    )
```
pared, line 136: Same for `plots_diff_parser`

efiop: Oops, nice catch!

PR #3842, opened 2020-05-21
addressing: allow stage name addressing without ":"

skshetry:
This PR enables users to run/work with stage via their name if they are in dvc.yaml file.

#### Example
```sh
$ dvc commit stage_name
```

There is a caveat though: what if the output and the stage name clashes?
Currently, there is no way to specify it is an output, but, there is way to indicate it's a stage via:
```sh
$ dvc commit :stage_name  # or,
$ dvc commit dvc.yaml:stage_name
```

The only good workaround is:
```sh
$ dvc checkout ./output-filename-that-clashes-with-stage-name
```

### DIFF
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -328,3 +328,9 @@ class NoOutputOrStage(DvcException):
     """
     Raised when the target is neither an output nor a stage name in dvc.yaml
     """
+
+    def __init__(self, target, file):
+        super().__init__(
+            f"'{target}' "
+            f"does not exist as an output or a stage name in '{file}'"
+        )
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -324,7 +324,7 @@ class IsADirectoryError(DvcException):
     """Raised when a file operation is requested on a directory."""
 
 
-class NoOutputOrStage(DvcException):
+class NoOutputOrStageError(DvcException):
     """
     Raised when the target is neither an output nor a stage name in dvc.yaml
     """
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -322,3 +322,9 @@ class RemoteCacheRequiredError(DvcException):
 
 class IsADirectoryError(DvcException):
     """Raised when a file operation is requested on a directory."""
+
+
+class NoOutputOrStage(DvcException):
+    """
+    Raised when the target is neither an output nor a stage name in dvc.yaml
+    """
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -6,7 +6,7 @@ from functools import wraps
 from funcy import cached_property, cat, first
 
 from dvc.config import Config
-from dvc.dvcfile import PIPELINE_FILE, Dvcfile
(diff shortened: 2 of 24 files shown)


skshetry:
Looks like we'll have a problem when we have a output and stage of same name. At the moment, the priority is for stage name.

I don't want to provide priority to outputs, as it really slows things down. Currently, there's no way to indicate that the file is an output, not a stage in case of collision.

```sh
$ ls
copy-foo-bar copy-foo-bar.dvc dvc.yaml dvc.lock
$ # dvc.yaml has stage with name `copy-foo-bar`
$ dvc checkout copy-foo-bar
A    bar
```

The workaround for now is:
```sh
$ dvc checkout ./copy-foo-bar
```

cc @efiop @shcheklein

shcheklein:
> The workaround for now is:

this is a good and clear workaround to my mind!

skshetry:
> > The workaround for now is:
> 
> this is a good and clear workaround to my mind!

Clear: yes, but *workaround* nonetheless. We are making a distinction 
between `./copy-foo-bar` and `copy-foo-bar` which I don't feel good about.

But, as it should happen rarely, and in docs, we should suggest users to not
clash stage name with folder name (to make their life easier for `-R`) and
output file/folder name.

dvc/repo/checkout.py
```diff
@@ -87,9 +95,7 @@ def _checkout(
                 filter_info=filter_info,
             )
             for key, items in result.items():
-                stats[key].extend(
-                    _fspath_dir(path, self.root_dir) for path in items
-                )
+                stats[key].extend(_fspath_dir(path) for path in items)
```
skshetry, line 98: Before, the checkout summary were shown as `path_in_repo`.  This should now actually show in terms of relpath.

dvc/exceptions.py
```diff
@@ -322,3 +322,15 @@ def __init__(self, path_info):
 
 class IsADirectoryError(DvcException):
     """Raised when a file operation is requested on a directory."""
+
+
+class NoOutputOrStage(DvcException):
```
efiop, line 327:
```suggestion
class NoOutputOrStageError(DvcException):
```

dvc/repo/__init__.py
```diff
@@ -314,9 +314,9 @@ def collect_granular(
                 # should say that both stage name and file does not exist.
                 if file and is_valid_filename(file):
                     raise
-                raise NoOutputOrStage(target, exc.file)
+                raise NoOutputOrStageError(target, exc.file) from exc
             except StageNotFound as exc:
-                raise NoOutputOrStage(target, exc.file)
+                raise NoOutputOrStageError(target, exc.file) from exc
```
skshetry: We concat message of every chain of exceptions. So, when we get `CheckoutErrorSuggestGit`,
we will get following error message:
```
$ dvc repro random_name
ERROR: Did you mean `git checkout random_name`?: 'random_name' does not exist as an output or a stage name in 'dvc.yaml': "Stage 'random_name' not found inside 'dvc.yaml' file"
```
So, I left it out, as *context* is anyway set if they are in `try...except` block.

skshetry: Alternative thing to do is only concat messages from 2 exceptions.

PR #3850, opened 2020-05-22
run: raise error when command is not specified

skshetry:

Previously, `run` used to show misleading `MissingOutputError` when no command was specified.

### DIFF
--- a/dvc/repo/run.py
+++ b/dvc/repo/run.py
@@ -45,6 +45,9 @@ def run(self, fname=None, no_exec=False, single_stage=False, **kwargs):
     from dvc.stage import PipelineStage, Stage, create_stage
     from dvc.dvcfile import Dvcfile, PIPELINE_FILE
 
+    if not kwargs.get("cmd"):
+        raise InvalidArgumentError("command is not specified")
+
     stage_cls = PipelineStage
     path = PIPELINE_FILE
     stage_name = kwargs.get("name")
--- a/tests/func/metrics/conftest.py
+++ b/tests/func/metrics/conftest.py
@@ -1,0 +1,1 @@
+from tests.func.plots import run_copy_metrics  # noqa: F401
--- a/tests/func/metrics/test_diff.py
+++ b/tests/func/metrics/test_diff.py
@@ -3,11 +3,10 @@ import json
 import yaml
 
 
-def test_metrics_diff_simple(tmp_dir, scm, dvc):
+def test_metrics_diff_simple(tmp_dir, scm, dvc, run_copy_metrics):
     def _gen(val):
-        tmp_dir.gen({"m.yaml": str(val)})
-        dvc.run(cmd="", metrics=["m.yaml"], single_stage=True)
-        dvc.scm.add(["m.yaml.dvc"])
+        tmp_dir.gen({"m_temp.yaml": str(val)})
+        run_copy_metrics("m_temp.yaml", "m.yaml", metrics=["m.yaml"])
         dvc.scm.commit(str(val))
 
     _gen(1)
@@ -19,13 +18,13 @@ def test_metrics_diff_simple(tmp_dir, scm, dvc):
     assert dvc.metrics.diff(a_rev="HEAD~2") == expected
 
 
-def test_metrics_diff_yaml(tmp_dir, scm, dvc):
+def test_metrics_diff_yaml(tmp_dir, scm, dvc, run_copy_metrics):
     def _gen(val):
         metrics = {"a": {"b": {"c": val, "d": 1, "e": str(val)}}}
-        tmp_dir.gen({"m.yaml": yaml.dump(metrics)})
-        dvc.run(cmd="", metrics=["m.yaml"], single_stage=True)
-        dvc.scm.add(["m.yaml.dvc"])
-        dvc.scm.commit(str(val))
+        tmp_dir.gen({"m_temp.yaml": yaml.dump(metrics)})
+        run_copy_metrics(
+            "m_temp.yaml", "m.yaml", metrics=["m.yaml"], commit=str(val)
+        )
 
     _gen(1)
     _gen(2)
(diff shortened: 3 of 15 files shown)


dvc/repo/run.py
```diff
@@ -45,6 +45,9 @@ def run(self, fname=None, no_exec=False, single_stage=False, **kwargs):
     from dvc.stage import PipelineStage, Stage, create_stage
     from dvc.dvcfile import Dvcfile, PIPELINE_FILE
 
+    if not kwargs.get("cmd"):
+        raise InvalidArgumentError("command is not specified")
+
```
skshetry, line 50: We have a lot of tests without any `cmd`. So, a lot of tests will fail here.

efiop: Btw, I think I know why we have a lot of tests doing that type of thing: those were mainly written before we had `dvc commit`, so that was the way we're lazy :slightly_smiling_face:

tests/func/test_ls.py
```diff
@@ -29,17 +30,18 @@ def match_files(files, expected_files):
...
+        p = pathlib.Path(sys.argv[1])
+        p.parent.mkdir(exist_ok=True); p.touch()
+    """
     )
     tmp_dir.scm_gen({"script.py": script}, commit="init")
     tmp_dir.dvc_gen({"dep": "content"}, commit="init dvc")
     dvc.run(
         **{
-            "command": "python script.py",
+            "cmd": "python script.py {}".format(os.path.join("out", "file")),
```
skshetry, line 44: The command never ran before, and never worked.

tests/func/metrics/conftest.py
```diff
@@ -0,0 +1 @@
+from tests.func.plots import run_copy_metrics  # noqa: F401
```
skshetry, line 1: Created a `run_copy_metrics` fixture, that'd copy files, mostly used as copying temporarily-dumped file into metrics/plots file. Also, as `metrics/params` mostly work with tags and commits, `run_copy_metrics` can set tags and commits itself.

tests/func/metrics/test_diff.py
```diff
@@ -3,11 +3,10 @@
...
 
 
-def test_metrics_diff_simple(tmp_dir, scm, dvc):
+def test_metrics_diff_simple(tmp_dir, scm, dvc, run_copy_metrics):
     def _gen(val):
-        tmp_dir.gen({"m.yaml": str(val)})
-        dvc.run(cmd="", metrics=["m.yaml"], single_stage=True)
-        dvc.scm.add(["m.yaml.dvc"])
+        tmp_dir.gen({"m_temp.yaml": str(val)})
+        run_copy_metrics("m_temp.yaml", "m.yaml", metrics=["m.yaml"])
```
skshetry, line 9: Similar code exists on metrics/params. A temp file with suffix `_temp/_t` is created which is copied (to have some `cmd` and generate metrics/plots as in the real usage).

tests/func/test_repro.py
```diff
@@ -405,7 +407,17 @@ def test(self):
...
-            ["run", "--no-exec", "--single-stage", "-f", DVC_FILE] + deps
+            [
+                "run",
+                "--no-exec",
+                "--single-stage",
+                "-f",
+                DVC_FILE,
+                *deps,
+                "ls {}".format(
+                    " ".join(dep for i, dep in enumerate(deps) if i % 2)
```
skshetry, line 418: On above code, `deps` are saved as list with sequence of `-d, dep` items (flattened out), i just extract every second element of the list.

tests/func/test_repro.py
```diff
@@ -1465,9 +1477,11 @@ def repro_dir(tmp_dir, dvc, run_copy):
     stages["dir_file_copy"] = stage
 
     last_stage = tmp_dir / "dir" / DVC_FILE
+    deps = [os.fspath(origin_copy_2), os.fspath(dir_file_copy)]
     stage = dvc.run(
+        cmd="echo {}".format(" ".join(deps)),
```
skshetry, line 1482: For these kind of stage that does not generate any outs, I have used `echo` or `ls` as a `cmd`.

tests/func/test_run_multistage.py
```diff
@@ -299,3 +301,16 @@ def test_run_params_no_exec(tmp_dir, dvc):
...
+    "kwargs",
+    [
+        {"outs": ["foo"], "deps": ["bar"]},
+        {"outs": ["foo"], "deps": ["bar"], "name": "copy-foo-bar"},
+    ],
+)
+def test_run_without_cmd(kwargs):
+    with pytest.raises(InvalidArgumentError) as exc:
+        Repo().run(**kwargs)
+    assert "command is not specified" == str(exc.value)
```
skshetry, line 316: This is the actual test for the code change.

tests/func/test_run_single_stage.py
```diff
@@ -91,7 +91,7 @@ def test(self):
 
         with self.assertRaises(DependencyDoesNotExistError):
             self.dvc.run(
-                cmd="",
+                cmd="command",
```
skshetry, line 94: A lot of tests here were `cmd` is never run (they eventually raise Exception before running the command), I have used `command`.

tests/func/test_run_single_stage.py
```diff
@@ -204,22 +204,30 @@ def test_non_normalized_paths(self):
 class TestRunStageInsideOutput(TestDvc):
     def test_cwd(self):
         self.dvc.run(
-            cmd="", deps=[], outs=[self.DATA_DIR], single_stage=True,
+            cmd=f"mkdir {self.DATA_DIR}",
```
skshetry, line 207: There should not too many of the cases were there are no deps but only outs, but in those cases I have used `mkdir` or `touch` as `cmd`.

skshetry:
Noticed that `python` scripts running on Windows uses different python interpreter, possibly 2.7 (as, `exist_ok` on makedirs didn't exist on 2.7, but is since available from python>3.2).

efiop:
> Noticed that python scripts running on Windows uses different python interpreter, possibly 2.7, as exist_ok usage on makedirs failed on a test, it is available from python>3.2).

@skshetry Oh, probably some env issues. Let's create an issue for it please to investigate further later.

PR #3858, opened 2020-05-22
remote: local: protect uploaded files

pared:

Related to #3416


dvc/remote/local.py
```diff
@@ -299,6 +299,7 @@ def _upload(
             from_file, tmp_file, name=name, no_progress_bar=no_progress_bar
         )
         os.rename(tmp_file, to_info)
+        self.protect(to_info)
```
efiop, line 302: We should protect before renaming to make the operation atomical :slightly_smiling_face:

PR #3865, opened 2020-05-23
snap: release non-stable-semver tags to beta

### LINKED ISSUE #3860: snap: add another channel for alpha versions
Snap supports 4 channels, atm DVC is only using

- `stable`==latest tag (even if alpha), and
- `edge`==latest master commit.

Should really use

- `stable`==latest release tag,
- `candidate` or `beta`==latest alpha tag, and
- `edge`==latest master commit (as is).

I'd prefer `candidate` (`beta` could be confusing since the current tags have `a` for alpha rather than `b`/`beta`). However, based on https://snapcraft.io/docs/channels#risk-levels, maybe `beta` is best:

> **candidate**: for users who need to test updates prior to stable deployment, or those verifying whether a specific issue has been resolved.
> > Releases in candidate are considered almost ready for going into stable, but need some additional real world experimentation before they move forward. Software reaching this stage will typically have passed all available QA and review processes, since users following it expect a high stability level. Should almost never break.
>
> **beta**: for users wanting to test the latest features, typically outside of a production environment.
> > Beta is the first level towards the stabilisation of what was before a fast moving stream of changes. Specific projects may have slightly different terminology for such releases (alpha, beta, etc) but all of these are welcome on this risk level. These releases will almost certainly have passed some sort of review and QA, but may still have unfinished parts. Breaking changes are still relatively common here.

casperdcl:
- [x] push non-stable tags to the snap `channel: beta`
- [x] ensure travis CI has the correct credentials
- [x] document
  - [x] website PR for later https://github.com/iterative/dvc.org/pull/1342
- fixes #3860

Also just rolled back last `stable` to match the new scheme:

```yaml
channels:
  latest/stable:    0.94.0              2020-05-23 (388) 104MB classic
  latest/candidate: ↑
  latest/beta:      1.0.0a4             2020-05-23 (441) 105MB classic
  latest/edge:      1.0.0a4-3-g5ab948f8 2020-05-23 (446) 105MB classic
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -101,7 +101,8 @@ Snap (Snapcraft/Linux)
    snap install dvc --classic
 
 This corresponds to the latest tagged release.
-Add ``--edge`` for the latest ``master`` version.
+Add ``--beta`` for the latest tagged release candidate,
+or ``--edge`` for the latest ``master`` version.
 
 Choco (Chocolatey/Windows)
 --------------------------
--- a/scripts/ci/before_install.sh
+++ b/scripts/ci/before_install.sh
@@ -52,7 +52,11 @@ elif [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
 fi
 
 if [[ -n "$TRAVIS_TAG" ]]; then
-  echo "export SNAP_CHANNEL=stable" >>env.sh
+  if [[ $(echo "$TRAVIS_TAG" | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$') ]]; then
+    echo "export SNAP_CHANNEL=stable" >>env.sh
+  else
+    echo "export SNAP_CHANNEL=beta" >>env.sh
+  fi
 else
   echo "export SNAP_CHANNEL=edge" >>env.sh
 fi


README.rst
```diff
@@ -101,7 +101,8 @@ Snap (Snapcraft/Linux)
    snap install dvc --classic
 
 This corresponds to the latest tagged release.
-Add ``--edge`` for the latest ``master`` version.
+Add ``--beta`` for the latest tagged release candidate,
```
jorgeorpinel, line 104: Hmm I see double back quotes all over this doc, are those needed? Would be best to have consistency (all other docs I think use single ones)

casperdcl: rst unfortunately isn't md. Python doc rendering ecosystems tend to work better with rst so I wouldn't recommend converting the readme to md (even though pypi has partial support for it now).

Also requested your review just in case you needed to update something in dvc.org

jorgeorpinel: Oh I didn't realize this is a rst file. OK

> Also requested your review just in case you needed to update something in dvc.org

- [ ] For this please open an issue or PR in the dvc.org repo if changes there are needed instead. You can probably check the install section to determine this.

casperdcl: ok done https://github.com/iterative/dvc/pull/3865/ - good to merge now.

casperdcl:
done (https://github.com/iterative/dvc.org/pull/1342) - fine to merge now

README.rst
jorgeorpinel: Much appreciated! I'll take that over 🙂

PR #3867, opened 2020-05-24
optimize local status

efiop:
```
#!/bin/bash                                   
                                              
set -e                                        
set -x                                        
                                              
rm -rf test_corentin                          
mkdir test_corentin                           
cd test_corentin                              
                                              
if [ ! -d "data" ]; then                      
        mkdir data                            
        for i in $(seq 1 400000); do          
                echo $i > data/$i             
        done                                  
fi                                            
                                              
rm -rf myrepo                                 
mkdir myrepo                                  
cd myrepo                                     
                                              
git init                                      
dvc init                                      
                                              
dvc config cache.type hardlink                
dvc config core.analytics false               

echo '*.py' > .dvcignore

cp ../data -a data                            
dvc add data                                  
dvc status                                    
python -mcProfile -o status.prof -m dvc status
```

Last `dvc status` results in ~50sec(way over 1minute if there is no unpacked dir) on current master and ~8sec with this PR.

The main issue that was causing this is `relpath` that we were calling for each file that we check for match in dvcignore. Plus some alog-the-way optimizations like `os.path.join` and saving `stat`s on `is_protected`. Also removing `unpacked dir` optimization, as it is no longer needed, mainly because of the recent optimizations that are done.

The issue was reported by the user https://discordapp.com/channels/485586884165107732/563406153334128681/713338211128311808 and my reproduction script is based on the dataset structure that he has.


efiop:
Will look if I can add a benchmark like that to dvc-bench before this is merged.

dvc/ignore.py
```diff
@@ -134,6 +140,8 @@ def isexec(self, path):
         return self.tree.isexec(path)
```
efiop, line 140: Also, we've discussed this before, but all of these methods don't take `dvcignore` into account, which is wrong.

courentin:
Thank you @efiop !
I can confirm that the `dvc status` is faster now !

Here is the call graph:
speech12




Also, I noticed that the more I have entries in my `.dvcignore`, the more it takes time to complete (tiny diff though less than 2sec)

efiop:
@courentin Thanks for the feedback! Yeah, we are definitely not fully done optimizing it, but this seems like a good start :wink:

tests/func/test_ignore.py
```diff
@@ -139,7 +149,7 @@ def test_match_nested(tmp_dir, dvc):
     )
 
     remote = LocalRemote(dvc, {})
-    result = {os.fspath(f) for f in remote.walk_files(".")}
```
efiop, line 142: Don't like that I have to rework this... Looking into it...

tests/unit/test_ignore.py
```diff
@@ -69,7 +69,6 @@ def test_ignore_from_file_should_filter_dirs_and_files():
         ),
         ("dont_ignore.txt", ["dont_ignore"], False),
         ("dont_ignore.txt", ["dont*", "!dont_ignore.txt"], False),
-        ("../../../something.txt", ["**/something.txt"], False),
```
efiop, line 72: it is too expensive for `match` to try to resolve paths like this.

Issue #3869, opened 2020-05-25
optimize dvcignore


efiop:
From https://github.com/iterative/dvc/pull/3867 it is apparent that we need to bench dvcignore and research if anything else there requires optimization. A user (@courentin ) reported that something simple like
```
data/something
bash_projects
mlruns
notebooks
externals
examples
.circleci
.vscode
.pytest_cache
wandb
```
takes a significant time.

- [ ] add benchmark for this to dvc-bench project

Suor:
Looked at this code and I see several issues with code, most of which also affect performance:

1. `DvcIgnore*` classes designed around using them in `walk()`, which leads to excessively complex code handling check for full path.
2. All the complexity above is added to `CleanTree`, which makes it go into `DvcIgnore*` domain. It should really look like:
    ```python
    def isfile(self, path):
        return self.tree.isfile(path) and not self.dvcignore.match(path)
    ```
3. `DvcIgnore*` hierarchy forces sequential check against all rules and regexes. This might be optimized by constructing single structure, i.e. a big regex or a prefix tree. This is complicated by the fact that dvcignores may contain negations though.
4. `pathspec` library also checks against all regexes sequentially, which adds to 3.
5. **High level issue**. I suspect that we recheck files found via walk, so we run ignores twice. It needs to be estimated whether this is an issue.

courentin:
Thank you @Suor 

It seems consistent with what I experienced. The more I add lines in `.dvcignore`, the slowest `dvc status` is.

karajan1001:
> Looked at this and I see several issues with code, most of which also affect performance:
> 
> 1. `DvcIgnore*` classes designed around using them in `walk()`, which leads to excessively complex code handling check for full path.
> 2. All the complexity above is added to `CleanTree`, which makes it go into `DvcIgnore*` domain. It should really look like:
>    ```python
>    def isfile(self, path):
>        return self.tree.isfile(path) and not self.dvcignore.match_file(path)
>    ```
> 3. `DvcIgnore*` hierarchy forces sequential check against all rules and regexes. This might be optimized by constructing single structure, i.e. a big regex or a prefix tree. This is complicated by the fact that dvcignores may contain negations though.
> 4. `pathspec` library also checks against all regexes sequentially, which adds to 3.
> 5. **High level issue**. I suspect that we recheck files found via walk, so we run ignores twice. It needs to be estimated whether this is an issue.

@Suor 
I also have an interest in this issue and had looked into the code. In my opinion:
[...]

PR #3876, opened 2020-05-26
ignore: CleanTree should always check ignores, not just for walk()

### LINKED ISSUE #3444: forbid using subrepo as a target

After introducing support for multiple DVC roots in #3257, in case of following setup:
```
.
├── .dvc
├── .git
└── subrepo
    ├── data
    ├── data.dvc
    ├── .dvc
    └── .gitignore
```
Running command `dvc push/pull/fetch subrepo/data.dvc` will have some unexpected results.
Eg. push will push to remote stored in parent repo. 
The reason for that is that we are not resolving Repo basing on `target`, but `cwd`. 
To solve this, we should either forbid initializing `Repo` under other `Repo`, or implement dynamic resolving of target paths. 
Other commands that need to be checked for this issue:
`destroy`, `remove`, `move`, `repro`, `status`, `import`, `get`, `root`, `update`, `pipeline`.

As simply forbidding seems like easier way, we will probably need to solve this problem anyway, to support `import` and  `get`.

Related to https://github.com/iterative/dvc.org/pull/1022

pmrowla:

* `CleanTree.exists()`/`isfile()`/`isdir()`/etc now work as expected.
* For a given `path` relative to tree root, CleanTree will now check that `path` and its parents are valid and not in our DVC ignores.

Will close #3444.

### DIFF
--- a/dvc/ignore.py
+++ b/dvc/ignore.py
@@ -5,7 +5,9 @@ from funcy import cached_property
 from pathspec import PathSpec
 from pathspec.patterns import GitWildMatchPattern
 
+from dvc.path_info import PathInfo
 from dvc.scm.tree import BaseTree
+from dvc.utils import relpath
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


efiop:
@pmrowla Could you please run the test from https://github.com/iterative/dvc/pull/3867 and see how that compares with and without your patch?

There is no doubt that we need to fix these methods, but maybe we'll have to do some lowhanger optimizations too right away :slightly_frowning_face: , if the performance degrades too much. We clearly need to revisit dvcignore in general https://github.com/iterative/dvc/issues/3869, but maybe there are some simple optimizations (e.g. checking `root` before running `matches` for each file/dir) that we could do right now, if needed.

pmrowla:
I'll look into the performance impact first thing tomorrow. Also need to fix the windows only issues from CI - looks like it's a problem with the case where local cache is on a different drive.

dvc/ignore.py
```diff
@@ -125,19 +127,79 @@ def tree_root(self):
...
+        # if path is outside of tree, assume this is a local remote/local cache
+        # link/move operation where we do not need to filter ignores
+        path = relpath(path, self.tree_root)
+        if path.startswith("..") or (
+            os.name == "nt"
+            and not os.path.commonprefix(
+                [os.path.abspath(path), self.tree_root]
+            )
+        ):
+            return True
```
pmrowla, line 190: This is non-intuitive, it seems to me that paths outside of repo root should by definition not exist in a repo CleanTree, but the problem is that for local cache/remote we do not currently differentiate between WorkingTree/CleanTree/filesystem paths.

This can be cleaned up by the remote/cache/tree separation:
    * local remote/cache should have their own working tree that would handle the case where DVC cache dir is outside of the repo root or on a different logical drive
    * DVC repo CleanTree could then disallow (ignore) paths outside of the repo root, since CleanTree paths would always and only be used to refer to paths inside the DVC repo, and not arbitrary local filesystem paths.

pmrowla:
@efiop across several profiler runs from #3867, `ignore.py` takes 5-6s of runtime for both master and this PR branch on my machine.

I agree that dvcignore/CleanTree performance definitely needs to be improved, but it seems to me like we can avoid making any optimization related changes in this PR, and handle #3869 in the next sprint

PR #3878, opened 2020-05-26
snap: restrict site-package checks

### LINKED ISSUE #3514: Adding file to dvc fails with Unknown argument(s): {'lock_args': (False,)}
When adding a file with `dvc add myfile` or `dvc push`, I get the following error:

```
ERROR: unexpected error - "Unknown argument(s): {'lock_args': (False,)}"
```

It used to work perfectly with the same version but we started getting this error today. Any idea what this could be?

**Config**
```
['remote "livebucket"']
url = s3://****/***
[core]
remote = livebucket
```

DVC version: 0.80.0
Platform: Ubuntu 
Install method: `pip install` in conda environment

casperdcl:
- [x] prevent snap's python looking for user site packages
- [ ] check if this really fixes #3514
- related: #3879 (`0.x` branch)
- fixes #3514
- reference https://github.com/snapcore/snapcraft/blob/19393ef36cd773a28131cec10cc0bfb3bf9c7e77/tools/snapcraft-override-build.sh#L18

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -38,6 +38,7 @@ parts:
         sed -rin 's/.*git.*diff.*--quiet.*//' $SNAPCRAFT_PART_SRC/dvc/version.py
     override-build: |
         snapcraftctl build
+        sed -ri 's/^(ENABLE_USER_SITE = )None$/\1False/' $SNAPCRAFT_PART_INSTALL/usr/lib/python*/site.py
         cp $SNAPCRAFT_PART_BUILD/scripts/completion/dvc.bash $SNAPCRAFT_PART_INSTALL/completion.sh
 apps:
   dvc:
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -38,6 +38,7 @@ parts:
         sed -rin 's/.*git.*diff.*--quiet.*//' $SNAPCRAFT_PART_SRC/dvc/version.py
     override-build: |
         snapcraftctl build
+        sed -ri 's/^(ENABLE_USER_SITE = )None$/\1False/' $SNAPCRAFT_PART_INSTALL/usr/lib/python*/site.py
```
efiop, line 41: Let's put a comment about what is going on here, please.

casperdcl: not actually sure the issue is fixed (https://github.com/iterative/dvc/issues/3514#issuecomment-634281454) - will do some more debugging

casperdcl: related: https://github.com/iterative/dvc/tree/snap-site-0.x

casperdcl: comment added

casperdcl:
@efiop as discussed this is ready for merge. I couldn't reproduce the original error but since this PR is the approach which snapcraft itself uses, hopefully it's fixed.

Issue #3317, opened 2020-02-13
remove: change default behaviour to `--purge`, port back patch from 1.0 branch


jorgeorpinel:
`dvc remote --outs` is the default behavior so will this option ever be used?

efiop:
It was meant for backward compatibility. We've planned to switch from it in 1.0 https://github.com/iterative/dvc/issues/763 and even have a patch there already. I guess we could consider switching it earlier. `dvc remove` needs some reconsideration in general, we are also thinking about it in terms of removing cache for a file, which is related to `gc` discussion.

jorgeorpinel:
Gotcha. Feel free to close this if it's included in some other issue or you feel it's not really impactful. Thanks for the explanation though!

efiop:
@jorgeorpinel Let's close it since it is fixed in 1.0 branch.

efiop:
On second thought... It is so trivial to fix that it is not worth waiting for 1.0 and we've been breaking worse stuff previously, so no need to be overprotective here. Keeping it open.

PR #3881, opened 2020-05-27
Implement removing of the stage

skshetry:
This PR adds support for removing stage entry for pipeline files and allows overwriting stage entry during `run`.

Backward Incompatible changes:
1. Changes `--overwrite-dvcfile` flag to `--overwrite` so that it makes sense for both kind of files.
2. Removes `--outs` flag from `dvc remove`.
3. `run` will always raise an exception if there already exists a stage, previously it'd prompt to overwrite.

PR will come shortly.

Closes https://github.com/iterative/dvc/issues/3317

### DIFF
--- a/dvc/command/remove.py
+++ b/dvc/command/remove.py
@@ -1,7 +1,6 @@
 import argparse
 import logging
 
-import dvc.prompt as prompt
 from dvc.command.base import CmdBase, append_doc_link
 from dvc.exceptions import DvcException
 
@@ -9,30 +8,10 @@ logger = logging.getLogger(__name__)
 
 
 class CmdRemove(CmdBase):
-    def _is_dvc_only(self, target):
-        if not self.args.purge:
-            return True
-
-        if self.args.force:
-            return False
-
-        msg = "Are you sure you want to remove '{}' with its outputs?".format(
-            target
-        )
-
-        if prompt.confirm(msg):
-            return False
-
-        raise DvcException(
-            "Cannot purge without a confirmation from the user."
-            " Use `-f` to force."
-        )
-
     def run(self):
         for target in self.args.targets:
             try:
-                dvc_only = self._is_dvc_only(target)
-                self.repo.remove(target, dvc_only=dvc_only)
+                self.repo.remove(target, outs=self.args.outs)
             except DvcException:
                 logger.exception(f"failed to remove '{target}'")
                 return 1
@@ -40,7 +19,7 @@ class CmdRemove(CmdBase):
 
 
 def add_parser(subparsers, parent_parser):
-    REMOVE_HELP = "Remove DVC-tracked files or directories."
+    REMOVE_HELP = "Remove stage entry and unprotect outputs"
     remove_parser = subparsers.add_parser(
         "remove",
         parents=[parent_parser],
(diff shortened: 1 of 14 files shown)


dvc/repo/run.py
```diff
@@ -74,10 +78,12 @@ def run(self, fname=None, no_exec=False, single_stage=False, **kwargs):
 
     dvcfile = Dvcfile(self, stage.path)
     if dvcfile.exists():
-        if stage_name and stage_name in dvcfile.stages:
+        if kwargs.get("overwrite", True):
+            dvcfile.remove_stage(stage)
```
skshetry, line 82: There's no need for prompt here imo. Users can simply add `--overwrite` and re-run, as to this point, we are not doing anything substantial.

efiop: @skshetry Agreed! This is in-line with `no-prompt` policy that is reflected in many tickets.

dvc/command/remove.py
```diff
@@ -1,38 +1,19 @@
...
-            " Use `-f` to force."
-        )
-
     def run(self):
         for target in self.args.targets:
             try:
-                dvc_only = self._is_dvc_only(target)
-                self.repo.remove(target, dvc_only=dvc_only)
+                self.repo.remove(
+                    target, purge_stage=self.args.purge, force=self.args.force,
```
efiop, line 15: So by default it will still only remove the outputs? If so, this is against the default behaviour we've intended for here in the ticket. How about we  purge by default? People were complaining a lot that by default it only removes the output and they wish for `--purge` by default, hence why we've introduced `--outs` to make the switch in the future (future is now :slightly_smiling_face: )

dvc/command/run.py
```diff
@@ -177,10 +177,10 @@ def add_parser(subparsers, parent_parser):
         help="Only create stage file without actually running it.",
     )
     run_parser.add_argument(
-        "--overwrite-dvcfile",
+        "--overwrite",
         action="store_true",
         default=False,
-        help="Overwrite existing DVC-file without asking for confirmation.",
+        help="Overwrite existing stage",
```
efiop, line 183:
```suggestion
        help="Overwrite existing stage.",
```

dvc/command/remove.py
skshetry: So, that'd be `--dvc-only` flag, right?

skshetry: And, no prompt, right? As it will be the default?

efiop: @skshetry So what've agreed on a year or more back, is that it should by default remove both outputs and the stage. And default behaviour would be without the prompt, yes.

But now I'm thinking that maybe it actually makes sense to remove the stage but keep the outputs by default(and, well, unprotect them too), to make it safer by-default. I guess the problem here is that the purpose of `dvc remove` now has changed. Previously it was mostly useful back in a unprotected-hardlink-by-default days, where it was used to remove the outputs so that you don't corrupt corresponding cache files by accidentally editing hardlink/symlink in-place. And now it is useful to remove stages from dvc.yaml.

skshetry: Makes sense to unprotect outs.

dvc/dvcfile.py
```diff
@@ -165,6 +165,9 @@ def remove_with_prompt(self, force=False):
 
         self.remove()
 
+    def remove_stage(self, stage):
```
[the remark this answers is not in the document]
skshetry: @pared, how about:
```suggestion
    def remove_stage(self, stage_name):
```

PR #3883, opened 2020-05-27
brancher: only yield git revisions that contain repo_root

### LINKED ISSUE #3821: dvc metrics -T raises exception if there are tags unrelated to the DVC sub-directory
DVC version: 0.94.0
Python version: 3.6.7
Binary: False
Package: pip
Supported remotes: http, https

```
NT> dvc metrics show -T
ERROR: unexpected error - [Errno 2] No such file
 
NT> dvc metrics show -T -v
2020-05-18 08:19:19,707 DEBUG: Trying to spawn '['c:\\program files\\python36\\python.exe', 'C:\\Program Files\\Python36
\\Scripts\\dvc', 'daemon', '-q', 'updater']'
2020-05-18 08:19:19,778 DEBUG: Spawned '['c:\\program files\\python36\\python.exe', 'C:\\Program Files\\Python36\\Script
s\\dvc', 'daemon', '-q', 'updater']'
2020-05-18 08:19:19,794 DEBUG: PRAGMA user_version;
2020-05-18 08:19:19,794 DEBUG: fetched: [(3,)]
2020-05-18 08:19:19,794 DEBUG: CREATE TABLE IF NOT EXISTS state (inode INTEGER PRIMARY KEY, mtime TEXT NOT NULL, size TE
XT NOT NULL, md5 TEXT NOT NULL, timestamp TEXT NOT NULL)
2020-05-18 08:19:19,825 DEBUG: CREATE TABLE IF NOT EXISTS state_info (count INTEGER)
2020-05-18 08:19:19,825 DEBUG: CREATE TABLE IF NOT EXISTS link_state (path TEXT PRIMARY KEY, inode INTEGER NOT NULL, mti
me TEXT NOT NULL)
2020-05-18 08:19:19,825 DEBUG: INSERT OR IGNORE INTO state_info (count) SELECT 0 WHERE NOT EXISTS (SELECT * FROM state_i
nfo)
2020-05-18 08:19:19,825 DEBUG: PRAGMA user_version = 3;
2020-05-18 08:19:19,965 DEBUG: SELECT count from state_info WHERE rowid=?
2020-05-18 08:19:19,981 DEBUG: fetched: [(38,)]
2020-05-18 08:19:19,981 DEBUG: UPDATE state_info SET count = ? WHERE rowid = ?
2020-05-18 08:19:20,059 ERROR: unexpected error - [Errno 2] No such file
  File "c:\program files\python36\lib\site-packages\dvc\main.py", line 49, in main
    ret = cmd.run()
  File "c:\program files\python36\lib\site-packages\dvc\command\metrics.py", line 49, in run
  File "c:\program files\python36\lib\site-packages\dvc\repo\metrics\__init__.py", line 13, in show
  File "c:\program files\python36\lib\site-packages\dvc\repo\__init__.py", line 30, in wrapper
  File "c:\program files\python36\lib\site-packages\dvc\repo\metrics\show.py", line 99, in show
    metrics = _collect_metrics(repo, targets, recursive)
  File "c:\program files\python36\lib\site-packages\dvc\repo\metrics\show.py", line 18, in _collect_metrics
    for stage in repo.stages:
  File "c:\program files\python36\lib\site-packages\funcy\objects.py", line 28, in __get__
  File "c:\program files\python36\lib\site-packages\dvc\repo\__init__.py", line 424, in stages
    return self._collect_stages()
  File "c:\program files\python36\lib\site-packages\dvc\repo\__init__.py", line 432, in _collect_stages
    for root, dirs, files in self.tree.walk(self.root_dir):
  File "c:\program files\python36\lib\site-packages\dvc\ignore.py", line 137, in walk
    for root, dirs, files in self.tree.walk(top, topdown):
  File "c:\program files\python36\lib\site-packages\dvc\scm\git\tree.py", line 139, in walk
    raise IOError(errno.ENOENT, "No such file")
FileNotFoundError: [Errno 2] No such file
 
 ```

My repo has lots of other code including other tags unrelated to DVC. I have initialised DVC and set up a pipeline in a sub-directory. I have a single tag that relates to a metric file and a few other tags that existed before DVC was initialised.
 
Does it expect all tags to contain metrics? It would seem so, given my “fix” below.
 
If I insert some diagnostics I can see that it is looking at a tag unrelated to my DVC directory and finding that the sub-directory does not exist in the GIT tree for a tag.
 
If I insert a try/except in site-packages/dvc/ignore.py:CleanTree:walk(), I can get it to work. I don’t suppose this is the best fix for this issue, but I’m unfamiliar with this code, so I’m not sure where it is iterating through the tags and I shouldn’t spend much more time on this.

__Suggestion__:
You might imagine a requirement where you only want to see metrics in tags that were branched from where “dvc init” was called, or only tags that are related to the sub-directory where “dvc init” was called.

### PR DESCRIPTION
**Pull request by pmrowla**

Brancher should only yield revisions that contain `repo.repo_root`.

If DVC repo was initialized with `subdir=True`, the DVC repo root directory may not exist at all in git revisions prior to when DVC was initialized. This will cause an exception when we try to `walk(repo.repo_root)` in `_collect_stages()`, since the walked directory doesn't exist.

Regular (non-subdir) DVC repos are not affected by this issue. In this case, even if a yielded git revision is from before DVC was initialized, the end result is that we walk the root git repo directory and do not find any DVC stages in that revision.

Fixes #3821.

### DIFF
--- a/dvc/ignore.py
+++ b/dvc/ignore.py
@@ -115,8 +115,12 @@ class DvcIgnoreFilter:
 
 
 class CleanTree(BaseTree):
-    def __init__(self, tree):
+    def __init__(self, tree, tree_root=None):
         self.tree = tree
+        if tree_root:
+            self._tree_root = tree_root
+        else:
+            self._tree_root = self.tree.tree_root
 
     @cached_property
     def dvcignore(self):
@@ -124,7 +128,7 @@ class CleanTree(BaseTree):
 
     @property
     def tree_root(self):
-        return self.tree.tree_root
+        return self._tree_root
 
     def open(self, path, mode="r", encoding="utf-8"):
         if self.isfile(path):
@@ -146,8 +150,11 @@ class CleanTree(BaseTree):
         )
 
     def _valid_dirname(self, path):
-        dirname, basename = os.path.split(os.path.normpath(path))
-        dirs, _ = self.dvcignore(os.path.abspath(dirname), [basename], [])
+        path = os.path.abspath(path)
+        if path == self.tree_root:
+            return True
+        dirname, basename = os.path.split(path)
+        dirs, _ = self.dvcignore(dirname, [basename], [])
         if dirs:
             return True
         return False
--- a/tests/func/metrics/test_show.py
+++ b/tests/func/metrics/test_show.py
@@ -1,5 +1,8 @@
+import os
+
 import pytest
 
+from dvc.repo import Repo
 from dvc.repo.metrics.show import NoMetricsError
 
 
@@ -55,4 +58,35 @@ def test_show_branch(tmp_dir, scm, dvc, run_copy_metrics):
     assert dvc.metrics.show(revs=["branch"]) == {
         "working tree": {"metrics.yaml": {"foo": 1}},
         "branch": {"metrics.yaml": {"foo": 2}},
+    }
+
+
+def test_show_subrepo_with_preexisting_tags(tmp_dir, scm):
+    tmp_dir.gen("foo", "foo")
+    scm.add("foo")
+    scm.commit("init")
+    scm.tag("no-metrics")
+
+    tmp_dir.gen({"subdir": {}})
+    subrepo_dir = tmp_dir / "subdir"
+    with subrepo_dir.chdir():
+        dvc = Repo.init(subdir=True)
+        scm.commit("init dvc")
+
+        subrepo_dir.gen("metrics.yaml", "foo: 1")
+        dvc.run(metrics=["metrics.yaml"], single_stage=True)
+
+    scm.add(
+        [
+            str(subrepo_dir / "metrics.yaml"),
+            str(subrepo_dir / "metrics.yaml.dvc"),
+        ]
+    )
+    scm.commit("init metrics")
+    scm.tag("v1")
+
+    expected_path = os.path.join("subdir", "metrics.yaml")
+    assert dvc.metrics.show(all_tags=True) == {
+        "working tree": {expected_path: {"foo": 1}},
+        "v1": {expected_path: {"foo": 1}},
     }
--- a/tests/func/metrics/test_show.py
+++ b/tests/func/metrics/test_show.py
@@ -74,7 +74,7 @@ def test_show_subrepo_with_preexisting_tags(tmp_dir, scm):
         scm.commit("init dvc")
 
         dvc.run(
-            cmd="echo 'foo: 1' > metrics.yaml",
+            cmd="echo foo: 1 > metrics.yaml",
             metrics=["metrics.yaml"],
             single_stage=True,
         )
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: tests/func/metrics/test_show.py [hunk 1]
```diff
@@ -48,3 +51,34 @@ def test_show_branch(tmp_dir, scm, dvc):
...
+    scm.tag("no-metrics")
+
+    tmp_dir.gen({"subdir": {}})
+    subrepo_dir = tmp_dir / "subdir"
+    with subrepo_dir.chdir():
+        dvc = Repo.init(subdir=True)
+        scm.commit("init dvc")
+
+        subrepo_dir.gen("metrics.yaml", "foo: 1")
+        dvc.run(metrics=["metrics.yaml"], single_stage=True)
```
skshetry, line 69: Needs command, and please rebase as https://github.com/iterative/dvc/pull/3850 has been merged.

skshetry, line 69:
```suggestion
        dvc.run(
            cmd="echo 'foo: 1' > metrics.yaml",
            metrics=["metrics.yaml"],
            single_stage=True,
        )
```

dvc/ignore.py
```diff
@@ -115,16 +115,20 @@ def __call__(self, root, dirs, files):
 
 
 class CleanTree(BaseTree):
-    def __init__(self, tree):
+    def __init__(self, tree, tree_root=None):
         self.tree = tree
+        if tree_root:
+            self._tree_root = tree_root
+        else:
+            self._tree_root = self.tree.tree_root
```
pmrowla, line 123: For subrepos, CleanTree root should be the subrepo root and not the git/scm root, otherwise we will try to filter out the actual subrepo.

This was only an issue for GitTrees from brancher (WorkingTree already uses DVC repo.root_dir as the tree root)

dvc/ignore.py
```diff
@@ -115,16 +115,20 @@ def __call__(self, root, dirs, files):
...
 
     @cached_property
     def dvcignore(self):
         return DvcIgnoreFilter(self.tree)
```
efiop, line 127: Hm, so dvcignore is still initialized with the host repo's root, right?

efiop: I guess it is an interesting question overall: should host's dvcignore affect subrepo? IIRC, git submodules are not affected by the host's .gitignore. CC @pared

pmrowla: Hmm, looks like you're right. `DvcIgnoreFilter.__init__` should probably also take a parameter for the cleantree root, since it will be different than the git tree root for subrepos.

And I don't think that git submodules are affected by the parent repo's gitignore.

pmrowla: Separate issue (#3894) has been filed for this bug, it seems like this could potentially be causing other issues in DVC besides dvcignore, where we incorrectly use the host repo DVC root instead of subrepo DVC root whenever we are using brancher.

PR #3899, opened 2020-05-28
use `selectors` instead of `select.select()`

### LINKED ISSUE #3888: tests: flaky ssh tests
https://travis-ci.com/github/iterative/dvc/jobs/340412779

```sh
tests/unit/remote/ssh/test_connection.py::test_hardlink 
[gw1] [ 80%] FAILED tests/unit/remote/ssh/test_connection.py::test_hardlink 
tests/unit/remote/ssh/test_connection.py::test_copy 
[gw1] [ 80%] FAILED tests/unit/remote/ssh/test_connection.py::test_copy 
```

I thought this was only failing on my machine, so I was ignoring it. But, looks like it occasionally happens on the CI. On my machine though this always appears.

Also, an important thing to note here is that, locally on my machine. it only fails when I'm running all of the tests, but when I rerun failed tests, it passes.

skshetry:
I noticed that we were going >1024 on file descriptors. And, as select() does not work with >1024 file descriptors, using `poll` might fix the problem. 

Though, it does not support Windows. So, I have used a higher level API [`selectors`](https://docs.python.org/3.7/library/selectors.html) that will use whatever is most efficient for the current platform. Though, due to lack of `*poll()` in Windows, it will use `select` there.

Closes #3888

### DIFF
--- a/dvc/remote/ssh/connection.py
+++ b/dvc/remote/ssh/connection.py
@@ -237,11 +237,15 @@ class SSHConnection:
             or channel.recv_ready()
             or channel.recv_stderr_ready()
         ):
-            import select
+            import selectors
+
+            selector = selectors.DefaultSelector()
+            selector.register(stdout.channel, selectors.EVENT_READ)
 
             got_chunk = False
-            readq, _, _ = select.select([stdout.channel], [], [], self.timeout)
-            for c in readq:
+            events = selector.select(self.timeout)
+            for key, _ in events:
+                c = key.fileobj
                 if c.recv_ready():
                     stdout_chunks.append(stdout.channel.recv(len(c.in_buffer)))
                     got_chunk = True


dvc/remote/ssh/connection.py
```diff
@@ -240,8 +241,11 @@ def execute(self, cmd):
             import select
 
             got_chunk = False
-            readq, _, _ = select.select([stdout.channel], [], [], self.timeout)
-            for c in readq:
+            poll = select.poll()
+            poll.register(stdout.channel, POLLIN)
```
efiop, line 245: Looks like this is not supported for windows and we'll have to use select() on it.

efiop:
@skshetry Unfortunately we won't see if this fixed it until we merge :slightly_smiling_face: Your sollution looks very good, let's fallback to select on windows and merge :slightly_smiling_face:

dvc/remote/ssh/connection.py
```diff
@@ -237,11 +237,15 @@ def execute(self, cmd):
             or channel.recv_ready()
             or channel.recv_stderr_ready()
         ):
-            import select
+            import selectors
```
efiop, line 240: Whoa, didn't know about this one! Neat!

PR #3905, opened 2020-05-28
dvc: replace `pipeline` with `dag`

### LINKED ISSUE #2392: pipeline: show: doesn't show all DAGs in the projects when ran without arguments
As per discussion with @pared we realized that the behavior of `dvc pipeline show --ascii` contradicts the [documentation](https://github.com/iterative/dvc.org/commit/796e79956413e35f5785c597fee76fb382561016). Current behavior looks for `Dvcfile` and fails if it is not found.

The expected behavior of `dvc pipeline show --ascii` is that it will find all `.dvc` files in the workspace and plot all pipelines. It does make sense IMHO that if a `.dvc` is provided then the yielded graph is showing the steps up to the file's corresponding stage.

### LINKED ISSUE #3661: Pipeline show for all a project
I was wondering if there is a way to print all the pipeline for a project?
I found that we can find only a pipeline for a DVC file.

efiop:
`pipeline list` is no more, it had some very questionable applications
(none to be precise) and was duplicating old `pipeline show` functionality.
We might introduce a different command to assist in listing (e.g. for
shell completion) in the future.

New `dvc dag` replaces `dvc pipeline show` and by-default will show
all pipelines in the project in ASCII (so there is no more --ascii)
and also `--dot` support, just like before. When given a target,
`dvc dag` (in any mode) will show DAG that leads to the target,
skipping the descendants, unless `--full` is specified.

Example commands:
```
dvc dag  # shows all pipelines in the project in ascii format
dvc dag --dot
dvc dag --vega (not implemented, just an idea on how it could be expanded in the future if we need to)
dvc dag mystage
dvc dag mystage --full
```

Fixes #3661
Fixes #2392 

https://github.com/iterative/dvc.org/pull/1383


dvc/command/pipeline.py
```diff
@@ -7,178 +7,133 @@
...
+                    return 1
+                target = stages[0]
+
+            G = _build(
+                self.repo.graph,
+                target=target,
+                commands=self.args.commands,
+                outs=self.args.outs,
+                full=self.args.full,
+            )
```
efiop, line 117: So now we build the DAG once and then visualize it as we please. Previously we used to have different methods to build it, which was too confusing and resulted in different inconsistent behavior for different flags.

dvc/command/pipeline.py
```diff
@@ -222,27 +167,14 @@ def add_parser(subparsers, parent_parser):
...
         default=False,
-        help="Print output files instead of paths to DVC-files.",
-    )
-    pipeline_show_parser.add_argument(
-        "-l",
-        "--locked",
-        action="store_true",
-        default=False,
-        help="Print locked DVC stages",
-    )
```
efiop, line 240: This option makes more sense for `dvc list` or `dvc pipeline list` then for `show`.

dvc/command/pipeline.py
```diff
@@ -198,16 +153,6 @@ def add_parser(subparsers, parent_parser):
...
-    PIPELINE_LIST_HELP = "List connected groups of stages (pipelines)."
-    pipeline_list_parser = pipeline_subparsers.add_parser(
-        "list",
-        parents=[parent_parser],
-        description=append_doc_link(PIPELINE_LIST_HELP, "pipeline/list"),
-        help=PIPELINE_LIST_HELP,
-        formatter_class=argparse.RawDescriptionHelpFormatter,
-    )
-    pipeline_list_parser.set_defaults(func=CmdPipelineList)
-
```
efiop, line 210: Maybe this command will comeback later, when we need it for shell completion.

efiop:
Had a discussion with @shcheklein and we came up with an idea to maybe rename it to `dvc dag`. E.g.
```
dvc dag
dvc dag --tree
dvc dag --dot
dvc dag --vega (теоретически в будущем)
dvc dag mystage
dvc dag mystage --full
dvc dag --commands
dvc dag --outs
```

dvc/command/dag.py
```diff
@@ -0,0 +1,192 @@
...
+            continue
+
+        root = _get_root(pipeline)
+
+        tree = Tree()
+        tree.create_node(root, root)
+        for parent, successors in nx.bfs_successors(pipeline, root):
+            for node in successors:
+                tree.create_node(node, node, parent=parent)
+
```
efiop, line 55: Could do better here for showing the whole project. Could totally show dep tree for each final output of the pipeline.

drorata:
I have two questions:
* What's the role of `Dvcfile` in the context of `dvc dag`?
* Is `dvc dag mystage --full` the same as `dvc dag --full`? And what about simply `dvc dag`? In this context, what about the case where the DAG have multiple connected components? I would argue that `dvc dag` and `dvc dag --full` would both return _all_ components while `dvc dag mystage --full` should return the whole connected component containing `mystage`.

efiop:
@drorata Thank you for the questions!

> What's the role of Dvcfile in the context of dvc dag?

None, it is no longer the default target.

> Is dvc dag mystage --full the same as dvc dag --full? 

If you only have 1 pipeline in your project, then yes, those will be the same and would be the same as `dvc dag`(no arguments).

> And what about simply dvc dag?

It will show you all the pipelines you have in your project in ASCII format.

> In this context, what about the case where the DAG have multiple connected components? I would argue that dvc dag and dvc dag --full would both return all components while dvc dag mystage --full should return the whole connected component containing mystage.

Correct, dvc dag and dvc dag --full will show the same thing: your whole DAG (all pipelines). And dvc dag mystage --full will show the whole pipeline that this stage belongs too.


Here is an example with two pipelines in the project: https://asciinema.org/a/keVBjk1IZMU1jvskw3WKTbocv

PR #3916, opened 2020-05-30
gdrive: add open

### LINKED ISSUE #3408: api: support streaming from Google Drive remotes

### LINKED ISSUE #3897: "AttributeError: NoneType object has no attribute 'wdir'" on dvc get-url call
## Bug Report

When calling ``dvc get-url gdrive://{mylink}`` from the command line, I am getting the following AttributeError: "NoneType' object has no attribute 'wdir'.  The full traceback is below.

I'm calling ``dvc get-url`` from outside the .dvc working directory.  A quick look at the source code suggests that this is related to output.LocalOutput being initialized with stage=None from the dependency.loads_from() function in repo.get_url.py.

**Output of `dvc version`:**
```sh
$ dvc version
DVC version: 0.94.0
Python version: 3.7.6
Binary: False
Package: pip
Supported remotes: gdrive, http, https
```

```
  File "d:\miniconda3\lib\site-packages\dvc\main.py", line 49, in main
    ret = cmd.run()
  File "d:\miniconda3\lib\site-packages\dvc\command\get_url.py", line 17, in run
  File "d:\miniconda3\lib\site-packages\dvc\repo\get_url.py", line 17, in get_url
  File "d:\miniconda3\lib\site-packages\dvc\dependency\__init__.py", line 83, in loads_from
  File "d:\miniconda3\lib\site-packages\dvc\dependency\__init__.py", line 68, in _get
    return LocalDependency(stage, p, info)
  File "d:\miniconda3\lib\site-packages\dvc\output\local.py", line 25, in __init__
    super().__init__(stage, path, *args, **kwargs)
  File "d:\miniconda3\lib\site-packages\dvc\output\base.py", line 98, in __init__
  File "d:\miniconda3\lib\site-packages\dvc\output\local.py", line 40, in _parse_path
    p = self.stage.wdir / p
AttributeError: 'NoneType' object has no attribute 'wdir'
```

casperdcl:
- [x] Add `open()` for `gdrive` remote
- [ ] update docs
- fixes #3408
- related #2865
- related #3897
- [x] depends https://github.com/iterative/PyDrive2/pull/39
  - [x] depends https://github.com/conda-forge/pydrive2-feedstock/pull/13
- [x] ~~depends https://github.com/iterative/PyDrive2/pull/40~~
- [x] tests in https://github.com/iterative/dvc/blob/1.0.0a6/tests/func/test_api.py#L60
  + [x] fix credentials https://travis-ci.com/github/iterative/dvc/jobs/341956240
  + [x] rebase onto current `master`

### DIFF
--- a/dvc/dependency/__init__.py
+++ b/dvc/dependency/__init__.py
@@ -3,6 +3,7 @@ from urllib.parse import urlparse
 
 import dvc.output as output
 from dvc.dependency.azure import AzureDependency
+from dvc.dependency.gdrive import GDriveDependency
 from dvc.dependency.gs import GSDependency
 from dvc.dependency.hdfs import HDFSDependency
 from dvc.dependency.http import HTTPDependency
@@ -19,6 +20,7 @@ from .repo import RepoDependency
 
 DEPS = [
     AzureDependency,
+    GDriveDependency,
     GSDependency,
     HDFSDependency,
     HTTPDependency,
@@ -33,6 +35,7 @@ DEP_MAP = {
     Schemes.SSH: SSHDependency,
     Schemes.S3: S3Dependency,
     Schemes.AZURE: AzureDependency,
+    Schemes.GDRIVE: GDriveDependency,
     Schemes.GS: GSDependency,
     Schemes.HDFS: HDFSDependency,
     Schemes.HTTP: HTTPDependency,
--- a/dvc/dependency/__init__.py
+++ b/dvc/dependency/__init__.py
@@ -3,7 +3,6 @@ from urllib.parse import urlparse
 
 import dvc.output as output
 from dvc.dependency.azure import AzureDependency
-from dvc.dependency.gdrive import GDriveDependency
 from dvc.dependency.gs import GSDependency
 from dvc.dependency.hdfs import HDFSDependency
 from dvc.dependency.http import HTTPDependency
@@ -20,7 +19,6 @@ from .repo import RepoDependency
 
 DEPS = [
     AzureDependency,
-    GDriveDependency,
     GSDependency,
     HDFSDependency,
     HTTPDependency,
@@ -35,7 +33,6 @@ DEP_MAP = {
     Schemes.SSH: SSHDependency,
     Schemes.S3: S3Dependency,
     Schemes.AZURE: AzureDependency,
-    Schemes.GDRIVE: GDriveDependency,
     Schemes.GS: GSDependency,
     Schemes.HDFS: HDFSDependency,
     Schemes.HTTP: HTTPDependency,
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -222,9 +222,12 @@ class ETagMismatchError(DvcException):
 
 
 class FileMissingError(DvcException):
-    def __init__(self, path):
+    def __init__(self, path, hint=None):
         self.path = path
-        super().__init__(f"Can't find '{path}' neither locally nor on remote")
+        self.hint = "" if hint is None else f". {hint}"
+        super().__init__(
+            f"Can't find '{path}' neither locally nor on remote{hint}"
+        )
 
 
 class DvcIgnoreInCollectedDirError(DvcException):
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -224,7 +224,7 @@ class ETagMismatchError(DvcException):
 class FileMissingError(DvcException):
     def __init__(self, path, hint=None):
         self.path = path
-        self.hint = "" if hint is None else f". {hint}"
+        hint = "" if hint is None else f". {hint}"
         super().__init__(
             f"Can't find '{path}' neither locally nor on remote{hint}"
         )
--- a/dvc/remote/gdrive.py
+++ b/dvc/remote/gdrive.py
@@ -35,7 +35,7 @@ class GDriveAuthError(DvcException):
         if cred_location:
             message = (
                 "GDrive remote auth failed with credentials in '{}'.\n"
-                "Backup first, remove of fix them, and run DVC again.\n"
+                "Backup first, remove or fix them, and run DVC again.\n"
                 "It should do auth again and refresh the credentials.\n\n"
                 "Details:".format(cred_location)
             )
--- a/dvc/remote/gdrive.py
+++ b/dvc/remote/gdrive.py
@@ -1,9 +1,11 @@
+import io
 import logging
 import os
 import posixpath
 import re
 import threading
 from collections import defaultdict
+from contextlib import contextmanager
 from urllib.parse import urlparse
 
 from funcy import cached_property, retry, wrap_prop, wrap_with
@@ -15,6 +17,7 @@ from dvc.progress import Tqdm
 from dvc.remote.base import BaseRemote, BaseRemoteTree
 from dvc.scheme import Schemes
 from dvc.utils import format_link, tmp_fname
+from dvc.utils.stream import IterStream
 
 logger = logging.getLogger(__name__)
 FOLDER_MIME_TYPE = "application/vnd.google-apps.folder"
@@ -388,6 +391,23 @@ class GDriveRemoteTree(BaseRemoteTree):
             bar_format=Tqdm.BAR_FMT_DEFAULT,
         ) as pbar:
             gdrive_file.GetContentFile(to_file, callback=pbar.update_to)
+
+    @contextmanager
+    @_gdrive_retry
+    def open(self, path_info, mode="r", encoding=None):
+        assert mode in {"r", "rt", "rb"}
+
+        item_id = self._get_item_id(path_info)
+        param = {"id": item_id}
+        # it does not create a file on the remote
+        gdrive_file = self._drive.CreateFile(param)
+        fd = gdrive_file.GetContentIOBuffer()
+        stream = IterStream(iter(fd))
+
+        if mode != "rb":
+            stream = io.TextIOWrapper(stream, encoding=encoding)
+
+        yield stream
 
     @_gdrive_retry
     def gdrive_delete_file(self, item_id):
--- a/dvc/remote/gdrive.py
+++ b/dvc/remote/gdrive.py
@@ -11,7 +11,7 @@ from urllib.parse import urlparse
 from funcy import cached_property, retry, wrap_prop, wrap_with
 from funcy.py3 import cat
 
-from dvc.exceptions import DvcException
+from dvc.exceptions import DvcException, FileMissingError
 from dvc.path_info import CloudURLInfo
 from dvc.progress import Tqdm
 from dvc.remote.base import BaseRemote, BaseRemoteTree
@@ -21,12 +21,6 @@ from dvc.utils.stream import IterStream
 
 logger = logging.getLogger(__name__)
 FOLDER_MIME_TYPE = "application/vnd.google-apps.folder"
-
-
-class GDrivePathNotFound(DvcException):
-    def __init__(self, path_info, hint):
-        hint = "" if hint is None else f" {hint}"
-        super().__init__(f"GDrive path '{path_info}' not found.{hint}")
 
 
 class GDriveAuthError(DvcException):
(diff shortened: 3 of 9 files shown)


dvc/dependency/__init__.py
```diff
@@ -19,6 +20,7 @@
 
 DEPS = [
     AzureDependency,
+    GDriveDependency,
```
shcheklein, line 23: is it actually needed to support open?

dvc/remote/gdrive.py
```diff
@@ -393,6 +396,23 @@ def _gdrive_download_file(
         ) as pbar:
             gdrive_file.GetContentFile(to_file, callback=pbar.update_to)
 
+    @contextmanager
```
shcheklein, line 399: just to double check - do we properly pass `close` down to the PyDrive?

dvc/remote/gdrive.py
```diff
@@ -393,6 +396,23 @@ def _gdrive_download_file(
...
+    @_gdrive_retry
+    def open(self, path_info, mode="r", encoding=None):
+        assert mode in {"r", "rt", "rb"}
```
shcheklein, line 402: should we raise an illegalargument here?

dvc/dependency/__init__.py
casperdcl: No - may have to revert `dependency` things for a different PR depending on @efiop feedback.

dvc/remote/gdrive.py
casperdcl: It's not done in other places in the code - would have to do everywhere if needed

dvc/remote/gdrive.py
casperdcl: pydrive2 doesn't expose any `close()` methods anywhere. I think the underlying `httplib2` closes connections before raising errors; so no special handling needed.

dvc/dependency/__init__.py
efiop: @casperdcl `get_file_checksum` is not implemented for GdriveRemote, so this is not needed.

tests/func/test_api.py
```diff
@@ -56,7 +56,9 @@ def test_get_url_requires_dvc(tmp_dir, scm):
         api.get_url("foo", repo=f"file://{tmp_dir}")
 
 
-@pytest.mark.parametrize("remote_url", all_remote_params, indirect=True)
+@pytest.mark.parametrize(
+    "remote_url", all_remote_params + [GDrive], indirect=True
```
efiop, line 60: Unless I'm missing something, this change should make all the tests in this test_api.py work with Gdrive.

casperdcl: nope because of https://github.com/iterative/dvc/blob/edfa1741db506ba12ee0244452349c1db059a026/tests/remotes.py#L144 requiring an argument (`self`).

tests/func/test_api.py
```diff
@@ -8,7 +8,7 @@
 from dvc.main import main
 from dvc.path_info import URLInfo
 from dvc.utils.fs import remove
-from tests.remotes import GCP, HDFS, OSS, S3, SSH, Azure, Local
+from tests.remotes import GCP, HDFS, OSS, S3, SSH, Azure, GDrive, Local
 
 remote_params = [S3, GCP, Azure, OSS, SSH, HDFS]
```
efiop, line 13: We should be able to simply add Gdrive to this list. Sorry for not mentioning it explicitly earlier.

tests/func/test_api.py
casperdcl: the only other classes that have that are `tests.remotes.HTTP` -> `TestHTTPRemote` (which doesn't actually look used?) and `tests.remotes.SSHMocked`

efiop: @casperdcl That's some primitive caching that was added. All other get_url's just generate a new one each time, we could try to do the same.

tests/func/test_api.py
casperdcl: yes wasn't sure if the other tests would pass - wanted to debug individually but changed now.

casperdcl:
hmm tests failing with

```
2020-05-31 19:42:01,222 ERROR: failed to push data to the cloud - GDrive remote auth failed with credentials in 'GDRIVE_CREDENTIALS_DATA'.
Backup first, remove of fix them, and run DVC again.
It should do auth again and refresh the credentials.
```

https://travis-ci.com/github/iterative/dvc/jobs/341956240

dvc/remote/gdrive.py
shcheklein: it should close if just `return` from the block ... we should check the logic carefully here

dvc/remote/gdrive.py
shcheklein: @efiop what do you think?

tests/remotes.py
```diff
@@ -141,10 +142,13 @@ class GDrive:
     def should_test():
         return os.getenv(GDriveRemote.GDRIVE_CREDENTIALS_DATA) is not None
 
-    def get_url(self):
-        if not getattr(self, "_remote_url", None):
```
shcheklein, line 145: removing this will break `TestGDriveRemoteCLI`  most likely

shcheklein: in general I believe we should probably introduce caching of get_url in for other remotes as well - it is strange to return different values within one test

tests/func/test_api.py
```diff
@@ -8,9 +8,9 @@
 from dvc.main import main
 from dvc.path_info import URLInfo
 from dvc.utils.fs import remove
-from tests.remotes import GCP, HDFS, OSS, S3, SSH, Azure, Local
+from tests.remotes import GCP, HDFS, OSS, S3, SSH, Azure, GDrive, Local
 
-remote_params = [S3, GCP, Azure, OSS, SSH, HDFS]
+remote_params = [S3, GCP, Azure, GDrive, OSS, SSH, HDFS]
```
shcheklein, line 13: we have a lot of tests for functions besides open here - do we expect all of them to pass? we'll need to implement more stuff then

dvc/remote/gdrive.py
efiop: @shcheklein Sorry or the delay. I think we could leave it with the assert for now, same as in other places.

tests/remotes.py
efiop: @shcheklein It actually shouldn't, because other remotes do that too and just reuse the results. Need to check maybe there is something wrong or special about gdrive...

tests/func/test_api.py
efiop: @shcheklein All the tests should work, there is nothing new required from GdriveRemotes. There might be some issues with Gdrive itself though (creds or something), that differentiates it from other remotes.

tests/remotes.py
shcheklein: @efiop I think they just call it once, in case of GDrive it just happened that we need it in multiple places. And in general, it's def not a good and safe interface - this value should be cached one way or another and stay the same during the test.

tests/func/test_api.py
shcheklein: Even it works out of the box, some things do not make much sense - e.g. what will `dvc.api.url` return for the object? Some URL that no tool (except DVC) can work with?

Re credentials - should be fine (we use an env var to set them, so I would expect external repo to pick it up as well). As I mentioned though, GDrive remote requires some special setup (e.g. it does not create a directory for you).

efiop:
Adjusted GDrive to have a `create_dir` method that is needed before you can use GDrive remote to push to it. Need to adjust those remote test classes to call it for the setup. @casperdcl could you take a look?

tests/func/test_api.py
shcheklein: regarding `get_url`, I think that's how the value should look like:

Screen Shot 2020-06-04 at 5 00 38 PM

(`id=` is the file ID that we can get via API)

dvc/remote/gdrive.py
casperdcl: upstream issue: https://github.com/iterative/PyDrive2/issues/44

tests/func/test_api.py
casperdcl: @shcheklein is this comment outdated since rebasing on master? Or does `tests/remotes.py` still need updating?

shcheklein: @casperdcl primarily gdrive logic should be updated I think, the one that generates URLs - right now they probably look like `gdrive://root/....` which is not very meaningful - no tools understand `gdrive://` URLs . Also, it probably returns path not ID, which is potentially not unique.

Ideally we would have an URL like on the image above (you can click and download in browser). Or it might be somethings else.

The bottom line here is that `get_url` for GDrive needs probably some research on its own and can be a separate PR, and for now we probably should throw NotImplemented.

Unless you guys see some easy and fast implementation.

casperdcl: I think you're talking about `api.get_url` rather than `tests.remotes.GDrive.get_url`. Pretty sure the latter is *meant* to work as-is after @efiop's changes but the tests are still failing requiring 2FA/confirmation codes.

efiop: @casperdcl You need to `create_dir`, `get_url` is not returning an already created dir.

casperdcl: Hmm. Any reason we can't move `create_dir` to within `get_url`?

efiop: @casperdcl Could do that, sure. Will require some work to make it do that without first passing it a dvc instance.

casperdcl: well the only other way I see is to rework the current pytest fixture (currently it hides the `GDrive` object so there's no way to call `create_dir`)

efiop: @casperdcl Both work. Choose the easiest one, we'll be revisiting this later anyway.

casperdcl: tried a simple change along the same lines as `setup_remote` but still failing with the same credentials error https://travis-ci.com/github/iterative/dvc/jobs/348374056#L7952

efiop: @casperdcl does it work locally?

tests/func/test_api.py
```diff
@@ -58,6 +63,7 @@ def test_get_url_requires_dvc(tmp_dir, scm):
 
 @pytest.mark.parametrize("remote_url", all_remote_params, indirect=True)
 def test_open(remote_url, tmp_dir, dvc):
+    ensure_dir(dvc, remote_url)
     run_dvc("remote", "add", "-d", "upstream", remote_url)
```
efiop, line 67: The reason why it doesn't work is because gdrive remote needs additional configuration. See TestGDriveRemoteCLI in test_data_cloud.py

tests/func/test_api.py
```diff
@@ -72,6 +122,7 @@ def test_open(remote_url, tmp_dir, dvc):
 @pytest.mark.parametrize("remote_url", all_remote_params, indirect=True)
 def test_open_external(remote_url, erepo_dir, setup_remote):
     setup_remote(erepo_dir.dvc, url=remote_url)
+    ensure_dir_scm(erepo_dir.dvc, remote_url)
```
casperdcl, line 125: @efiop I've added `ensure_dir_scm` here based off `setup_remote`.

<details>

- Using `ensure_dir` instead of `ensure_dir_scm`, there's
`ERROR: configuration error - config file error: remote 'upstream' doesn't exists.`
- Using `run_dvc("remote", "add", ...)` instead results in
`ERROR: configuration error - config file error: remote 'upstream' already exists. Use `-f|--force` to overwrite it.`

</details>

dvc/exceptions.py
```diff
@@ -222,9 +222,12 @@ def __init__(self, etag, cached_etag):
 
 
 class FileMissingError(DvcException):
-    def __init__(self, path):
+    def __init__(self, path, hint=None):
         self.path = path
-        super().__init__(f"Can't find '{path}' neither locally nor on remote")
+        self.hint = "" if hint is None else f". {hint}"
```
efiop, line 227: self.hint is not used

casperdcl: oops really silly. one of *those* days.

efiop:
Thanks @casperdcl ! For the record: s3 test is failing for unrelated reasons.

efiop:
@casperdcl Please don't forget about the docs :slightly_smiling_face:

PR #3921, opened 2020-05-31
completion: autogenerate

casperdcl:
ye moste beautiful autogeneration of autocompletion

- [x] ~~auto-generate `completion/dvc.bash` based on `dvc.cli`~~ use [iterative/shtab](https://github.com/iterative/shtab) to generate completions
  + [x] add `dvc completion [-s {bash,zsh}]` ~~`[-f FILE] [{dir}]`~~ subcommand
  + [x] add `choices=(Required|Optional).DVC_FILE` custom completion
  + [x] ~~add `_dvc_compgen_dirs` for e.g. output paths of `get`/`import`~~ use `choices=(Required|Optional).(FILE|DIR)` custom completion
    + [x] ~~introduce consistency calling DVC-files `targets` and other files `paths`?~~ not required ~~Required to choose between `_dvc_compgen_files` and `_dvc_compgen_DVCFiles`~~
    * [ ] double-check the current changes
      - [ ] exclude some hidden/beta commands?
- [x] make `snap` build its own completions
- [x] add CI tests
- [x] ~~potentially drop `_dvc_compgen_files` for `get`/`get-url`/`import`~~ `shtab` API offers separate completion of different args (i.e. don't complete url, but complete dirs for output path)
- [x] potentially remove `completion/dvc.bash` altogether
- [x] sort out `completion/dvc.zsh` too
- permanently puts #2985 to bed by loving *really* hard
- [x] depends on https://github.com/iterative/shtab/issues/2

much bash such python

# follow-up tasks

- [x] ~~add pre-commit hook~~ not needed as completions are dynamically generated now
- [ ] introduce consistency calling DVC-files `targets`, other files `paths`, and directories `dirs`
- [ ] introduce consistency around leading vs trailing whitespace when linebreaking strings
- [ ] update documentation https://github.com/iterative/dvc.org/issues/1447


scripts/completion/dvc.bash
```diff
@@ -4,63 +4,75 @@
...
+_dvc_cache_dir_COMPGEN=_dvc_compgen_files
+_dvc_checkout='--summary -d --with-deps -R --recursive -f --force --relink'
 _dvc_checkout_COMPGEN=_dvc_compgen_DVCFiles
 _dvc_commit='-f --force -d --with-deps -R --recursive'
 _dvc_commit_COMPGEN=_dvc_compgen_DVCFiles
-_dvc_config='-u --unset --local --system --global'
+_dvc_config='--global --system --local -u --unset'
+_dvc_daemon='updater analytics'
+_dvc_daemon_analytics=''
+_dvc_daemon_analytics_COMPGEN=_dvc_compgen_DVCFiles #?
```
casperdcl, line 24: need to check this line

scripts/completion/dvc.bash
```diff
@@ -4,63 +4,75 @@
...
+_dvc_pipeline_show='-c --commands -o --outs -l --locked --ascii --dot --tree'
 _dvc_pipeline_show_COMPGEN=_dvc_compgen_DVCFiles
-_dvc_plots='show diff'
-_dvc_plots_show='-t --template -o --out -x -y --show-json --no-csv-header --title --xlab --ylab'
-_dvc_plots_diff='-t --template --targets -o --out -x -y --show-json --no-csv-header --title --xlab --ylab'
-_dvc_pull='-j --jobs -r --remote -a --all-branches -T --all-tags -f --force -d --with-deps -R --recursive'
+_dvc_plots='show diff modify'
+_dvc_plots_diff='--targets -t --template -x -y --no-csv-header --title --xlab --ylab -o --out --show-vega'
+_dvc_plots_modify='-t --template -x -y --no-csv-header --title --xlab --ylab --unset'
+_dvc_plots_modify_COMPGEN=_dvc_compgen_DVCFiles #?
```
casperdcl, line 68: need to check this line

scripts/completion/dvc.bash
```diff
@@ -4,63 +4,75 @@
...
+_dvc_plots_show='-t --template -x -y --no-csv-header --title --xlab --ylab -o --out --show-vega'
+_dvc_plots_show_COMPGEN=_dvc_compgen_DVCFiles #?
```
casperdcl, line 70: need to check this line

scripts/completion/dvc.bash
```diff
@@ -1,66 +1,79 @@
 #!/usr/bin/env bash
```
skshetry, line 1: Is there a need to have `dvc.bash` in the codebase? Maybe we could create a command:
```console
$ dvc completions > /etc/bash_completion.d/dvc
```

casperdcl: yeah I'm not sure how much we want to bake this into `dvc` - it's a choice between having this nearly as a stand-alone module reusable for other CLI utils or something more tightly integrated.

casperdcl: also secondary point: even if we choose to bake this into `dvc` itself I would still recommend also supplying the rendered completions in the codebase as in my experience users tend to like the ability to `wget` them directly. I'd be fine with ignoring that for now and waiting for user requests though. Since most `dvc` users are some sort of devs then they might be fine with running `dvc completions | sudo tee /etc/bash_completion.d/dvc`.

dvc/command/completion.py
```diff
@@ -0,0 +1,203 @@
...
+                commands.append(cmd)
+                print_bash_commands(
+                    sub.choices[cmd], f"{prefix}_{cmd.replace('-', '_')}", fd
+                )
+        elif not any(i in sub.dest for i in UNCOMPLETABLE_POSITIONALS):
+            dest.append(sub.dest)
+        else:
+            logger.debug(f"uncompletable:{prefix}:{dest}")
+    if dest:
+        if not {"targets", "target"}.intersection(dest):
```
pared, line 54: I would consider adding exclusion for particular commands. As of now, we add `_dvc_compgen_files` for `get`, `get-url` and `import` too, which I think does not make sense.

casperdcl: i think it does make sense in terms of completing output directory path but that should be sorted out in a separate PR. Don't have a strong opinion on whether or not they should be excluded for now.

pared: Right, that makes sense. I was too much focused on the fact that, when importing, ideally you would get autocomplete for target repo.

pared:
@casperdcl you will be revisiting remaining points from the checklist?

casperdcl:
er, no - they are points I intended for reviewers to check.

efiop:
For the record, waiting for the same type of generation for zsh in this PR, to keep everything consistent. Might want to consider even creating a separate library for this, but not necessary at all and I'm not sure if it is a good fit for that.

casperdcl:
yes, though I still would like feedback specifically about the 3 commit suggestions above

scripts/completion/dvc.bash
efiop: @casperdcl `dvc daemon` is more of a service command, it is not meant to be ran by users. So we can safely skip it. But it accepts just files, not dvcfiles.

scripts/completion/dvc.bash
efiop: It is more complex than that (it wants specific output file/dir that is marked as plot in respective dvcfile), but `files` works fine for now.

scripts/completion/dvc.bash
efiop: same as with modify

efiop:
@casperdcl replied.

dvc/command/add.py
```diff
@@ -10,11 +10,11 @@
 class CmdAdd(CmdBase):
     def run(self):
         try:
-            if len(self.args.targets) > 1 and self.args.file:
+            if len(self.args.paths) > 1 and self.args.file:
```
efiop, line 13: As we've discussed, this is very weird that we need to touch the args names here. Feels like we need an explicit way of telling how to complete this specific arg. E.g. maybe some special marker/flag to complete to files/dvcfiles/(in future we would even want stages, but there is no CLI command to provide that yet)

dvc/command/completion.py
```diff
@@ -0,0 +1,203 @@
...
+GLOBAL_OPTIONS = ["-h", "--help", "-q", "--quiet", "-v", "--verbose"]
+ROOT_PREFIX = "_dvc"
+UNCOMPLETABLE_POSITIONALS = {
+    "rev",
+    "url",
+    "args",
+    "name",
+    "option",
+    "value",
+    "command",
```
efiop, line 18: This is hacky, but might be okay as a first step. Ideally we need some explicit way of telling this completion lib what to complete and with what.

dvc/command/completion.py
```diff
@@ -0,0 +1,203 @@
...
+}
+
+
+def print_bash_commands(parser, prefix=ROOT_PREFIX, fd=None):
```
efiop, line 22: doesn't need to be "public"

dvc/command/completion.py
```diff
@@ -0,0 +1,203 @@
...
+        if self.args.output == "-":
+            logger.debug("Writing bash completion to stdout")
+            fd = sys.stdout
+        else:
+            logger.info(f"Writing bash completion to {self.args.output}")
+            fd = open(self.args.output, "w")
+        parser = get_main_parser()
+        print_bash(parser, fd)
+        del fd
+        return 0
```
efiop, line 181: Why not just generate string and then print or write to file. There doesn't seem to be a reason to overcomplicate it with open/del/print_bash

efiop: Also, there are 0 tests here. Properly structuring code here would make it easier to test :slightly_smiling_face:

dvc/command/completion.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+def add_parser(subparsers, parent_parser):
+    COMPLETION_HELP = "Enable shell tab completion."
```
efiop, line 185: But it doesn't actually enable anything, it just generates the completion scripts.

dvc/command/completion.py
```diff
@@ -0,0 +1,203 @@
...
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+    completion_parser.add_argument(
+        "-o",
+        "--output",
+        help="Output filename for completion script. Use - for stdout.",
+        metavar="<path>",
+        default="-",
+    )
+    completion_parser.set_defaults(func=CmdCompletion)
```
efiop, line 203: You will have `zsh` soon and probably something else, so you should account for switching between those when you create a new command. Otherwise, it might break backward compatibility soon.

dvc/command/remote.py
```diff
@@ -106,22 +106,24 @@ def run(self):
 
 class CmdRemoteRename(CmdRemote):
     def _rename_default(self, conf):
-        if conf["core"].get("remote") == self.args.name:
-            conf["core"]["remote"] = self.args.new
+        if conf["core"].get("remote") == self.args.old_name:
```
efiop, line 109: Let's not touch unrelated code unless we really need too.

scripts/ci/check_patch.sh
```diff
@@ -10,3 +10,7 @@ python setup.py checkdocs
 
 # stop the build if there are any formatting errors
 pre-commit run --all-files
+
+# check bash completio
```
efiop, line 14:
```suggestion
# check bash completion
```

scripts/ci/check_patch.sh
```diff
@@ -10,3 +10,7 @@ python setup.py checkdocs
...
+python -m dvc completion -v -o scripts/completion/dvc.bash
```
efiop, line 15: Not sure if/how it runs right now, but deps are not installed yet, so you might very easily get an import error here.

scripts/completion/dvc.bash
efiop: If this is autogenerated already, I don't really see a point in keeping it here. It will just take time on pre-commit hooks for no reason. Let's delete these completion files and just change the docs to use `dvc completion` command.

dvc/command/add.py
casperdcl: sure, though on a related note for consistency's sake it's best to change these arg names.

efiop: @casperdcl Sure, but let's do that in a separate PR, no need to mix it into this one.

scripts/ci/check_patch.sh
casperdcl: outdated - removed

dvc/command/completion.py
casperdcl: added `--shell {bash,dvc}`

dvc/command/completion.py
casperdcl: outdated - removed

dvc/command/completion.py
casperdcl: outdated - removed

dvc/command/completion.py
casperdcl: outdated - removed. Also fixed (for `zsh` at least)

scripts/completion/dvc.bash
casperdcl: removed

dvc/command/add.py
casperdcl: sure; reverted as no longer needed

dvc/command/remote.py
casperdcl: reverted

dvc/command/completion.py
casperdcl: don't think tests are required here any more as we should rely on the upstream lib (`shtab`) being tested.

casperdcl:
@efiop any idea why the `isort` checks are failing? Passes locally for me.

dvc/command/completion.py
```diff
@@ -0,0 +1,79 @@
+import argparse
+import logging
+import os
+
+import shtab
+from dvc.command import choices
```
skshetry, line 6: @casperdcl, need a space here, that's why isort is failing.

```suggestion
import shtab

from dvc.command import choices
```

casperdcl: sure; but why are my local tests passing? `pre-commit clean && pre-commit run --all-files` passes...

casperdcl: just committed that suggestion and got further trolled by this: https://github.com/iterative/dvc/pull/4065

skshetry: @casperdcl, most likely, it's because you haven't added it to `known_third_party`?

casperdcl: `seed-isort-config`'s insisting on removing it when I try :thinking:

casperdcl: thx found the issue. Had `shtab` on my python path for development in such a way that `isort` didn't think it was third-party.

scripts/build_posix.sh
```diff
@@ -115,10 +115,10 @@ build_dvc() {
   # [1] https://github.com/iterative/dvc/issues/2585
   if [[ "$(uname)" == 'Linux' ]]; then
     mkdir -p $BUILD_DIR/$BASH_CMPLT_DIR
-    cp scripts/completion/dvc.bash $BUILD_DIR/$BASH_CMPLT_DIR/dvc
+    $LIB_DIR/dvc/dvc completion -s bash $BUILD_DIR/$BASH_CMPLT_DIR
 
     mkdir -p $BUILD_DIR/$ZSH_CMPLT_DIR
-    cp scripts/completion/dvc.zsh $BUILD_DIR/$ZSH_CMPLT_DIR
+    $LIB_DIR/dvc/dvc completion -s zsh $BUILD_DIR/$ZSH_CMPLT_DIR
```
casperdcl, line 121: this could create `_dvc` rather than `dvc.zsh` - let me know if that's not desired.

scripts/build_snap.sh
```diff
@@ -7,6 +7,10 @@ if [ ! -d "dvc" ]; then
   exit 1
 fi
 
+# generate bash completion script
+pip install .
+dvc completion -s bash -f completion.sh .
+
```
casperdcl, line 13: Ideally this should be done within the snapcraft build but haven't worked out how to get at its interpreter/virtual-env during the build process to run python -m dvc completion ....

There aren't any docs about this.

dvc/command/completion.py
```diff
@@ -0,0 +1,91 @@
...
+PREAMBLE = {
+    "bash": """
+# $1=COMP_WORDS[1]
+_dvc_compgen_DVCFiles() {
+  compgen -f -X '!*?.dvc' -- $1  # DVC-files
+  compgen -d -S '/' -- $1  # recurse into subdirs
+}
+""",
+    "zsh": "",
+}
```
casperdcl, line 25: maybe makes sense to move this to `dvc/command/choices.py` since that's where the `"DVCFile"` choice is defined?

scripts/build_snap.sh
casperdcl: nvm managed it by trawling through lots of source code.

dvc/command/completion.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class CmdCompletion(CmdBase):
```
skshetry, line 28:
```suggestion
class CmdCompletion(CmdBaseNoRepo):
```

dvc/command/base.py
```diff
@@ -31,9 +34,14 @@ def __init__(self, args):
...
-        self.repo = Repo()
-        self.config = self.repo.config
         self.args = args
+        try:
+            self.repo = Repo()
+        except NotDvcRepoError:
+            if args.cmd in HEADLESS_COMMANDS:
+                return
+            raise
+        self.config = self.repo.config
```
casperdcl, line 44: needed this to avoid

```
ERROR: you are not inside of a DVC repository (checked up to mount point '/')
```

when running `dvc completion` "headless" (no_scm, no_root) as required by the snap build.

casperdcl: replaced with `CmdBaseNoRepo` doh

casperdcl:
ready for review now. The DeepSource test can be ignored (`shell="bash"/"zsh"` is mistakenly viewed as a security issue, likely due to confusion with `Popen(shell=True)`)

dvc/command/completion.py
```diff
@@ -0,0 +1,91 @@
...
+    )
+    completion_parser.add_argument(
+        "dir",
+        help=(
+            "Output directory for completion script."
+            " Defaults to '-' for stdout (ignoring --file)."
+        ),
+        default="-",
+        nargs="?",
+        choices=choices.Required.DIR,
```
efiop, line 89: Doesn't look like it is required :) Btw, can't we distinguish required/optional from nargs?

dvc/command/daemon.py
```diff
@@ -64,5 +65,7 @@ def add_parser(subparsers, parent_parser):
         description=DAEMON_ANALYTICS_HELP,
         help=DAEMON_ANALYTICS_HELP,
     )
-    daemon_analytics_parser.add_argument("target", help="Analytics file.")
+    daemon_analytics_parser.add_argument(
+        "target", help="Analytics file.", choices=choices.Required.FILE
```
efiop, line 69: Is there a way to tell shtab that it shouldn't complete these commands? `daemon` is hidden (kinda) and shouldn't be completed.

casperdcl: I could add a `choices.Suppress`

dvc/command/completion.py
casperdcl: it is required. A bit of cleverness to get the desired behaviour.

- `nargs=?` allows no user-specified arg, resulting in default=-, ergo there is still a filled in. Forcefully unsetting (e.g. user providing empty string) will result in an error because this arg is very much required
- `choices.(Required|Optional)` is merely an extra error check and does correctly describe the fact that this is a required arg

We could use `choices.Optional` to essentially avoid `shtab` asserting non-null values, but this way seems nicer.

dvc/command/completion.py
```diff
@@ -0,0 +1,91 @@
...
+}
+PREAMBLE = {
+    "bash": """
+# $1=COMP_WORDS[1]
+_dvc_compgen_DVCFiles() {
+  compgen -f -X '!*?.dvc' -- $1  # DVC-files
+  compgen -d -S '/' -- $1  # recurse into subdirs
+}
+""",
+    "zsh": "",
```
efiop, line 24: Missing the new `dvc.yaml` files.

casperdcl: Hmm so we need:

- *.dvc
- Dvcfile
- dvc.yaml
- dvc.yml

and potentially case-insensitivity?

dvc/command/completion.py
efiop: @casperdcl Or move the contents of dvc/command/choices.py to dvc/command/completion.py :D I did find it confusing when reviewing.

dvc/command/daemon.py
efiop: @casperdcl For the whole `daemon` subcommand, right? Yeah, that might be alright.

dvc/command/completion.py
efiop: @casperdcl No, just

* *.dvc
* Dvcfile
* dvc.yaml

since those are the only *officially* supported ones. We'll add others later if we need to.

efiop: @casperdcl Also, command like `repro` should support getting particular stage name as an argument. We don't have a command to list stages right now in a form that would be useful in the completion, but just saying that we should be aware of that.

dvc/command/completion.py
efiop: I'm not sure, but maybe removing nargs would pretty much result in the same desired behaviour without causing this confusion?

dvc/command/completion.py
casperdcl: sure - I picked `choices` as a name though to match the `argparse` kwarg it's intended for...

dvc/command/completion.py
casperdcl: no - removing is the same as saying it's required to be explicitly specified. I'd prefer `nargs='?', default='-'` as the exact user-intuitive behaviour needed.

dvc/command/daemon.py
casperdcl: ah not for the whole subcommand - just for this argument.

Excluding an entire subcommand I think would make more sense higher-up... Something like:

```python
shtab.complete(parser, skip=["daemon"], ...)
```

dvc/command/completion.py
casperdcl: yes I was thinking about this. The way it's implemented now it'll be trivial to add (assuming something like `dvc stage list` will be provided in future)

dvc/command/daemon.py
efiop: @casperdcl That seems strange though :D Maybe there is something we can do in a more explicit way? `choices` worked out very nice, so I wonder if there is some appropriate trick here as well.

efiop: Not a deal breaker or anything, just wondering.

dvc/command/completion.py
casperdcl: moved

dvc/command/completion.py
casperdcl: updated

dvc/command/completion.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+
+        if self.args.dir == "-":
+            logger.debug("Writing tab completion to stdout")
+            print(script)
+        else:
+            fname = os.path.join(
+                self.args.dir, self.args.file or DEFAULT_FILENAME[shell],
+            )
+            logger.info(f"Writing tab completion to {fname}")
```
efiop, line 57:
```suggestion
            logger.info(f"Writing tab completion to '{fname}'")
```

dvc/command/completion.py
```diff
@@ -0,0 +1,100 @@
...
+        shell = self.args.shell
+        script = shtab.complete(
+            parser,
+            shell=shell,
+            preamble=PREAMBLE[shell],
+            choice_functions=CHOICE_FUNCTIONS[shell],
+        )
+
+        if self.args.dir == "-":
+            logger.debug("Writing tab completion to stdout")
```
efiop, line 51: Doesn't seem worth it.
```suggestion
```

dvc/command/completion.py
```diff
@@ -0,0 +1,100 @@
...
+            print(script)
+        else:
+            fname = os.path.join(
+                self.args.dir, self.args.file or DEFAULT_FILENAME[shell],
+            )
+            logger.info(f"Writing tab completion to {fname}")
+            with open(fname, "w") as fd:
```
efiop, line 58: Maybe not worth silently writing if that file already exists?

Also, would it be worth using some default paths? Or would that be too much of an assumption?

dvc/command/completion.py
```diff
@@ -0,0 +1,100 @@
+import argparse
+import logging
+import os
+
+import shtab
+
+from dvc.command.base import CmdBaseNoRepo, append_doc_link
+
+logger = logging.getLogger(__name__)
+DEFAULT_FILENAME = {"bash": "dvc", "zsh": "_dvc"}
```
efiop, line 10: Would it make sense to use `_dvc` for bash as well?

dvc/command/completion.py
```diff
@@ -0,0 +1,100 @@
...
+        if self.args.dir == "-":
+            logger.debug("Writing tab completion to stdout")
+            print(script)
+        else:
+            fname = os.path.join(
+                self.args.dir, self.args.file or DEFAULT_FILENAME[shell],
+            )
+            logger.info(f"Writing tab completion to {fname}")
+            with open(fname, "w") as fd:
+                print(script, file=fd)
```
efiop, line 59: Why not just
```suggestion
                fobj.write(script)
```
?

dvc/command/completion.py
```diff
@@ -0,0 +1,100 @@
...
+        script = shtab.complete(
+            parser,
+            shell=shell,
+            preamble=PREAMBLE[shell],
+            choice_functions=CHOICE_FUNCTIONS[shell],
+        )
+
+        if self.args.dir == "-":
+            logger.debug("Writing tab completion to stdout")
+            print(script)
```
efiop, line 52: I guess we could've used `logger.info` for consistency with other commands. Yeah, I know that `print` is more appropriate in CLI, we've been planning to change up that soon.

dvc/command/completion.py
```diff
@@ -0,0 +1,100 @@
...
+        default="bash",
+        choices=["bash", "zsh"],
+    )
+    completion_parser.add_argument(
+        "-f",
+        "--file",
+        help=(
+            "File name for output. Defaults depending on --shell:"
+            f" {DEFAULT_FILENAME}."
+        ),
```
efiop, line 88: the space should be trailing, not leading. Just consistency with the rest of the code.

dvc/command/completion.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+    completion_parser.add_argument(
+        "dir",
+        help=(
+            "Output directory for completion script."
+            " Defaults to '-' for stdout (ignoring --file)."
```
efiop, line 94:
```suggestion
            "Output directory for completion script. "
            "Defaults to '-' for stdout (ignoring --file)."
```

dvc/command/completion.py
```diff
@@ -0,0 +1,100 @@
...
+        help="Shell syntax for completions.",
+        default="bash",
+        choices=["bash", "zsh"],
+    )
+    completion_parser.add_argument(
+        "-f",
+        "--file",
+        help=(
+            "File name for output. Defaults depending on --shell:"
+            f" {DEFAULT_FILENAME}."
```
efiop, line 87:
```suggestion
            "File name for output. Defaults depending on --shell: "
            f"{DEFAULT_FILENAME}."
```

scripts/build_posix.sh
efiop: `_dvc` indeed seems like a better choice

dvc/command/completion.py
```diff
@@ -0,0 +1,100 @@
...
+        "-s",
+        "--shell",
+        help="Shell syntax for completions.",
+        default="bash",
+        choices=["bash", "zsh"],
+    )
+    completion_parser.add_argument(
+        "-f",
+        "--file",
+        help=(
```
efiop, line 85: This option conflicts with -f|--force convetion that we have.

dvc/command/completion.py
efiop: Also `output directory` and `-` for stdout seems conflicting as well :slightly_smiling_face: 

How about we get rid of `--file` and `dir` at all? If someone wants to write the completion script somewhere, he can simply `dvc completion bash > path/to/_dvc.bash`. No need to grow options that are not necessary.

dvc/command/completion.py
casperdcl: doesn't seem standard for bash

casperdcl: in any case we're removing this

dvc/command/completion.py
casperdcl: not true - there are dozens of cases in the rest of the code where the space is leading.

dvc/command/completion.py
casperdcl: removed

dvc/command/completion.py
casperdcl: removed

dvc/command/completion.py
casperdcl: removed

dvc/command/completion.py
casperdcl: removed

dvc/command/completion.py
casperdcl: removed

dvc/command/completion.py
casperdcl: removed

dvc/command/completion.py
casperdcl: removed

dvc/command/completion.py
casperdcl: removed

dvc/command/completion.py
casperdcl: reworded

dvc/command/completion.py
casperdcl: removed here anyway... will file an issue for later

dvc/command/completion.py
```diff
@@ -0,0 +1,68 @@
...
+    "zsh": "",
+}
+
+
+class Optional(shtab.Optional):
+    DVC_FILE = [shtab.Choice("DVCFile", required=False)]
+
+
+class Required(shtab.Required):
+    DVC_FILE = [shtab.Choice("DVCFile", required=True)]
```
casperdcl, line 32:
```suggestion
class Choices(shtab.Optional):
    DVC_FILE = [shtab.Choice("DVCFile", required=False)]
```

@efiop if we don't want to bother specifying `Required|Optional` in the codebase how about this?

efiop: @casperdcl specifying that seems useful, I wouldn't remove it.

efiop:
This is great. Thank you @casperdcl !

casperdcl:
Sure; feel free to add to #4070 with inevitable teething issues.

dvc/command/daemon.py
casperdcl: I think the existing `add_help=False` argument for subparsers can actually be used to skip instead. Released in `shtab==1.0.0` :)

Issue #3920, opened 2020-05-31
external workspaces


efiop:
We currently support [a so-called "external outputs" scenario](https://dvc.org/doc/user-guide/managing-external-data), that is based on creating separate external cache locations for each type of the output (e.g. s3, ssh etc). This scenario is unpolished and and even straight broken in some scenarios and is also constantly being misused with an intention of importing files from remote locations to local cache/workspace or remote. People often don't realise that this effectively extends your workspace outside of your local repo and needs proper isolation, in order to not run into conflict with your collegue running `dvc checkout` while you are working on the file/dir on external workspace.

This makes me think that we need to introduce proper terminology and abstraction to clearly state what this is all about and make this powerful feature usable. The solution is to introduce a concept of "workspace"s. It could possibly look something like this:

*** DRAFT ***

1) Define the external workspace you want to attach:
```
dvc workspace add myssh ssh://example.com/home/efiop
```
Now, unless explicitly configured otherwise, dvc will assume that you want to use `ssh://example.com/home/efiop/.dvc/cache` as a default cache for artifacts in that workspace. Similar to your local repo.

2) Use your workspace:
```
dvc add ssh://example.com/home/efiop/data
dvc run -o ssh://example.com/home/efiop/model.pkl ...
```
or with a special workspace-notation (similar to a so-called remote-notation that we currently have: remote://myremote/path)
```
dvc add ws://myssh/data
dvc run -o ws://myssh/model.pkl
```
This notation is nice because it allows you to redefine your workspaces (e.g. for each coworker to use his own home directory on the server) pretty easily (plus the config options can be set once in the config section for that workspace).

Current `cache.local/ssh/s3/gs/etc` sections will get removed, because it is wrong that we operate based on cache schema and not the workspace we are working on.

CC @PeterFogh , would appreciate what you think about this :slightly_smiling_face:

shcheklein:
> Current cache.local/ssh/s3/gs/etc sections will get removed, because it is wrong that we operate based on cache schema and not the workspace we are working on.

does it mean that we'll have to duplicate files on s3 (other cloud) on `dvc push`? it'll copying data from s3://something/.dvc/cache to s3://remote-storage?

> dvc run -o ssh://example.com/home/efiop/model.pkl

this notation does not solve one of the main problems - workspace isolation, so we'll have to use `ws` one similar to `remote` which kinda breaks the point, unless I'm missing something?

----------------------

I would say that ideally what we want is something like:

```
[...]

PR #3929, opened 2020-06-02
dvc: require `--external` for external outputs

### LINKED ISSUE #1545: add: from external FS or local external outputs
```
$ dvc add /Volumes/Seagate/storage/data/reddit-May2015_z.tsv.zip
```
It copies the file into the cache, removes the original file in the external drive and creates symlink instead (to the dvc cache).

Based on some offline discussions it confuses people. One of the biggest concerns - after removing DVC project (the entire dir with the code as well as DVC-cache) user sees a symlink to a file which does not exist.

1. Why should DVC work with external files? Is there any meaningful scenario? If not - let's throw an error when adding external files.
2. Even if we have a scenario for (1) let's not touch\remove external files?
3. As a result of the command, I have `reddit-May2015_z.tsv.zip.dvc` file but not the actual file in the workspace (like `reddit-May2015_z.tsv.zip`). It is not clear how to instantiate the file in the workspace?

efiop:
Fixes #1545
Related to #3920

https://github.com/iterative/dvc.org/pull/1387

### DIFF
--- a/dvc/command/run.py
+++ b/dvc/command/run.py
@@ -53,6 +53,7 @@ class CmdRun(CmdBase):
                 always_changed=self.args.always_changed,
                 name=self.args.name,
                 single_stage=self.args.single_stage,
+                external=self.args.external,
             )
         except DvcException:
             logger.exception("")
@@ -224,6 +225,12 @@ def add_parser(subparsers, parent_parser):
         action="store_true",
         default=False,
         help=argparse.SUPPRESS,
+    )
+    run_parser.add_argument(
+        "--external",
+        action="store_true",
+        default=False,
+        help="Allow outputs that are outside of the DVC repository.",
     )
     run_parser.add_argument(
         "command", nargs=argparse.REMAINDER, help="Command to execute."
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -19,6 +19,7 @@ from .utils import (
     check_circular_dependency,
     check_duplicated_arguments,
     check_missing_outputs,
+    check_no_externals,
     check_stage_path,
     compute_md5,
     fill_stage_dependencies,
@@ -52,7 +53,7 @@ def loads_from(cls, repo, path, wdir, data):
     return cls(**kw)
 
 
-def create_stage(cls, repo, path, **kwargs):
+def create_stage(cls, repo, path, external=False, **kwargs):
     from dvc.dvcfile import check_dvc_filename
 
     wdir = os.path.abspath(kwargs.get("wdir", None) or os.curdir)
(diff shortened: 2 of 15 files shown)


dvc/stage/__init__.py
```diff
@@ -32,6 +32,35 @@
 INVALID_STAGENAME_CHARS = set(string.punctuation) - {"_", "-"}
 
 
+class ExternalOutputsError(DvcException):
+    pass
+
+
+def _check_no_externals(outs):
```
skshetry, line 39: I think, we should have this check outside on `add()` as it's not really a stage that's external but an output.

dvc/stage/__init__.py
```diff
@@ -32,6 +32,35 @@
...
+    from urllib.parse import urlparse
+
+    # NOTE: preventing users from accidentally using external outputs. See
+    # https://github.com/iterative/dvc/issues/1545 for more details.
+
+    def _is_external(out):
+        # NOTE: if user uses `remote://` notation, he clearly knows what he is
+        # doing and we shouldn't error-out.
```
skshetry, line 47:
```suggestion
        # NOTE: on `remote://` notation, we shouldn't error-out.
```

efiop: @skshetry The new comment doesn't provide any hint on `why` though :slightly_smiling_face:

dvc/stage/__init__.py
efiop: @skshetry It is also used on `run` so it made sense to put it in one place. We also have some other out-relateted checks in create_stage already, like checking the cycles or output overlaps.

skshetry: Ahh, missed the `run`.

skshetry: Stage is already bloated, so I didn't want this check here. But, as it's also on `run`, it's okay, I guess.

dvc/stage/__init__.py
skshetry: @efiop, then let's make it gender-neutral. :smile:

dvc/stage/__init__.py
skshetry, line 39: So, let's move this to `stage.utils` then.

dvc/command/run.py
```diff
@@ -225,6 +226,12 @@ def add_parser(subparsers, parent_parser):
         default=False,
         help=argparse.SUPPRESS,
     )
+    run_parser.add_argument(
+        "--external",
+        action="store_true",
+        default=False,
+        help="Allow outputs that are outside of the DVC project.",
```
skshetry, line 233: Do we use `project`? Maybe:
```suggestion
        help="Allow outputs that are outside of the DVC repo.",
```

Or, simply `Allow external outputs`?

Issue #3882, opened 2020-05-27
remote: separate remote/cache/tree into distinct concepts


pmrowla:
Currently, `Remote` classes can contain methods that fall into three categories:
* remote operations - `upload`/`download`/etc
* cache operations - `save`/`checkout`/`link`/`move`/etc
* tree operations - `exists`/`isfile`/`isdir`/etc

Having all of these methods grouped into a single `Remote` class causes issues when we are dealing with paths that can come from different namespaces (trees). In particular, `LocalRemote` was originally written with the assumption that all paths for local remote or local cache operations are always filesystem paths, which is no longer the case given that we now need to support `GitTree` objects (via brancher and erepos).

Ex. `LocalRemote.exists(path)` is used in different places as a replacement for `os.path.exists()` to test if `path` exists in a local remote, if `path` exists in the local cache directory, and if `path` exists in a DVC repo, even though these are distinct operations.

Tasks:
* [x] separate remote/cache/tree methods
    * [x] tree methods -> `RemoteTree` (#3931, #3496)
    * [x] cache methods (#4019, #4041)
* [x] cache methods which take `path_info` parameters which can belong to a namespace other than `remote.tree` should take explicit `tree` parameters, so it is always clear what tree `path_info` applies to (#4019)
* [x] local cache should have its own tree instance that is separate from `repo.tree` (since local cache always operates on a filesystem working tree, but `repo.tree` does not) (#3991)
* [x] state should belong to local cache and not repo, and should only be used for working tree paths (#3991)

efiop:
> remote operations - upload/download/etc

Maybe `upload/download` are actually related to the tree, but `push/pull/status` are definitely remote concepts that will use `upload/download` in respective trees.

pmrowla:
Some minor cleanup left w/improving remote function naming, but after #4041 this refactor is pretty much done for now.

#4050 has been filed as a potential follow up issue.

PR #3931, opened 2020-06-02
remote: separate tree functions from main remote classes

pmrowla:

This PR does not change any existing behavior, it only moves an initial set of functions from `Remote` into `RemoteTree` classes (which can be accessed via `remote.tree`). The next set of functions to be moved for this refactor (separating remote and cache behavior) will also need to accept explicit tree parameters, so we need some minimal RemoteTree functionality before that.

`upload()` and `download()` will be moved into the trees in a future PR.

Related to #3882.


dvc/remote/azure.py
```diff
@@ -65,14 +99,7 @@ def get_etag(self, path_info):
     def get_file_checksum(self, path_info):
         return self.get_etag(path_info)
```
efiop, line 100: Should etag be part of the tree too? Kinda like stat.

dvc/remote/azure.py
```diff
@@ -65,14 +99,7 @@ def get_etag(self, path_info):
...
 
-    def remove(self, path_info):
-        if path_info.scheme != self.scheme:
-            raise NotImplementedError
-
-        logger.debug(f"Removing {path_info}")
-        self.blob_service.delete_blob(path_info.bucket, path_info.path)
-
-    def _list_paths(self, bucket, prefix, progress_callback=None):
+    def list_paths(self, bucket, prefix, progress_callback=None):
```
efiop, line 102: Isn't this `walk_files` ? :slightly_smiling_face:

dvc/remote/gs.py
```diff
@@ -65,11 +65,82 @@ def _upload_to_bucket(
             blob.upload_from_file(wrapped)
 
 
+class GSRemoteTree(BaseRemoteTree):
+    @property
+    def gs(self):
+        return self.remote.gs
```
efiop, line 71: This is kinda strange. Shouldn't the connection be part of the Tree?

efiop: Or is it just a convenience that will be fixed after the transition?

pmrowla: I think all of the connection/config related code for all of the remotes needs to be moved into the trees. This (and the equivalent properties in the other remotes) is just a temporary fix since I was trying to avoid moving everything all at once

dvc/remote/azure.py
pmrowla: It probably should be. I think it makes sense to implement `walk_files` for the remotes which still do not have it as a part of this overall refactoring, but in this PR I was only looking to start moving functions into the tree classes without changing existing behavior and potentially breaking things.

PR #3932, opened 2020-06-02
Fix quadratic behavior in Repo._collect_stages()

Suor:
We used to go through all stages when updating outs.


efiop:
Thanks @Suor ! :pray:

Issue #3924, opened 2020-06-01
metrics & plots: help output improvements [qa]


jorgeorpinel:
### Please provide information about your setup

DVC 1.0.0a6+283718

### Report

```console
λ dvc metrics -h
usage: dvc metrics [-h] [-q | -v] {add,show,diff,remove} ...

Commands to add, manage, collect, and display metrics.
Documentation: <https://man.dvc.org/metrics>

positional arguments:
  {add,show,diff,remove}
                        Use `dvc metrics CMD --help` to display command-specific help.
    add                 Mark a DVC-tracked file as a metric.
    show                Print metrics, with optional formatting.
    diff                Show changes in metrics between commits
    remove              Remove metric mark on a DVC-tracked file.
```
- [x] remove `add` and `remove`
- [x] desc: "Commands to display and compare metrics."
- [x] diff: "Show changes in metrics between commits." (add period: it's a full sentence)

```console
λ dvc metrics show -h
usage: dvc metrics show ...

positional arguments:
  targets             Metric files or directories (see -R) to show
```
- [x] `targets`: "Metric files ~~or directories (see -R)~~ to show" https://github.com/iterative/dvc/pull/3935

```console
λ dvc metrics diff -h
usage: dvc metrics diff ...

Show changes in metrics between commits
Documentation: ...
```
- "Show changes in metrics between commits." (end in period)

---

```console
λ dvc plots -h
usage: dvc plots [-h] [-q | -v] {show,diff} ...

Generating plots for metrics stored in structured files (JSON, CSV, TSV).
Documentation: <https://man.dvc.org/plots>

positional arguments:
  {show,diff}    Use `dvc plots CMD --help` to display command-specific help.
    show         Generate a plots image file from a metrics file.
    diff         Plot differences in metrics between commits in the DVC repository, or between the last commit and the workspace.
```
- desc: "Commands to visualize and compare plot metrics in structured files (JSON, CSV, TSV)."
- `show`: "Generate plot from a metrics file."
- `diff`: "Plot differences in metrics between commits."

```console
λ dvc plots show -h
usage: dvc plots show [-h] [-q | -v] [-t [TEMPLATE]] [-o OUT] [-x X] [-y Y] [--no-csv-header] [--show-json] [--title TITLE] [--xlab XLAB]
                      [--ylab YLAB]
                      [targets [targets ...]]

Generate a plots image file from a metrics file.
Documentation: <https://man.dvc.org/plots/show>

positional arguments:
[...]

PR #3935, opened 2020-06-02
Unify metrics behavior with dirs

### LINKED ISSUE #3930: dvc run -m/-M: inconsistent behavior with dirs
`dvc run` with `-M` flag doesn't accept directories as metrics whereas `dvc run` with `-m` flag does. This behavior should be unified.

**DVC version:**
- 0.94 (snap package)

**Additional Information :**
- OS: Ubuntu 18.04
- Discord context: https://discord.com/channels/485586884165107732/565699007037571084/717079938506686543
- Possibly related to #2973 

**Steps to reproduce**

1. Initialize git and dvc

2. Create `evaluate.py`:

```
import argparse
import os
import json

parser = argparse.ArgumentParser()
parser.add_argument("--metrics_dir", required=True)
args = parser.parse_args()
metrics_dir = args.metrics_dir

if not os.path.exists(metrics_dir):
    os.makedirs(metrics_dir)

with open(os.path.join(metrics_dir, "metrics1.json"), "w") as f:
    json.dump({"acc": 1.0}, f)

with open(os.path.join(metrics_dir, "metric2.json"), "w") as f:
    json.dump({"acc": 0.5}, f)
```

3. Create new stage with directory as cached metrics output:

```
$ dvc run -f metrics_cached.dvc -d evaluate.py -m metrics_cached  python3 evaluate.py --metrics_dir metrics_cached
Running command:
        python3 evaluate.py --metrics_dir metrics_cached

        git add metrics_cached.dvc .gitignore
```

4. Create new stage with directory as not cached metrics output:

```
$ dvc run -f metrics_not_cached.dvc -d evaluate.py -M metrics_not_cached  python3 evaluate.py --metrics_dir metrics_not_cached
Running command:
        python3 evaluate.py --metrics_dir metrics_not_cached
ERROR: failed to run command - directory 'metrics_not_cached' cannot be used as metrics.
```

Both commands should either accept directories as metrics either discard them

nik123:
Fixes #3930

### DIFF
--- a/dvc/output/base.py
+++ b/dvc/output/base.py
@@ -256,6 +256,9 @@ class BaseOutput:
                 )
             return
 
+        if self.metric:
+            self.verify_metric()
+
         assert not self.IS_DEPENDENCY
 
         if not self.changed():
--- a/dvc/output/base.py
+++ b/dvc/output/base.py
@@ -246,18 +246,16 @@ class BaseOutput:
 
         self.ignore()
 
+        if self.metric or self.plot:
+            self.verify_metric()
+
         if not self.use_cache:
             self.info = self.save_info()
-            if self.metric or self.plot:
-                self.verify_metric()
             if not self.IS_DEPENDENCY:
                 logger.debug(
                     "Output '%s' doesn't use cache. Skipping saving.", self
                 )
             return
-
-        if self.metric:
-            self.verify_metric()
 
         assert not self.IS_DEPENDENCY
 
--- a/dvc/command/run.py
+++ b/dvc/command/run.py
@@ -140,7 +140,7 @@ def add_parser(subparsers, parent_parser):
         "--metrics-no-cache",
         action="append",
         default=[],
-        help="Declare output metric file " "(do not put into DVC cache).",
+        help="Declare output metric file (do not put into DVC cache).",
         metavar="<path>",
     )
     run_parser.add_argument(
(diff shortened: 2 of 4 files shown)


dvc/output/base.py
```diff
@@ -256,6 +256,9 @@ def save(self):
                 )
             return
 
+        if self.metric:
+            self.verify_metric()
+
```
efiop, line 261: There is a
```
            if self.metric or self.plot:
                self.verify_metric()
```
a few lines above, under `if not self.use_cache`. How about we just move it outside of that if and check before that?

nik123: Fixed

efiop:
For the record: partial fix for https://github.com/iterative/dvc/issues/3924

efiop:
Looks great!

One more thing: could you please add a test for it?

nik123:
@efiop , should I fix latest DeepSource issues too?

BTW it seems that PyTest preferred over the UnitTest and all the future tests should be implemented via PyTests. Is it correct?

pmrowla:
> BTW it seems that PyTest preferred over the UnitTest and all the future tests should be implemented via PyTests. Is it correct?

Yes, this is correct

efiop:
Thank you @nik123 ! :pray: 

> @efiop , should I fix latest DeepSource issues too?

It is advisory, but it is not required. I just did a quick-n-dirty conversion to pytest for now, no biggie :slightly_smiling_face:

PR #3946, opened 2020-06-03
remote: finish separating RemoteTree methods

pmrowla:

* Moves remainder of tree-related methods out of Remote classes into the RemoteTree classes
* Implements `tree.walk_files()` for remotes that did not already have the method
* Makes `list_cache_paths()` use `tree.walk_files()` and behave the same for all cloud remotes except local/ssh

Related to #3882.

### DIFF
--- a/dvc/remote/azure.py
+++ b/dvc/remote/azure.py
@@ -76,7 +76,7 @@ class AzureRemoteTree(BaseRemoteTree):
         paths = self._list_paths(path_info.bucket, path_info.path)
         return any(path_info.path == path for path in paths)
 
-    def _list_paths(self, bucket, prefix, progress_callback=None):
+    def _list_paths(self, bucket, prefix):
         blob_service = self.blob_service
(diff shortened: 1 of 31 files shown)


dvc/remote/azure.py
```diff
@@ -15,56 +14,17 @@
 
 
 class AzureRemoteTree(BaseRemoteTree):
```
efiop, line 16: Would it make sense to make these just `Azure` or `AzureTree` and move them to `dvc/tree/azure.py` instead of continually thinking about them as remotes?

pmrowla: I think it's worth considering if we are going to start using these tree's for non-remote/cache related use cases (like workspaces).

efiop: We do that already for external outputs scenario :) Workspaces are a just a potential better abstraction for that.

But ok, let's merge and then do that later.

PR #3949, opened 2020-06-03
refactor/test multistage load for params and outputs

skshetry:

Part of #3693

It's more stricter and better-tested than before.

### DIFF
--- a/dvc/dependency/__init__.py
+++ b/dvc/dependency/__init__.py
@@ -1,6 +1,8 @@
 from collections import defaultdict
 from urllib.parse import urlparse
 
+from funcy import first
+
 import dvc.output as output
 from dvc.dependency.azure import AzureDependency
 from dvc.dependency.gs import GSDependency
@@ -87,26 +89,21 @@ def loads_from(stage, s_list, erepo=None):
     return ret
 
 
-def _parse_params(path_params):
-    path, _, params_str = path_params.rpartition(":")
-    params = params_str.split(",")
-    return path, params
-
-
 def loads_params(stage, s_list):
-    # Creates an object for each unique file that is referenced in the list
-    params_by_path = defaultdict(list)
-    for s in s_list:
-        path, params = _parse_params(s)
-        params_by_path[path].extend(params)
+    d = defaultdict(list)
+    for key in s_list:
+        if isinstance(key, str):
+            path = ParamsDependency.DEFAULT_PARAMS_FILE
+            params = [key]
+        else:
+            assert isinstance(key, dict)
+            path = first(key)
+            if not path:
+                continue
+            params = key[path]
+            assert isinstance(params, list)
+        d[path].extend(params)
 
-    d_list = []
-    for path, params in params_by_path.items():
-        d_list.append(
-            {
-                BaseOutput.PARAM_PATH: path,
-                ParamsDependency.PARAM_PARAMS: params,
-            }
-        )
-
-    return loadd_from(stage, d_list)
+    return [
+        ParamsDependency(stage, path, params) for path, params in d.items()
+    ]
--- a/dvc/dependency/__init__.py
+++ b/dvc/dependency/__init__.py
@@ -89,7 +89,7 @@ def loads_from(stage, s_list, erepo=None):
     return ret
 
 
-def loads_params(stage, s_list):
+def _merge_params(s_list):
     d = defaultdict(list)
     for key in s_list:
         if isinstance(key, str):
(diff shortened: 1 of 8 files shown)


dvc/dependency/__init__.py
```diff
@@ -87,26 +89,27 @@ def loads_from(stage, s_list, erepo=None):
     return ret
 
 
-def _parse_params(path_params):
```
skshetry, line 90: Moved this function to `repo.run()`.

dvc/dependency/param.py
```diff
@@ -40,7 +40,7 @@ def __init__(self, stage, path, params):
             info=info,
         )
 
-    def _dyn_load(self, values=None):
+    def inject_values(self, values=None):
```
skshetry, line 43: Renamed this function

dvc/repo/run.py
```diff
@@ -20,6 +20,19 @@ def is_valid_name(name: str):
     return not INVALID_STAGENAME_CHARS & set(name)
 
 
+def parse_params(path_params):
```
skshetry, line 23: This function outputs same structure as present in pipeline file.

tests/unit/output/test_load.py
```diff
@@ -0,0 +1,115 @@
...
+@pytest.mark.parametrize("key", [3, "list".split()])
+def test_load_from_pipeline_illegal_type(dvc, key):
+    stage = Stage(dvc)
+    with pytest.raises(ValueError):
+        output.load_from_pipeline(stage, [key], "outs")
+    with pytest.raises(ValueError):
+        output.load_from_pipeline(stage, [{"key": key}], "outs")
+
+
+def test_plots_load_from_pipeline(dvc):
```
skshetry, line 91: There are no tests for new `plots` props on `dvc.yaml`. Nearly messed up with it.

dvc/dependency/__init__.py
```diff
@@ -87,26 +89,31 @@ def loads_from(stage, s_list, erepo=None):
...
+                msg += f"'{type(key).__name__}'."
+                raise ValueError(msg)
+            path = first(key)
+            if not path:
+                continue
+            params = key[path]
+            if not isinstance(params, list):
+                msg = "Expected list of params for custom params file "
+                msg += f"'{path}', got '{type(params).__name__}'."
+                raise ValueError(msg)
```
skshetry, line 110: Didn't want to add assertions, so, I have used ValueError. They are purely internal, as error are likely to be raised by schema validators before even reaching here.

dvc/dependency/param.py
efiop: @skshetry Still looks odd. Maybe `reload/parse/etc`? :) Inject sounds strange.

skshetry: `inject` is a correct term here. But, it does look strange. `reload/parse` is not what it's doing. Could go with `load_values` or `fill_values`.

efiop: `fill_values` sounds better to me than `inject`.

skshetry: Done.

tests/unit/test_run.py
```diff
@@ -0,0 +1,36 @@
...
+        {"file2": ["param2"]},
+        {"file1": ["param2", "param3"]},
+        "param1",
+        "param2",
+        "param3",
+        {"file3": []},
+    ]
+
+
+@pytest.mark.parametrize("name", ["copy_name", "copy-name", "copy-name", "12"])
```
pared, line 29: why do we repeat copy-name here?

tests/unit/output/test_load.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def test_load_from_pipeline_accumulates_flag(dvc):
+    outs = output.load_from_pipeline(
+        Stage(dvc),
+        [
+            "file1",
+            {"file2": {"cache": False}},
+            {"file1": {"persist": False}},
+            {"file2": {"persist": False}},
```
pared, line 51: `persist` = `False` is default, right? Maybe for the accumulation test, it would be better to make it `True` and assert that `file2` is not cached and persisted? That way we are sure we are not using defaults.

tests/unit/test_run.py
skshetry: Thanks @pared. It should have been plain `copyname` (i improvised it to `copyName` now, you know just in case if I had missed uppercase letters to whitelist)

tests/unit/output/test_load.py
skshetry: That's a very good point. Fixed.

PR #3967, opened 2020-06-07
Performance improvement of dvcignore

### LINKED ISSUE #3869: optimize dvcignore

karajan1001:
Fix #3869

### DIFF
--- a/dvc/ignore.py
+++ b/dvc/ignore.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import re
+from itertools import groupby
 
 from funcy import cached_property
 from pathspec.patterns import GitWildMatchPattern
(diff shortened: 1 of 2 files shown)


pared:
@karajan1001 Windows build is failing.

dvc/ignore.py
```diff
@@ -27,7 +27,22 @@ def __init__(self, ignore_file_path, tree):
...
+        self.ignore_spec = (
+            re.compile("|".join(ignore_pattern_list))
+            if ignore_pattern_list
+            else None
+        )
+        self.include_spec = (
+            re.compile("|".join(include_pattern_list))
+            if include_pattern_list
+            else None
+        )
```
Suor, line 45: Aren't we loosing order here? I think that matters.

karajan1001: > @karajan1001 Windows build is failing.
Sorry for late, I went to the hospital last weekend but had recovered now.


> Aren't we loosing order here? I think that matters.

Yes，It losing order here。According to [gitignore](https://git-scm.com/docs/gitignore)
`
An optional prefix "!" which negates the pattern; any matching file excluded by a previous pattern will become included again. 
`
Only late rules can override early ones.

karajan1001:
> LGTM, though I would not close #3869 just yet.
> @karajan1001 you mentioned the flashtext library, and I could not find the description of the experiment conducted, that gave the author conclusion that it makes sense to use `trie`-based approach only above some number of keywords. What about matching few `gitignore` patterns against hundreds of files? Maybe there could be gain too.
> 
> Also, I wonder whether `re.compile` has limit of pattern size. Checked against 100k, seems to not be a problem. I did not find resources pointing to the limit.

Maybe this is because `regex` is written in `C`, and `falshtext` is in `Python`.

karajan1001:
@pared  @Suor 
Because there is still something wrong with my ASV. I can only give a command-line  result
Screenshot from 2020-06-13 20-40-30
This is the current version.
```
[ 81.82%] · For dvc commit fc42ca72 <1.0.0a1>:
[ 90.34%] ··· status.DVCIgnoreBench.time_status                                                                                                                          501±3ms
[ 90.91%] ··· status.DVCStatusBench.time_status                                                                                                                         84.3±2ms
```
This is the result after this submit.
```
[ 90.91%] · For dvc commit a86cedbf <1.0.0a1test>:
[ 99.43%] ··· status.DVCIgnoreBench.time_status                                                                                                                          104±1ms
[100.00%] ··· status.DVCStatusBench.time_status                                                                                                                       79.8±0.5ms
```

pared:
@karajan1001, thanks for this PR!

Now, that implementation of `dvcignore` changed, @Suor point starts to make sense. I did not foresee that in https://github.com/iterative/dvc-bench/pull/30
@karajan1001, as you are the original author of `dvcingore` benchmarks, would you be interested in reintroducing multiple `dvcignore` files benchmark? We could pull it to the extreme and do `30 files x 1 rule`, instead of original `3 x 10`.

PR #3972, opened 2020-06-08
Fix loading of params from run cache

skshetry:
This is more or less a dead code. The params and deps loaded from the stage cache are not used anywhere. I changed the interface when refactoring but missed on changing it here, but as it is an "unused" stuff, it didnot affect run-cache at all. Anyway, I have fixed it, and even though it's an internal API, I have added a test so that we can catch this soon. 

### DIFF
--- a/dvc/stage/cache.py
+++ b/dvc/stage/cache.py
@@ -80,27 +80,12 @@ class StageCache:
     def _create_stage(self, cache, wdir=None):
         from dvc.stage import create_stage, PipelineStage
 
-        params = []
-        for param in cache.get("params", []):
-            if isinstance(param, str):
-                params.append(param)
-                continue
-
-            assert isinstance(param, dict)
-            assert len(param) == 1
-            path = list(param.keys())[0]
-            params_list = param[path]
-            assert isinstance(params_list, list)
-            params.append(f"{path}:" + ",".join(params_list))
-
         stage = create_stage(
             PipelineStage,
             repo=self.repo,
             path="dvc.yaml",
             cmd=cache["cmd"],
             wdir=wdir,
-            params=params,
-            deps=[dep["path"] for dep in cache.get("deps", [])],
             outs=[out["path"] for out in cache["outs"]],
             external=True,
         )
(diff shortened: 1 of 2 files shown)


dvc/stage/cache.py
```diff
@@ -80,19 +80,8 @@ def _load(self, stage):
...
-
-            assert isinstance(param, dict)
-            assert len(param) == 1
-            path = list(param.keys())[0]
-            params_list = param[path]
-            assert isinstance(params_list, list)
-            params.append(f"{path}:" + ",".join(params_list))
-
+        params_cache = cache.get("params", {})
+        params = [{f: list(p.keys())} for f, p in params_cache.items()]
```
skshetry, line 84: @efiop, is params and deps even required here (I assume not, but just wanted to confirm)?

tests/unit/stage/test_cache.py
```diff
@@ -158,3 +162,30 @@ def test_stage_cache_wdir(tmp_dir, dvc, run_copy, mocker):
...
+    tmp_dir.gen({"params.yaml": yaml.dump(p1), "myparams.yaml": yaml.dump(p2)})
+    tmp_dir.gen("dep", "dependency")
+    stage = run_copy(
+        "dep",
+        "out",
+        params=["foo,bar", "myparams.yaml:baz,qux"],
+        single_stage=True,
+    )
+    with dvc.state:
+        cache_entry = dvc.stage_cache._load(stage)
```
skshetry, line 178: Maybe we should rename `StageCache` to `RunCache`. Just saying.
```suggestion
        cache_entry = dvc.run_cache._load(stage)
```

dvc/stage/cache.py
efiop: You are right, this is not used right now. We are filling everything from cached locks anyway, so this is not used. Let's get rid of it.

skshetry: Removed.

tests/unit/stage/test_cache.py
skshetry: Will do it later when I run into `stage/cache.py`.

PR #3973, opened 2020-06-08
[WIP] remote: use explicit tree parameter for cache operations

pmrowla:

* `remote.save()` now takes an explicit `tree` parameter
    * When `tree` is `remote.tree`, save is done via `move()` + `link()`
    * For all other trees, save is done by copying object from `tree` into `remote.tree`

### DIFF
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -400,17 +400,14 @@ class BaseRemote:
     def cache(self):
         return getattr(self.repo.cache, self.scheme)
 
-    def get_file_checksum(self, path_info):
-        raise NotImplementedError
-
-    def _calculate_checksums(self, file_infos):
+    def _calculate_checksums(self, file_infos, tree):
         file_infos = list(file_infos)
         with Tqdm(
             total=len(file_infos),
             unit="md5",
             desc="Computing file/dir hashes (only done once)",
         ) as pbar:
-            worker = pbar.wrap_fn(self.get_file_checksum)
+            worker = pbar.wrap_fn(tree.get_file_checksum)
             with ThreadPoolExecutor(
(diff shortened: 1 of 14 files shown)


dvc/remote/base.py
```diff
@@ -397,51 +400,44 @@ def supported(cls, config):
...
-    def _calculate_checksums(self, file_infos):
+    def _calculate_checksums(self, file_infos, tree):
         file_infos = list(file_infos)
         with Tqdm(
             total=len(file_infos),
             unit="md5",
             desc="Computing file/dir hashes (only done once)",
         ) as pbar:
-            worker = pbar.wrap_fn(self.get_file_checksum)
+            worker = pbar.wrap_fn(tree.get_file_checksum)
```
efiop, line 410: So we still don't have a cache/remote separation?

pmrowla: @efiop Not yet. I'm still thinking about the best way to go about that particular separation.

The "remote" functions like `push`/`pull`/`status` (which are currently in `LocalRemote`) are really functions for syncing the contents of two caches. So I've been wondering if they should just be abstracted into functions for syncing from "this/self" cache to/from any other cache, and then moved into `BaseRemote`?

And then the remote classes could still be renamed to `Cache` or `RemoteCache` (or something else along those lines), but there wouldn't be any further separation into new classes (other than the tree methods).

pmrowla: I know that we would still have the external output/dependency use case to consider, but it seems like in practice, that is still just a situation where we would have two "s3 caches"

* one which is used as the typical "dvc remote" for syncing local cache <-> s3
* and the second which is used as the s3 external out cache

pmrowla:
closed in favor of #4019

PR #3977, opened 2020-06-08
remote: tree: don't ignore exceptions

efiop:
Moving ignore logic to the remote itself, so that tree could behave
properly.

Related to #3897

### DIFF
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -1,10 +1,9 @@
-import errno
 import hashlib
 import itertools
 import json
 import logging
 import tempfile
-from concurrent.futures import ThreadPoolExecutor
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from copy import copy
 from functools import partial, wraps
 from multiprocessing import cpu_count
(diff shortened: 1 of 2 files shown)


dvc/remote/base.py
```diff
@@ -257,8 +239,7 @@ def _download_dir(
                 )
             )
             with ThreadPoolExecutor(max_workers=self.remote.JOBS) as executor:
-                futures = executor.map(download_files, from_infos, to_infos)
-                return sum(futures)
+                executor.map(download_files, from_infos, to_infos)
```
efiop, line 242: Oops, swallowing exceptions. Will fix.

dvc/remote/base.py
```diff
@@ -257,8 +241,11 @@ def _download_dir(
                 )
             )
             with ThreadPoolExecutor(max_workers=self.remote.JOBS) as executor:
-                futures = executor.map(download_files, from_infos, to_infos)
-                return sum(futures)
+                executor.map(
```
efiop, line 244: This is a tree method, so it also should raise an exception. Probably the first one that it gets, because it is similar to how you would implement it with `os.walk()`

dvc/remote/local.py
```diff
@@ -515,14 +535,14 @@ def _process(
 
         if download:
             func = partial(
-                remote.tree.download,
+                _log_exceptions(remote.tree.download, "download"),
```
efiop, line 538: Still ugly, but now explicit and tree is not affected by remote logic.

dvc/remote/base.py
```diff
@@ -257,8 +239,25 @@ def _download_dir(
...
+                # which might cause unexpected results in his pipeline.
+                for future in as_completed(futures):
+                    # NOTE: executor won't let us raise until all futures that
+                    # it has are finished, so we need to cancel them ourselves
+                    # before re-raising.
+                    exc = future.exception()
+                    if exc:
+                        for entry in futures:
+                            entry.cancel()
+                        raise exc
```
efiop, line 261: Also, just keeping this consistent for now. In the future we might want to condier adding something like `onerror` in `os.walk`, which will let us know if we should somehow process the exception and continue or abort.

efiop: Though, probably better to try to download to a temporary location first and delete it if unsuccessful. That would be consistent for now. Will look into Output.download() too.

PR #3985, opened 2020-06-09
tests: serialize pipeline file

skshetry:
This PR adds sorting for params keys as well. Before, it was only getting sorted for lockfile.
Rest are tests and a bit of refactoring that does not change anything.

Part of #3693

### DIFF
--- a/dvc/stage/serialize.py
+++ b/dvc/stage/serialize.py
@@ -3,14 +3,15 @@ from functools import partial
 from operator import attrgetter
 from typing import TYPE_CHECKING, List
 
-from funcy import lsplit, rpartial
+from funcy import post_processing
 
 from dvc.dependency import ParamsDependency
 from dvc.output import BaseOutput
-from dvc.stage.params import StageParams
-from dvc.stage.utils import resolve_wdir
 from dvc.utils.collections import apply_diff
 from dvc.utils.yaml import parse_yaml_for_update
+
+from .params import StageParams
+from .utils import resolve_wdir, split_params_deps
 
 if TYPE_CHECKING:
     from dvc.stage import PipelineStage, Stage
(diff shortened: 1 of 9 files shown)


dvc/stage/serialize.py
```diff
@@ -70,16 +82,15 @@ def _serialize_params(params: List[ParamsDependency]):
         dump = param_dep.dumpd()
         path, params = dump[PARAM_PATH], dump[PARAM_PARAMS]
         if isinstance(params, dict):
-            k = list(params.keys())
+            k = sorted(params.keys())
```
skshetry, line 85: We were not sorting params keys for pipeline file before, only for the lock. Now, the same behavior is for lockfile.

tests/unit/stage/test_serialize_pipeline_file.py
```diff
@@ -0,0 +1,196 @@
...
+        {"custom.yaml": ["baz", "pqrs", "wxyz"]},
+    ]
+
+
+@pytest.mark.parametrize(
+    "typ, extra",
+    [("plots", {"plot": True}), ("metrics", {"metric": True}), ("outs", {})],
+)
+def test_outs_and_outs_flags_are_sorted(dvc, typ, extra):
+    stage = create_stage(PipelineStage, dvc, deps=["input"], **kwargs)
```
skshetry, line 108: Maybe, we should implement `{metrics,plots}[_persist][_no_cache]` that are missing?

pared: @skshetry Could you elaborate?

skshetry: On `.dvc` and `dvc.yaml` files, we allow metrics and plots to be `cache: False` and `persist: True`. But, it cannot be set from outside (eg: `fill_stage_outputs` and `dvc run`).  So, I was just saying that maybe we should implement that for `run`.

efiop: @skshetry Let's not create those `dvc run` options for now, it is crowded there as it is. We clearly need to consider a different CLI approach, as `metrics_persist_no_cache` is just too ugly and doesn't scale if we introduce yet another option :slightly_smiling_face: We've been discussing something like using a comma to join those fstab-style, but, again, not worth bothering with that for now.

efiop:
For the record: snap is failing for unrelated reasons.

PR #3991, opened 2020-06-09
remote: use separate working tree instance for local cache

pmrowla:

Related to #3882

* Local cache/remote always uses its own working tree
    * local cache no longer affected by whether `repo.tree` is working or git tree
* State is tied to the local cache working tree rather than `repo.tree`
* CleanTree now ignores paths outside of the DVC tree/repo root
    * previously it was used interchangeably with working tree, and was expected to return true for paths outside of the DVC repo root in cases where local cache directory was moved or on a different logical drive
    * local cache working tree should now be used for handling all cache paths, not the DVC repo cleantree instance

### DIFF
--- a/dvc/state.py
+++ b/dvc/state.py
@@ -7,7 +7,6 @@ import sqlite3
 from urllib.parse import urlencode, urlunparse
 
 from dvc.exceptions import DvcException
-from dvc.scm.tree import WorkingTree
 from dvc.utils import current_timestamp, relpath, to_chunks
 from dvc.utils.fs import get_inode, get_mtime_and_size, remove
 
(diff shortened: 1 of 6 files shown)


dvc/state.py
```diff
@@ -89,10 +88,11 @@ class State:  # pylint: disable=too-many-instance-attributes
     MAX_INT = 2 ** 63 - 1
     MAX_UINT = 2 ** 64 - 2
 
-    def __init__(self, repo):
+    def __init__(self, local_cache):
+        repo = local_cache.repo
         self.repo = repo
         self.root_dir = repo.root_dir
-        self.tree = WorkingTree(self.root_dir)
+        self.tree = local_cache.tree.work_tree
```
efiop, line 95: After remote/cache/tree separation, the only default tree (aka cache.tree) that local cache is working on will be the work tree, right?

pmrowla: Yes, that's correct

efiop, line 95: @pmrowla State itself has nothing to do with the cache, but rather with the repos working tree, right? So maybe we should have `work_tree`(or wtree) in repo itself and make local cache and state use it? Just not quite obvious why state depends on local cache here.

pmrowla: It seemed to me that since state is only applicable whenever there is a local cache, and used as a cache for calculating md5s which will go into the local cache, it would make more sense to have state associated with the local cache's working tree rather than with a repo tree instance. The repo doesn't actually need a working tree to function (since we can use git trees), but local cache and state only use a working tree.

PR #3994, opened 2020-06-09
Refactor plots

### LINKED ISSUE #3818: Make plots reading and rendering API friendly
Context: this will help using it in viewer.

This things need to be implemented somewhat public:

- [ ] separate data and template extraction
- [ ] separate raw data fetching and post-processing (add rev and some json magic)
- [ ] provide a way to make a spec from raw data and template

The last point is culmination. The viewer scenario of using it is:
- parse repo, save raw plot data and templates (as text) to DB.
- on user request fetch data for selected commits and pass it to some template (postprocessing happens somewhere inbetween or within a template part) 

Ideally one won't need post-processing or if that is not possible it might be separated into postread and prerender parts. A thing to discuss.

This is not p0, since I'll be handling it in some dirty way for now myself, but still important. 

**NOTE:** this issue is scoped to plot data reading and plot rendering, plot data file discovery is discussed separately in #3807.

Suor:
- separate data collection and rendering
- provide a way to use custom template loader
- drop hard-coding datafiles in template feature (discussed this with @dmpetrov)

Closes #3818.


dvc/repo/plots/__init__.py
```diff
@@ -33,3 +104,113 @@ def modify(self, path, props=None, unset=None):
 
         dvcfile = Dvcfile(self.repo, out.stage.path)
         dvcfile.dump(out.stage, update_pipeline=True)
```
skshetry, line 106: @Suor, `plots modify` should not generate lockfile at all. Please pass `no_lock=True` here.

skshetry: I'll do it, as I am working on the plots as well.

Suor: This line is not changed in this PR) Should be fixed separately though.

Suor: Looks like `.dump_lock()` should be separated from `.dump()` otherwise this error will introduce itself somewhere else.

skshetry: I agree. But, for single stage file, we would have to make it a no-op.

Suor: That's ok.

skshetry: I think, it's the same thing compared to passing args. We just need to have a sane default.

skshetry: Initially, the idea was not to touch `dvc.yaml` at all, but as we are doing it quite often, it's better to protect `dvc.lock` rather than `dvc.yaml` by default.
So, we could change it to the following default:
```py
def dump(update_pipeline=True, lock=False)
```

The only place were we need to update `lock` are `dvc repro` and `dvc run`, at least atm.

dvc/repo/plots/__init__.py
```diff
@@ -33,3 +102,113 @@ def modify(self, path, props=None, unset=None):
...
+    else:
+        outs = (out for stage in repo.stages for out in stage.outs if out.plot)
+
+    return {str(out): _plot_props(out) for out in outs}
+
+
+def _plot_props(out):
+    if not out.plot:
+        raise DvcException(
+            f"'{out}' is not a plot. Use `dvc plots modify` to change that."
```
jorgeorpinel, line 130: plots modify can change a file that is not a plot into a plot? Or maybe I don't understand this message.

dvc/repo/plots/__init__.py
```diff
@@ -33,3 +102,113 @@ def modify(self, path, props=None, unset=None):
...
+        )
+    if isinstance(out.plot, list):
+        raise DvcException("Multiple plots per data file not supported.")
```
jorgeorpinel, line 133: Should this restriction be documented?

dvc/repo/plots/__init__.py
skshetry: Yes, it can. It's a feature.

dvc/repo/plots/__init__.py
skshetry: Did this work before? I am not sure, I think I messed this up with recent refactoring. :smile:

dvc/repo/plots/__init__.py
jorgeorpinel: OK thanks. What I meant is that I didn't understand what the sentence meant. Thinking about it now I think I get it: a regular output is being given to a plots command right? And you can make it into a plot with plots modify. So 2 things:

1. I recommend something like this for message for now: ``"'{out}' is found in stage '{stg}' as an output but not as a plot. You can make it into a plot with `plots modify ...` ``
2. Actually, what's the command to make an output into a plot with plots modify? I'd like to document this feature.
3. Why not just try to plot the output instead and throw a formatting error if it's not valid?

Thanks

dvc/repo/plots/__init__.py
jorgeorpinel: Idk whether it worked before TBH. Probably not. I guess this restriction is petty obvious i.e. no need to put it in docs explicitly for no. Thanks for the reply! Resolving.

dvc/repo/plots/__init__.py
skshetry: 1. Too long, and not any useful information than the one provided already.
2. `dvc plots modify output-file-name-that-is-not-a-plot.file_ext`
You can provide any other flags on the same command, like `--template` too.

jorgeorpinel: 2. Thanks

1. The existing string is hard to read and understand. Can we come up with a better suggestion if you mine isn't liked? "to change that" to change what?

jorgeorpinel: ``The given output file is not a known plot. Use `dvc plots modify {out}` to turn it into one.``

p.s. why is it bad that a message is relatively long? (2 sentences)

jorgeorpinel: Also, can plots modify make ANY file into a plot, or only other outputs? What about existing metrics?
Does it validate plot file format?
Thanks

Suor: > p.s. why is it bad that a message to the user is relatively long? (2 sentences)

The longer the message the smaller percentage of people will read it.

> Also, can plots modify make ANY file into a plot, or only other outputs?

Only outputs. In current dvc-file design only outputs may become metrics or plots. Metric might become a plot. Not sure whether it will stop being a metric.

> Does it validate plot file format?

No. `dvc run` doesn't validate it either. Strictly speaking we don't have such thing as validation here, we can check whether loading it results in an error though.

jorgeorpinel: > The longer the message the smaller percentage of people will read it.

Arguable. But if the text is hard to understand, it doesn't matter how many people read it. I guess I'll just sent a PR with a suggested text instead of arguing here.

Thanks for the other info. TBH it seems a little obscure/ arbitrary behavior but I'm not opposed to it. I just think it would be best to re-think it. Not really my realm so I'll just ping @shcheklein and @efiop here 🙂 

> dvc run doesn't validate it... we can check whether loading it results in an error

Please note causally I opened an issue about this (not related to this conversation originally): #4068

PR #3990, opened 2020-06-09
snap: build fixes

casperdcl:
- [x] fix #3989 `snapcraft --use-lxd` builds without `sudo`
- [x] enable security notifications https://snapcraft.io/blog/introducing-developer-notifications-for-snap-security-updates
- [x] add `v1` snap channels #3872
- [x] ~~reduce snap size https://forum.snapcraft.io/t/reducing-the-size-of-desktop-snaps/17280~~ doesn't work with `plugin: python` (also barely reduces size in this case - 107MB vs 97MB)

Issue #3989, opened 2020-06-09
snap: 'sudo' cannot be used with build provider 'lxd'


efiop:
https://travis-ci.com/github/iterative/dvc/jobs/346490988#L495

Issue #3872, opened 2020-05-25
snap: transition to v1


casperdcl:
## Anyone arriving here getting this message:

![image](https://user-images.githubusercontent.com/10780059/85142798-c9b5b000-b240-11ea-9ad0-75e2310c7d56.png)

**Running:**

- `snap refresh --channel=latest/beta dvc` (to upgrade now), or
- `snap refresh --channel=v0/stable dvc` (to never upgrade)

**makes the message go away.**

Otherwise (*if you ignore the message*) you'll be upgraded from `v0` to `v1` **automatically** in a few days.

See https://dvc.org/blog/june-20-community-gems for some notes on upgrading to v1 (it's simple!)

----

## Maintainer notes:

Snap packages automatically update to the most recent version in a channel.

The default channel is **`--latest/stable`** (with the slash, yes; not my design choice) and normally we push tagged releases to it.

During breaking transitions (v0 to v1), though, we need a transition phase.

- [x] #3999 push the last `0.x` to a new `--v0/stable` channel
- [x] #3990 push the current `1.x` to `--v1/beta` and `--latest/beta`
- [x] push a new `0.x-final` to **`--latest/stable`**
  + [x] #3999 this issues a warning in the update available and error messages. This warning will state the 2 options for the user:
    * `snap refresh --channel=v0/stable` to stick to the old version
    * `snap refresh --channel=latest/beta` & use new API/syntax
    * `snap refresh --channel=v1/beta` & use new API/syntax. This would stop them getting warnings about `--v2` in future. Probably bizarre ("I wanna upgrade now but never again") and not worth mentioning this option.

After the transition period (say 30 days?) we go back to normalcy:

- [x] close the `--v1/beta` channel. Users on it will automatically get shifted to `--v1/stable`.
- [x] regularly push `1.x` updates to `--v1/stable` and **`--latest/stable`**

Some people will be be hit by errors at this stage, but ~~serves them right for ignoring the transition warning messages~~ we tried our best.

I'm not happy with the channel naming conventions either but such is life: https://snapcraft.io/docs/channels. In fact we have to manually request `--v0`, `--v1` etc.

- [x] ~~Please bump here~~done: https://forum.snapcraft.io/t/track-request-for-dvc/17735
- [x] #3999, #4038 sync `v1` and `v0` snap build improvements (e.g. #3990, #4037)
- related #3777
- realted https://github.com/iterative/dvc.org/issues/906

casperdcl:
[...]

PR #3999, opened 2020-06-09
snap: v0: build fixes & updates

casperdcl:
- [x] sync `v1` fixes & updates #3990 <- #3989
- [x] add `v0` channels #3872
  + [x] auto-infer major version number
- [x] add transition warning messages #3872
  + > this issues a warning in the update available and error messages. This warning will state the 2 options for the user:
    > * `snap refresh --channel=v0/stable` to stick to the old version
    > * `snap refresh --channel=latest/beta` & use new API/syntax
  + [ ] (in the warning message) provide a link to an upgrade/breaking changes summary page on dvc.org
- [ ] maybe fix tests which have changed for v1? https://travis-ci.com/github/iterative/dvc/builds/171019482:

```
FAILED tests/func/test_repro.py::TestReproExternalSSH::test - dvc.exceptions....
FAILED tests/func/test_repro_multistage.py::TestReproExternalSSHMultiStage::test
```

### DIFF
--- a/dvc/updater.py
+++ b/dvc/updater.py
@@ -52,10 +52,7 @@ class Updater(object):  # pragma: no cover
             return
 
         if PKG == "snap":
-            # Ideally `if SNAP_CHANNEL.startswith(self.current): return`
-            # where `SNAP_CHANNEL` is:
-            #   snap info dvc | grep -E '^tracking:' | awk '{print $2}'
-            # However that's too slow to check
+            # hardcoded transition message
             self.current = f"v{version.parse(self.current).major}"
             self.latest = f"v{version.parse(self.current).major + 1}"
             self._notify()
--- a/dvc/updater.py
+++ b/dvc/updater.py
@@ -53,8 +53,9 @@ class Updater(object):  # pragma: no cover
 
         if PKG == "snap":
             # hardcoded transition message
-            self.current = f"v{version.parse(self.current).major}"
-            self.latest = f"v{version.parse(self.current).major + 1}"
+            version_major = version.parse(self.current).major
+            self.current = "v{}".format(version_major)
+            self.latest = "v{}".format(version_major + 1)
             self._notify()
             return
 
--- a/dvc/updater.py
+++ b/dvc/updater.py
@@ -48,7 +48,35 @@ class Updater(object):  # pragma: no cover
             logger.debug(msg.format(self.lock.lockfile, action))
 
     def check(self):
-        if os.getenv("CI") or env2bool("DVC_TEST") or PKG == "snap":
+        if os.getenv("CI") or env2bool("DVC_TEST"):
+            return
+
+        if PKG == "snap":
+            # Ideally need to check if SNAP_CHANNEL.startswith(self.current),
+            # where SNAP_CHANNEL is:
+            #   snap info dvc | grep -E '^tracking:' | awk '{print $2}'
+            # However a quicker way is to skip if a user-configurable variable
+            # `skip_update_check` is set
+            import subprocess
+
+            kwargs = {"close_fds": True}
+            if os.name == "nt":
+                kwargs["shell"] = True
+                cmd = []
+            else:
+                kwargs["shell"] = False
+                cmd = ["/bin/sh", "-c"]
+            p = subprocess.Popen(
+                cmd + ["snapctl", "get", "skip_update_check"], **kwargs
+            )
+            skip_update_check = p.communicate()[0]
+            if skip_update_check not in ("", "0", "false", "False"):
+                # frozen on current version so suppress notification
+                return
+
+            self.current = f"v{version.parse(self.current).major}"
+            self.latest = f"v{version.parse(self.current).major + 1}"
+            self._notify()
             return
 
         self._with_lock(self._check, "checking")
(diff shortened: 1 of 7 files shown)


scripts/ci/before_install.sh
```diff
@@ -53,10 +53,10 @@ fi
 
 if [[ -n "$TRAVIS_TAG" ]]; then
   if [[ $(echo "$TRAVIS_TAG" | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$') ]]; then
-    echo "export SNAP_CHANNEL=stable" >>env.sh
+    echo "export SNAP_CHANNEL=stable,v0/stable" >>env.sh
   else
-    echo "export SNAP_CHANNEL=beta" >>env.sh
+    echo "export SNAP_CHANNEL=beta,v0/beta" >>env.sh
```
efiop, line 58: I guess we could also parse the major version and push to `vN` channel?

casperdcl: that is a good point. properly future-proof. Will change.

casperdcl: done

dvc/updater.py
```diff
@@ -48,7 +48,35 @@ def _with_lock(self, func, action):
...
+            # where SNAP_CHANNEL is:
+            #   snap info dvc | grep -E '^tracking:' | awk '{print $2}'
+            # However a quicker way is to skip if a user-configurable variable
+            # `skip_update_check` is set
+            import subprocess
+
+            kwargs = {"close_fds": True}
+            if os.name == "nt":
+                kwargs["shell"] = True
+                cmd = []
```
efiop, line 65: Not sure I get why we do anything at all on windows.

casperdcl: ah good point - I just copy-pasted this from `stage.run`

scripts/ci/before_install.sh
```diff
@@ -51,12 +51,18 @@ elif [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
...
-    echo "export SNAP_CHANNEL=stable" >>env.sh
+    echo "export SNAP_CHANNEL=stable,v$TAG_MAJOR/stable" >>env.sh
   else
-    echo "export SNAP_CHANNEL=beta" >>env.sh
+    echo "export SNAP_CHANNEL=beta,v$TAG_MAJOR/beta" >>env.sh
   fi
 else
-  echo "export SNAP_CHANNEL=edge" >>env.sh
+  echo "export SNAP_CHANNEL=edge,v$TAG_MAJOR/edge" >>env.sh
 fi
```
efiop, line 68: Oops, missed that this is only for 0.94-dev branch. I guess we need to port to the master too.

dvc/updater.py
```diff
@@ -48,7 +48,35 @@ def _with_lock(self, func, action):
...
+            else:
+                kwargs["shell"] = False
+                cmd = ["/bin/sh", "-c"]
+            p = subprocess.Popen(
+                cmd + ["snapctl", "get", "skip_update_check"], **kwargs
```
efiop, line 70: Not a fan of creating processes on `updater`.

scripts/ci/before_install.sh
casperdcl: yes, will do later

dvc/updater.py
efiop: Also, this runs on every dvc command in the same process, so this will turn ugly slow.

casperdcl: yes I don't like it either but I don't see any other way apart from having two different 0.x releases: one which pushes to `v0/stable` without any warning logic, and one which for the next 30 days pushes to `latest/stable` with the warning but none of these `subprocess` checks.

efiop: Maybe it is not worth having a special notification. YOu've created separate channels that will make sure everything is setup properly in the future, so we could afford some people automatically upgrading to 1.0 here.

casperdcl: I don't know - based on discord it looked like at least one person had errors from an automatic upgrade before we rolled back.

efiop:
@casperdcl What's the status on this PR?

dvc/updater.py
casperdcl: removed

casperdcl:
I've thought long and hard about this - the only reservation I have is that the hardcoded warning will display even if users run `refresh --channel=v0/stable`. Ideally the warning should be suppressed in that case, but doing that would require having a different branch/tag pushing to `v0/stable` without the hardcoded warning.

dvc/updater.py
```diff
@@ -138,6 +148,17 @@ def _get_update_instructions(self):
...
+            "snap": (
+                "To upgrade to the latest major release,\n"
+                "run `{yellow}snap{reset} refresh --channel=latest/beta`, or\n"
+                "to stay on the current major release track,\n"
+                "run `{yellow}snap{reset} refresh --channel={current}/stable`"
+                "\n\n"
+                "{red}WARNING{reset}: ignoring this message will result in\n"
+                "snap automatically performing an upgrade soon.\n"
+                "More information can be found at\n"
+                "{blue}https://github.com/iterative/dvc/issues/3872{reset}"
```
casperdcl, line 160: Ideally need a link to migration/braking changes on dvc.org?

efiop:
@casperdcl Maybe let's not bother with the warning at all? Running snap from dvc is a definite no-go.

scripts/build_snap.sh
```diff
@@ -7,7 +7,14 @@ if [ ! -d "dvc" ]; then
   exit 1
 fi
 
-sudo snapcraft --use-lxd
+sg lxd -c snapcraft
+original_snap_name="$(ls dvc_*.snap)"
+mv "$original_snap_name" original.snap
+
+git apply scripts/remove_update_warning.patch
```
efiop, line 14: This will make `dvc --version` show up as dirty

casperdcl:
yes nothing's run each command - it's all hard-coded now for snap releases:

- no warning for `v0` channel
- warn for 0.x on `latest` channel

scripts/build_snap.sh
casperdcl: yes; I don't think that's an issue (`dvc-0.95.0-1+gs0m3sha`).

casperdcl:
@efiop this should be fine to merge now. The two unchecked boxes are optional follow-up items, really.

PR #4000, opened 2020-06-09
config/remote/cache: make --system/--global/--local mutually exclusive

efiop:

https://github.com/iterative/dvc.org/pull/1416

### DIFF
--- a/dvc/command/config.py
+++ b/dvc/command/config.py
@@ -50,21 +50,22 @@ class CmdConfig(CmdBaseNoRepo):
 
 
 parent_config_parser = argparse.ArgumentParser(add_help=False)
-parent_config_parser.add_argument(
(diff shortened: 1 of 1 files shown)


efiop:
Not a biggie or anything, just noticed while talking with a user.

shcheklein:
add a test?

efiop:
@shcheklein Was kinda resiting it, because it is declarative as is. But yeah, sure, will add one in a sec.

efiop:
Actually, after giving it another thought - will pass on this one, as it will require testing all of those commands (which is excessive) or unit testing specific parser, but it is not really guaranteed to be used. It is declarative as is and the potential bugs are pretty much non-existent (it will just overwrite with the latest flag), so it is not really worth the time in this case.

PR #4004, opened 2020-06-10
add/run/import-url: don't use -f as --file

efiop:
If anything, `-f` is closer to `--force` than to `--file`. So let's set
this straight.


jorgeorpinel:
Hi. I see a checkbox was added for this in iterative/dvc.org/issues/1255, thanks. But that's not really a catch all epic for every change from now on 😋

I may be wrong but this doesn't sound like a 1.0-specific change to me, can we open a separate issue or PR instead, please?

efiop:
@jorgeorpinel it is used all over the docs and I have no idea what to change there because of all of the 1.0 PRs hanging that affect all of the docs that I need to change. So I've added it to the epic.

jorgeorpinel:
Please open an issue and link to the core PR when submitting a PR is not possible. That's totally fine. Done: iterative/dvc.org/issues/1426

efiop:
@jorgeorpinel It was added to the epic ` get rid of -f in add/run/import(-url) per iterative/dvc#4004 and iterative/dvc#4011`, you can see it in the history of changes, but you've deleted it in the past two days.

jorgeorpinel:
I extracted it into iterative/dvc.org#1426 like I just mentioned. Thanks

PR #4005, opened 2020-06-10
RepoTree: work with dirty repos

### LINKED ISSUE #3974: make sure metrics/plots can compare dirty repo with the last commit by default
>  Which versions get compared by default? without sending arguments
> I expected the present metrics file to be plotted on top of the last commit but if I manually change this metrics file, the changes are not reflected in the plot.
>  Only if I run dvc repro in a way that generates a different metrics file does it pick up the changes and plots them.
>  Where is this version coming from? I haven't committed anything to Git.
>  On DVC 1.0

https://discordapp.com/channels/485586884165107732/563406153334128681/719342575835676733

efiop:
Currently `RepoTree.open` will throw FileNotFoundError if you have a
file that is tracked by dvc, but which has no md5 or a cache for which
is missing, even though you have the file itself in your workspace.

This corresponds very nicely to the `--no-commit` or `--no-exec`
scenario, where you might not want to commit your artifacts to cache
just yet.

Fixes #3974

### DIFF
--- a/dvc/repo/tree.py
+++ b/dvc/repo/tree.py
@@ -61,6 +61,10 @@ class DvcTree(BaseTree):
         except OutputNotFoundError as exc:
             raise FileNotFoundError from exc
 
+        # NOTE: this handles both dirty and checkout-ed out at the same time
+        if self.repo.tree.exists(path):
+            return self.repo.tree.open(path, mode=mode, encoding=encoding)
+
         if len(outs) != 1 or (
             outs[0].is_dir_checksum and path == outs[0].path_info
         ):
@@ -252,13 +256,9 @@ class RepoTree(BaseTree):
             encoding = None
 
         if self.dvctree and self.dvctree.exists(path):
-            try:
-                return self.dvctree.open(
-                    path, mode=mode, encoding=encoding, **kwargs
-                )
-            except FileNotFoundError:
-                if self.isdvc(path):
-                    raise
+            return self.dvctree.open(
+                path, mode=mode, encoding=encoding, **kwargs
+            )
         return self.repo.tree.open(path, mode=mode, encoding=encoding)
 
     def exists(self, path):
--- a/tests/func/metrics/test_diff.py
+++ b/tests/func/metrics/test_diff.py
@@ -153,3 +153,20 @@ def test_no_commits(tmp_dir):
     assert Git().no_commits
 
     assert Repo.init().metrics.diff() == {}
+
+
+def test_metrics_diff_dirty(tmp_dir, scm, dvc, run_copy_metrics):
+    def _gen(val):
+        tmp_dir.gen({"m_temp.yaml": str(val)})
+        run_copy_metrics("m_temp.yaml", "m.yaml", metrics=["m.yaml"])
+        dvc.scm.commit(str(val))
+
+    _gen(1)
+    _gen(2)
+    _gen(3)
+
+    tmp_dir.gen({"m.yaml": "4"})
+
+    expected = {"m.yaml": {"": {"old": 3, "new": 4, "diff": 1}}}
+
+    assert dvc.metrics.diff() == expected
--- a/tests/func/test_api.py
+++ b/tests/func/test_api.py
@@ -114,6 +114,10 @@ def test_missing(remote_url, tmp_dir, dvc):
     # Remove cache to make foo missing
     remove(dvc.cache.local.cache_dir)
 
+    api.read("foo")
+
+    remove("foo")
+
     with pytest.raises(FileMissingError):
         api.read("foo")
 
--- a/tests/func/test_update.py
+++ b/tests/func/test_update.py
@@ -52,9 +52,9 @@ def test_update_import_after_remote_updates_to_dvc(tmp_dir, dvc, erepo_dir):
     new_rev = None
     with erepo_dir.branch("branch", new=False), erepo_dir.chdir():
         erepo_dir.scm.repo.index.remove(["version"])
-        erepo_dir.dvc_gen("version", "updated")
-        erepo_dir.scm.add(["version", "version.dvc"])
-        erepo_dir.scm.commit("upgrade to DVC tracking")
+        erepo_dir.dvc_gen(
+            "version", "updated", commit="upgrade to DVC tracking"
+        )
         new_rev = erepo_dir.scm.get_rev()
 
     assert old_rev != new_rev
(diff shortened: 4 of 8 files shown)


efiop:
The failing test might be related to https://github.com/iterative/dvc/pull/3991 CC @pmrowla

tests/func/test_api.py
```diff
@@ -114,6 +114,10 @@ def test_missing(remote_url, tmp_dir, dvc):
     # Remove cache to make foo missing
     remove(dvc.cache.local.cache_dir)
 
+    api.read("foo")
```
efiop, line 117: There is no reason for this to not work. E.g. in a described --no-exec/--no-commit scenario.

tests/func/metrics/test_diff.py
```diff
@@ -153,3 +153,20 @@ def test_no_commits(tmp_dir):
     assert Git().no_commits
 
     assert Repo.init().metrics.diff() == {}
+
+
+def test_metrics_diff_dirty(tmp_dir, scm, dvc, run_copy_metrics):
```
efiop, line 158: Need to add a "dirty" test for plots too. Did test it manually though. Will add one later in this PR.

dvc/repo/tree.py
```diff
@@ -251,14 +251,10 @@ def open(self, path, mode="r", encoding="utf-8", **kwargs):
...
-                return self.dvctree.open(
-                    path, mode=mode, encoding=encoding, **kwargs
-                )
-            except FileNotFoundError:
-                if self.isdvc(path):
-                    raise
+        if self.dvctree and not self.repo.tree.exists(path):
+            return self.dvctree.open(
+                path, mode=mode, encoding=encoding, **kwargs
+            )
```
efiop, line 257: Was also thinking about moving this logic to `DvcTree.open`, but not quite sure. What do you think @pmrowla ?

pmrowla: I think moving this logic into `DvcTree.open` would work better. The `tests/func/test_update.py` test failure is because we need to always prefer `dvctree.open` over `repo.tree.open` for DVC outs.

For erepo import/get, `remote.save()` expects `dvctree.open` to fetch DVC outs into local cache as needed when the tree was initialized with `fetch=True`, so if we default to preferring `repo.tree.open` it will not do the fetch (it just calls builtin `open()` via CleanTree/WorkingTree)

If we wanted to keep this logic in `RepoTree` and not `DvcTree`, we would need something like
```py
        if self.dvctree and self.dvctree.exists(path):
            try:
                return self.dvctree.open(
                    path, mode=mode, encoding=encoding, **kwargs
                )
            except FileNotFoundError:
                pass
        return self.repo.tree.open(path)
```

But I think it would make more sense to handle this in `DvcTree.open` with something like
1. open cached version of dvc out when possible (same as existing `DvcTree.open` behavior)
2. try opening (dirty) working copy version of the dvc out if it exists as fallback
3. raise FileNotFoundError

efiop: For the record: discussed privately and looks like we do agree with the repo tree tests down below, that state that `dirty` version should actually be prefered over the cached one. We got a bit derailed by the failing test, but it might just be a bad test, I'll take a closer look.

dvc/remote/base.py
```diff
@@ -680,9 +680,7 @@ def _save_file(
                 with tree.open(path_info, mode="rb") as fobj:
                     # if tree has fetch enabled, DVC out will be fetched on
                     # open and we do not need to read/copy any data
-                    if not (
-                        tree.isdvc(path_info, strict=False) and tree.fetch
-                    ):
```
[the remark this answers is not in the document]
pmrowla: This looks kind of strange now, since we're calling `changed_cache` twice, before and after `tree.open`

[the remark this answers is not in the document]
pmrowla: I think that relying on this side effect the way it was originally written is correct, since we should only be calling `open` specifically when we need to download the cache file (which was checked when we call `changed_cache` before calling `open`)

With your changes to `DvcTree/RepoTree`, but without this change to `_save_file`, `tests/func/test_update.py::test_update_import_after_remote_updates_to_dvc` fails, but I think that is a bug in the test.

If I make this change in that test:
```diff
     with erepo_dir.branch("branch", new=False), erepo_dir.chdir():
         erepo_dir.scm.repo.index.remove(["version"])
         erepo_dir.dvc_gen("version", "updated")
-        erepo_dir.scm.add(["version", "version.dvc"])
+        erepo_dir.scm.add(["version.dvc"])
         erepo_dir.scm.commit("upgrade to DVC tracking")
         new_rev = erepo_dir.scm.get_rev()
```
it passes with the original `_save_file` implementation (but with your changes to prefer dirty version in DvcTree).

I think the issue is that `scm.add()` calls `git.index.add(<path>)` directly, which is not the same as `git add <path>`. It looks like `git.index.add` doesn't respect gitignores, which causes the dvc out `version` to also appear in the GitTree for that test.

In that test, when we `open("version")` to copy/fetch it into cache, `repo.tree.exists("version")` should be false, since it's a DVC out from our clean erepo clone tree, which does not have any dirty/modified copy of `version` at all. `version.dvc` should exist in the git tree, but not .gitignored DVC out file `version`.

I think the issue w/`scm.add()` not respecting .gitignore can be verified with
```py
def test_gittree_add_ignore(tmp_dir, scm):
    tmp_dir.gen({"foo": "foo", ".gitignore": "foo"})
    with tmp_dir.chdir():
        scm.add(["foo", ".gitignore"])
        scm.commit("init")
    tree = scm.get_tree("HEAD")
    assert not tree.exists(tmp_dir / "foo")
```
(test will fail in master)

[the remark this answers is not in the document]
efiop: Great explanation! You are absolutely right, we don't have to touch that line for that particular test right now.

The double `if changed_cache` is just our way of verifying the side-effect to tell if the cache was actually downloaded and there is nothing we need to do here.

Also, it seems like wiith the working tree we don't really go through that branch of execution (because of the `if tree`), so my change might not even affect it. So that also might be something that is just not abstracted yet.

I'll remove that change in favor of fixing the test itself, and will keep it in mind for the future. 

Thanks!

tests/func/test_update.py
```diff
@@ -52,9 +52,9 @@ def test_update_import_after_remote_updates_to_dvc(tmp_dir, dvc, erepo_dir):
     new_rev = None
     with erepo_dir.branch("branch", new=False), erepo_dir.chdir():
         erepo_dir.scm.repo.index.remove(["version"])
-        erepo_dir.dvc_gen("version", "updated")
```
efiop, line 55: Created https://github.com/iterative/dvc/issues/4015

Issue #1255, opened 2018-10-22
run: Warn about duplicated arguments


mroutis:
Summary
-------------
You can specify the same output more than one type without any hint.
It would be great to have at least a warning, helping the user with potential typos.

Expected behavior
------------------------
```bash
dvc run \
  --file double-output.dvc \
  --outs hello.txt \
  --outs hello.txt \
  'echo "hello" > hello.txt'

Warning: Output 'hello.txt' was specified more than one time
```

`cat double-output.dvc`:
```
cmd: echo "hello" > hello.txt
md5: ec93debb48f6f21a03f5dbe10fa0a6c2
outs:
- cache: true
  md5: b1946ac92492d2347c6235b4d2611184
  path: hello.txt
```

Actual behavior
--------------------
```bash
dvc run \
  --file double-output.dvc \
  --outs hello.txt \
  --outs hello.txt \
  'echo "hello" > hello.txt'
```

`cat double-output.dvc`:
```
cmd: echo "hello" > hello.txt
md5: ec93debb48f6f21a03f5dbe10fa0a6c2
outs:
- cache: true
  md5: b1946ac92492d2347c6235b4d2611184
  path: hello.txt
- cache: true
  md5: b1946ac92492d2347c6235b4d2611184
  path: hello.txt
```

System configuration
----------------------------
DVC version: `0.19.12`
Installation: `Arch Linux AUR package`

dixi3:
i'd like to work on this

PR #4011, opened 2020-06-10
run: use -f/--force to overwrite, fix msg

### LINKED ISSUE #4003: run: improve stage exists error message to hint how to overwrite (edit) it
Running `dvc run` again to edit an existing stage in the `dvc.yaml` is a very common case:

Would be nice to put a hint along with the error message that prints:

`ERROR: Stage 'prepare' already exists in 'dvc.yaml'. Add `--overwrite` option to edit the existing stage.`

instead of:

`ERROR: Stage 'prepare' already exists in 'dvc.yaml'.`

that it prints now.

skshetry:
Also, makes fixes for `--overwrite` rename to `--force`
and also remove `remove_with_prompt` which was only being
used with `force=True` anyway.

Fix #4003

I'll suggest this on `dvc.org`'s WIP PR.


jorgeorpinel:
Hi, how is this expressed in #1255 ? I'm not seeing it, thanks.

jorgeorpinel:
p.s. probably best to open an individual docs issue. That epic is for very specific 1.0 discrepancies. Let's try to keep the same process as always.

skshetry:
@jorgeorpinel, many of the commands were changed for 1.0. And, this is a very 1.0 discrepancy. And, it's not too important for `get-started`.

efiop:
@jorgeorpinel Same as with my `-f` change, there are too many conflicting docs to edit, so added a checkbox to that epic.

jorgeorpinel:
I don't see a checkbox for this in the epic actually. But can we please follow the regular process 🙏 sorry if the epic was confusing.

For this one I will just incorporate in iterative/dvc.org/pull/1420 since I'm just getting started. Thanks

shcheklein:
> For this one I will just incorporate in iterative/dvc.org/pull/1420

not a good idea I think, let's try to keep iterative/dvc.org/pull/1420 focused on `dvc run` only and do it faster, wdyt?

jorgeorpinel:
Sure. But this PR is now part of 1.0 pre-release `dvc run` and I'm just getting started with updating that cmd ref description and examples, so this behavior will already be included as I use the cmd to confirm everything. That's why it should be covered already in iterative/dvc.org/pull/1420

efiop:
> I don't see a checkbox for this in the epic actually.

There was: ` get rid of -f in add/run/import(-url) per iterative/dvc#4004 and iterative/dvc#4011`, you can see it in the history. Look like you've removed it in the past two days.

jorgeorpinel:
I extracted that other one you added into its own issue but it wasn't for this, unless I don't understand how "get rid of -f in run" per #4004 = "run: use -f/--force to overwrite" per this PR.

PR #4017, opened 2020-06-11
plots: fix a couple of bugs

Suor:
- running .plots.collect() not from repo root used to skip plot files
- make Plots.render() be more convenient to be called from outside


efiop:
@pared Could you take a post-merge look, please? @Suor needed this very urgently.

PR #4018, opened 2020-06-11
tests: cover run cache with functional tests

### LINKED ISSUE #4016: run cache: push/pull is broken
#### Setup:
```console
cd $(mktemp -d)
mkdir {remote,repo}
cd repo
dvc init --no-scm
echo "foo" > foo
dvc run -n copy-foo-bar -d foo -o bar "cp foo bar"
dvc remote add -d remote ../remote
```

Output:
```console
$ dvc push --run-cache -v
2020-06-11 12:50:21,102 DEBUG: fetched: [(3,)]
2020-06-11 12:50:21,103 DEBUG: fetched: [(3,)]
2020-06-11 12:50:21,103 ERROR: unexpected error - 'LocalRemote' object has no attribute 'upload'
  File "/home/saugat/repos/iterative/dvc/dvc/main.py", line 53, in main
    ret = cmd.run()
  File "/home/saugat/repos/iterative/dvc/dvc/command/data_sync.py", line 58, in run
    run_cache=self.args.run_cache,
  File "/home/saugat/repos/iterative/dvc/dvc/repo/__init__.py", line 35, in wrapper
  File "/home/saugat/repos/iterative/dvc/dvc/repo/push.py", line 17, in push
    used_run_cache = self.stage_cache.push(remote) if run_cache else []
  File "/home/saugat/repos/iterative/dvc/dvc/stage/cache.py", line 180, in push
    return self._transfer(remote.upload, self.repo.cache.local, remote)
AttributeError: 'LocalRemote' object has no attribute 'upload'
```

**Output of `dvc version`:**

```console
$ dvc version
DVC version: 1.0.0a10+4042d5
Python version: 3.6.6
Platform: Linux-5.5.2-arch1-1-x86_64-with-arch
Binary: False
Package: None
Filesystem type (cache directory): ('tmpfs', 'tmpfs')
Repo: dvc (no_scm)
Filesystem type (workspace): ('tmpfs', 'tmpfs')
```

cc @pmrowla

skshetry:
Also, I tried to fix #4016, and also fix discrepancies in push/fetch counts.

### DIFF
--- a/dvc/repo/fetch.py
+++ b/dvc/repo/fetch.py
@@ -72,7 +72,7 @@ def _fetch(
     if failed:
         raise DownloadError(failed)
 
-    return downloaded
+    return downloaded + len(used_run_cache)
 
 
 def _fetch_external(self, repo_url, repo_rev, files, jobs):
--- a/dvc/stage/cache.py
+++ b/dvc/stage/cache.py
@@ -177,11 +177,15 @@ class StageCache:
 
     def push(self, remote):
         remote = self.repo.cloud.get_remote(remote)
-        return self._transfer(remote.upload, self.repo.cache.local, remote)
+        return self._transfer(
+            remote.tree.upload, self.repo.cache.local.tree, remote.tree
+        )
 
     def pull(self, remote):
         remote = self.repo.cloud.get_remote(remote)
-        return self._transfer(remote.download, remote, self.repo.cache.local)
+        return self._transfer(
+            remote.tree.download, remote.tree, self.repo.cache.local.tree
+        )
 
     def get_used_cache(self, used_run_cache, *args, **kwargs):
         from dvc.cache import NamedCache
--- a/dvc/stage/cache.py
+++ b/dvc/stage/cache.py
@@ -6,6 +6,7 @@ import yaml
 from funcy import first
 from voluptuous import Invalid
 
+from dvc.remote.local import _log_exceptions
 from dvc.schema import COMPILED_LOCK_FILE_STAGE_SCHEMA
 from dvc.utils import dict_sha256, relpath
 from dvc.utils.fs import makedirs
@@ -178,13 +179,17 @@ class StageCache:
     def push(self, remote):
         remote = self.repo.cloud.get_remote(remote)
         return self._transfer(
-            remote.tree.upload, self.repo.cache.local.tree, remote.tree
+            _log_exceptions(remote.tree.upload, "upload"),
+            self.repo.cache.local.tree,
+            remote.tree,
         )
 
     def pull(self, remote):
         remote = self.repo.cloud.get_remote(remote)
         return self._transfer(
-            remote.tree.download, remote.tree, self.repo.cache.local.tree
+            _log_exceptions(remote.tree.download, "download"),
+            remote.tree,
+            self.repo.cache.local.tree,
         )
 
     def get_used_cache(self, used_run_cache, *args, **kwargs):
--- a/tests/func/test_run_cache.py
+++ b/tests/func/test_run_cache.py
@@ -1,0 +1,153 @@
+import os
+
+from dvc.dvcfile import PIPELINE_LOCK
+from dvc.utils import relpath
+
+
+def _recurse_count_files(path):
+    return len([os.path.join(r, f) for r, _, fs in os.walk(path) for f in fs])
+
+
+def test_push_pull(tmp_dir, dvc, erepo_dir, run_copy, setup_remote):
+    tmp_dir.gen("foo", "foo")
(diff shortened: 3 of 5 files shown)


dvc/repo/fetch.py
```diff
@@ -72,7 +72,7 @@ def _fetch(
     if failed:
         raise DownloadError(failed)
 
-    return downloaded
+    return downloaded + len(used_run_cache)
```
skshetry, line 75: `fetch` and `push/pull` for run-cache did not add up counts. Was this intentional, @efiop?

dvc/stage/cache.py
```diff
@@ -177,11 +177,15 @@ def _transfer(func, from_remote, to_remote):
 
     def push(self, remote):
         remote = self.repo.cloud.get_remote(remote)
-        return self._transfer(remote.upload, self.repo.cache.local, remote)
+        return self._transfer(
+            remote.tree.upload, self.repo.cache.local.tree, remote.tree
+        )
```
skshetry, line 182: Just changed it to `.tree` instead of changing everywhere in `_transfer()`.

tests/func/test_run_cache.py
```diff
@@ -0,0 +1,153 @@
...
+
+from dvc.dvcfile import PIPELINE_LOCK
+from dvc.utils import relpath
+
+
+def _recurse_count_files(path):
+    return len([os.path.join(r, f) for r, _, fs in os.walk(path) for f in fs])
+
+
+def test_push_pull(tmp_dir, dvc, erepo_dir, run_copy, setup_remote):
```
skshetry, line 11: The tests that are missing now are either low-level tests (which I don't really want to add), and more-higher level tests (eg: that covers CML like use-cases where you push to remote storage and remote branch, get contents to a different machine and try to repro or run).

dvc/repo/fetch.py
efiop: @skshetry to be honest, `--run-cache` was introduced asap as a hack to cover a specific use case, so I didn't add it to the count, hoping that we might find a better way (e.g. separate command). But it makes total sense to add it up, so let's keep it. :+1:

dvc/stage/cache.py
efiop: @skshetry `tree.upload/download` are not meant to swallow errors, unlike the remote, so it won't work :slightly_frowning_face:

skshetry: @efiop, so, what do you propose? I'm still wrapping my heads around tree. Is there any other interface, or are you proposing to accumulate the errors and continue and show error later?

efiop: @skshetry We just need to wrap those with `_log_exceptions` helper from `local.py`. That will do the trick for now.

dvc/repo/fetch.py
skshetry: @efiop, as far as I understand, `downloaded` should already have a total count, right? As we pass `used_run_cache` already.

efiop: @skshetry Correct. So indeed `+ len(used_run_cache)` is not needed here for now.

skshetry: But, that's not working at all. See the tests.

efiop: @skshetry Ok, the confusion here is because run cache consists of .dvc/cache/runs and the regular cache files assotiated with it. So `used_run_cache` is a list of run cache entries and regular `downloaded` accounts for regular cache files that include the ones used by used_run_cache. So `+ len(used_run_cache)` is correct if we want to account for run-cache entries and not only for the regular cache entries.

dvc/stage/cache.py
```diff
@@ -177,11 +178,19 @@ def _transfer(func, from_remote, to_remote):
 
     def push(self, remote):
         remote = self.repo.cloud.get_remote(remote)
-        return self._transfer(remote.upload, self.repo.cache.local, remote)
+        return self._transfer(
+            _log_exceptions(remote.tree.upload, "upload"),
```
efiop, line 182: Added this without moving it to utils or anything. This is a temporary hack, our tree/remote/cache separation will soon allow us to properly handle this. Ideally, we would want to make remote handle run-cache by itself, since it is really not that much different from regular cache, just that the location is unusual.

dvc/repo/fetch.py
skshetry: Ahh, okay I got it. It's already transfer via `RunCache`. Thanks for the explanation.

dvc/stage/cache.py
skshetry: Thanks for the help.

PR #4019, opened 2020-06-11
remote: separate cloud remote and cloud cache classes

pmrowla:

Related to #3882

* Supported cloud types now have distinct remote and cache classes (i.e. `S3Remote`, `S3Cache`, `LocalRemote`, `LocalCache`, etc)
* All `get_checksum` related functions are now moved into remote trees
* `cache.save()` now takes an explicit `tree` parameter
    * When `tree` is `remote.tree`, save is done via `move()` + `link()`
    * For all other trees, save is done by copying object from `tree` into `remote.tree`

### DIFF
--- a/dvc/remote/__init__.py
+++ b/dvc/remote/__init__.py
@@ -1,16 +1,25 @@
 import posixpath
 from urllib.parse import urlparse
 
-from dvc.remote.azure import AzureRemote
+from dvc.remote.azure import AzureCache, AzureRemote
 from dvc.remote.gdrive import GDriveRemote
-from dvc.remote.gs import GSRemote
-from dvc.remote.hdfs import HDFSRemote
+from dvc.remote.gs import GSCache, GSRemote
+from dvc.remote.hdfs import HDFSCache, HDFSRemote
 from dvc.remote.http import HTTPRemote
 from dvc.remote.https import HTTPSRemote
-from dvc.remote.local import LocalRemote
+from dvc.remote.local import LocalCache, LocalRemote
 from dvc.remote.oss import OSSRemote
-from dvc.remote.s3 import S3Remote
-from dvc.remote.ssh import SSHRemote
+from dvc.remote.s3 import S3Cache, S3Remote
+from dvc.remote.ssh import SSHCache, SSHRemote
+
+CACHES = [
+    AzureCache,
+    GSCache,
+    HDFSCache,
+    S3Cache,
+    SSHCache,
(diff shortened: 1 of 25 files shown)


dvc/remote/local.py
```diff
@@ -257,26 +262,119 @@ def wrapper(from_info, to_info, *args, **kwargs):
...
 
-    def __init__(self, repo, config):
-        super().__init__(repo, config)
-        self.cache_dir = config.get("url")
-
     @property
     def state(self):
         return self.repo.state
 
+    def get(self, md5):
```
pmrowla, line 280: Is this function needed? It looks like it was leftover from some old behavior, and is local only (not found in `BaseRemote`). It's still being tested in `tests/func/test_data_cloud` but I couldn't find anywhere else in DVC that it's actually used.

dvc/remote/base.py
```diff
@@ -389,188 +530,400 @@ def supported(cls, config):
...
+            removed = True
+        if removed:
+            self.index.clear()
+        return removed
+
+    def _remove_unpacked_dir(self, checksum):
+        pass
+
+
+class CacheMixin:
```
pmrowla, line 879: Cloud cache is technically not an extension to cloud remote, since we define remote as the cloud push/pull mirror for a local .dvc MD5 checksum-based cache. However, cloud caches still need to implement the remote methods for size estimation and garbage collection (so that we can display progress during `gc`).

Originally, I had separate `BaseCloud`, `RemoteMixin`, `CacheMixin`, classes, but it seemed strange to have it like that if every cache class was always going to include the `RemoteMixin` methods. So for now all the cloud cache classes are implemented as an extension of a cloud remote class + the cache mixin.

pmrowla:
Will need rebase after #3991 is merged to resolve `test_update_import_after_remote_updates_to_dvc` test failure (due to a State issue)

dvc/remote/__init__.py
```diff
@@ -26,21 +35,30 @@
 ]
 
 
-def _get(remote_conf):
-    for remote in REMOTES:
+def _get(remote_conf, remotes, default):
```
pmrowla, line 38: @efiop I think cleaning this up to use `tree.supported(...)` will also require moving scheme and the remaining constants into the tree classes like we discussed. And then once we have single `Remote` and `Cache` classes that take a tree as a parameter, we'll be able to get rid of these helper methods.

I think it would be best to leave this as-is for now, and then handle it in the next PR

Issue #4014, opened 2020-06-11
DVC 1.0 QA results, part 2


shcheklein:
```
DVC version: 1.0.0a10+4042d5
Python version: 3.7.7
Platform: Darwin-19.4.0-x86_64-i386-64bit
Binary: False
Package: None
Supported remotes: azure, gdrive, gs, hdfs, http, https, s3, ssh, oss
Repo: dvc, git
Filesystem type (workspace): ('apfs', '/dev/disk1s1')
```

No new _and_ major findings this time.

- [x] `dvc run -f` (ex `--overwrite`) changes order of stages in the file (#4035) -https://gist.github.com/shcheklein/30103d6fcb9d58e8be85e0be0f1b22a0
- [x] We should probably include newline at the end of the `dvc.yaml`. A lot of editors do this automatically meaning that we change hash of it (?) Having a new line at the end is a good practices for code files.
- [x] 💅 ~~no dot at the end in the `dvc --help` for unprotect~~ Too tedious to make it consistent everywhere.
- [x] 💅 ~~`or between a commit and the workspace.` is included for some `diff` commands, not for others in `dvc --help`~~
- [x] ~~check and update autocompletion scripts - we've done a lot of renaming/changes recently~~ (#3921)

A bunch of issues with plots, but I think we already have tickets for them or at least aware. A few other issues have been fixed yesterday ⚡

skshetry:
> dvc run -f (ex --overwrite) changes order of stages in the file

I would not like to spend time in preserving order on `-f`. Is there any strong reasons to?
I understand that by preserving order, we will be able to have a better diff and preserve comments + meta (which might also mean, accidentally it could copy unnecessary leftovers from previous entry as well, but my quick test did not show such behavior though). But, promising order  is a slippery slope. With `-f`, you can provide any options which essentially makes it a new stage (only the `name` is shared at minimum).

Again, by using `-f`, you are creating a new stage. So, it's better to treat it as such and append at the end.

> We should probably include newline at the end of the dvc.yaml

@shcheklein, can you double check this? This should already be the case.
Try with:
```console
tail -c1 dvc.yaml
```

shcheklein:
> Is there any strong reasons to?

`-f` is being used to _edit_ stages, not create new ones. If we allow _editing_  the stages we should be doing this w/o messing user defined order.

Alternative here - don't allow `-f`.

> But, promising order is a slippery slope. 

Not sure I understand this. You don't promise any oder - you just keep it.

[...]

PR #4035, opened 2020-06-12
run: maintain order when -f/--force overwrite stage

skshetry:
This commit preserves comments, meta and order inside a dvc.yaml
when `-f/--force` is used. But, it does not maintain comments/meta
for .dvc files.

Related: #4014

### DIFF
--- a/dvc/repo/run.py
+++ b/dvc/repo/run.py
@@ -1,4 +1,5 @@
 import os
+from contextlib import suppress
 
 from funcy import concat, first, lfilter
 
@@ -109,12 +110,12 @@ def run(self, fname=None, no_exec=False, single_stage=False, **kwargs):
         return None
 
     dvcfile = Dvcfile(self, stage.path)
-    if kwargs.get("force", True):
-        dvcfile.remove_stage(stage)
-    else:
-        _check_stage_exists(dvcfile, stage)
-
     try:
+        if kwargs.get("force", True):
+            with suppress(ValueError):
+                self.stages.remove(stage)
+        else:
+            _check_stage_exists(dvcfile, stage)
         self.check_modified_graph([stage])
     except OutputDuplicationError as exc:
         raise OutputDuplicationError(exc.output, set(exc.stages) - {stage})
--- a/tests/func/test_run_multistage.py
+++ b/tests/func/test_run_multistage.py
@@ -1,4 +1,5 @@
 import os
+import textwrap
 
 import pytest
 import yaml
@@ -6,6 +7,7 @@ import yaml
 from dvc.exceptions import InvalidArgumentError
 from dvc.repo import Repo
 from dvc.stage.exceptions import DuplicateStageName, InvalidStageName
+from dvc.utils.yaml import parse_yaml_for_update
 
 
 def test_run_with_name(tmp_dir, dvc, run_copy):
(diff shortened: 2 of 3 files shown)


tests/func/test_run_multistage.py
```diff
@@ -315,3 +318,43 @@ def test_run_without_cmd(kwargs):
...
         Repo().run(**kwargs)
     assert "command is not specified" == str(exc.value)
+
+
+def test_run_overwrite_order(tmp_dir, dvc, run_copy):
+    tmp_dir.gen({"foo": "foo", "foo1": "foo1"})
+    run_copy("foo", "bar", name="copy-foo-bar")
+    run_copy("bar", "foobar", name="copy-bar-foobar")
+
+    run_copy("foo1", "bar1", name="copy-foo-bar", force=True)
```
skshetry, line 328: don't know if the `force` being default was intentional on Python API, maybe for ease of use?
Anyway, a lot of tests depend of `force=True`.

tests/func/test_run_multistage.py
```diff
@@ -315,3 +318,43 @@ def test_run_without_cmd(kwargs):
...
+
+    data = parse_yaml_for_update(
+        (tmp_dir / PIPELINE_FILE).read_text(), PIPELINE_FILE
+    )
+    assert list(data["stages"].keys()) == ["copy-foo-bar", "copy-bar-foobar"]
+
+
+def test_run_overwrite_preserves_meta_and_comment(tmp_dir, dvc, run_copy):
```
skshetry, line 336: Is this a feature? I am not sure. Preserving this took less effort

tests/func/test_run_single_stage.py
```diff
@@ -1017,3 +1018,32 @@ def test_metrics_dir_not_cached(self, dvc):
                 metrics_no_cache=["dir"],
                 single_stage=True,
             )
+
+
+def test_run_force_doesnot_preserve_comments_and_meta(tmp_dir, dvc, run_copy):
```
skshetry, line 1023: Took a lot of time to find out why this was not the case. It's simple though: we never load `stage_text` on `run` and hence, when `dump`-ing, `to_single_stagefile` has no idea of `meta` or comments. Anyway, the PR is focused on order for `dvc.yaml`. I just added this test, if someone gets confused later.

dvc/repo/run.py
```diff
@@ -109,12 +110,12 @@ def run(self, fname=None, no_exec=False, single_stage=False, **kwargs):
...
-        dvcfile.remove_stage(stage)
-    else:
-        _check_stage_exists(dvcfile, stage)
-
     try:
+        if kwargs.get("force", True):
+            with suppress(ValueError):
+                self.stages.remove(stage)
+        else:
+            _check_stage_exists(dvcfile, stage)
```
skshetry, line 118: So, a bit more details on the implementation:

We check the graph for duplications. But, `check_modified_graph` appends the newly created `stage` and `self.stages` (which might contain the stage that we are trying to overwrite). For `-f`, we will be overwriting it anyway, so we can do
graph checks by removing the stage from self.stages and later appending the new stage.

 For `dvc.yaml`, we do a diff later when dumping, so, this preserves comments and meta.
For `.dvc` files, the `dvc run` never loads the file and therefore, does not set `_stage_text`, and hence, the `dump` of single stage file is never aware of `meta` and comments present before, so, it just overwrites them.

skshetry: This process is however slow, because there's clearly a `O(n)` for going throught the list of `self.stages`, and equality comparisons might not be faster (considering the findings with `relpath` :smile:).

Note, the equality of stages for `dvc.yaml` stages is the path relative to root of the repo and name of the stage whereas for `.dvc` files, it's just path rel. to the root.

tests/func/test_run_multistage.py
efiop: @skshetry Just noticing this myself as well and I think it is a bug, actually.

tests/func/test_run_multistage.py
efiop: @skshetry comments are indeed great to preserve, but meta indeed no longer seems very relevant. But at the same time I'm not sure if anyone is using it. But since we support it already - let's keep it for now.

PR #4037, opened 2020-06-13
CI: fix Mac builds

casperdcl:
- mac doesn't like `sed -r` (for `snapcraft` version detection)
- the logic is not needed anyway since `snapcraft` runs on linux
- continues #3999

Best to [view the diff ignoring whitespace](https://github.com/iterative/dvc/pull/4037/files?w=1)


casperdcl:
@efiop should be fine to merge (or could hijack this PR to fix the single failing SSH test)

PR #4041, opened 2020-06-15
remote: continue remote/cache/tree refactoring

pmrowla:

Related to #3882.

* Replaces `remote.Remote` and `remote.Cache` helper functions with actual `Remote` and `CloudCache` classes
* `Remote` and `CloudCache` constructors take a RemoteTree parameter
* `remote.get_cloud_tree` helper method can be used to return the proper tree for a given URL
* output/dependency/cache updated to use the new functionality

### DIFF
--- a/dvc/remote/__init__.py
+++ b/dvc/remote/__init__.py
@@ -1,45 +1,36 @@
 import posixpath
 from urllib.parse import urlparse
 
-from dvc.remote.azure import AzureCache, AzureRemote
-from dvc.remote.gdrive import GDriveRemote
-from dvc.remote.gs import GSCache, GSRemote
-from dvc.remote.hdfs import HDFSCache, HDFSRemote
-from dvc.remote.http import HTTPRemote
-from dvc.remote.https import HTTPSRemote
-from dvc.remote.local import LocalCache, LocalRemote
-from dvc.remote.oss import OSSRemote
-from dvc.remote.s3 import S3Cache, S3Remote
-from dvc.remote.ssh import SSHCache, SSHRemote
+from dvc.remote.azure import AzureRemoteTree
+from dvc.remote.gdrive import GDriveRemoteTree
+from dvc.remote.gs import GSRemoteTree
+from dvc.remote.hdfs import HDFSRemoteTree
+from dvc.remote.http import HTTPRemoteTree
+from dvc.remote.https import HTTPSRemoteTree
+from dvc.remote.local import LocalRemoteTree
+from dvc.remote.oss import OSSRemoteTree
+from dvc.remote.s3 import S3RemoteTree
+from dvc.remote.ssh import SSHRemoteTree
 
-CACHES = [
-    AzureCache,
-    GSCache,
-    HDFSCache,
-    S3Cache,
-    SSHCache,
-    # LocalCache is the default
-]
-
-REMOTES = [
-    AzureRemote,
-    GDriveRemote,
-    GSRemote,
-    HDFSRemote,
-    HTTPRemote,
-    HTTPSRemote,
-    S3Remote,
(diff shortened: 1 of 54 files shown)


dvc/remote/base.py
```diff
@@ -849,44 +702,243 @@ def exists_with_progress(path_info):
...
-            self.index.clear()
+
+        if removed and hasattr(remote, "index"):
+            remote.index.clear()
         return removed
 
-    def _remove_unpacked_dir(self, checksum):
-        pass
 
+class CloudCache:
```
efiop, line 893: Quite strange why it is CloudCache and not just Cache? And why LocalCache is the only exception?

efiop: Ah, I guess because of the push/pull. Ok, makes sense.

dvc/remote/__init__.py
```diff
@@ -90,3 +72,17 @@ def _resolve_remote_refs(config, remote_conf):
...
+    return _get_tree(remote_conf)(repo, remote_conf)
+
+
+def get_remote(repo, **kwargs):
+    tree = get_cloud_tree(repo, **kwargs)
+    if tree.scheme == "local":
+        return LocalRemote(tree)
+    if tree.scheme == "ssh":
+        return SSHRemote(tree)
+    return Remote(tree)
```
pmrowla, line 88: For initializing caches, we now just do
```
tree = get_cloud_tree(<remote config>)
cache = CloudCache(tree)
```

But for remotes, we still have this helper method. It would be ideal if we could just always use `Remote(tree)` but `LocalRemote` and `SSHRemote` still have their own overridden batch exists methods. It may be worth investigating whether or not the custom ssh/local methods are still needed with the other performance improvements that we have now.

efiop: Or maybe those custom methods could be moved to the trees. IIRC ssh uses some custom sftp channel handling, which might be abstracted away too. But surely could do that later. Maybe consider creating an issue for it.

dvc/remote/base.py
pmrowla: I went with CloudCache to avoid the collision with `dvc.cache.Cache`, but maybe this should be Cache, and the other one should be `RepoCache` or something along those lines.

efiop: @pmrowla Makes sense. I'm fine with CloudCache for now.

dvc/remote/__init__.py
pmrowla: One other question I had was whether or not the indexing related code should also just go into the trees at some point instead of being DVC Remote specific

efiop: @pmrowla Oh, that's a very good question! To me, it seems purely remote-related, as it solely relies on remote structure. But how do you see it?

dvc/remote/base.py
pmrowla: We can also get rid of LocalCache in the future by making push/pull/status more generic for syncing data between any Cache and any Remote, rather than having it be specific to LocalCache, but I don't think there's any pressing need for that right now.

dvc/remote/__init__.py
pmrowla: Since it's remote specific it makes sense to keep it in the Remote class for now, but I wasn't sure if indexing any kind of content (as opposed to just DVC Remotes) in a remote/cloud tree is something we might want to support in the future.

efiop: I'm not sure there is a good way to index arbitrary data in that way, it is really tailored for our  cache structure, so I really doubt that full index logic will ever make its way into the Tree. But, I'm sure that parts of it will, as many functions are actually quite generic (like walk_files replacing _list_cache).

PR #4049, opened 2020-06-15
get-url/import-url: save dependency before downloading

### LINKED ISSUE #3897: "AttributeError: NoneType object has no attribute 'wdir'" on dvc get-url call

efiop:
This allows us to properly process the errors like missing target.

Fixes #3897

### DIFF
--- a/dvc/output/local.py
+++ b/dvc/output/local.py
@@ -35,14 +35,14 @@ class LocalOutput(BaseOutput):
             # FIXME: if we have Windows path containing / or posix one with \
             # then we have #2059 bug and can't really handle that.
             p = self.TREE_CLS.PATH_CLS(path)
-            if not p.is_absolute():
+            if self.stage and not p.is_absolute():
                 p = self.stage.wdir / p
 
         abs_p = os.path.abspath(os.path.normpath(p))
         return self.TREE_CLS.PATH_CLS(abs_p)
 
     def __str__(self):
-        if not self.is_in_repo:
+        if not self.repo or not self.is_in_repo:
             return str(self.def_path)
 
         cur_dir = os.getcwd()
--- a/dvc/remote/local.py
+++ b/dvc/remote/local.py
@@ -56,7 +56,9 @@ class LocalRemoteTree(BaseRemoteTree):
 
     @property
     def state(self):
-        return self.repo.state
+        from dvc.state import StateNoop
+
+        return self.repo.state if self.repo else StateNoop()
 
     @cached_property
     def work_tree(self):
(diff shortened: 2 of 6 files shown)


dvc/output/local.py
```diff
@@ -35,14 +35,14 @@ def _parse_path(self, remote, path):
             # FIXME: if we have Windows path containing / or posix one with \
             # then we have #2059 bug and can't really handle that.
             p = self.TREE_CLS.PATH_CLS(path)
-            if not p.is_absolute():
+            if self.stage and not p.is_absolute():
```
skshetry, line 38: We should probably remove the dependency on `stage`, this is very much circular: `stage` needs outs/deps, and outs/deps require stages, so, when we construct a stage, there's no way beforehand to create outs/deps without creating a stage first without any deps/outs.

efiop: @skshetry you mean from the outputs/deps themselves? We still use it in some places, so easier said than done. Logically outs and deps do belong to the stage, but the way we create and fill them is not very friendly, I agree.

skshetry: What I am trying to say, is `Output` should not get access to the `Stage` at all. We could pass `wdir` and `repo` if they need it.

PR #4052, opened 2020-06-16
optimize: remove `fsync` when dumping rwlock data

skshetry:
Fixes #3653

#### Before:
Screenshot from 2020-06-16 20-00-43

### After:
Screenshot from 2020-06-16 20-02-49

(146s with 124.9s spent on fsync vs 14.8s cumulative)

Still 5-6 seconds are spent on 178461 `relpath()` calls. Will be handled on a separate PR.

* [x] ❗ I have followed the [Contributing to DVC](https://dvc.org/doc/user-guide/contributing/core) checklist.

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here. If the CLI API is changed, I have updated [tab completion scripts](https://github.com/iterative/dvc/tree/master/scripts/completion).

* [x] ❌ I will check DeepSource, CodeClimate, and other sanity checks below. (We consider them recommendatory and don't expect everything to be addressed. Please fix things that actually improve code or fix bugs.)

Thank you for the contribution - we'll try to review it as soon as possible. 🙏

Issue #3653, opened 2020-04-19
On Windows, dvc repro spends an awful lot of time in rwlock


charlesbaynham:
Windows 7 / 10, dvc 0.93.0+f5d89f (after merging #3645)

Running cProfile on `dvc repro -P` in a repository with around 200 steps all of which are up to date. On my system this takes 65s, of which 42s is spent in `_edit_rwlock`. 

I wonder, is it necessary to lock each individual file here? Couldn't the whole repository be locked instead? 

Maybe a broader question: what are the respective purposes of `.dvc/tmp/rwlock`, `.dvc/lock`and `.dvc/updater.lock`?

efiop:
> Running cProfile on dvc repro -P in a repository with around 200 steps all of which are up to date. On my system this takes 65s, of which 42s is spent in _edit_rwlock.

Thanks for the investigation! That is indeed pretty strange, we need to take a closer look. Could you show your `dvc version` output btw?

> I wonder, is it necessary to lock each individual file here? Couldn't the whole repository be locked instead?

Actually, that's what we used to do, but rwlock was introduced to support granular locking, so that you could run dvc in parallel.

> Maybe a broader question: what are the respective purposes of .dvc/tmp/rwlock, .dvc/lockand .dvc/updater.lock?

Useful link: https://dvc.org/doc/user-guide/dvc-files-and-directories  :slightly_smiling_face: `rwlock` is for granular locking of specific deps and outs, `lock` is for locking the whole repo (when you need to push/pull/etc or need to create/modify/delete dvc-file). `updater.lock` is for our updater worker, that checks for updates once a week.

efiop:
@charlesbaynham Btw, have you tried running that on Linux? Wonder if this is windows-specific or if we have troubles everywhere.

charlesbaynham:
Thanks for the run down @efiop, that's helpful. The version that ran the above was `0.93.0+f5d89f`, which is v0.93.0 + the changes in PR #3645. That's necessary because otherwise the effect is swamped by all the time spent reproducing duplicate steps. 

I'll give it a try on linux: annoyingly the repository I tested on has some case sensitivity problems which I've glossed over so far on Windows but which prevent it from working on linux, but I should fix that anyway. 

The cProfile results for my repo on Windows:
[repro_nowork.zip](https://github.com/iterative/dvc/files/4509341/repro_nowork.zip)

charlesbaynham:
[...]

PR #4053, opened 2020-06-16
cache relpath on stage

skshetry:
~~On top of #4052~~

A lot of `relpath` calls were being used by `networkx` indirectly through `Stage::__eq__()` and `Stage::__hash__()`. They would call `relpath` and `abspath` calls. This PR caches them.

Compare with #4052 with following:

Screenshot from 2020-06-16 20-26-29

Related #3653 

### DIFF
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -3,7 +3,7 @@ import os
 import string
 from collections import defaultdict
 
-from funcy import project
+from funcy import cached_property, project
 
 import dvc.dependency as dependency
 import dvc.prompt as prompt
@@ -127,6 +127,8 @@ class Stage(params.StageParams):
     @path.setter
     def path(self, path):
         self._path = path
+        self.__dict__.pop("path_in_repo", None)
+        self.__dict__.pop("relpath", None)
 
     @property
     def dvcfile(self):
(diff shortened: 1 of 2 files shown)


dvc/stage/__init__.py
```diff
@@ -172,11 +174,11 @@ def __eq__(self, other):
             and self.path_in_repo == other.path_in_repo
         )
 
-    @property
+    @cached_property
     def path_in_repo(self):
         return relpath(self.path, self.repo.root_dir)
 
-    @property
+    @cached_property
```
skshetry, line 181: There's a test `test_stage_strings_representation` that chdir to a different directory and checks for `relpath`. 

https://github.com/iterative/dvc/blob/a0c229cf9b9a33c1301c0ce0727a4d89fb73aaa1/tests/func/test_stage.py#L222

tests/func/test_stage.py
```diff
@@ -220,6 +220,10 @@ def test_stage_strings_representation(tmp_dir, dvc, run_copy):
     folder = tmp_dir / "dir"
     folder.mkdir()
     with folder.chdir():
+        # `Stage` caches `relpath` results, forcing it to reset
+        stage1.path = stage1.path
+        stage2.path = stage2.path
+
```
skshetry, line 226: Ugly hack, `chdir` is now not supported as we cache the `relpath`. It's fine for CLI use-case, but for API usecases, it sucks.

PR #4063, opened 2020-06-17
tests: create remote fixtures

efiop:
First steps in migrating to convenient cloud fixtures. External outs/deps tests and `put_object` tests are not converted for now, this PR is a pre-requisite. A lot of the stuff was just adapted for the new approach, remote helpers will be reworked some more in the next PR.


tests/func/test_api.py
```diff
@@ -5,93 +5,42 @@
...
+
+cloud_names = [
+    "s3",
+    "gs",
+    "azure",
+    "gdrive",
+    "oss",
+    "ssh",
+    "hdfs",
+    "http",
```
efiop, line 19: Also added http api test.

tests/func/test_api.py
```diff
@@ -5,93 +5,42 @@
...
-        GDrive.create_dir(dvc, url)
-        with dvc.config.edit() as conf:
-            conf["remote"][TEST_REMOTE].update(
-                gdrive_service_account_email="test",
-                gdrive_service_account_p12_file_path="test.p12",
-                gdrive_use_service_account=True,
-            )
-        dvc.scm.add(dvc.config.files["repo"])
-        dvc.scm.commit(f"modify '{TEST_REMOTE}' remote")
-
```
efiop, line 75: Finally this moved into the fixture itself. But the fixture itself could be  and will be improved further later.

skshetry:
@efiop, this looks good than before, though I don't like `TmpDir` getting more and more powerful. Feels like we are going the same old route (remember old `erepo_dir`?). 

What I'm saying is, I don't see any difference between:
```py
setup_remote(dvc, config=s3_config)
```

vs:

```python
tmp_dir.add_remote(config=s3_config)
```

efiop:
@skshetry Great point! What I really wanted is something like `dvc.remote.add`, but didn't want to expand the scope of this even further. It just feels like an overkill to have a separate fixture for that, especially since we no longer use it everywhere thanks to `local_remote` fixture, which is more intuitive. TmpDir will likely become a base class for working with external workspaces, just because it is is so convenient, and will get some additional properties/methods, which is not a bad thing as long as it is handy for writing the tests. I'll give it another look in the upcoming externa outs/deps PR.

Issue #4070, opened 2020-06-19
completion: misc updates


casperdcl:
Follow-up to #3921 

- [x] update `shtab` dependency (`>=1.0.0`, or likely `>=1.1.0`) #4076
- [x] hide/skip certain subcommands (e.g. `daemon`) #4076
  + `shtab>=1.0.0` https://github.com/iterative/dvc/pull/3921#discussion_r442787136
- [x] hide/skip certain arguments & options #4076
  + `shtab>=1.0.0` <- https://github.com/iterative/shtab/issues/3
- [ ] introduce consistency calling DVC-files `targets`, other files `paths`, and directories `dirs`
- [ ] introduce consistency around leading vs trailing whitespace when linebreaking strings

casperdcl:
based on the remaining checkboxes, this is no longer a `completion` issue (the completion issue is fixed). This is now a documentation issue - introducing naming consistency.

efiop:
Let's close for now.

PR #4076, opened 2020-06-21
completion: misc fixes

casperdcl:
- [x] update `shtab` dependency (`>=1.0.2`)
- [x] hide/skip certain subcommands (e.g. `daemon`)
  + `shtab>=1.0.0` https://github.com/iterative/dvc/pull/3921#discussion_r442787136
- [x] hide/skip certain arguments & options
  + `shtab>=1.0.0` <- https://github.com/iterative/shtab/issues/3
- [x] fix `zsh` subcommands with hyphens `-`
  + `shtab>=1.0.2` <- https://github.com/iterative/shtab/issues/4
- [x] re-introduce original metavar names prior to #3921
- part of #4070
- continues #3921

### DIFF
--- a/dvc/command/add.py
+++ b/dvc/command/add.py
@@ -66,6 +66,7 @@ def add_parser(subparsers, parent_parser):
         "targets",
         nargs="+",
         help="Input files/directories to add.",
+        metavar="targets",
         choices=completion.Required.FILE,
     )
     parser.set_defaults(func=CmdAdd)
--- a/dvc/command/cache.py
+++ b/dvc/command/cache.py
@@ -56,6 +56,7 @@ def add_parser(subparsers, parent_parser):
         help="Path to cache directory. Relative paths are resolved relative "
         "to the current directory and saved to config relative to the "
         "config file location.",
+        metavar="value",
         choices=completion.Required.DIR,
     )
     cache_dir_parser.set_defaults(func=CmdCacheDir)
--- a/dvc/command/checkout.py
+++ b/dvc/command/checkout.py
@@ -114,6 +114,7 @@ def add_parser(subparsers, parent_parser):
         nargs="*",
         help="DVC-files to checkout. Optional. "
         "(Finds all DVC-files in the workspace by default.)",
+        metavar="targets",
         choices=completion.Optional.DVC_FILE,
     )
     checkout_parser.set_defaults(func=CmdCheckout)
--- a/dvc/command/commit.py
+++ b/dvc/command/commit.py
@@ -67,6 +67,7 @@ def add_parser(subparsers, parent_parser):
         nargs="*",
         help="DVC-files to commit. Optional. "
         "(Finds all DVC-files in the workspace by default.)",
+        metavar="targets",
         choices=completion.Optional.DVC_FILE,
     )
     commit_parser.set_defaults(func=CmdCommit)
--- a/dvc/command/daemon.py
+++ b/dvc/command/daemon.py
@@ -66,6 +66,9 @@ def add_parser(subparsers, parent_parser):
         help=DAEMON_ANALYTICS_HELP,
     )
     daemon_analytics_parser.add_argument(
-        "target", help="Analytics file.", choices=completion.Required.FILE
+        "target",
+        help="Analytics file.",
+        metavar="target",
+        choices=completion.Required.FILE,
     )
     daemon_analytics_parser.set_defaults(func=CmdDaemonAnalytics)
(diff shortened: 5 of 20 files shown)


casperdcl:
Not a hack? Argparse always adds `[]` around options and `{}` around choices, and uses `dest` as the display name. `metavar` is the mechanism provided for the purpose of overriding all of this, and was used a lot in the codebase already.

The 29 places where `choices` were added in #3921 also needed `metavar` (added here in #4076) to control display.

There are 2 options:

1. make `shtab` infer completion type (e.g. `DVC_FILE`) based on `dest` (e.g. `target`)
    - decided against this as it forces devs to be consistent with their own naming conventions
2. explicitly mark completion types
  a) current approach: `add_argument(dest, choices=shtab.<TYPE>, metavar=dest)`
  b) `add_argument(shtab.Complete(dest, "<TYPE>"))`
    - `shtab.Complete` inherits from `str`

I prefer (1) followed by (2a) and then (2b), but let me know your thoughts

cc @iterative/engineering

efiop:
@casperdcl I mean that this metavar overwrite is only needed because we use shtab, we didn't use metavars before. Ideally, shtab shouldn't require the users to use such hacks, but at the same time this not that bad, just need to document it. Maybe it could even be used to our advantage if we could make the auto-generated metavar look better, but I'm not even sure it is possible.

casperdcl:
When you say "auto-generated metavar" what do you mean?

1. what's displayed with `add_argument("target", choices=[DVC_FILE])`
    + currently `{DVCFile}`, could be `{any-thing|*.dvc|can't avoid {} though}`
2. what's displayed with `add_argument("target", choices=[DVC_FILE], metavar="target")`
    + `target`
3. what could be displayed if we subclass & override the `add_argument` method to auto-fill `metavar`?

efiop:
@casperdcl I mean `add_argument("target")`, that would result in `target`. But when we start using shtab, we get `{DVCFile...}` or `url [{directory?}]`(which is absolutely confusing), which is a side-effect. Just wondering if we could avoid the side-effect or, possibly, embrace it and try to make it look nice and be helpful. Maybe the latter could be achieved by making shtab choices have their own nice-looking metavars, not sure.

casperdcl:
> I mean `add_argument("target")`, that would result in `target`.

yes

> But when we start using shtab, we get `{DVCFile...}`

That never happens (previously or currently).

Instead, with `add_argument("target", choices=[DVC_FILE])`, we used to get `{DVCFile}`, `[{DVCFile}]`, or `[{DVCFile} [{DVCFile} ...]]` for positional, optional, and optional list, respectively.

Now, with `add_argument("target", choices=[DVC_FILE], metavar="DVCFile")`, we get `DVCFile`, `[DVCFile]`, or `[DVCFile [DVCFile ...]]`, respectively, thus restoring the original look without the `{}`.

> or `[{directory?}]`

This was the case with `add_argument("--optional", choices=[OPTIONAL_DIR])`.

Now, with `add_argument("--optional", choices=[OPTIONAL_DIR], metavar="directory")`, we get `[directory]`, restoring the original look without `{}`. It also suppresses shtab's `?` feature which arguably could be removed.

The options we have are:

1. previous: use `choices` to specify type, accepting the extra `{}`
2. current: use `choices` to specify type and `metavar` to suppress `{}`
3. use `add_argument(ShtabString("target", "DVCFile"))`

In all proposed options, we have trivial full control over outputting `DVCFile` and `?` (i.e. it takes one line of code to globally change `[{DVCFile?}]` to `[{DVCFile}]` or `[{*.dvc|dvc.yml}]`).

> which is a side-effect.

There are 3 side-effects so I don't know which you refer to:

1. added `{}`
2. added `?`
3. rename `target` -> `DVCFile`

Side-effects (2) and (3) are trivially globally configurable by one line of code, side-effect (1) is a built-in argparse feature if using `choices`.

dvc/command/add.py
```diff
@@ -66,6 +66,7 @@ def add_parser(subparsers, parent_parser):
         "targets",
         nargs="+",
         help="Input files/directories to add.",
+        metavar="targets",
```
Suor, line 69: Can this be defaulted to arg name above?

Suor:
So the question is - can we make shtab to not use choices?

casperdcl:
> So the question is - can we make shtab to not use choices?

The only way I can see is the third option above with:

> `add_argument(ShtabString("target", type=DVC_FILE))` where `isinstance(ShtabString, str)`

casperdcl:
For the record: https://github.com/iterative/dvc/pull/4148 uses `add_argument("target").complete = DVC_FILE`

PR #4092, opened 2020-06-23
Enable pylint checks

skshetry:
It was getting ugly, so I decided to disable every checks and start enabling one-by-one that is preferred by us. 
If there's no consensus among us to enable particular checks, we should keep it disabled. 

Deepsource will be affected by this change (and, might return less error while in transition).


efiop:
> It was getting ugly, so I decided to disable every checks and start enabling one-by-one that is preferred by us.
If there's no consensus among us to enable particular checks, we should keep it disabled.

@skshetry The problem is that we will forget about them if we don't do it right away. Disabling everything doesn't make any sense, we will miss new features and will never enable them.

skshetry:
@efiop, No, I will keep making any PRs until we enable the checks that we need. I'd not suggest enabling everything and blacklisting some, at least for now. Let's go the other way round.

efiop:
> No, I will keep making any PRs until we enable the checks that we need. I'd not suggest enabling everything and blacklisting some, at least for now. Let's go the other way round.

So you plan on going through all avaliable checks one-by-one? What if there will be new features in the future? Or do you plan on starting with this and then switch this the other way around and enable everything and disable some?

skshetry:
@efiop, for the start, I'd love to keep things sane with pylint, and keep a whitelisted list of checks. Otherwise, we would just keep fighting over `pylint` and it's false-positive.

I'd recommend starting small and postpone the switching decisions for later.
See also: https://github.com/PyCQA/pylint/issues/3512, https://github.com/PyCQA/pylint/issues/746

efiop:
@skshetry We will have to enable everything and blacklist some in the near future, otherwise we will not even know that there is something useful (e.g. new) we can whitelist. I'm fine with the current approach, only if we switch in the near future.

.github/workflows/check-patch.yaml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
+    - name: Install requirements
+      run: bash ./scripts/ci/install.sh
```
skshetry, line 17: `pylint` requires all of the dependencies installed to check for `import-error`. Could either disable it (which is bad) or, need to install all of the dependency before-hand.

.pre-commit-config.yaml
```diff
@@ -23,6 +23,13 @@ repos:
           - flake8-string-format
     repo: https://gitlab.com/pycqa/flake8
     rev: 3.7.9
+  - repo: local
+    hooks:
+    - id: pylint
+      name: pylint
+      entry: pylint
+      language: system
+      types: [python]
```
skshetry, line 32: `pre-commit` creates an isolated virtualenv. But, for dynamic analysis such as `import-error`, `pylint` requires certain sets of dependencies installed, hence we need to use system's python (eg: from pip-installed one).

skshetry:
@efiop, I think we need to at least keep `pylint` only in Github Actions, or run `check-patch` in both the travis and GHA to not waste time for running tests even if the patch is wrongly formatted.

skshetry:
@efiop, also I think we could start with this PR.  I will follow-up with some PRs tomorrow.

.github/workflows/check-patch.yaml
```diff
@@ -0,0 +1,19 @@
+name: Lint Python
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
skshetry: I think it's okay to use the `latest` one. If it breaks, we will know.

.github/workflows/check-patch.yaml
```diff
@@ -0,0 +1,19 @@
...
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
+    - name: Install requirements
+      run: bash ./scripts/ci/install.sh
+    - name: Check Patch
+      run: ./scripts/ci/check_patch.sh
```
efiop, line 19: Looks like we can now remove it from travis?

.travis.yml
```diff
@@ -29,7 +29,7 @@ jobs:
     os: linux
     language: python
     python: 3.7
-    before_install:
+    before_install: bash ./scripts/ci/install.sh
     install:
     script: ./scripts/ci/check_patch.sh
```
efiop, line 34: If we have a github action for it already, then there is no need to run it again in travis.

.github/workflows/check-patch.yaml
skshetry: @efiop, mentioned here: https://github.com/iterative/dvc/pull/4092#issuecomment-648300217

efiop: Oops, missed that one. Ok, let's keep it for now.

PR #4101, opened 2020-06-24
optimize pre-commit, disable analytics on pre-commit run

skshetry:
Best viewed on [?w=1](4101/files?w=1).

Pre-commit waits for all the background tasks to complete, so the analytics was running twice when running pre-commit
which was making it slower.

Also, splitted class for the tests (though it's a Pytest one) into smaller functions.


tests/func/test_install.py
```diff
@@ -1,70 +1,69 @@
...
 import pytest
 
 from dvc.exceptions import GitHookAlreadyExistsError
 from dvc.utils import file_md5
 
-
-@pytest.mark.skipif(
-    sys.platform == "win32", reason="Git hooks aren't supported on Windows"
+pytestmark = pytest.mark.skipif(
+    sys.platform == "wind32", reason="Git hooks aren't supported on Windows"
```
skshetry, line 10: Oops
```suggestion
    sys.platform == "win32", reason="Git hooks aren't supported on Windows"
```

skshetry:
Also, it'd speed up if we use `.dvcignore`.

skshetry:
Local run of `pre-commit` was down to 1.18s from ~3.51s (depending on network conn.) for me.
With addition of `.dvcignore`, it reduced it further to 0.56s.

dvc/command/base.py
```diff
@@ -52,5 +56,7 @@ def run(self):
 
 
 class CmdBaseNoRepo(CmdBase):
+    check_updates = False
+
```
pared, line 60: I don't understand, why don't we want to check for updates on no-repo commands?

skshetry: As it is implemented right now, `Updater` uses `.dvc/tmp`. And, I haven't changed any implementation, Updater runs only on `CmdBase`, never on `CmdBaseNoRepo`.

dvc/command/git_hook.py
```diff
@@ -1,36 +1,37 @@
...
 
+        assert self.cmd
         try:
             repo = Repo()
             repo.close()
         except NotDvcRepoError:
             return 0
 
-        return self._run()
+        return main([self.cmd], disable_analytics=True)
```
efiop, line 20: Push/status/etc should trigger the analytics. The fact that it is slow means that we need to look deeper into it and optimize it.

dvc/command/git_hook.py
```diff
@@ -1,36 +1,37 @@
...
-
 from dvc.command.base import CmdBaseNoRepo, fix_subparsers
-from dvc.exceptions import NotDvcRepoError
-
-logger = logging.getLogger(__name__)
 
 
 class CmdHookBase(CmdBaseNoRepo):
+    cmd = None
+    collect_analytics = False
```
efiop, line 6: Arguably git-hook commands should also trigger the analytics. It is a piece of pretty useful knowledge for us.

dvc/command/git_hook.py
skshetry: Network connection do take some time.

dvc/command/git_hook.py
skshetry: Personally, I think it's too much to run anything else on git hooks. Sure, we can collect details if user are using pre-commit or not and how often they do, etc. But, collecting and sharing them at that instant feels wrong.

dvc/command/git_hook.py
efiop: @skshetry But that's why we launch it in a separate process. And that matters during regular `dvc` operation too, not only for hooks. There might be some regression.

skshetry: Nope, not a regression. `pre-commit` waits for all of the processes to finish.

efiop: @skshetry We launch analytics in a detached state, so it shouldn't wait for that process to finish. Unless there is something wrong on our side, of course.

dvc/command/git_hook.py
pmrowla: I think it's acceptable for us to run the analytics on git hooks given that users can opt out and disable them. Maybe we could just add a note in the docs that analytics will also be collected when dvc is run via the pre-commit hooks?

skshetry:
Looks like we don't want to disable analytics on pre-commit. As it's the gist of this PR, I am closing this for now.

dvc/command/git_hook.py
efiop: @pmrowla Not sure it is worth documenting explicitly, because we already say that we just collect the  analytics when you run our commands, and git-hook is a command too, so it seems logical that it wouldn't be an exception.

Issue #4104, opened 2020-06-24
`dvc pull` causes an error when there is an empty file in a remote


MGniew:
## Bug Report

Hi there, I hope I find you well :)
First of all, congratulation on a new release. It is absolutely awesome!

Recently I was working with a new dataset and the following error surprised me:
```
ERROR: failed to download 's3://blahblagh/myproject/d4/1d8cd98f00b204e9800998ecf8427e' to '.dvc/cache/d4/1d8cd98f00b204e9800998ecf8427e' - 'ContentLength' 
```
It turns out that one of my files was empty and therefore could not be downloaded from the remote host via https (S3).

**What is an issue:**

Empty files causes error when you are trying to download them from a remote.

**Steps to Reproduce:**

1. Initialize your git repo: `git init`
2. Initialize dvc: `dvc init`
3. Add a remote (I'm using Minio):

```
dvc remote add -d myremote s3://mybucket/path/to/dir
dvc remote modify myremote endpointurl https://object-storage.example.com
``` 
4. Create a directory with 2 example files:

```
mkdir data
echo "not empty" > data/not-empty.txt
> data/empty.txt
```
5. Upload them to the remote `dvc add data && dvc push`.
7. Files are now successfully uploaded. I can see them and download them manually from my remote storage.
6. Remove ./data and .dvc/cache
7. Try to download them: `dvc pull`

I dont think that this is an important bug (I simply found and deleted the empty file) but it might have other consequences that I can't think about.

**Output of `dvc version`:**

```console
$ DVC version: 1.0.1
Python version: 3.6.10
Platform: Linux-5.4.0-37-generic-x86_64-with-Ubuntu-20.04-focal
Binary: False
Package: pip
Supported remotes: http, https, s3
Cache: reflink - not supported, hardlink - supported, symlink - supported
Repo: dvc, git
```

**Additional Information (if any):**

It`s a KeyError exception:

```console
  File "..../venv/lib/python3.6/site-packages/dvc/remote/s3.py", line 333, in _download
    )["ContentLength"]
KeyError: 'ContentLength'
```

In case of lack of "ContentLength" header, it probably should assume it is 0 and create an empty file anyway.

efiop:
Hi @MGniew ! Are you using genuine s3 or some s3-compatible storage?

MGniew:
> Hi @MGniew ! Are you using genuine s3 or some s3-compatible storage?

Hi @efiop. Im using Minio (https://min.io/).

[...]

PR #4105, opened 2020-06-24
Update s3.py

MGniew:
Issue https://github.com/iterative/dvc/issues/4104

Yes, but most of the tests don`t work (before and after change)!

Before:
38 failed, 264 passed, 4 skipped, 565 error in 13.29 seconds 
After:
38 failed, 264 passed, 4 skipped, 565 error in 13.29 seconds 

### DIFF
--- a/dvc/remote/s3.py
+++ b/dvc/remote/s3.py
@@ -331,7 +331,7 @@ class S3RemoteTree(BaseRemoteTree):
         if total == 0:
             basedir = os.path.dirname(to_file)
             os.makedirs(basedir, exist_ok=True)
-            open(to_file, 'w').close()
+            open(to_file, "w").close()
             return
 
         with Tqdm(
--- a/dvc/remote/s3.py
+++ b/dvc/remote/s3.py
@@ -325,12 +325,15 @@ class S3RemoteTree(BaseRemoteTree):
             )
 
     def _download(self, from_info, to_file, name=None, no_progress_bar=False):
-        if no_progress_bar:
-            total = None
-        else:
-            total = self.s3.head_object(
-                Bucket=from_info.bucket, Key=from_info.path
-            )["ContentLength"]
+        total = self.s3.head_object(
+            Bucket=from_info.bucket, Key=from_info.path
+        ).get("ContentLength", 0)
+        if total == 0:
+            basedir = os.path.dirname(to_file)
+            os.makedirs(basedir, exist_ok=True)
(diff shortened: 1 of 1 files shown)


dvc/remote/s3.py
```diff
@@ -325,12 +325,15 @@ def _upload(self, from_file, to_info, name=None, no_progress_bar=False):
...
-                Bucket=from_info.bucket, Key=from_info.path
-            )["ContentLength"]
+        total = self.s3.head_object(
+            Bucket=from_info.bucket, Key=from_info.path
+        ).get("ContentLength", 0)
+        if total == 0:
+            basedir = os.path.dirname(to_file)
+            os.makedirs(basedir, exist_ok=True)
+            open(to_file, "w").close()
+            return
```
efiop, line 335: I don't think this one is needed, as it seems to work fine with regular s3.

MGniew: Sadly it wont work without this return (and creating file is also needed), because similar error is raised in boto3:

```
  File "venv/lib/python3.6/site-packages/s3transfer/download.py", line 346, in _submit
    response['ContentLength'])
KeyError: 'ContentLength'
```
So my solution basically doesn't run `self.s3.download_file` if the file is empty to overcome this issue.

sdatko:
Could you also provide tests to cover this scenario? It may be helpful in the future ;-)

dvc/remote/s3.py
MGniew: Ok sorry, I misunderstood your comment. Are you absolutely positive that it works with regular s3? If so there must be a 'ContentLength' set to zero there.

efiop: Yes, it works with s3 because it sets ContentLength to 0. Ok, so this is even worser than we thought. If s3transfer also doesn't work with it, then it is an even bigger problem for minio. Could you please report this issue to minio guys? Let's see what they have to say about it.

efiop: Sure, we could use this hack in dvc just for the minio for now, but ideally, we need to fix it in minio or in boto3/s3transfer as well.

MGniew: Ok, so I will make sure its is not my minio version problem (note my comment in the issue) and then (if necessary) I will report this problem to minio guys :) Keep in touch

MGniew:
Newest version of Minio returns ContentLength=0

PR #4114, opened 2020-06-25
version: redesign output format for the dvc version command #3499

### LINKED ISSUE #3499: version: redesign output
Current `dvc version` output is ugly, we could definitely do better:

```
DVC version: 0.89.0+9e96dd
Python version: 3.7.0
Platform: Linux-4.15.0-88-generic-x86_64-with-debian-buster-sid
Binary: False
Package: None
Filesystem type (cache directory): ('ext4', '/dev/sda1')
Filesystem type (workspace): ('ext4', '/dev/sda1')
```

Related https://github.com/iterative/dvc/issues/3185

sahilbhosale63:
Fix #3499 redesign the output format for the `dvc version` command.


skshetry:
@sahilbhosale63, thanks for the PR. CI is failing. Did you run pre-commit hooks as mentioned here: https://dvc.org/doc/user-guide/contributing/core.

Also, it'd be good to see the outputs for different DVC installations/repos. Thanks.

dvc/command/version.py
```diff
@@ -129,7 +127,12 @@ def get_supported_remotes():
             if not tree_cls.get_missing_deps():
                 supported_remotes.append(tree_cls.scheme)
 
-        return ", ".join(supported_remotes)
+        if supported_remotes.__len__() == 9:
+            return "All remotes"
+        elif supported_remotes.__len__() == 1:
```
skshetry, line 132: In Python, it's not good to use dunder methods directly(`__len__`). You can just use `len()`.

dvc/command/version.py
```diff
@@ -129,7 +127,12 @@ def get_supported_remotes():
             if not tree_cls.get_missing_deps():
                 supported_remotes.append(tree_cls.scheme)
 
-        return ", ".join(supported_remotes)
+        if supported_remotes.__len__() == 9:
```
skshetry, line 130: @sahilbhosale63, our support for remotes could change, it'd be great to find some way to get all the counts of the remotes that we support.

dvc/command/version.py
```diff
@@ -115,10 +111,12 @@ def get_linktype_support_info(repo):
                 os.unlink(dst)
             except DvcException:
                 status = "not supported"
-            cache.append(f"{name} - {status}")
+
+            if status == "supported":
+                cache.append(f"{name}")
```
skshetry, line 116: No need of f-strings here.
```suggestion
                cache.append(name)
```

sahilbhosale63:
Screenshot from 2020-06-26 21-54-06
@skshetry I want to commit the changes which I have made but when I run git commit some tests are failing. And also when I check if my changes are committed or not via git status then it shows "your branch is up to date with origin/master".

What's the problem? Can you please guide me here @skshetry?

dvc/command/version.py
```diff
@@ -10,7 +10,7 @@
 from dvc.exceptions import DvcException, NotDvcRepoError
 from dvc.scm.base import SCMError
 from dvc.system import System
-from dvc.utils import is_binary, relpath
+from dvc.utils.__init__ import relpath
```
pmrowla, line 13: this should just be 

```py
from dvc.utils import relpath
```

dvc/command/version.py
```diff
@@ -46,13 +45,11 @@ def run(self):
...
                 info.append(
-                    "Cache: {}".format(self.get_linktype_support_info(repo))
+                    "Cache types: " + self.get_linktype_support_info(repo)
                 )
                 if psutil:
                     fs_type = self.get_fs_type(repo.cache.local.cache_dir)
-                    info.append(
-                        f"Filesystem type (cache directory): {fs_type}"
-                    )
+                    info.append(f"Directory cache: {fs_type}")
```
pmrowla, line 52: `Directory cache` feels a bit off here. Seems like it should be `Cache directory:` or maybe just `Cache:`

dvc/command/version.py
```diff
@@ -129,7 +128,12 @@ def get_supported_remotes():
             if not tree_cls.get_missing_deps():
                 supported_remotes.append(tree_cls.scheme)
 
-        return ", ".join(supported_remotes)
+        if len(supported_remotes) == len(TREES):
+            return "All remotes"
+        elif len(supported_remotes) == 1:
+            return supported_remotes
+        else:
+            return ", ".join(supported_remotes)
```
pmrowla, line 136: nitpicky, but you don't need `elif`/`else` here
```suggestion
        if len(supported_remotes) == 1:
            return supported_remotes
        return ", ".join(supported_remotes)
```

dvc/command/version.py
skshetry: +1 for `Cache:`

dvc/command/version.py
skshetry, line 136: @sahilbhosale63, the plan was to make `Supports` use both remotes and cache types. But, as we didnot go through that, I think it's okay to not go "All remotes" route. But, it's your call.

dvc/command/version.py
skshetry: Also, we could change the order of this to be just above `Workspace`.

skshetry:
@sahilbhosale63, @pmrowla, maybe we could do something like this:
```
Build Info: Python 3.7.7 on Darwin-19.5.0-x86_64-i386-64bit (pip)
Build Info: Python 3.7.7 on Darwin-19.5.0-x86_64-i386-64bit (binary)
```

Also, it would be good to rename `Build Info` to `Platform` which is what it is.

Or, could add it to the version: `DVC version: 1.0.0 (pip)`?

pmrowla:
> Or, could add it to the version: `DVC version: 1.0.0 (pip)`?

This might work best just because the platform line is already going to be pretty long in a lot of cases (even without the installation method)

sahilbhosale63:
@skshetry can you please review this patch?

dvc/command/version.py
```diff
@@ -28,12 +28,11 @@ def run(self):
         from dvc.repo import Repo
 
         info = [
-            f"DVC version: {__version__}",
-            f"Python version: {platform.python_version()}",
-            f"Platform: {platform.platform()}",
-            f"Binary: {is_binary()}",
-            f"Package: {PKG}",
-            f"Supported remotes: {self.get_supported_remotes()}",
+            f"DVC version: {__version__} ({PKG})",
```
skshetry, line 31: I get:
```console
DVC version: 1.0.1+412f39 (None)
```
`PKG` can be None too. So, let's leave that out if it's None.

dvc/command/version.py
```diff
@@ -46,13 +45,11 @@ def run(self):
             # `dvc config cache.shared group`.
             if os.path.exists(repo.cache.local.cache_dir):
                 info.append(
-                    "Cache: {}".format(self.get_linktype_support_info(repo))
+                    "Cache types: " + self.get_linktype_support_info(repo)
```
skshetry, line 48: Let's use `{}` formated string as above with just the addtion of `Cache types:`.

dvc/command/version.py
```diff
@@ -46,13 +45,11 @@ def run(self):
...
                 )
                 if psutil:
                     fs_type = self.get_fs_type(repo.cache.local.cache_dir)
-                    info.append(
-                        f"Filesystem type (cache directory): {fs_type}"
-                    )
+                    info.append(f"Cache with {fs_type}")
```
skshetry, line 52: I agree with @pmrowla that it looks strange with mixed colon/no-colon output. Let's just use `Cache directory: ext4 on /dev/sda4` kind of output. Same for workspace too (Workspace directory?).

dvc/command/version.py
```diff
@@ -139,7 +144,7 @@ def _get_dvc_repo_info(repo):
     if repo.root_dir != repo.scm.root_dir:
         return "dvc (subdir), git"
 
-    return "dvc, git"
+    return "dvc + git"
```
skshetry, line 147: We need to make this uniform, either we should make it all commas or all plus-es. Could just use comma for now.

dvc/command/version.py
```diff
@@ -28,12 +28,11 @@ def run(self):
         from dvc.repo import Repo
 
         info = [
-            f"DVC version: {__version__}",
-            f"Python version: {platform.python_version()}",
-            f"Platform: {platform.platform()}",
-            f"Binary: {is_binary()}",
-            f"Package: {PKG}",
```
skshetry, line 35: @efiop, is `Binary` necessary? I assume we can deduct that through `PKG`, right?

sahilbhosale63:
rsz_screenshot_from_2020-07-21_20-55-51

@skshetry @pmrowla Does this looks good??

dvc/command/version.py
efiop: @skshetry Yeah, could totally live without `Binary`. PKG is enough.

dvc/command/version.py
jorgeorpinel: Same for cache types which seems to use `/` right now.

jorgeorpinel:
I just noticed this PR, thanks for the mention.

> Supports: All remotes

How about

Supported remotes: {List them always} — actually I think that's how it is now so maybe this doesn't need a change...

Also:
- `Cache types` -> `Supported cache types` ?
- `Workspace Directory` - lower case `directory`
- `Cache Directory` - same ^. And put after Workspace dir?
- `Repo: ` -> `Repo control: ` ?
---

Also keep in mind this may need updates to https://dvc.org/doc/command-reference/version or other docs after merge. Thanks

dvc/command/version.py
shcheklein: +1 for commas (since / is a path separator)
+1 to check other occurrences

shcheklein:
@sahilbhosale63 why do we need (`----------------`) separator?

How does the output looks like when no cache dir is available?
I don't see the install method ... is it None on your screenshot? should we use some human readable name?

sahilbhosale63:
@jorgeorpinel  https://github.com/iterative/dvc/issues/3499

sahilbhosale63:
> @sahilbhosale63 why do we need (`----------------`) separator?
> 
> How does the output looks like when no cache dir is available?
> I don't see the install method ... is it None on your screenshot? should we use some human readable name?

When there is no cache dir it gives a warning.

jorgeorpinel:
Sounds like that issue number should be in the PR description? Linked so merging this closes it. See https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords

sahilbhosale63:
> Sounds like that issue number should be in the PR description? Linked so merging this closes it. See https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords

Added #3499 issue number to the PR description with the **Fix** keyword.

sahilbhosale63:
The Travis Build test failures are unrelated.

dvc/command/version.py
```diff
@@ -27,13 +27,18 @@ class CmdVersion(CmdBaseNoRepo):
...
+
         info = [
-            f"DVC version: {__version__}",
-            f"Python version: {platform.python_version()}",
-            f"Platform: {platform.platform()}",
-            f"Binary: {is_binary()}",
-            f"Package: {PKG}",
-            f"Supported remotes: {self.get_supported_remotes()}",
+            f"DVC version: {__version__} {package}",
+            "--------------------------------- \n",
```
skshetry, line 38: No need of newline after the divider.

skshetry:
```suggestion
            "---------------------------------",
```

efiop:
Thank you @sahilbhosale63 ! :pray:

jorgeorpinel:
Created iterative/dvc.org/issues/1635 to update the docs.

PR #4122, opened 2020-06-26
pylint: enable additional checks

skshetry:

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -15,7 +15,11 @@ ignore-patterns=
 
 # Python code to execute, usually for sys.path manipulation such as
 # pygtk.require().
-#init-hook=
+init-hook="import sys; \
+           from pathlib import Path; \
+           from pylint.config import find_pylintrc; \
+           sys.path.append(Path(find_pylintrc()).parent / 'tests');"
+
 
 # Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
 # number of processors available to use.
@@ -28,7 +32,7 @@ limit-inference-results=100
 
 # List of plugins (as comma separated values of python modules names) to load,
 # usually to register additional checkers.
-load-plugins=pylint_pytest
+load-plugins=pylint_pytest,pylint_plugin_disable
 
 # Pickle collected data for later comparisons.
 persistent=yes
@@ -149,10 +153,8 @@ disable=format,
         exception-escape,
         comprehension-escape,
         # custom disables start here
-        protected-access,  # needs-refactor
         no-self-use,
         invalid-name,
-        blacklisted-name,
         # already a check enabled for `wildcard-import`
         unused-wildcard-import,
         misplaced-comparison-constant,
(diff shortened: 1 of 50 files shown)


.pylintrc
```diff
@@ -15,7 +15,11 @@ ignore-patterns=
 
 # Python code to execute, usually for sys.path manipulation such as
 # pygtk.require().
-#init-hook=
+init-hook="import sys; \
+           from pathlib import Path; \
+           from pylint.config import find_pylintrc; \
+           sys.path.append(Path(find_pylintrc()).parent / 'tests');"
+
```
skshetry, line 22: Required for loading the plugin.

dvc/remote/base.py
```diff
@@ -413,7 +419,7 @@ def upload(self, from_info, to_info, name=None, no_progress_bar=False):
 
         name = name or from_info.name
 
-        self._upload(
+        self._upload(  # noqa, pylint: disable=no-member
```
skshetry, line 422: I think, we should create an `_upload`/`_download` function in the base class that just throws `RemoteActionNotImplemented` error. It would make this cleaner (without ignore, that is), and provide stronger checks-and-balances.

tests/pylint_plugin_disable.py
```diff
@@ -0,0 +1,47 @@
...
+    "blacklisted-name": [
+        NameChecker.visit_global,
+        NameChecker.visit_assignname,
+    ],
+}
+
+
+def is_node_in_tests(node: "node_classes.NodeNG"):
+    module: "scoped_nodes.Module" = node.root()
+    return module.file.startswith(TESTS_FOLDER)
```
skshetry, line 34: @efiop, @pmrowla, @pared, I think we could do a similar prefix-based solution to reduce some `relpath`/`abspath` calls.

dvc/repo/tree.py
```diff
@@ -55,7 +55,9 @@ def _get_granular_checksum(self, path, out, remote=None):
                 return entry[out.remote.tree.PARAM_CHECKSUM]
         raise FileNotFoundError
 
-    def open(self, path, mode="r", encoding="utf-8", remote=None):
+    def open(
+        self, path, mode="r", encoding="utf-8", remote=None
+    ):  # pylint: disable=arguments-differ
```
pmrowla, line 60: Should we start preferring the use of **kwargs in these kinds of cases instead of suppressing this pylint test each time?

dvc/main.py
```diff
@@ -65,8 +65,8 @@ def main(argv=None):
     except DvcException:
         ret = 255
         logger.exception("")
-    except Exception as exc:  # pylint: disable=broad-except
-        if isinstance(exc, OSError) and exc.errno == errno.EMFILE:
+    except OSError as exc:
+        if exc.errno == errno.EMFILE:
```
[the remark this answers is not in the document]
efiop: This one is questionable, just duplicating the code :(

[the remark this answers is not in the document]
skshetry: I think it's worth it to separate exception here. The duplication is 2 lines and most of it is incidental, the return code and message can change independently like in the above try-except ladder.

I could do:
```python
if ret == 255:
  logger.info("unexpected error")
```

But, I don't think it's worth the hassle.

Issue #4050, opened 2020-06-16
refactor: consolidate tree classes


pmrowla:
Follow up to #3882.

Currently we have tree classes that all inherit from `dvc.scm.tree.BaseTree` spread out all over the code base w/`DvcTree`, `RepoTree`, SCM/git related trees, `BaseRemoteTree`, and all of the associated cloud remote trees. It may make more sense to group these in something like `dvc.tree` and `dvc.tree.cloud`. This would also allow us to simplify what is currently under `dvc.remote`, since we really only need single files for `Remote` and `CloudCache` now. (Cloud cache code could also potentially just be moved into `dvc.cache` now as well.)

efiop:
Ok, I think we are done for now. Core code is separated, but there might be some issues like https://github.com/iterative/dvc/pull/4244 which were there before this, but we weren't able to see them. We'll have to fix them as we go, especially since we are trying to use trees more and more and those bugs are getting more and more obvious.

I didn't move the tests around for now, because they are intermixed and it doesn't seem worth it to bother with in this sprint. But it is a technical debt, so we'll have to get to them at some point in the future. Luckily, they are easy to separate when writing new tests (similar how we are doing with unit -> pytest migration). Plus, `cache` and `remote` classes still have some cloud-specific implementations (e.g. local and ssh where we use some sftp channel magic) which will likely be slimmed down once we unify tree interfaces (see comment about it below). 

Next big thing here is to, of course, optimize dvcignore (something that is on our radar right now, has a separate ticket and PR), but , even more importantly, we will need to unify interfaces of our tree classes better. E.g. make them all raise appropriate OSErrors (or something else if we will be able to find something better) so we are able to handle them in a unified way. Same with return types, where it would be great to use type hints and use mypy to check them.

PR #4125, opened 2020-06-26
LocalRemoteTree: use repo tree as work_tree with local outputs

### LINKED ISSUE #4110: PermissionError on file which should be in .dvcignore.
## Bug Report
Getting a permission denied when I try to DVC add a file that is in my dvc ignore.
This is the `.dvcignore` file:
```
*/*/.thumbnail/
*/*/.exports/
```
Last few errors in traceback:
``` 
File "/home/andrea/projects/myproject/.venv/lib/python3.6/site-packages/dvc/stage/__init__.py", line 413, in commit
    out.commit()
  File "/home/andrea/projects/myproject/.venv/lib/python3.6/site-packages/dvc/output/base.py", line 281, in commit
    self.cache.save(self.path_info, self.cache.tree, self.info)
  File "/home/andrea/projects/myproject/.venv/lib/python3.6/site-packages/dvc/remote/base.py", line 1154, in save
    return self._save(path_info, tree, hash_, save_link, **kwargs)
  File "/home/andrea/projects/myproject/.venv/lib/python3.6/site-packages/dvc/remote/base.py", line 1161, in _save
    return self._save_dir(path_info, tree, hash_, save_link, **kwargs)
  File "/home/andrea/projects/myproject/.venv/lib/python3.6/site-packages/dvc/remote/base.py", line 1133, in _save_dir
    entry_info, tree, entry_hash, save_link=False, **kwargs
  File "/home/andrea/projects/myproject/.venv/lib/python3.6/site-packages/dvc/remote/base.py", line 1070, in _save_file
    self.tree.move(path_info, cache_info, mode=self.CACHE_MODE)
  File "/home/andrea/projects/myproject/.venv/lib/python3.6/site-packages/dvc/remote/local.py", line 138, in move
    move(from_info, to_info, mode=mode)
  File "/home/andrea/projects/myproject/.venv/lib/python3.6/site-packages/dvc/utils/fs.py", line 103, in move
    shutil.move(src, tmp)
  File "/usr/lib/python3.6/shutil.py", line 565, in move
    os.unlink(src)
```
Final Error:
```
PermissionError: [Errno 13] Permission denied: '/home/andrea/projects/coco-annotator/datasets/myproject/datasets/RoIDatasetValidv4/.exports/coco-1593013621.885437.json'
```

**Output of `dvc version`:**

```DVC version: 1.0.1
Python version: 3.6.9
Platform: Linux-5.3.0-51-generic-x86_64-with-Ubuntu-18.04-bionic
Binary: False
Package: pip
Supported remotes: azure, http, https, ssh
Filesystem type (cache directory): ('ext4', '/dev/nvme0n1p3')
Repo: dvc, git
Filesystem type (workspace): ('ext4', '/dev/nvme0n1p3')
```

pared:

EDIT:
Fixes #4110 
Fixes https://github.com/iterative/dvc/issues/4197

### DIFF
--- a/dvc/output/local.py
+++ b/dvc/output/local.py
@@ -5,6 +5,7 @@ from urllib.parse import urlparse
 from dvc.exceptions import DvcException
 from dvc.istextfile import istextfile
 from dvc.output.base import BaseOutput
+from dvc.scm.tree import is_working_tree
 from dvc.utils import relpath
 from dvc.utils.fs import path_isin
 
@@ -22,6 +23,8 @@ class LocalOutput(BaseOutput):
             path = relpath(path, stage.wdir)
 
         super().__init__(stage, path, *args, **kwargs)
+        if self.is_in_repo and self.repo and is_working_tree(self.repo.tree):
+            self.tree = self.repo.tree
 
     def _parse_path(self, tree, path):
         parsed = urlparse(path)
--- a/tests/func/test_ignore.py
+++ b/tests/func/test_ignore.py
@@ -364,3 +364,28 @@ def test_pattern_trie_tree(tmp_dir, dvc):
         )
         == ignore_pattern_bottom
     )
+
+
+def test_ignore_in_added_dir(tmp_dir, dvc):
+    tmp_dir.gen(
+        {
+            "dir": {
+                "sub": {
+                    "ignored": {"content": "ignored content"},
+                    "not_ignored": "not ignored content",
+                }
+            },
+            ".dvcignore": "**/ignored",
+        }
+    )
+    dvc.tree.__dict__.pop("dvcignore", None)
+
+    ignored_path = tmp_dir / "dir" / "sub" / "ignored"
+    assert not dvc.tree.exists(PathInfo(ignored_path))
+    assert ignored_path.exists()
+
+    dvc.add("dir")
+    shutil.rmtree(ignored_path)
+    dvc.checkout()
+
+    assert not ignored_path.exists()


tests/func/test_ignore.py
```diff
@@ -173,3 +173,27 @@ def test_ignore_blank_line(tmp_dir, dvc):
     tmp_dir.gen(DvcIgnore.DVCIGNORE_FILE, "foo\n\ndir/ignored")
 
     assert _files_set("dir", dvc.tree) == {"dir/other"}
+
+
+def test_ignore_in_added_dir(tmp_dir, dvc):
```
pared, line 178: I did consider making test for LocalRemoteTree only, but I guess, in the end, we want to be sure that one can actually ignore something in added directory.

dvc/ignore.py
```diff
@@ -216,7 +216,7 @@ def _parents_exist(self, path):
                 [os.path.abspath(path), self.tree_root]
             )
         ):
-            return False
+            return True
```
pmrowla, line 219: This works as a temporary solution, but it still seems unintuitive.

It seems to me that dvcignore should only ever apply to a DVC repo, and by definition anything outside the repo root directory should not exist in the clean tree.

dvc/remote/local.py
```diff
@@ -66,7 +67,7 @@ def work_tree(self):
         # GitTree arbitarily. When repo.tree is GitTree, local cache needs to
         # use its own WorkingTree instance.
         if self.repo:
-            return WorkingTree(self.repo.root_dir)
+            return CleanTree(WorkingTree(self.repo.root_dir))
```
pmrowla, line 70: Same as before, this works as part of a temporary fix but I'm not sure it makes sense as a real solution.

For local remotes, and DVC local cache we should need to wrap the work tree with CleanTree. With a local remote, it should only contain files which have been pushed to it, and we should not care about filtering/ignoring anything inside the remote. Likewise, for cache, we should not care about filtering anything inside `.dvc/cache`.

The main point of this tree refactor was that local remotes and local cache should only be dealing with paths inside the actual remote or cache. We run into issues when we start mixing remote/cache/repo tree paths and treating them all as just "local filesystem paths".

pmrowla:
These changes do fix the user issue, and CleanTree used to allow paths outside the repo root before the remote tree refactoring, so I'd be ok with merging this as a bug fix for now. But long term I think we should keep the other things I mentioned in mind.

pared:
@pmrowla I tried to make it less hackish behavior, and determine what tree should be used on `Output` level. The results are in newest commit.

I do believe that current behavior is still hackish.
What I believe we should do after this change:
1. rename `*RemoteTree` or `(Git/Working/Repo/DVC/Clean)Tree` - it gets really confusing which one we are using and where, and we are intermixing them, especially in `LocalRemoteTree`. Not sure how the renaming should go, so far I came up with `(Git/Working/Repo/DVC)FileSystem` since those objects provide fs-like operations, thought in that case CleanTree does not make sense, maybe it should be named `IgnoreWrapper` or something like that.
2. Related to 1: we are duplicating Local Filesystem functionality in WorkingTree and LocalRemoteTree. I think we should get rid of that duplication. First thing that comes to my mind is creating more `BaseTree` children, implementing fs operations for all remote types, and using those trees in `*RemoteTree`'s. That does not seem like perfect soultion as we already have this chierarchy `*Remote` -> `*RemoteTree` and now we would add another  thing: `*RemoteTree` -> `*Tree`. Probably that should be reconsidered.

dvc/output/local.py
```diff
@@ -23,6 +26,17 @@ def __init__(self, stage, path, *args, **kwargs):
...
+    @cached_property
+    def remote(self):
+        if self._remote:
+            return self._remote
+
+        work_tree = None
+        if self.repo and self.is_in_repo and is_working_tree(self.repo.tree):
+            work_tree = self.repo.tree
+        tree = LocalRemoteTree(self.repo, {}, work_tree)
+        return LocalRemote(tree)
```
efiop, line 38: Could we do this remote setting in `__init__`?  Just feels a bit weird introducing cached property & _remote mix in both here and base class for this hack.

pared: We could, but in that case, we will set remote inside parent constructor and override it here. Thats what I wanted to avoid creating `cached_property`

efiop: @pared but we can set it after `super()`, no?

efiop:
Had a similar issue with config and it really hurts :slightly_frowning_face: Thinking that maybe we should tackle https://github.com/iterative/dvc/issues/4050 right away and get rid of that is_working_tree mess once and for all. Otherwise, we might be introducing much more hidden issues than we are solving (though in this case this is clearly very serious).

Let's discuss this during planning today, maybe someone will be able to look into this part of trees deeply ASAP. If not, we'll have to merge the hack, indeed. :slightly_frowning_face:

tests/func/test_ignore.py
```diff
@@ -343,3 +343,27 @@ def test_pattern_trie_tree(tmp_dir, dvc):
         )
         == ignore_pattern_bottom
     )
+
+
+def test_ignore_in_added_dir(tmp_dir, dvc):
+    tmp_dir.dvc_gen(
```
efiop, line 349:
```suggestion
    tmp_dir.gen(
```

Issue #4131, opened 2020-06-29
Issues with dvc get and dvc get-url after 1.0


honnibal:
Hi,

I've been excited to explore DVC as I think it can offer a lot for some of the things I'm working on. I've had a couple of issues with `dvc get` and `dvc get-url` that I *think* are probably bugs. I've put them together to avoid creating lots of issues; apologies if you'd rather they were separate, I can move them into different threads.

I am testing as of v1.1.1.

## dvc get does not recreate directories correctly

After v1.0 `dvc get` pointed to a directory does not result in correctly placed files. All of the files are placed in the root, rathe than subdirectories. I've only tested this with a git repository that is not a DVC repository. It worked in v0.94, but does not work correctly in v1.0 or v1.1. `dvc import` works correctly.

The following example gets [this directory](https://github.com/explosion/projects/tree/master/nel-emerson/scripts):

```
$ dvc get https://github.com/explosion/projects nel-emerson
$ tree nel-emerson/
nel-emerson/
├── README.md
├── __init__.py
├── el_recipe.py
├── el_tutorial.py
├── emerson_annotated_text.jsonl
├── emerson_input_text.txt
├── entities.csv
├── nel_schema.png
├── notebook_video.ipynb
└── requirements.txt
0 directories, 10 files
```

And here's `dvc import`, which works correctly:

```
$ dvc import https://github.com/explosion/projects nel-emerson
$ tree nel-emerson/
nel-emerson/
├── README.md
├── __init__.py
├── input
│   └── entities.csv
├── prodigy
│   ├── emerson_annotated_text.jsonl
│   └── emerson_input_text.txt
├── requirements.txt
└── scripts
    ├── el_recipe.py
    ├── el_tutorial.py
    ├── nel_schema.png
    └── notebook_video.ipynb
3 directories, 10 files
```

## dvc get-url fails on URLs with weak ETAGs

I understand why `import-url` needs the strong ETag, but I don't think it makes sense for `get-url`, so I suspect this is a bug. If so it's a regression after v1.0, as it didn't happen in v0.94.

```
$ dvc get-url https://raw.githubusercontent.com/explosion/projects/master/nel-emerson/requirements.txt
[...]

PR #4133, opened 2020-06-30
RepoTree: preserve relative paths in copytree()

pmrowla:
Partial fix for #4131 (resolves the `dvc get` bug only).

### DIFF
--- a/dvc/repo/tree.py
+++ b/dvc/repo/tree.py
@@ -428,11 +428,12 @@ class RepoTree(BaseTree):
 
         for root, _, files in self.walk(top):
             root = PathInfo(root)
-            makedirs(dest, exist_ok=True)
+            dest_dir = dest / root.relative_to(top)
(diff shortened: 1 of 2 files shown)


dvc/repo/tree.py
```diff
@@ -428,11 +428,12 @@ def copytree(self, top, dest):
 
         for root, _, files in self.walk(top):
             root = PathInfo(root)
-            makedirs(dest, exist_ok=True)
+            dest_dir = dest / root.relative_to(top)
+            makedirs(dest_dir, exist_ok=True)
             for fname in files:
                 src = root / fname
                 with self.open(src, mode="rb") as fobj:
-                    copy_fobj_to_file(fobj, dest / fname)
```
pmrowla, line 435: `dest` here was the top level destination path, rather than the expected relative path

skshetry: Yeah, I should not have removed that without thinking once. My bad.

PR #4138, opened 2020-06-30
Reorder package widgets

dmpetrov:
Moving the most important widgets on the front


efiop:
For the record: Tests failing for unrelated reasons.

PR #4140, opened 2020-07-01
tests: remotes: use TmpDir-like fixtures

efiop:


tests/remotes/__init__.py
```diff
@@ -33,3 +43,26 @@
     ssh_remote,
     ssh_server,
 )
+
+
+@pytest.fixture
+def workspace(tmp_dir, dvc, request):
```
efiop, line 49: will introduce a similar `remote` fixture as well instead of current explicit ones.

tests/remotes/azure.py
```diff
@@ -30,7 +33,7 @@ def get_url():
 def azure():
     if not Azure.should_test():
         pytest.skip("no azure running")
-    yield Azure()
+    yield Azure(Azure.get_url())
```
efiop, line 36: This `CLS(CLS.get_url())` mess will be removed once no tests use it. Likely in this PR.

tests/remotes/base.py
```diff
@@ -1,7 +1,61 @@
...
+                else:
+                    self._gen(contents, prefix=path)
+            else:
+                path.parent.mkdir(parents=True)
+                if isinstance(contents, bytes):
+                    path.write_bytes(contents)
+                else:
+                    path.write_text(contents, encoding="utf-8")
+
+    def gen(self, struct, text=""):
```
efiop, line 49: This will be unified with TmpDir in the future.

tests/remotes/gs.py
```diff
@@ -64,24 +63,63 @@ def _get_storagepath():
     def get_url():
         return "gs://" + GCP._get_storagepath()
 
-    @classmethod
-    @contextmanager
-    def remote(cls, repo):
-        yield Remote(GSRemoteTree(repo, {"url": cls.get_url()}))
+    @property
+    def _gc(self):
+        from google.cloud.storage import Client
```
efiop, line 68: Cloud fixtures try to implement methods by itself without relying on Tree too much, so that our tests are separated from our code. But sometimes it is just too hard to do that so in stuff like ssh, we use SSHConnection for now, but I might switch to plain paramiko after all, to make it more isolated.

tests/remotes/base.py
```diff
@@ -1,7 +1,61 @@
...
+
+    def write_bytes(self, contents):
+        raise NotImplementedError
+
+    def read_text(self, encoding=None, errors=None):
+        raise NotImplementedError
+
+    def read_bytes(self):
+        raise NotImplementedError
+
```
efiop, line 32: All of these are your regular pathlib.Path's methods.

efiop: For now we have a bare minimum, but it is easy to extend with what we need in the future.

tests/remotes/__init__.py
efiop: Also, this really illustrates the need for https://github.com/iterative/dvc/issues/3920

tests/remotes/local.py
```diff
@@ -17,8 +17,12 @@ def get_url():
 
 
 @pytest.fixture
-def local_cloud():
-    yield Local()
+def local_cloud(make_tmp_dir):
+    ret = make_tmp_dir("local-cloud")
+    # FIXME: temporary hack
```
efiop, line 22: Yep, will fix in this PR.

tests/remotes/ssh.py
```diff
@@ -63,20 +66,73 @@ def get_url(user, port):  # pylint: disable=arguments-differ
...
+            username=TEST_SSH_USER,
+            key_filename=TEST_SSH_KEY_PATH,
+        )
+        try:
+            yield conn
+        finally:
+            conn.close()
+
+    def is_file(self):
+        with self._ssh() as _ssh:
```
efiop, line 92: Not very nice, might just create a connection and then add a closing logic to `close`, as it is really intended to be, and then register a fixture finaliser or something.

efiop:
For now running tests to see what i've missed in the ones that are not (easily) runnable locally.

tests/func/test_import_url.py
```diff
@@ -112,3 +112,34 @@ def test_import_url_with_no_exec(tmp_dir, dvc, erepo_dir):
     dvc.imp_url(src, ".", no_exec=True)
     dst = tmp_dir / "file"
     assert not dst.exists()
+
+
+@pytest.mark.parametrize(
+    "workspace",
+    [
+        pytest.lazy_fixture(cloud)
+        for cloud in ["local_cloud", "s3", "gs", "hdfs", "ssh"]
```
efiop, line 121: also forgot to add http here, will do.

tests/remotes/local.py
efiop: Keeping as is for now, will be removed after unification with TmpDir.

tests/func/test_repro.py
```diff
@@ -841,430 +818,6 @@ def test(self):
...
-
-        self.dvc.update([import_stage.path])
-        self.assertTrue(os.path.exists("import"))
-        self.assertTrue(filecmp.cmp("import", self.BAR, shallow=False))
-        self.assertEqual(self.dvc.status([import_stage.path]), {})
-
-        self.dvc.update([import_remote_stage.path])
-        self.assertEqual(self.dvc.status([import_remote_stage.path]), {})
-
-        stages = self.dvc.reproduce(cmd_stage.addressing)
```
efiop, line 983: Got rid of the repro test in favor of add/import/update/run(no-exec) tests because running commands is not trivial in a unified way, as there is no mechanism of translating our remote:// notation to a user-consumable one. So let's not have it for now (we are not loosing much) until https://github.com/iterative/dvc/issues/3920 is looked into, as this problem again illustrates the need for proper workspaces.

PR #4155, opened 2020-07-02
[RFC] tests: make read_text() work with directories

efiop:
This PR is just a showcase to see how we like it. I've converted a few
tests to it, but there will be much more to come.

```
assert (tmp_dir / "file") == "contents"
assert (tmp_dir / "dir").read_text() == {"subdir": {"file": "hello"}}
```

This approach is nicer because it is basically the same as what we feed
to `TmpDir.gen` and also allows us to use a declaratory asserts instead
of relying on fs files. The problem becomes much more apparent once
you get into remote workspaces, where we `trees_equal` doesn't work.

### DIFF
--- a/tests/dir_helpers.py
+++ b/tests/dir_helpers.py
@@ -221,6 +221,17 @@ class TmpDir(pathlib.Path):
         finally:
             self.scm.checkout(old)
 
+    def read_text(self, *args, **kwargs):  # pylint: disable=signature-differs
+        # NOTE: on windows we'll get PermissionError instead of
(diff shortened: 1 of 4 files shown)


efiop:
@Suor what do you think, is this following your philosophy as TmpDir creator? :slightly_smiling_face:

Suor:
I looked at how you:
```python
assert (tmp_dir / "file") == "contents"
```
and assumed you overloaded `__eq__` :)

Overall, I thought about that for a while, simply never needed that.

PR #4161, opened 2020-07-03
jsonschema: generate from schema

skshetry:

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -1,0 +1,258 @@
+{
+  "title": "dvc.yaml",
+  "type": "object",
+  "properties": {
+    "stages": {
+      "title": "Stages",
+      "description": "List of stages",
+      "type": "object",
+      "additionalProperties": {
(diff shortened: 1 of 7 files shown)


schema.json
```diff
@@ -0,0 +1,258 @@
+{
```
skshetry, line 1: Can also put this into https://github.com/iterative/pipeline-schema

scripts/jsonschema_gen.py
```diff
@@ -0,0 +1,109 @@
+try:
```
skshetry, line 1: Right now, it does not depend on any of the DVC. The schema is just ~~independent~~ duplicated. But, it cannot be used for validation, at least not yet. :slightly_frowning_face:

skshetry: Same here, if we don't want to maintain it, could push to separate repo.

skshetry:
Will do it in a separate repo.

PR #4166, opened 2020-07-04
Add more tests according to gitignore

### LINKED ISSUE #4103: .dvcignore is broken on negation when blacklisting all

I am not sure how far this extends to, but a lot of trials for `.dvcignore` failed when I blacklisted all and tried to whitelist some:

```
*
!scripts
```

```
*
!/scripts
```

```
/*
!/scripts/
```

```
/*
!/scripts/*
```

```
/*
!/scripts/**
```

```
/*
!scripts/
```

### What worked:
```
/*
!/scripts
```

Why do I feel suddenly the spelling of `scripts` is wrong? :smile:

karajan1001:
fix #4103

### DIFF
--- a/dvc/ignore.py
+++ b/dvc/ignore.py
@@ -44,11 +44,11 @@ class DvcIgnorePatterns(DvcIgnore):
 
     def __call__(self, root, dirs, files):
         files = [f for f in files if not self.matches(root, f)]
-        dirs = [d for d in dirs if not self.matches(root, d)]
+        dirs = [d for d in dirs if not self.matches(root, d, True)]
 
         return dirs, files
 
-    def matches(self, dirname, basename):
+    def matches(self, dirname, basename, is_dir=False):
         # NOTE: `relpath` is too slow, so we have to assume that both
         # `dirname` and `self.dirname` are relative or absolute together.
         prefix = self.dirname + os.sep
@@ -63,13 +63,19 @@ class DvcIgnorePatterns(DvcIgnore):
 
         if not System.is_unix():
             path = normalize_file(path)
-        return self.ignore(path)
+        return self.ignore(path, is_dir)
 
-    def ignore(self, path):
+    def ignore(self, path, is_dir):
         result = False
-        for ignore, pattern in self.ignore_spec:
-            if pattern.match(path):
-                result = ignore
+        if is_dir:
+            path_dir = f"{path}/"
+            for ignore, pattern in self.ignore_spec:
+                if pattern.match(path) or pattern.match(path_dir):
+                    result = ignore
+        else:
+            for ignore, pattern in self.ignore_spec:
+                if pattern.match(path):
+                    result = ignore
         return result
 
     def __hash__(self):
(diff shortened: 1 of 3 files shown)


karajan1001:
Two issues.

1. Other consecutive asterisks are considered regular asterisks and will match according to the previous rules.

> This is an issue of `PathSpec`.

2. It is not possible to re-include a file if a parent directory of that file is excluded. Git doesn’t list excluded directories for performance reasons, so any patterns on contained files have no effect, no matter where they are defined.
> Because path didn't distinguish with files in `dvcignore`. And `dir/` can't match pathname `dir`.

efiop:
@karajan1001 Sorry for the delay. This is outstanding work! Thank you so much for looking into this! 

> Other consecutive asterisks are considered regular asterisks and will match according to the previous rules.

Should we report it to pathspec and wait for the solution on their side? Or how do you think we should tackle it?

> It is not possible to re-include a file if a parent directory of that file is excluded. Git doesn’t list excluded directories for performance reasons, so any patterns on contained files have no effect, no matter where they are defined.

>  Because paths didn't distinguish with files in dvcignore. So dir/ can't match pathname dir.

Nice catch! So that one is solved and we are doing it the same way gitignore does it, right?

pared:
Seems like builds are failing.

karajan1001:
> Should we report it to pathspec and wait for the solution on their side? Or how do you think we should tackle it?
> 
> >
@efiop 
I'd like to report it to `PathSpec` and fix it myself, but only after work could I do it.

> Nice catch! So that one is solved and we are doing it the same way gitignore does it, right?
Yes, I had tested with git If the pattern format description is complex or not clear.
image
image
image
 

> Seems like builds are failing.
@pared  A bug from `pathspec` is waiting for fixing.

efiop:
> I'd like to report it to PathSpec and fix it myself, but only after work could I do it.

@karajan1001 Sounds great! :pray: So maybe let's remove or comment-out that particular test case for now, so we could merge?

karajan1001:
> > I'd like to report it to PathSpec and fix it myself, but only after work could I do it.
> 
> @karajan1001 Sounds great! 🙏 So maybe let's remove or comment-out that particular test case for now, so we could merge?

There is still something wrong with triple asterisks,  According to the document.

> Other consecutive asterisks are considered regular asterisks and will match according to the previous rules.

What the result from my computer shows that triple asterisks act like double ones.

image
image
image

And I found another problem.`data/*` can match `data/1/2` just like `data/**`. 

Seems that `Git`, the documents of `gitignore`, and `Pathspec` go different ways. 

I don't know which is the standard ones now.

pared:
Checked latest git master (2.26.0-rc1) and @karajan1001 is right, behavior of x/* is different that explicitly stated in gitignore docs. @karajan1001 that seems like a bug and potential contribution opportunity.

As to current build, I think we should comment out failing ones and wait till it gets done on PathSpec side. As to triple asterisk I would just leave it. If some will ever come to us with valid use case of triple asterisk other than 'match all', then we can consider it. There is no point in blindly following git docs.

PR #4183, opened 2020-07-08
tests: hdfs: use docker-based fixtures

### LINKED ISSUE #4054: tests: create azure/oss/hdfs/etc fixtures that would launch docker containers
Currently we set up azure/oss/hdfs servers before running the tests, which takes a long time. It would be great if those would be launched in pytest fixtures, so that it is easier to use locally and other tests could run while these are starting up.

efiop:
Fixed #4054


dvc/remote/hdfs.py
```diff
@@ -165,13 +165,14 @@ def get_file_hash(self, path_info):
         # NOTE: pyarrow doesn't support checksum, so we need to use hadoop
         regex = r".*\t.*\t(?P<checksum>.*)"
         stdout = self.hadoop_fs(
-            f"checksum {path_info.path}", user=path_info.user
+            f"checksum {path_info.url}", user=path_info.user
```
efiop, line 168: This worked only if your defaultFS matched the url you were connecting too.

PR #4192, opened 2020-07-11
pull: target message

dmpetrov:
I heard about this issue from 2 independent users. They didn't know that dvc support pull for a file.

`-R` option is minor and looks a bit confusing. Let's remove it from the target description.

- [x] Update docs if needed
https://github.com/iterative/dvc.org/pull/1384


efiop:
Need to update it in-sync with https://github.com/iterative/dvc.org/pull/1384/files . Including fetch/push/status. We'll take it from here.

PR #4196, opened 2020-07-12
remove: remove dvc.yaml and dvc.lock if they are empty

### LINKED ISSUE #4090: remove: remove dvc.yaml and dvc.lock if they are empty
https://github.com/iterative/dvc/pull/4074#issuecomment-648097445

```
$ cat dvc.lock
{}                                               
$ cat dvc.yaml
stages: {}                                       
```

nik123:
Fixes #4090 

https://github.com/iterative/dvc.org/issues/1564

### DIFF
--- a/dvc/dvcfile.py
+++ b/dvc/dvcfile.py
@@ -234,6 +234,8 @@ class PipelineFile(FileMixin):
         logger.debug("Removing '%s' from '%s'", stage.name, self.path)
         del d["stages"][stage.name]
         dump_yaml(self.path, d)
+        if len(d["stages"]) == 0:
+            self.remove(force=True)
 
 
 class Lockfile(FileMixin):
--- a/dvc/dvcfile.py
+++ b/dvc/dvcfile.py
@@ -233,9 +233,11 @@ class PipelineFile(FileMixin):
 
         logger.debug("Removing '%s' from '%s'", stage.name, self.path)
         del d["stages"][stage.name]
-        dump_yaml(self.path, d)
-        if len(d["stages"]) == 0:
-            self.remove(force=True)
+
+        if d["stages"]:
+            dump_yaml(self.path, d)
+        else:
+            super().remove()
 
 
 class Lockfile(FileMixin):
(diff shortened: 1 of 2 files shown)


dvc/dvcfile.py
```diff
@@ -234,6 +234,8 @@ def remove_stage(self, stage):
         logger.debug("Removing '%s' from '%s'", stage.name, self.path)
         del d["stages"][stage.name]
         dump_yaml(self.path, d)
+        if len(d["stages"]) == 0:
+            self.remove(force=True)
```
skshetry, line 238: Let's not dump if there's going to be no stages at all.

skshetry: Also, I think it's wrong to assume that lock file and dvc.yaml will always have the same no. of entries.

tests/func/test_dvcfile.py
```diff
@@ -285,6 +285,32 @@ def test_remove_stage_preserves_comment(tmp_dir, dvc, run_copy):
...
+    tmp_dir.gen(
+        "dvc.yaml",
+        textwrap.dedent(
+            """\
+            stages:
+                generate-foo:
+                    cmd: "echo foo > foo"
+                    outs:
+                    - foo"""
+        ),
```
skshetry, line 298: @nik123, we don't care about yaml structure here. It'd be better either to create stage via `run_copy` fixture or use `dump_yaml`.

dvc/dvcfile.py
nik123: > Let's not dump if there's going to be no stages at all.

Fixed

nik123: > Also, I think it's wrong to assume that lock file and dvc.yaml will always have the same no. of entries. I hope they'd be the same, but at the end, they are two different files (one of which, you're allowed to edit/remove yourselves).

I'm not sure I've got you right but now `dvc.lock` and `dvc.yaml` are removed independently. Please, check it out.

tests/func/test_dvcfile.py
nik123: Fixed

Issue #2799, opened 2019-11-16
ML experiments and hyperparameters tuning


dmpetrov:
UPDATE: Skip to https://github.com/iterative/dvc/issues/2799#issuecomment-650464000 for a summary and updated requirements.

## Problem

There are a lot of discussions on how to manage ML experiments with DVC. Today's DVC design allows ML experiments through Git-based primitives such as commits and branches. This works nicely for large ML experiments when code writing and testing required. However, this model is too heavy for the hyperparameters tuning stage when the user makes dozens of small, one-line changes in config or code. Users don't want to have dozens of Git-commits or branches.

## Requirements

A lightweight abstraction needs to be created in DVC to support hyperparameters-like tiny experiments without Git-commits. Hyperparameters tunning stage can be considered as a separate user activity outside of Git workflow. But the result of this activity still needs to be managed by Git preferably by a single commit.

High-level requirements to the hyperparameters tunning stage:
1. **Run.** Run dozens of experiments without committing any results into Git while keeping track of all the experiments. Each of the experiments includes a small config change or code change (usually, 1-2 lines).
2. **Compare.** A user should be able to compare two experiments: see diffs for code (and probably metrics)
2. **Visualize.** A user should be able to see all the experiments results: metrics that were generated. It might be some table with metrics or a graph. CSV table needs to be supported for custom visualization.
3. **Propagate.** Choose "the best" experiment (not necessarily the highest metrics) and propagate it to the workspace (bring all the config and code changes. Important: without retraining). Then it can be committed to Git. This is the final result of the current hyperparameter tunning stage. After that, the user can continue to work with a project in a regular Git workflow.
4. **Store.** Some (or all) of the experiments might be still useful (in additional to "the best" one). A user should be able to commit them to the Git as well. Preferably in a single commit to keep the Git history clean.
5. **Clean.** Not useful experiments should be removed with all the code and data artifacts that were created. A special subcommand of `dvc gc` might be needed.
6. [*] **Parallel.** In some cases, the experiments can be run in parallel which aligns with DVC parallel execution plans: #2212, #755. This might not be implemented now (in the 1st version of this feature) but it is important to support parallel execution by this new lightweight abstraction.
7. **Group.** Iterations of hyperparameters tuning might be not related to each other and need to be managed and visualized separately. Experiments need to be grouped somehow.

## What should NOT be covered by this feature?

[...]

PR #4199, opened 2020-07-13
Experiments proof of concept

pmrowla:
Related to #2799.

Implements initial POC support using a single git clone. UI is rough expected to change a lot.

* `dvc repro -e/--experiment` - creates a new experiment. Experiment is derived from the current HEAD commit in the user's repo and contains any changes in the user's current workspace.
* `dvc experiments show` - show experiment table. By default table includes the user's workspace, the current HEAD commit and any experiments derived from HEAD. `--all-commits/--all-tags/--all-branches` can be used to display more experiments.
* `dvc experiments diff` - diff 2 experiments. Currently just concats the output from `dvc metrics diff` and `dvc plots diff`.
* `dvc experiments checkout <experiment_rev>` - applies the changes from the specified experiment to the user's current workspace. If the changes would generate a git merge conflict, this command does nothing. `--force` can be used to replace the user's current workspace with the experiment changes. This has only been tested in cases where user checks out an experiment into a workspace with the original commit the experiment was derived from.
  **Note: using `--force` currently does a `git reset --hard` in your workspace**, the behavior you most likely want is to `git stash` your workspace changes before using `dvc experiments checkout`
* Experiments support is disabled by default.
    * Experiment related commands and args are hidden from the normal `dvc --help` output, but running `dvc experiments --help` can be used to show the usage for the experiments commands.
    * Experiments commands do nothing unless the `core.experiments` config option is set to true (or if DVC is run inside a testing/CI environment).

General thoughts/questions for next iteration:
* What is the desired behavior for `experiments diff` - mimic `metrics/params diff` formats or use the `show` table with added deltas?
* How to show data/code changes between experiments?
* `experiments checkout` feels a bit awkward to use, maybe it should be a `repro --experiment <rev>` command to reproduce a specific experiment instead?
* What kind of workflow to use for "promoting" an experiment commit to a "real" commit or branch in the main repo?

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -19,19 +19,20 @@ def _update_names(names, items):
             names.add(name)
 
 
-def _collect_names(experiments):
+def _collect_names(all_experiments):
     metric_names = set()
     param_names = set()
 
-    for exp in experiments.values():
-        _update_names(metric_names, exp.get("metrics", {}).items())
-        _update_names(param_names, exp.get("params", {}).items())
+    for _, experiments in all_experiments.items():
+        for exp in experiments.values():
+            _update_names(metric_names, exp.get("metrics", {}).items())
+            _update_names(param_names, exp.get("params", {}).items())
 
     return sorted(metric_names), sorted(param_names)
 
 
 def _collect_rows(
-    experiments, metric_names, param_names, include_rev=False, precision=None
+    base_rev, experiments, metric_names, param_names, precision=None
 ):
     from flatten_json import flatten
     from dvc.command.metrics import DEFAULT_PRECISION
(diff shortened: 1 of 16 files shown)


pmrowla:
Not really ready yet, but some basic things work:

[asciicast](https://asciinema.org/a/ikIHTJZPgpOfMR6uGM9UXblyX)

dvc/command/experiments.py
```diff
@@ -0,0 +1,358 @@
...
+
+def add_parser(subparsers, parent_parser):
+    EXPERIMENTS_HELP = "Commands to display and compare experiments."
+
+    experiments_parser = subparsers.add_parser(
+        "experiments",
+        parents=[parent_parser],
+        description=append_doc_link(EXPERIMENTS_HELP, "experiments"),
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
```
pmrowla, line 238: `help` param intentionally omitted here so that the command is hidden in the default `dvc`/`dvc --help` usage output

skshetry: In the long term, we should think of some way for this. Even though we had it hidden, users were still trying out new 1.0 docs in 0.94 `dvc`.  Something like `[experimental]` flags come to mind but would be difficult to read it at this higher-level (`add_parser`), or could introduce `DVC_EXPERIMENTAL_XXX_ENABLE` envs or maybe, even simpler name mangling till it's ready.

Not a big issue right now, just sharing the issue that I had. :slightly_smiling_face:. Definitely something we should consider, as we mostly work on HEAD.

pmrowla:
First post has been updated with details. All behavior changes are disabled unless `core.experiments` config option is set to true, so we should be able to merge and do future iterations on the experiments feature without breaking anything else.

kenahoo:
This feature piqued my interest.  Is there a writeup somewhere of what an "experiment" is, and how it relates to a pipeline and its results?  I didn't see any docs/explanation in the PR itself.

efiop:
@kenahoo Not yet, this is just a poc and is under active development. There are some discussions about it in https://github.com/iterative/dvc/issues/2799 .

PR #4219, opened 2020-07-16
Bump networkx

faucct:
To avoid conflict with https://github.com/aws-cloudformation/cfn-python-lint


pared:
@faucct so the problem here is that cfn uses at least `2.4`?

faucct:
Yep.

efiop:
Thanks!

faucct:
Can this get a release, please?

efiop:
@faucct Sure! We'll release a new version later today or tomorrow. We have some critical changes in the pipeline, so can't really release earlier. Thank you for the feedback!

faucct:
Good!

PR #4220, opened 2020-07-16
bump azure-storage-blob version to 12.3.2

### LINKED ISSUE #3546: remote: upgrade Azure Blob storage version
Hi!

At the moment `dvc[azure]` has as a requirement `azure = ["azure-storage-blob==2.1.0"]`

Recently there Python SDKs were massively improved to be more Pythonic and efficient. I was wondering if it would be possible to allow for the use of the latest version (12.1). Or if there are any blockers for this I can help with?

dtrifiro:
Fixes #3546

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -87,7 +87,7 @@ install_requires = [
 gs = ["google-cloud-storage==1.19.0"]
 gdrive = ["pydrive2>=1.4.14"]
 s3 = ["boto3>=1.9.201"]
-azure = ["azure-storage-blob==2.1.0", "knack"]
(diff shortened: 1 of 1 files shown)


skshetry:
@dtrifiro, is this backward compatible? I see that the version jumped from 2 to 12, so probably no. Let's see 
what happens with the tests.

dtrifiro:
Hi @skshetry, I see that the build failed (`The command "./scripts/ci/check_patch.sh" exited with 1.`), I'll see try to see what fails locally and get back to you.

dtrifiro:
Seems it's not backwards compatible, I'll see what I can do to update the azure storage version in the coming weeks. Thanks

Issue #4175, opened 2020-07-07
s3: provide a way to use key id and key secret directly


Suor:
For now the only way to pass credentials to S3 is using credentials file and the env var. This is very inconvenient when trying to use dvc as a library. For now I hacked it like:

```python
def pull(repo):
    # In memory config update to avoid validation
    repo.config["core"]["remote"] = "vdefault"
    repo.config["remote"]["vdefault"] = {"url": ..., "key_id": ..., "key_secret": ...}

    fix_s3()
    repo.pull()


from funcy import once, monkey, cached_property, wrap_prop  # noqa

@once
def fix_s3():
    import threading
    from dvc.remote.s3 import S3RemoteTree

    @monkey(S3RemoteTree, "s3")
    @wrap_prop(threading.Lock())
    @cached_property
    def s3(self):
        import boto3

        session = boto3.session.Session(
            profile_name=self.profile,
            region_name=self.region,
            # Pass these two from config to session
            aws_access_key_id=self.config.get("key_id"),
            aws_secret_access_key=self.config.get("key_secret"),
        )

        return session.client("s3", endpoint_url=self.endpoint_url, use_ssl=self.use_ssl)
```

Would be nice to have it inside `dvc.remote.S3RemoteTree`.

efiop:
@Suor Have you considered changing the env for that call?

Also, I suppose your request is only about internal api, right?

Suor:
Changing env for the process will leak it to other threads and also may leak to subsequent calls. Starting a new process is a solution, but a heavy one.

I am using `repo.cloud.pull()`, but `repo.pull()` should work the same.

Suor:
Ideally I want key_id/key_secret pair be added to config. Any reason you want to avoid it?

efiop:
> Ideally I want key_id/key_secret pair be added to config. Any reason you want to avoid it?

Same reason as with any plain-text passwords :slightly_smiling_face: We could make it api-only, but feels like a hack, still. I guess the only current nice way to do it is to just use a tempfile, dump keys into and and then use it as credentailpath in the config.

Suor:
That would work too. However, if we want api to be usable we should stop doing things like that) Non-saveable config option is fine with me too. Hack or not.

Suor:
[...]

PR #4224, opened 2020-07-17
S3: Use key_id and key_secret directly

farizrahman4u:
Fixes https://github.com/iterative/dvc/issues/4175

@efiop


efiop:
Hi @farizrahman4u !

Thank you for the PR! Please take a look at our contribution guide (it was in the PR hints). Right now our patch checks are failing, because you didn't have pre-commit hooks installed. :slightly_smiling_face:

dvc/tree/s3.py
```diff
@@ -54,18 +54,32 @@ def __init__(self, repo, config):
...
         shared_creds = config.get("credentialpath")
         if shared_creds:
             os.environ.setdefault("AWS_SHARED_CREDENTIALS_FILE", shared_creds)
+        self._validate_config()
+
+    def _validate_config(self):
+        if bool(self.key_id) != bool(self.key_secret):
+            raise DvcException(
+                "Provide either both or none of `s3_key_id` and `s3_key_secret`."
+            )
```
efiop, line 69: Does `boto3.session` require you pass both at the same time? Or will find the other from env/configs if you only give it one of the keys?

dvc/tree/s3.py
```diff
@@ -54,18 +54,32 @@ def __init__(self, repo, config):
...
-        session = boto3.session.Session(
-            profile_name=self.profile, region_name=self.region
-        )
+        session_opts = dict(profile_name=self.profile, region_name=self.region)
+
+        if self.key_id and self.key_secret:
+            session_opts["aws_access_key_id"] = self.key_id
+            session_opts["aws_secret_access_key"] = self.key_secret
+
+        session = boto3.session.Session(**session_opts)
```
efiop, line 82: Let's add a simple unit test to`tests/unit/remote/test_s3.py`

dvc/tree/s3.py
```diff
@@ -54,18 +54,32 @@ def __init__(self, repo, config):
 
         self._append_aws_grants_to_extra_args(config)
 
+        self.key_id = config.get("s3_key_id")
+        self.key_secret = config.get("s3_key_secret")
+
```
efiop, line 59: I don't think this is going to work, as you didn't enable these fields in `dvc/config.py`.

efiop:
Also, please feel free to mark your PR as WIP/draft if it is not ready for review yet :slightly_smiling_face: Thanks!

farizrahman4u:
@efiop CI seems to be cloning the main repo instead of the fork?

skshetry:
Looks like it is trying to run restyled's PR and is failing as the branch is no longer there. 
@farizrahman4u, please rebase this PR. Probably, that will fix the issue.

farizrahman4u:
@efiop Done.

dvc/config.py
```diff
@@ -149,6 +149,8 @@ class RelPath(str):
                     "profile": str,
                     "credentialpath": str,
                     "endpointurl": str,
+                    Optional("s3_key_id", default=""): str,
+                    Optional("s3_key_secret", default=""): str,
```
efiop, line 153: These are optional by default, you don't need to set them
```suggestion
                    "s3_key_id": str,
                    "s3_key_secret": str,
```

farizrahman4u: done

tests/unit/remote/test_s3.py
```diff
@@ -6,6 +6,8 @@
 bucket_name = "bucket-name"
 prefix = "some/prefix"
 url = f"s3://{bucket_name}/{prefix}"
+key_id = "key_id"
+key_secret = "key_secret"
```
efiop, line 10: a bit worried of some nasty bug like that :smiley: 
```suggestion
key_id = "key-id"
key_secret = "key-secret"
```

dvc/config.py
```diff
@@ -149,6 +149,8 @@ class RelPath(str):
                     "profile": str,
                     "credentialpath": str,
                     "endpointurl": str,
+                    "s3_key_id": str,
+                    "s3_key_secret": str,
```
efiop, line 153: Kinda have an urge to just do
```suggestion
                    "endpointurl": str,
                    "key_id": str,
                    "key_secret": str,
```
or, even better:
```suggestion
                    "endpointurl": str,
                    "access_key_id": str,
                    "secret_access_key": str,
```
to make it easier to find and more intuitive in general. Considering that `oss` has
```
                    "oss_key_id": str,
                    "oss_key_secret": str,
                    "oss_endpoint": str,
```
already (which we could consider unifying in the future).

efiop:
Btw, @farizrahman4u , in the PR header that you've removed, there was a bulletpoint about docs. In particular, we need to add the new config options to https://dvc.org/doc/command-reference/remote/modify (there is an edit button). Could you please submit a docs PR after we are done with config option naming here? :slightly_smiling_face:

farizrahman4u:
@efiop done renames and docs pr : https://github.com/iterative/dvc.org/pull/1589

dvc/tree/s3.py
```diff
@@ -54,6 +54,9 @@ def __init__(self, repo, config):
 
         self._append_aws_grants_to_extra_args(config)
 
+        self.access_key_id = config.get("access_key_id")
+        self.secret_access_key = config.get("secret_access_key")
```
Suor, line 58: Why do we assign it instead of using `self.config` directly later?

farizrahman4u: Following the pattern in other remotes (OSS, GDrive etc).

Issue #3436, opened 2020-03-04
Warn users when traversing special directories, recommend to ignore 


skshetry:
Currently, we ignore `.dvc`, `.git` and `.hg` directories by default. Still, we might be traversing through other special directories such as: `node_modules`, `.venv/venv/env`, `__pycache__`, `dist/eggs/wheels`, `htmlcov`, `docs`, `.mypy_cache`, `.pytest_cache`, etc.

We could throw warning if the users have not ignored these kind of directories and recommend them to add to `.dvcignore`. We could start with a small set of such directories.

### Context:

> Exactly. There can be tons of special/hidden dirs you don't want DVC to traverse (.idea/ .project/ 
> .env/ .ds_store/ venv/), and you can just add them to .gitignore or .dvcignore. Why are we secretly
> integrating with Mercurial if we don't actually support it? What if someone wants to version a .hg 
> dir? (edge case I know)
>
> It just seems very arbitrary to me, but OK maybe I'm missing something...

_Originally posted by @jorgeorpinel in https://github.com/iterative/dvc/pull/3257#discussion_r385515891_

shcheklein:
@skshetry please, consider adding a link to the discussion - I think it can be valuable.

PR #4229, opened 2020-07-17
Special dir warning #3436

aerubanov:
Added a warning message when traversing special directories and a test for that. Also removed unused arguments in a couple of tests in tests/func/test_ignore.py


dvc/ignore.py
```diff
@@ -148,6 +148,37 @@ def __eq__(self, other):
...
+        "__pycache__",
+        "dist",
+        "eggs",
+        "wheels",
+        "htmlcov",
+        "docs",
+        ".ipynb_checkpoints",
+        ".vscode",
+        ".github",
+    }
```
efiop, line 168: Hm, makes me wonder if we should really just ignore them by default (like we do with .git and .hg), and if some user really needs these, he could `!something` in dvcignore to include it.

Or, we might consider generating a default .dvcignore (kinda like github generates .gitignore when you create a project) to make it extra explicit. Though this is simple on `init`, but managing this stuff in existing projects is not too simple :slightly_frowning_face: 

I'm just worried about these warnings getting too annoying. Plus, unless I'm missing something, ignore execution order is not deterministic (that's actually is a bigger problem in general for us), so you never know if this filter will be executed before or after .dvcignores in which user has ignored a special dir explicitly. :slightly_frowning_face:

efiop: One more thing that this makes me think about is having some kind of safe guard, that would, for example, record the amount of time it took to walk through the directory and warn if it didn't find anything but it took, say, more than 5 sec.

aerubanov: This issue was opened after this conversation https://github.com/iterative/dvc/pull/3257. It was decided that it is better to warn the user if a special directory is found than to ignore it by default. But annoyance from warnings can be a problem. Maybe we should make it possible to disable this warning? From the other hand, the default .dvcignore also good idea because the ignore will be obvious to the user.

pared: What makes idea of default `.dvcignore` compelling, is the fact that is not only explicit, but also editable by user.

If we decide to go with warnings, they definitely need turn off mechanism, just like we did with `slow_link_guard`.

aerubanov: Now I soppouse that the default ``` .dvcignore```, is better option. It`s more explicite and easy to configure. Also the list of special dirs always will be incomplite, we can not add all posiible variants here. And default ``` .dvcignore``` provide additional example for users how to deal with such cases.

 @efiop, @pared, if you are agree with me, then I will change the code. Should I close this PR and open a new one later, or commit right here?

shcheklein: 👍 I love the idea of an explicit `.dvcignore`

pared: I wonder whether it should be default. Maybe we should add some option to init, or dedicate special command for that? It seems to me that `.dvcignore` was supposed to let user decide what is part of his project. If we always create default `.dvcignore` on init, we make it our choice. @efiop @aerubanov, what do you think?

aerubanov: I think that default ```.dvcignore``` should consist comment with link to the dvcignore documentation. And we should add option for init to filling ```.dvcignore``` with some content. Not all list of special folder, rather based on type of project specified by user.  Something like github gitignore template https://github.com/github/gitignore, but with fewer options ofcorse.

pared: @aerubanov I think you are right on this one.

aerubanov:
I add default .dvcignore with link on docs  and option for init to filling .dvcignore with some content based on project type specified by user. If there is no need for further changes to the functionality, I will add tests and changes to the documentation.

dvc/ignore_templates/__init__.py
```diff
@@ -0,0 +1,2 @@
+# templates for .dvcignore files
```
efiop, line 1: This doesn't really have to be a module, as there is no actual code. You can either keep it as datafiles or maybe actually make these into the python shape by declaring these as strings. The latter way is easier to manage, as things like setup.py and pyinstaller will have a much easier time collecting these automatically.

dvc/ignore_templates/java
```diff
@@ -0,0 +1,23 @@
+# Compiled class file
```
efiop, line 1: Hm, I also wonder if these are really worth stealing directly from https://github.com/github/gitignore . They would be hard to maintain, really. But I'm not able to see a better way right now...

dvc/repo/__init__.py
```diff
@@ -185,10 +185,22 @@ def find_dvc_dir(cls, root=None):
...
+    ):
         from dvc.repo.init import init
 
-        init(root_dir=root_dir, no_scm=no_scm, force=force, subdir=subdir)
+        init(
+            root_dir=root_dir,
+            no_scm=no_scm,
+            force=force,
+            subdir=subdir,
+            ignore_template_list=ignore_template_list,
```
efiop, line 202: People will probably want these even in existing projects, and they will probably forget to do this during the init phase. So we could add another command to install these, but maybe it is actually better to just document that people might want to get .dvcignore from https://github.com/github/gitignore  and use it.

This also makes me wonder if we should should really somehow reuse existing .gitignores (e.g. to not search for dvc-files and .dvcignores in .gitignore'd directories). But I suppose many people also want to exclude these dirs from dirs they are `dvc add`ing (or tracking), which doesn't play well, because we gitignore the data we are adding. :thinking:

aerubanov: @efiop, Ok. May be we should add command for downloading template content from https://github.com/github/gitignore and pasting it in ```.dvcignore```? Then, we need not to store it in dvc code base and mantain, and user need not to do it by hand. 

And also we can add second option for this command to read existing ```,gitignore```, exlude all records that was added by ```dvc add``` and paste it in ```.dvcignore```.  Is it possible to easily get a list of all tracked files? (I think it should be, I just don't know where to look).

What do you think of these two possibilities?

efiop: > @efiop, Ok. May be we should add command for downloading template content from https://github.com/github/gitignore and pasting it in .dvcignore? Then, we need not to store it in dvc code base and maintain, and user need not to do it by hand.

@aerubanov I don't think it is worth creating a special command. It is not hard to just go to a link in the docs and copypaste contents to .dvcignore (or use wget/curl/etc).

> And also we can add second option for this command to read existing ,gitignore, exlude all records that was added by dvc add and paste it in .dvcignore. Is it possible to easily get a list of all tracked files? (I think it should be, I just don't know where to look).

That sounds a bit complex :( We had a discussion with @pmrowla about it, and maybe we could just use `.gitignore` directly, the only thing that is problematic is not dvcignoring dvc-tracked files, but that should be solvable by telling dvcignore about them  when we collect the dag. We will soon have a pre-requisite PR that notifices dvcignore about those outputs for performace reasons and we might get back to this idea right after it.

aerubanov: Ok, i will make the necessary changes.

dvc/repo/init.py
```diff
@@ -43,7 +44,29 @@ def _welcome_message():
     logger.info(msg)
 
 
-def init(root_dir=os.curdir, no_scm=False, force=False, subdir=False):
+def generate_dvcignore(templates_list):
+    with open(".dvcignore", "w") as f:
+        f.write(
+            "# This is .dvcignore file."
+            " See more https://dvc.org/doc/user-guide/dvcignore\n"
+        )
```
skshetry, line 52: Good idea to create a default dvcignore. It'll make people aware of it. :+1: to that.

Regarding the comment, better to explain what it does. `This is .dvcignore file.` is clear from the filename.

Something like following, perhaps?
```suggestion
        f.write(
            "# add patterns of files dvc should ignore, can improve performance\n"
            "# Learn more at https://dvc.org/doc/user-guide/dvcignore\n"
        )
```

efiop: @skshetry Please see my comment above. Maybe we should just use .gitignore directly. Would love to know what you think about that.

aerubanov:
@efiop, @skshetry . I made the changes that we discussed above. Is there more to be done?

efiop:
Thank you @aerubanov ! :pray:

dvc/ignore.py
```diff
@@ -314,3 +314,18 @@ def check_ignore(self, target):
...
+
+
+def init(path):
+    dvcignore = os.path.join(path, DvcIgnore.DVCIGNORE_FILE)
+    if os.path.exists(dvcignore):
+        return dvcignore
+
+    with open(dvcignore, "w") as fobj:
+        fobj.write(
+            "# Add patterns of files dvc should ignore, which could improve\n"
```
jorgeorpinel, line 326:
```suggestion
            "# Add patterns of files DVC should ignore, which could improve\n"
```

PR #4240, opened 2020-07-19
dvc: get rid of BaseTree in favor of BaseRemoteTree

efiop:
Continuing #4050


dvc/scm/git/tree.py
```diff
@@ -52,7 +53,9 @@ def dvcignore(self):
         self.use_dvcignore = True
         return ret
 
-    def open(self, path, mode="r", encoding="utf-8"):
+    def open(
+        self, path, mode="r", encoding="utf-8"
+    ):  # pylint: disable=arguments-differ
```
efiop, line 58: These `arguments differ` will be addressed a bit later. We also need to stop setting mode and encoding like that, and instead do the same thing pathlib's Path does (we've started doing it in tests/remotes/ already).

Issue #4139, opened 2020-06-30
push: OOM when uploading 1.5TB(3.5Mfiles) dataset to local remote


lenrys29:
## Bug Report

**Output of `dvc version`:**
```console
DVC version: 1.0.1
Python version: 3.7.1
Platform: Linux-4.19.0-9-amd64-x86_64-with-debian-10.4
Binary: True
Package: deb
Supported remotes: azure, gdrive, gs, hdfs, http, https, s3, ssh, oss
Cache: reflink - supported, hardlink - supported, symlink - supported
Filesystem type (cache directory): ('xfs', '/dev/mapper/vg_data-lv_data')
Repo: dvc, git
Filesystem type (workspace): ('xfs', '/dev/mapper/vg_data-lv_data')
```

### Please provide information about your setup

We've got a single server with 1 big partition of 2 TO.

We are using git + dvc to manage our datasets, training, etc.

For now, we try to push 1,5TB (3,18M files) on 1 single repo to create our 1st original dataset

We are using reflink and xfs for storage.

SSH + local  usage for : repo / git / training

But when we are pushing dvc crashed :

```console
2020-06-29 21:36:34,255 DEBUG: Uploading '.dvc/cache/db/4b681a2fcb2d45436c2bc9420cb116' to '../../../../datasets/upper/idl/db/4b681a2fcb2d45436c2bc9420cb116'
2020-06-29 21:36:34,260 DEBUG: Uploading '.dvc/cache/8f/9081ae7243d5d9b2f63c6d18b27678' to '../../../../datasets/upper/idl/8f/9081ae7243d5d9b2f63c6d18b27678'
2020-06-29 21:36:34,603 DEBUG: Uploading '.dvc/cache/fd/7a74c5c5bc38a23845b0e79671b81f' to '../../../../datasets/upper/idl/fd/7a74c5c5bc38a23845b0e79671b81f'
2020-06-29 21:36:35,078 DEBUG: Uploading '.dvc/cache/9c/b4e48c89859fe5ded008d8f1e0e74a' to '../../../../datasets/upper/idl/9c/b4e48c89859fe5ded008d8f1e0e74a'
2020-06-29 21:36:35,095 DEBUG: Uploading '.dvc/cache/f7/6bd859740457c7b561d5fd5425083d' to '../../../../datasets/upper/idl/f7/6bd859740457c7b561d5fd5425083d'
[...]

PR #4242, opened 2020-07-19
Optimize dvcignore

karajan1001:
optimize #3869 


dvc/repo/__init__.py
```diff
@@ -99,7 +99,7 @@ def __init__(self, root_dir=None, scm=None, rev=None):
             )
 
         self.dvc_dir = os.path.join(self.root_dir, self.DVC_DIR)
-        self.config = Config(self.dvc_dir, tree=self.tree)
```
karajan1001, line 102: I don't know why Config should use a tree with dvcignore.

efiop: Good point. Looks like a leftover.

efiop: @karajan1001 ah, wait, it wasn't about dvcignore actually. But about config being able to access config from the git history (GitTree). So we need to rollback this change. Was it causing particular problems?

karajan1001: Ok.
Now We had three ignore levels. 
1. Nothing.
2. don't ignore files inside `.dvc` (used for DVC config)
3. ignore all `.dvc` , `.git` , etc, (used by users)

efiop: @karajan1001 You are right that config doesn't need dvcignore. We just use `self.tree` here as both convenience and necessity. There are really two levels: nothing or `with dvcignore` :slightly_smiling_face: Config clearly doesn't need dvcignore as you've pointed out, but it needs to use repo's tree (of sorts) so it works when we switch to other branches.

efiop: Seems like it would be nice to have an ability to disable dvcignore on particular tree calls. E.g.
```
with tree.open(..., use_dvcignore=False)
```
or maybe some other convenient way. As I've mentioned, we are in the middle of tree refactorings and we've been discussing whether `exists/isdir/isfile/open` need to care about dvcignore at all (tree.walk clearly does need to care about dvcignore). What do you think?

efiop:
For the record: that strange bug with the number of calls of file_md5 was caused by https://github.com/iterative/dvc/pull/4244 . Please rebase and it will go away :slightly_smiling_face:

dvc/ignore.py
```diff
@@ -176,60 +115,65 @@ def is_ignored_file(self, _):
...
-        for ignore in self.ignores:
-            dirs, files = ignore(root, dirs, files)
+        ignore_pattern = self._get_tire_pattern(root)
+        return ignore_pattern(root, dirs, files)
 
-        return dirs, files
+    def _get_tire_pattern(self, dirname):
+        ignore_pattern = self.ignores_trie_tree.longest_prefix(dirname).value
+        if not ignore_pattern:
+            return DvcIgnorePatterns([], dirname)
```
pared, line 154: Shouldn't we just return None here and react properly?
Now we will keep empty `DvcIgnorePatterns` for every directory in a tree, that does not containe `.dvcignore`. I think in big project, that might be causing some memory issues. And we already have [problems with that](https://github.com/iterative/dvc/issues/4139).

karajan1001: @pared 
Thank you,  I know it's not a good design, but can't find an elegant solution.

We are now using a trie tree, and add a default pattern to the root of the Repo. So every `dirname` inside the repo would not get an empty pattern, they at least would return the pattern of the Repo's root. 

But if we call with some `dirname` outside the Repo, then it would return a None. Although no code would do that now(`_outside_repo` prevents it), I still add this logic to make it more robust. And I think an exception handling logic inside `_get_tire_pattern` is better than handling it in every use case.

The best solution I thought is to add something like `DVCIgnorePatternNoop` to the root of the fs.  And it could be done in `Unix` like system where we have `/`, but the `Windows` uses a different logic, there is no root of `C:\`, `D:\` etc.

The memory usage would not be a question in this situation, because the trie tree only had nodes where a `.dvcignore` file exists. Other directories in a tree would fall to these nodes.

karajan1001:
~~~~~~~~~~~~~~~~~~~~~ Stack of Thread-32 (140416213042944) ~~~~~~~~~~~~~~~~~~~~~
  File "/opt/python/3.8.0/lib/python3.8/threading.py", line 890, in _bootstrap
    self._bootstrap_inner()
  File "/opt/python/3.8.0/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/opt/python/3.8.0/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/home/travis/virtualenv/python3.8.0/lib/python3.8/site-packages/mockssh/server.py", line 56, in handle_client
    command = self.command_queues[channel.chanid].get(block=True)
  File "/opt/python/3.8.0/lib/python3.8/queue.py", line 170, in get
    self.not_empty.wait()
  File "/opt/python/3.8.0/lib/python3.8/threading.py", line 302, in wait
    waiter.acquire()

pared:
@karajan1001 when is it showing up? Running tests?

karajan1001:
@pared
In (https://travis-ci.com/github/iterative/dvc/jobs/363240722)
After I fixed some typo error.

dvc/ignore.py
pared: @karajan1001 
Thanks for the explanation, I get it now. 
I would still say that more readable would be to just react on `None`, due to understanding that one needs to grasp during parsing the code. But that seems more like a personal opinion, so let's not dwell into it.

pared:
@karajan1001
I restarted the build, seems fine now.

dvc/ignore.py
karajan1001: @pared 
As `_get_tire_pattern` is a private function, I think it is OK to grasp `None` outside it, as the callings are all under control.

karajan1001:
@pared 

Excuse me,  how can I restart build myself? 
The network is unstable？

ERROR tests/func/test_api.py::test_get_url[oss] - Exception: Command docker-compose -f "C:\Users\travis\build\iterative\dvc\tests\docker-compose.yml" -p "pytest-dvc-test" up --build -d returned 1: """Creating network "pytest-dvc-test_default" with the default driver

pared:
@karajan1001 
You would need to have access to our builds. @efiop correct me if I am wrong but isn't it restricted to members of the organization?

efiop:
test failures are unrelated. Thank you! :pray:

Issue #3438, opened 2020-03-04
import: git sparse-checkout


casperdcl:
`dvc import https://some/git/repo/ some_file` should probably do a sparse checkout.

Currently looks like it does a full clone:

https://github.com/iterative/dvc/blob/05cc023164766fbbec91642d7d8af513c8d4079b/dvc/stage.py#L941 would call https://github.com/iterative/dvc/blob/05cc023164766fbbec91642d7d8af513c8d4079b/dvc/dependency/repo.py#L73-L81

On a related note, is the repo locally cached by default https://github.com/iterative/dvc/blob/05cc023164766fbbec91642d7d8af513c8d4079b/dvc/dependency/repo.py#L78-L79 so that `dvc update` won't re-clone?

efiop:
Related to https://github.com/iterative/dvc/issues/3473

Suor:
Local caches only used within one dvc execution or one python session. 

Sparse checkout or at least shallow checkouts might be used, this will complicate out `external_repo()` abstraction, including its caching. But looks like there is no way around.

casperdcl:
I think we had a discussion with @efiop and @shcheklein about which is more important to fix first (persistent repo cache or sparse checkout) and there was no clear outcome. Will open a new issue for persistent repo cache...

skshetry:
We don't have a persistent repo checkout, but we do shallow clone these days. Closing...

PR #4246, opened 2020-07-20
erepo: --rev BRANCH: shallow clone

### LINKED ISSUE #3473: 'dvc get' is cloning the full bitbucket repo when only one branch is needed

pmrowla:

Follow up to #3585.
Should fix #3473.

* erepo now clones with `--no-single-branch --depth=1 --branch=<branch>` when initialized with a valid branch or tag name (i.e. for `get/import --rev`)
    * if erepo is init'd with `for_write=True` (i.e. dvcx) we use the old full clone + writeable copy behavior
    * If shallow clone fails, we fallback to default full clone behavior
* If we reuse an already cloned repo and don't have the necessary revision, we expand the clone with `git fetch --unshallow` and then treat it like normal full clone

This does not address these other potential changes to erepo clone behavior:
* persistent cache #3496 
* sparse checkouts #3438

### DIFF
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -289,6 +289,16 @@ def _clone_default_branch(url, rev, for_write=False):
                     git.repo.git.fetch(unshallow=True)
                     shallow = False
                     CLONES[url] = (clone_path, shallow)
+                    if git.repo.head.is_detached:
+                        # If this is a detached head (i.e. we shallow
+                        # cloned a tag) checkout the default branch
+                        origin_refs = git.repo.remotes["origin"].refs
+                        ref = origin_refs["HEAD"].reference
+                        branch_name = ref.name.split("/")[-1]
+                        branch = git.repo.create_head(branch_name, ref)
+                        branch.set_tracking_branch(ref)
+                        branch.checkout()
+
                 logger.debug("erepo: git pull '%s'", url)
                 git.pull()
         else:
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -270,6 +270,19 @@ def _cached_clone(url, rev, for_write=False):
     return repo_path
 
 
+def _unshallow(git):
+    git.repo.git.fetch(unshallow=True)
+    if git.repo.head.is_detached:
+        # If this is a detached head (i.e. we shallow cloned a tag) switch to
+        # the default branch
+        origin_refs = git.repo.remotes["origin"].refs
+        ref = origin_refs["HEAD"].reference
+        branch_name = ref.name.split("/")[-1]
+        branch = git.repo.create_head(branch_name, ref)
+        branch.set_tracking_branch(ref)
+        branch.checkout()
+
+
 @wrap_with(threading.Lock())
 def _clone_default_branch(url, rev, for_write=False):
     """Get or create a clean clone of the url.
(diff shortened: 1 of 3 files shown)


dvc/external_repo.py
```diff
@@ -31,7 +32,7 @@ def external_repo(url, rev=None, for_write=False):
     logger.debug("Creating external repo %s@%s", url, rev)
     path = _cached_clone(url, rev, for_write=for_write)
     if not rev:
-        rev = "HEAD"
+        rev = "refs/remotes/origin/HEAD"
```
pmrowla, line 35: Since `rev=None` is supposed to mean tip on the default branch we want to use HEAD from the origin repo, not HEAD from our local clone working copy. Local HEAD points to the tip of whatever branch we first cloned from (which may not be the default branch), but here we want the tip of the default branch.

dvc/scm/git.py
```diff
@@ -109,14 +109,23 @@ def clone(url, to_path, rev=None):
...
                     url,
                     to_path,
                     env=env,  # needed before we can fix it in __init__
                     no_single_branch=True,
                     progress=pbar.update_git,
                 )
+                if shallow_branch is None:
+                    tmp_repo = clone_from()
+                else:
+                    tmp_repo = clone_from(branch=shallow_branch, depth=1)
```
casperdcl, line 128: seems essentially the same method as #3585 - just checking if @Suor has the same concern here as with https://github.com/iterative/dvc/pull/3585/files#r402728411

dvc/scm/git.py
```diff
@@ -5,7 +5,7 @@
 import shlex
 
 import yaml
-from funcy import cached_property
+from funcy import cached_property, partial
```
casperdcl, line 8: :eyes: any reason not `functools.partial`?

dvc/scm/git.py
pmrowla: In this case the caller has to explicitly specify that they want a shallow clone now, rather than making a choice based on the `rev` parameter (which we don't use when cloning from erepo).

dvc/scm/git.py
pmrowla: afaik `funcy.partial` just re-imports `functools.partial`. We use both elsewhere in DVC but I guess we should probably pick one for consistency.

casperdcl: doesn't make sense using things via indirect import if they're really the same thing

casperdcl: If it's easy to do, sounds like we may need some sort of linter (e.g. `pylint`) to detect indirect imports?

dvc/external_repo.py
Suor: I would be happy to see such things commented about in dvc code.

dvc/external_repo.py
```diff
@@ -265,36 +266,66 @@ def _cached_clone(url, rev, for_write=False):
...
+    git.repo.git.fetch(unshallow=True)
+    if git.repo.head.is_detached:
+        # If this is a detached head (i.e. we shallow cloned a tag) switch to
+        # the default branch
+        origin_refs = git.repo.remotes["origin"].refs
+        ref = origin_refs["HEAD"].reference
+        branch_name = ref.name.split("/")[-1]
+        branch = git.repo.create_head(branch_name, ref)
+        branch.set_tracking_branch(ref)
+        branch.checkout()
```
Suor, line 283: The code in this file goes from high level to lower, so this one should be placed after its use.

dvc/external_repo.py
```diff
@@ -265,36 +266,66 @@ def _cached_clone(url, rev, for_write=False):
...
             # Do not pull for known shas, branches and tags might move
             if not Git.is_sha(rev) or not git.has_rev(rev):
-                logger.debug("erepo: git pull %s", url)
+                if shallow:
+                    logger.debug("erepo: unshallowing clone for '%s'", url)
+                    _unshallow(git)
+                    shallow = False
+                    CLONES[url] = (clone_path, shallow)
+                logger.debug("erepo: git pull '%s'", url)
                 git.pull()
```
Suor, line 306: Do we need `git.pull` when we already called `.fetch()` via `_unshallow()`?

Suor:
Also, as I see if we clone the same url more than once then we use full clone. Is this intended? Should we explain our justification in a comment?

dvc/external_repo.py
pmrowla: We may need pull fast-forward behavior in certain situations, but this has been updated to just use a single `pull(unshallow=True)` call in `_unshallow()` instead of `fetch(unshallow=True)` followed by the extra `pull()`

pmrowla:
> Also, as I see if we clone the same url more than once then we use full clone. Is this intended? Should we explain our justification in a comment?

This is commented now before we call `_unshallow()`

PR #4247, opened 2020-07-20
get/list/import/api: subrepo support

skshetry:

This PR at this time only works with ExternalRepo (should `Repo()` needs subrepo support, or should that be via `RepoTree`? probably both).

### TODO:
- [x] More tests
- [ ] `subrepo` support inside `Repo`
- [x] What should be done for `dvcx`? Currently, it seems it's broken, well before this PR. And, `dvcx` depends on git repo to have a top-level `dvc` repo, which this PR does not work with at all.

### Possible future works
- [ ] Implement `RepoTree.stat()`?
- [ ] Granular configs for dvctree and dvcignore
- [ ] Subrepo support in DvcRepo?

### DIFF
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -1,14 +1,11 @@
 import logging
 import os
-from contextlib import contextmanager
 from functools import wraps
 
 from funcy import cached_property, cat, first
 
 from dvc.config import Config
 from dvc.dvcfile import PIPELINE_FILE, Dvcfile, is_valid_filename
-from dvc.exceptions import FileMissingError
-from dvc.exceptions import IsADirectoryError as DvcIsADirectoryError
 from dvc.exceptions import (
     NoOutputOrStageError,
     NotDvcRepoError,
@@ -153,22 +150,29 @@ class Repo:
     def __repr__(self):
         return f"{self.__class__.__name__}: '{self.root_dir}'"
 
+    @cached_property
+    def repo_tree(self):
+        return RepoTree(self.tree, [self], stream=True)
+
     @classmethod
     def find_root(cls, root=None, tree=None):
+        # TODO (@skshetry): Verify and refactor this
         root_dir = os.path.realpath(root or os.curdir)
 
-        if tree:
-            if tree.isdir(os.path.join(root_dir, cls.DVC_DIR)):
-                return root_dir
-            raise NotDvcRepoError(f"'{root}' does not contain DVC directory")
-
-        if not os.path.isdir(root_dir):
-            raise NotDvcRepoError(f"directory '{root}' does not exist")
+        is_dir = tree.isdir if tree else os.path.isdir
 
         while True:
             dvc_dir = os.path.join(root_dir, cls.DVC_DIR)
-            if os.path.isdir(dvc_dir):
+            if is_dir(dvc_dir):
                 return root_dir
+            if (
+                tree
+                and os.path.dirname(os.path.abspath(tree.tree_root))
+                == root_dir
+            ):
+                raise NotDvcRepoError(
+                    f"'{root}' does not contain DVC directory"
+                )
             if os.path.ismount(root_dir):
                 break
             root_dir = os.path.dirname(root_dir)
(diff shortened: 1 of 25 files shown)


dvc/repo/__init__.py
```diff
@@ -153,22 +150,29 @@ def tree(self, tree):
     def __repr__(self):
         return f"{self.__class__.__name__}: '{self.root_dir}'"
 
+    @cached_property
+    def repo_tree(self):
+        return RepoTree(self.tree, [self], stream=True)
```
skshetry, line 155: We need granular configs in each of the tree operations. I could get around with `stream_open()`  or just this ^ hack.
But, we need `fetch` and `stream` to be per-ops rather than per instance.

skshetry:
Tests are failing on Mac for some reasons, while it passes on all other OSes. :confused:

skshetry:
Was able to fix tests, but still some tests are failing in Mac due to some issues with Docker. cc @efiop

tests/func/test_get.py
```diff
@@ -252,3 +236,50 @@ def test_get_pipeline_tracked_outs(
...
+    sub_repos = [erepo / path for path in subrepo_paths]
+    filename = "output"
+    for repo in sub_repos:
+        make_subrepo(repo, erepo.scm, local_cloud.config)
+        repo.dvc_gen({filename: output}, commit="add subrepo")
+        repo.dvc.push()
+
+    for i, repo in enumerate(sub_repos):
+        Repo.get(
+            f"file:///{erepo}",
```
pared, line 281: Should we provide local URL with `file:///` prefix? It seems to me that it is passed straight to `external_repo` and we never actually call `_fix_upstream`, because `os.path.isdir` returns False.

tests/func/test_import.py
```diff
@@ -345,3 +348,45 @@ def test_local_import(tmp_dir, dvc, scm):
...
+    sub_repos = [erepo / path for path in subrepo_paths]
+    filename = "output"
+    for repo in sub_repos:
+        make_subrepo(repo, erepo.scm, local_cloud.config)
+        repo.dvc_gen({filename: output}, commit="add subrepo")
+        repo.dvc.push()
+
+    rev = erepo.scm.get_rev()
+    for i, repo in enumerate(sub_repos):
+        url = f"file:///{erepo}"
```
pared, line 385: Same as above

dvc/repo/tree.py
```diff
@@ -437,6 +481,27 @@ def copytree(self, top, dest):
...
+        for root, _, files in self.walk(top, **kwargs):
+            level = root.replace(top, "").count(os.sep)
+            indent = spacing * level
+            yield "{}{}/".format(indent, os.path.basename(root))
+            sub_indent = spacing * (level + 1)
+            length = len(files)
+            for i, f in enumerate(files):
+                yield "{}{}{}".format(
+                    sub_indent, tee if i + 1 != length else last, f
+                )
```
pared, line 507: +1, do you intend to keep it? Looks like debugging feature.

dvc/subrepos.py
```diff
@@ -0,0 +1,11 @@
+import os
+
+from funcy import collecting
+
+
+@collecting
+def find(tree, top=None):
+    top = top or tree.tree_root
+    for root, _, _ in tree.walk(top):
+        if tree.isdir(os.path.join(root, ".dvc")):
```
pared, line 10: Shouldn't we use `Repo.DVC_DIR`?

dvc/repo/tree.py
skshetry: yes, I'd like to keep this. :smile:

tests/func/test_get.py
skshetry: I think the intention was to avoid git's optimization, see `--local` option on [`git clone`](https://git-scm.com/docs/git-clone#_options). And, `local_cloud.config` is a temp dir, meaning it's an abspath, so it's fine.

But, by doing this, we might miss local git repos, so, I'll rollback (or, try using mix of both in tests).

dvc/repo/tree.py
```diff
@@ -245,78 +248,117 @@ class RepoTree(BaseTree):  # pylint:disable=abstract-method
...
 
-    def __init__(self, repo, **kwargs):
-        super().__init__(repo, {"url": repo.root_dir})
-        if hasattr(repo, "dvc_dir"):
-            self.dvctree = DvcTree(repo, **kwargs)
-        else:
-            # git-only erepo's do not need dvctree
-            self.dvctree = None
+    def __init__(
+        self, tree, subrepos=None, **kwargs
```
pmrowla, line 252: It looks a bit strange to pass a list of subrepos into RepoTree like this. It seems like a RepoTree should be able to find nested subrepos by walking itself.

PR #4256, opened 2020-07-22
tree,remote: add support for WebDAV

### LINKED ISSUE #1153: WebDav: new remote type

iksnagreb:

Should fix #1153
Took WebDAVURLInfo from #3647
Relates to https://github.com/iterative/dvc.org/pull/1617

### TODO
- [x] Tests (at least some unit tests to check whether a valid webdavclient3 can be constructed from config should be doable)
- [x] Provide documentation (after discussing the added configuration options for WebDAV)

### DIFF
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -120,10 +120,14 @@ HTTP_COMMON = {
     "ask_password": Bool,
 }
 WEBDAV_COMMON = {
-    "root": str,
     "user": str,
     "password": str,
     "ask_password": Bool,
+    "root": str,
+    "token": str,
+    "cert_path": str,
+    "key_path": str,
+    "timeout": Coerce(int),
 }
 
 SCHEMA = {
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -123,7 +123,6 @@ WEBDAV_COMMON = {
     "user": str,
     "password": str,
     "ask_password": Bool,
-    "root": str,
     "token": str,
     "cert_path": str,
     "key_path": str,
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -119,6 +119,13 @@ HTTP_COMMON = {
     "password": str,
     "ask_password": Bool,
 }
+WEBDAV_COMMON = {
+    "root": str,
+    "user": str,
+    "password": str,
+    "ask_password": Bool,
+}
+
 SCHEMA = {
     "core": {
         "remote": Lower,
@@ -199,6 +206,8 @@ SCHEMA = {
                 },
                 "http": {**HTTP_COMMON, **REMOTE_COMMON},
                 "https": {**HTTP_COMMON, **REMOTE_COMMON},
+                "webdav": {**WEBDAV_COMMON, **REMOTE_COMMON},
+                "webdavs": {**WEBDAV_COMMON, **REMOTE_COMMON},
                 "remote": {str: object},  # Any of the above options are valid
             }
         )
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -299,6 +299,11 @@ class HTTPError(DvcException):
         super().__init__(f"'{code} {reason}'")
 
 
+class WebdavConfigError(DvcException):
+    def __init__(self, host):
+        super().__init__(f"Configuration for WebDAV {host} is invalid.")
+
+
 class PathMissingError(DvcException):
     default_msg = (
         "The path '{}' does not exist in the target repository '{}'"
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -304,6 +304,11 @@ class WebDAVConfigError(DvcException):
         super().__init__(f"Configuration for WebDAV {host} is invalid.")
 
 
+class WebDAVConnectionError(DvcException):
+    def __init__(self, host):
+        super().__init__(f"Unable to connect to WebDAV {host}.")
+
+
 class PathMissingError(DvcException):
     default_msg = (
         "The path '{}' does not exist in the target repository '{}'"
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -299,7 +299,7 @@ class HTTPError(DvcException):
         super().__init__(f"'{code} {reason}'")
 
 
-class WebdavConfigError(DvcException):
+class WebDAVConfigError(DvcException):
     def __init__(self, host):
         super().__init__(f"Configuration for WebDAV {host} is invalid.")
 
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -299,16 +299,6 @@ class HTTPError(DvcException):
         super().__init__(f"'{code} {reason}'")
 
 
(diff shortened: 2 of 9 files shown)


setup.py
```diff
@@ -92,12 +92,13 @@ def run(self):
 oss = ["oss2==2.6.1"]
 ssh = ["paramiko>=2.5.0"]
 hdfs = ["pyarrow>=0.17.0"]
+webdav = ["webdavclient3==3.14.5"]
```
shcheklein, line 95: should we do >=?

iksnagreb: Yes, should be safe, the webdav3 API seems to be rather fixed. Just used what I got out of pip freeze...

dvc/tree/webdav.py
```diff
@@ -0,0 +1,190 @@
...
+
+class WebDAVTree(BaseTree):  # pylint:disable=abstract-method
+    # Use webdav scheme
+    scheme = Schemes.WEBDAV
+
+    # URLInfo for Webdav ~ replaces webdav -> http
+    PATH_CLS = WebDAVURLInfo
+
+    # Non traversable as walk_files is not implemented
+    CAN_TRAVERSE = False
```
skshetry, line 24: It'd be great if you could implement `walk_files`, no pressure though.

dvc/config.py
```diff
@@ -119,6 +119,17 @@ class RelPath(str):
     "password": str,
     "ask_password": Bool,
 }
+WEBDAV_COMMON = {
+    "user": str,
+    "password": str,
+    "ask_password": Bool,
+    "root": str,
```
skshetry, line 126: You probably don't need `root` here, as that `url` from `REMOTE_COMMON should work, no?

iksnagreb: Yes, probably it is not necessary any more, I used it to specify the location of the WebDAV API endpoint (e.g. `/public.php/webdav`). But I still have a bit of confusion there...

Since I now terminate the `makedirs` at the `path` part of the initial `path_info` (which comes from the `url` of `REMOTE_COMMON`), the `root` option might be removed to not confuse others as well.

I somehow still wonder if there could be a valid use case for the option... At least there must be a reason for `webdavclient3` to offer this option?

skshetry: Cannot we split that into `hostname` and `root` from the `url`:
Eg:  `http://owncloud.com/remote.php/webdav` -> `http://owncloud.com` and `/remote.php/webdav`?

dvc/tree/webdav.py
```diff
@@ -0,0 +1,190 @@
...
+    def _client(self):
+        # Import the webdav client library
+        from webdav3.client import Client
+
+        # Construct hostname from path_info
+        hostname = (
+            self.path_info.scheme.replace("webdav", "http")
+            + "://"
+            + self.path_info.host
+        )
```
skshetry, line 87: `port` is not passed, try something like following, perhaps?
```python
http_info = HTTPURLInfo(self.path_info.url)
hostname = http_info.replace(path="").url
```

iksnagreb: Yes, you are right, I somehow forgot that there is more in the `path_info` than `scheme` and `host`... Should be fixed now.

dvc/exceptions.py
```diff
@@ -299,6 +299,16 @@ def __init__(self, code, reason):
...
 
+class WebDAVConfigError(DvcException):
+    def __init__(self, host):
+        super().__init__(f"Configuration for WebDAV {host} is invalid.")
+
+
+class WebDAVConnectionError(DvcException):
+    def __init__(self, host):
+        super().__init__(f"Unable to connect to WebDAV {host}.")
+
```
efiop, line 310: These are only used in webdav.py, so let's move them there. Also i'm not sure we need a special WebDAVConfigError exception. We could just raise ConfigError or at least inherit from it.

dvc/tree/webdav.py
```diff
@@ -0,0 +1,219 @@
...
+            raise WebDAVConfigError(hostname)
+
+        # Check whether connection is valid (root should always exist)
+        if not client.check(self.path_info.path):
+            raise WebDAVConnectionError(hostname)
+
+        # Return constructed client (cached)
+        return client
+
+    # Checks whether file exists
```
efiop, line 121: Some of these comments are not useful, they are declaring obvious things.

dvc/tree/webdav.py
```diff
@@ -0,0 +1,219 @@
...
+    # Moves file/directory at remote
+    def move(self, from_info, to_info, mode=None):
+        # Webdav client move
+        self._client.move(from_info.path, to_info.path)
+
+    # Copies file/directory at remote
+    def copy(self, from_info, to_info):
+        # Webdav client copy
+        self._client.copy(from_info.path, to_info.path)
+
```
efiop, line 203: I don't think this method is needed.

dvc/tree/webdav.py
```diff
@@ -0,0 +1,219 @@
...
+    # Downloads file from remote to file
+    def _download(self, from_info, to_file, name=None, no_progress_bar=False):
+        # pylint: disable=unused-argument
+
+        # Webdav client download
+        self._client.download(from_info.path, to_file)
```
efiop, line 209: Shame this library doesn't support progress bars for download/upload, the ui will not be very helpful :(

We need to at least put a dummy progress bar for this.

dvc/tree/webdav.py
```diff
@@ -0,0 +1,219 @@
...
+
+    # Checks whether file exists
+    def exists(self, path_info):
+        # Use webdav check to test for file existence
+        return self._client.check(path_info.path)
+
+    # Gets file hash 'etag'
+    def get_file_hash(self, path_info):
+        # Use webdav client info method to get etag
+        etag = self._client.info(path_info.path)["etag"].strip('"')
```
efiop, line 129: Is etag or something like that part of the webdav standard?

iksnagreb: As WebDAV is an extension of HTTP, it should be part of the standard.

dvc/tree/webdav.py
iksnagreb: Yes, it is probably not needed. I have implemented it, because it comes almost for free as it directly matches the `webdavclient3` method (the same goes for move/remove). Shall I remove it?

efiop: Yeah, let's remove it to not leave dead code around. If we ever need it - it will be trivial to introduce.

skshetry:
We can also use [`wsgidav`](https://github.com/mar10/wsgidav) which has a [good documentation on how to use it as a library](https://wsgidav.readthedocs.io/en/latest/user_guide_lib.html). It requires a wsgi server (eg: gunicorn, etc), but probably we can get away with just [`wsgiref`](https://docs.python.org/3/library/wsgiref.html) from the Python's standard library.

PR #4258, opened 2020-07-22
add toml support for ParamsDependency

dtrifiro:
We've been testing DVC with a few of our projects and for some of those we use a toml config to store parameters.
This PR adds support for parameters file in toml format. Example:

```
dvc run -n train                      \
-d data/input.csv                   \
-o model/model.h5                \
-M model/train_history.csv   \
-p params.toml:model.training \
python train.py
```

Where `params.toml` looks something like this:

> [transformers.numerical_quantile_transformer]
> n_quantiles = 300
> output_distribution = "uniform"
> 
> [model.training]
> epochs = 50
> verbose = 1

### DIFF
--- a/dvc/dependency/param.py
+++ b/dvc/dependency/param.py
@@ -2,6 +2,7 @@ import os
 from collections import defaultdict
 
 import dpath.util
+import toml
 import yaml
 from voluptuous import Any
 
@@ -21,6 +22,8 @@ class ParamsDependency(LocalDependency):
     PARAM_PARAMS = "params"
     PARAM_SCHEMA = {PARAM_PARAMS: Any(dict, list, None)}
     DEFAULT_PARAMS_FILE = "params.yaml"
+    PARAMS_FILE_LOADERS = defaultdict(lambda: yaml.safe_load)
+    PARAMS_FILE_LOADERS.update({".toml": toml.load})
 
     def __init__(self, stage, path, params):
         info = {}
@@ -87,8 +90,10 @@ class ParamsDependency(LocalDependency):
 
         with self.repo.tree.open(self.path_info, "r") as fobj:
             try:
-                config = yaml.safe_load(fobj)
-            except yaml.YAMLError as exc:
+                config = self.PARAMS_FILE_LOADERS[
+                    self.path_info.suffix.lower()
+                ](fobj)
+            except (yaml.YAMLError, toml.TomlDecodeError) as exc:
                 raise BadParamFileError(
                     f"Unable to read parameters from '{self}'"
                 ) from exc
--- a/dvc/repo/params/show.py
+++ b/dvc/repo/params/show.py
@@ -1,5 +1,6 @@
 import logging
 
+import toml
 import yaml
 
 from dvc.dependency.param import ParamsDependency
@@ -34,8 +35,10 @@ def _read_params(repo, configs, rev):
 
         with repo.tree.open(config, "r") as fobj:
             try:
-                res[str(config)] = yaml.safe_load(fobj)
-            except yaml.YAMLError:
+                res[str(config)] = ParamsDependency.PARAMS_FILE_LOADERS[
+                    config.suffix.lower()
+                ](fobj)
+            except (yaml.YAMLError, toml.TomlDecodeError):
                 logger.debug(
                     "failed to read '%s' on '%s'", config, rev, exc_info=True
                 )
--- a/tests/unit/dependency/test_params.py
+++ b/tests/unit/dependency/test_params.py
@@ -1,4 +1,5 @@
 import pytest
+import toml
 import yaml
 
 from dvc.dependency import ParamsDependency, loadd_from, loads_params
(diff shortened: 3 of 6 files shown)


dvc/dependency/param.py
```diff
@@ -87,8 +88,15 @@ def read_params(self):
 
         with self.repo.tree.open(self.path_info, "r") as fobj:
             try:
-                config = yaml.safe_load(fobj)
-            except yaml.YAMLError as exc:
+                if self.path_info.suffix == ".yaml":
+                    config = yaml.safe_load(fobj)
```
skshetry, line 92: We use `yaml` to load json as well.

skshetry: Could just try checking `toml` suffix, and then fallback to `yaml.safe_load`.

dtrifiro: I just force-pushed a check for suffix (use `yaml.safe_load()` for json and yaml)

skshetry: Still this misses the `yml`. So, it's better to just:
```
if suffix == "toml":
    # toml load
else:
   # yaml load
```

tests/unit/dependency/test_params.py
```diff
@@ -52,7 +53,7 @@ def test_params_error(dvc, params):
         loads_params(Stage(dvc), params)
 
 
-def test_loadd_from(dvc):
+def test_load_from(dvc):
```
skshetry, line 56: Please don't change the name as this is testing `loadd()` method.

dvc/dependency/param.py
pmrowla: Is it actually safe for us to guess the file format from the suffix here? It seems like we'll potential to run into issues where users have params filenames with unexpected capitalization on case-sensitive filesystems, or where they are using `.yml` instead of `.yaml`.

It seems to me we should maybe just try to load as yaml (and json) first, then try toml if that fails, then error out?

dtrifiro: Agreed @pmrowla

skshetry: @pmrowla, if that's a concern, we can just do case-insensitive suffix checks. By doing the above, we'll be taxing users using TOML. Right now, it's just inside a params dependency, but for `dvc params diff/show`, it might get slow.

skshetry: Btw @dtrifiro, this probably won't make `dvc params diff/show` work. That'll require a change on:
https://github.com/iterative/dvc/blob/a4b3c6d872c06ef1259bb15bd0d69eaa4ccae7ff/dvc/repo/params/show.py#L29

dtrifiro: I added a loader map which is based on the lowercase file suffix, which defaults to `yaml.safe_load` for unknown suffixes. I also update `_read_params`

tests/unit/dependency/test_params.py
dtrifiro: Oops, didn't do this on purpose, I thought I added an extra d by mistake and did not check the diff before pushing.

dvc/repo/params/show.py
```diff
@@ -34,8 +35,10 @@ def _read_params(repo, configs, rev):
 
         with repo.tree.open(config, "r") as fobj:
             try:
-                res[str(config)] = yaml.safe_load(fobj)
-            except yaml.YAMLError:
+                res[str(config)] = ParamsDependency.PARAMS_FILE_LOADERS[
+                    config.suffix.lower()
```
pared, line 39: +1

shcheklein:
Please, don't forget to update the docs after this is merged.

dvc/dependency/param.py
efiop: I also agree with @pmrowla , it is better to just try to load and catch yaml formatting error than trying to judge by the extension.

dtrifiro: In the last force-push I ended up using another approach (default to `yaml.safe_load`, use `toml` if  lowercase suffix matches) because I felt a nested try/catch a bit awkward, especially if there are other loaders different from yaml/json and toml (not that I think there's much more, these 3 alone probably cover most of the use cases).

What is your preference?

Also, I saw that the Travis build fails on Windows and Mac (probably due to misconfiguration?) is there anything I need to do about that? Thanks

skshetry: @efiop, In case of unknown file format, sure, it could fallback to yaml; but, if the file is clearly a `toml` file, it should just try to load as toml and fail. 

This is what @dtrifiro is doing.

efiop: Makes sense, let's keep it. :+1:

efiop:
@dtrifiro Could you please submit a docs PR for  https://dvc.org/doc/command-reference/run ?

dtrifiro:
> @dtrifiro Could you please submit a docs PR for https://dvc.org/doc/command-reference/run ?


https://github.com/iterative/dvc.org/pull/1612

efiop:
@dtrifiro Thank you so much! :pray:

Issue #4233, opened 2020-07-17
repro: update status message for dvc-added files


sarthakforwet:
## Bug Report

As we know that running `dvc repro` command would reproduce the stages defined in `dvc.yaml` file. One thing we also know that `.dvc` file can be used as the beginning of a pipeline. If we consider a sample case where we have the follwing pipeline and we perform `dvc repro`. 

![image](https://user-images.githubusercontent.com/46163196/87825476-d7d50b80-c894-11ea-89c8-016daa1b2e6e.png)

As we can see the command is successful in its execution but the thing to note that it's considering `.dvc` files as a stage (which they are probably not) as can be seen from the first line of the output. 

Does the output corresponding to `.dvc` file needs some updations? 
**Output of `dvc version`:**

```console
$ dvc version
DVC version: 1.1.7
Python version: 3.8.3
Platform: Windows-10-10.0.19041-SP0
Binary: False
Package: pip
Supported remotes: http, https
Cache: reflink - not supported, hardlink - supported, symlink - supported
Repo: dvc, git
```

**Additional Information (if any):**

This issue is outcome of the discussion at [#1572(review)](https://github.com/iterative/dvc.org/pull/1572#discussion_r456230017)

efiop:
Hi @sarthakforwet ! Great points! The dag part is discussed in https://github.com/iterative/dvc/issues/4058 , but we do indeed need to reconsider `repro` output as well. Thank you for the feedback!

sarthakforwet:
Thanks @efiop.

Suggestion: We can change the output so that whenever a `.dvc` extension is detected, it would print "Deps" instead of "Stage".

Also, Can I also work on this issue and what would be the prerequisite for it? Thanks

efiop:
> Suggestion: We can change the output so that whenever a .dvc extension is detected, it would print "Deps" instead of "Stage".

Good suggestion, but I think we are better off using something like
```
if not isinstance(stage, PipelineStage) and stage.is_data_source:
    # new message
```
:slightly_smiling_face: Especially since `.dvc` might be a legacy pipeline file.

> Also, Can I also work on this issue and what would be the prerequisite for it? Thanks

Sure! Please feel free. Looks like you'll simply need to adjust the message in https://github.com/iterative/dvc/blob/b77ce02f6b7f66efa9fc73b7146f1c4bf4fa6da3/dvc/stage/__init__.py#L309 . Let us know if you'll have any questions. Thank you for looking into it! :pray:

[...]

PR #4269, opened 2020-07-23
repro: update status message for dvc-added files

sarthakforwet:
The reason for updating the status message is that `.dvc` files are not stages and hence we can just remove the prefix "Stage" from the output whenever a `.dvc` file would be detected.

References #4233

### DIFF
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -306,7 +306,10 @@ class Stage(params.StageParams):
     @rwlocked(read=["deps"], write=["outs"])
     def reproduce(self, interactive=False, **kwargs):
         if not (kwargs.get("force", False) or self.changed()):
-            logger.info("Stage '%s' didn't change, skipping", self.addressing)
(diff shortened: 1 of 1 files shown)


dvc/stage/__init__.py
```diff
@@ -306,7 +306,10 @@ def remove(self, force=False, remove_outs=True, purge=True):
     @rwlocked(read=["deps"], write=["outs"])
     def reproduce(self, interactive=False, **kwargs):
         if not (kwargs.get("force", False) or self.changed()):
-            logger.info("Stage '%s' didn't change, skipping", self.addressing)
+            if '.dvc' in self.adressing:
```
efiop, line 309: As I've mentioned in https://github.com/iterative/dvc/issues/4233#issuecomment-660721316 , using `.dvc` suffix is not the best approach:
```suggestion
            if not isinstance(stage, PipelineStage) and stage.is_data_source:
```

PR #1673, opened 2019-03-01
add: cleanup .gitignore decorator

pared:
Fixes #1641

PR #4282, opened 2020-07-25
`dvc check-ignore` command

### LINKED ISSUE #3736: Introduce check-ignore for dvc ignore command.
There should be functionality providing easy way to check whether some file is `dvc-ignored` or not. Related: #3538

karajan1001:
update some tests first
fix #3736

  iterative/dvc.org/pull/1629


dvc/ignore.py
```diff
@@ -217,3 +268,26 @@ def _outside_repo(self, path):
         ):
             return True
         return False
+
+    def check_ignore(self, targets):
+        check_results = []
+        for target in targets:
+            full_target = os.path.join(os.getcwd(), target)
```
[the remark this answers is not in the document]
efiop:
```suggestion
            full_target = os.path.abspath(target)
```

dvc/ignore.py
```diff
@@ -56,7 +75,7 @@ def __call__(self, root, dirs, files):
 
         return dirs, files
 
-    def matches(self, dirname, basename, is_dir=False):
+    def _get_normalize_path(self, dirname, basename):
```
pared, line 78: NOTE: line 88 `return False`. That does not match `_get..`

dvc/ignore.py
```diff
@@ -85,20 +108,46 @@ def ignore(self, path, is_dir):
...
+        path = self._get_normalize_path(dirname, basename)
+        return self._ignore_details(path, is_dir)
+
+    def _ignore_details(self, path, is_dir):
+        result = []
+        for ignore, pattern in zip(self.regex_pattern_list, self.pattern_list):
+            regex = re.compile(ignore[0])
+            # skip system pattern
+            if not pattern.file_info:
+                continue
```
pared, line 121: Won't that result in no matches for `.git` and return 1 when we do `check-ignore .git`?
I guess in one way this is right behavior, since its not specified in any file, but, its also against `tree.walk` behavior which will ignore `.git`. 
But I guess main point of `check-ignore` is to debug `.dvcignore`s, so its fine.

tests/func/test_check_ignore.py
```diff
@@ -0,0 +1,105 @@
...
+
+    assert main(["check-ignore", file]) == ret
+    assert output in caplog.text
+
+
+@pytest.mark.parametrize(
+    "file,ret,output",
+    [
+        ("file", 0, "{}:1:f*\tfile\n".format(DvcIgnore.DVCIGNORE_FILE)),
+        ("foo", 0, "{}:2:!foo\tfoo\n".format(DvcIgnore.DVCIGNORE_FILE)),
```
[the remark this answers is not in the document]
karajan1001: This is what I would like to discuss.

Yes In Git, 0 should be returned when a given path is ignored. But 0 returns on any pattern match the file either included or excluded, 1 returns when none of the pattern match. (It might be a bug or a feature not clearly documented. I sent an email to Git last time but no reply)
But I felt illogical on this results. 0 for ignored and 1 for included or not matched might be more reasonable. 

As a result. I left `matches` in
```
CheckIgnoreResult = namedtuple(
    "CheckIgnoreResult", ["file", "matches", "patterns"]
)
``` 
which could be used for marking whether it is ignored or not.

dvc/ignore.py
karajan1001: I just followed Git.
image
Maybe I can return 
```
$ dvc check-ignore .git
::System pattern   .git
$ echo $?
0
```

tests/func/test_check_ignore.py
```diff
@@ -0,0 +1,105 @@
...
+@pytest.mark.parametrize("path,ret", [({"dir": {}}, 0), ({"dir": "files"}, 1)])
+def test_check_ignore_dir(tmp_dir, dvc, path, ret):
+    tmp_dir.gen(DvcIgnore.DVCIGNORE_FILE, "dir/")
+    tmp_dir.gen(path)
+
+    assert main(["check-ignore", "-q", "dir"]) == ret
+
+
+def test_check_ignore_default_dir(tmp_dir, dvc):
+    tmp_dir.gen(DvcIgnore.DVCIGNORE_FILE, "dir/")
```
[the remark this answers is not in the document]
karajan1001: It could be removed

tests/func/test_check_ignore.py
```diff
@@ -0,0 +1,105 @@
...
+    assert main(["check-ignore", "-q", ".dvc"]) == 1
+
+
+def test_check_ignore_sub_repo(tmp_dir, dvc):
+    tmp_dir.gen(
+        {DvcIgnore.DVCIGNORE_FILE: "other", "dir": {".dvc": {}, "foo": "bar"}}
+    )
+
+    assert main(["check-ignore", "-q", os.path.join("dir", "foo")]) == 1
```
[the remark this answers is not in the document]
karajan1001: Throw exceptions at sub repo? A good suggestion,`OutOfRepoException` and could also be used at `dvc check-ignore ../`

dvc/ignore.py
karajan1001: Any `OutofRepoException` use case.

tests/func/test_check_ignore.py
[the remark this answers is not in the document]
pared: @karajan1001 
To me, it seems that either docs is wrong (because it says that 0 is returned if file is ignored) or implementation (which returns 0 upon any matching spec). I'll try to reach git-devel IRC, maybe we can get an answer faster. 
For now, we can leave it as is.

tests/func/test_check_ignore.py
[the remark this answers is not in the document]
pared: @karajan1001 
We can also create issue for that, after merge. There is big upcoming change with subrepos https://github.com/iterative/dvc/pull/4247
I guess the subrepo detection might overlap with this one.

dvc/ignore.py
pared: If that's not too much trouble, sure. But if its, feel free to keep it that way.

tests/func/test_check_ignore.py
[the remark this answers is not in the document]
karajan1001: @pared 
They including inside sub-rep, in `.dvc`, `.hg`,`.git`, and out of root dir are all `out of working space` I think.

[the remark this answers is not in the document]
karajan1001: @pared 
Inside sub-rep, in `.dvc`, `.hg`,`.git`, and out of root dir are all `out of working space` I think.

dvc/ignore.py
karajan1001: image
I think this is the same problem with the above one.
If 0 or 1 depends on whether the file is excluded, then we should return 0 here, but If it depends on whether any pattern in .dvcignore matches the file, then we should return 1 here.

tests/func/test_check_ignore.py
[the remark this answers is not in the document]
pared: @karajan1001 I agree

dvc/ignore.py
pared: @karajan1001 you mean the proper git return code?
To avoid unnecessary change requests, let's leave it as is.

karajan1001:
@pared 
Something to be noticed, replacing

```
if (conditions): 
    return False
```

with 

```
try: 
    ... 
except exceptions:
    return False
```

runs faster when `conditions` happens rarely, but it runs much slower when `conditions` is a common one. 
The exception handling cost lots of resources.

pared:
@karajan1001 
I think the most common out of workspace use case is with external dependencies. From the development point of view, I think we should stick to conditions because they are much more readable and don't require us to leave a long comment why are we doing things in a particular way.

karajan1001:
> @karajan1001
> I think the most common out of workspace use case is with external dependencies. From the development point of view, I think we should stick to conditions because they are much more readable and don't require us to leave a long comment why are we doing things in a particular way.

@pared 
Thank you, if-else is more readable and try-catch separates exception handling and function logic (All Exceptions been handled at one place and one level). The best practice might be sharing the same standard all over the project for all contributors. Because it makes us more ease to understand each other's code.
@efiop I'm not sure if this project has a universal rule or code style?

dvc/command/check_ignore.py
```diff
@@ -0,0 +1,71 @@
...
+        "--non-matching",
+        action="store_true",
+        default=False,
+        help="Show given paths which don’t match any pattern. "
+        "Only used when --details is enabled.",
+    )
+    parser.add_argument(
+        "targets",
+        nargs="+",
+        help="Input files/directories to check " "ignore patterns.",
```
jorgeorpinel, line 69: From https://github.com/iterative/dvc.org/pull/1629#pullrequestreview-457170973:

- Why "inputs"?
- "files/directories" -> "paths of files or directories"
- Should mention that Dvcignore patterns are also accepted as targets.

dvc/command/check_ignore.py
```diff
@@ -0,0 +1,71 @@
...
+        description=append_doc_link(ADD_HELP, "check-ignore"),
+        help=ADD_HELP,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+    parser.add_argument(
+        "-d",
+        "--details",
+        action="store_true",
+        default=False,
+        help="Show the exclude pattern together with the path.",
```
jorgeorpinel, line 56:
```suggestion
        help="Show the exclude pattern together with each target path.",
```
Per iterative/dvc.org/pull/1629

dvc/command/check_ignore.py
```diff
@@ -0,0 +1,71 @@
...
+    )
+    parser.add_argument(
+        "-n",
+        "--non-matching",
+        action="store_true",
+        default=False,
+        help="Show given paths which don’t match any pattern. "
+        "Only used when --details is enabled.",
```
jorgeorpinel, line 64:
```suggestion
        help="Show the target paths which don’t match any pattern. "
        "Only usable when `--details` is also employed",
```
Per iterative/dvc.org/pull/1629

jorgeorpinel: Could also be "Only has an effect if `--details`... (please update docs accordingly if you go this route).

jorgeorpinel: P/s why not when you use this option, --details is also enabled automaticlaly?

dvc/command/check_ignore.py
jorgeorpinel: From iterative/dvc.org/pull/1629:

> All fields in each line, except for `<target path>`, will be empty.

So why even use the special --details format? Maybe just output the target paths normally with this option? (Does it even require --details?)

jorgeorpinel:
p.s. being able to review docs and core at the same time has some great benefits, especially when the draft doc is so thorough and with comprehensive examples so kudos and thanks again @karajan1001 !

efiop:
@karajan1001 Sorry for the delay :)

> I'm not sure if this project has a universal rule or code style? Or at what condition should we raise an exception and what situation to use if-condition

There is no strict policy. As you've noticed we mix and match these at different places in the code. For example, stuff like `if self.tree.exists` is usually used because we didn't finalize tree unification yet, where we need to unify the exceptions that all tree types raise, so that we could handle them in a general way. And sometimes in places where performance doesn't matter much, we just use `if`s because they seem more readable in that particular place. So go with your gut in this case, we'll help out if we see that something could be improved. :slightly_smiling_face: 

As always, thank you for contributing amazing stuff! :pray:

dvc/command/check_ignore.py
```diff
@@ -0,0 +1,71 @@
...
+
+class CmdCheckIgnore(CmdBase):
+    def __init__(self, args):
+        super().__init__(args)
+        self.ignore_filter = self.repo.tree.dvcignore
+
+    def _show_results(self, results):
+        for result in results:
+            if result.matches or self.args.non_matching:
+                #
```
efiop, line 19:
```suggestion
```

dvc/main.py
```diff
@@ -93,7 +93,7 @@ def main(argv=None):
         # so won't be reused by any other subsequent run anyway.
         clean_repos()
 
-    if ret != 0:
+    if ret != 0 and ret != 1:
```
efiop, line 96: This will have consequences outside of `dvc check-ignore`. But on the other hand, we've heard people complain about `"Having any problems?"` message being too annoying on normal failures.

But actually we don't really need this, if `-q` works as `quiet`(logger.info should be silenced). But looking into it, I see that it is indeed broken right now, as we restore the loglevel a few lines above. This is out of scope for this PR, so let me adjust that on master so you could rebase.

karajan1001: Yes, I ought to submit an issue for it. We have only one return number that not showing "Having any problems?".

efiop: @karajan1001 Merged https://github.com/iterative/dvc/pull/4304 , please rebase on top of master, I it should be fixed.

karajan1001:
@pared @efiop 
According to discussions from in [Discord](https://discordapp.com/channels/485586884165107732/565699007037571084/738242622019076107)

Reverted to the previous version, but several problems still remained and need to create issues for them.

1.  Return 0 or 1 for files matching a `!` started pattern.
2. Result insistency of `__call__` ,`is_ignored_dir`, and `is_ignored_file` with paths outside the repo.

dvc/main.py
```diff
@@ -84,7 +84,7 @@ def main(argv=None):  # noqa: C901
         ret = 255
 
     try:
-        if ret != 0:
+        if ret != 0 and ret != 1:
```
efiop, line 87: I think this is no longer needed after the rebase
```suggestion
        if ret != 0:
```

karajan1001: @efiop 
Still got this warning after I deleted `ret != 1`, is this OK?
image

dvc/command/check_ignore.py
```diff
@@ -0,0 +1,70 @@
...
+        action="store_true",
+        default=False,
+        help="Show the target paths which don’t match any pattern. "
+        "Only usable when `--details` is also employed",
+    )
+    parser.add_argument(
+        "targets",
+        nargs="+",
+        help="Exact or wildcard paths of files or directories to check "
+        "ignore patterns.",
```
jorgeorpinel, line 68: That sentence is a bit confusing. How about "File or directory paths, or ignore patterns to check"

dvc/command/check_ignore.py
```diff
@@ -0,0 +1,70 @@
...
+        for target in self.args.targets:
+            result = self.ignore_filter.check_ignore(target)
+            self._show_results(result)
+            if result.match:
+                ret = 0
+        return ret
+
+
+def add_parser(subparsers, parent_parser):
+    ADD_HELP = "Debug DVC ignore/exclude files"
```
jorgeorpinel, line 40: How about "Check whether files or directories are excluded due to `.dvcignore`."
> Per https://github.com/iterative/dvc.org/pull/1629/files

karajan1001:
@jorgeorpinel 
I'm sorry Seems that it had been merged, and I can't change it anymore. 
Maybe we could open another Issue (GoodFirstOne) for these problems?

jorgeorpinel:
No problem, I'll submit a PR when I get a chance.

dvc/command/check_ignore.py
karajan1001: I think the targets we checked here are wildcard patterns not ignore patterns. It's a pattern used in file searches so the full name need not be typed provided by the file system. For example, we have 
```
aa
ab
bb
```
in our current path. 
The command `dvc check-ignore a*` equals to `dvc check-ignore aa ab`.  Our program received a list of files instead of a single pattern `a*`. 

Only my personal opinion, not good at English, maybe `ignore patterns` can represent the same meaning?

dvc/command/check_ignore.py
karajan1001: I think it's OK.

dvc/command/check_ignore.py
jorgeorpinel: > the targets we checked here are wildcard patterns not ignore patterns

Good to know, thanks! Confusing probably... but makes sense. So those depend on the OS? I.e. POSIX shells have accept different wildcards than, say, Windows PowerShell.

And yeah no worries, it's getting too deep. I'll take care of rewriting it as needed in iterative/dvc.org/pull/1629 and submit a matching PR to the core repo.

jorgeorpinel: Went for "File or directory paths to check (wildcards supported)" in iterative/dvc.org/pull/1629.

dvc/command/check_ignore.py
jorgeorpinel: Wrong handle 😆

dvc/command/check_ignore.py
jorgeorpinel: Just "File or directory paths to check" now... See #1673

dvc/main.py
skshetry: @karajan1001, or just do something like this for now:
```python
if ret == 1 and args and args.cmd == "check-ignore":
    pass
```

Not ideal, but should work for now.

Issue #4281, opened 2020-07-24
LockFile class dump/load uses different libraries


dtrifiro:
## Bug Report
`LockFile.dump()` calls `dvc.utils.yaml.dump_yaml()`,  which uses ruamel (`ruamel.yaml.YAML.dump`), whereas `LockFile.load()` calls `dvc.utils.yaml.parse_yaml()` which uses pyyaml (`yaml.load` with `SafeLoader`).

This results in a very subtle bug: ruamel uses the YAML 1.2 specification by default, whereas pyyaml uses the YAML 1.1 specification (https://pypi.org/project/PyYAML/).

This means that when dumping parameters to the lockfile, the 1.2 specification is used, which writes numbers in exponential notation like this: `1e-6`. The YAML 1.1 specification, however, expects exponential notation numbers to include a dot: `1.0e-6`, if it is not present, `1e-6` is read as a *string* instead of being read as a float.

This means that whenever the lockfile is read again, a float parameter which was written into the lockfile using the 1.2 specification, is read as a string instead of being read as a float, this results in `dvc status` always marking the parameters file as modified. When launching `dvc params diff` however, both `params.yaml` and `dvc.lock` are read using `yaml.safe_load` which uses the 1.1 specification thus resulting in an empty diff, which is kinda confusing: dvc is in a dirty status but `dvc params diff` shows nothing.

A (partial?) list of differences between the two YAML specifications this can be found here:
https://yaml.readthedocs.io/en/latest/pyyaml.html?highlight=specification

A discussion about pyaml's parsing of floats can be found here:
https://github.com/yaml/pyyaml/issues/173
See this comment about pyaml being focused on YAML spec 1.1. https://github.com/yaml/pyyaml/pull/174#issuecomment-405868251

## Dirty solution
Thanks to @ariciputi. This hack forces the ruamel yaml parser to use the 1.1 specification, thus solving the issue.

```In [65]: adict
Out[65]: {'hello': [1e-06, 1e-05, 0.0001, 0.001, 0.01]}
In [66]: cc = StringIO()
In [67]: yaml_12 = ruamel.yaml.YAML()
In [68]: yaml_12.dump(adict, cc)
In [69]: cc.seek(0)
Out[69]: 0
In [70]: cc.getvalue()
Out[70]: 'hello:\n- 1e-06\n- 1e-05\n- 0.0001\n- 0.001\n- 0.01\n'
In [71]: yaml_11 = ruamel.yaml.YAML()
[...]

PR #4295, opened 2020-07-28
force YAML 1.1 syntax when dumping yaml files with ruamel

dtrifiro:
workaround for #4281


skshetry:
Tests are failing because this workaround adds "% YAML 1.1` which is a yaml version directive to the start of the YAML files.

We need to find some way to avoid this (but still dump in 1.1 version).

skshetry:
just did a simple benchmarking for ruamel. It's twice as slow as pyyaml for `example-get-started`'s dvc.yaml load.

dtrifiro:
Development moved to https://github.com/iterative/dvc/pull/4380

Issue #4162, opened 2020-07-03
Developping a workflow for simultaneous work


bjonnh:
The idea would be to provide guidance to users in order to optimize parallel work to avoid conflicts when pushing on the same repository.

There are two aspects to that issue:
- organizational: what would be the workflow that would reduce chances of getting conflicts
- technical: can we make tools that would ease such conflicts

Organizational tools:
- a document that provides an example problematic workflow and how it can be overcome

Technical tools:
- a tool that would allow to merge two versions of dir contents  (could it work like a git rebase?)

efiop:
Discord context: https://discordapp.com/channels/485586884165107732/485596304961962003/728651039657558066

Possible solution: git merge driver.

efiop:
Hi @bjonnh !

We've added a git merge-driver (released in dvc 1.6.0) that should help with append-only datasets and created a doc about merge conflicts in general https://dvc.org/doc/user-guide/merge-conflicts (the recipe for setting up our merge-driver is there as well). Please take a look and let us know what you think :slightly_smiling_face:

bjonnh:
This is really nice! Thanks for all your work we'll try that on our repositories see how that goes.

efiop:
Closing for now. Feel free to reopen if you'll have a chance to give it a try and will have any feedback. :slightly_smiling_face:

Issue #4144, opened 2020-07-01
 `get-url` and `import-url` doesn't seem to work with S3 buckets anymore.


anotherbugmaster:
## Bug Report

1. Create an empty directory
2. `dvc init --no-scm`
3. `dvc import-url s3://some_bucket/some_target -v`
```
2020-07-01 17:14:02,947 DEBUG: fetched: [(3,)]                                  
2020-07-01 17:14:03,123 DEBUG: Removing output 'some_target' of stage: 'some_target.dvc'.
Importing 's3://some_bucket/some_target' -> 'some_target'
2020-07-01 17:14:03,123 DEBUG: Computed stage: 'some_target.dvc' md5: '2f8b87d3b22efd1638f414c3b3f65614'
2020-07-01 17:14:03,123 DEBUG: 'md5' of stage: 'some_target.dvc' changed.
2020-07-01 17:14:04,088 DEBUG: fetched: [(0,)]
2020-07-01 17:14:04,146 ERROR: failed to import s3://some_bucket/some_target. You could also try downloading it manually, and adding it with `dvc add`. - Current operation was unsuccessful because 's3://some_bucket/some_target' requires existing cache on 's3' remote. See <https://man.dvc.org/config#cache> for information on how to set up remote cache.
------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/dvc/command/imp_url.py", line 14, in run
    self.repo.imp_url(
  File "/usr/lib/python3.8/site-packages/dvc/repo/__init__.py", line 36, in wrapper
    ret = f(repo, *args, **kwargs)
  File "/usr/lib/python3.8/site-packages/dvc/repo/scm_context.py", line 4, in run
    result = method(repo, *args, **kw)
  File "/usr/lib/python3.8/site-packages/dvc/repo/imp_url.py", line 54, in imp_url
    stage.run()
  File "/home/anotherbugmaster/.local/lib/python3.8/site-packages/funcy/decorators.py", line 39, in wrapper
    return deco(call, *dargs, **dkwargs)
  File "/usr/lib/python3.8/site-packages/dvc/stage/decorators.py", line 35, in rwlocked
    return call()
  File "/home/anotherbugmaster/.local/lib/python3.8/site-packages/funcy/decorators.py", line 60, in __call__
    return self._func(*self._args, **self._kwargs)
  File "/usr/lib/python3.8/site-packages/dvc/stage/__init__.py", line 424, in run
[...]

PR #4298, opened 2020-07-28
dvc: introduce merge-driver

efiop:
Related to #4162

TODO:
- [ ] more tests

### DIFF
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -5,7 +5,12 @@ from copy import copy
 from shortuuid import uuid
 
 import dvc.prompt as prompt
-from dvc.exceptions import CheckoutError, ConfirmRemoveError, DvcException
+from dvc.exceptions import (
+    CheckoutError,
+    ConfirmRemoveError,
+    DvcException,
+    MergeError,
+)
 from dvc.path_info import WindowsPathInfo
 from dvc.progress import Tqdm
 from dvc.remote.slow_link_detection import slow_link_guard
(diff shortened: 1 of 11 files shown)


dvc/dvcfile.py
```diff
@@ -239,6 +259,9 @@ def remove_stage(self, stage):
         else:
             super().remove()
 
+    def merge(self, ancestor, other):
+        raise NotImplementedError
+
```
efiop, line 264: These stubs are for the future, as merging dvcfiles is potentially a handy thing to have.

dvc/cache/base.py
```diff
@@ -550,3 +555,70 @@ def get_files_number(self, path_info, hash_, filter_info):
...
+        result = [
+            {
+                self.tree.PARAM_RELPATH: relpath,
+                self.tree.PARAM_CHECKSUM: checksum,
+            }
+            for relpath, checksum in merged.items()
+        ]
+
+        # Sorting the list by path to ensure reproducibility
+        return sorted(result, key=itemgetter(self.tree.PARAM_RELPATH))
```
efiop, line 605: We have a similar thing in _collect_dir in BaseTree's, but these will be unified as a part of https://github.com/iterative/dvc/issues/4144 , once dir hash computation is detached from cache.

pared:
@efiop 
What about conflicting files? I prepared sample script:
```
#!/bin/bash

rm -rf repo
mkdir repo

pushd repo

git init --quiet
dvc init --quiet

git add -A
git commit -am "init"

dvc install
echo "*.dvc merge=dvc" >> .gitattributes
git add -A
git commit -m "setup hooks"

git checkout -b one
mkdir data
echo foo >> data/foo

dvc add data
git add -A
git commit -m "add foo"

git checkout master

git checkout -b two
mkdir data
echo bar >> data/foo

dvc add data
git add -A
git commit -m "add foo but bar"

git checkout master

git merge one --no-edit
git merge two --no-edit
```
Where I have 2 branches:
`one` that has `dvc add`-ed `data/foo` with `foo` content
`two` that has `dvc add`-ed `data/foo` with `bar` content

and then, on master I merge `one` and then `two`. 
My `data.dvc` gets silently overridden so that `data/foo` contains `bar` - the `two` branch content.
Shouldn't we get some error?

efiop:
@pared 

>git checkout master
> git merge one --no-edit
> # I think we should get error now

Not really, because there is nothing in master, so there is no conflict.

> git merge two --no-edit

And here you get a conflict, as expected.

pared:
@efiop sorry, I meant that I would expect error after the comment.
And yes, there is conflict, but my `.dvc` file gets Auto-merged. It is some indicator of failure, but it is easy to omit. I would expect that git will not try to auto-merge. Choosing which file is the proper one should be users responsibility. 
Now the content of `data/foo` depends on the order of `merge`:
```
git merge one --no-edit
git merge two --no-edit
# cat data/foo == bar
```

```
git merge two --no-edit
git merge one --no-edit
# cat data/foo == foo
```
If we were doing same thing in git, we would get unresolved conflict

efiop:
@pared Ah, got it. Yes, that's a bug, thank you! :pray: Finally getting to creating additional tests for it, will be sure to add one for that as well.

pared (approved):

PR #4304, opened 2020-07-30
dvc: fix --quiet mode

efiop:
Part of the issue is a bug from #3806                            
                                                                 
Kudos to @karajan1001 for finding this bug while working on #4282

### DIFF
--- a/dvc/main.py
+++ b/dvc/main.py
@@ -21,7 +21,7 @@ from dvc.utils import format_link
 logger = logging.getLogger("dvc")
 
 
-def main(argv=None):
+def main(argv=None):  # noqa: C901
     """Run dvc CLI command.
 
     Args:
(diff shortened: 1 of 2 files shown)


dvc/main.py
```diff
@@ -92,11 +101,3 @@ def main(argv=None):
         # Remove cached repos in the end of the call, these are anonymous
         # so won't be reused by any other subsequent run anyway.
         clean_repos()
```
efiop, line 103: Ideally these should be put into atexit kinda like https://github.com/python/cpython/blob/e1945307d36849f8be8937144cf3dd6ebab6274c/Lib/multiprocessing/util.py#L362 . Since they are not only relevant for CLI but for API as well. This also might be the reason why we are getting some weird fini errors in tests.

dvc/main.py
```diff
@@ -37,16 +37,16 @@ def main(argv=None):
...
+        level = None
+        if args.quiet:
+            level = logging.CRITICAL
+        elif args.verbose:
+            trace = args.verbose > 1
+            level = logging.TRACE if trace else logging.DEBUG
+
+        if level is not None:
+            logger.setLevel(level)
+
```
efiop, line 49: Got a bit too fancy with the magic :slightly_smiling_face:

PR #4305, opened 2020-07-30
commit: make sure deps/outs exist

### LINKED ISSUE #4266: 'NoneType' object is not subscriptable
## Bug Report

I came up with this issue by other paths, but I tried to find an easy way to isolate it.

#### steps to reproduce

```sh
dvc remote add s3 s3://
dvc remote default s3
dvc run --no-exec -n train \
  --outs-persist model.data \
   ./train.sh

git add dvc.yaml
dvc commit
```
outputs 
```sh
ERROR: unexpected error - 'NoneType' object is not subscriptable 
```

**Output of `dvc version`:**

```console
DVC version: 1.1.11
Python version: 3.7.6
Binary: False
Package: pip
Repo: dvc, git
Filesystem type (workspace): ('apfs', '/dev/disk1s5')
```

efiop:
Fixes #4266

### DIFF
--- a/dvc/dependency/repo.py
+++ b/dvc/dependency/repo.py
@@ -67,7 +67,7 @@ class RepoDependency(LocalDependency):
                     return self.repo.cache.local.tree.get_hash(path, tree=tree)
                 return tree.get_file_hash(path)
 
-    def status(self):
+    def workspace_status(self):
         current_checksum = self._get_checksum(locked=True)
         updated_checksum = self._get_checksum(locked=False)
 
@@ -75,6 +75,9 @@ class RepoDependency(LocalDependency):
             return {str(self): "update available"}
 
         return {}
+
+    def status(self):
+        return self.workspace_status()
 
     def save(self):
         pass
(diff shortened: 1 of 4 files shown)


dvc/dependency/repo.py
```diff
@@ -67,7 +67,7 @@ def _get_checksum(self, locked=True):
                     return self.repo.cache.local.tree.get_hash(path, tree=tree)
                 return tree.get_file_hash(path)
 
-    def status(self):
+    def workspace_status(self):
         current_checksum = self._get_checksum(locked=True)
         updated_checksum = self._get_checksum(locked=False)
```
efiop, line 73: This status hack backfired a bit. Need to set self.tree properly here (it will pretty much be RepoTree for the external repo that this belongs to), so that things like `self.exists()` work automatically.

dvc/output/base.py
```diff
@@ -280,6 +283,7 @@ def save(self):
         self.info = self.save_info()
 
     def commit(self):
+        assert self.info
```
efiop, line 286: Added this because https://github.com/iterative/dvc/blob/2f48bb4089c25c2144fa3c9d9d8d0ba69c4c3e51/dvc/cache/base.py#L280 shouldn't be there, really. It is a leftover from erepo/tree hacks we used to have, but some code is still using it, so need to get back to it in a separate PR.

PR #4309, opened 2020-07-31
remote: allow local options for repo remotes

### LINKED ISSUE #4276: remote local config does not see existing remote

**Output of `dvc version`:**

```console
DVC version: 1.1.10+1c9588
Python version: 3.8.4
Platform: macOS-10.15.4-x86_64-i386-64bit
Binary: False
Package: None
Repo: dvc, git
```

```
dvc remote add -d storage s3://storage
dvc remote modify storage region us-east-2 --local
```

complains:

```
ERROR: configuration error - config file error: remote 'storage' doesn't exists.
```

For example, it breaks the use case of using `dvc remote modify` to store credentials in the local config only.

efiop:
Fixes #4276

### DIFF
--- a/tests/func/test_remote.py
+++ b/tests/func/test_remote.py
@@ -232,6 +232,13 @@ def test_modify_missing_remote(tmp_dir, dvc):
     assert main(["remote", "modify", "myremote", "user", "xxx"]) == 251
 
 
+def test_modify_parent_remote(tmp_dir, dvc):
+    assert main(["remote", "add", "myremote", "http://example.com/path"]) == 0
+    assert (
+        main(["remote", "modify", "myremote", "user", "xxx", "--local"]) == 0
+    )
+
+
 def test_external_dir_resource_on_no_cache(tmp_dir, dvc, tmp_path_factory):
     # https://github.com/iterative/dvc/issues/2647, is some situations
     # (external dir dependency) cache is required to calculate dir md5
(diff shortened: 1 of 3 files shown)


tests/func/test_remote.py
```diff
@@ -232,6 +232,13 @@ def test_modify_missing_remote(tmp_dir, dvc):
     assert main(["remote", "modify", "myremote", "user", "xxx"]) == 251
 
 
+def test_modify_parent_remote(tmp_dir, dvc):
```
skshetry, line 235: What's a parent remote? Just find it a bit confusing.

tests/func/test_remote.py
```diff
@@ -232,6 +232,13 @@ def test_modify_missing_remote(tmp_dir, dvc):
...
+    assert main(["remote", "add", "myremote", "http://example.com/path"]) == 0
+    assert (
+        main(["remote", "modify", "myremote", "user", "xxx", "--local"]) == 0
+    )
```
skshetry, line 239: Should we also try reading the config? I understand that the issue was during `modify` but I don't want to trust that.

tests/func/test_remote.py
efiop: Yeah, called it parent for a lack of better word. In this case it means that .dvc/config is a parent config for .dvc/config.local

skshetry:
```suggestion
def test_remote_modify_local_on_repo_config(tmp_dir, dvc):
```

tests/func/test_remote.py
efiop: Good point! Added more throughout checks.

PR #4320, opened 2020-08-03
experiments: more intelligent checkout behavior

### LINKED ISSUE #4289: experiments: `core.experiments` config setting overridden in certain situations
If `core.experiments` is set in `.dvc/config` and not committed, experiments support will end up being disabled when we do `reset --hard` during `experiments checkout`. This also affects `.dvc/experiments` being added to `.gitignore`.

Temporary workarounds:
* Use `dvc config --local` when setting `core.experiments true`
* `git commit` changes to `.dvc/config` after doing `dvc config core.experiments true`, and before running any other experiments related commands

pmrowla:

Should fix #4289.

* No longer use `git reset --hard` when doing checkout.
* Changes from the experiment commit are preferred over workspace changes, but all other unrelated and unstaged/uncommitted workspace changes will now be preserved after `dvc experiments checkout`

### DIFF
--- a/dvc/repo/experiments/__init__.py
+++ b/dvc/repo/experiments/__init__.py
@@ -27,7 +27,15 @@ def hash_exp(stages):
 
 class UnchangedExperimentError(DvcException):
     def __init__(self, rev):
-        super().__init__("Experiment identical to baseline '{rev[:7]}'.")
+        super().__init__(f"Experiment identical to baseline '{rev[:7]}'.")
+        self.rev = rev
+
+
+class BaselineMismatchError(DvcException):
+    def __init__(self, rev):
+        super().__init__(
+            f"Experiment is not derived from current baseline '{rev[:7]}'."
+        )
         self.rev = rev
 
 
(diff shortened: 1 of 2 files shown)


dvc/repo/experiments/__init__.py
```diff
@@ -79,6 +88,13 @@ def exp_dvc(self):
 
         return Repo(self.exp_dvc_dir)
 
+    @contextmanager
+    def chdir(self):
+        cwd = os.getcwd()
+        os.chdir(self.exp_dvc.root_dir)
+        yield self.exp_dvc.root_dir
+        os.chdir(cwd)
+
```
efiop, line 97: Don't see it being used anywhere

pmrowla: It's used in `experiments show`, but got removed in the last PR by mistake.

https://github.com/iterative/dvc/blob/11d906ee517e1980ce9884427a33b17253b1e4c2/dvc/repo/experiments/show.py#L61

PR #4323, opened 2020-08-03
Two new modes `--all` and `--stdin` for check-ignore

### LINKED ISSUE #4321: Interactive mode and all patterns mode for dvc check-ignore command
Follow #3736

Two more modes: 
1. interactive mode `--stdin` which reads pathnames from the standard input.
2. all patterns mode `--all` which prints all pattern matches instead of only the last one.

karajan1001:
Fix #4321

Introduce new arguments --all for dvc check-ignore
Add tests for dvc check-ignore --all

### DIFF
--- a/dvc/command/check_ignore.py
+++ b/dvc/command/check_ignore.py
@@ -41,7 +41,7 @@ class CmdCheckIgnore(CmdBase):
             target = ask("")
             if target == "":
                 logger.info(
-                    "empty string is not a valid pathspec. please use . "
+                    "Empty string is not a valid pathspec. Please use . "
                     "instead if you meant to match all paths"
                 )
                 break
--- a/dvc/command/check_ignore.py
+++ b/dvc/command/check_ignore.py
@@ -61,7 +61,7 @@ class CmdCheckIgnore(CmdBase):
             raise DvcException("targets or --stdin needed")
 
         if self.args.stdin and self.args.targets:
-            raise DvcException("cannot have targets and --stdin both")
+            raise DvcException("cannot have both `targets` and `--stdin`")
 
         if self.args.non_matching and not self.args.details:
             raise DvcException("--non-matching is only valid with --details")
--- a/dvc/command/check_ignore.py
+++ b/dvc/command/check_ignore.py
@@ -56,7 +56,13 @@ class CmdCheckIgnore(CmdBase):
                 ret = 0
         return ret
 
-    def run(self):
+    def _check_args(self):
+        if not self.args.stdin and not self.args.targets:
+            raise DvcException("targets or --stdin needed")
+
+        if self.args.stdin and self.args.targets:
+            raise DvcException("cannot have targets and --stdin both")
+
         if self.args.non_matching and not self.args.details:
             raise DvcException("--non-matching is only valid with --details")
 
@@ -66,6 +72,8 @@ class CmdCheckIgnore(CmdBase):
         if self.args.quiet and self.args.details:
             raise DvcException("cannot both --details and --quiet")
 
+    def run(self):
+        self._check_args()
         if self.args.stdin:
             ret = self._interactive_mode()
         else:
@@ -115,7 +123,7 @@ def add_parser(subparsers, parent_parser):
     )
     parser.add_argument(
         "targets",
-        nargs="+",
+        nargs="*",
         help="Exact or wildcard paths of files or directories to check "
         "ignore patterns.",
     ).complete = completion.FILE
--- a/dvc/command/check_ignore.py
+++ b/dvc/command/check_ignore.py
@@ -42,7 +42,7 @@ class CmdCheckIgnore(CmdBase):
             if target == "":
                 logger.info(
                     "Empty string is not a valid pathspec. Please use . "
-                    "instead if you meant to match all paths"
+                    "instead if you meant to match all paths."
                 )
                 break
             if not self._check_one_file(target):
--- a/dvc/command/check_ignore.py
+++ b/dvc/command/check_ignore.py
@@ -4,6 +4,7 @@ import logging
 from dvc.command import completion
 from dvc.command.base import CmdBase, append_doc_link
 from dvc.exceptions import DvcException
+from dvc.prompt import path_input
 
 logger = logging.getLogger(__name__)
 
@@ -14,25 +15,61 @@ class CmdCheckIgnore(CmdBase):
         self.ignore_filter = self.repo.tree.dvcignore
 
     def _show_results(self, result):
-        if result.match or self.args.non_matching:
-            if self.args.details:
-                logger.info("{}\t{}".format(result.patterns[-1], result.file))
-            else:
-                logger.info(result.file)
+        if not result.match and not self.args.non_matching:
+            return
+
+        if self.args.details:
+            patterns = result.patterns
+            if not self.args.all:
+                patterns = patterns[-1:]
+
+            for pattern in patterns:
+                logger.info("{}\t{}".format(pattern, result.file))
+        else:
+            logger.info(result.file)
+
+    def _check_one_file(self, target):
+        result = self.ignore_filter.check_ignore(target)
+        self._show_results(result)
+        if result.match:
+            return 0
+        return 1
+
+    def _interactive_mode(self):
+        ret = 1
+        while True:
+            target = path_input()
+            if target == "":
+                logger.info(
+                    "empty string is not a valid pathspec. please use . "
+                    "instead if you meant to match all paths"
+                )
+                break
+            if not self._check_one_file(target):
+                ret = 0
+        return ret
+
+    def _normal_mode(self):
+        ret = 1
+        for target in self.args.targets:
+            if not self._check_one_file(target):
+                ret = 0
+        return ret
 
     def run(self):
         if self.args.non_matching and not self.args.details:
             raise DvcException("--non-matching is only valid with --details")
 
+        if self.args.all and not self.args.details:
+            raise DvcException("--all is only valid with --details")
+
         if self.args.quiet and self.args.details:
             raise DvcException("cannot both --details and --quiet")
 
-        ret = 1
-        for target in self.args.targets:
-            result = self.ignore_filter.check_ignore(target)
-            self._show_results(result)
-            if result.match:
-                ret = 0
+        if self.args.stdin:
+            ret = self._interactive_mode()
+        else:
+            ret = self._normal_mode()
         return ret
 
 
@@ -59,6 +96,21 @@ def add_parser(subparsers, parent_parser):
         action="store_true",
         default=False,
         help="Show the target paths which don’t match any pattern. "
+        "Only usable when `--details` is also employed",
+    )
+    parser.add_argument(
+        "--stdin",
+        action="store_true",
+        default=False,
+        help="Read pathnames from the standard input, one per line, "
+        "instead of from the command-line.",
+    )
+    parser.add_argument(
+        "-a",
+        "--all",
+        action="store_true",
+        default=False,
+        help="Show all of the patterns match the target paths. "
         "Only usable when `--details` is also employed",
     )
     parser.add_argument(
--- a/dvc/command/check_ignore.py
+++ b/dvc/command/check_ignore.py
@@ -70,7 +70,7 @@ class CmdCheckIgnore(CmdBase):
             raise DvcException("`--all` is only valid with `--details`")
 
         if self.args.quiet and self.args.details:
-            raise DvcException("cannot both --details and --quiet")
+            raise DvcException("cannot use both `--details` and `--quiet`")
 
     def run(self):
         self._check_args()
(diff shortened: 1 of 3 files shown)


dvc/command/check_ignore.py
```diff
@@ -61,6 +98,21 @@ def add_parser(subparsers, parent_parser):
...
+        "-a",
+        "--all",
+        action="store_true",
+        default=False,
+        help="Show all of the patterns match the target paths. "
+        "Only usable when `--details` is also employed",
+    )
     parser.add_argument(
         "targets",
         nargs="+",
```
pared, line 118: Target needs to be changed to `*` in order to make `stdin` work.

karajan1001:
> LGTM, one change request.
@pared 
Thank you, I'm so careless to write such a bug. 

And I had a problem. I tried to add a pytest to `--stdin` mode but failed.

I used monkeypatch to mock user input but was blocked by 
https://github.com/iterative/dvc/blob/35dd1fda9cfb4b645ae431f4621f2d1853d4fb7c/dvc/prompt.py#L11
Are there any methods to pass this line?

pared:
@karajan1001 
I guess most straightforward would be to mock `isatty` too, though I will see if there is something else to be done here.

karajan1001:
> @karajan1001
> I guess most straightforward would be to mock `isatty` too, though I will see if there is something else to be done here.

@pared  
Oh thank you very much, it works. But I don't know if there are some risks to influence other tests.

pared:
@karajan1001 
>But I don't know if there are some risks to influence other tests.

When you are using a mocker inside a test, patches and mocks are applied only to this test. So, if you are worried, whether it's possible to "leak" the mock to other tests - no, it should not happen. 
 
> I will see if there is something else to be done here

I think it's fine to use this patch. `isatty` tests whether we are working from terminal, so it is a situation we are trying to simulate.

jorgeorpinel:
Hi!

Maybe you can address https://github.com/iterative/dvc/pull/4282#pullrequestreview-460270636 and https://github.com/iterative/dvc/pull/4282#pullrequestreview-460272287 here @karajan1001 ? Thanks!

efiop:
> Maybe you can address #4282 (review) and #4282 (review) here @karajan1001 ? Thanks!

@jorgeorpinel Let's not do that. Let's not expand the scope of this PR, those changes are completely unrelated.

karajan1001:
> > Maybe you can address #4282 (review) and #4282 (review) here @karajan1001 ? Thanks!
> 
> @jorgeorpinel Let's not do that. Let's not expand the scope of this PR, those changes are completely unrelated.

I had considered doing so, but give up.  BTW, doc of this PR needs to be reviewed as well.

dvc/command/check_ignore.py
```diff
@@ -61,9 +106,24 @@ def add_parser(subparsers, parent_parser):
         help="Show the target paths which don’t match any pattern. "
         "Only usable when `--details` is also employed",
     )
+    parser.add_argument(
+        "--stdin",
+        action="store_true",
+        default=False,
+        help="Read pathnames from the standard input, one per line, "
+        "instead of from the command-line.",
+    )
```
efiop, line 115: What would be the typical use cases for --stdin? I've honestly never used it in git's check-ignore, so I'm not sure I understand who and why would use it. Just worried about a potentially unused/dead feature. `--all` is very useful, no doubt.

dvc/prompt.py
```diff
@@ -54,3 +54,12 @@ def password(statement):
...
     return getpass("")
+
+
+def path_input():
+    """Ask the user for a path.
+
+    Returns:
+        str: path entered by the user.
+    """
+    return _ask("")
```
efiop, line 65: This one is only used in check-ignore, so let's make `ask` public and use it directly in `check-ignore`.

dvc/command/check_ignore.py
karajan1001: @efiop 
Most use cases are inside a Linux pipeline, I think.
```
find . -ctime -7 | dvc check-ignore --stdin 
```

dvc/command/check_ignore.py
```diff
@@ -14,25 +15,69 @@ def __init__(self, args):
...
+            return 0
+        return 1
+
+    def _interactive_mode(self):
+        ret = 1
+        while True:
+            target = ask("")
+            if target == "":
+                logger.info(
+                    "empty string is not a valid pathspec. please use . "
```
efiop, line 44:
```suggestion
                    "Empty string is not a valid pathspec. Please use . "
```

dvc/command/check_ignore.py
```diff
@@ -14,25 +15,69 @@ def __init__(self, args):
...
+        return 1
+
+    def _interactive_mode(self):
+        ret = 1
+        while True:
+            target = ask("")
+            if target == "":
+                logger.info(
+                    "Empty string is not a valid pathspec. Please use . "
+                    "instead if you meant to match all paths"
```
efiop, line 45:
```suggestion
                    "instead if you meant to match all paths."
```

dvc/command/check_ignore.py
```diff
@@ -14,25 +15,69 @@ def __init__(self, args):
...
+            if not self._check_one_file(target):
+                ret = 0
+        return ret
+
+    def _check_args(self):
+        if not self.args.stdin and not self.args.targets:
+            raise DvcException("targets or --stdin needed")
+
+        if self.args.stdin and self.args.targets:
+            raise DvcException("cannot have targets and --stdin both")
```
efiop, line 64:
```suggestion
            raise DvcException("cannot have both `targets` and `--stdin`")
```

dvc/command/check_ignore.py
```diff
@@ -14,25 +15,69 @@ def __init__(self, args):
...
+    def _normal_mode(self):
+        ret = 1
+        for target in self.args.targets:
+            if not self._check_one_file(target):
+                ret = 0
+        return ret
+
+    def _check_args(self):
+        if not self.args.stdin and not self.args.targets:
+            raise DvcException("targets or --stdin needed")
```
efiop, line 61:
```suggestion
            raise DvcException("`targets` or `--stdin` needed")
```

dvc/command/check_ignore.py
```diff
@@ -14,25 +15,69 @@ def __init__(self, args):
...
+
+        if self.args.stdin and self.args.targets:
+            raise DvcException("cannot have both `targets` and `--stdin`")
 
-    def run(self):
         if self.args.non_matching and not self.args.details:
             raise DvcException("--non-matching is only valid with --details")
 
+        if self.args.all and not self.args.details:
+            raise DvcException("--all is only valid with --details")
```
efiop, line 70:
```suggestion
            raise DvcException("`--all` is only valid with `--details`")
```

dvc/command/check_ignore.py
```diff
@@ -14,25 +15,69 @@ def __init__(self, args):
...
 
-    def run(self):
         if self.args.non_matching and not self.args.details:
             raise DvcException("--non-matching is only valid with --details")
 
+        if self.args.all and not self.args.details:
+            raise DvcException("`--all` is only valid with `--details`")
+
         if self.args.quiet and self.args.details:
             raise DvcException("cannot both --details and --quiet")
```
efiop, line 73:
```suggestion
            raise DvcException("cannot use both `--details` and `--quiet`")
```

dvc/command/check_ignore.py
```diff
@@ -14,27 +15,71 @@ def __init__(self, args):
...
+        if self.args.quiet and self.args.details:
+            raise DvcException("cannot use both `--details` and `--quiet`")
+
+    def run(self):
+        self._check_args()
+        if self.args.stdin:
+            ret = self._interactive_mode()
+        else:
+            ret = self._normal_mode()
+        return ret
```
efiop, line 81:
```suggestion
            return self._interactive_mode()
        return self._normal_mode()
```

dvc/command/check_ignore.py
```diff
@@ -14,27 +15,69 @@ def __init__(self, args):
...
 
+    def _check_args(self):
+        if not self.args.stdin and not self.args.targets:
+            raise DvcException("`targets` or `--stdin` needed")
+
+        if self.args.stdin and self.args.targets:
+            raise DvcException("cannot have both `targets` and `--stdin`")
+
+        if self.args.non_matching and not self.args.details:
+            raise DvcException("--non-matching is only valid with --details")
```
efiop, line 67:
```suggestion
            raise DvcException("`--non-matching` is only valid with `--details`")
```

efiop:

jorgeorpinel:
> BTW, doc of this PR needs to be reviewed as well.

@karajan1001 is there a docs PR for this already? We haven't even merged the previous one 😅

karajan1001:
@jorgeorpinel 

Not already, I'm waiting for this one. And it would be created soon.

dvc/command/check_ignore.py
```diff
@@ -14,27 +15,71 @@ def __init__(self, args):
...
+        return 1
 
+    def _interactive_mode(self):
+        ret = 1
+        while True:
+            target = ask("")
+            if target == "":
+                logger.info(
+                    "Empty string is not a valid pathspec. Please use . "
+                    "instead if you meant to match all paths."
```
jorgeorpinel, line 45: What is a pathspec?
Can users actually send an empty string (e.g. `''`)? Or does this happen when you don't send any args? In the latter case the message could be simplified to something like "Please provide file or directory paths to check."

dvc/command/check_ignore.py
```diff
@@ -14,27 +15,71 @@ def __init__(self, args):
...
+        if self.args.non_matching and not self.args.details:
+            raise DvcException(
+                "`--non-matching` is only valid with `--details`"
+            )
+
+        if self.args.all and not self.args.details:
+            raise DvcException("`--all` is only valid with `--details`")
+
+        if self.args.quiet and self.args.details:
+            raise DvcException("cannot use both `--details` and `--quiet`")
```
jorgeorpinel, line 75: Maybe --quiet shouold just win here?

dvc/command/check_ignore.py
```diff
@@ -61,9 +106,24 @@ def add_parser(subparsers, parent_parser):
         help="Show the target paths which don’t match any pattern. "
         "Only usable when `--details` is also employed",
     )
+    parser.add_argument(
+        "--stdin",
+        action="store_true",
+        default=False,
+        help="Read pathnames from the standard input, one per line, "
+        "instead of from the command-line.",
```
jorgeorpinel, line 114: "Read path names interactively instead of providing `targets`."

jorgeorpinel: BTW how do you exit, Ctrl+C?

dvc/command/check_ignore.py
```diff
@@ -61,9 +106,24 @@ def add_parser(subparsers, parent_parser):
...
+    )
+    parser.add_argument(
+        "-a",
+        "--all",
+        action="store_true",
+        default=False,
+        help="Show all of the patterns match the target paths. "
+        "Only usable when `--details` is also employed",
```
jorgeorpinel, line 122: "Show all target paths and whether they match a pattern or not."

jorgeorpinel: I'd also move this option before -n, since it seems more useful IMO.

jorgeorpinel: > Show all target paths

Hmmm. I see that's not exactly what this option does. It shows all the patterns that match each target indeed so maybe just:

"Show all the patterns that match each target path."

And is there any way to have a --details list with both matched and not matched targets?

dvc/command/check_ignore.py
jorgeorpinel, line 114: "Read paths from standard input instead of providing `targets`."

dvc/command/check_ignore.py
jorgeorpinel: > And is there any way to have a --details list with both matched and not matched targets?

I see -n already does this 🙂 — will probably update in docs.

dvc/command/check_ignore.py
karajan1001: @jorgeorpinel 
1. `pathspec` means the target to be checked
2. It is in an interactive mode where users can type string as the target, and this happened when users type send an empty string.

dvc/command/check_ignore.py
karajan1001: > Maybe --quiet should just win here?

Yes, it used to be `--verbose` and `--quiet`. They might cause some problems, and had been fixed in #4304.
And for `--details` and `--quiet` we can just let `--quiet` win.

dvc/command/check_ignore.py
jorgeorpinel: > It is in an interactive mode
So maybe don't say anything and just go to the next `>` ? Seems intuitive enough that way. I just tried it and noticed it crashes after this error msg (i.e. it stops the interactive mode).

jorgeorpinel: p.s. when I send an empty string as target i.e. `dvc check-ignore ''` it just checks it (never matches even if `*` is a pattern in `.dvcignore`) and this error isn't presented. Seems slightly inconsistent. I vote to just remove it altogether 🙂

jorgeorpinel: Extracted to #4361

Issue #3369, opened 2020-02-20
dvc get/import support for subrepos


ehutt:
UPDATE: See https://github.com/iterative/dvc/issues/3369#issuecomment-588657129 for main purpose of this issue

---

If I have a data registry monorepo with several subrepos inside it, each pointing to it's own remote storage path, I want to be able to import and/or get data from a specific subrepo/remote location. 

My situation: I have an s3 bucket, `s3://datasets/` where I store multiple datasets, say `audio/` and `text/`. I want to track these data using dvc and a single, git-based data registry.  Then I want to be able to selectively push and pull/import data from s3 through this registry for use in my other dvc projects. 

So, I can make subrepos for `audio/` and `text/`, each initialized with its own dvc file and remote, and push data to s3 this way. Then, if I want to only download the audio data into a new project, then I can run something like `dvc import git@github.com/datasets/audio` and it will automatically pull from the correct path in s3, corresponding to the default remote for the audio subrepo. 

Thank you!

jorgeorpinel:
Hi @ehutt 

> I have a data registry monorepo with several subrepos inside it, each pointing to it's own remote storage 

At first I thought this was essentially the same as #2349, but I see its not about subdirectories but rather about Git subrepos that are DVC projects in a parent Git repo (which is not a DVC project). Am I correct? Maybe if you can provide a sample file hierarchy including .git and .dvc dirs, this could be cleared up.

> I have an s3 bucket where I store multiple datasets, say audio/ and text/. I want to track these using dvc...
> So, I can make subrepos for audio/ and text/, each initialized with its own dvc file and remote...

I'm not sure you need to use separate DVC projects for each sub-dataset. You could follow the pattern explained in https://dvc.org/doc/use-cases/data-registries which simply track multiple directories in a single DVC repo.

BTW, since your data is already in S3, you probably want to `dvc add` the sub-datasets as [external outputs](https://dvc.org/doc/user-guide/managing-external-data) in your data registry.

> Then I want to be able to selectively push and pull/import data from s3 through this registry for use in my other dvc projects...
> then I can run something like dvc import git@github.com/datasets/audio and it will automatically pull from the correct path in s3

Yep. `dvc get` or `dvc import` would be the commands for this. (Pulling and pushing are for synchronizing data between the cache of a DVC project and its remote storage.)

jorgeorpinel:
[...]

PR #4324, opened 2020-08-03
dvcignore: api to unhide subrepos directory

skshetry:

Part of #3369

I implemented this inside dvcignore, and it
should not require "internal" APIs hack as we discussed before.

Right now, `ignore_subrepos` will only work for the paths of subrepo,
not anything inside it. And the API user is more or less only
RepoTree, so should be safe (we switch DvcTree and repo.tree based
on path prefixes, so this is only required during `walk`, so as it
shows the repo that were previously dvcignored and so that we could
switch the trees later during walk.

So, if dir is `dir -> repo`
So, `tree.walk("dir")` will return `[]`,
whereas `tree.walk("dir", ignore_subrepos=False)` will return `["repo"]`
But, successive walk will just return `("repo", [], [])`, and could
just be ignored as a side-effect.

### DIFF
--- a/dvc/ignore.py
+++ b/dvc/ignore.py
@@ -6,12 +6,12 @@ from itertools import groupby, takewhile
 
 from pathspec.patterns import GitWildMatchPattern
 from pathspec.util import normalize_file
-from pygtrie import StringTrie
 
 from dvc.path_info import PathInfo
 from dvc.pathspec_math import PatternInfo, merge_patterns
 from dvc.system import System
 from dvc.utils import relpath
+from dvc.utils.collections import PathStringTrie
 
 logger = logging.getLogger(__name__)
 
@@ -156,7 +156,7 @@ class DvcIgnoreFilterNoop:
     def __init__(self, tree, root_dir):
         pass
 
-    def __call__(self, root, dirs, files):
+    def __call__(self, root, dirs, files, **kwargs):
         return dirs, files
 
     def is_ignored_dir(self, _):
@@ -183,10 +183,11 @@ class DvcIgnoreFilter:
 
         self.tree = tree
         self.root_dir = root_dir
-        self.ignores_trie_tree = StringTrie(separator=os.sep)
+        self.ignores_trie_tree = PathStringTrie()
         self.ignores_trie_tree[root_dir] = DvcIgnorePatterns(
             default_ignore_patterns, root_dir
         )
+        self._ignored_subrepos = PathStringTrie()
         self._update(self.root_dir)
 
     def _update(self, dirname):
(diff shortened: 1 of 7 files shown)


dvc/ignore.py
```diff
@@ -222,7 +240,8 @@ def _update(self, dirname):
     def _update_sub_repo(self, root, dirs):
         for d in dirs:
             if self._is_dvc_repo(root, d):
-                new_pattern = DvcIgnorePatterns(["/{}/".format(d)], root)
+                self._ignored_subrepos.add(f"{root}{os.sep}{d}")
```
efiop, line 243: Why not `os.path.join` ? If for perf reasons, then it is probably not a concern here, as we do this once for every parent in a path, which is negligible.

dvc/ignore.py
```diff
@@ -16,6 +16,16 @@
 logger = logging.getLogger(__name__)
 
 
+class PathStringTrie(StringTrie):
+    """Trie based on platform-dependent separator for pathname components."""
+
+    # pylint: disable=useless-super-delegation
+    def __init__(self, *args, separator=os.sep, **kwargs):
```
efiop, line 23: Doesn't look like anyone is going to use it anyways
```suggestion
    def __init__(self, *args, **kwargs):
```

dvc/ignore.py
```diff
@@ -222,7 +240,8 @@ def _update(self, dirname):
     def _update_sub_repo(self, root, dirs):
         for d in dirs:
             if self._is_dvc_repo(root, d):
-                new_pattern = DvcIgnorePatterns(["/{}/".format(d)], root)
+                self._ignored_subrepos.add(f"{root}{os.sep}{d}")
+                new_pattern = DvcIgnorePatterns([f"/{d}/"], root)
```
efiop, line 244: I know that fstrings are better than format() , but there wasn't really a reason to change it in this PR :slightly_smiling_face:

dvc/ignore.py
```diff
@@ -183,12 +193,20 @@ def __init__(self, tree, root_dir):
...
         )
+        self._ignored_subrepos = PrefixSet(factory=PathStringTrie)
         self._update(self.root_dir)
 
+    @property
+    def ignored_subrepos(self):
+        # as we go collecting subrepos dynamically, `_ignore_subrepos` will be
+        # populated which is useful for finding out if any subrepos were
+        # encountered along the way
+        return self._ignored_subrepos
```
efiop, line 208: Not sure I follow. Is it a draft for the future functionality? Right not it just acts as a proxy and seems meaningless.

dvc/ignore.py
```diff
@@ -222,7 +240,8 @@ def _update(self, dirname):
     def _update_sub_repo(self, root, dirs):
         for d in dirs:
             if self._is_dvc_repo(root, d):
```
efiop, line 242: Btw, looks like we might want to populate descendants with the last subrepo we've encountered. That way we will be able to distinguish uncollected path. Similar to how we do it with the ignore trie.

dvc/ignore.py
skshetry: I had to change that before when I was experimenting with it. And, f-string is easy to read and understand (here, speed does not matter but could have been another reason), and therefore, I kept it.

PR #4339, opened 2020-08-05
exception text: clarify error message

### LINKED ISSUE #4316: Did you forget to fetch? is generally not the solution to my problem
## Bug Report
I occasionally do ```dvc pull``` and get a failure with the message "Did you forget to fetch?". In my case the cause has always been that I've forgotten to push on another machine. The message could be expanded to cover this use case.

**Output of `dvc version`:**

```console
1.2.1
```

bobertlo:
The default error message is often invoked from running `dvc pull` and
in this case it is generally caused by not pushing to the remote. I
spent some time thinking about a replacement message and this seems to
be the most appropriate text I could think of.

Fixes #4316

  https://github.com/iterative/dvc.org/issues/1660

### DIFF
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -268,8 +268,11 @@ class CheckoutError(DvcException):
         self.stats = stats
         targets = [str(t) for t in target_infos]
         m = (
-            "Checkout failed for following targets:\n{}\nDid you "
-            "forget to fetch?".format("\n".join(targets))
+            "Checkout failed for following targets:\n{}\nIs your "
+            "cache up to date?\n{}".format(
+                "\n".join(targets),
+                format_link("https://error.dvc.org/missing-files"),
+            )
         )
         super().__init__(m)
 


dvc/exceptions.py
```diff
@@ -268,8 +268,8 @@ def __init__(self, target_infos, stats=None):
         self.stats = stats
         targets = [str(t) for t in target_infos]
         m = (
-            "Checkout failed for following targets:\n{}\nDid you "
-            "forget to fetch?".format("\n".join(targets))
+            "Checkout failed for following targets:\n{}\nIs your "
+            "cache up to date?".format("\n".join(targets))
```
efiop, line 272: Hm, not sure if this makes it clearer. The OP https://github.com/iterative/dvc/issues/4316 was mentioning that he forgets to push stuff on another machine, and this still doesn't seem to be directing in the right direction. Maybe we should actually add a link with common issues for such warning and have an article about it on dvc.org? Kinda like we do with https://dvc.org/doc/user-guide/troubleshooting (maybe that's even a good place for such hints as well). What do you think? CC @braaannigan

braaannigan: The cache is a bit mysterious for new users and this message wouldn't necessarily cause me to think about updating the cache from my other machine. I like the idea about having a link to a web page with common issues

pared: I am not sure changing the message will do the trick here. I wrote some thoughts on that under the [original issue](https://github.com/iterative/dvc/issues/4316#issuecomment-669138615).

bobertlo:
I guess I will admit this does not "solve" the original issue. This error message always really bothered me though :)

bobertlo:
I edited this to print a link to the troubleshooting page, edited in https://github.com/iterative/dvc.org/pull/1671 with a short explanation of common problems.

dvc/exceptions.py
```diff
@@ -268,8 +268,11 @@ def __init__(self, target_infos, stats=None):
         self.stats = stats
         targets = [str(t) for t in target_infos]
         m = (
-            "Checkout failed for following targets:\n{}\nDid you "
-            "forget to fetch?".format("\n".join(targets))
+            "Checkout failed for following targets:\n{}\nIs your "
+            "cache up to date?\n{}".format(
```
efiop, line 272: The cache question seems unclear, how about just:
```suggestion
            "Checkout failed for following targets:\n{}\n"
            "Visit {} for more info.".format(
```
?

bobertlo:
I suppose the errors provided during fetch are pretty clear. I mainly did not like the 'did you forget to fetch?' message. I can change this one I get near a terminal again.

bobertlo:
@efiop Okay, made that exact change. I agree that since we're linking to the troubleshooting page we don't need anything more than the above error messages in the context.

dvc/exceptions.py
skshetry: "Visit {} for more info." sounds like holding a user hostage. We do need better exception message, which can be done by making `_fetch` and `_checkout` collaborate with each other, but making `CheckoutError` message generic is not a solution, it only adds confusion.

bobertlo:
Should checkout be called automatically from `dvc pull` even if fetch
fails? If it stops there, fetch and checkout can just have proper error
messages for themselves. Otherwise, yes they will need to collaborate.

dvc/exceptions.py
efiop: @skshetry Good point. Maybe for a lack of better message (because a more advanced approach is needed), let's just append "Vinit {} for more info"(or someting like that) to the old message? Just in the interest of time, since the proper-proper solution will require quite a bit more effort.

bobertlo: @efiop I was sort of thinking the same thing myself. I updated it to that and removed the "Fixes #4316" since that issue goes deeper than this. I think other changes are unrelated to this PR at this point?

skshetry: @efiop, yeah, that sounds good. But let's keep the issue open.

PR #4340, opened 2020-08-05
import: fix importing into subdirectory bug

### LINKED ISSUE #4169: import: same command creates .dvc file in different locations
## DVC Version

```console
~/example-get-started$ dvc version
DVC version: 1.0.1
Python version: 3.7.1
Platform: Linux-4.15.0-88-generic-x86_64-with-debian-buster-sid
Binary: True
Package: deb
Filesystem type (cache directory): ('ext4', '/dev/mapper/host01--vg-root')
Repo: dvc, git
Filesystem type (workspace): ('ext4', '/dev/mapper/host01--vg-root')
```

## Reproduce

Run the same command:

`dvc import https://github.com/iterative/dataset-registry get-started/data.xml -o data/data.xml`

when `data` directory exists and when it does not.

In the first case we get:

```console
Importing 'get-started/data.xml (https://github.com/iterative/dataset-registry)' -> 'data/data.xml'

        git add data/data.xml.dvc
```

In the second case:

```console
Importing 'get-started/data.xml (https://github.com/iterative/dataset-registry)' -> 'data/data.xml'

        git add data.xml.dvc data/.gitignore
```

**There are two bugs:**

- [ ] It creates `data.xml.dvc` in different locations (`data/data.xml.dvc` vs `data.xml.dvc`)
- [x] It doesn't mention `.gitignore` in the first case in the `git add` hint.

pmrowla:

Fixes #4169.

Makes `dvc import ... -o subdir/file` behave the same whether or not `subdir` already exists (new output will be `subdir/file.dvc`)

docs: https://github.com/iterative/dvc.org/pull/1668

### DIFF
--- a/dvc/repo/imp_url.py
+++ b/dvc/repo/imp_url.py
@@ -2,7 +2,7 @@ import os
 
 from dvc.repo.scm_context import scm_context
 from dvc.utils import relpath, resolve_output, resolve_paths
-from dvc.utils.fs import makedirs, path_isin
+from dvc.utils.fs import path_isin
 
 from ..exceptions import OutputDuplicationError
 from . import locked
@@ -26,9 +26,6 @@ def imp_url(
         and path_isin(os.path.abspath(url), self.root_dir)
     ):
         url = relpath(url, wdir)
-
-    if not os.path.exists(wdir):
-        makedirs(wdir)
 
     stage = create_stage(
         Stage,
--- a/dvc/repo/imp_url.py
+++ b/dvc/repo/imp_url.py
@@ -2,7 +2,7 @@ import os
 
 from dvc.repo.scm_context import scm_context
 from dvc.utils import relpath, resolve_output, resolve_paths
-from dvc.utils.fs import path_isin
+from dvc.utils.fs import makedirs, path_isin
 
 from ..exceptions import OutputDuplicationError
 from . import locked
@@ -26,6 +26,9 @@ def imp_url(
         and path_isin(os.path.abspath(url), self.root_dir)
     ):
         url = relpath(url, wdir)
+
+    if not os.path.exists(wdir):
+        makedirs(wdir)
 
     stage = create_stage(
         Stage,
--- a/dvc/utils/__init__.py
+++ b/dvc/utils/__init__.py
@@ -373,11 +373,9 @@ def resolve_paths(repo, out):
 
     # NOTE: `out` might not exist yet, so using `dirname`(aka `wdir`) to check
     # if it is a local path.
-    if (
-        os.path.exists(dirname)  # out might not exist yet, so
-        and PathInfo(abspath).isin_or_eq(repo.root_dir)
-        and not contains_symlink_up_to(abspath, repo.root_dir)
-    ):
+    if PathInfo(abspath).isin_or_eq(
+        repo.root_dir
+    ) and not contains_symlink_up_to(abspath, repo.root_dir):
         wdir = dirname
         out = base
     else:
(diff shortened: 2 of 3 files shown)


dvc/utils/__init__.py
```diff
@@ -371,10 +372,8 @@ def resolve_paths(repo, out):
     dirname = os.path.dirname(abspath)
     base = os.path.basename(os.path.normpath(out))
 
-    # NOTE: `out` might not exist yet, so using `dirname`(aka `wdir`) to check
-    # if it is a local path.
     if (
-        os.path.exists(dirname)  # out might not exist yet, so
+        not urlparse(out).scheme
```
pmrowla, line 376: `out` may be a local path or a `remote://...` url, but `os.path.exists` isn't a good enough check whether or not `out` is a local path.

In the case where we are importing with `-o subdir/file`, if `subdir` doesn't already exist it is still a valid local output path (and we need to `makedirs(subdir)` later)

dvc/utils/__init__.py
```diff
@@ -363,19 +363,22 @@ def resolve_output(inp, out):
...
-    # if it is a local path.
+    scheme = urlparse(out).scheme
+    if os.name == "nt" and scheme == abspath.drive[0].lower():
+        # urlparse interprets windows drive letters as URL scheme
+        scheme = ""
     if (
-        os.path.exists(dirname)  # out might not exist yet, so
-        and PathInfo(abspath).isin_or_eq(repo.root_dir)
+        not scheme
+        and abspath.isin_or_eq(repo.root_dir)
```
pmrowla, line 381: `out` may be a local path or a `remote://...` url, but `os.path.exists` isn't a good enough check for whether or not `out` is a valid local path.

In the case where we are importing with `-o subdir/file`, if `subdir` doesn't already exist it is still a valid local output path (and we need to `makedirs(subdir)` later)

dvc/repo/imp_url.py
```diff
@@ -27,6 +27,9 @@ def imp_url(
     ):
         url = relpath(url, wdir)
 
+    if not os.path.exists(wdir):
+        makedirs(wdir)
+
```
efiop, line 32: Would it make sense to error-out if it doesn't exist? Kinda like `cp` and other tools do.

pmrowla: Creating the directory if it doesn't exist feels like the more natural `import -o` behavior for me, but given that the current docs for `-o` say:
> If an existing directory is specified, the target data will be placed inside.
it would also be okay for us to error out here.

shcheklein: please, don't forget to update the docs if we change this :)

efiop: @pmrowla We also don't makedirs in `dvc run`, but rather raise an error that wdir doesn't exist, so it seems it would be more consistent if we raise and error here too. Same with `dvc add` that shares the same logic with `dvc run`. I wonder why that https://github.com/iterative/dvc/blob/35dd1fda9cfb4b645ae431f4621f2d1853d4fb7c/dvc/stage/utils.py#L27 check wasn't triggered here as well :thinking:

pmrowla: The issue before was that we set `wdir` to `os.cwd()` if the output subdirectory didn't already exist, because we assumed that subdir not existing actually meant it was a remote URL and not a local path

pmrowla: Behavior has been updated here so that we use the correct subdir as `wdir`, and then error out in `utils.check_stage_path` if it doesn't already exist

PR #4363, opened 2020-08-08
[WIP] dependency: fine grained (user cmd filter)

### LINKED ISSUE #3439: Support function specific dependencies
Assume we have 4 stages:
* Stages `0` and `2` depend on the function `foo`
* Stages `1` and `3` depend on the function `bar`

In order to reduce code duplication we want to put the utilities `foo` and `bar` in one place; let them be in `utils.py`. So, `utils.py` becomes a dependency of all four stages.

Now... after some time we realize we have to update something in `bar`. So we edit `utils.py`. This, in turn, invalidate **all** stages. Obviously, a simple solution is to have `utils_0_2.py` which will be a dependency of `0` and `2` and similarly `utils_1_3.py`. But this solution conflicts with good practices of coding.

A possible solution could be to adopt some notation like the one used in `pytest` and specify the function _inside_ `utils.py`. For example, something like:

```
deps:
- md5: 409a438ee6ac70fc1f911631412325ea
  path: utils.py::foo
```

This is a continuation of [this discussion](https://discordapp.com/channels/485586884165107732/563406153334128681/684777057455964378).

casperdcl:
- [x] modify schema
  + [x] `dvc.yaml`
  + [x] `dvc.lock`
- [x] support running (from `dvc.yaml:stages.<stage>.deps.[].<path>.cmd`)
- [ ] add to `dvc.lock`
- [ ] add way to add user cmd filter via CLI API
  + maybe `dvc run -d "utils.py:extract_function.py --name check_db"`
- [ ] add tests
- fixes #3439

Note that this implementation

- uses `PARAM_FILTER = "cmd"`
- passes `path` as a positional argument to the user-defined `cmd`
- computes the md5sum of the user-defined `cmd` output
- only works on local files (not dirs & not remote paths)
- only works for dependencies (not outputs)
- would produce the old behaviour (whole-file hash) if setting `cmd: cat`
- was tested using https://github.com/casperdcl/dvc-udf

schema:

```yaml
# in dvc.yaml
utils.py:
  cmd: python extract_function.py --name check_db

# in dvc.lock
path: utils.py
cmd: python extract_function.py --name check_db
md5: s0m3h45h # computed via `{cmd} {path} | md5sum`
```

testing:

```bash
git clone https://github.com/casperdcl/dvc-udf
cd dvc-udf
pip install -r requirements.txt
dvc repro -f -v out  # assuming dvc installed from this PR
```


dvc/utils/__init__.py
```diff
@@ -58,6 +62,21 @@ def file_md5(fname, tree=None):
         open_func = open
 
     if exists_func(fname):
+        filtered = None
+        if cmd:
```
casperdcl, line 66: not sure if there are aspects of `stage.run.cmd_run` which should be used here

dvc/utils/__init__.py
```diff
@@ -80,6 +99,10 @@ def file_md5(fname, tree=None):
             with open_func(fname, "rb") as fobj:
                 _fobj_md5(fobj, hash_md5, binary, pbar.update)
 
+        if filtered is not None:
+            from dvc.utils.fs import remove
+
+            remove(filtered)
```
casperdcl, line 105: not sure if this is required - maybe automatically handled elsewhere (i.e. entire tmpdir deleted before exit)

dvc/dependency/__init__.py
```diff
@@ -49,9 +49,15 @@
 del SCHEMA[BaseOutput.PARAM_METRIC]
 SCHEMA.update(RepoDependency.REPO_SCHEMA)
 SCHEMA.update(ParamsDependency.PARAM_SCHEMA)
+SCHEMA.update({BaseOutput.PARAM_FILTER: str})
 
 
 def _get(stage, p, info):
+    if isinstance(p, dict):
+        p = list(p.items())
+        assert len(p) == 1
```
casperdcl, line 58: maybe assert not required (should be handled by schema)?

skshetry: Other CLI commands can create/load dependencies, skipping the schema. Good to have an assert.

efiop:
Closing for now, we'll get back after dep/out refactor to properly accomodate this.

PR #4368, opened 2020-08-10
Remove error message when exiting `--stdin` mode

### LINKED ISSUE #4361: check-ignore: exit --stdin silently on empty string [qa]
> Extracted from https://github.com/iterative/dvc/pull/4323#discussion_r467356171

When you use check-ignore in interactive mode, empty strings produce an error and exit. However empty strings are accepted (never match) without `--stdin`:

```console
λ cat .dvcignore
*

λ dvc check-ignore --stdin
 anything
anything
 # <- just hit Enter
Empty string is not a valid pathspec. Please use . instead if you meant to match all paths.

λ dvc check-ignore -dn ''
::
```

So I propose that a) empty strings in --stdin interactive mode simply cause the tool to exit back to the shell prompt and/or b) empty strings sent as `targets` should just be ignored? (exit tool silently)

karajan1001:
fix #4361
1. Remove error message when input a empty string
2. Remove error message when read input from stream
3. Remove error message when interrupt by `CTRL-C`

### DIFF
--- a/dvc/command/check_ignore.py
+++ b/dvc/command/check_ignore.py
@@ -38,12 +38,11 @@ class CmdCheckIgnore(CmdBase):
     def _interactive_mode(self):
         ret = 1
         while True:
-            target = ask("")
+            try:
+                target = ask("")
+            except (KeyboardInterrupt, EOFError):
+                break
             if target == "":
-                logger.info(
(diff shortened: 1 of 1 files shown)


jorgeorpinel:
Nice! I'll include the small docs changes for this in iterative/dvc.org/pull/1675 🙂

PR #4370, opened 2020-08-10
metrics: show: debug info on parsing problems

### LINKED ISSUE #4356: DVC metrics show doesn't find metrics files
## Bug Report

After marking some outputs as metrics using the `-M` option to `dvc run`, `dvc metrics show` tells me there are no metrics files. 

**Output of `dvc version`:**

```console
$ dvc version
DVC version: 1.4.0 (pip)
Platform: Python 3.6.9 on Linux-4.15.0-111-generic-x86_64-with-Ubuntu-18.04-bionic
Repo: dvc, git
```
as well as

```console
$dvc version
DVC version: 1.4.0 (pip)
Platform: Python 3.7.7 on Linux-4.19.132-1.pvops.qubes.x86_64-x86_64-with-fedora-30-Thirty
Workspace directory: ext4 on /dev/xvdb
Repo: dvc, git
```

Steps to reproduce:
```console
$ git clone --depth 1 --branch metrics-show-bug https://github.com/e3bo/random-walks.git
$ cd random-walks
$ dvc metrics show --verbose
2020-08-07 14:29:47,439 DEBUG: fetched: [(3,)]                                  
2020-08-07 14:29:47,459 DEBUG: fetched: [(0,)]
2020-08-07 14:29:47,466 ERROR: failed to show metrics - no metric files in this repository. Use `-m/-M` options for `dvc run` to mark stage outputs as  metrics.
  File "/home/user/venv/lib/python3.7/site-packages/dvc/command/metrics.py", line 49, in run
  File "/home/user/venv/lib/python3.7/site-packages/dvc/repo/metrics/__init__.py", line 8, in show
  File "/home/user/venv/lib/python3.7/site-packages/dvc/repo/__init__.py", line 34, in wrapper
  File "/home/user/venv/lib/python3.7/site-packages/dvc/repo/metrics/show.py", line 105, in show
    raise NoMetricsError()
dvc.exceptions.NoMetricsError: no metric files in this repository. Use `-m/-M` options for `dvc run` to mark stage outputs as  metrics.

2020-08-07 14:29:47,468 DEBUG: Analytics is enabled.
2020-08-07 14:29:47,522 DEBUG: Trying to spawn '['daemon', '-q', 'analytics', '/tmp/tmps0gegceh']'
2020-08-07 14:29:47,524 DEBUG: Spawned '['daemon', '-q', 'analytics', '/tmp/tmps0gegceh']
```

The `dvc.yaml` has the paths to the metrics files, which are present in the repository. It was generated using a script similar to `forecast.sh`. I asked about this problem on discord chat and ivan suggested I open a bug report here.

pared:

Fixes #4356

### DIFF
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -171,11 +171,7 @@ class BadMetricError(DvcException):
 
 
 class NoMetricsError(DvcException):
-    def __init__(self):
-        super().__init__(
-            "no metric files in this repository. Use `-m/-M` options for "
-            "`dvc run` to mark stage outputs as  metrics."
-        )
+    pass
 
 
 class NoPlotsError(DvcException):
--- a/dvc/repo/metrics/show.py
+++ b/dvc/repo/metrics/show.py
@@ -36,7 +36,7 @@ def _collect_metrics(repo, targets, recursive):
     return list(filter(_filter, metrics))
 
 
-def _extract_metrics(metrics):
+def _extract_metrics(metrics, path, rev):
     if isinstance(metrics, (int, float)):
         return metrics
 
@@ -45,9 +45,18 @@ def _extract_metrics(metrics):
 
     ret = {}
     for key, val in metrics.items():
-        m = _extract_metrics(val)
+        m = _extract_metrics(val, path, rev)
         if m:
             ret[key] = m
+        else:
+            logger.debug(
+                "Could not parse '%s' metric from '%s' at '%s' "
+                "due to its unsupported type: '%s'",
+                key,
+                path,
+                rev,
+                type(val).__name__,
+            )
 
     return ret
 
@@ -70,7 +79,7 @@ def _read_metrics(repo, metrics, rev):
             )
             continue
 
-        val = _extract_metrics(val)
+        val = _extract_metrics(val, metric, rev)
         if val:
             res[str(metric)] = val
 
@@ -88,6 +97,7 @@ def show(
     all_commits=False,
 ):
     res = {}
+    metrics_found = False
 
     for rev in repo.brancher(
         revs=revs,
@@ -96,13 +106,27 @@ def show(
         all_commits=all_commits,
     ):
         metrics = _collect_metrics(repo, targets, recursive)
+
+        if not metrics_found and metrics:
+            metrics_found = True
+
         vals = _read_metrics(repo, metrics, rev)
 
         if vals:
             res[rev] = vals
 
     if not res:
-        raise NoMetricsError()
+        if metrics_found:
+            msg = (
+                "Could not parse metric files. Use `-v` option to see more "
+                "details."
+            )
+        else:
+            msg = (
+                "no metric files in this repository. Use `-m/-M` options for "
+                "`dvc run` to mark stage outputs as  metrics."
+            )
+        raise NoMetricsError(msg)
 
     # Hide workspace metrics if they are the same as in the active branch
     try:


dvc/exceptions.py
```diff
@@ -178,6 +178,17 @@ def __init__(self):
         )
 
 
+class NoMetricsParsedError(DvcException):
+    def __init__(self, metrics):
+        lines = "\n".join(
+            f"{rev}: {', '.join(str(p) for p in paths)}"
+            for rev, paths in metrics.items()
```
shcheklein, line 185: output might become a bit long with this I guess. I wonder would be the best to just mention that `-v` is available? Also, considering that `-v` is still needed to understand what is actually going on, I guess.

In that case we could simplify the logic a bit - boolean flag below would be enough instead of a full dict.

dvc/repo/metrics/show.py
```diff
@@ -36,16 +36,20 @@ def _filter(path_info):
...
 
-def _extract_metrics(metrics):
+def _extract_metrics(metrics, path, rev):
     if isinstance(metrics, (int, float)):
         return metrics
 
     if not isinstance(metrics, dict):
+        logger.debug(
+            f"Could not parse metric '{path}' at '{rev}' due to its "
+            f"unsupported type: '{type(metrics).__name__}'"
```
shcheklein, line 46: we should print `key` here, ideally even full xpath within the file.

dvc/exceptions.py
```diff
@@ -178,6 +178,11 @@ def __init__(self):
         )
 
 
+class NoMetricsParsedError(DvcException):
+    def __init__(self,):
+        super().__init__("Coul not parse any metric. Use `-v` to investigate.")
```
shcheklein, line 183: @jorgeorpinel could you please review the message?

jorgeorpinel: How about something like "The metrics/plots file could not be parsed." Not sure we need the "use -v" part. What other info specific to this problem can the user get that way?

jorgeorpinel: Question: is this error output per bad metrics/plots file? Or only when all of the metrics/plots files are bad?

jorgeorpinel: How about something like "The metrics/plots file could not be parsed." for now?

Not sure we need the "use -v" part. What other info specific to this problem can the user get that way?

pared: @jorgeorpinel this error will only be thrown on occasion when there have been metrics found, but none could be parsed, so it should be rather rare. `-v` will allow user to see `debug` messages introduced in this change.

pmrowla: If we keep the message as-is, there's a typo in `Could`

pared: @pmrowla fixed

dvc/exceptions.py
```diff
@@ -178,6 +178,13 @@ def __init__(self):
         )
 
 
+class NoMetricsParsedError(DvcException):
+    def __init__(self,):
+        super().__init__(
+            "Could not parse any metric. Use `-v` to investigate."
```
shcheklein, line 184: `Could not parse metrics files. Use `-v` option to see more details.`

cc @jorgeorpinel - let's get this message right

dvc/repo/metrics/show.py
```diff
@@ -88,21 +93,30 @@ def show(
...
 
         if vals:
             res[rev] = vals
 
     if not res:
-        raise NoMetricsError()
+        if metrics_found:
+            raise NoMetricsParsedError()
+        else:
+            raise NoMetricsError()
```
efiop, line 119: Let's not create a yet another exception type here, since it is really the same error with just a tiny difference. Let's just raise with an explicit message,

dvc/exceptions.py
pared: Fixed the message

dvc/repo/metrics/show.py
```diff
@@ -96,13 +102,27 @@ def show(
...
         vals = _read_metrics(repo, metrics, rev)
 
         if vals:
             res[rev] = vals
 
     if not res:
-        raise NoMetricsError()
+        if metrics_found:
+            msg = (
+                "Could not parse metrics files. Use -v option to see more "
```
efiop, line 117:
```suggestion
                "Could not parse metrics files. Use `-v` option to see more "
```

dvc/repo/metrics/show.py
```diff
@@ -96,13 +102,27 @@ def show(
...
         vals = _read_metrics(repo, metrics, rev)
 
         if vals:
             res[rev] = vals
 
     if not res:
-        raise NoMetricsError()
+        if metrics_found:
+            msg = (
+                "Could not parse metrics files. Use `-v` option to see more "
```
efiop, line 117: for consistency with the origina message
```suggestion
                "Could not parse metric files. Use `-v` option to see more "
```

dvc/repo/metrics/show.py
```diff
@@ -45,9 +45,14 @@ def _extract_metrics(metrics):
...
     ret = {}
     for key, val in metrics.items():
-        m = _extract_metrics(val)
+        m = _extract_metrics(val, path, rev)
         if m:
             ret[key] = m
+        else:
+            logger.debug(
+                f"Could not parse '{key}' metric from '{path}' at '{rev}' "
+                f"due to its unsupported type: '{type(val).__name__}'"
```
efiop, line 54: Let's use a lazy notation, since there might be a lot of metrics present.

dvc/exceptions.py
jorgeorpinel: Hey sorry guys just noticed this mention. I did suggest "The metrics/plots file could not be parsed." before in https://github.com/iterative/dvc/pull/4370#discussion_r468221756 though. I don't think we need to mention -v, that applies to every error message.

Issue #4251, opened 2020-07-21
`get-url` should just `GET` files


skshetry:
If an endpoint does not support `HEAD`, but supports `GET`, `dvc get-url` will fail to work.

Eg: 
```console
$ dvc get-url https://api.travis-ci.com/v3/job/363241860/log.txt -v
2020-07-21 17:23:50,482 ERROR: failed to get 'https://api.travis-ci.com/v3/job/363241860/log.txt' - dependency 'https://api.travis-ci.com/v3/job/363241860/log.txt' does not exist
------------------------------------------------------------
Traceback (most recent call last):
  File "/home/saugat/repos/iterative/dvc/dvc/command/get_url.py", line 17, in run
    Repo.get_url(self.args.url, out=self.args.out)
  File "/home/saugat/repos/iterative/dvc/dvc/repo/get_url.py", line 19, in get_url
    dep.save()
  File "/home/saugat/repos/iterative/dvc/dvc/output/base.py", line 253, in save
    raise self.DoesNotExistError(self)
dvc.dependency.base.DependencyDoesNotExistError: dependency 'https://api.travis-ci.com/v3/job/363241860/log.txt' does not exist
------------------------------------------------------------
```

Just commenting `dep.save()` makes it work, as it does a HEAD request first which fails.

PR #4372, opened 2020-08-10
get-url: allow weak etag

pared:

Related to #4131

### DIFF
--- a/dvc/tree/http.py
+++ b/dvc/tree/http.py
@@ -136,12 +136,6 @@ class HTTPTree(BaseTree):  # pylint:disable=abstract-method
                 "Content-MD5 header for '{url}'".format(url=url)
             )
 
-        if etag.startswith("W/"):
-            raise DvcException(
-                "Weak ETags are not supported."
-                " (Etag: '{etag}', URL: '{url}')".format(etag=etag, url=url)
-            )
-
         return etag
 
     def _download(self, from_info, to_file, name=None, no_progress_bar=False):
--- a/dvc/tree/webdav.py
+++ b/dvc/tree/webdav.py
@@ -142,14 +142,6 @@ class WebDAVTree(BaseTree):  # pylint:disable=abstract-method
                 "Content-MD5 header for '{url}'".format(url=path_info.url)
             )
 
-        if etag.startswith("W/"):
-            raise DvcException(
-                "Weak ETags are not supported."
-                " (Etag: '{etag}', URL: '{url}')".format(
-                    etag=etag, url=path_info.url
-                )
-            )
-
         return etag
 
     # Checks whether path points to directory


dvc/tree/http.py
```diff
@@ -136,12 +136,6 @@ def get_file_hash(self, path_info):
                 "Content-MD5 header for '{url}'".format(url=url)
             )
 
-        if etag.startswith("W/"):
-            raise DvcException(
-                "Weak ETags are not supported."
-                " (Etag: '{etag}', URL: '{url}')".format(etag=etag, url=url)
-            )
-
```
skshetry, line 144: @pared, what about imports?

pared: Yeah, thats right, I changed the approach.

dvc/repo/get_url.py
```diff
@@ -16,6 +16,5 @@ def get_url(url, out=None):
 
     (dep,) = dependency.loads_from(None, [url])
     (out,) = output.loads_from(None, [out], use_cache=False)
-    dep.save()
```
pared, line 19: It seems to me that upon `get-url` there is no point in saving the dependency. We do not have any operations related to objects that are obtained using this method. Saving dependency makes sense in case of `import`, when later we reuse the information, for example to update the asset.

skshetry: Also fixes https://github.com/iterative/dvc/issues/4251

efiop: This makes us handle missing dependencies in a unified way though. Without it the user will get something random from the internals.

pared: @efiop, you mean, for example what happens upon downloading non-existient asset? Yeah that's true, but isn't `get-url` convenience method so that user does not have to work with other tools like, `wget`, `curl`? Treating it as normal dependency can limit its functionality, like in #4131. If we want to unify the handling of errors, maybe we should check if dependency exist and throw from inside the `get_url` method?

efiop: @pared That's would a yet another place for duplicating the logic. Currently, we have unified save() and we have non-unified tree exceptions, that might vary from FileNotFound for local to something completely odd in gdrive + some unexpected behaviours. In any case, this is clearly out of scope for this weak-etag PR.

pared: @efiop got it, though I don't think it makes sense to keep `dep.save()` just to unify its behavior on error, because main point of `save` is to save information about the dependency. Maybe we should split functionality, so that we could `check` the dependency without actually saving it?

pared: @efiop I pushed another way how we can solve it, thought in this version, #4251 is no longer fixed cc @skshetry 
I think that in this case we need to decide whether we prefer to have the error handling unified, or we treated `get-url` as convenience method for the user, and can live with some ambiguous errors. Considering there is no conceptual link between `get-url` results and other commands in the project, I would lean towards the latter.

efiop: @pared I agree, we need to unify tree exceptions and then we might not need to do anything else at all, as everything will just raise FileNotFoundError .

skshetry: For the `get` and `get-url`, is there even a need for checks? Should not `.download()` take responsibility (`download()` can depend on `tree` though)?

efiop: Yes, potentially unified tree exceptions is all we need, but I'm not 100% sure in this particular case. The fact right now is that `download()` will raise random exceptions, which are confusing. So using `save()` for now just provides a familiar import-like and `-d` like errors, which is good enough for now.

efiop: @pared I wonder if we can simply allow weak etags for import-url and `run -d` as well. We clearly can't use weak etags for caching, but it is good enough for dependencies. Or am I missing something?

efiop: @pared So if we can use weak etags, could we just remove that check in this PR? The `check()` changes work, but they won't work for `import-url`, so we will get different errors, which is not very nice.

pared: Rolled back to allowing weak etag, #4251 is no longer fixed by this issue. cc @skshetry

jorgeorpinel:
I unmarked the docs check box since "strong etag" is mentioned in:

- https://dvc.org/doc/user-guide/dvc-files-and-directories
- https://dvc.org/doc/user-guide/external-dependencies
- https://dvc.org/doc/command-reference/import-url

Seems like these (and possibly other) docs need an update per this?

pared:
@jorgeorpinel that is a good point, sorry for omitting that, I will create issue and assign myself.

jorgeorpinel:
Np, just checking 🙂

Issue #4346, opened 2020-08-06
experiments show: allow filtering/sorting specific metrics/params


pmrowla:
From #4345

> By default, if the console isn't wide enough to display the full table, params columns will be truncated first, giving priority to display the full experiment SHA and resulting metrics. In the future we should probably have options to only show certain metrics/params columns so that we don't need to truncate anything.

TODO

* [x] include/exclude columns
* [ ] sorting on specific columns

PR #4375, opened 2020-08-11
experiments: add options to include/exclude columns in `exp show`

pmrowla:

Related to #4346

* Adds `--include-metrics/--exclude-metrics` and `--include-params/--exclude-params` options to `dvc exp show`.
* Options take comma-separated list of metrics/params columns to include or exclude from the output table.
* If `--include...` is used, only columns specified in the list will be displayed. If `--exclude...` is used, all columns except those specified in the list will be displayed.
* Matches nested metrics/params, so `--include-params featurize` will match `featurize.max_features`, `featurize.ngrams`, etc
* Params and metrics columns lists are handled separately, so `dvc exp show --include-params=foo` will display all metrics columns plus the single param column `foo`.

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -2,6 +2,7 @@ import argparse
 import io
 import logging
 from collections import OrderedDict
+from itertools import groupby
 from typing import Iterable, Optional
 
 from dvc.command.base import CmdBase, append_doc_link, fix_subparsers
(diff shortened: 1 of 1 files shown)


pmrowla:
```

example-get-started git:cli-params  py:dvc ❯ dvc exp show --include-params featurize.max_features,featurize.ngrams --no-pager                              ⏎
┏━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
┃ Experiment  ┃     auc ┃ featurize.max_features ┃ featurize.ngrams ┃
┡━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
│ workspace   │ 0.61314 │ 1500                   │ 2                │
│ acd59eb     │ 0.61314 │ 1500                   │ 2                │
│ ├── be1e801 │  0.5125 │ 500                    │ 3                │
│ ├── 1dd825b │ 0.51676 │ 500                    │ 2                │
│ └── eb12d21 │ 0.54175 │ 500                    │ 1                │
└─────────────┴─────────┴────────────────────────┴──────────────────┘
```

pmrowla:
marking this WIP, will add support for filtering on higher level (i.e. `--include-params=featurize` would match `featurize.max_features`, `featurize.ngrams`, etc)

dvc/command/experiments.py
```diff
@@ -2,14 +2,57 @@
...
+                f" --exclude-{label}"
+            )
+
+    names = {tuple(name.split(".")) for name in names}
+
+    def _filter(filters, update_func):
+        filters = {tuple(name.split(".")) for name in filters}
+        for length, groups in groupby(filters, len):
+            for group in groups:
+                matches = [name for name in names if name[:length] == group]
```
pared, line 36: If we are going to the extent of matching on high level, maybe we could consider support for expressions?
Just an idea, and not for this change, I am pretty sure that parsing from the command line will be a headache if we were to go this way.

pmrowla: Yeah I was thinking about this too. Support for more complex expressions seems like a nice to have we could consider adding down the line, but would be overkill for now.

PR #4381, opened 2020-08-12
RepoTree: add support for subrepo traversal

skshetry:
Right now, no actual "functionality" is surfaced as
`subrepos` is always set to False by default for now.

Subrepos are collected dynamically and trie is created to track those repos inside `RepoTree`, and this does not incur any cost
as `.dvcignore` is shared among subrepos and `RepoTree` just utilizes them.

#### Possible alternative design
1. Create "MotherOfTree" that will be prefix-based and can switch between those just based on prefix. So, some separate tree class that works on top of RepoTree or  GitTree for submodules.
    cons: Hard to traverse, especially heirarchial traversal required for dvc repos, etc.

### DIFF
--- a/dvc/repo/ls.py
+++ b/dvc/repo/ls.py
@@ -76,9 +76,9 @@ def _ls(repo, path_info, recursive=None, dvc_only=False):
             if not recursive:
                 for dname in dirs:
                     info = PathInfo(root) / dname
-                    if not dvc_only or (
-                        tree.dvctree and tree.dvctree.exists(info)
-                    ):
+                    # pylint:disable=protected-access
+                    _, dvctree = tree._get_tree_pairs(info)  # noqa
+                    if not dvc_only or (dvctree and dvctree.exists(info)):
                         dvc = tree.isdvc(info)
                         path = str(info.relative_to(path_info))
                         ret[path] = {
--- a/dvc/repo/tree.py
+++ b/dvc/repo/tree.py
@@ -1,5 +1,11 @@
 import logging
 import os
+import threading
+from itertools import takewhile
+from typing import Optional, Tuple
+
+from funcy import wrap_with
+from pygtrie import StringTrie
 
 from dvc.dvcfile import is_valid_filename
 from dvc.exceptions import OutputNotFoundError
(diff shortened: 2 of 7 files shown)


dvc/repo/tree.py
```diff
@@ -248,78 +254,148 @@ class RepoTree(BaseTree):  # pylint:disable=abstract-method
...
+        repo_path = os.path.join(dir_path, Repo.DVC_DIR)
+        return self._main_repo.tree.isdir(repo_path, use_dvcignore=False)
+
+    def _get_tree_pairs(self, path) -> Tuple["BaseTree", Optional["DvcTree"]]:
+        path = os.path.abspath(path)
+        repo = self._get_repo(path)
+        if not repo:
+            # path could be outside of the repo, so we just send them the main
+            # tree instead
+            return self._main_repo.tree, self._dvctrees.get(self.root_dir)
```
skshetry, line 321: This is just temporary, will remove during "ExternalRepo" refactoring, as we will not support get/import for external outputs than the repo.

dvc/repo/ls.py
```diff
@@ -76,9 +76,9 @@ def onerror(exc):
             if not recursive:
                 for dname in dirs:
                     info = PathInfo(root) / dname
-                    if not dvc_only or (
-                        tree.dvctree and tree.dvctree.exists(info)
-                    ):
+                    # pylint:disable=protected-access
+                    _, dvctree = tree._get_tree_pairs(info)  # noqa
```
skshetry, line 80: Temporary for now, maybe we won't need it after `tree.stat()` impl.

dvc/repo/tree.py
```diff
@@ -248,78 +254,148 @@ class RepoTree(BaseTree):  # pylint:disable=abstract-method
     scheme = "local"
     PARAM_CHECKSUM = "md5"
 
-    def __init__(self, repo, **kwargs):
+    def __init__(self, repo, subrepos=False, repo_factory=None, **kwargs):
```
skshetry, line 257: `repo_factory` will be needed for `ExternalRepo` as we need to set `cache` and other configurations.

skshetry:
@efiop, I cannot split this into smaller PR, though more than half of it is tests.

dvc/repo/ls.py
efiop: So maybe let's start with `tree.stat()`?

dvc/repo/tree.py
efiop: Maybe let's start with that first? The current way is confusing, I'm afraid that it will be hard to clean it up once again, as the complexity will be growing more and more.

skshetry: It'd be helpful if you could point out what you find confusing.

dvc/repo/ls.py
skshetry: This is a pre-requisite for `tree.stat()`, as I would need to change things again to account for subrepos, and it's not that intrusive for now.

dvc/repo/tree.py
skshetry: > Maybe let's start with that first?

ExternalRepo will just use `repo_factory`. This is a requirement for `ExternalRepo`, so there will be no need for cleanups.

tests/unit/repo/test_repo_tree.py
```diff
@@ -177,3 +178,181 @@ def test_isdvc(tmp_dir, dvc):
...
+    with mock.patch.object(
+        tree,
+        "_get_repo",
+        side_effect=assert_tree_belongs_to_repo(subrepo1.dvc),
+    ):
+        assert tree.exists(subrepo1 / "foo") is True
+        assert tree.exists(subrepo1 / "bar") is False
+
+        assert tree.isfile(subrepo1 / "foo") is True
+        assert tree.isfile(subrepo1 / "foo") is True
```
pared, line 231: repeated line

tests/unit/repo/test_repo_tree.py
```diff
@@ -177,3 +178,181 @@ def test_isdvc(tmp_dir, dvc):
...
+    with mock.patch.object(
+        tree,
+        "_get_repo",
+        side_effect=assert_tree_belongs_to_repo(subrepo2.dvc),
+    ):
+        assert tree.exists(subrepo2 / "lorem") is True
+        assert tree.exists(subrepo2 / "ipsum") is False
+
+        assert tree.isfile(subrepo2 / "lorem") is True
+        assert tree.isfile(subrepo2 / "lorem") is True
```
pared, line 248: repeated line

dvc/repo/tree.py
```diff
@@ -248,78 +254,148 @@ class RepoTree(BaseTree):  # pylint:disable=abstract-method
...
+    def _is_dvc_repo(self, dir_path):
+        if not self._traverse_subrepos:
+            return False
+
+        from dvc.repo import Repo
+
+        repo_path = os.path.join(dir_path, Repo.DVC_DIR)
+        return self._main_repo.tree.isdir(repo_path, use_dvcignore=False)
+
+    def _get_tree_pairs(self, path) -> Tuple["BaseTree", Optional["DvcTree"]]:
```
pared, line 315: `_get_tree_pair`?

dvc/repo/tree.py
skshetry: I have a new alternate design, let's see.

skshetry:
Tests are failing due to #4229, as now we create a ".dvcignore" on every `dvc init`. So these 3 tests should add ".dvcignore" in it's expectations.

PR #4382, opened 2020-08-12
ui: '-j' option now mentions docs

### LINKED ISSUE #1310: Improve `-j` help messages
For example, `dvc gc -j N`. There are also other DVC commands that have this option.

`-j` specifies parallelism level for the command. 

It should specify that default number depends on a remote type and that to get defaults one should refer to docs.

nik123:
Fixes #1310 

I don't think that documentation should be updated.

### DIFF
--- a/dvc/command/data_sync.py
+++ b/dvc/command/data_sync.py
@@ -97,7 +97,8 @@ def shared_parent_parser():
         "-j",
         "--jobs",
         type=int,
-        help="Number of jobs to run simultaneously.",
+        help="Number of jobs to run simultaneously. "
+        "Refer to docs to get default value.",
         metavar="<number>",
     )
     parent_parser.add_argument(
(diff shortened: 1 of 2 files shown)


dvc/command/data_sync.py
```diff
@@ -97,7 +97,8 @@ def shared_parent_parser():
         "-j",
         "--jobs",
         type=int,
-        help="Number of jobs to run simultaneously.",
+        help="Number of jobs to run simultaneously. "
+        "Refer to docs to get default value.",
```
efiop, line 101: Maybe let's just copypaste from docs? E.g.
```
The default value is 4 * cpu_count(). For SSH remotes, the default is 4. Using more jobs may improve the overall transfer speed.
```

nik123: I've noticed that most help messages are short (2 sentences max) so I decided to use "refer to docs...". However, I have no strong opinion in that regard. If it's fine to have long messages I can copy text from cmd ref.

efiop: Yes, there is no limit for the help messages. We could totally copypaste those two sentences :) Thank you! :pray:

nik123: I've updated message

PR #4399, opened 2020-08-17
upgrade to isort5

skshetry:
* Imports are no longer moved to the top.
* `seed-isort-config` is no longer needed.
* Has a `black`-compatible profile/config.
* Also `isort`s the imports anywhere in the code
  (not just at the top).

### DIFF
--- a/dvc/repo/plots/data.py
+++ b/dvc/repo/plots/data.py
@@ -101,7 +101,7 @@ def _apply_path(data, path=None, **kwargs):
 def _lists(dictionary):
     for _, value in dictionary.items():
         if isinstance(value, dict):
-            yield from (_lists(value))
+            yield from _lists(value)
         elif isinstance(value, list):
             yield value
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -15,10 +15,6 @@ show_source=true
 count=true
 
 [isort]
-include_trailing_comma=true
+profile=black
 known_first_party=dvc,tests
-known_third_party=PyInstaller,RangeHTTPServer,boto3,colorama,configobj,distro,dpath,flaky,flufl,funcy,git,grandalf,mock,moto,nanotime,networkx,packaging,pathspec,pygtrie,pylint,pytest,requests,ruamel,setuptools,shortuuid,shtab,toml,tqdm,voluptuous,yaml,zc
 line_length=79
-force_grid_wrap=0
-use_parentheses=True
-multi_line_output=3
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,15 +4,11 @@ repos:
         language_version: python3
     repo: https://github.com/ambv/black
     rev: 19.10b0
-  - repo: https://github.com/asottile/seed-isort-config
-    rev: v2.1.1
-    hooks:
-      - id: seed-isort-config
   - hooks:
       - id: isort
         language_version: python3
     repo: https://github.com/timothycrosley/isort
-    rev: 4.3.21
+    rev: 5.4.2
   - hooks:
       - id: flake8
         language_version: python3
--- a/.restyled.yaml
+++ b/.restyled.yaml
@@ -10,7 +10,7 @@ restylers:
     interpreters:
     - python
   - name: isort
-    image: restyled/restyler-isort:v4.3.21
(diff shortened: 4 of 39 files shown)


dvc/repo/plots/data.py
```diff
@@ -101,7 +101,7 @@ def _apply_path(data, path=None, **kwargs):
 def _lists(dictionary):
     for _, value in dictionary.items():
         if isinstance(value, dict):
-            yield from (_lists(value))
+            yield from _lists(value)
```
skshetry, line 104: flake8 complained here.

dvc/repo/plots/data.py
```diff
@@ -114,10 +114,12 @@ def _find_data(data, fields=None, **kwargs):
...
-    for l in _lists(data):
-        if all(isinstance(dp, dict) for dp in l):
-            if set(first(l).keys()) & fields == fields:
-                return l
+    for lst in _lists(data):
+        if (
+            all(isinstance(dp, dict) for dp in lst)
+            and set(first(lst).keys()) & fields == fields
+        ):
+            return lst
```
skshetry, line 122: I also upgrade flake8, so it complained here which I refactored.

skshetry:
restyled-isort does not have isort5 support. will try adding later.

setup.cfg
```diff
@@ -15,10 +15,6 @@ show_source=true
 count=true
 
 [isort]
-include_trailing_comma=true
+profile=black
 known_first_party=dvc,tests
-known_third_party=PyInstaller,RangeHTTPServer,boto3,colorama,configobj,distro,dpath,flaky,flufl,funcy,git,grandalf,mock,moto,nanotime,networkx,packaging,pathspec,pygtrie,pylint,pytest,requests,ruamel,setuptools,shortuuid,shtab,toml,tqdm,voluptuous,yaml,zc
 line_length=79
```
skshetry, line 20: @efiop @pared @pmrowla, what do you guys say of increasing line length limit to black's default (i.e. 88)?

pmrowla:
failing CI test is specific to experiments and shouldn't block this PR (see #4401)

skshetry:
> failing CI test is specific to experiments and shouldn't block this PR (see #4401)

I'll rerun later. Thanks. Currently it's blocked by lack of support from restyled.

Keeping this as draft, till restyled-isort supports v5, see https://github.com/restyled-io/restylers/pull/86.

setup.cfg
pmrowla: Personally, I think 79 is fine. I'm not opposed to using 88, but I would prefer we don't go much higher than that in the future.

IMO Python projects that use 100+ tend to get hard to read, but I get that this is more of a personal taste issue than anything else.

pared: I have no strong opinion on that, only thing that bugs me is that this decision will require some refactor, as black tends to not remove `"` when its not needed anymore, so we will probably have a lot of lines like
`"This is " "the comment"`

skshetry: @pared, string handling has been significantly improved on black (https://github.com/psf/black/pull/1132) and should solve that issue. Though that'd be useful only if there were a new release. :(

skshetry:
PR has been merged and the new `restyled-isort` image is available. :tada:

setup.cfg
skshetry: Okay then, if no one has *strong* opinion, I'll be increasing the limit to 88.

tests/remotes/__init__.py
```diff
@@ -35,6 +35,14 @@
...
 
 
+TEST_REMOTE = "upstream"
+TEST_CONFIG = {
+    "cache": {},
+    "core": {"remote": TEST_REMOTE},
+    "remote": {TEST_REMOTE: {"url": ""}},
+}
+
+
```
skshetry, line 45:
```suggestion
```

Rebase duplicated this, removing

setup.cfg
efiop: Oops, missed this one. I would not increase it to 88. 80 is a good standard width that is used in many projects and is on the border of being revieable in github split screen view for PRs. Plus, if we change it now, we'll introduce inconsistency with the rest of the code base. So I'm not really seeing good reasons to go to 88, it won't make it easier to navigate.

PR #4402, opened 2020-08-17
experiments: abstract executor output collection using trees

pmrowla:

* Executors return a list of output paths via `executor.collect_output()`
* Collected paths should then be downloaded to experiments workspace with `local_tree.download(...)`

still need to investigate why stages aren't pickable/returnable across multiprocessing calls, but any changes related to that will probably come in a future PR

### DIFF
--- a/dvc/repo/experiments/__init__.py
+++ b/dvc/repo/experiments/__init__.py
@@ -450,7 +450,7 @@ class Experiments:
         for from_info in executor.collect_output():
             from_infos.append(from_info)
             fname = from_info.relative_to(src_tree.path_info)
-            names.append(fname)
+            names.append(str(fname))
             to_info = dest_tree.path_info / fname
             to_infos.append(dest_tree.path_info / fname)
             logger.debug(f"from '{from_info}' to '{to_info}'")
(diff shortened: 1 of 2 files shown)


dvc/repo/experiments/__init__.py
```diff
@@ -421,12 +434,42 @@ def _reproduce(self, executors: dict, jobs: Optional[int] = 1) -> dict:
...
+        func = partial(
+            _log_exceptions(dest_tree.download, "download"),
+            dir_mode=dest_tree.dir_mode,
+            file_mode=dest_tree.file_mode,
+        )
+        with Tqdm(total=total, unit="file", desc="Downloading") as pbar:
+            func = pbar.wrap_fn(func)
+            # TODO: parallelize this, currently --jobs for repro applies to
+            # number of repro executors not download threads
+            with ThreadPoolExecutor(max_workers=1) as dl_executor:
```
pmrowla, line 468: Will need to figure out better usage for this in the future, I think using `--jobs` for repro executors will probably end up changing, so at that point we can use it for download jobs the same way as w/other commands

pared: Maybe we should set some sensible default? Like max(2, cpu_count()/2)?

pared (approved):

efiop (approved):

PR #4403, opened 2020-08-17
update flake8 to latest version

skshetry:
There are some false positives in the older versions, particularly regarding type annotations that are blocking my other PR.

Also, there were some false negatives that seem to have been fixed, which has been adjusted.


skshetry:
This is not good, CI takes >35 mins to run. Should investigate why that is the case.

Issue #4401, opened 2020-08-17
experiments: test_new_simple is flaky


pmrowla:
Need to investigate flaky test, probably hitting some gitpython related race condition again

context:
https://travis-ci.com/github/iterative/dvc/jobs/372602915
https://travis-ci.com/github/iterative/dvc/jobs/372071149

skshetry:
Another failure on a [different test](https://travis-ci.com/github/iterative/dvc/jobs/373237624):
```
=================================== FAILURES ===================================
_____________________________ test_show_experiment _____________________________
[gw2] linux -- Python 3.7.1 /home/travis/virtualenv/python3.7.1/bin/python
tmp_dir = PosixTmpDir('/tmp/pytest-of-travis/pytest-0/popen-gw2/test_show_experiment0')
scm = Git: '/tmp/pytest-of-travis/pytest-0/popen-gw2/test_show_experiment0/.git'
dvc = Repo: '/tmp/pytest-of-travis/pytest-0/popen-gw2/test_show_experiment0'
    def test_show_experiment(tmp_dir, scm, dvc):
        tmp_dir.gen("copy.py", COPY_SCRIPT)
        tmp_dir.gen("params.yaml", "foo: 1")
        dvc.run(
            cmd="python copy.py params.yaml metrics.yaml",
            metrics_no_cache=["metrics.yaml"],
            params=["foo"],
            name="foo",
        )
        scm.add(["copy.py", "params.yaml", "metrics.yaml", "dvc.yaml", "dvc.lock"])
        scm.commit("baseline")
        baseline_rev = scm.get_rev()
        timestamp = datetime.fromtimestamp(
            scm.repo.rev_parse(baseline_rev).committed_date
        )
    
        dvc.reproduce(PIPELINE_FILE, experiment=True, params=["foo=2"])
        results = dvc.experiments.show()
    
        expected_baseline = {
            "metrics": {"metrics.yaml": {"foo": 1}},
            "params": {"params.yaml": {"foo": 1}},
            "queued": False,
            "timestamp": timestamp,
        }
        expected_params = {"foo": 2}
    
        assert set(results.keys()) == {"workspace", baseline_rev}
        experiments = results[baseline_rev]
        assert len(experiments) == 2
        for rev, exp in experiments.items():
            if rev == "baseline":
                assert exp == expected_baseline
            else:
                assert exp["metrics"]["metrics.yaml"] == expected_params
                assert exp["params"]["params.yaml"] == expected_params
[...]

PR #4415, opened 2020-08-18
Move to yaml1.2

### LINKED ISSUE #4281: LockFile class dump/load uses different libraries

skshetry:
This PR:
* gets rid of pyyaml
* uses ruamel everywhere
* tries to unite toml and yaml

https://github.com/iterative/dvc.org/issues/1704

Closes #4281

### DIFF
--- a/dvc/dependency/param.py
+++ b/dvc/dependency/param.py
@@ -2,12 +2,11 @@ import os
 from collections import defaultdict
 
 import dpath.util
-import toml
-import yaml
 from voluptuous import Any
 
 from dvc.dependency.local import LocalDependency
 from dvc.exceptions import DvcException
+from dvc.utils.serialize import PARSERS, ParseError
 
 
 class MissingParamsError(DvcException):
@@ -22,8 +21,6 @@ class ParamsDependency(LocalDependency):
     PARAM_PARAMS = "params"
     PARAM_SCHEMA = {PARAM_PARAMS: Any(dict, list, None)}
     DEFAULT_PARAMS_FILE = "params.yaml"
-    PARAMS_FILE_LOADERS = defaultdict(lambda: yaml.safe_load)
-    PARAMS_FILE_LOADERS.update({".toml": toml.load})
 
     def __init__(self, stage, path, params):
         info = {}
@@ -88,12 +85,12 @@ class ParamsDependency(LocalDependency):
         if not self.exists:
             return {}
 
+        suffix = self.path_info.suffix.lower()
+        parser = PARSERS[suffix]
         with self.repo.tree.open(self.path_info, "r") as fobj:
             try:
-                config = self.PARAMS_FILE_LOADERS[
-                    self.path_info.suffix.lower()
-                ](fobj)
-            except (yaml.YAMLError, toml.TomlDecodeError) as exc:
+                config = parser(fobj.read(), self.path_info)
+            except ParseError as exc:
                 raise BadParamFileError(
                     f"Unable to read parameters from '{self}'"
                 ) from exc
--- a/dvc/repo/metrics/show.py
+++ b/dvc/repo/metrics/show.py
@@ -1,12 +1,11 @@
 import logging
 import os
 
-import yaml
-
 from dvc.exceptions import NoMetricsError
 from dvc.path_info import PathInfo
 from dvc.repo import locked
 from dvc.repo.tree import RepoTree
+from dvc.utils.serialize import YAMLFileCorruptedError, parse_yaml
 
 logger = logging.getLogger(__name__)
 
@@ -72,8 +71,8 @@ def _read_metrics(repo, metrics, rev):
         try:
             with tree.open(metric, "r") as fobj:
                 # NOTE this also supports JSON
-                val = yaml.safe_load(fobj)
-        except (FileNotFoundError, yaml.YAMLError):
+                val = parse_yaml(fobj.read(), metric)
+        except (FileNotFoundError, YAMLFileCorruptedError):
             logger.debug(
                 "failed to read '%s' on '%s'", metric, rev, exc_info=True
             )
--- a/dvc/repo/plots/data.py
+++ b/dvc/repo/plots/data.py
@@ -5,9 +5,8 @@ import os
 from collections import OrderedDict
 from copy import copy
 
-import yaml
 from funcy import first
-from yaml import SafeLoader
+from ruamel.yaml import YAML
 
 from dvc.exceptions import DvcException
 
@@ -208,18 +207,7 @@ class CSVPlotData(PlotData):
 
 class YAMLPlotData(PlotData):
     def raw(self, **kwargs):
-        class OrderedLoader(SafeLoader):
-            pass
-
-        def construct_mapping(loader, node):
-            loader.flatten_mapping(node)
-            return OrderedDict(loader.construct_pairs(node))
-
-        OrderedLoader.add_constructor(
-            yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG, construct_mapping
-        )
-
-        return yaml.load(self.content, OrderedLoader)
+        return YAML().load(self.content)
 
     def _processors(self):
         parent_processors = super()._processors()
--- a/dvc/scm/git.py
+++ b/dvc/scm/git.py
@@ -5,7 +5,6 @@ import os
 import shlex
 from functools import partial
 
-import yaml
 from funcy import cached_property
 from pathspec.patterns import GitWildMatchPattern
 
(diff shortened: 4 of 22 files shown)


dvc/dependency/param.py
```diff
@@ -88,12 +85,12 @@ def read_params(self):
...
+        suffix = self.path_info.suffix.lower()
+        parser = PARSERS[suffix]
         with self.repo.tree.open(self.path_info, "r") as fobj:
             try:
-                config = self.PARAMS_FILE_LOADERS[
-                    self.path_info.suffix.lower()
-                ](fobj)
-            except (yaml.YAMLError, toml.TomlDecodeError) as exc:
+                config = parser(fobj.read(), self.path_info)
+            except ParseError as exc:
```
skshetry, line 93: This is not perfect as it still needs a `tree` to open it correctly and could be error-prone (depending on how/what encoding it's opened, _et cetera_). So, in the next PR, I'm thinking of changing the serialization APIs to make it optionally aware of the trees.
```
def load_yaml(path, tree=None):
   pass
```

This way, I can make `parse_yaml` somewhat _internal_ and so that we could get away without using it for the most problems.

dvc/repo/plots/data.py
```diff
@@ -208,18 +207,7 @@ def raw(self, header=True, **kwargs):  # pylint: disable=arguments-differ
...
-        def construct_mapping(loader, node):
-            loader.flatten_mapping(node)
-            return OrderedDict(loader.construct_pairs(node))
-
-        OrderedLoader.add_constructor(
-            yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG, construct_mapping
-        )
-
-        return yaml.load(self.content, OrderedLoader)
+        return YAML().load(self.content)
```
skshetry, line 210: @pared, this is a round-trip load, which returns OrderedDict, so it works the _same_  way.

skshetry: Might be slower though. :(

dvc/utils/__init__.py
```diff
@@ -237,18 +235,6 @@ def current_timestamp():
...
-
-
-def to_yaml_string(data):
-    stream = io.StringIO()
-    yaml = YAML()
-    yaml.default_flow_style = False
-    yaml.dump(data, stream)
-    return stream.getvalue()
-
-
```
skshetry, line 251: These weren't being used, seems leftover from `tags`. Moved these to `_yaml` and slightly changed the UI.

dvc/scm/git.py
```diff
@@ -349,8 +345,7 @@ def install(self, use_pre_commit_tool=False):
             return
 
         config["repos"].append(entry)
-        with open(config_path, "w+") as fobj:
-            yaml.dump(config, fobj)
+        dump_yaml(config_path, config)
```
skshetry, line 348: Thinking of introducing a different API here to reduce boilerplate of create + load + dump (and, it's error-prone to work with parsers directly).
```python
with modify_yaml(config_path, tree=repo.tree) as config:
    config["repos"] = config.get("repos", [])
    config["repos"].append(entry)
```

Should also make `experiments` and `dvcfile` dumps nicer.

skshetry:
Tests are failing because of https://github.com/iterative/dvc/issues/4401

dvc/dependency/param.py
```diff
@@ -2,12 +2,11 @@
 from collections import defaultdict
 
 import dpath.util
-import toml
```
skshetry, line 5: @pmrowla, [`tomlkit`](https://github.com/sdispater/tomlkit) preserves ordering and comments.

dvc/repo/metrics/show.py
```diff
@@ -72,8 +71,8 @@ def _read_metrics(repo, metrics, rev):
         try:
             with tree.open(metric, "r") as fobj:
                 # NOTE this also supports JSON
-                val = yaml.safe_load(fobj)
-        except (FileNotFoundError, yaml.YAMLError):
+                val = parse_yaml(fobj.read(), metric)
+        except (FileNotFoundError, YAMLFileCorruptedError):
```
skshetry, line 75: `FileNotFoundError`, is that even possible? Anyway, keeping as it is for now.

pmrowla (approved):

pared (approved):

dvc/utils/serialize/__init__.py
```diff
@@ -1,2 +1,8 @@
+from collections import defaultdict
+
+from ._common import *  # noqa, pylint: disable=wildcard-import
 from ._toml import *  # noqa, pylint: disable=wildcard-import
 from ._yaml import *  # noqa, pylint: disable=wildcard-import
+
+PARSERS = defaultdict(lambda: parse_yaml)  # noqa: F405
+PARSERS.update({".toml": parse_toml})  # noqa: F405
```
skshetry, line 8: Also, it'd be good to have JSON utils after the upcoming PR with trees are done. Maybe even move to `ujson` as it has now wheels on all platforms.

```
In [1]: %timeit json.loads('{"ja": "son"}')
2.32 µs ± 35 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [2]: %timeit YAML(typ='safe').load('{"ja": "son"}')
180 µs ± 1.28 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [3]: %timeit ujson.loads('{"ja": "son"}')
372 ns ± 13.4 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

dvc/dependency/param.py
skshetry: Done in #4425

PR #4416, opened 2020-08-18
Make repo lock reentrable and provide a context manager

Suor:
Can make `locked()` work both ways by:
```python
def locked(f):
    if isinstance(f, Repo):
        return lock_repo(f)

    ...
```

Or provide `Repo.locked()` context manager sharing implementation - less things to import.


efiop:
looks like rwlock is broken, need to take a closer look...

Suor:
@efiop should work now

efiop (approved):
Thank you so much for fixing it! :pray:

PR #4417, opened 2020-08-18
Speed up dvc files load

Suor:
See `._load()` implementation. Alternative is not using 3 io calls, but
simply use open and catch exceptions.

Reading is slow for old git revisions.

### DIFF
--- a/dvc/tree/git.py
+++ b/dvc/tree/git.py
@@ -1,8 +1,8 @@
 import errno
 import io
 import os
+from functools import lru_cache
 
-import methodtools
 from funcy import cached_property
 
 from dvc.exceptions import DvcException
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -81,6 +81,7 @@ def run(self):
     "dpath>=2.0.1,<3",
     "shtab>=1.3.0,<2",
     "rich>=3.0.5",
+    "methodtools>=0.4.2",
```
efiop, line 84: Doesn't seem like there is a conda package for it :slightly_frowning_face:  Need to use something else. Surprized you don't have something like this in funcy :wink:

dvc/tree/git.py
```diff
@@ -114,6 +115,7 @@ def _is_tree_and_contains(obj, path):
                 return True
         return False
 
+    @methodtools.lru_cache(maxsize=1)
```
efiop, line 118: This is neat. Indeed, we don't overwrite the history, so this is pretty safe.

setup.py
efiop: Why not just functool's lru_cache, btw? Was there a particular reason not to use it @Suor ?

dvc/tree/git.py
```diff
@@ -114,6 +115,7 @@ def _is_tree_and_contains(obj, path):
...
     def _git_object_by_path(self, path):
         import git
```
efiop, line 121: For the record: might want to check `tree = self.git.tree(self.rev)` logic down below, as there doesn't seem to be a reason to create new tree instance each time we want to check 1 object. Might mitigate this issue.

efiop: Update: caching tree doesn't replace this optimization (because we still need to walk through it), but actually speeds it up significantly on top. E.g. on master it takes ~4.5sec, with this PR - ~2.5sec and with tree caching ~1.5sec.

Suor:
Cool, thanks!

PR #4429, opened 2020-08-20
Mark test_new_simple as flaky

skshetry:
It is for short period of time till it gets resolved.
See: #4401


skshetry:
Probably better to fix the issue, I guess. When I locally inspect, I always get a string whereas in Travis, it seems it receives a `git.Commit` object.

Maybe just `commit.hexsha[:7]` would do.

skshetry:
Looks like https://github.com/iterative/dvc/pull/4429#issuecomment-677032398 won't solve problem as expectation is not to even get that error.

PR #4432, opened 2020-08-20
experiments: abstract executor tmpdir population using tree upload

### LINKED ISSUE #4401: experiments: test_new_simple is flaky

pmrowla:

Should fix #4401

testing with 
```
@pytest.mark.parametrize("runs", range(1000))
```

I no longer see any failures in a linux VM, before these changes I would see a few failures per 100 runs

### DIFF
--- a/dvc/repo/experiments/__init__.py
+++ b/dvc/repo/experiments/__init__.py
@@ -15,12 +15,13 @@ from typing import Iterable, Optional
 
 from funcy import cached_property, first
 
-from dvc.exceptions import DownloadError, DvcException
+from dvc.exceptions import DownloadError, DvcException, UploadError
 from dvc.path_info import PathInfo
 from dvc.progress import Tqdm
 from dvc.repo.experiments.executor import ExperimentExecutor, LocalExecutor
 from dvc.scm.git import Git
 from dvc.stage.serialize import to_lockfile
+from dvc.tree.repo import RepoTree
 from dvc.utils import dict_sha256, env2bool, relpath
 from dvc.utils.fs import remove
 
@@ -53,6 +54,8 @@ class UnchangedExperimentError(DvcException):
 
 class BaselineMismatchError(DvcException):
     def __init__(self, rev, expected):
+        if hasattr(rev, "hexsha"):
+            rev = rev.hexsha
         rev_str = f"{rev[:7]}" if rev is not None else "dangling commit"
         super().__init__(
             f"Experiment derived from '{rev_str}', expected '{expected[:7]}'."
@@ -249,6 +252,7 @@ class Experiments:
         """Commit stages as an experiment and return the commit SHA."""
         if not self.scm.is_dirty():
             raise UnchangedExperimentError(self.scm.get_rev())
+
         rev = self.scm.get_rev()
         exp_name = f"{rev[:7]}-{exp_hash}"
         if branch:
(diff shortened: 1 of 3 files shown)


dvc/tree/local.py
```diff
@@ -326,7 +326,7 @@ def _upload(
         )
 
         self.protect(tmp_file)
-        os.rename(tmp_file, to_info)
+        os.replace(tmp_file, to_info)
```
pmrowla, line 329: `os.rename` fails in windows if the destination path already exists. Given that the upload behavior for all of our trees is to replace an existing file, it should be safe for us to use `os.replace` here instead.

efiop: actually `_upload` shouldn't replace an already existing file. At least not in `push` logic, as we have a `status` stage before that to ensure that destination doesn't exist yet. All other cloud types will also error-out if `to_info` already exists. But indeed, the reason for `rename` is to just provide an atomic upload behavior, which is still atomic with `replace` logic, but I'm worried about this being different from all the other clouds. We could also catch that error and delete the temporary file, that would be an equivalent way to do this properly.

efiop: Just to clarify, this issue raised while uploading experiments results?

pmrowla: In this case this is while uploading the initial contents of the experiment temp directory (prior to this PR we just did direct local filesystem copy into the temp dir instead of using tree.upload).

I was init'ing an empty dvc repo in the temp directory, and then uploading the contents of the experiments workspace into that repo directory. But the uploaded .dvcignore conflicts with the default existing one and fails in windows (but in linux we silently replace the existing one on upload, since `os.rename` replaces files in posix platforms)

```
DEBUG    dvc.tree.base:base.py:387 Uploading '.dvc\experiments\.dvcignore' to '..\..\..\..\tmpkal7kqwc\.dvcignore'
ERROR    dvc.cache.local:local.py:40 failed to upload '.dvc\experiments\.dvcignore' to '..\..\..\..\tmpkal7kqwc\.dvcignore'
  File "C:\Users\travis\build\iterative\dvc\dvc\cache\local.py", line 30, in wrapper
    func(from_info, to_info, *args, **kwargs)
  File "C:\Users\travis\build\iterative\dvc\dvc\tree\base.py", line 395, in upload
    no_progress_bar=no_progress_bar,
  File "C:\Users\travis\build\iterative\dvc\dvc\tree\local.py", line 329, in _upload
    os.rename(tmp_file, to_info)
FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'C:\\Users\\travis\\AppData\\Local\\Temp\\tmpkal7kqwc\\.dvcignore.MgVcTNDu4KRnUV3yUYM2Mm.tmp' -> 'C:\\Users\\travis\\AppData\\Local\\Temp\\tmpkal7kqwc\\.dvcignore'
```

If the replace-existing behavior in linux is unintended, I can rework how I do the experiments upload

efiop: @pmrowla If you feel like `upload` fits very well, then maybe consider adding an `overwrite` flag or something for it to make that behavior explicit.

pared (approved):

dvc/tree/local.py
pmrowla: @efiop I'm still a bit confused on this because as far as I can tell, our other cloud trees do allow overwriting when using `upload` directly (which is what I am doing for experiments).

```
>>> s3 = get_remote(repo, name="s3-temp")
>>> s3
Remote: 's3://dvc-temp/peter-test'
>>> from_path = PathInfo("foo.txt")
>>> to_path = s3.tree.path_info / "foo.txt"
>>> s3.tree.upload(from_path, to_path)
>>> open(from_path).read()
'foo\n'
>>> with s3.tree.open(to_path) as s3obj:
...     s3obj.read()
...
'foo\n'
>>> open(from_path, "w").write("bar")
3
>>> s3.tree.upload(from_path, to_path)
>>> with s3.tree.open(to_path) as s3obj:
...     s3obj.read()
...
'bar'
```

pmrowla: when we do an actual cache `push` (using `LocalCache._process`) we do the status check, and won't try to `remote_tree.upload` any file that already exists (since it's a cache file, and we would be re-uploading the same exact content).

But for cases where we want to use `remote_tree.upload` directly, I don't really think we should need an extra flag for `overwrite`?

pmrowla:
this PR is now waiting on #4442

Issue #2180, opened 2019-06-24
status: granular output for directories


MikkelAntonsen:
Imagine dvc tracking a directory full of images and files containing labels.

If I delete/change/update a label or image, dvc will tell me that something in the directory has changed, but not exactly what. It would be nice is dvc status could be made more granular, returning something like 

```
data.dvc:
    outputs:
        data/foo: new
        data/bar: deleted
```

instead of 

```
data.dvc:
	changed outs:
		modified:           data
```

as it is today.

Thanks:)

efiop:
Discord context https://discordapp.com/channels/485586884165107732/485596304961962003/592632660539604993

efiop:
Entry points where to start looking at this:

1) Here is where the status is being created https://github.com/iterative/dvc/blob/0.41.3/dvc/output/base.py#L178
2) Need to make `Output.changed_checksum()` either more granular to report particular files in the directory that has changed or create even a new method that will return a status dict for a directory;

shcheklein:
we should be careful with this, imagine a directory with 1M new files. Probably we don't want to show all of them. I would say we should show a summary by default or at least on some threshold (on number of changed files per directory)

jorgeorpinel:
Once this is implemented, I recommend we deprecate the -R/--recursive option. As `status` seems to be recursive already, so just using a dir as target would suffice (full context: https://github.com/iterative/dvc/issues/4191).

pared:
I agree with @shcheklein. We could probably set some small default threshold of files to show (like 5, to not pollute the terminal) and add option to show all changes.
 In case of default dir `status` I wouldn't display it as `file` -> `status` mapping, but rather
`status` -> `list of files with that status` just to save some space (in case of explicit `show-all-changes` it makes more sense to me to show `file` -> `status` mapping, as one probably does not want to scroll through the terminal to see the status of a particular file.

jorgeorpinel:
> We could probably set some small default threshold of files to show (like 5

With pagination?

> I wouldn't display it as file -> status mapping, but rather status -> list of files

Agree. I think we do this in other commands already like metrics/params show/diff? And other output

pared:
> With pagination?

I would refrain from that in the default case. By default, I think it's better to "show the status of dir, but if it's small enough, we can go for particular files" rather than make user interact with it. 

> Agree. I think we do this in other commands already like metrics/params show/diff? And other output

Yes, we do, I was just referring to @MikkelAntonsen example.

[...]

PR #4433, opened 2020-08-20
Outputs as target supporting for `dvc status`

### LINKED ISSUE #4191: status: support outputs as targets [qa]
## Bug Report

UPDATE: Jump to https://github.com/iterative/dvc/issues/4191#issuecomment-657106691

---

Looks like `dvc status` is recursive by default now (maybe always was) from some testing I just did:

```console
λ dvc status
foo.dvc:
        changed outs:
                modified:           foo
data\raw.dvc:
        changed outs:
                modified:           data\raw

λ dvc status -R data/
data\raw.dvc:
        changed outs:
                modified:           data\raw
```

1. So it's really just useful to limit the status to specific directories (not really about recursion).
2. Why not just support dirs AND FILES as `targets` of the command, like all other commands that take targets (I think)?
  > Relater to https://github.com/iterative/dvc.org/pull/1384#issuecomment-656878839
3. In fact this is already supported by `status`, but only in remote mode (`-r` or `-c` options), which is kind of confusing (and complicates the docs).

```
DVC version: 1.1.2
Python version: 3.7.5
Binary: True
Package: exe
Cache: reflink - not supported, hardlink - supported, symlink - not supported
Filesystem type (cache directory): ('NTFS', 'C:\\')
Repo: dvc, git
```

karajan1001:
fix #4191
1. Add a related test which would fail on current version.

### DIFF
--- a/dvc/repo/status.py
+++ b/dvc/repo/status.py
@@ -10,31 +10,32 @@ from . import locked
 logger = logging.getLogger(__name__)
 
 
-def _joint_status(stages):
+def _joint_status(pairs):
     status_info = {}
 
-    for stage in stages:
+    for stage, filter_info in pairs:
         if stage.frozen and not stage.is_repo_import:
             logger.warning(
                 "{} is frozen. Its dependencies are"
                 " not going to be shown in the status output.".format(stage)
             )
-
-        status_info.update(stage.status(check_updates=True))
+        if not filter_info:
+            status_info.update(stage.status(check_updates=True))
+        else:
+            for out in stage.filter_outs(filter_info):
+                status_info.update(out.status())
 
     return status_info
 
 
 def _local_status(self, targets=None, with_deps=False, recursive=False):
-    if targets:
-        stages = cat(
-            self.collect(t, with_deps=with_deps, recursive=recursive)
-            for t in targets
-        )
-    else:
-        stages = self.collect(None, with_deps=with_deps, recursive=recursive)
+    targets = targets or [None]
+    pairs = cat(
+        self.collect_granular(t, with_deps=with_deps, recursive=recursive)
+        for t in targets
+    )
 
-    return _joint_status(stages)
+    return _joint_status(pairs)
 
 
 def _cloud_status(
--- a/tests/func/test_status.py
+++ b/tests/func/test_status.py
@@ -118,3 +118,23 @@ def test_status_recursive(tmp_dir, dvc):
             }
         ],
     }
+
+
+def test_status_outputs(tmp_dir, dvc, caplog):
+    dvc.run(
+        outs=["alice", "bob"],
+        cmd="echo alice>alice && echo bob>bob",
+        name="alice_bob",
+    )
+    tmp_dir.gen("alice", "bob")
+    tmp_dir.gen("bob", "alice")
+    assert main(["status", "alice_bob"]) == 0
+    assert "alice_bob:" in caplog.text
+    assert "changed outs:" in caplog.text
+    assert "modified:           alice" in caplog.text
+    assert "modified:           bob" in caplog.text
+    caplog.clear()
+
+    assert main(["status", "alice"]) == 0
+    assert "modified:           alice" in caplog.text
+    assert "modified:           bob" not in caplog.text
--- a/tests/func/test_status.py
+++ b/tests/func/test_status.py
@@ -121,8 +121,10 @@ def test_status_recursive(tmp_dir, dvc):
 
 
 def test_status_outputs(tmp_dir, dvc, caplog):
+    tmp_dir.dvc_gen({"foo": "foo", "bar": "bar"})
     dvc.run(
         outs=["alice", "bob"],
+        deps=["foo", "bar"],
         cmd="echo alice>alice && echo bob>bob",
         name="alice_bob",
     )
--- a/tests/func/test_status.py
+++ b/tests/func/test_status.py
@@ -128,8 +128,7 @@ def test_status_outputs(tmp_dir, dvc, caplog):
         cmd="echo alice>alice && echo bob>bob",
         name="alice_bob",
     )
-    tmp_dir.gen("alice", "bob")
-    tmp_dir.gen("bob", "alice")
+    tmp_dir.gen({"alice": "new alice", "bob": "new bob"})
     assert main(["status", "alice_bob"]) == 0
     assert "alice_bob:" in caplog.text
     assert "changed outs:" in caplog.text
--- a/tests/func/test_status.py
+++ b/tests/func/test_status.py
@@ -120,7 +120,7 @@ def test_status_recursive(tmp_dir, dvc):
     }
 
 
-def test_status_outputs(tmp_dir, dvc, caplog):
+def test_status_outputs(tmp_dir, dvc):
     tmp_dir.dvc_gen({"foo": "foo", "bar": "bar"})
     dvc.run(
         outs=["alice", "bob"],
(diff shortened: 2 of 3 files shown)


jorgeorpinel:
I marked the documentation checkbox as this is already supposed to be the case and mentioned in https://dvc.org/doc/command-reference/status. Thanks

karajan1001:
Only show outputs in a rough way, need to be rearranged and summarized in a new PR related to #2180.

tests/func/test_status.py
```diff
@@ -118,3 +118,25 @@ def test_status_recursive(tmp_dir, dvc):
...
+    assert main(["status", "alice_bob"]) == 0
+    assert "alice_bob:" in caplog.text
+    assert "changed outs:" in caplog.text
+    assert "modified:           alice" in caplog.text
+    assert "modified:           bob" in caplog.text
+    caplog.clear()
+
+    assert main(["status", "alice"]) == 0
+    assert "modified:           alice" in caplog.text
+    assert "modified:           bob" not in caplog.text
```
pared, line 142: Maybe it would be easier to check the status dictionary (like in test above that one, `test_status_recursive`), instead of caplog messages?

pared (approved):
LGTM, thanks @karajan1001!

dvc/repo/status.py
```diff
@@ -10,31 +10,32 @@
...
                 "{} is frozen. Its dependencies are"
                 " not going to be shown in the status output.".format(stage)
             )
-
-        status_info.update(stage.status(check_updates=True))
+        if not filter_info:
+            status_info.update(stage.status(check_updates=True))
+        else:
+            for out in stage.filter_outs(filter_info):
+                status_info.update(out.status())
```
skshetry, line 26: Should we push this to `Stage::status()` or `Stage::_status_outs`?

tests/func/test_status.py
```diff
@@ -118,3 +118,25 @@ def test_status_recursive(tmp_dir, dvc):
...
+    tmp_dir.dvc_gen({"foo": "foo", "bar": "bar"})
+    dvc.run(
+        outs=["alice", "bob"],
+        deps=["foo", "bar"],
+        cmd="echo alice>alice && echo bob>bob",
+        name="alice_bob",
+    )
+    tmp_dir.gen("alice", "bob")
+    tmp_dir.gen("bob", "alice")
+    assert main(["status", "alice_bob"]) == 0
```
skshetry, line 133: Better to use `dvc.status` here. Also, let's not test with stage name, this case should already be covered.

tests/func/test_status.py
```diff
@@ -118,3 +118,25 @@ def test_status_recursive(tmp_dir, dvc):
...
+def test_status_outputs(tmp_dir, dvc, caplog):
+    tmp_dir.dvc_gen({"foo": "foo", "bar": "bar"})
+    dvc.run(
+        outs=["alice", "bob"],
+        deps=["foo", "bar"],
+        cmd="echo alice>alice && echo bob>bob",
+        name="alice_bob",
+    )
+    tmp_dir.gen("alice", "bob")
+    tmp_dir.gen("bob", "alice")
```
skshetry, line 132: Swapping file contents will make it confusing unless that was the intention? Otherwise, I'll just suggest to:
```suggestion
    tmp_dir.gen({"alice": "new alice", "bob": "new bob"})
```

skshetry (approved):
Thanks @karajan1001 for the PR. Looks good to me, although, have a few suggestions.

dvc/repo/status.py
skshetry: Also, we should keep `{stage_name: stage_status}` format, to make it in line with `--show-json`.

karajan1001: > Thanks @karajan1001 for the PR. Looks good to me, although, have a few suggestions.
@skshetry , Thanks for your suggestions.

> Also, we should keep `{stage_name: stage_status}` format, to make it in line with `--show-json`.
Here, I just follow the format of `dvc status -c [outputs]`. Besides, if we show 

```
"alice_bob":  
        "changed outs": 
                "alice": "modified" 
```
We give misinformation that there is only one output "alice" in stage "alice_bob", here we need more emphasis on outputs than stages.

tests/func/test_status.py
karajan1001: Done.

dvc/repo/status.py
skshetry: @karajan1001, better to be consistent here. If we need better CLI output, we can handle that in `dvc/commands`. But, let's wait for others, let's see what they'll say. 

> It has a different format from the current Stage::status() and Stage::_status_outs now

Regarding  `status`, we could add `filter_info` to it so that you could just do
```suggestion
        status_info.update(stage.status(filter_outs=filter_info))
```

But this would make sense if we made result consistent.

tests/func/test_status.py
karajan1001: @pared 
Already Done.

dvc/repo/status.py
karajan1001: > @karajan1001, better to be consistent here. If we need better CLI output, we can handle that in `dvc/commands`. But, let's wait for others, let's see what they'll say.
> 
> > It has a different format from the current Stage::status() and Stage::_status_outs now
> 
> Regarding `status`, we could add `filter_info` to it so that you could just do
> 
> But this would make sense if we made result consistent.

@skshetry 
I found `filter_info` hard to use. Can we get rid of it? `collect_granular` returns a list of `outputs` and `stages`, and they share the same API `.status()` which returns the result. But seems that `filter_info` is at a finer granularity than outputs, as outputs may be a directory.

pared:
@karajan1001 is there anything that you would like to add here? Or can we drop the `[WIP]`? :)

karajan1001:
> @karajan1001 is there anything that you would like to add here? Or can we drop the `[WIP]`? :)

@pared
I'm sorry, I don't know what is [WIP] and googled it a moment ago. Now I can say yes.

pared:
@karajan1001 no problem, it just felt like the PR is complete, thats why I asked :)

skshetry:
Would have been better to have a consistent format: https://github.com/iterative/dvc/pull/4433/files#r475538104. :man_shrugging:

pared:
@skshetry sorry, my bad. Let's finish the discussion here and prepare follow up PR?

dvc/repo/status.py
pared: I would agree with @skshetry here, 
```
"alice_bob":  
        "changed outs": 
                "alice": "modified" 
```
In that case we inform user that only one out has changed, but that does not mean that this particular stage has only this output. I would say that `status` primary function is to inform about changes, and there is no need to list all of its outputs, if they did not change.

karajan1001: > I would agree with @skshetry here,
> 
> ```
> "alice_bob":  
>         "changed outs": 
>                 "alice": "modified" 
> ```
> 
> In that case we inform user that only one out has changed, but that does not mean that this particular stage has only this output. I would say that `status` primary function is to inform about changes, and there is no need to list all of its outputs, if they did not change.

@pared 
How about the `status -c`? Do they also need to follow this format, or just keep the current one?

pared: @karajan1001 This is a good question.
So in case of  our current test, if we commit modified `alice` we will get `Data and pipelines up to date` on `dvc status`
and we will get `new: alice` on `dvc status -c`. As this issue is about filtering the `status` I would keep output of `dvc status -c` as is and create issue discussing whether we should have consistent output format for both `dvc status` and `dvc status -c`.

PR #4434, opened 2020-08-20
WIP: metadata() on RepoTree

skshetry:


dvc/tree/_metadata.py
```diff
@@ -0,0 +1,55 @@
+from dataclasses import dataclass, field
```
efiop, line 1: Why `_metadata.py` though?

skshetry: I didn't want to expose it outside of the `tree`s. But, don't have a strong opinion.

efiop (approved):
This is really cool! Really eager to merge and work on top :)

dvc/tree/_metadata.py
efiop: Good point. We don't use that technique anywhere else for files in the module, but this is indeed a very suitable place to start. Let's keep it.

skshetry:
@efiop, there are a few things missing:
1. tests
2. add `dataclasses` to requirements for python3.6.
3. I'd like to deprecate `isdvc()`, as it's confusing on what it means. It has 3 different meanings (which Metadata has too, probably I'll remove it as well).

PR #4437, opened 2020-08-20
Enable profiling from cli

skshetry:
```console
$ dvc list . --cprofile
# prints all stats
$ dvc list . --cprofile dvc.prof
# dumps cprofile results in a dvc.prof file
```

Just a simple patch that I was using locally, thought I'd share.

#### Why not `python -m cProfile -m dvc`?
Not available on Python3.6 :(. I'm fine keeping it just for me as well. :smile: 
And, that command feels like it's quite far.

### DIFF
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -137,6 +137,14 @@ def get_parent_parser():
     """
     parent_parser = argparse.ArgumentParser(add_help=False)
 
+    parent_parser.add_argument(
+        "--cprofile",
+        action="store_true",
+        default=False,
+        help=argparse.SUPPRESS,
+    )
+    parent_parser.add_argument("--cprofile-dump", help=argparse.SUPPRESS)
+
     log_level_group = parent_parser.add_mutually_exclusive_group()
     log_level_group.add_argument(
         "-q", "--quiet", action="count", default=0, help="Be quiet."
--- a/dvc/main.py
+++ b/dvc/main.py
@@ -2,6 +2,7 @@
 
 import errno
 import logging
+from contextlib import contextmanager
 
 from dvc import analytics
 from dvc.cli import parse_args
@@ -19,6 +20,25 @@ from dvc.utils import format_link
 "".encode("idna")
 
 logger = logging.getLogger("dvc")
+
+
+@contextmanager
+def profile(enable, dump):
+    if not enable:
+        yield
+        return
+
+    import cProfile
+
+    prof = cProfile.Profile()
+    prof.enable()
+
+    yield
+
+    if not dump:
+        prof.print_stats(sort="cumtime")
+        return
+    prof.dump_stats(dump)
 
 
 def main(argv=None):  # noqa: C901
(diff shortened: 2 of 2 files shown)


dvc/main.py
```diff
@@ -50,8 +74,9 @@ def main(argv=None):  # noqa: C901
 
         logger.trace(args)
 
-        cmd = args.func(args)
-        ret = cmd.run()
+        with profile(enable=args.profile):
+            cmd = args.func(args)
+            ret = cmd.run()
```
efiop, line 79: arguably analytics code down below might also be of some interest to us :smiley:

efiop (approved):
This is really cool! I was also thinking about adding `--pdb` option, so it simply jumps into pdb on error. I think I even have that patch somewhere...

dvc/main.py
```diff
@@ -21,6 +23,28 @@
...
+
+    import cProfile
+    import os
+
+    prof = cProfile.Profile()
+    prof.enable()
+
+    yield
+
+    dump_location = os.getenv("DVC_PROFILE_DUMP")
```
efiop, line 40: Let's have it as a CLI option, no one will remember about this one otherwise.

dvc/cli.py
```diff
@@ -145,6 +145,13 @@ def get_parent_parser():
         "-v", "--verbose", action="count", default=0, help="Be verbose."
     )
 
+    parent_parser.add_argument(
+        "--profile",
```
efiop, line 149: A bit worried about this colliding with `profile` from s3 and other similar things in the future.

efiop: Maybe `--cprofile` ?

skshetry: Yup, could do. How about just `--prof`?

efiop: @skshetry IIRC, argparse counts `--prof` as `--profile` if you have it, so that won't work :D

dvc/main.py
```diff
@@ -21,6 +22,26 @@
 logger = logging.getLogger("dvc")
 
 
+@contextmanager
+def profile(arg):
+    if arg == -1:
```
skshetry, line 27: `arg` is not pretty, but we can now do following:
```
$ dvc list . --cprofile
# prints all stats
$ dvc list . --cprofile dvc.prof
# dumps cprofile results in a dvc.prof file
```

efiop: @skshetry Worried about it being confusing when you don't place it at the end of the command. How about just `--cprofile-out` ? We could potentially do `--cprofile-sort` in the future too.

efiop (approved):

jorgeorpinel:
Nice one. It definitely requires docs updates though, no? I've unmarked the check box for now.

skshetry:
It's meant for us, to quickly find out performance bottlenecks while running a dvc command. It could also be helpful for supporting users for help. 
The users can easily just do `python -m cProfile -m dvc <command>`, so no need for documentation.

jorgeorpinel:
Hidden feature? Got it. Is it also hidden from `-h` output?

skshetry:
@jorgeorpinel, yes it's hidden.

PR #4442, opened 2020-08-21
tree: add overwrite parameter in tree.upload

pmrowla:

(pushing this PR to upstream since we need to run all of the cloud tests)

From https://github.com/iterative/dvc/pull/4432#discussion_r473892090

Currently we have inconsistent behavior between clouds when using `tree.upload` directly. For AzureTree and Windows LocalTree, `tree.upload` will fail if the destination path already exists. On all other platforms the destination path will be always be overwritten if it already exists.

Changes:
* Check for file existence in `tree.upload` if `overwrite=False` and fail if the file already exists.
* `_upload` on all platforms now overwrites by default (since the check for file existence is done in the calling `upload` method).

### DIFF
--- a/dvc/tree/azure.py
+++ b/dvc/tree/azure.py
@@ -167,7 +167,7 @@ class AzureTree(BaseTree):
             with Tqdm.wrapattr(
                 fobj, "read", desc=name, total=total, disable=no_progress_bar
             ) as wrapped:
-                blob_client.upload_blob(wrapped)
+                blob_client.upload_blob(wrapped, overwrite=True)
 
     def _download(
         self, from_info, to_file, name=None, no_progress_bar=False, **_kwargs
--- a/dvc/tree/local.py
+++ b/dvc/tree/local.py
@@ -326,7 +326,7 @@ class LocalTree(BaseTree):
         )
 
         self.protect(tmp_file)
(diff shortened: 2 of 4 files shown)


dvc/tree/local.py
```diff
@@ -326,7 +326,7 @@ def _upload(
         )
 
         self.protect(tmp_file)
-        os.rename(tmp_file, to_info)
+        os.replace(tmp_file, to_info)
```
pmrowla, line 329: `os.replace` provides the same atomicity guarantee as `os.rename`, now that we check for file existence in `BaseTree.upload`, we should be able to safely default to overwrite behavior in `_upload` across all of our cloud specific trees.

dvc/tree/azure.py
```diff
@@ -167,7 +167,7 @@ def _upload(
             with Tqdm.wrapattr(
                 fobj, "read", desc=name, total=total, disable=no_progress_bar
             ) as wrapped:
-                blob_client.upload_blob(wrapped)
+                blob_client.upload_blob(wrapped, overwrite=True)
```
pmrowla, line 170: same as the change in LocalTree:
> now that we check for file existence in `BaseTree.upload`, we should be able to safely default to overwrite behavior in `_upload` across all of our cloud specific trees.

PR #4452, opened 2020-08-24
experiments: support binary and untracked files in experiment patches

### LINKED ISSUE #4451: experiments: experiment git patching will fail if patch contains binary files
Example from experiment using tensorflow ModelCheckpoint files

```
2020-08-24 12:56:57,888 ERROR: failed to apply experiment changes.
  File "/Users/pmrowla/git/dvc/dvc/repo/experiments/__init__.py", line 529, in checkout_exp
    self.repo.scm.repo.git.apply(tmp, reverse=True)
  File "/Users/pmrowla/.virtualenvs/dvc/lib/python3.8/site-packages/git/cmd.py", line 542, in <lambda>
    return lambda *args, **kwargs: self._call_process(name, *args, **kwargs)
  File "/Users/pmrowla/.virtualenvs/dvc/lib/python3.8/site-packages/git/cmd.py", line 1005, in _call_process
    return self.execute(call, **exec_kwargs)
  File "/Users/pmrowla/.virtualenvs/dvc/lib/python3.8/site-packages/git/cmd.py", line 822, in execute
    raise GitCommandError(command, status, stderr_value, stdout_value)
git.exc.GitCommandError: Cmd('git') failed due to: exit code(1)
  cmdline: git apply --reverse /var/folders/s_/j8cg24s945dcvlh77c677f080000gn/T/tmptqvdsgr5
  stderr: 'error: missing binary patch data for 'model.hdf5'
error: binary patch does not apply to 'model.hdf5'
error: model.hdf5: patch does not apply'
```

In this case, `model.hdf5` is untracked by both git and dvc. Unless a repro artifact is explicitly gitignored, we should be including it with our experiment/executor output, so we should be using `git diff --full-index --binary` and `git stash --include-untracked` when generating our patches

pmrowla:

Will fix #4451

### DIFF
--- a/dvc/repo/experiments/__init__.py
+++ b/dvc/repo/experiments/__init__.py
@@ -182,7 +182,9 @@ class Experiments:
         # patch user's workspace into experiments clone
         tmp = tempfile.NamedTemporaryFile(delete=False).name
         try:
-            self.repo.scm.repo.git.diff(patch=True, output=tmp)
+            self.repo.scm.repo.git.diff(
+                patch=True, full_index=True, binary=True, output=tmp
+            )
             if os.path.getsize(tmp):
                 logger.debug("Patching experiment workspace")
                 self.scm.repo.git.apply(tmp)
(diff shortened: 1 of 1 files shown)


pmrowla (commented):
One concern here would be that we could potentially be sticking large binary files into git, which git is bad at handling and why dvc exists in the first place. (But only in the experiments workspace, and not the user's main repo)

I'm not sure if there's a cleaner way for us to handle this situation other than documenting that best practices here would be to either mark these types of large binary intermediate outputs as dvc outs or to gitignore them.

pmrowla:
As a potential future improvement, we could maybe consider doing something like generating directory containing all of the untracked files generated by an experiment run (along with a manifest or something containing the original relpaths for the untracked files), and then storing that as an experiment-specific dvc-tracked directory. But in the meantime, this bugfix is needed for testing/development purposes.

pared (approved):

efiop (approved):

PR #4456, opened 2020-08-24
add py support for ParamsDependency

aandrusenko:
* add py support for ParamsDependency

This PR adds support for parameters file in ".py" format. Example:

```
dvc run -n train -d stages/train.py -p params.py:SEED,Train python stages/train.py
```

Where params.py looks something like this:
```
SEED = 1

class Train:
    lr = 0.0041
    epochs = 75
    layers = 9
```

dvc.lock looks like this:
```
train:
  cmd: python stages/train.py
  deps:
  - path: stages/train.py
    md5: 23be4307b23dcd740763d5fc67993f11
  params:
    params.py:
      SEED: 1
      Train:
        epochs: 75
        layers: 9
        lr: 0.0041
```

### DIFF
--- a/dvc/utils/serialize/_py.py
+++ b/dvc/utils/serialize/_py.py
@@ -1,0 +1,94 @@
+import ast
+
+from funcy import reraise
+
+from ._common import ParseError, _load_data
+
+
+class PythonFileCorruptedError(ParseError):
+    def __init__(self, path, message="PY file structure is corrupted"):
+        super().__init__(path, message)
+
+
+def load_py(path, tree=None):
+    return _load_data(path, parser=parse_py, tree=tree)
+
+
+def parse_py(text, path):
+    """Parses text from .py file into Python structure."""
+    with reraise(SyntaxError, PythonFileCorruptedError(path)):
+        tree = ast.parse(text, filename=path)
+
+    result = _ast_tree_to_dict(tree)
+    return result
+
+
+def _ast_tree_to_dict(tree, only_self_params=False):
+    """Parses ast trees to dict."""
+    result = {}
+    for _body in tree.body:
+        try:
+            if isinstance(_body, (ast.Assign, ast.AnnAssign)):
+                result.update(_ast_assign_to_dict(_body, only_self_params))
+            elif isinstance(_body, ast.ClassDef):
+                result.update({_body.name: _ast_tree_to_dict(_body)})
+            elif (
+                isinstance(_body, ast.FunctionDef) and _body.name == "__init__"
+            ):
+                result.update(_ast_tree_to_dict(_body, only_self_params=True))
+        except ValueError:
+            continue
+        except AttributeError:
+            continue
+    return result
+
+
+def _ast_assign_to_dict(assign, only_self_params=False):
+    result = {}
+
+    if isinstance(assign, ast.AnnAssign):
+        name = _get_ast_name(assign.target, only_self_params)
+    elif len(assign.targets) == 1:
+        name = _get_ast_name(assign.targets[0], only_self_params)
+    else:
+        raise AttributeError
+
+    if isinstance(assign.value, ast.Dict):
+        _dct = {}
+        for key, val in zip(assign.value.keys, assign.value.values):
+            _dct[_get_ast_value(key)] = _get_ast_value(val)
+        result[name] = _dct
+    elif isinstance(assign.value, ast.List):
+        result[name] = [_get_ast_value(val) for val in assign.value.elts]
+    elif isinstance(assign.value, ast.Set):
+        values = [_get_ast_value(val) for val in assign.value.elts]
+        result[name] = set(values)
+    elif isinstance(assign.value, ast.Tuple):
+        values = [_get_ast_value(val) for val in assign.value.elts]
+        result[name] = tuple(values)
+    else:
+        result[name] = _get_ast_value(assign.value)
+
+    return result
+
+
+def _get_ast_name(target, only_self_params=False):
+    if hasattr(target, "id") and not only_self_params:
+        result = target.id
+    elif hasattr(target, "attr") and target.value.id == "self":
+        result = target.attr
+    else:
+        raise AttributeError
+    return result
+
+
+def _get_ast_value(value):
+    if isinstance(value, ast.Num):
+        result = value.n
+    elif isinstance(value, ast.Str):
+        result = value.s
+    elif isinstance(value, ast.NameConstant):
+        result = value.value
+    else:
+        raise ValueError
+    return result
--- a/dvc/utils/serialize/_py.py
+++ b/dvc/utils/serialize/_py.py
@@ -1,12 +1,16 @@
 import ast
+from contextlib import contextmanager
 
 from funcy import reraise
 
-from ._common import ParseError, _load_data
+from ._common import ParseError, _dump_data, _load_data, _modify_data
+
+_PARAMS_KEY = "__params_old_key_for_update__"
+_PARAMS_TEXT_KEY = "__params_text_key_for_update__"
 
 
 class PythonFileCorruptedError(ParseError):
-    def __init__(self, path, message="PY file structure is corrupted"):
+    def __init__(self, path, message="Python file structure is corrupted"):
         super().__init__(path, message)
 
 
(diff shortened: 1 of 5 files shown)


tests/func/params/test_show.py
```diff
@@ -24,6 +24,16 @@ def test_show_toml(tmp_dir, dvc):
     }
 
 
+def test_show_py(tmp_dir, dvc):
+    tmp_dir.gen("params.py", "class Config:\n    foo = 42\n")
+    dvc.run(
+        cmd="echo params.py", params=["params.py:foo"], single_stage=True
```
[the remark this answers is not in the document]
aandrusenko: Not right now, all `Config` parameters are parsed into a dictionary, so only `foo`.
If you want, for example, all classes to be parsed into a dictionary individually, I can do.
But if you don't use classes at all, it's very difficult to separate the wheat from the chaff when the ".py" file is parsed. The ".py" can contain imports, random global variables, etc.

dvc/utils/serialize/_py.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def dump_py(path, data, tree=None):
+    return _dump_data(path, data, dumper=_dump, tree=tree)
+
+
+@contextmanager
+def modify_py(path, tree=None):
+    with _modify_data(path, parse_py_for_update, dump_py, tree=tree) as d:
+        yield d
```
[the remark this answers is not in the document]
aandrusenko: I wanted to consult. How realistic is it to update the ".py" file from the dictionary. For example, something like this:
```
dict = {"foo": 1, "train": {"bar": 2}}

param = "    {name} = {value}\n"
if dict in dict:
    text = "class {class_name}:\n{params}\n".format(
        class_name="Train",
        params=[param.format(name=n, value=v) for n, v in dict["train"].items()]
    )

text += "class Config:\n{params}\n".format(
    params=[param.format(name=n, value=v) for n, v in dict.items()]
)

# save text to ".py" file
```
Even the simplest version looks very strange. I think this will be extremely difficult to maintain. Maybe you have some ideas.

tests/func/params/test_show.py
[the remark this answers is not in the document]
efiop: @aandrusenko `Config` class convention seems like a pretty strong one, I'm not sure how common it is. My guess would be that having global variables is actually more common than a special class.

dvc/utils/serialize/_py.py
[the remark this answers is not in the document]
skshetry: This will take quite an effort to do correctly, especially with preserving comments, typed comments, etc. It'll need to use `tokenize` and `ast` (which btw does not support typing comments till 3.8, so we might need to use `typed_ast`). 

And, again using these, it's hard to do round-trip correctly, needs more time and effort.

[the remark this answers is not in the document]
skshetry: For now, `modify_*` is used for hyperparameter optimizations, which is an under-development feature. We can make without it as well for now.

aandrusenko:
@skshetry, I changed the approach to parsing the file. Now all global variables and classes are converted to a parameter dictionary by `ast` module. Parameters are used with `Config.foo`.

dvc/utils/serialize/_py.py
```diff
@@ -0,0 +1,89 @@
...
+            continue
+        except AttributeError:
+            continue
+    return result
+
+
+def ast_assign_to_dict(assign):
+    result = {}
+
+    if len(assign.targets) == 1:
```
skshetry, line 49: Looks like this does not work with annotations based assignment.
```python
is_dir: bool = True
```

But, for now, we could totally move without it.

dvc/utils/serialize/_py.py
```diff
@@ -0,0 +1,89 @@
+import ast
```
skshetry, line 1: I believe this is not cross-python-version compatible, right? Not a blocker though.

skshetry (commented):
Thanks @aandrusenko :slightly_smiling_face: 
This looks great and works amazingly well. 
Due to the nature of this PR, it might take a few days/deep-dives to get it merged. I hope that's okay.

Meanwhile, can you rebase? Thanks again, this is a great and very helpful feature.  :tada:

dvc/utils/serialize/_py.py
aandrusenko: I fixed it.

dvc/utils/serialize/_py.py
aandrusenko: ast uses the compiler of the installed python. As I understand it, ast is compatible https://stackoverflow.com/a/26656125

dvc/utils/serialize/_py.py
```diff
@@ -0,0 +1,96 @@
...
+
+from funcy import reraise
+
+from ._common import ParseError, _load_data
+
+
+class PYFileCorruptedError(ParseError):
```
skshetry, line 8:
```suggestion
class PythonFileCorruptedError(ParseError):
```

dvc/utils/serialize/_py.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def ast_tree_to_dict(tree):
+    """Parses ast trees to dict."""
+    result = {}
+    for _body in tree.body:
+        try:
+            if isinstance(_body, ast.Assign) or isinstance(
+                _body, ast.AnnAssign
+            ):
```
skshetry, line 33:
```suggestion
            if isinstance(_body, (ast.Assign, ast.AnnAssign)):
```

dvc/utils/serialize/_py.py
```diff
@@ -0,0 +1,96 @@
...
+                result.update(ast_assign_to_dict(_body))
+            elif isinstance(_body, ast.ClassDef):
+                result.update({_body.name: ast_tree_to_dict(_body)})
+            elif (
+                isinstance(_body, ast.FunctionDef) and _body.name == "__init__"
+            ):
+                result.update(ast_tree_to_dict(_body))
```
skshetry, line 40: Looks like this reads from variable instead of `self.*` right?

```python
class Klass:
    def __init__(self):
       self.a = 3
```

I'd expect `Klass.a` be 3 here. If it's difficult, we could remove this functionality.

dvc/utils/serialize/_py.py
```diff
@@ -0,0 +1,96 @@
...
+    elif isinstance(assign.value, ast.Set):
+        values = [get_ast_value(val) for val in assign.value.elts]
+        result[name] = set(values)
+    elif isinstance(assign.value, ast.Tuple):
+        values = [get_ast_value(val) for val in assign.value.elts]
+        result[name] = tuple(values)
+    else:
+        result[name] = get_ast_value(assign.value)
+
+    return result
```
skshetry, line 74: This does not work with recursive data. Not a problem though, just a note.

dvc/utils/serialize/_py.py
```diff
@@ -0,0 +1,96 @@
...
+
+def get_ast_value(value):
+    if isinstance(value, ast.Num):
+        result = value.n
+    elif isinstance(value, ast.Str):
+        result = value.s
+    elif isinstance(value, ast.NameConstant):
+        result = value.value
+    else:
+        raise ValueError
```
skshetry, line 95: Note that `ast.{Num/Str/NameConstant}` has been deprecated in Python3.8, but still works on 3.8.

skshetry: Fine as-is for now though.

dvc/utils/serialize/_py.py
skshetry: What I meant by that is:
```python
lst = [1, 2, 3, 4, [1, 2, 3, 4]]
```

dvc/utils/serialize/_py.py
```diff
@@ -0,0 +1,96 @@
...
+def parse_py(text, path):
+    """Parses text from .py file into Python structure."""
+    with reraise(SyntaxError, PYFileCorruptedError(path)):
+        tree = ast.parse(text, filename=path)
+
+    result = ast_tree_to_dict(tree)
+    return result
+
+
+def ast_tree_to_dict(tree):
```
skshetry, line 26: Let's make this an internal API as we do `*` imports.
https://github.com/iterative/dvc/blob/79e8e4edafb8a4745200f0a62f49c29ef074bdea/dvc/utils/serialize/__init__.py#L6

```suggestion
def _ast_tree_to_dict(tree):
```

skshetry: Applies to all other functions except `load_py` and ` parse_py`.

skshetry (commented):
Looks good overall. Most of the suggestions below are minor refactoring and a question about variables inside `__init__` of the class.

dvc/utils/serialize/_py.py
skshetry: For future reference only. Not sure how dot-based access would even work here. Again, no need to implement.

dvc/utils/serialize/_py.py
aandrusenko: Right, `Klass.a` be 3. It's works. Added to tests.

skshetry: Ahh, I thought it didn't. Now, I see how it works, but might be confusing to the users, especially in the following scenarios:
1.
```python
x = 3
config.x = 3
```
`parse_py` gives me result `{"x": 3}`.

2. 
```python
class Config:

   def __init__(self):
      self.container.value = 5
       self.value = 1
       value = 4  # used for something else
```

`parse_py` gives me result `{"Config": {"value": "4"}}`, instead of expected "1" (neither 4 or 5 is expected).

pared (approved):
Looks good to me, great job @aandrusenko. :+1:

dvc/utils/serialize/_py.py
aandrusenko: Yeah, added a flag to read only self parameters

skshetry: Hi @aandrusenko, this problem still exists, right? See the other two scenarios with `self.container.value` and `container.x` above.

dvc/utils/serialize/_py.py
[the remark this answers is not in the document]
skshetry: Is it possible to implement "modification" with AST now? We are working on a new feature, Experiments ([working draft](https://github.com/iterative/dvc/wiki/Experiments-development-status#repro)), which automatically modifies parameters based on hyper-parameter optimization, so it'd be great to have that if it's possible.

dvc/utils/serialize/_py.py
skshetry: We could either ignore these, or return a dict containing ```json
{"Config": {"container": {"value": 5}}}
```

Comment (aandrusenko) [reply dvc/utils/serialize/_py.py]: Fixed, reads only `self.value` without any implementations

File: dvc/utils/serialize/_py.py (hunk 2 above)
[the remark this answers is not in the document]
Comment (aandrusenko) [reply dvc/utils/serialize/_py.py]: I added support for py parameter modifications in the last commit. It looks scary, but it seems to work. Will not work with dictionaries or parameters that span multiple lines, but updating simple parameters should work.

File: dvc/utils/serialize/_py.py (hunk 7 above)
Comment (skshetry) [reply dvc/utils/serialize/_py.py]: Hi, this still fails for non-class attr assignment, right? 
```python
>>> parse_py("x = 4; config.x = 3", "file name does not matter")
{'x': 3}
```

dvc/utils/serialize/_py.py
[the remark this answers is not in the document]
skshetry: That seems to work for the simpler cases we are targeting, it's actually pretty clever. But, I am worried about edge-cases as `dump` is line-based (+ AST of course) whereas `load` is pure-ast based, which might leave users confused on why something got modified vs some that didn't. 

cc @efiop @pared @pmrowla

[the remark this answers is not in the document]
pared: So what options do we have here?
1. Support `modify` and just accept the fact that users might get their params file modified if they use `modify` - that will be confusing, will result in noise in PR's and commits.
2. Don't support it - we need this feature for experiments, but for now we can just "Error: Use yaml or json please" our way out of this for now.
3. Support modify and write our own `dump`-er - that seems like a lot of maintaining

I think out of those 3, the best is still 1 - cosmetic changes during dump will be present, but it can be simply avoided using `black` in pre-commit, and we still preserve the functionality required for experiments.

As to  `value = {"hello": "world"}` case, I would forbid that, at least for now.

dvc/utils/serialize/_py.py
aandrusenko: @skshetry, fixed, but I have no more ideas how to ignore `self.x` outside the class, `ast` has no more functionality. In my opinion, it is strange to meet `self.x` outside the class, because `self` is not defined and params.py cannot even be imported

dvc/utils/serialize/_py.py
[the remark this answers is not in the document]
aandrusenko: @skshetry, so what should I do, leave it as it is now?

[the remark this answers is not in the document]
skshetry: @pmrowla ping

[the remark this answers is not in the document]
pmrowla: Seems like leaving this as-is for now will be best, I think having this partial support for modifying python params is better than nothing at all.

When experiments is ready for release we will just have to document that support for python params is limited to these simple cases, and that yaml/toml should be preferred if using the experiments features.

skshetry (approved):
Looks good, @aandrusenko. Would you mind creating a docs PR on the https://github.com/iterative/dvc.org?

jorgeorpinel:
Awesome! I haven't read the whole conversation above but I have a Q: What happens if the user doesn't have Python installed and tries to use this feature? I know, it's a probably an edge case but it's possible to install DVC without Python with binary builds so... Just curious.

skshetry:
@jorgeorpinel, it depends on the python version the DVC was installed with, so it's tied to the same python version as well.

skshetry:
@aandrusenko, can you rebase? There are merge conflicts.

efiop (approved):
Thanks @aandrusenko ! :pray:

jorgeorpinel:
> it depends on the python the DVC was installed with, so it's tied to the same python version

@skshetry what about binary installers? Those don't require a pre-existing Python in the system. That's what I meant. Will this work without a system Python in that case?

PR #4458, opened 2020-08-24
api: refactor to use metadata and RepoTree

### LINKED ISSUE #3182: api: get_url() returns path to `.dir` for directory
Currently, `get_url()` returns path to `.dir` files. I'd propose to have some way to parse `get_url()` and return list or dicts (or, iterator) of  files inside the directory (`get_url(recurse=True)`?).

`dvc get . dir/ --show-url` works the same way.

skshetry:
This does not yet support subrepos as it requires us to work on setting up cache correctly which is pending.
After that, it should just be changing `subrepos=False` to `True`.

Still unsure about _proper_ exceptions here, but I did try to unify them somehow (still does not look good though).

Closes #3182

### DIFF
--- a/dvc/api.py
+++ b/dvc/api.py
@@ -2,16 +2,17 @@ import os
 from contextlib import _GeneratorContextManager as GCM
 from contextlib import contextmanager
 
-from dvc.exceptions import DvcException, NotDvcRepoError
+from funcy import reraise
+
+from dvc.exceptions import (
+    NotDvcRepoError,
+    OutputNotFoundError,
+    PathMissingError,
+)
 from dvc.external_repo import external_repo
+from dvc.path_info import PathInfo
 from dvc.repo import Repo
-
-
-class UrlNotDvcRepoError(DvcException):
-    """Thrown if the given URL is not a DVC repository."""
-
-    def __init__(self, url):
-        super().__init__(f"'{url}' is not a DVC repository.")
+from dvc.tree.repo import RepoTree
 
 
 def get_url(path, repo=None, rev=None, remote=None):
@@ -20,17 +21,26 @@ def get_url(path, repo=None, rev=None, remote=None):
     in a DVC repo. For Git repos, HEAD is used unless a rev argument is
     supplied. The default remote is tried unless a remote argument is supplied.
 
-    Raises UrlNotDvcRepoError if repo is not a DVC project.
+    Raises OutputNotFoundError if the file is not a dvc-tracked file.
 
     NOTE: This function does not check for the actual existence of the file or
     directory in the remote storage.
     """
     with _make_repo(repo, rev=rev) as _repo:
-        if not isinstance(_repo, Repo):
-            raise UrlNotDvcRepoError(_repo.url)  # pylint: disable=no-member
-        out = _repo.find_out_by_relpath(path)
-        remote_obj = _repo.cloud.get_remote(remote)
-        return str(remote_obj.tree.hash_to_path_info(out.checksum))
+        tree = RepoTree(_repo, fetch=True)
+        path_info = PathInfo(_repo.root_dir) / path
+
+        meta = tree.metadata(path_info)
+        exc = OutputNotFoundError(path)
+        if not meta.is_dvc:
+            raise exc
+
+        with reraise(FileNotFoundError, exc):
+            _, hash_ = tree.get_hash(PathInfo(_repo.root_dir) / path)
+
+        assert hash_ and meta.repo
+        cloud = meta.repo.cloud
+        return cloud.get_url_for(remote, checksum=hash_)
 
 
 def open(  # noqa, pylint: disable=redefined-builtin
@@ -74,10 +84,14 @@ class _OpenContextManager(GCM):
 
 def _open(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
     with _make_repo(repo, rev=rev) as _repo:
-        with _repo.open_by_relpath(
-            path, remote=remote, mode=mode, encoding=encoding
-        ) as fd:
-            yield fd
+        tree = RepoTree(_repo, stream=True, fetch=True)
+
+        path = PathInfo(_repo.root_dir) / path
+        with reraise(FileNotFoundError, PathMissingError(path, repo)):
+            with tree.open(
+                path, remote=remote, mode=mode, encoding=encoding
+            ) as fd:
+                yield fd
 
 
 def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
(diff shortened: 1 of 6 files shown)


dvc/api.py
```diff
@@ -20,17 +21,22 @@ def get_url(path, repo=None, rev=None, remote=None):
...
     NOTE: This function does not check for the actual existence of the file or
     directory in the remote storage.
     """
     with _make_repo(repo, rev=rev) as _repo:
-        if not isinstance(_repo, Repo):
-            raise UrlNotDvcRepoError(_repo.url)  # pylint: disable=no-member
-        out = _repo.find_out_by_relpath(path)
-        remote_obj = _repo.cloud.get_remote(remote)
-        return str(remote_obj.tree.hash_to_path_info(out.checksum))
+        tree = RepoTree(_repo, fetch=True)
```
efiop, line 30:
```suggestion
        tree = DvcTree(_repo, fetch=True)
```

dvc/api.py
```diff
@@ -20,17 +21,22 @@ def get_url(path, repo=None, rev=None, remote=None):
...
+        meta = tree.metadata(PathInfo(_repo.root_dir) / path)
```
efiop, line 31: Actually no need for this, we could just use `tree.get_hash(PathInfo(_repo.root_dir) / path)` directly

dvc/api.py
```diff
@@ -74,10 +80,14 @@ def __getattr__(self, name):
...
-        ) as fd:
-            yield fd
+        tree = RepoTree(_repo, stream=True, fetch=True)
+
+        path = PathInfo(_repo.root_dir) / path
+        with reraise(FileNotFoundError, PathMissingError(path, repo)):
+            with tree.open(
+                path, remote=remote, mode=mode, encoding=encoding
+            ) as fd:
+                yield fd
```
efiop, line 90: Doesn't look like there is a need for this change. Repo.open_by_relpath already does all of this.

skshetry: `open_by_relpath` will most likely go away, as `RepoTree` supports subrepos directly.

efiop: @skshetry Agreed, it was only used for the API, so we could delete it in that case.

dvc/api.py
skshetry: `DvcTree` is mostly internal, and does not support subrepos.

skshetry: Okay, I think I get why you are suggesting that. As `DvcTree` does not support subrepos, we need to implement `RepoTree.get_dvctree(path)` or `RepoTree.get_repo(repo_path)`, as `get_hash` also hashes git-files.

skshetry: Or, `get_hash(dvc_only=True)`? Thoughts?

efiop: Yeah, I suggest it because this is not meant to work with git files. Not a big fan of `dvc_only=True` though, but we could simply do:
```
if not tree.isdvc(path):
    raise ....
hash = tree.get_hash() 
```
:slightly_smiling_face:

efiop: Theoretically we could've made hash part of the metadata, but that might make it complicated for files inside dirs, as you'll need to parse the dir_cache (as we do in _get_granular_checksum, which is not that bad)

efiop: Btw, this will automatically close #3182  :slightly_smiling_face:

skshetry: Okay, did as you recommended, but `metadata`/`hash`/`RepoTree` all gets mixed. Feels like that thing could be a part of `RepoTree` on itself.

dvc/api.py
skshetry: Exceptions do not look good, I do have to tell you that because `Repo` and `ExternalRepo` threw different exception before, but now, I am throwing `PathMissingError`. Unified, but quite verbose.

tests/func/test_api.py
```diff
@@ -47,10 +46,10 @@ def test_get_url_external(erepo_dir, cloud):
 def test_get_url_requires_dvc(tmp_dir, scm):
     tmp_dir.scm_gen({"foo": "foo"}, commit="initial")
 
-    with pytest.raises(UrlNotDvcRepoError, match="not a DVC repository"):
+    with pytest.raises(OutputNotFoundError, match="with output 'foo'"):
```
efiop, line 49: This is changing the API, we shouldn't do it in this PR if we can avoid it.

efiop: Plus `OutputNotFoundError` is an internal thing, that shouldn't be exposed to api users.

tests/func/test_api.py
```diff
@@ -164,12 +163,12 @@ def test_open_not_cached(dvc):
     dvc.run(
         single_stage=True,
         metrics_no_cache=[metric_file],
-        cmd=(f'python -c "{metric_code}"'),
+        cmd=f'python -c "{metric_code}"',
```
efiop, line 166: ?

tests/func/test_api.py
skshetry: We do have to break things here as it's not precise enough. OutputNotFoundError is more of a correct term here.

skshetry:
The suggestions cannot be applied without complete refactoring of the `dvc.external_repo`, was trying to break cycle here, but looks like not. Keeping it as a draft till the refactoring is done.

pared:
Looks good, though I would also say that `DvcTree` belongs in `get_url`.

pared:
sorry @skshetry, clicked in close and comment

tests/func/test_api.py
skshetry: Also, we already threw `OutputNotFoundError`. It was just not documented. The following threw `OutputNotFoundError`.

https://github.com/iterative/dvc/blob/334556f07dc511927543218d4a2a1a1c1c83ed65/dvc/api.py#L31

efiop: @skshetry I think that one was caught by https://github.com/iterative/dvc/blob/master/dvc/external_repo.py#L51 and re-raised properly.

skshetry: Ahh missed that. So, we could raise the same exc here?

efiop: @skshetry But this particular test doesn't catch that one, but rather UrlNotDvcRepoError and it should stay that way. The issue is that we are using RepoTree instead of DvcTree right now, which works for git-only repos, when it shouldn't. Could do something simple like `if not tree.dvc_tree: raise UrlNotDvcRepoError` though, if subrepos are a concern.

skshetry:
Closing as https://github.com/iterative/dvc/pull/4465 has the same changes.

PR #4459, opened 2020-08-24
diff: Add sort logic to sort diff ouput irrespective of output format…

PuneethaPai:
…(default/json/md)

Fixes one of the task mentioned in Issue: #2982 
* [x] consider sorting default output

I have sorted output irrespective of output format to maintain consistency.

### DIFF
--- a/tests/unit/command/test_diff.py
+++ b/tests/unit/command/test_diff.py
@@ -123,7 +123,7 @@ def test_diff_show_md_and_hash(mock_show_md, mocker, caplog, show_hash):
 
     diff = {}
     show_hash = show_hash if show_hash else False
-    mocker.patch("dvc.repo.Repo.diff", return_value=diff)
+    mocker.patch("dvc.repo.Repo.diff", return_value=diff.copy())
 
     assert 0 == cmd.run()
     mock_show_md.assert_called_once_with(diff, show_hash)
--- a/tests/unit/command/test_diff.py
+++ b/tests/unit/command/test_diff.py
@@ -47,11 +47,12 @@ def test_show_hash(mocker, caplog):
         "added": [],
         "deleted": [
             {"path": os.path.join("data", ""), "hash": "XXXXXXXX.dir"},
-            {"path": os.path.join("data", "bar"), "hash": "00000000"},
             {"path": os.path.join("data", "foo"), "hash": "11111111"},
+            {"path": os.path.join("data", "bar"), "hash": "00000000"},
         ],
         "modified": [
-            {"path": "file", "hash": {"old": "AAAAAAAA", "new": "BBBBBBBB"}}
+            {"path": "file2", "hash": {"old": "AAAAAAAA", "new": "BBBBBBBB"}},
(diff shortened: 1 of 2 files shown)


tests/unit/command/test_diff.py
```diff
@@ -162,10 +171,10 @@ def test_show_md():
         "| Status   | Path     |\n"
         "|----------|----------|\n"
         "| added    | file     |\n"
+        "| deleted  | zoo      |\n"
         "| deleted  | data{sep}    |\n"
-        "| deleted  | data{sep}bar |\n"
         "| deleted  | data{sep}foo |\n"
-        "| deleted  | zoo      |\n"
+        "| deleted  | data{sep}bar |\n"
```
efiop, line 177: Hm, are you sure the order is correct? Maybe I'm missing something obvious.

efiop (commented):
Thanks for the PR @PuneethaPai ! :pray: Please see some comments above.

tests/unit/command/test_diff.py
PuneethaPai: I simply edited tests after fixing logic 
Thanks for noticing @efiop . Will check.

PuneethaPai: @efiop I remember now. Since we have already sorted entries in main method CmdDiff.run(), I don't want format specific methods to sort the entries. They take care only of the format to display, without altering any order. I kept the test with unordered entries, explicitly for this and Obviously I forgot 🤦 😄

Main method takes care of sorting, removing hash if not required, etc. Let me know if you want to make it clear. I can think of adding doc string, if required.

Thanks

PuneethaPai:
Just a small change in test code to pass dict copy instead of original `diff` dict.

In python both `list` and `dict` when modified by called function, it gets modified in original function as well. (Pass by value, Pass by reference scenario)

In test I was passing `diff` dict which could have lead to bug, hence now I pass copy of this. (Thus even if the copy gets modified original `diff` dict stays the same.)

tests/unit/command/test_diff.py
efiop: Ah, makes sense! Ok, let's leave it as is. Thank you!

efiop (approved):
Thank you so much @PuneethaPai ! :pray:

jorgeorpinel:
Hi! Does this change the output in the examples of https://dvc.org/doc/command-reference/diff perhaps?

Esp. in [Comparing tags or branches](https://dvc.org/doc/command-reference/diff#example-comparing-tags-or-branches) and [https://dvc.org/doc/command-reference/diff#example-using-different-output-formats](https://dvc.org/doc/command-reference/diff#example-using-different-output-formats) I guess.

efiop:
@jorgeorpinel Good catch! Adjusted in https://github.com/iterative/dvc.org/commit/ee2182aa3c53cea295865adc796c1001587f63bb . Thank you!

jorgeorpinel:
Thanks! That was a quick one 🙂

Issue #3633, opened 2020-04-14
Pipeline variables from params file


dmpetrov:
With the introduction of the new multiple-stage pipeline, we will need to find a way of defining variables in the pipeline. For example, the intermediate file name `cleansed.csv` is used from two stages in the following pipeline and it needs to be defined into a variable:

```yaml
stages:
    process:
        cmd: "./process.bin --input data --output cleansed.csv"
        deps:
             - path: data/
        outs:
             - path: cleansed.csv

    train:
        cmd: "python train.py"
        deps:
             - path: cleansed.csv
             - path: train.py
             - path: params.yaml
                params:
                     lr: 0.042
                     layers: 8
                     classes: 4
        outs:
             - path: model.pkl
             - path: log.csv
                cache: true
             - path: summary.json
```

We need to solve two problems here:
1. Define a variable in one place and reuse it from multiple places/stages.
2. Often users prefer to read file names from config files (like in the `train` stage), not from the command line (like in the `process` stage).

We can solve both of the problems using a single abstraction - `parameters file variable`:
```yaml
stages:
    process:
        cmd: ./process.bin
        outs:
             - path: "params.yaml:cleansed_file_name"
        ....
    train:
        cmd: "python train.py"
        deps:
             - path: "params.yaml:cleansed_file_name"
```

This feature is useful in the current DVC design as well. It is convenient to read file names from params file and still define dependency properly like `dvc run -d params.yaml:input_file -o params.yaml:model.pkl`

karajan1001:
In my case I stored these global variables (vocabulary file, cleaned data file, label name file) in a configuration file as default. And the processing cli can accept all these variables to override the defualt one. But my configuration file parsing and merging is coded inside the program, and had to be re-coded in every program.

tall-josh:
Before I knew about multiple-stage pipelines I used the `params` functionality and hacked together a script to build a richer `dvc run` command. I used the same `dot.notation` you can use in the params command line args ie: `dvc run -p params.yaml:data.output,train ...` In the end my `params.yaml` looks like:

```yaml
data:
    url: s3://mydata.npz
    out_path: data/mydata.npz
data/-o:  # Each element in this list will be
          # added as a  -o <arg>
    # For access strings within the current stage
    # you can also use '.out_path'
    - data.out_path
data/script:
    - python src/cli/data_download.py
    - --data-size
    - .url  # equivalent to data.url
    - --out-path
    - .out_path # equivalent to data.out_path

train:
[...]

PR #4463, opened 2020-08-25
WIP dvc: implement `vars` templating in multistage dvc file

tall-josh:

---

**This is a WIP I have made this PR do get the ball rolling and hopefully get some guidance as it's the first contribution I've made to an opensource project before**

## ToDo

  - [ ] The contribution tasks above
  - [x] Add `dvc render` to the cli
  - [ ] Optionally save the rendered `dvc.yaml`

PR adds templating to the `dvc.yaml` file. This is an effort to reduce duplication of arguments in the `dvc.yaml` and make the multistage dvc file more declarative. An optional `vars` key to the `dvc.yaml` file. The values under this key are rendered into the standard multistage dvcfile that is defined below.

eg:

```yaml
vars:
  project_name: my-project
  train_out: "./training/{{ project_name }}"
  eval_out: "./eval/{{ project_name }}"
  data_dir: my-data
  epochs: 20
  
stages:
  train:
    wdir: .
    cmd: >-
      train_model --out-dir {{ train_out }} --epochs {{ epochs }} --data {{ data_dir }}
    deps:
      - "{{ data_dir }}"
    outs:
      - "{{ train_out }}"
  eval:
    wdir: .
    cmd: >-
        eval_model --out-dir {{ eval_out }} --data {{ data_dir }}
    deps:
      - "{{ data_dir }}"
      - "{{ train_out }}"
    outs:
      - "{{ eval_out }}"
```


dvc/command/render.py
```diff
@@ -0,0 +1,62 @@
+import argparse
+import logging
+
+from dvc.command.base import CmdBase, append_doc_link
+from dvc.utils.serialize._yaml import dumps_yaml, render_dvc_template
+
+logger = logging.getLogger(__name__)
+
+
+class CmdRender(CmdBase):
```
efiop, line 10: What use case do you see for a separate `render` command? I see that you've already implemented automatic rendering, so not quite sure what is the use case for a separate command.

dvc/utils/serialize/_yaml.py
```diff
@@ -3,6 +3,7 @@
 from contextlib import contextmanager
 
 from funcy import reraise
+from jinja2 import Template
```
efiop, line 6: Hm, not quite sure that jinja template is the best format we can use in dvc files, feels like it is still a bit unusual, at least coming from python. It is mature and powerful though, no doubt about it :slightly_smiling_face:

efiop (commented):
Thanks for the contribution @tall-josh ! :pray: Please see some comments above.

dvc/command/render.py
tall-josh: Thanks for the feedback!

Ahhh, this is mainly for a user to debug their `dvc.yaml`. Just to check if the rendered stage and variables are doing what you want them to.

dvc/utils/serialize/_yaml.py
tall-josh: Yeah, I went with this because it was relatively easy to use and mature. We also touched on this in issue #3633. If you/anyone has any recommendations I'd be open to taking a crack at implementing it.

tall-josh: Hey @efiop 

Would a  custom syntax be more acceptable? I have been using this fork quite a lot recently, if we were to go down the path of a custom templating syntax I think the following features are important. I welcome any feedback :-)

  - Support for looping flagged command-line arguments. ie: I currently use this pattern frequently. (I'll show it in Jinja for now)
```yaml
vars:
  repeated_arg:
    - thing
    - another_thing
    - another_other_thing
stages:
  foo:
    wdir: .
    cmd: >-
      python foo.py
      {% for x in repeated_arg %}
      -a "{{ x }}"
      {% endfor %}

# Renders as
vars:
  repeated_arg:
    - thing
    - another_thing
    - another_other_thing
stages:
  foo:
    wdir: .
    cmd: >-
      python foo.py
      -a "thing"
      -a "anothet_thing"
      -a "another_othert_thing"
```
  - Similarly, looping over key-value pairs would be handy. For example, if a command has optional flagged args you could template the function call in the `stage` and simply update the `vars` as need.
  - On a similar vein, supporting a combination of positional and flagged args would be handy. Potentially using a list of tuples, ie:

```
vars:
  mixed_args:
    - pos-arg-1
    - pos-arg-2
    - ['--flag-arg-1', 42]
    - ['--flag-arg-2', 3.14159]
```
  - The templated syntax should stand out so it is easy to identify at a glance
  - Templated `dvc.yaml` should still be valid yaml (not sure about this, but intuitively I like the idea and have been striving for it)
  - Be able to pickup the stdout of a stage and refer to it in subsequent stages. I have not done this but I have seen it in Kubeflow pipelines and like the idea. ie:

```yaml
stages:
  one:
    wdir: .
    cmd: "echo Bailey"
  two:
    wdir: .
    cmd: "echo {{ stages.one.stdout }} is the greatest doggo in the universe!!!"

# After stage one completes stage two renders as

  two:
    wdir: .
    cmd: "echo Bailey is the greatest doggo in the universe!!!"
```
  _Though maybe this is getting a bit distracted from the initial goal_

---

Also pulling in interested parties from #3633

@dsuess @dmpetrov @karajan1001 @skshetry  @elgehelge @jcpsantiago

jcpsantiago:
I very much like this! Would you be able to do `{% for this, that in repeated_args %}` with this setup? i.e. can elements of `repeated_args` be objects?

efiop:
@jcpsantiago Really sorry for the silence, we are having some discussions internally on where we want to move with this. So far we think that jinja is a powerful and popular templating solution :ok_hand: , but just not sure if the syntax fits into what we see for dvc.yaml files. It is also clear that, as you've mentioned above, we will need additional tricks to make the parametrization/hparamsearch scenarios work. We understand that we need to somehow move this forward, just need to dedicate some more time to think about this.

tall-josh:
@efiop 

Thanks for the feedback guys!

@jcpsantiago

As it stands, using Jinja2 currently allows `{% for this, that in repeated_args %}` however I don't think you could then do something like `this.foo` or `that.bar` within the loop. Is that what you mean when you say;

> can elements of `repeated_args` be objects?

Presumably, this is a way to support the  looping over params you mention in the  [#3633 Comment](https://github.com/iterative/dvc/issues/3633#issuecomment-677298728)

---

@efiop

Thanks again for the feedback, great to hear you're considering this. I'll sit tight and tinker until you have time to think about how you want to move on this. Cheers

dvc/command/render.py
```diff
@@ -0,0 +1,62 @@
...
+    def run(self):
+        with open(self.args.path, encoding="utf-8") as fd:
+            text = fd.read()
+        dvc_dict, vars_dict = render_dvc_template(text)
+        vars_dict = {"vars": vars_dict}
+        vars_str = dumps_yaml(vars_dict)
+
+        if self.args.only_vars:
```
karajan1001, line 18: Too many return points in one function. Can we make it more compact?

dvc/utils/serialize/_yaml.py
karajan1001: Haha, I only had used `jinja2` before, used in an inner annotation platform. It is a powerful tool, maybe over powerful and complex here?

dvc/command/render.py
tall-josh: Yeah definitely. This was tied Josh coding :-p

dvc/utils/serialize/_yaml.py
```diff
@@ -14,14 +15,77 @@ def __init__(self, path):
         super().__init__(path, "YAML file structure is corrupted")
 
 
-def load_yaml(path, tree=None):
-    return _load_data(path, parser=parse_yaml, tree=tree)
+def recursive_render(tpl, values, max_passes=100):
```
skshetry, line 18: This is for nested interpolation, right?
```jinja
{{ var {{ var }} }}
```

dvc/utils/serialize/_yaml.py
```diff
@@ -14,14 +15,77 @@ def __init__(self, path):
...
+    for _ in range(max_passes):
+        curr = Template(prev).render(**values)
+        if curr != prev:
+            prev = curr
+        else:
+            return curr
+    raise RecursionError("Max resursion depth reached")
+
+
+def render_vars(dvc_dict):
```
skshetry, line 44: @tall-josh, this interpolates values in `vars` using data from `vars` itself, right?
```jinja
vars:
  foo: 3
  bar: {{ foo }}
```

skshetry: How important is this scenario?

tall-josh: Sorry for the delay. Yeah it does. I have used this a lot. ie:

```yaml
vars:
  my_labeling_project: 123
  data_out: /storage/data/{{ my_labeling_project }}_data
  src: s3://labeled-data/projects/{{ my_labeling_project }}
stages:
  download:
    cmd: python download_data.py --src {{src}} --out  {{ data_out }}
    outs:
      - {{ data_out }}
```

Without this I would have to repeat the `my_labeling_project`

dvc/utils/serialize/_yaml.py
tall-josh: Yep. Not the most efficient thing, just trying to get something working. My team and I have been using this feature in our internal projects. It's been quite nice!

skshetry:
> it's the first contribution I've made to an opensource project before

@tall-josh, thank you so much for contributing and choosing DVC to contribute. I have to decline this PR, as you are already familiar with the plan, and I hope you'd provide us with feedback as always.

Please don't get discouraged by this and I hope to see you with many contributions in terms of feedbacks and code. 

Again, thanks for *getting the ball rolling*. :+1:

PR #4465, opened 2020-08-25
support subrepos for get/list/api/import/ls

### LINKED ISSUE #3180: api: get_url does not support links for granular file
`get_url` does not support granular files access.

Eg:
```sh
mkdir dir
echo "foo" > dir/file

dvc add foo
python -c '
import dvc.api
dvc.api.get_url("dir/file"
'
```

throws following exception:
```
OutputNotFoundError: unable to find DVC-file with output 'dir/file'
```

This also affects #3156 which introduces `dvc get --show-url`.

### LINKED ISSUE #3369: dvc get/import support for subrepos

skshetry:

This adds support for get/list/api/import/ls on a subrepo.

Also fixes #3180, by adding a granular url support for `dvc.api.get_url`
And, of course it fixes #3369.

### DIFF
--- a/dvc/api.py
+++ b/dvc/api.py
@@ -2,16 +2,16 @@ import os
 from contextlib import _GeneratorContextManager as GCM
 from contextlib import contextmanager
 
-from dvc.exceptions import DvcException, NotDvcRepoError
+from funcy import reraise
+
+from dvc.exceptions import (
+    NotDvcRepoError,
+    OutputNotFoundError,
+    PathMissingError,
+)
 from dvc.external_repo import external_repo
+from dvc.path_info import PathInfo
 from dvc.repo import Repo
-
-
-class UrlNotDvcRepoError(DvcException):
-    """Thrown if the given URL is not a DVC repository."""
-
-    def __init__(self, url):
-        super().__init__(f"'{url}' is not a DVC repository.")
 
 
 def get_url(path, repo=None, rev=None, remote=None):
@@ -20,17 +20,22 @@ def get_url(path, repo=None, rev=None, remote=None):
     in a DVC repo. For Git repos, HEAD is used unless a rev argument is
     supplied. The default remote is tried unless a remote argument is supplied.
 
-    Raises UrlNotDvcRepoError if repo is not a DVC project.
+    Raises OutputNotFoundError if the file is not a dvc-tracked file.
 
     NOTE: This function does not check for the actual existence of the file or
     directory in the remote storage.
     """
     with _make_repo(repo, rev=rev) as _repo:
-        if not isinstance(_repo, Repo):
-            raise UrlNotDvcRepoError(_repo.url)  # pylint: disable=no-member
-        out = _repo.find_out_by_relpath(path)
-        remote_obj = _repo.cloud.get_remote(remote)
-        return str(remote_obj.tree.hash_to_path_info(out.hash_info.value))
+        path_info = PathInfo(_repo.root_dir) / path
+        with reraise(FileNotFoundError, PathMissingError(path, repo)):
+            metadata = _repo.repo_tree.metadata(path_info)
+
+        if not metadata.is_dvc:
+            raise OutputNotFoundError(path, repo)
+
+        cloud = metadata.repo.cloud
+        hash_info = _repo.repo_tree.get_hash(path_info)
+        return cloud.get_url_for(remote, checksum=hash_info.value)
 
 
 def open(  # noqa, pylint: disable=redefined-builtin
@@ -97,7 +102,7 @@ def _make_repo(repo_url=None, rev=None):
     repo_url = repo_url or os.getcwd()
     if rev is None and os.path.exists(repo_url):
         try:
-            yield Repo(repo_url)
+            yield Repo(repo_url, subrepos=True)
             return
         except NotDvcRepoError:
             pass  # fallthrough to external_repo
--- a/dvc/dependency/repo.py
+++ b/dvc/dependency/repo.py
@@ -1,9 +1,4 @@
-import os
-
 from voluptuous import Required
-
-from dvc.exceptions import OutputNotFoundError
-from dvc.path_info import PathInfo
 
 from .local import LocalDependency
 
@@ -42,30 +37,16 @@ class RepoDependency(LocalDependency):
     def __str__(self):
         return "{} ({})".format(self.def_path, self.def_repo[self.PARAM_URL])
 
-    def _make_repo(self, *, locked=True):
+    def _make_repo(self, *, locked=True, **kwargs):
         from dvc.external_repo import external_repo
 
         d = self.def_repo
         rev = (d.get("rev_lock") if locked else None) or d.get("rev")
-        return external_repo(d["url"], rev=rev)
+        return external_repo(d["url"], rev=rev, **kwargs)
 
     def _get_hash(self, locked=True):
-        from dvc.tree.repo import RepoTree
-
-        with self._make_repo(locked=locked) as repo:
-            try:
-                return repo.find_out_by_relpath(self.def_path).hash_info
-            except OutputNotFoundError:
-                path = PathInfo(os.path.join(repo.root_dir, self.def_path))
-
-                # we want stream but not fetch, so DVC out directories are
-                # walked, but dir contents is not fetched
-                tree = RepoTree(repo, stream=True)
-
-                # We are polluting our repo cache with some dir listing here
-                if tree.isdir(path):
-                    return self.repo.cache.local.tree.get_hash(path, tree=tree)
-                return tree.get_file_hash(path)
+        with self._make_repo(locked=locked, fetch=False, stream=True) as repo:
+            return repo.get_checksum(self.def_path)
 
     def workspace_status(self):
         current = self._get_hash(locked=True)
@@ -86,14 +67,15 @@ class RepoDependency(LocalDependency):
         return {self.PARAM_PATH: self.def_path, self.PARAM_REPO: self.def_repo}
 
     def download(self, to):
-        with self._make_repo() as repo:
+        cache = self.repo.cache.local
+
+        with self._make_repo(cache_dir=cache.cache_dir) as repo:
             if self.def_repo.get(self.PARAM_REV_LOCK) is None:
                 self.def_repo[self.PARAM_REV_LOCK] = repo.get_rev()
 
-            cache = self.repo.cache.local
-            with repo.use_cache(cache):
-                _, _, cache_infos = repo.fetch_external([self.def_path])
-            cache.checkout(to.path_info, cache_infos[0])
+            _, _, cache_infos = repo.fetch_external([self.def_path])
+
+        cache.checkout(to.path_info, cache_infos[0])
 
     def update(self, rev=None):
         if rev:
--- a/dvc/dependency/repo.py
+++ b/dvc/dependency/repo.py
@@ -45,6 +45,8 @@ class RepoDependency(LocalDependency):
         return external_repo(d["url"], rev=rev, **kwargs)
 
     def _get_hash(self, locked=True):
+        # we want stream but not fetch, so DVC out directories are
+        # walked, but dir contents is not fetched
         with self._make_repo(locked=locked, fetch=False, stream=True) as repo:
             return repo.get_checksum(self.def_path)
 
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -6,10 +6,12 @@ from contextlib import contextmanager
 from distutils.dir_util import copy_tree
 from typing import Iterable
 
-from funcy import cached_property, retry, wrap_with
+from funcy import cached_property, reraise, retry, wrap_with
 
+from dvc.cache import Cache
 from dvc.config import NoRemoteError, NotDvcRepoError
 from dvc.exceptions import (
+    DvcException,
     FileMissingError,
     NoOutputInExternalRepoError,
     NoRemoteInExternalRepoError,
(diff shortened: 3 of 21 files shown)


tests/remotes/__init__.py
```diff
@@ -40,7 +40,7 @@ def docker():
     import os
 
     # See https://travis-ci.community/t/docker-linux-containers-on-windows/301
-    if os.environ.get("TRAVIS") and os.name == "nt":
+    if os.environ.get("TRAVIS") or os.name == "nt":
```
skshetry, line 43: Oops.

dvc/dependency/repo.py
```diff
@@ -42,32 +41,18 @@ def repo_pair(self):
...
-                # We are polluting our repo cache with some dir listing here
-                if tree.isdir(path):
-                    return self.repo.cache.local.tree.get_hash(
-                        path, tree=tree
-                    )[1]
-                return tree.get_file_hash(path)
+        with self._make_repo(locked=locked, fetch=False, stream=True) as repo:
+            path = PathInfo(os.path.join(repo.root_dir, self.def_path))
+            tree = repo.repo_tree
+            return tree.get_hash(path)
```
skshetry, line 55: This does not support subrepos yet, nor does it ignore subrepos which it should.

dvc/external_repo.py
```diff
@@ -139,110 +140,143 @@ def download_update(result):
...
+        self.cache_types = cache_types
         self.tree = self.scm.get_tree(rev)
+        self.tree_confs = kwargs
 
-    @cached_property
-    def scm(self):
-        return Git(self.root_dir)
+        self.config = {"cache": {"dir": self.cache_dir}}
+        self.cache = Cache(self)
+        self.state = StateNoop()
```
skshetry, line 276: The biggest issues/controversial thing could be these two. I tried to unify them though, rather than trying to avoid, as most of the trees already try to access `repo.cache`, which the cache then tries to use `State`.

dvc/tree/repo.py
```diff
@@ -365,7 +365,10 @@ def metadata(self, path):
         if not stat_result and not dvc_meta:
             raise FileNotFoundError
 
-        meta = dvc_meta or Metadata(path_info=path_info)
+        meta = dvc_meta or Metadata(
+            path_info=path_info,
+            repo=self._get_repo(str(path_info)) or self._main_repo,
+        )
```
skshetry, line 372: This is a bit fragile, not sure how it'll hold for subrepos support.

dvc/dependency/repo.py
```diff
@@ -88,14 +73,13 @@ def dumpd(self):
...
+        with self._make_repo(cache_dir=cache.cache_dir) as repo:
             if self.def_repo.get(self.PARAM_REV_LOCK) is None:
                 self.def_repo[self.PARAM_REV_LOCK] = repo.get_rev()
 
-            cache = self.repo.cache.local
-            with repo.use_cache(cache):
-                _, _, cache_infos = repo.fetch_external([self.def_path])
-            cache.checkout(to.path_info, cache_infos[0])
+            _, _, cache_infos = repo.fetch_external([self.def_path])
+        cache.checkout(to.path_info, cache_infos[0])
```
skshetry, line 82: I am not sure if this will cause problems, maybe performance as they use different `State`? cc @efiop

dvc/external_repo.py
```diff
@@ -139,110 +140,143 @@ def download_update(result):
...
 
     def close(self):
-        if "scm" in self.__dict__:
-            self.scm.close()
+        self.scm.close()
```
pared, line 279: Shouldn't we check if scm is not None? Its optional for `__init__`.

dvc/dependency/repo.py
efiop: Performance - not so much, but `dvc get` on a drive that is on nfs/cifs might hang because of the state, hence why we've used Noop.

efiop: Plus, seems like this will also not use the cache.type that is configured in the repo for `dvc import`, that's a bigger problem.

skshetry: cache types are with `cache` as it is what is doing checkout. Plus, we can always pass `cache_types` to `external_repo`.

dvc/external_repo.py
skshetry: Good point @pared. I'll adjust it later.

dvc/external_repo.py
pared: Hmm, maybe we should have something that will be a common ancestor between  `Repo` and `BaseExternalRepo`? 
Modeling `ExternalGitRepo` to partially match the structure of `Repo` is probably not good for the long run.

dvc/dependency/repo.py
skshetry: Let me explain a bit what I was saying. This change makes the `ExternalRepo/ExternalGitRepo` and all the subrepos it discovers to have separate cache and state instance, but you can set the same cache directory and same cache types beforehand.

For imports, host's cache is doing the checkout, external_repo only downloads it to the cache. So, this works with by just setting `cache_dir` on `external_repo()`.

For get, we use `tmp_dir` and use separate cache types which can be passed to the `external_repo()` and it'll share those cache dir and cache types (but have separate instances of Cache/State).

```python
with external_repo(url, cache_dir="", cache_types=["reflink"]):
    pass
```

^

pared (commented):
I think this is a good change, "configuring" external repo from the start feels much better than using nested `with` statements which change existing instance.

dvc/dependency/repo.py
efiop: @skshetry Sounds good.

efiop:
Looks good! But I see that a lot of the changes are not actually about subrepos and could be introduced without mentioning subrepos at all (e.g. that `get_hash()` piece).

skshetry:
@efiop, subrepo support is already there. This PR is more about enabling it, which is not possible without changing APIs of External(Git)Repo, which in turn affects `dependency/repo.py` and friends.

dvc/api.py
```diff
@@ -2,35 +2,40 @@
...
-            raise UrlNotDvcRepoError(_repo.url)  # pylint: disable=no-member
-        out = _repo.find_out_by_relpath(path)
-        remote_obj = _repo.cloud.get_remote(remote)
-        return str(remote_obj.tree.hash_to_path_info(out.checksum))
+        path_info = PathInfo(_repo.root_dir) / path
+        with reraise(FileNotFoundError, PathMissingError(path, repo)):
+            metadata = _repo.repo_tree.metadata(path_info)
+
+        if not metadata.output_exists:
+            raise OutputNotFoundError(path, repo)
```
skshetry, line 34: Will see what I can do here. But I am for not using DvcTree, and throwing OutputNotFoundError.

tests/func/test_api.py
```diff
@@ -47,10 +46,10 @@ def test_get_url_external(erepo_dir, cloud):
 def test_get_url_requires_dvc(tmp_dir, scm):
     tmp_dir.scm_gen({"foo": "foo"}, commit="initial")
 
-    with pytest.raises(UrlNotDvcRepoError, match="not a DVC repository"):
+    with pytest.raises(OutputNotFoundError, match="output 'foo'"):
```
skshetry, line 49: Will try fixing it later.

skshetry:
Tests are pending, will try to add tomorrow. At a quick look, it feels better than before. 

Anyway, except for the tests, there might be only cosmetic changes hereafter. But, depends on the test failures.
Still need to figure out how to handle remote-related failures for `dvc list` as we cannot expect user with multiple repos to provide all of the remotes configurations, so it should try not to visit those repositories.

dvc/api.py
```diff
@@ -2,35 +2,40 @@
...
-            raise UrlNotDvcRepoError(_repo.url)  # pylint: disable=no-member
-        out = _repo.find_out_by_relpath(path)
-        remote_obj = _repo.cloud.get_remote(remote)
-        return str(remote_obj.tree.hash_to_path_info(out.hash_info.value))
+        path_info = PathInfo(_repo.root_dir) / path
+        with reraise(FileNotFoundError, PathMissingError(path, repo)):
+            metadata = _repo.repo_tree.metadata(path_info)
+
+        if not metadata.is_dvc:
+            raise OutputNotFoundError(path, repo)
```
skshetry, line 34: Looked into this, we used to throw `OutputNotFoundError` before as well, so this does not change anything.

Regarding `UrlNotDvcRepoError`, it's not precise to throw now as a git-repo can have subrepo inside of it.

dvc/external_repo.py
```diff
@@ -28,30 +31,44 @@
 logger = logging.getLogger(__name__)
 
 
+class IsADVCRepoError(DvcException):
```
skshetry, line 34: Thrown when the path is a root of the dvc repo during `get/import`.

dvc/external_repo.py
skshetry: @pared, let's duck type, for now. Let's reconsider in the future.

tests/func/test_run_multistage.py
```diff
@@ -306,9 +305,9 @@ def test_run_params_no_exec(tmp_dir, dvc):
         {"outs": ["foo"], "deps": ["bar"], "name": "copy-foo-bar"},
     ],
 )
-def test_run_without_cmd(kwargs):
+def test_run_without_cmd(tmp_dir, dvc, kwargs):
     with pytest.raises(InvalidArgumentError) as exc:
-        Repo().run(**kwargs)
+        dvc.run(**kwargs)
```
skshetry, line 310: Previously, we used to wait for 5s for lock timeout, so this used to work. But, since we reduced the timeout, this has become flakey.

skshetry:
@efiop @pared @pmrowla, this is ready for a review. The only issue right now is with `dvc ls` as it currently throws error when the path is not found in remote or if it cannot connect to a remote. This is a pre-existing issue (exaggerated by subrepo support), which I'll try fixing in a separate PR.

dvc/tree/git.py
```diff
@@ -74,9 +74,10 @@ def dvcignore(self):
         return cls(self, root)
 
     def open(
-        self, path, mode="r", encoding="utf-8"
+        self, path, mode="r", encoding=None
     ):  # pylint: disable=arguments-differ
         assert mode in {"r", "rb"}
+        encoding = encoding or "utf-8"
```
pared, line 80: Hmm, maybe `api.open` should use `utf-8` by default?

pared (changes requested):
LGTM, some minor things.

dvc/tree/git.py
efiop: @pared Ideally - no, it should use your default system encoding. This is how things like pathlib's Path do things. I believe, Saugat just changed this for a better practice for optional args and to match BaseTree.open (there are still some discrepancies in tree methods, that I left untouched when adding these trees).

skshetry: Yeah, it should use default system encoding, I just tried to make `api.open()` work without changing any behavior here.
Something to fix it later.

tests/unit/test_external_repo.py
```diff
@@ -0,0 +1,76 @@
...
+    for repo in subrepos:
+        make_subrepo(repo, erepo_dir.scm)
+
+    for repo in subrepos + [erepo_dir]:
+        with repo.chdir():
+            repo.scm_gen("foo", "foo", commit=f"git add {repo}/foo")
+            repo.dvc_gen("bar", "bar", commit=f"dvc add {repo}/bar")
+
+    with external_repo(str(erepo_dir)) as repo:
+        spy = mocker.patch.object(repo, "make_repo", wraps=repo.make_repo)
```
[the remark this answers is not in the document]
skshetry: @pared, yes we should `spy` most of the time. But, `spy` is not really a replacement for `wraps`, as `wraps` is very powerful in that you can wrap it with a different function. 

Looks like I could use `spy` here, I wanted to check for the return value with `wraps` before but I splitted that into a separate test.

tests/unit/test_external_repo.py
```diff
@@ -0,0 +1,76 @@
...
+        list(repo.repo_tree.walk(repo.root_dir))  # drain
+        assert spy.call_count == len(subrepos)
+
+        paths = [os.path.join(repo.root_dir, path) for path in subrepo_paths]
+        spy.assert_has_calls([call(path) for path in paths], any_order=True)
+
+
+@pytest.mark.parametrize("top_level_dvc", [False, True])
+def test_subrepo_is_constructed_properly(
+    tmp_dir, dvc, scm, mocker, make_tmp_dir, top_level_dvc
```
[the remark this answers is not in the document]
skshetry: Good catch, @pared. The root being a dvc repo or not has to be parametrized. Looks like I added `dvc` there by mistake, 
I'll keep this parametrization and remove `dvc`.

pared (approved):

pmrowla (approved):

PR #4469, opened 2020-08-25
github actions tests: initial

### LINKED ISSUE #4418: migrate from travis to github actions

pared:

Partially solves #4418

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -24,17 +24,19 @@ jobs:
       with:
         python-version: ${{ matrix.pyv }}
     - name: install
-      run: pip install ".[all,tests]"
+      run: |
+        pip install ".[all,tests]"
     - name: setup git
       run: |
         git config --global user.email "dvctester@example.com"
         git config --global user.name "DVC Tester"
     - name: run tests
-      run: python -m tests -n=2
-    - name: run codecov
-      if: github.event.action == 'push'
-      run: |
-        pip install codecov
-        codecov
+      run: python -m tests -n=2 --cov-report=xml
+    - name: upload coverage report
+      if: github.event.name == 'push'
+      uses: codecov/codecov-action@v1.0.13
+      with:
+        file: ./coverage.xml
+        fail_ci_if_error: true
 
 
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -1,0 +1,40 @@
+name: Tests build
+
+on: [push, pull_request]
+
+env:
+  DVC_TEST: "true"
+  HOMEBREW_NO_AUTO_UPDATE: 1
+  SHELL: /bin/bash
+
+jobs:
+  tests:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, windows-2019, macos-10.15]
+        pyv: [3.6, 3.7, 3.8]
+        exclude:
+          - os: windows-2019
+            pyv: 3.8
+    steps:
+    - uses: actions/checkout@v2.3.2
+    - name: Set up Python
+      uses: actions/setup-python@v2.1.2
+      with:
+        python-version: ${{ matrix.pyv }}
+    - name: install
+      run: pip install ".[all,tests]"
+    - name: setup git
+      run: |
+        git config --global user.email "dvctester@example.com"
+        git config --global user.name "DVC Tester"
+    - name: run tests
+      run: python -m tests -n=2
+    - name: run codecov
+      if: github.event.action == 'push'
+      run: |
+        pip install codecov
+        codecov
+
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,32 +32,6 @@ jobs:
     before_install: bash ./scripts/ci/install.sh
     install:
     script: ./scripts/ci/check_patch.sh
-  # test jobs
-  - name: "3.7 on Windows"
-    stage: test
-    os: windows           # Windows 10.0.17134 N/A Build 17134
-    language: shell       # 'language: python' is an error on Travis CI Windows
-    env:
-      - PYTHON_VERSION=3.7.5
-      - PATH=/c/Python37:/c/Python37/Scripts:$PATH
-  - name: "3.8 on Windows"
-    os: windows           # Windows 10.0.17134 N/A Build 17134
-    language: shell       # 'language: python' is an error on Travis CI Windows
-    env:
-      - PYTHON_VERSION=3.8.0
-      - PATH=/c/Python38:/c/Python38/Scripts:$PATH
-  - name: "3.6 on Linux"
-    language: python
-    python: 3.6
-  - name: "3.7 on Linux"
-    language: python
-    python: 3.7
-  - name: "3.8 on Linux"
-    language: python
-    python: 3.8
-  - name: "3.7 on Mac"
-    os: osx
-    osx_image: xcode11.3
   # build jobs
   - name: "Mac pkg"
     stage: build
--- a/tests/func/test_add.py
+++ b/tests/func/test_add.py
@@ -208,9 +208,14 @@ def test_add_file_in_dir(tmp_dir, dvc):
         ),
         (pytest.lazy_fixture("gs"), "md5", "8c7dd922ad47494fc02c388e12c00eac"),
         (
-            pytest.lazy_fixture("hdfs"),
-            "checksum",
-            "000002000000000000000000a86fe4d846edc1bf4c355cb6112f141e",
+            pytest.param(
+                pytest.lazy_fixture("hdfs"),
+                "checksum",
+                "000002000000000000000000a86fe4d846edc1bf4c355cb6112f141e",
+                marks=pytest.mark.xfail(
+                    reason="https://github.com/iterative/dvc/issues/4418"
+                ),
+            )
         ),
     ],
     indirect=["workspace"],
--- a/tests/func/test_api.py
+++ b/tests/func/test_api.py
@@ -4,7 +4,7 @@ import platform
 import pytest
 
 from dvc import api
-from dvc.exceptions import FileMissingError, OutputNotFoundError
+from dvc.exceptions import DvcException, FileMissingError, OutputNotFoundError
 from dvc.path_info import URLInfo
 from dvc.utils.fs import remove
 from tests.unit.tree.test_repo import make_subrepo
@@ -91,7 +91,7 @@ def test_open(tmp_dir, dvc, remote):
         pytest.param(
             pytest.lazy_fixture("ssh"),
             marks=pytest.mark.xfail(
-                raises=FileNotFoundError,
+                raises=DvcException,
                 condition=platform.system() == "Windows",
                 reason="https://github.com/iterative/dvc/issues/4418",
             ),
--- a/tests/func/test_api.py
+++ b/tests/func/test_api.py
@@ -1,4 +1,5 @@
 import os
+import platform
 
 import pytest
 
@@ -15,10 +16,16 @@ cloud_names = [
     "gdrive",
     "oss",
     "ssh",
-    "hdfs",
     "http",
 ]
-clouds = [pytest.lazy_fixture(cloud) for cloud in cloud_names]
+clouds = [pytest.lazy_fixture(cloud) for cloud in cloud_names] + [
+    pytest.param(
+        pytest.lazy_fixture("hdfs"),
+        marks=pytest.mark.xfail(
+            reason="https://github.com/iterative/dvc/issues/4418"
+        ),
+    )
+]
 all_clouds = [pytest.lazy_fixture("local_cloud")] + clouds
 
 # `lazy_fixture` is confusing pylint, pylint: disable=unused-argument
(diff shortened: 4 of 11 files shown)


skshetry:
```
E       AssertionError: assert {'version.dvc...available'}}]} == {'dir.dvc': [...available'}}]}
E         Omitting 1 identical items, use -vv to show
E         Right contains 1 more item:
E         {'dir.dvc': [{'changed deps': {'dir (/tmp/pytest-of-runner/pytest-0/popen-gw2/erepo25)': 'update '
E                                                                                                  'available'}}]}
E         Full diff:
E           {
E         -  'dir.dvc': [{'changed deps': {'dir (/tmp/pytest-of-runner/pytest-0/popen-gw2/erepo25)': 'update '
E         -                                                                                          'available'}}],
E            'version.dvc': [{'changed deps': {'version (/tmp/pytest-of-runner/pytest-0/popen-gw2/erepo25)': 'update '
E                                                                                                            'available'}}],
E           }
```

This test fails from time to time. It's very hard to reproduce, and I have tried to debug on failure, but haven't had much luck.

Once when I was able to reproduce, I debugged through it and found that somehow the "dir.dvc"'s `Output::checksum` gets changed when calculating the checksum, so when it later compares, it does not find any differences at all.
But I couldn't figure out why/when that happened.

pared:
@skshetry thanks for this info, that saved me a lot of time for sure :)

tests/func/test_repro.py
```diff
@@ -829,19 +829,19 @@ def test(self):
         self.dvc.run(
             fname=stage,
             outs=[fname],
-            cmd=f"echo $SHELL > {fname}",
+            cmd=f"echo $HOME > {fname}",
```
efiop, line 832: For the record: discussed that this change pretty much breaks the test, there should be another way to make it work in GHA.

tests/func/test_update.py
```diff
@@ -44,6 +44,8 @@ def test_update_import(tmp_dir, dvc, erepo_dir, cached):
 
     assert old_rev != new_rev
 
+    logger = logging.getLogger(__name__)
+    logger.error("#### FUN BEGINS #####")
```
efiop, line 48: If you are only having troubles with this test (and maybe the other one about SHELL), feel free to just mark them as optional and create an issue for it to look right after GHA migration. The main point of this PR is to migrate to GHA, so no reason to hold it just because of those few tests, as we can always just fix them right after.

tests/remotes/__init__.py
```diff
@@ -40,7 +40,7 @@ def docker():
     import os
 
     # See https://travis-ci.community/t/docker-linux-containers-on-windows/301
-    if os.environ.get("TRAVIS") and os.name == "nt":
+    if os.environ.get("GITHUB_ACTION") and os.name == "nt":
```
skshetry, line 43: @pared, use `CI`, this way it will be CI/CD-agnostic and easier to migrate in future.
```suggestion
    if os.environ.get("CI") and os.name == "nt":
```

pared: Good idea, thoguh I wonder whether we should have any check besides "nt". Docker itself does not allow pulling linux images on windows, so its not like temporary check.

pared: Interestingly, windows doc's seems to tell different story than tests log:
https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/linux-containers#:~:text=Docker%20has%20been%20able%20to,machine%20running%20on%20Hyper%2DV.
maybe there needs to be some configuration done to be able to run `docker` fixture on our CI pipeline.

efiop:
This is great :fire:

Oh, I see only windows 3.8 fails. Could create a p0 ticket and just skip 3.8 in this PR for now. Could  also mark it as "allow_failure" so we have a constant reminder :slightly_smiling_face:

tests/func/test_cache.py
```diff
@@ -164,6 +165,10 @@ def test_abs_path(self):
         self.assertEqual(config["cache"]["dir"], dname.replace("\\", "/"))
 
     def test_relative_path(self):
+        # Fixme
```
efiop, line 168: Let's use consistent comments that are saying that this is temporarily disabled and have a link to the github issue.

tests/func/test_cache.py
```diff
@@ -164,6 +165,10 @@ def test_abs_path(self):
...
+        if platform.system() == "Darwin":
+            return
```
efiop, line 170: Also, better to use `@pytest.skipif` than to write this by hand :slightly_smiling_face:

.github/workflows/build.yaml
```diff
@@ -0,0 +1,31 @@
...
+  HOMEBREW_NO_AUTO_UPDATE: 1
+  SHELL: /bin/bash
+
+jobs:
+  tests:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, windows-2019, macos-10.15]
+        pyv: [3.6, 3.7, 3.8]
```
efiop, line 16: THis looks great. So we can get rid of some of our ugly scripts/ci, right? Doesn't look like you use them at all, actually.

tests/func/test_api.py
```diff
@@ -83,6 +85,10 @@ def test_open(tmp_dir, dvc, remote):
     ],
 )
 def test_open_external(erepo_dir, cloud):
+    if isinstance(cloud, SSHMocked) and platform.system() == "Windows":
```
efiop, line 88: You could just use `pytest.param(pytest.lazy_fixture("ssh"), marks=pytest.mark.xfail(...))` in `@parametrize` above.

.github/workflows/build.yaml
pared: I'll get rid of them, just trying to make it work now

.travis.yml
```diff
@@ -23,41 +21,6 @@ stages:
...
-    python: 3.6
-  - name: "3.7 on Linux"
-    language: python
-    python: 3.7
-  - name: "3.8 on Linux"
-    language: python
-    python: 3.8
-  - name: "3.7 on Mac"
-    os: osx
-    osx_image: xcode11.3
```
pared, line 60: Maybe we should leave travis builds until we make all tests on GHA work?

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,35 @@
...
+      with:
+        python-version: ${{ matrix.pyv }}
+    - name: install
+      run: pip install ".[all,tests]"
+    - name: setup git
+      run: |
+        git config --global user.email "dvctester@example.com"
+        git config --global user.name "DVC Tester"
+    - name: run tests
+      run: python -m tests -n=2
```
skshetry, line 33: `codecov` is missing.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - name: run codecov
+      if: github.event.action == 'push'
+      run: |
+        pip install codecov
+        codecov
```
pared, line 38: @efiop could you set `CODECOV_TOKEN`?

efiop: Added to master, should be set.

tests/func/test_add.py
```diff
@@ -208,9 +208,13 @@ def test_add_file_in_dir(tmp_dir, dvc):
         ),
         (pytest.lazy_fixture("gs"), "md5", "8c7dd922ad47494fc02c388e12c00eac"),
         (
-            pytest.lazy_fixture("hdfs"),
-            "checksum",
-            "000002000000000000000000a86fe4d846edc1bf4c355cb6112f141e",
+            # FIXME https://github.com/iterative/dvc/pull/4469
```
efiop, line 211: You are linking PR and not an issue with all the checkboxed for all the failures in this PR.

tests/func/test_add.py
```diff
@@ -208,9 +208,13 @@ def test_add_file_in_dir(tmp_dir, dvc):
...
+            pytest.param(
+                pytest.lazy_fixture("hdfs"),
+                "checksum",
+                "000002000000000000000000a86fe4d846edc1bf4c355cb6112f141e",
+                marks=pytest.mark.xfail(reason="GHA fail"),
```
efiop, line 216: Not informative, could've added url here instead of comment.

tests/func/test_api.py
```diff
@@ -69,20 +74,31 @@ def test_open(tmp_dir, dvc, remote):
...
+                "hdfs",
+                "http",
+            ]
+        ],
+        pytest.param(
+            pytest.lazy_fixture("ssh"),
+            marks=pytest.mark.xfail(
+                raises=FileNotFoundError, condition="sys.platform == 'win32'"
+            ),
+        ),
```
efiop, line 94: This happens only in this PR, but there is no sign of that in comments/reasons.

efiop: Is there a master ticket with all the things that need to be fixed?

tests/func/test_api.py
efiop: ping

tests/func/test_cache.py
```diff
@@ -164,6 +165,10 @@ def test_abs_path(self):
         self.assertEqual(config["cache"]["dir"], dname.replace("\\", "/"))
 
     def test_relative_path(self):
+        # FIXME https://github.com/iterative/dvc/issues/4418
+        if platform.system() == "Darwin":
+            pytest.skip("Failing on Github actions")
+
```
efiop, line 171: Same as https://github.com/iterative/dvc/pull/4469/files#r482911109

tests/func/test_import_url.py
```diff
@@ -115,22 +115,26 @@ def test_import_url_with_no_exec(tmp_dir, dvc, erepo_dir):
     assert not dst.exists()
 
 
-@pytest.mark.parametrize(
-    "workspace",
-    [
-        pytest.lazy_fixture("local_cloud"),
-        pytest.lazy_fixture("s3"),
-        pytest.lazy_fixture("gs"),
+import_clouds = [
```
efiop, line 118: Not sure I understand the point of this change.

tests/func/test_run_multistage.py
```diff
@@ -386,10 +386,14 @@ def test_run_overwrite_preserves_meta_and_comment(tmp_dir, dvc, run_copy):
...
-            "0000020000000000000000003dba826b9be9c6a8e2f8310a770555c4",
-            "00000200000000000000000075433c81259d3c38e364b348af52e84d",
+            # FIXME https://github.com/iterative/dvc/pull/4469
+            pytest.param(
+                pytest.lazy_fixture("hdfs"),
+                "checksum",
+                "0000020000000000000000003dba826b9be9c6a8e2f8310a770555c4",
+                "00000200000000000000000075433c81259d3c38e364b348af52e84d",
+                marks=pytest.mark.xfail(reason="GHA fail"),
+            )
```
efiop, line 396: https://github.com/iterative/dvc/pull/4469/files#r482911109

tests/func/test_update.py
```diff
@@ -8,6 +8,11 @@
 
 @pytest.mark.parametrize("cached", [True, False])
 def test_update_import(tmp_dir, dvc, erepo_dir, cached):
+
+    # FIXME https://github.com/iterative/dvc/issues/4418
+    if not cached:
+        pytest.skip("Failing on Github actions")
+
```
efiop, line 15: https://github.com/iterative/dvc/pull/4469/files#r482911109

tests/unit/output/test_local.py
```diff
@@ -26,6 +28,11 @@ def test_save_missing(self):
 
 
 def test_str_workdir_outside_repo(erepo_dir):
+
+    # FIXME https://github.com/iterative/dvc/issues/4418
+    if platform.system() == "Windows":
+        pytest.skip("Failing on Github ations")
+
```
efiop, line 35: https://github.com/iterative/dvc/pull/4469/files#r482911109

.github/workflows/tests.yaml
efiop: @skshetry pointed out that this is not the best idea to add it publically (which I did), so now I've added a CODECOV_TOKEN secret. Should be usable now.

skshetry: @pared, [`codecov-action`](https://github.com/codecov/codecov-action) supports tokenless uploads using Github Actions, only caveat being we need to change report format to xml: https://github.com/codecov/codecov-action#example-workflowyml-with-codecov-action

.travis.yml
skshetry: Let's not remove that till we are running all of the tests, we might introduce regressions and not even know it.

.github/workflows/tests.yaml
pared: @efiop , sorry I meant to set it as repository secret, should have explained that

tests/func/test_api.py
pared: sorry, I did mark in  parametrization and did not remove this check

tests/func/test_api.py
pared: I will include the list in original issue

.travis.yml
pared: @skshetry after talk with @efiop we decided that we could skip those builds, as long as we keep the original Issue open and focus on fixing leftover tests.

skshetry: Up to you, but I'll still prefer to use Travis as a source of _truth_ for CI, as it might still take a week or more for that to complete. Anyway, the package builds will still be running there.

efiop:
@pared Is this WIP?

.travis.yml
efiop: @skshetry you are right, that would be ideal, but we are starting to have more issues with travis and fixing both seems like a bit too much. If we can make gha work reasonably well, then we could just get rid of travis and focus on final gha adjustments.

pared:
@efiop yes

.travis.yml
skshetry: @efiop, I'm just stating my preference, but if it takes too much effort, let's get rid of it then.

efiop:
@pared Looks like tests were flaky. Maybe let's merge and work on top?

efiop:
Eager to start adding other useful things like pre-commit update hooks, dependabot and so on :slightly_smiling_face: And packages, of course :slightly_smiling_face:

efiop:
This is a great start, let's work on top. Thank you @pared ! :pray:

PR #2957, opened 2019-12-15
[WIP][Showcase] dvc: use dynamic scope to stop callback/flag passing

Suor:
The idea is to simplify function signatures and make intermediate calls not need to know when some high-level code is talking some low-level one. 

Showcased on annoying enough `no_progress_bar` and `progress_callback`. Might be also used to simplify:
- combining pbars
- jobs
- force
- other cli or config options only low level code needs to know about

PR #4470, opened 2020-08-25
status: no warnings for missing state

### LINKED ISSUE #4383: status: ignoring cache that is not in local cache nor in remote
`dvc status -c` should probably exit with non-zero code if some files are not presented neither in cache, neither in remote.

It is possible that some files are not presented neither in local cache, neither in remote. In this case `dvc status -c` will print out warning and exit with zero code:

```
$dvc status -c
name: data.csv, md5: 746308829575e17c3331bbcb00c0898b
Data and pipelines are up to date.
```

It may happen if Alice executes `dvc add data.csv`, `git add data.csv`, `git commit` and `git push` but doesn't execute `dvc push` and Bob executes `dvc status -c data.csv` on another machine.

There should be a way in such case to exit with non-zero from `dvc status` command. It will allow to handle such errors in different scripts.

Context: https://discordapp.com/channels/485586884165107732/563406153334128681/743159360745898044

nik123:
Fixes #4383 (again)

After  PR #4398 it occured, that `dvc status -c` output is bloated and informs about "missing" files twice (took from https://github.com/iterative/dvc/pull/4398#issuecomment-678673973):

> ```
> $ dvc status -c -R .
> WARNING: Some of the cache files do not exist neither locally nor on remote. Missing cache files:
> name: data-in-cache2-not-in-remote.txt, md5: 599b9eea7a2c4f97be18c077d2df4789
> 	new:                data-in-cache-1-not-in-remote.txt
> 	missing:            data-in-cache-2-not-in-remote.txt
> ```
>If `missing` is now an expected status, the WARNING is probably no longer needed. So it could just output something like:
>
> ```
> $ dvc status -c  # Assumes default remote exists, otherwise -r myremote
> new:                data-in-cache-1-not-in-remote.txt
> missing:            data-in-cache2-not-in-remote.txt
> ```

The WARNING shouldn't shown for `dvc status -c`. However it should be shown for all other commands (`fetch`, `gc`, etc.). This PR implements such behavior.

**P.R.:** I didn't come up with anything better than adding `log_missing=True` param into `dvc.data_cloud.DataCloud.status` and passing it all the way down `dvc.cache.local._make_status`. It's ugly. If there any better way please inform me about it. 

issue in dvc.or repo already exist: https://github.com/iterative/dvc.org/issues/1701

### DIFF
--- a/dvc/cache/local.py
+++ b/dvc/cache/local.py
@@ -114,6 +114,7 @@ class LocalCache(CloudCache):
         jobs=None,
         show_checksums=False,
         download=False,
+        log_missing=True,
     ):
         # Return flattened dict containing all status info
         dir_status, file_status, _ = self._status(
@@ -122,6 +123,7 @@ class LocalCache(CloudCache):
             jobs=jobs,
             show_checksums=show_checksums,
             download=download,
+            log_missing=log_missing,
         )
         return dict(dir_status, **file_status)
 
@@ -132,6 +134,7 @@ class LocalCache(CloudCache):
         jobs=None,
         show_checksums=False,
         download=False,
+        log_missing=True,
     ):
         """Return a tuple of (dir_status_info, file_status_info, dir_contents).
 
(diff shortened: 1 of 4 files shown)


pared:
@nik123 the change looks good. I agree we should probably avoid passing down the flag, though I do not see any alternative as of now. Maybe its time we revisit https://github.com/iterative/dvc/pull/2957

pared (approved):

dvc/cache/local.py
```diff
@@ -114,6 +114,7 @@ def status(
         jobs=None,
         show_checksums=False,
         download=False,
+        log_missing=True,
```
efiop, line 117: We've been thinking about revisit logging/info interface for a long time, since passing a chain of such flags is not the prettiest thing ever :slightly_frowning_face: But there is no simple better way for now, so this will work for now :+1:

efiop (approved):
Thank you @nik123 !

PR #4471, opened 2020-08-26
tests: speedup lock tests

skshetry:
Patched to reduce timeout from default 5s to 0.01s

#### Before
```console
$ pytest tests/func/test_lock.py
2 passed in 10.20s
```

#### After
```console
$ pytest tests/func/test_lock.py
2 passed in 0.20s
```

### DIFF
--- a/tests/func/test_lock.py
+++ b/tests/func/test_lock.py
@@ -1,23 +1,23 @@
-import os
+import pytest
 
-from dvc.lock import Lock, LockError
+from dvc.lock import FAILED_TO_LOCK_MESSAGE, Lock, LockError
 from dvc.main import main
-from tests.basic_env import TestDvc
 
 
-class TestLock(TestDvc):
-    def test_with(self):
-        lockfile = os.path.join(self.dvc.dvc_dir, "tmp", "lock")
-        lock = Lock(lockfile)
-        with lock:
-            with self.assertRaises(LockError):
-                lock2 = Lock(lockfile)
-                with lock2:
-                    pass
+def test_with(tmp_dir, dvc, mocker):
+    # patching to speedup tests
+    mocker.patch("dvc.lock.DEFAULT_TIMEOUT", 0.1)
 
-    def test_cli(self):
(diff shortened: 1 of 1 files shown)


skshetry:
@efiop, why is there a 5s sleep? I see that it is from a very old commit 068f7428522f95c2. Is there any reason why we don't retry it in a 0.1s or so?

https://github.com/iterative/dvc/blob/bc7d952e88553eda33bb377a333ab018a1e854ef/dvc/lock.py#L58-L63

tests/func/test_lock.py
```diff
@@ -1,23 +1,23 @@
...
-        lockfile = os.path.join(self.dvc.dvc_dir, "tmp", "lock")
-        lock = Lock(lockfile)
-        with lock:
-            with self.assertRaises(LockError):
-                lock2 = Lock(lockfile)
-                with lock2:
-                    pass
+def test_with(tmp_dir, dvc, mocker):
+    # patching to speedup tests
+    mocker.patch("dvc.lock.DEFAULT_TIMEOUT", 0.01)
```
skshetry, line 9: Reduced it further to 0.01.

pared (commented):
LGTM, though I have to admit that I don't recall why we had 5s timeout.

efiop:
@skshetry Indeed, we could reduce the timeout, but increase the number of tries. 5s was just a random number that we thought was ok to sleep for.

tests/func/test_lock.py
efiop: So maybe let's just adjust it in the code as discussed above?

skshetry: This patch will be required after that too unless we decrease the timeout to a very low value.

Let's say we reduced the timeout to 2s. As these tests are negative tests, they will wait for 2s anyway.

efiop: @skshetry Good point, so we will have to patch both timeout and the number of tries.

efiop (approved):

PR #4479, opened 2020-08-26
version: improve link type detection fall back

### LINKED ISSUE #2788: version: improve link type detection fall back
**To reproduce:**

`dvc version` outside of any DVC repo.

**Output:**

```WARNING: Unable to detect supported link types, as cache directory '.dvc/cache' doesn't exist. It is usually auto-created by commands such as `dvc add/fetch/pull/run/import`, but you could create it manually to enable this check.```

It should instead generate a regular output and either omit the link type completely or print some special value "link: (run inside repo)"

### PR DESCRIPTION
**Pull request by IvanRubanov**
Create temporary file in cache dir. And use it to detect the supported cache types. Clear after cache type detection.

Fixes #2788

### DIFF
--- a/dvc/command/version.py
+++ b/dvc/command/version.py
@@ -59,9 +59,7 @@ class CmdVersion(CmdBaseNoRepo):
                     fs_type = self.get_fs_type(repo.cache.local.cache_dir)
                     info.append(f"Cache directory: {fs_type}")
             else:
-                info.append(
-                    "Cache types: " + error_link("no-dvc-cache")
-                )
+                info.append("Cache types: " + error_link("no-dvc-cache"))
 
         except NotDvcRepoError:
             root_directory = os.getcwd()
--- a/dvc/command/version.py
+++ b/dvc/command/version.py
@@ -10,7 +10,7 @@ from dvc.command.base import CmdBaseNoRepo, append_doc_link
 from dvc.exceptions import DvcException, NotDvcRepoError
 from dvc.scm.base import SCMError
 from dvc.system import System
-from dvc.utils import relpath
+from dvc.utils import error_link
 from dvc.utils.pkg import PKG
 from dvc.version import __version__
 
@@ -59,12 +59,8 @@ class CmdVersion(CmdBaseNoRepo):
                     fs_type = self.get_fs_type(repo.cache.local.cache_dir)
                     info.append(f"Cache directory: {fs_type}")
             else:
-                logger.warning(
-                    "Unable to detect supported link types, as cache "
-                    "directory '{}' doesn't exist. It is usually auto-created "
-                    "by commands such as `dvc add/fetch/pull/run/import`, "
-                    "but you could create it manually to enable this "
-                    "check.".format(relpath(repo.cache.local.cache_dir))
+                info.append(
+                    "Cache types: " + error_link("unable-to-detect-cache-type")
                 )
 
         except NotDvcRepoError:
--- a/dvc/command/version.py
+++ b/dvc/command/version.py
@@ -60,7 +60,7 @@ class CmdVersion(CmdBaseNoRepo):
                     info.append(f"Cache directory: {fs_type}")
             else:
                 info.append(
-                    "Cache types: " + error_link("unable-to-detect-cache-type")
+                    "Cache types: " + error_link("no-dvc-cache")
                 )
 
         except NotDvcRepoError:

### CODE REVIEW

File: dvc/command/version.py [hunk 1]
```diff
@@ -49,23 +50,24 @@ def run(self):
...
+
+            info.append(
+                "Cache types: {}".format(self.get_linktype_support_info(repo))
+            )
+            if psutil:
+                fs_type = self.get_fs_type(repo.cache.local.cache_dir)
+                info.append(f"Cache directory: {fs_type}")
+
+            if create_tmp_cache_dir:
+                shutil.rmtree(repo.cache.local.cache_dir, ignore_errors=True)
```
efiop, line 70: A bit worried about this racing in shared cache scenarios, where cache dir might be accidentally deleted. Also, we might simply not have any rights to create that cache_dir, so btter to wrap `makedirs` above in a try&except logic. Also, we will be leaving traces in a form of parent directories that didn't exist before. But if we are not able to create even a temporary dir, then we will still need to show some message about it, which is actually the same thing as was suggested in the original ticket by @shcheklein as a first option. That makes me wonder if we really shouldn't overcomplicate this and just stick to the first option and just not even try to create a temporary directory.

efiop (commented):
Hi @IvanRubanov !

Thank you so much for contributing! :pray: Please see a comment above.

dvc/command/version.py
IvanRubanov: Thanks for detailed explanation about shared cache scenario.
Just to be sure I understood your message correctly. The final solution is to remove warning completely. No new code required. Is it correct?

efiop: @IvanRubanov Yes, I think removing it and, for example, adding
```
Cache types: https://error.dvc.org/unable-to-detect-cache-type
```
to the `dvc version`  output would be the best overall solution here in terms of complexity and effectiveness. Thank you so much for looking into this!

IvanRubanov:
@efiop could you check this PR once again?

dvc/command/version.py
```diff
@@ -59,12 +58,9 @@ def run(self):
...
             else:
-                logger.warning(
-                    "Unable to detect supported link types, as cache "
-                    "directory '{}' doesn't exist. It is usually auto-created "
-                    "by commands such as `dvc add/fetch/pull/run/import`, "
-                    "but you could create it manually to enable this "
-                    "check.".format(relpath(repo.cache.local.cache_dir))
+                info.append(
+                    "Cache types: "
+                    + "https://error.dvc.org/unable-to-detect-cache-type"
```
efiop, line 63:
```suggestion
                info.append(
                    "Cache types: " + error_link("unable-to-detect-cache-type")
```
error_link helper is defined in `dvc.utils`. Could you also add such a section to https://dvc.org/doc/user-guide/troubleshooting, please? There is an `edit` button at the bottom right corner.

dvc/command/version.py
IvanRubanov: Done.

dvc/command/version.py
IvanRubanov: Done. Updated PR + created new PR for web site documentation.

efiop (approved):
Thank you @IvanRubanov ! :pray:

dvc/command/version.py
```diff
@@ -59,12 +59,8 @@ def run(self):
...
                     info.append(f"Cache directory: {fs_type}")
             else:
-                logger.warning(
-                    "Unable to detect supported link types, as cache "
-                    "directory '{}' doesn't exist. It is usually auto-created "
-                    "by commands such as `dvc add/fetch/pull/run/import`, "
-                    "but you could create it manually to enable this "
-                    "check.".format(relpath(repo.cache.local.cache_dir))
+                info.append(
+                    "Cache types: " + error_link("unable-to-detect-cache-type")
```
jorgeorpinel, line 63: Hi! Should some info on what the problem is be presented to the user at least? Other than just `Cache types: http://error...` Maybe I'm not seeing the full output just from this code.

jorgeorpinel: Also could we shorten the link path a bit? Maybe `"no-dvc-cache"`

jorgeorpinel: (Please see also https://github.com/iterative/dvc.org/pull/1751#pullrequestreview-480112616.)

efiop: @jorgeorpinel Adding more info will defeat the purpose of this PR, will make the desc too long.

efiop, line 63:
```suggestion
                    "Cache types: " + error_link("no-dvc-cache")
```

efiop:
Thank you @IvanRubanov ! :pray:

PR #4480, opened 2020-08-27
s3: provide more helpful messages on common errors

### LINKED ISSUE #4478: s3: better error on missing credentials
A colleague was recently onboarding to our dvc project and received this error:
```
$ dvc pull
ERROR: unexpected error - Unable to locate credentials
```

I believe that if this was a better error message, they may have figured it out on their own.  It wasn't clear that they were missing **AWS** credentials when running a **dvc** command.  Maybe you could say something more like "Unable to access s3://... (the remote url).  Do you have a network connection and have the proper credentials set up?"  which (I think) would address the problem more generally for various remote providers while giving a little more information to the user to figure it out.

efiop:
Fix #4478

### DIFF
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -162,7 +162,7 @@ SCHEMA = {
                     "endpointurl": str,
                     "access_key_id": str,
                     "secret_access_key": str,
-                    Optional("listobjects", default=False): Bool,
+                    Optional("listobjects", default=False): Bool,  # obsoleted
                     Optional("use_ssl", default=True): Bool,
                     "sse": str,
                     "sse_kms_key_id": str,
--- a/dvc/tree/s3.py
+++ b/dvc/tree/s3.py
@@ -94,7 +94,7 @@ class S3Tree(BaseTree):
             link = error_link("connection-error")
             name = self.endpoint_url or "AWS S3"
             raise DvcException(
-                f"Unable to connect to '{name}'. Are you online? {link}"
+                f"Unable to connect to '{name}'. {link}"
             ) from exc
 
     @contextmanager
--- a/dvc/tree/s3.py
+++ b/dvc/tree/s3.py
@@ -10,7 +10,7 @@ from dvc.exceptions import DvcException, ETagMismatchError
 from dvc.path_info import CloudURLInfo
 from dvc.progress import Tqdm
 from dvc.scheme import Schemes
-from dvc.utils import format_link
+from dvc.utils import error_link
 
 from .base import BaseTree
 
@@ -86,12 +86,12 @@ class S3Tree(BaseTree):
         try:
             yield self.s3
         except NoCredentialsError as exc:
-            link = format_link("https://error.dvc.org/no-credentials")
+            link = error_link("no-credentials")
             raise DvcException(
                 f"Unable to find AWS credentials. {link}"
             ) from exc
         except EndpointConnectionError as exc:
-            link = format_link("https://error.dvc.org/connection-error")
+            link = error_link("connection-error")
             name = self.endpoint_url or "AWS S3"
             raise DvcException(
                 f"Unable to connect to {name}. Are you connected to the "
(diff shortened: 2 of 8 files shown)


dvc/config.py
```diff
@@ -162,7 +162,7 @@ class RelPath(str):
                     "endpointurl": str,
                     "access_key_id": str,
                     "secret_access_key": str,
-                    Optional("listobjects", default=False): Bool,
+                    Optional("listobjects", default=False): Bool,  # obsoleted
```
efiop, line 165: Obsoleted since we are using a high-level resource API now, which just does the switch automatically.

dvc/tree/s3.py
```diff
@@ -75,24 +73,52 @@ def s3(self):
 
         session = boto3.session.Session(**session_opts)
 
-        return session.client(
+        return session.resource(
```
efiop, line 76: `resource` is a higher level API, so we don't have to manually parse-out HTTP codes.

dvc/tree/s3.py
```diff
@@ -75,24 +73,52 @@ def s3(self):
...
+            link = format_link("https://error.dvc.org/no-credentials")
+            raise DvcException(
+                f"Unable to find AWS credentials. {link}"
+            ) from exc
+        except EndpointConnectionError as exc:
+            link = format_link("https://error.dvc.org/connection-error")
+            name = self.endpoint_url or "AWS S3"
+            raise DvcException(
+                f"Unable to connect to {name}. Are you connected to the "
+                f"internet? {link}"
```
jorgeorpinel, line 99:
```suggestion
                f"Unable to connect to '{name}'. Are you online?"
```

dvc/tree/s3.py
```diff
@@ -75,24 +72,50 @@ def s3(self):
...
+                    f"Bucket '{bucket}' does not exist. {link}"
+                ) from exc
+
+    @contextmanager
+    def _get_obj(self, path_info):
+        with self._get_bucket(path_info.bucket) as bucket:
+            try:
+                yield bucket.Object(path_info.path)
+            except bucket.meta.client.exceptions.NoSuchKey as exc:
+                raise DvcException(f"{path_info.url} does not exist") from exc
```
efiop, line 118: We'll move to FileNotFoundError in the future, but for now just doing the same thing that we used to.

dvc/tree/s3.py
```diff
@@ -75,24 +72,49 @@ def s3(self):
...
+        )
 
         try:
-            obj = s3.head_object(Bucket=bucket, Key=path, *args, **kwargs)
-        except Exception as exc:
-            raise DvcException(f"s3://{bucket}/{path} does not exist") from exc
-        return obj
+            yield self.s3
+        except NoCredentialsError as exc:
+            link = error_link("no-credentials")
```
jorgeorpinel, line 89:
```suggestion
            link = error_link("no-aws-credentials")
```

efiop: @jorgeorpinel Actually keeping it generic so that section could be expanded with recipes for gs/azure/etc.

dvc/tree/s3.py
```diff
@@ -75,24 +72,49 @@ def s3(self):
...
+            raise DvcException(
+                f"Unable to find AWS credentials. {link}"
+            ) from exc
+        except EndpointConnectionError as exc:
+            link = error_link("connection-error")
+            name = self.endpoint_url or "AWS S3"
+            raise DvcException(
+                f"Unable to connect to '{name}'. Are you online?"
```
skshetry, line 97: "Are you online?" does not add any information that "Unable to connect" does not provide. It's redundant.

dvc/tree/s3.py
jorgeorpinel: Ah ok good idea.

dvc/tree/s3.py
jorgeorpinel: There's multiple reasons why you wouldn't be able to connect. Not being online is just the most simple case. It's a good tip I think.

skshetry: It's a tip that does not help in any way. Maybe, the mice have chewed the cables, or maybe the user typed "google" into Google and broke the whole internet.

PR #4481, opened 2020-08-27
lock: retry when trying to acquire lock

skshetry:
Previously, it'd sleep for 5s. Now, it should retry 6 times
before timing out.

Also reduced the timeout to 3s.

### DIFF
--- a/dvc/lock.py
+++ b/dvc/lock.py
@@ -2,17 +2,17 @@
 
 import hashlib
 import os
-import time
 from datetime import timedelta
 
 import flufl.lock
 import zc.lockfile
+from funcy import retry
 
 from dvc.exceptions import DvcException
 from dvc.progress import Tqdm
 from dvc.utils import format_link
 
-DEFAULT_TIMEOUT = 5
+DEFAULT_TIMEOUT = 3
 
 FAILED_TO_LOCK_MESSAGE = (
     "cannot perform the command because another DVC process seems to be "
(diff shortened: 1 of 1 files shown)


dvc/lock.py
```diff
@@ -56,13 +56,10 @@ def _do_lock(self):
...
-        try:
-            self._do_lock()
-            return
-        except LockError:
-            time.sleep(DEFAULT_TIMEOUT)
-
-        self._do_lock()
+        retries = 6
+        delay = DEFAULT_TIMEOUT / retries
+        lock_retry = retry(retries, LockError, timeout=delay)(self._do_lock)
```
efiop, line 61: `delay` is clearly incorrect terminology :slightly_smiling_face: Let's just
```suggestion
        lock_retry = retry(RETRIES, LockError, timeout=TIMEOUT)(self._do_lock)
```
with, e.g.
```
TIMEOUT = 0.1
RETRIES = 50
```

skshetry: Why do you think so? 
From [funcy's documentation](https://funcy.readthedocs.io/en/stable/flow.html#retry):
> There will be a delay in `timeout` seconds between tries

efiop: @skshetry Precisely because it is called `timeout` and not `delay`. Dividing stuff is not obvious, better to keep explicit TIMEOUT and RETRIES (and not hardcode it).

skshetry: Looks like timeout has two different meanings? I was only familiar with one.

efiop: @skshetry They are pretty much the same, but you are thinking about the total lock timeout, while I'm talking about this particular piece of code, where timeout is about the time between tries.

skshetry: @efiop, `HardlinkLock` uses `DEFAULT_TIMEOUT` as well. I don't want to have two different timeouts, so I'd have to do something like following:
```suggestion
        delay = DEFAULT_TIMEOUT / RETRIES
        lock_retry = retry(RETRIES, LockError, timeout=delay)(self._do_lock)
```

which is no different from this PR.

efiop (approved):

Issue #4042, opened 2020-06-15
Provide suggestions on typo of command


skshetry:
```console
$ git commti
git: 'commti' is not a git command. See 'git --help'.

The most similar command is
        commit
```

vs

```console
$ dvc commti
ERROR: argument COMMAND: invalid choice: 'commti' (choose from 'init', 'get', 'get-url', 'destroy', 'add', 'remove', 'move', 'unprotect', 'run', 'repro', 'pull', 'push', 'fetch', 'status', 'gc', 'import', 'import-url', 'config', 'checkout', 'remote', 'cache', 'metrics', 'params', 'install', 'root', 'list', 'freeze', 'unfreeze', 'dag', 'daemon', 'commit', 'diff', 'version', 'update', 'git-hook', 'plots')
usage: dvc [-q | -v] [-h] [-V] COMMAND ...

Data Version Control

optional arguments:
  -q, --quiet    Be quiet.
  -v, --verbose  Be verbose.
  -h, --help     Show this help message and exit.
  -V, --version  Show program's version.

Available Commands:
  COMMAND        Use `dvc COMMAND --help` for command-specific help.
    init         Initialize DVC in the current directory.
    get          Download file or directory tracked by DVC or by Git.
    get-url      Download or copy files from URL.
    destroy      Remove DVC-files, local DVC config and data cache.
    add          Track data files or directories with DVC.
    remove       Remove stage entry and unprotect outputs
    move         Rename or move a DVC controlled data file or a directory.
    unprotect    Unprotect tracked files or directories (when hardlinks or symlinks have been enabled with `dvc config cache.type`)
    run          Generate a stage file from a command and execute the command.
    repro        Reproduce complete or partial pipelines by executing their stages.
    pull         Download tracked files or directories from remote storage.
    push         Upload tracked files or directories to remote storage.
    fetch        Get tracked files or directories from remote storage into the cache.
    status       Show changed stages, compare local cache and a remote storage.
    gc           Garbage collect unused objects from cache or remote storage.
    import       Download file or directory tracked by DVC or by Git into the workspace, and track it.
    import-url   Download or copy file from URL and take it under DVC control.
    config       Get or set config options.
    checkout     Checkout data files from cache.
    remote       Set up and manage data remotes.
    cache        Manage cache settings.
    metrics      Commands to display and compare metrics.
    params       Commands to display params.
    install      Install DVC git hooks into the repository.
    root         Return the relative path to the root of the DVC project.
[...]

PR #4483, opened 2020-08-27
dvc: display command suggestions for small typos

santos22:

---

Addresses https://github.com/iterative/dvc/issues/4042

The Levenshtein distance algorithm used can be found at https://www.python-course.eu/levenshtein_distance.php

### DIFF
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -161,7 +161,7 @@ class DvcParser(argparse.ArgumentParser):
             self.error(msg % " ".join(argv), getattr(args, "func", None))
         return args
 
-    def get_cmd_choices(self):
+    def _get_cmd_choices(self):
         """Keep track of public dvc commands and hidden dvc commands.
 
         Returns:
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -142,17 +142,16 @@ class DvcParser(argparse.ArgumentParser):
         _find_parser(self, cmd_cls)
 
     def parse_args(self, args=None, namespace=None):
-        if args is None:
-            args = sys.argv[1:]
-        else:
-            args = list(args)
-
         # NOTE: overriding to provide a more granular help message.
         # E.g. `dvc plots diff --bad-flag` would result in a `dvc plots diff`
         # help message instead of generic `dvc` usage.
         try:
             args, argv = self.parse_known_args(args, namespace)
         except argparse.ArgumentError:
+            if args is None:
+                args = sys.argv[1:]
+            else:
+                args = list(args)
             self._find_cmd_suggestions(args)
         else:
             if argv:
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -142,7 +142,9 @@ class DvcParser(argparse.ArgumentParser):
         # NOTE: this is a custom check to see if any suggestions can
         # be displayed to users in case of small typos
         # E.g. `dvc commti` would display
-        # `The most similar command(s) are commit`
+        # The most similar commands are
+        #         commit
+        #         completion
         if args is None:
             args = sys.argv[1:]
         else:
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -85,6 +85,70 @@ COMMANDS = [
 ]
 
 
+def _ld(source, target):
+    """Compare similarity (Levenshtein distance) of typed command to dvc command.
+
+    Args:
+        source: typed command.
+        target: dvc command.
+
+    Returns:
+        Integer value representing Levenshtein distance between two strings.
+    """
+    rows = len(source) + 1
+    cols = len(target) + 1
+
+    d = [[0 for i in range(cols)] for j in range(rows)]
+
+    for i in range(rows):
+        d[i][0] = i
+
+    for i in range(cols):
+        d[0][i] = i
+
+    for col in range(1, cols):
+        for row in range(1, rows):
+            if source[row - 1] == target[col - 1]:
+                cost = 0
+            else:
+                cost = 1
+
+            a = d[row - 1][col] + 1
+            b = d[row][col - 1] + 1
+            c = d[row - 1][col - 1] + cost
+
+            d[row][col] = min(a, b, c)
+
+    return d[-1][-1]
+
+
+def _find_cmd_suggestions(cmd_arg, cmd_choices):
+    """Find similar command suggestions for a typed command that contains typos.
+
+    Args:
+        cmd_arg: command argument typed in.
+        cmd_choices: list of valid dvc commands to measure against.
+
+    Returns:
+        String with command suggestions to display to the user if any exist.
+    """
+
+    suggestions = []
+    min_distance = 2
+
+    for command in cmd_choices:
+        levenshtein_distance = _ld(cmd_arg, command)
+        if levenshtein_distance <= min_distance:
+            suggestions.append(command)
+
+    suggestion_str = ""
+    if suggestions:
+        suggestion_str += "\n\nThe most similar command(s) are\n"
+        for suggestion in suggestions:
+            suggestion_str += f"\t\n{suggestion}"
+    return suggestion_str
+
+
 def _find_parser(parser, cmd_cls):
     defaults = parser._defaults  # pylint: disable=protected-access
     if not cmd_cls or cmd_cls == defaults.get("func"):
@@ -103,11 +167,23 @@ def _find_parser(parser, cmd_cls):
 class DvcParser(argparse.ArgumentParser):
     """Custom parser class for dvc CLI."""
 
+    cmd_choices = []
+
     def error(self, message, cmd_cls=None):  # pylint: disable=arguments-differ
         logger.error(message)
         _find_parser(self, cmd_cls)
 
     def parse_args(self, args=None, namespace=None):
+        # NOTE: this is a custom check to see if any suggestions can
+        # be displayed to users in case of small typos
+        # E.g. `dvc commti` would display
+        # `The most similar command(s) are commit`
+        if args and args[0] not in self.cmd_choices:
+            cmd_suggestions = _find_cmd_suggestions(args[0], self.cmd_choices)
+            if cmd_suggestions:
+                sys.stderr.write(cmd_suggestions)
+                sys.exit(2)
+
         # NOTE: overriding to provide a more granular help message.
         # E.g. `dvc plots diff --bad-flag` would result in a `dvc plots diff`
         # help message instead of generic `dvc` usage.
@@ -216,6 +292,8 @@ def get_main_parser():
 
     for cmd in COMMANDS:
         cmd.add_parser(subparsers, parent_parser)
+
+    parser.cmd_choices = list(subparsers.choices.keys())
 
     return parser
 
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -86,7 +86,7 @@ COMMANDS = [
 ]
 
 
-def _find_cmd_suggestions(cmd_arg, cmd_choices):
+def _find_cmd_suggestions(cmd_arg, cmd_choices, cmd=None):
     """Find similar command suggestions for a typed command that contains typos.
 
     Args:
@@ -96,9 +96,14 @@ def _find_cmd_suggestions(cmd_arg, cmd_choices):
     Returns:
         String with command suggestions to display to the user if any exist.
     """
-    suggestion_str = (
-        f"dvc: '{cmd_arg}' is not a dvc command. See 'dvc --help'\n"
-    )
+    if cmd:
+        suggestion_str = (
+            f"dvc: '{cmd} {cmd_arg}' is not a dvc command. See 'dvc --help'\n"
+        )
+    else:
+        suggestion_str = (
+            f"dvc: '{cmd_arg}' is not a dvc command. See 'dvc --help'\n"
+        )
     suggestions = get_close_matches(cmd_arg, cmd_choices)
     if not suggestions:
         return suggestion_str
@@ -109,7 +114,10 @@ def _find_cmd_suggestions(cmd_arg, cmd_choices):
         suggestion_str += "\nThe most similar command is"
 
     for suggestion in suggestions:
-        suggestion_str += f"\n\t{suggestion}"
+        if cmd:
+            suggestion_str += f"\n\t{cmd} {suggestion}"
+        else:
+            suggestion_str += f"\n\t{suggestion}"
 
     return suggestion_str
 
@@ -133,10 +141,26 @@ class DvcParser(argparse.ArgumentParser):
     """Custom parser class for dvc CLI."""
 
     cmd_choices = []
+    sub_cmd_choices = {}
 
     def error(self, message, cmd_cls=None):  # pylint: disable=arguments-differ
         logger.error(message)
         _find_parser(self, cmd_cls)
+
+    def parse_known_args(self, args=None, namespace=None):
+        if len(args) == 2 and args[0] in self.sub_cmd_choices:
+            sub_cmd_choices = self.sub_cmd_choices[args[0]]
+
+            if args[1] in sub_cmd_choices:
+                return super().parse_known_args(args, namespace)
+
+            sub_cmd_suggestions = _find_cmd_suggestions(
+                args[1], sub_cmd_choices, args[0]
+            )
+            logger.error(sub_cmd_suggestions)
+            raise DvcParserError
+
+        return super().parse_known_args(args, namespace)
 
     def parse_args(self, args=None, namespace=None):
         # NOTE: this is a custom check to see if any suggestions can
@@ -145,11 +169,14 @@ class DvcParser(argparse.ArgumentParser):
         # The most similar commands are
         #         commit
         #         completion
+        # E.g. `dvc remote modfiy` would display
+        # The most similar commands are
+        #         remote modify
         if args is None:
             args = sys.argv[1:]
         else:
             args = list(args)
-        if args and args[0] not in self.cmd_choices:
+        if len(args) == 1 and args[0] not in self.cmd_choices:
             cmd_suggestions = _find_cmd_suggestions(args[0], self.cmd_choices)
             logger.error(cmd_suggestions)
             raise DvcParserError
@@ -265,6 +292,14 @@ def get_main_parser():
 
     parser.cmd_choices = list(subparsers.choices.keys())
     parser.cmd_choices.extend(help_action.option_strings)
+
+    for cmd, subparser in subparsers.choices.items():
+        actions = subparser._actions  # pylint: disable=protected-access
+        for action in actions:
+            if not isinstance(action.choices, dict):
+                # NOTE: we are only interested in subparsers
+                continue
+            parser.sub_cmd_choices[cmd] = list(action.choices.keys())
 
     return parser
 
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -86,7 +86,7 @@ COMMANDS = [
 ]
 
 
-def _find_cmd_suggestions(cmd_arg, cmd_choices, cmd=None):
+def _get_cmd_suggestion_msg(cmd_arg, cmd_choices, cmd=None):
     """Find similar command suggestions for a typed command that contains typos.
 
     Args:
@@ -216,10 +216,10 @@ class DvcParser(argparse.ArgumentParser):
             and args[0] not in cmd_choices
             and args[0] not in hidden_cmds
         ):
-            cmd_suggestions = _find_cmd_suggestions(
+            cmd_suggestion_msg = _get_cmd_suggestion_msg(
                 args[0], list(cmd_choices.keys())
             )
-            logger.error(cmd_suggestions)
+            logger.error(cmd_suggestion_msg)
             raise DvcParserError
 
         # NOTE: Check nested dvc subcommands for suggestions
@@ -229,10 +229,10 @@ class DvcParser(argparse.ArgumentParser):
         if len(args) == 2 and args[0] in cmd_choices:
             sub_cmd_choices = cmd_choices[args[0]]
             if sub_cmd_choices and args[1] not in sub_cmd_choices:
-                sub_cmd_suggestions = _find_cmd_suggestions(
+                cmd_suggestion_msg = _get_cmd_suggestion_msg(
                     args[1], sub_cmd_choices, args[0]
                 )
-                logger.error(sub_cmd_suggestions)
+                logger.error(cmd_suggestion_msg)
                 raise DvcParserError
 
 
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -147,16 +147,18 @@ class DvcParser(argparse.ArgumentParser):
         else:
             args = list(args)
 
-        self._find_cmd_suggestions(args)
-
         # NOTE: overriding to provide a more granular help message.
         # E.g. `dvc plots diff --bad-flag` would result in a `dvc plots diff`
         # help message instead of generic `dvc` usage.
-        args, argv = self.parse_known_args(args, namespace)
-        if argv:
-            msg = "unrecognized arguments: %s"
-            self.error(msg % " ".join(argv), getattr(args, "func", None))
-        return args
+        try:
+            args, argv = self.parse_known_args(args, namespace)
+        except argparse.ArgumentError:
+            self._find_cmd_suggestions(args)
+        else:
+            if argv:
+                msg = "unrecognized arguments: %s"
+                self.error(msg % " ".join(argv), getattr(args, "func", None))
+            return args
 
     def _get_cmd_choices(self):
         """Keep track of public dvc commands and hidden dvc commands.
@@ -290,6 +292,7 @@ def get_main_parser():
         parents=[parent_parser],
         formatter_class=argparse.RawTextHelpFormatter,
         add_help=False,
+        exit_on_error=False,
     )
 
     # NOTE: We are doing this to capitalize help message.
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -208,12 +208,12 @@ class DvcParser(argparse.ArgumentParser):
         Raises:
             dvc.exceptions.DvcParserError: raised for found suggestions.
         """
+        cmd_choices, hidden_cmds = self.get_cmd_choices()
+
         # NOTE: Check top level dvc commands for suggestions
         # E.g. `dvc commti` would display
         # The most similar command is
         #         commit
-        cmd_choices, hidden_cmds = self.get_cmd_choices()
-
         if (
             len(args) >= 1
             and args[0] not in cmd_choices
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -3,6 +3,7 @@ import argparse
 import logging
 import os
 import sys
+from difflib import get_close_matches
 
 from .command import (
     add,
@@ -85,61 +86,17 @@ COMMANDS = [
 ]
 
 
-def _ld(source, target):
-    """Compare similarity (Levenshtein distance) of typed command to dvc command.
-
-    Args:
-        source: typed command.
-        target: dvc command.
-
-    Returns:
-        Integer value representing Levenshtein distance between two strings.
-    """
-    rows = len(source) + 1
-    cols = len(target) + 1
-
-    d = [[0 for i in range(cols)] for j in range(rows)]
-
-    for i in range(rows):
-        d[i][0] = i
-
-    for i in range(cols):
-        d[0][i] = i
-
-    for col in range(1, cols):
-        for row in range(1, rows):
-            if source[row - 1] == target[col - 1]:
-                cost = 0
-            else:
-                cost = 1
-
-            a = d[row - 1][col] + 1
-            b = d[row][col - 1] + 1
-            c = d[row - 1][col - 1] + cost
-
-            d[row][col] = min(a, b, c)
-
-    return d[-1][-1]
-
-
 def _find_cmd_suggestions(cmd_arg, cmd_choices):
     """Find similar command suggestions for a typed command that contains typos.
 
     Args:
         cmd_arg: command argument typed in.
-        cmd_choices: list of valid dvc commands to measure against.
+        cmd_choices: list of valid dvc commands to match against.
 
     Returns:
         String with command suggestions to display to the user if any exist.
     """
-
-    suggestions = []
-    min_distance = 2
-
-    for command in cmd_choices:
-        levenshtein_distance = _ld(cmd_arg, command)
-        if levenshtein_distance <= min_distance:
-            suggestions.append(command)
+    suggestions = get_close_matches(cmd_arg, cmd_choices)
 
     suggestion_str = ""
     if suggestions:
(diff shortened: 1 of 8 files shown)


shcheklein:
@santos22 this looks cool! 😎 thanks for the contribution! Is there a way to simplify it a bit by using some library? I really hope there are some fuzzy match/search libraries out there that we could use? Hopefully something that integrates with argparse.

santos22:
I'm not sure if there is anything that integrates with argparse but I believe the following library might help with the matching: https://github.com/seatgeek/fuzzywuzzy

I went with the current implementation because it didn't require any external libraries. I believe a well-tested library may be more reliable 😄

shcheklein:
@santos22 yep, makes sense. That library is GPL-2 which is not compatible with Apache 2-0 license :(

santos22:
What about the `difflib` module in the Python library? The following seems helpful: https://docs.python.org/3/library/difflib.html#difflib.get_close_matches

I will test it out and report back.

santos22:
I used the tests cases found in this pull request to test out the `get_close_matches()` method - got back the same results and definitely easier to follow:
Screen Shot 2020-08-28 at 7 43 42 PM

shcheklein:
@santos22 yep! that looks great! thanks 🙏

santos22:
Awesome. I will update the PR with those changes shortly 👍

santos22:
I can fixup the second commit into the first. I also split out the unit test since "commti" now returns two suggestions instead of one.

dvc/cli.py
```diff
@@ -102,11 +123,23 @@ def _find_parser(parser, cmd_cls):
...
     def error(self, message, cmd_cls=None):  # pylint: disable=arguments-differ
         logger.error(message)
         _find_parser(self, cmd_cls)
 
     def parse_args(self, args=None, namespace=None):
+        # NOTE: this is a custom check to see if any suggestions can
+        # be displayed to users in case of small typos
+        # E.g. `dvc commti` would display
+        # `The most similar command(s) are commit`
+        if args and args[0] not in self.cmd_choices:
```
pmrowla, line 137: When DVC is run from the command line, args will be `None` here (meaning argparse should default to using `sys.argv`), so the suggested command code is never hit (except when we explicitly call `parse_args` with an argument list from the unit tests).

pmrowla: We'll need to do the same kind of thing that [argparse](https://github.com/python/cpython/blob/38e32872eb3cf0dc9dd8cef9b05e47ba03638d34/Lib/argparse.py#L1774) does here:

```suggestion
        if args is None:
            args = sys.argv[1:]
        else:
            args = list(args)
        if args and args[0] not in self.cmd_choices:
```

dvc/cli.py
```diff
@@ -102,11 +123,23 @@ def _find_parser(parser, cmd_cls):
...
+            cmd_suggestions = _find_cmd_suggestions(args[0], self.cmd_choices)
+            if cmd_suggestions:
+                sys.stderr.write(cmd_suggestions)
+                sys.exit(2)
```
pmrowla, line 141: We generally prefer using the logger rather than writing directly to stderr, and in the same vein, it will be cleaner for us to use the existing exception handling for argument parsing errors, rather than exiting immediately

```suggestion
                logger.error(cmd_suggestions)
                raise DvcParserError
```

dvc/cli.py
```diff
@@ -204,6 +237,8 @@ def get_main_parser():
     for cmd in COMMANDS:
         cmd.add_parser(subparsers, parent_parser)
 
+    parser.cmd_choices = list(subparsers.choices.keys())
```
pmrowla, line 240: It looks a bit strange to be setting this here, and my first thought was that `DvcParser.cmd_choices` should be a function or property that can dynamically figure out what the command choices are.

But for the scope of this PR, I think this is fine. (Unless you really feel like digging into `argparse.ArgumentParser` internals 😄 )

dvc/cli.py
```diff
@@ -84,6 +85,26 @@
...
+    Args:
+        cmd_arg: command argument typed in.
+        cmd_choices: list of valid dvc commands to match against.
+
+    Returns:
+        String with command suggestions to display to the user if any exist.
+    """
+    suggestions = get_close_matches(cmd_arg, cmd_choices)
+
+    suggestion_str = ""
```
pmrowla, line 100: So right now, (once the `sys.argv` issue is resolved), we will just exit with the `most similar commands are...` message, but don't actually note that the problem is due to an invalid command. I think we should start `suggestion_str` with a line similar to git's:

```
git: 'commti' is not a git command. See 'git --help'.
```

pmrowla (changes requested):
This PR looks good for the most part, but there's a couple issues I noticed. The main problem is that I noticed when I run
```
$ dvc commti
```
I still get the original full usage statement, rather than the message with command suggestions. It looks like functionality-wise, everything is working in the unit tests, just not when we parse actual command line arguments.

pmrowla:
> I can fixup the second commit into the first. I also split out the unit test since "commti" now returns two suggestions instead of one.

It's fine to leave everything in separate commits, since they will all be squashed later when the PR is merged. And in the meantime, keeping the commits separate makes it a bit easier to review.

dvc/cli.py
santos22: Sounds good. I noticed there was an `error()` method on the `DvcParser` class but didn't use it because this is where the full usage statement comes from.

Should the full usage statement not be used at all?

dvc/cli.py
santos22: Yea it does feel a bit strange but this appears to be the only place we have access to the `choices` field. I can dig into the internals some more and see if we can get this elsewhere.

santos22:
Is there a way to test the `dvc` command with these changes in my development environment? Do I need to specify anything or are the changes loaded automatically?

dvc/cli.py
pmrowla: Right, in this case I don't think we want to include the full usage statement here, so using `logger.error` instead of `self.error` is fine. If the invalid command looks like a small typo, we can just include `See 'dvc --help'` with the new suggested commands error message.

dvc/cli.py
pmrowla: I took a quick look at the argparse code as well, and I think it's probably possible to do something with the `ArgumentParser._subparsers` property, but nothing stood out as a quick and obvious solution to me. What you have now should be fine for our purposes, so don't worry about spending too much time digging into this one.

pmrowla:
> Is there a way to test the `dvc` command with these changes in my development environment? Do I need to specify anything or are the changes loaded automatically?

If you install from pip using the `-e` option, changes you make to any of the dvc python files should be loaded automatically.

```
pip install -e .[tests]
```

santos22:
These changes have broken the `--help` command. Will fix this!

dvc/cli.py
```diff
@@ -204,6 +250,9 @@ def get_main_parser():
     for cmd in COMMANDS:
         cmd.add_parser(subparsers, parent_parser)
 
+    parser.cmd_choices = list(subparsers.choices.keys())
+    parser.cmd_choices.extend(help_action.option_strings)
```
santos22, line 254: Functionally this works but it is just as strange as the line before. `--help/-h` aren't dvc commands but they function like arguments which is why they needed to be included in `cmd_choices`.

tests/func/test_cli.py
```diff
@@ -175,11 +175,17 @@ class A:
...
 
-def test_unknown_command_help(capsys):
+def test_unknown_command_help(caplog, capsys):
     try:
         _ = parse_args(["unknown"])
     except DvcParserError:
         pass
+    assert (
+        "dvc: 'unknown' is not a dvc command. See 'dvc --help'" in caplog.text
+    )
```
santos22, line 185: Test expectations changed here since we're not printing the full usage statement for unknown commands.

pmrowla (approved):
LGTM, but would like to note that for completely unknown commands (with 0 possible suggestions) we will now display
```
❯ dvc asdf
ERROR: dvc: 'asdf' is not a dvc command. See 'dvc --help'
```
rather than the full usage statement. This is fine with me (and is the same as git's behavior), but we could also consider using the new message only when there are possible suggested commands, and fall back to the full usage statement when there are no suggestions.

pmrowla:
@santos22 one of my colleagues noted that we will only support suggestions for top level DVC commands here, but not for any of the nested subcommands. Would you be able to take a look and see whether or not adding support for subcommands is doable?

i.e. it would be nice if we could get something like:
```
$ dvc remote modfiy ...
ERROR: dvc: 'remote modfiy' is not a dvc command. See 'dvc --help'

The most similar commands are
    remote modify
```

I know that git only supports suggestions for top level commands, but subcommand suggestions would be a nice thing to have in DVC.

santos22:
@pmrowla the functionality is there for nested subcommands suggestions - can do a refactor pass shortly but wanted to push up the latest changes to show my approach!

santos22:
I believe the following should be the complete list of nested subcommands we want to support:
```
'remote': ['add', 'default', 'modify', 'list', 'remove', 'rename']
'cache': ['dir']
'metrics': ['show', 'diff']
'params': ['diff']
'daemon': ['updater', 'analytics']
'git-hook': ['pre-commit', 'post-checkout', 'pre-push', 'merge-driver']
'plots': ['show', 'diff', 'modify']
'experiments': ['show', 'checkout', 'diff']
'exp': ['show', 'checkout', 'diff']
```

dvc/cli.py
```diff
@@ -102,11 +139,47 @@ def _find_parser(parser, cmd_cls):
 class DvcParser(argparse.ArgumentParser):
     """Custom parser class for dvc CLI."""
 
+    cmd_choices = []
+    sub_cmd_choices = {}
```
pmrowla, line 143: These can probably be combined into a single dict (`sub_cmd_choices.keys()` is equivalent to `cmd_choices`). For commands with no subcommands, they can just be mapped to `None` or an empty list.

pmrowla (changes requested):

pmrowla (approved):

dvc/cli.py
```diff
@@ -102,11 +139,47 @@ def _find_parser(parser, cmd_cls):
...
     def parse_args(self, args=None, namespace=None):
+        # NOTE: this is a custom check to see if any suggestions can
+        # be displayed to users in case a command contains typos
+        # E.g. `dvc commti` would display
+        # The most similar commands are
+        #         commit
+        #         completion
+        # E.g. `dvc remote modfiy` would display
+        # The most similar commands are
+        #         remote modify
```
skshetry, line 157: Is this example relevant here?

dvc/cli.py
```diff
@@ -102,11 +139,47 @@ def _find_parser(parser, cmd_cls):
...
+            )
+            logger.error(cmd_suggestions)
+            raise DvcParserError
+
+        # NOTE: this is a custom check to see if any suggestions can
+        # be displayed to users in case a nested subcommand contains typos
+        # E.g. `dvc remote modfiy` would display
+        # The most similar commands are
+        #         remote modify
+        if len(args) == 2 and args[0] in self.cmd_choices:
```
skshetry, line 174: Can we make this kind of typos work?
```console
$ git remot list
git: 'remot' is not a git command. See 'git --help'.

The most similar command is
	remote

$ dvc remot list
< prints out help message>
```

tests/func/test_cli.py
```diff
@@ -206,3 +212,42 @@ def test_unknown_subcommand_help(capsys):
...
+        pass
+    assert "\n\nThe most similar command is\n\tremote modify\n" in caplog.text
+
+    try:
+        _ = parse_args(["git-hook", "postcommit"])
+    except DvcParserError:
+        pass
+    assert "The most similar commands are" in caplog.text
+    assert "git-hook pre-commit" in caplog.text
+    assert "git-hook post-checkout" in caplog.text
```
skshetry, line 253: Can we prevent/avoid showing hidden commands/subcommands, eg: `exp` and `git-hook`? But, it's a minor issue.

skshetry (commented):
Looks good, @santos22. I have a few questions inlined.

tests/func/test_cli.py
```diff
@@ -206,3 +212,42 @@ def test_unknown_subcommand_help(capsys):
     captured = capsys.readouterr()
     help_output = captured.out
     assert output == help_output
+
+
+def test_similar_command_suggestions(caplog):
+    suggestions = {
+        "addd": "add",
+        "stats": "status",
+    }
```
pared, line 221: This could be unified with `@pytest.mark.parametrize`.

tests/func/test_cli.py
```diff
@@ -206,3 +212,42 @@ def test_unknown_subcommand_help(capsys):
...
+            f"\n\nThe most similar command is\n\t{suggestion}\n" in caplog.text
+        )
+
+    try:
+        _ = parse_args(["commti"])
+    except DvcParserError:
+        pass
+    assert "The most similar commands are" in caplog.text
+    assert "commit" in caplog.text
+    assert "completion" in caplog.text
```
pared, line 237: We could make two tests out of this.
eg
`test_similar_command_single_suggestion`
`test_similar_command_multiple_suggestions`

pared (approved):
Thanks @santos22! Looks good, two minor things about tests.

tests/func/test_cli.py
santos22: pytest is still new to me but awesome that this feature exists. Updated the tests to use this 👍

dvc/cli.py
santos22: Yep!

dvc/cli.py
santos22: This can be removed.

tests/func/test_cli.py
santos22: I believe the following should not be shown either: `daemon`, `completion`, `experiments` since I do not see them in the command reference docs. I will update this test too.

dvc/command/git_hook.py
```diff
@@ -93,7 +93,7 @@ def add_parser(subparsers, parent_parser):
 
     git_hook_subparsers = git_hook_parser.add_subparsers(
         dest="cmd",
-        help="Use `dvc daemon CMD --help` for command-specific help.",
+        help="Use `dvc git-hook CMD --help` for command-specific help.",
```
santos22, line 96: I believe this was wrong 😅

tests/func/test_cli.py
santos22: I went ahead and excluded those hidden commands/subcommands. This required adding another attribute to the `DvcParser` class for now unless there is a way to determine a command is hidden by the information found in each subparser.

dvc/cli.py
santos22:
```bash
$ dvc % dvc remot list
ERROR: dvc: 'remot' is not a dvc command. See 'dvc --help'

The most similar commands are
        remote
        remove
        root
```

dvc/cli.py
```diff
@@ -102,11 +139,48 @@ def _find_parser(parser, cmd_cls):
 class DvcParser(argparse.ArgumentParser):
     """Custom parser class for dvc CLI."""
 
+    cmd_choices = {}
+    hidden_cmds = ["completion", "daemon", "exp", "experiments", "git-hook"]
```
efiop, line 143: Doesn't seem like `completion` should be hidden though. Also, the rest of the commands are already marked as hidden (e.g. `add_help=False` in dvc/command/daemon.py). Would it be possible to just detect that while we parse `cmd_choices`? Would make maintaining this easier :slightly_smiling_face:

dvc/cli.py
```diff
@@ -204,6 +278,21 @@ def get_main_parser():
     for cmd in COMMANDS:
         cmd.add_parser(subparsers, parent_parser)
 
+    for cmd, subparser in subparsers.choices.items():
+        if cmd in parser.hidden_cmds:
+            continue
+        parser.cmd_choices[cmd] = []
```
efiop, line 284: It is a little weird that we parse it outside of parser itself and then set the `cmd_choices` here. Feels like it should be done in the parser itself, because it looks to be spreading the context of this feature without a good reason.

efiop (changes requested):
Hi @santos22 ! Thank you so much for contributing! And thank you for your patience, it is a pretty complex change, so we want to make this as easy to maintain as possible :slightly_smiling_face:

dvc/cli.py
santos22: Definitely. Will dig into more of the `argparse` internals to see if we can detect hidden commands.

dvc/cli.py
santos22: Are you suggesting moving this logic into a method on the `DvcParser` class? The reason this was done outside of the parser was because we have access to `subparsers` here but maybe we can pass that in as a parameter.

efiop: Hm, aren't those also part of the root parser?

santos22: These were part of the root parser 👍 updated this.

dvc/cli.py
```diff
@@ -213,6 +313,8 @@ def get_main_parser():
     for cmd in COMMANDS:
         cmd.add_parser(subparsers, parent_parser)
 
+    parser.set_cmd_choices()
```
efiop, line 316: Could we do this on-demand when we have to process the typo?

dvc/cli.py
```diff
@@ -103,11 +140,48 @@ def _find_parser(parser, cmd_cls):
...
+        # The most similar command is
+        #         commit
+        if args is None:
+            args = sys.argv[1:]
+        else:
+            args = list(args)
+        if (
+            len(args) >= 1
+            and args[0] not in self.cmd_choices
+            and args[0] not in self.hidden_cmds
```
efiop, line 163: Could we make these private properties (e.g. cmd_choices -> _cmd_choices) and get them dynamically instead of calling set_cmd_choices() explicitly?

dvc/cli.py
santos22: The typos get processed in `parse_args()` so this could be set one time here:
```python
def parse_args(self, args=None, namespace=None):
    if not self._cmd_choices and not self._hidden_cmds:
        self.set_cmd_choices()
    
    if args is None:
        args = sys.argv[1:]
    else:
        args = list(args)
    ...
```

efiop: @santos22 Do we need to cache it though? Unless I'm missing something, parse_args runs once, so we could just collect stuff we need once on error.

santos22: No need to cache. I think instead of having `_cmd_choices` and `_hidden_cmds` be class properties they could be returned by the `set_cmd_choices()` method (we could rename it to `get_cmd_choices()`).

And maybe lines 158-185 could be put in a new method called `find_cmd_suggestions` to clean up the `parse_args` method.

So `parse_args` would look like:
```python
    def parse_args(self, args=None, namespace=None):
        cmd_choices, hidden_cmds = self.set_cmd_choices()

        if args is None:
            args = sys.argv[1:]
        else:
            args = list(args)

        self.find_cmd_suggestions(args, cmd_choices, hidden_cmds)

        # NOTE: overriding to provide a more granular help message.
        # E.g. `dvc plots diff --bad-flag` would result in a `dvc plots diff`
        # help message instead of generic `dvc` usage.
        args, argv = self.parse_known_args(args, namespace)
        if argv:
            msg = "unrecognized arguments: %s"
            self.error(msg % " ".join(argv), getattr(args, "func", None))
        return args
```

dvc/cli.py
```diff
@@ -85,6 +86,42 @@
...
+        String with command suggestions to display to the user if any exist.
+    """
+    if cmd:
+        suggestion_str = (
+            f"dvc: '{cmd} {cmd_arg}' is not a dvc command. See 'dvc --help'\n"
+        )
+    else:
+        suggestion_str = (
+            f"dvc: '{cmd_arg}' is not a dvc command. See 'dvc --help'\n"
+        )
```
efiop, line 106: it is strange to have it duplicated like so, just use a base string and add what you need dynamically

dvc/cli.py
```diff
@@ -108,6 +145,13 @@ def error(self, message, cmd_cls=None):  # pylint: disable=arguments-differ
         _find_parser(self, cmd_cls)
 
     def parse_args(self, args=None, namespace=None):
+        if args is None:
+            args = sys.argv[1:]
+        else:
+            args = list(args)
+
+        self.find_cmd_suggestions(args)
+
```
efiop, line 154: Do we need to do it before we parse the args though? Seems like we can do that after a failure (`if argv` down below).

dvc/cli.py
```diff
@@ -117,6 +161,83 @@ def parse_args(self, args=None, namespace=None):
             self.error(msg % " ".join(argv), getattr(args, "func", None))
         return args
 
+    def get_cmd_choices(self):
```
efiop, line 164: no need to keep it public
```suggestion
    def _get_cmd_choices(self):
```

dvc/cli.py
```diff
@@ -117,6 +161,83 @@ def parse_args(self, args=None, namespace=None):
...
+                    )
+                    for action in actions:
+                        if not isinstance(action.choices, dict):
+                            # NOTE: we are only interested in subcommands
+                            continue
+                        cmd_choices[cmd].extend(action.choices.keys())
+
+        return cmd_choices, hidden_cmds
+
+    def find_cmd_suggestions(self, args):
```
efiop, line 200:
```suggestion
    def _find_cmd_suggestions(self, args):
```

efiop:
This looks better and better, just a few steps to make it mergeable and maintainable for us. :ok_hand:

Shame the turnaround time seems slow from our side, so please feel free to leave a comment once you think it is ready for another round. Just makes it easier for us to understand that the iteration is complete. Thank you! :pray:

dvc/cli.py
santos22: I thought so too but the reason I left `self.find_cmd_suggestions(args)` before the `parse_known_args` method was because `parse_known_args` throws its own internal error if the choices passed in are invalid. So the following:

```python
def parse_args(self, args=None, namespace=None):
    if args is None:
        args = sys.argv[1:]
    else:
        args = list(args)

    # NOTE: overriding to provide a more granular help message.
    # E.g. `dvc plots diff --bad-flag` would result in a `dvc plots diff`
    # help message instead of generic `dvc` usage.
    args, argv = self.parse_known_args(args, namespace)
    if argv:
        self._find_cmd_suggestions(args)

        msg = "unrecognized arguments: %s"
        self.error(msg % " ".join(argv), getattr(args, "func", None))
    return args
```

Leads to this:
Screen Shot 2020-10-24 at 6 04 49 PM

```python
2020-10-24 18:01:25,643 ERROR: argument cmd: invalid choice: 'dif' (choose from 'show', 'diff', 'modify')
```

[Python source](https://github.com/python/cpython/blob/b084d1b97e369293d2d2bc0791e2135822c923a8/Lib/argparse.py#L2499)

dvc/cli.py
santos22: Updated this 👍

santos22:
@efiop no problem about the turnaround; I find myself busy with other things and I understand the dvc team may be busy as well. Hopefully these changes can go out soon with no issues!

dvc/cli.py
efiop: @santos22 would it make sense to catch it and then process the suggestions?

dvc/cli.py
```diff
@@ -117,6 +158,83 @@ def parse_args(self, args=None, namespace=None):
...
+                    )
+                    for action in actions:
+                        if not isinstance(action.choices, dict):
+                            # NOTE: we are only interested in subcommands
+                            continue
+                        cmd_choices[cmd].extend(action.choices.keys())
+
+        return cmd_choices, hidden_cmds
+
+    def _find_cmd_suggestions(self, args):
```
efiop, line 197: It is really weird how we have this method and a function above with the same name, need better name for one or the other.

dvc/cli.py
```diff
@@ -117,6 +158,83 @@ def parse_args(self, args=None, namespace=None):
...
+        """Find similar command suggestions for commands and subcommands
+        that contain typos. Show suggestions for public commands, do not show
+        suggestions for hidden commands.
+
+        Args:
+            args: argument strings.
+
+        Raises:
+            dvc.exceptions.DvcParserError: raised for found suggestions.
+        """
```
efiop, line 207: This one is obvious, no need for a docstring in this private method
```suggestion
```

efiop (changes requested):
Unfortunately, after a closer inspection I have to state that this will require quite a bit more work on top to make it maintainable for us. I'm not sure we should put so much additional time into it, even reviewing takes a lot of time and iterations, we are already on 63 comments in this PR :slightly_frowning_face: Capasity of the core team is pretty limited, so I'm not sure we will be able to put the effort into these iterations in a timely manner.

I didn't ask this initially thinking that you've done the research, but maybe there are some good libraries that already provide this functionality that we could just start using instead of having a custom implementation?

santos22:
I'm all for making this maintainable. I quickly looked up some things initially but wasn't able to find a library that would allow us to provide "suggestions" using argparse. I recently found the following repo but it's a more simple implementation of what we have accomplished so far using the `get_close_matches()` method: https://github.com/xmonader/suggestparse

I'd be happy to come back to this if this is still needed/when the core team has more resources.

efiop:
@santos22 Sounds good. Then let's continue our iterations :ok_hand:

karajan1001:
@santos22 
Thanks for your contributions, I think it is a great work and could be helpful to our uses.
But I have a suggestion that can we treat the command and the subcommand to a whole or to one string?
There are two major advantages to this.
First, we can remove the complex `if-else` control logic.
Second, the subcommand might provide useful information for us to determine the main one. For example, only `dvc remote` can `modify` while `dvc remove` can't. So the `dvc remoe modify` should suggests to `dvc remote modify` not to `dvc remove`.

santos22:
I see the `dulwich` requirement was recently upgraded. I am getting a `ModuleNotFoundError` exception when attempting to commit a change - is there anything I need to run to fix this? I'm not very familiar with setup.py.
Screen Shot 2020-11-06 at 6 18 33 PM

https://github.com/iterative/dvc/commits/master/setup.py

efiop:
@santos22 Just run `pip install -e '.[all,tests]'` and that should fix it :slightly_smiling_face:

dvc/cli.py
santos22: I think that would make sense but I'm not sure this is possible with Python 3.7.

The following still leads to the same error message:
```python
try:
    args, argv = self.parse_known_args(args, namespace)
except ArgumentError:
    self._find_cmd_suggestions(args)
else:
    if argv:
        msg = "unrecognized arguments: %s"
        self.error(msg % " ".join(argv), getattr(args, "func", None))
finally:
    return args
```

The `exit_on_error` argument when initializating `DvcParser` seems useful but it is available only in Python 3.9.

See https://docs.python.org/3/library/argparse.html#exit-on-error

santos22: @efiop any further thoughts on this?

santos22: @efiop thoughts on the current iteration?

santos22:
@efiop thoughts on the current iteration?

dvc/cli.py
```diff
@@ -108,6 +142,13 @@ def error(self, message, cmd_cls=None):  # pylint: disable=arguments-differ
         _find_parser(self, cmd_cls)
 
     def parse_args(self, args=None, namespace=None):
+        if args is None:
+            args = sys.argv[1:]
+        else:
+            args = list(args)
```
efiop, line 148: Why do we need to do this during normal operation?

dvc/cli.py
```diff
@@ -108,6 +142,13 @@ def error(self, message, cmd_cls=None):  # pylint: disable=arguments-differ
...
+
+        self._find_cmd_suggestions(args)
```
efiop, line 150: Can we do this when we run into an error instead of doing this in advance?

efiop: Looks like currently we pretty much do the arg validation instead of argparse and raise an error if we don't recognise the command. Could we only kick this when argparse reports the error?

dvc/cli.py
santos22: Based on this following comment from Peter https://github.com/iterative/dvc/pull/4483#discussion_r479999776
> When DVC is run from the command line, args will be None here (meaning argparse should default to using sys.argv), so the suggested command code is never hit (except when we explicitly call parse_args with an argument list from the unit tests).

dvc/cli.py
santos22: The code now tries to find command suggestions when `argparse` throws an error. This currently only works for single commands (e.g. `dvc ad`) and not subcommands (e.g. `dvc remote modiyf`) so I still need to figure that out.

Raising an error when we don't recognize a command requires setting `exit_on_error=False` when initializing DvcParser. This requires Python 3.9.

More info can be found here: https://docs.python.org/3/library/argparse.html#exit-on-error

dvc/cli.py
santos22: Without those lines, we get the following error:
```python
ERROR: unexpected error - object of type 'NoneType' has no len()
```

efiop: @santos22 we didn't need that before, so probably caused by changes in this PR. argparse takes it from somewhere without this change, so I wonder why we need this.

santos22: Right. These are the internal `argparse` methods we are using:
```python
def parse_args(self, args=None, namespace=None):
    args, argv = self.parse_known_args(args, namespace)
    if argv:
        msg = _('unrecognized arguments: %s')
        self.error(msg % ' '.join(argv))
    return args

def parse_known_args(self, args=None, namespace=None):
    if args is None:
        # args default to the system args
        args = _sys.argv[1:]
    else:
        # make sure that args are mutable
        args = list(args)
    ...
```
[Source](https://github.com/python/cpython/blob/b084d1b97e369293d2d2bc0791e2135822c923a8/Lib/argparse.py#L1817)

It appears `parse_known_args()` defaults to the system args for us. This method is the one that throws the `argument cmd: invalid choice` error which we aim to override with a command suggestions error.

Because we are working with the `args` parameter outside of this method, we have to use those very same lines.

dvc/cli.py
santos22: Okay, I was looking at the wrong `add_parser()` method for subparsers when setting `exit_on_error`.

This has been fixed, tests are passing and command suggestions are shown when argparse throws an error for subcommands:
Screen Shot 2020-12-22 at 9 30 11 PM

This is definitely in better shape than where it was a couple of weeks/months ago 👍 

@efiop any issues with some of this requiring Python 3.9? How should we proceed?

santos22: I see `sys.version_info` is used throughout the code for checking Python versions: https://github.com/iterative/dvc/search?q=%22version_info%22

Maybe this can be used to determine whether or not to initialize `DvcParser` with `exit_on_error` set but that feels wrong.

efiop: > @efiop any issues with some of this requiring Python 3.9? How should we proceed?

@santos22 I haven't looked deep into it, so not sure what is the best solution here. We could either fix old versions or disable suggestions for older python versions, if it is too hard.

efiop:
@santos22 Closing this PR for now, since there have been too many iterations/comments and this PR is unmanageable at this point (even expanding the comments to find the mention is hard). Please feel free to submit a new version in a new PR when you feel comfortable with it. Thanks again for your work! :pray:

PR #4486, opened 2020-08-28
external: avoid unnecessary hash computation

### LINKED ISSUE #4485: Computing file/dir hashes multiple times (when reportedly only done once reported)
## Bug Report

Performed a dvc init --subdir in an example/DL folder (part of a larger project)
set to use common shared cache
Did not define remote (yet)

created examples/DL/dataset1 and examples/DL/dataset2 sub directories
cd examples/DL/dataset1
ran 2 dvc imports (images.dvc and labels.dvc from dataset1.git) - remote A

cd examples/DL/dataset2
ran 2 dvc imports (images.dvc and labels.dvc from dataset2.git) - remote B

The files were already in cache... During each import/received message indication "Computing file/dir hashes (only done once)".  In total this took ~40 minutes (though all files were already in cache).

cd examples/DL
dvc pull  (expecting it to say all files up to date), as this is a new dvc repo - not tracking anything other than the dvc imports.... But instead I get:
"Computing file/dir hashes (only done once)"- and waiting another 40 min.

dataset1.git and dataset2.git where unchanged between import and pull attempt

repeating dvc pull - same behavior (Computing hashes)

It appears if you issue the dvc pull command in a project that has a dvc import(s) - it ends up  recomputing the hashes every time? This doesn't seem right - is there a way to prevent?

dvc update - works as I would expect...  But this almost implies that you can't easily have a dvc project that tracks outputs (e.g. created model) - when you have dvc imports?  This doesn't sound right either. 

Verified that the dvc pull behavior is as expected in the dataset1 repo checkout

Also - I had set dvc config --global core.checksum_jobs 32
Top reveals a single core being used ~100% - but this machine is virtually idle (24 cores total).  I had assumed that option would perform the hash computation in parallel (there are certainly plenty of CPU and IO resources available) - but the option didn't seem to have much effect / if any.  
htop only revealed 3 threads/processes:

**Output of `dvc version`:**

```console
DVC version: 1.6.1 (conda)
Platform: Python 3.8.5 on Linux-4.15.0-64-generic-x86_64-with-glibc2.10
Supports: azure, gdrive, gs, hdfs, http, https, s3, ssh, oss
Repo: dvc (subdir), git
```

efiop:
Fixes #4485

### DIFF
--- a/dvc/cache/local.py
+++ b/dvc/cache/local.py
@@ -437,13 +437,25 @@ class LocalCache(CloudCache):
 
     @index_locked
     def pull(self, named_cache, remote, jobs=None, show_checksums=False):
-        return self._process(
+        ret = self._process(
             named_cache,
             remote,
             jobs=jobs,
             show_checksums=show_checksums,
             download=True,
         )
+
+        if not remote.tree.verify:
+            for checksum in named_cache.scheme_keys("local"):
+                cache_file = self.tree.hash_to_path_info(checksum)
+                if self.tree.exists(cache_file):
+                    # We can safely save here, as existing corrupted files will
+                    # be removed upon status, while files corrupted during
+                    # download will not be moved from tmp_file
+                    # (see `BaseTree.download()`)
+                    self.tree.state.save(cache_file, checksum)
+
+        return ret
 
     @staticmethod
     def _log_missing_caches(hash_info_dict):
--- a/dvc/tree/dvc.py
+++ b/dvc/tree/dvc.py
@@ -224,6 +224,17 @@ class DvcTree(BaseTree):  # pylint:disable=abstract-method
     def isexec(self, path):  # pylint: disable=unused-argument
         return False
 
+    def get_dir_hash(self, path_info, **kwargs):
+        outs = self._find_outs(path_info, strict=True)
+        if len(outs) == 1 and outs[0].is_dir_checksum:
+            if self.fetch or self.stream:
+                # pull dir cache if needed, because the other code expects
+                # to have .dir cache file present after hash calculation.
+                outs[0].get_dir_cache(**kwargs)
+            return outs[0].tree.PARAM_CHECKSUM, outs[0].checksum
+
+        return super().get_dir_hash(path_info, **kwargs)
+
     def get_file_hash(self, path_info):
         outs = self._find_outs(path_info, strict=False)
         if len(outs) != 1:
--- a/dvc/tree/dvc.py
+++ b/dvc/tree/dvc.py
@@ -24,6 +24,9 @@ class DvcTree(BaseTree):  # pylint:disable=abstract-method
     as a fallback.
     """
 
+    scheme = "local"
+    PARAM_CHECKSUM = "md5"
+
     def __init__(self, repo, fetch=False, stream=False):
         super().__init__(repo, {"url": repo.root_dir})
         self.fetch = fetch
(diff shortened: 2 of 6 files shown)


dvc/cache/local.py
```diff
@@ -437,14 +437,26 @@ def push(self, named_cache, remote, jobs=None, show_checksums=False):
...
+                cache_file = self.tree.hash_to_path_info(checksum)
+                if self.tree.exists(cache_file):
+                    # We can safely save here, as existing corrupted files will
+                    # be removed upon status, while files corrupted during
+                    # download will not be moved from tmp_file
+                    # (see `BaseTree.download()`)
+                    self.tree.state.save(cache_file, checksum)
+
+        return ret
+
```
efiop, line 459: More natural to have it here and when it is here, it is obvious that we should use `self.tree.state` and not `self.repo.state`

dvc/tree/dvc.py
```diff
@@ -224,6 +238,19 @@ def isdvc(self, path, recursive=False, strict=True):
     def isexec(self, path):  # pylint: disable=unused-argument
         return False
 
+    def get_dir_hash(self, path_info, **kwargs):
+        try:
+            outs = self._find_outs(path_info, strict=True)
+            if len(outs) == 1 and outs[0].is_dir_checksum:
+                out = outs[0]
+                # other code expects us to fetch the dir at this point
```
efiop, line 246: We could do the same thing for a subdir in a cached dir, but instead we just fallback to `super().get_dir_hash` since it is simpler. This will be cleaned up a bit more after https://github.com/iterative/dvc/issues/4144

dvc/tree/repo.py
```diff
@@ -306,6 +306,17 @@ def walk_files(self, top, **kwargs):  # pylint: disable=arguments-differ
             for fname in files:
                 yield PathInfo(root) / fname
 
+    def get_dir_hash(self, path_info, **kwargs):
+        if not self.exists(path_info):
+            raise FileNotFoundError
```
efiop, line 311: We really need to start unifying exceptions in trees so we could use exception flow control instead of abusing `self.exists`.

dvc/tree/repo.py
```diff
@@ -306,6 +306,17 @@ def walk_files(self, top, **kwargs):  # pylint: disable=arguments-differ
...
+        _, dvc_tree = self._get_tree_pair(path_info)
+        if dvc_tree and dvc_tree.exists(path_info):
+            try:
+                return dvc_tree.get_dir_hash(path_info, **kwargs)
+            except OutputNotFoundError:
+                pass
+        return super().get_dir_hash(path_info, **kwargs)
```
skshetry, line 318: This does not look right. The directory could be mixed as well, i.e. containing both dvc tracked files and git-tracked files.

skshetry: The following could be reproduced with:
```python
def test_import_mixed_dirs(tmp_dir, erepo_dir):
    erepo_dir.gen({"dir": {"foo": "foo", "bar": "bar"}})
    with (erepo_dir / "dir").chdir():
        erepo_dir.dvc.add("foo")

    erepo_dir.scm.add(["dir/foo.dvc", "dir/bar"])
    erepo_dir.scm.commit("add dir")

    from dvc.repo import Repo

    Repo.get(os.fspath(erepo_dir), "dir")
    assert (tmp_dir / "dir").read_text() == {"foo": "foo", "bar": "bar"}
```

efiop: Thank you @skshetry !

PR #4487, opened 2020-08-28
dvc: only config logger if main() called

Suor:
Makes it more friendly to be used as a library. Some setup part is still
called on import, e.g. adding trace logging level.

### DIFF
--- a/dvc/logger.py
+++ b/dvc/logger.py
@@ -180,10 +180,12 @@ def disable_other_loggers():
             logger.disabled = True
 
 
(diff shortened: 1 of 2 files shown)


pared (approved):

Suor:
Not sure what's happening in test setup so far.

efiop:
@Suor It would be great if you could fix your PR when/if you have time. :slightly_smiling_face: We weren't able to get to it ourselves just yet.

efiop:
Closing as stale.

PR #4488, opened 2020-08-28
plots: add `dvc plots -e/--experiment` option

### LINKED ISSUE #4449: experiments: provide access to experiment plots
While thinking about #4448 it occurred to me that there's currently no way to use `dvc plots` with experiment revisions. Implementation-wise, we should only need a `dvc exp plots ...` subcommand that runs regular plots in the experiments workspace, the same way that show/diff work for experiments.

One concern would be how to handle situations where someone wants to save an experiments plot - do we need functionality to git add & commit plots output into the internal experiment branches? Or can we just say for experiments, generated plots should always be considered temporary, and not be committed anywhere?

@dmpetrov @pared

pmrowla:

Will close #4449.

* Regarding docs, this option is currently hidden from the help/usage output for `dvc plots diff` the same way that the option is hidden in `dvc repro`.
* By default, if a single experiment revision is provided, we will diff against the baseline/parent rev for that experiment, rather than diffing against HEAD/workspace.

### DIFF
--- a/dvc/command/plots.py
+++ b/dvc/command/plots.py
@@ -86,7 +86,12 @@ class CmdPlotsShow(CmdPlots):
 
 class CmdPlotsDiff(CmdPlots):
     def _func(self, *args, **kwargs):
-        return self.repo.plots.diff(*args, revs=self.args.revisions, **kwargs)
+        return self.repo.plots.diff(
+            *args,
+            revs=self.args.revisions,
+            experiment=self.args.experiment,
+            **kwargs,
+        )
 
 
 class CmdPlotsModify(CmdPlots):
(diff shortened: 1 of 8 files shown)


dvc/repo/plots/diff.py
```diff
@@ -1,13 +1,23 @@
-def _revisions(revs, is_dirty):
+def _revisions(repo, revs, experiment):
     revisions = revs or []
+    if experiment and len(revisions) == 1:
+        baseline = repo.experiments.get_baseline(revisions[0])
+        if baseline:
+            revisions.append(baseline[:7])
```
pared, line 6: :+1:

pared (approved):
Looks good to me, though

dvc/command/plots.py
```diff
@@ -151,6 +156,13 @@ def add_parser(subparsers, parent_parser):
         help="Plots file to visualize. Shows all plots by default.",
         metavar="<path>",
     ).complete = completion.FILE
+    plots_diff_parser.add_argument(
+        "-e",
+        "--experiment",
```
efiop, line 161: No strong opinion, but consider just `--exp` instead of both. `-e` might come in handy later for the environment or something.

PR #4490, opened 2020-08-28
Keep `status` output format consistents

### LINKED ISSUE #4482: Discuss whether we should have consistent output format for both `dvc status` and `dvc status -c`.
> So in case of  our current test, if we commit modified `alice` we will get `Data and pipelines up to date` on `dvc status`
and we will get `new: alice` on `dvc status -c`. As this issue is about filtering the `status` I would keep output of `dvc status -c` as is and create issue discussing whether we should have consistent output format for both `dvc status` and `dvc status -c`.

_Originally posted by @pared in https://github.com/iterative/dvc/pull/4433#discussion_r477095154_

@skshetry  we can discuss the output and format and the `Stage::status()` with filter_info here

karajan1001:
fix #4482 
According to the discussion. Make `DVC status [outputs]` output format the same with `DVC status [stages]` instead of `DVC status -c [outputs]`

### DIFF
--- a/dvc/repo/status.py
+++ b/dvc/repo/status.py
@@ -19,11 +19,9 @@ def _joint_status(pairs):
                 "{} is frozen. Its dependencies are"
                 " not going to be shown in the status output.".format(stage)
             )
-        if not filter_info:
-            status_info.update(stage.status(check_updates=True))
-        else:
-            for out in stage.filter_outs(filter_info):
-                status_info.update(out.status())
+        status_info.update(
+            stage.status(check_updates=True, filter_info=filter_info)
+        )
 
     return status_info
 
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -469,13 +469,13 @@ class Stage(params.StageParams):
             return "failed", exc.target_infos
 
     @rwlocked(read=["deps", "outs"])
-    def status(self, check_updates=False):
+    def status(self, check_updates=False, filter_info=None):
         ret = []
         show_import = self.is_repo_import and check_updates
 
         if not self.frozen or show_import:
             self._status_deps(ret)
-        self._status_outs(ret)
+        self._status_outs(ret, filter_info=filter_info)
         self._status_always_changed(ret)
         self._status_stage(ret)
         return {self.addressing: ret} if ret else {}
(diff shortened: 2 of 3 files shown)


pared (approved):

skshetry (approved):
@karajan1001, I understand that the output format is not good. I find it quite ugly too (what's with the indent, right? :smile: ). If you have a good output format, please do suggest.

I have no issues with the previous output format, I only wanted the dvc's _internals_ to be consistent, as it could affect other `--show-json` functionalities. :slightly_smiling_face: 

@karajan1001, Thanks for the contributions. :tada:

skshetry:
Maybe we should go with git-like status:
```
modified: foo.txt (dvc.yaml:stage_one)
modified: foobar (foobar.dvc)
modified: dir
```

cc @karajan1001

karajan1001:
> Maybe we should go with git-like status:
> 
> ```
> modified: foo.txt (dvc.yaml:stage_one)
> modified: foobar (foobar.dvc)
> modified: dir
> ```
> 
> cc @karajan1001

@skshetry 
I had considered this format too, in the previous PR. Both are ok I think.

efiop:

jorgeorpinel:
Hi there! This PR probably implies several changes to the sample outputs of `status` commands throughout all docs. I started doing some of these changes in https://github.com/iterative/dvc.org/pull/1801/commits/6fc5e8b7aafa4a5af4d67ee1dd6502e078521b94 but we should probably double check them all. Would you like to try that, @karajan1001 ?

skshetry:
@jorgeorpinel, this does not change any output formats at all. Let's say you have a stage, "stage1" with outputs "out1" and "out2".

So, if you do `dvc status stage1`, you'll see:
```console
$ dvc status stage1
stage1:
	changed outs:
		deleted:            out1
		deleted:            out2
```

And, if you just do `dvc status out1`, you'll see filtered results:
```console
$ dvc status out1
stage1:
	changed outs:
		deleted:            out1
```

This PR just rollbacked a different output format that was temporarily there.

jorgeorpinel:
Okay, well I noticed we have some really outdated status outputs in docs thanks to this issue. We'll address that separately, then. Thanks

Issue #3069, opened 2020-01-06
Support other sha hash


adrinjalali:

efiop:
Hi @adrinjalali ! In our case md5 is not used for cryptography, it is simply used for file hashing, where it is still suitable (at least I'm not aware of any issues with it in such scenarios), so it is safe to use. If you need to encrypt your data, you could do that pretty easily on the remote itself (e.g. `sse` option for s3 remote) or encrypt in your pipeline along the way and only track those encrypted files. We also have https://github.com/iterative/dvc/issues/1317 for in-house encryption, please feel free to leave a comment there, so it is easier for us to estimate the priority.

adrinjalali:
I understand these hashes shouldn't be the only defense layer against a malicious player, but it is still a very useful thing. Imagine a malicious player who has write access to the storage (s3 bucket in this case). They can take a model file, modify it in a malicious way in a way the md5 is kept unchanged, and upload the new one. The rest of the pipeline now won't detect a change in the file, and will serve the new wrong model.

Again, I understand this should not be seen as the only way to secure the models, but in places where software goes through security audit before being used, md5 is one of the first things they're gonna use to reject it.

Another point is that having a SHA is not expensive or complicated at all, so why not include it?

JavierLuna:
I agree with @adrinjalali, using MD5 to check for integrity is [far from ideal](https://www.mscs.dal.ca/~selinger/md5collision/).

As SHA-1 [is starting to show collisions](https://security.googleblog.com/2017/02/announcing-first-sha1-collision.html) as well, DVC could join `git` and [replace the hashing function to SHA-256](https://stackoverflow.com/questions/28159071/why-doesnt-git-use-more-modern-sha/47838703#47838703).

efiop:
@adrinjalali @JavierLuna Thanks for the info, guys! We will support other hash functions eventually, it just wasn't the focus for us, but we will reconsider it. We've had an attempt from an outside contributor https://github.com/iterative/dvc/pull/1920 , but he wasn't able to finalise it :slightly_frowning_face:  If anyone is interested in contributing this feature(similar to #1920, where sha is a opt-in config option for now), we will be happy to help with everything we can.

I'll leave this issue opened in addition to https://github.com/iterative/dvc/issues/1676 , as this current one is about simply replacing md5 with something else (possibly as an option), but #1676 is originally about some custom hash functions.

JavierLuna:
I'd like to help with both issues!
[...]

Issue #1676, opened 2019-03-01
Alterative hashings for data


dunstantom:
Since md5 is sensitive to the order and format of the data, simple changes to the schema (eg. swapping two columns) or changing the type of a column (eg. integer to float) leads to new hash values and duplicated datasets. There are some alternatives that attempt to address this, such as UNF (http://guides.dataverse.org/en/latest/developers/unf/index.html). 

It would be great to specify an alternative hash function in DVC, particularly to be able to provide a user-defined function.

efiop:
Hi @dunstantom !

That is a very interesting idea! So is the cache duplication the main motivation for this request? Or do you also want to use something different than md5 for checking if your dependencies are changed in pipeline scenario?

Thanks,
ruslan

dunstantom:
I would say mainly it is having an alternative for md5. On one hand, one could argue about how SHA1 compares to md5, but on the other hand (and what I had in mind) we may have our own preferences on what kind of changes necessitate recalculating/updating a dataset. This ties in to limiting cache duplication as well, but more from the perspective of avoiding unnecessary computation rather than a concern of having storage for the cache. Hope that answers your question.

efiop:
@dunstantom Thank you for your answer! 🙂 We do think about moving to SHA too, but other users also requested some faster hashing as well, so looks like it is a general problem of supporting different hashing algos . Related to https://github.com/iterative/dvc/issues/1568 . This is something we'll take a closer look later. Thanks for the feedback! 🙂

rasmuse:
@efiop I asked earlier this week in #1568 about the possibility to use other hash algorithms. I'm shifting that part of the discussion here since this issue seems to be a better match for my request (#1568 could be solved in principle by changing to _one_ other hash algorithm, i.e., just switch the md5 function for another one).

Short report on my progress: I made a local `git checkout -b not-only-md5` yesterday. I notice that the code assumes in many places that each Remote implementation is tied to a single hash algoritm. This is most evident from the class-level constant `PARAM_CHECKSUM` (the name of the hash algorithm).

My first idea about how to implement this is to move the info about the hash algorithm to the Output and to let each instance of an Output have its own hash algoritm. At present the `OutputBase` class has a property `checksum` that depends on its `self.remote.PARAM_CHECKSUM`, and my idea would be to invert this relationship so that the Remote would get Output-specific information on the hash algorithm from the Output.

But I only know a very small part of the code base, so I'm hoping you can help me judge if this idea makes any sense at all.

[...]

PR #4495, opened 2020-08-30
dvc: use HashInfo

efiop:
Initial implementation of HashInfo to avoid using explicit tuples and have a handier way of working with hashes (e.g. comparing hashes of different types).

Related to #4144 , #3069 , #1676

### DIFF
--- a/dvc/hash_info.py
+++ b/dvc/hash_info.py
@@ -1,0 +1,10 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class HashInfo:
+    name: str
+    value: str
+
+    def __bool__(self):
+        return bool(self.value)
--- a/dvc/tree/base.py
+++ b/dvc/tree/base.py
@@ -15,6 +15,7 @@ from dvc.exceptions import (
     DvcIgnoreInCollectedDirError,
     RemoteCacheRequiredError,
 )
+from dvc.hash_info import HashInfo
 from dvc.ignore import DvcIgnore
 from dvc.path_info import PathInfo, URLInfo
 from dvc.progress import Tqdm
@@ -242,7 +243,7 @@ class BaseTree:
         )
 
         if not self.exists(path_info):
-            return self.PARAM_CHECKSUM, None
+            return None
 
         # pylint: disable=assignment-from-none
         hash_ = self.state.get(path_info)
@@ -260,17 +261,17 @@ class BaseTree:
             hash_ = None
 
         if hash_:
-            return self.PARAM_CHECKSUM, hash_
+            return HashInfo(self.PARAM_CHECKSUM, hash_)
 
         if self.isdir(path_info):
-            typ, hash_ = self.get_dir_hash(path_info, **kwargs)
+            hash_info = self.get_dir_hash(path_info, **kwargs)
         else:
-            typ, hash_ = self.get_file_hash(path_info)
+            hash_info = self.get_file_hash(path_info)
 
-        if hash_ and self.exists(path_info):
-            self.state.save(path_info, hash_)
+        if hash_info and self.exists(path_info):
+            self.state.save(path_info, hash_info.value)
 
-        return typ, hash_
+        return hash_info
 
     def get_file_hash(self, path_info):
         raise NotImplementedError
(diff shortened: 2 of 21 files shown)


dvc/tree/base.py
```diff
@@ -242,7 +243,7 @@ def get_hash(self, path_info, **kwargs):
         )
 
         if not self.exists(path_info):
-            return self.PARAM_CHECKSUM, None
+            return None
```
skshetry, line 246: Can we just do `return HashInfo(self.PARAM_CHECKSUM, None)` here? Suggesting this because otherwise there will be 3 kinds of possible results here:
1. None
2. Regular HashInfo
3. HashInfo with no value.

dvc/tree/base.py
```diff
@@ -294,8 +295,8 @@ def path_to_hash(self, path):
         return "".join(parts)
 
     def save_info(self, path_info, **kwargs):
-        typ, hash_ = self.get_hash(path_info, **kwargs)
-        return {typ: hash_}
+        hash_info = self.get_hash(path_info, **kwargs)
+        return {hash_info.name: hash_info.value}
```
skshetry, line 299: Let's create `hash_info.to_dict()` here?

dvc/tree/s3.py
```diff
@@ -332,10 +333,7 @@ def _copy(cls, s3, from_info, to_info, extra_args):
 
     def get_file_hash(self, path_info):
         with self._get_obj(path_info) as obj:
-            return (
-                self.PARAM_CHECKSUM,
-                obj.e_tag.strip('"'),
-            )
+            return HashInfo(self.PARAM_CHECKSUM, obj.e_tag.strip('"'),)
```
skshetry, line 336:
```suggestion
            return HashInfo(self.PARAM_CHECKSUM, obj.e_tag.strip('"'))
```

dvc/hash_info.py
```diff
@@ -0,0 +1,10 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class HashInfo:
```
skshetry, line 5: Loved this. I was having a lot of issues related to hashes when working with subrepos, mostly my fault but would have made debugging way easier.

skshetry: Also, a few `property`s would make it even better. Eg: `is_dir_checksum`?  But looks like it's not needed right now.

skshetry: We should try doing this more in the _internals_ instead of plain dict/tuples/list. Not saying we should use dataclass, but we can use `typing.NamedTuple/typing.TypedDict/dataclasses`. :slightly_smiling_face: 

The one that I can remember right now is `StatusInfo`. :)

dvc/tree/base.py
efiop: @skshetry `save_info` will go away soon, so I've decided not to implement it for now. `to_dict` might be useful in serialization (as well as `from_dict` in loading), but again no need to add it in this initial PR, especially since loading/dumping will be revisited in general.

dvc/tree/s3.py
efiop: Good catch! Thanks!

dvc/hash_info.py
efiop: > Also, a few propertys would make it even better. Eg: is_dir_checksum? But looks like it's not needed right now.

Exactly! :slightly_smiling_face: Yeah, just don't want to push everything in this PR. Plus `is_dir` actually will require some stuff on index/tree side, so don't want to mix those up. Clearly it will be implemented in the near future though :slightly_smiling_face:

efiop: > We should try doing this more in the internals instead of plain dict/tuples/list. Not saying we should use dataclass, but we can use typing.NamedTuple/typing.TypedDict/dataclasses/objects. slightly_smiling_face

Sorry, didn't get your point.

skshetry: > Sorry, didn't get your point.

Just trying to echo your point, that we should do this more instead of using plain dict/tuples. :arrow_down: 

> avoid using explicit tuples

PR #4496, opened 2020-08-30
dvc: show "Having any troubles?" only on unexpected errors 

### LINKED ISSUE #4126: Consider removing the support link from the CLI output
This is definitely a matter of opinion, but I wanted to raise my own reaction and see how others feel.

Personally, I _hate_ the support link that's printed after every error, so much so that I'm thinking of making myself an alias or something that strips it out.

The message is uninformative, especially after the first time it's shown. It also really feels out of place for an open-source tool. As a user I might write a dozen commands in a few minutes that each end with exit 1. So my screen is just full of this message, `Having any troubles? Hit us up at https://dvc.org/support, we are always happy to help!`, over and over again. This quickly creates quite the opposite feeling from what you're going for.

Anyway, this is definitely subjective, and it intersects with strategic considerations etc. So, up to you! But I thought I'd at least mention my reaction to it.

efiop:
Fixes #4126

### DIFF
--- a/dvc/command/destroy.py
+++ b/dvc/command/destroy.py
@@ -25,7 +25,7 @@ class CmdDestroy(CmdBase):
                 )
 
             self.repo.destroy()
-        except Exception:  # noqa, pylint: disable=broad-except
+        except DvcException:
             logger.exception("failed to destroy DVC")
             return 1
         return 0
--- a/dvc/main.py
+++ b/dvc/main.py
@@ -87,8 +87,9 @@ def main(argv=None):  # noqa: C901
     except DvcException:
         ret = 255
         logger.exception("")
-    except OSError as exc:
-        if exc.errno == errno.EMFILE:
+    except Exception as exc:  # noqa, pylint: disable=broad-except
+        # pylint: disable=no-member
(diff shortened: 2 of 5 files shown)


dvc/command/destroy.py
```diff
@@ -25,7 +25,7 @@ def run(self):
                 )
 
             self.repo.destroy()
-        except Exception:  # noqa, pylint: disable=broad-except
+        except DvcException:
             logger.exception("failed to destroy DVC")
```
efiop, line 29: Just to unify the behaviour among all of the commands. We'll also need to reconsider(probably remove) Captain Obvious messages like this one.

dvc/main.py
```diff
@@ -97,17 +98,10 @@ def main(argv=None):  # noqa: C901
             )
         else:
             logger.exception("unexpected error")
-        ret = 255
-    except Exception:  # noqa, pylint: disable=broad-except
-        logger.exception("unexpected error")
+            logger.info(FOOTER)
```
efiop, line 101: Might also consider moving it to `error`, but for now leaving as is.

PR #4503, opened 2020-08-31
experiments: support column ordering & sorting in `dvc exp show`

### LINKED ISSUE #4346: experiments show: allow filtering/sorting specific metrics/params

pmrowla:

Will close #4346.

* When using `--include-metrics/--include-params`, columns will be displayed left-to-right in the order specified by the `--include` option.
* `--sort-by <column_name>` and `--sort-order asc|desc` can be used to sort related experiments on the specified metric or param column.
    * Only nested experiments will be sorted, baseline (formal) commits will still be displayed in the git history order.
    * `--sort-order` defaults to ascending
    * Only sorting by a single column is supported

```
❯ dvc exp show --include-params=train.seed,featurize --sort-by=featurize.max_features --sort-order desc --all-commits --no-pager
┏━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
┃ Experiment            ┃ Created      ┃     auc ┃ train.seed ┃ featurize.max_features ┃ featurize.ngrams ┃
┡━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
│ workspace             │ -            │ 0.58164 │ 20170428   │ 1500                   │ 3                │
│ executor-tree         │ Aug 28, 2020 │ 0.50822 │ 20170428   │ 1500                   │ 4                │
│ └── *41a9f4b          │ Aug 28, 2020 │       - │ 20170428   │ 456                    │ 5                │
│ 6d18fc1               │ Aug 24, 2020 │ 0.58164 │ 20170428   │ 1500                   │ 3                │
│ ├── 5840202           │ Aug 24, 2020 │ 0.50822 │ 20170428   │ 1500                   │ 4                │
│ ├── 3c8db11           │ Aug 24, 2020 │ 0.50298 │ 20170428   │ 500                    │ 4                │
│ └── *3b6e442          │ Aug 25, 2020 │       - │ 20170428   │ 123                    │ 5                │
│ 0562956               │ Aug 24, 2020 │ 0.61314 │ 20170428   │ 1500                   │ 2                │
│ ├── e6d1716           │ Aug 24, 2020 │ 0.58164 │ 20170428   │ 1500                   │ 3                │
│ └── *c34fd7e          │ Aug 25, 2020 │       - │ 20170428   │ 456                    │ 5                │
│ acd59eb               │ Jul 22, 2020 │ 0.61314 │ 20170428   │ 1500                   │ 2                │
...
│ 0-git-init            │ Jun 20, 2020 │       - │ 20170428   │ 1500                   │ 3                │
└───────────────────────┴──────────────┴─────────┴────────────┴────────────────────────┴──────────────────┘
```

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -149,10 +149,8 @@ def _sort_exp(experiments, sort_by, typ, reverse):
                 item = {fname: item}
             if sort_by in item:
                 val = item[sort_by]
-                break
-        else:
-            val = None
-        return (val is None, val)
+                return (val is None, val)
+        return (True, None)
 
     ret.update(sorted(experiments.items(), key=_sort, reverse=reverse))
     return ret
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -29,10 +29,10 @@ def _filter_names(
                 f" --exclude-{label}"
             )
 
-    names = {tuple(name.split(".")) for name in names}
+    names = [tuple(name.split(".")) for name in names]
 
     def _filter(filters, update_func):
-        filters = {tuple(name.split(".")) for name in filters}
+        filters = [tuple(name.split(".")) for name in filters]
         for length, groups in groupby(filters, len):
             for group in groups:
                 matches = [name for name in names if name[:length] == group]
@@ -41,18 +41,18 @@ def _filter_names(
                     raise InvalidArgumentError(
                         f"'{name}' does not match any known {label}"
                     )
-                update_func(matches)
+                update_func({match: None for match in matches})
 
     if include:
-        ret = set()
+        ret = OrderedDict()
         _filter(include, ret.update)
     else:
-        ret = set(names)
+        ret = OrderedDict({name: None for name in names})
 
     if exclude:
         _filter(exclude, ret.difference_update)
 
-    return {".".join(name) for name in ret}
+    return [".".join(name) for name in ret]
 
 
 def _update_names(names, items):
@@ -61,7 +61,7 @@ def _update_names(names, items):
             item = flatten(item)
             names.update(item.keys())
         else:
-            names.add(name)
+            names[name] = None
 
 
 def _collect_names(all_experiments, **kwargs):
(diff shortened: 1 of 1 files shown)


dvc/command/experiments.py
```diff
@@ -41,18 +41,18 @@ def _filter(filters, update_func):
                     raise InvalidArgumentError(
                         f"'{name}' does not match any known {label}"
                     )
-                update_func(matches)
+                update_func({match: None for match in matches})
 
     if include:
-        ret = set()
+        ret = OrderedDict()
```
pmrowla, line 47: Use `OrderedDict()` with `None` values to get set-like behavior with preserved ordering

dvc/command/experiments.py
```diff
@@ -122,6 +134,30 @@ def _collect_rows(
...
+            if isinstance(item, dict):
+                item = flatten(item)
+            else:
+                item = {fname: item}
+            if sort_by in item:
+                val = item[sort_by]
+                break
+        else:
+            val = None
+        return (val is None, val)
```
skshetry, line 155: for-else is a bit confusing. `return` based will be more readable.
```suggestion
            if sort_by in item:
                val = item[sort_by]
                return (val is None, val)
        return (True, None)
```

dvc/command/experiments.py
skshetry: Or, use a `dict`?

skshetry (approved):
Looks good. Just one suggestion.

dvc/command/experiments.py
pmrowla: I thought the preserved key ordering for regular dict technically isn't guaranteed before 3.7, so I thought it would be preferred to just use `OrderedDict`?

skshetry: yes, you are right that it is only on `>3.6`. My point was that, as CPython and Python are closely tied, no one is going to change that behavior in 3.6, so should be safe to just use `dict` for convenience. Saying that I have no issues with OrderedDict, it was just a suggestion.

pared (approved):

PR #4505, opened 2020-08-31
tree.repo: fix race condition when walking subrepos

skshetry:
The subrepos were not removed from getting traversed, due to which it was causing race conditions, as we were expecting
different directories to be traversed but as subrepos were still there, `next(repo_walk)` could consume from the subrepo instead of the expected directories.

This was too hard to debug :disappointed_relieved:.

### DIFF
--- a/tests/unit/tree/test_repo.py
+++ b/tests/unit/tree/test_repo.py
@@ -182,7 +182,7 @@ def test_isdvc(tmp_dir, dvc):
 
 
 def make_subrepo(dir_, scm, config=None):
-    dir_.mkdir(parents=True)
+    dir_.mkdir(parents=True, exist_ok=True)
     with dir_.chdir():
         dir_.scm = scm
         dir_.init(dvc=True, subdir=True)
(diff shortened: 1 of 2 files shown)


tests/unit/tree/test_repo.py
```diff
@@ -398,3 +398,54 @@ def test_get_hash_cached_granular(tmp_dir, dvc, mocker):
         "md5", "8d777f385d3dfec8815d20f7496026dc",
     )
     assert dvc_tree_spy.called
+
+
+@pytest.mark.parametrize("traverse_subrepos", [True, False])
+def test_walk_nested_subrepos(tmp_dir, dvc, scm, traverse_subrepos):
```
skshetry, line 404: The tests setup is a bit lengthy.

skshetry:
To clarify: the bug might affect any directory that contains at least a sub-repo and other files/dirs when `walk`-ing.

tests/unit/tree/test_repo.py
skshetry: Could not reduce it. Tell me if it's confusing.

efiop (approved):

efiop:
Good catch!

PR #4511, opened 2020-09-01
dvc: separate cache and tree hashes

efiop:
Related to #4144 , #3069 , #1676

### DIFF
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -1,6 +1,7 @@
 import json
 import logging
 from copy import copy
+from operator import itemgetter
 
 from shortuuid import uuid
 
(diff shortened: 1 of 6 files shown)


dvc/cache/base.py
```diff
@@ -255,10 +256,52 @@ def _cache_is_copy(self, path_info):
...
+
+        return hash_info, to_info
+
+    def save_dir_info(self, dir_info, hash_info=None):
+        if hash_info and not self.changed_cache_file(hash_info):
+            return hash_info
+
+        hi, tmp_info = self._get_dir_info_hash(dir_info)
+        if hash_info:
+            assert hi == hash_info
```
efiop, line 288: Will go away in the next PR when we support md5s for external dirs. Just a sanity check for now.

dvc/hash_info.py
```diff
@@ -7,6 +7,7 @@
 class HashInfo:
     name: str
     value: str
+    dir_info: list = field(default=None, compare=False)
```
skshetry, line 10: Is it possible to have a non-nullable HashInfo, i.e. without `name` and `value` being None?
I understand that it's just a simple container, but I want to avoid possibilities of `hash_info` being:
1. `None`
2. A regular hash_info
3. hash_info with `value=None`.

1 and 2 are fine, it'd be safer to avoid 3rd case if we can.

efiop: > Is it possible to have a non-nullable HashInfo, i.e. without name and value being None?

@skshetry it is not, but we allow it. Very good points! I also thought about it, but decided not to mess with it for now. Ideally thinking about making this dataclass read-only, so everything will be set when creating an instance and so we can put appropriate checks/asserts for such things

PR #4517, opened 2020-09-02
dvc: override replace in HTTPURLInfo to include extra_parts

### LINKED ISSUE #4508: http remote: url query is ignored
## Bug Report

In the .dvc/config file, I provide a query for the http remote

```py
['remote "http"']
    url = http://localhost:8080/?param1=value1&param2=value2
```

When I receive the download/upload request in the server, the query is not received, I only get the file path

```
# sample
GET /85/ced7cc818c291d93a831db49318ff1
```
I would expect to get something like this

```
GET /85/ced7cc818c291d93a831db49318ff1?param1=value1&param2=value2
```

After checking the source code, looks like the parameters are correctly parsed at 

```
dvc/tree/http.py:45 - self.path_info = self.PATH_CLS(url)
```

but for some reason, it is lost in the upload/download functions from the same file. I've tried to trace the execution, but it is too hard. Maybe you can give me some guidance.

**Output of `dvc version`:** 
// Installed from source

```console
$ dvc version
DVC version: 1.6.6+747cb6.mod 
Platform: Python 3.6.9 on Linux-5.3.0-28-generic-x86_64-with-Ubuntu-18.04-bionic
Supports: All remotes
Repo: dvc, git
```

MetalBlueberry:
Fix #4508

### DIFF
--- a/dvc/path_info.py
+++ b/dvc/path_info.py
@@ -253,6 +253,15 @@ class HTTPURLInfo(URLInfo):
         self.query = p.query
         self.fragment = p.fragment
 
+    def replace(self, path=None):
+        return self.from_parts(
+            *self._base_parts,
+            params=self.params,
+            query=self.query,
+            fragment=self.fragment,
+            path=path,
+        )
+
     @classmethod
     def from_parts(
         cls,
--- a/tests/unit/test_path_info.py
+++ b/tests/unit/test_path_info.py
@@ -87,11 +87,10 @@ def test_path_info_as_posix(mocker, path, as_posix, osname):
     assert PathInfo(path).as_posix() == as_posix
 
 
-@pytest.mark.parametrize("cls", [HTTPURLInfo])
-def test_url_replace_path(cls):
+def test_url_replace_path():
     url = "https://user@test.com/original/path;p=par?q=quer#frag"
 
-    u = cls(url)
+    u = HTTPURLInfo(url)
     new_u = u.replace("/new/path")
 
     assert u.path == "/original/path"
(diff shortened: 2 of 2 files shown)


dvc/path_info.py
```diff
@@ -253,6 +253,15 @@ def __init__(self, url):
         self.query = p.query
         self.fragment = p.fragment
 
+    def replace(self, path=None):
```
efiop, line 256: Could you  also add a simple unit test to `tests/unit/test_path_info.py`, please?

MetalBlueberry:
I have one question, Should we include this change in URLInfo class instead of HTTPURLInfo? From my point of view, a replace method should only replace the given parameter and leave the others with the same value. In this case, the path.

efiop:
> Should we include this change in URLInfo class instead of HTTPURLInfo

remotes that use URLInfo don't actually have fragments/queries/params (at least from what I know), so I would rather not adjust it, unless someone has a use case. With HTTP it is pretty clear that it should be supported there.

tests/unit/test_path_info.py
```diff
@@ -85,3 +85,21 @@ def test_https_url_info_str():
 def test_path_info_as_posix(mocker, path, as_posix, osname):
     mocker.patch("os.name", osname)
     assert PathInfo(path).as_posix() == as_posix
+
+
+@pytest.mark.parametrize("cls", [HTTPURLInfo])
```
efiop, line 90: No need for this, we could just use `HTTPURLInfo` directly in the test.

MetalBlueberry: okey, I just followed the pattern from previous tests.

efiop (approved):
Thank you! :pray:

PR #4518, opened 2020-09-02
diff: use RepoTree to compare directory contents

pmrowla:

Partially addresses #2982.

* Directory file contents is now diff'd properly (instead of only showing the top level `dir/` added/modified/deleted status)
* Uncached directory hashes are fetched from remotes using `RepoTree(stream=True)`

This does not implement pulling all uncached file contents by default, or the file size related tasks.

### DIFF
--- a/dvc/repo/diff.py
+++ b/dvc/repo/diff.py
@@ -2,6 +2,7 @@ import os
 
 from dvc.repo import locked
 from dvc.tree.local import LocalTree
+from dvc.tree.repo import RepoTree
 
 
 @locked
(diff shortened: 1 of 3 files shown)


pared (approved):

dvc/repo/diff.py
```diff
@@ -85,3 +47,63 @@ def _exists(output):
...
+        repo_tree = RepoTree(repo, stream=True)
+        for fname in repo_tree.walk_files(output.path_info):
+            if on_working_tree:
+                hash_ = repo.cache.local.tree.get_hash(fname).value
+            else:
+                hash_ = repo_tree.get_hash(fname).value
+            yield str(fname), hash_
+    except NoRemoteError:
+        # if dir hash is missing from cache, and no remote to pull it from,
+        # there is nothing we can do here
```
efiop, line 108: Should we print a warning or at least a debug message?

dvc/repo/diff.py
```diff
@@ -85,3 +47,63 @@ def _exists(output):
...
+
+        directory: "data/"
+        file:      "data"
+    """
+
+    return dict(_output_paths(repo))
+
+
+def _output_paths(repo):
+    on_working_tree = isinstance(repo.tree, LocalTree)
```
efiop, line 68: Does `on_working_tree` matter anymore though? `RepoTree` will just use `repo.tree` and will act accordingly if it is a wtree or a git rev tree.

dvc/repo/diff.py
```diff
@@ -85,3 +47,63 @@ def _exists(output):
...
+
+
+def _dir_output_paths(repo, output, on_working_tree):
+    from dvc.config import NoRemoteError
+
+    try:
+        repo_tree = RepoTree(repo, stream=True)
+        for fname in repo_tree.walk_files(output.path_info):
+            if on_working_tree:
+                hash_ = repo.cache.local.tree.get_hash(fname).value
```
skshetry, line 102: We cannot use `repo_tree` because of the `state`, right?

pmrowla: `repo.diff` is wrapped with `@lock` so we shouldn't have issues using `state` here I think?

skshetry: I guess, I was trying to point out that `BaseTree` uses `StateNoop`. So, subsequent operations might become slower? I am not quite sure though.
Anyway, the question was if we could replace this `repo.cache.local.tree` with just `repo_tree.get_hash`?

skshetry (approved):
Looks good in general, but I am not sure if this is a good thing, as output could get quite lengthy.

dvc/repo/diff.py
pmrowla: So the issue here is that DvcTree (and RepoTree) does not re-compute dir output hashes when the workspace is dirty. I'm not sure that we would want to change that behavior just for `dvc diff`, it seems like in a majority of scenarios we would want to re-use the already computed hashes for dir outputs when we are using DvcTree and RepoTree.

efiop: That's actually an inconsistency that should be solved in RepoTree. Ideally, we should be able to work with both dirty state and clean repos. But could start with the current approach too.

pmrowla: Filed #4523 for this issue

dvc/repo/diff.py
```diff
@@ -85,3 +50,57 @@ def _exists(output):
...
+    def _exists(output):
+        if on_working_tree:
+            return output.exists
+        return True
+
+    def _to_path(output):
+        return (
+            str(output)
+            if not output.is_dir_checksum
+            else os.path.join(str(output), "")
```
skshetry, line 83: @pmrowla, is this intended? The output will have both directory and it's contents, right? Eg:
```
Added:
    data/
    data/1.txt
```

pmrowla: Yes, it's intended that it should show both the directory and the contents

PR #4530, opened 2020-09-04
diff: add state for files missing from both workspace and cache

pmrowla:

Partially addresses #2982.

Docs PR https://github.com/iterative/dvc.org/pull/1758

* When using `dvc diff` to compare local workspace against a git revision, diff output will include a new file state to indicate a file is missing from both the workspace and the cache (and most likely needs to be pulled), instead of showing that the file was deleted.
* `--hide-missing` can be used to suppress this new file state entirely.
* When comparing only git revisions (and not the workspace) this status will be suppressed by default.

[asciicast](https://asciinema.org/a/thHUinCl2BD3rAMcU5cEYZufO)

### DIFF
--- a/dvc/command/diff.py
+++ b/dvc/command/diff.py
@@ -17,12 +17,15 @@ def _digest(checksum):
     return "{}..{}".format(checksum["old"][0:8], checksum["new"][0:8])
 
 
-def _show_md(diff, show_hash=False):
+def _show_md(diff, show_hash=False, show_missing=True):
     from dvc.utils.diff import table
 
     header = ["Status", "Hash", "Path"] if show_hash else ["Status", "Path"]
     rows = []
-    for status in ["added", "deleted", "modified"]:
+    statuses = ["added", "deleted", "modified"]
+    if show_missing:
+        statuses.append("not in cache")
+    for status in statuses:
         entries = diff.get(status, [])
         if not entries:
             continue
@@ -39,7 +42,7 @@ def _show_md(diff, show_hash=False):
 
 class CmdDiff(CmdBase):
     @staticmethod
-    def _format(diff):
+    def _format(diff, show_missing=True):
         """
         Given a diff structure, generate a string of paths separated
         by new lines and grouped together by their state.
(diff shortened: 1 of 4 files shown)


dvc/repo/diff.py
```diff
@@ -104,3 +114,13 @@ def _dir_output_paths(repo_tree, output):
             yield str(fname), repo_tree.get_hash(fname).value
     except NoRemoteError:
         logger.warning("dir cache entry for '%s' is missing", output)
+
+
+def _filter_missing(repo, paths):
+    repo_tree = RepoTree(repo, stream=True)
+    for path in paths:
+        metadata = repo_tree.metadata(path)
+        if metadata.is_output or metadata.part_of_output:
```
skshetry, line 123:
```suggestion
        if metadata.is_dvc:
```

dvc/command/diff.py
```diff
@@ -69,12 +72,16 @@ def _format(diff):
...
+            "not in cache": colorama.Fore.YELLOW,
         }
 
         summary = {}
         groups = []
 
-        for state in ["added", "deleted", "modified"]:
+        states = ["added", "deleted", "modified"]
+        if show_missing:
+            states.append("not in cache")
```
skshetry, line 83: What do you think of `uncached`? I don't like that, yes, it's ugly; but it _fits_ with others?

dvc/command/diff.py
```diff
@@ -105,17 +112,23 @@ def _format(diff):
...
+            fmt += ", {not in cache} not in cache"
+        groups.append(fmt.format_map(summary))
 
         return "\n\n".join(groups)
 
     def run(self):
         try:
             diff = self.repo.diff(self.args.a_rev, self.args.b_rev)
             show_hash = self.args.show_hash
+            show_missing = False if self.args.b_rev else self.args.show_missing
```
skshetry, line 129: Can we keep this in sync with argument, i.e as `hide_missing`?

dvc/command/diff.py
pmrowla: Yeah I think `uncached` or `missing` would be cleaner as well. I went with `not in cache` here since that's what we return from `output.status()`, but I'm not opposed to changing it.

skshetry: `missing` seems ambiguous, `uncached` looks ugly. `missing cache` is the same as `not in cache`. :confused: 
I'll leave that up-to-you.

efiop (approved):

Issue #4523, opened 2020-09-03
DvcTree/RepoTree should re-compute dir hash for dirty outputs


pmrowla:
From https://github.com/iterative/dvc/pull/4518#discussion_r482763844

When local workspace is dirty, the existing dir cache and hashes are reused for dir outputs in DvcTree (and RepoTree by extension). If the output is dirty (new/modified/removed files in the directory) `get_hash()/get_dir_hash()` should re-compute the hash rather than using the existing (clean) hash from cache/state.

Once DvcTree is updated to handle dirty workspace in this scenario, `dvc diff` should be updated to use `RepoTree.get_hash()` everywhere.

pmrowla:
Related bug: `DvcTree.walk` in a dirty workspace still yields all file names in the original dir cache for the directory out, even if the file has been deleted in the local workspace. This makes diff ignore deleted files inside an output dir. Even though diff can see that the dir has been modified, it can't see which individual files have been removed.

```
example-get-started git:master  py:dvc ❯ dvc diff
example-get-started git:master  py:dvc ❯ rm data/features/test.pkl
example-get-started git:master  py:dvc ❯ dvc diff
Modified:
    data/features/

files summary: 0 added, 0 deleted, 0 modified, 0 not in cache
```

Resolving the original issue and re-computing dir cache for the dirty workspace will fix this bug without needing to modify `DvcTree.walk`, since it should be yielding filenames from the the updated (dirty) dir cache file list.

PR #4548, opened 2020-09-09
tree: open: handle dirty state in RepoTree

efiop:
We used to use that logic in DvcTree back before we had proper trees.
Now we have a RepoTree, that combines git/workspace tree with the
DvcTree and dirty state clearly belongs to the git/workspace and not
the dvc itself, as dvc state only depends on dvc-files and lock-files.

Prerequisite for #4523

### DIFF
--- a/dvc/tree/dvc.py
+++ b/dvc/tree/dvc.py
@@ -66,10 +66,6 @@ class DvcTree(BaseTree):  # pylint:disable=abstract-method
         except OutputNotFoundError as exc:
             raise FileNotFoundError from exc
 
-        # NOTE: this handles both dirty and checkout-ed out at the same time
-        if self.repo.tree.exists(path):
-            return self.repo.tree.open(path, mode=mode, encoding=encoding)
-
         if len(outs) != 1 or (
             outs[0].is_dir_checksum and path == outs[0].path_info
         ):
--- a/dvc/tree/repo.py
+++ b/dvc/tree/repo.py
@@ -151,9 +151,13 @@ class RepoTree(BaseTree):  # pylint:disable=abstract-method
             encoding = None
 
         tree, dvc_tree = self._get_tree_pair(path)
-        if dvc_tree and dvc_tree.exists(path):
-            return dvc_tree.open(path, mode=mode, encoding=encoding, **kwargs)
-        return tree.open(path, mode=mode, encoding=encoding)
+        try:
+            return tree.open(path, mode=mode, encoding=encoding)
+        except FileNotFoundError:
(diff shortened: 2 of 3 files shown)


pmrowla:
This looks good for dirty state w/`open()`, but we still don't handle dirty dirs in `RepoTree.walk()`.

If I add the following test to `tests/unit/tree/test_repo.py`:
```py
def test_walk_dirty(tmp_dir, dvc, scm):
    from dvc.utils.fs import remove

    tmp_dir.dvc_gen(
        {"dir": {"foo": "foo1", "bar": "bar1"}},
        commit="init",
    )
    tmp_dir.gen({"dir": {"bar": "bar2", "baz": "baz1"}})
    remove(str(tmp_dir / "foo"))
    tree = RepoTree(dvc)

    expected = [
        PathInfo("dir") / "bar",
        PathInfo("dir") / "baz",
    ]

    actual = []
    for root, dirs, files in tree.walk("dir"):
        for entry in dirs + files:
            actual.append(os.path.join(root, entry))

    assert (tmp_dir /"dir" / "bar").read_text() == "bar2"
    expected = [str(path) for path in expected]
    assert set(actual) == set(expected)
    assert len(actual) == len(expected)

```

We get
```
E       AssertionError: assert {'dir/baz', 'dir/foo', 'dir/bar'} == {'dir/baz', 'dir/bar'}
E         Extra items in the left set:
E         'dir/foo'
E         Full diff:
E         - {'dir/baz', 'dir/bar'}
E         + {'dir/baz', 'dir/foo', 'dir/bar'}
E         ?             +++++++++++

/Users/pmrowla/git/dvc/tests/unit/tree/test_repo.py:524: AssertionError
```

So for a dirty directory and `RepoTree.walk()` we handle modified and added files for a dirty directory, but not removed since `DvcTree.walk()` yields every expected file in the directory (as it should), but `RepoTree.walk()` does not check to make sure the dvctree files exist in the dirty local workspace.

efiop:
@pmrowla Correct, this is me just starting out migrating the dirty functionality from DvcTree to RepoTree. Walk, get_hash and other calls will need to be fixed.

efiop:
@pmrowla btw, your test seems incorrect. You delete `tmp_dir / foo`, but you probably meant `tmp_dir / "dir" / "foo"`, right?

pmrowla:
> @pmrowla btw, your test seems incorrect. You delete `tmp_dir / "foo"`, but you probably meant `tmp_dir / "dir" / "foo"`, right?

Good catch, that was my mistake, I tried it again with the fixed test and it passes 👍

PR #4553, opened 2020-09-09
Allow config option for HTTP method for uploads (Artifactory)

rfdickerson:
# Use case

Our MLE team at Expedia Group uses JFrog's Artifactory for storing binary artifacts. While Artifactory supports an HTTP/HTTPS interface, instead of POST it only accepts PUT for the new artifacts. This PR provides an alternative in this case.

Documentation change:
https://github.com/iterative/dvc.org/pull/1769

[JFrog Artifactory HTTP REST API](https://www.jfrog.com/confluence/display/JFROG/Artifactory+REST+API)

---
This change is [Reviewable](https://reviewable.io/reviews/iterative/dvc/4553)

### DIFF
--- a/dvc/tree/http.py
+++ b/dvc/tree/http.py
@@ -54,7 +54,7 @@ class HTTPTree(BaseTree):  # pylint:disable=abstract-method
         self.ask_password = config.get("ask_password", False)
         self.headers = {}
         self.ssl_verify = config.get("ssl_verify", True)
-        self.method = config.get("http_method", "POST")
+        self.method = config.get("method", "POST")
 
     def _auth_method(self, path_info=None):
         from requests.auth import HTTPBasicAuth, HTTPDigestAuth
--- a/dvc/tree/http.py
+++ b/dvc/tree/http.py
@@ -54,7 +54,7 @@ class HTTPTree(BaseTree):  # pylint:disable=abstract-method
         self.ask_password = config.get("ask_password", False)
         self.headers = {}
         self.ssl_verify = config.get("ssl_verify", True)
-        self.http_method = config.get("http_method", "POST")
+        self.method = config.get("http_method", "POST")
 
     def _auth_method(self, path_info=None):
         from requests.auth import HTTPBasicAuth, HTTPDigestAuth
@@ -195,7 +195,7 @@ class HTTPTree(BaseTree):  # pylint:disable=abstract-method
                             break
                         yield chunk
 
-        response = self.request(self.http_method, to_info.url, data=chunks())
+        response = self.request(self.method, to_info.url, data=chunks())
         if response.status_code not in (200, 201):
             raise HTTPError(response.status_code, response.reason)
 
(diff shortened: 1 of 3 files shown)


dvc/tree/http.py
```diff
@@ -194,7 +195,7 @@ def chunks():
                             break
                         yield chunk
 
-        response = self.request("POST", to_info.url, data=chunks())
+        response = self.request(self.http_method, to_info.url, data=chunks())
```
efiop, line 198: Hi @rfdickerson !

I suppose you need to use PUT in your scenario, right? Could you elaborate on what your scenario is?

rfdickerson: Absolutely! Our company at Expedia group uses JFrog's Artifactory for storing binary artifacts. While Artifactory supports an HTTP/HTTPS interface, instead of POST it only accepts PUT for uploading new artifacts. This PR provides an alternative in this case.

efiop: PUT seems like a more appropriate method than POST, considering that we already know the URL of cache file that we are uploading. I don't have a good statistics on how often people use one or another, so not sure if maybe we should consider switching to PUT by default along the way. We'll keep an eye on it, maybe we'll switch the defaults in the future.

Let's keep it as is for now. Thank you so much!

dvc/tree/http.py
```diff
@@ -54,6 +54,7 @@ def __init__(self, repo, config):
         self.ask_password = config.get("ask_password", False)
         self.headers = {}
         self.ssl_verify = config.get("ssl_verify", True)
+        self.http_method = config.get("http_method", "POST")
```
efiop, line 57: Btw, are you using dvc through API? For CLI we also need to add this to https://github.com/iterative/dvc/blob/master/dvc/config.py#L121

efiop: Also, would changing the name from `http_method` to `method` make sense to you?

efiop:
Could you please contribute a doc fix for HTTP in https://dvc.org/doc/command-reference/remote/modify to include this new config option?

And one more thing, I see that github is not able to tie your commit's author with your github profile. Likely you just need to add the email used in those commits to https://github.com/settings/emails

dmpetrov:
> JFrog's Artifactory for storing binary artifacts

> it only accepts PUT for the new artifacts

@rfdickerson thank you for this Artifactory insight and the change 🚀

We had an internal discussion about a month ago regarding possible integration with tools like Artifactory and realized that we are lacking hands-on experience with these tools.

I'd really appreciate it if you could share your experience. Please let me know if we can set up a quick chat - I can shoot an email to the address from your profile.

rfdickerson:
@dmpetrov would love a quick chat and maybe a quick screen share demo of it working. Expedia address would work fine.

dvc/tree/http.py
rfdickerson: yep! fixed

dvc/tree/http.py
```diff
@@ -54,6 +54,7 @@ def __init__(self, repo, config):
         self.ask_password = config.get("ask_password", False)
         self.headers = {}
         self.ssl_verify = config.get("ssl_verify", True)
+        self.method = config.get("http_method", "POST")
```
efiop, line 57:
```suggestion
        self.method = config.get("method", "POST")
```

efiop (approved):
Thank you @rfdickerson ! :pray:

PR #4554, opened 2020-09-10
tree: repo: handle dirty state

### LINKED ISSUE #4523: DvcTree/RepoTree should re-compute dir hash for dirty outputs

efiop:
DvcTree itself shouldn't care about the dirty state at all, as there is really no such thing for it, since its state is determined by the dvcfiles and lockfiles present.

What's a little weird here is that if a cached file/directory doesn't exist in the workspace, it is not considered dirty and we just treat it as a dvc output. But if it is modified, then we do consider it as dirty. Have to think about this a bit more... 

Fixes #4523 

### DIFF
--- a/dvc/tree/dvc.py
+++ b/dvc/tree/dvc.py
@@ -103,14 +103,14 @@ class DvcTree(BaseTree):  # pylint:disable=abstract-method
         try:
             self.metadata(path)
             return True
-        except OutputNotFoundError:
+        except FileNotFoundError:
             return False
 
     def isdir(self, path):  # pylint: disable=arguments-differ
         try:
             meta = self.metadata(path)
             return meta.isdir
-        except OutputNotFoundError:
+        except FileNotFoundError:
             return False
 
     def check_isdir(self, path_info, outs):
@@ -123,22 +123,17 @@ class DvcTree(BaseTree):  # pylint:disable=abstract-method
         if out.path_info == path_info:
             return True
 
-        # for dir checksum, we need to check if this is a file inside the
-        # directory
         try:
             self._get_granular_checksum(path_info, out)
             return False
         except FileNotFoundError:
-            # path may be an untracked file from a dirty workspace
-            if self.repo.tree.isfile(path_info):
-                return False
             return True
 
     def isfile(self, path):  # pylint: disable=arguments-differ
         try:
             meta = self.metadata(path)
             return meta.isfile
-        except OutputNotFoundError:
+        except FileNotFoundError:
             return False
 
     def _fetch_dir(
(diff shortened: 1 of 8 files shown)


dvc/tree/git.py
```diff
@@ -216,7 +216,11 @@ def to_ctime(git_time):
             return sec + nano_sec / 1000000000
 
         obj = self._git_object_by_path(path)
-        if obj is None:
+        if (
+            obj is None
+            or self.dvcignore.is_ignored_dir(path)
+            or self.dvcignore.is_ignored_file(path)
```
efiop, line 222: Potential for optimization. Probably faster to check dvcignore than to `_git_object_by_path`

dvc/tree/local.py
```diff
@@ -160,8 +160,12 @@ def isexec(self, path):
         mode = os.stat(path).st_mode
         return is_exec(mode)
 
-    @staticmethod
-    def stat(path):
+    def stat(self, path):
+        if self.dvcignore.is_ignored_dir(
+            path
+        ) or self.dvcignore.is_ignored_file(path):
```
efiop, line 166: Could probably use `self.dvcignore.check_ignore`

dvc/tree/dvc.py
```diff
@@ -103,14 +103,14 @@ def exists(self, path):  # pylint: disable=arguments-differ
         try:
             self.metadata(path)
             return True
-        except OutputNotFoundError:
+        except FileNotFoundError:
```
efiop, line 106: Could've made this a separate change, but in new code it felt too weird that I have to catch an internal exception.

efiop:
Looking into performance in the mean time...

tests/func/test_tree.py
```diff
@@ -191,6 +191,7 @@ def test_repotree_walk_fetch(tmp_dir, dvc, scm, local_remote):
     out = tmp_dir.dvc_gen({"dir": {"foo": "foo"}}, commit="init")[0].outs[0]
     dvc.push()
     remove(dvc.cache.local.cache_dir)
+    remove(tmp_dir / "dir")
```
efiop, line 194: This is because otherwise `tmp_dir / "dir"` is dirty.

Issue #4446, opened 2020-08-21
plots/metrics: accept any viable target


pared:
Running:
```
#!/bin/bash

rm -rf repo
mkdir repo

pushd repo

git init --quiet
dvc init --quiet


echo "{'metric':0.95}" >> metric.json
git add -A
git commit -am "initial"
git tag v1


echo "{'metric':0.96}" >> metric.json
git add -A
git commit -am "second"
git tag v2

dvc metrics diff v1 v2 -v --targets metric.json
```
yields no result. `metrics diff/show` should be able to handle non-`metrics` targets that can be interpreted by our parsers.

dmpetrov:
`dvc params diff` should follow the same logic. But first, `--targets` needs to be introduced for params.

dmpetrov:
Metrics/plots/params difference for Git-files is a useful scenario even outside of the data-files or pipeline context. It feels like it is a separate functionality in DVC that needs to be fully supported.

For example, it might be used in CI/CD to compare metrics/plots/params with master or previous commits in a repo when DVC is not used to generate reports like this - https://github.com/iterative/cml_dvc_case/pull/4

Additional scenarios to support:
1. metrics/plots/params diff for not-DVC repos (with `--targets` specified).
2. metrics/plots/params diff for DVC repos without any pipelines (without `--targets` specified). So, a dummy\phony stage might be needed.

pared:
@dmpetrov I see here following scenarios (lets take `plots` as example command):

1. DvcRepo, file that is not marked as plot, but we are able to `dvc plots show` and `dvc plots diff` - in this case providing `targets` is required - because we need to know what we are trying to parse.

2. DvcRepo, some file is marked as plot, but not using `dvc run --plots` (but, for example `dvc plots add {file}`). Later we are able to `dvc plots show/diff` without `targets`, and dvc is able to find it by itself.

3. GitRepo, we are able to `dvc plots show/diff` - in this case there is no way to mark file as plot beforehand, and that's why we need to provide `targets`.

4. Not a repo, we are able to `dvc plots show` - `targets` required

Is there any prioritization between those scenarios?

dmpetrov:
@pared All looks good!

Re (4). It is not a high-priority scenario. But it would be amazing to support it if not a big implementation overhead. `dvc plots show` makes sense for just a file outside of any repo. The `diff`s - probably not (we can imaging a scenario with two files diff, but cmd signature needs to be changed. it is probably does not worth it).

[...]

PR #4562, opened 2020-09-11
brancher: make brancher part of scm

pared:

Prerequisite for #4446 

In order to make plots/metrics/params `diff` accept any viable target, we need to make brancher repo agnostic. Considering that `diff` makes sense only in case of some kind of `scm`, it makes sense to make brancher part of `scm`.

### DIFF
--- a/dvc/scm/base.py
+++ b/dvc/scm/base.py
@@ -3,6 +3,7 @@
 import os
 
 from dvc.exceptions import DvcException
+from dvc.tree import LocalTree
 
 
 class SCMError(DvcException):
@@ -165,3 +166,15 @@ class Base:
 
     def close(self):
         """ Method to close the files """
+
+    def brancher(
+        self,
+        root_dir,
+        revs=None,
+        all_branches=False,
+        all_tags=False,
+        all_commits=False,
+    ):  # pylint: disable=no-self-use, unused-argument
+        yield "workspace", LocalTree(
+            None, {"url": root_dir}, use_dvcignore=True
+        )
--- a/tests/func/metrics/test_show.py
+++ b/tests/func/metrics/test_show.py
@@ -17,7 +17,7 @@ def test_show_simple(tmp_dir, dvc, run_copy_metrics):
     run_copy_metrics(
         "metrics_t.yaml", "metrics.yaml", metrics=["metrics.yaml"],
     )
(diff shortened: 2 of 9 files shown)


pared:
@Suor does the viewer interact with brancher in any way? In this change I intended to unify the branches behaviour and return "workspace" instead of ""/"workspace" depending on repository status, not sure if that won't interfere with something at `viewer`.

pmrowla (approved):

tests/func/metrics/test_show.py
```diff
@@ -17,23 +17,25 @@ def test_show_simple(tmp_dir, dvc, run_copy_metrics):
     run_copy_metrics(
         "metrics_t.yaml", "metrics.yaml", metrics=["metrics.yaml"],
     )
-    assert dvc.metrics.show() == {"": {"metrics.yaml": 1.1}}
+    assert dvc.metrics.show() == {"workspace": {"metrics.yaml": 1.1}}
```
efiop, line 20: This is a breaking API change, people might rely on this, so normally we can't just break it like that, unless we have a very good reason.

efiop (changes requested):
This is breaking our API, could you clarify the reasons to do that?

dvc/scm/base.py
```diff
@@ -165,3 +166,15 @@ def belongs_to_scm(self, path):
...
+
+    def brancher(
+        self,
+        root_dir,
+        revs=None,
+        all_branches=False,
+        all_tags=False,
+        all_commits=False,
+    ):  # pylint: disable=no-self-use, unused-argument
+        yield "workspace", LocalTree(
```
Suor, line 178: Do we need to return workspace from here? It looks like it makes mess, i.e. we need to skip it later.

Suor:
Another thing is a branch or tag might be named workspace. In this case on metric show that branch is indistinguishable from actual workspace.

dvc/scm/base.py
efiop: Also `LocalTree` has absolutely nothing to do with Base SCM

pared:
Seems like starting from redesigning the brancher was wrong idea. Not only it would break the current functionality, but also the end result could be achieved in [other way](https://github.com/iterative/dvc/pull/4627). Closing for now.

PR #4564, opened 2020-09-12
`check-ignore` show sub-repo patterns

### LINKED ISSUE #4555: check-ignore: should ignore subrepos
Subrepos are a part of dvcignore now (e.g. https://github.com/iterative/dvc/blob/7fa90e425d1c3f2b1771b541ccca4ac788d1abf9/dvc/ignore.py#L296 ) so this test https://github.com/iterative/dvc/blob/7fa90e425d1c3f2b1771b541ccca4ac788d1abf9/tests/func/test_check_ignore.py#L91 should be backwards.

karajan1001:
fix #4555
1. Sub-repo pattern now would be considered in `check-ignore` command.
2. New outputs for sub-repo patterns in a `detail`
3. Related tests.

### DIFF
--- a/dvc/ignore.py
+++ b/dvc/ignore.py
@@ -231,7 +231,7 @@ class DvcIgnoreFilter:
                     root, set()
                 ) | {d}
                 pattern_info = PatternInfo(
-                    f"/{d}/", "(inside sub_repo:{})".format(d)
+                    f"/{d}/", "in sub_repo:{}".format(d)
                 )
                 new_pattern = DvcIgnorePatterns([pattern_info], root)
                 old_pattern = self.ignores_trie_tree.longest_prefix(root).value
--- a/dvc/ignore.py
+++ b/dvc/ignore.py
@@ -230,7 +230,10 @@ class DvcIgnoreFilter:
                 self._ignored_subrepos[root] = self._ignored_subrepos.get(
                     root, set()
                 ) | {d}
-                new_pattern = DvcIgnorePatterns([f"/{d}/"], root)
+                pattern_info = PatternInfo(
+                    f"/{d}/", "(inside sub_repo:{})".format(d)
+                )
+                new_pattern = DvcIgnorePatterns([pattern_info], root)
                 old_pattern = self.ignores_trie_tree.longest_prefix(root).value
                 if old_pattern:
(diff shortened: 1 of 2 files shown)


efiop (approved):
Looks great! Btw, looks like we can now use
```
return self.check_ignore(path).match
```
in `DvcIgnoreFilter.is_ignored()`.

Btw, can we assume that the performance of `check_ignore` on par with `is_ignore_dir/file`?

karajan1001:
@efiop

1. Currently `_is_ignored` is returns `False` for files in some default ignored paths ('.dvc', '.git', '/hg'), but `_is_ignored` don't. This follows the result and logic of `git check-ignore`, also the return value of the `dvc check-ignore` which returns whether there is any pattern matches the target, not whether the target is ignored or not. 

Maybe we can do it in our own ways.

2. Yes, their algo complexity is the same.

efiop:
@karajan1001 Ah, I think there is a misunderstanding here. I meant `is_ignored` https://github.com/iterative/dvc/blob/bc5611968f44979016bf87904e0abc2d35a046c2/dvc/ignore.py#L337 , not `_is_ignored`

karajan1001:
> @karajan1001 Ah, I think there is a misunderstanding here. I meant `is_ignored`
> 
> https://github.com/iterative/dvc/blob/bc5611968f44979016bf87904e0abc2d35a046c2/dvc/ignore.py#L337
> 
> , not `_is_ignored`

Oh, I see now, but the conclusion is the same. Currently `check_ignore` only examines the matches of the patterns disregarding the type of the matched pattern, excluded one or included one. This follows the logic from Git, maybe it is the time to change it now.

efiop:
@karajan1001 ah, got it. Current behaviour makes total sense. Thank you for clarifying!

efiop:
@karajan1001 Than you so much for contributing! :pray: 

As I sign of gratitude and trust, we've sent you an invitation to join this repo as a collaborator, which means that you will now have rights to review PRs, etc](https://docs.github.com/en/github/setting-up-and-managing-organizations-and-teams/repository-permission-levels-for-an-organization#repository-access-for-each-permission-level) if you so choose :slightly_smiling_face:

PR #4116, opened 2020-06-25
Issue #4025:  cli: added cwd flag

JosephTLucas:
Added -C and --cwd as global flags to change working directory.
Added Unit Test and modified `repro` unit code and unit tests accordingly.

Closes #4025

* [X] ❗ I have followed the [Contributing to DVC](https://dvc.org/doc/user-guide/contributing/core) checklist.

* [ ] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here.

* [X] ❌ I will check DeepSource, CodeClimate, and other sanity checks below. (We consider them recommendatory and don't expect everything to be addressed. Please fix things that actually improve code or fix bugs.)

Thank you for the contribution - we'll try to review it as soon as possible. 🙏

---
If you like the approach, I'll submit another PR documenting the flag.

PR #4566, opened 2020-09-13
cli: add --cd flag to change directories before executing commands

### LINKED ISSUE #4025: Use a DVC command, but without having to change directories (like git -C <dir>)
Similar to:  [ Starting git 1.8.5 (Q4 2013)](https://github.com/git/git/blob/5fd09df3937f54c5cfda4f1087f5d99433cce527/Documentation/RelNotes/1.8.5.txt#L115-L116), you will be able to "use a Git command, but without having to change directories.  Just like "make -C <directory>", "git -C <directory> ..." tells Git to go there before doing anything else.

Could a similar to Git and Make  '-C' option be added to DVC?   

The particular use case I have in mind, if DVC is used from Jupyter, spurious `%cd` -  add quite a bit of clutter:

```
   %cd  $datasets
   !dvc pull --force --recursive xxxx --quiet
   %cd -
```

You can end up in a wrong directory and then things fail.  Much nicer, if one will be able to write:

```
   !dvc -C $datasets pull --force --recursive xxxx --quiet
```

bobertlo:
Fixes #4025

I was initally going to use `--cwd` and remove `--cwd` from `repro`, but that left `repro` with the very strange `args.c` value. So I turned to the tar manpage for inspiratoin and found the long form of `-C` is `--cd`. Tar is good enough for me.

I actually ran into this problem yesterday trying to launch dvc from ssh for a dumb reason.

I also added a docs PR:
https://github.com/iterative/dvc.org/pull/1780

### DIFF
--- a/dvc/command/repro.py
+++ b/dvc/command/repro.py
@@ -96,7 +96,8 @@ def add_parser(subparsers, parent_parser):
         "-c",
         "--cwd",
         default=os.path.curdir,
-        help="Directory within your repo to reproduce from.",
+        help="Directory within your repo to reproduce from. Note: deprecated "
+        "by `dvc --cd <path>`.",
         metavar="<path>",
     )
     repro_parser.add_argument(
--- a/dvc/command/repro.py
+++ b/dvc/command/repro.py
@@ -15,8 +15,7 @@ logger = logging.getLogger(__name__)
 class CmdRepro(CmdBase):
     def run(self):
         saved_dir = os.path.realpath(os.curdir)
-        if self.args.cwd:
-            os.chdir(self.args.cwd)
+        os.chdir(self.args.cwd)
 
         # Dirty hack so the for loop below can at least enter once
         if self.args.all_pipelines:
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -1,6 +1,7 @@
 """DVC command line interface"""
 import argparse
 import logging
+import os
 import sys
 
 from .command import (
@@ -189,6 +190,14 @@ def get_main_parser():
         action=VersionAction,
         nargs=0,
         help="Show program's version.",
+    )
+
+    parser.add_argument(
+        "--cd",
+        default=os.path.curdir,
+        metavar="<path>",
+        help="Change to directory before executing.",
+        type=str,
     )
 
     # Sub commands
--- a/dvc/command/base.py
+++ b/dvc/command/base.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from abc import ABC, abstractmethod
 
 logger = logging.getLogger(__name__)
@@ -32,6 +33,8 @@ class CmdBase(ABC):
         from dvc.repo import Repo
         from dvc.updater import Updater
 
+        os.chdir(args.cd)
+
         self.repo = Repo()
         self.config = self.repo.config
         self.args = args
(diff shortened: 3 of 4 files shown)


dvc/command/init.py
```diff
@@ -13,7 +13,7 @@ def run(self):
 
         try:
             self.repo = Repo.init(
-                ".",
+                self.args.cd,
```
bobertlo, line 16: It occurred to me later maybe this would be better as a chdir in the CmdBaseNoRepo (w/e?) constructor?

dvc/command/repro.py
```diff
@@ -15,8 +15,7 @@
 class CmdRepro(CmdBase):
     def run(self):
         saved_dir = os.path.realpath(os.curdir)
-        if self.args.cwd:
```
bobertlo, line 18: This is always initialized with os.curdir if not given a value. Alternative would be `if self.args.cwd != os.curdir`

pared (approved):
@bobertlo Looks good to me, lets please wait for some more reviews, as it impact all commands.

pmrowla (commented):
Functionality-wise this LGTM, but it looks a bit strange for us to have different naming conventions for `dvc -C/--cd` and `dvc repro -c/--cwd`. Seems to me that we should be using the same set of short/long flags for both of those options.

bobertlo:
@pmrowla the idea is not breaking existing functionality. With arparse, having `--cwd` on the global level would break `--cwd` functionality on the `repro` command. Also `-C` is much more standard than `-c` in this context.

I've seen at least one other issue where replacing argparse would be beneficial but otherwise this is the best I could think of so far?

I think the audience using `-c`/`--cwd` is limited enough that backward compatibility would be sufficient?

pmrowla:
w/argparse you can get around the naming conflict by specifying the `dest` variable name for either one of the conflicting options (so internally we would just have something like `args.cwd` and `args.repro_cwd`), but still keep consistent usage in terms of what the user would see.

I'm not sure how many DVC users actually use (or are even aware of) `repro -c/--cwd`, but I still think it looks odd to have two different sets of option flags that have the same effect. (maybe deprecating the repro option is something we should consider)

bobertlo:
@pmrowla thanks for the great info on argparse. When I started working on this I saw #4116 and assumed the `repro` arguments would be left for compatibility and deprecated. I was trying to be as unobtrusive as possible but do feel `-C` is a much more intuitive namespace for this than `-c`.

The PR as is should not break any current functionality but I would be happy to make changes if a consensus is reached.

efiop (changes requested):
Sorry for the delay.

This looks great! Just 2 minor comments:

1) How about we only keep `--cd` and don't use `-C` for now? It seems like this option is for a pretty specific advanced usage, so taking up a nice short `-C` option feels a bit wasteful. We could always consider bringing it back in the future.

2) Great point about `repro -c`! Let's mark it as obsolete in dvc and docs right now and we'll drop it in 2.0

bobertlo:
Ok. I will get on this ASAP.

1. Sure. I only chose `-C` because (iirc) it was mentioned in the original issue? Also, `-C` has a strong precedent here from that flags use in both `tar` and `make`. I like the flag but don't feel strongly. I think anyone who needs this flag will be able to find it and being able to use `-C` won't actually have much of a difference at all. 
2. I will make sure to update the docs for `repro -c` at the same time.

bobertlo:
Okay, I have simply removed the `-C` flag here and we should be good to go on this PR. I will review the docs PR and update the syntax and look at showing `repro -c` as deprecated.

bobertlo:
@efiop okay, I have also added a little note to the help message for `dvc repro` about the -c flag being deprecated.

This should be ready for review now.

efiop (approved):
Thanks!

PR #4569, opened 2020-09-14
tree: validate resolved remote:// sections

efiop:
Kudos to @PeterFogh for catching this one :pray: 

### DIFF
--- a/dvc/tree/__init__.py
+++ b/dvc/tree/__init__.py
@@ -78,5 +78,11 @@ def _resolve_remote_refs(repo, remote_conf):
 
 
 def get_cloud_tree(repo, **kwargs):
+    from dvc.config import SCHEMA, ConfigError, Invalid
+
     remote_conf = _get_conf(repo, **kwargs)
+    try:
+        remote_conf = SCHEMA["remote"][str](remote_conf)
+    except Invalid as exc:
+        raise ConfigError(str(exc)) from None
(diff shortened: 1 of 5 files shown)


dvc/tree/__init__.py
```diff
@@ -78,5 +78,11 @@ def _resolve_remote_refs(repo, remote_conf):
 
 
 def get_cloud_tree(repo, **kwargs):
+    from dvc.config import SCHEMA, ConfigError, Invalid
+
     remote_conf = _get_conf(repo, **kwargs)
+    try:
+        remote_conf = SCHEMA["remote"][str](remote_conf)
```
efiop, line 85: One might ask on why not to do this automatically during config verification (e.g. dvc/config.py::ByUrl), but that one has a very limited scope, so it is not trivial to make it resolve urls like this. Plus that would be static, while this resolution is dynamic.

dvc/tree/base.py
```diff
@@ -75,8 +75,8 @@ def __init__(self, repo, config):
     @cached_property
     def hash_jobs(self):
         return (
-            self.config.get("hash_jobs")
-            or (self.repo and self.repo.config["core"].get("hash_jobs"))
+            self.config.get("checksum_jobs")
```
efiop, line 78: This is a nasty one caught by the verification. Config schema never allowed for `hash_jobs`, so we were effectively not using `checksum_jobs` at all all this time. But now we do.

PR #4570, opened 2020-09-14
Fix error during external outputs checkout

skshetry:
A user in discord complained that external outputs' checkout
was failing. Looks like, during refactor of `HashInfo`,
this was forgotten, and as no tests existed before, we had
a regression from 1.6.5 onwards.

This might have been avoided with Static Typings. :sparkles:

Discord context: https://discord.com/channels/485586884165107732/485596304961962003/755067231339020358
 
### DIFF
--- a/dvc/cache/local.py
+++ b/dvc/cache/local.py
@@ -94,11 +94,7 @@ class LocalCache(CloudCache):
     def already_cached(self, path_info):
         assert path_info.scheme in ["", "local"]
 
-        current = self.tree.get_hash(path_info)
-        if not current:
-            return False
-
-        return not self.changed_cache(current)
+        return super().already_cached(path_info)
 
     def _verify_link(self, path_info, link_type):
         if link_type == "hardlink" and self.tree.getsize(path_info) == 0:
--- a/tests/func/test_checkout.py
+++ b/tests/func/test_checkout.py
@@ -852,22 +852,18 @@ def test_checkouts_for_pipeline_tracked_outs(tmp_dir, dvc, scm, run_copy):
     assert set(dvc.checkout()["added"]) == {"bar", "ipsum"}
 
 
-def test_checkout_external_modify(tmp_dir, dvc, scm, s3):
-    def add_and_commit(paths, message):
-        scm.add(paths)
-        scm.commit(message)
+@pytest.mark.parametrize("workspace", [pytest.lazy_fixture("s3")], indirect=True)
+def test_checkout_external_modified_file(tmp_dir, dvc, scm, mocker, workspace):
+    # regression: happened when file in external output changed and checkout
+    # was attempted without force, dvc checks if it's present in its cache
+    # before asking user to remove it.
+    workspace.gen("foo", "foo")
+    dvc.add(str(workspace / "foo"), external=True)
+    scm.add(["foo.dvc"])
+    scm.commit("add foo")
 
-    s3.gen({"dir": {"foo": "foo"}})
+    workspace.gen("foo", "foobar")  # messing up the external outputs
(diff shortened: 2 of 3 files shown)


pared (approved):

efiop (approved):
Great catch!

tests/func/test_checkout.py
```diff
@@ -850,3 +850,24 @@ def test_checkouts_for_pipeline_tracked_outs(tmp_dir, dvc, scm, run_copy):
...
+    add_and_commit(["dir.dvc"], "add dir")
+
+    with tmp_dir.branch("branch1", new=True):
+        s3.gen({"dir": {"bar": "bar"}})
+        add_and_commit(["dir.dvc"], "modify dir")
+
+    dvc.checkout(force=True)
+    assert (s3 / "dir").is_dir()
+    assert (s3 / "dir" / "foo").read_text() == "foo"
+    assert not (s3 / "dir" / "bar").exists()
```
skshetry, line 873: This test does not seem to be testing the regression. Need to see when `safe_remove` is triggered.

skshetry: And, I forgot to `dvc add` as well. Not sure, why it's not failing now on the `master`.

dvc/cache/local.py
```diff
@@ -94,11 +94,7 @@ def hashes_exist(
     def already_cached(self, path_info):
         assert path_info.scheme in ["", "local"]
 
-        current = self.tree.get_hash(path_info)
-        if not current:
-            return False
-
-        return not self.changed_cache(current)
+        return super().already_cached(path_info)
```
efiop, line 97: :heart:

tests/func/test_checkout.py
```diff
@@ -850,3 +850,24 @@ def test_checkouts_for_pipeline_tracked_outs(tmp_dir, dvc, scm, run_copy):
 
     (tmp_dir / "ipsum").unlink()
     assert set(dvc.checkout()["added"]) == {"bar", "ipsum"}
+
+
+def test_checkout_external_modify(tmp_dir, dvc, scm, s3):
```
efiop, line 855:
```suggestion
@pytest.mark.parametrize("workspace", pytest.lazy_fixture("s3"), indirect=True)
def test_checkout_external_modify(tmp_dir, dvc, scm, workspace):
```

efiop: This will setup cache for you, and you can then use `workspace` object as remote tmp_dir

tests/func/test_checkout.py
```diff
@@ -850,3 +850,22 @@ def test_checkouts_for_pipeline_tracked_outs(tmp_dir, dvc, scm, run_copy):
...
+    # was attempted without force, dvc checks if it's present in its cache
+    # before asking user to remove it.
+    workspace.gen("foo", "foo")
+    dvc.add(str(workspace / "foo"), external=True)
+    scm.add(["foo.dvc"])
+    scm.commit("add foo")
+
+    workspace.gen("foo", "foobar")  # messing up the external outputs
+    mocker.patch("dvc.prompt.confirm", return_value=True)
+    dvc.checkout()
```
efiop, line 869: Maybe
```suggestion
    dvc.checkout(force=True)
```

skshetry: This bug is only in effect on `force = False`. When checkout is forced, we never check for `already_cached`.

skshetry: Instead, we just remove it, without checking if it's cached.

efiop:
Looks like https://pypi.org/project/google-api-python-client/ updated to 1.12.0 an hour ago and broke gdrive.

efiop:
@skshetry Thanks for fixing pydrive2! :pray:

Issue #4498, opened 2020-08-30
Experiments: checkpoints tracking


dmpetrov:
# Checkpoints

The deep learning training process is very slow for complex neuro net models. Checkpoints are often used to track the progress of the training. Usually, a checkpoint is created for every N (1-100) epochs. For example, user's code can save models for every 10 epochs:
- `checkpoint/model_10.h5`
- `checkpoint/model_20.h5`
- `checkpoint/model_30.h5`
...
- `checkpoint/model_240.h5`

**Epoche** is the abstraction to index and name the checkpoints. Usually, code is not limited by max epoch and runs forever until the job won't be canceled (Ctrl+C or SIGTERM or `kill 1234`).

The checkpoints give to users many benefits:
1. **[Resume training]** Stop the training and resume it back later (might be in another machine).
2. **[Forward learning]** Training with a new hypothesis (set of parameters) starting from an old checkpoint. Stop training, change code, or hyper-params and restart it back using the last `checkpoint/model_240.h5`.
3. **[Backward learning]** Return back to a previous checkpoint If starts overfitting. Then resume the training with slightly different params (see 2).

So, the final model can be a result of many steps and combinations of 1-3 (and not reproducible).

# Problems

This process does not fit to DVC model for several reasons:
1. DVC is designed to work with a clean namespace and file versions. So, all the checkpoints should be just a version of the same file like `./model.h5`.
2. A checkpoint is a circular dependency. DVC has some support of that with `--outs-persist`.
3. DVC was not designed for jobs that run forever (until it will be killed form outside).

# Solution

## Checkpoint versions

New experiments #2799 can be a good abstraction for checkpoints and can solve problem 1 (checkpoint file versions). Example:

```
$ dvc exp show --no-pager
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━
┃ Experiment                         ┃ accuracy ┃    loss ┃ time_process ┃ epoch
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━
│ workspace                          │  0.61124 │ 2.31573 │     2153.649 │  370
│ d1c200e (2020-08-27 11:18:07)      │  0.61124 │ 2.31573 │     2153.649 │  370
[...]

PR #4573, opened 2020-09-15
experiments: support committing to existing experiment branches

pmrowla:

* Prerequisite for #4498
* Adds support for committing to an existing experiment branch rather than always creating a new branch per experiment
* This will not be exposed via any `repro`/`run` commands, for now it will only be used when running checkpoint experiments

### DIFF
--- a/dvc/repo/experiments/__init__.py
+++ b/dvc/repo/experiments/__init__.py
@@ -196,6 +196,7 @@ class Experiments:
         params: Optional[dict] = None,
         branch: Optional[str] = None,
         allow_unchanged: Optional[bool] = False,
+        apply_workspace: Optional[bool] = True,
         **kwargs,
     ):
         """Stash changes from the current (parent) workspace as an experiment.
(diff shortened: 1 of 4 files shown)


dvc/repo/experiments/__init__.py
```diff
@@ -72,14 +73,17 @@ class Experiments:
 
     EXPERIMENTS_DIR = "experiments"
     PACKED_ARGS_FILE = "repro.dat"
-    STASH_MSG_PREFIX = "dvc-exp-"
+    STASH_MSG_PREFIX = "dvc-exp:"
     STASH_EXPERIMENT_RE = re.compile(
-        r"(?:On \(.*\): )dvc-exp-(?P<baseline_rev>[0-9a-f]+)$"
+        r"(?:On \(.*\): )"
+        r"dvc-exp:(?P<baseline_rev>[0-9a-f]+)(:(?P<branch>.+))?$"
```
pmrowla, line 79: `:` is an illegal character for git branch/tag/ref names so we can use it as a separator for this regex

dvc/repo/experiments/__init__.py
```diff
@@ -184,30 +190,54 @@ def _checkout_default_branch(self):
...
+        else:
+            rev = self.scm.get_rev()
+
+        if apply_workspace:
+            # patch user's workspace into experiments clone
+            #
+            # TODO: patching changes into an extended branch will require
+            # propagating changes down the full branch, since the workspace
+            # patch will be based on the original branch point, not the tip
+            # of the experiment branch
```
pmrowla, line 227: This will be addressed in a follow-up PR, still thinking about how to best handle the case where the user wants to edit some arbitrary checkpoint commit and then start a new branch with those changes

pared (approved):

efiop (approved):

Issue #4519, opened 2020-09-02
config: set cache per project at a global/system level


wdixon:
Our team starting use DVC, working on different projects and different machines using some shared servers - (large # of cpu/gpu/memory, etc..)

The shared cache config is somewhat cumbersome - as its specific to a group of repos...  One set of projects and remotes will use one cache, and another set would use a different cache, as there are different restrictions on data and who has access to the information.  We have specifically used dvc.local to assign cache.dir to keep path/machine specific attributes out of git; however, this does mean that folks must remember to set the local cache.dir whenever they clone or setup a new training project w/dvc import (otherwise they end up creating a copy of the data).  Users could set global - but this is cumbersome working with different caches and different projects (set/unset/etc/get wrong!).

What might be nice - if the cache could be configured based on the remote...  This way a system and or global configuration could associate the correct shared cache dir.

**somewhat better global config:**
['remote "prject_a"']
    url = s3://dvc-bucket/datasets/project_a_dataset
    endpointurl = https://some.endpoint.com
    profile = some-profile
    cache_dir = "/project_a/shared_dvc/cache

Even better would be to allow configuration based on url matching, avoiding the need for a any remote or profile naming convention...  This could enable the use of a system or global config that has bucket specific settings...  as the bucket itself defines access control, and already implies a grouping of what data might be shared in a cache.

**even better system or global config:**
['remote "s3://dvc-bucket-a/datasets/*"']
    cache_dir = "/project_a/shared_dvc/cache
    profile = some-profile

['remote "s3://dvc-bucket-b/*"']
    cache_dir = "/project_b/shared_dvc/cache
    profile = some-profile

shcheklein:
@wdixon thanks! Few questions here:

- How do you define which remote to use per project? Do you save the name in the project's config? As a workaround for configuring cache locations, I guess, you can come up with a convention around cache locations (using symlinks to actual location if needed) and save those into the project's config, WDYT? Something like - for a `remoteA`  we always have `/mnt/cache/remoteA`.

- How will the default DVC caching mechanism (when you run `dvc add`, `dvc run`, etc) pick which cache to use? The one attached to the default remote? Does it mean we'll need to add `--remote` to those commands as well?

- It feels that this can be a bit more general problem. Let's say we have a project with some configuration, name, etc - we can come up with a mechanism to match it in the config and override settings. In you case we would be matching based on the default remote path/name. I wonder if there are some similar mechanism in git, or some other tools?

[...]

Issue #4040, opened 2020-06-14
Introducing cache types: data, metrics and plots, run-cache and per-file


dmpetrov:
Today we use a single remote to store all the data (which can be redefined by `--remote` option).  However, there are different types of information in the cache:
1. Data files
2. Metrics and plots files
3. Run-cache files

There might be several reasons (data sensitivity or optimizations) to store these artifacts in different remotes:
1. **Data** cannot be moved outside of a company data center (outside of a country) while it's ok to use clouds (S3) for **metrics** and **run-cache** for performance or for simplifying the user experience.
2. **Data** needs to be duplicated among several buckets/clouds for redundancy while it is not a requirement for **metrics**.
3. **Metrics** and **run-cache** - tons of small files while data is large data files. Users might need to use separate storage for the optimization reason.
4. It might be required in the future when some remotes might need auto-compression #1239 or splitting data by blocks #829 while the other files don't need that.

It would be greater to introduce "types" of remote to group some of the cache types.

## Proposal ideas

This is the very first iteration on the subject and I don't have a clear proposal yet. We need to take a look at some analogs and possible solutions. How I see it for now:
```bash
$ dvc remove add --default --type data mydata ssh://some/dir
$ dvc remove add --default --type metrics,plots,run-cache mymeta s3://mybucket
$ dvc remove add --default --type data,metrics,plots --redundancy mybackup ssh://backup/dir
```

### Remote per file

This use case can be extended to per-file scenarios. Sometimes a special remote is required for a specific file (one bucket for data sources, the other for the pipeline derivative data sources). An extreme case - imported data sources. It would be great to have the information about remotes and use a single command `dvc push` to use all the remotes instead of specifying `--remove` option all the time.

### Change remotes globally

See #2960

### External workspaces

It might be also related to https://github.com/iterative/dvc/issues/3920

## Random ideas

Does it make sense to use the new `dvc.yaml` for storing all this information or part of these?

pared:
Related to #2095

Suor:
There is also a use case where we need to separate data as well. So on top of `--type` their might be some glob pattern or smth.

Issue #2095, opened 2019-06-05
Specify what should be pushed to different remotes


nirtiac:
I need to be able to push versioned trained models to a production server. Both data and models will be created on one machine and versioned using DVC. However, I have sensitive information in my training server data that must never be pushed off the server. 

As it stands, we are currently figuring out how to push models to a remote that can be tied back to their versioned data/code within DVC while never also pushing the data. This requires a solution outside of DVC. It would be great if DVC could provide the capability to specify file types that can or can't be pushed to a remote once committed.

Suor:
Sounds like this is not about file types, but a need to specify what to push in general. What if you have files of same type ones with sensitive data and others not? You have this nice property the latter being models, which won't always be the case.

nirtiac:
Yep, very true @Suor  - that would be even better!

shcheklein:
so, add a flag per output, something like `--out-local` to avoid pushing it to remote at all? it should be pretty straightforward. My only concern that number of different options for different types of outputs double every time. We should think about some general mechanism to specify options per output?

jorgeorpinel:
It seems we hear this feature request often on the chat, and there's even a very similar SO question: [How to use different remotes for different folders?
](https://stackoverflow.com/questions/58952962/how-to-use-different-remotes-for-different-folders): "I want my data and models stored in separate Google Cloud buckets. The idea is that I want to be able to share the data with others without sharing the models."

The current solution is to setup the remotes, and then use `dvc push -r {REMOTE} {data/model}` individually for each data/model file.

As suggested in this issue though, it would be useful to setup a map of certain outputs/directories/file types to specific default remotes. For example we could add a `--out` option to `dvc remote default` (somehow recorded in .dvc/config) so DVC knows that one output file should be pushed to a specific remote by default (unless overwritten with `dvc push -r`, of course. The option could be called something else like `--map` and accept not only outputs but any file/dir path or even file type filter (e.g. something like `**/*.csv`).

Suor:
Mapping could be done two ways:
- specify outputs, possibly patterns, in a remote config
- specify remote in an output definition in a stage file

dmpetrov:
@nirtiac for data from different storages or with different access permissions, you can use new DVC functionality - `dvc import`. This works only in multiple-repository environment (not monorepo) but it potentially gives you an elegant solution of separating data remotes.

[...]

PR #4581, opened 2020-09-20
Add the option for imports to be backed up to the remote

### LINKED ISSUE #4527: import: Allow pushing imported files to remote
Currently, files imported by `dvc import` come into the local cache, but do not get `push`ed to any remotes. This can be useful in some cases, however at other times it might be desirable to make full backups of imported files, while still retaining the ability to update the local copies via `dvc update`. 

Discussion with @efiop in Discord at https://discordapp.com/channels/485586884165107732/563406153334128681/751199400931491900 and https://discordapp.com/channels/485586884165107732/565699007037571084/751203534351106119.

To provide this option, it should be enough to just alter https://github.com/iterative/dvc/blob/79e8e4edafb8a4745200f0a62f49c29ef074bdea/dvc/output/base.py#L463-L467 to not always skip imports, according to some flag in the dvcfile. 

For the CLI, we though `dvc import --backup` conveys the meaning. For the DVCfile, an addition to `outs`:

```yaml
deps: ...
outs:
- md5: 123456
  path: my_file.txt
  backup: True
```

So for imports with `backup: True`, dvc knows:

* On update / import it should store the files in the local cache (like it already does)
* On push it should transfer them to the remote (this is new, but easy to implement by avoiding the code branch linked above)
* On fetch / pull treat these like normal files from e.g. `dvc add`, so do not interact with the import source, only with the DVC remote

The idea is for imports with `--backup` to be severed from their import source completely unless the user calls `dvc update` explicitly.

charlesbaynham:

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here: https://github.com/iterative/dvc.org/pull/1788

Closes #4527 - see this and Discord at https://discordapp.com/channels/485586884165107732/563406153334128681/751199400931491900 and https://discordapp.com/channels/485586884165107732/565699007037571084/751203534351106119 for some discussion.

### DIFF
--- a/dvc/command/imp.py
+++ b/dvc/command/imp.py
@@ -19,6 +19,7 @@ class CmdImport(CmdBase):
                 rev=self.args.rev,
                 no_exec=self.args.no_exec,
                 desc=self.args.desc,
+                store=self.args.store,
             )
         except DvcException:
             logger.exception(
@@ -80,6 +81,16 @@ def add_parser(subparsers, parent_parser):
         help=(
             "User description of the data (optional). "
             "This doesn't affect any DVC operations."
+        ),
+    )
+    import_parser.add_argument(
+        "-s",
+        "--store",
+        action="store_true",
+        default=False,
+        help=(
+            "Enable pushing the imported data to remote storage, and "
+            "pulling it from there (instead of reimporting)."
         ),
     )
     import_parser.set_defaults(func=CmdImport)
--- a/dvc/dependency/__init__.py
+++ b/dvc/dependency/__init__.py
@@ -51,12 +51,14 @@ DEP_MAP = {
 
 # NOTE: schema for dependencies is basically the same as for outputs, but
 # without output-specific entries like 'cache' (whether or not output is
-# cached, see -o and -O flags for `dvc run`) and 'metric' (whether or not
-# output is a metrics file and how to parse it, see `-M` flag for `dvc run`).
+# cached, see -o and -O flags for `dvc run`), 'metric' (whether or not output
+# is a metric file and how to parse it, see `-M` flag for `dvc run`) and
+# `store` (whether to push outputs to the remote)
 SCHEMA = output.SCHEMA.copy()
 del SCHEMA[BaseOutput.PARAM_CACHE]
 del SCHEMA[BaseOutput.PARAM_METRIC]
 del SCHEMA[BaseOutput.PARAM_DESC]
+del SCHEMA[BaseOutput.PARAM_STORE]
 SCHEMA.update(RepoDependency.REPO_SCHEMA)
 SCHEMA.update(ParamsDependency.PARAM_SCHEMA)
 
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -265,16 +265,32 @@ class UploadError(DvcException):
 
 
 class CheckoutError(DvcException):
-    def __init__(self, target_infos, stats=None):
+    def __init__(self, target_infos, stats=None, failed_not_stored=None):
         self.target_infos = target_infos
         self.stats = stats
-        targets = [str(t) for t in target_infos]
+        targets = []
+        for t in target_infos:
+            if failed_not_stored and t in failed_not_stored:
+                targets.append(
+                    "{} (marked with 'store: false' in its dvc file)".format(t)
+                )
+            else:
+                targets.append(str(t))
+
         m = (
-            "Checkout failed for following targets:\n{}\nIs your "
-            "cache up to date?\n{}".format(
+            "Checkout failed for following targets:\n{}\n\n"
+            "Is your cache up to date?\n{}".format(
                 "\n".join(targets), error_link("missing-files"),
             )
         )
+
+        if failed_not_stored:
+            m = m + (
+                "\n\nNote that 'store: false' outputs are not "
+                "searched for in remote storage, only in the cache. "
+                "See {}".format(error_link("checkout-no-store"),)
+            )
+
         super().__init__(m)
 
 
--- a/dvc/output/base.py
+++ b/dvc/output/base.py
@@ -55,6 +55,15 @@ class OutputIsIgnoredError(DvcException):
         super().__init__(f"Path '{match.file}' is ignored by\n{lines}")
 
 
+class OutputStoreCacheIncompatible(DvcException):
+    def __init__(self, path):
+        msg = f"""Incompatible options for output '{path}'.
+    The .dvc file specifies an output with both `cache == False` and
+    `store == True` which is not possible to satisfy.
+    """
+        super().__init__(msg)
+
+
 class BaseOutput:
     IS_DEPENDENCY = False
 
@@ -76,6 +85,7 @@ class BaseOutput:
     PARAM_PLOT_HEADER = "header"
     PARAM_PERSIST = "persist"
     PARAM_DESC = "desc"
+    PARAM_STORE = "store"
 
     METRIC_SCHEMA = Any(
         None,
@@ -105,6 +115,7 @@ class BaseOutput:
         persist=False,
         checkpoint=False,
         desc=None,
+        store=True,
     ):
         self._validate_output_path(path, stage)
         # This output (and dependency) objects have too many paths/urls
@@ -128,6 +139,8 @@ class BaseOutput:
         self.use_cache = False if self.IS_DEPENDENCY else cache
         self.metric = False if self.IS_DEPENDENCY else metric
         self.plot = False if self.IS_DEPENDENCY else plot
+        self.store = True if self.IS_DEPENDENCY else store
+
         self.persist = persist
         self.checkpoint = checkpoint
         self.desc = desc
@@ -328,6 +341,9 @@ class BaseOutput:
         if self.checkpoint:
             ret[self.PARAM_CHECKPOINT] = self.checkpoint
 
+        if self.store == self.stage.is_repo_import:
+            ret[self.PARAM_STORE] = self.store
+
         return ret
 
     def verify_metric(self):
@@ -484,10 +500,16 @@ class BaseOutput:
         """
 
         if not self.use_cache:
+            if self.store:
+                raise OutputStoreCacheIncompatible(self)
             return NamedCache()
 
-        if self.stage.is_repo_import:
+        if not self.store:
             cache = NamedCache()
+            if not self.stage.is_repo_import:
+                return cache
+            # If this out isn't backed up but it is an import, then it's an
+            # external dependancy and can be retrieved from its source
             (dep,) = self.stage.deps
             cache.external[dep.repo_pair].add(dep.def_path)
             return cache
--- a/dvc/repo/checkout.py
+++ b/dvc/repo/checkout.py
@@ -89,6 +89,7 @@ def checkout(
         "deleted": [],
         "modified": [],
         "failed": [],
+        "failed_not_stored": [],
     }
     if not targets:
         targets = [None]
@@ -117,8 +118,13 @@ def checkout(
             for key, items in result.items():
                 stats[key].extend(_fspath_dir(path) for path in items)
 
-    if stats.get("failed"):
-        raise CheckoutError(stats["failed"], stats)
+    if stats.get("failed") or stats.get("failed_not_stored"):
+        raise CheckoutError(
+            stats["failed"] + stats["failed_not_stored"],
+            stats,
+            failed_not_stored=stats["failed_not_stored"],
+        )
 
     del stats["failed"]
+    del stats["failed_not_stored"]
     return stats
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -238,10 +238,9 @@ class Stage(params.StageParams):
 
     @property
     def is_repo_import(self):
-        if not self.is_import:
-            return False
-
-        return isinstance(self.deps[0], dependency.RepoDependency)
+        return len(self.deps) == 1 and isinstance(
+            self.deps[0], dependency.RepoDependency
+        )
 
     @property
     def is_checkpoint(self):
@@ -540,6 +539,8 @@ class Stage(params.StageParams):
                 return None, []
             return "modified" if modified else "added", [out.path_info]
         except CheckoutError as exc:
+            if not out.store:
+                return "failed_not_stored", exc.target_infos
             return "failed", exc.target_infos
 
     @rwlocked(read=["deps", "outs"])
--- a/tests/func/test_data_cloud.py
+++ b/tests/func/test_data_cloud.py
@@ -6,6 +6,7 @@ import pytest
 from flaky.flaky_decorator import flaky
 
 from dvc.cache import NamedCache
+from dvc.exceptions import DvcException
 from dvc.external_repo import clean_repos
 from dvc.main import main
 from dvc.remote.base import (
@@ -305,7 +306,7 @@ def test_verify_hashes(
 @pytest.mark.parametrize(
     "erepo", [pytest.lazy_fixture("git_dir"), pytest.lazy_fixture("erepo_dir")]
 )
-def test_pull_git_imports(tmp_dir, dvc, scm, erepo):
+def test_pull_git_imports(tmp_dir, dvc, scm, erepo, local_remote):
     with erepo.chdir():
         erepo.scm_gen({"dir": {"bar": "bar"}}, commit="second")
         erepo.scm_gen("foo", "foo", commit="first")
@@ -314,6 +315,7 @@ def test_pull_git_imports(tmp_dir, dvc, scm, erepo):
     dvc.imp(os.fspath(erepo), "dir", out="new_dir", rev="HEAD~")
 
     assert dvc.pull()["fetched"] == 0
+    assert dvc.push() == 0
 
     for item in ["foo", "new_dir", dvc.cache.local.cache_dir]:
         remove(item)
@@ -327,6 +329,72 @@ def test_pull_git_imports(tmp_dir, dvc, scm, erepo):
 
     assert (tmp_dir / "new_dir").exists()
     assert (tmp_dir / "new_dir" / "bar").read_text() == "bar"
+
+
+@pytest.mark.parametrize(
+    "erepo", [pytest.lazy_fixture("git_dir"), pytest.lazy_fixture("erepo_dir")]
+)
+def test_pull_git_imports_with_store(tmp_dir, dvc, scm, erepo, local_remote):
+    with erepo.chdir():
+        erepo.scm_gen({"dir": {"bar": "bar"}}, commit="second")
+        erepo.scm_gen("foo", "foo", commit="first")
+
+    dvc.imp(os.fspath(erepo), "foo", store=True)
+    dvc.imp(os.fspath(erepo), "dir", out="new_dir", store=True, rev="HEAD~")
+
+    assert dvc.pull()["fetched"] == 0
+    assert dvc.push() == 3
+
+    for item in ["foo", "new_dir", dvc.cache.local.cache_dir]:
+        remove(item)
+    os.makedirs(dvc.cache.local.cache_dir, exist_ok=True)
+    clean_repos()
+
+    # Close the repo since otherwise Windows holds references open which
+    # prevent deletion
+    erepo.close()
+    remove(erepo)
+
+    assert dvc.pull(force=True)["fetched"] == 2
+
+    assert (tmp_dir / "foo").exists()
+    assert (tmp_dir / "foo").read_text() == "foo"
+
+    assert (tmp_dir / "new_dir").exists()
+    assert (tmp_dir / "new_dir" / "bar").read_text() == "bar"
+
+
+def test_store_false_pull_error(tmp_dir, dvc, scm, local_remote, caplog):
+    tmp_dir.dvc_gen("foo", "Locally added", commit="add no-store file")
+    tmp_dir.dvc_gen("bar", "Another local file", commit="add normal file")
+
+    dvc.push()
+    clean(["foo", "bar"], dvc)
+    dvc.pull()
+
+    config = load_yaml(tmp_dir / "foo.dvc")
+    config["outs"][0]["store"] = False
+    dump_yaml(tmp_dir / "foo.dvc", config)
+
+    dvc.push()
+    clean(["foo", "bar"], dvc)
+
+    missing_error = (
+        "Cache 'md5: d089d68656330fd712dad63400c8710f' not found. "
+        "File 'foo' won't be created"
+    )
+    missing_error_hints = (
+        "foo (marked with 'store: false' in its dvc file)",
+        (
+            "Note that 'store: false' outputs are not searched for in remote "
+            "storage, only in the cache. See"
+        ),
+    )
+    caplog.clear()
+    with pytest.raises(DvcException) as excinfo:
+        dvc.pull()
+    assert missing_error in caplog.text
+    assert all(h in str(excinfo.value) for h in missing_error_hints)
 
 
 def test_pull_external_dvc_imports(tmp_dir, dvc, scm, erepo_dir):
--- a/dvc/hash_info.py
+++ b/dvc/hash_info.py
@@ -21,6 +21,9 @@ class HashInfo:
     def __bool__(self):
         return bool(self.value)
 
+    def __str__(self):
+        return f"{self.name}: {self.value}"
+
     @classmethod
     def from_dict(cls, d):
         _d = d.copy() if d else {}
--- a/dvc/output/__init__.py
+++ b/dvc/output/__init__.py
@@ -69,6 +69,7 @@ SCHEMA[BaseOutput.PARAM_CHECKPOINT] = bool
 SCHEMA[HashInfo.PARAM_SIZE] = int
 SCHEMA[HashInfo.PARAM_NFILES] = int
 SCHEMA[BaseOutput.PARAM_DESC] = str
+SCHEMA[BaseOutput.PARAM_STORE] = bool
 
 
 def _get(
@@ -81,6 +82,7 @@ def _get(
     persist=False,
     checkpoint=False,
     desc=None,
+    store=True,
 ):
     parsed = urlparse(p)
 
@@ -97,6 +99,7 @@ def _get(
             persist=persist,
             checkpoint=checkpoint,
             desc=desc,
+            store=store,
         )
 
     for o in OUTS:
@@ -112,6 +115,7 @@ def _get(
                 persist=persist,
                 checkpoint=checkpoint,
                 desc=desc,
+                store=store,
             )
     return LocalOutput(
         stage,
@@ -124,6 +128,7 @@ def _get(
         persist=persist,
         checkpoint=checkpoint,
         desc=desc,
+        store=store,
     )
 
 
@@ -131,6 +136,7 @@ def loadd_from(stage, d_list):
     ret = []
     for d in d_list:
         p = d.pop(BaseOutput.PARAM_PATH)
+        store = d.pop(BaseOutput.PARAM_STORE, not stage.is_repo_import)
         cache = d.pop(BaseOutput.PARAM_CACHE, True)
         metric = d.pop(BaseOutput.PARAM_METRIC, False)
         plot = d.pop(BaseOutput.PARAM_PLOT, False)
@@ -142,6 +148,7 @@ def loadd_from(stage, d_list):
                 stage,
                 p,
                 info=d,
+                store=store,
                 cache=cache,
                 metric=metric,
                 plot=plot,
@@ -161,6 +168,7 @@ def loads_from(
     plot=False,
     persist=False,
     checkpoint=False,
+    store=True,
 ):
     return [
         _get(
@@ -172,6 +180,7 @@ def loads_from(
             plot=plot,
             persist=persist,
             checkpoint=checkpoint,
+            store=store,
         )
         for s in s_list
     ]
--- a/dvc/repo/imp_url.py
+++ b/dvc/repo/imp_url.py
@@ -19,6 +19,7 @@ def imp_url(
     frozen=True,
     no_exec=False,
     desc=None,
+    store=True,
 ):
     from dvc.dvcfile import Dvcfile
     from dvc.stage import Stage, create_stage, restore_meta
@@ -34,14 +35,15 @@ def imp_url(
     ):
         url = relpath(url, wdir)
 
+    kwargs = {"outs": [out]} if store else {"outs_no_store": [out]}
     stage = create_stage(
         Stage,
         self,
         fname or path,
         wdir=wdir,
         deps=[url],
-        outs=[out],
         erepo=erepo,
+        **kwargs,
     )
     restore_meta(stage)
     if stage.can_be_skipped:
--- a/dvc/stage/utils.py
+++ b/dvc/stage/utils.py
@@ -54,6 +54,7 @@ def fill_stage_outputs(stage, **kwargs):
         "outs_no_cache",
         "outs",
         "checkpoints",
+        "outs_no_store",
     ]
 
     stage.outs = []
@@ -62,6 +63,7 @@ def fill_stage_outputs(stage, **kwargs):
             stage,
             kwargs.get(key, []),
             use_cache="no_cache" not in key,
+            store="no_store" not in key and "no_cache" not in key,
             persist="persist" in key,
             metric="metrics" in key,
             plot="plots" in key,
--- a/tests/func/test_checkout.py
+++ b/tests/func/test_checkout.py
@@ -617,6 +617,7 @@ def test_checkout_stats_on_failure(tmp_dir, dvc, scm):
     assert exc.value.stats == {
         **empty_checkout,
         "failed": ["foo"],
+        "failed_not_stored": [],
         "modified": ["other"],
     }
 
--- a/tests/func/test_import.py
+++ b/tests/func/test_import.py
@@ -31,6 +31,31 @@ def test_import(tmp_dir, scm, dvc, erepo_dir):
         "rev_lock": erepo_dir.scm.get_rev(),
     }
 
+    outs = stage.dumpd()["outs"]
+    assert outs[0]["path"] == "foo_imported"
+    assert "store" not in outs[0]
+    assert "md5" in outs[0]
+
+
+def test_import_with_store(tmp_dir, scm, dvc, erepo_dir):
+    with erepo_dir.chdir():
+        erepo_dir.dvc_gen("foo", "foo content", commit="create foo")
+
+    stage = dvc.imp(os.fspath(erepo_dir), "foo", "foo_imported", store=True)
+
+    assert os.path.isfile("foo_imported")
+    assert (tmp_dir / "foo_imported").read_text() == "foo content"
+    assert scm.repo.git.check_ignore("foo_imported")
+    assert stage.deps[0].def_repo == {
+        "url": os.fspath(erepo_dir),
+        "rev_lock": erepo_dir.scm.get_rev(),
+    }
+
+    outs = stage.dumpd()["outs"]
+    assert outs[0]["path"] == "foo_imported"
+    assert outs[0]["store"]
+    assert "md5" in outs[0]
+
 
 @pytest.mark.parametrize("src_is_dvc", [True, False])
 def test_import_git_file(tmp_dir, scm, dvc, git_dir, src_is_dvc):
--- a/tests/unit/command/test_imp.py
+++ b/tests/unit/command/test_imp.py
@@ -16,6 +16,7 @@ def test_import(mocker):
             "version",
             "--desc",
             "description",
+            "--store",
         ]
     )
     assert cli_args.func == CmdImport
@@ -33,6 +34,7 @@ def test_import(mocker):
         rev="version",
         no_exec=False,
         desc="description",
+        store=True,
     )
 
 
@@ -51,6 +53,7 @@ def test_import_no_exec(mocker):
             "--no-exec",
             "--desc",
             "description",
+            "--store",
         ]
     )
 
@@ -67,4 +70,5 @@ def test_import_no_exec(mocker):
         rev="version",
         no_exec=True,
         desc="description",
+        store=True,
     )


shcheklein:
@charlesbaynham thanks for the PR!! 🙏 

@efiop @charlesbaynham two questions to clarify:

1. Why don't we use `push: true/false` or `save: true/false` ? To make it a bit more generic since it can be applied to other outputs as well, right? I remember we had tickets/requests to turn off pushing some artifacts. This PR can cover that to my mind.

2. Do we take into account this new field in the `dvc gc -c`?

jorgeorpinel:
Hi! I also have some questions (left in the issue: https://github.com/iterative/dvc/issues/4527#issuecomment-696260103) — mostly arguing against having an explicit `backup` in all `outs` going fwd.

charlesbaynham:
@jorgeorpinel from your comments in the linked issue:
> > The idea is for imports with --backup to be severed from their import source completely unless the user calls dvc update explicitly.
> 
> Does `update` reset backup to false? What about re-importing (running `import` again but without `--backup`?

`update` does not reset `backup` to false, it leaves it as it was. Re-importing (without the `--backup` flag), however, does remove the flag. So you can use `import` to alter the settings of the import, or `update` to retrigger it with whatever settings it was set up with.

charlesbaynham:
> > The backup entry in outs should be added for all outs, with the same meaning... without imports having to be treated differently to normal files.
> > To keep backwards compatibility, dvc files without backup entries should have default values of backup = False for imports and backup = True for everything else.
> 
> So if we're still having to differentiate .dvc files based on contents of `deps` (regular vs. import) for backward compatibility, doesn't that defeat the point of adding the field to all outs going fwd for consistency? I'm of the inclination that `backup` should only have an effect on import .dvc files, TBH as it seems pretty edge and not something mainstream we want to be super visible.

In a way yes, we are still having to go to `deps` in order to find out what to do with our `outs`. But the difference is that we're only doing in the dvc code once - when `outs` gets loaded from the file - rather than every time we need to use it. That keeps concerns separate (i.e. dependencies vs. outputs) since code which only deals with outputs can now be decoupled from the `deps` logic fully. 

In terms of clutter in the `.dvc` file, the only files which would be altered by this change would be those that explicitly use the `--backup` option. And if a canny user wants to manually add a `backup: false` field to their `dvc add`ed file then I don't see why not let them: the logic is cleaner and easier and it won't cause them any surprises. 

> Also, what about `outs` in `dvc.yaml` files? would those also always have the backup field? (This would affect docs BTW) — again, feels like a big change for a small use case.

Again, `outs` could have the field if you wanted to add it, but they wouldn't get it by default.

charlesbaynham:
@shcheklein 

> @charlesbaynham thanks for the PR!! 🙏
You're welcome!

> @efiop @charlesbaynham two questions to clarify:
> 
>     1. Why don't we use `push: true/false` or `save: true/false` ? To make it a bit more generic since it can be applied to other outputs as well, right? I remember we had tickets/requests to turn off pushing some artifacts. This PR can cover that to my mind.

Yes, I'm not totally sold on the naming yet. We discussed `push` but decided that it would be confusing as a flag for the import command. (`dvc import --push ...` implies an action to me, but actually it's marking the file for an action in the future). `cache` is taken already. `save` is quite good. I quite like having the same flag in `outs` and in the `import` command, but that's not essential. 

Very open to suggestions for better names!

>     2. Do we take into account this new field in the `dvc gc -c`?

That's not something I've thought about at all I'm afraid. It might "just work" since the alterations to `get_used_cache` should mark the imports files as being part of the cache. But I don't know anything about the `dvc gc` code so I'm not sure.

jorgeorpinel:
> the difference is that we're only doing in the dvc code once - when outs gets loaded from the file - rather than every time we need to use it

OK, but it's too low level IMO. Implementation shouldn't dictate the requirements 🙂. The question is how should it work so it provides the desired utility, without complicating the other 95% of use cases (that may not need it)? I still think the `backup` field only makes sense for imports, at least for now. Up to @shcheklein and @efiop though.

> the only files which would be altered by this change would be those that explicitly use the --backup option

Sorry, I'm confused. I thought that "The backup entry in outs should be added for all outs" (from https://github.com/iterative/dvc/issues/4527#issuecomment-695189230)

> I don't see why not let them: the logic is cleaner and easier and it won't cause them any surprises.

It's not about not letting users do something hypothetical. It's just that applying this globally seems out of scope for #4527. It's also speculative to assume that such change would be welcome generally. We don't know if there's a need for `backup: false` — you can just not use remote storage, for example.

> > what about outs in dvc.yaml files?
>
> Again, outs could have the field if you wanted

So this has been tested with dvc.yaml files?

Thanks

jorgeorpinel:
> Very open to suggestions for better names!

I don't like `save`. How about `store` (as in "remote storage")?

jorgeorpinel:
Oh and finally, from https://github.com/iterative/dvc.org/pull/1788#pullrequestreview-492848747:

> > what happens to a regular output after it has been pushed if you then set it to backup:false manually, and run pull?
>
> you get an error for pull/fetch if the file isn't already in your local cache:
> ```
> WARNING: Cache 'HashInfo(name='md5', value='bbe02f946d5455d74616fc9777557c22', dir_info=None)' not found. > > File 'file.txt' won't be created.                                          
> 1 file failed                                                                                                                                                                        
> ERROR: failed to pull data from the cloud - Checkout failed for following targets:
> file.txt
> Is your cache up to date?
> ```

Could the error message be less cryptic and provide a hint on what's the cause? (backup: false) But see, this change can get complicated very quickly 😋

charlesbaynham:
> > the difference is that we're only doing in the dvc code once - when outs gets loaded from the file - rather than every time we need to use it
> 
> OK, but it's too low level IMO. Implementation should probably not define requirements 🙂. The question is how should it work so it provides the desired utility, without complicating the other 95% of use cases (that may not need it)? I still think the `backup` field only makes sense for imports, at least for now. Up to @shcheklein and @efiop though.
> 
> > the only files which would be altered by this change would be those that explicitly use the --backup option
> 
> Sorry, I'm confused. I thought that "The backup entry in outs should be added for all outs" (from [#4527 (comment)](https://github.com/iterative/dvc/issues/4527#issuecomment-695189230))

Ah, I realise I've been unclear. When I wrote that issue, I was expecting to go through adding `backup: true` to all outs. However, this PR doesn't do that: it only adds it to files where the presence of the `--backup` flag changes the `backup` setting away from its default. Its default is `true` for all files except imports, where it's `false`. So the only dvc files that will be altered are imports which were made with the `--backup` flag, so no changes to existing repo. 

What I mean by "added to all outs" is that it's part of the schema, and will be understood by dvc if it's manually added to an `out` of something that isn't an import. But I haven't added anything on the CLI to achieve this, since it's an obscure feature with cryptic error messages (see https://github.com/iterative/dvc/pull/4581#issuecomment-696538097). So for someone only using dvc via the cli, they'll never see `backup` in their dvc files unless they call `import --backup`.

charlesbaynham:
> Oh and finally, from [iterative/dvc.org#1788 (review)](https://github.com/iterative/dvc.org/pull/1788#pullrequestreview-492848747):
> 
> > > what happens to a regular output after it has been pushed if you then set it to backup:false manually, and run pull?
> > 
> > 
> > you get an error for pull/fetch if the file isn't already in your local cache:
> > ```
> > WARNING: Cache 'HashInfo(name='md5', value='bbe02f946d5455d74616fc9777557c22', dir_info=None)' not found. > > File 'file.txt' won't be created.                                          
> > 1 file failed                                                                                                                                                                        
> > ERROR: failed to pull data from the cloud - Checkout failed for following targets:
> > file.txt
> > Is your cache up to date?
> > ```
> 
> Could the error message be less cryptic and provide a hint on what's the cause? (backup: false) But see, this change can get complicated very quickly 😋

Can do if you think it's worth it. I didn't write that error message, it's coming from `checkout` - it's the normal error if you try to checkout when your cache is missing a file. 

I feel like if a power user wants to alter the dvc files under-the-hood then they shouldn't expect finely crafted error messages. But if that's something you'd usually support, I'm sure we can add some code to `checkout` which identifies `backup: false` stages.

charlesbaynham:
OK, name poll! @efiop , @shcheklein , @jorgeorpinel could you please vote with emojis for the name of the flag to be added to `import`/entry to `outs`:

`backup` = 🚀
`push` = 🎉
`save` = 😄
`store` = ❤️

shcheklein:
> We don't even know if there's a need for backup: false — you can just not use remote storage, for example.

Good point. The point is, that we do have tickets/requests for this. Something like: "I want to push everything except my models ... or except data." Usually because of some regulations/security.

Another **potentially** relevant issue- support remote per output (type). I.e. push models into one remote, data into another etc.

Before we introduce any flags into DVC files (that we'll have to support, document etc), I would def spend a few hours trying to see if we can generalize this.

charlesbaynham:
@shcheklein Do you have any examples of these requests? I had a browse through the issues but I couldn't find any. I did find this one (https://github.com/iterative/dvc/issues/4040) which is the separate-remotes-by-data-type issue.

shcheklein:
@charlesbaynham sure, some of them that I was able to find within a few minutes: 

https://github.com/iterative/dvc/issues/2095
https://github.com/iterative/dvc/issues/4040
https://stackoverflow.com/questions/58952962/how-to-use-different-remotes-for-different-folders

A lot of requests come from Discord or "use-case" discussion, os it's not that easy to find them usually.

jorgeorpinel:
Thanks for the clarifications @charlesbaynham !

> it's an obscure feature with cryptic error messages
> if a power user wants to alter the dvc files under-the-hood then they shouldn't expect finely crafted error messages

It won't be that obscure after we document it 🙂 (merging iterative/dvc.org/pull/1788)

> I didn't write that error ... it's the normal error if you try to checkout when your cache is missing a file.
> we can add some code to checkout which identifies backup: false stages

Got it. I think it would be worth handling the specific case, if it's not too hard (implementation can dictate this req. 😉)

> we do have tickets/requests for this. Something like: "I want to push everything except my models
> support remote per output

True @shcheklein! But like you mentioned, most such requests are for configuring default remotes per output (#4519 is another example). If that supports `remote: null/false`, then it's equivalent to the proposed `store: false` here. The Q is: in that case, should `backup/store: false` have the same effect? Or be ignored by non-imports.

I still incline to only affect imports, but I have no strong opinion after reconsidering.

charlesbaynham:
@jorgeorpinel You wrote in #2095 :

> so, add a flag per output, something like --out-local to avoid pushing it to remote at all? it should be pretty straightforward. My only concern that number of different options for different types of outputs double every time. We should think about some general mechanism to specify options per output?

I completely agree about the doubling options in 

https://github.com/iterative/dvc/blob/0cef951257f0d670c321c44364cfd0cdb2339c55/dvc/stage/utils.py#L43-L68

and in the associated CLI. That's exactly what I had to do here, although I didn't add `persist` or `no_cache` variants since `import` won't create these, and that's currently the only way of making `no_backup` files via the CLI. I also didn't add `--outs-no-backup` to `dvc run`, although I could do. But a more scalable method would definitely be better, especially if we're considering adding differing default remotes etc.

For the `dvc` file, how about a field called `remote`or `push_remote`  for each `out`? This could take the special values `true` to mean "push like normal" or `false` to mean "don't push at all" (equivalent to `backup: false`). In the future, it could also take strings (e.g. "data") to mean "only push to the remote called "data".

jorgeorpinel:
> Ivan wrote that, not me 😋 

My take on this is the scope keeps growing and getting complicated. Probably just implement backup mode for imports as a first step, then decide about the rest in #2095?

charlesbaynham:
> (Ivan wrote that, not me 😋)
> 
> My take on this is the scope keeps growing and getting complicated. Probably just implement backup mode for imports as a first step, then decide about the rest in #2095?

So he did, sorry @shcheklein ! The true / false setting for "backup mode" in the dvc file (name tbc) could later be extended to also accept (a list of) named remotes for #2095 later on. If that's the plan, we should choose the name accordingly now.

shcheklein:
> The true / false setting for "backup mode" in the dvc file (name tbc) could later be extended to also accept (a list of) named remotes for #2095 later on. If that's the plan, we should choose the name accordingly now.

yep, that's exactly my concerns/points! Expanding YAML schema is one of those things that we should be doing carefully- since we'll have to support it, it will complicate code potentially (e.g. `if import_stage and some_flag = true` is not as good as a general code `if `out.push`: add_to_array_to_push`)

Also, even if we don't cover different remote names here (it can be stretching it too far), we should at least consider covering a very close case- optionally disable push/save for regular outputs.

jorgeorpinel:
I think the name `store` still works for all cases. For now it can just accept true/false (and apply to imports only). Later it can accept remote names and apply to all outs.

jorgeorpinel:
BTW if/when `store` applies to all outs, I think the default value should be `true` everywhere including for imports, but import should generate a .dvc file with `store: false`. That would address my earlier consistency concerns.

jorgeorpinel:
p.s. I'm still of the idea it should only apply to imports for now though.

charlesbaynham:
> 
> 
> BTW if/when `store` applies to all outs, I think the default value should be `true` everywhere including for imports, but `dvc import(-url)` should generate a .dvc file with explicit `store: false` by default. That would address my earlier consistency concerns (no need to differentiate import .dvc files from other outs).

That's how I originally wanted to do it, but this causes a problem for backwards compatibility. You need a way to migrate existing import `.dvc` files without changing their behaviour, which means you still need some code which uses the old logic.

charlesbaynham:
I've renamed `backup` to `store` as I think we've agreed. Current status is that `store` :

* Defaults to `true` for normal files and to `false` for imports
* Is only added to DVC files by the CLI when you call `dvc import --store` (otherwise, the defaults reproduce current behaviour)
* **Will** affect normal files if `store: false` is added manually to the DVC file, preventing them from being pushed to any remote

@jorgeorpinel I've also added a hint to the error message. You now get e.g.:

```
$ dvc pull
WARNING: Cache 'HashInfo(name='md5', value='31ebdfce8b77ac49d7f5506dd1495830', dir_info=None)' not found. File 'no_store.txt' won't be created.                                
1 file failed_not_stored                                                                                                                                                       
ERROR: failed to pull data from the cloud - The following files are marked as `store: false` so the remote was not searched, but are not present in local cache:
no_store.txt

Checkout failed for following targets:
no_store.txt
Is your cache up to date?
```

dvc/exceptions.py
```diff
@@ -261,6 +261,15 @@ def __init__(self, target_infos, stats=None):
...
             )
         )
+
+        if failed_not_stored:
+            prepend = (
+                "The following files are marked as `store: false` so the "
+                "remote was not searched, but are not present in local "
+                "cache:\n{}".format("\n".join(failed_not_stored))
+            )
+            m = prepend + "\n\n" + m
```
jorgeorpinel, line 271:
```suggestion
            m = (
                "The following outputs are marked as `store: false` so "
                "remote storage was disabled, and are not present in the "
                "cache:\n{}\n\n".format("\n".join(failed_not_stored))
            ) + m
```
> May need wrapping?

jorgeorpinel: And how about also introducing a link to somewhere in the docs where this will be explained? (In anticipation of iterative/dvc.org/pull/1788)

jorgeorpinel:
```
ERROR: failed to pull data from the cloud - The following files...
no_store.txt

Checkout failed for following targets:
no_store.txt
```

Those 2 new lines and repeated list items may be confusing though. Not sure how to improve the UI here and it's not a blocking problem but would be nice to try 🙂

dvc/output/base.py
```diff
@@ -53,6 +53,15 @@ def __init__(self, match):
         super().__init__(f"Path '{match.file}' is ignored by\n{lines}")
 
 
+class OutputStoreCacheIncompatible(DvcException):
+    def __init__(self, path):
+        msg = f"""Incompatible options for output '{path}'.
+    The dvc file specifies an output with both "cache == False" and
+    "store == True" which is not possible to satisfy.
+    """
```
jorgeorpinel, line 61: Looks like the indentation is off but not sure.

tests/func/test_data_cloud.py
```diff
@@ -308,6 +310,68 @@ def test_pull_git_imports(tmp_dir, dvc, scm, erepo):
...
+    clean(["foo"], dvc)
+
+    missing_error = (
+        "Cache 'HashInfo(name='md5', "
+        "value='d089d68656330fd712dad63400c8710f', dir_info=None)' not found. "
+        "File 'foo' won't be created"
+    )
+    missing_error_hint = (
+        "The following files are marked as `store: false` so the remote was "
+        "not searched, but are not present in local cache:\nfoo"
```
jorgeorpinel, line 366: I guess this would need
```suggestion
        "The following outputs are marked as `store: false` so remote "
        "storage was disabled, and are not present in the cache:\nfoo"
```

jorgeorpinel (approved):
Left some final comments but LGTM in general. Let's just do another pass of iterative/dvc.org/pull/1788 after this is approved by the engineering team, please (if you don't have the capacity just ping me there and I'll take care of it, np). Thank you @charlesbaynham

dvc/output/base.py
charlesbaynham: I think that's right: I copied the other examples I saw. That'll produce:

```
Incompatible options for output 'path'.
    The dvc file specifies an output with both "cache == False" and
    "store == True" which is not possible to satisfy.
```

dvc/exceptions.py
charlesbaynham: I've added a docs link like this:
```
The following outputs are marked as `store: false` so remote storage was disabled, and are not present in the cache:
foo
<https://error.dvc.org/checkout-failed-no-store>
Checkout failed for following targets:
foo
Is your cache up to date?
``` 
and a section to the docs in https://github.com/iterative/dvc.org/pull/1788.

dvc/exceptions.py
```diff
@@ -261,6 +261,17 @@ def __init__(self, target_infos, stats=None):
                 "\n".join(targets), error_link("missing-files"),
             )
         )
+
+        if failed_not_stored:
+            m = (
+                "The following outputs are marked as `store: false` so "
+                "remote storage was disabled, and are not present in the "
+                "cache:\n{}\n{}\n".format(
```
jorgeorpinel, line 269:
```suggestion
                "The following outputs are marked as `store: false` (which"
                "disables remote storage) and they are not present in the "
                "cache:\n{}\n{}\n".format(
```

jorgeorpinel: > The following outputs are marked as `store: false` so ...
foo
\<https://error.dvc.org/checkout-failed-no-store\>
Checkout failed for following targets:
foo
Is your cache up to date?
\<https://error.dvc.org/missing-files\>

Hmmm yeah this is a little confusing. I guess it's best to have too much info rather than too little but ideally this should be closer to

> Checkout failed for **the** following targets:
> \* foo (marked with `store: false` in foo.dvc)
> \* bar
>
> Is your cache up to date?
\<https://error.dvc.org/missing-files\>
>
> Note that `store: false` outputs are not searched for in remote storage, only in the cache, see \<https://error.dvc.org/checkout-failed-no-store\>.

But even that is a little long... Not sure! I'm not trying to block this merge either, could be addressed separately I suppose.

charlesbaynham: Done! It's now

```
Checkout failed for following targets:
foo (marked with 'store: false' in dvc file)

Is your cache up to date?

Note that 'store: false' outputs are not searched for in remote storage, only in the cache. See <https://error.dvc.org/checkout-failed-no-store>
```

I didn't include a link to the DVC file since that'd involve a bit of refactoring which didn't seem worth it.

dvc/exceptions.py
```diff
@@ -251,16 +251,32 @@ def __init__(self, amount):
...
-    def __init__(self, target_infos, stats=None):
+    def __init__(self, target_infos, stats=None, failed_not_stored=None):
         self.target_infos = target_infos
         self.stats = stats
-        targets = [str(t) for t in target_infos]
+        targets = []
+        for t in target_infos:
+            if failed_not_stored and t in failed_not_stored:
+                targets.append(
+                    "{} (marked with 'store: false' in dvc file)".format(t)
```
jorgeorpinel, line 261:
```suggestion
                    "{} (marked with 'store: false' in its .dvc file)".format(t)
```

tests/func/test_data_cloud.py
```diff
@@ -308,6 +310,72 @@ def test_pull_git_imports(tmp_dir, dvc, scm, erepo):
...
+    dvc.push()
+    clean(["foo", "bar"], dvc)
+
+    missing_error = (
+        "Cache 'HashInfo(name='md5', "
+        "value='d089d68656330fd712dad63400c8710f', dir_info=None)' not found. "
+        "File 'foo' won't be created"
+    )
+    missing_error_hints = (
+        "foo (marked with 'store: false' in dvc file)",
```
jorgeorpinel, line 366:
```suggestion
        "foo (marked with 'store: false' in its .dvc file)",
```

dvc/exceptions.py
jorgeorpinel: Lovely

dvc/command/imp.py
```diff
@@ -65,4 +66,15 @@ def add_parser(subparsers, parent_parser):
...
     )
+    import_parser.add_argument(
+        "-s",
+        "--store",
+        action="store_true",
+        default=False,
+        help=(
+            "Store imported files in this repository's DVC remote, and "
+            "retrieve files from the remote on 'dvc pull' "
+            "instead of reimporting."
```
jorgeorpinel, line 77: How about

"Enable pushing the imported data to remote storage, and pulling it from there (instead of reimporting)."

? Thanks

jorgeorpinel (dismissed):
A few last wording suggestions ☝️  but LGTM!

dvc/exceptions.py
```diff
@@ -251,16 +251,32 @@ def __init__(self, amount):
...
             "cache up to date?\n{}".format(
                 "\n".join(targets), error_link("missing-files"),
             )
         )
+
+        if failed_not_stored:
+            m = m + (
+                "\n\nNote that 'store: false' outputs are not "
+                "searched for in remote storage, only in the cache. "
+                "See {}".format(error_link("checkout-failed-no-store"),)
```
jorgeorpinel, line 277:
```suggestion
                "See {}".format(error_link("checkout-no-store"),)
```
Merging this per https://github.com/iterative/dvc.org/pull/1788#pullrequestreview-501490410

dvc/exceptions.py
```diff
@@ -251,16 +251,32 @@ def __init__(self, amount):
...
+                targets.append(
+                    "{} (marked with 'store: false' in dvc file)".format(t)
+                )
+            else:
+                targets.append(str(t))
+
         m = (
-            "Checkout failed for following targets:\n{}\nIs your "
+            "Checkout failed for following targets:\n{}\n\nIs your "
             "cache up to date?\n{}".format(
```
jorgeorpinel, line 268: Copy edit:
```suggestion
            "Checkout failed for the following targets:\n{}\n\n
            "Is your cache up to date?\n{}".format(
```

dvc/exceptions.py
```diff
@@ -251,16 +251,32 @@ def __init__(self, amount):
...
                 "\n".join(targets), error_link("missing-files"),
             )
         )
+
+        if failed_not_stored:
+            m = m + (
+                "\n\nNote that 'store: false' outputs are not "
+                "searched for in remote storage, only in the cache. "
+                "See {}".format(error_link("checkout-no-store"),)
```
jorgeorpinel, line 277: Wait... Why is this a "checkout" problem though? The checkout would fail with or without `store: false` simply because the files are not in the cache, no?

Fetching or pulling is where `store: false` matters as I understand this. So shouldn't it be a fetch error/hint (info) message?

charlesbaynham: It's only the checkout that's a problem: `dvc fetch` happens fine because it doesn't attempt to retrieve the `store: false` files. It's only when you checkout (which is part of `dvc pull`) when it matters that they're not in the cache. 

So yes, it's the same as doing a checkout of a normal file that isn't in the cache. The additional error messages are just a hint to the user as to why that might have happened.

jorgeorpinel: > only the checkout that's a problem... The additional error messages are just a hint to the user as to why that might have happened.

Hmm.. If I just run checkout, I don't need to know anything about remote storage. These are basically unrelated things. The existing message about the files not being is the cache is enough there, I think. (Sorry, I know I already made you change it a lot 😅.)

> checkout (which is part of dvc pull

Good point. So pull itself is also not the issue.

I think this hint should happen independently at fetch, so that if you use fetch or pull and nothing is downloaded at least you get a message about a possible confusion. In the case of pull, if the files are also not in cache and you get that other (chekout) error, the combination of both messages should be what the user needs to know.

jorgeorpinel: But lets ask @efiop or/and @shcheklein to please confirm this before I make you go redo all the error handing again... Thanks!

charlesbaynham: I don't think the hint can happen at fetch, because there's no error at that point: not pulling these files from the remote is the expected, intended behaviour so showing an error message would be strange. 

The error only occurs because the files aren't available locally when you checkout, so I think it's definitely a checkout error, not a fetch error. So I don't think there's anywhere you could show the hint other than at checkout, since that's the only place that an error happens.

jorgeorpinel: It's not an error, it would be an INFO message (a hint).

> not pulling these files from the remote is the expected
> I don't think there's anywhere you could show the hint other than at checkout

Shouldn't fetch be always checking the `store` value now anyway? So there case to print the hint should be there somewhere already.

I still think this situation has nothing to do with checkout, which should be completely disconnected from remote storage. @shcheklein @efiop can you please corroborate? Thanks

charlesbaynham: This would be the place for a message on fetch:

https://github.com/iterative/dvc/blob/68471ebbf073b7c893ba3f9af17a09dc9425d761/dvc/output/base.py#L482-L490

Could be quite noisy though if a message here got fired for every import marked as `store: false`

efiop:
Sorry for the delay @charlesbaynham , we are not able to review this right now, but we'll try to get back to this in the next sprint. Thanks a lot for contributing!

jorgeorpinel:
Hi!

> Defaults to true for normal files and to false for imports

As recently mentioned in iterative/dvc.org/issues/1865, it seems like external outputs would conflict with this default value, as they cannot be pushed to remote storage. Have we tried this with the current changes? Thanks

dvc/repo/checkout.py
```diff
@@ -58,6 +58,7 @@ def checkout(
         "deleted": [],
         "modified": [],
         "failed": [],
+        "failed_not_stored": [],
```
pared, line 61: I don't think we should introduce new status. What matters from the point of view of `checkout` is that particular file failed. We should not spread logic related to `import` into `checkout`. Also, if we were to go with that status, we would need to revisit all situations, where remote does not contain particular hash (eg forgot to push) and in that case `failed_not_stored` would probably be more appropriate than just `failed`.

tests/func/test_data_cloud.py
```diff
@@ -308,6 +310,72 @@ def test_pull_git_imports(tmp_dir, dvc, scm, erepo):
...
+def test_store_false_pull_error(tmp_dir, dvc, scm, local_remote, caplog):
+    tmp_dir.dvc_gen("foo", "Locally added", commit="add no-store file")
+    tmp_dir.dvc_gen("bar", "Another local file", commit="add normal file")
+
+    dvc.push()
+    clean(["foo", "bar"], dvc)
+    dvc.pull()
+
+    with open(tmp_dir / "foo.dvc", "a") as F:
+        F.write("  store: false\n")
```
pared, line 355: I think it would be safer to `parse_yaml`, edit the dictionary and `dump_yaml`

dvc/stage/__init__.py
```diff
@@ -466,6 +465,8 @@ def _checkout(out, **kwargs):
                 return None, []
             return "modified" if modified else "added", [out.path_info]
         except CheckoutError as exc:
+            if not out.store:
+                return "failed_not_stored", exc.target_infos
```
pared, line 469: I think we should check here that stage is not imported, out.store == False, and then just `logger.debug` the message that originally was introduced to `CheckoutError`. That way we get rid of new status for checkout, and are able to pass suggestion why we think it fails.

pared (changes requested):
Great job @charlesbaynham! Sorry, it took so long to review it. I like the idea of introducing `store` param. Not sure we need "failed_not_stored" though.

tests/func/test_data_cloud.py
charlesbaynham: Makes sense, done

dvc/stage/__init__.py
charlesbaynham: We'd presumably want to set the logging level to warning, not debug, since otherwise users wouldn't see the hint unless they thought to use verbose mode. 

I agree that'll result in cleaner code, but it's a slightly coarser user experience. Instead of a nicely formatted list of files like

```
The following outputs are marked as `store: false` so remote storage was disabled, and are not present in the cache:
foo
<https://error.dvc.org/checkout-failed-no-store>
Checkout failed for following targets:
foo
Is your cache up to date?
```

You'll get the old error message interspersed with (potentially lots) of warnings about `store: false`.

dvc/repo/checkout.py
charlesbaynham: I wrote a bit below, but I'd point out that the logic here isn't logically linked to `import`: `store: false` is a property of an `out`, which just happens to be set by `import`. 

Also, `failed_not_stored` above wasn't mean to mean "these files failed because they aren't stored on the remote", it's meant to mean "these files failed and they have `store: false` set". I agree it's a little unsatisfying, but I don't think checkout is really being coupled to the import logic here: it's being linked to the settings of an `out`, but that's reasonable since that's what `checkout` is concerned with.

dvc/stage/__init__.py
charlesbaynham: Thanks @pared, and don't worry about the delay! 

I agree that the `failed_not_stored` addition to the statistics in `checkout` is a bit clunky, but it's the only way I could think of to provide a nice error message to the user for files with `store: false`. Seems like a shame to lose that: I don't think churning out logger messages is as nice a UX. Can you think of a way to inform `CheckoutError` objects created in `checkout` about which `outs` have `store: false` set on them without adding another category to `stats`?

charlesbaynham:
Thanks @pared, and don't worry about the delay! 

I agree that the `failed_not_stored` addition to the statistics in `checkout` is a bit clunky, but it's the only way I could think of to provide a nice error message to the user for files with `store: false`. Seems like a shame to lose that: I don't think churning out logger messages is as nice a UX. Can you think of a way to inform `CheckoutError` objects created in `checkout` about which `outs` have `store: false` set on them without adding another category to `stats`?

charlesbaynham:
On the other hand, marking files other than imports as `store: false` is quite a corner case, so maybe it's not unreasonable that the code is kept a bit cleaner at the expense of a slightly uglier, but just as informative error message.

dvc/stage/__init__.py
charlesbaynham: - [ ] PS wherever the warning ends up, I should add "if they're imports, try importing them" as a hint to the user

dvc/repo/checkout.py
```diff
@@ -100,8 +101,13 @@ def checkout(
             for key, items in result.items():
                 stats[key].extend(_fspath_dir(path) for path in items)
 
-    if stats.get("failed"):
-        raise CheckoutError(stats["failed"], stats)
+    if stats.get("failed") or stats.get("failed_not_stored"):
+        raise CheckoutError(
+            stats["failed"] + stats["failed_not_stored"],
+            stats,
+            failed_not_stored=stats["failed_not_stored"],
```
pared, line 108: Can we extract it inside CheckoutError from `stats`?

charlesbaynham: Yes we could do that, but we'd still have to keep the `failed_not_stored` entry on l.61 above. Unless we made it a `defaultdict` like in 
https://github.com/iterative/dvc/blob/68471ebbf073b7c893ba3f9af17a09dc9425d761/dvc/stage/__init__.py#L450-L457

That way `checkout` doesn't need to know about the structure of stats except for how to check if there are errors in 
https://github.com/iterative/dvc/blob/68471ebbf073b7c893ba3f9af17a09dc9425d761/dvc/repo/checkout.py#L104-L105

Alternatively, since stats is becoming complex, it could be made into an object with a `.failed` property which `checkout` could test. Then `CheckoutError` could extract the details without `checkout` needing to know about them.

dvc/repo/checkout.py
pared: You are right, I exaggerated here, the thing is about stage and outs rather than `import`.
Having said so, `failed_not_stored` is a piece of information connecting `stage` with `checkout`. 
It is the context for the reason why checkout for particular stage failed. The thing about `checkout` is that it is supposed to give summarized information about the state of the cache. So, till today we have been having 4 basic "stats" - which aggregate all possible outcomes of checkout.

Introducing hinting as an integral part of checkout is precedence that might result in more and more stats fields: in most cases, we can pinpoint why the checkout failed - and suggest user what is the problem. In long term, this approach is unmaintainable, since there will be a lot of use cases and reasons to fail.  

For example: what if in the current state of change, the output will fail to checkout for a different reason, for example, access constraints? Wouldn't it mean that the error would still point to out being not stored if `out.store` is true?

charlesbaynham: > For example: what if in the current state of change, the output will fail to checkout for a different reason, for example, access constraints? Wouldn't it mean that the error would still point to out being not stored if out.store is true?

Hmm yes, you're right about that. It's a hint rather than an error, but still. 

I also completely get your point about maintainability if we try to hint for every possible failure mode. 

That said, I do think this failure mode is a bit special, since it results from the use of a feature entirely within dvc (i.e. "store: false"). Compared to e.g. an access constraint or a network failure which are external failures, this one is due to settings which we have complete knowledge about. 

I'm not strongly in favor of keeping the hint, maybe a logger warning (or info?) is the easiest way around the problem.

pared: > It's a hint rather than an error, but still.

This is a problem about this approach: we would introduce new `checkout` status, in order to be able to hint user, knowing its just a hunch, while the actual reason for failure might be different. I totally get that this error comes "from design" and its detection is a bit different than other errors. I just think remodeling whole `checkout` in order to give user a hint what is wrong is a bit too much. 

Don't get me wrong, I am all for hinting, and we actually have been doing it quite a lot at a time, but introducing multiple hints over time results in a lot of `WARNING` log entries even in case of simple use cases, which gets users confused. 

That being said, in case of more advanced users, it would be desirable to be able to point them to potential solution. And that is why I suggested `logger.debug` on checkout - its a hint, does not require major changes in code and can be implemented in single place, the actual origin of `CheckoutError`. I guess it makes sense to make it into `warning`, as this after checkout, user won't be able to reproduce the results (since there will be no data to check out).

tests/func/test_data_cloud.py
```diff
@@ -308,6 +310,73 @@ def test_pull_git_imports(tmp_dir, dvc, scm, erepo):
...
+    assert dvc.pull(force=True)["fetched"] == 2
+
+    assert (tmp_dir / "foo").exists()
+    assert (tmp_dir / "foo").read_text() == "foo"
+
+    assert (tmp_dir / "new_dir").exists()
+    assert (tmp_dir / "new_dir" / "bar").read_text() == "bar"
+
+
+def test_store_false_pull_error(tmp_dir, dvc, scm, local_remote, caplog):
```
pared, line 346: Hmm, I wonder whether we need any communication with user (the "out" is not stored) at all. Do we actually have a situation, where this information would pop up? 

If we `import` without `--store` option, DVC (upon pull) will try to re-import the file. So there will actually be no situation when we try to checkout `out` from cache, that, at the same time is not stored.

charlesbaynham: Sure, unless you've e.g. run `git clean -xdf` and wiped your cache, or cloned someone else's repository. I hope the situation wouldn't happen often, but it is possible so we ought to have a test with some kind of well-defined behavior for that scenario, whatever that behavior is.

dvc/repo/checkout.py
charlesbaynham: Alright, I'm convinced :) I'll go ahead and revert the hinting stuff.

dvc/exceptions.py
jorgeorpinel: @pared what do you think about this discussion? Since I see you reviewed this from the eng. team. Thanks

dvc/exceptions.py
```diff
@@ -251,16 +251,32 @@ def __init__(self, amount):
...
-    def __init__(self, target_infos, stats=None):
+    def __init__(self, target_infos, stats=None, failed_not_stored=None):
         self.target_infos = target_infos
         self.stats = stats
-        targets = [str(t) for t in target_infos]
+        targets = []
+        for t in target_infos:
+            if failed_not_stored and t in failed_not_stored:
+                targets.append(
+                    "{} (marked with 'store: false' in its dvc file)".format(t)
```
jorgeorpinel, line 261:
```suggestion
                    "{} (marked with `store: false` in its dvc file)".format(t)
```

dvc/output/base.py
```diff
@@ -53,6 +53,15 @@ def __init__(self, match):
         super().__init__(f"Path '{match.file}' is ignored by\n{lines}")
 
 
+class OutputStoreCacheIncompatible(DvcException):
+    def __init__(self, path):
+        msg = f"""Incompatible options for output '{path}'.
+    The dvc file specifies an output with both "cache == False" and
```
jorgeorpinel, line 59:
```suggestion
    The .dvc file specifies an output with both "cache == False" and
```

jorgeorpinel: Also maybe use `cache: false` and `store: true` (in back quotes) since that's how they're shown in .dvc files.

dvc/exceptions.py
pared: @jorgeorpinel 
As far as I understand from our most recent conversation (https://github.com/iterative/dvc/pull/4581#discussion_r509366337)

We are going to change the approach and hint user (via warning message) during `stage` checkout. Making this error message obsolete. @charlesbaynham am I right here?

charlesbaynham: That's my understanding too. I just haven't done it yet!

pared: @charlesbaynham no worries, I am not trying to rush you, just clearing things up, so that we all are on the same page :)

jorgeorpinel: OK then, resolving here. Thanks

dvc/output/base.py
```diff
@@ -53,6 +53,15 @@ def __init__(self, match):
         super().__init__(f"Path '{match.file}' is ignored by\n{lines}")
 
 
+class OutputStoreCacheIncompatible(DvcException):
+    def __init__(self, path):
+        msg = f"""Incompatible options for output '{path}'.
+    The .dvc file specifies an output with both "cache == False" and
+    "store == True" which is not possible to satisfy.
```
efiop, line 60:
```suggestion
    The .dvc file specifies an output with both `cache == False` and
    `store == True` which is not possible to satisfy.
```

dvc/output/base.py
```diff
@@ -311,6 +325,9 @@ def dumpd(self):
         if self.persist:
             ret[self.PARAM_PERSIST] = self.persist
 
+        if self.store == self.stage.is_repo_import:
```
efiop, line 328: probably
```suggestion
        if self.stage.is_repo_import and self.store:
```
?

dvc/output/base.py
```diff
@@ -86,6 +96,7 @@ class BaseOutput:
     IsNotFileOrDirError = OutputIsNotFileOrDirError
     IsStageFileError = OutputIsStageFileError
     IsIgnoredError = OutputIsIgnoredError
+    StoreCacheIncompatible = OutputStoreCacheIncompatible
```
efiop, line 99:
```suggestion
```

dvc/output/base.py
```diff
@@ -458,10 +474,16 @@ def get_used_cache(self, **kwargs):
         """
 
         if not self.use_cache:
+            if self.store:
+                raise self.StoreCacheIncompatible(self)
```
efiop, line 478:
```suggestion
                raise OutputStoreCacheIncompatible(self)
```

efiop:
@charlesbaynham Thanks a lot for your patience. Really sorry for this is taking so long.

This PR looks great, just a few minor details to adjust. Let me see if we can take it from here...

efiop:
@charlesbaynham One more thing, could you please check the committer email in your commits and add it to https://github.com/settings/emails so that github is able to associate the great work you've done with your github account?

dvc/output/base.py
charlesbaynham: No, actually that is right (although unclear): the purpose of this is to only add `store: <something>` to the DVC file if it differs from the default. Default is `store: true` for most outputs and `store: false` for repo imports. So this code writes the value of `store` to the DVC file only if 

* it's a repo import and `store: true`
* or it's a normal import and `store: false`

charlesbaynham:
> @charlesbaynham Thanks a lot for your patience. Really sorry for this is taking so long.
> 
> This PR looks great, just a few minor details to adjust. Let me see if we can take it from here...

Not at all, and sorry I haven't got around to this: it's on my list! Although I'm very happy for you to take it from here if you want to, of course.

dvc/dependency/__init__.py
```diff
@@ -48,11 +48,13 @@
...
 # without output-specific entries like 'cache' (whether or not output is
-# cached, see -o and -O flags for `dvc run`) and 'metric' (whether or not
-# output is a metrics file and how to parse it, see `-M` flag for `dvc run`).
+# cached, see -o and -O flags for `dvc run`), 'metric' (whether or not output
+# is a metric file and how to parse it, see `-M` flag for `dvc run`) and
+# `store` (whether to push outputs to the remote)
 SCHEMA = output.SCHEMA.copy()
 del SCHEMA[BaseOutput.PARAM_CACHE]
 del SCHEMA[BaseOutput.PARAM_METRIC]
+del SCHEMA[BaseOutput.PARAM_STORE]
```
karajan1001, line 57: These lines are not coupling outputs with dependencies. Maybe some refactoring is needed. But this is not related to this PR.

efiop:
@charlesbaynham Thanks again for the PR and all the effort you've put into it! :pray: Even though the functionality is pretty simple and nicely implemented by you, the main reason why this hasn't been merged yet is that `store` option really highlighted the current mess of terminology and defaults that we have going on between dvcfiles created by `dvc add`, `dvc import`, `dvc import-url` and `Stage` class. This is apparent in other areas as well, but this might be the final nail :slightly_smiling_face:

We are currently starting to work on 2.0 changes and there are some plans to at least reconsider dvc file format as well as internal Stage class split, which might help set this functionality straight as well. We'll be having some discussions about it this week, so still keeping this PR on hold for now. :slightly_frowning_face:

charlesbaynham:
Hey @efiop, thanks for letting me know! Do say if there's anything I can do to help

efiop:
It has been a while, but we now have more proper mechanisms to handle multiple remotes, as well as `remote` option for outputs. E.g. imports effectively mean smth like
```
remote:
   pull: source_remote
   push: false
```
and "backups" could mean just setting `push` to the remote you want to backup to. We will be soon ready to come back to this. Thank you so much for your patience! :pray:

skshetry:
Closing as stale. Thanks @charlesbaynham.

PR #4582, opened 2020-09-21
Do not require account-level permissions for Azure tree

mhworth:

Background: We only issue container-level, user-delegated SAS tokens to our developers for the Azure blob store. That is a hard requirement due to some HIPAA security needs. To be very specific, I generate the SAS like so:

```
az storage container generate-sas --account-name my-account --name my-container --permissions acdlrw --expiry 2020-09-23 --auth-mode login --as-user
```

As long as I construct the connection string correctly, everything works fine except for this check:

https://github.com/iterative/dvc/blob/15488ebefd5d1bcaad80526472b04b46f89a71cb/dvc/tree/azure.py#L91-L94

This pull request gives a pass if that check fails for authentication error purposes. I believe this should be backwards compatible, since:

a) If they were using container-level SAS tokens this wouldn't have worked for them in the first place
b) If the get_container_properties() function fails with the ResourceNotFoundError, it will revert to the old behavior

I think requiring people to make their own containers is an acceptable solution too, but the goal here was to maintain backwards compability.

With this fix, I'm able to use the Container-level SAS's described above.

### DIFF
--- a/dvc/tree/azure.py
+++ b/dvc/tree/azure.py
@@ -95,7 +95,7 @@ class AzureTree(BaseTree):
             container_client.create_container()
         except HttpResponseError as e:
             # client may not have account-level privileges
-            if e.status_code != 403:
+            if exc.status_code != 403:
                 raise
 
         return blob_service
--- a/dvc/tree/azure.py
+++ b/dvc/tree/azure.py
@@ -93,8 +93,11 @@ class AzureTree(BaseTree):
             container_client.get_container_properties()
         except ResourceNotFoundError:
             container_client.create_container()
-        except HttpResponseError:
-            pass  # Client may not have account-level privileges
+        except HttpResponseError as e:
+            if e.status_code == 403:
+                pass  # client may not have account-level privileges
+            else:
+                raise
 
         return blob_service
 
--- a/dvc/tree/azure.py
+++ b/dvc/tree/azure.py
@@ -94,9 +94,8 @@ class AzureTree(BaseTree):
         except ResourceNotFoundError:
             container_client.create_container()
         except HttpResponseError as e:
-            if e.status_code == 403:
-                pass  # client may not have account-level privileges
-            else:
+            # client may not have account-level privileges
+            if e.status_code != 403:
                 raise
 
         return blob_service
--- a/dvc/tree/azure.py
+++ b/dvc/tree/azure.py
@@ -69,6 +69,7 @@ class AzureTree(BaseTree):
         # pylint: disable=no-name-in-module
         from azure.core.exceptions import ResourceNotFoundError
         from azure.storage.blob import BlobServiceClient
+        from azure.core.exceptions import HttpResponseError
 
         logger.debug(f"URL {self.path_info}")
 
@@ -92,6 +93,8 @@ class AzureTree(BaseTree):
             container_client.get_container_properties()
         except ResourceNotFoundError:
             container_client.create_container()
+        except HttpResponseError:
+            pass  # Client may not have account-level privileges
 
         return blob_service
 


dvc/tree/azure.py
```diff
@@ -92,6 +93,8 @@ def blob_service(self):
             container_client.get_container_properties()
         except ResourceNotFoundError:
             container_client.create_container()
+        except HttpResponseError:
```
efiop, line 96: Maybe we could catch a more specific exception here? Or are you getting a generic HttpResponseError? If so, what code does have?

mhworth: It's a generic one unfortunately, status code is 403:

```
(Pdb) e
HttpResponseError('This request is not authorized to perform this operation.\nRequestId:xyz\nTime:2020-09-21T10:16:42.1106259Z\nErrorCode:AuthorizationFailure\nError:None')
(Pdb) dir(e)
['__cause__', '__class__', '__context__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__setstate__', '__sizeof__', '__str__', '__subclasshook__', '__suppress_context__', '__traceback__', '__weakref__', '_parse_odata_body', 'additional_info', 'args', 'error', 'error_code', 'exc_msg', 'exc_traceback', 'exc_type', 'exc_value', 'inner_exception', 'message', 'model', 'raise_with_traceback', 'reason', 'response', 'status_code', 'with_traceback']
(Pdb) e.status_code
403
(Pdb) e.reason
'This request is not authorized to perform this operation.'
```

efiop: @mhworth That's actually pretty good, `403` is `Forbidden`, better than nothing. Let's check that this is indeed 403. Otherwise looks good.

mhworth: Sounds good! Just pushed up an edit that checks to see if status_code == 403 and then reraises if not.

dvc/tree/azure.py
```diff
@@ -92,6 +93,11 @@ def blob_service(self):
             container_client.get_container_properties()
         except ResourceNotFoundError:
             container_client.create_container()
+        except HttpResponseError as e:
+            if e.status_code == 403:
+                pass  # client may not have account-level privileges
+            else:
+                raise
```
efiop, line 100:
```suggestion
            # client may not have account-level privileges
            if e.status_code != 403:
                raise
```

dvc/tree/azure.py
```diff
@@ -92,6 +93,10 @@ def blob_service(self):
             container_client.get_container_properties()
         except ResourceNotFoundError:
             container_client.create_container()
+        except HttpResponseError as e:
```
efiop, line 96:
```suggestion
        except HttpResponseError as exc:
```

dvc/tree/azure.py
```diff
@@ -92,6 +93,10 @@ def blob_service(self):
...
+            # client may not have account-level privileges
+            if e.status_code != 403:
```
efiop, line 98:
```suggestion
            if exc.status_code != 403:
```

efiop: Just for consistency and to not annoy the linter :)

mhworth: No problem :)

dvc/tree/azure.py
efiop: Looks like this one was forgotten :) Can't add it myself, as your organization forbids edits from maintainers in your PRs.

efiop:
Merged in https://github.com/iterative/dvc/pull/4583 

Thank you! :pray:

mhworth:
@efiop 🎉  No problem! Thanks for reviewing.

PR #4590, opened 2020-09-22
metrics: accept any viable target in DvcRepo

pared:

### DIFF
--- a/dvc/repo/metrics/show.py
+++ b/dvc/repo/metrics/show.py
@@ -11,28 +11,31 @@ logger = logging.getLogger(__name__)
 
 
 def _collect_metrics(repo, targets, recursive):
-    metrics = set()
 
+    if targets:
+        target_infos = [
+            PathInfo(os.path.abspath(target)) for target in targets
+        ]
+        tree = RepoTree(repo)
+
+        rec_files = []
+        if recursive:
+            for target_info in target_infos:
+                if tree.isdir(target_info):
+                    rec_files.extend(list(tree.walk_files(target_info)))
+
+        result = [t for t in target_infos if tree.isfile(t)]
+        result.extend(rec_files)
+
+        return result
+
+    metrics = set()
     for stage in repo.stages:
         for out in stage.outs:
             if not out.metric:
                 continue
-
             metrics.add(out.path_info)
-
-    if not targets:
-        return list(metrics)
-
-    target_infos = [PathInfo(os.path.abspath(target)) for target in targets]
-
-    def _filter(path_info):
-        for info in target_infos:
-            func = path_info.isin_or_eq if recursive else path_info.__eq__
-            if func(info):
-                return True
-        return False
-
-    return list(filter(_filter, metrics))
+    return list(metrics)
 
 
 def _extract_metrics(metrics, path, rev):
--- a/tests/func/metrics/test_show.py
+++ b/tests/func/metrics/test_show.py
@@ -110,3 +110,41 @@ def test_missing_cache(tmp_dir, dvc, run_copy_metrics):
     remove(stage.outs[0].cache_path)
 
     assert dvc.metrics.show() == {"": {"metrics.yaml": 1.1}}
+
+
+def test_show_non_metric(tmp_dir, dvc):
+    tmp_dir.gen("metrics.yaml", "foo: 1.1")
+
+    assert dvc.metrics.show(targets=["metrics.yaml"]) == {
+        "": {"metrics.yaml": {"foo": 1.1}}
+    }
+
+
+def test_show_non_metric_branch(tmp_dir, scm, dvc):
+    tmp_dir.scm_gen("metrics.yaml", "foo: 1.1", commit="init")
+    with tmp_dir.branch("branch", new=True):
+        tmp_dir.scm_gen("metrics.yaml", "foo: 2.2", commit="other")
+
+    assert dvc.metrics.show(targets=["metrics.yaml"], revs=["branch"]) == {
+        "workspace": {"metrics.yaml": {"foo": 1.1}},
+        "branch": {"metrics.yaml": {"foo": 2.2}},
+    }
+
+
+def test_non_metric_and_recurisve_show(tmp_dir, dvc, run_copy_metrics):
+    tmp_dir.gen(
+        {"metrics_t.yaml": "foo: 1.1", "metrics": {"metric1.yaml": "bar: 1.2"}}
+    )
+
+    metric2 = os.fspath(tmp_dir / "metrics" / "metric2.yaml")
+    run_copy_metrics("metrics_t.yaml", metric2, metrics=[metric2])
+
+    assert dvc.metrics.show(
+        targets=["metrics_t.yaml", "metrics"], recursive=True
+    ) == {
+        "": {
+            os.path.join("metrics", "metric1.yaml"): {"bar": 1.2},
+            os.path.join("metrics", "metric2.yaml"): {"foo": 1.1},
+            "metrics_t.yaml": {"foo": 1.1},
+        }
+    }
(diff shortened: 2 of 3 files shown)


pared:
@jorgeorpinel Do you think its worth mentioning in docs that from this change on `dvc metrics` will work with any file if its provided via `targets`?

dvc/repo/metrics/show.py
```diff
@@ -22,17 +22,29 @@ def _collect_metrics(repo, targets, recursive):
...
-
-    def _filter(path_info):
-        for info in target_infos:
-            func = path_info.isin_or_eq if recursive else path_info.__eq__
-            if func(info):
-                return True
-        return False
-
-    return list(filter(_filter, metrics))
+    else:
```
efiop, line 25: This `else` is not needed, as you `return` above anyway

dvc/repo/metrics/show.py
```diff
@@ -22,17 +22,29 @@ def _collect_metrics(repo, targets, recursive):
...
+                    )
+                    if func(info):
+                        return True
+                return False
+
+            result.extend(list(filter(_filter, metrics)))
+
+        tree = RepoTree(repo)
+        for target in targets:
+            if tree.exists(target) and not tree.isdir(target):
```
efiop, line 45:
```suggestion
            if tree.isfile(target):
```
?

efiop: Not strictly equivalent, but wonder if that's what you meant here.

tests/func/metrics/test_show.py
```diff
@@ -110,3 +110,31 @@ def test_missing_cache(tmp_dir, dvc, run_copy_metrics):
     remove(stage.outs[0].cache_path)
 
     assert dvc.metrics.show() == {"": {"metrics.yaml": 1.1}}
+
```
efiop, line 113: Let's also add a test for showing something that is not metric in other branches.

tests/func/metrics/test_show.py
```diff
@@ -110,3 +110,31 @@ def test_missing_cache(tmp_dir, dvc, run_copy_metrics):
...
+    run_copy_metrics("metrics_t.yaml", metric1, metrics=[metric1])
+    run_copy_metrics("metrics_t.yaml", metric2, metrics=[metric2])
+
+    result = dvc.metrics.show(
+        targets=["metrics_t.yaml", "metrics"], recursive=True
+    )
+    assert result == {
+        "": {
+            "metrics/metric1.yaml": {"foo": 1.1},
+            "metrics/metric2.yaml": {"foo": 1.1},
```
efiop, line 137: Will it be / for windows too though?

tests/func/metrics/test_show.py
```diff
@@ -110,3 +110,31 @@ def test_missing_cache(tmp_dir, dvc, run_copy_metrics):
...
+    }
+
+
+def test_show_non_metric_and_recursive_metric(tmp_dir, dvc, run_copy_metrics):
+    tmp_dir.gen({"metrics_t.yaml": "foo: 1.1", "metrics": {}})
+
+    metric1 = os.fspath(tmp_dir / "metrics" / "metric1.yaml")
+    metric2 = os.fspath(tmp_dir / "metrics" / "metric2.yaml")
+    run_copy_metrics("metrics_t.yaml", metric1, metrics=[metric1])
+    run_copy_metrics("metrics_t.yaml", metric2, metrics=[metric2])
```
efiop, line 129: These are not really related to this PR, right? Or did you want to test if recursive works for dirs that don't have official metrics? Just a bit confused.

pared: I wanted to verify whether `show` will work as intended if we mix `non-metric` and `metric` approach in case of `recursive=true`

dvc/repo/metrics/show.py
pared: Well I wanted to omit directories in case of `recursive`, so applying some logic `isfile` seems like a better check.

tests/func/metrics/test_show.py
efiop: Makes sense.

dvc/repo/metrics/show.py
```diff
@@ -23,16 +23,26 @@ def _collect_metrics(repo, targets, recursive):
...
 
-    def _filter(path_info):
-        for info in target_infos:
-            func = path_info.isin_or_eq if recursive else path_info.__eq__
-            if func(info):
-                return True
-        return False
+        def _filter(path_info):
+            for info in target_infos:
+                func = path_info.isin_or_eq if recursive else path_info.__eq__
```
efiop, line 34: Does this make any sense now? You are under `if recursive` already

dvc/repo/metrics/show.py
```diff
@@ -23,16 +23,25 @@ def _collect_metrics(repo, targets, recursive):
...
+                    return True
+            return False
 
-    return list(filter(_filter, metrics))
+        result.extend(list(filter(_filter, metrics)))
+
+    tree = RepoTree(repo)
+    for target in targets:
+        if tree.isfile(target):
+            result.append(PathInfo(target))
```
efiop, line 43: Seems like you can simply reuse `target_infos` from above.

dvc/repo/metrics/show.py
```diff
@@ -23,16 +23,23 @@ def _collect_metrics(repo, targets, recursive):
     if not targets:
         return list(metrics)
 
+    result = []
     target_infos = [PathInfo(os.path.abspath(target)) for target in targets]
+    if recursive:
```
efiop, line 28: Will recursive work with unofficial metrics? I don't see a test for it and the code seems like it won't work.

pared: I am not sure how that would work, I mean, in that case we would need to try to show all files for particular directory, since they are not marked as metrics. So the question here is whether we want to support that.

efiop: @pared That seems like a natural approach. You have a test with official metrics dir in this PR, what changes if they are not declared as official metrics?

dvc/repo/metrics/show.py
```diff
@@ -23,16 +23,23 @@ def _collect_metrics(repo, targets, recursive):
...
+                if path_info.isin_or_eq(info):
+                    return True
+            return False
 
-    return list(filter(_filter, metrics))
+        result.extend(list(filter(_filter, metrics)))
+
+    tree = RepoTree(repo)
+    for target_info in target_infos:
+        if tree.isfile(target_info):
```
efiop, line 40: What if a target is a file and `recursive==True`? Will it be added to the `result` twice?

dvc/repo/metrics/show.py
pared: On the other hand, it kinda makes sense to have special dir for metrics and check all of them out. Ill try to modify the change.

efiop: Also note that we try to parse metric files, so if something is not a metric we'll just ignore it and won't show it.

dvc/repo/metrics/show.py
efiop: This will probably become obsoleted once you implement recursive for unofficial metrics above...

tests/func/metrics/test_show.py
```diff
@@ -110,3 +110,42 @@ def test_missing_cache(tmp_dir, dvc, run_copy_metrics):
...
+    tmp_dir.scm_gen("metrics.yaml", "foo: 1.1", commit="init")
+    with tmp_dir.branch("branch", new=True):
+        tmp_dir.scm_gen("metrics.yaml", "foo: 2.2", commit="other")
+
+    assert dvc.metrics.show(targets=["metrics.yaml"], revs=["branch"]) == {
+        "workspace": {"metrics.yaml": {"foo": 1.1}},
+        "branch": {"metrics.yaml": {"foo": 2.2}},
+    }
+
+
```
efiop, line 133: Missing a test for recursive directory with unofficial metrics.

pared: Since both official and unofficial metrics are gathered the same way now, I would argue that `test_non_metric_and_recurisve_show` already covers that. Anyhow, I modified it so that it checks both official and unofficial metrics.

efiop: Thank you! :pray:

efiop (approved):
Thanks!

jorgeorpinel:
> Do you think its worth mentioning in docs

Yes! Very much 🙂 — thanks for submitting a docs PR BTW.

PR #4591, opened 2020-09-22
experiments: checkpoints proof of concept

### LINKED ISSUE #4498: Experiments: checkpoints tracking

pmrowla:

Will close #4498.

* `dvc exp run --checkpoint` can be used to reproduce a checkpoint experiment
    * Checkpoint stages should be generated using `--always-changed` and with `--outs-persist` for the intermediate checkpoint outputs
    * When running a checkpoint experiment, a new commit in the experiment branch will be generated each time the running stage calls `dvc.api.make_checkpoint()` (or generates the appropriate `.dvc/tmp/DVC_CHECKPOINT`) signal file
    * Reproduction and checkpoint generation will run forever until the user code exits or is killed via Ctrl+C
* `dvc exp run --continue <checkpoint_exp_rev>` can be used to resume a prior checkpoint experiment. Execution will be resumed from the tip of the checkpoint branch.
* Checkpoint experiments can be viewed as normal in `dvc exp show`

Known issues (needs further investigation after this PR):
* Git-python sometimes throws broken pipe errors when checkpoint runs are killed via Ctrl-C
* Reproducing a baseline commit does not work properly, for a checkpoint experiment to be generated properly, there needs to be some change versus the baseline/parent commit (either in the workspace or via the `repro --params` option)
* Sorting in `dvc exp show` will not work properly if the table contains checkpoint experiments, but filtering should work as expected

Features that will need follow up PR:
* Branching from some commit in the middle of a checkpoint experiment and then resuming a "new" branch with potentially modified code/params is not yet supported

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -1,13 +1,17 @@
 import argparse
 import io
 import logging
+import os
 from collections import OrderedDict
 from datetime import date
 from itertools import groupby
 from typing import Iterable, Optional
 
+from dvc.command import completion
 from dvc.command.base import CmdBase, append_doc_link, fix_subparsers
-from dvc.command.metrics import DEFAULT_PRECISION
+from dvc.command.metrics import DEFAULT_PRECISION, _show_metrics
+from dvc.command.status import CmdDataStatus
+from dvc.dvcfile import PIPELINE_FILE
 from dvc.exceptions import DvcException, InvalidArgumentError
 from dvc.utils.flatten import flatten
 
@@ -373,6 +377,63 @@ class CmdExperimentsDiff(CmdBase):
         return 0
 
 
+class CmdExperimentsRun(CmdBase):
+    def run(self):
+        if not self.repo.experiments:
+            return 0
+
+        saved_dir = os.path.realpath(os.curdir)
+        os.chdir(self.args.cwd)
+
+        # Dirty hack so the for loop below can at least enter once
+        if self.args.all_pipelines:
+            self.args.targets = [None]
+        elif not self.args.targets:
+            self.args.targets = self.default_targets
+
+        ret = 0
+        for target in self.args.targets:
+            try:
+                stages = self.repo.reproduce(
+                    target,
+                    single_item=self.args.single_item,
+                    force=self.args.force,
+                    dry=self.args.dry,
+                    interactive=self.args.interactive,
+                    pipeline=self.args.pipeline,
+                    all_pipelines=self.args.all_pipelines,
+                    run_cache=not self.args.no_run_cache,
+                    no_commit=self.args.no_commit,
+                    downstream=self.args.downstream,
+                    recursive=self.args.recursive,
+                    force_downstream=self.args.force_downstream,
+                    experiment=True,
+                    queue=self.args.queue,
+                    run_all=self.args.run_all,
+                    jobs=self.args.jobs,
+                    params=self.args.params,
+                    checkpoint=(
+                        self.args.checkpoint or self.args.checkpoint_continue
+                    ),
+                    checkpoint_continue=self.args.checkpoint_continue,
+                )
+
+                if len(stages) == 0:
+                    logger.info(CmdDataStatus.UP_TO_DATE_MSG)
+
+                if self.args.metrics:
+                    metrics = self.repo.metrics.show()
+                    logger.info(_show_metrics(metrics))
+
+            except DvcException:
+                logger.exception("")
+                ret = 1
+                break
+
+        os.chdir(saved_dir)
+        return ret
+
+
 def add_parser(subparsers, parent_parser):
     EXPERIMENTS_HELP = "Commands to display and compare experiments."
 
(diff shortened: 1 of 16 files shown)


dvc/command/experiments.py
```diff
@@ -552,3 +613,156 @@ def add_parser(subparsers, parent_parser):
         metavar="<n>",
     )
     experiments_diff_parser.set_defaults(func=CmdExperimentsDiff)
+
+    EXPERIMENTS_RUN_HELP = (
+        "Reproduce complete or partial experiment pipelines."
+    )
+    experiments_run_parser = experiments_subparsers.add_parser(
```
pmrowla, line 620: all experiments run/repro behavior will be moved from `dvc repro` to `dvc exp run` in a follow-up PR, for now `repro -e` is just duplicated here so that checkpoint runs work correctly

pmrowla:
[asciicast](https://asciinema.org/a/OglsZBFwoIALfwsd98DdoipTn)

efiop:
@pmrowla What's up with the broken pipe in the recording?

efiop (approved):
Great start!

Issue #4574, opened 2020-09-15
dvc version: show external cache and remotes being used


skshetry:
This would have made my life easier when investigating https://github.com/iterative/dvc/pull/4570.

Another question to ask is that `dvc version` can only have a limited amount of information. Should there be `dvc version --json`?

PR #4598, opened 2020-09-22
Show current external cache and remotes being used if they exist

codingsett:

Resolves: #4574

### DIFF
--- a/dvc/info.py
+++ b/dvc/info.py
@@ -4,9 +4,9 @@ import pathlib
 import platform
 import uuid
 
+from dvc.config import SCHEMA
 from dvc.exceptions import DvcException, NotDvcRepoError
 from dvc.repo import Repo
-from dvc.scheme import Schemes
 from dvc.scm.base import SCMError
 from dvc.system import System
 from dvc.tree import TREES
@@ -36,7 +36,6 @@ def get_dvc_info():
 
     try:
         repo = Repo()
-
         # cache_dir might not exist yet (e.g. after `dvc init`), and we
         # can't auto-create it, as it might cause issues if the user
         # later decides to enable shared cache mode with
@@ -48,6 +47,13 @@ def get_dvc_info():
             if psutil:
                 fs_type = get_fs_type(repo.cache.local.cache_dir)
                 info.append(f"Cache directory: {fs_type}")
+                external_cache = repo.cache.local.tree.config.get("url", None)
+                remote_cache = _get_external_cache(repo)
+                if external_cache:
+                    info.append(f"External file cache: {external_cache}")
+                if remote_cache:
+                    info.append(f"External remote cache: {remote_cache}")
+
         else:
             info.append("Cache types: " + error_link("no-dvc-cache"))
 
@@ -62,9 +68,7 @@ def get_dvc_info():
             info.append(f"Workspace directory: {fs_root}")
         info.append("Repo: {}".format(_get_dvc_repo_info(repo)))
 
-        cache_location, remote_url = _get_external_remotes_cache(repo)
-        if cache_location:
-            info.append(f"External cache: {cache_location}")
+        remote_url = _get_external_remotes(repo)
         if remote_url:
             info.append(f"Remote url: {remote_url}")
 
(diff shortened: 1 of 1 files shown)


dvc/info.py
```diff
@@ -134,3 +142,26 @@ def _get_dvc_repo_info(self):
...
     return "dvc, git"
+
+
+def _get_external_remotes_cache(self):
+    cache_info = self.config.get("cache", {})
+
+    cache_vars = [
+        val for val in Schemes.__dict__.values() if isinstance(val, str)
+    ]
+    cache_key = list(set(cache_info.keys()).intersection(set(cache_vars)))
```
efiop, line 153: This works, but the list of accepted external caches is more limited: https://github.com/iterative/dvc/blob/38dd1c9fe7d0418868cd3a2d0962df3dd49ead71/dvc/config.py#L146 .

Also, it is not easy to spot, but `dir` https://github.com/iterative/dvc/blob/38dd1c9fe7d0418868cd3a2d0962df3dd49ead71/dvc/config.py#L154 is also a special case of `local`.

dvc/info.py
```diff
@@ -134,3 +142,26 @@ def _get_dvc_repo_info(self):
...
+    cache_location, remote_url = "", ""
+
+    if cache_info and cache_key:
+        cache_location = cache_info[cache_key[0]]
+
+    if core_info.get("remote", False):
+        remote = core_info["remote"]
+        remote_url = (
+            self.config.get("remote", {}).get(remote, {}).get("url", None)
+        )
```
efiop, line 165: Seems like `cache` and `remote` parts are not really related in this helper, so could be separated into separate helpers.

dvc/info.py
```diff
@@ -134,3 +142,26 @@ def _get_dvc_repo_info(self):
...
+    cache_vars = [
+        val for val in Schemes.__dict__.values() if isinstance(val, str)
+    ]
+    cache_key = list(set(cache_info.keys()).intersection(set(cache_vars)))
+
+    core_info = self.config.get("core", {})
+    cache_location, remote_url = "", ""
+
+    if cache_info and cache_key:
+        cache_location = cache_info[cache_key[0]]
```
efiop, line 159: A repo might have multiple caches (e.g. one for s3, gs, etc). The original ticket mostly focuses on local external cache dir (yeah, I bit confusing, I know :slightly_smiling_face: ), so `local` or `dir` keys in `cache` config section.

dvc/info.py
codingsett: But isn't _local_ considered as external since it's not in the parent directory?

dvc/info.py
codingsett: Yes I agree but won't _cache_info_ show the current cache in use because I assume you can't use multiple caches at the same time or am I wrong on this?

dvc/info.py
codingsett: I agree with you here, I will make changes and push.

dvc/info.py
efiop: Yeah, pretty much whenever cache dir is outside of repo it is considered external. Just saying that configuring `dir` here is kinda like creating a separate section and setting `cache.local`. Again, I know that this part is a bit confusing :slightly_smiling_face:

dvc/info.py
efiop: You can, see https://dvc.org/doc/user-guide/managing-external-data . It is a rather confusing part of dvc subsystems :slightly_frowning_face: In this case, I would say that @skshetry meant only `local` and `dir`, as those are the most useful local cache dir config options that many people use. The rest are used by a few people in a very advanced scenarios, and we could totally just ignore those :slightly_smiling_face:

efiop:
For the record: please don't mind windows tests, we are in the middle of migrating from travis to github actions, so have a few flaky ones :slightly_smiling_face: Sorry for inconvenience.

dvc/info.py
skshetry: @efiop, I meant all kinds of cache types in the dvc version or at least in the log during `--debug`, see the discord discussion for the motivation: https://discord.com/channels/485586884165107732/485596304961962003/755067231339020358.

codingsett: @sanketsaurav all kinds of caches should now show up in the output of `dvc version` command.

dvc/info.py
```diff
@@ -47,6 +47,13 @@ def get_dvc_info():
             if psutil:
                 fs_type = get_fs_type(repo.cache.local.cache_dir)
                 info.append(f"Cache directory: {fs_type}")
+                external_cache = repo.cache.local.tree.config.get("url", None)
```
efiop, line 50:
```suggestion
                external_cache = repo.cache.local.cache_dir
```

dvc/info.py
```diff
@@ -47,6 +47,13 @@ def get_dvc_info():
             if psutil:
                 fs_type = get_fs_type(repo.cache.local.cache_dir)
                 info.append(f"Cache directory: {fs_type}")
+                external_cache = repo.cache.local.cache_dir
+                remote_cache = _get_external_cache(repo)
+                if external_cache:
+                    info.append(f"External file cache: {external_cache}")
+                if remote_cache:
+                    info.append(f"External remote cache: {remote_cache}")
```
efiop, line 55: Seems like we can simply embed the cache dir into the `Cache directory` message above.

The remote caches are not dependant on `psutil`, so there is no reason to keep them under this `if`.

dvc/info.py
```diff
@@ -134,3 +146,39 @@ def _get_dvc_repo_info(self):
...
 
     return "dvc, git"
+
+
+def _get_external_remotes(self):
+
+    core_info = self.config.get("core", {})
+    remote_url = ""
+    if core_info.get("remote", False):
+        remote = core_info["remote"]
```
efiop, line 156: Seems wasteful to get it twice. Also `False` is not required.

efiop: Btw, we could `repo.cloud.get_remote` to get the remote name.

dvc/info.py
codingsett: Will cache_dir output the `external_cache_url` also? because i assumed it only outputs `directory_path` for caches

efiop: In that particular case - yes, because we are talking about the local (external or not) cache. Again, I do understand that this might be a bit confusing. :slightly_frowning_face: 

So default, config's `cache.local` and `cache.dir` will be covered by `repo.cache.local.cache_dir` above. And the rest should be then standalone (e.g. config's `cache.s3`, `cache.ssh`, etc)

codingsett: Yeah i get you now😃

efiop:
Closing as stale.

PR #4607, opened 2020-09-23
ci: get rid of travis

efiop:


snap/snapcraft.yaml
```diff
@@ -29,7 +29,6 @@ parts:
     - git
     override-pull: |
         snapcraftctl pull
-        snapcraftctl set-version $(cd $SNAPCRAFT_PART_SRC && git describe --tags)
```
efiop, line 32: For some reason there were no tags present when built by the action. Using a more appropriate `dvc --version` below instead.

.github/workflows/packages.yaml
```diff
@@ -0,0 +1,126 @@
...
+    - uses: actions/checkout@v2
+    - name: Build the package
+      uses: snapcore/action-build@v1
+      id: snapcraft
+    - name: Test the package
+      run: |
+        sudo snap install --dangerous --classic ${{ steps.snapcraft.outputs.snap }}
+        dvc version
+        sudo snap remove dvc
+    - name: Publish to the edge channel
```
efiop, line 113: publishing in all packages hasn't been tested for obvious reasons, so we'll be keeping an eye on it in the next release (we'll do a dummy minor one to test out).

.github/workflows/packages.yaml
```diff
@@ -0,0 +1,126 @@
+name: Build packages
+on: [push, pull_request, release]
```
efiop, line 2: Keeping push and pull_request because it matches the previous behavior. But could consider limiting to release-only in the near future.

.github/workflows/check-patch.yaml
```diff
@@ -14,6 +14,8 @@ jobs:
...
         python-version: 3.8
     - name: Install requirements
-      run: bash ./scripts/ci/install.sh
-    - name: Check Patch
-      run: ./scripts/ci/check_patch.sh
+      run: pip install ".[all,tests]" Pygments collective.checkdocs pre-commit
+    - name: Check README
+      run: python setup.py checkdocs
+    - name: Check formatting
+      run: pre-commit run --all-files
```
efiop, line 21: @pared Btw, would it make sense to include this into the test workflow as a separate job? Just a bit weird that this is a separate one, because it is just a yet another test.

efiop: Also, have you considered `fail-fast: false` for the test build matrix? Seems strange that all jobs are getting cancelled because of one flaky one.

scripts/ci/install.sh
```diff
@@ -1,22 +0,0 @@
...
-
-git config --global user.email "dvctester@example.com"
-git config --global user.name "DVC Tester"
-
-if [[ "$TRAVIS_SECURE_ENV_VARS" == "true" ]]; then
-  aws configure set aws_access_key_id $AWS_ACCESS_KEY_ID
-  aws configure set aws_secret_access_key $AWS_SECRET_ACCESS_KEY
-  aws configure set region us-west-2
-
-  openssl enc -d -aes-256-cbc -md md5 -k $GCP_CREDS -in scripts/ci/gcp-creds.json.enc -out scripts/ci/gcp-creds.json
```
efiop, line 21: For the record: we'll need to re-create gcp creds anyway.

.github/workflows/packages.yaml
skshetry: If you remove this for `push`/`pull_request`, consider making it `manual_dispatch` as it'd come handy to build these manually if required during PRs.

skshetry (approved):
Looks good to me.

.github/workflows/check-patch.yaml
pared: @efiop yeah, I guess it makes sense. In `dvc-bench` we have style checks run before test. If we want to restrict check to particular os-es, we can do that by using `if` statement. Also `fail-fast` sounds like a good idea.

pared (approved):

PR #4613, opened 2020-09-24
command: print the current cache directory when no cmd flags are provided

### LINKED ISSUE #1998: have `dvc cache dir` print the current cache dir when no cmd flags are used
> https://discordapp.com/channels/485586884165107732/565699007037571084/577605813271658519

Currently `dvc cache dir` by itself throws an error and display's the command help. It would be useful to, instead, print the location of the currently used cache dir, so that users can always know what it is (even if it's not specified in `.dvc/config` and thus unknown to `dvc config cache.dir`).

girish946:
comamnd 'dvc cache dir' now prints the current cache dir
when no cmd flags are provided.

Fixes #1998

Signed-off-by: Girish Joshi <girish946@gmail.com>

### DIFF
--- a/dvc/command/cache.py
+++ b/dvc/command/cache.py
@@ -7,6 +7,9 @@ from dvc.command.config import CmdConfig
 
 class CmdCacheDir(CmdConfig):
     def run(self):
+        if not self.args.value:
+            print(self.config["cache"]["dir"])
+            return 0
         with self.config.edit(level=self.args.level) as edit:
             edit["cache"]["dir"] = self.args.value
         return 0
@@ -55,6 +58,8 @@ def add_parser(subparsers, parent_parser):
         "value",
         help="Path to cache directory. Relative paths are resolved relative "
         "to the current directory and saved to config relative to the "
-        "config file location.",
+        "config file location. If no path is provided, it returns the "
+        "current cache directory.",
+        nargs="?",
     ).complete = completion.DIR
     cache_dir_parser.set_defaults(func=CmdCacheDir)
--- a/dvc/command/cache.py
+++ b/dvc/command/cache.py
@@ -7,7 +7,7 @@ from dvc.command.config import CmdConfig
 
 class CmdCacheDir(CmdConfig):
     def run(self):
-        if not self.args.value:
+        if self.args.value is None and not self.args.unset:
             print(self.config["cache"]["dir"])
             return 0
         with self.config.edit(level=self.args.level) as edit:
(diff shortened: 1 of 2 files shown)


dvc/command/cache.py
```diff
@@ -7,6 +7,9 @@
 
 class CmdCacheDir(CmdConfig):
     def run(self):
+        if not self.args.value:
+            print(self.config["cache"]["dir"])
+            return 0
```
efiop, line 12: We have a similar logic in `dvc config`: https://github.com/iterative/dvc/blob/738dfd49cf403412b018677e62b0b68e85c80886/dvc/command/config.py#L19 , but it also checks for --unset and uses logger.info  (our current convention). Maybe we could do the same here?

efiop (changes requested):
Hi @girish946 !

Thank you for your patience! Looks good! Please see a comment above.

dvc/command/cache.py
girish946: > We have a similar logic in `dvc config`:


>  , but it also checks for --unset and uses logger.info (our current convention). Seems like we could do the same here.

done.

girish946:
Hi @efiop,
Thanks for the review.
I've made the change mentioned above.

dvc/command/cache.py
```diff
@@ -7,6 +7,9 @@
 
 class CmdCacheDir(CmdConfig):
     def run(self):
+        if self.args.value is None and not self.args.unset:
+            print(self.config["cache"]["dir"])
```
efiop, line 11:
```suggestion
            logger.info(self.config["cache"]["dir"])
```

dvc/command/cache.py
```diff
@@ -7,6 +7,9 @@
 
 class CmdCacheDir(CmdConfig):
     def run(self):
+        if self.args.value is None and not self.args.unset:
+            logger.info(self.config["cache"]["dir"])
```
efiop, line 11: Unlike https://github.com/iterative/dvc/blob/738dfd49cf403412b018677e62b0b68e85c80886/dvc/command/config.py#L19 , we are not taking `level` into account :slightly_frowning_face:

girish946: My bad. I should have looked at logging.
Thanks for correcting.
Will take a pull and test it once.

efiop:
Thank you @girish946 ! :pray:

jorgeorpinel:
Hi @girish946 would you like to also update the doc in https://dvc.org/doc/command-reference/cache/dir to reflect the change? (See https://dvc.org/doc/user-guide/contributing/docs) Np either way, just lmk please so we take care of it otherwise. Best

girish946:
Hi @jorgeorpinel 
Thanks for reminding me about the documentation part that is remaining.
I'll update it in a day or two.

Best.

jorgeorpinel:
Sounds good, thanks @girish946

PR #4626, opened 2020-09-28
diff/dvctree: optimize dir cache access

### LINKED ISSUE #4580: diff: appears to hang for large directories
Looks like diff may be trying to re-fetch dir cache for every file in the directory instead of only fetching dir cache once and then walking it.

discord context: https://discordapp.com/channels/485586884165107732/485596304961962003/756269362482118656

`dvc diff -v` output for example-get-started
```
example-get-started git:master  py:dvc ❯ dvc diff -v
2020-09-18 10:01:15,523 DEBUG: Check for update is enabled.
2020-09-18 10:01:15,525 DEBUG: fetched: [(3,)]                            
2020-09-18 10:01:15,548 DEBUG: Path '/Users/pmrowla/git/scratch/example-get-started/data/prepared' inode '26117533'
2020-09-18 10:01:15,549 DEBUG: fetched: [('d93097474331d010b3b28be06f81968e', '23861319', '20b786b6e6f80e2b3fcf17827ad18597.dir', '1600390781154654976')]
2020-09-18 10:01:15,550 DEBUG: Assuming '/Users/pmrowla/git/scratch/example-get-started/.dvc/cache/20/b786b6e6f80e2b3fcf17827ad18597.dir' is unchanged since it is read-only
2020-09-18 10:01:15,551 DEBUG: Assuming '/Users/pmrowla/git/scratch/example-get-started/.dvc/cache/20/b786b6e6f80e2b3fcf17827ad18597.dir' is unchanged since it is read-only
2020-09-18 10:01:15,551 DEBUG: Assuming '/Users/pmrowla/git/scratch/example-get-started/.dvc/cache/20/b786b6e6f80e2b3fcf17827ad18597.dir' is unchanged since it is read-only
2020-09-18 10:01:15,551 DEBUG: Assuming '/Users/pmrowla/git/scratch/example-get-started/.dvc/cache/20/b786b6e6f80e2b3fcf17827ad18597.dir' is unchanged since it is read-only
2020-09-18 10:01:15,552 DEBUG: Path '/Users/pmrowla/git/scratch/example-get-started/data/features' inode '26117536'
2020-09-18 10:01:15,552 DEBUG: fetched: [('9b93598d23ee81a582ecf3a954adab85', '10895463', 'a1414b22382ffbb76a153ab1f0d69241.dir', '1600390781159751936')]
2020-09-18 10:01:15,553 DEBUG: Assuming '/Users/pmrowla/git/scratch/example-get-started/.dvc/cache/a1/414b22382ffbb76a153ab1f0d69241.dir' is unchanged since it is read-only
2020-09-18 10:01:15,554 DEBUG: Assuming '/Users/pmrowla/git/scratch/example-get-started/.dvc/cache/a1/414b22382ffbb76a153ab1f0d69241.dir' is unchanged since it is read-only
2020-09-18 10:01:15,554 DEBUG: Assuming '/Users/pmrowla/git/scratch/example-get-started/.dvc/cache/a1/414b22382ffbb76a153ab1f0d69241.dir' is unchanged since it is read-only
2020-09-18 10:01:15,554 DEBUG: Assuming '/Users/pmrowla/git/scratch/example-get-started/.dvc/cache/a1/414b22382ffbb76a153ab1f0d69241.dir' is unchanged since it is read-only
2020-09-18 10:01:15,555 DEBUG: Path '/Users/pmrowla/git/scratch/example-get-started/model.pkl' inode '26117539'
2020-09-18 10:01:15,555 DEBUG: fetched: [('1599633290750908672', '2732613', 'c8d307aa005d6974a8525550956d5fb3', '1600390781163895040')]
2020-09-18 10:01:15,555 DEBUG: Path '/Users/pmrowla/git/scratch/example-get-started/scores.json' inode '25389029'
2020-09-18 10:01:15,555 DEBUG: fetched: [('1599209322085689088', '27', '0d2602f440d13372435d91d3bfa17652', '1600390781164561920')]
2020-09-18 10:01:15,555 DEBUG: Path '/Users/pmrowla/git/scratch/example-get-started/prc.json' inode '25389028'
2020-09-18 10:01:15,556 DEBUG: fetched: [('1599209322085403904', '2252', 'a85c2e5f38affad070f1a6e0f837d0a9', '1600390781165104896')]
2020-09-18 10:01:15,556 DEBUG: Path '/Users/pmrowla/git/scratch/example-get-started/data/data.xml' inode '26117532'
2020-09-18 10:01:15,556 DEBUG: fetched: [('1599633321385457920', '37891850', 'a304afb96060aad90176268345e10355', '1600390781165520896')]
2020-09-18 10:01:15,586 DEBUG: Assuming '/Users/pmrowla/git/scratch/example-get-started/.dvc/cache/20/b786b6e6f80e2b3fcf17827ad18597.dir' is unchanged since it is read-only
2020-09-18 10:01:15,586 DEBUG: Assuming '/Users/pmrowla/git/scratch/example-get-started/.dvc/cache/20/b786b6e6f80e2b3fcf17827ad18597.dir' is unchanged since it is read-only
2020-09-18 10:01:15,587 DEBUG: Assuming '/Users/pmrowla/git/scratch/example-get-started/.dvc/cache/20/b786b6e6f80e2b3fcf17827ad18597.dir' is unchanged since it is read-only
...
2020-09-18 10:01:15,605 DEBUG: Assuming '/Users/pmrowla/git/scratch/example-get-started/.dvc/cache/a1/414b22382ffbb76a153ab1f0d69241.dir' is unchanged since it is read-only
2020-09-18 10:01:15,606 DEBUG: fetched: [(22,)]
2020-09-18 10:01:15,608 DEBUG: Analytics is disabled.
```

pmrowla:

Will fix #4580.

* `HashInfo.dir_info` is now stored as a dict instead of list, dir cache is still kept in the list format everywhere else
* `DvcTree._get_granular_checksum` now uses `out.hash_info.dir_info` to lookup file hashes

### DIFF
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -263,6 +263,8 @@ class CloudCache:
         from dvc.utils import tmp_fname
 
         # Sorting the list by path to ensure reproducibility
+        if isinstance(dir_info, dict):
+            dir_info = self._from_dict(dir_info)
         dir_info = sorted(dir_info, key=itemgetter(self.tree.PARAM_RELPATH))
 
         tmp = tempfile.NamedTemporaryFile(delete=False).name
@@ -275,7 +277,7 @@ class CloudCache:
 
         hash_info = self.tree.get_file_hash(to_info)
         hash_info.value += self.tree.CHECKSUM_DIR_SUFFIX
-        hash_info.dir_info = dir_info
+        hash_info.dir_info = self._to_dict(dir_info)
 
         return hash_info, to_info
 
@@ -299,15 +301,14 @@ class CloudCache:
 
     def _save_dir(self, path_info, tree, hash_info, save_link=True, **kwargs):
         if not hash_info.dir_info:
-            hash_info.dir_info = tree.cache.get_dir_cache(hash_info)
-        hi = self.save_dir_info(hash_info.dir_info, hash_info)
-        for entry in Tqdm(
-            hi.dir_info, desc="Saving " + path_info.name, unit="file"
-        ):
-            entry_info = path_info / entry[self.tree.PARAM_RELPATH]
-            entry_hash = HashInfo(
-                self.tree.PARAM_CHECKSUM, entry[self.tree.PARAM_CHECKSUM]
+            hash_info.dir_info = self._to_dict(
+                tree.cache.get_dir_cache(hash_info)
             )
+        hi = self.save_dir_info(hash_info.dir_info, hash_info)
+        for relpath, entry_hash in Tqdm(
+            hi.dir_info.items(), desc="Saving " + path_info.name, unit="file"
+        ):
+            entry_info = path_info / relpath
             self._save_file(
                 entry_info, tree, entry_hash, save_link=False, **kwargs
             )
@@ -615,7 +616,9 @@ class CloudCache:
 
     def _to_dict(self, dir_info):
         return {
-            entry[self.tree.PARAM_RELPATH]: entry[self.tree.PARAM_CHECKSUM]
+            entry[self.tree.PARAM_RELPATH]: HashInfo(
+                self.tree.PARAM_CHECKSUM, entry[self.tree.PARAM_CHECKSUM]
+            )
             for entry in dir_info
         }
 
@@ -623,9 +626,9 @@ class CloudCache:
         return [
             {
                 self.tree.PARAM_RELPATH: relpath,
-                self.tree.PARAM_CHECKSUM: checksum,
+                hash_info.name: hash_info.value,
             }
-            for relpath, checksum in dir_dict.items()
+            for relpath, hash_info in dir_dict.items()
         ]
 
     @staticmethod
@@ -692,3 +695,8 @@ class CloudCache:
             return hash_info
 
         return self.save_dir_info(hash_info.dir_info, hash_info)
+
+    def set_dir_info(self, hash_info):
+        assert hash_info.isdir
+
+        hash_info.dir_info = self._to_dict(self.get_dir_cache(hash_info))
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -615,12 +615,17 @@ class CloudCache:
         )
 
     def _to_dict(self, dir_info):
-        return {
-            entry[self.tree.PARAM_RELPATH]: HashInfo(
-                self.tree.PARAM_CHECKSUM, entry[self.tree.PARAM_CHECKSUM]
-            )
-            for entry in dir_info
-        }
+        info = {}
+        for entry in dir_info:
+            relpath = None
+            hash_info = None
+            for key, value in entry.items():
+                if key == self.tree.PARAM_RELPATH:
+                    relpath = value
+                else:
+                    hash_info = HashInfo(key, value)
+            info[relpath] = hash_info
+        return info
 
     def _from_dict(self, dir_dict):
         return [
--- a/dvc/repo/diff.py
+++ b/dvc/repo/diff.py
@@ -111,7 +111,7 @@ def _dir_output_paths(repo_tree, output):
 
     try:
         for fname in repo_tree.walk_files(output.path_info):
-            yield str(fname), repo_tree.get_hash(fname).value
+            yield str(fname), repo_tree.get_file_hash(fname).value
     except NoRemoteError:
         logger.warning("dir cache entry for '%s' is missing", output)
 
--- a/dvc/tree/dvc.py
+++ b/dvc/tree/dvc.py
@@ -56,12 +56,16 @@ class DvcTree(BaseTree):  # pylint:disable=abstract-method
         assert isinstance(path_info, PathInfo)
         if not self.fetch and not self.stream:
             raise FileNotFoundError
+
+        # NOTE: use string paths here for performance reasons
+        path_str = path_info.as_posix()
+        out_path_str = out.path_info.as_posix()
         dir_cache = out.get_dir_cache(remote=remote)
         for entry in dir_cache:
             entry_relpath = entry[out.tree.PARAM_RELPATH]
             if os.name == "nt":
                 entry_relpath = entry_relpath.replace("/", os.sep)
-            if path_info == out.path_info / entry_relpath:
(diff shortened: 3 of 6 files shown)


pmrowla:
Current changes in a repo w/directory containing 1K files cuts cprofile time for `dvc diff` from ~20s to ~2s. Still investigating why diff remains very slow with larger directories (100K+ files)

dvc/tree/dvc.py
```diff
@@ -11,6 +12,24 @@
 logger = logging.getLogger(__name__)
 
 
+# cache metadata for sequential exists/isdir/isfile/etc calls
+@lru_cache(maxsize=1)
+def _get_metadata(tree, path_info):
```
pmrowla, line 17: same `lru_cache(maxsize=1)` optimization that we use in git tree `_get_object_by_path`

efiop: With git it is pretty safe, as tree is read only, tied to a specific reference. Are there scenarios in which this could cause a bug?

efiop: Btw, which method was causing too much `metadata` calls? Would it be solved by our planned exception unification and moving to an exception-based workflow?

pmrowla: In `RepoTree.get_hash` we end up making several repeated `metadata` calls through `DvcTree.exists`. I think moving to an exception based workflow and avoiding all the `exists` calls would definitely help in this case.

pmrowla: I suppose in theory we could run into issues here with the dvc repo (and repo tree) being modified in between metadata calls. I can adjust this so that we only cache metadata calls within the scope of `dvctree.walk`, since the tree should not be changing within a single `walk` call.

pmrowla: Ended up reverting this change, as the main performance blocker in our metadata calls was still `_get_granular_checksum`. As noted before though, moving to an exception based workflow should cut down on duplicated metadata calls

pmrowla:
For a repo containing a single tracked directory with 100k files, cprofiled `dvc diff` in master previously took almost 2 hours on my machine, with these changes it's down to around 3 minutes. 3 minutes still seems kind of slow for a diff command, especially since we display no progress or status information, but it's better than before at least.

master
Screen Shot 2020-09-29 at 6 23 10 PM

PR branch
Screen Shot 2020-09-29 at 6 23 31 PM

dvc/tree/dvc.py
```diff
@@ -45,18 +47,39 @@ def _is_cached(out):
...
+        path_str = str(path)
+        hash_info = self._dir_entry_hashes.get(path_str)
+        if hash_info:
+            return hash_info
+        raise FileNotFoundError
+
+    def _update_dir_entry_hashes(self, out, remote):
+        # cache the most recently used output dir cache to avoid expensive
+        # repeated lookups of individual files within the same large output dir
         dir_cache = out.get_dir_cache(remote=remote)
```
efiop, line 67: So this particular one is slow, right?

Note that we have dir_cache caching in the Cache itself https://github.com/iterative/dvc/blob/7b590e7c5d9361a611443c057313a6686b146f50/dvc/cache/base.py#L64 , wonder why that one wasn't enough and we need to introduce another level.

efiop: Probably because iterating over the whole dir_cache to find one hash is slow. Do I understand correctly that `walk` was slow? If so, would it be possible to bulk up the operation somehow so we don't `metadata()` for each file in a giant dataset?

dvc/repo/diff.py
```diff
@@ -111,7 +111,7 @@ def _dir_output_paths(repo_tree, output):
 
     try:
         for fname in repo_tree.walk_files(output.path_info):
-            yield str(fname), repo_tree.get_hash(fname).value
+            yield str(fname), repo_tree.get_file_hash(fname).value
```
efiop, line 114: So this enters a giant dataset and thus has to call get_file_hash for each individual file in it, which is slow, right?

Hm, feels like we could bulk this up somehow. Maybe we could call `get_hash(output.path_info)`  and then use `.dir_info`(part of HashInfo) to receive it in bulk? Clearly there are some things like the dirty state that we'll need to keep in mind, but maybe that could be a reasonable approach to save up on metadata. Classic stat&walk problem :slightly_smiling_face:

dvc/tree/dvc.py
pmrowla: Right, `walk` was slow because we would re-iterate over the whole dir_cache when looking for each individual entry/file hash

dvc/repo/diff.py
pmrowla: This could work, but part of the issue right now is that dir_info/dir_cache is a list, so looking up files by the relpath field is slow

pmrowla: For this diff case we could just get `.dir_info` and then iterate over that instead of calling `tree.walk()`, but for me this really feels like something that should be abstracted away inside of `walk()` (and that `get_file_hash` inside of a walk should just work without introducing any giant performance overhead)

efiop: Really good point about the dir_info being a list! We could totally switch to dicts at least in a cached version, that will make it easier to work with everywhere. Wonder if that would mitigate the performance issue here, as it seems much more intuitive than the current two-level cache.

dvc/cache/base.py
```diff
@@ -614,18 +615,25 @@ def get_files_number(self, path_info, hash_info, filter_info):
...
+        for entry in dir_info:
+            relpath = None
+            hash_info = None
+            for key, value in entry.items():
+                if key == self.tree.PARAM_RELPATH:
+                    relpath = value
+                else:
+                    hash_info = HashInfo(key, value)
+            info[relpath] = hash_info
+        return info
```
pmrowla, line 628: This looks a bit ugly but dir_info may actually contain hashes with a different type than `cache.tree.PARAM_CHECKSUM` - i.e. dir_info contains S3 etag entries, but we are working in local cache (w/md5 `PARAM_CHECKSUM`)

pared (approved):

pared:
@pmrowla there are some merge conflicts

efiop (approved):

PR #4627, opened 2020-09-28
metrics/plots: initial support non-dvc repos

efiop:
Initial hacky support, that still tries to create `.dvc` directory as a side-effect. That side effect will be addressed in the upcoming 
PR. This feature in general makes us really rethink what a dvc repo is and whether having `.dvc` is really required.

### DIFF
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -80,7 +80,14 @@ class Repo:
     from dvc.repo.status import status
     from dvc.repo.update import update
 
-    def __init__(self, root_dir=None, scm=None, rev=None, subrepos=False):
+    def __init__(
+        self,
+        root_dir=None,
+        scm=None,
+        rev=None,
+        subrepos=False,
+        uninitialized=False,
+    ):
         from dvc.cache import Cache
         from dvc.data_cloud import DataCloud
         from dvc.lock import make_lock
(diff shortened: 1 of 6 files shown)


dvc/repo/__init__.py
```diff
@@ -80,7 +80,14 @@ class Repo:
...
     from dvc.repo.update import update
 
-    def __init__(self, root_dir=None, scm=None, rev=None, subrepos=False):
+    def __init__(
+        self,
+        root_dir=None,
+        scm=None,
+        rev=None,
+        subrepos=False,
+        uninitialized=False,
```
efiop, line 89: We might consider getting rid of this in the future (for now needed for backward compatibility). Currently `dvc init` doesn't do much useful things and all `.dvc` contents could be (and should probably) be created dynamically.

Even things like `dvc add` could potentially just work in a pure git repo. Just something to think about...

jorgeorpinel:
> This feature in general makes us really rethink what a dvc repo is and whether having .dvc is really required.

Interesting (and I've noticed this mentioned elsewhere) can you elaborate a little? Or is there a discussion somewhere I can keep an eye and participate on this idea? Thanks

efiop:
@jorgeorpinel Not yet, only verbal discussions at this point. We'll probably create something in the future.

PR #4641, opened 2020-09-29
WIP: Custom parametrization/imports

skshetry:
Expansion is based on `${var}` syntax, whereas `${{ var }}` can be
used for auto-tracking + expansion.

It needs the strucuture for `dvc.yaml` to be:
```yaml
args:
  imports:
    - params.yaml
  constants:
    a: 3

stages:
  ...
```

Auto-tracking is limited to imports from `params.yaml`.

I don't intend to merge this, this is just a simple-prototype to see how it feels. Looking to ask @dmpetrov for feedback later, this only works on happy-path cases and will report very bad error messages.

### DIFF
--- a/dvc/dvcfile.py
+++ b/dvc/dvcfile.py
@@ -105,7 +105,7 @@ class FileMixin:
         with self.repo.tree.open(self.path) as fd:
             stage_text = fd.read()
         d = parse_yaml(stage_text, self.path)
-        self.validate(d, self.relpath)
+        # self.validate(d, self.relpath)
         return d, stage_text
 
     @classmethod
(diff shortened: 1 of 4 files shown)


dvc/dvcfile.py
```diff
@@ -225,9 +225,13 @@ def stage(self):
...
     @property
     def stages(self):
+        from dvc.parsing import DataLoader
+
         data, _ = self._load()
+        spec = DataLoader(data)
+        resolved = spec.resolve()
         lockfile_data = self._lockfile.load()
-        return StageLoader(self, data.get("stages", {}), lockfile_data)
+        return StageLoader(self, resolved.get("stages", {}), lockfile_data)
```
skshetry, line 234: Hack for now, we'll need to short-circuit this function to the repo somehow.

I like `repo.stages.get()` syntax (that means, we will need to think something for current `repo.stages`.

PR #4646, opened 2020-09-30
Fixed the bug #4251, and fixed a minor bug which causes of OS default character.

### LINKED ISSUE #4251: `get-url` should just `GET` files

hl-a-k:

Fix #4251  
The reason we do this `save()` call here is just to check if the source exists. So I replace this with 
`
if in_repo:
    dep.save()
`
By the way, fixed a minor bug.
bug description:
1, environment: OS which default character is not 'UTF-8'. For example windows 10 which default character is 'gbk'.
2, operation:
python setup.py install
  File "setup.py", line 149, in <module>
    long_description=open("README.rst", "r").read(),
UnicodeDecodeError: 'gbk' codec can't decode byte 0xa2 in position 41: illegal multibyte sequence

### DIFF
--- a/dvc/repo/get_url.py
+++ b/dvc/repo/get_url.py
@@ -16,6 +16,7 @@ def get_url(url, out=None):
 
     (dep,) = dependency.loads_from(None, [url])
     (out,) = output.loads_from(None, [out], use_cache=False)
-    dep.save()
+    if not dep.exists:
+        raise dep.DoesNotExistError(dep)
     dep.download(out)
     out.save()
--- a/dvc/repo/get_url.py
+++ b/dvc/repo/get_url.py
@@ -6,7 +6,7 @@ from dvc.utils import resolve_output
 
 
 @staticmethod
-def get_url(url, out=None):
+def get_url(url, out=None, in_repo=True):
     out = resolve_output(url, out)
 
     if os.path.exists(url):
@@ -16,6 +16,7 @@ def get_url(url, out=None):
 
     (dep,) = dependency.loads_from(None, [url])
     (out,) = output.loads_from(None, [out], use_cache=False)
-    dep.save()
+    if in_repo:
+        dep.save()
     dep.download(out)
     out.save()
--- a/dvc/repo/get_url.py
+++ b/dvc/repo/get_url.py
@@ -6,7 +6,7 @@ from dvc.utils import resolve_output
 
 
 @staticmethod
-def get_url(url, out=None, in_repo=True):
+def get_url(url, out=None):
     out = resolve_output(url, out)
 
     if os.path.exists(url):
@@ -16,7 +16,6 @@ def get_url(url, out=None, in_repo=True):
 
     (dep,) = dependency.loads_from(None, [url])
     (out,) = output.loads_from(None, [out], use_cache=False)
-    if in_repo:
-        dep.save()
+    dep.save()
     dep.download(out)
     out.save()
--- a/setup.py
+++ b/setup.py
@@ -147,7 +147,7 @@ setup(
     name="dvc",
     version=version,
     description="Git for data scientists - manage your code and data together",
-    long_description=open("README.rst", "r").read(),
+    long_description=open("README.rst", "r", encoding="UTF-8").read(),
     author="Dmitry Petrov",
     author_email="dmitry@dvc.org",
     download_url="https://github.com/iterative/dvc",
--- a/dvc/command/get_url.py
+++ b/dvc/command/get_url.py
@@ -14,7 +14,7 @@ class CmdGetUrl(CmdBaseNoRepo):
         from dvc.repo import Repo
 
         try:
-            Repo.get_url(self.args.url, out=self.args.out)
+            Repo.get_url(self.args.url, out=self.args.out, in_repo=False)
             return 0
         except DvcException:
             logger.exception(f"failed to get '{self.args.url}'")
--- a/dvc/command/get_url.py
+++ b/dvc/command/get_url.py
@@ -14,7 +14,7 @@ class CmdGetUrl(CmdBaseNoRepo):
         from dvc.repo import Repo
 
         try:
-            Repo.get_url(self.args.url, out=self.args.out, in_repo=False)
+            Repo.get_url(self.args.url, out=self.args.out)
             return 0
         except DvcException:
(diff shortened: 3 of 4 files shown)


skshetry:
Hi, @hl-a-k. Thanks for contributing. :tada:  :slightly_smiling_face: 

It'd be helpful if you could provide more information regarding the PR/fix in the descriptin above. Thanks, we'll review it soon.

dvc/repo/get_url.py
```diff
@@ -16,6 +16,7 @@ def get_url(url, out=None):
 
     (dep,) = dependency.loads_from(None, [url])
     (out,) = output.loads_from(None, [out], use_cache=False)
-    dep.save()
+    if in_repo:
+        dep.save()
```
efiop, line 20: The unfortunate issue with not calling `save` is that our exceptions in the logic down below are not well unified, so it won't present a nice error when, for example, the source doesn't exist :slightly_frowning_face: I.e. `download` will trigger a method in one of the *Tree classes (in dvc/tree/*), and they all might throw a different backend-specific error, that is pretty ugly :slightly_frowning_face: We are on our way to unifying tree exceptions, but we are not there yet.

Might be missing something, but the only reason we do this `save()` call here is to check if the source exists. If we replace this with `if not dep.exists(): raise DependencyDoesNotExistError(...)`, that should do the trick for now

Also, if I understand correctly, you've introduced this flag to avoid changing the API behavior, right? So like for a backward compatibility? If so, we could totally simply switch to the new behaviour and remove it.

setup.py
```diff
@@ -147,7 +147,7 @@ def run(self):
     name="dvc",
     version=version,
     description="Git for data scientists - manage your code and data together",
-    long_description=open("README.rst", "r").read(),
+    long_description=open("README.rst", "r", encoding="UTF-8").read(),
```
efiop, line 150: Oh, nice catch!

efiop:
@hl-a-k Thank you for your patience and contribution :slightly_smiling_face: A few comments in addition to the ones above:

1) Looks like github is not able to associate your commits with your github profile. Please go to https://github.com/settings/emails and add the email used in your commits.

2) Please adjust the PR title to something more meaningful.

dvc/repo/get_url.py
hl-a-k: You're right. By introducing this flag, we can keep the original logic unaffected.

hl-a-k:
> @hl-a-k Thank you for your patience and contribution 🙂 A few comments in addition to the ones above:
> 
> 1. Looks like github is not able to associate your commits with your github profile. Please go to https://github.com/settings/emails and add the email used in your commits.
> 2. Please adjust the PR title to something more meaningful.
> 3. Please use Fix #issue-number (e.g. Fix #4251) in your PR description, so github will automatically close the issue when your PR is merged.
> 
> Thank you! 🙏
> 
> Btw, the reason we didn't reply earlier is that we were waiting for the PR description update and when you did edit it, github didn't send a notification (it doesn't for message edits, so that's expected). Please feel free to leave a comment in such cases next time, we'll be sure to get to it ASAP. 🙂

hl-a-k:
I have completed the changes.

efiop:
@hl-a-k Looks like you didn't add `zwh@geeboo.cn`(used in your commits) to https://github.com/settings/emails . If you did, then maybe it is just github that needs some time to reparse this :slightly_smiling_face:

dvc/repo/get_url.py
efiop: Makes sense. So let's get rid of it and just replace
```
if in_repo:
    dep.save()
```
with
```
if not dep.exists():
    raise dep.DoesNotExistError(dep)
```

hl-a-k: Changes committed. Seems nothing happened. ^_^

hl-a-k:
> @hl-a-k Looks like you didn't add `zwh@geeboo.cn`(used in your commits) to https://github.com/settings/emails . If you did, then maybe it is just github that needs some time to reparse this 🙂

Sorry, I don't use `zwh@geeboo.cn` anymore. I have updated my git configuration now.

dvc/repo/get_url.py
```diff
@@ -16,6 +16,7 @@ def get_url(url, out=None):
 
     (dep,) = dependency.loads_from(None, [url])
     (out,) = output.loads_from(None, [out], use_cache=False)
-    dep.save()
+    if not dep.exists:
```
efiop, line 19: Sorry, I totally forgot that the issue itself was about `exist` not working for some urls :slightly_frowning_face: So this won't actually solve it. My mistake, sorry for the confusion.

Ok, let me see if the exception approach mentioned above is feasible to implement...

hl-a-k: At least, this approach work for #4251 . I have tested.

hl-a-k: How is the current situation？

efiop: Didn't get to research this yet. Need to confirm this and we'll merge if it is indeed fixed. Looks good as is. Thanks!

efiop: @hl-a-k Thanks for your patience! Indeed, it was the `get_hash` that was causing that issue, so the current solution works like charm. Thank you! :pray:

efiop (approved):
Thank you! :rocket:

PR #4647, opened 2020-09-30
Run CI for Python3.9

skshetry:
Run CI for Python3.9, skips hdfs tests for 3.9
d673e2a

Also, `dvc[hdfs]` install will be a no-op till we have
a pyarrow wheel for 3.9.

Note that this does not add an entry to the classifier
claiming it supports 3.9, that will be added when 3.9
gets released (or, wait for pyarrow to be released).

### DIFF
--- a/dvc/scm/git.py
+++ b/dvc/scm/git.py
@@ -177,7 +177,7 @@ class Git(Base):
 
         gitignore = os.path.join(ignore_file_dir, self.GITIGNORE)
 
-        if not path_isin(gitignore, os.path.realpath(self.root_dir)):
+        if not path_isin(os.path.realpath(gitignore), self.root_dir):
             raise FileNotInRepoError(path)
 
         return entry, gitignore
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,4 +1,10 @@
 import os
+import sys
+
+# FIXME: Search and replace these from the tests if pyarrow wheel is available
+PY39 = sys.version_info >= (3, 9, 0)
+PYARROW_NOT_AVAILABLE = "pyarrow not available yet for Python3.9"
+
 
 # Increasing fd ulimit for tests
 if os.name == "nt":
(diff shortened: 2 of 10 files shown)


skshetry:
These test failures look completely unrelated. Looks like CI is not that stable.

tests/unit/test_info.py
```diff
@@ -6,6 +6,12 @@
 
 from dvc.info import get_dvc_info, psutil
 
+# PYthon's version is in the shape of:
```
efiop, line 9:
```suggestion
# Python's version is in the shape of:
```

tests/unit/output/test_hdfs.py
```diff
@@ -1,7 +1,11 @@
+import pytest
+
 from dvc.output.hdfs import HDFSOutput
+from tests import PY39, PYARROW_NOT_AVAILABLE
 from tests.unit.output.test_local import TestLocalOutput
 
 
+@pytest.mark.skipif(PY39, reason=PYARROW_NOT_AVAILABLE)
```
efiop, line 8: I guess we could've created a custom marker, but this will do too.

efiop (approved):

tests/__init__.py
```diff
@@ -1,4 +1,10 @@
 import os
+import sys
+
+# FIXME: Search and replace these from the tests if pyarrow wheel is available
```
efiop, line 4: This will get lost. How about a ticket? (Yeah, that one could be lost to).

skshetry: It won't, as I will come to this next week, as Python3.9 gets released.

dvc/scm/git.py
```diff
@@ -177,7 +177,7 @@ def _get_gitignore(self, path):
 
         gitignore = os.path.join(ignore_file_dir, self.GITIGNORE)
 
-        if not path_isin(gitignore, os.path.realpath(self.root_dir)):
+        if not path_isin(os.path.realpath(gitignore), self.root_dir):
             raise FileNotInRepoError(path)
```
skshetry, line 181: This exception here does not seem very precise,

efiop (approved):

efiop:
Thank you! :pray:

PR #4660, opened 2020-10-02
tests: cloud: ssh: assert test execution on same drive as cache

pared:

### DIFF
--- a/tests/func/test_api.py
+++ b/tests/func/test_api.py
@@ -1,11 +1,10 @@
 import os
-import platform
 
 import pytest
 from funcy import first, get_in
 
 from dvc import api
-from dvc.exceptions import DvcException, FileMissingError, OutputNotFoundError
+from dvc.exceptions import FileMissingError, OutputNotFoundError
 from dvc.path_info import URLInfo
 from dvc.utils.fs import remove
 from tests.unit.tree.test_repo import make_subrepo
@@ -26,7 +25,7 @@ def test_get_url(tmp_dir, dvc, remote):
 
 
 @pytest.mark.parametrize("cloud", clouds)
-def test_get_url_external(erepo_dir, cloud):
+def test_get_url_external(tmp_dir, erepo_dir, cloud):
     erepo_dir.add_remote(config=cloud.config)
     with erepo_dir.chdir():
         erepo_dir.dvc_gen("foo", "foo", commit="add foo")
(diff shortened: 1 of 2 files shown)


tests/func/test_api.py
```diff
@@ -62,29 +61,20 @@ def test_open(tmp_dir, dvc, remote):
...
+            "gdrive",
+            "oss",
+            "http",
+            "hdfs",
+            "ssh",
+        ]
     ],
 )
-def test_open_external(erepo_dir, cloud):
+def test_open_external(tmp_dir, erepo_dir, cloud):
```
efiop, line 77: Why is `tmp_dir` needed now?

tests/remotes/ssh.py
```diff
@@ -27,14 +27,23 @@ def get_url(user, port):  # pylint: disable=arguments-differ
             # a proper posixpath.
             # To do that, we should construct a posixpath that would be
             # relative to the server's root.
-            # In our case our ssh server is running with `c:/` as a root,
```
pared, line 30: Not always true.

tests/func/test_api.py
pared: Github actions clones the repository on `D`, so SSH server is initialized in a different drive than our storage. In order to be able to use SSH, we need to use Posix  style paths, and during SSH initialization we omit the drive information. This results in a situation, where we cannot access our storage. Added check which will let developers know about such situation.

tests/func/test_api.py
```diff
@@ -26,7 +25,7 @@ def test_get_url(tmp_dir, dvc, remote):
 
 
 @pytest.mark.parametrize("cloud", clouds)
-def test_get_url_external(erepo_dir, cloud):
+def test_get_url_external(tmp_dir, erepo_dir, cloud):
```
pared, line 28: In this case, api.get_url() has been returning non-existing path - cache was on `C` drive, while SSH server was on `D` (in case of github actions + Win). I think that we should check whether the path exist on `api.get_url()`. I would presume that in practical application user would like to follow up on `get_url` with download or something similar.
@efiop what do you think?

tests/func/test_api.py
efiop: Oh, great catch! :fire:  I remember having some ssh tests like these too, could you please grep the tests to see if there are any ssh tests like that?

Also seems like using `tmp_dir` here is still more of a hack, because the only reason we use it is for ssh to be initialized properly. Maybe we could alter the ssh fixture to start sftp where we want it (on C i suppose)?

tests/func/test_api.py
efiop: @pared `get_url` might not have access to the remote to be able to check for existence. I would rather not do that for now.

pared:
Ok, so I tried creating our own SSH server that would start on the proper drive, but the problem is that server runs on thread - being susceptible to us changing the directory during test execution. If we go back to test directory after starting server, `cwd` for it changes too. It could probably work if the server was started in its own process, but that would require some more changes. For now I would go with hack - using `tmp_dir` and create a new issue which aim would be to "keep" server in the proper place.

pared:
@efiop I checked other tests
`tests/unit/remote/ssh/test_ssh.py::test_hardlink_optimization`
seems to be using the same hack. I did not found other tests that would use clouds or `ssh` specifically and have unused `tmp_dir`.
If we agree to my proposed [approach](https://github.com/iterative/dvc/pull/4660#issuecomment-708431849), I will include tests in new issue.

tests/remotes/ssh.py
```diff
@@ -27,13 +27,19 @@ def get_url(user, port):  # pylint: disable=arguments-differ
...
             # [1]https://tools.ietf.org/html/draft-ietf-secsh-filexfer-13#section-6
             drive, path = os.path.splitdrive(path)
-            assert drive.lower() == "c:"
+
+            # Hackish way to make sure SSH server runs on same drive as tests
+            # and temporary directories are.
+            # Context: https://github.com/iterative/dvc/pull/4660
+            test_drive, _ = os.path.splitdrive(os.getcwd())
+            if drive.lower() != test_drive.lower():
+                raise Exception("Did you forget to use `tmp_dir?`")
```
efiop, line 41: :+1:

efiop (approved):

PR #4662, opened 2020-10-03
Azure service principal authentication WIP

### LINKED ISSUE #4635: Azure service principal authentication
I want to be able to authenticate against my remote using service principal (service account) credentials. The motivation being that these carry some identity when transacting against the data in the cloud which is a necessity for my auditing solution for sensitive data. 

[This](https://stackoverflow.com/questions/54324146/azure-datalake-with-dvc) is another individual after the same feature.
[This](https://discord.com/channels/485586884165107732/485586884165107734/784482317699121203) (edit by @shcheklein)
https://github.com/iterative/dvc/issues/5045 - another user asking about this

I'm keen to implement this and put up a PR.

One gotcha straight off the bat:

- Testing will require live cloud credentials as I cant really find a local emulator for this kind of auth. After speaking with some contributors it seems like live credentials exist for AWS, but not necessarily Azure. Would this be something the project is willing to add?

JordanSimba:
Add support for service principal and default credentials in config and from env vars.
Corresponding issue [here](https://github.com/iterative/dvc/issues/4635)

Fixes #4635


efiop (commented):
Hi @JordanSimba !

Thanks for the PR! :pray: 

Could you please rebase and check the failing tests?

dvc/exceptions.py
```diff
@@ -338,3 +337,9 @@ def __init__(self, target, file):
 
 class MergeError(DvcException):
     pass
+
+
+class AuthInfoMissingError(DvcException):
+    """
+    Raised when not enough information is provided for auth against remote.
+    """
```
efiop, line 345: Let's keep it in `dvc/tree/azure.py` for now, since it is only used there. I see that you've probably added it here for a potential to use it in the future in other remotes, which is good thinking :ok_hand: , but let's just keep it local for now. :slightly_smiling_face:

dvc/tree/azure.py
```diff
@@ -36,6 +38,59 @@ def __init__(self, repo, config):
             container = self._az_config.get("storage", "container_name", None)
             self.path_info = self.PATH_CLS(f"azure://{container}")
 
+        self._get_string_token_credential(config)
+
+        if not (self._conn_str or self._credential):
+            # from azure.identity import DefaultAzureCredential
```
efiop, line 44: Leftover?

dvc/tree/azure.py
```diff
@@ -36,6 +38,59 @@ def __init__(self, repo, config):
...
+                logging.warning(
+                    "You must enter an account name for this auth flow"
+                )
+
+            if config.get("azcli_credential"):
+                self._credential = ChainedTokenCredential(AzureCliCredential())
+            else:
+                self._credential = self._create_client_secret_credential(
+                    config
+                )
```
efiop, line 62: Btw, this whole logic is coming from azcli, right? If you've used specific azcli sources, please consider putting a link here as a reference for the future. :slightly_smiling_face:

efiop (commented):
Overall this looks great :fire: ! Just a few minor comments to address.

dvc/exceptions.py
JordanSimba: So my approach of throwing an exception if insufficient info was present was causing an issue with the `dvc\tests\unit\dependency\test_azure.py` test due to the `TestLocalDependency` superclass to `TestAzureDependency`.

 `dvc\tests\unit\dependency\test_azure.py` -> test_save_missing()` was breaking. Drilling down, I noticed that an empty config object was being passed into the tree. 
I saw no way of injecting a appropriate config object, so instead, rather than throwing the exception, I simply write an error log with some message!

dvc/tree/azure.py
JordanSimba: Good spot, deleted now!

dvc/exceptions.py
JordanSimba: So my approach of throwing an exception if insufficient info was present was causing an issue with the dvc\tests\unit\dependency\test_azure.py test due to the TestLocalDependency superclass to TestAzureDependency.

dvc\tests\unit\dependency\test_azure.py -> test_save_missing()` was breaking. Drilling down, I noticed that an empty config object was being passed into the tree.
I saw no way of injecting a appropriate config object, so instead, rather than throwing the exception, I simply write an error log with some message!

dvc/tree/azure.py
JordanSimba: Good spot, deleted now!

dvc/tree/azure.py
JordanSimba: Added a link to a page that would explain the library I've brought in.

JordanSimba:
Linting keeps failing with these warnings
```
************* Module tests.unit.tree.test_azure
tests/unit/tree/test_azure.py:2:0: E0611: No name 'identity' in module 'azure' (no-name-in-module)
************* Module dvc.tree.azure
dvc/tree/azure.py:43:12: E0611: No name 'identity' in module 'azure' (no-name-in-module)
dvc/tree/azure.py:63:8: E0611: No name 'identity' in module 'azure' (no-name-in-module)
```
But all tests pass which kind of implies indeed the imported modules exist.

efiop:
@JordanSimba Please check the formatting, looks like you need to hit your code with `black` one more time.

dvc/exceptions.py
```diff
@@ -1,4 +1,3 @@
-"""Exceptions raised by the dvc."""
```
efiop, line 1: Unrelated change, let's not do this

dvc/tree/azure.py
```diff
@@ -36,9 +37,62 @@ def __init__(self, repo, config):
...
+            tenant_id=tenant_id,
+        )
+
+    def _get_string_token_credential(self, config):
+        # Get `token` like credential to pass to
+        # blob service client. String like
+        # conn_str, sas_token, account_key
+        self._conn_str = config.get("connection_string") or self._az_config.get(
+            "storage", "connection_string", None
+        )
```
efiop, line 95: Not worth a separate method, just inline it

dvc/tree/azure.py
```diff
@@ -36,9 +37,62 @@ def __init__(self, repo, config):
...
+                AzureCliCredential,
+                ChainedTokenCredential,
+            )
+
+            account_name = config.get("account_name")
+            self._account_url = f"https://{account_name}.blob.core.windows.net"
+            if not account_name:
+                logging.warning(
+                    "You must enter an account name for this auth flow"
+                )
```
efiop, line 53: Should this be an error with a raised exception?

dvc/tree/azure.py
```diff
@@ -36,9 +37,62 @@ def __init__(self, repo, config):
...
+
+        tenant_id = config.get("tenant_id") or self._az_config.get(
+            "tenant", "id", None
+        )
+
+        if not (client_id and client_secret and tenant_id):
+            logging.error(
+                "Not enough information to authenticate to azure remote."
+            )
+            return None
```
efiop, line 81: Should this be an exception or are we able to work without the cred?

dvc/tree/azure.py
```diff
@@ -36,9 +37,62 @@ def __init__(self, repo, config):
...
+        )
+
+        tenant_id = config.get("tenant_id") or self._az_config.get(
+            "tenant", "id", None
+        )
+
+        if not (client_id and client_secret and tenant_id):
+            logging.error(
+                "Not enough information to authenticate to azure remote."
+            )
```
efiop, line 80: Should this really be an error? Seems like we are able to work without it just fine.

dvc/exceptions.py
JordanSimba: Dunnon how this, happened. I've put it back

dvc/tree/azure.py
JordanSimba: The `__init__` method is getting way too crowded, it was a little hard to read. That's why I decided to make a the method `_get_string_token_credential` to hold this bit of logic.

dvc/tree/azure.py
JordanSimba: Raising an exception broke a number of tests, one of which was `dvc\tests\unit\dependency\test_azure.py` because somewhere in the execution, a config object is created, one which I can't get a handle on, so the exception is raised and the test fails... this way is more in keeping with the previous behaviour where the user will just fail to authenticate against the remote and there will be logging if they wish to find more information about why

dvc/tree/azure.py
JordanSimba: Its more for information if the user is having troubles authenticating against the remote, if they put insufficient env vars or variables in the config.
Before hand if the user didn't put their token then all that would happen is that they would fail to authenticate against the remote when they tried to push.

JordanSimba:
Documentation PR at https://github.com/iterative/dvc.org/pull/1938

dvc/tree/azure.py
```diff
@@ -29,21 +30,77 @@ class AzureTree(BaseTree):
...
+
+        if not (self._conn_str or self._credential):
+            from azure.identity import (
+                AzureCliCredential,
+                ChainedTokenCredential,
+            )
+
+            if not self._storage_account:
+                logging.warning(
+                    "You must enter a storgae account name for this auth flow"
```
shcheklein, line 51: storgae -> storage 

(could you please open it in one of the IDEs that could check for typos?)

dvc/tree/azure.py
```diff
@@ -29,21 +30,77 @@ class AzureTree(BaseTree):
...
 
+        self._account_url = None
         url = config.get("url", "azure://")
         self.path_info = self.PATH_CLS(url)
 
         if not self.path_info.bucket:
             container = self._az_config.get("storage", "container_name", None)
             self.path_info = self.PATH_CLS(f"azure://{container}")
 
+        self._get_string_token_credential(config)
```
shcheklein, line 41: minor: getters (usually) do not have side effects. Rename to `init_string_...`. Should improve readability. Alternatively (may be even better) return values and assign them here as a tuple.

dvc/tree/azure.py
```diff
@@ -29,21 +30,77 @@ class AzureTree(BaseTree):
...
+
+    def _get_string_token_credential(self, config):
+        # Get `token` like credential to pass to
+        # blob service client. String like
+        # conn_str, sas_token, account_key
         self._conn_str = config.get(
             "connection_string"
         ) or self._az_config.get("storage", "connection_string", None)
 
         self._account_url = None
```
shcheklein, line 96: not needed here anymore?

dvc/tree/azure.py
```diff
@@ -29,21 +30,77 @@ class AzureTree(BaseTree):
...
+        self._get_string_token_credential(config)
+
+        if not (self._conn_str or self._credential):
+            from azure.identity import (
+                AzureCliCredential,
+                ChainedTokenCredential,
+            )
+
+            if not self._storage_account:
+                logging.warning(
```
shcheklein, line 50: if we must enter it, let's raise an exception here?

dvc/tree/azure.py
shcheklein: Essentially, when I read this, I have to jump and parse the whole `_get_string_token_credential` to understand which variables it sets inside. Makes twice harder to review and read the code.

dvc/tree/azure.py
shcheklein: It should be `logging.warning` then to my mind. Error usually means that we should stop.

dvc/tree/azure.py
```diff
@@ -20,6 +20,7 @@ class AzureTree(BaseTree):
     PATH_CLS = CloudURLInfo
     REQUIRES = {
         "azure-storage-blob": "azure.storage.blob",
+        "azure-identity": "azure.identity",
```
shcheklein, line 23: what are the dependencies of this sub-package? we had some problems before with azure-core. We need to check that this one doesn't pull a lot of stuff with it.

dvc/tree/azure.py
```diff
@@ -29,21 +30,77 @@ class AzureTree(BaseTree):
     def __init__(self, repo, config):
         super().__init__(repo, config)
 
+        self._account_url = None
         url = config.get("url", "azure://")
         self.path_info = self.PATH_CLS(url)
 
         if not self.path_info.bucket:
             container = self._az_config.get("storage", "container_name", None)
             self.path_info = self.PATH_CLS(f"azure://{container}")
```
shcheklein, line 39: what happens if container is None here? should we throw an exception?

dvc/tree/azure.py
```diff
@@ -29,21 +30,77 @@ class AzureTree(BaseTree):
...
+            )
+
+            if not self._storage_account:
+                logging.warning(
+                    "You must enter a storgae account name for this auth flow"
+                )
+
+            # Microsoft azure docs
+            # https://docs.microsoft.com/en-us/python/api/overview/azure/identity-readme?view=azure-python
+            if config.get("azcli_credential"):
```
shcheklein, line 56: instead of creating our own variable to differentiate the flow, can we use `DefaultAzureCredential` (may be disable some flows? hard to tell) as described in the link above? It can handle a lot of different cases.

the only exception that we will have is when we see that our config defines one of the variable like `client_id` or something- then we can first redefined ClientSecretCredential

JordanSimba: My thought on this was that users should be able to put auth variables in both the config and env vars and this freedom leads to situations where some info may be missing. In such a case, we should notify the user about missing info... If we just used `DefaultAzureCredential` it would be harder to relay what the possible issue is. But you are right, we could use `DefaultAzureCredential` but at the cost of more control in error handling etc ...

shcheklein: I mean that we can still allow defining our own client_id, etc if that's what you are referring to? (correct me if I didn't get the idea). We can just do a similar thing to the AWS, and even `DefaultAzureCredential` - it can detect all the available information (including our config) and decide which auth flow to use, depending on the fields present.

E.g. we don't detect any custom flow related fields in the DVC config, only `url`. We could use `DefaultCredentialsFlow` (can it handle AD, btw?)

if we detect at least one of the client_id, we can decide to do what you do with create custom Credential object (btw, does it handle env vars?).

The point here is- it would be great to use our config as a way to override certain fields (like env vars are being used for the same purpose in the Azure's lib). But otherwise rely on the Azure's logic to determine which one to use?

(I'm not 100% about all of this ^^, I've just started looking into this. So need your clarifications and help, if already gave it a thought)

dvc/tree/azure.py
JordanSimba: Yup. Removed

dvc/tree/azure.py
JordanSimba: From https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/identity/azure-identity/setup.py
```
    install_requires=[
        "azure-core<2.0.0,>=1.0.0",
        "cryptography>=2.1.4",
        "msal<2.0.0,>=1.6.0",
        "msal-extensions~=0.3.0",
        "six>=1.6",
    ],
    extras_require={
        ":python_version<'3.0'": ["azure-nspkg"],
        ":python_version<'3.5'": ["typing"],
    }
```

dvc/tree/azure.py
JordanSimba: You're right! Changed

dvc/tree/azure.py
JordanSimba: I ran into an issue with throwing exceptions. This is what I've mentions to @efiop in a comment above. Any idea how to get over this?
"
Raising an exception broke a number of tests, one of which was dvc\tests\unit\dependency\test_azure.py because somewhere in the execution, a config object is created, one which I can't get a handle on, so the exception is raised and the test fails... this way is more in keeping with the previous behaviour where the user will just fail to authenticate against the remote and there will be logging if they wish to find more information about why
"

dvc/tree/azure.py
JordanSimba: Yeah, just added a spell check, good spot!

dvc/tree/azure.py
shcheklein: Yep, thanks! should be fine I think .... @efiop could you confirm?

dvc/tree/azure.py
JordanSimba: This is also a good point. I've chosen the tuple approach and moved the side effects out, I don't think the side effects should have been there in the first place!

dvc/tree/azure.py
JordanSimba: I'm not entirely sure, I'll try have to try it out because I can't tell exactly from the code.

dvc/tree/azure.py
JordanSimba: I'm not sure I get what you mean by this line

'DefaultAzureCredential - it can detect all the available information (including our config) and decide which auth flow to use, depending on the fields present.' 

Could you write some psuedo code just so I can get a feeling of what you mean?

shcheklein: Corrected it a bit `it` was not meant to apply to the `DefaultAzureCredential`. Sorry for the confusion.

I'm not 100% familiar with Azure, but I meant something like this:

```python
cred = DefaultAzureCredential(secret_id=conf.secret_id, tenant_id=conf.tenant_id....)
BlobServiceClient(account_url, credential=cred)
```

if `DefaultAzureCredential` doesn't allow overriding values like this, chain it with one of the existing that allow.

Goal here is to let the regular Azure app logic apply. And let our users overwrite some values with config, similar to how would they do with env variables.

I thinks that's how our AWS, GCP remotes work (I don't remember them having any options to enable any specific flow via config).

Again, take this with a grain of salt, since I'm not Azure and especially Azure AD expert. I you see that I'm missing something, could you in turn clarify the logic similar to this https://docs.microsoft.com/en-us/python/api/overview/azure/identity-readme?view=azure-python#defaultazurecredential . What applies first, second, etc ... it would greatly help to review this.

JordanSimba: Apologies for the delay in the response.
The `DefaultAzureCredeneital` itself is quite a high level piece, only orchestrating the auth flow, and the of the auth methods it supports, none are extensible to our use case.

The `DefaultAzureCredential` does the following strategy

> `Envrionment -> ManagedIdentity -> VisualStudioCode -> AzureCLI -> Interactive`

Reading over the constructors for the classes corresponding

shcheklein: > The purpose of each of the auth flow strategies is to produce a credential (some object that can be used to authenticate with), and the DefaultAzureCredential is just a big 'switchboard' and whatever you select the output should be a fully qualified credential

this is exactly what we need to my mind! The lib is handling everything that you would need. Best case you don't have to run `dvc remote modify` at all - you rely on all the regular ways - Environment, ManagedIdentity, etc, etc.

Check this function for S3: https://github.com/iterative/dvc/blob/e1b82c5222930c55886ca16a48c3a223d05b4af0/dvc/tree/s3.py#L64

we pass only values that we have, everything else is handled by some `boto` default logic inside (it takes missing values from env, then from AWS config file, etc, etc). 

-------------

But I see that Azure is different indeed - https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/identity/azure-identity/azure/identity/_credentials/environment.py

It always checks existence of all the fields. 

It still doesn't mean to my mind that we can't or should not rely on the `DefaultAzureCredential`, we can chain it with our own `ConfigCredential`. `ConfigCredential` can behave exactly like `EnvironmentCredential` - check the same variable combinations. If neither of them properly defined -  raise an exception. Otherwise run `DefaultAzureCredential` (probably if none of the AZURE specific fields exist in the DVC remote config?).

In that case there are two questions:

1. Still not clear why do we need an additional `azcli_credential`?
2. Check that CredentialUnavailableError is raised via auze cli if I define only a single env var. To make sure that's it's all or nothing (it's clear that it doesn't override, but it's not clear if it ignores incomplete configs).

JordanSimba: The penny had dropped. 

I understand what you mean... I can just take the logic I've implemented and put it in a class and add this to the chain.
Roger that! If this is correct let me know and I'll just jiggle some code around and update the PR.

Also, in regards to your two questions at the end:

1.  I've chosen to be verbose, so one can look at the config and know what auth flow was chosen, this was just my preference over basically automagic auth.

2. For the Azure CLI Auth flow, you shouldn't need anything other than to have authenticated your az cli. But looking at some of the Auth strategies, whether the CredentialUnavailableError is all or nothing depends on the strategy

shcheklein: > I understand what you mean... I can just take the logic I've implemented and put it in a class and add this to the chain.

yes, it can be a class if chaining is easy. The most important here is to behave similar to how Azure behaves with env credentials, cli, etc and try to reuse as much as possible of the logic they provide

> I've chosen to be verbose, so one can look at the config and know what auth flow was chosen, this was just my preference over basically automagic auth.

yep, I understand. My concern here is that it's not very well aligned with the Azure's style, with other DVC remotes. Also, it's not scalable - since Azure provides so many different ways to auth, we'll have to add a lot of other flags.

> But looking at some of the Auth strategies, whether the CredentialUnavailableError is all or nothing depends on the strategy

hmm, that's interesting ... could you give an example where it's not "all"?

also, the question was more about how does it swallow the `CredentialUnavailableError` in some cases and try next method, or it raises it to the client right away

shcheklein: > If this is correct let me know and I'll just jiggle some code around and update the PR.

thanks 🙏

JordanSimba: What I meant is that the author of an with strategy defines what data must be present and what can be omitted, but there must be some notion of "all" which is all data required to make a valid credential.

Here, the chained token credential file shows how the exceptions are handled if I'm not mistaken.
https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/identity/azure-identity/azure/identity/_credentials/chained.py

From line 67 on.

Apologies, I'm on my phone so no nicely formatted responses

shcheklein: > Apologies, I'm on my phone so no nicely formatted responses

no worries at all

> From line 67 on.

yep, that's exactly what I was asking about! I would probably prefer the fail fast strategy, but it's fine as we have it. Can we enable Azure's logs when we run DVC in verbose mode `dvc something -v`? So, that users (and us debugging) see the sequence of credentials it attempts, failures, etc. @efiop do we have something similar for other libs, a mechanism to enable logs? (may be `-vvv` ?)

at least we should log a warning for the class we'll introduce ConfigCredentials or something.

----------

a separate question - how all this auth relates to the current `connectionString` one?

efiop: We don't have a special mechanism for it, but one could check the loglevel for `dvc` and tweak the lib loglevel accordingly

JordanSimba: @shcheklein I'm too sure what you mean about the logging.

---

Something like this 

```
class ServicePrincipalConfigCredential:
    def __init__(self, config):
        self._credential = None
        self._credential = self._create_client_secret_cred(config)

    def _create_client_secret_cred(self, config):
        client_id = config.get("client_id") 
        client_secret = config.get("client_secret") 
        tenant_id = config.get("tenant_id") 

        if not (client_id and client_secret and tenant_id):
            return None

        return ClientSecretCredential(
            client_id=client_id,
            client_secret=client_secret,
            tenant_id=tenant_id,
        )

    def get_token(self): 
        if not self._credential:
            message = (
                "Dvc authentication unavailable. "
                "Some service principle auth data missing from config."
            )
            raise CredentialUnavailableError(message=message)
        return self._credential

```

There isn't a way to marry the string tokens into a similar auth strategy to put into the `ChainedTokenCredential`, so I've ended up with something like this... Not pushed  yet 
```
...
        self._credential = (
            config.get("sas_token") or 
            self._az_config.get("storage", "key", None) or 
            self._az_config.get("storage", "sas_token", None) or
            ChainedTokenCredential(
                ServicePrincipalConfigCredential(config),
                EnvironmentCredential(),
                AzureCliCredential()
            )
        )
```

ConnectionString string is quite different, and I'm unable to handle it in the same way as these other ways of authenticating.

shcheklein: This looks good and close what I had in mind!! Thanks! A few questions:


`ConnectionString string is quite different`

yes, I think it's fine - it's just a matter of prioritizing, we should have a clear algorithm here. When does it expect/try a connection string, when goes into ChainedCredentials path.

---------

>so I've ended up with something like this... Not pushed yet

so, why not use DefaultCredentials after all? chain them with something

I also don't quite understand `sas_token`, `key` part - what do they mean? or do they come from the connection string?

---------------

>  I'm too sure what you mean about the logging.

if credentials like `EnvironmentCredential` fail for some reason (missing info) I would love to see at least in the logs what's going on here.

JordanSimba: Down the line in the file there is a `blob_service` function/property, this attempts to make a `blob_service_client (bsc)` first by using the `self._conn_str` if it is present, otherwise create a `bsc` with the `self._credential` instead.

```
def blob_service(self):
        if self._conn_str:
            logger.debug(f"Using connection string '{self._conn_str}'")
            blob_service = BlobServiceClient.from_connection_string(
                self._conn_str, credential=self._credential
            )
        else:
            logger.debug(f"Using account url '{self._account_url}'")
            blob_service = BlobServiceClient(
                self._account_url, credential=self._credential
            )
```

In terms of an algorithm, there is a little problem in that the `ChainedTokenCredential` is only checked when some called (in this case the blob_service_client) tries to use it, so hence why I use it last in the code below.
```
self._credential = (
            config.get("sas_token") or 
            config.get("storage_key) or
            self._az_config.get("storage", "key", None) or 
            self._az_config.get("storage", "sas_token", None) or
            ChainedTokenCredential(
                ServicePrincipalConfigCredential(config),
                EnvironmentCredential(),
                AzureCliCredential()
            )
        )
```
If none of the auth strategies are fulfilled when running dvc in verbose mode, I get an exception from the `blob_service_client`

```
...
azure.core.exceptions.ClientAuthenticationError: Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
RequestId:c7aa5197-c01e-0050-0be3-cbe1be000000
Time:2020-12-06T15:25:15.0932609Z
ErrorCode:AuthenticationFailed
...

```

---

- `DefaultCredentials` is not extensible... it's an opinionated offering exposed by the SDK, for a custom auth flow, we have `ChainedTokenCredential` available, and bits like `EnvironmentCredential()` and `AzureCliCredential()` are also available so users of the SDK can make their own auth flow if priorities differ from `DefaultCredentials`.

- In my opinion, using `DefaultCredentials` would make the approach less obvious and less extensible for the chaining.

- `sas_token` and `key` are alternative credentials one can use to create a `blob_service_client`, they were supported before my pull request and rightfully so! The more auth methods the better in terms of features.
SAS_TOKENS are more secure than `storage_keys` and `connection_strings` since they can be scoped more finely and have an expiry.

---

I can't see any way to get at the logs from the azure, hence why in my original implementation I took the responsibility of checking what info was available and logging it.

As it's been in the currently released version of the azure remote, there are no exceptions thrown on construction, the failure comes when trying to use the `blob_service_client`... I think this is an okay approach, since it will tell you about some auth error.

shcheklein: @JordanSimba thanks! I really appreciate all this clarifications. I'm reading the docs now, to make sense of some things. I'll keep it updated. So far I have a few additional questions:

> DefaultCredentials is not extensible

yes, it's expected. I was thinking about this I guess:

```python
self._credential = (
            config.get("sas_token") or 
            config.get("storage_key) or
            self._az_config.get("storage", "key", None) or 
            self._az_config.get("storage", "sas_token", None) or
            ChainedTokenCredential(
                ServicePrincipalConfigCredential(config),
                DefaultCredentials()
            )
        )
```

this way we keep `ManagedIdentify` as well, right? Why can't we go like this? Or at least, why don't we include `ManagedIdentify` and other types of auth?

> `sas_token` and `key` are alternative credentials 

The question I still have is why `self._az_config.get("storage", "sas_token", None)` is needed. I would expect `AzureCliCredential` to handle this? I'll read a bit more about this.

> I can't see any way to get at the logs from the azure, 

you could pass `logging_enable=True`? not sure how easy is that, we need to enable logger for the library, DVC suppresses it by default - would need some help from @efiop and the team here

shcheklein: I guess `AzureCliCredential` from the identity package handles only AAD related stuff, so no token/key type.

One more question though:

https://docs.microsoft.com/en-us/python/api/overview/azure/identity-readme?view=azure-python#environment-variables

describes a few types of Auth - certificate, username, secret - should we let users override all of them?

JordanSimba: You're right I think this would also work. 
If it doesn't I will explicitly add the other auth strategies.
```
ChainedTokenCredential(
                ServicePrincipalConfigCredential(config),
                DefaultCredentials()
            )
```
A little harder for people who aren't familiar with the library to skim, but there is a link to the docs for `DefaultCredentials` in the pyfile.

---
- Yes, I've looked through the identity sdk, and there isn't support for string tokens such as `sas` or `conn_str`. It's a shame.
But I'll keep an eye out... if some solution comes about, I'd be more than happy to unify them in this azure remote!

---
### Questions

- When you say let the users 'override', what exactly do you mean?
- Is `logging_enabled=True` a feature from the dvc side? I can't see anything about it in the azure sdk code.

shcheklein: > When you say let the users 'override', what exactly do you mean?

It's not that important to be honest. What I wanted to say is that we should allow users specify all supported combinations via DVC config, right? (not only `secret`, but `certificate`, `password`, etc)

I mean that when we allow users set `secret_id` for example in the DVC config we are effectively introducing one more layer on top of the auth methods (environment, CLI etc - all that stuff that `DefaultCredentials` provides). Why do I call it "override"? Let's say I usually use regular env vars like `AZURE_TENANT_ID` to pass values to `az` or `dvc`, and now I change my `.dvc/config.local` to specify `tenant_id` - it means that it'll take precedence, right? before the `AZURE_TENANT_ID` value.

> Is logging_enabled=True a feature from the dvc side? I can't see anything about it in the azure sdk code.

it seems to me that it's part of the `DefaultCredentials` itself - you can pass it inside.

JordanSimba: @shcheklein Apologies for leaving this for a while and dredging this up from the past... work got a little busy. 
I think we both needed a refresher...

My rebuttal:
 
- Could not find `logging_enabled` in the API for any of the credentials
- I kinda would prefer if this PR in, since I really only intended to account for traditional service principal auth ... Although the current implementation will support the `secret` `certificate` and `password` as environment variables, the pattern is laid down and anyone who wants to extend this to be supported in the config can do so... thoughts?


Thanks @isidentical for nudging me

.pylintrc
```diff
@@ -293,7 +293,7 @@ ignored-classes=optparse.Values,thread._local,_thread._local,Dvcfile
 # (useful for modules/projects where namespaces are manipulated during runtime
 # and thus existing member attributes cannot be deduced by static analysis. It
 # supports qualified module names, as well as Unix pattern matching.
-ignored-modules=
+ignored-modules=azure
```
shcheklein, line 296: @JordanSimba could you remind me please, why do we need this?

JordanSimba: I kept getting linting failures for Azure identity module which is imported

shcheklein: @skshetry @efiop hey guys, you should know more on these two things ^^ could you help us please?

dvc/tree/azure.py
shcheklein: No worrier @JordanSimba, and it's fair that you didn't ask for all the additional stuff. My main goal when I was reviewing this was: 

- naming and use facing API changes - if we introduce new config vars we should name them properly, we should introduce them only when we need them (thus initial flag to enable CLI caught my attention, thus we changed the logic to use Default mechanism, etc)
- simplicity - we don't have capacity to maintain custom logic, it creates a burden for us, new contributors, etc. So, if we could reuse something - we should be doing this.
- consistency - e.g. if one of the credential mechanisms fails if one field is missing, our custom mechanism should behave the same. Names should be close, etc
- only after that completeness

To some extent, the previous version was in an a pretty bad shape and it was impossible to add new changes that would not make it even worse. I'm really sorry that it happened with your PR and that it takes so long :(

@isidentical I think overall it's in a good shape. The approach makes sense to me. I think it covers pretty much any cases now. I think we can take it over and help @JordanSimba? Some questions that I would try to address/check:

- minor questions left with linters/tests not working, we pinged @efiop and @skshetry 
- tests - check that coverage is enough, review them, I didn't do that
- names for config vars - use `azure_secret_id` for example vs just `secret_id`. As more I think about it the more I realize we should be using prefixes (to make variable names exactly the same as env vars here https://docs.microsoft.com/en-us/python/api/overview/azure/identity-readme?view=azure-python#environment-variables) - cc @efiop. We have a good opportunity here.
- add missing vars from here to our config - https://docs.microsoft.com/en-us/python/api/overview/azure/identity-readme?view=azure-python#environment-variables. Our custom credentials can detect the right combination and initialize one of the existing, provided by the Azure lib.
- loggers - good to have. When auth fails, it's not clear from the user perspective what's going here. E.g. one env var is not set.

JordanSimba: I really appreciate the review @shcheklein, it has been thorough and I've learnt quite a bit more about Azure identity, so it's no problem. I just wish I was abit quicker do we could have reached this point earlier.
I don't mind helping out at all...

On Monday
- I can remove the line in the linter config to ensure there is still an issue.

- I can rename the config variables to match the environment variables (although the environment cars can get a little lengthy)

Leave the existing `connection string` and `sas token` as is for backward compatibility?

- Add the extra classes for `certificate` and `password` in the config. Or do you reckon make the one existing class more complex? Personally I would use separate classes.

Just FYI, With the credentials moved into the `chained credential` any errors will be moved to run time.

shcheklein: Makes sense. Regarding the naming I would not make it a blocker and actually keep it as is. I would wait for @efiop first to confirm and make the decision here, since it affects other remotes. So, if there is not response from him by the time everything else is ready we can merge this I think. My comment was mostly to the team.

> Add the extra classes for certificate and password in the config. Or do you reckon make the one existing class more complex? Personally I would use separate classes.

hard to tell to be honest w/o a bit of a research, I would check how ENV credentials work that Azure lib provides. It should be analyzing the same set of variables to my mind. Again, if no one from the team can address this or you feel it becomes too complicated, we can merge w/o it. And do other stuff on top of it when time comes. This is not a blocker.

> Just FYI, With the credentials moved into the chained credential any errors will be moved to run time.

just to confirm - by run time you mean downstream in the Azure lib code?

JordanSimba: @shcheklein 'by run time you mean downstream in the Azure lib code?' - Yes 

---

Okay, I'll keep an eye out on the PR as usual 👍

.pylintrc
isidentical: `azure.identity` uses the `azure`'s namespace, though apparently pylint fails to detect this. IMHO it is safe to ignore it here rather than keeping `# pylint disable=no-name-in-module` everywhere.

efiop:
@JordanSimba Thank you so much for your PR! We are slowly migrating to fsspec and @isidentical found that adlfs already supports various auth mechanisms, including service principal auth, so we've decided to go with #5463 for now. It will be released later today in 2.0 pre-release (2.0.0a2) and is available from the upstream. Please give it a try and let us know if it works for you. Thank you! :pray:

PR #4673, opened 2020-10-07
TestCmdCacheDir: roll back darwin skip

pared:

Related to #4418

### DIFF
--- a/tests/func/test_cache.py
+++ b/tests/func/test_cache.py
@@ -1,5 +1,4 @@
 import os
-import platform
 import stat
 
 import configobj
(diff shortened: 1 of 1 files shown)


pared:
The fails are related to the fact that in case of `macos` runner, `mktemp()` is creating temporary dir `/var/...`. `/var` is symlink to `/private/var`. 
There seem to be some problem with evaluating the external cache paths, dvc is unable to `makedirs` cache subfolders for a relative path. Sample `error` - ed path that DVC was unable to create: 
`/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/dvc-test.1111.0kupq4di.adgsrJYt9gRMpKwwgUKWM2/.dvc/../../../../../../../../var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/dvc-test.1111.p0h4ct6b.A9R9vcyvc3VLyaVM9oDd5X/ac` 
As we can see, this path contains "symlink version" relative path to external cache.

Wrapping the external cache path in `os.path.realpath` solves the problem.

I tried reproducing the problem on Linux, by creating `/tmp/var` which is a symlink to `/tmp/private/var` and reproducing the test from within `/tmp/var/test_workspace` but I was unable so far. Seems like there is more to that than just symlink being in the path of the repo or external cache. I have tried debugging it via GitHub actions, but was unsuccessful. I am creating an issue to investigate it further.

efiop (approved):

PR #4679, opened 2020-10-08
Parametrization support on dvc.yaml

skshetry:
This:
* does not yet support parametrizing from params (only from provided TEST_DATA). Will be fixed in the upcoming PRs.
* hidden behind a flag for now ( enable via `dvc config feature.parametrization true`). Unlikely to change till we have loops.
* Syntax is same: `${ var }`. Also kept `${{ var }}` syntax for now. Thinking of using double braces, just in case for forward/backward compatibility as users might already be using single braces for bash expansion. 
* Escaping is done by `\${ var}` or `\${{ var }}`, i.e. backslash(`\`).

Part of #3633

### DIFF
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -224,6 +224,8 @@ SCHEMA = {
         "row_limit": All(Coerce(int), Range(1)),
         "row_cleanup_quota": All(Coerce(int), Range(0, 100)),
     },
+    # section for experimental features
+    "feature": {Optional("parametrization", default=False): Bool},
 }
 COMPILED_SCHEMA = Schema(SCHEMA)
 
--- a/dvc/parsing/__init__.py
+++ b/dvc/parsing/__init__.py
@@ -1,9 +1,12 @@
+import logging
 from itertools import starmap
 
 from funcy import join
 
 from .context import Context
 from .interpolate import resolve
+
+logger = logging.getLogger(__name__)
 
 STAGES = "stages"
 
@@ -15,6 +18,7 @@ class DataResolver:
 
     def _resolve_entry(self, name, definition):
         stage_d = resolve(definition, self.context)
+        logger.trace("Resolved stage data for '%s': %s", name, stage_d)
         return {name: stage_d}
 
     def resolve(self):
--- a/dvc/parsing/__init__.py
+++ b/dvc/parsing/__init__.py
@@ -1,0 +1,23 @@
+from itertools import starmap
+
+from funcy import join
+
+from .context import Context
+from .interpolate import resolve
+
+STAGES = "stages"
+
+
+class DataResolver:
+    def __init__(self, d):
+        self.context = Context()
+        self.data = d
+
+    def _resolve_entry(self, name, definition):
+        stage_d = resolve(definition, self.context)
+        return {name: stage_d}
+
+    def resolve(self):
+        stages = self.data.get(STAGES, {})
+        data = join(starmap(self._resolve_entry, stages.items()))
+        return {**self.data, STAGES: data}
--- a/dvc/parsing/context.py
+++ b/dvc/parsing/context.py
@@ -1,0 +1,61 @@
+from collections.abc import Collection, Mapping, Sequence
+
+# for testing purpose
+# FIXME: after implementing of reading of "params".
+TEST_DATA = {
+    "__test__": {
+        "dict": {"one": 1, "two": 2, "three": "three", "four": "4"},
+        "list": [1, 2, 3, 4, 3.14],
+        "set": {1, 2, 3},
+        "tuple": (1, 2),
+        "bool": True,
+        "none": None,
+        "float": 3.14,
+        "nomnom": 1000,
+    }
+}
+
+
+class Context:
+    def __init__(self, data=None):
+        self.data = data or TEST_DATA
+
+    def select(self, key):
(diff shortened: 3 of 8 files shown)


tall-josh:
Hot off the press!!!

tall-josh:
Hey @skshetry can you provide a sample `dvc.yaml` to play with this?

I've tried:

```yaml
args | params | vars:
  ...
stages:
  ...
```

and various combinations from [your investigation in 3633](https://github.com/iterative/dvc/issues/3633#issuecomment-697637518)

skshetry:
@tall-josh, sorry, this does not yet support that. Right now, you can only use this test data for interpolation:
https://github.com/iterative/dvc/blob/9591cb4bf86005fe1cbe90823ec834463f3a243c/dvc/parsing/context.py#L5

```yaml
stages:
  stage1:
    cmd: echo ${__test__.dict.one}
````
As you see, this is not useful in any way. That "args" will come in upcoming PRs, as we are still discussing about few things including:
1. What to call that? "args" or anything else?
2. Should we allow multiple imports? If yes, should we merge them or namespace them?
3. Should the imports be explicit or be implicitly default to "params.yaml"?

**Comment by tall-josh** [comment]
Ahhh all good thanks for putting in the hard-yards @skshetry 

My two cents is:

  1. `args | vars | params | vals | ...` I dont have a strong preference for.
  2. Personally I think multiple params files could get cumbersome/confusing.
  3. I prefer explicit, I've never been a fan of secret sauce.

**Comment by skshetry** [comment]
`tests/func/test_api.py::test_open[azure]` failure is a known issue, needs investigation why it's flakey (see N1 on #4653).

**Review by pmrowla — APPROVED** [review]
Nice start, LGTM.

Also just noting that I'd prefer going with the double-brace syntax so that it is more explicitly separate from bash expansion

File: dvc/config.py [hunk 1]
```diff
@@ -224,6 +224,8 @@ class RelPath(str):
         "row_limit": All(Coerce(int), Range(1)),
         "row_cleanup_quota": All(Coerce(int), Range(0, 100)),
     },
+    # section for experimental features
+    "feature": {Optional("parametrization", default=False): Bool},
```
efiop, line 228: Maybe `experimental` after all? Just to emphasize that this is for experimental features. `feature` could potentially be used for non-experimental but opt-in features.

dvc/parsing/__init__.py
```diff
@@ -0,0 +1,27 @@
+import logging
```
efiop, line 1: About the general file structure. Maybe let's move `dvc/dvcfile.py` -> `dvc/dvcfile/__init__.py` and `dvc/parsing` -> `dvc/dvcfile/parsing`, since parsing is only for dvcfiles.

dvc/parsing/context.py
```diff
@@ -0,0 +1,61 @@
+from collections.abc import Collection, Mapping, Sequence
+
+# for testing purpose
+# FIXME: after implementing of reading of "params".
+TEST_DATA = {
```
efiop, line 5: A bit weird having it here, why not just have it in the tests themselves?

dvc/config.py
skshetry: Is there any difference between the opt-in feature and experimental feature? In theory, yes, but does not seem to be any different in practice.

dvc/parsing/__init__.py
skshetry: Yeah, I guess, but I am hesitant to make that decision yet. Everything from this PR is just expendables. I'm sure, it will find a better home.

skshetry: And, hesitant to keep that in dvcfile not yet, as most of the logic might need to be moved to "StageLoader" which in turn has to be moved to dvcfile in the future.

dvc/parsing/context.py
skshetry: For me to quickly test/qa. Nothing to do with actual test. 🙂

efiop: @skshetry Still don't understand why we need to keep it in the main codebase.

dvc/parsing/__init__.py
efiop: @skshetry Ok, your call.

dvc/config.py
efiop: @skshetry There is a difference in how people will understand it. At least in my view, `feature` is expected to contain well-off stable or almost stable opt-in functionality, while `experimental`(aka `experimental feature`) has the unstable nature of the section in the name itself, so you really know what you are getting into.

skshetry: In that case, "opt-in" features are those that are documented and experimental features are not documented.

dvc/parsing/context.py
skshetry: Could you please suggest where to keep these? These are not tests, so they are not in "tests" folder, it's only for me to play with. 

It will be most likely be gone in the next PR.

efiop: @skshetry I guess it is just not clear to why are they there in the first place. If you think you need it for now and you will get rid of it in the next iteration, then please feel free to keep it, I'm fine with it.

efiop (approved):

efiop:
@skshetry Please feel free to merge when you feel like it.

PR #4680, opened 2020-10-08
tests build: buil on push/pr to master

pared:

Currently each time anyone pushes anything to any fork of DVC, the test build is triggered. I think we should not allow that. Since the only builds that are interesting to us are the ones for master commits/pull requests.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -1,9 +1,16 @@
 name: Tests build
 
-on: [push, pull_request]
+on:
+  pull_request:
+    branches:
+      - master
+  push:
+    branches:
+      - master
 
 jobs:
   lint:
+    if: ${{ github.repository }} == 'iteratie/dvc'
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2.3.2
@@ -18,6 +25,7 @@ jobs:
     - name: Check formatting
       run: pre-commit run --all-files
   tests:
+    if: ${{ github.repository }} == 'iteratie/dvc'
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false


efiop:
> Currently each time anyone pushes anything to any fork of DVC, the test build is triggered. I think we should not allow that. Since the only builds that are interesting to us are the ones for master commits/pull requests.

Are fork builds visible in our actions or only in fork's actions?

Currently, you only allow `master` to be built with our GHA, which is not the best idea, as it prevents people from using dev branches (e.g. we sometimes use those to backport something to an older release and release an old version e.g. 0.94). If fork's builds are accounted as our main repo builds then maybe you really meant something like `github.repository == 'iterative/dvc'`

pared:
> Are fork builds visible in our actions or only in fork's actions?

Yes, fork builds are not visible in our repo.

> as it prevents people from using dev branches

It does not, user can just add the branch to "branches" and build will be done on this particular branch too.

> `github.repository == 'iterative/dvc'`

This is also a good idea,
I think we should combine both conditions - in current state builds should be happening on PR's and pushes to master, only in our repository, expanding permissions to any other situation is a bit wasteful in my opinion.

efiop (approved):

.github/workflows/tests.yaml
```diff
@@ -1,9 +1,16 @@
...
+  pull_request:
+    branches:
+      - master
+  push:
+    branches:
+      - master
 
 jobs:
   lint:
+    if: ${{ github.repository }} == 'iteratie/dvc'
```
skshetry, line 13: I don't quite understand why this change was necessary, as the workflow for forks run on their own repository.
There is a spelling error too. `s/iteratie/iterative`.

pared: So, since we promote PR's from users repositories, it is just wasteful to run full builds. Typo is a good point, and since the recent builds are still building, it seems like its not respected. Will get rid of that.

Issue #3642, opened 2020-04-17
Consistent cmd option order?


jorgeorpinel:
> Extracted from https://github.com/iterative/dvc/issues/3608#issuecomment-615008799

To match docs, where options have been ordered manually from more to less important (at least that's the idea).

Example difference:

```
λ dvc commit -h
usage: dvc commit [-h] [-q | -v] [-f] [-d] [-R] [targets [targets ...]]
...

optional arguments:
  -h, --help            show this help message and exit
  -q, --quiet           Be quiet.
  -v, --verbose         Be verbose.
  -f, --force      Commit even if hash value for dependencies/outputs changed.
  -d, --with-deps  Commit all dependencies of the specified target.
  -R, --recursive  Commit cache for subdirectories of the specified directory.
```
(DVC 0.93.0)

vs. https://dvc.org/doc/command-reference/commit#options:

![image](https://user-images.githubusercontent.com/1477535/79532553-1e6b5d80-803b-11ea-84a3-87c93af4eb5e.png)

jorgeorpinel:
I'm wondering whether -h -q -v can be moved to the bottom of all cmds easily as a first step..

efiop:
@jorgeorpinel They probably could be moved with some hack or something, but I would urge you to consider if this is really worth it.

efiop:
@jorgeorpinel I guess worth elaborating that moving -h/-v would require a non-trivial hack for argparse, which is probably too much effort for this kind of change. Not as easy as moving few lines around, unfortunately :slightly_frowning_face:

jorgeorpinel:
Yes, -h -q -v def. don't seem worth trying to move. I'm not sure about the other ones though, let's say you have 20 lines in your terminal vertically (most people will have more but still) you use -h on a big command like run and scroll back to the top of the output. Shouldn't the more commonly used options be shown first? It could potentially save time and make the experience a little better.

It's definitely marginal so I'm not obsessing over this or anything. We can just leave it open and if we never get to it just close it at some point. Or close it if you feel strongly against this, np on my side 🙂

efiop:
@jorgeorpinel Those are really good points! If we look at it that way, maybe it is worth hacking it at some point. Also, interesting to note that things like `awscli` don't even use argparse and use their own implementation to be able to polish the experience. I don't have a strong opinion.

efiop:
[...]

PR #4681, opened 2020-10-08
dvc: update order of -q -v -h args

naspeh:
Related to #3642

This is one possible solution for ordering these arguments. If this solution is acceptable, then other subcommands should be updated in a similar way.

### See `-q` `-v` and `-h` at the end:
```
$ dvc commit -h   
usage: dvc commit [-f] [-d] [-R] [-q | -v] [-h] [targets [targets ...]]

Save changed data to cache and update DVC-files.
Documentation: <https://man.dvc.org/commit>

positional arguments:
  targets          DVC-files to commit. Optional. (Finds all DVC-files in the workspace by default.)

optional arguments:
  -f, --force      Commit even if hash value for dependencies/outputs changed.
  -d, --with-deps  Commit all dependencies of the specified target.
  -R, --recursive  Commit cache for subdirectories of the specified directory.
  -q, --quiet      Be quiet.
  -v, --verbose    Be verbose.
  -h, --help       Show this help message and exit.

$ dvc init -h  
usage: dvc init [--no-scm] [-f] [--subdir] [-q | -v] [-h]

Initialize DVC in the current directory. Expects directory
to be a Git repository unless --no-scm option is specified.
Documentation: <https://man.dvc.org/init>

optional arguments:
  --no-scm       Initiate DVC in directory that is not tracked by any SCM tool (e.g. Git).
  -f, --force    Overwrite existing '.dvc/' directory. This operation removes local cache.
  --subdir       Necessary for running this command inside a subdirectory of a parent SCM repository.
  -q, --quiet    Be quiet.
  -v, --verbose  Be verbose.
  -h, --help     Show this help message and exit.

$ dvc -h       
usage: dvc [-V] [--cd <path>] [-q | -v] [-h] COMMAND ...

Data Version Control

optional arguments:
  -V, --version  Show program's version.
  --cd <path>    Change to directory before executing.
  -q, --quiet    Be quiet.
  -v, --verbose  Be verbose.
  -h, --help     Show this help message and exit.
...
```

### DIFF
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -129,8 +129,11 @@ class VersionAction(argparse.Action):  # pragma: no cover
         sys.exit(0)
 
 
-def add_common_args(parser):
+def add_common_args(parser, *, func=None):
     """Add common arguments shared among all the commands."""
+    if func:
+        parser.set_defaults(func=func)
+
     parser.add_argument(
         "--cprofile",
         action="store_true",
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -129,19 +129,7 @@ class VersionAction(argparse.Action):  # pragma: no cover
         sys.exit(0)
 
 
-def get_parent_parser():
-    """Create instances of a parser containing common arguments shared among
-    all the commands.
-
-    When overwriting `-q` or `-v`, you need to instantiate a new object
-    in order to prevent some weird behavior.
-    """
-    parent_parser = argparse.ArgumentParser(add_help=False)
-    add_common_args(parent_parser)
-    return parent_parser
-
-
-def add_common_args(parser, add_help=False):
+def add_common_args(parser):
     """Add common arguments shared among all the commands."""
     parser.add_argument(
         "--cprofile",
@@ -159,15 +147,15 @@ def add_common_args(parser, add_help=False):
         "-v", "--verbose", action="count", default=0, help="Be verbose."
     )
 
-    if add_help:
-        # NOTE: We are doing this to capitalize help message.
-        parser.add_argument(
-            "-h",
-            "--help",
-            action="help",
-            default=argparse.SUPPRESS,
-            help="Show this help message and exit.",
-        )
+    # NOTE: We are doing this to capitalize help message and move it to the
+    # bottom of all agruments.
+    parser.add_argument(
+        "-h",
+        "--help",
+        action="help",
+        default=argparse.SUPPRESS,
+        help="Show this help message and exit.",
+    )
 
 
 def get_main_parser():
@@ -198,7 +186,7 @@ def get_main_parser():
         type=str,
     )
 
-    add_common_args(parser, add_help=True)
+    add_common_args(parser)
 
     # Sub commands
     subparsers = parser.add_subparsers(
@@ -210,12 +198,8 @@ def get_main_parser():
 
     fix_subparsers(subparsers)
 
-    parent_parser = get_parent_parser()
     for cmd in COMMANDS:
-        subparser = cmd.add_parser(subparsers, parent_parser)
-        if isinstance(subparser, argparse.ArgumentParser):
-            add_common_args(subparser, add_help=True)
-
+        cmd.add_parser(subparsers, add_common_args)
     return parser
 
 
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -137,16 +137,21 @@ def get_parent_parser():
     in order to prevent some weird behavior.
     """
     parent_parser = argparse.ArgumentParser(add_help=False)
+    add_common_args(parent_parser)
+    return parent_parser
 
-    parent_parser.add_argument(
+
+def add_common_args(parser, add_help=False):
+    """Add common arguments shared among all the commands."""
+    parser.add_argument(
         "--cprofile",
         action="store_true",
         default=False,
         help=argparse.SUPPRESS,
     )
-    parent_parser.add_argument("--cprofile-dump", help=argparse.SUPPRESS)
+    parser.add_argument("--cprofile-dump", help=argparse.SUPPRESS)
 
-    log_level_group = parent_parser.add_mutually_exclusive_group()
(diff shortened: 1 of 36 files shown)


efiop (changes requested):
Hi @naspeh !

Thank you for the PR! Interesting approach!

Unfortunately, it is clear that this requires a lot of effort to convert the whole codebase and complicates the maintenance, as we are straying away from the vanilla argparse workflow.

Considering that the only thing that that ticket complains about is the order of options in the help output, maybe there is another way to achieve it more gracefully? :slightly_smiling_face:

naspeh:
@efiop What do you think about such a complete solution?

I think this is pretty safe refactoring, as if you forget to add `add_help=False,` in subparsers, then you will get an error right away like this:
```
$ git df
diff --git a/dvc/command/get.py b/dvc/command/get.py
index eb654c88..115e7098 100644
--- a/dvc/command/get.py
+++ b/dvc/command/get.py
@@ -55,7 +55,6 @@ def add_parser(subparsers, add_common_args):
     get_parser = subparsers.add_parser(
         "get",
         description=append_doc_link(GET_HELP, "get"),
-        add_help=False,
         help=GET_HELP,
     )

$ dvc experiments -h         
ERROR: unexpected error - argument -h/--help: conflicting option strings: -h, --help
```

:)

dvc/cli.py
```diff
@@ -129,59 +129,48 @@ def __call__(self, parser, namespace, values, option_string=None):
         sys.exit(0)
 
 
-def get_parent_parser():
-    """Create instances of a parser containing common arguments shared among
-    all the commands.
+def add_common_args(parser, *, func=None):
+    """Add common arguments shared among all the commands."""
+    if func:
+        parser.set_defaults(func=func)
```
efiop, line 135: This part is not really related to the common args, so passing `func` to `add_common_args` is confusing, as the name doesn't really reflect the action that well.

An alternative spin on that would be to just call `add_common_args(parser)` and not mess with `set_defaults`, but that will still require us calling that helper in every command. Which seems ok to me if there is no better way to do it.

Considering that all we want to do is make -h/-v appear last in the help section, maybe there are some argparse mechanisms we could leverage? E.g. maybe we could alter the way it formats the help in our DvcParser? It might turn out impossible or too hacky, but just want to consider it as it seems to have some potential to keep things localized and avoid us relying on the order in which we add things in the code.

naspeh: > [Explicit is better than implicit.](https://www.python.org/dev/peps/pep-0020/)

This is an explicit solution. It uses `agrparse` as it was designed. `agrparse` preserves the order of how you add your arguments and show them in that order, so it adds `-h` in init method and handles `parents` here as well:
https://github.com/python/cpython/blob/master/Lib/argparse.py#L1732-L1747
all later calls to `add_argument` will go down the list.

Now I do it in `agrparse` way and add these agrguments at the end, so it can be displayed as we want.

I don't understand why I should think about some magic way of just ordering some arguments differently. Yes, we can try to reorder [_ActionsContainer._actions](https://github.com/python/cpython/blob/fa1d83db62a545580d9a1a585b2c1fb55961a5c3/Lib/argparse.py#L1451), but it would be hacky and magical as you said, I'm not interested in implementing such a solution.

----

I agree on the name is confusing.

> An alternative spin on that would be to just call `add_common_args(parser)` and not mess with `set_defaults`

If you look at the commit history:
 [Move "parser.set_defaults(func=...)" to "add_common_args" as well](https://github.com/iterative/dvc/pull/4681/commits/ad86cbd64a3fa95b8b0099bfa8f59cfb99e2230a)
this move was in the last commit as I have seen repeatable two lines like
```py
    commit_parser.set_defaults(func=CmdCommit)
    add_common_args(commit_parser)
```

>  but that will still require us calling that helper in every command.

We already have assumptions for every command:
- we should call `add_parser` with `parents`
- we should call `parser.set_defaults(func=...)` as well

-----

I see a few options for this PR:
1. I revert my last commit with moving `parser.set_defaults(func=...)` into `add_common_args`
2. I rename `add_common_args` and keep `parser.set_defaults(func=...)` in it, maybe `add_common_parts`, or any better name
3. I just close this PR.

I'm fine with every option.

efiop: Don't get me wrong, I'm all the way for explicit solutions :slightly_smiling_face:

I totally see your points :+1: , but just one more thing: you've probably considered this already, but just wanted to ask if you've considered using argparser's `formatting_class`?

efiop: Btw, one more thing that I also got confused by: the ticket is about formatting in general and dvc.org seems to have a simple alphabetically sorted list of options, which we could do in dvc as well. Though the implementation for both approaches could be very similar. :slightly_smiling_face:

naspeh: @efiop please check my alternate solution here: #4695

efiop: @naspeh Looks pretty good! :slightly_smiling_face: 

So have you considered `formatting_class`? It is an explicit potential solution that is provided by argparse specifically to customize help output.

naspeh: I have checked `formatter_class` option, but couldn't get quick result, we have subparsers-subparsers-subparser and you need to traverse all of these deep nestings to handle it properly. Haven't seen a near solution and come up with another one.

efiop: @naspeh Got it, makes sense :ok_hand: 

It is important for this to be easy to maintain too, so a format_class-based solution seems promising. Seems like we could override `HelpFormatter.add_arguments` and either sort actions or just move -v/-h/-q to the end. I would probably go with sorting since it makes more sense. That way we won't need to override unrelated methods like we do in https://github.com/iterative/dvc/pull/4695 , would keep the original code and would format help where it is expected - in help formatter. What do you think?

naspeh: Closing both my PRs. Up to you.

efiop: @naspeh Thanks for the PRs! :pray:

Issue #1618, opened 2019-02-16
status: "Data and pipelines are up to date." when using --cloud


ghost:
I'm not sure if `Pipeline is up to date. Nothing to reproduce.` is the best message to describe the `status` of a remote storage.

When using `fetch/pull/push` you are not **reproducing** anything, you are **syncing** the cache.

I would suggest something more _ad hoc_ like the following: `Cache is up to date. Nothing to sync with '{}' remote`.

Also, it can be confusing if `dvc status` shows you some "changes" and `dvc status -c` outputs "Nothing to reproduce"

efiop:
Related https://github.com/iterative/dvc/issues/1202

jorgeorpinel:
@mroutis I believe the message has changed at this point, does this issue desc. need an update? Thanks

efiop:
@jorgeorpinel indeed current message is
```
➜  dvc git:(master) ✗ dvc status -c   
Data and pipelines are up to date.
```
But the gist of the issue stays the same. Updated the desc though. Thank you for the heads up! 🙂

dashohoxha:
Maybe the best solution is to rename `dvc status --cloud` to `dvc remote status` or `dvc remote diff`. Then it is going to have a completely different set of messages from `dvc status`.

I think this idea (splitting `dvc status` and `dvc status -c` into two different commands) has been proposed previously, but I couldn't find the relevant issues.

kurianbenoy:
Can I reserve this issue to be used for today's Devsprints?

efiop:
@kurianbenoy Sure!

tenitski:
Looking at this issue 👀

efiop:
Fixed by #4694

PR #4694, opened 2020-10-11
#1618 Custom message when cache and remote are in sync

tenitski:
Addresses  #1618 issue

Update `dvc status -r X` or `dvc status -c` commands by printing out a more specific message if cache and remote are in sync with each other.

### DIFF
--- a/dvc/command/status.py
+++ b/dvc/command/status.py
@@ -69,14 +69,11 @@ class CmdDataStatus(CmdDataBase):
                 self._show(st, indent)
             elif not self.repo.stages:
                 logger.info(self.EMPTY_PROJECT_MSG)
-            elif self.args.cloud:
-                logger.info(
-                    self.IN_SYNC_MSG.format(
-                        remote=self.repo.config["core"].get("remote")
-                    )
+            elif self.args.cloud or self.args.remote:
+                remote = self.args.remote or self.repo.config["core"].get(
+                    "remote"
                 )
-            elif self.args.remote:
-                logger.info(self.IN_SYNC_MSG.format(remote=self.args.remote))
+                logger.info(self.IN_SYNC_MSG.format(remote=remote))
             else:
                 logger.info(self.UP_TO_DATE_MSG)
 
--- a/dvc/command/status.py
+++ b/dvc/command/status.py
@@ -69,12 +69,14 @@ class CmdDataStatus(CmdDataBase):
                 self._show(st, indent)
             elif not self.repo.stages:
                 logger.info(self.EMPTY_PROJECT_MSG)
-            elif self.args.cloud or self.args.remote:
+            elif self.args.cloud:
                 logger.info(
                     self.IN_SYNC_MSG.format(
                         remote=self.repo.config["core"].get("remote")
                     )
                 )
+            elif self.args.remote:
+                logger.info(self.IN_SYNC_MSG.format(remote=self.args.remote))
             else:
                 logger.info(self.UP_TO_DATE_MSG)
 
--- a/dvc/command/status.py
+++ b/dvc/command/status.py
@@ -11,6 +11,7 @@ class CmdDataStatus(CmdDataBase):
     STATUS_LEN = 20
     STATUS_INDENT = "\t"
     UP_TO_DATE_MSG = "Data and pipelines are up to date."
+    IN_SYNC_MSG = "Cache and remote '{remote}' are in sync."
     EMPTY_PROJECT_MSG = (
         "There are no data or pipelines tracked in this project yet.\n"
         "See {link} to get started!"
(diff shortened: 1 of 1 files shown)


dvc/command/status.py
```diff
@@ -11,6 +11,7 @@ class CmdDataStatus(CmdDataBase):
     STATUS_LEN = 20
     STATUS_INDENT = "\t"
     UP_TO_DATE_MSG = "Data and pipelines are up to date."
+    IN_SYNC_MSG = "Cache and remote '{remote}' are in sync."
```
tenitski, line 14: There could be a better phrased message, I'm open to suggestions.

dvc/command/status.py
```diff
@@ -68,6 +69,12 @@ def run(self):
                 self._show(st, indent)
             elif not self.repo.stages:
                 logger.info(self.EMPTY_PROJECT_MSG)
+            elif self.args.cloud or self.args.remote:
+                logger.info(
+                    self.IN_SYNC_MSG.format(
+                        remote=self.repo.config["core"].get("remote")
```
efiop, line 75: Notice that we have `--remote` CLI flag too, and config['core']['remote'] is the default one that might not be set  :slightly_smiling_face:

tenitski: Oh, good point, updated.

dvc/command/status.py
```diff
@@ -68,6 +69,14 @@ def run(self):
...
             elif not self.repo.stages:
                 logger.info(self.EMPTY_PROJECT_MSG)
+            elif self.args.cloud:
+                logger.info(
+                    self.IN_SYNC_MSG.format(
+                        remote=self.repo.config["core"].get("remote")
+                    )
+                )
+            elif self.args.remote:
+                logger.info(self.IN_SYNC_MSG.format(remote=self.args.remote))
```
efiop, line 79: @tenitski Thanks! Looks good! One minor thing: let's keep one branch of execution:
```
            elif self.args.cloud:
                remote = self.args.remote or self.repo.config["core"].get("remote")
                logger.info(self.IN_SYNC_MSG.format(remote=remote))
```
just a bit easier to comprehend that way :slightly_smiling_face:

tenitski: Done. Linter insisted on breaking it into a few more lines though

efiop: Yeah, that's expected, linter sometimes makes it a little ugly :smile: Could refactor to fit into 80 chars or could leave as is, I'm fine with either :ok_hand:

efiop (approved):

dvc/command/status.py
efiop: Hm, `Cache and remote are in sync` sounds a bit like cache and remote have the same list of cache files, which is actually not true, because we care about the cache for the local workspace or branches/tags (--all-tags, --all-branches). But this is still a good start :+1:

efiop:
One more thing, in the PR header there was a checkbox for the docs. This PR will require doc change, so could you please at least create an issue for it on https://github.com/iterative/dvc.org/issues ?

tenitski:
When I created the PR I checked out docs page for `dvc status` but could not see there anything that would become out of date because of this change. Can you point out a specific place please?

dvc/command/status.py
tenitski: I used this paragraph from `dvc status` as a guide for what is happening and how it should be called:
> For new and deleted data, the cache is different from remote storage. Bringing the two into sync requires dvc pull or dvc push. For the typical process to update the workspace, see Sharing Data And Model Files.

tenitski: So if you don't need to do anything to bring them into sync I thought we can call the state as "are in sync" :)

dvc/command/status.py
tenitski: Makes sense, I didn't realise line length was the trigger.

dvc/command/status.py
tenitski: I'm keen to work out the best message cause this PR is all about making the message more accurate. Having an OK message kind of defeats the purpose...

efiop: @tenitski Thanks for the attention to details! :pray: 

@jorgeorpinel could you take a look, please? :pray:

efiop:
@tenitski Oh, good to know :ok_hand: Pretty much https://dvc.org/doc/command-reference/status will need some adjustment in `If no differences are detected, dvc status prints` paragraph and there might be some now incorrect "Data and pipelines are up to date" messages all across the docs (i.e. will have to grep https://github.com/iterative/dvc.org/tree/master/content for it and adjust messages for -c/-r invocation).

tenitski:
> @tenitski Oh, good to know 👌 Pretty much https://dvc.org/doc/command-reference/status will need some adjustment in `If no differences are detected, dvc status prints` paragraph and there might be some now incorrect "Data and pipelines are up to date" messages all across the docs (i.e. will have to grep https://github.com/iterative/dvc.org/tree/master/content for it and adjust messages for -c/-r invocation).

Oh, yeah, makes sense. I didn't think about other pages which may refer to this message.

dvc/command/status.py
jorgeorpinel: Minor: based on the `elif` order below wouldn't it be better to put this new constant before `UP_TO_DATE_MSG`?
As for the text itself, it seems good but mentioning the cache seems too low-level. Maybe "This repo is in sync with the remote." ? Also, may be no need to specify which remote since the user should know that already... But I guess for when you have a default remote, to make sure you're aware of which one was checked, appending the remote name at the end is not a bad idea.

PR #4695, opened 2020-10-11
dvc: update order of -q -v -h args [v2]

naspeh:
An alternative solution for #4681

I highlighted the idea here, if it's more suitable then I need to finalize it, adding `add_help=False` and cleaning `parent_parser` everywhere.

It works as previous PR:
```
$ dvc add -h        
usage: dvc add [-R] [--no-commit] [--external] [--file <filename>] [-q | -v] [-h] targets [targets ...]

Track data files or directories with DVC.
Documentation: <https://man.dvc.org/add>

positional arguments:
  targets            Input files/directories to add.

optional arguments:
  -R, --recursive    Recursively add files under directory targets.
  --no-commit        Don't put files/directories into cache.
  --external         Allow targets that are outside of the DVC repository.
  --file <filename>  Specify name of the DVC-file this command will generate.
  -q, --quiet        Be quiet.
  -v, --verbose      Be verbose.
  -h, --help         Show this help message and exit.

$ dvc init -h       
usage: dvc init [--no-scm] [-f] [--subdir] [-q | -v] [-h]

Initialize DVC in the current directory. Expects directory
to be a Git repository unless --no-scm option is specified.
Documentation: <https://man.dvc.org/init>

optional arguments:
  --no-scm       Initiate DVC in directory that is not tracked by any SCM tool (e.g. Git).
  -f, --force    Overwrite existing '.dvc/' directory. This operation removes local cache.
  --subdir       Necessary for running this command inside a subdirectory of a parent SCM repository.
  -q, --quiet    Be quiet.
  -v, --verbose  Be verbose.
  -h, --help     Show this help message and exit.

$ dvc -h            
usage: dvc [-V] [--cd <path>] [-q | -v] [-h] COMMAND ...

Data Version Control

optional arguments:
  -V, --version  Show program's version.
  --cd <path>    Change to directory before executing.
  -q, --quiet    Be quiet.
  -v, --verbose  Be verbose.
  -h, --help     Show this help message and exit.
```

### DIFF
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -117,6 +117,37 @@ class DvcParser(argparse.ArgumentParser):
             self.error(msg % " ".join(argv), getattr(args, "func", None))
         return args
 
+    def parse_known_args(self, args=None, namespace=None):
+        self.add_common_args()
+
+        # TODO: temp, all subparser should be called with add_help=False
+        if self.prog in ("dvc", "dvc add", "dvc init"):
+            self.add_argument(
+                "-h",
+                "--help",
+                action="help",
+                default=argparse.SUPPRESS,
+                help="Show this help message and exit.",
+            )
+        return super().parse_known_args(args, namespace)
+
+    def add_common_args(self):
+        self.add_argument(
+            "--cprofile",
+            action="store_true",
+            default=False,
+            help=argparse.SUPPRESS,
+        )
+        self.add_argument("--cprofile-dump", help=argparse.SUPPRESS)
+
+        log_level_group = self.add_mutually_exclusive_group()
+        log_level_group.add_argument(
+            "-q", "--quiet", action="count", default=0, help="Be quiet."
+        )
+        log_level_group.add_argument(
+            "-v", "--verbose", action="count", default=0, help="Be verbose."
+        )
+
 
 class VersionAction(argparse.Action):  # pragma: no cover
     # pylint: disable=too-few-public-methods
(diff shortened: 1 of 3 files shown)


dvc/cli.py
```diff
@@ -117,6 +117,37 @@ def parse_args(self, args=None, namespace=None):
             self.error(msg % " ".join(argv), getattr(args, "func", None))
         return args
 
+    def parse_known_args(self, args=None, namespace=None):
+        self.add_common_args()
+
+        # TODO: temp, all subparser should be called with add_help=False
+        if self.prog in ("dvc", "dvc add", "dvc init"):
```
efiop, line 124: Could you explain this line? Are we going to have to add each command to this list?

naspeh: nope, see my `TODO` comment, I haven't finished the implementation. If this solution works, I need go through all `add_parser` methods and remove `parent_parser` and add `add_help=False` (I wrote the same in PR description). After full cleanup, I can remove this block at all - this why it says `temp` :).

dvc/cli.py
```diff
@@ -117,6 +117,37 @@ def parse_args(self, args=None, namespace=None):
             self.error(msg % " ".join(argv), getattr(args, "func", None))
         return args
 
+    def parse_known_args(self, args=None, namespace=None):
+        self.add_common_args()
```
efiop, line 121: Btw, this might mess with `shtab`, as now `-v/-q/etc` are added dynamically when command is being parsed. formatter-based proposed potential solution from https://github.com/iterative/dvc/pull/4681#discussion_r502950934 doesn't seem to suffer from that.

PR #4701, opened 2020-10-12
experiments: support `exp run --checkpoint` for baseline commits

pmrowla:

* Checkpoint runs for a baseline commit with no param/source modifications now works as expected - experiment will be reproduced in a new checkpoint branch
* Checkpoint stage subprocess will be terminated when an error occurs during checkpoint generation

### DIFF
--- a/dvc/stage/run.py
+++ b/dvc/stage/run.py
@@ -157,9 +157,7 @@ def _checkpoint_run(stage, callback_func, done_cond, proc):
                 _run_callback(stage, callback_func)
             except Exception:  # pylint: disable=broad-except
                 logger.exception(
-                    "Error generating checkpoint for %s, "
-                    "stage will be aborted",
-                    stage,
+                    "Error generating checkpoint, %s will be aborted", stage
                 )
                 proc.terminate()
             finally:
--- a/dvc/stage/run.py
+++ b/dvc/stage/run.py
@@ -39,9 +39,9 @@ def warn_if_fish(executable):
 
 
 @unlocked_repo
-def cmd_run(stage, *args, checkpoint=False, **kwargs):
+def cmd_run(stage, *args, checkpoint_func=None, **kwargs):
     kwargs = {"cwd": stage.wdir, "env": fix_env(None), "close_fds": True}
-    if checkpoint:
+    if checkpoint_func:
         # indicate that checkpoint cmd is being run inside DVC
         kwargs["env"].update({"DVC_CHECKPOINT": "1"})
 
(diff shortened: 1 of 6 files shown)


dvc/stage/run.py
```diff
@@ -144,14 +148,23 @@ def checkpoint_monitor(stage, callback_func):
...
     """Run callback_func whenever checkpoint signal file is present."""
     signal_path = os.path.join(stage.repo.tmp_dir, CHECKPOINT_SIGNAL_FILE)
     while True:
         if os.path.exists(signal_path):
-            _run_callback(stage, callback_func)
-            logger.debug("Remove checkpoint signal file")
-            os.remove(signal_path)
+            try:
+                _run_callback(stage, callback_func)
+            except Exception:  # pylint: disable=broad-except
```
pmrowla, line 158: To handle exceptions that occur during checkpoint commit generation (outside of the main thread) we either need to catch general `Exception` here or override [threading.excepthook](https://docs.python.org/3/library/threading.html#threading.excepthook) with our own handler, but I think it's cleaner/easier to read this way

dvc/stage/run.py
```diff
@@ -144,14 +148,23 @@ def checkpoint_monitor(stage, callback_func):
...
+                logger.exception(
+                    "Error generating checkpoint for %s, "
```
pmrowla, line 160: this will be properly single-quoted by `Stage.__str__`

pmrowla:
If an error occurs during checkpoint generation, we will end up showing 2 separate error messages, one with the checkpoint error, and one for when we SIGTERM the stage process.
```console
$ dvc exp run --checkpoint
What's next?
------------
- Check out the documentation: <https://dvc.org/doc>
- Get help and share ideas: <https://dvc.org/chat>
- Star us on GitHub: <https://github.com/iterative/dvc>
WARNING: No file hash info found for 'foo'. It won't be created.
        python checkpoint.py
Updating lock file 'dvc.lock'
ERROR: Error generating checkpoint, stage: 'foo' will be aborted - Checkpoint experiment containing '69c1970' already exists. To resume the experiment, run:

        dvc exp run --continue 69c1970
ERROR: failed to reproduce 'dvc.yaml': failed to run: python checkpoint.py, exited with -15
```

Not sure whether or not we should suppress the second error?

efiop (approved):

efiop:
@pmrowla The first one is in the worker thread?

pmrowla:
> @pmrowla The first one is in the worker thread?

@efiop yeah, the first one is the logged exception from the worker thread, and then the second message is logged when the Popen'd stage cmd process ends. The second one is actually dependent on the stage cmd, if it's a command that will return 0 on SIGTERM (because it expects to be long-running and manually cancelled) then we would just exit cleanly without the second message.

efiop:
@pmrowla Got it. Yeah, don't have an answer right away. Your call :slightly_smiling_face:

PR #4706, opened 2020-10-13
Organize metrics as table

indhupriya:

### DIFF
--- a/dvc/command/metrics.py
+++ b/dvc/command/metrics.py
@@ -28,6 +28,9 @@ def _show_metrics(
     header_set = set()
     rows = []
 
+    if precision is None:
+        precision = DEFAULT_PRECISION
+
     def _round(val):
         if isinstance(val, float):
             return round(val, precision)
@@ -234,6 +237,15 @@ def add_parser(subparsers, parent_parser):
             "If any target is a directory, recursively search and process "
             "metrics files."
         ),
+    )
+    metrics_show_parser.add_argument(
+        "--precision",
+        type=int,
+        help=(
+            "Round metrics to `n` digits precision after the decimal point. "
+            f"Rounds to {DEFAULT_PRECISION} digits by default."
+        ),
+        metavar="<n>",
     )
     metrics_show_parser.set_defaults(func=CmdMetricsShow)
 
--- a/dvc/command/metrics.py
+++ b/dvc/command/metrics.py
@@ -12,7 +12,11 @@ DEFAULT_PRECISION = 5
 
 
 def _show_metrics(
-    metrics, all_branches=False, all_tags=False, all_commits=False
+    metrics,
+    all_branches=False,
+    all_tags=False,
+    all_commits=False,
+    precision=None,
 ):
     from dvc.utils.diff import format_dict, table
     from dvc.utils.flatten import flatten
@@ -23,6 +27,12 @@ def _show_metrics(
     with_rev = any([all_branches, all_tags, all_commits])
     header_set = set()
     rows = []
+
+    def _round(val):
+        if isinstance(val, float):
+            return round(val, precision)
+        return val
+
     for _branch, val in metrics.items():
         for _fname, metric in val.items():
             if not isinstance(metric, dict):
@@ -44,7 +54,7 @@ def _show_metrics(
             flattened_val = flatten(format_dict(metric))
 
             for i in header:
-                row.append(flattened_val.get(i))
+                row.append(_round(flattened_val.get(i)))
             rows.append(row)
     header.insert(0, "Path")
     if with_rev:
--- a/dvc/command/metrics.py
+++ b/dvc/command/metrics.py
@@ -14,31 +14,45 @@ DEFAULT_PRECISION = 5
 def _show_metrics(
     metrics, all_branches=False, all_tags=False, all_commits=False
 ):
-    from dvc.utils.diff import format_dict
+    from dvc.utils.diff import format_dict, table
     from dvc.utils.flatten import flatten
 
     # When `metrics` contains a `None` key, it means that some files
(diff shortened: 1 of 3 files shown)


dvc/command/metrics.py
```diff
@@ -285,3 +298,9 @@ def add_parser(subparsers, parent_parser):
         metavar="<n>",
     )
     metrics_diff_parser.set_defaults(func=CmdMetricsDiff)
+
+
+# import json
+# s = '{"branch_1": {"metrics.json": {"a": 0, "b": {"ad": 0.0, "bc": 0.0}}}}'
+# s_d = json.loads(s)
+# print(_show_metrics(s_d,all_branches=True))
```
efiop, line 306: Leftovers?

dvc/command/metrics.py
```diff
@@ -14,31 +14,44 @@
...
+            flattened_val = flatten(format_dict(metric))
 
-            lines.append(f"\t{fname}:")
-            for key, value in flatten(format_dict(metric)).items():
-                lines.append(f"\t\t{key}: {value}")
+            for i in header:
+                if i in flattened_val:
+                    row.append(flattened_val[i])
+                else:
+                    row.append(None)
```
efiop, line 46:
```suggestion
                row.append(flattened_val.get(i))
```

dvc/command/metrics.py
```diff
@@ -14,31 +14,44 @@
...
+            rows.append(row)
+    header.insert(0, "Path")
+    if all_branches or all_tags or all_commits:
+        header.insert(0, "Branch")
```
efiop, line 50: Maybe `Revision`? `Branch` is too specific.

dvc/command/metrics.py
```diff
@@ -14,31 +14,44 @@
...
+            for key, _val in flatten(format_dict(metric)).items():
+                header_set.add(key)
+    header = sorted(header_set)
     for branch, val in metrics.items():
-        if all_branches or all_tags or all_commits:
-            lines.append(f"{branch}:")
-
         for fname, metric in val.items():
+            row = []
+            if all_branches or all_tags or all_commits:
```
efiop, line 34: Let's have 
```
with_rev = any(all_branches, all_tags, all_commits)
```
somewhere above and use `with_rev` here and in the `if` further below.

efiop (commented):
Looks good! Let's fix up the tests and cleanup some minor things in the code and we should be good :ok_hand:

dvc/command/metrics.py
indhupriya: Yeah, I'll remove this.

indhupriya:
HI @efiop I have made the requested changes. All tests have passed. I have added an empty header in case the metrics are not an instance of dictionary.

efiop:
@indhupriya Looking great now! One more thing, how about we also round the values? The same way we do in `metrics diff`(see `_round`). That will make it a bit nicer.

dvc/command/metrics.py
```diff
@@ -12,33 +12,57 @@
...
     missing = metrics.pop(None, None)
+    with_rev = any([all_branches, all_tags, all_commits])
+    header_set = set()
+    rows = []
 
-    lines = []
-    for branch, val in metrics.items():
-        if all_branches or all_tags or all_commits:
-            lines.append(f"{branch}:")
+    def _round(val):
```
efiop, line 31: Same as in `diff`
```suggestion
    if precision is None:
        precision = DEFAULT_PRECISION

    def _round(val):
```

tests/unit/command/test_metrics.py
```diff
@@ -226,9 +226,6 @@ def test_metrics_show_with_valid_falsey_values():
         all_branches=True,
     ) == textwrap.dedent(
         """\
```
efiop, line 228: Since the new format is more advanced, we need to add some more unit tests like this one. E.g. multiple Revisions, No revision, multiple metrics, precision. E.g. see `diff` tests above.

dvc/command/metrics.py
```diff
@@ -12,33 +12,57 @@
 
 
 def _show_metrics(
-    metrics, all_branches=False, all_tags=False, all_commits=False
+    metrics,
+    all_branches=False,
+    all_tags=False,
+    all_commits=False,
+    precision=None,
```
efiop, line 19: This one is not used right now, we need to add `--precision` to `metrics show` parser. See `--precision` at the very bottom of this file.

dvc/command/metrics.py
```diff
@@ -12,33 +12,57 @@
...
 
-            lines.append(f"\t{fname}:")
-            for key, value in flatten(format_dict(metric)).items():
-                lines.append(f"\t\t{key}: {value}")
+            for i in header:
+                row.append(_round(flattened_val.get(i)))
+            rows.append(row)
+    header.insert(0, "Path")
+    if with_rev:
+        header.insert(0, "Revision")
```
efiop, line 61: I think there should be a cleaner way to do this. E.g. no need to flatten the same metrics twice, etc. Give it a look, see if you can clean up this code a bit more.

efiop:
@indhupriya Just a few more comments to address and we should be good to merge :ok_hand: 

Thank you! :pray:

tests/unit/command/test_metrics.py
indhupriya: Ill add more tests

dvc/command/metrics.py
indhupriya: Any suggestions for this @efiop? The first flattening is required to get the header names

efiop: There is no reason to flatten it twice though. You could flatten and then reuse both.

efiop:
Thank you @indhupriya ! Let's merge and work on top.

PR #4715, opened 2020-10-14
checkpoints: create final checkpoint commit after repro when needed

### LINKED ISSUE #4708: checkpoints: create final checkpoint on job termination
This was marked as a research TODO in the current code, creating a final checkpoint should resolve:

> checkpoint (model file) is not preserved if I cancel a job
> no issues with the model If I train till the end - it seems like the `--persist` was just ignored

pmrowla:

Will fix #4708.

### DIFF
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -121,8 +121,7 @@ class Repo:
             self.tree = LocalTree(self, {"url": self.root_dir}, **tree_kwargs)
 
         self.config = Config(self.dvc_dir, tree=self.tree)
-        no_scm = self.config["core"].get("no_scm", False)
-        self.scm = scm if scm else SCM(self.root_dir, no_scm=no_scm)
+        self._scm = scm
 
         # used by RepoTree to determine if it should traverse subrepos
         self.subrepos = subrepos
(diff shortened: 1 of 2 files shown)


dvc/repo/experiments/__init__.py
```diff
@@ -563,22 +563,26 @@ def _checkpoint_callback(rev, executor, unchanged, stages):
...
-        # NOTE: our cached GitPython Repo instance cannot be re-used if the
-        # checkpoint run was interrupted via SIGINT, so we need this hack
-        # to create a new git repo instance after checkpoint runs.
-        del self.scm
+            # NOTE: GitPython Repo instances cannot be re-used after
+            # process has received SIGINT or SIGTERM, so we need this hack
+            # to re-instantiate git instances after checkpoint runs. See:
+            # https://github.com/gitpython-developers/GitPython/issues/427
+            del self.repo.scm
+            del self.scm
```
pmrowla, line 580: The leftover `BrokenPipeError` from the initial checkpoints demo was because we were previously only re-instantiating `experiments.scm` after checkpoint runs (and not parent `repo.scm`).

As previously discussed, the need for this hack may be avoided by moving to dulwich or pygit, but checkpoints is currently the only place in DVC where we should be doing any git operations after SIGINT/SIGTERM, so this workaround works for now.

dvc/repo/__init__.py
```diff
@@ -158,6 +157,13 @@ def __init__(
         self.plots = Plots(self)
         self.params = Params(self)
 
+    @cached_property
+    def scm(self):
```
pared, line 161: +1

PR #4733, opened 2020-10-16
fix making too many ssh connections when pushing run cache (#4732)

### LINKED ISSUE #4732: `dvc push --run-cache` is incredibly slow on "ssh" remote
## Bug Report
A command `dvc push --run-cache` takes very long time (more than hour in my case) if there are a lot files in `dvc_cache/runs`. The reason is making a new ssh connection for each file to check if it exist on the remote.

**Output of `dvc version`:**

```console
$ dvc version
1.8.4+21a6be
```

```
$ dvc push --run-cache --verbose
2020-10-16 14:36:14,695 DEBUG: Check for update is enabled.
2020-10-16 14:36:14,697 DEBUG: fetched: [(3,)]                        
2020-10-16 14:36:14,818 DEBUG: Establishing ssh connection with 'xxx' through port '22' as user 'dvc'
2020-10-16 14:36:15,457 DEBUG: Establishing ssh connection with 'xxx' through port '22' as user 'dvc'
2020-10-16 14:36:16,070 DEBUG: Establishing ssh connection with 'xxx' through port '22' as user 'dvc'
```
and so on, hundreds of time.

peper0:

Fixes #4732.

### DIFF
--- a/dvc/tree/ssh/__init__.py
+++ b/dvc/tree/ssh/__init__.py
@@ -218,6 +218,7 @@ class SSHTree(BaseTree):
 
         # See dvc/remote/local/__init__.py - hardlink()
         if self.getsize(from_info) == 0:
+
             with self.ssh(to_info) as ssh:
                 ssh.sftp.open(to_info.path, "w").close()
 
--- a/dvc/tree/ssh/__init__.py
+++ b/dvc/tree/ssh/__init__.py
@@ -3,7 +3,7 @@ import io
 import logging
 import os
 import threading
-from contextlib import closing, contextmanager
+from contextlib import closing, contextmanager, suppress
 from urllib.parse import urlparse
 
 from funcy import first, memoize, silent, wrap_with
@@ -173,7 +173,8 @@ class SSHTree(BaseTree):
     def walk_files(self, path_info, **kwargs):
         with self.ssh(path_info) as ssh:
             for fname in ssh.walk_files(path_info.path):
-                yield path_info.replace(path=fname)
+                with suppress(GeneratorExit):  # dropping further walk in the middle doesn't hurt
+                    yield path_info.replace(path=fname)
 
     def remove(self, path_info):
         if path_info.scheme != self.scheme:
--- a/dvc/tree/ssh/__init__.py
+++ b/dvc/tree/ssh/__init__.py
@@ -3,7 +3,7 @@ import io
 import logging
 import os
 import threading
-from contextlib import closing, contextmanager, suppress
+from contextlib import closing, contextmanager
 from urllib.parse import urlparse
 
 from funcy import first, memoize, silent, wrap_with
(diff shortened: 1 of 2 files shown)


dvc/tree/ssh/__init__.py
```diff
@@ -215,7 +219,6 @@ def hardlink(self, from_info, to_info):
 
         # See dvc/remote/local/__init__.py - hardlink()
         if self.getsize(from_info) == 0:
-
             with self.ssh(to_info) as ssh:
```
efiop, line 222:
```suggestion

            with self.ssh(to_info) as ssh:
```

dvc/tree/ssh/__init__.py
```diff
@@ -174,7 +174,11 @@ def isfile(self, path_info):
     def walk_files(self, path_info, **kwargs):
         with self.ssh(path_info) as ssh:
             for fname in ssh.walk_files(path_info.path):
-                yield path_info.replace(path=fname)
+                try:
+                    yield path_info.replace(path=fname)
+                except GeneratorExit:
+                    # raised when generator is dropped in the middle of walk
+                    break
```
efiop, line 181: Great catch!

I guess the reason is https://github.com/iterative/dvc/blob/de01b11483c11d1b0cf55caf264b7bc65f3c8347/dvc/tree/pool.py#L14 , and not this particular piece of code. Also, this bug might affect other remote types as well, so even more reason to fix it there.

peper0: The line that you pointed is responsible for dropping connections, that was used when some unhandled exception was raised. Maybe it's a bit aggressive but imo it's not a bad solution - possibly some bad state of connection was a cause for this exception?
The code that I fixed was buggy for sure. It's completely valid to start iterating `walk_files()` generator and abandon it without iterating through all elements. But without my fix, such a usage raises an exception, which is wrong.

pmrowla: It should be valid to not catch `GeneratorExit` in a generator though, since `GeneratorExit` isn't even an error condition (which is why it inherits from `BaseException` and not `Exception`). Normally you catch `GeneratorExit` inside a generator function if you have to run cleanup code before exiting the function. We are not doing at all here, so this empty `except GeneratorExit` looks strange.

Basically, a generator raising `GeneratorExit` on `break` is expected behavior. The bug is that we incorrectly treat this expected behavior as an error in the connection pool wrapper.

The fix for this should probably be to catch `Exception`, and maybe also `KeyboardInterrupt`, instead of `BaseException` in the connection pool code.

peper0: Ok, you both are right. I misunderstood a bit how `GeneratorExit` works in this case.

efiop: Maybe @Suor could shed some light on the reasoning behind BaseException in get_connection?

Suor: Not sure about here, but on pool we should watch for everything, including. The logic is clear:
- we pop a connection from a pool
- if something happened during that connection use then we close it
- if connection was used successfully we return it to the pool
- we always reraise so that behavior is not altered

efiop:
@Mergifyio backport 1.11

efiop:
@peper0 Thank you for your patience!

Can confirm this change definitely helps. With it `push --run-cache` with 2K run-cache entries in the remote takes around 3minutes (still a lot, we need status-like approach for run-cache, same as we do with regular `dvc push/pull` for cache files), but without this fix it takes more than an hour.

A more faster cache-like approach for transfering run-cache will be implemented in the future.

peper0:
Thanks a lot!

efiop:
@peper0 Thank you! 1.11.6 will be out on pypi in a few minutes https://github.com/iterative/dvc/releases/tag/1.11.6 , other packages will be released later today.

peper0:
This bug is still present in 1.11.15. It seems that previous fix didn't work. I created a new PR: https://github.com/iterative/dvc/pull/5458

PR #4734, opened 2020-10-16
Implement foreach ... in loop in dvc.yaml

skshetry:

Here's a wiki documenting the feature: https://github.com/iterative/dvc/wiki/Parametrization#foreach--in.
And, following are the examples for the foreach:

1. Looping through a list
```yaml
stages:
  build:
    foreach:
    - foo
    - bar
    - baz
    in:
      cmd: python script.py ${item}
```

2. Looping through a dictionary
```yaml
stages:
  build:
    foreach:
      models:
        us:
          thresh: 10
        gb:
          thresh: 15
    in:
      cmd: python script.py ${item}
```

3. Looping throuhg values in [vars](https://github.com/iterative/dvc/wiki/Parametrization#vars-section) or params.yaml
```yaml
stages:
  build:
    foreach: ${models}
    in:
      cmd: python script.py --thresh ${item.thresh}
```

You might notice `set` in the wiki. That is currently not implemented.

### DIFF
--- a/dvc/parsing/__init__.py
+++ b/dvc/parsing/__init__.py
@@ -1,6 +1,7 @@
 import logging
 import os
 from collections import defaultdict
+from collections.abc import Mapping, Sequence
 from copy import deepcopy
 from itertools import starmap
 from typing import TYPE_CHECKING
@@ -25,6 +26,10 @@ VARS_KWD = "vars"
 WDIR_KWD = "wdir"
 DEFAULT_PARAMS_FILE = ParamsDependency.DEFAULT_PARAMS_FILE
 PARAMS_KWD = "params"
+FOREACH_KWD = "foreach"
+IN_KWD = "in"
+
+DEFAULT_SENTINEL = object()
 
 
 class DataResolver:
@@ -50,6 +55,11 @@ class DataResolver:
 
     def _resolve_entry(self, name: str, definition):
         context = Context.clone(self.global_ctx)
+        if FOREACH_KWD in definition:
+            assert IN_KWD in definition
+            return self._foreach(
+                context, name, definition[FOREACH_KWD], definition[IN_KWD]
+            )
         return self._resolve_stage(context, name, definition)
 
     def resolve(self):
@@ -114,3 +124,21 @@ class DataResolver:
             return self.wdir
         wdir = resolve(wdir, context)
         return self.wdir / str(wdir)
+
+    def _foreach(self, context: Context, name: str, foreach_data, in_data):
+        def each_iter(value, key=DEFAULT_SENTINEL):
+            c = Context.clone(context)
+            c["item"] = value
+            if key is not DEFAULT_SENTINEL:
+                c["key"] = key
+            suffix = str(key if key is not DEFAULT_SENTINEL else value)
+            return self._resolve_stage(c, f"{name}-{suffix}", in_data)
+
+        iterable = resolve(foreach_data, context)
+        if isinstance(iterable, Sequence):
+            gen = (each_iter(v) for v in iterable)
+        elif isinstance(iterable, Mapping):
+            gen = (each_iter(v, k) for k, v in iterable.items())
+        else:
+            raise Exception(f"got type of {type(iterable)}")
+        return join(gen)
--- a/tests/func/test_stage_resolver.py
+++ b/tests/func/test_stage_resolver.py
@@ -233,3 +233,72 @@ def test_with_templated_wdir(tmp_dir, dvc):
             }
         },
     )
+
+
+def test_simple_foreach_loop(tmp_dir, dvc):
+    iterable = ["foo", "bar", "baz"]
+    d = {
+        "stages": {
+            "build": {
+                "foreach": iterable,
+                "in": {"cmd": "python script.py ${item}"},
+            }
+        }
+    }
+
+    resolver = DataResolver(dvc, PathInfo(str(tmp_dir)), d)
+    assert resolver.resolve() == {
+        "stages": {
+            f"build-{item}": {"cmd": f"python script.py {item}"}
+            for item in iterable
+        }
+    }
+
+
+def test_foreach_loop_dict(tmp_dir, dvc):
+    iterable = {"models": {"us": {"thresh": 10}, "gb": {"thresh": 15}}}
+    d = {
+        "stages": {
+            "build": {
+                "foreach": iterable,
+                "in": {"cmd": "python script.py ${item}"},
+            }
+        }
+    }
+
+    resolver = DataResolver(dvc, PathInfo(str(tmp_dir)), d)
+    assert resolver.resolve() == {
+        "stages": {
+            f"build-{key}": {"cmd": f"python script.py {item}"}
+            for key, item in iterable.items()
+        }
+    }
+
+
+def test_foreach_loop_templatized(tmp_dir, dvc):
+    params = {"models": {"us": {"thresh": 10}}}
+    vars_ = {"models": {"gb": {"thresh": 15}}}
+    dump_yaml(tmp_dir / DEFAULT_PARAMS_FILE, params)
+    d = {
+        "vars": vars_,
+        "stages": {
+            "build": {
+                "foreach": "${models}",
+                "in": {"cmd": "python script.py --thresh ${item.thresh}"},
+            }
+        },
+    }
+
+    resolver = DataResolver(dvc, PathInfo(str(tmp_dir)), d)
+    assert_stage_equal(
+        resolver.resolve(),
+        {
+            "stages": {
+                "build-gb": {"cmd": "python script.py --thresh 15"},
+                "build-us": {
+                    "cmd": "python script.py --thresh 10",
+                    "params": ["models.us.thresh"],
+                },
+            }
+        },
+    )
--- a/tests/func/test_stage_resolver.py
+++ b/tests/func/test_stage_resolver.py
@@ -260,8 +260,8 @@ def test_foreach_loop_dict(tmp_dir, dvc):
     d = {
         "stages": {
             "build": {
-                "foreach": iterable,
-                "in": {"cmd": "python script.py ${item}"},
+                "foreach": iterable["models"],
+                "in": {"cmd": "python script.py ${item.thresh}"},
             }
         }
     }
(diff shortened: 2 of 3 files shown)


shcheklein:
This is excellent stuff, @skshetry !! I'll take a look this weekend and give some feedback if I have anything!

Should we invite some folks from the tickets related to this? To give it a try as well?

skshetry:
@shcheklein, there are few bugs that I need to fix. Then I'll write up in that ticket. Thanks.

skshetry:
Just noticed that `set` brokes auto-tracking. Keeping it as draft.

skshetry:
Btw, there's a wiki for this feature, as I have planned: https://github.com/iterative/dvc/wiki/Parametrization

srp-31:
This is great stuff. Waiting for it to be merged!

jorgeorpinel:
Woah super interesting stuff. Definitely a common need from what I remember (support/feature requests). Also a lot to document eventually so thanks for maintaining that wiki for now @skshetry!

I have a bunch of QA-related thoughts but the only thing that jumps out to me at this stage is about the foreach syntax. I find it a little confusing to call the thing to do `in:`. How about copying Python's `for` syntax more or less?

```yaml
...
for: item  # just a name to refer later
in: ${items}  # taken from `var` or params file in this case but cold be a hardcoded list/dict
do:
  cmd: # uses ${item.foo}, ${item.bar}, etc.
```

^ This way also the user names the index so that `item` is not a special keyword.

jorgeorpinel:
Alternatively, just rename it `foreach`/`do` (no "in" needed with "foreach", I think).

jorgeorpinel:
OR something like

```yaml
...
foreach:
  item:  # Could also be named by the user with this syntax
    - 2
    - 4
    ...
  cmd: # uses ${item}
```

tests/func/test_stage_resolver.py
```diff
@@ -233,3 +233,72 @@ def test_with_templated_wdir(tmp_dir, dvc):
...
+                "foreach": iterable,
+                "in": {"cmd": "python script.py ${item}"},
+            }
+        }
+    }
+
+    resolver = DataResolver(dvc, PathInfo(str(tmp_dir)), d)
+    assert resolver.resolve() == {
+        "stages": {
+            f"build-{key}": {"cmd": f"python script.py {item}"}
```
pared, line 272: The resolved `cmd` section is 
`'python script.py {\'us\': {\'thresh\': 10}, \'gb\': {\'thresh\': 15}}'`. Not sure whether this was the anticipated result. I think It might be more explicit if we would provide expected and resolved dictionary instead of building it from `iterable`.

dvc/parsing/__init__.py
```diff
@@ -25,6 +26,10 @@
 WDIR_KWD = "wdir"
 DEFAULT_PARAMS_FILE = ParamsDependency.DEFAULT_PARAMS_FILE
 PARAMS_KWD = "params"
+FOREACH_KWD = "foreach"
+IN_KWD = "in"
+
+DEFAULT_SENTINEL = object()
```
pared, line 32: Can't we use `None` instead of `DEFAULT_SENTINEL`? Seems like it's used only to check whether the key has been provided?

pared (changes requested):
This feature is cool! Left two comments.

jorgeorpinel (changes requested):
Other than my syntax suggestion (see https://github.com/iterative/dvc/pull/4734#issuecomment-712658511 above), I noticed that the [`use`](https://github.com/iterative/dvc/wiki/Parametrization#use-section) and [`vars`](https://github.com/iterative/dvc/wiki/Parametrization#vars-section) sections are already included in this PR (added a mention to the PR desc.) so I tested that first:

> Using DVC `1.8.4+8be0d8` at the moment

- Should `use` be called `params`? It's not clear what we mean by "use" e.g. dvc.yaml pipelines use all sort of things, like deps and outs.
- [ ] `use` accepts inexistent paths, resulting in silently not loading any params file, even if params.yaml is present. Should it at least warn that the given file doesn't exist and so no params will be available in dvc.yaml as vars? ~~Or warn AND load params.yaml (if it exists, and include this note in the warning).~~ (It's nice to have the `use: none` trick actually.)
- [ ] If `use` is given a directory path, it prints a generic `unexpected error - [Errno 13] Permission denied: '{dirname}'`.
- When vars used in dvc.yaml don't exist in params file or `vars`, you get `ERROR: unexpected error - Could not find '{key}' in {}: '{key}'` — the error could be a little more informative, I think.
- [ ] `vars` is supposed to get recursively merged with the params file but I always get an error such as `unexpected error - Cannot overwrite as key {keyname} already exists in {'{keyname}': {value}}` (I'm using the same dictionary in both the params file and in vars, but with different values).
  The `{value}` shown comes from params.yaml if no `use` is given, except if I `use: none` (inexistent file) in which case I still get the error (weird) but the value from `vars` is printed in the error msg.

tests/func/test_stage_resolver.py
skshetry: Yup, that should have been `item.thresh`. Regarding interpolating dict, it will be handled later.

dvc/parsing/__init__.py
skshetry: Not trying to assume to much about the parametrized data.

skshetry:
@jorgeorpinel, thanks for the QA. 

>     * Should `use` be called `params`? It's not clear what we mean by "use" e.g. dvc.yaml pipelines use all sort of things, like deps and outs.

`params` already has a different format/schema in stages, which might be confusing. But, maybe, it's not? 
Let see, I'll focus on trying to unify them by some means (but so far I am not seeing it clearly).

>     * [ ]  `use` accepts inexistent paths, resulting in silently not loading any params file, even if params.yaml is present. Should it at least warn that the given file doesn't exist and so no params will be available in dvc.yaml as `{$vars}`? ~Or warn AND load params.yaml (if it exists, and include this note in the warning).~ (It's nice to have the `use: none` trick actually.)

It might be difficult to keep backward compatibility and try to give good error message. The behavior definitely needs to be improved/fixed, will be fixed in successive PRs.

>     * [ ]  If `use` is given a directory path, it prints a generic `unexpected error - [Errno 13] Permission denied: '{path}'`.

Same as above.

>     * When vars used in dvc.yaml don't exist in params file or `vars`, you get `ERROR: unexpected error - Could not find '{key}' in {}: '{key}'` — the error could be a little more informative, I think.

Same as above.

>     * [ ]  `vars` is supposed to get recursively merged with the params file but I always get an error such as `unexpected error - Cannot overwrite as key {keyname} already exists in {'{keyname}': {value}}` (I'm using the same dictionary in both the params file and in `vars`, but with different values).

It recursively merges the dictionary, but if that's the same file, it'll have the same keys at the root level, which is not allowed.
So, we do merge things, but does not allow it to be overwritten.



> OR something like
> 
> ```yaml
> ...
> foreach:
>   item:  # Could also be named by the user with this syntax
>     - 2
>     - 4
>     ...
>   cmd: # uses ${item}
> ```

This way, `item` will be a reserved keyword. Though, I'll consider it (syntax is of course not set in stone).

jorgeorpinel:
Np. Actually that was just the first round of QA. I have several other things but I didn't want to post everything at once. I see this is merged though, is there a PR that continues it, for the remaining QA rounds? Thanks.

For now, some answers to the comments above:

> This way, item will be a reserved keyword

True. Actually there's a better way, since `in:` is always followed by `cmd:` what about just skip `in` and go straight to `cmd`? That addresses my concern around the semantics of "foreach ... in ...". Like this:

```yaml
stages:
  something:
    foreach: ...
    cmd: python script.py --thresh ${...}
```

> params already has a different format/schema in stages, 

Also true. So maybe `params-file`, `import`, or something else? My main suggestion is that `uses` is not clear (again, semantics). It's not a major thing but why not improve that if we can.

> > use accepts inexistent paths, resulting in silently not loading any params file, even if params.yaml is present
> > If use is given a directory path, it prints a generic unexpected error
>
> The behavior definitely needs to be improved/fixed, will be fixed in successive PRs.

OK, what PRs are those? I was asked QA this but I'm not sure how best to keep track of the issues I find if the PR is merged. The most obvious idea is to open issues for each of my check boxes above but... The point of QA is to avoid issues in the first place 🙂 

> we do merge things, but does not allow it to be overwritten

Got it. So the error message could be a little better and probably overwriting would be valuable. But fine, that's not very important right now. Did you check why the strange logic in the last edge case I mentioned? That behavior seems a little unexpected:

> The {value} shown comes from params.yaml if no use is given, except if I use: none (inexistent file) in which case I still get the error (weird) but the value from vars is printed

jorgeorpinel:
Continued in #4854

jcpsantiago:
This does exactly what I needed and discussed in one of the original feature discussion issues. Thanks everyone 🙏 🔥

skshetry:
@jcpsantiago, please note that this is a experimental feature, which is not stable yet and might change. Please do take a look, and tell us what you think. :slightly_smiling_face:

jcpsantiago:
sure, I'm modifying my pipeline to test it ;) @skshetry there is no mention about the behaviour of `dvc metrics` and `dvc plots` in the wiki. Are these commands already taking into account multiple metrics/plot files?

skshetry:
> there is no mention about the behaviour of `dvc metrics` and `dvc plots` in the wiki. 

@jcpsantiago, that should work, as this has nothing to do with `dvc metrics` and `dvc plots`.

Regarding mentioning them in `dvc.yaml` it should be working for `metrics` and `plots` as well. I guess the think that's not working is, interpolating stuffs for a key of the dictionary.
So, the following should work:
```yaml
outputs: 
  - ${foo}
  - ${bar}
```

But, the following does not work yet:
```yaml
outputs:
  - ${foo}:
      persist: True
```

skshetry:
@jcpsantiago, The latter case with keys being interpolated, should work now. PTAL.

jcpsantiago:
@skshetry which branch should I checkout?

jcpsantiago:
@skshetry everything works. I'm close to tears by how beautiful this is 😭 I'm having a very nerdy moment seeing DVC plow through multiple models and then showing me the plots and metrics on the terminal all in one go. Awesome work everyone!

PR #4737, opened 2020-10-17
command: add dvc push instruction after dvc repro

### LINKED ISSUE #2359: add `dvc push` to the instructions after `dvc repro`
(See description in https://github.com/iterative/dvc/issues/2359#issuecomment-517826724 below.)

jvfe:
Adds a logger.info() statement to remind the user to run dvc push after
running dvc repro.

Closes #2359

### DIFF
--- a/dvc/command/repro.py
+++ b/dvc/command/repro.py
@@ -30,14 +30,15 @@ class CmdRepro(CmdBase):
 
                 if len(stages) == 0:
                     logger.info(CmdDataStatus.UP_TO_DATE_MSG)
+                else:
+                    logger.info(
+                        "Use `dvc push` to send your updates to "
+                        "remote storage."
+                    )
 
                 if self.args.metrics:
                     metrics = self.repo.metrics.show()
                     logger.info(_show_metrics(metrics))
-
-                logger.info(
-                    "Use 'dvc push' to send your updates to remote storage."
-                )
 
             except DvcException:
                 logger.exception("")
--- a/dvc/command/repro.py
+++ b/dvc/command/repro.py
@@ -35,6 +35,10 @@ class CmdRepro(CmdBase):
                     metrics = self.repo.metrics.show()
                     logger.info(_show_metrics(metrics))
 
+                logger.info(
+                    "To save your updates to remote storage, run 'dvc push'."
+                )
+
             except DvcException:
                 logger.exception("")
                 ret = 1
--- a/dvc/command/repro.py
+++ b/dvc/command/repro.py
@@ -36,7 +36,7 @@ class CmdRepro(CmdBase):
                     logger.info(_show_metrics(metrics))
 
                 logger.info(
-                    "To save your updates to remote storage, run 'dvc push'."
+                    "Use 'dvc push' to send your updates to remote storage."
                 )
 
             except DvcException:


dvc/command/repro.py
```diff
@@ -35,6 +35,10 @@ def run(self):
                     metrics = self.repo.metrics.show()
                     logger.info(_show_metrics(metrics))
 
+                logger.info(
+                    "To save your updates to remote storage, run 'dvc push'."
```
pared, line 39: Maybe something like "Use 'dvc push' to send your updates to remote storage."?

pared (approved):
@jvfe I am sorry for such a late response. The change looks good, though I have a suggestion about phrasing. Don't feel obliged to comply, I am not a native speaker.

dvc/command/repro.py
jvfe: No problem, done!

pmrowla (approved):

dvc/command/repro.py
```diff
@@ -35,6 +35,10 @@ def run(self):
                     metrics = self.repo.metrics.show()
                     logger.info(_show_metrics(metrics))
 
+                logger.info(
+                    "Use 'dvc push' to send your updates to remote storage."
+                )
```
pared, line 40: Two more things:
1. please use ` instead of ' - to keep consistency with other help messages
2. I think we should move this message few lines up and log it under `if len(stages)==0` in `else` cause, so that we will not get this message when there is nothing actually reproduced.

What do you think @jvfe?

jvfe: Sorry, I don't really understand what you mean by using "instead of" in this context, could you explain further?

Also, sounds good putting an else statement below if `len(stages)==0`, like this, right?

```py
if len(stages) == 0:
    logger.info(CmdDataStatus.UP_TO_DATE_MSG)
else:
    logger.info(
        "Use 'dvc push' to send your updates to remote storage."
    )
```

pared: @jvfe I mean backticks (  **`**  ) instead of ticks (  **'**  ) - sorry its hard to highlit them, due to markdown formatting.

jvfe: Ok, thanks for explaining! I've done what you suggested but had to split the message in two lines or the pre-commit hook for flake8 would complain.

pared:
Seems like builds are failing for an unrelated reason.

dvc/command/repro.py
```diff
@@ -30,6 +30,11 @@ def run(self):
 
                 if len(stages) == 0:
                     logger.info(CmdDataStatus.UP_TO_DATE_MSG)
+                else:
+                    logger.info(
+                        "Use `dvc push` to send your updates to "
+                        "remote storage."
```
jorgeorpinel, line 36: Minor: Should we update some of the console examples in https://dvc.org/doc/command-reference/repro? Several show the command's output.

> p.s. looks like that could fit in a single-line-string?

pared: This is a good point, but there is a problem:
In the example, in `repro` we modify code only, without influence on the output of the stage...
Seems to me like we should not display this help message there. So I would say we should not update the command reference but fix the underlying issue.

PR #4738, opened 2020-10-17
get: checking target path before downloading

### LINKED ISSUE #3673: get/import: check target path before downloading
```console
λ dvc version
DVC version: 0.93.0
Python version: 3.7.5
Binary: True
Package: exe
Cache: reflink - not supported, hardlink - supported, symlink - not supported
Filesystem type (cache directory): ('NTFS', 'C:\\')
Repo: dvc, git

λ echo data > data.xml
λ la
...
-rw-r--r-- 1 usd grp        5 Apr 23 21:31 data.xml
λ dvc get https://github.com/iterative/dataset-registry get-started/data.xml -o data.xml
# HERE I SEE THE DOWNLOAD BAR FIRST, and the whole transfer is done up to 100%
ERROR: unexpected error - [WinError 5] Access is denied: 'C:\\Users\\poj12\\DVC-repos\\dvc-get-started\\data.xml.Cf4LfW3t5gZ7aK9F2qXquP'
λ la
...
-r--r--r-- 1 usd grp 37916850 Apr 23 21:26 data.xml
-rw-r--r-- 1 usd grp 37916850 Apr 23 21:26 data.xml.Cf4LfW3t5gZ7aK9F2qXquP
```

The result is I end up with the download  having overwritten the small data.xml I created, but with an error message, and with the tmp file not deleted either, as it can be seen above.

julsal:
Fixes #3673

### DIFF
--- a/tests/func/test_get.py
+++ b/tests/func/test_get.py
@@ -1,10 +1,11 @@
 import logging
 import os
 
+import mock
 import pytest
 
 from dvc.cache import Cache
-from dvc.exceptions import PathMissingError
+from dvc.exceptions import ConfirmRemoveError, PathMissingError
 from dvc.external_repo import IsADVCRepoError
 from dvc.main import main
 from dvc.repo import Repo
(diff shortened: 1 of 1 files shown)


dvc/external_repo.py
```diff
@@ -195,11 +195,26 @@ def get_external(self, path, dest):
...
+                if not prompt.confirm(msg):
+                    raise ConfirmRemoveError(str(path_info))
+                force = True
+
             repo = metadata.repo
             cache = repo.cache.local
             # fetch DVC and git files to tmpdir cache, then checkout
             save_info = self._fetch_to_cache(path_info, repo, None)
-            cache.checkout(PathInfo(dest), save_info)
+            cache.checkout(dest_path_info, save_info, force=force)
```
efiop, line 217: Hi @julsal !

Thanks for the PR!

By default `force=False`, so if the output already exists and it differs from the desired one, then we will prompt user for removal confirmation. E.g.
```
#!/bin/bash                                                                       
                                                                                  
set -e                                                                            
set -x                                                                            
                                                                                  
rm -rf mytest                                                                     
mkdir mytest                                                                      
cd mytest                                                                         
                                                                                  
echo data > data                                                                  
dvc get https://github.com/iterative/dataset-registry get-started/data.xml -o data # will prompt
```
So this whole change is not really needed :slightly_frowning_face: You probably got confused by running `dvc get ...` twice, in which case it won't prompt you simply because the files are identical, so it doesn't have to prompt.

efiop (changes requested):

dvc/external_repo.py
julsal: Hi @efiop,

Thanks for the quick reply.

The change comes in the context of #3673, which shows that dvc asks for confirmation only after downloading the file. For this reason I introduced an earlier prompt. `force=True` comes to avoid another prompt.

Does it make sense when considering the issue?

efiop: @julsal Ah, I see. Indeed, there was a misunderstanding on my side. Thank you for pointing out!

Just to clarify, this is a good PR :+1: and counts as a pass for the application purposes, but we want to reconsider from the tree arch perspective, which will require some more effort. So just to not waste your more of your time on this, we could close the PR for now.

julsal: @efiop I agree. A future deeper refactoring might find a more elegant solution. Nonetheless, if you find useful, I changed the PR to keep only the test cases for file replacement.

efiop (approved):

dvc/external_repo.py
efiop: Thank you @julsal  :pray:

PR #4740, opened 2020-10-19
experiments: fix checkpoints sorting issues in `dvc exp show`

### LINKED ISSUE #4709: checkpoints: minor table/UI issues
* [ ] checkpoints are not grouped properly during sorting (#4740)
* [ ] checkpoint branches are not sorted as expected by default - currently branches are returned alphabetically by branch name rather than by the most recent tip commit timestamp (#4740)
* [x] small floating point numbers are rounded to `0.0` instead of being displayed in scientific notation (#4730)

pmrowla:

Will close #4709.

* Checkpoint branches are now ordered by tip commit timestamp by default (newest to oldest)
* When using `dvc exp show --sort-by`, checkpoint branches will be sorted on the tip commit metric/param values. All other commits within the same checkpoint branch will remain grouped together and ordered from newest to oldest.

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -155,7 +155,11 @@ def _sort_exp(experiments, sort_by, typ, reverse):
         ret = OrderedDict()
 
     def _sort(item):
-        _, exp = item
+        rev, exp = item
+        tip = exp.get("checkpoint_tip")
+        if tip and tip != rev:
+            # Sort checkpoint experiments by tip commit
+            return _sort((tip, experiments[tip]))
         for fname, item in exp.get(typ, {}).items():
             if isinstance(item, dict):
                 item = flatten(item)
--- a/dvc/scm/git.py
+++ b/dvc/scm/git.py
@@ -294,8 +294,18 @@ class Git(Base):
     def active_branch(self):
         return self.repo.active_branch.name
 
-    def list_branches(self):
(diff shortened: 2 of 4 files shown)


dvc/command/experiments.py
```diff
@@ -155,7 +155,11 @@ def _sort_exp(experiments, sort_by, typ, reverse):
         ret = OrderedDict()
 
     def _sort(item):
-        _, exp = item
+        rev, exp = item
+        tip = exp.get("checkpoint_tip")
+        if tip and tip != rev:
```
pmrowla, line 160: if `tip == rev`, this is tip of a checkpoint branch and we want to use the actual metric/param values from this experiment

dvc/scm/git.py
```diff
@@ -294,8 +294,18 @@ def is_dirty(self, **kwargs):
...
+
+        Any provided kwargs will be passed into `sorted()` to sort the list of
+        git head (branch) objects. By default branches are listed
+        alphabetically.
+        """
+        if kwargs:
+            heads = sorted(self.repo.heads, **kwargs)
+        else:
+            heads = self.repo.heads
+        return [h.name for h in heads]
```
efiop, line 308: Maybe let's sort where we need to? This is a bit confusing with this kwargs & sorted shenanigans and doesn't belong to this method.

PR #4747, opened 2020-10-19
Support for WebHDFS

### LINKED ISSUE #4687: add support for WebHDFS
We currently support real HDFS by using pyarrow, but the problem with it is that it still requires libhdfs, which has to be installed separately and in some cases (on osx) is not at all trivial to install (brew's hadoop package doesn't provide it, for example). See https://github.com/iterative/dvc/issues/4656 Also [pyarrow itself has caused us some problems](https://github.com/iterative/dvc/issues?q=is%3Aissue+pyarrow+), as it requires wheels to be present for all platforms.

WebHDFS is often enabled by default alongside pure HDFS and there are lots of available libs for working with it https://pypi.org/search/?q=hdfs . It would be great if we could implement some basic support for it to have it as a potential alternative to pure HDFS. To do that, we need to support urls that look like `webhdfs://example.com/path`, e.g.:

```
dvc remote add my webhdfs://example.com/path
```

which requires creating `dvc/tree/webhdfs.py` with an appropriate Tree class, which should at least implement these methods:

```
exists
_upload
_download
```
which will allow us to use `dvc push/fetch/pull/(status -c)`. One could use one of the existing implementation for other clouds in dvc/tree, e.g. https://github.com/iterative/dvc/blob/master/dvc/tree/oss.py  as a template.

tomasfarias:
* [✅ ] ❗ I have followed the [Contributing to DVC](https://dvc.org/doc/user-guide/contributing/core) checklist.

* [✔️] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here.

This is a draft PR to close #4687. Documentation is required ~but no issue or separate PR has been created yet~, is tracked by https://github.com/iterative/dvc.org/issues/1899, and being worked on in https://github.com/iterative/dvc.org/pull/1916.

As I work on the PR, I would like some input on:
* Choice of library: I went with [hdfs](https://pypi.org/project/hdfs/) as it seems to be the most stable.
* Scope: assuming consensus on library, there's some optional parameters that could be supported on connection (for example, a `timeout` or `proxy`), as well as different connection clients (an `InsecureClient` using `user` and a `TokenClient`). Should the connection parameters be limited? I have currently prioritized the use of a config file as it's the preferred method for `hdfs` (the library), and left `InsecureClient` and `TokenClient` as fallbacks if no config file exists.
* Whether I'm over-complicating functional testing by creating a `WebHDFS` class similar to `HDFS` (available [here](https://github.com/iterative/dvc/blob/master/tests/remotes/hdfs.py#L15)). It smells a bit to me as it's mostly passing calls to the underlying client, but I'm not too certain on how functional tests use `HDFS` 😃.

### DIFF
--- a/dvc/tree/webhdfs.py
+++ b/dvc/tree/webhdfs.py
@@ -7,6 +7,7 @@ from urllib.parse import urlparse
 
 from funcy import cached_property, wrap_prop
 
+from dvc.hash_info import HashInfo
 from dvc.path_info import CloudURLInfo
 from dvc.progress import Tqdm
 from dvc.scheme import Schemes
@@ -32,6 +33,7 @@ class WebHDFSTree(BaseTree):
     PATH_CLS = CloudURLInfo
     REQUIRES = {"hdfs": "hdfs"}
     REGEX = r"^webhdfs://((?P<user>.*)@)?.*$"
+    PARAM_CHECKSUM = "checksum"
 
     def __init__(self, repo, config):
         super().__init__(repo, config)
@@ -73,11 +75,11 @@ class WebHDFSTree(BaseTree):
 
             if self.token is not None:
                 client = hdfs.TokenClient(  # pylint: disable=no-member
-                    http_url, self.token
+                    http_url, token=self.token, root="/"
                 )
             else:
                 client = hdfs.InsecureClient(  # pylint: disable=no-member
-                    http_url, self.path_info.user
+                    http_url, user=self.path_info.user, root="/"
                 )
 
         return client
@@ -116,11 +118,21 @@ class WebHDFSTree(BaseTree):
         self.hdfs_client.delete(path_info.path)
 
     def exists(self, path_info, use_dvcignore=True):
+        assert not isinstance(path_info, list)
+        assert path_info.scheme == "webhdfs"
+
         status = self.hdfs_client.status(path_info.path, strict=False)
         return status is not None
 
     def get_file_hash(self, path_info):
-        return self.hdfs_client.checksum(path_info.path)
+        checksum = self.hdfs_client.checksum(path_info.path)
+        print(checksum)
+        return HashInfo(self.PARAM_CHECKSUM, checksum["bytes"])
+
+    def copy(self, from_info, to_info, **_kwargs):
+        with self.hdfs_client.read(from_info.path) as reader:
+            content = reader.read()
+        self.hdfs_client.write(to_info.path, data=content)
 
     def _upload(
         self, from_file, to_info, name=None, no_progress_bar=False, **_kwargs
--- a/dvc/tree/webhdfs.py
+++ b/dvc/tree/webhdfs.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import threading
+from contextlib import contextmanager
 
 from funcy import cached_property, wrap_prop
 
@@ -57,6 +58,15 @@ class WebHDFSTree(BaseTree):
                 )
 
         return client
+
+    @contextmanager
+    def open(self, path_info, mode="r", encoding=None):
+        assert mode in {"r", "rt", "rb"}
+
+        with self.hdfs_client.read(
+            path_info.path, encoding=encoding
+        ) as reader:
+            yield reader.read()
 
     def walk_files(self, path_info, **kwargs):
         yield self.hdfs_client.walk(
--- a/dvc/tree/webhdfs.py
+++ b/dvc/tree/webhdfs.py
@@ -65,7 +65,7 @@ class WebHDFSTree(BaseTree):
         logger.debug("HDFSConfig: %s", self.hdfscli_config)
 
         try:
-            client = hdfs.config.Config(self.hdfscli_config).get_client(
+            return hdfs.config.Config(self.hdfscli_config).get_client(
                 self.alias
             )
         except hdfs.util.HdfsError as exc:
--- a/dvc/tree/webhdfs.py
+++ b/dvc/tree/webhdfs.py
@@ -134,6 +134,7 @@ class WebHDFSTree(BaseTree):
         self.hdfs_client.write(to_info.path, data=content)
 
     def move(self, from_info, to_info, mode=None):
+        self.hdfs_client.makedirs(to_info.parent.path)
         self.hdfs_client.rename(from_info.path, to_info.path)
 
     def _upload(
--- a/dvc/tree/webhdfs.py
+++ b/dvc/tree/webhdfs.py
@@ -1,6 +1,9 @@
 import logging
+import os
 import threading
+from collections import deque
 from contextlib import contextmanager
+from urllib.parse import urlparse
 
 from funcy import cached_property, wrap_prop
 
@@ -13,20 +16,44 @@ from .base import BaseTree
 logger = logging.getLogger(__name__)
 
 
+def update_pbar(pbar):
+    """Update pbar to accept the two arguments passed by hdfs"""
+
+    def update(_, bytes_transfered):  # No use for first argument (path)
+        if bytes_transfered == -1:
+            return
+        pbar.update(bytes_transfered)
+
+    return update
+
+
 class WebHDFSTree(BaseTree):
     scheme = Schemes.WEBHDFS
     PATH_CLS = CloudURLInfo
     REQUIRES = {"hdfs": "hdfs"}
+    REGEX = r"^webhdfs://((?P<user>.*)@)?.*$"
 
     def __init__(self, repo, config):
         super().__init__(repo, config)
 
+        self.path_info = None
         url = config.get("url")
-        self.path_info = self.PATH_CLS(url) if url else None
+        if not url:
+            return
+
+        parsed = urlparse(url)
+        user = parsed.username or config.get("user")
+
+        self.path_info = self.PATH_CLS.from_parts(
+            scheme="webhdfs",
+            host=parsed.hostname,
+            user=user,
+            port=parsed.port,
+            path=parsed.path,
+        )
 
         self.hdfscli_config = config.get("hdfscli_config")
         self.token = config.get("webhdfs_token")
-        self.user = config.get("user")
         self.alias = config.get("webhdfs_alias")
 
     @wrap_prop(threading.Lock())
@@ -40,15 +67,17 @@ class WebHDFSTree(BaseTree):
         try:
             client = hdfs.config.Config(  # pylint: disable=no-member
                 self.hdfscli_config
-            ).get(self.alias)
+            ).get_client(self.alias)
         except hdfs.util.HdfsError:  # pylint: disable=no-member
+            http_url = f"http://{self.path_info.host}:{self.path_info.port}"
+
             if self.token is not None:
                 client = hdfs.TokenClient(  # pylint: disable=no-member
-                    self.path_info.url, self.token
+                    http_url, self.token
                 )
             else:
                 client = hdfs.InsecureClient(  # pylint: disable=no-member
-                    self.path_info.url, self.user
+                    http_url, self.path_info.user
                 )
 
         return client
@@ -63,9 +92,22 @@ class WebHDFSTree(BaseTree):
             yield reader.read()
 
     def walk_files(self, path_info, **kwargs):
-        yield self.hdfs_client.walk(
-            path_info.path, depth=kwargs.get("depth", 0)
-        )
+        if not self.exists(path_info):
+            return
+
+        root = path_info.path
+        dir_queue = deque([root])
+
+        while dir_queue:
+            for path, dirs, files in self.hdfs_client.walk(
+                dir_queue.pop(), depth=kwargs.get("depth", 0)
+            ):
+                new_dirs = [os.path.join(path, dir_) for dir_ in dirs]
+                dir_queue.extend(new_dirs)
+
+                for file_ in files:
+                    path = os.path.join(path, file_)
+                    yield path_info.replace(path=path)
 
     def remove(self, path_info):
         if path_info.scheme != self.scheme:
@@ -84,10 +126,14 @@ class WebHDFSTree(BaseTree):
         self, from_file, to_info, name=None, no_progress_bar=False, **_kwargs
     ):
         with Tqdm(desc=name, disable=no_progress_bar, bytes=True) as pbar:
-            self.hdfs_client.upload(to_info.path, from_file, progress=pbar)
+            self.hdfs_client.upload(
+                to_info.path, from_file, progress=update_pbar(pbar)
+            )
 
     def _download(
         self, from_info, to_file, name=None, no_progress_bar=False, **_kwargs
     ):
         with Tqdm(desc=name, disable=no_progress_bar, bytes=True) as pbar:
-            self.hdfs_client.download(from_info.path, to_file, progress=pbar)
+            self.hdfs_client.download(
+                from_info.path, to_file, progress=update_pbar(pbar)
+            )
--- a/dvc/tree/webhdfs.py
+++ b/dvc/tree/webhdfs.py
@@ -32,7 +32,6 @@ class WebHDFSTree(BaseTree):
     scheme = Schemes.WEBHDFS
     PATH_CLS = CloudURLInfo
     REQUIRES = {"hdfs": "hdfs"}
-    REGEX = r"^webhdfs://((?P<user>.*)@)?.*$"
     PARAM_CHECKSUM = "checksum"
 
     def __init__(self, repo, config):
--- a/dvc/tree/webhdfs.py
+++ b/dvc/tree/webhdfs.py
@@ -1,8 +1,11 @@
 import logging
+import os
+import threading
 import shutil
 
 from funcy import cached_property, wrap_prop
 
+from dvc.path_info import CloudURLInfo
 from dvc.progress import Tqdm
 
 from .base import BaseTree
@@ -13,7 +16,7 @@ logger = logging.getLogger(__name__)
 class WebHDFSTree(BaseTree):
     PATH_CLS = CloudURLInfo
     REQUIRES = {"hdfs": "hdfs"}
-    
+
     def __init__(self, repo, config):
         super().__init__(repo, config)
 
@@ -25,7 +28,7 @@ class WebHDFSTree(BaseTree):
             or os.getenv("HDFSCLI_CONFIG")
             or "~/.hdfscli.cfg"
         )
-        
+
         self.token = (
             config.get("webhdfs_token")
         )
@@ -40,11 +43,10 @@ class WebHDFSTree(BaseTree):
     @cached_property
     def hdfs_client(self):
         import hdfs
-        
+
         logger.debug("URL: %s", self.path_info)
         logger.debug("HDFSConfig: %s", self.hdfscli_config)
 
-        
         try:
             client = hdfs.config.Config(self.hdfscli_config).get(self.alias)
         except hdfs.HdfsError:
@@ -58,7 +60,7 @@ class WebHDFSTree(BaseTree):
     def exists(self, path_info, use_dvcignore=True):
         status = self.hdfs_client.status(path_info.path, strict=False)
         return status is not None
-    
+
     def _upload(
         self, from_file, to_info, name=None, no_progress_bar=False, **_kwargs
     ):
--- a/dvc/tree/webhdfs.py
+++ b/dvc/tree/webhdfs.py
@@ -127,7 +127,10 @@ class WebHDFSTree(BaseTree):
     ):
         with Tqdm(desc=name, disable=no_progress_bar, bytes=True) as pbar:
             self.hdfs_client.upload(
-                to_info.path, from_file, progress=update_pbar(pbar)
+                to_info.path,
+                from_file,
+                overwrite=True,
+                progress=update_pbar(pbar),
             )
 
     def _download(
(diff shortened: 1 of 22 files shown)


tomasfarias:
Unsure what the linting errors mean, just noticed they also happen in my local dev environment, but overlooked them:
```
scripts/innosetup/dvc_up.bmp.dvc:
	changed outs:
		not in cache:       scripts/innosetup/dvc_up.bmp
scripts/innosetup/dvc.ico.dvc:
	changed outs:
		not in cache:       scripts/innosetup/dvc.ico
scripts/innosetup/dvc_left.bmp.dvc:
	changed outs:
		not in cache:       scripts/innosetup/dvc_left.bmp
```

Could someone guide me to fix this?

pmrowla:
> Unsure what the linting errors mean, just noticed they also happen in my local dev environment, but overlooked them:

The linting failure is from pylint, not the dvc pre-commit check:

```
pylint...................................................................Failed
- hook id: pylint
- exit code: 6

Registered custom plugin. Some checks will be disabled for tests.
************* Module tests.remotes.hdfs
tests/remotes/hdfs.py:152:8: W0621: Redefining name 'hdfs' from outer scope (line 143) (redefined-outer-name)
tests/remotes/hdfs.py:158:8: E1129: Context manager 'InsecureClient' doesn't implement __enter__ and __exit__. (not-context-manager)
...
```

tomasfarias:
@pmrowla Thanks for pointing it out, I had mistakenly used `return` instead of `yield`. 

Linting errors appeared fixed in my local environment so now I'm waiting for checks to pass 🤞

tomasfarias:
Windows py3.9 tests appeared to have been... cancelled? I'm a bit unsure what could have caused that, but I'm taking a look.
```
tests/func/test_repro.py::TestReproFrozen::test 
tests/func/test_import_url.py::test_import_url_dir[local_cloud-dc24e1271084ee317ac3c2656fb8812b-b6dcab6ccd17ca0a8bf4a215a37d14cc.dir] 
^C
Error: The operation was canceled.
```

dvc/tree/webhdfs.py
```diff
@@ -0,0 +1,89 @@
...
+    def __init__(self, repo, config):
+        super().__init__(repo, config)
+
+        url = config.get("url")
+        self.path_info = self.PATH_CLS(url) if url else None
+
+        self.hdfscli_config = (
+            config.get("hdfscli_config")
+            or os.getenv("HDFSCLI_CONFIG")
+            or "~/.hdfscli.cfg"
```
efiop, line 30: Is it the same for windows?

tests/remotes/hdfs.py
```diff
@@ -144,3 +144,62 @@ def hdfs(hdfs_server):
...
+    def read_text(self, encoding=None, errors=None):
+        if not encoding:
+            encoding = locale.getpreferredencoding(False)
+        assert errors is None
+        return self.read_bytes().decode(encoding)
+
+
+@pytest.fixture
+def webhdfs(hdfs_server):
+    port = "50070"
```
efiop, line 203: Hm, this doesn't look right. The line you've added to the docker compose file means that that port will be exposed but with a new dynamic number, so you need to ask docker for it. E.g. see how we do `port = docker_services.port_for("hdfs", 8020)` above. Maybe consider making `hdfs_server` fixture return a port map like {8020: ..., 50070: ...} instead of just port for 8020.

Btw, are there any python emulators/mockers for webhdfs? There are none for real hdfs (at least I'm not aware of them), so that's why we use real hadoop here in a docker container. Just wondering.

dvc/tree/webhdfs.py
```diff
@@ -0,0 +1,89 @@
...
+                    self.path_info.url, self.user
+                )
+
+        return client
+
+    def walk_files(self, path_info, **kwargs):
+        yield self.hdfs_client.walk(
+            path_info.path, depth=kwargs.get("depth", 0)
+        )
+
```
efiop, line 65: I guess we could add `open()` support right away as well. There is no `open` in this library, but `read()` seems to do the same thing, we just need to assert that no one is trying to use `open` with `w` flag.

setup.py
```diff
@@ -95,13 +95,14 @@ def run(self):
 
 # Remove the env marker if/when pyarrow is available for Python3.9
 hdfs = ["pyarrow>=0.17.0;  python_version < '3.9'"]
+webhdfs = ["hdfs==2.5.8"]
```
efiop, line 98: The library looks pretty good, but there are a few concerns:

1) Conda package for 2.5.8 hasn't been merged yet https://github.com/conda-forge/python-hdfs-feedstock :slightly_frowning_face: 

2) No wheels https://pypi.org/project/hdfs/#files :slightly_frowning_face:

dvc/tree/webhdfs.py
tomasfarias: The library `hdfs` calls `os.path.expanduser` to generate the appropriate default path in case nothing is passed, in particular `os.path.expanduser` takes care of dealing with Windows' paths. However, since my original approach would default to `"~/.hdfscli.cfg"`, `hdfs` never used it's own default path which called `os.path.expanduser`. This would have caused issues on Windows, like you suggested. 

I've removed the fallbacks so that `None` would be passed if nothing is set in `config`. `hdfs` already handles the default path anyways, so they were redundant.

tests/remotes/hdfs.py
tomasfarias: I think I confused `webhdfs` with a different service... It should all be the same port, just different scheme. I'm now using the same port as `hdfs`.

No emulators or mockers that I am aware unfortunately.

dvc/tree/webhdfs.py
tomasfarias: I gave it a shot at implementing `open` using `read` as suggested 😄

setup.py
tomasfarias: I agree with the concerns, unfortunately I couldn't find a library in PyPI that looked any better. The other ones I saw are in Beta or Alpha versions, and even less active than `hdfs`. A few examples: https://pypi.org/project/WebHDFS/, https://pypi.org/project/PyHDFS/, https://github.com/pywebhdfs/pywebhdfs. I'm open to switching libraries if a better one is found though (either in this PR or a different one).

I do have a use case at work where having having `webhdfs` support would allow us to get rid of `pyarrow` dependencies, and simplify our onboarding process. Not a pressing concern though, and regardless I'm also happy to help with the implementation.

efiop: Sounds good! :+1: Thank you!

tests/remotes/hdfs.py
```diff
@@ -144,3 +144,62 @@ def hdfs(hdfs_server):
...
+    def read_text(self, encoding=None, errors=None):
+        if not encoding:
+            encoding = locale.getpreferredencoding(False)
+        assert errors is None
+        return self.read_bytes().decode(encoding)
+
+
+@pytest.fixture
+def webhdfs(hdfs_server):
+    port = hdfs_server
```
efiop, line 203: You are using 8020(or the corresponding dynamic port to be correct). That doesn't seem right. Aren't you supposed to use 50070?

tomasfarias: You are right, probably caused by working late at night 😓. Latest commit gets the corresponding dynamic port for 50070 instead and has `hdfs_server` return both in a `dict`.

tomasfarias:
I'm confused about `flake8` complaining about `C408 Unnecessary dict call - rewrite as a literal.` in files not modified by this PR. Merged master changes locally and ran `pre-commit run --all-files` and got the same results.

Full error:
```
dvc/repo/__init__.py:117:23: C408 Unnecessary dict call - rewrite as a literal.
        tree_kwargs = dict(use_dvcignore=True, dvcignore_root=self.root_dir)
                      ^
tests/unit/command/test_diff.py:16:10: C408 Unnecessary dict call - rewrite as a literal.
        (dict(old="1234567890", new="0987654321"), "12345678..09876543"),
         ^
dvc/tree/s3.py:67:24: C408 Unnecessary dict call - rewrite as a literal.
        session_opts = dict(profile_name=self.profile, region_name=self.region)
                       ^
tests/func/test_tree.py:274:10: C408 Unnecessary dict call - rewrite as a literal.
    kw = dict(ignore_subrepos=False)
         ^
dvc/progress.py:34:22: C408 Unnecessary dict call - rewrite as a literal.
    BYTES_DEFAULTS = dict(
```

tests/docker-compose.yml
```diff
@@ -11,9 +11,12 @@ services:
...
   hdfs:
     image: rkuprieiev/docker-hdfs
+    hostname: localhost
     ports:
       - "8020"
       # NOTE: having this port as dynamic one will require modifying
       # `dfs.datanode.address` in `hdfs-site.xml` and probably something
       # else, so using default one for now.
       - "50010:50010"
+      - "50075:50075"
```
efiop, line 21: Do we need this one?

tomasfarias: We kind of do: when creating a file `hdfs` makes an initial `PUT` request to port `50070`. This request is answered with a temporary redirect to the datanode listening in port `50075`, not the dynamic port. So without having the port as static `hdfs` will fail when following the redirect, as it will attempt to request `50075`.

efiop (changes requested):
Looks great! :fire: 

We are missing some tests for the functionality that is added though. E.g. open/get_file_hash/read_text and stuff like that are the things that need to be tested in things like import-url, get-url, api (just see where `hdfs` fixture is used in the test suite), so far only the push/pull is really tested in any way.

tomasfarias:
> 
> 
> Looks great! 🔥
> 
> We are missing some tests for the functionality that is added though. E.g. open/get_file_hash/read_text and stuff like that are the things that need to be tested in things like import-url, get-url, api (just see where `hdfs` fixture is used in the test suite), so far only the push/pull is really tested in any way.

Thanks for the review @efiop. I agree that more testing is needed, I had a bit of trouble navigating the test suite which is why I only focused on what was mentioned explicitly in #4687. I have been busy with my day job, but I will have enough time to complete testing by adding your suggestions and also doing some documentation work before next week 💪

tomasfarias:
`webhdfs` fixture should now be added to remaining functionality tests: basically wherever I saw the `hdfs` fixture being used, I added `webhdfs` too. Making tests pass required implementing `get_file_hash` and `copy` in `WebHDFSTree` which were missing.

tomasfarias:
A couple of Windows tests are failing. Seems unrelated to the changes in this PR because tests are not particularly using `webhdfs`, and it seems other PRs are also getting similar failures.

efiop:
Inded, windows test failures are unrelated, we are working on solving that in the upstream. Thank you for your patience!

dvc/tree/webhdfs.py
```diff
@@ -0,0 +1,153 @@
...
+        pbar.update(bytes_transfered)
+
+    return update
+
+
+class WebHDFSTree(BaseTree):
+    scheme = Schemes.WEBHDFS
+    PATH_CLS = CloudURLInfo
+    REQUIRES = {"hdfs": "hdfs"}
+    REGEX = r"^webhdfs://((?P<user>.*)@)?.*$"
```
efiop, line 35: Ah, this is just a leftover. Same in hdfs tree.
```suggestion
```

dvc/tree/webhdfs.py
```diff
@@ -0,0 +1,152 @@
...
+        import hdfs
+
+        logger.debug("URL: %s", self.path_info)
+        logger.debug("HDFSConfig: %s", self.hdfscli_config)
+
+        try:
+            client = hdfs.config.Config(  # pylint: disable=no-member
+                self.hdfscli_config
+            ).get_client(self.alias)
+        except hdfs.util.HdfsError:  # pylint: disable=no-member
```
efiop, line 72: can we be more specific about the error that we catch here?

dvc/tree/webhdfs.py
```diff
@@ -0,0 +1,152 @@
...
+    @wrap_prop(threading.Lock())
+    @cached_property
+    def hdfs_client(self):
+        import hdfs
+
+        logger.debug("URL: %s", self.path_info)
+        logger.debug("HDFSConfig: %s", self.hdfscli_config)
+
+        try:
+            client = hdfs.config.Config(  # pylint: disable=no-member
```
efiop, line 69: Pylint is confused by hdfs library, right? Maybe better to add it to the pylint config instead of manually disabling in each line?

dvc/tree/webhdfs.py
```diff
@@ -0,0 +1,152 @@
...
+    def walk_files(self, path_info, **kwargs):
+        if not self.exists(path_info):
+            return
+
+        root = path_info.path
+        dir_queue = deque([root])
+
+        while dir_queue:
+            for path, dirs, files in self.hdfs_client.walk(
+                dir_queue.pop(), depth=kwargs.get("depth", 0)
```
efiop, line 104: I don't think `depth` is ever used anywhere

dvc/tree/webhdfs.py
```diff
@@ -0,0 +1,152 @@
...
+            ):
+                new_dirs = [os.path.join(path, dir_) for dir_ in dirs]
+                dir_queue.extend(new_dirs)
```
efiop, line 107: Are you sure we need this? From the first glance seems like hdfs's `walk` already does everything for us.

dvc/tree/webhdfs.py
```diff
@@ -0,0 +1,152 @@
...
+        status = self.hdfs_client.status(path_info.path, strict=False)
+        return status is not None
+
+    def get_file_hash(self, path_info):
+        checksum = self.hdfs_client.checksum(path_info.path)
+        return HashInfo(self.PARAM_CHECKSUM, checksum["bytes"])
+
+    def copy(self, from_info, to_info, **_kwargs):
+        with self.hdfs_client.read(from_info.path) as reader:
+            content = reader.read()
```
efiop, line 132: Is this reading everything into the memory?

dvc/tree/webhdfs.py
tomasfarias: Good call, I can just remove it then 👍

dvc/tree/webhdfs.py
tomasfarias: It made some sense if using `depth != 0`, but like you clarified we are not using it anywhere, so it should be safe to remove this.

dvc/tree/webhdfs.py
tomasfarias: I did just that. Saves 4 disable comments.

dvc/tree/webhdfs.py
tomasfarias: Tried adding a comment clarifying what we expect to fail there. Is that clear enough? Should some logging be added to?

dvc/tree/webhdfs.py
```diff
@@ -0,0 +1,143 @@
...
+            # No hdfs config file was found, or parsing failed. Fallback to
+            # clients using Hadoop delegation token or username
+            http_url = f"http://{self.path_info.host}:{self.path_info.port}"
+
+            if self.token is not None:
+                client = hdfs.TokenClient(http_url, token=self.token, root="/")
+            else:
+                client = hdfs.InsecureClient(
+                    http_url, user=self.path_info.user, root="/"
+                )
```
efiop, line 81: From https://github.com/iterative/dvc/pull/4747#discussion_r517441658 

You are still catching a generic error, is there a specific exc that is raised for the config error?

dvc/tree/webhdfs.py
efiop: See https://github.com/iterative/dvc/pull/4747#discussion_r519189887

dvc/tree/webhdfs.py
tomasfarias: I see what you mean. Not really, `hdfs` only wraps errors in `HDFSError`s, see: https://github.com/mtth/hdfs/blob/master/hdfs/config.py#L76

efiop: @tomasfarias :( Ok, can we check the exc message just to make sure we are not misinterpreting the error?

dvc/tree/webhdfs.py
```diff
@@ -0,0 +1,143 @@
...
+            path=parsed.path,
+        )
+
+        self.hdfscli_config = config.get("hdfscli_config")
+        self.token = config.get("webhdfs_token")
+        self.alias = config.get("webhdfs_alias")
+
+    @wrap_prop(threading.Lock())
+    @cached_property
+    def hdfs_client(self):
```
efiop, line 61: Btw, should we use connection pool like we do in regular hdfs? (see get_connection)

dvc/tree/webhdfs.py
```diff
@@ -0,0 +1,143 @@
...
+        import hdfs
+
+        logger.debug("URL: %s", self.path_info)
+        logger.debug("HDFSConfig: %s", self.hdfscli_config)
+
+        try:
+            client = hdfs.config.Config(self.hdfscli_config).get_client(
+                self.alias
+            )
```
efiop, line 70: Really minor: Let's just return here and below, since we are not really doing anything with the `client`.
```suggestion
            return hdfs.config.Config(self.hdfscli_config).get_client(
                self.alias
            )
```

efiop:
@tomasfarias Btw, about the docs. Could you please submit a doc PR? It will be much easier that way, since you are already deep into hdfscli usage.

dvc/tree/webhdfs.py
tomasfarias: I'm not too sure... `hdfs.client.Client` doesn't really have a `close` method. We could pool the underlying `request.Session` object, but only if using `InsecureClient` and `TokenClient` directly (which happens in the `except` block), since `hdfs.config.Config.get_client` instantiates a client using the configuration file, no `session` argument is exposed.

dvc/tree/webhdfs.py
tomasfarias: Gave it a shot. Really would have preferred different exception types since there's 3 potential error messages 😢

tomasfarias:
@efiop Will work on docs next 👍 

As a sidenote:

Ubuntu tests for python<3.9 are failing. Raw logs show there was an issue instantiating the Hadoop file system in S3, which shouldn't be related to this PR. Example error:

```
2020-11-07T17:06:20.3943432Z java.util.ServiceConfigurationError: org.apache.hadoop.fs.FileSystem: Provider org.apache.hadoop.fs.s3a.S3AFileSystem could not be instantiated
2020-11-07T17:06:20.3945203Z 	at java.util.ServiceLoader.fail(ServiceLoader.java:224)
2020-11-07T17:06:20.3946217Z 	at java.util.ServiceLoader.access$100(ServiceLoader.java:181)
2020-11-07T17:06:20.3947276Z 	at java.util.ServiceLoader$LazyIterator.next(ServiceLoader.java:377)
2020-11-07T17:06:20.3948233Z 	at java.util.ServiceLoader$1.next(ServiceLoader.java:445)
2020-11-07T17:06:20.3949500Z 	at org.apache.hadoop.fs.FileSystem.loadFileSystems(FileSystem.java:2623)
2020-11-07T17:06:20.3950943Z Caused by: java.lang.NoClassDefFoundError: com/amazonaws/AmazonServiceException
```

dvc/tree/webhdfs.py
```diff
@@ -0,0 +1,153 @@
...
+        checksum = self.hdfs_client.checksum(path_info.path)
+        hash_info = HashInfo(self.PARAM_CHECKSUM, checksum["bytes"])
+
+        hash_info.size = self.hdfs_client.status(path_info.path)["length"]
+        return hash_info
+
+    def copy(self, from_info, to_info, **_kwargs):
+        with self.hdfs_client.read(from_info.path) as reader:
+            content = reader.read()
+        self.hdfs_client.write(to_info.path, data=content)
```
efiop, line 134: WebHDFS doesn't have a serverside copy or just our lib doesn't support it?

And one more thing, let's add `move` implementation too through hdfs.rename. That will speed up this advanced usecase.

efiop (changes requested):
Looks great! One more comment above. And please rebase on top of master. I think we will be ready to merge after that.

dvc/tree/webhdfs.py
tomasfarias: WebHDFS doesn't have serverside copy. All operations listed here: https://hadoop.apache.org/docs/r1.0.4/webhdfs.html#Operations.

Added a move implementation via rename

dvc/tree/webhdfs.py
```diff
@@ -0,0 +1,157 @@
...
+            )
+        except hdfs.util.HdfsError as exc:
+            exc_msg = str(exc)
+            errors = (
+                "No alias specified",
+                "Invalid configuration file",
+                f"Alias {self.alias} not found",
+            )
+            if not any(err in exc_msg for err in errors):
+                raise
```
efiop, line 79: Let's leave a comment why we do such a weird thing. Also, this might be a good PR/issue for hdfs library :wink:

efiop (approved):
Looks good! Some minor stuff, but we can handle it later on top.

Btw, I assume you've tried pushing/pulling some big datasets/datafiles and progress bars work? :slightly_smiling_face:

tomasfarias:
@efiop Progress bars worked but without `total` they were looking kind of wonky 😅 I hope you don't mind if I push one last commit to fix them, sorry for dismissing your review.

At least now I have a gif:
dvc_upload

efiop:
@tomasfarias Is this still WIP? If not, please remove the prefix.

efiop:
Thank you @tomasfarias ! :pray:

Issue #4746, opened 2020-10-19
dvc checkout --ignore-missing;  build workspace even if some data is missing in local cache


DanielHiggs:
Currently 'dvc checkout' will fail with an error if the local cache does not contain all of the data specified by the repository and it will not construct any workspace directories.

For situations when having a partial subset of the repository's data is still useful, it would be nice to have an '--ignore-missing' option for the checkout command. When this option is invoked, 'dvc checkout' will create a workspace containing links to only those files which exist in the local cache, while only warning the user of the files that are unavailable instead of producing an error.

pmrowla:
Worth nothing that we likely need this same behavior internally for checkpoint experiments, so it would not be too difficult to just add the `dvc checkout` command option once the checkpoints work is completed.

dberenbaum:
@pmrowla Now that checkpoint experiments have been implemented for awhile, what is the level of effort to add this?

christian-steinmeyer:
Just to clarify: When using the proposed `--ignore-missing` flag, warnings about missing files would still be printed, right?

If so, I'd propose to also make this option configurable as per config file, e.g.:
```
[core]
    ignore_missing = true
```

frerksaxen:
I'd love to have this option since I usually add multiple stages in my dvc.yaml file, where outputs of one stage are dependencies of other stages. Calling `dvc pull` for the first time after defining these additional stages will fail because some dependencies have simply not yet been created at all. The `--ignore-missing` flag would really help in this situation, because it is not clear if `dvc pull` has finished pulling all available dependencies or if it just stopped at the first dependency it couldnt find.

hemker:
I would also like to see this feature, because it simplifies the reproducibility of a partially executed pipeline on a remote server, for instance with Gitlab. Otherwise there would be the need to heavily customize the `.gitlab-ci.yaml` files for specific target pulling and force-reproducing.

dberenbaum:
> @pmrowla Now that checkpoint experiments have been implemented for awhile, what is the level of effort to add this?

@pmrowla Following up on this. How much effort is it to add `--ignore-missing`?

PR #4750, opened 2020-10-20
checkpoints: allow missing outputs on checkout and stage.run

### LINKED ISSUE #4707: checkpoints: combining persist and no-persist outputs does not work as expected
```
$ dvc exp run --continue ca9b885
...
WARNING: Cache 'HashInfo(name='md5', value='241381b47884cd995b0b1cc57c737e50.dir', dir_info=None)' not found. File 'logs' won't be created.
WARNING: Cache 'HashInfo(name='md5', value='7969e8c1d96502dce39bd3c6d69feb40', dir_info=None)' not found. File 'model.h5' won't be created.
ERROR: Checkout failed for following targets:
logs
Is your cache up to date?
```

```
stages:
  train:
    cmd: python train.py
    deps:
    - users.csv
    - train.py
    - mymodel.py
    params:
    - epochs
    - learning_rate
    - log_file
    - dvc_logs_dir
    - dropout
    outs:
    - logs
    - model.h5:
        persist: true
    metrics:
    - summary.json:
        cache: false
    plots:
    - logs.csv:
        cache: false
    always_changed: true
```

Easiest fix would be to just require that all outputs be persist for checkpoints, but I need to look into whether there is a better workaround

pmrowla:

Will close #4707, may also be related to #4746.

* Adds support for allowing missing outputs in `repo.checkout` and `stage.run` (previously we supported missing `persist` outs for checkpoints, but we actually need this for regular outs as well).

For checkpoint experiments, we want to support the potential use case where a stage may contain both intermediate outputs (that persist) and regular (non-persist) outputs. Depending on when the checkpoint run is interrupted, the regular outputs may not have been created yet, but this should not be considered a failure state for checkpoint experiments.

### DIFF
--- a/dvc/output/base.py
+++ b/dvc/output/base.py
@@ -282,6 +282,9 @@ class BaseOutput:
         self.hash_info = self.get_hash()
 
     def commit(self):
+        if not self.exists:
+            raise self.DoesNotExistError(self)
+
         assert self.hash_info
         if self.use_cache:
             self.cache.save(self.path_info, self.cache.tree, self.hash_info)
@@ -326,7 +329,7 @@ class BaseOutput:
         progress_callback=None,
         relink=False,
         filter_info=None,
-        allow_persist_missing=False,
+        allow_missing=False,
     ):
         if not self.use_cache:
             if progress_callback:
@@ -345,7 +348,7 @@ class BaseOutput:
                 filter_info=filter_info,
             )
         except CheckoutError:
-            if self.persist and allow_persist_missing:
+            if allow_missing:
                 return None
             raise
 
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -8,6 +8,7 @@ from funcy import cached_property, project
 import dvc.dependency as dependency
 import dvc.prompt as prompt
 from dvc.exceptions import CheckoutError, DvcException, MergeError
+from dvc.output.base import OutputDoesNotExistError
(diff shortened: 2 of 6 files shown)


dvc/output/base.py
```diff
@@ -282,6 +282,9 @@ def save(self):
         self.hash_info = self.get_hash()
 
     def commit(self):
+        if not self.exists:
+            raise self.DoesNotExistError(self)
```
pmrowla, line 286: Unsure if it would be better for us to skip this exists call for performance reasons, and just catch (and allow) any exception on `commit()` in `Stage.run()` when we want `allow_exists` behavior?

dvc/stage/__init__.py
```diff
@@ -416,12 +421,23 @@ def changed_entries(self):
...
     @rwlocked(write=["outs"])
-    def commit(self):
+    def commit(self, allow_missing=False):
         for out in self.outs:
-            out.commit()
+            try:
+                out.commit()
+            except OutputDoesNotExistError:
+                if not allow_missing:
+                    raise
```
pmrowla, line 430: see related comment on `out.commit()`

dvc/output/base.py
efiop: Yeah, but the problem is that tree exceptions are not unified yet, so it will be all over the place :slightly_frowning_face: But since you are only working with local trees, you might skip this check and just catch FileNotFoundError even right now. Up to you.

PR #4751, opened 2020-10-20
Support parametrization from params or vars section in dvc.yaml

skshetry:

This  PR was splitted from https://github.com/iterative/dvc/pull/4734, as it became quite large without the tests. I have also removed `foreach` and `set` support from there to reduce scope.

This PR is also quite large, but the Context is well-tested and is about more than half of it.  Though, I have left the tests for the newer syntax, which will come in separate PRs.


skshetry:
Merged witout review, as I don't expect anyone to go through this large piece of code. Also, it's very core to the next iteration, and din't think  of how I could split this  (as this is already split).

PR #4758, opened 2020-10-21
checkpoints: `--continue` from last run experiment by default

### LINKED ISSUE #4756: checkpoints: `--continue` with no arguments should resume the most recent experiment
Now that branches are sortable by most recent commit, this just requires a command update

pmrowla:

Will close #4756

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -710,12 +710,16 @@ def add_parser(subparsers, parent_parser):
     experiments_run_parser.add_argument(
         "--continue",
         type=str,
+        nargs="?",
         default=None,
+        const=":last",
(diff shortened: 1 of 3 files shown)


dvc/command/experiments.py
```diff
@@ -710,12 +710,16 @@ def add_parser(subparsers, parent_parser):
     experiments_run_parser.add_argument(
         "--continue",
         type=str,
+        nargs="?",
         default=None,
+        const=":last",
```
pmrowla, line 715: Use `:` prefix to ensure we can't collide with a valid branch or tag name (currently branch names are only auto-generated but we may support user named branches or tags in the future)

PR #4760, opened 2020-10-21
Display coverage report on the GHA log and measure branch coverage

skshetry:

This prints coverage report on Github Actions log, and also enables measuring of branch coverage (eg: `if` conditionals).
Our coverage may decrease, but it will provide us better sense of our coverage.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -50,7 +50,7 @@ jobs:
         git config --global user.email "dvctester@example.com"
         git config --global user.name "DVC Tester"
     - name: run tests
-      run: python -m tests -ra --cov-report=xml
+      run: python -m tests -ra --cov-report=xml --cov-report=term
     - name: upload coverage report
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yaml
```diff
@@ -44,13 +44,14 @@ jobs:
         python-version: ${{ matrix.pyv }}
     - name: install
       run: |
-        pip install ".[all,tests]"
+        pip install wheel
+        pip install -e ".[all,tests]"
```
skshetry, line 48: Coverage was being counted from both local directory and the installed package before, thus reducing the coverage. Also, it's a good practice to use local directory for running the tests (hence, the _editable_ mode).

`pip install wheel`, because I saw a lot of packages were being installed on legacy mode.

pared (approved):

efiop:
Thank you @skshetry ! :pray:

Issue #4520, opened 2020-09-02
support adding/transfering data straight to cache/remote 


efiop:
We often see people trying to use `--external` to add some big dataset that they have on an external drive, where they also have an external cache dir. People often do that because they can't/don't want to copy their data to dvc repo to `dvc add` it normally, e.g. because their HDD/SSD won't be able to physically fit two copies of that dataset.

Same thing with s3/gs/etc, where people want to just move data straight to their remote, without having to download/add/push it, because, again, it might not even fit on their local machine.

That's why it would be great to introduce feature(s) to be able to move(or copy) data straight to cache/remote from it's original location. Potentially this is not only useful for `dvc add` but also for `dvc import[-url]`, where you want to use some data (e.g. through streaming with our API) in your project, that won't fit on your machine.

Related to https://github.com/iterative/dvc/issues/3920

jorgeorpinel:
Would this also solve #3305?

jorgeorpinel:
And yes, this feature would be useful. It sounds like a "lightweight" data registry (see https://dvc.org/doc/use-cases/data-registries) for people to use DVC as an intermediary between data already hosted on cloud storage and other projects. So they can access the data with a unified interface (DVC CLI or API).

efiop:
> Would this also solve #3305?

No, it won't. These are unrelated. Unless I'm missing something.

tomasfarias:
Hey all, I'll be tackling this issue next 👀

Let me know if I understand the problem clearly: although datasets in external drives/s3/etc can be `dvc add`ed with `--external` and later `dvc push`ed to a remote, we would like to streamline the process via a new feature(s). 

From the users perspective, I think something like a `dvc transfer` command that combines a download of the data, `dvc add` and `dvc push` would work well. Of course, under the hood we would not actually download the whole dataset, to avoid size issues as @efiop described.

An initial idea I had involved downloading chunks of a dataset at a time to avoid having to download the whole dataset. I'm aware at least s3 and WebHDFS support iterating over chunks as [`botocore`](https://github.com/boto/botocore) has an [`iter_chunks`](https://github.com/boto/botocore/blob/master/botocore/response.py#L104) method for it, and `HdfsCLI`'s `Client.read` supports a [`chunk_size`](https://github.com/mtth/hdfs/blob/master/hdfs/client.py#L642) argument. 

[...]

PR #4778, opened 2020-10-23
add: fix issue when adding already tracked symlinked files

### LINKED ISSUE #4654: Cannot dvc add an already tracked file that lies in a folder
## Bug Report

**dvc version:** 1.8.0

----**Use case problem**----
So I have folder that looks like that:
```
.
├── Annotations.json -> /media/newhdd/data/ki67_er_pr/data/dvc_cache/44/7b62afed38955cb2a2f4ca35c5133c
├── Annotations.json.dvc
└── annotator
    ├── ki67_20x__perez_2019_04_12__7__1___Annotation_0_hotspot_0.json -> /media/newhdd/data/ki67_er_pr/data/dvc_cache/12/6285610f25480393ca3e153378352b
    ├── ki67_20x__perez_2019_04_12__7__1___Annotation_0_hotspot_0.json.dvc
    ├── ki67_20x__perez_2019_04_12__7__1___Annotation_0_hotspot_1.json -> /media/newhdd/data/ki67_er_pr/data/dvc_cache/6b/303f04558c946332e3c746781ac940
    └── ki67_20x__perez_2019_04_12__7__1___Annotation_0_hotspot_1.json.dvc

1 directory, 6 files
```

When I run `dvc add annotator/ki67_20x__perez_2019_04_12__7__1___Annotation_0_hotspot_0.json` -> I get 
```ERROR: output 'annotator/ki67_20x__perez_2019_04_12__7__1___Annotation_0_hotspot_0.json' is already specified in stage: 'annotator/ki67_20x__perez_2019_04_12__7__1___Annotation_0_hotspot_0.json.dvc'.```  :exclamation: 

But when I do `cd annotator && dvc add ki67_20x__perez_2019_04_12__7__1___Annotation_0_hotspot_0.json` -> I get correctly :white_check_mark: 
```
Stage is cached, skipping                                                                                                                
  0% Add|                                                                                                       |0/0 [00:02,     ?file/s]
```
And the process is skipped.

If I `cat ki67_20x__perez_2019_04_12__7__1___ Annotation_0_hotspot_0.json.dvc` I get:
```
outs:
- md5: 126285610f25480393ca3e153378352b
  path: ki67_20x__perez_2019_04_12__7__1___Annotation_0_hotspot_0.json
```

### PR DESCRIPTION
**Pull request by pmrowla**

Will fix #4654.

* `dvc add <symlinked_file>` now works as expected. The DVC file will be placed in the same directory as the original symlink
* `dvc add`'ing symlinked a directory or a file inside a symlinked directory is now forbidden.

Docs PR: https://github.com/iterative/dvc.org/pull/1890

### DIFF
--- a/dvc/utils/__init__.py
+++ b/dvc/utils/__init__.py
@@ -372,9 +372,15 @@ def resolve_paths(repo, out):
     elif contains_symlink_up_to(dirname, repo.root_dir) or (
         os.path.isdir(abspath) and System.is_symlink(abspath)
     ):
-        raise DvcException(
-            "Cannot add files inside symlinked directories to DVC."
+        msg = (
+            "Cannot add files inside symlinked directories to DVC. "
+            "See {} for more information."
+        ).format(
+            format_link(
+                "https://dvc.org/doc/user-guide/troubleshooting#add-symlink"
+            )
         )
+        raise DvcException(msg)
     else:
         wdir = dirname
         out = base
--- a/dvc/utils/__init__.py
+++ b/dvc/utils/__init__.py
@@ -353,7 +353,9 @@ def resolve_paths(repo, out):
     from urllib.parse import urlparse
 
     from ..dvcfile import DVC_FILE_SUFFIX
+    from ..exceptions import DvcException
     from ..path_info import PathInfo
+    from ..system import System
     from .fs import contains_symlink_up_to
 
     abspath = PathInfo(os.path.abspath(out))
@@ -364,15 +366,18 @@ def resolve_paths(repo, out):
     if os.name == "nt" and scheme == abspath.drive[0].lower():
         # urlparse interprets windows drive letters as URL scheme
         scheme = ""
-    if (
-        not scheme
-        and abspath.isin_or_eq(repo.root_dir)
-        and not contains_symlink_up_to(dirname, repo.root_dir)
+
+    if scheme or not abspath.isin_or_eq(repo.root_dir):
+        wdir = os.getcwd()
+    elif contains_symlink_up_to(dirname, repo.root_dir) or (
+        os.path.isdir(abspath) and System.is_symlink(abspath)
     ):
+        raise DvcException(
+            "Cannot add files inside symlinked directories to DVC."
+        )
+    else:
         wdir = dirname
         out = base
-    else:
-        wdir = os.getcwd()
 
     path = os.path.join(wdir, base + DVC_FILE_SUFFIX)
 
--- a/dvc/utils/__init__.py
+++ b/dvc/utils/__init__.py
@@ -367,7 +367,7 @@ def resolve_paths(repo, out):
     if (
         not scheme
         and abspath.isin_or_eq(repo.root_dir)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: dvc/utils/__init__.py [hunk 1]
```diff
@@ -367,7 +367,7 @@ def resolve_paths(repo, out):
     if (
         not scheme
         and abspath.isin_or_eq(repo.root_dir)
-        and not contains_symlink_up_to(abspath, repo.root_dir)
+        and not contains_symlink_up_to(dirname, repo.root_dir)
```
pmrowla, line 370: We want to place the `.dvc` files into `cwd()` when the path leading up to the output contains a link, rather than when the output itself is a link, so we only need to check up to `dirname` here rather than `abspath`.

If the output itself is a symlink, it should not matter to DVC, as we will end up copying the file contents into DVC cache, and then replacing the original link with our own link or copy to the DVC cache entry. In that case, it is still safe for us to place the `.dvc` into the same directory as the original symlink.

tests/func/test_add.py
```diff
@@ -827,3 +829,26 @@ def test_add_symlink(tmp_dir, dvc):
     assert not (
         tmp_dir / ".dvc" / "cache" / "37" / "b51d194a7513e45b56f6524f2d51f2"
     ).is_symlink()
+
+    # Test that subsequent add succeeds
+    # See https://github.com/iterative/dvc/issues/4654
+    dvc.add(os.path.join("dir", "foo"))
```
pared, line 835: +1

pared (approved):
LGTM!

tests/func/test_add.py
```diff
@@ -422,52 +421,6 @@ def test_should_collect_dir_cache_only_once(mocker, tmp_dir, dvc):
...
-
-
-class TestShouldAddDataFromInternalSymlink(SymlinkAddTestBase):
-    def _get_data_dir(self):
-        return self.DATA_DIR
-
-    def test(self):
-        self._test()
-
-
```
pmrowla, line 470: These cases are covered by the new pytest based tests

pmrowla:
After discussion with @efiop, the prior behavior regarding adding symlinked dirs or files inside symlinked dirs was too ambiguous and is now forbidden. This use-case should instead be handled by implementing proper support for adding external data directly to cache - see #4520.

dvc/utils/__init__.py
```diff
@@ -364,15 +366,18 @@ def resolve_paths(repo, out):
...
-        and abspath.isin_or_eq(repo.root_dir)
-        and not contains_symlink_up_to(abspath, repo.root_dir)
+
+    if scheme or not abspath.isin_or_eq(repo.root_dir):
+        wdir = os.getcwd()
+    elif contains_symlink_up_to(dirname, repo.root_dir) or (
+        os.path.isdir(abspath) and System.is_symlink(abspath)
     ):
+        raise DvcException(
+            "Cannot add files inside symlinked directories to DVC."
```
efiop, line 376: Looks like we should make it a bit more helpful. Maybe a link to troubleshooting doc?

efiop (approved):
Thank you! :pray:

PR #4769, opened 2020-10-21
Experiments: allow setting of list parameters

sjawhar:
* [x] ❗ I have followed the [Contributing to DVC](https://dvc.org/doc/user-guide/contributing/core) checklist.

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here.

Thank you for the contribution - we'll try to review it as soon as possible. 🙏

Will close #4768

As stated in the issue, I'm sure more work will be needed, but this little tweak was enough to at least allow the experiment to run. Some more work would be required to get parameter diffs to show up nicely in `dvc exp show` (screenshot of current display is included below, see the change in activation function), but that might be out of scope for this issue.

image

If there are no major objections, I can go ahead and write up some tests and finish this PR off proper-like.

PR #4780, opened 2020-10-23
Simpler param updates with python-benedict

sjawhar:

## Description of Changes
Following up on my last [comment](https://github.com/iterative/dvc/pull/4769#issuecomment-714869426) on #4769 

So I did some testing and my idea to use [benedict](https://github.com/fabiocaccamo/python-benedict) to simplify the code actually works great... ~almost~. (**Update:** figured it out) As you can see in the changes here, you can replace a lot of code with a single built-in function call. This also has the benefit of using more standard `[0]` syntax for list accessing.

~I added more tests to make sure that everything was working, then I ran into this strange behavior on a single test case:~
image

~The params are being correctly updated and written, but for some reason the experiment doesn't seem to actually run the copy step. It's strange, and I would appreciate any help chasing this down...~

I just needed to add `goo` to the list of stage params :facepalm: 

Linting is failing on changes I didn't make.

## TODO
- [x] What's up with that weird behavior?!
- [x] Run the full test suite to make sure I didn't break anything elsewhere :smile:

### DIFF
--- a/dvc/repo/experiments/__init__.py
+++ b/dvc/repo/experiments/__init__.py
@@ -312,31 +312,18 @@ class Experiments:
 
     def _update_params(self, params: dict):
         """Update experiment params files with the specified values."""
+        from benedict import benedict
+
         from dvc.utils.serialize import MODIFIERS
 
         logger.debug("Using experiment params '%s'", params)
-
-        # recursive dict update
-        def _update(dict_, other):
-            for key, value in other.items():
-                if isinstance(dict_, list) and key.isdigit():
-                    key = int(key)
-                if isinstance(value, Mapping):
-                    if isinstance(dict_, list):
-                        fallback_value = dict_[key]
-                    else:
-                        fallback_value = dict_.get(key, {})
-                    dict_[key] = _update(fallback_value, value)
-                else:
-                    dict_[key] = value
-            return dict_
 
         for params_fname in params:
             path = PathInfo(self.exp_dvc.root_dir) / params_fname
             suffix = path.suffix.lower()
             modify_data = MODIFIERS[suffix]
             with modify_data(path, tree=self.exp_dvc.tree) as data:
-                _update(data, params[params_fname])
+                benedict(data).merge(params[params_fname], overwrite=True)
 
         # Force params file changes to be staged in git
         # Otherwise in certain situations the changes to params file may be
--- a/dvc/repo/experiments/run.py
+++ b/dvc/repo/experiments/run.py
@@ -12,7 +12,6 @@ def _parse_params(path_params: Iterable):
     from ruamel.yaml import YAMLError
 
     from dvc.dependency.param import ParamsDependency
-    from dvc.utils.flatten import unflatten
     from dvc.utils.serialize import loads_yaml
 
     ret = {}
@@ -31,7 +30,7 @@ def _parse_params(path_params: Iterable):
                 )
         if not path:
             path = ParamsDependency.DEFAULT_PARAMS_FILE
-        ret[path] = unflatten(params)
+        ret[path] = params
     return ret
 
 
--- a/tests/func/experiments/test_experiments.py
+++ b/tests/func/experiments/test_experiments.py
@@ -66,27 +66,45 @@ def test_update_with_pull(tmp_dir, scm, dvc, exp_stage, mocker):
 
 
 @pytest.mark.parametrize(
-    "change, expected",
+    "changes, expected",
     [
-        ["foo.1.baz=3", "foo: [bar: 1, baz: 3]"],
-        ["foo.0=bar", "foo: [bar, baz: 2]"],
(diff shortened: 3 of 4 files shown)


dvc/repo/experiments/run.py
```diff
@@ -31,7 +30,7 @@ def _parse_params(path_params: Iterable):
                 )
         if not path:
             path = ParamsDependency.DEFAULT_PARAMS_FILE
-        ret[path] = unflatten(params)
+        ret[path] = params
```
sjawhar, line 33: To be clear, this now returns a dict like:
```
{'path.to.key': yamlValue1, 'path[2]key': yamlValue2}
```

dvc/repo/experiments/__init__.py
```diff
@@ -311,31 +310,18 @@ def _unpack_args(self, tree=None):
 
     def _update_params(self, params: dict):
         """Update experiment params files with the specified values."""
+        from benedict import benedict
```
sjawhar, line 313: Not sure if you prefer local or global imports. Since this is only needed for the beta experiments feature, I figured local was appropriate.

tests/func/experiments/test_experiments.py
```diff
@@ -94,27 +94,33 @@ def test_update_with_pull(tmp_dir, scm, dvc, mocker):
...
-    "change, expected",
+    "changes, expected",
     [
-        ["foo.1.baz=3", "foo: [bar: 1, baz: 3]"],
-        ["foo.0=bar", "foo: [bar, baz: 2]"],
-        ["foo.1=- baz\n- goo", "foo: [bar: 1, [baz, goo]]"],
+        [["foo=baz"], "{foo: baz, goo: {bag: 3}}"],
+        [["foo=baz,goo=bar"], "{foo: baz, goo: bar}"],
+        [["goo.bag=4"], "{foo: [bar: 1, baz: 2], goo: {bag: 4}}"],
+        [["foo[0]=bar"], "{foo: [bar, baz: 2], goo: {bag: 3}}"],
```
sjawhar, line 102: Note the change in syntax, which I think is more in line with [your future plans](https://github.com/iterative/dvc/wiki/Parametrization#expression-syntax).

pmrowla:
This is very cool, thanks for the PR!

Since this will add a dependency I want to discuss this internally with the rest of the DVC team first before merging it, but everything looks good to me for now.

tests/func/experiments/test_experiments.py
```diff
@@ -94,27 +94,45 @@ def test_update_with_pull(tmp_dir, scm, dvc, mocker):
...
+            ["goo.bag=4"],
+            "{foo: [bar: 1, baz: 2], goo: {bag: 4}, lorem: false}",
+        ],
+        [["foo[0]=bar"], "{foo: [bar, baz: 2], goo: {bag: 3}, lorem: false}"],
+        [
+            ["foo[1].baz=3"],
+            "{foo: [bar: 1, baz: 3], goo: {bag: 3}, lorem: false}",
+        ],
+        [
+            ["foo[1]=- baz\n- goo"],
```
skshetry, line 111: This syntax is not easy to understand/use. At a first glance, I associated those `=-` together and assumed it was removing items from the list.

Ideally, I'd like to have the following syntax:
```console
$ dvc exp run --params "foo[1]=[baz, goo]" "bar={a:2, b: 3}"
```

I understand that we split it up by comma and achieve that. We should consider using a different delimiter, maybe even space and make it look json-like.

skshetry: It seems a natural progression to support appending and removing similar to [Hydra's syntax](https://hydra.cc/docs/next/advanced/override_grammar/basic/) (they use `~foo` to remove and `+foo=value` to append). 

Or, maybe even use Hydra later directly.

skshetry (approved):

tests/func/experiments/test_experiments.py
sjawhar: Note that I didn't change anything regarding the syntax. This is the syntax already accepted by the yaml parser being used, which is separate from this change. I'm simply documenting what currently works.

That said, I agree that it it would be nice to be able to specify arguments that include commas, and I think that your suggestion of simply specifying a list instead of a single string would be the most straight-forward way of doing it.

pmrowla:
Rebased this PR and created #4883 for future consideration regarding hydra

pmrowla:
Thanks for the PR @sjawhar!

efiop:
@sjawhar @pmrowlah Unfortunately there is no such package in conda, which means we can't use benedict as is. Is there an alternative lib that has conda package?

pmrowla:
Hydra does have a conda package (https://github.com/iterative/dvc/issues/4883). This can probably just be reverted until we get to that issue.

efiop:
@pmrowla Since hydra would change the syntax for this, I guess we better to do the switch before 2.0 if we indeed decide to go with it? Your call here. Could revert, or could submit benedict to conda-forge if we want (will be happy to help with sending one, we've sent a few already).

PR #4781, opened 2020-10-23
test_update_import: roll back skip

pared:

Related to #4418.


pared:
Well, that one is easy. Seems like the problem disappeared. 
I though some change has fixed the issue, but to my surprise, my own run with `skip` removed on top of original commit introducing github actions is also passing. It seems like something must have changed in a way how the runner (or one of the actions) is handling our project and now it passes without the original problem.

pmrowla (approved):

PR #4785, opened 2020-10-25
http: raise error on unsuccessful http statuses on `exists`

### LINKED ISSUE #4288: http: do not ignore HTTP statuses when working with HTTP(S) remotes
On master.

Exists check now simply says yes if request is successful. So If we get 403 Forbidden or 401 Auth Required these are ignored, which looks the same as absent files to dvc. The simplest fix would be using `res.raise_for_status()` in `.exists()` check unless we have success or 404.

damakuno:

This PR is to fix #4288 . It is still a work in progress as I haven't written any tests for the changes I have implemented, and I need some guidance on where/how to write tests for the project. Do I add to the unit test script in `dvc/tests/unit/remote/test_http.py`? 

I'm not sure if this requires documentation, but I'll create a PR for it if needed.

### DIFF
--- a/dvc/tree/http.py
+++ b/dvc/tree/http.py
@@ -141,7 +141,13 @@ class HTTPTree(BaseTree):  # pylint:disable=abstract-method
         return response
 
     def exists(self, path_info, use_dvcignore=True):
-        return bool(self._head(path_info.url))
+        res = self._head(path_info.url)
+        if res.status_code == 404:
+            return False
+        if bool(res):
+            return True
+        res.raise_for_status()
+        return False
 
     def get_file_hash(self, path_info):
         url = path_info.url
--- a/dvc/tree/http.py
+++ b/dvc/tree/http.py
@@ -146,8 +146,7 @@ class HTTPTree(BaseTree):  # pylint:disable=abstract-method
             return False
         if bool(res):
             return True
-        res.raise_for_status()
-        return False
+        raise HTTPError(res.status_code, res.reason)
 
     def get_file_hash(self, path_info):
         url = path_info.url
--- a/tests/unit/remote/test_http.py
+++ b/tests/unit/remote/test_http.py
@@ -127,23 +127,27 @@ def test_http_method(dvc):
     assert isinstance(tree._auth_method(), HTTPBasicAuth)
 
 
-def test_exists(dvc, http, mocker):
+def test_exists(mocker):
     import io
 
     import requests
 
+    from dvc.path_info import URLInfo
+
     res = requests.Response()
     res.raw = io.StringIO("foo")
-    tree = HTTPTree(dvc, http.config)
+    tree = HTTPTree(None, {})
 
     mocker.patch.object(tree, "request", return_value=res)
 
+    url = URLInfo("https://example.org/file.txt")
+
     res.status_code = 200
-    assert tree.exists(http / "file.txt") is True
+    assert tree.exists(url) is True
 
     res.status_code = 404
-    assert tree.exists(http / "file.txt") is False
+    assert tree.exists(url) is False
 
     res.status_code = 403
     with pytest.raises(HTTPError):
-        tree.exists(http / "file.txt")
+        tree.exists(url)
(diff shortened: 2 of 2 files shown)


damakuno:
I'm a bit stumped on the tests. I think I need to test 3 scenarios:

1. When the response returns 404
2. When the response returns 200 OK
3. Other types responses

I'm just not sure how to mock responses from the requests module?
I'm thinking about something like this, with comments on things I have yet to implement:
```python
def test_exists(dvc):
    from dvc.path_info import URLInfo
    config = {
        "url": "http://127.0.0.1",
        "path_info": "file.html"
    }
    path_info = URLInfo(config["url"])
    tree = HTTPTree(dvc, config)

    # mock a successful response that returns 200 OK
    assert tree.exists(path_info) is True
    # mock a response that returns 404
    assert tree.exists(path_info) is False
    # mock a HTTP Error response that is not 404 to 
    with pytest.raises(HTTPError):
        tree.exists(path_info)
```

skshetry:
@damakuno, sadly we don't have a good way to test _errors_ in http remote, as we are using this _hack_ for tests:
https://github.com/iterative/dvc/blob/59d5c73c5e5fd6e0c5bba7f6fc7ba5677380ed40/tests/utils/httpd.py#L11

The corresponding pytest http fixture is [here](https://github.com/iterative/dvc/blob/59d5c73c5e5fd6e0c5bba7f6fc7ba5677380ed40/tests/remotes/http.py).

The short-term solution can be mocking, but it's a code smell. But, it's the only one that I see for now: 
```
def test_exists(dvc, http, mocker):
    tree = HTTPTree(dvc, http.config)

    resp = requests.Response()
    resp.status_code = requests.codes.FORBIDDEN
    resp.raw = io.StringIO("boo!!!")

    mocker.patch.object(tree, "request", return_value=resp)
    with pytest.raises(HTTPError):
        tree.exists(http / "file.txt")
```

dvc/tree/http.py
```diff
@@ -141,7 +141,13 @@ def _head(self, url):
         return response
 
     def exists(self, path_info, use_dvcignore=True):
-        return bool(self._head(path_info.url))
+        res = self._head(path_info.url)
+        if res.status_code == 404:
+            return False
+        if bool(res):
+            return True
+        res.raise_for_status()
```
skshetry, line 149: It'd be better if we could wrap this with `HttpError` of dvc's itself.

damakuno: hmm perhaps by using a `try`...`catch` block to do that?

damakuno: or just raising dvc's `HTTPError` directly

damakuno:
How do I resolve the failing code coverage check? I'm not quite sure what's the issue

skshetry:
@damakuno, nothing to do with the coverage, as you have already added the tests.

tests/unit/remote/test_http.py
```diff
@@ -125,3 +125,25 @@ def test_http_method(dvc):
...
+
+
+def test_exists(dvc, http, mocker):
+    import io
+
+    import requests
+
+    res = requests.Response()
+    res.raw = io.StringIO("foo")
+    tree = HTTPTree(dvc, http.config)
```
skshetry, line 137: We just discussed this in the team, and they pointed out that we could get rid of `dvc` and `http` fixtures as we are already mocking the response.

@damakuno, let's remove `dvc` and `http` from above fixture and let's do the following:
```suggestion
    tree = HTTPTree(None, {})
```

tests/unit/remote/test_http.py
```diff
@@ -125,3 +125,25 @@ def test_http_method(dvc):
...
+
+    mocker.patch.object(tree, "request", return_value=res)
+
+    res.status_code = 200
+    assert tree.exists(http / "file.txt") is True
```
skshetry, line 142: Also, since we don't have `http` now, we need to create `url` that's a `URLInfo` (it comes from `dvc.path_info`).

```python
    url = URLInfo("https://example.org/file.txt")
```

And, now, pass those url to the `exists`:
```suggestion
    assert tree.exists(url) is True
```

tests/unit/remote/test_http.py
damakuno: I've added a commit with this change

tests/unit/remote/test_http.py
damakuno: I've added a commit with this change

skshetry (approved):
Thanks, @damakuno for contributing. The changes look good, and can be merged. I added a comment to clarify on the test. :tada:

PR #4603, opened 2020-09-23
params, plots, metrics: unify collection

pared:
* [x] ❗ I have followed the [Contributing to DVC](https://dvc.org/doc/user-guide/contributing/core) checklist.

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here.

Thank you for the contribution - we'll try to review it as soon as possible. 🙏
Related to #4446

PR #4789, opened 2020-10-26
metrics/plots: test metrics collection on no target on revision

pared:

Related to #4446, #4603 

In this change I introduce test checking behavior common both to `metrics` and `plots`. In order to prevent test duplication among `metrics` and `plots` I created `test_common` under `tests/func/metrics`. I also moved
`tests/func/metrics` -> `tests/func/metrics/metrics`
`tests/func/plots` -> `tests/func/metrics/plots`
So that we would have logic package structure, where common tests would be above particular subcommand tests. Not sure if naming of "main" package  is accurate. Considering that in our [docs](https://dvc.org/doc/command-reference/metrics#types-of-metrics) we name both `plots` and `metrics` as `metrics` I decided to go this way.


pmrowla (approved):

efiop (approved):

jorgeorpinel:
> Considering that in our docs we name both plots and metrics as metrics I decided to go this way.

Yeah it's tricky. I'm considering suggesting "simple metrics" vs. "plot metrics" too. Just a note 🙂

jorgeorpinel:
> p.s. so it could be metrics/simple and metrics/plots

PR #4801, opened 2020-10-29
add: warn on cache link errors instead of failing

### LINKED ISSUE #4797: dvc add deletes file when cache type fails on Windows
### Setup
* Windows client, where support for symlinks clearly didn't get applied. This has happened on all our Windows machines. It may have something to do with the way our corporate IT has set up Windows.
* Shared dvc cache on a samba mount. Samba server happens to be 3000 miles away from client.
* cache type symlink

### Situation
* Try to add a file via a line like `dvc add foo.dat`.
* File gets hashed, appears to get uploaded to cache, but then no .dvc file is created, the local copy of `foo.data` is _deleted_, and then dvc raises a "No cache types left to try" error when trying to symlink from the cache back to the local directory. 
* Basically, you're left with no local copy of the file, but also no md5 hash of the file to locate it in the cache!
* Change the local cache.type to `copy` and everything works fine.

I understand the symlink creation failing, but I don't understand why that error should result in the loss of data!

```
DVC version: 1.8.1 (exe)
 Platform: Python 3.7.9 on Windows-10-10.0.17763-SP0
 Supports: All remotes
 Cache types: symlink
 Cache directory: NTFS on Z:\
 Workspace directory: NTFS on D:\
 Repo: dvc, git
```

pmrowla:

Will fix #4797.
Docs PR https://github.com/iterative/dvc.org/pull/1892

* `output.commit()` no longer fails if file cannot be linked during `cache.save()` (as long as the file was still moved to cache successfully)
* `dvc add` will succesfully create .dvc file if the link step fails, and will warn user about the link failure and suggest re-running `dvc checkout`

Link failure during add now looks like:

```
$ dvc add foo
100% Add|██████████████████████████████████████████████████████████████████████████████████████████████████████████████|1/1 [00:00,  3.35file/s]
WARNING: Some targets could not be linked from cache to workspace.
See <https://dvc.org/doc/user-guide/troubleshooting#cache-types> for more information.
To re-link these targets, reconfigure cache types and run:

        dvc checkout foo.dvc

        git add foo.dvc

$ ls -l
total 4.0K
-rw-r--r-- 1 pmrowla staff 58 Oct 29 12:07 foo.dvc
```

Link failure during `dvc checkout` fails like before, but will include the paths which failed and the troubleshooting link.

```
$ dvc checkout foo.dvc
ERROR: No possible cache types for 'foo'. See <https://dvc.org/doc/user-guide/troubleshooting#cache-types> for more information.
```

### DIFF
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -343,7 +343,7 @@ class MergeError(DvcException):
 class CacheLinkError(DvcException):
     SUPPORT_LINK = "See {} for more information.".format(
         format_link(
-            "https://dvc.org/doc/user-guide/" "troubleshooting#cache-types"
+            "https://dvc.org/doc/user-guide/troubleshooting#cache-types"
         )
     )
 
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -338,3 +338,18 @@ class NoOutputOrStageError(DvcException):
 
 class MergeError(DvcException):
     pass
+
+
+class CacheLinkError(DvcException):
+    SUPPORT_LINK = "See {} for more information.".format(
+        format_link(
+            "https://dvc.org/doc/user-guide/" "troubleshooting#cache-types"
+        )
+    )
+
+    def __init__(self, path_infos):
+        msg = "No possible cache link types for '{}'. {}".format(
+            ", ".join([str(path) for path in path_infos]), self.SUPPORT_LINK,
+        )
+        super().__init__(msg)
+        self.path_infos = path_infos
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -8,6 +8,7 @@ from shortuuid import uuid
 
 import dvc.prompt as prompt
 from dvc.exceptions import (
+    CacheLinkError,
     CheckoutError,
     ConfirmRemoveError,
     DvcException,
@@ -192,7 +193,7 @@ class CloudCache:
                 )
                 del link_types[0]
 
-        raise DvcException("no possible cache types left to try out.")
+        raise CacheLinkError([to_info])
 
     def _do_link(self, from_info, to_info, link_method):
         if self.tree.exists(to_info):
--- a/dvc/repo/add.py
+++ b/dvc/repo/add.py
@@ -6,6 +6,7 @@ import colorama
 from dvc.dvcfile import Dvcfile, is_dvc_file
 
 from ..exceptions import (
+    CacheLinkError,
     OutputDuplicationError,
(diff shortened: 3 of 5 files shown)


dvc/exceptions.py
```diff
@@ -338,3 +338,18 @@ def __init__(self, target, file):
 
 class MergeError(DvcException):
     pass
+
+
+class CacheLinkError(DvcException):
+    SUPPORT_LINK = "See {} for more information.".format(
+        format_link(
+            "https://dvc.org/doc/user-guide/" "troubleshooting#cache-types"
```
pared, line 346:
```suggestion
            "https://dvc.org/doc/user-guide/troubleshooting#cache-types"
```

pared (approved):
Looks great!
One tiny `black` leftover.

efiop (approved):

dvc/exceptions.py
```diff
@@ -338,3 +338,18 @@ def __init__(self, target, file):
...
+
+class CacheLinkError(DvcException):
+    SUPPORT_LINK = "See {} for more information.".format(
+        format_link(
+            "https://dvc.org/doc/user-guide/troubleshooting#cache-types"
+        )
+    )
+
+    def __init__(self, path_infos):
+        msg = "No possible cache link types for '{}'. {}".format(
```
jorgeorpinel, line 351: Used to be "no possible cache types left to try" which is more of a phrase. Is that still what we mean by "No possible cache link types"? It looks like a typo in "not possible" but even then the phrase wouldn't make that much sense. Sorry, not sure what to suggest, as I don't understand the conditions that could lead to this error.

pmrowla: For `dvc checkout` it will occur in the same situations as the old "left to try" error. The "left to try" qualifier no longer seemed necessary to me given that we have the troubleshooting page w/a more thorough explanation now

jorgeorpinel: I see. What were those situations again? 🙂

jorgeorpinel: Oh I see it's explained in https://github.com/iterative/dvc.org/pull/1892#pullrequestreview-521909163:

> if your configured cache types are unsupported for your filesystem... or where a platform normally supports your configured link type(s) but the actual link operation itself fails due to some (non-DVC related) problem

jorgeorpinel: So how about something like

> Unable to create any of the configured cache link types: '{cache.type list}'.
> Affected files: '{list of outputs}'.
> See {link} for more details.

PR #4802, opened 2020-10-29
[#4136] fixed recalculating relative paths in config

### LINKED ISSUE #4136: conf: credentialpath is not recalculated when it's a relpath
This one supposed to be relative to config file and converted to absolute and back on config load/save. This is done for `cache.dir`, local urls and `gdrive_user_credentials_file`, but not for `credentialpath`, which is used in s3 and gs.

Example: `cache.dir` - https://dvc.org/doc/command-reference/cache/dir#description.

godfryd:
Other paths beside gdrive_user_credentials_file and cache.dir should be recalculated.
Added the following ones:
- gdrive_service_account_p12_file_path
- credentialpath
- keyfile
- cert_path
- key_path

Will fix #4136.


efiop (changes requested):
Thanks for the PR! Looks great!

Could we slim down the tests to just test that the paths are adjusted properly when loading a config with these fields? Current tests do the job, but they are quite heavy for such a simple function. Also, I see a lot of code duplication in the tests, would it be possible to reasonably us e pytest parametrize helper?

tests/func/remote/test_setting_paths.py
```diff
@@ -0,0 +1,37 @@
...
+        ("keyfile", "ssh://user@example.com:1234/path/to/dir"),
+        ("cert_path", "webdavs://example.com/files/USERNAME/"),
+        ("key_path", "webdavs://example.com/files/USERNAME/"),
+    ],
+)
+def test_relative_config_setting(tmp_dir, dvc, field, remote_url):
+    a_file = os.path.join("secrets", "file.txt")
+
+    remote_name = "some_name"
+    assert main(["remote", "add", "-d", remote_name, remote_url,]) == 0
```
efiop, line 25: Linter CI job is complaining about this one
```suggestion
    assert main(["remote", "add", "-d", remote_name, remote_url]) == 0
```

efiop (commented):
Looks much better now! Thank you! :pray: 

What do you think about creating instead a unit test(maybe -ish) for the `Config` class, to check that when it contains all of these fields - they are are all getting resolved properly when loading?

tests/func/remote/test_setting_paths.py
godfryd: fixed

godfryd:
I'm not sure. Current tests involve setting paths and loading them in a config. Proposed test would do only the second part.

tests/func/remote/test_setting_paths.py
```diff
@@ -0,0 +1,37 @@
...
+    # We need to load repo again to test updates to the config
+    str_path = os.fspath(tmp_dir)
+    repo = Repo(str_path)
+
+    # Check that in config we got the path relative to the config file itself
+    # Also, we store posix path even on Windows
+    config = configobj.ConfigObj(repo.config.files["repo"])
+    assert config[f'remote "{remote_name}"'][field] == posixpath.join(
+        "..", "secrets", "file.txt"
+    )
```
efiop, line 37: Seems like this is testing how it is written to the config file, but not really how it is loaded. Or am I missing something?

efiop:
@godfryd The setting part is pretty safe. You are currently using CLI and realoding the whole repo instance in order to test it, but you could really just do something like:
```
with config.edit() as conf:
    # set the options you want to test

# check if written paths are correct

# load config and check that it contains what we expect (relative paths are evaluated correctly)
```
which should be easily parametrizable (or could even shove everything into one test without parametrizations), simple and lightweight. What do you think?

efiop (changes requested):
Looks great! Please take a look at the windows tests, looks like there is a bug
```
2020-11-07T09:49:32.2041318Z     def test_load_relative_paths(dvc, field, remote_url):
2020-11-07T09:49:32.2041690Z         # set field to test
2020-11-07T09:49:32.2042032Z         with dvc.config.edit() as conf:
2020-11-07T09:49:32.2042463Z             conf["remote"]["test"] = {"url": remote_url, field: "file.txt"}
2020-11-07T09:49:32.2042793Z     
2020-11-07T09:49:32.2043269Z         # check if written paths are correct
2020-11-07T09:49:32.2043710Z         dvc_dir = dvc.config.dvc_dir
2020-11-07T09:49:32.2044307Z >       assert dvc.config["remote"]["test"][field] == os.path.join(
2020-11-07T09:49:32.2044735Z             dvc_dir, "..", "file.txt"
2020-11-07T09:49:32.2045003Z         )
2020-11-07T09:49:32.2045425Z E       AssertionError: assert 'C:\\Users\\r...\\../file.txt' == 'C:\\Users\\r...\..\\file.txt'
2020-11-07T09:49:32.2046433Z E         - C:\Users\runneradmin\AppData\Local\Temp\pytest-of-runneradmin\pytest-0\popen-gw3\test_load_relative_paths_cert_0\.dvc\..\file.txt
2020-11-07T09:49:32.2047086Z E         ?                                                                                                                         ^
2020-11-07T09:49:32.2047786Z E         + C:\Users\runneradmin\AppData\Local\Temp\pytest-of-runneradmin\pytest-0\popen-gw3\test_load_relative_paths_cert_0\.dvc\../file.txt
2020-11-07T09:49:32.2048624Z E         ?     
```

dvc/config.py
```diff
@@ -377,7 +377,14 @@ def _load_paths(conf, filename):
...
                 return path
-            return RelPath(os.path.join(abs_conf_dir, path))
+
+            # on windows convert slashes to backslashes
+            # to have path compatible with abs_conf_dir
+            if os.path.sep == "\\" and "/" in path:
+                path = path.replace("/", "\\")
+
+            path = RelPath(os.path.join(abs_conf_dir, path))
+            return path
```
efiop, line 387:
```suggestion
            return RelPath(os.path.join(abs_conf_dir, path))
```

dvc/config.py
```diff
@@ -389,7 +395,8 @@ def _to_relpath(conf_dir, path):
         if isinstance(path, RelPath) or not os.path.isabs(path):
             path = relpath(path, conf_dir)
 
-        return PathInfo(path).as_posix()
+        path = PathInfo(path).as_posix()
+        return path
```
efiop, line 399:
```suggestion
        return PathInfo(path).as_posix()
```

efiop (approved):
Looks great! Thank you! :pray:

Issue #4440, opened 2020-08-21
experiments: need commands for managing staged/stashed exp list


pmrowla:
Currently the experiments stash list can only be managed manually by using `git stash ...` in the experiments workspace. Ideally we should be able to at least clear the full list and delete specific experiments via `dvc exp` commands.

Also, since we preserve stashed experiments in the event `repro --exp` fails for that experiment, it may also be useful to be able to retrieve information about the failure, and then edit/rerun/delete the stashed experiment as needed.

pmrowla:
This should get resolved by something similar to `dvc exp gc ...` which will garbage collect un-needed experiments and their related cache artifacts. This needs to be implemented before release, but is not a priority issue yet

pmrowla:
Used object collection for experiments will depend on the exp sharing feature, since collection will require walking the custom exp refs rather than `.dvc/experiments` branches. So the remaining `gc` task is on hold until exp sharing is done.

PR #4803, opened 2020-10-29
experiments: add `dvc exp gc`

pmrowla:

Related to #4440.

* Adds `dvc exp gc` which can be used to remove experiments which are no longer needed
* `--workspace`/`--all-branches`/`--all-tags`/`--all-commits` are equivalent to the regular `dvc gc` flags - experiments derived from the specified set of commits will be preserved and all other experiments will be removed
* This command will not remove any objects from DVC cache, `dvc gc -e/--exp` option to preserve used-cache for experiments will be added in a follow up PR

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -8,6 +8,7 @@ from datetime import date, datetime
 from itertools import groupby
 from typing import Iterable, Optional
 
+import dvc.prompt as prompt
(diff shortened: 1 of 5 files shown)


pared (approved):
Looks great! 
I wonder whether all the tests couldn't be packed into one, but I guess reading such tests would be much harder than it is now.

PR #4811, opened 2020-10-30
Feature/add cache status command

### LINKED ISSUE #4020: introduce cache check/repair command(s)
Not 100% sure about the naming, but the idea is for dvc to make sure that it has all the permissions in-check for the cache that it will use. E.g. https://github.com/iterative/dvc/issues/3964 , where we had an issue with shared cache.

`dvc cache check` could walk around the cache, making sure it is able to read and write to it (though read-only is also fine for some scenarios, especially since we no longer have the unpacked dir) and `dvc cache repair` would possibly ask for sudo and chmod everything as it sees fit.

lsc20051426:
Fixes #4020

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here.  https://github.com/iterative/dvc.org/issues/1895


dvc/command/cache.py
```diff
@@ -18,6 +23,86 @@ def run(self):
...
 
+class CmdCacheStatus(CmdBase):
+    def run(self):
+        self.permission_check()
+        self.show_dvc_files_status()
+        return 0
+
+    def permission_check(self):
+        cache_dir = self.config["cache"]["dir"]
+        logger.info("Step 1: Permission Check on: {}".format(cache_dir))
```
jorgeorpinel, line 34:
```suggestion
        logger.info("Step 1: Permissions check on: '{}'".format(cache_dir))
```

dvc/command/cache.py
```diff
@@ -18,6 +23,86 @@ def run(self):
...
+            )
+        )
+        logger.info(
+            "\tExist: {}".format(
+                "OK" if os.access(cache_dir, os.F_OK) else "Not OK"
+            )
+        )
+
+    def show_dvc_files_status(self):
+        logger.info("Step 2: DVC Files and Cache status.")
```
jorgeorpinel, line 52: What does "DVC Files" refer to specifically? Maybe
```suggestion
        logger.info("Step 2: .dvc/ internal files and cache status")
```

jorgeorpinel: Just making it clear it doesn't refer to .dvc files, or dvc.yaml files, or does it?

tests/unit/command/test_cache.py
```diff
@@ -0,0 +1,15 @@
...
+    cmd = args.func(args)
+
+    mocker.patch("os.access", return_value=True)
+
+    assert 0 == cmd.run()
+    assert "Step 1: Permission Check on:" in caplog.text
+    assert "Read: OK" in caplog.text
+    assert "Write: OK" in caplog.text
+    assert "Exist: OK" in caplog.text
+    assert "Step 2: DVC Files and Cache status." in caplog.text
```
jorgeorpinel, line 15: (Would need corresponding text changes too)

dvc/command/cache.py
lsc20051426: @jorgeorpinel how about `versioned files and cache status`?

jorgeorpinel: OK seeing the table headers I think something like this would be accurate:
```suggestion
        logger.info("Step 2: Cache details of all DVC-tracked files:")
```

jorgeorpinel: Or maybe even skip this INFO message? It's what th user is expecting when they run `dvc cache status`

@lsc20051426 could you maybe paste here an example output of this new command would look like? To make sure we're on the same page. Like this

```console
$ dvc cache status
Step 1: ...
...
Step 2: ...
(Table)
```

Thanks

lsc20051426: @jorgeorpinel 
Currently, the output sample:
```
dvc cache status
Step 1: Permission Check on: xxx/.dvc/cache
	Read: OK
	Write: OK
	Exist: OK
Step 2: DVC Files and Cache status.
File/Directory    Name                   Size    CachePath                                         CacheSize
File              data/images/cat1.jpg   31926   .dvc/cache/c6/b4d357699d641823c537514996a44f      31926
File              data/images/cat2.jpeg  31291   .dvc/cache/27/e39df5f434f513fafda870624e063d      31291
File              samples/output1.txt    10      .dvc/cache/3f/a9c057aa667dd5869a249b13574057      10
Dir               data/images            -       .dvc/cache/0f/56c851f043425b39afb8742a2849c3.dir  -
File              data/model.v1.txt      4       .dvc/cache/48/d6d64719872cef5d4d3541ffc4a4a0      4
```
I think `Cache details of all DVC-tracked files:` is a good INFO message.

efiop (changes requested):
Hi @lsc20051426 !

Thank you for the PR!

Unfortunately, this PR doesn't really implement what we were intending in the original issue. For example:

1) cache permissions mentioned in the ticket are not about whether or not we can read/write the cache dir itself(the meaning of W_OK/R_OK/F_OK is a bit different for the dirs, btw), but rather that we are able to access it its subdirs and cache files and their permissions in general match our expectations.

2) The way you walk through the cache files and dvc files works for some cases, but it should really account for our dvc.yaml and dvc.lock pipeline files and go from the DAG to the cache. But that functionality in `dvc cache status` is also questionable, as we don't really care about our DAG and it's outputs in this command.

I'm really sorry, but I would suggest just closing this PR, since this needs to be reconsidered and redone from scratch :slightly_frowning_face:

lsc20051426:
> Hi @lsc20051426 !
> 
> Thank you for the PR!
> 
> Unfortunately, this PR doesn't really implement what we were intending in the original issue. For example:
> 
> 1. cache permissions mentioned in the ticket are not about whether or not we can read/write the cache dir itself(the meaning of W_OK/R_OK/F_OK is a bit different for the dirs, btw), but rather that we are able to access it its subdirs and cache files and their permissions in general match our expectations.
> 2. The way you walk through the cache files and dvc files works for some cases, but it should really account for our dvc.yaml and dvc.lock pipeline files and go from the DAG to the cache. But that functionality in `dvc cache status` is also questionable, as we don't really care about our DAG and it's outputs in this command.
> 
> I'm really sorry, but I would suggest just closing this PR, since this needs to be reconsidered and redone from scratch 🙁

Hi @efiop no worries. I will just close the PR first and pick up some other issues with label `good first issue`. thanks for your information.

dvc/command/cache.py
jorgeorpinel: Thanks for the update!

Issue #2992, opened 2019-12-21
better message when trying to link from files that belong to another user


qiuwei:
**Please provide information about your setup**
DVC version(i.e. `dvc --version`), Platform and method of installation (pip, homebrew, pkg Mac, exe (Windows), DEB(Linux), RPM(Linux))
DVC version: 0.77.3, Platform linux, method of installation: anaconda

Problem:
Our team has a shared data directory, all of the files and directories belong to the same group.
The members of the group have full read and write access.

I create dvc repository with `--no-scm` option successfully. However when I add any file whose owner is not the current user, dvc fails very badly.
It panics with the following error,
```
ERROR: unexpected error - [Errno 1] Operation not permitted: '/gendata/.dvc/cache/80/b1aeab28f6ddf9ba779b2681923442.ZKMhApEkd5VrDDarerdKmU'
```
 However, the file which I was adding is removed from the directory.

Maybe at the moment dvc doesn't support shared repository? 
However, I consider the error message is not clear enough, and it should never remove user's files when it fails.

shcheklein:
@qiuwei that's sounds really bad, but it's not exactly clear what's happening here - could you please give us a little bit more details?

 > Our team has a shared data directory

is it `/gendata`? do you use the same dir for the DVC cache? is DVC cache set to be shared as well?

> DVC doesn't support shared repository?

where is your repository located? the same directory as data?

chloe-wang:
Yes, it's /gendata.

I could provide a minimal working exmaple to reproduce.

step 1: create `proj1` directory
step 2: inside the `proj1` directory, initialize a dvc directory without scm.
```
dvc init --no-scm
```
step 3: create a file `test.txt`, change the owner of the file to someone else other than the current user, and change the permission bit to 664.  Note that the two users should be in the same group, so that the current user still has access to the `test.txt`.

step 4: add `test.txt` to dvc
```
dvc add test.txt
```
then you would be able to see the error mesage like:
```
ERROR: unexpected error - [Errno 1] Operation not permitted: '/gendata/.dvc/cache/80/b1aeab28f6ddf9ba779b2681923442.ZKMhApEkd5VrDDarerdKmU'

```

efiop:
Hi @qiuwei ! Could you please add `-v` to the command that is failing and show us full log?

efiop:
@qiuwei Also, take a look at `dvc config cache.shared group` config option, it allows you to use your cache directory that is shared between users.

[...]

PR #4813, opened 2020-10-30
utils/fs: exception handling in `move`

### LINKED ISSUE #4348: Parquet file owned by other user gets deleted upon dvc add
## Bug Report
 
On a Linux machine, in a folder `foo/` shared by various users, but not owned by me, I did:
```
git init .
dvc init .
git add .dvc
git commit -m "first commit"
```
The folder already had a file `bar.parquet` sitting in it, not owned by me either.
I did:
```
dvc add bar.parquet
```
And got the error
```
Adding...
ERROR: unexpected error - [Errno 1] Operation not permitted: '/baz/.dvc-cache/d0/3ea995971fc314c3087abbb3106e92.L6rCVNZWCFrzQRk53wdSCS'
```
 
**Most importantly**, the `foo/bar.parquet` file got deleted by the `dvc add` command above.
 
Note that I am using a shared `dvc` cache located at `/baz/.dvc-cache/`.
 
**Output of `dvc version`:**
 
```console
$ dvc version
1.3.1
```
 
I could not reproduce the bug. After recreating `bar.parquet` manually, with me being the owner, I could run `dvc add bar.parquet` successfully.

danielfleischer:
When file is not owned by us, chmoding will naturally fail;
now it would throw an exception and return the file back from `tmp` to the original `src`,
letting the user decide what to do. Should fix #4348 and #2992.


dvc/utils/fs.py
```diff
@@ -101,8 +101,13 @@ def move(src, dst, mode=None):
     else:
         shutil.move(src, tmp)
 
-    if mode is not None:
-        os.chmod(tmp, mode)
+    try:
```
skshetry, line 104: Can we check the permissions before we move it?

danielfleischer: Thanks for the feedback, it is better to first do the check.

Issue #829, opened 2018-06-30
Split data into blocks 


efiop:
As @shcheklein suggested, we should consider splitting data into small blocks to track data changes more efficiently. Example: giant file that has one line appended to it.

sotte:
For a private data sharing/deduplication tool I've used a buzhash/rolling hash for chunking big files into small chunks pretty successfully. The [borg backup tool](https://github.com/borgbackup/borg) has a nice implementation of a [buzhash chunker](https://github.com/borgbackup/borg/blob/master/src/borg/chunker.pyx). Maybe it's worth checking out.

efiop:
@Witiko noted that there is ficlonerange that we could use to create file by reflinking smaller files into it. https://github.com/iterative/dvc/issues/2073#issuecomment-497864614

salotz:
This is a pretty good solution to the chunking deduplication problem. I've used it with success for backing things up: https://duplicacy.com/.
If anything you could use the algorithm.

jorgeorpinel:
the way Git handles this is

> looks for files that are named and sized similarly, and stores just the deltas from one version of the file to the next
> See https://git-scm.com/book/en/v2/Git-Internals-Packfiles

So that would be another possible approach — equivalent to chunking down to the byte. Mentioned in #1487 BTW:

> 4. Diff's for dataset versions (see 2.1.). Which files were added\deleted\modified.

Either approach complicates the file linking from cache to workspace anyway.

karajan1001:
> As @shcheklein suggested, we should consider splitting data into small blocks to track data changes more efficiently. Example: giant file that has one line appended to it.

@efiop There is a problem with it that if the new line was inserted on top of the file, or in the middle.

sotte:
> > As @shcheklein suggested, we should consider splitting data into small blocks to track data changes more efficiently. Example: giant file that has one line appended to it.
> 
> @efiop There is a problem with it that if the new line was inserted on top of the file, or in the middle.

There are methods that can deal with inserts at various positions. I mentioned this earlier:
https://github.com/iterative/dvc/issues/829#issuecomment-438705585

Most of the chunks/blocks of a file stay the same, only chunks/blocks that change are actually added. Many backup tools use similar methods.

efiop:
For the record: another thing to consider is to revisit `.dir` format to be able to split it into smaller files. E.g.
`123456.dir` could look like (it might still be json or not)
```
{
   "dir": "654321.dir"
   "foo": "...",
[...]

PR #4820, opened 2020-11-01
[#4819] fixed failing to push some files to http remote

### LINKED ISSUE #4819: Pushing multiple files to http remote fails
## Bug Report
When trying to push over 100 files via HTTP/s, with a 5 Mbp/s upload connection, multiple files fail to upload because of a `socket timeout` exception. 

**Output of `dvc version`:**

```console
DVC version: 1.9.1 (pip)
Platform: Python 3.7.7 on Darwin-19.6.0-x86_64-i386-64bit
Supports: All remotes
```

Changing the `REQUEST_TIMEOUT` to 60 seconds (same as GS timeout) in debug mode solves the issue.

simonlousky:

* Increased the `HTTPTree.REQUEST_TIMEOUT` from 10 to 60 seconds, as it was not changed after implementing upload for http.
 
Will fix #4819

### DIFF
--- a/dvc/tree/http.py
+++ b/dvc/tree/http.py
@@ -33,7 +33,7 @@ class HTTPTree(BaseTree):  # pylint:disable=abstract-method
 
     SESSION_RETRIES = 5
     SESSION_BACKOFF_FACTOR = 0.1
-    REQUEST_TIMEOUT = 10
+    REQUEST_TIMEOUT = 60
     CHUNK_SIZE = 2 ** 16
 
     def __init__(self, repo, config):


dvc/tree/http.py
```diff
@@ -33,7 +33,7 @@ class HTTPTree(BaseTree):  # pylint:disable=abstract-method
 
     SESSION_RETRIES = 5
     SESSION_BACKOFF_FACTOR = 0.1
-    REQUEST_TIMEOUT = 10
+    REQUEST_TIMEOUT = 60
```
efiop, line 36: This might also be not enough if the file is big enough and/or the connection is slow enough. What is the biggest file size that you are uploading? We are having a similar discussion about webdav where apparently chunked upload is not always possible and so that makes https://github.com/iterative/dvc/issues/829 more relevant. Maybe it does here too.

efiop (approved):
Thank you for the PR! Looks good! Please see one comment above.

For the record: windows failures are unrelated.

dvc/tree/http.py
simonlousky: The chunk size has never caused a problem for a single large file in my case. Only having too many concurrent files uploading would make some of them time out. 
In general, with a slow enough connection, or too many concurrent jobs, some file will eventually fail to upload. 
This relates to other issues discussing the implementation of retries.

PR #4828, opened 2020-11-02
remote: avoid chunking on webdav. Fixes #4796

### LINKED ISSUE #4796: Pushing artifacts via WebDAV results in a 411 Length Required response
## Bug Report

I am trying to connect to a remote via WebDAV. I can correctly setup user and password along with the url, but when I try to push the artifacts I get a ```411 Length Required``` response. How can I solve the missing header problem?

DVC version: 1.9.0 (brew)
Platform: Python 3.9.0 on macOS-10.15.7-x86_64-i386-64bit
Supports: azure, gdrive, gs, http, https, s3, ssh, oss, webdav, webdavs
Repo: dvc, git

LucaButera:

### DIFF
--- a/dvc/tree/webdav.py
+++ b/dvc/tree/webdav.py
@@ -216,31 +216,44 @@ class WebDAVTree(BaseTree):  # pylint:disable=abstract-method
                 )
 
     # Uploads file to remote
-    def _upload(self, from_file, to_info, name=None, no_progress_bar=False):
+    def _upload(
+        self,
+        from_file,
+        to_info,
+        name=None,
+        no_progress_bar=False,
+        chunked=False,
+    ):
         # First try to create parent directories
         self.makedirs(to_info.parent)
 
-        # Progress from HTTPTree
-        def chunks():
-            with open(from_file, "rb") as fd:
-                with Tqdm.wrapattr(
-                    fd,
-                    "read",
-                    total=None
-                    if no_progress_bar
-                    else os.path.getsize(from_file),
-                    leave=False,
-                    desc=to_info.url if name is None else name,
-                    disable=no_progress_bar,
-                ) as fd_wrapped:
-                    while True:
-                        chunk = fd_wrapped.read(self.CHUNK_SIZE)
-                        if not chunk:
-                            break
-                        yield chunk
+        if chunked:
+            # Progress from HTTPTree
+            def chunks():
+                with open(from_file, "rb") as fd:
+                    with Tqdm.wrapattr(
+                        fd,
+                        "read",
+                        total=None
+                        if no_progress_bar
+                        else os.path.getsize(from_file),
+                        leave=False,
+                        desc=to_info.url if name is None else name,
+                        disable=no_progress_bar,
+                    ) as fd_wrapped:
+                        while True:
+                            chunk = fd_wrapped.read(self.CHUNK_SIZE)
+                            if not chunk:
+                                break
+                            yield chunk
 
-        # Upload to WebDAV via buffer
-        self._client.upload_to(buff=chunks(), remote_path=to_info.path)
+            # Upload to WebDAV via buffer
+            self._client.upload_to(buff=chunks(), remote_path=to_info.path)
+        else:
+            # Upload to WebDAV without chunking
+            self._client.upload_file(
+                local_path=from_file, remote_path=to_info.path
+            )
 
     # Queries size of file at remote
     def _file_size(self, path_info):
--- a/dvc/tree/webdav.py
+++ b/dvc/tree/webdav.py
@@ -251,9 +251,23 @@ class WebDAVTree(BaseTree):  # pylint:disable=abstract-method
             self._client.upload_to(buff=chunks(), remote_path=to_info.path)
         else:
             # Upload to WebDAV without chunking
-            self._client.upload_file(
-                local_path=from_file, remote_path=to_info.path
-            )
+            # self._client.upload_file(
+            #    local_path=from_file, remote_path=to_info.path
+            # )
+            with open(from_file, "rb") as fd:
+                with Tqdm.wrapattr(
+                    fd,
+                    "read",
+                    total=None
+                    if no_progress_bar
+                    else os.path.getsize(from_file),
+                    leave=False,
+                    desc=to_info.url if name is None else name,
+                    disable=no_progress_bar,
+                ) as fd_wrapped:
+                    self._client.upload_to(
+                        buff=fd_wrapped, remote_path=to_info.path
+                    )
 
     # Queries size of file at remote
     def _file_size(self, path_info):
--- a/dvc/tree/webdav.py
+++ b/dvc/tree/webdav.py
@@ -3,7 +3,7 @@ import os
 import threading
 from collections import deque
 
-from funcy import cached_property, wrap_prop
+from funcy import cached_property, nullcontext, wrap_prop
 
 from dvc.config import ConfigError
 from dvc.exceptions import DvcException
@@ -222,15 +222,21 @@ class WebDAVTree(BaseTree):  # pylint:disable=abstract-method
         # First try to create parent directories
         self.makedirs(to_info.parent)
 
+        file_size = os.path.getsize(from_file)
         with open(from_file, "rb") as fd:
-            with Tqdm.wrapattr(
-                fd,
-                "read",
-                total=None if no_progress_bar else os.path.getsize(from_file),
-                leave=False,
-                desc=to_info.url if name is None else name,
-                disable=no_progress_bar,
-            ) as fd_wrapped:
+            progress_context = (
+                nullcontext(fd)
+                if file_size == 0
+                else Tqdm.wrapattr(
+                    fd,
+                    "read",
+                    total=None if no_progress_bar else file_size,
+                    leave=False,
+                    desc=to_info.url if name is None else name,
+                    disable=no_progress_bar,
+                )
+            )
+            with progress_context as fd_wrapped:
                 self._client.upload_to(
                     buff=fd_wrapped, remote_path=to_info.path
                 )
--- a/dvc/tree/webdav.py
+++ b/dvc/tree/webdav.py
@@ -217,57 +217,25 @@ class WebDAVTree(BaseTree):  # pylint:disable=abstract-method
 
     # Uploads file to remote
     def _upload(
-        self,
-        from_file,
-        to_info,
-        name=None,
-        no_progress_bar=False,
-        chunked=False,
+        self, from_file, to_info, name=None, no_progress_bar=False,
     ):
         # First try to create parent directories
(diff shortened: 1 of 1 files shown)


LucaButera:
@skshetry I have tested the "not-chunked" with webdav both in pushing and pulling files. 
I report that:
1) Large files are correctly loaded up to ~7GB, bigger than that returns an error on unexpected file size. I assume that the timeout (ori similar) occurs and the server only gets part of the file.
2) Pushing normally sized files works even for an high number of them (around a thousand) concurrently. Anyway some of them return a Code 403 or 423 (a small number though). Trying to push agains results in the missing files being correctly uploaded.
3) Pulling works correctly.

At this points I just miss how to implement the selection between chunked an not chunked.

dvc/tree/webdav.py
```diff
@@ -216,31 +216,44 @@ def _download(self, from_info, to_file, name=None, no_progress_bar=False):
...
+                        disable=no_progress_bar,
+                    ) as fd_wrapped:
+                        while True:
+                            chunk = fd_wrapped.read(self.CHUNK_SIZE)
+                            if not chunk:
+                                break
+                            yield chunk
+
+            # Upload to WebDAV via buffer
+            self._client.upload_to(buff=chunks(), remote_path=to_info.path)
```
skshetry, line 251: This is a dead code, right?

dvc/tree/webdav.py
```diff
@@ -216,31 +216,44 @@ def _download(self, from_info, to_file, name=None, no_progress_bar=False):
...
+        else:
+            # Upload to WebDAV without chunking
+            self._client.upload_file(
```
skshetry, line 254: @LucaButera, this won't show up the progress bar. Can you check the snippets on the issue with progress bar? Please see if works.

skshetry (commented):
Hi, @LucaButera, thanks for the contribution. :tada:  I have a few suggestions above.

dvc/tree/webdav.py
LucaButera: I am not completely sure about that. I may be wrong but from my understanding chunking could work on webdav, or am I missing something? If only "non-chunked" has to be supported then you're right, it is dead code.

dvc/tree/webdav.py
LucaButera: From my testing the progress bar does work, but instead of having a global progress bar with "sub-bars" for each file you only get a single bar which updates for each file sent (more or less). This is perfectly fine for uploads of folders containing many reasonably sized files. It may actually become a problem when you upload one (or more) huge files. In such cases you have no information about the progress.

skshetry: Can you check with the snippet here that I mentioned: https://github.com/iterative/dvc/issues/4796#issuecomment-719181936?

LucaButera: Here's what I found out:
The snippet works flawlessly. It gives per file progress bar which updates correctly even in the case of a single, really big, file.

A few things that may or may not be related that I wish to mention:
* When querying the remote cache the progress bar starts from ```0/512``` even though the files are more than that. In fact, after it reaches 512, it starts going up like ```513/?``` and so on.
* After pushing a 7GB file, cache querying became extremely slow. I am not sure if it is just because the cache itself was bigger. Previously my remote contained 2.9GB divided in ~900 files and it seemed insanely fast in proportion.
* During push some files generate an error 403 or 423. Again this disappears the second time you try to push the files that gave the error.

Let me know what you think about this and how to proceed, thank you!

dvc/tree/webdav.py
```diff
@@ -216,12 +216,44 @@ def _download(self, from_info, to_file, name=None, no_progress_bar=False):
...
+        name=None,
+        no_progress_bar=False,
+        chunked=False,
+    ):
         # First try to create parent directories
         self.makedirs(to_info.parent)
 
-        # Progress from HTTPTree
-        def chunks():
+        if chunked:
```
skshetry, line 230: Let's remove the `if chunked` block, as we have discussed in #4796.

dvc/tree/webdav.py
LucaButera: @skshetry Did you have time to read the issues I mentioned? Do you think they are just server related problems or could there be some issues with dvc?

skshetry: @LucaButera, I am not sure of 1. 

For the 2, can you post the log? This patch should not affect that. Maybe, you were uploading the folders/files too-many times during the tests, that the local cache was getting invalidated. We try to guess the number of files to reduce no. of requests, so, that's my guess. Maybe, same thing is affecting 1.

For 3, it might happen with Owncloud/NextCloud if you are deleting/adding files too quickly, re in your test. So, it's not a DVC's issue. If that happens, retry again in a while.

Anyway, it should not be related to this patch. Please create a issue if that happens later.

dvc/tree/webdav.py
```diff
@@ -216,31 +216,26 @@ def _download(self, from_info, to_file, name=None, no_progress_bar=False):
...
-                    while True:
-                        chunk = fd_wrapped.read(self.CHUNK_SIZE)
-                        if not chunk:
-                            break
-                        yield chunk
-
-        # Upload to WebDAV via buffer
-        self._client.upload_to(buff=chunks(), remote_path=to_info.path)
+        # Upload to WebDAV without chunking
+        # Using non chunked progress
```
skshetry, line 226: This is unnecessary, as seen from the code below. Maybe add the link to the issue if you want to, otherwise we could skip it.

dvc/tree/webdav.py
```diff
@@ -216,31 +216,26 @@ def _download(self, from_info, to_file, name=None, no_progress_bar=False):
...
+        with open(from_file, "rb") as fd:
+            with Tqdm.wrapattr(
+                fd,
+                "read",
+                total=None if no_progress_bar else os.path.getsize(from_file),
+                leave=False,
+                desc=to_info.url if name is None else name,
+                disable=no_progress_bar,
+            ) as fd_wrapped:
+                self._client.upload_to(
```
skshetry, line 236: @LucaButera, can you test with an empty file? If that works well, it should be good to merge.

LucaButera: I think we have a problem with this. I get this error:

```
  File "/Users/lucabutera/dvc/dvc/cache/local.py", line 32, in wrapper
    func(from_info, to_info, *args, **kwargs)
  File "/Users/lucabutera/dvc/dvc/tree/base.py", line 356, in upload
    self._upload(  # noqa, pylint: disable=no-member
  File "/Users/lucabutera/dvc/dvc/tree/webdav.py", line 234, in _upload
    self._client.upload_to(
  File "/Users/lucabutera/dvc/.env/lib/python3.8/site-packages/webdav3/client.py", line 66, in _wrapper
    res = fn(self, *args, **kw)
  File "/Users/lucabutera/dvc/.env/lib/python3.8/site-packages/webdav3/client.py", line 438, in upload_to
    self.execute_request(action='upload', path=urn.quote(), data=buff)
  File "/Users/lucabutera/dvc/.env/lib/python3.8/site-packages/webdav3/client.py", line 208, in execute_request
    response = self.session.request(
  File "/Users/lucabutera/dvc/.env/lib/python3.8/site-packages/requests/sessions.py", line 530, in request
    resp = self.send(prep, **send_kwargs)
  File "/Users/lucabutera/dvc/.env/lib/python3.8/site-packages/requests/sessions.py", line 643, in send
    r = adapter.send(request, **kwargs)
  File "/Users/lucabutera/dvc/.env/lib/python3.8/site-packages/requests/adapters.py", line 469, in send
    for i in request.body:
TypeError: 'CallbackIOWrapper' object is not iterable
```

Which I think is related to the file being empty.
I created the file with ```touch filename.txt```

skshetry: @LucaButera, you might be able to solve this with passing `disable=True` when the `size` is `0` (see the `total` kwarg is getting passed the total size).
If that does not work, we can use a conditional context manager [`nullcontext`](https://funcy.readthedocs.io/en/stable/flow.html?highlight=nullcontext#nullcontext) in the case of `0` size, otherwise use `Tqdm.wrappattr`.

LucaButera: @skshetry 
So, I tried to set ```disable=no_progress_bar or os.path.getsize(from_file) == 0``` to no success. It yields the same error as before. 

Using ```nullcontext```, instead, solves the issue. I implemented it in this fashion:
```
with open(from_file, "rb") as fd:
    progress_context = nullcontext(fd) if file_size == 0 else Tqdm.wrapattr(fd, "read", ...)
    with progress_context as fd_wrapped:
        self._client.upload_to(buff=fd_wrapped, remote_path=to_info.path)
```
but then another error occurs:
```
  File "/Users/lucabutera/dvc/dvc/cache/local.py", line 32, in wrapper
    func(from_info, to_info, *args, **kwargs)
  File "/Users/lucabutera/dvc/dvc/tree/base.py", line 356, in upload
    self._upload(  # noqa, pylint: disable=no-member
  File "/Users/lucabutera/dvc/dvc/tree/webdav.py", line 236, in _upload
    self._client.upload_to(
  File "/Users/lucabutera/dvc/.env/lib/python3.8/site-packages/webdav3/client.py", line 66, in _wrapper
    res = fn(self, *args, **kw)
  File "/Users/lucabutera/dvc/.env/lib/python3.8/site-packages/webdav3/client.py", line 438, in upload_to
    self.execute_request(action='upload', path=urn.quote(), data=buff)
  File "/Users/lucabutera/dvc/.env/lib/python3.8/site-packages/webdav3/client.py", line 226, in execute_request
    raise ResponseErrorCode(url=self.get_url(path), code=response.status_code, message=response.content)
webdav3.exceptions.ResponseErrorCode: Request to https://<user>@drive.switch.ch/remote.php/dav/files/<user>/datasets/d4/1d8cd98f00b204e9800998ecf8427e failed with code 411 and message: b'<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">\n<html><head>\n<title>411 Length Required</title>\n</head><body>\n<h1>Length Required</h1>\n<p>A request of the requested method PUT requires a valid Content-length.<br />\n</p>\n<hr>\n<address>Apache/2.4.18 (Ubuntu) Server at a01.drive.switch.ch Port 80</address>\n</body></html>\n'
```

Since I don't see the ```Content-lenght``` explicitly set anywhere I think it gets set somewhere at a lower level when the size is other than 0. Can this be the problem?

Regarding you last suggestion, to use ```Tqdm.wrapattr```, I am not sure what you mean. I can try that if you elaborate further.

In the end, do we wish that an empty file gets uploaded anyway?

skshetry: It'd be great, if we could support this. It seems if the file is empty, `requests` sends a chunked-encoding request anyway, which is what is causing the error. Let's ignore this for now, as it only affects you.

skshetry (approved):

PR #4832, opened 2020-11-03
utils/fs: checking files ownership in 'move' (#4348)

### LINKED ISSUE #2992: better message when trying to link from files that belong to another user

### LINKED ISSUE #4348: Parquet file owned by other user gets deleted upon dvc add

danielfleischer:
Files are checked for ownership by trying to `chmod` them before moving them around; if fail, return a verbose exception.
Previous behavior: moving files to a temp folder and then do the check; when fail, files seemed to be missing.

Should fix #4348 and maybe fix #2992.

### DIFF
--- a/dvc/utils/fs.py
+++ b/dvc/utils/fs.py
@@ -106,7 +106,7 @@ def move(src, dst, mode=None):
 
     if os.path.islink(src):
         shutil.copy(src, tmp)
-        os.unlink(src)
+        _unlink(src, _chmod)
     else:
         shutil.move(src, tmp)
 
--- a/dvc/utils/fs.py
+++ b/dvc/utils/fs.py
@@ -95,14 +95,18 @@ def move(src, dst, mode=None):
     dst = os.path.abspath(dst)
     tmp = f"{dst}.{uuid()}"
 
+    try:
+        if mode is not None:
+            os.chmod(src, mode)
+    except OSError:
+        # File not owned by us, raise exception
+        raise DvcException(f"File not owned by user:  {src}")
+
     if os.path.islink(src):
         shutil.copy(src, tmp)
         os.unlink(src)
     else:
         shutil.move(src, tmp)
-
-    if mode is not None:
-        os.chmod(tmp, mode)
 
     shutil.move(tmp, dst)
 
--- a/dvc/utils/fs.py
+++ b/dvc/utils/fs.py
@@ -6,9 +6,10 @@ import stat
 import sys
 
 import nanotime
+from funcy import reraise
 from shortuuid import uuid
 
-from dvc.exceptions import DvcException
+from dvc.exceptions import DvcException, FileOwnershipError
 from dvc.system import System
 from dvc.utils import dict_md5
 
@@ -95,11 +96,9 @@ def move(src, dst, mode=None):
     dst = os.path.abspath(dst)
     tmp = f"{dst}.{uuid()}"
 
-    try:
+    with reraise(OSError, FileOwnershipError(src)):
         if mode is not None:
             os.chmod(src, mode)
-    except OSError:
-        raise DvcException(f"File not owned by user:  {src}")
 
     if os.path.islink(src):
         shutil.copy(src, tmp)
--- a/dvc/utils/fs.py
+++ b/dvc/utils/fs.py
@@ -99,7 +99,6 @@ def move(src, dst, mode=None):
         if mode is not None:
             os.chmod(src, mode)
     except OSError:
-        # File not owned by us, raise exception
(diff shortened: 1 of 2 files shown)


dvc/utils/fs.py
```diff
@@ -95,15 +95,19 @@ def move(src, dst, mode=None):
     dst = os.path.abspath(dst)
     tmp = f"{dst}.{uuid()}"
 
+    try:
+        if mode is not None:
+            os.chmod(src, mode)
+    except OSError:
+        # File not owned by us, raise exception
```
pared, line 102: The exception string seems to pass the same information. Do we need the comment?

pared: Also, since we are only raising another exception, we can make use of `reraise`, an example is [here](https://github.com/iterative/dvc/blob/411b38896014ae004d0782e25dbd49f5d33af2b7/dvc/external_repo.py#L170)

danielfleischer: 1. Good point. 
2. Not sure what you meant so I introduced a new exception and used `reraise`.

dvc/utils/fs.py
```diff
@@ -95,15 +96,16 @@ def move(src, dst, mode=None):
     dst = os.path.abspath(dst)
     tmp = f"{dst}.{uuid()}"
 
+    with reraise(OSError, FileOwnershipError(src)):
```
pared, line 99: We could move that below `mode` check.

pared (approved):
Looks great! One minor thing

dvc/utils/fs.py
pared: EDIT: @danielfleischer I am sorry for messing up, I just realized that your original approach is better:
So, the fact that we catch `OSError` does not mean that its a problem with permissions. To make this error accurate, we should actually check for error code.
So, I would recommend rolling back to the first version of change, 
catch the `OSError as e` and then check whether the error code is not `errno.EPERM` (and possibly `errno.EACCES`)
Just like we are doing [here](https://github.com/iterative/dvc/blob/411b38896014ae004d0782e25dbd49f5d33af2b7/dvc/utils/fs.py#L118).
And only then raise `FileOwnershipError`
[Error codes reference](https://docs.python.org/3.8/library/errno.html)

pared (changes requested):

dvc/utils/fs.py
danielfleischer: File not owned by user will lead to `errno.EPERM`. The question is are there more cases that will lead to permission error which are not related to ownership and then decide what's more helpful for the user: permission problem or ownership problem.

pared (approved):

pared:
@danielfleischer looks good, the tests `test_add_symlink_file` seems to be failing on windows due to this change.

danielfleischer:
Made a small fix; used `_unlink` instead, it's safer. Now all Windows tests pass.

pared (approved):

skshetry (approved):

pared:
@danielfleischer Thank you very much!

Issue #3256, opened 2020-01-30
Incorporate file sizes in dvc file


dmpetrov:
If we add file sizes in DVC-files (when we calculate checksum - so, no extra reads) it will help us to show this info in `dvc diff`/`dvc list` and other commands with no I/O or computational overhead.

Related to #2982

shcheklein:
I like the idea! On the other hand it might complicate the merge, diffs will become bigger? Also, are there any other fields that potentially could be useful (names? modes? type?).

Just a though to consider - does it makes sense to revisit the discussion around taking md5 files hashes out? Potentially all this meta information can go into the same place.

mroutis:
> Just a though to consider - does it makes sense to revisit the discussion around taking md5 files hashes out? Potentially all this meta information can go into the same place.

Yes :slightly_smiling_face:

dmpetrov:
> does it makes sense to revisit the discussion around taking md5 files hashes out? Potentially all this meta information can go into the same place.

💯

efiop:
For the record, might be solved in https://github.com/iterative/dvc/issues/1871 .

pmrowla:
From #2982
> we should store file size in addition to hashes. It will give us an ability to show file sizes in `dvc diff`.

- [ ] store file size together with checksums (it might be a separate issue)
- [ ] display file size in diff (like metrics diff). It might be disabled by default.

MetalBlueberry:
I've noticed that the size of the whole tracked dir is registered in .dvc file. but there is no data for individual files inside the tracked directory. I would like to have this information so I can display a list of files&size inside the directory without downloading any file.

is it possible to add the size to the json file generated to track contents inside the directory?

efiop:
Hi @MetalBlueberry !

Great question! Indeed, we are thinking about adding size to the .dir cache file, but adding those right now will result in older dvc versions registering it as a cache corruption and also us not being able to self-validate .dir files without filtering them first (md5 of them shouldn't depend on `size` fields) https://github.com/iterative/dvc/issues/4841

We will also add support for these to `dvc diff/list/status`.

efiop:
Most likely will be replaced by https://github.com/iterative/dvc/issues/8884 , but if .dir-s stay in some shape or form we'll redesign them from scratch. So closing this.

PR #4836, opened 2020-11-03
dvc: add size/nfiles for deps/outs

efiop:
Related to #3256

https://github.com/iterative/dvc.org/pull/1907

### DIFF
--- a/dvc/stage/serialize.py
+++ b/dvc/stage/serialize.py
@@ -134,18 +134,20 @@ def to_pipeline_file(stage: "PipelineStage"):
 def to_single_stage_lockfile(stage: "Stage") -> dict:
     assert stage.cmd
 
+    def _dumpd(item):
+        ret = [
+            (item.PARAM_PATH, item.def_path),
+            *item.hash_info.to_dict().items(),
+        ]
+
+        if item.size is not None:
+            ret.append((item.PARAM_SIZE, item.size))
+        return OrderedDict(ret)
+
     res = OrderedDict([("cmd", stage.cmd)])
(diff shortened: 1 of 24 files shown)


tests/func/test_add.py
```diff
@@ -272,6 +279,8 @@ def test_add_external_dir(tmp_dir, dvc, workspace, hash_name, hash_value):
     assert (tmp_dir / "dir.dvc").read_text() == (
         "outs:\n"
         f"- {hash_name}: {hash_value}\n"
+        "  size: 11\n"
+        "  nfiles: 2\n"
         "  path: remote://workspace/dir\n"
```
efiop, line 284: This should really be:
```
- path: ...
  md5: ...
  size: ...
  nfiles: ...
```
added a checkbox to https://github.com/iterative/dvc/issues/4841 , but we could do this right away, as it will only affect the newly written files, which is a bit annoying but not that bad.

dvc/stage/serialize.py
```diff
@@ -134,18 +134,18 @@ def to_pipeline_file(stage: "PipelineStage"):
 def to_single_stage_lockfile(stage: "Stage") -> dict:
     assert stage.cmd
 
+    def _dumpd(item):
+        ret = [
+            (item.PARAM_PATH, item.def_path),
+            *item.hash_info.to_dict().items(),
+        ]
+
+        return OrderedDict(ret)
```
efiop, line 143: This should eventually turn into a special Dumper for outputs or into `Output.dumpd_lock`(or smth like that). We've discussed it before.

Issue #4821, opened 2020-11-01
Checkpoint: separate model tuning and redesign behavior by a new command


dmpetrov:
The current checkpoint behavior works nice in the model tuning stage but it is very painful to make code/model redesign changes and start training from scratch. It seems like we need to separate these types of users' activities by two different commands `dvc exp run` and `dvc exp continue`.

Another confusion with the current design - it is not correct to require checkpoints to be `--always-changed`. They should be reproducible by default. `dvc exp run` should start from scratch if no direct ask for `--continue`. Also, I’m not 100% sure that `--persist` is the right behavior. It feels like checkpoint is a special type of outputs.

**How checkpoint reproducibility should look like**:
```
$ dvc exp run # runs (let say) 27 epochs then Ctrl+C
$ dvc exp run # it should recognize checkpoints and say something like:
Stage 'train' didn't change. To continue training checkpoints use 'dvc exp continue' command.
```

**What is the next after canceling:**
1. **Continue training with no changes**. `dvc exp continue` - starts from the last step. It should add new checkpoints to the same experiment (the same level of hierarchy)
2. **Change workspace**. `vi train.py` & `dvc exp run` will start training from scratch (in a regular way with removing all outputs)
3. **Change workspace and continue**. `vi train.py` & `dvc exp continue` should _fail_ and say that it is not possible since workspace was changed - please run another experiment `dvc exp run`
4. **Manual checkout of a checkpoint**. `dvc exp checkout e4253be`, changing code/params `vi train.py` and run it as a separate experiment: it can be a new training `dvc exp run` or continue with the latest checkpoint `dvc exp continue`. In any of these cases, it creates a new experiment (it would be amazing to show it as a new level in the hierarchy since it was inherited from another experiment - nice-to-have priority).
5. **Automatic checkpoint change**. The same as the previous one but automated with hyper params like `dvc exp continue --params train.dropout 0.18 e4253be` or starting from scratch `dvc exp run --params train.dropout 0.18 e4253be`

UPDATE 11/2 1am PST: I forgot to specify sha `e4253be` in the last scenario - updated.

Any feedback is highly appreciated. CC @iterative/engineering

pmrowla:
It seems to me like we will end up needing to define a proper checkpoint output type after all, rather than using the `persist` + `always_changed` workaround to get the internal behavior we need.

Just to clarify the current behavior, regarding 

[...]

PR #4845, opened 2020-11-05
experiments: add checkpoint output type

pmrowla:

Related to #4821, prerequisite for checkpoints CLI reorganization

* Defines `checkpoint` output type
    * Internal behavior for stages with checkpoint outs is currently equivalent to `out.persist = True` and `stage.always_changed = True`
    * In the future checkpoint outs should have unique behavior: stages w/checkpoint outs should be run-cachable (`persist` disables run cache) and stages should not be marked always changed (this will require limited support for circular dependencies in DVC)
* `dvc run -c/--checkpoints <output_path> ...` can be used to generate pipeline file w/checkpoint outputs
    * `--checkpoints` from the CLI also implies `--no-exec` since checkpoint stages are expected to be long running/manually killed operations, and normally only intended to be reproduced inside `dvc exp run`


pmrowla:
Not sure if run should use `-c/--outs-checkpoint` instead?

dvc/repo/reproduce.py
```diff
@@ -13,11 +13,19 @@
...
-        kwargs["checkpoint_func"] = partial(_run_callback, checkpoint_func)
+    if stage.is_checkpoint:
+        if checkpoint_func:
+            kwargs["checkpoint_func"] = partial(_run_callback, checkpoint_func)
+        else:
+            logger.warning(
+                "Checkpoint stages are not fully supported in 'dvc repro'. "
+                "Checkpoint stages should be reproduced with 'dvc exp run' "
+                "or 'dvc exp resume'."
+            )
```
pmrowla, line 28: I considered erroring out here and went with the warning in the end, but I don't feel too strongly either way.

Running a stage w/checkpoint outs from `dvc repro` won't actually break anything, it will just be handled the same as a stage with persist outs - the main difference being that depending on the user code this may be a command that runs forever until manually killed.

pmrowla:
Merging since the internal functionality is required for the rest of the checkpoints cleanup. UI/CLI flag changes can be made later as needed

skshetry:
@pmrowla, I'm sorry if you have already mentioned this elsewhere, but I am having trouble understanding the need for storing checkpoint outputs in the dvc.yaml itself. Is this for _sharing_ the experiment progress with others? Could that be kept in it's own vault (`git-stash`) separately?

pmrowla:
@skshetry checkpoints are an output type that need to be preserved like any other output, and they aren't exactly the same as persist outputs (although currently the difference doesn't really matter since we don't support using the previous commit's output as a true dependency for the subsequent commit).

@dmpetrov has suggested that we could maybe consider making `checkpoint` a stage level flag rather than output level though.

PR #4848, opened 2020-11-05
dvc: include cache urls in info output.

### LINKED ISSUE #4574: dvc version: show external cache and remotes being used

codito:
Show caches and remotes in the info output to assist in diagnosis of issues.

Fixes #4574.

Sample output

```sh
> dvc version
DVC version: 1.9.1+8f6350
Platform: Python 3.8.6 on Linux-5.9.6-arch1-1-x86_64-with-glibc2.2.5
Supports: All remotes
Cache directory: ext4 on /dev/sda4
Caches: ssh, local
Remotes: ssh, azure
Workspace directory: ext4 on /dev/sda4
Repo: dvc, git
```

This change will require an update to the docs for [version](https://dvc.org/doc/command-reference/version) command. Will create a docs PR if the output format looks good for this change.

---

### DIFF
--- a/dvc/info.py
+++ b/dvc/info.py
@@ -8,7 +8,7 @@ from dvc.exceptions import DvcException, NotDvcRepoError
 from dvc.repo import Repo
 from dvc.scm.base import SCMError
 from dvc.system import System
-from dvc.tree import TREES
+from dvc.tree import TREES, get_cloud_schemes
 from dvc.utils import error_link
 from dvc.utils.pkg import PKG
 from dvc.version import __version__
@@ -51,7 +51,9 @@ def get_dvc_info():
             info.append("Cache types: " + error_link("no-dvc-cache"))
 
         info.append(f"Caches: {_get_caches(repo.cache)}")
-        info.append(f"Remotes: {_get_configured_remotes(repo)}")
+
+        remotes = ", ".join(get_cloud_schemes(repo)) or "None"
+        info.append(f"Remotes: {remotes}")
 
     except NotDvcRepoError:
         pass
--- a/dvc/info.py
+++ b/dvc/info.py
@@ -67,20 +67,11 @@ def get_dvc_info():
 
 
 def _get_caches(cache):
-    from dvc.config import SCHEMA
-
-    caches = set()
-    for cache_type in SCHEMA["cache"].keys():
-        # Ignore any non cache types, e.g. LOCAL_COMMON objects
-        if not isinstance(cache_type, str):
-            continue
-
-        # Ignore the cache types not available in the repo
-        cache_instance = getattr(cache, cache_type, None)
-        if not cache_instance:
-            continue
-
-        caches.add(cache_type)
+    caches = (
+        cache_type
+        for cache_type, cache_instance in cache.by_scheme()
+        if cache_instance
+    )
 
     # Caches will be always non-empty including the local cache
     return ", ".join(caches)
--- a/dvc/info.py
+++ b/dvc/info.py
@@ -50,6 +50,11 @@ def get_dvc_info():
         else:
             info.append("Cache types: " + error_link("no-dvc-cache"))
 
+        cache_urls = _get_cache_urls(repo.cache)
+        if cache_urls:
+            info.append("Cache paths:")
+            info.extend(f"  {c[0]}: {c[1]}" for c in cache_urls)
+
     except NotDvcRepoError:
         pass
     except SCMError:
@@ -61,6 +66,23 @@ def get_dvc_info():
             info.append(f"Workspace directory: {fs_root}")
         info.append("Repo: {}".format(_get_dvc_repo_info(repo)))
     return "\n".join(info)
+
+
+def _get_cache_urls(cache):
+    from dvc.config import SCHEMA
+
+    for cache_type in SCHEMA["cache"].keys():
+        # Ignore any non cache types, e.g. LOCAL_COMMON objects
+        if not isinstance(cache_type, str):
+            continue
+
+        # Ignore the cache types not available in the repo
+        cache_instance = getattr(cache, cache_type, None)
+        if not cache_instance:
+            continue
+
+        cache_url = cache_instance.tree.config["url"]
+        yield (cache_type, cache_url)
 
 
 def _get_linktype_support_info(repo):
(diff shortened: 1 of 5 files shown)


efiop (changes requested):
Hi @codito !

Thanks for the PR!

Seems like this is duplicating the info a little bit, as local cache and `cache directory` are the same thing pretty much
```
Cache directory: ext4 on /dev/sda4
Cache paths:
  local: /home/arun/src/gh/dvc/build/tmp/.dvc/cache
  ssh: ssh://arun@192.168.1.99/home/arun/tmp/cache
```
so what we probably want is something that will aggregate cache info by the type. E.g. (pretty rough but still)
```
Cache:
    Local:
        types: hardlink, symlink
        path: /path/to/cache (ext4 on /dev/sda4)
    SSH:
         types: reflink, hardlink
         url: ssh://example.com/path
    ...
```
or something like that. In general, this does raise the question about reconsidering `version` formatting again, since it is clearly not ideal. :slightly_frowning_face:

codito:
@efiop thanks, great idea to segregate information by cache type. I wasn't sure if multi level indents would affect readability and verbosity, below are two additional variations for us to evaluate. I think option 1 may provide us the right balance between segregation and readability. What do you think?

For the remote cache types, is it important to include the `link types`? I am not sure if `version` command should perform complex operations with side effects e.g. attempt to link/unlink for finding link types.

---

Option 1: 2 level indent with cache type specified.

```
DVC version: 1.9.1+86607a.mod
Platform: Python 3.8.6 on Linux-5.9.6-arch1-1-x86_64-with-glibc2.2.5
Supports: All remotes
Cache (local):
  Url: /home/arun/src/gh/dvc/build/tmp/.dvc/cache
  File system: ext4 on /dev/sda4
  Supported link types: hardlink, symlink
Cache (ssh):
  Url: ssh://arun@192.168.1.99/home/arun/tmp/cache
Workspace directory: ext4 on /dev/sda4
Repo: dvc, git
```

Option 2: concatenated cache info in single line

```
DVC version: 1.9.1+86607a.mod
Platform: Python 3.8.6 on Linux-5.9.6-arch1-1-x86_64-with-glibc2.2.5
Supports: All remotes
Cache (type: url, link type, file system type):
  local: /home/arun/src/gh/dvc/build/tmp/.dvc/cache, (hardlink, symlink), ext4 on /dev/sda4
  ssh: ssh://arun@192.168.1.99/home/arun/tmp/cache
Workspace directory: ext4 on /dev/sda4
Repo: dvc, git
```

skshetry:
@codito, the original issue is about showing which kinds of remote and caches are being used, i.e. `ssh`/`s3`, etc. Users won't be comfortable with sharing the "urls" to their remotes and caches.

codito:
@skshetry thanks for clarifying! It simplifies the fix by a great extent. Will the below suffice then?

```
Cache directory: ext4 on /dev/sda4
Cache protocols: local, ssh
```

codito:
Updated PR to include both caches and remotes. Since we're enumerating only the configured values, I believe it is fair to assume that existing `Cache types` and `Cache directory` fields are only for `local` cache.

```sh
> dvc version
DVC version: 1.9.1+8f6350
Platform: Python 3.8.6 on Linux-5.9.6-arch1-1-x86_64-with-glibc2.2.5
Supports: All remotes
Cache directory: ext4 on /dev/sda4
Caches: ssh, local
Remotes: ssh, azure
Workspace directory: ext4 on /dev/sda4
Repo: dvc, git
```

dvc/info.py
```diff
@@ -63,6 +66,26 @@ def get_dvc_info():
     return "\n".join(info)
 
 
+def _get_caches(cache):
```
efiop, line 69: Note that we've recently introduced `repo.cache.by_scheme` helper, that should be handy here, as you won't need to parse config SCHEMA anymore.

codito: Thanks for sharing, it simplified the logic.

dvc/info.py
```diff
@@ -111,6 +125,26 @@ def _get_supported_remotes():
     return ", ".join(supported_remotes)
 
 
+def _get_configured_remotes(repo):
+    from urllib.parse import urlparse
+
+    from dvc.scheme import Schemes
+
+    # Get supported schemes, exclude the private variables
+    available_remotes = {
```
codito, line 134: Would it make sense to move this block to `scheme.py`? We are exposing internals of another class here.

efiop: Hm, actually we don't really need use `available_remotes`.

Seems like the below code with urlparse doesn't really take into account local remotes. And also doesn't take into account `remote://` notation (this is a tribal knowledge, so no wonder it is not obvious). The most reliable way to overcome all of that is to use `get_cloud_tree`(dvc/tree/__init__.py) to resolve it for us. The only issue with it is that it tries to initialize the remote, which is not really desired, so it would be great if we cold use only the `_get_conf` part of it.

codito: thanks, introduced a new utility to enumerate cloud schemes for a repo in the latest commit. `info.py` looks much cleaner with this :)

efiop (approved):
Thank you @codito ! :pray:

jorgeorpinel:
Hi! Quite late to this party but I have some feedback: I think that `Remotes` vs. `Supports` can kind of confusing. Same with `Caches` vs. `Cache types`. Users will have to go to the docs constantly to remember what the difference is. Also, why "caches"? Can you have more than one cache? Maybe the name should be "Link types used"

Suggestion: Separate the output of the `version` command into a **general** section and a **project** section, like:

```console
$ dvc version
Platform: Python 3.8.6 on Linux-5.9.6-arch1-1-x86_64-with-glibc2.2.5

DVC Version: 1.10.1
    Remotes supported: {Always list, even "all"}
    Link types supported: hardlink, symlink, copy

Project location: /home/user/mydvcproject/
    Repo: DVC, Git
    Workspace fs: ext4 on /dev/sda4
    Cache fs: ext4 on /dev/sda4
    Link types used: ssh, local
    Remotes used: ssh, azure
```

> Other small improvements applied ^

Issue #4847, opened 2020-11-05
tree: use trie in DvcTree (or in Repo)


efiop:
We are only using outputs there and we have the metadata now, so it makes sense to just construct the trie from the repo dag and have a nice and convenient way of working with it internally instead of doing dag lookups. Potentially worth doing the same thing in the repo itself, as trie is often handier than dag (though we are already using the tree(not trie) in such places). One repo trie with both outs and deps might be a solution too.

This became relevant while working on abstracting dir_info away into itsown class and analysing the way we do DvcTree.walk() now (we use trie there already).

efiop:
Already using repo.outs_trie

PR #4851, opened 2020-11-06
dvc: use DirInfo instead of dicts/lists

efiop:
Currently, we are converting dir_info to/from lists all the time.
The reason is that dir_info is stored as list of dicts in *.dir files,
but that makes it hard to work with. In addition to that, we will likely
be changing .dir file format in the near future #829, so we need to
abstract away dir_info into something that we won't care how it will be
stored on disk.

Related #3256
Related #4847

### DIFF
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -1,21 +1,18 @@
 import json
 import logging
 from copy import copy
-from operator import itemgetter
 
 from funcy import decorator
 from shortuuid import uuid
 
 import dvc.prompt as prompt
+from dvc.dir_info import DirInfo
 from dvc.exceptions import (
     CacheLinkError,
     CheckoutError,
     ConfirmRemoveError,
     DvcException,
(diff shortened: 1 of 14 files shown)


dvc/cache/base.py
```diff
@@ -623,77 +598,19 @@ def get_files_number(self, path_info, hash_info, filter_info):
...
-            )
-
         try:
-            return sum(_getsize(entry) for entry in dir_info)
+            return sum(
+                self.tree.getsize(self.tree.hash_to_path_info(hi.value))
+                for _, hi in dir_info.items()
+            )
         except FileNotFoundError:
             return None
```
efiop, line 615: Ideally `size` should be added to dir_info entries when loading it, so that we could simply use DirInfo.size instead of doing this manual labor. Will look into it...

efiop: Ok, not at this point. Keeping as is for now.

PR #4855, opened 2020-11-06
checkpoints: `exp run` and `exp res[ume]` refactor

pmrowla:

Related to #4821.

* checkpoint stage detection is handled automatically based on whether or not a stage contains `checkpoint` outputs
* `dvc exp run --checkpoint` is replaced with `dvc exp run`
    * To force reset an existing checkpoint run from scratch, use `dvc exp run -f`.
* `dvc exp run --continue` is replaced with `dvc exp res[ume]`

### DIFF
--- a/dvc/repo/reproduce.py
+++ b/dvc/repo/reproduce.py
@@ -13,7 +13,6 @@ logger = logging.getLogger(__name__)
 def _reproduce_stage(stage, **kwargs):
     def _run_callback(repro_callback):
         _dump_stage(stage)
-        logger.debug(f"{repro_callback} ([{stage}])")
         repro_callback([stage])
 
     checkpoint_func = kwargs.pop("checkpoint_func", None)
--- a/dvc/repo/reproduce.py
+++ b/dvc/repo/reproduce.py
@@ -3,6 +3,7 @@ from functools import partial
 
 from dvc.exceptions import InvalidArgumentError, ReproductionError
 from dvc.repo.scm_context import scm_context
+from dvc.stage.run import CheckpointKilledError
 
 from . import locked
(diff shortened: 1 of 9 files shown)


dvc/repo/reproduce.py
```diff
@@ -154,33 +154,7 @@ def _reproduce_stages(
...
-                    G.copy().reverse(copy=False), stage
-                )
-                all_pipelines += reversed(list(nodes))
-            else:
-                all_pipelines += nx.dfs_postorder_nodes(G, stage)
-
-    pipeline = []
-    for stage in all_pipelines:
-        if stage not in pipeline:
-            pipeline.append(stage)
```
pmrowla, line 183: This code is just moved into `_get_pipeline()` to fix flake8 complexity error

dvc/stage/run.py
```diff
@@ -159,17 +154,31 @@ def _checkpoint_run(stage, callback_func, done_cond, proc):
...
+    if os.name == "nt":
+        return _kill_nt(proc)
+    proc.terminate()
+    proc.wait()
+
+
+def _kill_nt(proc):
+    # windows stages are spawned with shell=True, proc is the shell process and
+    # not the actual stage process - we have to kill the entire tree
+    subprocess.call(["taskkill", "/F", "/T", "/PID", str(proc.pid)])
```
pmrowla, line 176: This is something that we could consider doing via `psutil`, which has the logic needed to actually walk Windows process trees and find each child process individually (native python does not provide an easy way to do that). psutil has Windows wheels for <3.9 but still does not have any wheels for linux/mac, so if we were going to re-introduce it as a dependency it would have to be Windows only.

But calling `taskkill` here also works for our purposes

PR #4860, opened 2020-11-08
[WIP] Osf remote

### LINKED ISSUE #4176: Support Open Science Foundation?
Hi there; I love this project! What a great tool.

I'm wondering if you could add support for an [open science framework](https://osf.io/) project as a remote. It is for hosting data as part of a published research report. They have a [python api](https://github.com/osfclient/osfclient) for sending and retrieving files, so I'm *guessing* it would be pretty easy to add this feature. If it's not too difficult for someone new to this project, I would be happy to help out with adding this feature myself, if you could provide me some pointers on where to get started.

aerubanov:
close #4176 

I have implemented the OSFTree class and I open this pull request to get feedback on these changes. Now I will write the tests and make changes to the documentation.

Note on method ```_download``` in ```dvc/tree/osf.py```. The currently used osfclient library does not have a public attribute of the File class to get the file size. Although the API itself has this feature. I opend the pull request with the necessary changes https://github.com/osfclient/osfclient/pull/185. So it should be possible to add tqdm progress bar for ```_downlod``` method after next osfclient release.

### DIFF
--- a/.dvcignore
+++ b/.dvcignore
@@ -1,2 +1,0 @@
-# add patterns of files dvc should ignore, can improve performance
-# Learn more at https://dvc.org/doc/user-guide/dvcignore
--- a/.dvcignore
+++ b/.dvcignore
@@ -1,0 +1,2 @@
+# add patterns of files dvc should ignore, can improve performance
+# Learn more at https://dvc.org/doc/user-guide/dvcignore
--- a/dvc/tree/osf.py
+++ b/dvc/tree/osf.py
@@ -24,6 +24,7 @@ class OSFAuthError(DvcException):
         )
         super().__init__(message)
 
+
 class OSFTree(BaseTree):
     scheme = Schemes.OSF
     REQUIRES = {"ofsclient": "osfclient"}
--- a/dvc/tree/osf.py
+++ b/dvc/tree/osf.py
@@ -41,6 +41,8 @@ class OSFTree(BaseTree):
         self.password = os.getenv(
             "OSF_PASSWORD", None
         )  # need for private projects
+        if self.password is None:
+            self.password = config.get("password")
         logger.debug(OSFTree)
 
     @wrap_prop(threading.Lock())
--- a/dvc/tree/osf.py
+++ b/dvc/tree/osf.py
@@ -87,7 +87,6 @@ class OSFTree(BaseTree):
         return any(path_info.path == path for path in paths)
 
     def _list_paths(self):
-        # print(list(self.storage.files))
         for file in self.storage.files:
             yield file.path
 
@@ -117,16 +116,16 @@ class OSFTree(BaseTree):
         self, from_info, to_file, name=None, no_progress_bar=False, **_kwargs
     ):
         file = self._get_file_obj(from_info)
-
         # hack to get size of file
         # This will be no longer needed after the
         # pull-request https://github.com/osfclient/osfclient/pull/185 merge
-
-        # pylint: disable=W0212
-        resp = self.osf._get(file._endpoint)
-        json = self.osf._json(resp, 200)
-        total = self.osf._get_attribute(json, "data", "attributes", "size")
-        # pylint: enable=W0212
+        total = None
+        if not no_progress_bar:
+            # pylint: disable=W0212
+            resp = self.osf._get(file._endpoint)
+            json = self.osf._json(resp, 200)
+            total = self.osf._get_attribute(json, "data", "attributes", "size")
+            # pylint: enable=W0212
 
         with open(to_file, "wb") as fobj:
             with Tqdm.wrapattr(
@@ -142,6 +141,6 @@ class OSFTree(BaseTree):
             with Tqdm.wrapattr(
                 fobj, "read", desc=name, total=total, disable=no_progress_bar
             ) as wrapped:
-                self.project.storage().create_file(
+                self.storage.create_file(
                     to_info.path, wrapped, force=False, update=False
                 )
--- a/dvc/tree/osf.py
+++ b/dvc/tree/osf.py
@@ -5,8 +5,8 @@ import threading
 from funcy import cached_property, wrap_prop
 
 from dvc.exceptions import DvcException
-from dvc.hash_info import HashInfo
 from dvc.path_info import URLInfo
+from dvc.hash_info import HashInfo
 from dvc.progress import Tqdm
 from dvc.scheme import Schemes
 
@@ -23,7 +23,6 @@ class OSFAuthError(DvcException):
             " to a private project."
         )
         super().__init__(message)
-
 
 class OSFTree(BaseTree):
     scheme = Schemes.OSF
--- a/dvc/tree/osf.py
+++ b/dvc/tree/osf.py
@@ -19,8 +19,9 @@ class OSFAuthError(DvcException):
     def __init__(self):
         message = (
             "OSF authorization failed. Please check or provide"
-            " osf_username and password. See "
-            "https://bit.ly/375On32 for details"
+            " user and password. See "
+            "https://dvc.org/doc/command-reference/remote"
+            " for details"
         )
         super().__init__(message)
 
@@ -36,13 +37,9 @@ class OSFTree(BaseTree):
 
         self.path_info = self.PATH_CLS(config["url"])
 
-        self.osf_username = config.get("username")
+        self.user = config.get("user")
         self.project_guid = config.get("project")
-        self.password = os.getenv(
-            "OSF_PASSWORD", None
-        )  # need for private projects
-        if self.password is None:
-            self.password = config.get("password")
+        self.password = os.getenv("OSF_PASSWORD", config.get("password"))
         logger.debug(OSFTree)
 
     @wrap_prop(threading.Lock())
@@ -59,7 +56,7 @@ class OSFTree(BaseTree):
         from osfclient.exceptions import UnauthorizedException
 
         osf = self.osf
-        osf.login(self.osf_username, self.password)
+        osf.login(self.user, self.password)
         try:
             project = osf.project(self.project_guid)
         except UnauthorizedException:
--- a/dvc/tree/osf.py
+++ b/dvc/tree/osf.py
@@ -77,7 +77,7 @@ class OSFTree(BaseTree):
         return md5
 
     def exists(self, path_info, use_dvcignore=True):
-        paths = self.list_paths()
+        paths = self._list_paths()
         return any(path_info.path == path for path in paths)
 
     def _list_paths(self):
--- a/dvc/tree/osf.py
+++ b/dvc/tree/osf.py
@@ -5,7 +5,6 @@ import threading
 from funcy import cached_property, wrap_prop
 
 from dvc.exceptions import DvcException
-from dvc.path_info import URLInfo
 from dvc.hash_info import HashInfo
 from dvc.path_info import URLInfo
 from dvc.progress import Tqdm
--- a/dvc/tree/osf.py
+++ b/dvc/tree/osf.py
@@ -1,0 +1,92 @@
+import logging
+import os
+import threading
+
+from funcy import cached_property, wrap_prop
+
+from dvc.hash_info import HashInfo
+from dvc.progress import Tqdm
+from dvc.scheme import Schemes
+
+from .base import BaseTree
+
+logger = logging.getLogger(__name__)
+
+
+class OSFTree(BaseTree):
+    scheme = Schemes.OSF
+    REQUIRES = {"ofsclient": "osfclient"}
+    PARAM_CHECKSUM = "md5"
+
+    def __init__(self, repo, config):
+        super().__init__(repo, config)
+
+        self.osf_username = config.get("osf_username")
+        self.project = config.get("project")
+        self.password = os.getenv(
+            "OSF_PASSWORD", None
+        )  # need for private projects
+
+    @wrap_prop(threading.Lock())
+    @cached_property
+    def storage(self):
+        import osfclient
+
+        osf = osfclient.OSF()
+        osf.login(self.osf_username, self.password)
+        storage = osf.project(self.project).storage()
+        return storage
+
+    def _get_file_obj(self, path_info):
+        for file in self.storage.files:
+            if file.path == path_info.path:
+                return file
+
+    def get_md5(self, path_info):
+        file = self._get_file_obj(path_info)
+        md5 = file.hashes.get("md5")
+        return md5
+
+    def exists(self, path_info, use_dvcignore=True):
+        paths = self.list_paths()
+        return any(path_info.path == path for path in paths)
+
+    def _list_paths(self):
+        for file in self.storage.files:
+            yield file.path
+
+    def walk_files(self, path_info, **kwargs):
+        for fname in self._list_paths():
+            if fname.endswith("/"):
+                continue
+
+            yield path_info.replace(path=fname)
+
+    def remove(self, path_info):
+        if path_info.scheme != self.scheme:
+            raise NotImplementedError
+
+        file = self._get_file_obj(path_info)
+        file.remove()
+        logger.debug(f"Removing {path_info}")
+
+    def get_file_hash(self, path_info):
+        return HashInfo(self.PARAM_CHECKSUM, self.get_md5(path_info))
+
+    def _download(self, from_info, to_file, **_kwargs):
+        file = self._get_file_obj(from_info)
+        # total = file._get_attribute(file, "attributes", "self", "size")
+        with open(to_file, "wb") as fobj:
+            file.write_to(fobj)
+
+    def _upload(
+        self, from_file, to_info, name=None, no_progress_bar=False, **_kwargs
+    ):
+        total = os.path.getsize(from_file)
+        with open(from_file, "rb") as fobj:
+            with Tqdm.wrapattr(
+                fobj, "read", desc=name, total=total, disable=no_progress_bar
+            ) as wrapped:
+                self.storage.create_file(
+                    to_info.path, wrapped, force=False, update=False
+                )
--- a/dvc/tree/osf.py
+++ b/dvc/tree/osf.py
@@ -6,7 +6,7 @@ from funcy import cached_property, wrap_prop
 
 from dvc.exceptions import DvcException
 from dvc.hash_info import HashInfo
-from dvc.path_info import CloudURLInfo
+from dvc.path_info import URLInfo
 from dvc.progress import Tqdm
 from dvc.scheme import Schemes
 
@@ -29,7 +29,7 @@ class OSFTree(BaseTree):
     scheme = Schemes.OSF
     REQUIRES = {"ofsclient": "osfclient"}
     PARAM_CHECKSUM = "md5"
-    PATH_CLS = CloudURLInfo
+    PATH_CLS = URLInfo
 
     def __init__(self, repo, config):
         super().__init__(repo, config)
@@ -100,7 +100,7 @@ class OSFTree(BaseTree):
 
     def _download(self, from_info, to_file, **_kwargs):
         file = self._get_file_obj(from_info)
-        # total = file._get_attribute(file, "attributes", "self", "size")
+
         with open(to_file, "wb") as fobj:
             file.write_to(fobj)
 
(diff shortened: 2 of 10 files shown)


efiop:
Hi @aerubanov ! Great stuff! :tada: Please check fix styling as noted in the failing `lint` job. Be sure to install our `pre-commit` hooks in your dev environment, so it checks everything for you when you do any commit.

Also, added `[WIP]` prefix to this PR title, so people know that this is not ready for review.

efiop:
Btw, regarding
> Note on method _download in dvc/tree/osf.py. The currently used osfclient library does not have a public attribute of the File class to get the file size. Although the API itself has this feature. I opend the pull request with the necessary changes osfclient/osfclient#185. So it should be possible to add tqdm progress bar for _downlod method after next osfclient release.

could we for now just use a hack for it? Or is it not possible to easily access that hidden attribute from outside?

aerubanov:
Ok, I'll try.

efiop:
@aerubanov Please rebase on top of master. Also, in the PR header there was a reminder about docs.

Btw, what about tests for this? Are there any emulators?

aerubanov:
@efiop Opps, I have not completed the merger process. Thank you for noticing.

About the documentation. I was distracted by my studies, but in the next couple of days I will open a pull request at dvc.org.

As for testing. OSF provides a separate url with API for testing applications (see testing section). It looks like we can replace the url in osfclient and thus test our code. But I have my doubts about this being a good idea.For example, if such tests will not break when running in CI/CD pipline.
But without it, we can only test the part of the code that does not include interaction with the OSF server. And there is not much such code.  What do you think?

efiop:
@aerubanov Looks like something went wrong with your merge. The changes now include other people's changes, which makes it really hard to review. Could you try:
```
git remote add upstream https://github.com/iterative/dvc
git fetch upstream
git rebase upstream/master
```
and when everything looks good:
```
git push -f
```
?

efiop:
> As for testing. OSF provides a separate url with API for testing applications (see testing section https://developer.osf.io/#tag/General-Usage). It looks like we can replace the url in osfclient and thus test our code. But I have my doubts about this being a good idea.For example, won't such tests break when running in CI/CD pipline?
But without it, we can only test the part of the code that does not include interaction with the OSF server. And there is not much such code. What do you think?

So we would have to create account for testing purposes but use the test endpoint instead of the production one? If so, that might be an option, but that will definitely slow down the testing. Maybe there are some better options (emulator, mocker or smth like that). Maybe take a look at the osf library you use, it might have some good recipes.

aerubanov:
@efiop, Thank you for your advice! I found in the code of ```osfclient``` mock-objects, which I suppose allow to write tests without using API queries https://github.com/osfclient/osfclient/blob/master/osfclient/tests/test_uploading.py. I will try to implement it. I think it's better suited to our task than using test endpoint.

aerubanov:
Open PR for docs https://github.com/iterative/dvc.org/pull/1958. Starting to work on the tests.

aerubanov:
I wrote tests for osf_tree.  @efiop, please look at my code and specify what needs to be improved.

dvc/tree/osf.py
```diff
@@ -0,0 +1,146 @@
...
+            " osf_username and password."
+            " It is also possible that you do not have access"
+            " to a project."
+        )
+        super().__init__(message)
+
+
+class OSFTree(BaseTree):
+    scheme = Schemes.OSF
+    REQUIRES = {"ofsclient": "osfclient"}
```
jorgeorpinel, line 31: Hi. So this will require the user to `pip install osfclient` themselves? Or can it be an optional dependency when you `pip install dvc[osf]`?

dvc/config.py
```diff
@@ -178,6 +179,12 @@ class RelPath(str):
                     "grant_full_control": str,
                     **REMOTE_COMMON,
                 },
+                "osf": {
+                    "osf_username": str,
+                    "project": str,
+                    "password": str,
```
jorgeorpinel, line 185: Should `osf_username` just be `username` like all types?

dvc/tree/osf.py
```diff
@@ -0,0 +1,146 @@
...
+logger = logging.getLogger(__name__)
+
+
+class OSFAuthError(DvcException):
+    def __init__(self):
+        message = (
+            "OSF authorization failed. Please check or provide"
+            " osf_username and password."
+            " It is also possible that you do not have access"
+            " to a project."
```
jorgeorpinel, line 24: This should be enough:
```suggestion
            "OSF authorization failed. Please check or provide"
            " your username and password."
```
Althouth it could also use "See {short link to https://dvc.org/doc/command-reference/remote/modify} for details."

setup.py
```diff
@@ -152,6 +155,7 @@ def run(self):
     "filelock",
     "black==19.10b0",
     "wsgidav",
+    "osfclient==0.0.4",
```
jorgeorpinel, line 158: So this will require the user to `pip install osfclient` themselves? Or can it be an optional dependency when you `pip install dvc[osf]`?

jorgeorpinel: p.s. asking per iterative/dvc.org/pull/1958

skshetry: @jorgeorpinel, this is only for the tests. There's `osf` above already.

jorgeorpinel: OK. My question is in general about the package though, not this specific line. Thanks

setup.py
```diff
@@ -100,12 +99,16 @@ def run(self):
 hdfs = ["pyarrow>=2.0.0;  python_version < '3.9'"]
 webhdfs = ["hdfs==2.5.8"]
 webdav = ["webdavclient3>=3.14.5"]
+osf = ["osfclient==0.0.4"]
```
skshetry, line 102: @jorgeorpinel, this is what makes `dvc[osf]` works. So, it's optional. The below one is for the tests.

aerubanov:
I changed osf_username to username and changed the message in OSFAuthError.

efiop:
> I changed osf_username to username and changed the message in OSFAuthError.

@aerubanov Let's do `user`, because other remotes have it already (e.g. `ssh`)

.dvcignore
```diff
@@ -0,0 +1,2 @@
+# add patterns of files dvc should ignore, can improve performance
```
shcheklein, line 1: does DVC creates this file automatically? Probably we don't need it in this PR?

dvc/tree/osf.py
```diff
@@ -0,0 +1,145 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class OSFAuthError(DvcException):
+    def __init__(self):
+        message = (
+            "OSF authorization failed. Please check or provide"
+            " osf_username and password. See "
+            "https://bit.ly/375On32 for details"
```
shcheklein, line 23: hmm, could we use the dvc.org/doc troubleshooting mechanism here? bitly in the exceptions looks a bit strange, probably we don't want to depend on bitly

dvc/tree/osf.py
```diff
@@ -0,0 +1,145 @@
...
+from .base import BaseTree
+
+logger = logging.getLogger(__name__)
+
+
+class OSFAuthError(DvcException):
+    def __init__(self):
+        message = (
+            "OSF authorization failed. Please check or provide"
+            " osf_username and password. See "
```
shcheklein, line 22: is it now `username`?

dvc/tree/osf.py
```diff
@@ -0,0 +1,145 @@
...
+        self.path_info = self.PATH_CLS(config["url"])
+
+        self.osf_username = config.get("username")
+        self.project_guid = config.get("project")
+        self.password = os.getenv(
+            "OSF_PASSWORD", None
+        )  # need for private projects
+        if self.password is None:
+            self.password = config.get("password")
+        logger.debug(OSFTree)
```
shcheklein, line 46: can be too verbose? should we make a message that includes what we need?

dvc/tree/osf.py
```diff
@@ -0,0 +1,145 @@
...
+    @cached_property
+    def project(self):
+        from osfclient.exceptions import UnauthorizedException
+
+        osf = self.osf
+        osf.login(self.osf_username, self.password)
+        try:
+            project = osf.project(self.project_guid)
+        except UnauthorizedException:
+            raise OSFAuthError
```
shcheklein, line 66: should we chain exceptions?

dvc/tree/osf.py
```diff
@@ -0,0 +1,145 @@
...
+        osf = osfclient.OSF()
+        return osf
+
+    @wrap_prop(threading.Lock())
+    @cached_property
+    def project(self):
+        from osfclient.exceptions import UnauthorizedException
+
+        osf = self.osf
+        osf.login(self.osf_username, self.password)
```
shcheklein, line 62: why is it osf_username, but `password` (no `osf_` prefix)?

dvc/tree/osf.py
```diff
@@ -0,0 +1,145 @@
...
+        if not no_progress_bar:
+            # pylint: disable=W0212
+            resp = self.osf._get(file._endpoint)
+            json = self.osf._json(resp, 200)
+            total = self.osf._get_attribute(json, "data", "attributes", "size")
+            # pylint: enable=W0212
+
+        with open(to_file, "wb") as fobj:
+            with Tqdm.wrapattr(
+                fobj, "read", desc=name, total=total, disable=no_progress_bar
```
shcheklein, line 131: hmm, should it be `write` here?

setup.py
```diff
@@ -100,12 +99,16 @@ def run(self):
...
 # gssapi should not be included in all_remotes, because it doesn't have wheels
 # for linux and mac, so it will fail to compile if user doesn't have all the
 # requirements, including kerberos itself. Once all the wheels are available,
 # we can start shipping it by default.
 ssh_gssapi = ["paramiko[invoke,gssapi]>=2.7.0"]
-all_remotes = gs + s3 + azure + ssh + oss + gdrive + hdfs + webhdfs + webdav
+all_remotes = (
+    gs + s3 + azure + ssh + oss + gdrive + hdfs + webhdfs + webdav + osf
```
shcheklein, line 109: I wonder if at some point we'll have to do multiple bundles? DVC can become too heavy.

.dvcignore
aerubanov: I suppose I accidentally run ``dvc init`` and this file was generated. I'll delete it.

dvc/tree/osf.py
aerubanov: Can I see somewhere an example of how to use the dvc.org/doc troubleshooting mechanism? Does it somehow allow you not to use a long link (like https://dvc.org/doc/command-reference/remote) in a error message?

dvc/tree/osf.py
```diff
@@ -0,0 +1,145 @@
...
+            if file.path == path_info.path:
+                return file
+
+    def get_md5(self, path_info):
+        file = self._get_file_obj(path_info)
+        md5 = file.hashes.get("md5")
+        return md5
+
+    def exists(self, path_info, use_dvcignore=True):
+        paths = self._list_paths()
```
shcheklein, line 85: this can be extremely inefficient if remote is large. Is there a call that checks a singe file?

DVC most likely expects for remotes to implement `exists` as a single call that check one specific file. Is it possible here?

In general for all remotes we have a pretty good mechanism that given a few settings and "pin-point" `exists` + `walk_files` implementations can decide which specific mechanism to use to run `dvc status`, for example (a lot of exists calls or one single list).

@pmrowla could you take a look and recommend something in this case please?

shcheklein:
@aerubanov thanks for contributing this 🙏 

A few questions (most of them due to lack of knowledge on my end anything about OSF, so sorry if they look stupid) and some notes to check:

1. `username/password` - is it the only auth flow possible for the service? No OAuth or something like that
2. What are the API rate limits, size limits for the service?
3. We should handle gracefully `import-url`, `dvc add --external osf://something`
4. Is there only way for OSF to do full file listings? (to remove, to do pretty much anything)?

dvc/tree/osf.py
shcheklein: yep, please try to grep in sources links like `error.dvc.org` . E.g. `https://error.dvc.org/missing-files` goes to the docs page. You can make a section on that page that provides more details, additional links, etc. 

Or in this specific case we might just provide a link to the `dvc remote modify` page? It'll have the OSF section?

setup.py
aerubanov: But you can use ``pip install dvc[osf]`` to use dvc with the selected remote storage. I suppose that ```all_remotes``` is mainly needed to install the development and testing environment.

dvc/tree/osf.py
aerubanov: @shcheklein, @pmrowla  every item in OSF storage (folder or file) is a node of tree and store links to a childes. Class ```Storage``` in ```osfclient``` lib provide ```files``` atribute (iterator) to iterate ower these tree.  Is there any way we can determine which folder the file stored in based on the ```path_info``` value? If so, I think it is possible to avoid going through the whole tree here.

aerubanov:
@shcheklein ,
1. For acces to OSF API it is possible to use username/password or token. But auth by token for `osfclient` still in development. There are other ways to authorize for the web client, but for us they are not relevant.
2 There is limit for storage size: 5 GB for private project and 50GB for public. I haven't seen any mention of API rate limit.
3 In fact, the API endpoint is encapsulated in `osfclient'. So I used a url like `osf://osf.io/path/to/dir' in the PR documentation for clarity. But 'osf.io' can be replaced with something else and nothing will break. If the question was about that.
4 See my answer above.

dvc/tree/osf.py
shcheklein: @aerubanov yes, `path_info` is the regular path pretty much - `a/b/c`, so you do know `a/b` - directories path. Is it enough to check if efficiently?

> Class Storage in osfclient lib provide files atribute (iterator) to iterate ower these tree. 

so, is it lazy? or does it fetch the whole tree from the remote when you first initialize this?

setup.py
shcheklein: you can. The problem that we have brew, deb, win, dmg, and a few other ways we deliver this. And those include all the dependencies.

shcheklein:
@aerubanov thanks for the explanation!

> There is limit for storage size: 5 GB for private project and 50GB for public.

what would be the main use case for this integration? upload machine learning models? Is there any bandwidth limits?

>  I haven't seen any mention of API rate limit.

can we ask them? it would be pretty bad user experience if we hit it and `dvc pull` will start failing. Similar to the Google Drive we can retry on those. It's a matter of single decorator usually.

if `rclone` supports it, I would check their docs and implementation code also

> If the question was about that.

no, there are "advanced" DVC commands, like `dvc import-url` or `dvc add --external`. No need to implement them. But we should make sure that we raise NonImplemented or something.

aerubanov:
@shcheklein, I believe that the main use case for  OSF as dvc remote is to share data and models for publications. OSF allows you to get DOI for you project and cite it in a papers. Also,  5 GB for private project and 50GB for public avaible for free. It is possible to increase storage size with add-on storage connection (S3, GoogleDrive, Dropbox etc.). I have understood what the problem with API limits can be. I will try to deal with this issue in more detail.  So far I have only found additional limits on the size of one file ([5GB by derfault](https://help.osf.io/hc/en-us/articles/360019737894#what-is-the-individual-file-size-limit), can be vary for add-ons). 

As for `rclone` support, I found only issue with this feature request https://github.com/rclone/rclone/issues/3754.

OK, I'll check what I need to do to work properly with `dvc import-url` and other.

dvc/tree/osf.py
aerubanov: Thank you for the advice, I'll see what needs to be done.

aerubanov: Yes, I opened PR with changes in docs and there will be a section for OSF on `dvc remote modify` page.

dvc/tree/osf.py
aerubanov: Yes, it is lazy. I think that we can firstly iterate over folders, and then iterate over files only in selected folder to find some specific file. It will be faster then iterate over all files in storage.

aerubanov: I need to take another look at the `osfclient' code to be 100% sure, but I think I can implement it.

aerubanov:
@shcheklein, I did the following:
1. Changed the `_get_file_obj()` method to make it more efficient as we discussed above. The `exist()` method now uses it too.
2. Add `open()` method which raise `RemoteActionNotImplemented` for handle `dvc import-url`. But i do not sure thet it actually need, becouse implementation  `dvc.tree.base` already do this. 
3. About API rate limit. There is a limit on the total number of requests per day - 10 000 and also with a large number of requests to to download/upload files error 429 can be obtained. But `osfclient` prevents this by adding a period of time between such requests. I don't know, do we still have to handle such situations ourselves? Considering that uploading/downloading can be in several threads.

Please write what you think about it.

shcheklein:
@aerubanov thanks!

> Changed the _get_file_obj() method to make it more efficient as we discussed above. The exist() method now uses it too.

yep, good. Still very inefficient to my mind, but it looks like it's a limitation of the Python client library (not sure about API itself). I can't dig into this unfortunately.

> Add open() method which raise RemoteActionNotImplemented for handle dvc import-url.

I would just try to run it and `dvc run -d osf://something` and see if exception makes sense (not implemented). If it loooks good - then we are fine.

>  There is a limit on the total number of requests per day - 10 000

:( (this is very low) ... what happens after? 429 - infinite wait?

> I don't know, do we still have to handle such situations ourselves?

Probably not, if library handles this already.

> Also, 5 GB for private project and 50GB for public avaible for free.

what happens after - do we handle this gracefully?


Next steps would be for someone more familiar with the code base to review this (cc @efiop ), and @pmrowla help us setup remote specific variables to make.

aerubanov:
> but it looks like it's a limitation of the Python client library (not sure about API itself)

I would say that this is a limitation of the API itself.

> :( (this is very low) ... what happens after? 429 - infinite wait?

Yes. But once again, this is for free storage. And I don't think anybody will use OSF as the main remote. I see the following scenario (mainly for research): when you get to the final version, you combine code from the Github and dvc data in the OSF project, freeze it and assign a DOI. This allows you to share your research.

Perhaps I should to specify this in the documentation as well.

> what happens after - do we handle this gracefully?

I think I should definitely do it.

@shcheklein, thanks for your comments!

aerubanov:
I added code for correct handling of `import-url`.

 I also found that when the number of requests/storage volume exceeded, `osfclient` raise RuntimeError with message "Response has status code 4XX". We can catch it and display message with link on docs (https://developer.osf.io/#tag/Errors-and-Error-Codes for example). But I do not sure that it actually needed. What do you think?

shcheklein:
> I also found that when the number of requests/storage volume exceeded, osfclient raise RuntimeError with message "Response has status code 4XX". We can catch it and display message with link on docs (https://developer.osf.io/#tag/Errors-and-Error-Codes for example). But I do not sure that it actually needed. What do you think?

hmm ... does it have any other explanation besides the error code?

it's definitely not the best user experience to my mind, if we can wrap it (don't like that we'll be wrapping a built-in exception - I would raise a ticket with them to improve it) let's do this. It's a matter of not even providing more information, but also the way we present this - if we wrap into a DvcException (or something inherited) it doesn't "unexpected crash" anymore, it's an expected limit.

aerubanov:
Add handling of API errors.

aerubanov:
@efiop, @pmrowla I would like to ask your advice on further code changes. @shcheklein mentioned above that need to set up remote specific variables to make.

efiop:
Hi @aerubanov ! I'm sorry, but we are working on preparing 2.0 release for January right now and we might not have the capacity to get back to this PR until then. We won't be able to merge it until 2.0 anyway, as there is a feature freeze in place, so this might have to wait for now. We'll get back to this sooner if we'll have the opportunity. Really sorry for the inconvenience. :slightly_frowning_face:

aerubanov:
@efiop, Okay, I understand, don't worry about it. Thanks for the quick response!

efiop:
Hi @aerubanov ! Quick update: we are migrating to [fsspec](https://github.com/iterative/dvc/issues/5162) and that will be our plugin mechanism for the future. We won't be able to provide proper support for osf ourselves, so the best way to go about it would be for you to implement osffs using fsspec and maintain that implementation (try registering it in fsspec, other people might find it useful too). On our end, we will allow using that as a plugin as seamlessly as possible.

PR #4862, opened 2020-11-08
gdrive import-url

julsal:

This PR addresses the `import-url` command for gdrive urls. It's important to highlight two aspects of this maintenance:

1. [Google Drive API always requires authentication](https://developers.google.com/drive/api/v3/about-auth), even if the gdrive:// url is publicly available.  There are two heterodox alternatives (none of them included in this PR) to be considered:

    -  keep internally a token associated with a kind of default DVC account to be used in the case of publicly available urls.
    -  convert the user url into the format `https://www.googleapis.com/drive/v2/files/FILE_ID_HERE?key=YOUR_API_KEY` and [download it without the GDrive API](https://stackoverflow.com/questions/31840617/google-drive-requires-authentication-despite-public-folder-access). In this case, we cannot access in advance any metadata, such as file name or hash.

2. Different from other urls, the gdrive url doesn't contain the file name on it. It's just gdrive:// followed by a meaningless code. As 
 the import function defines the local file name very early in the execution, far from the moment when we can access GDrive service to request the file name, two functions were refactored: `stage.create_stage` and `utils.resolve_paths`. The refactoring represents most of the effort associated with this PR.


dvc/tree/gdrive.py
```diff
@@ -128,16 +118,17 @@ def __init__(self, repo, config):
...
-        self._gdrive_user_credentials_path = (
-            tmp_fname(os.path.join(self.repo.tmp_dir, ""))
-            if os.getenv(GDriveTree.GDRIVE_CREDENTIALS_DATA)
-            else config.get(
-                "gdrive_user_credentials_file",
-                os.path.join(
-                    self.repo.tmp_dir, self.DEFAULT_USER_CREDENTIALS_FILE,
-                ),
+
+        if os.getenv(GDriveTree.GDRIVE_CREDENTIALS_DATA):
```
shcheklein, line 122: @julsal is it a purely cosmetic change? could we please, keep changes like this for a separate PR? Two reasons- they are subjective and might create some discussion that are not relevant, they extremely slow down the review process since you have to spend some time looking into them to see if there is some relevant logic changes. Let's keep the for the next, housekeeping iteration please. Let's try to keep only relevant changes in this (non small already PR). Thanks 🙏

dvc/path_info.py
```diff
@@ -341,5 +357,16 @@ class WebDAVURLInfo(URLInfo):
     @cached_property
     def url(self):
         return "{}://{}{}".format(
-            self.scheme.replace("webdav", "http"), self.netloc, self._spath
+            self.scheme.replace("webdav", "http"), self.netloc, self.spath
         )
+
+
+class GDriveURLInfo(CloudURLInfo):
```
shcheklein, line 364: @julsal moving it here is - first, a cosmetic change, that I would not do as part of this PR, second, is arguable- e.g. my preference would be to move WebDav out of it- so that logic that is very specific to a particular cloud should stay in that cloud

dvc/tree/gdrive.py
```diff
@@ -510,19 +509,29 @@ def _path_to_item_ids(self, path, create, use_cache):
...
+
         try:
-            self._get_item_id(path_info)
-        except FileMissingError:
+            item_id = self._get_item_id(path_info)
+            if isinstance(path_info, GDriveURLInfo):
+                file_repr = self._drive.CreateFile({"id": item_id})
+                file_repr.FetchMetadata()
+
+        except (FileMissingError, ApiRequestError):
```
shcheklein, line 534: ApiRequestError -> `false` transition doesn't look right to me, ApiRequestError might mean pretty much everything, probably we even have some retry logic that will be broken by this - need to double check this

dvc/tree/gdrive.py
```diff
@@ -510,19 +509,29 @@ def _path_to_item_ids(self, path, create, use_cache):
         )
 
     def _get_item_id(self, path_info, create=False, use_cache=True, hint=None):
-        assert path_info.bucket == self._bucket
-
-        item_ids = self._path_to_item_ids(path_info.path, create, use_cache)
-        if item_ids:
-            return min(item_ids)
+        if isinstance(path_info, GDriveURLInfo):
```
shcheklein, line 512: not sure I understand this logic, to be honest ... GDriveURLInfo can be pretty much anything, not only `grive://ID` - it can be `gdrive://root/some/path/here`, etc

dvc/tree/gdrive.py
julsal: Sure! I found it difficult to understand, so I ended up changing. But, I'll make the PR free of cosmetics.

dvc/tree/gdrive.py
julsal: In this case, the logic is definitively limited. I thought `gdrive://ID` was the unique form of gdrive url.

dvc/tree/gdrive.py
julsal: Good point! I'll check a more precise form to check the file doesn't exist.

dvc/path_info.py
julsal: Sure, I'm reversing. The idea was to concentrate all related classes in the same file, but I agree that's a cosmetic change.

dvc/tree/gdrive.py
julsal: @shcheklein is the `gdrive://` scheme a DVC convention? I didn't find any reference in the API documentation or on the web.

shcheklein: yes, `gdrive://`, it is described here - https://dvc.org/doc/user-guide/setup-google-drive-remote#url-format.

dvc/tree/gdrive.py
```diff
@@ -510,20 +532,41 @@ def _path_to_item_ids(self, path, create, use_cache):
...
+        assert isinstance(path_info, GDriveURLInfo)
         try:
-            self._get_item_id(path_info)
+            item_id = self._get_item_id(path_info)
+            file_repr = self._drive.CreateFile({"id": item_id})
+            file_repr.FetchMetadata()
+
         except FileMissingError:
             return False
+        except ApiRequestError as e:
```
shcheklein, line 559: could you please clarify, when does it happen? (why `FileMissingError` is not enough). It feels that it would be better to encapsulate this into the function that raises the `FileMissingError`, wdyt?

dvc/tree/gdrive.py
```diff
@@ -510,20 +532,41 @@ def _path_to_item_ids(self, path, create, use_cache):
...
+            error_code = e.error.get("code", 0)
+            if error_code == 404:
+                return False
+            else:
+                raise DvcException(
```
shcheklein, line 564: I think we don't do this in other places right now. DVC automatically converts it upstream (probably with less details). It feels that it should be something more generic, wrapping ApiRequestError in every gdrive function probably is not very scalable. Can it be a decorator? Do we need one at all? How bad is it w/o this wrapper?

dvc/utils/__init__.py
```diff
@@ -349,45 +349,61 @@ def resolve_output(inp, out):
     return out
 
 
-def resolve_paths(repo, out):
+def resolve_paths(repo, out, force_out=False):
```
shcheklein, line 352: @julsal could you please clarify the essence of this change?

dvc/path_info.py
```diff
@@ -57,6 +57,16 @@ def __repr__(self):
     def __fspath__(self):
         return pathlib.PurePath.__str__(self)
 
+    def __eq__(self, other):
```
shcheklein, line 60: is it needed for this PR? (or just a warning fix?)

dvc/path_info.py
```diff
@@ -148,12 +163,12 @@ def _fill_parts(self, scheme, host, user, port, path):
         self.port = int(port) if port else self.DEFAULT_PORTS.get(self.scheme)
 
         if isinstance(path, _URLPathInfo):
-            self._spath = str(path)
+            self.spath = str(path)
             self._path = path
```
shcheklein, line 167: do we use `_spath` anywhere? is it effectively undefined in the `else`? what is the reason to rename it, btw (private to public)?

tests/unit/tree/test_gdrive.py
```diff
@@ -0,0 +1,44 @@
...
+from tests.basic_env import TestDvc
+from tests.utils import load_gdrive_credentials
+
+
+class TestGDriveURL(TestDvc):
+    def setUp(self):
+        super().setUp()
+        success = load_gdrive_credentials(self.dvc.root_dir)
+        if not success:
+            pytest.skip("no gdrive-user-credentials.json available")
```
shcheklein, line 15: We'll need to run this on CI. I think reading creds from the file is not a very regular approach for that. We already should have some tests (they might be located in the remotes). I think we read credentials from the env var. Could you please check? Does it make sense to use the same approach here?

tests/unit/tree/test_gdrive.py
```diff
@@ -0,0 +1,44 @@
...
+        ]
+
+        self.invalid_paths = [
+            "============",
+            "16onq6BZiiUFj083XloYVk7LDDpklDr7h/fake_dir",
+        ]
+
+        self.tree = GDriveTree(self.dvc, {})
+
+    def test_get_file_name(self):
```
shcheklein, line 31: How about we stub the API response? At least if you run locally? I know that we don't do that yet for other remotes, but what's your take on this?

dvc/stage/__init__.py
```diff
@@ -60,21 +60,60 @@ def loads_from(cls, repo, path, wdir, data):
...
+    :returns: a *cls* object representing the stage.
+    """
     from dvc.dvcfile import check_dvc_filename
 
     wdir = os.path.abspath(kwargs.get("wdir", None) or os.curdir)
-    path = os.path.abspath(path)
-    check_dvc_filename(path)
     check_stage_path(repo, wdir, is_wdir=kwargs.get("wdir"))
+
+    # path is None because we define its filename from the dependency.
```
shcheklein, line 87: @julsal please, correct me if I'm wrong here. My understanding is that we do this (the `get_file_name` approach) only to cover the case when we pass `gdrive://ID` to the `dvc import-url` since we don't know the file name in advance?

dvc/tree/gdrive.py
julsal: When the url has no path (`gdrive://ID`) it doesn't raise `FileMissingError`. I agree it's a good idea to encapsulate in the same function that raises the `FileMissingError`. However, the way it's written, the change will affect other function and may seem too complicated for the problem it is solving, so I tried something simpler. I'll propose an encapsulated solution now.

dvc/tree/gdrive.py
julsal: Encapsulating it to raise a `FileMissingError` might solve everything.

dvc/utils/__init__.py
julsal: The original function receives `repo` and `out`, returning `path` (the `.dvc` file path), `wdir` (the dir where the file should be saved, if `out` is not a path out of `repo`, and a new `out` (usually the `normpath(out)`). 

When `out` is `None`, the name should be inferred from the original, which also impacts `path`. As we don't know the file name at this point  (for the case of `gdrive://ID`), the function now doesn't deal with `path`, and wait to define it into `stage.create_stage`.

One important aspect is that, when `out` is a relative path containing a folder (i.e. `data/file.txt`), `data/` should be incorporated by `wdir` and `out` becomes only `file.txt`. If `out` points to a dir (i.e. `path/to/dir`), the path goes to `wdir` and `out` becomes `None`.

In  `repo.add._create_stages`, `resolve_paths` should always return `out != None`, even when `out` points to a directory. In its default behavior, `resolve_paths` appends the full path to `wdir`. `force_out` guarantee the new `out` keeps the leaf dir. So, if the input is `out=path/to/dir` and `force_out=True`, the new `out` is `dir`, and `path/to` goes to `wdir`.

tests/unit/tree/test_gdrive.py
julsal: Definitively, playing with the existing mechanism is a better approach. I'll address it.

tests/unit/tree/test_gdrive.py
julsal: I feel more confident when we test with the real system. What is the negative side of it?

shcheklein: Yep, I usually call them integration or e2e tests. And we definitely should have them. There is cost to this though- they are slow (you can write 100 of them) and not easy to setup locally. That's why my personal preference is to have a pyramid of tests- e.g 90% are regular unit tests that can be run locally or in CI and use mocks/stubs for the external APIs. 10% everything else- and developers can run them on CI.

tests/unit/tree/test_gdrive.py
shcheklein: Thanks 🙏

dvc/tree/gdrive.py
julsal: I decided to not use a decorator because it wouldn't allow me to include the file name in the `FileMissingError` message.

tests/unit/tree/test_gdrive.py
julsal: I've changed to rely on the env var. When I ran the tests with the env var configured, the code executed some tests cases that had being skipped. One of them is `test_data_cloud.test_cloud_cli`, which is now failing. I checked, and it fails when trying to upload an empty file (which raises an exception because Google Drive server returns a 400 error code). I'm wondering if this error is associated to the changes I didn't, or any kind of misconfiguration in either my Google Drive account or its credentials.

dvc/path_info.py
julsal: I pressed "resolve conversation" before "comment". He is the explanation: Yes, it's used in `BaseOutput.get_file_name`.

dvc/tree/gdrive.py
julsal: Done!

dvc/tree/gdrive.py
julsal: I didn't use an decorator to allow adding the file name in the `FileMissingError` message.

dvc/path_info.py
julsal: Sorry, again pressing "Resolve Conversation" without pressing "Comment". Yes, it's needed. I got some cases where the code failed because the relative path was not correctly compared.

dvc/stage/__init__.py
julsal: Yes, that's correct.

dvc/path_info.py
shcheklein: sorry, I had a loaded question (multiple points at once)

correct me if I'm wrong, your answer means that `spath` (no underscore `_`) is used outside, right? It makes sense then.

my first question was about `self._path = path` - do we still need `_spath`? And if we do, it looks like it might be undefined in certain cases? Is it expected?

julsal:
As I got one merge conflict with master, I rebased my branch. Now most of the builds are broken.

dvc/path_info.py
julsal: I didn't analyzed them deeply. I just made `spath` public. I didn't change the logics, so I don't know if it's expected `_path` being undefined in some cases. May I put a `TODO` comment here to investigate in another PR?

tests/unit/tree/test_gdrive.py
julsal: I've just run the tests in the `master` branch and I got the same error in `test_data_cloud.test_cloud_cli`, so I'm assuming it's a misconfiguration in my environment.

tests/unit/tree/test_gdrive.py
julsal: Sure, I'll address it.

dvc/path_info.py
shcheklein: Ah sorry, my bad, @julsal ! For some reason I mistakenly read `self._path = path` as `self._spath = path` and thought that we forgot to change it one place. No need to analyze this, it's not relevant at all. My mistake, sorry about the confusion.

dvc/tree/gdrive.py
```diff
@@ -562,7 +604,9 @@ def remove(self, path_info):
         self.gdrive_delete_file(item_id)
 
     def get_file_hash(self, path_info):
-        raise NotImplementedError
+        item_id = self._get_item_id(path_info)
+        file_repr = self._drive.CreateFile({"id": item_id})
+        return HashInfo(self.PARAM_CHECKSUM, file_repr["md5Checksum"])
```
shcheklein, line 609: btw, what happens if item_id represents a directory?

dvc/tree/gdrive.py
```diff
@@ -510,22 +541,33 @@ def _path_to_item_ids(self, path, create, use_cache):
...
+        item_ids = self._path_to_item_ids(
+            path_info.path, create, use_cache, path_info.bucket
+        )
         if item_ids:
-            return min(item_ids)
+            item_id = min(item_ids)
+            if item_id == path_info.bucket:
+                # if item_id doesn't exist, _fetch_metadata
+                # raises FileMissingError
+                self._fetch_metadata(item_id)
```
shcheklein, line 558: minor: fetch only specific fields?

shcheklein:
@julsal looks good overall (besides fixing tests and some minor stuff)! My take on this so far:

- Let's confirm with Ruslan (@efiop ) all the changes to the stage create and resolve path. It concerns me to be honest that we modified (increased complexity) them just for the sake of `gdrive://ID` . We can potentially decide that in that specific scenario we expect users to provide us a file name.
- Check what happens if we try to `import-url` a dir
- Check that `remote://` notation works fine
- Stubbing tests - up to you. I think it's fine to keep them as-is. I think it can be a separate task to improve them all later. 
- The whole logic of these functions - get_item_id, item_ids, exists - etc - feels like we'll need another iteration at some point.

tests/unit/tree/test_gdrive.py
julsal: As you suggested, I'll leave for a future separated task.

dvc/tree/gdrive.py
julsal: Solved!

dvc/tree/gdrive.py
julsal: Adding one filed might improve the performance. Solved!

julsal:
@shcheklein Does the `import-url` work for any source? The [documentation says](https://dvc.org/doc/command-reference/import-url) it doesn't work for HDFS, HTTP, and WebDav, but what about the others?

My tests break apparently because the local directory isn't created. Reading the code, I didn't find how the dir is created in the context of another source (S3, for instance). Is this a general problem, that needs to be addressed?

efiop:
> @shcheklein Does the import-url work for any source? The documentation says it doesn't work for HDFS, HTTP, and WebDav, but what about the others?

@julsal HDFS works though, looks like the docs are out-of-date. HTTP/WebDav don't, but all others do support it (e.g. s3). See https://github.com/iterative/dvc/blob/556143fd7e55bb66fd5e8477217ad218058e2333/tests/func/test_import_url.py#L178 with the full list.

> My tests break apparently because the local directory isn't created. Reading the code, I didn't find how the dir is created in the context of another source (S3, for instance). Is this a general problem, that needs to be addressed?

Could you elaborate on what breaks? Were you looking for dvc/tree/s3.py::S3Tree.makedirs() ? Or are you talking about test fixtures? Sorry, I'm a bit out of context.

julsal:
> > @shcheklein Does the import-url work for any source? The documentation says it doesn't work for HDFS, HTTP, and WebDav, but what about the others?
> 
> @julsal HDFS works though, looks like the docs are out-of-date. HTTP/WebDav don't, but all others do support it (e.g. s3). See
> 
> https://github.com/iterative/dvc/blob/556143fd7e55bb66fd5e8477217ad218058e2333/tests/func/test_import_url.py#L178
> 
> with the full list.
> > My tests break apparently because the local directory isn't created. Reading the code, I didn't find how the dir is created in the context of another source (S3, for instance). Is this a general problem, that needs to be addressed?
> 
> Could you elaborate on what breaks? Were you looking for dvc/tree/s3.py::S3Tree.makedirs() ? Or are you talking about test fixtures? Sorry, I'm a bit out of context.

Thanks, @efiop. When debugging `import-url` with a `gdrive` dir, it raises an exception because the local directory doesn't exist at the moment the files are downloaded. Nonetheless, I'll investigate the code for S3 to identify what's missing.

julsal:
@efiop , I managed to identify the bug in GDrive after comparing to the S3 implementation. Thanks! It's now solved.

The challenge now is the builds. In some builds, one test is failing because it's running despite calling `pytest.skip`. For other OS/versions it works. I've already tried different strategies to skip (annotation in different levels, function calls...), but they continue failing. Are you aware of a problem in this regard?

tests/unit/command/test_imp_url.py
```diff
@@ -44,3 +49,57 @@ def test_import_url_no_exec(mocker):
...
+@pytest.mark.parametrize(
+    "path, auth",
+    [
+        ("1nKf4XcsNCN3oLujqlFTJoK5Fvx9iKCZb", False),
+        ("1syA-26p7tehWyUiMPPk_s0hsFN0Nr_kX", True),
+        ("16onq6BZiiUFj083XloYVk7LDDpklDr7h/dir/data.txt", True),
+        ("16onq6BZiiUFj083XloYVk7LDDpklDr7h/dir", True),
+    ],
+)
+def test_import_url_gdrive(dvc, path, auth):
```
efiop, line 76: This doesn't seem like a unit test. Is there a reason why we can't just test it in `tests/func/test_import_url.py::test_import_url_dir` like we do with all other cloud types?

tests/unit/command/test_imp_url.py
```diff
@@ -44,3 +49,57 @@ def test_import_url_no_exec(mocker):
...
+
+
+def test_import_url_https(dvc):
+    cli_args = parse_args(
+        ["import-url", "https://data.dvc.org/get-started/data.xml"]
+    )
+    assert cli_args.func == CmdImportUrl
+
+    cmd = cli_args.func(cli_args)
+    assert cmd.run() == 0
```
efiop, line 61: Not clear how this is related to this change. Also, this is not really a unit test.

dvc/path_info.py
```diff
@@ -57,6 +57,16 @@ def __repr__(self):
...
 
+    def __eq__(self, other):
+        if isinstance(other, str):
+            return self.fspath == os.path.abspath(other)
+        else:
+            return super().__eq__(other)
+
+    # pylint: disable=W0235
+    def __hash__(self):
+        return super().__hash__()
```
efiop, line 68: Not needed?

dvc/stage/__init__.py
```diff
@@ -60,21 +60,60 @@ def loads_from(cls, repo, path, wdir, data):
...
+    fill_stage_dependencies(
+        stage, **project(kwargs, ["deps", "erepo", "params"])
+    )
+
+    param_outs = kwargs.get("outs")
+
+    if param_outs:
+        outs = []
+        for i, o in enumerate(param_outs):
+            outs.append(o or stage.deps[i].get_file_name())
```
efiop, line 99: This will require establishing a connection, which will break --no-exec case.

Seems like we are overcomplicating this. If the name from the url is ugly, we could either leave it ugly or user could use `-o` to specify the name he wants. What do you think?

tests/unit/command/test_imp_url.py
julsal: If I understood correctly, there is no mock service for gdrive. So, I wrote the test relying on the actual service.

tests/unit/command/test_imp_url.py
julsal: I added the test to understand the behavior of the code for this case. I decided to keep it, so another functionality could keep being tested.

dvc/path_info.py
julsal: When we change `__eq__` , we also need to provide an implementation for `__hash__`.

tests/unit/command/test_imp_url.py
efiop: But we don't need this. We have https://github.com/iterative/dvc/blob/2f620e9226f6a337cd04da9748b4f40cd19efd1f/tests/func/test_import_url.py#L131 already, which, unlike this test, doesn't rely on external resources at all. Plus it is absolutely unrelated to this PR, making it harder to review. Let's remove it.

dvc/stage/__init__.py
julsal: It's not a matter of being ugly. Url in the format `gdrive://ID` doesn't provide the name of the file/directory at all. As the unique way to get it is establishing a connection, I don't consider we are overcomplicating. However, if the connection should be avoided, the code can either make `-o` mandatory for this case or use the ID as the file/directory name.

tests/unit/command/test_imp_url.py
julsal: Sure. Done!

tests/unit/command/test_imp_url.py
efiop: There is no mock, correct. There is a `gdrive` fixture though, with real gdrive (e.g. gs also relies on real google-cloud-storage). E.g. see tests/remotes/http.py , where it implements the minimal required methods (mkdir, write_text, write_bytes) for the fixture to be able to use it in `test_import_url`. `gdrive` fixture will handle skipping for you too automatically.

dvc/stage/__init__.py
efiop: Sure, I understand that.

> However, if the connection should be avoided, the code can either make -o mandatory for this case or use the ID as the file/directory name.

Let's not make the behavior dependant on whether or not we have a connection, that would be against the regular dvc behavior and will be a hell of a thing to maintain and explain to people. Let's just keep it simple: unless -o is provided, parse the name from the URL, no matter if it is gdrive://ID or gdrive://ID/some/path.

julsal: No problem. I'll change it and revert other changes, including those in `stage.create_stage` and `utils.resolve_paths`.

Regarding the build / tests, could you check the message above I'm reporting the errors I'm getting?

tests/unit/command/test_imp_url.py
julsal: I got it. I'll adapt this test to follow the format of those in `tests/func/test_import_url.py::test_import_url_dir`.

dvc/stage/__init__.py
efiop: > Regarding the build / tests, could you check the message above I'm reporting the errors I'm getting?

@julsal The message you've posted doesn't say anything about the tests that are failing, it only says that they are indeed failing. There should be a verbose log above it for each test that has failed.

tests/remotes/gdrive.py
```diff
@@ -1,41 +1,104 @@
...
+
+    def write_bytes(self, contents):
+        f = tempfile.NamedTemporaryFile()
+        f.write(contents)
+        f.flush()
+
+        local = PathInfo(pathlib.Path(f.name))
+        self.tree.upload(local, self._get_gdrive_path_info)
+
+        f.close()
```
efiop, line 83: just use `with`?

tests/remotes/gdrive.py
```diff
@@ -1,41 +1,104 @@
...
     def get_url():
         # NOTE: `get_url` should always return new random url
         return "gdrive://" + GDrive._get_storagepath()
 
+    def is_file(self):
+        return self.tree.isfile(self._get_gdrive_path_info)
+
+    def is_dir(self):
+        return self.tree.isdir(self._get_gdrive_path_info)
+
```
efiop, line 66: These and read_text/bytes are not used, unless I'm missing something.

tests/remotes/gdrive.py
```diff
@@ -45,7 +108,7 @@ def gdrive(make_tmp_dir):
     # NOTE: temporary workaround
     tmp_dir = make_tmp_dir("gdrive", dvc=True)
 
-    ret = GDrive(GDrive.get_url())
-    tree = GDriveTree(tmp_dir.dvc, ret.config)
+    ret = GDrive(GDrive.get_url(), tmp_dir.dvc)
+    tree = ret.tree
```
efiop, line 112: Shame gdrive is so complex that we have to reulse GdriveTree in our test fixtures :slightly_frowning_face: But oh well...

dvc/output/gdrive.py
```diff
@@ -0,0 +1,7 @@
+from dvc.output.base import BaseOutput
+
+from ..tree.gdrive import GDriveTree
+
+
+class GDriveOutput(BaseOutput):
+    TREE_CLS = GDriveTree
```
efiop, line 7: GDriveOutput is never used. We only need the Dependency for import-url.

dvc/tree/gdrive.py
```diff
@@ -527,42 +543,47 @@ def exists(self, path_info, use_dvcignore=True):
...
+        self._fetch_metadata(file_repr, fields="mimeType")
+        return file_repr["mimeType"] == FOLDER_MIME_TYPE
 
-        for item in self._gdrive_list(query):
-            parent_id = item["parents"][0]["id"]
-            yield posixpath.join(
-                self._ids_cache["ids"][parent_id], item["title"]
-            )
+    def isfile(self, path_info):
+        return self.exists(path_info) and not self.isdir(path_info)
```
efiop, line 553: Can we do it in one API call? For GDrive it is quite important, as there are pretty strict quotas.

dvc/path_info.py
```diff
@@ -120,11 +120,17 @@ def __repr__(self):
...
         assert not p.query and not p.params and not p.fragment
         assert p.password is None
-        self._fill_parts(p.scheme, p.hostname, p.username, p.port, p.path)
+
+        if keep_host_case:
+            host = p.netloc.split(":")[0]
+        else:
+            host = p.hostname
+
+        self._fill_parts(p.scheme, host, p.username, p.port, p.path)
```
efiop, line 134: Do I understand that this is not needed, as we've been restoring it GDriveURLInfo before?

julsal: Before adding here, I had removed from `GDriveURLInfo`. I did this change because the previous version wasn't working correctly during a debugging scenario. Yes, I know, it doesn't make sense, but it's the truth.

dvc/tree/gdrive.py
julsal: Done!

dvc/output/gdrive.py
julsal: Done!

tests/remotes/gdrive.py
julsal: Done!

tests/remotes/gdrive.py
julsal: Done!

julsal:
all review comments now are solved

dvc/dependency/gdrive.py
```diff
@@ -0,0 +1,16 @@
...
+
+class GDriveDependency(BaseDependency, BaseOutput):
+    TREE_CLS = GDriveTree
+
+    def __init__(self, stage, path, info=None, **wkwargs):
+        repo = stage.repo if stage else None
+        if "tree" not in wkwargs:
+            base_path = GDriveURLInfo(path).replace(path="").url
+            tree = GDriveTree(repo, {"url": base_path})
+            wkwargs["tree"] = tree
```
efiop, line 15: Could you clarify why we need this for gdrive but not for any other dep?

julsal: In `GDriveTree`, most of the methods rely on `_get_item_id`, which depends on `self._bucket` to work properly. `BaseOutput` (the class from which `GDriveDependency` inherits) has a `tree` attribute, which, by default, is instantiated with an empty configuration:

https://github.com/iterative/dvc/blob/689a569246634ed1f8367b7a0359e334964f1252/dvc/output/base.py#L124

However, to have `self._bucket` correctly configured, we need to pass `url` in the configuration. I presume the other `Dependency` classes work as expected with a "non-configured" `tree`.

dvc/path_info.py
efiop: @julsal So is this actually needed for this PR? If not, let's not do it.

tests/remotes/gdrive.py
```diff
@@ -1,41 +1,85 @@
...
+            return {
+                "url": self.url,
+            }
+        else:
+            return {
+                "url": self.url,
+                "gdrive_service_account_email": "test",
+                "gdrive_service_account_p12_file_path": "test.p12",
+                "gdrive_use_service_account": True,
+            }
```
efiop, line 41: This looks confusing, could you clarify what is going on?

Also no need for `else` if you return before it.

tests/remotes/gdrive.py
```diff
@@ -1,41 +1,85 @@
...
 
     @staticmethod
     def _get_storagepath():
         return TEST_GDRIVE_REPO_BUCKET + "/" + str(uuid.uuid4())
 
+    @cached_property
+    def _get_gdrive_path_info(self):
+        return GDriveURLInfo(self.url)
+
```
efiop, line 50: Not sure I understand the point of this, could you clarify, please? Why can't we just use `self` instead?

tests/remotes/gdrive.py
```diff
@@ -1,41 +1,85 @@
...
 
 TEST_GDRIVE_REPO_BUCKET = "root"
 
 
 class GDrive(Base, CloudURLInfo):
+    def __init__(self, url, dvc):
+        CloudURLInfo.__init__(self, url)
+        Base.__init__(self, url)
+        self.tree = GDriveTree(dvc, self.config)
+
```
efiop, line 24: Maybe just
```suggestion

    @cached_property
    def tree(self):
        return GDriveTree(dvc, self.config)

```
? Same as in other fixtures.

tests/remotes/gdrive.py
julsal: because `self` is not an instance of `GDriveURLInfo`. So, it breaks some `assert` conditions.

efiop: Ah, good point! I didn't realize that we didn't use it before here, it is by an accident. Does it break if we switch?

tests/remotes/gdrive.py
julsal: I need to pass `dvc` to the class, as it comes from `tmp_dir.dvc`. And `tree` is public.

https://github.com/iterative/dvc/blob/689a569246634ed1f8367b7a0359e334964f1252/tests/remotes/gdrive.py#L90-L93

dvc/path_info.py
julsal: Done!

tests/remotes/gdrive.py
julsal: My fault. Done!

dvc/dependency/gdrive.py
efiop: Correct, other trees handle that as a special case: https://github.com/iterative/dvc/blob/96ead9425af0c3e3503ebb335fd3edacd2c5f6e1/dvc/tree/http.py#L43 Could we do the same for gdrive?

tests/remotes/gdrive.py
julsal: Thanks for questioning again. In fact, the `assert` is not there anymore after I reverted some old changes. `self` works nicely.

julsal:
@efiop builds are breaking in the `set up Python` phase likely because of this https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

Everything else is solved.

dvc/dependency/gdrive.py
julsal: If I do the same for `GDriveTree`, I'll need to change the solution of `_get_item_id` and other methods to not rely on `self._bucket`, which will demand a change at least in `_path_to_item_ids`, `_get_cached_item_ids`, `_path_to_item_ids` and `_list_params`.

julsal: To make my comment more clear: `GDriveTree` needs `url` (and the `_bucket` from it) to work properly. If we allow a `url=None`, we need to change at least those methods I mentioned to get the `bucket` from the `info_path` passed by the param of the public methods. Such a change may also affect the cache mechanism that relies on `_bucket`.

efiop:
@julsal set-env issue was fixed in master. Please rebase.

dvc/dependency/gdrive.py
efiop: I see. The problem with using `self._bucket` for things like import is that you are mixing up the concepts of tree and the path that belongs to that tree. `self.path_info`(along with self._bucket) should only be used explicitly, like in s3 and other trees. If we are operating with a third-party `path_info`, we should use it or otherwise there might be a difference in tree and path configuration that we won't notice.

julsal: I've written new code without relying on `self._bucket`.

I've noticed, however,  in the current tests,`GDriveDependency` is always getting a `tree` based on the remote tree, because the tests are based on `remote://` urls.

https://github.com/iterative/dvc/blob/7501302baa6f724a2e962446249044a563b44568/dvc/dependency/__init__.py#L65-L67

So, the case where `self._bucket` is `None` was never tested. Now I added a test to import a `gdrive://` url.

The branch was also rebased.

tests/func/test_import_url.py
```diff
@@ -206,4 +209,45 @@ def test_import_url_dir(tmp_dir, dvc, workspace, stage_md5, dir_md5):
...
+def assert_local_dir(dvc, tmp_dir, dir_name):
+    assert set(os.listdir(tmp_dir / dir_name)) == {"file", "subdir"}
+    assert (tmp_dir / dir_name / "file").read_text() == "file"
+    assert list(os.listdir(tmp_dir / dir_name / "subdir")) == ["subfile"]
+    assert (tmp_dir / dir_name / "subdir" / "subfile").read_text() == "subfile"
+
     assert dvc.status() == {}
+
+
+def test_import_gdrive_url_dir(gdrive, tmp_dir, dvc):
```
efiop, line 222: Could you clarify why we need this test when we already have gdrive in `test_import_url_dir` above?

julsal: I explained in the other conversion. I'll copy it here:

I've noticed, in the current tests, `GDriveDependency` is always getting a tree based on the remote tree, because the tests use `remote://` urls.

https://github.com/iterative/dvc/blob/7501302baa6f724a2e962446249044a563b44568/dvc/dependency/__init__.py#L65-L67

So, the case where `self._bucket` is `None` was never tested. This test using a `gdrive://` url solves this gap.

tests/remotes/gdrive.py
efiop: @julsal Would it make sense to turn it around? The mkdir below is a hack, we didn't need tree before anywhere so we've initialized it there. Now we need the tree here, first and foremost, so it makes sense utilize the tree in the hack and not vice versa.

tests/func/test_import_url.py
efiop: @julsal After your recent change `self._bucket` shouldn't be used anywhere. If some path_info` is supplied, we should always use `path_info.bucket` and not use `self`. Which means that we can remove `self._bucket` shortcut and just not have to test it anymore.

julsal: `self._bucket` was removed. The method `GDriveTree.gdrive_delete_file` receives `item_id` instead of `path_info`. Should I rely on `self.path_info.bucket` as I did, or should I change its signature to get `path_info` (public method)?

Should I remove the `test_import_gdrive_url_dir`? Removing `self._bucket` doesn't change the fact that `GDriveDependency` is always getting a tree based on the remote tree when we test with `remote://` urls. Additionally, it tests the particularities of `gdrive://ID`, as:

https://github.com/iterative/dvc/blob/e90ae72d41413ebc0ea8e96465d67073be113f42/dvc/tree/gdrive.py#L86-L87

if self._spath is empty, `is_relative_to` get lost:

https://github.com/iterative/dvc/blob/6324bd8a69ddb7ca4bd4d32f6e2ae2b368ce266d/dvc/tree/base.py#L330

This is a case I spotted because of the new test. The current test works because the url has `path`.

efiop:
Moving to https://github.com/iterative/dvc/pull/4932 , as upstream branches have access to secret gdrive creds.

efiop:
Ah, wait. I think when you push a branch with the same name to the upstream, it makes the secrets available for the original one. Let's see...

efiop:
Nope, unfortunately it doesn't :( So moving to https://github.com/iterative/dvc/pull/4932 after all.

PR #4864, opened 2020-11-09
api: add support for simple wildcards

### LINKED ISSUE #4816: Filename patters in import and pull
Sometimes only a subset of files is needed when the user runs `import` or `pull` data from a data directory. It is convenient to define a file pattern for an import.

From https://discuss.dvc.org/t/working-with-a-small-subset-of-remote-data/541
Related issue: https://github.com/iterative/dvc/issues/4815

Patterns to implement:
- [ ] simple wildcard `dvc pull cats-dogs/data/train/dogs/*.img`
- [ ] whole wildcard `dvc pull cats-dogs/data/train/{dogs,cats}/???.img`
- [ ] globstar/ricursive `dvc pull cats-dogs/data/train/**/*.img`
- [ ] iterator `dvc pull cats-dogs/data/train/dogs/%C.img?counter=1:100`
- [ ] date `dvc pull users/%Y/%m/%d/users.csv?startdata=2020-09-01,enddate=now,ignoremissing`

The first three patterns should use a regular Unix file syntax. While the last two require a special language to define the pattern - we need to find a good examples.

ju0gri:
Related to #4816.

Signed-off-by: Ioana Grigoropol <ioana.grigoropol@gmail.com>

https://github.com/iterative/dvc.org/issues/1928

### DIFF
--- a/dvc/command/add.py
+++ b/dvc/command/add.py
@@ -20,6 +20,7 @@ class CmdAdd(CmdBase):
                 no_commit=self.args.no_commit,
                 fname=self.args.file,
                 external=self.args.external,
+                glob=self.args.glob,
             )
 
         except DvcException:
@@ -56,6 +57,12 @@ def add_parser(subparsers, parent_parser):
         action="store_true",
         default=False,
         help="Allow targets that are outside of the DVC repository.",
+    )
+    parser.add_argument(
+        "--glob",
+        action="store_true",
+        default=False,
+        help="Allows targets containing shell-style wildcards.",
     )
     parser.add_argument(
         "--file",
--- a/dvc/repo/add.py
+++ b/dvc/repo/add.py
@@ -1,4 +1,3 @@
-import glob
 import logging
 import os
 
@@ -24,7 +23,13 @@ logger = logging.getLogger(__name__)
 @locked
 @scm_context
 def add(
-    repo, targets, recursive=False, no_commit=False, fname=None, external=False
+    repo,
+    targets,
+    recursive=False,
+    no_commit=False,
+    fname=None,
+    external=False,
+    glob=False,
 ):
     if recursive and fname:
         raise RecursiveAddingWhileUsingFilename()
@@ -58,7 +63,12 @@ def add(
                 )
 
             stages = _create_stages(
-                repo, sub_targets, fname, pbar=pbar, external=external
+                repo,
+                sub_targets,
+                fname,
+                pbar=pbar,
+                external=external,
+                enable_glob=glob,
             )
 
             try:
@@ -150,16 +160,21 @@ def _find_all_targets(repo, target, recursive):
     return [target]
 
 
-def _create_stages(repo, targets, fname, pbar=None, external=False):
+def _create_stages(
+    repo, targets, fname, pbar=None, external=False, enable_glob=False
+):
+    import glob
+
     from dvc.stage import Stage, create_stage
 
-    expanded_targets = []
-
-    for target in targets:
-        found_target = glob.glob(target, recursive=True)
-        if not found_target:
-            found_target = [target]
-        expanded_targets.extend(found_target)
+    if enable_glob:
+        expanded_targets = [
+            exp_target
+            for target in targets
+            for exp_target in glob.iglob(target, recursive=True)
+        ]
+    else:
+        expanded_targets = targets
 
     stages = []
     for out in Tqdm(
--- a/dvc/repo/add.py
+++ b/dvc/repo/add.py
@@ -1,3 +1,4 @@
+import glob
 import logging
 import os
 
@@ -152,12 +153,19 @@ def _find_all_targets(repo, target, recursive):
 def _create_stages(repo, targets, fname, pbar=None, external=False):
     from dvc.stage import Stage, create_stage
 
-    stages = []
+    expanded_targets = []
 
+    for target in targets:
+        found_target = glob.glob(target, recursive=True)
+        if not found_target:
+            found_target = [target]
+        expanded_targets.extend(found_target)
+
+    stages = []
     for out in Tqdm(
-        targets,
+        expanded_targets,
         desc="Creating DVC-files",
-        disable=len(targets) < LARGE_DIR_SIZE,
+        disable=len(expanded_targets) < LARGE_DIR_SIZE,
         unit="file",
     ):
         path, wdir, out = resolve_paths(repo, out)
--- a/dvc/repo/add.py
+++ b/dvc/repo/add.py
@@ -68,7 +68,7 @@ def add(
                 fname,
                 pbar=pbar,
                 external=external,
-                enable_glob=glob,
+                glob=glob,
             )
 
             try:
--- a/dvc/repo/add.py
+++ b/dvc/repo/add.py
@@ -163,7 +163,7 @@ def _find_all_targets(repo, target, recursive):
 def _create_stages(
     repo, targets, fname, pbar=None, external=False, glob=False
 ):
-    import glob
+    from glob import iglob
 
     from dvc.stage import Stage, create_stage
 
--- a/dvc/repo/add.py
+++ b/dvc/repo/add.py
@@ -167,7 +167,7 @@ def _create_stages(
 
     from dvc.stage import Stage, create_stage
 
-    if enable_glob:
+    if glob:
         expanded_targets = [
             exp_target
             for target in targets
--- a/dvc/repo/add.py
+++ b/dvc/repo/add.py
@@ -171,7 +171,7 @@ def _create_stages(
         expanded_targets = [
             exp_target
             for target in targets
-            for exp_target in glob.iglob(target, recursive=True)
+            for exp_target in iglob(target, recursive=True)
         ]
     else:
         expanded_targets = targets
--- a/dvc/repo/add.py
+++ b/dvc/repo/add.py
@@ -161,7 +161,7 @@ def _find_all_targets(repo, target, recursive):
 
 
 def _create_stages(
-    repo, targets, fname, pbar=None, external=False, enable_glob=False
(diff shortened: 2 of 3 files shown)


dvc/repo/add.py
```diff
@@ -160,27 +160,33 @@ def _create_stages(repo, targets, fname, pbar=None, external=False):
...
-            fname or path,
-            wdir=wdir,
-            outs=[out],
-            external=external,
-        )
-        if stage:
-            Dvcfile(repo, stage.path).remove()
+        import glob
+
+        outs = glob.glob(out)
```
efiop, line 165: Expanded out means more targets, so looks like we should do that before starting the `targets` loop/progress-bar above.

dvc/repo/add.py
```diff
@@ -152,21 +152,29 @@ def _find_all_targets(repo, target, recursive):
...
     for out in Tqdm(
-        targets,
+        expanded_targets,
         desc="Creating DVC-files",
-        disable=len(targets) < LARGE_DIR_SIZE,
+        disable=len(expanded_targets) < LARGE_DIR_SIZE,
         unit="file",
     ):
-        path, wdir, out = resolve_paths(repo, out)
+        path, wdir, resolved_out = resolve_paths(repo, out)
```
efiop, line 171: Let's not rename it since it is not needed and unrelated.

dvc/repo/add.py
```diff
@@ -152,21 +152,29 @@ def _find_all_targets(repo, target, recursive):
 def _create_stages(repo, targets, fname, pbar=None, external=False):
     from dvc.stage import Stage, create_stage
 
-    stages = []
+    expanded_targets = []
+    import glob
```
efiop, line 156: Let's move the import up(at least in this helper) so it is cleaner.

efiop:
> Fixes #4816.

It doesn't really fix it, just use `Related` please, so that github doesn't close it automatically :slightly_smiling_face:

tests/func/test_add.py
```diff
@@ -190,6 +190,38 @@ def test_add_file_in_dir(tmp_dir, dvc):
...
+    subdir_path = os.path.join("dir", "subdir", "subdata*")
+
+    stages = dvc.add(subdir_path)
+
+    expected_def_paths = ["subdata", "subdata123"]
+    expected_rel_paths = [
+        os.path.join("dir", "subdir", "subdata") + ".dvc",
+        os.path.join("dir", "subdir", "subdata123") + ".dvc",
+    ]
+    for stage in stages:
```
efiop, line 214: Would test the length too.

tests/func/test_add.py
```diff
@@ -190,6 +190,38 @@ def test_add_file_in_dir(tmp_dir, dvc):
...
+            "dir": {
+                "subdir": {
+                    "subdata": "subdata content",
+                    "esubdata": "extra subdata content",
+                    "subdata123": "subdata content 123",
+                }
+            }
+        }
+    )
+    subdir_path = os.path.join("dir", "subdir", "subdata*")
```
efiop, line 205: Should we test more features or just single wildcards?

ju0gri: I'll add some more use cases for shell-style wildcards, and possible some negative ones for recursive matching which will not work as it is now, unless we enable the recursive flag. do we want that ?

efiop: @ju0gri Note that the more the tests - the more expensive it get's, as we run them in real filesystem. I suppose os.glob has some spec that it sticks to, so it probably doesn't make sense to test it all.

Recursive flag is for double wildcards only, right? Definitely seems useful. Any reason why it is disabled by default? Maybe some perf issues (except for the obvious misuse).

ju0gri: Yes, the recursive flag enables `**`. According to their docs: 
> Using the “**” pattern in large directory trees may consume an inordinate amount of time.

ju0gri: Would adding one tests for each use case defined in `fnmatch` manual be an acceptable compromise ?
```
* | matches everything
? | matches any single character
[seq] | matches any character in seq
[!seq] | matches any character not in seq
** | searches recursively
```

efiop: > Yes, the recursive flag enables **. According to their docs:

Sure, I am wondering whether it has some hidden cons. E.g. maybe it is considerably slower by itself when that pattern is supported. I guess need to give it a shot/look into internals for that.

efiop: > Would adding one test for each use case defined in fnmatch manual be an acceptable compromise ?

Sounds good.

Btw, pathspec might not be needed in some cases. E.g. `dvc add '?*[seq]**'`, where all the special stuff is already shieled by the single quotes. Current patch enabled globbing by default, but it looks like we should probably make it optional.

ju0gri: Yes, makes sense. I am bad with naming, but added an argument called `pathpattern` for the `add` command to enable globing of the passed in target. Also, as I am not very familiar with the structure of the repo, are there other places where I need to add coverage for the new argument ?

dvc/command/add.py
```diff
@@ -57,6 +58,12 @@ def add_parser(subparsers, parent_parser):
         default=False,
         help="Allow targets that are outside of the DVC repository.",
     )
+    parser.add_argument(
+        "--pathpattern",
```
efiop, line 62: How about `--glob`?

dvc/repo/add.py
```diff
@@ -57,7 +63,12 @@ def add(
                 )
 
             stages = _create_stages(
-                repo, sub_targets, fname, pbar=pbar, external=external
+                repo,
+                sub_targets,
+                fname,
+                pbar=pbar,
+                external=external,
+                enable_glob=glob,
```
efiop, line 71:
```suggestion
                glob=glob,
```

dvc/repo/add.py
```diff
@@ -149,15 +160,27 @@ def _find_all_targets(repo, target, recursive):
     return [target]
 
 
-def _create_stages(repo, targets, fname, pbar=None, external=False):
+def _create_stages(
+    repo, targets, fname, pbar=None, external=False, enable_glob=False
```
efiop, line 164:
```suggestion
    repo, targets, fname, pbar=None, external=False, glob=False
```

dvc/repo/add.py
```diff
@@ -149,15 +160,27 @@ def _find_all_targets(repo, target, recursive):
     return [target]
 
 
-def _create_stages(repo, targets, fname, pbar=None, external=False):
+def _create_stages(
+    repo, targets, fname, pbar=None, external=False, glob=False
+):
+    import glob
```
efiop, line 166:
```suggestion
    from glob import iglob
```

dvc/repo/add.py
```diff
@@ -149,15 +160,27 @@ def _find_all_targets(repo, target, recursive):
...
-def _create_stages(repo, targets, fname, pbar=None, external=False):
+def _create_stages(
+    repo, targets, fname, pbar=None, external=False, glob=False
+):
+    from glob import iglob
+
     from dvc.stage import Stage, create_stage
 
-    stages = []
+    if enable_glob:
```
efiop, line 170:
```suggestion
    if glob:
```

dvc/repo/add.py
```diff
@@ -149,15 +160,27 @@ def _find_all_targets(repo, target, recursive):
...
+    from glob import iglob
+
     from dvc.stage import Stage, create_stage
 
-    stages = []
+    if glob:
+        expanded_targets = [
+            exp_target
+            for target in targets
+            for exp_target in glob.iglob(target, recursive=True)
```
efiop, line 174:
```suggestion
            for exp_target in iglob(target, recursive=True)
```

efiop (approved):
Looks good! The only thing left now is to create a ticket for docs or create a doc PR, as noted in the PR checkbox.

efiop:
Thank you!

dvc/command/add.py
```diff
@@ -57,6 +58,12 @@ def add_parser(subparsers, parent_parser):
         default=False,
         help="Allow targets that are outside of the DVC repository.",
     )
+    parser.add_argument(
+        "--glob",
+        action="store_true",
+        default=False,
+        help="Allows targets containing shell-style wildcards.",
+    )
```
jorgeorpinel, line 66: Hi! A few questions (related to documenting this per iterative/dvc.org/issues/1928):

- What does "glob" abbreviate or stand for?
- Why is a flag needed to allow/enable wildcards?
- Is there a short description of what "shell-style" means exactly? Or a link to some Linux doc, Python package, or something specifying what can be done in terms of wildcards?

karajan1001: I think it's more natural to use wildcards in the default mode without any flag, just like what we do in Unix. And it‘s better to extract the `glob` to some common function that could be used in all commands.

efiop: @karajan1001 I personally wasn't ready to enable this by default right away as I'm not sure it won't affect negatively some existing usecases :slightly_frowning_face:  Once we're ready for it, it will be as simple as turning that flag on by default. For PS users `--glob` will be a temporary inconvenience, but at least the functionality is being slowly filled up.

@jorgeorpinel Maybe just forget about the doc ticket for it for now, we'll be adding more stuff for it in the future.

> What does "glob" abbreviate or stand for?

https://docs.python.org/3/library/glob.html 

> Why is a flag needed to allow/enable wildcards?

Usually, your shell handles the expansion and it will still do that. Our --glob is useful for programatic use or in shells that don't support explansion natively (e.g. powershell). I'm personally not ready to enable by default right now.

> Is there a short description of what "shell-style" means exactly? Or a link to some Linux doc, Python package, or something specifying what can be done in terms of wildcards?

https://docs.python.org/3/library/glob.html

jorgeorpinel: Cool, just wanted to understand this. The initial docs PR is ready for merge.

PR #4867, opened 2020-11-10
dvc: refactor cache/remote relations

efiop:
The final separation that puts this Cache/Remote/Tree trio in their proper places.

No new logic added, just tossing around methods to make them lend in logical places (except for minor gc message adjustments and new internal helpers). E.g. index is for Remote, but listing the cache is for Cache.


dvc/cache/base.py
```diff
@@ -646,3 +635,292 @@ def set_dir_info(self, hash_info):
 
         hash_info.dir_info = self.get_dir_cache(hash_info)
         hash_info.nfiles = hash_info.dir_info.nfiles
+
+    def list_paths(self, prefix=None, progress_callback=None):
```
efiop, line 639:
```suggestion
    def _list_paths(self, prefix=None, progress_callback=None):
```

dvc/cache/base.py
```diff
@@ -646,3 +635,292 @@ def set_dir_info(self, hash_info):
...
+            path_info = self.tree.path_info
+            prefix = False
+        if progress_callback:
+            for file_info in self.tree.walk_files(path_info, prefix=prefix):
+                progress_callback()
+                yield file_info.path
+        else:
+            yield from self.tree.walk_files(path_info, prefix=prefix)
+
+    def list_hashes(self, prefix=None, progress_callback=None):
```
efiop, line 656:
```suggestion
    def _list_hashes(self, prefix=None, progress_callback=None):
```

dvc/cache/base.py
```diff
@@ -646,3 +635,292 @@ def set_dir_info(self, hash_info):
...
+    def _max_estimation_size(self, hashes):
+        # Max remote size allowed for us to use traverse method
+        return max(
+            self.tree.TRAVERSE_THRESHOLD_SIZE,
+            len(hashes)
+            / self.tree.TRAVERSE_WEIGHT_MULTIPLIER
+            * self.tree.LIST_OBJECT_PAGE_SIZE,
+        )
+
+    def estimate_remote_size(self, hashes=None, name=None):
```
efiop, line 691:
```suggestion
    def _estimate_remote_size(self, hashes=None, name=None):
```

dvc/cache/base.py
```diff
@@ -646,3 +635,292 @@ def set_dir_info(self, hash_info):
...
+
+            remote_hashes = set(hashes)
+            if remote_hashes:
+                remote_size = total_prefixes * len(remote_hashes)
+            else:
+                remote_size = total_prefixes
+            logger.debug(f"Estimated remote size: {remote_size} files")
+        return remote_size, remote_hashes
+
+    def list_hashes_traverse(
```
efiop, line 726:
```suggestion
    def _list_hashes_traverse(
```

dvc/cache/base.py
```diff
@@ -646,3 +635,292 @@ def set_dir_info(self, hash_info):
...
+            if self.tree.is_dir_hash(hash_):
+                # backward compatibility
+                # pylint: disable=protected-access
+                self._remove_unpacked_dir(hash_)
+            self.tree.remove(path_info)
+            removed = True
+
+        return removed
+
+    def list_hashes_exists(self, hashes, jobs=None, name=None):
```
efiop, line 828:
```suggestion
    def _list_hashes_exists(self, hashes, jobs=None, name=None):
```

efiop: Finally all of these ugly methods are where they belong and can be made private.

dvc/cache/base.py
```diff
@@ -646,3 +635,292 @@ def set_dir_info(self, hash_info):
...
+        cache can be estimated, since the cache is evenly distributed according
+        to hash.
+
+        Returns:
+            A list with hashes that were found in the remote
+        """
+        # Remotes which do not use traverse prefix should override
+        # hashes_exist() (see ssh, local)
+        assert self.tree.TRAVERSE_PREFIX_LEN >= 2
+
```
efiop, line 891: Previosly we checked the index here, but that was mixing up remotes and caches, which is incorrect, so now we first check the index in Remote(because our current indexes are local indexes for _remote_ caches) and then fallback to checking cache, which is naturual and provides nice intuitive isolation.

dvc/path_info.py
```diff
@@ -121,11 +121,15 @@ class URLInfo(_BasePath):
...
+        if isinstance(url, URLInfo):
+            parts = (url.scheme, url.host, url.user, url.port, url.path)
+        else:
+            p = urlparse(url)
+            assert not p.query and not p.params and not p.fragment
+            assert p.password is None
+            parts = (p.scheme, p.hostname, p.username, p.port, p.path)
 
-        self._fill_parts(p.scheme, p.hostname, p.username, p.port, p.path)
+        self._fill_parts(*parts)
```
efiop, line 132: Oh, actually new logic here. Added this to make `URLInfo(URLInfo())` work the same way as `PathInfo(PathInfo())` and pathlib's `Path(Path())` work. Let me check if we still need it and if we do -- will add a unit test.

dvc/remote/__init__.py
```diff
@@ -1,13 +1,10 @@
...
 from .local import LocalRemote
-from .ssh import SSHRemote
 
 
 def get_remote(repo, **kwargs):
     tree = get_cloud_tree(repo, **kwargs)
     if tree.scheme == "local":
         return LocalRemote(tree)
-    if tree.scheme == "ssh":
-        return SSHRemote(tree)
```
efiop, line 12: We never needed an SSHRemote, as all of the custom stuff was actually about SSHCache.

dvc/remote/base.py
```diff
@@ -47,123 +90,393 @@ def __repr__(self):
...
-            A list with hashes that were found in the remote
-        """
-        # Remotes which do not use traverse prefix should override
-        # hashes_exist() (see ssh, local)
-        assert self.tree.TRAVERSE_PREFIX_LEN >= 2
+    def gc(self, *args, **kwargs):
+        removed = self.cache.gc(*args, **kwargs)
+
+        if removed:
+            with self.index:
```
efiop, line 97: Even though it is not quite the same, better to use `index_locked` decorator for symmetry.

dvc/remote/base.py
```diff
@@ -47,123 +90,393 @@ def __repr__(self):
...
+            named_cache,
+            jobs=jobs,
+            show_checksums=show_checksums,
+            download=download,
+            log_missing=log_missing,
+        )
+        return dict(dir_status, **file_status)
 
+    def hashes_exist(self, hashes: set, **kwargs):
         hashes = set(hashes)
```
efiop, line 124: This conversion that we do might be causing things like https://github.com/iterative/dvc/issues/4139 . Might want to check.

dvc/tree/local.py
```diff
@@ -32,7 +25,6 @@ class LocalTree(BaseTree):
     PARAM_CHECKSUM = "md5"
     PARAM_PATH = "path"
     TRAVERSE_PREFIX_LEN = 2
-    UNPACKED_DIR_SUFFIX = ".unpacked"
 
     CACHE_MODE = 0o444
     SHARED_MODE_MAP = {None: (0o644, 0o755), "group": (0o664, 0o775)}
```
efiop, line 30: This needs to be moved to cache as well.

Issue #4504, opened 2020-08-31
configuration based job limit


wdixon:
Seems that the current DVC push defaults --jobs to 4 * cpu_count()...  This is probably too high for larger machines - leading to file descriptor limits...    It would be nice to have this as a config option, so you don't have to always specify --jobs to limit the concurrency.

### Please provide information about your setup

**Output of `dvc version`:**

```console
DVC version: 1.6.3+828d1a 
---------------------------------
Platform: Python 3.8.5 on Linux-4.15.0-64-generic-x86_64-with-glibc2.10
Supports: http, https, s3
```

**Additional Information (if any):**

If applicable, please also provide a `--verbose` output of the command, eg: `dvc add --verbose`.

jenishah:
Hi, I would like to  work on this. Can someone suggest me on how the new [schema](https://github.com/iterative/dvc/blob/fcdb503b4eff1297e5c1c1ed18f5606fe780e481/dvc/config.py#L134) should look like ?

efiop:
@jenishah Sure! You will simply need to add `jobs` to `REMOTE_COMMON` there and that should be enough.

I159:
Hey there. I'm going to start it tonight.

efiop:
Fixed by https://github.com/iterative/dvc/pull/4872 . Thanks @I159 ! :pray:

PR #4872, opened 2020-11-10
[WIP]4504: configuration-based job limit.

I159:
## Configuration-based workers number limit. 

Introduced configuration value for the jobs number and prioritised over the default `tree` variable.

### Unknown

I haven't understood how and where the remote config used. I'm pretty sure that there is at least one place where the logic should be applied. 

### DIFF
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -103,6 +103,7 @@ class RelPath(str):
 REMOTE_COMMON = {
     "url": str,
     "checksum_jobs": All(Coerce(int), Range(1)),
+    Optional("jobs", default=4): int,
     Optional("no_traverse"): Bool,  # obsoleted
     "verify": Bool,
 }
@@ -135,6 +136,7 @@ SCHEMA = {
     "core": {
         "remote": Lower,
         "checksum_jobs": All(Coerce(int), Range(1)),
+        Optional("jobs", default=4): int,
         Optional("interactive", default=False): Bool,
         Optional("analytics", default=True): Bool,
         Optional("hardlink_lock", default=False): Bool,
--- a/dvc/tree/base.py
+++ b/dvc/tree/base.py
@@ -80,13 +80,16 @@ class BaseTree:
 
     @cached_property
     def jobs_count(self):
+        """Number of workers per cpu core."""
+
         # NOTE: if there wasn't the same snippet here I'd rather do it in the following
         # style:
         # repo_config = self.repo.config['core'] if self.repo else {}
-        # return self.config.get('jobs', repo_config.get('jobs', self.JOBS))
-        return self.config.get("jobs") or (
+        # jobs_per_core = self.config.get('jobs', repo_config.get('jobs', self.JOBS))
+        jobs_per_core = self.config.get("jobs") or (
             (self.repo and self.repo.config["core"].get("jobs")) or self.JOBS
         )
+        return jobs_per_core * cpu_count()
 
     @cached_property
     def hash_jobs(self):
--- a/dvc/tree/base.py
+++ b/dvc/tree/base.py
@@ -79,6 +79,16 @@ class BaseTree:
         self.path_info = None
 
     @cached_property
+    def jobs_count(self):
+        # NOTE: if there wasn't the same snippet here I'd rather do it in the following
+        # style:
+        # repo_config = self.repo.config['core'] if self.repo else {}
+        # return self.config.get('jobs', repo_config.get('jobs', self.JOBS))
+        return self.config.get("jobs") or (
+            (self.repo and self.repo.config["core"].get("jobs")) or self.JOBS
+        )
+
+    @cached_property
     def hash_jobs(self):
         return (
             self.config.get("checksum_jobs")
@@ -413,7 +423,7 @@ class BaseTree:
                     dir_mode=dir_mode,
                 )
             )
-            with ThreadPoolExecutor(max_workers=self.JOBS) as executor:
+            with ThreadPoolExecutor(max_workers=self.jobs_count) as executor:
                 futures = [
                     executor.submit(download_files, from_info, to_info)
                     for from_info, to_info in zip(from_infos, to_infos)


dvc/remote/ssh.py
```diff
@@ -29,6 +29,7 @@ def _exists(chunk_and_channel):
 
         with self.tree.ssh(path_infos[0]) as ssh:
             channels = ssh.open_max_sftp_channels()
+            # NOTE: is the config-based max workers limit relevant here?
```
I159, line 32: Hi, @efiop. May I ask for some clarification of the design?

 I'm not sure, how and should I use the configuration-based jobs (workers) limit here? Why is it based on sftp channels count there and on the CPU cores count in the base tree?

I159 (commented):
Clarifications requested.

dvc/cache/ssh.py
```diff
@@ -30,6 +30,7 @@ def _exists(chunk_and_channel):
 
         with self.tree.ssh(path_infos[0]) as ssh:
             channels = ssh.open_max_sftp_channels()
+            # NOTE: is the config-based max workers limit relevant here?
```
efiop, line 33: It is not. We open as many sftp channels as we can per JOBS ssh connections.

dvc/config.py
```diff
@@ -103,6 +103,7 @@ class RelPath(str):
 REMOTE_COMMON = {
     "url": str,
     "checksum_jobs": All(Coerce(int), Range(1)),
+    Optional("jobs", default=4): int,
```
efiop, line 106: Why 4 though? Each tree has itsown default number of jobs. Won't this `default=4` always override it?

dvc/tree/base.py
```diff
@@ -48,7 +48,7 @@ class BaseTree:
     scheme = "base"
     REQUIRES = {}
     PATH_CLS = URLInfo
-    JOBS = 4 * cpu_count()
+    JOBS = 4
```
efiop, line 51: This is misleading. And also seems incorrect, as there is no way to have `jobs = 1` in the config, as it will now always multiply it by NCPU in `jobs_count` below.

dvc/tree/base.py
```diff
@@ -77,6 +77,19 @@ def __init__(self, repo, config):
...
+    def jobs_count(self):
+        """Number of workers per cpu core."""
+
+        # NOTE: if there wasn't the same snippet here I'd rather do it in the following
+        # style:
+        # repo_config = self.repo.config['core'] if self.repo else {}
+        # jobs_per_core = self.config.get('jobs', repo_config.get('jobs', self.JOBS))
+        jobs_per_core = self.config.get("jobs") or (
+            (self.repo and self.repo.config["core"].get("jobs")) or self.JOBS
+        )
```
efiop, line 90: Existing --jobs and cls.JOBS were never about jobs-per-core, but rather an absolute number of jobs. This breaks SSH, for example, which has JOBS=4 but now it will spawn 4*NCPU

dvc/config.py
I159: I kept it close to the previous implementation, but more flexible. As I understand it now, the config-based value is not optional and must always be in the config. Is it?

dvc/tree/base.py
I159: Ok, got it, is the value must be independent of CPU count?

dvc/tree/base.py
I159: Thank you! It is clear now.

I159 (commented):
@efiop, thank you for the review! I'm implementing the improvements.

dvc/cache/ssh.py
I159: Great! Thank you!

dvc/config.py
```diff
@@ -103,6 +103,7 @@ class RelPath(str):
 REMOTE_COMMON = {
     "url": str,
     "checksum_jobs": All(Coerce(int), Range(1)),
+    "jobs": All(Coerce(int), Range(1)),
```
I159, line 106: With no defaults and fancy multiplications, the field is required and could be only `1 <=`.

dvc/config.py
efiop: No, it should stay optional.

dvc/tree/base.py
efiop: Yes, same as it was before.

dvc/config.py
efiop: >  the field is required

That's actually false :slightly_smiling_face:  Unless `Required()` is used, every field is considered `Optional`. We are using `voluptuous` for Schema, which has such default settings. A bit confusing, I know :slightly_smiling_face:

dvc/tree/base.py
```diff
@@ -48,7 +48,6 @@ class BaseTree:
     scheme = "base"
     REQUIRES = {}
     PATH_CLS = URLInfo
-    JOBS = 4 * cpu_count()
```
efiop, line 51: Doesn't seem like there is a need to delete this. Or am I missing something?

dvc/tree/base.py
```diff
@@ -404,7 +408,7 @@ def _download_dir(
                     dir_mode=dir_mode,
                 )
             )
-            with ThreadPoolExecutor(max_workers=self.JOBS) as executor:
```
efiop, line 407: So now `self.JOBS`( a default number of jobs provided by each tree) is not used at all. This doesn't look right.

efiop: @I159 Take a look at `hash_jobs` property, you need to do pretty much the same thing.

dvc/tree/base.py
I159: Well... after I understood the way how `voloptus` validator types work I see that there is no need to delete it, even more, it is breaking the logic :)

dvc/tree/base.py
I159: Yep, I reached this idea. 🙏🏼

I159 (commented):
The comments understood and addressed. Thanks!

dvc/tree/base.py
I159, line 51: Since multiplication of a default jobs count by the number of CPU cores caused exceeding of file descriptor limit on the machines with multiple CPU cores I see a reason to avoid the multiplication everywhere including the default class attribute.

dvc/config.py
```diff
@@ -135,6 +136,7 @@ class RelPath(str):
     "core": {
         "remote": Lower,
         "checksum_jobs": All(Coerce(int), Range(1)),
+        "jobs": All(Coerce(int), Range(1)),
```
efiop, line 139: Let's not add this one, as it might be harmful for some remotes.
```suggestion
```

dvc/tree/base.py
efiop: @I159 This is out of scope for this change. Let's keep the defaults the same they were before.

dvc/tree/base.py
```diff
@@ -77,6 +77,17 @@ def __init__(self, repo, config):
         self.verify = config.get("verify", self.DEFAULT_VERIFY)
         self.path_info = None
 
+    @cached_property
+    def jobs_count(self):
```
efiop, line 81: Let's name it
```suggestion
    def jobs(self):
```

dvc/config.py
I159: Could you please explain the usage difference between `"core"` and `REMOTE_COMMON`. I found that they are not always overlapping (used together) and to me, it looked like `"jobs"` field has to be in `"core"` as well for the cases when it used not extended with the `REMOTE_COMMON`. I definitely hadn't investigated all the architecture of the configs. So, I'm wondering, why is it ok to have the `"jobs"` field only in the `REMOTE_COMMON`?

dvc/tree/base.py
I159: Got it, reverting the default. 🙏🏼

I159 (commented):
@efiop, thanks for all the clarifications and driving the change!

I159:
Looks simple enough to be correct now. Are there any missed places to use the newly created property?

dvc/config.py
efiop: Not sure why it was marked as resolved. Unresolving...

@I159 `core` is a common section for everything. Like `core` section in git config. `REMOTE_COMMON` is a common set of options used in `remote *` sections, like in git config. Putting jobs (which only have sense for remotes) into the core section doesn't make much sense, as it is not generic enough to be applied to every remote type and might break some. There are some usecases where it could be useful, but I would rather not add it until someone asks for it.

dvc/tree/base.py
```diff
@@ -77,6 +77,17 @@ def __init__(self, repo, config):
         self.verify = config.get("verify", self.DEFAULT_VERIFY)
         self.path_info = None
 
+    @cached_property
+    def jobs(self):
+        # NOTE: alternative style.
+        # repo_config = self.repo.config['core'] if self.repo else {}
+        # return self.config.get("jobs", repo_config.get("jobs", self.JOBS))
```
efiop, line 84:
```suggestion
```

efiop (approved):
Looks good! :tada: There are two things missing here:

1) a test (see `test_remote_with_hash_jobs` test in tests/unit/test_remote.py and just copy it there for `jobs`)
2) Docs issue or PR, as noted in the PR checkboxes

tests/unit/remote/test_remote.py
```diff
@@ -1,5 +1,4 @@
 import pytest
-
 from dvc.tree import get_cloud_tree
```
efiop, line 2: Unrelated change
```suggestion

from dvc.tree import get_cloud_tree
```

tests/unit/remote/test_remote.py
```diff
@@ -1,5 +1,4 @@
 import pytest
-
```
I159: Ouch, that is my auto-formatter.

efiop (approved):

dvc/tree/base.py
```diff
@@ -77,6 +77,15 @@ def __init__(self, repo, config):
         self.verify = config.get("verify", self.DEFAULT_VERIFY)
         self.path_info = None
 
+    @cached_property
+    def jobs(self):
+        """Number of jobs (workers) set in a configuration"""
```
efiop, line 82:
```suggestion
```

efiop:
Thank you @I159 ! :pray:

jorgeorpinel:
Hi! Quick question, this didn't close #4504?

efiop:
@jorgeorpinel Indeed, thanks for the heads up! :pray:

PR #4890, opened 2020-11-13
experiments: preserve file permissions across tmpdir execution

### LINKED ISSUE #4850: Experiments: Executable permissions are removed
## Bug Report
It seems that running `dvc exp run` results in +x permissions being removed from files.

**Before**
```
-rwxr-xr-x  1 nte nte    541 Nov  4 23:58 diff-report.sh
-rw-rw-r--  1 nte nte    212 Oct 19 16:49 docker-compose.util.yml
-rw-rw-r--  1 nte nte    591 Oct 19 16:49 docker-compose.yml
-rw-r--r--  1 nte nte   4507 Nov  5 16:43 dvc.lock
-rw-r--r--  1 nte nte   2428 Nov  4 23:10 dvc.yaml
drwxr-xr-x  3 nte nte   4096 Nov  5 16:46 model
drwxr-xr-x  3 nte nte   4096 Nov  4 23:10 notebooks
-rw-r--r--  1 nte nte    910 Nov  5 16:43 params.yaml
-rw-r--r--  1 nte nte 245082 Nov  4 17:11 plots.html
-rwxr-xr-x  1 nte nte   1485 Nov  5 21:15 run-experiment.sh
```

**After**
```
-rw-r--r--  1 nte nte    541 Nov  6 00:16 diff-report.sh
-rw-rw-r--  1 nte nte    212 Oct 19 16:49 docker-compose.util.yml
-rw-rw-r--  1 nte nte    591 Oct 19 16:49 docker-compose.yml
-rw-r--r--  1 nte nte   4505 Nov  6 00:16 dvc.lock
-rw-r--r--  1 nte nte   2428 Nov  4 23:10 dvc.yaml
drwxr-xr-x  3 nte nte   4096 Nov  6 00:16 model
drwxr-xr-x  3 nte nte   4096 Nov  4 23:10 notebooks
-rw-r--r--  1 nte nte    870 Nov  6 00:16 params.yaml
-rw-r--r--  1 nte nte 245082 Nov  4 17:11 plots.html
-rw-r--r--  1 nte nte   1491 Nov  6 00:16 run-experiment.sh
```

Notice the different permissions in the first and last file.

Running inside a Docker container on Linux Ubuntu 20.04

**Output of `dvc version`:**

```console
$ dvc version
DVC version: 1.9.1 (pip)
Platform: Python 3.8.0 on Linux-5.4.0-52-generic-x86_64-with-glibc2.27
Supports: http, https, s3
Repo: dvc, git
```

pmrowla:

Will fix #4850.

* `BaseTree.upload()` now supports `file_mode` kwarg for applicable trees (currently LocalTree only).
    * Behaves the same way as `download(file_mode=...)` - if the tree supports modes, the specified mode will be applied.
    * For LocalTree, if mode is unspecified file will default to `LocalTree.CACHE_MODE` (same as existing behavior)

### DIFF
--- a/dvc/tree/base.py
+++ b/dvc/tree/base.py
@@ -225,7 +225,7 @@ class BaseTree:
         raise RemoteActionNotImplemented("reflink", self.scheme)
 
     @staticmethod
-    def protect(path_info):
+    def protect(path_info, mode=None):
         pass
 
     def is_protected(self, path_info):
@@ -338,7 +338,14 @@ class BaseTree:
         hash_info.size = dir_info.size
         return hash_info
 
-    def upload(self, from_info, to_info, name=None, no_progress_bar=False):
+    def upload(
+        self,
+        from_info,
+        to_info,
+        name=None,
+        no_progress_bar=False,
+        file_mode=None,
+    ):
         if not hasattr(self, "_upload"):
             raise RemoteActionNotImplemented("upload", self.scheme)
 
@@ -357,6 +364,7 @@ class BaseTree:
             to_info,
             name=name,
             no_progress_bar=no_progress_bar,
+            file_mode=file_mode,
         )
 
     def download(
--- a/dvc/tree/base.py
+++ b/dvc/tree/base.py
@@ -225,7 +225,7 @@ class BaseTree:
         raise RemoteActionNotImplemented("reflink", self.scheme)
 
     @staticmethod
-    def protect(path_info, mode=None):
+    def protect(path_info):
         pass
 
     def is_protected(self, path_info):
--- a/dvc/tree/gdrive.py
+++ b/dvc/tree/gdrive.py
@@ -564,7 +564,9 @@ class GDriveTree(BaseTree):
     def get_file_hash(self, path_info):
         raise NotImplementedError
 
-    def _upload(self, from_file, to_info, name=None, no_progress_bar=False):
+    def _upload(
+        self, from_file, to_info, name=None, no_progress_bar=False, **_kwargs
+    ):
         dirname = to_info.parent
         assert dirname
         parent_id = self._get_item_id(dirname, True)
--- a/dvc/tree/local.py
+++ b/dvc/tree/local.py
@@ -275,9 +275,9 @@ class LocalTree(BaseTree):
         else:
             self._unprotect_file(path)
 
-    def protect(self, path_info):
+    def protect(self, path_info, mode=None):
         path = os.fspath(path_info)
-        mode = self.CACHE_MODE
+        mode = mode if mode is not None else self.CACHE_MODE
 
         try:
             os.chmod(path, mode)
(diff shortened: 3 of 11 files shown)


dvc/tree/base.py
```diff
@@ -330,7 +330,14 @@ def get_dir_hash(self, path_info, **kwargs):
...
         return hash_info
 
-    def upload(self, from_info, to_info, name=None, no_progress_bar=False):
+    def upload(
+        self,
+        from_info,
+        to_info,
+        name=None,
+        no_progress_bar=False,
+        file_mode=None,
```
pmrowla, line 339: `file_mode` is already supported in `BaseTree.download()`, so I think specifying mode here fits into the existing tree API

dvc/tree/gdrive.py
```diff
@@ -564,7 +564,9 @@ def remove(self, path_info):
     def get_file_hash(self, path_info):
         raise NotImplementedError
 
-    def _upload(self, from_file, to_info, name=None, no_progress_bar=False):
+    def _upload(
+        self, from_file, to_info, name=None, no_progress_bar=False, **_kwargs
```
skshetry, line 568: Why `_kwargs`? Because of the pylint?

pmrowla: Several of the other trees already use `_kwargs` here, so this was keeping that style. May make more sense to just change everything to `kwargs` though.

pared (approved):

dvc/tree/local.py
```diff
@@ -325,7 +331,7 @@ def _upload(
             from_file, tmp_file, name=name, no_progress_bar=no_progress_bar
         )
 
-        self.protect(tmp_file)
+        self.protect(tmp_file, mode=file_mode)
```
efiop, line 334: The `protect` naming no longer makes sense if it can accept just arbitrary modes :slightly_frowning_face: But I guess it didn't make sense before, this should really be something like `self.chmod(tmp_file, file_mode)` and we should somehow handle `protect` logic in a more logical way.

tests/func/experiments/test_experiments.py
```diff
@@ -34,6 +39,13 @@ def test_new_simple(tmp_dir, scm, dvc, exp_stage, mocker):
     ).read_text() == "foo: 2"
 
 
+@pytest.mark.skipif(os.name == "nt", reason="Not supported for Windows.")
```
efiop, line 42: NTFS also doesn't have +x, right? Just clarifying.

We have some `not supported on windows` tests that are really about ntfs and not windows itself. Just important for us to keep that in mind :slightly_smiling_face:

pmrowla: in Windows Python's `os.chmod` ignores everything except for the user read/write bits, regardless of filesystem, so for this test I think it's not NTFS specific.

As far as using NTFS outside of Windows, user's would probably run into issues w/experiments and properly preserving file mode unless they use git with `core.filemode false`, but that's not really a DVC specific issue

dvc/tree/local.py
```diff
@@ -325,7 +332,10 @@ def _upload(
             from_file, tmp_file, name=name, no_progress_bar=no_progress_bar
         )
 
-        self.protect(tmp_file)
+        if file_mode is not None:
+            self.chmod(tmp_file, file_mode)
+        else:
+            self.protect(tmp_file)
```
pmrowla, line 338: It still looks weird to be calling `protect` from inside `tree._upload` but chmod is separated from protect now

PR #4892, opened 2020-11-13
ci: add macos-11 @ py3.8, py3.9 to build matrix

pmrowla:

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -37,8 +37,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-18.04, windows-2019, macos-10.15]
+        os: [ubuntu-18.04, windows-2019, macos-10.15, macos-11.0]
         pyv: ["3.6", "3.7", "3.8", "3.9"]
+        exclude:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -37,8 +37,13 @@ jobs:
...
       fail-fast: false
       matrix:
-        os: [ubuntu-18.04, windows-2019, macos-10.15]
+        os: [ubuntu-18.04, windows-2019, macos-10.15, macos-11.0]
         pyv: ["3.6", "3.7", "3.8", "3.9"]
+        exclude:
+          - os: macos-11.0
+            pyv: "3.6"
+          - os: macos-11.0
+            pyv: "3.7"
```
pmrowla, line 46: Based on reading some homebrew tickets, Python 3.8 and 3.9 builds properly on both arm and x86, 3.7 builds appears to still be flaky. Upstream cpython looks like they aren't going to support macos 11 at all on <3.7.

Looks like macos 11 includes 3.8 as the default system python version, and the homebrew default is already 3.9, so it's unlikely that we would even need to support 3.7 in macos 11.

PR #4901, opened 2020-11-16
dvcfile: add desc field for outs/stages

efiop:
Outs desc will be used by Viewer right away. Stage desc is for people to
start using now, but we won't be using it in viewer until DAG support is
added there.                                                            

This doesn't yet preserve the `desc` when re-`dvc add/run` something.
Same as with `meta`, that needs to be addressed separately in the follow-up PR.

https://github.com/iterative/dvc.org/pull/1951

### DIFF
--- a/dvc/command/add.py
+++ b/dvc/command/add.py
@@ -21,6 +21,7 @@ class CmdAdd(CmdBase):
                 fname=self.args.file,
                 external=self.args.external,
                 glob=self.args.glob,
+                desc=self.args.desc,
             )
 
         except DvcException:
@@ -68,6 +69,15 @@ def add_parser(subparsers, parent_parser):
         "--file",
         help="Specify name of the DVC-file this command will generate.",
         metavar="<filename>",
+    )
+    parser.add_argument(
+        "--desc",
+        type=str,
+        metavar="<text>",
+        help=(
+            "User description of the data (optional). "
+            "This doesn't affect any DVC operations."
+        ),
     )
     parser.add_argument(
         "targets", nargs="+", help="Input files/directories to add.",
--- a/dvc/command/imp.py
+++ b/dvc/command/imp.py
@@ -18,6 +18,7 @@ class CmdImport(CmdBase):
                 fname=self.args.file,
                 rev=self.args.rev,
                 no_exec=self.args.no_exec,
+                desc=self.args.desc,
             )
         except DvcException:
             logger.exception(
@@ -71,5 +72,14 @@ def add_parser(subparsers, parent_parser):
         action="store_true",
         default=False,
         help="Only create DVC-file without actually downloading it.",
+    )
+    import_parser.add_argument(
+        "--desc",
+        type=str,
+        metavar="<text>",
+        help=(
+            "User description of the data (optional). "
+            "This doesn't affect any DVC operations."
+        ),
     )
     import_parser.set_defaults(func=CmdImport)
--- a/dvc/command/imp_url.py
+++ b/dvc/command/imp_url.py
@@ -16,6 +16,7 @@ class CmdImportUrl(CmdBase):
                 out=self.args.out,
                 fname=self.args.file,
                 no_exec=self.args.no_exec,
+                desc=self.args.desc,
             )
         except DvcException:
             logger.exception(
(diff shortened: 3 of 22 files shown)


dvc/stage/serialize.py
```diff
@@ -119,6 +119,7 @@ def to_pipeline_file(stage: "PipelineStage"):
 
     outs, metrics, plots = _serialize_outs(stage.outs)
     res = [
+        (stage.PARAM_DESC, stage.desc),
         (stage.PARAM_CMD, stage.cmd),
```
efiop, line 123: Adding it before `cmd` since it seems to feel more natural.

dvc/stage/utils.py
```diff
@@ -199,6 +199,7 @@ def get_dump(stage):
     return {
         key: value
         for key, value in {
+            stage.PARAM_DESC: stage.desc,
```
efiop, line 202: Same as in https://github.com/iterative/dvc/pull/4901/files#r524478517

dvc/dependency/__init__.py
```diff
@@ -53,6 +53,7 @@
 SCHEMA = output.SCHEMA.copy()
 del SCHEMA[BaseOutput.PARAM_CACHE]
 del SCHEMA[BaseOutput.PARAM_METRIC]
+del SCHEMA[BaseOutput.PARAM_DESC]
```
efiop, line 56: This desperately needs a refactor :slightly_smiling_face:

dvc/output/base.py
```diff
@@ -293,7 +295,12 @@ def commit(self):
             self.cache.save(self.path_info, self.cache.tree, self.hash_info)
 
     def dumpd(self):
-        ret = copy(self.hash_info.to_dict())
+        ret = {}
+
+        if not self.IS_DEPENDENCY and self.desc:
+            ret[self.PARAM_DESC] = self.desc
```
efiop, line 301: Ugly because Output/DependencyBase needs a refactor (will try to send one after this PR). But the jist is that we're setting desc first, as it is more intuitive to have before the hash.

dvc/repo/add.py
```diff
@@ -194,6 +183,8 @@ def _create_stages(
         )
         if stage:
```
efiop, line 184: Desc is worth taking into account when checking if the stage is cached. Right now it ignores it :slightly_frowning_face:

dvc/command/add.py
```diff
@@ -69,6 +70,9 @@ def add_parser(subparsers, parent_parser):
         help="Specify name of the DVC-file this command will generate.",
         metavar="<filename>",
     )
+    parser.add_argument(
+        "--desc", type=str, metavar="<description>", help="Data description.",
```
jorgeorpinel, line 74: I think we use pseudo var types in `metavar`? Like
```suggestion
        "--desc", type=str, metavar="<string>", help="Data description.",
```
Would just need to double check with existing ones (maybe use `<text>` instead if it's already out there).

jorgeorpinel: Also, `Data description`: Maybe we can be slightly more informative like "User description of the data (optional)". Maybe even append "This has no effect for DVC." or something like that (if that's true)?

dvc/command/imp.py
```diff
@@ -72,4 +73,7 @@ def add_parser(subparsers, parent_parser):
         default=False,
         help="Only create DVC-file without actually downloading it.",
     )
+    import_parser.add_argument(
+        "--desc", type=str, metavar="<description>", help="Data description.",
```
jorgeorpinel, line 77: Same

dvc/command/imp_url.py
```diff
@@ -67,4 +68,7 @@ def add_parser(subparsers, parent_parser):
         default=False,
         help="Only create DVC-file without actually downloading it.",
     )
+    import_parser.add_argument(
+        "--desc", type=str, metavar="<description>", help="Data description.",
```
jorgeorpinel, line 72: "

dvc/stage/params.py
```diff
@@ -11,3 +11,4 @@ class StageParams:
     PARAM_PARAMS = "params"
     PARAM_METRICS = "metrics"
     PARAM_PLOTS = "plots"
+    PARAM_DESC = "desc"
```
efiop, line 14: We currently have PARAM_DESC in outs and in stages, because there is no general logic present for common fields. We'll be likely introducing them in the future with things like `if:`/etc, but for now keeping as is.

Issue #2042, opened 2019-05-22
Support declaring environment variables in DVC file


robogeek:
Obviously some command-line tool configuration is set using environment variables.  DVC does not support setting environment variables in a DVC file.  Neither [DVC file reference](https://dvc.org/doc/user-guide/dvc-file-format) nor [dvc run command reference](https://dvc.org/doc/commands-reference/run) show anything about setting an environment variable.

As to how it should be visible to the users:

* `dvc run -e VARIABLE=value` -- support setting this option multiple times
* In the top level of a DVC file have an `env` tag, like so:

```yaml
cmd: echo This is a command
env:
    - VARIABLE_NAME=value
    - VARIABLE_NAME2=value
    - VARIABLE_NAME3=value
```

OR

```yaml
cmd: echo This is a command
env:
    VARIABLE_NAME: value
    VARIABLE_NAME2: value
    VARIABLE_NAME3: value
```

For comparison, this is exactly how environment variables are defined in a Docker Compose file: https://docs.docker.com/compose/compose-file/#environment

efiop:
Great idea @robogeek ! We've actually been discussing something like this with @mroutis and @pared . Along with dvc being able to resolve things like `$(VARIABLE_NAME)` in commands and paths, it should get us covered for the parametrization feature as well. :slightly_smiling_face:

shcheklein:
might be a duplicate of https://github.com/iterative/dvc/issues/1416 to some extent, there was a discussion about introducing `-e`

efiop:
@shcheklein Indeed, closing in favor of that.

shcheklein:
@robogeek please, give a vote to that initial ticket. And thanks for the feedback :)

Issue #1416, opened 2018-12-05
pipelines: parametrize using environment variables / DVC properties


prihoda:
It would be useful if I could parametrize my pipeline using environment variables, which could be read from a properties file specified using `dvc config env my.properties`. DVC would load those environment variables when running the command.

For example, I could have this properties file:
```
DVC_NICKNAME=David
```

And run:
```
dvc run -o hello.txt 'echo "Hello ${DVC_NICKNAME}!" > hello.txt'
dvc run -o cheers.txt 'echo "Cheers ${DVC_NICKNAME}!" > cheers.txt'
```
And produce "Hello David!" and "Cheers David!" files. 

Users would just have to make sure to quote the command or use interactive mode #1415.

The DVC file would contain the variable reference:
```
cmd: echo "Hello ${DVC_NICKNAME}!" > hello.txt
```

The value would be added to the environment by DVC at DVC startup so it would be handled natively by the shell.

In order for `dvc status` to be able to detect that variables in a stage changed, we can calculate the internal md5 checksum on contents with the variable values injected in place of the variable names, so that it would be handled as if the contents of the DVC file changed. This can be done using [os.path.expandvars](https://docs.python.org/3/library/os.path.html#os.path.expandvars). But unfortunately, this would just replace variable references used directly in the shell command, it would not cover cases where you're using the environment variable inside a script. The only foolproof way would be force the user to explicitly request environment variables that would be injected from the properties file, e.g. using `dvc run -e DVC_NICKNAME -e DVC_OTHER`. That would basically allow adding additional "env dependencies" to stages.

It would be nice to inject the variables also into paths to dependencies, so that you can parametrize those as well. Could also be done using [os.path.expandvars](https://docs.python.org/3/library/os.path.html#os.path.expandvars). This would change the DAG dynamically, but AFAIK it should actually magically work without breaking anything, right? As long as you just initialize the environment at each DVC startup and call expandvars when reading deps paths.

mroutis:
This one is interesting, @prihoda :thinking: 

I've never encounter the of using varibables in DVC commands, maybe my use cases are very simple :sweat_smile: !

This looks like a _Makefile_ behavior, where you can define variables on the top of your file and use them in the rules.

I prefer being "explicit than implicit" and I'll think it twice before introducing this request.
Let's leave this open and wait for some thumbs up or comments from other users :slightly_smiling_face: 

[...]

PR #4902, opened 2020-11-16
live logs initial

pared:

Related to https://github.com/iterative/dvc.org/pull/1966


dvc/repo/plots/__init__.py
```diff
@@ -53,18 +53,31 @@ def collect(self, targets=None, revs=None):
...
+                for path, repo_path in plot_files:
+                    data[rev].update({repo_path: {"props": props}})
+
+                    # Load data from git or dvc cache
+                    try:
+                        with tree.open(path) as fd:
+                            data[rev][repo_path]["data"] = fd.read()
+                    except FileNotFoundError:
+                        # This might happen simply because cache is absent
+                        pass
```
pared, line 80: We start to support `dir` plots: all the files in the directory marked as `plots` will be treated the same way, and additional config provided with `dvc plots modify` will be applied to each one of them.
I intend to write support for "granular" config in such use case, though it will probably require some reworking on `dvc.yaml` side and is not essential to make logs/live metrics work.

dvc/repo/plots/data.py
```diff
@@ -152,7 +146,7 @@ def _append_revision(data_points, revision, **kwargs):
 
 class PlotData:
     REVISION_FIELD = "rev"
-    INDEX_FIELD = "index"
+    INDEX_FIELD = "step"
```
pared, line 149: Seems like `step` is generic enough to replace `index` - in both cases they do not make too much sense when plotting anything, so we agreed with @dmpetrov that we can replace it for the sake of "additional config"-less integration with logs.

efiop: What about backward compatibility?

pared: @efiop What do you mean by backward compatibility?
`INDEX_FIELD` is appended dynamically to data if `-x` [option is not provided](https://github.com/iterative/dvc/blob/4cf2f8139bdd2c30d439dea1bb7375ddb63e46d0/dvc/repo/plots/data.py#L132). So, if the question is whether we will break repositories already using `plots`? No, we will not, we will just replace `index` with `step` in the actual plot.
But if you mean whether default behavior of plots will change, then yes, since the default `x` axis name changes.

dvc/api/dvclive.py
```diff
@@ -0,0 +1,14 @@
...
+from dvc.repo import Repo
+
+
+def summary(path: str, revs: List[str] = None):
+    try:
+        root = Repo.find_root()
+    except NotDvcRepoError:
+        root = os.getcwd()
+
+    Repo(root_dir=root, uninitialized=True).logs.summarize(path, revs)
```
pared, line 14: Not sure we need that after introducing `logs` command. Might get deleted later.

dvc/command/logs.py
```diff
@@ -0,0 +1,43 @@
...
+import logging
+
+from dvc.command import completion
+from dvc.command.base import CmdBase
+from dvc.exceptions import DvcException
+
+logger = logging.getLogger(__name__)
+
+
+class CmdLogs(CmdBase):
```
pared, line 11: We need this command for a scenarion as in https://github.com/iterative/dvclive/issues/7

dvc/repo/logs.py
```diff
@@ -0,0 +1,35 @@
...
+import os
+from typing import List
+
+from dvc.repo import Repo
+from dvc.visualization import embed, metrics_embedding, plots_embeddings
+
+logger = logging.getLogger(__name__)
+
+
+class Logs:
```
pared, line 11: Naming `dvclive`/`logs` is not set in stone. It will probably be changed. Ideas are welcomed.

jorgeorpinel:
Hi. Is there an OP (issue) explaining this feature and/or with specs/design? Thanks!

skshetry:
@pared, is this ready for a review?

pared:
@skshetry sorry, not yet

dvc/stage/run.py
```diff
@@ -56,6 +56,8 @@ def prepare_kwargs(stage, checkpoint_func=None):
         # indicate that checkpoint cmd is being run inside DVC
         kwargs["env"].update(_checkpoint_env(stage))
 
+    kwargs["env"].update(stage.environment)
```
pared, line 59: @pmrowla 
Introducing `checkpoints` and `dvclive` requires us to set `env` for particular stages.
It seems to me that it worth handling on `stage` level, and might help us if we ever decide to support user-defined variables in stages. If we decide to go that way, I can move `_checkpoint_env` to `stage`.
Related: #2042 #1416 #3633

dvc/repo/collect.py
```diff
@@ -42,7 +42,7 @@ def _collect_paths(
         if recursive and tree.isdir(path_info):
             target_infos.update(set(tree.walk_files(path_info)))
 
-        if not tree.isfile(path_info):
```
pared, line 45: Supporting `dir plots` requires to pass this check for dirs too.

dvc/repo/metrics/show.py
```diff
@@ -55,7 +68,7 @@ def _read_metrics(repo, metrics, rev):
 
     res = {}
     for metric in metrics:
-        if not tree.exists(metric):
```
pared, line 58: Supporting `dir plots` allows us to `target` dirs during `collect`. Since we also support `target`-ing non-metrics, it is possible that someone will run `dvc metrics show dir`. In that situation, we would try to parse dir with yaml parser. I am narrowing this check, to avoid such a situation.

dvc/repo/plots/__init__.py
```diff
@@ -77,7 +90,7 @@ def render(data, revs=None, props=None, templates=None):
         plots = _prepare_plots(data, revs, props)
 
         result = {}
-        for datafile, desc in plots.items():
+        for datafile, desc in sorted(plots.items()):
```
pared, line 93: Just to keep our html consistient.

dvc/repo/plots/__init__.py
```diff
@@ -95,16 +108,28 @@ def render(data, revs=None, props=None, templates=None):
...
         targets = [targets] if isinstance(targets, str) else targets or []
         for target in targets:
             rpath = relpath(target, self.repo.root_dir)
-            if not any("data" in d[rpath] for d in data.values()):
+            if not any(
+                "data" in rev_data[key]
+                for rev_data in data.values()
+                for key, d in rev_data.items()
+                if rpath in key
+            ):
```
pared, line 132: since we support `dir plots` we need to consider possibility that data paths overlap.
Eg:
`dvc plots show directory` might result in {"directory/plot1.json":{...}, "directory/plot2.json":{}}.

tests/func/test_dvclive.py
```diff
@@ -0,0 +1,103 @@
...
+        import sys
+        r = 5
+        for i in range(r):
+           dvclive.log("loss", -i/5)
+           dvclive.log("accuracy", i/5)"""
+)
+
+
+@pytest.mark.parametrize("summary", (True, False))
+def test_export_config_tmp(tmp_dir, dvc, mocker, summary):
```
pared, line 20: placeholder test until we release `dvclive`

dvc/output/base.py
```diff
@@ -76,6 +78,8 @@ class BaseOutput:
     PARAM_PLOT_HEADER = "header"
     PARAM_PERSIST = "persist"
     PARAM_DESC = "desc"
+    PARAM_DVCLIVE = "dvclive"
```
efiop, line 81: Maybe just `live`?

dvc/schema.py
```diff
@@ -77,6 +83,7 @@
         Any(str, OUT_PSTAGE_DETAILED_SCHEMA)
     ],
     Optional(StageParams.PARAM_PLOTS): [Any(str, PLOT_PSTAGE_SCHEMA)],
+    Optional(StageParams.PARAM_DVCLIVE): [Any(str, DVCLIVE_PSTAGE_SCHEMA)],
```
efiop, line 86: Since there can only be 1 dvclive entry, maybe we could limit it in schema too in addition to your assert?

dvc/command/live.py
```diff
@@ -0,0 +1,44 @@
...
+
+class CmdLive(CmdBase):
+    UNINITIALIZED = True
+
+    def run(self):
+        try:
+            self.repo.dvclive.summarize(self.args.target, self.args.rev)
+        except DvcException:
+            logger.exception("")
+            return 1
```
efiop, line 19:
```suggestion
        self.repo.dvclive.summarize(self.args.target, self.args.rev)
```

dvc/command/live.py
```diff
@@ -0,0 +1,44 @@
...
+
+        return 0
+
+
+def add_parser(subparsers, parent_parser):
+
+    logs_parser = subparsers.add_parser(
+        "live",
+        parents=[parent_parser],
+        formatter_class=argparse.RawDescriptionHelpFormatter,
```
efiop, line 29: No help?

dvc/repo/dvclive.py
```diff
@@ -0,0 +1,52 @@
...
+    if has_summary:
+        return out.path_info.with_suffix(".json")
+    return None
+
+
+class DvcLive:
+    def __init__(self, repo: Repo):
+        self.repo = repo
+
+    def summarize(self, path: str, revs: List[str] = None):
```
efiop, line 29: Maybe let's do `summarize` -> `show` so it matches the other commands (metrics, plots, etc)? As I understand it is not user facing. Also `dvclive,py` -> `live.py`

dvc/stage/utils.py
```diff
@@ -57,6 +57,27 @@ def fill_stage_outputs(stage, **kwargs):
...
     stage.outs = []
+
+    dvclive_l = kwargs.get("dvclive", [])
+    if dvclive_l:
+        if len(dvclive_l) != 1:
+            from dvc.exceptions import DvcException
+
+            raise DvcException("Only one dvclive output allowed!")
+
+        from dvc.output import BaseOutput
```
efiop, line 68: Please don't import mid-method unless you have a very good reason to do that.

efiop: Might want to consider moving this dvclive section into a separate helper.

dvc/visualization.py
```diff
@@ -0,0 +1,46 @@
+from typing import Dict, Iterable, List
```
efiop, line 1: for the record: `visualization.py` sounds weird, discussed possibly renaming it and maybe even moving it somewhere else to utils or somtehing.

dvc/output/base.py
pared: Renamed, as well as other internal names.

dvc/api/dvclive.py
```diff
@@ -0,0 +1,14 @@
...
+
+from dvc.exceptions import NotDvcRepoError
+from dvc.repo import Repo
+
+
+def summary(path: str, revs: List[str] = None):
+    try:
+        root = Repo.find_root()
+    except NotDvcRepoError:
+        root = os.getcwd()
```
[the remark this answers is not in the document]
pared: This method will be called by `dvclive` lib to continuously produce a training summary report (HTML). In order to do that, we need a repo. But we need to take into account, that one might actually use `dvclive` without using DVC. In that case, we can do similar thing as we do for `plots` and `metrics`: uninitialized repo. I will prepare test for such use case.

dvc/visualization.py
pared: moved to `dvc/utils/html.py`, also refactored embedding of plots and metrics.

dvc/stage/utils.py
pared: created `_load_live_outputs`

dvc/repo/dvclive.py
pared: done, also adjusted returned values to match `metrics` / `plots`

dvc/command/live.py
pared: added description

dvc/command/live.py
```diff
@@ -0,0 +1,95 @@
...
+        help=DIFF_HELP,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+    live_diff_parser.add_argument(
+        "--revs",
+        nargs="*",
+        default=None,
+        help="Git revision (e.g. SHA, branch, tag)",
+        metavar="<commit>",
+    )
```
pared, line 80: To make `diff` compatible with `metrics` and `plots` we would need to make revisions the default argument. In order to do that, we would need to handle `no-targets` use case (when `repo` automatically gathers `live` outputs). As its still not decided whether this command will remain, and its not essential for `dvclive` to work, I leave it as is for now.

dvc/command/live.py
```diff
@@ -0,0 +1,95 @@
...
+    _add_common_arguments(live_diff_parser)
```
efiop, line 81: You are probably looking for `parents=`? https://github.com/iterative/dvc/blob/77ff24260eee5a2b2dc15b7ad9ccea456160f248/dvc/command/data_sync.py#L128

dvc/output/base.py
```diff
@@ -591,3 +600,19 @@ def merge(self, ancestor, other):
         self.hash_info = self.cache.merge(
             ancestor_info, self.hash_info, other.hash_info
         )
+
+    @cached_property
+    def environment(self) -> Dict[str, str]:
```
efiop, line 605:
```suggestion
    def env(self) -> Dict[str, str]:
```

dvc/stage/__init__.py
```diff
@@ -251,6 +251,15 @@ def is_checkpoint(self):
         """
         return any(out.checkpoint for out in self.outs)
 
+    @property
+    def environment(self) -> Dict[str, str]:
```
efiop, line 255:
```suggestion
    def env(self) -> Dict[str, str]:
```

dvc/stage/serialize.py
```diff
@@ -59,15 +61,18 @@ def _serialize_out(out):
...
-    outs, metrics, plots = [], [], []
+    outs, metrics, plots, live = [], [], [], None
     for out in sort_by_path(outputs):
         bucket = outs
         if out.plot:
             bucket = plots
         elif out.metric:
             bucket = metrics
+        elif out.live:
+            live = _serialize_out(out)
```
efiop, line 72: Just as an extra caution
```suggestion
            assert live is None
            live = _serialize_out(out)
```

dvc/command/run.py
```diff
@@ -162,6 +164,19 @@ def add_parser(subparsers, parent_parser):
...
+        default=[],
+        help=argparse.SUPPRESS,
+        metavar="<path>",
+    )
+    run_parser.add_argument(
+        "--dvclive-no-summary",
+        action="store_true",
+        default=False,
+        help=argparse.SUPPRESS,
+    )
```
efiop, line 179: Maybe `--live`? Also, you use `append`, but only one live output is allowed right now, right?

efiop (changes requested):
Please see the comments above.

Looks pretty good overall :+1: Please take a look at coverage comments though, some important new pieces of code seem to be not covered. Also, probably test refactor could've been introduced separately as a pre-requisite. It is not a blocker though. :slightly_smiling_face:

skshetry:
@pared, is there a demo repo/script to try it out with `dvclive`? 🙂

pared:
@skshetry 
There is wiki page with examples: https://github.com/iterative/dvclive/wiki Ping me if you don't have access

skshetry:
Thanks @pared, I have access to it.

dvc/command/run.py
pared: I would argue that in case of user facing interface its good to go with full name, so there is no need in remembering that "`dvclive` type of output is named as `live` in dvc itself". Though this is more elegant. I will change it, we can always redo it later.

efiop (approved):

PR #4908, opened 2020-11-19
dag: support output as target

### LINKED ISSUE #4904: dag: not accepting output files/folders

Running `dvc dag` with optional target argument that is an output file or a folder produces an error:

```console
$ dvc dag postprocessed-data/data.pickle
ERROR: failed to show a pipeline for 'None' - "Stage 'postprocessed-data/data.pickle' not found inside 'dvc.yaml' file"
```
This is despite the fact that there *is* a stage with output dependency `postprocessed-data/data.pickle` in file `dvc.yaml`.

**Output of `dvc version`:**

```console
$ dvc version
DVC version: 1.9.1 (pip)
Platform: Python 3.7.6 on Darwin-19.6.0-x86_64-i386-64bit
Supports: azure, gs, http, https, s3
Repo: dvc, git
```

efiop:
Fixes #4904 

Docs already say that we can do that, so no need to update them.

### DIFF
--- a/dvc/command/dag.py
+++ b/dvc/command/dag.py
@@ -2,7 +2,6 @@ import argparse
 import logging
 
 from dvc.command.base import CmdBase, append_doc_link
-from dvc.exceptions import DvcException
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


dvc/command/dag.py
```diff
@@ -72,7 +72,8 @@ def run(self):
         try:
             target = None
             if self.args.target:
-                stages = self.repo.collect(self.args.target)
+                pairs = self.repo.collect_granular(self.args.target)
+                stages = [stage for stage, _ in pairs]
```
efiop, line 76: For `--outs` we really need to not ignore the `filter_info` we get from pairs and possibly only show the filtered outputs in the outs DAG.

efiop: Fixed.

dvc/command/dag.py
```diff
@@ -30,80 +29,91 @@ def _show_dot(G):
...
     else:
-        H = G
+        G = repo.graph
 
-    if outs:
-        G = nx.DiGraph()
-        for stage in H.nodes:
-            G.add_nodes_from(stage.outs)
+        def _relabel(stage):
+            return stage.addressing
```
efiop, line 65: This whole `_transform` method is only here because `str(stage)` is not using `stage.addressing`. Need to consider changing `Stage.__str__` to it.

dvc/command/dag.py
```diff
@@ -30,80 +29,91 @@ def _show_dot(G):
...
-
-            return 0
-        except DvcException:
-            msg = "failed to show "
-            if self.args.target:
-                msg += f"a pipeline for '{target}'"
-            else:
-                msg += "pipelines"
-            logger.exception(msg)
-            return 1
```
efiop, line 106: Unrelated :slightly_frowning_face: , but we are gradually going away from KO errors like this in other places, so did it here too.

dvc/repo/__init__.py
```diff
@@ -648,6 +552,8 @@ def close(self):
         self.scm.close()
 
     def _reset(self):
+        self.__dict__.pop("outs_trie", None)
+        self.__dict__.pop("outs_graph", None)
```
efiop, line 556: Both of these will be very handy in https://github.com/iterative/dvc/issues/4847

PR #4932, opened 2020-11-20
import-url: add support for gdrive

efiop:
Created from https://github.com/iterative/dvc/pull/4862 to be able to use secret gdrive credentials.

### DIFF
--- a/dvc/config.py
+++ b/dvc/config.py
@@ -147,6 +147,7 @@ SCHEMA = {
     "cache": {
         "local": str,
         "s3": str,
+        "gdrive": str,
         "gs": str,
         "hdfs": str,
         "webhdfs": str,
--- a/dvc/dependency/__init__.py
+++ b/dvc/dependency/__init__.py
@@ -3,6 +3,7 @@ from urllib.parse import urlparse
 
 import dvc.output as output
 from dvc.dependency.azure import AzureDependency
+from dvc.dependency.gdrive import GDriveDependency
 from dvc.dependency.gs import GSDependency
 from dvc.dependency.hdfs import HDFSDependency
 from dvc.dependency.http import HTTPDependency
@@ -22,6 +23,7 @@ from .repo import RepoDependency
 
 DEPS = [
     AzureDependency,
+    GDriveDependency,
     GSDependency,
     HDFSDependency,
     HTTPDependency,
@@ -39,6 +41,7 @@ DEP_MAP = {
     Schemes.SSH: SSHDependency,
     Schemes.S3: S3Dependency,
     Schemes.AZURE: AzureDependency,
+    Schemes.GDRIVE: GDriveDependency,
     Schemes.GS: GSDependency,
     Schemes.HDFS: HDFSDependency,
     Schemes.HTTP: HTTPDependency,
--- a/dvc/dependency/gdrive.py
+++ b/dvc/dependency/gdrive.py
@@ -1,0 +1,7 @@
+from dvc.dependency.base import BaseDependency
+from dvc.output import BaseOutput
+from dvc.tree import GDriveTree
+
+
+class GDriveDependency(BaseDependency, BaseOutput):
+    TREE_CLS = GDriveTree
--- a/dvc/tree/gdrive.py
+++ b/dvc/tree/gdrive.py
@@ -12,6 +12,7 @@ from funcy import cached_property, retry, wrap_prop, wrap_with
 from funcy.py3 import cat
 
 from dvc.exceptions import DvcException, FileMissingError
+from dvc.hash_info import HashInfo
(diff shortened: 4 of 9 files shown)


efiop:
https://github.com/iterative/dvc/pull/4862#issuecomment-731467552

efiop:
@julsal Looks like it wasn't working :( Fixed the walk_files bug, but clearly there are more.

julsal:
@efiop 
For all UNIX-based builds:
gdrive authentication is somehow misconfigured. `GDriveTree` is trying to get a token from `CommandLineAuth()`. Maybe the problem is in the local client_secrets.json.

Some of the builds also show an HTTP 403 "User Rate Limit Exceeded", however, I presume it's due to the misconfiguration. From the logs, it was not possible to identify whether the error comes from either a misconfiguration (request without valid credentials) or an actual limited account.

Windows builds:
They raise a permission error. I [found](https://stackoverflow.com/questions/6416782/what-is-namedtemporaryfile-useful-for-on-windows) that a `tempfile.NamedTemporaryFile` cannot be read on Windows before closing. So, I changed the test code and created a pull request to `dvc/gdrive-import`. However, I'm afraid the windows builds might end up being affected by the auth problem too.

julsal:
@efiop After the last push, `NamedTemporaryFile` issue in Windows is solved. Now, errors are consistent in all builds (the authorization issue). Please, let me know if there is something I can do with this regard.

skshetry:
@julsal, is there anything special that `import-url` requires? Other GDrive tests are passing.

julsal:
@skshetry there is nothing special with this test. Locally, I'm testing with the same content that I got after configuring GDrive by the command line approach. BTW, this PR didn't change anything regarding the authentication method.

If it helps, this is the content of the `GDRIVE_CREDENTIALS_DATA` I'm currently using locally:

```json
{
   "access_token":"<token-here>",
   "client_id":"<id-here>",
   "client_secret": "<sec-here>",
   "refresh_token":"<token-here>",
   "token_expiry":"2020-11-05T12:07:44Z",
   "token_uri":"https://oauth2.googleapis.com/token",
   "user_agent":null,
   "revoke_uri":"https://oauth2.googleapis.com/revoke",
   "id_token":null,
   "id_token_jwt":null,
   "token_response":{
      "access_token":"<token-here>",
      "expires_in":3599,
      "refresh_token":"<token-here>",
      "scope":"https://www.googleapis.com/auth/drive.appdata https://www.googleapis.com/auth/drive",
      "token_type":"Bearer"
   },
   "scopes":[
      "https://www.googleapis.com/auth/drive",
      "https://www.googleapis.com/auth/drive.appdata"
   ],
   "token_info_uri":"https://oauth2.googleapis.com/tokeninfo",
   "invalid":false,
   "_class":"OAuth2Credentials",
   "_module":"oauth2client.client"
}```

**Comment by shcheklein** [comment]
@julsal it's a tricky one. Let me help you with it. 

Our tests rely of the credentials file that is generated with a `p12` key ([service account](https://dvc.org/doc/user-guide/setup-google-drive-remote#using-service-accounts)). When you do auth with a `p12` key for the first time it generates the same `.dvc/tmp/gdrive-user-credentials.json`, but content of that file is different from the regular command line auth. It contains `_private_key_pkcs12` inside, which doesn't expire. The content of this file is actually saved into the `GDRIVE_CREDENTIALS_DATA` secret in the CI system.

Thus, when you see in tests:

```
@cached_property
    def config(self):
        return {
            "url": self.url,
            "gdrive_service_account_email": "test",
            "gdrive_service_account_p12_file_path": "test.p12",
            "gdrive_use_service_account": True,
        }
```

(disregard `test.12` and `test` values- they are being ignored and set to pass the config validation. They are ignored when GDRIVE_CREDENTIALS_DATA is present already. This is a minor bug here to my mind in the validation code)

in tests, it's important that the system picks up the `gdrive_use_service_account` and goes in the right Auth flow here:

```
            if self._use_service_account:
                gauth.ServiceAuth()
            else:
                gauth.CommandLineAuth()
                GDriveTree._validate_credentials(gauth, auth_settings)
```

In your case it looks like tree object is initialized for the second (?) time and

`self._use_service_account = config.get("gdrive_use_service_account")`

is not set and/or not picked up. It goes into the `CommandLineAuth`, which breaks on the `GDRIVE_CREDENTIALS_DATA` it doesn't expect. It's interesting that at first I see that it picks it up, and only on the second time it breaks. 

-----------

I hope that makes more sense. Let me know if I can help you in some other way here.

efiop:
Thank you @julsal ! :pray: We'll take it over from now.

efiop:
No capacity to revisit this right now. Closing for now, will likely get back after 2.0 and some fsspec-inpired modifications

PR #4934, opened 2020-11-21
Allow user to pass a list of commands for one single stage

### LINKED ISSUE #4373: Support multiline command?
Something like done in Github actions with run and script?
```yaml
stages:
  stage1:
    cmd: |
      cmd1
      cmd2
```

Or, perhaps:
```yaml
stages:
  stage1:
    cmd:
      - cmd1
      - cmd2
```

Or, maybe both?

Need to double check if we will be able to keep the dump in this shape or form.

#### Why?
Better expression

#### Current Workaround:
```yaml
stages:
  stage1:
    cmd: >-
      cmd1
      && cmd2
      && cmd3
```

ClementWalter:

Fix #4373 
Doc PR: https://github.com/iterative/dvc.org/pull/1980

### DIFF
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -143,7 +143,7 @@ class Stage(params.StageParams):
 
         self.repo = repo
         self._path = path
-        self.cmd = cmd
+        self.cmd = " && ".join(cmd) if isinstance(cmd, list) else cmd
         self.wdir = wdir
         self.outs = outs
         self.deps = deps
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -143,7 +143,7 @@ class Stage(params.StageParams):
 
         self.repo = repo
         self._path = path
-        self.cmd = " && ".join(cmd) if isinstance(cmd, list) else cmd
+        self.cmd = cmd
         self.wdir = wdir
         self.outs = outs
         self.deps = deps
--- a/dvc/stage/run.py
+++ b/dvc/stage/run.py
@@ -79,7 +79,7 @@ def cmd_run(stage, *args, checkpoint_func=None, **kwargs):
         threading._MainThread,  # pylint: disable=protected-access
     )
     for _cmd in cmd:
-        logger.info("> %s", _cmd)
+        logger.info("$ %s", _cmd)
         old_handler = None
         p = None
 
--- a/dvc/stage/run.py
+++ b/dvc/stage/run.py
@@ -79,7 +79,7 @@ def cmd_run(stage, *args, checkpoint_func=None, **kwargs):
         threading._MainThread,  # pylint: disable=protected-access
     )
     for _cmd in cmd:
-        logger.info("\trunning %s", _cmd)
+        logger.info("> %s", _cmd)
         old_handler = None
         p = None
 
--- a/dvc/stage/run.py
+++ b/dvc/stage/run.py
@@ -46,13 +46,13 @@ def warn_if_fish(executable):
 @unlocked_repo
 def cmd_run(stage, *args, checkpoint_func=None, **kwargs):
     kwargs = {"cwd": stage.wdir, "env": fix_env(None), "close_fds": True}
+    cmd = stage.cmd if isinstance(stage.cmd, list) else [stage.cmd]
     if checkpoint_func:
         # indicate that checkpoint cmd is being run inside DVC
         kwargs["env"].update(_checkpoint_env(stage))
 
     if os.name == "nt":
         kwargs["shell"] = True
-        cmd = stage.cmd
     else:
         # NOTE: when you specify `shell=True`, `Popen` [1] will default to
         # `/bin/sh` on *nix and will add ["/bin/sh", "-c"] to your command.
@@ -70,32 +70,33 @@ def cmd_run(stage, *args, checkpoint_func=None, **kwargs):
         kwargs["shell"] = False
         executable = os.getenv("SHELL") or "/bin/sh"
         warn_if_fish(executable)
-        cmd = _nix_cmd(executable, stage.cmd)
+        cmd = [_nix_cmd(executable, _cmd) for _cmd in cmd]
 
     main_thread = isinstance(
         threading.current_thread(),
         threading._MainThread,  # pylint: disable=protected-access
     )
-    old_handler = None
-    p = None
+    for _cmd in cmd:
+        old_handler = None
+        p = None
 
-    try:
-        p = subprocess.Popen(cmd, **kwargs)
-        if main_thread:
-            old_handler = signal.signal(signal.SIGINT, signal.SIG_IGN)
+        try:
+            p = subprocess.Popen(_cmd, **kwargs)
+            if main_thread:
+                old_handler = signal.signal(signal.SIGINT, signal.SIG_IGN)
 
-        killed = threading.Event()
-        with checkpoint_monitor(stage, checkpoint_func, p, killed):
-            p.communicate()
-    finally:
-        if old_handler:
-            signal.signal(signal.SIGINT, old_handler)
+            killed = threading.Event()
+            with checkpoint_monitor(stage, checkpoint_func, p, killed):
+                p.communicate()
+        finally:
+            if old_handler:
+                signal.signal(signal.SIGINT, old_handler)
 
-    retcode = None if not p else p.returncode
-    if retcode != 0:
-        if killed.is_set():
-            raise CheckpointKilledError(stage.cmd, retcode)
-        raise StageCmdFailedError(stage.cmd, retcode)
+        retcode = None if not p else p.returncode
+        if retcode != 0:
+            if killed.is_set():
+                raise CheckpointKilledError(stage.cmd, retcode)
+            raise StageCmdFailedError(stage.cmd, retcode)
 
 
 def run_stage(stage, dry=False, force=False, checkpoint_func=None, **kwargs):
--- a/dvc/stage/run.py
+++ b/dvc/stage/run.py
@@ -21,7 +21,9 @@ class CheckpointKilledError(StageCmdFailedError):
     pass
 
 
-def _nix_cmd(executable, cmd):
+def _make_cmd(executable, cmd):
+    if executable is None:
+        return cmd
     opts = {"zsh": ["--no-rcs"], "bash": ["--noprofile", "--norc"]}
     name = os.path.basename(executable).lower()
     return [executable] + opts.get(name, []) + ["-c", cmd]
@@ -53,6 +55,7 @@ def cmd_run(stage, *args, checkpoint_func=None, **kwargs):
 
     if os.name == "nt":
         kwargs["shell"] = True
+        executable = None
     else:
         # NOTE: when you specify `shell=True`, `Popen` [1] will default to
         # `/bin/sh` on *nix and will add ["/bin/sh", "-c"] to your command.
@@ -70,18 +73,18 @@ def cmd_run(stage, *args, checkpoint_func=None, **kwargs):
         kwargs["shell"] = False
         executable = os.getenv("SHELL") or "/bin/sh"
         warn_if_fish(executable)
-        cmd = [_nix_cmd(executable, _cmd) for _cmd in cmd]
 
     main_thread = isinstance(
         threading.current_thread(),
         threading._MainThread,  # pylint: disable=protected-access
     )
     for _cmd in cmd:
+        logger.info("\trunning %s", _cmd)
         old_handler = None
         p = None
 
         try:
-            p = subprocess.Popen(_cmd, **kwargs)
+            p = subprocess.Popen(_make_cmd(executable, _cmd), **kwargs)
             if main_thread:
                 old_handler = signal.signal(signal.SIGINT, signal.SIG_IGN)
 
@@ -95,8 +98,8 @@ def cmd_run(stage, *args, checkpoint_func=None, **kwargs):
         retcode = None if not p else p.returncode
         if retcode != 0:
             if killed.is_set():
-                raise CheckpointKilledError(stage.cmd, retcode)
-            raise StageCmdFailedError(stage.cmd, retcode)
+                raise CheckpointKilledError(_cmd, retcode)
+            raise StageCmdFailedError(_cmd, retcode)
 
 
 def run_stage(stage, dry=False, force=False, checkpoint_func=None, **kwargs):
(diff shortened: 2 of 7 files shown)


dvc/stage/__init__.py
```diff
@@ -121,7 +121,7 @@ def __init__(
 
         self.repo = repo
         self._path = path
-        self.cmd = cmd
+        self.cmd = " && ".join(cmd) if isinstance(cmd, list) else cmd
```
efiop, line 124: Will && work the same in all shells?

Also, this is will make the `repro` create a long and ugly `cmd1 && cmd2 && cmd3 ...` in the dvc file instead of previously present nice list of commands. Doesn't seem like a good behavior.

efiop:
Hi @ClementWalter !

Thanks for the PR! :pray: Note that your commit is not recognised by github as coming from your github profile. You need to add the email used in the commit to https://github.com/settings/emails . :slightly_smiling_face:

dvc/stage/__init__.py
skshetry: The behaviour should be similar to the multiline string, rather than this `&&`.

This approach will fail on serialization/deserialization, so we need to support reading this command as a list directly on the following:
https://github.com/iterative/dvc/blob/6237a665c355020bf9c083851c8a89e874d2d437/dvc/stage/run.py#L47

And, also support list as `cmd` on the `dvc.lock` file.

efiop:
Also, please don't forget to add the issue number or even "fixes #issuenumber" to the PR desc(or commit), so github tracks it automatically. #4373  :slightly_smiling_face:

dvc/stage/__init__.py
karajan1001: cmd2,3,4... would stop if cmd1 returns some number else than 0.

ClementWalter:
thanks for the feedback. Actually I was guessing it was hacky but, so easy to write down that I give it a try. I'll work on this tonight

dvc/stage/__init__.py
ClementWalter: > cmd2,3,4... would stop if cmd1 returns some number else than 0.

I think it is the desired behavior as stated in the [issue](https://github.com/iterative/dvc/issues/4373): 

> This looks great and I think it should behave like cmd1 && cmd2 && cmd3, meaning that if one command fails then the rest is not executed and results in a stage failure.

ClementWalter:
This might be better but I am not sure about the loop and eventually I am not sure where to add tests for this change

dvc/stage/run.py
```diff
@@ -70,32 +70,33 @@ def cmd_run(stage, *args, checkpoint_func=None, **kwargs):
...
+                p.communicate()
+        finally:
+            if old_handler:
+                signal.signal(signal.SIGINT, old_handler)
+
+        retcode = None if not p else p.returncode
+        if retcode != 0:
+            if killed.is_set():
+                raise CheckpointKilledError(stage.cmd, retcode)
+            raise StageCmdFailedError(stage.cmd, retcode)
```
efiop, line 99: This will result in the whole list of commands being added to the exception. Feels like it should be a bit more granular in case of a list?

scripts/schema/dvc-yaml.json
```diff
@@ -102,8 +102,8 @@
           "type": "object",
           "properties": {
             "cmd": {
-              "type": "string",
-              "description": "Command to run"
+              "type": ["string", "array"],
+              "description": "Command(s) to run"
```
efiop, line 106: :pray:

efiop (commented):
Looks great! Just a few more comments.

dvc/stage/run.py
efiop: Also, in `run_stage`, there is a logger.info message about running the commands and right now it will show just a plain single-line list of commands:
```
        ['echo foo > foo', 'exit 1', 'echo bar > bar']
```
maybe let's adjust it for the list?

ClementWalter: I have moved the command logging to the `cmd_run` function and updated the `_cmd` arg of the errors

dvc/stage/run.py
```diff
@@ -110,11 +114,8 @@ def run_stage(stage, dry=False, force=False, checkpoint_func=None, **kwargs):
 
     callback_str = "callback " if stage.is_callback else ""
     logger.info(
-        "Running %s" "stage '%s' with command:",
-        callback_str,
-        stage.addressing,
+        "Running %s" "stage '%s':", callback_str, stage.addressing,
```
jorgeorpinel, line 117: Hi! Why no longer "with command" ? Is there a full example of what the output of run will be for both a single command stage and a multi-line? Thanks!

jorgeorpinel: And this code is also used by repro right? (Or does it need similar update there?)

ClementWalter: Actually I have moved the "command" to a new logging line in `cmd_run` (line 82) so as to be able to log the current command. This code is also used by repro indeed.

efiop (commented):
Looks good! We are lacking a simple func or unit test to make sure that we are actually able to run list of commands though. Could you add one, please? Ir could be a unit test for `run_cmd` or it could be a simple func test (adding a simple `repo.repro` test to `tests/func/test_repro_multistage.py` would be enough).

dvc/stage/run.py
```diff
@@ -46,13 +48,14 @@ def warn_if_fish(executable):
 @unlocked_repo
 def cmd_run(stage, *args, checkpoint_func=None, **kwargs):
     kwargs = {"cwd": stage.wdir, "env": fix_env(None), "close_fds": True}
+    cmd = stage.cmd if isinstance(stage.cmd, list) else [stage.cmd]
```
skshetry, line 51: The issue you linked to is also about multiline string.
Right now, it's not officially supported, but works depending on the shell.
What do you think of it? Should we support it?

dvc/stage/run.py
```diff
@@ -70,32 +73,33 @@ def cmd_run(stage, *args, checkpoint_func=None, **kwargs):
...
-        cmd = _nix_cmd(executable, stage.cmd)
 
     main_thread = isinstance(
         threading.current_thread(),
         threading._MainThread,  # pylint: disable=protected-access
     )
-    old_handler = None
-    p = None
+    for _cmd in cmd:
+        logger.info("\trunning %s", _cmd)
```
skshetry, line 82: Let's not add `running` on each command of the list as there's already a line for the whole stage.
Let's either print them all as a list before running or one-by-one:
```console
Running stage "build":
	- echo hello
	- echo world
```

But I can see that the output gets messy if the script output gets mixed.
We could think of removing indent and add a ">" before the script:
```console
Running stage "build":
> echo hello
hello
> echo world
world
```

skshetry (commented):
Looks good overall. I have 2 questions regarding UI and multiline support.

dvc/stage/run.py
jorgeorpinel: But it doesn't seem to output "with command(s):" anymore (per https://github.com/iterative/dvc/pull/4934#discussion_r533048696 below) — that's what I'm referring to. It would be great if you can provide a multi-cmd dvc.yaml sample + its repro output, at least for the docs review 🙏. That will make things easier 🙂

dvc/stage/run.py
ClementWalter: I am not sure about your question: the issue mentions:
 - piping: this shout stay as is as stated by @efiop 
 - `&&`: the purpose of this pr: officially allowing the user to define sequential commands in a single stage from its yaml definition.

I don't see much space to think about something else

dvc/stage/run.py
ClementWalter: I like the suggestion with `>`, accepted ;)

dvc/stage/run.py
skshetry: @ClementWalter, as you are proposing to close the issue related to multiline/multiple commands, I just wanted to know what your opinion is on that, and if we should support it, as this list and the multiline string seems to be allowing the user the same thing.

ClementWalter: I think I don't understand what you are trying to ask me

ClementWalter: ok you mean @skshetry if all this work is worth doing because eventually it is only syntactic sugar for yaml?

dvc/stage/run.py
ClementWalter: It is not clear to me where I should add this example

ClementWalter: Resolving here because I have added tests with multiple commands; in the dvc.doc I will add some examples

ClementWalter:
> Looks good! We are lacking a simple func or unit test to make sure that we are actually able to run list of commands though. Could you add one, please? Ir could be a unit test for `run_cmd` or it could be a simple func test (adding a simple `repo.repro` test to `tests/func/test_repro_multistage.py` would be enough).

Just added two tests @efiop tell me what do you think about them

dvc/stage/run.py
```diff
@@ -70,32 +73,33 @@ def cmd_run(stage, *args, checkpoint_func=None, **kwargs):
...
 
-        killed = threading.Event()
-        with checkpoint_monitor(stage, checkpoint_func, p, killed):
-            p.communicate()
-    finally:
-        if old_handler:
-            signal.signal(signal.SIGINT, old_handler)
+            killed = threading.Event()
+            with checkpoint_monitor(stage, checkpoint_func, p, killed):
+                p.communicate()
```
efiop, line 93: @pmrowla Could you please confirm that this will work as expected with checkpoints?

tests/func/test_repro_multistage.py
```diff
@@ -518,3 +515,44 @@ def test_repro_multiple_params(tmp_dir, dvc):
...
+    assert lines == ["First command", "Second command"]
+
+
+def test_repro_list_of_commands_raise_and_stops_after_failure(tmp_dir, dvc):
+    dvcfile_content = {
+        "stages": {
+            "multi_commands": {
+                "cmd": [
+                    """echo "First command" > foo""",
+                    """ehco "Second command" >> foo""",
```
efiop, line 545: Lol, took me a second to realize what's going on :smile: Could make this a bit more obvious, but up to you.

tests/func/test_repro_multistage.py
```diff
@@ -518,3 +515,44 @@ def test_repro_multiple_params(tmp_dir, dvc):
...
+            }
+        }
+    }
+    dump_yaml(PIPELINE_FILE, dvcfile_content)
+
+    with pytest.raises(ReproductionError):
+        dvc.reproduce(target="multi_commands")
+    with open(Path(tmp_dir) / "foo", "r") as foo:
+        lines = foo.read().splitlines()
+    assert lines == ["First command"]
```
efiop, line 558:
```suggestion
    (tmp_dir / "foo").read_text() == "First command"
```
?

tests/func/test_repro_multistage.py
```diff
@@ -518,3 +515,44 @@ def test_repro_multiple_params(tmp_dir, dvc):
...
+                ],
+                "outs": ["foo"],
+            }
+        }
+    }
+    dump_yaml(PIPELINE_FILE, dvcfile_content)
+    dvc.reproduce(target="multi_commands")
+    with open(Path(tmp_dir) / "foo", "r") as foo:
+        lines = foo.read().splitlines()
+    assert lines == ["First command", "Second command"]
```
efiop, line 536:
```suggestion
    assert (tmp_dir / "foo").read_text() == "First command\nSecond command"
```

tests/func/test_repro_multistage.py
ClementWalter: changes `"""` to `'` hopefully it is clearer

tests/func/test_repro_multistage.py
ClementWalter: 👍  I didn't know this command

dvc/stage/run.py
ClementWalter: ok I think I got it, just re-reading from the beginning. You mean: should we also support the parallel case, ie multi line string where every line should a command that runs in parallel?

dvc/stage/run.py
pmrowla: This should work as expected

dvc/stage/run.py
efiop: @ClementwWalter I believe @skshetry was wondering about your opinion on whether or not we should treat multiline string as a list of commands when running it. https://github.com/iterative/dvc/issues/4373#issuecomment-728359883 I've mentioned that it should probably stay as it is, but there is an idea that maybe we should handle that better, similar as gha yaml does it. Let's keep that issue opened for that last question.

ClementWalter: @efiop yes! actually re-reading this morning and re-reading yaml parsing rules it became 100% clear to me what was @skshetry's point, sorry about that yesterday. My guess is that we should not try to parse user input. Maybe we could add a flag in the stage definition about the desired behavior with a list of command: stop on fail or keep going?

tests/func/test_repro_multistage.py
```diff
@@ -517,37 +517,36 @@ def test_repro_multiple_params(tmp_dir, dvc):
...
-                    'echo "Second command" >> foo',
-                ],
-                "outs": ["foo"],
-            }
-        }
-    }
-    dump_yaml(PIPELINE_FILE, dvcfile_content)
-    dvc.reproduce(target="multi_commands")
-    assert (tmp_dir / "foo").read_text() == "First command\nSecond command\n"
+    (tmp_dir / "dvc.yaml").write_text(
```
ClementWalter, line 520: cool, I was not sure if it was better to write yaml directly or to use the `dump_yaml` func indeed.

ClementWalter: should the `PIPELINE_FILE` variable be used instead of direct `dvc.yaml` ?

tests/func/test_repro_multistage.py
```diff
@@ -517,37 +517,36 @@ def test_repro_multiple_params(tmp_dir, dvc):
...
+        dedent(
+            """\
+            stages:
+              multi:
+                cmd:
+                - echo foo>foo
+                - echo bar>bar
```
ClementWalter, line 527: much better because actually my version crashed on windows, thanks

tests/func/test_repro_multistage.py
efiop: Doesn't really matter in the test.

efiop:
Thank you @ClementWalter ! :rocket:

ClementWalter:
thanks you too! how is it that not all tests passes?

efiop:
@ClementWalter Temporary issues with windows tests related to gitpython, please don't mind it, your tests have passed. :+1:

PR #4938, opened 2020-11-23
diff: add support for --targets

### LINKED ISSUE #3388: diff: support targets
```sh
dvc diff -t models HEAD^ HEAD
```
```sh
ERROR: unrecognized arguments: -t HEAD
```

Ubuntu 18.04
Dvc 0.84.0

sandeepmistry:

***

Will close #3388.

The filter code needs to be improved (something smarter than starts with) and I think an error needs to be outputted if the target parameter is not found in the list of tracked files or folders. The documentation will also need to be updated accordingly once this is merged.

For now I've started with the original behaviour the issue started with (optional `--target` arg). Let me know if the behaviour from comment https://github.com/iterative/dvc/issues/3388#issuecomment-590692837 would be preferred (`dvc diff [-h] [-q | -v] [-t TARGET] [--show-json] [--show-hash] [--show-md] [--hide-missing] [a_rev] [b_rev] [targets [targets]]`).

### DIFF
--- a/dvc/command/diff.py
+++ b/dvc/command/diff.py
@@ -174,9 +174,8 @@ def add_parser(subparsers, parent_parser):
         "--targets",
         nargs="*",
         help=(
-            "Source paths to a data files or directories. Default None. "
-            "If not specified, compares all files and directories `). "
-            "that are under DVC control in the current working space."
+            "Limit command scope to these tracked files or directories. "
+            "Accepts one or more file paths."
         ),
     ).complete = completion.FILE
     diff_parser.add_argument(
--- a/dvc/command/diff.py
+++ b/dvc/command/diff.py
@@ -127,7 +127,9 @@ class CmdDiff(CmdBase):
 
     def run(self):
         try:
-            diff = self.repo.diff(self.args.a_rev, self.args.b_rev)
+            diff = self.repo.diff(
+                self.args.a_rev, self.args.b_rev, self.args.target
+            )
             show_hash = self.args.show_hash
             hide_missing = self.args.b_rev or self.args.hide_missing
             if hide_missing:
@@ -166,6 +168,15 @@ def add_parser(subparsers, parent_parser):
         description=append_doc_link(DIFF_DESCRIPTION, "diff"),
         help=DIFF_DESCRIPTION,
         formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+    diff_parser.add_argument(
+        "-t",
+        "--target",
+        help=(
+            "Source path to a data file or directory. Default None. "
+            "If not specified, compares all files and directories "
+            "that are under DVC control in the current working space."
+        ),
     )
     diff_parser.add_argument(
         "a_rev",
--- a/dvc/command/diff.py
+++ b/dvc/command/diff.py
@@ -5,6 +5,7 @@ import os
 
 import colorama
 
+from dvc.command import completion
 from dvc.command.base import CmdBase, append_doc_link
 from dvc.exceptions import DvcException
 
@@ -128,7 +129,7 @@ class CmdDiff(CmdBase):
     def run(self):
         try:
             diff = self.repo.diff(
-                self.args.a_rev, self.args.b_rev, self.args.target
+                self.args.a_rev, self.args.b_rev, self.args.targets
             )
             show_hash = self.args.show_hash
             hide_missing = self.args.b_rev or self.args.hide_missing
@@ -170,14 +171,14 @@ def add_parser(subparsers, parent_parser):
         formatter_class=argparse.RawDescriptionHelpFormatter,
     )
     diff_parser.add_argument(
-        "-t",
-        "--target",
+        "--targets",
+        nargs="*",
         help=(
-            "Source path to a data file or directory. Default None. "
-            "If not specified, compares all files and directories "
+            "Source paths to a data files or directories. Default None. "
+            "If not specified, compares all files and directories `). "
             "that are under DVC control in the current working space."
         ),
-    )
+    ).complete = completion.FILE
     diff_parser.add_argument(
         "a_rev",
         help="Old Git commit to compare (defaults to HEAD)",
--- a/dvc/path_info.py
+++ b/dvc/path_info.py
@@ -19,7 +19,23 @@ class _BasePath:
         return self.isin_or_eq(other) or other.isin(self)
 
     def isin_or_eq(self, other):
+        if isinstance(other, list):
+            others = other
+
+            for other in others:
+                if self.isin_or_eq(other):
+                    return True
+
+            return False
+
         return self == other or self.isin(other)  # pylint: disable=no-member
+
+    def isinside(self, others):
+        for other in others:
+            if other.isin(self):
+                return True
+
+        return False
 
 
 class PathInfo(pathlib.PurePath, _BasePath):
--- a/dvc/path_info.py
+++ b/dvc/path_info.py
@@ -19,23 +19,7 @@ class _BasePath:
         return self.isin_or_eq(other) or other.isin(self)
 
     def isin_or_eq(self, other):
-        if isinstance(other, list):
-            others = other
-
-            for other in others:
-                if self.isin_or_eq(other):
-                    return True
-
-            return False
-
         return self == other or self.isin(other)  # pylint: disable=no-member
-
-    def isinside(self, others):
-        for other in others:
-            if other.isin(self):
-                return True
-
-        return False
 
 
 class PathInfo(pathlib.PurePath, _BasePath):
--- a/dvc/repo/diff.py
+++ b/dvc/repo/diff.py
@@ -118,16 +118,16 @@ def _output_paths(repo, targets):
     for stage in repo.stages:
         for output in stage.outs:
             if _exists(output):
-                if targets is None or any(
-                    output.path_info == target for target in targets
-                ):
+                yield_output = targets is None or any(
+                    output.path_info.isin_or_eq(target) for target in targets
+                )
+
+                if yield_output:
                     yield _to_path(output), _to_checksum(output)
 
-                    if output.is_dir_checksum:
-                        yield from _dir_output_paths(repo_tree, output)
-
-                elif output.is_dir_checksum and any(
-                    target.isin(output.path_info) for target in targets
+                if output.is_dir_checksum and (
+                    yield_output
+                    or any(target.isin(output.path_info) for target in targets)
                 ):
                     yield from _dir_output_paths(repo_tree, output, targets)
 
--- a/dvc/repo/diff.py
+++ b/dvc/repo/diff.py
@@ -42,14 +42,10 @@ def diff(self, a_rev="HEAD", b_rev=None, targets=None):
 
     if targets is not None:
         # check for overlapping missing targets between a_rev and b_rev
-        missing_targets = list(
-            set(missing_targets[a_rev]) & set(missing_targets[b_rev])
-        )
-
-        if len(missing_targets):
-            # invalid targets supplied, raise error
-            for missing_target in missing_targets:
-                raise PathMissingError(missing_target, self)
+        for target in set(missing_targets[a_rev]) & set(
+            missing_targets[b_rev]
+        ):
+            raise PathMissingError(target, self)
 
     old = results[a_rev]
     new = results[b_rev]
@@ -101,7 +97,6 @@ def _paths_checksums(repo, targets):
 def _output_paths(repo, targets):
     repo_tree = RepoTree(repo, stream=True)
     on_working_tree = isinstance(repo.tree, LocalTree)
-    no_targets = targets is None
 
     def _exists(output):
         if on_working_tree:
@@ -123,14 +118,16 @@ def _output_paths(repo, targets):
     for stage in repo.stages:
         for output in stage.outs:
             if _exists(output):
-                yield_output = no_targets or output.path_info.isin_or_eq(
-                    targets
-                )
-                if yield_output:
+                if targets is None or any(
+                    output.path_info == target for target in targets
+                ):
                     yield _to_path(output), _to_checksum(output)
 
-                if output.is_dir_checksum and (
-                    yield_output or output.path_info.isinside(targets)
+                    if output.is_dir_checksum:
+                        yield from _dir_output_paths(repo_tree, output)
+
+                elif output.is_dir_checksum and any(
+                    target.isin(output.path_info) for target in targets
                 ):
                     yield from _dir_output_paths(repo_tree, output, targets)
 
@@ -140,7 +137,9 @@ def _dir_output_paths(repo_tree, output, targets=None):
 
     try:
         for fname in repo_tree.walk_files(output.path_info):
-            if targets is None or fname.isin_or_eq(targets):
+            if targets is None or any(
+                fname.isin_or_eq(target) for target in targets
+            ):
                 yield str(fname), repo_tree.get_file_hash(fname).value
     except NoRemoteError:
         logger.warning("dir cache entry for '%s' is missing", output)
--- a/dvc/repo/diff.py
+++ b/dvc/repo/diff.py
@@ -9,7 +9,7 @@ logger = logging.getLogger(__name__)
 
 
 @locked
-def diff(self, a_rev="HEAD", b_rev=None):
+def diff(self, a_rev="HEAD", b_rev=None, target=None):
     """
     By default, it compares the workspace with the last commit's tree.
 
@@ -28,7 +28,7 @@ def diff(self, a_rev="HEAD", b_rev=None):
             # brancher always returns workspace, but we only need to compute
             # workspace paths/checksums if b_rev was None
             continue
-        results[rev] = _paths_checksums(self)
+        results[rev] = _paths_checksums(self, target)
 
     old = results[a_rev]
     new = results[b_rev]
@@ -62,7 +62,7 @@ def diff(self, a_rev="HEAD", b_rev=None):
     return ret if any(ret.values()) else {}
 
 
-def _paths_checksums(repo):
+def _paths_checksums(repo, target):
     """
     A dictionary of checksums addressed by relpaths collected from
     the current tree outputs.
@@ -74,10 +74,10 @@ def _paths_checksums(repo):
         file:      "data"
     """
 
-    return dict(_output_paths(repo))
+    return dict(_output_paths(repo, target))
 
 
-def _output_paths(repo):
+def _output_paths(repo, target):
     repo_tree = RepoTree(repo, stream=True)
     on_working_tree = isinstance(repo.tree, LocalTree)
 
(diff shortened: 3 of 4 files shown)


dvc/command/diff.py
```diff
@@ -167,6 +169,15 @@ def add_parser(subparsers, parent_parser):
         help=DIFF_DESCRIPTION,
         formatter_class=argparse.RawDescriptionHelpFormatter,
     )
+    diff_parser.add_argument(
+        "-t",
+        "--target",
```
efiop, line 174: How about `--targets`, since we have it in `dvc plots/metrics` already?

dvc/repo/diff.py
```diff
@@ -101,6 +101,9 @@ def _to_checksum(output):
     for stage in repo.stages:
         for output in stage.outs:
             if _exists(output):
+                # if a target was supplied, filter out non-target files
+                if target is not None and not str(output).startswith(target):
```
efiop, line 105: startswith is susceptible to a bug when you don't take into account the path separator. Also `str(output)` is not the best way to get an output path. Please check out `output.path_info`.

dvc/repo/diff.py
```diff
@@ -101,6 +101,9 @@ def _to_checksum(output):
...
+                    continue
                 yield _to_path(output), _to_checksum(output)
                 if output.is_dir_checksum:
                     yield from _dir_output_paths(repo_tree, output)
```
efiop, line 109: what if `target` points to the insides of a directory?

dvc/command/diff.py
sandeepmistry: @efiop makes sense, incorporated in fa6c3b4d3868b240c12200c8dc8082e1bfc4ab3b.

dvc/repo/diff.py
sandeepmistry: Addressed in @efiop makes sense, incorporated in fa6c3b4d3868b240c12200c8dc8082e1bfc4ab3b.

dvc/repo/diff.py
```diff
@@ -124,3 +129,33 @@ def _filter_missing(repo, paths):
             out = metadata.outs[0]
             if out.status().get(str(out)) == "not in cache":
                 yield path
+
+
+def _filter_diff_targets(repo, targets, old, new):
```
sandeepmistry, line 134: @efiop do you think this filtering code belongs here or would it be better in `dvc/command/diff.py`?

sandeepmistry:
@efiop please take a look at the latest changes when you get a chance.

If things look good I will proceed to added tests and updating the documentation.

dvc/repo/diff.py
pmrowla: Being able to filter diffs via the internal API may be useful in the future, and is not really a UI specific issue, I think it makes sense to have the filtering code here in `repo/diff.py`.

Is there a particular reason you moved this code out into a separate function, rather than filtering by targets in `_paths_checksums()` like your original implementation? In DVC we may have repos containing millions of files that we need to iterate through in `diff()`. For performance reasons it's probably better for us to only do that iteration once, and filter out by target as we go in `_paths_checksums()`. But with this filtering code called after `_paths_checksums()`, we will end up iterating over the entire list of paths in the repo a second time.

sandeepmistry: I moved it out of `_paths_checksums()` to report errors on any targets that do not exists (this matches `git diff` behaviour).

If this is not desirable let me know and I can move things back.

pmrowla: So something like
```python
repo_tree = RepoTree(repo, stream=True)
missing_targets = {target for target in targets if not repo_tree.exists(target)}
```
could be used inside the `brancher` loop where we call `_paths_checksums(self)` to accomplish the same thing with much less performance overhead. At that point you only need to error out if there is any overlap between the missing target sets for both revisions after they've been retrieved.

sandeepmistry: @pmrowla thanks for reviewing and the feedback!

I've incorporated your suggestions in https://github.com/iterative/dvc/pull/4938/commits/132636fc0198a5ef52b5c5e1922e81e9d7c951af.

dvc/path_info.py
```diff
@@ -18,8 +18,24 @@ def overlaps(self, other):
         return self.isin_or_eq(other) or other.isin(self)
 
     def isin_or_eq(self, other):
+        if isinstance(other, list):
```
pmrowla, line 21: This function should really be left alone, it's meant to specifically compare two `BasePath` instances.

dvc/path_info.py
```diff
@@ -18,8 +18,24 @@ def overlaps(self, other):
...
+            return False
+
         return self == other or self.isin(other)  # pylint: disable=no-member
 
+    def isinside(self, others):
+        for other in others:
+            if other.isin(self):
+                return True
+
+        return False
```
pmrowla, line 37: Likewise, this is not necessary, it does the same thing as `isin`, and there's cleaner ways to handle comparing against a list of `others`.

dvc/path_info.py
pmrowla: You can accomplish this when you actually want to do the comparison with expressions like

```python
if any((self.isin_or_eq(other) for other in others)):
    ...
```

dvc/repo/diff.py
```diff
@@ -101,17 +123,25 @@ def _to_checksum(output):
     for stage in repo.stages:
         for output in stage.outs:
             if _exists(output):
-                yield _to_path(output), _to_checksum(output)
-                if output.is_dir_checksum:
-                    yield from _dir_output_paths(repo_tree, output)
+                yield_output = no_targets or output.path_info.isin_or_eq(
+                    targets
+                )
```
pmrowla, line 128: As mentioned above, this can be done using `any()` without requiring the change to `isin_or_eq`.

dvc/repo/diff.py
```diff
@@ -101,17 +123,25 @@ def _to_checksum(output):
...
+                if yield_output:
+                    yield _to_path(output), _to_checksum(output)
+
+                if output.is_dir_checksum and (
+                    yield_output or output.path_info.isinside(targets)
+                ):
+                    yield from _dir_output_paths(repo_tree, output, targets)
```
pmrowla, line 135: `isin()` will not cover every case here, I think we need to use `path_info.overlaps(...)` instead.

Let's say, `output` is a tracked directory `dir/`, which contains a file `foo`. `output.path_info` here will contain the path to the tracked dir: `dir/`.

If our target is `dir/foo`, `output.path_info.isin_or_eq(target)` will be False. In this case the output `dir/` itself should not be yielded (which you have correctly accounted for already), but the file `dir/foo` within `output` does need to be yielded (meaning we still have to call `_dir_output_paths(...)`

dvc/repo/diff.py
```diff
@@ -23,12 +24,32 @@ def diff(self, a_rev="HEAD", b_rev=None):
...
+    if targets is not None:
+        # check for overlapping missing targets between a_rev and b_rev
+        missing_targets = list(
+            set(missing_targets[a_rev]) & set(missing_targets[b_rev])
+        )
+
+        if len(missing_targets):
+            # invalid targets supplied, raise error
+            for missing_target in missing_targets:
+                raise PathMissingError(missing_target, self)
```
pmrowla, line 52: This can be simplified

```suggestion
        for target in set(missing_targets[a_rev]) & set(missing_targets[b_rev]):
            raise PathMissingError(target, self)
```

pmrowla (changes requested):
@sandeepmistry Nice progress on this PR! There's still a couple minor issues that I think need to be cleaned up, but for the most part it looks good to me.

dvc/path_info.py
sandeepmistry: Incorporated in https://github.com/iterative/dvc/pull/4938/commits/395167309234a0fb80cd66a416d3b52cc43f2e1e.

dvc/path_info.py
sandeepmistry: Incorporated in https://github.com/iterative/dvc/pull/4938/commits/395167309234a0fb80cd66a416d3b52cc43f2e1e.

dvc/repo/diff.py
sandeepmistry: Incorporated in https://github.com/iterative/dvc/pull/4938/commits/395167309234a0fb80cd66a416d3b52cc43f2e1e.

dvc/repo/diff.py
sandeepmistry: Incorporated in 3951673.

sandeepmistry:
@pmrowla thanks for the another round of suggestions and feedback - I think I've addressed everything in the last round.

I've opened PR for the documentation updates for this as well https://github.com/iterative/dvc.org/pull/2002.

dvc/repo/diff.py
```diff
@@ -101,17 +118,29 @@ def _to_checksum(output):
     for stage in repo.stages:
         for output in stage.outs:
             if _exists(output):
-                yield _to_path(output), _to_checksum(output)
-                if output.is_dir_checksum:
-                    yield from _dir_output_paths(repo_tree, output)
+                if targets is None or any(
+                    output.path_info == target for target in targets
```
pmrowla, line 122: Seems like this should be `output.path_info.isin_or_eq(target)` and not `==`?

If I have
```
.
└── dir
    └── output
```
and `targets=["dir"]`, this won't match `output`, since `output.path_info` will be `dir/output`

pmrowla (changes requested):

dvc/repo/diff.py
sandeepmistry: So that case will be covered by the `elif` path below:
```python
                elif output.is_dir_checksum and any(
                     target.isin(output.path_info) for target in targets
                 ):
                     yield from _dir_output_paths(repo_tree, output, targets)
```

I decided to have two paths because for the `== and output.is_dir_checksum` path `_dir_output_paths` can be called without any `targets` to filter because we want all files in that folder for the diff.

pmrowla: What about the case where `output` is a file output and not a directory output (where `output.is_dir_checksum` is false)?

sandeepmistry: Please disregard my previous comment, for now.

sandeepmistry: re: https://github.com/iterative/dvc/pull/4938#discussion_r535952997

I had missed the `dvc add dir/output` vs `dvc add dir` case in my original tests in https://github.com/iterative/dvc/pull/4938/commits/4f3a99a85ab98a81b8af9279fab768bb452bcc78. With your explanation now understand some of the simplifications made in https://github.com/iterative/dvc/pull/4938/commits/395167309234a0fb80cd66a416d3b52cc43f2e1e do not cover this any more.

I've added this scenarios to the tests in https://github.com/iterative/dvc/pull/4938/commits/4b4f04b49099ae24f48d248a7a787540efd36ab8 along with going back to the logic from https://github.com/iterative/dvc/pull/4938/commits/2b3e888758f08ea2464eede67481c62c3a8d346d but using `any` like you suggested.

tests/func/test_diff.py
```diff
@@ -264,3 +264,127 @@ def test_no_commits(tmp_dir):
     assert Git().no_commits
 
     assert Repo.init().diff() == {}
+
+
+def test_targets(tmp_dir, scm, dvc):
```
pmrowla, line 269: Can this maybe be broken up into separate tests? 150 lines for a single test is pretty long, and it will be easier to maintain `diff` in general if we can see that we are specifically failing when "`target` is inside a dir out", or "dir out is inside `target`", etc., just by looking at a test case name. As opposed to seeing "we fail inside one of 7 or 8 possible situations when `target` is set".

tests/func/test_diff.py
```diff
@@ -264,3 +264,148 @@ def test_no_commits(tmp_dir):
...
+    }
+
+    assert dvc.diff(targets=["dir"]) == expected_result
+    assert dvc.diff(targets=["dir" + os.path.sep]) == expected_result
+
+
+def test_targets_single_file_in_dir(tmp_dir, scm, dvc):
+    setup_targets_test(tmp_dir)
+
+    dvc.diff(targets=["dir" + os.path.sep])
```
sandeepmistry, line 329: @efiop @pmrowla There's some weirdness here if this line is removed, an error is thrown:

```
E       dvc.config.NoRemoteError: config file error: no remote specified. Create a default remote with
E           dvc remote add -d <remote name> <remote url>
```

I'm not sure if `setup_targets_test` is causing this or if an actual issue in dvc.

tests/func/test_diff.py
```diff
@@ -264,3 +264,148 @@ def test_no_commits(tmp_dir):
...
+        "deleted": [],
+        "modified": [],
+        "not in cache": [],
+    }
+
+
+def test_targets_two_files_in_dir(tmp_dir, scm, dvc):
+    setup_targets_test(tmp_dir)
+
+    dvc.diff(targets=["dir" + os.path.sep])
```
sandeepmistry, line 342: https://github.com/iterative/dvc/pull/4938/files#r540622220 ... same thing here.

tests/func/test_diff.py
sandeepmistry: @pmrowla understood, I've attempted to make this change in https://github.com/iterative/dvc/pull/4938/commits/480d951e308a2ba4a9788842b0d0a1180b47d224. Let me know if this was what you expected.

tests/func/test_diff.py
pmrowla: It's because you are deleting the cache directory in `setup_targets_test`. We delete the cache directory in some of the existing diff tests because we are specifically testing for the `cache missing` diff state there. For these `targets` related tests you should not need to delete cache or unlink state in your setup function.

sandeepmistry: Thanks! Updated in https://github.com/iterative/dvc/pull/4938/commits/1b6ee154c09370f4cd16aa2b58c5a457850a3259.

dvc/command/diff.py
```diff
@@ -167,6 +170,15 @@ def add_parser(subparsers, parent_parser):
...
         formatter_class=argparse.RawDescriptionHelpFormatter,
     )
+    diff_parser.add_argument(
+        "--targets",
+        nargs="*",
+        help=(
+            "Source paths to a data files or directories. Default None. "
+            "If not specified, compares all files and directories `). "
+            "that are under DVC control in the current working space."
+        ),
```
jorgeorpinel, line 180: How about "Limit command scope to these tracked files or directories. Accepts one or more file paths."

jorgeorpinel:
Should this PR be linked to #3388 ?

Also could the `[targets]` argument be implicit (after the revs) instead of a `--targets` flag? Per https://github.com/iterative/dvc/issues/3388#issuecomment-590658614 and https://github.com/iterative/dvc/issues/3388#issuecomment-590692837

Thanks

efiop:
> Should this PR be linked to #3388 ?

@jorgeorpinel Good point.

> Also could the [targets] argument be implicit (after the revs) instead of a --targets flag? Per #3388 (comment) and #3388 (comment)

No, we can't do it right now. --targets is consistent with dvc metrics/plots diff.

pmrowla:
The changes look good to me, but CLI `dvc diff` errors out with any arguments now.

The error is also happening in master, so it doesn't look like a problem with these diff changes. Appears to be related to #5035? @skshetry 

using diff in `iterative/example-get-started` master: 
```
❯ dvc diff -v                                                                                                             ⏎
2020-12-12 14:48:41,069 DEBUG: Check for update is enabled.
2020-12-12 14:48:41,084 DEBUG:     3.84 ms in resolving values
2020-12-12 14:48:41,135 DEBUG:     3.00 ms in resolving values
2020-12-12 14:48:41,469 ERROR: unexpected error - 'NoneType' object has no attribute 'parts'
  File "/Users/pmrowla/git/dvc/dvc/main.py", line 90, in main
    ret = cmd.run()
  File "/Users/pmrowla/git/dvc/dvc/command/diff.py", line 130, in run
    diff = self.repo.diff(self.args.a_rev, self.args.b_rev)
  File "/Users/pmrowla/git/dvc/dvc/repo/__init__.py", line 53, in wrapper
  File "/Users/pmrowla/git/dvc/dvc/repo/diff.py", line 31, in diff
    results[rev] = _paths_checksums(self)
  File "/Users/pmrowla/git/dvc/dvc/repo/diff.py", line 77, in _paths_checksums
    return dict(_output_paths(repo))
  File "/Users/pmrowla/git/dvc/dvc/repo/diff.py", line 106, in _output_paths
    yield from _dir_output_paths(repo_tree, output)
  File "/Users/pmrowla/git/dvc/dvc/repo/diff.py", line 113, in _dir_output_paths
    for fname in repo_tree.walk_files(output.path_info):
  File "/Users/pmrowla/git/dvc/dvc/tree/repo.py", line 362, in walk_files
    for root, _, files in self.walk(top, **kwargs):
  File "/Users/pmrowla/git/dvc/dvc/tree/repo.py", line 329, in walk
    if not self.exists(top):
  File "/Users/pmrowla/git/dvc/dvc/tree/repo.py", line 166, in exists
    meta = dvc_tree.metadata(path)
  File "/Users/pmrowla/git/dvc/dvc/tree/dvc.py", line 266, in metadata
    outs = self._find_outs(path_info, strict=False, recursive=True)
  File "/Users/pmrowla/git/dvc/dvc/tree/dvc.py", line 42, in _find_outs
    outs = self.repo.find_outs_by_path(path, *args, **kwargs)
  File "/Users/pmrowla/git/dvc/dvc/repo/__init__.py", line 384, in find_outs_by_path
    outs = outs or self.outs_graph
  File "/Users/pmrowla/.virtualenvs/dvc/lib/python3.8/site-packages/funcy/objects.py", line 28, in __get__
  File "/Users/pmrowla/git/dvc/dvc/repo/__init__.py", line 362, in outs_graph
    return build_outs_graph(self.graph, self.outs_trie)
  File "/Users/pmrowla/git/dvc/dvc/repo/graph.py", line 139, in build_outs_graph
    dep_key = dep.path_info.parts
AttributeError: 'NoneType' object has no attribute 'parts'
2020-12-12 14:48:41,992 DEBUG: Version info for developers:
DVC version: 1.11.0+96a97e
```

skshetry:
>   File "/Users/pmrowla/git/dvc/dvc/repo/graph.py", line 139, in build_outs_graph
>     dep_key = dep.path_info.parts
> AttributeError: 'NoneType' object has no attribute 'parts'

This issue is from a `RepoDependency` having a path info as `None`:
https://github.com/iterative/dvc/blob/b0b04d44b1082bfe0ad172f455b50849f6030e21/dvc/dependency/repo.py#L25

The fix could be to either make `Repo.find_outs_by_path` use `repo.graph` or make `outs_trie`/`outs_graph` skip `RepoDependency` (cc @efiop WDYT?)

skshetry:
@Mergifyio rebase

pmrowla (approved):

dvc/command/diff.py
efiop, line 180: Per https://github.com/iterative/dvc/pull/4938/files#r541138957
```suggestion
        help=(
            "Limit command scope to these tracked files or directories. "
            "Accepts one or more file paths."
        ),
```

efiop (approved):

efiop:
Thank you @sandeepmistry ! :pray:

dvinit (commented):
-ignore comment

Issue #4697, opened 2020-10-11
Add type-hints


lafrenierejm:
There are lots of type hints throughout dvc's codebase, but I couldn't find any mention of which type checker the hints are expected to pass against.  I would propose mentioning it in the README and running the checker as part of the test suite.

skshetry:
@lafrenierejm, we don't really use/have type-hints. We'd love to start using it, but so far, we haven't been able to get to it. The type-checks you are seeing are mostly for "documentation/autocompletion", we do not enforce it (yet).

lafrenierejm:
> We'd love to start using [a type checker], but so far, we 
> haven't been able to get to it.

Are there any preferences about which type checker(s) is used? 
mypy [0] is the reference implementation, but there are also FOSS 
tools from Google, Facebook, and Microsoft that all promise better 
performance than mypy and various levels of strictness

[0]: http://mypy-lang.org/

--
Joseph LaFreniere

skshetry:
@lafrenierejm, I'd say mypy, but don't really have any favourites. If it's easy to start with, let's say, pytype, we can totally go with that. I assume, it won't be difficult to migrate to "stricter" tools after we have considerable amounts of type-hints, compared to starting-out completely blank.

@lafrenierejm, are you interested in contributing?

lafrenierejm:
> @lafrenierejm, are you interested in contributing?

Yes, I am.

> @lafrenierejm, I'd say mypy, but don't really have any favourites. If it's easy to start with, let's say, pytype, we can totally go with that.

Sure!  Starting with pytype starts good.

> I assume, it won't be difficult to migrate to "stricter" tools after we have considerable amounts of type-hints, compared to starting-out completely blank.

Correct, it should be fine.

skshetry:
@lafrenierejm, wow, thanks for the interest.

Let's start small, with just simple "not-so strict" config and run those in Github Actions. And, we can expand to other modules thenafter.

If you need any help, feel free to ask us here in the issue/PR or in the [#dev-talk](https://dvc.org/chat) channel on Discord.

skshetry:
Hi @lafrenierejm, I'll be repurposing this issue as for type checks, I guess that's fine.

lafrenierejm:
@skshetry Yes, of course that's fine.  I'm opening a PR now to run pytype during CI.  The build is currently failing the type check, so I'll plan to continue pushing to the branch to resolve the type issues until the PR is suitable to merge.

PR #4944, opened 2020-11-23
types: fix typechecking errors in dvc/repo

skshetry:
Part of #4697.

### DIFF
--- a/dvc/repo/get_url.py
+++ b/dvc/repo/get_url.py
@@ -5,7 +5,6 @@ import dvc.output as output
 from dvc.utils import resolve_output
 
 
-@staticmethod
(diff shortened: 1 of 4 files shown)


dvc/repo/get_url.py
```diff
@@ -5,7 +5,6 @@
 from dvc.utils import resolve_output
 
 
-@staticmethod
```
efiop, line 8: Interesting that this didn't break it.

efiop (approved):

dvc/repo/get_url.py
skshetry: It is legal, mypy just does not like it.

PR #4945, opened 2020-11-23
types: fix typechecking errors around a few random places

skshetry:
Part of #4697 

### DIFF
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -3,6 +3,7 @@ import json
 import logging
 from concurrent.futures import ThreadPoolExecutor
 from copy import copy
+from typing import Optional
 
 from funcy import decorator
 from shortuuid import uuid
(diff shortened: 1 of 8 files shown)


dvc/cache/base.py
```diff
@@ -16,7 +17,9 @@
     DvcException,
 )
 from dvc.progress import Tqdm
-from dvc.remote.slow_link_detection import slow_link_guard
+from dvc.remote.slow_link_detection import (  # type: ignore[attr-defined]
```
skshetry, line 20: Need to refactor this module.

dvc/path_info.py
```diff
@@ -253,8 +253,6 @@ def path(self):
 
 
 class HTTPURLInfo(URLInfo):
-    __hash__ = URLInfo.__hash__
```
skshetry, line 256: Not sure why this was there. Maybe, the `__hash__` are not compatible because of the below overrides?

dvc/stage/serialize.py
```diff
@@ -57,6 +57,7 @@ def _serialize_out(out):
     return out.def_path if not flags else {out.def_path: flags}
 
 
+@no_type_check
```
skshetry, line 60: Type checking it will be hard, at least, I don't want to attempt for now. Doing it will cause cascade of failures In `dvc/stage` module.

dvc/path_info.py
efiop: @skshetry Because we override `__eq__` below?

skshetry: Yup, figured it out. Thanks.

Issue #4935, opened 2020-11-22
exp show: `--exclude` throws AttributeError


skshetry:
### Please provide information about your setup
```
dvc exp show --exclude-metrics train -v
2020-11-22 14:31:39,852 DEBUG: Check for update is enabled.
2020-11-22 14:31:39,880 DEBUG: fetched: [(3,)]                        
2020-11-22 14:31:41,553 DEBUG: fetched: [(12331,)]
2020-11-22 14:31:42,627 ERROR: unexpected error - 'collections.OrderedDict' object has no attribute 'difference_update'
------------------------------------------------------------
Traceback (most recent call last):
  File "/home/saugat/repos/iterative/dvc/dvc/main.py", line 90, in main
    ret = cmd.run()
  File "/home/saugat/repos/iterative/dvc/dvc/command/experiments.py", line 335, in run
    _show_experiments(
  File "/home/saugat/repos/iterative/dvc/dvc/command/experiments.py", line 265, in _show_experiments
    metric_names, param_names = _collect_names(
  File "/home/saugat/repos/iterative/dvc/dvc/command/experiments.py", line 84, in _collect_names
    metric_names = _filter_names(
  File "/home/saugat/repos/iterative/dvc/dvc/command/experiments.py", line 61, in _filter_names
    _filter(exclude, ret.difference_update)
AttributeError: 'collections.OrderedDict' object has no attribute 'difference_update'
------------------------------------------------------------
```

PR #4946, opened 2020-11-23
types: fix typechecking errors (2)

skshetry:
Part of #4697

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -51,13 +51,13 @@ def _filter_names(
                 update_func({match: None for match in matches})
 
     if include:
-        ret = OrderedDict()
+        ret: OrderedDict = OrderedDict()
         _filter(include, ret.update)
     else:
         ret = OrderedDict({name: None for name in names})
 
     if exclude:
-        _filter(exclude, ret.difference_update)
+        _filter(exclude, ret.difference_update)  # type: ignore[attr-defined]
 
     return [".".join(name) for name in ret]
 
--- a/dvc/dependency/base.py
+++ b/dvc/dependency/base.py
@@ -1,3 +1,5 @@
+from typing import Type
+
 from dvc.exceptions import DvcException
 
 
@@ -21,9 +23,11 @@ class DependencyIsStageFileError(DvcException):
 class BaseDependency:
     IS_DEPENDENCY = True
 
-    DoesNotExistError = DependencyDoesNotExistError
-    IsNotFileOrDirError = DependencyIsNotFileOrDirError
-    IsStageFileError = DependencyIsStageFileError
+    DoesNotExistError = DependencyDoesNotExistError  # type: Type[DvcException]
+    IsNotFileOrDirError = (
+        DependencyIsNotFileOrDirError
+    )  # type: Type[DvcException]
+    IsStageFileError = DependencyIsStageFileError  # type: Type[DvcException]
 
     def update(self, rev=None):
         pass
--- a/dvc/output/base.py
+++ b/dvc/output/base.py
@@ -1,6 +1,7 @@
 import logging
 import os
 from copy import copy
+from typing import Type
 from urllib.parse import urlparse
 
 from voluptuous import Any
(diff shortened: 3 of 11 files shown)


dvc/output/gs.py
```diff
@@ -1,7 +1,7 @@
-from dvc.output.s3 import S3Output
+from dvc.output.base import BaseOutput
```
skshetry, line 1: Looks like a mistake.

dvc/dependency/base.py
```diff
@@ -1,19 +1,23 @@
-from dvc.exceptions import DvcException
+from dvc.output.base import (
+    OutputDoesNotExistError,
+    OutputIsNotFileOrDirError,
+    OutputIsStageFileError,
+)
 
 
-class DependencyDoesNotExistError(DvcException):
+class DependencyDoesNotExistError(OutputDoesNotExistError):
```
skshetry, line 8: Need to do this to make `BaseDependency` work. Otherwise, it will complain because `BaseOutput` has different exception type whereas here, it's different.

skshetry: Though, need to figure out how to fix the messages, right now I have broken it.

dvc/command/experiments.py
```diff
@@ -51,13 +51,13 @@ def _filter(filters, update_func):
...
     if include:
-        ret = OrderedDict()
+        ret: OrderedDict = OrderedDict()
         _filter(include, ret.update)
     else:
         ret = OrderedDict({name: None for name in names})
 
     if exclude:
-        _filter(exclude, ret.difference_update)
+        _filter(exclude, ret.difference_update)  # type: ignore[attr-defined]
```
skshetry, line 60: Should be fixed when #4935 is fixed.

dvc/output/base.py
```diff
@@ -21,15 +21,32 @@
 logger = logging.getLogger(__name__)
 
 
-class OutputDoesNotExistError(DvcException):
+def _format_msg(msg, path: "BaseOutput"):
```
skshetry, line 24: @efiop, would love to hear what you think. This one is a major pain here, if it's solved, we can start using `mypy` more.

dvc/output/base.py
```diff
@@ -85,9 +102,9 @@ class BaseOutput:
         },
     )
 
-    DoesNotExistError = OutputDoesNotExistError
-    IsNotFileOrDirError = OutputIsNotFileOrDirError
-    IsStageFileError = OutputIsStageFileError
+    DoesNotExistError = DoesNotExistError
+    IsNotFileOrDirError = IsNotFileOrDirError
+    IsStageFileError = IsStageFileError
```
skshetry, line 107: Will fix these APIs after completing this round.

dvc/output/base.py
skshetry: Though, now these exceptions won't be two different things. :(

dvc/dependency/base.py
```diff
@@ -21,9 +23,11 @@ def __init__(self, path):
 class BaseDependency:
     IS_DEPENDENCY = True
 
-    DoesNotExistError = DependencyDoesNotExistError
-    IsNotFileOrDirError = DependencyIsNotFileOrDirError
-    IsStageFileError = DependencyIsStageFileError
+    DoesNotExistError = DependencyDoesNotExistError  # type: Type[DvcException]
```
skshetry, line 26: Found a way to override subtypes. :)

dvc/command/experiments.py
skshetry, line 60: This should be fixed when #4935 is fixed.

dvc/dependency/base.py
skshetry: So, no need for a hack/refactor for now. cc @efiop.

dvc/output/gs.py
efiop: @skshetry Not sure if this is needed anymore, but in the past this was done because gs shared logic with s3. Since tree refactor it might indeed no longer be needed.

dvc/dependency/base.py
efiop: That's hacky :D But will do until we refactor outs/deps.

efiop (approved):

PR #4947, opened 2020-11-23
types: add types to dvc/tree

skshetry:
Part of #4697

### DIFF
--- a/dvc/tree/base.py
+++ b/dvc/tree/base.py
@@ -2,6 +2,7 @@ import logging
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from functools import partial
 from multiprocessing import cpu_count
+from typing import Any, ClassVar, Dict, Optional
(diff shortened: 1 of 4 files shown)


dvc/tree/base.py
```diff
@@ -46,8 +47,8 @@ def use_state(call):
 
 class BaseTree:
     scheme = "base"
-    REQUIRES = {}
-    PATH_CLS = URLInfo
+    REQUIRES: ClassVar[Dict[str, str]] = {}
+    PATH_CLS = URLInfo  # type: Any
```
skshetry, line 51: Hack, not sure how to fix this. I think the best thing to do is, force types on all classes that override the types. But, it still won't solve our problem with function signature violating LSP. Generics might solve, but the `mypy` does not work on all cases. :(

efiop (approved):

PR #4948, opened 2020-11-23
types: fix type errors on dvc.collect

skshetry:
Part of #4697

### DIFF
--- a/dvc/types.py
+++ b/dvc/types.py
@@ -1,0 +1,10 @@
+from typing import TYPE_CHECKING, Union
+
+from dvc.path_info import PathInfo, URLInfo
+
+if TYPE_CHECKING:
+    from os import PathLike
+
+StrPath = Union[str, "PathLike[str]"]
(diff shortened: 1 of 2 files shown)


dvc/types.py
```diff
@@ -0,0 +1,10 @@
+from typing import TYPE_CHECKING, Union
+
+from dvc.path_info import PathInfo, URLInfo
+
+if TYPE_CHECKING:
+    from os import PathLike
+
+StrPath = Union[str, "PathLike[str]"]
+DvcPath = Union[PathInfo, URLInfo]
```
skshetry, line 9: I am not sure what to call this union. Suggestions would be appreciated.

efiop: @skshetry I guess in an ideal world we would've made URLInfo inherit from PurePath. Could be done.

efiop (approved):

dvc/types.py
skshetry: We should do this soon, as we cannot type check our tree function without this.

PR #4950, opened 2020-11-23
parametrization: prevent dumping parametrized stages

skshetry:
This is a bit hackish, but does the job done.
It feels like we need some unified way to be able to load
and dump, but not sure how. Till then, this data class
will be helpful in communicating between those.

### DIFF
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -2,6 +2,8 @@ import logging
 import os
 import string
 from collections import defaultdict
+from dataclasses import dataclass
+from typing import Optional
 
 from funcy import cached_property, project
 
@@ -58,6 +60,12 @@ def loads_from(cls, repo, path, wdir, data):
         ),
     }
     return cls(**kw)
+
+
+@dataclass
+class Meta:
+    parametrized: bool = False
+    generated_from: Optional[str] = None
 
 
 def create_stage(cls, repo, path, external=False, **kwargs):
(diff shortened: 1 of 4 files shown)


dvc/stage/__init__.py
```diff
@@ -131,6 +139,7 @@ def __init__(
         self._stage_text = stage_text
         self._dvcfile = dvcfile
         self.desc = desc
+        self.meta = Meta()
```
efiop, line 142: This is not the same as `meta` field, right? If so, could we change the name? It is really confusing otherwise.

skshetry: No, we don't keep `meta` around. It's only on `_stage_text`. Will change it to `RawData`/`raw_data`.

dvc/stage/loader.py
```diff
@@ -106,13 +107,23 @@ def __getitem__(self, name):
...
+        stage = self.load_stage(
             self.dvcfile,
             name,
             self.resolved_data[name],
             self.lockfile_data.get(name, {}),
         )
 
+        group, *keys = name.rsplit(JOIN, maxsplit=1)
+        if group and keys and name in self.stages_data:
+            stage.raw_data.generated_from = group
```
pared, line 119: It seems to me that in the test case `build@us` `generated_from` is not set - shouldn't it be set to `build`?

pared (commented):
One question

dvc/stage/loader.py
skshetry: Silly mistake. It should have been
```suggestion
        if group and keys and name not in self.stages_data:
            stage.raw_data.generated_from = group
```

pared (approved):

efiop (approved):

PR #4951, opened 2020-11-24
Add mypy config, fix type errors in a few remaining places

skshetry:
I haven't added it to the CI yet, as it will only confuse the contributors. We should add it after we have significant type coverage.
To run that locally, just run:
```
$ mypy -p dvc
```


dvc/parsing/context.py
```diff
@@ -42,7 +42,6 @@ def __init__(self, key, new, into):
             )
             return
 
-        assert isinstance(new, Node) and isinstance(into[key], Node)
```
skshetry, line 45: There was already a "if" clause checking it's negation. :)

dvc/tree/ssh/connection.py
```diff
@@ -70,16 +70,18 @@ def close(self):
         self._ssh.close()
 
     def st_mode(self, path):
+        lstat = None
         with suppress(FileNotFoundError):
-            return self.sftp.lstat(path).st_mode
```
skshetry, line 74: `mypy` had problem figuring out return type here. Dunno why.

dvc/utils/serialize/_common.py
```diff
@@ -1,34 +1,76 @@
...
-    with open_fn(path, encoding="utf-8") as fd:
+    with open_fn(path, encoding="utf-8") as fd:  # type: ignore
         return parser(fd.read(), path)
 
 
-def _dump_data(path, data, dumper, tree=None):
+def _dump_data(path, data: Any, dumper: DumperFn, tree: "BaseTree" = None):
     open_fn = tree.open if tree else open
-    with open_fn(path, "w+", encoding="utf-8") as fd:
+    with open_fn(path, "w+", encoding="utf-8") as fd:  # type: ignore
```
skshetry, line 61: We need to fix `PathInfo` situation soon, if we need to keep adding more types.

mypy.ini
```diff
@@ -0,0 +1,17 @@
+[mypy]
```
efiop, line 1: Could we add it to pyproject.toml instead? Just to avoid having a yet another config file.

dvc/utils/serialize/_common.py
```diff
@@ -1,34 +1,70 @@
 """Common utilities for serialize."""
 import os
 from contextlib import contextmanager
+from typing import TYPE_CHECKING, Any, Callable, ContextManager, Dict, Union
+
+from typing_extensions import Protocol
```
skshetry, line 6: Need to add this package on conda_forge too.

mypy.ini
skshetry: @efiop, it does not support `pyproject.toml` yet, but we can add it to `setup.cfg`.

efiop (approved):

Issue #4897, opened 2020-11-15
Sharing for experiments and checkpoints


dmpetrov:
We need a way to share experiments between remote executors and a local environment.

Scenarios:
- Run an experiment through CI/CD (without auto-commits) and see the results in a local environment
- Run an experiment in a remote machine (yeah, it is not implemented yet but the experiment architecture should support it).
- Run parallel experiments in a local machine and see all the runs from a single place.

An "experiment" can be:
- regular experiment `dvc exp`
- an experiment with checkpoints. All the checkpoints need to be shared as well.

What is important:
1. **Addressation** - experiments and checkpoints name should be consistent among the remote and local runs.
2. **Names** - it should be a clear connection (ideally 1-1) between experiment names and Git branches in the main repo.
3. **Local exp to branch** - we need a way (command) to convert regular experiments (w/ and w/o checkpoints) to Git branches in the main repo.
4.  **Run-cache exp to branch** - we need a way (command) to convert experiments from run-cache (w/ and w/o checkpoints) to Git branches in the main repo.

EDITS: It extends #4821

pmrowla:
I've been thinking about this a bit more, and I think custom git refs may work for us here rather than storing patches in DVC run-cache or exp-cache.

We've discussed the potential of using a custom git ref namespace for experiments before, but it was tabled since github actions cannot be triggered when pushing a custom (non-branch or tag) ref.

However, in this case even though we have to push an actual branch to trigger a github actions (or gitlab) CI/CD build, I think we can still leverage custom refs as a cleaner way to store experiment patches.

Internally on the DVC side, we will need to support `git push` and `git fetch`'ing from our experiments workspace (`.dvc/experiments`) to and from a custom refs namespace (like `refs/exp/...` or `refs/dvcexp/...`). This would also facilitate sharing experiments between team members even without taking the CI/CD use case into consideration. 

* `dvc exp push` (or maybe `dvc push --exp`) would push all "local" experiment branches to an upstream git remote (plus the run cache for whatever we have already run locally)
* `dvc exp pull` would pull everything from the upstream `refs/exp` into our local experiments workspace (plus the associated run cache).
* We would probably also need commands to list and gc experiments in the upstream git repo

On the CI/CD side, we are still limited by needing to push actual branches or PRs in order to trigger a build, so either

* the user needs to set up an actual repo level github branch with the changes they want to test and push it
* or we trigger it ourselves by pushing a PR branch via something like `dvc exp deploy`

[...]

PR #4952, opened 2020-11-24
git: add support for working with refspecs and stashes

pmrowla:

Prerequisite for #4897.

- Refactors `dvc.scm.git` to support the use of multiple backends
- Adds `GitPythonBackend` and `DulwichBackend`
- Adds support for refspecs and stashes


dvc/scm/git.py
```diff
@@ -89,10 +90,24 @@ def __init__(self, root_dir=os.curdir, search_parent_directories=True):
         self.ignored_paths = []
         self.files_to_track = set()
 
+        self._dulwich_repo = None
+
+    @property
+    def dulwich_repo(self):
```
efiop, line 96: Thank you so much for using dulwich right away! :pray: Is the functionality sufficient?

Also, why not cached_property?

pmrowla: For the refs related features dulwich will work for us. We just have to make sure to do the necessary conversion from strings to bytes when making dulwich calls.

The dulwich stash implementation is mostly incomplete, but the parts of it which are implemented work well enough.

And I used property here because we want to call `_dulwich_repo.close()` in `scm.close()`, but we won't always be instantiating the dulwich repo

pared:
LGTM so far

dvc/scm/backend/gitpython.py
```diff
@@ -0,0 +1,50 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class GitPythonBackend(BaseGitBackend):  # pylint:disable=abstract-method
+    """git-python Git backend."""
+
+    @property
+    def git(self):
+        return self.scm.repo.git
```
pmrowla, line 14: This plus the empty `__init__` are placeholders until all of the current `dvc.scm.git` git-python functionality is moved into this backend (or removed entirely)

dvc/scm/git.py
```diff
@@ -89,10 +95,16 @@ def __init__(self, root_dir=os.curdir, search_parent_directories=True):
         self.ignored_paths = []
         self.files_to_track = set()
 
+        self.backends = [backend(self) for backend in self.DEFAULT_BACKENDS]
```
pmrowla, line 98: Eventually `self.repo` should go away entirely (replaced by the `GitPythonBackend` in `self.backends`)

dvc/scm/git.py
```diff
@@ -491,3 +497,160 @@ def resolve_commit(self, rev):
...
+    _ignored = partialmethod(_backend_func, "_ignored")
+    set_ref = partialmethod(_backend_func, "set_ref")
+    get_ref = partialmethod(_backend_func, "get_ref")
+    remove_ref = partialmethod(_backend_func, "remove_ref")
+    push_refspec = partialmethod(_backend_func, "push_refspec")
+    fetch_refspecs = partialmethod(_backend_func, "fetch_refspecs")
+    _stash_iter = partialmethod(_backend_func, "_stash_iter")
+    _stash_push = partialmethod(_backend_func, "_stash_push")
+    _stash_apply = partialmethod(_backend_func, "_stash_apply")
+    reflog_delete = partialmethod(_backend_func, "reflog_delete")
```
pmrowla, line 519: I originally had this set up using class introspection and `setattr` calls in `__init__()`, but when using runtime instrospection, mypy and pylint will end up complaining about missing attributes any time we try to call a backend function. I'm not sure whether or not there's a proper way to get introspection + type checking to work nicely?

dvc/scm/git.py
```diff
@@ -491,3 +497,160 @@ def resolve_commit(self, rev):
...
+        self,
+        message: Optional[str] = None,
+        include_untracked: Optional[bool] = False,
+    ) -> Optional[str]:
+        if not self.scm.is_dirty(untracked_files=include_untracked):
+            logger.debug("No changes to stash")
+            return None
+
+        logger.debug("Stashing changes in '%s'", self.ref)
+        rev, reset = self.scm._stash_push(  # pylint: disable=protected-access
```
pmrowla, line 620: The backend `_stash...` functions are marked as protected to make it obvious that when using git stash you should be using this higher level `Stash` class and not the low level backend functions.

pmrowla:
@efiop take a look at this PR and see what you think about splitting the git internals into backends this way.

dvc/scm/backend/base.py
```diff
@@ -0,0 +1,106 @@
...
+if TYPE_CHECKING:
+    from dvc.scm.git import Git
+
+
+class NoGitBackendError(SCMError):
+    def __init__(self, func):
+        super().__init__(f"No valid Git backend for '{func}'")
+
+
+class BaseGitBackend:
```
efiop, line 15: Let's use abc right away
```suggestion
class BaseGitBackend(ABC):
```

dvc/scm/backend/base.py
```diff
@@ -0,0 +1,106 @@
...
+    """Base Git backend class."""
+
+    def __init__(self, scm: "Git", **kwargs):
+        self.scm = scm
+        self.root_dir = os.fspath(scm.root_dir)
+
+    def close(self):
+        pass
```
efiop, line 23:
```suggestion
    @abstractmethod
    def close(self):
        pass
```

efiop: and other methods/properties as well.

dvc/scm/git.py
```diff
@@ -491,3 +497,160 @@ def resolve_commit(self, rev):
...
+        rev = self.stash.push(**kwargs)
+        try:
+            yield rev
+        finally:
+            if rev:
+                logger.debug("Restoring stashed workspace")
+                self.stash.pop()
+
+
+class Stash:
```
efiop, line 581: Feels like we could make git.py -> git/__init__.py and have backend in git/backend/* and stash in a git/stash.py.

efiop: Btw, stash is purely git concept, right? Meaning that we don't ever need it for no-scm case and other things like mercurial don't have it? Talking about hg just theoretically, there is no plans to support hg right now, but the arch here assumes that there might be alternative scm implementations.

efiop (commented):
@pmrowla Thanks! This looks really good! Just a few comments. I really like the direction where this is going so far. Potentially we could even have a config option in the future to switch between backends. :nail_care: :nail_care: :nail_care:

dvc/scm/git.py
pmrowla: It's been a while since I used hg for anything, but yeah I think there's no stash equivalent out of the box for hg.

efiop (approved):

PR #4968, opened 2020-11-25
s3: use mocked s3 instead of the real one

skshetry:
Looks like it works in the newer versions of `moto`.
Not removing the fixtures just yet, maybe in the next PR if this gets approved.

### DIFF
--- a/tests/func/test_api.py
+++ b/tests/func/test_api.py
@@ -75,7 +75,7 @@ def test_open(tmp_dir, dvc, remote):
     [
         pytest.lazy_fixture(cloud)
         for cloud in [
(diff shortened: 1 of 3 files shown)


tests/func/test_api.py
```diff
@@ -128,7 +128,7 @@ def test_open_granular(tmp_dir, dvc, remote):
     [
         pytest.lazy_fixture(cloud)
         for cloud in [
-            "real_s3",  # NOTE: moto's s3 fails in some tests
+            "s3",
```
skshetry, line 131: Not sure why it fails on the CI as it is passing locally. From the error, it seems that the response is empty.

PR #4976, opened 2020-11-26
repro: support glob/foreach-group to run at once through CLI

### LINKED ISSUE #4886: api: `Repo.stage` for loading/dumping stages
Right now, this _service_ was provided through `Dvcfile.stages` and `Dvcfile.stage`. The problem with the API is it requires us to create `Dvcfile` each time even if we don't need it. 

We already have wrappers around it due to above reason: 

https://github.com/iterative/dvc/blob/3daeeaffdb43989283719d9cd8a7a4825dda19d5/dvc/repo/__init__.py#L258
https://github.com/iterative/dvc/blob/3daeeaffdb43989283719d9cd8a7a4825dda19d5/dvc/repo/__init__.py#L266

This might get more complicated with group stage run target, i.e. name of the stage group of a foreach.
So, it'd be great to have a simple wrapper that'd look like this:
```python
>>> repo.stage.load_one("build", file="dvc.yaml")
>>> repo.stage.load_all(file="data.csv.dvc")
>>> repo.stage.load(target="", file="")
```

The API might improve, when trying to improve this. That is what I have in mind.

### LINKED ISSUE #4912: repro: glob-style selection of stages
It would be useful to be able to select stages by glob, either at the top level or within a specific `dvc.yaml` file.

For example, if I have `eval/dvc.yaml` that evaluates several models over several data sets, I would like to be able to reproduce the evaluations over one data set with something like:

    dvc repro eval/dvc.yaml:AZ-*

Or:

    dvc repro --stages='AZ-*' eval/dvc.yaml

With the much-improved support for executing unrelated jobs in parallel in recent versions of DVC, this will enable me to submit each data set as a separate job to the cluster's batch scheduler, along with I imagine a host of other use cases.

### LINKED ISSUE #4958: make dvc repro STAGE running all parametrized stages
```yaml
stages:
  build:
    foreach:
      - 15
      - 25
      - 35
    do:
      cmd: echo ${item}
```

```console
jc@mathai3:~/Downloads/dvc_test$ dvc repro build@15
Running callback stage 'build@15' with command:
	echo 15
15
Use `dvc push` to send your updates to remote storage.
```

The following version doesn't work, even though `build` is defined in `stages` section.

```console
jc@mathai3:~/Downloads/dvc_test$ dvc repro build
ERROR: "Stage 'build' not found inside 'dvc.yaml' file"
```

I wish we could make it equivalent to

```console
for item in 15 25 35; do
    dvc repro build@$item
done
```

Furthermore, depends on whether there's a lock, multiple stages can be run in parallel.

skshetry:
This PR adds a way to be able to run the following stage `foreach` group through a single target `build`:
```yaml
stages:
  build:
    foreach:
      - 15
      - 25
      - 35
    do:
      cmd: echo ${item}
```

Eg:
```console
$ dvc repro build
```

Also, this PR adds a `--glob` flag so that you can achieve the same with:
```console
$ dvc repro "build@*" --glob
```

Also note that, globbing is only allowed for the stage name, not to the file.
So, the following would work:
```console
$ dvc repro "dvc.yaml:build*" --glob
```
But, the below one won't:
```console
$ dvc repro "**/dvc.yaml:build*" --glob
```

Regarding the implementation, I have moved all of the target parsing logic inside `dvc.repo.stage` in a `StageLoad` class.
So, enabling these features on a `checkout` or similar commands should be straightforward.

Fixes #4912
Fixes #4886
Fixes #4958

Tests will follow tomorrow.

### DIFF
--- a/dvc/command/repro.py
+++ b/dvc/command/repro.py
@@ -63,6 +63,7 @@ class CmdRepro(CmdBase):
             "recursive": self.args.recursive,
             "force_downstream": self.args.force_downstream,
             "pull": self.args.pull,
+            "glob": self.args.glob,
         }
 
 
@@ -174,6 +175,12 @@ def add_arguments(repro_parser):
             "Try automatically pulling missing cache for outputs restored "
             "from the run-cache."
         ),
+    )
+    repro_parser.add_argument(
+        "--glob",
+        action="store_true",
+        default=False,
+        help="Allows targets containing regex filter to repro.",
     )
 
 
--- a/dvc/command/repro.py
+++ b/dvc/command/repro.py
@@ -180,7 +180,7 @@ def add_arguments(repro_parser):
         "--glob",
         action="store_true",
         default=False,
-        help="Allows targets containing regex filter to repro.",
+        help="Allows targets containing shell-style wildcards.",
     )
 
 
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -302,7 +302,7 @@ class Repo:
         recursive=False,
         graph=None,
         accept_group: bool = False,
-        filter_regex: bool = False,
+        glob: bool = False,
     ):
         if not target:
             return list(graph) if graph else self.stages
@@ -313,7 +313,7 @@ class Repo:
             )
 
         stages = self.stage.from_target(
-            target, accept_group=accept_group, filter_regex=filter_regex
+            target, accept_group=accept_group, glob=glob
         )
         if not with_deps:
             return stages
@@ -335,7 +335,7 @@ class Repo:
         with_deps: bool,
         recursive: bool,
         accept_group: bool,
-        filter_regex: bool,
+        glob: bool,
     ):
         # Optimization: do not collect the graph for a specific target
         file, name = parse_target(target)
@@ -357,10 +357,7 @@ class Repo:
                     )
         elif not with_deps and is_valid_filename(file):
             stages = self.stage.load_all(
-                file,
-                name,
-                accept_group=accept_group,
-                filter_regex=filter_regex,
+                file, name, accept_group=accept_group, glob=glob,
             )
 
         return stages, file, name
@@ -372,7 +369,7 @@ class Repo:
         recursive=False,
         graph=None,
         accept_group: bool = False,
-        filter_regex: bool = False,
+        glob: bool = False,
     ):
         """
         Priority is in the order of following in case of ambiguity:
@@ -385,7 +382,7 @@ class Repo:
             return [(stage, None) for stage in self.stages]
 
         stages, file, _ = self._collect_specific_target(
-            target, with_deps, recursive, accept_group, filter_regex
+            target, with_deps, recursive, accept_group, glob
         )
         if not stages:
             if not (recursive and self.tree.isdir(target)):
@@ -403,7 +400,7 @@ class Repo:
                     recursive,
                     graph,
                     accept_group=accept_group,
-                    filter_regex=filter_regex,
+                    glob=glob,
                 )
             except StageFileDoesNotExistError as exc:
                 # collect() might try to use `target` as a stage name
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -1,6 +1,6 @@
 import logging
 import os
-from contextlib import contextmanager
+from contextlib import contextmanager, suppress
 from functools import wraps
 from typing import TYPE_CHECKING
 
@@ -8,7 +8,7 @@ from funcy import cached_property, cat
 from git import InvalidGitRepositoryError
 
 from dvc.config import Config
-from dvc.dvcfile import PIPELINE_FILE, Dvcfile, is_valid_filename
+from dvc.dvcfile import PIPELINE_FILE, is_valid_filename
 from dvc.exceptions import FileMissingError
 from dvc.exceptions import IsADirectoryError as DvcIsADirectoryError
 from dvc.exceptions import (
@@ -17,6 +17,7 @@ from dvc.exceptions import (
     OutputNotFoundError,
 )
 from dvc.path_info import PathInfo
+from dvc.repo.stage import StageLoad
 from dvc.scm import Base
 from dvc.scm.base import SCMError
 from dvc.tree.repo import RepoTree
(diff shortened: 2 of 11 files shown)


dvc/repo/__init__.py
```diff
@@ -165,6 +166,7 @@ def __init__(
 
         self.cache = Cache(self)
         self.cloud = DataCloud(self)
+        self.stage = StageLoad(self)
```
skshetry, line 169: Suggestion for better naming would be appreciated.

dvc/stage/loader.py
```diff
@@ -133,6 +133,11 @@ def __len__(self):
     def __contains__(self, name):
         return name in self.resolved_data
 
+    def is_foreach_generated(self, name: str):
```
skshetry, line 136: Temporary solution for now, will be fixed during optimization.

dvc/repo/__init__.py
```diff
@@ -339,44 +329,64 @@ def _collect_pipeline(self, stage, graph=None):
...
-
-        # Optimization: do not collect the graph for a specific target
         if not file:
             # parsing is ambiguous when it does not have a colon
             # or if it's not a dvcfile, as it can be a stage name
             # in `dvc.yaml` or, an output in a stage.
             logger.debug(
                 "Checking if stage '%s' is in '%s'", target, PIPELINE_FILE
             )
-            if not (recursive and os.path.isdir(target)):
```
skshetry, line 371: `os.path.isdir` might have been a bug.

skshetry:
Implemented a real globbing, rather than a regex.

dvc/command/repro.py
```diff
@@ -175,6 +176,12 @@ def add_arguments(repro_parser):
             "from the run-cache."
         ),
     )
+    repro_parser.add_argument(
+        "--glob",
+        action="store_true",
+        default=False,
+        help="Allows targets containing regex filter to repro.",
```
skshetry, line 183:
```suggestion
        help="Allows targets containing shell-style wildcards.",
```

efiop (approved):

pared (approved):

dvc/repo/__init__.py
skshetry: @pared, will make it `StageLoader`, and change the other one to `MultiStageLoader`. 

> Maybe stage_load or stage_resolver will be ok, depending on what name we choose in the end.

`stage_load,load_one()` looks a bit odd to me. :) But, I understand your concern.

jorgeorpinel:
> this PR adds a --glob flag

Nice! But the same option name in `add` seems to mean something different? see #4864. We should probably rename one or both of them. Why "globbing" BTW? I'm not sure where this term comes from (I guess also the `glob` Python module) and I'm guessing most users will have no clue. Thanks

johnnychen94:
> Why "globbing" BTW? I'm not sure where this term comes from (I guess also the glob Python module) and I'm guessing most users will have no clue. Thanks

It comes from the UNIX shell `glob` program and is a part of POSIX specification. https://pubs.opengroup.org/onlinepubs/009696899/functions/glob.html

jorgeorpinel:
Thanks Johny. My comment is about the UI. We can't assume users have such a specific UNIX/Linux or Python experience (or a Ph.D 😉). `glob` sounds like "global". `--wildcard` or something like that would seem more intuitive IMO.

skshetry:
@jorgeorpinel, our users must be more familiar with `glob` than someone like us.

shcheklein:
I agree with @jorgeorpinel that it's a technical term being used in UI, which is not ideal. You have to explain it in docs using usual, less technical terms - e.g. path matches pattern. Then why not call it `--name-pattern`, or something like this. Check other alternatives how CLI tools do this (if any). I would though mention that we are going to remove the option I hope? And make it a default behavior? (at least that's the idea for `dvc add` and other commands), not sure about this case.

off-topic @skshetry it reminds me the `yield` discussion that we had internally :) - it's good, technically correct, but not not widely known - thus we have `do` - which it intuitive and friendly. Same here to my mind.

efiop:
@jorgeorpinel @shcheklein term `globbing` is more popular. `wildcard` is just 1 globbing pattern. `--name-pattern` is just odd. I don't see an alternative to `--glob` right now. Please don't focus on this too much, as we've discussed earlier, this option might even go away in the future, but even if not - there doesn't seem to be a better alternative anyway.

shcheklein:
I didn't know the `glob` term. Even today I doubt I would even start looking for a `dvc stage names glob support`. My 2cs - it's fine to keep it if we gonna remove (why don't we remove it for repro right away?). Otherwise I agree with @jorgeorpinel on this - if we can find a human readable option - let's do that.

efiop:
For the record: We've discussed --glob in https://github.com/iterative/dvc/pull/4864#discussion_r521956584 before. Just keeping it as is until we add support for it to more commands and enable it by default.

jorgeorpinel:
> we are going to remove the option I hope? And make it a default behavior?

When will we know? Is there an issue to follow on this? I can create it otherwise. Please lmk

> We've discussed --glob in #4864 (comment) before

We have but specifically for add. My first impression here is that it does something quite different for run (using glob internally, but that's an implementation detail).

> term globbing is more popular

This [trend](https://trends.google.com/trends/explore?cat=5&q=globbing,glob,wildcard) suggest otherwise but it's also not definitive proof. In any case it's also about intuitiveness I think.

> Please don't focus on this too much, as this option might even go away in the future

OK let's move on! But I fear that we may be underestimating important details that may hinder the product in the long-run, and may be very harder to even detect later.

To me the takeaway is whether we consider UI/UX part of the dev process, or whether it's OK that it's secondary. My bias is towards the former because, like Ivan mentioned, it's harder to ignore these confusions when writing/reviewing docs anyway. But I realize that we would need a slightly different workflow so that everyone has time to think and give feedback on these things — not having to discuss after the PR is merged. Food for thought

johnnychen94:
I think it should either be `--glob` or `--regex`. Even with the word `--name-pattern` you'll still need to explain what kind of rule it follows, is it glob rule, PCRE regex, or Python regex; `--name-pattern` doesn't give a hint of what rule I should use from its name.

`--glob-match`, maybe?

efiop:
@jorgeorpinel I understand your perspective, let's indeed try to sync on this. The issue for --glob is https://github.com/iterative/dvc/issues/4816 . We are just waiting for `--glob` to get supported by more commands and to actually test it better before releasing it into the wild.

PR #4977, opened 2020-11-26
New option `jobs` for `dvc import`

### LINKED ISSUE #4838: import: `--jobs` option
`dvc import` Provides an easy way to reuse files or directories tracked in any DVC repository. This external tracked data might be stored in a remote DVC repository. In this situation `--job` which controls the parallelism level for DVC to download data from remote storage would be a useful option.

karajan1001:
Fixes #4838

  iterative/dvc.org/pull/2008

jobs3
It works on my computer now, but some more tests are still needed.

### DIFF
--- a/dvc/command/imp.py
+++ b/dvc/command/imp.py
@@ -19,6 +19,7 @@ class CmdImport(CmdBase):
                 rev=self.args.rev,
                 no_exec=self.args.no_exec,
                 desc=self.args.desc,
+                jobs=self.args.jobs,
             )
         except DvcException:
             logger.exception(
@@ -81,5 +82,16 @@ def add_parser(subparsers, parent_parser):
             "User description of the data (optional). "
             "This doesn't affect any DVC operations."
         ),
+    )
+    import_parser.add_argument(
+        "-j",
+        "--jobs",
+        type=int,
+        help=(
+            "Number of jobs to run simultaneously. "
+            "The default value is 4 * cpu_count(). "
+            "For SSH remotes, the default is 4. "
+        ),
+        metavar="<number>",
     )
     import_parser.set_defaults(func=CmdImport)
--- a/dvc/dependency/repo.py
+++ b/dvc/dependency/repo.py
@@ -68,14 +68,14 @@ class RepoDependency(LocalDependency):
     def dumpd(self):
         return {self.PARAM_PATH: self.def_path, self.PARAM_REPO: self.def_repo}
 
-    def download(self, to):
+    def download(self, to, jobs=None):
         cache = self.repo.cache.local
 
         with self._make_repo(cache_dir=cache.cache_dir) as repo:
             if self.def_repo.get(self.PARAM_REV_LOCK) is None:
                 self.def_repo[self.PARAM_REV_LOCK] = repo.get_rev()
 
-            _, _, cache_infos = repo.fetch_external([self.def_path])
+            _, _, cache_infos = repo.fetch_external([self.def_path], jobs=jobs)
 
         cache.checkout(to.path_info, cache_infos[0])
 
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -55,7 +55,6 @@ def loads_from(cls, repo, path, wdir, data):
                 Stage.PARAM_ALWAYS_CHANGED,
                 Stage.PARAM_MD5,
                 Stage.PARAM_DESC,
-                Stage.PARAM_JOBS,
                 "name",
             ],
         ),
@@ -122,7 +121,6 @@ class Stage(params.StageParams):
         stage_text=None,
         dvcfile=None,
         desc=None,
-        jobs=None,
     ):
         if deps is None:
             deps = []
@@ -141,7 +139,6 @@ class Stage(params.StageParams):
         self._stage_text = stage_text
         self._dvcfile = dvcfile
         self.desc = desc
-        self.jobs = jobs
         self.raw_data = RawData()
 
     @property
@@ -475,7 +472,8 @@ class Stage(params.StageParams):
             self.remove_outs(ignore_remove=False, force=False)
 
         if not self.frozen and self.is_import:
-            sync_import(self, dry, force, jobs=self.jobs)
+            jobs = kwargs.get("jobs", None)
+            sync_import(self, dry, force, jobs)
         elif not self.frozen and self.cmd:
             run_stage(self, dry, force, **kwargs)
         else:
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -472,7 +472,7 @@ class Stage(params.StageParams):
             self.remove_outs(ignore_remove=False, force=False)
 
         if not self.frozen and self.is_import:
-            jobs = kwargs.get("jobs", None)
+            jobs = kwargs.get("jobs")
             sync_import(self, dry, force, jobs)
         elif not self.frozen and self.cmd:
             run_stage(self, dry, force, **kwargs)
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -55,6 +55,7 @@ def loads_from(cls, repo, path, wdir, data):
                 Stage.PARAM_ALWAYS_CHANGED,
                 Stage.PARAM_MD5,
                 Stage.PARAM_DESC,
+                Stage.PARAM_JOBS,
                 "name",
             ],
         ),
@@ -121,6 +122,7 @@ class Stage(params.StageParams):
         stage_text=None,
         dvcfile=None,
         desc=None,
+        jobs=None,
     ):
         if deps is None:
             deps = []
@@ -139,6 +141,7 @@ class Stage(params.StageParams):
         self._stage_text = stage_text
         self._dvcfile = dvcfile
         self.desc = desc
+        self.jobs = jobs
         self.raw_data = RawData()
 
     @property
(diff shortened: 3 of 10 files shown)


dvc/command/imp.py
```diff
@@ -82,4 +83,15 @@ def add_parser(subparsers, parent_parser):
             "This doesn't affect any DVC operations."
         ),
     )
+    import_parser.add_argument(
+        "-j",
+        "--jobs",
+        type=int,
+        help=(
+            "Number of jobs to run simultaneously. "
```
shcheklein, line 91: I know this might come from some other DVC commands, but let's reconsider this message please?

`Number of jobs` is not very informative. Number of parallel connections? Number of download jobs?

efiop: @shcheklein Let's not do that please, it is totally out of scope for this PR.

efiop: Though I do see your point :slightly_frowning_face:

efiop: Could just refer to `dvc pull/fetch/status` here, to smooth this out. E.g. `Please refer to `dvc fetch` help for description` or something like that.

shcheklein: I'm not referring to push/pull or any other commands (we can create a ticket for this if needed). `this message` in my comment was about this specific help message only.

efiop: @shcheklein I understand that. This one was copied over from push/pull/etc, so the questions might arise there as well. If we use `Please refer to dvc fetch ...` here we'll dodge the bullet :smile:

shcheklein: hmm ... I guess, my take on this it's fine to change only here (and later propagate if needed, and if it's needed at all). Also `Please refer to dvc fetch` will force us to kinda go and see `fetch`, and what's there, etc, etc. Also it complicates UX. To be honest, I haven't see that kind of redirects in the help messages.

karajan1001: I agreed that we can change it here, and submit another one for push/pull/etc.

efiop (commented):
Looks good! Please check the unit tests, looks like you need to add the new flag there.

karajan1001:
@efiop 

There are two problem left 

1. `jobs` argument is not stored in the `dvcfile`.  

Now we have ways of setting `jobs` in `dvc import` and `dvc pull` after cloning a project with `dvc import`.
But we can't set it in `dvc repro` if the imported dataset changed.

2. `help` message mentioned above. 

---------

> FAILED tests/func/test_api.py::test_open_external[azure]
Is this fail caused by this PR?

efiop:
> jobs argument is not stored in the dvcfile.

@karajan1001 Great point! For now we could just keep it as is and not include jobs into dvcfile. `jobs` kinda depends on your particular connection to the server, so it doesn't feel right to have it in dvcfile for everyone. Feels like there should be a config-level thing to set it. There is a ticket for it already, IIRC. So I suggest just leaving as is for now. Very good observation though!

> help message mentioned above.

:+1: 

> Is this fail caused by this PR?

No, that's just azure being flaky :( Don't mind it.

dvc/stage/__init__.py
```diff
@@ -55,6 +55,7 @@ def loads_from(cls, repo, path, wdir, data):
                 Stage.PARAM_ALWAYS_CHANGED,
                 Stage.PARAM_MD5,
                 Stage.PARAM_DESC,
+                Stage.PARAM_JOBS,
```
efiop, line 58: Let's not add it, this is too specific for your particular connection and might be undesired for other users.

efiop: Ah, I see that you don't save it, you just utilize the loading this way. Hm, I guess we could keep it as is :+1: Let me see...

dvc/stage/__init__.py
```diff
@@ -472,7 +475,7 @@ def run(
             self.remove_outs(ignore_remove=False, force=False)
 
         if not self.frozen and self.is_import:
-            sync_import(self, dry, force)
+            sync_import(self, dry, force, jobs=self.jobs)
```
efiop, line 478: I think you should be able to avoid setting and using `self.jobs` and `PARAM_JOBS`. Just pass `jobs=...` to `run` in `dvc/repo/imp_url.py`

tests/func/test_import.py
```diff
@@ -330,7 +346,7 @@ def test_pull_no_rev_lock(erepo_dir, tmp_dir, dvc):
 
 def test_import_from_bare_git_repo(
     tmp_dir, make_tmp_dir, erepo_dir, local_cloud
-):
+):  # pylint:disable=unused-argument
```
efiop, line 349: I don't think these changes are needed. You probably were missing some dependencies for pylint or something.

dvc/stage/__init__.py
karajan1001: Yes, without this, I can't create a stage with the `jobs` member variable. Stage params' saving is in `dvc.schema.py`.

Here we have another two ways to pass this new variable:

1. Pass `jobs` through `Stage.run()`, but more ugly.
2. Add a new set or fill method? something like `fill_stage_outputs` and `fill_stage_dependencies`, more specific.

dvc/stage/__init__.py
karajan1001: I had considered this before but `stage.run` didn't have any variable before. This represents that it all depends on member variables not on the running status. Adding a `variable` to it break this, and this would cause some inconsistency of `dvc repo`. 

> Feels like there should be a config-level thing to set it. There is a ticket for it already, IIRC. So I suggest just leaving as is for now. Very good observation though!

Maybe we should abandon `--jobs` and makes config effect?

tests/func/test_import.py
karajan1001: It's annoying, I just don't know how to fix this `pylint` error. I'll try to solve it later.

efiop: @karajan1001 It only shows up locally? Try re-installing `pip install '.[all,tests]'`.

dvc/stage/__init__.py
efiop: @karajan1001 Good point! Important consistency here is fetch/pull and import. `repro` is a bit odd like that with `import`s, where it tries to `run` instead of trying to `pull`. That's something that we are trying to fix with `dvc repro --pull` flag right now that will become default behaviour in the future. So ideally there should indeed be a way of telling `import` that it should load config options from some config section and it should indeed save it in dvcfile. Smth like
```
remote:
  config:
    from-remote: myremote  # handy to not have to remember remote name in the project you are importing from
```
if we can make a generalization like that, then `--jobs` would indeed be no longer needed. This would be a more holistic solution, no doubt about that.

dvc/dependency/repo.py
```diff
@@ -68,14 +68,14 @@ def save(self):
...
-    def download(self, to):
+    def download(self, to, jobs=None):
         cache = self.repo.cache.local
 
         with self._make_repo(cache_dir=cache.cache_dir) as repo:
             if self.def_repo.get(self.PARAM_REV_LOCK) is None:
                 self.def_repo[self.PARAM_REV_LOCK] = repo.get_rev()
 
-            _, _, cache_infos = repo.fetch_external([self.def_path])
+            _, _, cache_infos = repo.fetch_external([self.def_path], jobs=jobs)
```
skshetry, line 78: It wasn't quite apparent how this was being used. Finally, at the 10th/11th level, I found it using `repo.cloud.pull()`. That's quite deep. :(

dvc/stage/__init__.py
```diff
@@ -472,7 +472,8 @@ def run(
             self.remove_outs(ignore_remove=False, force=False)
 
         if not self.frozen and self.is_import:
-            sync_import(self, dry, force)
+            jobs = kwargs.get("jobs", None)
```
skshetry, line 475:
```suggestion
            jobs = kwargs.get("jobs")
```

dvc/stage/params.py
```diff
@@ -12,3 +12,4 @@ class StageParams:
     PARAM_METRICS = "metrics"
     PARAM_PLOTS = "plots"
     PARAM_DESC = "desc"
+    PARAM_JOBS = "jobs"
```
skshetry, line 15: Is this used anywhere?

dvc/dependency/repo.py
karajan1001: Same to me, they need refactoring.

dvc/stage/params.py
karajan1001: It used to be used for argument passing but deprecated now. We can remove it.

skshetry (approved):
Just one suggestion, looks good to me otherwise.

https://github.com/iterative/dvc/pull/4977/files#r535817055

dvc/command/imp.py
jorgeorpinel: In fact per #4838 this is `--jobs` option has a very specific mening? "This external tracked data might be stored in a remote DVC repository. In this situation --job which controls the parallelism level for DVC to download data from remote storage." so seems like it's not really the same as in other commands?

> Please refer to dvc fetch will force us to kinda go and see fetch... I haven't see that kind of redirects in the help messages.

Unrelated, but `dvc get-url -h` used to refer to `import-url` (for the `url` arg. details) I think. Now only `import-url` has the full list of URLs.

dvc/stage/__init__.py
karajan1001: @skshetry . With this change tests on Windows would fail, I have reverted it.

dvc/command/imp.py
karajan1001: @jorgeorpinel `dvc import` clones an external repo, and then pull down the data. So I think it means the same as what in `dvc pull`.

efiop (approved):

efiop:
Thank you, @karajan1001 ! :pray:

dvc/stage/__init__.py
efiop: @karajan1001 Oops, didn't notice this before merging. I think windows tests were failing for an unrelated reason.  We've fixed them yesterday, they were failing because of gitpython.

Issue #3803, opened 2020-05-14
`dvc init` gives : No such file or directory: '.dvc/.gitignore'


courentin:
Hello,

I've encounter the error bellow when re-initilizing a dvc project in my repo.

`dvc init -v` gives:

<details open>
<summary>Error log details</summary>

```
2020-05-14 23:04:22,539 ERROR: unexpected error - [Errno 2] No such file or directory: '.dvc/.gitignore'
------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/corentinhembise/.local/share/virtualenvs/speech-XlnoNrSS/lib/python3.6/site-packages/dvc/main.py", line 48, in main
    ret = cmd.run()
  File "/Users/corentinhembise/.local/share/virtualenvs/speech-XlnoNrSS/lib/python3.6/site-packages/dvc/command/init.py", line 19, in run
    subdir=self.args.subdir,
  File "/Users/corentinhembise/.local/share/virtualenvs/speech-XlnoNrSS/lib/python3.6/site-packages/dvc/repo/__init__.py", line 160, in init
    init(root_dir=root_dir, no_scm=no_scm, force=force, subdir=subdir)
  File "/Users/corentinhembise/.local/share/virtualenvs/speech-XlnoNrSS/lib/python3.6/site-packages/dvc/repo/init.py", line 107, in init
    scm.add([os.path.join(dvc_dir, scm.ignore_file)])
  File "/Users/corentinhembise/.local/share/virtualenvs/speech-XlnoNrSS/lib/python3.6/site-packages/dvc/scm/git/__init__.py", line 236, in add
    self.repo.index.add(paths)
  File "/Users/corentinhembise/.local/share/virtualenvs/speech-XlnoNrSS/lib/python3.6/site-packages/git/index/base.py", line 740, in add
    entries_added.extend(self._entries_for_paths(paths, path_rewriter, fprogress, entries))
  File "/Users/corentinhembise/.local/share/virtualenvs/speech-XlnoNrSS/lib/python3.6/site-packages/git/util.py", line 68, in wrapper
    return func(self, *args, **kwargs)
  File "/Users/corentinhembise/.local/share/virtualenvs/speech-XlnoNrSS/lib/python3.6/site-packages/git/index/util.py", line 91, in set_git_working_dir
    return func(self, *args, **kwargs)
[...]

PR #4978, opened 2020-11-26
dvc: more explict error during init when .dvc is ignored by git

### LINKED ISSUE #3738: dvc init fails when .dvc directory is git-ignored
Tested on Windows (binary & pip install) in Windows Terminal (Powershell) and in Git Bash.

**Issue**
I tried to run "dvc init" in a repository that had an ".*" entry in its .gitignore. With -v flag, it fails with the following message:
```
  File "dvc\main.py", line 49, in main
  File "dvc\command\init.py", line 21, in run
  File "dvc\repo__init.py", line 155, in init
  File "dvc\repo\init.py", line 108, in init
  File "dvc\scm\git__init.py", line 194, in add
  File "site-packages\git\index\base.py", line 740, in add
  File "site-packages\git\util.py", line 68, in wrapper
  File "site-packages\git\index\util.py", line 91, in set_git_working_dir
  File "site-packages\git\index\base.py", line 626, in _entries_for_paths
  File "site-packages\git\index\base.py", line 587, in _store_path
FileNotFoundError: [WinError 2] Das System kann die angegebene Datei nicht finden: '.dvc\.gitignore'
```

**Steps to reproduce:**
1. mkdir example
2. cd example
3. git init
4. echo ".*" >> .gitignore
5. dvc init -v  # this fails!
_this is how I fixed it:_
6. echo "!.dvc" >> .gitignore
7. dvc init -fv

What's confusing is that after step 5, .dvc exists and has a few items in there. .dvc/config is even committed to git. That's why I didn't consider that my .gitignore could have anything to do with it. I only figured out the reason with the help of skshetry in Discord (thanks!). It would have been helpful if there was a better error message, or the documentation would mention that ".dvc" can't be under .gitignore.

dudarev:
Fixes #3738

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ neatlynx/__pycache__
 *.pyc
 .env/
 .env2.7/
+.python-version
 
 .dvc.conf.lock
 .DS_Store
--- a/dvc/repo/init.py
+++ b/dvc/repo/init.py
@@ -7,6 +7,7 @@ from dvc.ignore import init as init_dvcignore
 from dvc.repo import Repo
 from dvc.scm import SCM
 from dvc.scm.base import SCMError
+from dvc.scm.git import Git
 from dvc.utils import relpath
 from dvc.utils.fs import remove
 
@@ -50,6 +51,16 @@ def init(root_dir=os.curdir, no_scm=False, force=False, subdir=False):
             "`--subdir` if initializing inside a subdirectory of a parent SCM "
             "repository.".format(repo=root_dir)
         )
+
+    if isinstance(scm, Git):
+        if scm.is_ignored(dvc_dir):
+            raise InitError(
+                "{dvc_dir} is ignored by Git. \n"
+                "Make sure that it's tracked, "
+                "for example, by adding '!.dvc' to .gitignore.".format(
+                    dvc_dir=dvc_dir
+                )
+            )
 
     if os.path.isdir(dvc_dir):
         if not force:
--- a/dvc/repo/init.py
+++ b/dvc/repo/init.py
@@ -7,7 +7,6 @@ from dvc.ignore import init as init_dvcignore
 from dvc.repo import Repo
 from dvc.scm import SCM
 from dvc.scm.base import SCMError
-from dvc.scm.git import Git
 from dvc.utils import relpath
 from dvc.utils.fs import remove
 
@@ -61,9 +60,7 @@ def init(root_dir=os.curdir, no_scm=False, force=False, subdir=False):
 
     if os.path.isdir(dvc_dir):
         if not force:
-            raise InitError(
-                f"'{relpath(dvc_dir)}' exists. Use `-f` to force."
-            )
+            raise InitError(f"'{relpath(dvc_dir)}' exists. Use `-f` to force.")
 
         remove(dvc_dir)
 
--- a/dvc/repo/init.py
+++ b/dvc/repo/init.py
@@ -46,28 +46,23 @@ def init(root_dir=os.curdir, no_scm=False, force=False, subdir=False):
         scm = SCM(root_dir, search_parent_directories=subdir, no_scm=no_scm)
     except SCMError:
         raise InitError(
-            "{repo} is not tracked by any supported SCM tool (e.g. Git). "
+            f"{root_dir} is not tracked by any supported SCM tool (e.g. Git). "
             "Use `--no-scm` if you don't want to use any SCM or "
             "`--subdir` if initializing inside a subdirectory of a parent SCM "
-            "repository.".format(repo=root_dir)
+            "repository."
         )
 
-    if isinstance(scm, Git):
-        if scm.is_ignored(dvc_dir):
-            raise InitError(
-                "{dvc_dir} is ignored by Git. \n"
-                "Make sure that it's tracked, "
-                "for example, by adding '!.dvc' to .gitignore.".format(
-                    dvc_dir=dvc_dir
-                )
-            )
+    if scm.is_ignored(dvc_dir):
+        raise InitError(
+            f"{dvc_dir} is ignored by your SCM tool. \n"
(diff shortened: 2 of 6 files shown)


dvc/scm/git.py
```diff
@@ -183,7 +183,7 @@ def _get_gitignore(self, path):
 
         return entry, gitignore
 
-    def _ignored(self, path):
+    def is_ignored(self, path):
```
dudarev, line 186: I renamed it because it's used as a public method in the PR.

tests/func/test_init.py
```diff
@@ -111,3 +111,17 @@ def test_gen_dvcignore(tmp_dir):
         "# https://dvc.org/doc/user-guide/dvcignore\n"
     )
     assert text == (tmp_dir / ".dvcignore").read_text()
+
+
+def test_init_when_ignored_by_git(tmp_dir, scm, monkeypatch, caplog):
```
pmrowla, line 116: I don't think the `monkeypatch` fixture is needed for this test

tests/func/test_init.py
```diff
@@ -111,3 +111,17 @@ def test_gen_dvcignore(tmp_dir):
...
+    # https://github.com/iterative/dvc/issues/3738
+    tmp_dir.gen({".gitignore": ".*"})
+    with caplog.at_level(logging.ERROR, logger="dvc"):
+        assert main(["init"]) == 1
```
pmrowla, line 120: It may be better to test `Repo.init()` here rather than using `main` (since this is an issue with the `init` internals, not necessarily with only the command line `dvc init`).

So something like
```python
    with pytest.raises(InitError):
        # test using DvcRepo.init() instead of main()
        ...
```

pmrowla (commented):
Thanks for the PR @dudarev! It looks good so far, just a few small suggestions

.gitignore
```diff
@@ -5,6 +5,7 @@ neatlynx/__pycache__
 *.pyc
 .env/
 .env2.7/
+.python-version
```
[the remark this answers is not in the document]
dudarev: Yes, `.python-version` is used by `pyenv`. It's typical to ignore it in projects intended to be ran in multiple environments:

https://github.com/github/gitignore/blob/master/Python.gitignore#L88

Note that `pyenv` with `pyenv-virtualenv` plugin is just a convenience wrapper on top of virtualenv. This is a very common way to use it, so `virtualenv` is still being used. `.python-version` is a convenient feature that switches local shell env to what's mentioned in that file. It's typically a single line of text there.

tests/func/test_init.py
dudarev: Good catch. Removing it.

dvc/repo/init.py
```diff
@@ -51,6 +52,16 @@ def init(root_dir=os.curdir, no_scm=False, force=False, subdir=False):
...
 
+    if isinstance(scm, Git):
+        if scm.is_ignored(dvc_dir):
+            raise InitError(
+                "{dvc_dir} is ignored by Git. \n"
+                "Make sure that it's tracked, "
+                "for example, by adding '!.dvc' to .gitignore.".format(
+                    dvc_dir=dvc_dir
+                )
+            )
```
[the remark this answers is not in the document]
dudarev: Very good points.

- f-strings are my personal choice too, but in other strings in this file I saw usage of `format`, so I wanted to preserve the style. Based on the comment, I refactor both my addition and two old occurrences to f-strings.
- I've added `is_ignored` to `scm.Base` and modified the error message to say:

> {dvc_dir} is ignored by your SCM tool. 
> Make sure that it's tracked, for example, by adding '!.dvc' to .gitignore.

tests/func/test_scm.py
```diff
@@ -92,8 +92,8 @@ def test_ignored(tmp_dir, scm):
     tmp_dir.gen({"dir1": {"file1.jpg": "cont", "file2.txt": "cont"}})
     tmp_dir.gen({".gitignore": "dir1/*.jpg"})
 
-    assert scm._ignored(os.fspath(tmp_dir / "dir1" / "file1.jpg"))
-    assert not scm._ignored(os.fspath(tmp_dir / "dir1" / "file2.txt"))
+    assert scm.is_ignored(os.fspath(tmp_dir / "dir1" / "file1.jpg"))
+    assert not scm.is_ignored(os.fspath(tmp_dir / "dir1" / "file2.txt"))
```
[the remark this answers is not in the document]
dudarev: Usage of `os.fspath` is removed.

tests/func/test_init.py
dudarev: It may be better to leave testing with `main` and checking `caplog` in this functional test. If only exception is checked when `DvcRepo.init()` is called it may not test the full scenario that happened to users.

In issues

https://github.com/iterative/dvc/issues/3738
https://github.com/iterative/dvc/issues/3803

the users did not see clear message about the issue. Even though the exception when calling the `init` method has that message it could happen in the future that it gets suppressed somehow and is not displayed to users. Testing it with `main` would make it more robust.

pmrowla (approved):

pared (approved):

dvc/repo/init.py
```diff
@@ -45,19 +45,22 @@ def init(root_dir=os.curdir, no_scm=False, force=False, subdir=False):
...
             "`--subdir` if initializing inside a subdirectory of a parent SCM "
-            "repository.".format(repo=root_dir)
+            "repository."
+        )
+
+    if scm.is_ignored(dvc_dir):
+        raise InitError(
+            f"{dvc_dir} is ignored by your SCM tool. \n"
+            "Make sure that it's tracked, "
+            "for example, by adding '!.dvc' to .gitignore."
```
jorgeorpinel, line 58: Thanks fot this update!

BTW what do you guys think about just saying "Git" for 2.x +?

I don't know of any plans to support other SCMs soon anyway... Docs already only ever mention Git.

Issue #4906, opened 2020-11-18
Standardize issue names


MetalBlueberry:
Hi all

I've been recently reading through the issues in this repository.

The use of tags is clear enough and I think that is what allows you to manage the high volume of issues. The problem comes when you want to see if an issue already exists before opening it. 

If you check the current list, you can notice a title pattern like *topic: description" but is not consistent across issues. 

here are some examples

- `dvc dag` not accepting output files/folders 
- [WIP] outs: add desc field 
- [WIP] 'dvc vdir' command for updating large remote datasets 
- Do we need `exec` in git hooks?
- experiments: support branching w/modified params from intermediate checkpoint  
- Run-cache for experiments and checkpoints 
- azure: AzureTree.exists and import-url fix for folders 
- dvc: add --pdb flag  
- Git submodule ignored in the dvc dependency  

I think that we can edit the issue template to recommend the patten "topic: description"

- dag: not accepting output files

You have more experience than me and probably you will have a better idea of how to write issue titles, I just want to start the conversation :)

efiop:
Hi @MetalBlueberry ! Makes total sense! We try to stick to the standard `topic: description`, but looks like not everyone is used to such naming and we used to edit the titles but then just gave up unless the title is too random/confusing. Adding that to the issue template makes a lot of sense to me. Let us know if you are keen on submitting a low-hanger PR for that :wink:

karajan1001:
I wonder what is the difference between a topic and a label? in permission?

efiop:
@karajan1001 Correct, we can assign some default labes to some issue templates (e.g. bug/feature), but that's about it, the user won't be able to set other labels himself unless we decide to add like a bot to do that for him through some commands(e.g. `/label testing`). Issue names help to navigate the repo better too, so those are really complementary to each other. :slightly_smiling_face:

MetalBlueberry:
@efiop I will be glad to open a PR, but first, we need to agree on what "topic" means.

for example, if you find the CRLF issue #4658  What would be the best topic: description? 
- add: incorrect hash calculation
- hash: CRLF is ignored during hash calculation
- add: CRLF and LF line endings cannot be modified after dvc add

If we do not define what a topic is, it will be still hard to find related issues. do you have any idea?

also giving a hint of what the description should be will help. In this case, I think the description for an issue should be the consequence of the issue. in the CRLF case, option 3 will be the best.

karajan1001:
There are at least three kinds of issues.
[...]

PR #4991, opened 2020-11-29
update issue template

MetalBlueberry:
After #4906 discussion, here is the merge request.

This aims to standardize the issue names and bodies.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -3,9 +3,36 @@ name: "\U0001F41B Bug Report"
 about: Create a bug report to help us improve DVC
 ---
 
-## Bug Report
+# Bug Report
 
-### Please provide information about your setup
+## Issue name
+
+Issue names must follow the pattern `command: description` where the command is the dvc command that you are trying to run. The description should describe the consequence of the bug. Example: `repro: doesn't detect input changes`
+
+Delete this section before publish
+
+## Description
+
+A clear and concise description of what the bug is.
+
+### Reproduce
+
+Step list of how to reproduce the bug
+
+1. dvc init
+2. Copy dataset.zip to the directory
+3. dvc add dataset.zip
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -3,9 +3,36 @@ name: "\U0001F41B Bug Report"
...
 
-## Bug Report
+# Bug Report
 
-### Please provide information about your setup
+## Issue name
+
+Issue names must follow the pattern `command: description` where the command is the dvc command that you are trying to run. The description should describe the consequence of the bug. Example: `repro: doesn't detect input changes`
+
+Delete this section before publish
```
pmrowla, line 12: This kind of thing can be done with comments, so the user will see the message when they open the bug report, but don't have to explicitly remove all of the instructions every time they file a report.

the template would just look something like
```
# Bug Report

<!--
Issue names should follow the pattern ...
Example: `repro: doesn't detect input changes`
-->

...
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -3,9 +3,36 @@ name: "\U0001F41B Bug Report"
...
+
+A clear and concise description of what the bug is.
+
+### Reproduce
+
+Step list of how to reproduce the bug
+
+1. dvc init
+2. Copy dataset.zip to the directory
+3. dvc add dataset.zip
```
pared, line 24: maybe lets add "for example"

pared (approved):
Looks good to me!
I just wonder whether users will be reading the template, but I guess we won't know until we try.

karajan1001 (approved):
SGTM.

.github/ISSUE_TEMPLATE/bug_report.md
MetalBlueberry: This is a good idea! I basically added this to information blocks so they do not appear by default.

Issue #4907, opened 2020-11-18
dvc: preserve desc when re-add/import[-url]/run'ing


efiop:
Currently `desc` is not preserved if you decide to update your dataset with `dvc add`. E.g.
```
#!/bin/bash                             
                                        
set -e                                  
set -x                                  
                                        
rm -rf myrepo                           
mkdir myrepo                            
cd myrepo                               
git init                                
dvc init                                
                                        
echo foo > foo                          
dvc add foo --desc 'my data description'
cat foo.dvc                             
# outs:
# - md5: d3b07384d113edec49eaa6238ad5ff00
#   size: 4                              
#   path: foo                            
#   desc: my data description            
                                        
echo foo >> foo                         
dvc add foo                             
cat foo.dvc
# outs:
# - md5: d3b07384d113edec49eaa6238ad5ff00
#   size: 4                              
#   path: foo
```

PR #4994, opened 2020-11-30
stage: move is_cached logic to Stage.can_be_skipped

efiop:
Having that logic in `create_stage` is unnatural and also has some
side-effects for unsuspecting users. For example, it will check the
filesystem for existing dvcfile and will try to load a corresponding old
stage. I've adjusted it with minimal changes to better reflect the logic
behind it.

Pre-requisite for #4907

### DIFF
--- a/dvc/repo/imp_url.py
+++ b/dvc/repo/imp_url.py
@@ -44,7 +44,7 @@ def imp_url(
         erepo=erepo,
     )
 
-    if stage is None:
+    if stage.can_be_skipped:
         return None
 
     if desc:
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -87,18 +87,6 @@ def create_stage(cls, repo, path, external=False, **kwargs):
     check_circular_dependency(stage)
     check_duplicated_arguments(stage)
 
-    if stage and stage.dvcfile.exists():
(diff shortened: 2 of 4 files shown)


dvc/stage/__init__.py
```diff
@@ -87,18 +87,6 @@ def create_stage(cls, repo, path, external=False, **kwargs):
...
-        ignore_run_cache = (
-            not kwargs.get("run_cache", True) or has_persist_outs
-        )
-        if has_persist_outs:
-            logger.warning("Build cache is ignored when persisting outputs.")
-
-        if not ignore_run_cache and stage.can_be_skipped:
-            logger.info("Stage is cached, skipping")
-            return None
-
```
efiop, line 101: Minimal changes to get this out of the way from create_stage. `can_be_skipped` and `is_cached` relations still raise questions. These days the connection to actual run_cache is more apparent, but in this functionality we mixup run_cache and avoiding to write the same contents to the dvcfile. Those need to be separate.

dvc/repo/imp_url.py
```diff
@@ -44,7 +44,7 @@ def imp_url(
         erepo=erepo,
     )
 
-    if stage is None:
+    if stage.can_be_skipped:
```
efiop, line 47: Just a bit more intuitive than the old one.

Issue #2381, opened 2019-08-08
Avoid shell (re-)interpretation in `dvc run`


kenahoo:
When the command for `dvc run` is complicated enough, or needs internal quoting, it can be somewhat difficult to get the quoting correct:

```
# Fails
% dvc run -f thing.dvc -d TGP_Opp_2014.csv Rscript -e 'saveRDS(1:10, "ops-data.rds")'
Running command:
    Rscript -e saveRDS(1:10, "ops-data.rds")
zsh:1: unknown file attribute: 1
ERROR: failed to run command - stage 'thing.dvc' cmd 'Rscript -e saveRDS(1:10, "ops-data.rds")' failed

# Fails
% dvc run -f thing.dvc -d TGP_Opp_2014.csv 'Rscript -e "saveRDS(1:10, 'ops-data.rds')"'
Running command:
	Rscript -e "saveRDS(1:10, ops-data.rds)"
Error in saveRDS(1:10, ops - data.rds) : object 'ops' not found
Execution halted
ERROR: failed to run command - stage 'thing.dvc' cmd 'Rscript -e "saveRDS(1:10, ops-data.rds)"' failed

# Succeeds
% dvc run -f thing.dvc -d TGP_Opp_2014.csv 'Rscript -e "saveRDS(1:10, \"ops-data.rds\")"'

# Succeeds
% dvc run -f thing.dvc -d TGP_Opp_2014.csv 'Rscript -e "saveRDS(1:10, '\''ops-data.rds'\'')"'

# Succeeds
% dvc run -f thing.dvc -d TGP_Opp_2014.csv 'Rscript -e "saveRDS(1:10, '"'"'ops-data.rds'"'"')"'
```

The reason for the failure is that the shell evaluates the command arguments *twice* - once when passing the arguments to `dvc run`, and again when `dvc run` actually executes the command to create the artifact.  Once this is understood, the behavior makes consistent sense, but is still a bit tricky to deal with sometimes.  

(The last example succeeds because of the fun fact that the best way to escape single-quotes in a single-quoted string is to use the pattern`'"'"'`.  The penultimate example is similar, but using `'\''` instead.  Simple backslash quoting of single quotes does not work - see https://stackoverflow.com/questions/1250079/how-to-escape-single-quotes-within-single-quoted-strings)

[...]

PR #4995, opened 2020-11-30
dvc: add --no-shell argument for dvc run

Melevir:
Allows to execute dvc run commands directly, without shell.
It comes in handy when solving commands quotation problems.
Details: #2381

Documentation issue: https://github.com/iterative/dvc.org/issues/1984 (will make pr after merging of this one).

### DIFF
--- a/dvc/command/repro.py
+++ b/dvc/command/repro.py
@@ -64,6 +64,7 @@ class CmdRepro(CmdBase):
             "force_downstream": self.args.force_downstream,
             "pull": self.args.pull,
             "glob": self.args.glob,
+            "use_shell": not self.args.no_shell,
         }
 
 
@@ -181,6 +182,12 @@ def add_arguments(repro_parser):
         action="store_true",
         default=False,
         help="Allows targets containing shell-style wildcards.",
+    )
+    repro_parser.add_argument(
+        "--no-shell",
+        action="store_true",
+        default=False,
+        help="Executes command without shell, directly.",
     )
 
 
--- a/dvc/command/run.py
+++ b/dvc/command/run.py
@@ -255,10 +255,7 @@ def add_parser(subparsers, parent_parser):
         "--no-shell",
         action="store_true",
         default=False,
-        help=(
-            "Executes command without shell, directly "
-            "feeding it to subprocess"
-        ),
+        help="Executes command without shell, directly.",
     )
     run_parser.add_argument(
         "command", nargs=argparse.REMAINDER, help="Command to execute."
--- a/dvc/command/run.py
+++ b/dvc/command/run.py
@@ -58,6 +58,7 @@ class CmdRun(CmdBase):
                 single_stage=self.args.single_stage,
                 external=self.args.external,
                 desc=self.args.desc,
+                use_shell=not self.args.no_shell,
             )
         except DvcException:
             logger.exception("")
@@ -248,6 +249,15 @@ def add_parser(subparsers, parent_parser):
         help=(
             "User description of the stage (optional). "
             "This doesn't affect any DVC operations."
+        ),
+    )
+    run_parser.add_argument(
+        "--no-shell",
+        action="store_true",
+        default=False,
+        help=(
+            "Executes command without shell, directly "
+            "feeding it to subprocess"
         ),
     )
     run_parser.add_argument(
--- a/dvc/stage/cache.py
+++ b/dvc/stage/cache.py
@@ -166,7 +166,7 @@ class StageCache:
         dump_yaml(tmp, cache)
         self.tree.move(PathInfo(tmp), path)
 
-    def restore(self, stage, run_cache=True, pull=False, **kwargs):
+    def restore(self, stage, run_cache=True, pull=False):
         if stage.is_callback or stage.always_changed:
             raise RunCacheNotFoundError(stage)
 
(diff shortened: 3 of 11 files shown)


dvc/stage/cache.py
```diff
@@ -166,7 +166,7 @@ def save(self, stage):
         dump_yaml(tmp, cache)
         self.tree.move(PathInfo(tmp), path)
 
-    def restore(self, stage, run_cache=True, pull=False):
+    def restore(self, stage, run_cache=True, pull=False, **kwargs):
```
efiop, line 169: You could `pop` in `use_shell=kwargs.get("use_shell", True),` above and you won't have to make this change :slightly_smiling_face:

efiop (changes requested):
Hi @Melevir !

Thanks for the PR! :pray:

I'm afraid this PR doesn't have much sense without `repro` handled. We won't be able to merge this as is. :slightly_frowning_face:

Melevir:
Are current changes are ok besides lack of `repro` command update?
I'm new in the project, so I need early review :)

dvc/command/run.py
```diff
@@ -250,6 +251,15 @@ def add_parser(subparsers, parent_parser):
             "This doesn't affect any DVC operations."
         ),
     )
+    run_parser.add_argument(
+        "--no-shell",
+        action="store_true",
+        default=False,
+        help=(
+            "Executes command without shell, directly "
+            "feeding it to subprocess"
```
efiop, line 260: This sounds odd. `subprocess` seems a semi-implementation detail. Maybe there is a better way to describe it?

tests/func/test_run_single_stage.py
```diff
@@ -1032,3 +1032,18 @@ def test_run_force_doesnot_preserve_comments_and_meta(tmp_dir, dvc, run_copy):
 
     assert "comment" not in (tmp_dir / "bar.dvc").read_text()
     assert "meta" not in (tmp_dir / "bar.dvc").read_text()
+
+
+def test_run_noshell_skips_extra_quoting(dvc, mocker):
+    cmd = "Rscript -e 'saveRDS(1:10, \"ops-data.rds\")'"
```
efiop, line 1038: Since this is not running, let's try to keep it as simple as possible, so people are not encouraged to figure out what this does :slightly_smiling_face: e.g.
```suggestion
    cmd = "myscript arg1 arg2"
```

efiop:
> Are current changes are ok besides lack of repro command update?
> I'm new in the project, so I need early review :)

Yep, looks pretty good so far. But repro is a pretty important one. Not sure if you've checked out our get-started guide yet, if you didn't - I highly recommend doing so dvc.org/doc/start :slightly_smiling_face:

tests/func/test_run_single_stage.py
efiop: Or there was some intention behind it?

efiop:
@Melevir Btw, in the PR title you seem to be using `—no-shell` instead of `--no-shell`. Really minor, but just looks odd.

dvc/command/repro.py
```diff
@@ -182,6 +183,12 @@ def add_arguments(repro_parser):
         default=False,
         help="Allows targets containing shell-style wildcards.",
     )
+    repro_parser.add_argument(
+        "--no-shell",
+        action="store_true",
+        default=False,
+        help="Executes command without shell, directly.",
+    )
```
efiop, line 191: This is not needed. Repro should use params recorded in dvcfile. `shell: true/false` should be added to dvcfile (written by `dvc run`) so repro could use it.

skshetry: @efiop, should not this be a runtime option, rather than a dvcfile thing.

efiop: @skshetry No, it should be recorded into dvcfile and then used in `repro`. Otherwise, you pretty much break repro.

Melevir: So, if `dvc run` executed with `--no-shell` option, it should add line to dvcfile and `repro` should catch and use it. 
An I right?

efiop: @Melevir Correct.

efiop: Just a bit more context. You create a pipeline with `dvc run` and expect `dvc repro` to be able to reproduce it (the whole pipeline). So `dvc repro --no-shell` will affect all commands in your pipeline, which is clearly undesired, as that might break some commands. `shell: true/false` is a command-level option, so it should be recorded. Same as in things like GHA, where it is per-command option.

Melevir: Sounds legit. Thanks for explanation.

Melevir: Just to make it clear: we are adding `shell` line on top of file, at the same level as `stages` since it works for all commands. Am I right?
It looks like I should also update schema of the file.

efiop: @Melevir Actually no, `shell/no-shell` is cmd-level, so it shouldn't be applied to all commands/stages. It should be an optional field that complements each `cmd`. E.g.
```
stages:
  features:
    cmd: myfeaturescmd
    ...
  training:
    cmd: mytrainingcmd
    shell: false
    deps: ....
```

efiop:
Closing as stale.

PR #5010, opened 2020-12-01
import: support globbing 

ju0gri:
Rel. #4816

Signed-off-by: Ioana Grigoropol <ioana.grigoropol@gmail.com>

* [X] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here - https://github.com/iterative/dvc.org/pull/1991

### DIFF
--- a/dvc/command/imp.py
+++ b/dvc/command/imp.py
@@ -87,6 +87,6 @@ def add_parser(subparsers, parent_parser):
         "--glob",
         action="store_true",
         default=False,
-        help="Allows importing paths containing shell-style wildcards.",
+        help="Allows import paths containing shell-style wildcards.",
     )
     import_parser.set_defaults(func=CmdImport)
--- a/dvc/command/imp.py
+++ b/dvc/command/imp.py
@@ -19,6 +19,7 @@ class CmdImport(CmdBase):
                 rev=self.args.rev,
                 no_exec=self.args.no_exec,
                 desc=self.args.desc,
+                glob=self.args.glob,
             )
         except DvcException:
             logger.exception(
@@ -81,5 +82,11 @@ def add_parser(subparsers, parent_parser):
             "User description of the data (optional). "
             "This doesn't affect any DVC operations."
         ),
+    )
+    import_parser.add_argument(
+        "--glob",
+        action="store_true",
+        default=False,
+        help="Allows importing paths containing shell-style wildcards.",
     )
     import_parser.set_defaults(func=CmdImport)
--- a/dvc/repo/imp_url.py
+++ b/dvc/repo/imp_url.py
@@ -1,10 +1,10 @@
 import os
 
 from dvc.repo.scm_context import scm_context
-from dvc.utils import relpath, resolve_output, resolve_paths
+from dvc.utils import format_link, relpath, resolve_output, resolve_paths
 from dvc.utils.fs import path_isin
 
-from ..exceptions import OutputDuplicationError
+from ..exceptions import DvcException, OutputDuplicationError
 from . import locked
 
 
@@ -19,10 +19,12 @@ def imp_url(
     frozen=True,
     no_exec=False,
     desc=None,
+    glob=False,
 ):
     from dvc.dvcfile import Dvcfile
     from dvc.stage import Stage, create_stage, restore_meta
 
+    orig_out = out
     out = resolve_output(url, out)
     path, wdir, out = resolve_paths(self, out)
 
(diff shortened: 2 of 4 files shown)


dvc/command/imp.py
```diff
@@ -82,4 +83,10 @@ def add_parser(subparsers, parent_parser):
             "This doesn't affect any DVC operations."
         ),
     )
+    import_parser.add_argument(
+        "--glob",
+        action="store_true",
+        default=False,
+        help="Allows importing paths containing shell-style wildcards.",
```
jorgeorpinel, line 90:
```suggestion
        help="Allows import paths containing shell-style wildcards.",
```

jorgeorpinel: So it only affects the `path` argument right? Not `url`

dvc/repo/imp_url.py
```diff
@@ -34,6 +36,35 @@ def imp_url(
...
+            msg = (
+                "Cannot import multiple files or directories at once"
+                "See {} for more information."
+            ).format(
+                format_link(
+                    "https://dvc.org/doc/user-guide/"
+                    "troubleshooting#import-wildcard"
+                )
+            )
+            raise DvcException(msg)
```
jorgeorpinel, line 61: Hmmm so what's the point of wildcards in `import`? 🙂

jorgeorpinel: p.s. will this affect `import-url`? How about `get` and `get-url`? Thanks. Cc @efiop

dvc/command/imp.py
ju0gri: @jorgeorpinel  yes, that is my understanding.

ju0gri:
Is there a way to re-trigger the tests pipeline?

efiop:
@ju0gri Looks like you need to rebase your patch and fix
```
    def test_import_wildcard(tmp_dir, scm, dvc, erepo_dir):
2020-12-04T16:09:52.9319685Z         with erepo_dir.chdir():
2020-12-04T16:09:52.9320300Z             erepo_dir.dvc_gen("foo", "foo content", commit="create foo")
2020-12-04T16:09:52.9321043Z             erepo_dir.dvc_gen("foo123", "foo 123 content", commit="create foo")
2020-12-04T16:09:52.9321646Z             erepo_dir.dvc_gen(
2020-12-04T16:09:52.9322375Z                 "efoothisistoocomplicatedtotype",
2020-12-04T16:09:52.9323093Z                 "extra foo content",
2020-12-04T16:09:52.9323579Z                 commit="create foo",
2020-12-04T16:09:52.9323978Z             )
2020-12-04T16:09:52.9324295Z     
2020-12-04T16:09:52.9324889Z         stage = dvc.imp(os.fspath(erepo_dir), "efoo*", "efoo_imported", glob=True)
2020-12-04T16:09:52.9325636Z         assert os.path.isfile("efoo_imported")
2020-12-04T16:09:52.9326367Z         assert (tmp_dir / "efoo_imported").read_text() == "extra foo content"
2020-12-04T16:09:52.9327142Z >       assert scm.repo.git.check_ignore("efoo_imported")
2020-12-04T16:09:52.9328110Z E       AttributeError: 'Git' object has no attribute 'repo'
2020-12-04T16:09:52.9328656Z 
2020-12-04T16:09:52.9329247Z /home/runner/work/dvc/dvc/tests/func/test_import.py:58: AttributeError
```
it is one of the recent changes to the upstream, sorry for the inconvenience.

dvc/repo/imp_url.py
efiop: Yeah, in case of `import` it is only useful for local git repos, so more of a synthetic use case. Also the current implementation doesn't work for `import` at all, because for `import` there are no local files that iglob could work on, because the target repo is not even cloned yet. Not sure how the tests work here.

efiop: > p.s. will this affect import-url? How about get and get-url? Thanks. Cc @efiop

Yes it will. No, it won't affect get and get-url, at least right now. In the future get and get-url should support glob too.

ju0gri: > Seems like the current implementation doesn't work for `import` at all, because for `import` there are no local files that iglob could work on, because the target repo is not even cloned yet. Not sure how the tests work here.

I got a bit confused myself when I implemented this, and I digged deeper into how import works, and for this test `test_import_wildcard`, here is my understanding:

- there is a repository `erepo_dir` created locally with 3 files: `foo`, `foo123` and `efoothisistoocomplicatedtotype`; this is created somewhere locally, e.g.: `'/private/var/folders/_3/lmzh2stn56z_459q83p871_h0000gn/T/pytest-of-juogri/pytest-122/erepo0'`
- the command issued for importing `efoo*` as `efoo_imported` 
- the current implementation uses absolute path for the repository to do globbing; feels hacky and only works if the repo exists locally
- resolves the path to `'/private/var/folders/_3/lmzh2stn56z_459q83p871_h0000gn/T/pytest-of-juogri/pytest-122/erepo0/efoo*'`, and then finds `'efoothisistoocomplicatedtotype'`

What would be a less hacky approach here that would work for all import cases ?

efiop: It needs to do virtual glob on the cloned repo, which means that this PR will need to be re-done from scratch. Let's close this PR for now, since the most useful part is not really implemented.

efiop:
Closing due to https://github.com/iterative/dvc/pull/5010#discussion_r548794803

PR #5018, opened 2020-12-03
Enable parametrization by default

skshetry:


dvc/config.py
```diff
@@ -234,7 +234,7 @@ class RelPath(str):
         "row_cleanup_quota": All(Coerce(int), Range(0, 100)),
     },
     # section for experimental features
-    "feature": {Optional("parametrization", default=False): Bool},
+    "feature": {Optional("parametrization", default=True): Bool},
```
skshetry, line 237: keeping it till the official release.

skshetry: Maybe, it'd be better if we allowed anything in here and don't complain.

efiop: Yeah, could allow anything in this section. Someone is definitely using this option with viewer :smile: 

Let's add a comment in the code that this is no longer used. Aslo no reason to set default at all, since it is ignored anyway, right?

efiop (approved):

dvc/config.py
skshetry: @efiop, not really. This is still used, so that you could explicitly disable it.

efiop: @skshetry Could you elaborate on a scenario for that? Debugging?

skshetry: No reason, just wanted to provide a way to disable for preview releases.

skshetry: I'll just remove those logics.

Issue #4578, opened 2020-09-17
Preserve permissions of stored files


mgkwill:
The goal of this feature would be to preserve permission of files added to dvc. In our particular use-case it is useful to add a directory of scripts, binaries and config to dvc. When we pull those files or get them, they have lost original permissions requiring us to restore permission manually or with something like acl.

This feature would add a new attribute in the .dvc file.
eg: filemode
Or something more OS agnostic.

This feature would add a new --track-permissions flag:
dvc add <file/dir> --track-permissions

Later when a user pulls or gets files/dir with dvc the permissions are restored.

If track-permissions is not used, files are treated like they are now.

skshetry:
@mgkwill, it might be tricky for a file inside a DVC-tracked directory (we might need to add to `.dir` file meaning we'll have this information in two places). For a dvc-tracked file, it might be trivial to add the support.

shcheklein:
> we might need to add to .dir file meaning we'll have this information in two places

not exactly related, but it might be a good opportunity to include file sizes finally into dir files?

efiop:
@mgkwill Do you mean simple user execution permission (like git) or something else as well?

efiop:
@skshetry @shcheklein about sizes and metadata, there was an idea of having metadata stored separately from the regular cache files, so it is easier to manage ACL for it without giving access to the data contents. Just for the record.

efiop:
Update: we can't really store perms for files in datasets right now, as `.dir` cache format doesn't currently allow for expansion. We'll be changing that in 2.0 very soon. But we could totally start with something like:
```
$ dvc add binary
...
$ cat binary.dvc
outs:                                  
- md5: 9152ab3f172b3e23fd6b1a3ab0e1d150                    
  path: foo                            
  mode: 0775
```

dudarev:
I am picking up this issue.

dudarev:
@efiop I've created initial draft PR for this issue: https://github.com/iterative/dvc/pull/5036 It's far from ready to review. I'm following your recent work to add size into .dvc files from https://github.com/iterative/dvc/commit/e1b82c5222930c55886ca16a48c3a223d05b4af0

A few questions:

1. Currently .dvc file looks like

```
- md5: d41d8cd98f00b204e9800998ecf8427e
  size: 0
  mode: '0o100644'
  path: binary
```

Is `mode` OK in such Python 3 octal notation, or should `0o` better be removed?

[...]

PR #5036, opened 2020-12-05
[WIP] save permissions in dvc files

dudarev:
This is work in progress for #4578 

Modification of tests/func/test_add.py:test_add to output mode parameter in .dvc file.
Implementation of this test.
Other tests are not validated and not addressed yet.

### DIFF
--- a/dvc/state.py
+++ b/dvc/state.py
@@ -455,7 +455,10 @@ class State(StateBase):  # pylint: disable=too-many-instance-attributes
             return None
 
         self._update_state_record_timestamp_for_inode(actual_inode)
-        return HashInfo("md5", value, size=int(actual_size))
+
+        mode = oct(os.stat(path).st_mode)
+
+        return HashInfo("md5", value, size=int(actual_size), mode=mode)
 
     def save_link(self, path_info):
         """Adds the specified path to the list of links created by dvc. This
--- a/dvc/hash_info.py
+++ b/dvc/hash_info.py
@@ -11,12 +11,14 @@ DirInfo = Dict[str, str]
 class HashInfo:
     PARAM_SIZE = "size"
     PARAM_NFILES = "nfiles"
+    PARAM_MODE = "mode"
 
     name: Optional[str]
     value: Optional[str]
     dir_info: Optional[DirInfo] = field(default=None, compare=False)
     size: Optional[int] = field(default=None, compare=False)
     nfiles: Optional[int] = field(default=None, compare=False)
+    mode: Optional[str] = field(default=None, compare=False)
 
     def __bool__(self):
         return bool(self.value)
@@ -43,6 +45,9 @@ class HashInfo:
             ret[self.PARAM_SIZE] = self.size
         if self.nfiles is not None:
             ret[self.PARAM_NFILES] = self.nfiles
+        if self.mode is not None:
+            ret[self.PARAM_MODE] = self.mode
+
         return ret
 
     @property
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -135,6 +135,7 @@ class Stage(params.StageParams):
         dvcfile=None,
         desc=None,
         meta=None,
+        mode=None,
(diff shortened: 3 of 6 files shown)


shcheklein:
@dudarev could you please summarize what are the next steps here?

dudarev:
@shcheklein the next steps could be:

- make all tests pass
- figure out how to test mode change on checkout (test on Windows either should be skipped or set only allowed permission byte)
- implement chmod on checkout
- fix all tests if they will be broken on the previous step
--- tasks below this line probably can be a separate ticket ---
- think of various error handling scenarios
- implement error handling

shcheklein:
@dudarev sounds about right to me! I guess we can move forward with this and other folks will review and chime in along the way!

dvc/state.py
```diff
@@ -455,7 +455,10 @@ def get(self, path_info):
             return None
 
         self._update_state_record_timestamp_for_inode(actual_inode)
-        return HashInfo("md5", value, size=int(actual_size))
+
+        mode = oct(os.stat(path).st_mode)
```
efiop, line 459: Is full mode intentional? Or did you only mean to include permission bits?

efiop:
@dudarev Hi! Thanks for the PR! 🚀  A few comments:

1) Could you please rename it to reflect the contents?
2) Please add a reference for the ticket, so github could track it.
3) Please install pre-commit hooks as noted in https://dvc.org/doc/user-guide/contributing/core . Currently linter checks fail.

Thanks! 🙂

dvc/state.py
```diff
@@ -455,7 +455,10 @@ def get(self, path_info):
...
+
+        return HashInfo("md5", value, size=int(actual_size), mode=mode)
```
efiop, line 461: mode is not really used by state, so we shouldn't include it here (we can probably do that in Output.save with self.tree.stat(self.path_info).st_mode. Also mode is not really related to HashInfo so it shoudn't be.a part of it, unless there are good reasons to include it. 🙂

dvc/state.py
dudarev: I'm leaning now more towards only saving user executable bit, so we could have a boolean parameter:

```
is_user_executable: false/true
```

dvc/state.py
dudarev: Yes, it makes sense not to include `mode` as part of hash. Should content of output file looks like:

```
outs:
- md5: acbd18db4cc2f85cedef654fccc4a4d8
  size: 3
  path: foo
- is_user_executable: false
```

So there will be two separate outputs. Will it affect anything that uses those files?

Could you provide a link to a place where `outs` are set in the code? I'm having trouble finding it quickly.

dudarev:
@efiop Thank you for your feedback!

1. I renamed the PR to better reflect its content.
2. Also added the reference to the ticket in the description.
3. I installed pre-commit hook when working on the previous PR but this time commited from IDE that did not use it. I'll be more mindful about commit from git cli in the following PRs.

You have a very good point about storing only user execution permission and not having `mode` in `HashInfo`. I asked a few questions to your comments above that could streamline the development. Please have a look.

dvc/state.py
efiop: @dudarev It should be more like:
```
outs:
- md5: acbd18db4cc2f85cedef654fccc4a4d8
  size: 3
  path: foo
  is_user_executable: false
```
Since that relates to the same output. Maybe you are confused about the way that entry is generated, see https://github.com/iterative/dvc/blob/f09c27e74c0a79c30c2c09a6b82201f4930ba63b/dvc/output/base.py#L299 . We can just add `ret[self.PARAM_ISEXEC] = self.isexec` there and parse it on the roundtrip similar to other fields.

> Could you provide a link to a place where outs are set in the code? I'm having trouble finding it quickly.

Sure! https://github.com/iterative/dvc/blob/f09c27e74c0a79c30c2c09a6b82201f4930ba63b/dvc/output/base.py#L289

Issue #5037, opened 2020-12-05
Add ability to skip "dirty" commits during `dvc gc`


courentin:
Following https://github.com/iterative/dvc/issues/3885, I'd like to add the ability to skip commits that generate errors during `dvc gc`.

To sum-up, running `dvc gc` in a codebase that contains a commit where dvc is in a dirty state (eg. the `dvc.yaml` is badly formatted, one of the output is dvcignored, etc.) make it failing.

Thus, if feel assuming that **every** commits are clean and let dvc in a usable state is a bit too optimistic. Except if you have a very rigorous process with dvc, it is almost impossible.

Thus, I'd suggest 2 things.
First adding a log to explain where (which branch/commit) the error has been raised.
Second, adding an option to gc: `dvc gc --skip-failing-collect`. Not sure of the naming but the behaviour would be to skip the collection of branches/commits that produce `DvcException`.

What do think?
I've started an implementation that I'll link to this issue.

skshetry:
@courentin, I think `--ignore-revs`/`--ignore-commits` are more reliable than `--skip-failing-collect` as it is clear on what your intentions are, considering `gc` is a dangerous command.

courentin:
Yes, explicit is better than implicit indeed!

amka66:
I think this is not a feature request but rather a bug.

I would like to reduce the cache as it is eating up my hard drive, but i cannot do it due to one or two broken .yaml files in my git commit history.
@efiop or anyone, please help, what can i do?

Unless there is a reasonable solution, I am stuck we a full hard drive. This requires a quick resolutions. Thanks!

WilliamHarvey97:
I had this issue. Nothing to add here, but this looks quiet serious to me because it makes `dvc gc -A` such a dangerous command.

dberenbaum:
> @courentin, I think `--ignore-revs`/`--ignore-commits` are more reliable than `--skip-failing-collect` (being a conditional flag) as it is clear on what your intentions are, considering `gc` is a dangerous command. What do you think?

In addition, we recently introduced `-n` to various `dvc exp` commands to specify the last n commits. The same option might be useful for `gc` to keep some limited history.

daavoo:
> In addition, we recently introduced -n to various dvc exp commands to specify the last n commits. The same option might be useful for gc to keep some limited history.

https://github.com/iterative/dvc/issues/7695

PR #5038, opened 2020-12-05
Add context about failing branches during error in used_cache()

courentin:

This PR is an implementation of #5037.
It is my first contribution to the dvc project, feel free to challenge it.

### DIFF
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -293,6 +293,7 @@ class Repo:
         jobs=None,
         recursive=False,
         used_run_cache=None,
+        ignore_dvc_exceptions=False,
     ):
         """Get the stages related to the given target and collect
         the `info` of its outputs.
@@ -338,6 +339,10 @@ class Repo:
                     cache.update(used_cache, suffix=suffix)
             except DvcException as e:
                 logger.error(f"Cache for '{branch}' could not be collected")
+                if ignore_dvc_exceptions:
+                    logger.debug(e)
+                    continue
+
                 raise e
 
         if used_run_cache:
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -8,7 +8,7 @@ from funcy import cached_property, cat
 from git import InvalidGitRepositoryError
 
 from dvc.config import Config
-from dvc.exceptions import FileMissingError
+from dvc.exceptions import DvcException, FileMissingError
 from dvc.exceptions import IsADirectoryError as DvcIsADirectoryError
 from dvc.exceptions import NotDvcRepoError, OutputNotFoundError
 from dvc.path_info import PathInfo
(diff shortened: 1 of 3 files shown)


courentin:
[This](https://github.com/iterative/dvc.org/blob/master/content/docs/command-reference/gc.md) needs to be updated

dvc/repo/__init__.py
```diff
@@ -319,22 +320,30 @@ def used_cache(
...
+                    used_cache = stage.get_used_cache(
+                        remote=remote,
+                        force=force,
+                        jobs=jobs,
+                        filter_info=filter_info,
+                    )
+                    cache.update(used_cache, suffix=suffix)
+            except DvcException as e:
+                logger.error(f"Cache for '{branch}' could not be collected")
+                if ignore_dvc_exceptions:
```
efiop, line 342: Seems like using `--force` would be intuitive here.
```suggestion
                if ignore_dvc_exceptions or force:
```

efiop (commented):
Thank you for the PR! :pray: 

Ideally, this should skip particular malformed dvcfiles and not the whole branch. We have similar requests for metrics/params/plots parsing too. But this would be a great start too.

dvc/repo/__init__.py
courentin: Do you suggest using force and ignore_dvc_exceptions or just force?
Indeed, `--force` seems more coherent with options across dvc commands

courentin:
Yes indeed, not sure how to do it however

skshetry:
> Yes indeed, not sure how to do it however

There's not an easy way right now for this. I'd imagine something like the following, as [stage collection error can be handled using a callback](https://github.com/iterative/dvc/blob/a7d665157cd838a9406e30c249065111d93a9e4a/dvc/repo/stage.py#L344):
```python
def func(path, exc):
  print(f"Collecting '{path}' failed and was skipped.")

with repo.handle_error("stage_collection", func):
   stages = repo.stages
```

Something like that should work for the CLI usage, but as an API, it might not as we may have collected the stages before running
`metrics.show()` for example.

@efiop @pmrowla @pared if you have better suggestions.

dvc/repo/__init__.py
efiop: @courentin Seems like `--force` would be enough.

efiop:
@skshetry Maybe just something generic and explicit like a `onerror`(similar to os.walk) for `collect*` methods?

dvc/repo/__init__.py
courentin: Hum, `--force` is already used in the gc command "Force garbage collection - automatically agree to all prompts".

courentin:
Do you mind if after finding a good wording for this option, I try to merge this PR as it is and we can discuss a more general approach in an issue or discussion?

skshetry:
> Do you mind if after finding a good wording for this option, I try to merge this PR as it is and we can discuss a more general approach in an issue or discussion?

We don't expect you to handle this right away. This was just us discussing a long-term solution, sorry for the noise here. 
The PR looks good, I'm fine with either of the `--force` or the `--skip-failing-collect`, as long as they are well documented (just don't like the conditional flag on `gc`, that's all :) ). 

It'd be great if you could respond to https://github.com/iterative/dvc/issues/5037#issuecomment-741709787 and https://github.com/iterative/dvc/issues/5066, that'd help us prioritize (if those features would be helpful for you). Thanks for the contributions, @courentin.

dvc/repo/__init__.py
courentin: @efiop Don't we want to differentiate `--force` where the behaviour is "Force garbage collection - automatically agree to all prompts" and `--skip-failing-collect` so that we do one without the other?

WDYT?

I'd like to merge this PR so that our team can clean their workspace :)

efiop: @courentin Do you have a scenario in which those could be used separately?

courentin: I'd say if we want to ignore failing commits but not agree to all prompts. But to be honest I'm fine with using only `--force`, I let you decide what's best for the product ;)

efiop:
Closing as outdated. @courentin Thank you for your patience 🙏 , we've readjusted error handling for our metrics/plots/etc, and brancher is on our todo list for the near future. So please stay tuned.

PR #5044, opened 2020-12-07
git: use backend-provided trie in GitTree

efiop:
Very rough WIP to see how this goes. On gitlab example `dvc metrics diff`(aka the using GitTree part) upstream dvc takes ~26sec, and now it takes with both dulwich and gitpython backends around 7sec (clearly there are more optimizations to be done here). The main difference is that we build a trie that is then pretty simple to use and also more efficient than just walking through the git obj tree every time. The trie is getting fully built right now, but it could be easily built dynamically too.

### DIFF
--- a/dvc/scm/git/backend/dulwich.py
+++ b/dvc/scm/git/backend/dulwich.py
@@ -1,16 +1,53 @@
 import fnmatch
+import locale
 import logging
 import os
-from io import BytesIO
+import stat
+from io import BytesIO, StringIO
 from typing import Callable, Iterable, Optional, Tuple
 
 from dvc.scm.base import SCMError
-from dvc.tree.base import BaseTree
 from dvc.utils import relpath
 
+from ..objects import GitObject
 from .base import BaseGitBackend
 
 logger = logging.getLogger(__name__)
+
+
+class DulwichObject(GitObject):
+    def __init__(self, repo, name, mode, sha):
+        self.repo = repo
+        self._name = name
+        self._mode = mode
+        self.sha = sha
+
+    def open(self, mode: str = "r", encoding: str = None):
+        if not encoding:
+            encoding = locale.getpreferredencoding(False)
+        # NOTE: we didn't load the object before as Dulwich will also try to
+        # load the contents of it into memory, which will slow down Trie
+        # building considerably.
+        obj = self.repo[self.sha]
+        data = obj.as_raw_string()
+        if mode == "rb":
+            return BytesIO(data)
+        return StringIO(data.decode(encoding))
+
+    @property
+    def name(self) -> str:
+        return self._name
+
+    @property
+    def mode(self) -> int:
+        return self._mode
+
+    def scandir(self) -> Iterable["DulwichObject"]:
+        tree = self.repo[self.sha]
+        for entry in tree.iteritems():  # noqa: B301
+            yield DulwichObject(
+                self.repo, entry.path.decode(), entry.mode, entry.sha
+            )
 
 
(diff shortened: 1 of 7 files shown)


dvc/scm/git/backend/dulwich.py
```diff
@@ -102,8 +169,12 @@ def list_tags(self) -> Iterable[str]:
     def list_all_commits(self) -> Iterable[str]:
         raise NotImplementedError
 
-    def get_tree(self, rev: str, **kwargs) -> BaseTree:
-        raise NotImplementedError
+    def get_trie(self, rev: str, **kwargs) -> DulwichTrie:
+        from dulwich.objectspec import parse_tree
+
+        rev = self.scm.resolve_rev(rev)
```
efiop, line 175: Using `self.scm` is a bad thing (we do it in stash code too). Caused by a lack of implementation for `dulwich`. Will be solved in the future.

dvc/scm/git/backend/gitpython.py
```diff
@@ -1,19 +1,25 @@
...
 from dvc.scm.base import CloneError, RevError, SCMError
-from dvc.tree.base import BaseTree
 from dvc.utils import fix_env, is_binary
 
-from .base import BaseGitBackend
+from .base import BaseGitBackend, BaseGitObject, BaseGitTrie
 
 logger = logging.getLogger(__name__)
 
+GIT_MODE_DIR = 0o40000
```
efiop, line 20: this is just S_ISDIR, need to get rid of it.

dvc/tree/git.py
```diff
@@ -76,96 +47,86 @@ def dvcignore(self):
     def open(
         self, path, mode="r", encoding=None
     ):  # pylint: disable=arguments-differ
-        assert mode in {"r", "rb"}
-        encoding = encoding or "utf-8"
```
efiop, line 80: This wasn't right, we need to use locale to get default encoding, same as we do in our remote fixtures. Windows tests fail because of this now :slightly_frowning_face:

dvc/tree/git.py
```diff
@@ -76,96 +47,90 @@ def dvcignore(self):
...
-        # and `open` with default "r" mode returns str)
-        data = obj.data_stream.read()
-        if mode == "rb":
-            return io.BytesIO(data)
-        return io.StringIO(data.decode(encoding))
+            raise FileNotFoundError(errno.ENOENT, msg, path)
+
+        # NOTE: this is incorrect, we need to use locale to determine default
+        # encoding.
+        encoding = encoding or "utf-8"
```
efiop, line 63: Keeping for now to not mess with tests.

dvc/tree/git.py
```diff
@@ -180,66 +106,42 @@ def walk(
...
-                entry.inode,
-                entry.dev,
-                0,
-                entry.uid,
-                entry.gid,
-                entry.size,
-                # git index has no atime equivalent, use mtime
-                to_ctime(entry.mtime),
-                to_ctime(entry.mtime),
-                to_ctime(entry.ctime),
```
efiop, line 240: This was useless for our purposes, so I've dropped it for simplicity.

efiop:
Ok, tests for multiple backends will be introduced in the followup. There is a lot of stuff wrong with scm tree tests to add backends testing right now. Will add before pygit2 tree.

pared:
Note: researching https://iterative.github.io/dvc-bench/index.html#collect.TraverseGitRepoBench.time_repo_traversing
shows that this is the change responsible for decrease of benchmark execution time from 32 s to 11 s.

PR #5061, opened 2020-12-08
dvc: preserve exec bit for tracked files

dudarev:
For ticket #4578

Based on the discussion in https://github.com/iterative/dvc/pull/5036 I've added `isexec` parameter.

The plan for this PR:

* [x] add `test_add_executable` with an executable file
* [x] make all tests pass
* [x] figure out how to test mode change on checkout
* [x] implement chmod on checkout
* [x] fix all tests if they will be broken on the previous step
* [x] <s>disable this feature on Windows and create a separate ticket to enable it there</s>

------

https://github.com/iterative/dvc.org/pull/2039


dvc/output/base.py
```diff
@@ -135,6 +136,9 @@ def __init__(
         self.path_info = self._parse_path(tree, path)
         if self.use_cache and self.cache is None:
             raise RemoteCacheRequiredError(self.path_info)
+        self.is_user_executable = os.path.isfile(path) and os.access(
+            path, os.X_OK
+        )
```
efiop, line 141: The file might not exist here or might not be in the same state as the rest of the info. In this place we should load the isexec from the info provided, e.g. how `metric` is loaded above.

The actual stat should happen in `save()` method, as pointed out in https://github.com/iterative/dvc/pull/5036#discussion_r537530670 .

dvc/output/base.py
```diff
@@ -328,6 +332,8 @@ def dumpd(self):
         if self.checkpoint:
             ret[self.PARAM_CHECKPOINT] = self.checkpoint
 
+        ret[self.PARAM_IS_USER_EXECUTABLE] = self.is_user_executable
+
```
efiop, line 336: same as in other fields above
```suggestion
        if self.is_user_executable:
            ret[self.PARAM_IS_USER_EXECUTABLE] = self.is_user_executable
        
```

dvc/output/base.py
```diff
@@ -76,6 +76,7 @@ class BaseOutput:
     PARAM_PLOT_HEADER = "header"
     PARAM_PERSIST = "persist"
     PARAM_DESC = "desc"
+    PARAM_IS_USER_EXECUTABLE = "is_user_executable"
```
efiop, line 79: I guess we could even use `exec: true/false`, as it is specified for a particular output and it doesn't seem like we will have another use for `exec` here in the future. Could also use `isexec`. The current `is_user_executable` seems a bit long. WDYT?

efiop (commented):
Thanks for the PR! The plan looks pretty good :slightly_smiling_face:

dvc/output/base.py
skshetry: Also, `is` is very much unnecessary. `exec` or `executable` is fine imo.

dudarev: @efiop @skshetry good suggestion about a shorter variable name. It fits the style of this project well. I hesitated a lot, probably too much, between `exec` and `isexec` and decided to go with `exec`. Unfortunately, as a result, I've got a `pylint` error:

```
W0622: Redefining built-in 'exec' (redefined-builtin)
```

so I've started using `isexec` instead. Please let me know if you have a stronger opinion about using `executable` instead.

dvc/output/base.py
dudarev: Thank you for the clarification. I moved `stat` to `save()`.

dvc/hash_info.py
```diff
@@ -11,6 +11,7 @@
 class HashInfo:
     PARAM_SIZE = "size"
     PARAM_NFILES = "nfiles"
+    PARAMS_TO_IGNORE = ("isexec",)
```
dudarev, line 14: @efiop I'm not sure if this is a good approach and there could be a better way to ignore `isexec` from .dvc file in `HashInfo`. I plan to look at a better way to handle it later but maybe you have some quick suggestions?

efiop: @dudarev Take a look at `loadd_from` in `dvc/output/__init__.py`. You just need to pop the value for isexec from info, same as we do with things like `desc`.

dudarev: Thanks! I've modified the code to pop `isexec` in `loadd_from`.

dvc/output/base.py
```diff
@@ -287,6 +291,9 @@ def save(self):
             return
 
         self.hash_info = self.get_hash()
+        self.isexec = os.path.isfile(self.def_path) and os.access(
+            self.def_path, os.X_OK
+        )
```
efiop, line 296:
```suggestion
        self.isexec = self.tree.isfile(self.path_info) and self.tree.isexec(self.path_info)
```

dvc/stage/utils.py
```diff
@@ -66,6 +66,7 @@ def fill_stage_outputs(stage, **kwargs):
             metric="metrics" in key,
             plot="plots" in key,
             checkpoint="checkpoints" in key,
+            isexec="isexec" in key,
```
efiop, line 69: isexec can't be set through CLI
```suggestion
```

tests/func/test_add.py
```diff
@@ -57,6 +57,21 @@ def test_add(tmp_dir, dvc):
     }
 
 
+def test_add_executable(tmp_dir, dvc):
+    tmp_dir.dvc_gen_exec({"foo": "foo"})
```
efiop, line 61: This is the only place where `dvc_gen_exec` is used, better to just inline it here

tests/func/test_checkout.py
```diff
@@ -56,6 +56,35 @@ def _test_checkout(self):
...
 
+class TestCheckoutExecutable(TestRepro):
+
+    EXECUTABLE = "executable"
+    EXECUTABLE_CONTENT = "executable content"
+
+    def setUp(self):
+        super().setUp()
+
+        self.create_executable(self.EXECUTABLE, self.EXECUTABLE_CONTENT)
```
efiop, line 67: The only place where `create_executable` is used, better to inline it here too :)

Btw, we are migrating from unittest to pytest tests, so it would be great to re-write this one with pytest too.

dvc/cache/base.py
```diff
@@ -436,6 +444,10 @@ def _checkout_file(
         if progress_callback:
             progress_callback(str(path_info))
 
+        if isexec:
+            st = os.stat(path_info.fspath)
+            os.chmod(path_info.fspath, st.st_mode | stat.S_IEXEC)
+
```
efiop, line 450: IIRC chmod won't modify the mtime(we use it to detect changes in the workspace), so there is nothing cache should worry about when doing chmod +x. The chmod part is more related to the output itself, than to the cache (notice how we are just passing through the isexec, but don't really use anything cache-specific). That's why the saving of isexec is not in `cache.save` but is actually in `output.save`. So let's move this to `OutputBase.checkout`, right after `cache.checkout` call. Also note that it is better to use tree.* methods instead of `os.*`, as that helps to keep the abstraction for different types of outputs :slightly_smiling_face:

efiop (commented):
Also need to add PARAM_ISEXEC to https://github.com/iterative/dvc/blob/76c6c5427816b25a81d90f86d3af1cb19779802f/dvc/stage/utils.py#L178 , so it is not taking it into account when computing stage hash. This combined with os.access(X_OK) is resulting in windows failures, so we can't really skip those tests.

dvc/output/base.py
dudarev: Good point! I've done this and also updated `utils:is_exec` function to return boolean by casting to `bool`:

https://github.com/iterative/dvc/blob/167ab79b8029d3272e6aa056c12305b6c9befeee/dvc/utils/__init__.py#L463

dvc/stage/utils.py
dudarev: Removed it

tests/func/test_add.py
dudarev: I inlined `dvc_gen_exec` code here.

tests/func/test_checkout.py
dudarev: Inlined `created_executable` here too.

dvc/cache/base.py
dudarev: Good point! I moved this code to `BaseOutput.checkout` and used `tree.*` methods. There is one minor stylistic issue: all methods in `BaseTree` are using `path_info` as an argument, I've added `BaseTree.isexec(path)` since it was not there and all children classes use `path` argument in that method:

https://github.com/iterative/dvc/blob/2ade7f7ba56c1b76924f9b5b7f379225b89e1914/dvc/tree/base.py#L198

dudarev: I modified isexec to use `path_info` as an argument

dudarev:
`test_add_executable` fails on Windows, because `isexec` is not set on Windows. I need to investigate:

https://github.com/iterative/dvc/blob/2ade7f7ba56c1b76924f9b5b7f379225b89e1914/tests/func/test_add.py#L70

```
E         {'outs': [{'md5': 'acbd18db4cc2f85cedef654fccc4a4d8', 'path': 'foo', 'size': 3}]} != {'outs': [{'isexec': True, 'md5': 'acbd18db4cc2f85cedef654fccc4a4d8', 'path': 'foo', 'size': 3}]}
```

tests/func/test_add.py
```diff
@@ -57,6 +58,27 @@ def test_add(tmp_dir, dvc):
     }
 
 
+def test_add_executable(tmp_dir, dvc):
+
+    paths = tmp_dir.gen({"foo": "foo"}, "")
+    # make files executable
+    for p in paths:
+        st = os.stat(p)
+        os.chmod(p, st.st_mode | stat.S_IEXEC)
```
efiop, line 67:
```suggestion
    tmp_dir.gen("foo", "foo")
    st = os.stat("foo")
    os.chmod("foo", st.st_mode | stat.S_IEXEC)
```

tests/func/test_add.py
```diff
@@ -57,6 +58,27 @@ def test_add(tmp_dir, dvc):
...
+    tmp_dir.dvc_add(paths, commit=None)
```
efiop, line 66:
```suggestion
    dvc.add("foo")
```

tests/func/test_add.py
```diff
@@ -57,6 +58,25 @@ def test_add(tmp_dir, dvc):
     }
 
 
+def test_add_executable(tmp_dir, dvc):
+
```
efiop, line 62:
```suggestion
```

tests/func/test_add.py
```diff
@@ -57,6 +58,24 @@ def test_add(tmp_dir, dvc):
...
+        "outs": [
+            {
+                "md5": "acbd18db4cc2f85cedef654fccc4a4d8",
+                "path": "foo",
+                "size": 3,
+                "isexec": True,
+            }
+        ],
+    }
+
```
efiop, line 77:
```suggestion
    }
    assert os.stat("foo").st_mode & stat.S_IEXEC
```

efiop:
Adjusted the test in https://github.com/iterative/dvc/pull/5061#discussion_r546280722 and discovered that when `dvc add`ing an exec file, after we are done it no longer stays as executable. Likely need to chmod in `OutputBase.commit` after saving to cache.

dvc/output/base.py
```diff
@@ -362,6 +373,10 @@ def checkout(
                 filter_info=filter_info,
                 **kwargs,
             )
+            if self.isexec:
+                st = os.stat(self.path_info.fspath)
```
efiop, line 377: We are using >=3.6, where all os methods are totally happy with Path-like objects.
```suggestion
                st = os.stat(self.path_info)
```

tests/func/test_add.py
```diff
@@ -57,6 +58,25 @@ def test_add(tmp_dir, dvc):
     }
 
 
+def test_add_executable(tmp_dir, dvc):
+    tmp_dir.gen("foo", "foo")
+    st = os.stat("foo")
+    os.chmod("foo", st.st_mode | stat.S_IEXEC)
```
dudarev, line 64: @efiop this test fails on Windows because it's not possible to set `S_IEXEC` bit as mentioned in `os.chmod` docs:

https://docs.python.org/3/library/os.html#os.chmod

> Note Although Windows supports chmod(), you can only set the file’s read-only flag with it (via the stat.S_IWRITE and stat.S_IREAD constants or a corresponding integer value). All other bits are ignored.

Should we just skip this test on Windows?

efiop: @dudarev Sure, but I'm more worried about `dvc add` not preserving exec bit https://github.com/iterative/dvc/pull/5061#issuecomment-748525155 . I've pushed the fix that fixed the `add` and now I see that checkout test broke, it might've been working by accident :slightly_frowning_face:

dvc/output/base.py
```diff
@@ -373,14 +378,12 @@ def checkout(
...
             )
-            if self.isexec:
-                st = os.stat(self.path_info.fspath)
-                os.chmod(self.path_info.fspath, st.st_mode | stat.S_IEXEC)
             return res
         except CheckoutError:
             if allow_missing or self.checkpoint:
                 return None
             raise
+        self.set_exec()
```
dudarev, line 386: @efiop  This line will probably not execute because of `return` on line 381

dvc/output/base.py
```diff
@@ -362,10 +378,12 @@ def checkout(
                 filter_info=filter_info,
                 **kwargs,
             )
+            return res
         except CheckoutError:
             if allow_missing or self.checkpoint:
                 return None
             raise
+        self.set_exec()
```
efiop, line 386: Oops, my fault :)

dvc/output/base.py
efiop: Yep, put it there by accident. Thanks!

tests/func/test_add.py
efiop: Thinking about the windows, tests shouldn't be skipped, but actually we should check that things are working as expected even for windows.

dudarev: In the checkout test I was testing with `os.access(self.EXECUTABLE, os.X_OK)`. Maybe on Windows `os.access` returns True in this case.

For testing scenarios on Windows. Since it's not possible to set `S_IEXEC`, the expected behavior is probably: if .dvc file has `isexec: True` in it, the file still remains not executable on Windows after checkout. Or rather it remains the same as without that `isexec` property.

dvc/cache/base.py
```diff
@@ -420,7 +420,12 @@ def safe_remove(self, path_info, force=False):
         self.tree.remove(path_info)
 
     def _checkout_file(
-        self, path_info, hash_info, force, progress_callback=None, relink=False
+        self,
+        path_info,
+        hash_info,
+        force,
+        progress_callback=None,
+        relink=False,
```
efiop, line 428:
```suggestion
        self, path_info, hash_info, force, progress_callback=None, relink=False
```

dudarev: I think this was done by `black`

dvc/output/base.py
```diff
@@ -287,6 +291,13 @@ def save(self):
             return
 
         self.hash_info = self.get_hash()
+        self.isexec = self.tree.isfile(self.path_info) and self.tree.isexec(
+            self.path_info
+        )
+
+    def set_exec(self):
+        if self.isexec:
```
efiop, line 299:
```suggestion
        if self.isfile() and self.isexec:
```

dvc/output/base.py
```diff
@@ -287,6 +291,13 @@ def save(self):
             return
 
         self.hash_info = self.get_hash()
+        self.isexec = self.tree.isfile(self.path_info) and self.tree.isexec(
+            self.path_info
+        )
```
efiop, line 296:
```suggestion
        self.isexec = self.isfile() and self.tree.isexec(self.path_info)
```

dvc/tree/base.py
```diff
@@ -195,6 +195,15 @@ def isfile(self, path_info):
         """
         return True
 
+    def set_exec(self, path_info):
```
efiop, line 198: Note for us, for the future: time to finally move BaseTree to ABC...

dvc/tree/local.py
```diff
@@ -152,8 +152,12 @@ def remove(self, path_info):
...
         makedirs(path_info, exist_ok=True, mode=self.dir_mode)
 
-    def isexec(self, path):
-        mode = os.stat(path).st_mode
+    def set_exec(self, path_info):
+        mode = self.stat(path_info).st_mode
+        self.chmod(path_info, mode | stat.S_IEXEC)
+
+    def isexec(self, path_info):
+        mode = os.stat(path_info).st_mode
```
efiop, line 160:
```suggestion
        mode = self.stat(path_info).st_mode
```

tests/func/test_add.py
```diff
@@ -57,6 +58,35 @@ def test_add(tmp_dir, dvc):
...
+    dvc.add("foo")
+
+    actual = load_yaml("foo.dvc")
+    expected = {
+        "outs": [
+            {
+                "md5": "acbd18db4cc2f85cedef654fccc4a4d8",
+                "path": "foo",
+                "size": 3,
+                "isexec": True,
```
efiop, line 74: oops
```suggestion
```

efiop (approved):
Thank you! :pray:

efiop:
One more thing would be to use this in `dvc list`, but that could be added later.

dvc/output/base.py
skshetry: > good suggestion about a shorter variable name.

The suggestion was regarding the keyword in the dvc.yaml/dvcfile, not the variable name. I have no strong opinion towards either `exec` or `isexec`, but the decision should not be based on that pylint complained.

jorgeorpinel: I think that `isexec` seems better than `exec` as the latter sounds like "execute this output: true" which is confusing.

jorgeorpinel:
Will there be a WARN msg or something on Windows? Does it even matter that much? (I think you can usually try to execute any file on Win, it's more about the file extension).

efiop:
@jorgeorpinel No warning. It is a minor feature that is safe to ignore on Windows. Same thing with git, so shouldn't be a big suprise for anyone, but we'll see if someone complains.

jorgeorpinel:
K thanks. I was just double checking (for docs) as I read about that warning somewhere (in the OP I think).

PR #5064, opened 2020-12-09
experiments: support experiments cache collection for exp push/pull/gc

### LINKED ISSUE #4440: experiments: need commands for managing staged/stashed exp list

### LINKED ISSUE #4897: Sharing for experiments and checkpoints

pmrowla:

Will close #4897.
Will close #4440.

* Brancher now supports `all_experiments` for iterating over experiment commits
* `--all-commits`/`all_commits` for any DVC command now explicitly refers to all commits reachable through standard Git refs: `HEAD`, `refs/tags/*`, `refs/heads/*`, `remotes/refs/*`.
    * Previously, it referred to all commits reachable through any ref in `refs/*`.
    * This should not visibily affect behavior for DVC users unless they were using their own custom ref namespaces.
    * Experiments are explicitly excluded from this list of commits.
* Adds `dvc gc --all-experiments` option to preserve cache for experiments.
* `dvc exp push` now pushes DVC cache for the experiment by default
    * `--no-cache` will skip pushing DVC cache (only pushes to Git)
    * `-r/--remote`, `-j/--jobs`, `--run-cache` can be used and behave the same as for `dvc push`
* `dvc exp pull` now pulls DVC cache for the experiment by default
    * New args are the same as `dvc exp push`

### DIFF
--- a/dvc/repo/experiments/utils.py
+++ b/dvc/repo/experiments/utils.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, Generator
+from typing import TYPE_CHECKING, Generator, Iterable
 
 from .base import EXEC_NAMESPACE, EXPS_NAMESPACE, EXPS_STASH, ExpRefInfo
 
@@ -71,3 +71,15 @@ def remote_exp_refs_by_baseline(
         if ref.startswith(EXEC_NAMESPACE) or ref == EXPS_STASH:
             continue
         yield ExpRefInfo.from_ref(ref)
+
+
+def exp_commits(
+    scm: "Git", ref_infos: Iterable["ExpRefInfo"] = None
+) -> Generator[str, None, None]:
+    """Iterate over all experiment commits."""
+    shas = set()
+    refs = ref_infos if ref_infos else exp_refs(scm)
+    for ref_info in refs:
+        shas.update(scm.branch_revs(str(ref_info), ref_info.baseline_sha))
+        shas.add(ref_info.baseline_sha)
+    yield from shas
--- a/dvc/scm/git/backend/gitpython.py
+++ b/dvc/scm/git/backend/gitpython.py
@@ -215,10 +215,15 @@ class GitPythonBackend(BaseGitBackend):  # pylint:disable=abstract-method
         return [t.name for t in self.repo.tags]
 
     def list_all_commits(self):
+        head = self.get_ref("HEAD")
+        if not head:
+            # Empty repo
+            return []
+
         return [
             c.hexsha
             for c in self.repo.iter_commits(
-                rev="HEAD", branches=True, tags=True, remotes=True,
+                rev=head, branches=True, tags=True, remotes=True,
             )
         ]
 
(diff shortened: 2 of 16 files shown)


dvc/scm/git/backend/gitpython.py
```diff
@@ -215,7 +215,17 @@ def list_tags(self):
         return [t.name for t in self.repo.tags]
 
     def list_all_commits(self):
-        return [c.hexsha for c in self.repo.iter_commits("--all")]
```
pmrowla, line 218: `git rev-parse --all` returns every commit reachable by any ref in `refs/*`. If this is actually the behavior we want, we need to adjust this to be `git rev-parse --exclude=<experiment_refs> --all`. But it seems to me like the actual desired behavior for `--all-commits` is commits reachable from "standard" git refs (`HEAD`, `refs/tags`, `refs/heads` (branches), `remotes/*/refs/tags`, `remotes/*/refs/heads`).

Unless there's actually DVC users out there that are using their own custom refspecs, this change should not affect anything.

efiop (approved):

dvc/repo/experiments/utils.py
```diff
@@ -71,3 +71,15 @@ def remote_exp_refs_by_baseline(
         if ref.startswith(EXEC_NAMESPACE) or ref == EXPS_STASH:
             continue
         yield ExpRefInfo.from_ref(ref)
+
+
+def exp_commits(
+    scm: "Git", ref_infos: Iterable["ExpRefInfo"] = None
+) -> Generator[str, None, None]:
```
skshetry, line 78: I think, you can just do
```suggestion
) -> Iterator[str]:
```

skshetry (approved):

pared (approved):

Issue #4297, opened 2020-07-28
commit: support granularity


pared:
Update: see https://github.com/iterative/dvc/issues/4297#issuecomment-739848041

~~version : `1.2.2`~~

~~commiting output instead of stage, when there is no `dvc.yaml` results in:~~

~~`ERROR: failed to commit data - 'dvc.yaml' does not exist.`~~

~~in `0.94.0` it used to be:
`ERROR: failed to commit data - bad DVC-file name 'data'. DVC-files should be named 'Dvcfile' or have a '.dvc' suffix (e.g. 'data.dvc').`~~

~~Reproduction script:~~
```
#!/bin/bash

rm -rf repo
mkdir repo

pushd repo
git init --quiet
dvc init --quiet

echo data >> data
dvc add data

git add -A
git commit -am "init"

dvc commit data
```
NOTE:
When `dvc.yaml` is present, error is more readable:
`failed to commit data - "Stage 'data' not found inside 'dvc.yaml' file"
`

efiop:
The problem here is that we are using `collect` instead of `collect_granular` when resolving the target. Likely just need to do the switch. CC @skshetry or am I missing something?

skshetry:
@efiop, I don't think it was ever supported. Switching from `collect` to `collect_granular` would help, though it won't help with outputs being granular, it will commit all of the stage outputs.

efiop:
@skshetry Thanks! So looks like collect_granular + filter_info support for `commit` and we should be set :slightly_smiling_face:

efiop:
Discussed with @skshetry that we could start with just output-level granularity, e.g.
```
dvc add datafile
...
dvc commit datafile

dvc add datadir
...
dvc commit datadir
```
More granular commits like
```
dvc add datadir
....
dvc commit datadir/subdir/file
```
would require a bit more work on `cache.save` side, for it to accept filter_info and handle it properly, so we could do it as the next step after output-level granularity.

mbiesek:
@efiop I would like to start working on this.

efiop:
@mbiesek Great! Let us know if you'll have any questions :slightly_smiling_face:

skshetry:
@pmrowla, can we close this one?

dberenbaum:
@skshetry Can we close this one as completed or maybe open a new issue/edit the title to reflect the remaining work you would like to address? I think the original issue is solved, right?

PR #5088, opened 2020-12-13
commit: support granularity (#4297)

mbiesek:

### DIFF
--- a/dvc/output/base.py
+++ b/dvc/output/base.py
@@ -303,13 +303,21 @@ class BaseOutput:
         if self.isfile() and self.isexec:
             self.tree.set_exec(self.path_info)
 
-    def commit(self):
+    def commit(self, filter_info=None):
         if not self.exists:
             raise self.DoesNotExistError(self)
 
-        assert self.hash_info
+        hash_info = (
+            self.cache.get_hash(self.tree, filter_info)
+            if filter_info
+            else self.hash_info
+        )
+        path_info = filter_info if filter_info else self.path_info
+
+        assert hash_info
+
         if self.use_cache:
-            self.cache.save(self.path_info, self.cache.tree, self.hash_info)
+            self.cache.save(path_info, self.cache.tree, hash_info)
             self.set_exec()
 
     def dumpd(self):
--- a/dvc/output/base.py
+++ b/dvc/output/base.py
@@ -307,17 +307,15 @@ class BaseOutput:
         if not self.exists:
             raise self.DoesNotExistError(self)
 
-        hash_info = (
-            self.cache.get_hash(self.tree, filter_info)
-            if filter_info
-            else self.hash_info
-        )
-        path_info = filter_info if filter_info else self.path_info
-
-        assert hash_info
+        assert self.hash_info
 
         if self.use_cache:
-            self.cache.save(path_info, self.cache.tree, hash_info)
+            self.cache.save(
+                self.path_info,
+                self.cache.tree,
+                self.hash_info,
+                filter_info=filter_info,
+            )
             self.set_exec()
 
     def dumpd(self):
--- a/tests/func/test_commit.py
+++ b/tests/func/test_commit.py
@@ -83,6 +83,12 @@ def test_commit_no_exec(tmp_dir, dvc):
     assert dvc.status(stage.path) == {}
 
 
+def test_commit_granularity(tmp_dir, scm, dvc):
+    tmp_dir.dvc_gen("foo", "foo", commit="add files")
+    dvc.commit("foo")
+    assert dvc.status() == {}
+
+
 def test_commit_no_exec_missing_dep(tmp_dir, dvc):
     stage = dvc.run(
         name="my", cmd="mycmd", deps=["dep"], outs=["out"], no_exec=True
--- a/tests/func/test_commit.py
+++ b/tests/func/test_commit.py
@@ -1,4 +1,4 @@
-from os import fspath
+import os
 
 import pytest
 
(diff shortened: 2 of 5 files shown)


dvc/output/base.py
```diff
@@ -288,13 +288,13 @@ def save(self):
...
-    def commit(self):
+    def commit(self, filter_info=None):
         if not self.exists:
             raise self.DoesNotExistError(self)
-
+        path_info = filter_info if filter_info else self.path_info
         assert self.hash_info
         if self.use_cache:
-            self.cache.save(self.path_info, self.cache.tree, self.hash_info)
+            self.cache.save(path_info, self.cache.tree, self.hash_info)
```
efiop, line 297: `self.hash_info` doesn't really correspond to the `filter_info`, we need to either use that "partial" hash for filter_info or teach `self.cache.save` use `self.hash_info` but take `filter_info` into consideration.

dvc/repo/stage.py
```diff
@@ -31,6 +31,9 @@ class StageInfo(NamedTuple):
     stage: "Stage"
     filter_info: Optional[PathInfo] = None
 
+    def commit(self) -> None:
+        self.stage.commit(filter_info=self.filter_info)
+
```
efiop, line 36: Could you explain why we need this and can't just pass filter_info to stage.commit() in dvc/repo/commit.py?

tests/unit/tree/test_repo.py
```diff
@@ -46,6 +46,11 @@ def test_open_dirty_no_hash(tmp_dir, dvc):
         assert fobj.read() == "file"
 
 
+def test_commit_granularity(tmp_dir, scm, dvc):
+    tmp_dir.dvc_gen("foo", "foo", commit="add files")
+    dvc.commit("foo")
```
efiop, line 51: Seems like this should be in `tests/func/test_commit.py`

Also, need to better check that commit actually works.

efiop (approved):
Hi @mbiesek !

Thanks for the PR! :pray: The WIP skeleton looks pretty good! Please see a few questions above.

tests/unit/tree/test_repo.py
mbiesek: Not sure if my current changes are sufficient

dvc/repo/stage.py
mbiesek: Done - changed solution

dvc/output/base.py
mbiesek: Done

efiop:
@mbiesek Linters fail. Please install our pre-commit hooks, as noted in the contrib guide. You'll need to `pre-commit run --all-files` to fix/check already commited changes.

dvc/output/base.py
```diff
@@ -288,10 +290,11 @@ def save(self):
 
         self.hash_info = self.get_hash()
 
-    def commit(self):
+    def commit(self, filter_info=None):
         if not self.exists:
             raise self.DoesNotExistError(self)
-
+        if filter_info:
+            self.hash_info = self.get_hash(filter_info)
```
efiop, line 297: Why are we changing `self.hash_info` though?

dvc/output/base.py
```diff
@@ -288,10 +290,11 @@ def save(self):
...
         assert self.hash_info
         if self.use_cache:
             self.cache.save(self.path_info, self.cache.tree, self.hash_info)
```
efiop, line 300: Now `self.path_info` doesn't correspond to `self.hash_info`, as `self.hash_info` is for `filter_info` and not for `self.path_info` :slightly_frowning_face:

efiop: You could do something like
```suggestion
            self.cache.save(self.path_info, self.cache.tree, self.hash_info, filter_info=filter_info)
```
and let the cache properly filter `hash_info` on itsown (that way we will avoid computing the hash twice as you currently do in `get_hash(filter_info)`).

Or you could
```
self.cache.save(filter_info, self.cache.tree, hash_info)
```
where hash_info is self.hash_info but filtered according to `filter_info`. Notice that you don't have to do a full `get_hash` call for `filter_info`, you could just filter the existing `self.hash_info` according to `filter_info`.

mbiesek: Doene I choose to perform decision on output/base.py layer

dvc/dependency/param.py
```diff
@@ -105,7 +105,7 @@ def read_params(self):
                 pass
         return ret
 
-    def get_hash(self):
+    def get_hash(self, filter_info=None):
```
efiop, line 108: Seems unused?

dvc/output/base.py
```diff
@@ -288,13 +289,19 @@ def save(self):
...
         self.hash_info = self.get_hash()
 
-    def commit(self):
+    def commit(self, filter_info=None):
         if not self.exists:
             raise self.DoesNotExistError(self)
 
-        assert self.hash_info
+        hash_info = (
+            self.get_hash(filter_info) if filter_info else self.hash_info
```
efiop, line 297: We need to filter the hash_info (done in memory from existing self.hash_info.dir_info) and not re-compute the hash (actually goes to the filesystem and computes the hash from scratch). So seems like you don't really need to change `get_hash()` method at all.

mbiesek: Regarding changing get_hash method I remove this change but I am not sure how I could filter hash_info with information about filter_info in the moment of calling `dvc/output/base.py commit` method since self.hash_info.dir_info is None in this moment.

tests/func/test_commit.py
```diff
@@ -83,6 +83,12 @@ def test_commit_no_exec(tmp_dir, dvc):
     assert dvc.status(stage.path) == {}
 
 
+def test_commit_granularity(tmp_dir, scm, dvc):
+    tmp_dir.dvc_gen("foo", "foo", commit="add files")
+    dvc.commit("foo")
+    assert dvc.status() == {}
```
skshetry, line 89: This test only tests the `collect_granular`, not the granularity of the `commit`.

efiop:
Thank you @mbiesek !

Issue #5117, opened 2020-12-16
Parameter Diff while using Experimental Parameterization feature


bharathc346:
# Bug Report

### Reproduce
In `master` branch I have a `params.yaml` like so: 
```
datasets:
  - A
  - B
task: classification
seed: 42
feature_type: SparseDataset
training:
  alpha: 0.01
```
In `exp` branch I have a `params.yaml` like so:
```
datasets:
  - A
  - B
task: classification
seed: 42
feature_type: SparseDataset
training:
  alpha: 10.0
```

I have a training stage in `dvc.yaml` like so:
```
train_model:
    foreach: ${datasets}
    do:
      cmd: >-
        python scripts/train.py
        -- dataset ${item}
    params:
        - training.alpha
```

When I run `dvc params diff` in my `exp` branch I get
`ERROR: unexpected error - duplicated key 'training.alpha'`

Example:

1. dvc init
2. Copy dataset.zip to the directory
3. dvc add dataset.zip
4. dvc run -d dataset.zip -o model ./train.sh
5. modify dataset.zip
6. dvc repro

### Expected
Path         Param           Old    New                               
params.yaml  training.alpha  0.01   10.0

### Environment information
Mac OS Catalina 10.15.7
Python 3.7.9

**Output of `dvc version`:**

```console
$ dvc version
```
Tried on different dvc versions (1.11.8, master, 1.10.2) and only works on 1.10.2

**Additional Information (if any):**
Output of `dvc params diff -v`: 

```
2020-12-16 13:07:56,623 DEBUG: Check for update is enabled.
2020-12-16 13:07:56,639 DEBUG: fetched: [(3,)]
2020-12-16 13:07:56,703 DEBUG:    37.35 ms in resolving values
2020-12-16 13:07:56,824 DEBUG:    34.89 ms in resolving values
2020-12-16 13:07:56,851 DEBUG: fetched: [(5,)]
2020-12-16 13:07:56,858 DEBUG: fetched: [(3,)]
2020-12-16 13:07:56,911 DEBUG:    35.37 ms in resolving values
2020-12-16 13:07:56,946 DEBUG: fetched: [(5,)]
2020-12-16 13:07:56,949 ERROR: unexpected error - duplicated key 'training.alpha'
------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/bharathc/miniconda3/envs/dev/lib/python3.7/site-packages/dvc/main.py", line 90, in main
    ret = cmd.run()
[...]

PR #5121, opened 2020-12-17
params/metrics diff: collect from revisions/workspace twice

skshetry:
During diff, we were loading the stages from the current `repo.tree`'s state, from the workspace and from the revisions provided, which would collect 3-4 times (noticed while looking at the logs in https://github.com/iterative/dvc/issues/5117#issue-769158964).
This is good for using `params.diff()` and `metrics.diff()` as an API, but slows down the cli usage.

### DIFF
--- a/dvc/repo/metrics/diff.py
+++ b/dvc/repo/metrics/diff.py
@@ -3,12 +3,10 @@ from dvc.utils.diff import diff as _diff
 from dvc.utils.diff import format_dict
 
 
-def _get_metrics(repo, *args, rev=None, **kwargs):
+def _get_metrics(repo, *args, revs=None, **kwargs):
     try:
-        metrics = repo.metrics.show(
-            *args, **kwargs, revs=[rev] if rev else None
-        )
-        return metrics.get(rev or "", {})
+        metrics = repo.metrics.show(*args, **kwargs, revs=revs)
+        return metrics
     except NoMetricsError:
         return {}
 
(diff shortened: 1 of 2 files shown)


dvc/repo/metrics/diff.py
```diff
@@ -19,8 +17,12 @@ def diff(repo, *args, a_rev=None, b_rev=None, **kwargs):
 
     with_unchanged = kwargs.pop("all", False)
 
-    old = _get_metrics(repo, *args, **kwargs, rev=(a_rev or "HEAD"))
-    new = _get_metrics(repo, *args, **kwargs, rev=b_rev)
+    a_rev = a_rev or "HEAD"
+    b_rev = b_rev or "workspace"
```
skshetry, line 21: Note that the changes here are not equivalent, as the previous one would default to the `repo.tree`'s state whereas here it's the workspace.

skshetry: If we want to make it equivalent, we will likely need a knob in the `brancher` to not yield the "workspace".

pared: Considering that help message (for `metrics diff`) mentions that `b_rev` defaults to `workspace`, I would say its actually bug fixed. Also, `plots diff` defaults to `workspace`.

pared (approved):

efiop (approved):

PR #5123, opened 2020-12-17
params diff: skip params dependency for vars

skshetry:
Previously, `configs` which was a list of PathInfo was being used
to check if the `vars` should be skipped or not.

But, the file is stored as a string that made that check redundant
which might have caused #5117 as they were being stored twice,
once as an expanded paramdependency and next as a flattened dict
from vars, which later during `flatten`, it might have panicked
from that duplication.

### DIFF
--- a/dvc/repo/params/show.py
+++ b/dvc/repo/params/show.py
@@ -50,17 +50,20 @@ def _read_params(repo, configs, rev):
     return res
 
 
-def _add_vars(repo, configs, params):
+def _add_vars(repo, params):
+    vars_params = {}
     for stage in repo.stages:
         if isinstance(stage, PipelineStage) and stage.tracked_vars:
             for file, vars_ in stage.tracked_vars.items():
                 # `params` file are shown regardless of `tracked` or not
                 # to reduce noise and duplication, they are skipped
-                if file in configs:
+                if file in params:
                     continue
 
-                params[file] = params.get(file, {})
-                params[file].update(vars_)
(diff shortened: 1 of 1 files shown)


dvc/repo/params/show.py
```diff
@@ -50,17 +51,18 @@ def _read_params(repo, configs, rev):
     return res
 
 
-def _add_vars(repo, configs, params):
+def _add_vars(repo, params):
```
pared, line 54: I think that maybe we should change the naming, too. Now we are not adding anything. How about `_stage_vars`, to stress the fact that those are params loaded from the stage?

pared (approved):

dvc/repo/params/show.py
skshetry, line 54:
```suggestion
def _collect_vars(repo, params):
```

dvc/repo/params/show.py
```diff
@@ -70,7 +72,11 @@ def show(repo, revs=None):
     for branch in repo.brancher(revs=revs):
         configs = _collect_configs(repo, branch)
         params = _read_params(repo, configs, branch)
-        _add_vars(repo, configs, params)
+        vars_params = _add_vars(repo, params)
```
skshetry, line 75:
```suggestion
        vars_params = _collect_vars(repo, params)
```

pmrowla (approved):

skshetry:
@Mergifyio backport 1.11

Issue #4841, opened 2020-11-04
2.0 release checklist


efiop:
2.0 will be a major release, which is a good chance to change some things we've been talking about in a potentially non-backward compatible way. Creating this ticket as a reminder for smaller things, as they might get forgotten while working on major planned stuff. Please feel free to add other points

* ~~rename .dvc/cache into .dvc/objects(or smth like that)~~ delayed for 3.0
* ~~get rid of CRLF conversion when computing md5~~ https://github.com/iterative/dvc/issues/4658 will be handled in 3.0 cache
* ~~either get rid of *.dvc files or at least fix the order of fields in it. E.g.~~ leaving for 3.0
```
- md5: 456789
  path: myfile
```
should really be
```
- path: myfile
  md5: 456789
```
so that we create less confusion when merging/deleting/--no-exec'ing.
- ~~change cache/remote format from:~~ https://github.com/iterative/dvc/issues/829
```
12/3456
98/7654
...
runs/
```
to
```
objects/  # for potential chunking https://github.com/iterative/dvc/issues/829
files/  # same as our current cache files, aka assembled objects
dirs/  # for .dir files, will help us quicker list .dir files in push/pull/status optimization (currently we just walk the whole remote looking for .dir files)
runs/
tmp/  # for .tmp files that we use for universal atomic upload/download (download .tmp + rename) ?
...
```
also worth making `files/` use `12/123456` format like git and our runs/ do instead of the current `12/3456`

- ~~Change .dir format~~ https://github.com/iterative/dvc/issues/829#issuecomment-711137723  https://github.com/iterative/dvc/issues/3791
- ~~Drop dvc run --single-stage https://github.com/iterative/dvc/issues/3936 ?~~ Not a blocker, could be removed later.
- ~~switch to sha256(512?) as a default algo and drop dosunix for md5~~ https://github.com/iterative/dvc/issues/829
- [x] <s>drop .dvc files for `dvc add/import` in favor of special fields in dvc.yaml </s> Decided to do that after 2.0. Clear that we will be moving this way so no reason to refactor old .dvc file schema.
- ~~revisit API exceptions~~ (ping @skshetry ) For 3.0
- ~~(optional) revisit config options for remotes, maybe get create an alias to later get rid of things like gdrive_*~~ a topic for discussion
- [x] adjust conda package to not install extras by default https://github.com/conda-forge/dvc-feedstock/issues/168
[...]

PR #5125, opened 2020-12-17
Cleanup debug log messages when loading stages

skshetry:
1. Simplified callback stage debug message
2. Reduced no lock entry message to trace
   If lock file does not have some entry, it will simply complain
   that lock file is not up to date once.
   If there is no lock file or empty lock file, it will complain that lock file not found.
3. Annoying parametrization related wdir-interpolated debug message removed.

### DIFF
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -256,7 +256,9 @@ class Stage(params.StageParams):
             return False
 
         if self.is_callback:
-            logger.debug("%s has a command but no dependencies", self.addressing)
+            logger.debug(
+                "%s has a command but no dependencies", self.addressing
+            )
             return True
 
         if self.always_changed or self.is_checkpoint:
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -256,7 +256,7 @@ class Stage(params.StageParams):
             return False
 
         if self.is_callback:
-            logger.debug("%s has no command and dependencies", self.addressing)
+            logger.debug("%s has a command but no dependencies", self.addressing)
             return True
 
         if self.always_changed or self.is_checkpoint:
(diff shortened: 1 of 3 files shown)


dvc/stage/__init__.py
```diff
@@ -256,12 +256,7 @@ def changed_deps(self):
             return False
 
         if self.is_callback:
```
efiop, line 258: The whole `callback` logic can be deleted now, it has been obsoleted since 1.0

dvc/stage/__init__.py
```diff
@@ -256,12 +256,7 @@ def changed_deps(self):
...
-            logger.debug(
-                '%s is a "callback" stage '
-                "(has a command and no dependencies) and thus always "
-                "considered as changed.",
-                self,
-            )
+            logger.debug("%s has no command and dependencies", self.addressing)
```
skshetry, line 259:
```suggestion
            logger.debug("%s has a command but no dependencies", self.addressing)
```

dvc/stage/__init__.py
skshetry: How so? The callback is still present on the outputs during run/repro, so let's remove it later at the same time.

dvc/stage/__init__.py
```diff
@@ -256,12 +256,7 @@ def changed_deps(self):
             return False
 
         if self.is_callback:
-            logger.debug(
-                '%s is a "callback" stage '
-                "(has a command and no dependencies) and thus always "
-                "considered as changed.",
-                self,
-            )
+            logger.debug("%s has a command but no dependencies", self.addressing)
```
skshetry, line 259:
```suggestion
            logger.debug(
                "%s has a command but no dependencies",
                self.addressing
            )
```

dvc/stage/__init__.py
efiop: We have a checkbox https://github.com/iterative/dvc/issues/4841 . Sure, can do that separately, just saying.

efiop (approved):

PR #5139, opened 2020-12-20
tree: local: remove misc redundant logic

efiop:
Non-lazy logging, fspath, redundant exist check, etc.

### DIFF
--- a/dvc/tree/local.py
+++ b/dvc/tree/local.py
@@ -128,12 +128,10 @@ class LocalTree(BaseTree):
                 yield PathInfo(f"{root}{os.sep}{file}")
 
     def is_empty(self, path_info):
-        path = path_info.fspath
-
-        if self.isfile(path_info) and os.path.getsize(path) == 0:
+        if self.isfile(path_info) and os.path.getsize(path_info) == 0:
             return True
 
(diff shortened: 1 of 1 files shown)


dvc/tree/local.py
```diff
@@ -292,14 +281,15 @@ def _unprotect_dir(self, path):
             self._unprotect_file(fname)
 
     def unprotect(self, path_info):
-        path = path_info.fspath
-        if not os.path.exists(path):
-            raise DvcException(f"can't unprotect non-existing data '{path}'")
```
efiop, line 297: We have a test for broken symlink removal (`dvc.remove()`)that checks for this exception. It should really be ignored on a higher level, not raised in a special exception.

dvc/tree/local.py
```diff
@@ -142,12 +140,7 @@ def remove(self, path_info):
         if isinstance(path_info, PathInfo):
             if path_info.scheme != "local":
                 raise NotImplementedError
-            path = path_info.fspath
-        else:
-            path = path_info
-
-        if self.exists(path):
-            remove(path)
+        remove(path_info)
```
efiop, line 143: `remove` is ok with nonexistent path_info

PR #5145, opened 2020-12-22
erepo: get rid of get_checksum

efiop:
The intended behavior is to not ignore subrepos but not follow them. Think about it as `os.walk` seeing symlinks but not following them unless `followlinks` is set to `True`.

This is one of the pre-requisites to simplifying and possibly getting rid of ExternalRepo.

### DIFF
--- a/dvc/tree/repo.py
+++ b/dvc/tree/repo.py
@@ -309,7 +309,13 @@ class RepoTree(BaseTree):  # pylint:disable=abstract-method
                 yield from self._walk(repo_walk, None, dvcfiles=dvcfiles)
 
     def walk(
-        self, top, topdown=True, onerror=None, dvcfiles=False, **kwargs
+        self,
+        top,
+        topdown=True,
+        onerror=None,
+        dvcfiles=False,
+        follow_subrepos=None,
+        **kwargs
     ):  # pylint: disable=arguments-differ
         """Walk and merge both DVC and repo trees.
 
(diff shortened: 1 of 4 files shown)


dvc/tree/repo.py
```diff
@@ -336,13 +342,17 @@ def walk(
...
+            ignore_subrepos = not follow_subrepos
+
         tree, dvc_tree = self._get_tree_pair(top)
         repo_exists = tree.exists(top)
         repo_walk = tree.walk(
             top,
             topdown=topdown,
             onerror=onerror,
-            ignore_subrepos=not self._traverse_subrepos,
+            ignore_subrepos=ignore_subrepos,
```
efiop, line 355: We were confusing seeing and following subrepos before, but in the future there should be two options: `ignore_subrepos` and `follow_subrepos`.

dvc/tree/repo.py
```diff
@@ -336,13 +342,17 @@ def walk(
                 onerror(NotADirectoryError(top))
             return
 
+        ignore_subrepos = not self._traverse_subrepos
+        if follow_subrepos is not None:
```
skshetry, line 346:
```suggestion
        if follow_subrepos:
```

dvc/tree/repo.py
```diff
@@ -336,13 +342,17 @@ def walk(
...
+            ignore_subrepos = not follow_subrepos
```
skshetry, line 347: This is a bit confusing.

skshetry (approved):

dvc/tree/repo.py
efiop: @skshetry Agreed. I think `_traverse_subrepos` might go away in favor of explicit follow_subrepos + ignore_subrepos.

Issue #5150, opened 2020-12-22
diff: detect renames


cgebbe:
File renaming currently shows up in DVC DIFF <another_git_branch> as:

Added: (new filename after renaming)
Deleted: (old filename)

Would it be possible to group this information into a new category "Renamed" so that the overview is a bit clearer?

P.s.: If somebody could point me into the right direction, would like to create a pull request myself.

efiop:
More context: https://discuss.dvc.org/t/merging-of-files-in-dvc-tracked-directories/599/5

pmrowla:
Keeping this issue open for now since there are a few outstanding tasks that will need to be addressed eventually

* [ ] Make renaming detection in multiple matching file case consistent with git's
* [ ] Support similarity/rename detection once we have support for rolling hash based file chunking
* [ ] Consider options similar to git's `--no-renames`

efiop:
Closing since the rest of tasks don't seem to be in demand.

PR #5160, opened 2020-12-25
dvc.command.diff: add renamed detection feature

dkropachev:
This PR is my application for a job
fixes https://github.com/iterative/dvc/issues/5150

* [ ] 📖 ~~If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here.~~

### DIFF
--- a/tests/func/params/test_diff.py
+++ b/tests/func/params/test_diff.py
@@ -1,3 +1,7 @@
+import os
+
+import pytest
+
 from dvc.utils import relpath
 from dvc.utils.serialize import dump_yaml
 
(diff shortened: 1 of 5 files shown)


pmrowla (changes requested):
PR looks pretty good so far, thanks!

One question I have is about the case where we have more than one file with the same hash.

If we remove the originals and then add a new file with the same hash, we will end up showing one arbitrary/random path as renamed, and the rest as removed. Behavior-wise, this seems correct, and git will do the same kind of thing and arbitrarily choose one original `.dvc` file as "renamed" and the rest as "deleted". My concern is that right now we may end up choosing a different file to show as "renamed" than git.

Playing around with it, using the following simple example:
```bash
git init
dvc init
echo "foo" > foo.txt
echo "foo" > bar.txt
dvc add foo.txt bar.txt
git add .
git commit -m "init"

dvc remove foo.txt.dvc bar.txt.dvc
rm bar.txt
mv foo.txt baz.txt
dvc add baz.txt
git add .
git commit -m "rename"

git --no-pager diff --name-status HEAD~1 HEAD

dvc diff HEAD~1 HEAD
```

git will show `foo` as removed and `bar -> baz` as the rename

```
$ git --no-pager diff --name-status HEAD~1 HEAD
M       .gitignore
R077    bar.txt.dvc     baz.txt.dvc
D       foo.txt.dvc
```

However, in DVC we end up showing `bar` as removed and `foo -> baz` as the rename.

```
$ dvc diff HEAD~1 HEAD
Deleted:
    bar.txt

Renamed:
    foo.txt -> baz.txt

files summary: 0 added, 1 deleted, 1 renamed, 0 modified
```

This is a relatively minor issue, but I'm wondering if there is a way for us to leverage git's diff here so that our rename matches git's.

tests/func/params/test_diff.py
```diff
@@ -14,6 +16,76 @@ def test_diff_no_changes(tmp_dir, scm, dvc):
     assert dvc.params.diff() == {}
 
 
+def test_diff_rename_folder(tmp_dir, scm, dvc):
+    paths = tmp_dir.gen(
+        {"dir": {"file": "text1", "subdir": {"file2": "text2"}}}
+    )
+    tmp_dir.dvc_add(paths, commit="commit #1")
```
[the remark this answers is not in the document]
dkropachev: fixed

dkropachev:
> PR looks pretty good so far, thanks!
> 
> One question I have is about the case where we have more than one file with the same hash.
> 
> If we remove the originals and then add a new file with the same hash, we will end up showing one arbitrary/random path as renamed, and the rest as removed. Behavior-wise, this seems correct, and git will do the same kind of thing and arbitrarily choose one original `.dvc` file as "renamed" and the rest as "deleted". My concern is that right now we may end up choosing a different file to show as "renamed" than git.
> 
> Playing around with it, using the following simple example:
> 
> ```shell
> git init
> dvc init
> echo "foo" > foo.txt
> echo "foo" > bar.txt
> dvc add foo.txt bar.txt
> git add .
> git commit -m "init"
> 
> dvc remove foo.txt.dvc bar.txt.dvc
> rm bar.txt
> mv foo.txt baz.txt
> dvc add baz.txt
> git add .
> git commit -m "rename"
> 
> git --no-pager diff --name-status HEAD~1 HEAD
> 
> dvc diff HEAD~1 HEAD
> ```
> 
> git will show `foo` as removed and `bar -> baz` as the rename
> 
> ```
> $ git --no-pager diff --name-status HEAD~1 HEAD
> M       .gitignore
> R077    bar.txt.dvc     baz.txt.dvc
> D       foo.txt.dvc
> ```
> 
> However, in DVC we end up showing `bar` as removed and `foo -> baz` as the rename.
> 
> ```
> $ dvc diff HEAD~1 HEAD
> Deleted:
>     bar.txt
> 
> Renamed:
>     foo.txt -> baz.txt
> 
> files summary: 0 added, 1 deleted, 1 renamed, 0 modified
> ```
> 
> This is a relatively minor issue, but I'm wondering if there is a way for us to leverage git's diff here so that our rename matches git's.

I think it is right way to go, but there are following reasons to address this in other PR:
1. Currently scm.diff is not supporting reference resolution in same capacity as it is done for dvc.diff, in particilar it won't accept HEAD~1
2. Implementing reference resolution for scm will grow this PR beyond reasonable limits
3. In either case dvc needs it's own diff for NoSCM case

Peter, what you think ?

pmrowla:
> I think it is right way to go, but there are following reasons to address this in other PR:
> 
> 1. Currently scm.diff is not supporting reference resolution in same capacity as it is done for dvc.diff, in particilar it won't accept HEAD~1
> 2. Implementing reference resolution for scm will grow this PR beyond reasonable limits
> 3. In either case dvc needs it's own diff for NoSCM case
> 
> Peter, what you think ?

Regarding the diff limitation, ref resolution is already supported. You just need to call `rev = scm.resolve_rev("HEAD~1")` and pass the resolved SHA into `scm.diff()`. Although, the current implementation of `scm.diff` is probably also not really want we want here, since dulwich will return the text for the actual diff, and all we really want to know is the file status without the content changes.

For the no SCM case, it is fine for us to just choose whatever file we want as the "renamed" one, since we don't have to worry about matching with git.

I'm not saying that this is something that needs to go into this PR, but I think it's something that should be discussed.

karajan1001:
A good job, however, seems that it can only detect two files with exactly the same contents? 

Here I would recommend two algorithms [SimHash](https://en.wikipedia.org/wiki/SimHash) and [Locality-sensitive hashing
](https://en.wikipedia.org/wiki/Locality-sensitive_hashing)

I used to use them in the work of article deduplication of the news crawled by the web-crawlers.

skshetry:
> Seems that it can only detect two files with exactly the same contents? 

We should be able to support partial renaming when we have chunking support. But, it comes at an expense of making things slow (`git` only does this to a certain threshold). For now, this is good-enough solution, I think.

dkropachev:
Sooo, could you gyus merge it? :]

pmrowla:
@dkropachev have you looked into whether or not checking the git diff for renamed filenames is doable? As I noted earlier, the git ref resolution is already supported in DVC via `scm.resolve_rev()`.

dkropachev:
> @dkropachev have you looked into whether or not checking the git diff for renamed filenames is doable? As I noted earlier, the git ref resolution is already supported in DVC via `scm.resolve_rev()`.

Sorry, totally missed that part.
I've looked into the possibility of using scm.diff for that purpose, it is doable, key points:
1. Since dulwich operates only on git tree level it will be required to do something with "workspace" `revision`, we can temporary commit workspace files to the local git repo.
2. Also since dulwich uses it's own renaming detector logic (dulwich.diff_tree.py:RenameDetector) we won't get 100% same result is binary git
3. Small rewriting of DulwichBackend.diff will be required

dkropachev:
I've also found a bug that is triggered when files are added that have same hash as old files.

pmrowla:
> I've looked into the possibility of using scm.diff for that purpose, it is doable, key points:
> 
> 1. Since dulwich operates only on git tree level it will be required to do something with "workspace" `revision`, we can temporary commit workspace files to the local git repo.
> 2. Also since dulwich uses it's own renaming detector logic (dulwich.diff_tree.py:RenameDetector) we won't get 100% same result is binary git
> 3. Small rewriting of DulwichBackend.diff will be required

Thanks for looking into it, we can hold off on doing this for now then. It may be better for us to use pygit2's (libgit's) diff instead of dulwich since it has support for workspace diffs.

We're also already looking into doing rolling hash based chunking for files, so it will make more sense for us to just improve the diff renaming handling after that's done.

tests/func/params/test_diff.py
```diff
@@ -14,6 +16,107 @@ def test_diff_no_changes(tmp_dir, scm, dvc):
...
+
+
+def test_diff_rename_file(tmp_dir, scm, dvc):
+    paths = tmp_dir.gen(
+        {"dir": {"file": "text1", "subdir": {"file2": "text2"}}}
+    )
+    tmp_dir.dvc_add(paths, commit="commit #1")
+    (tmp_dir / "dir" / "file").replace(tmp_dir / "dir" / "subdir" / "file3")
+    tmp_dir.dvc_add(paths, commit="commit #2")
+    assert dvc.diff("HEAD~1") == {
```
pmrowla, line 95: Can we add test cases (or probably just parametrize the existing tests) that test the default diff case (uncommitted workspace changes against `HEAD`)?

Essentially we will want to make sure that something like:
```bash
echo "foo" > foo.txt
dvc add foo.txt
git add . && git commit -m "add foo"
dvc move foo bar
dvc diff
```
also shows the rename for `foo` -> `bar`

You'll have to rebase this PR against master to get the fix from https://github.com/iterative/dvc/pull/5205

pmrowla (changes requested):
Looks good for the most part, but we should have test cases for uncommitted workspace renames as well.

tests/func/params/test_diff.py
dkropachev: Tests added, rebased

pmrowla (approved):

Issue #1819, opened 2019-04-01
test: migrate from unittest to pytest


efiop:
Need to stop using unittest's TestCase and use pytest instead.

Suor:
Should we raise this in priority? 

Also, looks like we have 3 parts here:
- [ ] migrate `test_data_cloud.py`, look at `tests/remotes.py`
- [ ] migrate external dep tests
- [ ] migrate all other unit tests

efiop:
@Suor This one is still nice-to-have, but it doesn't mean that we can't take it 🙂 Considering the data_cloud complications, I suspect that it will take a bit longer to resolve this one, so I would take it after we migrate to new fixtures. Let's not add this one to the sprint for now.

Suor:
Maybe open a free hunt on it? @efiop @mroutis @pared @skshetry?

efiop:
@Suor Let's not do that in this sprint for now.

efiop:
Migrating tests/func/test_data_cloud.py

skshetry:
The unittest based tests have started to rot as we rarely update them compared to the fixtures. The `diskcache` changed removed the `repo.state` context manager, which made the tests prone to failure raising PermissionError in Windows as we are not closing the sqlite connection (which we still aren't). It was quite easy to debug and fix on pytest, but quite difficult on unittest based tests. Also to solve it, I had to add more hacks which I don't like (see: https://github.com/iterative/dvc/pull/5962).

nik123:
In one of my PRs I've been asked https://github.com/iterative/dvc/pull/6339#discussion_r673345814 to replace `unittest.mock` with `pytest-mock`. Is replacement of `unittest.mock` with `pytest-mock` is a part of this issue?

karajan1001:
> In one of my PRs I've been asked [#6339 (comment)](https://github.com/iterative/dvc/pull/6339#discussion_r673345814) to replace `unittest.mock` with `pytest-mock`. Is replacement of `unittest.mock` with `pytest-mock` is a part of this issue?

I think so.

skshetry:
Closed by #8523.

PR #5163, opened 2020-12-26
checkout: fix partial file checkouts in the same directory

### LINKED ISSUE #5158: pull/checkout: pulling multiple files consecutively from a folder checks out only first
# Bug Report

## Description

When attempting to pull individual files from dvc tracked folder, only the first pull creates file links. Similar behavior with `checkout` unless `--relink` is used.

### Reproduce
0. (git + dvc init)

1. mkdir db
2. touch db/a.txt && touch db/b.txt
3. dvc add db
4. git add db.dvc && git commit -m "dvc"
5. dvc remote add -d tmp /tmp/remote
6. dvc push
7. rm db/*
8. dvc pull db/a.txt
9. dvc pull db/b.txt

### Expected

db/a.txt & db/b.txt exist (When in fact b.txt doesn't). 

Getting messages:
dvc pull db/a.txt --> 1 file modified
dvc pull db/b.txt --> Everything is up to date. 

### Environment information

DVC version: 1.11.9 (pip)
Platform: Python 3.6.10 on Linux-4.18.0-1024-azure-x86_64-with-debian-buster-sid
Supports: http, https, s3
Cache directory: ext4 on /dev/sda1
Caches: local
Remotes: local
Workspace directory: ext4 on /dev/sda1
Repo: dvc, git

isidentical:
Resolves #5158

### DIFF
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -89,7 +89,10 @@ class CloudCache:
 
         return DirInfo.from_list(d)
 
-    def _get_filtered_hash_info(self, filter_info, hash_info, path_info):
+    def _filter_hash_info(self, hash_info, path_info, filter_info):
+        if not filter_info:
+            return hash_info
+
         dir_info = self.get_dir_cache(hash_info)
         hash_key = filter_info.relative_to(path_info).parts
 
@@ -135,31 +138,22 @@ class CloudCache:
             logger.debug("'%s' doesn't exist.", path)
             return True
 
-        if not hash_info:
-            logger.debug("hash value for '%s' is missing.", path_info)
+        hi = self._filter_hash_info(hash_info, path_info, filter_info)
+        if not hi:
+            logger.debug("hash value for '%s' is missing.", path)
             return True
 
-        if path != path_info:
-            filtered_hash_info = self._get_filtered_hash_info(
-                path, hash_info, path_info
-            )
-            if filtered_hash_info is not None:
-                path_info = filter_info
-                hash_info = filtered_hash_info
-
-        if self.changed_cache(hash_info):
-            logger.debug(
-                "cache for '%s'('%s') has changed.", path_info, hash_info
-            )
+        if self.changed_cache(hi):
+            logger.debug("cache for '%s'('%s') has changed.", path, hi)
             return True
 
-        actual = self.tree.get_hash(path_info)
-        if hash_info != actual:
+        actual = self.tree.get_hash(path)
+        if hi != actual:
             logger.debug(
                 "hash value '%s' for '%s' has changed (actual '%s').",
-                hash_info,
+                hi,
                 actual,
-                path_info,
+                path,
             )
             return True
 
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -89,7 +89,7 @@ class CloudCache:
 
         return DirInfo.from_list(d)
 
-    def changed(self, path_info, hash_info):
+    def changed(self, path_info, hash_info, filter_info=None):
         """Checks if data has changed.
 
         A file is considered changed if:
@@ -101,26 +101,28 @@ class CloudCache:
         Args:
             path_info: dict with path information.
             hash: expected hash value for this data.
+            filter_info: an optional argument to target a specific file.
 
         Returns:
             bool: True if data has changed, False otherwise.
         """
 
+        real_path_info = filter_info or path_info
         logger.trace(
-            "checking if '%s'('%s') has changed.", path_info, hash_info
+            "checking if '%s' ('%s') has changed.", path_info, hash_info
         )
 
-        if not self.tree.exists(path_info):
-            logger.debug("'%s' doesn't exist.", path_info)
+        if not self.tree.exists(real_path_info):
+            logger.debug("'%s' doesn't exist.", real_path_info)
             return True
 
         if not hash_info:
-            logger.debug("hash value for '%s' is missing.", path_info)
+            logger.debug("hash value for '%s' is missing.", real_path_info)
             return True
 
-        if self.changed_cache(hash_info):
+        if self.changed_cache(hash_info, filter_info=filter_info):
             logger.debug(
-                "cache for '%s'('%s') has changed.", path_info, hash_info
+                "cache for '%s'('%s') has changed.", real_path_info, hash_info
             )
             return True
 
@@ -521,7 +523,7 @@ class CloudCache:
             failed = path_info
 
         elif not relink and not self.changed(
-            filter_info or path_info, hash_info
+            path_info, hash_info, filter_info=filter_info
         ):
             logger.trace("Data '%s' didn't change.", path_info)
             skip = True
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -101,15 +101,25 @@ class CloudCache:
         Args:
             path_info: dict with path information.
             hash: expected hash value for this data.
-            filter_info: an optional argument to target a specific file.
+            filter_info: an optional argument to target a specific path.
 
         Returns:
             bool: True if data has changed, False otherwise.
         """
 
-        path = filter_info or path_info
+        hash_origin = path_info
+        if filter_info:
+            path = filter_info
+            dir_info = self.get_dir_cache(hash_info)
+            hash_key = path.relative_to(path_info).parts
+            if hash_key in dir_info.trie:
+                hash_info = dir_info.trie[hash_key]
+                hash_origin = filter_info
+        else:
+            path = path_info
+
         logger.trace(
-            "checking if '%s' ('%s') has changed.", path_info, hash_info
+            "checking if '%s'('%s') has changed.", path_info, hash_info
         )
 
         if not self.tree.exists(path):
@@ -126,13 +136,13 @@ class CloudCache:
             )
             return True
 
-        actual = self.tree.get_hash(path_info)
+        actual = self.tree.get_hash(hash_origin)
         if hash_info != actual:
             logger.debug(
                 "hash value '%s' for '%s' has changed (actual '%s').",
                 hash_info,
                 actual,
-                path_info,
+                hash_origin,
             )
             return True
 
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -107,22 +107,22 @@ class CloudCache:
             bool: True if data has changed, False otherwise.
         """
 
-        real_path_info = filter_info or path_info
+        path = filter_info or path_info
         logger.trace(
             "checking if '%s' ('%s') has changed.", path_info, hash_info
         )
 
-        if not self.tree.exists(real_path_info):
-            logger.debug("'%s' doesn't exist.", real_path_info)
+        if not self.tree.exists(path):
+            logger.debug("'%s' doesn't exist.", path)
             return True
 
         if not hash_info:
-            logger.debug("hash value for '%s' is missing.", real_path_info)
+            logger.debug("hash value for '%s' is missing.", path_info)
             return True
 
         if self.changed_cache(hash_info, filter_info=filter_info):
             logger.debug(
-                "cache for '%s'('%s') has changed.", real_path_info, hash_info
+                "cache for '%s'('%s') has changed.", path_info, hash_info
             )
             return True
 
@@ -136,7 +136,7 @@ class CloudCache:
             )
             return True
 
-        logger.trace("'%s' hasn't changed.", path_info)
+        logger.trace("'%s' hasn't changed.", path)
         return False
 
     def link(self, from_info, to_info):
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -104,9 +104,9 @@ class CloudCache:
                 filtered_dir_info.trie[key[depth:]] = value
         except KeyError:
             return None
-        else:
-            filtered_hash_info, _ = self._get_dir_info_hash(filtered_dir_info)
-            return filtered_hash_info
+
+        filtered_hash_info, _ = self._get_dir_info_hash(filtered_dir_info)
+        return filtered_hash_info
 
     def changed(self, path_info, hash_info, filter_info=None):
         """Checks if data has changed.
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -2,6 +2,7 @@ import itertools
 import json
 import logging
 from concurrent.futures import ThreadPoolExecutor
+from contextlib import suppress
 from copy import copy
 from typing import Optional
 
@@ -31,6 +32,10 @@ class DirCacheError(DvcException):
         super().__init__(
             f"Failed to load dir cache for hash value: '{hash_}'."
         )
+
+
+class _CantFilterCache(ValueError):
+    ...
 
 
 @decorator
@@ -103,8 +108,7 @@ class CloudCache:
             for key, value in dir_info.trie.items(hash_key):
                 filtered_dir_info.trie[key[depth:]] = value
         except KeyError:
-            # The directory doesn't exist on the trie, fallback to the original
-            return path_info, hash_info
+            raise _CantFilterCache
         else:
             filtered_hash_info, _ = self._get_dir_info_hash(filtered_dir_info)
             return filter_info, filtered_hash_info
@@ -127,14 +131,7 @@ class CloudCache:
             bool: True if data has changed, False otherwise.
         """
 
-        if filter_info is not None and filter_info != path_info:
-            path = filter_info
-            path_info, hash_info = self._get_filtered_hash_info(
-                path, hash_info, path_info
-            )
-        else:
-            path = path_info
-
+        path = filter_info or path_info
         logger.trace(
             "checking if '%s'('%s') has changed.", path_info, hash_info
         )
@@ -146,6 +143,12 @@ class CloudCache:
         if not hash_info:
             logger.debug("hash value for '%s' is missing.", path_info)
             return True
+
+        with suppress(_CantFilterCache):
+            if path != path_info:
+                path_info, hash_info = self._get_filtered_hash_info(
+                    path, hash_info, path_info
+                )
 
         if self.changed_cache(hash_info):
             logger.debug(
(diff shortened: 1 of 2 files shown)


tests/func/test_checkout.py
```diff
@@ -68,6 +68,37 @@ def test(self):
...
+
+        file1, file2 = files
+
+        self.dvc.checkout(targets=file1)
+        self.assertTrue(os.path.exists(self.CO_DIR))
+        self.assertTrue(os.path.exists(file1))
+        self.assertFalse(os.path.exists(file2))
+
+        self.dvc.checkout(targets=file2)
+        self.assertTrue(os.path.exists(file2))
```
efiop, line 99: We are currently continuing to migrate from unittest-style tests to pytest-style tests https://github.com/iterative/dvc/issues/1819 , so all new tests need to be in pytest-style. Could you please rewrite it? E.g. see examples at the bottom of this test_checkout.py file.

efiop: Also, we need to test a case for checking out a subdirectory in a tracked directory, that might break.

dvc/cache/base.py
```diff
@@ -520,7 +520,9 @@ def checkout(
             self.safe_remove(path_info, force=force)
             failed = path_info
 
-        elif not relink and not self.changed(path_info, hash_info):
+        elif not relink and not self.changed(
+            filter_info or path_info, hash_info
+        ):
```
efiop, line 525: hash_info is a hash_info of the full dataset, right? But filter_info might be about a subset. Please check `changed()` method, I don't think it is ready for this.

tests/func/test_checkout.py
isidentical: Applied both of them

dvc/cache/base.py
isidentical: You are right. In 323d20d, I tried to address by using filter_info in changed if the checks related to paths, and using the original path_info otherwise (like comparing the hashs)

dvc/cache/base.py
```diff
@@ -101,26 +101,28 @@ def changed(self, path_info, hash_info):
         Args:
             path_info: dict with path information.
             hash: expected hash value for this data.
+            filter_info: an optional argument to target a specific file.
 
         Returns:
             bool: True if data has changed, False otherwise.
         """
 
+        real_path_info = filter_info or path_info
```
skshetry, line 110: `real` *adj* might be confusing here. How about just `path`?

tests/func/test_checkout.py
```diff
@@ -880,3 +880,31 @@ def test_checkout_executable(tmp_dir, dvc):
...
+        db_dir / filename for filename in ("a.txt", "b.txt", sub_dir / "c.txt")
+    ]
+
+    shutil.rmtree(db_dir)
+
+    dvc.checkout(targets=[str(file_1)])
+    assert db_dir.exists()
+    assert file_1.exists()
+    assert not file_2.exists()
+    assert not sub_dir.exists()
```
skshetry, line 902: There's an easier way to express this in a more succinct way:
```suggestion
    assert db_dir.read_text() == {"a.txt": "a"}
```
This will check the contents as well as the folder.

tests/func/test_checkout.py
```diff
@@ -880,3 +880,31 @@ def test_checkout_executable(tmp_dir, dvc):
...
+
+    dvc.checkout(targets=[str(sub_dir_file)])
+    assert sub_dir.exists()
+    assert sub_dir_file.exists()
+    assert not file_2.exists()
```
skshetry, line 907: Similarly as https://github.com/iterative/dvc/pull/5163/files#r549324917,
```suggestion
    assert db_dir.read_text() == {"a.txt": "a", "sub_dir": {"c.txt": "c"}}
```

tests/func/test_checkout.py
```diff
@@ -880,3 +880,31 @@ def test_checkout_executable(tmp_dir, dvc):
...
+
+
+def test_checkout_partial(tmp_dir, dvc):
+    tmp_dir.dvc_gen(
+        {"db": {"a.txt": "a", "b.txt": "b", "sub_dir": {"c.txt": "c"}}}
+    )
+
+    db_dir = tmp_dir / "db"
+    sub_dir = db_dir / "sub_dir"
+    file_1, file_2, sub_dir_file = [
```
skshetry, line 892: Should we also add a test for `sub_dir`?

tests/func/test_checkout.py
```diff
@@ -880,3 +880,31 @@ def test_checkout_executable(tmp_dir, dvc):
...
+        db_dir / filename for filename in ("a.txt", "b.txt", sub_dir / "c.txt")
+    ]
+
+    shutil.rmtree(db_dir)
+
+    dvc.checkout(targets=[str(file_1)])
```
skshetry, line 898: *Nitpick*: `checkout` supports a single `str` argument as well. 
```suggestion
    dvc.checkout(targets=str(file_1))
```

tests/func/test_checkout.py
isidentical: I wasn't aware of this functionality, thanks @skshetry for letting me know!

tests/func/test_checkout.py
isidentical: Can you be more explicit about the case? We already have a test case to checkout a partial file from a subdirectory but if you mean anything, we can add that too!

tests/func/test_checkout.py
isidentical: ah, cool, thanks!

dvc/cache/base.py
```diff
@@ -101,24 +101,26 @@ def changed(self, path_info, hash_info):
         Args:
             path_info: dict with path information.
             hash: expected hash value for this data.
+            filter_info: an optional argument to target a specific file.
```
efiop, line 104: filter_info could also target a subdir, not only a sepcific file

dvc/cache/base.py
```diff
@@ -101,24 +101,26 @@ def changed(self, path_info, hash_info):
...
 
         Returns:
             bool: True if data has changed, False otherwise.
         """
 
+        path = filter_info or path_info
         logger.trace(
-            "checking if '%s'('%s') has changed.", path_info, hash_info
+            "checking if '%s' ('%s') has changed.", path_info, hash_info
```
efiop, line 112: Minor: this is an unrelated change. This should either be a separate commit or should not be in this PR at all. 
```suggestion
            "checking if '%s'('%s') has changed.", path_info, hash_info
```

dvc/cache/base.py
```diff
@@ -101,24 +101,26 @@ def changed(self, path_info, hash_info):
...
 
         if not hash_info:
             logger.debug("hash value for '%s' is missing.", path_info)
             return True
 
-        if self.changed_cache(hash_info):
+        if self.changed_cache(hash_info, filter_info=filter_info):
             logger.debug(
                 "cache for '%s'('%s') has changed.", path_info, hash_info
             )
```
efiop, line 126: In the logic below we compare `get_hash(path_info)` and `hash_info`, which means that if the directory you are trying to checkout a subdir/subfile of has some unrelated subdir/subfile changed, this method will think that the subdir/subfile is changed and will checkout it again. That's not right.

tests/func/test_checkout.py
efiop: @isidentical I believe @skshetry is saying that we should test `dvc checkout db/sub_dir` too. I left a comment about it as well.

dvc/cache/base.py
isidentical: From what I understand, we only have access to the hash_info that already presented in the dvcfile which is usually the base directory. So, this behavior seems correct yet inefficient but I failed to find any better alternative with the current dvcfile format. Do you have any hints for me regarding this?

tests/func/test_checkout.py
isidentical: Oh, right! Sure, can add it.

dvc/cache/base.py
efiop: @isidentical Take a look at `set_dir_info` helper. It tries to load `hash_info.dir_info` if the full info for `hash_info` is present in the cache. We could use that too here (or maybe `get_dir_cache` directly) to get a more complete info, then filter it out and use it for a correct comparison. If dir cache is not available - we could indeed stick to the current behavior as a fallback.

dvc/cache/base.py
```diff
@@ -101,40 +101,52 @@ def changed(self, path_info, hash_info):
...
             bool: True if data has changed, False otherwise.
         """
 
+        hash_origin = path_info
+        if filter_info:
+            path = filter_info
+            dir_info = self.get_dir_cache(hash_info)
+            hash_key = path.relative_to(path_info).parts
+            if hash_key in dir_info.trie:
+                hash_info = dir_info.trie[hash_key]
```
efiop, line 116: Are you sure this works correctly for a subdirectory?

efiop: Looks like this will be skipped for a subdir, right? But you can actually compute the hash for the subdir if you have hashes for all the files inside of it. Take a look at how we compute hash_info for the directory.

isidentical: I tried to address this problem in the 1ced36c, would you mind checking the latest revision out?

dvc/cache/base.py
```diff
@@ -101,40 +117,53 @@ def changed(self, path_info, hash_info):
...
 
         Returns:
             bool: True if data has changed, False otherwise.
         """
 
+        hash_origin = path_info
+        path = filter_info or path_info
+
         logger.trace(
             "checking if '%s'('%s') has changed.", path_info, hash_info
```
efiop, line 130: Needs update

dvc/cache/base.py
```diff
@@ -101,40 +117,53 @@ def changed(self, path_info, hash_info):
...
             return True
 
         if not hash_info:
             logger.debug("hash value for '%s' is missing.", path_info)
             return True
 
-        if self.changed_cache(hash_info):
+        if self.changed_cache(hash_info, filter_info=filter_info):
             logger.debug(
                 "cache for '%s'('%s') has changed.", path_info, hash_info
```
efiop, line 143: will need update too

dvc/cache/base.py
```diff
@@ -101,40 +117,53 @@ def changed(self, path_info, hash_info):
...
             hash: expected hash value for this data.
+            filter_info: an optional argument to target a specific path.
 
         Returns:
             bool: True if data has changed, False otherwise.
         """
 
+        hash_origin = path_info
+        path = filter_info or path_info
+
```
efiop, line 128: The naming is really confusing. Same with `_get_specialized_hash_info`

dvc/cache/base.py
```diff
@@ -101,40 +117,53 @@ def changed(self, path_info, hash_info):
...
+        # Instead of directly checking whether filter_info is present or not,
+        # we compare the target path with the path_info because sometimes
+        # filter_info might be equal to the path_info and in those cases we do
+        # not need to specialize the hash.
+        if path != path_info:
+            hash_origin, hash_info = self._get_specialized_hash_info(
+                path, hash_info, hash_origin
+            )
+
+        actual = self.tree.get_hash(hash_origin)
```
efiop, line 156: The logic here is really confusing. Maybe just filter the hash and then use a helper method to check if the target has changed.

dvc/cache/base.py
```diff
@@ -89,7 +89,23 @@ def load_dir_cache(self, hash_info):
...
+        else:
+            depth = len(hash_key)
+            specialized_dir_info = DirInfo()
+            try:
+                for key, value in dir_info.trie.items(hash_key):
+                    specialized_dir_info.trie[key[depth:]] = value
+            except KeyError:
+                return hash_origin, hash_info
+            else:
+                return path_info, self._get_dir_info_hash(specialized_dir_info)
```
efiop, line 106: This method is really confusing and is doing too much. The fallback logic here doesn't really apply to the filtered hash_info (btw `filtered` is probably a better term than `specialized` here).

Also note how giant the `else` statement is after `if & return`. We don't have this check enabled in pylint yet, but we do appreciate no-else-after-return unless there is a good reason to do otherwise.

efiop (commented):
We are on the right path, but need to clean up the code a bit :slightly_smiling_face:

dvc/cache/base.py
isidentical: I've kept the comparison logic in place, I think it seems a bit more simple now since we moved to the filtering part to the upper side. Could you check it again? (if you still find it complicated, I can move it out as well, as a separate function)

dvc/cache/base.py
```diff
@@ -89,7 +89,29 @@ def load_dir_cache(self, hash_info):
 
         return DirInfo.from_list(d)
 
-    def changed(self, path_info, hash_info):
+    def _get_filtered_hash_info(
+        self, filtered_path_info, hash_info, hash_origin
```
efiop, line 93: Doesn't seem like we need to introduce alternative names, only makes it more confusing
```suggestion
        self, filter_info, hash_info, path_info
```

dvc/cache/base.py
```diff
@@ -101,40 +123,53 @@ def changed(self, path_info, hash_info):
         Args:
             path_info: dict with path information.
             hash: expected hash value for this data.
+            filter_info: an optional argument to target a specific path.
 
         Returns:
             bool: True if data has changed, False otherwise.
         """
 
+        hash_origin = path_info
```
efiop, line 132: This assignment is probably not needed. You can probably just use `path_info`.

isidentical: Not really, this is the origin of the `hast_info`. If we can successfully filter it, we change it below to the `filter_info`. And when comparing hashes, we use this location for obtaining the current cache

dvc/cache/base.py
```diff
@@ -101,17 +121,26 @@ def changed(self, path_info, hash_info):
...
             bool: True if data has changed, False otherwise.
         """
 
+        if filter_info is not None and filter_info != path_info:
+            path = filter_info
+            path_info, hash_info = self._get_filtered_hash_info(
+                path, hash_info, path_info
+            )
+        else:
+            path = path_info
```
efiop, line 136: Notice how we don't really need filtered hash_info until `changed_cache` call below. We shouldn't try to filter it until we really need it, as it might be costly for big datasets. The `trace()` message that uses it can be removed or modified to just include filter_info too, so you don't have to have filtered hash_info for it.

dvc/cache/base.py
```diff
@@ -89,7 +89,27 @@ def load_dir_cache(self, hash_info):
...
+        filtered_dir_info = DirInfo()
+        try:
+            for key, value in dir_info.trie.items(hash_key):
+                filtered_dir_info.trie[key[depth:]] = value
+        except KeyError:
+            # The directory doesn't exist on the trie, fallback to the original
+            return path_info, hash_info
+        else:
+            filtered_hash_info, _ = self._get_dir_info_hash(filtered_dir_info)
+            return filter_info, filtered_hash_info
```
efiop, line 110: Note how the first return value is either `filter_info` or `path_info`(just a binary switch). There should be a nicer way to handle this.

isidentical: They both are `PathInfo`s, and I don't know what else I can return there. Maybe just raise an exception when we can't filter (just like a signal) and then use catch it right away?

dvc/cache/base.py
```diff
@@ -89,7 +89,26 @@ def load_dir_cache(self, hash_info):
...
+        depth = len(hash_key)
+        filtered_dir_info = DirInfo()
+        try:
+            for key, value in dir_info.trie.items(hash_key):
+                filtered_dir_info.trie[key[depth:]] = value
+        except KeyError:
+            return None
+        else:
+            filtered_hash_info, _ = self._get_dir_info_hash(filtered_dir_info)
+            return filtered_hash_info
```
efiop, line 109: Just a bit flatter, since we `return` in `except` anyway.
```suggestion

        filtered_hash_info, _ = self._get_dir_info_hash(filtered_dir_info)
        return filtered_hash_info
```

dvc/cache/base.py
```diff
@@ -89,7 +89,26 @@ def load_dir_cache(self, hash_info):
...
+        depth = len(hash_key)
+        filtered_dir_info = DirInfo()
+        try:
+            for key, value in dir_info.trie.items(hash_key):
+                filtered_dir_info.trie[key[depth:]] = value
+        except KeyError:
+            return None
+
+        filtered_hash_info, _ = self._get_dir_info_hash(filtered_dir_info)
+        return filtered_hash_info
```
efiop, line 109:
```suggestion
        return self._get_dir_info_hash(filtered_dir_info)[0]
```

dvc/cache/base.py
```diff
@@ -101,23 +120,33 @@ def changed(self, path_info, hash_info):
...
-            logger.debug("'%s' doesn't exist.", path_info)
+        if not self.tree.exists(path):
+            logger.debug("'%s' doesn't exist.", path)
             return True
 
         if not hash_info:
             logger.debug("hash value for '%s' is missing.", path_info)
             return True
 
+        if path != path_info:
```
efiop, line 142:
```suggestion
        if filter_info:
```

isidentical: This is not always true, since filter_info might equal to the path_info sometimes.

efiop: ah, good point!

dvc/cache/base.py
```diff
@@ -101,23 +120,33 @@ def changed(self, path_info, hash_info):
...
+            filtered_hash_info = self._get_filtered_hash_info(
+                path, hash_info, path_info
+            )
+            if filtered_hash_info is not None:
+                path_info = filter_info
```
efiop, line 147: Might be missing something, but how about we don't do this assignment and just use `path` down below to avoid confusion? 

We'll need to do something like `if filtered_hash_info is None: path = path_info` as a fallback though.

efiop: @isidentical `path_info` is coming in as an argument, so feels a bit weird to have `path` and then also overwrite `path_info`, which results in a confusion of not having to figure out if `path_info` is what came to us or is the new filtered value.

isidentical:
```
        if path != path_info:
            filtered_hash_info = self._get_filtered_hash_info(
                path, hash_info, path_info
            )
            if filtered_hash_info is None:
                path = path_info
            else:
                hash_info = filtered_hash_info
```
something like this?

efiop (approved):
Thank you!

PR #5173, opened 2020-12-28
[WIP] dvc: display command suggestions for small typos

santos22:
command line programs like "git" provide suggestions
for invalid commands that are similar to valid
commands - these changes allow for suggestions
to be displayed when invalid dvc commands are typed

Fixes https://github.com/iterative/dvc/issues/4042

---
See https://github.com/iterative/dvc/pull/4483, closed PR with lots of comments about approach and feedback (closed due to lots of comments)

The code is currently in a good state and one thing left to resolve is handling different Python versions since these changes require Python 3.9

### DIFF
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -3,6 +3,7 @@ import argparse
 import logging
 import os
 import sys
+import types
 from difflib import get_close_matches
 
 from ._debug import add_debugging_flags
@@ -273,19 +274,68 @@ def get_parent_parser():
     return parent_parser
 
 
+def add_parser_exit_on_error(self, name, **kwargs):
+    """Workaround for dynamically setting exit_on_error for subparsers on
+    Python 3.9. See more info at:
+        https://github.com/python/cpython/blob/master/Lib/argparse.py
+    """
+    if sys.version_info >= (3, 9, 0):
+        kwargs["exit_on_error"] = False
+
+    # set prog from the existing prefix
+    if kwargs.get("prog") is None:
+        kwargs["prog"] = "%s %s" % (
+            self._prog_prefix,  # pylint: disable=protected-access
+            name,
+        )  # pylint: disable=protected-access
+
+    aliases = kwargs.pop("aliases", ())
+
+    # create a pseudo-action to hold the choice help
+    if "help" in kwargs:
+        help = kwargs.pop("help")  # pylint: disable=redefined-builtin
+        choice_action = self._ChoicesPseudoAction(
+            name, aliases, help
+        )  # pylint: disable=protected-access
+        self._choices_actions.append(
+            choice_action
+        )  # pylint: disable=protected-access
+
+    # create the parser and add it to the map
+    parser = self._parser_class(**kwargs)
+    self._name_parser_map[name] = parser  # pylint: disable=protected-access
+
+    # make parser available under aliases also
+    for alias in aliases:
+        self._name_parser_map[
+            alias
+        ] = parser  # pylint: disable=protected-access
+
+    return parser
+
+
 def get_main_parser():
     parent_parser = get_parent_parser()
 
     # Main parser
     desc = "Data Version Control"
-    parser = DvcParser(
-        prog="dvc",
-        description=desc,
-        parents=[parent_parser],
-        formatter_class=argparse.RawTextHelpFormatter,
-        add_help=False,
-        exit_on_error=False,
-    )
+    if sys.version_info >= (3, 9, 0):
+        parser = DvcParser(
+            prog="dvc",
+            description=desc,
+            parents=[parent_parser],
+            formatter_class=argparse.RawTextHelpFormatter,
+            add_help=False,
+            exit_on_error=False,
+        )
+    else:
+        parser = DvcParser(
+            prog="dvc",
+            description=desc,
+            parents=[parent_parser],
+            formatter_class=argparse.RawTextHelpFormatter,
+            add_help=False,
+        )
 
     # NOTE: We are doing this to capitalize help message.
     # Unfortunately, there is no easier and clearer way to do it,
@@ -326,6 +376,10 @@ def get_main_parser():
     )
 
     fix_subparsers(subparsers)
+
+    subparsers.add_parser = types.MethodType(
+        add_parser_exit_on_error, subparsers
+    )
 
     for cmd in COMMANDS:
         cmd.add_parser(subparsers, parent_parser)
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -275,6 +275,7 @@ def get_parent_parser():
 
 
 def add_parser_exit_on_error(self, name, **kwargs):
+    # pylint: disable=protected-access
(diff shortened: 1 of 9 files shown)


dvc/cli.py
```diff
@@ -205,6 +375,10 @@ def get_main_parser():
 
     fix_subparsers(subparsers)
 
+    subparsers.add_parser = types.MethodType(
+        add_parser_exit_on_error, subparsers
+    )
```
santos22, line 380: Decided to override the behavior of `add_parser` here in one place for all subparsers.

The logic in `add_parser_exit_on_error()` is exactly the same as in the **argparse** package with the exception of the `sys.version_info >= (3, 9, 0)` check in the beginning that sets `exit_on_error=False`.

I believe this approach is cleaner than manually updating each of the `subparsers.add_parser()` calls. As for the new unit tests, the `@pytest.mark.skipif` decorator means these tests won't run when running tests in environments less than Python 3.9.

santos22:
@efiop any further thoughts on this iteration? The test failures seem to be unrelated to these changes.

dvc/cli.py
santos22: Also, this automatically handles any new DVC commands that are added. For example, the new **stage** command is displayed as a suggestion:

Screen Shot 2021-01-20 at 3 58 40 PM

dvc/cli.py
```diff
@@ -116,11 +151,96 @@ def parse_args(self, args=None, namespace=None):
...
-            self.error(msg % " ".join(argv), getattr(args, "func", None))
-        return args
+        try:
+            args, argv = self.parse_known_args(args, namespace)
+        except argparse.ArgumentError:
+            if args is None:
+                args = sys.argv[1:]
+            else:
+                args = list(args)
+            self._find_cmd_suggestions(args)
```
efiop, line 161: Let's just let the helper decide that it needs to use sys.argv, no need to do this right here.
```suggestion
            self._find_cmd_suggestions(args)
```

dvc/cli.py
```diff
@@ -154,18 +274,61 @@ def get_parent_parser():
...
+            exit_on_error=False,
+        )
+    else:
+        parser = DvcParser(
+            prog="dvc",
+            description=desc,
+            parents=[parent_parser],
+            formatter_class=argparse.RawTextHelpFormatter,
+            add_help=False,
+        )
```
efiop, line 331: No need to duplicate so much code. You could override `DvcParser.__init__` and just set `exit_on_error` there if it is supported.

tests/func/test_cli.py
```diff
@@ -200,11 +203,21 @@ class A:
         self.assertEqual(parent_dir, current_dir)
 
 
-def test_unknown_command_help(capsys):
+@pytest.mark.skipif(
+    not PY39,
+    reason="exit_on_error for command suggestions requires Python 3.9",
```
efiop, line 208: Could we mock sys.exit(or whatever argparse uses) to still test this?

dvc/cli.py
```diff
@@ -154,18 +274,61 @@ def get_parent_parser():
     return parent_parser
 
 
+def add_parser_exit_on_error(self, name, **kwargs):
```
efiop, line 277: Btw, `exit_on_error` calls `error` method of the parser, which is overriden in `DvcParser`, so seems like it shouldn't exit right now. Or am I missing something?

tests/func/test_cli.py
santos22: Are we mocking `SystemExit` only for this test? Would the following work?
```python
def test_unknown_command_help(caplog, capsys):
    if not PY39:
        with pytest.raises(SystemExit):
            _ = parse_args(["unknown"])
    try:
        _ = parse_args(["unknown"])
    ...
```

dvc/cli.py
santos22: I can move this if statement inside of `_find_cmd_suggestions` 👍

dvc/cli.py
santos22: I'm not sure I follow. `exit_on_error` calls `error` method of the parser if it is set to `True`.

```python
# parse the arguments and exit if there are any errors
if self.exit_on_error:
    try:
        namespace, args = self._parse_known_args(args, namespace)
    except ArgumentError:
        err = _sys.exc_info()[1]
        self.error(str(err))
else:
    namespace, args = self._parse_known_args(args, namespace)
```

https://github.com/python/cpython/blob/master/Lib/argparse.py#L1849

`logger.error` is then called in the `_find_cmd_suggestions` method when displaying command suggestions.

dvc/cli.py
```diff
@@ -116,11 +151,96 @@ def parse_args(self, args=None, namespace=None):
...
+                    actions = (
+                        subparser._actions  # pylint: disable=protected-access
+                    )
+                    for action in actions:
+                        if not isinstance(action.choices, dict):
+                            # NOTE: we are only interested in subcommands
+                            continue
+                        cmd_choices[cmd].extend(action.choices.keys())
+
+        return cmd_choices, hidden_cmds
```
efiop, line 202: You can simply not return hidden_cmds, just don't include them in cmd_choices

dvc/cli.py
```diff
@@ -116,11 +151,96 @@ def parse_args(self, args=None, namespace=None):
...
+        # The most similar command is
+        #         remote modify
+        if len(args) == 2 and args[0] in cmd_choices:
+            sub_cmd_choices = cmd_choices[args[0]]
+            if sub_cmd_choices and args[1] not in sub_cmd_choices:
+                cmd_suggestion_msg = _get_cmd_suggestion_msg(
+                    args[1], sub_cmd_choices, args[0]
+                )
+                logger.error(cmd_suggestion_msg)
+                raise DvcParserError
```
efiop, line 243: This will fallthrough if len(args) is bigger than 2 and will not raise a error up the stack.

dvc/cli.py
efiop: You don't need exit_on_error at all, our `DvcParser.error()` will not sys.exit, just catch DvcParserError instead of ArgumentError

tests/func/test_cli.py
efiop: We don't need to mock SystemExit, DvcParser doesn't use it.

efiop (commented):
Current implementation both prints help and then git-like message about similarity, which doesn't look good. The ui and the code are still problematic here.

Looks like this will require quite a bit more time to get right. So far we've spent around half a year for the previous PR iteration and this one combined and I feel bad for you spending so much time on this :slightly_frowning_face: We are also spending time on reviewing this and I don't think this is close to being mergeable and maintainable. I don't think it is worth spending more of your time on this, maybe we'll get back to properly implementing it in the future.

santos22:
@efiop yea I think that makes sense. No problem for me since I was able to learn more about building a command line experience and more about pytest. Closing this for now and hopefully it gets shipped in the future. Thanks for all of the feedback!

Issue #4604, opened 2020-09-23
Local config user respected in "dvc pull" but not "dvc list"


Erotemic:
## Bug Report

My computer username (joncrall) is different than my credentials that I use to access the dvc remote cache. In my local config I've modified the dvc remote to register my appropriate remote username (jon.crall). 

Running: 

```
dvc list -v git@myrepo.com/myrepo.git path/to/data
```

I get: 
```
2020-09-23 16:26:45,710 DEBUG: Collecting information from remote cache...                                                            
2020-09-23 16:26:45,712 DEBUG: Establishing ssh connection with '<internal-remote-store>' through port '22' as user 'joncrall'                       
2020-09-23 16:26:46,412 ERROR: unexpected error - No existing session                                                                 
```
 and then there is a big long `paramiko.ssh_exception.SSHException` that ultimately stems from the fact that it did not pick up my usename in my local config, which looks like this:

```
['remote "public-storage"']
    user = jon.crall
```

For reference my normal dvc/config looks like:

```
[cache]
    type = "reflink,symlink,copy"
[core]
    remote = public-storage
['remote "public-storage"']
    url = ssh://an/internal/url/dvc-caches/public
```

However, if I use something like: `dvc pull -v` it works 

```
2020-09-23 16:31:15,614 DEBUG: Establishing ssh connection with '<internal-remote-store>' through port '22' as user 'jon.crall'                      
```

Also, if I modify my global $HOME/.cache/dvc/config to include the custom "user" it works. 

### Please provide information about your setup

**Output of `dvc version`:**

```
DVC version: 1.7.4+e1b344.mod 
---------------------------------
Platform: Python 3.8.2 on Linux-5.4.0-47-generic-x86_64-with-glibc2.10
Supports: http, https, ssh
Cache types: hardlink, symlink
Cache directory: ext4 on /dev/sdd1
Workspace directory: ext4 on /dev/sdd1
Repo: dvc, git
```
I've also tested on pypi latest of 1.7.4

I'm currently trying to trace where local credentials are loaded in the pull command and see if I can port that to the list command, but I figured Id make an issue and see if this is a known bug first.

Erotemic:
[...]

Issue #3590, opened 2020-04-04
`dvc list`: handle local repos differently?


ahmed-shariff:
When I run the command `dvc list .` from any sub-directory of the project I get the following error:
```
ERROR: failed to list '.' - Failed to clone repo '.' to '/tmp/tmp2qmfnem7dvc-clone': Cmd('git') failed due to: exit code(128)
  cmdline: git clone --no-single-branch -v . /tmp/tmp2qmfnem7dvc-clone
  stderr: 'fatal: repository '.' does not exist
'
```

Though it works when executed from the root directory of the project

DVC: 0.91.1 (arch linux;pip)

----------

UPDATED (@shcheklein): 

repurposed it a bit - https://github.com/iterative/dvc/issues/3590#issuecomment-612558038

efiop:
Hi @ahmed-shariff !

dvc list expects to receive a URL to a git repo, which `.` isn't when you are not in the git repo root. Same as you can't `git clone` that directory. Theoretically, we could check if URL that you pass is a git repo subdir, but I'm not sure if it is worth the effort and it also will probably lead to misuse, where people would try to use it as `ls` in their subdirs.

ahmed-shariff:
I see. Thank you for the clarification.

shcheklein:
To be honest I think it makes sense to handle this as a special case:

* don't clone
* if URL is a local path that is part of the DVC repo - show the result of `dvc list <path-to-Git-root> URL`. It means if I run `dvc list .` I just see the files in the current location - it is probably the most expected result.

@iterative/engineering @casperdcl thoughts?

(reopening, since it's annoying to remember the special syntax when I deal with the local repo, and other users were caught by surprise)

casperdcl:
@shcheklein I agree

jamessergeant:
I'm using a local remote and I receive the same error when running `dvc list <path-to-local-remote>`. Is this expected behaviour?

For reference the "local remote" is actually a mounted network drive.

jorgeorpinel:
Hi @jamessergeant, `dvc list` expects the path or URL to the DVC repository itself, not to a remote storage location.

jorgeorpinel:
In fact I believe it doesn't check remotes at all to produce the list. Whether the data exists in remote storage is not guaranteed by `dvc list`. You have to attempt `dvc get` or `dvc import` to find out.

jorgeorpinel:
[...]

PR #5174, opened 2020-12-29
erepo: make repo configuration explicit

efiop:
A more explicit way to deal with cache/remote adjustments that we do for external repos.

Pre-requisite for #3590 and #4604

### DIFF
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -8,7 +8,6 @@ from typing import Dict, Iterable
 
 from funcy import cached_property, reraise, retry, wrap_with
 
-from dvc.cache import Cache
 from dvc.config import NoRemoteError, NotDvcRepoError
 from dvc.exceptions import (
     DvcException,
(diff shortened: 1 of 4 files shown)


dvc/external_repo.py
```diff
@@ -87,6 +86,30 @@ def clean_repos():
...
+            # original repo's cache location.
+            name = "auto-generated-upstream"
+            return {
+                "core": {"remote": name},
+                "remote": {name: {"url": repo.cache.local.cache_dir}},
+            }
+
+        # Use original remote to make sure that we are using correct url,
+        # credential paths, etc if they are relative to the config location.
+        return {"remote": {name: repo.config["remote"][name]}}
```
efiop, line 108: Previously we only fixed the url, but we actually want paths to keys and credentials too (e.g. there might be something set in original .dvc/config.local, that it makes sense to re-use as well)

dvc/external_repo.py
```diff
@@ -102,19 +125,29 @@ def __init__(
...
-        self.cache_dir = cache_dir or self._get_cache_dir()
-        self.cache_types = cache_types
-
-        self._setup_cache(self)
-        self._fix_upstream(self)
         self.tree_confs = kwargs
 
+        self._cache_config = {
+            "cache": {
+                "dir": cache_dir or self._get_cache_dir(),
```
efiop, line 134: Btw, we are and have been using the same cache dir for the main repo and its subrepos. Looks intentional, but just want to put a mind note...

dvc/external_repo.py
```diff
@@ -102,19 +125,29 @@ def __init__(
...
+        if os.path.isdir(url):
+            config.update(_get_remote_config(url))
+
+        super().__init__(
+            root_dir,
+            scm=scm,
+            rev=rev,
+            uninitialized=uninitialized,
+            config=config,
+        )
```
efiop, line 149: Looks like this could be unified with `make_repo` factory. But that's for the followup pr.

dvc/external_repo.py
```diff
@@ -218,58 +251,13 @@ def _get_tree_for(self, repo, **kwargs):
...
-            src_repo.close()
+        config = self._cache_config.copy()
+        if os.path.isdir(self.url):
+            rel = os.path.relpath(path, self.root_dir)
+            repo_path = os.path.join(self.url, rel)
+            config.update(_get_remote_config(repo_path))
+        return Repo(path, scm=self.scm, rev=self.get_rev(), config=config)
 
     @wrap_with(threading.Lock())
     def _get_cache_dir(self):
```
efiop, line 263: This should be out of this class, it is more about the url caching mechanism than ExternalRepo itself. Will likely get revisited when introducing global persistent git repo cache.

skshetry (approved):
Looks great.

PR #5175, opened 2020-12-29
experiments: fix checkout error for unexisting outputs

### LINKED ISSUE #5155: exp: tries to checkout not existing output
# Bug Report

```bash
#! /bin/bash

# init
cd "$(mktemp -d)"
git init && dvc init
git add -A
git commit -m "initial commit"

# track dependency
echo "foo" > foo
dvc add foo
git add -A
git commit -m "second commit"

# creating YAML file
dvc run --no-exec -n copy-foo-bar -d foo -o bar "cp foo bar"

# experiments run
dvc exp run
```

### Expected:
command is run and `bar` is produced which is then checked out.

### Actual:
```
ERROR: Checkout failed for following targets:                                                             
bar
Is your cache up to date?
```

Maybe I am doing something wrong here?

pmrowla:

Will fix #5155.

* Unexisting outputs will be ignored on initial checkout (before experiment reproduction)
* Staged Git changes will be reset prior to running experiments in workspace (otherwise stash merge conflicts may occur if a staged file was previously an untracked file).

### DIFF
--- a/dvc/repo/experiments/__init__.py
+++ b/dvc/repo/experiments/__init__.py
@@ -328,6 +328,15 @@ class Experiments:
 
     def reproduce_one(self, queue=False, tmp_dir=False, **kwargs):
         """Reproduce and checkout a single experiment."""
+        if not (queue or tmp_dir):
+            staged, _, _ = self.scm.status()
+            if staged:
+                logger.warning(
+                    "Your workspace contains staged Git changes which will be "
+                    "unstaged before running this experiment."
+                )
+                self.scm.reset()
+
         stash_rev = self.new(**kwargs)
         if queue:
             logger.info(
--- a/dvc/repo/experiments/base.py
+++ b/dvc/repo/experiments/base.py
@@ -102,7 +102,7 @@ class ExpRefInfo:
         self, baseline_sha: Optional[str] = None, name: Optional[str] = None,
     ):
         self.baseline_sha = baseline_sha
-        self.name = name
+        self.name: str = name if name else ""
 
     def __str__(self):
         return "/".join(self.parts)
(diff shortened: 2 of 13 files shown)


skshetry:
@pmrowla, looks like there's a minor UI issue, when I run the script mentioned in https://github.com/iterative/dvc/pull/5175#issue-546345699.
```console

	git add /tmp/tmp.5rCmTvomI8/dvc.yaml /tmp/tmp.5rCmTvomI8/foo dvc.lock
 
Reproduced experiment(s): exp-a34f1

	dvc exp branch <exp>
```

skshetry:
@pmrowla, when I change the script from `dvc exp run` to `git add dvc.yaml && dvc exp run --temp`, there's no lock file at all. I have to force the `apply` to bring the results, but the help message for `--temp` suggests that it should by default apply unless `--queue` is specified.

>  --temp                Run this experiment in a separate temporary
                        directory instead of your workspace. Only applies
                        when running a single experiment without --queue.

pmrowla:
> @pmrowla, when I change the script from `dvc exp run` to `git add dvc.yaml && dvc exp run --temp`, there's no lock file at all. I have to force the `apply` to bring the results, but the help message for `--temp` suggests that it should by default apply unless `--queue` is specified.
> 
> > --temp                Run this experiment in a separate temporary
> > directory instead of your workspace. Only applies
> > when running a single experiment without --queue.

This is expected behavior. When we run in temp directories (via `--temp` or `--queue`), the results are not automatically applied to the user's workspace.

pmrowla:
If `dvc.yaml` is untracked and is not staged, it is expected that the command will fail, as untracked files are not copied into the temp directory before running (so `dvc.yaml` will not exist in the temp directory). When running in temp directories, new files need to be staged using `git add` before running the experiment.

skshetry:
> If dvc.yaml is untracked and is not staged, it is expected that the command will fail

@pmrowla, would it be possible to check this and error out like how `pre-commit` does?
```console
$ pre-commit run                  
[ERROR] Your pre-commit configuration is unstaged.
`git add .pre-commit-config.yaml` to fix this.
```
`pre-commit` might not be the best example, but just wanted to demo the error message.

Issue #4742, opened 2020-10-19
Make that `dvc repro --pull` pulls all missing files.


peper0:
According to help for `dvc repro`:
```
  --pull                Try automatically pulling missing cache for outputs
                        restored from the run-cache.
```
and that's what it does. Pulls missing files that are **outputs restored from the run-cache**. But if there are outputs missing from "sources" (i.e. dvc files having only output and no command at all), it won't pull them. These must be pulled separately with a `pull` command.

Why not change it into "pull whatever is missing and necessary for this repro"? Is there any use case where a user want to download automatically some missing files, but not all of them?

efiop:
Hi @peper0 !

It is the plan :slightly_smiling_face: For now it is limited to run-cache, but we'll add support for data sources soon. Stay tuned! :slightly_smiling_face:

dberenbaum:
Comments from https://github.com/iterative/dvc/issues/4871:

> Currently the `--pull` flag have no impact when used together with the `--downstream` flag. When working in a new repo, one must pull manually all the `deps` of stage `A` (if there is too much data handled with dvc to run `dvc pull`) before running `dvc repro --downstream A`. It would be great if `dvc repro --downstream --pull A` would automatically pull `A` deps.
> 
> Linked to #4870 Discord discussion: https://discord.com/channels/485586884165107732/485596304961962003/775412039321387008

> Thinking more about it it would be great if `--pull` would work with `-s|--single-item` too.

dberenbaum:
Another request in discord: https://discord.com/channels/485586884165107732/485596304961962003/945713992825991188

francesco086:
Hi, I am leaving my comment as a user as suggested by @dberenbaum, I hope it can be helpful.

Let's take the example A -> B -> C mentioned above. After defining the pipeline I want to run each stage in a dedicated container using an orchestrator like Argo (or Kubeflow or Airflow or ...).
So, for example, I will have a step in Argo where I want to execute the stage B. Then I would like to be able to do somthing along these lines:
```sh
git clone git-repo-containing-the-dvc-files
cd git-repo-containing-the-dvc-files
dvc repro B --new-flag
dvc add -R .
dvc push -R .
git add -A .
git commit -m "executed stage B"
git push
```
(`--new-flag` could be something like `--pull-direct-dependency-only`)

The command `dvc repro B --new-flag` should ideally pull (download) `a.json` (output of stage A) only if it changed, and it should run the stage B again only if there was any change.

[...]

PR #5185, opened 2020-12-31
run-cache: skip unhashable stages

### LINKED ISSUE #5182: run-cache: `restore` does not check if stage has outs or not
```bash
#! /bin/bash

set -e
set -x

cd "$(mktemp -d)"
git init && dvc init

echo 'var: foo' > params.yaml

# create a stage without any outputs
dvc run -v -n echo -p var "echo hello world"
 
# change `cmd` so that it's not cached
# and, `-f` to not throw duplicate stage error
dvc run -f -v -n echo -p var "echo hello" 
```

#### Analysis
This happens because it seems we are not skipping if the stage does not have any outputs.

efiop:
Fixes #5182

### DIFF
--- a/dvc/stage/cache.py
+++ b/dvc/stage/cache.py
@@ -34,18 +34,26 @@ def _get_cache_hash(cache, key=False):
     )
 
 
-def _get_stage_hash(stage):
-    if not (stage.cmd and stage.deps and stage.outs):
-        return None
+def _can_hash(stage):
+    if stage.is_callback or stage.always_changed:
+        return False
+
+    if not all([stage.cmd, stage.deps, stage.outs]):
+        return False
 
     for dep in stage.deps:
         if not (dep.scheme == "local" and dep.def_path and dep.get_hash()):
-            return None
+            return False
 
     for out in stage.outs:
         if out.scheme != "local" or not out.def_path or out.persist:
-            return None
+            return False
 
+    return True
+
+
+def _get_stage_hash(stage):
+    assert _can_hash(stage)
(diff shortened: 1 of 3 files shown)


tests/unit/stage/test_cache.py
```diff
@@ -208,14 +208,22 @@ def _mode(path):
...
+)
+def test_unhashable(tmp_dir, dvc, mocker, kwargs):
+    from dvc.stage import Stage, create_stage
     from dvc.stage.cache import RunCacheNotFoundError, StageCache
 
-    repo = mocker.Mock(spec=Repo)
-    cache = StageCache(repo)
-    stage = Stage(repo, always_changed=True)
+    cache = StageCache(dvc)
+    stage = create_stage(Stage, path="stage.dvc", repo=dvc, **kwargs)
```
efiop, line 226: Too lazy to mock this out perfectly (as in many other tests). We need some nice factory for stages to use in tests :slightly_frowning_face:

dvc/stage/cache.py
```diff
@@ -167,7 +172,7 @@ def save(self, stage):
         self.tree.move(PathInfo(tmp), path)
 
     def restore(self, stage, run_cache=True, pull=False):
-        if stage.is_callback or stage.always_changed:
+        if not _can_hash(stage):
```
efiop, line 175: Better approach here would be to try to checkout/pull anything that is up-to-date(e.g. `dvc add`ed files and `always_changed` stages, but that needs to be done on a higher level. Will be resolved in https://github.com/iterative/dvc/issues/4742

Issue #1407, opened 2018-12-04
run: abandon `callback stage` feature in favor of `--always-changed`


efiop:
We had `--always-changed` for a long time now(since before 1.0), so we can drop support for callback stages (see Stage.is_callback).

PR #5187, opened 2020-12-31
dvc: redefine is_callback as cmd and no outs/deps

efiop:
Old cmd + no deps definiton has been obsoleted since 1.0 in favor
of --always-changed and doesn't affect our ability to read history for
metrics/plots/etc so we are safe to drop the support for 2.0 onwards.

Fixes https://github.com/iterative/dvc/issues/1407

We don't have any mention of callbacks since 1.0 so no changes for docs required.

### DIFF
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -228,7 +228,7 @@ class Stage(params.StageParams):
         A callback stage is always considered as changed,
         so it runs on every `dvc repro` call.
         """
-        return not self.is_data_source and len(self.deps) == 0
+        return self.cmd and not any((self.deps, self.outs))
 
     @property
     def is_import(self):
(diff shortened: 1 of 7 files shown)


efiop:
Though this might still be useful for cmd-only stages. Or maybe we should error-out in that case and require `always_changed` to be set, as cmd-only stage is probably a mistake in most cases :thinking:

efiop:
Thinking some more about it, it makes sense to get rid of is_callback, but actually forbid cmd with no deps/outs, as that is a 100% an error.

efiop:
The harmful behavior whas the fact that having no deps caused always_changed-like behavior, which could often be undesired. The case where we don't have outs or deps is still a corner-case though, where we can assume either that such stage is always changed or always unchanged. Both are useless, but the former one makes more sense to me and that is something that I would expect from dvc. We could consider forbidding `dvc stage add` with no deps/outs later though, but I don't remember people running into this for a long time now, so probably not worth bothering with.

Issue #5157, opened 2020-12-24
init: maybe `dvc init` should display that dvc repository was initialized


skshetry:
```console
$ dvc init
You can now commit the changes to git.

What's next?
------------
- Check out the documentation: <https://dvc.org/doc>
- Get help and share ideas: <https://dvc.org/chat>
- Star us on GitHub: <https://github.com/iterative/dvc>
```

Should we add information regarding what was initialized?

For example, when I do `git init`, it displays following information:

```console
$ git init
Initialized empty Git repository in /tmp/tmp.jJg5TbJ68v/.git/
```

Maybe, we could add similar information at the top line?
```console
$ dvc init
Initialized DVC repository. You can now commit the changes to git.

...
```

PR #5190, opened 2021-01-02
show more info when successfully init dvc repo

Njuhobby:
solving issue: https://github.com/iterative/dvc/issues/5157


dvc/repo/init.py
```diff
@@ -84,6 +84,10 @@ def init(root_dir=os.curdir, no_scm=False, force=False, subdir=False):
 
     if scm.ignore_file:
         scm.add([os.path.join(dvc_dir, scm.ignore_file)])
-        logger.info("\nYou can now commit the changes to git.\n")
```
skshetry, line 87: This will break `dvc init --no-scm` case. Let's print this message as-is and put the "initialized" message before this `if`.

skshetry: ping @Njuhobby

dvc/repo/init.py
```diff
@@ -82,6 +82,8 @@ def init(root_dir=os.curdir, no_scm=False, force=False, subdir=False):
         [config.files["repo"], dvcignore, proj.plots.templates.templates_dir]
     )
 
+    logger.info("\nInitialized DVC repository.\n")
```
skshetry, line 85: Let's not add a newline before this message.
```suggestion
    logger.info("Initialized DVC repository.\n")
```

dvc/repo/init.py
Njuhobby: Nice. Didn't realize dvc init could run in no-scm environment. Was wondering why the message was put under the if statement.

dvc/repo/init.py
```diff
@@ -82,6 +82,8 @@ def init(root_dir=os.curdir, no_scm=False, force=False, subdir=False):
         [config.files["repo"], dvcignore, proj.plots.templates.templates_dir]
     )
 
+    logger.info("Initialized DVC repository.\n")
+
     if scm.ignore_file:
         scm.add([os.path.join(dvc_dir, scm.ignore_file)])
         logger.info("\nYou can now commit the changes to git.\n")
```
skshetry, line 89: This will make it print two lines between these messages. Let's remove this too.
```suggestion
        logger.info("You can now commit the changes to git.\n")
```

skshetry (approved):
Looks good, @Njuhobby. Thanks a lot for contributing to the DVC. 🙏🏽

Issue #5025, opened 2020-12-04
Consistency of desc fields in dvc.yaml and .dvc files


jorgeorpinel:
> Extracted from https://github.com/iterative/dvc.org/pull/1951#pullrequestreview-532832838

The new `desc` field made me rethink the ability to have user data in DVC files. Expand below for the current options (this emoji 👉 points to unique characteristics of each one):

<details>

**YAML comments**

- 👉 Part of the YAML format anyway, I guess
- Only meaningful to the user
- DVC should ignore (but preserve) these.

**`meta` field**

- A valid field in dvc.yaml, at the stage level — part of its [schema](https://github.com/iterative/dvcyaml-schema/blob/master/schema.json)
- A valid field in .dvc files, 👉 at the file level (these files were considered stages in 0.x)
- 👉 Accepts any YAML structure.
- Only meaningful to the user
- DVC should ignore (but preserve) these.

**`desc` field**

- A valid field in dvc.yaml, at the stage 👉 and/or output levels — part of its [schema](https://github.com/iterative/dvcyaml-schema/blob/master/schema.json)
- A valid field in .dvc files, 👉 at the output level (I suppose these files are considered something closer to outputs in 1.0+ ?)
- 👉 Only accepts strings.
- Only meaningful to the user
- 👉 `dvc add/run --desc` can be used to write this.
- DVC ignores (but preserves) it otherwise?
- 👉 DVC Viewer requires it? See https://github.com/iterative/viewer/issues/1371#issuecomment-738549604

</details>

### TL;DR

I see significant overlap. What are we trying to solve and what's the ideal design to implement it?

> At the same time these are probably the least relevant fields in the file formats but still, I suspect at least one of these options is redundant.

pmrowla:
I think the reasoning is that `desc` is not only relevant to the user, it's also relevant to the viewer. The field can specifically be used to add a text description to an output, that will be displayed in the viewer.

`meta` can contain literally anything and generally only means anything to the user that put something into the `meta` field, so it does not make sense for the viewer to try and figure out what is inside `meta` and then attempt to display whatever it might be.

efiop:
Doesn't look like there is anything left to say here. Those are just different by definition. Closing as stale, happy to reopen if anyone else has to say anything here.

jorgeorpinel:
> desc is not only relevant to the user, it's also relevant to the viewer
> meta can contain literally anything and generally only means anything to the user

[...]

PR #5191, opened 2021-01-03
Add stages command

### LINKED ISSUE #3743: Quick listing of stages
Before [last release](https://github.com/iterative/dvc/releases/tag/1.0.0a0), it was handy to reproduce my stages due to easy filename completion in the command line, due to the stages being files in the directory. All I had to do was to type `dvc repro`, start typing the name and hit TAB.

Now, stages are fields in a YAML-formatted single file. If I don't know **exactly** the stage name, I must open the file, look for the part of the file where the stage name that I am looking for is located at, and check or copy-paste from there. Then leave the file and type.

The ideal feature would be to auto-complete with TAB, just like before, but this can be outside the technical scope of DVC (a dirty fix would be to have empty files with stage names, but I don't think that's a good solution...). Therefore, I think a feature that could improve usability would be listing of stages. There could be a new option in the **repro** command such as `dvc repro -l`. This command would parse the dvc.yaml and list the stage names so that the user could type them by seeing the desired stage name on the same screen since it has just been printed out.

skshetry:
Resolves #3743.

		Not looking into adding a docs yet, keeping this command as experimental for some time.

Screenshot from 2021-01-09 10-17-22


dvc/command/stages.py
```diff
@@ -0,0 +1,312 @@
...
+        for idx, column in enumerate(title_table.columns):
+            if not idx:
+                continue
+            column.no_wrap = True
+            column.justify = "right"
+        yield title_table
+        yield ""
+
+        # Table of ancestor nodes
+        nodes_table = Table.grid(padding=(0, 1), expand=True)
```
skshetry, line 203: Not sure if we need to so DAG, or just the next-to nodes. We could change it later.

dvc/command/stages.py
```diff
@@ -0,0 +1,312 @@
...
+            title.append(f":{stage.name}")
+    else:
+        title.append(address, style=file_style)
+    return title
+
+
+def prepare_description(stage: Stage) -> Union[Markdown, Text]:
+    # Description
+    if stage.desc:
+        # not sure if this was ever intended, but markdown looks nice
```
skshetry, line 159: I have raised a question in https://github.com/iterative/dvc/issues/5025#issuecomment-753631328 regarding the intention of `desc`.

dvc/command/repro.py
```diff
@@ -51,7 +51,7 @@ def add_arguments(repro_parser):
         "targets",
         nargs="*",
         help=f"Stages to reproduce. '{PIPELINE_FILE}' by default.",
-    ).complete = completion.DVC_FILE
+    ).complete = completion.DVCFILES_AND_STAGE
```
skshetry, line 54: Adding this only to the `repro` for now. The performance is really bad, considering the reason I would do a completion for is to be faster.

pared:
@skshetry do we intend to have tests for `stages`?

skshetry:
@pared, the autocompletion-related feature `stages -sq` is a very thin wrapper on `dvc.repo.stage`. The long-form list output is mostly presentational with very little logic and is experimental for now. I'm sure this will change. Till now, I was just toying with `rich` and its feature, so I haven't thought much about the information that we should display there.

The feedback on that and the design would be very much appreciated. Thanks.

pared:
@skshetry Thanks,
So far it seems very clear and understandable. I'll play around with it and see if I can come up with something useful.

pared (approved):
Support for markdown in description looks great!

pmrowla (approved):
LGTM.

Only comment would be that having the entire tabled centered instead of left-aligned looks off to me in a wider terminal window, but that's a UI personal preference thing.

Screen Shot 2021-01-05 at 5 18 30 PM

dvc/command/stages.py
```diff
@@ -0,0 +1,306 @@
...
+            nodes_table.add_row(Padding(nodes_text, LEFT_PAD))
+            yield nodes_table
+            yield ""
+
+        desc = prepare_description(stage)
+        yield Padding(desc, LEFT_PAD)
+
+    def column(renderable):
+        """Constrain width and align to center to create a column."""
+        return Align.center(renderable, width=LAYOUT_WIDTH, pad=False)
```
skshetry, line 209: @pmrowla, could you please try changing this `center` function to `left` and see? 
Do you prefer it that way? How does it look? I have no strong opinion on the centering.

pmrowla: I think it looks better with it `left` aligned.

Screen Shot 2021-01-05 at 5 33 09 PM

skshetry: @pmrowla, done. Thanks for the suggestion.

dvc/command/stages.py
```diff
@@ -0,0 +1,324 @@
...
+        if nodes_text:
+            nodes_table.add_row(Padding(nodes_text, LEFT_PAD))
+            yield nodes_table
+            yield ""
+
+        desc = prepare_description(stage)
+        yield Padding(desc, LEFT_PAD)
+
+    def column(renderable):
+        """Constrain width and align to center to create a column."""
```
skshetry, line 209:
```suggestion
        """Constrain width and align column to the left."""
```

dvc/command/stages.py
skshetry: Talked with Dmitry on this. He said that the dependent nodes are a good default choice. We will come to it later if needed.

dvc/command/repro.py
skshetry: The performance isn't _that_ bad. 😄  At least in our example-get-started repo.

shcheklein:
Looks great overall! I have a few concerns though to be honest:

1. A separate command - it feels if overlaps with `dvc dag` to some extent. Even if not, do we actually need the whole new command for this? Should we merge `dag` into this one `dvc stages --dag`?
2. Interface looks refreshing, but a bit too massive to my mind. Feels like `-s` can be default? And even if we add more details - let's add them as `ls` does (compact table, parsable with regular tools like awk - it's quite important to my mind)
3. It's better to always provide `--show-json`, or something similar for machine readable interface. So that UI does not become an API (as much as possible). In case `awk` parsable table is possible, it might we okay to do both?

skshetry:
>     1. A separate command - it feels if overlaps with `dvc dag` to some extent. Even if not, do we actually need the whole new command for this? Should we merge `dag` into this one `dvc stages --dag`?

`dag` is more of a visual command whereas my intention for this command is to just list stages in the dvc.yaml. Anything other than that is optional, the command is only trying to be helpful.

There was a discussion on showing `status` here or move this to the `dvc status`. But I am not sure about this command in the longer term or whether we should move them to `status`/`dag` or move them here. 

>     2. Interface looks refreshing, but a bit too massive to my mind. Feels like `-s` can be default? 

Can that be the other way around? Encourage people to use `--short`/`-s` for parsing/grepping? 

> And even if we add more details - let's add them as `ls` does (compact table, parsable with regular tools like awk - it's quite important to my mind)

`ls` might not work as the columns might vary based on description/outputs/dependent nodes/dependencies, etc. 
Unless we choose to not show them at all and just show the numbers of them, it will be difficult to parse anyway.

>     3. It's better to always provide `--show-json`, or something similar for machine readable interface. So that UI does not become an API (as much as possible). In case `awk` parsable table is possible, it might we okay to do both?

`-s` should be machine-readable. The current format that's required for autocompletion could be changed to a different option and we could just print out in the following format:
```
# <name>:<sp><desc>
train: Train *model*
evaluate: Evaluate *model*
```
And, we can easily provide all of the `stage` internals through `--show-json`. I have no issues with that.

Alternately, we could make `repro`/`exp` return a list of the stages, as a help command for the stages.

shcheklein:
> dag is more of a visual command

but it's not like it has to be, right? also, we can make `--dag` option. It just feels that we have to much overlap, and it's hard to justify the whole top level command for this.

> Could that be the other way around? Encourage people to use --short/-s for parsing/grepping?

I don't know. It's my personal bias I guess from regular CLI tools, like ls. It still feels we can make a bit more compact. Also if we can make it parsable by default it's better to my mind. We should have strong reasons to not do so.

> ls might not work as the columns might vary based on description/outputs/dependent nodes/dependencies

skip values? not sure, would need to see the whole table schema I guess

> -s should be machine-readable.

agreed, if we can make regular output grep/awk friendly - better do that. But `--show-json` I mention for consistency, an interface we guarantee consistenct?

> And, we can easily provide all of the stage internals through --show-json. I have no issues with that.

👍 

> Alternately, we could make repro/exp return a list of the stages, as a help command for the stages.

make sense, might be an option as well

skshetry:
> but it's not like it has to be, right?

I'll think about it. But, let's leave it for now? WDYT?

@shcheklein, `-s` is now machine-readable. For autocompletion in bash, I introduced `--list-names` and for `zsh`, instead of parsing in auto-completion scripts, I introduced `--compact` flag. PTAL.

> 
> > And, we can easily provide all of the stage internals through --show-json. I have no issues with that.

Let's wait for someone to ask for, I think.

shcheklein:
@skshetry I have some bad feeling about this to be honest. In this specific case `for now` doesn't apply well since we are introducing the whole new command, an option which provides certain API for other parts, etc. 

Also, DVC 2.0 is a very good moment to deprecate/break a bit things.

> Let's wait for someone to ask for, I think.

yep, I guess we can wait. 

@iterative/dvc @dmpetrov what are thoughts on this, guys?

skshetry:
> @skshetry I have some bad feeling about this to be honest. In this specific case `for now` doesn't apply well since we are introducing the whole new command, an option which provides certain API for other parts, etc.

But, it's main role is to list stages, right? which it does. I am not sure how it can be a separate subcommand/mixed with other, except `dvc repro list`, I think. Other commands have a different meaning/usecase.

skshetry:
@shcheklein, I intend to move ahead with this for now. As I have said it in the description, the list-layout is experimental. The documented feature of this is the autocompletion (we can improve upon the flags till 2.0 release if needed). If we decide to remove or make it a subcommand by release, I'm happy to do it. Otherwise, I'll keep it undocumented it for now.

shcheklein:
@skshetry sounds good if it's hidden, but I also don't quite understand why are we rushing here (can't wait for other folks to review/brainstorm). It's not something that blocks anything, right?

skshetry:
> @skshetry sounds good if it's hidden, but I also don't quite understand why are we rushing here (can't wait for other folks to review/brainstorm).

I have some performance improvements on top of this (completion is slow, as the startup time is not that great), but I'll do it anyway without this, I guess.

jorgeorpinel:
Woa, interesting!

- I agree that the centering is unexpected (feel like a retro DOS app which is fun but probably too much?)
- Agree about `-s` default and more compact details
- Agree that `dag` is a separate thing (as long as it makes sense that this command doesn't validates DAGs).
- Cool md support but maybe add a flag to enable that (otherwise truncate `desc` heavily).
- ➕1 on all stage details via --show-json (if requested)
- `dvc repro list` feels like something `repro` shouldn't worry about.

> The documented feature of this is the autocompletion

What docs changes does that entail? Thanks

jorgeorpinel:
> the list-layout is experimental

On this. I'm not sure about listing deps and outs (or even more details) at least initially (until someone asks for it?). You could cat dvc.yaml for that (mentioned in https://github.com/iterative/dvc/issues/3743#issuecomment-757087326). We "sell" dvc.yaml as human-friendly so I think we expect users to be OK with parsing simple YAML.

skshetry:
> * I agree that the centering is unexpected (feel like a retro DOS app which is fun but probably too much?)

UI is no longer centered. The demo is a bit old.

skshetry:
> We "sell" dvc.yaml as human-friendly so I think we expect users to be OK with eye-parsing simple YAML.

Tell this to a user with 100 stages/~1300 lines of `dvc.yaml`.

It's human-friendly in terms of readability and writability and that too applies only to a certain extent, it does not enlighten the user as to their relationship/purpose just by looking at it.

jorgeorpinel:
> a user with 100 stages/~1300 lines of dvc.yaml

Is there a real life example of this out there, and most importantly has it been reported/requested by a user with such huge dvc.yaml files? In any case how does this command help with that? The output is even longer than the DVC files.

> It's human-friendly in terms of readability

Yes, that's my point.

> it does not enlighten the user as to their relationship/purpose just by looking at it

OK but again, how does this command help with that? It basically reformats the same info from dvc.yaml. It doesn't make any "enlightening" interpretations.

jorgeorpinel:
As a reminder, I'm not arguing against the command in general. I'm questioning the value of showing all the details from dvc.yaml (except in `--show-json`, as a programmatic interface to out dvc.yaml format) until we have evidence of this need from users. But if that exists and I'm not aware of it, maybe you can help us by sharing the full context?

jorgeorpinel:
p.s. the purpose in #3743 doesn't seem to require full stage details IMO, including the dvc.yaml location seems enough.

jorgeorpinel:
p.p.s. any details that can fit in a very compact manner (one line) I wouldn't see a problem with, if the output remains easily readable.

PR #5194, opened 2021-01-04
dvc: handle multiline command as list

### LINKED ISSUE #4373: Support multiline command?

efiop:
Implements an approach proposed by @skshetry.

Fixes #4373

https://github.com/iterative/dvc.org/pull/2071

### DIFF
--- a/dvc/stage/run.py
+++ b/dvc/stage/run.py
@@ -47,7 +47,7 @@ def warn_if_fish(executable):
 
 def _enforce_cmd_list(cmd):
     assert cmd
-    return cmd if isinstance(cmd, list) else [cmd]
+    return cmd if isinstance(cmd, list) else cmd.splitlines()
 
 
 def prepare_kwargs(stage, checkpoint_func=None):
--- a/tests/func/test_repro_multistage.py
+++ b/tests/func/test_repro_multistage.py
@@ -529,7 +529,14 @@ def test_repro_multiple_params(tmp_dir, dvc):
     assert dvc.reproduce(stage.addressing) == [stage]
 
 
-def test_repro_list_of_commands_in_order(tmp_dir, dvc):
+@pytest.mark.parametrize("multiline", [True, False])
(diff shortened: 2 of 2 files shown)


dvc/stage/run.py
```diff
@@ -47,7 +47,7 @@ def warn_if_fish(executable):
 
 def _enforce_cmd_list(cmd):
     assert cmd
-    return cmd if isinstance(cmd, list) else [cmd]
+    return cmd if isinstance(cmd, list) else cmd.splitlines()
```
skshetry, line 50: A minor thing: What should we do in the `dvc.lock` file? Should we keep them in a single format (like either a list or a multiline?)? Just that when the user changes the multiline string to a list or vice versa, it should not affect the lock file.

It might be too much to support that way, but just saying that in principle, it'd be better to have a single format in the lock file, but I think it happens rarely in practice and the list and the multiline string might never be equivalent when user refactors them.

skshetry (approved):
What were the recent motivations for this? 🙂  Anyway, this is good.

dvc/stage/run.py
efiop: @skshetry Great point! This PR still uses old schema where we save the cmd as is, but it makes total sense to just enforce the list in the new lock format. I guess we can do it in https://github.com/iterative/dvc/pull/5128 ?

efiop:
> What were the recent motivations for this? slightly_smiling_face Anyway, this is good.

Just addressing your point that we should handle multiline as a list, so we could close that ticket :slightly_smiling_face:

PR #5198, opened 2021-01-04
Implement straight-to-remote functionality

isidentical:
Part of https://github.com/iterative/dvc/issues/4520

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here: iterative/dvc.org/pull/2091

Internal Tracker
- [x] If the provider supports it, show the total size in the progress bar (maybe with another patch, since this PR is too big already)


efiop:
For the record: while we are reviewing the technical part, let's start creating docs for this, to make sure we are on the same page in terms of CLI semantics.

dvc/tree/base.py
```diff
@@ -59,6 +59,7 @@ class BaseTree:
     TRAVERSE_PREFIX_LEN = 3
     TRAVERSE_THRESHOLD_SIZE = 500000
     CAN_TRAVERSE = True
+    CHUNK_SIZE = 64 * 1024 * 1024  # 64 MiB
```
efiop, line 62: Could you elaborate why this particular chunk size?

dvc/repo/imp_url.py
```diff
@@ -27,6 +29,9 @@ def imp_url(
     out = resolve_output(url, out)
     path, wdir, out = resolve_paths(self, out)
 
+    if to_remote and no_exec:
+        raise ValueError("to-remote option can't be combined with no_exec")
```
efiop, line 33: We tend to use `InvalidArgumentError` for this. You can also call the options by their CLI names and then you won't really need same checks in CLI. E.g. see dvc/repo/run.py.

dvc/tree/base.py
isidentical: In general, the job count is 16 for a normal computer (4 cores), so I wanted to fix it to 1 GiB at most (`64 * 16 = 1024`). For very good end computers (like 8 cores or something like that) it will grow up gradually to 2GB (and 4GB, ...). Making it 128 is also an option, but 64 MiB per chunk seems not that bad (256 MiB is I think very huge, at least that is what my previous experience with it showed).

dvc/repo/imp_url.py
isidentical: Ah, I see. Thanks!

dvc/command/add.py
```diff
@@ -74,6 +77,24 @@ def add_parser(subparsers, parent_parser):
         help="Specify name of the DVC-file this command will generate.",
         metavar="<filename>",
     )
+    parser.add_argument(
+        "-o",
+        "--out",
+        help="Destination path to put files to.",
+        metavar="<filename>",
```
jorgeorpinel, line 84: Consistent with `import`:
```suggestion
        help="Destination path to put files to",
        metavar="<path>",
```

dvc/command/add.py
```diff
@@ -74,6 +77,24 @@ def add_parser(subparsers, parent_parser):
         help="Specify name of the DVC-file this command will generate.",
         metavar="<filename>",
     )
+    parser.add_argument(
+        "-o",
+        "--out",
+        help="Destination path to put files to.",
```
jorgeorpinel, line 83: Question: will this work even without `--to-remote` ?

dvc/command/add.py
```diff
@@ -74,6 +77,24 @@ def add_parser(subparsers, parent_parser):
...
+        metavar="<filename>",
+    )
+    parser.add_argument(
+        "--to-remote",
+        action="store_true",
+        default=False,
+        help="Download it directly to the remote",
```
jorgeorpinel, line 90: How about something like
```suggestion
        help="Implies `--external`. Don't put `targets` in the workspace "
              "nor cache them, but push them to remote storage.",
```

jorgeorpinel: But now that I see this I realize it's indeed a little unexpected that `add` includes `push`...

dvc/command/add.py
```diff
@@ -74,6 +77,24 @@ def add_parser(subparsers, parent_parser):
...
+    )
+    parser.add_argument(
+        "-r",
+        "--remote",
+        help="Remote storage to download to",
```
jorgeorpinel, line 95: I think this should definitely be included in `--to-remote` (let it accept a remote name as argument) to simplify the UI. `add --to-remote --remote` looks fort of repetitive anyway.

dvc/repo/add.py
```diff
@@ -38,6 +40,26 @@ def add(
         raise RecursiveAddingWhileUsingFilename()
 
     targets = ensure_list(targets)
+
+    invalid_opt = None
+    if to_remote:
+        message = "{option} can't be used with --to-remote"
+        if len(targets) != 1:
+            invalid_opt = "multiple targets"
```
jorgeorpinel, line 48: Why can't you add multiple targets (or `--recursive`) straight-to-remote? (only the external ones or checking that all are external). If it's for simplicity of implementation for now, then I understand.

dvc/repo/add.py
```diff
@@ -38,6 +40,26 @@ def add(
...
+        elif no_commit:
+            invalid_opt = "--no-commit option"
+        elif recursive:
+            invalid_opt = "--recursive option"
+    else:
+        message = "{option} can't be used without --to-remote"
+        if kwargs.get("out"):
+            invalid_opt = "--out"
```
jorgeorpinel, line 56: Similar Q. Why introduce this limitation? Might as well enable `add --out` for any file along the way? Someone could find it useful perhaps e.g. add the same data twice (with different resulting names) to do different transformations (an alternative to branching) — not sure.

dvc/command/add.py
isidentical: We actually though about it and decided for the sake of consistency with other commands who works on the remote, it would better to just split it out.

shcheklein: also, in most cases it'll be just `--to-remote`, right? 

also, can an option in argparse be made to optionally accept a value - in this case we would have to make `--to-remote [remote-name]` where remote-name is optional (default remote if not specified)

dvc/repo/add.py
shcheklein: > add the same data twice

doesn't work (outputs overlap) - it'll complain

dvc/repo/add.py
isidentical: > Why can't you add multiple targets (or --recursive) straight-to-remote? 

It was just for mirroring the behavior for `import-url` (which this actually initially implemented in), though it should be trivial to support adding multiple targets if we make `-o`/`--out` take multiple arguments and then figure out a way to match the outputs (like what happens if the user supplies less output destinations than the total number of targets, seems a bit ambigous since it is an optional value).

dvc/command/add.py
isidentical: > also, in most cases it'll be just --to-remote, right?

yes

> also, can an option in argparse be made to optionally accept a value - in this case we would have to make --to-remote [remote-name] where remote-name is optional (default remote if not specified)

yes, the implementation is possible.

dvc/repo/add.py
shcheklein: I agree with @isidentical , let's not complicate (especially ambiguity around `-o` here) it for now. Let's wait for someone to ask about this.

dvc/command/add.py
shcheklein: > yes, the implementation is possible.

then, I guess, I don't have a strong preference I think. `--remote` is good for consistency, but I would not say it looks critical to me, `--to-remote` already reminds it a lot.

jorgeorpinel: > also, in most cases it'll be just --to-remote, right?

I'm not sure that most users employ default remotes. Do we know this?

I see the consistency argument but it's not that big a deal, and forcing the exact flag name causes other issues e.g. just `add --remote` is meaningless (and forbidden, I think, or it it silently ignored?) + an extra flag to document and explain 😋

dvc/repo/add.py
jorgeorpinel: OK. I wasn't proposing to enable this, I was asking.

dvc/repo/add.py
jorgeorpinel: Hmm I don't see this. If the output paths are (given explicitly) different, where's the overlap?

Maybe my example isn't great though, because it's equivalent to making copies of the data and `dvc add` them all (which is trivial and we already support).

dvc/command/add.py
shcheklein: > I'm not sure that most users employ default remotes. Do we know this?

that's why we have default remotes at all. if we are not sure about this we should reconsider the whole logic we work with remotes then

> add --remote is meaningless 

agreed, I would not do this. may be keep only `--to-remote <optional-name>`

dvc/repo/add.py
shcheklein: okay, yep, I'm not exactly right here. If `--out` controls path data will be located within the project, then yes it can be done probably without causing an overlap (but it'll duplicate data potentially, or rename it in the source? not sure about the semantics here). I wouldn't complicate it for now anyway, let's again wait for someone asking this. Since otherwise we'll have to think at least about semantics, and it looks a a bit like a different feature.

jorgeorpinel: It's a different feature indeed. Out of scope for this PR

The only thing is that `add --out` by itself seems to have a meaning but will be forbidden (for now). Not a big deal unless someone complains though

BTW @isidentical maybe some of these restrictions should be handled by printing a warning and ignoring the flag, instead of stopping and throwing an error? Idk, up to you.

isidentical: > maybe some of these restrictions should be handled by printing a warning and ignoring the flag, instead of stopping and throwing an error? Idk, up to you.

I'd personally choose explicit errors rather than some warnings, though I am open to that if it is more convenient for users.

jorgeorpinel: OK. Well, if you're not sure maybe check what DVC does in places where similar limitations already exist. See `repro --recursive/all-pipelines/downstream`, `run --always-changed`, and `check-ignore --all` — those are all the options with documented restrictions (but maybe there's more).

shcheklein: +1 for being strict - errors by default :)

dvc/command/add.py
```diff
@@ -74,6 +78,35 @@ def add_parser(subparsers, parent_parser):
         help="Specify name of the DVC-file this command will generate.",
         metavar="<filename>",
     )
+    parser.add_argument(
+        "-o",
+        "--out",
+        help="Destination path to put files to.",
+        metavar="<path>",
```
jorgeorpinel, line 85: Now that I think about it. I don't see why --out relates especially to transferring data straight to remote. Can you remind me why this is needed or desired please @isidentical ?

shcheklein: @jorgeorpinel you need to fill in the `out` path in the .dvc files that is being created. When you do the regular `dvc add something`. `something` becomes that value. In this case since data is remote you need to specify the way it should be "mapped" to the project's tree.

jorgeorpinel: @shcheklein but you still need to specify a target e.g. `add /tmp/fileordir --to-remote`, so it's possible to use the base name (`fileordir` in this case). That's how the default value for `out` is picked for `import-url`, for example. That's why I find it strange that to-remote requires an explicit output path.

I think `add --out` is a completely separate feature. BTW I would agree on introducing it (but not just for to-remote cases — that's the strange part IMO).

shcheklein: `import-url` has `[out]` for the same reason - to specify where do you want to put that data. It's like with `wget`, `curl`, etc - URLs come often in some raw format that you don't want to use in the workspace. For usual `dvc add` you don't need it, data comes from the workspace already.

If we by default use the same logic as in `import-url` to determine names, then this feature makes sense to my mind and more or less cosistent.

jorgeorpinel: > If we by default use the same logic as in import-url to determine names, then this feature makes sense

🤝 

WDYT @isidentical ?

shcheklein: Just to clarify, we still keep `--out`, we make it (keep it) as not required.

isidentical: If I am not misunderstanding, it already works in that way. If you don't pass something with `--out`, it will fall back to the strategy that import-url uses.

jorgeorpinel:
Oops I meant to resolve https://github.com/iterative/dvc/pull/5198#pullrequestreview-569646959 and accidentally I clicked on "Ready for review". Unfortunately you can't go back to draft status but we won't review this deeply until you request so, @isidentical . Thanks

jorgeorpinel:
Oh actually this is the core PR so I will have nothing to do with deeply reviewing this anyway. Sorry for the mistaken click guys!

isidentical:
> Oops I meant to resolve my #5198 (review) above and accidentally I clicked on "Ready for review". Unfortunately you can't go back to draft status... Sorry!

No worries, eventually I was going to also do that to get @efiop 's reviews!

efiop (approved):
Great stuff! :fire:

Work from this PR is needed for other PRs (e.g. overlapping with vdir functionality and even erepo refactor), so merging this as-is for now. Let's address the CLI and other leftovers on top.

dvc/cache/base.py
```diff
@@ -247,6 +248,116 @@ def _save_file(self, path_info, tree, hash_info, save_link=True, **kwargs):
...
+    def _transfer_file_as_whole(self, from_tree, from_info):
+        from dvc.utils import tmp_fname
+
+        # When we can't use the chunked upload, we have to first download
+        # and then calculate the hash as if it were a local file and then
+        # upload it.
+        local_tree = self.repo.cache.local.tree
+        local_info = local_tree.path_info / tmp_fname()
+
+        from_tree.download(from_info, local_info)
```
efiop, line 260: This might not fit on the local filesystem :slightly_frowning_face: We could download a chunk and upload a chunk though and then assemble them. Or could just forbid transfering files as a whole for now.

jorgeorpinel:
Working on iterative/dvc.org/pull/2091, I'm assuming the concerns left above will be addressed in a follow-up PR. Specifically:

- `--to-remote` option desc: https://github.com/iterative/dvc/pull/5198#pullrequestreview-566897428
- Combine `--to-remote` + `--remote`: https://github.com/iterative/dvc/pull/5198#discussion_r556712627
- Make sure the implementation doesn't download files locally: https://github.com/iterative/dvc/pull/5198#pullrequestreview-571128999 just above
- Other Qs: https://github.com/iterative/dvc/pull/5198#pullrequestreview-571128999

jorgeorpinel:
Another thing to consider (BTW I meant considered/discussed, not necessarily addressed):

Similar to combining --to-remote and --remote for simplicity/usability, should --external (in `add`) be implied with this? (applies to --to-cache too) As of now `add --to-remote /ext/path` fails if I don't also add `--external`:


Cc @dberenbaum

jorgeorpinel:
with error
```
Adding...
ERROR: Output(s) outside of DVC project: /home/jop/existing-data. See <https://dvc.org/doc/user-guide/managing-external-data> for more info.
```

jorgeorpinel:
p.s. on that error message (unrelated to this issue, sorry) should it give a HINT to retry with `--external`?

efiop:
@jorgeorpinel Good catch with `--external`! @isidentical Could you please adjust that? :pray:

isidentical:
I believe this already works with straigh-to-cache when the given output is within the repo, `dvc add /some/stuff -o stuff`. Though for the case of `dvc add --to-remote`, wouldn't it make sense to keep it with explicit `--external` since the actions related to that file in the future would affect paths outside of the project?

jorgeorpinel:
I haven't tested `--to-cache` yet so not sure about the first part of your comment for now, @isidentical.

And sorry, I didn't understand the relationship between requiring explicit `--external` with `--to-remote` to the fact that the data is external (affects external paths). Conceptually though, I don't think we consider consider it "outside of the project" once it's added. In fact we call it an "extended workspace" i.e. DVC "owns" that external location (they become links to the cache and no other system should try to control the same location).

Anyway, my main rationale is that saving users from unnecessarily typing a flag that is always required is good for the UX (similar as with merging `--to-remote` and `--remote`).

isidentical:
Would you create a ticket for the things in your mind (--external, --to-remote + --remote etc) @jorgeorpinel, it is hard to debate these stuff under a merged PR.

jorgeorpinel:
Yeah let me do that because actually I think I'm still confused here 😅

PR #5202, opened 2021-01-04
debug functions: have a good default value

skshetry:
I had to keep providing default value before when profiling
simple functions as an API. I added a default now, and
also made `debugtools` customizable so that you could
edit the code to, for example, have `pdb` enabled by default.

### DIFF
--- a/dvc/_debug.py
+++ b/dvc/_debug.py
@@ -3,7 +3,7 @@ from contextlib import ExitStack, contextmanager
 
 
 @contextmanager
-def instrument(html_output):
+def instrument(html_output=False):
     """Run a statistical profiler"""
     try:
         from pyinstrument import Profiler  # pylint: disable=import-error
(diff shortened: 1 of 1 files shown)


dvc/_debug.py
```diff
@@ -56,14 +56,17 @@ def debug():
...
+        if kw.get("pdb"):
             stack.enter_context(debug())
-        if args.cprofile:
-            stack.enter_context(profile(args.cprofile_dump))
-        if args.instrument:
-            stack.enter_context(instrument(args.instrument_open))
+        if kw.get("cprofile") or kw.get("cprofile_dump"):
+            stack.enter_context(profile(kw.get("cprofile_dump")))
+        if kw.get("instrument") or kw.get("instrument_open"):
+            stack.enter_context(instrument(kw.get("instrument_open", False)))
```
skshetry, line 69: Both --cprofile` and `--cprofile-dump` flags are not required from now on. Similar behaviour for `--instrument(-*)`

PR #5211, opened 2021-01-06
config command: lazily import

skshetry:
I don't see any difference that much while doing `time`, as I think the `dvc.config` modules are loaded eventually anyway, but the startup time has improved.

We need to split the schema for the config and the `Config` class so that they are not compiled when importing.

### Before:
Screenshot from 2021-01-06 15-18-43

### After:
Screenshot from 2021-01-06 15-18-50

### DIFF
--- a/dvc/command/config.py
+++ b/dvc/command/config.py
@@ -3,9 +3,6 @@ import logging
 import os
 
 from dvc.command.base import CmdBaseNoRepo, append_doc_link
-from dvc.config import Config, ConfigError
-from dvc.repo import Repo
-from dvc.utils.flatten import flatten
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


dvc/command/config.py
```diff
@@ -77,7 +76,11 @@ def _list(self):
         return 0
 
     def _get(self, remote, section, opt):
-        levels = [self.args.level] if self.args.level else Config.LEVELS[::-1]
+        from dvc.config import ConfigError
+
+        levels = (
+            [self.args.level] if self.args.level else self.config.LEVELS[::-1]
+        )
```
efiop, line 83: hm, accident?
```suggestion
        levels = [self.args.level] if self.args.level else self.config.LEVELS[::-1]
```

efiop (approved):

PR #5214, opened 2021-01-06
Move misc. imports from command, make them lazy

skshetry:

### DIFF
--- a/dvc/command/dag.py
+++ b/dvc/command/dag.py
@@ -2,7 +2,6 @@ import argparse
 import logging
 
 from dvc.command.base import CmdBase, append_doc_link
-from dvc.stage import Stage
 
 logger = logging.getLogger(__name__)
 
@@ -52,6 +51,8 @@ def _collect_targets(repo, target, outs):
 
 def _transform(repo, outs):
     import networkx as nx
+
+    from dvc.stage import Stage
 
     def _relabel(node) -> str:
         return node.addressing if isinstance(node, Stage) else str(node)
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -13,8 +13,6 @@ from dvc.command.metrics import DEFAULT_PRECISION
 from dvc.command.repro import CmdRepro
 from dvc.command.repro import add_arguments as add_repro_arguments
 from dvc.exceptions import DvcException, InvalidArgumentError
-from dvc.repo.experiments import Experiments
-from dvc.scm.git import Git
 from dvc.utils.flatten import flatten
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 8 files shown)


dvc/command/dag.py
```diff
@@ -53,6 +52,8 @@ def _collect_targets(repo, target, outs):
 def _transform(repo, outs):
     import networkx as nx
 
+    from dvc.stage import Stage
```
skshetry, line 55: `dvc.stage` imports a lot of things (eg: all deps and output which then import trees, etc.). So, lazily importing here.

dvc/command/experiments.py
```diff
@@ -601,6 +601,7 @@ def run(self):
 
 def add_parser(subparsers, parent_parser):
     EXPERIMENTS_HELP = "Commands to run and compare experiments."
+    LAST_CHECKPOINT = ":last"
```
skshetry, line 604: @pmrowla, above import `dvc.repo.experiment` was importing git-related stuffs, dvcfiles and stages which are slow. Instead of refactoring there, I choose to keep it simple by duplicating.

dvc/command/repro.py
```diff
@@ -50,7 +49,7 @@ def add_arguments(repro_parser):
     repro_parser.add_argument(
         "targets",
         nargs="*",
-        help=f"Stages to reproduce. '{PIPELINE_FILE}' by default.",
+        help="Stages to reproduce. 'dvc.yaml' by default.",
```
skshetry, line 52: Like I said above, importing `dvcfile` is slow. And, `dvc.yaml` is set in stone, so no need to import.

dvc/dvcfile.py
```diff
@@ -137,16 +136,17 @@ def merge(self, ancestor, other):
 
 class SingleStageFile(FileMixin):
     from dvc.schema import COMPILED_SINGLE_STAGE_SCHEMA as SCHEMA
+    from dvc.stage.loader import SingleStageLoader as LOADER
```
skshetry, line 139: Same here `stage` import is slow, making it a bit lazy.

dvc/exceptions.py
```diff
@@ -3,8 +3,6 @@
 
 from funcy import first
 
-from dvc.utils import error_link, format_link, relpath
```
skshetry, line 6: `dvc.utils` imports a lot of things. Making it lazy here as `dvc.exceptions` is imported in almost all of the commands that will affect everything (even `dvc -v`)

dvc/dvcfile.py
efiop: Seems a bit weird, maybe let's just import in place?

skshetry: We do it in the `Repo`. 🙂

efiop: Touche :slightly_smiling_face:

efiop (approved):

PR #5236, opened 2021-01-07
cli: remove: fix error message for the case when file is not under control (#4497)

### LINKED ISSUE #4497: remove one of many stage outputs: misleading error message
Users cannot remove a data file if it is only one output of many for a stage. It is a valid behavior that needs a better error message:

```
$ dvc remove model.h5
ERROR: failed to remove 'model.h5' - "Stage 'model.h5' not found inside 'dvc.yaml' file"
```

The pipeline:
```
$ cat dvc.yaml
stages:
  train:
    cmd: python train.py
    deps:
    - users.csv
    - train.py
    - mymodel.py
    params:
    - epochs
    - opt
    - log_file
    - dropout
    outs:
    - logs
    - model.h5       # <-- trying to remove this one
    metrics:
    - summary.json:
        cache: false
    plots:
    - logs.csv:
        cache: false
```

DiPaolo:
Partially fix #4497 

### DIFF
--- a/dvc/command/remove.py
+++ b/dvc/command/remove.py
@@ -10,19 +10,11 @@ logger = logging.getLogger(__name__)
 
 class CmdRemove(CmdBase):
     def run(self):
-        from dvc.stage.exceptions import StageNotFound
-
         for target in self.args.targets:
             try:
                 self.repo.remove(target, outs=self.args.outs)
-            except StageNotFound:
-                logger.error(
-                    f"failed to remove '{target}'. Make sure the target"
-                    f" is either a .dvc file or a name of the stage"
-                )
-                return 1
-            except DvcException:
-                logger.exception(f"failed to remove '{target}'")
+            except DvcException as exc:
+                logger.error(exc)
                 return 1
         return 0
 
--- a/dvc/command/remove.py
+++ b/dvc/command/remove.py
@@ -4,6 +4,7 @@ import logging
 from dvc.command import completion
 from dvc.command.base import CmdBase, append_doc_link
 from dvc.exceptions import DvcException
+from dvc.stage.exceptions import StageNotFound
 
 logger = logging.getLogger(__name__)
 
@@ -13,6 +14,9 @@ class CmdRemove(CmdBase):
         for target in self.args.targets:
             try:
                 self.repo.remove(target, outs=self.args.outs)
+            except StageNotFound:
+                logger.error(f"{target} is not under DVC control")
+                return 1
             except DvcException:
                 logger.exception(f"failed to remove '{target}'")
                 return 1
--- a/dvc/command/remove.py
+++ b/dvc/command/remove.py
@@ -4,18 +4,22 @@ import logging
 from dvc.command import completion
 from dvc.command.base import CmdBase, append_doc_link
 from dvc.exceptions import DvcException
-from dvc.stage.exceptions import StageNotFound
 
 logger = logging.getLogger(__name__)
 
 
 class CmdRemove(CmdBase):
     def run(self):
+        from dvc.stage.exceptions import StageNotFound
+
         for target in self.args.targets:
             try:
                 self.repo.remove(target, outs=self.args.outs)
             except StageNotFound:
-                logger.error(f"{target} is not under DVC control")
+                logger.error(
+                    f"failed to remove '{target}'. Make sure the target"
+                    f" is either a .dvc file or a name of the stage"
+                )
                 return 1
             except DvcException:
                 logger.exception(f"failed to remove '{target}'")
--- a/dvc/stage/exceptions.py
+++ b/dvc/stage/exceptions.py
@@ -80,7 +80,9 @@ class MissingDataSource(DvcException):
         super().__init__(msg)
 
 
-class StageNotFound(DvcException):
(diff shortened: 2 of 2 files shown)


dvc/command/remove.py
```diff
@@ -13,6 +14,9 @@ def run(self):
         for target in self.args.targets:
             try:
                 self.repo.remove(target, outs=self.args.outs)
+            except StageNotFound:
+                logger.error(f"{target} is not under DVC control")
```
skshetry, line 18: We don't support removing an output, only the `.dvc` file or the stage completely. If we want to support it, we need to make the `repo.remove` API support it.

skshetry: Let me comment on the issue itself.

dvc/command/remove.py
```diff
@@ -13,6 +14,9 @@ def run(self):
...
+                return 1
             except DvcException:
                 logger.exception(f"failed to remove '{target}'")
```
skshetry, line 21: For now, we can improve the UI by just removing this log.

dvc/command/remove.py
```diff
@@ -4,6 +4,7 @@
 from dvc.command import completion
 from dvc.command.base import CmdBase, append_doc_link
 from dvc.exceptions import DvcException
+from dvc.stage.exceptions import StageNotFound
```
skshetry, line 7: Note that importing internal modules in `dvc/command` costs us the startup time. If this is needed, it could be imported in `Cmd::run()` below. But, probably it's not required.

skshetry:
ping @DiPaolo, is there anything that we could do to move this PR forward?

DiPaolo:
@skshetry I made fixes according to your notes. I've improved error message for the user in that specific case, because IMHO it makes sense to provide the user more information. The text has been taken from your reply in the issue itself. Now it looks like:
```ERROR: failed to remove 'data/description.json'. Make sure the target is either a .dvc file or a name of the stage```

File: dvc/command/remove.py (hunk 3 above)
Comment (DiPaolo) [reply dvc/command/remove.py]: [offtopic]
I've fixed that. What is the process of reviewing such changes? Should I press "Resolve Conversion"? Will the changes automatically be displayed here? Are there any other actions required by my side?
[/offtopic]

**Comment by skshetry** [comment]
@DiPaolo, we need to support file as a target in the `remove`, so I am not sure if this new error message makes sense. 

We do need to clean up the message, especially the "failed to remove" one. So, either we need to add support for it, or just do the cleanup in this PR. 

Regarding the support for files, the change should be pretty trivial, we just need `collect_granular` here, similar to what's in the `repo.commit`:
https://github.com/iterative/dvc/blob/228b3b16e50ba9074d7471810129079af641ad11/dvc/repo/commit.py#L44

**Comment by DiPaolo** [comment]
@skshetry let me give it another try... I prefer the first option you've suggested - just to clean-up the message. Adding support of files as a target is out of the scope of this issue. So, to track it correctly, I will create a separate issue once the PS is accepted, if you don't mind.

File: dvc/stage/exceptions.py [hunk 4]
```diff
@@ -80,7 +80,7 @@ def __init__(self, missing_files):
         super().__init__(msg)
 
 
-class StageNotFound(KeyError, DvcException):
+class StageNotFound(DvcException):
```
skshetry, line 83: Why do we need to remove `KeyError`?

DiPaolo: ```KeyError``` was the reason why we have extra quotes in the initial messages. To fix that, I've removed it. But before I double-checked the usage of ```StageNotFound```, and didn't find a reason for inheritance from ```KeyError```. So it looked safe for me to remove that inheritance.

skshetry: Internally, StageLoader still works like a dictionary. So, I'd still prefer that we keep the semantics of `StageNotFound` being a KeyError as well.

Here's one example:
https://github.com/iterative/dvc/blob/273518afab02a7b229fc0ef0eba43a556fb80a44/dvc/repo/stage.py#L184

We can get around the quoting behaviour by using a custom `__str__` implementation.
```suggestion
    __str__ = DvcException.__str__
```

DiPaolo: Done

skshetry (approved):
Looks good, thanks @DiPaolo. 🙂

dvc/command/remove.py
```diff
@@ -13,8 +13,8 @@ def run(self):
         for target in self.args.targets:
             try:
                 self.repo.remove(target, outs=self.args.outs)
-            except DvcException:
-                logger.exception(f"failed to remove '{target}'")
+            except DvcException as exc:
+                logger.error(exc)
```
skshetry, line 17: Forgot that this hides the `__cause__`. Will update.

skshetry: Done in #5287

DiPaolo: @skshetry my bad. Thank you

DiPaolo:
> @skshetry let me give it another try... I prefer the second option you've suggested - just to clean-up the message. Adding support of files as a target is out of the scope of this issue. So, to track it correctly, I will create a separate issue once the PR is accepted, if you don't mind.

Here it is #5288

PR #5237, opened 2021-01-08
fix mypy errors; add mypy precommit hooks

### LINKED ISSUE #4697: Add type-hints

skshetry:
Closes #4697

### DIFF
--- a/dvc/repo/add.py
+++ b/dvc/repo/add.py
@@ -1,5 +1,6 @@
 import logging
 import os
+from typing import TYPE_CHECKING
 
 import colorama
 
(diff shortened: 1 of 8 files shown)


dvc/repo/add.py
```diff
@@ -14,20 +15,29 @@
...
+
 logger = logging.getLogger(__name__)
 
 
 @locked
 @scm_context
 def add(
-    repo, targets, recursive=False, no_commit=False, fname=None, **kwargs,
+    repo,
+    targets: "TargetType",
```
skshetry, line 29: I'd have used `TargetType` and `ensure_list` in more places, but most of the places need to leave `None` as-is to be used in `collect` or `used_cache`, so we need something like `ensure_list_or_none()`. Let me know if that makes sense, I'll add them in the next PR.

```python
def ensure_list_or_none(item: Union[List[str], str, None]) -> Optional[List[str]]:
    if item is None:
        return item
    if isinstance(item, str):
       return [item]
    return list(item)
```

pared (approved):

dvc/repo/add.py
skshetry: I am talking about the following stuff that are repeated on most of the API.
https://github.com/iterative/dvc/blob/ccc17917cfcb1b9ab8bc40d5ffdc8bcd02a1040a/dvc/repo/push.py#L22-L23

efiop: A bit ugly, but makes sense.

efiop (approved):
Thank you @skshetry ! :pray:

PR #5239, opened 2021-01-08
S3Tree._list_paths now recognizes the max_items limit

isidentical:

The `MaxKeys` represents the number of items to fetch in one request, if `limit` is not supplied this takes forever and potentially creates tons of requests (I found out about this because my code literally stuck on `from_tree.isdir()` for a very long time on a massive dataset). This is what `limit` is for. We could have also potentially fixed this in the client code by changing `list()` into a `next(<test>, False)` call but fixing this makes it redundant and just a fix of a code smell so not going to bother with it.

```
from dvc.repo import Repo
from dvc.tree import get_cloud_tree

URL = "remote://private/dataset-registry-private"
repo = Repo()

from_tree = get_cloud_tree(repo, url=URL)
from_info = from_tree.path_info
import time
start = time.perf_counter()
print(from_tree.isdir(from_info))
print("Took: ", time.perf_counter() - start)
```
This now results within a second

https://github.com/boto/boto3/issues/631

### DIFF
--- a/dvc/tree/s3.py
+++ b/dvc/tree/s3.py
@@ -202,6 +202,7 @@ class S3Tree(BaseTree):
         kwargs = {"Prefix": path_info.path}
         if max_items is not None:
             kwargs["MaxKeys"] = max_items
+            kwargs["limit"] = max_items
 
         with self._get_bucket(path_info.bucket) as bucket:
             for obj_summary in bucket.objects.filter(**kwargs):
--- a/dvc/tree/s3.py
+++ b/dvc/tree/s3.py
@@ -199,13 +199,13 @@ class S3Tree(BaseTree):
         return path_info.path in self._list_paths(path_info)
 
     def _list_paths(self, path_info, max_items=None):
-        kwargs = {"Prefix": path_info.path}
-        if max_items is not None:
-            kwargs["MaxKeys"] = max_items
-            kwargs["limit"] = max_items
-
         with self._get_bucket(path_info.bucket) as bucket:
-            for obj_summary in bucket.objects.filter(**kwargs):
+            obj_summaries = bucket.objects.filter(Prefix=path_info.path)
+            if max_items is not None:
+                obj_summaries = obj_summaries.page_size(max_items).limit(
+                    max_items
+                )
+            for obj_summary in obj_summaries:
                 yield obj_summary.key
 
     def walk_files(self, path_info, **kwargs):
--- a/tests/func/test_s3.py
+++ b/tests/func/test_s3.py
@@ -116,6 +116,6 @@ def test_copy_multipart_preserve_etag():
 def test_s3_isdir(tmp_dir, dvc, s3):
     s3.gen({"data": {"foo": "foo"}})
     tree = S3Tree(dvc, s3.config)
-    
+
     assert not tree.isdir(s3 / "data" / "foo")
     assert tree.isdir(s3 / "data")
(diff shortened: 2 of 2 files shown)


dvc/tree/s3.py
```diff
@@ -202,6 +202,7 @@ def _list_paths(self, path_info, max_items=None):
         kwargs = {"Prefix": path_info.path}
         if max_items is not None:
             kwargs["MaxKeys"] = max_items
+            kwargs["limit"] = max_items
```
efiop, line 205: Thinking about it again, your alternative suggestion with `limit()` makes more sense. Even navigating the docs, this keyword is not really listed there hinting that this is not an official API, but `limit()` helper is. Let's use limit intead then.

dvc/tree/s3.py
```diff
@@ -199,12 +199,13 @@ def isfile(self, path_info):
...
     def _list_paths(self, path_info, max_items=None):
-        kwargs = {"Prefix": path_info.path}
-        if max_items is not None:
-            kwargs["MaxKeys"] = max_items
-
         with self._get_bucket(path_info.bucket) as bucket:
-            for obj_summary in bucket.objects.filter(**kwargs):
+            obj_summaries = bucket.objects.filter(Prefix=path_info.path)
+            if max_items is not None:
+                obj_summaries = obj_summaries.page_size(max_items).limit(
```
efiop, line 205: Do we even need to alter page size anymore?

isidentical: If we don't add it, it will try to fetch a list 1000 summaries and return the first one AFAIK.

isidentical: You can also try it with the code that I gave in the description that is also much faster with `page_size()`. Runs with `page_size()`;
```
 $ python t.py
Took:  2.2795308610002394
(.venv38) (Python 3.8.5+) [ 11:53ÖÖ ]  [ isidentical@desktop:~/temp_dvc(master✗) ]
 $ python t.py
Took:  2.2522677979995933
(.venv38) (Python 3.8.5+) [ 11:53ÖÖ ]  [ isidentical@desktop:~/temp_dvc(master✗) ]
 $ python t.py
Took:  2.1077582990001247
```
Runs without `page_size()`;
```
 $ python t.py
Took:  3.382246265999129
(.venv38) (Python 3.8.5+) [ 11:53ÖÖ ]  [ isidentical@desktop:~/temp_dvc(master✗) ]
 $ python t.py
Took:  3.211584769998808
(.venv38) (Python 3.8.5+) [ 11:53ÖÖ ]  [ isidentical@desktop:~/temp_dvc(master✗) ]
 $ python t.py
Took:  5.401903420999588
```

tests/func/test_s3.py
```diff
@@ -111,3 +111,17 @@ def test_copy_multipart_preserve_etag():
...
+    path_info = S3Tree.PATH_CLS("s3://bucket/data/foo")
+    tree = S3Tree(dvc, {"url": str(path_info.parent)})
+
+    s3 = tree.s3.meta.client
+    s3.create_bucket(Bucket=path_info.bucket)
+    s3.put_object(Bucket=path_info.bucket, Key=path_info.path, Body="data")
+
+    tree.makedirs(path_info.parent)
+    assert not tree.isdir(path_info)
+    assert tree.isdir(path_info.parent)
```
efiop, line 127: The tests above are a bit obsolete, we have a more handy way to do this now:
```suggestion
def test_s3_isdir(tmp_dir, dvc, s3):
    s3.gen({"data": {"foo": "foo"}})
    tree = S3Tree(dvc, s3.config)
    
    assert not tree.isdir(s3 / "data" / "foo")
    assert tree.isdir(s3 / "data")
```
also please avoid using `scm` fixture unless you need git, as it slows down the test.

dvc/tree/s3.py
efiop: Thanks for clarifying! :pray:

tests/func/test_s3.py
isidentical: Thanks for the tips!

efiop (approved):
Great job! Thank you! :pray:

PR #5246, opened 2021-01-11
Implement optional dos2unix feature

MetalBlueberry:

Hi again =)

Related to #4658, I've implemented the [option to disable](https://github.com/iterative/dvc/issues/4658#issuecomment-703790216) dos2unix fron the dvc configuration file.

We are starting to use dvc internally, but we are facing issues with pdf or image/tiff formats because they are accidentally identified as text. This means that every /r/n is replaced to /n and the md5 hash calculation no longer matches the file content.

We rolled out our own version of dvc internally with this patch, but it is hard to explain to people that they have to install it from another source. For that reason, I've invested a little bit more time to properly add the parameter and submit this PR. Hopefully, someone else will be able to enjoy this feature.

By default, the parameter dos2unix under **core** configuration will be True. So dvc will behave as usual for everyone. In case someone notices errors due to the dos2unix feature, they can simply set dos2unix to False. 

I know that changing this value in an already existing repo may have weird issues, like dvc detecting changes on unchanged files. I think this won't be a usual case because either you have problems and disable it, or you never notice that this option exists. Usually, this value will be set at project creation and never modified again.

I will write the documentation page, but first, I would like to have your feedback.

Thanks for your support!

### DIFF
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -117,6 +117,7 @@ SCHEMA = {
         Optional("autostage", default=False): Bool,
         Optional("experiments"): Bool,  # obsoleted
         Optional("check_update", default=True): Bool,
+        Optional("dos2unix", default=True): Bool,
     },
     "cache": {
         "local": str,
--- a/dvc/tree/local.py
+++ b/dvc/tree/local.py
@@ -303,7 +303,8 @@ class LocalTree(BaseTree):
         return stat.S_IMODE(mode) == self.CACHE_MODE
 
     def get_file_hash(self, path_info):
-        hash_info = HashInfo(self.PARAM_CHECKSUM, file_md5(path_info)[0],)
+        to_unix=self.repo.config["core"].get("dos2unix")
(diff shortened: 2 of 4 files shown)


efiop (commented):
Looks pretty good, thank you!

Though this makes me think if we should indeed do the next step and just switch to no-dos2unix by default in 2.0, to not waste this chance. @pmrowla has been working on the new way to store data more efficiently, and we though that maybe it is better to wait for that to do this switch, but that big feature will likely become default only in 3.0 :(

If we want to do the switch in 2.0, we'll likely need to put a cache version/schema file in cache/remote, use it to verify that cache/remote is of expected version and provide a trivial migration script.

efiop:
Just an update: @pmrowla is working on dropping CRLF for dvc 2.0. Closing this PR for now.

MetalBlueberry:
@efiop, Can you leave a reference to the @pmrowla  work? My team will need to understand the new storage mechanism as we are maintaining our own server implementation to parse dvc files.
Thank you!

efiop:
@MetalBlueberry Sure, we use https://github.com/iterative/dvc/issues/4658 to track progress on that.

PR #5250, opened 2021-01-12
dvc: bump to 2.0.0a0

skshetry:
Looking at the log is confusing, as the version is stuck in `1.11.0` whereas the recent one is `1.11.10`.

As the updater is also dependent on that version, it prints an ugly update message while running too.
https://github.com/iterative/dvc/issues/5023#issuecomment-758450519

Even though we are not releasing right now, it'd be better to bump the version as it can be no longer considered as 1.X.

### DIFF
--- a/dvc/version.py
+++ b/dvc/version.py
@@ -6,7 +6,7 @@
 import os
 import subprocess
 
-_BASE_VERSION = "1.11.0"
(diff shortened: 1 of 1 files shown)


efiop:
Thank you!

PR #5257, opened 2021-01-13
checkpoints/live: move execution env to stage

pared:

This PR aims to make `Output` source of the execution environment for `Stage.run`. It seems that outputs are the cause of `env.`
Related: https://github.com/iterative/dvc/pull/4902#discussion_r542551178

### DIFF
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -273,16 +273,21 @@ class Stage(params.StageParams):
                     int(config.get(BaseOutput.PARAM_LIVE_REPORT, True))
                 )
         elif out.checkpoint and checkpoint_func:
-            from dvc.env import DVC_CHECKPOINT, DVC_ROOT
+            from dvc.env import DVC_CHECKPOINT
 
-            env.update({DVC_CHECKPOINT: "1", DVC_ROOT: self.repo.root_dir})
+            env.update({DVC_CHECKPOINT: "1"})
         return env
 
     def env(self, checkpoint_func=None) -> Env:
+        from dvc.env import DVC_ROOT
+
         env: Env = {}
+        if self.repo:
+            env.update({DVC_ROOT: self.repo.root_dir})
+
         for out in self.outs:
             current = self._read_env(out, checkpoint_func=checkpoint_func)
-            if any(current.keys() and env.keys()):
+            if set(env.keys()).intersection(set(current.keys())):
                 raise DvcException("Duplicated env variable")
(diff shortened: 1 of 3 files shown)


pmrowla:
This looks ok to me with the way checkpoints and dvclive are currently defined, but there has been some discussion on  whether or not checkpoint should just be a stage level flag in the future, rather than an output level flag. Not sure if the same consideration would apply to dvclive?

pared:
@pmrowla 
Well, as of now, there can be only one dvclive defined. And `env` is resource used mainly by stage, so it does make sense to me to define it on `stage` level. I can move `env` out of `out` into `stage`. How about that?

pmrowla:
> @pmrowla
> Well, as of now, there can be only one dvclive defined. And `env` is resource used mainly by stage, so it does make sense to me to define it on `stage` level. I can move `env` out of `out` into `stage`. How about that?

I think having it in `stage` makes more sense 👍

pared:
@pmrowla moved `env` to `stage`

efiop:
@pared Looks like PR and commit titles need an update too?

dvc/stage/__init__.py
```diff
@@ -252,13 +252,36 @@ def is_checkpoint(self):
...
+                env[DVCLIVE_SUMMARY] = str(
+                    int(config.get(BaseOutput.PARAM_LIVE_SUMMARY, True))
+                )
+                env[DVCLIVE_REPORT] = str(
+                    int(config.get(BaseOutput.PARAM_LIVE_REPORT, True))
+                )
+        elif out.checkpoint and checkpoint_func:
+            from dvc.env import DVC_CHECKPOINT, DVC_ROOT
+
+            env.update({DVC_CHECKPOINT: "1", DVC_ROOT: self.repo.root_dir})
```
pmrowla, line 275: This isn't because of your changes, but looking at this now, I think we should be setting `DVC_ROOT` all the time in the calling `env()` function. (`DVC_CHECKPOINT` should still only be set inside this `if` case)

For experiments that are run in temp-dirs (and not only checkpoint experiments), it will always be useful for user's code to be able to see where it is being run, since they may need to work with relpaths from that directory (instead of relpaths inside their git/dvc repo workspace). And outside of the experiment context, it won't hurt us to set `DVC_ROOT` for regular `repro` runs as well.

pared: Good point!

pmrowla (approved):

PR #5260, opened 2021-01-13
system: prevent mypy from checking windows-related method

pared:

mypy fails on non-windows machines because the `sys` module does not have `getwindowsversion`.

### DIFF
--- a/dvc/system.py
+++ b/dvc/system.py
@@ -10,9 +10,9 @@ from dvc.exceptions import DvcException
 logger = logging.getLogger(__name__)
 
 if (
-    platform.system() == "Windows"
+    sys.platform == "win32"
     and sys.version_info < (3, 8)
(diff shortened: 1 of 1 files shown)


efiop:
@karajan1001 has also ran into this https://discord.com/channels/485586884165107732/565699007037571084/798846655935152158

dvc/system.py
```diff
@@ -12,7 +12,7 @@
 if (
     platform.system() == "Windows"
```
skshetry, line 13: @pared, what happens if we change this to following?
```suggestion
    sys.platform == "win32"
```

Does that work?

skshetry: But I am fine with this change as well. 🙂

pared: it does, will change the PR

skshetry (approved):

PR #5261, opened 2021-01-13
tree: implement getsize()

isidentical:

### DIFF
--- a/dvc/tree/base.py
+++ b/dvc/tree/base.py
@@ -226,20 +226,7 @@ class BaseTree:
     def getsize(self, path_info):
         if self.isfile(path_info):
             return self._getsize(path_info)
-
-        try:
-            file_iter = self.walk_files(path_info)
-        except NotImplementedError:
-            return None
-
-        total = 0
-        for file in file_iter:
-            size = self.getsize(file)
-            if size is None:
-                return None
-            total += size
-
-        return total
+        raise NotImplementedError
 
     def remove(self, path_info):
         raise RemoteActionNotImplemented("remove", self.scheme)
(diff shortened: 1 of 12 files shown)


dvc/tree/base.py
```diff
@@ -220,6 +220,27 @@ def walk_files(self, path_info, **kwargs):
...
 
+    def _getsize(self, path_info):
+        return None
+
+    def getsize(self, path_info):
+        if self.isfile(path_info):
+            return self._getsize(path_info)
+
+        try:
+            file_iter = self.walk_files(path_info)
```
efiop, line 231: I guess this becomes more of `du` than `getsize` at this point. In transfer you'll likely need to walk and transfer at the same time, and walking files here once again will be wasteful in terms of API calls.

efiop: If you only really need getsize for files right now, I would just stick to that. `du` will be useful later, but there seems to be no need to implement it right now.

isidentical: Sure! Since I will always pass a file in the end, it wouldn't matter much for me. Will only make this available for files

dvc/tree/base.py
```diff
@@ -226,20 +226,7 @@ def _getsize(self, path_info):
     def getsize(self, path_info):
         if self.isfile(path_info):
```
efiop, line 227: This will waste an API call, no need for it.

isidentical: What about calling it `get_file_size`?

efiop: @isidentical No need, `getsize` is always called for a particular file. Like os.getsize and like https://github.com/intake/filesystem_spec/blob/master/fsspec/spec.py states (our long-term goal).

efiop (approved):

PR #5265, opened 2021-01-13
stage collection: do not collect .gitignored dvcfiles

### LINKED ISSUE #5244: stage collection: ignore git-ignored directories?
`dvc status` in our repo is terribly slow (~6s), and I think a project of a reasonable size of our users are slow as well, as we might be traversing through very dense directories (eg: `node_modules`/`venv`, and user's dvc-tracked data).

We have discussed before to get rid of `.dvcignore`, but I see it to be a different thing, useful in, for example, ignoring `.DS_Store`/`logs`, etc in dvc-tracked artefacts.

We could opt for a simple solution and just ignore the git-ignored dvcfiles and directories and never traverse them. In addition to this, as we also add user's data in the `gitignore`, we can also get significant speedups just by doing this.

Also, design-wise, we won't need to mix `gitignore`s and `dvcignore`s in the `tree`s.

##### Before
```console
  Time (mean ± σ):      4.240 s ±  0.192 s    [User: 3.535 s, System: 0.474 s]
  Range (min … max):    4.141 s …  4.781 s    10 runs
```

#### After
```console
Benchmark #1: dvc status with dulwich cached
  Time (mean ± σ):     512.1 ms ±  73.2 ms    [User: 387.6 ms, System: 43.3 ms]
  Range (min … max):   436.2 ms … 704.0 ms    10 runs
```

Also, note that we can change this behaviour in 2.0 release, as this could be considered a breaking change.

skshetry:
Closes #5244 

A few fixes has been made along the way, including:
* better error message if the dvc.yaml and .dvc files are dvc-ignored.
  ```console
  $ echo "dvc.yaml" >> .dvcignore
  $ dvc repro
  ERROR: 'dvc.yaml' is dvc-ignored. 
  ```
* better error message if the dvc.yaml and .dvc files are git-ignored.
  ```console
  $ echo "dvc.yaml" >> .gitignore
  $ dvc repro
  ERROR: 'dvc.yaml' is git-ignored.
  ```

An open question is, what should we do about lockfiles if they are git-ignored or dvc-ignored?
Right now, we assume the lockfile as empty if they are dvc-ignored and we don't care about git-ignores
for the lockfiles. I feel that this behaviour is good, but if we want them to be consistent, please do tell me.

Another thing to keep in mind is, since the implementation depends on `isinstance(repo.tree, LocalTree)`, if we start using LocalTree for `get/list/import`, it'll break, though I don't think it should as we better apply it to only the user's workspace (IIRC `dvcx` uses this mode).
Though, this might be handled by a state (eg: `repo.is_external_repo` or something like that).

### DIFF
--- a/dvc/repo/stage.py
+++ b/dvc/repo/stage.py
@@ -13,6 +13,7 @@ from typing import (
     Tuple,
     Union,
 )
+from functools import partial
 
 from dvc.exceptions import (
     DvcException,
@@ -371,13 +372,33 @@ class StageLoad:
                 the collection.
         """
         from dvc.dvcfile import is_valid_filename
+        from dvc.tree.local import LocalTree
+
+        scm = self.repo.scm
+        sep = os.sep
+        outs: Set[str] = set()
+
+        is_local_tree = isinstance(self.tree, LocalTree)
+
+        def is_ignored(path):
+            # apply only for the local tree
+            return is_local_tree and scm.is_ignored(path)
+
+        def is_dvcfile_and_not_ignored(root, file):
+            return is_valid_filename(file) and not is_ignored(
+                f"{root}{sep}{file}"
+            )
+
+        def is_out_or_ignored(root, directory):
+            dir_path = f"{root}{sep}{directory}"
+            return dir_path in outs or is_ignored(dir_path)
 
         stages = []
-        outs: Set[str] = set()
         for root, dirs, files in self.tree.walk(self.repo.root_dir):
-            for file_name in filter(is_valid_filename, files):
-                file_path = os.path.join(root, file_name)
+            dvcfile_filter = partial(is_dvcfile_and_not_ignored, root)
 
+            for file in filter(dvcfile_filter, files):
+                file_path = os.path.join(root, file)
                 try:
                     new_stages = self.load_file(file_path)
                 except DvcException as exc:
@@ -393,5 +414,5 @@ class StageLoad:
                     for out in stage.outs
                     if out.scheme == "local"
                 )
-            dirs[:] = [d for d in dirs if os.path.join(root, d) not in outs]
+            dirs[:] = [d for d in dirs if not is_out_or_ignored(root, d)]
         return stages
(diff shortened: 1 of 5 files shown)


dvc/repo/stage.py
```diff
@@ -371,13 +372,33 @@ def collect_repo(self, onerror: Callable[[str, Exception], None] = None):
...
-        stages = []
+        scm = self.repo.scm
+        sep = os.sep
         outs: Set[str] = set()
+
+        is_local_tree = isinstance(self.tree, LocalTree)
+
+        def is_ignored(path):
+            # apply only for the local tree
+            return is_local_tree and scm.is_ignored(path)
```
skshetry, line 385: This check might not work everywhere, especially in `for_write=True` scenarios in `dvcx`. But as we are adjusting our APIs in `ExternalRepo`, `Repo.clone()` should set some kind of state to workaround that.

efiop (approved):

pared:
> An open question is, what should we do about lockfiles if they are git-ignored or dvc-ignored?

I think users should be able to ignore lockfile. Dvc is able (to some extent) to work without it, so removing this possibility would limit `dvc`'s functionality. And I guess one can come with some use cases where this is actually a feature. For example any educational material that's purpose is to show how running the pipeline looks like.

pmrowla (approved):

pmrowla:
> > An open question is, what should we do about lockfiles if they are git-ignored or dvc-ignored?
> 
> I think users should be able to ignore lockfile. Dvc is able (to some extent) to work without it, so removing this possibility would limit `dvc`'s functionality. And I guess one can come with some use cases where this is actually a feature. For example any educational material that's purpose is to show how running the pipeline looks like.

One thing to keep in mind is that experiments runs in temp directories (and on remote machines in the future) have to git commit `dvc.lock` in order for us to retrieve the experiment result. We can (and already do) force-add gitignored lock files internally for these types of runs, and on the user's local `exp apply` the dvc.lock file would no longer be staged/committed like normal. But this does seem like the kind of thing that may lead to unexpected behavior in the future and it may be safer for us to forbid it?

pared:
@pmrowla Fair point, I didn't consider experiments in my thought process. In that case it might be better to just require it.

PR #5268, opened 2021-01-14
remove .gitignore entries after output changes

### LINKED ISSUE #5156: repro/run: remove .gitignore entries when no longer needed
> Similar to #4962

If I change the name of an output in my code and dvc.yaml and re-run repro, the new name gets added to .gitignore, but the old one is not removed.

One reason I may be changing the name is to liberate the previous one and perhaps add it to Git. So I think this would be an intuitive/expected behavior.

karajan1001:
fix #5156

1. Add a failed test
2. solve it.

### DIFF
--- a/dvc/dvcfile.py
+++ b/dvc/dvcfile.py
@@ -361,6 +361,24 @@ class Lockfile(FileMixin):
         version = LOCKFILE_VERSION.V2.value  # pylint:disable=no-member
         return {SCHEMA_KWD: version}
 
+    def _remove_unused_outs_in_gitignore(
+        self, orig_stage: dict, new_stage: dict
+    ):
+        def get_out_path_set(stage_data: dict) -> set:
+            stage_outs = stage_data.get("outs", [])
+            path_list = []
+            for out in stage_outs:
+                if isinstance(out, dict) and "path" in out:
+                    path_list.append(out["path"])
+            return set(path_list)
+
+        orig_set = get_out_path_set(orig_stage)
+        new_set = get_out_path_set(new_stage)
+        remove_set = orig_set - new_set
+        for path in remove_set:
+            path = os.path.abspath(path)
+            self.repo.scm.ignore_remove(path)
+
     def dump(self, stage, **kwargs):
         stage_data = serialize.to_lockfile(stage)
 
@@ -378,13 +396,16 @@ class Lockfile(FileMixin):
                     logger.info("Generating lock file '%s'", self.relpath)
 
             data["stages"] = data.get("stages", {})
-            modified = data["stages"].get(stage.name, {}) != stage_data.get(
-                stage.name, {}
-            )
+            orig_stage_data = data["stages"].get(stage.name, {})
+            new_stage_data = stage_data.get(stage.name, {})
+            modified = orig_stage_data != new_stage_data
+
             if modified:
                 logger.info("Updating lock file '%s'", self.relpath)
 
             data["stages"].update(stage_data)
+
+        self._remove_unused_outs_in_gitignore(orig_stage_data, new_stage_data)
 
         if modified:
             self.repo.scm.track_file(self.relpath)
--- a/tests/func/test_repro.py
+++ b/tests/func/test_repro.py
@@ -7,7 +7,7 @@ from pathlib import Path
 import pytest
 from mock import patch
 
-from dvc.dvcfile import DVC_FILE, Dvcfile
+from dvc.dvcfile import DVC_FILE, PIPELINE_FILE, Dvcfile
 from dvc.exceptions import (
     CyclicGraphError,
     ReproductionError,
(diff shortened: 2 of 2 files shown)


dvc/dvcfile.py
```diff
@@ -332,6 +332,24 @@ def latest_version_info(self):
...
+            for out in stage_outs:
+                if isinstance(out, dict) and "path" in out:
+                    path_list.append(out["path"])
+            return set(path_list)
+
+        orig_set = get_out_path_set(orig_stage)
+        new_set = get_out_path_set(new_stage)
+        remove_set = orig_set - new_set
+        for path in remove_set:
+            path = os.path.abspath(path)
```
karajan1001, line 350: Any better solutions to get the abs path?

karajan1001:
This is not an atomic operation. Add the new outs to `.gitignore` happens when we save outputs while removing the old outs in `.gitignore` requires information from `dvc.lock`. For convenience, I make removing old outputs at `dvc.lock` dumping .

karajan1001:
@Mergifyio rebase

Can this work?

karajan1001:
@Mergifyio rebase

tests/func/test_repro.py
```diff
@@ -1247,3 +1247,27 @@ def test_repro_when_cmd_changes(tmp_dir, dvc, run_copy, mocker):
...
+        a = f.readlines()
+        assert "/{}\n".format(old) in a and "/{}\n".format(new) not in a
+    stage_yaml = load_yaml(PIPELINE_FILE)
+    stage_yaml["stages"][stage_name]["cmd"] = stage_yaml["stages"][stage_name][
+        "cmd"
+    ].replace(old, new)
+    stage_yaml["stages"][stage_name]["outs"].remove(old)
+    stage_yaml["stages"][stage_name]["outs"].append(new)
+    dump_yaml(PIPELINE_FILE, stage_yaml)
+    main(["repro", "copy-foo-bar"])
```
karajan1001, line 1269:
```
    dvc.reproduce(stage.addressing)
```
are better?

dvc/dvcfile.py
```diff
@@ -332,6 +332,24 @@ def latest_version_info(self):
         version = LOCKFILE_VERSION.V2.value  # pylint:disable=no-member
         return {SCHEMA_KWD: version}
 
+    def _remove_unused_outs_in_gitignore(
+        self, orig_stage: dict, new_stage: dict
+    ):
+        def get_out_path_set(stage_data: dict) -> set:
+            stage_outs = stage_data.get("outs", [])
```
skshetry, line 339: The issue #5156 is pretty broad, though I like your approach of using `dvc.lock` data to figure out what to remove.

Though I see a few problems regarding the implementation:
1. Lockfiles don't have `wdir` (it's read from `dvc.yaml`). So we don't have a way to know where the `.gitignore` file is. You cannot reliably use `wdir` from `dvc.yaml` or `stage` arg passed in `lockfile.dump()` as they might have been updated/changed/removed as well. We might end up deleting the user's `.gitignore` entries this way even if we try.
2. `Lockfile` does not feel like a good place to determine this (see that you have to parse `outs`). We need some abstraction to hide this somewhere. 

 (Maybe we need a synchronized dumping abstraction that takes care of this stuff together. We have been discussing of introducing a `repo.index.update()` abstraction that might take the responsibility, but we are still not sure of it.)

skshetry: Please ignore the 2) point, just trying to provide the context. We can accept the contribution as-is for now, if we are able to solve 1) somehow.

karajan1001: @skshetry 
I agreed, file operations ( `.gitignore`, `dvc.yaml`, `dvc.lock`) are too scattered and hard to maintain. It took me half an hour to found where the outputs are added to `.gitignore`.

karajan1001: The same with #5293. We need to summarize the DVC index files (`dvc.yaml`, `dvc.lock`) information to make the final modification decision.

skshetry: @karajan1001, we don't have enough information anyway. We cannot depend on `dvc.yaml` data for this at all, as this is the "source of truth" that resulted in old .gitignore entries being no longer valid. So we cannot trust it, not even for `wdir` as it could have changed as well.

Feels like keeping 3 kinds of data format (dvc.lock, dvc.yaml and .gitignore) in sync is not a good idea?

karajan1001: @skshetry 
Not only `.gitignore` files but also `dvc.yaml` and `dvc.lock` can we have inside a `wdir`. Data tracked in one `dvc.yaml` might be moved to another one. Gathering all of the information from these dvcfiles and lockfiles and keep them in sync is not an easy work.

karajan1001:
> This is not an atomic operation. Add the new outs to `.gitignore` happens when we save outputs while removing the old outs in `.gitignore` requires information from `dvc.lock`. For convenience, I make removing old outputs at `dvc.lock` dumping .

And there is another problem because my solution only considers one stage file. We only know that in this stage the old file has been removed, no one can guarantee that it is not moved to some other stage.  For example, there were three stages `A`, `B`, `C`, and `C` depends on `A` and `B`. 

A  B
|.   /
|. /
C

originally file `a` belongs to stage `B`,  then we moved file `a` to stage `A` and reproduce the `C`. Then file `a1 would first be added to `.gitignore` by stage `A` then delete by stage `B`.  The correct result would be that `a` still in `.gitignore` due to stage `A`.

efiop:
Discussed with @karajan1001 and @skshetry during planning and decided that we don't have enough info to solve this automatically (e.g. our lock files lack wdir), so closing this PR for now. In the future we'll potentially introduce some kind of gc-like mechanism to cleanup leftovers in the repo.

PR #5270, opened 2021-01-14
scm: implement pygit2 resolve rev and native checkout

pmrowla:

Implements required functionality for workspace `git checkout` and `git checkout -b` (`get_ref/set_ref/iter_refs`, `resolve_rev`, `reset`) in pygit2 backend.
Related to #2215

### DIFF
--- a/dvc/scm/git/backend/gitpython.py
+++ b/dvc/scm/git/backend/gitpython.py
@@ -375,7 +375,10 @@ class GitPythonBackend(BaseGitBackend):  # pylint:disable=abstract-method
 
         if name == "HEAD":
             try:
-                return self.repo.head.commit.hexsha
+                if follow or self.repo.head.is_detached:
+                    return self.repo.head.commit.hexsha
+                else:
+                    return f"refs/heads/{self.repo.active_branch}"
             except (GitCommandError, ValueError):
                 return None
         elif name.startswith("refs/heads/"):
--- a/dvc/scm/git/backend/gitpython.py
+++ b/dvc/scm/git/backend/gitpython.py
@@ -370,9 +370,7 @@ class GitPythonBackend(BaseGitBackend):  # pylint:disable=abstract-method
         except GitCommandError as exc:
             raise SCMError(f"Failed to set ref '{name}'") from exc
 
-    def get_ref(
-        self, name: str, follow: Optional[bool] = True
-    ) -> Optional[str]:
+    def get_ref(self, name: str, follow: bool = True) -> Optional[str]:
         from git.exc import GitCommandError
 
         if name == "HEAD":
--- a/dvc/scm/git/backend/gitpython.py
+++ b/dvc/scm/git/backend/gitpython.py
@@ -496,7 +496,8 @@ class GitPythonBackend(BaseGitBackend):  # pylint:disable=abstract-method
         try:
             self.git.stash("apply", rev)
         except GitCommandError as exc:
-            if "CONFLICT" in str(exc):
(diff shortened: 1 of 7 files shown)


pmrowla:
Investigating why checkpoint resume breaks with the pygit2 checkout, everything else in DVC appears to work properly with it.

dvc/scm/git/backend/pygit2.py
```diff
@@ -223,7 +280,20 @@ def diff(self, rev_a: str, rev_b: str, binary=False) -> str:
         raise NotImplementedError
 
     def reset(self, hard: bool = False, paths: Iterable[str] = None):
-        raise NotImplementedError
+        from pygit2 import GIT_RESET_HARD, GIT_RESET_MIXED, IndexEntry
+
+        self.repo.index.read()
```
pmrowla, line 285: `libgit2/pygit2` does not automatically write Git index file to disk in most cases, it's kept in-memory only until the repo object is closed. Since we are regularly swapping backends between git operations, we have to be careful to always explicitly read/write the `pygit2` to and from disk when doing any Git index related operations.

dvc/scm/git/backend/pygit2.py
```diff
@@ -93,7 +93,25 @@ def commit(self, msg: str, no_verify: bool = False):
...
-        raise NotImplementedError
+        from pygit2 import GitError
+
+        if create_new:
+            commit = self.repo.revparse_single("HEAD")
+            new_branch = self.repo.branches.local.create(branch, commit)
+            self.repo.checkout(new_branch)
+        else:
+            if branch == "-":
+                branch = "@{-1}"
```
pmrowla, line 104: For whatever reason, libgit2's `resolve_refish` doesn't handle `-` shorthand, probably because it's specific to `git checkout/switch` and not a formal/general git ref shorthand symbol

tests/func/test_install.py
```diff
@@ -50,7 +50,7 @@ def test_post_checkout(self, tmp_dir, scm, dvc):
         os.unlink("file")
         scm.install()
 
-        scm.checkout("new_branch", create_new=True)
+        scm.gitpython.git.checkout("-b", "new_branch")
```
pmrowla, line 53: libgit2 does not run git hooks on any command, and internally we don't need to run this hook.

For this test we want to make sure that the installed DVC post-checkout hook is run when a user does command-line `git checkout`, not that it is run on DVC `scm.checkout`.

pmrowla:
`experiments/test_gc` still failing because something in the changes is causing carriage returns to get committed on windows even with `core.autocrlf` is set to true.

dvc/scm/git/backend/gitpython.py
```diff
@@ -370,14 +370,15 @@ def set_ref(
...
+    def get_ref(self, name: str, follow: bool = True) -> Optional[str]:
         from git.exc import GitCommandError
 
         if name == "HEAD":
             try:
-                return self.repo.head.commit.hexsha
+                if follow or self.repo.head.is_detached:
+                    return self.repo.head.commit.hexsha
+                else:
+                    return f"refs/heads/{self.repo.active_branch}"
```
efiop, line 381: A bit offtopic and not important in such a small `else` statement, but what do you guys think about enabling no-else-after-return check in our linters? I seem to stumble upon it regularly.
```suggestion
                return f"refs/heads/{self.repo.active_branch}"
```

efiop (approved):
Great stuff :rocket:

pared (approved):

PR #5273, opened 2021-01-14
repo: Add repro from given out names functionality

### LINKED ISSUE #3875: repro: support running stage via it's output's name
We should probably accept an output as a target.
It can solve some problems with autocomplete and improve experience. Related #3743 (especially last comments).
```sh
$ dvc repro model.pkl
```

*Originally posted by @shcheklein in #3777*

Vonski:
I'm not sure if I should have added special test for this functionality.

Should the `--glob` functionality be preserved? If so, should handling for it be added to `collect_granular(...)`?

Fixes #3875

@pared

### DIFF
--- a/dvc/command/repro.py
+++ b/dvc/command/repro.py
@@ -11,10 +11,6 @@ logger = logging.getLogger(__name__)
 
 class CmdRepro(CmdBase):
     def run(self):
-        if self.args.glob and self.args.glob_stages:
-            raise ValueError(
-                "--glob and --glob-stages options are mutually exclusive."
-            )
         stages = self.repo.reproduce(**self._repro_kwargs)
         if len(stages) == 0:
             logger.info(CmdDataStatus.UP_TO_DATE_MSG)
@@ -151,22 +147,23 @@ def add_arguments(repro_parser):
             "from the run-cache."
         ),
     )
-    repro_parser.add_argument(
+    glob_group = repro_parser.add_mutually_exclusive_group()
+    glob_group.add_argument(
         "--glob",
         action="store_true",
         default=False,
         help=(
             "Allows targets containing shell-style wildcards to match out "
-            "files.",
+            "files."
         ),
     )
-    repro_parser.add_argument(
+    glob_group.add_argument(
         "--glob-stages",
         action="store_true",
         default=False,
         help=(
             "Allows targets containing shell-style wildcards to match stage "
-            "names.",
+            "names."
         ),
     )
 
--- a/dvc/command/repro.py
+++ b/dvc/command/repro.py
@@ -50,7 +50,7 @@ def add_arguments(repro_parser):
     repro_parser.add_argument(
         "targets",
         nargs="*",
-        help="Stages to reproduce. 'dvc.yaml' by default.",
+        help="Stages or output files to reproduce. 'dvc.yaml' by default.",
     ).complete = completion.DVCFILES_AND_STAGE
     repro_parser.add_argument(
         "-f",
--- a/dvc/command/repro.py
+++ b/dvc/command/repro.py
@@ -50,7 +50,7 @@ def add_arguments(repro_parser):
     repro_parser.add_argument(
         "targets",
         nargs="*",
-        help="Stages or output files to reproduce. 'dvc.yaml' by default.",
+        help="Stage names or stage outputs to reproduce. 'dvc.yaml' by default.",
     ).complete = completion.DVCFILES_AND_STAGE
     repro_parser.add_argument(
         "-f",
--- a/dvc/command/repro.py
+++ b/dvc/command/repro.py
@@ -50,7 +50,8 @@ def add_arguments(repro_parser):
     repro_parser.add_argument(
         "targets",
         nargs="*",
-        help="Stage names or stage outputs to reproduce. 'dvc.yaml' by default.",
+        help="Stage names or stage outputs to reproduce. 'dvc.yaml' by "
+        "default.",
     ).complete = completion.DVCFILES_AND_STAGE
     repro_parser.add_argument(
         "-f",
--- a/dvc/command/repro.py
+++ b/dvc/command/repro.py
@@ -11,6 +11,10 @@ logger = logging.getLogger(__name__)
 
 class CmdRepro(CmdBase):
     def run(self):
+        if self.args.glob and self.args.glob_stages:
+            raise ValueError(
+                "--glob and --glob-stages options are mutually exclusive."
+            )
         stages = self.repo.reproduce(**self._repro_kwargs)
         if len(stages) == 0:
             logger.info(CmdDataStatus.UP_TO_DATE_MSG)
@@ -42,6 +46,7 @@ class CmdRepro(CmdBase):
             "force_downstream": self.args.force_downstream,
             "pull": self.args.pull,
             "glob": self.args.glob,
+            "glob_stages": self.args.glob_stages,
         }
 
 
@@ -150,7 +155,19 @@ def add_arguments(repro_parser):
         "--glob",
         action="store_true",
         default=False,
-        help="Allows targets containing shell-style wildcards.",
+        help=(
+            "Allows targets containing shell-style wildcards to match out "
+            "files.",
+        ),
+    )
+    repro_parser.add_argument(
+        "--glob-stages",
+        action="store_true",
+        default=False,
+        help=(
+            "Allows targets containing shell-style wildcards to match stage "
+            "names.",
+        ),
     )
 
 
--- a/dvc/command/repro.py
+++ b/dvc/command/repro.py
@@ -153,8 +153,8 @@ def add_arguments(repro_parser):
         action="store_true",
         default=False,
         help=(
-            "Allows targets containing shell-style wildcards to match out "
-            "files."
+            "Allows targets containing shell-style wildcards to match stage "
+            "outputs."
         ),
     )
     glob_group.add_argument(
--- a/dvc/command/repro.py
+++ b/dvc/command/repro.py
@@ -50,8 +50,8 @@ def add_arguments(repro_parser):
     repro_parser.add_argument(
         "targets",
         nargs="*",
-        help="Stage names or stage outputs to reproduce. 'dvc.yaml' by "
-        "default.",
+        help="Stage names or stage outputs to reproduce. './dvc.yaml' "
+        "by default.",
     ).complete = completion.DVCFILES_AND_STAGE
     repro_parser.add_argument(
         "-f",
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -455,7 +455,7 @@ class Repo:
                 return True
 
             if glob:
-                return fnmatch.fnmatch(out.path_info.name, path)
+                return fnmatch.fnmatch(str(out.path_info), path)
 
             return False
 
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -1,3 +1,4 @@
+import fnmatch
 import logging
 import os
 from contextlib import contextmanager
@@ -436,7 +437,9 @@ class Repo:
         error_handler = self.stage_collection_error_handler
         return self.stage.collect_repo(onerror=error_handler)
 
-    def find_outs_by_path(self, path, outs=None, recursive=False, strict=True):
+    def find_outs_by_path(
+        self, path, outs=None, recursive=False, strict=True, glob=False
+    ):
         # using `outs_graph` to ensure graph checks are run
         outs = outs or self.outs_graph
 
@@ -450,6 +453,9 @@ class Repo:
 
             if recursive and out.path_info.isin(path_info):
                 return True
+
+            if glob:
+                return fnmatch.fnmatch(out.path_info.name, path)
 
             return False
 
--- a/dvc/repo/reproduce.py
+++ b/dvc/repo/reproduce.py
@@ -124,7 +124,10 @@ def reproduce(
     else:
         for target in targets:
             granular_stages = self.stage.collect_granular(
-                target, recursive=recursive, accept_group=accept_group
+                target,
+                recursive=recursive,
+                accept_group=accept_group,
+                glob=glob,
             )
             stages.update([stage_info.stage for stage_info in granular_stages])
 
--- a/dvc/repo/reproduce.py
+++ b/dvc/repo/reproduce.py
@@ -92,7 +92,8 @@ def reproduce(
     from .graph import get_pipeline, get_pipelines
 
     glob = kwargs.pop("glob", False)
-    accept_group = not glob
+    glob_stages = kwargs.pop("glob_stages", False) and not glob
+    accept_group = not glob and not glob_stages
 
     if isinstance(targets, str):
         targets = [targets]
@@ -128,6 +129,7 @@ def reproduce(
                 recursive=recursive,
                 accept_group=accept_group,
                 glob=glob,
+                glob_stages=glob_stages,
             )
             stages.update([stage_info.stage for stage_info in granular_stages])
 
--- a/dvc/repo/reproduce.py
+++ b/dvc/repo/reproduce.py
@@ -92,7 +92,7 @@ def reproduce(
     from .graph import get_pipeline, get_pipelines
 
     glob = kwargs.pop("glob", False)
-    glob_stages = kwargs.pop("glob_stages", False) and not glob
+    glob_stages = kwargs.pop("glob_stages", False)
     accept_group = not glob and not glob_stages
 
     if isinstance(targets, str):
--- a/dvc/repo/reproduce.py
+++ b/dvc/repo/reproduce.py
@@ -123,14 +123,10 @@ def reproduce(
                     stages.add(stage)
     else:
         for target in targets:
-            stages.update(
-                self.stage.collect(
-                    target,
-                    recursive=recursive,
-                    accept_group=accept_group,
-                    glob=glob,
-                )
+            granular_stages = self.stage.collect_granular(
+                target, recursive=recursive, accept_group=accept_group
             )
+            stages.update([stage_info.stage for stage_info in granular_stages])
 
     return _reproduce_stages(self.graph, list(stages), **kwargs)
 
--- a/dvc/repo/stage.py
+++ b/dvc/repo/stage.py
@@ -74,11 +74,12 @@ def _collect_specific_target(
     with_deps: bool,
     recursive: bool,
     accept_group: bool,
+    glob: bool = False,
 ) -> Tuple[StageIter, "OptStr", "OptStr"]:
     from dvc.dvcfile import is_valid_filename
 
     # Optimization: do not collect the graph for a specific target
-    file, name = parse_target(target)
+    file, name = parse_target(target, isa_glob=glob)
 
     # if the target has a file, we can load directly from it.
     if not file:
@@ -90,12 +91,14 @@ def _collect_specific_target(
         logger.debug(msg, target, PIPELINE_FILE)
         if not (recursive and loader.fs.isdir(target)):
             stages = _maybe_collect_from_dvc_yaml(
-                loader, target, with_deps, accept_group=accept_group,
+                loader, target, with_deps, accept_group=accept_group, glob=glob
             )
             if stages:
                 return stages, file, name
     elif not with_deps and is_valid_filename(file):
-        stages = loader.load_all(file, name, accept_group=accept_group)
+        stages = loader.load_all(
+            file, name, accept_group=accept_group, glob=glob
+        )
         return stages, file, name
     return [], file, name
 
@@ -372,6 +375,7 @@ class StageLoad:
         recursive: bool = False,
         graph: "DiGraph" = None,
         accept_group: bool = False,
+        glob: bool = False,
     ) -> List[StageInfo]:
         """Collects a list of (stage, filter_info) from the given target.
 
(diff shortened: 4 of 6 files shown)


dvc/repo/reproduce.py
```diff
@@ -119,14 +119,10 @@ def reproduce(
...
     else:
         for target in targets:
-            stages.update(
-                self.stage.collect(
-                    target,
-                    recursive=recursive,
-                    accept_group=accept_group,
-                    glob=glob,
-                )
+            granular_stages = self.stage.collect_granular(
```
pared, line 122: Hmm, if we are to change this method, we will effectively make `glob` stop working. We should not do that. 
Ideally we should support `glob` for `collect_granular`. 
But, we could also use `collect_granular` as a fallback in a situation when `glob` is not provided and `collect` returns no results. That way it will be easier but we will need to create issue for supporting `glob` in `collect_granular` ad do it in the future anyway.

pared (changes requested):
One comment,
In any case, test in `test_repro` veryfing that repro by output is possible would be nice.

dvc/repo/reproduce.py
Vonski: Second approach is inconsistent, because it would not let user to do `repro` by outs with `glob` option. I'll try to implement `glob` handling logic in `collect_granular`. One more question. There wasn't any tests fails connected to missing `glob` functionality. Should I prepare a test that validates if globbing is possible?

skshetry: @Vonski, this approach also does not work with `accept_group` which supports running multiple `foreach` generated stages, as `collect_granular` does not support that yet (which is yet another thing that introduces an ambiguity for target resolution).

Vonski: Ok, as far as I understand the only acceptable solution is to upgrade `collect_granular` to support globbing and `accept_group`. May I try to do it here?

Vonski:
> One comment,
> In any case, test in `test_repro` veryfing that repro by output is possible would be nice.

I've added one, is this check correct for this functionality?

dvc/repo/reproduce.py
skshetry: `accept_group` seems to be implemented in the `collect_granular`. Sorry for misleading.

pared: @Vonski Sure, go ahead, if you stumble upon any problems, feel free to ping us.

Vonski: Now `collect_granular` handles `glob` also.

pared (approved):

dvc/repo/__init__.py
```diff
@@ -410,6 +413,9 @@ def func(out):
             if recursive and out.path_info.isin(path_info):
                 return True
 
+            if glob:
+                return fnmatch.fnmatch(out.path_info.name, path)
```
skshetry, line 417: `name` is a basename, we should be checking for the `str(path_info)` which should also make it work with `../data/**` kind of globs.

dvc/repo/stage.py
```diff
@@ -299,14 +303,16 @@ def collect_granular(
             return [StageInfo(stage) for stage in self.repo.stages]
 
         stages, file, _ = _collect_specific_target(
-            self, target, with_deps, recursive, accept_group
+            self, target, with_deps, recursive, accept_group, glob=glob
```
skshetry, line 306: Now that `repro` will support output name as a target, I don't think we should `glob` for the stage names at all. Or, we could think of a different flags to glob stage names.

skshetry: Doing this way, if the glob does match within the `dvc.yaml`, it will not try to collect from other stages. Globbing outputs should be global.

skshetry:
cc'ing @dberenbaum @shcheklein

Right now, we have a `--glob` only for the stage names. I'd say that we change it to `--glob-stages` and use `--glob` for outputs-based glob. 🙂

pared:
@Vonski seems like we have a consensus. We need another `glob` flag. As @skshetry mentioned `--glob-stages` should be for globbing in context of stages, and `--glob` should be for outputs. Would you be willing to introduce this?

Vonski:
> @Vonski seems like we have a consensus. We need another `glob` flag. As @skshetry mentioned `--glob-stages` should be for globbing in context of stages, and `--glob` should be for outputs. Would you be willing to introduce this? 

@pared Yeah, sure. I assume that it should be introduced in this PR. Correct me if I'm wrong.

pared:
@Vonski yes, the change shall be included here

skshetry:
@Vonski, ping. What's the status of this? 

Looking at this, it seems we are asking quite a big feature/change. If you feel the same, how about we start by supporting the `repro` from given output name (just the title of the PR)? We could introduce other changes in successive PRs. What do you think? 🙂

Vonski:
@skshetry Honestly, I don't know. I couldn't find time to think about it yet. It looks like today I'll have some, so in 12 hours from now I should be able to answer the question, or maybe even to push some implemented solution. What do you say?

dvc/repo/__init__.py
Vonski: I changed that and added new test to check how globbing with directories in path works.

Vonski:
@skshetry @pared I pushed a preview of my idea of mutually exclusive split of old `glob` flag, into new `glob` for outs paths, and `glob_stages` for stages names. Please check if the logic of `reproduce` is correct. Is this assumption about mutual exclusiveness acceptable?

As I understand, I'll also need to change some code for command repro to accept new `--glob-stages` flag. Is that correct? If so, I would appreciate if you could point out the files that should be modified to achieve that.

pared (commented):
@Vonski 

> Is this assumption about mutual exclusiveness acceptable

Yes, if there will be need to support that, we will continue with this change. For now is fine. We should probably check in the `repro` command and throw Exception mentioning that both cannot be provided.

> As I understand, I'll also need to change some code for command repro to accept new --glob-stages flag.

Correct, here is what I think needs to be done:
- [x] - add `--glob-stages` in [repro command](https://github.com/iterative/dvc/blob/f2b216261ed84cce39449710f04cbe252a5a349e/dvc/command/repro.py#L149) - take a look at `--glob` to see how to do that
- [x] - parse `--glob-stages` in [_repro_kwargs](https://github.com/iterative/dvc/blob/f2b216261ed84cce39449710f04cbe252a5a349e/dvc/command/repro.py#L29)
- [ ] - fix test for parsing [command](https://github.com/iterative/dvc/blob/f2b216261ed84cce39449710f04cbe252a5a349e/tests/unit/command/test_repro.py#L4)

Vonski:
@pared I hope to find some time for it before the end of the week, fyi.

Vonski:
@pared I added requested changes.

dvc/command/repro.py
```diff
@@ -11,6 +11,10 @@
 
 class CmdRepro(CmdBase):
     def run(self):
+        if self.args.glob and self.args.glob_stages:
+            raise ValueError(
+                "--glob and --glob-stages options are mutually exclusive."
+            )
```
pared, line 17: +1

pared (approved):

dvc/command/repro.py
```diff
@@ -150,7 +155,19 @@ def add_arguments(repro_parser):
         "--glob",
         action="store_true",
         default=False,
-        help="Allows targets containing shell-style wildcards.",
+        help=(
+            "Allows targets containing shell-style wildcards to match out "
+            "files.",
```
skshetry, line 160: comma inside the parentheses turn into a tuple. This is making `dvc repro -h` fail.

skshetry (changes requested):
@Vonski, this looks great. I have some minor issues inlined, and I see a few other issues that we might need to solve soon after this PR gets merged:

1. Improving error message
```console
$ dvc repro "model.pkl"   
ERROR: 'model.pkl' does not exist as an output or a stage name in 'dvc.yaml': Stage 'model.pkl' not found inside 'dvc.yaml' file
$ dvc repro --glob "model-*.pkl"
ERROR: 'model-*.pkl' does not exist as an output or a stage name in 'dvc.yaml': Stage 'model-*.pkl' not found inside 'dvc.yaml' file
```
We have to fix this separately and try to provide precise error messages in terms when globbing. The original message is not ideal either.

2. `glob` and `glob-stages`
We were using `glob` for globbing stages in both the cli option and in the internal API. But now, we are inconsistent in our internal API (except `collect_granular`). So, we should migrate other `glob` kwargs to `glob_stages` for reducing confusion. 

But let's do this in separate PRs, and fix minor issues inlined in the review comments.

dvc/command/repro.py
skshetry: Same issue in the `glob-stages` too.

Vonski:
I'll try to do that today/tomorrow fyi.

skshetry (approved):
@Vonski, could you please update the targets help message:
```
Stages to reproduce. 'dvc.yaml' by default.
```

to something more appropriate with the change? And, would you be able to submit a docs patch in the [dvc.org](https://github.com/iterative/dvc.org)?

The change looks good to me otherwise.

Vonski:
@skshetry Yes, sure. I'll try do that today.

Vonski:
Related https://github.com/iterative/dvc.org/pull/2362

dvc/command/repro.py
```diff
@@ -42,14 +42,15 @@ def _repro_kwargs(self):
...
+            "glob_stages": self.args.glob_stages,
         }
 
 
 def add_arguments(repro_parser):
     repro_parser.add_argument(
         "targets",
         nargs="*",
-        help="Stages to reproduce. 'dvc.yaml' by default.",
+        help="Stages or output files to reproduce. 'dvc.yaml' by default.",
```
jorgeorpinel, line 53: could be a dir.
```suggestion
        help="Stages or outputs to reproduce. 'dvc.yaml' by default.",
```

dvc/command/repro.py
```diff
@@ -146,11 +147,24 @@ def add_arguments(repro_parser):
...
-    repro_parser.add_argument(
+    glob_group = repro_parser.add_mutually_exclusive_group()
+    glob_group.add_argument(
         "--glob",
         action="store_true",
         default=False,
-        help="Allows targets containing shell-style wildcards.",
+        help=(
+            "Allows targets containing shell-style wildcards to match out "
+            "files."
```
jorgeorpinel, line 157:
```suggestion
            "Allows targets containing shell-style wildcards to match stage "
            "outputs."
```

dvc/repo/stage.py
```diff
@@ -391,18 +397,25 @@ def collect_granular(
 
             (see `collect()` for other arguments)
         """
+        if glob and glob_stages:
+            raise InvalidArgumentError(
+                "--glob and --glob-stages options are mutually exclusive."
```
jorgeorpinel, line 402:
```suggestion
                "The --glob and --glob-stages options are mutually exclusive."
```

jorgeorpinel:
> Right now, we have a --glob only for the stage names. I'd say that we change it to --glob-stages and use --glob for outputs-based glob.

Q. Couldn't changing the behavior of `--glob` be a breaking change for some users? (Better leave for 3.0?)

jorgeorpinel:
p.s. I do like that it's more consistent with `add/pull/push --glob` now though.

dvc/command/repro.py
jorgeorpinel: Actually, can we avoid the term "output"? Per https://github.com/iterative/dvc.org/pull/2357#pullrequestreview-629433013 cc @shcheklein 

Maybe:
```suggestion
        help="Stages or data files or directories to reproduce. 'dvc.yaml' by default.",
```

jorgeorpinel: TBH I'm not sure it's undesirable to use the term "output" here, since it's part of the dvc.yaml schema (and probably in many UI strings of `dvc`), but let's try.

- [ ] Note for self: review other existing places where "output" can be avoided.

dvc/command/repro.py
jorgeorpinel: Same. Any way to avoid the term "output" here? Thanks

dvc/command/repro.py
Vonski: From the user perspective it seems misleading to me to use "data files" term here. Current code doesn't let you to reproduce anything by dependencies names, which I would assume is possible after reading "data files" in help for targets. We could add more descriptive term like "... files defined as outputs from stages ..." there. What do you think?

shcheklein: my 2cs - I think in this specific case `output files` are fine (note! output files != output)

shcheklein: I mean ... may be even `outputs` are fine @jorgeorpinel :) If we indeed mean them in end-user meaning (a short cut to the results or output files or more precise form that @Vonski mentioned). What I'd like to avoid I guess, is when we use `outputs` for a command that takes a path to _any_ DVC-tracked file, model, directory. E.g. `dvc add`-ed files. They are called "outputs" internally (and even in the .dvc file), but from the readers/end user perspective those are datasets, data, or files, _input_ data (when there is a pipeline).

In this specific case (my understanding) that what we add is an ability to take any "pipeline produced output" and pass it into `repro`. In this case it's more or less clear I think - that yes, there are pipelines - they have inputs and outputs, and you can pass one of those.

Sorry, for a long description, was trying to unload my thought process :)

jorgeorpinel: Got it. So "outputs" should be OK here, but as an alternative we can go with "stage outputs", though in this case we would need something like the following (a little repetitive but clear):
```suggestion
        help="Stage names or stage outputs to reproduce. 'dvc.yaml' by default.",
```

dvc/command/repro.py
jorgeorpinel: UPDATE: This suggestion seems OK after all 👍

dvc/command/repro.py
jorgeorpinel: > 'dvc.yaml' by default.

BTW, that's neither a stage name nor a stage output. Should it be "Everything in /dvc.yaml by default" ?

dvc/command/repro.py
```diff
@@ -42,14 +42,16 @@ def _repro_kwargs(self):
...
         }
 
 
 def add_arguments(repro_parser):
     repro_parser.add_argument(
         "targets",
         nargs="*",
-        help="Stages to reproduce. 'dvc.yaml' by default.",
+        help="Stage names or stage outputs to reproduce. 'dvc.yaml' by "
+        "default.",
```
jorgeorpinel, line 54:
```suggestion
        help="Stage names or stage outputs to reproduce. './dvc.yaml' "
        "by default.",
```

pared:
@jorgeorpinel 
>  Couldn't changing the behavior of --glob be a breaking change for some users? (Better leave for 3.0?)

Yes, this is a breaking change, however, after it, we still would be able to retain "old" behavior with the new flag `--glob-stages`. I think that because of that we should proceed and involve this change in the current version. 

@dberenbaum @shcheklein am I right here?

dberenbaum:
> @jorgeorpinel
> 
> > Couldn't changing the behavior of --glob be a breaking change for some users? (Better leave for 3.0?)
> 
> Yes, this is a breaking change, however, after it, we still would be able to retain "old" behavior with the new flag `--glob-stages`. I think that because of that we should proceed and involve this change in the current version.
> 
> @dberenbaum @shcheklein am I right here?

Hm, if someone is scripting their use of `dvc repro --glob` (like in cml), then this seems like it's going to break their workflow, so I'm not sure it's a good idea to introduce without sufficient warning. Maybe it would be better to go with `--glob/--glob-outs` instead of `--glob/--glob-stages`?

By the way, forgetting about glob, what happens if there is a stage and an output in a different stage with the same name?

pared:
@dberenbaum

> By the way, forgetting about glob, what happens if there is a stage and an output in a different stage with the same name?

In that case, both stages shall be reproduced.

pared:
@dberenbaum 

> I'm not sure it's a good idea to introduce without sufficient warning. Maybe it would be better to go with --glob/--glob-outs instead of --glob/--glob-stages?

It is possible that that we might break some users scripts. I see 2 solutions to this problem:
1. accept the fact that something might fail
2. use `--glob-outs/--glob` approach.

While I am not a fan of breaking the consistency between minor dvc versions, it seems to me that in this case it would be justified  by the fact (noted by @jorgeorpinel) that `add/pull/push` commands target `outputs` when using `--glob`. 

So if we go with `--glob-outs` that puts us in situation where one needs to use different option when using `add/pull/push` (`--glob`) and other one (`--glob-outs`) when using `repro` command. From UI point of view it would seem to me to be tiresome to remember that fact.

dberenbaum:
Before 1.0, users always specified outputs as targets to `repro`, right?

> > By the way, forgetting about glob, what happens if there is a stage and an output in a different stage with the same name?
> 
> In that case, both stages shall be reproduced.

Hmm, so do we recommend users avoid this? How would someone repro one or the other? Is this the same behavior for `push`/`pull`?

> While I am not a fan of breaking the consistency between minor dvc versions, it seems to me that in this case it would be justified by the fact (noted by @jorgeorpinel) that `add/pull/push` commands target `outputs` when using `--glob`.

Makes sense, but those commands are inherently about files and `repro` is inherently about stages. Do `push`/`pull` glob stages?

pared:
> Hmm, so do we recommend users avoid this? How would someone repro one or the other? Is this the same behavior for push/pull?

We don't do that in any way - glob option is there to switch this on or off. If one wants to reproduce only particular output, `dvc repro` should be called for particular stage. Using `glob` is a convenience method allowing user to target multiple stages containing assets matching the `target`. In case of `push`/ `pull` its similar - one should not use `glob` if aiming to interact with some particular target. If users would still like to use glob to repro only particular output, full path needs to be provided.

> Makes sense, but those commands are inherently about files and repro is inherently about stages. Do push/pull glob stages?

No, it does not.

dberenbaum:
Thanks @pared, makes sense!

I'm still unsure about introducing a breaking change in a minor release like this. Would be interested in getting thoughts from others @jorgeorpinel @efiop. Is changing from `--glob` to glob for stages to `--glob`/`--glob-stages` to glob for files/stages enough of a breaking change to warrant holding off until a major release?

efiop:
Btw, `--glob` is like that right now because we want to turn it on by default once we are ready. And we won't be able to turn both --glob and --glob-stages on without mixing them together. And that's actually probably what we really want here in the long term: to make `--glob` work for both stages and outputs and even dvc-files - that would be the best ui, really. So right now it would be nice to merge `--glob` and `--glob-stages` into one `--glob`, if that makes sense.

@skshetry considering the implications that we already have in `collect` logic for outputs/stages/granular outputs, what do you think about ^^^ ?

jorgeorpinel:
> better to go with --glob/--glob-outs ? ... add/pull/push commands target outputs when using --glob.

I'd also err on the side of not breaking anyone's scripts for now and do `--glob/--glob-outs` temporarily.

Also keep in mind `repro` may be phased out into 3.0? In favor of `exp run`. Will all this globbig make sense for `exp run`? (Do you "run" a data artifact)?

> if there is a stage and an output (in a different stage) with the same name ... In that case, both stages shall be reproduced

Would be ideal to error-out instead with an appropriate message, and a way to specify whether you meant the stage or the output. But prob not critical so could be left for a follow-up PR. And again will this all pass on from `repro` to `exp run`? 🙂 

>  would be nice to merge --glob and --glob-stages into one --glob

Agree. Another route to consider: leave `--glob` changes out of this PR is possible (wouldn't support file/dir paths for now).

skshetry:
> And we won't be able to turn both --glob and --glob-stages on without mixing them together.

It will be kinda confusing mixing outputs and stages both for glob. I have been against the change for supporting output names as a target in `repro`, but with `glob` too in the mix, it adds more confusion.

Internally, it complicates things as well, as it's already quite hard to support multiple targets, optimizing for stages names and directories and provide good error message as well. So I'd prefer avoiding it as much as possible.

> even dvc-files

Why would we need dvc-files here?

@dberenbaum, @jorgeorpinel, @efiop, as mentioned by @pared, I don't think the `glob` discussion is worth it for blocking this PR, especially considering the PR is sitting there for more than 3 months now. Let's discuss this in a separate issue.

efiop:
Thanks for a great discussion, guys! :pray: 

We had a great discussion with @skshetry about this today and so, to summarize: we can't merge this as is, as it is breaking backward compatibility and creates a problem for us in the near future when we will enable --glob by default. The latter problem applies to alternative `--glob-outs` approach, so it is also a nogo. That being said, I see that @skshetry has drawn a pretty nice plan in https://github.com/iterative/dvc/pull/5273#issuecomment-800824215 , where he raised a concern that this PR is doing too much, and I tend to agree with him. In order to support outs globbing we need to first support using outputs in a separate PR (along with figuring out a way to distinguish output name from a stage name in a deterministic way, which is not trivial) and then we can come back to the globbing itself that should be part of regular `--glob` (there are lots of questions about distinguishing stages/outputs there too, so it will be quite a substantial effort to get it right).

@Vonski Thank you so much for your contributions! :pray: I hope you are not feeling discouraged by this, it is just a classic issue that turned out to be much more complex than originally anticipated, and all of these discussions and research are arguably an even bigger contribution to the project than the PR itself, so we truly appreciate it. 

So there are three paths we could take from here:

1) close this PR and start from scratch, tackling output support first, merging it and then proceeding with globbing
2) repurpose this PR for only tackling outputs (without globbing) and only then proceeding with globbing in a separate PR
3) considering that the original issue doesn't seem to be in high demand (it is a bit surprising, but still), we might just decide that less is more and not implement it for now, until there is a clear demand from users

@Vonski What do think?

efiop:
Closing for now.

Vonski:
Ah sorry I'm late. I would appreciate to start from scratch and try to push issue forward anyway. I think 1. option is the best idea under the circumstances. I am also a little more engaged in some other things right now, so I have less time to make things happen here, but if it is not a problem, I want to help with that. I hope I'll find some time this week to create another PR for this.

efiop:
@Vonski Sounds good! Thank you! :pray:

PR #5275, opened 2021-01-15
tree: Introduce the new .ls() API

### LINKED ISSUE #5274: tree, import-url: Introduce a new .ls() API for faster dir_info generation (including collection of hash_infos for entries)
Currently, we list files and then fetch every file individually to get their e-tag, which is very costly both in terms of time and the number of API requests we do, so instead, we can retrieve both `size` and `e_tag` directly on the listing (object summaries) and reduce them the amount of requests we do from `n/1000 + n` to only `n/1000`. This shows a %60+ speedup on my local system (low internet speed).

isidentical:
Resolves #5274 

Introduces a new `Tree.ls()` API for listings

master:
```
Importing 's3://dvc-temp/import-url' -> 'import-url'                  
        git add import-url.dvc
dvc import-url s3://dvc-temp/import-url  57,20s user 2,65s system 27% cpu 3:37,47 total
```

this PR:
```
 $ time dvc import-url s3://dvc-temp/import-url
Importing 's3://dvc-temp/import-url' -> 'import-url'                  
        git add import-url.dvc
dvc import-url s3://dvc-temp/import-url  37,69s user 2,57s system 65% cpu 1:01,20 total
```

### DIFF
--- a/dvc/tree/azure.py
+++ b/dvc/tree/azure.py
@@ -23,6 +23,8 @@ class AzureTree(BaseTree):
         "knack": "knack",
     }
     PARAM_CHECKSUM = "etag"
+    DETAIL_FIELDS = frozenset(("etag", "size"))
+
     COPY_POLL_SECONDS = 5
     LIST_OBJECT_PAGE_SIZE = 5000
 
@@ -166,7 +168,6 @@ class AzureTree(BaseTree):
                 yield file_info, {
                     "size": blob.size,
                     "e_tag": blob.etag.strip('"'),
-                    "last_modified": blob.last_modified,
                 }
             else:
                 yield file_info
@@ -186,15 +187,6 @@ class AzureTree(BaseTree):
         )
         properties = blob_client.get_blob_properties()
         return properties.size
-
-    def _iter_hashes(self, path_info, **kwargs):
-        for file_info, details in self.ls(
-            path_info, recursive=True, detail=True
-        ):
-            self._check_ignored(file_info)
-            yield file_info, HashInfo(
-                self.PARAM_CHECKSUM, details["e_tag"], size=details["size"],
-            )
 
     def get_file_hash(self, path_info):
         return HashInfo(self.PARAM_CHECKSUM, self.get_etag(path_info))
--- a/dvc/tree/azure.py
+++ b/dvc/tree/azure.py
@@ -162,14 +162,14 @@ class AzureTree(BaseTree):
         for blob in container_client.list_blobs(
             name_starts_with=path_info.path
         ):
-            file_info = path_info.replace(path=blob.name)
             if detail:
-                yield file_info, {
+                yield {
+                    "name": blob.name,
                     "size": blob.size,
                     "etag": blob.etag,
                 }
             else:
-                yield file_info
+                yield blob.name
 
     def remove(self, path_info):
         if path_info.scheme != self.scheme:
--- a/dvc/tree/azure.py
+++ b/dvc/tree/azure.py
@@ -154,6 +154,8 @@ class AzureTree(BaseTree):
             yield path_info.replace(path=fname)
 
     def ls(self, path_info, recursive=False, detail=False):
+        assert recursive
+
         container_client = self.blob_service.get_container_client(
             path_info.bucket
         )
@@ -161,13 +163,10 @@ class AzureTree(BaseTree):
             name_starts_with=path_info.path
         ):
             file_info = path_info.replace(path=blob.name)
-            if not recursive and file_info.parent != path_info:
-                continue
-
             if detail:
                 yield file_info, {
                     "size": blob.size,
-                    "e_tag": blob.etag.strip('"'),
+                    "etag": blob.etag,
                 }
             else:
                 yield file_info
--- a/dvc/tree/azure.py
+++ b/dvc/tree/azure.py
@@ -153,7 +153,9 @@ class AzureTree(BaseTree):
 
             yield path_info.replace(path=fname)
 
-    def ls(self, path_info, recursive=False, detail=False):
+    def ls(
+        self, path_info, detail=False, recursive=False
+    ):  # pylint: disable=arguments-differ
         assert recursive
 
         container_client = self.blob_service.get_container_client(
--- a/dvc/tree/azure.py
+++ b/dvc/tree/azure.py
@@ -151,6 +151,25 @@ class AzureTree(BaseTree):
 
             yield path_info.replace(path=fname)
 
+    def _dir_info(self, path_info):
+        container_client = self.blob_service.get_container_client(
+            path_info.bucket
+        )
+        for blob in container_client.list_blobs(
+            name_starts_with=path_info.path
+        ):
+            yield blob.name, {
+                "size": blob.size,
+                "e_tag": blob.etag.strip('"'),
+                "last_modified": blob.last_modified,
+            }
+
+    def info(self, path_info):
+        if self.isfile(path_info):
+            _, details = next(self._dir_info(path_info))
+            return details
+        return self._dir_info(path_info / "")
+
     def remove(self, path_info):
         if path_info.scheme != self.scheme:
             raise NotImplementedError
@@ -166,6 +185,13 @@ class AzureTree(BaseTree):
         )
         properties = blob_client.get_blob_properties()
         return properties.size
+
+    def _iter_hashes(self, path_info, **kwargs):
+        for file_info, details in self.info(path_info):
+            self._check_ignored(file_info)
+            yield file_info, HashInfo(
+                self.PARAM_CHECKSUM, details["e_tag"], size=details["size"],
+            )
 
     def get_file_hash(self, path_info):
         return HashInfo(self.PARAM_CHECKSUM, self.get_etag(path_info))
--- a/dvc/tree/azure.py
+++ b/dvc/tree/azure.py
@@ -151,24 +151,25 @@ class AzureTree(BaseTree):
 
             yield path_info.replace(path=fname)
 
-    def _dir_info(self, path_info):
+    def ls(self, path_info, recursive=False, detail=False):
         container_client = self.blob_service.get_container_client(
             path_info.bucket
         )
         for blob in container_client.list_blobs(
             name_starts_with=path_info.path
         ):
-            yield blob.name, {
-                "size": blob.size,
-                "e_tag": blob.etag.strip('"'),
-                "last_modified": blob.last_modified,
-            }
+            file_info = path_info.replace(path=blob.name)
+            if not recursive and file_info.parent != path_info:
+                continue
 
-    def info(self, path_info):
-        if self.isfile(path_info):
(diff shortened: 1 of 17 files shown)


efiop (commented):
Sorry for the delay.

Great research! The PR looks good, but I'm wondering if there is a way to properly generalize this  for our trees. Something interesting that pyarrow uses with their fsspec filesystems is a FileSelector https://arrow.apache.org/docs/python/generated/pyarrow.fs.FileSelector.html#pyarrow.fs.FileSelector . E.g. we use it in our HDFSTree right now for walking https://github.com/iterative/dvc/blob/master/dvc/tree/hdfs.py#L103 . Maybe we could do something similar here too? WDYT?

dvc/tree/azure.py
```diff
@@ -151,6 +151,25 @@ def walk_files(self, path_info, **kwargs):
...
+            name_starts_with=path_info.path
+        ):
+            yield blob.name, {
+                "size": blob.size,
+                "e_tag": blob.etag.strip('"'),
+                "last_modified": blob.last_modified,
+            }
+
+    def info(self, path_info):
+        if self.isfile(path_info):
```
efiop, line 168: This `isfile` wastes an API call and `info()`(like stat()) should return info for the path that is given, not try to walk it recursively (e.g. how fsspec's stat/info does).

dvc/tree/s3.py
```diff
@@ -227,6 +228,20 @@ def walk_files(self, path_info, **kwargs):
...
+        assert recursive
+
+        with self._get_bucket(path_info.bucket) as bucket:
+            for obj_summary in bucket.objects.filter(Prefix=path_info.path):
+                file_info = path_info.replace(path=obj_summary.key)
+                if detail:
+                    yield file_info, {
+                        "size": obj_summary.size,
+                        "etag": obj_summary.e_tag.strip('"'),
+                    }
```
efiop, line 241: This should return just a dict with a string as a name. Creating lots of `file_info`s will be really costly.

isidentical: > Also doesn't comply with fsspec

I am aware of this, though I thought it would comply with the general sense of our trees (exchanging file_info's everywhere [walk_files, upload, download, etc.]).

efiop: I see, you are right. But let's try to comply with fsspec if we can. This looks like a good place to do it, as you'll need to call `.replace` only once instead of in every `ls` implementation.

efiop:
Btw, what about other clouds? E.g. gs, gdrive, etc?

isidentical:
> Btw, what about other clouds? E.g. gs, gdrive, etc?

I didn't check them, yet. I guess it requires a bit of more research. Possibly I'll apply the same approach as `upload_fobj`, and submit another pull request for the remaining tree providers with a similar API (even though they don't offer *hash* information).

tests/remotes/webdav.py
```diff
@@ -17,19 +16,27 @@
...
+    _DIR_PATH = None
+
     @staticmethod
     def get_url(port):  # pylint: disable=arguments-differ
         return f"webdav://localhost:{port}"
 
     def mkdir(self, mode=0o777, parents=False, exist_ok=False):
-        assert parents
-        makedirs(self.path, exist_ok=exist_ok)
+        self.dir_path.mkdir(parents=parents, exist_ok=True)
```
isidentical, line 26: @efiop this is because, when you are generating a directory with 2 files in it (`{'data': {'foo': 'foo', 'bar': 'bar'}}`) it tries to recreate the parent for 2 times, which results with a `DirectoryExists` error. 
https://github.com/iterative/dvc/blob/fc77618ad1eb9d0bb306a85e4f750660b5285318/tests/remotes/base.py#L43

Apparently all other trees have some sort of `assert not exist_ok` in their `mkdir()` implementation, so instead of passing this in the `Base.gen()` I just fixed here.
https://github.com/iterative/dvc/blob/fc77618ad1eb9d0bb306a85e4f750660b5285318/tests/remotes/s3.py#L77
https://github.com/iterative/dvc/blob/fc77618ad1eb9d0bb306a85e4f750660b5285318/tests/remotes/ssh.py#L84
etc.

tests/remotes/webdav.py
```diff
@@ -17,19 +16,25 @@
...
     def write_bytes(self, contents):
-        Path(self.path).write_bytes(contents)
+        self.dir_path.write_bytes(contents)
 
     def write_text(self, contents, encoding=None, errors=None):
-        Path(self.path).write_text(contents, encoding=encoding, errors=errors)
+        self.dir_path.write_text(contents, encoding=encoding, errors=errors)
+
+    @property
+    def dir_path(self):
```
isidentical, line 35: The `self.path` points to the root `/data` rather then a relative position in the given temporary directory, since it tries to obtain it from the url (`webdav://stuff:stufff/stuff`). So this points to the temporary directory location

dvc/tree/gs.py
```diff
@@ -143,6 +144,22 @@ def walk_files(self, path_info, **kwargs):
...
+        assert recursive
+
+        for blob in self.gs.bucket(path_info.bucket).list_blobs(
+            prefix=path_info.path
+        ):
+            if detail:
+                yield {
+                    "type": "file",
+                    "name": blob.name,
+                    "md5": blob.md5_hash.encode().hex(),
```
efiop, line 157: Unrelated to this PR since we've been using it for some time, but `md5_hash` is not supported for composite objects :slightly_frowning_face: And I don't remember when we did that switch from etag and why :slightly_frowning_face:

dvc/tree/base.py
```diff
@@ -202,6 +204,9 @@ def walk_files(self, path_info, **kwargs):
         """
         raise NotImplementedError
 
+    def ls(self, path_info, recursive=False, detail=False):
+        raise RemoteActionNotImplemented("ls", self.scheme)
```
efiop, line 208: Let's try match fsspec's ls as much as we can. `recursive` is our extension supported only by some clouds.
```suggestion
    def ls(self, path_info, detail=False, **kwargs):
        raise RemoteActionNotImplemented("ls", self.scheme)
```

tests/remotes/webdav.py
efiop: So now `exist_ok` is just ignored, which is not nice :( We should go the other way around and pass exist_ok=True in `_gen`.

efiop: Those asserts in s3 and others are just defensive ones and could be changed accordingly.

isidentical: > Those asserts in s3 and others are just defensive ones and could be changed accordingly.

So should I remove all of them?

tests/remotes/webdav.py
efiop: Could we do it the way FakeHadoopFileSystem does it? `Webdav._DIR_PATH = Path(directory)` is awkward.

tests/remotes/webdav.py
efiop: @isidentical Only where it makes sense.

isidentical: Well, this is going to be passed to all of the trees (wherever you use `.gen()`)

efiop: @isidentical I understand. It does make sense for gen to pass `exist_ok=True` to makedirs, as you've explained above :pray: . So it makes to adjust the incorrect/improper asserts in particular makedirs implementations. E.g. for `s3` there are no dirs at all and that assert is just defensive to make sure we are not doing anything weird, which means that we are totally fine removing it.

tests/remotes/webdav.py
isidentical: @efiop one problem with initializing this in the init rather than as a class var is that, we won't be inheriting it in the sub paths and then there would be no point unless I am missing something

efiop: @isidentical Hm, not sure I follow. We have `_hdfs_root` in hdfs, that we use as a "server root", which seems to have the same meaning as `_DIR_PATH`.

isidentical: So are you proposing to put it as a global-var?

efiop: @isidentical the root? Yeah, if we can. Notice how we are then using uuid to isolate. So we emulate having one real server that way.

isidentical: I think it should work now, though we don't clean the directory though I am not sure whether we need an `atexit.register()` call or just let it handled by the interpreter close.

tests/remotes/webdav.py
```diff
@@ -1,44 +1,52 @@
...
 def webdav_server(test_config, tmp_path_factory):
     test_config.requires("webdav")
 
     host, port = "localhost", 0
-    directory = os.fspath(tmp_path_factory.mktemp("http"))
-    dirmap = {"/": directory}
+    root_id = str(uuid.uuid4())
+    root_dir = _WEBDAV_DIR / root_id
+    root_dir.mkdir()
+    dirmap = {f"/{root_id}": os.fspath(_WEBDAV_DIR / root_id)}
```
efiop, line 49: Do we actually need the dirmap? The root_id situation is pretty confusing.

Seems like we could just do
```
url = f"webdav://localhost:{webdav_server.port}/{uuid.uuid4()}"
```
in `webdav` fixture and remove `get_url` method (it is a legacy one, we don't really need it, iirc).

efiop: > Do we actually need the dirmap?

I mean we need the dirmap for `/` to point to _WEBDAV_DIR, but do we need the `root_id` mapping? I might be missing something.

isidentical: > in webdav fixture and remove get_url method (it is a legacy one, we don't really need it, iirc).

This is how we actually isolate each test's own directory

efiop: Ah, maybe I'm not understanding the `dirmap` correctly. So we have to specify those subdirs in the dirmap or otherwise we won't be able to create/access them?

efiop: @skshetry Could you take a look at webdav fixture changes, please? We might be missing something.

skshetry: Looks like I did not test the `write_text`/`mkdir` and seems to have been never used before, so didn't notice.

Instead of depending on a certain directory, maybe it'd better to just use a WebDAV client here, similar to what we do for http. What do you guys think? I'll make a PR to update it if we decide so.

skshetry: I guess, in the longer term, we should move away from this `PathInfo` to something that makes it easy to share a state.

efiop: @skshetry Yeah, that would be pretty good as well. You mean a follow-up PR, right?

skshetry: @efiop, no need to block @isidentical on this, I think.  But, could create a PR now if you like.

efiop: @skshetry Ok, then let's keep this PR as is for now. Just as long as you are fine with the webdav changes for now. Thank you! :pray:

PR #5277, opened 2021-01-15
hdfs: copy: don't read everything into memory

efiop:
Regression from #4973 and #4747

### DIFF
--- a/dvc/tree/hdfs.py
+++ b/dvc/tree/hdfs.py
@@ -2,6 +2,7 @@ import io
 import logging
 import os
 import re
+import shutil
 import subprocess
 from collections import deque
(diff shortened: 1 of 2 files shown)


dvc/tree/hdfs.py
```diff
@@ -162,7 +163,7 @@ def copy(self, from_info, to_info, **_kwargs):
                     with closing(
                         hdfs.open_output_stream(tmp_info.path)
                     ) as tmp_fobj:
-                        tmp_fobj.write(from_fobj.read())
```
efiop, line 165: No pbars in `copy()` methods (--external scenario), but that's a common issue in all clouds, so not touching it here yet.

PR #5279, opened 2021-01-16
tests: add basic [web]hdfs mockers

efiop:
Raw POC for hdfs and webhdfs. Both mockers are library-level, not emulating real hdfs protocol or http.

Now all [web]hdfs tests run on all operating systems (previously only linux).

### DIFF
--- a/dvc/tree/webhdfs.py
+++ b/dvc/tree/webhdfs.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import posixpath
 import shutil
 import threading
 from contextlib import contextmanager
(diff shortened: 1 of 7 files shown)


dvc/tree/webhdfs.py
```diff
@@ -105,10 +107,9 @@ def walk_files(self, path_info, **kwargs):
             return
 
         root = path_info.path
-        for path, _, files in self.hdfs_client.walk(root):
-            for file_ in files:
-                path = os.path.join(path, file_)
```
efiop, line 110: This was modifying the `path` used during walking.

PR #5280, opened 2021-01-17
term: replace 'DVC-file' and 'stage file' with correct terms (#3960)

### LINKED ISSUE #3960: term: don't use "DVC-file" or "stage file" anymore, use ".dvc file"
To match iterative/dvc.org/issues/1366 — which does not include updating any `dvc` command example output.

Let's update all user output strings please, and hopefully debug output too. And then open a docs PR to reflect any of these changes, as usual.

DiPaolo:
Fixes #3960 

Documentation on dvc.org was already updated before.

### DIFF
--- a/dvc/command/repro.py
+++ b/dvc/command/repro.py
@@ -92,8 +92,8 @@ def add_arguments(repro_parser):
         "--pipeline",
         action="store_true",
         default=False,
-        help="Reproduce the whole pipeline that the specified stage file "
-        "belongs to.",
+        help="Reproduce the whole pipeline that the specified targets "
+        "belong to.",
     )
     repro_parser.add_argument(
         "-P",
--- a/dvc/dependency/base.py
+++ b/dvc/dependency/base.py
@@ -17,7 +17,7 @@ class DependencyIsNotFileOrDirError(DvcException):
 
 class DependencyIsStageFileError(DvcException):
     def __init__(self, path):
-        super().__init__(f"Stage file '{path}' cannot be a dependency.")
+        super().__init__(f"DVC file '{path}' cannot be a dependency.")
 
 
 class BaseDependency:
--- a/dvc/dvcfile.py
+++ b/dvc/dvcfile.py
@@ -70,9 +70,9 @@ def is_lock_file(path):
 def check_dvc_filename(path):
     if not is_valid_filename(path):
         raise StageFileBadNameError(
-            "bad DVC-file name '{}'. DVC-files should be named "
-            "'Dvcfile' or have a '.dvc' suffix (e.g. '{}.dvc').".format(
-                relpath(path), os.path.basename(path)
+            "bad DVC file name '{}'. DVC files should be named "
+            "'{}' or have a '.dvc' suffix (e.g. '{}.dvc').".format(
+                relpath(path), PIPELINE_FILE, os.path.basename(path)
             )
         )
 
@@ -118,7 +118,7 @@ class FileMixin:
     def _load(self):
         # it raises the proper exceptions by priority:
         # 1. when the file doesn't exists
-        # 2. filename is not a DVC-file
+        # 2. filename is not a DVC file
         # 3. path doesn't represent a regular file
         # 4. when the file is git ignored
         if not self.exists():
--- a/dvc/output/base.py
+++ b/dvc/output/base.py
@@ -46,7 +46,7 @@ class OutputAlreadyTrackedError(DvcException):
 
 class OutputIsStageFileError(DvcException):
     def __init__(self, path):
-        super().__init__(f"Stage file '{path}' cannot be an output.")
+        super().__init__(f"DVC file '{path}' cannot be an output.")
 
 
 class OutputIsIgnoredError(DvcException):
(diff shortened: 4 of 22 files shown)


.pylintrc
```diff
@@ -178,6 +178,9 @@ disable=format,
         # we use it to speedup/optimize
         import-outside-toplevel,
         fixme,
+        # temporarily; should be fixed in pylint 2.6.1 or later
+        # (see https://github.com/PyCQA/pylint/issues/3882)
+        unsubscriptable-object,
```
pmrowla, line 183: Don't think this is something that needs to be committed as it is not failing in CI?

dvc/command/repro.py
```diff
@@ -92,7 +92,7 @@ def add_arguments(repro_parser):
         "--pipeline",
         action="store_true",
         default=False,
-        help="Reproduce the whole pipeline that the specified stage file "
+        help="Reproduce the whole pipeline that the specified targets "
         "belongs to.",
```
pmrowla, line 96:
```suggestion
        "belong to.",
```

dvc/dvcfile.py
```diff
@@ -65,7 +65,7 @@ def is_lock_file(path):
 def check_dvc_filename(path):
     if not is_valid_filename(path):
         raise StageFileBadNameError(
-            "bad DVC-file name '{}'. DVC-files should be named "
+            "bad DVC file name '{}'. DVC files should be named "
             "'Dvcfile' or have a '.dvc' suffix (e.g. '{}.dvc').".format(
```
pmrowla, line 69: I think we can get rid of the `Dvcfile` part of this error message entirely since we are completely dropping single-stage support in 2.0?

pmrowla (changes requested):
Thanks for the PR, just a few minor suggestions.

Also, this needs a rebase now after some of the changes in master.

.pylintrc
DiPaolo: Reverted

dvc/command/repro.py
DiPaolo: Thanks. Fixed

dvc/dvcfile.py
DiPaolo: I would leave it as it is for now. Later it can be changed closely to 2.0. Anyway, I'm not so informed about 2.0 specific, so it could be difficultly to me to change the message.

dvc/command/repro.py
DiPaolo: [offtopic] @pmrowla should I click **Resolve conversation** button here when I fixed a reviewer's note? Or the reviewer should do it himself?

dvc/dvcfile.py
pmrowla: The current master branch is for 2.0 so it is safe to change it. The message should be something like:

`bad DVC file name '{}'. DVC files should be named '{PIPELINE_FILE}' or have a '.dvc' suffix (e.g. '{}.dvc')`

dvc/command/repro.py
pmrowla: For something minor like this it's fine to mark it resolved yourself, but if you're unsure about something it's also fine to leave it for the reviewer to decide

dvc/output/base.py
```diff
@@ -116,7 +116,7 @@ def __init__(
         # This output (and dependency) objects have too many paths/urls
         # here is a list and comments:
         #
-        #   .def_path - path from definition in stage file
+        #   .def_path - path from definition in dvc.yaml file
```
pmrowla, line 119:
```suggestion
        #   .def_path - path from definition in DVC file
```

dvc/output/base.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, path):
 
 class OutputIsStageFileError(DvcException):
     def __init__(self, path):
-        super().__init__(f"Stage file '{path}' cannot be an output.")
+        super().__init__(f"dvc.yaml file '{path}' cannot be an output.")
```
pmrowla, line 49:
```suggestion
        super().__init__(f"DVC file '{path}' cannot be an output.")
```

dvc/dependency/base.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, path):
 
 class DependencyIsStageFileError(DvcException):
     def __init__(self, path):
-        super().__init__(f"Stage file '{path}' cannot be a dependency.")
+        super().__init__(f"dvc.yaml file '{path}' cannot be a dependency.")
```
pmrowla, line 20:
```suggestion
        super().__init__(f"DVC file '{path}' cannot be a dependency.")
```

pmrowla (changes requested):
Internally, outputs and dependencies apply to both pipeline (`dvc.yaml`) outputs and data files (`.dvc`) so there's a few places we should still be using the generic "DVC file"

pmrowla:
hey @DiPaolo, any updates on this?

DiPaolo:
@pmrowla I've pushed an update.

pmrowla (approved):

pmrowla:
Thanks for the PR!

PR #5281, opened 2021-01-18
use `repo.stage.create_from_cli` to create stages

skshetry:
Refactoring and introducing `repo.stage.create_from_cli` instead of using ugly `utils` (directly).

Also clarified API names, docstring and code, especially around `validate_state`/`check_graphs`.

### DIFF
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -280,7 +280,7 @@ class Repo:
 
         self.scm.ignore_list(flist)
 
-    def check_modified_graph(self, new_stages):
+    def check_modified_graph(self, new_stages, old_stages=None):
         """Generate graph including the new stage to check for errors"""
         # Building graph might be costly for the ones with many DVC-files,
         # so we provide this undocumented hack to skip it. See [1] for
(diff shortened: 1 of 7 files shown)


dvc/repo/__init__.py
```diff
@@ -280,7 +280,7 @@ def _ignore(self):
 
         self.scm.ignore_list(flist)
 
-    def check_modified_graph(self, new_stages):
+    def check_modified_graph(self, new_stages, old_stages=None):
```
pared, line 283: Does it make sense to have "old_stages"? Since `repo.stages` are called in `validate_state`, won't they be already cached here?

pared (commented):
LGTM, one question.

dvc/repo/__init__.py
skshetry: Yes, they will be cached if `force=True` in the `validate_state`. And, in that case, we are (force) removing an existing stage with the new one in a new (shallow) copy of `repo.stages` to check for the graph. 

Otherwise, `repo.stages` is used if `force=False`.

pared (approved):

PR #5282, opened 2021-01-18
Introduce stage:list command with autocompletions

skshetry:
Autocompletion only limited to the `dvc repro` for now.

### DIFF
--- a/dvc/command/completion.py
+++ b/dvc/command/completion.py
@@ -8,10 +8,7 @@ from dvc.command.base import CmdBaseNoRepo, append_doc_link
 logger = logging.getLogger(__name__)
 FILE = shtab.FILE
 DIR = shtab.DIRECTORY
-DVC_FILE = {
-    "bash": "_dvc_compgen_DVCFiles",
-    "zsh": "_files -g '(*?.dvc|Dvcfile|dvc.yaml)'",
-}
+
 PREAMBLE = {
     "bash": """
 # $1=COMP_WORDS[1]
@@ -21,8 +18,59 @@ _dvc_compgen_DVCFiles() {
   compgen -f -X '!*Dvcfile' -- $1
   compgen -f -X '!*dvc.yaml' -- $1
 }
-""",
-    "zsh": "",
+
+_dvc_compgen_stages() {
+    local _dvc_stages=($(dvc stage list -q --names-only))
+    compgen -W "${_dvc_stages[*]}" -- $1
+}
+_dvc_compgen_stages_and_files() {
+    _dvc_compgen_DVCFiles $1
+    _dvc_compgen_stages $1
+}
+    """,
+    "zsh": """
+_dvc_compadd_DVCFiles() {
+    _files -g '(*?.dvc|Dvcfile|dvc.yaml)'
+}
+_dvc_compadd_stages() {
+    # this will also show up the description of the stages
+    _describe 'stages' "($(_dvc_stages_output))"
+}
+
+_dvc_stages_output() {
+  dvc stage list -q | awk '{
+    # escape possible `:` on the stage name
+    sub(/:/, "\\\\\\\\:", $1);
+    # read all of the columns except the first
+    # reading `out` from $2, so as not to have a leading whitespace
+    out=$2; for(i=3;i<=NF;i++){out=out" "$i};
+    # print key, ":" and then single-quote the description
+    # colon is a delimiter used by `_describe` to separate field/description
+    print $1":""\\047"out"\\047"
+    # single quote -> \\047
+    }'
+}
+
+_dvc_compadd_stages_and_files() {
+    _dvc_compadd_DVCFiles
+    _dvc_compadd_stages
+}
+    """,
+}
+
+DVC_FILE = {
+    "bash": "_dvc_compgen_DVCFiles",
+    "zsh": "_dvc_compadd_DVCFiles",
+}
+
+STAGE = {
+    "bash": "_dvc_compgen_stages",
+    "zsh": "_dvc_compadd_stages",
+}
+
+DVCFILES_AND_STAGE = {
+    "bash": "_dvc_compgen_stages_and_files",
+    "zsh": "_dvc_compadd_stages_and_files",
 }
 
 
--- a/dvc/command/stage.py
+++ b/dvc/command/stage.py
@@ -302,7 +302,13 @@ def add_parser(subparsers, parent_parser):
         formatter_class=argparse.RawDescriptionHelpFormatter,
     )
     stage_list_parser.add_argument(
-        "targets", nargs="*", default=["dvc.yaml"], help="Stages to list from",
+        "targets",
+        nargs="*",
+        default=["dvc.yaml"],
+        help=(
+            "Show stages from a dvc.yaml/.dvc file or a directory. "
+            "'dvc.yaml' by default"
+        ),
     )
     stage_list_parser.add_argument(
         "--all",
@@ -321,7 +327,7 @@ def add_parser(subparsers, parent_parser):
         "--recursive",
         action="store_true",
         default=False,
-        help="List all stages inside the directory.",
+        help="List all stages inside the specified directory.",
     )
     stage_list_parser.add_argument(
         "--names-only",
--- a/dvc/command/stage.py
+++ b/dvc/command/stage.py
@@ -1,10 +1,108 @@
 import argparse
 import logging
+from itertools import chain, filterfalse
+from typing import TYPE_CHECKING, Dict, Iterable, List
 
 from dvc.command import completion
 from dvc.command.base import CmdBase, append_doc_link, fix_subparsers
 
+if TYPE_CHECKING:
+    from dvc.output.base import BaseOutput
+    from dvc.stage import Stage
+
 logger = logging.getLogger(__name__)
+
+MAX_TEXT_LENGTH = 80
+ELLIPSIS = "…"
+
+
+def generate_description(stage: "Stage") -> str:
+    def part_desc(outs: Iterable["BaseOutput"]) -> str:
+        return ", ".join(out.def_path for out in outs)
+
+    if not stage.deps and not stage.outs:
+        return "No outputs or dependencies."
+
+    if not stage.outs and stage.deps:
+        return "Stage depends on " + part_desc(stage.deps)
+
+    def is_plot_or_metric(out: "BaseOutput"):
+        return bool(out.plot) or bool(out.metric)
+
+    desc: List[str] = []
+
+    outs = list(filterfalse(is_plot_or_metric, stage.outs))
+    if outs:
+        desc.append("Produces " + part_desc(outs))
+
+    plots_and_metrics = list(filter(is_plot_or_metric, stage.outs))
+    if plots_and_metrics:
+        desc.append("Generates " + part_desc(plots_and_metrics))
+
+    return "; ".join(desc)
+
+
+def prepare_description(
+    stage: "Stage", max_length: int = MAX_TEXT_LENGTH
+) -> str:
+    desc = stage.short_description() or generate_description(stage)
+    if len(desc) > max_length:
+        return desc[: max_length - 1] + ELLIPSIS
+    return desc
+
+
+def prepare_stages_data(
+    stages: Iterable["Stage"],
+    description: bool = True,
+    max_length: int = MAX_TEXT_LENGTH,
+) -> Dict[str, str]:
+    return {
+        stage.addressing: prepare_description(stage, max_length=max_length)
+        if description
+        else ""
+        for stage in stages
+    }
+
+
+class CmdStageList(CmdBase):
+    def _get_stages(self) -> Iterable["Stage"]:
+        if self.args.all:
+            stages: List["Stage"] = self.repo.stages  # type: ignore
+            logger.trace(  # type: ignore[attr-defined]
+                "%d no. of stages found", len(stages)
+            )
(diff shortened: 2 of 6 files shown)


dvc/command/completion.py
```diff
@@ -21,8 +18,59 @@
...
+    _dvc_compgen_stages $1
+}
+    """,
+    "zsh": """
+_dvc_compadd_DVCFiles() {
+    _files -g '(*?.dvc|Dvcfile|dvc.yaml)'
+}
+_dvc_compadd_stages() {
+    # this will also show up the description of the stages
+    _describe 'stages' "($(_dvc_stages_output))"
```
skshetry, line 37: Maybe it'd be better to create a hidden flag for completions instead of using `awk`?

pared: I think it makes sense, that would make completions "test"-able to some extent.

pared (approved):
Played around a bit on my own machine. There seems to be a bit of delay, but its much better than when we were testing it on https://github.com/iterative/dvc/pull/5191 :rocket:

skshetry:
> There seems to be a bit of delay

Sadly, most of the time is just the startup time. If you are not in `shtab >= 1.3.4` version, then it might be `pkg_resources` as well. Or, if you are in Python3.9, don't forget to `pip install ruamel.yaml.clib`.

For the example-get-started, it takes ~300ms and for a quite large dvc.yaml file with ~100 stages, it takes ~400ms.

dvc/command/stage.py
```diff
@@ -194,3 +292,41 @@ def add_parser(subparsers, parent_parser):
...
+    STAGE_LIST_HELP = "List stages."
+    stage_list_parser = stage_subparsers.add_parser(
+        "list",
+        parents=[parent_parser],
+        description=append_doc_link(STAGE_LIST_HELP, "stage/list"),
+        help=STAGE_LIST_HELP,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+    stage_list_parser.add_argument(
+        "targets", nargs="*", default=["dvc.yaml"], help="Stages to list from",
```
shcheklein, line 305: Files to list from?

skshetry: @shcheklein, it can also be a directory/foreach-stage-name/stage-name.

shcheklein: kk, my point that this combination `default=["dvc.yaml"]` and `"Stages to list from"` sounds strange a bit and it's not clear what does it exactly expect/can take.

skshetry: This has been a common problem in most of our commands these days as they all support these kinds of target spec. I don't have a good idea to make it concise and also document it clearly.

I changed it to the following (removed `or a stage name` to keep it _short_ for now):
> Show stages from a `dvc.yaml`/`.dvc` file or a directory. 'dvc.yaml' by default.

I looked into `pip install` and `git add`, and they mostly seem to not explain positional arguments.

1. `pip`
```console
$ pip install -h

Usage:   
  pip install [options] <requirement specifier> [package-index-options] ...
  pip install [options] -r <requirements file> [package-index-options] ...
  pip install [options] [-e] <vcs project url> ...
  pip install [options] [-e] <local project path> ...
  pip install [options] <archive url/path> ...
```

2. `git`
```console
$ git add -h
usage: git add [<options>] [--] <pathspec>...
```

maybe we could do similar to pip in our commands?

efiop (approved):

dvc/command/stage.py
shcheklein: You are right, but in this case it should be less generic as `targets`. In case of pip `local project path` is a good example. It doesn't require any further explanations.

Git is so-so. They use `pathspec` but they are very consistent and it's easy to google it (it's very specific). Also, they have examples usually around.

shcheklein:
A few quick notes on the workflow:

- since it affects UI/UX - it would be great to see some summary to review and ask @jorgeorpinel or @dberenbaum to review messages/descriptions
- do we have a ticket for dvc.org on this? (it's checked, but I don't see it)

dvc/command/stage.py
```diff
@@ -1,11 +1,109 @@
...
+
+def generate_description(stage: "Stage") -> str:
+    def part_desc(outs: Iterable["BaseOutput"]) -> str:
+        return ", ".join(out.def_path for out in outs)
+
+    if not stage.deps and not stage.outs:
+        return "No outputs or dependencies."
+
+    if not stage.outs and stage.deps:
+        return "Stage depends on " + part_desc(stage.deps)
```
jorgeorpinel, line 27: Maybe just "Depends on " since other strings are short like that.

dvc/command/stage.py
```diff
@@ -1,11 +1,109 @@
...
+
+    def is_plot_or_metric(out: "BaseOutput"):
+        return bool(out.plot) or bool(out.metric)
+
+    desc: List[str] = []
+
+    outs = list(filterfalse(is_plot_or_metric, stage.outs))
+    if outs:
+        desc.append("Produces " + part_desc(outs))
```
jorgeorpinel, line 36: "Outputs " ?

dvc/command/stage.py
```diff
@@ -1,11 +1,109 @@
...
+
+    plots_and_metrics = list(filter(is_plot_or_metric, stage.outs))
+    if plots_and_metrics:
+        desc.append("Generates " + part_desc(plots_and_metrics))
```
jorgeorpinel, line 40: "Outputs metric: "

jorgeorpinel (commented):
Stage list description UI text suggestions

jorgeorpinel:
(Most of the strings in https://github.com/iterative/dvc/pull/5282/files#diff-cd4d8ab7bbb98bdac7ebf8f89f4c65961526eea6bd78a142e11d4da9835e2bf7 are not covered in tests according to Codecov BTW, not sure if that's a problem)

dvc/command/stage.py
skshetry: That's too long. I'd onl like to add a single verb here.

jorgeorpinel: OK. How is it indicated that this refers to metrics or plots? Would be great to have an updated example or summary as suggested by Ivan 🙂

jorgeorpinel: Yep I tried this in a test repo I have and it's not clear what the difference between "Produces" (soon Outputs?) and "Generates" is.

```console
$ dvc stage list
ab   Produces ab
abm  Generates ab.met
```

jorgeorpinel: Maybe also use "Outputs" and append "metric" or "plot" at the end of the line?

dvc/command/stage.py
skshetry: We could also replace similar to the pip with:
```console
$ dvc stage list -h

Usage:
	dvc stage list [path to dvc.yaml]...
	dvc stage list -R <dir>...
	dvc stage list [path to dvc.yaml][:<stage_name>]...
	dvc stage list --all
```

dvc/command/stage.py
```diff
@@ -194,3 +292,47 @@ def add_parser(subparsers, parent_parser):
...
+        help=(
+            "Show stages from a dvc.yaml/.dvc file or a directory. "
+            "'dvc.yaml' by default"
+        ),
+    )
+    stage_list_parser.add_argument(
+        "--all",
+        action="store_true",
+        default=False,
+        help="List all of the stages in the repo.",
```
jorgeorpinel, line 317: ``"List all of the stages in the repo regardless of `targets`."`` (if I'm understanding this correctly)

dvc/command/stage.py
skshetry: Adding metric and plot is one more word for each of the outputs. Would be better to find a single word to explain all these kinds of output.

dvc/command/stage.py
```diff
@@ -194,3 +292,47 @@ def add_parser(subparsers, parent_parser):
...
+    )
+    stage_list_parser.add_argument(
+        "--fail",
+        action="store_true",
+        default=False,
+        help="Fail immediately if there's an error.",
```
jorgeorpinel, line 323: Doesn't error imply failing already?
Not sure how to try this locally, what kind of error? Thanks

dvc/command/stage.py
skshetry: Remember that user will have the output path that is already quite lengthy. I am trying to keep it within a single line.

dvc/command/stage.py
```diff
@@ -194,3 +292,47 @@ def add_parser(subparsers, parent_parser):
...
+        "--recursive",
+        action="store_true",
+        default=False,
+        help="List all stages inside the specified directory.",
+    )
+    stage_list_parser.add_argument(
+        "--names-only",
+        action="store_true",
+        default=False,
+        help="List only the name of the stages.",
```
jorgeorpinel, line 336: Minor: `"List only stage names."`

jorgeorpinel (commented):
`stage list -h` string suggestions and some Qs:

dvc/command/stage.py
skshetry: Syntax Errors in yaml

dvc/command/stage.py
```diff
@@ -194,3 +292,47 @@ def add_parser(subparsers, parent_parser):
...
+        help=STAGE_LIST_HELP,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+    stage_list_parser.add_argument(
+        "targets",
+        nargs="*",
+        default=["dvc.yaml"],
+        help=(
+            "Show stages from a dvc.yaml/.dvc file or a directory. "
+            "'dvc.yaml' by default"
```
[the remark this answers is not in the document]
skshetry: > And, so this doesn't accept stage names?

It does, but I removed it to keep it short. 

> "Limit command scope to the stages in these .dvc or dvc.yaml files."

I don't think _limit_ applies here, it's more like changing the scope completely. Though I don't like the `repro`'s either.

dvc/command/stage.py
jorgeorpinel: Oops sorry super long reply. I know this is merged so OK. But for the record, keeping "produces" and "generates" seems to me  like the worst alternative. It's unclear why the former refers to regular outputs and the latter to dependencies.

jorgeorpinel: I mean metrics/plots

jorgeorpinel: \* super late reply 😅

dvc/command/stage.py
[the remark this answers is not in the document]
jorgeorpinel: "Limit" applies as much as in all the existing `targets` arg descriptions I think. It's usually dvc.yaml by default, and the base use case of this is to narrow it down.

dvc/command/stage.py
jorgeorpinel: OK. So my point was that "fails if there's an error" doesn't explain much. Maybe we can update these messages in some upcoming PR?

skshetry: I have opened https://github.com/iterative/dvc/issues/5390 regarding discussion of UI messages. Will update the flags as that will most likely affect it.

dvc/command/stage.py
```diff
@@ -1,11 +1,109 @@
...
+        if description
+        else ""
+        for stage in stages
+    }
+
+
+class CmdStageList(CmdBase):
+    def _get_stages(self) -> Iterable["Stage"]:
+        if self.args.all:
+            stages: List["Stage"] = self.repo.stages  # type: ignore
```
jorgeorpinel, line 70: One last post-merge comment here. Should it be `dvc stages add/list`? In plural like `params/metrics/plots/experiments` (Only `remote` is in singular, which is OK for Git familiarity I think)

skshetry: Good point. Subcommand should usually be plural, which we should better discuss on #5392 as a guideline.

dvc/command/stage.py
jorgeorpinel: Yeah I found that issue after checking this. Cool, thanks!

dvc/command/stage.py
skshetry: I find "regardless of `targets`" to be redundant due to the "all of the stages" there.

dvc/command/stage.py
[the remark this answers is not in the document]
skshetry: I don't prefer `Limit` here as the main use of `targets` here is to change the scope of `dvc.yaml` to be used, rather than narrowing it.

dvc/command/stage.py
skshetry: Tried to address this on #5499, please take a look.

dvc/command/stage.py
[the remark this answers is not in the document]
jorgeorpinel: OK then something like `.dvc file or dvc.yaml files to show stages from". Should def. be plural at least since it's "targets".

dvc/command/stage.py
jorgeorpinel: It's good to be redundant sometimes. We need to be extra clear. Users are not machines.

dvc/command/stage.py
jorgeorpinel: #5499 LGTM 👍

dvc/command/stage.py
jorgeorpinel: > Users are not machines. 

By this I mean that we often describe things correctly but people still get confused. All the commands an options are not obvious at all and as the developers we are biased because we already deeply understand them.

PR #5284, opened 2021-01-18
cli: add: add an --output argument

mrstegeman:
Allow users to add files from outside of the repo. This is a first
step toward achieving #4657.

References iterative/dvc.org#2101

### DIFF
--- a/dvc/command/add.py
+++ b/dvc/command/add.py
@@ -18,6 +18,12 @@ class CmdAdd(CmdBase):
             if len(self.args.targets) > 1 and self.args.file:
                 raise RecursiveAddingWhileUsingFilename()
 
+            if len(self.args.targets) != 1 and self.args.output:
+                logger.error(
+                    "cannot add more than one file when using --output"
+                )
+                return 1
+
             self.repo.add(
                 self.args.targets,
                 recursive=self.args.recursive,
@@ -26,6 +32,7 @@ class CmdAdd(CmdBase):
                 external=self.args.external,
                 glob=self.args.glob,
                 desc=self.args.desc,
+                output=self.args.output,
             )
 
         except DvcException:
@@ -82,6 +89,13 @@ def add_parser(subparsers, parent_parser):
             "User description of the data (optional). "
             "This doesn't affect any DVC operations."
         ),
+    )
+    parser.add_argument(
+        "-o",
+        "--output",
+        type=str,
+        metavar="<path>",
+        help="Output path for the file, if it lives outside the repo.",
     )
     parser.add_argument(
         "targets", nargs="+", help="Input files/directories to add.",
--- a/dvc/repo/add.py
+++ b/dvc/repo/add.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import shutil
 from typing import TYPE_CHECKING
 
 import colorama
@@ -7,6 +8,7 @@ import colorama
 from ..exceptions import (
     CacheLinkError,
     OutputDuplicationError,
+    OutputNotCreatedError,
     OverlappingOutputPathsError,
     RecursiveAddingWhileUsingFilename,
 )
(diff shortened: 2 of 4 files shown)


dvc/repo/add.py
```diff
@@ -38,6 +41,20 @@ def add(
...
     targets = ensure_list(targets)
+
+    if output:
+        try:
+            dname = os.path.dirname(output)
+
+            if dname and not os.path.isdir(dname):
+                os.makedirs(dname)
+
+            shutil.copy2(targets[0], output)
```
efiop, line 52: This will be really slow with big files/dirs. We could be able to add it without this uneeded copy.

tests/func/test_add.py
```diff
@@ -59,6 +60,40 @@ def test_add(tmp_dir, dvc):
...
+def test_add_output(tmp_dir, dvc):
+    (fd, fname) = tempfile.mkstemp()
+    os.write(fd, b"foo")
+    os.close(fd)
+
+    ret = main(["add", "--output", "foo", "bar", "baz"])
+    assert ret == 1
+
+    md5, _ = file_md5(fname)
+    stage = dvc.add(fname, output="foo")[0]
```
efiop, line 72: You are actually implementing straight-to-cache functionality https://github.com/iterative/dvc/issues/4520 for local files here :slightly_smiling_face: We have a WIP for a general straight-to-remote, where we also use `-o` in add, they are in no way contradictory though.

tests/func/test_add.py
```diff
@@ -59,6 +60,40 @@ def test_add(tmp_dir, dvc):
...
+        "outs": [
+            {
+                "md5": "acbd18db4cc2f85cedef654fccc4a4d8",
+                "path": "foo",
+                "size": 3,
+            }
+        ],
+    }
+
+    os.unlink(fname)
```
efiop, line 94: maybe just use tmp_path fixture or smth like that to avoid doing this manually and using file descriptors?

dvc/repo/add.py
efiop: Also, you are checking that there is only one target in CLI, but not in API (they could be used separately), so `targets[0]` needs some reassurance.

dvc/command/add.py
```diff
@@ -83,6 +90,13 @@ def add_parser(subparsers, parent_parser):
             "This doesn't affect any DVC operations."
         ),
     )
+    parser.add_argument(
+        "-o",
+        "--output",
+        type=str,
+        metavar="<path>",
+        help="Output path for the file, if it lives outside the repo.",
+    )
```
jorgeorpinel, line 99: This will clash with `--out` being introduced in #5198

efiop: It should be `--out`, but it won't clash otherwise. They are actually the same thing :)

jorgeorpinel: p.s. looking at this to try to understand iterative/dvc.org/pull/2101

Can someone explain what this does with an example perhaps? 

It's also not obvious to me how this is a step towards #4657 which doesn't seem to involve external data (but this does).

Thanks!

jorgeorpinel: I just noticed your answer @efiop but my questions remain because

> They are actually the same thing

`--out` in #5198 is not stand-alone.

jorgeorpinel: BTW should this maybe be a stand-alone, general option that `add` can always use (whether the target is external or not)? As disucssed in https://github.com/iterative/dvc/pull/5198#pullrequestreview-569646959 🙂

efiop:
@mrstegeman We've introduced `--out` flag as a part of straight-to-remote functionality https://github.com/iterative/dvc/issues/4520 , take a look at the upstream, maybe that's enough to get you going with #4657 We are currently working on straight-to-cache that avoids copying the artifact.

PR #5285, opened 2021-01-19
push: Optimize pushing behavior for big datasets

isidentical:
This patch basically rewrites the pushing logic's multitasking system to work gradually instead of pre-allocating all tasks. Some results (3x10k (3 different targets being pushed with 10k files in them each), 50k, 250k) (blue is the master) (these graphs are constructed from the data taken in every 4 seconds and compares the RSS)
image
image
image

### DIFF
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -1,10 +1,10 @@
 import errno
 import hashlib
+import itertools
 import logging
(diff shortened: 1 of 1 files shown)


pmrowla (approved):
LGTM, great optimization!

Should note that this is related to https://github.com/iterative/dvc/issues/4139

PR #5286, opened 2021-01-19
pygit/dulwich: implement stash functionality

pmrowla:

Related to #2215

### DIFF
--- a/dvc/scm/git/backend/pygit2.py
+++ b/dvc/scm/git/backend/pygit2.py
@@ -334,6 +334,8 @@ class Pygit2Backend(BaseGitBackend):  # pylint:disable=abstract-method
         return str(oid), False
 
     def _stash_apply(self, rev: str):
+        from pygit2 import GitError
+
         from dvc.scm.git import Stash
 
         # libgit2 stash apply only accepts refs/stash items by index. If rev is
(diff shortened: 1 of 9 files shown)


dvc/scm/git/backend/pygit2.py
```diff
@@ -193,15 +193,18 @@ def set_ref(
         message: Optional[str] = None,
         symbolic: Optional[bool] = False,
     ):
+        if message:
+            # pygit2 reflog message handling is bugged in 1.4.0
+            # see: https://github.com/libgit2/pygit2/issues/1061
```
pmrowla, line 198: upstream PR to fix this has been merged but unsure when it will actually be available i na release.

https://github.com/libgit2/pygit2/pull/1062

efiop: @pmrowla Let's just ask :)

pmrowla:
Still WIP pending https://github.com/dulwich/dulwich/pull/841

pmrowla:
With the current changes in this PR (including the upstream dulwich PR) and the changes that were merged into master today, experiments functionality should be completely independent of git-python. With the git-python backend disabled, `tests/func/experiments/` passes except for a couple specific tests that use CLI git directly to validate experiments behavior.

pmrowla:
There is some kind of pygit2 related race condition that only shows up on linux for python < 3.8, where the git index is not written/flushed properly on `stash.push/pop` that I'm still debugging

pmrowla:
As far as I can tell, there is no actual issue with our stash implementation, but for some reason, on linux in our pytest environment, pygit2/libgit2's stash_save is flaky. Using either `dvc.scm.git.stash` or pygit2 directly, for some reason libgit2 sometimes generates empty stash commits. Even though libgit see's that a file has changed and should be stashed, it stores an empty delta in the actual commit:

```
2021-03-08 08:16:39,847 DEBUG: Stashing changes in 'refs/stash'
status before stash
{'file': 256, 'untracked': 128}
stashed diff:
None
```
Before libgit2's `git_stash_save()`, it sees that `file` is modified (256 is `GIT_STATUS_WT_MODIFIED`), but the stash commit contains an empty diff. Note that libgit2 is trying to stash something, if `git_stash_save()` sees an unmodified working tree it will error out with a "nothing to do" error (which would be raised as a pygit exception).

So for whatever reason, in this flaky case, libgit2 see's that it should stash changes from `file`, but ends up stashing an empty diff instead of reading the current contents of `file`.

This is unreproducible outside of our test environment (identical tests in pygit2's unit test environment always pass), and all of the functional tests which use pygit2's stashing also pass. Will mark this test as flaky, and revisit it in the future if/when needed.

efiop (approved):
Great research! :rocket:

Issue #5266, opened 2021-01-13
dvc exp show: table display problem


patzy:
# Bug Report

## Description

I have a (toy) project with two pipeline folder (`pip1` and pip2`), each one has its own `params.yaml` and `metrics.json` both containing the same fields with différent values.

Parameters:
```
$ dvc params diff --all
Path              Param       Old    New                              
pip1/params.yaml  some_param  42     42
pip2/params.yaml  some_param  33     33
```
Metrics:
```
$ dvc metrics diff --all
Path               Metric    Old    New    Change                     
pip1/metrics.json  accuracy  —      0.42   —
pip2/metrics.json  accuracy  —      0.33   —
```
Experiments table:
```
$ dvc exp show --no-pager
```
![image](https://user-images.githubusercontent.com/151000/104496463-53600d00-55d9-11eb-8dba-67ef8e06b75b.png)

(I included a screenshot to avoir copy-paste formatting errors).

Project layout:
```
$ tree
.
├── README.md
├── data
│   ├── data.txt
│   ├── dvc.yaml
│   ├── data_count.txt
│   ├── dvc.lock
│   ├── dvc.yaml~
│   └── data_copy.txt
├── pip2
│   ├── params.yaml
│   ├── count_out.txt
│   ├── dvc.yaml
│   ├── dvc.lock
│   ├── dvc.yaml~
│   └── metrics.json
└── pip1
    ├── params.yaml
    ├── dvc.lock
    ├── dvc.yaml~
    ├── metrics.json
    ├── data_out.txt
    └── dvc.yaml
```

### Reproduce
1. create two pipeline folders and `dvc.yaml` files
2. create one `params.yaml` file for each folder containing the same variable (with different values)
3. create one `metrics.json` file for each folder containing the same fields (with different values)
4. Make sure `dvc.yaml` files each refer to their proper `params.yaml` and `metrics.json` files
5. run `dvc exp show`

### Expected

Proper display of the informations in the table (not sure how to display different pipelines)

### Environment information

```console
$ dvc version
DVC version: 2.0.0a0+e5f91f 
---------------------------------
Platform: Python 3.9.1 on Linux-5.10.4-arch2-1-x86_64-with-glibc2.32
Supports: http, https, s3, ssh
Cache types: hardlink, symlink
Cache directory: ext4 on /dev/sdb1
Caches: local
Remotes: None
Workspace directory: ext4 on /dev/sdb1
Repo: dvc, git
```

[...]

PR #5297, opened 2021-01-19
metrics/plots/params: return ordered

### LINKED ISSUE #5009: dvc: metrics.show() and params.show() order
It looks like these go in somewhat random order. Is it possible to make them reflect the order in `dvc.yaml` file? Since Python 3.6 dicts preserve order so we can still use that as an output type.

pared:

Related to #5009 

Since viewer is going to rely on this behaviour I decided to write tests for it.

EDIT:
Introduced tests also verify that use case described [here](https://github.com/iterative/dvc/issues/4446#issuecomment-757373208) is avialable.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -138,3 +138,17 @@ def pytest_configure(config):
             enabled_remotes.discard(remote_name)
         if enabled:
             enabled_remotes.add(remote_name)
+
+
+@pytest.fixture
+def dummy_stage(tmp_dir, dvc):
+    def make(path="dvc.yaml", name="dummy_stage", **kwargs):
+        from dvc.stage import PipelineStage, create_stage
+
+        stage = create_stage(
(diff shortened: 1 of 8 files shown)


tests/conftest.py
```diff
@@ -138,3 +138,17 @@ def pytest_configure(config):
             enabled_remotes.discard(remote_name)
         if enabled:
             enabled_remotes.add(remote_name)
+
+
+@pytest.fixture
+def dummy_stage(tmp_dir, dvc):
```
pared, line 144: If the idea sticks, I can refactor tests using `create_stage`, there is quite few of them.

pmrowla (approved):
LGTM

One thing to note is that the original issue should remain open after this PR, as it still needs to be addressed in `dvc exp show`. There's another metrics/params column related `exp show` issue (#5266) and I should be able to finish both issues together next week.

PR #5304, opened 2021-01-21
remove lock-file stages when pipeline file changed manually

### LINKED ISSUE #5293: manually deleted stages not removed from lock file
# Bug Report

## Description

Stages removed from a pipeline by editing `dvc.yaml` are not removed from `dvc.lock` after `dvc repro`.

### Reproduce

1. create a pipeline with two stages (stage_a, stage_b)
2. `dvc repro`
3. `dvc.lock` refers to stage_a and stage_b
4. edit `dvc.yaml` and remove one stage (i.e. stage_b)
5. `dvc repro`
6. stage_b is still mentionned in `dvc.lock`

### Expected

Manually deleted stages should have their corresponding section removed from `dvc.lock`

### Environment information

**Output of `dvc version`:**

```console
$ dvc version
DVC version: 1.11.10 (pip)
Platform: Python 3.9.1 on Linux-5.10.4-arch2-1-x86_64-with-glibc2.32
Caches: local
Remotes: None
Repo: dvc, git
```

Discord context https://discord.com/channels/485586884165107732/563406153334128681/801147361837187102

karajan1001:
fix #5293 

1. [x] Add bad case to tests
2. [ ] Solve these bad case.

### DIFF
--- a/tests/func/test_repro.py
+++ b/tests/func/test_repro.py
@@ -7,7 +7,7 @@ from pathlib import Path
 import pytest
 from mock import patch
 
-from dvc.dvcfile import DVC_FILE, Dvcfile
+from dvc.dvcfile import DVC_FILE, PIPELINE_FILE, PIPELINE_LOCK, Dvcfile
 from dvc.exceptions import (
     CyclicGraphError,
     ReproductionError,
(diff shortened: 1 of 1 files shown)


karajan1001:
@efiop 
Excuse me, having some trouble here.
Currently, we modify the lock-file after every stage reproducing. This is reasonable because the following stages in the pipeline may fail. So all of our modifications to the lock-file are at stage level. 

https://github.com/iterative/dvc/blob/02d9684c339e80f7f50a0a278720273b97d91598/dvc/repo/reproduce.py#L52-L56

Here all of our modifications should focus on the current stage, modifying other stages here is not a good design.

We lack a place where we manage the lock-file at the pipeline or higher level now.

skshetry:
This should be possible with the current abstraction as well, but I am not quite sure if we should create a separate command or prune those stale lock entries each time on `dvc repro` and `dvc exp run`.

The original reason for not doing this automatically was to try to reduce conflicts when the user is moving through different branches. Also, we didn't have the run-cache before, so moving to a different branch might mean that it had to be rerun. But that argument is invalid now.

Another thing to consider is how it will affect the `dvc exp`, considering how the ordering of those entries change. Is it really that big issue? 

If they are not big issues, we could implement an automatic pruning of those entries, otherwise, we'd need to add a separate command/flag for this.

efiop:
For the record: Agreed to close it for now.

PR #4721, opened 2020-10-14
lock: added troubleshooting doc link in error message (#4325)

harsh8398:
Resolves: https://github.com/iterative/dvc/issues/4325
PR merged in dvc.org: https://github.com/iterative/dvc.org/pull/1842

PR #5307, opened 2021-01-21
tree: Implement upload_fobj protocol to all trees

isidentical:
# All trees
## Previously implemented
- [x] s3
- [x] ssh
- [x] local
## Implemented in this PR
- [x] azure
- [x] gdrive
- [x] gs
- [x] hdfs
- [x] oss
- [x] webhdfs
- [x] webdav / webdavs?
- [x] http / https?

### DIFF
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -249,33 +249,16 @@ class CloudCache:
                     if not (
                         tree.isdvc(path_info, strict=False) and tree.fetch
                     ):
-                        self.tree.copy_fobj(fobj, cache_info)
+                        self.tree.upload_fobj(
+                            fobj, cache_info, no_progress_bar=True
+                        )
                 callback = kwargs.get("download_callback")
                 if callback:
                     callback(1)
 
         self.tree.state.save(cache_info, hash_info)
 
-    def _transfer_file_as_whole(self, from_tree, from_info):
-        from dvc.utils import tmp_fname
-
-        # When we can't use the chunked upload, we have to first download
-        # and then calculate the hash as if it were a local file and then
-        # upload it.
-        local_tree = self.repo.cache.local.tree
-        local_info = local_tree.path_info / tmp_fname()
-
-        from_tree.download(from_info, local_info)
-        hash_info = local_tree.get_file_hash(local_info)
-
-        self.tree.upload(
-            local_info,
-            self.tree.hash_to_path_info(hash_info.value),
-            name=from_info.name,
-        )
-        return hash_info
-
-    def _transfer_file_as_chunked(self, from_tree, from_info):
+    def _transfer_file(self, from_tree, from_info):
         from dvc.utils import tmp_fname
         from dvc.utils.stream import HashedStreamReader
 
@@ -296,14 +279,6 @@ class CloudCache:
 
         hash_info = stream_reader.hash_info
         self.move(tmp_info, self.tree.hash_to_path_info(hash_info.value))
-        return hash_info
-
-    def _transfer_file(self, from_tree, from_info):
-        try:
-            hash_info = self._transfer_file_as_chunked(from_tree, from_info)
-        except RemoteActionNotImplemented:
-            hash_info = self._transfer_file_as_whole(from_tree, from_info)
-
         return hash_info
 
     def _transfer_directory_contents(self, from_tree, from_info, jobs, pbar):
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -249,9 +249,7 @@ class CloudCache:
                     if not (
                         tree.isdvc(path_info, strict=False) and tree.fetch
                     ):
-                        self.tree.upload_fobj(
-                            fobj, cache_info, no_progress_bar=True
-                        )
+                        self.tree.upload_fobj(fobj, cache_info)
                 callback = kwargs.get("download_callback")
                 if callback:
                     callback(1)
--- a/dvc/tree/azure.py
+++ b/dvc/tree/azure.py
@@ -174,23 +174,20 @@ class AzureTree(BaseTree):
         blob_client = self.blob_service.get_blob_client(
             to_info.bucket, to_info.path
         )
-        blob_client.max_block_size = self.CHUNK_SIZE
-        blob_client.min_large_block_upload_threshold = self.CHUNK_SIZE
         blob_client.upload_blob(fobj, overwrite=True, max_concurrency=1)
 
     def _upload(
         self, from_file, to_info, name=None, no_progress_bar=False, **_kwargs
     ):
-
-        blob_client = self.blob_service.get_blob_client(
-            to_info.bucket, to_info.path
-        )
         total = os.path.getsize(from_file)
         with open(from_file, "rb") as fobj:
-            with Tqdm.wrapattr(
-                fobj, "read", desc=name, total=total, disable=no_progress_bar
-            ) as wrapped:
-                blob_client.upload_blob(wrapped, overwrite=True)
+            self.upload_fobj(
+                fobj,
+                to_info,
+                desc=name,
+                total=total,
+                no_progress_bar=no_progress_bar,
+            )
 
     def _download(
         self, from_info, to_file, name=None, no_progress_bar=False, **_kwargs
--- a/dvc/tree/azure.py
+++ b/dvc/tree/azure.py
@@ -170,15 +170,13 @@ class AzureTree(BaseTree):
     def get_file_hash(self, path_info):
         return HashInfo(self.PARAM_CHECKSUM, self.get_etag(path_info))
 
-    def upload_fobj(self, fobj, to_info, no_progress_bar=False, **pbar_args):
+    def _upload_fobj(self, fobj, to_info):
         blob_client = self.blob_service.get_blob_client(
             to_info.bucket, to_info.path
         )
         blob_client.max_block_size = self.CHUNK_SIZE
         blob_client.min_large_block_upload_threshold = self.CHUNK_SIZE
-
-        with Tqdm.wrapattr(fobj, "read", bytes=True, **pbar_args) as wrapped:
-            blob_client.upload_blob(wrapped, overwrite=True, max_concurrency=1)
+        blob_client.upload_blob(fobj, overwrite=True, max_concurrency=1)
 
     def _upload(
         self, from_file, to_info, name=None, no_progress_bar=False, **_kwargs
--- a/dvc/tree/azure.py
+++ b/dvc/tree/azure.py
@@ -170,6 +170,16 @@ class AzureTree(BaseTree):
     def get_file_hash(self, path_info):
         return HashInfo(self.PARAM_CHECKSUM, self.get_etag(path_info))
 
+    def upload_fobj(self, fobj, to_info, no_progress_bar=False, **pbar_args):
+        blob_client = self.blob_service.get_blob_client(
+            to_info.bucket, to_info.path
+        )
+        blob_client.max_block_size = self.CHUNK_SIZE
+        blob_client.min_large_block_upload_threshold = self.CHUNK_SIZE
+
+        with Tqdm.wrapattr(fobj, "read", bytes=True, **pbar_args) as wrapped:
+            blob_client.upload_blob(wrapped, overwrite=True, max_concurrency=1)
+
     def _upload(
         self, from_file, to_info, name=None, no_progress_bar=False, **_kwargs
     ):
@@ -183,16 +193,6 @@ class AzureTree(BaseTree):
                 fobj, "read", desc=name, total=total, disable=no_progress_bar
             ) as wrapped:
                 blob_client.upload_blob(wrapped, overwrite=True)
-
-    def upload_fobj(self, fobj, to_info, no_progress_bar=False, **pbar_args):
-        blob_client = self.blob_service.get_blob_client(
-            to_info.bucket, to_info.path
-        )
-        blob_client.max_block_size = self.CHUNK_SIZE
-        blob_client.min_large_block_upload_threshold = self.CHUNK_SIZE
-
-        with Tqdm.wrapattr(fobj, "read", bytes=True, **pbar_args) as wrapped:
-            blob_client.upload_blob(wrapped, overwrite=True, max_concurrency=1)
 
     def _download(
         self, from_info, to_file, name=None, no_progress_bar=False, **_kwargs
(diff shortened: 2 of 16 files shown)


dvc/tree/gdrive.py
```diff
@@ -345,29 +345,12 @@ def _gdrive_shared_drive_id(self, item_id):
...
     @_gdrive_retry
-    def _gdrive_upload_file(
-        self,
-        parent_id,
-        title,
-        no_progress_bar=False,
-        from_file="",
-        progress_name="",
-    ):
+    def _gdrive_upload_fobj(self, fobj, parent_id, title):
```
isidentical, line 348: This is a bit unclear, internally pydrive2 uses `MediaIoBaseUpload`
https://github.com/iterative/PyDrive2/blob/ee368315132bd31bb643bf3d8af088cc512e39f3/pydrive2/files.py#L776-L778

the initializer of this wrapper says chunk size should be lower than 5MB
https://googleapis.github.io/google-api-python-client/docs/epy/googleapiclient.http.MediaIoBaseUpload-class.html#__init__ 

though the default chunk size is 100 MB (if I am not missing anything) (line #4721 https://googleapis.github.io/google-api-python-client/docs/epy/googleapiclient.http-pysrc.html#MediaIoBaseUpload.__init__

cc: @shcheklein

dvc/tree/webdav.py
```diff
@@ -192,6 +192,23 @@ def move(self, from_info, to_info, mode=None):
         # Webdav client move
         self._client.move(from_info.path, to_info.path)
 
+    def upload_fobj(self, fobj, to_info, no_progress_bar=False, **pbar_args):
+        # this is not exactly a chunked upload, since it doesn't quite work
+        # in the way. internally it uses requests, so in theory we should be
+        # able to pass a generator that yields chunks though it is not working
+        # for some reason.
```
isidentical, line 199: The comment explains it up. When I checked the source code, I noticed it just passes the `buff=` directly as the `data=` argument of `requests.<method>`, which can actually accept a generator that yields chunks though it didn't work for some reason (~stuck in the test for infinite time), so this doesn't use a chunked approach.

shcheklein: btw, if it's an easy fix we have a full control of PyDrive2, we maintain it

dvc/tree/gdrive.py
shcheklein: Yes, information is confusing, probably they forgot to update the comment, or it should be interpreted in some other way. I see for example, that `rclone` has 8MB - https://forum.rclone.org/t/google-drive-and-optimal-drive-chunk-size/1186/11 ... and people mention 256M when they deal with very large files. I think it would be a good option to have, and we might indeed consider making it smaller by default to reduce memory pressure.

dvc/tree/webdav.py
isidentical: this is for `webdav`

dvc/tree/gdrive.py
isidentical: we use `64 MiB` on all other providers, though I think there is no way (as of now) to pass the chunk_size into `MediaIoBaseUpload`. Maybe we should consider adding it as a feature to `pydrive2`?

dvc/tree/webdav.py
efiop: @isidentical Yeah, webdav library is pretty odd with `buff`, even on dowload_to it writes the whole file into the buffer, which seems to defeat the purpose of it. I ran into it too in https://github.com/iterative/dvc/pull/5324 . Looks like we will need to use iter_chunks directly and have some fobj-like object that will read from it on demand.

dvc/tree/oss.py
```diff
@@ -116,6 +116,21 @@ def remove(self, path_info):
         logger.debug(f"Removing oss://{path_info}")
         self.oss_service.delete_object(path_info.path)
 
+    def upload_fobj(self, fobj, to_info, no_progress_bar=False, **pbar_args):
+        with Tqdm.wrapattr(
+            fobj, "read", disable=no_progress_bar, bytes=True, **pbar_args
+        ) as wrapped:
```
efiop, line 122: Sucks how we have to repeat the pbar logic in every implementation. It is clear that in the end we will do what fsspec does and create an AbstractFile, so that pbar could be used once up the stack. Not quite sure it is worth to bother with right now...

dvc/tree/webdav.py
skshetry: > stuck in the test for infinite time

We use `WSGIServer` for the tests which does not support chunked encoding.

The chunking upload does work if you try it out on a real server. We had support for chunking upload before on `_upload` but we removed it, as the Webdav uses `PUT` to transfer, which does not have good support for chunking upload in older versions (see: https://github.com/iterative/dvc/issues/4796).

isidentical: Thanks for the info @skshetry

dvc/tree/oss.py
isidentical: Should I hoist the upload_fobj into the BaseTree to manage the progress bar, or just leave it (just like upload/download)?

efiop: @isidentical If you can extract it like that with a reasonable amount of effort it would indeed be amazing! If not - we can leave it as is.

dvc/tree/webdav.py
isidentical: I have adjusted my comment with this information

dvc/tree/local.py
```diff
@@ -242,6 +243,8 @@ def get_file_hash(self, path_info):
     def getsize(path_info):
         return os.path.getsize(path_info)
 
+    _upload_fobj = partialmethod(copy_fobj, chunk_size=BaseTree.CHUNK_SIZE)
```
efiop, line 246: Btw, we have `copy_fobj` method here and in BaseTree, do we still need it or is it the same as upload_fobj?

isidentical: `copy_fobj` is basically `uplad_fobj` *without* progress bar.

isidentical: we could change all occurences to use `upload_fobj(no_progress_bar=True)`

isidentical: + and also without chunking

efiop: @isidentical It not having a progress bar is probably a bug :)

dvc/cache/base.py
```diff
@@ -249,33 +249,14 @@ def _save_file(self, path_info, tree, hash_info, save_link=True, **kwargs):
...
-        hash_info = local_tree.get_file_hash(local_info)
-
-        self.tree.upload(
-            local_info,
-            self.tree.hash_to_path_info(hash_info.value),
-            name=from_info.name,
-        )
-        return hash_info
-
-    def _transfer_file_as_chunked(self, from_tree, from_info):
```
efiop, line 278: :fire: :fire: :fire:

dvc/tree/s3.py
```diff
@@ -375,15 +363,20 @@ def upload_fobj(self, fobj, to_info, no_progress_bar=False, **pbar_args):
...
+                disable=no_progress_bar, total=total, bytes=True, desc=name
             ) as pbar:
-                s3.meta.client.upload_fileobj(
-                    fobj,
-                    to_info.bucket,
-                    to_info.path,
-                    Config=config,
-                    Callback=pbar.update,
+                obj.upload_file(
+                    from_file, Callback=pbar.update, ExtraArgs=self.extra_args,
```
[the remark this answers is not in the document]
isidentical: Yeah, that makes sense!

dvc/tree/azure.py
```diff
@@ -170,6 +170,14 @@ def getsize(self, path_info):
     def get_file_hash(self, path_info):
         return HashInfo(self.PARAM_CHECKSUM, self.get_etag(path_info))
 
+    def _upload_fobj(self, fobj, to_info):
+        blob_client = self.blob_service.get_blob_client(
+            to_info.bucket, to_info.path
+        )
+        blob_client.max_block_size = self.CHUNK_SIZE
+        blob_client.min_large_block_upload_threshold = self.CHUNK_SIZE
```
efiop, line 178: Btw, this might result in timeout errors. Maybe let's just let it use default chunk size? IIRC s3 (and maybe azure, but in gs we do this ourselves, see dynamic_chunk_size) tries to adjust chunk size until it stops running into timeout errors.

efiop: Same question for the rest of remotes.

Also, if we use defaults here, I guess we could reuse `upload_fobj` in `_upload` down below.

efiop:
Whoa, something strange is going on with the tests. I don't think it is related to your PR.

efiop:
@isidentical Fixed the issue in https://github.com/iterative/dvc/pull/5333 . Please rebase.

dvc/tree/azure.py
isidentical: > Same question for the rest of remotes.

What do you suggest? Lowering it down to something like 8MB or just removing it from azure/s3?

efiop: @isidentical I think we can just not use `CHUNK_SIZE` in s3 (possibly others) to let them figure out the best chunk size themselves. Unless I'm missing something.

isidentical: Alright, I'll strip the `CHUNK_SIZE` from all providers tomorrow (in order to first test under heavy conditions to see whether we would experience memory issues or not)

dvc/tree/azure.py
```diff
@@ -170,19 +170,24 @@ def getsize(self, path_info):
...
 
-    def _upload(
-        self, from_file, to_info, name=None, no_progress_bar=False, **_kwargs
-    ):
-
+    def _upload_fobj(self, fobj, to_info):
         blob_client = self.blob_service.get_blob_client(
             to_info.bucket, to_info.path
         )
+        blob_client.upload_blob(fobj, overwrite=True, max_concurrency=1)
```
efiop, line 177: Is `max_concurency=1` needed? If so, does it affect the upload speed? We didn't set it to 1 earlier.

dvc/tree/oss.py
```diff
@@ -116,6 +117,18 @@ def remove(self, path_info):
...
+        cursor = 0
+        while True:
+            chunk = fobj.read(self.CHUNK_SIZE)
+            if not chunk:
+                break
+
+            result = self.oss_service.append_object(
+                to_info.path, cursor, chunk
+            )
+            cursor = result.next_position
```
efiop, line 130: The chunk size might once again be too much for oss to handle :( Can we just use `put_object(fobj)`?

dvc/tree/webhdfs.py
```diff
@@ -100,7 +100,7 @@ def open(self, path_info, mode="r", encoding=None, **kwargs):
         with self.hdfs_client.read(
             path_info.path, encoding=encoding
         ) as reader:
-            yield reader.read()
+            yield reader
```
efiop, line 103: Just making sure this is not an accidental change.

isidentical: `reader.read()` will return string/bytes, `open()` should return an IO interface that we can call `.read()` on demand

efiop: ah, strange that it didn't fail in the tests before. Probably I've missed something. Which test started failing now, btw?

isidentical: > Which test started failing now, btw?

https://github.com/iterative/dvc/blob/1a43bfc29d49050b3e7f83d204d9105f9d615503/tests/func/test_tree.py#L326-L327

dvc/tree/s3.py
```diff
@@ -353,6 +353,13 @@ def get_file_hash(self, path_info):
                 size=obj.content_length,
             )
 
+    def _upload_fobj(self, fobj, to_info):
+        from boto3.s3.transfer import TransferConfig
+
+        config = TransferConfig(max_concurrency=1, use_threads=False)
```
efiop, line 359: same question as for azure

dvc/tree/webhdfs.py
efiop: @isidentical Thank you! :pray:

dvc/tree/s3.py
efiop: @isidentical And one more question: when we don't limit the concurency here, are there any specific requirements for the fobj that it is uploading from? E.g. if there are multiple threads, do they just `read()` sequentially or do they try to do some seek() magic?

isidentical: > E.g. if there are multiple threads, do they just read() sequentially or do they try to do some seek() magic?

The interface that `to_remote` passes to them doesn't implement the full file-like protocol (no seek etc), so in theory they are not able to do that.

efiop: So you've tried it and it didn't break, right?

dvc/tree/oss.py
```diff
@@ -38,6 +38,7 @@ class OSSTree(BaseTree):  # pylint:disable=abstract-method
     PARAM_CHECKSUM = "etag"
     COPY_POLL_SECONDS = 5
     LIST_OBJECT_PAGE_SIZE = 100
+    CHUNK_SIZE = 64 * 1024 * 1024
```
efiop, line 41: I guess no longer needed?
```suggestion
```

dvc/tree/base.py
```diff
@@ -59,6 +59,8 @@ class BaseTree:
     TRAVERSE_PREFIX_LEN = 3
     TRAVERSE_THRESHOLD_SIZE = 500000
     CAN_TRAVERSE = True
+
+    # Needed for some providers, and http open()
     CHUNK_SIZE = 64 * 1024 * 1024  # 64 MiB
```
efiop, line 64: Looks like it is now only used in http, so I guess no need to have it in the base class.
```suggestion
```

dvc/tree/http.py
```diff
@@ -169,6 +172,18 @@ def get_file_hash(self, path_info):
 
         return HashInfo(self.PARAM_CHECKSUM, etag)
 
+    def _upload_fobj(self, fobj, to_info):
+        def chunks(fobj):
+            while True:
+                chunk = fobj.read(self.CHUNK_SIZE)
```
efiop, line 178: There is a potential for timeout errors with 64mb chunk size, but if it was that size before we can keep it for now.

dvc/tree/base.py
isidentical: It's used in the `open()` call of `_transfer_file()`

efiop (approved):
Great stuff! :fire:

Issue #5255, opened 2021-01-12
add: Fails to add file to shared cache on windows drive mount, regression from 1.7.2


gamis:
# Bug Report

## Description

We have a shared dvc cache on a ZFS file server shared via Samba. On Windows machines, it is mounted as a drive letter. At one point, we could `dvc add` a file from a windows machine, and dvc would move the file to the cache and replace the local file with a symlink, as expected.

Somewhere between v1.7.2 and v1.11.10, it stopped working. Specifically, this is what appears to happen:

1. The file hash is calculated as expected.
2. The file bytes are copied to the cache under the correct name, but with the random suffix, denoting that the copy is in progress.
3. It never seems to be able to rename the cache file to it's final file name without the random suffix. No error is emitted, even with `-v`.
4. No .dvc file is created.
5. It errors out at a point where it's probably trying to replace the local file with a symlink. Output below.

Note that this happens even if I change the cache.type to `copy`, even if I run from an Admin-elevated cmd. 

Note also that the same operation runs just fine using a Ubuntu machine connecting to the same cache. 

### Reproduce

```console
mkdir D:\dvcregression
mkdir Z:\dvcregression_cache
cd D:\dvcregression
git init
dvc init
dvc config cache.dir Z:\dvcregression_cache
# copy in file TC3-096.oct to D:\dvcregression
dvc add -v TC3-096.oct
```

And you get the error pasted below.

```console
2021-01-12 15:44:59,435 DEBUG: Check for update is enabled.
2021-01-12 15:44:59,514 DEBUG: Trying to spawn '['daemon', '-q', 'updater']'
2021-01-12 15:45:03,013 DEBUG: Spawned '['daemon', '-q', 'updater']'
2021-01-12 15:45:03,420 DEBUG: Adding 'TC3-096.oct' to '.gitignore'.
2021-01-12 15:45:03,451 DEBUG: Path 'D:\dvcregression\TC3-096.oct' inode '5598938201872874496'
2021-01-12 15:45:03,451 DEBUG: fetched: []
2021-01-12 15:45:04,107 DEBUG: Path 'TC3-096.oct' inode '5598938201872874496'
2021-01-12 15:45:04,107 DEBUG: fetched: []
2021-01-12 15:45:04,107 DEBUG: {'TC3-096.oct': 'modified'}
2021-01-12 15:45:04,107 DEBUG: Path 'D:\dvcregression\TC3-096.oct' inode '5598938201872874496'
[...]

PR #5314, opened 2021-01-22
dvc: refactor cache saving

efiop:
Moving cache-related modes from the tree to the cache itself. On our way to unify transfer/push/pull/save functionality into one, so that we could use them in the 3.0 cache format.

Pre-requisite for #5255 

### DIFF
--- a/dvc/cache/local.py
+++ b/dvc/cache/local.py
@@ -9,7 +9,7 @@ from dvc.path_info import PathInfo
 from dvc.progress import Tqdm
 
 from ..utils import relpath
-from ..utils.fs import copyfile, remove, walk_files
+from ..utils.fs import copyfile, remove, umask, walk_files
 from .base import CloudCache
 
 logger = logging.getLogger(__name__)
@@ -24,6 +24,14 @@ class LocalCache(CloudCache):
         super().__init__(tree)
         self.cache_dir = tree.config.get("url")
 
+        shared = tree.config.get("shared")
+        if shared:
+            self._file_mode = 0o664
+            self._dir_mode = 0o2775
+        else:
+            self._file_mode = 0o666 & ~umask
+            self._dir_mode = 0o777 & ~umask
+
     @property
     def cache_dir(self):
         return self.tree.path_info.fspath if self.tree.path_info else None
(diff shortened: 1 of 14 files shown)


dvc/cache/local.py
```diff
@@ -24,6 +24,14 @@ def __init__(self, tree):
         super().__init__(tree)
         self.cache_dir = tree.config.get("url")
 
+        shared = tree.config.get("shared")
+        if shared:
+            self._file_mode = 0o664
+            self._dir_mode = 0o2775
```
efiop, line 30: Same as git, using S_ISGID sticky bit, this is a more proper way to handle shared cache scenario.

dvc/cache/local.py
```diff
@@ -24,6 +24,14 @@ def __init__(self, tree):
...
+        else:
+            self._file_mode = 0o666 & ~umask
+            self._dir_mode = 0o777 & ~umask
```
efiop, line 33: Previously we've used hardcoded modes, which was not right. Now properly using umask, same as git.

dvc/remote/base.py
```diff
@@ -463,14 +450,27 @@ def create_taskset(amount):
...
+        if self.tree.scheme == "local":
+            with self.tree.state:
+                for checksum in named_cache.scheme_keys("local"):
+                    cache_file = self.tree.hash_to_path_info(checksum)
+                    if self.tree.exists(cache_file):
+                        hash_info = HashInfo(
+                            self.tree.PARAM_CHECKSUM, checksum
+                        )
+                        self.tree.state.save(cache_file, hash_info)
+                        self.cache.protect(cache_file)
```
efiop, line 470: Same as "trust" logic in `pull`. This really emphasizes that push and pull are the same thing: transering objects from one cache to another, so the receiving cache should be able to save objects as it pleases. This and the "trust" logic below will move to cache.save() in the future.

dvc/stage/cache.py
```diff
@@ -170,12 +166,12 @@ def save(self, stage):
...
 
         path = PathInfo(self._get_cache_path(cache_key, cache_value))
-        self.tree.makedirs(path.parent)
+        self.repo.cache.local.makedirs(path.parent)
         tmp = tempfile.NamedTemporaryFile(delete=False, dir=path.parent).name
         assert os.path.exists(path.parent)
         assert os.path.isdir(path.parent)
         dump_yaml(tmp, cache)
-        self.tree.move(PathInfo(tmp), path)
+        self.repo.cache.local.move(PathInfo(tmp), path)
```
efiop, line 174: Logic related to saving stage cache object, should really be in the cache itself.

PR #5320, opened 2021-01-24
[WIP] pull: Improve logging when files are missing from remote and cache

### LINKED ISSUE #5248: pull: warning on failed objects is too low level
## DVC Version

```
DVC version: 1.11.0+322fc6
Platform: Python 3.8.6 on macOS-10.15.6-x86_64-i386-64bit
Supports: All remotes
Cache directory: apfs on /dev/disk1s1
Caches: local
Remotes: dropbox
Workspace directory: apfs on /dev/disk1s1
Repo: dvc, git
```

## Reproduce

Clean remote storage, pull a file:

`dvc pull images.tar.dvc`

## Output

```
name: images.tar, md5: 721d011095cf62dd7cd5be2ea2c269b7
WARNING: Cache 'HashInfo(name='md5', value='721d011095cf62dd7cd5be2ea2c269b7', dir_info=None, size=192139264, nfiles=None)' not found. File 'images.tar' won't be created.
1 file failed
ERROR: failed to pull data from the cloud - Checkout failed for following targets:
images.tar
Is your cache up to date?
```

## Problems

- ~~`HashInfo(name='md5', value='721d011095cf62dd7cd5be2ea2c269b7', dir_info=None, size=192139264, nfiles=None)` is not acceptable for UI~~ https://github.com/iterative/dvc/pull/5319
- WARNING prefix looks ugly
- warnings are probably not needed (especially if it's causing ERROR at the end - ERROR in this case should be better)

gth828r:
pull: Improve logging when files are missing

 * When the file is missing on the remote, report the issue at the
   debug log level. Let the exception be the message that gets
   reported back to the user by default.
 * Improve the logging when a file is missing from the cache
   * Report the issue at the debug log level. Let the exception be the
     message that gets reported back to the user by default.
   * Change the log message to be more human readable by a user.

Fixes #5248

### DIFF
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -700,12 +700,13 @@ class CloudCache:
         elif self.changed_cache(
             hash_info, path_info=path_info, filter_info=filter_info
         ):
-            if not quiet:
-                logger.warning(
-                    "Cache '%s' not found. File '%s' won't be created.",
-                    hash_info,
-                    path_info,
-                )
+            msg = (
+                "Cache entry with hash '%s' not found. File '%s' won't be "
+                "created."
+            )
+            logger.debug(
+                msg, hash_info.value, path_info,
+            )
             self.safe_remove(path_info, force=force)
             failed = path_info
 
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -519,4 +519,4 @@ class Remote:
                 "Some of the cache files do not exist neither locally "
                 "nor on remote. Missing cache files:\n{}".format(missing_desc)
             )
-            logger.warning(msg)
+            logger.debug(msg)


gth828r:
What this change effectively does is takes the two warnings that were being printed by default and only makes them visible in verbose mode. Alternatively, we could also set the log level to info to one of both of the messages, which would have the effect of printing out the messages without any type of log level prefix. From the sound of the issue, the goal was to only print out the error level message on runs without the verbose flag set.

dvc/cache/base.py
```diff
@@ -700,12 +700,13 @@ def checkout(
...
-                    "Cache '%s' not found. File '%s' won't be created.",
-                    hash_info,
-                    path_info,
-                )
+            msg = (
+                "Cache entry with hash '%s' not found. File '%s' won't be "
+                "created."
+            )
+            logger.debug(
+                msg, hash_info.value, path_info,
```
pmrowla, line 708:
```suggestion
                msg, hash_info, path_info,
```

`HashInfo` has a `__str__` method now, and we don't want to drop the hash type from the message either, since it may not always be an `md5` depending on the remote/cache type.

pmrowla (changes requested):
I'm not sure this change is enough, as the error message for missing files is still misleading. If a file which needs to be fetched does not exist in the remote, what happens now is that we only display the "checkout failed" message:

```
$ dvc pull                                                                                                    ⏎
1 file failed
ERROR: failed to pull data from the cloud - Checkout failed for following targets:
data/data.xml
Is your cache up to date?
```

In this case the real cause of the problem is not that the (local cache) checkout step failed, it's that the (remote) fetch step failed. It seems to me like what is really needed here for `dvc fetch/pull` would be a "is your remote up to date" hint (because someone else forgot to `dvc push`) rather than "is your (local) cache up to date" (because you forgot to `dvc fetch/pull`) hint.

Should note that the troubleshooting link does correctly address the "someone else probably needs to push" issue for `dvc pull` and is still displayed when local checkout fails but the file is actually available to be fetched/pulled.

dvc/cache/base.py
gth828r: > In this case the real cause of the problem is not that the (local cache) checkout step failed, it's that the (remote) fetch step failed. It seems to me like what is really needed here for dvc fetch/pull would be a "is your remote up to date" hint (because someone else forgot to dvc push) rather than "is your (local) cache up to date" (because you forgot to dvc fetch/pull) hint.

I think you are right about the error message. I had not thought deeply about what might be a common cause of this error case, but you are right that it is likely that someone else forgetting to push is likely a common cause.

> HashInfo has a __str__ method now, and we don't want to drop the hash type from the message either, since it may not always be an md5 depending on the remote/cache type.

I had mistakenly assumed that only one hash type was used, but if that is not the case then we should include the hash type. I will need update the message format string though, as the change I made to the message doesn't make sense with the raw hash_info put in where just the hash was.

With that said, I had intentionally moved away from passing the raw hash_info object because I felt that the claim in #5248 that the message was too low level still applied even if we only printed the hash_info str representation in verbose mode (via only logging at DEBUG level). If this representation is OK to print out in verbose mode, then I agree it is fine to just go back to passing the hash_info object. What do you think @pmrowla?

pmrowla: Keeping the hash type in the output should be preferred for the debug output because it is not always going to be an MD5 (it could also be an S3 or HTTP etag for example). The main complaint before was that printing the entire `repr()` for hash_info objects was too verbose, but it will be a more nicely formatted string in the form `md5: 12345...` now that https://github.com/iterative/dvc/pull/5319 is merged

gth828r: Ok, I am up to speed now! Thanks for pointing out the recent change.

Regardless, it sounds like more thought and additional changes are probably required to truly address the original issue, and I am not going to be able to get back to this for a bit. If it makes sense to avoid me sitting on this task without progress, I'm happy to close out this MR so that someone else can pick it up. Otherwise, I can pick it back up later (probably in a week or two).

pmrowla: I'll close this for now then, but I'm not sure if anyone else will pick up this issue in the next couple of weeks. If it hasn't been addressed by then, feel free to re-open your PR whenever you have the time for it.

PR #5324, opened 2021-01-25
RepoTree: get rid if implicit fetching

efiop:
When we collect cache for pull/fetch or run status -c, we implicitly try to fetch the dir cache entry from the remote. And that's the behaviour that is expected from RepoTree too. There have been multiple times where we've tried to use Repo.repo_tree just to realise that it tries to fetch automatically, which is often not needed. 

This PR is more of a POC that tries to get rid of `fetch` flag and the implicit fetching during walk() that it implies in favor of just fetching stuff explicitly where we really need it (repo has `fetch` method just for that already).

Need to check performance and also 2 `webdav` tests fail now because they don't have open() support, and Tree doesn't fetch things automatically, as expected by the api code.

Depends on https://github.com/iterative/dvc/pull/5307

### DIFF
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -246,12 +246,7 @@ class CloudCache:
         else:
             if self.changed_cache(hash_info):
                 with tree.open(path_info, mode="rb") as fobj:
-                    # if tree has fetch enabled, DVC out will be fetched on
-                    # open and we do not need to read/copy any data
-                    if not (
-                        tree.isdvc(path_info, strict=False) and tree.fetch
-                    ):
-                        self.tree.upload_fobj(fobj, cache_info)
+                    self.tree.upload_fobj(fobj, cache_info)
                 callback = kwargs.get("download_callback")
                 if callback:
                     callback(1)
@@ -444,7 +439,9 @@ class CloudCache:
             )
 
         if not hash_info:
-            hash_info = tree.get_hash(path_info, **kwargs)
+            kw = kwargs.copy()
+            kw.pop("download_callback", None)
+            hash_info = tree.get_hash(path_info, **kw)
             if not hash_info:
                 raise FileNotFoundError(
                     errno.ENOENT, os.strerror(errno.ENOENT), path_info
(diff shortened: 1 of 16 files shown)


dvc/repo/fetch.py
```diff
@@ -98,6 +99,16 @@ def cb(result):
...
             for path in files:
                 path_info = root / path
+                try:
+                    used = repo.used_cache(
+                        [os.fspath(path_info)],
+                        force=True,
+                        jobs=jobs,
+                        recursive=True,
+                    )
+                    cb(repo.cloud.pull(used, jobs))
```
efiop, line 109: Strictly speaking this is not really needed, as `cache.save()` will fetch everything anyway, but I've added this for now because logic around this (e.g. tests) expects specific exceptions and also because this might be faster than `save()`, as we didn't optimize it yet.

dvc/tree/dvc.py
```diff
@@ -149,17 +133,14 @@ def _fetch_dir(
...
-        # pull dir contents if needed
-        if self.fetch and out.changed_cache(filter_info=filter_info):
-            used_cache = out.get_used_cache(filter_info=filter_info)
-            downloaded = self.repo.cloud.pull(used_cache, **kwargs)
-            if download_callback:
-                download_callback(downloaded)
+        dir_cache = out.dir_cache
+        hash_info = out.cache.save_dir_info(dir_cache)
+        if hash_info != out.hash_info:
+            from dvc.output.base import CollectCacheError
```
efiop, line 139: A temporary workaround, we should be able to open cache dir remotely without fetching it.

dvc/cache/base.py
```diff
@@ -444,7 +439,9 @@ def save(self, path_info, tree, hash_info, save_link=True, **kwargs):
             )
 
         if not hash_info:
-            hash_info = tree.get_hash(path_info, **kwargs)
+            kw = kwargs.copy()
+            kw.pop("download_callback", None)
```
efiop, line 443: This is temporary, we'll start computing hash on-the-fly in the followup PRs.

Issue #5330, opened 2021-01-25
Support standard HTTP authentication protocol for basic authentication.


simonlsk:
Currently, it is possible to manually set authentication to `basic`, then set `user` and `password` as config parameters. This means saving a password in plain text in the `config` or `config.local` file, or alternatively use `ask_password` which is also sub-optimal. 

The standard HTTP protocol, which is also used by git, is the following:
1. Try to access URL (plain request)
2. Receive `401` with header `WWW-Authenticate: Basic realm=<realm>`
3. Locally fetch username and password - preferably store the password in the OS password store (i.e Keychain for MacOS)
4. Retry to access URL again with basic auth header.

This will make it possible to configure an HTTP remote with a single line
```dvc remote add origin https://dagshub.com/user/repo.dvc```
And directly be able to run
```dvc push -r origin```

guysmoilov:
I would add - this would also make it much easier to implement
`dvc push -r https://my-dvc-remote/` directly, 
without configuring anything, much like how `git clone` works.

efiop:
Thanks for the feature request, guys! We'll be happy to accept a PR, if you feel like contributing it :slightly_smiling_face:

PR #5331, opened 2021-01-25
[WIP] Support HTTP authentication prompt after 401

simonlsk:
Addresses issue #5330.

Affected components: 
HTTP tree

When HTTP request returns 401 with `www-authenticate`, the username is prompted if not existent, then the password is prompted if inexistent or fetched from OS secret manager.
If `ask_password` is set to `true` prompt for password anyway.

### DIFF
--- a/dvc/tree/http.py
+++ b/dvc/tree/http.py
@@ -5,7 +5,6 @@ import threading
 from funcy import cached_property, wrap_prop, wrap_with
 
 import dvc.prompt as prompt
-from dvc.config import merge
 from dvc.exceptions import DvcException, HTTPError
 from dvc.hash_info import HashInfo
 from dvc.path_info import HTTPURLInfo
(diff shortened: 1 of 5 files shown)


dvc/tree/http.py
```diff
@@ -56,18 +63,47 @@ def __init__(self, repo, config):
         self.ssl_verify = config.get("ssl_verify", True)
         self.method = config.get("method", "POST")
 
+    @memoize
+    def _get_user(self, host):
+        user = ask_username(host)
+        self.path_info.user = user
+        config_level = "local"
+        with self.repo.config.edit(config_level) as conf:
```
efiop, line 71: There have been a few questions about this before, but our remotes don't modify their own config. Some remotes like `gdrive`, save things to .dvc/tmp, but don't modify the config. Looks like we don't really need to do it, right? We can ask again or make the user set `user` in .dvc/config.local by himself. Or am I missing something?

efiop:
Oops, we had an issue with a recent pytest version. Fixed in upstream, please rebase.

dvc/tree/http.py
simonlsk: I see how it breaks the design a little bit. 
Asking again / asking the user to run a command is a bit unnatural UX (different than git for instance). Can you try to help me find a more appropriate way to save the username as a side effect, that will be satisfying design-wise?

efiop: @simonlsk Great point. Where does git store the username in such cases?

simonlsk: That's a good question, my git installation set my `credential.helper` as `osxkeychain` and I am only prompted once about any combination of `user, host, password`. I thought that you could only fetch a password once you know the `user, host` combination and that the username was being saved in  `.git/config`, but I was wrong. I just read [the code in git](https://github.com/git/git/blob/306ee63a703ad67c54ba1209dc11dd9ea500dc1f/contrib/credential/osxkeychain/git-credential-osxkeychain.c) and it looks like can it fetch entries without knowing the username.
I will try to workaround this, and let you know.

simonlsk:
There is a security issue using keyring for that purpose. Closing the PR for now.

PR #5341, opened 2021-01-27
list: don't clone local repos

efiop:
Pre-requisite for #3590 

### DIFF
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -59,12 +59,16 @@ def external_repo(
         url=url,
         scm=None if for_write else Git(root_dir),
         rev=None if for_write else rev,
-        subrepos=not for_write,
-        uninitialized=True,
         config=config,
         repo_factory=make_repo,
         **kwargs,
     )
+
+    if "subrepos" not in repo_kwargs:
+        repo_kwargs["subrepos"] = True
+
+    if "uninitialized" not in repo_kwargs:
+        repo_kwargs["uninitialized"] = True
 
     repo = Repo(**repo_kwargs)
 
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -202,6 +202,18 @@ class Repo:
     def __str__(self):
         return self.url or self.root_dir
 
+    @staticmethod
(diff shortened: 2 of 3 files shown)


dvc/external_repo.py
```diff
@@ -59,13 +59,17 @@ def make_repo(path, **_kwargs):
...
         repo_factory=make_repo,
         **kwargs,
     )
 
+    if "subrepos" not in repo_kwargs:
+        repo_kwargs["subrepos"] = True
+
+    if "uninitialized" not in repo_kwargs:
+        repo_kwargs["uninitialized"] = True
+
```
efiop, line 72: These should be explicit as in the ls.py below. Just keeping this as a temporary compatibility layer for other code.

dvc/repo/__init__.py
```diff
@@ -202,6 +202,18 @@ def __init__(
...
+    @staticmethod
+    @contextmanager
+    def open(url, *args, **kwargs):
+        if os.path.isdir(url):
+            yield Repo(url, *args, **kwargs)
+            return
+
+        from dvc.external_repo import external_repo
+
+        with external_repo(url, *args, **kwargs) as repo:
```
efiop, line 214: What this really is is a `Repo.clone` with a temporary dir. Having `Repo.open()` is very convenient when you just want to access the repo, no matter local or remote (in which case temp clone is totally fine).

dvc/repo/__init__.py
```diff
@@ -202,6 +202,18 @@ def __init__(
     def __str__(self):
         return self.url or self.root_dir
 
+    @staticmethod
+    @contextmanager
+    def open(url, *args, **kwargs):
+        if os.path.isdir(url):
```
skshetry, line 208: just an `exists` check would be enough here, I think.

efiop: Good point, might improve the local error.

efiop:
Ok, we'll need to revisit CLI semantics for the `list` itself (e.g. to make the docs/help make sense for `dvc list` and `dvc list path`). Merging so we could use open() in other places.

PR #5343, opened 2021-01-27
update: implement --to-remote flag

isidentical:

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here. (Awaiting iterative/dvc.org#2091)

### DIFF
--- a/dvc/stage/imports.py
+++ b/dvc/stage/imports.py
@@ -14,11 +14,11 @@ def update_import(stage, rev=None, to_remote=False, remote=None, jobs=None):
     stage.deps[0].update(rev=rev)
     frozen = stage.frozen
     stage.frozen = False
-    if to_remote:
-        return _update_import_on_remote(stage, remote, jobs)
-
     try:
-        stage.reproduce()
+        if to_remote:
+            _update_import_on_remote(stage, remote, jobs)
+        else:
+            stage.reproduce()
     finally:
         stage.frozen = frozen
 
--- a/dvc/stage/imports.py
+++ b/dvc/stage/imports.py
@@ -3,10 +3,20 @@ import logging
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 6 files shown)


dvc/stage/imports.py
```diff
@@ -3,10 +3,20 @@
...
+    stage.outs[0].hash_info = stage.repo.cloud.transfer(
+        url, jobs=jobs, remote=remote, command="update"
+    )
+
+
+def update_import(stage, rev=None, to_remote=False, remote=None, jobs=None):
     stage.deps[0].update(rev=rev)
     frozen = stage.frozen
     stage.frozen = False
+    if to_remote:
```
efiop, line 17: Looks like we are forgetting to set `frozen` back

dvc/stage/imports.py
```diff
@@ -3,10 +3,20 @@
 logger = logging.getLogger(__name__)
 
 
-def update_import(stage, rev=None):
+def _update_import_on_remote(stage, remote, jobs):
+    url = stage.deps[0].path_info.url
+    stage.outs[0].hash_info = stage.repo.cloud.transfer(
+        url, jobs=jobs, remote=remote, command="update"
+    )
```
efiop, line 10: Hm, seems like this is something we are doing over and over in add/import and now update. Makes me think this should really be some kind of flag/option for `stage.save()` and `out.save()` that would make it transfer stuff. We've had this discussion before and it wasn't clear back then, but maybe it is a good time. Or at least we could consider it and do it in the follow up.

isidentical: I also thought the same, and I think it would somehow make sense to have a function to do this in either stage or output.

efiop (approved):

PR #5344, opened 2021-01-27
tests: replace create_stage with fixture

pared:

related to #5297

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -141,13 +141,15 @@ def pytest_configure(config):
 
 
 @pytest.fixture
-def dummy_stage(tmp_dir, dvc):
-    def make(path="dvc.yaml", name="dummy_stage", **kwargs):
-        from dvc.stage import PipelineStage, create_stage
+def make_stage(tmp_dir, dvc):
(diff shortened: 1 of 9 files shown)


tests/conftest.py
```diff
@@ -141,13 +141,15 @@ def pytest_configure(config):
 
 
 @pytest.fixture
-def dummy_stage(tmp_dir, dvc):
-    def make(path="dvc.yaml", name="dummy_stage", **kwargs):
-        from dvc.stage import PipelineStage, create_stage
+def make_stage(tmp_dir, dvc):
```
skshetry, line 144: @pared, we might not need this fixture at all if we decide on `dvc.stage.add()` (just need to type a bit more though).
The only thing that needs to be decided is whether it fits into the current `dvc.stage` or should it be a different part of the abstraction (`dvc.index`?).

pared: makes sense, do we have a discussion for that somewhere?

pared:
Closing for now after discussion with @skshetry. Seems like soon we will be able to replace `create_stage` with some different concept. I will get back to this change then.

PR #5346, opened 2021-01-27
pull, fetch: optimize memory consuption by ~%35

isidentical:
(250k files, master=blue), It is safe to assume that 750mb~ is the total memory consumption up to the point of future object creation. After that (750-1.3 GB) is the real part that is being optimized away. Generalizes the same approach used to optimize `push` in my previous PR. 
image

### DIFF
--- a/dvc/remote/base.py
+++ b/dvc/remote/base.py
@@ -4,7 +4,7 @@ import itertools
 import logging
 from concurrent import futures
 from concurrent.futures import ThreadPoolExecutor
-from functools import wraps
+from functools import partial, wraps
 
 from dvc.exceptions import DownloadError, UploadError
 from dvc.hash_info import HashInfo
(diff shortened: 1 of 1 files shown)


isidentical:
I guess the last part of #4139

dvc/remote/base.py
```diff
@@ -339,63 +339,74 @@ def _process(
...
+                dir_contents,
+                missing_files,
+                jobs,
+                func,
+            )
 
         return total
 
-    def _upload_plans(
+    def _process_plans(
```
isidentical, line 354: I felt like `_process` was actually doing so much, so extracted the actual transfer functionality aside.

pmrowla (approved):
LGTM

Unrelated to this PR, but at some point we should probably revisit/refactor the top level push/pull/status calls to use typing, and also use things like named tuples. The way we pass around the different `NamedCache` instances and upload/download plan tuples is non-intuitive, and maybe it's just me, but I always have to re-read a big chunk of this module to remember what we are doing.

Issue #5347, opened 2021-01-27
abstract cli ui utilities, move away from `logger.info` for display


skshetry:
It's hard to describe, but we are trying to move away from `logging` for UI related prints.
To do so, we have to start consolidating the utilities in one place, decide default colours for the output, reuse certain components (table, etc.). 

Then slowly, start replacing the usages of logger internally for printing.

skshetry:
Closing, as we have moved away from `logger.info` in the CLI, and created some abstractions. Other stuff will be added as I go through the commands based on necessity. :)

PR #5348, opened 2021-01-27
introduce dvc.io module to consolidate io utilities

skshetry:
Related to https://github.com/iterative/dvc/issues/5347

### DIFF
--- a/dvc/ui/__init__.py
+++ b/dvc/ui/__init__.py
@@ -1,0 +1,152 @@
+import sys
+from collections import defaultdict
+from typing import Any, Dict, Iterable, Optional, TextIO
+
+from funcy import cached_property
+
+from dvc.progress import Tqdm
+from dvc.utils import colorize
+
+NEWLINE = "\n"
+SEP = " "
+
+
+class Formatter:
+    def __init__(self, theme: Dict = None, defaults: Dict = None) -> None:
+        theme = theme or {
+            "success": {"color": "green", "style": "bold"},
+            "warn": {"color": "yellow"},
+            "error": {"color": "red", "style": "bold"},
+        }
+        self.theme = defaultdict(lambda: defaults or {}, theme)
+
+    def format(self, message: str, style: str = None, **kwargs) -> str:
+        return colorize(message, **self.theme[style])
+
+
+class Console:
+    def __init__(
+        self,
+        formatter: Formatter = None,
+        output: TextIO = None,
+        error: TextIO = None,
+        disable: bool = False,
+    ) -> None:
+        self._input: TextIO = sys.stdin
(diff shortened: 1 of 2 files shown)


efiop:
Maybe `io` -> `ui`? `io` has lowlevel connotation, at least to me. `ui` is logically about printing, tables, etc.

pmrowla:
> Maybe `io` -> `ui`? `io` has lowlevel connotation, at least to me. `ui` is logically about printing, tables, etc.

Same for me, especially in the python sense, `io` as a module sounds more like something for directly manipulating streams (same as python stdlib `io`)

dvc/io/__init__.py
```diff
@@ -0,0 +1,152 @@
...
+            "warn": {"color": "yellow"},
+            "error": {"color": "red", "style": "bold"},
+        }
+        self.theme = defaultdict(lambda: defaults or {}, theme)
+
+    def format(self, message: str, style: str = None, **kwargs) -> str:
+        return colorize(message, **self.theme[style])
+
+
+class IO:
```
pmrowla, line 27: I'm still in favor of naming this something besides `IO`, whether it's `UI` or `Console` or something else entirely.

`IO` to me still makes it look like this is supposed to be related to `io.IOBase` or `io.TextIOBase` for reading/writing text streams. We do have a `write` method, but it isn't really an arbitrary stream write, it's a formatted wrapper for `print`.

dvc/io/__init__.py
```diff
@@ -0,0 +1,152 @@
...
+        self.disabled: bool = disable
+
+    def success(self, message: str) -> None:
+        self.write(message, style="success")
+
+    def error(self, message: str) -> None:
+        self.error_write(message, style="error")
+
+    def warn(self, message: str) -> None:
+        self.error_write(message, style="warn")
```
pmrowla, line 49: Is the idea here that we will eventually replace the use of `logging` with these methods?

Or are these just for console UI output messages which would be separate from internal logging calls?

skshetry: If we start printing in colors, they will come in handy.

skshetry:
@pmrowla @pared @efiop  @isidentical, guys, PTAL again. I changed the module name to `dvc.ui` and the class to `Console` (as I plan to add more `ui` related components to the module, though you'll probably use it as `ui.success` etc.).

skshetry:
I planned to add `rich_table` support in this same PR, but looks like it'll take some time to unify it, especially the styling (and/or adding markdown/csv/json formats into it). So, I'd like to merge this and work on `Table` on a different PR.

pared (approved):

efiop (approved):
Great start! :fire:

PR #5350, opened 2021-01-28
stage:add: use remainder as a command script

### LINKED ISSUE #5342: stage:add: decide on `-c` command vs remainder arg for user's script
On `dvc run`, the remainder positional argument is considered as a script. 
```console
$ dvc run -n stage_name -o bar -w data echo bar > bar
```

But, we have heard complaints from users that it's confusing, as trying to add certain flags to `dvc run` at the end is now assumed to be part of the user's script. 

So, on `stage:add`, we moved it under the `-c`/`--command` flag and now needs to be explicitly specified and quoted.
```console
$ dvc stage add stage_name -o bar -w data -c "echo bar > bar"
```

@dmpetrov later pointed out that it breaks the autocompletion for users. This discussion was without any DVC's autocompletion installed.

With DVC's autocompletion installed, the autocompletion anyway does not work as it is assumed as a plain string.

Anyway, we have to either choose between`-c` and the remainder arg in `run`.

cc @dberenbaum @shcheklein @jorgeorpinel

skshetry:
Fixes #5342 

https://github.com/iterative/dvc.org/issues/2134

### DIFF
--- a/dvc/command/stage.py
+++ b/dvc/command/stage.py
@@ -310,7 +310,9 @@ def add_parser(subparsers, parent_parser):
         help=STAGE_ADD_HELP,
         formatter_class=argparse.RawDescriptionHelpFormatter,
     )
-    stage_add_parser.add_argument("name", help="Name of the stage to add")
+    stage_add_parser.add_argument(
+        "-n", "--name", help="Name of the stage to add", required=True
+    )
     _add_common_args(stage_add_parser)
     stage_add_parser.set_defaults(func=CmdStageAdd)
 
--- a/dvc/command/stage.py
+++ b/dvc/command/stage.py
@@ -248,7 +248,7 @@ def _add_common_args(parser):
         metavar="<filename>",
     )
     parser.add_argument(
-        "-C",
+        "-c",
         "--checkpoints",
         action="append",
         default=[],
--- a/dvc/command/stage.py
+++ b/dvc/command/stage.py
@@ -105,10 +105,32 @@ class CmdStageList(CmdBase):
         return 0
 
 
+def parse_cmd(commands: List[str]) -> str:
+    """
+    We need to take into account two cases:
+
+    - ['python code.py foo bar']: Used mainly with dvc as a library
+    - ['echo', 'foo bar']: List of arguments received from the CLI
+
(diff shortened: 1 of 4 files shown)


pmrowla:
If we're going to drop `-c` for commands should we go back to `-c/--checkpoints` instead of `-C`

pmrowla (approved):

dvc/command/stage.py
```diff
@@ -254,6 +276,12 @@ def _add_common_args(parser):
             "This doesn't affect any DVC operations."
         ),
     )
+    parser.add_argument(
+        "cmd",
+        nargs=argparse.REMAINDER,
```
skshetry, line 281: Looks like this is greedy.

skshetry: We might need to move this back to `-n` like in `dvc run`, if I could not find a way.

pared (approved):

dvc/command/stage.py
efiop: `-n` is a safe choice

skshetry: Okay, so it is greedy. The following works:
```console
$ dvc stage add -d foo -o bar stage_name echo foo bar
```
But, the following does not:
```console
$ dvc stage add stage_name -d foo -o bar echo foo bar
```
It seems `REMAINDER` starts eating up rest of the arguments when the parser starts parsing positional arguments.

skshetry: Yup, looks like it's correct. Note that `REMAINDER` has been removed from documentation as it's considered buggy.
https://bugs.python.org/issue17050

skshetry: Though it seems, we can achieve best of the both worlds (remainder + `-c`) by making `-c` a remainder rather than a positional argument.

dberenbaum: My vote is to mimic the 1.0 dvc run syntax unless there's a compelling justification for changing it:

`$ dvc stage add -n stage_name -d foo -o bar rest of the things are command`

Any concerns besides what was in #5342?

skshetry: Moving back to `-n` then.

Issue #5247, opened 2021-01-12
experiments: checkpoints cleanup/research issues


pmrowla:
* [x] Investigate whether we can have finer control over `rich` table column collapse behavior - `--no-pager` with wide tables results in several empty columns that make output unreadable (#5381)
* [x] Preserve metrics/params column order (related to #5009)
* [x] Add short option for `dvc exp show --all-commits` (#5361)
* [x] `dvc repro` should just fail on `checkpoint = true` stages instead of issuing warning and handling as a persist out (#5258)
* [x] ~~`exp res` should be possible after `run --no-exec --checkpoints ...`~~
* [x] Unify `run/resume`, for checkpoints re-introduce `reset` (#5351)
* [x] Investigate duplicate commit issue from #5311

PR #5351, opened 2021-01-28
experiments: drop `exp resume` in favor of auto-resuming in `exp run`

pmrowla:

Related to #5247

* `dvc exp run` will now resume the previous checkpoint run by default (if it exists)
    * `dvc exp run -r/--rev <experiment_rev>` can be used to explicitly resume from a specific revision, if it is omitted the resume point will be determined based on the workspace state (using last run/applied experiment). This option will likely only be useful for queued/parallel runs.
* `dvc exp resume` has been removed
* `dvc exp reset` should be used to explicitly reset/restart a checkpoint run

### DIFF
--- a/dvc/repo/experiments/__init__.py
+++ b/dvc/repo/experiments/__init__.py
@@ -82,7 +82,6 @@ class Experiments:
         r"^(?P<baseline_rev>[a-f0-9]{7})-(?P<exp_sha>[a-f0-9]+)"
         r"(?P<checkpoint>-checkpoint)?$"
     )
-    LAST_CHECKPOINT = ":last"
     EXEC_TMP_DIR = "exps"
 
     StashEntry = namedtuple(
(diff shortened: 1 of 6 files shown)


dvc/repo/experiments/__init__.py
```diff
@@ -192,23 +191,30 @@ def _stash_exp(
...
+                        if branch:
+                            branch_name = ExpRefInfo.from_ref(branch).name
+                        else:
+                            branch_name = ""
                         if self.scm.is_dirty():
-                            logger.debug(
-                                "Resume new checkpoint branch with "
-                                "modifications"
+                            logger.info(
+                                "Modified checkpoint experiment based on "
```
pared, line 200: What about following scenario:
1. Someone has `checkpoint` experiment
2. It runs for some epochs in workspace
3. We edit workspace (eg code) and run `dvc exp run`
4. `dvc exp apply`
5. We have repo state which we will not be able to reproduce.
Do we intend to allow that? Maybe we should warn user here that it might result in unreproducible experiment?

pared (approved):
LGTM, one question

dvc/repo/experiments/__init__.py
pmrowla: Editing code in the workspace is supported, and this essentially creates a new branch from the last applied/run checkpoint commit. So if you run 5 epochs, and then want to make a branch from the 3rd epoch w/some modified params or code, the workflow is to do 

```
dvc exp run  # initial run to generate 5 epochs
dvc exp apply <SHA_for_3rd_epoch_commit>
vim script.py  # edit code
dvc exp run # run some more epochs w/that modification
```

at the end of this, you'd end up with 2 separate checkpoint refs (git branches) that share the same first 3 epochs (commits) and diverge after that point

This is all still reproducible (as long as the user's code for a single epoch is reproducible) and internally is essentially the same thing as doing `dvc repro` + `git commit` 5 times (once for each epoch), then checking out the 3rd commit, and starting a new git branch

PR #5352, opened 2021-01-28
tests: reduce checkpoint experiment iterations

pmrowla:

Cuts `tests/func/experiments` from 137s to 84s on my machine

### DIFF
--- a/dvc/api/__init__.py
+++ b/dvc/api/__init__.py
@@ -139,4 +139,4 @@ def make_checkpoint():
         fobj.flush()
         os.fsync(fobj.fileno())
     while os.path.exists(signal_file):
-        sleep(1)
+        sleep(0.1)
--- a/tests/func/experiments/conftest.py
+++ b/tests/func/experiments/conftest.py
@@ -4,6 +4,7 @@ import pytest
 
 from tests.func.test_repro_multistage import COPY_SCRIPT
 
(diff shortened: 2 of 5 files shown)


pmrowla:
Still need further investigation into whether we can optimize the multiprocessing overhead for experiment executors, but in the meantime this should still help

dvc/api/__init__.py
```diff
@@ -139,4 +139,4 @@ def make_checkpoint():
         fobj.flush()
         os.fsync(fobj.fileno())
     while os.path.exists(signal_file):
-        sleep(1)
+        sleep(0.1)
```
efiop, line 142: we could move this into a constant and patch it only for tests. Kinda like we do with moto https://github.com/iterative/dvc/blob/08300d3bcf51f7bd9c8509924110416461b11b60/tests/func/test_s3.py#L13

pared (approved):

dvc/api/__init__.py
pmrowla: Dmitry has mentioned before that he thought this 1s sleep could be sped up in general, so this change doesn't really need to be for tests only.

efiop: Still better to not hardcode such things :slightly_smiling_face:

Issue #5353, opened 2021-01-28
pull: "glob" flag is not passed to command


pbielak:
Hi,
I installed the current master version of DVC and I noticed that, although there is the `--glob` parameter for the `pull` command, it is not being used in the actual code, i.e.,
```py
# dvc/command/data_sync.py 
class CmdDataPull(CmdDataBase):
    (...)
    def run(self):
        from dvc.exceptions import CheckoutError, DvcException
        try:
            stats = self.repo.pull(
                targets=self.args.targets,
                jobs=self.args.jobs,
                remote=self.args.remote,
                all_branches=self.args.all_branches,
                all_tags=self.args.all_tags,
                all_commits=self.args.all_commits,
                with_deps=self.args.with_deps,
                force=self.args.force,
                recursive=self.args.recursive,
                run_cache=self.args.run_cache,
               # glob not filled here -> glob=False
            )
            self.log_summary(stats)
```
The `glob` parameter is always set to the default value (`False`).

I already prepared a fix for that, which I will link in a moment.

Best regards,
Piotr

PR #5354, opened 2021-01-28
Pass missing `glob` flag to `pull` command

pbielak:
The `--glob` flag is parsed from the command line, but in the
`CmdDataPull` class in the invocation the parameter (`self.args.glob`)
is not passed to the `pull` method implementation. Hence, always the
default value (False) is used. This commit fixes this issue and sets the
`glob` parameter properly.

Related to #5353

### DIFF
--- a/dvc/command/data_sync.py
+++ b/dvc/command/data_sync.py
@@ -37,6 +37,7 @@ class CmdDataPull(CmdDataBase):
                 force=self.args.force,
                 recursive=self.args.recursive,
                 run_cache=self.args.run_cache,
+                glob=self.args.glob,
             )
             self.log_summary(stats)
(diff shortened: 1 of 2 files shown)


dvc/command/data_sync.py
```diff
@@ -37,6 +37,7 @@ def run(self):
                 force=self.args.force,
                 recursive=self.args.recursive,
                 run_cache=self.args.run_cache,
+                glob=self.args.glob,
```
efiop, line 40: Looks like we are also missing it in fetch, but that will require some additional modification, so out of scope for this PR.

efiop (approved):
Thank you @pbielak ! :pray:

PR #5360, opened 2021-01-29
exp apply: force apply by default

pmrowla:

After discussion with @dmpetrov and @dberenbaum, overwriting any conflicting workspace changes with `--force` should be the default behavior when applying an experiment commit. Due to the nature of how we hide git usage behind experiments, it is not obvious to the user why applying an experiment can cause git conflicts, and in typical use cases for `exp apply`, overwriting conflicting changes is likely the behavior the user was already expecting to begin with.

* `--no-force` can be used to disable overwriting existing changes

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -764,10 +764,10 @@ def add_parser(subparsers, parent_parser):
         formatter_class=argparse.RawDescriptionHelpFormatter,
     )
     experiments_apply_parser.add_argument(
-        "-f",
-        "--force",
-        action="store_true",
-        help="Overwrite any conflicting changes.",
+        "--no-force",
+        action="store_false",
+        dest="force",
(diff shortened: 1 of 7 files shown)


pared (approved):

jorgeorpinel:
I don't think this was documented.

pmrowla:
this PR is from before there were actual docs for experiments commands. It most likely got lost in the shuffle between when we stopped using the wiki for docs, and when my original command ref PR was opened and then closed (to be split up and managed separately on the docs repo side)

jorgeorpinel:
@pmrowla feel free to contribute something to #2421 (if you have a chance) and I'll polish it up. Thanks

jorgeorpinel:
> this PR is from before there were actual docs for experiments commands

Makes sense!

jorgeorpinel:
I just happened to walk into this when updating all the `exp` cmd options (#2421).

PR #5361, opened 2021-01-29
exp show: cleanup UI issues

pmrowla:

Related to #5247

* Adds short form `-A/--all-commits` for `dvc exp show`
* Adds `-n/--num <num>` option for `dvc exp show` to show the last `n` commits from HEAD (defaults to 1)
* Fix issue where duplicate (empty) commit would be created for checkpoint runs with untracked files present in workspace

### DIFF
--- a/tests/func/experiments/test_show.py
+++ b/tests/func/experiments/test_show.py
@@ -206,7 +206,7 @@ def test_show_multiple_commits(tmp_dir, scm, dvc, exp_stage):
 
     expected = {"workspace", init_rev, next_rev}
     results = dvc.experiments.show(num=2)
-    assert set(results.keys()) == {"workspace", init_rev, next_rev}
+    assert set(results.keys()) == expected
 
     expected = {"workspace"} | set(scm.branch_revs("master"))
     results = dvc.experiments.show(all_commits=True)
--- a/tests/func/experiments/test_show.py
+++ b/tests/func/experiments/test_show.py
@@ -192,3 +192,17 @@ def test_show_filter(tmp_dir, scm, dvc, exp_stage, capsys):
     cap = capsys.readouterr()
 
     assert f"{div} foo {div}" not in cap.out
+
+
+def test_show_multiple_commits(tmp_dir, scm, dvc, exp_stage):
+    init_rev = scm.get_rev()
+    tmp_dir.scm_gen("file", "file", "commit")
+    next_rev = scm.get_rev()
+
+    expected = {"workspace", init_rev, next_rev}
+    results = dvc.experiments.show(num=2)
+    assert set(results.keys()) == {"workspace", init_rev, next_rev}
+
+    expected = {"workspace"} | set(scm.branch_revs("master"))
+    results = dvc.experiments.show(all_commits=True)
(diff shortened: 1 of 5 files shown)


pmrowla:
Questions for @dberenbaum & @dmpetrov:

* Does `-A` work as a short form for `exp show --all-commits`? `-a` would be nicer but we would end up conflicting with the existing usage for `-a/--all-branches` that is used everywhere else in DVC
* Should `-n` default to something other than 1 (i.e. should we show the last 2 or 3 commits by default instead of only HEAD)?

dberenbaum:
I want to make sure I'm understanding the use case. These refer to permanent commits on the workspace branch, right? So a workflow where this would be useful looks something like:

1. `git commit -m "set up dvc project"`
2. `dvc exp run` (one or more times)
3. `dvc exp apply`
4. `git commit -m "keep best experiment"`
5. `dvc exp show -A`/`dvc exp show -n 2` (to see experiments from step 2)

The options and defaults you chose look good.

One thing I find a little confusing is that the experiments are associated with the commit in step 1 rather than the commit in step 4. It makes sense since experiments can't be associated with a nonexistent commit when they are run, but it might need to be explained to users in the docs/user guides to avoid confusion.

pmrowla:
> I want to make sure I'm understanding the use case. These refer to permanent commits on the workspace branch, right?

This is correct, these options refer to the number of "parent" git commits that we display in the table - which normally only defaults to the single HEAD commit.

> One thing I find a little confusing is that the experiments are associated with the commit in step 1 rather than the commit in step 4. It makes sense since experiments can't be associated with a nonexistent commit when they are run, but it might need to be explained to users in the docs/user guides to avoid confusion.

Experiments are grouped by their shared parent git commit, in the docs and code this is usually referred to as the baseline or parent commit, and experiments are described as being derived from that baseline/parent.

As you noted, there's not really any way for DVC to know if the follow-up commit in step 4 is actually related to any of those experiments so we group them by the shared parent git commit instead.

dberenbaum:
> Experiments are grouped by their shared parent git commit, in the docs and code this is usually referred to as the baseline or parent commit, and experiments are described as being derived from that baseline/parent.

Parent commit seems clear to me since it describes the relationship of the commits and is consistent with the existing git terminology. We have so many new concepts and terms for experiments that we may need a few iterations of the documentation to get them right.

tests/func/experiments/test_show.py
```diff
@@ -192,3 +192,25 @@ def test_show_filter(tmp_dir, scm, dvc, exp_stage, capsys):
...
+    init_rev = scm.get_rev()
+    tmp_dir.scm_gen("file", "file", "commit")
+    next_rev = scm.get_rev()
+
+    with pytest.raises(InvalidArgumentError):
+        dvc.experiments.show(num=-1)
+
+    expected = {"workspace", init_rev, next_rev}
+    results = dvc.experiments.show(num=2)
+    assert set(results.keys()) == {"workspace", init_rev, next_rev}
```
pared, line 209:
```suggestion
    assert set(results.keys()) == expected
```

pared (approved):

PR #5363, opened 2021-01-29
params: support unicode characters

### LINKED ISSUE #5357: repro: a Greek letter in parameters name cause an error
# Bug Report

## Description

`dvc repro --dry` throws the error
```
ERROR: failed to reproduce 'dvc.yaml': Parameters 'background_fits.mÎ©b_bgd_range' are missing from 'settings.json'.
```
clearly misreplacing the greek symbol `Ω`

The name of the parameter is correct in dvc.yaml
```
    - settings.json:
      - background_fits.mΩb_bgd_range
```

It is correct in settings.json:
```
       "background_fits": {
                              "mΩb_bgd_range": [
                                                 6.1,
                                                 6.4
                                               ]
                          },
```
Also fine in dvc.lock
```
      background_fits.mΩb_bgd_range:
      - 6.1
      - 6.4
```

### Environment information

**Output of `dvc version`:**

```console
$ dvc version
```

```
λ dvc version
DVC version: 1.11.10 (exe)
Platform: Python 3.7.9 on Windows-10-10.0.19041-SP0
Supports: All remotes
Cache types: hardlink
Cache directory: NTFS on C:\
Caches: local
Remotes: None
Workspace directory: NTFS on C:\
Repo: dvc, git
```

isidentical:
resolves #5357

### DIFF
--- a/tests/unit/test_params.py
+++ b/tests/unit/test_params.py
@@ -25,4 +25,10 @@ def test_params_order(tmp_dir, dvc, dummy_stage):
 def test_repro_unicode(tmp_dir, dvc, dummy_stage):
     tmp_dir.gen({"settings.json": '{"Ω_value": 1}'})
     dummy_stage(params=[{"settings.json": ["Ω_value"]}])
-    dvc.reproduce(dry=True)
+    (stage,) = dvc.reproduce(dry=True)
+    stage.cmd = "foo"
+    stage.dvcfile._lockfile.dump(stage)
+
+    dvc.remove(stage.name)
+    assert not (tmp_dir / "dvc.yaml").exists()
+    assert not (tmp_dir / "dvc.lock").exists()
--- a/tests/unit/test_params.py
+++ b/tests/unit/test_params.py
@@ -20,3 +20,9 @@ def test_params_order(tmp_dir, dvc, dummy_stage):
 
     # params are sorted during dumping, therefore p1 is first
     assert list(dvc.params.show()[""]) == ["p1.yaml", p2_path, "params.yaml"]
+
+
+def test_repro_unicode(tmp_dir, dvc, dummy_stage):
+    tmp_dir.gen({"settings.json": '{"Ω_value": 1}'})
+    dummy_stage(params=[{"settings.json": ["Ω_value"]}])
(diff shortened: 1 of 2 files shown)


skshetry (approved):

skshetry:
Looks like there are two more places to solve this where we are not doing this that might affect `dvc remove` command.

https://github.com/iterative/dvc/blob/08300d3bcf51f7bd9c8509924110416461b11b60/dvc/dvcfile.py#L279
https://github.com/iterative/dvc/blob/08300d3bcf51f7bd9c8509924110416461b11b60/dvc/dvcfile.py#L383

tests/unit/test_params.py
```diff
@@ -20,3 +20,15 @@ def test_params_order(tmp_dir, dvc, dummy_stage):
...
     # params are sorted during dumping, therefore p1 is first
     assert list(dvc.params.show()[""]) == ["p1.yaml", p2_path, "params.yaml"]
+
+
+def test_repro_unicode(tmp_dir, dvc, dummy_stage):
+    tmp_dir.gen({"settings.json": '{"Ω_value": 1}'})
+    dummy_stage(params=[{"settings.json": ["Ω_value"]}])
+    (stage,) = dvc.reproduce(dry=True)
+    stage.cmd = "foo"
+    stage.dvcfile._lockfile.dump(stage)
```
skshetry, line 30: Let's just use `stage.dump()` to both files. This way, both the `remove_stage` will be used. 🙂

isidentical: I thought `dummy_stage()` already did it

isidentical: *it doesn't create a `dvc.lock` file but rather only dumps the stage itself to the `dvc.yaml` 
https://github.com/iterative/dvc/blob/08300d3bcf51f7bd9c8509924110416461b11b60/tests/conftest.py#L148-L151

skshetry: Okay, let's leave it then. I made a mistake on that `stage.dump` not dumping lockfile by default.  Considering the callee of it, it's not a bug, but we should keep the behavior same here. I'll look into later.

Issue #5371, opened 2021-02-01
dvc.yaml: very hard to edit, can we detect and do human readable/helpful schema errors?


shcheklein:
As we are migrating from using `dvc run`, `dvc import-url`, and other helpers when we create stages, it makes creating a proper dvc.yaml very painful. Trying to add a stage, I'm getting errors like this:

```
ERROR: unable to read: 'dvc.yaml', YAML file structure is corrupted: mapping values are not allowed in this context
  in "<unicode string>", line 7, column 14
```

or this:

```
ERROR: unable to read: 'dvc.yaml', YAML file structure is corrupted: while parsing a block collection
  in "<unicode string>", line 6, column 7
did not find expected '-' indicator
  in "<unicode string>", line 7, column 9
```

or this:

```
'dvc.yaml' format error: expected str @ data['stages']['prepare_genome']['deps'][0]
```

(from https://github.com/iterative/dvc/issues/5370)

or infamous one with `size`.

From my personal experience you basically have to go to docs and manually compare and experiment until it's fixed.

I wonder if we can produce some meaningful error messages, include examples if needed of the affected section?

skshetry:
We might be able to add a linter to provide better error message during syntax errors (see: https://github.com/adrienverge/yamllint).

dberenbaum:
More generically, how can stage creation in `dvc.yaml` be easier? If `dvc run` is deprecated, users may find it harder to create stages. Does this need to be prioritized before the 2.0 release?

dberenbaum:
I forgot about `dvc stage add`, which should help users create stages without manually editing `dvc.yaml` and make this less high priority. Regardless, it would be a useful feature for users who do manually edit `dvc.yaml`.

dberenbaum:
`voluptuous` is already being used to perform validation and provide error messages like the last one, right? I wonder if custom exceptions would be more helpful, or if another data validation library provides more useful messages.

skshetry:
One good workaround around this would be to try to reparse the yaml using `YAML(type="rt")` on error which provides better error messages and contexts around it (compared to the C-parser).

```
ScannerError: while scanning a simple key
  in "<unicode string>", line 2, column 1:
    hi
    ^ (line: 2)
could not find expected ':'
  in "<unicode string>", line 2, column 3:
    hi
      ^ (line: 2)
```

Also, the provided exception has more details in it this way, which we can use to wrap it and provide better (colorful and structured) error messages.

skshetry:
[...]

Issue #5370, opened 2021-01-31
deps is inconsistent between dvc.yaml and .dvc files


antonkulaga:
When I have something like:
```
    deps:
      - path: http://ftp.ensembl.org/pub/release-102/fasta/homo_sapiens/dna/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz
        etag: W/"5f8d62ed-34863818"
        cache: false
```
It works find in .dvc files, but when I add it to one of the stages, I get an error as it assumes deps are string only.
```
stages:
  prepare_genome:
    frozen: true
    deps:
      - path: http://ftp.ensembl.org/pub/release-102/fasta/homo_sapiens/dna/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz
        etag: W/"5f8d62ed-34863818"
        cache: false
    cmd: bin/prepare_genome.sh
    out:
      - data/gwas/homo_sapiens_ensembl_102/Homo_sapiens.GRCh38.dna.primary_assembly.fa
```
the error is:
```
 'dvc.yaml' format error: expected str @ data['stages']['prepare_genome']['deps'][0]
```
the dvc version is 1.11.10 (latest .deb file).
I suggest that "deps" should behave the same in .dvc and dvc.yaml as otherwise it creates a lot of confusion.

skshetry:
@antonkulaga, the old `.dvc` file support is going to be removed in the upcoming release next month.  It's only supported for `dvc add/import`.

antonkulaga:
@skshetry I was not talking about having stages in .dvc files, but having:
```
deps:
      - path: http://ftp.ensembl.org/pub/release-102/fasta/homo_sapiens/dna/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz
        etag: W/"5f8d62ed-34863818"
        cache: false
```
inside of the dependencies in dvc.yaml So far only "str" is supported for deps in dvc.yaml, I believe it should also support etags for external dependencies

skshetry:
@antonkulaga, the checksums are kept in lockfile. You just have to write the entry as follows and then `dvc repro` it later:
```yaml
stages:
  prepare_genome:
    cmd: bin/prepare_genome.sh
    deps:
      - http://ftp.ensembl.org/pub/release-102/fasta/homo_sapiens/dna/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz
    out:
[...]

Issue #5312, opened 2021-01-22
Future dvc.yaml 2.0 improvements


jorgeorpinel:
UPDATE: Jump to https://github.com/iterative/dvc/issues/5312#issuecomment-766506842 for remaining discussions

> Consolidates #5164, #5165, and #5180.
> For context, see existing docs in https://dvc.org/doc/user-guide/dvc-files/advanced-dvc-yaml.

Some remaining concerns in order or relevance (only 1 and 2.1 seem more or less worth considering in the immediate term).

> The min. purpose of this issue is to at least decide which, if any, to address, and that we can do it in the future without breaking backward-compatibility.

### 1. Behavior: default params file & merging of values

We decided to always include * from params.yaml in `vars`. Trying to redefine values (via other file includes or write-in vars) fails except if they're objects that can be merged (no leaf node conflicts).

What about reconsidering the "merging" of vals/objects? Instead we could rename tree paths so there's no possibility of conflicts (@shcheklein's idea). E.g.:

```yaml
# params.yaml
foo: "foo"
```
```js
// params.json
{"foo": "bar"}
```
```yaml
vars:
- foo # use as ${foo}, ${params.yaml.foo}, or ${paramsyaml.foo}
- params.json:foo # use as ${params.json.foo} or ${paramsjson.foo}
```
```yaml
vars:
- foo # use as ${params.yaml:foo} or ${paramsyaml.foo}
- params.json:foo # use as ${params.json.foo} or ${paramsjson.foo}
- foo: "baz" # use as ${foo} or ${globals.foo}
```

### 2. Syntax (minor)

Is it worth renaming certain keywords for accuracy? The suggestions below come from the terminology that we ended up using (so far) in docs:

2.1 `vars:` aren't really variables (not using that term in docs). How about `include:` or `load:` to invoke an action; `values:` (we use that term a lot in docs),`const:`, or `globals/locals:` for a descriptive term?
> The most accurate term would be `params:`, but that is taken for the local scope... I wonder if we could overload the section for both things (when it assigns a value it's a SET, then it just lists the key it's a GET).

[...]

PR #5373, opened 2021-02-01
don't allow loading stages without `cmd`

skshetry:
This was loosened when introducing parametrization.
But, `foreach`..`do` should be considered first
before the stage's regular structure. And, cmd is made
required (as it should have been)

Related: #5371, #5370, #5312

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -145,8 +145,9 @@ def dummy_stage(tmp_dir, dvc):
     def make(path="dvc.yaml", name="dummy_stage", **kwargs):
         from dvc.stage import PipelineStage, create_stage
 
+        cmd = kwargs.get("cmd", "command")
         stage = create_stage(
-            PipelineStage, dvc, path, name=name, cmd="", **kwargs
+            PipelineStage, dvc, path, name=name, cmd=cmd, **kwargs
         )
         stage.dump()
         return stage
--- a/tests/unit/command/test_status.py
+++ b/tests/unit/command/test_status.py
@@ -6,7 +6,7 @@ from dvc.cli import parse_args
 from dvc.command.status import CmdDataStatus
 
 
-def test_cloud_status(mocker):
+def test_cloud_status(tmp_dir, dvc, mocker):
     cli_args = parse_args(
         [
             "status",
--- a/dvc/schema.py
+++ b/dvc/schema.py
@@ -1,4 +1,4 @@
-from voluptuous import Any, Exclusive, Optional, Required, Schema
+from voluptuous import Any, Optional, Required, Schema
 
 from dvc import dependency, output
 from dvc.hash_info import HashInfo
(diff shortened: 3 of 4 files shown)


tests/conftest.py
```diff
@@ -145,8 +145,9 @@ def dummy_stage(tmp_dir, dvc):
     def make(path="dvc.yaml", name="dummy_stage", **kwargs):
         from dvc.stage import PipelineStage, create_stage
 
+        cmd = kwargs.get("cmd", "command")
         stage = create_stage(
-            PipelineStage, dvc, path, name=name, cmd="", **kwargs
+            PipelineStage, dvc, path, name=name, cmd=cmd, **kwargs
```
skshetry, line 150: I'll look into this on later iterations. It should never be empty.

tests/unit/command/test_status.py
```diff
@@ -6,7 +6,7 @@
 from dvc.command.status import CmdDataStatus
 
 
-def test_cloud_status(mocker):
+def test_cloud_status(tmp_dir, dvc, mocker):
```
skshetry, line 9: We might need to move some of the logic in `status` command to API.

pared:
This PR contradicts the use case specified in [here](https://github.com/iterative/dvc/issues/4446#issuecomment-757373208). I think its not common use case, but still we have been working on 4446 to make it work, @dmpetrov what is the desired behavior?

skshetry:
@pared, I assumed "dummy stage" being a high-level term. Could you please elaborate on that?
Do we need to support it in the `dvc.yaml` file?

pared:
@skshetry so the `dummy stage` was supposed to be an empty stage that would only contain `params/metrics/plots` so that one could trigger default behavior for `dvc {params/plots/metrics} show/diff`. I am not sure what is the particular use case but it was one of the requirements for #4446 - I guess we need some high-level perspective here.

efiop:
@pared That's more of a workaround for our current inability to put metrics/etc into the dvc.yaml without creating a pipeline stage. If you remember, we were discussing something like:
```
data:
- data1
imports:
...
metrics:
- metric1
- metric2
plots:
...   
stages:
....
```

for the future. Do we need that workaround now? If we do, we could consider proper approach instead of bringing back a hack. CC @dberenbaum

dberenbaum:
Would metrics optionally have dependencies in the schema above?

efiop:
> Would metrics optionally have dependencies in the schema above?

@dberenbaum Not really. At least we didn't see any need for that yet. Theoretically the proposed schema could be extended with anything under specific `metrics` entry, but the scenario for declaring metric dependencies without a stage doesn't make sense. The only case where we handle such a thing is in `import`s as an edge case - if there is one dependency and one output and no command we assume that this is import.

dberenbaum:
I was assuming that all metrics would be defined at the top level, but it sounds like the idea is to support top-level metrics and metrics defined within stages?

efiop:
@dberenbaum I was also thinking about that, but not as a new dag in top-level metrics, but more of a high-er level thing, where in stages there are only deps and outputs(e.g. metrics.json is an output) but then at the top level you also declare that this is not only an output but also a metric. But it seems like it is convenient to have support for both options: declare metrics inside of the stages and have an ability to declare them at the top level if they don't have their own stage or choose not to have it.

PR #5375, opened 2021-02-01
Amazon S3: change parsing of path

### LINKED ISSUE #5130: s3 remote: unexpected error with session token
# Bug Report

## Description

When using a session token for aws s3 remote the dvc push produces the following error:
ERROR: unexpected error - An error occurred (AccessDenied) when calling the ListObjects operation: Access Denied

### Reproduce

1. Configure ~/aws/credentials with access_key, secret_access_key and session_token; profile name: myprofile
2. dvc remote add -d myremote s3://mybucket/path
3. dvc remote modify myremote profile myprofile
5. dvc add test.txt
4. dvc push

Alternative:
1. Configure an aws sso profile named myprofile
2. aws sso login --profile myprofile
3. dvc remote add -d myremote s3://mybucket/path
4. dvc remote modify myremote profile myprofile
5. dvc add test.txt
6. dvc push

### Expected

test.txt is pushed to the S3 bucket

### Environment information

**Output of `dvc version`:**

```console
$ dvc version
DVC version: 1.11.7 (brew)
Platform: Python 3.9.1 on macOS-11.0.1-x86_64-i386-64bit
Supports: azure, gdrive, gs, http, https, s3, ssh, oss, webdav, webdavs
Caches: local
Remotes: s3
Repo: dvc, git
```

Christoph-1:
Fixes #5130

### DIFF
--- a/dvc/tree/s3.py
+++ b/dvc/tree/s3.py
@@ -210,7 +210,7 @@ class S3Tree(BaseTree):
 
     def _list_paths(self, path_info, max_items=None):
         with self._get_bucket(path_info.bucket) as bucket:
-            obj_summaries = bucket.objects.filter(Prefix=path_info.path)
+            obj_summaries = bucket.objects.filter(Prefix=(path_info / "/").path)
             if max_items is not None:
                 obj_summaries = obj_summaries.page_size(max_items).limit(
                     max_items


dvc/tree/s3.py
```diff
@@ -210,7 +210,7 @@ def getsize(self, path_info):
 
     def _list_paths(self, path_info, max_items=None):
         with self._get_bucket(path_info.bucket) as bucket:
-            obj_summaries = bucket.objects.filter(Prefix=path_info.path)
+            obj_summaries = bucket.objects.filter(Prefix=(path_info / "/").path)
```
pmrowla, line 213: This fix won't be sufficient because `path_info` won't always be the path to a directory here.

S3 supports arbitrary prefixes, meaning `path_info` could be a path to something like `ab/c` where we want to retrieve the list of any key in the S3 bucket that starts with the string `ab/c`, (so it would match `ab/cdef0`, `ab/cdef1`, etc). In this case, adding the trailing slash will end up returning the wrong set of keys.

efiop: Great point @pmrowla !

efiop:
For the record: this might be fixed in `s3fs`. We'll be migrating to it soon and will double-check this issue.

isidentical:
We have rewritten our s3 implementation, so this patch became obsolete. Let's discuss whether the underlying issue is resolved or not in the issue (if not, perhaps a patch against s3fs would be more suitable (depending on the problem)).

PR #5378, opened 2021-02-01
config: Support ~/.aws/config parsing

### LINKED ISSUE #4359: Support additional configuration parameters for S3 transfer
**Problem**
When using S3 as remote for DVC, boto3 reads also the configuration file `~/.aws/config` typically used by the original AWS CLI tool. Not all configuration parameters from this file have an effect in DVC at the moment. For instance, the last line `multipart_threshold = 500MB` will be skipped by DVC while the previous line `signature_version = s3` will be applied when pushing data into S3:      
```ini
[default]
output = json
s3 =
  signature_version = s3
  multipart_threshold = 500MB
```
**Suggestion**
Suggestion is to support optional yet useful parameters from `~/.aws/config` as per [AWS CLI S3 documentation](https://docs.aws.amazon.com/cli/latest/topic/s3-config.html) to control S3 transfers for performance reasons or to unlock the advanced configuration capabilities. 

These parameters should be passed to TransferConfig:
- `max_concurrent_requests` - The maximum number of concurrent requests.
- `max_queue_size` - The maximum number of tasks in the task queue.
- `multipart_threshold` - The size threshold the CLI uses for multipart transfers of individual files.
- `multipart_chunksize` - When using multipart transfers, this is the chunk size that the CLI uses for multipart transfers of individual files.
- `max_bandwidth` - The maximum bandwidth that will be consumed for uploading and downloading data to and from Amazon S3.

If it is reasonable and doable - support also the following set of parameters:
- `use_accelerate_endpoint` - Use the Amazon S3 Accelerate endpoint for all s3 and s3api commands. You must first enable S3 Accelerate on your bucket before attempting to use the endpoint. This is mutually exclusive with the use_dualstack_endpoint option.
- `use_dualstack_endpoint` - Use the Amazon S3 dual IPv4 / IPv6 endpoint for all s3 and s3api commands. This is mutually exclusive with the use_accelerate_endpoint option.
- `addressing_style` - Specifies which addressing style to use. This controls if the bucket name is in the hostname or part of the URL. Value values are: path, virtual, and auto. The default value is auto.
- `payload_signing_enabled` - Refers to whether or not to SHA256 sign sigv4 payloads. By default, this is disabled for streaming uploads (UploadPart and PutObject) when using https.

These values must be set under the top level `s3` key in the AWS Config File, which has a default location of `~/.aws/config`. Below is an example configuration:
```ini
[profile development]
aws_access_key_id=foo
aws_secret_access_key=bar
s3 =
  max_concurrent_requests = 20
  max_queue_size = 10000
  multipart_threshold = 64MB
  multipart_chunksize = 16MB
  max_bandwidth = 50MB/s
  use_accelerate_endpoint = true
  addressing_style = path
```

isidentical:

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here. (iterative/dvc.org#2140)

resolves #4359

### DIFF
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -138,6 +138,7 @@ SCHEMA = {
                     "region": str,
                     "profile": str,
                     "credentialpath": str,
+                    "configpath": str,
                     "endpointurl": str,
                     "access_key_id": str,
                     "secret_access_key": str,
--- a/dvc/tree/s3.py
+++ b/dvc/tree/s3.py
@@ -98,9 +98,11 @@ class S3Tree(BaseTree):
         from boto3.s3.transfer import TransferConfig
         from botocore.configloader import load_config
 
-        config = load_config(
-            os.environ.get("AWS_CONFIG_FILE", _AWS_CONFIG_PATH)
-        )
+        config_path = os.environ.get("AWS_CONFIG_FILE", _AWS_CONFIG_PATH)
+        if not os.path.exists(config_path):
+            return None
+
+        config = load_config(config_path)
         profile = config["profiles"].get(self.profile or "default")
         if not profile:
             return None
--- a/dvc/tree/s3.py
+++ b/dvc/tree/s3.py
@@ -11,11 +11,13 @@ from dvc.hash_info import HashInfo
 from dvc.path_info import CloudURLInfo
 from dvc.progress import Tqdm
 from dvc.scheme import Schemes
-from dvc.utils import error_link
+from dvc.utils import conversions, error_link
 
 from .base import BaseTree
 
 logger = logging.getLogger(__name__)
+
+_AWS_CONFIG_PATH = os.path.expanduser("~/.aws/config")
 
 
 class S3Tree(BaseTree):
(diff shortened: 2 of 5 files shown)


dvc/config_schema.py
```diff
@@ -138,6 +138,7 @@ class RelPath(str):
                     "region": str,
                     "profile": str,
                     "credentialpath": str,
+                    "configpath": str,
```
shcheklein, line 141: qq - is it different from the `credentialpath`?

isidentical: `credentialpath` is the replacement for `~/.aws/credentials` file, and `configpath` would be the replacement for `~/.aws/config`

shcheklein: yep, I see.

dvc/tree/s3.py
```diff
@@ -60,6 +62,56 @@ def __init__(self, repo, config):
         if shared_creds:
             os.environ.setdefault("AWS_SHARED_CREDENTIALS_FILE", shared_creds)
 
+        config_path = config.get("configpath")
+        if config_path:
+            os.environ.setdefault("AWS_CONFIG_FILE", config_path)
+        self._transfer_config = None
+
+    # https://github.com/aws/aws-cli/blob/0376c6262d6b15dc36c82e6da6e1aad10249cc8c/awscli/customizations/s3/transferconfig.py#L107-L113
+    _TRANSFER_CONFIG_ALIASES = {
```
isidentical, line 71: The exposed `TransferConfig` by `boto` doesn't accept `max_bandwith`, so we don't support it even though aws-cli does (through [s3transfer](https://pypi.org/project/s3transfer/))

dvc/tree/s3.py
```diff
@@ -60,6 +62,56 @@ def __init__(self, repo, config):
...
+
+    def _process_config(self):
+        from boto3.s3.transfer import TransferConfig
+        from botocore.configloader import load_config
+
+        config_path = os.environ.get("AWS_CONFIG_FILE", _AWS_CONFIG_PATH)
+        if not os.path.exists(config_path):
+            return None
+
+        config = load_config(config_path)
```
isidentical, line 105: The underlying function calls `os.path.expanduser` anyways, so we don't need to expand the user's config path again (we do it for the `_AWS_CONFIG_PATH`, since it is a global and might be required elsewhere)

dvc/utils/conversions.py
```diff
@@ -0,0 +1,24 @@
+# https://github.com/aws/aws-cli/blob/5aa599949f60b6af554fd5714d7161aa272716f7/awscli/customizations/s3/utils.py
+
+MULTIPLIERS = {
```
isidentical, line 3: These are kept in the same way as the `aws-cli`'s source

pared (approved):
LGTM!

dvc/tree/s3.py
```diff
@@ -11,12 +11,14 @@
...
 from dvc.progress import Tqdm
 from dvc.scheme import Schemes
-from dvc.utils import error_link
+from dvc.utils import conversions, error_link
 
 from .base import BaseTree
 
 logger = logging.getLogger(__name__)
 
+_AWS_CONFIG_PATH = os.path.expanduser("~/.aws/config")
```
efiop, line 20: Does it work on windows? Or this is how awscli does it too?

isidentical:
```
and place it in ``~/.aws/credentials`` (or in
``%UserProfile%\.aws/credentials`` on Windows). If you wish to place the
shared credentials file in a different location than the one specified
above, you need to tell aws-cli where to find it. Do this by setting the
appropriate environment variable:
```

Which is what `os.path.expanduser` does on the windows

> On Windows, USERPROFILE will be used if set, otherwise a combination of HOMEPATH and HOMEDRIVE will be used. An initial ~user is handled by stripping the last directory component from the created user path derived above.

efiop: @isidentical Right, but will the `/` be handled properly? Or do we need to `os.path.join(os.path.expanduser("~"), ".aws", "config")`?

isidentical: Ah completely forgot about that 🤦🏻 fixing it

isidentical: Fixed this, and modified one of the tests to actually use this path.

dvc/utils/conversions.py
```diff
@@ -0,0 +1,24 @@
+# https://github.com/aws/aws-cli/blob/5aa599949f60b6af554fd5714d7161aa272716f7/awscli/customizations/s3/utils.py
```
efiop, line 1: Since this whole file is awscli-specific, we could move this stuff to `dvc/tree/s3.py` or create something like `dvc/tree/s3/utils.py` and it put it there. This piece is quite small, so could go with the former no problem.

isidentical: I think keeping here is OK (even though this is a pre-emptive assumption) I believe this utility might come in handy in other places since it does a very general job. The awscli-specific part here is the handling of IEC suffixes, though it doesn't differ much and something we can ignore.

efiop: Ok, let's keep it then :+1:

PR #5381, opened 2021-02-02
exp show: customize rich table rendering behavior

### LINKED ISSUE #5247: experiments: checkpoints cleanup/research issues

pmrowla:

Will close #5247

* Updates rich version to 9.x, uses native rich pager support
* Re-enables color/styling in exp show table
* When using `--no-pager`, columns will be collapsed in right-to-left order if the table cannot fit into the console
    * At least one metric column and one params column will always be displayed whenever possible.
    * Params will be collapsed before metrics
    * If the table still does not fit after collapsing all but one metric and param column, will fallback to rich's folding behavior for the remaining columns

TODO

* [x] Wrap rich column/table classes to collapse/merge metrics & params columns which can't be displayed due to terminal width

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -302,7 +302,7 @@ def _parse_filter_list(param_list):
 
 
 def _experiments_table(all_experiments, **kwargs):
-    from rich.table import Table
+    from dvc.utils.table import Table
 
     include_metrics = _parse_filter_list(kwargs.pop("include_metrics", []))
     exclude_metrics = _parse_filter_list(kwargs.pop("exclude_metrics", []))
@@ -347,10 +347,13 @@ def _add_data_columns(table, names, **kwargs):
     count = Counter(
         name for path in names for name in names[path] for path in names
     )
+    first = True
     for path in names:
         for name in names[path]:
             col_name = name if count[name] == 1 else f"{path}:{name}"
+            kwargs["collapse"] = False if first else True
             table.add_column(col_name, **kwargs)
+            first = False
 
 
 def _format_json(item):
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -318,11 +318,21 @@ def _experiments_table(all_experiments, **kwargs):
     )
 
     table = Table()
-    table.add_column("Experiment", no_wrap=True)
+    table.add_column(
+        "Experiment", no_wrap=True, header_style="black on grey93"
+    )
     if not kwargs.get("no_timestamp", False):
-        table.add_column("Created")
-    _add_data_col(table, metric_names, justify="right", no_wrap=True)
-    _add_data_col(table, param_names, justify="left")
+        table.add_column("Created", header_style="black on grey93")
+    _add_data_columns(
+        table,
+        metric_names,
+        justify="right",
+        no_wrap=True,
+        header_style="black on cornsilk1",
+    )
+    _add_data_columns(
+        table, param_names, justify="left", header_style="black on light_cyan1"
+    )
 
     for base_rev, experiments in all_experiments.items():
         for row, _, in _collect_rows(
@@ -333,7 +343,7 @@ def _experiments_table(all_experiments, **kwargs):
     return table
 
 
-def _add_data_col(table, names, **kwargs):
+def _add_data_columns(table, names, **kwargs):
     count = Counter(
         name for path in names for name in names[path] for path in names
     )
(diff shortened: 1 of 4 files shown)


pmrowla:
Collapse behavior example using jameson-metrics:

current master:
```
$ dvc exp show --no-pager
┏━━━━━━━━━━━━┳━┳━━━━━━━━━┳━━━━━┳━━━━━┳━━━━━━━━┳━━━━━━━━━┳━┳━┳━┳━┳━┳━┳
┃ Experiment ┃ ┃     AUC ┃  TP ┃  TN ┃   took ┃    loss ┃ ┃ ┃ ┃ ┃ ┃ ┃
┡━━━━━━━━━━━━╇━╇━━━━━━━━━╇━━━━━╇━━━━━╇━━━━━━━━╇━━━━━━━━━╇━╇━╇━╇━╇━╇━╇
│ workspace  │ │ 0.75265 │ 190 │ 894 │ 206.43 │ 0.85513 │ │ │ │ │ │ │
│ master     │ │ 0.75265 │ 190 │ 894 │ 206.43 │ 0.85513 │ │ │ │ │ │ │
└────────────┴─┴─────────┴─────┴─────┴────────┴─────────┴─┴─┴─┴─┴─┴─┴
```

after this PR:
```
$ dvc exp show --no-pager
┏━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━┳━━━━━┳━┳━━━━━━━━┳━┓
┃ Experiment ┃ Created      ┃     AUC ┃  TP ┃  TN ┃ ┃ L4size ┃ ┃
┡━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━╇━━━━━╇━╇━━━━━━━━╇━┩
│ workspace  │ -            │ 0.75265 │ 190 │ 894 │ │ 24     │ │
│ master     │ Feb 29, 2020 │ 0.75265 │ 190 │ 894 │ │ 24     │ │
└────────────┴──────────────┴─────────┴─────┴─────┴─┴────────┴─┘
```

dberenbaum:
@pmrowla I like your idea to show the first listed metrics and parameters rather than let rich decide, and as you mentioned, maybe the first metric should even be on the right to be next to the first parameter and consistent with viewer.

A couple small suggestions to try (interested in feedback on whether others find these useful @pmrowla @dmpetrov @skshetry):

- Provide a hint above or below the table with something like: `6 of 14 columns shown. Use --include-metrics/--include-params to specify which columns to show`.
- A suggestion from <span>#</span>8 in https://medium.com/@jdxcode/12-factor-cli-apps-dd3c227a0e46 is: "Never output table borders. It’s noisy and a huge pain for parsing." This would at least solve the issue of the empty columns:
```console

  Experiment   Created            AUC    TP    TN     L4size    

  workspace    -              0.75265   190   894     24        
  master       Feb 29, 2020   0.75265   190   894     24        
```

One bigger suggestion is to provide a csv or tsv output option for a few reasons:
- Users likely will have no problem reading csv files into other tools and formatting, sorting, and visualizing however they want, giving users more flexibility and making it less necessary to solve all those issues in cli output.
- Number of columns does not need to be limited to screen width, so the leftmost column that summarizes a complex tree structure can be expanded into multiple columns, one for each level of the tree (and probably at least one additional one for checkpoints). Even rows that are mostly empty are fine, so more info can be provided by adding columns.
- Provides an easily parseable serialization format for saving, sharing, etc.

pmrowla:
> * A suggestion from #8 in https://medium.com/@jdxcode/12-factor-cli-apps-dd3c227a0e46 is: "Never output table borders. It’s noisy and a huge pain for parsing." This would at least solve the issue of the empty columns:

This seems a bit more personal preference? Not having borders makes it harder to read bigger tables imo, but I don't have any strong opinions about it either way.

I do think that we shouldn't worry about making the table machine-parsable though. If users want machine readable data they should just be using a better output format like json (or potentially csv).

pmrowla:
Will update table class location once https://github.com/iterative/dvc/pull/5348 is merged

Issue #4529, opened 2020-09-04
add: improve message for overlapping outputs


shcheklein:
### Please provide information about your setup

**Output of `dvc version`:**

```console
$ dvc version
DVC version: 1.6.6+79e8e4
---------------------------------
Platform: Python 3.8.5 on macOS-10.15.6-x86_64-i386-64bit
Supports: All remotes
Cache types: reflink, hardlink, symlink
Cache directory: apfs on /dev/disk1s1
Workspace directory: apfs on /dev/disk1s1
Repo: dvc, git
```

### Reproduce

Run `dvc add` on a directory with some data twice and get something like:

```console
(.env) √ Projects/test-dvc % dvc add data
Stage is cached, skipping
```

### Problem

~~`stage` term is not applicable here. This message should be fixed/changed.~~ https://github.com/iterative/dvc/pull/5385

### Bonus

Error message on the overlapping data should be avoiding term `output`. Instead we should be using `directory` or `file`. Also should we recommend `dvc add data` instead of commit (might be easier for users, no need to learn a new command).

```
(.env) √ Projects/test-dvc % dvc add data/file3
Adding...
ERROR: Cannot add 'data/file3', because it is overlapping with other DVC tracked output: 'data'.
To include 'data/file3' in 'data', run 'dvc commit data.dvc'
```

gsvolt:
@shcheklein - may I work on this issue?

shcheklein:
@gsvolt yes! absolutely, feel free to take it! thanks 🙏

DiPaolo:
Is it fine to use a message similar I've tried to use at https://github.com/iterative/dvc/pull/5236? Something like this:
```WARNING 'data' is already under DVC control``` 

If it's okay, I will fix the issue soon.

**Comment by efiop** [comment]
Repurposing for the "bonus"

**Comment by skshetry** [comment]
Note that the message these days says something like following:
```
ERROR: bad DVC file name 'data/file3.dvc' is git-ignored.
```

Issue #5368, opened 2021-01-29
run: run is computing checksums even though --no-exec is specified


Honzys:
# Bug Report

run: run cache is not ignored when --no-exec

## Description

When you specify `--no-exec` param, dvc run command should be treated asi if the `--no-run-cache` was also specified. I guess it makes no sense to check the cache when `--no-exec` is toggled, or am I wrong?

The problem is when I run `dvc run --no-exec ...` with dependency pointing to a large folder, it takes a long time to execute this command. Also after the stage is created and I actually want to run the stage using `dvc repro` it takes a long time to execute that command too.

Our use-case is that we generate all the stages with `--no-exec` param before we run it for real. If we somehow change the pipeline we would rerun the "generation of all stages with `--no-exec` param" and then again run it for real after it's refreshed.

Would it makes sense to set the `--no-run-cache` to `True` when `--no-exec` is specified? Or am I overlooking something?

Thank you very much !

### Reproduce

Example:

1. Create a large folder, ideally with lots of images
2. `dvc run --no-exec --force --deps folder_path -n stage` # This should be done in instance
3. `dvc repro stage` # This should take some time to compute checksum for dependencies
4. `dvc run --no-exec --force --deps folder_path -n stage` # This should be done in instance (but it actually will take some time)

### Expected

I would expect that steps 2 and 4 would be completed aproximately in the same amount of time.

### Environment information

**Output of `dvc version`:**

```console
DVC version: 1.11.13 (pip)
---------------------------------
Platform: Python 3.6.8 on Linux-4.18.0-259.el8.x86_64-x86_64-with-centos-8
Supports: http, https, s3
Cache types: reflink, hardlink, symlink
Cache directory: xfs on *********
Caches: local
Remotes: s3
Workspace directory: xfs on *********
Repo: dvc, git
```

efiop:
Hi @Honzys !

I'm not able to reproduce the issue with 1.11.13 or newer dvc. I see that the commands you've provided are not the actual commands that you've used (e.g. there is no command in the `dvc run`), so probably something important is missing here. Could you provide actual commands or provide a reproducer script that actually runs, please?

Honzys:
@efiop Thank you for your response.

I am sorry, I wrote the reproducer as a "pseudocode" just to show the basic idea.

Unfortunately I cannot create exact reproducer to this issue (since I cannot provide you the dependency folder). But I can show you some other info.

[...]

Issue #3362, opened 2020-02-19
Previous import stage disappears on import failure


skshetry:
```sh
$ dvc import /path/to/existing/repo file
$ ls
file file.dvc
$ dvc import /path/to/not-existing/repo file
$ test -f file.dvc || echo $? # no files or import stages
1
```

Same thing happens when the repo is valid but the file does not exist.

efiop:
@skshetry This issue affects any dvc command such as `dvc add/run/import/import-url`. The reason is that when you run the same command again, you are explicitly overwriting the dvc-file and if the operation fails we don't want to give you a false sense of successful operation. E.g. if you would run `dvc checkout` after that you won't run the risk of getting an old file and not even realising that.

skshetry:
@efiop, I think the behaviour should be more robust. Right now, I guess this is a limitation of the way we collect stages or build graph ( that could be solved by `update/replace` methods in `repo.index` with a `trie`?)

skshetry:
These days we fail with:
```console
unexpected error - [Errno 17] File exists: 'file'
```

For `--no-download` case, we do end up overwriting as the file won't exist in the workspace which is preventing this from happening. It is not ideal, but closing for now, as the above mentioned issue is somewhat fixed.

PR #5385, opened 2021-02-02
dvc: get rid of can_be_skipped logic

efiop:
This is a legacy logic that we used to have before run-cache. At this point it is not needed and is broken in some scenarios.

Part of https://github.com/iterative/dvc/issues/4841
Related to #5368
Fixes https://github.com/iterative/dvc/issues/4529

### DIFF
--- a/dvc/repo/add.py
+++ b/dvc/repo/add.py
@@ -255,12 +255,9 @@ def _create_stages(
             external=external,
         )
         restore_meta(stage)
-        if stage.can_be_skipped:
-            stage = None
-        else:
-            Dvcfile(repo, stage.path).remove()
-            if desc:
-                stage.outs[0].desc = desc
+        Dvcfile(repo, stage.path).remove()
+        if desc:
+            stage.outs[0].desc = desc
 
         repo._reset()  # pylint: disable=protected-access
 
--- a/dvc/repo/imp_url.py
+++ b/dvc/repo/imp_url.py
@@ -57,8 +57,6 @@ def imp_url(
         erepo=erepo,
     )
     restore_meta(stage)
-    if stage.can_be_skipped:
-        return None
 
     if desc:
         stage.outs[0].desc = desc
--- a/dvc/repo/run.py
+++ b/dvc/repo/run.py
@@ -22,8 +22,6 @@ def run(
     from dvc.stage.utils import validate_state
 
     stage = self.stage.create_from_cli(**kwargs)
-    if run_cache and stage.can_be_skipped:
-        return None
 
     validate_state(self, stage, force=force)
 
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -32,7 +32,6 @@ from .utils import (
     fill_stage_dependencies,
     fill_stage_outputs,
     get_dump,
-    stage_dump_eq,
 )
 
 if TYPE_CHECKING:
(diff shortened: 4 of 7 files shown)


dvc/repo/add.py
```diff
@@ -255,12 +255,9 @@ def _create_stages(
             external=external,
         )
         restore_meta(stage)
-        if stage.can_be_skipped:
-            stage = None
-        else:
-            Dvcfile(repo, stage.path).remove()
-            if desc:
-                stage.outs[0].desc = desc
```
skshetry, line 263: hmm, I find this helpful though.

dvc/repo/add.py
```diff
@@ -255,12 +255,9 @@ def _create_stages(
             external=external,
         )
         restore_meta(stage)
-        if stage.can_be_skipped:
-            stage = None
```
skshetry, line 259: hmm, I find this helpful though.

efiop: @skshetry But it was only used for this semi-run-cache, while it wasn't the behavior in real run-cache. So inconsistent and confusing :slightly_frowning_face: Thinking if maybe we could introduce some flag that will signal that the stage was actually ran. E.g. `if stage.ran: ...` or something.

efiop: Or `stage.restored`. Seems to be more useful than returning `None`, as you have more information and more control.

skshetry: but, for `dvc add`-ed files, it would use `changed_outs()` result to determine this, no?

efiop: @skshetry You mean `ran/restored`? Those are close to meaningless for `dvc add`, as they are more of a pipeline thing. We were calling `commit` in `is_cached` anyway, so without it we'll just verify that outputs and cache are in place. If anyone needs to know whether or not there have been any workspace changes, they could call `status` before running `dvc add`.

skshetry: The `stage = None` is definitely confusing, there's even a comment for that:

https://github.com/iterative/dvc/blob/9ad1e0e15925ec3ca0d741d64bdf65d293a83f58/dvc/repo/trie.py#L10

But I find this hack simpler than proper solution. The `dvc add` has been kind of broken for `--file` for long, but this might break for non `--file` case as well, by removing `.dvc` file.

Consider the following scenario:
```console
$ echo foo > foo
$ echo bar > bar
$ dvc add foo bar
$ ls
bar bar.dvc foo foo.dvc 
$ rm foo
$ echo foobar > bar
$ dvc add foo bar
$ ls  # no foo.dvc file
bar bar.dvc
```
```

Comment (efiop) [reply dvc/repo/add.py:259]: Oh, good point! Need to remove that filtering too.

Comment (skshetry) [reply dvc/repo/add.py:259]: I meant the following:
```console
$ dvc add foo
Stage is cached, skipping                                                                        
  0% Add|                                                               |0/0 [00:00,     ?file/s]
```
vs the current PR:
```console
$ dvc add foo
100% Add|███████████████████████████████████████████████████████████████|1/1 [00:00,  4.34file/s]

	git add foo.dvc
```

dvc/repo/add.py
```diff
@@ -255,12 +255,9 @@ def _create_stages(
...
-        else:
-            Dvcfile(repo, stage.path).remove()
-            if desc:
-                stage.outs[0].desc = desc
+        Dvcfile(repo, stage.path).remove()
+        if desc:
+            stage.outs[0].desc = desc
 
         repo._reset()  # pylint: disable=protected-access
```
skshetry, line 262: Unrelated to this PR, we can check the graph's correctness first without doing other stuff and fail. I don't think we need to keep resetting it.

dvc/repo/add.py
efiop: @skshetry Previous one was confusing https://github.com/iterative/dvc/issues/4529 .

The current one is related to the way we have the progress bar in `add()`. Definitely could consider running the targets through `dvc status` first to filter-out unchanged ones and not reflect them in the progress bar. Or to somehow pass the pbar into the saving logic and let it handle the progress. To me 0% seem confusing, and the current PR behavior makes more sense, as it is a corner case of you already having the data in the cache (e.g. some collegue already added it to the sshared cache, or you've already added it in some other branch).

skshetry (approved):
Looks good to me. 👍🏽

dvc/repo/add.py
skshetry: My concern was with "is cached" part, which I find helpful, and there's no need for a hint in that case as well. But, fine for now. 
I'll create a separate issue with `dvcfile.remove()` which I find unsafe here.

skshetry: That issue already exists 😄  https://github.com/iterative/dvc/issues/3362

PR #5397, opened 2021-02-03
stage: implement add/create methods in stage loader

skshetry:

### DIFF
--- a/dvc/repo/stage.py
+++ b/dvc/repo/stage.py
@@ -3,7 +3,7 @@ import logging
 import os
 import typing
 from contextlib import suppress
-from functools import partial
+from functools import partial, wraps
 from typing import (
     Callable,
     Iterable,
@@ -21,6 +21,7 @@ from dvc.exceptions import (
     OutputNotFoundError,
 )
 from dvc.path_info import PathInfo
+from dvc.repo import lock_repo
 from dvc.utils import parse_target, relpath
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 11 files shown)


dvc/repo/__init__.py
```diff
@@ -482,6 +463,25 @@ def open_by_relpath(self, path, remote=None, mode="r", encoding=None):
     def close(self):
         self.scm.close()
 
+    @contextmanager
+    def lock_repo(self):
```
pared, line 467: Usage in tests indicates that it might be used via `Repo.lock_repo(repo)`, maybe we should not use `self` here, but `repo` instead? It would look similar to what we are doing in many `dvc.repo.*` packages and also indicate that one not necessarily needs an object to call it. I don't have a strong opinion on that.

pared (approved):

dvc/repo/__init__.py
skshetry: @pared, that's just an easy way to test a `contextmanager`. You can just use it as:
```python:
with repo.lock_repo():
  pass
```

pared: I guess I was too fast with that comment, test led me to believe that there might be the use case where you lock repo twice, which will probably not be a prevalent use case, now that I think about that.

skshetry: > where you lock repo twice

@pared, this was meant to be used in the Viewer to prevent calling `_reset()` too many times when calling Repo's top-level APIs.

cc @efiop let me know if we should maintain compatibility for the Viewer.
It made sense to me to move it inside `Repo` as  I see it as a part of the `Repo` itself and that it would get into cyclic import issue. But, looks like it won't get into cyclic import issue.

efiop: @skshetry Is this change needed for this PR?

dvc/repo/stage.py
```diff
@@ -99,24 +99,100 @@ def _collect_specific_target(
...
+        from dvc.stage import PipelineStage, Stage, create_stage, restore_meta
+        from dvc.stage.exceptions import InvalidStageName
+        from dvc.stage.utils import (
+            is_valid_name,
+            prepare_file_path,
+            prepare_kwargs_from_cli,
+            validate_state,
+        )
+
+        if from_cli:
```
efiop, line 173: I might be missing something here,  but this `from_cli` logic is pretty weird, leaking cli into this lowlevel stage creation. It only signals that some of the arguments might be in string format, right? Or is there some additional logic/behavior that we need to apply?

efiop: Feels like this should either not have cli in mind (e.g. from the API standpoint, the user might try to pass strings as arguments and we need to deal with them here) or the cli itself should properly prepare all the arguments before passing to `stage.add/create()` API

skshetry: cli details already leak into `dvc.run()` as they are 1:1 mapping with `dvc run` command. The `stage.add()` and `stage.create()` are meant to be high level API similar to `dvc run`. 

So, to remove this logic from here, either we would have to consider `dvc run` as an exception for now, and duplicate this logic on cli for stage:add and the `dvc.run()`, or bring it here.

The most important one is the parsing of `params` that we need to do inside `dvc run`, for which if we choose to change API, we could be able to move to `cli` logic.

skshetry: To clarify, I'd like to use the `params` as dict/list in API/Tests as well in addition to the cli str format.

dvc/repo/__init__.py
skshetry: > Is this change needed for this PR?

No, I just find it easier it being in `Repo`, easier to test/mock, and no need to import it at all. And, it depends too much on the private variables anyway. Rolled back for now.

dvc/repo/stage.py
skshetry: I am merging this for now. Let's discuss it (or, should I open a new issue?), and I'll update it swiftly. Thanks.

skshetry: nvmd, moved parsing logic outside.

efiop: Sorry, forgot to react. Makes sense in the current circumstances. :+1:

PR #5403, opened 2021-02-03
tree: stop using self.path_info outside of __init__

efiop:
The fact that our trees have `self.path_info` is really confusing and some cloud implementations sometimes accidentally use it instead of the explicitly provided `path_info`. As an endgoal, we should completely separate `self.path_info` and the tree (aka filesystem) itself. So the tree should accept itsown options during initialization instead of the whole config section with the url.

We've been doing things this way because long ago we only had remotes that acted as both remotes, caches and trees at the same time. `self.path_info` is used as a `cache` location, so it has nothing to do with the tree(aka filesystem).

We are already following that isolation nicely in the most places, but this PR tries to cleanup a few leftovers outside `__init__`, so we could get rid of `self.path_info` in the upcoming PRs.

This also makes us 1 step closer to complying with fsspec.

### DIFF
--- a/dvc/tree/gdrive.py
+++ b/dvc/tree/gdrive.py
@@ -113,6 +113,7 @@ class GDriveTree(BaseTree):
             )
 
         self._bucket = self.path_info.bucket
+        self._path = self.path_info.path
         self._trash_only = config.get("gdrive_trash_only")
         self._use_service_account = config.get("gdrive_use_service_account")
         self._service_account_email = config.get(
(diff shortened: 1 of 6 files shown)


dvc/tree/gdrive.py
```diff
@@ -113,6 +113,7 @@ def __init__(self, repo, config):
             )
 
         self._bucket = self.path_info.bucket
+        self._path = self.path_info.path
```
efiop, line 116: This one doesn't belong here, but gdrive is the only implementation that is trying to use it. We've ran into it during `import-url` work, will send a separate PR for that.

dvc/tree/http.py
```diff
@@ -56,20 +55,16 @@ def __init__(self, repo, config):
         self.ssl_verify = config.get("ssl_verify", True)
         self.method = config.get("method", "POST")
 
-    def _auth_method(self, path_info=None):
+    def _auth_method(self):
         from requests.auth import HTTPBasicAuth, HTTPDigestAuth
 
-        if path_info is None:
-            path_info = self.path_info
```
efiop, line 63: _auth_method was never called with explicit path_info, so consider this removing unused code :smile:

PR #5411, opened 2021-02-04
refactor webdav to use client; extract write_text to base remote fixture

skshetry:
- Moves `write_text` to `Base` to reduce duplications
- Move `Webdav` fixture to use `webdav` client rather than creating files/directories in the workspace.
  This was done before to speed up tests, but we found out that they never worked before as they were never used.

From https://github.com/iterative/dvc/pull/5275#discussion_r569483334


efiop (approved):
Nice one!

PR #5412, opened 2021-02-04
cache: generalize save and checkout

efiop:
Pre-requisites for merging save and transfer as well as separating trees and their path_info's, and for the upcoming ODB.

### DIFF
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -115,7 +115,7 @@ class CloudCache:
 
         return self._get_dir_info_hash(filtered_dir_info)[0]
 
-    def changed(self, path_info, hash_info, filter_info=None):
+    def changed(self, path_info, tree, hash_info, filter_info=None):
         """Checks if data has changed.
 
         A file is considered changed if:
@@ -138,7 +138,7 @@ class CloudCache:
             "checking if '%s'('%s') has changed.", path_info, hash_info
         )
 
-        if not self.tree.exists(path):
+        if not tree.exists(path):
             logger.debug("'%s' doesn't exist.", path)
             return True
 
@@ -151,7 +151,7 @@ class CloudCache:
             logger.debug("cache for '%s'('%s') has changed.", path, hi)
             return True
 
-        actual = self.tree.get_hash(path)
+        actual = tree.get_hash(path)
         if hi != actual:
             logger.debug(
                 "hash value '%s' for '%s' has changed (actual '%s').",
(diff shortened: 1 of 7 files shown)


dvc/cache/base.py
```diff
@@ -138,7 +138,7 @@ def changed(self, path_info, hash_info, filter_info=None):
             "checking if '%s'('%s') has changed.", path_info, hash_info
         )
 
-        if not self.tree.exists(path):
```
efiop, line 141: One of the sources of dvcignore-related problems, as `self.tree`(cache tree) doesn't use it.

dvc/cache/base.py
```diff
@@ -218,40 +218,22 @@ def _do_link(self, from_info, to_info, link_method):
...
 
         cache_info = self.tree.hash_to_path_info(hash_info.value)
-        if tree == self.tree:
-            if self.changed_cache(hash_info):
+        if self.changed_cache(hash_info):
+            with tree.open(path_info, mode="rb") as fobj:
+                # using our makedirs to create dirs with proper permissions
                 self.makedirs(cache_info.parent)
-                self.move(path_info, cache_info)
+                self.tree.upload_fobj(fobj, cache_info)
```
efiop, line 229: Compared to the previous implementation, we don't move things to cache but rather copy them. This might be a problem in case you want to add a file that can only fit once in your filesystem, but that's rather unusual and you have `--to-remote` now to help you with that. A giant PRO here is that it is much easier to recover from `dvc add` failures (no space, permission problems, etc), as the file/dir is lying right there, intact. Keeping old behavior is not a problem and fits into the arch, so I'll bring that back for now, but will start a discussion elsewhere.

efiop: On the other hand we were about to look into recovering from such errors and this is a good and intuitive way to do it. We've received multiple reports over this year of dvc badly handling initial `dvc add` when there are problems with space or permissions and that's especially bad with directories, where we might be stuck in a limbo of half the files being in cache and half being in the workspace, recovering from such things is a nightmare. Keeping the safer new behavior after all.

shcheklein: sorry not much context on my end, just reading your comments. it means we'll hit some performance issues as well, right? E.g. will it do a full copy even on a FS that supports reflinks?

efiop: Correct. 1 copy when adding the file, that's it.

shcheklein: so, the process itself will be way way longer on these modern file systems. Not sure this is good to be honest.

shcheklein: ... or any file system where people set symlinks/hardlinks I guess

efiop: Agreed, it would be better to be able to properly recover. I'm keeping it in this PR for now(in WIP), will tackle transfer now and will then bring the optimization back.

efiop:
Tried to combine this with `transfer`, but our State got in the way again, as it is not really meant for multithreading(we were noticing some race conditions in trees too because of it). Might be a time to use some ORM to handle that for us.

dvc/cache/base.py
efiop: Ok, checked the next steps with `transfer`, there will be a bit more steps required to make it merge nicely. So brought back the `move` optimization for now. So the behavior is unchanged.

dvc/cache/local.py
```diff
@@ -81,10 +81,10 @@ def hashes_exist(
             )
         ]
 
-    def already_cached(self, path_info):
+    def already_cached(self, path_info, tree):
         assert path_info.scheme in ["", "local"]
 
-        return super().already_cached(path_info)
+        return super().already_cached(path_info, tree)
```
efiop, line 87: Will be handled in the followup related to checkout.

PR #5414, opened 2021-02-04
pygit2, gitpython: handle relpaths in reset/checkout_index backends

rogermparent:
While working on vscode-dvc, both @mattseddon and I found that this change allows
us to run `dvc exp run`, and we cannot run it with dvc master otherwise.

Neither of us know exactly what the other effects of this change are, and I'm hoping 
someone more knowledgeable can find out- after which we can also probably give 
this PR a better name.

### DIFF
--- a/dvc/scm/git/backend/pygit2.py
+++ b/dvc/scm/git/backend/pygit2.py
@@ -183,7 +183,10 @@ class Pygit2Backend(BaseGitBackend):  # pylint:disable=abstract-method
         raise NotImplementedError
 
     def is_ignored(self, path: str) -> bool:
-        return self.repo.path_is_ignored(relpath(path, self.root_dir))
+        rel = relpath(path, self.root_dir)
+        if os.name == "nt":
+            rel.replace("\\", "/")
+        return self.repo.path_is_ignored(rel)
 
     def set_ref(
         self,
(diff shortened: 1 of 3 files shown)


efiop:
For the record: this happens when dvc root and git root don't match. Particularly https://github.com/iterative/vscode-dvc/tree/master/demo .

efiop:
Also, looks like `gs` tests started failing for unrelated reasons. Looking into it...

dvc/scm/git/backend/pygit2.py
```diff
@@ -287,7 +287,7 @@ def reset(self, hard: bool = False, paths: Iterable[str] = None):
             tree = self.repo.revparse_single("HEAD").tree
             for path in paths:
                 rel = relpath(path, self.root_dir)
-                obj = tree[relpath(rel, self.root_dir)]
```
efiop, line 290: @pmrowla Looks like this was a typo?

pmrowla:
Thanks for catching this issue! Looks like we also have a similar bug in the gitpython backend, so I fixed it there as well and updated our unit tests to catch this case.

efiop:
For the record: windows tests are failing on pygit: https://github.com/iterative/dvc/pull/5414/checks?check_run_id=1835638237

PR #5421, opened 2021-02-06
cache: use add() to add objects to cache

efiop:
This makes all cache objects go through the same logic, so that all of
the appropriate rules are applied to them (e.g. proper permissions).

In the end our save/transfer logic should be outside of cache (same as
checkout) and should build appropriate objects and then just save them
into cache (aka odb) with `cache.add()`(aka `odb.add_object()`).

Pre-requisite for #5337

### DIFF
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -141,25 +141,30 @@ class CloudCache:
             "Created '%s': %s -> %s", self.cache_types[0], from_info, to_info,
         )
 
+    def add(self, path_info, tree, hash_info, **kwargs):
+        if not self.changed_cache_file(hash_info):
+            return
+
+        cache_info = self.tree.hash_to_path_info(hash_info.value)
+        # using our makedirs to create dirs with proper permissions
+        self.makedirs(cache_info.parent)
+        if isinstance(tree, type(self.tree)):
+            self.tree.move(path_info, cache_info)
+        else:
+            with tree.open(path_info, mode="rb") as fobj:
+                self.tree.upload_fobj(fobj, cache_info)
+            tree.state.save(path_info, hash_info)
+        self.protect(cache_info)
+        self.tree.state.save(cache_info, hash_info)
+
+        callback = kwargs.get("download_callback")
+        if callback:
+            callback(1)
+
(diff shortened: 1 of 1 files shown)


dvc/cache/base.py
```diff
@@ -141,25 +141,30 @@ def _do_link(self, from_info, to_info, link_method):
...
+                self.tree.upload_fobj(fobj, cache_info)
+            tree.state.save(path_info, hash_info)
+        self.protect(cache_info)
+        self.tree.state.save(cache_info, hash_info)
+
     def _save_file(self, path_info, tree, hash_info, **kwargs):
         assert hash_info
 
-        cache_info = self.tree.hash_to_path_info(hash_info.value)
-        if self.changed_cache(hash_info):
```
efiop, line 148: This `cached_cache` was costing us one `isdir()` call :man_facepalming: :man_facepalming: :man_facepalming: Should've used `changed_cache_file` all along.

dvc/cache/base.py
```diff
@@ -141,25 +141,30 @@ def _do_link(self, from_info, to_info, link_method):
             "Created '%s': %s -> %s", self.cache_types[0], from_info, to_info,
         )
 
+    def add(self, path_info, tree, hash_info):
```
efiop, line 144: This should really be:

```suggestion
    def add(self, object):
```

but that will come later (e.g. an object might be a plain file, a `.dir` or a run cache).

dvc/cache/base.py
```diff
@@ -215,9 +219,13 @@ def _transfer_directory(
...
 
         with Tqdm(total=1, unit="Files", disable=no_progress_bar) as pbar:
-            for entry_info, entry_hash in self._transfer_directory_contents(
+            for (
+                entry_info,
+                (entry_tmp_info, entry_hash),
+            ) in self._transfer_directory_contents(
                 from_tree, from_info, jobs, pbar
             ):
+                self.add(entry_tmp_info, self.tree, entry_hash)
```
efiop, line 228: Now we are checking if the file is already in cache before doing a final move, so need to not forget to clean it up.

PR #5423, opened 2021-02-06
utils: don't use unnecessary exists() calls

efiop:

### DIFF
--- a/dvc/checkout.py
+++ b/dvc/checkout.py
@@ -64,7 +64,10 @@ def _changed(path_info, tree, hash_info, cache, filter_info=None):
         logger.debug("cache for '%s'('%s') has changed.", path, hi)
         return True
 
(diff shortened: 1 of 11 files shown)


dvc/checkout.py
```diff
@@ -64,7 +60,12 @@ def _changed(path_info, tree, hash_info, cache, filter_info=None):
         logger.debug("cache for '%s'('%s') has changed.", path, hi)
         return True
 
-    actual = tree.get_hash(path)
+    try:
+        actual = tree.get_hash(path)
```
efiop, line 64: Some clouds don't properly wrap exceptions into OSErrors, but the followup PR will move `get_hash` out of trees into a separate helper and those clouds will use stat() that uses ls() that handles exceptions properly.

PR #5424, opened 2021-02-08
dvc: introduce File/Tree cache objects

efiop:
Pre-requisite for #5337 and #829

### DIFF
--- a/dvc/cache/base.py
+++ b/dvc/cache/base.py
@@ -1,8 +1,5 @@
-import errno
 import itertools
-import json
 import logging
-import os
 from concurrent import futures
 from concurrent.futures import ThreadPoolExecutor
 from copy import copy
@@ -18,16 +15,9 @@ from dvc.remote.slow_link_detection import (  # type: ignore[attr-defined]
     slow_link_guard,
 )
 
-from ..tree.base import RemoteActionNotImplemented
+from ..objects import File, HashFile, ObjectFormatError, Tree
 
 logger = logging.getLogger(__name__)
-
-
-class DirCacheError(DvcException):
-    def __init__(self, hash_):
-        super().__init__(
-            f"Failed to load dir cache for hash value: '{hash_}'."
-        )
 
 
 @decorator
@@ -52,40 +42,6 @@ class CloudCache:
             self.DEFAULT_CACHE_TYPES
         )
         self.cache_type_confirmed = False
-        self._dir_info = {}
-
-    def get_dir_cache(self, hash_info):
-        assert hash_info
-
-        dir_info = self._dir_info.get(hash_info.value)
-        if dir_info:
-            return dir_info
-
-        try:
-            dir_info = self.load_dir_cache(hash_info)
-        except DirCacheError:
-            dir_info = DirInfo()
-
-        self._dir_info[hash_info.value] = dir_info
-        return dir_info
-
-    def load_dir_cache(self, hash_info):
-        path_info = self.tree.hash_to_path_info(hash_info.value)
-
-        try:
-            with self.tree.open(path_info, "r") as fobj:
-                d = json.load(fobj)
-        except (ValueError, FileNotFoundError) as exc:
-            raise DirCacheError(hash_info) from exc
-
-        if not isinstance(d, list):
-            logger.error(
-                "dir cache file format error '%s' [skipping the file]",
-                path_info,
-            )
-            d = []
-
-        return DirInfo.from_list(d)
 
     def link(self, from_info, to_info):
         self._link(from_info, to_info, self.cache_types)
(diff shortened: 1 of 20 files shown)


dvc/objects.py
```diff
@@ -0,0 +1,142 @@
...
+
+    @classmethod
+    def load(cls, odb, hash_info):
+        return odb.get(hash_info)
+
+    def commit(self, odb, **kwargs):
+        odb.add(self, **kwargs)
+
+
+class File(HashFile):
```
efiop, line 53: For 3.0 we'll be able to introduce smth like `Blob` that will do the chunking like git does

dvc/objects.py
```diff
@@ -0,0 +1,142 @@
...
+        super().__init__(*args, **kwargs)
+        self.src = None
+
+    @classmethod
+    def stage(cls, odb, path_info, tree, **kwargs):
+        hash_info = tree.get_hash(path_info, **kwargs)
+        raw = odb.get(hash_info)
+        obj = cls(raw.path_info, raw.tree, hash_info)
+        obj.src = HashFile(path_info, tree, hash_info)
+        return obj
```
efiop, line 64: Git staging would've read the object into memory, but we obviously can't do that, so we do it the lazy way: we memorize the original file location (along with it's filesystem aka tree) so that we can later save it.

dvc/objects.py
```diff
@@ -0,0 +1,142 @@
...
+
+class ObjectError(DvcException):
+    pass
+
+
+class ObjectFormatError(ObjectError):
+    pass
+
+
+class HashFile:
```
efiop, line 16: A lot of this is inspired by dulwich. E.g. this is a generalized version of ShaFile https://github.com/dulwich/dulwich/blob/35fdccdd4eff2c95137b3d32c0320b805d8f3320/dulwich/objects.py#L256 , since we support many types of hashes and we also support virtual filesystems.

dvc/objects.py
```diff
@@ -0,0 +1,142 @@
...
+        raw = odb.get(hash_info)
+        obj = cls(raw.path_info, raw.tree, hash_info)
+        obj.src = HashFile(path_info, tree, hash_info)
+        return obj
+
+    def commit(self, odb, **kwargs):
+        self.src.commit(odb, **kwargs)
+
+
+class Tree(HashFile):
```
efiop, line 70: This is the current Tree (don't confuse with our dvc/tree tree's, those will become FileSystem's soon), and it will be possible to tweak it to start saving empty dirs or change the format of how it is stored in cache.

dvc/objects.py
```diff
@@ -0,0 +1,142 @@
...
+        for (
+            key,
+            hash_info,
+        ) in self.hash_info.dir_info.trie.iteritems(  # noqa: B301
+            **kwargs
+        ):
+            yield (key, hash_info)
+
+    @classmethod
+    def stage(cls, odb, path_info, tree, **kwargs):
```
efiop, line 90: Similar to File, we are using a lazy approach for files, but actually save the '.dir' cache file, same as it was before.

dvc/output/base.py
```diff
@@ -389,24 +404,30 @@ def checkout(
                 )
             return None
 
+        obj = self.get_obj(filter_info=filter_info)
+        if not obj and (filter_info and filter_info != self.path_info):
+            # backward compatibility
```
efiop, line 409: Currently, if you try to checkout a nonexistent directory/path in a cached directory, we'll just silently swallow it :( That nees to be fixed, but I didn't to minimise the func test changes.

dvc/output/base.py
```diff
@@ -389,24 +404,30 @@ def checkout(
...
+            return None
+
+        added = not self.exists
```
efiop, line 412: this is what checkout() was doing, but now in an explicit way, since we care about status with output-level granularity, not finer

dvc/output/base.py
```diff
@@ -431,29 +452,49 @@ def move(self, out):
             self.repo.scm.ignore(self.fspath)
 
     def get_files_number(self, filter_info=None):
```
efiop, line 454: get_files_number/get_dir_cache are needed mostly for push/pull and really shouldn't be in the Output, as those things should be figured out on the odb/object level.

dvc/objects.py
```diff
@@ -0,0 +1,142 @@
...
+
+class ObjectFormatError(ObjectError):
+    pass
+
+
+class HashFile:
+    def __init__(self, path_info, tree, hash_info):
+        self.path_info = path_info
+        self.tree = tree
+        self.hash_info = hash_info
```
efiop, line 20: hash_info for us is like git's object id (aka oid, aka object sha), since, again, we support multiple hashes.

dvc/objects.py
```diff
@@ -0,0 +1,142 @@
...
+        return 1
+
+    def __str__(self):
+        return f"object {self.hash_info}"
+
+    def __bool__(self):
+        return bool(self.hash_info)
+
+    def check(self):
+        actual = self.tree.get_hash(self.path_info)
```
efiop, line 32: this really should be something like
```suggestion
        actual = self.tree.get_hash(self.path_info, hash_type=odb.hash_type)
```
but that will be introduced later.

dvc/cache/base.py
```diff
@@ -275,89 +230,9 @@ def cache_is_copy(self, path_info):
...
-
-    def _save(self, path_info, tree, hash_info, **kwargs):
-        to_info = self.tree.hash_to_path_info(hash_info.value)
-        logger.debug("Saving '%s' to '%s'.", path_info, to_info)
-
-        if tree.isdir(path_info):
-            self._save_dir(path_info, tree, hash_info, **kwargs)
-        else:
-            self._save_file(path_info, tree, hash_info, **kwargs)
+    def save(self, obj, **kwargs):
```
efiop, line 234: Needs some state adjustments and we'll get rid of it.

dvc/cache/base.py
```diff
@@ -141,31 +97,39 @@ def _do_link(self, from_info, to_info, link_method):
             "Created '%s': %s -> %s", self.cache_types[0], from_info, to_info,
         )
 
-    def add(self, path_info, tree, hash_info, **kwargs):
-        if not self.changed_cache_file(hash_info):
+    def get(self, hash_info):
+        """ get raw object """
+        return HashFile(
+            # Prefer string path over PathInfo when possible due to performance
```
efiop, line 103: Very soon this will no longer be needed, when we detach trees from their self.path_info's.

efiop:
Ok, moving to state thread changes...

PR #5427, opened 2021-02-08
[WIP] gdrive: .ls()/.walk()/.walk_dir()

### LINKED ISSUE #5410: gdrive: _list_paths/walk_files returns both files and directories
This is a pre-requisite to properly handle `import-url`s and to generalize GdriveTree to the level of other trees so we could work with arbitrary fs tree structures. We should implement fsspec-like `ls()` and use it to implement `walk()` properly. Note that there is `_ids_cache` primitive caching that will definitely get in the way as it makes some strong assumptions about current location, so we might need to generalize it for the whole tree (kinda like fsspec's dircache).

https://github.com/intake/gdrivefs/blob/master/gdrivefs/core.py might be useful for inspiration.

isidentical:

Resolves #5410

### DIFF
--- a/dvc/tree/gdrive.py
+++ b/dvc/tree/gdrive.py
@@ -4,7 +4,7 @@ import os
 import posixpath
 import re
 import threading
-from collections import defaultdict
+from collections import defaultdict, deque
 from contextlib import contextmanager
 from urllib.parse import urlparse
 
@@ -88,6 +88,7 @@ class GDriveURLInfo(CloudURLInfo):
 class GDriveTree(BaseTree):
     scheme = Schemes.GDRIVE
     PATH_CLS = GDriveURLInfo
+    PARAM_CHECKSUM = "checksum"
     REQUIRES = {"pydrive2": "pydrive2"}
     # Always prefer traverse for GDrive since API usage quotas are a concern.
     TRAVERSE_WEIGHT_MULTIPLIER = 1
@@ -477,7 +478,7 @@ class GDriveTree(BaseTree):
             return self._ids_cache["dirs"].get(path, [])
         return []
 
-    def _path_to_item_ids(self, path, create, use_cache):
+    def _path_to_item_ids(self, path, create=False, use_cache=True):
         item_ids = self._get_cached_item_ids(path, use_cache)
         if item_ids:
             return item_ids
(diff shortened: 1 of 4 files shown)


dvc/tree/gdrive.py
```diff
@@ -510,35 +511,68 @@ def exists(self, path_info, use_dvcignore=True):
...
+    ):  # pylint: disable=arguments-differ
+        if recursive:
+            for entry in self.walk_files(path_info, detail=detail):
+                if detail:
+                    yield entry
+                else:
+                    yield str(entry)
+            return None
+
+        parent_ids = self._path_to_item_ids(path_info.path)
```
shcheklein, line 525: minor: a bit confusing name here - they are just ids here, not parent ids as far as I understand

dvc/tree/gdrive.py
```diff
@@ -510,35 +511,68 @@ def exists(self, path_info, use_dvcignore=True):
...
+                    yield {
+                        "type": "file",
+                        "name": item_path,
+                        "size": item["fileSize"],
+                        "checksum": item["md5Checksum"],
+                    }
+            else:
+                yield item_path
+
+        if parent_ids:
```
shcheklein, line 550: since we cache only `min()` next time this call can return a different result?

dvc/tree/gdrive.py
```diff
@@ -510,35 +511,68 @@ def exists(self, path_info, use_dvcignore=True):
...
     def walk_files(self, path_info, **kwargs):
-        use_prefix = kwargs.pop("prefix", False)
-        if path_info == self.path_info or not use_prefix:
-            prefix = None
-        else:
-            prefix = path_info.path
-        for fname in self._list_paths(prefix=prefix, **kwargs):
-            yield path_info.replace(fname)
+        if kwargs.pop("prefix", False):
+            path_info = path_info / ""
```
shcheklein, line 573: hmm, not sure I understand what's going on here ... could you please clarify? is it equivalent? do we change upstream a bit to pass path/prefix in a different way?

dvc/tree/gdrive.py
isidentical: No, not really. I can manually verify that caching works, and also this usage is something that I borrowed from the code above. Though I am not quite sure about why and how (probably the parent_ids is always 1-length list? maybe not but it seems that way)

dvc/tree/gdrive.py
isidentical: I think this should be obsolete now

isidentical:
GDrive tests doesn't work on my repo, so closing this PR in order to get an upstream branch

PR #5428, opened 2021-02-08
Optimaze dvc update

### LINKED ISSUE #4889: Optimize import-url on directories, do not download existing files again
The use case is described here:

https://discord.com/channels/485586884165107732/485596304961962003/773882224248225812
https://discord.com/channels/485586884165107732/485596304961962003/776205388654968892

----------------

I have a bucket with images that is being updated (new images, etc) regularly:

`gs://bucket/dataset/images`

I don't want to use `dvc add`, primarily because I have other projects, websites using this bucket directly.
I still want to use these images in the DVC project, download them, download new, etc (basically do sync operation):

```bash
dvc run -n download_images 
              --outs-persist-no-cache dataset/images 
              "gsutil -m cp -n -r gs://bucket/dataset/images dataset/"
# -m -> multiprocess
# -n -> not downloading existing files (even though i wouldnt care because i pay the access anyway and the traffic is free) but its faster
```

This can be done with `import-url`. This is what `import-url` was made for:

```
dvc import-url gs://bucket/dataset/images
```

The problem is that `import-url` downloads all the images again on every change in the remote directory.

We should somehow utilize existing knowledge about the local cache to avoid downloading existing images and download only changes.

aerubanov:
It's PR close #4889. If dependensy is a directory, we compare has of individual files in dep and out and download/remove file if differences were founded. Comparison is possible if dep and out use the same hash functions. Otherwise the processing is the same as before. Also I added tests.

There are a few things that raise doubts in my mind:
- in ```tests/unit/stage/test_imports.py``` I wrote mocks for Stage and Tree classes. I think it would be better to use something that already exists instead, but I haven't found anything that is suitable. Maybe I just missed something.
- Do I need to add any changes to the documentation?

I would be grateful for any advice on further improvement.

### DIFF
--- a/tests/unit/stage/test_imports.py
+++ b/tests/unit/stage/test_imports.py
@@ -1,67 +1,40 @@
-from collections import namedtuple
-
 from dvc.path_info import PathInfo
 from dvc.stage.imports import get_dir_changes, update_import_dir
 
 
-class MockTree:
-    def __init__(self, files):
-        self.files = files
+def test_get_dir_changes(tmp_dir, dvc):
 
-    def get_file_hash(self, file):
-        result = namedtuple("HashInfo", "value")
-        return result(file)
+    imp = tmp_dir.gen({"datadir": {"foo": "foo", "bar": "bar"}})[0]
 
-    def walk_files(self, *args, **kwargs):
-        return iter(self.files)
+    out = tmp_dir / "out"
 
-    def download(self, *args, **kwargs):
-        pass
+    stage = dvc.imp_url(str(imp), str(out))
 
-    def remove(self, *args, **kwargs):
-        pass
+    (tmp_dir / "datadir" / "foo").write_text("test")
+    (tmp_dir / "datadir" / "foo1").write_text("test_1")
 
-
-class MockStage:
-    class TestOut:
-        def __init__(self, path_info=None, tree=None):
-            self.path_info = path_info
-            self.tree = tree
-
-        def update(self, *args, **kwargs):
-            pass
-
-    def __init__(self, out, dep, dep_path=None, out_path=None):
-        self.outs = [self.TestOut(path_info=out_path, tree=out)]
-        self.deps = [self.TestOut(path_info=dep_path, tree=dep)]
-
-    def save_deps(self, *args, **kwargs):
-        pass
-
-
-def test_get_dir_changes():
-    out_tree = MockTree(["file1", "file2", "file3"])
-    dep_tree = MockTree(["file1", "file2", "file4"])
-    stage = MockStage(out_tree, dep_tree)
     files_to_down, files_to_rem = get_dir_changes(stage)
-    assert files_to_down == ["file4"]
-    assert files_to_rem == ["file3"]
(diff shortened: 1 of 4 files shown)


aerubanov:
I broke some tests in ```tests/func/test_update.py```. The problem is caused by the fact that  ```self.deps[0].path_info``` is None in these tests. And I pass this to ```tree.isdir()``` in ```stage.update()``` and it breaks. I'm looking at how to fix this.

aerubanov:
Some tests are still failed.But I think I fixed the ones that were broken from the changes in this PR.

aerubanov:
Updated PR to make it compatible with dvc 2.0.

tests/unit/stage/test_imports.py
```diff
@@ -0,0 +1,58 @@
...
+def test_update_import_dir(tmp_dir, dvc, mocker):
+    imp = tmp_dir.gen({"data": {"file3": "foo", "file4": "bar"}})[0]
+
+    out = tmp_dir / "out"
+
+    stage = dvc.imp_url(str(imp), str(out))
+    save_deps = mocker.patch.object(stage, "save_deps")
+    remove = mocker.patch.object(stage.outs[0].fs, "remove")
+    download = mocker.patch.object(stage.deps[0].fs, "download")
+    with mocker.patch(
```
karajan1001, line 49: A warning here, according to https://pypi.org/project/pytest-mock/

> The purpose of this plugin is to make the use of context managers and function decorators for mocking unnecessary, so it will emit a warning when used as such.

karajan1001:
> There are a few things that raise doubts in my mind:
> 
> * in `tests/unit/stage/test_imports.py` I wrote mocks for Stage and Tree classes. I think it would be better to use something that already exists instead, but I haven't found anything that is suitable. Maybe I just missed something.
> * Do I need to add any changes to the documentation?
> 
> I would be grateful for any advice on further improvement.

Maybe we can merge the three tests into one? We could check:
1. If the already exists files were still downloaded using `mocker`.
2. The final contents of the path.
in one test.

aerubanov:
@karajan1001 , yes, I think it is possible to replace tests by one. I`ll try to implement this.

efiop:
@aerubanov Thanks for the PR! Excuse us for not getting back to you earlier, but I think the tests could wait. I don't think we will be able to merge this PR or quickly adapt it, as it has a lot of fragile assumptions that don't really work in a general case (e.g. md5 and md5dos2unix that we use for local objects are not the same). We might have to close this PR and implement this optimization later, after migrating to a new hash for our objects. :slightly_frowning_face: Just wanted to make it clear to not waste more of your time.

aerubanov:
@efiop, yes, this optimization will work much better if all objects will use the same hash functions. So I close this PR and will add comment in related issue.

efiop:
@aerubanov Sounds great! Really sorry for wasting your effort here, we really appreciate all of the contributions you've made. Just to be clear, the fragility I was talking about was actually coming from the way we dealt with directories before, and that's why we are introducing `dvc/objects/` for the past month, so that we could accommodate such optimizations in a natural and robust way. Thank you so much for the PR! :pray:

PR #5433, opened 2021-02-09
gdrive: Implement .ls() protocol / proper tree walking with .walk_files()

isidentical:

### DIFF
--- a/dvc/tree/gdrive.py
+++ b/dvc/tree/gdrive.py
@@ -4,7 +4,7 @@ import os
 import posixpath
 import re
 import threading
-from collections import defaultdict, deque
+from collections import defaultdict
 from contextlib import contextmanager
 from urllib.parse import urlparse
 
@@ -286,20 +286,21 @@ class GDriveTree(BaseTree):
             ),
         }
 
-        self._cache_path_id(self._path, cache["root_id"], cache)
+        self._cache_path_id(self._path, cache["root_id"], cache=cache)
 
         for item in self._gdrive_list(
             "'{}' in parents and trashed=false".format(cache["root_id"])
         ):
             item_path = (self.path_info / item["title"]).path
-            self._cache_path_id(item_path, item["id"], cache)
+            self._cache_path_id(item_path, item["id"], cache=cache)
 
         return cache
 
-    def _cache_path_id(self, path, item_id, cache=None):
+    def _cache_path_id(self, path, *item_ids, cache=None):
         cache = cache or self._ids_cache
-        cache["dirs"][path].append(item_id)
-        cache["ids"][item_id] = path
+        for item_id in item_ids:
+            cache["dirs"][path].append(item_id)
+            cache["ids"][item_id] = path
 
     @cached_property
     def _list_params(self):
@@ -511,15 +512,55 @@ class GDriveTree(BaseTree):
         else:
             return True
 
+    def _gdrive_list_ids(self, query_ids):
+        query = " or ".join(
+            f"'{query_id}' in parents" for query_id in query_ids
+        )
+        query = f"({query}) and trashed=false"
+        return self._gdrive_list(query)
+
+    def _ls_recursive(self, path_info, detail=False):
+        root_path = path_info.path
+        seen_paths = set()
+
+        dir_ids = [self._ids_cache["ids"].copy()]
+        while dir_ids:
+            query_ids = {
+                dir_id: dir_name
+                for dir_id, dir_name in dir_ids.pop().items()
+                if posixpath.commonpath([root_path, dir_name])
+                if dir_name not in seen_paths
+            }
+            if not query_ids:
+                continue
+
+            seen_paths.update(query_ids.values())
+
+            new_query_ids = {}
+            dir_ids.append(new_query_ids)
+            for item in self._gdrive_list_ids(query_ids):
+                parent_id = item["parents"][0]["id"]
+                item_path = posixpath.join(query_ids[parent_id], item["title"])
+                if item["mimeType"] == FOLDER_MIME_TYPE:
+                    new_query_ids[item["id"]] = item_path
+                    self._cache_path_id(item_path, item["id"])
+                    continue
+
+                if detail:
+                    yield {
+                        "type": "file",
+                        "name": item_path,
+                        "size": item["fileSize"],
+                        "checksum": item["md5Checksum"],
+                    }
+                else:
+                    yield item_path
+
     def ls(
         self, path_info, detail=False, recursive=False
     ):  # pylint: disable=arguments-differ
         if recursive:
-            for entry in self.walk_files(path_info, detail=detail):
-                if detail:
-                    yield entry
-                else:
-                    yield entry.path
+            yield from self._ls_recursive(path_info, detail=detail)
             return None
 
         cached = path_info.path in self._ids_cache["dirs"]
@@ -531,11 +572,8 @@ class GDriveTree(BaseTree):
         if not dir_ids:
             return None
 
-        query = " or ".join(f"'{dir_id}' in parents" for dir_id in dir_ids)
-        query = f"({query}) and trashed=false"
-
         root_path = path_info.path
-        for item in self._gdrive_list(query):
+        for item in self._gdrive_list_ids(dir_ids):
             item_path = posixpath.join(root_path, item["title"])
             if detail:
                 if item["mimeType"] == FOLDER_MIME_TYPE:
@@ -551,31 +589,11 @@ class GDriveTree(BaseTree):
                 yield item_path
 
         if not cached:
-            self._cache_path_id(root_path, min(dir_ids))
-
-    def walk(
-        self, path_info, prefix=None, detail=False
-    ):  # pylint: disable=unused-argument
-        stack = deque([path_info])
-        while stack:
-            path_info = stack.pop()
-            files, directories = [], []
-            for entry in self.ls(path_info, detail=True):
-                path = path_info.replace(path=entry["name"])
-                if entry["type"] == "file":
-                    if detail:
-                        files.append(entry)
-                    else:
-                        files.append(path)
-                else:
-                    directories.append(path)
-
-            yield path_info, directories, files
-            stack.extendleft(directories)
+            self._cache_path_id(root_path, dir_ids)
 
     def walk_files(self, path_info, **kwargs):
-        for _, _, file_infos in self.walk(path_info, **kwargs):
-            yield from file_infos
+        for filename in self.ls(path_info, recursive=True):
+            yield path_info.replace(path=filename)
 
     def remove(self, path_info):
         item_id = self._get_item_id(path_info)
--- a/dvc/tree/gdrive.py
+++ b/dvc/tree/gdrive.py
@@ -528,13 +528,13 @@ class GDriveTree(BaseTree):
             query_ids = {
                 dir_id: dir_name
                 for dir_id, dir_name in dir_ids.pop().items()
-                if posixpath.commonpath([root_path, dir_name])
-                if dir_name not in seen_paths
+                if posixpath.commonpath([root_path, dir_name]) == root_path
+                if dir_id not in seen_paths
             }
             if not query_ids:
                 continue
 
-            seen_paths.update(query_ids.values())
+            seen_paths |= query_ids.keys()
 
             new_query_ids = {}
             dir_ids.append(new_query_ids)
--- a/dvc/tree/gdrive.py
+++ b/dvc/tree/gdrive.py
@@ -4,7 +4,7 @@ import os
 import posixpath
 import re
 import threading
-from collections import defaultdict
+from collections import defaultdict, deque
 from contextlib import contextmanager
 from urllib.parse import urlparse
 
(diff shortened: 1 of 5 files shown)


dvc/tree/gdrive.py
```diff
@@ -510,35 +511,71 @@ def exists(self, path_info, use_dvcignore=True):
...
 
     def walk_files(self, path_info, **kwargs):
-        use_prefix = kwargs.pop("prefix", False)
-        if path_info == self.path_info or not use_prefix:
-            prefix = None
-        else:
-            prefix = path_info.path
-        for fname in self._list_paths(prefix=prefix, **kwargs):
-            yield path_info.replace(fname)
+        for _, _, file_infos in self.walk(path_info, **kwargs):
```
shcheklein, line 577: do we replace the prefix logic completely everywhere in the `status` optimization logic?

dvc/tree/gdrive.py
```diff
@@ -510,35 +511,71 @@ def exists(self, path_info, use_dvcignore=True):
...
+                        files.append(entry)
+                    else:
+                        files.append(path)
+                else:
+                    directories.append(path)
+
+            yield path_info, directories, files
+            stack.extendleft(directories)
 
     def walk_files(self, path_info, **kwargs):
```
shcheklein, line 576: just for the record and so that we are on the same page ... we are making an extra call here now to list `00`, `01`, ... `ff` directories first, right? Before we were directly taking `dir_ids` from the cache (since we expected cache to contain only storage related ids in it)

dvc/tree/gdrive.py
```diff
@@ -510,35 +511,71 @@ def exists(self, path_info, use_dvcignore=True):
...
+            files, directories = [], []
+            for entry in self.ls(path_info, detail=True):
+                path = path_info.replace(path=entry["name"])
+                if entry["type"] == "file":
+                    if detail:
+                        files.append(entry)
+                    else:
+                        files.append(path)
+                else:
+                    directories.append(path)
```
shcheklein, line 571: should we still append entry if details is true? for the sake of type consistency?

dvc/tree/gdrive.py
```diff
@@ -510,35 +511,71 @@ def exists(self, path_info, use_dvcignore=True):
...
+                        "type": "file",
+                        "name": item_path,
+                        "size": item["fileSize"],
+                        "checksum": item["md5Checksum"],
+                    }
+            else:
+                yield item_path
+
+        if not cached:
+            self._cache_path_id(root_path, min(dir_ids))
```
shcheklein, line 554: going back to this:

you need to run a loop on `dir_ids` and add all of them into cache for the same path (that's the current cache behavior and that's how GD is organized - multiple dirs with the same path), or you could change the interface and accept a list there.

dvc/tree/gdrive.py
isidentical: > we are making an extra call here now to list 00, 01, ... ff directories first, right?

yes. We actually have the ids, so we don't make any extra calls to list the directory names themselves but we make extra calls to list their contents. So if you run `dvc gc -c ` twice and if some directories are empty, previous implementation would result within a single call (since it doesn't actually have to take responsibility of other stuff, and works recursively) but now it makes additional calls for contents of each directory which is a bit slower (but that is how fsspec does it too). CC: @efiop , we had some internal discussion with Ruslan (in case of I am missing something)

dvc/tree/gdrive.py
isidentical: yeap, makes sense.

dvc/tree/gdrive.py
shcheklein: okay, so it's even 1 vs 256+ sequential requests, correct? `gc` is fine, I'm worried a bit that this is also part of the regular `push` flow as far as I remember.

Can it be/should it be part of the `ls(recursive=True)` though?

It's a bit strange btw that ls depends on walk_files and walk_file depends on ls. Feels like `ls` should be self contained. And it can be optimized when `recursive=True` to batch/parallelize requests. Probably it won't be recursive itself then.

isidentical: > Can it be/should it be part of the ls(recursive=True) though?

As far as I can tell, not really. At least not without knowing the whole structure of the tree (not just the initials levels), we can initially go with the first one and then visit the remaining parts though that feels like very cumbersome logic so I am not sure about it. 

I might be wrong, though it seems like there is no change in the push stuff? (have to re-check tomorrow)

shcheklein: > I might be wrong, though it seems like there is no change in the push stuff? (have to re-check tomorrow)

it depends on how `push` decides to do the check for the existing files. It kinda runs status before. (as far as I remember)

> As far as I can tell, not really.

could you elaborate please?

>  At least not without knowing the whole structure of the tree (not just the initials levels), we can initially go with the first one and then visit the remaining parts though that feels like very cumbersome logic so I am not sure about it.

you just accumulate directories (similar to your deque) and batch them when you do a call. Yes, it will require some additional logic to accumulate directories within the ls itself and batch them. Doesn't sound very terrible to me.

(again, discuss with @efiop the priorities here, I'm commenting and reviewing since I have a bit of experience with GD and understand some implications. It might be fine to create tickets for some things. It's great to have a plan at least if we are migrating to the new interface).

isidentical: > you just accumulate directories (similar to your deque) and batch them when you do a call. Yes, it will require some additional logic to accumulate directories within the ls itself and batch them. Doesn't sound very terrible to me.

Sure, that's what I meant. Though as I stated earlier, this would be very local and unfortunately won't be available on other solutions in the future (like gdrivefs). Will try to work on this tomorrow though 👀

isidentical: I have an implementation ready, which should work much efficiently though it is only acceptable if we refactor `.walk_files()` to use `.ls(recursive=True)` instead of `.walk()`. By doing that, we can also remove the `.walk()` method for now since there won't be any use case for it. Would it be a problem to disregard fsspec's way on this @efiop?

efiop: @isidentical Doesn't seem like that would be a problem. If we don't need walk - let's get rid of it :slightly_smiling_face:

dvc/tree/gdrive.py
isidentical: I think it should work effectively for all recursive calls with/without prefix

isidentical:
Timings (on the request of @efiop) looks ~same (beside the normal deviation)

master:
```console
 $ time dvc push dataset1
21 files pushed                                                                                                                     
dvc push dataset1  1,98s user 0,28s system 10% cpu 22,415 total                                                                     

 $ time dvc status -c
Cache and remote 'origin' are in sync.                                                                                              
dvc status -c  0,82s user 0,09s system 34% cpu 2,639 total

 $ rm -rf dataset1 .dvc/cache

 $ time dvc pull dataset1
A       dataset1/                                                                                                                   
1 file added and 20 files fetched                                                                                                   
dvc pull dataset1  1,76s user 0,31s system 26% cpu 7,689 total                                                                      

 $ rm -rf dataset1 .dvc/cache dataset1.dvc

 $ time dvc gc -c -f -a

WARNING: This will remove all cache except items used in the workspace and all git branches of the current repo.
No unused 'local' cache to remove.                                                                                                  
dvc gc -c -f -a  1,37s user 0,10s system 6% cpu 22,645 total                                                                        

 $ time dvc gc -c -f -a                                                               

WARNING: This will remove all cache except items used in the workspace and all git branches of the current repo.
No unused 'local' cache to remove.                                                                                                  
No unused cache to remove from remote.                                                                                              
dvc gc -c -f -a  0,90s user 0,08s system 33% cpu 2,940 total
```

this PR:
```console
 $ time dvc push dataset1
21 files pushed                                                                                                                     
dvc push dataset1  2,70s user 0,40s system 12% cpu 25,019 total                                                                     

 $ time dvc status -c    
Cache and remote 'origin' are in sync.                                                                                              
dvc status -c  0,84s user 0,08s system 35% cpu 2,592 total

 $ rm -rf dataset1 .dvc/cache 

 $ time dvc pull dataset1
A       dataset1/                                                                                                                   
1 file added and 20 files fetched                                                                                                   
dvc pull dataset1  1,61s user 0,29s system 24% cpu 7,898 total                                                                      
(.venv38) (Python 3.8.5+) [  4:42ÖS ]  [ isidentical@desktop:~/temp_dvc3(master✗) ]
 $ rm -rf dataset1 .dvc/cache dataset1.dvc 

 $ time dvc gc -c -f -a
WARNING: This will remove all cache except items used in the workspace and all git branches of the current repo.
No unused 'local' cache to remove.                                                                                                  
dvc gc -c -f -a  1,42s user 0,12s system 6% cpu 22,898 total                                                                        

 $ time dvc gc -c -f -a
WARNING: This will remove all cache except items used in the workspace and all git branches of the current repo.
No unused 'local' cache to remove.                                                                                                  
No unused cache to remove from remote.                                                                                              
dvc gc -c -f -a  0,87s user 0,10s system 30% cpu 3,173 total
```

dvc/tree/gdrive.py
shcheklein: yeah, this question I think is not about this local change, more about the upstream changes:

https://github.com/iterative/dvc/blob/00461bb5dff898f8add7fb2311105bb24294f001/dvc/cache/base.py#L762

do we change this logic (that utilizes prefix I think) to estimate the remote size?

dvc/tree/gdrive.py
```diff
@@ -510,35 +512,88 @@ def exists(self, path_info, use_dvcignore=True):
...
+        root_path = path_info.path
+        seen_paths = set()
+
+        dir_ids = [self._ids_cache["ids"].copy()]
+        while dir_ids:
+            query_ids = {
+                dir_id: dir_name
+                for dir_id, dir_name in dir_ids.pop().items()
+                if posixpath.commonpath([root_path, dir_name])
+                if dir_name not in seen_paths
```
shcheklein, line 532: probably should be by `dir_id` also? since GD allows multiple dirs with the same path

dvc/tree/gdrive.py
```diff
@@ -510,35 +512,88 @@ def exists(self, path_info, use_dvcignore=True):
...
+            }
+            if not query_ids:
+                continue
+
+            seen_paths.update(query_ids.values())
+
+            new_query_ids = {}
+            dir_ids.append(new_query_ids)
+            for item in self._gdrive_list_ids(query_ids):
```
shcheklein, line 541: at some point we probably should check what is the limit here - how many `ids` we can pass at once into the API

dvc/tree/gdrive.py
```diff
@@ -510,35 +512,88 @@ def exists(self, path_info, use_dvcignore=True):
...
+                parent_id = item["parents"][0]["id"]
+                item_path = posixpath.join(query_ids[parent_id], item["title"])
```
shcheklein, line 543: would `query_ids[parent_id] / item["title"]` work?

dvc/tree/gdrive.py
```diff
@@ -510,35 +512,88 @@ def exists(self, path_info, use_dvcignore=True):
...
+    def _ls_recursive(self, path_info, detail=False):
+        root_path = path_info.path
+        seen_paths = set()
+
+        dir_ids = [self._ids_cache["ids"].copy()]
+        while dir_ids:
+            query_ids = {
+                dir_id: dir_name
+                for dir_id, dir_name in dir_ids.pop().items()
+                if posixpath.commonpath([root_path, dir_name])
```
shcheklein, line 531: maybe I'm missing something, but what happens if `dir_name` path is shorter than the `root_path` (e.g. cached from the previous run). They will have the common prefix right, but we'll end up querying a directory outside the `root_path`?

isidentical: Can you give an example? I don't use commonprefix, but rather use commonpath so it should actually work fine if I am not missing something

dvc/tree/gdrive.py
isidentical: nop, contents are all strings

dvc/tree/gdrive.py
isidentical: It seems like the limit is ~600 folders or 30k chars, https://stackoverflow.com/questions/29738020/google-drive-api-limit-on-search-query-parameter

dvc/tree/gdrive.py
isidentical: Makes sense, yes!

dvc/tree/gdrive.py
shcheklein: 👍

dvc/tree/gdrive.py
shcheklein: 👍 no need to address it now probably

dvc/tree/gdrive.py
shcheklein: sure, I was thinking about the case like this:

```
root_path = `a/b/c`
dir_name = `a/b`
```

it will return `a/b`, right? (which seems to be wrong here)

isidentical: Oh, right! Guess we have to add a condition the commonpath is the rootpath.

efiop:
For the record: @isidentical created a very handy test fixture using gdrivefs in https://github.com/iterative/dvc/pull/5427/files , but it needs a pydata token to work. I'll try to find the original creds we are using in CI and will add it to the secrets so we could bring that fixture back (maybe in the followup PR, not a blocker).

efiop:
Hey guys, sorry for suddenly jumping in, merging this as is because it is a pre-requisite for another tree patch. Let's please address the rest in the follow-up PR.

Makes me wonder if we are ready to introduce a dev branch :thinking:

PR #5439, opened 2021-02-10
cache: move unrelated helpers out of CloudCache class

efiop:
Moving almost as--is for now, but both transfer and link could use some refactoring.

Pre-requisite for #5337 and #829

### DIFF
--- a/dvc/cache/local.py
+++ b/dvc/cache/local.py
@@ -55,11 +55,6 @@ class LocalCache(CloudCache):
 
         makedirs(path_info, exist_ok=True, mode=self._dir_mode)
 
-    def link(self, from_info, to_info):
-        super().link(from_info, to_info)
-        if self.cache_types[0] not in ("symlink", "hardlink"):
-            os.chmod(to_info, self._file_mode)
-
     def hash_to_path(self, hash_):
         # NOTE: `self.cache_path` is already normalized so we can simply use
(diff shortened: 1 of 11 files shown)


dvc/cache/local.py
```diff
@@ -55,11 +55,6 @@ def makedirs(self, path_info):
 
         makedirs(path_info, exist_ok=True, mode=self._dir_mode)
 
-    def link(self, from_info, to_info):
-        super().link(from_info, to_info)
-        if self.cache_types[0] not in ("symlink", "hardlink"):
-            os.chmod(to_info, self._file_mode)
```
efiop, line 61: This was only needed to put normal permissions on reflinks (as they initially adopt the original ones), moved to `reflink()`.

dvc/checkout.py
```diff
@@ -61,6 +71,80 @@ def _remove(path_info, tree, cache, force=False):
...
+            logger.debug(
+                "Cache type '%s' is not supported: %s", link_types[0], exc
+            )
+            del link_types[0]
+
+    raise CacheLinkError([to_info])
+
+
+def _link(cache, from_info, to_info):
+    assert cache.tree.isfile(from_info)
```
efiop, line 119: Wasting a call :( As noted in the PR desc, link() could use some refactoring. Keeping as it was before for now, but will get back to this shortly during optimization.

PR #5463, opened 2021-02-15
azure: migrate AzureFileSystem to fsspec

### LINKED ISSUE #5435: azure: QA our implementation
People have been complaining about our azure remotes having other issues with authentication, e.g.

https://github.com/iterative/dvc/issues/4534
https://github.com/iterative/dvc/issues/4635

Azure support is very important for us, but unfortunately, no one in the core team is using it regularly (except for the tests that we have, which are clearly not enough). Need to take a look into it, check different auth options, check performance, etc.

isidentical:

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here. (iterative/dvc.org/pull/2226)

Resolves #5435

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -293,7 +293,7 @@ ignored-classes=optparse.Values,thread._local,_thread._local,Dvcfile
 # (useful for modules/projects where namespaces are manipulated during runtime
 # and thus existing member attributes cannot be deduced by static analysis. It
 # supports qualified module names, as well as Unix pattern matching.
-ignored-modules=
+ignored-modules=azure
 
 # Show a hint with possible names when a member name was not found. The aspect
 # of finding the hint is based on edit distance.
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -183,6 +183,9 @@ SCHEMA = {
                     "connection_string": str,
                     "sas_token": str,
                     "account_name": str,
+                    "tenant_id": str,
+                    "client_id": str,
+                    "client_secret": str,
                     **REMOTE_COMMON,
                 },
                 "oss": {
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -179,7 +179,12 @@ SCHEMA = {
                     "webhdfs_alias": str,
                     **REMOTE_COMMON,
                 },
-                "azure": {"connection_string": str, **REMOTE_COMMON},
+                "azure": {
+                    "connection_string": str,
+                    "sas_token": str,
+                    "account_name": str,
+                    **REMOTE_COMMON,
+                },
                 "oss": {
                     "oss_key_id": str,
                     "oss_key_secret": str,
--- a/dvc/fs/azure.py
+++ b/dvc/fs/azure.py
@@ -63,13 +63,9 @@ class AzureFileSystem(BaseFileSystem):
 
     @wrap_prop(threading.Lock())
     @cached_property
-    def client(self):
-        import adlfs.spec
+    def fs(self):
         from adlfs import AzureBlobFileSystem
 
-        assert self.connection_string or self.account_name
-
-        adlfs.spec.FORWARDED_BLOB_PROPERTIES.append("etag")
         file_system = AzureBlobFileSystem(
             account_name=self.account_name,
             connection_string=self.connection_string,
@@ -89,10 +85,10 @@ class AzureFileSystem(BaseFileSystem):
     def open(
         self, path_info, mode="r", **kwargs
     ):  # pylint: disable=arguments-differ
-        return self.client.open(self._with_bucket(path_info), mode=mode)
+        return self.fs.open(self._with_bucket(path_info), mode=mode)
 
     def exists(self, path_info, use_dvcignore=False):
-        return self.client.exists(self._with_bucket(path_info))
+        return self.fs.exists(self._with_bucket(path_info))
 
     def _strip_bucket(self, entry):
         _, entry = entry.split("/", 1)
@@ -114,7 +110,7 @@ class AzureFileSystem(BaseFileSystem):
     ):  # pylint: disable=arguments-differ
         path = self._with_bucket(path_info)
         if recursive:
-            for root, _, files in self.client.walk(path, detail=detail):
+            for root, _, files in self.fs.walk(path, detail=detail):
                 if detail:
                     files = files.values()
                 yield from self._strip_buckets(files, detail, prefix=root)
@@ -127,10 +123,10 @@ class AzureFileSystem(BaseFileSystem):
             yield path_info.replace(path=file)
 
     def remove(self, path_info):
-        self.client.delete(self._with_bucket(path_info))
+        self.fs.delete(self._with_bucket(path_info))
 
     def info(self, path_info):
-        info = self.client.info(self._with_bucket(path_info)).copy()
+        info = self.fs.info(self._with_bucket(path_info)).copy()
         info["name"] = self._strip_bucket(info["name"])
         return info
 
@@ -150,7 +146,7 @@ class AzureFileSystem(BaseFileSystem):
                 total=total,
                 no_progress_bar=no_progress_bar,
             )
-        self.client.invalidate_cache(self._with_bucket(to_info.parent))
+        self.fs.invalidate_cache(self._with_bucket(to_info.parent))
 
     def _download(
         self, from_info, to_file, name=None, no_progress_bar=False, **pbar_args
--- a/dvc/fs/azure.py
+++ b/dvc/fs/azure.py
@@ -1,7 +1,7 @@
 import logging
 import os
+import shutil
 import threading
-from datetime import datetime, timedelta
 
 from funcy import cached_property, wrap_prop
 
@@ -17,42 +17,37 @@ logger = logging.getLogger(__name__)
 class AzureFileSystem(BaseFileSystem):
     scheme = Schemes.AZURE
     PATH_CLS = CloudURLInfo
-    REQUIRES = {
-        "azure-storage-blob": "azure.storage.blob",
-        "knack": "knack",
-    }
     PARAM_CHECKSUM = "etag"
     DETAIL_FIELDS = frozenset(("etag", "size"))
-
-    COPY_POLL_SECONDS = 5
-    LIST_OBJECT_PAGE_SIZE = 5000
+    REQUIRES = {
+        "adlfs": "adlfs",
+        "knack": "knack",
+    }
 
     def __init__(self, repo, config):
         super().__init__(repo, config)
 
-        url = config.get("url", "azure://")
+        url = config.get("url")
         self.path_info = self.PATH_CLS(url)
         self.bucket = self.path_info.bucket
 
         if not self.bucket:
             container = self._az_config.get("storage", "container_name", None)
-            self.path_info = self.PATH_CLS(f"azure://{container}")
-            self.bucket = self.path_info.bucket
+            url = f"azure://{container}"
 
-        self._conn_str = config.get(
-            "connection_string"
-        ) or self._az_config.get("storage", "connection_string", None)
+        self.path_info = self.PATH_CLS(url)
+        self.bucket = self.path_info.bucket
 
-        self._account_url = None
-        if not self._conn_str:
-            name = self._az_config.get("storage", "account", None)
-            self._account_url = f"https://{name}.blob.core.windows.net"
-
-        self._credential = config.get("sas_token") or self._az_config.get(
-            "storage", "sas_token", None
+        self.connection_string = config.get(
+            "connection_string",
+            self._az_config.get("storage", "connection_string", None),
         )
-        if not self._credential:
-            self._credential = self._az_config.get("storage", "key", None)
+        self.account_name = config.get(
+            "account_name", self._az_config.get("storage", "account", None)
+        )
+        self.sas_token = config.get(
+            "sas_token", self._az_config.get("storage", "sas_token", None)
+        )
 
     @cached_property
     def _az_config(self):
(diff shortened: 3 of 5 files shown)


dvc/fs/azure.py
```diff
@@ -68,158 +63,108 @@ def _az_config(self):
...
+        assert self.connection_string or self.account_name
 
-        sas_token = generate_blob_sas(
-            blob_client.account_name,
-            blob_client.container_name,
-            blob_client.blob_name,
-            account_key=blob_client.credential.account_key,
-            permission=BlobSasPermissions(read=True),
-            expiry=expires_at,
+        adlfs.spec.FORWARDED_BLOB_PROPERTIES.append("etag")
```
isidentical, line 72: Contributed to the upstream, awaiting new release

dvc/fs/azure.py
```diff
@@ -68,158 +63,108 @@ def _az_config(self):
...
-        blob_client = self.blob_service.get_blob_client(
-            path_info.bucket, path_info.path
-        )
-        properties = blob_client.get_blob_properties()
-        return {
-            "type": "file",
-            "size": properties.size,
-            "etag": properties.etag.strip('"'),
-        }
+        self.client.invalidate_cache()
```
isidentical, line 132: The `invalidate_cache` calls here and in the download() are still on my radar, but I believe this might be related to a bug in the upstream (still didn't got the standalone reproducer, but the tests fail without them since the dircache somehow holds the containerless-path but the `_ls_from_cache` function tries to match it with the path that also includes the container).

dvc/fs/azure.py
```diff
@@ -68,158 +63,108 @@ def _az_config(self):
...
-        return etag.strip('"')
-
-    def _generate_download_url(self, path_info, expires=3600):
-        from azure.storage.blob import (  # pylint:disable=no-name-in-module
-            BlobSasPermissions,
-            generate_blob_sas,
-        )
-
-        expires_at = datetime.utcnow() + timedelta(seconds=expires)
+    def client(self):
```
efiop, line 66: Let's name it `fs` for consistency. E.g. we do that in memory.py

dvc/config_schema.py
```diff
@@ -179,7 +179,12 @@ class RelPath(str):
                     "webhdfs_alias": str,
                     **REMOTE_COMMON,
                 },
-                "azure": {"connection_string": str, **REMOTE_COMMON},
+                "azure": {
+                    "connection_string": str,
```
shcheklein, line 183: - we want to support Azure ServicePrincipal credentials on the DVC config level (tenant_id, client_id, client_secret) - should we do this as part of this PR, or keep it purely as a migration one?

dvc/config_schema.py
```diff
@@ -179,7 +179,12 @@ class RelPath(str):
...
+                    "sas_token": str,
+                    "account_name": str,
```
shcheklein, line 185: I see that it's already not only about  the migration, meaning that we need to address principal credentials, ideally other env vars Azure lib supports as we have been discussing in the second Azure PR

setup.py
```diff
@@ -96,7 +96,7 @@ def run(self):
 gs = ["google-cloud-storage==1.19.0"]
 gdrive = ["pydrive2>=1.6.3", "six >= 1.13.0"]
 s3 = ["boto3>=1.9.201"]
-azure = ["azure-storage-blob>=12.0", "knack"]
+azure = ["adlfs>=0.6.2", "knack"]
```
shcheklein, line 99: a thing to check - it brings more dependencies? I remember we had some troubles with some Azure libs .. @efiop should have a better idea :)

dvc/config_schema.py
isidentical: We talked about this with @efiop, I didn't have chance to test it personally today but after verifying it works I'll add it tomorrow since it is already supported in adlfs
https://github.com/dask/adlfs/blob/2bd9d9302e5d3511ce636246cbd2d7770000f00c/adlfs/spec.py#L304-L309

dvc/config_schema.py
isidentical: agreed

shcheklein:
A few other things we paid attention in the second PR that implements principal auth:

- support for all the regular auth flows (CLI configured, config, AD, env vars - https://docs.microsoft.com/en-us/python/api/overview/azure/identity-readme?view=azure-python#defaultazurecredential:

it is handled like this in the second PR, and it covers pretty much any possible Azure auth (+ our own config):

```
self._credential = (
            config.get("sas_token") or 
            config.get("storage_key) or
            self._az_config.get("storage", "key", None) or 
            self._az_config.get("storage", "sas_token", None) or
            ChainedTokenCredential(
                ServicePrincipalConfigCredential(config),
                DefaultCredentials()
            )
        )
```

- names for the config vars - I come to this again and again, we should discuss this before we merge any of those PRs - should they start with `azure_` or not? If they start, they will have the same names as env vars which makes it easier to read and explain to my mind

- https://docs.microsoft.com/en-us/python/api/overview/azure/identity-readme?view=azure-python#environment-variable - if this lib supports all possible Azure auth flows, we should probably add config values for these env vars.

- auth errors - how does it detect and fail if one of the methods fail. We should be friendly and explain that e.g. we were trying ENV vars, one of them is not set, etc.

setup.py
efiop: We had problems with old azure-storage-blob, now that's adlfs's problem to handle it properly and they do :) More dependencies also because it supports datalake, which we've been asked about before but never added support for(we'll need to do some adjustments to support it with fsspec too). So that's also justifiable.

dvc/config_schema.py
isidentical: These are the environment variables that adlfs fallbacks to (unless we give it from our config / local azure config):
https://github.com/dask/adlfs/blob/f7e810ab5294037f0d7ad5db16e480e119a48c3f/adlfs/spec.py#L375-L383

Maybe we could just document those?

isidentical: Actually we already handle it in our `_az_config` section (knack fallbacks to reading environment variables with `AZURE_{SECTION}_{KEY}` format;
https://github.com/iterative/dvc/blob/53ba51b5fed1ed19f03c3050ccd83ca128ce9012/dvc/fs/azure.py#L57-L67

though we don't comply with Azure docs there (this is also the old behavior, not something new);
https://docs.microsoft.com/en-us/python/api/overview/azure/identity-readme?view=azure-python#service-principal-with-secret

isidentical:
> should they start with azure_ or not?

Wouldn't it make sense to just follow s3's convention (`access_key_id` etc instead of `aws_s3_access_key_id`/`s3_access_key_id`)? 

> we should probably add config values for these env vars.

So I believe what we need is;
```
- connection_string
- account_name
- account_key
- sas_token
- client_id
- client_secret
- tenant_id
```

the corresponding environment variables provided by knack would be;
```
AZURE_STORAGE_CONNECTION_STRING
AZURE_STORAGE_ACCOUNT_NAME
AZURE_STORAGE_ACCOUNT_KEY
...
```

> auth errors - how does it detect and fail if one of the methods fail

If you provide multiple connection options (like both connection_string + account_name/account_key), it has an internal order (connection_string => principal => account_key => sas_token => default) and it will return the connection created only with the first one. If it can not succeed, it won't fall back to other options (which seems like the right choice, since you should generally offer only one reliable method). We can capture the `ValueError` and try to print a pretty message maybe?

shcheklein:
> Wouldn't it make sense to just follow s3's convention (access_key_id etc instead of  aws_s3_access_key_id/s3_access_key_id)?

good question! do you mean DVC convention or AWS itself? DVC has different examples (oss, gdrive both use prefixes). AWS itself has `aws_` prefix for all env vars, and for _some_ config vars. In our case we need to discuss, deprecate one of the ways we do it now.

also, https://rclone.org/dropbox/ - rclone has prefixes, probably to be explicit and/or have the same names as env vars.

> So I believe what we need is;

yep, it's a good start. Eventually support https://docs.microsoft.com/en-us/python/api/overview/azure/identity-readme?view=azure-python#environment-variables and the same order

> it won't fall back to other options which seems like the right choice, since you should generally offer only one reliable method

yep, agreed

> We can capture the ValueError and try to print a pretty message maybe?

that's exactly what I meant! Not the implementation-wise, but from the product perspective. it should be clear what auth method it tried and why did it fail


-------


what about https://docs.microsoft.com/en-us/python/api/overview/azure/identity-readme?view=azure-python#defaultazurecredential - does this lib support all possible ways that Azure lib itself supports?

isidentical:
> good question! do you mean DVC convention or AWS itself? DVC has different examples (oss, gdrive both use prefixes). AWS itself has aws_ prefix for all env vars, and for some config vars. In our case we need to discuss, deprecate one of the ways we do it now.

One thing to also note is whether to use `AZURE_BLAH_BLAH` or `AZURE_STORAGE_BLAH_BLAH`. Both adlfs and knack (azure-config parser that defaults to using environment variables) uses `AZURE_STORAGE_` prefix though the microsoft link you shared says `AZURE_`. (I'd prefer the `AZURE_` prefix just to comply with docs).

> what about https://docs.microsoft.com/en-us/python/api/overview/azure/identity-readme?view=azure-python#defaultazurecredential - does this lib support all possible ways that Azure lib itself supports?

It seems to be supporting, though it would be extra trouble for us to integrate it with already working adlfs. Do you a particular reason for us to use this?

shcheklein:
> It seems to be supporting, though it would be extra trouble for us to integrate it with already working adlfs. Do you a particular reason for us to use this?

I would expect more like this coming, that's why I think it would be great to rely on Azure lib itself to do Auth, we won't have to implement it on our won -  https://github.com/iterative/dvc/issues/5435#issuecomment-780602025

The question is - how easy it would be to change `adlfs` to rely on Azure DefaultCredentials flow? Or override their auth flow and pass an already authenticated object?

> I'd prefer the AZURE_ prefix just to comply with docs

yep, agreed. But I guess, please confirm with the team, share my thoughts if it helps. I think everyone should be onboard with the decision.

isidentical:
> The question is - how easy it would be to change adlfs to rely on Azure DefaultCredentials flow? Or override their auth flow and pass an already authenticated object?

I have to check, though as far as I can see all potential flows are somewhat supported and the commented way is also available (by passing credentials object directly). I also don't think that is a great idea to support browser-login on a platform where there are 5 different ways, though might be a future enhancement. I'll need to play with it, though it wouldn't be to hard to subclass adlfs's implementation and override the connection method (I don't feel so great about it, especially when we have not much of a gain now). 

> yep, agreed. But I guess, please confirm with the team, share my thoughts if it helps. I think everyone should be onboard with the decision.

CC: @iterative/dvc

dvc/fs/azure.py
```diff
@@ -1,58 +1,94 @@
...
+
+        if not any(login_info.values()):
+            login_info["credentials"] = DefaultAzureCredential()
+
+        for login_method, required_keys in [  # noqa
+            ("connection string", ["connection_string"]),
+            ("account key", ["account_name", "account_key"]),
+            ("SAS token", ["account_name", "sas_token"]),
+            ("anonymous login", ["account_name"]),
+            (f"default credentials ({_DEFAULT_CREDS_STEPS})", ["credentials"]),
```
isidentical, line 84: @shcheklein would this work for showing user which method we tried (this order comes from the adlfs)? I couldn't figure out how else I can describe default credentials, so I just appended the documentation link which includes all the stages that meant to be tried out.

shcheklein: look fine to me. I'm not sure about names and the order, etc - I would need to go too deep into details to say something meaningful re the details

shcheklein: I don't see though our own credentials -when we take principal things like token_id etc and pass into adlfs, or when we read and parse config. Is is part of the `default credentials` in this case?

isidentical: The `default credentials` happen when you don't specify anything in your .dvc/config / ~/.azure/config or in the appropriate environment variables (AZURE_STORAGE_* stuff). If we can't find anything in those, we fall back to default credentials which covers stuff like AZURE_* environment variables etc. 

> look fine to me. I'm not sure about names and the order, etc - I would need to go too deep into details to say something meaningful re the details

There is not much about the order we can do, though I was wondering what do you think about the `login_method` and the AzureAuthError message we raise below

shcheklein: yep, but I don't see in the list principal service auth defined in .dvc/config - client_id, tenant_id, etc ... is it intentional?

isidentical: Ah my bad🤦🏻! I forgot to add these options, thanks for catching it up

efiop (approved):
Thank you @isidentical ! And thanks @shcheklein for reviews :slightly_smiling_face: 

For the record: we'll also double check the performance.

isidentical:
It appears to be that, we introduced a regression on pushing big files (upload/upload_fobj) [it doubled x2], beside that if I am not missing anything rest seems same (ignoring deviation). I'll create a patch for the push regression
https://gist.github.com/isidentical/63ed3653a3113ea7fc253befa1e11845#file-before
https://gist.github.com/isidentical/63ed3653a3113ea7fc253befa1e11845#file-after

.pylintrc
```diff
@@ -293,7 +293,7 @@ ignored-classes=optparse.Values,thread._local,_thread._local,Dvcfile
 # (useful for modules/projects where namespaces are manipulated during runtime
 # and thus existing member attributes cannot be deduced by static analysis. It
 # supports qualified module names, as well as Unix pattern matching.
-ignored-modules=
+ignored-modules=azure
```
shcheklein, line 296: @isidentical do we need it here?

dvc/fs/azure.py
shcheklein: resolved, but I don't see the change? forgot to push?

dvc/fs/azure.py
```diff
@@ -1,58 +1,94 @@
...
+        )
+        login_info["account_key"] = config.get(
+            "account_key", self._az_config.get("storage", "key", None)
+        )
+        login_info["sas_token"] = config.get(
+            "sas_token", self._az_config.get("storage", "sas_token", None)
         )
-        if not self._credential:
-            self._credential = self._az_config.get("storage", "key", None)
+        login_info["tenant_id"] = config.get("tenant_id")
```
shcheklein, line 72: can they also com from storage?

.pylintrc
isidentical: yes

dvc/fs/azure.py
isidentical: No, at least it isn't listed here https://docs.microsoft.com/en-us/cli/azure/azure-cli-configuration#cli-configuration-values-and-environment-variables

.pylintrc
shcheklein: can we put a comment why do we need it here then? it's very hard to guess and will be very hard to understand when can we remove it

dvc/fs/azure.py
shcheklein: gotcha, thanks!

.pylintrc
isidentical: the comment above seems to (partially) explain the situation, do you mean like a comment that points to the file this is required?

shcheklein: yeah, may be it's me ... but for me it would be useful here to have something like - `azure - because we do .... in the .... and it's needed ....`

not a strong opinion, @isidentical .... just from my practice it's really hard to maintain all these settings, keep them up to date, etc.

PR #5466, opened 2021-02-15
live integration: get rid of dvc from live

pared:

## Important
Tests are failing because `dvclive` with corresponding change needs to be released. If change is accepted, please do not merge, I need to synchronize `dvclive` with `dvc` release to make the time window of incompatible versions as small as possible.


tests/func/experiments/conftest.py
```diff
@@ -67,7 +67,8 @@ def exp_stage(tmp_dir, scm, dvc):
 
 
 @pytest.fixture
-def checkpoint_stage(tmp_dir, scm, dvc):
+def checkpoint_stage(tmp_dir, scm, dvc, mocker):
+    mocker.patch("dvc.stage.run.MonitorConfig.AWAIT", 0.01)
```
pared, line 71: +/- 40% time less for `test_checkpoints.py`

dvc/stage/run.py
```diff
@@ -145,48 +170,109 @@ def run_stage(
...
-        yield monitor_thread
-    finally:
-        done.set()
-        monitor_thread.join()
-
-
-def _checkpoint_run(stage, callback_func, done, proc, killed):
-    """Run callback_func whenever checkpoint signal file is present."""
-    signal_path = os.path.join(stage.repo.tmp_dir, CHECKPOINT_SIGNAL_FILE)
+@dataclass
```
skshetry, line 173: Looks like we need a scheduler already for the runner. :)

skshetry: I was just saying that all the logic in the `stage/run` does not belong here. 

We should think of decoupling all of the logic here and in the `repro` to a `Runner` and `TaskScheduler` respectively. `Runner` should support hooks (`on_start`/`on_end`/`on_failure`/`every(1s)` etc.) and actually running the stages.
And, the `TaskScheduler` that schedules stages to be run from the given DAG (which would help us parallelize in the future).

We'd likely come to it later though, just sharing my thoughts 🙂 . But this also looks good for now.

pared: @skshetry That sounds like a great idea!

tests/func/test_live.py
```diff
@@ -1,12 +1,13 @@
...
 from textwrap import dedent
 
 import pytest
 from funcy import first
 
 from dvc import stage as stage_module
-from dvc.exceptions import MetricsError
+from dvc.exceptions import MetricDoesNotExistError, MetricsError
 
 LIVE_SCRITP = dedent(
```
dberenbaum, line 12: Can the spelling be fixed here and where it's used below as long as other changes are being made?

```suggestion
LIVE_SCRIPT = dedent(
```

dberenbaum (commented):
I want to make sure I understand what's happening here. You are not only making dvclive independent from dvc, but also keeping dvclive-like html output functionality implemented independently within dvc? Or do I have that completely wrong @pared ?

pared:
@dberenbaum You got it right.
We want to make `live` as lightweight as possible in its basic installation, so instead of adding `dvc` as a dependency, this change aims to introduce checkpoint-like behaviour.
`live`, if it finds `.dvc` dir, will create, under proper path kind of semafor file that dvc will be reacting to. If file is present, prepare summary and delete file, wait for another one and repeat.

dberenbaum:
Thanks, @pared! So how does dvclive generate the html summary if no `.dvc` dir is found?

pared:
@dberenbaum Without `dvc`, `dvclive` is unable to produce a summary. It can work only as a logger in that case.

dberenbaum:
Got it. Thanks, @pared! It seems like `live` without `dvc` is not a particularly likely regular scenario, since the functionality will be limited to logging metrics to a file, which users can do pretty easily themselves. However, maybe it's a better code design to keep them separate?

pared:
So the reasoning for the removal was to provide potential users with a logger that comes with no strings attached, and yet allows close integration with `dvc`.  While it definitely does not sound like a popular scenario, keeping dependencies as scarce as possible was mainly for this purpose.

dvc/stage/run.py
```diff
@@ -145,48 +170,109 @@ def run_stage(
...
+class MonitorConfig:
+    name: str
+    stage: "Stage"  # type: ignore[name-defined] # noqa: F821
```
pmrowla, line 176: This doesn't need to be ignored, you just need

```
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from dvc.stage import Stage
...
    stage: "Stage"
```

dvc/stage/run.py
```diff
@@ -145,48 +170,109 @@ def run_stage(
...
+                    "Error running '%s' task, '%s' will be aborted",
+                    config.name,
+                    config.stage,
                 )
-                _kill(proc)
-                killed.set()
+                _kill(config.proc)
+                config.killed.set()
             finally:
                 logger.debug("Remove checkpoint signal file")
```
pmrowla, line 273: This message should probably be generalized now

pmrowla: Also feels like the top level `_...` methods here should go into the base `Monitor` as class methods now (w/the checkpoint callback one in `CheckpointMonitor`)

dvc/stage/run.py
```diff
@@ -145,48 +170,109 @@ def run_stage(
...
+            )
+        )
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        # make sure all data is visualied after training
+        self.config.task()
+        super().__exit__(exc_type, exc_val, exc_tb)
+
+
+def _monitor_loop(config: MonitorConfig):
```
pmrowla, line 259: One general design question, does it make sense for us to have separate threads watching for both the checkpoint signal files and the dvclive signal files? Or should we have one monitor thread that checks for multiple signals (and runs their callbacks) sequentially like

```
while True:
    for signal_path, task in signals_to_watch:
        if os.path.exists(signal_path):
            try:
                task()
            ...
```

I guess the real question is would running both callbacks at the same time in parallel threads potentially cause any issues?

pmrowla: When the user creates the checkpoint signal, their code is supposed to block so that we know the user won't be touching the DVC workspace until we signal them to continue (by removing the signal file). This way we know that we can do whatever is needed in the workspace to create the checkpoint commits safely, and I'm assuming that dvclive works the same way?

In the event that a user creates both a checkpoint and dvclive signal file at the same time, I'm not sure that it is safe for dvclive to be doing anything while the workspace may be in the state where we are moving HEAD around and creating git commits during checkpoint creation

pared: @pmrowla Your proposition makes more sense, Ill introduce necessary changes

pared: Redesigned so that we have single monitor with multiple tasks provided as needed, depending on stage content.

pmrowla (approved):
LGTM

dberenbaum (approved):
Sorry, changed the status on this, but I don't have anymore comments.

pared:
Thanks for review! After discussing with @efiop we decided that I will give it a try and try to prepare `TaskScheduler` and `Runner` in this change.

skshetry:
@pared, I'll share some libraries that I was looking for inspiration, might be helpful/useful: [`invoke`](https://github.com/pyinvoke/invoke/blob/ed94c59f1eacc700dac3815530493e0809fe001d/invoke/runners.py#L47), [`doit`](https://github.com/pydoit/doit/blob/d1a1b7b3abc7641d977d3b78b580d97aea4e27ea/doit/runner.py#L20) and [`prefect`](https://github.com/PrefectHQ/prefect/blob/master/src/prefect/executors/dask.py) (mostly for experiment executors, generalizing it for `repro`).

pared:
@skshetry thanks, though it was misunderstanding between me and @efiop  - in this change we will just try to move `Monitor` logic out of `cmd_run`. So, definining `Scheduler` and `Runner` will not be a scope of this change.

tests/func/test_live.py
```diff
@@ -81,19 +89,11 @@ def make(summary=True, html=True, live=None, live_no_cache=None):
     yield make
 
 
-@pytest.mark.parametrize("report", (True, False))
+@pytest.mark.parametrize("html", (True, False))
 @pytest.mark.parametrize("summary", (True, False))
-def test_export_config_tmp(tmp_dir, dvc, mocker, summary, report):
```
pared, line 86: remove the `test_export_config_tmp` test - now that `live` is online `test_export_config` tests the same things

pared:
Extracted monitor logic out of `stage/run`.
Looking at problems that I will face releasing this small change, and looking at
[this](https://github.com/iterative/dvc/pull/5513/files#r581670930) change I came to a conclusion that it would be better to have those tests inside `dvclive`. That leads to the need of having fixtures reused between projects. Similar problem in `dvcx` was solved by copying the fixtures. It seems that official way of handling this kind of issue is to create a [`pytest-plugin`](https://docs.pytest.org/en/latest/writing_plugins.html). I think we should prepare the plugin. It could already be used in `dvcx`, `dvclive` and potentially in `dvc-bench`, once we find a capacity to get back to making the benchmarks pytest-like.

dvc/repo/live.py
```diff
@@ -1,15 +1,30 @@
...
+def create_summary(out):
+    from dvc.utils.html import write
+
+    assert out.live and out.live["html"]
+
+    metrics, plots = out.repo.live.show(str(out.path_info))
+
+    html_path = out.path_info.with_suffix(".html")
+    write(html_path, plots, metrics)
+    logger.info(f"\nfile://{os.path.abspath(html_path)}")
```
dberenbaum, line 25: It doesn't look like there's much to this HTML summary (either here or in https://github.com/iterative/dvc/blob/master/dvc/utils/html.py). Would it make sense to build out summary capability in dvclive? That would allow for complete dvclive functionality without dvc.

Also, could the summary output format be abstracted in dvclive so that non-HTML outputs could be built? For example, a matplotlib output that auto-updates, or even a very basic cli output. Different output types could enable users to see something like a progress bar for their model performance without opening a separate page that needs to be manually refreshed.

dberenbaum:
What is the intended user experience for including dvclive in a dvc stage?

dvc/repo/live.py
pared: @dberenbaum 
This question has few levels.
1.  > Would it make sense to build out summary capability in dvclive? That would allow for complete dvclive functionality without dvc

I guess it would make sense, but it would mean copying not only html module and this particular function, but quite a lot of code from `metrics` and `plots` (`live.show` is actually calling them) - so all the functionality related to vega-js would probably need to be copied too, the templates for example. Also having it in dvc allows us to use a command `dvc live` to visualize the `live` outputs. Its not too flashy as of today, but AFAIK, it is supposed to be iterated upon in the future.

> non-HTML outputs could be built ?

Probably, but that leaves us with a lot of questions related to plots itself - how would it work on different os-es? What about some pipeline running in a server where we connect via SSH? Also, what to do with `plots` command? Do we want to keep the HTML functionality? The plots were made in certain way in order to not have to support our own plotting framework, and use something that probably all of the users share - having browser. I think visualizing `live` using different means will be hard both to maintain and explain to users.

pared:
@dberenbaum I guess the main point is that if you use `dvclive` in your code, you don't need to specify live logs directory inside the code. So, if you run `dvc run ... --live logs` - the information that `logs` is supposed to be te live logs dir will be passed to the code. And if code itself does not specify `init` method, provided directory will be used.

dvc/repo/live.py
dberenbaum: Including these features in dvclive doesn't need to impact dvc at all. No need to block this PR or the release.

For users to be able to get value out of dvclive without dvc, could we build a basic summary capability in dvclive? I was wondering whether we could adapt what's currently in https://github.com/iterative/dvc/blob/master/dvc/utils/html.py as a starting point, but it doesn't need to be if there's a simpler way. Looking at that file, it looked like the data structures to feed to `HTML.write()` are reasonably simple, and the data is already being collected by dvclive.

Thoughts? I can open a new issue to discuss if it's worthwhile.

dberenbaum:
> @dberenbaum I guess the main point is that if you use `dvclive` in your code, you don't need to specify live logs directory inside the code. So, if you run `dvc run ... --live logs` - the information that `logs` is supposed to be te live logs dir will be passed to the code. And if code itself does not specify `init` method, provided directory will be used.

Is this being documented? I had completely missed this until now 😄 . I see it's in the help for `dvc stage add` and `dvc run`, but I don't see it in the command reference docs or anywhere else. CC @jorgeorpinel 

As another feature request _for the future_, would it be possible to simplify how to write code when wanting to use both checkpoints and dvclive? Right now, I'm writing code like:

```python
for _ in epochs:
    train()
    metrics = evaluate()
    with open("metrics.json", "w") as f:
        json.dump(metrics, f)
    for k, v in metrics.items():
        dvclive.log(k, v)
    dvclive.next_step()
    make_checkpoint()
```

This seems redundant. If I use a dvclive callback, I'm not even sure how I would write out the metrics or define the checkpoints for dvc since my code wouldn't iterate through each epoch.

pmrowla (approved):

Issue #5445, opened 2021-02-10
add: --to-remote question or improvement


jorgeorpinel:
> Follow up to https://github.com/iterative/dvc/pull/5198#issuecomment-774299750, #5301, and https://github.com/iterative/dvc.org/pull/2172#discussion_r573963049:

### Question

`add --to-remote` is a bit strange because normally `add` doesn't move target data, rather tracks it in-place (analog to `git add`). But `--to-remote` implies that external data will be moved into the workspace at some point, which we skip for now but "pre-push" (transfer) it to remote storage (for later `pull/fetch`).

As of now `add --to-remote` has a similar result to `get-url` + `add` + `push` + `remove`, `gc`. So OK, maybe it's nice to have a shortcut to all that, but **we already have `import-url (--to-remote)`** to achieve the same.

The only difference vs. importing is that the data source is not recorded as a dependency in the .dvc file. So you can't `update` it or `unfreeze`+`repro` it. However I don't see any use cases where you would want to prevent the .dvc from having this `dep`, as you can simply never `update` or `unfreeze` it.

TLDR: I think `import-url --to-remote` is enough and what we should recommend for these situations. And `add --to-remote` breaks the Git analogy. Cc @dberenbaum 

### Improvement

But if we keep it, an improvement would be to NOT require the `--external` flag with it (cc @isidentical). This saves the user from typing a flag that is always needed, but also make sense since the data is not actually being treated as external in the sense that it won't be tracked/controlled in it's original location (requiring external cache, etc.).

jorgeorpinel:
This probably applies to `add --to-cache` too BTW.

shcheklein:
My 2cs:

- add is more natural to find (I need to _add_ data to DVC but it's too large to do it with a regular `dvc add` on this machine - etc - I'll be searching for something like "How do add large data to DVC" or something.
- `import-url` has a different semantics (.dvc file is different). So, I am not sure why would keep only one, even though they are similar.
- agreed on removing `--external` constraint, `--to-remote` should imply it. Since it's probably almost always the case that we need both.

jorgeorpinel:
> add is more natural to find (I need to add data... searching for something like "How do add large data to DVC

I get the semantics part but that can probably addressed with good docs. Or maybe `import` could have a better name or become a flag in `add` (just a thought).

[...]

PR #5473, opened 2021-02-16
add, import-url: always resolve the output in the local path for --to-remote

isidentical:
During the discussions on #5445, it seemed `--external` is redundantly needed when trying to add an absolute path in the local system without `-o`. This wasn't the case for remote locations, such as `s3://` or `azure://`. So this PR forbids the usage of `--external` with `to-remote`/`to-cache` and normalizes absolute system paths just like other remote providers.

- [ ] dvc.org docs PR 🙂

### DIFF
--- a/dvc/utils/__init__.py
+++ b/dvc/utils/__init__.py
@@ -357,7 +357,7 @@ def resolve_output(inp, out):
     return ret
 
 
-def resolve_paths(repo, out):
+def resolve_paths(repo, out, always_local=False):
     from urllib.parse import urlparse
 
     from ..dvcfile import DVC_FILE_SUFFIX
@@ -391,6 +391,9 @@ def resolve_paths(repo, out):
         raise DvcException(msg)
     else:
         wdir = dirname
+        out = base
+
+    if always_local:
         out = base
 
     path = os.path.join(wdir, base + DVC_FILE_SUFFIX)
--- a/tests/func/test_add.py
+++ b/tests/func/test_add.py
@@ -6,7 +6,6 @@ import shutil
 import stat
 import textwrap
 import time
-from tempfile import gettempdir
 
 import colorama
 import pytest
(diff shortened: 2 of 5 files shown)


tests/func/test_add.py
```diff
@@ -1013,12 +1015,42 @@ def test_add_to_remote(tmp_dir, dvc, local_cloud, local_remote):
     assert local_remote.hash_to_path_info(hash_info.value).read_text() == "foo"
 
 
+def test_add_to_remote_absolute(tmp_dir, dvc, local_remote):
+    from shortuuid import uuid
+
+    temp_name = uuid()
+    temp_file = os.path.join(gettempdir(), temp_name)
+    with open(temp_file, "w") as stream:
+        stream.write("foo")
```
skshetry, line 1024: Use `tmp_path` fixtures here? 🙂

tests/func/test_import_url.py
```diff
@@ -340,6 +341,22 @@ def test_import_url_to_remote_directory(tmp_dir, dvc, workspace, local_remote):
         )
 
 
+def test_import_url_to_remote_absolute(tmp_dir, dvc, local_remote):
+    from shortuuid import uuid
+
+    temp_name = uuid()
+    temp_file = os.path.join(gettempdir(), temp_name)
+    with open(temp_file, "w") as stream:
+        stream.write("foo")
```
skshetry, line 350: Same here, we try to use fixture as much as possible. 🙂

dvc/utils/__init__.py
```diff
@@ -357,7 +357,7 @@ def resolve_output(inp, out):
     return ret
 
 
-def resolve_paths(repo, out):
+def resolve_paths(repo, out, always_local=False):
```
skshetry, line 360: IIUC, `always_local` is a high-level concept. Here, it does not convey what it does properly: using the base path as an output name. But, I don't have a good name to suggest either (that is succinct), so I'm fine with it as well.  😄

tests/func/test_add.py
isidentical: Isn't `tmp_path` gives the same path as the path of the repo? I need an external path outside of a repo

skshetry: Ah, right. There's `make_tmp_dir` fixture, that should do I think.

isidentical: Do you mean `tmp_path_factory.mktemp`?

skshetry: @isidentical, this one:

https://github.com/iterative/dvc/blob/9909ee85a1f61491a959962477208291e38f8f24/tests/dir_helpers.py#L263

This factory will give you a `tmp_dir` (aka `Path`) like fixture.

isidentical: Thanks!

dvc/utils/__init__.py
isidentical: I also couldn't think of a better one 🤔

skshetry (approved):

PR #5475, opened 2021-02-16
update: forbid updating repo-imports with --to-remote

isidentical:

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here. (will mention in iterative/dvc.org/pull/2175)

### DIFF
--- a/dvc/stage/imports.py
+++ b/dvc/stage/imports.py
@@ -1,9 +1,16 @@
 import logging
 
+from dvc.exceptions import InvalidArgumentError
+
 logger = logging.getLogger(__name__)
 
 
 def _update_import_on_remote(stage, remote, jobs):
+    if stage.is_repo_import:
+        raise InvalidArgumentError(
+            "Can't update a repo import with --to-remote"
+        )
+
     url = stage.deps[0].path_info.url
     stage.outs[0].hash_info = stage.repo.cloud.transfer(
         url, jobs=jobs, remote=remote, command="update"
--- a/tests/func/test_update.py
+++ b/tests/func/test_update.py
@@ -3,6 +3,7 @@ import os
 import pytest
 
 from dvc.dvcfile import Dvcfile
+from dvc.exceptions import InvalidArgumentError
 from tests.unit.fs.test_repo import make_subrepo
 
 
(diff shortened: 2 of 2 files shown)


efiop (approved):

jorgeorpinel:
Hi @isidentical I see there's a docs PR for this (https://github.com/iterative/dvc.org/pull/2175/files), thanks!

I'm just not sure this was mentioned completely? What was forbidden exactly? It seems like update works perfectly well with --to-remote imports. Maybe I just don't understand this PR.

dvc/stage/imports.py
```diff
@@ -1,9 +1,16 @@
...
+from dvc.exceptions import InvalidArgumentError
+
 logger = logging.getLogger(__name__)
 
 
 def _update_import_on_remote(stage, remote, jobs):
+    if stage.is_repo_import:
+        raise InvalidArgumentError(
+            "Can't update a repo import with --to-remote"
+        )
```
jorgeorpinel, line 12: What's a repo import?

jorgeorpinel:
Oh I see "repo import" means a `dvc import` (not `import-url`). OK please see https://github.com/iterative/dvc.org/pull/2175#pullrequestreview-596702408

jorgeorpinel:
BTW I'm not sure that term will mean anything for users. If that exception is a user message I would rephrase to something more explicit like "Data imported with `dvc import` can't be updated --to-remote." or "Only data imported with `dvc import-url --to-remote` can be updated `--to-remote`."

dvc/stage/imports.py
jorgeorpinel: BTW I'm not sure that term will mean anything for users. If that exception is a user message I would rephrase to something more explicit like "Data imported with `dvc import` can't be updated `--to-remote`." or "Only data imported with `dvc import-url --to-remote` can be updated `--to-remote`."

isidentical:
> Oh I see "repo import" means a dvc import (not import-url). OK please see iterative/dvc.org#2175 (review)

I agree that the error message is a bit term-focused, even though it is the correct term. Let's just use `Only data imported with dvc import-url can be updated with --to-remote.`

jorgeorpinel:
Yes or maybe more general like "Data imported from other DVC or Git repositories can't be updated to-remote." to avoid terminology completely. Although maybe mentioning the exact command is better. Up to you

jorgeorpinel:
p.s. is there a core issue to put in these ideas?

isidentical:
> p.s. is there a core issue to keep track of these ideas?

nop, though feel free to just create an issue and CC me

PR #5479, opened 2021-02-16
fs: detach state from fs

efiop:
We used to think this way, but state is not really an fs concept
(fsspec's dircache is though), as it is a cache of hashes of objects
that we are staging, and the hash doesn't depend on fs, as it is usually
computed by us by reading from fs.

Pre-requisite for diskcache-based state.

### DIFF
--- a/dvc/checkout.py
+++ b/dvc/checkout.py
@@ -167,7 +167,7 @@ def _checkout_file(
         _link(cache, cache_info, path_info)
         modified = True
 
-    fs.state.save(path_info, obj.hash_info)
+    fs.repo.state.save(path_info, fs, obj.hash_info)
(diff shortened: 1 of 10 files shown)


dvc/checkout.py
```diff
@@ -217,7 +217,7 @@ def _checkout_dir(
         or modified
     )
 
-    fs.state.save(path_info, obj.hash_info)
+    fs.repo.state.save(path_info, fs, obj.hash_info)
```
efiop, line 220: Looks a bit odd with `fs.repo`, but we'll be getting rid of `fs.repo` in the folloups and it will be cleaner. For now detaching the sate from fs.

PR #5499, opened 2021-02-19
stage:list: update output and help text

skshetry:
* Replaced 'Produces' with 'Outputs'
* Replaced 'Stage depends on' with just 'Depends on'
* Changes help text of --names-only and --fail

Addressing some of the review comments from https://github.com/iterative/dvc/pull/5282#pullrequestreview-571831191.

### DIFF
--- a/dvc/command/stage.py
+++ b/dvc/command/stage.py
@@ -348,7 +348,7 @@ def add_parser(subparsers, parent_parser):
         "--fail",
         action="store_true",
         default=False,
-        help="Do not suppress any syntax errors, fail immediately.",
+        help="Fail immediately, do not suppress any syntax errors.",
     )
     stage_list_parser.add_argument(
         "-R",
--- a/dvc/command/stage.py
+++ b/dvc/command/stage.py
@@ -22,10 +22,10 @@ def generate_description(stage: "Stage") -> str:
         return ", ".join(out.def_path for out in outs)
 
     if not stage.deps and not stage.outs:
-        return "No outputs or dependencies."
+        return "No outputs or dependencies"
 
     if not stage.outs and stage.deps:
-        return "Stage depends on " + part_desc(stage.deps)
+        return "Depends on " + part_desc(stage.deps)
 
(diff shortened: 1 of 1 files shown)


dvc/command/stage.py
```diff
@@ -34,7 +34,7 @@ def is_plot_or_metric(out: "BaseOutput"):
 
     outs = list(filterfalse(is_plot_or_metric, stage.outs))
     if outs:
-        desc.append("Produces " + part_desc(outs))
+        desc.append("Outputs " + part_desc(outs))
```
skshetry, line 37: @jorgeorpinel, still not sure about `metrics/params` part, so I am leaving `Generates` part as-is. 

For others, here's the discussion around it: https://github.com/iterative/dvc/pull/5282#discussion_r560641053

dberenbaum: I don't see why "Generates" is any more specific to metrics or plots than "Outputs." Maybe "Reports" would be a bit more specific, but I'm not sure it's worth it to have a different verb. That being said, this is truly nitpicking and seems unlikely to even be noticed by users, so I'm fine to leave as is if this is your preference @skshetry .

dvc/command/stage.py
```diff
@@ -348,7 +348,7 @@ def add_parser(subparsers, parent_parser):
         "--fail",
         action="store_true",
         default=False,
-        help="Fail immediately if there's an error.",
+        help="Do not suppress any syntax errors, fail immediately.",
```
dberenbaum, line 351: Again nitpicking, but I'd suggest switching the order so the primary effect of the flag is obvious from the first couple words.

```suggestion
        help="Fail immediately, do not suppress any syntax errors.",
```

dberenbaum (approved):
Left a couple very minor comments, but up to you to decide whether to accept them @skshetry .

dvc/command/stage.py
skshetry: I think `Reports` is what I am looking for, though I find it to be a bit _formal_. Thanks @dberenbaum.

PR #5505, opened 2021-02-22
state: use diskcache

efiop:
Initial migration to diskcache. Same as our `State`, diskcache's `Cache` is basically a sqlite database underneath. Saves us a lot of code (need to migrate indexes too in the follow-up PR), supports multiprocessing/threading(a big problem with our in-house state), allows us to use pbar even when using state (better ui) and even though we are now saving hashes of files in datasets to state, we are still getting the same level of performance, even without advanced tuning.

Con is that we are removing `nolock` option from state, which allowed us to somewhat work on nfs, but that was dangerous (we were relying on keeping our repo lock) and hard to maintain. We'll do a round of QA specifically for nfs/cifs to see how dvc works with it in general and we'll likely make state dir configurable, to allow placing it in /tmp or elsewhere on normal fs.

### DIFF
--- a/dvc/api/__init__.py
+++ b/dvc/api/__init__.py
@@ -29,7 +29,7 @@ def get_url(path, repo=None, rev=None, remote=None):
             raise OutputNotFoundError(path, repo)
 
         cloud = metadata.repo.cloud
-        hash_info = get_hash(path_info, _repo.repo_fs, "md5")
+        hash_info = get_hash(path_info, _repo.repo_fs, "md5", _repo.state)
         return cloud.get_url_for(remote, checksum=hash_info.value)
 
 
--- a/dvc/api/__init__.py
+++ b/dvc/api/__init__.py
@@ -29,7 +29,7 @@ def get_url(path, repo=None, rev=None, remote=None):
             raise OutputNotFoundError(path, repo)
 
         cloud = metadata.repo.cloud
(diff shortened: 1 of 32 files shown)


dvc/api/__init__.py
```diff
@@ -29,7 +29,7 @@ def get_url(path, repo=None, rev=None, remote=None):
             raise OutputNotFoundError(path, repo)
 
         cloud = metadata.repo.cloud
-        hash_info = get_hash(path_info, _repo.repo_fs, "md5")
+        hash_info = get_hash(path_info, _repo.repo_fs, "md5", _repo.odb)
```
efiop, line 32: This temporary and will be `fs.info(path_info).get("md5")` in the future, so nevermind the ugliness.

dvc/checkout.py
```diff
@@ -30,7 +30,7 @@ def _changed(path_info, fs, obj, cache):
         return True
 
     try:
-        actual = get_hash(path_info, fs, obj.hash_info.name)
+        actual = get_hash(path_info, fs, obj.hash_info.name, cache)
```
efiop, line 33: `get_hash` shouldn't be used directly, we should use `stage()` instead. Again, this will be addressed a bit later, need to move staging/state/dir_info code around a bit first.

PR #5513, opened 2021-02-24
tests: temporarily skip dvclive tests

efiop:
Our builds are timing out because pip is confused by

> dvclive 0.0.1a1 requires dvc==2.0.0a0, but you'll have dvc 2.0.0a1+e61136 which is incompatible.

and tries really hard to resolve conflicts.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,6 @@ install_requires = [
     "pyparsing==2.4.7",
     "typing_extensions>=3.7.4",
     "fsspec>=0.8.5",
-    "dvclive>=0.0.1a0",
     "diskcache>=5.2.1",
 ]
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -88,7 +88,6 @@ def run(self):
     "pyparsing==2.4.7",
     "typing_extensions>=3.7.4",
     "fsspec>=0.8.5",
-    "dvclive>=0.0.1a0",
```
efiop, line 91: I guess this was a typo and it was meant for `tests`, right? cc @pared

tests/func/test_live.py
```diff
@@ -54,6 +54,11 @@ def dump(value, path):
 
 @pytest.fixture
 def live_stage(tmp_dir, scm, dvc):
```
efiop, line 56: Need to consider not using dvclive in the dvc func tests here or creating separate integration tests for dvclive.

efiop:
Merging to unblock all other PRs.

setup.py
pared: Right, sorry for that

tests/func/test_live.py
pared: Hmm, maybe we should move them into `dvclive` package

Issue #5500, opened 2021-02-19
gs:  QA our implementation 


isidentical:
Some notable google storage issues
- #5488
- #3005
- #1522
- ~~#1675~~ (not gs relevant)
- #5509

PR #5526, opened 2021-02-26
gs: Migrate Google Storage to FSSpec 

isidentical:
Part of #5500. I haven't started doing QA yet and testing the issues, only the draft for the initial fsspec migration. Still WIP.


isidentical:
Here is the benchmark results for operations on 2 different subjects (one big file, a directory of 5K small files); https://gist.github.com/isidentical/3b625e6da048c512093781ed03c69176

They are nearly identical, the push seems a bit faster and the pull seems a bit slower. Though it might be just the deviation.

dvc/fs/gs.py
```diff
@@ -1,228 +1,43 @@
...
-from dvc.progress import Tqdm
 from dvc.scheme import Schemes
 
-from .base import BaseFileSystem
+from .fsspec_wrapper import FSSpecWrapper
 
-logger = logging.getLogger(__name__)
 
-
-def dynamic_chunk_size(func):
```
isidentical, line 19: Underlying fsspec upload is a bit complicated then a just a call (due to their own chunking) and they use a static threshold. I couldn't seem to trigger it with the defaults (5 MiB) when uploading 8 x 100 MiB files (I was thinking it might bottleneck, but it didn't). So we might be just safe on it, without actually worrying to much.

dvc/fs/gs.py
```diff
@@ -1,228 +1,43 @@
...
+        login_info["token"] = config.get("credentialpath")
+        return login_info
 
-    def isfile(self, path_info):
-        if path_info.path.endswith("/"):
-            return False
+    def _entry_hook(self, entry):
+        if "md5Hash" in entry:
+            entry["md5"] = base64.b64decode(entry["md5Hash"]).hex()
+        return entry
```
efiop, line 36: I guess no longer be needed if we decide to switch to etag https://github.com/iterative/dvc/issues/5527 Out of scope for this PR though.

efiop (approved):

dvc/fs/gs.py
isidentical: yep. This is temporary.

dvc/fs/fsspec_wrapper.py
```diff
@@ -37,10 +39,20 @@ def _entry_hook(self, entry):
...
+            entry["name"].endswith("/")
+            and entry["type"] == "file"
+            and entry["size"] == 0
+        )
 
     def isfile(self, path_info):
-        return self.fs.isfile(self._with_bucket(path_info))
+        if not self.exists(path_info):
+            return False
+        return not self.isdir(path_info)
```
efiop: We need a comment (maybe with a link to an issue) to clarify why we are doing such strange things.

dvc/fs/fsspec_wrapper.py
```diff
@@ -11,6 +12,7 @@ class FSSpecWrapper(BaseFileSystem):
     def fs(self):
         raise NotImplementedError
 
+    @lru_cache(512)
```
efiop, line 15: Why do we need it bt? Isinstance is too slow?

isidentical: It is mostly for `.isdir()/.isfile()` since they call `.exists()` + `.info()` which under the hood calls this function with the same arguments.

dvc/fs/fsspec_wrapper.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def _entry_hook(self, entry):
+        """Simple hook method to be overriden when wanted to process
+        entries within info() and ls(detail=True) calls"""
+        return entry
+
+    def isdir(self, path_info):
+        if not self.exists(path_info):
+            return False
+
```
isidentical, line 44:
```suggestion
# Empty directories in prefix-based storages are represented as
# empty files. The check below ensures that this is either a non-empty
# directory or a file that is empty and its name contains a trailing slash.
# https://github.com/iterative/dvc/issues/5521

```
@efiop how does this look like?

dvc/fs/fsspec_wrapper.py
```diff
@@ -0,0 +1,127 @@
...
+        entry = self.info(path_info)
+        return entry["type"] == "directory" or (
+            entry["name"].endswith("/")
+            and entry["type"] == "file"
+            and entry["size"] == 0
+        )
```
efiop, line 50: This is for `isdir` in `import-url` for directories, right? It makes importing empty dirs possible for adlfs and gsfs, correct? Could we just filter-out `dir/` objects in `_collect_dir` instead?

isidentical: > Could we just filter-out dir/ objects in _collect_dir instead?

We need to take care of this stuff in the download(); https://github.com/iterative/dvc/blob/92b95bae9b0bfed1321c3406ff2295c3118be028/dvc/fs/base.py#L276-L279

Issue #5528, opened 2021-02-26
Add Sonatype Nexus remote repository support


BenLinnik:
Sonatype Nexus is a famous and widely used repository manager used to securely store packages and has an excellent user access control.

Usually it is used to host R, Java, Python and other build artifacts. Nevertheless, its hosted RAW repository type allows it to store any files and data! 
It would be nice to add this famous, tested and flexible artifactory as a remote storage type to DVC and be able store ML data.

![image](https://user-images.githubusercontent.com/12953323/109366876-d640dc80-7894-11eb-9a4f-10be67c000f4.png)

BenLinnik:
@efiop Privet Ruslan :)
This is not just wishful thinking, I actually implemented this already- I'm running all pre-commit hooks to make a Pull request with this feature included. Please give me some more minutes 🚀

efiop:
Hi @BenLinnik !

Please note that we are migrating to fsspec https://github.com/iterative/dvc/issues/5162 and there are a few remote PRs pending already for other clouds (e.g. osf, dropbox, etc). We will not be merging new implementations unless they comply with [fsspec](https://github.com/intake/filesystem_spec). At the same time it is becoming harder and harder for dvc team to maintain quality support for the variety of clouds on our own, so the absolute best way to go about it would be for you to follow fsspec and maintain your implementation (e.g. how [s3fs](https://github.com/dask/s3fs) does it) in your own package (and also maybe try to get it accepted into https://github.com/intake/filesystem_spec/tree/master/fsspec/implementations). That way it will be really easy for us to accept it, or, if our plugin mechanism will be ready by that time, just use your implementation automatically through fsspec registry or some explicit option.

BenLinnik:
Oh ok, this is new to me. I will have a look into it. Please keep up the good work!

efiop:
@BenLinnik Thank you! :pray: And sorry for making you now adapt your PR again, we are in the middle of migrating ourselves, so there is a mixup in the code :slightly_frowning_face: We'll document it properly in the contribution guide after the 2.0 release. Thanks again for contributing!

efiop:
Doesn't seem like there is much demand for this at all. Closing.

PR #5529, opened 2021-02-27
Add Nexus3 artifactory support

BenLinnik:

I know there is a feature freeze, and I know this PR will not be accepted ✋ ; see #5528. But for future reference and people interested in this integration, I nevertheless created this PR so that the community can benefit from it ❤️ . I don't want my work to be for nothing😞 

Given my code, which methods of [fsspec](https://github.com/intake/filesystem_spec) do I have to support so that DVC can work with it? `mkdir`, `rmdir`, `info`, `modified`, `ls`, etc. I'm asking because I will maybe refactor this to be a [fsspec implementation](https://github.com/intake/filesystem_spec/tree/master/fsspec/implementations) as suggested by @efiop.

### DIFF
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -126,6 +126,7 @@ SCHEMA = {
         "webhdfs": str,
         "ssh": str,
         "azure": str,
+        "nexus3": str,
         # This is for default local cache
         "dir": str,
         **LOCAL_COMMON,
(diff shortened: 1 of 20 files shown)


efiop:
As discussed in the original issue, we won't be able to maintain this implementation ourselves, so we need someone to implement it based on https://github.com/intake/filesystem_spec and maintain it himself. Soon dvc's plugin mechanism will be ready and we'll pick up such filesystems automatically. Closing for now

Issue #5477, opened 2021-02-16
exp run: validate params filename and parameter names


pmrowla:
One thing I noticed when trying the changes from https://github.com/iterative/dvc/issues/5302 is that if the user specifies a parameter or params file that doesn't exist (i.e. typo's a param name) we just silently accept it and create the new file or new entry in the existing params file (this behavior existed before the `-S/--set-param` change as well).

If the specified parameter doesn't exist in the target stage/pipeline we should probably throw an error instead of silently ignoring it.

dberenbaum:
Need to consider what's expected here. It could be useful for new users doing experiments who have no prior dvc experience to setup experiments without ever touching `params.yaml` by adding parameters this way. Maybe a warning is best?

mattlbeck:
I have started to use `dvc exp run` in earnest recently. I have found this behaviour particularly unexpected and had assumed that misspelling a parameter would result in the experiment not running. 

In my opinion the current behaviour is pretty dangerous and a major source of experiment error if the user does not validate that the params were correctly applied themselves afterwards.

I appreciate allowing users to stay away from params files, but I think some `--verify-params` flag would be very much appreciated at least.

mmeendez8:
I understand @mattlbeck and @pmrowla point here.

I would like to add another view point here. When dealing with DL frameworks, most time you have model config and training parameters already defined in the framework configuration, see for example:

- [mmdetection config files](https://github.com/open-mmlab/mmdetection/blob/master/configs/_base_/models/retinanet_r50_fpn.py)
- [detectron config files](https://github.com/facebookresearch/Detectron/blob/master/configs/12_2017_baselines/e2e_faster_rcnn_R-101-FPN_1x.yaml)

So if all parameters have to exist in `params.yaml`, I am forced to duplicate all my config parameters twice!

dberenbaum:
@mattlbeck Do you think a warning would be sufficient, or do you think it's important for the experiment to fail?

@mmeendez8 You only need to add to `params.yaml` those parameters that you change via experiments, right? So basically validating the parameters means you have to edit `params.yaml` directly instead of using `--set-param` if you are introducing a new parameter?

mattlbeck:
If it is important to maintain a route to modifying parameters solely via the CLI, then I would suggest a new `--verify-params` flag that fails the experiment when new params are declared via a `dvc exp run`. 

[...]

PR #5531, opened 2021-03-01
Introduce pydantic schema

skshetry:
From the [dvcyamlschema/gen.py](https://github.com/iterative/dvcyaml-schema/blob/724d2ba40d13978334f53f988b19b2b7510bad97/gen.py), added some 2.0 stuff and improved schema in the hopes that this will replace the Voluptuous schema we have for `dvc.yaml` at least.

I'm still hesitant to go full-blown with Pydantic yet, as making both of these schemas compatible is harder than I thought previously, which means there might be some bugs lurking around. The tests do pass with this new schema, but they don't tests the negative cases well. And, we also need to consider performance (voluptuous is faster, but is it worth to continue using it as it's DX is very bad?).

This pydantic schema can be used to provide `jsonschema` as well, you can just use it as:

```python
from dvc.schema.dvc_yaml import Schema

print(Schema.schema_json(indent=4))
```

The `dvcyaml_schema` repo will simply use this API if/when this gets merged, which will make it closer to where we work.
See also https://github.com/iterative/dvc/issues/5371#issuecomment-787670244, pydantic has a bit better error messages than Voluptuous.

Related: https://github.com/iterative/dvcyaml-schema/issues/7


dvc/dvcfile.py
```diff
@@ -122,7 +123,7 @@ def _check_gitignored(self):
         if self._is_git_ignored():
             raise FileIsGitIgnored(self.path)
 
-    def _load(self):
+    def _load(self, use_pydantic: bool = False):
```
skshetry, line 126: To test new schema, you can toggle this to `True`, and it should start using the new schema for validation.

pared (approved):

skshetry:
Closing it for now, will come back to it after UI is done. Unfortunately, I haven't been able to work more on this right now, so this needs more research.

As this PR stands, it is only meant to change the validation schema, so I can't provide any compelling argument other than to say that we are maintaining this schema already on `dvcyaml-schema` for generating `schema.json` and right now, it provides benefits of both generating `schema.json` and validating `dvc.yaml` file with a single schema definition.

It feels like cheating if I make other arguments just to make a point on its potential benefits that we might get in the future, which I am quite not sure of myself. I do believe we'll come to this again soon. 🙂

daavoo:
> Closing it for now, will come back to it after UI is done. Unfortunately, I haven't been able to work more on this right now, so this needs more research.
> 
> As this PR stands, it is only meant to change the validation schema, so I can't provide any compelling argument other than to say that we are maintaining this schema already on `dvcyaml-schema` for generating `schema.json` and with this PR, it provides benefits of both generating `schema.json` and validating `dvc.yaml` file with a single schema definition.
> 
> It feels like cheating if I make other arguments just to make a point on its potential benefits that we might get in the future, which I am quite not sure of myself. I do believe we'll come to this again soon. slightly_smiling_face

I'm not aware of the internal implications of this change but I think that having the schema inside the core repo makes a lot of sense. Issues related with having `dvcyaml-schema` outdated have been raised at least a few times. (And some discord messages, i.e. recently https://discord.com/channels/485586884165107732/485596304961962003/869072338778468373)

Probably not enough to be considered an argument at this point but I think that adding this feature (and introducing internal usage of pydantic) could open new possibilities regarding `params` validation (https://github.com/iterative/dvc/issues/5477) and enhancements to the overall `params` UI. I was working on a proposal about having internal `params` validation and potentially allowing users to provide it's own `params` schemas. Having this P.R. as foundation would make it way easier to implement.

SoyGema:
Under a context that is not relevant to this conversation and without being aware of this specific initiative , maybe worth citing that found  that [voluptuous](https://pypi.org/project/voluptuous/) is no longer actively maintained and failing with [pip-rating](https://github.com/Nekmo/pip-rating) . This could nudge the decision beyond any DVC functionality and add a maintenance dimension to the  discussion. 


Captura de pantalla 2023-10-20 a las 16 41 26

PR #5534, opened 2021-03-02
gs: use etag for the checksum parameter

### LINKED ISSUE #5527: Switch to using etag from md5 for google storage dependencies
In this way, it appears to be that we don't support composite objects due to they don't have an attached md5 hash: https://cloud.google.com/storage/docs/composite-objects#metadata

> Composite objects do not have an MD5 hash metadata field.

I don't know how breaking the transition would be, though 2.0 seems like the best place to target considering it won't take much effort to switch those in the codebase. Related #5500

isidentical:
Switch from `md5` to `etag` as the checksum parameter. Due to `etag`
is not preserved while doing server side copies, this patch drops support
for external gs outputs.

Resolves #5527

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here. (iterative/dvc.org/pull/2249)

### DIFF
--- a/dvc/dependency/gs.py
+++ b/dvc/dependency/gs.py
@@ -1,6 +1,8 @@
 from dvc.dependency.base import BaseDependency
-from dvc.output.gs import GSOutput
+from dvc.output.base import BaseOutput
+
+from ..fs.gs import GSFileSystem
 
 
-class GSDependency(BaseDependency, GSOutput):
-    pass
+class GSDependency(BaseDependency, BaseOutput):
+    FS_CLS = GSFileSystem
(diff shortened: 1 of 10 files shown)


efiop (approved):

shcheklein:
@isidentical @efiop @jorgeorpinel is it mentioned in the docs somewhere 🤔 ?

efiop:
@shcheklein Yes, https://github.com/iterative/dvc.org/issues/2228#issuecomment-788813586

shcheklein:
@efiop sorry for a question that was not clear enough, I was asking primarily if we mention in the docs somewhere that we use specifically md5 or etag or something else for a specific remote.

efiop:
@shcheklein Ah, I see. I don't think there's anything left.

jorgeorpinel:
> asking primarily if we mention in the docs somewhere that we use specifically md5 or etag or something else for a specific remote

We mention which type of hash is used per output/location combo in https://dvc.org/doc/user-guide/project-structure/dvc-files#output-entries so I guess we should update that indeed. Will do ⌛

jorgeorpinel:
Good point, OK then!

PR #5535, opened 2021-03-02
gs: Use rm() instead of delete() calls

isidentical:
delete is a direct alias to rm(), though apparently it is kind of
broken on the latest release of gcsfs. This doesn't change any behavior
but only uses rm() directly instead of the given alias.


isidentical:
Upstream fix: https://github.com/dask/gcsfs/pull/348

Issue #5521, opened 2021-02-25
get-url, import-url: handle empty directory imports


isidentical:
When getting/importing an empty directory, we don't take any actions in the workspace since we don't create the parent directories unless there are files.
```
dvc import-url some://empty/dir/
dvc get-url some://empty/dir/
```

isidentical:
Awaiting fixes on the upstream implementations (adlfs, gcsfs) before we can act.

jorgeorpinel:
> When getting/importing an empty directory, we don't take any actions in the workspace

Actually it currently fails with some repetitive and contradicting messaging 🙂 

```console
$  dvc import-url ../ext/empty
Importing '../ext/empty' -> 'empty'
WARNING: '../ext/empty' is empty.
WARNING: '../ext/empty' is empty.
ERROR: failed to import ../ext/empty. You could also try downloading it manually, and adding it with `dvc add`. - output 'empty' does not exist
```

Contradiction: "is empty" vs. "does not exist".

DVC 2.0.5 (pip)

isidentical:
Awaiting: https://github.com/intake/filesystem_spec/issues/562#issuecomment-869664069

PR #5536, opened 2021-03-02
import-url: support importing empty directories on object storage services

isidentical:
Empty directories are basically files with a trailing slash at the end of their names,
with empty contents on object service providers (such as azure, google storage). This
patch implements the `is_empty()` method for the `FSSpecWrapper` to determining whether
a path is empty (directory or a file) and also tweaks `.isdir()` in order to see empty
directories.

The base class is only used by azure and gcsfs for now, but if we go on the route of
migrating normal providers (like proper filesystems, not just object storages), we might
need to do these tweaks on a subclass (like `ObjectStorageFSSpecWrapper` etc), but I
don't think it is necessary as of now.

The `is_empty` method is already there, and also used in the importing part to
warn users that they are importing something empty (https://github.com/iterative/dvc/blob/e385e59a098cd661ea1ea4257c9a5735c5c32bc6/dvc/output/base.py#L252-L254)
so this is not a new method. Just for the reference, as far as I can locate there
is no variant of this method in the fsspec.

Related #5521

### DIFF
--- a/dvc/fs/fsspec_wrapper.py
+++ b/dvc/fs/fsspec_wrapper.py
@@ -42,6 +42,11 @@ class FSSpecWrapper(BaseFileSystem):
         if not self.exists(path_info):
             return False
 
+        # Directory in object storages are interpreted differently
+        # among different fsspec providers, so this logic is a temporary
+        # measure for us to adapt as of now. It checks whether it is a
+        # directory (as in a prefix with contents) or whether it is an empty
+        # file where it's name ends with a forward slash
         entry = self.info(path_info)
         return entry["type"] == "directory" or (
             entry["size"] == 0
--- a/dvc/fs/fsspec_wrapper.py
+++ b/dvc/fs/fsspec_wrapper.py
@@ -1,5 +1,6 @@
 import os
 import shutil
+from functools import lru_cache
 
 from dvc.progress import Tqdm
 
@@ -11,6 +12,7 @@ class FSSpecWrapper(BaseFileSystem):
     def fs(self):
         raise NotImplementedError
 
+    @lru_cache(512)
     def _with_bucket(self, path):
         if isinstance(path, self.PATH_CLS):
             return f"{path.bucket}/{path.path}"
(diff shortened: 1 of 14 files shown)


tests/remotes/base.py
```diff
@@ -16,7 +16,7 @@ def is_dir(self):
     def exists(self):
         raise NotImplementedError
 
-    def mkdir(self, mode=0o777, parents=False, exist_ok=False):
+    def mkdir(self, mode=0o777, parents=False, exist_ok=False, empty=False):
```
efiop, line 19: It's pretty odd that `empty` is leaking into this fixture. How about we create separate test cases for such filesystems? They can use raw fs interface, no need to use this fixture.

isidentical: > It's pretty odd that empty is leaking into this fixture.

I didn't want to override whole `_gen()` function just to support it

> How about we create separate test cases for such filesystems?

Or perhaps we can just to it explicitly on the testcase with determining what the contents should be. If you think that it is a bit weird to do this on spot, we can always hoist the same code into `_gen()`.

dvc/fs/fsspec_wrapper.py
```diff
@@ -37,10 +39,32 @@ def _entry_hook(self, entry):
...
+        entry = self.info(path_info)
+        return entry["type"] == "directory" or (
+            entry["size"] == 0
+            and entry["type"] == "file"
+            and entry["name"].endswith("/")
+        )
 
     def isfile(self, path_info):
-        return self.fs.isfile(self._with_bucket(path_info))
+        if not self.exists(path_info):
```
isidentical, line 58: @efiop forgot to mention this during our meeting, that the reason we have this guards instead of a try/except is that if I do this on isdir()
```
def isdir():
    try:
       ...
    except FileNotFoundError:
       return False
```

then `isfile()` is just going to negate the `False` to `True` even though the `False` is coming because the file doesn't exist (not because the file is not a directory). Since it is impossible to distinguish these operations, and for the sake of not repeating the same logic (if a file is empty and has trailing slash return False from isfile) I just used isdir() and negated the result but I have a pre-guard that ensures the file exists in the first place so that we won't return True when the file doesn't exist.

efiop (approved):

jorgeorpinel:
Hi, interesting! What's the use case for this? Bootstrapping (i.e. you import dir that's currently empty but expected to have data later) for later `update`? If so should we add a new example on this to https://dvc.org/doc/command-reference/import-url?

Similar to to-cache/to-remote cases cc @shcheklein

shcheklein:
@jorgeorpinel it's an edge case. I don't see this worth creating a separate example or even mentioning to be honest in docs in some special way.

jorgeorpinel:
Okdk. I thought there was a broader use case for this.

Issue #5451, opened 2021-02-11
params: untracked values in params files are not ignored in params/exp commands


geblanco:
# Bug Report

exp show: Invalid or wrong values for column names and values

## Description

I have an experiment trying different parameters, it is the last stage of the pipeline, and the only with checkpointed outputs. Following this procedure: run the experiment (`dvc exp run`), change a parameter in `params.yaml` and run it again, do it several times. To visualize the results I run `dvc exp show`, but I get wrong column names and values.

I have a doubt about the intended behavior of this feature, when `showing` experiments, the idea is to show the comparison of metrics along parameters. It is expected to show the difference in metrics for ALL stages that contain metrics and parameters, or only those being experimented with (the ones that are run when iterating experiments)?

Trying to hide the trash columns with `--exclude-metrics` allows me to remove some columns, though some others are not recognized. But, in any case, it doesn't remove the strange field values. I include a photo to avoid breaking the format:

![2021-02-11-171417_1895x84_scrot](https://user-images.githubusercontent.com/6652222/107664541-cd44ee00-6c8c-11eb-9863-0a91b054e5f1.png)

### Reproduce

Reproducing is a bit computationally expensive, it requires training a bert model and doing some operations on it's embeddings, it can be accomplished in the following way (at least one GPU with enough RAM, 5-8hrs of computations).

```console
mkdir ./venv
virtualenv ./venv/dvc-reproduce
source ./venv/dvc-reproduce/bin/activate
git clone https://github.com/geblanco/quail-experiments
cd quail-experiments
git checkout restrucutre-experiments
mkdir ~/dvc_cache
dvc config --local core.remote local_dvc
dvc remote add --local -d local_dvc "~/dvc_cache"
dvc repro merge_norm_with_lengths_oversample_05
dvc exp run
```

### Expected

The command should show a table with the parameters tracked by the stage, it's changes and results in metrics.

### Environment information

**Output of `dvc version`:**

```console
$ dvc version
2.0.0a0+02cfe3
```

**Additional Information (if any):**

I have the intuition that it has to do with the parsing of parameters from a file other than `params.yaml`. To be more clear, here is the output in json of the `show` command:

There are various entries for `data/specs/bert-train-no-empty-answers_experiment.json` stated as parameters, with the whole file contents as parameters, although the stage using that file only uses a part of it as parameters (see stage bert-train-no-empty-answers in `dvc.yaml` below)

```console
dvc exp show --show-json
```
```json
{                                                                     
[...]

PR #5541, opened 2021-03-03
params: show only params files that are actually used

### LINKED ISSUE #5539: Inconsistiency in `params.yaml`
While working on #5451 I stumbled upon an inconsistency in our tests.

If we take a look at params test from [here](https://github.com/iterative/dvc/blob/c3a0e8733e799e3c49e5ab6b9a8e32f365553f97/tests/unit/test_params.py#L6):
```python
def test_params_order(tmp_dir, dvc):
    tmp_dir.gen(
        {
            "params.yaml": dumps_yaml({"p1": 1}),
            "p1.yaml": dumps_yaml({"p2": 1}),
            "sub": {"p2.yaml": dumps_yaml({"p3": 1})},
        }
    )

    p2_path = os.path.join("sub", "p2.yaml")
    dvc.stage.add(
        params=[{p2_path: ["p3"]}, {"p1.yaml": ["p2"]}],
        cmd="cmd1",
        name="stage1",
    )
    with (tmp_dir / "sub").chdir():
        dvc.stage.add(params=["p1"], cmd="cmd2", name="stage2")

    # params are sorted during dumping, therefore p1 is first
    assert list(dvc.params.show()[""]) == ["p1.yaml", p2_path, "params.yaml"]
```

We expect that providing `p1` param while in `subdir` will resolve into `{repo_root}/params.yaml:foo`

On the other hand, one of the tests for [experiments](https://github.com/iterative/dvc/blob/c3a0e8733e799e3c49e5ab6b9a8e32f365553f97/tests/func/experiments/test_experiments.py#L490):
```python
def test_subdir(tmp_dir, scm, dvc, workspace):
    subdir = tmp_dir / "dir"
    subdir.gen("copy.py", COPY_SCRIPT)
    subdir.gen("params.yaml", "foo: 1")

    with subdir.chdir():
        dvc.run(
            cmd="python copy.py params.yaml metrics.yaml",
            metrics_no_cache=["metrics.yaml"],
            params=["foo"],
            name="copy-file",
            no_exec=True,
        )
        scm.add(
            [subdir / "dvc.yaml", subdir / "copy.py", subdir / "params.yaml"]
        )
        scm.commit("init")

        results = dvc.experiments.run(
            PIPELINE_FILE, params=["foo=2"], tmp_dir=not workspace
        )
        assert results

    exp = first(results)
    ref_info = first(exp_refs_by_rev(scm, exp))

    fs = scm.get_fs(exp)
    for fname in ["metrics.yaml", "dvc.lock"]:
        assert fs.exists(subdir / fname)
    with fs.open(subdir / "metrics.yaml") as fobj:
        assert fobj.read().strip() == "foo: 2"

    assert dvc.experiments.get_exact_name(exp) == ref_info.name
    assert scm.resolve_rev(ref_info.name) == exp
```
In this case `foo` resolves to `{repo_root}/subdir/params.yaml`

It would be possible that both use cases to be correct, if we expect some kind of "fallback" (if not `params.yaml` in `cwd`, try in root, or previous dir) but that looks to me like bug, because we expose ourselves to risk that introducing `params.yaml` in some subdirectory will change the source of parameters, which will result in indirect influence on some pipeline just by creating file with certain param name inside.

So, I wanted to raise a discussion, whether we assume that there is only single default `{repo_root}/params.yaml`? Or, when we are specifying `--params foo` we expect `{cwd}/params.yaml` to exist and contain `foo`?

pared:

Closes #5539

### DIFF
--- a/tests/unit/test_params.py
+++ b/tests/unit/test_params.py
@@ -6,9 +6,12 @@ from dvc.utils.serialize import dumps_yaml
 def test_params_order(tmp_dir, dvc):
     tmp_dir.gen(
         {
-            "params.yaml": dumps_yaml({"p1": 1}),
+            "params.yaml": dumps_yaml({"p1": 5}),
             "p1.yaml": dumps_yaml({"p2": 1}),
-            "sub": {"p2.yaml": dumps_yaml({"p3": 1})},
+            "sub": {
+                "p2.yaml": dumps_yaml({"p3": 1}),
+                "params.yaml": dumps_yaml({"p1": 1}),
+            },
         }
     )
 
@@ -22,7 +25,11 @@ def test_params_order(tmp_dir, dvc):
         dvc.stage.add(params=["p1"], cmd="cmd2", name="stage2")
 
     # params are sorted during dumping, therefore p1 is first
-    assert list(dvc.params.show()[""]) == ["p1.yaml", p2_path, "params.yaml"]
+    assert list(dvc.params.show()[""]) == [
+        "p1.yaml",
+        p2_path,
+        os.path.join("sub", "params.yaml"),
+    ]
 
 
 def test_repro_unicode(tmp_dir, dvc):
--- a/dvc/repo/params/show.py
+++ b/dvc/repo/params/show.py
@@ -4,7 +4,6 @@ from typing import TYPE_CHECKING, List
 
 from dvc.dependency.param import ParamsDependency
 from dvc.exceptions import DvcException
-from dvc.path_info import PathInfo
 from dvc.repo import locked
 from dvc.repo.collect import collect
 from dvc.scm.base import SCMError
@@ -36,12 +35,6 @@ def _collect_configs(repo: "Repo", rev, targets=None) -> List["DvcPath"]:
         rev=rev,
     )
     path_infos.extend([p.path_info for p in params])
-    if not targets:
-        default_params = (
-            PathInfo(repo.root_dir) / ParamsDependency.DEFAULT_PARAMS_FILE
-        )
-        if default_params not in path_infos:
-            path_infos.append(default_params)
     return list(path_infos)
 
 


tests/unit/test_params.py
```diff
@@ -22,7 +25,11 @@ def test_params_order(tmp_dir, dvc):
         dvc.stage.add(params=["p1"], cmd="cmd2", name="stage2")
```
skshetry, line 25: I wonder why this didn't fail before.

skshetry: Ahh, because it does not check if the dependencies exist.

skshetry (approved):
Looks good to me. But this default behaviour has been there since we introduced `params`, same as #5451 (introduced in #3631), so cc'ing @efiop if this was intended.

efiop:
It was intended originally, but now it is a mixup. Need to take a closer look and also look at the docs, as those might be affected. IIRC the logic was that people might use it without our pipelines (similar to same assumptions/discussions around metrics/plots). I don't feel comfortable just disabling this behavior, because it might turn out that it is needed in some scenario. We might need to double check params workflow first.

skshetry:
@efiop, isn't expecting a `params.yaml` file at the root of a git repository quite an assumption?

efiop:
@skshetry Hard to tell. Params/metrics and plots are all filled with assumptions that were provided by users while developing it, and I also get constantly confused by it. The change seems logical, but we just need to make sure it fits into the workflow and we def need to check the docs.

pared:
@efiop

So the whole issue started in https://github.com/iterative/dvc/issues/5451, where user has quite extensive pipeline, and running `exp show` yielded unwanted results. This change was a prerequisite to make sure only used parameters are shown. 

> Params/metrics and plots are all filled with assumptions that were provided by users while developing it, and I also get constantly confused by it.

I guess we all are. 
I think I jumped the horse here. The way to fix #5451 and not break just released 2.0.0 is to provide new flag that will allow user to show only used params, and keep this issue in mind when releasing 3.0. That way we keep the default behaviour.

pared:
Closing, we do not intend to change default behaviour for now.
Context: 
https://github.com/iterative/dvc/issues/5451#issuecomment-793992639

PR #5554, opened 2021-03-05
gdrive: use the normal file backend for saved credentials

### LINKED ISSUE #5551: gdrive: looking for user credentials when configured with service account
# Bug Report

dvc push looking for user credentials when configured with service account
## Description

I have set up dvc with gdrive, created a google cloud project, and then created a service account according to: https://dvc.org/doc/user-guide/setup-google-drive-remote. I then `dvc add` a file and `dvc push`, and receive an error indicating that my gdrive user credentials file is missing.

here is the trace when running `dvc push -v`:
```
2021-03-04 18:47:25,775 DEBUG: Check for update is enabled.
2021-03-04 18:47:26,472 DEBUG: Preparing to upload data to 'gdrive://1qOaTb9rdhBlUdHeilikWMzAu7PI_KH7m'
2021-03-04 18:47:26,472 DEBUG: Preparing to collect status from gdrive://1qOaTb9rdhBlUdHeilikWMzAu7PI_KH7m
2021-03-04 18:47:26,472 DEBUG: Collecting information from local cache...
2021-03-04 18:47:26,475 DEBUG: Collecting information from remote cache...                                         
2021-03-04 18:47:26,475 DEBUG: Querying 1 hashes via object_exists
2021-03-04 18:47:28,678 DEBUG: GDrive remote auth with config '{'client_config_backend': 'settings', 'client_config_file': 'client_secrets.json', 'save_credentials': True, 'oauth_scope': ['https://www.googleapis.com/auth/drive', 'https://www.googleapis.com/auth/drive.appdata'], 'save_credentials_file': '/home/jaheyer/chipbrain/chipbrain-transformers/.dvc/tmp/gdrive-user-credentials.json', 'get_refresh_token': True, 'save_credentials_backend': 'json', 'service_config': {'client_user_email': 'dvc-remote-storage@dvc-remote-storage-305303.iam.gserviceaccount.com', 'client_json_file_path': '/home/jaheyer/chipbrain/chipbrain-transformers/.dvc/dvc-remote-storage-305303-2331a8a33e40.json'}}'.
2021-03-04 18:47:28,680 ERROR: failed to push data to the cloud - Failed to authenticate GDrive remote: [Errno 2] No such file or directory: '/home/jaheyer/chipbrain/chipbrain-transformers/.dvc/tmp/gdrive-user-credentials.json'
  File "/home/jaheyer/miniconda3/envs/transformers/lib/python3.7/site-packages/dvc/fs/gdrive.py", line 258, in _drive
    gauth.ServiceAuth()
  File "/home/jaheyer/miniconda3/envs/transformers/lib/python3.7/site-packages/pydrive2/auth.py", line 98, in _decorated
    self.LoadCredentials()
  File "/home/jaheyer/miniconda3/envs/transformers/lib/python3.7/site-packages/pydrive2/auth.py", line 329, in LoadCredentials
    self.LoadCredentialsFileJson()
  File "/home/jaheyer/miniconda3/envs/transformers/lib/python3.7/site-packages/pydrive2/auth.py", line 366, in LoadCredentialsFileJson
    filename=credentials_file, scopes=scopes
  File "/home/jaheyer/miniconda3/envs/transformers/lib/python3.7/site-packages/oauth2client/service_account.py", line 219, in from_json_keyfile_name
    with open(filename, 'r') as file_obj:
FileNotFoundError: [Errno 2] No such file or directory: '/home/jaheyer/chipbrain/chipbrain-transformers/.dvc/tmp/gdrive-user-credentials.json'
```
### Reproduce
N/A - don't know how to create a MWE involving service accounts and credentials. 

### Expected

I expect DVC to push my files to the gdrive.
### Environment information

**Output of `dvc version`:**

```console
$ dvc version
DVC version: 2.0.1 (pip)
Platform: Python 3.7.10 on Linux-4.15.0-46-generic-x86_64-with-debian-buster-sid
Supports: gdrive, http, https
Cache directory: ext4 on /dev/sda
Caches: local
Remotes: gdrive
Workspace directory: ext4 on /dev/sda
Repo: dvc, git
```

contents of my .dvc/config:
```
[core]
    remote = storage
['remote "storage"']
    url = gdrive://1qOaTb9rdhBlUdHeilikWMzAu7PI_KH7m
    gdrive_use_service_account = true
    gdrive_service_account_json_file_path = dvc-remote-storage-305303-2331a8a33e40.json
    gdrive_service_account_user_email = dvc-remote-storage@dvc-remote-storage-305303.iam.gserviceaccount.com
```

I have used user credentials to authenticate and push to this drive folder in the past, but since then I have `rm -rf .dvc/`, uninstalled dvc 1.3 and installed dvc 2.0.1

isidentical:
Resolves #5551, ~~awaiting iterative/pydrive2#97~~

### DIFF
--- a/dvc/fs/gdrive.py
+++ b/dvc/fs/gdrive.py
@@ -242,7 +242,9 @@ class GDriveFileSystem(BaseFileSystem):
                 "client_json_file_path": self._service_account_json_file_path,
             }
             if is_credentials_temp:
-                auth_settings["service_config"] = temporary_save_path
+                auth_settings["service_config"][
+                    "client_json_file_path"
+                ] = temporary_save_path
 
         else:
             auth_settings["client_config"] = {
--- a/dvc/fs/gdrive.py
+++ b/dvc/fs/gdrive.py
@@ -3,6 +3,7 @@ import logging
 import os
 import posixpath
 import re
+import tempfile
 import threading
 from collections import defaultdict
 from contextlib import contextmanager
@@ -208,8 +209,17 @@ class GDriveFileSystem(BaseFileSystem):
         from pydrive2.auth import GoogleAuth
         from pydrive2.drive import GoogleDrive
 
-        if os.getenv(GDriveFileSystem.GDRIVE_CREDENTIALS_DATA):
-            with open(self._gdrive_user_credentials_path, "w") as cred_file:
+        temporary_save_path = self._gdrive_user_credentials_path
+        is_credentials_temp = os.getenv(
+            GDriveFileSystem.GDRIVE_CREDENTIALS_DATA
+        )
+        if self._use_service_account:
+            temporary_save_path = os.path.join(
+                tempfile.gettempdir(), "google-creds.json"
+            )
+
+        if is_credentials_temp:
+            with open(temporary_save_path, "w") as cred_file:
                 cred_file.write(
                     os.getenv(GDriveFileSystem.GDRIVE_CREDENTIALS_DATA)
                 )
@@ -231,6 +241,9 @@ class GDriveFileSystem(BaseFileSystem):
                 "client_user_email": self._service_account_user_email,
                 "client_json_file_path": self._service_account_json_file_path,
             }
+            if is_credentials_temp:
+                auth_settings["service_config"] = temporary_save_path
+
         else:
             auth_settings["client_config"] = {
                 "client_id": self._client_id or self.DEFAULT_GDRIVE_CLIENT_ID,
@@ -266,8 +279,8 @@ class GDriveFileSystem(BaseFileSystem):
         except Exception as exc:
             raise GDriveAuthError(self.credentials_location) from exc
         finally:
-            if os.getenv(GDriveFileSystem.GDRIVE_CREDENTIALS_DATA):
-                os.remove(self._gdrive_user_credentials_path)
+            if is_credentials_temp:
+                os.remove(temporary_save_path)
 
         return GoogleDrive(gauth)
 
--- a/dvc/fs/gdrive.py
+++ b/dvc/fs/gdrive.py
@@ -217,6 +217,7 @@ class GDriveFileSystem(BaseFileSystem):
         auth_settings = {
             "client_config_backend": "settings",
             "save_credentials": True,
+            "save_credentials_backend": "file",
             "save_credentials_file": self._gdrive_user_credentials_path,
             "get_refresh_token": True,
             "oauth_scope": [
@@ -226,13 +227,11 @@ class GDriveFileSystem(BaseFileSystem):
         }
 
         if self._use_service_account:
-            auth_settings["save_credentials_backend"] = "json"
             auth_settings["service_config"] = {
                 "client_user_email": self._service_account_user_email,
                 "client_json_file_path": self._service_account_json_file_path,
             }
         else:
-            auth_settings["save_credentials_backend"] = "file"
             auth_settings["client_config"] = {
                 "client_id": self._client_id or self.DEFAULT_GDRIVE_CLIENT_ID,
                 "client_secret": self._client_secret
--- a/setup.py
+++ b/setup.py
@@ -95,7 +95,7 @@ install_requires = [
 # Extra dependencies for remote integrations
 
 gs = ["gcsfs>=0.7.2"]
-gdrive = ["pydrive2>=1.7.3", "six >= 1.13.0"]
+gdrive = ["pydrive2>=1.8.1", "six >= 1.13.0"]
 s3 = ["boto3>=1.9.201"]
 azure = ["adlfs>=0.6.3", "azure.identity>=1.4.0", "knack"]
 # https://github.com/Legrandin/pycryptodome/issues/465


dvc/fs/gdrive.py
```diff
@@ -208,15 +209,25 @@ def _drive(self):
...
 
-        if os.getenv(GDriveFileSystem.GDRIVE_CREDENTIALS_DATA):
-            with open(self._gdrive_user_credentials_path, "w") as cred_file:
+        temporary_save_path = self._gdrive_user_credentials_path
+        is_credentials_temp = os.getenv(
+            GDriveFileSystem.GDRIVE_CREDENTIALS_DATA
+        )
+        if self._use_service_account:
+            temporary_save_path = os.path.join(
+                tempfile.gettempdir(), "google-creds.json"
```
isidentical, line 218: The additional logic is needed because the saved file (cached authorization info) is different than the actual service json. The one that we have secret on the CI is the exported raw credentials json file. But the one pydrive's newer versions expect from this is the cached version. So we have to designate the temporary file as the raw credentials instead of the serialized ones.

efiop (approved):

dvc/fs/gdrive.py
efiop: Oops, didn't realise that this is litterally `/tmp/google-creds.json`. Indeed, that's seems a bit odd.

efiop: @isidentical Can we keep it in `.dvc/tmp` ?

isidentical: I didn't want to depend on `self.repo` in this logic (which is how we get `.dvc/tmp`, `self.repo.tmp_dir`) since we intend to remove `repo` access. Do you know a better way to get this?

dvc/fs/gdrive.py
```diff
@@ -208,15 +209,25 @@ def _drive(self):
         from pydrive2.auth import GoogleAuth
         from pydrive2.drive import GoogleDrive
 
-        if os.getenv(GDriveFileSystem.GDRIVE_CREDENTIALS_DATA):
-            with open(self._gdrive_user_credentials_path, "w") as cred_file:
+        temporary_save_path = self._gdrive_user_credentials_path
+        is_credentials_temp = os.getenv(
+            GDriveFileSystem.GDRIVE_CREDENTIALS_DATA
+        )
+        if self._use_service_account:
```
shcheklein, line 216: `GDRIVE_CREDENTIALS_DATA` this way has two difference meanings in different situations. I think this is confusing, complicates the code, etc. Can we just do the same as we had before? If we don't change anything and just revert the CI variable value, I think it should work? Or cache the saved credentials instead of the raw JSON file?

isidentical: I found that logic a bit weird, considering it would throw a random error when somebody tries to run `GDRIVE_CREDENTIALS_DATA=... dvc status -c` in case of `...` is the raw JSON file.

dvc/fs/gdrive.py
shcheklein: As I mentioned above - I see where it comes from, but I feel we don't need these changes. We can keep everything the way it was before all these changes (cache on CI saved credentials instead of the raw JSON file). Otherwise, if we want to support an env var we should introduce another one, since it has a different semantics.

efiop: @isidentical `tmp_dir` will need to be passed as an option to the fs, it is fine. We might also switch to using proper global directories for it (e.g. using appdirs) in the future.

efiop: @isidentical Wdyt about @shcheklein 's suggestion?

dvc/fs/gdrive.py
shcheklein: yep, I suggest we do a separate env variable - GDRIVE_SERVICE_ACCOUNT_DARA?  It's bad to mix two different meanings into the same variable.

> considering it would throw a random error when somebody tries to run

handle and show a nice message?

dvc/fs/gdrive.py
isidentical: >  cache on CI saved credentials instead of the raw JSON file

Can do it. This would require us to switch the CI variable to a serialized version of this raw file @efiop.

shcheklein: > to a serialized version of this raw file

not raw but the one that GDrive saves into `.dvc/tmp/....`? right? just to be sure we are on the same page...

> Otherwise, if we want to support an env var we should introduce another one, since it has a different semantics.

I still like the idea behind this PR (to make it work with the raw file), @isidentical I think it's fine to create a ticket? It just feels that we need to do it bit differently (var name, path?, change docs, etc)

isidentical: > not raw but the one that GDrive saves into .dvc/tmp/....? right? just to be sure we are on the same page...

I guess we have to find proper names for these 2 different files since it might become very confusing (serialized file / raw file maybe? or cached file / raw file). Yes, that's what I meant by "serialized version of this raw file". 

> I still like the idea behind this PR (to make it work with the raw file), @isidentical I think it's fine to create a ticket? It just feels that we need to do it bit differently (var name, path?, change docs, etc)

My aim was make the the CI work, so I didn't see this as a feature though can create a ticket.

dvc/fs/gdrive.py
isidentical: > handle and show a nice message?

It is an arbitrary key error :( Probably won't even worth the effort though if we want, I guess the best place is to that is wrapping around it in the pydrive2.

shcheklein: Agreed. 

```
 Raises:
            ValueError, if the credential type is not :data:`SERVICE_ACCOUNT`.
            KeyError, if one of the expected keys is not present in
                the keyfile.
```

handle this in the `auth.py` `ServiceAuth()` (?) makes total sense

PR #5582, opened 2021-03-09
scm: add `force` parameter to BaseGitBackend.checkout()

duijf:
The `GitPythonBackend.checkout()` method already implicitly supported this parameter because it forwards `**kwargs` to the backend. `pygit2` didn't support this yet.

This patch fixes this inconsistency.

Let me know if you'd like to see a test for this.

Related to https://github.com/iterative/dvc/issues/2215

### DIFF
--- a/dvc/scm/git/backend/base.py
+++ b/dvc/scm/git/backend/base.py
@@ -67,7 +67,11 @@ class BaseGitBackend(ABC):
 
     @abstractmethod
     def checkout(
-        self, branch: str, create_new: Optional[bool] = False, **kwargs,
+        self,
+        branch: str,
+        create_new: Optional[bool] = False,
+        force: bool = False,
+        **kwargs,
     ):
         pass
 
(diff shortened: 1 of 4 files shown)


pmrowla:
I'll review this tomorrow, but it looks fine at a glance.

Suor:
Note to guys and especially to @duijf: don't rebase and force push it, viewer currently refers to this commit :fearful:

pmrowla (approved):

duijf:
Thanks for the review @pmrowla! 

>  if you need to pass flags into command-line git which aren't implemented in all of the dvc scm backends, you can just use `scm.gitpython` directly (there's already code in the viewer that does this).

Noted! Thanks for the explanation :+1: 

> We'll be dropping the `**kwargs` usage in all of the dvc.scm.Git methods to make it more obvious if this kind of thing happens in the future (instead of silently accepting a kwarg that never gets forwarded into dulwich/pygit).

This seems like a sensible thing to do! We don't want to accidentally rely on implementation details that aren't part of the official API surface.

> viewer currently refers to this commit

Thanks for making this explicit @Suor! Will point the viewer to the commit on `master` now that this has been merged. Will post an update here once addressed. (And will communicate this myself if we ever need to do this again in the future)

Issue #5580, opened 2021-03-09
gdrive: separate cached/serialized data


isidentical:
In GDrive we have 2 types of credential files, one is a regular one that we export from GCP console (a json file) and the other one is the version that automatically gets created by pydrive if the `save_credentials` option is true. They have a different schema so that makes them incompatible. 

Regarding this issue, currently `GDRIVE_CREDENTIALS_DATA` environment variable can be both used for storing user credentials (these are always the 'serialized' ones) and also to store raw json credentials. @shcheklein suggested that we should separate the environment variables where we pass the serialized data and the raw data. This would allow people to use serialized *service* account credentials with `GDRIVE_CREDENTIALS_DATA` and use their raw *service* account credentials with another env var (something like `GDRIVE_SERVICE_DATA`). 

One thing to note in this serialization is that, whether we are going to implement the same in our fixture too (since it is independent). It seems to be that even though the underlying authentication libraries and formats are different they seem to be compatible (with the exception of not being represented as a service account in python level, but I think there is nothing that troubles that) 

tl;dr here is the 2 primary questions (I have a naive patch)
- How should we name the second environment variable that holds *raw credentials*
- Should we implement the same logic to the tests (which I guess it the primary user of this feature?)

isidentical:
CC: @shcheklein @efiop (from the conversation on the #5554)

shcheklein:
> Should we implement the same logic to the tests (which I guess it the primary user of this feature?)

I would add to this that CML (and similar cases when people run DVC in CI)  is the main user for this feature. So it makes it more important than our tests. Good catch on this credentials issue, @isidentical .

And our tests can definitely use the primary JSON file (it's easier for us to manage I think, no need to generate credentials locally?)

> How should we name the second environment variable that holds raw credentials

`GDRIVE_SERVICE_ACCOUNT_DATA` sounds reasonable to me

PR #5583, opened 2021-03-09
gdrive: change the temporary directory location to .dvc/tmp

isidentical:
This doesn't change any behavior regarding our serialization (the issue for that is #5580) but only changes the path so that we would unblock the release.

### DIFF
--- a/dvc/fs/gdrive.py
+++ b/dvc/fs/gdrive.py
@@ -3,7 +3,6 @@ import logging
 import os
 import posixpath
 import re
-import tempfile
 import threading
 from collections import defaultdict
 from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


dvc/fs/gdrive.py
```diff
@@ -127,6 +126,9 @@ def __init__(self, repo, config):
         self._client_id = config.get("gdrive_client_id")
         self._client_secret = config.get("gdrive_client_secret")
         self._validate_config()
+        self._gdrive_service_credentials_path = tmp_fname(
+            os.path.join(self.repo.tmp_dir, "")
```
efiop, line 130: Let's use some prefix to identify it as a one belonging to gdrive?

isidentical: we clean this file, so it doesn't exist after the auth call (also this is what we do for the normal files [non service accounts])

efiop (approved):

Issue #5553, opened 2021-03-05
exp run: reset when the lock file is in Git


dmpetrov:
Currently, the `dvc exp run --reset` option sets the checkpoints files to the latest commited `dvc.lock` (if that file is in Git). In many cases (probably in most of the cases), the training needs to be started from scratch but that seems not possible in the current design.

We should introduce something like `--hard-reset` to start training from scratch even for commited `dvc.lock`?

Also, I'm wondering about inconsistency between commited `dvc.lock` and not commited. It looks like `--reset` has a different behavior. Is there a way to unify it? The `--hard-reser` option does not unify it and I'm not sure it is the best way to solve the problem.

dberenbaum:
> Also, I'm wondering about inconsistency between commited `dvc.lock` and not commited. It looks like `--reset` has a different behavior. Is there a way to unify it? The `--hard-reser` option does not unify it and I'm not sure it is the best way to solve the problem.

Could you be more specific with this point? What is the inconsistency/different behavior you see?

dmpetrov:
@dberenbaum from user's point of view, `dvc exp run --reset` has a different behavior when `dvc.lock` is commited and when it is not.

An example: if I need to start training from scratch is not enough to just run `dvc exp run --reset`. I need to check if lock file is commited or not.

After more thinking about this scenario... It feels like removing model and starting training from scratch is more frequent use case then starting from the previous commit. Instead of introducing `--hard-reset` we can change the default behavior to model removal while introduce `--reset-to-previous` for the case when the latest commited model is needed.

dberenbaum:
It seems reasonable for the default `--reset` to always run the experiment from scratch (as if the stage has never been run before). One concern I would have is to make sure it doesn't just drop `dvc.lock` in case there are other pipelines that are not part of the experiment.

Rather than have a `--reset-to-previous` flag, maybe it makes more sense to have an option to drop all experiments in the workspace? Something like `dvc exp remove --workspace`. People may want to clear their workspace and start over, even in experiments without checkpoints. I have run into lots of times where I thought, "Whoops, didn't mean to do that, let me get back to a clean workspace." See also #5437.

Thoughts @pmrowla ?

dmpetrov:
> default `--reset` to always run the experiment from scratch

Yes, it seems like the best option.


> Something like `dvc exp remove --workspace`

[...]

Issue #5567, opened 2021-03-08
exp run: multiple error messages


skshetry:
I had run this experiment before already. But now, it provides same error messages twice (from two different workers I think), and makes the output very ugly.

And even though it has failed, it suggests me to track files and also updates lock files, nothing wrong with it but is questionable.

```console
$ dvc exp run --set-param foo=13 -v
2021-03-08 17:34:44,409 DEBUG: Check for update is disabled.
2021-03-08 17:34:44,513 DEBUG: Detaching HEAD at 'HEAD'               
2021-03-08 17:34:44,515 DEBUG: Stashing workspace
2021-03-08 17:34:44,523 DEBUG: Stashing changes in 'refs/stash'
2021-03-08 17:34:44,537 DEBUG: Applying stash commit 'e9fc4a1adbb971cbf0dc9dad959f9a1186ab82b8'
2021-03-08 17:34:44,704 DEBUG: Using experiment params 'defaultdict(<class 'dict'>, {'params.yaml': {'foo': 13}})'
2021-03-08 17:34:44,716 DEBUG: Stashing changes in 'refs/exps/stash'
2021-03-08 17:34:44,740 DEBUG: Stashed experiment 'bfc3174' with baseline 'f9d0f0a' for future execution.
2021-03-08 17:34:44,742 DEBUG: Restoring stashed workspace
2021-03-08 17:34:44,742 DEBUG: Popping from stash 'refs/stash'
2021-03-08 17:34:44,742 DEBUG: Applying stash commit 'e9fc4a1adbb971cbf0dc9dad959f9a1186ab82b8'
2021-03-08 17:34:44,750 DEBUG: Dropping 'refs/stash@{0}'
2021-03-08 17:34:44,755 DEBUG: Removing '/tmp/tmp.WTJ1L9Of5F/.git/logs/refs/stash'
2021-03-08 17:34:44,757 DEBUG: Restore HEAD to 'master'
2021-03-08 17:34:44,760 DEBUG: Detaching HEAD at 'f9d0f0af0f3884238997ba7a1415f147fd3341e6'
2021-03-08 17:34:44,771 DEBUG: Removing 'dvc.lock'
2021-03-08 17:34:44,771 DEBUG: Popping from stash 'refs/exps/stash'
2021-03-08 17:34:44,772 DEBUG: Applying stash commit 'bfc31747adac7ade25bb20938a0cbd678c317336'
[...]

PR #5586, opened 2021-03-10
checkpoints: completely remove checkpoint outs on `exp run --reset`

pmrowla:

Will close https://github.com/iterative/dvc/issues/5553.

Docs PR: https://github.com/iterative/dvc.org/pull/2286

Adjusts `dvc exp run --reset` behavior for `checkpoint` outs:

- When `--reset` is used, any `checkpoint` outs will be removed entirely before running the experiment (regardless of whether or not a hash for the given out exists in `dvc.lock`)
- When `--reset` is used, all other (non-checkpoint) outs in `dvc.lock` will be unchanged
    - Previously, `--reset` would reset the entire `dvc.lock` file to HEAD (i.e `git checkout HEAD -- dvc.lock`)
- `--queue` now explicitly implies `--reset`, and `--reset` is now mutually exclusive with `--rev`
    - This was essentially already the case but we now explicitly error out if `--reset` is used in conjunction with `--rev`
    - Previously, using `--rev` would do the equivalent of `git reset --hard <rev>` (so the old behavior would do the same thing with or with `--reset`).
    - With the new behavior, it does not make sense to delete a checkpoint out at the same time as resuming from that checkpoint, so it's now an error condition
    - The docs for `--queue/--run-all` with regard to checkpoints have also been clarified

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -514,6 +514,11 @@ class CmdExperimentsRun(CmdRepro):
     def run(self):
         from dvc.command.metrics import _show_metrics
 
+        if self.args.reset and self.args.checkpoint_resume:
+            raise InvalidArgumentError(
+                "--reset and --rev are mutually exclusive."
+            )
+
         if self.args.reset:
             logger.info("Any existing checkpoints will be reset and re-run.")
 
--- a/dvc/output/base.py
+++ b/dvc/output/base.py
@@ -408,6 +408,7 @@ class BaseOutput:
         relink=False,
         filter_info=None,
         allow_missing=False,
+        checkpoint_reset=False,
         **kwargs,
     ):
         if not self.use_cache:
@@ -420,6 +421,11 @@ class BaseOutput:
         obj = self.get_obj(filter_info=filter_info)
         if not obj and (filter_info and filter_info != self.path_info):
             # backward compatibility
+            return None
+
+        if self.checkpoint and checkpoint_reset:
+            if self.exists:
+                self.remove()
             return None
 
         added = not self.exists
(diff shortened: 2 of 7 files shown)


pmrowla:
@dberenbaum I tested the new behavior against some of the toy checkpoints repos I have on hand (in addition to the CI tests), but would appreciate it if you can double check with a couple of your own repos as well.

dberenbaum (commented):
If there are checkpoints prior to HEAD, `--reset` will not drop the existing checkpoints, but it will create a new experiment from scratch and duplicate those checkpoints. Is it important that `--reset` drops those checkpoints to avoid duplication? Will it seem inconsistent that checkpoints are dropped when based off HEAD but otherwise are not?

pmrowla:
> If there are checkpoints prior to HEAD, `--reset` will not drop the existing checkpoints, but it will create a new experiment from scratch and duplicate those checkpoints. Is it important that `--reset` drops those checkpoints to avoid duplication? Will it seem inconsistent that checkpoints are dropped when based off HEAD but otherwise are not?

Experiments (checkpoints or not) are always considered unique per parent HEAD commit. Experiments derived from any other commit are considered separate, so we don't check to see if a matching experiment tied to any other commit in the entire git history. (But if an identical run does exist, we will still re-use run-cache for that experiment so execution would not be duplicated, and we would also re-use identical cache objects like with anything else in DVC)

As it stands right now, checkpoints behave the same way as standalone experiments, making `--reset` search for an identical checkpoint through the reset of the git history seems like it would be unnecessary (and probably unexpected behavior as well?)

dberenbaum (approved):
I'm fine with the behavior as is, but we might have to revisit this workflow if users get confused. Both this and #5567 are related to running experiments with different Git histories but matching dvc-tracked info.

pmrowla:
#5567 is bug that is not related to checkpoints

Issue #5595, opened 2021-03-11
exp run: --temp crashes


jorgeorpinel:
# Bug Report

## Description

`exp run --temp` throws a Python error

### Reproduce

```console
$ git clone git@github.com:iterative/example-get-started.git
$ cd example-get-started
$ dvc pull
$ dvc exp run --temp
'data/data.xml.dvc' didn't change, skipping
Stage 'prepare' didn't change, skipping
Stage 'featurize' didn't change, skipping
Stage 'train' didn't change, skipping
Stage 'evaluate' didn't change, skipping
Exception ignored in: <finalize object at 0x7ff0a023d6b0; dead>
Traceback (most recent call last):
  File "/usr/lib/python3.6/weakref.py", line 548, in __call__
    return info.func(*info.args, **(info.kwargs or {}))
  File "/usr/lib/python3.6/tempfile.py", line 938, in _cleanup
    _rmtree(name)
  File "/usr/lib/python3.6/shutil.py", line 477, in rmtree
    onerror(os.lstat, path, sys.exc_info())
  File "/usr/lib/python3.6/shutil.py", line 475, in rmtree
    orig_st = os.lstat(path)
FileNotFoundError: [Errno 2] No such file or directory: '/home/jop/DVC-repos/example-get-started/.dvc/tmp/exps/tmpnu_88y_x'
```

### Expected

Per 1. in #5593, it should generate an experiment (identical to `HEAD` in this case).

### Environment information

```console
$ dvc doctor
DVC version: 2.0.5
---------------------------------
Platform: Python 3.6.9 on Linux-5.4.72-microsoft-standard-WSL2-x86_64-with-Ubuntu-18.04-bionic
Supports: All remotes
Cache types: hardlink, symlink
Cache directory: ext4 on /dev/sdb
Caches: local
Remotes: https
Workspace directory: ext4 on /dev/sdb
Repo: dvc, git
```

pmrowla:
I'm able to reproduce this in Python 3.6 on linux but not in Python 3.8

```
(.venv-linux-38) vagrant@ubuntu-bionic:~/example-get-started$ dvc exp run --temp
'data/data.xml.dvc' didn't change, skipping
Stage 'prepare' didn't change, skipping
Stage 'featurize' didn't change, skipping
Stage 'train' didn't change, skipping
Stage 'evaluate' didn't change, skipping
(.venv-linux-38) vagrant@ubuntu-bionic:~/example-get-started$ dvc doctor
[...]

Issue #5478, opened 2021-02-16
exp: cleanup error during finalization


pared:
When running experiments in `tmp`, we get `FileNotFoundError` during cleanup.
```
#!/bin/bash

main=$(pwd)
rm -rf repo
mkdir repo
pushd repo

git init --quiet
dvc init --quiet

git add -A
git commit -am "init"
echo "param: 2" > params.yaml
echo "data" > data
dvc add data

dvc run -d data --params param -n train -o out 'cat params.yaml > out'
git add -A
git commit -am "run"

dvc exp run train -S param=3 --queue
dvc exp run --run-all
```

Result:
```
Queued experiment '17c7e16' for future execution.                     
'data.dvc' didn't change, skipping                                                                                                                
Running stage 'train':                                                                                                                            
> cat params.yaml > out
Updating lock file 'dvc.lock'                                                                                                                     
Exception ignored in: <finalize object at 0x7fb899ec3410; dead>                                                                                   
Traceback (most recent call last):
  File "/home/prd/development/venvs/dvc3.7/lib/python3.7/weakref.py", line 572, in __call__
    return info.func(*info.args, **(info.kwargs or {}))
  File "/home/prd/development/venvs/dvc3.7/lib/python3.7/tempfile.py", line 934, in _cleanup
    _rmtree(name)
  File "/home/prd/development/venvs/dvc3.7/lib/python3.7/shutil.py", line 485, in rmtree
    onerror(os.lstat, path, sys.exc_info())
  File "/home/prd/development/venvs/dvc3.7/lib/python3.7/shutil.py", line 483, in rmtree
    orig_st = os.lstat(path)
FileNotFoundError: [Errno 2] No such file or directory: '/home/prd/development/playground/5741_exp_ext_cache_symlink/repo/.dvc/tmp/exps/tmpjco8_3pa'

Reproduced experiment(s): exp-2f709
To apply the results of an experiment to your workspace run:

	dvc exp apply <exp>

To promote an experiment to a Git branch run:

	dvc exp branch <exp>
```
Expected result:

no error message

dvc version:
```
DVC version: 2.0.0a0+66f24b 
---------------------------------
Platform: Python 3.7.5 on Linux-5.4.0-65-generic-x86_64-with-Ubuntu-18.04-bionic
Supports: All remotes
```

pared:
Also, error shows up a lot while running `test_experiments`

pmrowla:
[...]

PR #5425, opened 2021-02-08
exp: cleanup executor temp directories ourselves

pmrowla:
* [x] ❗ I have followed the [Contributing to DVC](https://dvc.org/doc/user-guide/contributing/core) checklist.

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here.

Thank you for the contribution - we'll try to review it as soon as possible. 🙏

Will fix #5374

PR #5598, opened 2021-03-11
exp: subclass tempfile.TemporaryDirectory to handle cleanup

pmrowla:

Will close https://github.com/iterative/dvc/issues/5595
Will close https://github.com/iterative/dvc/issues/5478

After the changes to use `dvc.utils.fs.remove` instead of `TemporaryDirectory.cleanup()` in https://github.com/iterative/dvc/pull/5425, on Python <3.8 on Linux, a skipped/ignored exception message would be logged and visible during DVC runs (with or without `-v`). This happened when `cleanup()` was called on deletion of the tempdir object - the directory itself had already been manually removed by DVC, and `FileNotFoundError` is logged (but not re-raised) to note that `cleanup()` expected to remove a directory which no longer exists. In other python versions, the ignored exception is just silently handled (since it's not actually an error condition).

After this PR we just subclass `TemporaryDirectory` to use our `remove` so that `cleanup()` works as expected (without logging ignored exceptions) whether it is called manually or on object deletion.

### DIFF
--- a/dvc/repo/experiments/executor/local.py
+++ b/dvc/repo/experiments/executor/local.py
@@ -3,11 +3,21 @@ import os
 from tempfile import TemporaryDirectory
 from typing import Optional
 
-from dvc.utils.fs import remove
-
 from .base import BaseExecutor
 
 logger = logging.getLogger(__name__)
+
+
+class ExpTemporaryDirectory(TemporaryDirectory):
+    # Python's TemporaryDirectory cleanup shutil.rmtree wrapper does not handle
+    # git read-only dirs cleanly in Windows, so we use our own remove()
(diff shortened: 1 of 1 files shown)


skshetry:
Looks like this was fixed in [Python3.8](https://bugs.python.org/issue26660).

pmrowla:
@skshetry  Yeah, the original issue was resolved by just using `fs.remove()` to manually clean up the temp dir instead of relying on `TemporaryDirectory.cleanup()` (since our `remove` handles read-only files/dirs properly).

The follow up issue is that if you manually remove the temp directory, it apparently ends up being logged in a strange way on Linux in Python <3.8

PR #5599, opened 2021-03-11
stage add: unsuppress -c/--checkpoints help text

pmrowla:

Related to https://github.com/iterative/dvc.org/issues/2253

docs PR: https://github.com/iterative/dvc.org/pull/2290

### DIFF
--- a/dvc/command/stage.py
+++ b/dvc/command/stage.py
@@ -255,7 +255,8 @@ def _add_common_args(parser):
         "--checkpoints",
         action="append",
         default=[],
-        help=argparse.SUPPRESS,
+        help="Declare output file or directory that will be be handled as an "
+        "intermediate checkpoint upon exp run.",
         metavar="<filename>",
     ).complete = completion.FILE
     parser.add_argument(
--- a/dvc/command/stage.py
+++ b/dvc/command/stage.py
@@ -255,8 +255,8 @@ def _add_common_args(parser):
         "--checkpoints",
         action="append",
         default=[],
-        help="Declare output file or directory that will be be handled as an "
-        "intermediate checkpoint upon exp run.",
+        help="Declare checkpoint output file or directory for 'dvc exp run'. "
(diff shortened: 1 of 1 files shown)


dvc/command/stage.py
```diff
@@ -255,7 +255,8 @@ def _add_common_args(parser):
         "--checkpoints",
         action="append",
         default=[],
-        help=argparse.SUPPRESS,
+        help="Declare output file or directory that will be be handled as an "
+        "intermediate checkpoint upon exp run.",
```
dberenbaum, line 259: Any reason to leave out "Note that checkpoints are not supported by `dvc repro`."?

dberenbaum (approved):
One small question, but looks good!

dvc/command/stage.py
jorgeorpinel: Does `repro` error-out or just ignore them?

dberenbaum: Errors out:

```
ERROR: failed to reproduce 'dvc.yaml': Checkpoint stages are not supported in 'dvc repro'. Checkpoint stages must be reproduced with 'dvc exp run' or 'dvc exp resume'.
```

jorgeorpinel:
How does this work, something like

```console
$ dvc stage add -n mystage -d mydata -d myout1 -c myout2 ...
```

? I.e. `-c` replaces `-o`

dvc/command/stage.py
jorgeorpinel: OK my suggested text: ``"Declare checkpoint output (file or directory) for `dvc exp run`. Not compatible with `dvc repro`"``

pmrowla:
> How does this work, something like
> 
> ```
> $ dvc stage add -n mystage -d mydata -o myout1 -c myout2 -M ...
> ```
> 
> ? I.e. `-c` replaces `-o`

Yes, it works the same way as the metrics flags or `--outs-persist` for generating a specific type of output

Issue #5593, opened 2021-03-11
exp run: checks out data dependencies (destructive) [QA]


jorgeorpinel:
### ~~1. It produces a first unchanged exp~~
UPDATE: Addressed in #5600

```console
$ git clone git@github.com:iterative/example-get-started.git
$ cd example-get-started
$ dvc pull

$ dvc exp run
```

This works even when there are no changes to the committed project version (`HEAD`). Below we can see there are differences in metrics or params:

```
$  dvc exp show --no-pager
┏━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━
┃ Experiment    ┃ Created      ┃ avg_prec ┃ roc_auc ┃ prepare.split ┃ prepare.seed ┃ featurize.max_features ┃ featurize.ngrams ┃ train.seed ┃ train.n_est ┃
┡━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━
│ workspace     │ -            │  0.60405 │  0.9608 │ 0.2           │ 20170428     │ 3000                   │ 2                │ 20170428   │ 100         │
│ master        │ Mar 01, 2021 │  0.60405 │  0.9608 │ 0.2           │ 20170428     │ 3000                   │ 2                │ 20170428   │ 100         │
│ └── exp-44136 │ 02:22 AM     │  0.60405 │  0.9608 │ 0.2           │ 20170428     │ 3000                   │ 2                │ 20170428   │ 100         │
└───────────────┴──────────────┴──────────┴─────────┴───────────────┴──────────────┴────────────────────────┴──────────────────┴────────────┴─────────────┴─
```
> `exp diff` doesn't print anything.

Is there a use case for this? Otherwise I'd vote to block it.

### 2. It checks out data dependencies (destructive)

> Continuing the previous CLI block

```console
...
[...]

PR #5600, opened 2021-03-11
exp run: do nothing if result matches parent (HEAD)

pmrowla:

Related to https://github.com/iterative/dvc/issues/5593

If the result of `exp run` is identical to the parent commit (HEAD), do nothing. Previously, for `--temp` runs we would do nothing, but workspace runs would generate an empty named experiment which was identical to HEAD.

### DIFF
--- a/dvc/repo/experiments/__init__.py
+++ b/dvc/repo/experiments/__init__.py
@@ -725,10 +725,14 @@ class Experiments:
                     rel_cwd=relpath(os.getcwd(), self.scm.root_dir),
                 )
 
-                if not exec_result.exp_hash or not exec_result.ref_info:
+                if not exec_result.exp_hash:
                     raise DvcException(
                         f"Failed to reproduce experiment '{rev[:7]}'"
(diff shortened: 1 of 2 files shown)


skshetry:
@pmrowla, does this also solve #5567?

pmrowla:
@skshetry possibly? There was a bug with workspace runs where the workspace was considered dirty because of the way we pass repro args into the experiment (even though the end result would still be an empty git commit) that should be resolved by this PR.

That may be what was causing the "conflicts with existing experiment" issue you were seeing in that issue

skshetry:
@Mergifyio rebase

skshetry:
Looks like not.

```bash
cd "$(mktemp -d)" && git init && dvc init
dvc stage add -n echo -p foo -m metrics.yaml "echo params.yaml > metrics.yaml"
echo "foo: 10" > params.yaml
git add -A && git commit -m "initialize"
dvc exp run
dvc exp run
```

PR #5607, opened 2021-03-11
dvc: refactor objects

efiop:
DirInfo was created before we had proper objects, as a way to have a
structure that represents a staged directory. Right now we have a
dedicated Tree object that is specifically meant for that.

As a result, HashInfo no longer needs to have dir_info and we no longer
need `get_hash/get_dir_hash` helpers.

Still needs a few more smaller followups to finish migration to objects.


dvc/api.py
```diff
@@ -25,12 +24,20 @@ def get_url(path, repo=None, rev=None, remote=None):
...
-        if not metadata.is_dvc:
+        if (
+            not metadata.is_dvc
+            or len(metadata.outs) != 1
+            or not metadata.outs[0].hash_info
+        ):
             raise OutputNotFoundError(path, repo)
 
         cloud = metadata.repo.cloud
-        hash_info = get_hash(path_info, _repo.repo_fs, "md5", _repo.odb)
```
efiop, line 32: This was actively trying to compute a hash, which is clearly not something that `get_url` should do. What we really needed is a recorded md5, which is now easy to get from `dvcfs.info()`(part of fsspec).

dvc/api.py
```diff
@@ -25,12 +24,20 @@ def get_url(path, repo=None, rev=None, remote=None):
...
-        return cloud.get_url_for(remote, checksum=hash_info.value)
+        md5 = metadata.repo.dvcfs.info(path_info).get("md5")
+
+        if not md5:
+            raise OutputNotFoundError(path, repo)
```
efiop, line 38: Need to raise some other exc, this is not suitable.

PR #5611, opened 2021-03-12
exp run: fix issue where duplicate workspace runs would incorrectly conflict

pmrowla:

Will fix https://github.com/iterative/dvc/issues/5567

### DIFF
--- a/dvc/repo/experiments/executor/base.py
+++ b/dvc/repo/experiments/executor/base.py
@@ -218,6 +218,7 @@ class BaseExecutor(ABC):
         queue: Optional["Queue"] = None,
         rel_cwd: Optional[str] = None,
         name: Optional[str] = None,
+        log_errors: bool = True,
         log_level: Optional[int] = None,
     ) -> "ExecutorResult":
         """Run dvc repro and return the result.
@@ -234,7 +235,7 @@ class BaseExecutor(ABC):
 
         if queue is not None:
             queue.put((rev, os.getpid()))
-        if log_level is not None:
+        if log_errors and log_level is not None:
             cls._set_log_level(log_level)
 
         def filter_pipeline(stages):
@@ -246,7 +247,7 @@ class BaseExecutor(ABC):
         exp_ref: Optional["ExpRefInfo"] = None
         repro_force: bool = False
 
-        with cls._repro_dvc(dvc_dir, rel_cwd) as dvc:
+        with cls._repro_dvc(dvc_dir, rel_cwd, log_errors) as dvc:
             args, kwargs = cls._repro_args(dvc)
             if args:
                 targets: Optional[Union[list, str]] = args[0]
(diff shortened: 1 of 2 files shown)


dvc/repo/experiments/executor/base.py
```diff
@@ -406,16 +412,31 @@ def commit(
...
         scm.commit(f"dvc: commit experiment {exp_hash}", no_verify=True)
         new_rev = scm.get_rev()
-        scm.set_ref(branch, new_rev, old_ref=old_ref)
+        if check_exists:
+            orig_rev = scm.get_ref(branch)
+            if scm.diff(scm.get_ref(branch), new_rev):
+                if checkpoint:
+                    raise CheckpointExistsError(ref_info.name)
+                raise ExperimentExistsError(ref_info.name)
+            new_rev = orig_rev
```
pmrowla, line 437: If the new run generates an identical commit to an existing one, we should be reusing the existing commit (this logic was already happening for tempdir runs during `git fetch`, but not for workspace runs where we commit directly into the main git/dvc workspace)

skshetry: @pmrowla, is it possible to  find out before even running this, that the state is the same as before (might be problematic  with non-deterministic stages probably).

skshetry, line 437: These `if` branches are a bit hard to read.

dvc/repo/experiments/executor/base.py
```diff
@@ -406,16 +412,31 @@ def commit(
...
+            if scm.get_ref(branch):
+                if not force:
+                    check_exists = True
+                    logger.debug(
+                        "Reuse existing experiment branch '%s'", branch
+                    )
+                else:
+                    logger.debug(
+                        "Replace existing experiment branch '%s'", branch
+                    )
```
skshetry, line 424:
```suggestion
                logger.debug(
                    "%s existing experiment branch '%s'", "Replace" if force else "Reuse", branch
                )
```

skshetry (approved):
Looks good to me. I have made some minor comments inline.

dvc/repo/experiments/executor/base.py
pmrowla: I think it's difficult because we also include git repo/workspace modifications with experiments (and not just DVC dependencies). So while two pipeline runs might be identical (so they get hashed into matching stages with matching DVC-tracked deps/outs), there could be other changes in the repo that show up in git but not to DVC. So we have to also generate the final git commit and then see if that actually conflicts/diffs with the previous run.

If there are git differences, we can't really tell which experiment should be preferred (so we would error out and then require running with `-f/--force` to overwrite the existing one)

pmrowla: And yeah, also as you noted, I'm not sure we can rely on checkpoint stages to be deterministic, since they are persist outputs and it's not necessarily guaranteed that the user's code will always generate the identical sequence of checkpoints

PR #5631, opened 2021-03-15
update: use unresolved path with --to-remote calls

### LINKED ISSUE #5629: update: --to-remote doesn't use the specified config options
This is caught while I was doing the migration to s3fs

```py
# start 'moto3_server -p 6666'

git init
dvc init
mkdir /tmp/upstream

dvc remote add -d upstream /tmp/upstream
dvc remote add workspace s3://test-bucket-5
dvc remote modify workspace endpointurl http://127.0.0.1:6666/

dvc import-url remote://workspace/foo.txt --to-remote
dvc update foo.txt.dvc --to-remote
```

When doing `dvc update --to-remote`, it doesn't pass extra config options to the file system as a config option. It only passes the url.

isidentical:
Resolves #5629.

### DIFF
--- a/dvc/stage/imports.py
+++ b/dvc/stage/imports.py
@@ -12,7 +12,7 @@ def _update_import_on_remote(stage, remote, jobs):
             "be updated with --to-remote"
         )
 
-    url = stage.deps[0].path_info.url
+    url = stage.deps[0].def_path
     stage.outs[0].hash_info = stage.repo.cloud.transfer(
         url, jobs=jobs, remote=remote, command="update"
     )


isidentical:
I thought about adding a test though I don't think there is an easy way to verify this (considering each platform have different options and it would be hard to verify this works as is). Anyways, this will be tested by s3fs since it will be depending on it.

isidentical:
https://github.com/iterative/dvc/blob/4ed55d00511ea3d9115b76c463e1a466408b11ef/dvc/output/base.py#L123-L124

isidentical:
restarting CI....

efiop (approved):
Good catch. This will do for now, but we'll be refactoring it a bit soon with objects.

PR #5632, opened 2021-03-16
Using existing logging level if defined in addLoggingLevel

### LINKED ISSUE #4546: DVC v.1.6.6 cannot be imported after Bokeh v.22.1 as both uses logging levelName = "TRACE"
## Bug Report
It has been a long time since I updated my DVC version - previously, I have used version 0.84.0 - but I have now updated to version 1.6.6. However, I get an assert error, when I first import Bokeh, before I import DVC in my python script - as seen in the image below (it fails both with `import dvc` and `import dvc.api`).
dvc error

The assert shows that both DVC and Bokeh creates a loggingLevel and loggerClass, named "TRACE" and "trace", respectively.
I think DVC should handle - i.e. not fail at - this edge case, where TRACE already have been defined by another Python package or user code before DVC is imported.

PS: Thanks for the big improvements all DVC developer have done since version 0.84.0 - I'm so excited to use all of the new features 😄

**Output of `dvc version`:**

```console
$ dvc version
DVC version: 1.6.6 (conda)
Platform: Python 3.8.2 on Linux-5.4.0-45-generic-x86_64-with-glibc2.10
Supports: azure, gdrive, gs, hdfs, http, https, s3, ssh, oss
```

asford:

fixes #4546

### DIFF
--- a/dvc/logger.py
+++ b/dvc/logger.py
@@ -23,14 +23,19 @@ def addLoggingLevel(levelName, levelNum, methodName=None):
     Adds a new logging level to the `logging` module and the
     currently configured logging class.
 
+    Uses the existing numeric levelNum if already defined.
+
     Based on https://stackoverflow.com/questions/2183233
     """
     if methodName is None:
         methodName = levelName.lower()
 
-    assert not hasattr(logging, levelName)
-    assert not hasattr(logging, methodName)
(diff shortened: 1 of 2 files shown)


efiop (approved):
Thank you @asford ! :pray:

PR #5633, opened 2021-03-16
gs: process etag when it is available

isidentical:
Directories doesn't have etag entries, so we need to skip processing them on `_entry_hook`

### DIFF
--- a/dvc/fs/gs.py
+++ b/dvc/fs/gs.py
@@ -31,7 +31,8 @@ class GSFileSystem(FSSpecWrapper):
         return login_info
 
     def _entry_hook(self, entry):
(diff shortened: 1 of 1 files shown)


isidentical:
Moved to an upstream PR #5634

skshetry:
Btw, @isidentical, are you aware of `--enable-gs` flags in the pytest? I am not sure if we have shared this with you.

isidentical:
> Btw, @isidentical, are you aware of --enable-gs like flags in the pytest? I am not sure if we have shared this with you.

I use `--all` with `-k '[gs]'` though that seem to be a better solution

PR #5634, opened 2021-03-16
gs: process etag when it is available

isidentical:
Directories doesn't have etag entries, so we need to skip processing them on _entry_hook

### DIFF
--- a/dvc/fs/gs.py
+++ b/dvc/fs/gs.py
@@ -31,7 +31,8 @@ class GSFileSystem(FSSpecWrapper):
         return login_info
 
     def _entry_hook(self, entry):
-        entry["etag"] = base64.b64decode(entry["etag"]).hex()
+        if "etag" in entry:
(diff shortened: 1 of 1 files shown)


skshetry (approved):
Thanks a lot for the prompt fix @isidentical. Looks good to me.

efiop:
@isidentical Is it something that should be fixed upstream in gcfs, btw?

jorgeorpinel:
What hash is used for dirs on GCS? May need a quick update in https://dvc.org/doc/user-guide/project-structure/dvc-files#dependency-entries (GS is still supported as [external dependencies](https://dvc.org/doc/user-guide/external-dependencies) right?)

jorgeorpinel:
> May need a quick update in https://dvc.org/doc/user-guide/project-structure/dvc-files#dependency-entries

Doesn't even mention GCS atm... I think that was removed recently by @efiop and I keep getting confused about this sry 😅

PR #5652, opened 2021-03-19
diff: clear calculated renamed files from inverted hashmap

### LINKED ISSUE #5597: ERROR: unexpected error - list.remove(x): x not in list, if I execute "dvc diff HEAD~",  after  dvc version is updated from 1.1.15 to 2.0.5
# Bug Report

ERROR: unexpected error - list.remove(x): x not in list, if I execute "dvc diff HEAD~", after dvc version is updated from 1.1.15 to 2.0.5

## Description

"dvc diff" command is wrong in version 2.0.5.

### Reproduce

Step list of how to reproduce the bug

1. I made dvc repository in 1.11.15 and push data into GCS.
2. "pip install dvc --upgrade"
3. Create a file in a directory named foo managed by dvc
4. "dvc add foo"
5. "dvc diff HEAD~"
6. shown "ERROR: unexpected error - list.remove(x): x not in list"

### Expected

"dvc diff" command execute as well as 1.11.15.
looks like below.
```
Added:
    bar.txt
```

### Environment information

**Output of `dvc doctor`:**

```console
$ dvc doctor
```

```console
DVC version: 2.0.5 (pip)
Platform: Python 3.7.9 on Darwin-19.6.0-x86_64-i386-64bit
Supports: gs, hdfs, webhdfs, http, https
Cache directory: apfs on /dev/disk1s2
Caches: local
Remotes: gs
Workspace directory: apfs on /dev/disk1s2
Repo: dvc, git
```

part of error message of ```dvc diff --verbose HEAD~``` is below.
```
  File "venv/lib/python3.7/site-packages/dvc/main.py", line 50, in main
    ret = cmd.run()
  File "<reposistory>/venv/lib/python3.7/site-packages/dvc/command/diff.py", line 138, in run
    self.args.a_rev, self.args.b_rev, self.args.targets
  File "<reposistory>/venv/lib/python3.7/site-packages/dvc/repo/__init__.py", line 49, in wrapper
  File "<reposistory>/venv/lib/python3.7/site-packages/dvc/repo/diff.py", line 79, in diff
    deleted.remove(renamed_item["path"]["old"])
ValueError: list.remove(x): x not in list
```

version 2.0.0 is wrong too.
version 1.11.16 is correct.

Thanks.

skshetry:
We were trying to reuse the first encounter of the file from the inverted hashmap again and again, which was throwing ValueError.

Fixes #5597

### DIFF
--- a/dvc/repo/diff.py
+++ b/dvc/repo/diff.py
@@ -1,5 +1,7 @@
 import logging
 import os
+from collections import defaultdict
+from typing import Dict, List
 
 from dvc.exceptions import PathMissingError
 from dvc.objects.stage import get_file_hash
@@ -190,26 +192,19 @@ def _targets_to_path_infos(repo_fs, targets):
 
 
 def _calculate_renamed(new, old, added, deleted):
-    old_inverted = {}
+    old_inverted: Dict[str, List[str]] = defaultdict(list)
     # It is needed to be dict of lists to cover cases
     # when repo has paths with same hash
     for path, path_hash in old.items():
-        bucket = old_inverted.get(path_hash, None)
-        if bucket is None:
-            old_inverted[path_hash] = [path]
-        else:
-            bucket.append(path)
+        old_inverted[path_hash].append(path)
 
     renamed = []
     for path in sorted(added, key=len):
         path_hash = new[path]
-        old_paths = old_inverted.get(path_hash, None)
-        if not old_paths:
-            continue
-
+        old_paths = old_inverted[path_hash]
         try:
             iterator = enumerate(old_paths)
-            index = next((idx for idx, path in iterator if path in deleted))
+            index = next(idx for idx, path in iterator if path in deleted)
         except StopIteration:
             continue
 
--- a/dvc/repo/diff.py
+++ b/dvc/repo/diff.py
@@ -199,7 +199,7 @@ def _calculate_renamed(new, old, added, deleted):
         old_inverted[path_hash].append(path)
 
     renamed = []
-    for path in sorted(added, key=len):
+    for path in added:
         path_hash = new[path]
         old_paths = old_inverted[path_hash]
         try:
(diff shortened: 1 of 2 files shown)


dvc/repo/diff.py
```diff
@@ -201,18 +201,19 @@ def _calculate_renamed(new, old, added, deleted):
             bucket.append(path)
 
     renamed = []
-    for path in added:
+    for path in sorted(added, key=len):
```
skshetry, line 204: Sorting it to get the best results (in case of duplicated results). 

Before:
```console
$ dvc diff
Renamed:                                                              
    data/foo -> foo1
    foo -> data/foo1
```

After:
```console
$ dvc diff
Renamed:                                                              
    data/foo -> data/foo1
    foo -> foo1
```

dvc/repo/diff.py
```diff
@@ -190,29 +192,23 @@ def _targets_to_path_infos(repo_fs, targets):
...
-                break
-        if not old_path:
+        old_paths = old_inverted[path_hash]
+        try:
+            iterator = enumerate(old_paths)
+            index = next(idx for idx, path in iterator if path in deleted)
+        except StopIteration:
             continue
+
+        old_path = old_paths.pop(index)
```
skshetry, line 211: So, in the case of multiple files with the same hashes, the only first file was being used as an old file.

To figure out renames,
1. we compare newly added and old deleted files and find renamed files.
2. then, we remove those that are renamed entries from `added` section using new path.
3. then, we remove those that are renamed entries from `deleted` section using old path.

We can have the same hashes for multiple renamed entries, but here in 1st step, we were only using the first encountered one. But, in the 3rd step, we delete it from `deleted` entries which fail if there were multiple files with the same hashes.

skshetry:
Got a confirmation that it fixes the issue, waiting for the cause of the issue to be confirmed. Will add a test after that.

dvc/repo/diff.py
skshetry: Rolled back, we should figure out a better way to associate them in case of duplicated results (trie based approach could be useful). But, I don't think it's worth the effort.

efiop (approved):

PR #5653, opened 2021-03-19
ui: use global singleton UI object for printing

skshetry:
I understand this is a bit controversial, but it seems that using this `ui` object globally is the best thing to do.
DVC tries to provide a sane `API` as well as `CLI`. Extending api to accomodate `ui` object (which is for `cli`) does not seem like a good thing, as it extends our API surface and makes it ugly. We'd also need to pass them to the depth of our APIs.

Similarly, we cannot just use `ui` object in CLIs, as our cli commands are usually just wrapper to the `Repo` apis. So, even if we try hard to separate them and isolate it inside the `CLI`, it'd be quite difficult.

Looking at it, and the purpose that `ui` object serves, it seems to me that it is a global singleton object anyway (an instance in a single session of CLI). And, we use `loggers` in a similar way, so I find it fitting as per our needs, without adding any extra efforts.

### DIFF
--- a/dvc/ui/__init__.py
+++ b/dvc/ui/__init__.py
@@ -7,9 +7,6 @@ from funcy import cached_property
 from dvc.progress import Tqdm
 from dvc.utils import colorize
 
-NEWLINE = "\n"
-SEP = " "
-
 
 class Formatter:
     def __init__(self, theme: Dict = None, defaults: Dict = None) -> None:
@@ -30,14 +27,17 @@ class Console:
         formatter: Formatter = None,
         output: TextIO = None,
         error: TextIO = None,
-        disable: bool = False,
+        enable: bool = False,
     ) -> None:
         self._input: TextIO = sys.stdin
         self._output: TextIO = output or sys.stdout
         self._error: TextIO = error or sys.stderr
 
         self.formatter: Formatter = formatter or Formatter()
-        self.disabled: bool = disable
+        self._enabled: bool = enable
+
+    def enable(self):
+        self._enabled = True
 
     def success(self, message: str) -> None:
         self.write(message, style="success")
@@ -52,8 +52,8 @@ class Console:
         self,
         *objects: Any,
         style: str = None,
-        sep: str = SEP,
-        end: str = NEWLINE,
+        sep: str = None,
+        end: str = None,
         flush: bool = False,
     ) -> None:
         return self.write(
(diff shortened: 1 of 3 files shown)


dvc/ui/__init__.py
```diff
@@ -152,9 +152,11 @@ def table(self, header, rows, markdown: bool = False):
         self.write(ret)
 
 
-if __name__ == "__main__":
-    ui = Console()
+ui = Console()
```
skshetry, line 155: maybe it should be `console` now?

dvc/ui/__init__.py
```diff
@@ -69,12 +69,12 @@ def write(
...
         style: str = None,
-        sep: str = SEP,
-        end: str = NEWLINE,
+        sep: str = None,
+        end: str = None,
         file: TextIO = None,
         flush: bool = False,
     ) -> None:
-        if self.disabled:
+        if not self._enabled:
```
skshetry, line 77: Q: How important is thread-safety for us while printing? `logger` is thread safe, but this is not.

efiop:
@skshetry Could you clarify how we would change verbosity or in general tweak ui in cases like erepo?

skshetry:
> Could you clarify how we would change verbosity 

Not implemented, but verbosity could be change in a similar way. Probably, changing the API of `enable` to `configure`. 

```python
ui.configure(level="success", enable=True)
```

But, I think we should not do that, as I am thinking `ui.write` to be always visible (except on `--quiet`). Verbosity will display logging information as well.

> or in general tweak ui in cases like erepo?

I had implemented this before, but removed it to keep API simple. I am thinking of using [`contextvars`](https://docs.python.org/3/library/contextvars.html) here (requires compat package in 3.6, in stdlib since 3.7). It'll look something like following:

```python
with ui.enabled():
  ui.write("visible")
  with ui.disabled():
    ui.write("hidden")
    with ui.enabled():
      ui.write(" visible again")
    ui.write(" hidden again")
```
This should be thread and async safe.

efiop:
@skshetry Sorry for the delay. Looks pretty good. :+1:

dvc/ui/__init__.py
pared: So, ultimately, we are intending to use `ui.write` everywhere we use `logger` now? If so, I would say thread safety is pretty important as we already have some parts of code that utilize `ThreadPoolExecutor` and log something down the line (for example, [`_upload_plans`](https://github.com/iterative/dvc/blob/f2ba67aece1d5ff3bc5aaf2d40948606c3f7e5c8/dvc/remote/base.py#L370).

skshetry:
Okay, merging it for now.

dvc/ui/__init__.py
skshetry: thanks, @pared. Will come to it as we start using it in multithreaded applications.

PR #5668, opened 2021-03-22
fsspec: skip instance cache on all filesystems

isidentical:
fsspec caches the instances, so even though we instantiate a new instance it might retrieve one from the cache since we use the same authentication options. The retrieved instances come with their listing cache which causes problems on the test suite (considering they all share a single instance). This patch disables instance caching for all file systems, and also unifies the credentials APIs

### DIFF
--- a/dvc/fs/azure.py
+++ b/dvc/fs/azure.py
@@ -48,8 +48,6 @@ class AzureFileSystem(FSSpecWrapper):
         self.path_info = self.PATH_CLS(url)
         self.bucket = self.path_info.bucket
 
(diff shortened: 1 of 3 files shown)


efiop (commented):
Are we going to turn it on later instead of our `Pool`?

isidentical:
> Are we going to turn it on later instead of our Pool?

We might, though from what I see some of our test cases (I also hit this issue on the s3fs migration) affect others due to fsspec's dircache being not isolated. So a deeper research might be needed regarding this (maybe when migrating ssh/hdfs)

efiop (approved):
Sounds good. Let's keep that in mind.

PR #5584, opened 2021-03-09
plots: support opening plots file directly in the browser

skshetry:
Closes https://github.com/iterative/dvc/issues/4179
```
$ dvc plots show --open
```
I don't know if this is that useful, but I think this does save some time (I was feeling the need when I was going through `example-get-started`).

### Alternative:
No way to be compatible across different platforms. For me on Linux, it's following:
```
$ dvc plots show | xargs xdg-open
```

Working on macOS and on Windows (cc @jorgeorpinel?) confirmation would be great. 

* [x] ❗ I have followed the [Contributing to DVC](https://dvc.org/doc/user-guide/contributing/core) checklist.

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here.

Thank you for the contribution - we'll try to review it as soon as possible. 🙏

PR #5670, opened 2021-03-22
plots: quote and resolve paths properly

skshetry:
This commit fixes 3 things:
1. The path printed on the terminal is properly is quoted.  This was
problematic when the output is piped to `xdg-open` if the path contained
spaces or quotes or special characters.
```console
$ dvc plots show -o 'plots with spaces.html' | xargs xdg-open
xdg-open: unexpected argument 'with'
Try 'xdg-open --help' for more information.
```

2. The path is resolved. This is not an issue at all but makes the
output look nicer.
```console
$ dvc plots show -o '../plots.html' --open
file:///home/saugat/repos/iterative/example-get-started/../plots.html
```

3. The path with which the browser is opened with uses relative path.
This is done to solve issue in WSL, that the absolute path is not
accessible outside the WSL for host programs as the linux filesystem
is available on `/wsl$` path.
```console
$ dvc plots show --open
file:///home/saugat/example-get-started/plots.html
Start : This command cannot be run due to the error: The system cannot find the file specified.
At line:1 char:1
+ Start "/home/saugat/example-get-started/plots.html"
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Start-Process], InvalidOperationException
    + FullyQualifiedErrorId : InvalidOperationException,Microsoft.PowerShell.Commands.StartProcessCommand
```

4. On Windows, the file-uri was being constructed using platform-dependent `os.path.join`. This has been changed to use `as_uri()` from `pathlib.Path`.

With the fix to use relative path as argument to `webbrowser.open`, I expect WSL to be able to work from outside (tested) as well as within inside if needed (didnot test at all). This does work on Linux.

See https://github.com/iterative/dvc/pull/5584#issuecomment-801325773

### DIFF
--- a/dvc/command/plots.py
+++ b/dvc/command/plots.py
@@ -1,6 +1,5 @@
 import argparse
 import logging
-import os
 
 from dvc.command import completion
 from dvc.command.base import CmdBase, append_doc_link, fix_subparsers
@@ -23,6 +22,8 @@ class CmdPlots(CmdBase):
         return {k: v for k, v in props.items() if v is not None}
 
     def run(self):
+        from pathlib import Path
+
         if self.args.show_vega:
             if not self.args.targets:
                 logger.error("please specify a target for `--show-vega`")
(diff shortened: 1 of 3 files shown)


pared (approved):

skshetry:
Need to do a QA on Windows for the file-uri changes. Previously, the file-uri where being constructed through platform-specific paths, which means there were `\\` in them. Needs to be checked if those were valid before.

skshetry:
Was not able to test out on windows, I was curious if it worked before there, but probably it didn't.

PR #5674, opened 2021-03-23
Move tests requirements to test_requirements.txt

skshetry:
Plan is to add dependabot to update those packages required for testing, after this gets merged.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,6 @@
 import importlib.util
 import os
+from pathlib import Path
 
 from setuptools import find_packages, setup
 from setuptools.command.build_py import build_py as _build_py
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -113,49 +114,9 @@ def run(self):
...
-    "mypy",
-    "wsgidav",
-    "crc32c",
-    "google-cloud-storage==1.19.0",
-    # pypi doesn't allow for direct dependencies
-    #    "gdrivefs @ git+https://github.com/intake/gdrivefs.git",
-]
+tests_requirements = (
+    Path("test_requirements.txt").read_text().strip().splitlines()
+)
```
skshetry, line 119: Should we allow users to install with `[tests]` at all? I'd say that we should ask contributors to use `-r dev_requirements.txt` than using `[tests]` here. Anyway, avoiding that change/discussion for now. 

`dependabot` only understands pipfile/poetry lockfile/pip-tools and these requirements.txt.

efiop (approved):

Issue #5622, opened 2021-03-15
s3: s3fs migration


isidentical:
Roadmap;
- [x] A basic `S3FileSystem` class like the rest of providers
- [x] Passing the bare minimum tests (apparently moto doesn't work well with aiobotocore when used in wrapper mode, aio-libs/aiobotocore/issues/755, so we probably would have to rewrite the test remote using the server mode)
- [x] Porting `~/.aws/config` parsing
- [x] Permission/grant mangement
- [x] Port functional/unit tests (and drop the redundant ones)
- [x] etag-preserving server side copy implementation (possibly by overriding the s3fs's copy method while preserving the underlying logic) (dask/s3fs/pull/441)
- [x] support [exception unrolling](https://github.com/iterative/dvc/blob/dcc55aa645c361a865d24df33f6989bb15e13f24/dvc/logger.py#L153-L157) (dask/s3fs/pull/443)
- [x] s3fs doesn't support 3.6 (dask/s3fs/issues/445) [perhaps we can have a fork for 3.6 only and have it on pypi until we also drop the support, EOL is in 9 months]
- [ ] try to setup dvc bench
- [ ] end to end performance testing
- [ ] Collecting issues regarding s3 on the bug tracker and checking whether they still apply or not

Future;
- [ ] should we continue to wrap boto-based exceptions (e.g NoCreds/NoBucketFound etc)
- [ ] Look into whether we can preserve / port the `walk_files(prefix=)` optimization

isidentical:
It seems like the naive approach on s3fs (using the FSSpecWrapper without any additions) is about 1.5-3x slower than the current master of simple push/pull/status

Some stats regarding API calls;
```
       master         s3fs
GET     160            103
PUT      81             89
HEAD    132            278
POST      2              3
        375            473
```

Looking at the possible optimizations :eyes:

isidentical:
Here are some new numbers (real timings instead of request amounts), it appears to be that pull of small files is extremely slow (25x);
https://gist.github.com/isidentical/e0c33a5db41aca10509fcb5c95489247

isidentical:
I've re-implemented the push/pull logic in async way to see how much it affects, and it appears to be that this made the pull only 2x slower (compared to 9x slower in the previous version) though the push is now 4x slower (instead of 3x);

|                   | master    | s3fs      | s3fs with download/upload APIs | ASYNC s3fs with download/upload APIs |
[...]

PR #5683, opened 2021-03-24
s3: fsspec migration

isidentical:
Supersedes the first attempt on #5622.


dvc/fs/fsspec_wrapper.py
```diff
@@ -91,7 +91,17 @@ def copy(self, from_info, to_info):
...
-        return self.fs.exists(self._with_bucket(path_info))
+        # Some implementations don't use ls under the hood
+        # which is not an efficient behavior on DVC considering that
+        # we usually to a lot of stat calls. For increasing the cache
+        # construction, we will first try to see whether that directory
+        # exists by doing an ls() on it's parent, and if that fails we
+        # would fall back to the original implementation.
+        raw_path = self._with_bucket(path_info)
+        try:
+            return raw_path in self.fs.ls(self._with_bucket(path_info.parent))
```
shcheklein, line 102: hmm ... this feels like a very expensive op compared to a regular HEAD call (I would expect that's what `exists` do). Could you please clarify why is it needed? It was hard for me to understand it from the comment.

isidentical: Actually I'd say it is not that expensive (considering that the pagination works asyncronously). If the parent directory consists from 1000< items, then there is not much of an overhead. But if it has 8K items then the cost is about 3.5X worse for a single time. 

Considering that we do exists() calls for the same objects sometimes 2 sometimes even 3 (accomponied by .isfile()/.isdir()) I think it worths it. I am not really sure whether I should only apply this optimization to s3 implementation or in general, will actually intend to run benchmarks for the other remotes we have (adlfs/gcsfs) [currently still running more s3 benchmarks]). But I think it is a nice one (at least for s3)

shcheklein: Hmm ... I'm trying to wrap my mind around it :)

1. Okay, 8K works, but what if we have 1M?
2. Transferring even 1000 results is probably 1000x slower/ takes 1000x traffic? Means that in a slower network connection situation we might hit a bigger performance problem?
3. `Considering that we do exists() calls for the same objects sometimes 2 sometimes even 3` - can we improve/fix this? How the current approach helps us there. May be I'm missing something.


To be honest, it feels that I'm missing some important information here. Could explain me the comment about the `ls`? Let's put it this way - I'm not sure I understand why it's an optimization at all.

isidentical: > To be honest, it feels that I'm missing some important information here.

fsspec basically caches all the directory listings and later use those stuff on different calls (like info()/ls()/exists() etc). What we are basically doing here is sort of pre-populating the cache with the assumption of we will use later. Not sure how much I am missing, but it always shows a positive performance on my tests. 

> Okay, 8K works, but what if we have 1M?
> Transferring even 1000 results is probably 1000x slower/ takes 1000x traffic? Means that in a slower network connection situation we might hit a bigger performance problem?

Right, these cases seem problematic. Will think more about this one (maybe separate it out)

shcheklein: > fsspec basically caches all the directory listings and later use those stuff on different calls 

interesting! how/when do they invalidate this cache? 

> What we are basically doing here is sort of pre-populating the cache with the assumption of we will use later.

gotcha. Thanks. It makes sense. I wonder if we can do optimize this with a simple memoize decorator (I think `funcy` has it). But then we'll have the same question - when/how do we invalidate it? It can be a quite strong assumption and an important behavior detail since I can easily imagine where in certain cases we actually want to go and check an object again.

isidentical: Basically every operation (e.g copy, move, rm, upload) invalidates the cache for that directory, e.g;
https://github.com/iterative/dvc/blob/a3a55339abf133e8073a5c4220e7d61749b5f16f/dvc/fs/fsspec_wrapper.py#L141

By the way, I decided to drop this optimization for this PR. Instead I might go with #5689 which would definietly improve the amount of calls we do to exists() + info() (both in the master and then in this PR).

shcheklein: Yep, thanks, @isidentical . It's still really good to know about this built-in cache mechanism anyways. Since DVC is used in a distributed environment it might happen that it won't detect certain files when it "expects" them? I hope that it will be just a matter of some non-optimal behavior in some rare edge cases (e.g. dir uploaded twice, or something).

I wonder btw, why don't they just cache `exists` calls as well?

isidentical: > Since DVC is used in a distributed environment it might happen that it won't detect certain files when it "expects" them?

Unlike the internal database we use to hold stuff (diskcache), this is a per-process cache that is hold in memory so I'd except the amount of race conditions to be minimum (though it can still happen).

isidentical: > I wonder btw, why don't they just cache exists calls as well?

The structure they use for cache (dircache) is in the form of a directory where you'd keep `{'dir': [{'name': 'foo'}, {'name': 'bar'}]}` etc. So from what I can presume, individual items doesn't fit (AFAIK they can't simply create a new entry on the dircache without knowing all the contents, since the code assumes the dircache is complete and raises FileNotFoundError etc. when it can't find something over there).

isidentical:
I'm marking this PR ready to review. The test suite passes on every version, though since we are awaiting a new s3fs release we can't just merge it as is yet (only one line change in setup to point to pypi s3fs instead of github one solve this though, so it won't block reviews).

tests/remotes/s3.py
```diff
@@ -95,14 +102,72 @@ def read_text(self, encoding=None, errors=None):
...
 
-        boto3.client("s3").create_bucket(Bucket=TEST_AWS_REPO_BUCKET)
+    import shlex
+    import subprocess
+
+    import requests
+
+    proc = subprocess.Popen(
+        shlex.split(f"moto_server s3 -p {TEST_AWS_S3_PORT}")
+    )
```
efiop, line 147: Let's just use `motoserver/moto` docker image, same as we do with other remotes. (see tests/docker-compose.yml)

tests/docker-compose.yml
```diff
@@ -6,6 +6,10 @@ services:
     command: azurite -L -l /data --blobHost 0.0.0.0 --queueHost 0.0.0.0
     ports:
       - "10000"
+  motoserver:
+    image: motoserver/moto
+    ports:
+      - "5000:5000"
```
efiop, line 12: Looks like you are already logic for dynamic ports in `tests/remotes/s3.py`, so there shouldn't be any need to nail it down to 5000 on the host
```suggestion
      - "5000"
```

tests/remotes/s3.py
```diff
@@ -95,14 +100,58 @@ def read_text(self, encoding=None, errors=None):
...
+
+    with monkeypatch.context() as m:
+        m.setenv("AWS_ACCESS_KEY_ID", "testing")
+        m.setenv("AWS_SECRET_ACCESS_KEY", "testing")
+        m.setenv("AWS_SECURITY_TOKEN", "testing")
+        m.setenv("AWS_SESSION_TOKEN", "testing")
+        yield
+
+    if aws_creds.exists() and not initially_exists:
+        aws_creds.unlink()
```
efiop, line 124: I guess you meant smth like
```
        yield
    finally:
        if aws_creds.exists() and not initially_exists:
            aws_creds.unlink()
```

isidentical:
Benchmarks from the latest revision, same in general but there seems to be a slight improvement on `dvc status -c` which might be related with dircache on fsspec: https://gist.github.com/isidentical/9e0246c4fe700c019a49cd6ec6de2440

tests/docker-compose.yml
```diff
@@ -6,6 +6,10 @@ services:
     command: azurite -L -l /data --blobHost 0.0.0.0 --queueHost 0.0.0.0
     ports:
       - "10000"
+  motoserver:
+    image: motoserver/moto
+    ports:
+      - "5000"
```
skshetry, line 12: @isidentical, now that we are using docker 😩, please turn the following config off so that s3 tests are not run by default:

https://github.com/iterative/dvc/blob/31439754377d829da6babffcc720dd696554d8ab/tests/conftest.py#L22

efiop (approved):
:fire: 

And s3fs is out too https://pypi.org/project/s3fs/ (as well as gcsfs), just waiting on adlfs :slightly_smiling_face:

shcheklein:
hey, guys - quick ask - could we summarize (and may be visualize) before after? It's hard for me to figure out what do specific `dvc status -c` commands mean in the gist, what do we tests, what we don't test ...

(not a blocker for merge obviously, you decide - but would be really helpful to understand better what's going on and review)

🙏

isidentical:
> It's hard for me to figure out what do specific dvc status -c commands mean in the gist, what do we tests, what we don't test ...

Actually, it is a very simple example of push + pull of big single file / a dataset of a lot of small files with `status -c`'s in the middle of every operation (e.g before push / after push) etc. 

> (not a blocker for merge obviously, you decide - but would be really helpful to understand better what's going on and review)

I'll try to prepare a proper table with before/after of different scenerios.

isidentical:
We've discovered a bug that caused `exists()` calls on s3fs to take 6~ seconds for non-existent files, now this bug is resolved and it is taking 1.8-2.0~ seconds. This is still worse than our current implementation (for both existing and non-existing files, when the calls are made separately. This overhead is compensated when `list_hashes_exist` performs calls in pararel) (0.6-0.7~ seconds). For the last day, I've been debugging the cause and it appears to be the cost of `aiobotocore` + `aiohttp` (and all the underlying stuff).

efiop (approved):
Outstanding work :fire: 

Let's merge and move forward for now, [aiobotocore issue](https://github.com/aio-libs/aiobotocore/issues/866) is a problem, but we can live with it for now (we also have some time before the next release, etc).

Thank you! :pray:

PR #5688, opened 2021-03-25
[WIP] Dolt/integration

max-hoffman:


efiop:
@max-hoffman Could you please add a summary/explanation for the PR? Not everyone in the team is familiar with the dolt, so it would really help us to wrap our heads around it. Thanks! :pray:

max-hoffman:
Hey @efiop , yes my apologies, I have been working on a long-form write-up with a demo to address the how/why. Hopefully I'll have that in a sharable form in a couple days or worst-case next week.

A 30-second summary:
- Dolt is a Git-versioned SQL database. It can be used as a file, sort of like SQLite. It can be used as a server, compatible with the MYSQL wire protocol. The range of Git and SQL features are available through both, and everything funnels through Dolt's storage format that handles merges/diffs/etc.
- Git + MySQL creates features that are novel. Diffing/merging is fairly obvious, [Dolt SQL functions](https://docs.dolthub.com/interfaces/sql/dolt-sql-functions) is an interesting hybrid, our [system tables](https://docs.dolthub.com/interfaces/sql/dolt-system-tables#dolt-system-tables) are another good example.
- The files vs. server contention creates UX questions for us at the integration layer sometimes. Dolt is not a file-system, so it doesn't plug into DVC's commands in a one-to-one way like an object-store would.
- This PR routes some commands through dolt comparables (ex: we point dvc push/pull through dolt's push/pull).
- Dolt has its own object and tree hashing, reflected by the three-heads, which I used in place of an md5 hash (`{head}-{working}-{staging}.dolt`). The entire state of the database can be monitored or restored based on those three values.
- The rest of the changes are to make the unit-tests/small demo work as expected. Things like preventing removes on a Dolt database, or short-circuiting caching because we handle deltas, etc.

max-hoffman:
Hey folks, we spent some time last week rounding out this example: https://github.com/dolthub/dolt-integrations/pull/8

We are interested in blogging this, so excited to hear any feedback from your end.

efiop:
@max-hoffman Thanks for the short summary! Looks very interesting and useful for dolt users, but we clearly need to develop some kind of hooks/plugin mechanism that dolt might use to integrate with dvc, as maintaining dolt support in the upstream dvc would be rather challenging. What do you think about such potential approach? If you think it is suitable from your perspective, maybe you have some ideas for particular API/hooks/etc that you would need dvc to implement to support dolt?

max-hoffman:
Hey @efiop, thanks for taking a look!

One example of an external Dolt plugin would require DVC to expose a file interface. That interface would be used in DVC similar to Objects, where you can call things like `get_hash`, `has_changed`, `save`, `load`, `push`, `remote`, `pull`, `checkout` on objects that already have a pre-defined type, compared to switching on files, directories, remote types, etc individually in all of the above functions.

You already do a bit of this with remotes, something like [snakemake's remotes](https://github.com/snakemake/snakemake/tree/6f9f4a6cfb0cccfda8740d64c766ad368c52d78e/snakemake/remote) is one example where the main code is executed with a consistent class pattern that can be overridden.

A partial/easier implementation would expose only  `get_hash`, `has_changed`, `save`, `load`, `checkout` and a switch for turning off dvc remote interaction.

efiop:
@max-hoffman Thanks! We would really appreciate if you could formalize the requirements and propose a general mechanism (e.g. through https://github.com/iterative/enhancement-proposals or https://github.com/iterative/dvc/discussions for less formal prerequisited) in order for this to move forward. We won't be able to merge raw dolt support, it needs to be a 3rd party plugin.

Closing for now, since this needs a bunch of pre-requisites.

max-hoffman:
@efiop Thanks for taking the time! Let us know if you have any feedback before we publish our blog. Revisions and contributions are welcome.
https://dolthub.awsdev.ld-corp.com/blog/2021-04-12-dolt-dvc/

Issue #5679, opened 2021-03-23
Enable Vega plots without internet access


dberenbaum:
Plots are dependent on sourcing the javascript from vega, which makes the plots fail if that resource cannot be reached via internet. As a workaround, dvc can provide an option to replace the source of the javascript with some locally available copy of the vega javascript source.

efiop:
Fixed by #5694

PR #5694, opened 2021-03-25
html: enable providing custom html page templates

pared:

Initial implementation allowing custom HTML webpage template.
I decided just to allow users provide their own HTML, as long as there is `{plot_divs}` format string inside, so that we don't have to deal with additional args like `--vega-path` `--vega-lite-path` `--vega-embed-path` which would be necessary if we were to handle the template on our own, in `dvc` code.

How one can utilize new functionality :

```
#!/bin/bash

set -ex
rm -rf repo
mkdir repo
pushd repo

echo '[{"m":1},{"m":2},{"m":3}]' >> plot.json

echo '<html>
<head>
    <script type="text/javascript" src="VEGA_PATH"></script>
    <script type="text/javascript" src="VEGA_LITE_PATH"></script>
    <script type="text/javascript" src="VEGA_EMBED_PATH"></script>
</head>
<body>
	{plot_divs}
</body>
</html> ' >> template

wget https://cdn.jsdelivr.net/npm/vega@5.10.0 -O VEGA_PATH
wget https://cdn.jsdelivr.net/npm/vega-lite@4.8.1 -O VEGA_LITE_PATH
wget https://cdn.jsdelivr.net/npm/vega-embed@6.5.1 -O VEGA_EMBED_PATH

dvc plots show plot.json --html-template template
```

For sake of simplicity there is only `--html-template` option. However if we decide to go this way (allowing users to provide their own HTML template) I will prepare proper `dvc config`  command, that would allow doing the setup only once per project.

cc @JIoJIaJIu @mnrozhkov @dberenbaum

### DIFF
--- a/dvc/command/plots.py
+++ b/dvc/command/plots.py
@@ -44,7 +44,8 @@ class CmdPlots(CmdBase):
 
             rel: str = self.args.out or "plots.html"
             path = (Path.cwd() / rel).resolve()
-            write(path, plots)
+            write(path, plots=plots, template_path=self.args.html_template)
+
         except DvcException:
             logger.exception("")
             return 1
@@ -242,4 +243,10 @@ def _add_output_arguments(parser):
         action="store_true",
         default=False,
         help="Open plot file directly in the browser.",
+    )
+    parser.add_argument(
+        "--html-template",
+        default=None,
+        help="Custom HTML template for VEGA visualization.",
+        metavar="<path>",
     )
--- a/dvc/command/plots.py
+++ b/dvc/command/plots.py
@@ -5,7 +5,6 @@ from dvc.command import completion
 from dvc.command.base import CmdBase, append_doc_link, fix_subparsers
 from dvc.exceptions import DvcException
 from dvc.utils import format_link
-from dvc.utils.html import write
 
 logger = logging.getLogger(__name__)
 
@@ -44,7 +43,9 @@ class CmdPlots(CmdBase):
 
             rel: str = self.args.out or "plots.html"
             path = (Path.cwd() / rel).resolve()
-            write(path, plots=plots, template_path=self.args.html_template)
+            self.repo.plots.write_html(
+                path, plots=plots, html_template_path=self.args.html_template
+            )
 
         except DvcException:
             logger.exception("")
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -223,4 +223,5 @@ SCHEMA = {
         # enabled by default. It's of no use, kept for backward compatibility.
         Optional("parametrization", default=True): Bool
     },
+    "plots": {"html_template": str},
 }
--- a/dvc/utils/html.py
+++ b/dvc/utils/html.py
@@ -1,5 +1,8 @@
 from typing import Dict, List, Optional
 
+from dvc.exceptions import DvcException
+from dvc.types import StrPath
+
 PAGE_HTML = """<!DOCTYPE html>
 <html>
 <head>
@@ -9,7 +12,7 @@ PAGE_HTML = """<!DOCTYPE html>
     <script src="https://cdn.jsdelivr.net/npm/vega-embed@6.5.1"></script>
 </head>
 <body>
-    {divs}
+    {plot_divs}
 </body>
 </html>"""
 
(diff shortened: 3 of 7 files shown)


dvc/utils/html.py
```diff
@@ -54,13 +70,23 @@ def with_element(self, html: str) -> "HTML":
...
 
 def write(
-    path, plots: Dict[str, Dict], metrics: Optional[Dict[str, Dict]] = None
+    path: StrPath,
+    plots: Dict[str, Dict],
+    metrics: Optional[Dict[str, Dict]] = None,
+    template_path: Optional[StrPath] = None,
 ):
-    document = HTML()
+    print(metrics)
```
dberenbaum, line 83: Is this print statement intended to stay in here?

dberenbaum (approved):
LGTM. Another alternative would be to have a `cdn-path` or `vega-base` path or something that would replace the url used for all the vega scripts. We may end up revisiting the plots in the near future to address some of the other enhancements discussed, so whatever you and @JIoJIaJIu think is best for now.

pared:
Added config option for html_template, here is the sample script modified to use new option:

```
#!/bin/bash

set -ex
rm -rf repo
mkdir repo
pushd repo

git init --quiet
dvc init --quiet

git add -A
git commit -am "init"

path=$(pwd)
path=$path/.dvc

echo "<html>
<head>
    <script type=\"text/javascript\" src=\"$path/VEGA_PATH\"></script>
    <script type=\"text/javascript\" src=\"$path/VEGA_LITE_PATH\"></script>
    <script type=\"text/javascript\" src=\"$path/VEGA_EMBED_PATH\"></script>
</head>
<body>
	{plot_divs}
</body>
</html>" >> .dvc/template.html

dvc config plots.html_template template.html

wget https://cdn.jsdelivr.net/npm/vega@5.10.0 -O .dvc/VEGA_PATH
wget https://cdn.jsdelivr.net/npm/vega-lite@4.8.1 -O .dvc/VEGA_LITE_PATH
wget https://cdn.jsdelivr.net/npm/vega-embed@6.5.1 -O .dvc/VEGA_EMBED_PATH

mkdir sub
pushd sub
echo '[{"m":1},{"m":2},{"m":3}]' >> plot.json
dvc plots show plot.json 
```

Please note the fact that in this case using full paths in HTML template is necessity, that's the only way to make sure custom template works regardless where do we call `plots` from.

dberenbaum:
Fixes #5679

JIoJIaJIu (approved):
LGTM

dvc/config_schema.py
```diff
@@ -223,4 +223,5 @@ class RelPath(str):
         # enabled by default. It's of no use, kept for backward compatibility.
         Optional("parametrization", default=True): Bool
     },
+    "plots": {"html_template": str},
```
jorgeorpinel, line 226: Should it be just `plots.html`? To avoid overloading the term "template" (also associated with [Vega customizations](https://dvc.org/doc/command-reference/plots#plot-templates)).

dvc/command/plots.py
```diff
@@ -243,3 +245,9 @@ def _add_output_arguments(parser):
         default=False,
         help="Open plot file directly in the browser.",
     )
+    parser.add_argument(
+        "--html-template",
```
jorgeorpinel, line 249: Similarly, should it be just `--html`?

dvc/utils/html.py
```diff
@@ -20,9 +23,22 @@
...
 
+class MissingPlaceholderError(DvcException):
+    def __init__(self, placeholder):
+        super().__init__(f"HTML template has to contain '{placeholder}'.")
+
+
 class HTML:
-    def __init__(self):
-        self.elements = []
+    PLACEHOLDER = "plot_divs"
```
jorgeorpinel, line 32: What about `{dvc_plots}`? But...

jorgeorpinel: From https://github.com/iterative/dvc.org/issues/2402#issuecomment-865831930:

> think about going with pure HTML and chaning the way the "templates" work - my idea would be to remove the requirement for the placeholder and just insert the plots after <body> tag.

Is this a better implementation? Cc @pared thanks

jorgeorpinel (commented):
Post-merge discussion per https://github.com/iterative/dvc.org/issues/2402#issuecomment-866664740

dvc/utils/html.py
iesahin: I don't think removing the `PLACEHOLDER` is better. People may want to put other things in the `<body/>`, like branding, other kinds of images, or text.

jorgeorpinel: Good point.

PR #5696, opened 2021-03-26
pull: skip isdir() checks on downloading files

### LINKED ISSUE #5689: pull: implement a filesytstem.download_file protocol for skipping isdir() check
When pulling caches from the remote storage, we only pull files, not directories. But each `download()` call made by pull (for each cache file we want to retrive) costs an extra `isdir()` call which is not needed at all (considering that pull will always fetch files, not directories). 
https://github.com/iterative/dvc/blob/a3a55339abf133e8073a5c4220e7d61749b5f16f/dvc/remote/base.py#L311-L314
https://github.com/iterative/dvc/blob/a3a55339abf133e8073a5c4220e7d61749b5f16f/dvc/fs/base.py#L276-L280

We can get a pretty straightforward optimization by just implementing a `BaseFileSystem.download_file` method which would skip this check and go directly to the `download_file`. 

This API is fsspec-compatible, as it offers a [`get()`](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractFileSystem.get) API for getting a directory or a file recursively (just like our `download()`) and it offers a single [`get_file()`](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractFileSystem.get_file) API which fetches a single file from the remote. 

```sh
mkdir /tmp/test_dvc
pushd /tmp/test_dvc
git init > /dev/null
dvc init > /dev/null
dvc remote add -d default s3://dvc-temp/s3fs-migration/h5-s3fs/
python ~/tooling/create_dataset.py dataset1 10000
dvc add dataset1 > /dev/null
time dvc push dataset1 > /dev/null
time dvc status -c > /dev/null
rm -rf dataset1 .dvc/cache
time dvc status -c > /dev/null
time dvc pull dataset1 > /dev/null
popd
```

The performance improvement for the script above (creates 10k files and pulls them) was about %30 (8.30min with download_file, 11.30min on the master). The only concern of mine is whether this operation is going to change in the future ODB systems, but even then I think this API might be useful.

isidentical:
Resolves #5689

### DIFF
--- a/dvc/fs/base.py
+++ b/dvc/fs/base.py
@@ -1,6 +1,6 @@
 import logging
 from concurrent.futures import ThreadPoolExecutor, as_completed
-from functools import partial
+from functools import partial, partialmethod
 from multiprocessing import cpu_count
 from typing import Any, ClassVar, Dict, FrozenSet, Optional
 
@@ -258,6 +258,7 @@ class BaseFileSystem:
         name=None,
         no_progress_bar=False,
         jobs=None,
+        _only_file=False,
         **kwargs,
     ):
         if not hasattr(self, "_download"):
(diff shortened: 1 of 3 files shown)


efiop (approved):
:fire:

PR #5720, opened 2021-03-28
exp show: Allow for wildcard patterns in include and exclude params/m…

### LINKED ISSUE #5642: Allow for wildcard patterns in include and exclude params/metrics with dvc exp show
When using the `dvc exp show` command, you have the possibility to include and exclude both parameters and metrics using the arguments:
```
--include-params
--exclude-params
--include-metrics
--exclude-metrics
```

Instead of just providing an explicit list of parameters or metrics to include/exclude, it would be nice to have the possibility to select them using simple patterns using * as a wildcard

For example, in case I have a variety of different models  over time, each with their own test-set metrics json file I would now have to add the specific metric for each of the models I would like to see using `--include-metrics`.

With a wildcard, this could become:
```sh
dvc exp show --include-metrics-pattern "metrics/test*f1-score"
```
which would select all of the metrics that start with "metrics/test"  and end with "f1-score"

A `--exclude-metrics-pattern` argument could then be implemented in the same way. Also implementing the combination for parameters would allow for easily allowing to easily selecting only those parameters and metrics you want to have without having to put them all explicitly on the commandline

In theory, this could be extended to work not just with simple * wildcards but with complete regex patterns if needed.

daavoo:
…etrics

Fixes #5642

Related doc PR: https://github.com/iterative/dvc.org/pull/2334


pmrowla (commented):
One question I have about this is how we should handle matching across dictionary levels.

Currently, we support matching against nested dictionary levels using the `.` syntax to separate nested keys. So for a `params.yaml` like in our [example-get-started](https://github.com/iterative/example-get-started/blob/master/params.yaml)

```yaml
prepare:
  split: 0.20
  seed: 20170428

featurize:
  max_features: 3000
  ngrams: 2

train:
  seed: 20170428
  n_est: 100
  min_split: 64
```

You can currently do `--include-params prepare` to match everything in `prepare.*`.

in master:
```
dvc exp show --no-pager --include-params prepare
┏━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┓
┃ Experiment    ┃ Created      ┃     auc ┃ prepare.split ┃ prepare.seed ┃
┡━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━┩
│ workspace     │ -            │ 0.57756 │ 0.2           │ 20170428     │
│ master        │ Nov 11, 2020 │ 0.57756 │ 0.2           │ 20170428     │
...
```

I'm not sure what the best way for us to handle globbing would be - should it only match within a single nesting level, or should it match the entire string?

For reference, this PR currently only matches inside one nesting level (and not the entire string), so we get:

```
dvc exp show --no-pager --include-params \*split
ERROR: failed to show experiments - '*split' does not match any known params
```

```
dvc exp show --no-pager --include-params \*.split                                   ⏎
┏━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━━━━┓
┃ Experiment    ┃ Created      ┃     auc ┃ prepare.split ┃
┡━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━━┩
│ workspace     │ -            │ 0.57756 │ 0.2           │
│ master        │ Nov 11, 2020 │ 0.57756 │ 0.2           │
...
```

In this case, `*split` is only compared against the top level keys (`prepare`, `featurize`, `train`) and does not match anything. Using `*.split` matches `*` against the top level keys, and then `split` against the next level of keys (and matches `prepare.split`)

pmrowla:
Keeping this as-is and requiring separate glob patterns per-nesting level (separated by `.`) for matching within nested keys is fine with me, but we would have be sure to document it. @dberenbaum @jorgeorpinel thoughts?

pmrowla:
So on second thought, maybe it would be best to drop the current method of allowing `prepare` to match everything nested beneath `prepare.`, and only support matching via glob patterns (against the entire string). So the user would have to explicitly use `prepare*` or `prepare.*` to get the current `--include-params prepare` behavior.

daavoo:
> So on second thought, maybe it would be best to drop the current method of allowing `prepare` to match everything nested beneath `prepare.`, and only support matching via glob patterns (against the entire string). So the user would have to explicitly use `prepare*` or `prepare.*` to get the current `--include-params prepare` behavior.

I personally consider this the best solution from the perspectives of both user experience (easier to define patterns for complicated nested params) and maintenance (matching the full string will simplify the code) . I did not implemented it because, as you have noted, it introduces a backwards incompatible change and I though that was out of the scope in a first P.R (I have no problem with implementing the full string matching approach if you so choose).

daavoo:
> One question I have about this is how we should handle matching across dictionary levels.
> 
> Currently, we support matching against nested dictionary levels using the `.` syntax to separate nested keys. So for a `params.yaml` like in our [example-get-started](https://github.com/iterative/example-get-started/blob/master/params.yaml)
> 
> ```yaml
> prepare:
>   split: 0.20
>   seed: 20170428
> 
> featurize:
>   max_features: 3000
>   ngrams: 2
> 
> train:
>   seed: 20170428
>   n_est: 100
>   min_split: 64
> ```
> 
> You can currently do `--include-params prepare` to match everything in `prepare.*`.
> 
> in master:
> 
> ```
> dvc exp show --no-pager --include-params prepare
> ┏━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┓
> ┃ Experiment    ┃ Created      ┃     auc ┃ prepare.split ┃ prepare.seed ┃
> ┡━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━┩
> │ workspace     │ -            │ 0.57756 │ 0.2           │ 20170428     │
> │ master        │ Nov 11, 2020 │ 0.57756 │ 0.2           │ 20170428     │
> ...
> ```
> 
> I'm not sure what the best way for us to handle globbing would be - should it only match within a single nesting level, or should it match the entire string?
> 
> For reference, this PR currently only matches inside one nesting level (and not the entire string), so we get:
> 
> ```
> dvc exp show --no-pager --include-params \*split
> ERROR: failed to show experiments - '*split' does not match any known params
> ```
> 
> ```
> dvc exp show --no-pager --include-params \*.split                                   ⏎
> ┏━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━━━━┓
> ┃ Experiment    ┃ Created      ┃     auc ┃ prepare.split ┃
> ┡━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━━┩
> │ workspace     │ -            │ 0.57756 │ 0.2           │
> │ master        │ Nov 11, 2020 │ 0.57756 │ 0.2           │
> ...
> ```
> 
> In this case, `*split` is only compared against the top level keys (`prepare`, `featurize`, `train`) and does not match anything. Using `*.split` matches `*` against the top level keys, and then `split` against the next level of keys (and matches `prepare.split`)

I will add a test case for nested params as that use case is not currently covered.

dberenbaum:
Yeah, I prefer getting rid of the current approach and simplifying to a full string pattern match. The current way has some nice benefits for exact matches to filenames, sections, etc., but it's less intuitive and flexible. I don't think backwards incompatibility is a big deal since it's an experimental feature and I don't think current behavior is even documented.

My other question would be whether `fnmatch` is preferable to regex? `fnmatch` is probably simpler for most, but I'm sure some users will claim to want the power of regex (for example, if they have some convention like making certain parameters all caps).

pmrowla:
I think keeping it simple w/fnmatch is best for now. If/when we get enough users asking for regex we can consider it at that point.

@daavoo you can go ahead and make the changes to drop the existing behavior and only do the glob matching against the full param string

daavoo:
>My other question would be whether fnmatch is preferable to regex? fnmatch is probably simpler for most, but I'm sure some users will claim to want the power of regex (for example, if they have some convention like making certain parameters all caps).

> I think keeping it simple w/fnmatch is best for now. If/when we get enough users asking for regex we can consider it at that point.

Unless I am missing something, replacing `fnmatch` with `regex` would be an easy change to make, when the claim comes.

dvc/command/experiments.py
```diff
@@ -27,32 +27,23 @@ def _filter_name(names, label, filter_strs):
...
-                    ret[match_path].update(
-                        {".".join(match): None for match in matches}
+        for match_path in match_paths:
+            for f in filters:
+                matches = [
+                    name for name in names[match_path] if fnmatch(name, f)
+                ]
+                if not matches:
+                    raise InvalidArgumentError(
+                        f"{f} does not match any known {label}"
```
pmrowla, line 44:
```suggestion
                        f"'{f}' does not match any known {label}"
```

We usually wrap these kinds of strings in single quotes to make them easier to read in terminal output

dvc/command/experiments.py
```diff
@@ -27,32 +27,23 @@ def _filter_name(names, label, filter_strs):
 
     for filter_s in filter_strs:
         path, _, name = filter_s.rpartition(":")
-        path_filters[path].append(tuple(name.split(".")))
+        path_filters[path].append(name)
```
pmrowla, line 30: We should also add a note in the docs PR that by default, filter patterns will be applied to all params/metrics files, unless the filter specifically includes the `path:` prefix, and that wildcards are not supported in the path prefix itself (the file path must be explicit)

pmrowla (changes requested):
Looks good for the most part, just a couple minor notes.

dvc/command/experiments.py
dberenbaum: Hm, I guess I was thinking it would be a match against the full column header (`path:section.param` where paths are included) since this would be most intuitive for users. Seems like this is halfway there, matching everything after the path, which as you note requires some additional documentation.

pmrowla: @dberenbaum If I have a repo with 2 parameter files:

```yaml
# params.yaml
foo: 1
foobar: 2
```

```yaml
# other_params.yaml
foo: 3
```
The table will contain 3 columns: `params.yaml:foo`, `foobar`, `other_params.yaml:foo`

With the current implementation, `foo*` will match all 3 columns. With the proposed change, you would need `*foo*` to match all 3 columns.

The first pattern seems most intuitive to me, especially since in DVC we default to omitting pathnames for params/metrics whenever possible, but I'm not particularly opposed to going with the alternative either.

pmrowla: I know this is kind of a contrived example, but "params with matching prefixes in different files" does seem like a realistic/potential user scenario to me.

dberenbaum: Yeah, thanks for the example. Since dvc omits path names where possible, it's a little confusing regardless of the choice. My suggestion would make it really hard to get all the parameters from a single file, which seems like enough reason to discard it.

What about doing fnmatch separately on the path and the param, assuming the `:` is still needed if including a path, so that `p*:*` would include the first two columns? @pmrowla @daavoo up to you at this point.

dberenbaum (approved):
LGTM

dvc/command/experiments.py
pmrowla: I'm in favor of leaving this as-is for simplicities sake, if/when enough people ask for the additional matching we can address it at that time.

pmrowla (approved):

pmrowla:
@daavoo can you update the docs PR to reflect the path discussion (the wildcard pattern will not be applied to `path:` prefix)? Once that's done we can merge this PR

dvc/command/experiments.py
daavoo: Ok!

daavoo:
> @daavoo can you update the docs PR to reflect the path discussion (the wildcard pattern will not be applied to `path:` prefix)? Once that's done we can merge this PR

Done: https://github.com/iterative/dvc.org/pull/2334

Issue #5556, opened 2021-03-05
Issue with fetch from azure remote "no current event loop in thread"


friggog:
# Bug Report

I'm getting this error
```
ERROR: failed to download 'azure://container/***' to '.dvc\cache\***' - There is no current event loop in thread 'ThreadPoolExecutor-127_0
...
ERROR: failed to download 'azure://container/***' to '.dvc\cache\***' - There is no current event loop in thread 'ThreadPoolExecutor-127_14
ERROR: failed to fetch data from the cloud - 53 files failed to download
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x000002159F326E88
```
For a number of files (not all as far I can tell) when running 
`dvc fetch [some_files]` on Azure devops windows CI machine with SPA auth.

DVC v2.0.3, everything seems to run fine on my local windows machine

This still fails if I set the number of jobs to 1

This is working fine with DVC v1.11.10

efiop:
@friggog Please post verbose log for that command. E.g. `dvc fetch -v`. Also, please post `dvc doctor` output.

friggog:
```
DVC version: 2.0.3 (pip)
---------------------------------
Platform: Python 3.7.9 on Windows-10-10.0.14393-SP0
Supports: azure, hdfs, http, https, ssh
Cache types: <https://error.dvc.org/no-dvc-cache>
Caches: local
Remotes: azure
Workspace directory: NTFS on D:\
Repo: dvc (subdir), git
2021-03-05 16:23:34,513 DEBUG: Check for update is enabled.
2021-03-05 16:23:34,540 DEBUG: Trying to spawn '['daemon', '-q', 'updater']'
2021-03-05 16:23:34,590 DEBUG: Spawned '['daemon', '-q', 'updater']'
2021-03-05 16:23:34,619 DEBUG: Checking if stage 'FILE_1' is in 'dvc.yaml'
2021-03-05 16:23:35,958 DEBUG: Checking if stage 'FILE_2' is in 'dvc.yaml'
2021-03-05 16:23:35,959 DEBUG: Checking if stage 'FILE_3' is in 'dvc.yaml'
2021-03-05 16:23:35,960 DEBUG: Checking if stage 'DIR\FILE_4' is in 'dvc.yaml'
2021-03-05 16:23:36,759 DEBUG: Preparing to download data from 'azure://CONTAINER_NAME'
[...]

Issue #5511, opened 2021-02-23
More complete Azure authentication


friggog:
Continuing discussion from #5435.

I believe SPA has somewhat been addressed, but my two primary requests are:
- Automatic detection of azure auth env vars on azure devops CI when using [AzureCLI task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/deploy/azure-cli?view=azure-devops), these are added as `servicePrincipalId`, `servicePrincipalKey` and `tenantId`.
- Support for [RBAC](https://docs.microsoft.com/en-us/azure/role-based-access-control/overview) auth rather than relying on account key.

The code snippet below supports authentication for azure blob storage using SPA with the above env vars, falling back to RBAC auth through get_azure_cli_credentials. This triggers interactive login in the browser (through azure CLI) if the user hasn't logged in on that machine already, from then on it happens silently.

```python
from adal import AdalError
from azure.cli.core import get_default_cli
from azure.cli.core._profile import Profile
from azure.common.credentials import get_azure_cli_credentials, _CliCredentials, ServicePrincipalCredentials
from knack.util import CLIError
from typing import Union


def _get_credential() -> Union[_CliCredentials, ServicePrincipalCredentials]:
        if 'servicePrincipalId' in os.environ:
            tenant_id = os.environ['tenantId']
            client_id = os.environ['servicePrincipalId']
            client_secret = os.environ['servicePrincipalKey']
            return ServicePrincipalCredentials(tenant=tenant_id, client_id=client_id, secret=client_secret)

        try:
            credential, _ = get_azure_cli_credentials()        # throws FileNotFoundError on a clean machine
            credential.get_token('https://storage.azure.com')  # throws CLIError if token not available and
                                                               # throws AdalError if authentication has expired
        except (FileNotFoundError, CLIError, AdalError):
            # Azure CLI token is not available. Perform interactive authentication to create it.
            # This is equivalent to running "az login".

            # get_default_cli() is required when running on a clean machine.
            get_default_cli()

            # async_persist=False will cause credentials cache to be flushed to disk immediately (instead of waiting
            # for program exit).
            print('Please complete the login process in the web browser.')
            profile = Profile(async_persist=False)
            profile.find_subscriptions_on_login(
                interactive=True,
                username=None,
                password=None,
                is_service_principal=False,
                tenant=None,
            )
            print('Done.\n')
            credential, _ = get_azure_cli_credentials()

        return credential
```

[...]

PR #5725, opened 2021-03-29
azure: use account_name with default credentials

friggog:
This resolves the authentication issues highlighted in #5511. Before I was getting authentication failures with DefaultAzureCredential, now auth seems to succeed but I still face the problems in #5556 which make this hard to test beyond the fact that the error has changed from auth to threading.

EDIT: Hacked azure.identity.aio so that it doesn't complain by disabling asyncio.Lock - with this done I no longer get the treading errors of #5556 and auth works fine! Not really sure what is the cause of the threading issues and this clearly isn't a viable solution though...

### DIFF
--- a/dvc/fs/azure.py
+++ b/dvc/fs/azure.py
@@ -69,8 +69,8 @@ class AzureFileSystem(FSSpecWrapper):
         login_info["client_id"] = config.get("client_id")
         login_info["client_secret"] = config.get("client_secret")
 
-        if not any(login_info.values()):
-            login_info["credentials"] = DefaultAzureCredential()
+        if not any([a in login_info and a is not None for a in ["connection_string", "account_key", "sas_token", "client_secret"]]):
+            login_info["credential"] = DefaultAzureCredential(exclude_interactive_browser_credential=False)
 
         for login_method, required_keys in [  # noqa
             ("connection string", ["connection_string"]),
--- a/dvc/fs/azure.py
+++ b/dvc/fs/azure.py
@@ -50,6 +50,8 @@ class AzureFileSystem(FSSpecWrapper):
 
     def _prepare_credentials(self, config):
         from azure.identity.aio import DefaultAzureCredential
+        # Disable spam from failed cred types for DefaultAzureCredential
+        logging.getLogger("azure.identity.aio").setLevel(logging.ERROR)
 
         login_info = {}
         login_info["connection_string"] = config.get(
--- a/dvc/fs/azure.py
+++ b/dvc/fs/azure.py
@@ -69,7 +69,8 @@ class AzureFileSystem(FSSpecWrapper):
         login_info["client_id"] = config.get("client_id")
         login_info["client_secret"] = config.get("client_secret")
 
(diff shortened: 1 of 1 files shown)


dvc/fs/azure.py
```diff
@@ -69,8 +69,8 @@ def _prepare_credentials(self, config):
         login_info["client_id"] = config.get("client_id")
         login_info["client_secret"] = config.get("client_secret")
 
-        if not any(login_info.values()):
-            login_info["credentials"] = DefaultAzureCredential()
+        if not any([a in login_info and a is not None for a in ["connection_string", "account_key", "sas_token", "client_secret"]]):
```
isidentical, line 72:
```py
        if not any(
            [
                a in login_info and a is not None
                for a in [
                    "connection_string",
                    "account_key",
                    "sas_token",
                    "client_secret",
                ]
            ]
        ):
```

This condition seems like always succeeding. As you can see `a` is always a string from that list which is non none, and `login_info` always contains it (even though if you don't specify it in the config)

friggog: Ah sorry should be `login_info[a] is not None`

isidentical: Can you explain why you excluded other values from here (e.g `tenant_id`). AFAIK `DefaultCredentials` will override the specified `tenant_id` in your config so I am not so sure that would be wanted behavior.,

friggog: Just that all three of tenant_id, client_id and client_secret are required for SPA so it seemed superfluous to check against all

dvc/fs/azure.py
```diff
@@ -69,8 +69,9 @@ def _prepare_credentials(self, config):
         login_info["client_id"] = config.get("client_id")
         login_info["client_secret"] = config.get("client_secret")
 
-        if not any(login_info.values()):
-            login_info["credentials"] = DefaultAzureCredential()
+        non_cred_auth_methods = ["connection_string", "client_secret", "account_key", "sas_token"]
+        if not any([login_info.get(a, None) is not None for a in non_cred_auth_methods]):
```
isidentical, line 73: If what you want to exclude is only `account_name`, I'd go with
```suggestion
        if not any(value for key, value in login_info.items() if key != 'account_name'):
```

dvc/fs/azure.py
```diff
@@ -50,6 +50,8 @@ def __init__(self, repo, config):
 
     def _prepare_credentials(self, config):
         from azure.identity.aio import DefaultAzureCredential
+        # Disable spam from failed cred types for DefaultAzureCredential
+        logging.getLogger("azure.identity.aio").setLevel(logging.ERROR)
```
friggog, line 54: Not sure what the best way to do this is, but DefaultAzureCredential will log XXCredential.get_token failed blah blah loads of times (one for each thread?) for any auth methods it tries which aren't successful before finding one that is

isidentical: What is the level that they are logging those methods, if it is debug then it should be fine.

friggog: I think warning

isidentical:
@friggog can you run our linting hooks on this PR (via pre-commit).

isidentical (approved):

Issue #5388, opened 2021-02-02
support specifying custom certificates when accessing remotes


skshetry:
something like `certs` in the config for the remotes instead of using `AWS_CA_BUNDLE`/`REQUESTS_CA_BUNDLE`, etc.

Related: https://github.com/iterative/dvc/issues/3450, https://github.com/iterative/dvc.org/issues/1079

skshetry:
Closing as not planned. Looks like there is not much of an interest, and since we are using fsspec now, different filesystems will have different requirements, so we have to implement this on a per-filesystem basis.

PR #5732, opened 2021-03-30
Added SSL verification option

### LINKED ISSUE #3450: s3 and compatible: add a config option to support validate_certs
Similar to use_ssl, we should probably provide a config for this.

More context here - https://discordapp.com/channels/485586884165107732/485596304961962003/685129104214851595

NigelVanHattum:
Added the parameter to allow the use of SSL without the SSL verification. This uses the similar named boto3 parameter. 

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a [separate PR](https://github.com/iterative/dvc.org/pull/2341) (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here.

Fixes #3450

### DIFF
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -145,7 +145,7 @@ SCHEMA = {
                     "session_token": str,
                     Optional("listobjects", default=False): Bool,  # obsoleted
                     Optional("use_ssl", default=True): Bool,
-                    Optional("ssl_verify", default="true"): str,
+                    Optional("ssl_verify", default=True): bool,
                     "sse": str,
                     "sse_kms_key_id": str,
                     "acl": str,
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -145,7 +145,7 @@ SCHEMA = {
                     "session_token": str,
                     Optional("listobjects", default=False): Bool,  # obsoleted
                     Optional("use_ssl", default=True): Bool,
-                    Optional("ssl_verify", default=True): bool,
+                    Optional("ssl_verify", default=True): Bool,
                     "sse": str,
                     "sse_kms_key_id": str,
                     "acl": str,
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -145,7 +145,7 @@ SCHEMA = {
                     "session_token": str,
                     Optional("listobjects", default=False): Bool,  # obsoleted
                     Optional("use_ssl", default=True): Bool,
-                    Optional("verify_ssl", default=True): Bool,
+                    Optional("verify_ssl", default="true"): str,
                     "sse": str,
                     "sse_kms_key_id": str,
                     "acl": str,
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -145,7 +145,7 @@ SCHEMA = {
                     "session_token": str,
                     Optional("listobjects", default=False): Bool,  # obsoleted
                     Optional("use_ssl", default=True): Bool,
-                    Optional("verify_ssl", default="true"): str,
+                    Optional("ssl_verify", default="true"): str,
                     "sse": str,
                     "sse_kms_key_id": str,
                     "acl": str,
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -145,6 +145,7 @@ SCHEMA = {
                     "session_token": str,
                     Optional("listobjects", default=False): Bool,  # obsoleted
                     Optional("use_ssl", default=True): Bool,
+                    Optional("verify", default=True): Bool,
                     "sse": str,
                     "sse_kms_key_id": str,
                     "acl": str,
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -145,7 +145,7 @@ SCHEMA = {
                     "session_token": str,
                     Optional("listobjects", default=False): Bool,  # obsoleted
                     Optional("use_ssl", default=True): Bool,
-                    Optional("verify", default=True): Bool,
+                    Optional("verify_ssl", default=True): Bool,
                     "sse": str,
                     "sse_kms_key_id": str,
                     "acl": str,
--- a/dvc/fs/s3.py
+++ b/dvc/fs/s3.py
@@ -38,7 +38,7 @@ class S3FileSystem(BaseFileSystem):
         self.endpoint_url = config.get("endpointurl")
 
         self.use_ssl = config.get("use_ssl", True)
-        self.verify_ssl = config.get("verify_ssl", True)
+        self.ssl_verify = config.get("ssl_verify", True)
 
         self.extra_args = {}
 
@@ -137,7 +137,7 @@ class S3FileSystem(BaseFileSystem):
         return session.resource(
             "s3",
             endpoint_url=self.endpoint_url,
-            verify=self.verify_ssl,
+            verify=self.ssl_verify,
             use_ssl=self.use_ssl,
             config=boto3.session.Config(
                 signature_version="s3v4", s3=s3_config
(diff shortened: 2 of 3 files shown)


dvc/config_schema.py
```diff
@@ -145,6 +145,7 @@ class RelPath(str):
                     "session_token": str,
                     Optional("listobjects", default=False): Bool,  # obsoleted
                     Optional("use_ssl", default=True): Bool,
+                    Optional("verify", default=True): Bool,
```
efiop, line 148: `verify` is already used for different purpose (see REMOTE_COMMON above), so we need to use something else. E.g.
```suggestion
                    Optional("verify_ssl", default=True): Bool,
```
?

efiop:
@NigelVanHattum Thank you so much for the PR! :pray: Please note that github didn't recognize the email in your commit, so you might want to go to https://github.com/settings/emails and add it there, so it can associate your commit with your github profile.

dberenbaum:
Related to #3450?

NigelVanHattum:
> @NigelVanHattum Thank you so much for the PR! 🙏 Please note that github didn't recognize the email in your commit, so you might want to go to https://github.com/settings/emails and add it there, so it can associate your commit with your github profile.
It should be there

dvc/fs/s3.py
```diff
@@ -136,6 +137,7 @@ def s3(self):
         return session.resource(
             "s3",
             endpoint_url=self.endpoint_url,
+            verify=self.verify_ssl,
```
efiop, line 140: For the record, we are migrating to s3fs right now, and looks like we will be able to just use `client_kwargs` there https://github.com/dask/s3fs/blob/main/s3fs/core.py

efiop (commented):
Could you also add a simple unit test to https://github.com/iterative/dvc/blob/master/tests/unit/fs/test_s3.py ?

efiop:
@isidentical Could you take a look, please? Also, how do you want to go about the ordering of this and https://github.com/iterative/dvc/pull/5683 ?

dvc/config_schema.py
```diff
@@ -145,6 +145,7 @@ class RelPath(str):
                     "session_token": str,
                     Optional("listobjects", default=False): Bool,  # obsoleted
                     Optional("use_ssl", default=True): Bool,
+                    Optional("verify_ssl", default=True): Bool,
```
skshetry, line 148: While we are at it, we should also try to provide a way to specify a custom cert (see https://github.com/iterative/dvc/issues/5388). We could just use `verify_ssl` to accept a string to a path or create a separate config. 

Note that we have the `ssl_verify` option in HTTP remote already, so we are being inconsistent here. We should deprecate either of those. 🙂

NigelVanHattum: I agree with your first point, but I disagree with the second. As we are directly parsing this to boto3, here is what they say about the difference: 

- use_ssl (boolean) -- Whether or not to use SSL. By default, SSL is used. Note that not all services support non-ssl connections.

- verify (boolean/string) -- Whether or not to verify SSL certificates. By default SSL certificates are verified. You can provide the following values:
**False** - do not validate SSL certificates. SSL will still be used (unless use_ssl is False), but SSL certificates will not be verified.
**path/to/cert/bundle.pem** - A filename of the CA cert bundle to uses. You can specify this argument if you want to use a different CA cert bundle than the one used by botocore.

skshetry: @NigelVanHattum, not sure I understand. I am saying that maybe we should extend `verify_ssl` to also accept the path to the certificates (as you have written, they accept boolean and the path to cert, requests work in a similar fashion).

efiop: @skshetry Good point about `ssl_verify`, totally forgot about it. @NigelVanHattum would you be so kind to rename it once more, please? Sorry for the inconvenicence. 

Cert path option could probably wait until someone asks for it.

skshetry: > Cert path option could probably wait until someone asks for it.

@efiop, users are asking about this. We have been pointing users to `AWS_REQUESTS_BUNDLE` and `REQUESTS_CA_BUNDLE` for quite a long time now. They do work, but it always feels like a hack.

skshetry: I am not trying to block this PR, it could be introduced separately. This config does have its place, I just worry that we'll suggest this to the users instead of a more secure option in which the certificates are validated properly.

NigelVanHattum: Just added the option to give a path as a value to this parameter as well. I will change the name to ssl_verify

NigelVanHattum: > @NigelVanHattum, not sure I understand. I am saying that maybe we should extend `verify_ssl` to also accept the path to the certificates (as you have written, they accept boolean and the path to cert, requests work in a similar fashion). 🙂

Aah, I think I've read your comment wrong. I'm renaming the parameter

dvc/config_schema.py
```diff
@@ -145,6 +145,7 @@ class RelPath(str):
                     "session_token": str,
                     Optional("listobjects", default=False): Bool,  # obsoleted
                     Optional("use_ssl", default=True): Bool,
+                    Optional("ssl_verify", default="true"): str,
```
isidentical, line 148: If I do `dvc remote modify something ssl_verify true` or `dvc remote modify something ssl_verify false`, how will it automatically coerce these to bool values rather than paths? I might be missing something but it looks like it will always use these as strings, and even I say `"false"` it will interpret it either as a path or a true value.

efiop: It could probably be done with something like `supported_cache_type` that we use above, but it really shows that we are probably better off not combining bool and str in this option, as even the name is not really friendly for the `path` case. As @skshetry mentioned above, we could introduce that in a separate PR later, for now keeping `ssl_verify` only as Bool

isidentical: Agreed

NigelVanHattum: Suggestion, I could add an option to enter a path as a str. Boto can handle both on the verify param, so we only nee to discuss which one had priority above the other?

NigelVanHattum: @isidentical I've just made a function that parses this parameter. Also tested all cases on this function

dvc/fs/s3.py
```diff
@@ -20,6 +21,19 @@
...
+    """
+    Checks the type of the input parameter and returns the
+    boolean or certPath based on the input
+    """
+    if isinstance(ssl_verify_config_value, bool):
+        return ssl_verify_config_value
+    try:
+        return strtobool(ssl_verify_config_value)
+    except ValueError:
+        return ssl_verify_config_value
```
efiop, line 34: This should be on the config level, like `supported_cache_types` we've mentioned before. str to bool conversions don't belong on the fs level. If you don't want to mess around with it more - reverting to supporting only bool would be totally acceptable.

NigelVanHattum: Just going to try to implement it, otherwise I'll revert

NigelVanHattum: @efiop Could you show me where the value of supported_cache_type is used? I can't seem to find it anywhere

NigelVanHattum: Done

dvc/config_schema.py
```diff
@@ -145,6 +145,7 @@ class RelPath(str):
                     "session_token": str,
                     Optional("listobjects", default=False): Bool,  # obsoleted
                     Optional("use_ssl", default=True): Bool,
+                    Optional("ssl_verify", default=True): bool,
```
efiop, line 148:
```suggestion
                    Optional("ssl_verify", default=True): Bool,
```

efiop (approved):

efiop:
Thank you @NigelVanHattum !

Issue #5731, opened 2021-03-30
metrics diff: reports existing file as not existing


iksnagreb:
# Bug Report

`metrics diff`: reports existing file as not existing

## Description

Using dvc 2.0.13 to compare metrics of pipeline runs, dvc reports a parameters file (listed in the `vars` section of the `dvc.yaml`) as not existing and quits while the file is clearly existing in the workspace as well as the git HEAD (and other previous commits).

```console
ERROR: failed to show metrics diff - failed to parse 'vars' in 'exp/baseline/dvc.yaml': 'exp/project.yaml' does not exist
```

### Environment information

**Output of `dvc doctor`:**

```console
$ dvc doctor
DVC version: 2.0.13 (pip)
---------------------------------
Platform: Python 3.8.5 on Linux-5.4.0-70-generic-x86_64-with-glibc2.29
Supports: http, https, webdav, webdavs
Cache types: hardlink, symlink
Cache directory: ext4 on /dev/sda1
Caches: local
Remotes: webdavs, webdavs
Workspace directory: ext4 on /dev/sda1
Repo: dvc, git

```

**Additional Information (if any):**

[dvc.yaml ](https://gist.github.com/iksnagreb/c2fc28c1937d6b87340a16440472ce3d)
[verbose error log](https://gist.github.com/iksnagreb/16eefb74463da8af8f944a0bffd5eaec)

https://discord.com/channels/485586884165107732/485596304961962003/825417237333803019
https://discord.com/channels/485586884165107732/485596304961962003/826445437924802651

iksnagreb:
@skshetry dvc seems to be able to see the file:

```
exists locally True
# Global project parameters
project:
  # Project root directory
  root: ../../
  # Project binary/program directory
  bin: ../../bin
  # Data parameters
  data:
    # Root directory
    root: ../../data
  # Experiment parameters
  exp:
    # Root directory
    root: ../

exists in HEAD True
# Global project parameters
project:
  # Project root directory
  root: ../../
  # Project binary/program directory
  bin: ../../bin
  # Data parameters
  data:
    # Root directory
    root: ../../data
  # Experiment parameters
  exp:
    # Root directory
    root: ../
```

skshetry:
Thanks @iksnagreb, found the root cause of this. Seems like the issue is with using `..` in the path, which is not being resolved.

skshetry:
[...]

PR #5734, opened 2021-03-30
parsing: resolve relative paths when loading parameters

skshetry:
We were joining the paths in the following way: 

https://github.com/iterative/dvc/blob/09cb49542364694f2eec84f61b63e40ae8dcd05f/dvc/parsing/context.py#L404

This created the path such as `PosixPath: dir/../params.yaml`, which broke on GitFS when trying to get the key of the GitTrie,
as it is now changed to `('dir', '..', 'params.yaml')` parts instead of just `(params.yaml,)`.

Similarly, LocalFS.exists was also failing because of the unresolved path.

This is fixed by changing it to abspath first, removing those relative paths and then converted into `PathInfo`.

Closes https://github.com/iterative/dvc/issues/5731

### DIFF
--- a/dvc/parsing/context.py
+++ b/dvc/parsing/context.py
@@ -401,9 +401,9 @@ class Context(CtxDict):
     ):
         path, _, keys_str = item.partition(":")
         select_keys = lfilter(bool, keys_str.split(",")) if keys_str else None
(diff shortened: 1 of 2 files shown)


efiop (approved):
Thanks!

PR #5741, opened 2021-03-31
status: collect stages only once

### LINKED ISSUE #5621: status: looks like we are collecting stages twice
https://github.com/iterative/dvc/blob/4ed55d00511ea3d9115b76c463e1a466408b11ef/dvc/command/status.py#L70

We might end up collecting stages twice for status command if there are no changes, once for actual `repo.status` call and another in above if statement. 

This happens because we clear the collected stages when the `repo.status` is done with, which happens on the following decorator:

https://github.com/iterative/dvc/blob/4ed55d00511ea3d9115b76c463e1a466408b11ef/dvc/repo/status.py#L114

skshetry:
This is achieved by using `lock_repo` context manager that can be used in a recursive depth that only resets stages when it has left any lock_repo blocks.

Also added some UI related tests for the status which I found missing.

Fixes #5621

### DIFF
--- a/dvc/command/status.py
+++ b/dvc/command/status.py
@@ -44,21 +44,27 @@ class CmdDataStatus(CmdDataBase):
                 self._show(value, indent + 1)
 
     def run(self):
+        from dvc.repo import lock_repo
         from dvc.ui import ui
 
         indent = 1 if self.args.cloud else 0
-        try:
-            st = self.repo.status(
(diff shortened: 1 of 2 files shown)


pared (approved):

efiop (approved):

efiop:
@skshetry Nice catch!

Issue #5628, opened 2021-03-15
exp run --run-all: runs full pipeline, even stages that did not change


dberenbaum:
# Bug Report

## Description

Queueing stages and then running with `dvc exp --run-all` runs every stage in the pipeline, regardless of what stages have changed.

### Reproduce

```console
$ git clone git@github.com:iterative/example-get-started.git
$ cd example-get-started
$ pip install -r src/requirements.txt
$ dvc pull data/data.xml.dvc
$ dvc exp run -S train.n_est=200
$ dvc exp run -S train.n_est=300 --queue
$ dvc exp run --run-all --verbose
2021-03-15 15:18:33,934 DEBUG: Check for update is enabled.
2021-03-15 15:18:33,940 DEBUG: Reproducing experiment revs '6b69f41'
2021-03-15 15:18:34,059 DEBUG: Writing experiments local config '/private/tmp/example-get-started/.dvc/tmp/exps/tmpvrxfnzm9/.dvc/config.local'
2021-03-15 15:18:34,059 DEBUG: Init temp dir executor in dir '<ExpTemporaryDirectory '/private/tmp/example-get-started/.dvc/tmp/exps/tmpvrxfnzm9'>'
2021-03-15 15:18:34,525 DEBUG: Running repro in '/private/tmp/example-get-started/.dvc/tmp/exps/tmpvrxfnzm9'
2021-03-15 15:18:34,525 DEBUG: Removing '/private/tmp/example-get-started/.dvc/tmp/exps/tmpvrxfnzm9/.dvc/tmp/repro.dat'
2021-03-15 15:18:34,556 DEBUG: 'data/data.xml' doesn't exist.
2021-03-15 15:18:34,556 DEBUG: Checking out 'data/data.xml' with cache 'object md5: a304afb96060aad90176268345e10355'.
2021-03-15 15:18:34,557 DEBUG: Created 'reflink': ../../../cache/a3/04afb96060aad90176268345e10355 -> data/data.xml
2021-03-15 15:18:34,558 DEBUG: state save (4444713, 1615835726497615872, 37891850) a304afb96060aad90176268345e10355
2021-03-15 15:18:34,561 DEBUG: 'data/features' doesn't exist.
2021-03-15 15:18:34,561 DEBUG: Checking out 'data/features' with cache 'object md5: a079f4c29277a48a254ef3a0e695b8fe.dir'.
2021-03-15 15:18:34,561 DEBUG: Linking directory 'data/features'.
[...]

PR #5751, opened 2021-04-01
exp run: --reset no longer implies --force

pmrowla:

Will fix https://github.com/iterative/dvc/issues/5628

Docs PR: https://github.com/iterative/dvc.org/pull/2348

### DIFF
--- a/dvc/repo/experiments/__init__.py
+++ b/dvc/repo/experiments/__init__.py
@@ -324,7 +324,6 @@ class Experiments:
 
         if reset:
             self.reset_checkpoints()
-            kwargs["force"] = True
 
         if not (queue or tmp_dir):
             staged, _, _ = self.scm.status()
--- a/dvc/repo/experiments/executor/base.py
+++ b/dvc/repo/experiments/executor/base.py
@@ -271,7 +271,7 @@ class BaseExecutor(ABC):
             # - if run without --reset, the checkpoint out will be checked out
             #   using any hash present in dvc.lock (or removed if no entry
             #   exists in dvc.lock)
-            checkpoint_reset = kwargs.pop("reset", False)
+            checkpoint_reset: bool = kwargs.pop("reset", False)
             dvc_checkout(
                 dvc,
                 targets=targets,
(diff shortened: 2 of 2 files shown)


dberenbaum:
To summarize, `--reset` only implies `--force` if there is also a checkpoint out for any stage?

Assuming that's correct:
* What's the behavior of running `--reset` without checkpoint? Should this be allowed?
* If I have a stage with a checkpoint that is downstream of stages without checkpoints that were previously committed (by `dvc repro`), can I reset without having to run all of those upstream stages?

pmrowla:
> To summarize, `--reset` only implies `--force` if there is also a checkpoint out for any stage?

`--reset` no longer implies `--force` at all regardless of whether or not a stage contains `checkpoint` outs. Previously it implied `--force` all the time

> * What's the behavior of running `--reset` without checkpoint? Should this be allowed?

Specifying `--reset` for a stage without checkpoint outs now does nothing (the option is just silently ignored). Previously, this would cause the unexpected behavior of running non-checkpoint stages with `--force`.

> * If I have a stage with a checkpoint that is downstream of stages without checkpoints that were previously committed (by `dvc repro`), can I reset without having to run all of those upstream stages?

In this case, the `--reset` flag will be applied to all of your upstream stages. On `dvc exp run --reset <target>`, we do the equivalent of `dvc checkout --with-deps <target>`, and the `--reset` behavior (to drop existing checkpoint outs if they exist in dvc.lock) happens during this internal checkout step.

There is not currently a way to only `--reset` one specific stage. Doing so, would probably require some kind of separate `--reset-targets/--reset-all` flags, but I think that would be for a separate discussion/feature request.

dberenbaum:
> There is not currently a way to only `--reset` one specific stage

Does `--reset` force execution of non-checkpoint stages upstream of the checkpoint out?

pmrowla:
> Does `--reset` force execution of non-checkpoint stages upstream of the checkpoint out?

No, nothing will be changed for non-checkpoint stages so they will not be rerun (unless there is a checkpoint stage further upstream in the pipeline)

dberenbaum (approved):
LGTM

PR #5756, opened 2021-04-01
ui: unify tables

skshetry:

This PR try to unify the tables behind the same abstraction (with facade?). 

A few possible abstraction has been left out, eg: unifying pagers and trying to use rich more. A temporary workaround exists for rich console, as it didn't work well with capsys during testing, and I was unable to figure out why it's happening. 

I also added a support for borders, and row styles in this PR for the rich table.

### DIFF
--- a/dvc/types.py
+++ b/dvc/types.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, List, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
 from dvc.path_info import PathInfo, URLInfo
 
@@ -11,3 +11,5 @@ AnyPath = Union[str, DvcPath, StrPath]
 
 OptStr = Optional[str]
 TargetType = Union[List[str], str]
+DictStrAny = Dict[str, Any]
+DictAny = Dict[Any, Any]
--- a/dvc/ui/__init__.py
+++ b/dvc/ui/__init__.py
@@ -1,11 +1,21 @@
 import sys
 from collections import defaultdict
-from typing import Any, Dict, Iterable, Optional, TextIO
-
-from funcy import cached_property
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    Iterable,
+    Optional,
+    Sequence,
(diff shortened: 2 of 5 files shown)


pmrowla:
Rather than having `ui.table(..., markdown=False, rich=False)` with individual parameters for each output type, I guess I was envisioning having some base `Table` abstraction with subclasses for each rendering method? Something like

```py
class Table:
    rows = []
    headers = []
    ...
    def render(self):
       # return default plaintext table


class MarkdownTable(Table):
    def render(self):
        # return markdown formatted text


class RichTable(Table):
    def render(self):
        # return rich formatted text
```
and then we would just be able to do
```py
    table = ...Table(rows, headers, styles, pager=True)
    ui.write(table.render())
```
The specific table class (rich/markdown/etc) would be set as needed in CLI commands.

pmrowla:
I was also wondering if maybe we can use rich's existing Table class everywhere? Since `__str__` for any renderable type in rich return the plaintext (w/no styling) content, rendering the GFM version of Table should just be something along the lines of 

```py
def to_markdown(rich_table):
    table = [
        [str(cell) for cell in row.cells]
        for row in table.rows
    ]
    headers = ... # do the same (get header strings from table.columns)
    return tabulate(table, headers, tablefmt="github")
```

skshetry:
> Rather than having `ui.table(..., markdown=False, rich=False)` 

`ui.table()` API easily maps with the CLI arguments. Otherwise, we'd have to duplicate logic for using the appropriate table.

I thought of doing it in the similar way you propose (I was thinking of just extending RichTable to support markdown). We can consider them as an implementation detail and provide one API for a way to print a table. We can come back to it later to change if required (considering it to be an internal API).

In the end, like the way you propose, there is only one method/API that we really need to provide (`render()` vs `ui.table()`), so the inheritance/class solution felt like an overkill.

I am still looking to get rid of the old `tabulate` based table, and make `pager` work natively with `rich` and `ui.write()` without much effort.

pmrowla (approved):

efiop (approved):

pared (approved):

Issue #2575, opened 2019-10-05
ui: add a way to undo `dvc add` easily


Suor:
Maybe repurpose `dvc remove` for this, which at least be somewhat intuitive.

The hard part here is removing added file from cache: `dvc gc` is dangerous now and might be slow, simply removing cache file doesn't account for possible other refs to it.

This is inspired by https://github.com/iterative/dvc.org/issues/625 and a corresponding Stackoverflow question.

dashohoxha:
@Suor Are you suggesting that `dvc remove file.dvc` removes the cached outputs of `file.dvc` and the `file.dvc` itself, but does not touch the files on the work place? This would indeed undo `dvc add file`. Maybe it should also delete the output files from `.gitignore`, to be exactly the undo operation.

Suor:
@dashohoxha basically yes. We need to decide on retiring current `dvc remove` behaviour first though.

dashohoxha:
>  @Suor: We need to decide on retiring current dvc remove behaviour first though.

@shcheklein has already asked for `dvc remove` to be deprecated/removed. I proposed that maybe we should improve it, instead of removing. But now I think that your approach of re-purposing it is better.

mazzma12:
Hi, what about `dvc reset` to follow git naming?

dashohoxha:
@mazzma12 The analogy between Git and DVC is not perfect because DVC does not have an intermediate stage, before you commit. When you make `dvc add` it also does an implicit `dvc commit` (saving file to cache). To remove from Git a file that is committed by mistake you actually need `git rm --cached file && git commit --amend -CHEAD` (https://help.github.com/en/enterprise/2.15/user/articles/removing-files-from-a-repositorys-history#removing-a-file-added-in-the-most-recent-unpushed-commit)

But `dvc reset` makes sense to me, since `dvc remove` might seem a bit scary to a beginner. Even `dvc untrack` would be ok for me.

Suor:
git reset is a bad name, we shouldn't use it as a guide.

вс, 6 окт. 2019 г., 15:35 mazzma12 <notifications@github.com>:

casperdcl:
> git reset is a bad name, we shouldn't use it as a guide

maybe bad but the functionality will be understandable by end users without much explanation.

shcheklein:
[...]

PR #5757, opened 2021-04-01
remove: Add support for removing tracked files / directories

### LINKED ISSUE #5288: remove: support file as a target
### Description
As a further improvement of #4497, the support of using file as a target for ```dvc remove``` command should be added.

Here is what @skshetry says (https://github.com/iterative/dvc/pull/5236#issuecomment-759330299):
> we need to support file as a target in the remove, so I am not sure if this new error message makes sense.
> 
> We do need to clean up the message, especially the "failed to remove" one. So, either we need to add support for it, or just do the cleanup in this PR.
> 
> Regarding the support for files, the change should be pretty trivial, we just need collect_granular here, similar to what's in the repo.commit:
> 
> **dvc/dvc/repo/commit.py** (Line 44 in 228b3b1):
> ``` stages_info = self.stage.collect_granular( )```

@efiop made another point (https://github.com/iterative/dvc/issues/4497#issuecomment-759298988) based on a user's feedback:
> Another user ran into it https://discord.com/channels/485586884165107732/485596304961962003/798834932217937970

/cc @efiop @skshetry

### LINKED ISSUE #5772: remove: granular support
After implementing and reviewing #5757 we come to the conclusion that we should further discuss about how to approach adding "granular support" to `dvc remove` or even if it is needed.

---
## Context 

Given the following example yaml:

```yaml
train:
  cmd: python train.py data.csv
  deps:
    - data.csv
    - train.py
  outs:
    - logs
    - model.h5
```

```shell
$ ls
dvc.lock  dvc.yaml  data.csv  datacsv.dvc  logs  model.h5  train.py
```
With #5757 you can now run (previously raised `StageFileDoesNotExistError`):

```shell
dvc remove model.h5
```
And successfully remove only that output file.

```remove
$ ls
dvc.lock  dvc.yaml  data.csv  data.csv.dvc  logs  train.py
```

Howeve, in the same scenario, if we have several output files under `logs`:

```shell
$ ls logs
logs.csv logs.json
```
And run:

```shell
dvc remove logs.json
```
The current behavior (after #5757) will remove the full `logs` directory. See test case: 

https://github.com/iterative/dvc/blob/c871f3b5854ef73a4f2c6c16f760925b394be6b7/tests/func/test_remove.py#L51

---

## Discussion

Given the previous context, the question that arise is:

- Is this (from an user perspective) the expected behavior?

And further:

If this is the expected behavior, should we:

A) Leave the implementation as it is.
B) Add a warning to inform users that might not expect the bevaiour
C) other ideas

On the other hand, if this is not the expected behavior, should we:

A) Prevent the user from running the command (i.e. raising an exception)
B) Add support for granular remove, so that `dvc remove logs.json` only removes that specific file and not the full folder.
C) other ideas

daavoo:
Fixes #5288 
Fixes #5772

Docs: https://github.com/iterative/dvc.org/pull/2357


dvc/repo/remove.py
```diff
@@ -11,9 +11,16 @@
 
 @locked
 def remove(self: "Repo", target: str, outs: bool = False):
-    stages = self.stage.from_target(target)
+    stages_info = self.stage.collect_granular(target)
 
-    for stage in stages:
-        stage.remove(remove_outs=outs, force=outs)
+    for stage_info in stages_info:
```
skshetry, line 16: Let's just unpack them here, makes it easier.
```suggestion
    for stage, filter_info in stages_info:
```

dvc/repo/remove.py
```diff
@@ -11,9 +11,16 @@
...
+        stage = stage_info.stage
+        filter_info = stage_info.filter_info
+        if filter_info is not None:
+            # target is a specific output file
+            for out in stage.filter_outs(filter_info):
+                out.remove(True)
```
skshetry, line 22: This might not support granularity (eg: when the output is `data` and the target is `data/images/`).

skshetry: I am not sure if we need to support this though.

tests/func/test_remove.py
```diff
@@ -32,10 +32,29 @@ def test_remove(tmp_dir, scm, dvc, run_copy, remove_outs):
...
+def test_remove_file_as_target(tmp_dir, scm, dvc):
+    dvc.run(
+        name="my",
+        cmd='echo "hello" > foo && echo "hello" > bar',
+        deps=[],
+        outs=["foo", "bar"],
+    )
+
+    assert "/foo" in get_gitignore_content()
+    assert "/bar" in get_gitignore_content()
```
skshetry, line 44: This is not a part of `remove`, so we can assume it works (or tested for `dvc.run`).

tests/func/test_remove.py
```diff
@@ -32,10 +32,29 @@ def test_remove(tmp_dir, scm, dvc, run_copy, remove_outs):
...
+
+    dvc.remove("foo")
+    assert "/foo" not in get_gitignore_content()
+    assert "/bar" in get_gitignore_content()
```
skshetry, line 48: +1 to these assertions.

dvc/repo/remove.py
daavoo: I assume that setting a directory (i.e. `data/images`) as target might be out of the scope as the issue is currently described (We could extend this P.R. to support directories as target, if needed?) but I actually did not test to set output `data` and target something like `data/images/image_0.png`; I will add a test case for that.

daavoo: It seems that setting `outs=["foo"]` and running `dvc.remove("foo/file1"`) is removing all files under `foo`, not only the given target. I added a test for that and I'm reviewing the code

daavoo: @skshetry is the failing test case (which i have just added) supposed to pass, given the feature description?:

```
def test_remove_file_in_subdir_as_target(tmp_dir, scm, dvc):
    dvc.run(
        name="my",
        cmd='mkdir -p "foo" && echo "hello" > foo/file1 && echo "hello" > foo/file2',
        deps=[],
        outs=["foo"],
    )

    dvc.remove("foo/file1")
    # Fails here because `foo` is being completely removed, not just the file specified
    assert "/foo" in  get_gitignore_content()
    dvc.remove("foo/file2")
    assert not (tmp_dir / ".gitignore").exists()
```
Do we want to add support to that?

skshetry: @daavoo, that's expected, as we have not implemented the granular remove logic. Let's not implement this PR, let's discuss it in a separate issue.

To give you a more context, we have this implemented on the checkout. 

https://github.com/iterative/dvc/blob/c63457746332762ef6f203d2e71d8f7f4bde497a/dvc/repo/checkout.py#L113

Which we pass to Stage's checkout and then to the `Output`'s checkout:
https://github.com/iterative/dvc/blob/c63457746332762ef6f203d2e71d8f7f4bde497a/dvc/stage/__init__.py#L531

https://github.com/iterative/dvc/blob/c63457746332762ef6f203d2e71d8f7f4bde497a/dvc/output/base.py#L404-L413

So the support should be similarly implemented in the `Output` and this `filter_info` should be passed along.

skshetry: I am not sure how much this feature of granular remove makes sense.
Do the users expect this complete removal? Or, should we implement the granular remove, or should we forbid it?

Though forbidding it might not be a good idea for consistency. 🤷🏼

daavoo: > @daavoo, that's expected, as we have not implemented the granular remove logic. Let's not implement this PR, let's discuss it in a separate issue.

Sorry for my lack of understanding :sweat_smile: but by `Let's not implement this PR` you mean that:

A) We should close this P.R. and open a separate issue for discussing this feature and "granular remove"
B) We should leave this P.R. as it is, with the current implementation, and also open a separate issue for further discussion on extending the functionality?
C) None of the above  :sweat_smile:

skshetry: @daavoo, sorry for the typo. We can merge this and then implement granular remove later. Otherwise,  the  PR will get large and hard to review.

daavoo: #5772

dvc/repo/remove.py
```diff
@@ -11,9 +11,14 @@
 
 @locked
 def remove(self: "Repo", target: str, outs: bool = False):
-    stages = self.stage.from_target(target)
+    stages_info = self.stage.collect_granular(target)
```
jorgeorpinel, line 14: Does the `target` argument description string (`argparse`) need an update? 🙂

tests/func/test_remove.py
```diff
@@ -32,10 +32,40 @@ def test_remove(tmp_dir, scm, dvc, run_copy, remove_outs):
...
+    dvc.run(
+        name="my",
+        cmd='echo "hello" > foo && echo "hello" > bar',
+        deps=[],
+        outs=["foo", "bar"],
+    )
+
+    dvc.remove("foo")
+    assert "/foo" not in get_gitignore_content()
+    assert "/bar" in get_gitignore_content()
```
karajan1001, line 45: Maybe checking contents in `dvc.yaml` directly is better?

tests/func/test_remove.py
```diff
@@ -32,10 +32,40 @@ def test_remove(tmp_dir, scm, dvc, run_copy, remove_outs):
...
+
+    dvc.remove("bar")
+    assert not (tmp_dir / ".gitignore").exists()
+
+
+def test_remove_file_in_subdir_as_target(tmp_dir, scm, dvc):
```
karajan1001, line 51: These two tests can be merged into one? Generating a stage output both file and path, then remove them.

tests/func/test_remove.py
daavoo: I'm not sure if I fully understand your suggestion. As far as I know, using `dvc.remove("foo")` will not update the contents in `dvc.yaml`; it will remove `foo` from both the file system and the `.gitignore`

dvc/repo/remove.py
daavoo: Indeed, although I think I should wait for some clarifications regarding the doc update (https://github.com/iterative/dvc.org/pull/2357) before updating the argument description

tests/func/test_remove.py
daavoo: I felt like the 2 tests are better separated because one is asserting a successful behavior and the other is asserting that a specific exception is being raised

tests/func/test_remove.py
karajan1001: This confused me, because according to [doc](https://dvc.org/doc/command-reference/remove#remove) `dvc remove`  delete stages or `.dvc` files, including removing `.dvc` file or removing stages in `dvc.yaml` and removing outputs from .gitignore. I don't know the exact behavior of this feature, and seems that it wasn't defined clearly in [documents](https://github.com/iterative/dvc.org/pull/2357#discussion_r608201613)

daavoo: I understand the source of confusion, we surely need to improve the documentation related with this change

jorgeorpinel:
Will this fix #2575? If so please link them in the description of this PR.

daavoo:
> Will this fix #2575? If so please link them in the description of this PR.

It will not fix the original premise of the issue (you still have to use `dvc gc` for cleaning the cache) but it will not raise the exception described by https://github.com/iterative/dvc/issues/2575#issuecomment-707868202 and work as intended in that case.

efiop:
Unfortunately, even though this works, but the code is convoluted 🙁 We'll need a more general solution once we have a better architecture in this department. Closing this in favor of the path outlined in https://github.com/iterative/dvc/pull/6300

PR #5773, opened 2021-04-06
update: only update the missing hashes with --to-remote

### LINKED ISSUE #4992: cache: check and gracefully handle move to existing file
See https://github.com/iterative/dvc/issues/4876#issuecomment-726461567

isidentical:
`dvc update --to-remote` now only updates the missing cache files instead of syncing all over, as a pre-requisite this patch also merges `transfer()` logic with `stage()` and introduces the `stage(upload=...)` option for transferring to the remote cache. 

Fixes #4992


dvc/objects/file.py
```diff
@@ -51,3 +53,26 @@ def check(self, odb):
...
+class StreamFile(HashFile):
+    def __init__(self, path_info, fs, stream):
+        from dvc.utils.stream import HashedStreamReader
+
+        super().__init__(path_info, fs, None)
+        self.stream = HashedStreamReader(stream)
+
+    def sync(self, **upload_args):  # pylint: disable=unused-argument
+        with open(os.devnull, "wb") as stream:
+            shutil.copyfileobj(self.stream, stream)
```
efiop, line 67: Seems like the only difference between `StreamFile` and normal file staging is that we compute the hash not right away, but when `sync` is called, which is effectively the same as normal file staging, because we call `sync` right away in `_transfer_file`.

The only special case is `TransferredStreamFile` one, where we combine uploading to a temp location with on-the-fly hash computation. This makes me think that we need to get rid of `transfer()` in favor of `stage()` that will accept some kind of strategy flag. Like `stage(...., upload=True)`. What do you think?

dvc/objects/transfer.py
```diff
@@ -60,23 +69,62 @@ def create_tasks(executor, amount):
...
+                from_info,
+                jobs,
+                excluded_paths=frozenset(
+                    entry[tree.PARAM_RELPATH]
+                    for entry in entries
+                    if entry[PARAM_CHECKSUM] in existing
+                ),
+                no_progress_bar=no_progress_bar,
+                stream_cls=TransferredStreamFile,
+            ).as_dict()
```
efiop, line 118: Looks really complicated

efiop: Also, looks like we are effectively filtering out the already existing objects, but this is actually something that we can figure out during `save()`ing, as `odb.add` can figure it out on itsown. The only problem is that in order to know what to exclude, we need its hash_info, which will result in potentially unneeded uplading in case of TransferredStreamFile, hence you not using it in case of `update`.

efiop:
@isidentical Cool optimization! :fire: We've talked about getting rid of `transfer()` in favor of `stage()` before, I think it needs to be done as a pre-requisite for this optimization, as it seems `transfer` itself gets more complicated and actually `update` case is almost indistinguishable from normal `stage()`. And because of this additional similarity, I think it is worth doing that in this PR.

dvc/objects/transfer.py
```diff
@@ -85,8 +133,10 @@ def transfer(odb, from_fs, from_info, jobs=None, no_progress_bar=False):
             from_fs,
             from_info,
             jobs=jobs,
+            update=known_hash and known_hash.isdir,
```
efiop, line 136: Btw, `known_hash` pretty much doesn't matter at all here, as `save()` will call `odb.add()` that will check for existing object with the same object id (aka hash_info). So instead of passing `known_hash` to `transfer()` we coud've just passed `update=True/False`

isidentical: Agreed. It is a leftover from the previous logic and I kept it in case of might need it in the future though didn't remove. Will probably get rid of it when migrating to stage() (still thinking...  :eyes:)

efiop: @isidentical Makes sense! Btw, this is one of the points of confusion that `object` refactor was intended to clear up. We used to toss around hash_infos when we really meant to use objects. So if we assume that `known_hash` was needed here, we should've used `known_object`(or old_obj) instead. Just illustrating the point :wink: , we are on the same page otherwise.

dvc/remote/base.py
```diff
@@ -505,17 +507,37 @@ def pull(self, cache, named_cache, jobs=None, show_checksums=False):
...
-            from_fs,
             from_info,
+            from_fs,
+            "md5",
+            upload=upload,
             jobs=jobs,
             no_progress_bar=no_progress_bar,
         )
 
+        save(self.odb, obj)
```
isidentical, line 538: One thing that worries me is that, when `save()` calls `odb.add` with an `fs` of the same type (like both `fs` and `odb.fs` are on s3) then it `move()`s the file instead of `copy()` which contradicts with it's second behavior (just `upload_fobj`ing but not removing) so I guess we might need to introduce an option like `use_copy=...` to `save()` and `odb.add()` and use `self.fs.copy()` instead;

https://github.com/iterative/dvc/blob/2472c04873305fdfae5e93a0201d3af0d4689702/dvc/objects/db/base.py#L55-L59

CC: @efiop

efiop: @isidentical Good point! The purpose of that is to optimize for storage (e.g. `dvc add bigdatafile), but it indeed should be an explicit strategy option. Finding a good name for it is not easy though... Maybe `move=True/False` instead?

dvc/objects/db/base.py
```diff
@@ -53,7 +53,10 @@ def add(self, path_info, fs, hash_info, **kwargs):
         # using our makedirs to create dirs with proper permissions
         self.makedirs(cache_info.parent)
         if isinstance(fs, type(self.fs)):
-            self.fs.move(path_info, cache_info)
+            if move:
+                self.fs.move(path_info, cache_info)
+            else:
+                self.fs.copy(path_info, cache_info)
```
efiop, line 59: This one in preparation for https://github.com/iterative/dvc/issues/5811 , right? Maybe let's not add it for now and just fallback to `else: upload_fobj` that we have below.

dvc/objects/file.py
```diff
@@ -1,7 +1,6 @@
 import logging
 
 from .errors import ObjectFormatError
-from .stage import get_file_hash
```
efiop, line 4: This change seems unrelated.

dvc/objects/stage.py
```diff
@@ -1,91 +1,131 @@
...
 
     if not fs.exists(path_info):
         raise FileNotFoundError(
             errno.ENOENT, os.strerror(errno.ENOENT), path_info
         )
 
-    hash_info = _get_file_hash(path_info, fs, name)
+    target_info, obj = _get_file_hash(
+        path_info, fs, name, odb=odb, upload=upload
+    )
```
efiop, line 67: Unintuitive that `_get_file_hash` returns an object. There is probably no reason to pass `upload` to it, we can do it separately.

dvc/objects/stage.py
```diff
@@ -1,91 +1,131 @@
...
 
+    hash_info = obj.hash_info
     if state:
         assert ".dir" not in hash_info.value
-        state.save(path_info, fs, hash_info)
+        state.save(obj.path_info, obj.fs, obj.hash_info)
```
efiop, line 72: This one should actuallly stay the same, as this is us saving information about the original path_info and fs, notabout the temporary file on odb.fs

dvc/objects/stage.py
```diff
@@ -1,91 +1,131 @@
...
+
+    return target_info, obj
 
-    return hash_info
 
+def get_file_hash(path_info, fs, name, state=None, **kwargs):
+    _, obj = _get_file_obj(path_info, fs, name, state=state, **kwargs)
+    return obj.hash_info
```
efiop, line 79: How about we keep the `get_file_hash` as it was before, and then introduce smth like

```
def _get_file_obj(...):
    if upload:
       _upload_file(...)
    else
       hash_info = get_file_hash(...)
       obj = HashFile(...)
     return obj
```

Note that `_upload_file` is the one that could (and probably should) do `state.save()` for the new temporary object.

dvc/objects/db/base.py
isidentical: No, it is still needed. It removes the data from the original location if both the source and the remote storage have the same fs type.

dvc/objects/file.py
isidentical: No, this is for preventing circular import (it now imports this after modules are finalized)

efiop: Ah, makes sense. :+1:

dvc/objects/db/base.py
efiop: @isidentical I mean the `fs.copy` part. It could be:
```
if isinstance(fs, type(self.fs)) and move:
    self.fs.move()
else:
     with fs.open()...
         self.fs.upload_fobj()
```

or am I missing something?


On this topic, seems like we are finally reaching the point where we actually need to distinguish move from rename and copy from serverside-copy, so that we could provide the most efficient experience and nicest code. But it will do for now.

dvc/objects/stage.py
isidentical: > probably should

what would the benefit of doing a state.save() for it though? We still would have to recalculate the hash every time

dvc/objects/db/base.py
isidentical: That also works!

dvc/objects/stage.py
efiop: @isidentical Yeah, that was part of my original confusion. Just saying that if we do need to save it - we should do it in `_upload_file` logic. I agree that it is not needed for now.

However, it might be nice to state.save() for original file even in transfer case, so that if for some reason we fail later we won't have to download it again. Or if we'll run `update` - we'll be able to see if state.get() has an already available hash_info and we won't need to compute the hash again. It could be a separate PR though. Also since state only works for local files right now, the value for this is not that big.

dvc/objects/db/base.py
```diff
@@ -42,7 +42,7 @@ def get(self, hash_info):
             hash_info,
         )
 
-    def add(self, path_info, fs, hash_info, **kwargs):
+    def add(self, path_info, fs, hash_info, move=True, **kwargs):
```
efiop, line 45: For the record: `move` option is a bit clunky (same as `isinstance(fs, type(self.fs))`) but we kinda have to keep it for now, until we migrate to fsspec and will be able to distinguish rename/move copy/servercopy.

dvc/repo/add.py
```diff
@@ -173,12 +173,20 @@ def _process_stages(
...
+            obj = ostage(
+                out.odb,
+                from_fs.path_info,
+                from_fs,
+                "md5",
+                upload=True,
+                jobs=kwargs.get("jobs"),
             )
+            osave(out.odb, obj, move=False)
+            out.hash_info = obj.hash_info
```
efiop, line 189: I guess we could create a `out.transfer()` method, so we could reuse it with `odb=repo.cloud.odb` in --to-remote as well. But it could be done later in a followup.

efiop:
@isidentical Could you check the performance of transfer before and after this PR, please?

dvc/remote/base.py
```diff
@@ -505,17 +507,37 @@ def pull(self, cache, named_cache, jobs=None, show_checksums=False):
...
+        # When running import-url --to-remote / add --to-remote/-o ... we
+        # assume that it is unlikely that the odb will contain majority of the
+        # hashes, so we transfer everything as is (even if that file might
+        # already be in the cache) and don't waste an upload to scan the layout
+        # of the source location. But when doing update --to-remote, there is
+        # a high probability that the odb might contain some of the hashes, so
+        # we first calculate all the hashes (but don't transfer anything) and
+        # then only update the missing cache files.
+
+        upload = not (update and from_fs.isdir(from_info))
```
efiop, line 527: Why do we have to upload if we are updating a file though?

isidentical: because there is no point of calculating the hash, since we'd end up with downloading one time (calculating the hash) + odb.add()ing it (since it is unlikely that the file's hash won't be unchanged when you do update) which would cost download + upload. This optimization only works for directories

efiop: @isidentical Makes sense :+1: Btw, looks like if we would've saved to state during import, then we would be able to quickly tell that this file indeed changed and save us that downloading(for calculating the hash) that you've mentioned. Just a thought.

dvc/objects/stage.py
```diff
@@ -49,43 +66,64 @@ def get_file_hash(path_info, fs, name, state=None):
...
+                _get_file_obj,
+                fs=fs,
+                name=name,
+                odb=odb,
+                state=state,
+                upload=upload,
+            )
+        )
+        with ThreadPoolExecutor(
+            max_workers=kwargs.pop("jobs", fs.hash_jobs)
```
efiop, line 97: Btw, looks like we've lost
```
jobs = jobs or min((from_fs.jobs, odb.fs.jobs))
```
we used to have for transfer. We could bring it back here with `if upload` for now, to match previous behavior. But I wonder of there is a reasonable common approach here in general, looks like this finally pushes us to reconsider it (again, just a thought).

isidentical: > we used to have for transfer. We could bring it back here with if upload for now, to match previous behavior. But I wonder of there is a reasonable common approach here in general, looks like this finally pushes us to reconsider it (again, just a thought).

I also thought about it, and I think it makes sense to use `fs.hash_jobs` since we are actively calculating the hash in the meanwhile (but still the IO time is 2X than the normal [if you include upload], so maybe on `uplaod` it can be `fs.hash_jobs * 2`)

efiop: If we think about it, both jobs and hash_jobs are usually just N*CPU heuristics. Both uploading and hashing are likely IO bound, so there is probably a common number of jobs that benefits them. Once again this is probably a topic for research for after fsspec migration, since it will make it easier to experiment with that. But it is good that we are keeping it in mind.

isidentical:
Benchmark results: [master](https://gist.github.com/isidentical/1d15ad5c2d6a8052c13978223f23740c#file-master-txt), [this branch](https://gist.github.com/isidentical/1d15ad5c2d6a8052c13978223f23740c#file-this_branch-txt). The `import-url --to-remote` performance is reduced from 3.50 min to 2.10 min, and the `update --to-remote` performance is reduced from 9~ min to 3~ min.

dvc/objects/__init__.py
```diff
@@ -7,10 +8,22 @@
 logger = logging.getLogger(__name__)
 
 
-def save(odb, obj, **kwargs):
+def save(odb, obj, jobs=None, **kwargs):
```
isidentical, line 11: Other places in the codebase (e.g fetch) already pass `jobs=...` though `odb.add` doesn't use it (AFAIK). So we don't need to modify those

dvc/objects/__init__.py
```diff
@@ -7,10 +8,25 @@
...
+            for future in Tqdm(
+                as_completed(
+                    executor.submit(
+                        odb.add,
+                        entry.path_info,
+                        entry.fs,
+                        entry.hash_info,
+                        **kwargs
+                    )
+                    for _, entry in obj
```
efiop, line 23: Seems like you could `executor.map`, or am I missing something?

isidentical: We need to manually unpack some attributes of entry (`entry.path_info`, `entry.fs` etc) + add `**kwargs` to the calls. In theory we can do something like `executor.map(lambda entry: odb.add(...), ...)`, what would be the advantage with map though? It would just block the progress bar if there is some big file in the middle, which might not be the best UI.

efiop: Seemed nicer and shorter. E.g. see how we wrap the worker in tqdm in dvc/objects/stage.py, where we use it in map, so it doesn't block the pbar.

efiop:
@isidentical Btw, very interesting failures on windows. Looks like we have the same objects (e.g. two files with the same content) saved in parallel, which results in that move exception. We need to simply ignore it, as it is an expected behavior for when someone (e.g. another user) is saving stuff in parallel to the same cache dir.

dvc/objects/__init__.py
```diff
@@ -7,10 +8,25 @@
...
+                ),
+                total=len(obj),
+                unit="file",
```
efiop, line 26:
```suggestion
                desc="Saving files",
                total=len(obj),
                unit="file",
```

efiop (approved):
Looks great! :fire: As we've discussed above, there are a few things like using `verify` in `odb.add` and putting stage()&save() logic somewhere in one method, but let's deal with those on top.

PR #5778, opened 2021-04-07
serialize: py: try to track segments of the source

### LINKED ISSUE #5777: exp run --set-param fails to update param value
# Bug Report

`exp run --set-param params.py:TrainConfig.TEMP=0.0001 : does not update TrainConfig.TEMP=1e-8`

## Description

When trying to run an experiment --set-param does not update class variable TrainConfig.TEMP=1e-8 to TrainConfig=0.0001 declared and defined in params.py; however, --set-param does update TrainConfig.TEMP=0.0001 to TrainConfig:TEMP=1e-8. 

### Reproduce

Example:

1. Consider the Data Pipelines tutorial example with all the stages already defined in the dvc.yaml file. The yaml file contains stages prepare, featurize, train and evaluate.
2. Besides the params.yaml file also utilise the example params.py python file, provided [here](https://dvc.org/doc/command-reference/params#examples-python-parameters-file)
3. Add a variable TEMP=1e-8 to TrainConfig class in params.py file
```
# All standard variable types are supported.
BOOL = True
INT = 5
FLOAT = 0.001
STR = 'abc'
DICT = {'a': 1, 'b': 2}
LIST = [1, 2, 3]
SET = {4, 5, 6}
TUPLE = (10, 100)
NONE = None

# DVC can retrieve class constants and variables defined in __init__
class TrainConfig:

    EPOCHS = 70
    TEMP = 1e-8

    def __init__(self):
        self.layers = 5
        self.layers = 9  # TrainConfig.layers param will be 9
        self.sum = 1 + 2  # Will NOT be found due to the expression
        bar = 3  # Will NOT be found since it's locally scoped


class TestConfig:

    TEST_DIR = 'path'
    METRICS = ['metric']
```
4. Add below parameter dependencies to your Data Pipelines dvc.yaml file, such that your train stage looks like below
```
train:
    cmd: python src/train.py data/features model.pkl
    deps:
    - data/features
    - src/train.py
    params:
    - train.min_split
    - train.n_est
    - train.seed
    - params.py:
      - BOOL
      - INT
      - FLOAT
      - TrainConfig.EPOCHS
      - TrainConfig.TEMP
      - TrainConfig.layers
    outs:
    - model.pkl
```
5. dvc repro --force
6. dvc exp run --set-param params.py:TrainConfig.TEMP=0.0001
### Expected

A new experiment with param value TrainConfig.TEMP=0.0001
### Environment information

**Output of `dvc doctor`:**

```console
DVC version: 2.0.15 (pip)
Platform: Python 3.9.1 on Linux-5.4.0-70-generic-x86_64-with-glibc2.31
Supports: azure, hdfs, http, https
Cache directory: ecryptfs on /home/<username>/.Private
Caches: local
Remotes: local
Workspace directory: ecryptfs on /home/<username>/.Private
Repo: dvc, git
```

skshetry:
Fix #5777

We are working internally with the parsed object. So, as an example, if we have `1e-8` as a value in the code, we get it as `1e-08` which we try to replace later, but actually, the text in the code is `1e-8` due to which it fails. 

Looking at the code, I'd really love to see AST based solution (parse into ast -> modify ast -> dump ast) than the current solution where we:

1. Parse into AST
2. Convert AST to a dictionary. Also add some stuffs of interests into it such as original source, lineno etc.
3. Modify dict.
4. Modify the source based on the new dictionary (by replacing the old value with the new one).
5. Dump the source (with `ast.parse` sanity check).

It fails on step 4, as the old value that we get is parsed to 1e-08 which we are trying to replace, instead of the actual 1e-8. With this PR, it also keeps track of the code segments of the value in step 2 and then uses it later when modifying on step 4.  It is not perfect though, but it seemed to work nicely (as before). It does fallback to the original behavior if it does not find any segment.

#### TODO
- [ ] Error if the value that needs to be modified was not successfull.
- [ ] Supporting composite/nested items.
- [ ] Tests

### DIFF
--- a/dvc/utils/serialize/_py.py
+++ b/dvc/utils/serialize/_py.py
@@ -168,11 +168,8 @@ def _ast_assign_to_dict(
         v = assign.value
         offsets = slice(v.col_offset, v.end_col_offset)
         lno = assign.lineno - 1
-        result[name] = {
-            "lineno": lno,
-            "value": value,
-            "segment": src_lines[lno][offsets],
-        }
+        segment = src_lines.get(lno, []).get(offsets)
+        result[name] = {"lineno": lno, "value": value, "segment": segment}
     else:
         result[name] = value
 
--- a/dvc/utils/serialize/_py.py
+++ b/dvc/utils/serialize/_py.py
@@ -23,7 +23,8 @@ def parse_py(text, path):
     with reraise(SyntaxError, PythonFileCorruptedError(path)):
         tree = ast.parse(text, filename=path)
 
-    result = _ast_tree_to_dict(tree)
+    lines = text.splitlines()
+    result = _ast_tree_to_dict(tree, lines)
     return result
 
 
@@ -32,8 +33,9 @@ def parse_py_for_update(text, path):
     with reraise(SyntaxError, PythonFileCorruptedError(path)):
         tree = ast.parse(text, filename=path)
 
-    result = _ast_tree_to_dict(tree)
-    result.update({_PARAMS_KEY: _ast_tree_to_dict(tree, lineno=True)})
+    lines = text.splitlines()
+    result = _ast_tree_to_dict(tree, lines)
+    result.update({_PARAMS_KEY: _ast_tree_to_dict(tree, lines, lineno=True)})
     result.update({_PARAMS_TEXT_KEY: text})
     return result
 
@@ -53,10 +55,13 @@ def _dump(data, stream):
             if isinstance(value, dict):
                 lines = _update_lines(lines, old_dct[key], value)
             elif value != old_dct[key]["value"]:
+                old_value = old_dct[key]["value"]
                 lineno = old_dct[key]["lineno"]
-                lines[lineno] = lines[lineno].replace(
-                    f" = {old_dct[key]['value']}", f" = {value}"
-                )
+
+                segment = old_dct[key].get("segment")
+                old_segment = " = {}".format(segment or old_value)
+                new_segment = " = {}".format(value)
+                lines[lineno] = lines[lineno].replace(old_segment, new_segment)
             else:
                 continue
         return lines
@@ -86,7 +91,7 @@ def modify_py(path, fs=None):
         yield d
 
 
-def _ast_tree_to_dict(tree, only_self_params=False, lineno=False):
+def _ast_tree_to_dict(tree, src_lines, only_self_params=False, lineno=False):
     """Parses ast trees to dict.
 
     :param tree: ast.Tree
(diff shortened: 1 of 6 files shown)


dvc/utils/serialize/_py.py
```diff
@@ -23,7 +23,8 @@ def parse_py(text, path):
     with reraise(SyntaxError, PythonFileCorruptedError(path)):
         tree = ast.parse(text, filename=path)
 
-    result = _ast_tree_to_dict(tree)
+    lines = text.splitlines()
```
skshetry, line 26: @isidentical, would love to hear your thoughts on this module on how we can improve. Not really a priority for now though.

tests/unit/utils/serialize/test_python.py
```diff
@@ -0,0 +1,50 @@
...
+        "-1e+4",
+        "-1.0e4",
+        "-1.0e+4",
+        "-1.0e+04",
+        "-1e-4",
+        "-1.0e-4",
+        "-1.0e-04",
+    ],
+)
+def test_modify_override_floats(tmp_dir, val):
```
skshetry, line 33: This test seems wrong. It passes on older version of code as well.

dvc/utils/serialize/_py.py
isidentical: It would've been much simpler if we could used end_lineno/end_col_offsets (just like you did here) though they are only present in 3.8+ (instead of manually managing lines/location slices, there is even `ast.get_source_segment`). 

There are a few options (depending on how complex a system that we might want to go, but from what I see this is just a convenience module so I don't think there is much need for extreme stuff), but it might be easy for to just record all changes (a list of change objects `[Change(replacement='new_val', lineno=3, col_offset=20), ...]`) and then apply them one by one over the tokens (`tokenize.generate_tokens` + `tokenize.untokenize`). I might write a draft PR for that, after I get through how this works.

isidentical: @skshetry do we support replacing containers (like `A = [1,2,3,4]`, `-S a=[5,6,7,8]` or dicts etc) or only literals (like strings / numbers etc)?

skshetry: > do we support replacing containers

It seems we do, but only some of the times, otherwise it fails.

isidentical: It would have been simpler if we didn't support containers (actually anything represented as multi-token), though I assume we could still support selected number of literals (by keeping a stack of brackets/parens and selecting tokens until those brackets are balanced). 

Here is a simple snippet that would replace a segment of code with another only with known start location (for single tokens, one exception we might introduce later would be complex numbers `1 + 0j` etc). If we want to support containers, then would require a bit of more logic but still doable.

```py
import ast
import io
import tokenize
from itertools import chain
from dataclasses import dataclass

@dataclass
class Change:
    value: str
    lineno: int
    col_offset: int

    @classmethod
    def from_node(cls, node, value):
        return cls(value, node.lineno, node.col_offset)

def shift_tokens(tokens, start, offset):
    for index, token in enumerate(tokens[start:]):
        start_line, start_col, end_line, end_col = *token.start, *token.end
        tokens[start + index] = token._replace(
            start=(start_line, start_col + offset),
            end=(end_line, end_col + offset)
        )
    
def apply_changes(source, changes):
    buffer = io.StringIO(source)
    token_groups = defaultdict(list)
    for token in tokenize.generate_tokens(buffer.readline):
        token_groups[token.start[0]].append(token)

    offset_map = defaultdict(int)
    for change in changes:
        token_group = token_groups[change.lineno]
        start_token, end_token = None, None
        for index, token in enumerate(token_group):
            if token.start[1] == change.col_offset + offset_map[change.lineno]:
                break
        else:
            raise ValueError('corrupted ...')
        
        offset = len(change.value) - token.end[1]
        token_group[index] = token._replace(
            string=change.value,
            end=(token.end[0], len(change.value))
        )
        shift_tokens(token_group, index + 1, offset)
        offset_map[change.lineno] += offset
    
    tokens = chain.from_iterable(token_groups.values())
    return tokenize.untokenize(tokens)

def find_assign_value(tree, target_name):
    for node in ast.walk(tree):
        if isinstance(node, ast.Assign):
            assert len(node.targets) == 1
            [target] = node.targets
        elif isinstance(node, ast.AnnAssign):
            target = node.target
        else:
            continue
        
        if isinstance(target, ast.Name) and target.id == target_name:
            return node.value

source = """\
# imagine this is a file
temp = 1e-8
"""

tree = ast.parse(source)
node = find_assign_value(tree, 'temp')
change = Change.from_node(node, '0.0001')
print(apply_changes(source, [change]))
```

PR #5788, opened 2021-04-09
fsspec: update to 0.9 (including remotes)

isidentical:
With the adlfs's 0.7.1 release, we should be fine to update fsspec to 0.9 which would allow us to use newer versions of s3fs and gcsfs at the same time with adlfs.


efiop (approved):
:tada:

Issue #5645, opened 2021-03-18
Hadoop with Kerberos authentication


TM-Rin:
Greetings, fellow developers.
I am requesting adding a feature to accept `kerb_ticket` path while connecting to HDFS.
[PyArrow docs](http://arrow.apache.org/docs/python/generated/pyarrow.fs.HadoopFileSystem.html#pyarrow.fs.HadoopFileSystem)
Right now DVC does not use custom kerberos ticket paths and connecting with NULL ticket.

pared:
Seems like we should acknowlege the `kerb_ticket`, probably [here](https://github.com/iterative/dvc/blob/6026573860de2e395e4a7cdc3ab4b1739823af83/dvc/fs/hdfs.py#L98)

pared:
Related: #2283

naveen24691:
I have gone through the issues and related issues. It seems to be straight forward. I have raised a PR https://github.com/iterative/dvc/pull/5809  for this. Let me know if you are still planning to integrate this in the main branch.

PR #5809, opened 2021-04-13
Add support for kerberos token based auth

naveen24691:

Closes https://github.com/iterative/dvc/issues/5645

I have tested the kereberos authentication on my local Ambari hadoop cluster. I haven't written any unit test cases as there was no existing test cases for the HDFS connection. I think may be because of remote connections.

Edit: These failed tests doesn't seem to be specific to my changes. Can you please help me with some pointer here?

### DIFF
--- a/dvc/fs/hdfs.py
+++ b/dvc/fs/hdfs.py
@@ -84,8 +84,10 @@ class HDFSFileSystem(BaseFileSystem):
             path=parsed.path,
         )
 
+        self.kerb_ticket = config.get("kerb_ticket")
+
     @staticmethod
-    def hdfs(path_info):
+    def hdfs(path_info, kerb_ticket=None):
         import pyarrow.fs
 
         # NOTE: HadoopFileSystem is not meant to be closed by us and doesn't
@@ -100,7 +102,7 @@ class HDFSFileSystem(BaseFileSystem):
             path_info.host,
             path_info.port,
             user=path_info.user,
-            kerb_ticket=path_info.kerb_ticket
+            kerb_ticket=kerb_ticket
         )
 
     @contextmanager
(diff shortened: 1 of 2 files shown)


efiop:
Hi @naveen24691 !

Thanks for the PR!

Please check the CI, you need to fix styling of your PR and also looks like most of hdfs tests broke.

naveen24691:
Thanks, Yes I just realized that some of the test cases for HDFS is failing. I am working on the fix.

dvc/fs/hdfs.py
```diff
@@ -100,14 +102,15 @@ def close(self):
             path_info.host,
             path_info.port,
             user=path_info.user,
+            kerb_ticket=kerb_ticket,
```
efiop, line 105: Seems like you should be able to simply
```suggestion
            kerb_ticket=self.kerb_ticket,
```
and avoid the rest of the logic. Or does something break because of it?

efiop: If you are worried about `@staticmethod`, feel free to remove it, it is not used for any particular reason other that we didn't use any internal state before (aka `self`).

naveen24691: hdfs is a static method. I cant use the self directly. Hence the rest of the logic

naveen24691: Okay thanks.. I will remove the  staticmethod.

efiop:
@naveen24691 Thank you! Looks great! :fire: 

Could you also submit a doc PR to https://github.com/iterative/dvc.org , please?

naveen24691:
Yes, I will update that quickly. Thanks

efiop (approved):

naveen24691:
I have created the PR for the documentation: https://github.com/iterative/dvc.org/pull/2387

PR #5605, opened 2021-03-11
Dependency not consider dvcignore

karajan1001:
fixes #5256

1. add a new test
2. get rid of dvcignore in dependency

* [x] ❗ I have followed the [Contributing to DVC](https://dvc.org/doc/user-guide/contributing/core) checklist.

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here.

Thank you for the contribution - we'll try to review it as soon as possible. 🙏

Issue #5256, opened 2021-01-12
dependency: don't use dvcignore


efiop:
```
def test_run_dvcignored_dep(tmp_dir, dvc, run_copy):             
    tmp_dir.gen(                                                 
        {                                                        
            ".dvcignore": "dir\n",                               
            "dir": {"foo": "foo"},                               
        }                                                        
    )                                                            
    run_copy(                                                    
        os.path.join("dir", "foo"), "bar", name="copy-foo-to-bar"
    )                                                            
    assert (tmp_dir / "bar").read_text() == "foo"                
```

results in

```
/home/efiop/git/dvc/tests/dir_helpers.py:332: in run_copy                                                 
    return dvc.run(                                                                                       
/home/efiop/git/dvc/dvc/repo/__init__.py:49: in wrapper                                                   
    return f(repo, *args, **kwargs)                                                                       
/home/efiop/git/dvc/dvc/repo/scm_context.py:4: in run                                                     
    result = method(repo, *args, **kw)                                                                    
/home/efiop/git/dvc/dvc/repo/run.py:113: in run                                                           
    stage = create_stage(                                                                                 
/home/efiop/git/dvc/dvc/stage/__init__.py:84: in create_stage                                             
    fill_stage_dependencies(                                                                              
/home/efiop/git/dvc/dvc/stage/utils.py:97: in fill_stage_dependencies                                     
    stage.deps += loads_from(stage, deps or [], erepo=erepo)                                              
/home/efiop/git/dvc/dvc/dependency/__init__.py:95: in loads_from                                          
    return [_get(stage, s, info.copy()) for s in s_list]                                                  
/home/efiop/git/dvc/dvc/dependency/__init__.py:95: in <listcomp>                                          
    return [_get(stage, s, info.copy()) for s in s_list]                                                  
/home/efiop/git/dvc/dvc/dependency/__init__.py:81: in _get                                                
    return LocalDependency(stage, p, info)                                                                
/home/efiop/git/dvc/dvc/output/local.py:24: in __init__                                                   
    super().__init__(stage, path, *args, **kwargs)                                                        
/home/efiop/git/dvc/dvc/output/base.py:117: in __init__                                                   
    self._validate_output_path(path, stage)                                                               
[...]

PR #5812, opened 2021-04-14
Decoupling dvcignore and fs

karajan1001:
- [x] remove dvcignore from fs objects
- [x] attach dvcignore to repo
- [x] add ignore check to functions(add, checkout, output, pipeline, etc)
- [x] problem of subrepo
- [x] problem `.git` and `.dvc` ignore in fs
- [x] pass all tests
- [x] final check ( some new tests)
- [x] Performance guarantee.

### DIFF
--- a/dvc/checkout.py
+++ b/dvc/checkout.py
@@ -15,6 +15,7 @@ from dvc.objects.stage import stage
 from dvc.remote.slow_link_detection import (  # type: ignore[attr-defined]
     slow_link_guard,
 )
+from dvc.scheme import Schemes
 
 logger = logging.getLogger(__name__)
 
@@ -176,9 +177,11 @@ def _checkout_file(
 
 
 def _remove_redundant_files(path_info, fs, obj, cache, force):
-    existing_files = set(
-        fs.repo.dvcignore(fs.walk(path_info), walk_files=True)
-    )
+    if fs.scheme == Schemes.LOCAL and fs.repo:
+        walk_files = fs.repo.dvcignore(fs.walk(path_info), walk_files=True)
+    else:
+        walk_files = fs.walk_files(path_info)
+    existing_files = set(walk_files)
 
     needed_files = {path_info.joinpath(*key) for key, _ in obj}
     redundant_files = existing_files - needed_files
--- a/dvc/checkout.py
+++ b/dvc/checkout.py
@@ -203,11 +203,11 @@ def _checkout_dir(
     relink=False,
     dvcignore: Optional[DvcIgnoreFilter] = None,
 ):
-    modified = False, False
+    modified = False
     # Create dir separately so that dir is created
     # even if there are no files in it
     if not fs.exists(path_info):
-        modified = True  # type: ignore
+        modified = True
         fs.makedirs(path_info)
 
     logger.debug("Linking directory '%s'.", path_info)
@@ -223,14 +223,14 @@ def _checkout_dir(
             relink,
         )
         if entry_modified:
-            modified = True  # type: ignore
+            modified = True
 
     modified = (
         _remove_redundant_files(
             path_info, fs, obj, cache, force, dvcignore=dvcignore
         )
-        or modified,
-    )  # type: ignore
+        or modified
+    )
 
     fs.repo.state.save(path_info, fs, obj.hash_info)
 
--- a/dvc/checkout.py
+++ b/dvc/checkout.py
@@ -9,12 +9,14 @@ from dvc.exceptions import (
     ConfirmRemoveError,
     DvcException,
 )
+from dvc.ignore import DvcIgnoreFilter
 from dvc.objects import check, load
 from dvc.objects.errors import ObjectFormatError
 from dvc.objects.stage import stage
 from dvc.remote.slow_link_detection import (  # type: ignore[attr-defined]
     slow_link_guard,
 )
+from dvc.types import Optional
 
 logger = logging.getLogger(__name__)
 
@@ -175,8 +177,13 @@ def _checkout_file(
     return modified
 
 
-def _remove_redundant_files(path_info, fs, obj, cache, force):
-    existing_files = set(cache.repo.dvcignore.walk_files(fs, path_info))
+def _remove_redundant_files(
+    path_info, fs, obj, cache, force, dvcignore: Optional[DvcIgnoreFilter],
+):
+    if dvcignore:
+        existing_files = set(dvcignore.walk_files(fs, path_info))
+    else:
+        existing_files = set(fs.walk_files(path_info))
 
     needed_files = {path_info.joinpath(*key) for key, _ in obj}
     redundant_files = existing_files - needed_files
@@ -187,13 +194,20 @@ def _remove_redundant_files(path_info, fs, obj, cache, force):
 
 
 def _checkout_dir(
-    path_info, fs, obj, cache, force, progress_callback=None, relink=False,
+    path_info,
+    fs,
+    obj,
+    cache,
+    force,
+    progress_callback=None,
+    relink=False,
+    dvcignore: Optional[DvcIgnoreFilter] = None,
 ):
     modified = False, False
     # Create dir separately so that dir is created
     # even if there are no files in it
     if not fs.exists(path_info):
-        modified = True
+        modified = True  # type: ignore
         fs.makedirs(path_info)
 
     logger.debug("Linking directory '%s'.", path_info)
@@ -209,11 +223,14 @@ def _checkout_dir(
             relink,
         )
         if entry_modified:
-            modified = True
+            modified = True  # type: ignore
 
     modified = (
-        _remove_redundant_files(path_info, fs, obj, cache, force) or modified
-    )
+        _remove_redundant_files(
+            path_info, fs, obj, cache, force, dvcignore=dvcignore
+        )
+        or modified,
+    )  # type: ignore
 
     fs.repo.state.save(path_info, fs, obj.hash_info)
 
@@ -229,6 +246,7 @@ def _checkout(
     force=False,
     progress_callback=None,
     relink=False,
+    dvcignore: Optional[DvcIgnoreFilter] = None,
 ):
     if not obj.hash_info.isdir:
         ret = _checkout_file(
@@ -236,7 +254,14 @@ def _checkout(
         )
     else:
         ret = _checkout_dir(
-            path_info, fs, obj, cache, force, progress_callback, relink,
+            path_info,
+            fs,
+            obj,
+            cache,
+            force,
+            progress_callback,
+            relink,
+            dvcignore=dvcignore,
         )
 
     fs.repo.state.save_link(path_info, fs)
@@ -253,6 +278,7 @@ def checkout(
     progress_callback=None,
     relink=False,
     quiet=False,
+    dvcignore: Optional[DvcIgnoreFilter] = None,
 ):
     if path_info.scheme not in ["local", cache.fs.scheme]:
         raise NotImplementedError
@@ -269,7 +295,7 @@ def checkout(
         failed = path_info
 
     elif not relink and not _changed(path_info, fs, obj, cache):
-        logger.trace("Data '%s' didn't change.", path_info)
+        logger.trace("Data '%s' didn't change.", path_info)  # type: ignore
         skip = True
     else:
         try:
@@ -296,5 +322,12 @@ def checkout(
     logger.debug("Checking out '%s' with cache '%s'.", path_info, obj)
 
     return _checkout(
-        path_info, fs, obj, cache, force, progress_callback, relink,
+        path_info,
+        fs,
+        obj,
+        cache,
+        force,
+        progress_callback,
+        relink,
+        dvcignore=dvcignore,
     )
--- a/dvc/checkout.py
+++ b/dvc/checkout.py
@@ -15,7 +15,6 @@ from dvc.objects.stage import stage
 from dvc.remote.slow_link_detection import (  # type: ignore[attr-defined]
     slow_link_guard,
 )
-from dvc.scheme import Schemes
 
 logger = logging.getLogger(__name__)
 
@@ -177,12 +176,7 @@ def _checkout_file(
 
 
 def _remove_redundant_files(path_info, fs, obj, cache, force):
-    if fs.scheme == Schemes.LOCAL and cache.repo:
-        existing_files = set(
-            cache.repo.dvcignore.walk_files(fs.walk(path_info))
-        )
-    else:
-        existing_files = set(fs.walk_files(path_info))
+    existing_files = set(cache.repo.dvcignore.walk_files(fs, path_info))
 
     needed_files = {path_info.joinpath(*key) for key, _ in obj}
     redundant_files = existing_files - needed_files
--- a/dvc/checkout.py
+++ b/dvc/checkout.py
@@ -176,7 +176,11 @@ def _checkout_file(
 
 
 def _remove_redundant_files(path_info, fs, obj, cache, force):
-    existing_files = set(fs.walk_files(path_info))
+    if fs.repo:
+        ignore_filter = fs.repo.dvcignore
+    else:
+        ignore_filter = None
+    existing_files = set(fs.walk_files(path_info, ignore_filter=ignore_filter))
 
     needed_files = {path_info.joinpath(*key) for key, _ in obj}
     redundant_files = existing_files - needed_files
--- a/dvc/checkout.py
+++ b/dvc/checkout.py
@@ -177,11 +177,12 @@ def _checkout_file(
 
 
 def _remove_redundant_files(path_info, fs, obj, cache, force):
-    if fs.scheme == Schemes.LOCAL and fs.repo:
-        walk_files = fs.repo.dvcignore(fs.walk(path_info), walk_files=True)
+    if fs.scheme == Schemes.LOCAL and cache.repo:
+        existing_files = set(
+            cache.repo.dvcignore.walk_files(fs.walk(path_info))
+        )
     else:
-        walk_files = fs.walk_files(path_info)
-    existing_files = set(walk_files)
+        existing_files = set(fs.walk_files(path_info))
 
     needed_files = {path_info.joinpath(*key) for key, _ in obj}
     redundant_files = existing_files - needed_files
--- a/dvc/checkout.py
+++ b/dvc/checkout.py
@@ -176,11 +176,9 @@ def _checkout_file(
 
 
 def _remove_redundant_files(path_info, fs, obj, cache, force):
-    if fs.repo:
-        ignore_filter = fs.repo.dvcignore
-    else:
-        ignore_filter = None
-    existing_files = set(fs.walk_files(path_info, ignore_filter=ignore_filter))
+    existing_files = set(
+        fs.repo.dvcignore(fs.walk(path_info), walk_files=True)
+    )
 
     needed_files = {path_info.joinpath(*key) for key, _ in obj}
     redundant_files = existing_files - needed_files
--- a/dvc/dvcfile.py
+++ b/dvc/dvcfile.py
@@ -145,8 +145,8 @@ class FileMixin:
         # 3. path doesn't represent a regular file
         # 4. when the file is git ignored
         if not self.exists():
-            is_ignored = self.repo.fs.exists(self.path)
-            raise StageFileDoesNotExistError(self.path, dvc_ignored=is_ignored)
+            file_exist = self.repo.fs.exists(self.path)
+            raise StageFileDoesNotExistError(self.path, dvc_ignored=file_exist)
 
         self._verify_filename()
         if not self.repo.fs.isfile(self.path):
--- a/dvc/dvcfile.py
+++ b/dvc/dvcfile.py
@@ -124,7 +124,8 @@ class FileMixin:
         return relpath(self.path)
 
     def exists(self):
-        return self.repo.fs.exists(self.path)
+        is_ignored = not self.repo.dvcignore.is_ignored(self.path)
+        return self.repo.fs.exists(self.path) and is_ignored
 
     def _is_git_ignored(self):
         return is_git_ignored(self.repo, self.path)
@@ -143,8 +144,8 @@ class FileMixin:
         # 2. filename is not a DVC file
         # 3. path doesn't represent a regular file
         # 4. when the file is git ignored
-        is_ignored = self.repo.dvcignore.is_ignored(self.path)
-        if not self.exists() or is_ignored:
+        if not self.exists():
+            is_ignored = self.repo.fs.exists(self.path)
             raise StageFileDoesNotExistError(self.path, dvc_ignored=is_ignored)
 
         self._verify_filename()
--- a/dvc/dvcfile.py
+++ b/dvc/dvcfile.py
@@ -124,7 +124,7 @@ class FileMixin:
         return relpath(self.path)
 
     def exists(self):
-        is_ignored = self.repo.dvcignore.is_ignored(self.path)
+        is_ignored = self.repo.dvcignore.is_ignored_file(self.path)
         return self.repo.fs.exists(self.path) and not is_ignored
 
     def _is_git_ignored(self):
@@ -145,7 +145,7 @@ class FileMixin:
         # 3. path doesn't represent a regular file
         # 4. when the file is git ignored
         if not self.exists():
-            dvc_ignored = self.repo.dvcignore.is_ignored(self.path)
+            dvc_ignored = self.repo.dvcignore.is_ignored_file(self.path)
             raise StageFileDoesNotExistError(
                 self.path, dvc_ignored=dvc_ignored
             )
--- a/dvc/dvcfile.py
+++ b/dvc/dvcfile.py
@@ -143,8 +143,8 @@ class FileMixin:
         # 2. filename is not a DVC file
         # 3. path doesn't represent a regular file
         # 4. when the file is git ignored
-        if not self.exists():
-            is_ignored = self.repo.fs.exists(self.path, use_dvcignore=False)
+        is_ignored = self.repo.dvcignore.is_ignored(self.path)
+        if not self.exists() or is_ignored:
             raise StageFileDoesNotExistError(self.path, dvc_ignored=is_ignored)
 
         self._verify_filename()
--- a/dvc/dvcfile.py
+++ b/dvc/dvcfile.py
@@ -124,8 +124,8 @@ class FileMixin:
         return relpath(self.path)
 
     def exists(self):
-        is_ignored = not self.repo.dvcignore.is_ignored(self.path)
-        return self.repo.fs.exists(self.path) and is_ignored
+        is_ignored = self.repo.dvcignore.is_ignored(self.path)
+        return self.repo.fs.exists(self.path) and not is_ignored
 
     def _is_git_ignored(self):
         return is_git_ignored(self.repo, self.path)
--- a/dvc/dvcfile.py
+++ b/dvc/dvcfile.py
@@ -145,8 +145,10 @@ class FileMixin:
         # 3. path doesn't represent a regular file
         # 4. when the file is git ignored
         if not self.exists():
-            file_exist = self.repo.fs.exists(self.path)
-            raise StageFileDoesNotExistError(self.path, dvc_ignored=file_exist)
+            dvc_ignored = self.repo.dvcignore.is_ignored(self.path)
+            raise StageFileDoesNotExistError(
+                self.path, dvc_ignored=dvc_ignored
+            )
 
         self._verify_filename()
         if not self.repo.fs.isfile(self.path):
--- a/dvc/fs/azure.py
+++ b/dvc/fs/azure.py
@@ -53,7 +53,7 @@ class AzureAuthError(DvcException):
     pass
 
 
-class AzureFileSystem(FSSpecWrapper):
+class AzureFileSystem(FSSpecWrapper):  # pylint:disable=abstract-method
     scheme = Schemes.AZURE
     PATH_CLS = CloudURLInfo
     PARAM_CHECKSUM = "etag"
--- a/dvc/fs/dvc.py
+++ b/dvc/fs/dvc.py
@@ -168,23 +168,21 @@ class DvcFileSystem(BaseFileSystem):  # pylint:disable=abstract-method
         for dname in dirs:
             yield from self._walk(root / dname, trie)
 
-    def walk(self, path_info, **kwargs):
+    def walk(self, top, topdown=True, onerror=None, **kwargs):
         from pygtrie import Trie
 
-        topdown = kwargs.pop("topdown", True)
-        onerror = kwargs.pop("onerror", None)
         assert topdown
-        root = PathInfo(os.path.abspath(path_info))
+        root = PathInfo(os.path.abspath(top))
         try:
             meta = self.metadata(root)
         except FileNotFoundError:
             if onerror is not None:
-                onerror(FileNotFoundError(path_info))
+                onerror(FileNotFoundError(top))
             return
 
         if not meta.isdir:
             if onerror is not None:
-                onerror(NotADirectoryError(path_info))
+                onerror(NotADirectoryError(top))
             return
 
         trie = Trie()
--- a/dvc/fs/dvc.py
+++ b/dvc/fs/dvc.py
@@ -168,21 +168,23 @@ class DvcFileSystem(BaseFileSystem):  # pylint:disable=abstract-method
         for dname in dirs:
             yield from self._walk(root / dname, trie)
 
-    def walk(self, top, topdown=True, onerror=None, **kwargs):
+    def walk(self, path_info, **kwargs):
         from pygtrie import Trie
 
+        topdown = kwargs.pop("topdown", True)
+        onerror = kwargs.pop("onerror", None)
         assert topdown
-        root = PathInfo(os.path.abspath(top))
+        root = PathInfo(os.path.abspath(path_info))
         try:
             meta = self.metadata(root)
         except FileNotFoundError:
             if onerror is not None:
-                onerror(FileNotFoundError(top))
+                onerror(FileNotFoundError(path_info))
             return
 
         if not meta.isdir:
             if onerror is not None:
-                onerror(NotADirectoryError(top))
+                onerror(NotADirectoryError(path_info))
             return
 
         trie = Trie()
--- a/dvc/fs/hdfs.py
+++ b/dvc/fs/hdfs.py
@@ -159,15 +159,15 @@ class HDFSFileSystem(BaseFileSystem):
         if not topdown:
             yield root, dirs, nondirs
 
-    def walk(self, path_info, **kwargs):
-        if not self.isdir(path_info):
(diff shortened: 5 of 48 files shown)


karajan1001:
Currently affected files ( with at least one `fs.walk`, `fs.isfile`, `fs.isdir` or `fs.exists` in it)
```
dvc/checkout.py
dvc/config.py
dvc/dvcfile.py
dvc/fs/fsspec_wrapper.py
dvc/fs/memory.py
dvc/fs/repo.py
dvc/ignore.py
dvc/objects/db/base.py
dvc/objects/db/local.py
dvc/objects/stage.py
dvc/objects/transfer.py
dvc/output/base.py
dvc/parsing/context.py
dvc/remote/base.py
dvc/repo/__init__.py
dvc/repo/add.py
dvc/repo/brancher.py
dvc/repo/collect.py
dvc/repo/diff.py
dvc/repo/ls.py
dvc/repo/metrics/show.py
dvc/repo/params/show.py
dvc/repo/plots/__init__.py
dvc/repo/stage.py
dvc/state.py
dvc/utils/fs.py
dvc/utils/serialize/_common.py
```

Can be classified into five levels
1. `fs` level:
```
dvc/fs/fsspec_wrapper.py
dvc/fs/memory.py
dvc/fs/repo.py
dvc/remote/base.py # push / pull
dvc/utils/fs.py
dvc/utils/serialize/_common.py
```
2. `object` level:
```
dvc/objects/db/base.py
dvc/objects/db/local.py
dvc/objects/stage.py
dvc/objects/transfer.py
```
3. `output` level
```
dvc/output/base.py
dvc/dependency # inheriant from output
dvc/state.py
```

4. `command` level:
```
dvc/checkout.py
dvc/repo/__init__.py
dvc/repo/add.py
dvc/repo/brancher.py
dvc/repo/collect.py
dvc/repo/diff.py
dvc/repo/ls.py
dvc/repo/metrics/show.py
dvc/repo/params/show.py
dvc/repo/plots/__init__.py
dvc/repo/stage.py
```

5. `file` operations:
```
dvc/config.py
dvc/dvcfile.py
dvc/parsing/context.py
```

Originally `dvcignore` check was done in fs level, and affect all of the file levels above. In this PR, we have two choices:

1. on `output` level.
2. on `command` level. 

Currently the behavior of `DVC` or more like in `output` level. Ignore checking not only affect untracked files but also those which had already been added to the repo. For example
```
$ dvc add a
$ echo a >> .dvcignore
$ dvc status
```
would throw an exception
```
ERROR: Path '<dvc_root>/a' is ignored by
.dvcignore:5:a
```

And in [documents](https://dvc.org/doc/user-guide/project-structure/dvcignore-files)
> ⚠️ dvc run and dvc repro might remove ignored files. If they are not produced by a pipeline stage, they can be lost permanently.

Keep in mind that when you add .dvcignore patterns that affect an existing output, its status will change and DVC will behave as if that affected files were deleted.

For comparison, in Git `.gitignore` only affect untracked files. The behavior of `gitignore` is more like on `command` level (this is actually what dulwich do), and this implementation is more simple and its behavior is easier to predict.

Should we keep our current logic or switch to a more Git-like one? 

BTW, even if we decided to switch, it shouldn't be included in this PR.
@efiop  @dberenbaum

karajan1001:
@efiop 
Besides the functional problem above, we have another two code problem:
1. `dvcignore` used to couple with sub-repo and they need to be decoupled now.
2. Without `dvcignore`, `.git`, `.dvc` directory are now exposed to the filesystem. We need a new way to mask them.

efiop:
@karajan1001 Great summary! Indeed, there is an inconsistency where dvcignore could ignore already tracked output, even though it really should behave like `git add -f` for `.gitignore`ed file. I think it is more of an accident caused by the current fs & dvcignore coupling and indeed could be changed. I'm not sure there are even tests that check for that behavior right now.

> dvcignore used to couple with sub-repo and they need to be decoupled now.

Most of user-facing commands don't really walk into subrepos when walking the host repo, but when given a path that is in subrepo - will indeed go into the subrepo (it is a bit confusing, but it is really a more simple logic), while we do have some tests that check for recursive host walk that has to walk into subrepos as well. The latter one could be considered to be changed.

> Without dvcignore, .git, .dvc directory are now exposed to the filesystem. We need a new way to mask them.

Hopefully most of such cases could be covered by just filtering out `fs.walk()` results with `.dvcignore`, so it won't traverse into .git and .dvc when it is not needed.

skshetry:
@efiop, @karajan1001, what are the reasons for the decoupling?  IIRC we depend on this behaviour in a lot of places 
(stage collection, parametrization, checkouts, add/repro, get/import), so I see this as a necessity in a lot of places.

(maybe what we need is a better API but is decoupling/duplication worth it?).

efiop:
@skshetry Bugs like https://github.com/iterative/dvc/pull/5605 and migration to fsspec. Those could be solved in the current arch, but it gets clunky, so trying to explore if more explicit approach of applying dvcignore only where it is needed (similar to git) would fit better. We might decide that it is not worth it too.

karajan1001:
> @karajan1001 Great summary! Indeed, there is an inconsistency where dvcignore could ignore already tracked output, even though it really should behave like `git add -f` for `.gitignore`ed file. I think it is more of an accident caused by the current fs & dvcignore coupling and indeed could be changed. I'm not sure there are even tests that check for that behavior right now.

So, it is not a designed behavior but accidentally behaves like this. 

> > dvcignore used to couple with sub-repo and they need to be decoupled now.
> 
> Most of user-facing commands don't really walk into subrepos when walking the host repo, but when given a path that is in subrepo - will indeed go into the subrepo (it is a bit confusing, but it is really a more simple logic), while we do have some tests that check for recursive host walk that has to walk into subrepos as well. The latter one could be considered to be changed.

`The latter one could be considered to be changed.` Sorry I didn't understand this.

efiop:
@karajan1001 I mean that we can change the second part - behaviour that is only tested by us in syntetic tests but not actually used in any user-facing functionality. E.g. an ability to fs.walk("host_root") and go through host repo and into the subrepo could be considered to be dropped, as it is not used by any dvc commands.

Have to point out that this is an open-ended task, if you see that the current approach works better - we can keep it.

dvc/fs/local.py
```diff
@@ -94,18 +64,14 @@ def walk(
...
-
+            dirs[:] = list(filter(lambda i: i not in self.HIDDEN_DIRS, dirs))
             yield os.path.normpath(root), dirs, files
 
     def walk_files(self, path_info, **kwargs):
-        for root, _, files in self.walk(path_info):
+        ignore_filter = kwargs.get("ignore_filter", None)
+        for root, dirs, files in self.walk(path_info):
+            if ignore_filter:
+                dirs, files = ignore_filter(root, dirs, files)
```
efiop, line 74: How about we just call `ignore_filter` outside, when we need it? E.g. by wrapping walk_files result with it. That way it won't leak into the fs level.

dvc/fs/local.py
```diff
@@ -21,70 +19,42 @@ class LocalFileSystem(BaseFileSystem):
     PARAM_CHECKSUM = "md5"
     PARAM_PATH = "path"
     TRAVERSE_PREFIX_LEN = 2
+    HIDDEN_DIRS = {".git", ".hg", ".dvc"}
```
efiop, line 22: Should this be part of dvcignore or some additional filter? Seems like it doesn't really belong on fs level.

karajan1001:
Current `subrepo` problem.

Originally `subrepo`'s `dvcignore` attached to its own `fs`. So each fs has its own `dvcignore` patterns on it, and can be controlled individually. Now the fs of `subrepo` is directly called by the root `repo` 's `fs`. And our `dvcginore` pattern is bound on `subrepo` ignore. We can't get `subrepo` files while ignoring `subrepo`'s patterns.

Two solutions:
1.  Root repo didn't get `subrepo`'s files directly.
2. Dvcginore decoupling patterns` ignore and `subrepo`'s ignore.

karajan1001:
| tests| previous version | after this PR| diff |
|:---:|:---:|:---:|:---:|
| add.Add.time_cats_dogs copy| 38.6±0.7s |  40.0±0.6s | 3.6% |
| add.Add.time_cats_dogs symlink| 30.9±0.1s |  30.1±0.6s | - 2.6% |
| add.Add.time_cats_dogs hardlink| 38.1±0.1s |   37.0±0.2s |  - 2.9% |
|checkout.CheckoutBench.time_cats_dogs copy |  16.8±0.4s | 16.8±0.3s | 0.0%|
|checkout.CheckoutBench.time_cats_dogs symlink |  8.27±0.4s | 8.45±0.04s | 2.1%|
|checkout.CheckoutBench.time_cats_dogs hardlink |  13.4±0.4s | 13.7±0.1s | 2.2% |
| collect.CollectBench.time_stages_collection | 2.58±0s | 2.18±0s | - 15.5 % |
| collect.TraverseGitRepoBench.time_repo_traversing |  4.14±0s |  5.28±0s | 27.1% |
| imports.ImportBench.time_imports | 13.9±0m | 14.1±0m | 1.4 % |
| init.InitNoScmBench.time_init | 177±5ms | 177±2ms | 0.0% |
| init.InitScmBench.time_init | 197±3ms | 196±1ms | - 0.5 %|
| push.PushBench.time_cats_dogs | failed | failed
| startup.StartupBench.time_startup | 107±0ms | 110±0ms | 2.8% |
| status.DVCIgnoreBench.time_status |403±0ms | 406±0ms |0.7% |
| status.DVCStatusBench.time_status | 401±0ms | 401±0ms | 0.0% |

Most of the tests' performance changed less than 4% after this PR. Only two of them show an obvious diff
**collect.CollectBench.time_stages_collection ** accelerates about 15%
**collect.TraverseGitRepoBench.time_repo_traversing** slows about 27%
Currently, I'm looking for the reason for this.

karajan1001:
> tests	previous version	after this PR	diff
> add.Add.time_cats_dogs copy	38.6±0.7s	40.0±0.6s	3.6%
> add.Add.time_cats_dogs symlink	30.9±0.1s	30.1±0.6s	- 2.6%
> add.Add.time_cats_dogs hardlink	38.1±0.1s	37.0±0.2s	- 2.9%
> checkout.CheckoutBench.time_cats_dogs copy	16.8±0.4s	16.8±0.3s	0.0%
> checkout.CheckoutBench.time_cats_dogs symlink	8.27±0.4s	8.45±0.04s	2.1%
> checkout.CheckoutBench.time_cats_dogs hardlink	13.4±0.4s	13.7±0.1s	2.2%
> collect.CollectBench.time_stages_collection	2.58±0s	2.18±0s	- 15.5 %
> collect.TraverseGitRepoBench.time_repo_traversing	4.14±0s	5.28±0s	27.1%
> imports.ImportBench.time_imports	13.9±0m	14.1±0m	1.4 %
> init.InitNoScmBench.time_init	177±5ms	177±2ms	0.0%
> init.InitScmBench.time_init	197±3ms	196±1ms	- 0.5 %
> push.PushBench.time_cats_dogs	failed	failed	
> startup.StartupBench.time_startup	107±0ms	110±0ms	2.8%
> status.DVCIgnoreBench.time_status	403±0ms	406±0ms	0.7%
> status.DVCStatusBench.time_status	401±0ms	401±0ms	0.0%
> Most of the tests' performance changed less than 4% after this PR. Only two of them show an obvious diff
> **collect.CollectBench.time_stages_collection** accelerates about 15%
> **collect.TraverseGitRepoBench.time_repo_traversing** slows about 27%
> Currently, I'm looking for the reason for this.

Extra time cost comes from two steps.
1. Now we update two trees, 
2. Time cost in searching all dirs (potential subrepo).

After solving these two problems. New version runs a bit faster than before. This might because we move subrepo finding from `__call__` to `__update__` which would run only once.

| tests| previous version | after this PR| after optimization |diff |
|:---:|:---:|:---:|:---:|
| collect.CollectBench.time_stages_collection | 2.67±0s | 2.33±0s | 2.27±0s | - 15.0 % |
| collect.TraverseGitRepoBench.time_repo_traversing |  4.14±0s |  5.37±0s | 3.96±0s | -4.4% |

pared (changes requested):

karajan1001:
Newest result.
| tests| previous version | after this PR| diff |
|:---:|:---:|:---:|:---:|
| add.Add.time_cats_dogs copy| 40.4±0.7s |  36.4±0.5s | -9% |
| add.Add.time_cats_dogs symlink| 30.8±0.08s |  28.8±0.01s | - 6.5% |
| add.Add.time_cats_dogs hardlink| 36.5±0.2s |   35.0±0.04s |  - 4.2% |
|checkout.CheckoutBench.time_cats_dogs copy |  15.6±0.2s | 15.4±0.02s | -1.1%|
|checkout.CheckoutBench.time_cats_dogs symlink |  7.69±0.02s | 7.59±0.01s | -1.4%|
|checkout.CheckoutBench.time_cats_dogs hardlink |  12.4±0.05s | 12.5±0.02s | 0.8% |
| collect.CollectBench.time_stages_collection | 2.49±0s | 2.15±0s | - 13.7 % |
| collect.TraverseGitRepoBench.time_repo_traversing |  4.09±0s |  3.87±0s | -5.4% |
| imports.ImportBench.time_imports | 14.6±0m | 14.8±0m | 1.3 % |
| init.InitNoScmBench.time_init | 175±0.9ms | 175±0.7ms | 0.0% |
| init.InitScmBench.time_init | 193±0.7ms | 194±0.7ms | 0.5 %|
| push.PushBench.time_cats_dogs | failed | failed
| startup.StartupBench.time_startup | 104±0ms | 104±0ms | 0.0% |
| status.DVCIgnoreBench.time_status |403±0ms | 401±0ms | -0.5% |
| status.DVCStatusBench.time_status | 402±0ms | 398±0ms | -1.0% |

tests/func/test_ignore.py
```diff
@@ -4,83 +4,89 @@
...
 
 
-def test_ignore(tmp_dir, dvc, monkeypatch):
-    tmp_dir.gen({"dir": {"ignored": "text", "other": "text2"}})
-    tmp_dir.gen(DvcIgnore.DVCIGNORE_FILE, "dir/ignored")
-
-    dvc.fs._reset()
+def _get_pathinfo_set(dvc: Repo, path: PathInfo) -> list:
+    root = PathInfo(dvc.root_dir)
+    ls_list = dvc.ls(root, path.relpath(root), recursive=True)
```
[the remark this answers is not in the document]
karajan1001: It is an outdated design.

tests/func/test_ignore.py
```diff
@@ -4,83 +4,89 @@
...
     tmp_dir.gen({"dir": {"ignored": "...", "other": "text"}})
 
     tmp_dir.gen(DvcIgnore.DVCIGNORE_FILE, "ignored*")
-    dvc.fs._reset()
+    dvc._reset()
+
+    root_path = PathInfo(tmp_dir)
 
-    mtime, size = get_mtime_and_size("dir", dvc.fs)
+    mtime, size = get_mtime_and_size(root_path / "dir", dvc.fs)
```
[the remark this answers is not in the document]
karajan1001: Seems that originally abs_path transfer was in `fs.walk()`, now I moved it to dvcignore. Actually, I don't think `dvcignore` would do this. This transfer ought to be unified at some other level.

dvc/checkout.py
```diff
@@ -176,7 +177,11 @@ def _checkout_file(
 
 
 def _remove_redundant_files(path_info, fs, obj, cache, force):
-    existing_files = set(fs.walk_files(path_info))
+    if fs.scheme == Schemes.LOCAL and fs.repo:
+        walk_files = fs.repo.dvcignore(fs.walk(path_info), walk_files=True)
+    else:
+        walk_files = fs.walk_files(path_info)
+    existing_files = set(walk_files)
```
efiop, line 184: `walk_files` should probably be `files` to avoid confusion with the function.

How about:

```suggestion
    files = set(cache.repo.dvcignore(fs, walk_files=True))
```

unlike `fs.repo`, `cache.repo` is more persistent (though temporarily) so we can rely on it more.

Or maybe even:
```
    files = set(cache.repo.dvcignore.walk_files(fs, path_info))
```

so that dvcignore has `walk` and `walk_files` methods.

Also, for the record, I'm refactoring this right now to decouble repo from fs, and `fs.repo` will go away soon, as well as `cache.repo`. We'll likely start passing `dvcignore`(as well as state) to checkout as kwargs. No action needed in this PR, just FYI.

dvc/fs/repo.py
```diff
@@ -340,18 +329,34 @@ def walk(
...
+        else:
+            yield from self.walk_fs(
+                top,
+                topdown=topdown,
+                onerror=onerror,
+                dvcfiles=dvcfiles,
+                **kwargs
+            )
 
+    def walk_fs(
```
efiop, line 354: No need to have it as a public method.

```suggestion
    def _walk_fs(
```

Or there might be a better name too, not sure.

dvc/fs/repo.py
```diff
@@ -340,18 +329,34 @@ def walk(
...
+                ignore_subrepos=ignore_subrepos,
+            )
+        else:
+            yield from self.walk_fs(
+                top,
+                topdown=topdown,
+                onerror=onerror,
+                dvcfiles=dvcfiles,
+                **kwargs
+            )
```
efiop, line 352: Duplicated code :slightly_frowning_face: Seems like dvcignore noop would come in handy in cleaning it up.

I also see that the approach with `dvcignore.walk(fs, path_info)` from above will run into a problem of `walk_fs` having a custom name, so you probably need to apply dvcignore inside `walk_fs`, for fs/dvcfs walks.

efiop: For the record: that existing walk logic for fs & dvc_fs is quite messy, we are aware of that, so just take a look if it is possible to wrap with dvcignore without messing with it too much. The code itself has been due for a refactor for awhile now :slightly_frowning_face:

dvc/fs/repo.py
```diff
@@ -366,8 +371,8 @@ def walk(
 
         yield from self._walk(repo_walk, dvc_walk, dvcfiles=dvcfiles)
 
-    def walk_files(self, top, **kwargs):  # pylint: disable=arguments-differ
-        for root, _, files in self.walk(top, **kwargs):
+    def walk_files(self, path_info, **kwargs):
+        for root, _, files in self.walk(path_info):
```
efiop, line 375: looks like we've lost `kwargs`. Is it intentiopnal?

dvc/repo/add.py
```diff
@@ -209,7 +209,7 @@ def _find_all_targets(repo, target, recursive):
         return [
             os.fspath(path)
             for path in Tqdm(
-                repo.fs.walk_files(target),
+                repo.dvcignore(repo.fs.walk(target), walk_files=True),
```
efiop, line 212: How about
```suggestion
                repo.dvcignore.walk_files(repo.fs, target),
```
?

dvc/repo/collect.py
```diff
@@ -43,7 +43,9 @@ def _collect_paths(
     for path_info in path_infos:
 
         if recursive and fs.isdir(path_info):
-            target_infos.extend(fs.walk_files(path_info))
+            target_infos.extend(
+                repo.dvcignore(fs.walk(path_info), walk_files=True)
```
efiop, line 47:
```suggestion
                repo.dvcignore.walk_files(fs, path_info)
```

dvc/utils/fs.py
```diff
@@ -36,7 +37,11 @@ def get_mtime_and_size(path, fs):
     if fs.isdir(path):
         size = 0
         files_mtimes = {}
-        for file_path in fs.walk_files(path):
+        if fs.scheme == Schemes.LOCAL and fs.repo:
+            walk_iterator = fs.repo.dvcignore(fs.walk(path), walk_files=True)
+        else:
+            walk_iterator = fs.walk_files(path)
+        for file_path in walk_iterator:
```
efiop, line 44: hm, okay, here is where getting `dvcignore` is a bit of a challenge. I think we should pass it explicitly here instead of relying on `fs.repo`, which will go away soon. Please take a look if it is simple to do that right now, and if not - I'll adjust it in my PR on top.

```suggestion
        for fpath in dvcignore.walk_files(fs, path):
```

dvc/fs/repo.py
karajan1001: Just forgot it.

dvc/checkout.py
karajan1001: How about the `RepoFileSystem`? Seems that it's difficult to be decoupled from `repo`.

dvc/fs/repo.py
karajan1001: > I also see that the approach with dvcignore.walk(fs, path_info) from above will run into a problem of walk_fs having a custom name, so you probably need to apply dvcignore inside walk_fs, for fs/dvcfs walks.

Sorry, I don't get it. Currently, the `dvcignore` can work correctly in `subrepo`s. We don't need to pass `fs` arguments to it.

dvc/checkout.py
karajan1001: In my opinion, maybe we should pass a list of individual files instead of path and file systems other things in the last three cases.

dvc/fs/repo.py
efiop: @karajan1001 Actually not really. There is no need to wrap `self._walk`, for example. You just need to wrap `fs.walk`, as `dvc_fs.walk` already used dvcignore when it built the dag, so there is no reason to apply it again.

efiop: Also, if we'll be able to pull that off ^^ , then we could switch to:

```
dvcignore.walk(fs, path)
dvcignore.walk_files(fs, path)
```

instead of current

```
dvcignore.walk(fs.walk(path))
dvcignore.walk_files(fs.walk(path))
```

Unless I'm missing something, of course.

dvc/checkout.py
efiop: > In my opinion, maybe we should pass a list of individual files instead of path and file systems other things in the last three cases.

Could you elaborate, please?

dvc/checkout.py
```diff
@@ -176,7 +177,12 @@ def _checkout_file(
 
 
 def _remove_redundant_files(path_info, fs, obj, cache, force):
-    existing_files = set(fs.walk_files(path_info))
+    if fs.scheme == Schemes.LOCAL and cache.repo:
+        existing_files = set(
+            cache.repo.dvcignore.walk_files(fs.walk(path_info))
+        )
+    else:
+        existing_files = set(fs.walk_files(path_info))
```
efiop, line 185: Discusssed with @karajan1001 that we can move this check into `dvcignore` and do something like `dvcignore.walk_files(fs, path_info)` and let it ignore non-local filesystems

pared (commented):
LGTM, though I would like to take one more review tomorrow

dvc/fs/dvc.py
```diff
@@ -168,21 +168,23 @@ def _walk(self, root, trie, topdown=True, **kwargs):
         for dname in dirs:
             yield from self._walk(root / dname, trie)
 
-    def walk(self, top, topdown=True, onerror=None, **kwargs):
+    def walk(self, path_info, **kwargs):
```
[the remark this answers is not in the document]
karajan1001: `walk(top, topdown=True, onerror=None, followlinks=False)` os.walk is in this form.

efiop:
@karajan1001 Please rebase.

efiop:
@karajan1001 Could you rebase again, please? Sorry for that :slightly_frowning_face:

dvc/dvcfile.py
```diff
@@ -144,8 +145,8 @@ def _load(self):
         # 3. path doesn't represent a regular file
         # 4. when the file is git ignored
         if not self.exists():
-            is_ignored = self.repo.fs.exists(self.path, use_dvcignore=False)
-            raise StageFileDoesNotExistError(self.path, dvc_ignored=is_ignored)
+            file_exist = self.repo.fs.exists(self.path)
+            raise StageFileDoesNotExistError(self.path, dvc_ignored=file_exist)
```
pared, line 149: This should not look like that. In `exists` we have two statements (`exists and is_not_dvcignored`). Here we check `exists` and infer the information whether file is ignored. But to infer this knowledge we need to know how `exists` works. We could simply check here again if path is `dvcignored` and provide that into error constructor.

dvc/objects/stage.py
```diff
@@ -78,6 +78,11 @@ def _get_file_obj(path_info, fs, name, odb=None, state=None, upload=False):
 
 
 def _build_objects(path_info, fs, name, odb, state, upload, **kwargs):
+    dvcignore = kwargs.get("dvcignore", None)
```
pared, line 81:
```suggestion
def _build_objects(path_info, fs, name, odb, state, upload, dvcignore=None, **kwargs):
```

dvc/output/base.py
```diff
@@ -195,17 +196,35 @@ def get_hash(self):
...
+            self.fs,
+            self.odb.fs.PARAM_CHECKSUM,
+            dvcignore=self.dvcignore,
         ).hash_info
 
     @property
     def is_dir_checksum(self):
         return self.hash_info.isdir
 
+    def is_ignored(self, path) -> bool:
```
pared, line 213: I understand that later we need to use it to check `abs_path` but `is_ignored` method in context of `output` does not seem to make sense. Output is a concept of some singular asset and methods like `exists` `is_ignored` are regarding the output itself. I think that `is_ignored` should be implemented for `self.path_info` and in case of `_validate_output_paths` we should be using `dvcignore` object

pared (changes requested):
Some minor things. Looks really good. The amount of removed code and logic is amazing!

dvc/checkout.py
```diff
@@ -269,7 +295,7 @@ def checkout(
         failed = path_info
 
     elif not relink and not _changed(path_info, fs, obj, cache):
-        logger.trace("Data '%s' didn't change.", path_info)
+        logger.trace("Data '%s' didn't change.", path_info)  # type: ignore
```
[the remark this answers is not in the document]
efiop: @pared That will stringify it immediately, which defeats the purpose of lazy logging :slightly_frowning_face:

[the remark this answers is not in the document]
pared: Ahh, right, sorry @karajan1001 please disregard this one.

dvc/fs/hdfs.py
```diff
@@ -159,15 +159,15 @@ def _walk(self, hdfs, root, topdown=True):
         if not topdown:
             yield root, dirs, nondirs
 
-    def walk(self, path_info, **kwargs):
-        if not self.isdir(path_info):
+    def walk(self, top, topdown=True, onerror=None, **kwargs):
```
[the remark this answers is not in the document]
karajan1001: But the `pylint` suppressions are ugly. Do we have any methods to solve this?

dvc/output/base.py
karajan1001: @pared , Yes, and so I renamed it here.

dvc/output/base.py
```diff
@@ -195,17 +196,35 @@ def get_hash(self):
...
+            self.fs,
+            self.odb.fs.PARAM_CHECKSUM,
+            dvcignore=self.dvcignore,
         ).hash_info
 
     @property
     def is_dir_checksum(self):
         return self.hash_info.isdir
 
+    def check_path_dvcignore(self, path) -> bool:
```
pared, line 213: I think we can make it with underscore (`_check_path_dvcignore`), so that its not exposed outside, since we are using it internally for now.

pared (approved):
1 minor thingy, not a blocker

karajan1001:
Add new tests for #5256, and verified it would be solved automatically after this PR. Close #5605 for this.

karajan1001:
```
dvc/fs/s3.py:17:0: W0223: Method 'walk' is abstract in class 'BaseFileSystem' but is not overridden (abstract-method)
```
Any methods to suppress this other than using `# pylint`?

karajan1001:
> > Any methods to suppress this other than using # pylint?
> 
> I guess we need to override it and `raise NotImplementedError`

I remember that an abstract method must be implemented ( can't be abstract ) in subclasses in some other languages.

karajan1001:
> > > Any methods to suppress this other than using # pylint?
> > 
> > 
> > I guess we need to override it and `raise NotImplementedError`
> 
> I remember that an abstract method must be implemented ( can't be abstract ) in subclasses in some other languages.

image
Hmmm, It didn't work as well in `pylint`.

dvc/ignore.py
```diff
@@ -253,25 +262,50 @@ def _update_sub_repo(self, path):
...
+    def walk(self, fs: BaseFileSystem, path_info: AnyPath, **kwargs):
+        ignore_subrepos = kwargs.pop("ignore_subrepos", True)
+        if fs.scheme == Schemes.LOCAL:
+            for root, dirs, files in fs.walk(path_info, **kwargs):
+                dirs[:], files[:] = self(
+                    root, dirs, files, ignore_subrepos=ignore_subrepos
+                )
+                yield root, dirs, files
+        else:
+            yield from fs.walk(path_info)
```
efiop, line 285:
```suggestion
            yield from fs.walk(path_info, **kwargs)
```

dvc/repo/fetch.py
```diff
@@ -111,14 +111,7 @@ def cb(result):
...
                     pass
-                obj = stage(
-                    odb,
-                    path_info,
-                    repo.repo_fs,
-                    "md5",
-                    jobs=jobs,
-                    follow_subrepos=False,
-                )
+                obj = stage(odb, path_info, repo.repo_fs, "md5", jobs=jobs,)
```
efiop, line 114: There's an interesting question of whether we should apply dvcignore here instead of relying on it being applied in `repo_fs` :thinking: For now we could keep it as is, but it might be a good idea for the future.

dvc/output/base.py
```diff
@@ -195,17 +196,35 @@ def get_hash(self):
...
     def is_dir_checksum(self):
         return self.hash_info.isdir
 
+    def _check_path_dvcignore(self, path) -> bool:
+        if (
+            self.scheme == Schemes.LOCAL
+            and self.IS_DEPENDENCY is False
+            and self.repo
+        ):
+            if self.repo.dvcignore.is_ignored(path, ignore_subrepos=False):
```
efiop, line 219: We should probably pass `fs` to `is_ignored` too, so it could decide to just shortcircuit if it is not a local filesystem.

efiop: Also, should we use `self.dvcignore` here instead of `self.repo.dvcignore`?

dvc/utils/fs.py
```diff
@@ -30,13 +30,17 @@ def get_inode(path):
     return inode
 
 
-def get_mtime_and_size(path, fs):
+def get_mtime_and_size(path, fs, dvcignore=None):
     import nanotime
 
     if fs.isdir(path):
```
efiop, line 36: This should probably also use dvcignore, right?

efiop:
@karajan1001 @pared We are in a bit of a mess with legacy `walk()` and fsspec migration, the end goal is fsspec's `walk()`.  For now just adding walk to fsspec_wrapper and adding a pylint ignore to that line is fine, we'll adjust it in a followup.

dvc/output/base.py
```diff
@@ -248,14 +267,22 @@ def changed(self):
...
+        return None
+
     @property
     def is_empty(self):
         return self.fs.is_empty(self.path_info)
 
     def isdir(self):
+        if self._check_path_dvcignore(self.path_info):
+            return False
         return self.fs.isdir(self.path_info)
```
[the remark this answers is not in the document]
karajan1001: @efiop , `dvcignore` didn't test where the path is a file or a dir. The only difference of dirs and files are the matching of pattern like `paths/`.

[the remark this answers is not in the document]
efiop: @karajan1001 Right, I'm saying that you are using `is_ignored` which will interpret `path` as a local path and will try to use `os.path.isdir` https://github.com/iterative/dvc/blob/e7571ab3251ef4417fa196e7ad45b7f6f8a42ee3/dvc/ignore.py#L354 on it, while it should use `fs.isdir`, because `fs` might be a GitFileSystem, for example.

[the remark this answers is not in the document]
efiop: In general, right now `dvcignore` should only work for `fs`(and `path`) that only belong to `dvcignore.fs`(filesystem that dvcignore was collected from).

dvc/fs/azure.py
```diff
@@ -53,7 +53,7 @@ class AzureAuthError(DvcException):
     pass
 
 
-class AzureFileSystem(FSSpecWrapper):
+class AzureFileSystem(FSSpecWrapper):  # pylint:disable=abstract-method
```
efiop, line 56: There is a less dangerous way: you could add `walk` to `FSSpecWrapper`.

tests/unit/fs/test_repo.py
```diff
@@ -465,17 +467,6 @@ def test_repo_fs_no_subrepos(tmp_dir, dvc, scm):
...
-    assert fs.isfile(tmp_dir / "dir" / "repo" / "foo") is False
-    assert fs.isdir(tmp_dir / "dir" / "repo") is False
-    assert fs.isdir(tmp_dir / "dir") is True
-
-    assert fs.isdvc(tmp_dir / "lorem") is True
-    assert fs.isdvc(tmp_dir / "dir" / "repo" / "dir1") is False
-
-    assert fs.exists(tmp_dir / "dir" / "repo.txt") is True
-    assert fs.exists(tmp_dir / "repo" / "ipsum") is False
-
```
efiop, line 478: Hm, this is actually what we were talking about in https://github.com/iterative/dvc/pull/5812#discussion_r629539919 . Here we have a very strange situation where `fs.walk` is dvcignore aware, but `isdir/isdfile` are not, which is bad.  So we need to either use dvcignore in both, or just apply dvcignore where it is needed on top of RepoFileSystem, effectively decoupling it and dvcignore.

efiop: Simple fix for now is to just use `dvicgnore` in `RepoFileSystem`s `isfile/isdir/exists`, but we might want to think about futher decoupling here in the future.

dvc/fs/repo.py
```diff
@@ -164,6 +167,9 @@ def exists(self, path_info) -> bool:
     def isdir(self, path):  # pylint: disable=arguments-differ
         fs, dvc_fs = self._get_fs_pair(path)
 
+        if dvc_fs and dvc_fs.repo.dvcignore.is_ignored_dir(path):
```
karajan1001, line 170: ignore_subrepos is need here.

efiop:
To summarize the stuff to consider and handle in the followups:

1) Potential RepoFileSystem and dvcignore decoupling

2) Scan for potential existing tests that were relying on not using dvcignore. Those tests are bad and are a result of bad previous architecture.

3) We should make State use fs instead of os.path and possibly stop passing dvcignore to it into constructor or into separate methods (by possibly wrapping the filesystem with dvcignore before passing the fs to the methods)

4) dvcignore can be used even more granularly in thngs like stages and outputs (and other). E.g. only when dvc-adding, so we don't waste dvcignore calls for no reason.

5) We should stop using abspath in dvcignore. This is related to fsspec migration and making our filesystems only work with fspath-s

6) Stop using PathInfo in fs/dvcignore related applications. Again related to fsspec migration and making our filesystems only work with strings fspaths-s

PR #5818, opened 2021-04-15
exp: use exp baseline for DVC commands that compare to HEAD during `exp run`

### LINKED ISSUE #5750: exp show: only shows one row while `exp run` is running.
# Bug Report

## Description

When running `exp show` between checkpoints of `exp run` (between checkpoints while the repo isn't locked), the table only shows a `workspace` and a single row (I assume the one currently being run)

In my case, this is the normal table before a run:
regular-exp-table

the table during a run:
mid-run-exp-table

and the table after the run is complete:
after-run-table

### Reproduce

1. Set up `dvc-checkpoints-mnist` (or the demo project of the vscode extension)
2. Run `dvc exp run`
3. While `dvc exp run` is running, try running `dvc exp show` until the lock isn't present and you get a table.

### Expected

The `exp show` table displays the full experiments table, or at least refuses to display the incorrect table.

### Environment information

**Output of `dvc doctor`:**

```console
DVC version: 2.0.10 (pip)
Platform: Python 3.7.3 on Linux-5.8.0-3-amd64-x86_64-with-MX-19.3-patito_feo
Supports: http, https, s3
Caches: local
Remotes: https
Repo: dvc (subdir), git
```

I'm aware this may be an unavoidable part of how experiments are implemented, and if so we could also use this Issue to discuss how to best work around it both as a user and as an external consumer of DVC CLI's output.

pmrowla:

Will close #5750 

[asciicast](https://asciinema.org/a/xTMncLndoFtOkwuSARH1GLfY9)

This PR addresses the issue in `exp show`, `exp diff`, `metrics diff`, `params diff`, `plots diff`, `dvc diff`

### DIFF
--- a/dvc/repo/diff.py
+++ b/dvc/repo/diff.py
@@ -7,6 +7,7 @@ from dvc.exceptions import PathMissingError
 from dvc.objects.stage import get_file_hash
 from dvc.objects.stage import stage as ostage
 from dvc.repo import locked
+from dvc.repo.experiments.utils import fix_exp_head
 
 logger = logging.getLogger(__name__)
 
@@ -28,7 +29,8 @@ def diff(self, a_rev="HEAD", b_rev=None, targets=None):
 
     repo_fs = RepoFileSystem(self)
 
-    b_rev = b_rev if b_rev else "workspace"
+    a_rev = fix_exp_head(self.scm, a_rev)
+    b_rev = fix_exp_head(self.scm, b_rev) if b_rev else "workspace"
     results = {}
     missing_targets = {}
     for rev in self.brancher(revs=[a_rev, b_rev]):
--- a/dvc/repo/experiments/diff.py
+++ b/dvc/repo/experiments/diff.py
@@ -1,5 +1,6 @@
 import logging
 
+from dvc.repo.experiments.utils import fix_exp_head
 from dvc.utils.diff import diff as _diff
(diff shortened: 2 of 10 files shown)


pmrowla:
Regarding the question raised by @efiop during the sprint meeting: I looked into it and I don't think we can avoid moving `HEAD` during experiment (checkpoint) runs. CLI git internals are written to assume index/commit/merge/etc behaviors are always done done against `HEAD`, and dulwich/libgit2 follow the same design.

Any users that are negatively impacted by this behavior (detaching and moving Git `HEAD` during `exp run`) and don't want us to touch their Git workspace should just use queued/temp directory runs instead of running `exp run` in their workspace.

dberenbaum (approved):
Looks awesome!

pared (commented):
LGTM to me so far, is it still WIP?

pmrowla:
I marked it WIP until I get to fixing the test issue, but for the most part this is done @pared

jorgeorpinel:
This is some crazy magic, nice 🧙

pmrowla:
It's basically the same behavior that git does when you are doing something like a bisect or interactive rebase. git moves HEAD to whatever the "current" commit is as needed, but it keeps track of the original repo head in another ref and uses that as needed (including restoring everything when the bisect/rebase is finished)

PR #5827, opened 2021-04-16
get rid of disutils

### LINKED ISSUE #5817: remove use of distutils
`distutils` [has been deprecated](https://www.python.org/dev/peps/pep-0632) in 3.10 and will be removed in 3.12. 

The only usage I could find was in the `external_repo`:

https://github.com/iterative/dvc/blob/e0842ce01e4d78223c3302a378a7c766eacee1db/dvc/external_repo.py#L147

`distuils.dir_util.copy_tree` is similar to the `shutil.copytree`, but it also works with an existing dir (`shutil.copytree` got `dirs_exist_ok` param since Python3.8).

Anyway, in the code, we are copying the directory to a temporary directory, so we could just use the `shutil.copytree` there instead.
https://github.com/iterative/dvc/blob/31439754377d829da6babffcc720dd696554d8ab/dvc/external_repo.py#L156-L159

bhagone:

Closes #5817

### DIFF
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -144,7 +144,7 @@ def _cached_clone(url, rev, for_write=False):
     revision checked out. If for_write is set prevents reusing this dir via
     cache.
     """
-    from distutils.dir_util import copy_tree
+    from shutil import copytree
 
     # even if we have already cloned this repo, we may need to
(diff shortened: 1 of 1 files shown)


skshetry (approved):
thanks @bhagone for contributions, looks good to me 🎉

PR #5829, opened 2021-04-16
exp run: error out when trying to resume from intermediate checkpoint w/no modifications

### LINKED ISSUE #5737: Git compatibility for exp names
```bash
$ dvc exp show --no-pager
┏━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━┳━━━━━━━━┳━━━━━━━━━━┳━━━┳━━━━━━━━┳━━━┓
┃ Experiment            ┃ Created  ┃ step ┃   loss ┃ accuracy ┃ … ┃ epochs ┃ … ┃
┡━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━╇━━━━━━━━╇━━━━━━━━━━╇━━━╇━━━━━━━━╇━━━┩
│ workspace             │ -        │    9 │ 1.8085 │  0.52762 │ … │ 15     │ … │
│ master                │ 09:27 AM │    - │     15 │  2.1e-07 │ … │ 0.124  │ … │
│ │ ╓ exp-dba5c         │ 10:42 AM │    9 │ 1.8085 │  0.52762 │ … │ 15     │ … │
│ │ ╟ 00e68eb           │ 10:42 AM │    8 │  1.857 │  0.48758 │ … │ 15     │ … │
│ │ ╟ cda7883 (3263dc2) │ 10:42 AM │    7 │ 1.9092 │  0.44263 │ … │ 15     │ … │
│ │ ╓ exp-59c7a         │ 10:41 AM │    6 │ 1.9522 │  0.42667 │ … │ 15     │ … │
│ │ ╟ d721197           │ 10:41 AM │    5 │ 2.0103 │  0.36575 │ … │ 15     │ … │
│ │ ╟ 0d4e1a7           │ 10:40 AM │    4 │ 2.0717 │  0.30868 │ … │ 15     │ … │
│ │ ╟ 312a0e5           │ 10:40 AM │    3 │ 2.1351 │   0.2546 │ … │ 15     │ … │
│ │ ╟ a00b24c           │ 10:40 AM │    2 │  2.202 │  0.20255 │ … │ 15     │ … │
│ │ ╟ a371c0e           │ 10:39 AM │    1 │ 2.2746 │  0.15662 │ … │ 15     │ … │
│ ├─╨ 3da11aa           │ 10:39 AM │    0 │ 2.3516 │  0.12398 │ … │ 15     │ … │
└───────────────────────┴──────────┴──────┴────────┴──────────┴───┴────────┴───┘
```

```bash
# works just great!
$ git diff 00e68eb 0d4e1a7 -- params.yaml
diff --git a/params.yaml b/params.yaml
index 8f6f5fb..fedaf92 100644
...
...

# These two do not work :( 
$ git diff exp-dba5c 0d4e1a7 -- params.yaml 
fatal: bad revision 'exp-dba5c'
$ git diff dba5c 0d4e1a7 -- params.yaml
fatal: bad revision 'dba5c'
```

I'd love to have Git compatibility for exp names like `exp-dba5c` or at least the suffixes `dba5c`.

EDIT:
The same should ideally work for `dvc metrics` & `dvc params`.

pmrowla:

Will close #5737, see original issue for full discussion of changes

After this PR

- Checkpoints can only be resumed (with no modifications) from the head of a checkpoint run. (New checkpoint branches with modifications can still be created from any intermediate or head checkpoint)

See example:
1. apply ambiguous commit and run with no modifications (error with no modifications)
2. apply ambiguous commit that is an exact match for a head and run with no modifications (resume from the exact matching head)
3. apply non-ambiguous commit and run with no modifications (error with no modifications)
[asciicast](https://asciinema.org/a/BsdI6cf78qTYhlNTKsoApSm52)

### DIFF
--- a/dvc/repo/experiments/__init__.py
+++ b/dvc/repo/experiments/__init__.py
@@ -195,13 +195,23 @@ class Experiments:
                             branch_name = ExpRefInfo.from_ref(branch).name
                         else:
                             branch_name = f"{resume_rev[:7]}"
-                        if self.scm.is_dirty() or not branch:
+                        if self.scm.is_dirty():
                             logger.info(
                                 "Modified checkpoint experiment based on "
                                 "'%s' will be created",
                                 branch_name,
                             )
                             branch = None
+                        elif (
+                            not branch
+                            or self.scm.get_ref(branch) != resume_rev
+                        ):
+                            raise DvcException(
+                                f"Nothing to do for unchanged checkpoint "
+                                f"'{resume_rev[:7]}'. To resume from the head "
+                                "of this experiment, use "
+                                f"'dvc exp apply {branch_name}'."
+                            )
                         else:
                             logger.info(
                                 "Existing checkpoint experiment '%s' will be "
@@ -237,11 +247,15 @@ class Experiments:
                         baseline_rev[:7],
                     )
                 finally:
-                    # Reset any of our changes before prior unstashing
                     if resume_rev:
+                        # NOTE: this set_ref + reset() is equivalent to
+                        # `git reset orig_head` (our SCM reset() only operates
+                        # on HEAD rather than any arbitrary commit)
                         self.scm.set_ref(
                             "HEAD", orig_head, message="dvc: restore HEAD"
                         )
+                        self.scm.reset()
+                    # Revert any of our changes before prior unstashing
                     self.scm.reset(hard=True)
 
         return stash_rev
@@ -719,10 +733,10 @@ class Experiments:
 
         # NOTE: the stash commit to be popped already contains all the current
         # workspace changes plus CLI modifed --params changes.
-        # `reset --hard` here will not lose any data (pop without reset would
-        # result in conflict between workspace params and stashed CLI params).
-        self.scm.reset(hard=True)
-        with self.scm.detach_head(entry.rev):
+        # `checkout --force` here will not lose any data (popping stash commit
+        # will result in conflict between workspace params and stashed CLI
+        # params, but we always want the stashed version).
+        with self.scm.detach_head(entry.rev, force=True):
             rev = self.stash.pop()
             self.scm.set_ref(EXEC_BASELINE, entry.baseline_rev)
             if entry.branch:
--- a/dvc/repo/experiments/__init__.py
+++ b/dvc/repo/experiments/__init__.py
@@ -206,12 +206,35 @@ class Experiments:
                             not branch
                             or self.scm.get_ref(branch) != resume_rev
                         ):
-                            raise DvcException(
-                                f"Nothing to do for unchanged checkpoint "
-                                f"'{resume_rev[:7]}'. To resume from the head "
-                                "of this experiment, use "
-                                f"'dvc exp apply {branch_name}'."
-                            )
+                            msg = [
+                                (
+                                    "Nothing to do for unchanged checkpoint "
+                                    f"'{resume_rev[:7]}'. "
+                                )
+                            ]
+                            if branch:
+                                msg.append(
+                                    "To resume from the head of this "
+                                    "experiment, use "
+                                    f"'dvc exp apply {branch_name}'."
+                                )
+                            else:
+                                names = [
+                                    ref_info.name
+                                    for ref_info in exp_refs_by_rev(
+                                        self.scm, resume_rev
+                                    )
+                                ]
+                                if len(names) > 3:
+                                    names[3:] = [
+                                        f"... ({len(names) - 3} more)"
+                                    ]
+                                msg.append(
+                                    "To resume an experiment containing this "
+                                    "checkpoint, apply one of these heads:\n"
+                                    "\t{}".format(", ".join(names))
+                                )
+                            raise DvcException("".join(msg))
                         else:
                             logger.info(
                                 "Existing checkpoint experiment '%s' will be "
--- a/dvc/repo/experiments/__init__.py
+++ b/dvc/repo/experiments/__init__.py
@@ -194,8 +194,8 @@ class Experiments:
                         if branch:
                             branch_name = ExpRefInfo.from_ref(branch).name
                         else:
-                            branch_name = ""
-                        if self.scm.is_dirty():
+                            branch_name = f"{resume_rev[:7]}"
+                        if self.scm.is_dirty() or not branch:
                             logger.info(
                                 "Modified checkpoint experiment based on "
                                 "'%s' will be created",
@@ -447,22 +447,34 @@ class Experiments:
         """
         assert resume_rev
 
-        allow_multiple = "params" in kwargs
-        branch: Optional[str] = self.get_branch_by_rev(
-            resume_rev, allow_multiple=allow_multiple
-        )
-        if not branch:
-            raise DvcException(
-                "Could not find checkpoint experiment " f"'{resume_rev[:7]}'"
+        branch: Optional[str] = None
+        try:
+            allow_multiple = bool(kwargs.get("params", None))
+            branch = self.get_branch_by_rev(
+                resume_rev, allow_multiple=allow_multiple
             )
+            if not branch:
+                raise DvcException(
+                    "Could not find checkpoint experiment "
+                    f"'{resume_rev[:7]}'"
+                )
+            baseline_rev = self._get_baseline(branch)
+        except MultipleBranchError as exc:
+            baselines = {
+                info.baseline_sha
+                for info in exc.ref_infos
+                if info.baseline_sha
+            }
+            if len(baselines) == 1:
+                baseline_rev = baselines.pop()
+            else:
+                raise
 
-        baseline_rev = self._get_baseline(branch)
         logger.debug(
-            "Resume from checkpoint '%s' with baseline '%s'",
-            resume_rev,
+            "Checkpoint run from '%s' with baseline '%s'",
+            resume_rev[:7],
             baseline_rev,
         )
-
         return self._stash_exp(
             *args,
             resume_rev=resume_rev,
@@ -797,7 +809,10 @@ class Experiments:
         if not ref_infos:
             return None
         if len(ref_infos) > 1 and not allow_multiple:
-            raise MultipleBranchError(rev)
+            for ref_info in ref_infos:
+                if self.scm.get_ref(str(ref_info)) == rev:
+                    return str(ref_info)
+            raise MultipleBranchError(rev, ref_infos)
         return str(ref_infos[0])
 
     def get_exact_name(self, rev: str):
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -172,7 +172,10 @@ def _collect_rows(
                 else:
                     tree = "└──"
                 new_checkpoint = True
-            name = exp.get("name", rev[:7])
+            if "name" in exp:
+                name = f"{rev[:7]} [[bold]{exp['name']}[/]]"
+            else:
+                name = rev[:7]
 
         row = [
             name,
--- a/dvc/repo/experiments/apply.py
+++ b/dvc/repo/experiments/apply.py
@@ -32,7 +32,9 @@ def apply(repo, rev, force=True, **kwargs):
         raise InvalidExpRevError(rev) from exc
 
     stash_rev = exp_rev in exps.stash_revs
-    if not stash_rev and not exps.get_branch_by_rev(exp_rev):
+    if not stash_rev and not exps.get_branch_by_rev(
+        exp_rev, allow_multiple=True
+    ):
         raise InvalidExpRevError(exp_rev)
 
     # Note that we don't use stash_workspace() here since we need finer control
--- a/dvc/repo/experiments/base.py
+++ b/dvc/repo/experiments/base.py
@@ -74,12 +74,13 @@ class InvalidExpRevError(InvalidArgumentError):
 
 
 class MultipleBranchError(DvcException):
-    def __init__(self, rev):
+    def __init__(self, rev, ref_infos):
         super().__init__(
             f"Ambiguous commit '{rev[:7]}' belongs to multiple experiment "
             "branches."
         )
         self.rev = rev
+        self.ref_infos = ref_infos
 
 
 class ApplyConflictError(InvalidArgumentError):
--- a/dvc/scm/git/__init__.py
+++ b/dvc/scm/git/__init__.py
@@ -426,7 +426,7 @@ class Git(Base):
             commit = self.resolve_commit(parent)
 
     @contextmanager
-    def detach_head(self, rev: Optional[str] = None):
+    def detach_head(self, rev: Optional[str] = None, force: bool = False):
         """Context manager for performing detached HEAD SCM operations.
 
         Detaches and restores HEAD similar to interactive git rebase.
@@ -440,7 +440,7 @@ class Git(Base):
             rev = "HEAD"
         orig_head = self.get_ref("HEAD", follow=False)
         logger.debug("Detaching HEAD at '%s'", rev)
-        self.checkout(rev, detach=True)
+        self.checkout(rev, detach=True, force=force)
         try:
             yield self.get_ref("HEAD")
         finally:
--- a/tests/func/experiments/test_checkpoints.py
+++ b/tests/func/experiments/test_checkpoints.py
@@ -160,3 +160,27 @@ def test_resume_branch(tmp_dir, scm, dvc, checkpoint_stage, workspace):
     assert branch_rev == dvc.experiments.scm.gitpython.repo.git.merge_base(
         checkpoint_a, checkpoint_b
     )
+
+
+@pytest.mark.parametrize("workspace", [True, False])
+def test_resume_non_head_checkpoint(
+    tmp_dir, scm, dvc, checkpoint_stage, workspace
+):
+    orig_head = scm.get_rev()
+    results = dvc.experiments.run(
+        checkpoint_stage.addressing, params=["foo=2"], tmp_dir=not workspace
+    )
+    checkpoint_head = first(results)
+    orig_branch = dvc.experiments.get_branch_by_rev(checkpoint_head)
+
+    rev = list(scm.branch_revs(checkpoint_head, orig_head))[-1]
+    dvc.experiments.apply(rev)
+
+    with pytest.raises(DvcException):
+        dvc.experiments.run(checkpoint_stage.addressing, tmp_dir=not workspace)
+
+    results = dvc.experiments.run(
+        checkpoint_stage.addressing, params=["foo=100"], tmp_dir=not workspace
+    )
+    new_head = first(results)
+    assert orig_branch != dvc.experiments.get_branch_by_rev(new_head)
--- a/tests/func/experiments/test_show.py
+++ b/tests/func/experiments/test_show.py
@@ -111,6 +111,7 @@ def test_show_checkpoint(
     for i, rev in enumerate(checkpoints):
         if i == 0:
             name = dvc.experiments.get_exact_name(rev)
+            name = f"{rev[:7]} [{name}]"
             fs = "╓"
         elif i == len(checkpoints) - 1:
             name = rev[:7]
@@ -155,7 +156,7 @@ def test_show_checkpoint_branch(
     for rev in (checkpoint_a, checkpoint_b):
         ref = dvc.experiments.get_branch_by_rev(rev)
         ref_info = ExpRefInfo.from_ref(ref)
-        name = ref_info.name
+        name = f"{rev[:7]} [{ref_info.name}]"
         assert f"╓ {name}" in cap.out
     assert f"({branch_rev[:7]})" in cap.out
 


pmrowla:
Need to clean up test coverage, but want to confirm that this is the intended behavior @dberenbaum @dmpetrov

dberenbaum:
You decided to sneak in an updated table UI here 🤣 ? Should that be in a separate PR?

One thing to note is that if I `exp apply` on a non-head checkpoint and then `exp run` with no modifications, I will get duplicated checkpoints:
```console
$ dvc exp show --no-pager
┏━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━┳━━━━━━┳━━━━━━━━━━━━━━━━━━━┳━━━┓
┃ Experiment              ┃ Created      ┃ epoch ┃ mult ┃ params.yaml:start ┃ … ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━╇━━━━━━╇━━━━━━━━━━━━━━━━━━━╇━━━┩
│ workspace               │ -            │     6 │    6 │ 0                 │ 3 │
│ main                    │ Mar 10, 2021 │     - │    - │ 0                 │ 2 │
│ │ ╓ b885f8e [exp-54d76] │ 01:44 PM     │     6 │    6 │ 0                 │ 3 │
│ │ ╟ 9b63a8c             │ 01:44 PM     │     5 │    5 │ 0                 │ 3 │
│ │ ╟ 120bbdc (38a98e7)   │ 01:44 PM     │     4 │    4 │ 0                 │ 3 │
│ │ ╓ 38a98e7 [exp-e44da] │ 01:44 PM     │     3 │    3 │ 0                 │ 2 │
│ │ ╟ 6795f10             │ 01:44 PM     │     2 │    2 │ 0                 │ 2 │
│ │ ╟ ca05d38             │ 01:44 PM     │     1 │    1 │ 0                 │ 2 │
│ ├─╨ 20ba56f             │ 01:44 PM     │     0 │    0 │ 0                 │ 2 │
└─────────────────────────┴──────────────┴───────┴──────┴───────────────────┴───┘
$ dvc exp apply ca05d38
Changes for experiment 'ca05d38' have been applied to your current workspace.
$ dvc exp run
Modified checkpoint experiment based on 'ca05d38' will be created
'bar.dvc' didn't change, skipping
Running stage 'foo':
> python checkpoint.py
Updating lock file 'dvc.lock'
Checkpoint experiment iteration '2c9da59'.
Updating lock file 'dvc.lock'
Checkpoint experiment iteration 'e4d6c38'.

        git add params.yaml dvc.yaml scores.yaml checkpoint.py dvc.lock

Reproduced experiment(s): exp-bc2fa

        dvc exp branch <exp>
$ dvc exp show --no-pager
┏━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━┳━━━━━━┳━━━━━━━━━━━━━━━━━━━┳━━━┓
┃ Experiment              ┃ Created      ┃ epoch ┃ mult ┃ params.yaml:start ┃ … ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━╇━━━━━━╇━━━━━━━━━━━━━━━━━━━╇━━━┩
│ workspace               │ -            │     3 │    3 │ 0                 │ 2 │
│ no_checkpoint           │ Mar 10, 2021 │     - │    - │ 0                 │ 2 │
│ │ ╓ e4d6c38 [exp-bc2fa] │ 01:45 PM     │     3 │    3 │ 0                 │ 2 │
│ │ ╟ 2c9da59 (ca05d38)   │ 01:45 PM     │     2 │    2 │ 0                 │ 2 │
│ │ ╓ b885f8e [exp-54d76] │ 01:44 PM     │     6 │    6 │ 0                 │ 3 │
│ │ ╟ 9b63a8c             │ 01:44 PM     │     5 │    5 │ 0                 │ 3 │
│ │ ╟ 120bbdc (38a98e7)   │ 01:44 PM     │     4 │    4 │ 0                 │ 3 │
│ │ ╓ 38a98e7 [exp-e44da] │ 01:44 PM     │     3 │    3 │ 0                 │ 2 │
│ │ ╟ 6795f10             │ 01:44 PM     │     2 │    2 │ 0                 │ 2 │
│ │ ╟ ca05d38             │ 01:44 PM     │     1 │    1 │ 0                 │ 2 │
│ ├─╨ 20ba56f             │ 01:44 PM     │     0 │    0 │ 0                 │ 2 │
└─────────────────────────┴──────────────┴───────┴──────┴───────────────────┴───┘
```
I'm not sure if this should be okay, or if dvc should maybe throw an error here. I can't think of a reason why someone would do this on purpose, so I think it's fine as is, but wanted to call it out in case.

pmrowla:
> You decided to sneak in an updated table UI here 🤣 ? Should that be in a separate PR?

fwiw, the bold text for "regular" commits was already in DVC a while back but it had been temporarily disabled due to some other UI bug. We just forgot to turn it back on at some point. I only noticed it because I wanted to the `[exp-name]` in brackets stand out more relative to the `(abc123)` parent sha's.

> One thing to note is that if I `exp apply` on a non-head checkpoint and then `exp run` with no modifications, I will get duplicated checkpoints:
> 
> I'm not sure if this should be okay, or if dvc should maybe throw an error here. I can't think of a reason why someone would do this on purpose, so I think it's fine as is, but wanted to call it out in case.

This is basically the entire point of the original issue, and it will only happen if the non-head checkpoint you apply belongs to more than one possible experiment (since we cannot easily decide which one you meant to continue). In your example, `ca05d38` belongs to both `exp-54d76` and `exp-e44da`. Our only options here are:

- error out entirely (the original behavior and what we are trying to fix in this PR)
- just create a new branch, which will end up duplicating checkpoints (this PR)
- do an expensive computation and walk potentially huge git history trees to figure out which git branch should be continued (what we decided to avoid in the discussion before this PR)

After this PR, if you `exp apply` a non-head checkpoint that only belongs to a single branch (like `9b63a8c` from your example), and then run without modifications, we will resume that single branch, and you would not get duplicate checkpoints.

dberenbaum:
>     * error out entirely (the original behavior and what we are trying to fix in this PR)

I didn't think this was what we were fixing, so maybe I'm missing some scenario. I thought this PR was about reverting to commits that belong to multiple experiments. That can be useful to 1) keep some previous commit as the final experiment result; or 2) make some modification and continue running.

I don't see why anyone would want to revert to an old commit and regenerate existing checkpoint commits **if there are no modifications**. If someone does this, I think it's fine to throw an error. The behavior in this PR might also be fine (it leaves this option open in case there is some use case for it), but I wanted to make it clear.

pmrowla:
> I don't see why anyone would want to revert to an old commit and regenerate **existing** checkpoint commits **if there are no modifications**. If someone does this, I think it's fine to throw an error. The behavior in this PR might also be fine (it leaves this option open in case there is some use case for it), but I wanted to make it clear.

It seems like this still needs some discussion then? We can make it error out on `exp run` if there are no modifications, but my impression was that we wanted to avoid erroring here at all (since it is not obvious to the user why that would be an invalid workspace/experiment state)

@dberenbaum @dmpetrov 

This comment seemed to imply that there was a use case for just being able to do `exp run` from any applied checkpoint: https://github.com/iterative/dvc/issues/5737#issuecomment-820484391

> good point! In some cases, you do apply xxxx & run but it didn't work and you would like to continue the training in "old" checkpoint. However, the next training should start from head not from xxxx.

dberenbaum:
> good point! In some cases, you do apply xxxx & run but it didn't work and you would like to continue the training in "old" checkpoint. However, the next training should start from head not from xxxx.

Yes, I think we need @dmpetrov to clarify here.

I thought this meant something like:

```console
$ dvc exp show
┏━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━┳━━━━━━━━┓
┃ Experiment              ┃ Created  ┃ step ┃    acc ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━╇━━━━━━━━┩
│ workspace               │ -        │    1 │ 0.9133 │
│ live                    │ 09:19 AM │    - │      - │
│ │ ╓ f2450ba [exp-498bc] │ 11:06 AM │    1 │ 0.9133 │
│ │ ╟ 7b4421a (06b31c7)   │ 11:06 AM │    1 │ 0.9133 │
│ │ ╓ 06b31c7 [exp-794b3] │ 11:05 AM │    1 │ 0.9133 │
│ ├─╨ 7b35cb2             │ 11:05 AM │    0 │ 0.8523 │
└─────────────────────────┴──────────┴──────┴────────┘
```

The modifications in exp-498bc stalled my training progress, so I want to go back to exp-794b3 and keep going:

```
$ dvc exp apply 06b31c7
$ dvc exp run
```

This should ideally add more checkpoints to exp-794b3.

[Using this PR, I'm actually seeing that it adds to exp-498bc, so probably need to QA more once we decide expected behavior.]

> my impression was that we wanted to avoid erroring here at all (since it is not obvious to the user why that would be an invalid workspace/experiment state)

What if we provide a hint, like `A checkpoint based on 06b31c7 already exists. Make some modification to your experiment and try dvc exp run again, or use --force to create another unmodified checkpoint based on this commit.`

dberenbaum:
> Right. I'd prefer to "jump" to the head (with a hint/warning message I guess) and continue training from the head.
> 
> In the example above, a user can `apply 7b35cb2` but `exp run` will warn the user it is "jumping" to the head and start training from 06b31c7 which is head of exp-794b3.

Got it, thanks!

I'd prefer to give a warning and make the user jump to the head rather than assume and keep going. If the user does `apply 7b36cb2` and then `exp run`, they seemingly have made a mistake, but it could be:
1. They meant to keep training from the experiment head (06b31c7).
2. They meant to make some modifications on top of 7b36cb2 before doing `exp run`.
3. They think there was some issue with 06b31c7 (maybe there was some environment variable not tracked by dvc that was set incorrectly), and they want to force this checkpoint to be regenerated.

What do you think about summarizing this in a warning and making the user decide how to handle it (for example, user would have to run `dvc exp apply exp-794b3/06b31c7`)?

I'm also not sure if the jump ahead functionality would require walking through the git history @pmrowla.

dmpetrov:
> What do you think about summarizing this in a warning and making the user decide how to handle it

Are you talking about exit with an error message before a user can make a decision (run next command)? Starting training with a warning message might be a better option.

A summary in warning seems like a good experience for the first time. However, it is going to be annoying for experienced users. So, I'd prefer a short clear warning like `WARN: checkpoint '7b35cb2' is done - skipping to the head of 'exp-794b3'`

pmrowla:
> I'm also not sure if the jump ahead functionality would require walking through the git history @pmrowla.

jumping ahead automatically depends on how many branches the `exp apply`ed commit belongs to. If it only belongs to one experiment, jumping ahead automatically is straightforward. If it belongs to more than one branch, it requires walking git history to figure out which branch head to jump to.

dberenbaum:
> Are you talking about exit with an error message before a user can make a decision (run next command)? Starting training with a warning message might be a better option.

Correct.

Why do you think that users generally fall under category 1 above and not 2 or 3? I don't see that 1 is the obvious default. Why would an experienced user revert to 7b36cb2 if they mean to keep training from the head of exp-794b3?

Also, from the comment of @pmrowla above, it seems like this jump ahead functionality is more difficult to implement, so is it worth it?

pmrowla:
@dberenbaum @dmpetrov any resolution on the desired behavior here?

dmpetrov:
> Also, from the comment of @pmrowla above, it seems like this jump ahead functionality is more difficult to implement, so is it worth it?

I don't think it is an important decision, to be honest. We are talking about not a common scenario - applying a checkpoint, deciding not to cheng it, and continue the training. If the jump to head makes implementation more complicated I'd recommend failing with an error message.

dberenbaum:
Thanks, @dmpetrov and @pmrowla! @pmrowla let me know if that's enough info or if you want help on error messages or anything.

pmrowla:
@dberenbaum I've updated the asciinema rec w/the new behavior (error out when trying to resume w/no changes). Not sure how the error message should be worded, let me know what you think

iesahin:
> * They meant to keep training from the experiment head (06b31c7).
> 
> * They meant to make some modifications on top of 7b36cb2 before doing `exp run`.
> 
> * They think there was some issue with 06b31c7 (maybe there was some environment variable not tracked by dvc that was set incorrectly), and they want to force this checkpoint to be regenerated.

I think it may be better to let the user select one of these with a parameter, like `dvc exp run --start-from HEAD` or `dvc exp run --start-from exp-32`. It can be basically `dvc exp apply && dvc exp run` without committing. Users may have all sorts of reasons to start from an intermediate experiment, or revert back to `HEAD` or continue from the last one.

dmpetrov:
> Not sure if this is worth worrying about.

@dberenbaum that's an interesting finding. Could you please help me to understand this?

1. Was the 2nd experiment `exp-e44da` (commit `e8eb2b5`) a modification of the existing one `exp-e44da`? 
   - If yes then it looks like a bug - you cannot have all the checkpoints in the same line (from `763700d` to `90d4e73`).
   - if not - why `exp-e44da` & `exp-82e69` not in the same line initially?
2. Just to make sure - you have not done any modification after `dvc exp apply exp-e44da`?

dberenbaum:
> 1. Was the 2nd experiment `exp-e44da` (commit `e8eb2b5`) a modification of the existing one `exp-e44da`?

Yes, you got it.

> 2. Just to make sure - you have not done any modification after `dvc exp apply exp-e44da`?

Correct again.

dmpetrov:
> Yes, you got it.

thank you, @dberenbaum 

There we go:
> * If yes then it looks like a bug - you cannot have all the checkpoints in the same line (from `763700d` to `90d4e73`).

@pmrowla could you please check that.

pmrowla:
@dberenbaum @dmpetrov this is not a bug, it is working correctly. The issue w/the display is that git returns us branch (exp) names in any arbitrary order, and we display them in whatever order git gives them too us.

In the first `exp show`, you have 2 experiments containing the following commits

`exp-e44da`
- `763700d` (epoch 0)
- `5b86517` (epoch 1) (this is the branch/fork point)

`exp-82e69`
- `763700d` (epoch 0)
- `5b86517` (epoch 1) (this is the branch/fork point)
- `e8eb2b5` (epoch 3)
- `90d4e73` (epoch 5)

In the second `exp show`, you have 2 experiments which now contain the following commits

`exp-e44da`
- `763700d` (epoch 0)
- `5b86517` (epoch 1) (this is the branch/fork point)
- `f59c3ee` (epoch 2)
- `f49318b` (epoch 3)

`exp-82e69`
- `763700d` (epoch 0)
- `5b86517` (epoch 1) (this is the branch/fork point)
- `e8eb2b5` (epoch 3)
- `90d4e73` (epoch 5)

You can see that `exp-e44da` has (correctly) been resumed and extended by 2 new checkpoints. `exp-82e69` has not been modified at all.

Your git history looks like this:

tree

Note that git has no concept of which branch (green or blue) is an "original" and which is a "branch w/modification". In git internals, both branches have equal priority, all git cares about is that commits `f59ce33` and `e8eb2b5` share a parent (`5b86517`). When we ask git for a list of branches which are derived from our current HEAD to be displayed in the table, git can return them to us in any order, since git sees the branches as equivalent.

The confusion here is that in DVC we tend to think about experiments like this:

dvc tree

The tree on the left is what is "expected", where the green experiment is the "original", and the blue is the "branch". However, to git, the tree on the right is also equivalent and just as valid.

So there are two possible ways which this git history could be displayed in a flat table.

DVC's output in @dberenbaum's example is the tree on the right, with the blue side shown as a contiguous chain:
```
│ │ ╓ f49318b [exp-e44da]
│ │ ╟ f59c3ee (5b86517)
│ │ ╓ 90d4e73 [exp-82e69]
│ │ ╟ e8eb2b5
│ │ ╟ 5b86517
│ ├─╨ 763700d  
```

What you were probably "expecting" was the tree on the left, like this:
```
│ │ ╓ 90d4e73 [exp-82e69]
│ │ ╟ e8eb2b5  (5b86517)
│ │ ╓ f49318b [exp-e44da]
│ │ ╟ f59c3ee (5b86517)
│ │ ╟ 5b86517
│ ├─╨ 763700d  
```

These two tables are equivalent in terms of displaying a git tree in a flat structure. The way they are displayed in `exp show` depends on whatever arbitrary order git returns the branches to us, and the order is not guaranteed.

dmpetrov:
@pmrowla thank you for the detailed explanation. love the images 😍 

Yeah. The behavior seems correct with the set of assumptions that we had. The question is - how to improve it without the overhead (before we discussed that we can do that by identifying modifications that introduce a huge computational overhead).

Today's problems:
- **Problem 1**: We didn't recognize the modification `5b-e8`. `e8` looks like a regular checkpoint.
- **Problem 2**: Recognizing not-modifications by timestamps (as we discussed before) works only for "intermediate" checkpoints - if Dave would jump to 76 (not the head - 5b) then Problem-1 would not have happened.

It looks like we are talking about a special case when a user modifies the head of an exp, trains it, and then returns back to continue training the head. In all the other cases, the current approach works just fine.
@pmrowla & @dberenbaum could you please confirm that?

Also, @pmrowla could you please confirm that we recognize modifications by timestepms (in case of multiple parents - the older child is a regular checkpoint and the rest are modifications). If not, what algorithm we use - the one that the Git tool provides (no control over this)?

dberenbaum:
>     * **Problem 1**: We didn't recognize the modification `5b-e8`. `e8` looks like a regular checkpoint.

`e8eb2b5` was identified as a modification since a new experiment was created for it, but by the time of the second `exp show` it's unclear which is the "original" and which is the "modified" experiment.

>     * **Problem 2**: Recognizing not-modifications by timestamps (as we discussed before) works only for "intermediate" checkpoints - if Dave would jump to 76 (not the head - 5b) then Problem-1 would not have happened.

If I jump to `763700d`, I have to make a modification before a new checkpoint would be created off of it (since the next unmodified checkpoint already exists, see the asciinema video at the top of the discussion). In that case, it shouldn't be an issue since the table will look like:

```
┏━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━┳━━━━━━┳━━━━━━━━━━━━━━━━━━━┳━━━┓
┃ Experiment              ┃ Created      ┃ epoch ┃ mult ┃ params.yaml:start ┃ … ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━╇━━━━━━╇━━━━━━━━━━━━━━━━━━━╇━━━┩
│ workspace               │ -            │     6 │    6 │ 0                 │ 2 │
│ main                    │ Mar 10, 2021 │     - │    - │ 0                 │ 2 │
│ │ ╓ 2c8bf2f [exp-f2ff9] │ 01:07 PM     │     6 │    6 │ 0                 │ 2 │
│ │ ╟ 4852c3b (763700d)   │ 01:07 PM     │     3 │    3 │ 0                 │ 2 │
│ │ ╓ 90d4e73 [exp-82e69] │ 12:05 PM     │     5 │    5 │ 0                 │ 2 │
│ │ ╟ e8eb2b5 (5b86517)   │ 12:05 PM     │     3 │    3 │ 0                 │ 2 │
│ │ ╓ 5b86517 [exp-e44da] │ 12:05 PM     │     1 │    1 │ 0                 │ 2 │
│ ├─╨ 763700d             │ 12:05 PM     │     0 │    0 │ 0                 │ 2 │
└─────────────────────────┴──────────────┴───────┴──────┴───────────────────┴───┘
```

> It looks like we are talking about a special case when a user modifies the head of an exp, trains it, and then returns back to continue training the head. In all the other cases, the current approach works just fine.
> @pmrowla & @dberenbaum could you please confirm that?

As far as I know, that's correct.  I agree that it seems like a special case.

pmrowla:
> Also, @pmrowla could you please confirm that we recognize modifications by timestepms (in case of multiple parents - the older child is a regular checkpoint and the rest are modifications). If not, what algorithm we use - the one that the Git tool provides (no control over this)?

We don't do anything at all to actually "recognize modifications". We just get the list of experiment refs (exp branch tips) for a given HEAD from git - this list will be ordered by timestamp. The only other sorted/non-random exp order we can get from git would be alphabetical by exp name (which doesn't really make sense for us here).

We then start with the oldest branch tip, and essentially fill out the table data from the bottom up. So the branch with the oldest tip will always appear as a single contiguous run at the bottom of the table. Then we process the next oldest branch, and so on.

pmrowla:
After discussion with @dmpetrov we decided this latest behavior should be acceptable. The edge case where
> user modifies the head of an exp, trains it, and then returns back to continue training the head
looks a bit strange, but there is not an obvious solution (that would not come with a big performance hit) right now.

We can revisit this later as needed if/when this issue is brought up by enough users.

dvc/repo/experiments/__init__.py
```diff
@@ -194,14 +194,47 @@ def _stash_exp(
...
                                 branch_name,
                             )
                             branch = None
+                        elif (
+                            not branch
+                            or self.scm.get_ref(branch) != resume_rev
+                        ):
+                            msg = [
+                                (
+                                    "Nothing to do for unchanged checkpoint "
```
jorgeorpinel, line 211: "unchanged" here may be confusing since the last checkpoint can be continued from even if unchanged. I guess the point is that it's intermediate? Easiest solution: just remove "unchanged" in this line.

dvc/repo/experiments/__init__.py
```diff
@@ -194,14 +194,47 @@ def _stash_exp(
...
+                                    f"'{resume_rev[:7]}'. "
+                                )
+                            ]
+                            if branch:
+                                msg.append(
+                                    "To resume from the head of this "
+                                    "experiment, use "
```
jorgeorpinel, line 218: Is it obvious what "head" means? Sounds like a Git term. I'd be more explicit:

`"To resume this experiment as-is, apply the latest checkpoint: "`

jorgeorpinel: But there are more options right? So maybe

```
"Make changes to branch off from here (1234567) into a new experiment. "
"To continue this experiment as-is, apply the latest checkpoint with: ..."
```

dvc/repo/experiments/__init__.py
```diff
@@ -194,14 +194,47 @@ def _stash_exp(
...
+                                        self.scm, resume_rev
+                                    )
+                                ]
+                                if len(names) > 3:
+                                    names[3:] = [
+                                        f"... ({len(names) - 3} more)"
+                                    ]
+                                msg.append(
+                                    "To resume an experiment containing this "
+                                    "checkpoint, apply one of these heads:\n"
```
jorgeorpinel, line 234: Similar suggestion:
```
"Make changes to branch off from here (1234567) into a new experiment. "
"Otherwise, apply a n experiment containing this commit first:"
```

dvc/repo/experiments/__init__.py
```diff
@@ -447,22 +484,34 @@ def _resume_checkpoint(
...
-                "Could not find checkpoint experiment " f"'{resume_rev[:7]}'"
+        branch: Optional[str] = None
+        try:
+            allow_multiple = bool(kwargs.get("params", None))
+            branch = self.get_branch_by_rev(
+                resume_rev, allow_multiple=allow_multiple
             )
+            if not branch:
+                raise DvcException(
+                    "Could not find checkpoint experiment "
```
jorgeorpinel, line 495:
```suggestion
                    "Could not find checkpoint "
```

jorgeorpinel (commented):
UI review

jorgeorpinel:
> Checkpoints can only be resumed (with no modifications) from the head of a checkpoint run

https://dvc.org/doc/command-reference/exp/run#checkpoints could use a note about this.

And probably several `exp show` examples shuold be updated in general, esp. in https://dvc.org/doc/command-reference/exp/show

@pmrowla mind opening a PR or issue in docs? Thanks

jorgeorpinel:
> And probably several exp show examples shuold be updated in general

This part actually may have more to do with #5885 ... I'll comment over there.

Issue #5667, opened 2021-03-22
exp show: fails when dvc.yaml is invalid [QA]


jorgeorpinel:
If after running experiments you edit dvc.yaml and (accidentally) make it invalid, `exp show` will

```console
ERROR: failed to show experiments - unable to read: 'dvc.yaml', YAML file structure is corrupted: while scanning a simple key
```

(or some other dvc.yaml-related error). Why does `exp show` check dvc.yaml? Does it also check the consistency of all project dvc.yaml files like `status/repro/exp run`? (May hurt performance)

pmrowla:
Collecting metrics and params for a repo requires collecting stages from `dvc.yaml` (since we need the list of metrics outs and params dependencies for all stages in the pipeline).

I'm pretty sure this same issue isn't experiments specific and would apply to any DVC command that requires collecting stages/metrics/params.

shcheklein:
(Note: I didn't check the current implementation, so feel free to disregard if it's already done this waY)

My 2cs on this. It reminds me the problem that we had in the viewer. What if one of the commits is broken from the DVC perspective (raises something similar to the above, for example)? Should we fail? Should we show at least other commits? 

In the viewer we've decided that we parse commits (and we don't deal with workspace there) in isolation.  Each commit is independent and if it breaks we move to the next one. Each commit that viewer returns in its version of the "dvc exp show" has a status - ok/failed + error message. Otherwise it was almost impossible to deal with actual repos - people break things constantly. 

Also, let's keep in mind the VS code extension - it'll depend on `dvc exp show` and it will keep running it as we change `dvc.yaml`. It might happen that we'll run it when workspace is broken - we need to handle this to my mind, we still should show the table, we can assign some special flag/state to the broken commits/experiments.

dberenbaum:
Related convo in #5525 about showing as much as possible in plots/metrics/params/exp show, although I think a broken `dvc.yaml` is a bit different and tougher problem than the other examples discussed in either issue. Maybe better to leave this convo in #5525 and break out smaller issues as you QA @jorgeorpinel to keep each issue focused? It's our constant struggle to try to keep issues focused on a single topic 😄 .

jorgeorpinel:
> Collecting metrics and params for a repo requires collecting stages from dvc.yaml

Makes sense. But a main use of `exp show` is to see past experiment details, and past exps are guaranteed to have had valid dvc.yaml files, right?

[...]

Issue #5525, opened 2021-02-26
plots/metrics/params(?): stop throwing exceptions


pared:
After writing fix for #5490 It seems to me that we should be handling exceptions thrown during `plots/metrics/params` differently. 

For some time now we have been drifting towards approach that communicates problems with warnings instead of exceptions. And in case of `plots/metrics/params` commands it does make sense, as its ok for those commands to do "as much as possible" (eg, show all metrics besides that faulty one, instead of crippling whole command and not allowing user to see anything). This approach is also not ideal, every now and then I have discussion with @dmpetrov on why `plots/metrics` flood terminal with warnings in some cases, which reminds me of time when we used a lot of `warn` calls in other parts of code, and had the same problem.

I think that we should gather the exceptions preventing us from displaying particular `file` at particular `revision` and return them as part of result and let the error handling happen after doing as much as possible. Errors can still be logged to debug, and in case of `dvc plots/metrics` command we could issue single warning, like "10 out of 1000 plots could not be shown. Use -v to see why".

dberenbaum:
Candidate to convert from issue -> discussion?

dberenbaum:
We have had a couple people in Discord try to use `dvc exp show` to compare metrics across all commits, but it breaks because they have some old commit that throws an error, so I think this is probably higher priority now with experiments. See https://discord.com/channels/485586884165107732/563406153334128681/822468979364593704 for context.

jorgeorpinel:
> its ok for those commands to do "as much as possible"

Agree! In general commands that print any kind of report should try to recover from individual errors and still give a report with some `MISSING/INVALID` etc. fields ( + printing detailed warnings or non-fatal errors to stdout perhaps).

> Candidate to convert from issue -> discussion?

For now (if there's consensus on the proposal) I'd suggest breaking this into several `product` issues, one per reporting-command (diff, metrics/plots ..., params ..., list, exp show/diff) so we can QA each one. And/or keep this one as a `product` `epic`, and add a bunch of checkbox tasks.

> couple people in Discord try to use dvc exp show to compare metrics across all commits, but it breaks

Can you maybe explain how old commits can break `exp show` in #5667 @dberenbaum? Please 🙂

dberenbaum:
> Can you maybe explain how existing commits can break `exp show` in #5667 @dberenbaum? Please 🙂

Here's the discord context:

> my only issue is that the first time i ran the dvc pipeline I had an error in my metrics output json file because it had duplicate keys in the json file

[...]

PR #5845, opened 2021-04-19
metrics: Skip revisions with corrupt dvc.lock files

### LINKED ISSUE #5844: metrics show --all-commits fails on corrupt historic commits
When using `metrics show --all-commits` we got failures due to a historic commit where a `dvc.lock` file was committed with merge conflicts in it.

I think it would be best to skip (and log warnings) revisions where a `LockfileCorruptedError` is thrown.

jheister:
Fixes #5844

### DIFF
--- a/dvc/repo/metrics/show.py
+++ b/dvc/repo/metrics/show.py
@@ -1,6 +1,7 @@
 import logging
 from typing import List
 
+from dvc.dvcfile import LockfileCorruptedError
 from dvc.exceptions import (
     MetricDoesNotExistError,
     NoMetricsFoundError,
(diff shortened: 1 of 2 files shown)


pmrowla:
Hi @jheister, thanks for the PR. This does correct the specific `metrics show` issue you found in #5844, but as @dberenbaum noted already, this is really related to a larger problem in DVC (https://github.com/iterative/dvc/issues/5525).

There are more types of potential errors we need to be looking for here than just `LockfileCorruptedError`, and ideally the fix for this will go in `repo.collect` and possible `repo.brancher` since the underlying problem is not actually specific to metrics - it's really related to how we collect pipelines/stages across git commits in general.

I'll leave this PR open for now, but I'm not sure it will get merged since we are already planning on addressing the problem with a more general fix for #5525.

pared:
@jheister
Thank your for your contribution!
 As @pmrowla mentioned we are currently working on two issues originating in same problem: #5667 and #5525. While your solution is correct, it solves only one of multiple problems that can occur during collection of metrics. To get rid of the source problem we need to fix this problem for all possible Exceptions thrown (during collection) and for all commands related to collecting `metrics/params/plots` from the repository. So the fix here needs to cover `dvc plots/params/metrics/exp show` commands. As I mentioned we are working on that, and probably the fix will cover also this use case. We cannot merge this fix, until we have all the use cases covered (and once we have them this fix might be too specific to be included). Having said so I would like to keep this PR open until we propose the "general" solution, to be sure we do not omit this particular use case.

efiop:
Closing in favor of #5525

pared:
@jheister seems like #5984 fixed the issue

PR #5852, opened 2021-04-20
ui: tabular data

skshetry:

This PR combines the earlier `ui.table()` that we had, and provides a nice interfaces around the table as a part of the data structure. Please take a look at how we use this in #5872.

### DIFF
--- a/dvc/compare.py
+++ b/dvc/compare.py
@@ -1,0 +1,169 @@
+from itertools import chain, repeat, zip_longest
+from operator import itemgetter
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    ItemsView,
+    Iterable,
+    Iterator,
+    List,
+    MutableSequence,
+    Sequence,
+    Tuple,
+    Union,
+    overload,
+)
+
+from funcy import reraise
+
+if TYPE_CHECKING:
+    from dvc.ui.table import CellT
+
+
+class Column(List["CellT"]):
+    pass
+
+
+def with_value(value, default):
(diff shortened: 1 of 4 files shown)


dvc/compare.py
```diff
@@ -0,0 +1,169 @@
+from itertools import chain, repeat, zip_longest
```
efiop, line 1: Should this file go to dvc/ui or dvc/utils?

skshetry: @efiop, this is not really about UI, most of it is unifying the diff/show which I am thinking might be a part of the unified diff/show that @dberenbaum is proposing, so it's not just `utils` as well. It might be easier later this way. 🙂

skshetry: Though I am open for better name that covers both `show` and `diff`. 😄

skshetry: Ahh, I get the confusion now. #5872 will immediately follow after this, so I am trying to keep these show/diff in a single place.

dvc/compare.py
```diff
@@ -0,0 +1,169 @@
...
+    def __getitem__(self, item: Union[int, slice]):
+        func = itemgetter(item)
+        it = map(func, self.columns)
+        if isinstance(item, slice):
+            it = map(list, zip(*it))
+        return list(it)
+
+    @overload
+    def __setitem__(self, item: int, value: Sequence["CellT"]) -> None:
+        ...
```
efiop, line 88: Whoa, this is new :slightly_smiling_face:

efiop (approved):

Issue #2418, opened 2019-08-20
Investigate type for `RemoteLOCAL.protected`


pared:
https://github.com/iterative/dvc/blob/4fb4b84a5fd7d3bd0da20e9ab56a2fdbb766dbd9/dvc/remote/local/__init__.py#L66

is None, if not set. Investigate, why it does not have default value of False.

example fail:
put
`assert self.protected is not None`
right after it and run  `test_add.TestAddDirectory`

pared:
The issue no longer exists, closing.

PR #5859, opened 2021-04-21
exp run: `dvc commit` DVC-tracked data deps when stashing an experiment

### LINKED ISSUE #5593: exp run: checks out data dependencies (destructive) [QA]

pmrowla:

Will close #5593

- Data dependencies are now `dvc commit`ed internally when stashing an experiment so that any modifications to that data dep are preserved in both workspace and tempdir runs (previously the changes were dropped entirely by the `exp run` `dvc checkout` step).

### DIFF
--- a/tests/func/experiments/test_experiments.py
+++ b/tests/func/experiments/test_experiments.py
@@ -624,3 +624,47 @@ def test_fix_exp_head(tmp_dir, scm, tail):
     scm.set_ref(EXEC_BASELINE, "refs/heads/master")
     assert EXEC_BASELINE + tail == fix_exp_head(scm, head)
     assert "foo" + tail == fix_exp_head(scm, "foo" + tail)
+
+
+@pytest.mark.parametrize("workspace", [True, False])
+def test_modified_data_dep(tmp_dir, scm, dvc, workspace):
+    tmp_dir.dvc_gen("data", "data")
+    tmp_dir.gen("copy.py", COPY_SCRIPT)
+    tmp_dir.gen("params.yaml", "foo: 1")
+    exp_stage = dvc.run(
+        cmd="python copy.py params.yaml metrics.yaml",
+        metrics_no_cache=["metrics.yaml"],
+        params=["foo"],
+        name="copy-file",
+        deps=["copy.py", "data"],
+    )
+    scm.add(
+        [
+            "dvc.yaml",
+            "dvc.lock",
+            "copy.py",
+            "params.yaml",
+            "metrics.yaml",
+            "data.dvc",
+            ".gitignore",
+        ]
+    )
+    scm.commit("init")
+
+    tmp_dir.gen("params.yaml", "foo: 2")
+    tmp_dir.gen("data", "modified")
+
+    results = dvc.experiments.run(exp_stage.addressing, tmp_dir=not workspace)
+    exp = first(results)
+
+    for rev in dvc.brancher(revs=[exp]):
+        if rev != exp:
+            continue
+        with dvc.repo_fs.open(tmp_dir / "metrics.yaml") as fobj:
+            assert fobj.read().strip() == "foo: 2"
+        with dvc.repo_fs.open(tmp_dir / "data") as fobj:
+            assert fobj.read().strip() == "modified"
+
+    if workspace:
+        assert (tmp_dir / "metrics.yaml").read_text().strip() == "foo: 2"
+        assert (tmp_dir / "data").read_text().strip() == "modified"
--- a/tests/func/experiments/test_experiments.py
+++ b/tests/func/experiments/test_experiments.py
@@ -626,8 +626,11 @@ def test_fix_exp_head(tmp_dir, scm, tail):
     assert "foo" + tail == fix_exp_head(scm, "foo" + tail)
 
 
-@pytest.mark.parametrize("workspace", [True, False])
-def test_modified_data_dep(tmp_dir, scm, dvc, workspace):
+@pytest.mark.parametrize(
+    "workspace, params",
+    [(True, "foo: 1"), (True, "foo: 2"), (False, "foo: 1"), (False, "foo: 2")],
+)
+def test_modified_data_dep(tmp_dir, scm, dvc, workspace, params):
     tmp_dir.dvc_gen("data", "data")
     tmp_dir.gen("copy.py", COPY_SCRIPT)
     tmp_dir.gen("params.yaml", "foo: 1")
@@ -651,7 +654,7 @@ def test_modified_data_dep(tmp_dir, scm, dvc, workspace):
     )
     scm.commit("init")
 
-    tmp_dir.gen("params.yaml", "foo: 2")
+    tmp_dir.gen("params.yaml", params)
     tmp_dir.gen("data", "modified")
 
     results = dvc.experiments.run(exp_stage.addressing, tmp_dir=not workspace)
@@ -661,10 +664,10 @@ def test_modified_data_dep(tmp_dir, scm, dvc, workspace):
         if rev != exp:
             continue
         with dvc.repo_fs.open(tmp_dir / "metrics.yaml") as fobj:
-            assert fobj.read().strip() == "foo: 2"
+            assert fobj.read().strip() == params
         with dvc.repo_fs.open(tmp_dir / "data") as fobj:
             assert fobj.read().strip() == "modified"
 
     if workspace:
-        assert (tmp_dir / "metrics.yaml").read_text().strip() == "foo: 2"
+        assert (tmp_dir / "metrics.yaml").read_text().strip() == params
         assert (tmp_dir / "data").read_text().strip() == "modified"
--- a/tests/func/experiments/test_experiments.py
+++ b/tests/func/experiments/test_experiments.py
@@ -1,3 +1,4 @@
+import itertools
 import logging
 import os
 import stat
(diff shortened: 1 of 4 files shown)


tests/func/experiments/test_experiments.py
```diff
@@ -624,3 +624,47 @@ def test_fix_exp_head(tmp_dir, scm, tail):
...
+            "copy.py",
+            "params.yaml",
+            "metrics.yaml",
+            "data.dvc",
+            ".gitignore",
+        ]
+    )
+    scm.commit("init")
+
+    tmp_dir.gen("params.yaml", "foo: 2")
```
dberenbaum, line 654: Maybe include a test where params don't change (only data)?

dberenbaum: Also maybe test multiple modified data deps?

dberenbaum (commented):
Have a couple of test suggestions, but otherwise looks good!

The downside to this behavior will be that it will potentially take a long time to queue an experiment, right? Maybe we need to document that users should do `dvc checkout` to get back to their original data if they don't want to use the data changes in their workspace. Thoughts @jorgeorpinel?

jorgeorpinel:
Thanks @pmrowla 🙏 

> Data dependencies are now dvc commited internally when stashing an experiment ... previously the changes were dropped

Q#1. What happens to .dvc and dvc.lock files in the working tree if `run` fails after the commit? I'm guessing nothing (only changed in the tmp dir)

Q#2. by "stashin" we litterally mean `git stash`? I didn't realize we still use that internally.

> The downside to this behavior will be that it will potentially take a long time to queue an experiment, right? Maybe we need to document that users should do dvc checkout to get back to their original data

@dberenbaum idk if `commit` is usually reported as a slow command but I guess it can be for large datasets?

As for `dvc checkout` maybe I didn't get it but I don't think that would change anything (assuming .dvc and dvc.lock files aren't modified in the workspace).

pmrowla:
> The downside to this behavior will be that it will potentially take a long time to queue an experiment, right?

@dberenbaum This will depend on the complexity of the user's pipeline and how many data deps they have, but yes it will slow down queueing.

> Q#1. What happens to .dvc and dvc.lock files in the working tree if run fails after the commit? I'm guessing nothing (only changed in the tmp dir)

If `exp run` fails in a temp dir nothing in the workspace will change. If it fails in the workspace, the workspace will contain any (non-git-committed) changes to .dvc and dvc.lock files that were made prior to the failure. So .dvc files will be modified and contain the hashes from our `dvc commit` step.

> Q#2. by "stashing" do we litterally mean git stash? I didn't realize we still use that internally.

Queued experiments are git stash (merge) commits. We don't use the standard `git stash` ref directly, but functionality wise, the experiment queue is just a custom git stash.

@jorgeorpinel

tests/func/experiments/test_experiments.py
pmrowla: Updated so we test with and without params changes (so we ensure that a new exp branch can be triggered solely on modified data deps).

Testing w/multiple data deps should not be necessary here, we are re-using the existing `repo.commit`/`stage.commit` functionality, handling changes in multiple deps is already tested at the `commit` level elsewhere.

dberenbaum (changes requested):
I tried out the first test manually and it seems like there's maybe another test needed upstream somewhere.

Here's what I did:
```bash
mkdir repo
cd repo
git init
dvc init -q
git add .
git commit --quiet -m "init"
echo data > data
dvc add data
echo "import sys; import shutil; shutil.copyfile(sys.argv[1], sys.argv[2])" > copy.py
echo "foo: 1" > params.yaml
dvc run -n copy-file -M metrics.yaml -p foo -d copy.py -d data python copy.py params.yaml metrics.yaml
git add .
git commit -m "run stage"
echo modified > data
dvc exp run
```
The output from `dvc exp run` is:
```console
'data.dvc' didn't change, skipping
Stage 'copy-file' didn't change, skipping
```
No experiment is created, and `data` is reverted back from `modified` to `data`.

dberenbaum:
@jorgeorpinel I think we may just need to document that large changes to data dependencies in the workspace may slow down experiment queueing.

pmrowla:
> I tried out the first test manually and it seems like there's maybe another test needed upstream somewhere.

@dberenbaum bug was w/calling `exp run` without a target (the original test always called it with a specific target stage), should be fixed now and the test has been updated

dberenbaum (approved):

jorgeorpinel:
> If it fails in the workspace, the workspace will contain any (non-git-committed) changes to .dvc and dvc.lock files ... contain the hashes from our dvc commit step.

Hm... Is this something we should be worried about? Let's wait and see I guess.
Out of curiosity, how feasible would it be to make it a transaction-type operation so the initial `dvc commit` gets rolled back (or happens in a tmp dir and gets discarded) if the run fails?

> We don't use the standard git stash ref directly, but functionality wise, the experiment queue is just a custom git stash.

Should've mentioned it in your blog post!

> I think we may just need to document that large changes to data dependencies in the workspace may slow down experiment queueing.

Agree. Created #2418

pmrowla:
> > If it fails in the workspace, the workspace will contain any (non-git-committed) changes to .dvc and dvc.lock files ... contain the hashes from our dvc commit step.
> 
> Hm... Is this something we should be worried about? Let's wait and see I guess.
> Out of curiosity, how feasible would it be to make it a transaction-type operation so the initial `dvc commit` gets rolled back (or happens in a tmp dir and gets discarded) if the run fails?

We can do this if we want to, but I don't think this is what users would expect. If `dvc repro` fails at some random point, we don't roll anything back, and the failed state is left in the repo. `exp run` should work the same way for experiments, so that the user can debug what happened.

If the user wants to roll back the repo state they can use `git reset --hard` the same way that they would for a failed `repro`.

jorgeorpinel:
The initial `dvc commit` is also not obvious so who knows if having changed `md5` sums in metafiles will help users debug. But yeah just wondering, we should prob let it be unless users complain.

Issue #5768, opened 2021-04-05
update: when using --to-remote, only update missing cache files


isidentical:
Currently what `update --to-remote` does is the same exact thing with `import-url --to-remote` with only except that update is using the stage's url instead of taking it as an argument. This means that for each source file, even though they might already exist on our remote storage: we download them, calculate their hashes, and upload them to the remote storage. 

For `import --to-remote`; downloading, uploading, and hash calculation happens at the same time, which means we only do `1 download + 1 upload `. Though this is not as efficient as it might get for `update --to-remote` with a directory that consists of small files. For each file, do the same actions without ever checking whether they are present or not. Since we can not determine their presence beforehand (before actually downloading the file and calculating hash in chunks), we can optimize the `update` in a way that works for cases that the updated files are always less than total amount of files for a directory that consist of small files.

So instead of doing `1 download + 1 upload` on every case,  we can do `1 download` on best cases (when the file already exists) and `2 download + 1 upload` for worst cases (when the file doesn't exist on the cache).  This might be a bad optimization when the changes on the source is always bigger than the remaining, unchanged part though I think it is a bit unusual and rare case.

CC: @efiop @dberenbaum

dberenbaum:
> So instead of doing `1 download + 1 upload` on every case, we can do `1 download` on best cases (when the file already exists) and `2 download + 1 upload` for worst cases (when the file doesn't exist on the cache).

Sorry, @isidentical, can you explain in a bit more detail what's happening in each case? I'm not following why an extra download is required if the file doesn't exist.

isidentical:
> I'm not following why an extra download is required if the file doesn't exist.

For finding the only missing files, we need to calculate the hashes and find out which one of those hashes are not present in the remote. Unfortunately, we can not calculate the hashes without downloading all the files beforehand. So instead of wasting an upload on a file that might exist, we initially download everything (actually we don't download anything to anywhere, just calculate the hashes, but we waste the same time) and then find out the missing hashes and apply the normal procedure (like this time download normally as in chunks and upload that chunks to the remote). 

```
import-url --to-remote => download [from url] -> md5 -> upload [to remote storage]
update => download [from url] -> md5 -> download [from url, but only missing files on the remote storage] -> upload [to remote storage]
```

```

dberenbaum:
How does the md5 step work? Is it calculated per file? Is each file read into memory in its entirety?

dberenbaum:
[...]

PR #5860, opened 2021-04-21
objects: don't recompute hash on odb.add() if verify is False

isidentical:
When adding new objects with `odb.add()` when the `odb.verify` is set to False (or defaulted to False, which happens on almost all remotes except GDrive), instead of recomputing the hash again we'll simply check the existence of the cache file and not verify the contents. This speeds up the `dvc update --to-remote` by  %15 (from `2.49 minutes` to `2.24 minutes`). Part of #5768


efiop (approved):
For the record: test failures are unrelated.

PR #5861, opened 2021-04-21
output: introduce BaseOutput.transfer

isidentical:
Removes `DataCloud.transfer`/`BaseRemote.transfer` and introduces `BaseOutput.transfer`. Part of #5768

### DIFF
--- a/dvc/output/base.py
+++ b/dvc/output/base.py
@@ -17,6 +17,7 @@ from dvc.exceptions import (
     RemoteCacheRequiredError,
 )
 from dvc.hash_info import HashInfo
+from dvc.objects import save as osave
 from dvc.objects.db import NamedCache
 from dvc.objects.errors import ObjectFormatError
 from dvc.objects.stage import stage as ostage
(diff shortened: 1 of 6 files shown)


dvc/output/base.py
```diff
@@ -470,6 +471,48 @@ def move(self, out):
         if self.scheme == "local" and self.use_scm_ignore:
             self.repo.scm.ignore(self.fspath)
 
+    def transfer(
```
efiop, line 474: For the record: this might eventually turn into `commit` or get somehow unified with it, but at least having `transfer` here is great for now.

dvc/output/base.py
```diff
@@ -470,6 +471,48 @@ def move(self, out):
...
+        jobs=None,
+        update=False,
+        remote=None,
+        command=None,
+        no_progress_bar=False,
+    ):
+        from dvc.fs import get_cloud_fs
+
+        remote = self.repo.cloud.get_remote(remote, command)
+        odb = remote.odb
```
efiop, line 486: For `to-cache` it will be `self.odb`, right? Then maybe consider passing remote `odb` to `out.transfer` so it could use it and if not specified - fallback to self.odb.

efiop:
@isidentical Please rebase.

isidentical:
> @isidentical Please rebase.

Rebased (though I couldn't see any conflicts)

efiop:
@isidentical That was for codecov update :) Just to make CI not fail.

efiop (approved):

PR #5867, opened 2021-04-23
Add full stops in CLI help

jankrepl:

# Description

```
dvc --help
```
The description of some commands does not end with a full stop. It is a stupid PR but for some reason this inconsistency stood out in my eyes.

### DIFF
--- a/dvc/command/plots.py
+++ b/dvc/command/plots.py
@@ -95,7 +95,7 @@ class CmdPlotsModify(CmdPlots):
 def add_parser(subparsers, parent_parser):
     PLOTS_HELP = (
         "Commands to visualize and compare plot metrics in structured files "
-        "(JSON, YAML, CSV, TSV)"
+        "(JSON, YAML, CSV, TSV)."
     )
 
     plots_parser = subparsers.add_parser(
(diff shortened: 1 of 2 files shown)


efiop (approved):
Thanks!

dvc/command/plots.py
```diff
@@ -95,7 +95,7 @@ def run(self):
 def add_parser(subparsers, parent_parser):
     PLOTS_HELP = (
         "Commands to visualize and compare plot metrics in structured files "
-        "(JSON, YAML, CSV, TSV)"
+        "(JSON, YAML, CSV, TSV)."
```
jorgeorpinel, line 98: Many of these are not full not sentences though, so no full stop was needed.

jorgeorpinel: In this case this one was correct without period.

dvc/command/unprotect.py
```diff
@@ -23,7 +23,7 @@ def run(self):
 def add_parser(subparsers, parent_parser):
     UNPROTECT_HELP = (
         "Unprotect tracked files or directories (when hardlinks or symlinks "
-        "have been enabled with `dvc config cache.type`)"
+        "have been enabled with `dvc config cache.type`)."
```
jorgeorpinel, line 26: But this one is a sentence indeed so `.` = 👍

dvc/command/plots.py
jorgeorpinel: Oh well it's not the only "wrong" one (very minor too), we can let it be.

PR #5872, opened 2021-04-26
refactor diff and show to use new table ui and new data structure

### LINKED ISSUE #5576: ui: what should `--quiet` do for tabular outputs?
What does the `--quiet` mean for our DVC commands? 

It can have two different meanings:
1. Don't show up verbose error messages, but show messages that do matter (important warnings + what the user has asked for).
2. Display nothing at all.

I have seen a few tools using `--silent` instead for 2). But, I guess the most important thing is being consistent here.

The following are the ones that do print on `--quiet`:
1. `stage list`
2. `exp show`
3. `completion`
4. `remote default`

Other commands such as `metrics/params/plots` don't show anything at all (what's the use here of `--quiet`?).

P.S. What should `dvc repro` and `dvc exp run` do?

### LINKED ISSUE #5591: metrics show/diff: UnicodeEncodeError: 'ascii' codec can't encode character '\u2014'
# Bug Report

## Description

I have an encoding error on one of the remote server I use. I think it is linked to the default encoding which is configured on this machine. But I wanted to raise this issue just in case you didn't know that for some particular configuration, displaying metrics can fail. The problematic character is `—`, which is used when a metric value is missing. For the same project, `dvc metrics show` and `dvc metrics diff` display it normally on my local dev machine but on the remote server it fails with the encoding error.

### Reproduce

Basically, take a machine (or a docker container, see the image I used in the next section) with ascii encoding being used for the console and run `dvc metrics show` or `dvc metrics diff` in a context in which a metric value is missing.

For instance, make two metric files with different metric names. If you run `dvc metrics show`, it will try to display something like this:

```console
                                 Metric Name 1     Metric Name 2
metric_file_1.json               0.15833           —          
metric_file_2.json               —                 0.98103
```

If the default encoding is ascii, this will fail.

### Expected

I would expect the command to work in all contexts. For instance, the `—` character could be replaced with `-` or `NA` if it cannot be displayed.

### Environment information
I run my code in a docker container based on `nvidia/cuda:10.0-cudnn7-devel-ubuntu16.04`

**Output of `dvc doctor`:**

```console
$ dvc doctor
DVC version: 2.0.3 (pip)
Platform: Python 3.6.12 on Linux-3.10.0-1160.11.1.el7.x86_64-x86_64-with-debian-stretch-sid
Supports: hdfs, http, https, ssh
Cache directory: overlay on overlay
Caches: local
Remotes: ssh
Workspace directory: overlay on overlay
Repo: dvc, git
```

```console
$ dvc metrics show --verbose
  File "/opt/conda/envs/my-env/lib/python3.6/site-packages/dvc/logger.py", line 136, in emit
    msg, file=self.stream, end=getattr(self, "terminator", "\n")
  File "/opt/conda/envs/my-env/lib/python3.6/site-packages/tqdm/std.py", line 619, in write
    fp.write(s)
UnicodeEncodeError: 'ascii' codec can't encode character '\u2014' in position 301: ordinal not in range(128)

During handling of the above exception, another exception occurred:

  File "/opt/conda/envs/my-env/lib/python3.6/site-packages/dvc/main.py", line 50, in main
    ret = cmd.run()
  File "/opt/conda/envs/my-env/lib/python3.6/site-packages/dvc/command/metrics.py", line 100, in run
    logger.info(table)
  File "/opt/conda/envs/my-env/lib/python3.6/logging/__init__.py", line 1308, in info
    self._log(INFO, msg, args, **kwargs)
  File "/opt/conda/envs/my-env/lib/python3.6/logging/__init__.py", line 1444, in _log
    self.handle(record)
  File "/opt/conda/envs/my-env/lib/python3.6/logging/__init__.py", line 1454, in handle
    self.callHandlers(record)
  File "/opt/conda/envs/my-env/lib/python3.6/logging/__init__.py", line 1516, in callHandlers
    hdlr.handle(record)
  File "/opt/conda/envs/my-env/lib/python3.6/logging/__init__.py", line 865, in handle
    self.emit(record)
  File "/opt/conda/envs/my-env/lib/python3.6/site-packages/dvc/logger.py", line 142, in emit
    self.handleError(record)
  File "/opt/conda/envs/my-env/lib/python3.6/site-packages/dvc/logger.py", line 129, in handleError
    raise LoggingException(record)
dvc.logger.LoggingException: failed to log <LogRecord: dvc.command.metrics, 20, /opt/conda/envs/my-env/lib/python3.6/site-packages/dvc/command/metrics.py, 100
```

I cannot show the whole log because it contains confidential metrics but I think that with these information you will be able to understand where is the issue in the code of DVC. Otherwise, don't hesitate to ask me :wink:

skshetry:

We have introduce `ui.table()` elements and new data structure in #5852. This will unify all diffs/show into `dvc.compare` module and will make use of the new tabular data structure and the new `ui.table`. I am really sorry for quite a large PR (most of those are the test changes).

Fixes #5576 and fixes #5591.

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -447,6 +447,7 @@ class CmdExperimentsDiff(CmdBase):
         else:
             from dvc.compare import show_diff
 
+            precision = self.args.precision or DEFAULT_PRECISION
             diffs = [("metrics", "Metric"), ("params", "Param")]
             for idx, (key, title) in enumerate(diffs):
                 if idx:
@@ -459,7 +460,7 @@ class CmdExperimentsDiff(CmdBase):
                     no_path=self.args.no_path,
                     old=self.args.old,
                     on_empty_diff="diff not supported",
-                    precision=self.args.precision or DEFAULT_PRECISION,
+                    precision=precision if key == "metrics" else None,
                 )
 
         return 0
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -426,43 +426,6 @@ class CmdExperimentsApply(CmdBase):
         return 0
 
 
-def _show_diff(
-    diff,
-    title="",
-    markdown=False,
-    no_path=False,
-    old=False,
-    precision=DEFAULT_PRECISION,
-):
-    from dvc.utils.diff import table
-
-    rows = []
-    for fname, diff_ in diff.items():
-        sorted_diff = OrderedDict(sorted(diff_.items()))
-        for item, change in sorted_diff.items():
-            row = [] if no_path else [fname]
-            row.append(item)
-            if old:
-                row.append(_format_field(change.get("old"), precision))
-            row.append(_format_field(change["new"], precision))
-            row.append(
-                _format_field(
-                    change.get("diff", "diff not supported"), precision
-                )
-            )
-            rows.append(row)
-
-    header = [] if no_path else ["Path"]
-    header.append(title)
-    if old:
-        header.extend(["Old", "New"])
-    else:
-        header.append("Value")
-    header.append("Change")
-
-    return table(header, rows, markdown)
-
-
 class CmdExperimentsDiff(CmdBase):
     def run(self):
 
@@ -473,41 +436,38 @@ class CmdExperimentsDiff(CmdBase):
                 all=self.args.all,
                 param_deps=self.args.param_deps,
             )
-
-            if self.args.show_json:
-                import json
-
-                logger.info(json.dumps(diff))
-            else:
-                if self.args.precision is None:
-                    precision = DEFAULT_PRECISION
-                else:
-                    precision = self.args.precision
-
-                diffs = [("metrics", "Metric"), ("params", "Param")]
-                for key, title in diffs:
-                    table = _show_diff(
-                        diff[key],
-                        title=title,
-                        markdown=self.args.show_md,
-                        no_path=self.args.no_path,
-                        old=self.args.old,
-                        precision=precision,
-                    )
-                    if table:
-                        logger.info(table)
-                        logger.info("")
-
         except DvcException:
             logger.exception("failed to show experiments diff")
             return 1
+
+        if self.args.show_json:
+            import json
+
+            logger.info(json.dumps(diff))
+        else:
+            from dvc.compare import show_diff
+
+            diffs = [("metrics", "Metric"), ("params", "Param")]
+            for idx, (key, title) in enumerate(diffs):
+                if idx:
+                    logger.info("")
+
+                show_diff(
+                    diff[key],
+                    title=title,
+                    markdown=self.args.show_md,
+                    no_path=self.args.no_path,
+                    old=self.args.old,
+                    on_empty_diff="diff not supported",
+                    precision=self.args.precision or DEFAULT_PRECISION,
+                )
 
         return 0
 
 
 class CmdExperimentsRun(CmdRepro):
     def run(self):
-        from dvc.command.metrics import _show_metrics
+        from dvc.compare import show_metrics
 
         if self.args.checkpoint_resume:
             if self.args.reset:
@@ -538,7 +498,7 @@ class CmdExperimentsRun(CmdRepro):
         if self.args.metrics and results:
             metrics = self.repo.metrics.show(revs=list(results))
             metrics.pop("workspace", None)
-            logger.info(_show_metrics(metrics))
+            logger.info(show_metrics(metrics))
 
         return 0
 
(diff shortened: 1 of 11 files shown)


dvc/compare.py
```diff
@@ -0,0 +1,298 @@
...
+
+    def as_dict(self):
+        keys = self.keys()
+        return [dict(zip(keys, row)) for row in self]
+
+
+def _format_field(val: Any, precision: int = None) -> str:
+    def _format(_val):
+        if isinstance(_val, float) and precision:
+            fmt = f"{{:.{precision}f}}"
```
skshetry, line 176: I changed the precision from `g` specifier to `f` now, which is how other `params/metrics` diff worked.
This might affect experiments.

dvc/compare.py
```diff
@@ -0,0 +1,298 @@
...
+            return float(fmt.format(_val))
+        if isinstance(_val, Mapping):
+            return {k: _format(v) for k, v in _val.items()}
+        if isinstance(_val, list):
+            return [_format(x) for x in _val]
+        return _val
+
+    return str(_format(val))
```
skshetry, line 184: Rather than stringifying each element of the composite data (list/dict), I am doing this to the whole data to have an easier output. Might affect experiments.

Before: `["2", "3", "4"]`
After: `[2, 3, 4]`

dvc/command/experiments.py
```diff
@@ -473,41 +436,38 @@ def run(self):
...
+            import json
+
+            logger.info(json.dumps(diff))
+        else:
+            from dvc.compare import show_diff
+
+            diffs = [("metrics", "Metric"), ("params", "Param")]
+            for idx, (key, title) in enumerate(diffs):
+                if idx:
+                    logger.info("")
```
skshetry, line 453: Use of `logger.info` has a side-effect of breaking separation between tables on `--quiet`. Will fix when I come to replacing `logger.info`.

efiop:
@skshetry Please rebase. Also, are there any visual changes here?

skshetry:
> @skshetry Please rebase. Also, are there any visual changes here? If there are, could you show before/after?

The actual value of `value` in params.yaml being ` 1.123455`.
```
# before
```console
# before
$ dvc exp diff --precision 4
Path         Param            Old    New               Change
params.yaml  value            None   1.123        diff not supported

# after
dvc exp diff --precision 4
Path         Param            Old    New                Change
params.yaml  value            -      1.1235          diff not supported
```

There are three changes here:
1. Precision: We used to use `round()` in metrics diff/show and `%g` specifier in `exp diff/show`. This has been changed to `%f` so the no. of digits match the `--precision` value (as far I understand, `%g` goes for small repr). Similarly, as we used to use `round`, it might have weird output for some floats. They are now changed to always being a string.
2. `exp diff/show` used to show `None` values as it is as opposed to em-dash in `metrics/params diff/show`. This has been changed to fill the empty value instead.
3. The em-dash as a fill value has been changed to just a hyphen in `metrics/params diff/show` to fix #5591.

dvc/compare.py
pmrowla: This will break the experiments table, for very big or very small numbers we want to display them in scientific notation (with the correct precision), and for everything else we want to display them as regular decimals, which is why it was `g` in experiments. It seems like this should also just be the default behavior everywhere in DVC.

pmrowla: according to the python docs, to keep the number of digits when it's displayed as a regular decimal, you can use `#` in conjunction with `g`

so what we probably want here is 
`f"{{:#.{precision}g}}"`

pmrowla (changes requested):
see notes about using `g` instead of `f`

dvc/compare.py
skshetry: `#` specifier only affects the trailing zeros. The problem with `g` specifier is that it uses ` P − (exp + 1)` precision for fixed-point formatting and `P-1` for exponent-based formatting.

I guess it depends on what we are aiming for with `precision`, precision in terms of significant digits or in terms of fractional digits. The docs [points to the latter](https://dvc.org/doc/command-reference/exp/show#options):

>  - round decimal values to n digits of precision (5 by default). Applies to metrics only.

pmrowla: I think we should go with precision meaning significant digits and update the docs accordingly. @dberenbaum thoughts on this?

The switch to `g` for exp commands was made specifically because users had metrics/params with very small magnitudes, and using `f` will result in rounding the number down to zero, which is definitely not what DVC should be displaying in the exp table (or in the other diff/show commands).

skshetry: @pmrowla, significant digits makes sense to me as well. I changed to `g`, but I can rollback later if we decide so (changing to `g` does affect metrics/params diff/show). I'll update docs separately.

pmrowla (approved):

skshetry:
Guys, I am merging this as I'm getting blocked with the other PRs. Please do post-merge, I am really sorry about this. :(

PR #5878, opened 2021-04-26
fs: introduce tree.find()

### LINKED ISSUE #5877: fsspec: use find() for walk_files() and optimize status calls
Currently, we use `.ls(recursive=True)` for `walk_files()` under the hood, which is not a really efficient method especially for cloud providers where they don't actually offer a proper listing directory-based API without actually retrieving a part of the results. The `.ls(recursive=True)` is something that we intend to remove and replace in the future with the proper APIs that it is using (currently it is just a proxy to `walk()`).

The `walk()` calls for object-based storages are much much slower than the actual `find()` with passing the base prefix. It is always faster than the `walk()`, and in some cases, the gap between calls might go through over the roof (15-20x, if you have like 2000 hashes separated between 256 prefixes, then `walk()` will send an API request for each prefix `00/`, `01/` but the `find()` will only cost a couple of API requests and list the whole contents which are what we need for `walk_files()`). 

P.S: This is something that I've noticed while adding new s3fs benchmark scenarios, so this would also help on s3fs migration.

isidentical:
Resolves #5877. This patch also drops `ls(recursive=True)` option, and replaces all usages with the new `Tree.find()` (fsspec compliant API).

### DIFF
--- a/dvc/fs/fsspec_wrapper.py
+++ b/dvc/fs/fsspec_wrapper.py
@@ -33,15 +33,13 @@ class FSSpecWrapper(BaseFileSystem):
 
         return path or bucket
 
-    def _strip_buckets(self, entries, detail, prefix=None):
+    def _strip_buckets(self, entries, detail=False):
         for entry in entries:
             if detail:
                 entry = self._entry_hook(entry.copy())
                 entry["name"] = self._strip_bucket(entry["name"])
             else:
-                entry = self._strip_bucket(
-                    f"{prefix}/{entry}" if prefix else entry
-                )
+                entry = self._strip_bucket(entry)
             yield entry
 
     def _entry_hook(self, entry):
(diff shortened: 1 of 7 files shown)


dvc/fs/fsspec_wrapper.py
```diff
@@ -93,24 +91,29 @@ def copy(self, from_info, to_info):
...
+        files = self.fs.find(path, detail=detail)
+        if detail:
+            files = files.values()
+
+        # When calling find() on a file, it returns the same file in a list.
+        # For object-based storages, the same behavior applies to empty
+        # directories since they are represented as files. This condition
+        # checks whether we should yield an empty list (if it is an empty
+        # directory) or just yield the file itself.
+        if len(files) == 1 and files[0] == path and self.isdir(path_info):
```
isidentical, line 110: This check is intentionally omitted from the bare `ls()` since its main objective is to support directory listing when calling `walk_files()` and I didn't want to repeat it in 2 different places.

isidentical:
🤦🏻‍♀️ pushed branch to the wrong repo

Issue #5871, opened 2021-04-25
s3 performance is slow


tyommik:
# Bug Report

 DVC (s3 client)
- Bucket 40Gb - 410k files (each file <= 400KB)

```
DVC version: 2.0.17 (deb)
---------------------------------
Platform: Python 3.8.8 on Linux-5.4.0-70-generic-x86_64-with-glibc2.4
Supports: All remotes
Cache types: hardlink, symlink
Cache directory: ext4 on /dev/sda
Caches: local
Remotes: s3
Workspace directory: ext4 on /dev/sda
Repo: dvc, git
```

When I did dvc pull -j 20 maximum speed was 80 mbit/sec but average was about 40. I download the bucket for 220 minutes.

What I tried else:
dvc pull -j 80 - no improvements. 
awscli - aws tool can maximum **160** Mbit/sec downloading speed. I tried different settings, but I couldn't exceed the limit.
s4cmd - maximum I got is about **130** Mbit/sec, 64minutes to get the bucket. 
s5cmd - maximum **960** Mbit/sec and less than 10 minutes to download the whole bucket.

So you can see that storage performance is okay but the download speed of tools written on python can not  reach maximum.

### Reproduce

Profiler stat: https://disk.yandex.ru/d/XNajwHgWYlPSHA

pared:
Might be influenced by https://github.com/iterative/dvc/pull/5683

PR #5879, opened 2021-04-26
fs: introduce fs.find()

### LINKED ISSUE #5877: fsspec: use find() for walk_files() and optimize status calls

isidentical:
Resolves #5877. This patch also drops ls(recursive=True) option and replaces all usages with the new fs.find() (fsspec compliant API).

### DIFF
--- a/dvc/fs/fsspec_wrapper.py
+++ b/dvc/fs/fsspec_wrapper.py
@@ -33,15 +33,13 @@ class FSSpecWrapper(BaseFileSystem):
 
         return path or bucket
 
-    def _strip_buckets(self, entries, detail, prefix=None):
+    def _strip_buckets(self, entries, detail=False):
         for entry in entries:
             if detail:
                 entry = self._entry_hook(entry.copy())
                 entry["name"] = self._strip_bucket(entry["name"])
             else:
-                entry = self._strip_bucket(
-                    f"{prefix}/{entry}" if prefix else entry
-                )
+                entry = self._strip_bucket(entry)
             yield entry
 
     def _entry_hook(self, entry):
@@ -93,24 +91,29 @@ class FSSpecWrapper(BaseFileSystem):
     def exists(self, path_info, use_dvcignore=False):
         return self.fs.exists(self._with_bucket(path_info))
 
-    def ls(
-        self, path_info, detail=False, recursive=False
-    ):  # pylint: disable=arguments-differ
-        if self.isdir(path_info) and self.is_empty(path_info):
-            return None
-
+    def ls(self, path_info, detail=False):
         path = self._with_bucket(path_info)
-        if recursive:
-            for root, _, files in self.fs.walk(path, detail=detail):
-                if detail:
-                    files = files.values()
-                yield from self._strip_buckets(files, detail, prefix=root)
+        files = self.fs.ls(path, detail=detail)
+        yield from self._strip_buckets(files, detail=detail)
+
+    def find(self, path_info, detail=False):
+        path = self._with_bucket(path_info)
+        files = self.fs.find(path, detail=detail)
+        if detail:
+            files = files.values()
+
+        # When calling find() on a file, it returns the same file in a list.
+        # For object-based storages, the same behavior applies to empty
+        # directories since they are represented as files. This condition
+        # checks whether we should yield an empty list (if it is an empty
+        # directory) or just yield the file itself.
+        if len(files) == 1 and files[0] == path and self.isdir(path_info):
             return None
 
-        yield from self._strip_buckets(self.ls(path, detail=detail), detail)
+        yield from self._strip_buckets(files, detail=detail)
 
     def walk_files(self, path_info, **kwargs):
-        for file in self.ls(path_info, recursive=True):
+        for file in self.find(path_info):
             yield path_info.replace(path=file)
 
     def remove(self, path_info):
--- a/dvc/fs/base.py
+++ b/dvc/fs/base.py
@@ -175,8 +175,11 @@ class BaseFileSystem:
         """
         raise NotImplementedError
 
-    def ls(self, path_info, detail=False, **kwargs):
+    def ls(self, path_info, detail=False):
         raise RemoteActionNotImplemented("ls", self.scheme)
+
+    def find(self, path_info, detail=False):
+        raise RemoteActionNotImplemented("find", self.scheme)
 
     def is_empty(self, path_info):
         return False
--- a/dvc/fs/gdrive.py
+++ b/dvc/fs/gdrive.py
@@ -532,7 +532,7 @@ class GDriveFileSystem(BaseFileSystem):
         query = f"({query}) and trashed=false"
         return self._gdrive_list(query)
 
-    def _ls_recursive(self, path_info, detail=False):
+    def find(self, path_info, detail=False):
         root_path = path_info.path
         seen_paths = set()
 
@@ -569,13 +569,7 @@ class GDriveFileSystem(BaseFileSystem):
                 else:
                     yield item_path
 
-    def ls(
-        self, path_info, detail=False, recursive=False
-    ):  # pylint: disable=arguments-differ
-        if recursive:
-            yield from self._ls_recursive(path_info, detail=detail)
-            return None
-
+    def ls(self, path_info, detail=False):
         cached = path_info.path in self._ids_cache["dirs"]
         if cached:
             dir_ids = self._ids_cache["dirs"][path_info.path]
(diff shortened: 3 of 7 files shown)


isidentical:
See the comments from the old PR: https://github.com/iterative/dvc/pull/5878

efiop (approved):
:fire:

dvc/fs/fsspec_wrapper.py
```diff
@@ -93,24 +91,29 @@ def copy(self, from_info, to_info):
...
+        files = self.fs.find(path, detail=detail)
+        if detail:
+            files = files.values()
+
+        # When calling find() on a file, it returns the same file in a list.
+        # For object-based storages, the same behavior applies to empty
+        # directories since they are represented as files. This condition
+        # checks whether we should yield an empty list (if it is an empty
+        # directory) or just yield the file itself.
+        if len(files) == 1 and files[0] == path and self.isdir(path_info):
```
efiop, line 110: would calling full `isdir` be wasteful? Or does fsspec caching compensate for that?

isidentical: No, not really. We only call the `isdir()` if it is a file and see whether the file is actually a file or just a directory. So unless we start calling `walk_files()` on normal files, then this would cost nothing. 

The last check will simply ensure that the this is indeed a directory, and would only cost stuff when `import-url`in empty directories etc. Which I guess shouldn't be much of a big deal.

dvc/fs/fsspec_wrapper.py
```diff
@@ -93,24 +91,29 @@ def copy(self, from_info, to_info):
...
             return None
 
-        yield from self._strip_buckets(self.ls(path, detail=detail), detail)
+        yield from self._strip_buckets(files, detail=detail)
 
     def walk_files(self, path_info, **kwargs):
-        for file in self.ls(path_info, recursive=True):
+        for file in self.find(path_info):
             yield path_info.replace(path=file)
```
efiop, line 117: Looks like getting rid of `walk_files` will now be quite easy ;)

isidentical: yeah, `find()` is basically `walk_files` without path_infos

dberenbaum:
Similar to https://github.com/iterative/dvc/pull/5683#issuecomment-824356255, would it be realistic to summarize the performance differences here?

isidentical:
> Similar to #5683 (comment), would it be realistic to summarize the performance differences here?

Yeah, we were talking this with @efiop today. I'll probably end up with finalizing a framework where I can automatize running these kind of stuff.

isidentical:
To sum it up, the partial status used to take 54 seconds for azure and 32 seconds for google cloud and now it takes 4 seconds for azure and 3 seconds for google cloud. Pushing 1024 new files used to take 90/75 seconds respectively and now it it is taking 40/51. So for partial status, there is an improvement over 15x and for partial status the improvement is about 1.5x-2x for fsspec-based providers. 

before:
```
=====================================azure======================================                                                    
    Story: basic data cloud                                                                                                         
        push (1024 small files) took 21.9323 seconds                                                                                
        pull (1024 small files) took 34.3765 seconds                                                                                
    Story: cloud status
        fresh status (nothing missing on the remote) took 1.9444 seconds                                                            
        status (1024 files missing on the remote) took 54.6923 seconds                                                              
        push only new files (1024 new small files / 1024 existing small files) took 89.0522 seconds                                 
=======================================gs=======================================                                                    
    Story: basic data cloud                                                                                                         
        push (1024 small files) took 29.1087 seconds
        pull (1024 small files) took 32.3727 seconds
    Story: cloud status
        fresh status (nothing missing on the remote) took 2.5915 seconds
        status (1024 files missing on the remote) took 32.7394 seconds
        push only new files (1024 new small files / 1024 existing small files) took 76.1696 seconds
```

after:
```
=====================================azure======================================                                                    
    Story: basic data cloud                                                                                                         
        push (1024 small files) took 19.1998 seconds                                                                                
        pull (1024 small files) took 33.443 seconds                                                                                 
    Story: cloud status                                                                                                             
        fresh status (nothing missing on the remote) took 1.8277 seconds                                                            
        status (1024 files missing on the remote) took 4.4486 seconds                                                               
        push only new files (1024 new small files / 1024 existing small files) took 40.8264 seconds                                 
=======================================gs=======================================
    Story: basic data cloud
        push (1024 small files) took 27.9613 seconds
        pull (1024 small files) took 30.029 seconds
    Story: cloud status
        fresh status (nothing missing on the remote) took 2.5847 seconds
        status (1024 files missing on the remote) took 3.8758 seconds
        push only new files (1024 new small files / 1024 existing small files) took 51.135 seconds
```

Just for reference (this is the normal s3, not the fsspec one so nothing changed on it);
```
=======================================s3=======================================                                                    
    Story: basic data cloud                                                                                                         
        push (1024 small files) took 40.4514 seconds                                                                                
        pull (1024 small files) took 61.8855 seconds                                                                                
    Story: cloud status
        fresh status (nothing missing on the remote) took 2.7203 seconds                                                            
        status (1024 files missing on the remote) took 5.2392 seconds                                                               
        push only new files (1024 new small files / 1024 existing small files) took 76.0406 seconds                                 
```

dberenbaum:
Will this show up in any dvc-bench tests?

isidentical:
> Will this show up in any dvc-bench tests?

No, dvc-bench tests will use s3 as the remote and they are very limited. These improvements are present in azure and google storage

dberenbaum:
Would be great if we could add those at some point 😁 . @efiop Do you think this makes sense as a separate issue?

efiop:
@dberenbaum Sure, it is on our TODO list, @isidentical will likely repurpose his scripts into dvc-bench benchmarks in the near future (there are some PRs already).  https://github.com/iterative/dvc-bench/issues/252

shcheklein:
This is a really great start @isidentical . 

We need a bit more advanced (real life) benchmarks to my mind though. I would love to see cases like - 300K on remote + I'm pushing 1 file, 10 files, 1024 files (where is the threshold here when it stars listing all the files?), 1, 10, 1024 files on remote and I'm pushing 300K. (not specifically related to this change, but who knows? that's why we have tests)- I have a directory with 300K files and add 1 file and push it again (+different size on the remote end, up to 1M files).

I would also talk to @pmrowla - he has a lot of insights on different thresholds and optimizations that we made before.

+ compare this with some baseline - s3 tools optimized to the maximum performance (basically when network becomes a limiting factor)

isidentical:
> We need a bit more advanced (real life) benchmarks to my mind though. I would love to see cases like - 300K on remote + I'm pushing 1 file, 10 files, 1024 files (where is the threshold here when it stars listing all the files?), 1, 10, 1024 files on remote and I'm pushing 300K. (not specifically related to this change, but who knows? that's why we have tests)- I have a directory with 300K files and add 1 file and push it again (+different size on the remote end, up to 1M files).

From what I can see, it might take multiple hours per run (which needs to be doubled for also running on the baseline branch and then compared).

shcheklein:
okay, I got 300K from the top of my head, primary to emphasize the scale. My point is that on 1024 files we might have some short cuts in place, or dis-balance (remote - local cache) is not enough to cover certain optimizations, etc. That's what I'm worried about.

shcheklein:
(also, we might not need even to run all these benchmarks every day, but it's good to have them to run when we do major updates, or before a major release)

isidentical:
Ah, I see. I'll try to use for the s3fs PR (this one was a net gain in any scenario, so didn't bother much). Perhaps something like 16K/32K.

dberenbaum:
How are you generating data for these benchmark tests?

isidentical:
random bytes, not a real dataset.

dberenbaum:
I mean that s5cmd performance would be like a goal that dvc could try to get as close to possible to since its a library focused solely on speed of s3 transfers. We could include it as a benchmark to compare our performance, potentially having specified reasons why it's not realistic to reach that level of performance, and track whether dvc is getting closer or further from that benchmark over time.

shcheklein:
Reminds me this one - https://github.com/iterative/dvc/issues/5871 . @tyommik has done a good research, and indeed s5cmd was 6-7x faster than the regular aws cli.

isidentical:
I guess so, though it might better if we continue on a separate issue. Perhaps #5871, since it is a bit irrelevant for `find()` (it doesn't change anything s3-wise)

PR #5880, opened 2021-04-26
refactor exp show: use new abstractions

skshetry:

One visible change here is that now baseline rows are styled as bold.

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -123,8 +123,6 @@ def _collect_rows(
     sort_by=None,
     sort_order=None,
 ):
-    from rich.text import Text
-
     from dvc.scm.git import Git
 
     if sort_by:
@@ -147,8 +145,7 @@ def _collect_rows(
                 name_rev = base_rev[:7]
             else:
                 name_rev = base_rev
-            name = exp.get("name", name_rev)
-            text = Text(name, style="bold")
+            text = exp.get("name", name_rev)
         else:
             if tip:
                 parent_rev = exp.get("checkpoint_parent", "")
@@ -176,7 +173,7 @@ def _collect_rows(
             name = exp.get("name", rev[:7])
             text = f"{tree} {queued}{name}{parent}"
 
-        row = [text, _format_time(exp.get("timestamp"))]
+        row = [text, _format_time(exp.get("timestamp")), rev == "baseline"]
         _extend_row(
             row, metric_names, exp.get("metrics", {}).items(), precision
         )
@@ -287,7 +284,7 @@ def experiments_table(
 
     from dvc.compare import TabularData
 
-    headers = ["Experiment", "Created"]
+    headers = ["Experiment", "Created", "is_baseline"]
     td = TabularData(
         lconcat(headers, metric_headers, param_headers), fill_value=FILL_VALUE
     )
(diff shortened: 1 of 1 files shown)


dvc/command/experiments.py
```diff
@@ -145,8 +148,7 @@ def _collect_rows(
             else:
                 name_rev = base_rev
             name = exp.get("name", name_rev)
-            row.append(f"{name}")
-            style = "bold"
+            text = Text(name, style="bold")
```
skshetry, line 151: @pmrowla, was this a bug of not using it?

dvc/command/experiments.py
```diff
@@ -235,51 +235,31 @@ def _sort(item):
...
-        return "-"
+        return FILL_VALUE
     if timestamp.date() == date.today():
         fmt = "%I:%M %p"
     else:
         fmt = "%b %d, %Y"
     return timestamp.strftime(fmt)
 
 
-def _format_field(val, precision=DEFAULT_PRECISION):
```
skshetry, line 246: Waiting for _format_field to be in the `dvc.compare`.

dvc/command/experiments.py
```diff
@@ -308,44 +321,57 @@ def _experiments_table(all_experiments, **kwargs):
         include_params=include_params,
         exclude_params=exclude_params,
     )
+    metric_headers = _normalize_headers(metric_names)
+    param_headers = _normalize_headers(param_names)
```
skshetry, line 325: It could have simplified a lot of things here if we were working with a flattened/normalized params and metrics data.
Working with flattened headers but with expanded/normal params and metrics might cause unintended bugs. 
Leaving it for now though.

skshetry: Using the flattened data might make this code cleaner thanks to the Table, but as-is, it can provide very little advantage.

dvc/command/experiments.py
```diff
@@ -308,44 +321,57 @@ def _experiments_table(all_experiments, **kwargs):
...
-        "Experiment", no_wrap=True, header_style="black on grey93"
-    )
-    if not kwargs.get("no_timestamp", False):
-        table.add_column("Created", header_style="black on grey93")
-    _add_data_columns(
-        table,
+    td = experiments_table(
+        all_experiments,
+        metric_headers,
         metric_names,
```
skshetry, line 330: need to pass metric headers and metric names both here, as they are both required.

dvc/command/experiments.py
skshetry: And, since experiments are more like nested tables than a simple table, the abstraction does not work well for the experiments. For example, sorting the best experiment.

dvc/command/experiments.py
pmrowla: no, this style is meant to be applied to the entire row (which is why the original code yielded a tuple of `(row_data, row_style)`

skshetry: @pmrowla, but this wasn't being used at all.

dvc/command/experiments.py
```diff
@@ -308,44 +321,57 @@ def _experiments_table(all_experiments, **kwargs):
...
-    )
-    _add_data_columns(
-        table, param_names, justify="left", header_style="black on light_cyan1"
-    )
-
-    for base_rev, experiments in all_experiments.items():
-        for row, _, in _collect_rows(
-            base_rev, experiments, metric_names, param_names, **kwargs,
-        ):
-            table.add_row(*row)
```
pmrowla, line 333: @skshetry the confusion about the row style thing is that this was originally
```
for row, style in _collect_rows(...):
    ....
    table.add_row(*row, style=style)
```
it got turned off at some point during the pre-release testing phase due to a terminal compatibility bug and was never turned back on (by mistake). It should be turned back on now.

pmrowla: see also: https://github.com/iterative/dvc/pull/5829/files#diff-150402093379363d315fa80a2af5b8576a883a197951aeb66228e2a459630ec7

dvc/command/experiments.py
pmrowla: ah I see what you meant now. The row style was intended to be used later in the calling function (see the other comment below)

pmrowla (approved):

PR #5885, opened 2021-04-27
exp show: separate ui representation of exp names and identifiers

skshetry:

It might not be apparent from the PR on why I am doing this. Basically, I am trying to separate the presentation logic with the other business logics. The indent guide seems to be a `rich-table` only thing. For example, if we have a `--show-md`, it might not make much sense to have that tree (but it can be there if we want to on a separate presentation logic). 

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -138,16 +138,16 @@ def _collect_rows(
 
     new_checkpoint = True
     for i, (rev, exp) in enumerate(experiments.items()):
-        queued = "*" if exp.get("queued", False) else ""
-
+        queued = str(exp.get("queued") or "")
         tip = exp.get("checkpoint_tip")
         parent = ""
+        tree = ""
         if rev == "baseline":
             if Git.is_sha(base_rev):
                 name_rev = base_rev[:7]
             else:
                 name_rev = base_rev
-            text = exp.get("name", name_rev)
+            name = exp.get("name", name_rev)
         else:
             if tip:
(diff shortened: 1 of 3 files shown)


dvc/command/experiments.py
```diff
@@ -305,6 +318,19 @@ def experiments_table(
     return td
 
 
+def prepare_exp_id(kwargs):
```
skshetry, line 321: This is a bit ugly though. :(

dvc/command/experiments.py
```diff
@@ -164,7 +164,7 @@ def _collect_rows(
                         tree = "├─╨"
                     else:
                         tree = "│ ╟"
-                        parent = f" ({parent_rev[:7]})"
```
skshetry, line 167: I was trying to separate the `tree` logic, but the `new_checkpoint` state made it a bit hard. We could add this value to a column and later calculate the appropriate tree, but I don't think it's worth it considering this hacky alternate way as done here.

pared (approved):

pared:
Failing tests are unrelated

pmrowla (approved):

jorgeorpinel:
Hi. There's plenty `exp show` examples in docs, esp. in https://dvc.org/doc/command-reference/exp/show

Does this change it much or is it mainly #5829 ? We'll need a doc PR or issue either way please.

pmrowla:
This doesn't change the table at all. And the only thing that #5829 actually changes UI-wise is that SHA's are displayed in the row by default for experiments.

PR #5888, opened 2021-04-27
list: add --show-json option

### LINKED ISSUE #5881: list: Add --show-json (or similar flag)
In the vs code project we have a view that uses `dvc list . --dvc-only` to show all paths that are tracked by DVC in a tree view.  Reasons for this view, some discussion around it and a short demo are shown here: https://github.com/iterative/vscode-dvc/issues/318.

At the moment we take the stdout from the command, split the string into a list (using `\n` as a delimiter) and then post process to work out whether or not the paths relate to files or directories. I can see from the output of the command that directories are already highlighted: 

image

From the above I assume that the work to determine what the path is (file or dir) has already been done by the cli. Rather than working out this information again it would be ideal if the cli could pass us json that contains the aforementioned information.

This will reduce the amount of code required in the extension and should increase performance (ever so slightly).

Please let me know if any of the above is unclear.

Thanks

pared:

Fixes #5881

### DIFF
--- a/dvc/command/ls/__init__.py
+++ b/dvc/command/ls/__init__.py
@@ -34,7 +34,11 @@ class CmdList(CmdBaseNoRepo):
                 recursive=self.args.recursive,
                 dvc_only=self.args.dvc_only,
             )
-            if entries:
+            if self.args.show_json:
+                import json
+
+                logger.info(json.dumps(entries))
+            elif entries:
                 entries = _prettify(entries, sys.stdout.isatty())
                 logger.info("\n".join(entries))
             return 0
@@ -64,6 +68,9 @@ def add_parser(subparsers, parent_parser):
     )
     list_parser.add_argument(
         "--dvc-only", action="store_true", help="Show only DVC outputs."
+    )
+    list_parser.add_argument(
+        "--show-json", action="store_true", help="Show output in JSON format."
     )
     list_parser.add_argument(
         "--rev",
--- a/tests/unit/command/ls/test_ls.py
+++ b/tests/unit/command/ls/test_ls.py
@@ -1,3 +1,6 @@
+import json
+import logging
+
 from dvc.cli import parse_args
 from dvc.command.ls import CmdList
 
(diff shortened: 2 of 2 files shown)


karajan1001 (approved):
LGTM, but the Codecov said that it lacks tests.

pared:
@karajan1001 I wrote a test

dvc/command/ls/__init__.py
```diff
@@ -34,7 +34,11 @@ def run(self):
                 recursive=self.args.recursive,
                 dvc_only=self.args.dvc_only,
             )
-            if entries:
+            if self.args.show_json:
+                import json
+
+                logger.info(json.dumps(entries))
```
mattseddon, line 40: [Q] What is the format of this output? 

Is it:
```
[{"path1": "fileOrDirectory"}, {"path2": "fileOrDirectory"}, {"path3": "fileOrDirectory"}]
```
?

pared: @mattseddon 
Here is output for `dvc list --dvc-only git@github.com:iterative/dvc.git scripts/innosetup --show-json`
```bash
[{"isout": true, "isdir": false, "isexec": false, "path": "dvc.ico"}, {"isout": true, "isdir": false, "isexec": false, "path": "dvc_left.bmp"}, {"isout": true, "isdir": false, "isexec": false, "path": "dvc_up.bmp"}]
```

pared:
Failing tests are unrelated.

dvc/command/ls/__init__.py
mattseddon: Thank you

skshetry: @pared, maybe it'd be better to have the schema as an object rather than a list?

skshetry:
> @skshetry Does the table UI cover `dvc list`?

No, it does not, as there's not much to it regarding UI.

dvc/command/ls/__init__.py
pared: @skshetry I am not sure I follow - do you mean to have a schema and validate the resulting list or return the results as some kind of an object?

skshetry: @pared, Sorry, I meant the result that we return. List might not be that useful, I think. Either a dictionary (or, a hierarchy inside a dictionary) might be easier?

pared: @skshetry it might be that dict would be better, we could easily preserve dir structure, especially for (-R). @mattseddon what do you think about that? Would that be desirable feature for you?

mattseddon: I'm quite happy with what we have at the moment but I can deal with you updating to whatever you see fit. 

However, any dict with a 'hierarchy' would be undesirable as would require more parsing after we've run the command.

PR #5889, opened 2021-04-27
introduce tabview backend for tables

skshetry:

0. Checkout to this PR.
1. `pip install tabview`.
2. `dvc exp show` or `dvc metrics show`.

Keeping it as a draft for now. It could also leave as an optional flag, but right now for demonstration, it's on by default.

I'd keep this as one of our backend for the table, but more than that, I am not sure. Let's explore this TUI area, I think as metrics flow sideways, it's the best way to provide a good user experience for metrics/diff show. 

We could further explore this with auto-reloading as experiment runs, sorting, aggregating, highlighting experiments run, and adding keyboard shortcuts and commands. But that's far in the future and iff we decide to go with this.

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -339,6 +339,8 @@ def show_experiments(
     include_params = _parse_filter_list(kwargs.pop("include_params", []))
     exclude_params = _parse_filter_list(kwargs.pop("exclude_params", []))
 
+    tabview = kwargs.get("tabview", False)
+
     metric_names, param_names = _collect_names(
         all_experiments,
         include_metrics=include_metrics,
@@ -390,7 +392,11 @@ def show_experiments(
     td.drop("is_baseline")
 
     merge_headers = ["Experiment", "queued", "ident_guide", "parent"]
-    td.column("Experiment")[:] = map(prepare_exp_id, td.as_dict(merge_headers))
+
+    if not tabview:
+        td.column("Experiment")[:] = map(
+            prepare_exp_id, td.as_dict(merge_headers)
+        )
     td.drop(*merge_headers[1:])
 
     td.render(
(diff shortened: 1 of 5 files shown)


dvc/command/experiments.py
```diff
@@ -450,6 +457,7 @@ def run(self):
                 sort_order=self.args.sort_order,
                 precision=self.args.precision or DEFAULT_PRECISION,
                 pager=not self.args.no_pager,
+                tabview=True,
```
dberenbaum, line 460: Why true here?

dvc/command/metrics.py
```diff
@@ -50,6 +50,7 @@ def run(self):
                 all_tags=self.args.all_tags,
                 all_commits=self.args.all_commits,
                 precision=self.args.precision or DEFAULT_PRECISION,
+                tabview=True,
```
dberenbaum, line 53: Why true here?

dberenbaum (commented):
I'm a bit confused when tabview should be set to true.

More importantly, is this worth the added dependency, especially considering this doesn't look like a particularly active project: https://github.com/TabViewer/tabview?

dvc/command/experiments.py
skshetry: I have written it above.

> for demonstration, it's on by default.

skshetry:
@dberenbaum, I know tabview is dead now, the point of the PR is to explore TUI-based `metrics/exp show`. There's [`visidata`](https://github.com/saulpw/visidata) if we want to get serious (albeit with not so good API) or if it's just simpler usecases, tabview works really well and is simple.

dvc/command/experiments.py
dberenbaum: Sorry, I missed the explanation above!

dberenbaum:
Can it have something like a `style` arg that can support whatever (markdown, rich, tabview, etc.)?

PR #5893, opened 2021-04-27
output: relative paths outside of repo

### LINKED ISSUE #5870: dvc add causes ERROR without saying that's the problem
# Bug Report

## Description
dvc add causes ERROR without saying that's the problem. 
We need clear explanation that is wrong

### Reproduce

(venv) (base) insaf@MacBook-Pro-Insaf test_gan % dvc add ../tabgan/Research/data/credit/credit.gz
Adding...                                                                                                                                                  
ERROR: /Users/insaf/Desktop/ml projects/tabgan/Research/data/credit/credit.gz

### Expected
Add file to dvc config file

### Environment information

**Output of `dvc doctor`:**

```console
$ dvc doctor
```
(venv) (base) insaf@MacBook-Pro-Insaf test_gan % dvc doctor
DVC version: 2.0.18 (osxpkg)
Platform: Python 3.8.9 on macOS-10.16-x86_64-i386-64bit
Supports: All remotes
Caches: local
Remotes: None
Workspace directory: apfs on /dev/disk1s1s1
Repo: dvc, git

Please check https://github.com/iterative/dvc/wiki/Debugging-DVC on ways to gather more information regarding the issue.

If the issue is regarding the performance, please attach the profiling information and the benchmark comparisons.
-->

efiop:
Fixes #5870 + improves internal exc message.

### DIFF
--- a/dvc/output/local.py
+++ b/dvc/output/local.py
@@ -48,7 +48,11 @@ class LocalOutput(BaseOutput):
         return self.FS_CLS.PATH_CLS(abs_p)
 
     def __str__(self):
-        if not self.repo or not self.is_in_repo:
+        if (
+            not self.repo
+            or urlparse(self.def_path).scheme == "remote"
+            or os.path.isabs(self.def_path)
+        ):
             return str(self.def_path)
 
         cur_dir = os.getcwd()
@@ -63,8 +67,15 @@ class LocalOutput(BaseOutput):
 
     @property
     def is_in_repo(self):
-        def_scheme = urlparse(self.def_path).scheme
-        return def_scheme != "remote" and not os.path.isabs(self.def_path)
+        if urlparse(self.def_path).scheme == "remote":
+            return False
+
+        if os.path.isabs(self.def_path):
+            return False
+
+        if self.repo and not self.path_info.isin(self.repo.root_dir):
+            return False
+        return True
 
     def dumpd(self):
         ret = super().dumpd()
--- a/dvc/scm/git/__init__.py
+++ b/dvc/scm/git/__init__.py
@@ -151,7 +151,9 @@ class Git(Base):
         gitignore = os.path.join(ignore_file_dir, self.GITIGNORE)
 
         if not path_isin(os.path.realpath(gitignore), self.root_dir):
-            raise FileNotInRepoError(path)
+            raise FileNotInRepoError(
+                f"'{path}' is outside of git repository '{self.root_dir}'"
+            )
 
         return entry, gitignore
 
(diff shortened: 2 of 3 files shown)


dvc/output/local.py
```diff
@@ -48,7 +48,11 @@ def _parse_path(self, fs, path):
...
     def __str__(self):
-        if not self.repo or not self.is_in_repo:
+        if (
+            not self.repo
+            or urlparse(self.def_path).scheme == "remote"
+            or os.path.isabs(self.def_path)
+        ):
             return str(self.def_path)
 
         cur_dir = os.getcwd()
```
efiop, line 58: We expect ([and have tests for it](https://github.com/iterative/dvc/blob/b2f71d459bab129833840c8c5517f28156ff6162/tests/unit/output/test_local.py#L31)) that `str()` of an external output with a relative path in `def_path` should be a proper `relpath` and not just a `def_path`, which no longer fits corrected `is_in_repo`.

dvc/scm/git/__init__.py
```diff
@@ -151,7 +151,9 @@ def _get_gitignore(self, path):
         gitignore = os.path.join(ignore_file_dir, self.GITIGNORE)
 
         if not path_isin(os.path.realpath(gitignore), self.root_dir):
-            raise FileNotInRepoError(path)
+            raise FileNotInRepoError(
+                f"'{path}' is outside of git repository '{self.root_dir}'"
+            )
```
efiop, line 156: Just to improve the message in case this ever falls through the cracks again.

dvc/output/local.py
karajan1001: Excuse me ,what is the difference between them? `def_path` and `relpath`, one relative to current dir another to repo root?

skshetry: @karajan1001, `def_path` is what is defined (or, to-be defined) in the `dvc.yaml` and `.dvc` files. `relpath` depends on what we are using for (based on stage's wdir or root of the repo or the current working directory).

tests/func/test_add.py
```diff
@@ -347,6 +347,25 @@ def test_add_external_file(tmp_dir, dvc, workspace, hash_name, hash_value):
     assert dvc.status() == {}
 
 
+def test_add_external_relpath(tmp_dir, dvc, local_cloud):
+    fpath = local_cloud / "file"
+    fpath.write_text("file")
```
pared, line 352:
```suggestion
     fpath, = local_cloud.gen("file", "file")
```

pared (approved):

dvc/output/local.py
karajan1001: @skshetry , so maybe `__repr__` should return `def_path` and `__str__` returns `relpath` ?

skshetry: @karajan1001, I'd expect `__repr__` to return something like `Output: {def_path}` or `Output: {resolved_path}` as it makes it easier to debug. `output.def_path` is also much better than `repr(output)`.

PR #5901, opened 2021-04-30
simplify experiments row collection logic

skshetry:

Still trying to get rid of `tree` from the `collect_rows`. I will follow up more in the upcoming PR, and also move the `[[bold]]` with the `rich.text` format as it messes up the output for other kinds of tables.

Please do review the changes correctly, I tried to have it equivalent, but might have missed something.


pmrowla (approved):
LGTM, tested it with a few checkpoint repos I have on hand and the output is consistent w/what it was before this PR

skshetry:
Thanks, @pmrowla for the review, I have one more follow-up PR to this.

PR #5903, opened 2021-04-30
exp show: decoupling tree and styling from row collection

skshetry:
Trying to remove tree and the bold/row-styling part even more

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -5,7 +5,7 @@ from datetime import date, datetime
 from fnmatch import fnmatch
 from typing import TYPE_CHECKING, Dict, Iterable, Optional
 
-from funcy import constantly, iffy, lmap
+from funcy import lmap
 
 import dvc.prompt as prompt
 from dvc.command import completion
@@ -17,6 +17,8 @@ from dvc.exceptions import DvcException, InvalidArgumentError
 from dvc.utils.flatten import flatten
 
 if TYPE_CHECKING:
+    from rich.text import Text
+
     from dvc.compare import TabularData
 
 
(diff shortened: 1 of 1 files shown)


dvc/command/experiments.py
```diff
@@ -116,6 +118,16 @@ def _collect_names(all_experiments, **kwargs):
     return metric_names, param_names
 
 
+experiment_types = {
```
skshetry, line 121: @pmrowla, let me know if you have better naming suggestions. Feels like I am artificially making one here.

dvc/command/experiments.py
```diff
@@ -321,17 +328,30 @@ def experiments_table(
...
     exp_name = kwargs["Experiment"]
-    ident = kwargs.get("ident_guide")
-    queued = kwargs.get("queued")
+    rev = kwargs["rev"]
+    typ = kwargs.get("typ", "baseline")
+
+    if typ == "baseline" or not exp_name:
+        text = Text(exp_name or rev)
+    else:
+        text = Text.assemble(rev, " [", (exp_name, "bold"), "]")
```
skshetry, line 341: Text is equivalent to `str` + styling, so it will be compatible with other tables too.

dvc/command/experiments.py
pmrowla: These names are fine w/me but maybe something like this would also work

```
    "branch_tip": "│ ╓",
    "branch_commit": "│ ╟",
    "branch_base": "├─╨", # or maybe branch_start
    "single_commit": "├──",
    "baseline": "",

skshetry: `branch` might be confusing with the git terminology. I instead added it to the non-checkpoint based commits.
Let me know what you think.

pmrowla: the thing is that experiments really are just branches, but what you have now is fine with me

skshetry: I meant that they themselves are the branches of the commit-graph/tree, and this UI representation is all about showing those branches, no?

If you are fine with naming, please do a review. Thanks.

pmrowla (approved):

PR #5904, opened 2021-04-30
metrics diff/show: rollback to rounding fractional digits with precision

skshetry:

### DIFF
--- a/dvc/compare.py
+++ b/dvc/compare.py
@@ -172,11 +172,17 @@ class TabularData(MutableSequence[Sequence["CellT"]]):
         ]
 
 
-def _format_field(val: Any, precision: int = None) -> str:
+def _normalize_float(val: float, precision: int):
+    return f"{val:.{precision}g}"
+
+
+def _format_field(
+    val: Any, precision: int = None, round_digits: bool = False
+) -> str:
     def _format(_val):
         if isinstance(_val, float) and precision:
-            fmt = f"{{:.{precision}g}}"
-            return fmt.format(_val)
+            func = round if round_digits else _normalize_float
+            return func(_val, precision)
         if isinstance(_val, Mapping):
             return {k: _format(v) for k, v in _val.items()}
         if isinstance(_val, list):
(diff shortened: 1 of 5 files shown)


dvc/compare.py
```diff
@@ -172,11 +172,17 @@ def as_dict(self, cols: Iterable[str] = None) -> Iterable[Dict[str, str]]:
...
+
+
+def _format_field(
+    val: Any, precision: int = None, round_digits: bool = False
+) -> str:
     def _format(_val):
         if isinstance(_val, float) and precision:
-            fmt = f"{{:.{precision}g}}"
-            return fmt.format(_val)
+            func = round if round_digits else _normalize_float
```
skshetry, line 184: I didn't go for  a `f` specifier for rounding because a) we used to use `round` before and, b) `f` is a fixed width formatting, so there could be trailing zeros which `round` does not have.

efiop (approved):
Thank you!

dberenbaum:
What's the background on the rollback? What's the precision behavior for `exp show` now? I thought this all came up to try to resolve a discrepancy between `metrics diff` and `exp show`?

skshetry:
> What's the background on the rollback? 

@efiop and I had some doubts about the new behaviour, so we decided to rollback to not block the release.

@dberenbaum, I'll leave this up to you now to clarify what should be the expected and correct behaviour for the `--precision` here (I know we have discussed this but would be good to have more closer look?).

Another thing that is unclear to me is, why do we have `--precision` flag? For what purpose do the users use that flag?
Or, is it us, trying to provide a way to change the default behaviour that we apply?

> What's the precision behavior for `exp show` now? 

It is not affected. It's same as before, with it affecting significant digits.

dberenbaum:
> @dberenbaum, I'll leave this up to you now to clarify what should be the expected and correct behaviour for the `--precision` here (I know we have discussed this but would be good to have more closer look?).

Looking at some examples and in Studio, it seems like `g` format is probably best here. Are there specific concerns that were raise about that format or changing formats in general?



> Another thing that is unclear to me is, why do we have `--precision` flag? For what purpose do the users use that flag?
> Or, is it us, trying to provide a way to change the default behaviour that we apply?

I would think it's primary use is to give users some flexibility to make outputs more readable without dvc having to always guess what format is best. One user suggested it could also be useful to filter metrics to only those with a difference greater than the specified precision (see https://github.com/iterative/dvc/discussions/5790), but I lean towards not adopting this interpretation for now.

Issue #5568, opened 2021-03-08
pager: Should DVC respect `PAGER` envvars?


skshetry:
I noticed that when experimenting with `experiments`, the pager were not being honoured. 

`PAGER=cat git branch` works but `PAGER=cat dvc exp show` does not.

pmrowla:
We specifically look for `DVC_PAGER` in commands that support the pager (so `dag` and `exp show` right now), but I'm not sure of the original reasoning for not checking `PAGER` when `dvc dag` was implemented

efiop:
Probably it seemed like a good idea to have a dedicated env var at that time :man_shrugging: , but we could and should definitely support `PAGER` too, just make it have a lower prioirity then `DVC_PAGER`.

iesahin:
Also, my 2c, `dvc exp show` and `dvc exp list` has `--no-pager` parameters. Other commands can have these as well.

iesahin:
And `dvc dag` should output a `\n` at the end: 

![Screen Shot 2021-03-14 at 13 13 54](https://user-images.githubusercontent.com/476310/111064809-3fe0fd80-84c7-11eb-8aba-d4a3f04e2114.png)

PR #5906, opened 2021-04-30
respect $PAGER env vars, make -FRSX default in less

### LINKED ISSUE #5793: dvc: expand on `--no-pager` option
1. ~~Make it the default for `dvc exp show`?~~ No longer needed after #5906
2. Include it in `dvc dag`?
3. Maybe put it directly in `dvc --no-pager` so it works for any commands that need it in the future.
4. Any other existing commands that need it? E.g. `list` or `stage list`

Rel. #4060

skshetry:
If $LESS is not set, -FSRX is set by default.
Though, regardless of $LESS set, dvc will always
use -RS, -R for rendering ansi escape sequences for coloring
and -S for horizontal scrolling for exp show, as otherwise
the output will be wrapped.

Closes #5793 and #5568


dvc/utils/pager.py
```diff
@@ -14,9 +14,33 @@
...
+def prepare_default_pager(
+    clear_screen: bool = False,
+    quit_if_one_screen: bool = True,
+    ansi_escapes: bool = True,
+    chop_long_lines: bool = True,
+    no_init: bool = True,
+    no_tilde: bool = False,
+) -> str:
+    args = [DEFAULT_PAGER]
+    if clear_screen:
```
shcheklein, line 30: the only question- do all versions of less have these arguments? (e.g. GNU vs BSD)?

skshetry: @shcheklein. the new flags that we have started to use with this PR is `--quit-if-one-screen/-F` and `--no-init/-X` (introduced ~2000 as per the [changelog](https://github.com/gwsw/less/blob/master/NEWS.VER)). And, git already defaults to `-FRX` so that's another point for the confidence. `-S` seems to be there forever, and the latest change was on `-c` which had a behavior change in 2005 (which we are not using now).

dvc/utils/pager.py
```diff
@@ -30,20 +54,35 @@ def find_pager():
...
+    # pylint: disable=redefined-outer-name
+    pager = os.getenv(DVC_PAGER)
+    if not pager:
+        pager = os.getenv(PAGER_ENV)
+    if not pager:
+        if os.system(f"({DEFAULT_PAGER}) 2>{os.devnull}") != 0:
+            logger.warning(
+                "Unable to find `less` in the PATH. Check out "
+                "{} for more info.".format(
+                    format_link("https://man.dvc.org/pipeline/show")
```
skshetry, line 66: Will update the link separately.

dvc/utils/pager.py
shcheklein: yep, sorry my question was not exactly clear. It was not about versions in terms of older versions vs newer, it's about different implementations (like `grep` for example has a bit different arguments on Mac and Linux). Anyways, your point about `And, git already defaults to -FRX so that's another point for the confidence` makes it look safe.

skshetry: @shcheklein, I don't think there are variants of `less` at all, the versions might be different though. I tested the one that comes in macOS (less 487 2016 version) and the latest one (less 581 2021 version).

pmrowla (approved):

skshetry:
@MergifyIO rebase

dberenbaum (approved):
LGTM! I wonder if there's a good way to test these UI features, but that's a separate discussion.

jorgeorpinel:
Hi. Does https://dvc.org/doc/command-reference/dag#paginating-the-output need a quick update after this? And maybe `--no-pager` in that ref's Options as well as in https://dvc.org/doc/command-reference/exp/show#options. Any others? Thanks

skshetry:
@jorgeorpinel, I have a patch somewhere, got lost between other stuff. I don't think we need to add anything on `--no-pager`.

jorgeorpinel:
Sure! By patch do you mean for the docs? Well even if you can summarize what the change should be in an issue there that would be great. Thanks

Issue #5886, opened 2021-04-27
move: creates inconsistency when moving tracked file into tracked folder 


mistermult:
# Bug Report

## Description
Assume that you track the whole folder `dir1` with `dir1.dvc`. You cannot add an extra file `dir1/a.txt` and track with `dir1/a.txt.dvc` (with the command `dvc add a.txt` you get an error). However, you can first create a file `b.txt` outside `dir1` and move it with the .dvc-file into  `dir1` with `dvc move`.

### Reproduce

```bash
dvc init
mkdir dir1
dvc add dir1

# Error as excected
touch dir1/a.txt
dvc add dir1/a.txt

# No error (unexpected)
touch b.txt
dvc add b.txt
dvc move b.txt
dvc move b.txt dir1

# Result
ls dir1
## a.txt  b.txt  b.txt.dvc

# After a commit we get
dvc commit
cat dir1.dvc 
#outs:
#- md5: c53e75a492b0e8ccefef15fbf24d4c22.dir
#  size: 70
#  nfiles: 3
#  path: dir1
```

### Expected
I excpeted one of the two cases:
1. One should not be able to move a tracked file into a tracked folder and get an error.
2. Only the `b.txt` should be moved to the folder `dir1`. `b.txt.dvc` should be removed (or modified if it tracks other files).

### Environment information

Sorry, dvc doctor hangs on my system. Produced on Linux with ext4.

karajan1001:
@mistermult , thanks.

> Sorry, dvc doctor hangs on my system. Produced on Linux with ext4.
 
Strange, but I had tested on the latest version, it can be reproduced. In my opinion, it is a bug.

pmrowla:
for reference, after the fix this `dvc move` operation (trying to move a tracked file into an already tracked dir) will result in an error before anything is actually modified

PR #5907, opened 2021-05-03
dvcfile/stage: check if dvcfile would be gitignored on stage creation

### LINKED ISSUE #5650: add: confusing interaction with gitignore'd subdirectories
# Bug Report

## Description

`dvc add` will happily complete when targeting subdirectories that appear transiently in a `.gitignore`. The tracked data has a corresponding `.dvc` file generated and appears in the cache correctly, but attempts to push to a remote is met with a silent failure, a `dvc remove` with a more vocal error, and most alarmingly a `dvc pull` deletes the data from the workspace (although it can still be found in the cache).

The behaviour is confusing and difficult to debug due to the fact that everything pretends to work or fails silently except for `dvc remove`, which was the only clue as to what was causing this.

### Reproduce

```console
$ git init
$ dvc init
$ dvc remote add -d ...
$ mkdir datadir/subdir
$ echo "data" > datadir/subdir/data.txt
$ echo "datadir/" > .gitignore
$ dvc add datadir/subdir/
$ dvc status
There are no data or pipelines tracked in this project yet. 
See <https://dvc.org/doc/start> to get started!
$ dvc push
Everything is up to date.
$ dvc remove datadir/subdir.dvc
ERROR: 'datadir/subdir.dvc' is git-ignored.
$ dvc pull
D       datadir/subdir/                                                                                                             
1 file deleted
```

### Expected

If tracking gitignored directories is not allowed, `dvc add` should fail with the same error as `dvc remove`.

I don't fully understand why adding gitignored subdirectories is not allowed however. Is the issue more that the generated '.dvc' file is transiently gitignored?

### Environment information

**Output of `dvc doctor`:**

```console
$ dvc doctor
DVC version: 2.0.6 (pip)
Platform: Python 3.7.3 on Linux-5.4.0-67-generic-x86_64-with-debian-bullseye-sid
Cache directory: ext4 on /dev/mapper/ubuntu--vg-root
Caches: local
Remotes: s3
Workspace directory: ext4 on /dev/mapper/ubuntu--vg-root
Repo: dvc, git
```

pmrowla:

Will close #5650

- Creating a stage (or pipeline) with a `.dvc` or `dvc.yaml` path which would end up gitignored is now explicitly prohibited

### DIFF
--- a/dvc/dvcfile.py
+++ b/dvc/dvcfile.py
@@ -13,7 +13,6 @@ from dvc.stage.exceptions import (
     StageFileBadNameError,
     StageFileDoesNotExistError,
     StageFileFormatError,
-    StageFileIgnoredError,
     StageFileIsNotDvcFileError,
 )
 from dvc.types import AnyPath
@@ -38,8 +37,12 @@ PIPELINE_LOCK = "dvc.lock"
 
 
 class FileIsGitIgnored(DvcException):
-    def __init__(self, path):
-        super().__init__(f"'{path}' is git-ignored.")
+    def __init__(self, path, pipeline_file=False):
+        super().__init__(
+            "{}'{}' is git-ignored.".format(
+                "bad DVC file name " if pipeline_file else "", path,
+            )
+        )
 
 
 class LockfileCorruptedError(DvcException):
@@ -79,7 +82,7 @@ def is_git_ignored(repo, path):
         return False
 
 
-def check_dvc_filename(repo, path):
+def check_dvcfile_path(repo, path):
     if not is_valid_filename(path):
         raise StageFileBadNameError(
             "bad DVC file name '{}'. DVC files should be named "
@@ -89,7 +92,7 @@ def check_dvc_filename(repo, path):
         )
 
     if is_git_ignored(repo, path):
-        raise StageFileIgnoredError(relpath(path))
+        raise FileIsGitIgnored(relpath(path), True)
 
 
 class FileMixin:
@@ -128,7 +131,7 @@ class FileMixin:
 
     def _verify_filename(self):
         if self.verify:
-            check_dvc_filename(self.repo, self.path)
+            check_dvcfile_path(self.repo, self.path)
 
     def _check_gitignored(self):
         if self._is_git_ignored():
@@ -193,7 +196,7 @@ class SingleStageFile(FileMixin):
 
         assert not isinstance(stage, PipelineStage)
         if self.verify:
-            check_dvc_filename(self.repo, self.path)
+            check_dvcfile_path(self.repo, self.path)
         logger.debug(
             "Saving information to '{file}'.".format(file=relpath(self.path))
         )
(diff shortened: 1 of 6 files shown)


pmrowla:
#5886 is related and will be addressed in a followup PR (the issue there is that we don't perform any of the stage name validation at all on the `dvc move` destination)

dvc/dvcfile.py
```diff
@@ -66,7 +67,19 @@ def is_lock_file(path):
...
+
+    try:
+        return isinstance(repo.fs, LocalFileSystem) and repo.scm.is_ignored(
+            path
+        )
+    except NoSCMError:
+        return False
+
+
+def check_dvc_filename(repo, path):
```
pared, line 82: Shouldn't we rename it? It is no longer about filename. Maybe `check_dvc_file` or `check_dvc_file_path`?

pared (approved):
LGTM, thought I would take a look at tests - they do not look like usual git stash failures.

dvc/stage/exceptions.py
```diff
@@ -44,6 +44,11 @@ class StageFileBadNameError(DvcException):
     pass
 
 
+class StageFileIgnoredError(StageFileBadNameError):
+    def __init__(self, path):
+        super().__init__(f"bad DVC file name '{path}' is git-ignored.")
+
+
```
skshetry, line 51: We already have `FileIsGitIgnored`, do we need this at all?

pmrowla: I added this since `FileIsGitIgnored` is specific to `dvc/dvcfile.py`, and there is other code elsewhere where we specifically catch & handle `StageFileBadNameError`

pmrowla:
> LGTM, thought I would take a look at tests - they do not look like usual git stash failures.

same tests are failing intermittently in master, it's something related to external outputs: https://github.com/iterative/dvc/runs/2492419791?check_suite_focus=true

skshetry:
> > LGTM, thought I would take a look at tests - they do not look like usual git stash failures.
> 
> same tests are failing intermittently in master, it's something related to external outputs: https://github.com/iterative/dvc/runs/2492419791?check_suite_focus=true

Probably a regression from https://github.com/iterative/dvc/pull/5893. That's why we should not be ignoring any failures.

dvc/stage/exceptions.py
skshetry: Hmm, maybe I am missing something here, but I could only find this during exception use during the checkout (where this does not matter) and during `dump` and `create_stage` where we never catch them.

Btw, `create_stage` might not be a good place to check them, we can just check this during the `dump`.

pmrowla: I'm not sure that moving the check to `dump` works for us, since the `dump` happens last in add/move (after we've already copied/moved objects into cache and linked them back into the workspace). If we are going to fail because the destination `.dvc`/`dvc.yaml`/`dvc.lock` file would be in an invalid location, that failure should happen before we actually do any other processing on the output itself.

pmrowla: changed to use `FileIsGitIgnored` in f9f45e2

PR #5911, opened 2021-05-03
fs: make constructor closer to fsspec

efiop:
Todo:

~~* [ ] get rid of fs.path_info~~(will be a followup, need to get rid of fs_root first).

Depends on: #5683 and #5812 , will rebase when those are merged.

### DIFF
--- a/dvc/checkout.py
+++ b/dvc/checkout.py
@@ -150,7 +150,14 @@ def _cache_is_copy(cache, path_info):
 
 
 def _checkout_file(
-    path_info, fs, obj, cache, force, progress_callback=None, relink=False,
+    path_info,
+    fs,
+    obj,
+    cache,
+    force,
+    progress_callback=None,
+    relink=False,
+    state=None,
 ):
     """The file is changed we need to checkout a new copy"""
     modified = False
@@ -170,7 +177,9 @@ def _checkout_file(
         _link(cache, cache_info, path_info)
         modified = True
 
-    fs.repo.state.save(path_info, fs, obj.hash_info)
+    if state:
+        state.save(path_info, fs, obj.hash_info)
+
     if progress_callback:
         progress_callback(str(path_info))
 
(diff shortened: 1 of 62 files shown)


dvc/checkout.py
```diff
@@ -296,5 +319,5 @@ def checkout(
     logger.debug("Checking out '%s' with cache '%s'.", path_info, obj)
 
     return _checkout(
-        path_info, fs, obj, cache, force, progress_callback, relink,
+        path_info, fs, obj, cache, force, progress_callback, relink, state,
```
efiop, line 322: Need to start using kwargs, this is unwieldy already.

dvc/fs/__init__.py
```diff
@@ -87,4 +87,20 @@ def get_cloud_fs(repo, **kwargs):
...
+    if "checksum_jobs" not in remote_conf:
+        checksum_jobs = repo.config["core"].get("checksum_jobs")
+        if checksum_jobs:
+            remote_conf["checksum_jobs"] = checksum_jobs
+
+    cls = get_fs_cls(remote_conf)
+
+    if isinstance(cls, GDriveFileSystem):
+        remote_conf["gdrive_credentials_tmp_dir"] = repo.tmp_dir
+
```
efiop, line 105: These config options would fit nicely into a dynamic config schema, if we had one.

dvc/state.py
```diff
@@ -72,7 +72,7 @@ def __init__(self, root_dir=None, tmp_dir=None, dvcignore=None):
         self.tmp_dir = tmp_dir
         self.root_dir = root_dir
         self.dvcignore = dvcignore
-        self.fs = LocalFileSystem(None, {"url": self.root_dir})
+        self.fs = LocalFileSystem()
```
efiop, line 75:
```suggestion
        self.fs = LocalFileSystem(url=self.root_dir)
```

PR #5953, opened 2021-05-04
move: validate new stage before move

### LINKED ISSUE #5886: move: creates inconsistency when moving tracked file into tracked folder

pmrowla:

Will close #5886 

- `dvc move` now uses `repo.stage.create` to create the new destination stage instead of just modifying the existing `stage.path` in order to do the necessary validation (check if dest overlaps with existing output, new `.dvc` file would be gitignored, etc)

### DIFF
--- a/dvc/stage/utils.py
+++ b/dvc/stage/utils.py
@@ -330,7 +330,7 @@ def validate_state(
 def validate_kwargs(single_stage: bool = False, fname: str = None, **kwargs):
     """Prepare, validate and process kwargs passed from cli"""
     cmd = kwargs.get("cmd")
-    if not cmd:
+    if not cmd and not single_stage:
         raise InvalidArgumentError("command is not specified")
 
     stage_name = kwargs.get("name")
--- a/tests/func/test_move.py
+++ b/tests/func/test_move.py
@@ -226,6 +226,11 @@ def test_move_gitignored(tmp_dir, scm, dvc):
     with pytest.raises(FileIsGitIgnored):
         dvc.move("foo", "dir")
 
+    assert (tmp_dir / "foo").read_text() == "foo"
+    assert (tmp_dir / "foo.dvc").exists()
+    assert not (tmp_dir / "dir" / "foo").exists()
+    assert not (tmp_dir / "dir" / "foo.dvc").exists()
+
 
 def test_move_output_overlap(tmp_dir, dvc):
     from dvc.exceptions import OverlappingOutputPathsError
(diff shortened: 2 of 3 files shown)


dvc/stage/utils.py
```diff
@@ -330,7 +330,7 @@ def validate_state(
 def validate_kwargs(single_stage: bool = False, fname: str = None, **kwargs):
     """Prepare, validate and process kwargs passed from cli"""
     cmd = kwargs.get("cmd")
-    if not cmd:
+    if not cmd and not single_stage:
```
pmrowla, line 333: not sure if this should be a separate `data_source` kwarg instead of using `single_stage`? @skshetry

tests/func/test_move.py
```diff
@@ -211,3 +213,34 @@ def test_should_move_to_dir_on_non_default_stage_file(tmp_dir, dvc):
...
+    assert (tmp_dir / "foo").read_text() == "foo"
+    assert (tmp_dir / "foo.dvc").exists()
+    assert not (tmp_dir / "dir" / "foo").exists()
+    assert not (tmp_dir / "dir" / "foo.dvc").exists()
+
+
+def test_move_output_overlap(tmp_dir, dvc):
+    from dvc.exceptions import OverlappingOutputPathsError
+
+    tmp_dir.dvc_gen({"foo": "foo", "dir": {"bar": "bar"}})
```
dberenbaum, line 238: Does this generate `dir.dvc` or `bar.dvc`? Should this test that `dir.dvc` does not change (and maybe that `dir/bar` still exists)?

Also, should both of these tests check that the `.gitignore` files haven't changed?

pmrowla: This generates `dir.dvc`.

Regarding not making gitignore changes on failed stage creation, we check for that in other func tests, we shouldn't need to duplicate that here.

dvc/stage/utils.py
skshetry: We still support stages in `.dvc` file, so it's not just about `data_source`. Maybe in 3.0? 😄

Issue #5781, opened 2021-04-07
reduce ci failures


skshetry:
The CI has not been stable for a few weeks now, and we don't pay enough attention to it.
I get bothered by these failures quite a lot (don't you?) and find them embarrassing. These failures waste everyone's time to retry them and debug unexpected failures. Similarly, tests failures on the master and release branch do not provide a good impression of the project.

Taking a quick look at it, it seems there are few issues that we are having:
1. Package build failures (thanks to @efiop, these have been fixed).
2. Intermittent failures, especially due to remote that are real (mostly on master).
3. Intermittent failures but repeated running does not help (eg: `test_stash_pop` is run 5 times max, but still it fails half the time, `gdrive` auth errors).
4. Test failures based on the phase of the moon (`get/import` test failures)

I know we have increased our tests from ~800 a year ago to ~2400 at the moment, and not all of them might be of high quality. But I think we should focus on making tests infrastructure sound and tests more robust, so we could write even more and _iterate_ 🚀  faster.

Let's investigate/discuss how we could improve upon this. 🙂

skshetry:
Also, few questions that come up with this:
1. Who should be responsible for the fixes?
2. How often/When should we assign someone to fix it?
3. How do we keep track of failures and failure rates?

karajan1001:
How about make a instance (container) running tests repeatedly. It can provide us
1. failure rate.
2. most frequent reasons.
3. comparison to previous version.

daviddanielarch:
Maybe there is no need to run the tests repeteadely, it might be possible to retrieve failed executions by using the Actions API.
It would be useful for future runs to generate an artifact for test runs on a programatic format like junit XML, that would make analyzing test failures easier.

For tests interacting with real APIs like https://github.com/iterative/dvc/blob/master/tests/func/test_api.py **test_open_external** which is failing on this run for example https://github.com/iterative/dvc/runs/2325104896, and that are going to be run frequently maybe its worth to use mocks instead of real connections (https://github.com/googleapis/google-api-python-client/blob/master/docs/mocks.md). 

Is it worth it to run those type of tests on every merge to master? Maybe using scheduled actions https://jasonet.co/posts/scheduled-actions/ and running those type of tests weekly makes more sense. Ideally those tests should only be run when there is a change on the library connecting to the third party provider, can you get that type of granularity using Actions?

karajan1001:
@daviddanielarch 
[...]

PR #5959, opened 2021-05-04
tests: rerun gdrive and gs on failures

skshetry:
We just need to mark them with `needs_internet` marker.
Unfortunately, it's not possible to do that automatically
with the fixtures as fixtures are per-session and we need
something on higher level (which the marker is not passed
through in the pytest).

Part of https://github.com/iterative/dvc/issues/5781

### DIFF
--- a/tests/unit/remote/test_remote_tree.py
+++ b/tests/unit/remote/test_remote_tree.py
@@ -32,6 +32,7 @@ def remote(request, dvc):
     return get_remote(dvc, **cloud.config)
 
 
+@pytest.mark.needs_internet
 @pytest.mark.parametrize("remote", remotes, indirect=True)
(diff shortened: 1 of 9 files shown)


tests/unit/remote/test_remote_tree.py
```diff
@@ -32,6 +32,7 @@ def remote(request, dvc):
     return get_remote(dvc, **cloud.config)
 
 
+@pytest.mark.needs_internet
```
skshetry, line 35: Whoever decided to run gs and s3 inside a unit test 👏🏼.

skshetry:
Note that the test failures in 3.8 and 3.9 will be fixed by #5954.

efiop (approved):

PR #5962, opened 2021-05-04
try to close dvc.state on exit

skshetry:
I have a feeling that it is causing the issues with shutil.copytree and some PermissionError during 'init', as we are not closing it between multiple 'main()' calls. But I am not quite sure, and this seems like a good thing to do anyway.

The error message could be seen in https://github.com/iterative/dvc/runs/2498067244#step:10:5463:
```console
 D:\a\dvc\dvc\tests\basic_env.py:110: UserWarning: Failed to remove test dir: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\dvc-test.6432.9i20lxfe.CKjKY6gtu6d33MzUazViLz\\.dvc\\tmp\\links\\cache.db'
5464
    warnings.warn("Failed to remove test dir: " + str(exc))
```

And, I think it might be causing [these kinds of failures](https://github.com/iterative/dvc/runs/2498067244#step:10:5042) in Windows as well.

Part of https://github.com/iterative/dvc/issues/5781.

### DIFF
--- a/dvc/command/init.py
+++ b/dvc/command/init.py
@@ -39,14 +39,14 @@ class CmdInit(CmdBaseNoRepo):
         from dvc.repo import Repo
 
         try:
-            self.repo = Repo.init(
+            with Repo.init(
                 ".",
                 no_scm=self.args.no_scm,
                 force=self.args.force,
                 subdir=self.args.subdir,
-            )
-            self.config = self.repo.config
-            _welcome_message()
+            ) as repo:
+                self.config = repo.config
+                _welcome_message()
         except InitError:
             logger.exception("failed to initiate DVC")
             return 1
--- a/dvc/info.py
+++ b/dvc/info.py
@@ -31,34 +31,32 @@ def get_dvc_info():
     ]
 
     try:
-        repo = Repo()
+        with Repo() as repo:
+            # cache_dir might not exist yet (e.g. after `dvc init`), and we
+            # can't auto-create it, as it might cause issues if the user
+            # later decides to enable shared cache mode with
+            # `dvc config cache.shared group`.
+            if os.path.exists(repo.odb.local.cache_dir):
+                info.append(
+                    "Cache types: {}".format(_get_linktype_support_info(repo))
+                )
+                fs_type = get_fs_type(repo.odb.local.cache_dir)
+                info.append(f"Cache directory: {fs_type}")
+            else:
+                info.append("Cache types: " + error_link("no-dvc-cache"))
 
-        # cache_dir might not exist yet (e.g. after `dvc init`), and we
-        # can't auto-create it, as it might cause issues if the user
-        # later decides to enable shared cache mode with
-        # `dvc config cache.shared group`.
-        if os.path.exists(repo.odb.local.cache_dir):
-            info.append(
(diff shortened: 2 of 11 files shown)


dvc/main.py
```diff
@@ -52,7 +52,7 @@ def main(argv=None):  # noqa: C901
 
         with debugtools(args):
             cmd = args.func(args)
-            ret = cmd.run()
+            ret = cmd.do_run()
```
skshetry, line 55: I do like `run` for aesthetics. Maybe I should do metaclass hacks here to keep `run`. I'm feeling lazy to change other subclasses.

skshetry (commented):
I need to monitor the CI if there are any improvements. I still see failures and warnings, but I'll take a more closer look tomorrow on what's happening with Windows.

efiop (dismissed):

skshetry:
@MergifyIO rebase

dvc/command/init.py
```diff
@@ -39,14 +39,14 @@ def run(self):
         from dvc.repo import Repo
 
         try:
-            self.repo = Repo.init(
+            with Repo.init(
```
skshetry, line 42: `init` is used in a lot of places in our tests, causing a lot of PermissionErrors  if we are not closing the cache handles.

dvc/info.py
```diff
@@ -31,34 +31,32 @@ def get_dvc_info():
...
-            fs_type = get_fs_type(repo.odb.local.cache_dir)
-            info.append(f"Cache directory: {fs_type}")
-        else:
-            info.append("Cache types: " + error_link("no-dvc-cache"))
-
-        info.append(f"Caches: {_get_caches(repo.odb)}")
-
-        info.append(f"Remotes: {_get_remotes(repo.config)}")
-
+        with Repo() as repo:
```
skshetry, line 34: Same here as above, when we are loading info, we should be closing the caches. A lot of warnings/failures were from this and the init.

tests/func/test_cli.py
```diff
@@ -60,31 +60,42 @@ def test(self):
                 arg2,
             ]
         )
-
-        self.assertIsInstance(args.func(args), CmdRun)
+        cmd_cls = args.func(args)
```
skshetry, line 63: This returns a new instance of `Repo`, so we should be closing this as well.

tests/func/test_run_single_stage.py
```diff
@@ -898,6 +898,8 @@ def setUp(self):
         super().setUp()
         ret = main(["config", "cache.type", "hardlink"])
         self.assertEqual(ret, 0)
+        self.dvc.close()
+
         self.dvc = DvcRepo(".")
```
skshetry, line 903: I did not want to waste too much time on why `self.dvc.config.load()` was not working. So I am going for close and reopen here.

skshetry:
There are two major issues left with the CI failures after this PR:
1. gdrive: This should be somewhat redued after we run it only once in the CI job.
2. Error during dvc fixture initialization: This is because of how we are copying temp directories. This needs more investigation (or, might need to be rolled back to just create each time). This is the one that is currently failing.

efiop (approved):

PR #5974, opened 2021-05-06
run gdrive tests in a single job, run all on schedule

skshetry:

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -7,6 +7,8 @@ on:
   push:
     branches:
       - master
+  schedule:
+    - cron: '5 1 * * *'  # every day at 01:05
 
 env:
   DVC_TEST: "true"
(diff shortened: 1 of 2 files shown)


efiop:
For the record: ideally, we'll soon create fsspec fs for gdrive in PyDrive2, so we could keep most of the tests there and also then add a gdrive test to dvc-bench.

This PR seems to be complicating the tests, but it is a good workaround that we c ould use for now. :+1:

skshetry:
> For the record: ideally, we'll soon create fsspec fs for gdrive in PyDrive2, so we could keep most of the tests there and also then add a gdrive test to dvc-bench.
> 
> This PR seems to be complicating the tests, but it is a good workaround that we c ould use for now.

Note that even if we do move them over to a different repo, we should not be running multiple jobs with the same credentials/token. Or, even better, we should create a new token/user every time we run the tests. It seems a lot of our issues is just because of being rate-limited.

shcheklein:
> Note that even if we do move them over to a different repo, we should not be running multiple jobs with the same credentials/token. Or, even better, we should create a new token/user every time we run the tests. It seems a lot of our issues is just because of being rate-limited.

A quick note: there is an official mechanism to expand limits. It can be expanded significantly. There is a form in the Google cloud, somewhere close to the tokens and their current usage rate.

efiop:
@shcheklein Thanks for the reminder! I've sent a request to increase the quota.

shcheklein:
:(

Maybe try to explain them that it is sporadic and they keep the per day limit the same, but increase the one per minute and/or per second?

efiop:
@shcheklein Btw, do we have exponential backoff implemented?

shcheklein:
@efiop yep, it's implemented.

skshetry:
> @shcheklein Btw, do you remember if we have exponential backoff implemented?

Back off might not help for the rate-limit. We also run gdrive tests 5 times if it fails. But it still fails.

shcheklein:
@skshetry so, it depends. It has multiple limits - per day, per hour, per second, etc. If we run out of the "per day" one - then, yes we are out of luck (in the failure message it should have a link to the dashboard where you could see what is happening).

skshetry:
@shcheklein, here's the last failure:
https://github.com/iterative/dvc/runs/2509376211?check_suite_focus=true#step:10:5256

skshetry:
@shcheklein, is it possible to see if we have crossed the `per 100 seconds per user` anytime of the day? It's very much possible that 2 pipelines (2 * 12 CI jobs) sometimes come in sync, that'd be just 416 API calls per job. 

The last screenshot says that it's a 3hr average.

shcheklein:
Good catch (3h average). It might be indeed.

Do we randomize runs? (it's a good practices anyway, and in this case it can reduce this this overlap, and we can see the result pretty quickly).

skshetry:
I ran the CI jobs 3 times. It failed on 2 jobs in the first one, 4 jobs in the second and 3 in the third one. 

We never seem to have crossed 2500 queries per 100 seconds (see the curve on the right side).
Screenshot from 2021-05-08 11-24-08


There seem to be some hidden limits though as we are getting high error rate (it's almost as equal as success rates during peak and about half on average.

Screenshot from 2021-05-08 11-29-15

shcheklein:
Can we contact them, ask to clarify what's happening here, or access query logs?

PR #5975, opened 2021-05-06
disable caching the test directories

skshetry:
https://github.com/iterative/dvc/runs/2498067244#step:10:5042

^ Trying to see if not caching those test directories gets rid of those errors.
I'll try running them a few times to see what happens.
 
### DIFF
--- a/tests/dir_helpers.py
+++ b/tests/dir_helpers.py
@@ -266,18 +266,10 @@ def make_tmp_dir(tmp_path_factory, request, worker_id):
     def make(name, *, scm=False, dvc=False, subdir=False):
         from dvc.repo import Repo
         from dvc.scm.git import Git
-        from dvc.utils.fs import fs_copy
 
-        cache = CACHE.get((scm, dvc, subdir))
-        if not cache:
-            cache = tmp_path_factory.mktemp("dvc-test-cache" + worker_id)
-            TmpDir(cache).init(scm=scm, dvc=dvc, subdir=subdir)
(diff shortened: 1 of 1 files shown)


tests/dir_helpers.py
```diff
@@ -266,18 +266,10 @@ def make_tmp_dir(tmp_path_factory, request, worker_id):
...
-        if not cache:
-            cache = tmp_path_factory.mktemp("dvc-test-cache" + worker_id)
-            TmpDir(cache).init(scm=scm, dvc=dvc, subdir=subdir)
-            CACHE[(scm, dvc, subdir)] = os.fspath(cache)
         path = tmp_path_factory.mktemp(name) if isinstance(name, str) else name
-        for entry in os.listdir(cache):
-            # shutil.copytree's dirs_exist_ok is only available in >=3.8
-            fs_copy(os.path.join(cache, entry), os.path.join(path, entry))
         new_dir = TmpDir(path)
+        new_dir.init(scm=scm, dvc=dvc, subdir=subdir)
```
skshetry, line 272: Tests become slower on Windows though.

skshetry:
Note that the test failures are unrelated, and will be fixed by https://github.com/iterative/dvc/pull/5962.

tests/dir_helpers.py
efiop: Also will probably start seeing more problems with gitpython again :(

I see that it is sqlite-related files that are causing those errors and there have been requests to move sqlite to /tmp so that we would be able to work on nfs and such, which will probably solve the original issue as well.

PR #5976, opened 2021-05-06
plots: handle NoRemoteError for dirs

efiop:
Studio ran into it for `dvclive` directory. CC @Suor 

https://sentry.io/organizations/iterative/issues/2382799173/?project=5220519&referrer=slack

### DIFF
--- a/dvc/repo/plots/__init__.py
+++ b/dvc/repo/plots/__init__.py
@@ -51,6 +51,7 @@ class Plots:
             }}}
         Data parsing is postponed, since it's affected by props.
         """
+        from dvc.config import NoRemoteError
         from dvc.fs.repo import RepoFileSystem
         from dvc.utils.collections import ensure_list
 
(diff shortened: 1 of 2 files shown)


dvc/repo/plots/__init__.py
```diff
@@ -71,10 +72,21 @@ def collect(
...
                     plot_files = []
-                    for pi in fs.walk_files(path_info):
-                        plot_files.append(
-                            (pi, relpath(pi, self.repo.root_dir))
+                    try:
+                        for pi in fs.walk_files(path_info):
+                            plot_files.append(
+                                (pi, relpath(pi, self.repo.root_dir))
+                            )
+                    except NoRemoteError:
```
efiop, line 80: Same as we handle it in `dvc diff` https://github.com/iterative/dvc/blob/77be01ee2d8bed167928e64ab6d29c6be87554d0/dvc/repo/diff.py#L168

efiop: Though alternative these days could be to use `onerror` too, but in this case we know that this dir is dvc-tracked, so this error could only be risen for the dir itself.

pared (approved):

dvc/repo/plots/__init__.py
shcheklein: (just trying to understand this better)

> Also NoRemoteError could be replaced with some standard errno

yep 👍 - it can potentially raise other things here? like remote is not available, or remote exists, but broken, right? 

not sure if we handle this well on the Studio side cc @Suor ... I mean the long tail of all other possible errors here ?

seems related to what @pared has been working on also (to not fail on errors on specific revisions), looks like it should be all part of that solution, wdyt guys?

PR #5984, opened 2021-05-07
[WIP] plots/metrics/params/experiments show: stop throwing exceptions

pared:

Related: #5525 #5667 #5845


pared:
In this change I try to provide a new way of handling errors. 
Lets take a look at sample repositry, with broken `dvc.yaml` on workspace and one of the experiments commits breaking `out.yaml` metric.

Providing test code for reproduction puproses:
```python
@pytest.fixture
def custom_exp_stage(tmp_dir, scm, dvc):
    dump_yaml(tmp_dir / "params.yaml", {"foo": 111, "bar": 222})
    dump_yaml(tmp_dir / "other.yaml", {"a": 333, "b": 999})
    CODE = dedent(
        """
        from ruamel.yaml import YAML
        ruyaml=YAML()
        with open('params.yaml', 'r') as fd:
          data=ruyaml.load(fd)
        with open('other.yaml', 'r') as fd:
          other=ruyaml.load(fd)
        
        def to_metric(d):
          res = {}
          for n,v in d.items():
            res[n + "_met"] = v/10
          return res

        with open('out.yaml', 'w') as fd:
          ruyaml.dump(to_metric(data), fd)
        with open('other_out.yaml', 'w') as fd:
          ruyaml.dump(to_metric(other), fd)"""
    )
    tmp_dir.gen("code.py", CODE)
    scm.add(["params.yaml", "other.yaml", "code.py"])
    scm.commit("init")

    stage = dvc.run(
        name="train",
        params=["foo,bar", "other.yaml:a,b"],
        metrics_no_cache=["out.yaml", "other_out.yaml"],
        cmd="python code.py",
        deps=["code.py"],
    )
    scm.add(
        ["dvc.yaml", "dvc.lock", ".gitignore", "out.yaml", "other_out.yaml"]
    )
    scm.commit("initial run done")
    return stage


def test_show_broken(tmp_dir, scm, dvc, custom_exp_stage):
    init_rev = scm.get_rev()
    dvc.experiments.run(
        custom_exp_stage.addressing, params=["foo=2"], name="good_experiment"
    )

    scm.gitpython.git.reset(init_rev, hard=True)
    BREAK_METRIC_CODE = dedent(
        """
    with open("out.yaml", "a") as fd:
        fd.write("break the yaml!")
    """
    )

    with open("code.py", "a") as fd:
        fd.write(BREAK_METRIC_CODE)

    dvc.experiments.run(custom_exp_stage.addressing, name="break_metric")
    scm.gitpython.git.reset(init_rev, hard=True)
    dvc.checkout()

    with open("dvc.yaml", "a") as fd:
        fd.write("break the yaml")

    assert main(["exp", "show", "--no-pager"]) == 0
```

In current master `DVC` version (lets assume `2.1.0`) Upon running `dvc exp show` we will get:

```bash
ERROR: failed to show experiments - unable to read: 'dvc.yaml', YAML file structure is corrupted: while scanning a simple key
  in "<unicode string>", line 17, column 1
could not find expected ':'
  in "<unicode string>", line 18, column 1
```

The result for this change:

```bash
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┓
┃ Experiment                    ┃ Created  ┃ other_out.yaml:a_met ┃ other_out.yaml:b_met ┃ out.yaml:foo_met ┃ out.yaml:bar_met ┃ params.yaml:foo ┃ params.yaml:bar ┃ other.yaml:a ┃ other.yaml:b ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━┩
│ workspace                     │ -        │ ERR_BRANCH           │ ERR_BRANCH           │ ERR_BRANCH       │ ERR_BRANCH       │ 111             │ 222             │ ERR_BRANCH   │ ERR_BRANCH   │
│ master                        │ 06:11 PM │ 33.3                 │ 99.9                 │ 11.1             │ 22.2             │ 111             │ 222             │ 333          │ 999          │
│ ├── bbfde7f [good_experiment] │ 06:11 PM │ 33.3                 │ 99.9                 │ 0.2              │ 22.2             │ 2               │ 222             │ 333          │ 999          │
│ └── a6b1a41 [break_metric]    │ 06:11 PM │ 33.3                 │ 99.9                 │ ERR_FILE         │ ERR_FILE         │ 111             │ 222             │ 333          │ 999          │
└───────────────────────────────┴──────────┴──────────────────────┴──────────────────────┴──────────────────┴──────────────────┴─────────────────┴─────────────────┴──────────────┴──────────────┘
```

skshetry:
Hi, @pared. Could you please provide me the issue tickets and other contexts when and why we started to suppress those error messages? I have been trying to understand the issue on generalizing this thing as a guideline, on why it's okay to suppress exceptions for some commands and not on others.

But I am having issues understanding why we don't print errors to the terminal saying that the "view" or "table" might not be "complete" because of such and such reasons and proceed with the rendering not-so complete table and then exit with 1 error code. This is not related to this PR, but the status quo.

Btw, have you looked into https://github.com/iterative/dvc/pull/5038? I feel like tolerating errors is a more general issue than the metrics/params/plots/experiments, and we need a general framework to do this.

Regarding the metrics/params/plots/experiments, with https://github.com/iterative/dvc/pull/5324, these commands started fetching files from remote in case it didn't exist before. You might have noticed that remote errors are making these commands fail now as we are only catching `DvcException` whereas the remotes don't have a standardized exceptions.
So this makes me think that we won't ever be able to catch every errors and even we do, it might not be a good solution.

Take an example of a recently cloned dvc repo with no cache but remote configured with credentials that are invalid. We'd have just failed before, because there's no way to proceed further. But now with this PR, we'd just be hammering the remote trying to find all the files in the remote. So, I don't feel good about catch-all here in the PR.

But I also don't think catching a set of exceptions is the answer. Earlier, `metrics show` was not showing all information because it was raising `OutputIsIgnoredError`, which I'd have never thought to whitelist to that set. This is because we have too many places where the errors can be raised (and, RepoFS is too complicated). 

So, I'd prefer we spend some time on having a general framework around this and try to harmonize exceptions. Regarding UI, I'd prefer we print error messages to `stderr` and then render the table, and maybe return the exit code.

pared:
> Could you please provide me the issue tickets and other contexts when and why we started to suppress those error messages? I have been trying to understand the issue on generalizing this thing as a guideline, on why it's okay to suppress exceptions for some commands and not on others.

There is none, the discussion regarding this behaviour is conducted under linked issues, from first message in this PR.
The guideline would be a good idea, i guess we need to fromalize this and inspect whether other commands could benefit from such behaviour. Lets start a discussion on that?

> But I am having issues understanding why we don't print errors to the terminal saying that the "view" or "table" might not be "complete" because of such and such reasons and proceed with the rendering not-so complete table and then exit with 1 error code. This is not related to this PR, but the status quo.

This might be my reluctance to use `logger` speaking, but we used to have a lot of issues which point was that logs generated during command execution were occluding the command result. After first use with failed revision, `WARNING` or `INFO` is informative. User checks what is wrong, understands. Yet, repo history does not get revritten, so now every time the user uses `dvc * show/diff`, the warning is still there. And if user runs the command yet again (with some other failed revision) there just no results for failed revs. Current apporach gives granularity, telling the user whether particular file and revision failed to be obtained or whether the file does not exists. I don't say that `ERR_FILE` and `ERR_REV` is proper naming, but it lets one quickly understand what is going on.

>Btw, have you looked into #5038? I feel like tolerating errors is a more general issue than the metrics/params/plots/experiments, and we need a general framework to do this.

I guess its related to first question. 

> Regarding the metrics/params/plots/experiments, with #5324, these commands started fetching files from remote in case it didn't exist before. You might have noticed that remote errors are making these commands fail now as we are only catching DvcException whereas the remotes don't have a standardized exceptions.
So this makes me think that we won't ever be able to catch every errors and even we do, it might not be a good solution.

Yes, in current master state we are not able to catch them all. This change however makes handling those errors possible. The `intercep_error` context manager was created in order to handle unexpected errors on revision. We know that in `exp/params/metrics/plots` commands we have 2 "levels" of potential failure: `revision` (remote fetch failed, wrongly formatted `dvc.yaml`) and `path` (wrongly saved metric, for example, or not saved at all). So we can intercept most of the errors (probably beside brancher ones). As we don't know what is the error, we have to assume that we  might be able to recover on different branch. It might not be the best solution but the alternative is implementing handling for all types of errors, which seems to bother users. They have to create issue because whole command fails, while they just wanted to check the metrics progress (#5845)

> Take an example of a recently cloned dvc repo with no cache but remote configured with credentials that are invalid. We'd have just failed before, because there's no way to proceed further. But now with this PR, we'd just be hammering the remote trying to find all the files in the remote. So, I don't feel good about catch-all here in the PR.

Yes this is unfortunate, but this could be handled later. Instead of erroring on every exception, suppress all and implement special handling for the ones we know how to handle. In this case we could implement onerror that raises exception when there is authorization problem, and ignores others. 

> So, I'd prefer we spend some time on having a general framework around this and try to harmonize exceptions. Regarding UI, I'd prefer we print error messages to stderr and then render the table, and maybe return the exit code.

Lets start the discussion then?

My main point in this change is that i believe we need a shift in how we handle errors for let's name it `read only` commands. They are safe, they will not break the repo, and their biggest value is providing even smallest piece of information about repository. In other places, we catch only particular errors, because its safer to break the execution than let it continue after unknown exception. Here it is different - unknown exception will not harm the repo, so its better to supress and do as much as possible.

dberenbaum:
@skshetry What should the table look like in your mind? Should the `a6b1a41` line be missing? Should that line exist but the `ERR_FILE` cells are empty?

> I'd prefer we print error messages to `stderr` and then render the table, and maybe return the exit code

If we are still going to continue and possibly encounter multiple errors, how do we decide which exit code to show?

@pared Is there a way to see the individual error messages?

pared:
@dberenbaum yes, they are put into log on debug level. I still consider putting them into trace, but no matter what we decide, it is still possible to see them.

As to exit code, it seems to me that `status` with failed revisions should still be 0. Other should be reserved to unexpected errors. But this is another subject for a more general discussion how do we handle excepitons in such cases.

skshetry:
> @skshetry What should the table look like in your mind? Should the `a6b1a41` line be missing? Should that line exist but the `ERR_FILE` cells are empty?
> 

I am not convinced that we need to indicate in the table itself, but I am not opposed to it.


> If we are still going to continue and possibly encounter multiple errors

I am not saying that we print all the logs. I feel that there should be at least a message printed in stderr indicating that the table is not complete to the user. 

> how do we decide which exit code to show?
If we decide to fail in all cases, I'd say we return 1 if it is minor failures and 2 if it is serious ([similar to `ls`](http://www.gnu.org/software/coreutils/manual/html_node/ls-invocation.html)).
If we decide not to fail in all cases, we could return 1 if it's serious/unhandled failures, else it'll be 0.

pared:
> If we decide to fail in all cases, I'd say we return 1 if it is minor failures and 2 if it is serious

This looks like a valid solution for this case. I guess we would need to analyze other commands whether they also need such granularity.

pared:
@skshetry I started a discussion regarding brancher: #6039

#6039 will deal with how to approach `onerror` and `brancher`.

With the problem how we should inform user about some problems, we should probably deal here.

So, during the demo i proposed 3 solutions:

1. Input error data into the table
```bash
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┓
┃ Experiment                    ┃ Created  ┃ other_out.yaml:a_met ┃ other_out.yaml:b_met ┃ out.yaml:foo_met ┃ out.yaml:bar_met ┃ params.yaml:foo ┃ params.yaml:bar ┃ other.yaml:a ┃ other.yaml:b ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━┩
│ workspace                     │ -        │ ERR_BRANCH           │ ERR_BRANCH           │ ERR_BRANCH       │ ERR_BRANCH       │ 111             │ 222             │ ERR_BRANCH   │ ERR_BRANCH   │
│ master                        │ 05:36 PM │ 33.3                 │ 99.9                 │ 11.1             │ 22.2             │ 111             │ 222             │ 333          │ 999          │
│ ├── f36320f [break_metric]    │ 05:36 PM │ 33.3                 │ 99.9                 │ ERR_FILE         │ ERR_FILE         │ 111             │ 222             │ 333          │ 999          │
│ └── 3b77c91 [good_experiment] │ 05:36 PM │ 33.3                 │ 99.9                 │ 0.2              │ 22.2             │ 2               │ 222             │ 333          │ 999          │
└───────────────────────────────┴──────────┴──────────────────────┴──────────────────────┴──────────────────┴──────────────────┴─────────────────┴─────────────────┴──────────────┴──────────────┘

```

2. Log warning when some errors occured
```bash
WARNING: Loading files for some revisions failed (`workspace, f36320f4f9f65d3a856ba621274b05e602fe12d7`) Use `-v` to get more info.
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┓
┃ Experiment                    ┃ Created  ┃ other_out.yaml:a_met ┃ other_out.yaml:b_met ┃ out.yaml:foo_met ┃ out.yaml:bar_met ┃ params.yaml:foo ┃ params.yaml:bar ┃ other.yaml:a ┃ other.yaml:b ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━┩
│ workspace                     │ -        │ !                    │ !                    │ !                │ !                │ 111             │ 222             │ !            │ !            │
│ master                        │ 05:36 PM │ 33.3                 │ 99.9                 │ 11.1             │ 22.2             │ 111             │ 222             │ 333          │ 999          │
│ ├── 3b77c91 [good_experiment] │ 05:36 PM │ 33.3                 │ 99.9                 │ 0.2              │ 22.2             │ 2               │ 222             │ 333          │ 999          │
│ └── f36320f [break_metric]    │ 05:36 PM │ 33.3                 │ 99.9                 │ !                │ !                │ 111             │ 222             │ 333          │ 999          │
└───────────────────────────────┴──────────┴──────────────────────┴──────────────────────┴──────────────────┴──────────────────┴─────────────────┴─────────────────┴──────────────┴──────────────┘
```

3. Mark failed files with (for example) `!` and provide some context with, for example a legend under the table:
```bash
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┓
┃ Experiment                    ┃ Created  ┃ other_out.yaml:a_met ┃ other_out.yaml:b_met ┃ out.yaml:foo_met ┃ out.yaml:bar_met ┃ params.yaml:foo ┃ params.yaml:bar ┃ other.yaml:a ┃ other.yaml:b ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━┩
│ workspace                     │ -        │ !                    │ !                    │ !                │ !                │ 111             │ 222             │ !            │ !            │
│ master                        │ 05:36 PM │ 33.3                 │ 99.9                 │ 11.1             │ 22.2             │ 111             │ 222             │ 333          │ 999          │
│ ├── 3b77c91 [good_experiment] │ 05:36 PM │ 33.3                 │ 99.9                 │ 0.2              │ 22.2             │ 2               │ 222             │ 333          │ 999          │
│ └── f36320f [break_metric]    │ 05:36 PM │ 33.3                 │ 99.9                 │ !                │ !                │ 111             │ 222             │ 333          │ 999          │
└───────────────────────────────┴──────────┴──────────────────────┴──────────────────────┴──────────────────┴──────────────────┴─────────────────┴─────────────────┴──────────────┴──────────────┘
! means some problem occured. Use `-v` to get more info.
```

I think we should agree which way we will go and keep it consistient for `params/metrics/plots diff` too.
My own preference would be 3.
1. cannot contain too much information due to lack of space
2. we have already been here - ususally users get frustrated with warnings that they are aware of and cannot disable

3. Seems to me like reasonable compromise. Though I am not sure its possible to implement it in pager. In the example i used `--no-pager`.

@skshetry @pmrowla  what do you think?

pmrowla:
With 3, your issue with the pager is just that currently we can only paginate the table itself in `ui`, right?

@skshetry It seems like our pager handling should be a level above that, so that we can paginate any UI output at all (and not just a single table). Then for 3 we would just write the table and then the error information and it would all show up in the pager.

This would also allow us to do things like writing a second (optional?) table that could potentially contain rows of commits we failed to parse and the full string error message for each row

pared:
@skshetry 

But when would we be displaying the error messages? Before showing the table or after? I like @pmrowla's idea for creating second table with errors, because it puts the errors into background - we keep the ability to show whatever works, and there is a way to show errors.

> We should not be overwhelming users with error messages.

The question here is how do we differentiate what is the important error. I think this is a good idea, though this assumes we create logic reacting to different errors. But what do we do with unknown errors?

> And, I'd prefer we don't print Use -v for more info.

Well, if one is aware that `-v` he/she will probably use it anyway, so we can just remove it.

dvc/command/plots.py
```diff
@@ -34,47 +37,78 @@ def run(self):
...
+                opened = webbrowser.open(rel)
+                if not opened:
+                    logger.error(
+                        "Failed to open. Please try opening it manually."
+                    )
+                    return_value = 1
+        else:
+            from dvc.ui import ui
+
+            ui.warn(
```
pared, line 83: We used to create file with no actual plots. After this change we will not create it if there are no plots.

pared:
Migth stil need to reiterate on type checks, though the implementation should be finished.

Some remarks:

1. For displaying failed metrics/params in experiments, we are showing only `!` for now, as it needs solving on its own (see discussion above)

2. We no longer produce empty plots file if no plots has been gathered during collection/rendering.

3. Introduced new class `Onerror` which is gathering data on errors encountered during processing our data. This class has `errors` dict that is supposed to store info on failed revisions and particular paths. If the whole revision failed (eg malformed `dvc.yaml`, `errors['{rev}']` will contain the exception raised. If only particular file failed, `{rev}.{path}` path will contain the error.

shcheklein:
@pared good stuff! For the sake of the VS Code team, could you please summarize how does it affect `--show-json`? do we see these granular errors there now and how?

pared:
@shcheklein In current state there is no change in `--show-json` - I think we need to discuss how can we support retriveing the errors too.

shcheklein:
@pared could you please coordinate with @mattseddon and/or @rogermparent on this? They have more information on how do we process data from the JSON output on the VS Code end.

pared:
Leaving PR as draft until we discuss what to do with `--show-json`

skshetry:
Users piping it to `jq` or similar tools may not usually care about the errors in the `json` itself. Also, users may not be running with `-o pipefail` set, so I think we should still print to stderr. In case of vscode, maybe we could introduce `--with-errors` or similar flag.

pared:
@skshetry you mean to replace logging to debug with logging to error?

rogermparent:
My gut feeling tends toward a solution similar to @skshetry's suggestion of a `--with-errors` option, but specifically geared toward machine-readable errors either by encoding the message with an error type in predictable JSON or simply associating each error type with a stable error code and outputting the error code alone, leaving it up to the consumer to display a relevant message (which can also be localized, particularly in vscode with its whole `.nls` thing).

Currently, the vscode extension handles errors by simply dumping `stderr`'s to a notification if a command fails, but that output is a little noisy in that context due to how the docs link interacts with VSCode stripping formatting notifications.
It's difficult to programmatically detect and react to specific errors it is without some sort of error code or other machine-oriented output- we've held off on it until now, and while we can work around the current error system with regex or something I think dvc CLI should make error handling as easy as possible to encourage more rich integrations.

dberenbaum:
As a reference, here's some guidelines from clig.dev:

> Catch errors and rewrite them for humans. If you’re expecting an error to happen, catch it and rewrite the error message to be useful. Think of it like a conversation, where the user has done something wrong and the program is guiding them in the right direction. Example: “Can’t write to file.txt. You might need to make it writable by running ‘chmod +w file.txt’.”
>
> Signal-to-noise ratio is crucial. The more irrelevant output you produce, the longer it’s going to take the user to figure out what they did wrong. If your program produces multiple errors of the same type, consider grouping them under a single explanatory header instead of printing many similar-looking lines.

tests/func/plots/test_show.py
```diff
@@ -395,24 +392,6 @@ def test_plot_cache_missing(tmp_dir, scm, dvc, caplog, run_copy_metrics):
     ]
 
 
-def test_throw_on_no_metric_at_all(tmp_dir, scm, dvc, caplog):
```
pared, line 398: Removing, its no longer relevant after removing throws.

pared:
After talking with @dberenbaum we decided to create differences how do we handle different use cases that are covered here.
I used following script:
```bash
#!/bin/bash

set -ex

pushd $TMPDIR

rm -rf test_workspace
mkdir test_workspace

pushd test_workspace

mkdir test_repo
pushd test_repo

echo -e "foo: 2\nbar: 3" > params.yaml
echo -e "a: 2\nb: 3" > other.yaml
echo -e "
from ruamel.yaml import YAML
import json

ruyaml=YAML()
with open('params.yaml', 'r') as fd:
  data=ruyaml.load(fd)

def to_metric(ds):
  res = {}
  for d in ds:
    for metric,val in d.items():
      res[metric + '_met']=val/10
  return res

with open('other.yaml', 'r') as fd:
  other=ruyaml.load(fd)


with open('output.yaml', 'w') as fd:
  ruyaml.dump(to_metric([data]), fd)
with open('other_out.yaml', 'w') as fd:
  ruyaml.dump(other, fd)

def to_plot(params):
  for key, val in params.items():
    plot_data = [{'m': params} for i in range(5)]
    with open(key+'_plot.json', 'w') as fobj:
      json.dump(plot_data, fobj)

to_plot(data)
to_plot(other)

" >> code.py

git init --quiet
dvc init --quiet

git add -A
git commit -am "init"

dvc run -p foo,bar -p other.yaml:a,b -d code.py -m output.yaml -m other_out.yaml --plots foo_plot.json --plots bar_plot.json --plots a_plot.json --plots b_plot.json -n train "python code.py"
git add -A
git commit -am "first run"
git tag v1

echo -e "foo: 4\nbar: 5" > params.yaml
echo -e "a: 1\nb: 7" > other.yaml

dvc repro train
git add -A
git commit -am "second run"
git tag v2
```

I created different edge use cases, by simply `echo MALFORMED >> dvc.yaml` in malformed `dvc.yaml` use cases, and `echo MALFORMED >> output.yaml` whend making (in this case) `output.yaml` metric malformed.

1. [`params diff` when `dvc.yaml` is malformed](https://asciinema.org/a/419545)
2. [`params diff` when params file is malformed](https://asciinema.org/a/419560)
3. [`metrics show` when `dvc.yaml` is malformed](https://asciinema.org/a/420380)
4. [`metrics show` when metrics file is malformed](https://asciinema.org/a/420383)
5. [`metrics diff` when `dvc.yaml` is malformed](https://asciinema.org/a/420381)
6. [`metrics diff` when metrics file is malformed](https://asciinema.org/a/420384)
7. [`plots show` when `dvc.yaml` is malformed](https://asciinema.org/a/420391)
8. [`plots show` when plot file is malformed](https://asciinema.org/a/420394)
9. [`plots diff` when `dvc.yaml` is malformed](https://asciinema.org/a/420392)
10. [`plots diff` when plot file is malformed](https://asciinema.org/a/420393)

dvc/repo/experiments/show.py
```diff
@@ -91,8 +111,9 @@ def show(
     sha_only=False,
     num=1,
     param_deps=False,
+    onerror: Optional[Callable] = None,
```
pmrowla, line 114: Is there a reason why we are defining our error handler at the command level instead of doing it within `repo.experiments.show`/`repo.metrics.show`?

At least for the show methods (not sure about diff yet), it seems to me like our API would be cleaner if we handled the errors with the `repo...show()` methods rather than the CLI commands. Since we already return a dictionary with revs as keys why not just collect the errors inside the `show()` methods and when some rev fails, just set that revision to something like

```py
{
  "workspace": {...},  # normal rev
  "abc123": {"errors": ["failed parsing foo.yaml"]},
  "def456": {...},  # normal rev
  ...
}
```

At the CLI level, we can collect revs from that dict that have non-empty `errors` and log them as needed with the `Command failed for files in the following revs...` message.

This would also be a straightforward way of passing error messages into VScode, since `--show-json` just dumps the returned dictionary from `metrics.show()`/`experiments.show()`

dvc/command/experiments.py
```diff
@@ -194,10 +215,19 @@ def _collect_rows(
             _format_time(exp.get("timestamp")),
             parent,
         ]
+        metric_names_filled = _fill_empty_values(
+            error_rev, metric_names, onerror=onerror
+        )
+        param_names_filled = _fill_empty_values(
+            error_rev, param_names, onerror=onerror
+        )
```
pmrowla, line 223: If we move the error handling into `experiments.show()` we could get rid of these helpers and simplify things here. (see my other comment)

W/the moved error handling, we can just pass the a fill character into `_extend_row` like
```
_extend_row(..., fill_value="!" if exp.get("errors") else "-")
```

---

I get that these helpers are trying to display as much information as possible for everything that we were able to successfully parse for a particular rev, but it seems like an unnecessary level of complexity to me.

(Without the helpers we will still display whatever metrics/params we were able to successfully collect, the only difference is that `-` or `!` will be used in every other empty column depending on whether or not any error occurred for that rev, rather than having `!` in specific columns)

dvc/repo/experiments/show.py
pared: This approach is possible, however I see some things to consider before:
1. In this case we would change how API works, before there was never `errors` in the results.
2. We would need to design how to convert Errors into JSON, as `msg` string is not enough for VScode team, which would like to handle some kinds of errors. So the error jsons would have to contain some information.
In some cases the whole revisions will fail (malformed `dvc.yaml`) and in other particular path (malformed metric file). How do we differentiate between those and let VScode handle the problems at most granular level possible?
Do we include `path,revision` fields in error json objects? Or should we expect that result json can contain the error at any level, like for  example:
```
{
  "workspace": {"path1": {"error":{"message":"File not found", "code": 404}}}, 
  "abc123": {"error": {"message":"failed parsing dvc.yaml", "code":123}},
  "def456": {...},  # normal rev
  ...
}
```

I would personally be leaning towards possible `error` field on any level. What do you think @pmrowla?
@rogermparent @mattseddon would this approach be easier/better for you to handle than `--with-errors` we've been talking about?

pmrowla: Regarding 1, we are already going to have to change the API in some way to pass errors to VSCode. I also don't think this is a problem because for anyone who was actually consuming `--show-json` output before, the commands were unusable entirely if an error occurred (since DVC would throw an exception and not return any json output at all).

On 2, what level of granularity we want to use for passing errors probably depends on the specific command/call. For `exp show` I don't think it makes sense to do anything more granular then per experiment rev. But for metrics/params it may be better to do it per file.

dvc/command/plots.py
skshetry: @pared, we haven't yet decided on colors yet, so we are using `ui.error_write` in other places. It does look good, so maybe that's okay, but just wanted to let you know.

dvc/repo/experiments/show.py
skshetry: I think it's okay to change the internal API if it's easier that way. 

Though I'd still prefer that we do not return errors in the `--show-json` without the user explicitly asking for it. We should try to serve more users than the VSCode, and generally, the users don't care about the errors. For VSCode, it could use a private API or a different one easily.

skshetry (approved):
Looks good to me, @pared. 🔥

dberenbaum:
@pared Looks great!

Why does https://asciinema.org/a/420383 have an empty revision in the warning message at the end?

What do these look like for `exp show/diff`? Maybe you can summarize? Or at most screenshots.

dvc/repo/experiments/show.py
dberenbaum: I'm not sure whether users want the errors in json. If they are parsing json, they might have similar needs to vscode (unless we end up with special error messages for vscode). Having both options might be good.

It might be just as important or more important to consider what would be best for a public API. Again, having both options seems nice as long as it's easy for users. It might even be good to default to `onerror = OnError` and make is so users can just pass `onerror = None` to exclude them (rather than have to import and pass `OnError` to get the errors).

rogermparent: > @pared: We would need to design how to convert Errors into JSON, as `msg` string is not enough for VScode team, which would like to handle some kinds of errors. So the error jsons would have to contain some information.

I think this is a consideration not just for VSCode, but any application consuming these errors programmatically. Many DVC error messages point to specific files or ids, such that the full error cannot completely be converted to something like an error code without some accompanying context. We _could_ parse these from strings, but specifically for machine-reading that seems like bad DX all around.

> would this approach be easier/better for you to handle than `--with-errors` we've been talking about?

I think both approaches shake out to be roughly even in difficulty, since we already have to preprocess all experiments/checkpoints  to get aggregate data like columns.

My personal opinion is that consistently using a separate error field containing error objects containing a unique code and all necessary context would be the most clean, flexible, and low-maintenance way to handle machine-readable errors- I would imagine something like that will be needed anyway for errors that aren't tied to a specific piece of data, so my preference is to be lazily consistent and use that necessary API for all errors instead of having two levels of errors that are consumed slightly differently.

pmrowla: > rather than have to import and pass `OnError` to get the errors

Users shouldn't have to import DVC's `OnError` class at all. What we should be documenting is just the callback format (what happens in `OnError.__call__`) so that users can handle errors however they want.

pared:
> @pared Looks great!
> 
> Why does https://asciinema.org/a/420383 have an empty revision in the warning message at the end?
> 
> What do these look like for `exp show/diff`? Maybe you can summarize? Or at most screenshots.

I think that in the beggining we had a concept that "current workspace" name is '' (literally nothing), and at some point someone started (it might have been me) naming it `workspace`. Since then we have this concept divergence which is not that easy to fix. But it will probably be fixed when we handle #6039

dvc/repo/experiments/show.py
dberenbaum: The important thing is that however errors end up being communicated in `--show-json`, it should be easy for users to get the same output through the API.

dvc/utils/__init__.py
```diff
@@ -480,3 +481,48 @@ def glob_targets(targets, glob=True, recursive=True):
...
+def intercept_error(onerror: Callable = None, **kwargs):
+    try:
+        yield
+    except Exception as e:  # pylint: disable=W0703
+        logger.debug("", exc_info=True)
+        if onerror is not None:
+            onerror(e, **kwargs)
+
+
+class Onerror:
```
pmrowla, line 496: Would also like to mention that I think the naming convention here is a bit confusing.

In our methods where we want to pass the error handling callback, we define the param as `onerror=None`, where `onerror` should be a callback function that accepts 2 args (exception instance and revision) and one kwarg (optional str path). (This callback format isn't mentioned anywhere, but should probably be mentioned in the `Repo.func()` level docstrings)

This class is also named `Onerror`, but what it does is handle collecting any errors passed into it (and they get passed via the callback method which is defined in `__call__`). I think my main issue with the naming is that it should be something along the lines of `ErrorCollector`. 

The way it's named now, it looks like users using the `Repo` API are supposed to subclass `Onerror`, and pass those subclassed objects into `onerror=...`, when all they really need to do is implement a callback method. (Because not everyone will want to collect errors, in some cases I may just want to check for a specific exception type to `pass`, and re-raise everything else.)

dvc/utils/__init__.py
```diff
@@ -480,3 +481,48 @@ def glob_targets(targets, glob=True, recursive=True):
...
+    def __init__(self):
+        self.errors = {}
+
+    def __call__(self, exception: Exception, revision: str, path: str = None):
```
pmrowla, line 500: Also, this is just my personal preference, but I'm not a big fan of using `__call__` in this way since I think it makes it less obvious what this is (a handler for the `onerror` callback method). 

I get that this:

```
class Onerror():
    def __call__(self, ...):
        ...

onerror = Onerror()
Repo.func(onerror=onerror)
```

is slightly less verbose than

```
class ErrorCollector():
    def onerror(self, ...):
        ...

errors = ErrorCollector()
Repo.func(onerror=errors.onerror)
```

but I think the 2nd way is more obvious about what it is doing

dvc/repo/experiments/show.py
dberenbaum: > consistently using a separate error field

I agree that having errors separate from the successful results seems best. If I use the API to retrieve experiment results and visualize them, having to handle errors at every possible level first seems painful.

dvc/repo/metrics/show.py
```diff
@@ -102,33 +99,21 @@ def show(
...
-
-    if not res:
-        if metrics_found:
-            raise NoMetricsParsedError("metrics")
-        elif targets:
-            raise MetricDoesNotExistError(targets)
-        else:
-            raise NoMetricsFoundError("metrics", "-m/-M")
+        res[rev] = error_handler(_gather_metrics)(
+            repo, targets, rev, recursive, onerror=onerror
```
pared, line 115: It would probably be easier to read if we would pass `onerror` directly to `error_handler`, but since we need that down the line (to handle errors araising during particular file parsing) I decided to left it in `kwargs`

dvc/repo/plots/__init__.py
```diff
@@ -62,91 +60,90 @@ def collect(
...
+                    plot_files.append((pi, relpath(pi, self.repo.root_dir)))
+            else:
+                plot_files = [
+                    (path_info, relpath(path_info, self.repo.root_dir))
+                ]
+
+            props = props or {}
+
+            @error_handler
+            def read(fs, path, props=None, rev_props=None, **kwargs):
```
pared, line 101: Moved reading plots files here. (Before it was in `to_datapoints`). We are moving parsing here to catch potential parsing error here, and not in `render` level. We could probably get rid of `to_datapoints` method at this point, but I am leaving it as is to not complicate this PR further.  Logic behind `to_datapoints` is strictly related to file structure (eg `csv` is list of datapoints while `json` can be a dict that needs to be explored for data) so it still makes sense to relate data transformation to file extension (which is what `to_datapoints` do now), but it seems to me that we will need to change the approach how we deal with plots data.
Previously it was:
1. read datafiles from revisions
2. parse, transform and render (in `render` method)

Now it is:
1. read and parse datafiles from revsions
2.transform and render

How it will need to be, when we start introducing images support:
1. read and parse (if applicable)
2. transform (if applicable)
3. render

tests/unit/test_compare.py
```diff
@@ -445,44 +490,54 @@ def test_metrics_show_mocked(mocker, markdown):
...
         """\
-        | Path   | diff   | new   | old   |
-        |--------|--------|-------|-------|
-        | x.b    | -      | 6     | 5     |
-        | a.d.e  | 1      | 4     | 3     |
-        | a.b.c  | 1      | 2     | 1     |
+        | Revision   | Path         | a   | b.ad   | b.bc   | c   |
+        |------------|--------------|-----|--------|--------|-----|
+        | branch_1   | metrics.json | -   | 1      | 2      | 4   |
+        | branch_2   | metrics.json | 1   | 3      | 4      | -   |
```
pared, line 540: Original PR:
https://github.com/iterative/dvc/pull/5161
The test is passing and from "behavioral" point of view is correct, it was checking how `diff` command results behave when put into `show_metrics`. Modified the test so that test data matches usual output fed to `show_metrics`.
I had to fix it here, since I am changing `metrics.show` results schema, and those tests would no longer be passing.

dvc/repo/plots/__init__.py
```diff
@@ -273,11 +243,27 @@ def _collect_plots(
...
+    rev_props = rev_props or {}
+    _, extension = os.path.splitext(path)
+    if extension in (".tsv", ".csv"):
+        header = {**rev_props, **props}.get("header", True)
+        if extension == ".csv":
+            return _load_sv(path=path, fs=fs, delimiter=",", header=header)
+        return _load_sv(path=path, fs=fs, delimiter="\t", header=header)
+    if extension in LOADERS or extension in (".yml", ".yaml"):
+        return LOADERS[extension](path=path, fs=fs)
+    raise PlotMetricTypeError(path)
```
pared, line 264: Moved reading plots files here. (Before it was in to_datapoints). We are moving parsing here to catch potential parsing error here, and not in render level. We could probably get rid of to_datapoints method at this point, but I am leaving it as is to not complicate this PR further. Logic behind `to_datapoints` is strictly related to file structure (eg csv is list of datapoints while json can be a dict that needs to be explored for data) so it still makes sense to relate data transformation to file extension (which is what `to_datapoints` do now), but it seems to me that we will need to change the approach how we deal with plots data.

Previously it was:
1. read datafiles from revisions
2. parse, transform and render (in render method)

Now it is:
1. read and parse datafiles from revsions
2. transform and render

How it will need to be, when we start introducing images support:
1. read and parse (if applicable)
2. transform (if applicable)
3. render

pared:
I still need to reimplement behaviour for edge cases introduced in [my previous comment](https://github.com/iterative/dvc/pull/5984#issuecomment-861433436). But I believe the new version is ready for review.

What changed?
Current approach to deal with errors in `metrics/params/experiments/plots` commands assumes introducing intermediate "layers" for the results on `revision` and `path` level. Original result data is stored under `data` key. Newly introduced `onerror`, by default, will be appending potential error under `error` key.

Example: 
- metrics:
was:

```
{
    "workspace": {
        "metrics.yaml": {
            "foo": 1
        }
    },
    "branch": {
        "metrics.yaml": {
            "foo": 2
        }
    }
}
```
is:
```
{
    "workspace": {
        "data": {
            "metrics.yaml": {
                "data": {
                    "foo": 1
                }
            }
        }
    },
    "branch": {
        "data": {
            "metrics.yaml": {
                "data": {
                    "foo": 2
                }
            }
        }
    }
}
```

In case of error:

```
{
    "branch": {
        "data": {
            "metrics.yaml": {
                "error": YAMLFileCorruptedError - instance
            }
        }
    },
    "master": {
        "data": {
            "metrics.yaml": {
                "data": {
                    "foo": 1
                }
            }
        }
    }
}
```
Question: in case of error, do we want to have empty `data` field? It seems to me that we should, so that API will be consistient.

rogermparent:
> What changed?
> Current approach to deal with errors in `metrics/params/experiments/plots` commands assumes introducing intermediate "layers" for the results on `revision` and `path` level. Original result data is stored under `data` key. Newly introduced `onerror`, by default, will be appending potential error under `error` key.
> 
> Example:

Looks usable on the Node/VSCode end!

> Question: in case of error, do we want to have empty `data` field? It seems to me that we should, so that API will be consistient.

There's not much of a difference between omitting the field or setting it to something falsy like `null`, but an empty object could make it slightly more difficult to distinguish between partial and no data. It's pretty trivial to handle either way, so no need to hesitate if we want to value consistency over slight QoL for empty checks.

pared:
@rogermparent 
I omitted one thing: serializing the errors for `--show-json`. I will introduce it today.
Would 
something like:
`{"error": {"type": "YAMLFileCorruptedError", "msg": "unable to read: 'metrics.yaml', YAML file structure is corrupted"}}` suffice for now?

rogermparent:
> Would something like:
> `{"error": {"type": "YAMLFileCorruptedError", "msg": "unable to read: 'metrics.yaml', YAML file structure is corrupted"}}` suffice for now?

Absolutely, I'd even say it would suffice for a while even if the ideal is fully JSON. If we can limit our string-based inference to pieces of data on some individual types instead of having to find the type itself through string inference, that should really ease things.

tests/func/metrics/test_show.py
```diff
@@ -173,24 +192,33 @@ def test_show_no_repo(tmp_dir):
...
 
-def test_metrics_show_no_target(tmp_dir, dvc):
-    with pytest.raises(MetricDoesNotExistError):
-        dvc.metrics.show(targets=["metrics.json"])
+    assert (
+        "'metrics.json' was not found in current workspace." in caplog.messages
+    )
 
 
 def test_show_no_metrics_files(tmp_dir, dvc, caplog):
```
dberenbaum, line 220: What is the behavior now if there are no metrics?

pmrowla:
Looks a lot cleaner and is easier to follow now vs the previous class based implementation 👍

tests/func/metrics/test_show.py
pared: @dberenbaum nothing, we will get no outuput. I am still working to bring back warning in case of exceptions and no results, as I did in [previous approach](https://github.com/iterative/dvc/pull/5984#issuecomment-861433436).

pared: warnings should be working as in the comment

dberenbaum: Can/should we assert that the warning is raised?

pared: @dberenbaum I guess this is a matter of discussion, whether we should assert such things as warnings. But I did it anyway, mostly because if we don't do that we have to check it manually.

Tests verifying the behavior are named `test_log_errors` and are created for all of `metrics/params/plots`. I did not create this mechansim for experiments, since the errors will be indicated in the table anyway.

pared:
@Suor @rogermparent Will you need some heads up before merging this?  It seems we might be done here soon.

dvc/repo/metrics/show.py
pared: Passing the `onerror` seems a little bit clumsy, maybe we should revisit context flags change? 
https://github.com/iterative/dvc/pull/2957

Suor:
@pared I guess will handle that while updating dvc.

dberenbaum (approved):

pmrowla (approved):

skshetry (approved):
Looks good @pared. I have some concerns about "data"/"error" being nested (like an onion? 🧅), but I also don't see how it could be kept in one place. I have inlined a few comments but look good overall. 👍🏼

dvc/command/metrics.py
```diff
@@ -33,17 +34,10 @@ def run(self):
...
         else:
             from dvc.compare import show_metrics
 
-            # When `metrics` contains a `None` key, it means that some files
-            # specified as `targets` in `repo.metrics.show` didn't contain any
-            # metrics.
-            missing = metrics.pop(None, None)
-            if missing:
-                raise BadMetricError(missing)
-
```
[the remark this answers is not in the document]
pared: Yes, I had 2 resons for removing that:
1. I was unable to retrigger this behaviour.
2. New approach assumes we do not throw exceptions but rather gather them. Lack of metrics will be represented as empty files content.

pared:
@Suor @rogermparent 

The change is on the master now.

dvc/command/metrics.py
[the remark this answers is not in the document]
dberenbaum: Should/can we at least provide a warning or something that some metrics are missing?

PR #6000, opened 2021-05-12
fs: get rid of fs_root

efiop:
`fs_root` no longer makes sense after decoupling `fs` and `dvcignore` and needs to go for fsspec migration.

### DIFF
--- a/dvc/ignore.py
+++ b/dvc/ignore.py
@@ -52,10 +52,10 @@ class DvcIgnorePatterns(DvcIgnore):
         ]
 
     @classmethod
-    def from_files(cls, ignore_file_path, fs):
+    def from_files(cls, ignore_file_path, fs, root):
         assert os.path.isabs(ignore_file_path)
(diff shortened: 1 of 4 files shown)


dvc/ignore.py
```diff
@@ -201,7 +201,7 @@ def _update_trie(self, dirname: str, trie: PathStringTrie) -> None:
         ignore_file_path = os.path.join(dirname, DvcIgnore.DVCIGNORE_FILE)
         if not matches and self.fs.exists(ignore_file_path):
             new_pattern = DvcIgnorePatterns.from_files(
-                ignore_file_path, self.fs
+                ignore_file_path, self.fs, self.root_dir,
```
efiop, line 204: Not always true for subrepos. Looking into it...

efiop: For the record: had a discussion with @karajan1001 . Agreed to keep it as is and stabilize it in https://github.com/iterative/dvc/issues/6001

PR #6002, opened 2021-05-13
dvcignore: refactor from_file(s)

efiop:
Slight adjustments to make it cleaner. Followup for #6000 

### DIFF
--- a/dvc/ignore.py
+++ b/dvc/ignore.py
@@ -52,16 +52,12 @@ class DvcIgnorePatterns(DvcIgnore):
         ]
 
     @classmethod
-    def from_files(cls, ignore_file_path, fs, root):
-        assert os.path.isabs(ignore_file_path)
-        dirname = os.path.normpath(os.path.dirname(ignore_file_path))
-        ignore_file_rel_path = os.path.relpath(ignore_file_path, root)
-        with fs.open(ignore_file_path, encoding="utf-8") as fobj:
+    def from_file(cls, path, fs, name):
+        assert os.path.isabs(path)
+        dirname = os.path.normpath(os.path.dirname(path))
+        with fs.open(path, encoding="utf-8") as fobj:
             path_spec_lines = [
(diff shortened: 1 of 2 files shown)


dvc/ignore.py
```diff
@@ -52,16 +52,12 @@ def __init__(self, pattern_list, dirname):
         ]
 
     @classmethod
-    def from_files(cls, ignore_file_path, fs, root):
-        assert os.path.isabs(ignore_file_path)
-        dirname = os.path.normpath(os.path.dirname(ignore_file_path))
-        ignore_file_rel_path = os.path.relpath(ignore_file_path, root)
-        with fs.open(ignore_file_path, encoding="utf-8") as fobj:
+    def from_file(cls, path, fs, name):
```
efiop, line 55: The name was misleading as it only expected one file.

efiop: Also `ignore_file_path` is unnecessarily long, as in this context it is very clear that we are talking about the dvcignore file path.

dvc/ignore.py
```diff
@@ -52,16 +52,12 @@ def __init__(self, pattern_list, dirname):
         ]
 
     @classmethod
-    def from_files(cls, ignore_file_path, fs, root):
-        assert os.path.isabs(ignore_file_path)
-        dirname = os.path.normpath(os.path.dirname(ignore_file_path))
-        ignore_file_rel_path = os.path.relpath(ignore_file_path, root)
```
efiop, line 58: This was really just a name for information purposes, so we could just pass it down here instead of using `fs.fs_root` or `root`, which previously gave an impression of being used here for some functional purposes.

karajan1001 (approved):
Next time we can get rid of `self.fs` is IgnoreFilter. It is only used in `_update` and can be replace by the fs arguments from `walk` and `is_ignored`

pared (approved):

PR #6004, opened 2021-05-13
ls: use ui.write

skshetry:

### DIFF
--- a/dvc/command/ls/__init__.py
+++ b/dvc/command/ls/__init__.py
@@ -1,11 +1,11 @@
 import argparse
 import logging
-import sys
 
 from dvc.command import completion
 from dvc.command.base import CmdBaseNoRepo, append_doc_link
 from dvc.command.ls.ls_colors import LsColors
 from dvc.exceptions import DvcException
+from dvc.ui import ui
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


dvc/command/ls/__init__.py
```diff
@@ -37,10 +37,10 @@ def run(self):
             if self.args.show_json:
                 import json
 
-                logger.info(json.dumps(entries))
+                ui.write(json.dumps(entries))
             elif entries:
-                entries = _prettify(entries, sys.stdout.isatty())
-                logger.info("\n".join(entries))
+                entries = _prettify(entries, with_color=True)
```
skshetry, line 42: We should always be printing in colors and we should let `colorama` to handle appropriate usage. It might help us with `--color-always` and `--no-color` flags (if we introduced it).

tests/func/test_ls.py
```diff
@@ -73,36 +73,6 @@ def test_ls_repo(tmp_dir, dvc, scm):
     )
 
 
-def test_ls_repo_with_color(tmp_dir, dvc, scm, mocker, monkeypatch, caplog):
```
skshetry, line 76: This test is splitted into two and moved to unit.

dvc/command/ls/__init__.py
skshetry: I have been thinking of moving to rich more and more to not deal with colors, pagers, etc. But,it's a bit slow, especially for `ls`. For 27000 files, `ui.write` is 45X  faster (55ms vs 2.5s).

Issue #5988, opened 2021-05-10
commit: granular commit poisons the cache


skshetry:
# Bug Report

It seems that `Output.commit()` uses the filter_info provided by the `collect_granular`, but it does not consider the existing hash_info to merge the records, so we are polluting/poisoning the cache with the invalid entries based on that filter_info instead of `os.path.join(path_info, filter_info)`. ​

https://github.com/iterative/dvc/blob/674693bb86a3df135405c4eedc0b65943d01ae2a/dvc/output/base.py#L325-L340

Also, this function only does the "saving to odb" part, the staging part is also done in `stage.save` which does not consider the filter_info at all, so we are still staging everything to the cache (and we are using that to dump in the stage).

https://github.com/iterative/dvc/blob/674693bb86a3df135405c4eedc0b65943d01ae2a/dvc/repo/commit.py#L62

### Script to reproduce
```bash
cd "$(mktemp -d)"
git init && dvc init
mkdir -p data/{foo,bar}
echo foo-1 > data/foo/1.txt

dvc add data
find .dvc/cache -name "*.dir"
# there should be just one file

echo bar-1 > data/bar/1.txt
echo foo-2 > data/foo/2.txt

dvc commit data/bar
find .dvc/cache -name "*.dir"
# there should be 3 files: 
# one is the old one
# the other one is a valid one from hash info in data.dvc
# the other one is with an invalid entries (look at the relpath)
cat data.dvc

dvc remote add -d local $(mktemp -d)
dvc push
```

How severe this behaviour depends on how sacred we consider the cache to be.
But it also stages everything from that stage, which is definitely a big issue. This might cause two issues:
- errors in pull/push as the new files do not yet exist. Only the .dir file exists.
- the staged files might not be what the user want, so we are staging garbage as well.

---

Anyway, to fix the issue, we should only selectively stage, then save those filtered files and we should merge it with older hash_info and dump.

PR #6008, opened 2021-05-14
output/remote: get rid of NamedCache

pmrowla:

prerequisite for #829 

- Updates our used cache collection and data cloud methods to just use collections of `HashFile/Tree` objects.
- `status -c` only displays a single revision suffix for path names instead of every possible one (see https://github.com/iterative/dvc/pull/6008#issuecomment-851943247)
- Does not modify any other cloud status/push/pull behavior
    - eventually we can just use the tree objects directly instead of the multiple dicts/sets of hashes for directory+file statuses that are currently used, but that will come in a later PR

TODO in followup PRs:

- refactor external deps into behaving like objects (instead of the current repopair behavior)
- migrate cloud status into ODB
- unify remote upload/download and local cache save into single odb.save


dvc/output/base.py
```diff
@@ -669,18 +636,23 @@ def get_used_cache(self, **kwargs):
...
+        if not obj:
+            obj = self.odb.get(self.hash_info)
 
-        ret.add_child_cache(
-            self.hash_info.value, self.collect_used_dir_cache(**kwargs),
-        )
+        return {obj}
 
-        return ret
+    def get_used_external(self, **kwargs) -> Dict["RepoPair", str]:
```
pmrowla, line 650: seems like we probably want to split the idea of "used objects" and "used external deps" all the way up our call stack, but for now its only done in `output` (with the stage/repo level `used_cache` functions returning tuples of `used_objects, used_external_deps`)

dvc/remote/base.py
```diff
@@ -218,9 +217,10 @@ def make_names(hash_, names):
 
         return dir_status, file_status, dir_contents
 
-    def _indexed_dir_hashes(self, cache, named_cache, dir_md5s):
+    def _indexed_dir_hashes(self, dir_objs):
         # Validate our index by verifying all indexed .dir hashes
         # still exist on the remote
+        dir_md5s = set(dir_objs.keys())
```
pmrowla, line 223: all of these `dir_md5`/`file_md5` sets/dicts should be dropped in favor of just using the tree objects, but I don't want to touch this functionality yet in this PR

dvc/objects/tree.py
```diff
@@ -115,24 +119,45 @@ def load(cls, odb, hash_info):
...
+        self, odb: "ObjectDB", prefix: Tuple[str], copy: bool = False
+    ) -> Optional[HashFile]:
+        """Return filter object(s) for this tree.
+
+        If copy is True, returned object will be a Tree containing
+        filtered entries, but with hash_info copied from the original tree.
+
+        If copy is False, returned object will be a raw HashFile or Tree with
+        newly computed hash_info for the filtered object.
+        """
```
pmrowla, line 138: The current behavior for `filter` is to return a new tree containing the filtered set of trie entries, and a newly computed digest for the entire tree. These filtered trees are used when we do granular stage/output operations (and are related to https://github.com/iterative/dvc/issues/5988)

For collecting used dir cache, what we really want is a filtered version of `self` which contains the original `self.hash_info`, but with the filtered trie entries.

The use of `copy` is kind of ugly right now, but I'm also not sure that we really need 2 different filter methods?

pmrowla:
Still need to look into how this affects UI for push/pull/status

pmrowla:
@efiop so what's in this PR can be reviewed but I don't think we should merge it yet, since what's in the PR now basically makes `status -c` unusable since we no longer track the NamedCache hash -> pathname mappings

ex:
```
$ dvc status -A -c
        deleted:            258622b1688250cb619f3c9ccaefb7eb
        deleted:            c157a79031e1c40f85931829bc5fc552
        deleted:            d3b07384d113edec49eaa6238ad5ff00
```

pmrowla:
As discussed, what `status -c` for objects should display is something along the lines of
```
dvc status -A -c
        new: 2 objects - 12345 byte(s)
        deleted:  1 object - 12345 byte(s)
        missing: ...
```

In the actual returned data (and `--show-json` output) for cloud status, we would have a dict where the status is mapped to a list of object ids (with sizes).

---

I'm going to open a follow up PR (rebased on top of this one) that will move cloud `status` into the ODB, and will handle the `status -c` UI changes. And this PR shouldn't be considered mergeable until the both PRs are ready, so that `status -c` stays usable in master and any releases that are made in the meantime

efiop:
For the record: had a discussion and decided to try to not change the UI aspect in this PR, possibly by adding filename to the obj for now.

pmrowla:
As discussed, collected used objects now have a `HashFile.name` field which will be populated w/the output's relpath. The old brancher suffixes have still been removed from the `name` field.

dvc/repo/__init__.py
```diff
@@ -370,9 +371,17 @@ def used_cache(
...
 
-        cache = NamedCache()
+        def _add_suffix(objs, suffix):
+            from dvc.objects.tree import Tree
+
+            for obj in objs:
+                obj.name += suffix
+                if isinstance(obj, Tree):
+                    for _, entry_obj in obj:
+                        entry_obj.name += suffix
```
pmrowla, line 384: This workaround is needed to preserve current `status -c` and `repo.status()` behavior.

Assuming that we want to preserve the existing `status -c` behavior going forward, it seems like what we will is for `repo.status()` to handle returning  the mapping of objects or object IDs to the set of revision+path pairs that point to each object, similar to what NamedCache did before.

But it should be handled at the repo level rather than passing string suffixes all the way down to the output `collect_used_cache` methods like we were doing before.

dvc/objects/file.py
```diff
@@ -1,17 +1,30 @@
...
+    from dvc.path_info import PathInfo
+
 logger = logging.getLogger(__name__)
 
 
 class HashFile:
-    def __init__(self, path_info, fs, hash_info):
+    def __init__(
+        self,
+        path_info: Optional["PathInfo"],
```
efiop, line 19: from dvc/types.py
```suggestion
        path_info: Optional["DvcPath"],
```

dvc/output.py
```diff
@@ -788,18 +793,14 @@ def unprotect(self):
             self.odb.unprotect(self.path_info)
 
     def get_dir_cache(self, **kwargs):
-
         if not self.is_dir_checksum:
             raise DvcException("cannot get dir cache for file checksum")
 
         try:
-            objects.check(self.odb, self.odb.get(self.hash_info))
+            obj = self.odb.get(self.hash_info)
```
efiop, line 800: Let's move it out of `try`, just to avoid making the `except` confusing because of `obj`  being used in both.

dvc/objects/db/base.py
```diff
@@ -316,14 +317,22 @@ def _remove_unpacked_dir(self, hash_):
         pass
 
     def gc(self, used, jobs=None):
+        used_hashes = set()
+        for obj in used:
+            used_hashes.add(obj.hash_info.value)
+            if isinstance(obj, Tree):
+                used_hashes.update(
+                    entry_obj.hash_info.value for _, entry_obj in obj
+                )
```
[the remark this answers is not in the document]
efiop: @skshetry That one will no longer be necessery after the next PR, as push/pull will no longer work with raw md5 strings, but rather through odb.add, which go through objects one-by-one.

dvc/remote/base.py
```diff
@@ -142,8 +143,20 @@ def _status(
         {dir_hash: set(file_hash, ...)} which can be used to map
         a .dir file to its file contents.
         """
-        logger.debug(f"Preparing to collect status from {self.odb.path_info}")
-        md5s = set(named_cache.scheme_keys(cache.fs.scheme))
+
+        logger.debug(
+            f"Preparing to collect status from {self.fs.odb.path_info}"
```
efiop, line 148:
```suggestion
            f"Preparing to collect status from {self.odb.path_info}"
```

efiop:
@pmrowla Let's address the review in the followup, please. Merging so you won't have to rebase it for `path_info` removal. Thank you! :pray:

dvc/dependency/repo.py
```diff
@@ -1,10 +1,17 @@
+from typing import NamedTuple, Optional
+
 from voluptuous import Required
 
 from dvc.path_info import PathInfo
 
 from .base import Dependency
 
 
+class RepoPair(NamedTuple):
```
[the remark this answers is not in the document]
pmrowla: We already were already use `RepoDependency.repo_pair` so it seemed to me that we should just have `RepoPair` as the named type. Either way, this is likely going away entirely in the future since we will move to handling external deps as a special object type (so RepoDependency will have an overridden `get_obj()` method, and then will be collected the same way as other objects)

PR #6013, opened 2021-05-15
check-ignore: raise exceptions for path in subrepo

karajan1001:
fix#6001
follow the behavior of Git


dvc/ignore.py
```diff
@@ -244,7 +261,7 @@ def _update_sub_repo(self, path, ignore_trie: PathStringTrie):
             return
 
         root, dname = os.path.split(path)
-        pattern_info = PatternInfo(f"/{dname}/", f"in sub_repo:{dname}")
+        pattern_info = PatternInfo(f"/{dname}/", SubrepoException(dname))
```
efiop, line 264: This is quite odd to include an exception like that. Maybe there is some nicer way to handle this case? Like, do we even subrepo trie anymore?

From the outside perspective (e.g. from `dvc/command/check_ignore.py`) we could very quickly and easilly check if path is a subrepo, by using something like:

```
repo = Repo(dirpath) # will walk up the tree from dirpath in search of .dvc dir
if self.repo.root_dir != repo.root_dir:
    # we are in subrepo
```

unless we have a nicer way of doing it. E.g. maybe have a `SubrepoPatternInfo`(or some flag in PatterInfo) so that in you could either raise in `_ignore_details` like you do right now or, maybe, just check for that on CLI level.

karajan1001: @efiop , I raise Exception here is because the return value of `git check-ignore`. In Git:
- return 0 when the file is matched by some pattern (both include or ignored )
- return 1 when the file is not matched by any pattern.
- return 128 when the file is out of repo ( both in parent path or in a subrepo )

But the 0 and 1 here are counterintuitive, many had asked why not set 0 for missed or matches by include pattern and 1 only for ignored one.

efiop: @karajan1001 I mean that it is odd to include an actual exception instance as a `file_info` in `PatternInfo`.

karajan1001:
Example show

image

performance analysis ( 

| tests| previous version | after this PR| diff |
|:---:|:---:|:---:|:---:|
| add.Add.time_cats_dogs copy| 34.9±0.4s |  36.8±0.3s | +5.4% |
| add.Add.time_cats_dogs symlink| 27.9±0.06s |  27.7±0.02s | - 0.8% |
| add.Add.time_cats_dogs hardlink| 34.1±0.2s |   34.3±0.2s |  +0.6% |
|checkout.CheckoutBench.time_cats_dogs copy |  13.4±0.01s | 13.3±0.02s | -0.8%|
|checkout.CheckoutBench.time_cats_dogs symlink |  5.16±0.09s| 5.40±0.09s | 4.6%|
|checkout.CheckoutBench.time_cats_dogs hardlink |  10.5±0.1s | 10.5±0.07s | 0.0% |
| collect.CollectBench.time_stages_collection | 2.46±0s | 2.45 ±0s | - 0.4 % |
| collect.TraverseGitRepoBench.time_repo_traversing |  4.14±0s |  4.06±0s | -2.0% |
| imports.ImportBench.time_imports | 13.9±0.7ms | 14.1±0m | 1.4 % |
| init.InitNoScmBench.time_init | 216±0.6ms | 216±0.8ms | 0.0% |
| init.InitScmBench.time_init | 230±0.8ms | 230±0.7ms | 0.0%|
| push.PushBench.time_cats_dogs | failed | failed
| startup.StartupBench.time_startup | 107±0ms | 106±0ms | -1.0% |
| status.DVCIgnoreBench.time_status |486±0ms | 487±0ms |0.2% |
| status.DVCStatusBench.time_status | 456±0ms | 456±0ms | 0.0% |

dvc/ignore.py
```diff
@@ -357,20 +367,32 @@ def _outside_repo(self, path):
...
         full_target = os.path.abspath(target)
-        if not self._outside_repo(full_target):
-            dirname, basename = os.path.split(os.path.normpath(full_target))
-            pattern = self._get_trie_pattern(dirname)
-            if pattern:
-                matches = pattern.matches(
-                    dirname, basename, os.path.isdir(full_target), True,
-                )
+        if self._outside_repo(full_target):
+            raise DvcException(
```
karajan1001, line 377: This is a problem here that, `check-ignore` is also used in `output validation`, although only for message show.
After this merged, these two exceptions and error message would also be thrown to 

https://github.com/iterative/dvc/blob/31439754377d829da6babffcc720dd696554d8ab/dvc/output/base.py#L626-L627

And the error message here would be shown.

karajan1001: Error message from cls.IsIgnoredError
```
f"Path '{match.file}' is ignored by\n{lines}")
```

karajan1001: Maybe we should not use `fatal` which follows `Git` here to make messages more adapt to our own.

karajan1001:
I add 10 subrepos to the ignore tests and get the following result.

| tests| previous version | after this PR| diff |
|:---:|:---:|:---:|:---:|
| status.DVCIgnoreBench.time_status |547±0ms | 552±0ms | +0.9% |
| status.DVCStatusBench.time_status | 467±0ms | 471±0ms | +0.8% |

Seems that the new version would slightly slower the `dvc status`. 

And I had created an issue with the PR following it.

By the way, the `dvc status` bench would fail because we replaced `DvcIgnore.DVCIGNORE_FILE` with `DVCIGNORE_FILE`

If this PR merged like this, we also need change the `dvc-bench` side.

skshetry:
@karajan1001, please rebase and run `pre-commit run --all-files`. I updated black yesterday which is causing the failure here.

dvc/ignore.py
efiop: Agreed, let's get rid of `fatal` prefix.

dberenbaum:
@efiop When will we be getting back to this one?

PR #6014, opened 2021-05-15
fsspec: split the cloud-specific functionality from the base wrapper

isidentical:


dvc/fs/fsspec_wrapper.py
```diff
@@ -164,3 +149,36 @@ def _download(
             ) as wrapped:
                 with open(to_file, "wb") as fdest:
                     shutil.copyfileobj(wrapped, fdest, length=fobj.blocksize)
+
+
+# pylint: disable=abstract-method
+class CloudFSSpecWrapper(FSSpecWrapper):
```
skshetry, line 155: `Cloud` might be misleading (isn't `gdrive` in cloud?). How about `ObjectFSWrapper`? It is a bit of a mouthful though.

isidentical: What about PrefixBasedFSWrapper?

skshetry: @isidentical, I'd prefer `Object-*`, but no strong opinions.

isidentical: Then Let's go with ObjectFSWrapper. Thanks for the suggestion!

karajan1001 (approved):

PR #6018, opened 2021-05-16
config, remote: Made S3 CA bundle customizable

### LINKED ISSUE #6012: ssl_verify key in remote config does not accept custom CA bundle path and aws config is ignored
The `ssl_verify` key in the remote config gets passed through to the `S3FileSystem` `client_kwargs`:
https://github.com/iterative/dvc/blob/89b40afee740146af42efeb8563c08053f984a88/dvc/fs/s3.py#L105
https://github.com/iterative/dvc/blob/89b40afee740146af42efeb8563c08053f984a88/dvc/fs/fsspec_wrapper.py#L17
https://github.com/iterative/dvc/blob/89b40afee740146af42efeb8563c08053f984a88/dvc/fs/s3.py#L154

These are in turn passed to the `aiobotocore.AioSession` :
https://github.com/dask/s3fs/blob/a3d7a946f85b6dbef62ab75c61fe1319a482c8ba/s3fs/core.py#L366

In the AioSession it checks if the `verify` key is set and if it isn't then it looks in the aws config:
https://github.com/aio-libs/aiobotocore/blob/2a7c7f5a8c7a61daebe484bc5a6f2232607af82c/aiobotocore/session.py#L70-L71
`verify` can either be a boolean _or_ a string, with the latter being a path to a custom CA bundle:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html (see the `verify` argument of the `client` method.)

However, the config schema for DVC only allows boolean for `ssl_verify` and defaults true:
https://github.com/iterative/dvc/blob/89b40afee740146af42efeb8563c08053f984a88/dvc/config_schema.py#L148

The result is that the aws config is never checked and a custom CA bundle cannot be used. If such a CA bundle is needed when trying to communicate to remote (e.g. using push or pull) the result is
```
[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate
```

I ran into this problem because my company uses a self-hosted S3 clone with a bundle of internally signed certificates. Setting the `AWS_CA_BUNDLE` environment variable did not resolve the issue. But modifying the config schema to accept a string:
```
Optional("ssl_verify", default=True): Any(Bool, str),
```
and running
```
dvc remote modify object-store ssl_verify "$HOME/.aws/cabundle.pem"
```
resolved the issue for me.

I'm happy to open a pull request to make the change to the config schema if that solution is acceptable, but it would be my first contribution (for any OSS project!), so it'll take extra time for me to setup my environment, etc.

rgvanwesep:
`botocore` allows a path to a custom CA bundle either by passing a path to the CA bundle file into the verify argument of `boto3.session.Session.client` (see https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html) or passing `None` (the default) which will fall back to the AWS config. Previously, the DVC config only accepted a
boolean into the `ssl_verify` option in the remote S3 config. This changes the DVC config to accept both string and `None` in addition to boolean and defaults to `None`. I also changed the default for `ssl_verfiy` to `None` in `BaseS3FileSystem`. Thus, if `ssl_verify` is not provided, `botocore` will fall back to the AWS config.

#### Testing

Unit tests to cover the changes to the config schema and additional `ssl_verify` types that will be passed into `S3FileSystem`. Also, ran
```
dvc push -r object-store data/cifar-10-python.tar.gz
```
in my work environment that has a private S3 endpoint that requires a custom CA bundle, both with and without `ssl_verify` specified in the DVC config. This was successful, showing that communication could be established. And I ran
```
dvc remote modify object-store ssl_verify "$HOME/.aws/cabundle.pem"
```
and confirmed that the custom CA bundle path was added to the config.

Fixes #6012

### DIFF
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -145,7 +145,7 @@ SCHEMA = {
                     "session_token": str,
                     Optional("listobjects", default=False): Bool,  # obsoleted
                     Optional("use_ssl", default=True): Bool,
-                    Optional("ssl_verify", default=True): Bool,
+                    Optional("ssl_verify", default=None): Any(Bool, str, None),
                     "sse": str,
                     "sse_kms_key_id": str,
                     "acl": str,
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -145,7 +145,7 @@ SCHEMA = {
                     "session_token": str,
                     Optional("listobjects", default=False): Bool,  # obsoleted
                     Optional("use_ssl", default=True): Bool,
-                    Optional("ssl_verify", default=None): Any(Bool, str, None),
+                    "ssl_verify": Any(Bool, str),
                     "sse": str,
                     "sse_kms_key_id": str,
                     "acl": str,
--- a/dvc/fs/s3.py
+++ b/dvc/fs/s3.py
@@ -102,7 +102,7 @@ class BaseS3FileSystem(FSSpecWrapper):  # pylint:disable=abstract-method
         client = login_info["client_kwargs"]
         client["region_name"] = config.get("region")
         client["endpoint_url"] = config.get("endpointurl")
-        client["verify"] = config.get("ssl_verify", True)
+        client["verify"] = config.get("ssl_verify")
 
         # encryptions
         additional = login_info["s3_additional_kwargs"]
--- a/tests/unit/fs/test_s3.py
+++ b/tests/unit/fs/test_s3.py
@@ -39,7 +39,15 @@ def test_verify_ssl_default_param(dvc):
     }
     fs = S3FileSystem(**config)
 
-    assert fs.fs_args["client_kwargs"]["verify"]
+    assert "client_kwargs" not in fs.fs_args
+
+    config = {
+        "url": url,
+        "endpointurl": "https://my.custom.s3:1234",
+    }
+    fs = S3FileSystem(**config)
+
+    assert "verify" not in fs.fs_args["client_kwargs"]
 
 
 def test_s3_config_credentialpath(dvc, monkeypatch):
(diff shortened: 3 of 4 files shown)


pmrowla:
@rgvanwesep this PR will require a docs update - in the `Click for Amazon S3` section in https://dvc.org/doc/command-reference/remote/modify#remote-modify should note that `ssl_verify` can be set to true/false or a `ca_bundle` path, and that if unset, we will fall back to verifying certificates using the AWS CLI config's `ca_bundle` (if it exists)

See https://dvc.org/doc/user-guide/contributing/docs#submitting-changes for info on submitting the docs PR, the file you need to update is: https://github.com/iterative/dvc.org/blob/master/content/docs/command-reference/remote/modify.md

rgvanwesep:
@pmrowla Thanks for pointing that out. I should be able to put up the docs PR today.

rgvanwesep:
The doc PR:
https://github.com/iterative/dvc.org/pull/2483

dvc/config_schema.py
```diff
@@ -145,7 +145,7 @@ class RelPath(str):
                     "session_token": str,
                     Optional("listobjects", default=False): Bool,  # obsoleted
                     Optional("use_ssl", default=True): Bool,
-                    Optional("ssl_verify", default=True): Bool,
+                    Optional("ssl_verify", default=None): Any(Bool, str, None),
```
efiop, line 148: I guess we could just

```suggestion
                    "ssl_verify": Any(Bool, str, None),
```

since they are optional by default.

rgvanwesep: Yeah, that makes sense. And at that point I think I can get rid of the `None` so that it is `"ssl_verify": Any(Bool, str),`. I'm testing the change now and should be able to push soon.

skshetry:
I thought this added support for http(s) as well. Looking into it. :)

PR #6029, opened 2021-05-18
ui: migrate completion to ui.write

skshetry:

### DIFF
--- a/dvc/command/completion.py
+++ b/dvc/command/completion.py
@@ -4,6 +4,7 @@ import logging
 import shtab
 
 from dvc.command.base import CmdBaseNoRepo, append_doc_link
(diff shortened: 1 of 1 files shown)


dvc/command/completion.py
```diff
@@ -95,7 +96,7 @@ def run(self):
         parser = get_main_parser()
         shell = self.args.shell
         script = shtab.complete(parser, shell=shell, preamble=PREAMBLE)
-        print(script)
+        ui.write(script, force=True)
```
skshetry, line 99: Note that `force=True` will print completion script even in the `--quiet` mode (which what was already there with `print()`).

karajan1001 (approved):

PR #6037, opened 2021-05-19
[RFC] ui: write gets a styled param with rich console printing

skshetry:
This PR adds a `styled=True` arg, which will use `rich` for printing, which means we can use it to format the string segments with `[red]` kind of syntax (see the changes in the checkout). 

Though I am still not sure that much of this PR, on whether this should be a special API (`printf`? 😄) or just the way it is in this PR.

I am just putting this out there, so any suggestions/feedback will be welcomed.

### DIFF
--- a/dvc/ui/__init__.py
+++ b/dvc/ui/__init__.py
@@ -7,7 +7,6 @@ from typing import (
     Iterable,
     Optional,
     Sequence,
-    TextIO,
     Union,
 )
 
@@ -58,15 +57,13 @@ class Console:
         style: str = None,
         sep: str = None,
         end: str = None,
-        flush: bool = False,
     ) -> None:
         return self.write(
             *objects,
             style=style,
             sep=sep,
             end=end,
-            file=sys.stderr,
-            flush=flush,
+            stderr=True,
         )
 
     def write(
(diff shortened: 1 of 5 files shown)


dvc/ui/__init__.py
```diff
@@ -75,16 +72,22 @@ def write(
...
     ) -> None:
+        sep = " " if sep is None else sep
+        end = "\n" if end is None else end
         if not self._enabled and not force:
             return
 
-        file = file or sys.stdout
+        if styled:
+            console = self.error_console if stderr else self.rich_console
+            return console.print(*objects, sep=sep, end=end)
```
skshetry, line 86: Note that `rich_console` expands the tab, whereas `print` does not. But, `logger.info` did, so we are still compatible.

pmrowla:
What about just using rich `console.print` all the time?

skshetry:
> What about just using rich `console.print` all the time?

@pmrowla, I have not been able to decide that yet. It's kinda slow and it's expected from a general-purpose renderer. 
I think I have to bite the bullet at some point (maybe after the migration is complete), but I am not sure right now.

skshetry:
Merging it as-is, as everyone in the team seems okay with it. We will reconsider moving completely to rich later.

Issue #5684, opened 2021-03-24
exp:Error "cannot stash changes - there is nothing to stash" on windows


gdiepen:
# Bug Report

## Description

After creating a new project and initializing dvc on it I would like to run an experiment with setting a parameter value. If I run this command on a linux machine it works, running the same command on windows gives me the following error:

```
ERROR: unexpected error - 'cannot stash changes - there is nothing to stash.'
```
The full output of the different command are:
![image](https://user-images.githubusercontent.com/2786853/112284660-c21caf00-8c89-11eb-9c99-567876cce258.png)

If I run exactly the same commands on exactly the same clean repo under linux, this works, as you can see in the screenshot below:
![image](https://user-images.githubusercontent.com/2786853/112284790-e6788b80-8c89-11eb-9a16-0c772a29839e.png)

### Reproduce

I created a very small demo project that depends on a value y (taken from y.txt) and a parameter x from params.yaml and it calculates the product as a 'metric'.

The steps will clone this test repository and initiate an experiment with setting the x parameter value to 40

1. git clone https://github.com/gdiepen/test_experiments_dvc
2. cd test_experiments_dvc
3. dvc exp run --set-param settings.x=40

If you run these steps on windows, you will get the error message about stash

### Expected

Under windows I would expect the same to happen, namely that an experiment is created which you can see in the `dvc exp show` command.

### Environment information

**Output of `dvc doctor`:**

```console
$ dvc doctor
DVC version: 2.0.6 (pip)
---------------------------------
Platform: Python 3.8.8 on Windows-10-10.0.18362-SP0
Supports: hdfs, http, https, s3
Cache types: <https://error.dvc.org/no-dvc-cache>
Caches: local
Remotes: None
Workspace directory: NTFS on C:\
Repo: dvc, git
```

**Additional Information (if any):**

Using the same version of git and the same version of dvc on a linux machine everything works, it only appears to happen on my windows machine.

gdiepen:
Did see that I made small mistake in the original dvc.yaml configuration and not include settings.x as a parameter on which the stage was depending. Just updated the repository I refer to for this (otherwise changing the parameters might not trigger a new run)

gdiepen:
After chat on discord, got asked to run the command:
`dvc exp run --set-param settings.x=40 -vv`
to get more verbose information.

The output of this command on a clean clone of the repo is the following:
```
[...]

Issue #5638, opened 2021-03-16
exp: running experiments is very slow


courentin:
# Bug Report

## Description

Hello!

I'm trying to test the suggestion made in #5557 but running `dvc exp run export_corpus_test_fr --set-param debug=true` is taking a lot of time (~8min) while reproducing the stage without experiment does not take so much time.
I tried to add as many directories in `.dvcignore` as I could but it doesn't change anything. So I suspect something in dvc.

### Environment information

I'm using the dvc version from master (862e18) and dulwich from master too (bbcc4b).

```console
DVC version: 2.0.5+862e18
---------------------------------
Platform: Python 3.7.6 on Darwin-20.2.0-x86_64-i386-64bit
Supports: http, https, s3
Cache types: reflink, hardlink, symlink
Cache directory: apfs on /dev/disk1s2s1
Caches: local
Remotes: s3
Workspace directory: apfs on /dev/disk1s2s1
Repo: dvc, git
```

**Additional Information:**

![Screenshot 2021-03-16 at 18 56 27](https://user-images.githubusercontent.com/12371832/111357145-520a9980-8689-11eb-9ca2-a25bb5a0524b.png)

Profiling information here: [exp.prof.zip](https://github.com/iterative/dvc/files/6151256/exp.prof.zip)

pmrowla:
The issue is with computing git status (via dulwich), not DVC status, so using `.dvcignore` won't address the issue.

pmrowla:
@courentin @ettadar do you have large untracked directories in your workspaces?

I know @ettadar's issue mentioned `.venv`. Is your venv directory gitignored? (if not, can you try adding it to your `.gitignore` and then see if the performance changes)

shcheklein:
If it is the case with `.venv` not being git-ignored,  it somewhat reminds me a discussion we had about DVC collecting stages and data files - should we ignore some common subset of directories (.env, node_modules, etc). I wonder if it makes sense in this kind of situations to at least write something if we some of those? Or may be mention in docs/faq?

ettadar:
> @courentin @ettadar do you have large untracked directories in your workspaces?

Nop.

> I know @ettadar's issue mentioned .venv in particular. Is your venv directory gitignored? (if not, can you try adding it to your .gitignore and then see if the performance changes)

Yep, it is already gitignored. This is the output of `git status`:
```
On branch develop
Your branch is up to date with 'origin/develop'.

Changes not staged for commit:
[...]

PR #6049, opened 2021-05-24
git: update dulwich to 0.20.22

pmrowla:

- Related to https://github.com/iterative/dvc/issues/2215
- Will close https://github.com/iterative/dvc/issues/5684
- Will close https://github.com/iterative/dvc/issues/5638

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,6 +10,8 @@ from .remotes import *  # noqa, pylint: disable=wildcard-import
 os.environ["DVC_TEST"] = "true"
 # Ensure progress output even when not outputting to raw sys.stderr console
 os.environ["DVC_IGNORE_ISATTY"] = "true"
+# Disable system git config
+os.environ["GIT_CONFIG_NOSYSTEM"] = "1"
 
 REMOTES = {
     # remote: enabled_by_default?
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -10,6 +10,8 @@
 os.environ["DVC_TEST"] = "true"
 # Ensure progress output even when not outputting to raw sys.stderr console
 os.environ["DVC_IGNORE_ISATTY"] = "true"
+# Disable system git config
+os.environ["GIT_CONFIG_NOSYSTEM"] = "1"
```
pmrowla, line 14: dulwich now reads the system-wide default config, which means it will read `core.autocrlf=true` for typical windows Git environments. This should be disabled in CI, as DVC does not read Git configs, and will not do any autocrlf normalization when importing from Git.

Several of our tests git-commit files with CRLF's (usually via Windows `echo` which behaves differently than unix `echo`). If `autocrlf=true`, the original workspace file will have`\r\n` line endings, but the git-committed (and git-index) version will have nomalized `\n` line endings.

If we `dvc import` or `dvc get` that git committed file, DVC will read/download the git-committed version containing `\n` endings (even on Windows). This will break tests on Windows which do a direct comparison between the "original" workspace file and the imported one (since the line endings will not match).

PR #6059, opened 2021-05-25
odb: don't use fs.path_info

efiop:
Pre-requisite for getting rid of `fs.path_info`

### DIFF
--- a/dvc/fs/webdav.py
+++ b/dvc/fs/webdav.py
@@ -71,20 +71,18 @@ class WebDAVFileSystem(BaseFileSystem):  # pylint:disable=abstract-method
         # Connection timeout
         self.timeout = config.get("timeout", 30)
 
-        # Get URL from configuration
-        self.url = config.get("url", None)
+        self.hostname = config["host"]
+        self.prefix = config.get("prefix")
+        self.user = config.get("user")
 
-        # If URL in config parse path_info
-        if self.url:
-            self.path_info = self.PATH_CLS(self.url)
-
-            # If username not specified try to use from URL
-            if self.user is None and self.path_info.user is not None:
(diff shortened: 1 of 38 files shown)


efiop:
For the record: Too much stuff in this PR, will split it up into smaller pre-requisites for odb.

efiop:
ok, rushing a bit, a follow-up will remove `path_info` from methods, simplifying the fsspec migration.

dvc/fs/webdav.py
```diff
@@ -71,20 +71,18 @@ def __init__(self, **config):
...
-        # Get URL from configuration
-        self.url = config.get("url", None)
-
-        # If URL in config parse path_info
-        if self.url:
-            self.path_info = self.PATH_CLS(self.url)
-
-            # If username not specified try to use from URL
-            if self.user is None and self.path_info.user is not None:
-                self.user = self.path_info.user
```
skshetry, line 83: Is this going to be implemented in the upcoming PR?

efiop: @skshetry Yeah, it needs to be a part of `_get_kwargs_from_urls`. I've noticed that webdav4 doesn't have it, is that intended?

skshetry: This is a dvc thingy, as it needs to ask for password in this case. Webdav4 does support urls having both password and username.

efiop: @skshetry Looking at other filesystems, it seems like `password` and `usertname` should be part of fs, not part of path, as otherwise you'll be forced to always carry those around like we did with path_info, which is not ideal. WDYT, should we adjust webdav4 to maybe only allow `base_url` with such options? Or am I missing something and this is actually intended?

PR #6060, opened 2021-05-25
Issue 4042 cmd suggestions

naveen24691:

This PR is for the issue: https://github.com/iterative/dvc/issues/4042

I am using regex to figure which command needs to be picked for suggestions. I have tested the regex most of the error scenarios of the argparse library. It is able to handle all the errors during argparse. Sadly, this regex approach is not very future proof but it was getting very complex to figure out the affected command.

Sample outputs :
```
$ dvc commt
ERROR: argument COMMAND: invalid choice: 'commt' (choose from 'init', 'get', 'get-url', 'destroy', 'add', 'remove', 'move', 'unprotect', 'run', 'repro', 'pull', 'push', 'fetch', 'status', 'gc', 'import', 'import-url', 'config', 'checkout', 'remote', 'cache', 'metrics', 'params', 'install', 'root', 'list', 'freeze', 'unfreeze', 'dag', 'daemon', 'commit', 'completion', 'diff', 'version', 'doctor', 'update', 'git-hook', 'plots', 'stage', 'experiments', 'exp', 'check-ignore', 'live')

The most similar commands are:
        commit

usage: dvc [-q | -v] [-h] [-V] [--cd <path>] COMMAND ...

Data Version Control

optional arguments:
  -q, --quiet        Be quiet.
  -v, --verbose      Be verbose.
  -h, --help         Show this help message and exit.
  -V, --version      Show program's version.
  --cd <path>        Change to directory before executing.

Available Commands:
  COMMAND            Use `dvc COMMAND --help` for command-specific help.
    init             Initialize DVC in the current directory.
    get              Download file or directory tracked by DVC or by Git.
    get-url          Download or copy files from URL.
    destroy          Remove DVC files, local DVC config and data cache.
    add              Track data files or directories with DVC.
    remove           Remove stages from dvc.yaml and/or stop tracking files or directories.
    move             Rename or move a DVC controlled data file or a directory.
    unprotect        Unprotect tracked files or directories (when hardlinks or symlinks have been enabled with `dvc config cache.type`).
    run              Generate a dvc.yaml file from a command and execute the command.
    repro            Reproduce complete or partial pipelines by executing their stages.
    pull             Download tracked files or directories from remote storage.
    push             Upload tracked files or directories to remote storage.
    fetch            Download files or directories from remote storage to the cache.
    status           Show changed stages, compare local cache and a remote storage.
    gc               Garbage collect unused objects from cache or remote storage.
    import           Download file or directory tracked by DVC or by Git into the workspace, and track it.
    import-url       Download or copy file from URL and take it under DVC control.
    config           Get or set config options.
    checkout         Checkout data files from cache.
    remote           Set up and manage data remotes.
    cache            Manage cache settings.
    metrics          Commands to display and compare metrics.
    params           Commands to display params.
    install          Install DVC git hooks into the repository.
    root             Return the relative path to the root of the DVC project.
    list             List repository contents, including files and directories tracked by DVC and by Git.
    freeze           Freeze stages or .dvc files.
    unfreeze         Unfreeze stages or .dvc files.
    dag              Visualize DVC project DAG.
    commit           Record changes to files or directories tracked by DVC by storing the current versions in the cache.
    completion       Generate shell tab completion.
    diff             Show added, modified, or deleted data between commits in the DVC repository, or between a commit and the workspace.
    version (doctor)
                     Display the DVC version and system/environment information.
    update           Update data artifacts imported from other DVC repositories.
    plots            Commands to visualize and compare plot metrics in structured files (JSON, YAML, CSV, TSV).
    stage            Commands to list and create stages.
    experiments (exp)
                     Commands to run and compare experiments.
    check-ignore     Check whether files or directories are excluded due to `.dvcignore`.

```

Another sample:

```
$ dvc remote remoev
ERROR: argument cmd: invalid choice: 'remoev' (choose from 'add', 'default', 'modify', 'list', 'remove', 'rename')

The most similar commands are:
        remove rename

usage: dvc remote [-h] [-q | -v] {add,default,modify,list,remove,rename} ...

Set up and manage data remotes.
Documentation: <https://man.dvc.org/remote>

positional arguments:
  {add,default,modify,list,remove,rename}
                        Use `dvc remote CMD --help` for command-specific help.
    add                 Add a new data remote.
    default             Set/unset the default data remote.
    modify              Modify the configuration of a data remote.
    list                List all available data remotes.
    remove              Remove a data remote.
    rename              Rename a DVC remote

optional arguments:
  -h, --help            show this help message and exit
  -q, --quiet           Be quiet.
  -v, --verbose         Be verbose.

```

I havent added the test cases for the string library yet. Let me know if something like this will be required in the future, then I will implement test cases as well.

### DIFF
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -136,8 +136,14 @@ class DvcParser(argparse.ArgumentParser):
                 0
             ].help
         )
+
+        parent_cmd = self.prog
+        command_text = (
+            "subcommand" if len(parent_cmd.split()) > 1 else "command"
+        )
         multi_line_message = [
-            f"dvc: '{self.last_invalid_command}' is not a valid dvc command. "
+            f"dvc: '{self.last_invalid_command}' is "
+            f"not a valid `{parent_cmd}` {command_text}. "
             f"{help_cmd}"
             "\t",
             "",
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -115,7 +115,7 @@ class DvcParser(argparse.ArgumentParser):
         _find_parser(self, cmd_cls)
 
     def _check_and_log_suggestions(self, message):
-        cmd = regex_search(string=message, regex="'(.*?)'")
+        cmd = regex_search(string=message, regex="[?=invalid choice: ]'(.*?)'")
         if not cmd:
             return
 
@@ -124,7 +124,6 @@ class DvcParser(argparse.ArgumentParser):
             return
 
         multi_line_message = [
-            "",
             "",
             "The most similar commands are:",
             f"\t{' '.join(suggestions)}",
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -46,7 +46,7 @@ from .command import (
 )
 from .command.base import fix_subparsers
 from .exceptions import DvcParserError
-from .utils.string import fuzzy_match, regex_search
+from .utils.string import fuzzy_match
 
 logger = logging.getLogger(__name__)
 
@@ -109,17 +109,19 @@ def _find_parser(parser, cmd_cls):
 class DvcParser(argparse.ArgumentParser):
     """Custom parser class for dvc CLI."""
 
+    last_invalid_command = None
+
     def error(self, message, cmd_cls=None):  # pylint: disable=arguments-differ
         logger.error(message)
-        self._check_and_log_suggestions(message)
+        if self.last_invalid_command:
+            self.print_suggestions()
+            raise DvcParserError()
         _find_parser(self, cmd_cls)
 
-    def _check_and_log_suggestions(self, message):
-        cmd = regex_search(string=message, regex="[?=invalid choice: ]'(.*?)'")
-        if not cmd:
-            return
-
-        suggestions = fuzzy_match(cmd, self._get_choices())
+    def print_suggestions(self):
+        suggestions = fuzzy_match(
+            self.last_invalid_command, self._get_choices()
+        )
         if not suggestions:
             return
 
@@ -155,6 +157,15 @@ class DvcParser(argparse.ArgumentParser):
             return []
 
         return cmd_choices[0]
+
+    def _check_value(self, action, value):
+        # NOTE: overriding to get the last invalid command
+        try:
+            self.last_invalid_command = None
+            super()._check_value(action, value)
+        except argparse.ArgumentError as arg_exc:
+            self.last_invalid_command = value
+            raise arg_exc
 
 
 class VersionAction(argparse.Action):  # pragma: no cover
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -112,26 +112,40 @@ class DvcParser(argparse.ArgumentParser):
     last_invalid_command = None
 
     def error(self, message, cmd_cls=None):  # pylint: disable=arguments-differ
-        logger.error(message)
-        if self.last_invalid_command:
-            self.print_suggestions()
+
+        if self.last_invalid_command and self.print_suggestions():
             raise DvcParserError()
+
+        logger.error(message)
         _find_parser(self, cmd_cls)
 
     def print_suggestions(self):
         suggestions = fuzzy_match(
             self.last_invalid_command, self._get_choices()
         )
-        if not suggestions:
-            return
 
+        subparser = self._subparsers
+        if (
+            not suggestions
+            or not subparser._group_actions  # pylint: disable=protected-access
+        ):
+            return False
+
+        help_cmd = (
+            subparser._group_actions[  # pylint: disable=protected-access
+                0
+            ].help
+        )
         multi_line_message = [
+            f"dvc: '{self.last_invalid_command}' is not a valid dvc command. "
+            f"{help_cmd}"
+            "\t",
             "",
             "The most similar commands are:",
             f"\t{' '.join(suggestions)}",
-            "",
         ]
         logger.info("\n".join(multi_line_message))
+        return True
 
     def parse_args(self, args=None, namespace=None):
         # NOTE: overriding to provide a more granular help message.
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -46,6 +46,7 @@ from .command import (
 )
 from .command.base import fix_subparsers
 from .exceptions import DvcParserError
+from .utils.string import fuzzy_match, regex_search
 
 logger = logging.getLogger(__name__)
 
@@ -110,7 +111,26 @@ class DvcParser(argparse.ArgumentParser):
 
     def error(self, message, cmd_cls=None):  # pylint: disable=arguments-differ
         logger.error(message)
+        self._check_and_log_suggestions(message)
         _find_parser(self, cmd_cls)
+
+    def _check_and_log_suggestions(self, message):
+        cmd = regex_search(string=message, regex="'(.*?)'")
+        if not cmd:
+            return
+
+        suggestions = fuzzy_match(cmd, self._get_choices())
+        if not suggestions:
+            return
+
+        multi_line_message = [
+            "",
+            "",
+            "The most similar commands are:",
+            f"\t{' '.join(suggestions)}",
+            "",
+        ]
+        logger.info("\n".join(multi_line_message))
 
     def parse_args(self, args=None, namespace=None):
         # NOTE: overriding to provide a more granular help message.
@@ -121,6 +141,21 @@ class DvcParser(argparse.ArgumentParser):
             msg = "unrecognized arguments: %s"
             self.error(msg % " ".join(argv), getattr(args, "func", None))
         return args
+
+    def _get_choices(self) -> list:
+        """ Gets the list of choices for dvc and its sub-commands"""
+        cmd_choices = [
+            list(action.choices.keys())
+            for action in self._actions
+            if action.dest == "cmd"
+            and isinstance(
+                action.choices, dict
+            )  # pylint: disable=protected-access
+        ]
+        if not cmd_choices:
+            return []
+
+        return cmd_choices[0]
 
 
 class VersionAction(argparse.Action):  # pragma: no cover
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -142,7 +142,7 @@ class DvcParser(argparse.ArgumentParser):
         return args
 
     def _get_choices(self) -> list:
-        """ Gets the list of choices for dvc and its sub-commands"""
+        """Gets the list of choices for dvc and its sub-commands"""
         cmd_choices = [
             list(action.choices.keys())
             for action in self._actions
--- a/tests/func/test_cli.py
+++ b/tests/func/test_cli.py
@@ -261,3 +261,12 @@ def test_unknown_subcommand_help(capsys):
     captured = capsys.readouterr()
     help_output = captured.out
     assert output == help_output
+
+
+def test_cli_suggestions():
+    try:
+        _ = parse_args(["psh"])
+        _ = parse_args(["fileas"])
+        raise AssertionError
+    except DvcParserError:
+        assert True
--- a/tests/func/test_cli.py
+++ b/tests/func/test_cli.py
@@ -265,7 +265,7 @@ def test_unknown_subcommand_help(capsys):
     assert output == help_output
 
 
-def test_cli_suggestions():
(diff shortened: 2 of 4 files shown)


naveen24691:
One more way to implement is to override the `_check_value `method and the `last_checked_cmd `attribute to the value. Use it to figure out whether to show the suggestions

dvc/cli.py
```diff
@@ -110,8 +111,26 @@ class DvcParser(argparse.ArgumentParser):
 
     def error(self, message, cmd_cls=None):  # pylint: disable=arguments-differ
         logger.error(message)
+        self._check_and_log_suggestions(message)
         _find_parser(self, cmd_cls)
 
+    def _check_and_log_suggestions(self, message):
+        cmd = regex_search(string=message, regex="[?=invalid choice: ]'(.*?)'")
```
karajan1001, line 118: Is any method to get the false cmd directly instead of parsing it from the error message?

karajan1001:
@naveen24691 Thanks for your contributions, looks pretty good to me. 

Can we use the subcommand to adjust the final suggestion,  For example, only `dvc remote` can modify while `dvc remove` can't. So the `dvc remoe modify` should be suggested to `dvc remote modify` instead of `dvc remove modify`

dvc/cli.py
pared: In current state we are running the whole command before trying to get the error message and match failed command. We should do that before DVC tries to execute the command, mostly due to fact that this check will run upon every exception that DVC faces.

@naveen24691 in `cli.py` there is argument parsing part, I think that potential misuse detection should be happening somewhere there.

naveen24691:
> 
> 
> @naveen24691 Thanks for your contributions, looks pretty good to me.
> 
> Can we use the subcommand to adjust the final suggestion, For example, only `dvc remote` can modify while `dvc remove` can't. So the `dvc remoe modify` should be suggested to `dvc remote modify` instead of `dvc remove modify`

I agree. I will add an additional check to recommend supported command only

dvc/cli.py
naveen24691: @karajan1001  We can override `_check_value ` method. This method is called to check if command is supported or not. I can maintain an attribute to store the last checked value and use it here instead in case of error. But it will be difficult to figure it out for which reason this error method was called. We still need to check if "invalid choice" text is present in the string or not. We know something is wrong, only when error method is called with  a string message from within the `argparse.ArgumentParser `class.


@pared I think this `DvcParser.error` method is called only during the argument parser step. Command is executed post the successful parsing. Please correct me If I am wrong

pared:
@naveen24691 I am sorry for lack of response, I did not notice your comment regarding my review.

You are right that `error` is called before the execution. Thic comment can be disregarded.
Regarding the change - I am not a fan of matching regex on error message, it seems that it will be vulnerable to simple string changes on side of argparse. Also currently on wrong command we will get, for example:

```bash
ERROR: argument COMMAND: invalid choice: 'plts' (choose from 'init', 'get', 'get-url', 'destroy', 'add', 'remove', 'move', 'unprotect', 'run', 'repro', 'pull', 'push', 'fetch', 'status', 'gc', 'import', 'import-url', 'config', 'checkout', 'remote', 'cache', 'metrics', 'params', 'install', 'root', 'list', 'freeze', 'unfreeze', 'dag', 'daemon', 'commit', 'completion', 'diff', 'version', 'doctor', 'update', 'git-hook', 'plots', 'stage', 'experiments', 'exp', 'check-ignore', 'live')

The most similar commands are:
	plots

usage: dvc [-q | -v] [-h] [-V] [--cd <path>] COMMAND ...

Data Version Control

optional arguments:
  -q, --quiet        Be quiet.
  -v, --verbose      Be verbose.
  -h, --help         Show this help message and exit.
  -V, --version      Show program's version.
  --cd <path>        Change to directory before executing.

Available Commands:
  COMMAND            Use `dvc COMMAND --help` for command-specific help.
    init             Initialize DVC in the current directory.
    get              Download file or directory tracked by DVC or by Git.
    get-url          Download or copy files from URL. 
...
```

So, the help message gets occluded by help message. I think that it should show up below the help message, or even the help message should not show up at all.

As @karajan1001 mentioned, did you consider different ways of validating the commands?
It seems to me that it would be possible to obtain list of commands via (for example) `self._subparsers._group_actions[0].choices`. Should we rely on parsing error message when it seems possible to get necessary data programatically and validate the command before error is thrown?  What do you thin @naveen24691?

naveen24691 (commented):
Thanks @pared , I will remove the help message and just show the suggestions. I agree, I also don't like parsing the error message. I think I have figured out a way to remove the parsing snippet and do it more concretely.

I will update once it is done. Thanks

pared:
> @karajan1001 I have checked the case you suggested. In case of, dvc remoe modify , parser immediately fails at remoe. It doesnt event get to parse modify. It will be difficult to add a check for the sub-command that way. We will have to explicitly parse the rest of the sub-command to generate correct combination. I think it will create duplicate work. Let me know what you think>

I think there is no need to match whole chain of commands. If `dvc remoe modify` fails, what we want is to get information that `remoe` is probably a `remote`. If `modify` is wrong too, we will get another error message.

pared (commented):
Looks good to me. The new problem is that one of the tests have been veryfing that we print help message on error. I guess my suggestion about removing that was wrong at least for now. I still believe that we are outputting too much on such simple use case, while we could do what git does. Having said so, this is not the scope of this PR, so @naveen24691 lets remove the `raise DvcParserError`, make the test pass and we can take care of messy messages in the future.

naveen24691:
I have implemented as suggested. I have also added test cases for string class. The failing test automatically resolved as if no suggestion found, returning earlier response.

Here is the sample output now:

```
$ git commmut
git: 'commmut' is not a git command. See 'git --help'.

The most similar command is
        commit
$ dvc addd
dvc: 'addd' is not a valid dvc command. Use `dvc COMMAND --help` for command-specific help.

The most similar commands are:
        add
$ dvc remote modf
dvc: 'modf' is not a valid dvc command. Use `dvc remote CMD --help` for command-specific help.

The most similar commands are:
        modify
$ dvc remot modify
dvc: 'remot' is not a valid dvc command. Use `dvc COMMAND --help` for command-specific help.

The most similar commands are:
        remote remove root
$ dvc unkown
ERROR: argument COMMAND: invalid choice: 'unkown' (choose from 'init', 'get', 'get-url', 'destroy', 'add', 'remove', 'move', 'unprotect', 'run', 'repro', 'pull', 'push', 'fetch', 'status', 'gc', 'import', 'import-url', 'config', 'checkout', 'remote', 'cache', 'metrics', 'params', 'install', 'root', 'list', 'freeze', 'unfreeze', 'dag', 'daemon', 'commit', 'completion', 'diff', 'version', 'doctor', 'update', 'git-hook', 'plots', 'stage', 'experiments', 'exp', 'check-ignore', 'live')
usage: dvc [-q | -v] [-h] [-V] [--cd <path>] COMMAND ...

Data Version Control

optional arguments:
  -q, --quiet        Be quiet.
  -v, --verbose      Be verbose.
  -h, --help         Show this help message and exit.
  -V, --version      Show program's version.
  --cd <path>        Change to directory before executing.

Available Commands:
  COMMAND            Use `dvc COMMAND --help` for command-specific help.
    init             Initialize DVC in the current directory.
    get              Download file or directory tracked by DVC or by Git.
    get-url          Download or copy files from URL.
    destroy          Remove DVC files, local DVC config and data cache.
    add              Track data files or directories with DVC.
    remove           Remove stages from dvc.yaml and/or stop tracking files or directories.
    move             Rename or move a DVC controlled data file or a directory.
    unprotect        Unprotect tracked files or directories (when hardlinks or symlinks have been enabled with `dvc config cache.type`).
    run              Generate a dvc.yaml file from a command and execute the command.
    repro            Reproduce complete or partial pipelines by executing their stages.
    pull             Download tracked files or directories from remote storage.
    push             Upload tracked files or directories to remote storage.
    fetch            Download files or directories from remote storage to the cache.
    status           Show changed stages, compare local cache and a remote storage.
    gc               Garbage collect unused objects from cache or remote storage.
    import           Download file or directory tracked by DVC or by Git into the workspace, and track it.
    import-url       Download or copy file from URL and take it under DVC control.
    config           Get or set config options.
    checkout         Checkout data files from cache.
    remote           Set up and manage data remotes.
    cache            Manage cache settings.
    metrics          Commands to display and compare metrics.
    params           Commands to display params.
    install          Install DVC git hooks into the repository.
    root             Return the relative path to the root of the DVC project.
    list             List repository contents, including files and directories tracked by DVC and by Git.
    freeze           Freeze stages or .dvc files.
    unfreeze         Unfreeze stages or .dvc files.
    dag              Visualize DVC project DAG.
    commit           Record changes to files or directories tracked by DVC by storing the current versions in the cache.
    completion       Generate shell tab completion.
    diff             Show added, modified, or deleted data between commits in the DVC repository, or between a commit and the workspace.
    version (doctor)
                     Display the DVC version and system/environment information.
    update           Update data artifacts imported from other DVC repositories.
    plots            Commands to visualize and compare plot metrics in structured files (JSON, YAML, CSV, TSV).
    stage            Commands to list and create stages.
    experiments (exp)
                     Commands to run and compare experiments.
    check-ignore     Check whether files or directories are excluded due to `.dvcignore`.
$
```

pmrowla:
> ```
> $ dvc remote modf
> dvc: 'modf' is not a valid dvc command. Use `dvc remote CMD --help` for command-specific help.
> 
> The most similar commands are:
>         modify
> ```

I think we should be clearer in this case that the problem is with a subcommand. With the way the error message is worded right now, it reads to me like we are suggesting that the user meant to run
```
$ dvc modify
```
rather than `remote modify`

It would be better if the error either read like:

> 'modf' is not a valid 'dvc remote' subcommand...

or just listed the entire command+subcommand string like:

> 'remote modf' is not a valid dvc command

naveen24691:
@pmrowla  Thanks for the suggestions. I have incorporated those changes as well. Please take a look.

```
$ dvc remot
dvc: 'remot' is not a valid `dvc` command. Use `dvc COMMAND --help` for command-specific help.

The most similar commands are:
        remote remove root

$ dvc remote modf
dvc: 'modf' is not a valid `dvc remote` subcommand. Use `dvc remote CMD --help` for command-specific help.

The most similar commands are:
        modify
```

pared (changes requested):
One thing regarding tests.
@naveen24691 sorry for prolonged time of response. I assigned myself to this PR so that won't happen again.

tests/func/test_cli.py
```diff
@@ -261,3 +261,12 @@ def test_unknown_subcommand_help(capsys):
...
     assert output == help_output
+
+
+def test_cli_suggestions():
+    try:
+        _ = parse_args(["psh"])
+        _ = parse_args(["fileas"])
+        raise AssertionError
+    except DvcParserError:
+        assert True
```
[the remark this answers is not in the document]
naveen24691: Yes, you are right. That is better way to implement. I have pushed the new commit

tests/func/test_cli.py
```diff
@@ -261,3 +263,9 @@ def test_unknown_subcommand_help(capsys):
     captured = capsys.readouterr()
     help_output = captured.out
     assert output == help_output
+
+
+def test_cli_suggestions():
+    with pytest.raises(DvcParserError):
+        _ = parse_args(["psh"])
+        _ = parse_args(["fileas"])
```
pared, line 271: One more thing here: if `psh` parsing raises `DVCParserError` we will never reach `fileas` part of the code.
What we could do here is:
```suggestion
@pytest.mark.parametrize("cmd", ("psh", "fileas"))
def test_cli_suggestions(cmd):
    with pytest.raises(DvcParserError):
        parse_args([cmd])
```

naveen24691: Thanks a lot! I learned something new today. This is much better way to implement this. Let me check and update the commit.

pared (commented):
LGTM, though i would add functional test that verifies whole process is working as intended (use `main` to call some commands and verify that `caplog` contains proper message)

efiop:
@naveen24691 Thank you for your patience! I'm worried that this increases the maintenance burden for us, so I can't merge this as is. We might consider using it as a library though, but even then I would wait until it is stable and commonly used. So closing for now 🙁

PR #6085, opened 2021-05-31
pin networkx to 2.5

skshetry:
2.1 is more than 3 years old, whereas the latest version 2.5 was released in Aug, 2020.
I am a bit hesitant to add an upper bound here as networkx seems to be requiring a lot of dependencies from the next version onwards.

2.1-2.3 versions were probably not compatible with Python3.9. 

Discord context: https://discord.com/channels/485586884165107732/485596304961962003/848943242481369088

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,7 @@ install_requires = [
     "flufl.lock>=3.2,<4",
     "win-unicode-console>=0.5; sys_platform == 'win32'",
     "pywin32>=225; sys_platform == 'win32'",
-    "networkx>=2.1",
+    "networkx~=2.5",
     "psutil>=5.8.0",
     "pydot>=1.2.4",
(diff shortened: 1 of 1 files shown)


efiop (approved):

casperdcl:
btw technically are you saying 2.4 would work too? In which case we should relax this to include 2.4 to avoid placing some users in dependency hell.

setup.py
```diff
@@ -73,7 +73,7 @@ def run(self):
     "flufl.lock>=3.2,<4",
     "win-unicode-console>=0.5; sys_platform == 'win32'",
     "pywin32>=225; sys_platform == 'win32'",
-    "networkx>=2.1",
+    "networkx~=2.5",
```
casperdcl, line 76:
```suggestion
networkx >=2.1,<2.6; python_version<'3.9'
networkx >=2.4,<2.6; python_version>='3.9'
```

PR #6086, opened 2021-05-31
fs: don't automatically create buckets

### LINKED ISSUE #2101: fs: don't automatically create buckets
~~We currently do that in azure https://github.com/iterative/dvc/blob/master/dvc/remote/azure.py#L90 .~~

Related https://github.com/iterative/dvc.org/issues/415

efiop:
Currently, we have a bit of a mess, where s3 and gs don't create buckets automatically, but azure and oss do. The safe approach here is to simply not create buckets in any of the filesystems. Local/ssh and others are a bit different, as there is no such thing as bucket there, and nonexistend directories are logically part of the path/prefix and not a standalone resource like buckets are.

Pre-requisite for getting rid of fs.path_info.

Fixes #2101

### DIFF
--- a/dvc/fs/oss.py
+++ b/dvc/fs/oss.py
@@ -45,7 +45,6 @@ class OSSFileSystem(BaseFileSystem):  # pylint:disable=abstract-method
         url = config.get("url")
         self.path_info = self.PATH_CLS(url) if url else None
 
-        self.bucket = self.path_info.bucket
         self.endpoint = config.get("oss_endpoint") or os.getenv("OSS_ENDPOINT")
 
         self.key_id = (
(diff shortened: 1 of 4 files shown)


dvc/fs/oss.py
```diff
@@ -68,25 +67,17 @@ def oss_service(self):
...
-
-        # Ensure bucket exists
-        try:
-            bucket.get_bucket_info()
-        except oss2.exceptions.NoSuchBucket:
-            bucket.create_bucket(
-                oss2.BUCKET_ACL_PUBLIC_READ,
-                oss2.models.BucketCreateConfig(
-                    oss2.BUCKET_STORAGE_CLASS_STANDARD
-                ),
```
efiop, line 82: These permissions are clearly bad :facepalm: Just shows how neglected oss has been :slightly_frowning_face:

efiop: For the record, new and improved implementation is coming soon in https://github.com/iterative/ossfs thanks to @karajan1001 's work.

isidentical (approved):
LGTM

jorgeorpinel:
Hi~ **Click for Microsoft Azure Blob Storage** in https://dvc.org/doc/command-reference/remote/modify may need a super quick update after this

image

If possible I'd double check in other places but I think that's the only mention.

The Q is, should we state somewhere (or in all the appropriate remotes) that DVC **does not** create buckets/containers?

efiop:
@jorgeorpinel Good catch, I didn't realize that we've documented that behavior already. Created https://github.com/iterative/dvc.org/pull/2546

Issue #6089, opened 2021-06-01
Old bug (dvc pull from S3 and md5 starts with `00`) reappears with latest release


jnd77:
# Bug Report

dvc pull does not seem to work for files whose md5 starts with 00, when remote is S3

## Description

This bug https://github.com/iterative/dvc/issues/4141 has reappeared with release 2.2.0 (it's working fine with 2.1.0).
Might be due to this commit:
https://github.com/iterative/dvc/pull/5683

Let me know if you need more information ... :)

isidentical:
Would you mind sending the full traceback?

isidentical:
I just tried this as a test case and it seemed to work (`363`'s hash is something that startswith `00`, which is confirmed in the last assert);

```py
@pytest.mark.parametrize("remote", [pytest.lazy_fixture("s3")], indirect=True)
def test_pull_00_prefix(tmp_dir, dvc, remote):
    tmp_dir.dvc_gen({"foo": "363"})
    dvc.push()
    clean(["foo"], dvc)
    
    stats = dvc.pull("foo")
    assert stats["fetched"] == 1
    assert stats["added"] == ["foo"]
    
    [cache_dir] = dvc.cloud.get_remote('upstream').fs.ls(remote)
    assert cache_dir.endswith('00')
```

pmrowla:
@isidentical I think we'd need to test with pulling more than one file at a time, if we're only trying to push/pull a single file, we do the direct `exists()` query instead of using walk/ls when we are checking whether objects exist in the remote.

It looks like we might not be handling the dvc specific `prefix` kwarg for `fs.walk_files()/walk()` any more, if we're just passing it directly into fsspec's `fs.ls` now, fsspec won't do anything with it.

see https://github.com/iterative/dvc/issues/4141#issuecomment-652798544 for an explanation of the original bug, and https://github.com/iterative/dvc/pull/4149#discussion_r448767923 for discussion on why we added the `prefix` kwarg

essentially , there's times that we do `walk("00/0", prefix=True)` where we really want to query for S3 keys that start with `00/0`, not keys inside `00/0/`

isidentical:
> essentially , there's times that we do walk("00/0", prefix=True) where we really want to query for S3 keys that start with 00/0, not keys inside 00/0/ (I'm not sure what API call s3fs will actually run in this case?)

[...]

PR #6096, opened 2021-06-02
fs: temporarily disable prefix-based search for fsspec-implementations

isidentical:
This is a temporary fix for #6089. It disables the partial prefix (`00/0`) search optimization entirely for all fsspec implementations and does it in the regular way (with only listing the full path of `00/`). This API is not actually supported to be optimized on the fsspec side so we will wait for all the clouds we support (s3, azure, google cloud) and then turn on the optimization (In the `ObjectFSWrapper`) this time with just fixing the custom `find()` method in the subclass.

### DIFF
--- a/dvc/objects/db/base.py
+++ b/dvc/objects/db/base.py
@@ -399,8 +399,15 @@ class ObjectDB:
         # hashes_exist() (see ssh, local)
         assert self.fs.TRAVERSE_PREFIX_LEN >= 2
 
+        # During the tests, for ensuring that the traverse behavior
+        # is working we turn on this option. It will ensure the
+        # list_hashes_traverse() is called.
+        always_traverse = getattr(self.fs, "_ALWAYS_TRAVERSE", False)
+
         hashes = set(hashes)
-        if len(hashes) == 1 or not self.fs.CAN_TRAVERSE:
+        if (
+            len(hashes) == 1 or not self.fs.CAN_TRAVERSE
+        ) and not always_traverse:
             remote_hashes = self.list_hashes_exists(hashes, jobs, name)
             return remote_hashes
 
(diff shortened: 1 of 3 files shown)


pmrowla (approved):

tests/func/test_data_cloud.py
```diff
@@ -41,6 +41,14 @@
...
 ]
 
+# Clouds that implement the general methods that can be tested
+# for functional tests that require extensive apis (e.g traversing
+# via walk_files())
+full_clouds = [
+    pytest.lazy_fixture(cloud)
+    for cloud in ["s3", "gs", "azure", "ssh", "hdfs"]
+]
+
```
efiop, line 51: Let's add these directly to `test_pull_00_prefix` parametrization, since they are not used anywhere else.

dvc/objects/db/base.py
```diff
@@ -399,8 +399,15 @@ def hashes_exist(self, hashes, jobs=None, name=None):
         # hashes_exist() (see ssh, local)
         assert self.fs.TRAVERSE_PREFIX_LEN >= 2
 
+        # During the tests, for ensuring that the traverse behavior
+        # is working we turn on this option. It will ensure the
+        # list_hashes_traverse() is called.
+        always_traverse = getattr(self.fs, "_ALWAYS_TRAVERSE", False)
+
```
efiop, line 406: It feels a bit weird to amend the core code like that only for the tests. I guess this is a hint for an upcoming refactor of this part of code :slightly_frowning_face:

efiop: @pmrowla Do you foresee this part being refactored to accommodate the test better? Maybe you have any suggestions?

efiop:
Ok, let's merge and release, but discuss ^^^

Thank you! :pray:

PR #6099, opened 2021-06-02
webdav4: migrate to new library 

skshetry:
Here's the fetch speed comparison between the two libraries:

### webdavclient3
[asciicast](https://asciinema.org/a/dSrJbWLziKcBgvGvU2DZqCScU)

### webdav4

[asciicast](https://asciinema.org/a/fReef1N6FsTZw8tv1WEHFgz7h)

TODO:
- [x] Test ask_password still works
- [x] Test token auth works.
- [x] Test auth still works when the URL contains a username.
- [x] Test external remote case.
- [ ] Test that it works if the server is accessed through a port.
- [ ] Test that it works if the server is behind a path (i.e. not in root URL).
- [ ] Test with Owncloud, NextCloud, and Yandex.
- [ ] Test with certs.

Additionally, while we are at it, it'd be great to support:
- [ ] Retry mechanism on upload
- [ ] Retry/reconnect mechanism on file obj using range (https://github.com/skshetry/webdav4/issues/4)
- [ ] `verify` certs.

### DIFF
--- a/dvc/fs/webdav.py
+++ b/dvc/fs/webdav.py
@@ -1,264 +1,88 @@
-import io
 import logging
 import os
 import threading
-from collections import deque
+from functools import lru_cache
 
-from funcy import cached_property, nullcontext, wrap_prop
+from funcy import cached_property, wrap_prop
 
-from dvc.config import ConfigError
-from dvc.exceptions import DvcException
-from dvc.path_info import HTTPURLInfo, WebDAVURLInfo
-from dvc.progress import Tqdm
+from dvc.path_info import WebDAVURLInfo
 from dvc.scheme import Schemes
 
-from .base import BaseFileSystem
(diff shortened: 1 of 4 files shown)


efiop:
> Here's the fetch speed comparison between the two libraries:

@skshetry x100 faster? :open_mouth:

skshetry:
> > Here's the fetch speed comparison between the two libraries:
> 
> @skshetry x100 faster?  If so, could you elaborate on key differences that allowed you to achieve that?

https://github.com/ezhov-evgeny/webdav-client-python-3/blob/09ae1414ddd331775686ebe55a625e732ef78b14/webdav3/client.py#L337

This has been recently fixed, but not released so far. Just a humblebrag

efiop:
For the record: discussed and decided to merge as-is and the leftover TODOs will be handled in the follow-up. Thank you @skshetry :pray:

Issue #3305, opened 2020-02-11
import: allow chaining imports somehow? (except circular)


jorgeorpinel:
> More context around https://discordapp.com/channels/485586884165107732/485596304961962003/676405501940072453

As of now, imported data is not cached by default, so you won't be able to import any imported data:

repo1 -> ✔️ dvc import data -> repo2 🙂 -> ✖️ dvc import data -> repo3 🙁 

Somehow allowing this could be useful for the case where you're building a data registry based on other previous smaller DVC repos, for example. Right now you have to `dvc get` and then `dvc add` those artifacts from scratch in the data registry (so they can be imported in further DVC repos).

Ruslan mentioned something about using "links" to implement this (on [Discord](https://discordapp.com/channels/485586884165107732/485596304961962003/676466989035945984)).

jorgeorpinel:
> except circular

This means repo1 -> import into repo2 -> import back to repo1 of course cannot be allowed.

So maybe the solution is that when you import an import stage, you simply copy the DVC-file as-is (with it's original source `repo` URL, `rev`, etc. and if the original `rev_lock` exists in the present repo, the import recognizes a circular import and fails.

efiop:
Just a note that we need to be careful about this and consider all the possible corner cases (e.g. circular dependencies). If we allow this, dvc will have to behave like a proper package manager when resolving dependencies, which is very hard (remember pip dep resolution PEP?).

tadejsv:
This would be a very important feature for me. The use case is the following: files (datasets, pretrained models...) is generated across many different repositories, so I need a central data registry to easily catalogue data. Also, in case some of the original data creating repositories are renamed or merged together, I only need to change the import in data registry and not have to track down every single user of the data.

pared:
It seems like this issue consist of two stages:
- [ ] error on chain import and suggest importing from original repo - this one is required to avoid confusion here, and for example in iterative/dvc.org#2079
- [ ] allow chaining imports

pared:
Ok, so to check if chaining imports works I created following test:
```python
def cleanup_repo(repo_dir):
    cache = repo_dir.dvc.config["cache"]["dir"]
    shutil.rmtree(cache)
    os.remove(repo_dir / "data")


def add_remote_and_push(repo_dir):
    repo_dir.add_remote(
        name="str", url=str(repo_dir) + "_storage", default=True
    )
    repo_dir.dvc.push()


@pytest.mark.parametrize("with_cleanup", [0, 1])
[...]

PR #6109, opened 2021-06-03
odb: treat external repo dependencies as objects

pmrowla:


dvc/repo/fetch.py
```diff
@@ -74,8 +79,8 @@ def fetch(
     except DownloadError as exc:
         failed += exc.amount
 
-    for repo_pair, files in used_external.items():
-        d, f = _fetch_external(self, repo_pair.url, repo_pair.rev, files, jobs)
+    if used_external:
+        d, f = _fetch_external(self, used_external, jobs)
```
pmrowla, line 83: We will be able to get rid of `used_external` and `_fetch_external()` entirely after push/pull are refactored to use unified `odb.save` (since `save` can handle being passed both regular and external objects)

dvc/repo/fetch.py
```diff
@@ -100,29 +102,15 @@ def cb(result):
...
-                save(odb, obj, jobs=jobs, download_callback=cb)
-    except CloneError:
-        failed += 1
-        logger.exception(
-            "failed to fetch data for '{}'".format(", ".join(files))
-        )
+    for obj in external_objs:
+        try:
+            save(odb, obj, jobs=jobs, download_callback=cb)
+        except (ObjectError, OSError):
```
pmrowla, line 108: `OSError`s during `odb.add` are equivalent to the old fetch/push/pull upload & download errors.

dvc/objects/external.py
```diff
@@ -0,0 +1,107 @@
...
+        from dvc.scm.base import CloneError
+
+        try:
+            return external_repo(self.repo_url, rev=self.repo_rev, **kwargs)
+        except CloneError as exc:
+            raise ObjectFormatError(
+                f"Could not clone external obj repo '{self.repo_url}'"
+            ) from exc
+
+    def get_staged(self, odb: Optional["ObjectDB"] = None):
```
pmrowla, line 71: It feels a bit off to have an object stage itself, but in this case external repo deps have to be staged before we can compute the hash_info

dvc/objects/__init__.py
```diff
@@ -1,14 +1,28 @@
...
 
-def save(odb, obj, jobs=None, **kwargs):
+def save(
+    odb: "ObjectDB",
+    obj: "HashFile",
+    jobs: Optional[int] = None,
+    **kwargs,
+):
+    if isinstance(obj, ExternalRepoFile):
+        obj = obj.get_staged(odb)
```
pmrowla, line 24: I'm not sure if we will have more object types that work like erepo files in the future, but we could maybe name this `obj.resolve(odb)` instead, and just have it default to a no-op for regular `HashFiles` (to get rid of this erepo file specific isinstance() check)

efiop:
For the record: discussed with @pmrowla that it might be worth it to not create a new object type, but rather to collect regular objects and return them with an odb from the erepo.

dvc/objects/__init__.py
```diff
@@ -1,14 +1,31 @@
...
+    from .db.base import ObjectDB
+    from .file import HashFile
+
 logger = logging.getLogger(__name__)
 
 
-def save(odb, obj, jobs=None, **kwargs):
+class UsedObjectsPair(NamedTuple):
+    remote: Optional["Remote"]
+    objs: Set["HashFile"]
```
pmrowla, line 20: This whole thing is temporary, I think what we want (for now) is to return a dict mapping odbs to a set of objects

dvc/dependency/repo.py
```diff
@@ -126,3 +92,94 @@ def changed_checksum(self):
...
+            locked=locked, cache_dir=local_odb.cache_dir
+        ) as repo:
+            used_objs = defaultdict(set)
+            rev = repo.get_rev()
+            if locked and self.def_repo.get(self.PARAM_REV_LOCK) is None:
+                self.def_repo[self.PARAM_REV_LOCK] = rev
+
+            path_info = PathInfo(repo.root_dir) / str(self.def_path)
+            try:
+                for odb, objs in repo.used_objs(
```
pmrowla, line 116: This ends up making us support #3305 - if there's a chained import here, we will end up recursively calling `dep.get_used_objs` for each repo in the chain.

dvc/objects/db/base.py
```diff
@@ -19,17 +19,33 @@ class ObjectDB:
     DEFAULT_CACHE_TYPES = ["copy"]
     CACHE_MODE: Optional[int] = None
 
-    def __init__(self, fs, path_info, **config):
+    def __init__(self, fs, path_info, tmp_dir, **config):
```
pmrowla, line 22: `tmp_dir` isn't actually used anywhere in ODB yet, but it will be once the current remote index is moved out of Remote into ODB.

In the meantime, this is needed to be able to construct a Remote object from an ODB instance.

dvc/repo/fetch.py
```diff
@@ -85,44 +102,51 @@ def fetch(
     return downloaded
 
 
-def _fetch_external(self, repo_url, repo_rev, files, jobs):
-    from dvc.external_repo import external_repo
-    from dvc.objects import save
-    from dvc.objects.stage import stage
-    from dvc.path_info import PathInfo
-    from dvc.scm.base import CloneError
+def _fetch_naive_objs(repo, objs, **kwargs):
```
pmrowla, line 105: Once the remote push/pull are unified into ODB save, we will be able to get rid of these separate fetch helpers

dvc/remote/base.py
```diff
@@ -86,6 +86,21 @@ def __repr__(self):
...
         )
 
+    def __eq__(self, other):
+        return self.odb == other.odb
+
+    def __hash__(self):
+        return hash(self.odb)
+
+    @classmethod
+    def from_odb(cls, odb):
```
pmrowla, line 96: This is a temporary workaround until the push/pull/save unification is completed (since we will no longer need Remote)

dvc/repo/status.py
```diff
@@ -91,22 +91,26 @@ def _cloud_status(
...
-    status_info = self.cloud.status(
-        used_objs, jobs, remote=remote, log_missing=False
-    )
-    for info in status_info.values():
-        name = info["name"]
-        status_ = info["status"]
-        if status_ == cloud.STATUS_OK:
+    for odb, objs in used.items():
+        if odb is not None:
+            # ignore imported objects
```
pmrowla, line 96: As discussed, we may want to start reporting status for imports later, but for now we can just skip any objects that come from non-default (None) odbs

dvc/objects/db/base.py
```diff
@@ -29,6 +29,23 @@ def __init__(self, fs, path_info, **config):
         self.cache_types = config.get("type") or copy(self.DEFAULT_CACHE_TYPES)
         self.cache_type_confirmed = False
         self.slow_link_warning = config.get("slow_link_warning", True)
+        self.tmp_dir = config.get("tmp_dir")
```
pmrowla, line 32: tmp_dir isn't actually used anywhere in ODB yet, but it will be once the current remote index is moved out of Remote into ODB.

In the meantime, this is needed to be able to construct a Remote object from an ODB instance.

dvc/repo/fetch.py
```diff
@@ -60,22 +60,39 @@ def fetch(
...
+                jobs=jobs,
+                remote=remote,
+                show_checksums=show_checksums,
+            )
+            downloaded += d
+            failed += f
+        elif isinstance(odb, GitObjectDB):
+            # objs contains staged import objects which should be saved
+            # last (after all other objects have been pulled)
+            external.update(objs)
```
efiop, line 82: Is this only to keep previous behavior or is there some additional reason to keep fetching these last?

dvc/repo/fetch.py
efiop: Just  to clarify, you mean that all three will go away, meaning that default odb, custom odb(e.g. for externals, or future per-output remotes) and git odb will all have the same fetching logic, right?

dvc/objects/db/git.py
```diff
@@ -0,0 +1,24 @@
...
+        super().__init__(fs, path_info)
+
+    def get(self, hash_info):
+        raise NotImplementedError
+
+    def add(self, path_info, fs, hash_info, move=True, **kwargs):
+        raise NotImplementedError
+
+    def gc(self, used, jobs=None):
+        raise NotImplementedError
```
efiop, line 24: So currently for git files we go the `stage() + save()` way for simplicity, but how do you see this odb in the future? Right now we want md5s from git files, but git odb can provide us with sha*s (and even then there are some details to it), so we'll need some translation layer? Or is there a more elegant way?

dvc/dependency/repo.py
```diff
@@ -85,39 +62,28 @@ def dumpd(self):
...
-                repo.fetch([path_info.fspath], jobs=jobs, recursive=True)
-            except (NoOutputOrStageError, NoRemoteError):
-                pass
-            obj = stage(
-                odb, path_info, repo.repo_fs, odb.fs.PARAM_CHECKSUM, jobs=jobs
-            )
-            save(odb, obj, jobs=jobs)
+        for odb, objs in self.get_used_objs().items():
+            if not isinstance(odb, GitObjectDB):
+                fetch_from_odb(self.repo, odb, objs, jobs=jobs)
```
efiop, line 71: Related to https://github.com/iterative/dvc/pull/6109/files#r649601307

dvc/dependency/repo.py
```diff
@@ -85,39 +62,28 @@ def dumpd(self):
...
 
+        obj = self.get_obj()
+        save(self.repo.odb.local, obj, jobs=jobs)
```
efiop, line 74: I guess this `save()` would not have been needed if `fetch_from_odb` would work `GitObjectDB`, so when `save`&`transfer` get unified, we'll just have (potentially) smth like `fetch(odb, objs)`, right? Just making sure I'm on the same page :)

dvc/objects/db/git.py
efiop: Or maybe we just need to convert them (aka stage()&save() like we do right now) on-the-fly from GitFS to LocalODB, so that we just use the resulting local objects right away? Since that is essentially the translation layer and once these files are converted into our md5-based objs, they are no longer in real git odb (i mean real-real git odb)

dvc/repo/fetch.py
pmrowla: This is to preserve the current fetch optimization for imported DVC objects and should be able to go away once pull/save are unified.

If we just call save() on the GitODB objects right now, we will end up streaming/downloading any imported DVC-tracked files, during the save(), but as we discussed it would be better to use fetch/pull to get all of the DVC-tracked files for now.

dvc/repo/fetch.py
pmrowla: Yes that's correct

dvc/dependency/repo.py
pmrowla: Yes this is correct, eventually we should just have the single odb-based fetch that works for all of our object/odb types

dvc/objects/db/git.py
pmrowla: Yeah I've been thinking about what we can do here, since git files won't be structured like a regular DVC ODB.

In theory though, if we move to using SHA's for git imports (or in general eventually), we could do an complete ODB implementation for translating git blob storage into DVC object storage. And lookups like hashes_exist could be done directly through git instead of using the fs walk/find method we do for other ODBs.

efiop:
For the record: had a meeting and decided to take a closer look at GitObjectDB, maybe we instead need a staging (aka temporary) db to keep git objects in and then just transfer them to local cache same way that we do that with imports.

Also, @pmrowla brought an idea that maybe we need to introduce smth like `read-only` flag for odbs, so we could use that to skip pushing to them. E.g. to avoid pushing imports.

pmrowla:
Staging ODB will addressed in a follow-up PR

jorgeorpinel:
I understand this implies that chained imports are now possible per https://github.com/iterative/dvc/issues/3305#issuecomment-871048690 🎉 If so, it may require some docs updates depending on the answer to https://github.com/iterative/dvc/issues/3305#issuecomment-871052955.

Issue #4420, opened 2020-08-19
unable to open database on mapped network drive


timbp:
I just installed dvc. My repository is on a mapped network drive, and dvc cannot open the database file. It seems to work perfectly if the repository is on the local C: drive. for the example below, I just made a repositiory on my home network storage, which is mapped to the Z: drive.

```
timbp@Eucalypt MINGW64 /z/MyVideos
$ git init
Initialized empty Git repository in //192.168.29.3/MyLibrary/MyVideos/.git/

timbp@Eucalypt MINGW64 /z/MyVideos (master)
$ dvc init

You can now commit the changes to git.

+---------------------------------------------------------------------+
|                                                                     |
|        DVC has enabled anonymous aggregate usage analytics.         |
|     Read the analytics documentation (and how to opt-out) here:     |
|              https://dvc.org/doc/user-guide/analytics               |
|                                                                     |
+---------------------------------------------------------------------+

What's next?
------------
- Check out the documentation: https://dvc.org/doc
- Get help and share ideas: https://dvc.org/chat
- Star us on GitHub: https://github.com/iterative/dvc

timbp@Eucalypt MINGW64 /z/MyVideos (master)
$ git status
On branch master

No commits yet

Changes to be committed:
  (use "git rm --cached <file>..." to unstage)
        new file:   .dvc/.gitignore
        new file:   .dvc/config
        new file:   .dvc/plots/confusion.json
        new file:   .dvc/plots/default.json
        new file:   .dvc/plots/scatter.json
        new file:   .dvc/plots/smooth.json

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        chasings.mp4

timbp@Eucalypt MINGW64 /z/MyVideos (master)
$ git commit -m "testing"
[master (root-commit) 1f5f95a] testing
 6 files changed, 128 insertions(+)
 create mode 100644 .dvc/.gitignore
 create mode 100644 .dvc/config
 create mode 100644 .dvc/plots/confusion.json
 create mode 100644 .dvc/plots/default.json
 create mode 100644 .dvc/plots/scatter.json
 create mode 100644 .dvc/plots/smooth.json

timbp@Eucalypt MINGW64 /z/MyVideos (master)
$ dvc add chasings.mp4
ERROR: unexpected error - unable to open database file

Having any troubles? Hit us up at https://dvc.org/support, we are always happy to help!

timbp@Eucalypt MINGW64 /z/MyVideos (master)
$ dvc status
[...]

Issue #5562, opened 2021-03-07
dvc: QA local/nfs/cifs/overlay/etc


efiop:
When working with local repos, dvc uses a lot of `exists()` and `stat()` calls, which is fine with normal filesystems, but could be extremely slow with filesystems like nfs/cifs/etc (e.g. `stat()` can be 100x slower). We should look into conserving such calls, similar to what we do with remotes, where we know that an api call to s3/gdrive/etc is pretty slow, so we try to do as little calls as possible.

* https://discord.com/channels/485586884165107732/728693131557732403/818168922393673738

Another issue with nfs/cifs/overlay is problems with sqlite database (e.g. freezeing locks) that we use for local optimization https://github.com/iterative/dvc/issues/4420 . Dvc itself is also often affected by it, which is why we provide an option to use `flufl.lock` as repo lock instead of flock-based lock. For sqlite it is not an option, so a solution might be to place it somewhere on a normal filesystem (e.g. /tmp is usually fine).

jorgeorpinel:
Idk if this is the right issue or if we need another one but users are reporting DVC 2.0 to be very slow with basic operations. Sounds like something to prioritize perhaps?

Also @dberenbaum @efiop @shcheklein WDYT about establishing some non-functional requirements such as basic operation speeds and designing some way to test them at least for major and minor releases (no patches)? Thanks

> Sorry, that should definitely be a separate issue.

efiop:
@jorgeorpinel Just for the record: Both links are not related to this issue.

> Also @dberenbaum @efiop @shcheklein WDYT about establishing some non-functional requirements such as basic operation speeds and designing some way to test them at least for major and minor releases (no patches)? Thanks

We have dvc-bench that tests scenarios end-to-end, we are filling it with stuff from time to time.

brbarkley:
Looking forward to this feature...

https://discord.com/channels/485586884165107732/563406153334128681/821510730565025823

brbarkley:
Any update on this issue? I see it's been delayed a few times.

efiop:
Hey @brbarkley , thanks for the interest! We are currently actively working on improving the internals in order for them to accommodate the optimizations that will be the result of this QA. So far ETA for the active phase of this ticket is around the beginning of June.

shcheklein:
For the record, from Discord:

[...]

PR #6111, opened 2021-06-04
Allow user to specify different state and index tmp dir locations, re: #4420

agurtovoy:
This PR introduces two new config options, `state.dir` and `index.dir`, which allow user to override state's cache and remote's index files location, correspondingly. The primary use case is to be able to move SQLite-based caches/indexes outside of the repository when the repository is located on a mounted volume (#4420), e.g. to be able to say something like

```sh
dvc config state.dir /tmp
dvc config index.dir /tmp
```

I debated the option names as well as whether this should be two or just one option, and am not fully sold on what I ended up with, so please treat this PR as it currently stands as a conversation starter. I'm willing to do the legwork of renaming/consolidating the config options, adding tests, and tweaking the docs as needed, or you can take over from here :)

### DIFF
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -119,32 +119,33 @@ class Repo:
 
         return root_dir, dvc_dir, tmp_dir
 
-    def _get_subsystem_tmp_dir(
-        self, subsystem, root_dir, tmp_dir, config: dict
-    ):
-        # NOTE: by default, store remote indexes and state's `links` and
-        # `md5s` caches in the repository itself to avoid any possible state
-        # corruption in 'shared cache dir' scenario, but allow user to
-        # override this through config when, say, the repository is located
-        # on a mounted volume — see
+    def _get_database_dir(self, db_name):
+        # NOTE: by default, store SQLite-based remote indexes and state's
+        # `links` and `md5s` caches in the repository itself to avoid any
+        # possible state corruption in 'shared cache dir' scenario, but allow
+        # user to override this through config when, say, the repository is
+        # located on a mounted volume — see
         # https://github.com/iterative/dvc/issues/4420
-        base_state_dir = config.get(subsystem, {}).get("dir", None)
-        if not base_state_dir:
-            return tmp_dir
+        base_db_dir = self.config.get(db_name, {}).get("dir", None)
+        if not base_db_dir:
+            return self.tmp_dir
 
         import hashlib
 
         from dvc.utils.fs import makedirs
 
-        root_dir_hash = hashlib.sha224(root_dir.encode("utf-8")).hexdigest()
-        state_tmp_dir = os.path.join(
-            base_state_dir,
+        root_dir_hash = hashlib.sha224(
+            self.root_dir.encode("utf-8")
+        ).hexdigest()
+
+        db_dir = os.path.join(
+            base_db_dir,
             self.DVC_DIR,
-            f"{os.path.basename(root_dir)}-{root_dir_hash[0:7]}",
+            f"{os.path.basename(self.root_dir)}-{root_dir_hash[0:7]}",
         )
 
-        makedirs(state_tmp_dir, exist_ok=True)
-        return state_tmp_dir
+        makedirs(db_dir, exist_ok=True)
+        return db_dir
 
     def __init__(
         self,
@@ -208,10 +209,8 @@ class Repo:
                 friendly=True,
             )
 
-            state_tmp_dir = self._get_subsystem_tmp_dir(
-                "state", self.root_dir, self.tmp_dir, self.config
-            )
-            self.state = State(self.root_dir, state_tmp_dir, self.dvcignore)
+            state_db_dir = self._get_database_dir("state")
+            self.state = State(self.root_dir, state_db_dir, self.dvcignore)
             self.odb = ODBManager(self)
 
             self.stage_cache = StageCache(self)
@@ -470,10 +469,8 @@ class Repo:
         return RepoFileSystem(self, subrepos=self.subrepos, **self._fs_conf)
 
     @cached_property
-    def index_tmp_dir(self):
-        return self._get_subsystem_tmp_dir(
-            "index", self.root_dir, self.tmp_dir, self.config
-        )
+    def index_db_dir(self):
+        return self._get_database_dir("index")
 
     @contextmanager
     def open_by_relpath(self, path, remote=None, mode="r", encoding=None):
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -119,6 +119,33 @@ class Repo:
 
         return root_dir, dvc_dir, tmp_dir
 
+    def _get_subsystem_tmp_dir(
+        self, subsystem, root_dir, tmp_dir, config: dict
+    ):
+        # NOTE: by default, store remote indexes and state's `links` and
+        # `md5s` caches in the repository itself to avoid any possible state
+        # corruption in 'shared cache dir' scenario, but allow user to
+        # override this through config when, say, the repository is located
+        # on a mounted volume — see
+        # https://github.com/iterative/dvc/issues/4420
+        base_state_dir = config.get(subsystem, {}).get("dir", None)
+        if not base_state_dir:
+            return tmp_dir
+
+        import hashlib
+
+        from dvc.utils.fs import makedirs
+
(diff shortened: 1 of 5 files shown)


agurtovoy:
Related umbrella issue: https://github.com/iterative/dvc/issues/5562

jmblackmer:
@agurtovoy How will this work if the repository and local storage cache are both located on NFS that is mounted on multiple machines?

agurtovoy:
@jmblackmer Good question. For what I understand, both state's cache and remote's index files are optional optimizations, so in theory it should be okay for every DVC client to have its own copy, but please don't quote me on that, haha.

efiop:
Hi @agurtovoy . Thanks for the PR! In regards to this issue, we were also considering keeping such files in /tmp(or in ~/.cache maybe or elsewhere) automatically, similar to tools like pre-commit. `/tmp` seems to be a safer choice, as many tools rely on it existing on a full-featured fs, but I'm not 100% sure about the best practices there and if our use case fits it.

Being able to at least configure the dir is a very useful feature that will unblock this whole scenario, but I'm wondering if there is a way to make it more seamless. Feel free to share your thoughts on this. :pray:

dvc/repo/__init__.py
```diff
@@ -184,10 +211,10 @@ def __init__(
                 friendly=True,
             )
 
-            # NOTE: storing state and link_state in the repository itself to
-            # avoid any possible state corruption in 'shared cache dir'
-            # scenario.
-            self.state = State(self.root_dir, self.tmp_dir, self.dvcignore)
+            state_tmp_dir = self._get_subsystem_tmp_dir(
+                "state", self.root_dir, self.tmp_dir, self.config
```
pared, line 215: In both use cases the only actual argument we provide is `state` or `index`, rest is repo class variables. Do we need to explicitly pass them?

dvc/repo/__init__.py
```diff
@@ -122,6 +122,33 @@ def _get_repo_dirs(
 
         return root_dir, dvc_dir, tmp_dir
 
+    def _get_subsystem_tmp_dir(
+        self, subsystem, root_dir, tmp_dir, config: dict
+    ):
```
pared, line 127: How about `_get_database_dir(db_name)`?
`state` and `index` are essentialy databases identified in config by their name. 
I think that `tmp` can be skipped, since from db point of view we don't care if its tmp or not. tmp is a naming convention that can be used by users to deduct what can be removed in project without breaking it severly.

pared (changes requested):
I like the idea with hashes - nice way to avoid project naming collisions.
The change looks good, though I would consider adding test for that. @agurtovoy would you be willing to write one?

agurtovoy:
Hey @efiop, I think when everything is local, co-locating these caches inside of the repo make sense; people do rename and move around repo folders, and doing so will bust the caches if they are not co-located. On the other hand, as a user, I haven't noticed a negative impact of busting a cache, maybe because our repos are still relatively small. Still, I assume that you introduced these caches for a reason, so personally I'd be inclined to not bust them for innocent scenarios like renaming/moving the repo locally.

At the same time, repos are put on mounted/network volumes frequently enough that a solution that just makes DVC work in that situation will likely save you a lot of support time. 

Personally, I'd go that route: detect when a database cache can't be co-located and move it to, say, `~/.cache`, but I'd leave the config option(s) there and have them override the default behavior when present. I used `/tmp` in my example above because our use case comes from running DVC in an AWS Lambda instance, and `/tmp` is the only local writeable dir we get, but `~/.cache` would probably be more idiomatic as the default fallback. I'd also consider having an option to disable these caches altogether to allow running DVC on a read-only local FS.

Hope this helps!

dvc/repo/__init__.py
agurtovoy: Good point, we don't.

efiop:
I guess another way to go about it is to have just 1 config option for tmp_dir itself, since it is unusual to want to have state and indexes in separate dirs. We might introduce a global/system cache for indexes in the future, but it will likely have a very different format anyway. I'm a bit hesitant to introduce 2 config options, as that seems excessive and will require introducing new options if we ever decide to use another sqlite db for something else.

Another interesting fact is that we have a special config option for hardlink-based locks (`core.hardlink_lock`) that we recommend using on funcy filesystems that don't support `flock`. And those locks are also located in `.dvc/tmp`, so if we introduce a config option for moving it we'll also no longer need `hardlink_lock` anymore (not saying that we should remove it right now, just saying that we could then consider it in the future).

So how about we simplify this PR and just introduce one config option to relocate `.dvc/tmp` and call it something like `repo_tmp_dir`(instead of `tmp_dir`, just hesitant to take such a general name for this)? WDYT?

jmblackmer:
@agurtovoy @efiop I see that this hasn't had much movement in the last few weeks. Is there another approach that is being taken or is this just being held up by the failed tests at this point?

agurtovoy:
@jmblackmer I got swamped and didn't have time to look into @efiop's suggestion of one config option for `tmp_dir` yet. IIRC, that's actually how I started with my original PR, but got concerned about how many different files it affected and decided to go with something more localized; having said that, I think it definitely deserves a second look. Given my limited availability for the next couple of weeks, I was half-hoping the DVC team could take this one.

pared:
@agurtovoy thanks for heads up! Usually we give users as much time as they need, and close if we see no activity in months. In situation like this we might be able to take over. Added to next sprint for evaluation.

jmblackmer:
Hi @efiop,

I see that this has been In Progress for a while. Do you have an idea of when this might be available for use?

efiop:
@jmblackmer Hi. Thank you for your interest! Unfortunately, other stuff gets in the way :slightly_frowning_face:  Ideally, I would love for someone to take a closer look here and reevaluate the approach. Config option is ok, but it would be great if we could just work by default, without making the user set this explicit option. Thinking that maybe moving `.dvc/tmp` permanently to `/tmp` could be a good idea, but I didn't spend enough time on it to be absolutely sure about it in general and about particular details. We have an upcoming NFS/etc story in our fs epic, we will definitely get to this then (prob a matter of weeks). If anyone has some good examples of other tools doing something similar, please feel free to share.

efiop:
@jmblackmer Btw, could you elaborate on your scenario? Maybe I could suggest some workaround (e.g. keeping the repo on normal fs but using nfs as cache).

jmblackmer:
@efiop My team works with data with a high number of restrictions, so the data has to stay on the NFS. The NFS is mounted to two HPC clusters (where we won't have access to /tmp) and a more general development server specifically for our team. In order to avoid moving the code around to different clusters (which would also require 3 caches, but I'm not sure that matters), most of our code lives next to the data that it works with and is backed up to our gitlab server. I'm trying to move away from that, but I'll need the automation support it and this is part of that automation.

efiop:
@jmblackmer So you are using literally the same git/dvc repo instance that stays on nfs? Or does every team member has it own clone of that repo? In the first case, repo locking is going to be tricky and I would not recommend using it like that. In the second case `/tmp` will be unique per every machine, so as long as there is only one guy working on his repo instance - it should be fine. I guess this is a good example where we indeed only need to move state and index somewhere outside, as other locks should be able to function properly on nfs and are actually needed in a shared repo instance scenario. I guess we might just want to clean up and merge this particular PR after all...

jmblackmer:
@efiop It's a small team with everyone working on different projects, so there is very little, if any, overlap of work within repos. Almost no merging and git acts mostly as a checkpoint system. The only overlap is between the data engineer, who gathers and pre-processes data in one project, and the data scientists, who take the output of the previous project and build their models off the data. We're hoping to use DVC with a shared cache to help move data between projects and to limit the amount of storage used on NFS.

jmblackmer:
@efiop I've only done proofs of concept with DVC, so I haven't been offering much input, but allowing us to specify the cache location would allow us to specify ~, which is probably the only location outside of the nfs where we would have write permissions on all 3 machines. Like you said, fixing the bug is better than working around it, but until we know how to do that, this branch would at least allow us to start using DVC.

gro1m:
@efiop @jmblackmer As also stated in the Discord chat this highly important for us and we made a package with this change (with resolved conflicts) and the PR #6359. We could also make this available in a PR here, but somebody would need to review and maybe make the changes needed to only use *one* directory or whatever... If that is of any help, please tell us.
 FYI: @meierale,  @nueschs

Issue #6106, opened 2021-06-03
list: fails if a DVC-tracked directory is empty but present


shcheklein:
# Bug Report

## Description

```
(.env) √ Projects/get-started-experiments % dvc list . data/fashion-mnist/prepared
ERROR: failed to list '.' - The path 'data/fashion-mnist/prepared' does not exist in the target repository
'/Users/ivan/Projects/get-started-experiments' neither as a DVC output nor as a Git-tracked file.
```

even though, this path does exist, and it is being tracked by DVC:

```yaml
schema: '2.0'
stages:
  prepare:
    cmd: python3 src/prepare.py
    deps:
    - path: data/fashion-mnist/raw/
      md5: 8a31b1338b609cf1133d1df639de606c.dir
      size: 30878645
      nfiles: 4
    - path: src/prepare.py
      md5: 2c9ddd60ad065771628c58c618663050
      size: 3881
    params:
      params.yaml:
        prepare.remix: false
        prepare.remix_split: 0.2
        prepare.seed: 20210428
    outs:
    - path: data/fashion-mnist/prepared
      md5: c5095cd3e98121bc8d2bebe27d702ab5.dir
      size: 54951020
      nfiles: 2
...
```

### Reproduce

```
git clone git@github.com:iterative/get-started-experiments.git
cd get-started-experiments
mkdir data/fashion-mnist/prepared
dvc list . data/fashion-mnist/prepared
```

gives:

> ERROR: failed to list '.' - The path 'data/fashion-mnist/prepared' does not exist in the target repository '/Users/ivan/Projects/test-bug' neither as a DVC output nor as a Git-tracked file.

### Expected

To see `prepared`.

### Environment information

```
(.env) √ Projects/test-bug % dvc version
DVC version: 2.3.0+eb46c0
---------------------------------
Platform: Python 3.8.9 on macOS-10.15.6-x86_64-i386-64bit
Supports: All remotes
Cache types: <https://error.dvc.org/no-dvc-cache>
Caches: local
Remotes: https
Workspace directory: apfs on /dev/disk1s1
Repo: dvc, git
```

dberenbaum:
@shcheklein could you give some context on priority?

shcheklein:
[...]

PR #6120, opened 2021-06-05
dvc ls: not raise PathMissingError on empty dir.

### LINKED ISSUE #5841: dvc list: Error on empty directory.
# Bug Report

Got error message on an empty directory, shouldn't it show nothing? like ls command.

## Description
image
Error when list a empty path, strange behavior.
Might relate to https://github.com/iterative/dvc/blob/daf07451f8e8f3e76a791c696b0ea175e8ed3ac1/dvc/repo/ls.py#L40-L41

### Reproduce

1. git init
2. dvc init
3. mkdir empty
4. dvc list . empty

### Expected
Show nothing like ls command
image

### Environment information
DVC version: 2.0.17+7e4851
Platform: Python 3.8.8 on macOS-10.16-x86_64-i386-64bit
Supports: All remotes
Caches: local
Remotes: None
Repo: dvc, git

**Output of `dvc doctor`:**

```console
$ dvc doctor
```

karajan1001:
fix #5841

### DIFF
--- a/tests/func/test_ls.py
+++ b/tests/func/test_ls.py
@@ -150,9 +150,20 @@ def test_ls_repo_with_path_dir_dvc_only_empty(tmp_dir, dvc, scm):
     tmp_dir.scm_gen(FS_STRUCTURE, commit="init")
     tmp_dir.dvc_gen(DVC_STRUCTURE, commit="dvc")
     tmp_dir.scm_gen({"folder/.keep": "content"}, commit="add .keep")
+    tmp_dir.scm_gen({"empty_scm_folder/": {}}, commit="add scm empty")
+    tmp_dir.dvc_gen({"empty_dvc_folder": {}}, commit="empty dvc folder")
 
     with pytest.raises(PathMissingError):
-        Repo.ls(os.fspath(tmp_dir), path="folder", dvc_only=True)
+        Repo.ls(os.fspath(tmp_dir), path="not_exist_folder")
+
+    assert Repo.ls(os.fspath(tmp_dir), path="empty_scm_folder") == []
+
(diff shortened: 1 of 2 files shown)


tests/func/test_ls.py
```diff
@@ -139,8 +139,8 @@ def test_ls_repo_with_path_dir_dvc_only_empty(tmp_dir, dvc, scm):
     tmp_dir.dvc_gen(DVC_STRUCTURE, commit="dvc")
     tmp_dir.scm_gen({"folder/.keep": "content"}, commit="add .keep")
 
-    with pytest.raises(PathMissingError):
-        Repo.ls(os.fspath(tmp_dir), path="folder", dvc_only=True)
+    files = Repo.ls(os.fspath(tmp_dir), path="folder", dvc_only=True)
```
efiop, line 142: Discussed with @karajan1001 that this fixes https://github.com/iterative/dvc/issues/6106 but #5841 is actually not a bug, but an intended behaviour as `folder` in this test doesn't exist from DvcFileSystem. So a proper solution here is to only use DvcFileSystem in case of `dvc_only=True` instead of using `RepoFileSystem` like we do right now.

karajan1001: @efiop Doesn't with `--dvc-only` flag should this exception been thrown out? in #5841 no `--dvc-only` flag on it.

efiop:
@dberenbaum Even earlier. After https://github.com/iterative/dvc/issues/7384 Thanks for the heads up!

dberenbaum:
@efiop Is this one ready for review now?

efiop (approved):

PR #6134, opened 2021-06-08
stage: monitor: don't forget to unprotect outs

### LINKED ISSUE #6083: checkpoint: write permission denied
Surely `checkpoint: true` should un-protect hardlink/symlinks?

```python
$ git clone https://github.com/iterative/dvc-checkpoints-mnist && cd dvc-checkpoints-mnist \
  && pip install -r requirements.txt && dvc exp run
...
...
  File "train.py", line 99, in main
    torch.save(model.state_dict(), "model.pt")
...
PermissionError: [Errno 13] Permission denied: 'model.pt'
$ ls -l model.pt | cut -d' ' -f1
-r--r--r--
```
```bash
$ dvc doctor
DVC version: 2.1.0 (conda)
Platform: Python 3.8.10 on Linux-4.15.0-58-generic-x86_64-with-glibc2.10
Supports: http, https
Caches: local
Remotes: None
Repo: dvc, git
```

efiop:
Fixes #6083

### DIFF
--- a/dvc/stage/monitor.py
+++ b/dvc/stage/monitor.py
@@ -75,6 +75,7 @@ class CheckpointTask(MonitorTask):
     def _run_callback(stage, callback_func):
         stage.save(allow_missing=True)
         stage.commit(allow_missing=True)
+        stage.unprotect_outs()
         logger.debug("Running checkpoint callback for stage '%s'", stage)
         callback_func()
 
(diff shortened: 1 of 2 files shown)


dvc/stage/monitor.py
```diff
@@ -75,6 +75,7 @@ def __init__(
     def _run_callback(stage, callback_func):
         stage.save(allow_missing=True)
         stage.commit(allow_missing=True)
+        stage.unprotect_outs()
```
efiop, line 78: A better approach here would be to have a method that could just save the outputs to cache without touching the originals, kinda like our `--to-cache/--to-remote` stuff. That would save us some unnecessary moving+linking+copying that we have right now and would turn that into just `copying`. Looks like it might be a special option for `commit` that would tell it to only `copy`. Might be okay to leave it as is for now though.

pared (approved):

PR #6137, opened 2021-06-08
setup: bump fsspec/s3fs/gcsfs to 2021.06.0

isidentical:


isidentical:
It seems like there are a lot of failures, due to possibly some changes on the `LocalFileSystem` on upstream. Looking into it :eyes:

isidentical:
pre-requisite: https://github.com/intake/filesystem_spec/issues/666

isidentical:
Okay, everything is passing now. Though we need a new release.

PR #6151, opened 2021-06-10
s3, gs, azure: re-enable prefix-based search optimization

### LINKED ISSUE #6089: Old bug (dvc pull from S3 and md5 starts with `00`) reappears with latest release

isidentical:
It was turned off due to not all providers supporting `find(<something>, prefix=<something>)` but now it is safe to turn that back on. Resolves #6089.

### DIFF
--- a/dvc/fs/fsspec_wrapper.py
+++ b/dvc/fs/fsspec_wrapper.py
@@ -96,7 +96,8 @@ class FSSpecWrapper(BaseFileSystem):
         files = self.fs.ls(path, detail=detail)
         yield from self._strip_buckets(files, detail=detail)
 
-    def find(self, path_info, detail=False):
+    # pylint: disable=unused-argument
+    def find(self, path_info, detail=False, prefix=None):
         path = self._with_bucket(path_info)
         files = self.fs.find(path, detail=detail)
         if detail:
@@ -105,7 +106,7 @@ class FSSpecWrapper(BaseFileSystem):
         yield from self._strip_buckets(files, detail=detail)
 
     def walk_files(self, path_info, **kwargs):
-        for file in self.find(path_info):
+        for file in self.find(path_info, **kwargs):
             yield path_info.replace(path=file)
 
     def remove(self, path_info):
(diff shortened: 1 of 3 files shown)


dvc/fs/fsspec_wrapper.py
```diff
@@ -169,9 +172,16 @@ def _isdir(self, path_info):
             and entry["name"].endswith("/")
         )
 
-    def find(self, path_info, detail=False):
-        path = self._with_bucket(path_info)
-        files = self.fs.find(path, detail=detail)
+    def find(self, path_info, detail=False, prefix=None):
```
isidentical, line 175: It might be a nice idea (for future refactors of remotes) to pass the prefix in here (`0/`), and use the actual path (`bucket/cache/00/`) as the path_info (to be fully fsspec-compliant).

dvc/fs/fsspec_wrapper.py
```diff
@@ -96,7 +96,8 @@ def ls(self, path_info, detail=False):
         files = self.fs.ls(path, detail=detail)
         yield from self._strip_buckets(files, detail=detail)
 
-    def find(self, path_info, detail=False):
+    # pylint: disable=unused-argument
+    def find(self, path_info, detail=False, prefix=None):
```
isidentical, line 100: I don't really know if there is a scenario where the `prefix` will not be `None` for this case (TRAVERSE_PREFIX_LEN=2, used by filesystems with real dirs etc) but if there is no such case, then perhaps we could add an `assert prefix is None` as a safe-guard.

dvc/fs/fsspec_wrapper.py
pmrowla: Yeah, it would make more sense for us to do
```
walk("bucket/00/", prefix="0")
```
instead of the current
```
walk("bucket/00/0", prefix=True)
```

pmrowla (approved):

PR #6162, opened 2021-06-14
add, import-url: preserve size for to-remote/to-cache

### LINKED ISSUE #6161: straight-to-remote: does not set size in the .dvc file
When I do `dvc add cat.jpg`, I get the following contents in `cat.jpg.dvc`:
```yaml
outs:
- md5: 4a94da5d72618376ab28d789300308da
  size: 2092576
  path: cat.jpg
```

And, when I do `dvc add --to-remote -r webdav remote://webdav-root/cat.jpg`, I get the following contents in `cat.jpg.dvc`:
```yaml
outs:
- md5: 4a94da5d72618376ab28d789300308da
  path: cat.jpg
  nfiles: 1
```

Both of those should create identical `.dvc` files.

isidentical:
Resolves #6161

### DIFF
--- a/dvc/utils/stream.py
+++ b/dvc/utils/stream.py
@@ -70,6 +70,7 @@ class HashedStreamReader(io.IOBase):
     def __init__(self, fobj):
         self.fobj = fobj
         self.md5 = hashlib.md5()
+        self.total_read = 0
         self.is_text_file = None
         super().__init__()
 
@@ -95,9 +96,12 @@ class HashedStreamReader(io.IOBase):
         else:
             data = chunk
         self.md5.update(data)
+        self.total_read += len(data)
 
         return chunk
 
     @property
     def hash_info(self):
-        return HashInfo(self.PARAM_CHECKSUM, self.md5.hexdigest(), nfiles=1)
+        return HashInfo(
+            self.PARAM_CHECKSUM, self.md5.hexdigest(), size=self.total_read
+        )
--- a/tests/func/test_add.py
+++ b/tests/func/test_add.py
@@ -1024,6 +1024,7 @@ def test_add_to_remote(tmp_dir, dvc, local_cloud, local_remote):
 
     hash_info = stage.outs[0].hash_info
     assert local_remote.hash_to_path_info(hash_info.value).read_text() == "foo"
+    assert hash_info.size == len("foo")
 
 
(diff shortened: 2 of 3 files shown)


dvc/utils/stream.py
```diff
@@ -70,6 +70,7 @@ class HashedStreamReader(io.IOBase):
     def __init__(self, fobj):
         self.fobj = fobj
         self.md5 = hashlib.md5()
+        self.total_read = 0
```
skshetry, line 73: We should probably extend `upload_fobj` with `size` for Webdav and HTTP servers as we want to avoid chunked encoding as much as possible (`PUT` does not have good support for `Transfer-Encoding: chunked`, see https://github.com/iterative/dvc/issues/4796). 

fsspec based file object already has a size attribute which we could use to pass that to `upload_fobj`.

isidentical: Would you mind extending it a bit?

skshetry: As WebDAV uses PUT for upload, there might be old servers that don't support chunked encoding. The same applies to HTTP. So, we try to avoid chunking upload as much as possible. But to do that, we need to set the `Content-Length` header with the expected file size. 

So, I was suggesting something like following:
```python
with fs.open(...) as fd:
    stream = HashedStreamReader(fd)
    size = getattr(fd, "size", None)
    odb.fs.upload_fobj(stream, size=size)
```

But looking at it right now, it seems to have been implemented already for the `pbar` args, but we could rephrase/repurpose it.
https://github.com/iterative/dvc/blob/5fb181fcf8b58e91268a51301bd6d43052921c76/dvc/objects/stage.py#L21-L23

isidentical: Ah, I see. Regarding the usage of `fd.size`, I see that it is not required in general so that is alright but just to keep in mind I have some ideas about making it lazy so that we don't make an additional API call every time (also it is only supported in AbstractBufferedFile's, so stuff like sshfs.SSHFile won't be supporting it).

skshetry (approved):

dvc/utils/stream.py
skshetry: Could you please share what you are thinking about sizes, only related to `sshfs` or in DVC itself? Anyway, I am moving with `size` stuff for now at I am considering it as broken.

Issue #5739, opened 2021-03-30
exp show: Allow running while DVC is locked


rogermparent:
While working on a way to consume `dvc exp show --show-json`'s output, we found that DVC being locked while running an experiment can pose an issue when trying to show the table, especially if a long-running experiment were to be started outside the repo and a user wanted to view it in VS Code.

Would it be possible to have DVC CLI not check for a lock on read-only commands like `exp show`? Doing so could be huge for applications that want to consume DVC's output like the VSCode extension.

#3783 may also be related in that is helps solve this sort of usage.

efiop:
@rogermparent Could you please elaborate? Some example would be great.

rogermparent:
> @rogermparent Could you please elaborate? Some example would be great.

Sure, any DVC project capable of `exp show` can serve as an example, but to be specific the project I encounter this issue on is an embedded `https://github.com/iterative/dvc-checkpoints-mnist` that we use to test the VSCode Extension against.

The extension uses `dvc exp show --show-json` to update the table, but this command can fail if it is run while another command, usually `dvc exp run`, is locking the repo. There are many ways around this on the VSCode Extension end, of course, but the question of "Why does a lock prevent a read-only command" has come up enough in discussion to warrant a ticket on the DVC CLI project, for clarification if nothing else.

There's also related discussion going on in #5750, some of it may actually be more appropriate for this Issue.

dberenbaum:
@rogermparent Is this an intermittent error, or is there some reproducible example you can provide of when this happens? I'm trying to reproduce it now but keep getting the output in #5750 instead of any kind of error. Also, do you have the error output?

pmrowla:
@dberenbaum this is an actual issue, but it will be difficult to reproduce manually due to the nature of the problem. The error  itself will be this message:
https://github.com/iterative/dvc/blob/02d9684c339e80f7f50a0a278720273b97d91598/dvc/lock.py#L19-L22

During `dvc exp run`, the repo will be unlocked while we run the user's pipeline commands. The repo *will* be completely locked while DVC is doing anything else (like writing git commits, managing exp refs, etc).

As far as actual CPU execution time goes, 99+% of the time will be spent executing the user's pipeline (giving the output in #5750). If you are trying to test it manually it's unlikely you will run `dvc exp show` at precisely the right time to trigger the problem raised in this issue. But if you wrote a script to constantly run `dvc exp show` in a loop, you would see this issue eventually.

rogermparent:
[...]

Issue #6050, opened 2021-05-24
exp show: show names for the queued experiments


aguschin:
It would be great to see the name of a queued experiment. Right now if we run, for example
```
dvc exp run --queue -S model.cnn.dropout=0.25 -n named-one
```
the table below won't show it's name right away:
<img width="1262" alt="image" src="https://user-images.githubusercontent.com/6797716/119343471-2aded280-bc9f-11eb-9325-d7a4e788986f.png">
only after we run it with `dvc exp run --run-all`:
<img width="1271" alt="image" src="https://user-images.githubusercontent.com/6797716/119343665-6ed1d780-bc9f-11eb-9982-389c5a4dd55f.png">

We could live fine without names, but if for example we want to review multiple queued experiments with `dvc exp show` and look for what we forgot to add to the queue, it would be helpful to see them.

Another use case which is not supported now, AFAK, is to remove queued experiments. If I added few experiments, I may want to remove one of them before running them all. Right now I need to reset all queued experiments with `dvc exp remove --queue` and add required ones again. This removal would be convenient to do if we could reference the queued experiments by their names. If it seems like a valid use case and dvc has no such feature, I can create a feature request.

dberenbaum:
#5965 also has some loosely related discussion about whether the final exp name should be shown in the queue from the beginning. @pmrowla what do you think? Not sure how important this is or how difficult to implement, but it does seems like it's a strict improvement on current behavior.

jorgeorpinel:
> If the user is specifying a name with -n/--name we can display it for queued experiments.

👍

dberenbaum:
> Another use case which is not supported now, AFAK, is to remove queued experiments. If I added few experiments, I may want to remove one of them before running them all. Right now I need to reset all queued experiments with `dvc exp remove --queue` and add required ones again. This removal would be convenient to do if we could reference the queued experiments by their names. If it seems like a valid use case and dvc has no such feature, I can create a feature request.

@aguschin This part got lost in the discussion, but see #6157.

@pmrowla I think it would be helpful to add this to the loose ends to clean up for experiments. It's not clear if or when we will have a separate `exp queue` command, or whether that will address this problem since the queue might still also show in `exp show`. How much work do you think it is to show named experiments in the queue?

pmrowla:
> @pmrowla How much work do you think it is to show named experiments in the queue?

@dberenbaum this is pretty much just a one liner UI change. We already have names for queued experiments internally, we just don't show them in the UI right now.

PR #6174, opened 2021-06-15
exp show: display running/queued state for experiments

### LINKED ISSUE #5965: exp show: include running experiment
Show currently running experiment in the table. cc @shcheklein @pmrowla

pmrowla:

Will close #5965

- For the DVC UI, this PR adds `State` column that will currently be Running, Queued or empty
    - If the table has no running or queued experiments the column will not be displayed
    - If `State` is `Running`, an additional `Executor` column will be displayed noting where the experiment is being run, currently this is limited to `local (workspace)` or `local (background)`
- For `--show-json` (VSCode), experiment dicts now have a `running` key that will be true or false, and an `executor` key which will be an optional string executor name
- Fixes bug where main repo would be locked during `--temp` or `--queue/--run-all` runs (related to #5739) which prevented `exp show` from being usable

Regular experiments example:
[asciicast](https://asciinema.org/a/bHRMvG8V7Oja8fuk8oVO27D6O)

Checkpoints example:
[asciicast](https://asciinema.org/a/1RVLUEvZBmAFxskL1C000SikI)

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -97,9 +97,16 @@ def _update_names(names, items):
 def _collect_names(all_experiments, **kwargs):
     metric_names = defaultdict(dict)
     param_names = defaultdict(dict)
+    include_state = False
+    include_executor = False
 
     for _, experiments in all_experiments.items():
         for exp in experiments.values():
+            if exp.get("running"):
+                include_state = True
+                include_executor = True
+            elif exp.get("queued"):
+                include_state = True
             _update_names(metric_names, exp.get("metrics", {}).items())
             _update_names(param_names, exp.get("params", {}).items())
     metric_names = _filter_names(
@@ -115,7 +122,7 @@ def _collect_names(all_experiments, **kwargs):
         kwargs.get("exclude_params"),
     )
 
-    return metric_names, param_names
+    return metric_names, param_names, include_state, include_executor
 
 
 experiment_types = {
@@ -136,6 +143,8 @@ def _collect_rows(
     precision=DEFAULT_PRECISION,
     sort_by=None,
     sort_order=None,
+    include_state=True,
+    include_executor=True,
 ):
     from dvc.scm.git import Git
 
@@ -155,7 +164,8 @@ def _collect_rows(
         elif exp.get("queued"):
             state = "Queued"
         else:
-            state = "-"
+            state = None
+        executor = exp.get("executor")
         is_baseline = rev == "baseline"
 
         if is_baseline:
@@ -197,8 +207,11 @@ def _collect_rows(
             typ,
             _format_time(exp.get("timestamp")),
             parent,
-            state,
         ]
+        if include_state:
+            row.append(state)
+        if include_executor:
+            row.append(executor)
         _extend_row(
             row, metric_names, exp.get("metrics", {}).items(), precision
         )
@@ -304,12 +317,18 @@ def experiments_table(
     sort_by=None,
     sort_order=None,
     precision=DEFAULT_PRECISION,
+    include_state=True,
+    include_executor=True,
 ) -> "TabularData":
     from funcy import lconcat
 
     from dvc.compare import TabularData
 
-    headers = ["Experiment", "rev", "typ", "Created", "parent", "State"]
+    headers = ["Experiment", "rev", "typ", "Created", "parent"]
+    if include_state:
+        headers.append("State")
+    if include_executor:
+        headers.append("Executor")
     td = TabularData(
         lconcat(headers, metric_headers, param_headers), fill_value=FILL_VALUE
     )
@@ -322,6 +341,8 @@ def experiments_table(
             sort_by=sort_by,
             sort_order=sort_order,
             precision=precision,
+            include_state=include_state,
+            include_executor=include_executor,
         )
         td.extend(rows)
 
@@ -361,7 +382,12 @@ def show_experiments(
     include_params = _parse_filter_list(kwargs.pop("include_params", []))
     exclude_params = _parse_filter_list(kwargs.pop("exclude_params", []))
 
-    metric_names, param_names = _collect_names(
+    (
+        metric_names,
+        param_names,
+        include_state,
+        include_executor,
+    ) = _collect_names(
         all_experiments,
         include_metrics=include_metrics,
         exclude_metrics=exclude_metrics,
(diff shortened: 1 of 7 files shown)


pmrowla:
Regarding docs, `exp show` usage examples need to be updated throughout the docs. I'm planning on adding a PR to display user-specified names for queued exps and will do a single docs PR w/the updated examples after that

dvc/command/experiments.py
```diff
@@ -150,7 +150,12 @@ def _collect_rows(
 
     new_checkpoint = True
     for i, (rev, exp) in enumerate(experiments.items()):
-        queued = str(exp.get("queued") or "")
+        if exp.get("running"):
+            state = "Run"
+        elif exp.get("queued"):
+            state = "Queue"
```
efiop, line 156: Should it be `Running`/`Queued`? Just feels a bit odd. Maybe this has been discussed somewhere. Just wondering.

pmrowla: Running/Queued makes more sense but I'm not really sure what's best here given that table width is a concern

dberenbaum (commented):
It looks good to me for non-checkpoint experiments.

For checkpoints experiments, it's a little confusing because:
1. The `Run` status shows next to the already completed checkpoint. Maybe we should be showing the experiment name on a separate line from individual checkpoints? That would also narrow the first column.
2. For workspace runs, the status is `Run` for both the workspace row and the experiment row (which only exists after the first checkpoint). So it can look like there are two experiments running, one of which got added after the first checkpoint. Is there a way to indicate that the workspace is the same as one of the experiments, at least while the experiment is running?

pmrowla:
@dberenbaum 

> 1. The `Run` status shows next to the already completed checkpoint. Maybe we should be showing the experiment name on a separate line from individual checkpoints? That would also narrow the first column.

What would we do here for regular experiments? When https://github.com/iterative/dvc/issues/6050 is done we will have names in the rows for these queued/running experiments, and will have the same issue. I don't think it makes sense to take up 2 rows for a single experiment.

> 2. For workspace runs, the status is `Run` for both the workspace row and the experiment row (which only exists after the first checkpoint). So it can look like there are two experiments running, one of which got added after the first checkpoint. Is there a way to indicate that the workspace is the same as one of the experiments, at least while the experiment is running?

It seems like what we are going to eventually need is to display information about where an experiment is running (for remote executors) but I'm not sure where to indicate that right now. We could omit `Run` entirely from the workspace row, and display the state for experiment rows as something like `Running (workspace)`/ `Running (background)` but this is going to keep making the table wider

dberenbaum:
> This makes sense to me because when you are resuming/continuing a checkpoint run, you are starting from the last completed checkpoint (rather than starting from the queued row that only has params and no metrics/outputs)

Extracted to #6194.

> It seems like what we are going to eventually need is to display information about where an experiment is running (for remote executors) but I'm not sure where to indicate that right now. We could omit `Run` entirely from the workspace row, and display the state for experiment rows as something like `Running (workspace)`/ `Running (background)` but this is going to keep making the table wider

Yes, we need a column like executor or location or something. Why not add it now? Maybe we can autohide it and the state column when it's empty (nothing is running)?

dvc/command/experiments.py
```diff
@@ -304,7 +309,7 @@ def experiments_table(
 
     from dvc.compare import TabularData
 
-    headers = ["Experiment", "rev", "queued", "typ", "Created", "parent"]
+    headers = ["Experiment", "rev", "typ", "Created", "parent", "State"]
```
skshetry, line 312: I am a bit confused about what it should be: `State` or a `Status`. Both seem to make sense to me. Is `State` most appropriate here?

pmrowla: I think either term works here. I just used `state` because `status` has some other connotations in DVC

pmrowla:
@dberenbaum I've added a separate `executor` column that will display where the experiment is running, and both `state` and `executor` are only shown if there is actually any running/queued experiments in the table. (see the 2 updated screen recordings in the top post)

dvc/command/experiments.py
```diff
@@ -299,12 +317,18 @@ def experiments_table(
...
     from funcy import lconcat
 
     from dvc.compare import TabularData
 
-    headers = ["Experiment", "rev", "queued", "typ", "Created", "parent"]
+    headers = ["Experiment", "rev", "typ", "Created", "parent"]
+    if include_state:
+        headers.append("State")
+    if include_executor:
+        headers.append("Executor")
```
skshetry, line 331: I wanted to get rid of these kinds of conditional checks in the table. And, it is not clear on what basis `include_state` and `include_executor` are set/works.

I'll suggest to get rid of these conditionals, and add the following in the header itself. 
```python
headers = ["Experiment", "rev", "typ", "Created", "parent", "State", "Executor"]
```

Also, get rid of all the `if` checks and keep everything in the table. Later you can simply drop them if they are empty (after [here](https://github.com/iterative/dvc/blob/61835f65116c3e5808c325d6cb22a12d36f70cde/dvc/command/experiments.py#L389)).

```python
if td.is_empty("State"):
    td.drop("State")
if td.is_empty("Executor"):
    td.drop("Executor")
```

And, in the implementation of `TabularData`, you could have something like this:
```python
def is_empty(self, name: str) -> bool:
    col = self.column(name)
    return not any(item != self._fill_value for item in col)
```

This should clean up the code here, and also make it safe and clear.

dberenbaum:
> I think that we are mixing two things in the `exp show` - `list` and `metrics/params show` (which is collapsed in non-pager mode). Would be great to revisit the table formats for 3.0 (especially considering remote executors might add more columns here). cc @dberenbaum

Do you mean that columns like status and executor should be shown in a separate command and not in `dvc exp show`?

dberenbaum:
@pmrowla Looks great!

One very minor comment: can we shorten `local (workspace)` and `local (background)` text to just `workspace` and `background` or `temp` (since that's the flag name) or similar? `local` seems redundant now, and even when we have remote executors, we can specify them by name.

dvc/command/experiments.py
```diff
@@ -381,16 +395,22 @@ def show_experiments(
     if no_timestamp:
         td.drop("Created")
 
+    for col in ("State", "Executor"):
```
dberenbaum, line 398: Is it possible for other columns to be empty? If so, would we want to hide them, too?

pmrowla: There shouldn't be any other columns in the table right now that can be empty

dberenbaum (approved):

PR #6175, opened 2021-06-15
import-url: compute the stage hash in --to-remote

### LINKED ISSUE #6166: straight-to-remote: missing `etag` or checksum on `import-url`
### Script to reproduce
```python
cd $(mktemp -d)
git init && dvc init
dvc remote add myremote $(mktemp -d) -f
dvc import-url https://data.dvc.org/get-started/data.xml data.xml \
    --to-remote -r myremote 
cat data.xml.dvc
dvc status
```

### Output
```console
$ cat data.xml.dvc
frozen: true
deps:
- path: https://data.dvc.org/get-started/data.xml
outs:
- md5: a304afb96060aad90176268345e10355
  nfiles: 1
  path: data.xml

$ dvc status
WARNING: stage: 'data.xml.dvc' is frozen. Its dependencies are not going to be shown in the status output.
data.xml.dvc:
        changed checksum
```

When importing locally, I do see the checksum, so it looks like the problem is only on straight-to-remote cases.

isidentical:
Resolves #6166

### DIFF
--- a/dvc/repo/imp_url.py
+++ b/dvc/repo/imp_url.py
@@ -76,6 +76,7 @@ def imp_url(
     elif to_remote:
         remote = self.cloud.get_remote(remote, "import-url")
         stage.outs[0].transfer(url, odb=remote.odb, jobs=jobs)
+        stage.md5 = stage.compute_md5()
     else:
         stage.run(jobs=jobs)
 
--- a/dvc/repo/imp_url.py
+++ b/dvc/repo/imp_url.py
@@ -76,6 +76,7 @@ def imp_url(
     elif to_remote:
         remote = self.cloud.get_remote(remote, "import-url")
         stage.outs[0].transfer(url, odb=remote.odb, jobs=jobs)
+        stage.save_deps()
         stage.md5 = stage.compute_md5()
     else:
         stage.run(jobs=jobs)
--- a/tests/func/test_import_url.py
+++ b/tests/func/test_import_url.py
@@ -274,6 +274,7 @@ def test_import_url_to_remote_single_file(
     url = "remote://workspace/foo"
     stage = dvc.imp_url(url, to_remote=True)
 
+    assert stage.deps[0].hash_info.value is not None
     assert not (tmp_dir / "foo").exists()
     assert (tmp_dir / "foo.dvc").exists()
 
--- a/tests/func/test_import_url.py
+++ b/tests/func/test_import_url.py
@@ -400,3 +400,18 @@ def test_import_url_empty_directory(tmp_dir, dvc, workspace):
     empty_dir = tmp_dir / "empty_dir"
     assert empty_dir.is_dir()
     assert tuple(empty_dir.iterdir()) == ()
+
+
+def test_import_url_to_remote_status(tmp_dir, dvc, local_cloud, local_remote):
+    local_cloud.gen("foo", "foo")
+
+    stage = dvc.imp_url(str(local_cloud / "foo"), to_remote=True)
+    assert stage.md5 is not None
+
+    status = dvc.status()
+    assert status["foo.dvc"] == [{"changed outs": {"foo": "not in cache"}}]
+
+    dvc.pull()
+
+    status = dvc.status()
+    assert len(status) == 0


tests/func/test_import_url.py
```diff
@@ -398,3 +398,18 @@ def test_import_url_empty_directory(tmp_dir, dvc, workspace):
     empty_dir = tmp_dir / "empty_dir"
     assert empty_dir.is_dir()
     assert tuple(empty_dir.iterdir()) == ()
+
+
+def test_import_url_to_remote_status(tmp_dir, dvc, local_cloud, local_remote):
+    local_cloud.gen("foo", "foo")
+
+    stage = dvc.imp_url(str(local_cloud / "foo"), to_remote=True)
```
karajan1001, line 406: It can be used in this way `str(Path)`? In the previous I'm always using `os.fspath(Path)` following other's work.

karajan1001 (approved):

tests/func/test_import_url.py
pmrowla: @karajan1001 either one can be used (see https://github.com/iterative/dvc/blob/master/dvc/path_info.py#L50)

dvc/repo/imp_url.py
```diff
@@ -76,6 +76,7 @@ def imp_url(
     elif to_remote:
         remote = self.cloud.get_remote(remote, "import-url")
         stage.outs[0].transfer(url, odb=remote.odb, jobs=jobs)
+        stage.md5 = stage.compute_md5()
```
skshetry, line 79: This fixes a part of the issue in #6166.  I still see the .dvc file without any `etag` or `size`.

```yaml
md5: da74a44aace86762bb08cf21d130f214
frozen: true
deps:
  - path: https://data.dvc.org/get-started/data.xml
outs:
  - md5: a304afb96060aad90176268345e10355
    nfiles: 1
    path: data.xml
```

Missing `etag` means that after:
1. `dvc import-url --to-remote`
2. `dvc pull`
3. `dvc update`,

... it will again try to download from remote. Unless I am missing something, this should set deps correctly here.

So, it should be something like following:
```suggestion
        stage.save_deps()
        stage.md5 = stage.compute_md5()
```

isidentical: > it will again try to download from remote

Isn't that already expected? The way that `import-url --to-remote` works, we disregard any remote-specific checksums (e.g etags), and compute md5 every time once for the each dependency/output pair.  So if you do `update --to-remote`, then it is going to fetch the file anyway if even it has not changed.

isidentical: Also `save_deps()` seems a very expensive operation with multiple exists calls + staging (https://github.com/iterative/dvc/blob/61835f65116c3e5808c325d6cb22a12d36f70cde/dvc/output.py#L409-L423) so I am not really sure if we have to do it every time considering that we already stage the *output* once, and doing this for the dependency might be too wasteful (unless I am missing a point in the code which prevents this).

skshetry: > Isn't that already expected? The way that import-url --to-remote works, we disregard any remote-specific checksums (e.g etags), and compute md5 every time once for the each dependency/output pair.

Yes, that's during `import-url --to-remote`. After that is complete, it should act like an `import-url` stage, I think. 
The user should be able to use `update` and/or `update --to-remote` on the stages generated by both means for different purposes. And, the behavior of `update` is that it does not re-fetch them if they don't change.

>  I am not really sure if we have to do it every time considering that we already stage the output once

You already have the access to the source file, and their info in `Output.transfer`. You can fetch the etag/checksum there already. The `size`/`nfiles` don't matter that much.

isidentical: > You already have the access to the source file, and their info in Output.transfer. You can fetch the etag/checksum there already. The size/nfiles don't matter that much.

If I am understanding what you are saying correctly, we don't collect information about per object details (e.g etags) during `Output.transfer`, since until now we only used md5s that are computed during transfer. So `save_deps()` would cost us N info() calls (where N is the number of files traversed), which in theory that we should be able to eliminate by propagating some infromation from `objects.stage` but there doesn't seem to be an easy way of doing so. 

I'm fine with changing this, but just to keep in mind that it will have downgrade the performance (unless we do some magicary stuff in `stage()` to collect etags alongside md5s).

isidentical: Also CC: @efiop @pmrowla

isidentical: After our discussion with @efiop, we decided to go with a `save_deps()` call for now and potentially optimizing this in the future (via some additional data attached to HashFiles). If the provider supports `find()`, then the current increase is about ~(n / 1000) API calls. If not, it will be a little bit higher but since we are transitioning to fsspec based systems they'll hopefully support this (and even cache it, since we are double traversing the next one should be free). 
https://github.com/iterative/dvc/blob/67c46b98dd9a0aa201d13ba1f12321a6270ac5bb/dvc/objects/stage.py#L102-L108

efiop (approved):

skshetry (approved):

skshetry:
@Mergifyio rebase

PR #6178, opened 2021-06-15
oss: migrate to ossfs

### LINKED ISSUE #6177: oss remote Migration to ossfs.
New https://github.com/iterative/ossfs released.
Need to migrate to it.

karajan1001:
Fix #6177

### DIFF
--- a/dvc/fs/oss.py
+++ b/dvc/fs/oss.py
@@ -53,15 +53,15 @@ class OSSFileSystem(ObjectFSWrapper):
     def _download(
         self, from_info, to_file, name=None, no_progress_bar=False, **pbar_args
     ):
-        total = self.fs.size(from_info.url)
+        # total = self.fs.size(from_info.url)
         with Tqdm(
             disable=no_progress_bar,
-            total=total,
+            # total=total,
             bytes=True,
             desc=name,
             **pbar_args,
         ) as pbar:
-            self.fs.download_file(
+            self.fs.get_file(
                 from_info.url,
                 to_file,
                 progress_callback=pbar.update_to,
--- a/dvc/fs/oss.py
+++ b/dvc/fs/oss.py
@@ -19,12 +19,18 @@ class OSSFileSystem(ObjectFSWrapper):
     PATH_CLS = CloudURLInfo
     REQUIRES = {"ossfs": "ossfs"}
     PARAM_CHECKSUM = "etag"
+    COPY_POLL_SECONDS = 5
+    LIST_OBJECT_PAGE_SIZE = 100
     DETAIL_FIELDS = frozenset(("etag", "size"))
 
     def _prepare_credentials(self, **config):
         login_info = {}
-        login_info["key"] = config.get("oss_key_id")
-        login_info["secret"] = config.get("oss_key_secret")
+        login_info["key"] = config.get("oss_key_id") or os.getenv(
+            "OSS_ACCESS_KEY_ID"
+        )
+        login_info["secret"] = config.get("oss_key_secret") or os.getenv(
+            "OSS_ACCESS_KEY_SECRET"
+        )
         login_info["endpoint"] = config.get("oss_endpoint")
         return login_info
 
--- a/dvc/fs/oss.py
+++ b/dvc/fs/oss.py
@@ -1,8 +1,10 @@
+import os
 import threading
 
 from funcy import cached_property, wrap_prop
 
 from dvc.path_info import CloudURLInfo
+from dvc.progress import Tqdm
 from dvc.scheme import Schemes
 
 from .fsspec_wrapper import ObjectFSWrapper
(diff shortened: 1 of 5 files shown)


dvc/objects/db/oss.py
```diff
@@ -0,0 +1,5 @@
+from .base import ObjectDB
+
+
+class OSSObjectDB(ObjectDB):
+    DEFAULT_VERIFY = True
```
isidentical, line 5: Would you mind documenting the purpose of this? From an outside look, it is not clear why we have this separate ODB for OSS, so a few inline comments might help.

efiop: @isidentical It is the same as with gdrive. This is the flag we use to tell dvc pull to not trust md5s and actually verify them by hand after downloading. A legacy thing that is a bit clumsy, it will be cleaned up during save/transfer unification, I think.

isidentical: > This is the flag we use to tell dvc pull to not trust md5s and actually verify them by hand after downloading. 

Yeah, I know what it does but what I mean is that why do we need to turn it on for OSS?

efiop: @isidentical Ah, gotcha. We've agreed with @karajan1001 to turn it on temporarily for now just as an extra assurance. We'll turn it off once we are ready. But yeah, a comment about it would be nice to avoid confusion :+1:

karajan1001: Will add them later. This PR is still in progress. Because of some of the performance optimizations. 
I will submit the result of dvc-bench for it.

karajan1001:
Have a question here its tests could not be started. It is because of the github secret?

efiop:
@mergifyio rebase

dvc/fs/oss.py
```diff
@@ -8,124 +7,65 @@
...
-    def remove(self, path_info):
-        if path_info.scheme != self.scheme:
-            raise NotImplementedError
-
-        logger.debug(f"Removing oss://{path_info}")
-        self._get_bucket(path_info.bucket).delete_object(path_info.path)
-
-    def _upload_fobj(self, fobj, to_info, **kwargs):
-        self._get_bucket(to_info.bucket).put_object(to_info.path, fobj)
+    def _upload_fobj(self, fobj, to_info, size=None):
```
karajan1001, line 35: There are some differences between the oss-emulator and the real servier. about 40% percent of the methods are unsupported in the emulator. If we do not use the emulator, these lines can be removed.

efiop: @karajan1001 Is this particular method require all of those methods to be implemented? Or are there any quick fixes for the emulator to get rid of the workaround logic?

efiop:
@karajan1001 btw, looks like it is time to create a conda package for ossfs, since it is part of our conda package, so we won't be able to release right now if we merge this. E.g. see https://github.com/conda-forge/staged-recipes/pull/15325

dvc/fs/oss.py
karajan1001: @efiop , Sorry not quite understand. Most of 40% percent of the methods which oss-emulator didn't support are not used in dvc. But this one in the base class would use them, and this would cause tests failure in `Linux`. And the `Macos` and `Windows` tests which uses real server would not be affected. If we replace the emulator with real server in `Linux` tests. We can pass it.

If you mean to repair the lacking maintenance `oss-emulator`. It requires weeks or even months of work.

efiop: @karajan1001  I thought that there is one or two broken/missing methods in emulator that we need to fix to make it work in these simple func tests, but if it will take so much time then yeah, let's just use a real thing for now. I would keep emulator fixtures too though (similar to s3 and real_s3 fixtures that we have), so that if someone will be willing to fix the emulator - he could in the future. Also, as a part of moving to plugins, we'll likely get rid of many oss tests from dvc itself in the future, so we'll later figure out what to do with those fixtures.

karajan1001:
image
Performance of these test

|tests \ version| old_version | after this PR | diff |
|:--:|:--:|:--:|:--:|
| gc | 1.74m | 2.00m| +15% |
| pull | 1.06m | 1.82m | +71% |
| push | 50.4s | 47.3s | -7% |

The increase of time cost in `pull` comes from two side 
1. Adding the progress bar
image 
2. File info get inside `ossfs.get_file` method.

The second one can be solved after https://github.com/iterative/ossfs/issues/27 been solved.

For the problem of `gc`, 
Old one
image
New one
image

The `delete_object` method from `oss2` takes a bit more time. Not look into it now.

dvc/fs/oss.py
```diff
@@ -8,124 +8,77 @@
...
-        if not kwargs.pop("prefix", False):
-            path_info = path_info / ""
-        for fname in self._list_paths(path_info):
-            if fname.endswith("/"):
-                continue
-
-            yield path_info.replace(path=fname)
+        return _OSSFileSystem(**self.fs_args)
 
     def remove(self, path_info):
```
karajan1001, line 44: Without these optimizations, the `dvc gc` would reduce about 5% of the speed.

karajan1001:
@efiop 
Default dircache of `fsspes` is not a good choice for file infos. Because the data structure are something like
```
    {"path0": [
        {"name": "path0/file0",
         "size": 123,
         "type": "file",
         ...
        },
        {"name": "path0/file1",
        },
        ...
        ],
     "path1": [...]
    }
```
which is `O(n)` in finding particular file info. Currently there are two choices:
 1. write a hash-based file info cache.
 2. follow what we did in s3 use a low-level API for `pull` 

The first one might make it difficult to merge `ossfs` into `fsspec`,  for the second choice I don't know if it is a temporary solution?

efiop:
@karajan1001 Do you mean that using fsspec's dircache doesn't speedup tests at all or that it is just not as optimal as it could be?

Also, maybe @isidentical has something to recommend here.

karajan1001:
Performance summary 

| time cost \ version| current version | last week version | reuse session version  | dircache version |
|:--:|:--:|:--:|:--:|:--:|
| gc | 100% |  +115% | 70% |  68% |
| pull | 100% | 171% |  118% | 36% |
| push | 100% | 97% | 64% |  72 %|

image
The results from the current version compare to reuse of sessions optimization.

image
The results from the current version compare to reuse of sessions optimization.

efiop:
@karajan1001 Those are amazing results! :fire:

dvc/fs/oss.py
```diff
@@ -8,124 +8,77 @@
...
-        if path_info.scheme != self.scheme:
-            raise NotImplementedError
-
         logger.debug(f"Removing oss://{path_info}")
```
efiop, line 45: This is expensive when removing lots of objects. We could make it lazy, but it is not really needed
```suggestion
```

dvc/objects/db/oss.py
```diff
@@ -0,0 +1,10 @@
+from .base import ObjectDB
+
+
+class OSSObjectDB(ObjectDB):
+    """
+    The oss remote is migrated to ossfs recently
+    add some additional verification for the data
```
efiop, line 7:
```suggestion
    Temporary extra verification
```

efiop (approved):
A few minor suggestions above. Looks great overall :fire: , please feel free to merge when you are ready.

karajan1001:
close for #6307

PR #6187, opened 2021-06-16
pass size to upload_fobj

skshetry:
This allows us to select between chunked and non-chunked uploading in Webdav, and hopefully in the HTTP itself in the future.

As an effect of this change, straight-to-remote will work even if the http/webdav server does not support chunked encoding over PUT.

This PR makes me think that if we should introduce `chunk`/`no_chunk` option, rather than always defaulting to `Content-Length` based uploads. 

See https://github.com/iterative/dvc/pull/6162#discussion_r651647741 for more information.

### DIFF
--- a/dvc/fs/base.py
+++ b/dvc/fs/base.py
@@ -235,14 +235,23 @@ class BaseFileSystem:
             no_progress_bar=no_progress_bar,
         )
 
-    def upload_fobj(self, fobj, to_info, no_progress_bar=False, **pbar_args):
+    def upload_fobj(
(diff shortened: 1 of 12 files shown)


efiop (approved):

jorgeorpinel:
Hi

> straight-to-remote will work even if the http/webdav server does not support chunked encoding over PUT

So before this `--to-remote` didn't work on HTTP or WebDAV? Still doesn't in HTTP? Thanks.

skshetry:
@jorgeorpinel, it did to some extent. This PR adds support for very old servers that don't support `Transfer-Encoding: chunked`.

PR #5833, opened 2021-04-16
azure: better error messages / allow_anonymous_login option

isidentical:
* [x] ❗ I have followed the [Contributing to DVC](https://dvc.org/doc/user-guide/contributing/core) checklist.

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here. (iterative/dvc.org/pull/2489)

Thank you for the contribution - we'll try to review it as soon as possible. 🙏

Part of #5797. Implements the `allow_anonymous_option` and support for anonymous logins.

PR #6190, opened 2021-06-17
azure: add allow_anonymous_login option to the schema

isidentical:
Follow up on https://github.com/iterative/dvc/pull/5833#discussion_r653374061

### DIFF
--- a/dvc/fs/azure.py
+++ b/dvc/fs/azure.py
@@ -147,7 +147,7 @@ class AzureFileSystem(ObjectFSWrapper):
         if (
             login_info["account_name"]
             and not any_secondary
-            and not config.get("allow_anonymous_login")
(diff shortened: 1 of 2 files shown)


dvc/fs/azure.py
```diff
@@ -147,7 +147,7 @@ def _prepare_credentials(self, **config):
         if (
             login_info["account_name"]
             and not any_secondary
-            and not config.get("allow_anonymous_login")
+            and not config.get("allow_anonymous_login", False)
```
efiop, line 150: Is this for explicitness? Or am I missing some other reason?

efiop (approved):

isidentical:
Normallay it would be None, and it would still be acceptable (it is a falsy
value), though just wanted to explicitly state that if it is not specified,
the option is going to be False.

PR #6191, opened 2021-06-17
push: prevent crash while using sqlite db on windows network share

tomrutter:
There is no strict convention for URI paths that refer to UNC shares on
windows. Although the current model is using the most common approach,
this is not the approach that sqlite expects, and a authority that is
blank or "localhost" is required. This change prevents a crash when pushing
a DVC repo that is hosted on a windows network share affected by
the change in python 3.8 that resolves network shares to their UNC
path when using os.path.abspath. Note that drive letter paths on windows
are handled identically to how they were beforehand.

### DIFF
--- a/dvc/state.py
+++ b/dvc/state.py
@@ -237,7 +237,7 @@ def _build_sqlite_uri(filename, options):
             # file://server/dir/file
             # doesn't work, so need to either use:
             # file:////server/dir/file
-            # or 
+            # or
             # file://localhost//server/dir/file
             # easiest to do the latter with current structure
             # In future versions of sqlite, this may not be necessary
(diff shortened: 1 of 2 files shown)


efiop:
Hi @tomrutter . Thanks for the PR!

Could you show the error you were getting, please? Please don't forget to add `-v` flag for verbose output.

We use raw sqlite only for remote indexes right now and plan on migrating to `diskcache`, that we already use for state. So I just wonder what error specifically you are getting right now.

efiop:
Closing as stale. For the record: we will migrate to diskcache after https://github.com/iterative/dvc/pull/6308

tomrutter:
This change is no longer relevant for the updated codebase since the recent changes to remote push/pull.

Issue #6192, opened 2021-06-17
parsing: Reraise `UnicodeDecodeError`s into something more structured


duijf:
This falls somewhere between a bug report and a feature request.

Currently, DVC assumes that all YAML/JSON/TOML files it reads are valid UTF-8:

https://github.com/iterative/dvc/blob/1d3524f14403363046048ea4aa348042fa4006f4/dvc/utils/serialize/_common.py#L48-L51

When dealing with user data in Studio, this isn't always the case though. (See https://github.com/iterative/viewer/pull/2188 and https://github.com/iterative/viewer/issues/2185).

If a file does not contain valid UTF-8, `fd.read()` raises a `UnicodeDecodeError`. 

- For interactive use on the CLI, this seems fine to me: DVC will catch the error and print an error message about the problem. Users can fix this if they whish.
- However, for Studio, we could use a more structured exception here (e.g. a subclass of `dvc.utils.serialize._common.ParseError`). We have some machinery for handling `DvcException`s and showing them to our users already. Reraising `UnicodeDecodeError`s in DVC means we do not have to special case it in Studio.

I can send a PR if you'd like. My current thinking is to add a subclass of `ParseError` for this case, but let me know if you don't

PR #6193, opened 2021-06-17
serialize: Wrap `UnicodeDecodeError`s

duijf:
Fixes https://github.com/iterative/dvc/issues/6192

This ensures that any `UnicodeDecodeError`s that can be raised by `file.read()` are raised in a more structured manner. Additionally, expose the path that the exception was raised in. This will help us handle problems with files like this in Studio.

Please let me know if you'd like to have a different exception name or there is anything else that needs to be considered here :)

(Also, apologies if I asked for the wrong people to review this. I used GitHub's reviewer suggestions)

### DIFF
--- a/dvc/utils/serialize/_common.py
+++ b/dvc/utils/serialize/_common.py
@@ -3,6 +3,7 @@ import os
 from contextlib import contextmanager
 from typing import TYPE_CHECKING, Any, Callable, ContextManager, Dict, Union
 
+from funcy import reraise
 from typing_extensions import Protocol
 
 from dvc.exceptions import DvcException
@@ -42,13 +43,24 @@ class ParseError(DvcException):
         from dvc.utils import relpath
 
         path = relpath(path)
+        self.path = path
         super().__init__(f"unable to read: '{path}', {message}")
+
+
+class EncodingError(ParseError):
+    """We could not read a file with the given encoding"""
+
+    def __init__(self, path: "AnyPath", encoding: str):
+        self.encoding = encoding
+        super().__init__(path, f"is not valid {encoding}")
 
 
 def _load_data(path: "AnyPath", parser: ParserFn, fs: "BaseFileSystem" = None):
     open_fn = fs.open if fs else open
-    with open_fn(path, encoding="utf-8") as fd:  # type: ignore
-        return parser(fd.read(), path)
+    encoding = "utf-8"
+    with open_fn(path, encoding=encoding) as fd:  # type: ignore
+        with reraise(UnicodeDecodeError, EncodingError(path, encoding)):
+            return parser(fd.read(), path)
 
 
 def _dump_data(path, data: Any, dumper: DumperFn, fs: "BaseFileSystem" = None):
--- a/tests/unit/utils/serialize/test_yaml.py
+++ b/tests/unit/utils/serialize/test_yaml.py
@@ -1,6 +1,6 @@
 import pytest
 
-from dvc.utils.serialize._yaml import YAMLError, parse_yaml
+from dvc.utils.serialize import EncodingError, YAMLError, load_yaml, parse_yaml
 
 
 def test_parse_yaml_duplicate_key_error():
@@ -12,3 +12,14 @@ def test_parse_yaml_duplicate_key_error():
     """
     with pytest.raises(YAMLError, match='found duplicate key "mykey"'):
         parse_yaml(text, "mypath")
+
+
+def test_parse_yaml_invalid_unicode(tmp_dir):
+    filename = "invalid_utf8.yaml"
+    tmp_dir.gen(filename, b"\x80some: stuff")
+
+    with pytest.raises(EncodingError) as excinfo:
+        load_yaml(tmp_dir / filename)
+
+    assert filename in excinfo.value.path
+    assert excinfo.value.encoding == "utf-8"


tests/unit/utils/serialize/test_yaml.py
```diff
@@ -12,3 +16,17 @@ def test_parse_yaml_duplicate_key_error():
     """
     with pytest.raises(YAMLError, match='found duplicate key "mykey"'):
         parse_yaml(text, "mypath")
+
+
+def test_parse_yaml_invalid_unicode():
```
skshetry, line 21: We use `tmp_dir` fixture for creating a temporary directory and generate the file using `tmp_dir.gen` here, similar to:
https://github.com/iterative/dvc/blob/d6afd26beae7c4d9b4192abfb4a37506aa2186e4/tests/func/metrics/test_diff.py#L128-L129

[tests/dir_helpers.py](https://github.com/iterative/dvc/blob/1d3524f14403363046048ea4aa348042fa4006f4/tests/dir_helpers.py) has some docstring on how to use it.

dvc/utils/serialize/_common.py
```diff
@@ -42,13 +43,23 @@ def __init__(self, path: "AnyPath", message: str):
         from dvc.utils import relpath
 
         path = relpath(path)
+        self.path = path  # For programmatic consumption, e.g. Studio.
```
skshetry, line 46: I don't think this comment is useful here.

dvc/utils/serialize/_common.py
```diff
@@ -42,13 +43,23 @@ def __init__(self, path: "AnyPath", message: str):
...
         super().__init__(f"unable to read: '{path}', {message}")
 
 
+class EncodingError(ParseError):
```
skshetry, line 50: Isn't this error a _Decoding_ error? :)

tests/unit/utils/serialize/test_yaml.py
```diff
@@ -1,6 +1,10 @@
+import os
+import tempfile
+
 import pytest
 
-from dvc.utils.serialize._yaml import YAMLError, parse_yaml
+from dvc.utils.serialize._common import EncodingError
+from dvc.utils.serialize._yaml import YAMLError, load_yaml, parse_yaml
```
skshetry, line 7: We should not be using a private API to test here. As we already doing it, I'm fine with it, but we should fix it later. 🙂

dvc/utils/serialize/_common.py
duijf: My thoughts when naming this one were "There is an error with the encoding (noun) of the file".

But as you point out, you can also read this as "An error happened while encoding the file", which isn't the case here.

I'll rename it :)

skshetry: Thinking about it again, I think it's okay. We might be able to reuse this during `dump/modify` without creating yet another `Exception`. And, `encoding` covers both `encode` and `decode` stuff anyway.

tests/unit/utils/serialize/test_yaml.py
duijf: Thanks! Wasn't aware of this convention. I've updated this test to import from `dvc.utils.serialize` instead.

dvc/utils/serialize/_common.py
duijf: Added it to explain why the field is there while it isn't used in DVC itself.

But that info will also be in the git history + we explicitly test that the field is there, so I'm also fine with removing it!

tests/unit/utils/serialize/test_yaml.py
duijf: Thanks for pointing this out! I have updated the test!

duijf:
Thanks for the review @skshetry! I have addressed your comments and pushed a new version :)

skshetry (approved):
Thank you so much @duijf for the PR and also for cleaning things up. 🙂

duijf:
> Thank you so much @duijf for the PR and also for cleaning things up.

You're welcome! I did end up renaming the exception. Would you like me to revert it to `EncodingError` so DVC can reuse it as per https://github.com/iterative/dvc/pull/6193#discussion_r654273061 ?

skshetry:
I am not sure how you handle errors on the Viewer side, but if you need to depend on `DecodingError` rather than `ParseError`,  having an `EncodingError` might be more future-proof. I am fine with it as-is on DVC's side.

Sorry for misleading you. 🙂

duijf:
> I am not sure how you handle errors on the Viewer side, but if you need to depend on DecodingError rather than ParseError, having an EncodingError might be more future-proof. I am fine with it as-is on DVC's side though.

Most of the time, we catch broad exceptions like `ParseError` and convert the exceptions into something that can be displayed to users. Depending on the use-case, we do sometimes inspect which subclass something is so we can give the user a hint on what to do next.

Considering the above, I went for some future-proofing here and changed the name back to `EncodingError`

> Sorry for misleading you

No worries! :)

I don't intend to make any further changes to this PR (unless somebody else would like to see additional changes of course). What's the merging process like? Are you OK if I merge this one or does the DVC team prefer to do this themselves?

skshetry:
> What's the merging process like? 

Our merging process is not formalized. If it's a small issue, we do merge them ourselves, but otherwise, we leave that up to @efiop to decide. 😄   I think it's better if we leave merging within the team (though I am a fan of allowing everyone to "merge when green").

PR #6195, opened 2021-06-18
objects: use separate staging ODB for staging trees

### LINKED ISSUE #5988: commit: granular commit poisons the cache

pmrowla:

Follow up to https://github.com/iterative/dvc/pull/6109.
Should fix #5988.

- `objects.stage(odb, ...)` now places staged trees in a separate ODB
    - When `odb` is local, staged objects are placed in `.dvc/tmp/staging`
    - Otherwise staged trees will be placed in a global memfs ODB
    - When using `objects.save(odb, obj, ...) with a staged tree object, it will be moved from the staging location (instead of copied) into the ODB whenever possible
- For (git) imports, staged trees are now stored in a temporary memfs based staging ODB
- For external outs that on filesystems that use non-md5 checksums, staged trees are still placed directly in the ODB since we need to get the checksum for the uploaded dir cache (and the remote fs location matters)
- Granular commit no longer poisons the local cache with filtered subdirectory `.dir` files
- Existing `tree.filter(copy=...)` logic has been replaced by separate `tree.get` and `tree.filter` methods
    - `tree.get(prefix=...)` returns a new `Tree` object that only contains the subtree (or singular HashFile) located at `prefix` (with the proper digest computed for the subtree).
    - `tree.filter(prefix=...)` returns a filtered copy of the tree object with an unmodified hash_info, but only contains trie entries inside `prefix`


dvc/repo/fetch.py
```diff
@@ -76,7 +76,7 @@ def fetch(
             )
             downloaded += d
             failed += f
-        elif isinstance(odb, GitObjectDB):
+        elif isinstance(odb.fs, MemoryFileSystem):
             # objs contains staged import objects which should be saved
             # last (after all other objects have been pulled)
             external.update(objs)
```
pmrowla, line 82: This is only temporary until fetch/save are unified

dvc/dependency/repo.py
```diff
@@ -62,16 +62,18 @@ def dumpd(self):
...
+        from dvc.fs.memory import MemoryFileSystem
         from dvc.objects import save
-        from dvc.objects.db.git import GitObjectDB
         from dvc.repo.fetch import fetch_from_odb
 
         for odb, objs in self.get_used_objs().items():
-            if not isinstance(odb, GitObjectDB):
+            if isinstance(odb.fs, MemoryFileSystem):
+                for obj in objs:
+                    save(self.repo.odb.local, obj, jobs=jobs)
```
pmrowla, line 72: This is only temporary until fetch/save are unified

dvc/objects/db/__init__.py
```diff
@@ -78,3 +79,12 @@ def __getattr__(self, name):
     def by_scheme(self):
         self._init_odb(self.CLOUD_SCHEMES)
         yield from self._odb.items()
+
+    @classmethod
+    def get_staging(cls):
```
pmrowla, line 84: This is `get_staging()` method instead of a `staging` property since we will be changing the state of `odb.read_only` depending on where we are in the staging process (we need to be able to `odb.add()` while staging an object, then we want to mark the odb instance as read_only once staging is completed, so we don't try to push those objects later.

Note that fsspec MemoryFS is global, so while individual odb instances are marked read_only for internal DVC purposes, every instance always points to a single shared/global memfs storage

dvc/fs/memory.py
```diff
@@ -1,34 +1,35 @@
...
-    def open(self, path_info, mode="r", encoding=None, **kwargs):
-        return self.fs.open(
-            path_info.fspath, mode=mode, encoding=encoding, **kwargs
-        )
+class MemoryFileSystem(FSSpecWrapper):  # pylint:disable=abstract-method
+    scheme = Schemes.MEMORY
+    PARAM_CHECKSUM = "md5"
+    PATH_CLS = CloudURLInfo
+    TRAVERSE_PREFIX_LEN = 2
+    DEFAULT_BLOCKSIZE = 4096
```
pmrowla, line 16: for whatever reason, fsspec's MemFile file-object does not define its own default blocksize, but our fsspec wrapper expects all of the cloud fobj's to define them

dvc/dependency/repo.py
```diff
@@ -112,64 +120,71 @@ def get_used_objs(
...
-            git_odb = GitObjectDB(repo.repo_fs, repo.root_dir)
-            used_objs[git_odb].add(staged_obj)
-            return used_objs
 
-    def _check_circular_import(self, odb):
+            filtered = self._filter_obj(staged_obj)
+            used_objs[staging].add(filtered)
+            return used_objs, filtered
+
+    def _check_circular_import(self, odb, objs):
```
skshetry, line 156: Is this checking for the possibility in all the chained imports?

dvc/output.py
```diff
@@ -402,7 +402,7 @@ def cache_path(self):
     def get_hash(self):
         if not self.use_cache:
             return ostage(
-                self.repo.odb.local,
+                self.repo.odb.get_staging(),
```
pmrowla, line 405: `objects.stage()` could maybe just use the memfs staging ODB by default now, and make `odb` an optional kwarg for cases where we explicitly want to stage an object in some other ODB?

efiop:
For the record: discussed that we'll likely need a local staging odb for `.dir` objects, since that's something that our internals expect when working with `state`.

dvc/dependency/repo.py
pmrowla: Yes, for chained imports the `erepo.used_objs()` call is recursive, so each "nested" erepo will also end up checking whether or not it has already been imported

tests/func/test_data_cloud.py
```diff
@@ -340,7 +340,7 @@ def test_pull_git_imports(tmp_dir, dvc, scm, erepo):
     os.makedirs(dvc.odb.local.cache_dir, exist_ok=True)
     clean_repos()
 
-    assert dvc.pull(force=True)["fetched"] == 2
+    assert dvc.pull(force=True)["fetched"] == 3
```
pmrowla, line 343: Previously, we would not count the dir cache object for git imports. After the staging changes, the git import dir cache object is odb.add'ed (from the memory ODB to the local ODB) upon fetch() and can be counted like a regular fetched dir cache object

dvc/objects/stage.py
```diff
@@ -151,19 +157,37 @@ def _get_tree_obj(path_info, fs, name, odb, state, upload, **kwargs):
...
+
     odb.add(tree.path_info, tree.fs, tree.hash_info)
-    if name != "md5":
+
+    raw = odb.get(tree.hash_info)
+    if name == "md5":
+        # cleanup the digest() generated memfs path if needed and return the
+        # staged tree object based on the ODB fs/path
+        if tree.fs.exists(tree.path_info):
+            tree.fs.remove(tree.path_info)
```
pmrowla, line 183: When staging in memory, the original `tree.path_info` will already be removed (since the `odb.add` will use `fs.move()`). When staging into local, it will be copied from memory into the local staging ODB, and we can explicitly remove the temp memfile here.

dvc/checkout.py
```diff
@@ -315,7 +316,7 @@ def checkout(
         skip = True
     else:
         try:
-            check(cache, obj)
+            check(cache, obj, skip_tree=skip_tree_check)
```
pmrowla, line 319: Depending on how we are doing checkout, `obj` may be a filtered memfs tree where we only need to checkout (link) the file entries, and don't care about the `.dir` cache being present

efiop: This does look a bit odd though, but it is clear that the reason is that our `.dir`'s are just weird like that when it comes to filtered things. If we were to switch to git-like Tree objects, this flag would not be needed, right? Just wondering if there is a more elegant way to go about this right now.

Another interesting question is if we even need this `check()` call here. Like, previously we've done it because we used `hash_info`s here and didn't know if the objects actually exist or not. And these days we handle real objects here at this point, so maybe we can just skip this unnecessary check. Maybe I'm missing something...

dvc/dependency/repo.py
```diff
@@ -112,64 +122,63 @@ def get_used_objs(
...
-                    recursive=True,
-                ).items():
-                    if odb is None:
-                        odb = repo.cloud.get_remote().odb
-                    self._check_circular_import(odb)
-                    used_objs[odb].update(objs)
-            except (NoRemoteError, NoOutputOrStageError):
-                pass
-
+            if not obj_only:
```
efiop, line 125: Could you elaborate on `obj_only`, please? Not sure I get it right away.

dvc/objects/db/base.py
```diff
@@ -30,6 +35,8 @@ def __init__(self, fs, path_info, **config):
         self.cache_type_confirmed = False
         self.slow_link_warning = config.get("slow_link_warning", True)
         self.tmp_dir = config.get("tmp_dir")
+        self.read_only = config.get("read_only", False)
+        self.staging: Optional["ObjectDB"] = config.get("staging", None)
```
efiop, line 39: Hm, could you elaborate why we need to know if it is a staging db in db itself and not just on a higher level, like `repo.odb.staging`? Is it because you could have two odbs (memfs and localfs), or maybe also some other reason?

dvc/objects/db/base.py
```diff
@@ -62,9 +75,18 @@ def get(self, hash_info):
...
+    def add(
+        self,
+        path_info: "AnyPath",
+        fs: "BaseFileSystem",
+        hash_info: "HashInfo",
+        move: bool = True,
+        **kwargs,
+    ):
+        if self.read_only:
+            raise ObjectError("Cannot write to read-only ODB")
```
efiop, line 87: Should we create a new exc class for these odb errors?

dvc/objects/db/base.py
```diff
@@ -140,9 +167,16 @@ def check(self, hash_info, check_hash=True):
         - Remove the file from cache if it doesn't match the actual hash
         """
 
+        if check_staging and self.staging is not None:
+            try:
+                self.staging.check(hash_info, check_hash=check_hash)
+                return
+            except Exception:  # pylint: disable=broad-except
```
efiop, line 174: Could you elaborate why it needs to be so broad, please?

dvc/objects/db/base.py
```diff
@@ -140,9 +167,16 @@ def check(self, hash_info, check_hash=True):
         - Remove the file from cache if it doesn't match the actual hash
         """
 
+        if check_staging and self.staging is not None:
```
efiop, line 170: Feels a bit strange that the odb itself knows that it is for staging. Feels like this needs to be checked somewhere on a higher level.

dvc/objects/db/base.py
```diff
@@ -360,6 +396,9 @@ def gc(self, used, jobs=None):
             self.fs.remove(path_info)
             removed = True
 
+        if self.staging is not None:
+            self.staging.gc(set(), jobs=jobs)
```
efiop, line 400: Again, pretty weird that odb itself knows that it is for staging and skips gc here. Am I missing something?

dvc/objects/db/base.py
efiop: Ah, so this is kinda like "my staging odb" from the odb's perspective. So this is mini-version of simple mirrors? :slightly_smiling_face: If so, it still feels like high-level things leaking into base odb logic. If we still need this to be this way, maybe it should be some kind of wrapper?

efiop (commented):
This is great :fire: Just a few questions about odb layering.

dvc/checkout.py
pmrowla: As long as we are always doing the correct `odb.save()` before `checkout` this check shouldn't be needed any more.

We could move it into `_checkout_file` and just check the individual file objects before linking if we want to be safe, but the dir-cache/tree obj check isn't needed at all.

dvc/dependency/repo.py
pmrowla: We used to have a lot of duplicated code in `get_obj` and `get_used_objs`/`get_used_cache` that did this same `_make_repo()` logic in both functions, so I unified them here.

The only difference is that for `get_used_objs` we need to do the full recursive used objects collection for imported outs, but for `get_obj` we can skip that collection (via `obj_only=True`)

dvc/objects/db/base.py
pmrowla: Yes it's more like a simple mirror from the ODB's perspective.

I did it this way because it seems like whether an tree is staged or actually saved to `repo.odb.local`, I should be able to use any ODB method other than linking with `repo.odb.local`, and the ODB should handle whether the tree object gets loaded from our "actual" cache or from the temp staging area.

dvc/objects/db/base.py
pmrowla: Going back to the mirror concept, it seemed like this still gets tied to the parent ODB to me.

dvc/objects/db/base.py
pmrowla: If we can't find the hash for any reason in the staging area, we fall back to checking the parent ODB

dvc/objects/db/base.py
pmrowla: If the parent ODB has a staging child ODB, it gc's it.

The child ODB won't know itself that it is for staging (it will have a None `staging` value)

dvc/checkout.py
efiop: > We could move it into _checkout_file and just check the individual file objects before linking if we want to be safe, but the dir-cache/tree obj check isn't needed at all.

Are per-file checks needed though? Seems like that is also pretty safe. Though our linking logic might be not raising proper exceptions when a src is missing.

pmrowla: I think we can get rid of the per-file checks then.

I'll double check what exceptions get raised when a src file is missing on link, but just to be clear, when a specific file is missing we should still checkout the rest of the file objects that do exist (and warn about the missing files), correct?

@efiop

efiop: @pmrowla Didn't mean to expand the scope here too much, we could get back to this in a followup and live with the flag  in the mean time.

> I'll double check what exceptions get raised when a src file is missing on link, but just to be clear, when a specific file is missing we should still checkout the rest of the file objects that do exist (and warn about the missing files), correct?

Even though it doesn't match current behaviour (we skip the whole dir if it is incomplete, which is very questionable), it does make sense to checkout what we can, yes.

pmrowla: Handling it in a follow-up is fine with me, and leaving this as-is with the extra flag for now will keep the current behavior since it still does the file checks before linking

dvc/objects/db/base.py
efiop: Ah, I see. It still seems weird to have one embedded into another like that. Like, there is one staging db per project (or two with mem and local?), but here we have a mechanism that assumes that any odb might have a staging odb inside. Trying to think if there is a way to put this high level logic above db class and not inside of it...

dvc/checkout.py
```diff
@@ -243,6 +251,9 @@ def _checkout_dir(
         or modified
     )
 
+    if failed:
+        raise CheckoutError(failed)
```
pmrowla, line 255: This will make us error out after checking out as much as we can (and before we save the dir entry in state).

This is a change from the current behavior, and I'm not sure how exactly we want to handle partial checkouts. (Should it just be a warning instead of an error?)

dvc/objects/db/__init__.py
```diff
@@ -77,4 +117,85 @@ def __getattr__(self, name):
...
-        yield from self._odb.items()
+        for scheme in [Schemes.LOCAL] + self.CLOUD_SCHEMES:
+            yield scheme, self._odb[scheme]
+
+    def get_staging(
+        self, scheme: Optional[str] = None
+    ) -> Optional["ObjectDB"]:
+        return self._staging.get(scheme or Schemes.MEMORY)
+
+    def load_from_state(
```
pmrowla, line 128: So one issue w/staging is that our local cache ODB and the local temp/staging ODB both share state.

In `objects.stage(odb, path_info, ...)` we used to check for a cached hash for `path_info` in `odb.state`, and then would try to load that object from `odb` (so we don't re-compute hashes for a path we've already staged).

This doesn't work when we are staging objects on one place and saving them elsewhere, we want `stage()` to load the object whether it's in the temp ODB or the main 'local cache' ODB, so staging has to be moved to a level that knows about state + all the possible ODBs where hashes from state can be stored.

It seemed to me that this fits at the ODBManager level, but I'm not sure if this should maybe go into a separate class for handling the "state + {set of ODBs that share state}" staging mechanic. (and then the repo ODBManager would also own an instance of that separate class)

pmrowla:
failing test passes for me locally, will have to investigate

dvc/checkout.py
```diff
@@ -310,21 +321,9 @@ def checkout(
         _remove(path_info, fs, cache, force=force)
         failed = path_info
 
-    elif not relink and not _changed(path_info, fs, obj, cache):
+    elif not relink and not _changed(path_info, fs, obj, cache, state=state):
         logger.trace("Data '%s' didn't change.", path_info)  # type: ignore
         skip = True
-    else:
-        try:
-            check(cache, obj)
```
pmrowla, line 318: This check no longer required, objects passed into `checkout()` should already have been checked. In the event that a file entry within a tree does not exist for some reason, we now raise the appropriate error when linking fails

dvc/dependency/repo.py
```diff
@@ -112,23 +120,25 @@ def get_used_objs(
...
+                            odb.read_only = True
+                        self._check_circular_import(objs)
+                        used_objs[odb].update(objs)
+                except (NoRemoteError, NoOutputOrStageError):
+                    pass
 
             try:
                 staged_obj = stage(
-                    local_odb,
+                    None,
```
pmrowla, line 141: This will stage imported trees in the memfs ODB. We could also just stage them in the local ODB's staging area, but persisting them doesn't save us anything at the moment.

Since we re-clone the erepo across `dvc` command runs, we won't have a state entry for the import's (clone dir) path, so we will end up having to recompute the entire dir cache each time whether or not we already have it in the local ODB (or local staging ODB).

If/when we cache the import erepo clone dir, we could benefit from staging these trees locally, since if the cached clone dir hasnt changed, we would have a valid state entry for the imported path and could actually load the existing hash from local ODB (or local ODB's staging area)

dvc/objects/__init__.py
```diff
@@ -38,18 +42,26 @@ def save(
             ):
                 future.result()
 
+            # if dir cache for this tree has already been staged, move the
+            # staged object rather than uploading/copying a new object
+            staging = get_staging(odb)
+            if staging.exists(obj.hash_info):
+                obj = staging.get(obj.hash_info)
+                kwargs["move"] = True
```
pmrowla, line 50: This could also be a flag for `save()` instead of the exists check, indicating that we already know the object has been explicitly staged (since `obj = objects.stage()` `save(odb, obj)` is a common pattern)

dvc/objects/stage.py
```diff
@@ -131,85 +145,151 @@ def _build_tree(path_info, fs, name, odb, state, upload, **kwargs):
...
+            staging.add(obj.path_info, obj.fs, obj.hash_info)
+            raw = staging.get(obj.hash_info)
+            # cleanup unneeded memfs tmpfile and return obj based on staging
+            # ODB fs/path
+            if obj.fs != staging.fs:
+                obj.fs.remove(obj.path_info)
+            obj.fs = raw.fs
+            obj.path_info = raw.path_info
+        else:
+            obj = _stage_external_tree_info(odb, obj, name)
```
pmrowla, line 288: Staged objects with non-MD5 checksums are still placed directly within the external ODB instead of inside the `/staging` path (so external cache is polluted here). But this is the only way for us to get the checksum for the resulting dir cache file, since checksums like etags are based on location and not file content

efiop (approved):

efiop:
Thanks @pmrowla ! :fire:

For the record: we've released 2.5.4 before merging, so we have a little bit of time in case we catch something here and before save/transfer unification.

PR #6214, opened 2021-06-23
Bump fsspec version

skshetry:
Testing new review assignment

### DIFF
--- a/dvc/fs/local.py
+++ b/dvc/fs/local.py
@@ -32,7 +32,9 @@ class LocalFileSystem(BaseFileSystem):
 
     def exists(self, path_info) -> bool:
         assert isinstance(path_info, str) or path_info.scheme == "local"
-        return self.fs.exists(path_info)
+        # NOTE: os.fspath could be removed after fsspec 2021.06.1
+        # https://github.com/intake/filesystem_spec/issues/666
+        return self.fs.exists(os.fspath(path_info))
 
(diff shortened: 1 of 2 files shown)


skshetry:
We need to wait for adlfs and gcsfs to make a release.

karajan1001 (approved):
Conda is still on `2021.6.0`?

efiop:
@karajan1001 Yeah, it will get updated automatically a bit later today.

efiop:
Ok, introduced a temporary hack for now.

dvc/fs/local.py
```diff
@@ -32,7 +32,9 @@ def open(path_info, mode="r", encoding=None, **kwargs):
 
     def exists(self, path_info) -> bool:
         assert isinstance(path_info, str) or path_info.scheme == "local"
-        return self.fs.exists(path_info)
+        # NOTE: os.fspath could be removed after fsspec 2021.06.1
+        # https://github.com/intake/filesystem_spec/issues/666
+        return self.fs.exists(os.fspath(path_info))
```
skshetry, line 37: Do we need to use `fsspec` here? I think this is the only place where we use it.

efiop: @skshetry The plan is to slowly start using fsspec's localfs instead of ours. So it is not necessary right now, sure.

PR #6220, opened 2021-06-24
separate validation and transformation of args from the add func

skshetry:
Just experimenting 🧪 with this, let me know what you think. 

The `validate` decorator can be used to validate and transform arguments and results from function calls (if `post=True`). Multiple validators can be provided, and the validators will get access to a dictionary that also supports dot-notation.

```python
def validator(args):
   assert args["x"] == args.x
```

Inside the validators, the arguments can also be modified/transformed, so it also acts as a converter/transformer.

Note that, there is one small difference. So, if you are calling it without default arguments, the function will receive the default arguments too.
i.e. calling function with signature `func(x, y, z=3)` as `func(x, y)` will call the function as `func(x, y, z=3)`.
It could be done, but I didn't want to complicate it.

Also note that the validator functions (or, args dotted notation specifically) are tied with the function signature. So, if we modify `dvc.add` function signature, we might need to modify the validator functions a little bit.

I could not find a good library that does this (except type-checking libraries).

Let me know what you think, or if you don't like the magic done here. ✨🪄✨  

### DIFF
--- a/dvc/repo/add.py
+++ b/dvc/repo/add.py
@@ -41,7 +41,7 @@ def check_arg_combinations(args):
 
     if to_remote or to_cache:
         message = "{option} can't be used with "
-        message += "--to-remote" if args.to_remote else "-o"
+        message += "--to-remote" if to_remote else "-o"
         if len(args.targets) != 1:
             invalid_opt = "multiple targets"
         elif args.no_commit:
--- a/dvc/repo/add.py
+++ b/dvc/repo/add.py
@@ -14,6 +14,7 @@ from ..exceptions import (
 from ..progress import Tqdm
 from ..repo.scm_context import scm_context
 from ..utils import LARGE_DIR_SIZE, glob_targets, resolve_output, resolve_paths
+from ..utils.collections import ensure_list, validate
 from . import locked
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


pared:
I love it. Validation checks are ussually just a clutter that makes debuggining harder. Extracting the logic out of their respective methods seems like a great idea!

karajan1001:
I have a question here, what is the advantage of it compare to directly packaging the arguments checking logic into some functions? This logic is still hard to read and write. And the decorator itself is not very friendly to new contributors.

karajan1001 (commented):
I have a question here, what is the advantage of it compare to directly packaging the arguments checking logic into some functions? This logic is still hard to read and write. And the decorator itself is not very friendly to new contributors.

skshetry:
>  This logic is still hard to read and write. And the decorator itself is not very friendly to new contributors.

Each layer of abstraction does make it hard for someone new. I'd say that decorators are very pythonic (just look at the no. of decorators already in the `add` function). 

I think it's a small price to pay for a nice separation between validation/transformation and actual logic. A new contributor does not need to understand all the inner workings of the decorator, a small learning curve is expected.

> I have a question here, what is the advantage of it compare to directly packaging the arguments checking logic into some functions?

To be able to keep this in the functions itself, you need to pass the args to a separate function, by converting all the received args to kwargs or duplicate the arguments anyway, which you don't need to do with this validator.

Eg: you'd very likely need to do this:
```
@locked
 @scm_context
 @validate(check_recursive_and_fname, transform_targets, check_arg_combinations)
 def add(  # noqa: C901
     repo,
     targets: "TargetType",
     recursive=False,
     no_commit=False,
     fname=None,
     to_remote=False,
     **kwargs,
 ):
    validate_args(
        repo=repo,
        targets=targets,
        recursive=recursive,
        no_commit=no_commit,
        fname=fname,
        to_remote=to_remote,
        **kwargs
    )
    kwargs = convert_kwargs(kwargs)
```

This is a lot of boilerplate to write in lots of places. And, you might need to do a lot of validations, which if you go with this route, you'll end up adding all of the validations in a single function. 

The proposal allows us to divide and conquer the validations more easily.

karajan1001:
@skshetry OK, until now I have noticed that it is on the repo level, not the API level. I used to write some checking logic on the API level. It is relatively easier. All of the arguments are in `self.args`, what you need is only.
```
self.check_args(self, ):
    self.rule_first()
    self.rule_second()
    self.rule_third()
```

skshetry:
> All of the arguments are in `self.args`, what you need is only.

yeah, this PR is partly influenced by `argparse`, so you get the `args` inside the validator function which works similar to how you described. As an example ⬇️ 

https://github.com/iterative/dvc/blob/09a014b2186b53b79cd39add9b2683dca8f3de80/dvc/repo/add.py#L27-L29

karajan1001 (approved):
Cool `_NamespacedDict` more elegant and easier to understand.

pared (approved):

PR #6221, opened 2021-06-24
[WIP] plots: html_template -> html

pared:

Adressing https://github.com/iterative/dvc/pull/5694#pullrequestreview-690401273

### DIFF
--- a/dvc/command/plots.py
+++ b/dvc/command/plots.py
@@ -45,7 +45,7 @@ class CmdPlots(CmdBase):
             rel: str = self.args.out or "plots.html"
             path = (Path.cwd() / rel).resolve()
             self.repo.plots.write_html(
-                path, plots=plots, html_template_path=self.args.html_template
+                path, plots=plots, html_template_path=self.args.html
             )
 
         except DvcException:
@@ -249,8 +249,8 @@ def _add_output_arguments(parser):
         help="Open plot file directly in the browser.",
     )
     parser.add_argument(
-        "--html-template",
+        "--html",
         default=None,
-        help="Custom HTML template for VEGA visualization.",
+        help="Custom HTML for VEGA visualization.",
         metavar="<path>",
     )
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -225,5 +225,5 @@ SCHEMA = {
         # enabled by default. It's of no use, kept for backward compatibility.
         Optional("parametrization", default=True): Bool
     },
-    "plots": {"html_template": str},
+    "plots": {"html": str},
 }
--- a/dvc/repo/plots/__init__.py
+++ b/dvc/repo/plots/__init__.py
@@ -242,7 +242,7 @@ class Plots:
     ):
         if not html_template_path:
             html_template_path = self.repo.config.get("plots", {}).get(
-                "html_template", None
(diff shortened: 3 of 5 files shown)


dvc/config_schema.py
```diff
@@ -225,5 +225,5 @@ class RelPath(str):
         # enabled by default. It's of no use, kept for backward compatibility.
         Optional("parametrization", default=True): Bool
     },
-    "plots": {"html_template": str},
+    "plots": {"html": str},
```
karajan1001, line 228: Can it compatible with the old schema? Some users might have an old version in their workspace.

pared: This is true, but the feature is relatively new and I think there is more harm in leaving `html_template` because users might associate `template` with vega templates and not understand what it actually is.

karajan1001: Can we adopt the old version, and transfer it to new ones? Or at least give some friendly messages when it happens.

dberenbaum: This feature was only introduced in the last minor version or so, and it was specifically developed for the Studio team, so I wouldn't expect many (probably any) other people are using it. Feel free to support both or whatever you think is best, but I don't think this is a breaking change worth worrying about.

pared: AFAIK @mnrozhkov is the person that needs to know about this change, also one user on discord. I will try to reach out to this user.

karajan1001 (approved):

pared:
@jorgeorpinel could you take a look if description change seems allright?

dvc/command/plots.py
```diff
@@ -249,8 +249,8 @@ def _add_output_arguments(parser):
         help="Open plot file directly in the browser.",
     )
     parser.add_argument(
-        "--html-template",
+        "--html",
```
efiop, line 252: This is backwards incompatible, right? Can we keep an alias or something?

dvc/config_schema.py
efiop: Studio guys will complain about problems with this, we should not break an already released schema. We __have__ to keep older name, but we could introduce an alias.

efiop (changes requested):
https://github.com/iterative/dvc/pull/6221/files#r663397354

dvc/config_schema.py
dberenbaum: My mistake, I didn't think it would be a big deal for studio.

jorgeorpinel: Related review, for the record: https://github.com/iterative/dvc/pull/5694#pullrequestreview-690401273.

dvc/command/plots.py
```diff
@@ -249,8 +249,8 @@ def _add_output_arguments(parser):
...
         default=None,
-        help="Custom HTML template for VEGA visualization.",
+        help="Custom HTML for VEGA visualization.",
```
jorgeorpinel, line 254:
```suggestion
        help="File with custom HTML container for DVC plots.",
```

efiop:
@pared What's the plan with this PR?

pared:
@efiop

Seems like there are not too many reports of user confusion. I think we can close for now and get back to it if we actually get any requests for it.

PR #6224, opened 2021-06-25
cleaning up add, progress bars and status

skshetry:
- Change in behaviour: Previously, `dvc add` would recursively find all files and then expand upon it with glob.
   This changes to first glob, and then recursively find all files, which matches with what the shell would do.

- Fixes progress bar for globbed targets (previously the no. for them was not accounted for).
- Reduces progress bars to 2. I have merged targets collection and stage creation into the same generator pipeline and hence replaced by a single progress bar. The other progress bar is during `Adding`.
- Also adds spinners while collecting cache and checking for graph correctness.
- Simplifies large dir recursive add warning
   i.e. removes the hint to remove the generated file. As I think if a user makes a mistake, they will cancel the operation in the middle, so no need to remove the generated file.
- Makes "recursive large directory add warning" more responsive. It'll print the message as soon as it crosses the threshold of 100, whereas before it'd have traverse through a (potentially) large directory before warning the user.
- Reorganises codebase to make it clear what `add` really does.

[asciicast](https://asciinema.org/a/YPXZhstyFDk1QdJv0tPTuZHgN)

- Note that the `tqdm` progress bars and messages might get interleaved which will be fixed in https://github.com/iterative/dvc/pull/6280.
- ~~I am not sure why tqdm progress bar becomes empty at the end and still looks like it's iterating, but it's not.~~

### DIFF
--- a/dvc/repo/add.py
+++ b/dvc/repo/add.py
@@ -94,17 +94,15 @@ def translate_graph_error(stages: Stages) -> Iterator[None]:
 def progress_iter(stages: Stages) -> Iterator["Stage"]:
     total = len(stages)
     desc = "Adding..."
-    with ui.progress(
-        stages, total=total, desc=desc, unit="file", leave=True
-    ) as pbar:
+    with ui.progress(total=total, desc=desc, unit="file", leave=True) as pbar:
         if total == 1:
             pbar.bar_format = desc
             pbar.refresh()
 
-        for item in pbar:
+        for stage in stages:
             if total > 1:
-                pbar.update_msg(f"{item.outs[0]}")
-            yield item
+                pbar.update_msg(f"{stage.outs[0]}")
+            yield stage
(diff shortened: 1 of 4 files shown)


dvc/repo/add.py
```diff
@@ -61,233 +74,190 @@ def check_arg_combinations(args):
...
-    expanded_targets = glob_targets(targets, glob=glob)
-
-    stages = []
-    for out in Tqdm(
-        expanded_targets,
-        desc="Creating DVC files",
-        disable=len(expanded_targets) < LARGE_DIR_SIZE,
-        unit="file",
-    ):
+        )
```
skshetry, line 233: These filterings are terribly slow for some reason. Drops down to `500files/sec` to about `60files/sec`.

efiop (approved):

jorgeorpinel:
Very nice!

PR #6229, opened 2021-06-25
Fix #6205: Fixed diff summary to only show states with values.

### LINKED ISSUE #6205: ui: fix `dvc diff` summary
```
files summary: 0 added, 1 deleted, 0 renamed, 0 modified, 1 not in cache
```

We should not have `0 ...` in the summary result.

hadasarik:

Fixes: https://github.com/iterative/dvc/issues/6205

### DIFF
--- a/dvc/command/diff.py
+++ b/dvc/command/diff.py
@@ -111,13 +111,8 @@ class CmdDiff(CmdBase):
         if not sum(summary.values()):
             return None
 
-        states_values = []
-        for state in states:
-            state_value = summary[state]
-            if state_value > 0:
-                states_values.append(f"{state_value} {state}")
-
-        ui.write("files summary: " + ", ".join(states_values))
+        states_summary = ", ".join(f"{summary[state]} {state}" for state in states if summary[state] > 0)
+        ui.write("files summary:", states_summary)
 
     def run(self):
         from dvc.exceptions import DvcException
--- a/dvc/command/diff.py
+++ b/dvc/command/diff.py
@@ -111,14 +111,15 @@ class CmdDiff(CmdBase):
         if not sum(summary.values()):
             return None
 
-        fmt = (
-            "files summary: {added} added, {deleted} deleted,"
-            " {renamed} renamed, {modified} modified"
-        )
-        if not hide_missing:
-            fmt += ", {not in cache} not in cache"
+        summary_message = "files summary: "
 
-        ui.write(fmt.format_map(summary))
+        for state in states:
+            state_value = summary[state]
+            if state_value > 0:
+                summary_message += f"{state_value} {state}, "
+
+        # Using [:-2] to remove last unnecessary ", " from message.
+        ui.write(summary_message[:-2])
 
     def run(self):
         from dvc.exceptions import DvcException
--- a/dvc/command/diff.py
+++ b/dvc/command/diff.py
@@ -111,15 +111,13 @@ class CmdDiff(CmdBase):
         if not sum(summary.values()):
             return None
 
-        summary_message = "files summary: "
-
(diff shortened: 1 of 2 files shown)


dvc/command/diff.py
```diff
@@ -111,14 +111,15 @@ def _show_diff(diff, hide_missing=False):
         if not sum(summary.values()):
             return None
 
-        fmt = (
-            "files summary: {added} added, {deleted} deleted,"
-            " {renamed} renamed, {modified} modified"
-        )
-        if not hide_missing:
-            fmt += ", {not in cache} not in cache"
```
pmrowla, line 119: After these changes we won't be accounting for `hide_missing`. The `not in cache` state should be hidden if the value is `0` or `hide_missing` is true

dvc/command/diff.py
```diff
@@ -111,14 +111,15 @@ def _show_diff(diff, hide_missing=False):
...
+        summary_message = "files summary: "
+
+        for state in states:
+            state_value = summary[state]
+            if state_value > 0:
+                summary_message += f"{state_value} {state}, "
 
-        ui.write(fmt.format_map(summary))
+        # Using [:-2] to remove last unnecessary ", " from message.
+        ui.write(summary_message[:-2])
```
pmrowla, line 122: It would be better to just generate the message without the extra `, ` entirely so that it doesn't need to be removed like this (maybe consider using `", ".join(...)`)

pmrowla (changes requested):
@hadasarik thanks for the PR! See my comments, there's a couple small issues

dvc/command/diff.py
hadasarik: It will, since `not in cache` is added to the list of states.

pmrowla: The issue is that `not in cache` is always in the list of states.

If `hide_missing` is True, that state should be hidden from the CLI output. But with your change, the `not in cache` state will always be appended to `states_values` (and output to the CLI) as long as the value for it is not zero.

`not in cache` state needs to be handled slightly differently than the other states - it should only be appended to `state_values` if the value is not zero AND `hide_missing` is False.

hadasarik: Look at line 83 of this file - `not in cache` is only added to the `states` list if the flag is set to `False`.

pmrowla: ah I see now, I missed that and was thinking states came directly from the diff

dvc/command/diff.py
```diff
@@ -111,14 +111,13 @@ def _show_diff(diff, hide_missing=False):
...
-        if not hide_missing:
-            fmt += ", {not in cache} not in cache"
+        states_values = []
+        for state in states:
+            state_value = summary[state]
+            if state_value > 0:
+                states_values.append(f"{state_value} {state}")
 
-        ui.write(fmt.format_map(summary))
+        ui.write("files summary: " + ", ".join(states_values))
```
skshetry, line 120: I personally find loops to be confusing for small stuff. Generators/Comprehensions might be clearer to the intent, eg:
```python
summary = ", ".join(generator)
```

Also, `ui.write` is basically a small wrapper around `print`, so you could get rid of "+" and just do:
```python
ui.write("files summary:", summary)
```

WDYT?

hadasarik: @skshetry  Fixed :)

pmrowla (approved):

jorgeorpinel:
Hi @hadasarik @skshetry @pmrowla 

This require small docs updates to https://dvc.org/doc/command-reference/diff and https://dvc.org/doc/command-reference/exp/apply please (find `files summary`).

Issue #5748, opened 2021-03-31
dvc cannot handle gitignore patterns


muaali:
**Problem Description:**

I am gitignoring most subdirectories of my `dataset` folder and trying to dvc track few files from my `dataset` subdirectories. Even after adding those files to dvc, it seems that dvc cannot track them. Local cache is built and they are hashed, but not tracked by dvc apparently.

**Error output:**
After adding `.dvc` files with `dvc add`, and running `dvc push` gives the following output:
```
There are no data or pipelines tracked in this project yet.           
See <https://dvc.org/doc/start> to get started!
```

**Debugging:**

```
dvc status
dvc status path/to/file.dvc
dvc status -R .
```
give output 
```
There are no data or pipelines tracked in this project yet.           
See <https://dvc.org/doc/start> to get started!
```

**Reproduce**

Create a folder structure as shown in the gitignore file below. Add `gitignore` to the main directory with the following content :
```
!/package_root/dataset/
/package_root/dataset/*
!/package_root/dataset/train_dataset/
/package_root/dataset/train_dataset/*
!/package_root/dataset/train_dataset/labels/
/package_root/dataset/train_dataset/labels/*
!/package_root/dataset/train_dataset/labels/*.dvc
!/package_root/dataset/train_dataset/Images/
/package_root/dataset/train_dataset/Images/*
!/package_root/dataset/train_dataset/Images/*.dvc

```
Run following commands: 

```
dvc add package_root/dataset/train_dataset/labels/labels.csv #labels.csv is a dummy csv file.
git add package_root/dataset/train_dataset/labels/labels.csv.dvc
dvc add package_root/dataset/train_dataset/Images/* #contain images, can be any image.
git add package_root/dataset/train_dataset/Images/*.dvc
dvc push
```
gives the output : 
```
There are no data or pipelines tracked in this project yet.           
See <https://dvc.org/doc/start> to get started!
```
And with more verbose output : 
`dvc push -v`
gives output
```
2021-03-31 16:18:23,019 DEBUG: Check for update is enabled.
2021-03-31 16:18:23,134 DEBUG: Preparing to upload data to 's3://<bucket_name>/<directory>'
2021-03-31 16:18:23,134 DEBUG: Preparing to collect status from s3://<bucket_name>/<directory>
2021-03-31 16:18:23,134 DEBUG: Collecting information from local cache...
2021-03-31 16:18:23,134 DEBUG: Collecting information from remote cache...                                                                                 
[...]

PR #6232, opened 2021-06-27
git: Keep trailing slash when checking if directory is gitignored

tpietruszka:
Closes https://github.com/iterative/dvc/issues/5748

When querying `dulwich` to check if a directory is gitignored, the path we query should include the trailing slash:
https://github.com/dulwich/dulwich/blob/76dd8356f9d8870f44226e4232c24fbc1e573e93/dulwich/ignore.py#L228-L231

Without it, we get "false positives" and effectively ignore some `.dvc` files that should not be ignored (see issue for details)

Remaining issues:
1. Is it reasonable to write unit tests for this? Can existing machinery for testing handle "un-ignores" (corresponding to gitignore lines prefixed with `!`)?
2. `is_ignored` [was type-hinted as expecting string-typed paths](https://github.com/tpietruszka/dvc/blob/master/dvc/scm/git/backend/dulwich.py#L281); I assumed that type, but 2 unit tests failed, as they were causing `PosixPath`-like types (temporary ones?) to be passed to this function; I changed everything to be passed as string - but another solution would be to  handle Path-likes. Not sure what's the general strategy here.

### DIFF
--- a/dvc/repo/stage.py
+++ b/dvc/repo/stage.py
@@ -462,7 +462,8 @@ class StageLoad:
 
         def is_out_or_ignored(root, directory):
             dir_path = f"{root}{sep}{directory}"
-            return dir_path in outs or is_ignored(dir_path)
+            # trailing slash needed to check if a directory is gitignored
+            return dir_path in outs or is_ignored(f"{dir_path}/")
 
         stages = []
         for root, dirs, files in self.repo.dvcignore.walk(
--- a/dvc/repo/stage.py
+++ b/dvc/repo/stage.py
@@ -463,7 +463,7 @@ class StageLoad:
         def is_out_or_ignored(root, directory):
             dir_path = f"{root}{sep}{directory}"
             # trailing slash needed to check if a directory is gitignored
-            return dir_path in outs or is_ignored(f"{dir_path}/")
+            return dir_path in outs or is_ignored(f"{dir_path}{sep}")
 
         stages = []
         for root, dirs, files in self.repo.dvcignore.walk(
--- a/dvc/scm/git/__init__.py
+++ b/dvc/scm/git/__init__.py
@@ -160,7 +160,7 @@ class Git(Base):
     def ignore(self, path):
         entry, gitignore = self._get_gitignore(path)
 
-        if self.is_ignored(path):
+        if self.is_ignored(str(path)):
             return
 
         msg = "Adding '{}' to '{}'.".format(relpath(path), relpath(gitignore))
--- a/dvc/scm/git/__init__.py
+++ b/dvc/scm/git/__init__.py
@@ -160,7 +160,7 @@ class Git(Base):
     def ignore(self, path):
         entry, gitignore = self._get_gitignore(path)
 
-        if self.is_ignored(str(path)):
+        if self.is_ignored(path):
             return
 
         msg = "Adding '{}' to '{}'.".format(relpath(path), relpath(gitignore))
--- a/dvc/scm/git/backend/dulwich.py
+++ b/dvc/scm/git/backend/dulwich.py
@@ -283,8 +283,8 @@ class DulwichBackend(BaseGitBackend):  # pylint:disable=abstract-method
         # `None` if it's not mentioned at all. `True` if it is ignored.
         relative_path = relpath(path, self.root_dir)
         # if checking a directory, a trailing slash must be included
-        if path[-1] == "/":
-            relative_path += "/"
+        if path[-1] == os.sep:
+            relative_path += os.sep
         return bool(self.ignore_manager.is_ignored(relative_path))
 
     def set_ref(
--- a/dvc/scm/git/backend/dulwich.py
+++ b/dvc/scm/git/backend/dulwich.py
@@ -281,9 +281,11 @@ class DulwichBackend(BaseGitBackend):  # pylint:disable=abstract-method
     def is_ignored(self, path: str) -> bool:
         # `is_ignored` returns `false` if excluded in `.gitignore` and
         # `None` if it's not mentioned at all. `True` if it is ignored.
-        return bool(
-            self.ignore_manager.is_ignored(relpath(path, self.root_dir))
-        )
+        relative_path = relpath(path, self.root_dir)
+        # if checking a directory, a trailing slash must be included
+        if path[-1] == "/":
+            relative_path += "/"
+        return bool(self.ignore_manager.is_ignored(relative_path))
 
     def set_ref(
         self,
--- a/dvc/scm/git/backend/dulwich.py
+++ b/dvc/scm/git/backend/dulwich.py
@@ -27,6 +27,8 @@ from ..objects import GitObject
 from .base import BaseGitBackend
 
 if TYPE_CHECKING:
+    from dvc.types import StrPath
+
     from ..objects import GitCommit
 
 logger = logging.getLogger(__name__)
@@ -278,12 +280,12 @@ class DulwichBackend(BaseGitBackend):  # pylint:disable=abstract-method
 
         return IgnoreFilterManager.from_repo(self.repo)
 
-    def is_ignored(self, path: str) -> bool:
+    def is_ignored(self, path: "StrPath") -> bool:
         # `is_ignored` returns `false` if excluded in `.gitignore` and
         # `None` if it's not mentioned at all. `True` if it is ignored.
         relative_path = relpath(path, self.root_dir)
         # if checking a directory, a trailing slash must be included
-        if path[-1] == os.sep:
+        if str(path)[-1] == os.sep:
             relative_path += os.sep
         return bool(self.ignore_manager.is_ignored(relative_path))
 
--- a/tests/func/test_scm.py
+++ b/tests/func/test_scm.py
@@ -99,6 +99,21 @@ def test_ignored(tmp_dir, scm):
     assert not scm.is_ignored(tmp_dir / "dir1" / "file2.txt")
 
 
+def test_ignored_dir_unignored_subdirs(tmp_dir, scm):
+    tmp_dir.gen({".gitignore": "data/**\n!data/**/\n!data/**/*.csv"})
+    scm.add([".gitignore"])
+    tmp_dir.gen(
+        {"data/raw/tracked.csv": "cont", "data/raw/not_tracked.json": "cont"}
+    )
+
+    assert not scm.is_ignored("data/raw/tracked.csv")
+    assert scm.is_ignored("data/raw/not_tracked.json")
+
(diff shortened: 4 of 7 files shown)


dvc/repo/stage.py
```diff
@@ -462,7 +462,8 @@ def is_dvcfile_and_not_ignored(root, file):
 
         def is_out_or_ignored(root, directory):
             dir_path = f"{root}{sep}{directory}"
-            return dir_path in outs or is_ignored(dir_path)
+            # trailing slash needed to check if a directory is gitignored
+            return dir_path in outs or is_ignored(f"{dir_path}/")
```
pmrowla, line 466: Path sep won't be `/` on windows

```suggestion
            return dir_path in outs or is_ignored(f"{dir_path}{sep}")
```

dvc/scm/git/backend/dulwich.py
```diff
@@ -281,9 +281,11 @@ def ignore_manager(self):
     def is_ignored(self, path: str) -> bool:
```
pmrowla, line 281: We can loosen the type-hinting here to allow path-like objects. If you update the imports at the top of this file to include
```
if TYPE_CHECKING:
    from dvc.types import StrPath
```

You can update this function declaration to:
```
def is_ignored(self, path: StrPath) -> bool
```
to allow the use of path-like objects. (`relative_path` will still always end up being a `str` after the `relpath()` call)

You will also have to update the function/typing declaration for `is_ignore` in the other `dvc/scm/git/backend/*.py` files

dvc/scm/git/backend/dulwich.py
```diff
@@ -281,9 +281,11 @@ def ignore_manager(self):
...
         # `is_ignored` returns `false` if excluded in `.gitignore` and
         # `None` if it's not mentioned at all. `True` if it is ignored.
-        return bool(
-            self.ignore_manager.is_ignored(relpath(path, self.root_dir))
-        )
+        relative_path = relpath(path, self.root_dir)
+        # if checking a directory, a trailing slash must be included
+        if path[-1] == "/":
+            relative_path += "/"
```
pmrowla, line 287: Use `os.sep` here

tests/func/test_scm.py
```diff
@@ -95,8 +95,8 @@ def test_ignored(tmp_dir, scm):
     tmp_dir.gen({"dir1": {"file1.jpg": "cont", "file2.txt": "cont"}})
     tmp_dir.gen({".gitignore": "dir1/*.jpg"})
 
-    assert scm.is_ignored(tmp_dir / "dir1" / "file1.jpg")
-    assert not scm.is_ignored(tmp_dir / "dir1" / "file2.txt")
+    assert scm.is_ignored(str(tmp_dir / "dir1" / "file1.jpg"))
+    assert not scm.is_ignored(str(tmp_dir / "dir1" / "file2.txt"))
```
pmrowla, line 99: After the typing update these `str()` calls should no longer be needed

dvc/scm/git/__init__.py
```diff
@@ -160,7 +160,7 @@ def _get_gitignore(self, path):
     def ignore(self, path):
         entry, gitignore = self._get_gitignore(path)
 
-        if self.is_ignored(path):
+        if self.is_ignored(str(path)):
```
pmrowla, line 163: After typing updates this `str()` should no longer be needed

tests/func/test_scm.py
pmrowla: Also, it would be great if you could add a test case for the gitignore patterns that we were not handling correctly to this file

pmrowla (changes requested):
Thanks for doing the research for this issue! Just a few small comments on the PR

tests/func/test_scm.py
pmrowla: You can generate an appropriate .gitignore the same way it's done in line 96

To get a .gitignore like:
```
dir/*
!dir/foo
```
you would generate it like:
```
tmp_dir.gen({".gitignore": "dir/*\n!dir/foo"})
```

dvc/repo/stage.py
tpietruszka: Indeed, I don't know how I forgot about this 

Sorry about the mess

tests/func/test_scm.py
tpietruszka: Done
Please LMK if sufficient

Thank you for the hint!

tests/func/test_stage.py
```diff
@@ -227,6 +227,13 @@ def test_parent_repo_collect_stages(tmp_dir, scm, dvc):
     assert deep_subrepo_stages != []
 
 
+def test_collect_repo_ignored_dir_unignored_pattern(tmp_dir, dvc, scm):
+    tmp_dir.gen({".gitignore": "data/**\n!data/**/\n!data/**/*.dvc"})
+    scm.add([".gitignore"])
+    (stage,) = tmp_dir.dvc_gen({"data/raw/tracked.csv": "5,6,7,8"})
+    assert dvc.stage.collect_repo() == [stage]
```
pmrowla, line 234: This is fine for ensuring that `collect_repo()` works properly, but can you add a similar test to `tests/func/test_scm.py` that ensures the `!` exclusions work properly?

using this example the checks would be something like:
```
assert scm.is_ignored("data/")
assert scm.is_ignored("data/raw/")
assert not scm.is_ignored("data/raw/tracked.csv.dvc")
```

tpietruszka: Right, makes sense. Added one

tests/func/test_scm.py
```diff
@@ -99,6 +99,21 @@ def test_ignored(tmp_dir, scm):
...
+    tmp_dir.gen({".gitignore": "data/**\n!data/**/\n!data/**/*.csv"})
+    scm.add([".gitignore"])
+    tmp_dir.gen(
+        {"data/raw/tracked.csv": "cont", "data/raw/not_tracked.json": "cont"}
+    )
+
+    assert not scm.is_ignored("data/raw/tracked.csv")
+    assert scm.is_ignored("data/raw/not_tracked.json")
+
+    assert not scm.is_ignored("data/")
```
tpietruszka, line 112: I was considering adding also `assert scm.is_ignored("data")`.
It passes, but it's kind of unclear to me what the outcome should be there - is it an universally correct behavior or just  a `dulwich` interpretation. 
I'd prefer not to add it after all

pmrowla: Testing it with CLI `git check-ignore` looks like it is supposed to pass, so it looks like dulwich's behavior is correct

As far as I can tell CLI git only ignores it if you explicitly include the trailing `/`

pmrowla: So in this case, CLI check-ignore is doing the `isdir` check on the supplied local filesystem path, and then compares it to the pattern with trailing slash if it is a directory.

But in our case, in `scm.is_ignored()` we are already assuming that the supplied path has had the trailing slash explicitly added if `path` is supposed to be a directory.

So in this check:
```
assert scm.is_ignored("data/raw")
```
we are explicitly checking the file path `data/raw` and not the directory path `data/raw/`. Meaning the assertion should pass.

It would be good to add a comment about this behavior in the test regarding why it will differ from CLI `git check-ignore` though.

---

And actually, for this test you shouldn't need to generate the file content at all (lines 105-107), as `scm.is_ignored` only deals with checking raw paths against the ignore patterns, and does not care whether or not the paths exist (or are files/dirs).

pmrowla: also, note that you can get the same behavior with CLI `git check-ignore` if you are comparing raw paths that don't exist in your local working directory

```
ignore-test git:master ✩   ❯ cat .gitignore
dir/**
!dir/**/
!dir/**/*.dvc

ignore-test git:master ✩   ❯ ls

ignore-test git:master ✩   ❯ git check-ignore dir

ignore-test git:master ✩   ❯ git check-ignore dir/                                                                                    ⏎
dir/

ignore-test git:master ✩   ❯ git check-ignore dir/subdir
dir/subdir

ignore-test git:master ✩   ❯ git check-ignore dir/subdir/
dir/subdir/

ignore-test git:master ✩   ❯ git check-ignore dir/subdir/foo.dvc

ignore-test git:master ✩   ❯ git check-ignore dir/subdir/foo                                                                          ⏎
dir/subdir/foo
```

although I'm not entirely sure why `dir/subdir/` doesn't pass here, as my understanding was that it should be excluded by `!dir/**/`

tpietruszka: Perhaps this is because of the `is_dir` check?  (since `dir/subdir/` is not an existing directory, perhaps the trailing slash gets stripped internally?)

In any case, do you think it would be okay to merge this as is? My reasoning is: let's test the behavior we actually need (focusing on existing files and dirs) and not all of the `dulwich`/`git check-ignore` contract, as that might be a huge task

pmrowla: The issue is that checking ignores against raw paths which don't exist locally is a needed use-case in DVC, as when we do operations with a specific git-revision (like `dvc push --rev=...`), we will be checking raw paths (that might not exist at all locally) directly against git trees.

So handling git-ignored/excluded `.dvc` files consistently is a requirement, whether we are pushing what's in the user's local workspace, or what is in some other git revision/branch (that is not checked-out in the git working tree)

pmrowla: Looking at the CGit source, they do only treat paths passed into `check-ignore` as directories if the path actually exists, and is a directory, and will end up removing the trailing slash from the path if the is_directory check fails before checking against the gitignore patterns

the paths are checked using this `is_directory()` function:
https://github.com/git/git/blob/142430338477d9d1bb25be66267225fb58498d92/abspath.c#L8

which confirms what you were thinking about the `git check-ignore` behavior

---

So at a minimum, given the gitignore patterns

```
dir/**
!dir/**/
!dir/**/*.dvc
```

I think we should be testing with the following raw paths:

```
assert not scm.is_ignored("dir")
assert scm.is_ignored("dir/file")
assert not scm.is_ignored("dir/subdir/")
assert scm.is_ignored("dir/subdir/file")
assert not scm.is_ignored("dir/subdir/file.dvc")
---

tpietruszka: I see. Didn't think about that. Added a comment as I understand it - and assertions for both existing and made-up paths

tpietruszka: Only saw your second comment after submitting.  Sorry about the commit spam :wink: 

Adding a few more assertions

pmrowla: Thanks! It looks like this PR should be good now as long as CI passes

pmrowla (approved):

pmrowla:
Thanks again for doing the legwork on this!

Issue #6227, opened 2021-06-25
add: performance and reliability issues


skshetry:
- [x] Repeated `dvc add` is not skipped.
  ```console
  $ dvc add data
  $ dvc add data
  ```
   In 1.X, it'd have been skipped. And, dvc still deletes the file and tries to restore it from the cache making it slower.

- [x] DVC uses move-then-checkout logic. It moves the file from the workspace to the cache and then checks it out again, rather than just using copy. 

    This is slow and might result in data loss if it happens to fail in between the operations.

- [x] DVC deletes the stage file, before even adding those files. This means that if the `dvc add` operation fails, the existing pointer file is lost, which is the only way to get access to the data.

- [x] DVC resets the stages multiple times (only if multiple targets are provided) and forces the stage recollection which is slow.

- [x] To the same effect, it resets the internal state of the repo after creating each stage, which also happens to reset dulwich's ignore manager, making it horribly slow if using too many targets (or, `-R`).

https://github.com/iterative/dvc/blob/4e792ae61c5927ab2e5f6a6914d985d43aa705b4/dvc/repo/add.py#L266

pared:
> DVC uses move-then-checkout logic. It moves the file from the workspace to the cache and then checks it out again, rather than just using copy.

Wasn't this intended to enforce cache link type? I guess in case of `copy` it would make sense but what about others?

skshetry:
For other links, the one I suggested was to change copy behaviour to be `move + link` that works *atomically*. 
@efiop also suggested using hardlinks instead.

skshetry:
Closed by

- https://github.com/iterative/dvc-data/pull/522
- https://github.com/iterative/dvc-data/pull/543
- https://github.com/iterative/dvc-data/pull/544
- https://github.com/iterative/dvc-data/pull/545
- https://github.com/iterative/dvc-data/pull/548
- https://github.com/iterative/dvc-data/pull/552
- https://github.com/iterative/dvc-data/pull/554
- https://github.com/iterative/dvc-data/pull/562
- https://github.com/iterative/dvc-data/pull/561
- https://github.com/iterative/dvc/pull/10513
- https://github.com/iterative/dvc/pull/10519

and, released in https://github.com/iterative/dvc/releases/tag/3.54.0.

PR #6239, opened 2021-06-28
add: do not delete stage files before add

skshetry:
Because of the way we collect stages and cache them, we were not able to
collect them for the `add` without removing them from the workspace.
As doing so, we'd have two same/similar stages - one collected from the
workspace and the other just created from the `dvc add` in-memory.

This would raise errors during graph checks, so we started to delete them
and reset them (which is very recently, see #2886 and #3349).

By deleting the file before we even do any checks, we are making DVC
fragile, and results in data loss for the users with even simple mistakes.
This should make it more reliable and robust.

And, recently, we have started to keep states of a lot of things, that by
resetting them on each stage, we waste a lot of performance, especially
on gitignores. We cache the Dulwich's IgnoreManager, which when reset
too many times will waste a lot of our time just collecting them again
next time (see #6227).

It's hard to say how much this improves, as this very much depends on
no. of gitignores in the repo (which can be assumed to be quite a number
for a DVC repo) and the number of files that we are adding
(eg: `-R` adding a large directory).

On a directory with 10,000 files (in a dataset-registry repo),
creating stages on `dvc add -R` went from 64 files/sec to 1.1k files/sec.

Also relevant discussion: https://github.com/iterative/dvc/issues/3362

### DIFF
--- a/dvc/repo/add.py
+++ b/dvc/repo/add.py
@@ -98,8 +98,11 @@ def add(  # noqa: C901
                 **kwargs,
             )
 
+            # remove existing stages that are to-be replaced with these
+            # new stages for the graph checks.
+            old_stages = set(repo.stages) - set(stages)
             try:
-                repo.check_modified_graph(stages)
+                repo.check_modified_graph(stages, list(old_stages))
             except OverlappingOutputPathsError as exc:
                 msg = (
                     "Cannot add '{out}', because it is overlapping with other "
@@ -233,7 +236,6 @@ def _create_stages(
     transfer=False,
     **kwargs,
 ):
-    from dvc.dvcfile import Dvcfile
     from dvc.stage import Stage, create_stage, restore_meta
 
     expanded_targets = glob_targets(targets, glob=glob)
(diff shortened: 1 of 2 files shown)


shcheklein:
Can we add some tests for this (failed op doesn't remove file for example?). Or prevent regression from happening in some other way in the future.

dvc/repo/add.py
```diff
@@ -98,8 +98,11 @@ def add(  # noqa: C901
                 **kwargs,
             )
 
+            # remove existing stages that are to-be replaced with these
+            # new stages for the graph checks.
+            old_stages = set(repo.stages) - set(stages)
             try:
-                repo.check_modified_graph(stages)
+                repo.check_modified_graph(stages, list(old_stages))
```
skshetry, line 105: This would be better reflected by `Index` here, as we are creating a new index. Index is a collection of stages, so if some stages underneath are changed, is it the same index? So, immutability is at the heart of the index and would have been better reflected here.
```suggestion
            new_index = repo.index.update(stages)
            try:
                new_index.check_graph()
```

skshetry:
> Can we add some tests for this (failed op doesn't remove file for example?). Or prevent regression from happening in some other way in the future.

Thanks, @ivan, I have added some tests. 🙂

efiop (approved):
We are doing stuff like this because we wanted to not leave a dvc-file in case of failure to not lead the users to believe that the command was actually successful, but that was a wrong approach. What we should've really done is worked on better indication of success/failure and avoid losing the information.

PR #6247, opened 2021-06-29
updater: only notify in dvc doctor

### LINKED ISSUE #5912: updater: don't remind about updates on each command run
We're getting reports that updater is too annoying (it prints the notification on every command that you run) and people just turn it off for good with a config option.

We could consider just warning once in, say, a week, but maybe we should just get rid of it as some people were suggesting (or create a separate `dvc check-updates` command). 

CC @dberenbaum @dmpetrov

efiop:
We've used updater for years now to notify our users ASAP about new
releases. These days dvc is more stable and we've actually been
receiving numerous complains about updater being too annoying and
even introduced a config flag to turn it off, which a lot of people do
and which defeats its purpose. Instead of introducing some complex logic
that would notify more rarely and once for the user instead of once for
every dvc project they have - this PR just turns it off for everything,
except `dvc doctor/version`, which is the command that people use when
they are running into issues and that will give them a hint to just try
out the newest version first, as it might be already fixed.

This PR is a more of a compromise, to get the most out of updater while
removing a pain point of it being too annoying, without investing much
time into it and not having to maintain complex unnecessary logic. We
might get back to introducing something more intricate in the future,
if the need arises (hence not removing the config option for now).

Fixes #5912

### DIFF
--- a/dvc/command/base.py
+++ b/dvc/command/base.py
@@ -33,17 +33,12 @@ class CmdBase(ABC):
 
     def __init__(self, args):
         from dvc.repo import Repo
-        from dvc.updater import Updater
 
         os.chdir(args.cd)
 
         self.repo = Repo(uninitialized=self.UNINITIALIZED)
         self.config = self.repo.config
         self.args = args
-        if self.repo.tmp_dir:
-            hardlink_lock = self.config["core"].get("hardlink_lock", False)
-            updater = Updater(self.repo.tmp_dir, hardlink_lock=hardlink_lock)
-            updater.check()
 
     def do_run(self):
         with self.repo:
--- a/dvc/command/version.py
+++ b/dvc/command/version.py
@@ -10,9 +10,21 @@ logger = logging.getLogger(__name__)
 class CmdVersion(CmdBaseNoRepo):
     def run(self):
         from dvc.info import get_dvc_info
+        from dvc.repo import NotDvcRepoError, Repo
 
         dvc_info = get_dvc_info()
         ui.write(dvc_info, force=True)
+
+        try:
+            with Repo() as repo:
+                from dvc.updater import Updater
+
+                hardlink_lock = repo.config["core"].get("hardlink_lock", False)
+                updater = Updater(repo.tmp_dir, hardlink_lock=hardlink_lock)
+                updater.check()
+        except NotDvcRepoError:
+            pass
+
         return 0
 
 


dvc/info.py
```diff
@@ -32,6 +33,10 @@ def get_dvc_info():
 
     try:
         with Repo() as repo:
+            hardlink_lock = repo.config["core"].get("hardlink_lock", False)
+            updater = Updater(repo.tmp_dir, hardlink_lock=hardlink_lock)
+            updater.check()
+
```
skshetry, line 39: The version info on the bottom makes more sense in the `version` output, like pip.
```console
...
You are using pip version 21.1.2; however, version 21.1.3 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

dvc/info.py
```diff
@@ -32,6 +33,10 @@ def get_dvc_info():
 
     try:
         with Repo() as repo:
+            hardlink_lock = repo.config["core"].get("hardlink_lock", False)
+            updater = Updater(repo.tmp_dir, hardlink_lock=hardlink_lock)
+            updater.check()
```
skshetry, line 38: This will run when we raise exceptions as well. It should be part of the `CmdVersion`, I think.

dberenbaum (approved):

skshetry:
@efiop, let me know if you'd like me to help with this. I'd like to prioritize this for the UI stuff. 

Regarding the exception-related logic, as we discussed making it dynamic, for now, we could just hack it with `updater=True|False`.

efiop:
@skshetry I can implement the proposed changes, no problem. But if you want to change more around there, and this PR would get in the way - we could close it.

dvc/info.py
efiop: On a second thought, both box and free-floating text (btw, red is a  color of errors in my mind, probably not the best for upgrade notes) look out of place in `dvc version`'s current output. Let's just keep the old box for now, since it looks like `dvc version` output needs to be reconsidered.

skshetry: > btw, red is a color of errors in my mind

red is just a suggestion, I don't see why it cannot be yellow-ish.

> look out of place in dvc version's current output

This is what this PR is about right? free-floating text is much simpler than the box and fits with the version output more than the box. 

> Plus upgrading instructions (esp multiline ones) don't quite fit in pip-like message.

Multiline is only a problem for the binary package, and even then, it could be much simpler:
```\n
To upgrade, uninstall `dvc` and reinstall from https://dvc.org.
```

> since it looks like dvc version output needs to be reconsidered.

I don't see how this is related. `updater` message was already there if it fits or not. I don't see how version output needs to be considered to accommodate this, I consider them to be separate things.

skshetry (approved):

pmrowla (approved):

skshetry:
The UI changes will come in the next PR.

PR #6254, opened 2021-06-30
push params in the experiments table to the right

skshetry:
Made a mistake during refactoring on styling before. Noticed while reviewing #6174.

#### Before
Screen Shot 2021-06-30 at 14 31 18

#### After

Screen Shot 2021-06-30 at 15 31 40

- metrics were left-aligned, whereas now it's right-aligned
- also fixed the mistake `params` as justify.

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -397,7 +397,7 @@ def show_experiments(
     styles.update(
         {
             header: {
-                "justify": "left" if typ == "metrics" else "params",
+                "justify": "right" if typ == "metrics" else "left",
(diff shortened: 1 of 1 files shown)


pmrowla:
I think your original code is right. Metrics are right-aligned and params are left-aligned to match the viewer behavior. For the columns where metrics & params meet, the right-most metric and left-most param values are supposed to appear directly adjacent to one another.

skshetry:
> I think your original code is right. 

Metric columns are still broken then, as they are left aligned.

pmrowla (approved):

PR #6273, opened 2021-07-02
doctor: show versions for fs requirements

### LINKED ISSUE #6248: doctor: should dvc doctor show remote-required dependency versions
Something that we've been experiencing lately is asking for an extra question regarding the version of their used remote dependency. We could try to annotate the dvc doctor's output with the stuff that we can get (optional). Something like this
```
 $ dvc doctor
DVC version: 2.4.2+f1bf10.mod 
Platform: Python 3.8.5+ on Linux-5.4.0-77-generic-x86_64-with-glibc2.29
...
Remotes: gs (gcsfs == 2021.06.01, ... == 1.0.2)
...
```

isidentical:
Resolves #6248
Previous output:
```
DVC version: 2.4.3+ce45ec 
Platform: Python 3.8.5+ on Linux-5.4.0-77-generic-x86_64-with-glibc2.29
Supports: s3, gs, http
Caches: local
Remotes: https
Workspace directory: ext4 on /dev/sda7
Repo: dvc, git
```

Current output:
```
DVC version: 2.4.3+fc7409 
Platform: Python 3.8.5+ on Linux-5.4.0-77-generic-x86_64-with-glibc2.29
Plugins:
        gs (gcsfs = 2021.6.1),
        http (requests = 2.25.1),
        s3 (s3fs = 2021.6.1, boto3 = 1.17.49),
Caches: local
Remotes: https
Workspace directory: ext4 on /dev/sda7
Repo: dvc, git
```

Current output of all remotes:
```
DVC version: 2.4.3+fc7409 
Platform: Python 3.8.5+ on Linux-5.4.0-77-generic-x86_64-with-glibc2.29
Plugins:
        azure (adlfs = 0.7.7, knack = 0.7.2, azure-identity = 1.5.0),
        gdrive (pydrive2 = 1.8.1),
        gs (gcsfs = 2021.6.1),
        hdfs (pyarrow = 2.0.0),
        webhdfs (hdfs = 2.5.8),
        http (requests = 2.25.1),
        https (requests = 2.25.1),
        s3 (s3fs = 2021.6.1, boto3 = 1.17.49),
        ssh (paramiko = 2.7.2),
        oss (oss2 = 2.6.1),
        webdav (webdav4 = 0.8.1),
        webdavs (webdav4 = 0.8.1)
Caches: local
Remotes: https
Workspace directory: ext4 on /dev/sda7
Repo: dvc, git
```

### DIFF
--- a/dvc/info.py
+++ b/dvc/info.py
@@ -2,7 +2,6 @@ import itertools
 import os
 import pathlib
 import platform
-import sys
 import uuid
 
 import psutil
@@ -21,10 +20,10 @@ if PKG is None:
 else:
     package = f"({PKG})"
 
-if sys.version_info < (3, 8):
-    import importlib_metadata
-else:
+try:
     import importlib.metadata as importlib_metadata
+except ImportError:  # < 3.8
+    import importlib_metadata  # type: ignore[no-redef]
 
 
 def get_dvc_info():
@@ -33,7 +32,7 @@ def get_dvc_info():
         "---------------------------------",
         f"Platform: Python {platform.python_version()} on "
         f"{platform.platform()}",
-        f"Plugins:{_get_supported_remotes()}",
+        f"Supports:{_get_supported_remotes()}",
     ]
 
     try:
--- a/dvc/info.py
+++ b/dvc/info.py
@@ -2,6 +2,7 @@ import itertools
 import os
 import pathlib
 import platform
+import sys
 import uuid
 
 import psutil
@@ -20,6 +21,11 @@ if PKG is None:
 else:
     package = f"({PKG})"
 
+if sys.version_info < (3, 8):
+    import importlib_metadata
+else:
+    import importlib.metadata as importlib_metadata
+
 
 def get_dvc_info():
     info = [
@@ -27,7 +33,7 @@ def get_dvc_info():
         "---------------------------------",
         f"Platform: Python {platform.python_version()} on "
         f"{platform.platform()}",
-        f"Supports: {_get_supported_remotes()}",
+        f"Plugins:{_get_supported_remotes()}",
     ]
 
     try:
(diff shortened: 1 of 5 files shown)


dvc/info.py
```diff
@@ -20,14 +21,19 @@
...
+
 
 def get_dvc_info():
     info = [
         f"DVC version: {__version__} {package}",
         "---------------------------------",
         f"Platform: Python {platform.python_version()} on "
         f"{platform.platform()}",
-        f"Supports: {_get_supported_remotes()}",
+        f"Plugins:{_get_supported_remotes()}",
```
efiop, line 36: Let's keep `Supports` for now, since we don't have plugins just yet and if we introduce this new term we'll have to change docs too.
```suggestion
        f"Supports: {_get_supported_remotes()}",
```

skshetry (approved):

setup.py
```diff
@@ -78,6 +78,7 @@ def run(self):
     "pydot>=1.2.4",
     "speedcopy>=2.0.1; python_version < '3.8' and sys_platform == 'win32'",
     "dataclasses==0.7; python_version < '3.7'",
+    "importlib-metadata==1.4; python_version < '3.8'",
```
efiop, line 81: Should it be `>=1.4`? @isidentical

isidentical: We depend on the stdlib version when the python version is greater than that. So for the sake of consistency among all versions, the 3.8's stdlib importlib_metadata package is the equivalent of `importlib-metadata==1.4`.

efiop: @isidentical Right, but it is still backward compatible, it is just that 1.4 corresponds to the features introduced in 3.8, but we are still using stdlib's importlib-metadata in 3.9, so is there a reason to keep this requirement so strict?

isidentical: The `importlib-metadata` in 3.8 is going to stay always the same, so if anything is broken in `1.4` or if there is any missing feature we wouldn't be benefitting from it. Obviously we can just change it, but I am not really sure what would it gain to extend the range.

efiop: @isidentical Just relaxing requirements that have no need to be so strict. Ran into some problems in https://github.com/conda-forge/dvc-feedstock/pull/225/files/5e4708490f97d5395e7273c879a280901c92c11a..5cf119bcc0c486819728c2c2f0a469ce8f53a3ba , so was wondering if this strict requirement is needed.

isidentical:
Right. I don't think there is any harm of relaxing the requirements if the
current version is blocking something.

PR #6281, opened 2021-07-05
staging: optimize the number of info() calls

isidentical:
We do 2 `.exists()` and `.isdir()` call for the same path in the same context, so we can reduce them into a single `info()` call and retrieve all information we need in one go. If the file is not there, it will raise an FileNotExistsError too.

### DIFF
--- a/dvc/fs/hdfs.py
+++ b/dvc/fs/hdfs.py
@@ -1,3 +1,4 @@
+import errno
 import io
 import logging
 import os
@@ -220,8 +221,19 @@ class HDFSFileSystem(BaseFileSystem):
 
     def info(self, path_info):
         with self.hdfs(path_info) as hdfs:
-            finfo = hdfs.get_file_info(path_info.path)
-            return {"size": finfo.size}
+            from pyarrow.fs import FileType
+
+            file_info = hdfs.get_file_info(path_info.path)
+            if file_info.type is FileType.Directory:
+                kind = "directory"
+            elif file_info.type is FileType.File:
+                kind = "file"
+            else:
+                raise FileNotFoundError(
+                    errno.ENOENT, os.strerror(errno.ENOENT), path_info
+                )
+
+            return {"size": file_info.size, "type": kind}
 
     def checksum(self, path_info):
         return HashInfo(
--- a/dvc/fs/webhdfs.py
+++ b/dvc/fs/webhdfs.py
@@ -123,7 +123,7 @@ class WebHDFSFileSystem(BaseFileSystem):  # pylint:disable=abstract-method
 
     def info(self, path_info):
         st = self.hdfs_client.status(path_info.path)
-        return {"size": st["length"]}
+        return {"size": st["length"], "type": "file"}
 
     def checksum(self, path_info):
         return HashInfo(
--- a/dvc/objects/stage.py
+++ b/dvc/objects/stage.py
@@ -1,5 +1,3 @@
-import errno
-import os
 from concurrent.futures import ThreadPoolExecutor
 from functools import partial
 
@@ -174,11 +172,7 @@ def stage(odb, path_info, fs, name, upload=False, **kwargs):
         isinstance(path_info, str) or path_info.scheme == fs.scheme
     )
 
-    if not fs.exists(path_info):
-        raise FileNotFoundError(
-            errno.ENOENT, os.strerror(errno.ENOENT), path_info
-        )
-
+    details = fs.info(path_info)
     state = odb.state
     # pylint: disable=assignment-from-none
     hash_info = state.get(path_info, fs)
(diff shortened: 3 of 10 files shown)


dvc/objects/stage.py
```diff
@@ -174,11 +172,7 @@ def stage(odb, path_info, fs, name, upload=False, **kwargs):
         isinstance(path_info, str) or path_info.scheme == fs.scheme
     )
 
-    if not fs.exists(path_info):
-        raise FileNotFoundError(
-            errno.ENOENT, os.strerror(errno.ENOENT), path_info
-        )
-
+    details = fs.info(path_info)
```
efiop, line 175: Ideally, we shouldn't need to call this here at all, but that requires logic below to properly raise FileNotFound (e.g. `state.get()` in case when it is not a `StateNoop`, and `_get_file_obj` if it is a Noop). I'm almost sure that should work already, but need to check that. Could you please give it a look?

isidentical: We still need to make an info call, but just later in that case. Not sure what would be changed? Also isn't for remote providers (e.g s3) the state is `StateNoop`?

efiop: @isidentical We would live `isdir()` as it was before. In current PR we are calling info() once in place of old exits(), but then only using it instead of `isdir`.  So my point is rather about `fs.exists()` not being needed right at that place, so `info()` call is also not needed there.

> Also isn't for remote providers (e.g s3) the state is StateNoop?

Correct, which means that in current form `state.get()`'s behavior will differ between `State` and `StateNoop`, which might be problematic. But leaving `info()` where it is right now, makes that part of the code the slightest bit more confusing than it was before (at least it was an explicit exists() call).

Thinking about it a bit more, it becomes clear to me that this is probably a topic for a followup, as that `state.get` call could also be moved a bit. Anyway, let's merge and we'll get back to the follups in the future.

dvc/fs/hdfs.py
```diff
@@ -220,8 +221,19 @@ def isdir(self, path_info):
...
+
+            file_info = hdfs.get_file_info(path_info.path)
+            if file_info.type is FileType.Directory:
+                kind = "directory"
+            elif file_info.type is FileType.File:
+                kind = "file"
+            else:
+                raise FileNotFoundError(
+                    errno.ENOENT, os.strerror(errno.ENOENT), path_info
+                )
```
efiop, line 234: This `else` is not obvious, could you elaborate why we raise this if the type doesn't match dir or file?

efiop: E.g. if you look at fsspec's hdfs: https://github.com/intake/filesystem_spec/blob/9c75d6267d930b4fb01333d22bc2bebf4836116c/fsspec/implementations/hdfs.py#L105

dvc/fs/webhdfs.py
```diff
@@ -123,7 +123,7 @@ def exists(self, path_info) -> bool:
 
     def info(self, path_info):
         st = self.hdfs_client.status(path_info.path)
-        return {"size": st["length"]}
+        return {"size": st["length"], "type": "file"}
```
efiop, line 126: I guess we could try to use `type` from the status call here?

isidentical: For providers that lack of support for directories (gdrive, webdhfs) it is a static. Otherwise it would change the current behavior for staging (in the past `isdir()` would always return `False` for the same path, now `info()['type'] == 'directory'` returning True).

dvc/fs/hdfs.py
isidentical: `get_file_info()` won't raise any files even if the `path_info` does not exists. It would return `FileType.NotFound` (or `FileType.Unknown` but I don't think that is a case that we need to handle separately).

dvc/fs/webhdfs.py
efiop: webhdfs doesn't have directories? O_o Am I missing something?

dvc/fs/hdfs.py
efiop: @isidentical Ah, ok, makes sense. :+1:

dvc/fs/webhdfs.py
isidentical: There is no `isdir()` method implemented, even though there are dirs in webhdfs we don't support them.

dvc/objects/stage.py
isidentical: > Correct, which means that in current form state.get()'s behavior will differ between State and StateNoop, which might be problematic. But leaving info() where it is right now, makes that part of the code the slightest bit more confusing than it was before (at least it was an explicit exists() call).

If we defer this check to `state.get()`, then for example when you are staging something over s3 and if the directory doesn't exist yet this might cause a different behavior. So that was my worry about not shifting this and depending solely on `state.get`

efiop: @isidentical Yeah, makes sense! That's what I meant in my last comment too, we'll likely move that `state.get` in the future and will get back to this particular place in the code.

dvc/fs/webhdfs.py
efiop: @isidentical Ah, gotcha. So you are just matching the issues with our old implementation? :slightly_smiling_face: If we can report proper type here with reasonable effort - we should, or we run a chance of running into this odd bug in the future. I guess we need to finally start migrating to fsspec's webhdfs/hdfs implementations...

isidentical: Yes. This is a best-effort optimization PR, so I did not try to implement anything new or change any behavior.

efiop (approved):

PR #6286, opened 2021-07-06
add: ignore duplicated targets?

skshetry:
```console
# before
$ dvc add foo foo
ERROR: output 'foo' is already specified in stages:

# now
```
Screen Shot 2021-07-07 at 12 41 34

### DIFF
--- a/dvc/repo/add.py
+++ b/dvc/repo/add.py
@@ -43,7 +43,14 @@ def check_recursive_and_fname(args):
 
 
 def transform_targets(args):
-    args.targets = ensure_list(args.targets)
+    from funcy import count_reps
+
+    counts = count_reps(ensure_list(args.targets))
+    dupes = [key for key, count in counts.items() if count > 1]
+    if dupes:
+        msg = ", ".join(f"[b]{key}[/]" for key in dupes)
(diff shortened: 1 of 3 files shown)


dberenbaum:
What about a warning? It's not clear to me whether this would ever be intended behavior, or whether it would be important for someone to know they had added the same target twice.

skshetry:
@dberenbaum, it could happen with typos when similar targets exist. The main reason for this PR is that our internals cannot handle this properly (especially the graph checks), so we have to handle it at the top. I'll add a warning then.

dvc/repo/add.py
```diff
@@ -43,7 +43,14 @@ def check_recursive_and_fname(args):
...
 
 def transform_targets(args):
-    args.targets = ensure_list(args.targets)
+    from collections import Counter
+
+    counter = Counter(ensure_list(args.targets))
+    dupes = counter.most_common(n=2)
+    if dupes:
+        msg = ", ".join(f"[b]{key}[/]" for key, _ in dupes)
+        ui.error_write(f"ignoring duplicated targets: {msg}", styled=True)
```
skshetry, line 52: hmm, the capitalization of message (or, not) could be another topic for the UI discussion.

dberenbaum (approved):
LGTM!

PR #6294, opened 2021-07-06
Fix typos discovered by codespell

cclauss:
This PR was created with the command [`codespell --ignore-words-list="ba,datas,fo,uptodate" -w`](https://github.com/codespell-project/codespell)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
     rev: 21.6b0
   - hooks:
       - id: codespell
-        args: [--ignore-words-list="ba,datas,fo,uptodate"]
+        args: [--ignore-words-list='ba,datas,fo,uptodate']
     repo: https://github.com/codespell-project/codespell
     rev: v2.1.0
   - hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,8 @@ repos:
     rev: 21.6b0
   - hooks:
       - id: codespell
-        args: [--ignore-words-list='ba,datas,fo,uptodate']
+        args: 
+          - ignore-words-list ba,datas,fo,uptodate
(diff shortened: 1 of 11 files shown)


tests/unit/test_updater.py
```diff
@@ -83,7 +83,7 @@ def test_check_update_respect_config(mock_check, result, updater, mocker):
         ("0.0.2", "0.0.3", True),
         ("0.0.2", "0.0.1", False),
     ],
-    ids=["uptodate", "behind", "ahead"],
+    ids=["up-to-date", "behind", "ahead"],
```
skshetry, line 86: I'd prefer the previous one, as it is easier to type in the terminal, in case we need to run that particular test.

skshetry:
`codespell` seems to offer a pre-commit hook as well. @cclauss, what do you think? How is it in terms of false-positives?

@iterative/dvc WDYT?

tests/unit/test_updater.py
cclauss: Reverted.

cclauss:
The pre-commit works well and the syntax `codespell—ignore-words-list="uptodate”` can deal with any false reports.

pared (commented):
LGTM, seems like we could add it to pre-commit.

.pre-commit-config.yaml
```diff
@@ -7,6 +7,12 @@ repos:
         language_version: python3
     repo: https://github.com/ambv/black
     rev: 21.6b0
+  - hooks:
+      - id: codespell
+        args: 
+          - ignore-words-list ba,datas,fo,uptodate
```
skshetry, line 13: I think you need to split this into two:
```suggestion
          - ignore-words-list
          - ba,datas,fo,uptodate
```

But let's see if this also works first though.

skshetry (approved):
Thanks @cclauss for the PR. 🙂

pared (approved):

PR #6295, opened 2021-07-07
ssh: migrate to sshfs

### LINKED ISSUE #6064: ssh: QA the implementation / migrate to fsspec
fsspec itself comes with a [naive implementation](https://github.com/intake/filesystem_spec/blob/master/fsspec/implementations/sftp.py) for ssh, though it might require a bit of extra work.

isidentical:
This is a draft PR to check CI against all python versions. Resolves #6064


dvc/fs/fsspec_wrapper.py
```diff
@@ -29,6 +29,12 @@ def _with_bucket(self, path):
     def _strip_bucket(self, entry):
         return entry
 
+    @staticmethod
+    def _get_kwargs_from_urls(urlpath):
```
karajan1001, line 33: fsspec also have this methods `_get_kwargs_from_urls`, but
1. it is a private method.
2. not have been implemented in all of the `fs`.

dvc/fs/ssh.py
```diff
@@ -0,0 +1,133 @@
...
+
+    DEFAULT_PORT = 22
+    PARAM_CHECKSUM = "md5"
+
+    def _with_bucket(self, path):
+        if isinstance(path, self.PATH_CLS):
+            path_str = path.path
+        else:
+            path_str = super()._with_bucket(path)
+        return os.path.normpath(path_str)
```
isidentical, line 40: It seems like windows failures are related with using forward-slashes, so normalizing them here resolves all the CI failures.

isidentical:
I was able to trace the failure on 3.6 to a missing feature in `unittest.mock.mock_open()`. After opening the file, `asyncssh` directly iterates on it; 
https://github.com/ronf/asyncssh/blob/e1c36548fcd2de23890a59e5583c9f08d2b09f1e/asyncssh/config.py#L274-L277

But the `__iter__()` protocol to the `mock_open()` was brought in the later versions, so it wasn't reading anything in the 3.6;
https://docs.python.org/3/library/unittest.mock.html#unittest.mock.mock_open

Since we already use `mock` in various places instead of `unittest.mock` to catch up with new features, replacing that solved the issue.

tests/unit/fs/test_tree.py
```diff
@@ -39,11 +39,9 @@ def test_get_cloud_fs_validate(tmp_dir, dvc):
         default=False,
     )
 
-    assert get_cloud_fs(dvc, name="base")[1] == {"host": "example.com"}
-    assert get_cloud_fs(dvc, name="first")[1] == {
```
karajan1001, line 43: Excuse me, any difference between these two forms?

dvc/utils/fs.py
```diff
@@ -233,3 +234,21 @@ def walk_files(directory):
...
+        yield tmp_info
+    except BaseException:
+        # Handle stuff like KeyboardInterrupt
+        # as well as other errors that might
+        # arise during file transfer.
+        with suppress(FileNotFoundError):
+            fs.remove(tmp_info)
+        raise
+    else:
+        fs.move(tmp_info, to_info)
```
karajan1001, line 254: use `finally` here to cover other exceptions?

isidentical: What do you mean by 'cover other exceptions'? `except BaseException:` already covers everything.

tests/unit/fs/test_tree.py
isidentical: the difference is that, now we check the precise key instead of the whole dict which might change in time depending to the config.

tests/remotes/ssh.py
```diff
@@ -15,6 +18,39 @@
 )
 
 
+class SSHMockHandler(Handler):
```
isidentical, line 21: See this for the reason, https://github.com/carletes/mock-ssh-server/issues/22

efiop: @isidentical Please add a comment with a link, so we don't forget to remove it in the future.

isidentical:
```suggestion
# See: https://github.com/carletes/mock-ssh-server/issues/22
class SSHMockHandler(Handler):
```

efiop (approved):

isidentical:
Releasing the non-alpha version of sshfs (2021.07.0) and then merge.

efiop:
@isidentical Seems like there is some kind of issue between our ThreadPoolExecutor and internal channel throttling. This is what I get when I push a dataset with 4 files (dataset-registry/fashion-mnist/raw) to 127.0.0.1 with the current master. It works if I reduce it to `-j1`, but, if i try to just `dvc push` everything even `-j1` no longer helps and I'm getting this same issue.

[log.log](https://github.com/iterative/dvc/files/6895415/log.log)

isidentical:
I've identified the underlying issue(s), will submit a patch! Thanks for the report Ruslan, it actually uncovered an fsspec bug too.

efiop:
@isidentical Thanks for a quick response! :heart:

PR #6296, opened 2021-07-07
Upgrade to modern Python syntax

cclauss:
This pull request was created with the command [`pyupgrade --py36-plus **/*.py`](https://github.com/asottile/pyupgrade#readme)

Python f-strings simplify the code and [should speed up execution](https://www.scivision.dev/python-f-string-speed). 


dvc/command/status.py
```diff
@@ -39,7 +39,7 @@ def _show(self, status, indent=0):
 
         for key, value in status.items():
             if isinstance(value, str):
-                ui.write("{}{}{}".format(ind, self._normalize(value), key))
+                ui.write(f"{ind}{self._normalize(value)}{key}")
```
pmrowla, line 42: I'm honestly not sure about all of these changes, as I personally find `.format()` more readable in these cases, rather than having function calls inside f-strings.

Problably something the whole @iterative/dvc team needs to weigh in on

cclauss: I too resisted the change to this new format but now, after a year of use, I find f-strings easier to read and write.  It is a choice for your team to make but these lines are shorter on the page and a [3.62 times runtime speedup](https://www.scivision.dev/python-f-string-speed) does not hurt.

pared (commented):
LGTM, I have not preference regarding format/fstring approach. If fstrings are faster it makes sense to use them from my point of view.
This change also modifies `logger` calls which we should probably lazily evaluate, but all occurences seem to be in non-conditional places, so that probably won't change much.

dvc/command/status.py
efiop: I guess in this particular case even `format` one was a bit ugly and needed smth like `nvalue = self._normalize(value)` so that then you could use `nvalue` in it instead of putting that chunk of code directly into it. If we adjust it like that new f-string will actually look better than the old format :slightly_smiling_face:

dvc/dvcfile.py
```diff
@@ -200,9 +200,7 @@ def dump(self, stage, **kwargs):
         assert not isinstance(stage, PipelineStage)
         if self.verify:
             check_dvcfile_path(self.repo, self.path)
-        logger.debug(
-            "Saving information to '{file}'.".format(file=relpath(self.path))
-        )
+        logger.debug(f"Saving information to '{relpath(self.path)}'.")
```
efiop, line 203: To address @pared 's point about lazy logging

```suggestion
        logger.debug("Saving information to '%s'", relpath(self.path))
```

efiop:
Seems like it would make sense to have this as a pre-commit hook as well.

cclauss:
Let's hold off of adding to pre-commit because some of our Python 3.6 and 3.7 (but not Python 3.8 and 3.9) test fail using the Python `unittest.mock` vs [the backported module](https://github.com/iterative/dvc/blob/master/test_requirements.txt#L13).

skshetry:
> Let's hold off of adding to pre-commit because some of our Python 3.6 and 3.7 (but not Python 3.8 and 3.9) test fail using the Python `unittest.mock` vs [the backported module](https://github.com/iterative/dvc/blob/master/test_requirements.txt#L13).

That's expected as the `unittest.mock` in 3.6 is buggy. And, since `mock` is now a rolling backport, it makes more sense to use it compared to `unittest.mock` (though a bit risky).

dvc/dvcfile.py
skshetry: This is unnecessary, as relpath is already in string.
```pycon
>>> path = os.getcwd()
>>> %timeit logger.debug("Saving information to '%s'.", relpath(path))
69 µs ± 840 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
>>> %timeit logger.debug(f"Saving information to '{relpath(path)}'.")
66.5 µs ± 1.42 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

efiop: @skshetry That is not for explicit convertion, but rather for lazyness. Or am I missing something and f-strings lazy here?

skshetry: you are right, f-strings are not lazy, but I don't think lazy-formatting is expensive here, it's the string conversion that is, which is done eagerly. And, this is not an expensive code path, so I'd prefer f-strings here.

This is the result that I get when the logger is disabled.
```pycon
>>> %timeit logger.debug(f"Saving information to '{relpath(path)}'.")
29 µs ± 1.03 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

>>> %timeit logger.debug("Saving information to '%s'.", relpath(path))
29.8 µs ± 790 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

(they are very close and flip flop on which one is better.)

efiop: right, I didn't mean that is was unbearably expensive here, just was thinking about consistency and developing a habit/rule so we always use lazy notation in logger. So could keep it as is, I'm. totally fine with this.

efiop (approved):
ok, looks like this works fine so far, so let's merge it as is. A pre-commit hook would be nice though 🙂

PR #6300, opened 2021-07-09
 introduce index

skshetry:

### DIFF
--- a/dvc/output.py
+++ b/dvc/output.py
@@ -1004,6 +1004,18 @@ class Output:
     def fspath(self):
         return self.path_info.fspath
 
+    @property
+    def is_decorated(self) -> bool:
+        return self.is_metric or self.is_plot
+
+    @property
+    def is_metric(self) -> bool:
+        return bool(self.metric) or bool(self.live)
+
+    @property
+    def is_plot(self) -> bool:
+        return bool(self.plot)
+
 
 ARTIFACT_SCHEMA = {
     **CHECKSUMS_SCHEMA,
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -3,9 +3,9 @@ import os
 from collections import defaultdict
 from contextlib import contextmanager
 from functools import wraps
-from typing import TYPE_CHECKING, Callable, Optional
+from typing import TYPE_CHECKING, Callable, Optional, Set
 
-from funcy import cached_property, cat
+from funcy import cached_property
 
 from dvc.exceptions import FileMissingError
 from dvc.exceptions import IsADirectoryError as DvcIsADirectoryError
@@ -14,11 +14,9 @@ from dvc.ignore import DvcIgnoreFilter
 from dvc.path_info import PathInfo
 from dvc.utils.fs import path_isin
 
-from .graph import build_graph, build_outs_graph, get_pipelines
-from .trie import build_outs_trie
-
 if TYPE_CHECKING:
     from dvc.fs.base import BaseFileSystem
+    from dvc.objects.file import HashFile
     from dvc.scm import Base
 
 
@@ -57,7 +55,6 @@ class Repo:
     DVC_DIR = ".dvc"
 
     from dvc.repo.add import add
-    from dvc.repo.brancher import brancher
     from dvc.repo.checkout import checkout
     from dvc.repo.commit import commit
     from dvc.repo.destroy import destroy
@@ -207,6 +204,12 @@ class Repo:
     def __str__(self):
         return self.url or self.root_dir
 
+    @cached_property
+    def index(self):
+        from dvc.repo.index import Index
+
+        return Index(self, self.fs)
+
     @staticmethod
     def open(url, *args, **kwargs):
         if url is None:
@@ -323,23 +326,10 @@ class Repo:
 
         self.scm.ignore_list(flist)
 
-    def check_modified_graph(self, new_stages, old_stages=None):
-        """Generate graph including the new stage to check for errors"""
-        # Building graph might be costly for the ones with many DVC-files,
-        # so we provide this undocumented hack to skip it. See [1] for
-        # more details. The hack can be used as:
-        #
-        #     repo = Repo(...)
-        #     repo._skip_graph_checks = True
-        #     repo.add(...)
-        #
-        # A user should care about not duplicating outs and not adding cycles,
-        # otherwise DVC might have an undefined behaviour.
-        #
-        # [1] https://github.com/iterative/dvc/issues/2671
-        if not getattr(self, "_skip_graph_checks", False):
-            existing_stages = self.stages if old_stages is None else old_stages
-            build_graph(existing_stages + new_stages)
+    def brancher(self, *args, **kwargs):
+        from dvc.repo.brancher import brancher
+
+        return brancher(self, *args, **kwargs)
 
     def used_objs(
         self,
@@ -373,16 +363,14 @@ class Repo:
         """
         used = defaultdict(set)
 
-        def _add_suffix(objs, suffix):
-            from dvc.objects.tree import Tree
+        def _add_suffix(objs: Set["HashFile"], suffix: str) -> None:
+            from itertools import chain
 
-            for obj in objs:
+            from dvc.objects import iterobjs
+
+            for obj in chain.from_iterable(map(iterobjs, objs)):
                 if obj.name is not None:
                     obj.name += suffix
-                if isinstance(obj, Tree):
-                    for _, entry_obj in obj:
-                        if entry_obj.name is not None:
-                            entry_obj.name += suffix
 
         for branch in self.brancher(
             revs=revs,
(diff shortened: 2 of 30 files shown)


pared:
Related #6039

dvc/repo/index.py
```diff
@@ -0,0 +1,275 @@
...
+
+            stages - provides direct access to this collection
+            outputs - provides direct access to the outputs
+            objects - provides direct access to the objects
+            graph -
+            ... and many more.
+
+        Index also provides ways to slice and dice this collection.
+        Some `views` might not make sense when sliced (eg: pipelines/graph).
+        """
```
pared, line 51: +1

dvc/repo/index.py
```diff
@@ -0,0 +1,275 @@
...
+    @cached_property
+    def _stages(self) -> "PSet['Stage']":
+        """Index without stages cannot exist.
+        So, stages being lazy, should not be depended upon.
+        """
+        from pyrsistent import pset
+
+        onerror = self.repo.stage_collection_error_handler
+        # pylint: disable=protected-access
+        return pset(self.stage_collector._collect_repo(onerror=onerror))
```
skshetry, line 68: The reason why this is immutable is that I feel that indices are inherently immutable. Another reason is that most of the indices are immutable in DVC, those that belong to some commits. Only that belongs to the workspace can be mutable, but generalizing this, I choose the lowest common denominator and made it immutable. Even if we don't use persistent data structures, I think it should behave in an immutable way.

dvc/repo/__init__.py
```diff
@@ -512,11 +485,7 @@ def close(self):
     def _reset(self):
         self.state.close()
         self.scm._reset()  # pylint: disable=protected-access
-        self.__dict__.pop("outs_trie", None)
-        self.__dict__.pop("outs_graph", None)
-        self.__dict__.pop("graph", None)
-        self.__dict__.pop("stages", None)
-        self.__dict__.pop("pipelines", None)
```
efiop, line 519: image

dvc/output.py
```diff
@@ -982,6 +980,18 @@ def merge(self, ancestor, other):
     def fspath(self):
         return self.path_info.fspath
 
+    @property
+    def is_decorated(self) -> bool:
+        return self.is_metric or self.is_plot
```
efiop, line 985: Not sure I understand why `decorated`, could you elaborate on naming?

dvc/repo/index.py
```diff
@@ -0,0 +1,275 @@
...
+
+        return self.filter(is_stage_inside_path)
+
+    @property
+    def outputs(self) -> Iterator["Output"]:
+        for stage in self:
+            yield from stage.outs
+
+    @property
+    def decorated_outputs(self) -> Iterator["Output"]:
```
efiop, line 114: Doesn't look like this is used anywhere.

dvc/repo/index.py
```diff
@@ -0,0 +1,275 @@
...
+                yield output
+
+    @property
+    def plots(self) -> Iterator["Output"]:
+        for output in self.outputs:
+            if output.is_plot:
+                yield output
+
+    @property
+    def dependencies(self) -> Iterator["Output"]:
```
efiop, line 132:
```suggestion
    def deps(self) -> Iterator["Dependency"]:
```
?

dvc/repo/index.py
```diff
@@ -0,0 +1,275 @@
...
+
+    @cached_property
+    def deterministic_hash(self) -> str:
+        # we can use this to optimize and skip opening some indices
+        # eg. on push/pull/fetch/gc --all-commits
+        import hashlib
+        import json
+
+        d = json.dumps(self.dumpd(), sort_keys=True).encode("utf-8")
+        return hashlib.md5(d).hexdigest()
```
efiop, line 254: We have `dict_md5` in utils, looks like you could use it.

efiop (commented):
Looks great! Let's clean it up, review and merge it 🙂

dvc/output.py
skshetry: `decorated` as in bearing extra items like `plots`, `metrics`, etc.
> decorate - **make (something)** look more attractive **by adding extra items or images to it**.

pared:
I just noted that I didn't wrote: LGTM so far.

dvc/repo/index.py
```diff
@@ -0,0 +1,275 @@
...
+        """
+        from pyrsistent import pset
+
+        onerror = self.repo.stage_collection_error_handler
+        # pylint: disable=protected-access
+        return pset(self.stage_collector._collect_repo(onerror=onerror))
+
+    @property
+    def stages(self) -> "PSet['Stage']":
+        return self._stages
```
Suor, line 72: Why do we need `._stages` prop? Why don't move everything here and make it cached property?

dvc/repo/index.py
```diff
@@ -0,0 +1,275 @@
...
+        return len(self.stages)
+
+    def __contains__(self, stage: "Stage") -> bool:
+        return stage in self.stages
+
+    def __iter__(self) -> Iterator["Stage"]:
+        yield from self.stages
+
+    def filter(self, filter_fn: Callable[["Stage"], bool]) -> "Index":
+        from pyrsistent import pset
```
Suor, line 95: It looks like you are importing it in several places like that.

tests/func/test_repo_index.py
```diff
@@ -0,0 +1,171 @@
...
+    dup_stage1 = Stage(dvc, path="path1")
+    dup_stage2 = Stage(dvc, path="path2")
+    dup_index = index.update([dup_stage1, dup_stage2])
+    assert not index.stages
+    assert len(new_index) == 1
+    assert new_index.stages[0] is new_stage
+    assert set(map(id, dup_index.stages)) == {id(dup_stage1), id(dup_stage2)}
+
+
+def test_discard_remove(dvc):
```
skshetry, line 154: Few tests are remaining, the implementation should be more or less complete.

dvc/repo/index.py
```diff
@@ -0,0 +1,285 @@
...
+        # pylint: disable=pointless-statement
+        print("no of stages", len(index.stages))
+    with log_durations(print, "building graph"):
+        index.build_graph()
+    with log_durations(print, "calculating hash"):
+        print(index.identifier)
+    with log_durations(print, "updating"):
+        index2 = index.update(index.stages)
+    with log_durations(print, "calculating hash"):
+        print(index2.identifier)
```
efiop, line 285:
```suggestion
```

skshetry: I think I'll keep these as it is quite helpful in checking bugs/performance issues quickly, unless there is some strong reason against this.

efiop: @skshetry This is not suitable for the core codebase, just create a test instead.

efiop: @skshetry Looks like we also still have these in parsing and in ui. They really should not be there.

skshetry: But I cannot run test on a real project, with this, I can do that by just `python dvc/repo/index.py`.

skshetry: > They really should not be there.

Could you please elaborate in why you don't like this?

I really like having small snippets as it allows me to document  example usages, play with and test quite easily.

Other projects also use them in a similar manner, eg: [rich](https://github.com/willmcgugan/rich/blob/542cfe1a9681569123366d51d5bb2e53eef8c62c/rich/table.py#L845).

skshetry: > no one is using it except you

Right, because it's not merged yet. Like I said above, it acts as a quick litmus test, especially for UI where I can quickly visually inspect if there's any issues.

> Keeping such snippets around is very ugly.

That's quite subjective, it's not a utility at all.

efiop: Oops, sorry, didn't see your last comment.

> Could you please elaborate in why you don't like this?
> I really like having small snippets as it allows me to document example usages, play with and test quite easily.

It is a test script that is not used anywhere in the code or during normal workflow. It is also not a generic debugging mechanism like --pdb/--cprofile/etc. It looks to me like an arbitrary test code that no one knows about, which makes the codebase smell, as it is percieved like some debugging code that we forgot to remove.

> Other projects also use them in a similar manner, eg: rich.

It might be more justifiable for rich, because it is a library, but it still looks to me as a code snippet that someone forgot to remove.

efiop: > Right, because it's not merged yet. Like I said above, it acts as a quick litmus test, especially for UI where I can quickly visually inspect if there's any issues.

The functionality makes sense to me - to quickly check if our subsystems work correctly. I only have a problem with code snippets scattered around the codebase, that no one even knows exist and could be ran. How about we create a command (`dvc test`? `dvc doctor`? `dvc debug`?), where we will collect all such self-checks, so that there is one place to test if dvc subsystems are operating correctly?

skshetry: For having a central way of running all these, it requires certain pre-conditions/post-conditions (eg: this needs to always be run inside a DVC repo). And, having a central way is not the point of this, tests: end-to-end or unit is there for that.

> but it still looks to me as a code snippet that someone forgot to remove.

It is very much intentional. They use it for showing example usages and demonstration, not to add quick visual inspection.

> no one even knows exist and could be ran

It's okay to not know they exist, as they are very local to the subsystem. Whenever someone happens to work in this module, they'll take a note of `if __name__ == "main"`. When refactoring, they'll be able to grep it or find usages of it to fix. And, as it's not critical to us, it's okay if it breaks occasionally. Though the parsing/UI has never been broken, they still work.

>  one place to test

The parsing one is a test and uses doctest, so we could run it always. But there is not much benefit on running others, as they serve different purpose.

efiop: @skshetry Are there any other examples besides rich? I feel like it is a very specific library, I would love to see some other examples in serious projects (cli tools would be great). Just trying to understand when and how this is used.

CC @iterative/dvc , guys, wdyt about this practice?

casperdcl: I'd completely agree with moving this to `tests/`

efiop: Ok, so looks like this was indeed a practice that wasn't aware of. Thanks to @isidentical and @skshetry for pointing out that this is used in things like cpython, doctests, rich. However, I wasn't able to find a single case in any of the projects that I've looked through (e.g. numpy, pyinstaller, duliwch, boto, moto, scipy, spacy, etc), so this still feels like a pretty cryptic practice to me, so I'm not personally 100% sold on it, it feels like these snippets belong somewhere else. But there doesn't seem to be much opposition to it and we already have a few of these in the codebase, so let's keep it for now, and get back to discussing this practice in general on retro or elsewhere.

efiop (approved):

PR #6307, opened 2021-07-13
oss: migrate to ossfs

### LINKED ISSUE #6177: oss remote Migration to ossfs.

karajan1001:
Fix #6177, migrate to ossfs

come from #6178 

Performance summary: 
> The latest version runs 1.5 times faster  in `gc` and `push` than the version we are using, and more than 2 times in `pull`.

| time cost \ version| current version | last week version | reuse session version  | dircache version (latest) |
|:--:|:--:|:--:|:--:|:--:|
| gc | 100% |  +115% | 70% |  68% |
| pull | 100% | 171% |  118% | 36% |
| push | 100% | 97% | 64% |  72 %|

image
The results from the current version compare to reuse of sessions optimization.

image
The results from the current version compare to reuse of sessions optimization.

### DIFF
--- a/dvc/fs/oss.py
+++ b/dvc/fs/oss.py
@@ -53,15 +53,15 @@ class OSSFileSystem(ObjectFSWrapper):
     def _download(
         self, from_info, to_file, name=None, no_progress_bar=False, **pbar_args
     ):
-        total = self.fs.size(from_info.url)
+        # total = self.fs.size(from_info.url)
         with Tqdm(
             disable=no_progress_bar,
-            total=total,
+            # total=total,
             bytes=True,
             desc=name,
             **pbar_args,
         ) as pbar:
-            self.fs.download_file(
+            self.fs.get_file(
                 from_info.url,
                 to_file,
                 progress_callback=pbar.update_to,
--- a/dvc/fs/oss.py
+++ b/dvc/fs/oss.py
@@ -19,12 +19,18 @@ class OSSFileSystem(ObjectFSWrapper):
     PATH_CLS = CloudURLInfo
     REQUIRES = {"ossfs": "ossfs"}
     PARAM_CHECKSUM = "etag"
+    COPY_POLL_SECONDS = 5
+    LIST_OBJECT_PAGE_SIZE = 100
     DETAIL_FIELDS = frozenset(("etag", "size"))
 
     def _prepare_credentials(self, **config):
         login_info = {}
-        login_info["key"] = config.get("oss_key_id")
-        login_info["secret"] = config.get("oss_key_secret")
+        login_info["key"] = config.get("oss_key_id") or os.getenv(
+            "OSS_ACCESS_KEY_ID"
+        )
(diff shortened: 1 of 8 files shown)


karajan1001:
[asciicast](https://asciinema.org/a/CPAuGIWn9RkHdlar0uPOEFhp7)

dvc/fs/oss.py
```diff
@@ -8,124 +8,73 @@
...
     def remove(self, path_info):
-        if path_info.scheme != self.scheme:
-            raise NotImplementedError
-
-        logger.debug(f"Removing oss://{path_info}")
-        self._get_bucket(path_info.bucket).delete_object(path_info.path)
-
-    def _upload_fobj(self, fobj, to_info, **kwargs):
-        self._get_bucket(to_info.bucket).put_object(to_info.path, fobj)
+        self.fs.rm_file(self._with_bucket(path_info))
```
isidentical, line 45: Why do you re-implement this? 
https://github.com/iterative/dvc/blob/f45311e658f4ce0ec537dfc421a4c45aa8efb13e/dvc/fs/fsspec_wrapper.py#L112-L113

karajan1001: `rm` equals to `expand` and `rm_file` which is slower than directly `rm_file` if we had already known the removed path_info is a `object`.

isidentical: Right. For a single file the behavior should be close, but it is possible that `expand_paths` calls on some providers might be a little bit problematic. Apparently s3fs doesn't implement `_rm_file()`, so it is a bit problematic to just switch it of on the `FsspecWrapper` but once it is done we should be safe with replacing this for all providers. Nice catch!

karajan1001: Yes looks like s3fs has no rm_file.

https://github.com/intake/filesystem_spec/blob/6de675e24adcbc1a8c3d74f5956a79bc95c34c2b/fsspec/spec.py#L884

https://github.com/dask/s3fs/blob/fc4489cfe749b2d3de62bde926ed93c05e27ce7a/s3fs/core.py#L1640

But I think that the best solution for `gc` is to call a `batch_delete` ( don't know the exact name of it in other remotes).

isidentical: Yea, that is among my plans. But we first probably want to go with parallelization of gc and then add this special layer which is going to be tricky since only a handful of providers support it.

isidentical:
Now that we are using the real OSS, do we need this?
https://github.com/iterative/dvc/blob/f45311e658f4ce0ec537dfc421a4c45aa8efb13e/tests/func/test_data_cloud.py#L35-L41

efiop:
@isidentical Great catch, indeed we don't need it anymore. We can safely remove xfail now.

isidentical (approved):
LGTM

efiop:

For the record: now waiting for conda package https://github.com/conda-forge/staged-recipes/pull/15612

Issue #6262, opened 2021-07-01
libify dvc.odb and objects


skshetry:
We should try to push io to the edge. `odb` being a core part of the DVC should be a good place to start.

I did a quick search, and it does not seem that we use the logger calls for the UI. so we are at a good place. We should change the logger to only output during the `--verbose` mode and hide otherwise. There is a warning and an error, but on both we do raise Exception, so we could reduce the level for both of them too.

https://github.com/iterative/dvc/blob/ce45ec8a2864f4d2ad02bb27c158f00f532affe2/dvc/objects/db/base.py#L153-L156

https://github.com/iterative/dvc/blob/ce45ec8a2864f4d2ad02bb27c158f00f532affe2/dvc/objects/tree.py#L112-L117

```console 
$ rg "logger.(info|warn|warning|debug|error|exception)"        
db/base.py
95:                logger.debug("'%s' file already exists, skipping", path_info)
154:            logger.warning("corrupted cache file '%s'.", obj.path_info)
197:                logger.debug(
207:                logger.debug(
254:            logger.debug(f"Estimated remote size: {remote_size} files")
320:        logger.debug(
369:        logger.debug(
456:            logger.debug(
466:        logger.debug("Querying '{}' hashes via traverse".format(len(hashes)))

db/local.py
108:            logger.debug("Unprotecting '%s'", path)
121:            logger.debug(

tree.py
113:            logger.error(
```

Also, we have 7 progress bars created inside them. We should try replacing them through callback/progress args. 

```console
$ rg "Tqdm\("
stage.py
82:    with Tqdm(

db/base.py
233:        with Tqdm(
293:        with Tqdm(
372:        with Tqdm(

db/ssh.py
56:        with Tqdm(

db/local.py
69:        for hash_ in Tqdm(

__init__.py
24:            for future in Tqdm(
```

Let's not add any UI messages or progress bars there. We might be able to extend this as we go outside eg: outputs/stages, etc. It looks very doable for the odb though, and let's keep it at that. 

@iterative/dvc

skshetry:
Calling it done at this point. We have extracted those into two libraries where more work will be done: https://github.com/iterative/dvc-objects and https://github.com/iterative/dvc-data.

PR #6322, opened 2021-07-15
 fsspec: implement callback mixins 

isidentical:
[asciicast](https://asciinema.org/a/NJlF0wwd2hSLb3yPqe7jg1N9C)

### DIFF
--- a/dvc/fs/fsspec_wrapper.py
+++ b/dvc/fs/fsspec_wrapper.py
@@ -7,6 +7,7 @@ from funcy import cached_property
 from dvc.progress import Tqdm
 
 from .base import BaseFileSystem
+from .local import LocalFileSystem
 
 
 # pylint: disable=no-member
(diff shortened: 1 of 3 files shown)


dvc/fs/fsspec_wrapper.py
```diff
@@ -215,3 +216,44 @@ def find(self, path_info, detail=False, prefix=None):
             return None
 
         yield from self._strip_buckets(files, detail=detail)
+
+
+_LOCAL_FS = LocalFileSystem()
+
+
+class CallbackMixin:
```
skshetry, line 224: I think it's a good opportunity to consider getting rid of progress bars from the filesystem and lifting these up. Related: #6262

isidentical: Interesting, though I don't see anything related to 'where to lift' these progress. Mind explaining it a bit further?

skshetry: Looking at it from the CLI, I expect these *updates* to come from each component that I am interacting with, rather than filesystem creating progress bars itself. So, I think the responsibility falls on individual components like `odb`/`checkout`/`repro`/`experiments` to provide these. These days I am leaning more and more towards the event design patterns (similar to [`pyee`](https://pyee.readthedocs.io/en/latest/)) with observable/pub-sub.
The decoupling that we will is good, but the debugging gets harder, it's a tradeoff.

I do think that the solutions for all of these components might differ based on their nature, so I would love to hear out proposals for #6262 for `odb` and friends. Not trying to block your PR, it looks good.

dvc/fs/oss.py
```diff
@@ -43,38 +42,3 @@ def fs(self):
 
     def remove(self, path_info):
         self.fs.rm_file(self._with_bucket(path_info))
-
-    def _upload(
-        self, from_file, to_info, name=None, no_progress_bar=False, **kwargs
```
karajan1001, line 48: @efiop 
Removing these lines will slow down `pull` by 20%, and `push` by about 100%.

isidentical: wait a minute, why would it slow down push/pull? There is no behavioral change

karajan1001: @isidentical , Sorry, hadn't seen that the `_upload` and `_download` had been updated. The old `open(fileA, 'w')` version are slower.

karajan1001:
New version (2021.7.5) of ossfs had been released with the new callbacks.

dvc/progress.py
```diff
@@ -161,3 +165,28 @@ def format_dict(self):
...
+        information is not already present on the caller. In that case, we'll
+        retrieve the size from an ``info()`` call."""
+        if size is None:
+            size = self.fs.info(self.path_info)["size"]
+        self.progress_bar.total = size
+        self.progress_bar.refresh()
+
+    def relative_update(self, inc=1):
+        self.progress_bar.update(inc)
+        self.progress_bar.refresh()
```
skshetry, line 188: In `tqdm`, `update()` decides when to refresh the progress bar, so you don't need to force it with `refresh`, might be slower otherwise. Same on the `absolute_update`.

efiop (approved):

PR #6326, opened 2021-07-16
setup: bump ossfs==2021.7.5

isidentical:


karajan1001:
Any better methods to test this?

https://github.com/iterative/dvc/blob/6c8673b1582fcf1a92264b90885e8c4f8df4f63d/tests/unit/remote/test_oss.py#L19-L21

is somewhat useless. Maybe we should delete them?

isidentical:
> is somewhat useless. Maybe we should delete them?

We can just use `fs_args`.

isidentical:
CC: @karajan1001

karajan1001 (approved):

PR #6332, opened 2021-07-19
Push running checkpoint to remote

### LINKED ISSUE #6182: checkpoints: flag to `exp push` && `push`
Provide a flag (e.g. `dvclive.next_step(push=True)` or `make_checkpoint(push=True)` or `DVC_EXP_AUTO_PUSH=<git_remote_name_or_url|true>`) to push things per-checkpoint.

Helps with resuming a long-running CI job after timeout.

Part of https://github.com/iterative/cml/issues/560.

karajan1001:
Fix #6182

  https://github.com/iterative/dvc.org/pull/2672


dvc/repo/experiments/executor/base.py
```diff
@@ -464,6 +466,23 @@ def checkpoint_callback(
...
             )
+
+            git_remote = os.environ.get("DVC_EXP_AUTO_PUSH", None)
+            if git_remote:
+                from dvc.repo.experiments.push import push
+
+                branch = dvc.experiments.get_branch_by_rev(
+                    exp_rev, allow_multiple=None
+                )
+                branch_name = ExpRefInfo.from_ref(branch).name
```
pmrowla, line 477:
```suggestion
                branch = scm.get_ref(EXEC_BRANCH, follow=False)
```
We don't need to do the reverse rev->branch lookup from here, since we set the appropriate refs ourselves during exp execution (this `branch` will be the complete `refs/exps/...` refname and can be passed directly into `experiments.push`)

dvc/repo/experiments/executor/base.py
```diff
@@ -464,6 +466,23 @@ def checkpoint_callback(
...
+                push(
+                    dvc,
+                    git_remote,
+                    branch_name,
+                    push_cache=True,
+                    run_cache=True,
+                )
+                logger.info({"pushed": branch_name})
```
pmrowla, line 485: This push also needs to happen for the final `executor.commit()` at the end of `executor.reproduce()` (line 359).

So maybe this env var check + push should maybe be done inside `commit()` so we don't end up missing additional calls in the future in case we add more places that commits can happen.

pmrowla (changes requested):
looks good, just a couple small comments

dvc/repo/experiments/executor/base.py
dberenbaum: This would also create more consistency between checkpoint and non-checkpoint experiments in cml. `dvc exp push` probably shouldn't be needed in the workflow in either case.

dvc/repo/experiments/executor/base.py
```diff
@@ -464,6 +466,20 @@ def checkpoint_callback(
             exp_rev = cls.commit(
                 scm, exp_hash, exp_name=name, force=force, checkpoint=True
             )
+
+            git_remote = os.environ.get("DVC_EXP_AUTO_PUSH", None)
```
efiop, line 470: Please put this env var into `dvc/env.py` and use it as a var.

casperdcl: what about a different meaning, e.g. `DVC_EXP_AUTO_PUSH=10` meaning "push every 10 checkpoints?"

karajan1001: @casperdcl 
`DVC_EXP_AUTO_PUSH` is the remote git repo for auto push. Maybe this name needs to be modified. But it is better to discuss this problem in the PR for `dvc.org`

tests/func/experiments/test_checkpoints.py
```diff
@@ -188,3 +192,73 @@ def test_resume_non_head_checkpoint(
...
+
+    # resume the remote checkpoint
+    os.environ.pop("DVC_EXP_AUTO_PUSH")
+    with git_upstream.chdir():
+        git_upstream.dvc.experiments.run(checkpoint_stage.addressing)
+    assert (git_upstream / "foo").read_text() == "6"
+
+
+def test_auto_push_error_url(dvc, scm, checkpoint_stage, local_remote):
+    os.environ["DVC_EXP_AUTO_PUSH"] = "none"
```
karajan1001, line 237: How to test the error message from an exception? None of the  `caplog.text`, `capsys.err` and `capsys.out` show this out besides the exception was inside the executor hard to be tested by `pytest`

tests/func/experiments/test_checkpoints.py
```diff
@@ -188,3 +192,73 @@ def test_resume_non_head_checkpoint(
...
+        == {}
+    )
+
+
+def test_auto_push_no_remote(dvc, scm, checkpoint_stage, git_upstream):
+    os.environ["DVC_EXP_AUTO_PUSH"] = git_upstream.url
+    assert (
+        dvc.experiments.run(checkpoint_stage.addressing, params=["foo=2"])
+        == {}
+    )
```
karajan1001, line 249: @dberenbaum 
Currently, all of the remote push errors occur after the stage repro, so the repro will stop after one epoch. I'm not sure what is the best way to handle it.
1. leave the repo to be after epoch1?
2. roll back to the init stage?
3. examine it before stage repro?
4. Don't stop running, just give users a warning.

tests/func/experiments/test_checkpoints.py
pmrowla: As you noted, you can't test for exceptions and logging messages here like usual, since experiment execution happens in a separate process.

But any exception that is raised in the experiment will be returned by `future.exception()` and collected in the main process:
https://github.com/iterative/dvc/blob/1e43c303a7bb7c2c34c721554c859ae3eb54c938/dvc/repo/experiments/__init__.py#L710

If you just want to test that a particular exception type generates a specific message, you can just patch the `future.exception()` call so that it returns the exception you want:
https://github.com/iterative/dvc/blob/master/dvc/repo/experiments/__init__.py#L710

You could also patch `future.exception` and check all of the times it is actually called, and look for whatever particular exception you are testing for

dvc/repo/experiments/executor/base.py
```diff
@@ -450,9 +452,32 @@ def _repro_args(cls, dvc):
...
+    @staticmethod
+    def _auto_push(git_remote: str, dvc: "Repo", scm: "Git"):
+        from dvc.repo.experiments.push import push
+
+        if git_remote == scm.root_dir:
+            logger.warning(
+                f"{DVC_EXP_AUTO_PUSH} {git_remote} is the running "
+                "repository auto push will not work"
+            )
+            return
```
pmrowla, line 467: Have you tested this behavior (and did the push actually fail)?

I would not expect this to error out, pushing to yourself is possible with CLI git, and just does nothing, since the refs will always be up to date.

A warning might be appropriate since the user was probably not trying to push to self, but I don't think this should be an error condition

dvc/repo/experiments/executor/base.py
```diff
@@ -450,9 +452,32 @@ def _repro_args(cls, dvc):
...
+
+        branch = scm.get_ref(EXEC_BRANCH, follow=False)
+        branch_name = ExpRefInfo.from_ref(branch).name
+        push(
+            dvc,
+            git_remote,
+            branch_name,
```
pmrowla, line 474: It would be better to call `experiments.push` with the full refname string returned by `scm.get_ref()` here, instead of using this `branch_name` (which is the shortened exp name). Otherwise we will end up doing the entire shortname to full ref string lookup/conversion again in `experiments.push`.

dvc/repo/experiments/executor/base.py
```diff
@@ -450,9 +452,32 @@ def _repro_args(cls, dvc):
...
+            push_cache=True,
+            run_cache=True,
+        )
+        logger.info({"pushed": branch_name})
```
pmrowla, line 478: I think this is probably unneeded (and too verbose). We already log when a push is completed in `experiments.push` at the `debug` level.

dvc/repo/experiments/executor/base.py
```diff
@@ -464,6 +489,10 @@ def checkpoint_callback(
             exp_rev = cls.commit(
                 scm, exp_hash, exp_name=name, force=force, checkpoint=True
             )
+
+            git_remote = os.environ.get(DVC_EXP_AUTO_PUSH, None)
+            if git_remote:
+                cls._auto_push(git_remote, dvc, scm)
```
pmrowla, line 498: This check and push still needs to be moved from `checkpoint_callback()` into the end of `commit()`. With the current behavior, the final commit at the end of `reproduce()` is not pushed.

tests/func/experiments/test_checkpoints.py
```diff
@@ -188,3 +192,73 @@ def test_resume_non_head_checkpoint(
...
+    git_upstream.scm.fetch_refspecs(str(tmp_dir), ["master:master"])
+
+    # without auto push
+    results = dvc.experiments.run(checkpoint_stage.addressing)
+    exp = first(results)
+    ref_info = first(exp_refs_by_rev(scm, exp))
+    assert git_upstream.scm.get_ref(str(ref_info)) is None
+
+    # add auto push
+    os.environ["DVC_EXP_AUTO_PUSH"] = remote
```
pmrowla, line 212: all of the test cases that reference this env var should probably use `dvc.env.DVC_EXP_AUTO_PUSH` instead of the raw string

pmrowla (changes requested):

dvc/repo/experiments/executor/base.py
pmrowla: Actually now that I'm thinking about it, one situation where this "push to self" behavior may be useful is if the user wants to ensure that DVC cache is always pushed for their experiments, but does not care about pushing the individual exp commits to a git remote.

dvc/repo/experiments/executor/base.py
casperdcl: > PR for `dvc.org`

I don't follow. I understand what you're using `DVC_EXP_AUTO_PUSH` for. I'm asking for:

- `DVC_EXP_AUTO_PUSH` to be changed to mean something else (push every N epochs)
- or a new variable `DVC_EXP_AUTO_PUSH_EVERY` added (push every N epochs)
- or 2 different variables with less ambiguous names `DVC_EXP_PUSH_CHECKPOINT=0`, `DVC_EXP_PUSH_REMOTE=""`

Basically I think a "push every N epochs" config should be supported in this PR rather than in a follow-up so that we can keep API consistent. WDYT @efiop?

dvc/repo/experiments/executor/base.py
```diff
@@ -457,25 +457,21 @@ def _repro_args(cls, dvc):
...
-            dvc,
+        dvc.experiments.push(
             git_remote,
-            branch_name,
+            branch,
             push_cache=True,
             run_cache=True,
         )
-        logger.info({"pushed": branch_name})
+        logger.debug(f"pushed {branch} to {git_remote}")
```
pmrowla, line 474:
```suggestion
```
This can just be removed, it ends up being a duplicate of the logging statement that is already done in `experiments.push`

dvc/repo/experiments/executor/base.py
```diff
@@ -457,25 +457,21 @@ def _repro_args(cls, dvc):
 
     @staticmethod
     def _auto_push(git_remote: str, dvc: "Repo", scm: "Git"):
-        from dvc.repo.experiments.push import push
-
         if git_remote == scm.root_dir:
             logger.warning(
```
pmrowla, line 461: one other comment on this is that we will end up repeating this warning for every commit, it would be better if we just checked for this condition once at the start of `reproduce()`

tests/func/experiments/test_checkpoints.py
dberenbaum: Good point!

From user standpoint, it's better to check before stage repro, right? Is there any downside to that approach for the user?

As far as whether to error or give a warning, considering this feature is primarily for CI, I don't think warnings will be that helpful, so I'd opt for an error. What do you think?

dvc/repo/experiments/executor/base.py
pmrowla: Is "push every N commits" actually a requirement for CML? This seems like an additional feature request that's unrelated to being able to recover checkpoints from a dead CI runner.

And it seems like there's still some behavior discussion that would need to be worked out there, i.e. what happens if I have it set to 10 and my experiment generates less than 10 commits, should DVC not push anything?

It also seems like something that should maybe be tied to dvclive and not DVC itself, since DVC doesn't even have a real concept of epoch count/iteration awareness (related dvclive ticket: https://github.com/iterative/dvclive/issues/113). Since dvclive *is* aware of actual epoch counts, it can just call `experiments.push` at the appropriate time without the need for adding more env vars

karajan1001: @pmrowla 
> And it seems like there's still some behavior discussion that would need to be worked out there, i.e. what happens if I have it set to 10 and my experiment generates less than 10 commits, should DVC not push anything?

The final push after commit is a solution, I think.

pmrowla: With regard to this PR, the `DVC_EXP_AUTO_PUSH` env var is intended to be used when there is a requirement that all commits be pushed as they are made (so for CI and other remote execution environments where the user is concerned about potential data loss)

I think granular control over DVC behavior relative to metric states/epoch counts/etc belongs in dvclive. Similar to the discussion over telling dvclive to stop iterating once some metric condition or epoch count is reached, it would be useful for users to be able to push on those kind of conditions as well via dvclive (something like `dvclive.next_step(push=...)` was even suggested in the original issue for this PR).

But this all seems unrelated to the immediate CML need for automated pushing on every commit to account for potentially dead runners (and the DVC fix for this CML requirement should not be dependent on dvclive), which is addressed by what's in this PR now

karajan1001: > It also seems like something that should maybe be tied to dvclive and not DVC itself, since DVC doesn't even have a real concept of epoch count/iteration awareness (related dvclive ticket: [iterative/dvclive#113](https://github.com/iterative/dvclive/issues/113)). Since dvclive _is_ aware of actual epoch counts, it can just call `experiments.push` at the appropriate time without the need for adding more env vars

Agreed

karajan1001: > I think granular control over DVC behavior relative to metric states/epoch counts/etc belongs in dvclive. Similar to the discussion over telling dvclive to stop iterating once some metric condition or epoch count is reached, it would be useful for users to be able to push on those kind of conditions as well via dvclive (something like `dvclive.next_step(push=...)` was even suggested in the original issue for this PR).

How about let the dvclive control `DVC_EXP_AUTO_PUSH` env dynamically?

pmrowla: I don't really see why dvclive would need to manipulate the env var when it can just call the push function directly itself and keep control over the specific push options.

(In CML to fix the dead runner issue we need to push all 3 of commits/dvc cache/run cache, but in normal user scenarios they probably do not need to push all 3 every time)

dvc/repo/experiments/executor/base.py
```diff
@@ -450,9 +455,27 @@ def _repro_args(cls, dvc):
             kwargs = {}
         return args, kwargs
 
+    @staticmethod
+    def _auto_push(git_remote: str, dvc: "Repo", scm: "Git"):
+        if git_remote == scm.root_dir:
+            logger.warning(
+                f"try to auto checkpoints to {git_remote} which is the "
+                "running repository dvc cache will be pushed to the "
+                "default remote while git references will not be pushed"
```
jorgeorpinel, line 464: I'm not sure I understand this message. Maybe something like

> Try to save the experiment status to {git_remote} after every checkpoint.
Data will be pushed to the default remote.

Not sure what Git refecences we're talking about. Meaning the experiments themselves? If so what gets saved to {git_remote} ?

pmrowla: We are `git push`ing the currently running experiment ref. In this condition, `git_remote` points to the current git repo itself, so nothing actually happens on the git side. However, when this env var is set, we also automate `dvc push`ing used cache objects and run cache for the currently running experiment as well.

So the side effect here is that nothing will be "pushed" anywhere on the git side of things, but DVC outputs and run cache will still be pushed to the default configured DVC remote.

pmrowla: The warning message should probably just be something like

> '{git_remote}' points to the current Git repo, experiment Git refs will not be pushed. DVC cache and run cache will automatically be pushed to the default DVC remote (if any) on each experiment commit.

dvc/repo/experiments/executor/base.py
casperdcl: > Is "push every N commits" actually a requirement for CML?

A stable API is a requirement. My point is if we (in this PR or in a future PR) implement "push every N commits" and other config options we'd probably change the API of this PR.

In summary, as a minimum requirement to approve this PR, I would rename `DVC_EXP_AUTO_PUSH` to `DVC_EXP_AUTOPUSH_REMOTE` or something similar.

dberenbaum: So your only concern for now is the name of this env var @casperdcl?

Seems like `DVC_EXP_AUTO_PUSH` does two things:
1. Flags whether to auto push.
2. Specifies the remote.

The name makes sense for the first purpose but is probably too generic for the second. We could:
1. Keep it as is.
2. Change the name to mention `REMOTE` or even `GIT_REMOTE`/`GITREMOTE`/`GIT`.
3. Split into two env vars.
4. Leave the remote to be configured in Git (see https://git-scm.com/docs/git-config#Documentation/git-config.txt-remotepushDefault) and make `DVC_EXP_AUTO_PUSH` a boolean.

Any preferences?

pmrowla: I don't think relying on the default git remote is a solution since it is probably not set in CI environments. I'd be in favor of just changing it to something with `GIT_REMOTE` or `GIT_URL`.

karajan1001: `GIT_REMOTE` didn't indicate anything about `DVC` and `AUTOPUSH`, maybe `DVC_CHECKPOINT_REMOTE` is better?

pmrowla: We meant that the name should include something like `GIT_REMOTE`/`GIT_URL`, not that the env var should just be `GIT_REMOTE`. So something like `DVC_EXP_AUTO_PUSH_GIT_REMOTE`, but preferably shorter.

I don't think using just `REMOTE` alone will work here since it's not obvious if it's a git remote or a dvc remote.

This is also not specific to checkpoints, regular experiments will end be pushed the same way if the env var is set.

casperdcl: > So your only concern for now is the name of this env var?

only blocking concern, yes (because *APIs are forever*).

I'd suggest:

- `DVC_EXP_CHECKPOINT_PUSH:bool`
- rely on Git remote (if CI envs don't set it, we can document that the user should `git remote add` etc. first)

introducing `DVC_EXP_GIT_REMOTE` to me seems necessary [iff](https://en.wikipedia.org/wiki/If_and_only_if) there's a reason for the user to deliberately configure multiple different Git remotes.

pmrowla: Configuring multiple git remotes is a pretty typical use case (original/`upstream` repo vs my fork/`origin`). If I'm working on a fork and want my run experiments pushed to the original repo but not my fork, I need to be able to configure the remote.

---

I'm also concerned that relying on the default git remote behavior is going to end up confusing for users in certain situations.

Default git remotes only apply to git branches. When I have a checked out git branch, and do
```
git push
```
with no remote specified, git will look at the repo config and checks for `remote.pushDefault` (never set by default in any git repo), and then `branch.<current_branch_name>.remote` (set when you configure a specific remote-tracking branch with `git push -u ...`) in that order. If neither of those are set, it will fallback to `origin`.

The issue here is that upstream tracking branches only apply to branches. DVC exps do not have upstream tracking branches. This means that "default" for DVC experiments is always going to be either `remote.pushDefault` or `origin`.

This is going to be confusing for users in the event that they are running experiments based on a branch with a configured non-`origin` upstream tracking remote. The end result is that users get this behavior:

```
git checkout mybranch  # checks out mybranch
git push -u upstream mybranch  # sets default push remote for mybranch to 'upstream'
DVC_EXP_AUTO_PUSH=1 dvc exp run  # runs an experiment and auto-pushes it to 'origin'
```

In this case it seems like what the user probably wanted (and probably would expect) is that the experiment would be pushed to `upstream`. But the exp ref is actually pushed to `origin`, since `branch.mybranch.remote = upstream` specifically only applies when git is pushing `refs/heads/mybranch` and does not apply to anything in any custom ref namespace.

This ambiguity is why `dvc exp push` always requires an explicit remote name.

karajan1001: I think that the behavior of `auto push` should go with `dvc exp push`.  Make different protocols for them will confuse users. If `dvc exp push` can run with a default remote, `auto push` should also support it, vice versa.

karajan1001: One possible solution is that 
1. `DVC_EXP_GIT_REMOTE` is used for the default exp push git remote. Both `dvc exp push` and `auto push` can use it (But the auto push cann't work without it, while `dvc exp push` can run git remote arguments).
2. `DVC_EXP_CHECKPOINT_PUSH`/ `DVC_EXP_CHECKPOINT_AUTO_PUSH` can be used for the open/close of auto_push

For the problem of `push every N epochs ` better to discuss with `dvclive` side?

karajan1001: > * `DVC_EXP_GIT_REMOTE` is used for the default exp push git remote. Both `dvc exp push` and `auto push` can use it (But the auto push cann't work without it, while `dvc exp push` can run git remote arguments).

But it is a totally new feature, involving changes on `dvc exp push` should not be mixed in this PR.

casperdcl: > Configuring multiple git remotes is a pretty typical use case

ok in that case I'd go with

> 2. Split into two env vars

i.e. `DVC_EXP_CHECKPOINT_PUSH:bool` and `DVC_EXP_GIT_REMOTE:str`?

Regarding:

> I think that the behavior of `auto push` should go with `dvc exp push`

This also makes sense to me

dberenbaum: > This is also not specific to checkpoints, regular experiments will end be pushed the same way if the env var is set.

I think we need to keep `CHECKPOINT` out of the names of these env vars for this reason.

---

Splitting into two env vars makes sense, but it does make the UX more complex. I think it's worth making sure this is really necessary before creating extra complexity for hypothetical future needs, so I have a couple follow-ups:

Why not rename the env var to `DVC_EXP_GIT_REMOTE` and make no other changes? That provides future flexibility but also keeps the UX concise.

Alternatively, I think we could reconsider having a default remote.

> Configuring multiple git remotes is a pretty typical use case (original/`upstream` repo vs my fork/`origin`). If I'm working on a fork and want my run experiments pushed to the original repo but not my fork, I need to be able to configure the remote.

1. Is this likely in CI?
2. Even locally, are users likely to do `git push -u upstream mybranch`? My typical workflow would be to pull from upstream and push to origin by default.
3. If we aren't following Git default remote configuration, why not document that `origin` is the default instead of requiring it to be explicit?

pmrowla: > Is this likely in CI?

in CI probably not

> Even locally, are users likely to do git push -u upstream mybranch? My typical workflow would be to pull from upstream and push to origin by default.

I do this regularly with DVC - we have CI configured to run a different set of actions on pull requests vs branches pushed directly to `iterative/dvc`, so there are times when I want to push a certain branch for specific features (that will require the stricter branch tests) to upstream rather than my personal fork

> If we aren't following Git default remote configuration, why not document that origin is the default instead of requiring it to be explicit?

This is an option, although as someone who is going to have to answer support questions regarding "why is the default for this experiment `origin` instead of the default I set on my git branch", my preference would be to continue requiring the explicit argument

karajan1001: @dberenbaum 
> Alternatively, I think we could reconsider having a default remote.

I think having a default remote for the `dvc exp push` is reasonable, both `dvc push` and `git push` have default remotes.
How about adding it to the `dvc.config`, it is more manageable. 

And if a default remote for `dvc exp push` is setted, the only env arguments here to control the auto push process is `DVC_EXP_CHECKPOINT_PUSH:bool`.

I don't think users will directly use auto push without using `dvc exp push` previously. If they are familiar with `dvc exp push`, and knows how to set up a default remote. The auto push configuration is quite a simple one.

karajan1001:
In summary, 

1. split the current env into two (confirm the usage with CML team).
2. create a new ticket for the default dvc exp push git remote.
3. auto push on every several epochs -- a new feature not to discuss here (also a new ticket ) .

Anything else.

casperdcl:
@iterative/cml

dvc/env.py
```diff
@@ -7,3 +7,5 @@
 DVCLIVE_HTML = "DVCLIVE_HTML"
 DVCLIVE_RESUME = "DVCLIVE_RESUME"
 DVC_IGNORE_ISATTY = "DVC_IGNORE_ISATTY"
+DVC_EXP_GIT_REMOTE = "DVC_EXP_GIT_REMOTE"
+DVC_EXP_CHECKPOINT_PUSH = "DVC_EXP_CHECKPOINT_PUSH"
```
pmrowla, line 11: I think this should be named something more like `DVC_EXP_AUTO_PUSH`

It is not specific to checkpoints, and should not have `CHECKPOINTS` in the name.

dvc/utils/__init__.py
```diff
@@ -512,3 +513,11 @@ def errored_revisions(rev_data: Dict) -> List:
...
             result.append(revision)
     return result
+
+
+def getenv_bool(name: str, default: str = "False"):
+    raw = os.getenv(name, default).title()
+    try:
+        return bool(ast.literal_eval(raw))
+    except ValueError:
+        return False
```
pmrowla, line 523: I don't think we should get into this kind behavior, in my opinion it would be better to just document that if the environment variable is set at all, it will be considered enabled.

but others may have different thoughts on this
cc: @iterative/dvc

pmrowla (changes requested):

dvc/env.py
karajan1001: Yes, because we had added auto push after the running result being committed. Other experiments have also been affected.

dvc/utils/__init__.py
casperdcl: I could see a scenario where a templated YAML workflow yields `env: VAR: false` but not `env: VAR: False`. Agree that it seems to much of a headache to support.

`if os.getenv(name, '')` seems fine to me...

dvc/env.py
```diff
@@ -7,3 +7,5 @@
 DVCLIVE_HTML = "DVCLIVE_HTML"
 DVCLIVE_RESUME = "DVCLIVE_RESUME"
 DVC_IGNORE_ISATTY = "DVC_IGNORE_ISATTY"
+DVC_EXP_GIT_REMOTE = "DVC_EXP_GIT_REMOTE"
+DVC_EXP_AUTO_PUSH = "DVC_EXP_CHECKPOINT_PUSH"
```
casperdcl, line 11: why the difference between var name & string? (Also what are the strings for?)

dvc/utils/__init__.py
efiop: We also have `env2bool` in dvc/utils, that we can use here instead.

dvc/scm/git/backend/dulwich.py
```diff
@@ -353,20 +353,20 @@ def iter_remote_refs(self, url: str, base: Optional[str] = None):
...
             client, path = get_transport_and_path(location)
+            if base:
+                yield from (
+                    os.fsdecode(ref)
+                    for ref in client.get_refs(path)
+                    if ref.startswith(os.fsencode(base))
+                )
+            else:
+                yield from (os.fsdecode(ref) for ref in client.get_refs(path))
+
```
pmrowla, line 364: Why is this code block moved?

`client.get_refs()` failing is not always going to be due to an invalid URL. `get_transport_and_path` does validation on whether or not the specified git remote exists, or whether or not an explicit URL is a valid git URL.

I get that you are trying to validate whether or not the URL is reachable before the experiment is run, but I don't think that is necessary at all. And even if you want to validate the URL by doing an `scm.iter_refs` call, you still don't need to move this code block to do it.

(and if `client.get_refs()` fails, it would be better to pass the actual exception for why it failed up the chain, instead of just assuming it was due to an invalid URL.)

pmrowla (changes requested):

dvc/repo/experiments/executor/base.py
```diff
@@ -248,6 +250,28 @@ def on_diverged_ref(orig_ref: str, new_rev: str):
...
+            )
+        try:
+            for ref in dvc.scm.iter_remote_refs(
+                git_remote, base=EXPS_NAMESPACE
+            ):
+                if ref:
+                    break
+        except BaseException as e:
+            print(e)
+            raise e
```
pmrowla, line 271: Why is this print statement needed?

In the event that we need to explicitly catch and re-log exceptions in a multiprocessing context, they are already caught and logged in the `_repro_dvc` context manager

dvc/repo/experiments/executor/base.py
```diff
@@ -248,6 +250,28 @@ def on_diverged_ref(orig_ref: str, new_rev: str):
             )
         return refs
 
+    @classmethod
+    def _auto_push_check(cls, dvc: "Repo"):
```
pmrowla, line 254: This is minor, but I would maybe call this function `_validate_remotes` or something along those lines. The way the PR looks right now it comes across as if this is going to be a modified version or wrapper around `_auto_push()`, when it's actually a function to validate whether or not we have git and dvc remotes to push to

dvc/scm/git/backend/dulwich.py
karajan1001: > I get that you are trying to validate whether or not the URL is reachable before the experiment is run, but I don't think that is necessary at all. 

Because the first push will happen after one iteration, stopping at that point will leave a dirty workspace.

pmrowla: Right, but that's not any different than any other failure that may occur during an experiment run, or during `dvc repro`.

The user's pipeline code itself could fail, or the `dvc push` could fail due to any number of reasons, whether it's because of a network error, or because the remote config is actually invalid. (We don't validate being able to connect to a DVC remote, all we do is validate that a default remote exists in the DVC config).

All of these cases will result in a "dirty" workspace, which is fine.

karajan1001: In some cases, training 1 epoch would cost hours or even days of time. And it will be really frustrating if got an error from DVC.

pmrowla: My point is that we are not checking for other things that can fail in DVC, after the epoch has been generated.

Just as an example, we do not verify whether or not it is possible to actually push the cache data to the default DVC remote. If I have mis-configured my DVC remote using either a URL that doesn't exist, or with the wrong authentication credentials, we will fail after we have generated a commit and try to auto-push it.

If we are going to do this kind of strict validation on the git remote, then we should also be doing it on the DVC remote. But in my opinion, both of these checks would be excessive.

---

I also still do not think we should be failing at all in the event that the automatic push fails. We should wrap any exceptions that occur when we try to push, and log them if it fails, but continue running the pipeline. If the user has misconfigured the git URL, they can just do a manual `dvc exp push <correct_git_url> ...` after execution has completed, rather than needing to re-run an entire experiment.

But this is maybe a question for @dberenbaum 

(And yes, this manual push isn't possible in the CML use case, but in the CML use case the key differences are that the user has probably already tested their code locally and that the upstream git URL or remote should be coming from the CI provider itself (i.e. from the github/gitlab environment))

dberenbaum: It looks the changes are outdated, so it seems like some of the specifics have changed, and I'll focus on the general user experience.

I think there was a previous question from @karajan1001 about whether to validate the url before running the experiment. This validation seems useful and I see little downside from a user perspective. Since dvc knows the experiment will be pushed, it would be helpful to avoid issues before spending time to run the experiment. I also think it makes sense to throw an error if validation fails since auto pushing is specifically for cml and similar workflows, where leaving a dirty workspace isn't helpful.

It would actually seem useful to have some basic validation of both the git and dvc remotes. As @pmrowla mentioned, in cml the `DVC_EXP_GIT_REMOTE` is unlikely to be a problem since it's being set automatically. However, the dvc remote is configured by the user, and it will be problematic if the auto push fails in cml.

Feel free to push back with reasons to avoid this validation or with questions of how strictly to validate.

pmrowla: > I think there was a previous question from @karajan1001 about whether to validate the url before running the experiment. This validation seems useful and I see little downside from a user perspective.

There are different possible levels of "validation". IMO validating that a `.git/config` remote named `origin` exists or that the string `git@github.com/my/repo.git` is a valid git URL but `s3://some_non_git_url` is not valid is fine. But I don't think we have to bother checking that `git@github.com/my/repo.git` is reachable before trying to push to it.

This is the exact same level of @karajan1001 's current validation for the DVC remote. We check that a default DVC remote exists in the repo config, but we do not do anything to verify that the remote URL is reachable, or that we have the required list/write/move/delete permissions to actually push to it.

IMO this is fine. But trying to connect to each remote before we actually run anything is overkill, and in the event that it fails we can handle it at that time.

> Since dvc knows the experiment will be pushed, it would be helpful to avoid issues before spending time to run the experiment. I also think it makes sense to throw an error if validation fails since auto pushing is specifically for cml and similar workflows, where leaving a dirty workspace isn't helpful.

I don't think that an automatic push errors in general should be considered a failure state for the entire workflow.  A push failure is an error that the user can recover from by just re-trying `dvc exp push` again after their experiment has completed.

So in the event that github or aws goes down, or that a gdrive remote gets API throttled, or any other kind of possible network error occurs (including a remote being unreachable due to a user misconfiguration), we should just log that "automatic push failed with error: ..." and continue running the pipeline.

> However, the dvc remote is configured by the user, and it will be problematic if the auto push fails in cml.

In this case, users should be testing their remote configurations themselves locally before they git commit things into their `.dvc/config`.

And in the event that they misconfigure it, we should just warn them, and continue running the experiment (for the same reasons I already mentioned)

pmrowla: At the end of the day I guess this is a product decision, and if we want to get into creating a test file and `dvc push`ing and then removing it, obviously we can do that. But I think it's overkill and an unnecessary extra step.

dvc/env.py
```diff
@@ -7,3 +7,5 @@
 DVCLIVE_HTML = "DVCLIVE_HTML"
 DVCLIVE_RESUME = "DVCLIVE_RESUME"
 DVC_IGNORE_ISATTY = "DVC_IGNORE_ISATTY"
+DVC_EXP_GIT_REMOTE = "DVC_EXP_GIT_REMOTE"
+DVC_EXP_AUTO_PUSH = "DVC_EXP_AUTO_PUSH "
```
pmrowla, line 11:
```suggestion
DVC_EXP_AUTO_PUSH = "DVC_EXP_AUTO_PUSH"
```

dvc/repo/experiments/executor/base.py
```diff
@@ -248,6 +250,27 @@ def on_diverged_ref(orig_ref: str, new_rev: str):
...
+                "(if any) on each experiment commit."
+            )
+        try:
+            for ref in dvc.scm.iter_remote_refs(
+                git_remote, base=EXPS_NAMESPACE
+            ):
+                if ref:
+                    break
+        except BaseException as e:
+            raise e
```
pmrowla, line 270: This try/except is unnecessary (nothing is actually done in the except block other than re-raising the exception)

dvc/repo/experiments/executor/base.py
```diff
@@ -450,9 +478,21 @@ def _repro_args(cls, dvc):
             kwargs = {}
         return args, kwargs
 
+    @staticmethod
+    def _auto_push(dvc: "Repo", scm: "Git"):
+        git_remote = os.getenv(DVC_EXP_GIT_REMOTE)
```
pmrowla, line 483: It's not really a requirement for this PR, but one thing to consider would be designing this to be more like

```py
def _exp_push(dvc, scm, git_remote, push_cache=True, run_cache=True):
    ...
    dvc.experiments.push(
        git_remote, branch, push_cache=push_cache, run_cache=run_cache
    )
```

It seems like eventually we are going to end up needing to provide some API that supports finer control over this pushing behavior, and triggering it won't always be based on env vars. So the automated push functionality should be designed with that in mind.

Also with the current implementation we end up re-reading environment vars over and over again. It would be cleaner to just read them all once at the start of `executor.reproduce()` and then pass the known values into functions where they are needed, with something like:

```py
    def reproduce(...):
        auto_push = env2bool(DVC_EXP_AUTO_PUSH)
        git_remote = os.getenv(DVC_EXP_GIT_REMOTE)
        ...
        if auto_push:
            _exp_push(..., git_remote, ...)
```

dvc/scm/git/backend/dulwich.py
dberenbaum: > There are different possible levels of "validation". IMO validating that a `.git/config` remote named `origin` exists or that the string `git@github.com/my/repo.git` is a valid git URL but `s3://some_non_git_url` is not valid is fine. But I don't think we have to bother opening a test connection to `git@github.com/my/repo.git` before trying to push to it.

How can we check that it's a valid git url? I agree that limiting the validation makes sense since it might not make sense to fail for random timeouts that have nothing to do with the remote configuration. Ideally, it seems like a retry might be helpful, but that's out of scope.

> I don't think that an automatic push errors in general should be considered a failure state for the entire workflow. A push failure is an error that the user can recover from by just re-trying `dvc exp push` again after their experiment has completed.

I think you mentioned before how this isn't helpful in the cml context since you won't be able to access the runner and debug. Are you thinking that, if there's something like a temporary network issue, then the next checkpoint might succeed, so the workflow should continue? Or are you thinking of other scenarios like remote executors where it will be possible to debug?

If there is a temporary network error for a single checkpoint push in the middle of the workflow, failing could be bad for the user since they might not be watching a long-running job and might return later to unexpectedly find that the job did not run to completion. That's why I agree that limiting validation makes sense, and I also agree that auto push errors should warn rather than interrupt the entire job.

However, in the initial validation, failing fast seems like a good user experience since they can debug quickly and resubmit their job. For the initial validation, I think it makes sense to do what we can to check if everything is configured correctly and working as expected and fail if not.

Since this feature is for @iterative/cml, it would be good to get their feedback on this.

pmrowla: > How can we check that it's a valid git url?

All of the git backends we use have calls that can do this check (to verify that either a string remote name like `origin` exists, or that a URL conforms to what git expects a git URL to look like)

Basically we would just have to make this a standalone scm call:

https://github.com/iterative/dvc/blob/fd76e0991f6dd47e554b874c2efb8c280dee96c0/dvc/scm/git/backend/dulwich.py#L353-L359

> I think you mentioned before how this isn't helpful in the cml context since you won't be able to access the runner and debug. Are you thinking that, if there's something like a temporary network issue, then the next checkpoint might succeed, so the workflow should continue? Or are you thinking of other scenarios like remote executors where it will be possible to debug?

Depending on the issue, yes it's entirely possible that after one exp commit fails to be pushed in CML, the subsequent push could succeed. But really I was just thinking of scenarios where users set up automatic pushing in general, and not necessarily in CI (this becomes relevant for local exp runs in the event that support for this gets added into dvclive)

dvc/repo/experiments/executor/base.py
```diff
@@ -248,6 +250,30 @@ def on_diverged_ref(orig_ref: str, new_rev: str):
...
+
+        if git_remote == dvc.root_dir:
+            logger.warning(
+                f"'{git_remote}' points to the current Git repo, experiment "
+                "Git refs will not be pushed. But DVC cache and run cache "
+                "will automatically be pushed to the default DVC remote "
+                "(if any) on each experiment commit."
+            )
+        try:
+            _, location = get_remote_repo(dvc.scm.dulwich.repo, git_remote)
```
karajan1001, line 268: One problem here. This method not only uses low-level Dulwich API but also can not raise proper Exception with a Git remote value `true`

pmrowla: Right, this should not go here. It needs to be added as a new `dvc.scm.git.backend` call (with the non-dulwich implementations set to raise `NotImplementedError`)

karajan1001: > Right, this should not go here. It needs to be added as a new `dvc.scm.git.backend` call (with the non-dulwich implementations set to raise `NotImplementedError`)

Another problem is that these lines will not raise any exception even if I set the git_remote to a none URL format (guess, this would be considered to some references of a URL). The exception only raises when it began to read out the data from the remote repo.

pmrowla: It's because dulwich just falls back to assuming `location` is supposed to be a local path (and assumes there will be a valid git repo at `file:///location`.
https://github.com/dulwich/dulwich/blob/952573478aa2b0833b8a51b7e741e9e13fc14060/dulwich/client.py#L2210

If dulwich returns a local client you can just check that the local path actually exists with:

```py
from dulwich.client import LocalGitClient
_, location = get_remote_repo(dvc.scm.dulwich.repo, git_remote)
client, path = get_transport_and_path(location)
if isinstance(client, LocalGitClient) and not os.path.exists(os.path.join("", path)):
    # this is an invalid local client/path
    raise ...
```

dvc/scm/git/backend/dulwich.py
```diff
@@ -646,3 +646,17 @@ def merge(
...
+        try:
+            _, location = get_remote_repo(self.repo, url)
+            client, path = get_transport_and_path(location)
+        except Exception as exc:
+            raise SCMError(
+                f"'{url}' is not a valid Git remote or URL"
+            ) from exc
+        if isinstance(client, LocalGitClient) and not os.path.exists(path):
+            raise SCMError(f"'{url}' is not a valid Git remote or URL")
```
pmrowla, line 662: We can make a dedicated `InvalidRemoteRepo` exception for this in `dvc.scm.base` since it seems like we may end up using it in other places as well in the future

dvc/repo/experiments/executor/base.py
```diff
@@ -450,9 +473,32 @@ def _repro_args(cls, dvc):
...
+                git_remote,
+                branch,
+                push_cache=push_cache,
+                run_cache=run_cache,
+            )
+        except BaseException:
+            logger.warning(
+                "Something went wrong while auto pushing checkpoint cache "
+                f"to the remote {git_remote}"
+            )
```
pmrowla, line 496:
```suggestion
        except BaseException as exc:
            logger.warning(
                "Something went wrong while auto pushing experiment "
                f"to the remote '{git_remote}': {exc}"
            )
```

tests/func/experiments/test_checkpoints.py
```diff
@@ -188,3 +194,103 @@ def test_resume_non_head_checkpoint(
...
+
+@pytest.fixture
+def clear_env(monkeypatch):
+    yield
+    monkeypatch.delenv(DVC_EXP_GIT_REMOTE, raising=False)
+    monkeypatch.delenv(DVC_EXP_AUTO_PUSH, raising=False)
+
+
+@pytest.mark.parametrize("use_url", [True, False])
+def test_auto_push_during_iterations(
```
pmrowla, line 207: This test can probably be simplified. For this test case, what's important is whether or not we automatically push when the env vars are set. We don't actually need to test what happens in `dvc.experiments.push` (since we already have other tests for that behavior).

I think all we really need to do here is `mocker.spy` `experiments.push` to confirm that it is called the expected number of times, with the expected arguments.

So if the env var is enabled, push should be called once per commit - including the final non-checkpoint commit) with the arguments we expect (the correct git URL, and with DVC cache + run cache set to true).

If the env var is disabled, it should not be called at all.

pmrowla (changes requested):

dvc/repo/experiments/executor/base.py
pmrowla: We need to log the exception itself so that the user can see why the push failed

tests/func/experiments/test_checkpoints.py
karajan1001: Good suggestion, before this I used to plan to add a new test case which will exit in running.

dvc/scm/git/backend/dulwich.py
```diff
@@ -642,3 +644,15 @@ def merge(
...
+    def validate_git_repo(self, url: str):
+        from dulwich.client import LocalGitClient, get_transport_and_path
+        from dulwich.porcelain import get_remote_repo
+
+        try:
+            _, location = get_remote_repo(self.repo, url)
+            client, path = get_transport_and_path(location)
+        except Exception as exc:
+            raise InvalidRemoteSCMRepo(url) from exc
+        if isinstance(client, LocalGitClient) and not os.path.exists(path):
```
pmrowla, line 657:
```suggestion
        if isinstance(client, LocalGitClient) and not os.path.exists(os.path.join("", path)):
```

We need the `os.path.join("", ...)` here since dulwich assumes `path` is really `file:///path` (and always adds the leading `/` to make it an abspath)

pmrowla (changes requested):
Looks good for the most part, I think this should be the last set of changes to do

tests/func/experiments/test_checkpoints.py
```diff
@@ -188,3 +195,90 @@ def test_resume_non_head_checkpoint(
...
+    auto_push_spy = mocker.spy(BaseExecutor, "_auto_push")
+
+    # without auto push
+    results = dvc.experiments.run(checkpoint_stage.addressing)
+    assert auto_push_spy.call_count == 0
+
+    # add auto push
+    monkeypatch.setenv(DVC_EXP_AUTO_PUSH, "true")
+    results = dvc.experiments.run(checkpoint_stage.addressing)
+    assert (tmp_dir / "foo").read_text() == "4"
```
casperdcl, line 233: not complaining, but this reminds me of
xkcd#221

skshetry: Your comment reminds me of this old blogpost: https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html.

dvc/repo/experiments/executor/base.py
```diff
@@ -248,6 +250,19 @@ def on_diverged_ref(orig_ref: str, new_rev: str):
             )
         return refs
 
+    @classmethod
+    def _validate_remotes(cls, dvc: "Repo", git_remote: Optional[str]):
+
+        if git_remote == dvc.root_dir:
```
dberenbaum, line 256: Is there a reason to have this explicit check? Would this qualify as a valid git remote? Is there some reason to think a user might set `DVC_EXP_GIT_REMOTE` to their local project's root dir?

dvc/repo/experiments/executor/base.py
```diff
@@ -248,6 +250,19 @@ def on_diverged_ref(orig_ref: str, new_rev: str):
...
+            logger.warning(
+                f"'{git_remote}' points to the current Git repo, experiment "
```
dberenbaum, line 258:
```
                f"'{git_remote}' points to the current Git repo, experiment "
```

I'm not sure this is helpful since a valid remote would also point to the current Git repo, right? Maybe something like "local workspace" makes more sense than "current Git repo"?

dvc/repo/experiments/executor/base.py
dberenbaum: Also, why fail for an invalid Git remote but only warn for this scenario?

pmrowla: `git_remote` is just a variable that is either a path/URL to a git repo (including local ones) or a remote name. If this check is hit it means it's a local path that points directly to the current DVC root directory, which in most cases is a valid git repo path.

If you do `DVC_EXP_GIT_REMOTE=.` the end result would be that we auto-push DVC cache for your local experiment runs. (The git push step becomes a no-op, the same thing that happens if you do a CLI `git push` to your own current repo directory)

tests/func/experiments/test_checkpoints.py
pmrowla: Like Casper said it's not a big deal, but we could actually get rid of these magic numbers since it comes from a defined tests constant:
https://github.com/iterative/dvc/blob/fb96b13f13948d775a3061c03868db3775f6c665/tests/func/experiments/conftest.py#L7

the `4` is the number of `exp run` calls multiplied by the number of iterations per `exp run` so it would be a bit clearer to write it as `2 * DEFAULT_ITERATIONS`

dvc/repo/experiments/executor/base.py
pmrowla: The local workspace is your current Git repo. `https://github.com/my/repo.git` is not your current Git repo - it's *a* remote Git repo, and potentially where your current Git repo is cloned from, but "current repo" would always be your current local workspace repo

pmrowla (approved):

casperdcl:
thanks @karajan1001 - is there a follow-up PR in https://github.com/iterative/dvc.org/pulls?

karajan1001:
@casperdcl related https://github.com/iterative/dvc/pull/6332

PR #6338, opened 2021-07-20
objects: fix gdrive CI issue

pmrowla:

### DIFF
--- a/dvc/data_cloud.py
+++ b/dvc/data_cloud.py
@@ -118,6 +118,7 @@ class DataCloud:
             jobs=jobs,
             src_index=get_index(odb),
             cache_odb=self.repo.odb.local,
+            verify=odb.verify,
         )
 
     def status(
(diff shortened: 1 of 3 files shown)


dvc/fs/base.py
```diff
@@ -182,7 +182,10 @@ def info(self, path_info):
         raise NotImplementedError
 
     def getsize(self, path_info):
-        return self.info(path_info).get("size")
+        size = self.info(path_info).get("size")
+        if size is not None:
+            size = int(size)
+        return size
```
efiop, line 188: Was gdrive returning string size? `fs.info` should always return size as int, so prob worth fixing in `gdrive.py` itself.

pmrowla: Yeah, there were 2 issues

- after the transfer change we were verifying the destination ODB on both push and pull, when previously we only verified on pull
- the verify on push caused us to read sizes from the gdrive fs, which made the issue w/gdrive returning string sizes show up

I'll move the conversion into our gdrive fs, still waiting for the branch build to finish to make sure there's nothing else

efiop: @pmrowla Nice catch! :fire:

efiop:
Thank you, @pmrowla ! :pray:

PR #6342, opened 2021-07-21
[testing] tests: ensure gdrive fixture URL is unique

pmrowla:

### DIFF
--- a/tests/func/test_data_cloud.py
+++ b/tests/func/test_data_cloud.py
@@ -68,6 +68,8 @@ def test_cloud(tmp_dir, dvc, remote):  # pylint:disable=unused-argument
             expected = kwargs.get(key, set())
             assert expected == {obj.hash_info for obj in getattr(status, key)}
 
+    odb = dvc.cloud.get_remote_odb("upstream")
+    assert not odb.exists(foo_hash)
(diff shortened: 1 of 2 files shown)


tests/func/test_data_cloud.py
```diff
@@ -68,6 +68,8 @@ def _check_status(status, **kwargs):
             expected = kwargs.get(key, set())
             assert expected == {obj.hash_info for obj in getattr(status, key)}
 
+    odb = dvc.cloud.get_remote_odb("upstream")
+    assert not odb.exists(foo_hash)
```
efiop, line 72: Hm, very interesting that this gets triggered, even though we are using unique urls :thinking:

pmrowla: Yeah I was initially thinking it that we were maybe hitting UUID collisions w/the URL generation, but it seems like it's probably something gdrive internals specific. I'll keep digging into it

efiop: @pmrowla Maybe let's just temporarily xfail or skip this test for gdrive and create a ticket. We'll be migrating gdrive to fsspec soon, and might take a look then.

PR #6357, opened 2021-07-22
setup: bump flatten-dict to 0.4.1

### LINKED ISSUE #6349: s3: config load takes an excessive amount of time when initializing an s3 fs
Noticed this when benchmarking the recent ODB changes. For initializing a single s3 fs instance we spend almost an entire second (on my machine) just flattening and unflattening config dictionaries, which seems like a lot of overhead for just merging a couple of small dict configs

https://github.com/iterative/dvc/blob/771b89ddf2aaac4bf05afa907eb34f7ada013b1a/dvc/fs/s3.py#L130-L136

(should note that this is in 3rd party `flatten_dict` library)

Screen Shot 2021-07-22 at 2 52 20 PM

For perspective, both before and after the odb transfer PR, this is from a `dvc status -c` that takes a total of ~3 seconds to complete. So this config handling is 1/3 of the total execution time, and actual s3/network operations are most of the remaining 2/3. (I would have expected the network I/O to be almost the entire runtime)

isidentical:
My [latest PR](https://github.com/ianlini/flatten-dict/pull/44) to flatten-dict should shave about 0.5 seconds on very complex environments for us, so upgrading it to 0.4.1. Resolves #6349

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -79,7 +79,7 @@ install_requires = [
     "speedcopy>=2.0.1; python_version < '3.8' and sys_platform == 'win32'",
     "dataclasses==0.7; python_version < '3.7'",
     "importlib-metadata>=1.4; python_version < '3.8'",
-    "flatten_dict>=0.3.0,<1",
+    "flatten_dict>=0.4.1,<1",
     "tabulate>=0.8.7",
     "pygtrie>=2.3.2",
     "dpath>=2.0.1,<3",


efiop:
> Error: Codecov: Uploader shasum does not match uploader hash: 011fd19e47133975f69a520d93dac4cc0f8b245c1d5b144aeac4b00ab6d9b716  codecov, public hash: 4368ca2271b06047b30fd0aff9d08336fa8fc7bbec897845fc702bf70bb9eaed  codecov

image

efiop (approved):
Nice catch @isidentical ! :fire: 


For the record: codecov issue is unrelated.

PR #6360, opened 2021-07-23
objects: use object IDs and references instead of naive objs in status/transfer

pmrowla:

- Adds `ReferenceHashFile` object type and `ReferenceObjectDB` ODB type
    - Ref ODB stores references to files outside the ODB fs, so the object at hash `abc123` is a serialized reference to some filesystem and path where the actual source file w/hash `abc123` exists
- `objects.stage` now stages objects as references in a memfs based ref ODB
- `objects.save` is removed in favor of stage + transfer from staging to ODB pattern
- `objects.status` and `objects.transfer` now accept a collection of object IDs (hashinfos) instead of naive objects
    - repo/stage/output object collection now collects object IDs instead of naive objects as well


dvc/objects/stage.py
```diff
@@ -267,26 +284,29 @@ def stage(
...
-                raw = odb.get(obj.hash_info)
-            else:
-                staging.add(obj.path_info, obj.fs, obj.hash_info)
-                raw = staging.get(obj.hash_info)
-            # cleanup unneeded memfs tmpfile and return obj based on staging
-            # ODB fs/path
-            if obj.fs != raw.fs:
-                obj.fs.remove(obj.path_info)
-            obj.fs = raw.fs
-            obj.path_info = raw.path_info
```
efiop, line 283: image

dvc/objects/db/reference.py
```diff
@@ -0,0 +1,66 @@
...
+    """Reference ODB.
+
+    File objects are stored as ReferenceHashFiles which reference paths outside
+    of the staging ODB fs. Tree objects are stored natively.
+    """
+
+    def get(self, hash_info: "HashInfo"):
+        if hash_info.isdir:
+            return super().get(hash_info)
+        path_info = self.hash_to_path(hash_info.value)
```
efiop, line 28: For the record: as we've discussed with @pmrowla before, it might make sense to either store these ref objects in a some subdir in odb (e.g. `refs/` or smth) or maybe use a file extention (similar to `.dir`, e.g. `.ref`) to distinguish these objects from actual real objects.

This might get useful in cases where you need to store actual objects as well as ref objects and is kinda like git's object header, but encoded into a path or filename, which allows us to very easilly distinguish them on sight and through API calls like `list_objects`.

efiop:
Btw, even though refodb is not able to make dvc provide 100% git-like staging experience (we can't save actual files since they are too big), we could maybe improve `--no-commit` by just showing that in `dvc status`. We won't be able to make your data roll back to the staged state, but at least we won't need to store md5 of an uncommited (to .dvc/cache) file/dir, which is less hazardous and so `uncommited` and `missing cache` cases could be more easilly distinguished in functionality that cares (e.g. `dvc status`, `dvc diff`, etc). This will, of course, be possible, if staging refodb becomes permanently saved to disk, which is a future possible step. But just wanted to mention this, since it seems to allow us for a better ui or better functionality in general in the future.

pmrowla:
For the record:

had discussion with @efiop regarding potential future use cases for refodb

- replacing get with the ability to "checkout" directly from a remote odb seems promising
- checkout itself would be generalized into a "transfer/link from source odb into checkout refodb" where the refodb contains references to the output paths

efiop:
@Mergifyio rebase

dvc/state.py
```diff
@@ -122,10 +126,10 @@ def get(self, path_info, fs):
 
         value = self.md5s.get(inode)
 
-        if not value or value[0] != mtime or value[1] != size:
+        if not value or value[0] != mtime or value[1] != str(size):
             return None
 
-        return HashInfo("md5", value[2], size=int(size))
+        return HashInfo("md5", value[2], size=size)
```
efiop, line 132: This breaks backward compatibility for this db :slightly_frowning_face:

tests/utils/__init__.py
```diff
@@ -42,3 +42,15 @@ def dump_sv(stream, metrics, delimiter=",", header=True):
...
+def clean_staging():
+    from dvc.fs.memory import MemoryFileSystem
+    from dvc.objects.stage import _STAGING_MEMFS_PATH
+
+    try:
+        MemoryFileSystem().fs.rm(
+            f"memory://{_STAGING_MEMFS_PATH}", recursive=True
+        )
+    except FileNotFoundError:
+        pass
```
efiop, line 56: Would it make sense to ovverride fsspec's memfs so that we could have isolated instances?

dvc/dependency/repo.py
```diff
@@ -91,31 +92,32 @@ def changed_checksum(self):
...
+            used_obj_ids = defaultdict(set)
             rev = repo.get_rev()
             if locked and self.def_repo.get(self.PARAM_REV_LOCK) is None:
                 self.def_repo[self.PARAM_REV_LOCK] = rev
 
             path_info = PathInfo(repo.root_dir) / str(self.def_path)
             if not obj_only:
                 try:
-                    for odb, objs in repo.used_objs(
+                    for odb, obj_ids in repo.used_objs(
```
efiop, line 120: Should we go `obj_ids` -> `oids` and coin this name? Sure, it is a bit more cryptic, but short and sweet and makes sense if you know what you are doing. Just feels like we should get the terminology straight right away so it is settled, as it will be used more and more around the code.

dvc/hash_info.py
```diff
@@ -14,6 +14,7 @@ class HashInfo:
     value: Optional[str]
     size: Optional[int] = field(default=None, compare=False)
     nfiles: Optional[int] = field(default=None, compare=False)
+    obj_name: Optional[str] = field(default=None, compare=False)
```
efiop, line 17: Hm, looks like `obj` is leaking here. I see that you only use it for status (prev obj.name), maybe there is a nicer way to handle that without leaking obj things into HashInfo?

pmrowla: This is the remnants of NamedCache - to keep the current `dvc status -c` behavior (and output workspace path names instead of hashes) we have to return something during used object collection that maps objects/oids to path names. 

We could collect an additional dictionary that maps oids to names (so used_objs would return tuples of `(odb, oids, oid_names)`) so that `dvc status` can do the hash->name lookup, but at this point, `oid_names` would be a hack that only exists for legacy UI purposes (and is pretty much just the old `NamedCache`).

IMO, since `status -c`/`push`/`pull` operate on object IDs and don't actually have anything to do with "workspace files" at all, the nicest way to handle it would be to get rid of the name thing entirely, and change `status -c` so that it just outputs counts and size information for "total objects that would be uploaded, downloaded, or are missing" rather than the current behavior of listing workspace filenames for every object that would be uploaded/downloaded.

(But this would obviously be a significant behavior change for `dvc status -c`)

dvc/state.py
pmrowla: I'll adjust it so that we go back to writing size as a string in the state DB.

The main thing here was that I don't think the `utils` function should have been returning both as strings (the returned `mtime` does have to be a string since for dirs it's actually the hash of all the file mtimes, but that doesn't apply to `size`)

efiop: @pmrowla Agreed, that's part of an old code that used to be directly in old state class 🙁

dvc/hash_info.py
efiop: @pmrowla Makes sense! Then we could keep obj_name for now and reconsider it in the near future when we'll be revisiting push/pull/status ui. 👍

tests/utils/__init__.py
pmrowla: So the way this works now is that we do have isolated instances per-ODB. `objects.stage(odb, ...)` will return a refodb placed at `memory://_STAGING_MEMFS_PATH/<hash of destination ODB path>`. So in practical terms, staging ODB URLs are unique per repo local cache directory

The issue is that for testing purposes, even with unique staging ODBs, we will still have to manually clear what's in memfs for certain tests. In real-world CLI use, staging is cleared after every DVC command (when the process terminates), but in our tests we just repeatedly call `dvc.func()` which does not clear anything in memfs, since it's all done in the same process.

pmrowla: Since we return the odb instance each time we `stage()` something, we could maybe just generate a new staging ODB path per `stage()` call (using uuids), but this would end up causing us to duplicate a lot of behavior internally, since we call `stage()` multiple times even in a single DVC command.

output.commit, output.save, checkout all end up staging the same path whenever we `dvc commit` something. So if each call used a unique ODB path, we would end up recalculating the entire staged dir each time instead of just re-using the staged ref objects from the initial call

efiop: @pmrowla Feels like we'll probably end up with 1 staging odb per repo instance or something like that in the future, and so it makes sense that it looks like it does in this PR 👍 Let's keep it for now.

pmrowla:
Thinking about staging some more (with regard to the partial add issue), we will need to always write staged trees to disk somewhere prior to the formal save/transfer (whether it's in the actual ODB or somewhere else like .dvc/tmp).

With the current behavior for this PR where it's all done in memory, we wouldn't be able to recover from a partial/failed add since we would lose the dir cache/tree for the original "complete" directory

efiop (approved):

efiop:
@pmrowla Thank you! :pray:

efiop:
For the record: some followups that we've discussed before:

1. obj type is part of oid
2. base odb should be aware of obj type (will likely result in removal of ReferenceODB)
3. use protobuf? (not needed right now, as we don't need to serialize anything to an actual disk, but still)
4. hash type might be part of oid but we could keep plain 12/34 structure for now in refs/
5. ref objects could be a compatibility layer between different hash types.
6. should be possible to transfer all objects from one odb to another (e.g. run cache too). Aka `cloning`

PR #6365, opened 2021-07-27
odb: use regular upload() when the source filesystem is local

isidentical:
In the old remote logic, for transfering to remotes we used `upload()` instead of `upload_fobj()` which on some backends defaulted to `upload_fobj` and on some others used the native transfer libraries logic. For some backends like oss / ssh etc. this is faster than the `upload_fobj`, so this patch adds an extra check to the ODB to match against whether the source file system is local or not. If it is the local filesystem, we can safely use `_upload()` if not we can still fallback to `_upload_fobj()`.

### DIFF
--- a/dvc/objects/db/base.py
+++ b/dvc/objects/db/base.py
@@ -86,6 +86,8 @@ class ObjectDB:
                 if not isinstance(from_info, from_fs.PATH_CLS):
                     from_info = from_fs.PATH_CLS(from_info)
                 self.fs.upload(from_info, to_info)
+            elif isinstance(self.fs, LocalFileSystem):
+                from_fs.download_file(from_info, to_info)
             else:
                 with from_fs.open(from_info, mode="rb") as fobj:
                     self.fs.upload_fobj(fobj, to_info)
--- a/dvc/objects/db/base.py
+++ b/dvc/objects/db/base.py
@@ -5,6 +5,7 @@ from concurrent.futures import ThreadPoolExecutor
 from copy import copy
 from typing import TYPE_CHECKING, Optional
 
+from dvc.fs.local import LocalFileSystem
 from dvc.objects.errors import ObjectDBPermissionError, ObjectFormatError
 from dvc.objects.file import HashFile
 from dvc.progress import Tqdm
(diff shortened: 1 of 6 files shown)


dvc/objects/db/base.py
```diff
@@ -75,6 +76,38 @@ def get(self, hash_info: "HashInfo"):
...
+    def _add_file(self, from_fs, from_info, to_info, move):
+        self.makedirs(to_info.parent)
+        use_move = isinstance(from_fs, type(self.fs)) and move
+        try:
+            if use_move:
+                self.fs.move(from_info, to_info)
+            elif isinstance(from_fs, LocalFileSystem):
+                if not isinstance(from_info, from_fs.PATH_CLS):
+                    from_info = from_fs.PATH_CLS(from_info)
+                self.fs.upload(from_info, to_info)
```
efiop, line 88: Random thought: I guess we could also put a check for `self.fs` being a local one and then using `from_fs.download`?

dvc/objects/db/base.py
```diff
@@ -75,6 +76,40 @@ def get(self, hash_info: "HashInfo"):
...
+            elif isinstance(self.fs, LocalFileSystem):
+                from_fs.download_file(from_info, to_info)
```
efiop, line 90: Btw, should both of these use `_file` methods?

isidentical: What do you mean by that?

efiop: @isidentical I see that we use `fs.upload` but `fs.download_file`. Wondering if they both should be `upload/download` or `upload_file/download_file` (likely the latter if we take fsspec's `get/put_file` future in the context, but maybe I'm missing something important here)

isidentical: We don't have `upload_file()`, so the `upload()` is `put_file`. And `download_file()` is the `get_file()`. The reason that I didn't use `download()` is that, it has an extra check against whether the source path is a directory, which is an waste of API call for this case where we know we are adding a single entity (file).

efiop: Makes sense, thank you! I guess we'll switch to get_file/put_file when reducing fsspec_wrapper in the future.

pmrowla:
LGTM

efiop (approved):

PR #6370, opened 2021-07-29
ssh: handle concurrency issues for raw command execution

isidentical:
From https://github.com/iterative/dvc/pull/6295#issuecomment-888495315. There are 2 primary issues, one is that we used to max out all the sessions with SFTP channels, so after a certain amount of operations there wasn't any channels to use for SSH. Now we allocate 2 channels for it. The second issue is that, `AsyncFileSystem` that fsspec offers overrides the `move()` command with it's own `copy()`+`remove()` implementation. Now that is also temporarily fixed on the sshfs, and will submit a patch against fsspec upstream.

### DIFF
--- a/dvc/fs/fsspec_wrapper.py
+++ b/dvc/fs/fsspec_wrapper.py
@@ -110,6 +110,9 @@ class FSSpecWrapper(BaseFileSystem):
         for file in self.find(path_info, **kwargs):
             yield path_info.replace(path=file)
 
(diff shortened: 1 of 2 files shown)


efiop (approved):

efiop:
Tested in my setup, works like charm now. Thank you! :pray:

isidentical:
Thanks for giving it a try @efiop!

PR #6378, opened 2021-08-02
Fix merging of params when empty

### LINKED ISSUE #6374: tests: test_experiments::test_untracked is failing in CI
may be related to existing flaky pygit tests issue?

skshetry:
The recent release of python-benedict==0.24.1 started ignoring empty dictionaries
during the merge. The way we expected to work was questionable and also how
benedict itself tries to handle the references to an empty dictionary is
also questionable.

I consider this to be a proper way to handle this even if this gets fixed
on the benedict side.

To demonstrate the problem with referencing an empty dictionary:
```pycon
>>> x = {}
>>> y = dict(x)
>>> y.update({"foo": "bar"})
>>> x
{}
>>> y
{'foo': 'bar'}
```

Benedict tries to handle in-place updates, so it's tricky for it to handle this case of empty dictionaries as the references are not preserved. And, since `modify_data` requires updating things in-place, we require references to be preserved. So, it seems it's our duty to try to preserve references than it being left to benedict alone.

Fixes #6374.

### DIFF
--- a/dvc/repo/experiments/__init__.py
+++ b/dvc/repo/experiments/__init__.py
@@ -352,8 +352,7 @@ class Experiments:
 
     def _update_params(self, params: dict):
         """Update experiment params files with the specified values."""
-        from benedict import benedict
-
+        from dvc.utils.collections import merge_params
         from dvc.utils.serialize import MODIFIERS
 
         logger.debug("Using experiment params '%s'", params)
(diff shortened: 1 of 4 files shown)


skshetry:
I'll try adding creating a patch to the upstream tomorrow, and we can proceed accordingly.

efiop (approved):
Thank you! :pray:

PR #6379, opened 2021-08-02
Unittest to pytest: moved few unittest tests to pytest 

nik123:
Part of #1819 

PR doesn't require documentation updates.

### DIFF
--- a/tests/func/test_init.py
+++ b/tests/func/test_init.py
@@ -1,53 +1,52 @@
 import logging
 import os
 
+import pytest
+
 from dvc.config import Config
 from dvc.exceptions import InitError
 from dvc.main import main
 from dvc.repo import Repo as DvcRepo
-from tests.basic_env import TestDir, TestGit
 
 
-class TestInit(TestGit):
-    def _test_init(self):
-        self.assertTrue(os.path.exists(DvcRepo.DVC_DIR))
-        self.assertTrue(os.path.isdir(DvcRepo.DVC_DIR))
-
-    def test_api(self):
-        DvcRepo.init().close()
-
-        self._test_init()
-
-    def test_cli(self):
-        ret = main(["init"])
-        self.assertEqual(ret, 0)
-
-        self._test_init()
+def _check_dirs():
+    assert os.path.exists(DvcRepo.DVC_DIR)
+    assert os.path.isdir(DvcRepo.DVC_DIR)
 
 
-class TestDoubleInit(TestInit):
-    def test(self):
-        ret = main(["init"])
-        self.assertEqual(ret, 0)
-        self._test_init()
-
-        ret = main(["init"])
-        self.assertNotEqual(ret, 0)
-        self._test_init()
-
-        ret = main(["init", "--force"])
-        self.assertEqual(ret, 0)
-        self._test_init()
+def test_api_init(scm):
+    DvcRepo.init().close()
+    _check_dirs()
 
 
-class TestInitNoSCMFail(TestDir):
-    def test_api(self):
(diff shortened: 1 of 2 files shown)


tests/func/test_scm.py
```diff
@@ -34,27 +34,27 @@ def test_init_sub_dir(tmp_dir):
...
-        self.assertTrue(G.is_tracked(self.FOO))
-        G.gitpython.repo.index.remove([self.FOO], working_tree=True)
-        self.assertFalse(G.is_tracked(foo))
-        self.assertFalse(G.is_tracked(self.FOO))
-        self.assertFalse(G.is_tracked("not-existing-file"))
+def test_commit(tmp_dir, scm):
+    tmp_dir.gen({"foo": "foo"})
+    scm.add(["foo"])
+    scm.commit("add")
+    assert "foo" in Repo(tmp_dir).git.ls_files()
```
karajan1001, line 41: How about this one.
```
scm.is_tracked("foo")
```

instead of 

```
    assert "foo" in Repo(tmp_dir).git.ls_files()
```

tests/func/test_init.py
```diff
@@ -1,53 +1,52 @@
...
+    assert os.path.exists(DvcRepo.DVC_DIR)
+    assert os.path.isdir(DvcRepo.DVC_DIR)
 
-    def test_api(self):
-        DvcRepo.init().close()
 
-        self._test_init()
+def test_api_init(scm):
+    DvcRepo.init().close()
+    _check_dirs()
```
skshetry, line 19: Just the check for `isdir` should be enough, so I think it'd be more clearer if we get rid of `_check_dirs` and replace them with `os.path.isdir(repo.DVC_DIR)`. 1 less indirection.

tests/func/test_scm.py
nik123: I've replaced `ls_files` with `is_tracked`. But I'm not sure if `test_commit` is really needed because it seems that `test_is_tracked` fully covers `test_commit` case.

pmrowla: So, the reason that this did not use `is_tracked` before is that `is_tracked` is still using our SCM class, and the tests for `is_tracked` itself also depends on the assumption that `scm.commit` works (so we want to test `scm.commit` independently of `is_tracked`.

To verify that `scm.commit` works, we were testing it directly using CLI git before (the `git.ls_files` call spawns an actual CLI `git ls-files` call via gitpython).

tests/func/test_scm.py
```diff
@@ -34,27 +34,27 @@ def test_init_sub_dir(tmp_dir):
...
-        self.assertTrue(G.is_tracked(self.FOO))
-        G.gitpython.repo.index.remove([self.FOO], working_tree=True)
-        self.assertFalse(G.is_tracked(foo))
-        self.assertFalse(G.is_tracked(self.FOO))
-        self.assertFalse(G.is_tracked("not-existing-file"))
+def test_commit(tmp_dir, scm):
+    tmp_dir.gen({"foo": "foo"})
+    scm.add(["foo"])
+    scm.commit("add")
+    assert scm.is_tracked("foo")
```
pmrowla, line 41:
```suggestion
    assert "foo" in scm.gitpython.git.ls_files()
```

(see my comments in the outdated thread)

tests/func/test_scm.py
nik123: ok. Now `test_commit` makes sense.

karajan1001 (approved):

efiop:
Thank you, @nik123 ! :pray:

PR #6382, opened 2021-08-03
Show name of experiments in the queue

### LINKED ISSUE #6050: exp show: show names for the queued experiments

karajan1001:
fix #6050

### DIFF
--- a/dvc/repo/experiments/__init__.py
+++ b/dvc/repo/experiments/__init__.py
@@ -880,7 +880,7 @@ class Experiments:
         ref = self.scm.describe(rev, base=EXPS_NAMESPACE, exclude=exclude)
         if ref:
             return ExpRefInfo.from_ref(ref).name
-        return None
+        return self.stash_revs[rev].name
 
     def get_running_exps(self) -> Dict[str, int]:
         """Return info for running experiments."""
--- a/dvc/repo/experiments/__init__.py
+++ b/dvc/repo/experiments/__init__.py
@@ -877,11 +877,13 @@ class Experiments:
         Prefers tags, branches (heads), experiments in that orer.
         """
         exclude = f"{EXEC_NAMESPACE}/*"
-        if rev in self.stash_revs:
-            return self.stash_revs[rev].name
         ref = self.scm.describe(rev, base=EXPS_NAMESPACE, exclude=exclude)
         if ref:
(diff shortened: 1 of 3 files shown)


dvc/repo/experiments/__init__.py
```diff
@@ -877,6 +877,8 @@ def get_exact_name(self, rev: str):
         Prefers tags, branches (heads), experiments in that orer.
         """
         exclude = f"{EXEC_NAMESPACE}/*"
+        if rev in self.stash_revs:
+            return self.stash_revs[rev].name
```
pmrowla, line 881: This stash check should be done last - experiments (including stashed experiments) should have lower priority than tags or branch names

pmrowla (changes requested):

dvc/repo/experiments/__init__.py
karajan1001: Reorder here would make the code a little corrupt. And shouldn't the stash ones be mutually exclusive with the commit ones? But I agreed that making commit ones ahead of stash ones is a logical/common choice, I also hesitated here before I wrote them down.

pmrowla: It's unlikely that a git tag or branch would ever point to one of our stash commits, but it's technically a legal state - you can manually tag any git commit, and manually create a `refs/heads/branch_name` entry that points to any git commit.
This includes our stash commits, so we have to account for it.

pmrowla (approved):

karajan1001:
image
Add a example to this.

PR #6385, opened 2021-08-04
Made DVCLive `path` optional

daavoo:

- Background: https://github.com/iterative/dvclive/issues/60

- `dvc.org` PR: https://github.com/iterative/dvc.org/pull/2686

This P.R. makes `--live` and `--live-no-cache` booleans by adding a new arg `--live-path`. 

The whole point is to make the CLI behavior consistent with the `dvclive` API, where `path` is an optional argument defaulting to `dvclive`.

The current call:

```bash
dvc stage add --name train --live dvclive python train.py
```

Would be now

```bash
dvc stage add --name train --live python train.py
```

And the path would stil be customizable with `--live-path`.


dberenbaum:
I think this is a more clear and consistent UI, but TBH I have some reservations:

1. It's a breaking change (not a big deal yet for dvclive but moreso for dvc).
2. This would add a 3rd dvclive option to an already crowded set of command options.
3. It's unclear if this makes it easier or harder for users depending on whether they would frequently use the default `dvclive` path (personally, I probably would use a different path that will make sense to people unfamiliar with dvclive).

daavoo:
> I think this is a more clear and consistent UI, but TBH I have some reservations:
> 
>     1. It's a breaking change (not a big deal yet for dvclive but moreso for dvc).
> 
>     2. This would add a 3rd dvclive option to an already crowded set of command options.
> 
>     3. It's unclear if this makes it easier or harder for users depending on whether they would frequently use the default `dvclive` path (personally, I probably would use a different path that will make sense to people unfamiliar with dvclive).

So I have an initial implementation that kind of addressed those points:

```python
    parser.add_argument(
         "--live", 
         nargs='?', const="dvclive",
         help="Declare output as dvclive.", metavar="<path>"
    )
```

Which would make it compatible with the previous behavior:

```bash
dvc stage add --live training_metrics --name train  python train.py
```

And also allow it to work with defaults:

```bash
dvc stage add --live --name train  python train.py
```

The problem is that it has a major gotcha when `--live` is the last argument used before the command definition (which, unfortunately is how we use it in the checkpoints docs). Which was the main reason for which I discarded it (although now I see your points it looks better than the current P.R.):

```bash
# Works fine
dvc stage add --name train  --live training_metrics  python train.py
# Gotcha, python is parsed as the `path` for `--live`
dvc stage add --name train  --live  python train.py
```

I`m not sure if this can be somehow handled properly by `argparse`, I don't think so. It could be handled in the `validate_args` and raising an exception.

It could  also be handled though on the documentation side, by stablishing a more consistent pattern of passing the *optional* arguments always before the *required* (`-n <name>`) and *positional* (`command`). I think it would make less likely for an user to encounter this gotcha.

pared:
> `dvc stage add -n train  --live  python train.py`

@daavoo I think it could work like this:

`dvc stage add -n train  --live --  python train.py`

daavoo:
> > `dvc stage add -n train  --live  python train.py`
> 
> @daavoo I think it could work like this:
> 
> `dvc stage add -n train --live -- python train.py`

Does that requires some change to `argparse`?. 

I tried it and:

```bash
dvc stage add -n foo --live -- echo "foo"
```

Results in:

```yaml
stages:
  foo:
    cmd: -- echo foo
    live:
      dvclive:
        summary: true
        html: true
```

pared:
@daavoo aah right, that would require adding default path on `dvc` side too...

pared (commented):
I don't think we should introduce this change in this form.
1. `live` and `live-no-cache` resembles other outputs types, so making it boolean will be confusing users, who are used to the fact that there should go path
2. to provide `--live path` one will need `--live --live-path path` which makes ux more complicated

daavoo:
> I don't think we should introduce this change in this form.
> 
>     1. `live` and `live-no-cache` resembles other outputs types, so making it boolean will be confusing users, who are used to the fact that there should go path
> 
>     2. to provide `--live path` one will need `--live --live-path path` which makes ux more complicated

Would it make sense to indroduce if the change is implemented as described in the commend above?

pared:
@daavoo 

Do you mean documenting it properly? In this case, we will be still making `live` inconsistent with other output types (no other option allows no arg).
I am not sure we actually need that functionality. 
Default `dvclive` path was introduced so that there are no problems with executing training code without dvc context - if implementing it on `dvc` side shows that much problems we might reconsider it. It would be a convenience for user to have that option, but if there is no easy way to do that, we can draw a line on dvc side.

daavoo:
> @daavoo
> 
> Do you mean documenting it properly? In this case, we will be still making `live` inconsistent with other output types (no other option allows no arg).
> I am not sure we actually need that functionality.
> Default `dvclive` path was introduced so that there are no problems with executing training code without dvc context - if implementing it on `dvc` side shows that much problems we might reconsider it. It would be a convenience for user to have that option, but if there is no easy way to do that, we can draw a line on dvc side.

Makes sense. I will be closing this for now.

This made me realized one potential point of discussion in the future. `dvclive` is somehow already inconsistent with the other outputs if we consider that the API is implemented outside `dvc`, so there are some things that feel weird from an user perspective. 

There are some potential problems with this, one minor would be the inconsistency that this P.R. tried to address. 

Other more relevant would be that any further options added to `dvclive.init` also need to be addressed in the `dvc stage add` side, potentially overburden the command. I feel like this has been a hidden consideration when discussing `init` arguments in `dvclive` (https://github.com/iterative/dvclive/issues/128).

When looking at other options that have subfields, i.e. `plots`, the subfields can't be modified directly from `dvc stage add` but only by manually edifting the `yaml` or with `dvc plots modify`. 

Would it make sense to follow a similar approach for `live`, in  order to keep `dvc stage add` clean from existing or future `live` subfields?

pared:
@daavoo this is a good point, and while I think that it would be good to have consistient behaviour here, we still have `dvc run` to handle, which requires all params upfront, while plots require them only when we intend to use `dvc plots show/diff`. This is something that have yet not been addressed. Could you create issue for that? We could host the discussion how we could apporach that there.

PR #6393, opened 2021-08-09
Remove a special queued experiments

### LINKED ISSUE #6157: exp: remove <queued_experiment>
Currently `exp remove` doesn't work for a single queued experiment. This is a problem if 1001 experiments are queued but the last one was an accident. Starting from scratch (`dvc exp remove --queue`) is a poor solution.

karajan1001:
fix #6157
1. dvc exp remove not accept queued experiments name
2. add some tests for this feature

### DIFF
--- a/dvc/repo/experiments/remove.py
+++ b/dvc/repo/experiments/remove.py
@@ -29,14 +29,14 @@ def remove(repo, refs_or_revs=None, queue=False, **kwargs):
 
 
 def _get_exp_stash_index(repo, ref_or_rev: str) -> Optional[int]:
-    stash_ref_infos = repo.experiments.stash_revs
-    print("*" * 100)
-    for rev, ref_info in stash_ref_infos.items():
-        print(rev, ref_info)
+    stash_revs = repo.experiments.stash_revs
+    for _, ref_info in stash_revs.items():
         if ref_info.name == ref_or_rev:
             return ref_info.index
-        if rev == ref_or_rev:
-            return ref_info.index
+    rev = repo.scm.resolve_rev(ref_or_rev)
+    if rev in stash_revs:
+        return stash_revs.get(rev).index
+
     return None
 
 
--- a/dvc/repo/experiments/remove.py
+++ b/dvc/repo/experiments/remove.py
@@ -23,8 +23,8 @@ def remove(repo, exp_names=None, queue=False, **kwargs):
         removed += len(repo.experiments.stash)
         repo.experiments.stash.clear()
     if exp_names:
-        remained = _remove_commited_experiments(repo, exp_names)
-        remained = _remove_queued_experiements(repo, remained)
+        remained = _remove_commited_exps(repo, exp_names)
+        remained = _remove_queued_exps(repo, remained)
         if remained:
             logger.warning(
                 "'{}' is neither a valid experiment reference"
@@ -81,7 +81,7 @@ def _get_ref(ref_infos, name, cur_rev) -> Optional[ExpRefInfo]:
     return ref_infos[0]
 
 
-def _remove_commited_experiments(repo, refs: List[str]) -> List[str]:
+def _remove_commited_exps(repo, refs: List[str]) -> List[str]:
     remain_list = []
     remove_list = []
     for ref in refs:
@@ -95,7 +95,7 @@ def _remove_commited_experiments(repo, refs: List[str]) -> List[str]:
     return remain_list
 
 
-def _remove_queued_experiements(repo, refs_or_revs: List[str]) -> List[str]:
+def _remove_queued_exps(repo, refs_or_revs: List[str]) -> List[str]:
     remain_list = []
     for ref_or_rev in refs_or_revs:
         stash_index = _get_exp_stash_index(repo, ref_or_rev)
--- a/dvc/repo/experiments/remove.py
+++ b/dvc/repo/experiments/remove.py
@@ -26,7 +26,7 @@ def remove(repo, exp_names=None, queue=False, **kwargs):
         remained = _remove_commited_exps(repo, exp_names)
         remained = _remove_queued_exps(repo, remained)
         if remained:
-            logger.warning(
+            raise InvalidArgumentError(
                 "'{}' is not a valid experiment".format(";".join(remained))
             )
         removed += len(exp_names) - len(remained)
--- a/dvc/repo/experiments/remove.py
+++ b/dvc/repo/experiments/remove.py
@@ -1,4 +1,5 @@
 import logging
+from typing import Optional
 
 from dvc.exceptions import InvalidArgumentError
 from dvc.repo import locked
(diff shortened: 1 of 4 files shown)


pmrowla:
So one other thing that will be needed is removing queued experiments by revision.

For regular experiments we only support removing them by ref (or shortened name), since the operation to remove a finished experiment is a git ref deletion, and using commit SHA's here would be ambiguous.

But for queued (stashed) experiments, they will only be named if the user provided one themselves with `exp run -n/--name`. Otherwise, it can only be identified by a git commit SHA. Both names and SHAs should be allowed for queued experiments since the remove operation is not actually a name/ref deletion, it is just a stash drop (removing a SHA from the git reflog).

dvc/repo/experiments/remove.py
```diff
@@ -61,3 +65,16 @@ def _get_ref(ref_infos, name, cur_rev):
...
+def _remove_exp_by_name(repo, exp_name: str):
+    ref_info = _get_exp_ref(repo, exp_name)
+    if ref_info is not None:
+        remove_exp_refs(repo.scm, [ref_info])
+    else:
+        stash_index = _get_exp_stash_index(repo, exp_name)
+        if stash_index is None:
+            raise InvalidArgumentError(
+                f"'{exp_name}' is not a valid experiment name"
+            )
```
pmrowla, line 79: We need to account for whether or not `exp_name` is the SHA for a stash revision here

pmrowla (changes requested):

karajan1001:
> But for queued (stashed) experiments, they will only be named if the user provided one themselves with `exp run -n/--name`. Otherwise, it can only be identified by a git commit SHA. Both names and SHAs should be allowed for queued experiments since the remove operation is not actually a name/ref deletion, it is just a stash drop (removing a SHA from the git reflog).

Yes, there comes another question should the committed ones also accept a revision name? It is natural for uses to extend the usage of queued ones to a committed one.

pmrowla:
> Yes, there comes another question should the committed ones also accept a revision name? It is natural for uses to extend the usage of queued ones to a committed one.

No, because the actual operations for removing a queued experiment and a regular one are two different things. Since removing a regular experiment is a git ref deletion, specifying a commit SHA is ambiguous - it's possible for multiple experiment refs to point to a single git SHA, so it is better for us to require the user to be explicit here (and use the ref/name).

Removing queued experiments is not deleting a git ref, it's dropping a git SHA from a list of stashed git SHAs.

pared (approved):

dvc/repo/experiments/remove.py
```diff
@@ -61,3 +69,17 @@ def _get_ref(ref_infos, name, cur_rev):
...
+def _remove_exp_by_ref_or_rev(repo, ref_or_rev: str):
+    ref_info = _get_exp_ref(repo, ref_or_rev)
+    if ref_info is not None:
+        remove_exp_refs(repo.scm, [ref_info])
+    else:
+        stash_index = _get_exp_stash_index(repo, ref_or_rev)
+        if stash_index is None:
+            raise InvalidArgumentError(
+                f"'{ref_or_rev}' is neither a valid experiment reference"
+                " nor a queued experiment revision"
```
karajan1001, line 83: Quite weird here, because the asymmetric in committed and queued experiments.

dvc/repo/experiments/remove.py
```diff
@@ -12,41 +13,48 @@
...
+            removed += 1
     return removed
 
 
-def _get_exp_refs(repo, exp_names):
+def _get_exp_stash_index(repo, ref_or_rev: str) -> Optional[int]:
+    stash_ref_infos = repo.experiments.stash_revs
+    print("*" * 100)
+    for rev, ref_info in stash_ref_infos.items():
+        print(rev, ref_info)
```
pmrowla, line 35: leftover print statements

dvc/repo/experiments/remove.py
```diff
@@ -12,41 +13,48 @@
...
+    if refs_or_revs:
+        for ref_or_rev in refs_or_revs:
+            _remove_exp_by_ref_or_rev(repo, ref_or_rev)
+            removed += 1
     return removed
 
 
-def _get_exp_refs(repo, exp_names):
+def _get_exp_stash_index(repo, ref_or_rev: str) -> Optional[int]:
+    stash_ref_infos = repo.experiments.stash_revs
```
pmrowla, line 32: `experiments.stash_revs` is not a list of `ref_infos`. It's a dictionary of stash entries (which are just loose individual git commits, not named git references): https://github.com/iterative/dvc/blob/38cd7d01c2fd14fc28c68f70bc963b105ce64db0/dvc/repo/experiments/__init__.py#L86

dvc/repo/experiments/remove.py
```diff
@@ -12,41 +13,48 @@
...
+    print("*" * 100)
+    for rev, ref_info in stash_ref_infos.items():
+        print(rev, ref_info)
+        if ref_info.name == ref_or_rev:
+            return ref_info.index
+        if rev == ref_or_rev:
```
pmrowla, line 38: This will not match the case where `ref_or_rev` is a shortened git SHA. It needs to be resolved into a full SHA with `rev = scm.resolve_rev(ref_or_rev)`

you should then be able to just do

```
if rev in stash_revs:
    index = stash_revs.get[rev].index
```

dvc/repo/experiments/remove.py
pmrowla: I would separate this into two methods, one for removing a complete experiment ref (only by ref name) and one for removing stash entries (by either name or revision).

karajan1001: But without a flag, we can not know if the input is a ref or a revision. Some ref might use a shortcut name for example.
`epoch_1_layers3` - > `ep1la3` hard to distinguish them.

dvc/repo/experiments/remove.py
karajan1001: How to set an auto-commit for the VSCODE, always get some changes not commited.

dvc/repo/experiments/remove.py
pmrowla: Right, you would start by calling
`_remove_exp_ref(name)`.

If that fails because an exact exp ref match for `name` does not exist, then you would try the stash remove function. Which also needs to try for an exact name match first, and then handling it as git rev second.

pmrowla: Basically you can parse the list of names that we get at the start into two separate lists before trying to remove anything at all. One list that matches exact refs, and a second list that does not match exact refs. Everything in the second list is either a stash entry (by name or rev) or is invalid entirely.

karajan1001:
[asciicast](https://asciinema.org/a/TXwGF1KpjDmDtmB8PCOLEiMMz)

Some example to this

dvc/repo/experiments/remove.py
```diff
@@ -21,32 +23,48 @@ def remove(repo, exp_names=None, queue=False, **kwargs):
...
-        ref_infos = list(_get_exp_refs(repo, exp_names))
-        remove_exp_refs(repo.scm, ref_infos)
-        removed += len(ref_infos)
+        remained = _remove_commited_exps(repo, exp_names)
+        remained = _remove_queued_exps(repo, remained)
+        if remained:
+            logger.warning(
+                "'{}' is neither a valid experiment reference"
+                " nor a revision of queued experiment".format(
+                    ";".join(remained)
```
pmrowla, line 32: This message can just be `'{}' is not a valid experiment`

dvc/repo/experiments/remove.py
```diff
@@ -21,32 +23,48 @@ def remove(repo, exp_names=None, queue=False, **kwargs):
         removed += len(repo.experiments.stash)
         repo.experiments.stash.clear()
     if exp_names:
-        ref_infos = list(_get_exp_refs(repo, exp_names))
-        remove_exp_refs(repo.scm, ref_infos)
-        removed += len(ref_infos)
+        remained = _remove_commited_exps(repo, exp_names)
+        remained = _remove_queued_exps(repo, remained)
+        if remained:
```
karajan1001, line 28: A behavior change here, because we had already removed all of the matched experiments, raise an exception here is needless.

tests/func/experiments/test_remove.py
```diff
@@ -0,0 +1,49 @@
+from funcy import first
+
+from dvc.repo.experiments.utils import exp_refs_by_rev
+
+
+def test_remove_experiments_by_ref(tmp_dir, scm, dvc, exp_stage, caplog):
+    results = dvc.experiments.run(exp_stage.addressing, params=["foo=2"])
+    exp = first(results)
+    ref_info = first(exp_refs_by_rev(scm, exp))
+
```
dberenbaum, line 10: Thanks for expanding on these! As long as we are adding more tests and conditions, what do you think about adding another experiment and asserting that it doesn't get removed? What about testing multiple refs passed at once (`dvc exp remove abc123 def456`)?

tests/func/experiments/test_remove.py
```diff
@@ -0,0 +1,49 @@
...
+    assert dvc.experiments.remove(["non-exist"]) == 0
+    assert ("'non-exist' is not a valid experiment") in caplog.text
+
+    removed = dvc.experiments.remove([str(ref_info)])
+    assert removed == 1
+    assert scm.get_ref(str(ref_info)) is None
+
+
+def test_remove_all_queued_experiments(tmp_dir, scm, dvc, exp_stage):
```
dberenbaum, line 19: What about testing that a committed exp doesn't get removed?

tests/func/experiments/test_remove.py
```diff
@@ -0,0 +1,49 @@
...
+            exp_stage.addressing, params=[f"foo={i}"], queue=True
+        )
+
+    assert len(dvc.experiments.stash) == queue_length
+    removed = dvc.experiments.remove(queue=True)
+    assert removed == queue_length
+    assert len(dvc.experiments.stash) == 0
+
+
+def test_remove_special_queued_experiments(tmp_dir, scm, dvc, exp_stage):
```
dberenbaum, line 32: More test case ideas to consider:
* Assert that committed exp is not removed here?
* Assert that multiple queued exp refs can be removed at once?
* Remove a mix of queued exps and committed exp refs?

karajan1001:
Better tests cases. 
1. remove a list of committed and non-exist exp.
2. Tests that committed exp not be affected by `remove --queue`.
3. remove a list of mixed queued and committed exp.

We still need to raise Exception in a mixed committed and non exist case. Because of  only by this the return values of `dvc exp remove` would be a non-zero value.

tests/func/experiments/test_remove.py
```diff
@@ -0,0 +1,74 @@
...
+
+    for i in range(queue_length):
+        results = dvc.experiments.run(
+            exp_stage.addressing, params=[f"foo={i}"]
+        )
+        ref_info = first(exp_refs_by_rev(scm, first(results)))
+        ref_list.append(str(ref_info))
+
+    with pytest.raises(InvalidArgumentError):
+        assert dvc.experiments.remove(ref_list[:2] + ["non-exist"])
```
dberenbaum, line 20: Would `dvc.experiments.remove(["non-exist"] + ref_list[:2])` have the same behavior?

karajan1001: of course.

pmrowla (approved):

Issue #6006, opened 2021-05-13
Remove pushed experiment from git remote


kwon-young:
Push an experiment to a git remote:

```
$ dvc exp push origin myexpe
```

Then:

```
$ dvc exp list origin
master:
   myexpe
```

Currently, only local experiment can be removed with:

```
$ dvc exp remove myexpe
```

but `dvc exp list origin` still shows `myexpe`.

I would like to also be able to remove a pushed experiment on the remote.

pmrowla:
For now you can just use git directly

```shell
# get list of exps pushed to origin
$ git ls-remote origin "refs/exps/*"

# remove one of the listed refs
$ git push -d origin refs/exps/path/to/ref
```

dberenbaum:
@pmrowla Do you think there's anything to discuss here, or is this a straightforward feature to implement?

pmrowla:
@dberenbaum it still needs a discussion of what the usage should look like

dberenbaum:
I think we actually need options for both `dvc exp remove --git-remote` (as mentioned above) and `dvc exp gc --git-remote` (to clean up lots of old experiments that aren't relevant anymore).

> @dberenbaum it still needs a discussion of what the usage should look like

Can you elaborate? Does the suggested syntax above help?

pmrowla:
@dberenbaum My original comment was regarding whether it should be tied to `exp remove/gc` or whether it should look more like git's `push -d` workflow.

Also, I do think it's worth asking the question of "is removing remote experiments necessary"? There are reasons to disallow it - the same reasons that github always keeps all pull request refs and always keeps references to all remote git branches that have ever existed (including remote branches that users manually to "delete").

Should removing remote exps actually delete refs, or should it just hide them in the way that github "removes" git branches (and just move them into a different namespace so they don't show up in git CLI "list remote branch" commands)?

Or maybe what we really need is a better way to filter the results of `exp list <git remote>` in the first place instead of any remove functionality at all?

pmrowla:
Also I know this came up in the CML auto-push discussion as well (since we will be generating a lot of experiments that way). Maybe the CML runner pushed exps should just go into their own namespace that can be regularly cleaned on the CML side

dberenbaum:
> My original comment was regarding whether it should be tied to `exp remove/gc` or whether it should look more like git's `push -d` workflow.
> 
[...]

PR #6394, opened 2021-08-09
Remove all of the experiments in the workspace

### LINKED ISSUE #5676: Remove all experiments in the workspace
Have a `dvc exp remove --workspace` command or similar to drop all experiments in the workspace. Should be equivalent to `dvc exp list --names-only | xargs dvc exp remove`. This allows users to reset their experiments workspace if they have made mistakes or were playing around with some changes they ultimately want to roll back.

Some background:
* https://discord.com/channels/485586884165107732/485596304961962003/823901428652965898
* https://github.com/iterative/dvc/issues/5553#issuecomment-791717274
* https://github.com/iterative/dvc/issues/5437#issuecomment-776861639

karajan1001:
fix #5676
1. remove all the experiments in workspace
2. add tests for it

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -750,6 +750,7 @@ class CmdExperimentsRemove(CmdBase):
             exp_names=self.args.experiment,
             queue=self.args.queue,
             workspace=self.args.workspace,
+            all_baseline=self.args.all,
         )
 
         return 0
@@ -1246,7 +1247,12 @@ def add_parser(subparsers, parent_parser):
     remove_group.add_argument(
         "--workspace",
         action="store_true",
-        help="Remove all experiments in current workspace.",
+        help="Remove all experiments based on current workspace.",
+    )
+    remove_group.add_argument(
+        "--all",
+        action="store_true",
+        help="Remove all experiments.",
     )
     experiments_remove_parser.add_argument(
         "experiment",
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -750,7 +750,7 @@ class CmdExperimentsRemove(CmdBase):
             exp_names=self.args.experiment,
             queue=self.args.queue,
             workspace=self.args.workspace,
-            all_baseline=self.args.all,
+            clear_all=self.args.all,
         )
 
         return 0
@@ -1245,11 +1245,13 @@ def add_parser(subparsers, parent_parser):
         "--queue", action="store_true", help="Remove all queued experiments."
     )
     remove_group.add_argument(
+        "-w",
         "--workspace",
         action="store_true",
         help="Remove all experiments based on current workspace.",
     )
     remove_group.add_argument(
+        "-A",
         "--all",
         action="store_true",
         help="Remove all experiments.",
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -747,7 +747,9 @@ class CmdExperimentsRemove(CmdBase):
     def run(self):
 
         self.repo.experiments.remove(
-            exp_names=self.args.experiment, queue=self.args.queue
+            exp_names=self.args.experiment,
+            queue=self.args.queue,
+            workspace=self.args.workspace,
         )
 
         return 0
@@ -1237,8 +1239,14 @@ def add_parser(subparsers, parent_parser):
         help=EXPERIMENTS_REMOVE_HELP,
         formatter_class=argparse.RawDescriptionHelpFormatter,
     )
-    experiments_remove_parser.add_argument(
+    remove_group = experiments_remove_parser.add_mutually_exclusive_group()
+    remove_group.add_argument(
         "--queue", action="store_true", help="Remove all queued experiments."
+    )
+    remove_group.add_argument(
+        "--workspace",
+        action="store_true",
+        help="Remove all experiments in current workspace.",
     )
     experiments_remove_parser.add_argument(
         "experiment",
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -749,7 +749,6 @@ class CmdExperimentsRemove(CmdBase):
         self.repo.experiments.remove(
             exp_names=self.args.experiment,
             queue=self.args.queue,
-            workspace=self.args.workspace,
             clear_all=self.args.all,
         )
 
@@ -1245,16 +1244,10 @@ def add_parser(subparsers, parent_parser):
         "--queue", action="store_true", help="Remove all queued experiments."
     )
     remove_group.add_argument(
-        "-w",
-        "--workspace",
-        action="store_true",
-        help="Remove all experiments based on current workspace.",
-    )
-    remove_group.add_argument(
         "-A",
         "--all",
         action="store_true",
-        help="Remove all experiments.",
+        help="Remove all committed experiments.",
     )
     experiments_remove_parser.add_argument(
         "experiment",
--- a/dvc/repo/experiments/remove.py
+++ b/dvc/repo/experiments/remove.py
@@ -7,21 +7,25 @@ from dvc.repo.scm_context import scm_context
 from dvc.scm.base import RevError
 
 from .base import EXPS_NAMESPACE, ExpRefInfo
-from .utils import exp_refs_by_name, remove_exp_refs
+from .utils import exp_refs, exp_refs_by_name, remove_exp_refs
 
 logger = logging.getLogger(__name__)
 
 
 @locked
 @scm_context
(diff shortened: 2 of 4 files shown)


dvc/repo/experiments/remove.py
```diff
@@ -5,21 +5,25 @@
...
         return 0
 
     removed = 0
-    if queue:
+    if queue or workspace:
         removed += len(repo.experiments.stash)
         repo.experiments.stash.clear()
+    if workspace:
+        ref_infos = list(exp_refs(repo.scm))
+        remove_exp_refs(repo.scm, ref_infos)
```
pmrowla, line 25: This will remove all experiments in the user's entire git repository. `exp remove --workspace` should only remove experiments that are based on the user's current `HEAD` (and not experiments based on other git commits)

https://github.com/iterative/dvc/issues/5676#issuecomment-858582527

so we can't use the general `exp_refs()` iterator here, it should be:
```
        head = repo.scm.get_ref()
        exp_refs_by_baseline(head)
```

pmrowla (changes requested):

dvc/repo/experiments/remove.py
dberenbaum: I think it's fine to leave the behavior as is and rename the flag to `--all`. That fits the requirement here and is easier to explain.

pmrowla: @karajan1001 can you please rename the flag to `--all`? If it is left as `--workspace` it will be confusing, since all other DVC commands use `--workspace` to mean `HEAD` + whatever is in the actual working directory

karajan1001: @pmrowla , @dberenbaum 
I think the feature `--all` is not as useful as `--workspace`?  Can add it sometime later?

karajan1001:
[asciicast](https://asciinema.org/a/6S1SQw8wG6JHwUt5Cp09mT3Bk)

dvc/repo/experiments/remove.py
dberenbaum: I think which is more useful depends. Most of the time, I would guess users won't notice a different between them. If they have experiments in other commits, they might want to keep those, but I think it's just as likely that they have forgotten about them or don't know they still exist and would be happy to drop them.

After the discussion in this ticket, I prefer `--all` because:

1. It's easier to explain and understand. It's not obvious what is a "workspace" experiment, and users might want to confirm that these are the experiments they want to remove.
2. It parallels `dvc exp list`. If we eventually want to do something like `dvc exp remove --remote`, having this parallel `dvc exp list` seems useful since it's easy to see what experiments will be deleted.
3. Similarly, in the future we could add the `--rev` option like in `dvc exp list`. This is probably more useful than the `--workspace/--all-branches/--all-tags/--all-commits` flags since `--rev` is more granular (and `--rev HEAD` can do the same as `--workspace`), and I don't see `--all-branches` or `--all-tags` ever being useful in `dvc exp remove`.

dvc/repo/experiments/remove.py
```diff
@@ -7,21 +7,37 @@
...
 @locked
 @scm_context
-def remove(repo, exp_names=None, queue=False, **kwargs):
-    if not exp_names and not queue:
+def remove(
+    repo,
+    exp_names=None,
+    queue=False,
+    workspace=False,
+    all_baseline=False,
```
pmrowla, line 27: Can we just name this `all_`? (and change `clear_all_baseline` to `clear_all`)

dvc/repo/experiments/remove.py
```diff
@@ -33,6 +49,33 @@ def remove(repo, exp_names=None, queue=False, **kwargs):
     return removed
 
 
+def _clear_workspace_stash(repo):
```
pmrowla, line 52:
```suggestion
def _clear_stash(repo):
```

This clears the entire stash, it is not specific to stashed experiments which only apply to the current workspace/HEAD

pmrowla (changes requested):

dvc/command/experiments.py
```diff
@@ -1237,9 +1240,20 @@ def add_parser(subparsers, parent_parser):
...
+    remove_group.add_argument(
         "--queue", action="store_true", help="Remove all queued experiments."
     )
+    remove_group.add_argument(
+        "--workspace",
+        action="store_true",
+        help="Remove all experiments based on current workspace.",
+    )
+    remove_group.add_argument(
+        "--all",
```
pmrowla, line 1253: we can add `-A` short option here to be consistent with other commands

pmrowla:
Also, don't forget that this will need a docs PR

dvc/repo/experiments/remove.py
```diff
@@ -7,21 +7,37 @@
...
 
     removed = 0
     if queue:
-        removed += len(repo.experiments.stash)
-        repo.experiments.stash.clear()
+        removed += _clear_stash(repo)
+    if workspace:
+        removed += _clear_workspace(repo)
+    if clear_all:
+        removed += _clear_all_baseline(repo)
```
pmrowla, line 39:
```suggestion
        removed += _clear_all(repo)
```

dvc/repo/experiments/remove.py
```diff
@@ -33,6 +49,33 @@ def remove(repo, exp_names=None, queue=False, **kwargs):
...
+    stashed = 0
+    stash_revs = repo.experiments.stash_revs
+    for _, ref_info in stash_revs.items():
+        if head == ref_info.baseline_rev:
+            repo.experiments.stash.drop(ref_info.index)
+            stashed += 1
+    return stashed + len(ref_infos)
+
+
+def _clear_all_baseline(repo):
```
pmrowla, line 71:
```suggestion
def _clear_all(repo):
```

dvc/repo/experiments/remove.py
```diff
@@ -33,6 +49,33 @@ def remove(repo, exp_names=None, queue=False, **kwargs):
...
+    ref_infos = list(exp_refs(repo.scm))
+    remove_exp_refs(repo.scm, ref_infos)
+    removed = len(repo.experiments.stash)
+    repo.experiments.stash.clear()
```
pmrowla, line 75: I'm not actually sure if `-A/--all` should affect queued experiments, since we have a separate option for that (`--queue`). It seems to mie that `-A` should only affect "full" experiments that have already been reproduced

maybe a question for @dberenbaum

dberenbaum: 🤔 It seems simpler to explain if it includes the queue, but it's probably more useful if it doesn't since we already have an option to remove the whole queue. Going back to the discussion above, `dvc exp list` ignores queued experiments, and I think it makes sense to do the same here.

karajan1001: @dberenbaum , Same with `-w/--workspace`, do not affect queued ones? 

table | workspace based | all based
------ | ------ | -------------
committed | 1 | 2
queued | 3  | 4

currently `--queue` removes `3` and `4`. After next change `--all` removes `1` and `2`, and workspace? removes `1` and `3` or only `1`

dberenbaum: Sorry, I thought we were getting rid of `--workspace`?

karajan1001: I keep both of them for now.

dberenbaum: I would recommend keeping one, and I prefer `--all` so we don't have to get into the confusing corner cases like the one you mentioned above. I think when you document the features, you will find it easier if you don't have to document `--workspace` 😄 .

As mentioned in the other discussion above, I think you can add `--rev` later if functionality similar to `--workspace` is needed.

dvc/repo/experiments/remove.py
```diff
@@ -53,9 +73,9 @@ def _get_exp_ref(repo, exp_name: str) -> Optional[ExpRefInfo]:
         if repo.scm.get_ref(exp_name):
             return ExpRefInfo.from_ref(exp_name)
     else:
-        exp_refs = list(exp_refs_by_name(repo.scm, exp_name))
-        if exp_refs:
-            return _get_ref(exp_refs, exp_name, cur_rev)
+        exp_ref_list = list(exp_refs_by_name(repo.scm, exp_name))
```
karajan1001, line 76: duplicated name with the function `exp_refs`

pmrowla (approved):
LGTM on the engineering side, but should hold off on merging until @dberenbaum double checks the final usage

dberenbaum (approved):
Looks great! Sorry for the lack of clarity on the syntax. Hopefully keeping it simple for now will be helpful, especially when we start expanding usage like in #6006.

PR #6400, opened 2021-08-09
Add auto-refresh option to html

daavoo:

Closes https://github.com/iterative/dvclive/issues/147

This P.R. enables the option to add an auto-refresh tag to the default HTML plot template. The option is not exposed to the public API (commands).

The main use case is DVCLive, as the HTML gets updated at each `step`. Not sure if this would be the best way to implement this feature.


karajan1001 (approved):
LGTM. but not familiar with html.

dvc/repo/live.py
```diff
@@ -17,7 +17,7 @@ def create_summary(out):
 
     html_path = out.path_info.with_suffix(".html")
 
-    out.repo.plots.write_html(html_path, plots, metrics)
+    out.repo.plots.write_html(html_path, plots, metrics, refresh_seconds=5)
```
dberenbaum, line 20: Are you thinking this would be configurable in dvclive?

daavoo: It could be, but I just implemented it as an optional arg because I wasn't sure about what would be a reasonable value to hardcode. If we decide to hardcode it, the introduced change might be simpler but we would still need to be able to deactivate the refresh.

pared (approved):
LGTM, do we have an issue for that?

daavoo:
> LGTM, do we have an issue for that?

Not really.

pared:
On thing though: what in the use case, where user has his own html template, and there is no refresh tag inside? Seems like it will fail now, won't it?

pared (changes requested):
As in comment.

daavoo:
> On thing though: what in the use case, where user has his own html template, and there is no refresh tag inside? Seems like it will fail now, won't it?

That it's True. I will fix that.

By the way does it makes sense to expose the refreshing option for `dvc plots`?

daavoo:
> > On thing though: what in the use case, where user has his own html template, and there is no refresh tag inside? Seems like it will fail now, won't it?
> 
> That it's True. I will fix that.
> 
> By the way does it makes sense to expose the refreshing option for `dvc plots`?

It looks like it doesn't fail. python str.format just silently ignores the non existing keys. 

It is covered by this test case:

https://github.com/iterative/dvc/blob/415c212453618a75f9a4ea3d5064c068090e72f7/tests/func/utils/test_html.py#L28

dvc/repo/live.py
daavoo: Both points make sense. I guess the discussion would be yet again how to expose this DVCLive options to DVC commands without poluting the args.

For point 1, feels like the kind of customization that you would currently apply with `dvc plots modify` or manually editing the `dvc.yaml`.

> revert back to no refresh before exiting

Sounds like a good idea but I'm not sure how we could implement that.

dberenbaum: Sounds good. Let's either open a separate issue or wait and see if there's a need.

Issue #5546, opened 2021-03-04
DVC import file hashing only runs on one CPU thread


raimannma:
# Bug Report

## Description

I want to import data from one repo into another, but hashing takes forever for me. As ruslan pointed out on [discord](https://discord.com/channels/485586884165107732/563406153334128681/817004690197577748), this is caused due to single thread only hashing on DVC imports.

For me this is a big problem, as pulling data takes more than an hour only for computing the hashes.

### Reproduce

1. Create 2 hosted repos (REPO_1, REPO_2)
2. Add large dataset to REPO_1
3. In REPO_2 `dvc import` the data from REPO_2 

### Expected

I expect the same performance as normal dvc file hashing on all my cpu threads, which should take for my 60k image dataset under 5min

### Environment information

- Ubuntu 20.04
- Self-hosted GitLab
- S3 Remote
- Dataset with ~60k images

**Output of `dvc version`:**

```console
$ dvc version
DVC version: 2.0.1 (pip)
---------------------------------
Platform: Python 3.8.5 on Linux-5.8.0-44-generic-x86_64-with-glibc2.29
Supports: All remotes
Cache types: hardlink, symlink
Cache directory: ext4 on /dev/nvme0n1p5
Caches: local
Remotes: s3
Workspace directory: ext4 on /dev/nvme0n1p5
Repo: dvc, git
```

efiop:
Discord context: https://discord.com/channels/485586884165107732/563406153334128681/817013134782890004

BeatWolf:
I'm also affected by this bug, and it sadly doesn't make dvc import useable right now (for datasets with many files)

BeatWolf:
Here is the cprofile of a dvc update. I took maybe 24 hours. There is one "issue" with it. It updated 2 folders, but as it took about 10 hours for the first, it asked for authentication again (for the git repository) so it must have been stuck for many hours. I hope that doesn't cause a problem with the profiling.
Here is the profile, i hope it helps! (as this is a huge issue for my use-case with DVC) https://drive.switch.ch/index.php/s/IKN2M5cRCOe3hwf

zimka:
Is there any news regarding this issue?
It looks like the speed difference between `dvc pull` and `dvc import` is not caused only by threading problem (in code ThreadPoolExecutor is used in both cases).
I have made some tests, and even on single core machine with --jobs=1 `dvc pull` worked for me several times faster than `dvc import` on `data/cat_dogs' dataset from dvc-bench

[...]

PR #6401, opened 2021-08-09
repofs: use underlying fs.download to download files

### LINKED ISSUE #6019: get: load performance is significantly (many times) lower than for pull (reproduced on dvc-bench dataset)
# Bug Report
`get` command for external repo works many times slower than clone + `pull` command for the same repo (checked on local remote). The same issue happens for the `import` command. 
For reproducible toy dataset (1Gb, 10^4 files) there is a 5x difference, measured by cProfile (archive attached). 
For my real data (which actually raised the issue in the first place) with 20Gb and 150k files the downgrade is huge: `pull` takes 30min, while `get` estimation was 350+hours!

## Reproduce
I have created a test [setup](https://github.com/zimka/example-get-started) to create toy dataset with random binary files and measure the difference on them, but I think it can be reproduced from scratch on any data. 
1. Create dataset, it should contain enough data and files (1Gb and 10^4 file in my case)
2. Set `local` remote, `push` data there, commit and push your repo on github
3. Clone your repo, use `pull` to load data (measure)
4. Use `get` to load data (measure)

I measured both ways with cProfile and `get` is 5x slower than `pull`.
[cProfiles2.zip](https://github.com/iterative/dvc/files/6489530/cProfiles2.zip)

The problem is, that it is probably not a constant ratio, but depends on the amount of files or data. For my real data with 160k files and 20Gb data, the `get` after 8 hours of exectuion(which by its own 10x longer) the ETA was 350+ hours, while `pull` took only 30min.
In my view, even if it is expected for `get` to work slow, such decrease is too bad to be bearable.

I have also noticed that CPU is used intensively during the `pull` and is not during the `get` execution. I tried to increase number of jobs for `get` in several times, but it has not affected the ETA anyhow.

### Environment information
**Output of `dvc doctor`:**

```console
$ dvc doctor
DVC version: 2.1.0 (pip)
Platform: Python 3.7.6 on Windows-10-10.0.19041-SP0
Supports: http, https
Cache types: hardlink
Cache directory: NTFS on C:\
Caches: local
Remotes: local
Workspace directory: NTFS on C:\
Repo: dvc, git
```

efiop:
This makes `dvc get` work just as fast as `import/pull`. The reason is that `open()` is harder to make fast (e.g. we need to know what block size to use, etc, for which we don't yet have a very good internal infrastructure), and it is much easier to just use native `fs.download` method that has been already tuned for maximum performance.

Related to #5546
Fixes #6019

### DIFF
--- a/dvc/fs/dvc.py
+++ b/dvc/fs/dvc.py
@@ -62,9 +62,7 @@ class DvcFileSystem(BaseFileSystem):  # pylint:disable=abstract-method
             return obj.hash_info
         raise FileNotFoundError
 
-    def open(  # type: ignore
-        self, path: PathInfo, mode="r", encoding=None, remote=None, **kwargs
-    ):  # pylint: disable=arguments-differ
+    def _get_fs_path(self, path: PathInfo, remote=None):
         try:
             outs = self._find_outs(path, strict=False)
         except OutputNotFoundError as exc:
@@ -92,16 +90,20 @@ class DvcFileSystem(BaseFileSystem):  # pylint:disable=abstract-method
             else:
                 checksum = out.hash_info.value
             remote_info = remote_odb.hash_to_path_info(checksum)
-            return remote_odb.fs.open(
-                remote_info, mode=mode, encoding=encoding
-            )
+            return remote_odb.fs, remote_info
 
         if out.is_dir_checksum:
             checksum = self._get_granular_hash(path, out).value
             cache_path = out.odb.hash_to_path_info(checksum).url
         else:
             cache_path = out.cache_path
-        return open(cache_path, mode=mode, encoding=encoding)
+        return out.odb.fs, cache_path
+
+    def open(  # type: ignore
+        self, path: PathInfo, mode="r", encoding=None, **kwargs
+    ):  # pylint: disable=arguments-renamed
+        fs, fspath = self._get_fs_path(path, **kwargs)
+        return fs.open(fspath, mode=mode, encoding=encoding)
 
     def exists(self, path):  # pylint: disable=arguments-renamed
         try:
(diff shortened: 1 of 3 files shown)


dvc/fs/dvc.py
```diff
@@ -253,3 +255,9 @@ def info(self, path_info):
                 ret[obj.hash_info.name] = obj.hash_info.value
 
         return ret
+
+    def _download(self, from_info, to_file, **kwargs):
```
efiop, line 259: For the record: this is a bit ugly, but it will turn into a cleaner and more proper `fs.get_file` when migrating dvcfs to fsspec.

dvc/fs/dvc.py
```diff
@@ -62,9 +62,7 @@ def _get_granular_hash(
             return obj.hash_info
         raise FileNotFoundError
 
-    def open(  # type: ignore
-        self, path: PathInfo, mode="r", encoding=None, remote=None, **kwargs
-    ):  # pylint: disable=arguments-differ
+    def _get_fs_path(self, path: PathInfo, remote=None):
```
efiop, line 65: This didn't handle chained imports before and neither does it handle them now. We'll need to generalize resolving logic we use in dvc/dependency/repo.py to use it here. Will prob look into it in a followup.

dvc/fs/dvc.py
```diff
@@ -253,3 +255,9 @@ def info(self, path_info):
                 ret[obj.hash_info.name] = obj.hash_info.value
 
         return ret
+
+    def _download(self, from_info, to_file, **kwargs):
+        fs, path = self._get_fs_path(from_info)
+        fs._download(  # pylint: disable=protected-access
```
efiop, line 261: This is again temporary, it will be replaced by just `get_file` in the future.

isidentical (approved):
LGTM

PR #6419, opened 2021-08-12
Allow user to specify different state and index tmp dir locations

nueschs:
The pull request [6111](https://github.com/iterative/dvc/pull/6111) intends fix issue #4420. However, it has fallen a bit behind master. This PR reincorporates the changes made by @agurtovoy into the current master.

  https://github.com/iterative/dvc.org/pull/2776


efiop:
@mergifyio rebase

efiop:
Looks like there are some windows failures :slightly_frowning_face:

meierale:
> Looks like there are some windows failures 🙁

In this PR there are some timeout problems in the tests which we could not figure out why they occur. (locally tests run fine)... in the original PR 6111 there were some windows errors however.

meierale:
Documentation is now added in https://github.com/iterative/dvc.org/pull/2776

meierale:
Looks like the tests for this PR (which did not use the `make_tmp_dir` fixture at first) caused other tests to fail... I don't quite understand why... but since I adjusted the tests the checks run properly.

efiop:
@meierale Thank you so much for updating the PR! 🙏

We've had some new requirements for state recently, which made me think about the path generation in this PR (one of the reasons for slow reviews and us not overtaking earlier). One thing that worries me here is that when setting `state.dir` config, users will likely expect it to use that dir directly (e.g. so that you could share state between two repos, which is useful for optimization), so it will be a source of confusion. I think it makes sense to allow for that scenario of sharing state(index too), but we need to cleanup a few things first (e.g. we are saving relpaths for links https://github.com/iterative/dvc/blob/e5cf9b203df01f13e5568c745906d74bc131235c/dvc/state.py#L151 , which clearly cannot be shared). Plus we have a few more requirements to support other fs types (e.g. s3) in the future.

This means that we cannot remake this PR into a pre-requisite for ^, because it will be fundamentally broken if anyone will try to use the same state.dir setting for multiple repos 🙁 We could just merge it and then reimplement it, but that will cause a problem in docs as I've mentioned before. But I guess the latter is probably the best way to handle it for now after all. Ideally we would merge and then do our adjustments right away before the next release (could do that in this PR, but that would be way out of scope).

So let's do this: we'll do one more release without this PR and will then merge and introduce our changes on top, to hopefully get the benefit of both changes in the release after.

efiop:
Thanks for the PR @nueschs ! 🙏 

For the record: we need to make our state work with any filesystems (e.g. s3) for multiple reasons these days (e.g. some incoming projects will need to use it), so we'll likely just put all state databases in one directory, that will then have a subdir for each filesystem (identified by the hash of its config), so that main directory will obey `state.dir` config option and will be sharable between different dvc repo instances.

Issue #6164, opened 2021-06-14
cls._from_parts(args, init=False) is broken on 3.10


isidentical:
Both the init parameter, and the `_init` method have been removed in the 3.10, see python/cpython/pull/19342. 
https://github.com/iterative/dvc/blob/19f2fe52e6d7615aa28388cf1acdea76a8a0345a/tests/dir_helpers.py#L84

PR #6422, opened 2021-08-13
add support for Python 3.10

skshetry:
`pyarrow` is not available, so we are skipping tests for it.
And, restricting `hdfs` extras to be not installed in Python3.10.

Pathlib removed `_init` method and `init` parameter, so just working
around that. It's not a big change, even compared to 3.9.

`pygit2` and `pyarrow` does not have wheels, `pywin32` is not available for Windows at this point (which is a critical dependency in the Windows for us).
If you can install `libgit2-1.1` and are on macOS/Linux, there should not be any issues on 3.10.
Ubuntu < 21.04 does not provide libgit2-1.1 which pygit2 requires.

In the CI, to be sure it works well with Python3.10, we run `macOS` with libgit2 installed from `brew`.

--- 
Not adding a trove classifier for Python3.10 support, as we should:
1. Have wheels for `pyarrow`/`pywin32`/`pygit2`.
2. Pass all tests in CI for all OSes.

When adding proper support, I should remember to check following in the future.:
1. Remove skips in the `hdfs` fixtures.
2. Unskip `pywin32` for Windows + Python >= 3.10.
3. Unskip `pyarrow` for Python >= 3.10.
4. Run tests on all OSes.
5. Also check for additional compatibility for `dulwich`/`ruamel.yaml`.

Closes https://github.com/iterative/dvc/issues/6164.


.github/workflows/tests.yaml
```diff
@@ -68,10 +69,14 @@ jobs:
         key: ${{ runner.os }}-pip-${{ hashFiles('setup.py') }}
         restore-keys: |
           ${{ runner.os }}-pip-
+    - name: install libgit2 on Python3.10 in macOS
+      if: matrix.pyv == '3.10.0-rc.1' && matrix.os == 'macos-10.15'
+      run: |
+        brew install libgit2
```
skshetry, line 75: Python3.10 only running in `macOS` as libgit2-1.1 is not available in Ubuntu (only >= 21.04 which is not supported in GHA).

PR #6426, opened 2021-08-15
du: add du command to summarize file sizes

### LINKED ISSUE #4682: Get dataset size from DVC data registry
Hi, I would like to request a feature to query dataset sizes from a DVC registry.

I have created a data registry using the documentation and added a dataset to my remote storage through the simple `dvc add my_dataset` and `dvc push` commands. 
Now, before I import the dataset from a registry I want to make sure I have enough disk space. So, something like the following would be really useful. 

```bash
dvc du -hs https://github.com/my_data_registry my_dataset
```
-  `du`: which stands for Disk Usage is a standard Unix command to query file sizes.
      options `h` and `s` represent the following:
       -  `h`: is a human-readable format so in bytes, megabytes, gigabytes etc
       -  `s`: is to sum the sizes of all files in a directory.

If the DVC team thinks there are better ways to do this without adding a new feature I'm all ears.

jonburdo:
Fixes #4682

@shcheklein 
@efiop


efiop:
For the record: opened by mistake.

PR #6431, opened 2021-08-16
[WIP] plots: introduce support for images

pared:


dberenbaum:
Hi @pared, what's the status on this PR? Is there anything we can do to move it along while you are out?

dvc/repo/collect.py
```diff
@@ -49,7 +49,6 @@ def _collect_paths(
                 )
             else:
                 logger.warning("'%s' was not found at: '%s'.", path_info, rev)
-            continue
```
pared, line 52: Removing - even if we want to have a warning that some file does not exist, we need to try to open that file later, so that we get `FileNotFound` in results, for API consistiency.

pared:
@dberenbaum I believe it should be fine as-is for now. There could be some polishing in the future, but it should be fine for now.

pared:
cc @Suor @rogermparent @mattseddon

pared:
Sample image plots rendering:
```
#!/bin/bash

set -ex

src=$(pwd)

pushd $TMPDIR

rm -rf test_workspace
mkdir test_workspace

pushd test_workspace
mkdir test_repo

pushd test_repo

git init --quiet
dvc init --quiet

echo -e "
from tensorflow.keras.callbacks import Callback
from tensorflow.keras.datasets import mnist
from tensorflow.keras.models import Sequential
from tensorflow.keras.layers import Dense, Activation
from tensorflow.keras import utils
from dvclive.keras import DvcLiveCallback

results = defaultdict(list) 

class MetricsCallback(Callback):
    def on_epoch_end(self, epoch: int, logs: dict = None):
        logs = logs or {}
        for metric, value in logs.items():
            results[metric].append(value)
        results["epoch"].append(epoch)


def load_data():
    (x_train, y_train), (x_test, y_test) = mnist.load_data()

    x_train = x_train.reshape(60000, 784)
    x_test = x_test.reshape(10000, 784)
    x_train = x_train.astype('float32')
    x_test = x_test.astype('float32')
    x_train /= 255
    x_test /= 255

    classes = 10
    y_train = utils.to_categorical(y_train, classes)
    y_test = utils.to_categorical(y_test, classes)
    return (x_train, y_train), (x_test, y_test)


def get_model():
    model = Sequential()

    model.add(Dense(256, input_dim=784))
    model.add(Activation('sigmoid'))
    model.add(Dense(10, input_dim=256))
    model.add(Activation('softmax'))

    model.compile(loss='categorical_crossentropy',
    metrics=['accuracy'], optimizer='sgd')
    return model


(x_train, y_train), (x_test, y_test) = load_data()
model = get_model()

model.fit(x_train,
          y_train,
          validation_data=(x_test, y_test),
          batch_size=128,
          epochs=5,
          callbacks=[MetricsCallback(), DvcLiveCallback()])

def write_results(results, d):
    import matplotlib
    import matplotlib.pyplot as plt
    import os
    os.makedirs(d)
    x = results.pop("epoch")
    for key, y in results.items():
        plt.figure()
        plt.plot(x, y)
        plt.ylabel(key)
        plt.savefig(os.path.join(d, key + ".png"))
    

write_results(results, "plots")
" >> train.py
dvc run -d train.py --plots plots --plots dvclive -n train python train.py
git add -A
git commit -am "init"

sed -i "" "s/sgd/adam/g" train.py
dvc repro train
dvc plots diff
```

dvc/repo/live.py
```diff
@@ -15,10 +17,10 @@ def create_summary(out):
 
     metrics, plots = out.repo.live.show(str(out.path_info))
 
-    html_path = out.path_info.with_suffix(".html")
+    html_path = out.path_info.fspath + "_dvc_plots"
```
daavoo, line 20: Is there a reason for this change? 

Should we update on DVCLive that `{path}.html` is no longer the default html output?

dberenbaum: Good catch! To support images, the image files must be stored alongside the html, so the output is now a directory instead of a single file.

I would probably opt to do something like `dvcvlive/dvc_plots` rather than append them with underscores. Any other ideas are welcome.

daavoo: I have no strong preference but if we go with `dvcvlive/dvc_plots`, the `.html` output will now be handled like the `.tsv` plots (i.e. based on `cache` option of the output). 

Previously, the `.html` was not tracked by DVC and it was up to the user whether to track it with Git/DVC or just ignore.

dberenbaum: Ah, that's probably why @pared did it this way. Maybe just `dvclive_plots` then?

Another option would be to reorganize dvclive outputs so that the tsv files are under a subdirectory like `dvclive/logs` or `dvclive/tsv` and the html is another subdirectory like `dvclive/plots` or `dvclive/html`.

daavoo: I kind of like `dvclive/tsv` & `dvclive/html`. 

Honestly, the word `plots` for referring to the HTML sounds kind of confusing to me because the `.tsv` files are also `dvc plots` in a sense.

Moving the outputs to subdirectories would require some changes in DVC core but it feels like a good direction. I might consider also moving the summary `.json`  to `dvclive/summary.json`.

daavoo: Maybe we can continue the discussion in https://github.com/iterative/dvclive/issues/148 as this is a comment in a closed P.R.

dvc/render/vega.py
```diff
@@ -0,0 +1,175 @@
...
+            "'{}' - file type error\n"
+            "Only JSON, YAML, CSV and TSV formats are supported.".format(file)
+        )
+
+
+class PlotDataStructureError(DvcException):
+    def __init__(self):
+        super().__init__(
+            "Plot data extraction failed. Please see "
+            "https://man.dvc.org/plot for supported data formats."
```
jorgeorpinel, line 31: Should be `/plots`

dvc/render/image.py
```diff
@@ -0,0 +1,64 @@
...
+        img_path = os.path.join(
+            static, f"{revision}_{filename.replace(os.sep, '_')}"
+        )
+        rel_img_path = relpath(img_path, page_dir_path)
+        with open(img_path, "wb") as fd:
+            fd.write(image_data)
+        return """
+        <div>
+            <p>{title}</p>
+            <img src="{src}">
```
jorgeorpinel, line 39: I think you can probably inject code here:

```python
>>> os.path.relpath("</img> <script>alert('oops');</script>")
"</img> <script>alert('oops');</script>"
```

PR #6432, opened 2021-08-16
fs: find(..., detail=True) now casts values properly

isidentical:
In the following statement we try to access the first index, which is not possible when using `dict.values()`.


isidentical:
pre-requisite; https://github.com/iterative/PyDrive2/pull/128

efiop (approved):

PR #6448, opened 2021-08-16
analytics: add DVC_ANALYTICS env var

efiop:
New special env var to force turn on/off analytics. If not defined - keeps old behavior. Needed for our vs-code plugin to force-disable analytics for them (we don't really care about that analytics and vs-code has its own analytics anyway).

Docs: https://github.com/iterative/dvc.org/pull/2724


dvc/env.py
```diff
@@ -9,3 +9,4 @@
 DVC_IGNORE_ISATTY = "DVC_IGNORE_ISATTY"
 DVC_EXP_GIT_REMOTE = "DVC_EXP_GIT_REMOTE"
 DVC_EXP_AUTO_PUSH = "DVC_EXP_AUTO_PUSH"
+DVC_ANALYTICS = "DVC_ANALYTICS"
```
skshetry, line 12: This should be called `DVC_NO_ANALYTICS` as it's sole purpose is to disable analytics.

See also: https://keepsake.ai/docs/learn/analytics

efiop: @skshetry It is also meant to force-enable analytics, so `NO_ANALYTICS` would get confusing.

skshetry: Why does the user need to force-enable it? Feels unnecessary.

efiop: For the record: continuing in https://github.com/iterative/dvc.org/pull/2724

Issue #6449, opened 2021-08-17
console briefly opens on Windows


skshetry:
At the end, when DVC tries to run analytics, a console briefly opens and closes. On older computers, it stays a bit longer.

Ideally, this should not be the case at all.

https://user-images.githubusercontent.com/18718008/129657817-f6a0b0c0-f8d7-450d-bec8-4e00827f499e.mp4

PR #6450, opened 2021-08-17
fix console appearing on daemon in Windows

skshetry:
Trying to fix https://github.com/iterative/dvc/issues/6449.

Would be great to have confirmation from someone using Windows.


dvc/daemon.py
```diff
@@ -12,29 +12,34 @@
...
+    # https://bugs.python.org/issue41619
+    creationflags = CREATE_NEW_PROCESS_GROUP | CREATE_NO_WINDOW
 
     startupinfo = STARTUPINFO()
     startupinfo.dwFlags |= STARTF_USESHOWWINDOW
 
-    _popen(
-        cmd, env=env, creationflags=creationflags, startupinfo=startupinfo
-    ).communicate()
+    _popen(cmd, env=env, creationflags=creationflags, startupinfo=startupinfo)
```
skshetry, line 46: We just need to spawn it, I don't think we need to wait for it.

shcheklein:
cc @mattseddon

dvc/daemon.py
```diff
@@ -12,29 +12,38 @@
...
     prefix = [sys.executable]
     if not is_binary():
-        prefix += [sys.argv[0]]
+        dvc_exec = sys.argv[0]
+        if os.name == "nt":
+            # python -m /path/to/dvc does not work?
+            # as /path/to/dvc is a path to .exe and itself is executable
+            prefix = [dvc_exec]
+        else:
+            prefix += [dvc_exec]
```
skshetry, line 27: Need to see if there's a better way to find the current executable.

karajan1001:
Sorry, could not see the console on my computer. Maybe because it is too fast?  
I'm using an AMD 3600 CPU and 32GB memory.

skshetry:
Did you check on master branch?

karajan1001:
> Did you check on master branch?

Yes. I'm on 2.6.0+a285d2

skshetry:
Do you have analytics enabled?

karajan1001:
> analytics

`dvc config core.analytics true`? I tried both `true` and `false`

skshetry:
@Mergifyio rebase

dvc/daemon.py
```diff
@@ -12,29 +12,34 @@
...
 
 def _popen(cmd, **kwargs):
     prefix = [sys.executable]
     if not is_binary():
-        prefix += [sys.argv[0]]
-
+        main_entrypoint = os.path.join(
+            os.path.abspath(os.path.dirname(__file__)), "__main__.py"
+        )
+        prefix += [main_entrypoint]
```
skshetry, line 24: @iterative/dvc, do you know a better way to get the executable? In Windows, the `sys.argv[0]` is `.env/Scripts/dvc` which is just a link to `.env/Scripts/dvc.exe` build from the `console_scripts`.  So, we were calling it as `/path/to/python .env/Scripts/dvc` which is not a python file and the daemon was failing on Windows.  

In Linux/MacOS, based on how you invoke, it's either `__main__.py` or the `.env/bin/dvc`.

efiop: @skshetry It is also important to not forget about our exe binary built by pyinstaller, IIRC `__main__.py` is not present there, but I might be mistaken.

skshetry: @efiop, It's inside the `if not is_binary():` block as before.

efiop: @skshetry Ah, sorry, totally blinded that one. :facepalm:

efiop (approved):
LGTM

PR #6453, opened 2021-08-17
analytics: introduce DVC_NO_ANALYTICS

efiop:
Replacing DVC_ANALYTICS. This is following our other env conventions and
will not have effect if set to false/no/etc, and will only work if set
to true/yes/y/etc. Will disable analytics even if it is explicitly
enabled in our configs.

Per discussion in https://github.com/iterative/dvc.org/pull/2724 . Kudos @skshetry


pared (approved):

dvc/analytics.py
```diff
@@ -44,8 +44,8 @@ def is_enabled():
     if env2bool("DVC_TEST"):
         return False
 
-    enabled = env2bool(DVC_ANALYTICS, None)
-    if enabled is None:
+    enabled = not env2bool(DVC_NO_ANALYTICS)
```
skshetry, line 47:
```suggestion
    enabled = not os.getenv("DVC_NO_ANALYTICS")
```
Usually, in these kind of `NO_*` environment variables, the value does not matter. You can check it with `HOMEBREW_NO_ANALYTICS=<value> brew analytics`.

Keepsake seems to be doing same: https://github.com/replicate/keepsake/blob/a8032b4e4387a3f445ad92b83a20275b4b4ec4a0/go/pkg/analytics/track.go#L24

See also: [NO_COLOR](https://no-color.org/).
> All command-line software which outputs text with ANSI color added should check for the presence of a NO_COLOR environment variable that, when present (**regardless of its value**), prevents the addition of ANSI color.

efiop: @skshetry You are right, we currently have a mixup of env2bool and getenv in dvc. So indeed to reason to stick to env2bool for convention. Adjusting.

skshetry (approved):

PR #6458, opened 2021-08-18
exp run: pass recursive args to checkout

skshetry:
Closes #6450.

### DIFF
--- a/dvc/repo/experiments/executor/base.py
+++ b/dvc/repo/experiments/executor/base.py
@@ -348,6 +348,7 @@ class BaseExecutor(ABC):
                     quiet=True,
                     allow_missing=True,
                     checkpoint_reset=checkpoint_reset,
+                    recursive=kwargs.get("recursive", False),
                 )
 
             checkpoint_func = partial(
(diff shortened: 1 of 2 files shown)


dvc/repo/experiments/executor/base.py
```diff
@@ -348,6 +348,7 @@ def filter_pipeline(stages):
                     quiet=True,
                     allow_missing=True,
                     checkpoint_reset=checkpoint_reset,
+                    recursive=kwargs.get("recursive", False),
```
skshetry, line 351: @pmrowla, Is it not possible to checkout as we go `repro`-ing? This way, it's may miss a lots of things. `foreach` based stages come to mind, as only `repro` supports them.

pmrowla: Yeah we could do that. I think we'd just need to pass a flag into repro (and subsequently `stage.run`) that indicates we need to `checkout(allow_missing=True)` dependencies prior to reproducing a stage, right?

pmrowla: The one thing I'm not sure on is how this affects the run cache lookup - do deps need to be checked out before that lookup or can we just skip the dep checkout entirely in the event we have a cached stage?

pmrowla (approved):
LGTM, it's probably worth adding the glob issue to the next sprint to address this properly

dvc/repo/experiments/executor/base.py
skshetry: Run-cache lookup does get affected, as if there's no dependency, there's no cached stage. This will work for workspace based run, but we may need to map dependencies into the tmpdir repo.

PR #6425, opened 2021-08-14
pyinstaller: build: check dvc doctor output

efiop:
Fixes https://github.com/iterative/cml/issues/692

PR #6459, opened 2021-08-18
[WIP] Add microbenchmarks

skshetry:
Just a simple proposal for the µbenchmarks.

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -8,6 +8,7 @@ pytest-mock==3.6.1
 pytest-lazy-fixture==0.6.3
 pytest-timeout==1.4.2
 pytest-docker==0.10.3
+pytest-benchmark==3.4.1
 
 flaky==3.7.0
 mock==4.0.3


efiop:
@skshetry Let's put them under a subdir in `tests/benchmarks/`, please. E.g. `tests/benchmarks/micro`. We'll likely be bringing dvc-bench benchmarks here too, so want this to be compatible.

tests/benchmarks/test_startup.py
```diff
@@ -0,0 +1,6 @@
+from tests.shell import Command
+
+
+def test_startup(dvc_cmd: Command, benchmark):
+    result = benchmark(dvc_cmd.args("--help"))
+    assert result.returncode == 0
```
efiop, line 6: We have `test_help` in `dvc-bench`. It should be under `cli/commands`, since we need an ablility to test any binary with this, no matter if it is installed with pip or deb/rom/etc.

skshetry:
Should we try to keep benchmarks isolated from DVC's codebase? This way, we may be easy to compare in terms of history, but we may need to duplicate lots of things.

efiop:
@skshetry It is also clear that we need to even make our test suit more isolated from dvc :slightly_smiling_face: But from benchmark framework perspective, it doesn't matter if it is here or in `dvc-bench` as long as we take care isolating benchmark themselves from the rest of the tests (just having a tests/benchmarks subdir seems to be enough). If anything, it is actually easier to harness https://github.com/iterative/dvc-bench/issues/241 when it is in this repo (we can just default to always running benchmarks from `master`). So I'm definitely tempted to carry benchmarks over here and just use `dvc-bench` repo for storing results with html.

skshetry:
Another thing that I want to do is add `tests/e2e`, catching smaller issues like https://github.com/iterative/dvc/pull/6425 early, as well as larger ones.

But we should ask @dberenbaum what his thoughts are, either to use `pytest` or separate `.t` test files with bash scripts, as it is going to be testing product scenarios.

dberenbaum:
> But we should ask @dberenbaum what his thoughts are, either to use `pytest` or separate `.t` test files with bash scripts ([git-example](https://github.com/git/git/blob/master/t/t0012-help.sh), [hg-example](https://github.com/indygreg/hg/blob/bookmarks/%40/tests/test-clone.t)), as it is going to be testing product scenarios.

I'm not familiar with this `.t` test file format. Is the question whether to put these benchmarks into pytest vs bash scripts? Can you think of any product implications that this choice would have?

What are these micro benchmarks going to be about? the tests I see so far are checking return codes and output. I would guess that benchmarks are about tracking running time. Is that what they will be?

efiop:
> Another thing that I want to do is add tests/e2e, catching smaller issues like #6425 early, as well as larger ones.

@skshetry That is what https://github.com/iterative/dvc-test (it is dead though) was meant to test before. But we also care about e2e story performance, so maybe `commands/stories`(in dvc-bench) benchmarks could provide us with both checks and performance reports.

My problem with `.t` right now is that it will be a yet another framework and language (bash is simple when doing simple scripts, but for good tests it requires a certain proficiency), and so it will take some time to establish and get used to it. But I'm still open to it, if we have the need and there are significant advantages.

efiop:
@skshetry Not sure if this covers all microbenchmarks, but from the ones discussed here, it seems like they are just "internal" benchmarks, right? If so, we clearly also need to benchmark things like object `stage()`, tree diff-ing and so on, which are "internal benchmarks" kinda. So maybe we could group them? E.g. `tests/benchmarks/internal`?

tests/benchmarks/test_stage_list.py
```diff
@@ -0,0 +1,54 @@
...
+    to: two
+"""
+
+
+def test_stage_list(tmp_dir, dvc, dvc_cmd, benchmark):
+    (tmp_dir / "dvc.yaml").write_text(PARAMETRIZED_DVC_YAML)
+    (tmp_dir / "params.yaml").write_text(PARAMS_YAML)
+
+    stage_add = dvc_cmd.args("stage", "list")
+    result = benchmark(stage_add)
```
pared, line 43:
```suggestion
    stage_list = dvc_cmd.args("stage", "list")
    result = benchmark(stage_list)
```

pared (approved):

skshetry:
> @skshetry Not sure if this covers all microbenchmarks, but from the ones discussed here, it seems like they are just "internal" benchmarks, right? If so, we clearly also need to benchmark things like object `stage()`, tree diff-ing and so on, which are "internal benchmarks" kinda. So maybe we could group them? E.g. `tests/benchmarks/internal`?

I don't see any difference between `stage()` and other APIs, at least not in microbenchmarks. And, I don't think we should worry about that right now. Creating too many folder structures will only cause pain in the short term.

Issue #6263, opened 2021-07-01
executor: setup and configure executors


dberenbaum:
Meta-issue. Please link related tickets back to this one.

### User story

I want to spin up an executor and tear it down.

### Requirements

* Define Terraform resources in DVC config files
* Create, update, and destroy resources from within DVC

### Tasks
Config-only commands:

* [x] `dvc machine add` (#6462)
* [x] `dvc machine remove` (#6462)
* [x] `dvc machine list` - (#6480)
* [x] `dvc machine modify` - (#6480)

Terraform wrapper commands
* [x] `dvc machine create` (#6462)
* [x] `dvc machine destroy` (#6462)
* [x] `dvc machine rename` (#6481)
* [x] `dvc machine status` - (#6482)
* [x]  `dvc machine ssh` - (#6483)

dberenbaum:
Terraform resources should be configurable in `.dvc/config` files - including support for system, user, etc. configuration.

casperdcl:
related: https://github.com/iterative/terraform-provider-iterative/issues/164#issuecomment-901084421

dberenbaum:
I'm a little concerned that we are using up the namespaces for `machine`, especially `list` and `status`. We will need some command that outputs table-like info about each machine (similar to `docker ps`): what machines have been created, how long have they been running, are they idle or running an experiment, etc. Where will that go?

pmrowla:
I think anything that is aware of whether or not some instance is running experiments or not would go under `dvc exp`. `machine list/status` are more like `docker machine ls`, which just tells you about machines that have been created and/or booted (and the separation between `list` and `status` is that `list` shows .dvc/config entries, and `status` shows terraform status). `machine` is currently not aware of anything that is actually running on the instance itself.

`docker ps` displays the state of deployed docker containers that are running on some machine (that may or may not have been created with `docker machine`). This is more like the actual "executor" concept in DVC (i.e. keeping track of running/deployed experiments, which is separate from terraform/machine instance status). It seems like we will eventually just want a separate `machine ps` or `exp ps` command that would be the equivalent to `docker ps` and shows the state of running experiments - including whether or not a potential executor is idle.

pmrowla:
[...]

Issue #6440, opened 2021-08-16
Epic: remote executors


dberenbaum:
Parent issue for remote executor issues.

Linked issues:

- [x] #6263 
- [x] #6264 
- [ ] #6265 
- [ ] #6266 
- [ ] #6267 
- [ ] #6268 
- [ ] #6269 
- [ ] #6270

dberenbaum:
Need to rescope this feature

karajan1001:
Someone asking for running DVC in a lambda instance

https://discord.com/channels/485586884165107732/563406153334128681/1051517217478553640

PR #6462, opened 2021-08-19
[WIP] terraform executor proof-of-concept

pmrowla:

Related to https://github.com/iterative/dvc/issues/6440
May close: https://github.com/iterative/dvc/issues/6263

### DIFF
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -118,6 +118,7 @@ SCHEMA = {
         Optional("autostage", default=False): Bool,
         Optional("experiments"): Bool,  # obsoleted
         Optional("check_update", default=True): Bool,
+        "executor": Lower,
     },
     "cache": {
         "local": str,
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -224,10 +224,16 @@ SCHEMA = {
         "row_limit": All(Coerce(int), Range(1)),  # obsoleted
         "row_cleanup_quota": All(Coerce(int), Range(0, 100)),  # obsoleted
     },
+    "executor": {
+        str: {
+            "cloud": All(Lower, Choices("aws", "azure")),
+        },
+    },
     # section for experimental features
     "feature": {
+        Optional("executor", default=False): Bool,
         # enabled by default. It's of no use, kept for backward compatibility.
-        Optional("parametrization", default=True): Bool
+        Optional("parametrization", default=True): Bool,
     },
     "plots": {"html_template": str},
 }
--- a/setup.py
+++ b/setup.py
@@ -112,6 +112,8 @@ webdav = ["webdav4>=0.9.0"]
 ssh_gssapi = ["sshfs[gssapi]>=2021.7.1"]
 all_remotes = gs + s3 + azure + ssh + oss + gdrive + hdfs + webhdfs + webdav
 
+terraform = ["python-terraform>=0.14", "jinja2>=2.0.0"]
+
 tests_requirements = (
     Path("test_requirements.txt").read_text().strip().splitlines()
 )
@@ -138,6 +140,7 @@ setup(
         "hdfs": hdfs,
         "webhdfs": webhdfs,
         "webdav": webdav,
+        "terraform": terraform,
         "tests": tests_requirements,
     },
     keywords="data-science data-version-control machine-learning git"
--- a/dvc/cli.py
+++ b/dvc/cli.py
@@ -18,6 +18,7 @@ from .command import (
     data_sync,
     destroy,
     diff,
+    executor,
     experiments,
     freeze,
     gc,
@@ -87,6 +88,7 @@ COMMANDS = [
     experiments,
     check_ignore,
     live,
+    executor,
 ]
 
 
--- a/dvc/command/executor.py
+++ b/dvc/command/executor.py
@@ -1,0 +1,257 @@
+import argparse
+
+from dvc.command.base import CmdBase, append_doc_link, fix_subparsers
+from dvc.command.config import CmdConfig
+from dvc.config import ConfigError
+from dvc.ui import ui
+from dvc.utils import format_link
+
+
+class CmdExecutorConfig(CmdConfig):
+    def __init__(self, args):
+
+        super().__init__(args)
+        if not self.config["feature"].get("executor", False):
+            raise ConfigError("executor feature is disabled")
+
+        if getattr(self.args, "name", None):
+            self.args.name = self.args.name.lower()
+
+    def _check_exists(self, conf):
+        if self.args.name not in conf["executor"]:
+            raise ConfigError(f"executor '{self.args.name}' doesn't exist.")
+
+
+class CmdExecutorAdd(CmdExecutorConfig):
+    def run(self):
+        if self.args.default:
+            ui.write(f"Setting '{self.args.name}' as a default executor.")
+
+        with self.config.edit(self.args.level) as conf:
+            if self.args.name in conf["executor"] and not self.args.force:
+                raise ConfigError(
+                    "executor '{}' already exists. Use `-f|--force` to "
+                    "overwrite it.".format(self.args.name)
+                )
+
+            conf["executor"][self.args.name] = {"cloud": self.args.cloud}
+            if self.args.default:
+                conf["core"]["executor"] = self.args.name
+
+        return 0
+
+
+class CmdExecutorRemove(CmdExecutorConfig):
+    def run(self):
+        with self.config.edit(self.args.level) as conf:
+            self._check_exists(conf)
+            del conf["executor"][self.args.name]
+
+        up_to_level = self.args.level or "repo"
+        # Remove core.executor refs to this executor in any shadowing configs
+        for level in reversed(self.config.LEVELS):
+            with self.config.edit(level) as conf:
+                if conf["core"].get("executor") == self.args.name:
+                    del conf["core"]["executor"]
+
+            if level == up_to_level:
+                break
+
+        return 0
+
+
+class CmdExecutorModify(CmdExecutorConfig):
+    def run(self):
+        from dvc.config import merge
+
+        with self.config.edit(self.args.level) as conf:
+            merged = self.config.load_config_to_level(self.args.level)
+            merge(merged, conf)
+            self._check_exists(merged)
+
+            if self.args.name not in conf["executor"]:
+                conf["executor"][self.args.name] = {}
+            section = conf["executor"][self.args.name]
+            if self.args.unset:
+                section.pop(self.args.option, None)
+            else:
+                section[self.args.option] = self.args.value
+        return 0
+
+
+class CmdExecutorDefault(CmdExecutorConfig):
+    def run(self):
+        if self.args.name is None and not self.args.unset:
+            conf = self.config.read(self.args.level)
+            try:
+                print(conf["core"]["executor"])
+            except KeyError:
+                ui.write("No default executor set")
+                return 1
+        else:
+            with self.config.edit(self.args.level) as conf:
+                if self.args.unset:
+                    conf["core"].pop("executor", None)
+                else:
+                    merged_conf = self.config.load_config_to_level(
+                        self.args.level
+                    )
+                    if (
+                        self.args.name in conf["executor"]
+                        or self.args.name in merged_conf["executor"]
+                    ):
+                        conf["core"]["executor"] = self.args.name
+                    else:
+                        raise ConfigError(
+                            "default executor must be present in executor "
+                            "list."
+                        )
+        return 0
+
+
+class CmdExecutorInit(CmdBase):
+    def run(self):
+        self.repo.executor.init(self.args.name)
+
+
+class CmdExecutorDestroy(CmdBase):
+    def run(self):
+        self.repo.executor.destroy(self.args.name)
+
+
+def add_parser(subparsers, parent_parser):
+    from dvc.command.config import parent_config_parser
+
+    executor_HELP = "Set up and manage cloud executors."
+    executor_parser = subparsers.add_parser(
+        "executor",
+        parents=[parent_parser],
+        description=append_doc_link(executor_HELP, "executor"),
+        # NOTE: suppress help during development to hide command
+        # help=executor_HELP,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+
+    executor_subparsers = executor_parser.add_subparsers(
+        dest="cmd",
+        help="Use `dvc executor CMD --help` for " "command-specific help.",
+    )
+
+    fix_subparsers(executor_subparsers)
+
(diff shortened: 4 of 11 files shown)


pmrowla:
[asciicast](https://asciinema.org/a/n3wZ8IG8ATny6pGb1TSHTZvhT)

pmrowla:
Not really sure what we want to do with the terraform CLI output. It seems like something we should suppress, but there's no other useful way for us to display progress since we have to wrap their CLI (and operations to manage cloud machines will generally be slow)

I left out an `dvcexecutor list` for now since I'm not sure how we want to differentiate between listing what's in the DVC repo config (like `remote list`) and actually listing state for whether or not instances of machines from the config have been started/destroyed

---

@dberenbaum take a look and see what you think so far

pmrowla:
I'm also not sold on the `executor` naming here. Something like `machine` (like how docker uses `docker-machine`)might be better, since as it stands right now there is overlap between `dvc.executor` as in "cloud machine executor instance" and `dvc.repo.experiments.executor` as in "experiment pipeline executor"

I think the eventual workflow will basically be something along  the lines of
```
machine = Executor(Machine)Backend.init(machine_name)
executor = machine.get_executor()  # return a pipeline/experiment executor instance that will run on the desired machine
executor.reproduce()
```

pmrowla:
Also we probably don't actually need the jinja dependency, since the tf files we will generating are simple enough, but I left it in for now since the initial dvc-tpi example also used it.

dberenbaum:
Great stuff!

Responses to your comments:

> Not really sure what we want to do with the terraform CLI output. It seems like something we should suppress, but there's no other useful way for us to display progress since we have to wrap their CLI (and operations to manage cloud machines will generally be slow)

Not sure either. Should it take a long time to create/destroy machines? Maybe @skshetry can weigh in here.

> I left out a `dvc executor list` implementation for now since I'm not sure how we want to differentiate between listing what's in the DVC repo config (like `remote list`) and actually listing state for whether or not instances of machines from the config have been started/destroyed

Why not start with just the repo config and we can add more later? Even if the output format has to change, I don't see much downside unless I'm missing something. See also https://www.notion.so/iterative/Remote-Executors-215a0767aa0341d892b52767e58e2ac6#2281a6b533e4433496d0dfb8db6bf96b for ideas.

> I'm also not sold on the `executor` naming here. Something like `machine` (like how docker uses `docker-machine`) might be better, since as it stands right now there is overlap between `dvc.executor` as in "cloud machine executor instance" and `dvc.repo.experiments.executor` as in "experiment pipeline executor"

No strong feelings here. We can get some other opinions on this before we commit to the name.

> Also we probably don't actually need the jinja dependency, since the tf files we will generating are simple enough, but I left it in for now since the initial dvc-tpi example also used it.

Okay. We might need jinja anyway when we introduce `dvc exp init`. I forgot that it's not already used for plots.

---

Other thoughts:

* `python-terraform` is only available up to 0.10.1 on pypi (https://pypi.org/simple/python-terraform/), meaning that the pip install fails for me. It's not hard to work around, but it raises questions about this library, which I know you mentioned as well. 0.10.1 was released in June 2019 (https://github.com/beelit94/python-terraform/tags) and there's only one other release since then, which wasn't even uploaded to pypi.
* Do we need `dvc executor default`? How would this work?
* If a user has an existing terraform resource or directory with a `main.tf` (and possibly state files), is there a way for us to manage this? Not sure we need to support this yet, but we should consider whether it will be easy to implement.
* Why make `cloud` a positional argument? This might relate to the previous bullet and making sure we retain flexibility.
* Should we add a `rename` subcommand to be consistent with `remote`?
* Is `init` a good name for what it does? It looks like it does `terraform init && terraform apply`. Is this going to be confusing for terraform users to have `init` actually create instances? Maybe `start` or `create` make sense?

pmrowla:
> Why not start with just the repo config and we can add more later? Even if the output format has to change, I don't see much downside unless I'm missing something. See also https://www.notion.so/iterative/Remote-Executors-215a0767aa0341d892b52767e58e2ac6#2281a6b533e4433496d0dfb8db6bf96b for ideas.

This is fine with me, but this type of stuff can also be split out into separate tasks
 
> * Do we need `dvc executor default`? How would this work?

I'm not really sure, it will depend on where we want to go with usage I think. I was thinking along the lines of `local` is basically a default (and currently the only option in dvc), but users might not want that behavior. But it may make more sense for users to be specifying alternative default executors on a per-pipeline or per-stage basis in `dvc.yaml` like we've discussed before.

> * If a user has an existing terraform resource or directory with a `main.tf` (and possibly state files), is there a way for us to manage this? Not sure we need to support this yet, but we should consider whether it will be easy to implement.

We will probably want a `executor.terraform_dir` or similar config option, that would allow users to specify where their existing terraform environment is (and override the use of `.dvc/tmp/exec/terraform`

> * Why make `cloud` a positional argument? This might relate to the previous bullet and making sure we retain flexibility.

Mostly just for config schema validation reasons. It's possible to modify it with `dvc executor modify myexecutor cloud azure` after the fact (the same way you can modify remote URLs after the fact). But I'm not opposed to changing the usage around.

> * Should we add a `rename` subcommand to be consistent with `remote`?

It seems like we will need one, but I left it out for now for the same reasons as `list`. I don't think `executor rename` can be a pure config command - it will probably need to modify terraform state to rename existing machines that have already been started (via `terraform mv`)

> * Is `init` a good name for what it does? It looks like it does `terraform init && terraform apply`. Is this going to be confusing for terraform users to have `init` actually create instances? Maybe `start` or `create` make sense?

`init` and `destroy` make sense to me from a "design a generic workflow that in theory could run on a backend other than terraform" standpoint, but `start` or `create` would work too.

pmrowla:
Also, I should note that for anyone testing this, I'm pretty sure that creating more than one executor will not work properly since the current `main.tf` generation just replaces your entire `main.tf` file instead of properly adding/removing additional resources to an existing tf file.

(this will need to be addressed in the future)

pmrowla:
Also, regarding jinja, we could also drop it if we opt to make DVC-generated terraform configs non-human-readable and output them as JSON (`.tf.json`) instead of terraform HCL (`.tf`).

I guess the relevant question here would be whether or not we intentionally want the DVC-generated terraform configs to be user-editable. Users are always going to be able to manually modify the files, but writing them in JSON will make it more obvious that they are not supposed to be manually edited (and that users should use `dvc executor ...` to manage them instead)

pmrowla:
Regarding managing multiple executors, it seems like what we will need is separate terraform directories per executor (so `.dvc/tmp/exec/terraform/<executor_name>/main.tf` instead of a single top level `main.tf` containing multiple resources.

We want to individually provision/destroy the machines per DVC executor (i.e. a single terraform resource), but terraform is intended to be used on a per-configuration (encompassing all possible resources) basis rather than a per-resource basis. `apply/destroy -target=...` does support targeting individual resources, but the docs and the CLI include warnings about how you aren't supposed to use terraform in that way.

The thing to note here would be that doing separate configurations per executor means we will have multiple terraform states to manage rather than a single state w/all of our possible resources.

dberenbaum:
> This is fine with me, but this type of stuff can also be split out into separate tasks

Sounds good.

> I'm not really sure, it will depend on where we want to go with usage I think. I was thinking along the lines of `local` is basically a default (and currently the only option in dvc), but users might not want that behavior. But it may make more sense for users to be specifying alternative default executors on a per-pipeline or per-stage basis in `dvc.yaml` like we've discussed before.

Consistent with above, let's leave it out for now and come back to it.

> Mostly just for config schema validation reasons. It's possible to modify it with `dvc executor modify myexecutor cloud azure` after the fact (the same way you can modify remote URLs after the fact). But I'm not opposed to changing the usage around.

My concern with the usage is that there could be times when `cloud` is not a relevant option (such as connecting to an existing resource like an internal host).

> It seems like we will need one, but I left it out for now for the same reasons as `list`. I don't think `executor rename` can be a pure config command - it will probably need to modify terraform state to rename existing machines that have already been started (via `terraform mv`)

We can leave `rename` for a separate future task.

> Regarding managing multiple executors, it seems like what we will need is separate terraform directories per executor (so `.dvc/tmp/exec/terraform/<executor_name>/main.tf` instead of a single top level `main.tf` containing multiple resources.

👍

casperdcl:
- `.json` vs `.tf`: I think the underlying https://github.com/iterative/terraform-provider-iterative supports both (https://github.com/iterative/terraform-provider-iterative/issues/180)
- I think it makes much more sense to abstract the terraform stuff here to a separate python library (https://github.com/iterative/terraform-provider-iterative/issues/164#issuecomment-901084421)
  + `pip install tpi` could handle downloading `terraform.go` binaries (or even compiling them)
  + `pip install dvc[terraform/tpi/machine/...]` won't have to directly handle any dependencies

/CC @iterative/cml

0x2b3bfa0:
> * `pip install tpi` could handle [downloading `tpi.go`](https://github.com/iterative/terraform-provider-iterative/releases) & `terraform.go` binaries (or even compiling them)

If we plan to stick with Go, as per https://github.com/iterative/terraform-provider-iterative/issues/164#issuecomment-903144469, we could provide a Python wrapper package including precompiled wheels.

pmrowla:
> My concern with the usage is that there could be times when `cloud` is not a relevant option (such as connecting to an existing resource like an internal host).

In this case I think `cloud` would be `ssh`, and then we would have `ssh` options specific to DVC for hostname/port (since it's not an terraform-provider-iterative "cloud")

dvc/config_schema.py
```diff
@@ -224,10 +225,16 @@ class RelPath(str):
         "row_limit": All(Coerce(int), Range(1)),  # obsoleted
         "row_cleanup_quota": All(Coerce(int), Range(0, 100)),  # obsoleted
     },
+    "machine": {
+        str: {
+            "cloud": All(Lower, Choices("aws", "azure")),
```
casperdcl, line 230: maybe

```suggestion
            "cloud": All(Lower, Choices("aws", "azure", "gcp", "kubernetes")),
```

/CC @iterative/cml

setup.py
```diff
@@ -138,6 +140,7 @@ def run(self):
         "hdfs": hdfs,
         "webhdfs": webhdfs,
         "webdav": webdav,
+        "terraform": terraform,
```
casperdcl, line 143:
```suggestion
        "machine": terraform,
```

dvc/config_schema.py
dberenbaum: See also https://github.com/iterative/dvc/blob/f77a5d2402600272eee6fff5de9cd21ca511c60a/dvc/machine/__init__.py#L91

setup.py
0x2b3bfa0: :100: Or whatever name we choose for cloud execution capabilities.

0x2b3bfa0:
> In this case I think `cloud` would be `ssh`, and then we would have `ssh` options specific to DVC for hostname/port (since it's not an terraform-provider-iterative "cloud")

The question is, should we support DVC logging in through SSH to an unknown environment and acting like it knows what is there and how to find it? Cloud execution is "easy" because we provide the system images and know how to provision them, but accessing a custom system is a completely different scenario.

The only way of mitigating compatibility issues I can think of would be providing a reference Ansible role or relying on Docker on the target systems, but still... 🤔

dberenbaum:
We aren't going to ssh into an existing machine anytime soon most likely, but keeping it in mind as a possible option is what matters for now.

Both `ssh` and `kubernetes` seem like odd values to provide for the `cloud` argument IMO. I guess we are copying that from the terraform-provider-iterative syntax. It seems more natural to call it something more generic like `provider` since that's the equivalent in terraform. Not a blocker, maybe it's just me 🤷 .

dvc/config_schema.py
pmrowla: Do gcp and kubernetes work with the existing TPI? I was just going off of what's documented in the readme right now

> The Iterative Provider currently supports AWS and Azure. Google Cloud Platform is not currently supported.

pmrowla: Either way, this kind of thing can also be addressed later (https://github.com/iterative/dvc/issues/6480)

This PR is not meant to be a complete/finished implementation in any way

setup.py
pmrowla: > Both `ssh` and `kubernetes` seem like odd values to provide for the `cloud` argument IMO. I guess we are copying that from the terraform-provider-iterative syntax. It seems more natural to call it something more generic like `provider` since that's the equivalent in terraform. Not a blocker, maybe it's just me 🤷 .

Yeah, it's `cloud` because that's the name of the TPI option. Basically I think the `.dvc/config` options should 1:1 map with the TPI ones. It doesn't really make sense for us to say "dvc config should have `provider = aws` but if you are using TPI yourself you need to use `cloud = aws`"

pmrowla:
> Both `ssh` and `kubernetes` seem like odd values to provide for the `cloud` argument IMO. I guess we are copying that from the terraform-provider-iterative syntax. It seems more natural to call it something more generic like `provider` since that's the equivalent in terraform. Not a blocker, maybe it's just me 🤷 .

Yeah, it's `cloud` because that's the name of the TPI option. Basically I think the `.dvc/config` options should 1:1 map with the TPI ones. It doesn't really make sense for us to say "dvc config should have `provider = aws` but if you are using TPI yourself you need to use `cloud = aws`"

pmrowla:
> I'm inclined to think that Kubernetes should be our go–to abstraction layer for on–premises executor support; trying to operate at lower levels could be risky for maintenance, support and user experience.

It would be better to have this kind of discussion in notion - in this case there's a dedicated story for managing existing/custom resources, but it's not a requirement that we are planning on addressing any time in the near future.

setup.py
pmrowla: I named the extra `terraform` for now because the `dvc machine` feature itself is supposed to be independent of whatever backend is actually used to provision machines (in this case, the only supported backend right now would be `terraform`). So in this case to be able to specifically use `dvc machine` with the terraform-based backend, the install extra is `terraform`.

Hypothetically, if someone else wanted to write a different backend (that wrapped something like docker-machine for provisioning instead of terraform) you could do that, and then install `dvc[docker-machine]` to use it instead of `dvc[terraform]`.

The more realistic example would be that eventually we may support arbitrary ssh machines in general (without any provisioning backend), in which case you should just be able to use `dvc machine` out of the box without installing any extras at all.

dvc/config_schema.py
casperdcl: Yes `gcp` & `kubernetes` are in the same category as `dvclive` - they *should* work even though the docs aren't there yet.

setup.py
casperdcl: yes fully agree, however... do we realistically think we may support other `machine` backends? As-is, TPI is a lot of work. For that reason I'm not convinced it's exactly the same as `dvc[remote]`.

pmrowla:
Using a wiki page makes the most sense for now I think

Issue #5989, opened 2021-05-10
exp show: using wrong results for display


pared:
```bash
DVC version: 2.1.0+fc4e31 (osxpkg)
---------------------------------
Platform: Python 3.8.2 on macOS-11.3.1-x86_64-i386-64bit
Supports: All remotes
Cache types: reflink, hardlink, symlink
Cache directory: apfs on /dev/disk1s1s1
Caches: local
Remotes: https
Workspace directory: apfs on /dev/disk1s1s1
Repo: dvc, git

```

Seems like #5880 introduced some bug:
Given experiments runs:
```python
def test_something(tmp_dir, scm, dvc, exp_stage):
    baseline_rev = scm.get_rev()
    exp1 = dvc.experiments.run(exp_stage.addressing, params=["foo=2"])
    exp2 = dvc.experiments.run(exp_stage.addressing, params=["foo=3"])

    result = dvc.experiments.show()
    rev1 = first(exp1)
    rev2 = first(exp2)

    assert result[baseline_rev][rev1]["params"]["params.yaml"] == {"foo": 2}
    assert result[baseline_rev][rev2]["params"]["params.yaml"] == {"foo": 3}
```
Note that the asserts seem to verify that internals work correctly
if we go to the test directory and run `dvc exp show`, we will get:
```bash
┏━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━┳━━━━━┓
┃ Experiment    ┃ Created  ┃ foo ┃ foo ┃
┡━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━╇━━━━━┩
│ workspace     │ -        │   3 │ 3   │
│ master        │ 04:54 PM │   3 │ 3   │
│ ├── exp-e0d04 │ 04:54 PM │   1 │ 1   │
│ └── exp-42ced │ 04:54 PM │   1 │ 1   │
└───────────────┴──────────┴─────┴─────┘
```

while it should more like:

```bash
┏━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━┳━━━━━┓
┃ Experiment    ┃ Created  ┃ foo ┃ foo ┃
┡━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━╇━━━━━┩
│ workspace     │ -        │   3 │ 3   │
│ master        │ 04:54 PM │   1 │ 1   │
│ ├── exp-42ced │ 04:54 PM │   2 │ 2   │
[...]

PR #6468, opened 2021-08-23
Show csv format of experiments

### LINKED ISSUE #5446: exp show csv/tsv option
See https://github.com/iterative/dvc/pull/5381#issuecomment-773555945 for context. 

This is NOT a requirement for 2.0 release, but a nice-to-have that could be added after release.

Proposed format:
```
"Branch","Experiment","Revision","Queued","Checkpoint Tip","Checkpoint Parent","Created",...(params and metrics columns as they appear in CLI)
"workspace","baseline",,false,,,,...
"master","baseline","bcbee294fc6df6d096acdf1872a54e05818ce427",false,,,"2021-02-09T13:05:04",...
"master","exp-cd493","bcbee294fc6df6d096acdf1872a54e05818ce427",false,,,"2021-02-09T13:05:04",...
"master","named-exp","d4caebb5162878d16ea89ec125f88cb595a08491",false,,,"2021-02-10T11:18:00",...
"master","exp-deb9f","24f0f31b6daa46d57b7cd55d9238d55971b8a2b2",true,,,"2021-02-10T14:13:00",...
"checkpoints_branch","exp-c5edd","73b69e032634e69a0f96fb8f6c41d51485a9428d",false,"73b69e032634e69a0f96fb8f6c41d51485a9428d","0a7d8aa973c72bb88cb308b06273777260f21de6","2021-02-10T14:46:14",...
"checkpoints_branch","exp-c5edd","0a7d8aa973c72bb88cb308b06273777260f21de6",false,"73b69e032634e69a0f96fb8f6c41d51485a9428d","bcbee294fc6df6d096acdf1872a54e05818ce427","2021-02-10T14:46:13",...
```

Questions:
- Do the columns names make sense?
- Is any info missing?
- Is "Checkpoint Tip" redundant if "Experiment ID" is included in each row? Would a checkpoint boolean column be more helpful (indicating whether or not this is a checkpoint)?

@pmrowla @dmpetrov

karajan1001:
fix #5446 

1. add --show-csv to dvc exp show
2. add tests for dvc exp show --show-csv

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -463,18 +463,9 @@ def _format_json(item):
 
 
 class CmdExperimentsShow(CmdBase):
-    @staticmethod
-    def _show_csv(td):
-        _post_process_td(td)
-        ui.write(td.to_csv())
-
-    @staticmethod
-    def _show_json(all_experiments):
+    def run(self):
         import json
 
-        ui.write(json.dumps(all_experiments, default=_format_json))
-
-    def run(self):
         try:
             all_experiments = self.repo.experiments.show(
                 all_branches=self.args.all_branches,
@@ -489,7 +480,7 @@ class CmdExperimentsShow(CmdBase):
             return 1
 
         if self.args.show_json:
-            self._show_json(all_experiments)
+            ui.write(json.dumps(all_experiments, default=_format_json))
             return 0
 
         metric_names, param_names = _collect_names(
@@ -511,7 +502,7 @@ class CmdExperimentsShow(CmdBase):
         )
 
         if self.args.show_csv:
-            self._show_csv(td)
+            ui.write(td.to_csv(), end="")
         else:
             row_styles = lmap(baseline_styler, td.column("typ"))
             styles = _get_styles(metric_names, param_names)
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -326,6 +326,7 @@ def _parse_filter_list(param_list):
 
 def experiments_table(
     all_experiments,
+    headers,
     metric_headers,
     metric_names,
     param_headers,
@@ -340,15 +341,6 @@ def experiments_table(
 
     from dvc.compare import TabularData
 
-    headers = [
-        "Experiment",
-        "rev",
-        "typ",
-        "Created",
-        "parent",
-        "State",
-        "Executor",
-    ]
     td = TabularData(
         lconcat(headers, metric_headers, param_headers), fill_value=fill_value
     )
@@ -412,13 +404,25 @@ def show_experiments(
         exclude_params=exclude_params,
     )
 
+    headers = [
+        "Experiment",
+        "rev",
+        "typ",
+        "Created",
+        "parent",
+        "State",
+        "Executor",
+    ]
+
     names = {**metric_names, **param_names}
     counter = Counter(flatten_list([list(a.keys()) for a in names.values()]))
+    counter.update(headers)
     metric_headers = _normalize_headers(metric_names, counter)
     param_headers = _normalize_headers(param_names, counter)
 
     td = experiments_table(
         all_experiments,
+        headers,
         metric_headers,
         metric_names,
         param_headers,
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -511,7 +511,11 @@ class CmdExperimentsShow(CmdBase):
 
             ui.write(json.dumps(all_experiments, default=_format_json))
         else:
-            precision = None if self.args.show_csv else self.args.precision
+            precision = (
+                self.args.precision or None
+                if self.args.show_csv
+                else DEFAULT_PRECISION
+            )
             fill_value = "" if self.args.show_csv else FILL_VALUE
             iso = True if self.args.show_csv else False
 
@@ -916,7 +920,6 @@ def add_parser(subparsers, parent_parser):
     experiments_show_parser.add_argument(
         "--precision",
         type=int,
-        default=DEFAULT_PRECISION,
         help=(
             "Round metrics/params to `n` digits precision after the decimal "
             f"point. Rounds to {DEFAULT_PRECISION} digits by default."
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -417,7 +417,7 @@ def _post_process_td(td):
     return td
 
 
-def get_styles(metric_names: Dict, param_names: Dict) -> Dict:
+def _get_styles(metric_names: Dict, param_names: Dict) -> Dict:
     metric_headers = _normalize_headers(metric_names)
     param_headers = _normalize_headers(param_names)
     headers = {"metrics": metric_headers, "params": param_headers}
@@ -514,7 +514,7 @@ class CmdExperimentsShow(CmdBase):
             self._show_csv(td)
         else:
             row_styles = lmap(baseline_styler, td.column("typ"))
-            styles = get_styples(metric_names, param_names)
+            styles = _get_styles(metric_names, param_names)
             _post_process_td(td)
             td.render(
                 pager=not self.args.no_pager,
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -415,9 +415,6 @@ def show_experiments(
     metric_headers = _normalize_headers(metric_names, counter)
     param_headers = _normalize_headers(param_names, counter)
 
-    precision = None if show_csv else kwargs.get("precision")
-    fill_value = "" if show_csv else FILL_VALUE
-
     td = experiments_table(
         all_experiments,
         metric_headers,
@@ -426,8 +423,8 @@ def show_experiments(
         param_names,
         kwargs.get("sort_by"),
         kwargs.get("sort_order"),
-        precision,
-        fill_value,
+        None if show_csv else kwargs.get("precision"),
+        "" if show_csv else FILL_VALUE,
         True if show_csv else False,
     )
 
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -384,21 +384,9 @@ def baseline_styler(typ):
     return {"style": "bold"} if typ == "baseline" else {}
 
 
-def show_experiments(
-    all_experiments, pager=True, no_timestamp=False, **kwargs
-):
-    include_metrics = _parse_filter_list(kwargs.pop("include_metrics", []))
-    exclude_metrics = _parse_filter_list(kwargs.pop("exclude_metrics", []))
-    include_params = _parse_filter_list(kwargs.pop("include_params", []))
-    exclude_params = _parse_filter_list(kwargs.pop("exclude_params", []))
-
-    metric_names, param_names = _collect_names(
-        all_experiments,
-        include_metrics=include_metrics,
-        exclude_metrics=exclude_metrics,
-        include_params=include_params,
-        exclude_params=exclude_params,
-    )
+def _experiments_td(
+    all_experiments, metric_names, param_names, no_timestamp=False, **kwargs
+) -> "TabularData":
     metric_headers = _normalize_headers(metric_names)
     param_headers = _normalize_headers(param_names)
 
@@ -419,14 +407,21 @@ def show_experiments(
     for col in ("State", "Executor"):
         if td.is_empty(col):
             td.drop(col)
+    return td
 
-    row_styles = lmap(baseline_styler, td.column("typ"))
 
+def _post_process_td(td):
     merge_headers = ["Experiment", "rev", "typ", "parent"]
     td.column("Experiment")[:] = map(prepare_exp_id, td.as_dict(merge_headers))
     td.drop(*merge_headers[1:])
+    return td
 
+
+def get_styples(metric_names: Dict, param_names: Dict) -> Dict:
+    metric_headers = _normalize_headers(metric_names)
+    param_headers = _normalize_headers(param_names)
     headers = {"metrics": metric_headers, "params": param_headers}
+
     styles = {
         "Experiment": {"no_wrap": True, "header_style": "black on grey93"},
         "Created": {"header_style": "black on grey93"},
@@ -447,13 +442,7 @@ def show_experiments(
         }
     )
 
-    td.render(
-        pager=pager,
-        borders=True,
-        rich_table=True,
-        header_styles=styles,
-        row_styles=row_styles,
-    )
+    return styles
 
 
 def _normalize_headers(names):
@@ -474,6 +463,17 @@ def _format_json(item):
 
 
 class CmdExperimentsShow(CmdBase):
+    @staticmethod
+    def _show_csv(td):
+        _post_process_td(td)
+        ui.write(td.to_csv())
+
+    @staticmethod
+    def _show_json(all_experiments):
+        import json
+
+        ui.write(json.dumps(all_experiments, default=_format_json))
+
     def run(self):
         try:
             all_experiments = self.repo.experiments.show(
@@ -489,21 +489,39 @@ class CmdExperimentsShow(CmdBase):
             return 1
 
         if self.args.show_json:
-            import json
+            self._show_json(all_experiments)
+            return 0
 
-            ui.write(json.dumps(all_experiments, default=_format_json))
+        metric_names, param_names = _collect_names(
+            all_experiments,
+            include_metrics=_parse_filter_list(self.args.include_metrics),
+            exclude_metrics=_parse_filter_list(self.args.exclude_metrics),
+            include_params=_parse_filter_list(self.args.include_params),
+            exclude_params=_parse_filter_list(self.args.exclude_params),
+        )
+
+        td = _experiments_td(
+            all_experiments,
+            metric_names,
+            param_names,
+            no_timestamp=self.args.no_timestamp,
+            sort_by=self.args.sort_by,
+            sort_order=self.args.sort_order,
+            precision=self.args.precision or DEFAULT_PRECISION,
+        )
+
+        if self.args.show_csv:
+            self._show_csv(td)
         else:
-            show_experiments(
-                all_experiments,
-                include_metrics=self.args.include_metrics,
-                exclude_metrics=self.args.exclude_metrics,
-                include_params=self.args.include_params,
-                exclude_params=self.args.exclude_params,
-                no_timestamp=self.args.no_timestamp,
-                sort_by=self.args.sort_by,
-                sort_order=self.args.sort_order,
-                precision=self.args.precision or DEFAULT_PRECISION,
+            row_styles = lmap(baseline_styler, td.column("typ"))
+            styles = get_styples(metric_names, param_names)
+            _post_process_td(td)
+            td.render(
                 pager=not self.args.no_pager,
+                borders=True,
+                rich_table=True,
+                header_styles=styles,
+                row_styles=row_styles,
             )
         return 0
 
@@ -881,6 +899,12 @@ def add_parser(subparsers, parent_parser):
         action="store_true",
         default=False,
         help="Print output in JSON format instead of a human-readable table.",
+    )
+    experiments_show_parser.add_argument(
+        "--show-csv",
+        action="store_true",
+        default=False,
+        help="Print output in csv format instead of a human-readable table.",
     )
     experiments_show_parser.add_argument(
         "--precision",
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -279,12 +279,14 @@ def _sort_exp(experiments, sort_path, sort_name, typ, reverse):
 def _format_time(datetime_obj, fill_value=FILL_VALUE, iso=False):
     if datetime_obj is None:
         return fill_value
+
+    if iso:
+        return datetime_obj.isoformat()
+
     if datetime_obj.date() == date.today():
         fmt = "%I:%M %p"
     else:
         fmt = "%b %d, %Y"
-    if iso:
-        return datetime_obj.isoformat()
     return datetime_obj.strftime(fmt)
 
 
@@ -423,9 +425,9 @@ def show_experiments(
         param_names,
         kwargs.get("sort_by"),
         kwargs.get("sort_order"),
-        None if show_csv else kwargs.get("precision"),
-        "" if show_csv else FILL_VALUE,
-        True if show_csv else False,
+        kwargs.get("precision"),
+        kwargs.get("fill_value"),
+        kwargs.get("iso"),
     )
 
     if no_timestamp:
@@ -509,6 +511,10 @@ class CmdExperimentsShow(CmdBase):
 
             ui.write(json.dumps(all_experiments, default=_format_json))
         else:
+            precision = None if self.args.show_csv else self.args.precision
+            fill_value = "" if self.args.show_csv else FILL_VALUE
+            iso = True if self.args.show_csv else False
+
             show_experiments(
                 all_experiments,
                 include_metrics=self.args.include_metrics,
@@ -518,7 +524,9 @@ class CmdExperimentsShow(CmdBase):
                 no_timestamp=self.args.no_timestamp,
                 sort_by=self.args.sort_by,
                 sort_order=self.args.sort_order,
-                precision=self.args.precision or DEFAULT_PRECISION,
+                precision=precision,
+                fill_value=fill_value,
+                iso=iso,
                 pager=not self.args.no_pager,
                 show_csv=self.args.show_csv,
             )
@@ -908,6 +916,7 @@ def add_parser(subparsers, parent_parser):
     experiments_show_parser.add_argument(
         "--precision",
         type=int,
+        default=DEFAULT_PRECISION,
         help=(
             "Round metrics/params to `n` digits precision after the decimal "
             f"point. Rounds to {DEFAULT_PRECISION} digits by default."
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -141,6 +141,8 @@ def _collect_rows(
     precision=DEFAULT_PRECISION,
     sort_by=None,
     sort_order=None,
+    fill_value=FILL_VALUE,
+    iso=False,
 ):
     from dvc.scm.git import Git
 
@@ -161,8 +163,8 @@ def _collect_rows(
         elif exp.get("queued"):
             state = "Queued"
         else:
-            state = FILL_VALUE
-        executor = exp.get("executor", FILL_VALUE)
+            state = fill_value
+        executor = exp.get("executor", fill_value)
         is_baseline = rev == "baseline"
 
         if is_baseline:
@@ -202,12 +204,12 @@ def _collect_rows(
             exp_name,
             name_rev,
             typ,
-            _format_time(exp.get("timestamp")),
+            _format_time(exp.get("timestamp"), fill_value, iso),
             parent,
             state,
             executor,
         ]
-        fill_value = FILL_VALUE_ERRORED if results.get("error") else FILL_VALUE
+        fill_value = FILL_VALUE_ERRORED if results.get("error") else fill_value
         _extend_row(
             row,
             metric_names,
@@ -274,14 +276,16 @@ def _sort_exp(experiments, sort_path, sort_name, typ, reverse):
     return ret
 
 
-def _format_time(timestamp):
-    if timestamp is None:
-        return FILL_VALUE
-    if timestamp.date() == date.today():
+def _format_time(datetime_obj, fill_value=FILL_VALUE, iso=False):
+    if datetime_obj is None:
+        return fill_value
+    if datetime_obj.date() == date.today():
         fmt = "%I:%M %p"
     else:
         fmt = "%b %d, %Y"
-    return timestamp.strftime(fmt)
+    if iso:
+        return datetime_obj.isoformat()
+    return datetime_obj.strftime(fmt)
 
 
 def _extend_row(row, names, items, precision, fill_value=FILL_VALUE):
@@ -327,6 +331,8 @@ def experiments_table(
     sort_by=None,
     sort_order=None,
     precision=DEFAULT_PRECISION,
+    fill_value=FILL_VALUE,
+    iso=False,
 ) -> "TabularData":
     from funcy import lconcat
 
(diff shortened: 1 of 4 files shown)


pmrowla:
Is just dumping the exact `exp show` CLI table content into a CSV actually the desired behavior here? What we get right now is:

```
Experiment,Created,avg_prec,roc_auc,prepare.split,prepare.seed,featurize.max_features,featurize.ngrams,train.seed,train.n_est,train.min_split,max_features
workspace,-,0.60405,0.9608,0.2,20170428,3000,2,20170428,100,64,2500
bee447d,"Jun 02, 2021",0.67038,0.96693,0.2,20170428,3000,2,20170428,100,64,-
master,"May 29, 2021",0.60405,0.9608,0.2,20170428,3000,2,20170428,100,64,-
└── ac627ad [exp-44136],01:14 PM,0.60405,0.9608,0.2,20170428,3000,2,20170428,100,64,2500
cc51022,"May 28, 2021",0.55259,0.91536,0.2,20170428,1500,2,20170428,50,2,-
```

It seems like the merged `Experiment` column from the show table should actually be broken up into separate columns for a CSV formatted output (i.e. having separate columns for SHA, name, and parent instead of using the `tree character + sha + [name]` format used in the CLI output).

Also it seems like we should be outputting the ISO8601 timestamps for the created column, instead of the formatted output we use in the CLI table.

It also seems like we should just be not outputting the `-` for empty cells as well (in a CSV you can just output the empty cell (so for `col1,,col3` col2 is empty)

Basically I think we need to be doing more than just using the existing `TabularData` and calling `to_csv()` on it. If we are writing to CSV, the data itself needs to be formatted differently than if we are outputting to the CLI. For comparison, the JSON output for one of these experiment rows is:
```json
{
  "ac627ad7b1996289695199c6c0ffb74bb1a57c9f":{
    "data":{
      "timestamp":"2021-08-23T13:14:33",
      "params":{
        "params.yaml":{
          "data":{
            "prepare":{
              "split":0.2,
              "seed":20170428
            },
            "featurize":{
              "max_features":3000,
              "ngrams":2
            },
            "train":{
              "seed":20170428,
              "n_est":100,
              "min_split":64
            },
            "max_features":2500
          }
        }
      },
      "queued":false,
      "running":false,
      "executor":null,
      "metrics":{
        "scores.json":{
          "data":{
            "avg_prec":0.6040544652105823,
            "roc_auc":0.9608017142900953
          }
        }
      },
      "name":"exp-44136"
    }
  }
}
```
You can see that in the JSON, we use full git SHAs (not shortened ones), full precision for floating point numbers, and detailed timestamps. I think it should be the same if we are writing to CSV. For CSV output we should probably also always be including the full path for metrics/params files instead of dropping the default path segment in column headers.

dvc/command/experiments.py
```diff
@@ -419,14 +407,21 @@ def show_experiments(
...
-    row_styles = lmap(baseline_styler, td.column("typ"))
 
+def _post_process_td(td):
     merge_headers = ["Experiment", "rev", "typ", "parent"]
     td.column("Experiment")[:] = map(prepare_exp_id, td.as_dict(merge_headers))
     td.drop(*merge_headers[1:])
+    return td
 
+
+def get_styples(metric_names: Dict, param_names: Dict) -> Dict:
```
pmrowla, line 420:
```suggestion
def get_styles(metric_names: Dict, param_names: Dict) -> Dict:
```

dberenbaum:
> It seems like the merged `Experiment` column from the show table should actually be broken up into separate columns for a CSV formatted output (i.e. having separate columns for SHA, name, and parent instead of using the `tree character + sha + [name]` format used in the CLI output).

Can we just drop https://github.com/iterative/dvc/blob/bbbf15c75c0fa751ef81fc0e526f2b940c7b58e6/dvc/command/experiments.py#L468?

> Also it seems like we should be outputting the ISO8601 timestamps for the created column, instead of the formatted output we use in the CLI table.

I think it's okay as is and some users may prefer human readable dates. Many csv readers will either automatically handle date parsing or will provide an option to provide the date format.

> It also seems like we should just be not outputting the `-` for empty cells as well (in a CSV you can just output the empty cell (so for `col1,,col3` col2 is empty)

This would be nice if it's not too much trouble. It could also probably be handled pretty easily by the user if it is too messy to do on the dvc side.

> Basically I think we need to be doing more than just using the existing `TabularData` and calling `to_csv()` on it. If we are writing to CSV, the data itself needs to be formatted differently than if we are outputting to the CLI.

I think the differences between the table and `--show-json` are actually confusing and not well documented today. Users might expect `--show-csv` output to look as much as possible like the table output, especially since it's a similar structure.

I agree that some users will probably want output more similar to `--show-json` since it's better structured and more informative, but it seems reasonable to start with the existing table output since it's much simpler and probably matches some users' expectations ("I want this table as a csv").

pmrowla:
> Can we just drop
> 
> https://github.com/iterative/dvc/blob/bbbf15c75c0fa751ef81fc0e526f2b940c7b58e6/dvc/command/experiments.py#L468

Yes

> I think it's okay as is and some users may prefer human readable dates. Many csv readers will either automatically handle date parsing or will provide an option to provide the date format.

The thing is that the timestamps are timestamps though (date + time) and in the table we only display times for experiments that are from the current day, and dates for everything else (to match the default viewer behavior). This is fine for making the CLI table human readable but it seems like, it would still be more useful to have the date + time in the CSV output, even if we output it as human readable formatted "Month Day Year, hh:mm:ss" strings instead of an ISO8601 timestamp.

> This would be nice if it's not too much trouble. It could also probably be handled pretty easily by the user if it is too messy to do on the dvc side.

This can be done on the DVC side by changing the default empty data placeholder string (we want to use `""` instead of `"-"`)

> I agree that some users will probably want output more similar to `--show-json` since it's better structured and more informative, but it seems reasonable to start with the existing table output since it's much simpler and probably matches some users' expectations ("I want this table as a csv").

If this is really what users want then I don't have any objections to this PR, but I guess I don't really see the use case where users would want less detail in their CSV.

If users want to format columns in some particular way (string formatting for dates/float precision/etc) they can do it in their spreadsheet viewer. But especially with stuff like float precision - I can tell excel to only show 4 decimal places even if the original CSV data includes 10+ decimal places, but I cannot do it the other way around if DVC is only outputting the 4 decimal places to start with.

dberenbaum:
@pmrowla All of your points make sense. I'm not sure how much additional work each of those changes are. If they take significant additional effort, I think it's acceptable to leave them as is since it has parity with the existing table output.

skshetry:
TabularData is just a 2D data structure, it has nothing to do with CLI. We should easily be able to skip merging headers for the row index, skip formatting timestamp, precision (this is already possible by passing `precision=None`). Also, the fill value is hardcoded now which can be changed to just drill down, so that we can specify different fill values for csv vs non-csv.

Regarding the headers, ideally, we should always be keeping the headers with full prefixes, and only normalize them just before rendering. It may be a bit more involved so we could skip it for now.

dvc/command/experiments.py
```diff
@@ -384,21 +384,9 @@ def baseline_styler(typ):
...
-    exclude_params = _parse_filter_list(kwargs.pop("exclude_params", []))
-
-    metric_names, param_names = _collect_names(
-        all_experiments,
-        include_metrics=include_metrics,
-        exclude_metrics=exclude_metrics,
-        include_params=include_params,
-        exclude_params=exclude_params,
-    )
+def _experiments_td(
```
skshetry, line 387: I'd prefer not to have too many layers here. Initially, I divided it into `show_experiments` (that knows about CLI flags and knows how to render/style) and `experiments_table` that simply parses and does not care about CLI.

dvc/command/experiments.py
```diff
@@ -489,21 +489,39 @@ def run(self):
             return 1
 
         if self.args.show_json:
-            import json
+            self._show_json(all_experiments)
```
skshetry, line 492: Why do we need `_show_json` and `_show_csv`?

dvc/command/experiments.py
```diff
@@ -489,21 +489,39 @@ def run(self):
...
-                include_params=self.args.include_params,
-                exclude_params=self.args.exclude_params,
-                no_timestamp=self.args.no_timestamp,
-                sort_by=self.args.sort_by,
-                sort_order=self.args.sort_order,
-                precision=self.args.precision or DEFAULT_PRECISION,
+            row_styles = lmap(baseline_styler, td.column("typ"))
+            styles = _get_styles(metric_names, param_names)
+            _post_process_td(td)
+            td.render(
```
skshetry, line 519: We could extend `render` to also print csv for us, WDYT?

tests/unit/command/test_experiments.py
```diff
@@ -58,6 +58,32 @@ def test_experiments_diff(dvc, scm, mocker):
     )
 
 
+@pytest.mark.parametrize("show_json,show_csv", [(True, False), (False, True)])
+def test_experiments_show_format(dvc, scm, mocker, show_json, show_csv):
```
skshetry, line 62: We may need a functional test for this instead, as csv and the default output might have significant differences.

dvc/command/experiments.py
```diff
@@ -474,6 +463,17 @@ def _format_json(item):
 
 
 class CmdExperimentsShow(CmdBase):
+    @staticmethod
+    def _show_csv(td):
+        _post_process_td(td)
+        ui.write(td.to_csv())
```
skshetry, line 469: Also, `to_csv()` already adds a `\r\n`, so, there'll be two empty lines at the end, one from csv, and other from `ui.write`.
You may want to use `ui.write(..., end="")`.

dvc/command/experiments.py
karajan1001: One problems `--show-json` still outside this render.

karajan1001:
@pmrowla @dberenbaum for the precision problem `--precision` works on `--show-csv`

pmrowla:
> @pmrowla @dberenbaum for the precision problem `--precision` works on `--show-csv`

Yes, but there is no way for users to specify full precision right now. It's currently set using
```py
precision=self.args.precision or DEFAULT_PRECISION,
```
meaning that when `--precision` isn't used (and is `None`), it actually defaults to 5 instead of full precision.

This behavior is what we want for display in the CLI table by default, but not for a data format like CSV. It should default to full precision (meaning explicit `precision=None`) for `--show-csv`

dvc/command/experiments.py
skshetry: `json` is completely different format from table/csv.

dvc/command/experiments.py
```diff
@@ -399,8 +407,16 @@ def show_experiments(
         include_params=include_params,
         exclude_params=exclude_params,
     )
-    metric_headers = _normalize_headers(metric_names)
-    param_headers = _normalize_headers(param_names)
+
+    names = {**metric_names, **param_names}
+    counter = Counter(
```
karajan1001, line 412: for #5989 
The duplicated column name `foo` in the test stage will cause a wrong output in test.

tests/func/experiments/test_show.py
```diff
@@ -176,84 +177,84 @@ def test_show_checkpoint_branch(
     "i_metrics,i_params,e_metrics,e_params,included,excluded",
     [
         (
-            "foo",
-            "foo",
+            "f",
```
karajan1001, line 180: shorten the name of the parameters, because it is too long to collapse in outputs.

```
┏━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━┳━━━━━━━━━━━━━━━━━┳━━━┓
┃ Experiment ┃ metrics.yaml:foo ┃ metrics.yaml:bar ┃ … ┃ params.yaml:foo ┃ … ┃
┡━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━╇━━━━━━━━━━━━━━━━━╇━━━┩
│ workspace  │                1 │                1 │ 1 │ 1               │ 1 │
│ master     │                1 │                1 │ 1 │ 1               │ 1 │
└────────────┴──────────────────┴──────────────────┴───┴─────────────────┴───┘
```

Actually, in the previous test, we also have a column collapse in some of the test cases. But because there is a duplication in columns, we still get passed.

pmrowla: This makes the test very hard to read, it's probably better to mock/patch the `Console` object used in `dvc.ui` to have a large enough width to fit the full table (instead of the default behavior where rich auto-detects however wide your current terminal session is).

Similar to how we override the width when we use the pager output:
https://github.com/iterative/dvc/blob/dba0aaeab7be3d369d1e41b6d7cd905c04e981af/dvc/ui/table.py#L57

pmrowla (approved):
LGTM, should have @skshetry and @dberenbaum double check

dberenbaum (approved):
🎉

dvc/command/experiments.py
skshetry: @karajan1001, let's do the same for other headers that we have (like `Experiments`, `rev`, etc.) to reduce chances of collision. You can extract the headers from the following to a global variable and reuse them here:

https://github.com/iterative/dvc/blob/b3680b1a9357b206359b70c2753ec6cf9c4e1536/dvc/command/experiments.py#L341-L349

dvc/command/experiments.py
```diff
@@ -274,14 +276,16 @@ def _sort(item):
...
+def _format_time(datetime_obj, fill_value=FILL_VALUE, iso=False):
+    if datetime_obj is None:
+        return fill_value
+    if datetime_obj.date() == date.today():
         fmt = "%I:%M %p"
     else:
         fmt = "%b %d, %Y"
-    return timestamp.strftime(fmt)
+    if iso:
+        return datetime_obj.isoformat()
```
skshetry, line 287: Maybe let's do this after `if datetime_obj is None` branch, as it does not require `fmt`. It's a bit confusing right now.

dvc/command/experiments.py
```diff
@@ -410,7 +423,9 @@ def show_experiments(
         param_names,
         kwargs.get("sort_by"),
         kwargs.get("sort_order"),
-        kwargs.get("precision"),
+        None if show_csv else kwargs.get("precision"),
+        "" if show_csv else FILL_VALUE,
```
skshetry, line 427: How about we move these out of here to the `show_experiments`?

tests/func/experiments/test_show.py
```diff
@@ -269,6 +270,10 @@ def test_show_filter(
     included,
     excluded,
 ):
+    from contextlib import contextmanager
```
skshetry, line 273: For now, I think this workaround of setting console width is okay. But I'd like to see making columns collapse configurable and also get rid of this test with simpler versions, this is testing too much. :)

dvc/command/experiments.py
skshetry: Also, should we rename the column `typ` to `Type`?

skshetry: Also, the lowercase names, should we uppercase it?

tests/func/experiments/test_show.py
```diff
@@ -471,3 +492,47 @@ def test_show_with_broken_repo(tmp_dir, scm, dvc, exp_stage, caplog):
 
     paths = ["workspace", "baseline", "error"]
     assert isinstance(get_in(result, paths), YAMLFileCorruptedError)
+
+
+def test_show_csv(tmp_dir, scm, dvc, exp_stage, capsys):
```
skshetry, line 497: I think it's okay to just test `show_experiments` with fixed `all_experiments` value, and then write a small unit test for the command to check it is passing correct parameters to `show_experiments`,

Also, for the `show_experiments`, perhaps you can just pass captured stdout to `csv.DictReader` and read from that object?

tests/func/experiments/test_show.py
karajan1001: Yeah, but we had mixed too much in this PR, I‘d prefer opening a new `columns collapse configuration` ticket

tests/func/experiments/test_show.py
karajan1001: Too much functions and commands in `dvc.command.experiments`, we should split them into separate files. Currently, it is hard to find a place for the unit test.

tests/func/experiments/test_show.py
skshetry: I am not asking you to work on it, I was just making a note: :)

dvc/command/experiments.py
```diff
@@ -399,8 +409,13 @@ def show_experiments(
         include_params=include_params,
         exclude_params=exclude_params,
     )
-    metric_headers = _normalize_headers(metric_names)
-    param_headers = _normalize_headers(param_names)
+
+    names = {**metric_names, **param_names}
+    counter = Counter(
+        name for path in names for name in names[path] for path in names
```
karajan1001, line 415: $ names = {"a":{"b":1, "c":2}}
$ [name for path in names for name in names[path] for path in names]
> ['b', 'c']

$ names = {"a":{"b":1, "c":2}, "d":{}}
$ [name for path in names for name in names[path] for path in names]
> ['b', 'b', 'c', 'c']

A bug ?

tests/func/experiments/test_show.py
```diff
@@ -471,3 +492,47 @@ def test_show_with_broken_repo(tmp_dir, scm, dvc, exp_stage, caplog):
...
+    rev1 = first(result1)
+    ref_info1 = first(exp_refs_by_rev(scm, rev1))
+    result2 = dvc.experiments.run(exp_stage.addressing, params=["foo=3"])
+    rev2 = first(result2)
+    ref_info2 = first(exp_refs_by_rev(scm, rev2))
+
+    capsys.readouterr()
+    assert main(["exp", "show", "--show-csv"]) == 0
+    cap = capsys.readouterr()
+    print(cap.out)
```
karajan1001, line 515: should remove

karajan1001:
Current exp show like this

```
Experiment,rev,typ,Created,parent,avg_prec,roc_auc,prepare.split,prepare.seed,featurize.max_features,featurize.ngrams,train.seed,train.n_est,train.min_split
,workspace,baseline,,,0.5843640011189556,0.9544670443829399,0.2,20170428,3000,1,20170428,100,36
master,b05eecc,baseline,2021-08-02T16:48:14,,0.5325162867864254,0.9106964878520005,0.2,20170428,3000,1,20170428,100,2
exp-44136,ae99936,branch_commit,2021-08-31T14:56:55,,0.5843640011189556,0.9544670443829399,0.2,20170428,3000,1,20170428,100,36
exp-9fcef,8bc0b4d,branch_commit,2021-08-23T17:43:25,,0.5843640011189556,0.9544670443829399,0.2,20170428,3000,1,20170428,100,36
exp-aaa23,4d5e611,branch_commit,2021-08-23T17:43:17,,0.5950970297562502,0.9554043071921983,0.2,20170428,3000,1,20170428,100,72
exp-8ae22,567c4b8,branch_commit,2021-08-23T17:43:10,,0.6037301973188752,0.9557358950572323,0.2,20170428,3000,1,20170428,100,64
exp-a5f46,358e946,branch_base,2021-08-23T17:43:06,,0.576325443740785,0.955705227971449,0.2,20170428,3000,1,20170428,100,128
```

tests/unit/command/test_experiments.py
```diff
@@ -274,3 +277,128 @@ def test_experiments_remove(dvc, scm, mocker, queue, clear_all):
...
+
+
+def test_show_experiments(capsys):
+    show_experiments(
+        all_experiments, precision=None, fill_value="", iso=True, show_csv=True
+    )
+    cap = capsys.readouterr()
+    assert (
+        "Experiment,rev,typ,Created,parent,State,scores.json:"
+        "featurize.max_features,scores.json:featurize.ngrams,"
```
karajan1001, line 386: In this test, some arguments appear twice in different files, it can be used for test #5989

dvc/command/experiments.py
pmrowla: I think the bug is just that there is an extra `for path in names` in the original expression for some reason.

But what this does is count the number of times a particular metric or param name is used. Since different metrics/params files can use the same names for values (like `pipeline1/params.yaml:foo` and `pipeline2/params.yaml:foo`) we count the number of times that `foo` is used, and if it is once we can just display it as `foo` in the table. If it is used more than once, we have to display the columns using the full `path:foo` header instead.

tests/unit/command/test_experiments.py
karajan1001: Add a duplicated `parent` column in `params.yaml`

dvc/command/experiments.py
karajan1001: > Also, the lowercase names, should we capitalize it?

for the revs and parent I think we should capitalize them, but for the user-defined ones, capitalization might cause confusion to the users, and make them hard to manage in code ( It is easy to capitalize a string but hard to recover it )

skshetry (approved):

karajan1001:
Current edition

```bash
Experiment,Rev,Type,Created,Parent,avg_prec,roc_auc,prepare.split,prepare.seed,featurize.max_features,featurize.ngrams,train.seed,train.n_est,train.min_split
,workspace,baseline,,,0.5843640011189556,0.9544670443829399,0.2,20170428,3000,1,20170428,100,36
master,b05eecc,baseline,2021-08-02T16:48:14,,0.5325162867864254,0.9106964878520005,0.2,20170428,3000,1,20170428,100,2
exp-44136,ae99936,branch_commit,2021-08-31T14:56:55,,0.5843640011189556,0.9544670443829399,0.2,20170428,3000,1,20170428,100,36
exp-9fcef,8bc0b4d,branch_commit,2021-08-23T17:43:25,,0.5843640011189556,0.9544670443829399,0.2,20170428,3000,1,20170428,100,36
exp-aaa23,4d5e611,branch_commit,2021-08-23T17:43:17,,0.5950970297562502,0.9554043071921983,0.2,20170428,3000,1,20170428,100,72
exp-8ae22,567c4b8,branch_commit,2021-08-23T17:43:10,,0.6037301973188752,0.9557358950572323,0.2,20170428,3000,1,20170428,100,64
exp-a5f46,358e946,branch_base,2021-08-23T17:43:06,,0.576325443740785,0.955705227971449,0.2,20170428,3000,1,20170428,100,128
```

tests/func/experiments/test_show.py
```diff
@@ -471,3 +492,46 @@ def test_show_with_broken_repo(tmp_dir, scm, dvc, exp_stage, caplog):
 
     paths = ["workspace", "baseline", "error"]
     assert isinstance(get_in(result, paths), YAMLFileCorruptedError)
+
+
+def test_show_csv(tmp_dir, scm, dvc, exp_stage, capsys):
```
skshetry, line 497: Do we need this test? Can this test be replaced with mock test that checks if `show_experiments` is being called correctly? WDYT? No strong opinion though.

Eg:
https://github.com/iterative/dvc/blob/2b63cedfcd89484bf1ec93e86610e3009b4fa6da/tests/unit/command/test_experiments.py#L61

dvc/command/experiments.py
skshetry: @karajan1001, I was only talking about that particular list: {typ, rev, parent}.

tests/func/experiments/test_show.py
karajan1001: Agreed, In my previous version, what I planned is that we only test `_show_csv` (you asked why I had this function) being called, the value of `all_experiments` (already tested properly), and the `to_csv` function. But I didn't test the code between `all_experiments` and `to_csv` in `show_experiments` in it. And for now as the `show_experiments` had been tested fully, we can just test the call from interface to the `show_experiments`.

dvc/command/experiments.py
karajan1001: @skshetry 
But one problem here, "rev" is consisitent with the some other functions for example in `dvc/repo/plots/template.py`, `./dvc/scm/git/__init__.py` and `./dvc/api.py`
while `typ` and `parent` are only used here. So I will only modify  `typ` and `parent` here.

skshetry: But they are not part of a UI, they are mostly part of an API or a schema where that makes sense. Please check `dvc metrics show --all-commits` for example.

efiop:
@karajan1001 Looks like this is a little flaky:

```
2021-09-13T19:02:32.4081032Z =================================== FAILURES ===================================
2021-09-13T19:02:32.4082061Z ________________________________ test_show_csv _________________________________
2021-09-13T19:02:32.4083946Z [gw0] linux -- Python 3.9.6 /opt/hostedtoolcache/Python/3.9.6/x64/bin/python
2021-09-13T19:02:32.4085135Z 
2021-09-13T19:02:32.4086660Z tmp_dir = PosixTmpDir('/tmp/pytest-of-runner/pytest-0/popen-gw0/test_show_csv0')
2021-09-13T19:02:32.4090688Z scm = Git: '/tmp/pytest-of-runner/pytest-0/popen-gw0/test_show_csv0/.git'
2021-09-13T19:02:32.4092684Z dvc = Repo: '/tmp/pytest-of-runner/pytest-0/popen-gw0/test_show_csv0'
2021-09-13T19:02:32.4093854Z exp_stage = Stage: 'copy-file'
2021-09-13T19:02:32.4094866Z capsys = <_pytest.capture.CaptureFixture object at 0x7f281995ef10>
2021-09-13T19:02:32.4095580Z 
2021-09-13T19:02:32.4096279Z     def test_show_csv(tmp_dir, scm, dvc, exp_stage, capsys):
2021-09-13T19:02:32.4097073Z         baseline_rev = scm.get_rev()
2021-09-13T19:02:32.4097695Z     
2021-09-13T19:02:32.4098326Z         def _get_rev_isotimestamp(rev):
2021-09-13T19:02:32.4099176Z             return datetime.fromtimestamp(
2021-09-13T19:02:32.4100403Z                 scm.gitpython.repo.rev_parse(rev).committed_date
2021-09-13T19:02:32.4101301Z             ).isoformat()
2021-09-13T19:02:32.4101872Z     
2021-09-13T19:02:32.4102739Z         result1 = dvc.experiments.run(exp_stage.addressing, params=["foo=2"])
2021-09-13T19:02:32.4103660Z         rev1 = first(result1)
2021-09-13T19:02:32.4104405Z         ref_info1 = first(exp_refs_by_rev(scm, rev1))
2021-09-13T19:02:32.4105425Z         result2 = dvc.experiments.run(exp_stage.addressing, params=["foo=3"])
2021-09-13T19:02:32.4106363Z         rev2 = first(result2)
2021-09-13T19:02:32.4107086Z         ref_info2 = first(exp_refs_by_rev(scm, rev2))
2021-09-13T19:02:32.4107756Z     
2021-09-13T19:02:32.4108382Z         capsys.readouterr()
2021-09-13T19:02:32.4109421Z         assert main(["exp", "show", "--show-csv"]) == 0
2021-09-13T19:02:32.4110204Z         cap = capsys.readouterr()
2021-09-13T19:02:32.4110886Z         assert (
2021-09-13T19:02:32.4112108Z             "Experiment,rev,typ,Created,parent,metrics.yaml:foo,params.yaml:foo"
2021-09-13T19:02:32.4115556Z             in cap.out
2021-09-13T19:02:32.4116708Z         )
2021-09-13T19:02:32.4117749Z         assert ",workspace,baseline,,,3,3" in cap.out
2021-09-13T19:02:32.4118815Z         assert (
2021-09-13T19:02:32.4119855Z             "master,***,baseline,***,,1,1".format(
2021-09-13T19:02:32.4121030Z                 baseline_rev[:7], _get_rev_isotimestamp(baseline_rev)
2021-09-13T19:02:32.4122034Z             )
2021-09-13T19:02:32.4122852Z             in cap.out
2021-09-13T19:02:32.4123651Z         )
2021-09-13T19:02:32.4124450Z >       assert (
2021-09-13T19:02:32.4125381Z             "***,***,branch_base,***,,2,2".format(
2021-09-13T19:02:32.4126502Z                 ref_info1.name, rev1[:7], _get_rev_isotimestamp(rev1)
2021-09-13T19:02:32.4127487Z             )
2021-09-13T19:02:32.4128288Z             in cap.out
2021-09-13T19:02:32.4129087Z         )
2021-09-13T19:02:32.4132373Z E       AssertionError: assert 'exp-42ced,b84cb50,branch_base,2021-09-13T19:01:43,,2,2' in 'Experiment,rev,typ,Created,parent,metrics.yaml:foo,params.yaml:foo\r\n,workspace,baseline,,,3,3\r\nmaster,9279735,bas...exp-42ced,b84cb50,branch_commit,2021-09-13T19:01:43,,2,2\r\nexp-e0d04,755d134,branch_base,2021-09-13T19:01:43,,3,3\r\n'
2021-09-13T19:02:32.4137095Z E        +  where 'exp-42ced,b84cb50,branch_base,2021-09-13T19:01:43,,2,2' = <built-in method format of str object at 0x7f2903a639e0>('exp-42ced', 'b84cb50', '2021-09-13T19:01:43')
2021-09-13T19:02:32.4139193Z E        +    where <built-in method format of str object at 0x7f2903a639e0> = '***,***,branch_base,***,,2,2'.format
2021-09-13T19:02:32.4141910Z E        +    and   'exp-42ced' = ExpRefInfo(baseline_sha='9279735db68b206e8725ba24c7b28fab99d3b93e', name='exp-42ced').name
2021-09-13T19:02:32.4144543Z E        +    and   '2021-09-13T19:01:43' = <function test_show_csv.<locals>._get_rev_isotimestamp at 0x7f2819b9fd30>('b84cb506452c41f40b30a22d8a43a3b547fa2d0c')
2021-09-13T19:02:32.4149434Z E        +  and   'Experiment,rev,typ,Created,parent,metrics.yaml:foo,params.yaml:foo\r\n,workspace,baseline,,,3,3\r\nmaster,9279735,bas...exp-42ced,b84cb50,branch_commit,2021-09-13T19:01:43,,2,2\r\nexp-e0d04,755d134,branch_base,2021-09-13T19:01:43,,3,3\r\n' = CaptureResult(out='Experiment,rev,typ,Created,parent,metrics.yaml:foo,params.yaml:foo\r\n,workspace,baseline,,,3,3\r\n...36mhttps://man.dvc.org/config#core\x1b[39m>\r                                                                      \r').out
2021-09-13T19:02:32.4152843Z 
2021-09-13T19:02:32.4153848Z /home/runner/work/dvc/dvc/tests/func/experiments/test_show.py:526: AssertionError
```

could you take a look, please?

Issue #6421, opened 2021-08-13
`dvc exp pull <refname>` error: ERROR: unexpected error - 'str' object has no attribute 'baseline_sha'


patzy:
# Bug Report

## Issue name

When there are multiple experiments pushed to a git remote with the same name, trying to use `dvc pull` instructs to use the full refname to decide which one to pull. Doing so outputs an error, although the experiment seems to have been pulled correctly.

## Description

For example:
```console
$ dvc exp pull origin RandomForest
ERROR: Ambiguous name 'RandomForest' refers to multiple experiments in '{git_remote}'. Use full refname to pull one of the following:

	refs/exps/0a/57c2dab7898260bfdecce271f6b7f9cc22cd6b/RandomForest
	refs/exps/8e/96906ff946d98335625c8525d0dc0b54474665/RandomForest
```

Doing so prints an error:
```console
$ dvc exp pull origin refs/exps/0a/57c2dab7898260bfdecce271f6b7f9cc22cd6b/RandomForest
ERROR: unexpected error - 'str' object has no attribute 'baseline_sha'                                                                                                                         
```

But when listing experiments it seems it has been pulled correctly:
```console
$ dvc exp list --all
0a57c2d:                                                              
	RandomForest
```

### Reproduce

1. Create multiple experiments with the same name (this is possible by doing a git commit between experiments).
2. Push experiments as they're created
3. Clone the repo and attempt to pull a named experiment (since there are multiple ones with the same name you'll get the Ambiguous name error and have to pull by refname).
4. Pull experiment using refname

### Expected

There should be no error message ?

Also maybe make the doc more clear about this kind of edge cases ? (Seems to me we should not be able to create another experiment with the same name like this ?)

### Environment information

**Output of `dvc doctor`:**

```console
$ dvc doctor
DVC version: 2.5.4 (pip)
---------------------------------
Platform: Python 3.9.6 on Linux-5.13.7-arch1-1-x86_64-with-glibc2.33
Supports:
	http (requests = 2.26.0),
	https (requests = 2.26.0),
	s3 (s3fs = 2021.6.1, boto3 = 1.17.49),
	ssh (paramiko = 2.7.2)
Cache types: hardlink, symlink
Cache directory: ext4 on /dev/sdb1
Caches: local
Remotes: s3
Workspace directory: ext4 on /dev/sdb1
Repo: dvc, git
```

**Additional Information (if any):**

Full verbose output of the command:
```console
[...]

Issue #5392, opened 2021-02-02
Common UI improvements


dberenbaum:
Goals for UI:

1. Abstract UI implementation to apply high-level functions/classes to all UI components
  - [x] #5347 and #5348 
2. Establish CLI/UI guidelines

3. Improve formatting, style, and functionality

@skshetry @efiop @pmrowla

dberenbaum:
See #5357 and #5348 for abstraction

dberenbaum:
Referenced in https://github.com/iterative/dvc/pull/5282#discussion_r570719275: `Subcommand should usually be plural`

jorgeorpinel:
What about extracting all user messages to a central location for easy access to the prod team to help with? Could even help for future i18n ([more context](https://github.com/iterative/dvc.org/issues/2108#issuecomment-774580721)).

dberenbaum:
See https://github.com/iterative/dvc/discussions/5891 for `--quiet` guidelines

dberenbaum:
@skshetry Are there other issues that can be linked here?

skshetry:
Some update on this - We decided to adopt https://clig.dev as a base guideline for the CLI applications (I would suggest everyone read this). I started going through our commands and fix where we violate this and make some improvements along the way. 

I had to start somewhere, so I have chosen data management commands - starting with `add`.  I noticed some performance and reliability issues on `add`: #6227, #6226, and #6228, which I'll be going to fix, but before I need to do some cleanups.

After that, I'll focus on improving data management commands (`add/commit/remove/move/checkout/destroy`).

skshetry:
Another thing that I'd like to open the discussion is the UI messages for the `add`.

- The `add` command is quiet about what is done, at the end, there's only a progress bar (and, an ugly hint). What should it say on success?
- Do we even need a progress bar for a single target to add (eg: `dvc add data`)?
- Do we need progress bars at all, even for multiple targets?
- How should DVC show progress bars/status on places where it might take a lot of time? eg: recursively adding a large directory, collecting lots of stages, etc? 

To compare (and, spark the discussion, here are the asciinema recordings for multiple targets add and single targets add).

#### Single target
[![asciicast](https://asciinema.org/a/EMl7w6cAF27eobziIXKUC23UV.svg)](https://asciinema.org/a/EMl7w6cAF27eobziIXKUC23UV)

#### Multiple targets
[...]

PR #6471, opened 2021-08-23
Clean up remotes's exps

### LINKED ISSUE #6006: Remove pushed experiment from git remote

karajan1001:
fix #6006 

1. add a new argument `--git-remote` to `dvc exp remove`
2. add some tests for it


karajan1001:
@dberenbaum @pmrowla 
Should I support a more granular remove like in `dvc exp push`. The current version will clean up all of the exps in the remove repo.

dvc/scm/git/backend/dulwich.py
```diff
@@ -399,9 +399,11 @@ def push_refspec(
             ) from exc
 
         def update_refs(refs):
+            from dulwich.objects import ZERO_SHA
+
             new_refs = {}
             for ref, value in zip(dest_refs, values):
-                if ref in refs:
+                if ref in refs and value != ZERO_SHA:
```
karajan1001, line 406: Delete something is also a kind of overwrite. Maybe 

```
if ref in refs:
    local_sha = value
```

Here and add a `--force` flag to it?

is better?

pmrowla: If you are explicitly deleting a remote ref, the assumption is that it already exists, and that you want to delete it. I don't see any reason to need a separate `-f/--force` in this case. (This is the same way that `git push -d` works - you don't need to do `git push -f -d`)

pmrowla: Also, could you clarify why the `and value != ZERO_SHA` check is needed here?

karajan1001: And here what I suggest is that it is better to error out here in `push_refspec` because the `ZERO_SHA` is overwritten to the remote. But we can add the `force=True` flag in the call 

```
repo.scm.push_refspec(remote, None, ref_name)
```
I agreed that we don't need to use `--force` flag here. all of the operations in `dvc exp remove` are dangerous, do not need to remind users.

pmrowla: Pushing the `None` source refspec is an explicit remove operation, it shouldn't require an additional `force=True`.

dberenbaum:
> @dberenbaum @pmrowla
> Should I support a more granular remove like in `dvc exp push`. The current version will clean up all of the exps in the remove repo.

Yes, I think granular removal is needed here. My preferred usage would be to make `--git-remote` act like the existing syntax but operating on the remote git repo instead of the local one:

* `dvc exp remove --git-remote origin abc123` removes experiment `abc123` from `origin`.
* `dvc exp remove --git-remote origin --all` removes all experiments from `origin`.
* `dvc exp remove --git-remote origin --queue` does nothing or throws an error.

dvc/scm/git/backend/dulwich.py
dberenbaum: > I agreed that we don't need to use `--force` flag here. all of the operations in `dvc exp remove` are dangerous, do not need to remind users.

In general, it might be helpful to list the refs found for removal and ask for confirmation by default. That's true for all of `dvc exp remove`, not just with `--git-remote`. I'll open a separate issue for that.

dvc/command/experiments.py
```diff
@@ -1249,6 +1250,12 @@ def add_parser(subparsers, parent_parser):
         action="store_true",
         help="Remove all committed experiments.",
     )
+    remove_group.add_argument(
+        "-r",
+        "--git-remote",
```
jorgeorpinel, line 1255: I don't think we should use `-r` here since that's used for `--remote` in other commands (possibly here later?)
I think argument/flag consistency is a guideline in https://clig.dev/
How about `-g`? or nothing

dvc/command/experiments.py
```diff
@@ -1249,6 +1250,12 @@ def add_parser(subparsers, parent_parser):
...
+        metavar="<git_remote>",
```
jorgeorpinel, line 1256: Should be something non-repetitive. AFAIR we try to use pseudo types here? In this case `<text>` or maybe `<name>`

dvc/command/experiments.py
```diff
@@ -1249,6 +1250,12 @@ def add_parser(subparsers, parent_parser):
...
+        help="Name of the Git remote to GC all of the experiment branches.",
```
jorgeorpinel, line 1257: Why GC? Isn't this `exp remove`?

jorgeorpinel: Also what's the default? `origin`, I assume. Thanks

jorgeorpinel:
Thinking about it, I'm a bit confused about the command semantics. `dvc push/pull` use the default DVC remote or accepts a `--remote`, `exp push/pull/remove` however take a remote as argument. Seems inconsistent.

Maybe the argument these last accept (should be optional) can be the git remote, resembling `git push/pull`. And then use the default DVC remote or take one via `-r`.

dvc/command/experiments.py
karajan1001: Can we use `--remote` here? because `dvc exp remove` act only on refs not caches.

dvc/command/experiments.py
karajan1001: `<remove name or url>` is this too long?

dvc/command/experiments.py
pmrowla: There is no default, if `--git-remote` is not specified it removes a local experiment ref and not a remote one

karajan1001: I think it is better to use "Name or URL of the Git remote to delete all of the experiment references."" 

> Also what's the default? origin, I assume. Thanks
No default for this

pmrowla:
> Thinking about it, I'm a bit confused about the command semantics. `dvc push/pull` use the default DVC remote or accept a `--remote` otherwise. `dvc exp push/pull/remove` however expect a remote as argument. Seems inconsistent.

The mandatory argument for `exp push/pull` is the git remote. DVC data for the experiments is pushed to the default DVC remote if `-r/--remote` is not provided, same as `dvc push/pull`.

> Maybe the argument these last accept should be the git remote (perhaps made optional if the branch has a default upstream), resembling `git push/pull`. And then use the default DVC remote or take one via `-r`. Cc @dberenbaum

Git branch upstream/tracking remotes are specifically only for git branch refs, they don't apply to any other git ref (including experiments). The reason we require explicit git remote arguments is because the "correct" git behavior for "default remotes" will end up being confusing for users.

dvc/command/experiments.py
dberenbaum: I would defer to @jorgeorpinel, but I think `--git-remote` would be better for consistency and clarity.

dvc/command/experiments.py
dberenbaum: We are using `metavar="<git_remote>"` already for other `exp push/pull`. If you think it's unhelpful @jorgeorpinel, can we extract to a separate issue and keep as is for now for consistency and to not block this pr?

dvc/repo/experiments/utils.py
```diff
@@ -33,7 +33,7 @@ def exp_refs_by_name(
 ) -> Generator["ExpRefInfo", None, None]:
     """Iterate over all experiment refs matching the specified name."""
     for ref_info in exp_refs(scm):
-        if ref_info.name == name:
+        if ref_info.name == name or str(ref_info) == name:
```
pmrowla, line 36: If `name` is an absolute ref (meaning it starts with `refs/`) we don't have to do anything to find the matching ref, we can just return the absolute ref.

Doing the string search/loop iteration is potentially slow, which was why we had the `if name.startswith("refs/")` shortcut checks in the calling functions. The bug in the shortcut implementation was just that we were returning the absolute refs as strings and not as `ExpRefInfo`s (https://github.com/iterative/dvc/issues/6421)

Basically we just want something like:
```py
if name.startswith("refs/"):
    try:
        yield ExpRefInfo.from_ref(name)
    except InvalidExpRefError:
        pass
    return
```

pmrowla (changes requested):

dvc/repo/experiments/pull.py
```diff
@@ -36,9 +36,6 @@ def on_diverged(refname: str, rev: str) -> bool:
 
 
 def _get_exp_ref(repo, git_remote, exp_name):
-    if exp_name.startswith("refs/"):
-        return exp_name
-
```
pmrowla, line 41: All the instances of this should not be removed (see my other comment)

dvc/repo/experiments/utils.py
karajan1001: Yes, but I guess for the remote refs, this would not cause any performance issue? Because we didn't initialize the remote repo previously?

karajan1001: Maybe we should rename it to `exp_ref_info_by_ref`?

pmrowla: If the user is providing an absolute `refs/...` string, we can just assume that it exists, either locally or in the remote, and error out later when we try to do whatever we need to with that ref if it turns out to not exist. We don't need to iterate over anything in this case.

It doesn't matter  `ExpRefInfo.from_ref("refs/...")` doesn't matter if it's a local or remote ref, the string is the same either way.

dvc/repo/experiments/remove.py
```diff
@@ -96,17 +97,22 @@ def _get_ref(ref_infos, name, cur_rev) -> Optional[ExpRefInfo]:
     return ref_infos[0]
 
 
-def _remove_commited_exps(repo, refs: List[str]) -> List[str]:
+def _remove_commited_exps(repo, remote: str, refs: List[str]) -> List[str]:
```
pmrowla, line 100: `remote` should be an optional parameter optional here and in `_get_exp_ref`

dvc/repo/experiments/pull.py
```diff
@@ -36,6 +36,9 @@ def on_diverged(refname: str, rev: str) -> bool:
 
 
 def _get_exp_ref(repo, git_remote, exp_name):
+    if exp_name.startswith("refs/"):
+        return exp_name
```
pmrowla:
```suggestion
        return ExpRefInfo.from_ref(exp_name)
```

dvc/repo/experiments/remove.py
```diff
@@ -67,14 +67,21 @@ def _get_exp_stash_index(repo, ref_or_rev: str) -> Optional[int]:
     return None
 
 
-def _get_exp_ref(repo, remote: str, exp_name: str) -> Optional[ExpRefInfo]:
+def _get_exp_ref(
+    repo, remote: Optional[str], exp_name: str
+) -> Optional[ExpRefInfo]:
+    if exp_name.startswith(EXPS_NAMESPACE):
+        if repo.scm.get_ref(exp_name):
```
pmrowla, line 74: I don't think this check is actually necessary, we can just return the refinfo directly. Also, we can just check if it starts with `refs/` and not `refs/exps`. 

Essentially, if the user gives us a non-exp `refs/...` string, it won't match anything in the rest of this function. Likewise, if `get_ref()` fails, it also means that nothing would match in the rest of this function. So it's faster for us to just return the result of `ExpRefInfo.from_ref()` immediately.

karajan1001:
Fixed #6421 and modified a test line for it.

dvc/repo/experiments/utils.py
```diff
@@ -115,3 +115,27 @@ def fix_exp_head(scm: "Git", ref: Optional[str]) -> Optional[str]:
         if name == "HEAD" and scm.get_ref(EXEC_BASELINE):
             return "".join((EXEC_BASELINE, tail))
     return ref
+
+
+def get_exp_ref_list(
+    repo, exp_name: str, git_remote: Optional[str] = None
+) -> List[ExpRefInfo]:
```
pmrowla, line 122: I don't think there will be any use case where we need the list of multiple possible experiments returned by this function. In the event that we wanted multiple matches, we would just use `exp_refs_by_name` directly. If you want to reorganize it this way, this function should probably return a single experiment, and it should raise an appropriate "ambiguous name" error if there is more than one possible matching experiment.

Also, all of the functions in this `utils` module are just doing Git operations, so they take SCM/Git instances as a parameter, rather than a DVC `Repo` instance. It should be the same for this function as well.

To keep parity with the other git libraries we use, it may also make more sense to name it something like `resolve_exp_ref`.

tests/func/experiments/test_remote.py
```diff
@@ -151,7 +151,7 @@ def test_pull(tmp_dir, scm, dvc, git_downstream, exp_stage, use_url):
 
     git_downstream.scm.remove_ref(str(ref_info))
 
-    downstream_exp.pull(remote, str(ref_info))
+    downstream_exp.pull(remote, str(ref_info), pull_cache=True)
```
pmrowla, line 154: Is there a reason for this change? We have a dedicated end-to-end test for pushing and pulling cache (`test_push_pull_cache`)

dvc/repo/experiments/utils.py
pmrowla: This change can be removed, as discussed we should be doing the check and using `from_ref` elsewhere here. This function is specifically for searching for experiments by the `name` segment of an experiment ref

tests/func/experiments/test_remote.py
karajan1001: for the #6421 ,

pmrowla: So that bug only occurred when we pull cache, but it's not obvious why setting `pull_cache` here should affect whether or not using a full refname works or not.

What we really want for #6421 is a unit test that verifies that the get_exp_ref utils function always returns an `ExpRefInfo`, whether it is passed a shortened name or a full `refs/...` string.

dvc/repo/experiments/utils.py
karajan1001: Are there any other files for the common function of exp?

tests/func/experiments/test_remote.py
karajan1001: Yes but there are two problems.
1. the original function had been removed.
2. we already have a type examine for the new function .
So a functional test or no test and let the issue submitter test him/herself.

pmrowla: Type checking isn't a replacement for a unit test, the new function (`resolve_exp_ref`) should get a unit test instead of this `pull_cache` change.

dvc/repo/experiments/utils.py
pmrowla: @karajan1001

tests/func/experiments/test_utils.py
```diff
@@ -0,0 +1,26 @@
...
+    if test_remote:
+        remote = git_upstream.url
+        dvc.experiments.push(remote, ref_info.name)
+
+    if full_name:
+        exp_name = str(ref_info)
+    else:
+        exp_name = ref_info.name
+
+    assert resolve_exp_ref(scm, exp_name, remote).name == ref_info.name
```
pmrowla, line 26:
```suggestion
    assert str(resolve_exp_ref(scm, exp_name, remote)) == str(ref_info)
```

We should be checking that we get the correct ref entirely, not just the name segment.

Actually it would be even better to just define `__hash___()` properly for ExpRefInfo's so that we can just do `if ref_info1 == ref_info2`, but it doesn't have to be done in this PR

pmrowla: Also I think this should really be in it's own unit test (it probably also belongs in `tests/unit/experiments/test_utils.py` since it's not a func test). And it should be tested using SCM only, without a full DVC repo instance. Basically something like

```py
@pytest.mark.parametrize("name_only", [True, False])
def test_resolve_exp_ref(tmp_dir, scm, name_only):
    tmp_dir.scm_gen("foo", "foo", commit="init")
    rev = scm.get_rev()
    ref = "/".join([EXPS_NAMESPACE, "ab", "c123", "foo"])
    scm.gitpython.set_ref(ref, rev)
    ref_info = resolve_exp_ref(scm, "foo" if name_only else ref)
    assert isinstance(ref_info, ExpRefInfo)
    assert str(ref_info) == ref
```

Ideally it should also be parametrized for remotes, to do the setup for testing a remote ref you just need
```
scm.push_refspec(remote, ref, ref)
```
(see https://github.com/iterative/dvc/blob/master/tests/unit/scm/test_git.py#L231 for examples on how to set up the git remote)

dvc/command/experiments.py
jorgeorpinel: Yeah it's a mix in the codebase. At some point I reviewed a lot of metavars and changed their descriptions to metatype annotations but maybe missed some and new ones may be repetitive/unhelpful here. Maybe @skshetry wants to consider this for https://github.com/iterative/dvc/issues/5392 🙂 Up to you guys

dvc/command/experiments.py
jorgeorpinel: > There is no default

What does `exp remove -g 1234567` do? Errors out? If so OK 👍 thanks

dvc/command/experiments.py
```diff
@@ -1249,6 +1250,13 @@ def add_parser(subparsers, parent_parser):
         action="store_true",
         help="Remove all committed experiments.",
     )
+    experiments_remove_parser.add_argument(
+        "-g",
+        "--git-remote",
+        metavar="<git_remote>",
+        help="Name or URL of the Git remote to delete the experiment "
+        "references",
```
jorgeorpinel, line 1258: > Follow-up of https://github.com/iterative/dvc/pull/6471#pullrequestreview-736872715
```suggestion
        help="Name or URL of the Git remote to remove the experiment "
        "from",
```

jorgeorpinel:
> The mandatory argument for exp push/pull is the git remote.

Oh ok thanks @pmrowla I was confused then. That's consistent with `dvc push/pull`.

dvc/command/experiments.py
```diff
@@ -1249,6 +1250,12 @@ def add_parser(subparsers, parent_parser):
         action="store_true",
         help="Remove all committed experiments.",
     )
+    experiments_remove_parser.add_argument(
+        "-g",
+        "--git-remote",
+        metavar="<git_remote>",
+        help="Name or URL of the Git remote to remove the experiment " "from",
```
jorgeorpinel, line 1257:
```suggestion
        help="Name or URL of the Git remote to remove the experiment from",
```

tests/func/experiments/conftest.py
```diff
@@ -102,7 +102,7 @@ def git_upstream(tmp_dir, erepo_dir):
 @pytest.fixture
 def git_downstream(tmp_dir, erepo_dir):
     url = "file://{}".format(tmp_dir.resolve().as_posix())
-    erepo_dir.scm.gitpython.repo.create_remote("upstream", url)
-    erepo_dir.remote = "upstream"
+    erepo_dir.scm.gitpython.repo.create_remote("downstream", url)
+    erepo_dir.remote = "downstream"
```
pmrowla, line 106: This change shouldn't be necessary.

The `git_upstream` fixture creates an git repo that is "upstream" from tmp_dir. So the local `tmp_dir` git repo contains a remote named `upstream` that points to `erepo_dir`. This allows simulating that `tmp_dir` is a local repo, and `erepo_dir` is a server repo.

The `git_downstream` remote creates a repo that is downstream from `tmp_dir`. Meaning that it simulates `tmp_dir` being the "server" repo, and `repo_dir` being a downstream clone of `tmp_dir`. So the git repo in `tmp_dir` contains no remotes at all, and the git repo in `erepo_dir` contains a single remote named `upstream` which points to `tmp_dir`.

tests/unit/repo/experiments/test_utils.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+@pytest.fixture
+def git_upstream(scm, make_tmp_dir, name="origin"):
+    remote_dir = make_tmp_dir("git-remote", scm=True)
+    url = f"file://{remote_dir.resolve().as_posix()}"
+    scm.gitpython.repo.create_remote(name, url)
+    remote_dir.remote = name
+    remote_dir.url = url
+    return remote_dir
```
pmrowla, line 14: Why is this fixture needed? As far as I can tell it's the same as the existing `git_upstream` fixture?

If the issue is just fixture scope/availability, you can move the git fixtures from `tests/func/experiments/conftest` into the global `tests/dir_helpers` module.

(Or just add the appropriate import into `tests/conftest.py`)

tests/conftest.py
```diff
@@ -155,3 +155,28 @@ def pytest_configure(config):
             enabled_remotes.discard(remote_name)
         if enabled:
             enabled_remotes.add(remote_name)
+
+
+@pytest.fixture
+def git_upstream(tmp_dir, erepo_dir, git_dir, request):
```
pmrowla, line 161: It's probably better to put these in `tests/dir_helpers.py` with the other `tmp_dir` related fixtures

karajan1001: I'm just going to push this change.

karajan1001:
If
1. there is no `__init__.py` in the the pytest package ( directory ).
2. Two test file share a same file name.
There would be a failure like this.

```
import file mismatch:
imported module 'test_utils' has this __file__ attribute:
  /home/runner/work/dvc/dvc/tests/unit/repo/experiments/test_utils.py
which is not the same as the test file we want to collect:
  /home/runner/work/dvc/dvc/tests/unit/stage/test_utils.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
```

pmrowla (approved):

pmrowla:
Looks good, thanks for being patient with all the review requests 👍

dvc/command/experiments.py
```diff
@@ -1249,6 +1250,12 @@ def add_parser(subparsers, parent_parser):
         action="store_true",
         help="Remove all committed experiments.",
     )
+    experiments_remove_parser.add_argument(
```
dberenbaum, line 1253: Should this be part of the mutually exclusive group:

```suggestion
    remove_group.add_argument(
```

Using `-g` with `--queue` doesn't make sense. `dvc exp remove --all -g` doesn't seem to do anything, and I think we discussed that `--all` is probably not wanted for now either since it may be dangerous/unexpected in a collaborative repo.

dberenbaum (changes requested):
One minor issue with the arg parsing. Otherwise looks great!

dvc/command/experiments.py
karajan1001: If we will not have any of these kinds of feature requests.

dberenbaum (approved):

PR #6479, opened 2021-08-24
[6478] Fix 401 unauthorized for directories in basic auth protected http remote

### LINKED ISSUE #6478: dvc pull -r http_remote: fails with 401 - unauthorized for directory cache files, when basic auth is set properly
# Bug Report

## Description
When running dvc pull for an HTTP cache protected with basic auth, directory files read cause:
```
ERROR: unexpected error - 401 Client Error: Unauthorized for url: http://<remote_url>/c2/f5094a84cd17aeeca0524bd422b5ab.dir
```

### Reproduce
* Git clone repository with a DVC HTTP remote protected with basic auth
* Repository should have a directory tracked by DVC.
* Configure basic auth
   ```
    dvc remote add <remote_name> http://<remote_url>
    dvc remote modify origin --local auth basic
    dvc remote modify origin --local user <user>
    dvc remote modify origin --local password <password>
    ```
* Run DVC pull -r <remote_name>

### Expected
Pull all the files from the cache, including the directory and its content.

### Environment information
* MacOS 11.4
* DVC 2.6.4 / 2.6.0
* HTTP remote - Private repository on DAGsHub

**Output of `dvc doctor`:**

```console
DVC version: 2.6.4
Platform: Python 3.7.7 on Darwin-20.5.0-x86_64-i386-64bit
Supports:
	gdrive (pydrive2 = 1.7.0),
	hdfs (pyarrow = 2.0.0),
	webhdfs (hdfs = 2.5.8),
	http (requests = 2.24.0),
	https (requests = 2.24.0)
Caches: local
Remotes: http, http
Workspace directory: apfs on /dev/disk1s2s1
Repo: dvc, git
```

The origin of the issue seems to be in file `dvc/objects/tree.py` in the function `load` in line 106.
This line ultimately leads to an HTTP request to be done without any `auth` specified in the `open` method of the class `BaseFileSystem`, when it calls `open_url`

simonlsk:

This should fix #6478 

### DIFF
--- a/dvc/fs/http.py
+++ b/dvc/fs/http.py
@@ -65,9 +65,6 @@ class HTTPFileSystem(BaseFileSystem):  # pylint:disable=abstract-method
                 self.headers.update({self.custom_auth_header: self.password})
         return None
 
-    def _generate_download_url(self, path_info):
-        return path_info.url
-
     @wrap_prop(threading.Lock())
     @cached_property
     def _session(self):
@@ -191,10 +188,12 @@ class HTTPFileSystem(BaseFileSystem):  # pylint:disable=abstract-method
             )
 
     def open(self, path_info, mode: str = "r", encoding: str = None, **kwargs):
-        return super().open(
-            path_info,
-            mode,
-            encoding,
+        from dvc.utils.http import open_url
+
+        return open_url(
+            path_info.url,
+            mode=mode,
+            encoding=encoding,
             auth=self._auth_method(path_info),
             **kwargs,
         )
--- a/dvc/fs/http.py
+++ b/dvc/fs/http.py
@@ -190,6 +190,15 @@ class HTTPFileSystem(BaseFileSystem):  # pylint:disable=abstract-method
                 desc=name or to_info.url,
             )
 
+    def open(self, path_info, mode: str = "r", encoding: str = None, **kwargs):
+        return super().open(
+            path_info,
+            mode,
+            encoding,
+            auth=self._auth_method(path_info),
+            **kwargs,
+        )
+
     @staticmethod
     def _content_length(response) -> Optional[int]:
         res = response.headers.get("Content-Length")
--- a/dvc/utils/http.py
+++ b/dvc/utils/http.py
@@ -24,13 +24,15 @@ def open_url(url, mode="r", encoding=None, **iter_opts):
 
 
 @contextmanager
-def iter_url(url, chunk_size=io.DEFAULT_BUFFER_SIZE):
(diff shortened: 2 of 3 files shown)


efiop:
@simonlsk Looks like it is breaking our tests, could you take a look, please? 

@isidentical FYI

simonlsk:
@efiop, sorry about that, that should be fixed now.

dvc/fs/http.py
```diff
@@ -190,6 +190,15 @@ def _upload(
...
 
+    def open(self, path_info, mode: str = "r", encoding: str = None, **kwargs):
+        return super().open(
+            path_info,
+            mode,
+            encoding,
+            auth=self._auth_method(path_info),
+            **kwargs,
+        )
+
```
efiop, line 201: Looks like our httpfs is still using an old `_generate_download_url` approach, which we are getting rid of in favor of just using fsspec-compatible filesystems. So to be correct, we just need to move stuff from `BaseFileSystem.open` into `HTTPFileSystem.open` here, so we won't need to do this confusing `super()` call at all. @simonlsk Could you adjust that, please?

dvc/fs/http.py
```diff
@@ -190,6 +190,15 @@ def _upload(
...
             )
 
+    def open(self, path_info, mode: str = "r", encoding: str = None, **kwargs):
+        return super().open(
+            path_info,
+            mode,
+            encoding,
+            auth=self._auth_method(path_info),
+            **kwargs,
+        )
```
efiop, line 200:
```suggestion
            return open_url(
                path_info.url,
                mode=mode,
                encoding=encoding,
                auth=self._auth_method(path_info),
                **kwargs,
            )
```

efiop: And we can remove `if hasattr(self, "_generate_download_url"):` block from `BaseFileSystem.open` and we can also remove `_generate_download_url` everywhere (http was the only one still using it)

dvc/utils/http.py
```diff
@@ -24,13 +24,15 @@ def open_url(url, mode="r", encoding=None, **iter_opts):
 
 
 @contextmanager
-def iter_url(url, chunk_size=io.DEFAULT_BUFFER_SIZE):
+def iter_url(url, auth=None, chunk_size=io.DEFAULT_BUFFER_SIZE):
```
efiop, line 27:
```suggestion
def iter_url(url, chunk_size=io.DEFAULT_BUFFER_SIZE, auth=None):
```

efiop:
@simonlsk Thanks for the PR! :pray: Looks good! Please see a few minor comments above.

efiop (approved):
Thank you! :pray:

Issue #6423, opened 2021-08-13
See how much new version of python-benedict impacts us


skshetry:
The recent release of `python-benedict` has monkeypatched `JSONEncoder` to always use python-based encoder (instead of C-based encoder.

This monkeypatching happens during import.
https://github.com/iterative/dvc/blob/a7f01b65a5cd1654b74e4aa5651063f2b118831d/dvc/utils/collections.py#L85

As we only use it during `exp run --set-params`, we need to see how much it affects us. Also how important is the performance for the `exp run`? Do we save large directories and files during `exp run`?

---

Note: Simple benchmark shows it causes 2x-4x degradation.

efiop:
@skshetry Should we put an upper cap for it for now?

also, for the record: this is clearly +1 to the microbenchmarks that you've proposed before.

skshetry:
µbenchmarks do show a 2-4x slowdown, but it still feels theoretical. We lack benchmarks for pipelines, so it's hard to tell.

skshetry:
Discussed with @efiop to get rid of the dependency. We need to see what it'll take to replace it.

pmrowla:
May be worth revisiting https://github.com/iterative/dvc/issues/4883

mattlbeck:
Benedict's merge module appears to be fairly self-contained /blob/c98c471065ae84b4752a87f1bd63fe3987783663/benedict/core/merge.py . A possible stopgap could be to remove the bendict import and copy this module into the DVC codebase instead. I have done something similar for keypath parsing currently in #6521 (8d219dc4c4be678f8e12a57f766d282d13997443).

skshetry:
I think I'm more inclined on working around benedict than replacing it. We should rollback this particular change:

https://github.com/fabiocaccamo/python-benedict/blob/master/benedict/dicts/__init__.py#L283-L285

daavoo:
It is worth having a dependency that requires workarounds if we only use ~20 lines of that dependency?

Considering it causes other side effects/bugs like https://github.com/iterative/dvc/issues/6476

skshetry:
@mattlbeck, we still need to unflatten those and it comes from [`keylist_util.py`](https://github.com/fabiocaccamo/python-benedict/blob/c98c471065ae84b4752a87f1bd63fe3987783663/benedict/dicts/keylist/keylist_util.py). 

I have not been able to work on this. Quicker way is to workaround benedict, that should help us move ahead #6476 and #6521.

PR #6484, opened 2021-08-24
Added `test_set_empty_params`

### LINKED ISSUE #6476: exp run: unexpected error - cannot represent an object: {'lr': 0.1}
# Bug Report

## Description

I have updated DVC version from 2.5.4 to latest version 2.6.4.

I am running my experiments with `dvc exp run`. I use -S option to use different parameters between experiments. Following the [get-started example](https://github.com/iterative/example-get-started/blob/master/src/prepare.py#L10) I load my params.yaml in my script and update my model config programatically. 

If my `params.yaml` is emtpy and I run `dvc exp run -S optimizer.lr=0.1`:

- `dvc 2.5.4`: experiment runs without problems
- `dvc 2.6.4`: experiment crashes returning a confusing error:` ERROR: unexpected error - cannot represent an object: {'lr': 0.1}`

### Reproduce

1. dvc init
2. Copy dataset.zip to the directory
3. dvc add dataset.zip
4. create empty params.yaml
5. dvc exp run -S optimizer.lr=0.1

-->

### Expected

Experiment runs without returning any error.

### Environment information

**Output of `dvc doctor`:**

```console
$ dvc doctor
DVC version: 2.6.4 (pip)
Platform: Python 3.8.10 on Linux-4.15.0-96-generic-x86_64-with-glibc2.10
Supports:
        gdrive (pydrive2 = 1.9.1),
        http (requests = 2.26.0),
        https (requests = 2.26.0)
Caches: local
Remotes: None
Repo: dvc, git
```

I wouldn't like to have to set my `params.yaml` manually. This would force to replicate my config twice. Most DL frameworks have config files where model description and training hyperparameters are stored. So I would like to update those values simply using the -S flag.

daavoo:

- Added explicit test for https://github.com/iterative/dvc/issues/5477
- Added regression test for  #6476

### DIFF
--- a/dvc/utils/serialize/_yaml.py
+++ b/dvc/utils/serialize/_yaml.py
@@ -2,7 +2,6 @@ import io
 from collections import OrderedDict
 from contextlib import contextmanager
 
-from benedict.dicts import benedict
 from funcy import reraise
 
 from ._common import ParseError, _dump_data, _load_data, _modify_data
@@ -51,7 +50,6 @@ def _get_yaml():
     # tell Dumper to represent OrderedDict as normal dict
     yaml_repr_cls = yaml.Representer
     yaml_repr_cls.add_representer(OrderedDict, yaml_repr_cls.represent_dict)
-    yaml_repr_cls.add_representer(benedict, yaml_repr_cls.represent_dict)
 
     return yaml
 
--- a/dvc/utils/serialize/_yaml.py
+++ b/dvc/utils/serialize/_yaml.py
@@ -50,7 +50,6 @@ def _get_yaml():
     # tell Dumper to represent OrderedDict as normal dict
     yaml_repr_cls = yaml.Representer
     yaml_repr_cls.add_representer(OrderedDict, yaml_repr_cls.represent_dict)
-
     return yaml
 
 
--- a/dvc/utils/serialize/_yaml.py
+++ b/dvc/utils/serialize/_yaml.py
@@ -1,8 +1,8 @@
 import io
 from collections import OrderedDict
 from contextlib import contextmanager
-from benedict.dicts import benedict
 
+from benedict.dicts import benedict
 from funcy import reraise
 
 from ._common import ParseError, _dump_data, _load_data, _modify_data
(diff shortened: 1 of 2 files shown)


dvc/utils/serialize/_yaml.py
```diff
@@ -50,6 +51,8 @@ def _get_yaml():
     # tell Dumper to represent OrderedDict as normal dict
     yaml_repr_cls = yaml.Representer
     yaml_repr_cls.add_representer(OrderedDict, yaml_repr_cls.represent_dict)
+    yaml_repr_cls.add_representer(benedict, yaml_repr_cls.represent_dict)
```
skshetry, line 54: benedict should never escape when we merge params from `--set-params`. This may result in loss of commits. 
A fix might be to merge the params in the file recursively from the `--set-params` benedict dictionary.

daavoo: Not sure I'm fully understanding, what do you mean by `escape` ?

daavoo: Removed that and replaced with just a regression test currently checking that a exception is being raised. I guess that if we go with https://github.com/iterative/dvc/issues/6423#issuecomment-908028959 we should make that test pass

skshetry: #6476 is more urgent than #6423, though we could do it together too.

dvc/utils/serialize/_yaml.py
```diff
@@ -50,6 +50,7 @@ def _get_yaml():
     # tell Dumper to represent OrderedDict as normal dict
     yaml_repr_cls = yaml.Representer
     yaml_repr_cls.add_representer(OrderedDict, yaml_repr_cls.represent_dict)
+
```
daavoo, line 53:
```suggestion
```

dvc/utils/serialize/_yaml.py
daavoo: I'm not sure but maybe we could add this tests for now and discuss in https://github.com/iterative/dvc/pull/6521 how we want to handle #5477 (as it would make #6476 no longer a bug but a different exception)

skshetry:
Sorry for the late review. I have incorporated this test in https://github.com/iterative/dvc/pull/6706#discussion_r718207470, as it gets automatically fixed after updating benedict to the latest version. This bug happened because we were handling an empty dictionary in a different way because of a bug in benedict which we don't need to handle now at all.

PR #6486, opened 2021-08-24
[WIP] dvcfile: support remote per output

efiop:
Related to #2095

### DIFF
--- a/dvc/output.py
+++ b/dvc/output.py
@@ -81,6 +81,7 @@ def loadd_from(stage, d_list):
         desc = d.pop(Output.PARAM_DESC, False)
         isexec = d.pop(Output.PARAM_ISEXEC, False)
         live = d.pop(Output.PARAM_LIVE, False)
+        remote = d.pop(Output.PARAM_REMOTE, None)
         ret.append(
             _get(
                 stage,
@@ -94,6 +95,7 @@ def loadd_from(stage, d_list):
                 desc=desc,
                 isexec=isexec,
                 live=live,
+                remote=remote,
             )
         )
     return ret
@@ -109,6 +111,7 @@ def loads_from(
     checkpoint=False,
     isexec=False,
     live=False,
+    remote=None,
 ):
     return [
         _get(
@@ -122,6 +125,7 @@ def loads_from(
             checkpoint=checkpoint,
             isexec=isexec,
             live=live,
+            remote=remote,
         )
         for s in s_list
     ]
@@ -162,7 +166,6 @@ def load_from_pipeline(stage, data, typ="outs"):
     metric = typ == stage.PARAM_METRICS
     plot = typ == stage.PARAM_PLOTS
     live = typ == stage.PARAM_LIVE
-
     if live:
         # `live` is single object
         data = [data]
@@ -185,6 +188,7 @@ def load_from_pipeline(stage, data, typ="outs"):
                 Output.PARAM_CACHE,
                 Output.PARAM_PERSIST,
                 Output.PARAM_CHECKPOINT,
+                Output.PARAM_REMOTE,
             ],
         )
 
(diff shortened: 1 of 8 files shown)


dberenbaum:
🔥 

How does this work with other `dvc push` flags, and can we test those? Specifically:
* `dvc push -r`: does `-r` override the file-specific remote?
* `dvc push --run-cache`: what happens with files that need to be pushed to reconstruct results from the run-cache?

dvc/output.py
```diff
@@ -843,7 +849,8 @@ def get_dir_cache(self, **kwargs):
         try:
             objects.check(self.odb, obj)
         except FileNotFoundError:
-            self.repo.cloud.pull([obj.hash_info], **kwargs)
+            remote = self.repo.cloud.get_remote_odb(self.remote)
+            self.repo.cloud.pull([obj.hash_info], odb=remote, **kwargs)
```
skshetry, line 853: We need support for better error messages, like where the remote has been specified in case of `NoRemoteError` or `RemoteNotFound`.

dvc/output.py
```diff
@@ -185,6 +188,7 @@ def load_from_pipeline(stage, data, typ="outs"):
                 Output.PARAM_CACHE,
                 Output.PARAM_PERSIST,
                 Output.PARAM_CHECKPOINT,
+                Output.PARAM_REMOTE,
```
skshetry, line 191: Would there be a need for a stage-level remote?

efiop: Maybe in the future, but not for now.

skshetry: okay, I was just wondering if it makes sense in the stage-level, or should there be a different keyword there?

skshetry: `push_to: <remote>` or something.

efiop: It might be that we'll need to support pushing/pulling granularly in the future, but for now this is a symmetrical. Interesting to note is that our imports are kinda like `pull: orig_remote, push: false`, since they are pulled from external repo remote but not pushed anywhere by default. So that might be a better way to handle https://github.com/iterative/dvc/pull/4581 (e.g. `push: mybackup` or just `true` for default remote). But that's beyond the scope for this PR. Or do you see where `remote: ` might get problematic in the future?

efiop:
> dvc push -r: does -r override the file-specific remote?

No, it doesn't. This is similar to how `-r` doesn't override remote for `dvc import`-ed artifacts. So the order is:

1. `remote: ` per output
2. `-r`
3. `core.remote` from config

> dvc push --run-cache: what happens with files that need to be pushed to reconstruct results from the run-cache?

So we've talked about --run-cache before and it is a little special. It is only pushed/pulled with an explicit `--run-cache` flag, so there is less of a chance of accidentally pushing data. But still, this is a risk. The problem here is that regular `pushing/pulling` of artifacts and run-cache are really different and somewhat separate things, so I think that maybe what we really need here is a stage-level run-cache options to be able to turn off run-cache (similar to `always_changed`) explicitly, to specify if it needs to be pushed or if it is allowed to be pulled and stuff like that. E.g.

```
run-cache: false/true
```
```
run-cache:
    push: false/myremote/etc
    pull: false/myremote/etc   # could be useful for slow speeds or giant artifacts that are faster to generate locally. Currently if one uses `dvc repro --pull` , we will download artifacts no matter what, so this option could be useful.
```
etc. A bit worried about overengineering this one, need to think a bit more about it, but it is obvious that needs to be handled in a special way.

shcheklein:
@efiop could you please provide a brief summary in the description on how will it look like (or docs PR?)? Otherwise it makes it hard to review for folks who are not familiar with the code base.

efiop:
@shcheklein Sure, this is WIP still, so I didn't add it yet. Will do.

PR #6493, opened 2021-08-26
Added `GitAuthError`

daavoo:

To match updates in docs. From https://github.com/iterative/dvc.org/pull/2731#pullrequestreview-737914588


pmrowla:
> Even better would be to prompt for username and password and try again. Is there a reason we wouldn't want to do that?

Using promps won't work in automated contexts (like the CML auto-push feature) but it would be possible for us to support it. The main issue will just be implementing proper support for git HTTP auth (see dulwich tests for examples: https://github.com/dulwich/dulwich/blob/76dd8356f9d8870f44226e4232c24fbc1e573e93/dulwich/tests/test_client.py#L565)

But it should be noted that even if we add the prompts and username/password HTTP support for git operations, it won't work for pushing to github (https://github.blog/2020-12-15-token-authentication-requirements-for-git-operations/)

dberenbaum:
> Using prompts won't work in automated contexts (like the CML auto-push feature) but it would be possible for us to support it. 

Right, if we make it easier in non-automated scenarios, there could be confusion if people try to migrate to something like CML. Hopefully they would realize any manual steps would need to be automated? Is it better to give flexibility or require the preferred workflow? It seems like generally we try to let users make these decisions themselves. On the other hand, we don't prompt for remote credentials, but that may be because it would get too complex to handle.

> The main issue will just be implementing proper support for git HTTP auth (see dulwich tests for examples: https://github.com/dulwich/dulwich/blob/76dd8356f9d8870f44226e4232c24fbc1e573e93/dulwich/tests/test_client.py#L565)

Can we do this:
1. Catch the `HTTPUnauthorized` exception
2. Prompt for the username and password
3. Recreate the git client with those keyword args
4. Retry
5. Throw if there is still an exception
 
Does dvc need to worry about the details of http auth beyond asking for username and password?

> But it should be noted that even if we add the prompts and username/password HTTP support for git operations, it won't work for pushing to github (https://github.blog/2020-12-15-token-authentication-requirements-for-git-operations/)

From a quick test, it looks like users should be able to push to github with http if they replace their passwords with their tokens.

daavoo:
> Can we do this:
> 
>     1. Catch the `HTTPUnauthorized` exception
> 
>     2. Prompt for the username and password
> 
>     3. Recreate the git client with those keyword args
> 
>     4. Retry
> 
>     5. Throw if there is still an exception
> 
> 
> Does dvc need to worry about the details of http auth beyond asking for username and password?

This would be prompted on each command run (i.e. `dvc exp push`), right? I feel like that could be somehow annoying.

Not sure if that would be a good user experience or if it would be more helpful to just raise an exception with a link to the docs.

dvc/repo/experiments/base.py
```diff
@@ -95,6 +96,15 @@ def __init__(self, name):
         self.name = name
 
 
+class GitRemoteAuthError(GitAuthError):
```
pmrowla, line 99: This seems like it might not belong in `experiments`?

Maybe `scm.base.GitAuthError` should just have the more detailed error message by default, and then rather than having the per-command error link and a repeated "git auth requirements" docs section in every `exp` command that does remote git operations, we could just link to a single troubleshooting entry (https://dvc.org/doc/user-guide/troubleshooting) page regarding git auth instead?

daavoo: Makes sense.

dberenbaum:
> This would be prompted on each command run (i.e. dvc exp push), right? I feel like that could be somehow annoying.
>
> Not sure if that would be a good user experience or if it would be more helpful to just raise an exception with a link to the docs.

It could be annoying, but is it any more annoying than doing `git push` via HTTP? SSH seems like the way to go, but I worry that it will cause trouble for some users (for example, if SSH port is blocked).

We could start with the warning and keep the HTTP authorization open as a separate issue to see if users do encounter it (or wait to see if someone opens that issue themselves).

skshetry:
For autopush, we may need some kind of memoization like the way we do for http(s) filesystems.

https://github.com/iterative/dvc/blob/6b582bd4d9428c4c3152ea502aff12f3f092e745/dvc/fs/http.py#L12-L18

dberenbaum:
> Not sure if there is a way of synthesizing those cases for the CLI help or if we should just say `any valid URL` and leave the detailed info for the docs.

Is "any valid URL" correct? It seems more like it's "any supported URL" since I think the HTTP URL is also valid according to Git. I also think we could probably summarize behavior as not supporting HTTP authentication.  Cc @jorgeorpinel

dvc/command/experiments.py
```diff
@@ -1112,9 +1112,9 @@ def add_parser(subparsers, parent_parser):
         nargs="?",
         default=None,
         help=(
-            "Optional Git remote name or Git URL. If provided, experiments "
-            "from the specified Git repository will be listed instead of "
-            "local experiments."
+            "Optional Git remote name or a valid Git URL. If provided, "
+            "experiments from the specified Git repository will be "
+            "listed instead of local experiments."
```
jorgeorpinel, line 1117: Small copy edit
```suggestion
            "Optional Git remote name or valid Git URL. If provided, "
            "experiments from the specified repository will be "
            "listed instead of local ones."
```

dvc/command/experiments.py
```diff
@@ -1168,7 +1168,7 @@ def add_parser(subparsers, parent_parser):
     )
     experiments_push_parser.add_argument(
         "git_remote",
-        help="Git remote name or Git URL.",
+        help="Git remote name or a valid Git URL.",
```
jorgeorpinel, line 1171:
```suggestion
        help="Git remote name or valid Git URL",
```

dvc/command/experiments.py
```diff
@@ -1224,7 +1224,7 @@ def add_parser(subparsers, parent_parser):
     )
     experiments_pull_parser.add_argument(
         "git_remote",
-        help="Git remote name or Git URL.",
+        help="Git remote name or a valid Git URL.",
```
jorgeorpinel, line 1227:
```suggestion
        help="Git remote name or valid Git URL",
```

jorgeorpinel (commented):
💅 

> See also https://github.com/iterative/dvc.org/pull/2731

dvc/scm/base.py
```diff
@@ -47,6 +47,15 @@ def __init__(self, url: str):
         super().__init__(msg)
 
 
+class GitAuthError(SCMError):
+    def __init__(self, url: str):
+        super().__init__(
+            f"Invalid authentication for Git Remote: '{url}'"
```
jorgeorpinel, line 53:
```suggestion
            f"HTTP Git authentication id not supported: '{url}'"
```

jorgeorpinel:
> Is "any valid URL" correct? It seems more like it's "any supported URL"
summarize behavior as not supporting HTTP authentication

Good points @dberenbaum. I updated my suggestions ☝️

daavoo:
Need to add tests catching the exception

dberenbaum:
Pending discussion in https://github.com/iterative/dvc.org/pull/2731#discussion_r705593107.

karajan1001:
So when is it ready? Still a draft one?

pmrowla (approved):

jorgeorpinel (approved):

daavoo:
> So when is it ready? Still a draft one?

I guess that we can now consider this ready (docs PR was also approved https://github.com/iterative/dvc.org/pull/2731#issuecomment-917312199) and have https://github.com/iterative/dvc/issues/6586 as a follow-up.

PR #6501, opened 2021-08-30
machine: add `dvc machine ssh`

### LINKED ISSUE #6483: machine: provide SSH shell access to created instances
`dvc machine ssh` should drop users into an ssh session on the remote instance (similar to `docker-machine ssh` or `vagrant ssh`).

Since we default to relying on terraform to auto-generate SSH keys, there is no convenient way for users to ssh into the remote machine (unless they manually read `tfstate` to get the private key, write their own PEM file, and use the correct `ssh ...` flags). DVC should automate all of that for the user.

pmrowla:

Will close #6483

* Adds `dvc machine ssh` command to connect to `dvc machine create`ed remote machine instances - shell/PTY session is done via asyncssh and does not require the user to have a separate OpenSSH/putty/etc SSH client on their machine (meaning it should work from Windows hosts out of the box as long as `dvc[terraform]` is installed)
* Adds `ssh` requirements to the `terraform` install extras - eventually `ssh` may need to be added to the default installed remotes if/when `dvc machine` has basic functionality independent of terraform

### DIFF
--- a/tests/unit/machine/test_terraform.py
+++ b/tests/unit/machine/test_terraform.py
@@ -1,5 +1,6 @@
 import json
 import os
+from unittest.mock import ANY
 
 import pytest
 from python_terraform import IsFlagged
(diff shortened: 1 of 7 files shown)


tests/unit/machine/test_terraform.py
```diff
@@ -127,3 +128,27 @@ def test_instances(tmp_dir, terraform, resource):
...
     assert list(terraform.instances(resource)) == [expected]
+
+
+# NOTE: native MagicMock doesn't support async context managers in Python < 3.8
+class ConnectionMock(MagicMock):
+    async def __aenter__(self, *args, **kwargs):
+        return MagicMock()
+
+    async def __aexit__(self, *args, **kwargs):
+        return MagicMock()
```
skshetry, line 139: You might be able to use `mock` package, which should have this backported.
```python

import mock

mocker.patch(
    "asyncssh.connect", return_value=mock.MagicMock()
)

```

PR #6521, opened 2021-09-01
exp run: throw error if new parameters are added by --set-param

### LINKED ISSUE #5477: exp run: validate params filename and parameter names

mattlbeck:
Throws an error if `dvc run exp --set-param` references a parameter key path
that is not in the target config file.

Resolves #5477 

### DIFF
--- a/tests/func/experiments/test_experiments.py
+++ b/tests/func/experiments/test_experiments.py
@@ -406,7 +406,7 @@ def test_no_scm(tmp_dir):
 @pytest.mark.parametrize("workspace", [True, False])
 def test_untracked(tmp_dir, scm, dvc, caplog, workspace):
     tmp_dir.gen("copy.py", COPY_SCRIPT)
-    tmp_dir.gen("params.yaml", "foo: 1")
+    tmp_dir.scm_gen("params.yaml", "foo: 1", commit="track params")
     stage = dvc.run(
         cmd="python copy.py params.yaml metrics.yaml",
         metrics_no_cache=["metrics.yaml"],


mattlbeck:
This is now the minimum required to resolve the main topic of #5477 . I expect there should be some additional change to allow setting new parameters via `exp run` if the user explicitly wants to do this e.g. downgrade to a warning or use an additional flag. This is still up for discussion.

mattlbeck:
Just noticed #6423 . I will have a think about other ways to implement the keypath set operation that doesn't use benedict.

tests/func/experiments/test_experiments.py
```diff
@@ -399,9 +418,7 @@ def test_untracked(tmp_dir, scm, dvc, caplog, workspace):
 
     # copy.py is untracked
     with caplog.at_level(logging.ERROR):
-        results = dvc.experiments.run(
-            stage.addressing, params=["foo=2"], tmp_dir=True
-        )
+        results = dvc.experiments.run(stage.addressing, tmp_dir=True)
```
mattlbeck, line 421: This needs careful review to make sure I am not invalidating this test.

I don't believe this call to exp run needs to be modifying the parameter as there appears to be no params file at the time of running (thus throws the new error unexpectedly) and the file is checked afterwards, as it is in the second exp run in the test. But there could be something to the test that I am not understanding.

dvc/repo/experiments/__init__.py
```diff
@@ -350,6 +350,34 @@ def _pack_args(self, *args, **kwargs):
...
+        b_src = benedict(src)
+
+        return list(set(to_update.keys()) - set(b_src.keypaths(indexes=True)))
+
+    def _format_new_params_msg(self, new_params, config_path):
+        """Format an error message for when new parameters are identified"""
+        new_param_count = len(new_params)
+        additional_msg = "is"
+        if new_param_count > 1:
+            pluralise = "s are" if new_param_count > 2 else " is"
```
karajan1001, line 372: I think here we have at least 2 params, it is always `are` here

dvc/repo/experiments/__init__.py
```diff
@@ -350,6 +350,34 @@ def _pack_args(self, *args, **kwargs):
...
+            additional_msg = (
+                f"and {new_param_count-1} other " f"parameter{pluralise}"
```
karajan1001, line 374: Maybe list all of the new_params are better? `param_a,param_b,param_c,...`

karajan1001:
Any conclusions?

mattlbeck:
I am torn between adding a new `--add-param` flag or just downgrading to a warning. Would love @dberenbaum's input befor proceeding.

daavoo:
> Just noticed #6423 . I will have a think about other ways to implement the keypath set operation that doesn't use benedict.

Do you see this viable (implementing without using benedict)? As commented in https://github.com/iterative/dvc/pull/6484#discussion_r696278507 adding another place where benedict is used could be a drawback

dberenbaum:
Thanks for taking this on yourself @mattlbeck! We haven't seen a great use case yet for using `--set-param` to add a new parameter, and I'm not in favor of adding more flags unless they add obvious value since this command is busy. If an error is more helpful than a warning for CI or greater visibility, then we can leave the usage as is and revisit if users find it overly strict (assuming the implementation questions like benedict usage get resolved of course).

mattlbeck:
@dberenbaum sounds very sensible. I would personally prefer an error too - will leave this as it's current implementation and move on to address the earlier feedback.

mattlbeck:
> > Just noticed #6423 . I will have a think about other ways to implement the keypath set operation that doesn't use benedict.
> 
> Do you see this viable (implementing without using benedict)? As commented in [#6484 (comment)](https://github.com/iterative/dvc/pull/6484#discussion_r696278507) adding another place where benedict is used could be a drawback

Commit 8d219dc4c4be678f8e12a57f766d282d13997443 adapts benedict's keypath parsing into the dvc codebase. I have taken the strategy of minimising the amount of code I copy from benedict, so instead of constructing a list of keypaths from the src dict I parse the incoming key paths and test them against the src dict. This may open up some corner cases that I haven't thought of yet...

mattlbeck:
Does this change need a mention in the docs?

dberenbaum:
> Does this change need a mention in the docs?

Yes, see the note in https://dvc.org/doc/user-guide/experiment-management/running-experiments#updating-experiment-parameters-on-the-fly. Thanks for checking!

dvc/repo/experiments/__init__.py
```diff
@@ -362,6 +401,10 @@ def _update_params(self, params: dict):
             suffix = path.suffix.lower()
             modify_data = MODIFIERS[suffix]
             with modify_data(path, fs=self.repo.fs) as data:
+                new_params = self._get_new_params(data, params[params_fname])
+                if new_params:
+                    error_msg = self._format_new_params_msg(new_params, path)
+                    raise DvcException(error_msg)
```
daavoo, line 407: Rasing a new exception created for this case (`ParamDoesNotExistError`?) or an existing one (`MissingParamsError`? from `dvc.dependency.param`) would be probably better than raising the generic `DvcException`.

dvc/repo/experiments/__init__.py
```diff
@@ -350,6 +352,44 @@ def _pack_args(self, *args, **kwargs):
...
+                subdict = subdict[key]
+            except (KeyError, IndexError, TypeError) as e:
+                logger.debug(e)
+                return True
+        return False
+
+    def _format_new_params_msg(self, new_params, config_path):
+        """Format an error message for when new parameters are identified"""
+        new_param_count = len(new_params)
+        pluralise = "s were" if new_param_count > 1 else " was"
```
jorgeorpinel, line 386:
```suggestion
        pluralise = "s are" if new_param_count > 1 else " is"
```

dvc/repo/experiments/__init__.py
```diff
@@ -350,6 +352,44 @@ def _pack_args(self, *args, **kwargs):
...
+        param_list = ", ".join(new_params)
+        return (
+            f"{new_param_count} parameter{pluralise} missing "
+            f"from {config_path}: {param_list}"
```
jorgeorpinel, line 390:
```suggestion
            f"from '{config_path}': {param_list}"
```

mattlbeck:
@dberenbaum I think the PR is just awaiting a further review after having addressed the feedback

skshetry:
@mattlbeck, let's just use `benedict` for now, we'll look for better alternatives later.

skshetry:
Hi, @mattlbeck, let's use `benedict`. But you may need to import from `dvc.collections._benedict` like done here:

dvc/repo/experiments/__init__.py
```diff
@@ -362,6 +387,10 @@ def _update_params(self, params: dict):
             suffix = path.suffix.lower()
             modify_data = MODIFIERS[suffix]
             with modify_data(path, fs=self.repo.fs) as data:
+                new_params = self._get_new_params(data, params[params_fname])
+                if new_params:
+                    error_msg = self._format_new_params_msg(new_params, path)
+                    raise MissingParamsError(error_msg)
```
skshetry, line 393: @mattlbeck, could you please move this to `merge_params`? That way, you can just test it via unittest, though you may need to adjust existing ones. Although, we could extend `merge_params`'s API with `only_update=True|False`.

dvc/repo/experiments/__init__.py
```diff
@@ -350,6 +351,30 @@ def _pack_args(self, *args, **kwargs):
...
+
+        src is a traditionally nested dict structure representing a parameter
+        config file. to_update is a flat dict with keypath keys representing
+        param names.
+        """
+        from dvc.utils._benedict import benedict
+
+        b_src = benedict(src)
+
+        return list(set(to_update.keys()) - set(b_src.keypaths(indexes=True)))
```
skshetry, line 366: Does this work for `optimizer.lr=0.01` like nested-parameters?

tests/func/experiments/test_experiments.py
pmrowla: This test is checking for the behavior that happens when any important files are untracked (completely new file from git's perspective). In this case, with the new behavior we end up hitting the error for having a new parameter first (before we get to the relevant checks for the untracked files behavior).

It would be better to revert this change and also generate `params.yaml` as a git-tracked file from the start (so that the params existence is irrelevant for this test).

So these changes should be reverted, and line 409 should become:

```py
    tmp_dir.scm_gen("params.yaml", "foo: 1", commit="track params")
```

pmrowla (changes requested):

dvc/repo/experiments/__init__.py
jorgeorpinel: Thanks

Issue #6320, opened 2021-07-15
oss: get endpoint url from the environment if it is possible


isidentical:
https://github.com/iterative/dvc/blob/6275f5d0d738c0bc1611bde0be0afa0e07ba76cd/dvc/fs/oss.py#L34
Would probably require something like `os.getenv("OSS_ENDPOINT")`

Also ossfs should not fail if it can't find the endpoint url, see iterative/ossfs#48

karajan1001:
@isidentical 
> Would probably require something like `os.getenv("OSS_ENDPOINT")`

Good suggestion.

> Also ossfs should not fail if it can't find the endpoint url, see iterative/ossfs#48

`oss` is different from `s3`, its buckets are binding with the endpoint. The addresses of `oss` are not in a similar form like `s3` or `azure` but something like `BucketName.Endpoint`. Without an endpoint, you can't get access to the bucket.

isidentical:
> oss is different from s3, its buckets are binding with the endpoint. The addresses of oss are not in a similar form like s3 or azure but something like BucketName.Endpoint. Without an endpoint, you can't get access to the bucket. The only special endpoint is the accelerate endpoint, it can be used for all buckets, but it requires you to open the accelerate service in advance.

Right. That is really weird, but apparently there seems to be a solution. [OSS domain names](https://partners-intl.aliyun.com/help/doc-detail/31834.htm?spm=a2c63.p38356.b99.36.270824fcs2Pdte) states:

> All requests except the GetService request that are sent to OSS include third-level domains. The third-level domains contain the bucket information.

And a call to [`GetService`](https://www.alibabacloud.com/help/doc-detail/31957.htm) can yield the endpoint url, `<ExtranetEndpoint>oss-cn-shanghai.aliyuncs.com</ExtranetEndpoint>`. What we could do is just giving user a warning about they did not set their bucket's endpoint, and if they intend to do use this functionality later they can set it but if it is a simple `import-url` for example, then they can just fallback to the default solution which we resolve the bucket's location at runtime. Though after seeing this, i think it is not a priority any more. One thing that is important is just adding a check for `endpoint_url` being None, and raising the proper error then.

PR #6524, opened 2021-09-02
setup: bump fsspec/s3fs/gcsfs/adlfs/sshfs to 08.01

isidentical:


setup.py
```diff
@@ -90,19 +90,19 @@ def run(self):
...
 # Extra dependencies for remote integrations
 
-gs = ["gcsfs==2021.7.0"]
+gs = ["gcsfs==2021.8.1"]
 gdrive = ["pydrive2[fsspec]>=1.9.1"]
-s3 = ["s3fs==2021.8.0", "aiobotocore[boto3]>1.0.1"]
-azure = ["adlfs==2021.7.1", "azure-identity>=1.4.0", "knack"]
+s3 = ["s3fs==2021.8.1", "aiobotocore[boto3]>1.0.1"]
+azure = ["adlfs==2021.8.2", "azure-identity>=1.4.0", "knack"]
 oss = ["ossfs==2021.7.5"]
```
efiop, line 104: I guess we need dependabots for our fsspec packages and for dvc, to test with the latest released fsspec version and remind us about it.
```suggestion
oss = ["ossfs==2021.8.0"]
```

karajan1001:
For the OSSFS, 2021.8.0. It now gets endpoint from env variable, related to #6320

isidentical:
Seems like we are getting timeouts due to pip's new dependency resolution system.

isidentical:
depends on: https://github.com/iterative/PyDrive2/pull/133

isidentical:
Seems like some stuff changed regarding how the links are handled in the fsspec, which makes our tests fail: reported at https://github.com/intake/filesystem_spec/issues/742

isidentical:
Let's temporarily switch to `os.path.lexists()` instead of the `self.fs.exists()`, since otherwise it would block progress in other features as well (e.g http)

skshetry (approved):

PR #6525, opened 2021-09-02
http: migrate to fsspec & aiohttp

### LINKED ISSUE #6413: http: migrate to fsspec
`http` is an important remote, and we should migrate it to fsspec. There is already a built-in http filesystem inside `fsspec`, but it seem to lack a couple of features we depend. So there should be some benchmarkings to determine whether we should create a sync, `requests` backed `httpfs` or just use the `HTTPFileSystem` from fsspec.

isidentical:
Resolves #6413

Differences compared to the old HTTPFileSystem (requests based):
- We no longer have `digest` authentication since it is not supported by the aiohttp, at least not yet. (there is an old PR, ~3 years old that is still open). Instead of a deprecation period, now the config option `auth` with the value `digest` will raise an error saying that the option is not available. In the future, if this feature gets accepted to the aiohttp we could bring back the support, but it doesn't seem to be going to happen anytime soon.

### DIFF
--- a/dvc/fs/http.py
+++ b/dvc/fs/http.py
@@ -1,20 +1,12 @@
-import logging
-import os.path
 import threading
-from typing import Optional
-from urllib.parse import urlparse
 
-from funcy import cached_property, memoize, wrap_prop, wrap_with
+from funcy import cached_property, memoize, wrap_with
 
 from dvc import prompt
-from dvc.exceptions import DvcException, HTTPError
 from dvc.path_info import HTTPURLInfo
-from dvc.progress import Tqdm
 from dvc.scheme import Schemes
 
-from .base import BaseFileSystem
-
-logger = logging.getLogger(__name__)
+from .fsspec_wrapper import CallbackMixin, FSSpecWrapper, NoDirectoriesMixin
 
 
 @wrap_with(threading.Lock())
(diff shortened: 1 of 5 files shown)


efiop:
@isidentical Btw, since we are depending on fsspec's core more and more (e.g. that lexist issue you've mentioned today), let's setup testing for upstream fsspec? E.g. as a daily build.

isidentical:
Makes sense!

efiop:
@isidentical Also, could you please elaborate in the PR description on the things we are missing here and other differences introduced by this? E.g. digest auth.

skshetry:
+1 for [httpx](https://www.python-httpx.org).

efiop:
@skshetry could you elaborate, please?

dvc/fs/http.py
```diff
@@ -26,179 +18,102 @@ def ask_password(host, user):
...
-            self.upload_fobj(
-                fobj,
-                to_info,
-                size=None if no_progress_bar else os.path.getsize(from_file),
-                no_progress_bar=no_progress_bar,
-                desc=name or to_info.url,
+            from aiohttp_retry import ExponentialRetry
+            from aiohttp_retry import RetryClient as ClientSession
+        except ImportError:
+            from aiohttp import ClientSession
```
pared, line 99: Is this situation even possible? It seems that both `aiohttp` and `aiohttp-retry` are installed in the same setup option (`http`).

pared (approved):
LGTM, one question.

setup.py
```diff
@@ -103,6 +103,7 @@ def run(self):
 azure = ["adlfs==2021.8.2", "azure-identity>=1.4.0", "knack"]
 oss = ["ossfs==2021.8.0"]
 ssh = ["sshfs>=2021.8.1"]
+http = ["fsspec[http]>=2021.8.1", "aiohttp-retry==2.4.5"]
```
skshetry, line 106: DVC provides http support by default, so this should be part of `install_requires`.

pared (approved):

dvc/fs/http.py
isidentical: That is right (forget about this part), thanks!

efiop:
@isidentical Please clarify the path we will take about digest auth. I know that we've talked about it privately, but we need it to be visible for everyone :)

isidentical:
@efiop updated the PR description.

efiop:
For the record: digest auth from our perception seems to be used rare enough that it makes sense for us to take the risk and to drop it for now. If there will be demand to bring it back, we will consider investing time in it. This is bad to drop a feature like that, but in this case it is worth it to move our filesystems/objects forward.

jorgeorpinel:
Hi. Do we know in what exact release (DVC version) this will be applied? For https://github.com/iterative/dvc.org/pull/2802#issuecomment-916507027

efiop:
@jorgeorpinel 2.7.1

PR #6528, opened 2021-09-02
machine: move to `tpi[ssh]`

casperdcl:
- depends on https://github.com/iterative/tpi
  + :sunglasses: removes need to [manually download terraform](https://www.terraform.io/downloads.html)
- fixes https://github.com/iterative/terraform-provider-iterative/issues/164
- related #6526
- related #6502
- related #6462

The failing test/lint predates this PR.


dvc/machine/__init__.py
```diff
@@ -59,9 +64,18 @@ def __init__(
...
-
-        self.initialized: Dict[str, BaseMachineBackend] = {}
+        self.backends: Dict[str, "BackendCls"] = {}
+        for key in selected:
+            cls = self.DEFAULT.get(key)
+            if cls is None:
+                raise DvcException(
+                    f"'dvc machine' backend '{key}' is missing required "
+                    "dependencies. Install them with:\n"
+                    f"\tpip install dvc[{key}]"
```
pmrowla, line 74: This will be raised when `feature.machine` is enabled and the `tpi` package is missing. When `feature.machine` is disabled (by default) this code block won't be hit (since backends aren't initialized).

pmrowla (approved):

PR #6537, opened 2021-09-06
machine: support instance size & startup script config

pmrowla:

* `instance_size` should be any string `instance_size` accepted by terraform-provider-iterative
* `startup_script` should be (config-relative) path to a shell script file (rather than the raw string used in terraform-provider-iterative)

### DIFF
--- a/dvc/machine/__init__.py
+++ b/dvc/machine/__init__.py
@@ -30,10 +30,11 @@ sudo apt-get update
 sudo apt-get install --yes python3 python3-pip
 python3 -m pip install --upgrade pip
 
-cd /etc/apt/sources.list.d
+pushd /etc/apt/sources.list.d
 sudo wget https://dvc.org/deb/dvc.list
 sudo apt-get update
 sudo apt-get install --yes dvc
+popd
 """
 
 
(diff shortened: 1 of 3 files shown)


dvc/machine/__init__.py
```diff
@@ -25,6 +25,17 @@
...
 
+DEFAULT_STARTUP_SCRIPT = """#!/bin/bash
+sudo apt-get update
+sudo apt-get install --yes python3 python3-pip
+python3 -m pip install --upgrade pip
+
+cd /etc/apt/sources.list.d
+sudo wget https://dvc.org/deb/dvc.list
+sudo apt-get update
+sudo apt-get install --yes dvc
```
isidentical, line 36: Perhaps use `pushd`/`popd` to ensure the cwd is not changed?

isidentical (approved):
LGTM (beside the question)

dvc/machine/__init__.py
pmrowla: Terraform runs the startup script in its own shell instance, so I don't think changing the cwd will actually affect anything else. But it's probably still good to make this change anyways in case we add more actions to the install script later

PR #6539, opened 2021-09-06
extract transfer related logic from odb to fs.utils.transfer

skshetry:

### DIFF
--- a/dvc/fs/utils.py
+++ b/dvc/fs/utils.py
@@ -1,0 +1,57 @@
+import logging
+import os
+from typing import TYPE_CHECKING
+
+from .local import LocalFileSystem
+
+if TYPE_CHECKING:
+    from dvc.types import AnyPath, DvcPath
+
+    from .base import BaseFileSystem
+
+logger = logging.getLogger(__name__)
+
+
+def transfer(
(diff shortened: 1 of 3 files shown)


dvc/fs/utils.py
```diff
@@ -0,0 +1,57 @@
...
+
+if TYPE_CHECKING:
+    from dvc.types import AnyPath, DvcPath
+
+    from .base import BaseFileSystem
+
+logger = logging.getLogger(__name__)
+
+
+def transfer(
```
skshetry, line 15: I am planning to add support for the callback in the `transfer` here, so extracting to `fs.utils`. `move` will definitely cause an issue, not sure how to handle that without an extra `size` call.

skshetry: Also, `upload`/`download_file`/`upload_fobj` requires support for callbacks. 
I am thinking of keeping the progress bar support in the filesystem that we have for now, if `callback` argument is not passed. Then, replace it with appropriate callbacks as needed and then get rid of the Tqdm progress bars inside fs.

skshetry:
@Mergifyio rebase

efiop (approved):

PR #6545, opened 2021-09-07
Set timeout for jobs in Github Actions

skshetry:
The jobs in CI are getting frozen sometimes, either of an issue with `pytest-xdist` or pip resolver due to some weird constraints set by us or one of our dependencies.

The default timeout is 360 minutes (6hrs), but our test completes on ~18 minutes at maximum, so I think 30mins should be enough to account for occasional CI slowdown.

Also keeping the total test job timeout at 45 mins, we usually require just 4 mins of extra time to do installation and setup but giving it 11 more minutes just in case.

Similarly, `lint` completes in <5 mins usually, but giving it a 10 mins timeout.

This should give us early feedback on the issue.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -17,7 +17,7 @@ known_first_party=dvc,tests
 line_length=79
 
 [tool:pytest]
-timeout = 600
+timeout = 100
 timeout_method = thread
 log_level = debug
 addopts = -ra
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -17,7 +17,7 @@ known_first_party=dvc,tests
 line_length=79
 
 [tool:pytest]
-timeout = 600
+timeout = 100
```
skshetry, line 20:
```suggestion
timeout = 200
```

pmrowla (approved):

isidentical (approved):

PR #6546, opened 2021-09-07
fs.upload: add support for optional callback

skshetry:
Also removes _upload methods in other fses, and replaces them
with fsspec-compatible put_file which supports fsspec callback
even if they are not fsspec based filesystem.

### DIFF
--- a/dvc/fs/fsspec_wrapper.py
+++ b/dvc/fs/fsspec_wrapper.py
@@ -3,8 +3,9 @@ import shutil
 from functools import lru_cache
 
 from funcy import cached_property
+from tqdm.utils import CallbackIOWrapper
 
-from dvc.progress import Tqdm
+from dvc.progress import DEFAULT_CALLBACK, Tqdm
 
 from .base import BaseFileSystem
 from .local import LocalFileSystem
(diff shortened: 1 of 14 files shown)


dvc/fs/s3.py
```diff
@@ -204,24 +204,18 @@ def _get_obj(self, path_info):
...
-            **pbar_args,
-        ) as pbar:
-            obj = self._get_obj(to_info)
-            obj.upload_file(
-                from_file,
-                Callback=pbar.update,
-                ExtraArgs=self.fs_args.get("s3_additional_kwargs"),
-                Config=self._transfer_config,
-            )
+        callback.set_size(os.path.getsize(from_file))
```
skshetry, line 210: I am not sure if we need `self.makedirs` here in the `put_file` these days, cc @isidentical.

tests/unit/remote/test_remote_tree.py
```diff
@@ -91,8 +91,8 @@ def test_walk_files(remote):
 def test_copy_preserve_etag_across_buckets(cloud, dvc):
     cloud.gen(FILE_WITH_CONTENTS)
     rem = _get_odb(dvc, cloud.config)
-    s3 = rem.fs
-    s3.fs.mkdir("another/")
+    s3 = rem.fs.s3
+    s3.create_bucket(Bucket="another")
```
skshetry, line 95: The `s3.fs.mkdir` was failing locally for me, though I am not sure why. I replaced this with the one that shares the intention in a better way.

dvc/fs/fsspec_wrapper.py
```diff
@@ -263,22 +255,18 @@ class CallbackMixin:
     """Use the native ``get_file()``/``put_file()`` APIs
     if the target filesystem supports callbacks."""
 
-    def _upload(
-        self, from_file, to_info, name=None, no_progress_bar=False, **pbar_args
+    @staticmethod
+    def put_file_compat(
```
skshetry, line 259: Temporary, I'm planning to get rid of `_upload`/`_download` from `CallbackMixin` and then this `CallbackMixin` will just be a way to provide compat support for callback. Naming is bad, but will just be replace by `put_file` after I get rid of `_download`.

dvc/progress.py
```diff
@@ -182,9 +181,29 @@ def set_size(self, size):
...
     def relative_update(self, inc=1):
         self.progress_bar.update(inc)
+        super().relative_update(inc)
 
     def absolute_update(self, value):
         self.progress_bar.update_to(value)
+        super().absolute_update(value)
+
+
+def tdqm_or_callback_wrapped(
```
skshetry, line 195: This codebase is a bit weird, as we need to support both tqdm and `FsspecCallback` for now.

dvc/progress.py
```diff
@@ -182,9 +181,29 @@ def set_size(self, size):
             size = self.fs.info(self.path_info)["size"]
         self.progress_bar.total = size
         self.progress_bar.refresh()
+        super().set_size(size)
```
skshetry, line 184: These will be useful in tests, as we can introspect size and value easily.

PR #6551, opened 2021-09-07
build: copy fs.http dependencies 

skshetry:
See https://github.com/iterative/dvc/runs/3535516030.


scripts/pyinstaller/hooks/hook-dvc.py
```diff
@@ -14,6 +14,8 @@
 datas += copy_metadata("ossfs")
 datas += copy_metadata("sshfs")
 datas += copy_metadata("webdav4")
+datas += copy_metadata("aiohttp")
+datas += copy_metadata("aiohttp_retry")
```
skshetry, line 18: There's no way to validate it in CI, we need to merge this and see if this worked or not.

skshetry:
Merging as-is, will check CI if it works or not.

PR #6552, opened 2021-09-07
add contrib avatars

casperdcl:
Replaces the old https://github.com/sourcerer-io/sourcerer-app/issues/632

preview: https://github.com/iterative/dvc/pull/6552/files?short_path=7b3ed02#diff-7b3ed02bc73dc06b7db906cf97aa91dec2b2eb21f2d92bc5caa761df5bbc168f or https://github.com/casperdcl/dvc/tree/contrib-avatars#contributing

/CC @jendefig

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -222,7 +222,9 @@ Contributing
 |Maintainability| |Donate|
 
 Contributions are welcome! Please see our `Contributing Guide <https://dvc.org/doc/user-guide/contributing/core>`_ for more
-details.
+details. Thanks to all our contributors!
+
+|Contribs|
 
 Mailing List
 ============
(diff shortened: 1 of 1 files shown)


shcheklein:
( can we update the dvc.org as well? I think its README also has sourcerer still)

shcheklein (approved):

skshetry:
Do we need this these days? since there is a Contributor section in the right hand side in Github now.

casperdcl:
> Do we need this these days?

as @jendefig noted, would make a difference at https://pypi.org/project/dvc etc.

jendefig:
> Do we need this these days? since there is a Contributors section in the right hand side in Github now.

When someone goes to PyPi.org it's good for people to see how many contributors. Builds community, encourages contribution, shows how many people are behind the project which provide a trust factor.

jendefig:
Thanks @casperdcl!!!!! 🙏

skshetry (approved):

isidentical (approved):

PR #6558, opened 2021-09-08
migrate use of upload_fobj to use transfer

skshetry:
Migrate away from `upload_fobj` to `fs.utils.transfer`
Also extend `fs.utils.transfer` to support file object upload and bytes upload.

Pre-requisite for merging `fs.upload` and `fs.upload_fobj` together. `fs.upload_fobj` won't be able to support `callback.get_size` as we don't have that information in the `fs.upload` as there's no `src_fs` and `src_info` to query to, so we are achieving that consistency in terms of callback support with the use of `fs.utils.transfer`.

Also pre-requisite for https://github.com/iterative/dvc/pull/6546.

### DIFF
--- a/dvc/fs/utils.py
+++ b/dvc/fs/utils.py
@@ -1,6 +1,7 @@
 import logging
 import os
-from typing import TYPE_CHECKING
+from io import BytesIO
+from typing import TYPE_CHECKING, BinaryIO, Union
 
 from .local import LocalFileSystem
 
@@ -18,9 +19,25 @@ def transfer(
     to_fs: "BaseFileSystem",
     to_info: "DvcPath",
     move: bool = False,
+    content: Union[bytes, BinaryIO] = None,
 ) -> None:
     use_move = isinstance(from_fs, type(to_fs)) and move
     try:
+        if content:
+            if isinstance(content, bytes):
+                fobj: BinaryIO = BytesIO(content)
+                size = len(content)
+            else:
+                fobj = content
+                size = from_fs.getsize(from_info)
+
+            desc = (
+                from_info.name
+                if isinstance(from_info, from_fs.PATH_CLS)
+                else from_info
+            )
+            return to_fs.upload_fobj(fobj, to_info, size=size, desc=desc)
+
         if use_move:
             return to_fs.move(from_info, to_info)
 
@@ -33,7 +50,8 @@ def transfer(
             return from_fs.download_file(from_info, to_info)
 
         with from_fs.open(from_info, mode="rb") as fobj:
-            return to_fs.upload_fobj(fobj, to_info)
+            size = from_fs.getsize(from_info)
+            return to_fs.upload_fobj(fobj, to_info, size=size)
     except OSError as exc:
         # If the target file already exists, we are going to simply
         # ignore the exception (#4992).
--- a/dvc/objects/db/reference.py
+++ b/dvc/objects/db/reference.py
@@ -1,6 +1,4 @@
-import io
 import logging
-import os
 from typing import TYPE_CHECKING, Dict
 
 from dvc.scheme import Schemes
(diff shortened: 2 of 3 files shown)


dvc/fs/utils.py
```diff
@@ -18,9 +19,19 @@ def transfer(
     to_fs: "BaseFileSystem",
     to_info: "DvcPath",
     move: bool = False,
+    content: Union[bytes, object] = None,
 ) -> None:
     use_move = isinstance(from_fs, type(to_fs)) and move
     try:
+        if content:
```
skshetry, line 26: This is a bit overloaded though, but moving it with for now.

dvc/fs/utils.py
```diff
@@ -18,9 +19,19 @@ def transfer(
...
+            if isinstance(content, bytes):
+                fobj: "BinaryIO" = BytesIO(content)
+                size = len(content)
+            else:
+                fobj = content
+                size = from_fs.getsize(from_info)
+            return to_fs.upload_fobj(fobj, to_info, size=size)
```
skshetry, line 33: `fs.utils.transfer` is the only place where we use `upload_fobj`.

dvc/fs/utils.py
```diff
@@ -18,9 +19,19 @@ def transfer(
     to_fs: "BaseFileSystem",
     to_info: "DvcPath",
     move: bool = False,
+    content: Union[bytes, "BinaryIO"] = None,
```
skshetry, line 22:
```suggestion
    content: Union[bytes, BinaryIO] = None,
```

dvc/fs/utils.py
```diff
@@ -18,9 +19,19 @@ def transfer(
...
 ) -> None:
     use_move = isinstance(from_fs, type(to_fs)) and move
     try:
+        if content:
+            if isinstance(content, bytes):
+                fobj: "BinaryIO" = BytesIO(content)
```
skshetry, line 28:
```suggestion
                fobj: BinaryIO = BytesIO(content)
```

isidentical (approved):

dvc/fs/utils.py
```diff
@@ -18,9 +19,19 @@ def transfer(
...
     use_move = isinstance(from_fs, type(to_fs)) and move
     try:
+        if content:
+            if isinstance(content, bytes):
+                fobj: BinaryIO = BytesIO(content)
+                size = len(content)
+            else:
+                fobj = content
+                size = from_fs.getsize(from_info)
+            return to_fs.upload_fobj(fobj, to_info, size=size)
```
skshetry, line 33:
```suggestion
            return to_fs.upload_fobj(fobj, to_info, size=size, desc=from_info.name)
```

efiop (approved):

dvc/objects/db/reference.py
```diff
@@ -60,25 +58,19 @@ def _add_file(
...
-                and exc.__context__
-                and isinstance(exc.__context__, FileExistsError)
-            ):
-                logger.debug("'%s' file already exists, skipping", to_info)
-            else:
-                raise
+        content = ref_file.to_bytes()
+        fs.utils.transfer(
+            from_fs, from_info, self.fs, to_info, move=move, content=content
+        )
```
skshetry, line 74: @pmrowla, what do you think of this? Is `transfer` needed here, or should it use more primitive functions like `pipe_file`?

pmrowla: The `upload_fobj` for reference objects isn't really a file transfer, it's just a `write()` of the serialized reference data from memory.

It seems to me like we should still have a distinction between a file transfer and a direct binary data write/upload, since in this case there really isn't a `from_fs` and `from_info` that we are transferring.

One alternative would maybe be adding something like `ReferenceHashFile.to_memfile()` where it returns a `memfs://some_tmp_pathname` path to a `MemoryFileSystem` entry (containing the serialized data). And then we could do an actual "file transfer" using the memfs and memfs_path_info without needing the `content` field?

skshetry: yeah, initially I started with two methods, but at the end `fs.utils.write` is going to be equivalent to `upload_fobj`, so I don't see much difference there. With `transfer`, I tried to push the problem of getting `size` for the callback support inside the `transfer` itself, rather than putting that burden on the caller.

We do the following when transferring files straight to the remote, which is just a file transfer at the end:
```python
 with fs.open(path_info, mode="rb", chunk_size=fs.CHUNK_SIZE) as stream:
         stream = HashedStreamReader(stream)
         upload_odb.fs.upload_fobj(
             stream, tmp_info, desc=path_info.name, size=fs.getsize(path_info)
         )
```

But maybe the best thing to do here is put that burden on the caller, rather than trying to generalize it in `transfer`.

skshetry:
Closing this PR as we'll likely put the burden of doing `callback.set_size` on the caller itself.

PR #6560, opened 2021-09-08
project: formalize extra dependencies

### LINKED ISSUE #6559: project: formalize extra dependency types
While working on #6540, I noticed that it would be a bit simpler if we had dependencies organized as sole files instead of the current way (so i can simply strip the version specifiers out on a certain directory to test with the newer versions). A simple way of doing this is just creating a `requirements/` directory and adding entries for each remote we have as a new file. This will also allow dependabot to upgrade our remote dependencies automatically (just like it is doing for the tests).

isidentical:
Resolves #6559


requirements/azure.txt
```diff
@@ -0,0 +1,3 @@
+adlfs==2021.8.2
```
skshetry, line 1: We also need to include these files in the source distributions. Please add the following contents in the [`./MANIFEST.in`](https://github.com/iterative/dvc/blob/378486dbf53271ec894b4380f60ec02dc3351516/MANIFEST.in):

```
recursive-include requirements *.txt
```

skshetry (approved):

setup.py
```diff
@@ -97,29 +97,22 @@ def run(self):
 
 
 # Extra dependencies for remote integrations
+extra_requirements = {
+    path.stem: path.read_text().strip().splitlines()
+    for path in Path("requirements").glob("*.txt")
+}
```
skshetry, line 103: I am having a second thought here if we should move the requirements to `setup.cfg` and use pyup.io. 

That's what I use in webdav4: https://github.com/skshetry/webdav4/blob/main/setup.cfg.

isidentical: I also agree that is a better option, especially if we are migrating the regular metadata as well.

PR #6570, opened 2021-09-09
fs: merge upload_fobj and upload

skshetry:

### DIFF
--- a/dvc/fs/base.py
+++ b/dvc/fs/base.py
@@ -247,6 +247,7 @@ class BaseFileSystem:
             ) as wrapped:
                 # `size` is used to provide hints to the WebdavFileSystem
                 # for legacy servers.
+                # pylint: disable=no-member
                 return self.upload_fobj(wrapped, to_info, size=total)
 
         if from_info.scheme != "local":
--- a/dvc/fs/base.py
+++ b/dvc/fs/base.py
@@ -217,44 +217,49 @@ class BaseFileSystem:
             return False
         return hash_.endswith(cls.CHECKSUM_DIR_SUFFIX)
 
-    def upload(self, from_info, to_info, name=None, no_progress_bar=False):
-        if not hasattr(self, "_upload"):
-            raise RemoteActionNotImplemented("upload", self.scheme)
+    def upload(
+        self,
+        from_info,
+        to_info,
+        total=None,
+        desc=None,
+        no_progress_bar=False,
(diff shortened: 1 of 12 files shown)


dvc/fs/base.py
```diff
@@ -217,44 +217,49 @@ def is_dir_hash(cls, hash_):
...
+                "read",
+                disable=no_progress_bar,
+                bytes=True,
+                total=total,
+                desc=desc,
+                **pbar_args,
+            ) as wrapped:
+                # `size` is used to provide hints to the WebdavFileSystem
+                # for legacy servers.
+                return self.upload_fobj(wrapped, to_info, size=total)
```
skshetry, line 250:
```suggestion
                # pylint: disable=no-member
                return self.upload_fobj(wrapped, to_info, size=total)
```

pmrowla (approved):

efiop:
@skshetry Btw, could you clarify how you see it in relation to fsspec?

dvc/fs/base.py
```diff
@@ -217,44 +217,50 @@ def is_dir_hash(cls, hash_):
...
 
         if to_info.scheme != self.scheme:
             raise NotImplementedError
 
-        if from_info.scheme != "local":
-            raise NotImplementedError
-
         logger.debug("Uploading '%s' to '%s'", from_info, to_info)
+        if is_file_obj:
+            with Tqdm.wrapattr(
```
karajan1001, line 239: This TQDM wrapper in different levels for `upload_fobj` and `_upload`. Hard to maintain them.

skshetry: This is temporary.

skshetry:
I don't think I have any motivations to unify them from fsspec perspective. My first preference there would be to have an explicit API like `upload_fileobj`/`upload_fobj` and my second choice would be to extend `pipe_file` to support file objects. 
`upload` and `upload_fobj` are quite different things in the filesystem layer.

For us, what we need is a `fs.utils.write` like method (discussed in https://github.com/iterative/dvc/pull/6558#discussion_r704927903) that can also provide callback support, which is more of a service layer for fs.

I am assuming `upload_fobj` as a fsspec's API at least for now, so I have tried to not add callback support there, as the caller can do so themselves quite easily. 

I'll consider `upload` API to be that `fs.utils.write`, that extends fsspec API for us, rather than being fsspec-related.

Issue #6265, opened 2021-07-01
executors: share repo with executor


dberenbaum:
Meta-issue. Please link related tickets back to this one.

### User story

I want to share my code with my executor.

### Requirements

* Local repo code is accessible on the executor and reflects the local workspace at the time of running `dvc exp run --executor`.

dberenbaum:
Is this best shared directly over ssh or should it be transferred via github or other git remotes if possible?

dberenbaum:
Need to rescope

PR #6589, opened 2021-09-10
[WIP] exp: add initial SSH executor

pmrowla:

Related to https://github.com/iterative/dvc/issues/6265


pmrowla:
related: https://github.com/dulwich/dulwich/pull/902

dvc/repo/experiments/executor/ssh.py
```diff
@@ -0,0 +1,142 @@
...
+    def _git_client_args(fs):
+        kwargs = {
+            "password": fs.fs_args.get("password"),
+            "key_filename": first(fs.fs_args.get("client_keys", [])),
+        }
+        if os.name != "win32":
+            kwargs["ssh_command"] = (
+                "ssh -o UserKnownHostsFile=/dev/null "
+                "-o StrictHostKeyChecking=no"
+            )
```
pmrowla, line 101: TODO: this needs to be moved into dvc-machine specific ssh args before this PR is mergeable (we don't want to disable host checking for non-terraform generated hosts)

pmrowla: Upon further review, this probably does need to be handled either here in the ssh executor or in the dulwich git backend. Dulwich chooses an ssh vendor type (openssh client, putty, etc) depending on local platform, so these args are really per git/executor and not per dvc-machine

PR #6591, opened 2021-09-10
objects.transfer: minor refactoring, move lazier taskset inside custom executor

skshetry:


dvc/utils/threadpool.py
```diff
@@ -0,0 +1,32 @@
...
+
+        it = zip(*iterables)
+        tasks = create_taskset(self.max_workers * 5)
+        while tasks:
+            done, tasks = futures.wait(
+                tasks, return_when=futures.FIRST_COMPLETED
+            )
+            for fut in done:
+                yield fut.result()
+            tasks.update(create_taskset(len(done)))
```
efiop, line 32: Btw, there was a plan to try using async here e.g. with fsspec's put/get methods. Maybe it would be reasonable to do that instead now?

skshetry: Could you please elaborate? I am not sure I understand.

efiop: So we use thread pool executor right now to parallelize uploading/downloading of objects, but fsspec provides `put/get` methods that accept batches and  async filesystems do that with async, effectively managing the coroutines, instead of us manually using threads.

skshetry: So what you are suggesting is to have something like TransferManager that handles multithreaded and async transfer as well right?

dvc/objects/transfer.py
```diff
@@ -38,109 +37,48 @@ def wrapper(path_info, *args, **kwargs):
...
-        bound_file_ids = set()
-        dir_obj: Optional["Tree"] = None
-        for odb in (cache_odb, src):
-            if odb is not None:
-                try:
-                    dir_obj = Tree.load(odb, dir_hash)
-                    break
-                except (FileNotFoundError, ObjectFormatError):
-                    pass
+        dir_obj = find_tree_by_obj_id([cache_odb, src], dir_hash)
```
skshetry, line 74: Why do we call it `dir_obj`, not `tree`?

dvc/utils/threadpool.py
efiop: @skshetry Ah, no, I meant that we possibly won't need transfer manager at all, and could just `fs.put(obj_list)` and let it do its thing. So we would be effectively delegating transfer management to particular filesystem's put/get methods.

efiop: @skshetry Btw, happy to jump on a call to talk about it a bit more.

dvc/objects/transfer.py
efiop: It's a terminology mess (+ .dir objects as we call them, and not .tree). Also driving me insane for 3.0 :slightly_smiling_face:

skshetry: `.dir` is a physical repr thing similar to how it's json and kept in some structure, it's fine till 3.0.  Though I'd say that we `s/dir_obj/tree` in all of these abstractions right away.

dvc/utils/threadpool.py
skshetry: I think we should wait for fsspec changes to be done, this will likely require us to have colored APIs too (`transfer_async`, `odb.add_async` etc.).  And, there's a problem of running fsspec's sync FS in async mode too. 

As this change is very minor restructuring, I'd prefer to merge this as-is.

dvc/objects/transfer.py
pmrowla: It's a lot of leftovers from when remote was migrated to odb, and yeah it would be good to just drop the old terminology entirely whenever possible

Issue #6594, opened 2021-09-10
diff: use object diff


efiop:
Currently `dvc/repo/diff.py` is quite a bit outdated, we didn't touch it since before introducing objects, so it is still based on hash_info, which makes it suboptimal and very confusing. What we should do instead is introduce `dvc/objects/diff.py` and implement generic diffing logic for `Tree` objects, that could then be used in `dvc diff`. Also it is important to remember that even though right now we don't provide one `Tree` for the whole repo (kinda like git commits do), we will do that soon, which will again simplify `dvc diff` logic, that will then use `objects.diff` even more directly.

Btw, we are already doing basic object diff when merging trees in our git merge-driver (see `merge` in `dvc/objects/tree.py`).

efiop:
We also clearly need this for checkout.

dberenbaum:
@efiop Is this still p1?

skshetry:
`diff` uses `index.diff` now. Closing. 

- https://github.com/iterative/dvc/pull/8930

PR #6595, opened 2021-09-10
diff: don't forget to use dvcignore and state

efiop:
Also no longer using `repofs` for staging. Significantly speeds up `dvc diff`:

Before:
```
$ pytest tests/benchmarks/cli/commands/test_diff.py --size large
test_diff-diff-noop     13.3057
test_diff-diff          43.4180
```

After:
```
$ pytest tests/benchmarks/cli/commands/test_diff.py --size large
test_diff-diff-noop      2.6598
test_diff-diff          30.0593
```

Really, this is more of a temporary mitigation before https://github.com/iterative/dvc/issues/6594

### DIFF
--- a/dvc/repo/diff.py
+++ b/dvc/repo/diff.py
@@ -4,7 +4,6 @@ from collections import defaultdict
 from typing import Dict, List
 
 from dvc.exceptions import PathMissingError
-from dvc.objects.stage import get_file_hash
 from dvc.repo import locked
 from dvc.repo.experiments.utils import fix_exp_head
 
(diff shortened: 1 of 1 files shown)


shcheklein:
Is it easy/makes sense to do a benchmark for this? :)

efiop:
@shcheklein We already have one (the results are in the description), will be adding some amendments though.

dvc/repo/diff.py
```diff
@@ -131,45 +130,47 @@ def _to_path(output):
...
 
+            if on_working_fs:
+                _, obj = ostage(
+                    repo.odb.local,
+                    output.path_info,
+                    repo.odb.local.fs,
+                    "md5",
+                    dry_run=True,
+                    dvcignore=output.dvcignore,
+                )
```
efiop, line 147: For the record: We really shouldn't be calling this in such a verbose form, but this and other points are related to https://github.com/iterative/dvc/issues/6594

PR #6600, opened 2021-09-13
output: use remote kwarg instead of odb

### LINKED ISSUE #6596: pull -r remote: Disregards specified http remote
# Bug Report

## Description
When running `dvc pull -r <remote>`, the requests sent in order to retrieve directory data are sent to the default remote instead of the specified one.

### Reproduce
1. `git clone repo` (that has dvc directories)
2. `dvc remote add <remote> <http-remote>`
3. `dvc pull -r <remote>`

### Expected
DVC pull should try to fetch the data from the specified remote

### Environment information
MacOS

**Output of `dvc doctor`:**

```console
DVC version: 2.7.2 (pip)
Platform: Python 3.7.6 on Darwin-20.6.0-x86_64-i386-64bit
Supports:
	http (aiohttp = 3.7.4.post0, aiohttp-retry = 2.4.5),
	https (aiohttp = 3.7.4.post0, aiohttp-retry = 2.4.5)
Cache directory: apfs on /dev/disk1s2s1
Caches: local
Remotes: https, http
Workspace directory: apfs on /dev/disk1s2s1
Repo: dvc, git
```

The bug seems to have been introduced with #6486 
In the following code (dvc/output.py):
```
def get_dir_cache(self, **kwargs):
        if not self.is_dir_checksum:
            raise DvcException("cannot get dir cache for file checksum")

        obj = self.odb.get(self.hash_info)
        try:
            objects.check(self.odb, obj)
        except FileNotFoundError:
            remote = self.repo.cloud.get_remote_odb(self.remote). # <---- here self.remote is None if nothing is specified for the output instead of falling back to kwargs["remote"]
            self.repo.cloud.pull([obj.hash_info], odb=remote, **kwargs)
```

efiop:
Fixes #6596

### DIFF
--- a/dvc/output.py
+++ b/dvc/output.py
@@ -851,7 +851,7 @@ class Output:
         except FileNotFoundError:
             if self.remote:
                 kwargs["remote"] = self.remote
-            self.repo.cloud.pull([obj.hash_info], odb=remote, **kwargs)
+            self.repo.cloud.pull([obj.hash_info], **kwargs)
 
         if self.obj:
             return self.obj
--- a/dvc/output.py
+++ b/dvc/output.py
@@ -849,7 +849,8 @@ class Output:
         try:
             objects.check(self.odb, obj)
         except FileNotFoundError:
-            remote = self.repo.cloud.get_remote_odb(self.remote)
+            if self.remote:
+                kwargs["remote"] = self.remote
             self.repo.cloud.pull([obj.hash_info], odb=remote, **kwargs)
 
         if self.obj:
--- a/tests/func/test_data_cloud.py
+++ b/tests/func/test_data_cloud.py
@@ -654,6 +654,7 @@ def test_output_remote(tmp_dir, dvc, make_remote):
         "b8a9f715dbb64fd5c56e7783c6820a61",
     }
 
+
 def test_target_remote(tmp_dir, dvc, make_remote):
     make_remote("default", default=True)
     make_remote("myremote", default=False)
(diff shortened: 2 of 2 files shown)


dvc/output.py
```diff
@@ -849,7 +849,8 @@ def get_dir_cache(self, **kwargs):
         try:
             objects.check(self.odb, obj)
         except FileNotFoundError:
-            remote = self.repo.cloud.get_remote_odb(self.remote)
+            if self.remote:
```
efiop, line 852: This is getting irritating, will need to look into getting rid of `get_dir_cache` soon. :disappointed:

skshetry: Shouldn't `-r <remote>` be of higher priority?

skshetry: My second priority would be to skip not specified remotes.

skshetry: Though, for `dir` cache, I would not care that much. 🤷🏼

efiop: `remote: ` has higher prioirity than `-r`, as documented in https://github.com/iterative/dvc.org/pull/2761/files . Similar how `dvc import`ed remotes are not overriden by `-r`

dvc/output.py
```diff
@@ -849,7 +849,8 @@ def get_dir_cache(self, **kwargs):
...
+                kwargs["remote"] = self.remote
             self.repo.cloud.pull([obj.hash_info], odb=remote, **kwargs)
```
dberenbaum, line 854:
```suggestion
            self.repo.cloud.pull([obj.hash_info], **kwargs)
```

efiop:
ci failure is unrelated

PR #6609, opened 2021-09-14
Fix ci failures

karajan1001:
related to ci failures, and related to #6468 

The current ci failure was caused by wrong order of the experiments.
Because our test script runs so fast that they might share the same
commit timestamp, that makes them in an uncertain order. And then gives
a wrong branch_base.

### DIFF
--- a/tests/func/experiments/test_show.py
+++ b/tests/func/experiments/test_show.py
@@ -495,6 +495,8 @@ def test_show_with_broken_repo(tmp_dir, scm, dvc, exp_stage, caplog):
 
 
 def test_show_csv(tmp_dir, scm, dvc, exp_stage, capsys):
+    import time
+
     baseline_rev = scm.get_rev()
 
     def _get_rev_isotimestamp(rev):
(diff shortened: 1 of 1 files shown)


karajan1001:
And as @skshetry had suggested we can replace this e2e functional test with unit tests. And in the unit test, the man-made `all_experiements` data will not have this problem.

efiop (approved):
This works for now, but it is a slippery slope 😄 We have a long story of abusing time.sleep in the past.

tests/func/experiments/test_show.py
```diff
@@ -505,6 +507,7 @@ def _get_rev_isotimestamp(rev):
     result1 = dvc.experiments.run(exp_stage.addressing, params=["foo=2"])
     rev1 = first(result1)
     ref_info1 = first(exp_refs_by_rev(scm, rev1))
+    time.sleep(1)
```
shcheklein, line 510: does it makes sense to add a comment here?

karajan1001: Only a temporary solution, I will create a new PR and replace this test with a serial of unit tests with man-made data to avoid the original problem.

PR #6610, opened 2021-09-14
fs.download: add callback support

skshetry:
With this, we won't have any progress bars in our filesystem, except on `base.py` where we have 3 progress bars on `upload`, `download_file`, and `download_dir`. `download_dir` requires nested/branched callback, so it has been tricky to remove those right away. Will be fixed in upcoming PRs.

With this PR, all the filesystems will use `get_file` irrespective of being fsspec-based or not to download files, and has callback support.

### DIFF
--- a/dvc/fs/fsspec_wrapper.py
+++ b/dvc/fs/fsspec_wrapper.py
@@ -5,7 +5,7 @@ from functools import lru_cache
 from funcy import cached_property
 from tqdm.utils import CallbackIOWrapper
 
-from dvc.progress import DEFAULT_CALLBACK, Tqdm
+from dvc.progress import DEFAULT_CALLBACK
 
 from .base import BaseFileSystem
 from .local import LocalFileSystem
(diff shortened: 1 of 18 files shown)


dvc/fs/fsspec_wrapper.py
```diff
@@ -252,35 +242,36 @@ def ls(self, *args, **kwargs):
 
 
 class CallbackMixin:
```
skshetry, line 244: CallbackMixin is now a compat-layer for filesystems that don't support callbacks yet.

tests/func/test_fs.py
```diff
@@ -439,5 +439,41 @@ def test_upload_callback(tmp_dir, dvc, cloud):
...
+    [
+        pytest.lazy_fixture("azure"),
+        pytest.lazy_fixture("gs"),
+        pytest.param(
+            pytest.lazy_fixture("gdrive"),
+            marks=pytest.mark.xfail(
+                reason="https://github.com/iterative/PyDrive2/issues/136"
+            ),
+        ),
+        pytest.lazy_fixture("gdrive"),
```
skshetry, line 459:
```suggestion
```

efiop (approved):

PR #6615, opened 2021-09-14
ssh: ask password before setting it

isidentical:
Resolves [the following error](https://discuss.dvc.org/t/how-do-i-use-dvc-with-ssh-remote/279/21?u=isidentical)


efiop:
@mergifyio rebase

pared (approved):

dberenbaum:
Could this be caught with a test passing expected config fields to `_prepare_credentials()` and checking the expected return?

PR #6618, opened 2021-09-15
gdrive: strip bucket

skshetry:
This was giving me the entries such as `root/root/dir/{foo,bar}` instead of just `root/dir/{foo,bar}`. Though I am not sure if this has any effects on existing users.

### DIFF
--- a/dvc/fs/gdrive.py
+++ b/dvc/fs/gdrive.py
@@ -296,6 +296,15 @@ class GDriveFileSystem(FSSpecWrapper):  # pylint:disable=abstract-method
 
         return super()._with_bucket(path)
 
+    def _strip_bucket(self, entry):
+        try:
+            bucket, path = entry.split("/", 1)
+        except ValueError:
+            # If there is no path attached, only returns
+            # the bucket (top-level).
+            bucket, path = entry, None
+        return path or bucket
+
     def upload_fobj(self, fobj, to_info, **kwargs):
         rpath = self._with_bucket(to_info)
         self.makedirs(os.path.dirname(rpath))
--- a/tests/func/test_fs.py
+++ b/tests/func/test_fs.py
@@ -493,9 +493,7 @@ def test_download_callback(tmp_dir, dvc, cloud):
                 os.name == "nt", reason="unsupported on Windows."
             ),
         ),
-        pytest.param(
-            pytest.lazy_fixture("gdrive"), marks=pytest.mark.xfail(strict=True)
-        ),
+        pytest.lazy_fixture("gdrive"),
     ],
 )
 def test_download_dir_callback(tmp_dir, dvc, cloud):


shcheklein:
@skshetry what are we try to solve here? may be I can give a bit more context on GDrive urls.

skshetry:
@shcheklein, when trying to use `fs.walk_files(GDriveURLInfo("gdrive://root/dir"))`, it returns the following list of values:
```
>>> fs.walk_files(GDriveURLInfo("gdrive://root/5511199/dir"))
[GDrive: 'gdrive://root/root/5511199/dir/bar', GDrive: 'gdrive://root/root/5511199/dir/foo']
# instead of 
# [GDrive: 'gdrive://root/5511199/dir/bar', GDrive: 'gdrive://root/5511199/dir/foo']
```

You can see this here on a new test that I added: https://github.com/iterative/dvc/runs/3598940548?check_suite_focus=true#step:12:5139.

This happens because on `walk_files`, we replace the `path` that we get from the `.find` method:
https://github.com/iterative/dvc/blob/f3010d1d65bf2eb181c63f4bf0c3ca01b837f50b/dvc/fs/fsspec_wrapper.py#L113-L115

And, the `.find` method returns complete path, i.e. `root/5511199/dir/foo` and `root/5511199/dir/bar`. GDriveURLInfo also has additional component: `bucket` which is set as `root`, so this replaces path and keeps the bucket too, making the `root` appear twice.

We can solve this in two ways: either by replacing the bucket and path at the same time or don't consider the bucket at all in our dvc (so, everything will be relative to the bucket/remote). I went with the second option.

I was wondering if this bug affected existing users.

shcheklein:
> I went with the second option, which is what we do in s3/azure, etc.

A few things to consider - `root` is not the only option. 

(to be honest bucket is not the right term at all for the GDrive, to my mind in general URL can be `gdrive:///file-id` - mind the tripple `///`)

A possible valid URL is `gdrive://<drive-id>` or `gdrive://<directory-id>`. Would this change affect such URLs in any way?

shcheklein:
Overall the change makes sense to me. I don't see from the top of my head where it can break anything.

dvc/fs/gdrive.py
```diff
@@ -296,6 +296,15 @@ def _with_bucket(self, path):
 
         return super()._with_bucket(path)
 
+    def _strip_bucket(self, entry):
```
shcheklein, line 299: we have similar implementation in the `ObjectFSWrapper` ? what is the difference here?

skshetry: ObjectFSWrapper is for object-based filesystems, which has some shared methods that we need, especially for directories-like path `/dir/`, etc. It is copied from `ObjectFSWrapper` as you said. This is a temporary stuff that we need, until we get rid of `PathInfo`/`CloudURLInfo`.

skshetry:
> (to be honest bucket is not the right term at all for the GDrive, - mind the tripple `///`)

yeah, I was quoting internal terminology that we use on `GDriveURLInfo`.

> A possible valid URL is `gdrive://<drive-id>` or `gdrive://<directory-id>`. Would this change affect such URLs in any way?

The <directory-id> or `<drive-id>` will be considered as a bucket in that case, so I think it should still work. 

>  to my mind in general URL can be `gdrive:///file-id`

I am not sure about `///file-id`, did we support that before?

shcheklein:
> I am not sure about `///file-id`, did we support that before?

not it's not supported and was never supported, It's me thinking about use cases like `dvc get url gdrive://something`, etc ... in some case it makes sense that `host` part of the URL is empty.

Again, not related to this PR.

isidentical (approved):

PR #6619, opened 2021-09-15
tests: scriptify function into module

skshetry:
so that we can write tests having larger side-effects than
the run_copy that we use as a hack.

Let me know what you think, it is magicky for sure, it'd great if we could run python code inside it. But we can't, and also we usually need better scripts for tests than `run_copy`.


pared (approved):
Great idea!

isidentical (approved):

Issue #6441, opened 2021-08-16
exp init: create command to append a default experiment stage


dberenbaum:
Requirements:

* Create default paths for data, params, metrics, plots, code, and models.
* Create `dvc exp init` command that takes the command as input and appends a stage to dvc.yaml using default paths.
* Make reasonable assumptions about what to cache, what to make a dependency vs. output, etc. (see https://www.notion.so/iterative/Initialize-Experiments-cec7b7c6e3d2451490491d22dfb4e63f).

Issue #6446, opened 2021-08-16
Epic: exp init


dberenbaum:
Parent issue for lightweight command for initializing experiments.

Linked issues:

- [ ] #6441 
- [ ] #6442 
- [ ] #6443 
- [ ] #6444 
- [ ] #6445

karajan1001:
Maybe we can have a label for these summary?

skshetry:
From the @dberenbaum's comment on https://github.com/iterative/dvc/pull/6681#issuecomment-928180859.

>How do you anticipate this (dvclive scenario) working for non-interactive mode?
>
> I'm still wondering whether the template option is needed (solely as a cli option, not as an actual template; it could be renamed) to separate the dvclive scenario and provide flexibility to add other scenarios in the future. What do you think?


One way may be to make `--live` mandatory for dvclive scenario unless `live` is set on the config (though there's a question about what to do if `metrics` and `plots` are also specified in the config).

Another would be to create a `--template` or something similar flag (`--with={live, etc.}`)?

dberenbaum:
> Another would be to create a `--template` or something similar flag (`--with={live, etc.}`)?

Although it adds another option, I like this mainly because it's more explicit. Now that we don't have templates, it's less transparent what the command and various options are doing. Having a separate option for this suggests that it's not just another path to add but actually changes the stage configuration. We can also further explain it in the help text for that option.

On a similar note, at least interactive mode should probably have a confirmation so that users can see what will be added to `dvc.yaml` beforehand. There could be an option to do the same outside of interactive mode (`--dry`?). Not sure if this has been discussed before for `dvc stage add`.

skshetry:
Tasks that remains:
- [ ] Improve UI/CLI, messaging, and prompts
- [ ] Add config support
- [ ] Add tests/QA

daavoo:
Might be worth revisiting https://github.com/iterative/dvcyaml-schema/issues/7 to better support the workflow: 

`dvc exp init` for simple cases / basic bootstrapping -> manually edit `dvc.yaml` for customization?

skshetry:
With dvcyaml-schema, my goal was to have same schema for JSONSchema and validation. We have been using pydantic for JSONSchema generation and voluptuous for validation.

There are few issues with using pydantic:
[...]

PR #6621, opened 2021-09-15
exp init: add simple dvc exp init command

skshetry:
Simpler, initial implementation for `exp init`.

Closes https://github.com/iterative/dvc/issues/6441.

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -1360,7 +1360,7 @@ def add_parser(subparsers, parent_parser):
     )
     experiments_remove_parser.set_defaults(func=CmdExperimentsRemove)
 
-    EXPERIMENTS_INIT_HELP = "Create experiments."
+    EXPERIMENTS_INIT_HELP = "Initialize experiments."
     experiments_init_parser = experiments_subparsers.add_parser(
         "init",
         parents=[parent_parser],
@@ -1396,12 +1396,12 @@ def add_parser(subparsers, parent_parser):
     experiments_init_parser.add_argument(
         "--code",
         help="Path to the source file or directory "
-        "which your experiment depends",
+        "which your experiments depend",
     )
     experiments_init_parser.add_argument(
         "--data",
         help="Path to the data file or directory "
-        "which your experiment depends",
+        "which your experiments depend",
     )
     experiments_init_parser.add_argument(
         "--models",
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -1,5 +1,6 @@
 import argparse
 import logging
+import os
 from collections import Counter, OrderedDict, defaultdict
 from datetime import date, datetime
 from fnmatch import fnmatch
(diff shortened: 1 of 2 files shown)


dvc/command/experiments.py
```diff
@@ -789,6 +790,62 @@ def run(self):
...
+        name = self.args.name or self.DEFAULT_NAME
+        stage = self.repo.stage.add(
+            name=name,
+            cmd=cmd,
+            deps=[code, data],
+            outs=[models],
+            params=[{params_path: params}],
+            metrics_no_cache=[metrics],
+            plots_no_cache=[plots],
+            live=dvclive,
```
dberenbaum, line 840: Default stage can ignore `--dvclive` arg. We probably need conditional arg parsing based on the template, throwing an error if a provided argument is irrelevant for that template.

dvc/command/experiments.py
```diff
@@ -1303,6 +1360,68 @@ def add_parser(subparsers, parent_parser):
     )
     experiments_remove_parser.set_defaults(func=CmdExperimentsRemove)
 
+    EXPERIMENTS_INIT_HELP = "Create experiments."
```
skshetry, line 1363:
```suggestion
    EXPERIMENTS_INIT_HELP = "Initialize experiments."
```

dvc/command/experiments.py
skshetry: We don't have a default stage/template yet, so I left it with the default. I'll change this when we have templates.

dvc/command/experiments.py
```diff
@@ -789,6 +790,62 @@ def run(self):
...
+        stage = self.repo.stage.add(
+            name=name,
+            cmd=cmd,
+            deps=[code, data],
+            outs=[models],
+            params=[{params_path: params}],
+            metrics_no_cache=[metrics],
+            plots_no_cache=[plots],
+            live=dvclive,
+            force=True,
```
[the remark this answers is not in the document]
skshetry: I'll add the `force` option internally on the next PR. :)

dvc/command/experiments.py
```diff
@@ -1303,6 +1360,68 @@ def add_parser(subparsers, parent_parser):
...
+        help="Run the experiment after initializing it",
+    )
+    experiments_init_parser.add_argument(
+        "--interactive",
+        "-i",
+        action="store_true",
+        help="Prompt for values that are not provided",
+    )
+    experiments_init_parser.add_argument(
+        "--template", help="Stage template to use to fill with provided values"
```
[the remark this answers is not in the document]
skshetry: Will add defaults in the upcoming PRs, as the place where I'll keep those defaults may constantly change. :)

skshetry:
@dberenbaum, I am merging this and will follow up on your comments in the future PRs. The command is hidden meanwhile.

dvc/command/experiments.py
```diff
@@ -1303,6 +1360,68 @@ def add_parser(subparsers, parent_parser):
...
+    )
+    experiments_init_parser.add_argument(
+        "--explicit", help="Only use the path values explicitly provided"
```
dberenbaum, line 1391: By the way, feel free to suggest a better name for this option. Maybe `--ignore-missing`?

PR #6624, opened 2021-09-15
get rid of pytest-timeout

skshetry:
Trying to see if getting rid of this dependency, our issue with the timeout
gets solved or not.

We use timeout-minutes anyway on GHA, so we probably don't need it.
Locally, users can monitor this themselves.


skshetry:
Am I being lucky here, or did this really fixed the issue? 😂

skshetry:
I was just being lucky. 😂

PR #6628, opened 2021-09-16
Dvclive output subdirs

daavoo:

None of the changes are strictly required by DVCLive but I feel like the existing output structure could be revisited after the addition of `image` support and the changes to `html` generation (which is now a folder).

This P.R. makes `live` outputs to be ordered in subdirs divided by plot type. Currently `linear` and `images`.
Summary is now generated under `{live_path}/summary.json` instead of `{live_path}.json`.
HTML is now generated under `{live_path}/html` instead of `{live_path}_dvc_plots`.


pared:
We still need to make compatible changes on `dvclive` side, right?

daavoo:
> We still need to make compatible changes on `dvclive` side, right?

Yep. Just wanted to check how this would look on `DVC` side and what are your thoughs

pared:
I believe this is a great idea.  Generating summary and HTML is along the `live` was awkward. Inside, I think its great. One thing to consider is `.dvcignore` -ing HTML if it's produced so that we don't store it.

dberenbaum:
Related: https://github.com/iterative/dvclive/issues/148 and https://github.com/iterative/dvc/pull/6431#discussion_r698317884.

Great idea. I have two questions:

1. Is the images path needed now?
2. Is `linear` the best name for the `.tsv` directory? You might be thinking ahead to https://github.com/iterative/dvclive/issues/82, but I think `linear` is not that intuitive in dvclive for users, especially since it contains the `.tsv` and not the plot. Other ideas:
  a. `metrics`: the `.tsv` files are step-based logs of what would generally be considered metrics in dvc and similar tools.
  b. `scalars`: describes the data type being logged (this is also what's used in tensorboard for their linear plots).

dberenbaum:
Related complaint: https://discord.com/channels/485586884165107732/485596304961962003/915439735256801352. Maybe we should consider reopening this?

dberenbaum:
> There are a few different things that don't seem strictly related to the P.R:

💯 It's not strictly related. Let me open a separate issue to discuss.

Issue #4410, opened 2020-08-18
Make Output files Optional (dvc run)


kailingding:
Hi, I've been using DVC for a while now. Loved it so far.

Recently, I'm working on a pipeline that doesn't always output files that are defined in `dvc run -o` command. In other words, if there is a DVC feature that can cache an output file if existed and ignore it if such a file does not exist, that would be great! Because when running experiments, output files defined with `-o` would not always be outputed.

Here's an example of what I exactly mean:

score.py
```python
import json

if __name__ == "__main__":
    score = {"rmse": 0.98}

    saved = True
    if saved:
        with open("./score.json", 'w') as fp:
            json.dump(score, fp)
```

Terminal
```
>>> git init
>>> dvc init
>>> dvc run -n score  -o ./score.json  python ./score.py
```

After running commands above, a `score.json` would be cached by DVC. 

But when I set
```python
saved = False
```
And run the following command
```
dvc repro ./dvc.yaml
```

The following error occurs
```
ERROR: failed to reproduce 'dvc.yaml': output 'score.json' does not exist
```

If there is a feature that can solve this problem, that would be awesome!

shcheklein:
Discord context: https://discordapp.com/channels/485586884165107732/485596304961962003/745002571353030717

efiop:
For the record: clearly we can't make these optional by-default, as that will be too error-prone in most of the cases. We could consider introducing a special flag for such options, but that would only make sense if we would also make a similar optional dependency flag for stages that might depend on this file, otherwise you might as well just not specify the output at all.

I see that you've already discussed using empty files as placeholders and I would also suggest doing that instead of a special optional dep/out flag. Seems like that workaround is currently working for you just fine, so I would stick to it, since it is a lot less hacky than optional deps/outs that we would have to introduce otherwise. That said, I can see that being a potential advanced feature in the future.

shcheklein:
@kailingding btw, what type of outputs do you need this feature for first? It looks like your case is about logging (optionally) some ML metrics, right? is it a regular DVC metrics files, or some plots (`dvc plots`)?

I'm asking primarily for two reasons - if it's a metrics file, then due to #https://github.com/iterative/dvc/issues/4068#issuecomment-676540417 empty file might not an option? Also, if we this feature is needed mostly for end results (loggers) we might solve it faster by allowing optional leaf outputs I guess.

kailingding:
[...]

PR #6630, opened 2021-09-16
exp init: add basic template support

skshetry:
This adds basic support for templating on `dvc exp init`.
`interactive` and `explicit` are not supported yet. The custom templates are supported, 
but right now, as we don't initialize it in `dvc init`, it has to be added manually in `.dvc/stages` by the
user.  I don't want to add that support yet before having this stable, as we may have some backward incompatible changes, so for testing, you'd have to do `mkdir .dvc/stages` yourself for a while.

The default templates are not usable yet for binary packages users.
Note that the stage is appended after the template has been parsed, validated and then processed by stage when dumping, so it may not reflect the template very well, there may be formatting differences.

I tried using direct updates first, but doing that, it leaves formatting issues generated by templates (like whitespaces, empty newlines, etc.) and comment gets left-shifted. So, I went with the `stage.dump()` option. We could roll back to it if we want to in the future.

In terms of code, all of the stuff related to initialization has been moved to `experiments.init` module, but I am still hesitant to add it to `repo.experiments` as a new API.

Also, it adds a dependency on `jinja2`. Regarding tests, I have only added simple ones and I don't plan to add until this feature starts taking shape.

### DIFF
--- a/dvc/repo/experiments/init.py
+++ b/dvc/repo/experiments/init.py
@@ -59,7 +59,7 @@ def init(
 ):
     from jinja2 import Environment
 
-    from dvc.dvcfile import make_dvcfile
+    from dvc.utils import relpath
     from dvc.stage import check_circular_dependency, check_duplicated_arguments
     from dvc.stage.loader import StageLoader
     from dvc.utils.serialize import LOADERS, parse_yaml_for_update
@@ -94,7 +94,7 @@ def init(
 
     # render, parse yaml and then validate schema
     rendered = template.render(**context, param_names=param_names)
-    template_path = os.path.relpath(template.filename)
+    template_path = relpath(template.filename)
     data = parse_yaml_for_update(rendered, template_path)
     try:
         validated = STAGE_SCHEMA(data)
--- a/dvc/repo/experiments/init.py
+++ b/dvc/repo/experiments/init.py
@@ -59,9 +59,9 @@ def init(
 ):
     from jinja2 import Environment
 
-    from dvc.utils import relpath
     from dvc.stage import check_circular_dependency, check_duplicated_arguments
     from dvc.stage.loader import StageLoader
+    from dvc.utils import relpath
     from dvc.utils.serialize import LOADERS, parse_yaml_for_update
 
     data = compact(data)  # remove None values
--- a/dvc/repo/experiments/init.py
+++ b/dvc/repo/experiments/init.py
@@ -100,7 +100,7 @@ def init(
         validated = STAGE_SCHEMA(data)
     except MultipleInvalid as exc:
         raise DvcException(
-            f"template '{template_path}'"
+            f"template '{template_path}' "
             "failed schema validation while rendering"
         ) from exc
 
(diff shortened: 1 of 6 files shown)


dvc/repo/experiments/init.py
```diff
@@ -0,0 +1,121 @@
...
+
+    # render, parse yaml and then validate schema
+    rendered = template.render(**context, param_names=param_names)
+    template_path = os.path.relpath(template.filename)
+    data = parse_yaml_for_update(rendered, template_path)
+    try:
+        validated = STAGE_SCHEMA(data)
+    except MultipleInvalid as exc:
+        raise DvcException(
+            f"template '{template_path}'"
```
skshetry, line 103:
```suggestion
            f"template '{template_path}' "
```

dvc/repo/experiments/init.py
```diff
@@ -0,0 +1,121 @@
...
+    assert "params" in context
+    # See https://github.com/iterative/dvc/issues/6605 for the support
+    # for depending on all params of a file.
+    param_path = str(context["params"])
+    _, ext = os.path.splitext(param_path)
+    param_names = list(LOADERS[ext](param_path))
+
+    # render, parse yaml and then validate schema
+    rendered = template.render(**context, param_names=param_names)
+    template_path = os.path.relpath(template.filename)
```
skshetry, line 97:
```suggestion
    template_path = relpath(template.filename)
```

dvc/repo/experiments/init.py
```diff
@@ -0,0 +1,121 @@
...
+    data: Dict[str, Optional[object]],
+    template_name: str = None,
+    interactive: bool = False,
+    explicit: bool = False,
+    template_loader: Callable[["Repo"], "BaseLoader"] = get_loader,
+    force: bool = False,
+):
+    from jinja2 import Environment
+
+    from dvc.dvcfile import make_dvcfile
```
skshetry, line 62:
```suggestion
    from dvc.utils import relpath
```

dvc/repo/experiments/init.py
```diff
@@ -0,0 +1,121 @@
...
+    data: Dict[str, Optional[object]],
+    template_name: str = None,
+    interactive: bool = False,
+    explicit: bool = False,
+    template_loader: Callable[["Repo"], "BaseLoader"] = get_loader,
+    force: bool = False,
+):
+    from jinja2 import Environment
+
+    from dvc.stage import check_circular_dependency, check_duplicated_arguments
```
skshetry, line 62:
```suggestion
    from dvc.utils import relpath
    from dvc.stage import check_circular_dependency, check_duplicated_arguments
```

pared:
LGTM so far, 
I have a question regarding:

> The default templates are not usable yet for binary packages users.

Did you already think about how to include that support? I have a similar problem in #6550 and my research led me to `pkg_resources`. I guess it would be good to agree on some convention.

dberenbaum:
> Did you already think about how to include that support? I have a similar problem in #6550 and my research led me to `pkg_resources`. I guess it would be good to agree on some convention.

Yes, it would be great to maintain consistency with #6550 and #4410.

skshetry:
This is stale, superseded by #6681.

skshetry:
> Did you already think about how to include that support? 

I have not research about this yet, as I don't want to add the support till we are done with all the changes with `exp init`. :)

PR #6631, opened 2021-09-16
[WIP] diff: don't use path_info

efiop:
On `get-started-experiments` this shaves off around ~3.5 seconds from 7 seconds total in `dvc diff`.


efiop:
Ok, closing in favor of https://github.com/iterative/dvc/issues/6594 , since have to mess with path_info anyway.

Issue #6482, opened 2021-08-24
machine: list status of created instances


pmrowla:
* [ ] `dvc machine status` should list status of running (`dvc machine create`'d ) instances

Output should be similar to what is displayed in `terraform show` (cloud, IP, instance sizes, etc)

Related to #6263

PR #6633, opened 2021-09-16
`dvc machine`: add a new command `rename` to it

### LINKED ISSUE #6481: machine: support renaming machines
* [ ] `dvc machine rename` should rename `.dvc/config` machine entries
* [ ] `dvc machine rename` should also `terraform state mv` existing instances

Related to #6263

karajan1001:
fix #6481
1. add new command `dvc machine rename`
2. add a unit test for the call.
3. add several functional tests for it.t push myself fix6481


dvc/command/machine.py
```diff
@@ -104,6 +104,53 @@ def run(self):
         return 0
 
 
+class CmdMachineRename(CmdBase):
+    def _check_exists(self, conf):
```
karajan1001, line 108: Didn't use multi inheritance here, because it is hard to maintain and we only need this `_check_exists` from `CmdMachineConfig` here.

dvc/machine/__init__.py
```diff
@@ -200,3 +203,17 @@ def get_sshfs(self, name: Optional[str]):
...
+
+        cloud_type = self._get_config(name=old_name)["cloud"]
+        backend = self.CLOUD_BACKENDS[cloud_type]
+        old_path = os.path.join(self.machine_space, backend, old_name)
+        new_path = os.path.join(self.machine_space, backend, new_name)
+        if os.path.exists(new_path):
+            raise MachineExistError(new_name, new_path)
+        if not os.path.exists(old_path):
+            return
+        move(old_path, new_path)
```
pmrowla, line 219: I think this should be handled in the `MachineBackend` classes instead of at the manager level. Moving the directory may not necessarily be the intended operation here, and the specifics of the `machine/<backend>` directory structure is intended to only have meaning for a given backend. 

Also, the backend will likely have to do some additional operations beyond just moving/renaming the directory. (Such as accounting for already-running machine instances that use the old name)

pmrowla:
@karajan1001 I think it may be beneficial to look into doing `machine status` (https://github.com/iterative/dvc/issues/6482) before `machine rename`.

Properly supporting rename will require checking for running instances and using `terraform state mv` (https://www.terraform.io/docs/cli/commands/state/mv.html) to rename existing resources, it's a bit more involved than only needing to rename the config section and moving the `tmp/machine/...` dirs around.

Doing the `status` work first will help give you a better idea of how the TPI wrapper works, and how terraform state works as well, that should make it clearer how `rename` will need to be addressed.

pmrowla:
Basically, with the directory move only implementation from this PR, if you do

```
$ dvc machine add foo ...
$ dvc machine create foo
$ dvc machine rename foo bar
```

The repo will now be in a state in which the user cannot destroy the created/running `foo` machine instance.

`dvc machine destroy bar` will fail because there will be not `tfstate` resource named `bar` (since the `foo` instance was not renamed through terraform). (And `dvc machine destroy foo` will fail because the DVC config entry for `foo` no longer exists).

In order to destroy the existing resource the user would have to manually do
```
$ cd .dvc/tmp/machine/terraform/bar
$ terraform destroy
```

And if the user happened to run `dvc machine create bar` before realizing the old `foo` resource was still running, the original terraform config for `foo` would be completely overwritten by `machine create bar`, and the user would not be able to destroy `foo` at all (unless they log into the AWS/azure web console and manually destroy it through the web UI)

karajan1001:
related to https://github.com/iterative/tpi/issues/12

pmrowla (commented):
This LGTM for the most part, but I think we need to consider it blocked by https://github.com/iterative/tpi/issues/15 for the time being.

As discussed with @karajan1001, we need to move the per-machine workspace directory handling out of TPI and back into dvc-machine. (At which point the `shutil.mv` behavior on `dvc machine rename` will become less confusing)

dberenbaum:
Sorry, I'm way too late in commenting here, but I'm not sure this is worth the effort it's taking. How important is it really to have the ability to rename a machine? In particular, is it really needed to rename a machine with running instances? I think throwing an error asking users to shut down their instances before renaming is sufficient.

pmrowla:
related: https://github.com/iterative/dvc/pull/6748 (moves the per-name directory handling back into DVC and out of TPI, so the rename operation should make more sense on the DVC side now)

karajan1001:
```
Or maybe rename specifically should only support the repo and local levels?
```

Similiar problems in `remote config`. Maybe it is better to make some of the configurations never global?

karajan1001:
> ```
> Or maybe rename specifically should only support the repo and local levels?
> ```
> 
> Similiar problems in `remote config`. Maybe it is better to make some of the configurations only above `repo` level?

PR #6641, opened 2021-09-17
tests: additional callbacks test for remaining filesystems

skshetry:
Gdrive will fail unless the `pydrive2` is updated.

Added a few more remotes for tests of the callbacks as well.

### DIFF
--- a/tests/func/test_fs.py
+++ b/tests/func/test_fs.py
@@ -429,6 +429,8 @@ def test_fs_makedirs_on_upload_and_copy(dvc, cloud):
         pytest.lazy_fixture("s3"),
         pytest.lazy_fixture("ssh"),
         pytest.lazy_fixture("webhdfs"),
+        pytest.lazy_fixture("webdav"),
+        pytest.lazy_fixture("http"),
     ],
 )
 def test_upload_callback(tmp_dir, dvc, cloud):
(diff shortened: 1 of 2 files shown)


tests/func/test_fs.py
```diff
@@ -450,24 +452,23 @@ def test_upload_callback(tmp_dir, dvc, cloud):
     [
         pytest.lazy_fixture("azure"),
         pytest.lazy_fixture("gs"),
-        pytest.param(
-            pytest.lazy_fixture("gdrive"),
-            marks=pytest.mark.xfail(
-                reason="https://github.com/iterative/PyDrive2/issues/136"
-            ),
-        ),
```
skshetry, line 458: `gdrive` will fail until the dependency is updated.

tests/func/test_fs.py
```diff
@@ -507,3 +509,59 @@ def test_download_dir_callback(tmp_dir, dvc, cloud):
...
+    callback = fsspec.Callback()
+    fs.download(tmp_dir / "file", tmp_dir / "file2", callback=callback)
+
+    size = os.path.getsize(tmp_dir / "file")
+    assert (tmp_dir / "file2").read_text() == "file"
+    assert callback.size == size
+    assert callback.value == size
+
+    if fs_type == "git":
+        pytest.skip("gitfs does not support download_dir")
```
skshetry, line 530: `download_dir` is not working, not sure if we use `fs.download_dir` for git filesystems.

efiop (approved):

efiop:
For the record: ci failures are unrelated.

PR #6645, opened 2021-09-17
gha: increase timeouts

efiop:
Looks like we are running into timeouts from time to time, need to investigate what's going on.


karajan1001 (approved):
Some of my PRs also failed after 30minute at 99%.

efiop:
Ok, let's just increse them for now, will need to get back with a more in-depth investigation later. Looks like scm tests take awhile, +1 reason to extract it into a separate project.

PR #6647, opened 2021-09-19
command: Show dvc list output in pretty JSON format

eggqq007:
1. update ls __init__.py to handle --show-pretty-json
2. introduce pygments into requirements/default.txt

Closes https://github.com/iterative/dvc.org/issues/2838

### DIFF
--- a/dvc/command/ls/__init__.py
+++ b/dvc/command/ls/__init__.py
@@ -35,9 +35,7 @@ class CmdList(CmdBaseNoRepo):
                 dvc_only=self.args.dvc_only,
             )
             if self.args.show_json:
-                import json
-
-                ui.write(json.dumps(entries))
+                ui.write_json(entries)
             elif entries:
                 entries = _prettify(entries, with_color=True)
                 ui.write("\n".join(entries))
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -36,7 +36,7 @@ tabulate>=0.8.7
 pygtrie>=2.3.2
 dpath>=2.0.2,<3
 shtab>=1.3.4,<2
-rich>=10.0.0
+rich>=10.9.0
 dictdiffer>=0.8.1
 python-benedict>=0.21.1
 pyparsing==2.4.7
--- a/dvc/ui/__init__.py
+++ b/dvc/ui/__init__.py
@@ -74,6 +74,20 @@ class Console:
             styled=styled,
         )
 
+    def write_json(
+        self,
+        data: Any,
+        indent: int = 2,
+    ) -> None:
+        if sys.stdout.isatty():
+            from rich.json import JSON
+
+            j = JSON.from_data(data, indent=indent)
+            return self.write(j, styled=True)
+        import json
+
+        return self.write(json.dumps(data))
+
     def write(
         self,
         *objects: Any,


dvc/command/ls/__init__.py
```diff
@@ -72,6 +84,11 @@ def add_parser(subparsers, parent_parser):
     list_parser.add_argument(
         "--show-json", action="store_true", help="Show output in JSON format."
     )
+    list_parser.add_argument(
+        "--show-pretty-json",
```
skshetry, line 88: We could just prettify the JSON by default by checking if we are in `isatty` mode and then prettify that output.

skshetry:
Btw, why not just do `dvc list --show-json . | python -m rich.json -`?

requirements/default.txt
```diff
@@ -44,3 +44,4 @@ typing_extensions>=3.10.0.2
 fsspec[http]>=2021.8.1
 aiohttp-retry==2.4.5
 diskcache>=5.2.1
+pygments>=2.10.0
```
skshetry, line 47: We depend on `rich`, so you could just do something like following:
```python
from rich.json import JSON

obj = JSON.from_data(data)
ui.write(obj, styled=True)
```

dvc/command/ls/__init__.py
eggqq007: Thanks for the suggestion! The patch was updated

dvc/command/ls/__init__.py
```diff
@@ -37,7 +38,12 @@ def run(self):
             if self.args.show_json:
                 import json
 
-                ui.write(json.dumps(entries))
+                if sys.stdout.isatty():
+                    from rich.json import JSON
+
+                    ui.write(JSON.from_data(entries), styled=True)
+                else:
+                    ui.write(json.dumps(entries))
```
skshetry, line 46: Let's move this to [`dvc/ui/__init__.py`](https://github.com/iterative/dvc/blob/e81c88e501b79d2e1873b109f56332a4ad3903b7/dvc/ui/__init__.py) as `ui.write_json`, so that we can reuse it in other parts of dvc.

Also, support for `indent=` and `prettify=True|False` that keeps the old behaviour would be nice to have. 🙂

skshetry: Also, we may need to bump `rich` requirement, as `from_data` is only available since 10.9.0.

eggqq007: Sure, that makes sense.  patch updated.

skshetry (approved):

skshetry:
The PR looks good. @iterative/dvc, if you have any opinions if dvc should format json. Otherwise, we'll be formatting in other places as well.

dvc/command/ls/__init__.py
```diff
@@ -35,9 +36,10 @@ def run(self):
                 dvc_only=self.args.dvc_only,
             )
             if self.args.show_json:
-                import json
-
-                ui.write(json.dumps(entries))
+                if sys.stdout.isatty():
```
efiop, line 39: We could hide this check under ui.write_json (if rich doesn't do that already for us) CC @skshetry

efiop: Though that would get a little confusing with `prettify`, as it is not clear if it should enforce it or be more like "prettify, but only if it is a tty". Still, it feels like `tty` belongs to ui.

efiop: Btw, is there any significant performace hit for prettifying? If not, maybe we should indeed consider just showing pretty json by default?

eggqq007: Thanks for the comments!. It is hard to tell the performance gap between json.dumps and prettifying. I can do some evaluation on it later. Besides that, I tend to believe the prettifying is not the bottleneck since dvc list should fetch repo info from remote first which may spend more time.

eggqq007: I did two simple tests on it. 
1. A python array has 100000 element, each element is a dict which contains 3 key-value.  json.dump takes about 0.05s, rich.json.from_date takes about 2.6s

2. dvc list takes about 1.8s to fetch https://github.com/iterative/example-get-started while prettifying  takes 0.03s

skshetry: I don't think performance should be a concern. We could choose a certain threshold like 1000 or so entries when we fallback to `json.dumps(indent=4)` instead of prettifying it. No one should expect us to prettify the 100000 many elements of entries.

efiop: regretfully `dvc list`(and others) are not yet fully optimized themselves and having 100K entries in the resulting json is not unusual for our use case, so looks like prettifying is kinda expensive to do always. Plus we have other projects depending on this json output, so that migth be slowing them down too. But `isatty` check kinda solves that, because if someone will be using the json output - they will redirect it somewhere anyway, so we should be safe and efficient, while being pretty. That should make the threshold approach unnecessary.

eggqq007: Thanks for the comments! I move the isatty into ui __init__.py and the patch updated. :)

skshetry (approved):
We discussed this today in the team, and everyone is okay with this getting merged. For consistency, we'd like this to be used in all of the other commands that has `--json` flags (though in a separate PR).

isidentical (approved):

skshetry:
@eggqq007, would you like to create a PR (or, issue) for https://github.com/iterative/dvc/pull/6647#pullrequestreview-765482070? Thanks.

eggqq007:
> @eggqq007, would you like to create a PR (or, issue) for [#6647 (review)](https://github.com/iterative/dvc/pull/6647#pullrequestreview-765482070)? Thanks for this PR.

Sure, new issue in https://github.com/iterative/dvc/issues/6712

PR #6648, opened 2021-09-19
checkout: start switching to obj-based workflow 

efiop:
First steps in migrating checkout to object-based workflow. Introduces `dvc.objects.diff` that can be used in other applications (e.g. in `dvc diff`) later. Allows us to stage existing target first(multithreaded, benefiting from all current and future improvements in `dvc.objects.stage`), then compare the resulting object with the end result that we are trying to achieve and use that diff to only apply actual changes.

`diff()` has some pre-requisites for supporting granular `Tree`s in the future, but for the most part this is trying to carry over legacy things, that will be revisited in the followups.

Pre-requisite for #6594

Todo:
- [x] investigate checkpoints test failing
- [x] check benchmarks

### DIFF
--- a/dvc/checkout.py
+++ b/dvc/checkout.py
@@ -1,4 +1,5 @@
 import logging
+from itertools import chain
 
 from shortuuid import uuid
 
@@ -10,52 +11,19 @@ from dvc.exceptions import (
     DvcException,
 )
 from dvc.ignore import DvcIgnoreFilter
-from dvc.objects import check, load
 from dvc.objects.db.slow_link_detection import (  # type: ignore[attr-defined]
     slow_link_guard,
 )
-from dvc.objects.errors import ObjectFormatError
+from dvc.objects.diff import ROOT
+from dvc.objects.diff import diff as odiff
 from dvc.objects.stage import stage
+from dvc.objects.tree import Tree
 from dvc.types import Optional
 
 logger = logging.getLogger(__name__)
 
 
-def _changed(path_info, fs, obj, cache, state=None):
-    logger.trace("checking if '%s'('%s') has changed.", path_info, obj)
-
-    try:
-        check(cache, obj)
-    except (FileNotFoundError, ObjectFormatError):
-        logger.debug(
-            "cache for '%s'('%s') has changed.", path_info, obj.hash_info
-        )
-        return True
-
-    actual = state.get(path_info, fs) if state else None
-    if actual is None:
-        try:
-            actual = stage(
-                cache, path_info, fs, obj.hash_info.name, dry_run=True
-            )[1].hash_info
-        except FileNotFoundError:
-            logger.debug("'%s' doesn't exist.", path_info)
-            return True
-
-    if obj.hash_info != actual:
-        logger.debug(
-            "hash value '%s' for '%s' has changed (actual '%s').",
-            obj.hash_info,
-            actual,
-            path_info,
-        )
-        return True
-
-    logger.trace("'%s' hasn't changed.", path_info)
-    return False
-
-
-def _remove(path_info, fs, cache, force=False):
+def _remove(path_info, fs, in_cache, force=False):
     if not fs.exists(path_info):
         return
 
@@ -63,13 +31,7 @@ def _remove(path_info, fs, cache, force=False):
         fs.remove(path_info)
         return
 
-    current = stage(cache, path_info, fs, fs.PARAM_CHECKSUM, dry_run=True)[
-        1
-    ].hash_info
-    try:
-        obj = load(cache, current)
-        check(cache, obj)
-    except (FileNotFoundError, ObjectFormatError):
+    if not in_cache:
         msg = (
             f"file/directory '{path_info}' is going to be removed. "
             "Are you sure you want to proceed?"
@@ -98,7 +60,7 @@ def _verify_link(cache, path_info, link_type):
 
 def _do_link(cache, from_info, to_info, link_method):
     if cache.fs.exists(to_info):
-        raise DvcException(f"Link '{to_info}' already exists!")
+        cache.fs.remove(to_info)  # broken symlink
 
     link_method(from_info, to_info)
 
@@ -130,7 +92,7 @@ def _link(cache, from_info, to_info):
     try:
         _try_links(cache, from_info, to_info, cache.cache_types)
     except FileNotFoundError as exc:
-        raise CheckoutError([to_info]) from exc
+        raise CheckoutError([str(to_info)]) from exc
 
 
 def _cache_is_copy(cache, path_info):
@@ -160,7 +122,7 @@ def _cache_is_copy(cache, path_info):
 def _checkout_file(
     path_info,
     fs,
-    obj,
+    change,
     cache,
     force,
     progress_callback=None,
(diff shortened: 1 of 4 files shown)


dvc/objects/diff.py
```diff
@@ -0,0 +1,135 @@
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, List, Optional, Tuple
+
+if TYPE_CHECKING:
+    from .file import HashFile
+
+ADD = "add"
+MODIFY = "modify"
+DELETE = "delete"
+UNCHANGED = "unchanged"
```
efiop, line 10: We'll add rename detection too in the followup, when working on `dvc diff`.

dvc/objects/diff.py
```diff
@@ -0,0 +1,135 @@
...
+        if not isinstance(other, KeyObjPair):
+            return False
+
+        if self.key != other.key or bool(self.obj) != bool(other.obj):
+            return False
+
+        if not self.obj:
+            return False
+
+        return self.obj.hash_info == other.obj.hash_info
```
efiop, line 32: Need to reconsider existing `obj.__eq__` to be effectively `self.hash_info == other.hash_info`, as the current one (includes fs and path_info comparison too, which doesn't work with staged objects 🙁 ) makes it awkward to compare in applications like this.

dvc/objects/diff.py
```diff
@@ -0,0 +1,135 @@
...
+    def __bool__(self):
+        return bool(self.added or self.modified or self.deleted)
+
+
+ROOT = ("",)
+
+
+def diff(
+    old: Optional["HashFile"], new: Optional["HashFile"], cache
+) -> DiffResult:
```
efiop, line 76: In git, you are only dealing with trees, but in our case we don't yet have a repo `Tree` that we could use (plus a few more issues), so we try to be a bit more general and support `file`&`tree` diff too, which results in some seemingly odd hacks like `ROOT`. Should be able to get rid of those once we introduce `repo.tree` and `view`s.

dvc/objects/diff.py
```diff
@@ -0,0 +1,135 @@
...
+        elif change.typ == MODIFY:
+            ret.modified.append(change)
+        elif change.typ == DELETE:
+            ret.deleted.append(change)
+        else:
+            assert change.typ == UNCHANGED
+            if not change.new.in_cache and not isinstance(
+                change.new.obj, Tree
+            ):
+                ret.modified.append(change)
```
efiop, line 132: yet another thing that differs us from git, is that our objects could be missing from our object store (aka odb, aka cache), so we need this as a part of diff. (e.g. we already do that in `dvc diff`)

tests/func/test_add.py
```diff
@@ -496,27 +496,27 @@ def test_should_update_state_entry_for_directory_after_add(
 
     ret = main(["add", "data"])
     assert ret == 0
-    assert file_md5_counter.mock.call_count == 3
+    assert file_md5_counter.mock.call_count == 4
```
efiop, line 499: This is because we previously called `stage` for every file in a directory when running `checkout()`, and now we just `stage` the whole directory in bulk, so that's +1 `file_md5` for `tree.digest()`.

Side note: we should revisit these kinds of tests and probably just have some unit tests instead.

tests/func/experiments/test_checkpoints.py
```diff
@@ -143,7 +143,7 @@ def test_resume_branch(tmp_dir, scm, dvc, checkpoint_stage, workspace):
         checkpoint_stage.addressing,
         checkpoint_resume=branch_rev,
         params=["foo=100"],
-        tmp_dir=not workspace,
+        tmp_dir=True,  # not workspace,
```
efiop, line 146: Looks like preceeding `apply` doesn't perform `dvc checkout` correctly, because checkout things that existing `foo` is already `2` instead of `4` because state db tells it so. Investigating if this is my patch or a problem with state 😱

tests/func/test_add.py
skshetry: yeah, With `objects`, `dvc add` should not care about file_md5 at all.

dvc/objects/diff.py
```diff
@@ -0,0 +1,135 @@
...
+        if not self.obj:
+            return False
+
+        return self.obj.hash_info == other.obj.hash_info
+
+
+@dataclass
+class Change:
+    old: KeyObjPair
+    new: KeyObjPair
```
skshetry, line 38: What's `old` and what's `new` in the objects?

dvc/objects/diff.py
```diff
@@ -0,0 +1,135 @@
...
+    from .file import HashFile
+
+ADD = "add"
+MODIFY = "modify"
+DELETE = "delete"
+UNCHANGED = "unchanged"
+
+
+@dataclass
+class KeyObjPair:
```
skshetry, line 14: `KeyObjPair` looks a bit weird, maybe `FileChange`/`DiffFile`, etc. as this relates to a single file.

dvc/objects/diff.py
```diff
@@ -0,0 +1,135 @@
...
+    in_cache: bool
+    key: Tuple[str]
+    obj: Optional["HashFile"] = field(default=None)
+
+    def __bool__(self):
+        return bool(self.obj)
```
skshetry, line 20: Of all the protocols, I like `__bool__` the least, but probably it makes sense here.

dvc/checkout.py
```diff
@@ -169,135 +130,109 @@ def _checkout_file(
...
-    else:
-        existing_files = set(fs.walk_files(path_info))
+    old = None
+    try:
+        _, old = stage(
+            cache,
+            path_info,
+            fs,
+            obj.hash_info.name if obj else cache.fs.PARAM_CHECKSUM,
+            dry_run=True,
```
skshetry, line 173: Not related to this PR, `stage` but not really looks weird. `dry_run=True` should probably be just `hash()/`hash_file()`.

dvc/checkout.py
```diff
@@ -169,135 +130,109 @@ def _checkout_file(
...
+        entry_path = (
+            path_info.joinpath(*change.old.key)
+            if change.old.key != ROOT
+            else path_info
+        )
+        _remove(entry_path, fs, change.old.in_cache, force=force)
 
     failed = []
-    for entry_key, entry_obj in obj:
+    for change in diff.added + diff.modified:
```
skshetry, line 209: concatenation might be expensive.
```suggestion
    for change in chain(diff.added, diff.modified):
```

dvc/objects/diff.py
efiop: Not sure I understand the question. Could you elaborate, please? `old` and `new` are `diff` terms, rather than `object` terms. E.g. dulwich uses `tree1` and `tree2` https://github.com/dulwich/dulwich/blob/81bd07a84254036af7e95268921d653450aa4b3e/dulwich/diff_tree.py#L166 , but I find that more confusing than old and new, as with the latter you at least instinctively know the direction of the difference.

dvc/objects/diff.py
efiop: Yeah, didn't want to use `file` there, as it could be used for trees too in the near future (and actually now too when ROOT used to be a file but now a tree), but that's still better than `pair` and we can adjust it later anyway. Will adjust.

dvc/checkout.py
efiop: I also don't like `dry_run`, could probably consider removing it and defaulting to `True`. IIRC the only difference is that we'll use a staging reference odb, which should be fine for internals. I don't think we should use `hash/hash_file`, because those are misleading, we should not be operating on hashes, but on objects here.

dvc/checkout.py
efiop: Nice catch!

dvc/checkout.py
skshetry: I was talking about something like [`git_odb_hashfile`](https://libgit2.org/libgit2/#HEAD/group/odb/git_odb_hashfile) or [`pygit2.hash_file`](https://www.pygit2.org/objects.html#pygit2.hashfile). Even though it has `hash` in the name, it's similar to what `stage(dry_run=True)` is doing.

efiop:
benchmarks so far:

master:
```
---------------------------------------------
test_checkout-checkout                    7.8 sec
test_checkout-checkout-noop          0.95 sec
test_checkout-checkout-update       25 sec
---------------------------------------------
```

this PR:
```
------------------------------------------
test_checkout-checkout                  12.2 sec
test_checkout-checkout-noop        2 sec
test_checkout-checkout-update     5.8 sec
------------------------------------------
```

so we've gained better performance in complex cases, but lost performance in simple cases because we've also adopted all performance problems of `stage()` that need to be solved anyway (but they are common for multiple commands, so one optimization round there will improve many), plus diffing big trees is kinda slow in python when using a naive approach (need to polish it out and/or add a simple c extension, e.g. like dulwich). Still, this is moving us in the right direction and allows us to move along with `dvc diff` and `dvc status` and put all problems into one batch (e.g. `stage()` + obj `diff()`), so merging as is (no new release for now though).

PR #6649, opened 2021-09-20
machine: Add a new command `dvc machine status`

karajan1001:
fix#6482
1. add a new command `dvc machien status`
2. add a unit test for the cli call.
3. add a unit test for the shown result.

### DIFF
--- a/dvc/command/machine.py
+++ b/dvc/command/machine.py
@@ -143,6 +143,21 @@ class CmdMachineCreate(CmdBase):
         return 0
 
 
+class CmdMachineStatus(CmdBase):
+    def run(self):
+        if self.repo.machine is None:
+            raise MachineDisabledError
+
+        all_status = self.repo.machine.status(self.args.name)
+        for i, status in enumerate(all_status):
+            ui.write(f"instance_num_{i+1}:")
+            keys = list(status.keys())
+            keys.sort()
+            for key in keys:
+                ui.write(f"\t{key:20}: {status[key]}")
+        return 0
+
+
 class CmdMachineDestroy(CmdBase):
     def run(self):
         if self.repo.machine is None:
@@ -305,6 +320,19 @@ def add_parser(subparsers, parent_parser):
         "name", help="Name of the machine to create."
     )
     machine_create_parser.set_defaults(func=CmdMachineCreate)
+
+    machine_STATUS_HELP = "List the status of a running machine."
+    machine_status_parser = machine_subparsers.add_parser(
+        "status",
+        parents=[parent_config_parser, parent_parser],
+        description=append_doc_link(machine_STATUS_HELP, "machine/status"),
+        help=machine_STATUS_HELP,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+    machine_status_parser.add_argument(
+        "name", help="Name of the running machine."
+    )
+    machine_status_parser.set_defaults(func=CmdMachineStatus)
 
     machine_DESTROY_HELP = "Destroy an machine instance."
     machine_destroy_parser = machine_subparsers.add_parser(
--- a/dvc/machine/__init__.py
+++ b/dvc/machine/__init__.py
@@ -2,8 +2,10 @@ import logging
 import os
 from typing import (
     TYPE_CHECKING,
+    Any,
     Dict,
     Iterable,
+    Iterator,
     Mapping,
     Optional,
     Tuple,
@@ -200,3 +202,7 @@ class MachineManager:
     def run_shell(self, name: Optional[str]):
         config, backend = self.get_config_and_backend(name)
         return backend.run_shell(**config)
+
+    def status(self, name: Optional[str]) -> Iterator[Dict[Any, Any]]:
+        config, backend = self.get_config_and_backend(name)
+        return backend.instances(**config)
--- a/tests/func/machine/__init__.py
+++ b/tests/func/machine/__init__.py
@@ -1,0 +1,10 @@
+import textwrap
+
+CONFIG_TEXT = textwrap.dedent(
+    """\
+        [feature]
+            machine = true
+        ['machine \"foo\"']
+            cloud = aws
+    """
+)
--- a/tests/func/machine/test_machine_config.py
+++ b/tests/func/machine/test_machine_config.py
@@ -5,14 +5,7 @@ import pytest
 
 from dvc.main import main
 
-CONFIG_TEXT = textwrap.dedent(
-    """\
-        [feature]
-            machine = true
-        ['machine \"foo\"']
-            cloud = aws
-    """
-)
+from . import CONFIG_TEXT
 
 
 @pytest.mark.parametrize(
--- a/tests/func/machine/test_machine_status.py
+++ b/tests/func/machine/test_machine_status.py
@@ -1,0 +1,42 @@
+import json
+
+from tpi.terraform import TerraformBackend
+
+from dvc.main import main
+
+from . import CONFIG_TEXT
+
+STATUS = """{
+    "aws_security_group": null,
+    "cloud": "aws",
+    "id": "iterative-2jyhw8j9ieov6",
+    "image": "ubuntu-bionic-18.04-amd64-server-20210818",
+    "instance_gpu": null,
+    "instance_hdd_size": 35,
+    "instance_ip": "123.123.123.123",
+    "instance_launch_time": "2021-08-25T07:13:03Z",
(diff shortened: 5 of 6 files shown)


dvc/command/machine.py
```diff
@@ -143,6 +143,21 @@ def run(self):
         return 0
 
 
+class CmdMachineStatus(CmdBase):
+    def run(self):
+        if self.repo.machine is None:
+            raise MachineDisabledError
+
+        all_status = self.repo.machine.status(self.args.name)
+        for i, status in enumerate(all_status):
```
skshetry, line 152: Use `start=1` instead of `i+1` below if you need 1-based indexing, makes intention clearer.
```suggestion
        for i, status in enumerate(all_status, start=1):
```

dvc/command/machine.py
```diff
@@ -143,6 +143,21 @@ def run(self):
...
+            ui.write(f"instance_num_{i+1}:")
+            keys = list(status.keys())
+            keys.sort()
+            for key in keys:
+                ui.write(f"\t{key:20}: {status[key]}")
```
skshetry, line 157:
```suggestion
            for key, value in sorted(status.items()):
                ui.write(f"\t{key:20}: {value}")
```

pared (approved):

dberenbaum:
@karajan1001 How much extra work would it take to add an asciinema or other video inside of these PRs as an example? It might also make it easier to demo the feature.

pmrowla:
One thing we should consider is sanitizing auth/private data related fields in the CLI commands. In `terraform show` they remove things like SSH private keys and startup scripts from their CLI output (and just display something like `<sensitive data removed>`). They only include those fields when it's in an output format intended for machine/programmatic consumption (i.e. `terraform show -json`)

We should probably do the same thing and not display those fields by default (and include them with `--show-json`).

I'm not sure if there's a straightforward way to get the list of what fields terraform considers sensitive from their CLI, or if it's hard-coded in the provider settings somewhere (@iterative/cml team may know?)

casperdcl:
@pmrowla we hardcode a whitelist in CML https://github.com/iterative/cml/blob/2e9a21ddb4b4eb1261a75ba72f89b98b1972d063/bin/cml/runner.js#L191-L212 and a blacklist in TPI (e.g. https://github.com/iterative/terraform-provider-iterative/blob/732d0919577867d40b7ff635dd0667caf19d3cb9/iterative/resource_machine.go#L106)

karajan1001:
> @pmrowla we hardcode a whitelist in CML https://github.com/iterative/cml/blob/2e9a21ddb4b4eb1261a75ba72f89b98b1972d063/bin/cml/runner.js#L191-L212 and also mark sensitive things in TPI.go (e.g. https://github.com/iterative/terraform-provider-iterative/blob/732d0919577867d40b7ff635dd0667caf19d3cb9/iterative/resource_machine.go#L106) for `terraform show`

Any way to read this white list? Looks like I need to hardcode it first and sync manually first.

karajan1001:
> @karajan1001 How much extra work would it take to add an asciinema or other video inside of these PRs as an example? It might also make it easier to demo the feature.

[asciicast](https://asciinema.org/a/ycr3AtAhHBn9FFtVDCYZ08rd0)

tests/func/machine/__init__.py
```diff
@@ -0,0 +1,10 @@
+import textwrap
+
+CONFIG_TEXT = textwrap.dedent(
+    """\
+        [feature]
+            machine = true
+        ['machine \"foo\"']
```
casperdcl, line 7:
```suggestion
        ['machine "foo"']
```

casperdcl (approved):
`SENSITIVE_FIELDS = {"ssh_private", "startup_script"}` seems enough

dvc/machine/__init__.py
```diff
@@ -200,3 +202,7 @@ def get_sshfs(self, name: Optional[str]):
     def run_shell(self, name: Optional[str]):
         config, backend = self.get_config_and_backend(name)
         return backend.run_shell(**config)
+
+    def status(self, name: Optional[str]) -> Iterator[Dict[Any, Any]]:
```
casperdcl, line 206: perhaps move `SENSITIVE_FIELDS` to this function?

```suggestion
    def status(self, name: str, hide_sensitive: bool=True) -> Iterator[Dict[Any, Any]]:
```

casperdcl: @karajan1001 this was just a stub - the `SENSITIVE_FIELDS` logic would have to be moved to this class

karajan1001: Discussed this with peter yesterday. Here we use will use a white list to limit the showing fields to only in (name, cloud, instance size, IP address) for now. The upcoming `--show-json` can output the whole config in json format in the future.

karajan1001:
[asciicast](https://asciinema.org/a/CKQgYvM8U6hvaQZ2CuNhICRdm)

casperdcl (approved):

dberenbaum:
UI comments:

* It's unclear whether this is intended to show a single machine/instance or all of them. It looks like it's set up to show multiple instances per machine if they exist, which sounds nice for the future. In that case, why not be able to show instances of all machines at once?
* Terraform output doesn't conform with our UI guidelines. If we want to have raw terraform output as an option, I think we also need more typical DVC output to summarize the relevant info in a table or something that looks cleaner and easier to parse. We also probably need to consider which fields are actually relevant. For example, the ssh key info is probably too much detail, especially since DVC has a built-in `dvc machine ssh` command.

casperdcl:
@dberenbaum I think suppressing terraform's CLI output may be an upstream feature request related to https://github.com/iterative/tpi/blob/c7259a8fea023797058deaf487700645df5fe210/tpi/__init__.py#L60

dvc/command/machine.py
```diff
@@ -143,6 +143,22 @@ def run(self):
         return 0
 
 
+class CmdMachineStatus(CmdBase):
+    SHOWN_FIELD = ["name", "cloud", "instance_hdd_size", "instance_ip"]
```
dberenbaum, line 147: Can we include `instance_launch_time`?

dvc/command/machine.py
```diff
@@ -143,6 +143,22 @@ def run(self):
...
+
+    def run(self):
+        if self.repo.machine is None:
+            raise MachineDisabledError
+
+        all_status = self.repo.machine.status(self.args.name)
+        for i, status in enumerate(all_status, start=1):
+            ui.write(f"instance_num_{i}:")
```
dberenbaum, line 155: Could this be a table, more like dvc stage list, to make it easier to parse? We are trying to have default outputs that are both easy to read and parse if possible.

dberenbaum:
> > Any way to read this white list? Looks like I need to hardcode it first and sync manually first.
> 
> `SENSITIVE_FIELDS = {"ssh_private", "startup_script"}` seems enough

Thanks, @casperdcl! Should I open an issue?

Also, is there a "local" option supported for tpi cloud? @karajan1001 @pmrowla Would that be helpful for integration tests and development generally?

casperdcl:
@dberenbaum were you intending to quote https://github.com/iterative/dvc/pull/6649#issuecomment-931371524 instead?

dberenbaum:
> @dberenbaum were you intending to quote [#6649 (comment)](https://github.com/iterative/dvc/pull/6649#issuecomment-931371524) instead?

Yup, thanks. I'll edit it to reference the correct quote.

dvc/command/machine.py
karajan1001: Agreed

pmrowla:
Regarding the terraform CLI output, we originally decided to leave it enabled by default (and not suppress it) since there is no other way for us to reflect status while `terraform ...` commands are running (and we expect them to be slow).

Since we can't hook into the terraform process itself to setup any kind of status callbacks, the only other option on the DVC UI side would be to just display something like a spinner (we can't provide any progressbar-like estimate of when the terraform command will complete).

skshetry:
Another way might be to only show the tail of the output limited to `n` lines, just like the progress bar (on how they clear up the screen), and dump whatever we have on failure.

dberenbaum:
We don't want to fully suppress the output for commands where progress is important (like `dvc machine add`), but for `dvc machine status`, is it needed?

On the other hand, I can imagine some people might want the option to see the raw terraform output if they are familiar with that or want more granular info, so it might not be bad as an option (or maybe include in verbose logging?). If we do show it, it would be helpful to somehow tag it or separate it from output controlled by DVC.

PR #6650, opened 2021-09-20
command: Should not consume 'print' to output default remote info

eggqq007:
1. should not consume 'print' to output default remote info,
   otherwise the --quiet would not work.

### DIFF
--- a/dvc/command/remote.py
+++ b/dvc/command/remote.py
@@ -86,7 +86,7 @@ class CmdRemoteDefault(CmdRemote):
         if self.args.name is None and not self.args.unset:
             conf = self.config.read(self.args.level)
             try:
-                print(conf["core"]["remote"])
+                ui.write(conf["core"]["remote"])
             except KeyError:
                 ui.write("No default remote set")
                 return 1


dvc/command/remote.py
```diff
@@ -86,7 +86,7 @@ def run(self):
         if self.args.name is None and not self.args.unset:
             conf = self.config.read(self.args.level)
             try:
-                print(conf["core"]["remote"])
+                ui.write(conf["core"]["remote"])
```
skshetry, line 89: @eggqq007, could you please describe your use case? Do you want to check if the default is set or not?

skshetry: We don't treat `--quiet` as a no output mode. We still output the data that the user asked for, but silence all the other outputs that DVC usually throws (so that `dvc metrics diff --show-md` still works with `--quiet`, etc).

This is a bit of an edge-case here as you may want to either check what the default is or test if the default remote has been set.

eggqq007: Thanks for the reply. emm, I think  updating the dvc document in https://dvc.org/doc/command-reference/remote/add  is a candidate.  It says "-q, --quiet - do not write anything to standard output. Exit with 0 if no problems arise, otherwise 1."

dberenbaum: Current behavior seems consistent with the guideline in https://github.com/iterative/dvc/discussions/5891. I think it even gives the behavior you are looking for @eggqq007, but it's inverted (there is output printed only if the default remote has been set). The command reference definitely needs an update, though, and probably for other commands also.

skshetry: Yeah, I think this could be interpreted as both, so we should probably not output on quiet here, as it's may not just be the data user asked for. And, as per guideline, we only print the output on `--quiet` where it makes the most sense.

skshetry (approved):

dvc/command/remote.py
dberenbaum: Sorry, that's not what I meant, @skshetry. I was suggesting that we close this PR and make a docs update.

Currently, dvc prints output on `--quiet` if the default remote has been set but not if it hasn't been set. As you mention, the default remote name may be what the user is requesting. Even if the user only wants to know whether the default remote has been set, it seems useful to print something only when the remote has been set, so that users could check not only by exit code but also by whether there was any output. By never printing the output on `--quiet`, isn't it making both use cases more difficult?

skshetry: Ahh, okay. I was mostly sticking to the first point of the guideline:

> If the command's (or, the option) that sole purpose is to display the information, let show it in any case.

I guess we need to clarify if this belongs to the second one or not:

> If the information is very important to the user, print warnings and error messages to the stderr even in --quiet mode wherever appropriate.

I have mostly chosen to err on the side of caution, and support `--quiet` only if the command's/flag's sole purpose is to show that information, if `--quiet` has other uses too, I have kept that so far (as we have discussed, `--quiet` on the `dvc metrics diff` probably is an error on the user's side).

`dvc status` does the same thing when used with `--quiet`, in fact it's what we have been recommending for use in the scripts. 

To resolve this, I guess we need to either suggest alternatives here, or clarify the guideline then.

dberenbaum: I'm not following. Isn't the sole purpose of `dvc remote default` (without any args/options) and `dvc status` to display information? What do you think `dvc remote default --quiet` should show according to the guideline?

efiop: > Isn't the sole purpose of dvc remote default (without any args/options) and dvc status to display information?

Actually no, `dvc status -q` without output is used to indicate whether there are any changes, same as `git status -q`. `dvc remote default` can be used similarly in scripts - to tell user if it has default remote or not. So I think that making `-q` suppress it is justifiable beacuse `dvc remote default` has two purposes: with and without printing.

jorgeorpinel:
Will someone be able to address https://github.com/iterative/dvc.org/issues/2840 too? 🙂

dvc/command/remote.py
dberenbaum: Discussed this as a group. I was wrong about `dvc status -q`, which never prints anything. In that case, it seems fine to do the same here and merge this PR. I don't think it can cause much harm, and it aligns with the current documentation. Sorry for the oversight and all the noise here.

dberenbaum (approved):

skshetry:
Thanks @eggqq007 🙂

Issue #6285, opened 2021-07-06
More informative message when dvc.yaml is malformed


nsorros:
I know we are supposed to let `dvc run` construct the `dvc.yaml` but I tend to write them manually as they are quite easy to construct. Often though I am receiving slightly difficult to decipher errors about the format. An example is below
```
stages:
  train:
    cmd:
      - spacy train config.cfg
    deps:
      - config.cfg
    outs:
      models/spacy-model
```

this throws `ERROR: 'dvc.yaml' format error: extra keys not allowed @ data['stages']['train']['cmd']`

it is quite difficult to decipher that this is missing a `-` in `outs`. not sure why it complains about `cmd` either. would it be possible to give back feedback closer to the error?

dberenbaum:
Related: #5371

PR #6654, opened 2021-09-20
exceptional exceptions

skshetry:
Proposing a `__pretty_exc__` data model for exceptions that can provide prettier exceptions. 

With an example on #5371, the whole thing is a bit fragile though.
Also related: https://github.com/iterative/dvc/issues/6285

---

Screen Shot 2021-09-20 at 14 30 36
Screen Shot 2021-09-20 at 14 31 17
Screen Shot 2021-09-20 at 14 37 42

### DIFF
--- a/dvc/logger.py
+++ b/dvc/logger.py
@@ -142,9 +142,22 @@ class LoggerHandler(logging.StreamHandler):
         super().handleError(record)
         raise LoggingException(record)
 
+    def emit_pretty_exception(self, exc, verbose: bool = False):
+        return exc.__pretty_exc__(verbose=verbose)
+
     def emit(self, record):
         """Write to Tqdm's stream so as to not break progress-bars"""
         try:
+            if record.exc_info:
+                _, exc, *_ = record.exc_info
+                if hasattr(exc, "__pretty_exc__"):
+                    try:
+                        self.emit_pretty_exception(exc, verbose=_is_verbose())
+                        if not _is_verbose():
+                            return
+                    except Exception:  # fallback to usual method
+                        pass
+
             msg = self.format(record)
             Tqdm.write(
                 msg, file=self.stream, end=getattr(self, "terminator", "\n")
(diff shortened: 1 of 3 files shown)


dvc/logger.py
```diff
@@ -142,9 +142,22 @@ def handleError(self, record):
         super().handleError(record)
         raise LoggingException(record)
 
+    def emit_pretty_exception(self, exc, verbose: bool = False):
+        return exc.__pretty_exc__(verbose=verbose)
```
skshetry, line 146: A better solution would be to just handle it in `dvc.main` and introduce `DvcException.pretty_print()` kinda API, but right now, we just use `logger.exception` everywhere which makes it tricky, will have to get rid of that. For now, we are stuck with this hack.

pared (approved):
Nice!

efiop (approved):
+1, great stuff.

skshetry:
> +1, great stuff. Btw, IIRC rich is doing that same thing? Or was it some other project?

Rich has a [`__rich_repr__`](https://rich.readthedocs.io/en/stable/pretty.html#rich-repr-protocol) protocol with which you can add support for rich output for custom objects.

This might be temporary, I am more inclined now to introduce `DvcException.show()` similar to in [`Click`](https://click.palletsprojects.com/en/8.0.x/exceptions/#which-exceptions-exist), which we can handle in `dvc.main.main`. But this requires us to get rid of `logger.exception` calls that we have.

dberenbaum:
@skshetry Any idea why I get the old output when trying out the example in #6285?

```
$ dvc -v status
ERROR: 'dvc.yaml' format error: extra keys not allowed @ data['stages']['train']['cmd']
$ dvc doctor
DVC version: 2.7.4+cace8e
Platform: Python 3.9.5 on macOS-11.5.2-arm64-arm-64bit
Supports:
        s3 (s3fs = 2021.8.0, boto3 = 1.17.5)
Cache directory: apfs on /dev/disk3s1s1
Caches: local
Remotes: local, s3
Repo: dvc, git
```

skshetry:
@dberenbaum, schema validation is not in this PR, only limited to the yaml syntax errors.

PR #6657, opened 2021-09-20
fix dvc[all]

casperdcl:
`requirements["all"]` should be a `list[str], not `list[list[str]]`

- related https://github.com/iterative/dvc-snap/issues/24


setup.py
```diff
@@ -37,9 +38,14 @@ class build_py(_build_py):
...
-            fobj.write("# AUTOGENERATED at build time by setup.py\n")
-            fobj.write(f'__version__ = "{version}"\n')
+        (Path(path) / "version.py").write_text(
+            dedent(
+                """\
+                # AUTOGENERATED at build time by setup.py
+                fobj.write(f'__version__ = "{version}"
+                """
+            )
+        )
```
casperdcl, line 48: just keeping `pylint` happy

setup.py
```diff
@@ -37,9 +38,14 @@ class build_py(_build_py):
...
         path = os.path.join(self.build_lib, "dvc")
         self.mkpath(path)
-        with open(os.path.join(path, "version.py"), "w") as fobj:
-            fobj.write("# AUTOGENERATED at build time by setup.py\n")
-            fobj.write(f'__version__ = "{version}"\n')
+        (Path(path) / "version.py").write_text(
+            dedent(
+                """\
+                # AUTOGENERATED at build time by setup.py
+                fobj.write(f'__version__ = "{version}"
```
skshetry, line 45: Missing closing parentheses
```suggestion
                fobj.write(f'__version__ = "{version}"')
```

casperdcl: no just missing `Ctrl`+`S` :D

setup.py
```diff
@@ -35,11 +36,16 @@
 #   - https://github.com/python/mypy
 class build_py(_build_py):
     def pin_version(self):
-        path = os.path.join(self.build_lib, "dvc")
-        self.mkpath(path)
-        with open(os.path.join(path, "version.py"), "w") as fobj:
-            fobj.write("# AUTOGENERATED at build time by setup.py\n")
-            fobj.write(f'__version__ = "{version}"\n')
+        ver_file = Path(self.build_lib) / "dvc" / "version.py"
```
skshetry, line 39: Maybe just the `file` is enough?

casperdcl: idk lol; not even the point of this PR; darn it `pylint`

efiop: Pyint started complaining about it in this PR? If so, just add flags to ignore it in these lines. Let's not change unrelated logic, please.

casperdcl: I assume the `pytlint` CI is currently configured to complain about touched files in their entriety so if you don't want PRs to slowly fix unrelated issues then isn't it better to switch off that option?

I'd prefer to either

- revert in entirety & merge with failing pylint, or
- merge as-is including tangential fixes

in contrast AFAIK

- reverting fixes, adding tangential ignore directives, and merging

is the worst of all combinations unless I'm missing something?

efiop: @casperdcl was pre-commit ci failing or our `Tests / lint`?

setup.py
```diff
@@ -36,9 +36,9 @@
 #   - https://github.com/python/mypy
 class build_py(_build_py):
     def pin_version(self):
-        path = os.path.join(self.build_lib, "dvc")
-        self.mkpath(path)
```
efiop, line 39: A little worried about using our mkdir instead of self.mkpath, as it might have some hidden logic/mode/etc in it.

setup.py
casperdcl: `pre-commit ci` as well as local `pre-commit` hook.

setup.py
casperdcl: fixed

efiop (approved):

PR #6662, opened 2021-09-21
webhdfs: migrate to fsspec

### LINKED ISSUE #6583: webhdfs: migrate to fsspec
To complete the fsspec migration, we need to migrate webhdfs to the fsspec.

isidentical:
Resolves #6583, awaiting an fsspec release.


dvc/fs/webhdfs.py
```diff
@@ -66,115 +32,71 @@ def _get_kwargs_from_urls(urlpath):
...
-            logger.debug("URL: %s", http_url)
-
-            if self.token is not None:
-                client = hdfs.TokenClient(http_url, token=self.token, root="/")
-            else:
-                client = hdfs.InsecureClient(
-                    http_url, user=self.user, root="/"
+            config = Config(path)
+        except HdfsError:
+            return None
```
efiop, line 44: For the record: discussed that we might just drop hdfscli config, since we are no longer using it as a backend.

dvc/fs/webhdfs.py
```diff
@@ -66,115 +32,71 @@ def _get_kwargs_from_urls(urlpath):
...
+        return WebHDFS(**self.fs_args)
 
     def checksum(self, path_info):
-        return HashInfo(
-            "checksum",
-            self.hdfs_client.checksum(path_info.path)["bytes"],
-            size=self.hdfs_client.status(path_info.path)["length"],
-        )
+        path = self._with_bucket(path_info)
+        ukey = self.fs.ukey(path)
```
efiop, line 98: Will need to properly distinguish between `checksum` and `ukey` in dvc in the future. I did a mistake of using `checksum` for refodb in staging, while we could really use ukey there instead.

tests/remotes/webhdfs.py
```diff
@@ -58,155 +57,7 @@ def read_text(self, encoding=None, errors=None):
...
-
-        makedirs(os.path.dirname(local_path), exist_ok=True)
-        with open(local_path, "wb") as writer:
-            with self.read(hdfs_path, **kwargs) as reader:
-                for chunk in reader:
-                    writer.write(chunk)
-
-
-@pytest.fixture
-def webhdfs(mocker):
```
efiop, line 209: A bit sad that we no longer have mocked webhdfs, but we could think about mocking with localfs in the future. Or maybe mocking a whole webhdfs server. Not for this PR, of course.

efiop:
@mergifyio rebase

Issue #6627, opened 2021-09-16
commands: enhance our profiling capabilities


pared:
We should consider using a framework that has better multi-thread capabilities than currently used cProfile.

Some options:
https://github.com/sumerc/yappi
https://github.com/benfred/py-spy
http://kcachegrind.sourceforge.net/html/Home.html

Related: 
#5594
https://github.com/iterative/dvc/issues/2891#issuecomment-561951733

pared:
This has been hanging for some time. Let's close it. We can reopen if `yappi` will be not enough in the future.

PR #6664, opened 2021-09-21
debug: add yappi, callgrind output support

skshetry:
#### How to use:
0. `pip install yappi`
1. add `--yappi` at the end of any command.
2. Install `kcachegrind`/`qcachegrind`.
3. Open the file, which is in the form of `callgrind.dvc-{id}.out`format with `{k,q}cachegrind`.

I haven't really used this that much, so you'd have to see how this works. :)

Related: https://github.com/iterative/dvc/issues/6627 (not closing it yet, let's see how this works).

Screen Shot 2021-09-21 at 22 20 34

### DIFF
--- a/dvc/_debug.py
+++ b/dvc/_debug.py
@@ -6,6 +6,34 @@ if TYPE_CHECKING:
 
 
 @contextmanager
+def yappi_profile(save_output: bool = False, wall_clock: bool = True):
+    try:
+        import yappi  # pylint: disable=import-error
+    except ImportError:
+        print("Failed to run profiler, yappi is not installed")
+        yield
+        return
+
+    import time
+
+    yappi.set_clock_type("wall" if wall_clock else "cpu")
+
+    yappi.start()
+    yield
+    yappi.stop()
+
+    if save_output:
+        stats = yappi.get_func_stats()
+        uid = time.strftime("%Y%m%d_%H%M%S")
+        stats.save(f"callgrind.dvc-{uid}.out", "callgrind")
+    else:
+        yappi.get_func_stats().print_all()
+        yappi.get_thread_stats().print_all()
+
+    yappi.clear_stats()
+
+
+@contextmanager
 def instrument(html_output=False):
     """Run a statistical profiler"""
     try:
@@ -70,6 +98,8 @@ def debugtools(args: "Namespace" = None, **kwargs):
             stack.enter_context(profile(kw.get("cprofile_dump")))
         if kw.get("instrument") or kw.get("instrument_open"):
             stack.enter_context(instrument(kw.get("instrument_open", False)))
+        if kw.get("yappi"):
+            stack.enter_context(yappi_profile(save_output=True))
         yield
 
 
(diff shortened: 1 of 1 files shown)


efiop:
Side note: would be amazing to support these flags in `dvc-bench`. https://github.com/iterative/dvc-bench/issues/301

efiop (approved):
Great stuff! 🔥

dvc/_debug.py
```diff
@@ -5,6 +5,34 @@
...
+    yappi.set_clock_type("wall" if wall_clock else "cpu")
+
+    yappi.start()
+    yield
+    yappi.stop()
+
+    if save_output:
+        stats = yappi.get_func_stats()
+        uid = time.strftime("%Y%m%d_%H%M%S")
+        stats.save(f"callgrind.dvc-{uid}.out", "callgrind")
```
skshetry, line 28: Went autogenerated way with this yappi, let me know if you want to use `--cprofile-dump dump.prof` like flags 😄.

skshetry:
> Let's add to https://github.com/iterative/dvc/wiki/Debugging-DVC ?

Working on it. :)

dvc/_debug.py
efiop: Might need that option for https://github.com/iterative/dvc-bench/issues/301 , so as long as we could introduce it later - it is fine.

skshetry:
Btw, we should probably have debug builds for binaries.

dvc/_debug.py
skshetry: Let me work on that then. :)

dberenbaum:
Any example output you can show? Or is that part of adding it to the Wiki?

skshetry:
> Any example output you can show? Or is that part of adding it to the Wiki?

Added example output (it is part of kcachegrind though). There's not much to this than adding `--yappi` to the command, and then later opening the `callgrind.*` file through kcachegrind.

dvc/_debug.py
```diff
@@ -70,6 +99,13 @@ def debugtools(args: "Namespace" = None, **kwargs):
             stack.enter_context(profile(kw.get("cprofile_dump")))
         if kw.get("instrument") or kw.get("instrument_open"):
             stack.enter_context(instrument(kw.get("instrument_open", False)))
+        if kw.get("yappi"):
+            from datetime import datetime
+
+            output = "callgrind.dvc-{0:%Y%m%d}_{0:%H%M%S}.out"
+            stack.enter_context(
+                yappi_profile(path=lambda: output.format(datetime.now()))
```
efiop, line 107:
```suggestion
                yappi_profile(path_factory=lambda: output.format(datetime.now()))
```
?

efiop (approved):
🔥 

These options are hidden anyway, so we can tweak them as we go.

dvc/_debug.py
skshetry: yeah, we could do that too, but I am optimizing the API here. Note that this `yappi_profile` (along with all the others) can be used as a decorator and a context manager on any functions that you want to profile:

```python
c = count(start=1)


@yappi_profile(path=lambda: next(c))
def _diff(*args, **kwargs):
    pass
```

PR #6677, opened 2021-09-23
ssh: use a docker-based testing environment

### LINKED ISSUE #6673: ssh: switch to a docker-based alternative from mock-ssh server
It appears to be that we are in an impasse caused by the oddities of asyncssh + mock-ssh-server. There were already some bugs (e.g decoding commands) that made us patch the actual mock-ssh-server to function, but while researching #6514 I noticed that it  works very weirdly due to a couple issues I was able to observe (path separator conflict, trying to `mkdir()` root of the filesystem and getting permission error etc.). Due to these and making this more maintainable and similiar to it's real environment, we should switch to a docker-based solution

isidentical:
Resolves #6673

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -23,7 +23,7 @@ REMOTES = {
     "http": True,
     "oss": False,
     "s3": False,
-    "ssh": True,
+    "ssh": False,
     "webdav": True,
 }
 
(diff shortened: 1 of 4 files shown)


tests/func/test_fs.py
```diff
@@ -468,6 +469,7 @@ def test_download_callback(tmp_dir, dvc, cloud, local_cloud):
     fs = cls(**config)
 
     (tmp_dir / "to_upload").write_text("foo")
+    cloud.mkdir(exist_ok=True, parents=True)
```
skshetry, line 472: If we need for ssh, this should already be created in `ssh` fixture. Note  that `oss` fixture does not implement anything, it's mostly a wrapper for config right now, so it fails on `mkdir`.

efiop (approved):

PR #6661, opened 2021-09-21
merge yaml parsing and schema validation into one, aka strictyaml.load

skshetry:
Closes https://github.com/iterative/dvc/issues/6290.

To solve #6285 and #5371 (i.e. better error messaging from schema errors and yaml syntax errors), I reckoned that we need to combine both yaml parsing and schema validation into one single pipeline. So, this merges both `parse_yaml` and `validate_schema` into one `strictyaml.load()` and may help us in providing better error messages. This is really useful for  loading `dvc.yaml`/`dvc.lock` and `.dvc` files as they are validated YAML files.

Moves previous changes from #6654 into strictyaml.

Schema error messaging is not in this PR, which will come after this gets merged. #6290 has been fixed by this change, as it now raises proper `EncodingError`. 

* [x] ❗ I have followed the [Contributing to DVC](https://dvc.org/doc/user-guide/contributing/core) checklist.

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here.

Thank you for the contribution - we'll try to review it as soon as possible. 🙏

PR #6679, opened 2021-09-23
better exception messages for schema errors

skshetry:
Related PRs
- #5371
- #6285
- #6654
- #6661

More details are in inlined comments.


dvc/dvcfile.py
```diff
@@ -39,11 +37,6 @@ def __init__(self, path, pipeline_file=False):
         )
 
 
-class LockfileCorruptedError(DvcException):
```
skshetry, line 42: Got rid of LockfileCorruptedError and StageFileFormatError.

dvc/utils/strictyaml.py
```diff
@@ -6,24 +6,43 @@
...
+
+def make_relpath(path: str) -> str:
+    import os
+
+    from dvc.utils import relpath
+
+    rel = relpath(path)
+    prefix = ""
+    if not rel.startswith(".."):
+        prefix = "./" if os.name == "posix" else ".\\"
```
skshetry, line 44:
```
'dvc.yaml' is corrupted.
```

^ This kind of message was confusing, at least I find it to be. As it may be talking about `dvc.yaml` file in a different place in the workspace or in the user's working directory. Added `./dvc.yaml`/`.\dvc.yaml` to make it obvious that it's relative path.

dvc/utils/strictyaml.py
```diff
@@ -51,13 +73,15 @@ def __init__(self, path: str, yaml_text: str, exc: Exception) -> None:
         self.path: str = path
         self.yaml_text: str = yaml_text
         self.exc: Exception = exc
+
+        merge_conflicts = merge_conflict_marker.search(self.yaml_text)
+        self.hint = " (possible merge conflicts)" if merge_conflicts else ""
```
skshetry, line 78: Now we can detect if we have a merge conflicts on the file.

dvc/utils/strictyaml.py
```diff
@@ -101,25 +125,148 @@ def prepare_code(mark: "StreamMark") -> "Syntax":
...
-            ui.error_write(message, styled=True)
 
+        rel = make_relpath(self.path)
+        message = f"'{rel}' is corrupted{self.hint}"
+        lines.insert(0, f"[red]{message}.")
+        for line in lines:
+            ui.error_write(line, styled=True)
+
+
+def determine_linecol(
```
skshetry, line 140: Line numbers for schema validation errors may be inaccurate sometimes. Please read the doc below for more technical information.

tests/func/utils/test_strict_yaml.py
```diff
@@ -0,0 +1,291 @@
...
+  group:
+    foreach: 3
+    do:
+      cmd: python script${i}.py
+"""
+
+FOREACH_SCALAR_VALUE_OUTPUT = """\
+'./dvc.yaml' validation failed.
+
+expected dict, in stages -> group -> foreach, line 3, column 5
```
skshetry, line 164: Should have been something like following, but we have issues with validating "EitherOr" schema (at least it starts throwing too many errors in those cases).
```console
expected dict and list, in stages -> group -> foreach, line 3, column 5
```

dvc/dvcfile.py
skshetry: To clarify, this was removed because the pretty_print is implemented on the `YAMLValidationError` side, it's quite tricky to pass that context of data/text/path, etc. to other exceptions that are reraised. And, that too in two of these exceptions.

tests/func/utils/test_strict_yaml.py
```diff
@@ -0,0 +1,310 @@
...
+    mocker.patch.object(console, "print", print_with_fixed_width)
+
+    tmp_dir.gen("dvc.yaml", text)
+
+    capsys.readouterr()  # clear outputs
+    assert main(["stage", "list"]) != 0
+    out, err = capsys.readouterr()
+
+    assert not out
+    assert expected in err
```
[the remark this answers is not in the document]
skshetry: Did you try locally?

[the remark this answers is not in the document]
skshetry: Maybe we need to update rich, `pip install -U rich`.

dvc/utils/strictyaml.py
skshetry: Absolute path may be quite lengthy. We don't have guidelines for that, but for UI, we have always used relpaths.

tests/func/utils/test_strict_yaml.py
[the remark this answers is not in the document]
skshetry: Or, you may be missing `dpath`. That was the issue for me, if I recall correctly.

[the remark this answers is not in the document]
dberenbaum: Got it, must have been a dependency issue, thanks!

tests/func/utils/test_strict_yaml.py
```diff
@@ -0,0 +1,310 @@
...
+stages:
+  stage1
+    cmd: python script.py
+"""
+
+MAPPING_VALUES_NOT_ALLOWED_OUTPUT = """\
+'./dvc.yaml' is corrupted.
+
+Mapping values are not allowed {}, in line 3, column 8
+  3 │   cmd: python script.py""".format(
```
dberenbaum, line 37: Would it be helpful to show line 2 also?

tests/func/utils/test_strict_yaml.py
```diff
@@ -0,0 +1,310 @@
...
+    outs:
+      - logs:
+          cache: false
+      metrics:
+"""
+
+NO_HYPEN_INDICATOR_IN_BLOCK_OUTPUT = """\
+'./dvc.yaml' is corrupted.
+
+While parsing a block collection, in line 5, column 7
```
dberenbaum, line 55: I find this a lot easier to understand by looking at the entire `outs:` block.

tests/func/utils/test_strict_yaml.py
```diff
@@ -0,0 +1,310 @@
...
+  - logs:
+      cache: false
+      persist: true
+      remote:
+"""
+
+NULL_VALUE_ON_OUTS_OUTPUT = """\
+'./dvc.yaml' validation failed: 2 errors.
+
+expected a dictionary, in stages -> cmd, line 2, column 3
```
dberenbaum, line 130: Why are this one and the next few missing a stage name?

tests/func/utils/test_strict_yaml.py
dberenbaum: Not ideal, but at least it points to the right place!

dberenbaum (commented):
🔥 This looks great!

What do you think about showing more lines generally, even when the specific line is known? I think showing a line above and below, or the whole parent block, provides a lot of useful context, and we could still highlight the primary line.

tests/func/utils/test_strict_yaml.py
skshetry: This is a syntax error. So we don't really have a way to go one node above, as we don't even get a semi-parsed dictionary from ruamel. The lines that we show are what we get from the error directly. However, we can show `n` lines before and after that point.

tests/func/utils/test_strict_yaml.py
skshetry: 🤦🏼‍♂️

tests/func/utils/test_strict_yaml.py
skshetry: For the schema errors, we can traverse upwards, but we don't have a way to know the line ranges, `endpos` in particular. There may be a workaround, but they are a slippery slope.

tests/func/utils/test_strict_yaml.py
skshetry: I have plans to work on that in the near future. :)

dvc/utils/strictyaml.py
dberenbaum: That works, should we document that as a guideline?

tests/func/utils/test_strict_yaml.py
dberenbaum: Yeah, personally I'd be in favor of generally including a line or two above and below and highlighting the error line to give more context, but curious if you think this is helpful or too noisy.

skshetry: It may be noisy, as it will be about 5lines of text per error message. With 2 errors on average, it's 12 lines of the message (2*5 lines + 1 main header line + 1 separation line), which is half of `80*24` standard terminal.

I haven't seen other programming languages do this, but this is YAML 😄, so adding context does make sense, although it's better to be precise in error messages than the code snippets. Everything after that message is a best effort.

dberenbaum: True. There's no obvious right answer, so up to you.

pmrowla: At least for code related things, I prefer when linting tools default to displaying the single line and then allow the user to configure displaying `n` lines before after via env vars or command flags

pmrowla (approved):

PR #6681, opened 2021-09-24
exp init: support interactive and explicit options

skshetry:
This PR adds support for `--interactive` and `--explicit` options. This does not have support for templates, which still requires some thought. This change also makes `dvclive` optional, it'll be only applied if the `--live` flag is used or the user responds interactively (where it'll still be asked for by default).

Also adds the `--force` flag to mirror `dvc stage add`.

This includes stuff we have discussed in https://github.com/iterative/dvc/pull/6637#issuecomment-924897599 so far.
If we still need templates, it could be done over this.

Supersedes #6630 and #6637.

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -1,14 +1,14 @@
 import argparse
 import logging
 import os
-from collections import Counter, OrderedDict, defaultdict
+from collections import ChainMap, Counter, OrderedDict, defaultdict
 from datetime import date, datetime
 from fnmatch import fnmatch
 from typing import TYPE_CHECKING, Dict, Iterable, Optional
 
-from funcy import lmap
+from funcy import compact, lmap, post_processing
+from rich.prompt import Prompt
 
-from dvc import prompt
 from dvc.command import completion
 from dvc.command.base import CmdBase, append_doc_link, fix_subparsers
 from dvc.command.metrics import DEFAULT_PRECISION
@@ -669,7 +669,7 @@ class CmdExperimentsGC(CmdRepro):
         logger.warning(msg)
 
         msg = "Are you sure you want to proceed?"
-        if not self.args.force and not prompt.confirm(msg):
+        if not self.args.force and not ui.confirm(msg):
             return 1
 
         removed = self.repo.experiments.gc(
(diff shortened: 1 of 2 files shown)


dvc/command/experiments.py
```diff
@@ -800,45 +844,100 @@ class CmdExperimentsInit(CmdBase):
...
+            "models": "Path to a [b]model[/b] file/directory",
+            "metrics": "Path to a [b]metrics[/b] file",
+            "params": "Path to a [b]parameters[/b] file",
+            "plots": "Path to a [b]plots[/b] file/directory",
+            "live": "Path to log [b]dvclive[/b] outputs",
+        }
+        message = (
+            "This command will guide you to set up your first stage in "
+            "[green]dvc.yaml[/green].\n"
+        )
```
skshetry, line 863: There were some suggestions to do `dvc init` if we detect that there's no `dvc` repo. `dvc exp init` is likely going to be used for the first time, so I added that up, but we could autodetect the presence of `dvc.yaml` file as well to have heading or not.

dberenbaum: Awesome, that sounds great!

dberenbaum (commented):
The non-dvclive scenario looks good, although I think we could still guide the users more with the messaging, assuming that they don't know anything about dvc experiments, stages, etc.

For dvclive, I still think it needs to be separate or handled a bit differently because:
* It should assume the models output has `checkpoint: true`.
* It should not ask for metrics or plots since dvclive should handle these. In more complex stages, maybe there could be scenarios where users have additional custom metrics/plots, but I think this would confuse new users.

skshetry:
>  although I think we could still guide the users more with the messaging, assuming that they don't know anything about dvc experiments, stages, etc.

Could you please suggest some examples? Also it may be a bit noisy.

> For dvclive, I still think it needs to be separate or handled a bit differently because:
> 
> * It should assume the models output has `checkpoint: true`.
> * It should not ask for metrics or plots since dvclive should handle these. In more complex stages, maybe there could be scenarios where users have additional custom metrics/plots, but I think this would confuse new users.

I think the easiest way to do that is to ask the user about `dvclive` before metrics/plots and then skip prompts after that.

dberenbaum:
> Could you please suggest some examples? Also it may be a bit noisy.

See the comment in https://github.com/iterative/dvc/pull/6637#issuecomment-924897599:

> Suggested wording: Path to data file/directory on which the experiment depends [leave blank for "data", type "n" to omit]: (we might need to state "on which the experiment depends" or similar since "data" is very generic). We can also tweak this later if we agree on the blank vs. "n" input behavior.

As mentioned there, we could tweak it later, so no need to block this PR to agree on the wording.

---

> I think the easiest way to do that is to ask the user about `dvclive` before metrics/plots and then skip prompts after that.

Why not just ask at the start like: "Do you want to log training steps iteratively with dvclive?"

How do you anticipate this working for non-interactive mode?

I'm still wondering whether the template option is needed (it could be renamed) to separate the dvclive scenario and provide flexibility to add other scenarios in the future. What do you think?

PR #6684, opened 2021-09-24
state: cleanup redundant or localfs-specific logic

efiop:
Pre-requisite for making state work with other filesystems.

### DIFF
--- a/dvc/state.py
+++ b/dvc/state.py
@@ -79,11 +79,6 @@ class State(StateBase):  # pylint: disable=too-many-instance-attributes
         if not isinstance(fs, LocalFileSystem):
             return
 
-        assert isinstance(path_info, str) or path_info.scheme == "local"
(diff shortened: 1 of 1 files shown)


dvc/state.py
```diff
@@ -111,17 +106,12 @@ def get(self, path_info, fs):
...
-        path = os.fspath(path_info)
-
-        # NOTE: use os.path.exists instead of LocalFileSystem.exists
-        # because it uses lexists() and will return True for broken
-        # symlinks that we cannot stat() in get_mtime_and_size
-        if not os.path.exists(path):
+        try:
+            mtime, size = get_mtime_and_size(path_info, fs, self.dvcignore)
+        except FileNotFoundError:
             return None
```
efiop, line 112: Uses `stat()` inside, so this is no longer relevant.

PR #6706, opened 2021-09-29
benedict: monkeypatch, performance improvements, bugfixes

### LINKED ISSUE #6423: See how much new version of python-benedict impacts us

### LINKED ISSUE #6476: exp run: unexpected error - cannot represent an object: {'lr': 0.1}

skshetry:
Fixes #6476, fixes #6423 and closes #6484.

### DIFF
--- a/dvc/utils/collections.py
+++ b/dvc/utils/collections.py
@@ -82,19 +82,10 @@ def chunk_dict(d: Dict[_KT, _VT], size: int = 1) -> List[Dict[_KT, _VT]]:
 
 def merge_params(src: Dict, to_update: Dict) -> Dict:
     """Recursively merges params with benedict's syntax support in-place."""
-    from benedict import benedict
+    from ._benedict import benedict
 
     data = benedict(src)
-    if src:
-        data.merge(to_update, overwrite=True)
-    else:
-        # NOTE: the following line may seem like an unnecessary duplication
-        # data.merge might affect the `src` if it's not empty, so we cannot
-        # check `if src` later, as it may have been mutated already.
-        data.merge(to_update, overwrite=True)
-        # benedict has issues keeping references to an empty dictionary
-        # see: https://github.com/iterative/dvc/issues/6374.
-        src.update(data)
+    data.merge(to_update, overwrite=True)
     return src
 
 
--- a/tests/unit/utils/test_collections.py
+++ b/tests/unit/utils/test_collections.py
@@ -1,4 +1,7 @@
 # pylint: disable=unidiomatic-typecheck
+import json
+from json import encoder
+
 import pytest
 from mock import create_autospec
 
@@ -8,6 +11,7 @@ from dvc.utils.collections import (
     merge_params,
     validate,
 )
+from dvc.utils.serialize import dumps_yaml
 
 
 class MyDict(dict):
@@ -143,6 +147,12 @@ def test_post_validate_decorator(mocker):
     assert result == [1, 2]
 
 
+def is_serializable(d):
+    json.dumps(d)
+    dumps_yaml(d)
+    return True
+
+
 @pytest.mark.parametrize(
     "changes, expected",
     [
@@ -195,6 +205,8 @@ def test_merge_params(changes, expected):
     merged = merge_params(params, changes)
     assert merged == expected == params
     assert params is merged  # references should be preserved
+    assert encoder.c_make_encoder
+    assert is_serializable(params)
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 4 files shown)


dvc/utils/collections.py
```diff
@@ -82,19 +82,10 @@ def chunk_dict(d: Dict[_KT, _VT], size: int = 1) -> List[Dict[_KT, _VT]]:
...
-    if src:
-        data.merge(to_update, overwrite=True)
-    else:
-        # NOTE: the following line may seem like an unnecessary duplication
-        # data.merge might affect the `src` if it's not empty, so we cannot
-        # check `if src` later, as it may have been mutated already.
-        data.merge(to_update, overwrite=True)
-        # benedict has issues keeping references to an empty dictionary
-        # see: https://github.com/iterative/dvc/issues/6374.
-        src.update(data)
```
skshetry, line 97: This `else` was for a bug in the `benedict`, and it has been fixed. To resolve this bug, that monkeypatch I mentioned was introduced 😢.  We no longer need this `else` block, and tests should cover this if it breaks again in future.

tests/unit/utils/test_collections.py
```diff
@@ -203,10 +215,20 @@ def test_merge_params(changes, expected):
...
     assert params is merged  # references should be preserved
+    assert encoder.c_make_encoder
+    assert is_serializable(params)
+
+
+def test_benedict_rollback_its_monkeypatch():
+    from dvc.utils._benedict import benedict
+
+    assert benedict({"foo": "foo"}) == {"foo": "foo"}
+    assert encoder.c_make_encoder
```
skshetry, line 234: Test for the monkeypatch we rollback.

tests/unit/utils/test_collections.py
```diff
@@ -203,10 +215,20 @@ def test_merge_params(changes, expected):
         [{"foo": "baz"}, {"foo": "baz"}],
         [{"foo": "baz", "goo": "bar"}, {"foo": "baz", "goo": "bar"}],
         [{"foo[1]": ["baz", "goo"]}, {"foo": [None, ["baz", "goo"]]}],
+        [{"foo.bar": "baz"}, {"foo": {"bar": "baz"}}],
```
skshetry, line 218: Testing that it fixes #6476.

efiop (commented):
We've also talked that it might be easier to throw it out maybe, wdyt about that option right now?

skshetry:
With this https://github.com/iterative/dvc/pull/6521 also requiring significant part of benedict and that we need to imitate it's syntax, I think it's much easier to workaround performance issue.

I do believe that we should have these small utilities in DVC itself, and try to avoid small dependencies moving forward.

pmrowla (approved):
LGTM

It does seem like in the long run we should try to get away from having these kinds of dependencies in DVC, but in the short term benedict does work for us (with this patch), and this will unblock #6521

Issue #6488, opened 2021-08-25
metrics/params/exp: should the `--show-*` flags have a show prefix at all?


skshetry:
In `dvc metrics/params/exp`, we have `show-json/show-md/show-csv` flags. 

Should we just have `--json/--markdown/--md/--csv` flag instead? This is much easier to type, `show` is redundant here.

eg: `dvc exp show --show-json`

We can easily have an alias in the argparse for the flags for backward compatibility till 3.0.

PR #6711, opened 2021-09-30
cli: add short alias for --show-{csv,json,md}

skshetry:
As `--csv`/`--json`/`--md`

Fixes https://github.com/iterative/dvc/issues/6488. My plan is to replace options from docs soon and remove these options in 3.0 (though we could keep it, they won't cause any issues).

Documentation can be done after we merge this PR, as this is just an alias.


dvc/command/diff.py
```diff
@@ -200,9 +201,11 @@ def add_parser(subparsers, parent_parser):
         default=False,
     )
     diff_parser.add_argument(
+        "--md",
```
skshetry, line 204: I always try to use `-md` (single dash) on these abbreviations, so I am not a fan of this, but `--markdown` is too long. And `--md` is what we use in CML as well.

dberenbaum (approved):
Looks great! Are you planning a corresponding docs PR?

skshetry:
> Looks great! Are you planning a corresponding docs PR?

yes, I'll submit one tomorrow.

PR #6716, opened 2021-09-30
exp init: introduce `--with={default,live}` flag

skshetry:
This flag will select the different mode of outputs of stage.
With `--with=live`, it'll enable `live` setting in the stage.
On interactive mode, it'll always ask user if they want to
use live (unless `--with=live` is specified), and then it'll
skip metrics/plots question and moves on to ask for path to
dvclive path.

Also, if live setting is enabled in the stage, it'll set the
`models` to be a checkpoint output.

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -7,7 +7,7 @@ from fnmatch import fnmatch
 from typing import TYPE_CHECKING, Dict, Iterable, Optional
 
 from funcy import compact, lmap, post_processing
-from rich.prompt import Prompt
+from rich.prompt import Confirm, Prompt
 
 from dvc.command import completion
 from dvc.command.base import CmdBase, append_doc_link, fix_subparsers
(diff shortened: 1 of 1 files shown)


dvc/command/experiments.py
```diff
@@ -868,6 +849,57 @@ def init_interactive(self, defaults=None, show_heading: bool = False):
...
+                "code": "Path to a [b]code[/b] file/directory",
+                "data": "Path to a [b]data[/b] file/directory",
+                "models": "Path to a [b]model[/b] file/directory",
+            },
+            defaults=defaults,
+        )
+
+        if not live:
+            ui.error_write()
+            live = Confirm.ask(
```
skshetry, line 875: Will need to iterate on this separately.

dvc/command/experiments.py
```diff
@@ -1548,6 +1592,13 @@ def add_parser(subparsers, parent_parser):
     experiments_init_parser.add_argument(
         "--live", help="Path to log dvclive outputs for your experiments"
     )
+    experiments_init_parser.add_argument(
+        "--with",
```
skshetry, line 1596: Not so sure about this flag, can change to anything better.

dvc/command/experiments.py
```diff
@@ -868,6 +849,57 @@ def init_interactive(self, defaults=None, show_heading: bool = False):
...
+                "data": "Path to a [b]data[/b] file/directory",
+                "models": "Path to a [b]model[/b] file/directory",
+            },
+            defaults=defaults,
+        )
+
+        if not live:
+            ui.error_write()
+            live = Confirm.ask(
+                "Do you want to log training steps iteratively with "
```
dberenbaum, line 876: I don't think this makes sense for `dvc exp init -i --with=default`.

dberenbaum (commented):
It feels a little clunky to me to ask about dvclive in the middle of entering paths. I think it's better to require `--with` non-interactively (`--name`, `--run`, etc. aren't covered). It also fits with the description of `--interactive`: `Prompt for values that are not provided`.

Minor note: should `--template` be removed at this point?

dvc/command/experiments.py
skshetry: I prefer `--type`, `--stage` has a different meaning. Compared to `--format`, `--type` is more simpler.

dvc/command/experiments.py
```diff
@@ -868,6 +849,57 @@ def init_interactive(self, defaults=None, show_heading: bool = False):
...
+            "[green]dvc.yaml[/green].\n"
+        )
+        if show_heading:
+            ui.error_write(message, styled=True)
+
+        yield from self._prompt(
+            {
+                "cmd": "[b]Command[/b] to execute",
+                "code": "Path to a [b]code[/b] file/directory",
+                "data": "Path to a [b]data[/b] file/directory",
```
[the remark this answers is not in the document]
skshetry: I wanted to separate dependencies and then pipeline artifacts, but I don't need to. 😄

skshetry:
> It feels a little clunky to me to ask about dvclive in the middle of entering paths. I think it's better to require --with non-interactively (--name, --run, etc. aren't covered). 

I felt the same way, I thought it was just me. I'll make it a requirement then. :)

dberenbaum (approved):
LGTM!

My only concern is that we may need to come back to the idea of custom templates again in the future. For example, users who mainly plot image files might want plots with `cache: true`. We don't want to turn this into `dvc stage add`, but it seems useful to enable a consistent stage structure, especially if it could be shared across projects. This seems like it might require some redesign if we ever need it, but for now it's just a note to keep in mind.

PR #6723, opened 2021-10-01
Use latest pygit2 from master for 3.10

skshetry:


pmrowla:
The latest pygit2 master requires libgit2 1.3.x but the version available through homebrew is only 1.1.1

skshetry:
This link says it's 1.3: https://formulae.brew.sh/formula/libgit2

But the installed version is 1.2.

pmrowla:
Ah ok, looks like it was pinned on my machine for some reason

skshetry:
I am not really sure why it's installing 1.2 instead of 1.3.0. Locally it does install 1.3.0 for me.

pmrowla:
Do we need an explicit `brew update` step before `brew install` in GHA?

skshetry:
Probably because of `HOMEBREW_NO_AUTO_UPDATE: 1` envvar that we set, and the required dependencies are not updated and therefore fails to install 1.3.0.

skshetry:
@pmrowla, you are right, it was because of an older version of `brew`.

PR #6726, opened 2021-10-02
config: introduce core.tmp_dir to allow running on nfs/cifs/etc

### LINKED ISSUE #4420: unable to open database on mapped network drive

efiop:
`core.tmp_dir` is a one-stop-shop for making dvc work on nfs, as it covers all sqlite databases and all locks. After careful consideration of https://github.com/iterative/dvc/pull/6419 , it is clear that even though `state.dir` and `index.dir` could be useful in the future for sharing state and index db, but `core.tmp_dir` solves the issue of running on nfs/cifs/etc better in one option. `core.tmp_dir` also makes `core.hardlink_lock` no longer needed, because we used to use it with lock-less sqlite connection, but these days our sqlite databases are all using locks, so that option it is not enough anyway. That being said, `core.hardlink_lock` has to stay for now for backward compatibility, but will be removed in 3.0.

Also, please note that `state/core.dir` have not yet been officially released, so we are fine removing them without worrying about backward compatibility.

Compared to `state/core.dir` that were autogenerated based on a hash of `root_dir`, `tmp_dir` is taken litterally,  to avoid making any assumptions and to allow using the same one even if you move your repository around, which made `state/core.dir` lose track of old state.

Example for making 1 dvc repo work on nfs/cifs/etc:

```
dvc config core.tmp_dir $(mktemp -d) --local
```

Reverts #6419 
Fixes #4420

### DIFF
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -114,6 +114,7 @@ SCHEMA = {
         Optional("interactive", default=False): Bool,
         Optional("analytics", default=True): Bool,
         Optional("hardlink_lock", default=False): Bool,
+        "tmp_dir": str,
         Optional("no_scm", default=False): Bool,
         Optional("autostage", default=False): Bool,
         Optional("experiments"): Bool,  # obsoleted
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -226,8 +226,12 @@ SCHEMA = {
         )
     },
     "state": {
+        "dir": str,  # noop, reserved
         "row_limit": All(Coerce(int), Range(1)),  # obsoleted
         "row_cleanup_quota": All(Coerce(int), Range(0, 100)),  # obsoleted
+    },
+    "index": {
+        "dir": str,  # noop, reserved
     },
     "machine": {
         str: {
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -225,12 +225,8 @@ SCHEMA = {
         )
     },
     "state": {
-        "dir": str,
         "row_limit": All(Coerce(int), Range(1)),  # obsoleted
         "row_cleanup_quota": All(Coerce(int), Range(0, 100)),  # obsoleted
-    },
-    "index": {
-        "dir": str,
     },
     "machine": {
         str: {
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -120,34 +120,6 @@ class Repo:
 
         return root_dir, dvc_dir, tmp_dir
 
-    def _get_database_dir(self, db_name):
-        # NOTE: by default, store SQLite-based remote indexes and state's
-        # `links` and `md5s` caches in the repository itself to avoid any
-        # possible state corruption in 'shared cache dir' scenario, but allow
-        # user to override this through config when, say, the repository is
-        # located on a mounted volume — see
-        # https://github.com/iterative/dvc/issues/4420
-        base_db_dir = self.config.get(db_name, {}).get("dir", None)
-        if not base_db_dir:
-            return self.tmp_dir
-
-        import hashlib
-
-        from dvc.utils.fs import makedirs
-
-        root_dir_hash = hashlib.sha224(
-            self.root_dir.encode("utf-8")
-        ).hexdigest()
-
-        db_dir = os.path.join(
-            base_db_dir,
-            self.DVC_DIR,
-            f"{os.path.basename(self.root_dir)}-{root_dir_hash[0:7]}",
-        )
-
-        makedirs(db_dir, exist_ok=True)
-        return db_dir
-
     def __init__(
         self,
         root_dir=None,
@@ -211,8 +183,10 @@ class Repo:
                 friendly=True,
             )
 
-            state_db_dir = self._get_database_dir("state")
-            self.state = State(self.root_dir, state_db_dir, self.dvcignore)
+            # NOTE: storing state and link_state in the repository itself to
+            # avoid any possible state corruption in 'shared cache dir'
+            # scenario.
+            self.state = State(self.root_dir, self.tmp_dir, self.dvcignore)
             self.odb = ODBManager(self)
 
             self.stage_cache = StageCache(self)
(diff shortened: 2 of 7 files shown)


efiop:
FYI ^^^ @nueschs @meierale @agurtovoy

dvc/repo/__init__.py
```diff
@@ -211,8 +184,10 @@ def __init__(
                 friendly=True,
             )
 
-            state_db_dir = self._get_database_dir("state")
-            self.state = State(self.root_dir, state_db_dir, self.dvcignore)
+            # NOTE: storing state and link_state in the repository itself to
+            # avoid any possible state corruption in 'shared cache dir'
+            # scenario.
```
efiop, line 189: This is an old note that was about not saving state along with cache, so no longer relevant
```suggestion
```

dvc/config_schema.py
```diff
@@ -114,6 +114,7 @@ class RelPath(str):
         Optional("interactive", default=False): Bool,
         Optional("analytics", default=True): Bool,
         Optional("hardlink_lock", default=False): Bool,
+        "tmp_dir": str,
```
efiop, line 117: Same problem as with state/index.dir - there is a chance that someone will try to share them between repos, which might get confusing (e.g. same lock for all repos, so they can no longer run in parallel). We could create a dir inside based on root_dir hash like state/index.dir do, but that will result in state being lost when a repo dir is moved, unless we generate the tmp_dir and then record it back somehow (either back into the config, or by creating a file with a new generated path inside, that we will then dereference). I don't like that we are trying to be too smart with these approaches, as it might bite us. The simple approach is probably better for now.

dvc/config_schema.py
```diff
@@ -225,13 +226,9 @@ class RelPath(str):
         )
     },
     "state": {
-        "dir": str,
```
efiop, line 228: Can bet my money that some users already are using it in their git history, will will result in ConfigErrors for those revisions. Let's keep these options for now, but make them defunct. In the future these will be useful for sharing state/index.

efiop:
One more approach is to make index and state sharable by default (and put all of them into a certain shared dir, e.g. in ~) once we've generalized them and then just switch to hardlink lock instead of flock-based zc.lockfile by default, so that nfs users don't even need to play with config options at all 🤔 I'm keen on this approach, but we need some time to complete that, and in the mean time `tmp_dir` would do the trick.

efiop:
Thinking about it some more, let's indeed keep this as is for now, as once state/index are generalized we'll likely not need `tmp_dir` at all, while `state.dir/cache.dir` will still be useful. Closing for now. Sorry for the noise 🙂

PR #6728, opened 2021-10-02
http: explicitly retry all http-related errors

isidentical:
Beside the server-side errors like timeouts, also retry stuff that is happening on the client (e.g internet connection drop for a second).


isidentical:
Something I noticed while trying the `dvc get` example. Not sure if there is an easy way to imitate this on the tests, so not adding something for now.

efiop (approved):
Nice catch! 🔥

skshetry:
How does this work for `put_file`? We should not be retrying on those, as it has side effects.

Issue #6707, opened 2021-09-29
regression: dvc init is slow


skshetry:
The 3 major issue seems to be:
- [x] `tpi` (see https://github.com/iterative/tpi/issues/14) (~300ms)
- [ ] `fsspec` callback (~200ms)
- [x] prompts import on experiments from `rich` (for `dvc exp init`) (~50ms) #6738
- [ ] [`dvc.fs` importing fsspec by default](https://github.com/iterative/dvc/blob/612696cf4ae01885a89269b5d26b4631380614b8/dvc/fs/azure.py#L5) (~100ms)
(There may not be much point in making fsspec lazy, as we'll need local filesystem anyway).

The following is a startup log for `dvc init`.
<img width="1107" alt="Screen Shot 2021-09-29 at 17 52 16" src="https://user-images.githubusercontent.com/18718008/135265276-7f9279b5-8610-497a-8f59-654a7a859870.png">

Following is the benchmark for `dvc init`.
```console
$ hyperfine "dvc init" --warmup 3 -p "rm -rf .dvc"
Benchmark #1: dvc init
  Time (mean ± σ):      1.311 s ±  0.065 s    [User: 1.066 s, System: 0.203 s]
  Range (min … max):    1.230 s …  1.435 s    10 runs
```

`dvc init` crossing 1 sec is really bad, this was the result a month ago (before tpi and prompts, and with fsspec.Callback made lazy):
```console
$ hyperfine "dvc init" --warmup 3 -p "rm -rf .dvc"
Benchmark #1: dvc init
  Time (mean ± σ):     506.8 ms ±  31.4 ms    [User: 369.2 ms, System: 103.7 ms]
  Range (min … max):   477.4 ms … 566.6 ms    10 runs
```

skshetry:
This is why I prefer having benchmarks in the repo (or, at least run some tests in each PR). It's difficult to figure out later.

skshetry:
We have slowly regressed on `init` side (most likely due to fsspec imports) by 25% between `2.6` and `main`.

But nothing we can do at this time. Closing...

PR #6733, opened 2021-10-03
gha: run benchmarks on PRs

efiop:
Qucik&dirty way to run dvc-bench for our PRs to help with things like #6707

Need to clarify that this is not all-in-one regression detection, but rather this is specific to test regressions within one PR. We need a similar thing for specific dvc versions to detect slow regressions that accumulate over a long period of time, but that is out of scope for this PR.

### DIFF
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -1,0 +1,19 @@
+name: benchmarks
+on: [pull_request, workflow_dispatch]
+
+jobs:
+  bench:
+    name: run benchmarks
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-18.04]
+    steps:
+      - uses: actions/setup-python@v2
+        with:
+            python-version: 3.7
+      - uses: iterative/dvc-bench@master
+        with:
+            pytest_options: "${{ github.event_name == 'workflow_dispatch' && '' || '-k test_init -k test_help' }}"
+


efiop:
Ok, posting full tables as comments is too noisy, the last version runs regression checks for individual tests, which feels much nicer to me for PRs. That being said, it is clear that we shouldn't run all possible benchmarks for PRs, but rather keep key ones (e.g. init/add/etc, but not tutorial ones), so I'll try to condense those a little bit.

.github/workflows/benchmarks.yaml
```diff
@@ -0,0 +1,48 @@
+name: benchmarks
+on: pull_request
+
+env:
+  DVC_TEST: "true"
+
+jobs:
+  build:
```
skshetry, line 8: It'd be great if we could just keep all of this in the `dvc-bench` itself, with the [composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action).

efiop: For sure! Looking into it...

efiop: @skshetry One problem here is that composite actions don't support build matrixes https://github.com/actions/runner/issues/810 , so this is going to be a bit less flexible and we will have to run all benchmarks in one single job, which is fine for now with a small selection of benchmarks, but might not be in the long term. So I would probably wait with creating a composite action for now, until we are sure that it will be enough for our use case here. 🙁

skshetry: Why do we need matrix though?

efiop: @skshetry To run benchmarks in separate jobs, as they might take quite some time once you run a bunch of them. This is not an issue with 2 benchmarks that we have in this PR, but will be once we add some more, especially ones that work with large data.

skshetry: hmm, it's better to just run small benchmarks in PRs. The jobs already take quite a lot of time, which makes me slower.

efiop: @skshetry Agreed. So looks like it is pretty clear that this is more of a micro-benchmark run for all PRs, but we might need a fuller bigger benchmark run on-demand for some PRs (maybe on a regular basis too, to check for daily/weekly/etc regressions). Will create an action for now.

skshetry: We can run a full suite on PR on demand with comment or `workflow_dispatch` like we do for binary packages builds.

.github/workflows/benchmarks.yaml
```diff
@@ -0,0 +1,19 @@
...
+      fail-fast: false
+      matrix:
+        os: [ubuntu-18.04]
+    steps:
+      - uses: actions/setup-python@v2
+        with:
+            python-version: 3.7
+      - uses: iterative/dvc-bench@master
+        with:
+            pytest_options: "${{ github.event_name == 'workflow_dispatch' && '' || '-k test_init -k test_help' }}"
```
efiop, line 18: Could create a special mark for similar tests in dvc-bench (e.g. micro), but will leave it for later.

efiop:
Ok, let's try this out.

efiop:
@skshetry Oops, looking into it... Thank you! 🙏

PR #6740, opened 2021-10-04
Fix dvclive running subdir

daavoo:
Closes https://github.com/iterative/dvclive/issues/169

`DVCLIVE_PATH` env var was not taking into account the stage's workdir causing `dvclive` to write outputs to the wrong path on some occasions (i.e. when running `dvc repro "subdir"`)

### DIFF
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -273,7 +273,7 @@ class Stage(params.StageParams):
             from dvc.output import Output
             from dvc.schema import LIVE_PROPS
 
-            env[DVCLIVE_PATH] = str(os.path.relpath(out.path_info, self.wdir))
+            env[DVCLIVE_PATH] = relpath(out.path_info, self.wdir)
             if isinstance(out.live, dict):
                 config = project(out.live, LIVE_PROPS)
 
--- a/dvc/stage/__init__.py
+++ b/dvc/stage/__init__.py
@@ -273,7 +273,7 @@ class Stage(params.StageParams):
             from dvc.output import Output
             from dvc.schema import LIVE_PROPS
 
-            env[DVCLIVE_PATH] = str(out.path_info)
+            env[DVCLIVE_PATH] = str(os.path.relpath(out.path_info, self.wdir))
             if isinstance(out.live, dict):
                 config = project(out.live, LIVE_PROPS)
 
--- a/tests/func/test_live.py
+++ b/tests/func/test_live.py
@@ -7,6 +7,7 @@ from funcy import first
 
 from dvc import stage as stage_module
 from dvc.render.utils import get_files
+from tests.utils import cd
 
 LIVE_SCRIPT = dedent(
     """
@@ -255,3 +256,26 @@ def test_dvc_generates_html_during_run(tmp_dir, dvc, mocker, live_stage):
     live_stage(summary=True, live="logs", code=script)
 
     assert show_spy.call_count == 2
+
+
+def test_dvclive_running_subir(tmp_dir, scm, dvc):
+
+    tmp_dir.gen("subdir/train.py", LIVE_SCRIPT)
+    tmp_dir.gen("subdir/params.yaml", "foo: 1")
+    with cd("subdir"):
+        dvc.stage.add(
+            cmd="python train.py",
+            params=["foo"],
+            deps=["train.py"],
+            name="live_stage",
+            live="dvclive",
+            live_no_html=True,
+        )
+
+    scm.add(["subdir/dvc.yaml", "subdir/train.py", "subdir/params.yaml"])
+    scm.commit("initial: live_stage")
+
+    dvc.reproduce("subdir", recursive=True)
+
+    assert (tmp_dir / "subdir" / "dvclive").is_dir()
+    assert (tmp_dir / "subdir" / "dvclive.json").is_file()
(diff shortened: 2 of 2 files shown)


pmrowla (approved):

tests/func/test_live.py
```diff
@@ -255,3 +256,26 @@ def test_dvc_generates_html_during_run(tmp_dir, dvc, mocker, live_stage):
...
+            cmd="python train.py",
+            params=["foo"],
+            deps=["train.py"],
+            name="live_stage",
+            live="dvclive",
+            live_no_html=True,
+        )
+
+    scm.add(["subdir/dvc.yaml", "subdir/train.py", "subdir/params.yaml"])
+    scm.commit("initial: live_stage")
```
skshetry, line 276: Do we need this?

tests/func/test_live.py
```diff
@@ -255,3 +256,26 @@ def test_dvc_generates_html_during_run(tmp_dir, dvc, mocker, live_stage):
...
+    tmp_dir.gen("subdir/params.yaml", "foo: 1")
+    with cd("subdir"):
+        dvc.stage.add(
+            cmd="python train.py",
+            params=["foo"],
+            deps=["train.py"],
+            name="live_stage",
+            live="dvclive",
+            live_no_html=True,
+        )
```
skshetry, line 273: You may just be able to set `wdir=subdir` instead of using `cd`. That'll create a `dvc.yaml` file instead of `subdir/dvc.yaml`. Also note that `subdir` should exist before `dvc.stage.add`.

tests/func/test_live.py
```diff
@@ -255,3 +256,26 @@ def test_dvc_generates_html_during_run(tmp_dir, dvc, mocker, live_stage):
     live_stage(summary=True, live="logs", code=script)
 
     assert show_spy.call_count == 2
+
+
+def test_dvclive_running_subir(tmp_dir, scm, dvc):
+
+    tmp_dir.gen("subdir/train.py", LIVE_SCRIPT)
```
skshetry, line 263: Instead of using stringy paths:
```suggestion
    (tmp_dir / "subdir").gen("train.py", LIVE_SCRIPT)
```

or,
```suggestion
    tmp_dir.gen({"subdir": {"train.py": LIVE_SCRIPT}})
```

tests/func/test_live.py
```diff
@@ -255,3 +256,26 @@ def test_dvc_generates_html_during_run(tmp_dir, dvc, mocker, live_stage):
     live_stage(summary=True, live="logs", code=script)
 
     assert show_spy.call_count == 2
+
+
+def test_dvclive_running_subir(tmp_dir, scm, dvc):
```
skshetry, line 261:
```suggestion
def test_dvclive_running_subdir(tmp_dir, scm, dvc):
```

skshetry: or,
```suggestion
def test_dvclive_stage_with_different_wdir(tmp_dir, scm, dvc):
```

dvc/stage/__init__.py
```diff
@@ -273,7 +273,7 @@ def _read_env(self, out, checkpoint_func=None) -> Env:
             from dvc.output import Output
             from dvc.schema import LIVE_PROPS
 
-            env[DVCLIVE_PATH] = str(out.path_info)
+            env[DVCLIVE_PATH] = str(os.path.relpath(out.path_info, self.wdir))
```
skshetry, line 276: We usually use our own [`dvc.utils.relpath`](https://github.com/iterative/dvc/blob/0eda4bd91ceac9c5bfba30f52e7951b86ee7bfe9/dvc/utils/__init__.py#L314) which is already imported here.
```suggestion
            env[DVCLIVE_PATH] = str(relpath(out.path_info, self.wdir))
```

pared (approved):

tests/func/test_live.py
daavoo: I originally writed the test to reproduce the exact workflow the user reported in the bug. But you are right, using `wdir` is enough to cover the case, thanks!

tests/func/test_live.py
```diff
@@ -255,3 +255,23 @@ def test_dvc_generates_html_during_run(tmp_dir, dvc, mocker, live_stage):
     live_stage(summary=True, live="logs", code=script)
 
     assert show_spy.call_count == 2
+
+
+def test_dvclive_stage_with_different_wdir(tmp_dir, scm, dvc):
+
```
skshetry, line 261:
```suggestion
```

tests/func/test_live.py
```diff
@@ -255,3 +255,23 @@ def test_dvc_generates_html_during_run(tmp_dir, dvc, mocker, live_stage):
...
+    (tmp_dir / "subdir").gen("train.py", LIVE_SCRIPT)
```
skshetry, line 262: I think we should keep the path in envvar in an absolute form. The user's script could change the working directory.
```suggestion
    (tmp_dir / "subdir2").mkdir()
    (tmp_dir / "subdir").gen("train.py", "import os\nos.chdir('..')\n" + LIVE_SCRIPT)
```

skshetry (approved):
LGTM

tests/func/test_live.py
skshetry: @pared @daavoo, what do you guys think? Should I create an issue for this?

daavoo: https://github.com/iterative/dvclive/issues/171

PR #6750, opened 2021-10-06
exp init: iterate on ui, tree and show to-be-generated yaml content

skshetry:
It has some of the ideas that we have so far discussed. It's a bit ugly with lots of color palletes, structure mixed together, but should be the gist of it. This reminds me that we should standardize use of some colors (using accessible colors, colorblind friendly, etc.).

cc @dberenbaum


shcheklein:
@skshetry could you do a recording please? :)

dberenbaum:
Since this is hard to see, edit, or make suggestions for rapid iteration in GH, I made a google doc: https://docs.google.com/document/d/1XLe6XCWWNX_urXRyS2nUmhGiUwzPRAEjkl7KNoQsu8Y/edit?usp=sharing

dberenbaum:
Can we validate that dependencies exist at the time of entry during interactive mode? It would save a lot of trouble with having to redo the whole stage after it fails at the end.

PR #6752, opened 2021-10-06
plots: allow multiple outputs for --show-vega

pared:

### DIFF
--- a/dvc/command/live.py
+++ b/dvc/command/live.py
@@ -3,6 +3,7 @@ from pathlib import Path
 
 from dvc.command import completion
 from dvc.command.base import CmdBase, fix_subparsers
+from dvc.render.utils import match_renderers
 from dvc.ui import ui
 
 
@@ -16,7 +17,8 @@ class CmdLive(CmdBase):
             html_path = Path.cwd() / (self.args.target + "_html")
             from dvc.render.utils import render
 
-            index_path = render(self.repo, plots, metrics, html_path)
+            renderers = match_renderers(plots, self.repo.plots.templates)
+            index_path = render(self.repo, renderers, metrics, html_path)
             ui.write(index_path.as_uri())
             return 0
         return 1
--- a/dvc/command/plots.py
+++ b/dvc/command/plots.py
@@ -1,10 +1,12 @@
 import argparse
+import json
 import logging
 
 from dvc.command import completion
 from dvc.command.base import CmdBase, append_doc_link, fix_subparsers
 from dvc.exceptions import DvcException
-from dvc.render.utils import find_vega, render
+from dvc.render.utils import match_renderers, render
+from dvc.render.vega import VegaRenderer
 from dvc.ui import ui
 from dvc.utils import format_link
 
@@ -25,16 +27,6 @@ class CmdPlots(CmdBase):
     def run(self):
         from pathlib import Path
 
-        if self.args.show_vega:
-            if not self.args.targets:
-                logger.error("please specify a target for `--show-vega`")
-                return 1
-            if len(self.args.targets) > 1:
-                logger.error(
-                    "you can only specify one target for `--show-vega`"
-                )
-                return 1
-
         try:
             plots_data = self._func(
                 targets=self.args.targets, props=self._props()
@@ -46,9 +38,21 @@ class CmdPlots(CmdBase):
                     "visualization file will not be created."
                 )
 
+            renderers = match_renderers(
+                plots_data=plots_data, templates=self.repo.plots.templates
+            )
+
             if self.args.show_vega:
-                target = self.args.targets[0]
-                plot_json = find_vega(self.repo, plots_data, target)
+                vega_only = filter(
+                    lambda r: isinstance(r, VegaRenderer), renderers
+                )
+                plot_json = json.dumps(
+                    {
+                        renderer.filename: json.loads(renderer.as_json())
+                        for renderer in vega_only
+                    },
+                    indent=4,
+                )
                 if plot_json:
                     ui.write(plot_json)
                 return 0
@@ -57,7 +61,7 @@ class CmdPlots(CmdBase):
             path: Path = (Path.cwd() / rel).resolve()
             index_path = render(
                 self.repo,
-                plots_data,
+                renderers,
                 path=path,
                 html_template_path=self.args.html_template,
             )
--- a/dvc/render/utils.py
+++ b/dvc/render/utils.py
@@ -21,15 +21,6 @@ def group_by_filename(plots_data: Dict) -> List[Dict]:
     return grouped
 
 
-def find_vega(repo, plots_data, target):
-    found = dpath.util.search(plots_data, ["*", "*", target])
-    from dvc.render.vega import VegaRenderer
-
-    if found and VegaRenderer.matches(found):
-        return VegaRenderer(found, repo.plots.templates).as_json()
-    return ""
-
-
 def match_renderers(plots_data, templates):
     from dvc.render import RENDERERS
 
@@ -43,14 +34,12 @@ def match_renderers(plots_data, templates):
 
 def render(
     repo,
-    plots_data,
+    renderers,
     metrics=None,
     path=None,
     html_template_path=None,
     refresh_seconds=None,
 ):
-    # TODO we could probably remove repo usages (here and in VegaRenderer)
-    renderers = match_renderers(plots_data, repo.plots.templates)
     if not html_template_path:
         html_template_path = repo.config.get("plots", {}).get(
             "html_template", None
--- a/dvc/repo/live.py
+++ b/dvc/repo/live.py
@@ -6,7 +6,7 @@ from typing import TYPE_CHECKING, List, Optional
 
 from funcy import once_per_args
 
-from dvc.render.utils import render
+from dvc.render.utils import match_renderers, render
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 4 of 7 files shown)


pared:
Well it seems there is no necessity for mapping now, but I think it would be safer to have one, it gives a lot of context what particular vega contains, so probably sooner or later we will need that.

dberenbaum:
We will also start working on the experiments API in the future. We should think about what plots info will be useful to return in an API, which should probably resemble the outputs here.

pared:
@Mergifyio rebase

pared:
@rogermparent lets get back to this issue:

Initially, this PR was about allowing multiple targets for `--show-vega`, now after starting supporting images the scope starts to shift, and I started to think how to implement that, but after discussing a bit with @dberenbaum and @daavoo I start to wonder whether I am not too hastily getting into that.

How the results are going to be visualized in the extension? 
Wouldn't it make sense if both `dvc plots show` and extension results be similar? 
It seems to me that if we split the way we visualize our results, soon we will end up with different implementations. 
Do you think it would make sense, to, for example, embed HTML in the vscode? Would it be possible at all?

rogermparent:
> Initially, this PR was about allowing multiple targets for `--show-vega`, now after starting supporting images the scope starts to shift, and I started to think how to implement that, but after discussing a bit with @dberenbaum and @daavoo I start to wonder whether I am not too hastily getting into that.

It's possible, but I'd say that's more of a question about the DVC codebase. VSCode will need both features, and preferably in one command, but we can absolutely make do with vega plots first and static plots in a later PR.

> How the results are going to be visualized in the extension? Wouldn't it make sense if both `dvc plots show` and extension results be similar? It seems to me that if we split the way we visualize our results, soon we will end up with different implementations. Do you think it would make sense, to, for example, embed HTML in the vscode? Would it be possible at all?

We currently have a design that's reminiscent of the `dvc plots show` HTML, but a little more ambitious with features like collapsing and search.

static plots design screenshot

While similar, the differences in this design mean we're already intending to make a different implementation than `dvc plots show`'s output which serves a different, simpler purpose. 

As far as using HTML goes, this vega plots UI is also slated to be part of a larger "visualizations" webview shared with live plots, and also VSCode requires webviews use [URLs with a special `resource:` protocol to load local data](https://code.visualstudio.com/api/extension-guides/webview#loading-local-content) so using the HTML straight is out of the question.

Handling all of these considerations with `plots show` HTML would essentially require us to heavily pre-parse the given HTML and use the gathered data similarly to how we would use the JSON from `--show-vega`. While this is doable, we'd essentially be consuming the `plots show` HTML output as an API which would add friction in any changes to that HTML as we worry they could become breaking.

pared:
@rogermparent makes sense, thanks for the clarifications. Seems like we will need to pack it all into `--show-json`. Need to think about how to do that. Writing files and providing URLs which we discussed, sounds a bit clunky, maybe we will come up with something better.

dberenbaum:
Thanks @rogermparent and @pared!

Let me try to recap some of my questions and concerns from our discussion for @rogermparent (and probably useful for @shcheklein).

It seems like we are redoing the Python API as a JSON output. Do we want to do this? If so, it might clarify things and we can at least have consistency. On the other hand, it might be redundant and doesn't necessarily have a lot of general use unless others can't use Python.

> Handling all of these considerations with `plots show` HTML would essentially require us to heavily pre-parse the given HTML and use the gathered data similarly to how we would use the JSON from `--show-vega`. While this is doable, we'd essentially be consuming the `plots show` HTML output as an API which would add friction in any changes to that HTML as we worry they could become breaking.

This sounds ugly, but I'm not yet convinced it's much worse than the the current approach, and I'd rather start as simple as possible. My reasoning:
* The HTML already includes both the Vega and static plots. VS Code already has what they need, and no special output needs to be developed in DVC.
* VS Code can specify the output location and have all the data they need (the HTML and the static plot files). They can manage temporary directories and cleanup however they want.
* The HTML template is [configurable](https://dvc.org/doc/command-reference/plots#html-templates), so VS Code could use its own and have less risk of breaking changes.

shcheklein:
> This sounds ugly

Yep :)

It sounds a bit too risky to use HTML (and maintain its backward compatibility). We'll have to go away from this for sure and it doesn't sound worth spending time on implementing HTML parser.

What was wrong with initial proposal - expand `--show-vega` to return multiple targets.

For images - I would add a separate API that generates a directory with a predefined structure.

shcheklein:
> The HTML template is configurable, so VS Code could use its own and have less risk of breaking changes.

It doesn't help much unless `{plot_divs}` is extremely simple and we are 100% we won't be changing it in the future.

Again, just not sure what was so wrong with the initial proposed `--show-vega` with multiple targets and generating dir with images also doesn't sound that complicated.

pared:
@shcheklein it is not about complexity, we started questioning the necessity of the existence of duplicated API (JSON one) when we already have one (unofficial and pythonic, but still). Anyway it seems that HTML will be too complicated to handle.

> For images - I would add a separate API that generates a directory with a predefined structure.

How would that look like? New option? `--show-images`?

shcheklein:
> when we already have one

Yep, I get it. But we can't consume it now unless I'm missing something.

> How would that look like? New option? --show-images?

something like this, yep

dberenbaum:
Sorry for the delay. I wanted to think about it a bit. Overall, I'm still wondering whether it makes sense to implement new functionality right now rather than go as far as possible with existing functionality first.

> It seems like we are redoing the Python API as a JSON output. Do we want to do this? If so, it might clarify things and we can at least have consistency. On the other hand, it might be redundant and doesn't necessarily have a lot of general use unless others can't use Python.

@shcheklein What do you think about this? Do you see the current direction as a long-term solution? If you feel confident that it has long-term benefits, then sure, let's spend the time to get them right. I want to make sure we aren't prematurely committing to a more time-consuming short-term solution.

> Again, just not sure what was so wrong with the initial proposed `--show-vega` with multiple targets and generating dir with images also doesn't sound that complicated.

@pared mentioned that it seems like we are duplicating an existing API, which I worry has little value outside of this use case and could extend to other features and commands. Even if each request is not that complicated, it creates an inefficient workflow. The VS Code team has to figure out requirements ahead of time, relay them to the DVC team, wait for the DVC team to implement them, validate that it meets their requirements, then possibly iterate to resolve discrepancies.

rogermparent:
> it seems like we are duplicating an existing API, which I worry has little value outside of this use case and could extend to other features and commands.

While the CLI-as-API technique is ugly and comes with its own problems, we're certainly going to need some way to get these APIs into a non-Python format to encourage users to make DVC integrations that aren't locked to Python, and JSON is a pretty viable format for this considering JS is continuously engulfing all of software development and all other languages have some kind of JSON parser regardless.

Is there anywhere the Python plots API is documented? I don't see it in the dvc.org docs. I remember it being described in the last cross-team meeting as undocumented but internally used.

The VSCode team could use docs for the Python API to either create a small embedded Python script that translates it for us or use a fully-fledged Python->Node bridge like [boa](https://github.com/imgcook/boa) which @mattseddon has been experimenting with to minor success. This could be a good middle-ground where the duplicated API doesn't have to be in dvc itself, and the VSCode team can get it in quickly without putting possibly shoddy code in DVC itself, then we can work on it for a possible extraction and future public release for others in the "I want to use DVC API in Node" use case we find ourselves in.

Happy to continue discussion either here, in the cross-team meeting today, or both.

pared:
> How would that look like? New option? --show-images?

@shcheklein I am not a fan of such solution, because in that case we start to separate the plots by their type. To me, it seems that more viable option would be just a generic `dvc plots show --show-json` where the json contains information depending on plot type. Eg:
`{"type": "vega", "data": {...}}` would mean that this is a vega plot and `data` is (for example) vega dict. `{"type": "image", "URL": {...}}` would mean binary data type that user needs to access via URL.

shcheklein:
> would mean binary data type that user needs to access via URL.

@pared is it possible at all? Or do you want to generate the dir along the way?

shcheklein:
> On the other hand, it might be redundant and doesn't necessarily have a lot of general use unless others can't use Python.

@dberenbaum to @rogermparent point I think `--json` is a very regular thing to have. Since it usually builds on top of some regular internal Python API it's not a problem in terms of redundancy. It's just a matter of what downstream tools could consume. In our case, unless I'm missing some options, we don't have a way to consume Python, so one way or another we'll need a non-Python interface to that API. 

> @pared mentioned that it seems like we are duplicating an existing API

It should be done in a way that it's not duplicating an API, but rather provides an alternative interface to that API. (regular CLI itself is another alternative interface that ideally should be built on top of API).

pared:
> It should be done in a way that it's not duplicating an API, but rather provides an alternative interface to that API. (regular CLI itself is another alternative interface that ideally should be built on top of API) 

If we want to have an alternative UI to CLI then probably we would need to handle the images in a similar way as we do in a normal CLI call: we would need a dir to store the results of the visualization. That is also an option, though it will probably clutter the repo. But if default would be `.dvc/tmp` that might just do the trick.

rogermparent:
> But if default would be `.dvc/tmp` that might just do the trick.

VSCode can can access `.dvc/tmp` pretty easily regardless of the project (unlike `.git`), so I think if that's a solution we can definitely run with.

pared:
I have rebased this PR. We can merge it and I start working on more generic solution right away.

pared:
After some discussion and consideration, it seems that we will introduce new generic option that will handle `image` type plots too. It seems that there is no need to merge this change, as it would break previously existing `--show-vega`. Closing for now, we can reconsider that if we see users demand for it.

PR #6773, opened 2021-10-08
[WIP] use setuptools_scm and build_requires

skshetry:
Let's not include this in the next release as there are already too many changes, and we may need to test for binaries and other packages.

The local version is not the same that we had before, it's set to whatever default setuptools_scm uses, I didn't like to mess with it, as it makes things more complicated.

Please feel free to take over while I'm away.


skshetry:
Opened #6775 to build packages.

Issue #5515, opened 2021-02-24
params/metrics/exp diff: consistent output


jorgeorpinel:
> Extracted from https://github.com/iterative/dvc.org/pull/2200#discussion_r581757584

* `metrics diff` currently shows old, new, change
* `params diff` currently shows old, new
* `exp diff` currently shows new, change

Should they all be Old, New, Change? (Possibly with some flag to skip Old or Change)

jorgeorpinel:
cc @dberenbaum

dberenbaum:
Thanks, @jorgeorpinel! Let's refer this back to #5392. 

> Should they all be Old, New, Change?

👍

jorgeorpinel:
What do you mean? 🙂

jorgeorpinel:
BTW @pmrowla do you know if we've considered merging `dvc params/metrics/exp diff` into a single command (e.g. `dvc compare`)? Or maybe flags of `dvc diff`

pmrowla:
@jorgeorpinel the idea has come up before as something that we could consider doing in the future, but not seriously considered yet

dberenbaum:
> What do you mean @dberenbaum? 🙂

I just wanted to mention it so we have all of these UI issues linked in that one issue. No other action needed!

jorgeorpinel:
Git it 👍

daavoo:
Was about to comment this in #5392 (haven't seen any other related issues). I think that the terms `Old` / `New` used in `diff` sometimes could be confusing, given that users can pass `refs` in any order. 

For example they could run `dvc metrics diff {newer commit hash} {older commit hash}` and the table will show the values of the newer commit under the column `Old`.

dberenbaum:
cc @skshetry 

I'd prefer having the ref names instead of Old and New.

daavoo:
Two questions:

A) Should `Change` be included by default ?? 

I might be biased but in my personal experience, but the "Change" value is kind of awkward mostly (sometimes overflowing the float diff, others mixing diffs that don't make sense).

I would vote for hiding it by default and making it optional.

B) Should there be an option to hide `a_rev` (previously "Old") at all?

IMO the only case where it could make sense to not show `a_rev` is when calling `dvc {X} diff {b_rev}` (where `a_rev` is set to `HEAD` by default.)

Maybe instead of a flag for opting out `a_rev`, we could just hide it when equals to `HEAD`.

dberenbaum:
[...]

PR #6814, opened 2021-10-15
exp/metrics/params diff: Replaced `Old/New` with `a_rev/b_rev`

daavoo:

`dvc.org` P.R: https://github.com/iterative/dvc.org/pull/2930
Per https://github.com/iterative/dvc/issues/5515#issuecomment-883929040

---

- With targets.

Before:

```bash
dvc params diff cnn low-lr-experiment 
Path         Param     Old    New                                     
params.yaml  train.lr  0.01  0.001
```

After:

```bash
dvc params diff
Path         Param     cnn    low-lr-experiment                       
params.yaml  train.lr  0.01   0.001
```

---

- Default.

Before:

```bash
dvc params diff
Path         Param     Old    New                                     
params.yaml  train.lr  0.001  0.009
```

After:

```bash
dvc params diff
Path         Param     HEAD    workspace                              
params.yaml  train.lr  0.001   0.009
```

### DIFF
--- a/tests/func/metrics/test_diff.py
+++ b/tests/func/metrics/test_diff.py
@@ -187,9 +187,10 @@ def test_metrics_diff_cli(tmp_dir, scm, dvc, run_copy_metrics, caplog, capsys):
     assert main(["metrics", "diff", "HEAD~2"]) == 0
 
     captured = capsys.readouterr()
+
     assert captured.out == (
-        "Path    Metric    Old      New      Change\n"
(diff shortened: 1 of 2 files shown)


skshetry:
For 3.0, we should also change the internal API return type to use "a_rev", "b_rev", instead of "old"/"new". I don't like that this adds 2 parameters to all of the APIs for that, though we may need to wait till 3.0.

PR #6815, opened 2021-10-15
 LiveTask: write output path at start/end 

daavoo:

Closes https://github.com/iterative/dvclive/issues/177

Logging at the end of each step can mess up with some ML Framework progress bars, I moved the message to the instantiation of the monitor task, as the path won't change during training.

@pared Do you consider necessary logging at each `create_summary` call? If so, we could maybe keep the call but at `debug` level

### DIFF
--- a/dvc/repo/live.py
+++ b/dvc/repo/live.py
@@ -1,5 +1,4 @@
 import logging
-import os
 from typing import TYPE_CHECKING, List, Optional
 
 from dvc.render.utils import render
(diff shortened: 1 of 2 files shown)


pared:
@dberenbaum there is none, the main point was that ideally, we would only get a single path log, but usually, training produces some other logs, hiding static site URL somewhere in the beginning. Producing it every step makes it even more ugly, so I think let's log it once.

dberenbaum:
> training produces some other logs, hiding static site URL somewhere in the beginning

Good point, this is what I was missing. Maybe it would be nice to print once more at the end?

> Producing it every step makes it even more ugly, so I think let's log it once.

Sounds good for now.

PR #6827, opened 2021-10-18
Revert "webdav: limit httpx to fix the CI"

skshetry:
Reverts iterative/dvc#6800. Also bumps the minimum requirement for webdav4 to 0.9.3, which pins the httpx version.


efiop:
> I think we should still have options for "Rebase and Merge", which makes the most sense here.

For consistency squashing for now, but happy to discuss, seems like we might indeed be past the phase of squash-only.

Issue #6819, opened 2021-10-18
exp run: Parameter name does not support backslashes


DavidGOrtega:
# Bug Report

## Description
In  GH CI is useful to reuse the ENV var ```GITHUB_REPOSITORY``` if not the workflow is even uglier in bash.
A plausible solution would be to use a hash function internally to generate an "ID" from the name or throw an error with illegal chars.
If that would be accepted I can prepare a PR 🚀 

### Reproduce

In any experiment try to setup a name with backslashes 

```sh
dvc exp run --name wont/work
```

### Expected

To work or at least a log related to wrong naming.

### Environment information

```console
DVC version: 2.7.4 (deb)
---------------------------------
Platform: Python 3.8.11 on Linux-5.8.0-1042-azure-x86_64-with-glibc2.14
Supports:
	azure (adlfs = 2021.9.1, knack = 0.8.2, azure-identity = 1.6.1),
	gdrive (pydrive2 = 1.9.4),
	gs (gcsfs = 2021.8.1),
	hdfs (pyarrow = 5.0.0),
	webhdfs (hdfs = 2.6.0),
	http (aiohttp = 3.7.4.post0, aiohttp-retry = 2.4.5),
	https (aiohttp = 3.7.4.post0, aiohttp-retry = 2.4.5),
	s3 (s3fs = 2021.8.1, boto3 = 1.17.106),
	ssh (sshfs = 2021.8.1),
	oss (ossfs = 2021.8.0),
	webdav (webdav4 = 0.9.2),
	webdavs (webdav4 = 0.9.2)
Cache types: <https://error.dvc.org/no-dvc-cache>
Caches: local
Remotes: s3
Workspace directory: ext4 on /dev/root
Repo: dvc, git
```

Output logs:
```sh
2021-10-18 08:57:34,587 ERROR: Failed to reproduce experiment '42a05ee': expected str, bytes or os.PathLike object, not NoneType
------------------------------------------------------------
Traceback (most recent call last):
  File "dvc/repo/experiments/__init__.py", line 807, in _workspace_repro
  File "dvc/scm/git/__init__.py", line 343, in _backend_func
  File "dvc/scm/git/backend/dulwich.py", line 309, in set_ref
  File "os.py", line 806, in fsencode
TypeError: expected str, bytes or os.PathLike object, not NoneType

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "dvc/main.py", line 55, in main
[...]

PR #6848, opened 2021-10-21
exp run: add experiment name check.

### LINKED ISSUE #6652: exp run: Results cannot be found if experiment name includes space
# Bug Report

## Description 
When running an experiment with white space in the name, the resulting experiment does not show up in `exp show` later.

### Reproduce

1. `dvc exp run --name "some name": Experiments runs as expected.
2. `dvc exp show`: Experiment is not listed.

### Expected

I would expect that either

- `dvc exp run` should refuse to assign names with white space in them,
- or `dvc exp show` shows this experiment in the summary regardless of the white space in the name.

### Environment information

**Output of `dvc doctor`:**

```console
DVC version: 2.7.2 (pip)
Platform: Python 3.7.10 on Linux-5.14.5-arch1-1-x86_64-with-arch
```

### LINKED ISSUE #6674: exp run -n: check for conflicts before running experiment
`dvc exp run -n` will throw an error if a naming conflict arises, but it won't check until after the experiment has completed. 

I guess someone could re-run the experiment later with a new name and the run-cache should prevent the whole experiment from needing to be executed, but this doesn't seem intuitive, and it would be better to know immediately that the experiment name conflicts with an existing experiment.

Steps to reproduce:

```
$ dvc init
$ echo "p: 0" > params.yaml
$ dvc stage add -p p -n test "echo test"
$ dvc exp run -n test
Running stage 'test':
> echo test
test
Updating lock file 'dvc.lock'

        git add dvc.yaml dvc.lock params.yaml

Reproduced experiment(s): test

        dvc exp branch <exp> <branch>
$ dvc exp run -n test -S p=1
Running stage 'test':
> echo test
test
Updating lock file 'dvc.lock'

        git add params.yaml dvc.yaml dvc.lock
ERROR: Reproduced experiment conflicts with existing experiment 'refs/exps/d4/5e189b914d70882bd67b041110f3f27a30f941/test'. To overwrite the existing experiment run:

        dvc exp run -f ...

To run this experiment with a different name run:

        dvc exp run -n <new_name> ...
```

karajan1001:
fix #6674 and fix #6652 

1. Add experiment name check (https://git-scm.com/docs/git-check-ref-format)
2. Add duplicate exp name check.
3. Add some unit test for it.

@pmrowla , in #6819,  exp name with '/' inside are request. And these types of exp names are stored in a subdirectory.  For example in #6843, it is stored in `.git/refs/heads/dependabot/pip/pylint-2.11.1`. The subdirectory here is used for grouping branches. Currently, our `dvc exp` did not support this. It involves not only name checking here, maybe we should reopen #6819 and create a new PR for it?

### DIFF
--- a/dvc/repo/experiments/__init__.py
+++ b/dvc/repo/experiments/__init__.py
@@ -473,20 +473,16 @@ class Experiments:
             "\tdvc exp branch <exp> <branch>\n"
         )
 
-    def _validate_ref_name(self, name: Optional[str], **kwargs):
+    def _validate_new_ref(self, exp_ref: ExpRefInfo):
         from .utils import check_ref_format
 
-        if name is None:
+        if not exp_ref.name:
             return
 
-        baseline_sha = kwargs.get("baseline_rev") or self.repo.scm.get_rev()
-        exp_ref = ExpRefInfo(baseline_sha=baseline_sha, name=name)
         check_ref_format(self.scm, exp_ref)
 
-        reset = kwargs.get("reset", False)
-        force = kwargs.get("force", False)
-        if not (force or reset) and self.scm.get_ref(str(exp_ref)):
-            raise ExperimentExistsError(name)
+        if self.scm.get_ref(str(exp_ref)):
+            raise ExperimentExistsError(exp_ref.name)
 
     @scm_locked
     def new(self, *args, checkpoint_resume: Optional[str] = None, **kwargs):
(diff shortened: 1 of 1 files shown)


pmrowla:
> @pmrowla , in #6819, exp name with '/' inside are request. And these types of exp names are stored in a subdirectory. For example in #6843, it is stored in `.git/refs/heads/dependabot/pip/pylint-2.11.1`. The subdirectory here is used for grouping branches. Currently, our `dvc exp` did not support this. It involves not only name checking here, maybe we should reopen #6819 and create a new PR for it?

We should not support slashes in experiment names. Yes, the full ref can contain slashes, and we already use slashes in exp refs to organize them, but the user-readable (short) name should not contain slashes - the same as git branch (short) names.

pmrowla:
Ah ok, that's my misunderstanding then, I didn't think cgit allowed you to nest anything in `refs/heads`. I guess that could be re-opened as a feature request, but I'm still not sure we should support it, unless it becomes something that is more frequently requested from users. I'd prefer to just keep everything in `refs/exps/...` namespaces completely reserved for DVC use for now.

karajan1001:
> > I'd prefer to just keep everything in `refs/exps/...` namespaces completely reserved for DVC use for now.
> 
> Is it harder to handle nested refs in `refs/exps`? Or is there some other reason not to allow slashes? @karajan1001 makes a good point that this could be a helpful way to organize, plus it would solve #6819.

The `dulwich` backend support `/` in exp name, and actually our full reference is inside the subdirectory named by the `HEAD` revision. But still some problems on the DVC side, for example currently we use `/` to separate refname and the last part as the human-readable one, this should be changed. I don't think it is very difficult to support `/` in ref name, but still need some effort.

jorgeorpinel:
From what I understand this fixes bugs so no docs changes are needed (I checked the box).

karajan1001:
> > From what I understand this fixes bugs so no docs changes are needed (I checked the box).

This PR forbids the duplicate name and name with some illegal character (space, asterisk, etc), in my opinion, better to make an error msg if someone wants to use an illegal name.

jorgeorpinel:
OK I created https://github.com/iterative/dvc.org/pull/3065 then.

PR #6850, opened 2021-10-22
[WIP] api: traversable `files` API

skshetry:
WIP, I don't plan to merge this without more discussions/refinements. The implementation is quite straightforward (half of the changes are type-annotations). See: https://github.com/iterative/dvc/pull/6550#discussion_r728600334 for the motivation.

This API allows users to traverse through Repo, with `pathlib.Path`-like API.

eg:
```python
from dvc.api import files

with files() as root:
   (root / "data" / "data.xml").read_text()
   print((root / "data" / "data.xml").url(remote=None))
   print(list((root / "data").iterdir()))
```

Pathlib APIs supported: `.open`/`.read_text`/`.read_bytes`/`.iterdir`/`.exists`/`.is_dir`/`.isfile`.
Extended APIs: `.read` (equivalent to `dvc.api.read`) / `.url` (eq. to `dvc.api.get_url`).
Limited support for: `.glob`/`.rglob`

Also, all of the `dvc.api` have been migrated to use `files()` api.

---

Given that we are migrating to fsspec, and the fact that we have `RepoFileSystem`, this does mean that we'll have two similar APIs (though this one is built on top of it). The `RepoFileSystem` may be more powerful, but the pathlib-like API will definitely be user-friendly. We can keep this API private or leave this PR unmerged till we decide on the API.

### DIFF
--- a/dvc/api.py
+++ b/dvc/api.py
@@ -1,10 +1,22 @@
 import os
 from contextlib import _GeneratorContextManager as GCM
+from contextlib import contextmanager
+from typing import ContextManager, Iterator
 
-from funcy import reraise
-
-from dvc.exceptions import OutputNotFoundError, PathMissingError
+from dvc.exceptions import NoOutputInExternalRepoError, OutputNotFoundError
 from dvc.repo import Repo
+from dvc.repo_path import RepoPath
+
+
+def files(path=os.curdir, repo=None, rev=None) -> ContextManager[RepoPath]:
+    @contextmanager
+    def inner() -> Iterator["RepoPath"]:
+        with Repo.open(
+            repo, rev=rev, subrepos=True, uninitialized=True
+        ) as root_repo:
+            yield RepoPath(path, fs=root_repo.repo_fs)
+
+    return inner()
 
 
 def get_url(path, repo=None, rev=None, remote=None):
@@ -18,17 +30,11 @@ def get_url(path, repo=None, rev=None, remote=None):
     NOTE: This function does not check for the actual existence of the file or
     directory in the remote storage.
     """
-    with Repo.open(repo, rev=rev, subrepos=True, uninitialized=True) as _repo:
-        fs_path = _repo.fs.path.join(_repo.root_dir, path)
-        with reraise(FileNotFoundError, PathMissingError(path, repo)):
-            metadata = _repo.repo_fs.metadata(fs_path)
-
-        if not metadata.is_dvc:
-            raise OutputNotFoundError(path, repo)
-
-        cloud = metadata.repo.cloud
-        md5 = metadata.repo.dvcfs.info(fs_path)["md5"]
-        return cloud.get_url_for(remote, checksum=md5)
+    try:
+        with files(path, repo=repo, rev=rev) as path_obj:
+            return path_obj.url(remote=remote)
+    except NoOutputInExternalRepoError as exc:
+        raise OutputNotFoundError(exc.path, repo=repo)
 
 
 def open(  # noqa, pylint: disable=redefined-builtin
@@ -46,15 +52,15 @@ def open(  # noqa, pylint: disable=redefined-builtin
                 ) as fd:
             # ... Handle file object fd
     """
-    args = (path,)
-    kwargs = {
-        "repo": repo,
-        "remote": remote,
-        "rev": rev,
-        "mode": mode,
-        "encoding": encoding,
-    }
-    return _OpenContextManager(_open, args, kwargs)
+
+    def _open():
+        with files(path, repo=repo, rev=rev) as path_obj:
+            with path_obj.open(  # pylint: disable=not-context-manager
+                remote=remote, mode=mode, encoding=encoding
+            ) as fd:
+                yield fd
+
+    return _OpenContextManager(_open, (), {})
 
 
 class _OpenContextManager(GCM):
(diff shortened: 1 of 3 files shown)


dvc/api.py
```diff
@@ -1,11 +1,17 @@
...
-
-from dvc.exceptions import OutputNotFoundError, PathMissingError
-from dvc.path_info import PathInfo
+from dvc.exceptions import NoOutputInExternalRepoError, OutputNotFoundError
 from dvc.repo import Repo
+from dvc.repo_path import RepoPath
+
+
+@contextmanager
+def files(path=os.curdir, repo=None, rev=None) -> Iterator[RepoPath]:
```
daavoo, line 12: I don't have a better suggestion but IMO, the name `files` doesn't feel very intuitive (or related with what it returns).

pared: @daavoo `walk_files`?

daavoo: `files()`/`walk_files` sound to me like what I would expect to be returned by `os.listdir("root")`/`os.walk("root")`. 

However, it looks like what `files()` is returning is more similar to what `Path("root")` would return (unless I'm completely misunderstanding the example snippets in the description).

So it would make more sense a name that doesn't suggest that you are going to return an iterator of files instead of a context manager with a Path-like instance pointing to the root.

Does that make sense?

dberenbaum: I agree that the usage wasn't immediately intuitive to me. How much value is the 2-line `files()` function adding? It might be more clear to use `Repo.open()` and `RepoPath` directly since they follow conventions of Python built-ins.

skshetry: I am open for changing the name of the API, would love to hear other alternative namings. Regarding `Repo`, it's an undocumented API and to replace `files()`, it needs to interact with two different components.

dberenbaum:
`RepoPath` looks great. How is this or `files()` supposed to work for remote repos?

I tried:

```
>>> with files("src", repo="https://github.com/iterative/example-dvc-experiments.git") as root:
...     print(list(root.iterdir()))
...
  File "<stdin>", line 2, in <module>
  File "/Users/dave/Code/dvc/dvc/repo_path.py", line 201, in iterdir
    for _, dirs, files in repo_walk:
  File "/Users/dave/Code/dvc/dvc/fs/repo.py", line 410, in walk
    if not self.exists(top):
  File "/Users/dave/Code/dvc/dvc/fs/repo.py", line 235, in exists
    if dvc_fs.repo.dvcignore.is_ignored(fs, path_info):
  File "/Users/dave/Code/dvc/dvc/ignore.py", line 382, in is_ignored
    if fs.isfile(path):
  File "/Users/dave/Code/dvc/dvc/fs/git.py", line 67, in isfile
    key = self._get_key(path_info)
  File "/Users/dave/Code/dvc/dvc/fs/git.py", line 35, in _get_key
    relparts = path.relative_to(self._root).parts
  File "/opt/homebrew/Cellar/python@3.9/3.9.5/Frameworks/Python.framework/Versions/3.9/lib/python3.9/pathlib.py", line 929, in relative_to
    raise ValueError("{!r} is not in the subpath of {!r}"
ValueError: 'src' is not in the subpath of '/private/var/folders/24/99_tf1xj3vx8k1k_jkdmnhq00000gn/T/tmp787b4c4bdvc-clone' OR one path is relative and the other is absolute.
```

dberenbaum:
Things I find confusing:
1. `dvc.api.files()` returns a `RepoPath`, so do we need to document `RepoPath` anyway? Documenting this and `dvc.api.files()` seems like it might be more complex than using `RepoPath` directly.
2. Does the user need a context manager? It's more overhead and it feels weird combining `open` and `Path`. I guess without it, you would need to create a repo object whenever a `RepoPath` method is called. `pathlib.Path` calls `with self.open()` in every method where it opens a file, so maybe it's not too different?
3. If the user needs a context manager, it seems more convenient at the repo level than combining it with a path. Why should users need a separate context manager to open a different path in the same repo?

It would be more intuitive to me for users to directly use `RepoPath` (or `dvc.api.Path`), which could create/open repo instances as needed internally, or to have something like `dvc.api.repo_open()` and pass that to `RepoPath`.

skshetry:
>     1. `dvc.api.files()` returns a `RepoPath`, so do we need to document `RepoPath` anyway? Documenting this and `dvc.api.files()` seems like it might be more complex than using `RepoPath` directly.

Not sure I understand this, it should be fairly straightforward to document this. 
```files() returns a `RepoPath` object that is an extended version of pathlib.Path.```

I think you are looking it at being two different things, `files()` API and `RepoPath` API, but they are the same thing. Functions do have return types and that's how you interact with them, so they are not two things, but one. For example, `dvc.api.open` returns a `io.IOBase` compatible object that has methods like `read`.

>     2. Does the user need a context manager? It's more overhead and it feels weird combining `open` and `Path`. I guess without it, you would need to create a repo object whenever a `RepoPath` method is called. `pathlib.Path` calls `with self.open()` in every method where it opens a file, so maybe it's not too different?

We do need a context manager to manage resources for the `Repo`. DVC may be keeping handles to a lots of files (mostly git related libraries are doing this), keeping `sqlite` state open. We should always cleanup these resources. This is going to be a big issue on Windows if we don't manage them properly, though we should always clean up these resources in all platforms.

We already have `dvc.api.read` and `dvc.api.open` that does clone repo each time you call.
This API however is meant to be traversable, so we cannot do that on each `open` and `read`, so we do it only once.
Also, context manager is the pythonic way to manage resources.

>     3. If the user needs a context manager, it seems more convenient at the repo level than combining it with a path. Why should users need a separate context manager to open a different path in the same repo?

Users don't need to use separate context managers for different paths, the `files()` returns a root path, and users can use child objects using `root / <path>`.

> It would be more intuitive to me for users to directly use `RepoPath` (or `dvc.api.Path`), which could create/open repo instances as needed internally, or to have something like `dvc.api.repo_open()` and pass that to `RepoPath`.

`dvc.api.repo_open` still needs to return `Repo` instance, which has a huge API surface. We should be hiding those complexities with these APIs as we only need this thing for being traversable with files across the Repo. Also, why separate two different things if we can do so with a single API?

File: dvc/api.py (hunk 1 above)
Comment (daavoo) [reply dvc/api.py:12]: `get_repo_path`?

**Comment by skshetry** [comment]
From discussions in #7379, it seems we need to also expose a way to copy a file/directory similar to `dvc get`.

```
with files() as root:
    root.download(to="path")
```

dberenbaum:
> From discussions in #7379, it seems we need to also expose a way to copy a file/directory similar to `dvc get`.

What would be the benefit over exposing `Repo.get` as part of the public API? I'm not opposed to adding a way to download in the `file` API, but when trying to mimic the CLI, it might be more straightforward to use the `Repo` API.

skshetry:
> > From discussions in #7379, it seems we need to also expose a way to copy a file/directory similar to `dvc get`.
> 
> What would be the benefit over exposing `Repo.get` as part of the public API?

`Repo` API is not a public interface and the fact that it mimics CLI

dberenbaum:
> `Repo` API is not a public interface and the fact that it mimics CLI

Not sure I understand. My point is that it might be easier for users to follow an API that mimics the CLI for simple use cases. Anyway, let me open a separate discussion since this is really about `Repo`.

skshetry:
Closing for now, will need to revisit after all the filesystem changes.

PR #6868, opened 2021-10-26
exp init: initialize dvc repo if uninitialized

skshetry:
Also autostages dvc.yaml/.gitignore/param files by default.

The whole solution feels a bit hacky, as we need to create an uninitialized dvc repo, create a stage in memory and initialize dvc repo if there is no `.dvc` directory. The "creating stage" part does not really work without creating a proper DVC repo, so we have to hack it around.
For now, removing the `RemoteCacheRequired` exception throwing does seem to make it work, but it's hard to guarantee that as we have never supported that before.

### DIFF
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -8,7 +8,12 @@ from typing import TYPE_CHECKING, Dict, Iterable, Optional
 from funcy import compact, lmap
 
 from dvc.command import completion
-from dvc.command.base import CmdBase, append_doc_link, fix_subparsers
+from dvc.command.base import (
+    CmdBase,
+    CmdBaseNoRepo,
+    append_doc_link,
+    fix_subparsers,
+)
 from dvc.command.metrics import DEFAULT_PRECISION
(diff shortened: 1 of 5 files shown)


dberenbaum:
So if the command fails or the user decides not to keep the stage, is the dvc repo initialized?  Would it be better to add a hint to dvc init when running a command outside of a dvc repo?

daavoo:
> So if the command fails or the user decides not to keep the stage, is the dvc repo initialized? Would it be better to add a hint to dvc init when running a command outside of a dvc repo?

Upvoting for hint + docs updates

skshetry:
Closing for now, some of the pre-requisite were backported. It's a bit tricky considering we depend on `Repo` for so many things, and can be opened again if we think it's important to have.

skshetry:
> So if the command fails or the user decides not to keep the stage, is the dvc repo initialized? Would it be better to add a hint to dvc init when running a command outside of a dvc repo?

No, this PR initialized repo at the very end just before creating dvc.yaml file, but after getting confirmation from the user for stage.

PR #6872, opened 2021-10-26
exp init: ask for command before paths

skshetry:
This changes the ordering of prompts, now it'll ask for command first before showing the workspace tree and prompts for the paths.

This is an RFC PR. I find it a bit confusing as it seems to break the linear flow of the prompts. I have yet another suggestion to move this prompt to the end. cc @dberenbaum @daavoo 

### Before this PR

[asciicast](https://asciinema.org/a/RU9HNkdfuq40zKvSsAOgYTygg)

### After this PR

[asciicast](https://asciinema.org/a/2NmpIbMDJpzYe39iI7ei8LeGX)

### Alternative 

[asciicast](https://asciinema.org/a/lrRFXb0OfBmjHnPGRCc5nx07B)

### DIFF
--- a/dvc/repo/experiments/init.py
+++ b/dvc/repo/experiments/init.py
@@ -124,16 +124,14 @@ def init_interactive(
     show_tree: bool = False,
     live: bool = False,
 ) -> Dict[str, str]:
-    primary = lremove(
-        provided.keys(), ["cmd", "code", "data", "models", "params"]
-    )
+    command = provided.pop("cmd", None)
+    primary = lremove(provided.keys(), ["code", "data", "models", "params"])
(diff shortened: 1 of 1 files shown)


skshetry:
Okay, thinking about it more, I am not very sure about this. The command at the start makes more sense given the fact that all of the paths are the effect of it. It might be easier for us to even ask users the follow-up questions. But I still find it a bit confusing as the text jumps a few line after asking for the command.


So, I'd say let's go with the command at the first, and we can see what the feedbacks will be from other members, and we can adjust that later.

dberenbaum (approved):

PR #6869, opened 2021-10-26
machine status: include two new field in the output

karajan1001:
fix #6863 

1. include instance_type and instance_gpu in the output.
2. rerange the field by importance.

* [x] ❗ I have followed the [Contributing to DVC](https://dvc.org/doc/user-guide/contributing/core) checklist.

* [ ] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here.

Thank you for the contribution - we'll try to review it as soon as possible. 🙏

PR #6893, opened 2021-10-31
Machine: use table for machine list and status outputs 

### LINKED ISSUE #6892: Machine: use table for machine list and status outputs
Current, machine list output follows the `dvc config -l`. Which is not suitable in showing multiple machines. Considering use table to replace it.
```
$ dvc machine list
new.cloud=aws
myaws.cloud=aws
myaws.spot_price=100.00000
myaws.instance_hdd_size=10
myaws.instance_type=l
myaws.instance_gpu=k80
```

karajan1001:
fix #6892 

use table for the machine list outputs

Now the outputs looks like
```
$  dvc machine list
┏━━━━━━━┳━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┓
┃ name  ┃ cloud ┃ spot_price ┃ instance_hdd_size ┃ instance_type ┃ ssh_private ┃ instance_gpu ┃
┡━━━━━━━╇━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━━━━━┩
│ new   │ aws   │            │ 20                │               │ ***         │              │
│ myaws │ aws   │ 100.00000  │ 10                │ l             │             │ k80          │
└───────┴───────┴────────────┴───────────────────┴───────────────┴─────────────┴──────────────┘

$ dvc machine status foo
┏━━━━━━┳━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━┳━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┓
┃ name ┃ instance ┃ status  ┃ cloud ┃ instance_hdd_size ┃ instance_ip     ┃
┡━━━━━━╇━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━╇━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━┩
│ foo  │ num_1    │ running │ aws   │ 35                │ 123.123.123.123 │
└──────┴──────────┴─────────┴───────┴───────────────────┴─────────────────┘
```

### DIFF
--- a/dvc/command/machine.py
+++ b/dvc/command/machine.py
@@ -5,6 +5,7 @@ from dvc.command.config import CmdConfig
 from dvc.compare import TabularData
 from dvc.config import ConfigError
 from dvc.exceptions import DvcException
+from dvc.types import Dict, List
 from dvc.ui import ui
 from dvc.utils import format_link
 
@@ -232,8 +233,8 @@ class CmdMachineCreate(CmdBase):
 
 
 class CmdMachineStatus(CmdBase):
+    INSTANCE_FIELD = ["name", "instance", "status"]
     SHOWN_FIELD = [
-        "name",
         "cloud",
         "instance_ip",
         "instance_type",
@@ -241,23 +242,34 @@ class CmdMachineStatus(CmdBase):
         "instance_gpu",
     ]
 
-    def _show_machine_status(self, name: str):
-        ui.write(f"machine '{name}':")
-        all_status = list(self.repo.machine.status(name))
+    def _add_row(
+        self,
+        name: str,
+        all_status: List[Dict],
+        td: TabularData,
+    ):
+
         if not all_status:
-            ui.write("\toffline")
+            row = [name, None, "offline"]
+            td.append(row)
         for i, status in enumerate(all_status, start=1):
-            ui.write(f"\tinstance_num_{i}:")
+            row = [name, f"num_{i}", "running" if status else "offline"]
             for field in self.SHOWN_FIELD:
-                value = status.get(field, None)
-                ui.write(f"\t\t{field:20}: {value}")
+                value = str(status.get(field, ""))
+                row.append(value)
+            td.append(row)
 
     def run(self):
         if self.repo.machine is None:
             raise MachineDisabledError
 
+        td = TabularData(
+            self.INSTANCE_FIELD + self.SHOWN_FIELD, fill_value="-"
+        )
+
         if self.args.name:
-            self._show_machine_status(self.args.name)
+            all_status = list(self.repo.machine.status(self.args.name))
+            self._add_row(self.args.name, all_status, td)
         else:
             name_set = set()
             for level in self.repo.config.LEVELS:
@@ -265,8 +277,11 @@ class CmdMachineStatus(CmdBase):
                 name_set.update(conf.keys())
             name_list = list(name_set)
             for name in sorted(name_list):
-                self._show_machine_status(name)
+                all_status = list(self.repo.machine.status(name))
+                self._add_row(name, all_status, td)
 
+        td.dropna("cols", "all")
+        td.render()
         return 0
 
 
--- a/dvc/command/machine.py
+++ b/dvc/command/machine.py
@@ -2,6 +2,7 @@ import argparse
 
 from dvc.command.base import CmdBase, append_doc_link, fix_subparsers
 from dvc.command.config import CmdConfig
+from dvc.compare import TabularData
 from dvc.config import ConfigError
 from dvc.exceptions import DvcException
 from dvc.ui import ui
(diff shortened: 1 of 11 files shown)


karajan1001:
Wait for #6855  and #6869

karajan1001:
> * Can you drop the cell borders? Unlike `dvc exp show`, I think the output is simple enough, and it will be easier to parse the output with other commands.

I guess machine reads requires `--csv` or `--json` format, dropping borders will still face a different number of blanks.

dberenbaum:
> I guess machine reads requires `--csv` or `--json` format, dropping borders will still face a different number of blanks.

Sorry, I gave `dvc remote list` as an example, but I guess that's not aligned. I meant more like `dvc metrics show/diff`. One of our UI principles is to make output both machine and human readable by default where possible. Maybe @skshetry has thoughts on how this should look.

This also will different depending on the number of columns. If we follow the lead of `dvc remote list`, `dvc machine list` should only have the name and cloud. However, maybe this is different, since people could have multiple `machine` configs using the same cloud provider. I'd probably include at least `image`, but maybe not everything (`ssh_private` is probably not worth showing since it won't be visible anyway). Again, interested to hear from you or others in @iterative/dvc about what to include exclude in both `machine list` and `machine status`.

karajan1001:
Another example is the docker ps and docker images which show in a no border table format.

```
docker images
REPOSITORY                                TAG       IMAGE ID       CREATED         SIZE
ghcr.io/linuxserver/openssh-server        latest    dfcbb139af6d   4 weeks ago     46MB
motoserver/moto                           latest    13087ad4e36c   4 weeks ago     393MB
mcr.microsoft.com/azure-storage/azurite   3.14.2    d54d388ab8a4   8 weeks ago     199MB
mcr.microsoft.com/azure-storage/azurite   3.13.1    bfa88c85ec6b   4 months ago    179MB
rkuprieiev/docker-hdfs                    latest    77e89e3a81a4   13 months ago   1.73GB
rkuprieiev/oss-emulator                   latest    93d6d0f4a3de   16 months ago   843MB
```

dberenbaum:
Right, `dvc exp show` is a special case since we have ascii art all over that makes it already not machine readable. Otherwise, the docker-like output is preferable IMO.

dvc/command/machine.py
```diff
@@ -71,18 +72,57 @@ def run(self):
...
-                self.args.show_origin, self.config, level
-            )
+            self._hide_private(conf)
+            prefix = self._config_file_prefix(True, self.config, level)
             configs = list(self._format_config(conf, prefix))
             if configs:
                 ui.write("\n".join(configs))
+
+    def _show_table(self):
+        td = TabularData(self.TABLE_COLUMNS)
```
daavoo, line 109: Maybe use some `fill_value` ?

daavoo:
> > I guess machine reads requires `--csv` or `--json` format, dropping borders will still face a different number of blanks.
> 
> Sorry, I gave `dvc remote list` as an example, but I guess that's not aligned. I meant more like `dvc metrics show/diff`. One of our UI principles is to make output both machine and human readable by default where possible. Maybe @skshetry has thoughts on how this should look.

 I agree with @karajan1001 about requiring `--json` . I don't know if any of our tables is actually machine-readable; In the docs we describe `--json` as: `prints the command's output in JSON format (machine-readable) instead of a human-readable table`

> This also will different depending on the number of columns. If we follow the lead of `dvc remote list`, `dvc machine list` should only have the name and cloud. However, maybe this is different, since people could have multiple `machine` configs using the same cloud provider. I'd probably include at least `image`, but maybe not everything (`ssh_private` is probably not worth showing since it won't be visible anyway). Again, interested to hear from you or others in @iterative/dvc about what to include exclude in both `machine list` and `machine status`.

I find all current fields, shown in the description, pretty much useful (except for private fields). I can even have two machines with all same fields with the except `instance_gpu` and I would like to be able to distinguish. There is always the option of adding `--only-changed` or other stuff to filter.

dberenbaum:
Machine readable was the wrong term. It's about being able to grep and otherwise parse with unix-like tools. See https://github.com/iterative/dvc/discussions/6046 and https://clig.dev/#output. A lot of other commands have outputs that were designed before we had these guidelines. Is there a reason not to follow this guideline here?

dvc/command/machine.py
dberenbaum: Hm, makes sense I guess. It's probably easier to parse with a fill value also.

karajan1001:
@Mergifyio rebase

pmrowla:
The new table format looks fine to me.

I think it probably looks a bit cleaner without the table borders (and makes it more grep-able), but I don't have a strong opinion here either way.

karajan1001:
@dberenbaum 
Current format
```
$ dvc machine list
name    cloud    spot_price    instance_hdd_size    instance_type    ssh_private    instance_gpu
new     aws      -             20                   -                ***            -
myaws   aws      100.00000     10                   l                -              k80

$ dvc machine status
name    instance    status    cloud    instance_ip      instance_type    instance_hdd_size    instance_gpu
bar     -           offline   -        -                -                -                    -
foo     num_1       running   aws      123.123.123.123  m                35                   None
```

dberenbaum:
Looks great, thanks @karajan1001!

pmrowla (approved):

dvc/command/machine.py
```diff
@@ -2,8 +2,10 @@
 
 from dvc.command.base import CmdBase, append_doc_link, fix_subparsers
 from dvc.command.config import CmdConfig
+from dvc.compare import TabularData
 from dvc.config import ConfigError
 from dvc.exceptions import DvcException
+from dvc.types import Dict, List
```
skshetry, line 8: This should come from `typing` module.

dvc/ui/table.py
```diff
@@ -14,7 +14,7 @@
 SHOW_MAX_WIDTH = 1024
 
 
-CellT = Union[str, "RichText"]  # RichText is mostly compatible with str
+CellT = Union[str, "RichText", None]  # RichText is mostly compatible with str
```
skshetry, line 17: Please don't extend the types here. Table rendering does not support `None` for now.

karajan1001: Then what should I set here ?
https://github.com/iterative/dvc/blob/eac030598cdbfaba4c029f8e1131614916a6c707/dvc/command/machine.py#L253
It looks like an empty string didn't work in this case.

skshetry: Use a fill value "-" or just empty string `""`.

karajan1001: Fill value is a bit too hard code I think, and empty string `""` it was just because it gives empty result instead of a fill value made me using None in this case.

karajan1001: @skshetry How about adding a new type `InputCellT` which can accept `None` value for those input functions like `append`, `extend` or `row_from_dict`?

skshetry: @karajan1001, we'll need to fix that in https://github.com/iterative/dvc/issues/7167. Meanwhile, this may introduce unnecessary bugs, as rendering with `rich` table is not capable of supporting `None`, could you please roll back the type of `CellT`, and make adjustments, please?

karajan1001: > @karajan1001, we'll need to fix that in #7167. Meanwhile, this may introduce unnecessary bugs, as rendering with `rich` table is not capable of supporting `None`, could you please roll back the type of `CellT`, and make adjustments, please?

Sure.

Issue #6878, opened 2021-10-27
pyarrow 6 available for Python 3.10 


skshetry:
Pyarrow 6.0.0 added support for Python 3.10 and is now available. 

We don't need to update our requirement though, we would just need to get rid of python requirement below:
https://github.com/iterative/dvc/blob/0d895e26612a749ced21ac9afb226881b1a79cd1/setup.cfg#L104

We also need to unskip hdfs tests for Python 3.10 which were skipped in https://github.com/iterative/dvc/pull/6422.

@iterative/dvc

PR #6898, opened 2021-11-01
tests: skip based on conditional import on pyarrow

skshetry:
From now on, pyarrow is optional for the tests.

I am adding this PR before fixing https://github.com/iterative/dvc/issues/6878 to check whether we can depend on it in the future (eg: when we will be adding 3.11 support). I will remove requirements for python version on the next PR.

### DIFF
--- a/tests/remotes/hdfs.py
+++ b/tests/remotes/hdfs.py
@@ -2,7 +2,6 @@ import locale
 import os
 import platform
 import subprocess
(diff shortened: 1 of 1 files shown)


skshetry:
Test failures are unrelated, see https://github.com/iterative/dvc/issues/6899.

efiop (approved):

Issue #5012, opened 2020-12-02
dvc <command> fails with known numpy/win10 error; Windows 10 (20H2) for dvc version >=1.10.0; 


mzietlow:
## Bug Report
I have two systems running Win10, one on 2004, the other on 20H2 and I have installed DVC via https://github.com/iterative/dvc/releases.

When executing dvc <any-command>, I get a Win10 specific error message (see my output of $ dvc version). Roughly, this is caused by details in ucrtbase.dll which, in turn, lead to errors in numpy's fmod function.

I have rolled my system-wide numpy installations back to numpy==1.19.3 which alleviates the issue as described (in conjunction with other workarounds) in https://developercommunity.visualstudio.com/comments/1244066/view.html.

However, as dvc does not seem use my system-wide python interpreters, solving it for them doesn't get me off the ground.

#### Suggestion
Although I dislike pinning, I suggest to set numpy==1.19.3 **until** ucrtbase.dll is fixed.
Alternatively, users with Win10 {20H2, 2004} must either:
1. use dvc with version < 1.10.0, or 
2. install via pip and pin numpy == 1.19.3.

Maybe someone else is able to come up with a better solution.

### Please provide information about your setup

**Output of `dvc version`:**

```console
$ dvc version
ERROR: unexpected error - The current Numpy installation ('C:\\Program Files (x86)\\DVC (Data Version Control)\\numpy\\__init__.pyc') fails to pass a sanity check due to a bug in the windows runtime. See this issue for more information: https://tinyurl.com/y3dm3h86
Traceback (most recent call last):
  File "dvc\__main__.py", line 7, in <module>
  File "dvc\main.py", line 119, in main
  File "dvc\info.py", line 33, in get_dvc_info
  File "dvc\info.py", line 126, in _get_supported_remotes
  File "dvc\tree\base.py", line 104, in get_missing_deps
  File "importlib\__init__.py", line 127, in import_module
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\site-packages\PyInstaller\loader\pyimod03_importers.py", line 627, in exec_module
  File "site-packages\pyarrow\__init__.py", line 63, in <module>
[...]

PR #6906, opened 2021-11-02
build: remove numpy restriction

skshetry:
This was added for fixing https://github.com/iterative/dvc/issues/5012 on Windows. But it has been fixed for a long time now, so it should be okay to remove this. But we'll find out on the next release 😅.


skshetry:
Package builds test is here: https://github.com/iterative/dvc/actions/runs/1411045622

PR #6909, opened 2021-11-02
exp init: validate params as soon as it's entered in prompt

### LINKED ISSUE #6865: exp init: when to fail for non-existent params
From https://github.com/iterative/dvc/pull/6750#issuecomment-937219108:

> Can we validate that dependencies exist at the time of entry during interactive mode? It would save a lot of trouble with having to redo the whole stage after it fails at the end.

`exp init` fails if `params.yaml` does not exist since the params are parsed to generate the stage. Related: #6605.

In interactive mode, it would be nice to fail as soon as possible (right after path to params is provided).

For other dependencies, I wonder if `dvc exp init` should also validate that the paths exist. This could help guide users. It could be annoying if users want to do `dvc exp init` before creating projects files, but this seems unlikely to me outside of toy scenarios.

skshetry:

It'll give an error message to the user and re-prompt them if params validation fails.
This mechanism can be extended to other prompts, and we can show a warning message and/or re-prompt them based on validations.

Closes #6865. Part of #6446.

### DIFF
--- a/dvc/repo/experiments/init.py
+++ b/dvc/repo/experiments/init.py
@@ -1,9 +1,18 @@
 import logging
 import os
 from contextlib import contextmanager
-from typing import TYPE_CHECKING, Dict, Iterable, cast
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    cast,
+)
 
-from funcy import compact, lremove, post_processing
+from funcy import compact, lremove
 from rich.prompt import Confirm, Prompt
 from rich.rule import Rule
 from rich.syntax import Syntax
@@ -31,6 +40,10 @@ PROMPTS = {
     "plots": "Path to a [b]plots[/b] file/directory",
     "live": "Path to log [b]dvclive[/b] outputs",
 }
+
+
+class RetryPrompt(Exception):
+    """Used for signalling whether prompts should be retried again."""
 
 
 class RichInputMixin:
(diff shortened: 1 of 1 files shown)


pared (approved):
LGTM.
Please include what issue we fix in the commit comment.

dberenbaum:
Looks good. Some ideas on other potential ways of handling this:
* Adding some sort of wrapper for the error message, like "Please provide a path to a file to be parsed for parameters"?
* Retry rather than fail (prompt again).

skshetry:
Retry makes sense as users can always exit themselves here.

skshetry:
Btw, regarding checking if other dependencies exist from #6865, we can warn users just after they enter those values in the prompt after this PR. @dberenbaum WDYT?

dberenbaum:
> we can warn users just after they enter those values in the prompt after this PR

Makes sense, although it's hard to say if it's too noisy without seeing it. If you have time to try it out, it could be interesting.

skshetry:
@dberenbaum, please see if this PR works or not. I'll add the warning message on top of this (I will only include some checks for whether some dependencies exist or not, I may not be able to do more than that till next week).

dberenbaum:
I like it since it saves me from having to type everything else again (although asking for parameters first could also solve that problem). If we are going to retry on a missing path, can we also retry on failed validation of the file? We could do all the retry logic in a separate PR since it otherwise seems fine.

skshetry:
> If we are going to retry on a missing path, can we also retry on failed validation of the file?

Could you be more specific? What do you mean by failed validation of the file?

dberenbaum:
> Could you be more specific, please? Could you share some examples?
> What do you mean by failed validation of the file?

I mean any errors with parsing the params file (anything other than file does not exist). For example:

```
$ echo ": foo" > params.yaml
$ dvc exp init -i
This command will guide you to set up a default stage in dvc.yaml.
See https://dvc.org/doc/user-guide/project-structure/pipelines-files.

Command to execute: python train.py

Enter the paths for dependencies and outputs of the command.
DVC assumes the following workspace structure:
├── data
├── metrics.json
├── models
├── params.yaml
├── plots
└── src

Path to a code file/directory [src, n to omit]:
Path to a data file/directory [data, n to omit]:
Path to a model file/directory [models, n to omit]:
ERROR: unable to read: 'params.yaml', YAML file structure is corrupted: while parsing a block mapping
did not find expected key
  in "<unicode string>", line 1, column 1
```

What do you think about implementing a retry here also?

Again, this could all be in a separate PR if you prefer.

skshetry:
Closed by #6926.

Issue #6911, opened 2021-11-02
plots: loading templates from pyinstaller breaks binary builds


pared:
In #6550, as @skshetry [noted](https://github.com/iterative/dvc/pull/6550#issuecomment-943070426)
Templates were not added to binary builds.

As a result, when running `dvc init` using binary package, we would not produce the templates in `.dvc/plots`.

PR #6912, opened 2021-11-02
plots: revert templates in packages change

pared:
Fixes: #6911
Reverts: a1b146669c9d9241d424e10cc8266e0c5fdebf94

In #6550 we introduced templates in packages. While utilizing `importlib.resources` and including templates with our package makes sense for other types of builds, in the case of our binary packages and `pyinstaller` we cannot include package data, because of https://github.com/pyinstaller/pyinstaller/issues/2697.

What we could do with pyinstaller would be to create `resources` directory alongside `dvc` and include the templates there. That would require us to load our templates differently depending on whether we can load data from package (other builds) or `pyinstallers` `sys._MEIPASS` when using binary packages.

This seems much more error-prone than the previous approach (keeping predefined templates as python dicts in `dvc/repo/plots/template.py` and is harder to test (it would require us to test binary packages). 

That is why I decided to revert the change about templates in packages.

cc @dberenbaum

### DIFF
--- a/dvc/repo/plots/__init__.py
+++ b/dvc/repo/plots/__init__.py
@@ -187,7 +187,7 @@ class Plots:
         props = props or {}
         template = props.get("template")
         if template:
-            self.templates.load(template)
+            self.templates.get_template(template)
 
         (out,) = self.repo.find_outs_by_path(path)
         if not out.plot and unset is not None:
--- a/dvc/repo/plots/template.py
+++ b/dvc/repo/plots/template.py
@@ -1,23 +1,15 @@
 import json
 import os
-from typing import TYPE_CHECKING, Iterable, Optional
-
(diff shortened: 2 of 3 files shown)


pmrowla:
@pared What was the issue with adding data files in pyinstaller (it looks like you might have linked the wrong issue)?  Is there a reason we can't do something like they've documented here: https://pyinstaller.readthedocs.io/en/stable/spec-files.html#adding-files-to-the-bundle

pared:
@Mergifyio rebase

pared:
@Mergifyio rebase

efiop:
@pared What's the status here?

pared:
Awaiting review

efiop (approved):

PR #6913, opened 2021-11-02
live: Call webbrowser.open

daavoo:

Closes iterative/dvclive#180

### DIFF
--- a/dvc/repo/live.py
+++ b/dvc/repo/live.py
@@ -1,5 +1,10 @@
 import logging
+import webbrowser
+from pathlib import Path
+from platform import uname
 from typing import TYPE_CHECKING, List, Optional
+
+from funcy import once_per_args
 
 from dvc.render.utils import render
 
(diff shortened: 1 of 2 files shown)


dvc/repo/live.py
```diff
@@ -10,15 +14,27 @@
     from dvc.path_info import PathInfo
     from dvc.repo import Repo
 
+webbrowser_open = call_only_once(webbrowser.open)
```
skshetry, line 17: You can use [`funcy.once`](https://funcy.readthedocs.io/en/stable/flow.html#once) instead. Btw, is it possible to have multiple `live` outputs? How does it work with this?

daavoo: Didn't know about funcy.once, thanks!

>  Btw, is it possible to have multiple live outputs?

It is not possible. A stage can only have 1 `live` output. There might be multiple plots in the `live` output directory but that doesn't affect the number of times the HTML generation is called.

skshetry: But can multiple stages have multiple live outputs?

daavoo: Yes. 

I assume that you are pointing to better use `once_per_args`?

dvc/repo/live.py
```diff
@@ -10,15 +15,27 @@
...
+    index_path = render(
+        out.repo, plots, metrics=metrics, path=html_path, refresh_seconds=5
+    )
+
+    url = index_path.as_uri()
+
+    if "Microsoft" in uname().release:
+        url = Path(index_path) / "index.html"
+
+    webbrowser_open(url)
```
skshetry, line 38: Could you please move this piece to `ui`? Maybe as `ui.launch` (similar to [`click.launch`](https://click.palletsprojects.com/en/8.0.x/utils/#launching-applications)) or `ui.open_browser`? We can handle all `relpath`, `uri` and `WSL` stuff inside it. What do you think?

daavoo: Sounds good

daavoo: Opened #6993

karajan1001 (approved):
LGTM

tests/func/test_live.py
```diff
@@ -237,6 +237,7 @@ def test_live_checkpoints_resume(
 
 def test_dvc_generates_html_during_run(tmp_dir, dvc, mocker, live_stage):
     show_spy = mocker.spy(dvc.live, "show")
+    webbrowser_open = mocker.patch("dvc.repo.live.webbrowser_open")
```
skshetry, line 240: @daavoo, it seems that the above change affected all of the tests and now opens the browser while running the tests.

PR #6914, opened 2021-11-02
exp init: track data by default

skshetry:

### DIFF
--- a/dvc/repo/experiments/init.py
+++ b/dvc/repo/experiments/init.py
@@ -8,7 +8,7 @@ from rich.prompt import Prompt
 from rich.rule import Rule
 from rich.syntax import Syntax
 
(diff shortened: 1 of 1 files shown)


dberenbaum:
@skshetry What's the status on this? Do you think it makes sense? We can reconsider giving a hint instead if you think this goes too far.

dberenbaum:
@skshetry Are you working on this? I would say don't spend time on this yet until we have more of a chance to consider the best way to do it. For example, should we actually do a `dvc add` or just create a `.dvc` file with no commit and no version info?

skshetry:
@dberenbaum, I am not working on this, and I don't see an easier way to implement this without duplicating parts of `Repo.add`. 

Closing ...

Issue #4455, opened 2020-08-24
experiments: parallel coordinates plot 


dmpetrov:
I'd love to see a parallel coordinates plot for an experiment.

```
$ dvc exp show --no-pager --include-metrics accuracy --include-params epochs,opt.lr --all-commits
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━┳━━━━━━━━┓
┃ Experiment                        ┃ accuracy ┃ epochs ┃ opt.lr ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━╇━━━━━━━━┩
│ workspace                         │  0.48982 │ 2      │ 0.0    │
│ 794e714 (2020-08-24 03:01:44)     │  0.48982 │ 2      │ 0.0    │
│ ├── a0f6717 (2020-08-24 03:17:31) │  0.84703 │ 2      │ 1e-05  │
│ ├── 59b2759 (2020-08-24 03:12:42) │  0.86678 │ 2      │ 1e-05  │
│ ├── 26656e0 (2020-08-24 03:16:02) │  0.67587 │ 2      │ 0.0    │
│ └── 0a81605 (2020-08-24 03:16:42) │  0.81715 │ 2      │ 1e-05  │
│ 650049c (2020-08-21 20:43:31)     │  0.48607 │ 2      │ -      │
...
$ dvc exp plots 794e714 # show all ephemeral commits of a given commit as a  parallel coordinates plot 
file:///..../plot.png

$ dvc exp plots a0f6717 59b2759 ba39f4  # given set of ephemeral commits
file:///..../plot.png

$ dvc exp plots a0f6717 59b2759 ba39f4 --order epochs,opt.lr,accuracy  # change the order
file:///..../plot.png
```

![image](https://user-images.githubusercontent.com/10081867/91033995-1f983280-e5b9-11ea-9a09-807fe2c295f0.png)

pared:
Seems like it will involve creating a new template.

pmrowla:
See https://vega.github.io/vega-lite/examples/parallel_coordinate.html for example parallel coordinates plot schema

This will also require adding support for parameters in DVC plots, my understanding is that we cannot currently plot values from `params.yaml` since it is not a DVC out
@pared can you confirm this?

pared:
[...]

PR #6931, opened 2021-11-04
dvc render: Add RENDERERS registry.

daavoo:

Pre-requisite for #6840 -> #4455 

These changes *should* not break any existing behavior. 

Refactored existing `render` / `plots.template` logic in order to make it more "plugin" friendly so it's easier to add new renderers (i.e. plotly / parallel coordinates plot).

Open questions:

- Should `ImageRenderer` implement `as_json` (meaning a machine-shareable representation of the "plot"). Maybe we should use base64 for that (potentially even internally as `src`). 

- Should `ImageRenderer` be, somehow, template-based?. It looks that it could be generalized to follow a similar logic as `VegaRenderer` where DVC provides "anchor" data (the `src` in ``) and the template can customize how to insert that data in the HTML. 

- Is there a reason to have `templates` in `dvc.repo.plots`? It looks like there is the only remaining connection to other dvc subsystems that prevents `render` to be extracted as a separate library.

### DIFF
--- a/dvc/render/base.py
+++ b/dvc/render/base.py
@@ -1,0 +1,52 @@
+import abc
+from typing import TYPE_CHECKING, Dict
+
+from dvc.exceptions import DvcException
+
+if TYPE_CHECKING:
+    from dvc.types import StrPath
+
+
+REVISION_FIELD = "rev"
+INDEX_FIELD = "step"
+
+
+class BadTemplateError(DvcException):
+    pass
+
+
+class Renderer(abc.ABC):
+    def __init__(self, data: Dict, templates=None):
+        self.data = data
+        self.templates = templates
+
+        from dvc.render.utils import get_files
+
+        files = get_files(self.data)
+
+        # we assume comparison of same file between revisions for now
+        assert len(files) == 1
+        self.filename = files.pop()
+
+    def _convert(self, path: "StrPath"):
+        raise NotImplementedError
+
+    @property
+    @abc.abstractmethod
+    def DIV(self):
+        raise NotImplementedError
+
+    @property
+    @abc.abstractmethod
+    def SCRIPTS(self):
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def as_json(self):
+        raise NotImplementedError
+
+    def generate_html(self, path: "StrPath"):
+        """this method might edit content of path"""
+        partial = self._convert(path)
+        div_id = f"plot_{self.filename.replace('.', '_').replace('/', '_')}"
+        return self.DIV.format(id=div_id, partial=partial)
--- a/dvc/render/__init__.py
+++ b/dvc/render/__init__.py
@@ -1,33 +1,4 @@
-import logging
-from typing import TYPE_CHECKING, Dict
+from .image import ImageRenderer
+from .vega import VegaRenderer
 
-if TYPE_CHECKING:
-    from dvc.types import StrPath
-
-logger = logging.getLogger(__name__)
-
-
-class Renderer:
-    def __init__(self, data: Dict):
-        self.data = data
-
-        from dvc.render.utils import get_files
-
-        files = get_files(self.data)
-
-        # we assume comparison of same file between revisions for now
-        assert len(files) == 1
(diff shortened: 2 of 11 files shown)


pared:
> Should ImageRenderer implement as_json (meaning a machine-shareable representation of the "plot"). Maybe we should use base64 for that (potentially even internally as src).

I wouldn't do that until there is demand for that. As far as the discussions go with viewer team and @rogermparent, more desirable would be to get URL to the images.

> Should ImageRenderer be, somehow, template-based?. It looks that it could be generalized to follow a similar logic as VegaRenderer where DVC provides "anchor" data (the src in ) and the template can customize how to insert that data in the HTML.

Lets remember that from HTML point of view, Image and Vega behave very similar - they just provide the `<div>` containing the visualization. What would such templating do?

> Is there a reason to have templates in dvc.repo.plots? It looks like there is the only remaining connection to other dvc subsystems that prevents render to be extracted as a separate library.

Good point, there is no need for that, I moved it as is to `VegaRenderer` in order to not complicate #6431 which already was humonguous.

daavoo:
> I wouldn't do that until there is demand for that. As far as the discussions go with @rogermparent, more desirable would be to get URL to the images.

I don't really have the context from that discussions. I was just thinking that `base64` would *technically* be the "machine-readable" equivalent to the vega spec that we are returning in `--show-vega`, because the json actually includes the _parsed data_. 

Regardless, the bigger point was that all renderers should implement a method (`as_json`?) and we should rename `--show-vega` to `--show-json`. It won't matter much if the json representation of the images are URLs or base64, I don't really have a use case requirement for using `base64`.

> Lets remember that from HTML point of view, Image and Vega behave very similar - they just provide the <div> containing the visualization. What would such templating do?

For vega, we have a `DIV` in the renderer class:
 
https://github.com/iterative/dvc/blob/4175f9ffa4948157436283bef11153a7233e90ca/dvc/render/vega.py#L102-L109
 
And users can customize the vega spec (`{partial}`) via _templates_. 

For images, we also have a `DIV` in the renderer:

https://github.com/iterative/dvc/blob/4175f9ffa4948157436283bef11153a7233e90ca/dvc/render/image.py#L13-L18

But in this case users can't customize the `{partial}`, we hardcode the creation of a children `<div>`:

https://github.com/iterative/dvc/blob/4175f9ffa4948157436283bef11153a7233e90ca/dvc/render/image.py#L35-L41

My perspective here would be that users could use _templates_ to customize this children `<div>`.

However, I'm not sure if this makes sense or it is just trying too hard to mimic abstractions between renderers for plots and images.

pared:
> Regardless, the bigger point was that all renderers should implement a method (as_json?) and we should rename --show-vega to --show-json. It won't matter much if the json representation of the images are URLs or base64, I don't really have a use case requirement for using base64.

The renaming will probably take place, thought I would leave `--show-vega` as a hidden option for backward compatibility, and get rid of it in 3.0. Lets not focus too much on representation, for now I am just following what seems to be better fit for dependant projects, we can talk with VSCode team about that.

> My perspective here would be that users could use templates to customize this children <div>.

Makes sense, current state of image rendering is very ugly.

dvc/render/base.py
```diff
@@ -0,0 +1,52 @@
...
+
+        from dvc.render.utils import get_files
+
+        files = get_files(self.data)
+
+        # we assume comparison of same file between revisions for now
+        assert len(files) == 1
+        self.filename = files.pop()
+
+    def _convert(self, path: "StrPath"):
```
pared, line 31: Every time I see this name I am cringing, if you have some comments on this name feel free to share :D

pared (approved):

dvc/render/base.py
daavoo: Maybe in #6943 names (and logic) could be revisited

pared:
@Mergifyio rebase

pared:
@daavoo can you include the information that is related to #6840, #4455, and #6943 in the commit comment? Let's merge it, I am working on #6943 and it will be helpful there.

PR #6933, opened 2021-11-05
exp show: Add parallel coordinates plot.

### LINKED ISSUE #4455: experiments: parallel coordinates plot

daavoo:

Closes #4455

Introduces the following new arguments / behavior to `dvc exp show`:

- `--html` 🤷
- `--sort-by {foo}` Defines the column to create colorscale from (Uses `Experiment` as default) 
- `--out {foo}`
- `--open`
- 
## Examples

Using https://github.com/iterative/demo-fashion-mnist

- Interactivity

https://user-images.githubusercontent.com/12677733/144048287-4baa1784-08c0-4b74-8704-1bb5345f0af8.mp4

- Default: `dvc exp show -T --html`

newplot

- Customize color: `dvc exp show -T --html --sort-by accuracy`

newplot(1)

- Interaction with other options: `dvc exp show -T --html -sort-by accuracy --exclude-metrics loss`

newplot(2)

### DIFF
--- a/tests/func/experiments/test_show.py
+++ b/tests/func/experiments/test_show.py
@@ -637,9 +637,7 @@ def test_show_parallel_coordinates(tmp_dir, dvc, scm, mocker):
     assert main(["exp", "show", "--html"]) == 0
     kwargs = show_experiments.call_args[1]
 
-    assert kwargs["color_by"] == "Experiment"
     html_text = (tmp_dir / "dvc_plots" / "index.html").read_text()
-
     assert all(rev in html_text for rev in ["workspace", "master", "[exp-"])
 
     assert (
@@ -652,11 +650,10 @@ def test_show_parallel_coordinates(tmp_dir, dvc, scm, mocker):
     assert '"label": "metrics.yaml:bar"' not in html_text
 
     assert (
-        main(["exp", "show", "--html", "--color-by", "metrics.yaml:foo"]) == 0
+        main(["exp", "show", "--html", "--sort-by", "metrics.yaml:foo"]) == 0
     )
     kwargs = show_experiments.call_args[1]
 
-    assert kwargs["color_by"] == "metrics.yaml:foo"
     html_text = (tmp_dir / "dvc_plots" / "index.html").read_text()
     assert '"line": {"color": [2.0, 1.0, 2.0]' in html_text
 
@@ -669,3 +666,9 @@ def test_show_parallel_coordinates(tmp_dir, dvc, scm, mocker):
     assert main(["exp", "show", "--html", "--open"]) == 0
 
     webbroser_open.assert_called()
+
+    params_data = {"foo": 1, "bar": 1, "foobar": 2}
+    (tmp_dir / params_file).dump(params_data)
+    assert main(["exp", "show", "--html"]) == 0
+    html_text = (tmp_dir / "dvc_plots" / "index.html").read_text()
+    assert '{"label": "foobar", "values": [2.0, null, null]}' in html_text
--- a/tests/func/experiments/test_show.py
+++ b/tests/func/experiments/test_show.py
@@ -597,3 +597,75 @@ def test_show_only_changed(tmp_dir, dvc, scm, capsys):
     cap = capsys.readouterr()
 
     assert "bar" not in cap.out
+
+
+def test_show_parallel_coordinates(tmp_dir, dvc, scm, mocker):
+    from dvc.command.experiments import show
+
+    webbroser_open = mocker.patch("webbrowser.open")
+    show_experiments = mocker.spy(show, "show_experiments")
+
+    tmp_dir.gen("copy.py", COPY_SCRIPT)
+    params_file = tmp_dir / "params.yaml"
+    params_data = {
+        "foo": 1,
+        "bar": 1,
+    }
+    (tmp_dir / params_file).dump(params_data)
+
+    dvc.run(
+        cmd="python copy.py params.yaml metrics.yaml",
+        metrics_no_cache=["metrics.yaml"],
+        params=["foo", "bar"],
+        name="copy-file",
+        deps=["copy.py"],
+    )
+    scm.add(
+        [
+            "dvc.yaml",
+            "dvc.lock",
+            "copy.py",
+            "params.yaml",
+            "metrics.yaml",
+            ".gitignore",
+        ]
+    )
+    scm.commit("init")
+
+    dvc.experiments.run(params=["foo=2"])
+
+    assert main(["exp", "show", "--html"]) == 0
+    kwargs = show_experiments.call_args[1]
+
+    assert kwargs["color_by"] == "Experiment"
+    html_text = (tmp_dir / "dvc_plots" / "index.html").read_text()
+
+    assert all(rev in html_text for rev in ["workspace", "master", "[exp-"])
+
+    assert (
+        '{"label": "metrics.yaml:foo", "values": [2.0, 1.0, 2.0]}' in html_text
+    )
+    assert (
+        '{"label": "params.yaml:foo", "values": [2.0, 1.0, 2.0]}' in html_text
+    )
+    assert '"line": {"color": [2, 1, 0]' in html_text
+    assert '"label": "metrics.yaml:bar"' not in html_text
+
+    assert (
+        main(["exp", "show", "--html", "--color-by", "metrics.yaml:foo"]) == 0
+    )
+    kwargs = show_experiments.call_args[1]
+
+    assert kwargs["color_by"] == "metrics.yaml:foo"
+    html_text = (tmp_dir / "dvc_plots" / "index.html").read_text()
+    assert '"line": {"color": [2.0, 1.0, 2.0]' in html_text
+
+    assert main(["exp", "show", "--html", "--out", "experiments"]) == 0
+    kwargs = show_experiments.call_args[1]
+
+    assert kwargs["out"] == "experiments"
+    assert (tmp_dir / "experiments" / "index.html").exists()
+
+    assert main(["exp", "show", "--html", "--open"]) == 0
+
+    webbroser_open.assert_called()
--- a/dvc/command/experiments/show.py
+++ b/dvc/command/experiments/show.py
@@ -3,10 +3,12 @@ import logging
 from collections import Counter, OrderedDict, defaultdict
 from datetime import date, datetime
 from fnmatch import fnmatch
+from pathlib import Path
 from typing import TYPE_CHECKING, Dict, Iterable, Optional
 
 from funcy import lmap
 
+from dvc.command import completion
 from dvc.command.base import CmdBase, append_doc_link
 from dvc.command.metrics import DEFAULT_PRECISION
 from dvc.exceptions import DvcException, InvalidArgumentError
@@ -382,6 +384,7 @@ def show_experiments(
     no_timestamp=False,
     csv=False,
     markdown=False,
+    html=False,
     **kwargs,
 ):
     from funcy.seqs import flatten as flatten_list
@@ -429,7 +432,7 @@ def show_experiments(
         kwargs.get("iso"),
     )
 
-    if no_timestamp:
+    if no_timestamp or html:
         td.drop("Created")
 
     for col in ("State", "Executor"):
@@ -466,8 +469,20 @@ def show_experiments(
         }
     )
 
-    if kwargs.get("only_changed", False):
+    if kwargs.get("only_changed", False) or html:
         td.drop_duplicates("cols")
+
+    html_args = {}
+    if html:
+        td.dropna("rows", how="all")
+        td.column("Experiment")[:] = [
+            # remove tree characters
+            str(x).encode("ascii", "ignore").strip().decode()
+            for x in td.column("Experiment")
+        ]
+        out = kwargs.get("out") or "dvc_plots"
+        html_args["output_path"] = (Path.cwd() / out).resolve()
+        html_args["color_by"] = kwargs.get("sort_by") or "Experiment"
 
     td.render(
         pager=pager,
@@ -477,7 +492,12 @@ def show_experiments(
         row_styles=row_styles,
         csv=csv,
         markdown=markdown,
+        html=html,
+        **html_args,
     )
+
+    if html and kwargs.get("open"):
+        return ui.open_browser(Path(out) / "index.html")
 
 
 def _normalize_headers(names, count):
@@ -544,6 +564,9 @@ class CmdExperimentsShow(CmdBase):
                 csv=self.args.csv,
                 markdown=self.args.markdown,
                 only_changed=self.args.only_changed,
+                html=self.args.html,
+                out=self.args.out,
+                open=self.args.open,
             )
         return 0
 
@@ -692,5 +715,24 @@ def add_parser(experiments_subparsers, parent_parser):
             "Only show metrics/params with values varying "
             "across the selected experiments."
         ),
+    )
+    experiments_show_parser.add_argument(
+        "--html",
+        action="store_true",
+        default=False,
+        help="Generate a parallel coordinates plot from the tabulated output.",
+    )
+    experiments_show_parser.add_argument(
+        "-o",
+        "--out",
+        default=None,
+        help="Destination folder to save the HTML to",
+        metavar="<path>",
+    ).complete = completion.DIR
+    experiments_show_parser.add_argument(
+        "--open",
+        action="store_true",
+        default=False,
+        help="Open the HTML directly in the browser.",
     )
     experiments_show_parser.set_defaults(func=CmdExperimentsShow)
--- a/dvc/compare.py
+++ b/dvc/compare.py
@@ -159,7 +159,9 @@ class TabularData(MutableSequence[Sequence["CellT"]]):
 
         index_path = write(
             output_path,
-            renderers=[ParallelCoordinatesRenderer(self, color_by)],
+            renderers=[
+                ParallelCoordinatesRenderer(self, color_by, self._fill_value)
+            ],
         )
         return index_path.as_uri()
 
--- a/dvc/compare.py
+++ b/dvc/compare.py
@@ -153,6 +153,16 @@ class TabularData(MutableSequence[Sequence["CellT"]]):
             writer.writerow(row)
         return buff.getvalue()
 
+    def to_parallel_coordinates(self, output_path, color_by):
+        from dvc.render.html import write
+        from dvc.render.plotly import ParallelCoordinatesRenderer
+
+        index_path = write(
+            output_path,
+            renderers=[ParallelCoordinatesRenderer(self, color_by)],
+        )
+        return index_path.as_uri()
+
     def add_column(self, name: str) -> None:
         self._columns[name] = Column([self._fill_value] * len(self))
         self._keys.append(name)
@@ -173,6 +183,14 @@ class TabularData(MutableSequence[Sequence["CellT"]]):
 
         if kwargs.pop("csv", False):
             ui.write(self.to_csv(), end="")
+
+        elif kwargs.pop("html", False):
+            ui.write(
+                self.to_parallel_coordinates(
+                    kwargs["output_path"], kwargs.get("color_by")
+                )
+            )
+
         else:
             ui.table(self, headers=self.keys(), **kwargs)
 
--- a/dvc/render/plotly.py
+++ b/dvc/render/plotly.py
@@ -26,11 +26,15 @@ class ParallelCoordinatesRenderer(Renderer):
 
     # pylint: disable=W0231
     def __init__(
-        self, tabular_data: "TabularData", color_by: Optional[str] = None
+        self,
+        tabular_data: "TabularData",
+        color_by: Optional[str] = None,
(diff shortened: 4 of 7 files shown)


dvc/command/experiments.py
```diff
@@ -474,6 +476,18 @@ def show_experiments(
...
 
+    extra_render_args = {}
+    if html:
+        td.dropna("rows")
+        td.column("Experiment")[:] = [
+            # remove tree characters
+            str(x).encode("ascii", "ignore").strip().decode()
+            for x in td.column("Experiment")
+        ]
+        rel = kwargs.get("out") or "dvc_plots"
```
karajan1001, line 487: Better to use a variable for this `dvc_plots`? Keep it synchronize with what in `dvc/command/plots.py`

dvc/command/experiments.py
```diff
@@ -482,8 +496,26 @@ def show_experiments(
         row_styles=row_styles,
         csv=csv,
         markdown=markdown,
+        html=html,
+        **extra_render_args,
     )
 
+    if html and kwargs.get("open"):
```
karajan1001, line 503: I guess it is better to extract this opening of browsers and merge it with 
https://github.com/iterative/dvc/blob/3486435a7fcc76e031f7ed23335a7354441da634/dvc/command/plots.py#L69-L82

karajan1001: And @skshetry what do you think should it go with render with an argument or keep it in this show_experiment function?

dvc/command/experiments.py
```diff
@@ -525,6 +557,10 @@ def run(self):
             fill_value = "" if self.args.csv else FILL_VALUE
             iso = True if self.args.csv else False
 
+            if self.args.html:
+                self.args.only_changed = True
```
karajan1001, line 561: As this logic always goes with html, I guess it is better to move them into `show_experiments` or `show_experiments` will go wrong if it was called elsewhere without them.

dvc/command/experiments.py
daavoo: Yeah. @skshetry already suggested that in https://github.com/iterative/dvc/pull/6913#discussion_r748392459 So I guess now it makes even more sense as we have another place where its used

daavoo: I think it should receive a relpath and construct the abspath inside for non WSL cases

dvc/command/experiments.py
daavoo: Makes sense. I doubt I had is whether to use the same default folder as plots o to have a different one instead (i.e. dvc_exp?)

Wdyt?

cc @dberenbaum

dvc/command/experiments.py
daavoo: Opened #6993

dvc/command/experiments.py
daavoo: I tried to follow the logic added for `csv` above (i.e. set `iso = True`) which modifies arguments for `show_experiments`.

`show_experiments` won't really fail if these args have different values, these are just defaults that I assume to make the plot cleaner.

karajan1001: The logic here is a bit complex. Any better ways to handle them? @skshetry

daavoo: I moved the logic to inside `show_experiments`

dvc/command/experiments.py
daavoo: @karajan1001 Where this kind of constants should be placed in order to be accessed by different commands (i.e. `dvc plots` / `dvc exp`) ? 

A constant under `Repo` like the existing `Repo.DVC_DIR`?

dberenbaum:
My biggest question is whether this belongs in `dvc exp show` or might be better as a separate command, like `dvc exp plot`? A `dvc exp plot` command could also potentially show other things, like `dvc plots diff` between experiments, or checkpoint trends (like dvclive plots).

dberenbaum (commented):
Looks great!

One other comment: behavior seems kind of odd where a value is missing, like the points show up in unpredictable patterns. Would it make sense to have a set value for missing data so it all gets grouped together? Not sure if this is possible for numeric fields.

dvc/command/experiments.py
```diff
@@ -1045,6 +1085,34 @@ def add_parser(subparsers, parent_parser):
             "across the selected experiments."
         ),
     )
+    experiments_show_parser.add_argument(
+        "--html",
+        action="store_true",
+        default=False,
+        help="Generate a parallel coordinates plot from the tabulated output.",
+    )
```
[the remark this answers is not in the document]
daavoo: About that, should we introduce instead `--format` (or `--output-type` or something) and make (`csv`, `json`, `md`, `html`) just the valid values for this single option?

daavoo:
> One other comment: behavior seems kind of odd where a value is missing, like the points show up in unpredictable patterns. Would it make sense to have a set value for missing data so it all gets grouped together? Not sure if this is possible for numeric fields.

Could you share some example data for me to try?

dvc/command/experiments.py
```diff
@@ -1045,6 +1081,34 @@ def add_parser(subparsers, parent_parser):
...
+    experiments_show_parser.add_argument(
+        "--color-by",
```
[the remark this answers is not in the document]
daavoo: Really like the idea

dvc/command/experiments.py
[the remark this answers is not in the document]
dberenbaum: I like it, but not sure if it makes sense for other commands which mostly have only`--json` format.

dvc/command/experiments.py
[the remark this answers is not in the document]
dberenbaum: Also, can we use `--sort-order` if we do this to allow for inverting the scale?

dberenbaum:
> Could you share some example data for me to try?

Add `foo: 1` to `params.yaml` in your https://github.com/iterative/demo-fashion-mnist and do `dvc exp show -T --html`:

image

I would expect `foo` to have all the experiments grouped together except for `workspace`.

dberenbaum:
Does it make sense to always/ever include the experiment names? Is it possible to make a tooltip hover with experiment names? I think it can get distracting with lots of experiments:

image

To me, the main point of parallel coordinates plots is when there are too many to look at each individual experiment, so I wouldn't usually expect to see each experiment name.

dvc/command/experiments.py
karajan1001: I'm not quite sure, for myself I would place them to `dvc.plot.Plot`?

tests/func/experiments/test_show.py
```diff
@@ -582,3 +582,74 @@ def test_show_only_changed(tmp_dir, dvc, scm, capsys):
...
+        [
+            "dvc.yaml",
+            "dvc.lock",
+            "copy.py",
+            "params.yaml",
+            "metrics.yaml",
+            ".gitignore",
+        ]
+    )
+    scm.commit("init")
```
pared, line 618: How much do we need `bar`? Seems like besides that this is identical with `exp_stage` fixture

pared (changes requested):

dvc/command/experiments.py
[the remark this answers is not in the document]
daavoo: Sure

tests/func/experiments/test_show.py
daavoo: I think I was going to use it for checking that `--only-changed` is implicitly called when `--html` is passed so `bar` should not be present in the table

daavoo: I need to revise the tests as I don't see that check 😅

daavoo:
> > Could you share some example data for me to try?
> 
> Add `foo: 1` to `params.yaml` in your https://github.com/iterative/demo-fashion-mnist and do `dvc exp show -T --html`:
> 
> image
> 
> I would expect `foo` to have all the experiments grouped together except for `workspace`.

Actually `foo` should be discarded in this case by `drop_duplicates`, so it is technically a bug. Need to check it. 🤔

Anyhow would you expect to see `foo` in the plot if there is only 1 value?

dberenbaum:
> Actually `foo` should be discarded in this case by `drop_duplicates`, so it is technically a bug. Need to check it. 🤔
> 
> Anyhow would you expect to see `foo` in the plot if there is only 1 value?

That might be okay for now, although eventually we will need a better solution since it means missing a field for a single experiment makes it invisible, right? We might need an N/A value for each field where some data is missing, or we can just have a missing point and corresponding gap in the line for experiments that have no value for that field.

dvc/command/experiments.py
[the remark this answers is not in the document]
daavoo: About `--sort-by`, I implemented it and it kind of feels good for scalars but it feels a little strange for categorical values (i.e. `--sort-by params.yaml:optimizer`). Perhaps this is a problem derived from integrating the plot under `dvc exp show`; If we had a separate command, having a specific `--color-by` option would be more intuitive but I guess that the motivation here is to reduce the number of available options.

About `--sort-order`, what would you expect to be inverted? 
We don't currently expose the colorbar selection, would you expect to be the colors inverted or the ticks of the colorscale on the side?

dvc/command/experiments.py
daavoo: @pared thoughts?

dvc/command/experiments.py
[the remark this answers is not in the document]
dberenbaum: > feels a little strange for categorical values

Is this any different for the regular table view? It's an awkward sort column in either case.

> having a specific --color-by option would be more intuitive

Do you mean it's useful to color by a categorical column, but sorting by it is awkward? I think if I am coloring by that column, I still want a predictable order for it. Maybe I need to see an example to understand the problem.

> About --sort-order, what would you expect to be inverted?

If you invert the ticks, won't the colors of the individual lines be inverted? That's what I would expect.

dvc/command/experiments.py
```diff
@@ -471,9 +473,21 @@ def show_experiments(
...
     )
 
-    if kwargs.get("only_changed", False):
+    if kwargs.get("only_changed", False) or html:
         td.drop_duplicates("cols")
 
+    extra_render_args = {}
+    if html:
+        td.dropna("rows")
+        td.column("Experiment")[:] = [
```
daavoo, line 482: @dberenbaum continuing discussion here because it is easier for me to track parallel conversations on threaded comments than on P.R. level comments.

> Does it make sense to always/ever include the experiment names? Is it possible to make a tooltip hover with experiment names? I think it can get distracting with lots of experiments:
> 
> image
> 
> To me, the main point of parallel coordinates plots is when there are too many to look at each individual experiment, so I wouldn't usually expect to see each experiment name.

Looks like hovering on lines of these plots is not supported nor in the plans to implement unless a sponsor funds the feature (https://github.com/plotly/plotly.js/issues/3012).

Assuming hover is not an option, would you think that adjusting the height of the plot so that the labels are not that messy would help? 

I will also try to find some other alternatives

shcheklein: > Assuming hover is not an option, would you think that adjusting the height of the plot so that the labels are not that messy would help?

does it make to sponsor / work with them together?

daavoo: > Genuine question: do you think showing dozens of experiment names can be helpful? I wouldn't find it useful and would find differentiating the colors of each difficult, but maybe that's just me.

**Without** on hover tooltip, I would say that it is still better to have an awkward column (and work around making it less awkward) than not having a way to identify each individual line. **With** on hover tooltip, I would drop the `Experiment` column, no doubts.

> does it make to sponsor / work with them together?

It could be a worthy investment if we consider the underlying javascript component (where the changes should be made) to be reusable by other tools / products (i.e. `vscode`, `Studio`). 

Alternatively, if we consider this feature a blocker, we could explore other open-source alternatives for rendering the plot like: https://github.com/facebookresearch/hiplot . The renderer might be a little more tricky to implement but I think it would still be doable.

dvc/command/experiments.py
```diff
@@ -471,9 +473,21 @@ def show_experiments(
         }
     )
 
-    if kwargs.get("only_changed", False):
+    if kwargs.get("only_changed", False) or html:
         td.drop_duplicates("cols")
 
+    extra_render_args = {}
+    if html:
+        td.dropna("rows")
```
daavoo, line 481: > > Anyhow would you expect to see `foo` in the plot if there is only 1 value?
> 
> That might be okay for now, although eventually we will need a better solution since it means missing a field for a single experiment makes it invisible, right? We might need an N/A value for each field where some data is missing, or we can just have a missing point and corresponding gap in the line for experiments that have no value for that field.

I fixed 2 separated bugs:

[One](https://github.com/iterative/dvc/pull/7064) was unexpectedly making `foo` slip through the `only_changed` filter. So, in the example you shared before, the `foo` column won't be shown on the plot (whether this is desirable or not doesn't affect the fix).

The other was related to how `NaNs` were treated in categorical columns (unique values were being assigned to each NaN). The fix is now included as part of this p.r. making NaNs converge to a single point.

If we didn't filter the `foo` column per the first fix, it would look like the following:

```dvc exp show -T --html --sort-by foo```

newplot(3)

daavoo: > That might be okay for now, although eventually we will need a better solution since it means missing a field for a single experiment makes it invisible, right?

There are 2 different filters:

First, ` td.drop_duplicates("cols")` removes any columns with 0 variance. Here we could discuss if we want to consider `NaNs` as unique values (keep `foo` in the example`) or not, I don't have a strong opinion here.

After that, `td.dropna("rows")` drops rows with *any* NaNs. This was my decision to prevent populating the plot with useless lines (i.e. commits missing some metric/param) but also because NaNs are very problematic for scalar columns as it totally messes up the plot (scalar columns are converted to categorical if NaN is present). NaNs in categorical columns are ok.

We could rework the second filter in many ways:

- Using `all` option instead of `any` (works for useless lines but does not solve scalar columns with NaNs).
- Adding a new `thresh` option mimicking pandas (magic number hard to choose ).
- Being smart about choosing the fill value for NaNs in scalar columns.

> or we can just have a missing point and corresponding gap in the line for experiments that have no value for that field.

This is not supported by plotly.

dvc/command/experiments.py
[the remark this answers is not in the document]
daavoo: > Is this any different for the regular table view? It's an awkward sort column in either case.

In table it's an awkward sort but in plot, it is that + the option that defines the color so it makes it a little more strange.

> Do you mean it's useful to color by a categorical column, but sorting by it is awkward? I think if I am coloring by that column, I still want a predictable order for it. Maybe I need to see an example to understand the problem.

I just mean that the name is strange, sometimes I would just use a categorical column to color, don't caring about the order and the option being named `--sort-by` is strange.

> If you invert the ticks, won't the colors of the individual lines be inverted? That's what I would expect.

Ticks and colorscale are decoupled in plotly that's why I'm not sure what effect would be expected when using `--sort-order` in the plot.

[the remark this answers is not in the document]
dberenbaum: I see what you mean about the awkwardness of `--sort-by`. I would still vote to use it since otherwise we need a flag specific to this output type, plus we need to figure out how that should interact with `--sort-by` and `--sort-order` here.

For `--sort-order`, I would invert the ticks.

dvc/command/experiments.py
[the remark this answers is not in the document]
casperdcl: potentially https://docs.python.org/3/library/argparse.html#argument-groups

dvc/command/experiments.py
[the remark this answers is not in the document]
daavoo: `--sort-by` is already pushed. working on `--sort-order`

dvc/command/experiments.py
daavoo: @dberenbaum Here is the latest state:

- Rows are only dropped if *all* of the columns are NaN.

- `NaN` in **categorical** columns are  represented with a `Missing` value. This makes NaN lines converge to a single point:

newplot(13)

- `NaN` in **scalar** columns (creating `foo: 1` in the example repo) are represented as `null`, leaving the behavior dependent on `plotly` internals (for better or for worse). This makes NaN lines converge to a point beyond the limits of the plot:

newplot(12)

dberenbaum (approved):

dvc/command/experiments.py
daavoo: Resolving after offline discussion. We would keep the column as it is for now and discuss options for improving the line-experiment association in https://github.com/iterative/dvc/issues/7080

dvc/command/experiments.py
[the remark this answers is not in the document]
daavoo: Resolving. `--sort-order` won't be included as part of this P.R. Opened https://github.com/iterative/dvc/issues/7081 to track it

dvc/command/experiments.py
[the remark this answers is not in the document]
daavoo: Tracked in https://github.com/iterative/dvc/issues/7082

dvc/command/experiments.py
daavoo: Opened https://github.com/iterative/dvc/issues/7083 for improving the `Rows are only dropped if all of the columns are NaN` to only consider metrics and params columns.

pared (approved):

casperdcl:
just to check - the colourbar missing a label in the video/images :)

daavoo:
> just to check - the colourbar missing a label in the video/images :)

Fixed and pushed. Nothing like a good Q/A demo in All Hands

dvc/command/experiments.py
daavoo: Tracked in https://github.com/iterative/dvc/issues/7088

jorgeorpinel:
> `--html` 🤷

Why something as generic and conceptually disconnected as "HTML"? How about `--parallel-coordinates-plot`/`--pcp`? Or a separate command as @dberenbaum mentioned

jorgeorpinel:
> `--out {foo}`

I understand the default here is `dvc_plots/` but `dvc plots` uses `.` by default I think. Seems inconsistent?

daavoo:
> > `--out {foo}`
> 
> I understand the default here is `dvc_plots/` but `dvc plots (show)` uses `.` by default I think. Seems inconsistent?

It is the same for plots. At least in current version, DVC pots are saved in `dvc_plots`

jorgeorpinel:
According to https://dvc.org/doc/command-reference/plots/show#options `dvc plots show` uses the same path as the source image or data series file.

It would be more consistent to just use the cwd `.` I think, as we may be in a subdirectory or something. `dvc_plots/` seems to assume we're in the root of the repo.

Or are the docs out of date?

jorgeorpinel:
Also, the choice of `dvc_plots/` by default reinforces the argument that this feature should be in `dvc plots pcp` or maybe `dvc exp plots --pcp` (exp. if we plan to have more exp-specific plots).

jorgeorpinel:
> But maybe I'm missing a discussion that already happened (lots of collapsed comments above).

shcheklein:
To be honest I also don't quite understand why `--html`? :) @daavoo could shed some light on this, please?

shcheklein:
To be honest `--html` for me would mean an HTML table, not plot. I would consider renaming this to be explicit about its meaning. I think `--plot` is a bit more explicit. But can limit us later if we decide to introduce more types of plots. Probably `--pcp` is the best unfortunately.

daavoo:
> Or are the docs outdated?

Yes they are https://github.com/iterative/dvc.org/pull/3110

PR #6936, opened 2021-11-07
webhdfs: expose kerberos and https options

### LINKED ISSUE #6935: Allow passing fsspec options to WebHDFS to enable Kerberos and https
After #6662 it is no longer possible to use custom hdfscli clients with WebHDFS, as a result we are no longer able to connect using Kerberos or HTTPS.

fsspec allows passing in options to enable kerberos and https, we simply need to pass them through and expose them in the DVC config file:
https://filesystem-spec.readthedocs.io/en/latest/_modules/fsspec/implementations/webhdfs.html

This would also be an opportunity to remove the no longer used hdfscli options.

gudmundur-heimisson:
This will re-enable some of the functionality lost after #6662.

I have not created a documentation pull request for this yet, since I expect there may be some discussion around the naming or organization of these options, since this is my first contribution to this repo and I'm not sure about all the conventions.

Once the naming is settled I will of course create a documentation PR as well.

Note that I have renamed `webhdfs_token` to `token` since that is what is indicated in the current docs.

Please let me know if you want the naming convention of these to be changed.

Fixes #6935

### DIFF
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -179,10 +179,12 @@ SCHEMA = {
                 },
                 "hdfs": {"user": str, "kerb_ticket": str, **REMOTE_COMMON},
                 "webhdfs": {
-                    "hdfscli_config": str,
-                    "webhdfs_token": str,
-                    "user": str,
-                    "webhdfs_alias": str,
+                    "kerberos": Bool,
+                    "kerberos_principal": str,
+                    "proxy_to": str,
+                    "ssl_verify": Any(Bool, str),
+                    "token": str,
+                    "use_https": Bool,
                     **REMOTE_COMMON,
                 },
                 "azure": {
--- a/dvc/fs/webhdfs.py
+++ b/dvc/fs/webhdfs.py
@@ -32,7 +32,7 @@ class WebHDFSFileSystem(CallbackMixin, FSSpecWrapper):
 
     def _prepare_credentials(self, **config):
         self._ssl_verify = (
-            config.pop("ssl_verify") if "ssl_verify" in config else True
+            config.pop("ssl_verify", True)
         )
         config["kerb_kwargs"] = {}
         if "kerberos_principal" in config:
--- a/dvc/fs/webhdfs.py
+++ b/dvc/fs/webhdfs.py
@@ -31,9 +31,14 @@ class WebHDFSFileSystem(CallbackMixin, FSSpecWrapper):
         )
 
     def _prepare_credentials(self, **config):
-        if "webhdfs_token" in config:
-            config["token"] = config.pop("webhdfs_token")
-
+        self._ssl_verify = (
+            config.pop("ssl_verify") if "ssl_verify" in config else True
+        )
+        config["kerb_kwargs"] = {}
+        if "kerberos_principal" in config:
+            config["kerb_kwargs"]["principal"] = config.pop(
+                "kerberos_principal"
+            )
         return config
 
     @wrap_prop(threading.Lock())
@@ -41,7 +46,9 @@ class WebHDFSFileSystem(CallbackMixin, FSSpecWrapper):
     def fs(self):
         from fsspec.implementations.webhdfs import WebHDFS
 
-        return WebHDFS(**self.fs_args)
+        fs = WebHDFS(**self.fs_args)
+        fs.session.verify = self._ssl_verify
+        return fs
 
     def checksum(self, path_info):
         path = self._with_bucket(path_info)
--- a/dvc/fs/webhdfs.py
+++ b/dvc/fs/webhdfs.py
@@ -32,11 +32,9 @@ class WebHDFSFileSystem(CallbackMixin, FSSpecWrapper):
 
     def _prepare_credentials(self, **config):
         self._ssl_verify = config.pop("ssl_verify", True)
-        config["kerb_kwargs"] = {}
-        if "kerberos_principal" in config:
-            config["kerb_kwargs"]["principal"] = config.pop(
-                "kerberos_principal"
-            )
+        principal = config.pop("kerberos_principal", None)
+        if principal:
+            config["kerb_kwargs"] = {"principal": principal}
         return config
 
     @wrap_prop(threading.Lock())
--- a/dvc/fs/webhdfs.py
+++ b/dvc/fs/webhdfs.py
@@ -31,9 +31,7 @@ class WebHDFSFileSystem(CallbackMixin, FSSpecWrapper):
         )
 
     def _prepare_credentials(self, **config):
-        self._ssl_verify = (
-            config.pop("ssl_verify", True)
-        )
+        self._ssl_verify = config.pop("ssl_verify", True)
         config["kerb_kwargs"] = {}
         if "kerberos_principal" in config:
             config["kerb_kwargs"]["principal"] = config.pop(
--- a/setup.cfg
+++ b/setup.cfg
@@ -112,6 +112,7 @@ ssh_gssapi = sshfs[gssapi]>=2021.8.1
 webdav = webdav4>=0.9.3
 # not to break `dvc[webhdfs]`
 webhdfs =
+    requests-kerberos==0.13.0
 terraform = tpi[ssh]>=2.1.0
 tests =
     %(terraform)s
--- a/tests/unit/remote/test_webhdfs.py
+++ b/tests/unit/remote/test_webhdfs.py
@@ -1,21 +1,54 @@
+from unittest.mock import Mock, create_autospec
+
+import pytest
+import requests
+
 from dvc.fs.webhdfs import WebHDFSFileSystem
 
+host = "host"
+kerberos = False
+kerberos_principal = "principal"
+port = 12345
+proxy_to = "proxy"
+ssl_verify = False
+token = "token"
+use_https = True
 user = "test"
-webhdfs_token = "token"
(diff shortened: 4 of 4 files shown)


gudmundur-heimisson:
I'm not sure if I linked the issue correctly but it is #6935

efiop (approved):
Thank you so much! 🙏

efiop:
Oh, looks like we also didn't update the docs before 🙁 We'll need to https://dvc.org/doc/command-reference/remote/modify accordingly.

gudmundur-heimisson:
I will create a pull request in the dvc.org repo and link here once it's up for the documentation.

dvc/fs/webhdfs.py
```diff
@@ -31,17 +31,24 @@ def _get_kwargs_from_urls(urlpath):
         )
 
     def _prepare_credentials(self, **config):
-        if "webhdfs_token" in config:
-            config["token"] = config.pop("webhdfs_token")
-
+        self._ssl_verify = (
+            config.pop("ssl_verify") if "ssl_verify" in config else True
+        )
```
efiop, line 36:
```suggestion
        self._ssl_verify = (
            config.pop("ssl_verify", True)
        )
```

dvc/fs/webhdfs.py
```diff
@@ -31,17 +31,24 @@ def _get_kwargs_from_urls(urlpath):
         )
 
     def _prepare_credentials(self, **config):
-        if "webhdfs_token" in config:
-            config["token"] = config.pop("webhdfs_token")
-
+        self._ssl_verify = (
+            config.pop("ssl_verify", True)
+        )
```
efiop, line 36:
```suggestion
        self._ssl_verify = config.pop("ssl_verify", True)
```

dvc/fs/webhdfs.py
```diff
@@ -31,17 +31,22 @@ def _get_kwargs_from_urls(urlpath):
...
         return config
 
     @wrap_prop(threading.Lock())
     @cached_property
     def fs(self):
         from fsspec.implementations.webhdfs import WebHDFS
 
-        return WebHDFS(**self.fs_args)
+        fs = WebHDFS(**self.fs_args)
+        fs.session.verify = self._ssl_verify
```
efiop, line 48: Is this one the same as using `use_https`?

efiop: In the long term, it might be worth adding an option for `WebHDFS` class to pass some args to the session it creates.

dvc/fs/webhdfs.py
```diff
@@ -31,17 +31,22 @@ def _get_kwargs_from_urls(urlpath):
...
     def _prepare_credentials(self, **config):
-        if "webhdfs_token" in config:
-            config["token"] = config.pop("webhdfs_token")
-
+        self._ssl_verify = config.pop("ssl_verify", True)
+        config["kerb_kwargs"] = {}
+        if "kerberos_principal" in config:
+            config["kerb_kwargs"]["principal"] = config.pop(
+                "kerberos_principal"
+            )
```
efiop, line 39:
```suggestion
        principal = config.pop("kerberos_principal", None)
        if principal:
            config["kerb_kwargs"] = {"principal": principal}
```

dvc/fs/webhdfs.py
gudmundur-heimisson: I think it also might be useful to allow the caller to pass in its own session object, long term. Something like what pylivy does for example:
https://github.com/acroz/pylivy/blob/master/livy/session.py#L112

gudmundur-heimisson:
Here is the documentation pull request:
https://github.com/iterative/dvc.org/pull/3009

dvc/config_schema.py
```diff
@@ -179,10 +179,12 @@ class RelPath(str):
                 },
                 "hdfs": {"user": str, "kerb_ticket": str, **REMOTE_COMMON},
                 "webhdfs": {
-                    "hdfscli_config": str,
-                    "webhdfs_token": str,
-                    "user": str,
-                    "webhdfs_alias": str,
+                    "kerberos": Bool,
+                    "kerberos_principal": str,
+                    "proxy_to": str,
```
jorgeorpinel, line 184: Should it be `proxy_user` or `superuser` per https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/Superusers.html ? "proxy to" could refer to many things...

jorgeorpinel: Or `hadoop_proxy_user`

dvc/config_schema.py
```diff
@@ -179,10 +179,12 @@ class RelPath(str):
...
+                    "ssl_verify": Any(Bool, str),
+                    "token": str,
+                    "use_https": Bool,
```
jorgeorpinel, line 187: `use_https` - should it be swebhdfs per https://hadoop.apache.org/docs/r3.1.0/api/org/apache/hadoop/fs/SWebHdfs.html ? Or is the point that it enables HTTPS (SWebHdfs being an implementation detail)

jorgeorpinel: `ssl_verify` - if it's Hadoop specific should it be `hadoop_ssl_verify`?

dvc/config_schema.py
```diff
@@ -179,10 +179,12 @@ class RelPath(str):
...
                 "webhdfs": {
-                    "hdfscli_config": str,
-                    "webhdfs_token": str,
-                    "user": str,
-                    "webhdfs_alias": str,
+                    "kerberos": Bool,
+                    "kerberos_principal": str,
+                    "proxy_to": str,
+                    "ssl_verify": Any(Bool, str),
+                    "token": str,
```
jorgeorpinel, line 186: `token` - should it be `hadoop_delegation_token` to be explicit? Per https://hadoop.apache.org/docs/current/hadoop-aws/tools/hadoop-aws/delegation_tokens.html#Background:_Hadoop_Delegation_Tokens

dvc/config_schema.py
gudmundur-heimisson: Yes for this initial pull request I just copied the names of parameters from `fsspec` and left them as a first draft to have a discussion around.

I think `proxy_user` would be clearer, and in the hadoop docs I feel like I see proxy user more often than superuser.

If we do something like `hadoop_proxy_user` I feel like maybe we should prefix all the other options with `hadoop` or `webhdfs` as well to be consistent?

It seems like the other DVC protocol config options do not do this kind of prefixing except google drive, but I can see how it would make sense.

dvc/config_schema.py
gudmundur-heimisson: If my understanding of the Hadoop docs is correct, then `swebhdfs` protocol is `webhdfs` over `https`, but fsspec (and DVC) does not actually support `swebhdfs` protocol in URL strings and instead uses `webhdfs` with a `use_https` flag being true.

The question is if we want to create a new `swebhdfs` protocol to be consistent with the actual standard, or keep it as it is to avoid multiplying protocols on our end and just use this `use_https` flag.

As `fsspec` shows support for actually using `swebhdfs` in URL strings is rather inconsistent in the hadoop ecosystem it seems, so it wouldn't be crazy to just use `webhdfs` and then provide this flag.

Regarding the hadoop prefix see other comments.

dvc/config_schema.py
gudmundur-heimisson: I think this makes sense, with the previous caveat about prefixes.

dvc/config_schema.py
jorgeorpinel: Thanks @gudmundur-heimisson . It's an interesting question and I'm sure @efiop will know best what to do. I personally do like the prefixing idea

efiop:
@jorgeorpinel We've released this already, so I think renaming could wait as not critical. Also `hadoop_` prefixes are a different topic for all remotes, as we've discussed during our meeting.

skshetry:
This requires system libraries for kerberos to be installed, so I'd suggest to roll back this change. Strictly speaking, the `requests-kerberos` is not our dependency.

gudmundur-heimisson:
You don't need the system libraries of kerberos to be installed to install requests_kerberos, you can get a ticket from another machine and use that, we do that for some of our processes that use service accounts.

As an FYI, if the kerberos capability is removed then this renders DVC unusable with secured hdfs clusters over webhdfs, which means it would not be usable in an enterprise setting.

If you decide to roll this back I would recommend rolling back all the way to before fsspec was used for webhdfs so that at least the client can be customized, since otherwise it is impossible to use with a secured cluster.

skshetry:
@gudmundur-heimisson, sorry I meant to rollback the dependency that you have added in `webhdfs` extras in `setup.cfg`, not the whole PR/functionality (was not able to comment on code changes somehow). fsspec also does not specify them directly and expects uses to install them, and I think we should recommend the same here. Maybe we can print a message on the usages of those config if it's not installed?

I have been trying to [install it on CI](https://github.com/iterative/dvc-e2e/runs/4578653280?check_suite_focus=true#step:6:197) but one of it's indirect dependency `krb5` needs to be built before it can be used which requires system libraries.

Also see https://github.com/iterative/dvc-webhdfs/blob/0ce2ab527eb3ab68f3e5420b32caf773330174e6/.github/workflows/tests.yaml#L31 as well.

skshetry:
@gudmundur-heimisson, does `dvc[webhdfs_kerberos]` work for you instead?

- Please take a look at https://github.com/iterative/dvc/pull/7178.

PR #6955, opened 2021-11-10
localfs: remove LocalPath

efiop:
fs.path is meant to be lightweight, and we should try to avoid using os.path.

Leftover from #6946


efiop:
For the record: test failures are unrelated.

Issue #6956, opened 2021-11-10
Possible Issue with Python 3.9.8


PythonFZ:
# Bug Report
I am sorry in advance for this incomplete Bug Report.

I am writing a package that depends on / is built for DVC.
In our CI testing we encountered an Issue with the new Python 3.9.8 release which was not present in 3.9.7.
Running the same CI with with Python 3.9.7 did not fail.

Typically I would assume that this is an issue with our software, but it occurs in a subprocess call which I think comes from DVC, because the command `Running stage <stage name>` is run by DVC.

Here is part of the log:

```
Running stage 'BasicTest':
> r u n
---------------------------- Captured stderr setup -----------------------------
/bin/sh: 1: r: not found
ERROR: failed to reproduce 'dvc.yaml': failed to run: r u n, exited with 127
```

I tried to reproduce the issue locally but with conda I am currently unable to install the Python 3.9.8 but only 3.9.7.
I wanted to let you know in advance about this and maybe it is already know.
It does not happen in 3.7, 3.8, 3.10

Here is a Link to the failing CI test https://github.com/zincware/ZnTrack/runs/4168272021?check_suite_focus=true
Here is the same CI but restricted to 3.9.7 https://github.com/zincware/ZnTrack/runs/4168558848?check_suite_focus=true

efiop:
@PythonFZ Thanks for reporting this issue! 🙏 We are already working on it.

pared:
Not sure if its related but `--quiet` stopped working too... Invesitgating...

pared:
@PythonFZ thanks for the heads up! Looking at how the [issue develops](https://bugs.python.org/issue45235) it seems that this particular fix will be [scrapped](https://bugs.python.org/msg406123). The workaround, for now, will be to downgrade the CI python version to `3.9.7` and wait for the release after revert.

efiop:
Keeping open until we decide on https://github.com/iterative/dvc/pull/6958#issuecomment-965834949

efiop:
Looks like it is stale now. Closing

PR #6958, opened 2021-11-10
argparse: dont override cmd

pared:

Recent bugfix in 3.9 release: https://bugs.python.org/issue45235
Has caused a regression. There were 2 problems:

1. we used to override `cmd` value for commands like `run`, `stage` and `exp init`. This one is on us, and while the proposed workaround does not require it, I renamed stage-related `cmd` into `command` so it does not clash with `dvc {cmd}` commands.
2. The fix has caused some other problems, for example: `dvc status -q` would not pass `-q` down to the status command, while `dvc -q status` would. I tried resolving the problem but today was not enough. Since the bugfix in question is most likely to get [scrapped](https://bugs.python.org/msg406123) I want to limit the python version to an unaffected one, and wait for another release to get pushed. ~~Since all Python versions seem to be affected, I am limiting all versions just in case.~~ - can't do that due to different versions released for different os-es. Lets limit `3.9` and see how fast the revert gets released.

The action point here is that we need to observe python releases and make sure to remove constraints once we know that the new version gets released. If for some reason the fix stays in the current format, we will need to research why the flags have not been passed down and fix that.

Closes: #6956

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -45,7 +45,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04, windows-latest, macos-latest]
-        pyv: ["3.7", "3.8", "3.9", "3.10"]
+        pyv: ["3.7", "3.8", "3.9.7", "3.10"]
         exclude:
           # no wheels for pygit2 yet
           - os: windows-latest
--- a/dvc/command/experiments.py
+++ b/dvc/command/experiments.py
@@ -814,7 +814,7 @@ class CmdExperimentsInit(CmdBase):
     def run(self):
         from dvc.command.stage import parse_cmd
 
-        cmd = parse_cmd(self.args.cmd)
+        cmd = parse_cmd(self.args.command)
         if not self.args.interactive and not cmd:
             raise InvalidArgumentError("command is not specified")
 
@@ -1416,7 +1416,7 @@ def add_parser(subparsers, parent_parser):
         formatter_class=argparse.RawDescriptionHelpFormatter,
     )
     experiments_init_parser.add_argument(
-        "cmd",
+        "command",
         nargs=argparse.REMAINDER,
         help="Command to execute.",
         metavar="command",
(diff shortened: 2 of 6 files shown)


efiop:
Great catch with -q! I guess another question is if we might be better off just putting an version check and error-ing out on 3.9.8 and not use this fix at all? Seems like you found -q/-v/--pdb not working, who knows if there is another bug hidden somewhere, so maybe better to just forbid? WDYT?

efiop (approved):
Thank you! 🙏

efiop:
From discussions in PMs: even though `cmd` is not causing problems in older versions, strictly speaking it is a bug, so it is still useful to fix it, no matter where next python version will move.

pared:
It's worth noting here that this PR does not solve all problems.

We managed to limit the failing tests and potential bugs for users using `dvc run` `dvc stage` or `dvc exp init` commands.

The change included in https://bugs.python.org/issue45235 makes using our globally defined flags (eg `-q`, `-v`, `--pdb`, `--yappi`) quite hard.
For example if we want to run `dvc status -q`, the `quiet` option will not get passed into CmdStatus. If we run `dvc -q status` it will. 

Since this fix is supposed to be scrapped I do not investigate this particular issue further. User's might face some problems with globally defined options for python `3.9.8` version.

efiop:
> Since this fix is supposed to be scrapped I do not investigate this particular issue further. User's might face some problems with globally defined options for python 3.9.8 version.

@pared Should we forbid 3.9.8 then for now?

pared:
@efiop 

After the changes to `cmd` only non-passing tests are related to `--quiet` option - I think we can live with that.

Due to fact that the future of the new implementation is highly uncertain, I would wait with adjusting to it for the python core team move. If they will keep that change, then surely we will need to adjust. But as of today, it seems the whole change will be reverted, so we should wait and see.

efiop:
> After the changes to cmd only non-passing tests are related to --quiet option - I think we can live with that.

But it also means that it doesn't work for users, right? And some other options too? If so, it might get in the way of using dvc. Though it looks like 3.9.8 broke lots of packages anyway, so it might get removed from everywhere or patched everywhere. So we still need to either fix it or at least forbid it, so users don't think that this is our bug.

Issue #6836, opened 2021-10-19
split scm from dvc


skshetry:
We want to split scm from dvc into a separate package on its own. Ideally, we'd like to have a proper and stable API. Though I'd prefer we split as-is.

I want to split the `scm` from dvc, using `filter-repo` to keep history intact.

Though I still see a few issues that we need to fix:
- [x] DVC uses logging info for UI purposes, which dvc.scm still has.
      So we may need to get rid of logging.info calls, and make scm work in terms of proper exceptions/returns. (Important)
- [x] Migrate exceptions as they subclass `DvcException` and are part of our UI framework.
- [x] scm uses Tqdm progress bars for clone operation. We need to update it to use callbacks. (Medium priority)
- [ ] Migrating tests and fixtures used. We need to copy it or figure out a way to share fixtures.

As I want to split those from DVC with history intact, we should fix 1 and 2 in DVC itself and then need to split it out.

<details>
<summary>filter-repo how-to</summary>

I used the following snippet to split scm before.
```console
cd "$(mktemp -d)"
git clone git@github.com:iterative/dvc.git .
git filter-repo --path dvc/scm --path dvc/tree/git.py --path dvc/fs/git.py --tag-rename '':'scmrepo-' --path-rename dvc:scmrepo
```

The following structure is generated:
```console
tree .
.
├── scmrepo
│   ├── fs
│   │   └── git.py
│   └── scm
│       ├── __init__.py
│       ├── base.py
│       └── git
│           ├── __init__.py
│           ├── backend
│           │   ├── __init__.py
│           │   ├── base.py
│           │   ├── dulwich.py
│           │   ├── gitpython.py
│           │   └── pygit2.py
│           ├── objects.py
│           └── stash.py
```

</details>

skshetry:
We depend on more than GitFileSystem, specifically `Git` which supports multiple backends.

skshetry:
After #6961, we won't have any `logger.info` calls in `scm`, we still have some `logger.debug` calls. Most of those are in `scm.stash` and is being used by `Experiments`. I could remove those logs or move those calls to where it's being used.

But looking at the `Experiments`, I wonder if we should split the `stash`-ing part out of it to a separate place? This is not required for this, but it is nice to have. @pmrowla, WDYT? Does extracting that out make sense? Or, do you see another way we could do this?

pmrowla:
@skshetry I think for now we can just remove the debug logging calls from the separated SCM.

[...]

PR #6961, opened 2021-11-11
scm: split tracking of to-be git-ignored/git-staged into SCMContext

skshetry:
Tracking of files to be autostaged and gitignored are now kept track in SCMContext. I was trying to find a suitable place to do that, but looking at the code, it seems that we already had a place with `@scm_context`, so it felt natural to just extend what we have been doing already. And, the naming of `scm_context` made sense for this tracking purpose. So, from now on, tracking bits goes to `SCMContext` and the other goes to `scm` (which will eventually be extracted out from dvc).

It's a 2 different API but I think tracking stuff are best kept in a separate place, and keep the `scm` lightweight and state-free.

Part of https://github.com/iterative/dvc/issues/6836.

* [x] Tests for `scm_context`.


skshetry:
@efiop @pmrowla, this should be ready for review.

pmrowla (approved):

PR #6976, opened 2021-11-12
parsing: resolve wdir for empty string as curdir

skshetry:
This PR does fix the issue within the parsing resolver
when the `wdir` is passed as an empty string. When it is passed
as such, it interprets it as a current working directory which
is a correct assumption internally.

This still does not fix the issue with fs.path.join. Original issue
with `fs.path.join` is still there.

```pycon
>>> os.path.join('', 'dvc.yaml')
'dvc.yaml'
>>> fs.path.join('', 'dvc.yaml')
'/dvc.yaml'
```

This fixes the regression introduced with 9aec8f1926b, which failed parameters/vars loading from the file
as `fs.path.join` turned the file into absolute path. This happened when `wdir` is relative path as is the case
with most stages (default `wdir` is assumed to be `""` which is also a relpath). But the tests were using `tmp_dir.fspath`
due to which we were not able to catch this. One of the test has been adjusted to use relative wdir to account for this change.

### DIFF
--- a/dvc/parsing/__init__.py
+++ b/dvc/parsing/__init__.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from collections.abc import Mapping, Sequence
 from copy import deepcopy
 from typing import (
@@ -133,17 +134,18 @@ def make_definition(
 
 
 class DataResolver:
-    def __init__(self, repo: "Repo", wdir: str, d: dict):
+    def __init__(self, repo: "Repo", wdir: str = "", d: Dict = None):
         self.fs = fs = repo.fs
-        self.wdir = wdir
+        self.wdir = wdir or os.curdir
         self.relpath = relpath(fs.path.join(self.wdir, "dvc.yaml"))
 
+        d = d or {}
         vars_ = d.get(VARS_KWD, [])
         check_interpolations(vars_, VARS_KWD, self.relpath)
         self.context: Context = Context()
 
         try:
-            args = fs, vars_, wdir  # load from `vars` section
+            args = fs, vars_, self.wdir  # load from `vars` section
             self.context.load_from_vars(*args, default=DEFAULT_PARAMS_FILE)
         except ContextError as exc:
             format_and_raise(exc, "'vars'", self.relpath)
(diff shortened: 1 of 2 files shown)


dvc/parsing/__init__.py
```diff
@@ -133,17 +134,18 @@ def make_definition(
...
         self.relpath = relpath(fs.path.join(self.wdir, "dvc.yaml"))
 
+        d = d or {}
         vars_ = d.get(VARS_KWD, [])
         check_interpolations(vars_, VARS_KWD, self.relpath)
         self.context: Context = Context()
 
         try:
-            args = fs, vars_, wdir  # load from `vars` section
+            args = fs, vars_, self.wdir  # load from `vars` section
```
efiop, line 148: If we fix fs.path.join, will these code changes be necessary? Just checking.

skshetry: I think passing an empty string in place of `os.curdir` is still a correct behaviour to me. This could happen when we use `os.path.dirname` instead which just returns the first element of path which could just be an empty string.

efiop (approved):
Will look into adjusting path.join in the followup.

skshetry:
Still need to process the question: https://github.com/iterative/dvc/pull/6976#discussion_r748393667 in more detail especially the `.parent` which changed the behaviour (which returns empty string instead of the os.curdir).

skshetry:
so fixing `fs.path.join` should fix this issue. There's another issue with `.parent()` returning an empty string instead of os.curdir, but it does not change anything here. We cannot use `path or os.curdir` everywhere, so it's better to fix it in
that single place. I was confused as before changing `fs.path.join` to `os.path.join` did not fix the issue but looking closely, there was more `fs.path.join` that I needed to replace.

Closing ...

dvc/parsing/__init__.py
efiop: Btw, looks like the difference here is that `self.wdir` and `self.relpath` are really symbolic values that are not resolved/normalized for use in `self.fs`(aka they are not `fs_path`s). Right?

skshetry: They are `fs` paths, not a symbolic values. `wdir` is used for finding a complete path to the file from where variables/params are to be loaded.`relpath` is used on error messages, so it's more of a UI thing, but it is still a `fs` path.

Issue #6694, opened 2021-09-27
push: doesn't respect HTTP remote "method" config


tonyfm15:
# Bug Report

## Description

I configured my DVC remote as a [GitLab generic package repository](https://docs.gitlab.com/ee/user/packages/generic_packages/). Part of the configuration was setting the HTTP method to `PUT`. However, `dvc push` did not work unless I went into `fsspec/implementations/http.py` and hard-coded the `method` variable in `_put_file` to `"put"`.

### Reproduce

1. Add an HTTP remote
1. `dvc remote modify <remote> method PUT`
1. `dvc add` a data file
1. Add a breakpoint or print statement to `_put_file` in `fsspec/implementations/http.py` in order to see what HTTP method it's using (via the `method` variable)
1. `dvc push`
1. Confirm that `_put_file` is using `POST`

### Expected

`dvc push` should respect the `method` config parameter and set the HTTP method to `PUT`

### Environment information

**Output of `dvc doctor`:**

```console
$ dvc doctor
DVC version: 2.7.4 (brew)
---------------------------------
Platform: Python 3.9.7 on macOS-11.5.1-x86_64-i386-64bit
Supports:
        azure (adlfs = 2021.9.1, knack = 0.8.2, azure-identity = 1.6.1),
        gdrive (pydrive2 = 1.9.4),
        gs (gcsfs = 2021.8.1),
        http (aiohttp = 3.7.4.post0, aiohttp-retry = 2.4.5),
        https (aiohttp = 3.7.4.post0, aiohttp-retry = 2.4.5),
        s3 (s3fs = 2021.8.1, boto3 = 1.17.106),
        ssh (sshfs = 2021.8.1),
        webdav (webdav4 = 0.9.2),
        webdavs (webdav4 = 0.9.2)
Cache types: reflink, hardlink, symlink
Cache directory: apfs on /dev/disk1s1s1
Caches: local
Remotes: https
Workspace directory: apfs on /dev/disk1s1s1
Repo: dvc, git
```

**Additional Information (if any):**

```console
$ dvc push --verbose
dvc push --verbose
2021-09-27 18:10:15,263 DEBUG: Preparing to transfer data from '.dvc/cache' to <redacted-url>
2021-09-27 18:10:15,263 DEBUG: Preparing to collect status from <redacted-url>
2021-09-27 18:10:15,263 DEBUG: Collecting status from <redacted-url>
[...]

PR #6981, opened 2021-11-14
push: fix http storage ignoring method setting

ditod:
Add method overriding the default put_file in http.py
where the upload method is passed to the http implementation.

Fixes: #6694

### DIFF
--- a/dvc/fs/http.py
+++ b/dvc/fs/http.py
@@ -1,11 +1,13 @@
 import threading
+from typing import Any
 
 from funcy import cached_property, memoize, wrap_with
 
 from dvc import prompt
+from dvc.progress import DEFAULT_CALLBACK
 from dvc.scheme import Schemes
 
-from .fsspec_wrapper import FSSpecWrapper, NoDirectoriesMixin
+from .fsspec_wrapper import AnyFSPath, FSSpecWrapper, NoDirectoriesMixin
 
 
 @wrap_with(threading.Lock())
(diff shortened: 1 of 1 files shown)


dvc/fs/http.py
```diff
@@ -127,3 +129,14 @@ def fs(self):
...
         return path
+
+    def put_file(
+        self,
+        from_file: AnyFSPath,
+        to_info: AnyFSPath,
+        callback: Any = DEFAULT_CALLBACK,
+        **kwargs,
+    ) -> None:
+        kwargs["method"] = self.upload_method
```
skshetry, line 140: Let's use a `setdefault` here?

```suggestion
        kwargs.setdefault("method", self.upload_method)
```

dvc/fs/http.py
```diff
@@ -127,3 +129,14 @@ def fs(self):
...
+        self.fs.put_file(from_file, to_info, callback=callback, **kwargs)
```
skshetry, line 141: We don't want to override `FSSpecWrapper` as much as possible, so let's call `super().put_file()` here, as it already calls the appropriate underlying filesystem and takes care of invalidating the cache.

skshetry (approved):

pared (approved):

pared:
Shouldn't we test that? Not sure how complicated that would be, so we might want to make a separate task out of it.

PR #6985, opened 2021-11-15
scm: hint to autostage

dberenbaum:
Adds hint (when reminding to track files) to set core.autostage.

Related to #6929 and https://github.com/iterative/dvc.org/issues/2969.

Thoughts on this?

### DIFF
--- a/dvc/scm/git/__init__.py
+++ b/dvc/scm/git/__init__.py
@@ -258,7 +258,11 @@ class Git(Base):
             "\n"
             "To track the changes with git, run:\n"
             "\n"
-            "\tgit add {files}".format(files=files)
+            "\tgit add {files}\n"
+            "\n"
+            "To enable auto staging, run:\n"
+            "\n"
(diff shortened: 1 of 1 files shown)


daavoo (commented):
Like it. 

Minor consideration would be to *only* show the extra lines if the config option is *not* enabled

dberenbaum:
> Like it.
> 
> Minor consideration would be to _only_ show the extra lines if the config option is _not_ enabled

If `core.autostage` is enabled, this entire hint is hidden:

```
$ dvc config core.autostage true
$ dvc add foo
100% Adding...|████████████████████████████████████████████████|1/1 [00:00, 102.26file/s]
$ dvc config core.autostage false
100% Adding...|█████████████████████████████████████████████████|1/1 [00:00, 94.89file/s]

        git add foo.dvc

To enable auto staging, run:

        dvc config core.autostage true
```

daavoo (approved):

karajan1001 (approved):

PR #6993, opened 2021-11-17
ui: Add `open_browser`.

daavoo:
Use `webbrowser.open` and resolve relpath for non WSL plattforms.

For reusing in #6913 and #6933

### DIFF
--- a/dvc/ui/__init__.py
+++ b/dvc/ui/__init__.py
@@ -20,6 +20,7 @@ if TYPE_CHECKING:
     from rich.text import Text as RichText
 
     from dvc.progress import Tqdm
+    from dvc.types import StrPath
     from dvc.ui.table import Headers, Styles, TableData
 
 
@@ -259,6 +260,28 @@ class Console:
         import sys
 
         return sys.stdout.isatty()
+
+    def open_browser(self, file: "StrPath") -> int:
+        import webbrowser
+        from pathlib import Path
+        from platform import uname
+
+        from dvc.utils import relpath
+
+        path = Path(file).resolve()
+        url = (
+            relpath(path) if "Microsoft" in uname().release else path.as_uri()
+        )
+
+        opened = webbrowser.open(url)
+
+        if not opened:
+            ui.error_write(
+                f"Failed to open {url}. " "Please try opening it manually."
+            )
+            return 1
+
+        return 0
 
 
 ui = Console()
--- a/dvc/command/plots.py
+++ b/dvc/command/plots.py
@@ -62,23 +62,11 @@ class CmdPlots(CmdBase):
                 html_template_path=self.args.html_template,
             )
 
-            assert index_path.is_absolute()
-            url = index_path.as_uri()
-            ui.write(url)
+            ui.write(index_path.as_uri())
 
             if self.args.open:
-                import webbrowser
-                from platform import uname
+                return ui.open_browser(index_path)
 
-                if "Microsoft" in uname().release:
-                    url = Path(rel) / "index.html"
-
-                opened = webbrowser.open(url)
-                if not opened:
-                    ui.error_write(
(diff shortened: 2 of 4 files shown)


dvc/ui/__init__.py
```diff
@@ -260,6 +263,23 @@ def isatty(self) -> bool:
 
         return sys.stdout.isatty()
 
+    def open_browser(self, rel_parent_path, file: str = "index.html"):
```
pared, line 266: Not sure we should provide `rel_parent_path` and `file` separately:
The assumption that our webpages are created as `{path}/index.html` is a concept from our visualization code, UI should not be aware of this. I think we would be fine just providing a `path` (or rather `relpath` if we want to nudge user in proper direction.)

pared (approved):

dvc/ui/__init__.py
skshetry: I'll add a few requirements here to the `path`/`file` parameter here: 
1. It should accept both `os.Pathlike` and `str`.
2. The path could be either relative, absolute, or a non-normalized path (eg: `src/../index.html`) that we pass straight through CLI arguments.

daavoo: I guess that `open_browser` could just accept any kind of path and change the logic so that, for the WSL case, the given path is converted to relative.

daavoo: >     1. It should accept both `os.Pathlike` and `str`.

👍 

>     2. The path could be either relative, absolute, or a non-normalized path (eg: `src/../index.html`) that we pass straight through CLI arguments. The reason for the non-normalized path is that
>        `Path("/src/../index.html").as_uri()` will just create file uri like `file:///src/../index.html`.

Do you mean that we should normalize the path inside `open_browser`?

skshetry: > Do you mean that we should normalize the path inside open_browser?

yes, `Path(path).resolve()` should probably work.

dvc/ui/__init__.py
```diff
@@ -1,4 +1,7 @@
+import webbrowser
 from contextlib import contextmanager
+from pathlib import Path
+from platform import uname
```
skshetry, line 4: The `dvc.ui` module is meant to be imported globally as it's a singleton, so a great care has been taken to have almost no imports as possible. Please import this inside ' open_browser`.

dvc/ui/__init__.py
```diff
@@ -260,6 +262,26 @@ def isatty(self) -> bool:
 
         return sys.stdout.isatty()
 
+    def open_browser(self, file: "PathLike"):
+        import webbrowser
+        from pathlib import Path
+        from platform import uname
+
+        if "Microsoft" in uname().release:
+            url = str(Path(file).relative_to(Path.cwd()))
```
skshetry, line 271: Let's `resolve` both of them anyway. Also, for `relative_to` to work, [the `cwd` needs to be parent path](https://docs.python.org/3/library/pathlib.html#id6), so we should be using `dvc.utils.relpath`/`os.path.relpath` here.

```suggestion
            path = Path(file).resolve()
            url = relpath(path) if "Microsoft" in uname().release else path.uri()
```

dvc/ui/__init__.py
```diff
@@ -260,6 +262,26 @@ def isatty(self) -> bool:
 
         return sys.stdout.isatty()
 
+    def open_browser(self, file: "PathLike"):
```
skshetry, line 265: We should also support `str`, so it should really be `str | PathLike`. Let's use [`dvc.types.StrPath`](https://github.com/iterative/dvc/blob/a70b3087b4a2a8d604781a9d4ca34486ff90ee16/dvc/types.py#L6) which does that already.

```suggestion
    def open_browser(self, file: "StrPath"):
```

pared (approved):

PR #7006, opened 2021-11-19
exp init: timeout prompts' test

skshetry:
The `exp init` tests may just keep waiting on the prompts blocking tests.
The blocking of prompts is likely a failure in tests setup or a bug
in the implementation. It's hard to find out these failures or avoid it.

If they take longer than 1 second, they most likely have failed as
most of the tests take 0.25 seconds at max in Windows. So, using `pytest-timeout` is the
easiest approach here.


skshetry:
@Mergifyio rebase

efiop (approved):
Thank you!

Issue #6943, opened 2021-11-09
plots: extract data processing out of renderers


pared:
We need to extract conversion of `dvc.plots.show` data to vega-elgible datapoints out of `VegaRenderer`.
There are two main resons for that:
1. So that studio can use the same codebase for datapoints conversion as we do
2. So that we can remove dependency from Repo class from Renderers: https://github.com/iterative/dvc/blob/4175f9ffa4948157436283bef11153a7233e90ca/dvc/render/vega.py#L113 - template resolving and plot properties resolving is part of DVC job, Renderers should focus just on creating an image.

Related: #6931 #6923
cc @daavoo

Issue #6051, opened 2021-05-24
exp show: branch names are toggling randomly


aguschin:
# Bug Report

## Description

I cloned repo get-started-experiments and created new branch. Now `dvc exp show` shows me different  branch names from time to time. It also happens if you move back to `main` branch.

<img width="1160" alt="image" src="https://user-images.githubusercontent.com/6797716/119347212-e3a71080-bca3-11eb-9bfa-ccfd663a5718.png">

### Reproduce

1. git clone https://github.com/iterative/get-started-experiments 
2. cd get-started-experiments
3. dvc pull
4. git checkout -b other-branch
5. dvc exp show --no-pager

Run last command few times

### Environment information

**Output of `dvc doctor`:**

```console
$ dvc doctor
DVC version: 2.1.0 (brew)
---------------------------------
Platform: Python 3.9.5 on macOS-10.16-x86_64-i386-64bit
Supports: azure, gdrive, gs, http, https, s3, ssh, oss, webdav, webdavs
Cache types: reflink, hardlink, symlink
Cache directory: apfs on /dev/disk1s5s1
Caches: local
Remotes: https
Workspace directory: apfs on /dev/disk1s5s1
Repo: dvc, git
```

**Additional Information (if any):**

Tried the same with the latest dev version 2.1.0+9dda6b, the bug persists

casperdcl:
Ideally all aliases (refs with the same `HEAD`) should be shown, or at least the same one consistently (e.g. first alphabetically. I don't think Git keep a record of when an alias branch was created so no way to take last created, I think).

pmrowla:
`exp show` is being changed to always show git SHA's by default, along with a single branch or tag name (w/tag's given higher priority) for that SHA. The SHA will remain unchanged, and ideally we should consistently choose a ref, but I don't think this is high priority right now.

dberenbaum:
@pmrowla Isn't `exp show` limited to the experiments derived from the current HEAD by default? If so, would it make sense to at least prioritize the current branch over others?

pmrowla:
@dberenbaum sure, that's one way we could make it more consistent, at least for the case where HEAD is actually a branch, and there are no tags that also point to HEAD.

karajan1001:
@aguschin Does it still happen on your computer? I tried on my one can't reproduce it for now.

[...]

PR #7007, opened 2021-11-19
render: vega: data processing outside of renderers

pared:
Fixes: #6943

@Suor I am addressing #6943 here which is a result of our [discussion in iterative/viewer#2657](https://github.com/iterative/viewer/issues/2657#issuecomment-939674364)

`dvc/render/utils.py::to_datapoints` accepts `data` and `props`, so you should be able to provide your own properties, which should allow to address the disrepancies between `dvc` and studio plot properties resolving.

### DIFF
--- a/dvc/command/live.py
+++ b/dvc/command/live.py
@@ -3,7 +3,6 @@ from pathlib import Path
 
 from dvc.command import completion
 from dvc.command.base import CmdBase, fix_subparsers
-from dvc.render.utils import match_renderers
 from dvc.ui import ui
 
 
@@ -11,11 +10,12 @@ class CmdLive(CmdBase):
     UNINITIALIZED = True
 
     def _run(self, target, revs=None):
+        from dvc.render.utils import match_renderers, render
+
         metrics, plots = self.repo.live.show(target=target, revs=revs)
 
         if plots:
             html_path = Path.cwd() / (self.args.target + "_html")
-            from dvc.render.utils import render
 
             renderers = match_renderers(plots, self.repo.plots.templates)
             index_path = render(self.repo, renderers, metrics, html_path)
--- a/dvc/render/data.py
+++ b/dvc/render/data.py
@@ -1,0 +1,194 @@
+from copy import deepcopy
+from functools import partial
+from typing import Dict, List, Optional, Set, Union
+
+from funcy import first, project
+
+from dvc.exceptions import DvcException
+from dvc.render.base import INDEX_FIELD, REVISION_FIELD
+
+
+class FieldsNotFoundError(DvcException):
+    def __init__(self, expected_fields, found_fields):
+        expected_str = ", ".join(expected_fields)
+        found_str = ", ".join(found_fields)
+        super().__init__(
+            f"Could not find all provided fields ('{expected_str}') "
+            f"in data fields ('{found_str}')."
+        )
+
+
+class PlotDataStructureError(DvcException):
+    def __init__(self):
+        super().__init__(
+            "Plot data extraction failed. Please see "
+            "https://man.dvc.org/plots for supported data formats."
+        )
+
+
+def _filter_fields(datapoints: List[Dict], fields: Set) -> List[Dict]:
+    if not fields:
+        return datapoints
+    assert isinstance(fields, set)
+
+    new_data = []
+    for data_point in datapoints:
+        keys = set(data_point.keys())
+        if not fields <= keys:
+            raise FieldsNotFoundError(fields, keys)
+
+        new_data.append(project(data_point, fields))
+
+    return new_data
+
+
+def _lists(dictionary: Dict):
+    for _, value in dictionary.items():
+        if isinstance(value, dict):
+            yield from _lists(value)
+        elif isinstance(value, list):
+            yield value
+
+
+def _find_first_list(data: Union[Dict, List], fields: Set) -> List[Dict]:
+    fields = fields or set()
+
+    if not isinstance(data, dict):
+        return data
+
+    for lst in _lists(data):
+        if (
+            all(isinstance(dp, dict) for dp in lst)
+            # if fields is empty, it will match any set
+            and set(first(lst).keys()) & fields == fields
+        ):
+            return lst
+
+    raise PlotDataStructureError()
+
+
+def _append_index(datapoints: List[Dict]) -> List[Dict]:
+    if INDEX_FIELD in first(datapoints).keys():
+        return datapoints
+
+    for index, data_point in enumerate(datapoints):
+        data_point[INDEX_FIELD] = index
+    return datapoints
+
+
+class Converter:
+    """
+    Class that takes care of converting unspecified data blob
+    (Dict or List[Dict]) into datapoints (List[Dict]).
+    If some properties that are required by Template class are missing
+    ('x', 'y') it will attempt to fill in the blanks.
+    """
+
+    @staticmethod
+    def update(datapoints: List[Dict], update_dict: Dict):
+        for data_point in datapoints:
+            data_point.update(update_dict)
+        return datapoints
+
+    def __init__(self, plot_properties: Optional[Dict] = None):
+        plot_properties = plot_properties or {}
+        self.props = deepcopy(plot_properties)
+        self.inferred_props: Dict = {}
+
+        self.steps = []
+
+        self._infer_x()
+        self._infer_fields()
+
+        self.steps.append(
+            (
+                "find_data",
+                partial(
+                    _find_first_list,
+                    fields=self.inferred_props.get("fields", set())
+                    - {INDEX_FIELD},
+                ),
+            )
+        )
+
+        if not self.props.get("x", None):
+            self.steps.append(("append_index", partial(_append_index)))
+
+        self.steps.append(
+            (
+                "filter_fields",
+                partial(
+                    _filter_fields,
+                    fields=self.inferred_props.get("fields", set()),
+                ),
+            )
+        )
+
+    def _infer_x(self):
+        if not self.props.get("x", None):
+            self.inferred_props["x"] = INDEX_FIELD
+
+    def skip_step(self, name: str):
+        self.steps = [(_name, fn) for _name, fn in self.steps if _name != name]
+
+    def _infer_fields(self):
+        fields = self.props.get("fields", set())
+        if fields:
+            fields = {
+                *fields,
+                self.props.get("x", None),
+                self.props.get("y", None),
+                self.inferred_props.get("x", None),
+            } - {None}
+            self.inferred_props["fields"] = fields
+
+    def _infer_y(self, datapoints: List[Dict]):
+        if "y" not in self.props:
+            data_fields = list(first(datapoints))
+            skip = (
+                REVISION_FIELD,
+                self.props.get("x", None) or self.inferred_props.get("x"),
+            )
+            inferred_y = first(
+                f for f in reversed(data_fields) if f not in skip
+            )
+            if "y" in self.inferred_props:
+                previous_y = self.inferred_props["y"]
+                if previous_y != inferred_y:
+                    raise DvcException(
+                        f"Inferred y ('{inferred_y}' value does not match"
+                        f"previously matched one ('f{previous_y}')."
+                    )
+            else:
+                self.inferred_props["y"] = inferred_y
+
+    def convert(self, data):
+        """
+        Convert the data. Fill necessary fields ('x', 'y') and return both
+        generated datapoints and updated properties.
+        """
+        processed = deepcopy(data)
+
+        for _, step in self.steps:
+            processed = step(processed)
+
+        self._infer_y(processed)
+
+        return processed, {**self.props, **self.inferred_props}
+
+
+def to_datapoints(data: Dict, props: Dict):
+    converter = Converter(props)
+
+    datapoints = []
+    for revision, rev_data in data.items():
+        for _, file_data in rev_data.get("data", {}).items():
+            if "data" in file_data:
+                processed, final_props = converter.convert(
+                    file_data.get("data")
+                )
+
+                Converter.update(processed, {REVISION_FIELD: revision})
+
+                datapoints.extend(processed)
+    return datapoints, final_props
--- a/dvc/render/utils.py
+++ b/dvc/render/utils.py
@@ -21,14 +21,31 @@ def group_by_filename(plots_data: Dict) -> List[Dict]:
     return grouped
 
 
+def squash_plots_properties(data: Dict) -> Dict:
+    resolved: Dict[str, str] = {}
+    for rev_data in data.values():
+        for file_data in rev_data.get("data", {}).values():
+            props = file_data.get("props", {})
+            resolved = {**resolved, **props}
+    return resolved
+
+
 def match_renderers(plots_data, templates):
     from dvc.render import RENDERERS
 
     renderers = []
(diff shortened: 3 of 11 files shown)


tests/unit/render/test_vega.py
```diff
@@ -284,51 +281,19 @@ def test_metric_missing(tmp_dir, scm, dvc, caplog):
     assert first(plot_content["layer"])["encoding"]["y"]["field"] == "y"
 
 
-def test_custom_template(tmp_dir, scm, dvc, custom_template):
-    metric = [{"a": 1, "b": 2}, {"a": 2, "b": 3}]
```
pared, line 288: test_templates::test_load_template is responsible for testing custom template load now

tests/unit/render/test_vega.py
```diff
@@ -284,51 +281,19 @@ def test_metric_missing(tmp_dir, scm, dvc, caplog):
...
-    plot_content = json.loads(plot_string)
-    assert plot_content["data"]["values"] == [
-        {"a": 1, "b": 2, REVISION_FIELD: "workspace"},
-        {"a": 2, "b": 3, REVISION_FIELD: "workspace"},
-    ]
-    assert plot_content["encoding"]["x"]["field"] == "a"
-    assert plot_content["encoding"]["y"]["field"] == "b"
-
-
-def test_raise_on_no_template(tmp_dir, dvc):
```
pared, line 305: test_templates::test_raise_on_no_template

tests/unit/render/test_vega.py
```diff
@@ -348,37 +317,15 @@ def test_plot_choose_columns(tmp_dir, scm, dvc, custom_template):
     assert plot_content["encoding"]["y"]["field"] == "c"
 
 
-def test_plot_default_choose_column(tmp_dir, scm, dvc):
```
pared, line 351: test_plot_choose_columns and test_templates::test_load_templates test the same

tests/unit/render/test_vega.py
```diff
@@ -448,29 +395,3 @@ def test_find_vega(tmp_dir, dvc):
...
-            os.path.join("subdir", "template.json"),
-        ),
-        (
-            os.path.join(".dvc", "plots", "subdir", "template.json"),
-            os.path.join("subdir", "template"),
-        ),
-        ("template.json", "template.json"),
-    ],
-)
-def test_should_resolve_template(tmp_dir, dvc, template_path, target_name):
```
pared, line 469: moved to test_templates

dvc/render/utils.py
```diff
@@ -67,3 +100,94 @@ def render(
...
+    return datapoints
+
+
+def _append_revision(datapoints: List[Dict], revision) -> List[Dict]:
+    for data_point in datapoints:
+        data_point[REVISION_FIELD] = revision
+    return datapoints
+
+
+def to_datapoints(data: Dict, props: Dict):
```
Suor, line 171: Is this the blessed preprocessor util, which one should import and use? It can't be used by Studio at least in this form:
- we need to skip `_append_revision()`
- need to fill in props `x` and `y` based on data

Suor: Everything is so low-level and not really defined that I doubt it could be really reused.

pared: Ok, Ill modify it, how should

> need to fill in props x and y based on data

work?

dberenbaum: @Suor Can you clarify what you DO want this preprocessing util to do? I might be missing the context from previous conversations, but it's hard for me to tell from looking through the GitHub issues.

@pared Can we think about whether this would be useful when we start developing a public api to access plots data? For example, let's say I want to get all the data from one or multiple experiments to do my own analysis/visualization.

pared: @dberenbaum this is already in place: current `Repo.plots.show` will return mapping containing chosen targets from particular revisions. The change that we are doing here is supposed to convert this data returned from `Repo.plots.show` to datapoints which are a structure required by vega plots templates.

pared: @Suor sorry, should have checked `parse_data` before asking for review, I will look into moving the `x` and `y` resolution too. I will also think about how to alternate the structure so that those methods are not some randoms imported from `utils`.

pared:
@Suor introduced some changes.
Now the parse data should be looking something like: 

```
def parse_data(datafile, datablob, props, append_index=False):
    """This is basically VegaRenderer.get_vega() but for a single thing
       and with some ops and checks skipped"""
    from dvc.render.data import Converter

    # Copy it to not modify a passed value
    props = props.copy()
    props["append_index"] = append_index

    # Legacy unparsed data from database need to be parsed before preprocessing
    if isinstance(datablob, str):
        data = _parse_blob(datafile, datablob, props)
    else:
        data = datablob

    c = Converter(props)
    datapoints, all_props = c.convert(data)

    
    return all_props, datapoints 
```

Main changes:
- provided props are no longer modified. In the end of `convert` we will receive new dict that takes into account all the automagical behaviors
- `append_revision` is now a static method of `Converter` class to be used separately
- converter can now fail if it won't be provided with `y` in `props` and will detect that different field is used as y in different data parsed
- one can explicitly state with `props` `append_index` = False

Not sure if it's working thought, because I was unable to run the viewer tests as of today. Will try to achieve that tomorrow.

Suor:
That comes very close. This code should work:

```python
def parse_data(datafile, datablob, props, append_index=False):
    from dvc.render.data import Converter

    # `append_index` param means whether we should perform append_index step,
    # it should not affect props. Those will be passed to FE, which will do append_index itself.
    # The reason is not to store meaningless `step: i` in static data.
    # 
    # # Copy it to not modify a passed value
    # props = props.copy()
    # props["append_index"] = append_index

    # This will be dropped soon
    # Legacy unparsed data from database need to be parsed before preprocessing
    if isinstance(datablob, str):
        data = _parse_blob(datafile, datablob, props)
    else:
        data = datablob

    c = Converter(props)
    if not append_index:  # This will be always False soon
        del c.steps[1]  # Drop append_index step, fragile
    datapoints, all_props = c.convert(data)

    return all_props, datapoints  # Can revert on a caller side to make it simpler later
```

So if you clean it up it's just:

```python
def parse_data(data, props):
    from dvc.render.data import Converter

    c = Converter(props)
    del c.steps[1]  # Drop append_index step, fragile
    return c.convert(data)
```

As you can see the only lacking functionality is official steps manipulation. Any of the below will work:

```python
c = Converter(props, skip_steps=["append_index"])
# or 
c = Converter(props)
c.skip_step("append_index")
# or
c = Converter(props)
c.steps = [step for step in c.steps if step.name != "append_index"]
```

Whichever is easier to implement or makes more sense to you. BTW, having that you might integrate `append_revision` back as one of the steps if you want.

dvc/render/data.py
```diff
@@ -0,0 +1,199 @@
...
+            if keys & fields != fields:
+                raise FieldsNotFoundError(fields, keys)
+
+            to_del = keys - fields
+            for key in to_del:
+                del new_dp[key]
+            new_data.append(new_dp)
+        return new_data
+
+    return process
```
Suor, line 51: As an alternative can keep flat functions and use `partial()` when making steps. Can also make them named with something like:
```python
def named_partial(name, func, *args, **kwargs):
    p = partial(func, *args, **kwargs)
    p.name = name
    return p

c.steps = [
    named_partial("find_data", _find_data, fields=self.props.get("fields"),
    ...
]
```
Or put names outside of callable objects:
```python
c.steps = [
    ("find_data", partial(_find_data, fields=self.props.get("fields")),
    ...
]

```

dvc/render/data.py
```diff
@@ -0,0 +1,199 @@
...
+        self.props = plot_properties.copy()
+        self.inferred_props = {}
+
+        if self.props.get("append_index", True):
+            self._infer_x()
+
+        self._infer_fields()
+
+        self.steps = []
+        self.steps.append(_find_data(self.props.get("fields", None)))
```
Suor, line 118: Should this the `self.infered_props.get('fields')`?

dvc/render/data.py
pared: Great idea. It would also make skip functionality very easy.

pared: Implemented names outside of partial, to keep mypy happy.

pared:
@Suor added `skip_step`. I am keeping `append_revision`, as more generic `update` for datapoints update:
I am doing so that we don't have to pass `revision` into `convert`.

dvc/render/data.py
```diff
@@ -0,0 +1,213 @@
...
+
+def _lists(dictionary: Dict):
+    for _, value in dictionary.items():
+        if isinstance(value, dict):
+            yield from _lists(value)
+        elif isinstance(value, list):
+            yield value
+
+
+def _find_data(data: Union[Dict, List], fields: Set) -> List[Dict]:
```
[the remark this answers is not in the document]
pared: I am not sure how that would look like? The `if` check that is below, besides verifying that the fields match also verifies that all the elements of list are of `Dict` type. I am not sure it makes sense to duplicate this logic. I agree that this comment is less informative than it used to be, I will modify it and put it near the check.

pared:
@Mergifyio rebase

dvc/render/data.py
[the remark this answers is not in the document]
pared: @karajan1001 I just reread this conversation and I think understood your comment wrongly. I think it makes perfect sense to rename `_find_data`.

dvc/render/data.py
```diff
@@ -0,0 +1,211 @@
...
+
+def _lists(dictionary: Dict):
+    for _, value in dictionary.items():
+        if isinstance(value, dict):
+            yield from _lists(value)
+        elif isinstance(value, list):
+            yield value
+
+
+def _find_first_list(data: Union[Dict, List], fields: Set) -> List[Dict]:
```
karajan1001, line 58: I'm sorry I'm not quite familiar with the vega render but looks like this function is not deterministic, without
 any problem?

dvc/render/data.py
```diff
@@ -0,0 +1,211 @@
...
+        else:
+            self.inferred_props["x"] = INDEX_FIELD
+
+    def skip_step(self, name: str):
+        self.skip_steps.add(name)
+
+    def _infer_fields(self):
+        fields = self.props.get("fields", set())
+        if fields:
+            fields = {
```
karajan1001, line 142: How about 
```python
fields += self.props.get("x", Set()) + self.props.get("y", Set()) + self. inferred_props.get("x", Set())
```

dvc/render/data.py
pared: What do you mean by not deterministic? Are you talking about `_lists` sourcing from `Dict`? AFAIK, in python3 the Dicts are ordered.

dvc/render/data.py
```diff
@@ -0,0 +1,211 @@
...
+            )
+        )
+
+    def _infer_x(self):
+        if self.props.get("x", None):
+            self.skip_step("append_index")
+        else:
+            self.inferred_props["x"] = INDEX_FIELD
+
+    def skip_step(self, name: str):
```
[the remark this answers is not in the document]
pared: We do, `viewer` needs it.

dvc/render/data.py
karajan1001: It will give a different result even if we put the same arguments to it?

dvc/render/data.py
pared: `x` and `y` wil be strings, so we cannot use union to join the to fields

dvc/render/data.py
pared: I am not sure what is the situation when it would return different results? If you are talking about the dictionary and the `_lists` call, python dicts maintain the [order](https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6). We had few discussions (eg https://github.com/iterative/dvc/issues/5009) and AFAIR we decided to not put too much effort in asserting the order of dict "just in case", but react if default behavior will change.

dvc/render/data.py
```diff
@@ -0,0 +1,211 @@
...
+
+    new_data = []
+    for data_point in datapoints:
+        keys = set(data_point.keys())
+        if keys & fields != fields:
+            raise FieldsNotFoundError(fields, keys)
+
+        new_data.append(
+            {key: value for key, value in data_point.items() if key in fields}
+        )
```
Suor, line 45: May use `project()` from `funcy`:
```suggestion
        new_data.append(project(data_point, fields))
```

dvc/render/data.py
```diff
@@ -0,0 +1,211 @@
...
+
+def _filter_fields(datapoints: List[Dict], fields: Set) -> List[Dict]:
+    if not fields:
+        return datapoints
+    assert isinstance(fields, set)
+
+    new_data = []
+    for data_point in datapoints:
+        keys = set(data_point.keys())
+        if keys & fields != fields:
```
Suor, line 40:
```suggestion
        if not fields <= keys:
```
Only one set op and not need to create an intermediate set.

dvc/render/data.py
```diff
@@ -0,0 +1,211 @@
...
+        )
+
+    def _infer_x(self):
+        if self.props.get("x", None):
+            self.skip_step("append_index")
+        else:
+            self.inferred_props["x"] = INDEX_FIELD
+
+    def skip_step(self, name: str):
+        self.skip_steps.add(name)
```
Suor, line 137: Can simply drop from planned steps immediately
```suggestion
        self.steps = [(_name, fn) for _name, fn in self.steps if _name != name]
```

Suor (approved):
Looks good, thanks @pared. Some cosmetic suggestions below. Also do we need to add a test imitating Studio `parse_data()` scenario to not break things accidentally?

pared:
@Suor introduced the changes you suggested. Regarding the test for `viewer` use case, it seems to me that 
`test_convert_skip_step` covers that. Did you think about something else? Or should I stress the fact that this functionality is external so that nobody removes it in the future?

@karajan1001 I think we should leave the dict loading as it is. In  #5009 we agreed to leave that be unless core `dict` behavior changes. AFAIK https://github.com/iterative/dvc/issues/6051 is caused by how the git is working and not the order of parsing JSON.

dvc/render/data.py
```diff
@@ -0,0 +1,204 @@
...
+
+from funcy import first, project, reraise
+
+from dvc.exceptions import DvcException
+from dvc.render.base import INDEX_FIELD, REVISION_FIELD
+
+
+class FieldsNotFoundError(DvcException):
+    def __init__(self, expected_fields, found_fields):
+        self.expected_fields = expected_fields
```
karajan1001, line 13: Excuse me, what does these two member variable do? And what is the status of this PR?

pared: I was using them to reraise in some previous version of this change. Removing.
As of status, I am currently trying to check why benchmarks keep failing. Besides that I believe its ready to merge.

dvc/command/live.py
```diff
@@ -3,19 +3,19 @@
 
 from dvc.command import completion
 from dvc.command.base import CmdBase, fix_subparsers
-from dvc.render.utils import match_renderers
```
pared, line 6: This import was a cause for failing `help` benchmarks

dvc/render/data.py
pared: The PR is ready for review. @karajan1001 ping me if you want to review it once again. If not, I will squash the minor commits and merge.

karajan1001 (approved):

PR #7008, opened 2021-11-19
exp show: Use tag name for `all_tags`.

### LINKED ISSUE #6383: `dvc exp show --all-tags`: doesn't show the names of annotated tags.
# Bug Report

## Description

`dvc exp show --all-tags` lists commit id's which correspond to tags, but the tag names are not shown when they are created with `git tag -a`. 

image

image

image

### Reproduce

```
git clone https://github.com/iterative/get-started-experiments
dvc exp show --all-tags
```

### Expected

The `Experiments` column in the table should show tag names. 

### Environment information

Recent Ubuntu 20.04 WSL on

Edition	Windows 11 Pro
Version	21H2
Installed on	‎7/‎28/‎2021
OS build	22000.100
Experience	Windows Feature Experience Pack 421.18901.0.3

image

daavoo:
Close #6383

### DIFF
--- a/dvc/scm/git/backend/dulwich/__init__.py
+++ b/dvc/scm/git/backend/dulwich/__init__.py
@@ -326,6 +326,7 @@ class DulwichBackend(BaseGitBackend):  # pylint:disable=abstract-method
             raise SCMError(f"Failed to set '{name}'")
 
     def get_ref(self, name, follow: bool = True) -> Optional[str]:
+        from dulwich.objects import Tag
         from dulwich.refs import parse_symref_value
 
         name_b = os.fsencode(name)
@@ -342,6 +343,8 @@ class DulwichBackend(BaseGitBackend):  # pylint:disable=abstract-method
             except ValueError:
                 pass
         if ref:
+            if isinstance(self.repo[ref], Tag):
+                ref = self.repo.get_peeled(name_b)
             return os.fsdecode(ref)
         return None
 
--- a/tests/func/experiments/test_show.py
+++ b/tests/func/experiments/test_show.py
@@ -348,6 +348,38 @@ def test_show_multiple_commits(tmp_dir, scm, dvc, exp_stage):
     assert set(results.keys()) == expected
 
 
+def test_show_tags(tmp_dir, scm, dvc):
+    tmp_dir.scm_gen("file", "1", "commit")
+    scm.tag("light")
+    rev = scm.get_rev()
+
+    show = dvc.experiments.show(all_tags=True)
+    assert show[rev]["baseline"]["data"]["name"] == "light"
+
+    tmp_dir.scm_gen("file", "2", "commit")
+    scm.tag(["-a", "annot", "-m", ""])
+    rev = scm.get_rev()
+
+    show = dvc.experiments.show(all_tags=True)
+    assert show[rev]["baseline"]["data"]["name"] == "annot"
+
+    tmp_dir.scm_gen("file", "3", "commit")
+    scm.tag(["-a", "annotated", "-m", ""])
+    scm.tag(["-a", "pointer", "-m", "", "annotated"])
+    rev = scm.get_rev()
+
+    show = dvc.experiments.show(all_tags=True)
+    assert show[rev]["baseline"]["data"]["name"] == "pointer"
+
+    tmp_dir.scm_gen("file", "3", "commit")
+    scm.tag(["-a", "foo", "-m", ""])
+    scm.tag(["-a", "stacked", "-m", ""])
+    rev = scm.get_rev()
+
+    show = dvc.experiments.show(all_tags=True)
+    assert show[rev]["baseline"]["data"]["name"] == "foo"
+
+
 def test_show_sort(tmp_dir, scm, dvc, exp_stage, caplog):
     with caplog.at_level(logging.ERROR):
         assert main(["exp", "show", "--no-pager", "--sort-by=bar"]) != 0


dvc/repo/experiments/show.py
```diff
@@ -136,13 +136,14 @@ def show(
             all_branches=all_branches,
             all_tags=all_tags,
             all_commits=all_commits,
-            sha_only=True,
+            sha_only=False,
```
pmrowla, line 139: Functionality wise, this just reverts to using command line git (via git-python) to get the tag names, which isn't really a fix (since we eventually want to get rid of git-python entirely).

pmrowla (changes requested):
So the original issue isn't with tags in packed refs. The problem is with annotated (signed) tags. Normal/lightweight (unsigned) tags are already handled properly, even when they are in packed refs.

Lightweight tags are just regular git refs (where the ref points directly to a commit object's SHA).

Annotated tags are different in that the ref points to a (annotated) tag object SHA, and not a commit SHA. To go from the annotated tag name to the commit SHA, you have to actually read the tag object first to get the dereferenced commit SHA (this is generally referred to as `peeling` the tag object in git/dulwich/pygit/etc docs+code).

(see https://git-scm.com/docs/git-tag)

The underlying issue in DVC is that the current dulwich implementation for `git.describe` doesn't resolve annotated (signed) tags properly.
`https://github.com/iterative/dvc/blob/cb941a836d77c845f612be29e27ae80f7a2a126c/dvc/scm/git/backend/dulwich/__init__.py#L599`

I haven't investigated this issue deeply, but if I had to guess I would think that the actual bug is probably that dulwich `get_ref(annotated_tag)` is returning the SHA for the tag object itself, instead of returning the peeled commit SHA.

https://github.com/iterative/dvc/blob/cb941a836d77c845f612be29e27ae80f7a2a126c/dvc/scm/git/backend/dulwich/__init__.py#L613

(Note that `follow` is unrelated to tags - it's for whether or not to follow symbolic refs. For annotated tags `get_ref` should always return the peeled commit SHA.)

pmrowla:
Also, if you are trying to test against packed refs, you can just make some tags and then force git to pack everything in your repo with
```
git pack-refs --all
```

Don't do this in non-testing repos as it will pack all of your active branch refs (which will probably end up affecting performance of normal git command usage)

https://git-scm.com/docs/git-pack-refs

daavoo:
> 

Thanks! Indeed, I used that for testing on local repo but was asking more in the lines of: how to properly do this inside one of our tests? I didn't find a way to do it from Python (apart from subprocess call)

dvc/repo/experiments/show.py
daavoo: Totally right. 

The snippet you posted when guessing above is exactly where I ended up reaching debugging and it is indeed the source of the problem.

I was a little "afraid" of touching the `scm` module for fixing an (at least initially) `exp show` issue, especially given that `get_ref` is used widely across the experiments functionality (beyond `show`) and that I didn't know what `packed-refs` were or the difference between lightweight and annotated tags 😅 

So I added a "fix" that worked on the `experiments/show` level.

I will try to fix it on `scm` level without breaking anything

dvc/scm/git/backend/dulwich/__init__.py
```diff
@@ -342,6 +343,8 @@ def get_ref(self, name, follow: bool = True) -> Optional[str]:
             except ValueError:
                 pass
         if ref:
+            if isinstance(self.repo[ref], Tag):
+                ref = self.repo[ref].object[1]
```
daavoo, line 347: Does this makes sense @pmrowla ?

daavoo:
> Also, if you are trying to test against packed refs, you can just make some tags and then force git to pack everything in your repo with
> 
> ```
> git pack-refs --all
> ```
> 
> Don't do this in non-testing repos as it will pack all of your active branch refs (which will probably end up affecting performance of normal git command usage)
> 
> https://git-scm.com/docs/git-pack-refs



> > Also, if you are trying to test against packed refs, you can just make some tags and then force git to pack everything in your repo with
> > ```
> > git pack-refs --all
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Don't do this in non-testing repos as it will pack all of your active branch refs (which will probably end up affecting performance of normal git command usage)
> > https://git-scm.com/docs/git-pack-refs
> 
> Thanks! Indeed, I used that for testing on local repo but was asking more in the lines of: how to properly do this inside one of our tests? I didn't find a way to do it from Python (apart from subprocess call)

As the problem was really about annotated tags, not `packed refs`, there is not really need for that

pmrowla:
> Thanks! Indeed, I used that for testing on local repo but was asking more in the lines of: how to properly do this inside one of our tests? I didn't find a way to do it from Python (apart from subprocess call)

In tests you can use CLI git directly via the `scm` fixture and gitpython like

```
scm.gitpython.git.pack_refs(all=True)
```

(https://gitpython.readthedocs.io/en/stable/tutorial.html#using-git-directly)

dvc/scm/git/backend/dulwich/__init__.py
pmrowla: This is the right idea, but it's possible to have a longer chain of tag objects pointing to other tags which eventually point to a commit, so peeling the ref requires resolving the entire chain.

We should be able to do something like

```suggestion
            if isinstance(self.repo[ref], Tag):
                ref = self.repo.get_peeled(ref)
```

(note that I haven't tested this, you may need to dig a bit more into how `repo.get_peeled()` or `repo.refs.get_peeled()` works):
https://github.com/dulwich/dulwich/blob/ee231198600efcfa4874130f7fc619d8fee30c31/dulwich/repo.py#L708

daavoo: It works (at least for the test cases included)

dvc/scm/git/backend/dulwich/__init__.py
```diff
@@ -342,6 +343,8 @@ def get_ref(self, name, follow: bool = True) -> Optional[str]:
             except ValueError:
                 pass
         if ref:
+            if isinstance(self.repo[ref], Tag):
+                ref = self.repo.get_peeled(name_b)
```
pmrowla, line 347: This breaks the case where `name` is a symbolic reference and `follow = False`, like when `HEAD` points to `refs/heads/master`.  (see the current test failures)

Basically we need to catch `KeyError` when we do the `self.repo[ref]` check, and `pass` as long as `follow` is also `False`

pmrowla (changes requested):

skshetry:
@daavoo, please open this PR on the upstream. Really sorry for the inconvenience. 🙂

PR #7010, opened 2021-11-20
tests: run on Windows with Python 3.10

skshetry:
pygit2 wheels should be now available.
Had to skip pytest-docker for Windows + Python 3.10 due to a dependency issue, see: docker/docker-py#2902.


efiop (approved):
Thank you! 🙏

PR #7011, opened 2021-11-22
scm: remove ignore list

skshetry:
There is no `scm_context.ignore_list`. Instead of adding APIs, I am removing its use and replacing it with a loop (which is what it really did).

Looking at the code, I think we should really stop having side effects by just initializing a `Repo` instance:
https://github.com/iterative/dvc/blob/318b269ad118f198bb97b203e151529398da0f54/dvc/repo/__init__.py#L206-L219

### DIFF
--- a/tests/func/test_repo_index.py
+++ b/tests/func/test_repo_index.py
@@ -200,7 +200,7 @@ def test_dumpd(dvc):
 
 
 def test_unique_identifier(tmp_dir, dvc, scm, run_copy):
-    dvc.config["core"]["autostage"] = True
+    dvc.scm_context.autostage = True
     tmp_dir.dvc_gen("foo", "foo")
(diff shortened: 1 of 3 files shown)


tests/func/test_repo_index.py
```diff
@@ -200,7 +200,7 @@ def test_dumpd(dvc):
 
 
 def test_unique_identifier(tmp_dir, dvc, scm, run_copy):
-    dvc.config["core"]["autostage"] = True
+    dvc.scm_context.autostage = True
```
skshetry, line 203: Changing config won't really work as `SCMContext` determines `autostage` config when it's initialized.

skshetry:
One test failing due to a timeout added in #7006, will fix it separately.

efiop (approved):

Issue #6975, opened 2021-11-12
tests: wait for python 3.9 rerelase


pared:
The task to clean up after https://github.com/iterative/dvc/pull/6958.

Bug discussion: https://bugs.python.org/issue45235

It seems that python 3.9 gets rerelase. 
Other python versions should not be affected as the change has been already reverted, and the next release is python 3.10, on the 6th of December.

PR #7016, opened 2021-11-22
packages: build binary pkgs with Python 3.9

skshetry:


skshetry:
Package builds running at https://github.com/iterative/dvc/actions/runs/1489796368.

efiop:
@skshetry Did 3.9.8 get a rerelease?

skshetry:
> @skshetry Did 3.9.8 get a rerelease? #6975

yeah, there is a 3.9.9 release: https://www.python.org/downloads/release/python-399.

skshetry:
Looks like setup-python is still using `3.9.8`.

pared:
So we probably need to wait for them now?

skshetry:
> So we probably need to wait for them now?

From https://github.com/actions/virtual-environments:

> Regular Weekly Rhythm can be followed by watching Releases. Pre-release is created when deployment is started. As soon as deployment is finished, it is converted to release. You can also track upcoming changes using the awaiting-deployment label.

I already see pre-releases with `Python 3.9.9`, so we need to wait till they turn to `release`.

pared (approved):
Seems like the new 3.9 is on now.

PR #7017, opened 2021-11-22
dvc.scm: refactor exceptions

skshetry:
This PR splits scm-related exceptions into two kinds:
1. Internal scm exceptions that are in `dvc.scm.exceptions` and likely going to be available as `scmrepo.exceptions`.
2. External DVC-related exceptions in `dvc.scm.base` that is required for UI and dvc's API to work.
   This will be available in `dvc.scm` when we migrate.

So, due to 2, there are some exceptions with the same name in both 1 and 2, but they serve a different purpose and are just reraised from one to another.

This PR is still incomplete, in that we have to decide what to do with raised `SCMError` exceptions. As DVC is unaware of these exceptions, places where we are checking for `DvcException` or where we just exit from the `main`, it will throw errors with:
```
```

There may definitely be some UI issues with this breakage, which we need to discuss. The easiest path forward that I see is to make these changes and fix them in the future as we find them.

The other way would require us to go through chains of API calls where we raise `dvc.scm.exceptions.SCMError` and catch those and reraise those as `dvc.scm.base.SCMError`.

### DIFF
--- a/dvc/scm/__init__.py
+++ b/dvc/scm/__init__.py
@@ -31,3 +31,13 @@ def SCM(
         return NoSCM(root_dir)
 
     return Git(root_dir, search_parent_directories=search_parent_directories)
+
+
+def clone(url: str, to_path: str, **kwargs):
+    from .exceptions import CloneError as InternalCloneError
+    from .base import CloneError
+
+    try:
+        return Git.clone(url, to_path, **kwargs)
+    except InternalCloneError as exc:
+        raise CloneError(str(exc))
--- a/dvc/scm/__init__.py
+++ b/dvc/scm/__init__.py
@@ -26,11 +26,17 @@ def SCM(
     Returns:
         dvc.scm.base.Base: SCM instance.
     """
+    from dvc.scm.base import SCMError
+    from dvc.scm.exceptions import SCMError as InternalSCMError
 
-    if no_scm:
-        return NoSCM(root_dir)
-
-    return Git(root_dir, search_parent_directories=search_parent_directories)
+    try:
+        if no_scm:
+            return NoSCM(root_dir)
+        return Git(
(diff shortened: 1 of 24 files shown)


dvc/scm/__init__.py
```diff
@@ -26,8 +26,34 @@ def SCM(
...
+    from .base import CloneError
+    from .exceptions import CloneError as InternalCloneError
+
+    try:
+        return Git.clone(url, to_path, **kwargs)
+    except InternalCloneError as exc:
+        raise CloneError(str(exc))
+
+
+def resolve_rev(scm: "Git", rev: str) -> str:
```
pmrowla, line 52: Not related to this PR specifically, but we will probably need a `dvc.utils.resolve_rev` function that wraps the separate SCM package's `resolve_rev`.

The existing `scm.git.resolve_rev` method contains DVC specific behavior, where we extend the standard git ref (branch/tag) resolution to also include refs in the exps namespace. This should probably stay in DVC (and not the standalone package).

[`scm.git.__init__.py::resolve_rev`](https://github.com/iterative/dvc/blob/644e9ede924b6416b4ff714fdc0c3463190ab6f2/dvc/scm/git/__init__.py#L305)

pmrowla (approved):

dvc/scm/__init__.py
skshetry: @pmrowla, this `dvc/scm/__init__.py` will remain as `dvc/scm.py`. So, this `resolve_rev` is the one that will provide DVC specific behaviour. The following contents will be moved here in successive PRs:

efiop:
@skshetry could you rebase, please?

skshetry:
The new changes remaps internal `SCMError` into `DvcException` counterparts in `external_repo` and `experiments` (as @pmrowla suggested).

PR #7032, opened 2021-11-23
tests: patch webbrowser.open

### LINKED ISSUE #7014: tests: opens browser with plots when running tests
See https://github.com/iterative/dvc/pull/6913#discussion_r754003524.

Can be reproduced with `pytest tests/func/test_live.py`.

daavoo:
Prevent tests from opening browser windows.

Closes #7014

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -178,3 +178,9 @@ def custom_template(tmp_dir, dvc):
 
 
 scriptify_fixture = pytest.fixture(lambda: scriptify, name="scriptify")
+
+
+@pytest.fixture(scope="session", autouse=True)
+def webbrowser_open(session_mocker):
+    session_mocker.patch("dvc.repo.live.webbrowser_open")
+    session_mocker.patch("webbrowser.open")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -180,7 +180,6 @@ def custom_template(tmp_dir, dvc):
 scriptify_fixture = pytest.fixture(lambda: scriptify, name="scriptify")
 
 
-@pytest.fixture(scope="session", autouse=True)
-def webbrowser_open(session_mocker):
-    session_mocker.patch("dvc.repo.live.webbrowser_open")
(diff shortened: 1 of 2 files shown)


pared (approved):

tests/func/test_live.py
```diff
@@ -66,6 +66,7 @@ def live_stage(tmp_dir, scm, dvc, mocker):
         pytest.skip("no dvclive")
 
     mocker.patch("dvc.stage.run.Monitor.AWAIT", 0.01)
+    mocker.patch("dvc.repo.live.webbrowser_open")
```
skshetry, line 69: I'm worried that this may get unnoticed on new tests. We should mock `webbrowser.open` globally if that's possible.

skshetry (approved):

tests/conftest.py
```diff
@@ -178,3 +178,9 @@ def custom_template(tmp_dir, dvc):
 
 
 scriptify_fixture = pytest.fixture(lambda: scriptify, name="scriptify")
+
+
+@pytest.fixture(scope="session", autouse=True)
+def webbrowser_open(session_mocker):
+    session_mocker.patch("dvc.repo.live.webbrowser_open")
```
skshetry, line 185: Let's drop this? We may still want to test `webbrowser_open`?

daavoo: It is still possible to re-patch inside the test and it works as intended

daavoo: And without it, the global patch doesn't work for the test in `tests/func/live`

skshetry: I don't think it's a good idea to mock `dvc.repo.live.webbrowser_open` globally.

skshetry: Merging, let's see if we can fix this after https://github.com/iterative/dvc/pull/6993. We may be able to mock `ui.open_browser`.

skshetry: Okay, making the imports for `webbrowser` lazy makes it work. Pushing the changes.

PR #7039, opened 2021-11-24
scm: remove dependency from dvc's utils in scm

skshetry:
This removes dependency from dvc.utils and dvc.types from DVC.
Some utils like fix_env, is_binary have been reimplemented.
Regarding fix_env and is_binary, I think eventually this
should be a responsibility of DVC itself to set.

relpath implementation in DVC also works with paths in
different drives in Windows. I have changed this to use
os.path.relpath for now.


skshetry:
After this PR, there are no dependencies on `dvc` in `dvc.scm`. The only thing remaining is `dvc.fs.git` refactoring.

efiop (approved):

PR #7042, opened 2021-11-24
fs.git: refactor GitFileSystem to be fsspec compliant

skshetry:
Also adds a wrapper for DVC to use as GitFileSystem in dvc.fs.scm.
This is temporary, and will be renamed back to dvc.fs.git after the
split.

Note that this filesystem is not fully compliant, it's only
meant to be compatible with our previous implementation.

### DIFF
--- a/dvc/fs/git.py
+++ b/dvc/fs/git.py
@@ -1,52 +1,90 @@
 import errno
 import os
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    BinaryIO,
+    Callable,
+    Dict,
+    Optional,
+    Tuple,
+)
 
-from tqdm.utils import CallbackIOWrapper
+from fsspec.spec import AbstractFileSystem
 
-from dvc.utils import is_exec, relpath
+if TYPE_CHECKING:
+    from io import BytesIO
 
-from ..progress import DEFAULT_CALLBACK
-from .base import FileSystem
+    from dvc.scm.git import Git
+    from dvc.scm.git.objects import GitTrie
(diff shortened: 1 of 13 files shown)


dvc/fs/scm.py
```diff
@@ -0,0 +1,72 @@
+import os
```
skshetry, line 1: This file is named `fs/scm.py` because I need another file's history. This will be renamed to `fs/git.py` after the split, and the other file will be part of `scmrepo`.

dvc/fs/git.py
```diff
@@ -1,52 +1,88 @@
...
+        length = obj.seek(0, os.SEEK_END)
+        obj.seek(offset)
+    except (AttributeError, OSError):
+        return None
+    return length
 
-    sep = os.sep
 
-    scheme = "local"
+class GitFileSystem(AbstractFileSystem):
```
skshetry, line 32: This implementation is not complete but is compatible with the previous implementation. We may need to iterate separately on `scmrepo`.

dvc/fs/scm.py
efiop: @skshetry Btw, there will be lots of changes in these files, so maybe not worth bothering with preserving the history, if it will make it significantly easier to split.

skshetry: This should be the last thing in DVC.

efiop (approved):

efiop:
For the record: test failure is unrelated timeout issue.

Issue #6267, opened 2021-07-01
executor: run experiment on an executor


dberenbaum:
Meta-issue. Please link related tickets back to this one.

### User story

I want to run an experiment on my executor from my local machine.

### Requirements

* `dvc exp run --executor` successfully runs an experiment on a running executor instance with no other steps.

### Out of scope

* It's not yet necessary to monitor the experiment or its output while it runs or to be notified when the experiment completes.
* It's not yet necessary to queue or run multiple experiments on an executor.
* It's not yet necessary to create or destroy the executor instance as part of the experiment run.

dberenbaum:
It's important to keep the remote executor's experiment running even if the connection from the local machine breaks.

May need a `dvc exp collect` or similar command to check status of running experiments.

dberenbaum:
Need to rescope

Issue #4294, opened 2020-07-28
macOS: analytics daemon crashes due to fork related error


pmrowla:
After the daemon changes introduced in #4262 (f799e1e3), DVC on macOS crashes when we try to spawn the analytics daemon.

```
dvc git:f799e1e3 ✩  py:dvc ❯ dvc version
WARNING: Unable to detect supported link types, as cache directory '.dvc/cache' doesn't exist. It is usually auto-created by commands such as `dvc add/fetch/pull/run/import`, but you could create it manually to enable this check.
DVC version: 1.1.11+f799e1
Python version: 3.8.5
Platform: macOS-10.15.6-x86_64-i386-64bit
Binary: False
Package: None
Supported remotes: azure, gdrive, gs, hdfs, http, https, s3, ssh, oss
Repo: dvc, git
Filesystem type (workspace): ('apfs', '/dev/disk1s1')
objc[34576]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called.
objc[34576]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
```
![Screen Shot 2020-07-28 at 4 37 51 PM](https://user-images.githubusercontent.com/651988/88633994-bd026280-d0f0-11ea-9d1f-a60e755bbbfa.png)

Disabling analytics avoids the crash
```
dvc git:f799e1e3 ✩  py:dvc ❯ dvc config --local core.analytics false
dvc git:f799e1e3 ✩  py:dvc ❯ dvc version
WARNING: Unable to detect supported link types, as cache directory '.dvc/cache' doesn't exist. It is usually auto-created by commands such as `dvc add/fetch/pull/run/import`, but you could create it manually to enable this check.
DVC version: 1.1.11+f799e1
Python version: 3.8.5
Platform: macOS-10.15.6-x86_64-i386-64bit
Binary: False
Package: None
Supported remotes: azure, gdrive, gs, hdfs, http, https, s3, ssh, oss
Repo: dvc, git
Filesystem type (workspace): ('apfs', '/dev/disk1s1')
```

This problem appears to be related to `fork` being unsafe in macOS 

See related Python issue: https://bugs.python.org/issue33725 and Python 3.8 [multiprocessing docs](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods)

[...]

PR #7048, opened 2021-11-25
proc: initial (local) process manager implementation

pmrowla:

Related to https://github.com/iterative/dvc/issues/6267

Local prerequisite for https://github.com/iterative/dvc/discussions/7002

- Adds `dvc.proc` module for managing background processes
- Process state information is kept within the manager directory
- Processes are backgrounded and output is redirected to a file (stdin interactivity is not currently supported)
   - Backgrounding is handled through `multiprocessing.Process(daemon=True)`

Functionality is not used anywhere else in DVC yet (exp tempdir/queue runs will be the first place it gets added)

Python shell demo:

[asciicast](https://asciinema.org/a/K5DA7pbbrRH1LBHcHLbTNqBje)

### DIFF
--- a/dvc/proc/manager.py
+++ b/dvc/proc/manager.py
@@ -1,0 +1,92 @@
+"""Serverless process manager."""
+
+import json
+import logging
+import os
+from typing import Generator, List, Optional, Union
+
+from shortuuid import uuid
+
+from .process import ManagedProcess, ProcessInfo
+
+logger = logging.getLogger(__name__)
+
+
+class ProcessManager:
+    """Manager for controlling background ManagedProcess(es).
+
+    Spawned process entries are kept in the manager directory until they
+    are explicitly removed (with remove() or cleanup()) so that return
+    value and log information can be accessed after a process has completed.
+    """
+
+    def __init__(self, wdir: Optional[str] = None):
+        self.wdir = wdir or "."
+
+    def __iter__(self):
+        return self.processes()
+
+    def __getitem__(self, key: str) -> "ProcessInfo":
+        info_path = os.path.join(self.wdir, key, f"{key}.json")
+        try:
+            with open(info_path, encoding="utf-8") as fobj:
+                return ProcessInfo.from_dict(json.load(fobj))
+        except FileNotFoundError:
+            raise KeyError
+
+    def get(self, key: str, default=None):
+        try:
+            return self[key]
+        except KeyError:
+            return default
+
+    def processes(self) -> Generator["ProcessInfo", None, None]:
+        if not os.path.exists(self.wdir):
+            return
+        for name in os.listdir(self.wdir):
+            try:
+                yield self[name]
+            except KeyError:
+                continue
+
+    def spawn(self, args: Union[str, List[str]], name: Optional[str] = None):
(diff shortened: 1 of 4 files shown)


dvc/proc/manager.py
```diff
@@ -0,0 +1,92 @@
...
+
+        Raises:
+            ProcessNotTerminatedError if the specified process is still
+            running and was not forcefully killed.
+        """
+        raise NotImplementedError
+
+    def cleanup(self):
+        """Remove stale (terminated) processes from this manager."""
+        raise NotImplementedError
```
pmrowla, line 92: @karajan1001 the next set of tasks will be to fill out these methods.

- `send_signal`, `terminate`, `kill` are all related and would be grouped together in the first task. Implementation/behavior for these methods should work the same as they do in `subprocess` and `multiprocessing` (https://docs.python.org/3/library/subprocess.html#subprocess.Popen.send_signal).
- `remove` and `cleanup` can be implemented once `terminate`/`kill` are done (since force-removing a child process directory will require killing the process if it is still running)

dvc/proc/process.py
```diff
@@ -0,0 +1,198 @@
...
+        proc = mp.Process(
+            target=cls._spawn,
+            args=args,
+            kwargs=kwargs,
+            daemon=True,
+        )
+        proc.start()
+        # Do not terminate the child daemon when the main process exits
+        # pylint: disable=protected-access
+        mp.process._children.discard(proc)  # type: ignore[attr-defined]
```
pmrowla, line 192: This is kind of a hack, but even if we subclassed `multiprocessing.Process` we would just end up with an overridden

```
def start(self):
    super().start()
    _children.discard(self)
```

which I'm not sure is any cleaner


But in general, the idea behind using multiprocessing is that it allows us to use `spawn` instead of `fork` on windows + mac (https://github.com/iterative/dvc/issues/4294), and also allows us to directly run the target python function we want instead of adding the overhead from using `dvc.daemon` + a new dedicated CLI command

dvc/proc/process.py
```diff
@@ -0,0 +1,198 @@
...
+    PARAM_RETURNCODE = "returncode"
+
+    pid: int
+    stdin: Optional[str]
+    stdout: Optional[str]
+    stderr: Optional[str]
+    returncode: Optional[int]
+
+    @classmethod
+    def from_dict(cls, d):
```
skshetry, line 35: Isn't this equivalent to `ProcessInfo(**d)`?

dvc/proc/process.py
```diff
@@ -0,0 +1,198 @@
...
+        return cls(
+            d.get(cls.PARAM_PID),
+            d.get(cls.PARAM_STDIN),
+            d.get(cls.PARAM_STDOUT),
+            d.get(cls.PARAM_STDERR),
+            d.get(cls.PARAM_RETURNCODE),
+        )
+
+    def to_dict(self):
```
skshetry, line 44: There is [`dataclasses.asdict`](https://docs.python.org/3/library/dataclasses.html) that does this already.

karajan1001 (approved):

PR #7064, opened 2021-11-29
TabularData: Add ignore_empty flag to drop_duplicates.

daavoo:
- TabularData: Add ignore_empty flag to drop_duplicates

- TabularData: Fix drop_duplicates for rich_text NaN

### DIFF
--- a/dvc/compare.py
+++ b/dvc/compare.py
@@ -249,7 +249,7 @@ class TabularData(MutableSequence[Sequence["CellT"]]):
         else:
             self.drop(*to_drop)
 
-    def drop_duplicates(self, axis: str = "rows"):
+    def drop_duplicates(self, axis: str = "rows", ignore_empty: bool = True):
         if axis not in ["rows", "cols"]:
             raise ValueError(
                 f"Invalid 'axis' value {axis}."
@@ -260,9 +260,9 @@ class TabularData(MutableSequence[Sequence["CellT"]]):
             cols_to_drop: List[str] = []
             for n_col, col in enumerate(self.columns):
                 # Cast to str because Text is not hashable error
-                unique_vals = {
-                    str(x) for x in col if str(x) != self._fill_value
-                }
+                unique_vals = {str(x) for x in col}
+                if ignore_empty and self._fill_value in unique_vals:
+                    unique_vals -= {self._fill_value}
                 if len(unique_vals) == 1:
                     cols_to_drop.append(self.keys()[n_col])
             self.drop(*cols_to_drop)
--- a/dvc/compare.py
+++ b/dvc/compare.py
@@ -260,7 +260,9 @@ class TabularData(MutableSequence[Sequence["CellT"]]):
             cols_to_drop: List[str] = []
             for n_col, col in enumerate(self.columns):
                 # Cast to str because Text is not hashable error
-                unique_vals = {str(x) for x in col if x != self._fill_value}
+                unique_vals = {
+                    str(x) for x in col if str(x) != self._fill_value
+                }
                 if len(unique_vals) == 1:
                     cols_to_drop.append(self.keys()[n_col])
             self.drop(*cols_to_drop)
--- a/tests/unit/test_tabular_data.py
+++ b/tests/unit/test_tabular_data.py
@@ -226,30 +226,56 @@ def test_dropna(axis, how, data, expected):
         (
             "rows",
             [
-                ["foo", "", ""],
-                ["foo", "foo", ""],
+                ["foo", "-", "-"],
+                ["foo", "foo", "-"],
                 ["foo", "bar", "foobar"],
             ],
         ),
-        ("cols", [[""], ["foo"], ["foo"], ["bar"]]),
+        ("cols", [["-"], ["foo"], ["foo"], ["bar"]]),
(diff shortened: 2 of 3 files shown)


tests/unit/test_tabular_data.py
```diff
@@ -252,6 +252,32 @@ def test_drop_duplicates(axis, expected):
...
+    td.extend(
+        [
+            ["foo", None, ui.rich_text("-")],
+            ["foo", "foo"],
+            ["foo", "foo"],
+            ["foo", "bar", "foobar"],
+        ]
+    )
+
+    assert list(td) == [
```
karajan1001, line 269: LGTM for the most part has only one problem here. Empty values are not the same thing as a non-empty one, should we drop them here?

karajan1001 (approved):

tests/unit/test_tabular_data.py
daavoo: Not sure I fully understand

karajan1001: Here, we drop a column if all of them share the same value, or some of them are `None` and others share the same. But the `None` value is not the same thing as a definite one they might behave differently in some cases.

daavoo: Added a flag to configure: `nan_is_value`. 

```
    assert list(td) == [
        ["foo", "-", "-"],
        ["foo", "foo", "-"],
        ["foo", "foo", "-"],
        ["foo", "bar", "foobar"],
    ]

    td.drop_duplicates("cols", nan_is_value=True)

    assert list(td) == [
        ["-", "-"],
        ["foo", "-"],
        ["foo", "-"],
        ["bar", "foobar"],
    ]
```

Defaults to `False` to preserve old behavior

pmrowla:
If the problem here is that `TabularData` is being filled with a mix of `None` and `"-"` strings (i.e. `table._fill_value`), this seems like a bug that needs to be fixed in whatever is generating the `TabularData`, and not in specific UI functions that filter/render `TabularData`. (It seems like this will probably also show up in places other than `drop_duplicates` at some point)

`TabularData` should either

- always store `None` values as `None` (and replace them with `_fill_value` at the time they are rendered to UI)
- always replace `None` values with `_fill_value` at the time the value is added to the table/row/cell.

Looking at the code it looks like the 2nd option is what is supposed to be happening, so my understanding is that `drop_duplicates` should never be encountering `None` values in the first place.

daavoo:
> If the problem here is that TabularData is being filled with a mix of None and "-" strings (i.e. table._fill_value)

That's not the problem. 

One problem is that the table is being filled with a mix of `_fill_value` of type string and `_fill_value` of type `rich.Text` (from show_experiments).

This is addressed in the first commit https://github.com/iterative/dvc/pull/7064/commits/b09d353c04b2a3df41cb93855e2dfa305574ac6f

>  and not in specific UI functions that filter/render TabularData

The `nan_is_value` flag for `drop_duplicates` is not fixing any problem but allowing to configure 2 valid behaviors that depend on the use case.

`drop_duplicates` should remove columns where all rows have the same value. There is an ambiguity on how to treat NaNs, i.e the following column:

```
foo
foo
foo
None
```

For some use cases, could be considered all duplicates because `None` it's not considered a relevant value. For others, `None` is considered a relevant value and thus this column should not be dropped.

> always store None values as None (and replace them with _fill_value at the time they are rendered to UI)

I prefer this option. Was trying to not introduce many changes here but it's probably the way to go.

pmrowla:
> For some use cases, could be considered all duplicates because `None` it's not considered a relevant value. For others, `None` is considered a relevant value and thus this column should not be dropped.

Ah ok, that makes sense then.

> > always store None values as None (and replace them with _fill_value at the time they are rendered to UI)
> 
> I prefer this option. Was trying to not introduce many changes here but it's probably the way to go.

After reading your explanation, this PR is probably fine for now, maybe just open a separate issue regarding making `TabularData` store `None` types instead of the fill strings.

---

As a side note, I would say that NaN is not the same thing as a `None`/null (even if python doesnt have native types for nan and +/- infinity) and would maybe name this something else like `ignore_none`/`ignore_empty`/skip_empty` (to skip/ignore empty cells when determining if the value changed).

dvc/compare.py
```diff
@@ -260,7 +260,9 @@ def drop_duplicates(self, axis: str = "rows"):
             cols_to_drop: List[str] = []
             for n_col, col in enumerate(self.columns):
                 # Cast to str because Text is not hashable error
-                unique_vals = {str(x) for x in col if x != self._fill_value}
+                unique_vals = {str(x) for x in col}
+                if ignore_empty and self._fill_value in unique_vals:
```
karajan1001, line 264:
```
                if ignore_empty:
                   unique_vals -= {self._fill_value}
```
Set algorithm.

tests/unit/test_tabular_data.py
```diff
@@ -252,6 +252,55 @@ def test_drop_duplicates(axis, expected):
     assert list(td) == expected
 
 
+def test_drop_duplicates_ignore_empty():
```
karajan1001, line 255: A little bit duplicated with the previous test, could we merge them into one?

Issue #7062, opened 2021-11-29
proc: implement process signaling


pmrowla:
`ProcessManager` should implement signaling methods for local processes in the same way as `subprocess` and `multiprocessing`

https://docs.python.org/3/library/subprocess.html#subprocess.Popen.send_signal

- [ ] `send_signal`
- [ ] `terminate`
- [ ] `kill`

Follow up to https://github.com/iterative/dvc/pull/7048
Prerequisite for https://github.com/iterative/dvc/issues/6267

PR #7066, opened 2021-11-30
proc: implement process signaling(#7062)

karajan1001:
fix: #7062 
1. Add `send_signal`, `kill`, `terminate` to ProcessManager.
2. Add tests for each of them.

### DIFF
--- a/dvc/proc/exceptions.py
+++ b/dvc/proc/exceptions.py
@@ -18,3 +18,8 @@ class TimeoutExpired(DvcException):
         )
         self.cmd = cmd
         self.timeout = timeout
+
+
+class UnsupportedSignalError(DvcException):
+    def __init__(self, sig):
+        super().__init__(f"Unsupported signal: {sig}")
--- a/dvc/proc/manager.py
+++ b/dvc/proc/manager.py
@@ -3,10 +3,14 @@
 import json
 import logging
 import os
+import signal
+import sys
 from typing import Generator, List, Optional, Union
 
+from funcy.flow import reraise
 from shortuuid import uuid
 
+from .exceptions import UnsupportedSignalError
 from .process import ManagedProcess, ProcessInfo
 
 logger = logging.getLogger(__name__)
@@ -33,6 +37,12 @@ class ProcessManager:
                 return ProcessInfo.from_dict(json.load(fobj))
         except FileNotFoundError:
             raise KeyError
+
+    @reraise(FileNotFoundError, KeyError)
+    def __setitem__(self, key: str, value: "ProcessInfo"):
+        info_path = os.path.join(self.wdir, key, f"{key}.json")
+        with open(info_path, "w", encoding="utf-8") as fobj:
+            return json.dump(value.asdict(), fobj)
 
     def get(self, key: str, default=None):
         try:
@@ -63,17 +73,47 @@ class ProcessManager:
             pid,
         )
 
-    def send_signal(self, name: str, signal: int):
+    def send_signal(self, name: str, sig: int):
         """Send `signal` to the specified named process."""
-        raise NotImplementedError
+        process_info = self[name]
+        if sys.platform == "win32":
+            if sig not in (
+                signal.SIGTERM,
+                signal.CTRL_C_EVENT,
+                signal.CTRL_BREAK_EVENT,
+            ):
+                raise UnsupportedSignalError(sig)
 
-    def kill(self, name: str):
-        """Kill the specified named process."""
-        raise NotImplementedError
+        def handle_closed_process():
+            logging.warning(
+                f"Process {name} had already aborted unexpectedly."
+            )
+            process_info.returncode = -1
+            self[name] = process_info
+
+        if process_info.returncode is None:
+            try:
+                os.kill(process_info.pid, sig)
+            except ProcessLookupError:
+                handle_closed_process()
+                raise
+            except OSError as exc:
+                if sys.platform == "win32":
+                    if exc.winerror == 87:
+                        handle_closed_process()
+                        raise ProcessLookupError from exc
+                raise
 
     def terminate(self, name: str):
         """Terminate the specified named process."""
-        raise NotImplementedError
+        self.send_signal(name, signal.SIGTERM)
+
+    def kill(self, name: str):
+        """Kill the specified named process."""
+        if sys.platform == "win32":
+            self.send_signal(name, signal.SIGTERM)
+        else:
+            self.send_signal(name, signal.SIGKILL)
 
     def remove(self, name: str, force: bool = False):
         """Remove the specified named process from this manager.
--- a/tests/unit/proc/test_manager.py
+++ b/tests/unit/proc/test_manager.py
@@ -1,0 +1,84 @@
+import json
+import os
+import signal
+import sys
+
+import pytest
+
+from dvc.proc.exceptions import UnsupportedSignalError
+from dvc.proc.manager import ProcessManager
+from dvc.proc.process import ProcessInfo
+
+PID_FINISHED = 1234
+PID_RUNNING = 5678
+
+
+def create_process(root: str, name: str, pid: int, returncode=None):
+    info_path = os.path.join(root, name, f"{name}.json")
+    os.makedirs(os.path.join(root, name))
+    process_info = ProcessInfo(
+        pid=pid, stdin=None, stdout=None, stderr=None, returncode=returncode
+    )
+    with open(info_path, "w", encoding="utf-8") as fobj:
+        json.dump(process_info.asdict(), fobj)
+
+
+@pytest.fixture
+def finished_process(tmp_dir):
+    key = "finished"
+    create_process(tmp_dir, key, PID_FINISHED, 0)
+    return key
+
+
+@pytest.fixture
+def running_process(tmp_dir):
+    key = "running"
+    create_process(tmp_dir, key, PID_RUNNING)
+    return key
+
+
+def test_send_signal(tmp_dir, mocker, finished_process, running_process):
+    m = mocker.patch("os.kill")
+    process_manager = ProcessManager(tmp_dir)
+    process_manager.send_signal(running_process, signal.SIGTERM)
+    m.assert_called_once_with(PID_RUNNING, signal.SIGTERM)
+
+    m = mocker.patch("os.kill")
+    process_manager.send_signal(finished_process, signal.SIGTERM)
+    m.assert_not_called()
+
+    if sys.platform == "win32":
+        with pytest.raises(UnsupportedSignalError):
+            process_manager.send_signal(finished_process, signal.SIGABRT)
+
+
+def test_dead_process(tmp_dir, mocker, running_process):
+    process_manager = ProcessManager(tmp_dir)
+    with pytest.raises(ProcessLookupError):
+        process_manager.send_signal(running_process, signal.SIGTERM)
+    assert process_manager[running_process].returncode == -1
+
+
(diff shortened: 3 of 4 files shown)


dvc/proc/manager.py
```diff
@@ -63,17 +65,62 @@ def spawn(self, args: Union[str, List[str]], name: Optional[str] = None):
...
-        raise NotImplementedError
-
-    def kill(self, name: str):
-        """Kill the specified named process."""
-        raise NotImplementedError
-
-    def terminate(self, name: str):
-        """Terminate the specified named process."""
-        raise NotImplementedError
+    if sys.platform == "win32":
```
pmrowla, line 68:
```suggestion
    if os.name == "nt":
```

In the event that we are running in cygwin, we should still use the windows specific methods (`sys.platform` would be `cygwin` and not `win32`)

dvc/proc/manager.py
```diff
@@ -63,17 +65,62 @@ def spawn(self, args: Union[str, List[str]], name: Optional[str] = None):
...
+            handle = _winapi.OpenProcess(
+                PROCESS_TERMINATE, False, process_info.pid
+            )
+            try:
+                _winapi.TerminateProcess(handle, 1)
+            except PermissionError:
+                rc = _winapi.GetExitCodeProcess(handle)
+                if rc == _winapi.STILL_ACTIVE:
+                    raise
+                process_info.returncode = rc
```
pmrowla, line 112: One other suggestion, it may be better to just define this as `_terminate_nt(self, ...)`, and then call it from `send_signal()` for `SIGKILL`/`SIGTERM`. That way, we could keep all of the platform specific handling in `send_signal()`, and not need platform specific handling in `kill()`/`terminate()` at all (since they just call `self.send_signal()`.

dvc/proc/manager.py
```diff
@@ -63,17 +65,62 @@ def spawn(self, args: Union[str, List[str]], name: Optional[str] = None):
...
+            # pylint: disable=no-member
+            process_info = self[name]
+            if process_info.returncode is None:
+                return
+            if sig == signal.SIGTERM:
+                self.terminate(name)
+            elif sig == signal.CTRL_C_EVENT:
+                os.kill(process_info.pid, signal.CTRL_C_EVENT)
+            elif sig == signal.CTRL_BREAK_EVENT:
+                os.kill(process_info.pid, signal.CTRL_BREAK_EVENT)
```
pmrowla, line 81:
```suggestion
            if sig in (signal.SIGTERM, signal.SIGKILL):
                self.terminate(name)
            elif sig in (signal.CTRL_C_EVENT, signal.CTRL_BREAK_EVENT):
                os.kill(process_info.pid, sig)
```

dvc/proc/manager.py
```diff
@@ -63,17 +65,62 @@ def spawn(self, args: Union[str, List[str]], name: Optional[str] = None):
...
-    def terminate(self, name: str):
-        """Terminate the specified named process."""
-        raise NotImplementedError
+    if sys.platform == "win32":
+
+        def send_signal(self, name: str, sig: int):
+            """Send `signal` to the specified named process."""
+            # pylint: disable=no-member
+            process_info = self[name]
+            if process_info.returncode is None:
```
pmrowla, line 74:
```suggestion
            if process_info.returncode is not None:
```

pmrowla (changes requested):

dvc/proc/manager.py
pmrowla: Also, this could probably be broken up into a separate helper function for getting the windows process handle, which may be something we end up needing in the future.

pmrowla: Actually looking at the python [docs](https://docs.python.org/3/library/os.html), I think we can probably get rid of all of this code entirely and just call `os.kill(pid, -1)` instead of trying to call the windows `TerminateProcess()` ourselves

> Any other value for sig will cause the process to be unconditionally killed by the TerminateProcess API, and the exit code will be set to sig 

@karajan1001

karajan1001: Yes, but still need to handle these `SIGKILL` and `CTRL_C_EVENT` because they might not exist in some of the platforms

pmrowla: Right, I meant for the kill/terminate case on windows we can just call `os.kill(pid, -1)`, we don't need to take care of using windows process handles + `TerminateProcess()` ourselves

dvc/proc/manager.py
karajan1001: But one problem here, in the signal class, the signal constants' defination follows `sys.platform == "win32"`. So If we use `os.name == "nt"` here we would face a constants mismatch problem in `cygwin`?

dvc/proc/manager.py
pmrowla: I edited my change here, I thought `signal.SIGKILL` was aliased to `SIGTERM` on windows in python, but it's not defined at all.

dvc/proc/manager.py
pmrowla: It looks like I was mistaken on the `os.name`/`sys.platform` difference, I did some testing and for cygwin `os.name == posix`, so it won't make a difference either way.

dvc/proc/manager.py
karajan1001: I copied most of these codes from `Popen`, if we can safely use `os.kill` here I think we don't need this complex logic at all.

pmrowla: For windows we should be checking the signal value though.

The issue being that if you use `os.kill()` on windows without checking the signal value, it will terminate the process on any signal other than `CTRL_C_EVENT` and `CTRL_BREAK_EVENT`. I think we should error out on the unexpected values instead of always terminating the process.

dvc/proc/manager.py
```diff
@@ -63,17 +64,30 @@ def spawn(self, args: Union[str, List[str]], name: Optional[str] = None):
...
-
-    def kill(self, name: str):
-        """Kill the specified named process."""
-        raise NotImplementedError
+        # pylint: disable=no-member
+        process_info = self[name]
+        if os.name == "nt":
+            if sig not in (
+                signal.SIGTERM,
+                signal.CTRL_C_EVENT,  # type:ignore
```
karajan1001, line 74: Here only with `sys.platform` check can prevent these mypy checking disabled here.

pmrowla: If mypy prefers `sys.platform` then we can just use that

dvc/proc/manager.py
```diff
@@ -63,17 +64,30 @@ def spawn(self, args: Union[str, List[str]], name: Optional[str] = None):
...
+                signal.CTRL_BREAK_EVENT,  # type:ignore
+            ):
+                raise ValueError("Unsupported signal: {}".format(sig))
```
pmrowla, line 77:
```suggestion
                raise ValueError(f"Unsupported signal: {sig}")
```

tests/unit/proc/test_manager.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def create_process(root: str, name: str, pid: int, returncode=None):
+    info_path = os.path.join(root, name, f"{name}.json")
+    os.makedirs(os.path.join(root, name))
+    process_info = ProcessInfo(
+        pid=pid, stdin=None, stdout=None, stderr=None, returncode=returncode
+    )
+    with open(info_path, "w", encoding="utf-8") as f_w:
+        json.dump(process_info.asdict(), f_w)
```
pmrowla, line 21:
```suggestion
    with open(info_path, "w", encoding="utf-8") as fobj:
        json.dump(process_info.asdict(), fobj)
```

I think we generally just use `fobj` everywhere in the DVC code base, so we should try to stay consistent here

dvc/proc/manager.py
```diff
@@ -63,17 +65,29 @@ def spawn(self, args: Union[str, List[str]], name: Optional[str] = None):
...
+        process_info = self[name]
+        if sys.platform == "win32":
+            if sig not in (
+                signal.SIGTERM,
+                signal.CTRL_C_EVENT,
+                signal.CTRL_BREAK_EVENT,
+            ):
+                raise ValueError(f"Unsupported signal: {sig}")
+        if process_info.returncode is None:
+            os.kill(process_info.pid, sig)
```
pmrowla, line 79: We need to account for potential `OSError`s raised on `os.kill()`. I think in general we do want to re-raise any exceptions, but there is at least one special case we should account for:

If the spawned daemon process crashes (or is explicitly killed by some other event, like the user running `kill -9` or a system crash/shutdown), this PID will no longer exist. For this error case we should catch `ProcessLookupError`, and then explicitly set `process_info.returncode` to `-1` and update the info file (to indicate that the process was closed due to some unknown error), so that we don't try to treat it as a running process again in subsequent calls. After we've done that, we can then re-raise the original `ProcessLookupError` exception up the call stack.

pmrowla (changes requested):
@karajan1001 looks good for the most part now, but we should account for potential system errors when possible

dvc/proc/manager.py
```diff
@@ -34,6 +37,15 @@ def __getitem__(self, key: str) -> "ProcessInfo":
         except FileNotFoundError:
             raise KeyError
 
+    def __setitem__(self, key: str, value: "ProcessInfo"):
+        info_path = os.path.join(self.wdir, key, f"{key}.json")
+        try:
+            with open(info_path, "w", encoding="utf-8") as fobj:
+                return json.dump(value.asdict(), fobj)
+        except FileNotFoundError:
```
pared, line 45: We can use `funcy.reraise` here. Not a necessity though.

pared (approved):

dvc/proc/exceptions.py
```diff
@@ -18,3 +18,8 @@ def __init__(self, cmd, timeout):
         )
         self.cmd = cmd
         self.timeout = timeout
+
+
+class UnsupportSignalError(DvcException):
```
pmrowla, line 23:
```suggestion
class UnsupportedSignalError(DvcException):
```

karajan1001:
Looks like the `os.kill`'s behaviour is still different in the Windows platform.

```
            try:
>               os.kill(process_info.pid, sig)
E               OSError: [WinError 87] The parameter is incorrect
```

It might require us to fix it some time later.

pmrowla:
> Looks like the `os.kill`'s behaviour is still different in the Windows platform.
> 
> ```
>             try:
> >               os.kill(process_info.pid, sig)
> E               OSError: [WinError 87] The parameter is incorrect
> ```
> 
> It might require us to fix it some time later.

In that case we should also catch `OSError`, and on windows treat `exc.winerror == 87` the same as `ProcessLookupError`

dvc/proc/manager.py
```diff
@@ -63,17 +73,47 @@ def spawn(self, args: Union[str, List[str]], name: Optional[str] = None):
...
+            try:
+                os.kill(process_info.pid, sig)
+            except ProcessLookupError:
+                handle_closed_process()
+                raise
+            except OSError as exc:
+                if sys.platform == "win32":
+                    if exc.winerror == 87:
+                        handle_closed_process()
+                        raise ProcessLookupError from exc
```
pmrowla, line 104: Why not just

```suggestion
                if sys.platform == "win32" and exc.winerror == 87:
                    handle_closed_process()
                    raise ProcessLookupError from exc
```

karajan1001: Going this way would cause `pylint` error, no argument `winerror` in `exc`.

pmrowla (approved):

Issue #7063, opened 2021-11-29
proc: implement cleanup/removal of completed processes


pmrowla:
`ProcessManager` does not automatically remove completed processes from the managed directory (so that exit status and output logs can be retrieved completion). The manager API will still need to support cleaning up these process directories once they are no longer needed.

- [ ] `remove` - note that `remove()` should support forcefully terminating existing running processes
- [ ] `cleanup`

Follow up to https://github.com/iterative/dvc/pull/7048
Prerequisite for https://github.com/iterative/dvc/issues/6267

PR #7067, opened 2021-11-30
proc: implement cleanup/removal  (#7063)

karajan1001:
fix: #7063
Add `cleanup` and `removal` function to ProcessManager

### DIFF
--- a/dvc/proc/manager.py
+++ b/dvc/proc/manager.py
@@ -5,12 +5,12 @@ import logging
 import os
 import signal
 import sys
-from typing import Generator, List, Optional, Union
+from typing import Generator, List, Optional, Tuple, Union
 
 from funcy.flow import reraise
 from shortuuid import uuid
 
-from .exceptions import UnsupportedSignalError
+from .exceptions import ProcessNotTerminatedError, UnsupportedSignalError
 from .process import ManagedProcess, ProcessInfo
 
 logger = logging.getLogger(__name__)
@@ -27,8 +27,11 @@ class ProcessManager:
     def __init__(self, wdir: Optional[str] = None):
         self.wdir = wdir or "."
 
-    def __iter__(self):
-        return self.processes()
+    def __iter__(self) -> Generator[str, None, None]:
+        if not os.path.exists(self.wdir):
+            return
+        for name in os.listdir(self.wdir):
+            yield name
 
     def __getitem__(self, key: str) -> "ProcessInfo":
         info_path = os.path.join(self.wdir, key, f"{key}.json")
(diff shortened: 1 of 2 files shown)


dvc/proc/manager.py
```diff
@@ -85,8 +88,21 @@ def remove(self, name: str, force: bool = False):
...
-        raise NotImplementedError
+        if not os.path.exists(self.wdir):
+            return
+        for name in os.listdir(self.wdir):
+            try:
+                process_info = self[name]
+            except KeyError:
+                continue
+            if process_info.returncode is not None:
+                remove(os.path.join(self.wdir, name))
```
pmrowla, line 108:
```suggestion
        for process_info in self.processes():
            if process_info.returncode is not None:
                remove(os.path.join(self.wdir, name))
```

No reason to duplicate the iterator exception handling when we can just use `self.processes()`

karajan1001: `cleanup` has no name here?

pmrowla:
Can we mark this WIP until the other PR is merged (this is blocked until `kill()` is actually implemented)

dvc/proc/manager.py
karajan1001: Currently, the self.processes() don't provide name info. We have several choices:
1. add `name` to ProcessInfo. 
2. return a (name, ProcessInfo) in `processes()` .
3. return `name`/`key` list in `__iter__(self)`.

For myself, I'd more prefer `3`.

pmrowla: 3 is probably the best option here (see also https://github.com/iterative/dvc/pull/7048/files#r756744379)

pmrowla: > `cleanup` is at the manager level, and should remove all stale processes. To remove individual named processes, the caller should just use `remove()`

pmrowla: One thing I was thinking is that eventually `cleanup` may need a `force` parameter, to remove all processes whether or not they are still running, but that's not a requirement for now.

pmrowla:
@karajan1001 looks like the signal handling tests are flaky
```
        if process_info.returncode is None:
            try:
>               os.kill(process_info.pid, sig)
E               PermissionError: [Errno 1] Operation not permitted

/home/runner/work/dvc/dvc/dvc/proc/manager.py:104: PermissionError
```

Since the tests assign dummy integer PIDs for our running and terminated processes, you need mock out the actual `os.kill` calls. In this case, the test for "process doesn't exist" fails because the dummy PID is actually a valid process on the test machine

pmrowla (approved):

PR #7084, opened 2021-12-02
exp: refactor (local) executors to start using process manager

pmrowla:

This does not change any user-facing DVC behavior, other than exp pidfile location (which was undocumented/for internal use only)

related to: https://github.com/iterative/dvc/issues/6440
prerequisite for: https://github.com/iterative/dvc/issues/6267

Currently in the experiments API, we have a clear separation between queuing/staging an experiment (generating a stash commit with the desired changes), and execution/reproduction. The problem right now is that "execution" includes other operations that should not be part of the "execution" step. Essentially both populating the temporary workspace and calling `Repo.reproduce` must always be done as a single operation right now. This is OK in the current local execution workflow, but does not work when we get into the remote scenario - populating the workspace is a push operation that must happen on the local machine, and repro'ing the experiment in that workspace is an operation that happens on the remote machine.

This refactor cleans up the exp executors API so that there is a clearer separation between operations that are not actually dependent on each other, and will allow us to serialize the executor state so that we can do different operations in separate DVC commands/processes when needed.

- introduce ExecutorManager class
- move executor init/execution/collection from `experiments.__init__` into the
  manager classes
- make executors serializable via ExecutorInfo
- move workspace repro into (managed) WorkspaceExecutor
- remove old exp run "pidfile" behavior in favor of serializing ExecutorInfo within process manager subdirs (alongside where process manager pidfiles/logs/etc are stored)

### DIFF
--- a/dvc/repo/experiments/executor/manager.py
+++ b/dvc/repo/experiments/executor/manager.py
@@ -1,0 +1,234 @@
+import logging
+import os
+from abc import ABC
+from collections import deque
+from typing import TYPE_CHECKING, Deque, Dict, Optional, Tuple, Type
+
+from dvc.proc.manager import ProcessManager
+
+from ..base import (
+    EXEC_BASELINE,
+    EXEC_HEAD,
+    EXEC_MERGE,
+    CheckpointExistsError,
+    ExperimentExistsError,
+    ExpRefInfo,
+    ExpStashEntry,
+)
+from .base import EXEC_PID_DIR
+from .local import TempDirExecutor, WorkspaceExecutor
+
+if TYPE_CHECKING:
+    from scmrepo.git import Git
+
+    from dvc.repo import Repo
+
+    from .base import BaseExecutor
+
+logger = logging.getLogger(__name__)
+
+
+class BaseExecutorManager(ABC):
+    """Manages executors for a collection of experiments to be run."""
+
+    EXECUTOR_CLS: Type = WorkspaceExecutor
+
+    def __init__(
+        self,
+        scm: "Git",
+        wdir: str,
+    ):
+        from dvc.utils.fs import makedirs
+
+        self.scm = scm
+        makedirs(wdir, exist_ok=True)
+        self.wdir = wdir
+        self.executors = self._load_infos()
+        self._queue: Deque[Tuple[str, "BaseExecutor"]] = deque()
+        self.proc = ProcessManager(self.pid_dir)
+
+    @property
+    def pid_dir(self) -> str:
+        return os.path.join(self.wdir, EXEC_PID_DIR)
+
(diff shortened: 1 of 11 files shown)


dvc/repo/experiments/executor/manager.py
```diff
@@ -0,0 +1,317 @@
...
+            logger.debug("Collected experiment '%s'.", exp_rev[:7])
+            results[exp_rev] = exec_result.exp_hash
+        return results
+
+    def exec_queue(self, jobs: Optional[int] = 1):
+        """Run a single WorkspaceExecutor.
+
+        Workspace execution is done within the main DVC process
+        (rather than in multiprocessing context)
+        """
```
pmrowla, line 280: In theory we could get rid of this entirely and just use `WorkspaceExecutor` inside the generic base multiprocessing context, but for now keeping the workspace specific behavior at least makes it easier to test exp reproduction (since we don't have to deal with testing function calls from inside multiprocessing child workers).

dvc/repo/experiments/executor/manager.py
```diff
@@ -0,0 +1,340 @@
...
+            logger.debug("Collected experiment '%s'.", exp_rev[:7])
+            results[exp_rev] = exec_result.exp_hash
+        return results
+
+    def exec_queue(self, jobs: Optional[int] = 1, detach: bool = False):
+        """Run a single WorkspaceExecutor.
+
+        Workspace execution is done within the main DVC process
+        (rather than in multiprocessing context)
+        """
```
pmrowla, line 301: In theory we could get rid of this entirely and just use WorkspaceExecutor inside the generic base multiprocessing context, but for now keeping the workspace specific behavior at least makes it easier to test exp reproduction (since we don't have to deal with testing function calls from inside multiprocessing child workers).

dvc/repo/experiments/executor/manager.py
```diff
@@ -0,0 +1,340 @@
...
+        from multiprocessing import Manager
+
+        from dvc.stage.monitor import CheckpointKilledError
+
+        result: Dict[str, Dict[str, str]] = defaultdict(dict)
+
+        manager = Manager()
+        pid_q = manager.Queue()
+
+        with ProcessPoolExecutor(max_workers=jobs) as workers:
```
pmrowla, line 159: This should eventually go away entirely in favor of having "attached" execution done by `ProcessManager` (or rather via an actual task queue that works on top of `ProcessManager`).

In particular, we should not need to be managing multiprocessing worker signals/PIDs ourselves in exp (and it should all be abstracted away by the process manager).

pmrowla:
POC demo of features that are now possible for us after this PR:
[asciicast](https://asciinema.org/a/WkaeNrrTRzZZyQAOR7PjB7KbC)

karajan1001 (approved):
LGTM

PR #7087, opened 2021-12-03
tests: remove redundant remote tests

efiop:
With introduction of https://github.com/iterative/dvc-plugins,
all important tests are being ran there, so there is no need
for dvc repo to run them again.

This PR also removes some redundant/broken remote tests.


efiop:
Will get to tweaking `--enable-*` flags in a followup, didn't touch them here since they are still kinda used for git_server tests.

PR #7089, opened 2021-12-03
exp show: Include `deps` and `outs`.

### LINKED ISSUE #6434: `exp show`: Include data files.
If an user modifies data files in between `dvc exp run` calls, the change is not reflected in the table generated by `dvc exp show`.

In [Studio](https://studio.iterative.ai) users can view information of the data files associated to each `experiment`. I think it would be nice to include similar information in the table generated by `dvc exp show`.

Ideally, the feature would have the same options available for `metrics` and `params` (i.e. `--include-files`/`--exclude-files`).

daavoo:

https://github.com/iterative/dvc.org/pull/3220

---

Use `repo.index.deps` / `repo.index.outs` to collect dependencies / outputss associated with each experiment.

Closes #6434

## What's considered a dep

Currently, anything in `repo.index.deps` that is not a param dependency or an imported `.dvc` (because if the `.dvc` is used in the pipeline it would be a duplicated column). 

Studio filters git tracked files but I think that showing those files (i.e. `src` deps) is also valuable.

I think that more complicated internal filtering (i.e. considering removing intermediate deps) it's not worthy and problematic when considering all use cases. 

The table can get noisy but we provide the `--only-changed` flag (we could consider making it the default) and new improved filtering  #7141 that should make it easy to customize the table.

## JSON output

For `--json` output, this P.R. adds new `deps` and `outs` fields:

```json
{
        "baseline": {
            "data": {
                "deps": {
                    "copy.py": {
                        "hash": "561f068574ab2a132d304dca3dd6510d",
                        "size": 310,
                        "nfiles": None,
                    }
                },
                "metrics": {"metrics.yaml": {"data": {"foo": 1}}},
                "outs": {
                    "model.pkl": {
                        "hash": "fb7792b6596fd12502dd132c0aba0568",
                        "size": 2000,
                        "nfiles": None,
                    }
                },
                "params": {"params.yaml": {"data": {"foo": 1}}},
                "queued": False,
                "running": False,
                "executor": None,
                "timestamp": None,
            }
        }
    }
```

## Table

For the table, it creates a new type of colored columns and shows the `hash` (let the debate begin).

After some testing, it looks like the optimal value for showing in the data column highly varies between use cases. 

Given the limitations of the CLI, I opted for showing `hash` as it's the value that allows, IMO, to easily identify differences between rows.

From example-get-started:

- `dvc exp show --only-changed`

Captura de pantalla 2021-12-22 a las 20 11 16

- `dvc exp show --all-branches --only-changed`

Captura de pantalla 2021-12-22 a las 20 11 45

Some deps might not be relevant, filtering with #7141 :

- ` dvc exp show --all-branches --only-changed --drop '.+prepared|model'`

Captura de pantalla 2021-12-22 a las 20 15 31

### DIFF
--- a/dvc/repo/experiments/show.py
+++ b/dvc/repo/experiments/show.py
@@ -39,6 +39,16 @@ def _collect_experiment_commit(
         if params:
             res["params"] = params
 
+        res["deps"] = {
+            dep.def_path: {
+                "hash": dep.hash_info.value,
+                "size": dep.meta.size,
+                "nfiles": dep.meta.nfiles,
+            }
+            for dep in repo.index.deps
+            if type(dep).__name__ != "ParamsDependency" and dep.is_in_repo
+        }
+
         res["queued"] = stash
         if running is not None and exp_rev in running:
             res["running"] = True
--- a/dvc/repo/experiments/show.py
+++ b/dvc/repo/experiments/show.py
@@ -23,6 +23,8 @@ def _collect_experiment_commit(
     running=None,
     onerror: Optional[Callable] = None,
 ):
+    from dvc.dependency import ParamsDependency, RepoDependency
+
     res: Dict[str, Optional[Any]] = defaultdict(dict)
     for rev in repo.brancher(revs=[exp_rev]):
         if rev == "workspace":
@@ -46,7 +48,17 @@ def _collect_experiment_commit(
                 "nfiles": dep.meta.nfiles,
             }
             for dep in repo.index.deps
-            if type(dep).__name__ != "ParamsDependency" and dep.is_in_repo
+            if not isinstance(dep, (ParamsDependency, RepoDependency))
+        }
+
+        res["outs"] = {
+            out.def_path: {
+                "hash": out.hash_info.value,
+                "size": out.meta.size,
+                "nfiles": out.meta.nfiles,
+            }
+            for out in repo.index.outs
+            if not (out.is_metric or out.is_plot)
         }
 
         res["queued"] = stash
@@ -121,6 +133,7 @@ def show(
     param_deps=False,
     onerror: Optional[Callable] = None,
 ):
+
     if onerror is None:
         onerror = onerror_collect
 
--- a/tests/func/experiments/test_show.py
+++ b/tests/func/experiments/test_show.py
@@ -36,7 +36,6 @@ def make_executor_info(**kwargs):
 
 
 def test_show_simple(tmp_dir, scm, dvc, exp_stage):
-    print(dvc.experiments.show()["workspace"])
     assert dvc.experiments.show()["workspace"] == {
         "baseline": {
             "data": {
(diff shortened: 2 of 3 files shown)


tests/func/experiments/test_show.py
```diff
@@ -530,26 +553,26 @@ def _get_rev_isotimestamp(rev):
     capsys.readouterr()
     assert main(["exp", "show", "--csv"]) == 0
     cap = capsys.readouterr()
+    data_dep = first(x for x in dvc.index.deps if "copy.py" in x.fspath)
+    data_hash = data_dep.hash_info.value[:7]
```
daavoo, line 557: Didn't know how to something like the `ANY` usage above but for the type of assertions bellow

dberenbaum:
Looks good so far!

Should deps come at the end (after params)? The noise would be less of an issue if they are on the right.

daavoo:
> Looks good so far!
> 
> Should deps come at the end (after params)? The noise would be less of an issue if they are on the right.

Done

dvc/repo/experiments/show.py
```diff
@@ -40,6 +40,16 @@ def _collect_experiment_commit(
...
             res["params"] = params
 
+        res["deps"] = {
+            dep.def_path: {
+                "hash": dep.hash_info.value,
+                "size": dep.meta.size,
+                "nfiles": dep.meta.nfiles,
+            }
+            for dep in repo.index.deps
+            if type(dep).__name__ != "ParamsDependency" and dep.is_in_repo
```
skshetry, line 50:
```suggestion
            if not isinstance(dep, ParamsDependency) and dep.is_in_repo
```

skshetry: @daavoo, do we really need `dep.is_in_repo`? It may be slow for "in-repo" dependencies. If we don't need it strictly, we could just do:
```suggestion
            if not isinstance(dep, (ParamsDependency, RepoDependency))
```

dtrifiro (dismissed):

dberenbaum (dismissed):

dtrifiro (approved):

dberenbaum:
> Should deps come at the end (after params)? The noise would be less of an issue if they are on the right.

By the way, Studio has data files in between metrics and params, right? Do you know why this order was preferred, and what do you think? I like them at the end, but consistency with Studio makes sense.

daavoo:
> By the way, Studio has data files in between metrics and params, right? Do you know why this order was preferred, and what do you think? I like them at the end, but consistency with Studio makes sense.

I don't know Studio preferences. I think that in our case it makes sense to introduce them at the end as they didn't exist before and table can't get noisy without `--only-changed` flag.

dberenbaum:
@daavoo It looks like the data columns are in a random order and it sometimes changes on repeated calls to `exp show`. Can we sort in alphabetical order or some other predictable way that will make sense to users?

jorgeorpinel:
> or some other predictable way

Curious: what was the sorting in the end? I think the more intuitive would be as defined in dvc.yaml

daavoo:
> > or some other predictable way
> 
> Curious: what was the sorting in the end? I think the most natural would be as defined in dvc.yaml

Alphabetical (it is what's currently used in Studio, afaik)

PR #7108, opened 2021-12-07
plots: make templates generation explicit

pared:
Fixes: #6389

Changes in this PR:

- `.dvc/plots` will no longer be created upon `dvc init`
- user has to explicitly call `dvc plots templates` to create and populate `.dvc/plots`

TODO
- [ ] docs

### DIFF
--- a/dvc/repo/plots/template.py
+++ b/dvc/repo/plots/template.py
@@ -1,6 +1,6 @@
 import json
 import os
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
 
 from funcy import cached_property
 
@@ -16,6 +16,15 @@ class NoFieldInDataError(DvcException):
     def __init__(self, field_name):
         super().__init__(
             f"Field '{field_name}' does not exist in provided data."
+        )
+
+
+class TemplateContentDoesNotMatch(DvcException):
+    def __init__(self, template_name: str, path: str):
+        super().__init__(
+            f"Template '{path}' already exists "
+            f"and its content is different than '{template_name}' content. "
+            "Remove it manually if you want to recreate it."
         )
 
 
@@ -533,17 +542,40 @@ class PlotTemplates:
     def __init__(self, dvc_dir):
         self.dvc_dir = dvc_dir
 
-    def init(self):
+    def init(
+        self, output: Optional[str] = None, targets: Optional[List] = None
+    ):
         from dvc.utils.fs import makedirs
 
-        makedirs(self.templates_dir, exist_ok=True)
-        for t in self.TEMPLATES:
-            self._dump(t())
+        output = output or self.templates_dir
 
-    def _dump(self, template):
-        path = os.path.join(self.templates_dir, template.filename)
-        with open(path, "w", encoding="utf-8") as fd:
-            fd.write(template.content)
+        makedirs(output, exist_ok=True)
+
+        if targets:
+            templates = [
+                template
+                for template in self.TEMPLATES
+                if template.DEFAULT_NAME in targets
+            ]
+        else:
+            templates = self.TEMPLATES
+
+        for template in templates:
+            self._dump(template(), output)
+
+    def _dump(self, template: Template, output: str):
+        path = os.path.join(output, template.filename)
+
+        if os.path.exists(path):
+            with open(path, "r", encoding="utf-8") as fd:
+                content = fd.read()
+            if content != template.content:
+                raise TemplateContentDoesNotMatch(
+                    template.DEFAULT_NAME or "", path
+                )
+        else:
+            with open(path, "w", encoding="utf-8") as fd:
+                fd.write(template.content)
 
     def load(self, template_name=None):
         if not template_name:
--- a/tests/unit/command/test_plots.py
+++ b/tests/unit/command/test_plots.py
@@ -4,9 +4,10 @@ import posixpath
 from pathlib import Path
 
 import pytest
+from funcy import pluck_attr
 
 from dvc.cli import parse_args
-from dvc.commands.plots import CmdPlotsDiff, CmdPlotsShow
+from dvc.commands.plots import CmdPlotsDiff, CmdPlotsShow, CmdPlotsTemplates
 
 
 @pytest.fixture
@@ -342,3 +343,38 @@ def test_show_json_requires_out(dvc, mocker, capsys):
     )
     cmd = cli_args.func(cli_args)
     assert cmd.run() == 0
+
+
+@pytest.mark.parametrize("target", (("t1"), (None)))
+def test_plots_templates(tmp_dir, dvc, mocker, capsys, target):
+    assert not os.path.exists(dvc.plots.templates.templates_dir)
+    mocker.patch(
+        "dvc.commands.plots.CmdPlotsTemplates.TEMPLATES_CHOICES",
+        ["t1", "t2"],
+    )
+
+    arguments = ["plots", "templates", "--out", "output"]
+    if target:
+        arguments += [target]
+
+    cli_args = parse_args(arguments)
+    assert cli_args.func == CmdPlotsTemplates
+
+    init_mock = mocker.patch("dvc.repo.plots.template.PlotTemplates.init")
+    cmd = cli_args.func(cli_args)
+
+    assert cmd.run() == 0
+    out, _ = capsys.readouterr()
+
+    init_mock.assert_called_once_with(
+        output=os.path.abspath("output"), targets=[target] if target else None
+    )
+    assert "Templates have been written into 'output'." in out
+
+
+def test_plots_templates_choices(tmp_dir, dvc):
+    from dvc.repo.plots.template import PlotTemplates
+
+    assert CmdPlotsTemplates.TEMPLATES_CHOICES == list(
+        pluck_attr("DEFAULT_NAME", PlotTemplates.TEMPLATES)
+    )
--- a/dvc/commands/plots.py
+++ b/dvc/commands/plots.py
@@ -140,6 +140,37 @@ class CmdPlotsModify(CmdPlots):
         return 0
 
 
+class CmdPlotsTemplates(CmdBase):
+    TEMPLATES_CHOICES = [
+        "simple",
+        "linear",
+        "confusion",
+        "confusion_normalized",
+        "scatter",
+        "smooth",
+    ]
+
+    def run(self):
+        import os
+
+        try:
+            out = (
+                os.path.join(os.getcwd(), self.args.out)
+                if self.args.out
+                else self.repo.plots.templates.templates_dir
+            )
+
+            targets = [self.args.target] if self.args.target else None
+            self.repo.plots.templates.init(output=out, targets=targets)
+            templates_path = os.path.relpath(out, os.getcwd())
+            ui.write(f"Templates have been written into '{templates_path}'.")
+
+            return 0
+        except DvcException:
+            logger.exception("")
+            return 1
+
+
 def add_parser(subparsers, parent_parser):
     PLOTS_HELP = (
         "Commands to visualize and compare plot metrics in structured files "
@@ -176,7 +207,8 @@ def add_parser(subparsers, parent_parser):
         "Shows all plots by default.",
     ).complete = completion.FILE
     _add_props_arguments(plots_show_parser)
-    _add_output_arguments(plots_show_parser)
+    _add_output_argument(plots_show_parser)
+    _add_ui_arguments(plots_show_parser)
     plots_show_parser.set_defaults(func=CmdPlotsShow)
 
     PLOTS_DIFF_HELP = (
@@ -211,7 +243,8 @@ def add_parser(subparsers, parent_parser):
         "revisions", nargs="*", default=None, help="Git commits to plot from"
     )
     _add_props_arguments(plots_diff_parser)
-    _add_output_arguments(plots_diff_parser)
+    _add_output_argument(plots_diff_parser)
+    _add_ui_arguments(plots_diff_parser)
     plots_diff_parser.set_defaults(func=CmdPlotsDiff)
 
     PLOTS_MODIFY_HELP = (
@@ -236,6 +269,27 @@ def add_parser(subparsers, parent_parser):
         help="Unset one or more display properties.",
     )
     plots_modify_parser.set_defaults(func=CmdPlotsModify)
+
+    TEMPLATES_HELP = (
+        "Write built-in plots templates to a directory "
+        "(.dvc/plots by default)."
+    )
+    plots_templates_parser = plots_subparsers.add_parser(
+        "templates",
+        parents=[parent_parser],
+        description=append_doc_link(TEMPLATES_HELP, "plots/templates"),
+        help=TEMPLATES_HELP,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+    plots_templates_parser.add_argument(
+        "target",
+        default=None,
+        nargs="?",
+        choices=CmdPlotsTemplates.TEMPLATES_CHOICES,
+        help="Template to write. Writes all templates by default.",
+    )
+    _add_output_argument(plots_templates_parser, typ="templates")
+    plots_templates_parser.set_defaults(func=CmdPlotsTemplates)
 
 
 def _add_props_arguments(parser):
@@ -276,14 +330,17 @@ def _add_props_arguments(parser):
     )
 
 
-def _add_output_arguments(parser):
+def _add_output_argument(parser, typ="plots"):
     parser.add_argument(
         "-o",
         "--out",
         default=None,
-        help="Destination path to save plots to",
+        help=f"Directory to save {typ} to.",
         metavar="<path>",
     ).complete = completion.DIR
+
+
+def _add_ui_arguments(parser):
     parser.add_argument(
         "--show-vega",
         action="store_true",
--- a/dvc/repo/init.py
+++ b/dvc/repo/init.py
@@ -75,13 +75,10 @@ def init(root_dir=os.curdir, no_scm=False, force=False, subdir=False):
 
     proj = Repo(root_dir)
 
-    proj.plots.templates.init()
-
     with proj.scm_context(autostage=True) as context:
         files = [
             config.files["repo"],
             dvcignore,
-            proj.plots.templates.templates_dir,
         ]
         ignore_file = context.scm.ignore_file
         if ignore_file:
--- a/tests/unit/repo/plots/test_templates.py
+++ b/tests/unit/repo/plots/test_templates.py
@@ -2,7 +2,13 @@ import os
 
 import pytest
 
-from dvc.repo.plots.template import TemplateNotFoundError
+from dvc.repo.plots.template import (
+    LinearTemplate,
+    PlotTemplates,
+    ScatterTemplate,
+    TemplateContentDoesNotMatch,
+    TemplateNotFoundError,
+)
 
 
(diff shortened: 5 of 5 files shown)


dberenbaum (commented):
So it seems like this has two uses:

1. Make it easy to see and modify templates.
2. Get updated templates if the ones currently in `.dvc/plots` are outdated.

Is that what you have in mind @pared?

dvc/command/plots.py
```diff
@@ -198,6 +210,16 @@ def add_parser(subparsers, parent_parser):
     )
     plots_modify_parser.set_defaults(func=CmdPlotsModify)
 
+    SHOW_HELP = "Dump plots templates."
```
dberenbaum, line 213:
```suggestion
    SHOW_HELP = "Write plots template JSON schemas to files."
```

Not sure if this wording is great, but a bit more description would be nice. cc @jorgeorpinel

dvc/command/plots.py
```diff
@@ -198,6 +210,16 @@ def add_parser(subparsers, parent_parser):
...
+    plots_dump_parser = plots_subparsers.add_parser(
```
dberenbaum, line 214: Does it make sense to be able to specify a different path to dump to? Maybe I am worried about overwriting what I have now, or I've made modifications to the defaults.

Also, can I dump just one template? This seems less necessary, but it might be nice to not have to generate all the templates if I only want one, especially if the number of templates grows.

dvc/command/plots.py
jorgeorpinel: I skimmed through #6389 and didn't really get why we're dumping plot template files. Will they no longer be included in `.dvc/` ? (Compatibility-breaking change BTW?)

jorgeorpinel: UPDATE: Duh it's explained in the PR description... OK ⌛

jorgeorpinel: Idk if `dump` is a good name in the first place. Sounds like what we're really doing is generating default templates. Why not `dvc plots templates`?

```suggestion
    SHOW_HELP = "Write built-in plots templates to a directory."
```

dvc/command/plots.py
pared: I think it makes sense to extend the functionality of templates, though it raises another questions.
1. If we want to specify particular template, do we want any means of obtaining templates available for this particular repo? I think so, that would make the user experience much better than circling between terminal and docs.
2. If we agree in 1, should `templates` have subcommands? eg (`write` and `list`)? That would make sense and be unified with our other commands.

jorgeorpinel:
What do we think about renaming `plots dump` to `plots templates`? Is "dumping" the right concept? Sounds like data-related to me.

shcheklein:
@pared @dberenbaum qq - how will the workflow with plots look like? how will this affect docs?

jorgeorpinel:
I think you'll have to "dump" the templates first before you can use `dvc plots` commands with data series files. So I guess those commands will throw an ERROR with exact hint of what to do if no templates exist. Or maybe they can include this command automatically on first use and just keep `plots dump` as an optional helper (I'd still rename it `plots templates` though).

dberenbaum:
You should not have to dump the templates. The default templates are stored internally, so these should not be breaking changes. The goal is to provide a mechanism to update plots without breaking current functionality.

`dvc init` will no longer generate templates in `.dvc/plots`. For new repos, plots will default to the templates in the installed dvc version instead of the dvc version used during `dvc init`. Templates with the same name will be prioritized in the order of 1) explicit paths, 2) `.dvc/plots`, 3) internal defaults. The `templates` command allows users with old `.dvc/plots` templates to upgrade, as well as maintaining the current ability to view/edit those templates.

@pared Please correct any mistakes/oversights!

dvc/command/plots.py
dberenbaum: When dumping `targets`, I was thinking of the default templates, not any file, since otherwise it's just copying a file, right? Is there a use case you had in mind? I see that it works to do something like `dvc plots template linear`, but it's not that clear from the help description that `targets` doesn't have to specify existing files. If I already have `linear.json` in `.dvc/plots`, will `dvc templates linear` dump the internal version or the one on file?

Having a way to list templates seems great. Maybe `--list` can be an option to avoid subcommands?

pared:
@dberenbaum that's precisely what this change is about

dvc/command/plots.py
pared: I think we should throw an exception if we are trying to dump a template that is different from the existing one under the given path. Do you want a new one? Fine, remove the old one manually. No chance of accidental removal of custom changes.

pared: > Is there a use case you had in mind?

Not exactly, just wanted to clarify

dberenbaum: So is this the expected workflow to upgrade outdated templates in `.dvc/plots`?

1. `rm -rf .dvc/plots`
2. `dvc plots templates`

pared: Yes

dberenbaum: > If I already have `linear.json` in `.dvc/plots`, will `dvc templates linear` dump the internal version or the one on file?

So this will fail because it will overwrite `.dvc/plots/linear.json`, correct? What about `dvc templates linear -o new_linear.json`? Which template version will it dump?

pared: > So this will fail because it will overwrite .dvc/plots/linear.json, correct?

It will fail only if the content of the template differs.

> dvc templates linear -o new_linear.json?

About that, another question: do we want to specify particular file? Currently it is implemented that out is dir to dump templates to (to be able to dump few at once). But I guess it makes more sense to specify particular templates and particular paths.

> Which template version will it dump?

The one that is loaded from DVC code, so the newest one.

dberenbaum: > It will fail only if the content of the template differs.

Sounds good, although it also seems like more work for you to check if the contents are the same. It also seems fine to me to fail anytime a file would be overwritten.

> do we want to specify particular file?

No, sorry, I got confused. No need to specify a particular file.

> > Which template version will it dump?
> 
> The one that is loaded from DVC code, so the newest one.

So will `dvc plots templates` always operate on the built-in templates? Maybe we should specify in the help description if so.

pared: > No, sorry, I got confused. No need to specify a particular file.

I probably should explaining it in option description 

> So will dvc plots templates always operate on the built-in templates? Maybe we should specify in the help description if so.

Will do

jorgeorpinel:
> The default templates are stored internally

"Internally" as in inside a DVC package installation directory? Something like /etc/dvc/plots ? oic

> (This) command allows users with old .dvc/plots templates to upgrade

Wait. So the system-level plots templates are being updated compared to the repo-level ones as well? If so how will users know that there's newer templates available and why/how to get them? We could just tell them to `rm -d .dvc/plots` instead.

Sorry, still unclear to me why you'd ever need to dump default templates if they're always available in the installation files. Ultimately we can document that location for each platform so people can just `cp` them.

jorgeorpinel:
> unclear to me why you'd ever need to dump default templates

I guess as a helper to avoid figuring out the system file location for `cp`ing? If so maybe `plots template` could accept argument(s) i.e. `default`, `scatter`, etc. so you only dump the specific one(s) you want to edit.

dberenbaum:
> "Internally" as in inside a DVC package installation directory? Something like /etc/dvc/plots ? oic

That was the initial plan, but it caused difficulties with packaging. Instead, they are composed within the code, and there's no path where the template exists as a JSON file.

> We could just tell them to `rm -d .dvc/plots` instead.

> I guess as a helper to avoid figuring out the system file location on your platform?

Yup, you got it. As mentioned above, it's unfortunately difficult to keep copies of the templates in some easy-to-find system file location, so there's no way to see the templates without some command like this.

> If we do keep this, maybe `plots template` could accept an optional argument i.e. `default` (would be the default), `scatter`, `smooth`, etc. so you can only dump the specific one you want to customize. Not sure

Yeah, this has been added as well.

jorgeorpinel:
> they are composed within the code, and there's no path where the template exists as a JSON file

oic!

No more doubts on my end. Except that I still wonder how/if we'll message users of existing projects on older DVC versions on how to upgrade and if so (or as docs in general) I suggest we recommend the `rm .dvc/plots` route instead of this command.

dberenbaum:
> I suggest we recommend the `rm .dvc/plots` route instead of this command.

Right, we were still working through this when the PR first opened. In fact, they will need to do this since `dvc plots templates` will fail to overwrite an old version of the template (see https://github.com/iterative/dvc/pull/7108#discussion_r772249726).

However, if they (or users with new repos) want to see what the current templates look like (or to modify them), they will need this command.

pared:
@dberenbaum The change is ready.
I have a feeling that specifying directory as `-o` is a bit clunky, but I don't have a better idea if we don't want to make users dump the templates one by one. And in the current state, there is a choice. Also, plot `show` itself also requires dir as output (since the static page is now generated as dir), so it does make sense.

dvc/command/plots.py
```diff
@@ -274,14 +337,17 @@ def _add_props_arguments(parser):
...
 
 
-def _add_output_arguments(parser):
+def _add_output_argument(parser, typ="plots"):
     parser.add_argument(
         "-o",
         "--out",
         default=None,
-        help="Destination path to save plots to",
+        help=f"Destination path to save {typ} to",
```
dberenbaum, line 345:
```suggestion
        help=f"Directory to save {typ} to.",
```
Since you mentioned it seems awkward that `-o` specifies a directory instead of the actual filepath, should we be more explicit about it in help?

dberenbaum (changes requested):
Looking good!

dvc/command/plots.py
```diff
@@ -235,6 +273,31 @@ def add_parser(subparsers, parent_parser):
...
+        "templates",
+        parents=[parent_parser],
+        description=append_doc_link(TEMPLATES_HELP, "plots/templates"),
+        help=SHOW_HELP,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+    plots_templates_parser.add_argument(
+        "targets",
+        nargs="*",
+        help="Templates to write. Writes all templates by default.",
```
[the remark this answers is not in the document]
pared: Sounds good, though I am not sure it will be able to handle `all` as default, I will see what I can do.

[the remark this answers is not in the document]
dberenbaum: I think it's fine to make `nargs="?"` (only allow one template) to simplify. What do you think? In that case, I would adjust the help text:

```suggestion
        help="Template to write. Writes all templates by default.",
```

dberenbaum:
@pared What's the status for this?

pared:
Got distracted with flexible plots and support, will try to finish it this week.

pared:
@dberenbaum 
So the problem with `choices` is that it will be hard to implement default value use case: we would need to provide empty string or `none` as one of the options. 

I have been playing a bit with it and the nicest looking thing that I was able to make work is currently at the master:

1. default:

```bash
(dvc3.8) ➜  dvc git:(6389_optional_plots) ✗ dvc plots templates                                                                   0.000s 
Templates have been written into '.dvc/plots'.
(dvc3.8) ➜  dvc git:(6389_optional_plots) ✗ tree .dvc/plots                                                                       0.000s 
.dvc/plots
├── confusion.json
├── confusion_normalized.json
├── linear.json
├── scatter.json
├── simple.json
└── smooth.json

0 directories, 6 files
(dvc3.8) ➜  dvc git:(6389_optional_plots) ✗      
```

2. multiple choices available:

```bash
(dvc3.8) ➜  dvc git:(6389_optional_plots) dvc plots templates simple linear                                                       0.000s 
Templates have been written into '.dvc/plots'.
(dvc3.8) ➜  dvc git:(6389_optional_plots) ✗ tree .dvc/plots                                                                       0.000s 
.dvc/plots
├── linear.json
└── simple.json

0 directories, 2 files
(dvc3.8) ➜  dvc git:(6389_optional_plots) ✗        
```

3. help:
```bash
(dvc3.8) ➜  dvc git:(6389_optional_plots) ✗ dvc plots templates --help                                                            0.000s 
usage: dvc plots templates [-h] [-q | -v] [-o <path>]
                           [['simple', 'linear', 'confusion', 'confusion_normalized', 'scatter', 'smooth']
                           [['simple', 'linear', 'confusion', 'confusion_normalized', 'scatter', 'smooth'] ...]]

Write built-in plots templates to a directory (.dvc/plots by default).
Documentation: <https://man.dvc.org/plots/templates>

positional arguments:
  ['simple', 'linear', 'confusion', 'confusion_normalized', 'scatter', 'smooth']  Templates to write. Writes all templates by default.

optional arguments:
  -h, --help            show this help message and exit
  -q, --quiet           Be quiet.
  -v, --verbose         Be verbose.
  -o <path>, --out <path>
                        Directory to save templates to.
(dvc3.8) ➜  dvc git:(6389_optional_plots) ✗             
```

4. wrong argument:
```bash
(dvc3.8) ➜  dvc git:(6389_optional_plots) dvc plots templates wrong_template simple                                               0.000s 
ERROR: argument ['simple', 'linear', 'confusion', 'confusion_normalized', 'scatter', 'smooth']: Incorrect template name: 'wrong_template' not in 'simple, linear, confusion, confusion_normalized, scatter, smooth'
usage: dvc plots templates [-h] [-q | -v] [-o <path>]
                           [['simple', 'linear', 'confusion', 'confusion_normalized', 'scatter', 'smooth']
                           [['simple', 'linear', 'confusion', 'confusion_normalized', 'scatter', 'smooth'] ...]]

Write built-in plots templates to a directory (.dvc/plots by default).
Documentation: <https://man.dvc.org/plots/templates>

positional arguments:
  ['simple', 'linear', 'confusion', 'confusion_normalized', 'scatter', 'smooth']  Templates to write. Writes all templates by default.

optional arguments:
  -h, --help            show this help message and exit
  -q, --quiet           Be quiet.
  -v, --verbose         Be verbose.
  -o <path>, --out <path>
                        Directory to save templates to.
(dvc3.8) ➜  dvc git:(6389_optional_plots)             
```

tests/unit/command/test_plots.py
```diff
@@ -342,3 +343,38 @@ def test_show_json_requires_out(dvc, mocker, capsys):
...
+    )
+    assert "Templates have been written into 'output'." in out
+
+
+def test_plots_templates_choices(tmp_dir, dvc):
+    from dvc.repo.plots.template import PlotTemplates
+
+    assert CmdPlotsTemplates.TEMPLATES_CHOICES == list(
+        pluck_attr("DEFAULT_NAME", PlotTemplates.TEMPLATES)
+    )
```
pared, line 380: Had to add the names of the templates manually to avoid importing `plots` on every `cmd` use. This test has been created to make sure `plots` module and `templates` command are in sync.

dvc/command/plots.py
```diff
@@ -235,6 +273,31 @@ def add_parser(subparsers, parent_parser):
...
+    )
+    _add_output_argument(plots_templates_parser, typ="templates")
+    plots_templates_parser.add_argument(
+        "--list",
```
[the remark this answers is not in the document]
pared: removed for now, check https://github.com/iterative/dvc/pull/7108#issuecomment-1023132442 for my current proposition for this problem

tests/unit/command/test_plots.py
dberenbaum: Ah, nice idea!

dberenbaum:
Looks good @pared!

I was thinking in https://github.com/iterative/dvc/pull/7108#discussion_r780351114 that if you only allow one template (which seems reasonable to me), then it could look like:

```
    plots_templates_parser.add_argument(
        "targets",
        nargs="?",
        choices=CmdPlotsTemplates.TEMPLATES_CHOICES,
        help="Template to write. Writes all templates by default.",
    )
```

I don't think that requires you to manually check for valid template names (although you still need `TEMPLATE_CHOICES`). It also keeps the outputs a little simpler.

Multiple choices not supported:

```
dvc plots templates simple linear
ERROR: unrecognized arguments: linear
usage: dvc plots templates [-h] [-q | -v] [-o <path>]
                           [{simple,linear,confusion,confusion_normalized,scatter,smooth}]

Write built-in plots templates to a directory (.dvc/plots by default).
Documentation: <https://man.dvc.org/plots/templates>

positional arguments:
  {simple,linear,confusion,confusion_normalized,scatter,smooth}
                        Template to write. Writes all templates by default.

optional arguments:
  -h, --help            show this help message and exit
  -q, --quiet           Be quiet.
  -v, --verbose         Be verbose.
  -o <path>, --out <path>
                        Directory to save templates to.
```

Help:

```
dvc plots templates --help
usage: dvc plots templates [-h] [-q | -v] [-o <path>]
                           [{simple,linear,confusion,confusion_normalized,scatter,smooth}]

Write built-in plots templates to a directory (.dvc/plots by default).
Documentation: <https://man.dvc.org/plots/templates>

positional arguments:
  {simple,linear,confusion,confusion_normalized,scatter,smooth}
                        Templates to write. Writes all templates by default.

optional arguments:
  -h, --help            show this help message and exit
  -q, --quiet           Be quiet.
  -v, --verbose         Be verbose.
  -o <path>, --out <path>
                        Directory to save templates to.
```

Wrong argument:

```
dvc plots templates wrong_template
ERROR: argument targets: invalid choice: 'wrong_template' (choose from 'simple', 'linear', 'confusion', 'confusion_normalized', 'scatter', 'smooth')
usage: dvc plots templates [-h] [-q | -v] [-o <path>]
                           [{simple,linear,confusion,confusion_normalized,scatter,smooth}]

Write built-in plots templates to a directory (.dvc/plots by default).
Documentation: <https://man.dvc.org/plots/templates>

positional arguments:
  {simple,linear,confusion,confusion_normalized,scatter,smooth}
                        Template to write. Writes all templates by default.

optional arguments:
  -h, --help            show this help message and exit
  -q, --quiet           Be quiet.
  -v, --verbose         Be verbose.
  -o <path>, --out <path>
                        Directory to save templates to.
```

Now that you have done it this way, I'm fine with either. Do you prefer keeping the option for multiple targets?

pared:
@dberenbaum

Ok, so I guess my research was to shallow - as I mentioned earlier, providing `choices` requires to add `None` or `''` as option among choices. However as I am experimenting now it seems to be the case only in case of `nargs='*'` and not `nargs='?'`. In that case I think there is no point in providing validation and metavar just to be able to provide multiple targets. I will modify the PR.

pared:
@dberenbaum So I though (as per my previous comment) that we will be unable to use choices without adding `None` or `''` as one of the options, which would make the help ugly. But if we make it as you proposed (with `?` instead of `*`) it does work. So I think we will be better off going your way. I made necessary change.

dberenbaum (approved):

dvc/repo/plots/template.py
```diff
@@ -533,15 +542,38 @@ def get_template(self, template_name: str):
...
+        for template in templates:
+            self._dump(template(), output)
+
+    def _dump(self, template: Template, output: str):
+        path = os.path.join(output, template.filename)
+
+        if os.path.exists(path):
+            with open(path, "r", encoding="utf-8") as fd:
+                content = fd.read()
+            if content != template.content:
```
daavoo, line 572: Probably is simpler this way but if this is False there is no need for the re-write below.

daavoo (approved):

dvc/repo/plots/template.py
pared: fixed

dberenbaum:
@pared Is there a corresponding issue or PR in [dvc.org?](https://github.com/iterative/dvc.org)?

pared:
@dberenbaum not yet, creating

PR #7111, opened 2021-12-07
fs: path: use ntpath/posixpath

efiop:
This is less error-prone than implementing these methods ourselves.

I think we can still do better generalizing this for fsspec, but this should work reliably in the meantime.

A downside here is that we don't get speedups from doing pure string manipulation instead of `os.path.join`.

Followup for #6976

### DIFF
--- a/dvc/fs/path.py
+++ b/dvc/fs/path.py
@@ -1,34 +1,44 @@
+import ntpath
 import posixpath
 
 
 class Path:
-    """
-    Class for operations on simple string paths.
-
-    This is meant to be very efficient and so doesn't use os.path,
-    doesn't have any notion of cwd and assumes that we are always
-    operating on absolute paths.
-    """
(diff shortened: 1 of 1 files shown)


dvc/fs/path.py
```diff
@@ -1,34 +1,44 @@
...
 class Path:
-    """
-    Class for operations on simple string paths.
+    def __init__(self, sep):
+        if sep == posixpath.sep:
+            self.flavour = posixpath
+        elif sep == ntpath.sep:
+            self.flavour = ntpath
+        else:
+            raise ValueError(f"unsupported separator '{sep}'")
```
efiop, line 12: have to do it this way because dvcfs/repofs/gitfs are still using os.path as sep 🙁 They will be adjusted in the future and `fs.path` will be revisited as well.

Issue #7118, opened 2021-12-09
stage add: StageExternalOutputsError on symlinked cache


endremborza:
# Bug Report

## Description

`os.path.realpath` resolves a symlink here: https://github.com/iterative/dvc/blob/82c5caee27d4b5591d4ab0b07fd1a73064ba8bff/dvc/output.py#L393 so if you want to add a stage where the output is already present but is cached it will think that it is external.

### Reproduce

1. dvc init
2. dvc config cache.dir /some/external/dir
3. dvc config cache.type  symlink
4. run and cache something
5. dvc stage add /w --out something that has already been cached

### Expected

`Outputout.is_in_repo` should not be `False` for a symlink to the cache in the repo

### Environment information

```console
$ dvc doctor
DVC version: 2.9.1 (pip)
---------------------------------
Platform: Python 3.8.10 on Linux-5.11.0-41-generic-x86_64-with-glibc2.29
Supports:
        hdfs (fsspec = 2021.10.1, pyarrow = 4.0.0),
        webhdfs (fsspec = 2021.10.1),
        http (aiohttp = 3.7.4.post0, aiohttp-retry = 2.4.6),
        https (aiohttp = 3.7.4.post0, aiohttp-retry = 2.4.6),
        s3 (s3fs = 2021.10.1, boto3 = 1.17.106),
        ssh (sshfs = 2021.8.1)
Cache types: hardlink, symlink
Cache directory: ext4 on /dev/sda2
Caches: local
Remotes: local, local
Workspace directory: ext4 on /dev/sda2
Repo: dvc, git
```

pared:
@endremborza good catch! Does it break some use case? How did you stumble upon it?

endremborza:
I am working on a tool that manages pipelines that use dvc and it updates stages in dvc pipelines based on parts of the code. currently, this breaks with symlinked cache.

pmrowla:
This is also reproducible in `dvc add` and it affects any symlink that points outside the repo directory (whether or not it's linked to the cache dir should be irrelevant in this scenario).

```
test-stage-add git:master ✓ ✩  py:dvc ❯ la
total 8.0K
drwxr-xr-x  6 pmrowla staff 192 Dec 15 11:01 .dvc
-rw-r--r--  1 pmrowla staff 139 Dec 15 10:51 .dvcignore
drwxr-xr-x 11 pmrowla staff 352 Dec 15 11:09 .git
[...]

PR #7119, opened 2021-12-09
add abspath option to Output.is_in_repo

endremborza:
Just a very quick check for a fix for #7118 without downloading, running or linting anything, sorry if this distracting, I just didn't want to set up the environment on my new machine again and hoped the CI will resolve whether this breaks any tests.

### DIFF
--- a/dvc/output.py
+++ b/dvc/output.py
@@ -388,8 +388,9 @@ class Output:
         if os.path.isabs(self.def_path):
             return False
 
-        return self.repo and path_isin(
-            os.path.realpath(self.fs_path), self.repo.root_dir
+        return self.repo and (
+            path_isin(os.path.realpath(self.fs_path), self.repo.root_dir)
+            or path_isin(os.path.abspath(self.fs_path), self.repo.root_dir)
         )
 
     @property
--- a/tests/func/test_add.py
+++ b/tests/func/test_add.py
@@ -842,6 +842,22 @@ def test_add_symlink_file(tmp_dir, dvc):
     dvc.add(os.path.join("dir", "foo"))
 
 
+def test_add_symlink_file_pointing_out(make_tmp_dir, tmp_dir, dvc):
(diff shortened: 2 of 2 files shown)


endremborza:
test break seems to be unrelated, and this change solves my problem with the symlinked cache of #7118 , so please indicate possible needed modifications.

pared:
Fixes #7118 
@endremborza what use case this check break? Maybe we need a test for that.

endremborza:
I don'r know what might be broken by this, the CI I'm pretty sure is not failing due to the change.

pmrowla:
I think @pared's point was that this PR should probably add a test case for the original issue (where adding a symlinked file to an alternate cache location caused some problem)

karajan1001:
> I don'r know what might be broken by this, the CI I'm pretty sure is not failing due to the change.

Let me rerun it, the CI is still not very stable 🤔 ?

pmrowla:
Thanks for adding the test @endremborza

This still needs a bit more investigation on our part, but this PR may be an acceptable short-term workaround.

It's clear that resolving the entire output path doesn't work in this case (since the symlink to a file outside the repo should be considered valid), so what we probably need to do is actually adjust the stage/output/dep paths so that they are properly aligned up to the resolved repo path (as described in https://github.com/iterative/dvc/pull/5954)

@skshetry you did the investigation on the original PR, WDYT?

pmrowla:
> @skshetry Pinging about the question above. 🙏

@dberenbaum this was discussed in a sprint meeting and the result was noted in the original issue (this bug will likely be resolved by the DVC data changes)

https://github.com/iterative/dvc/issues/7118#issuecomment-1040269129

skshetry:
I'm sorry to close this one, but it looks like this requires further investigation/research into the issue. Please follow issue #7668 for updates. 🙂

dberenbaum:
@iterative/dvc Is there opposition to revisiting this?

Issue #5489, opened 2021-02-17
Sharing multiple experiments


dberenbaum:
`dvc exp push` and `dvc exp pull` each take a single experiment. The experiments command is all about running multiple experiments, so most real-world use cases will involve wanting to push or pull multiple (or even all) experiments. For the initial proof of concept, it made sense to focus on one, but there should be ways to share multiple experiments at once. This could include:

- Specifying a list of experiment IDs
- Combining with `exp list` to get a particular subset of experiments to share
- Sharing all experiments in a branch
- Sharing all experiments from a given parent commit
- Sharing all experiments for the last n commits

dberenbaum:
Related (duplicate?): #4649

dberenbaum:
Seems like this could be handled using flags similar to those found in other dvc commands already:

```
  -w, --workspace       Push experiments derived from the current workspace.
  -a, --all-branches    Push experiments derived from the tip of all Git branches.
  -T, --all-tags        Push experiments derived from all Git tags.
  -A, --all-commits     Push experiments derived from all Git commits.
 ```

The same options could probably apply `dvc exp list`.

dberenbaum:
@pmrowla Are there any issues or complications with pushing multiple experiments that are different from pushing a single experiment?

pmrowla:
@dberenbaum the complication is regarding how to handle conflicts. I suppose if `--force` is given we can just assume the user wants to force push everything, but if `--force` is not used, do we abort as soon as there is any conflict, or do we behave like `dvc push` and continue trying to push everything else?

dberenbaum:
A user requested this in https://discord.com/channels/485586884165107732/485596304961962003/895329378110677003.

dberenbaum:
Moving the priority of this up since we are promoting the collaborative aspect of experiment versioning.

pmrowla:
See https://github.com/iterative/dvc/pull/7131#pullrequestreview-829717037 regarding potentially splitting this issue into separate tasks/steps

karajan1001:
According to https://github.com/iterative/dvc/pull/7131#pullrequestreview-829717037
Split this into four steps.
[ ]  add --rev and updated -n=-1 behavior to exp show
[ ] add the new flags (including -A/--all-commits, -a/--all-branches, --all-tags) to exp list
[ ] make all the flags above to support `pull/push`
[ ] make all the flags above to support `remove`

dberenbaum:
[...]

PR #7135, opened 2021-12-13
exp: split exp sub command (#5489)

karajan1001:
related to #5489
Current dvc/command/experiments.py is too long for switch between cmd class
and the argument parser.

Split the subcommand in dvc/command/experiments.py to individual files
```
deleted:    dvc/command/experiments.py
```

```
new file:   dvc/command/experiments/__init__.py
new file:   dvc/command/experiments/branch.py
new file:   dvc/command/experiments/data_sync.py
new file:   dvc/command/experiments/diff.py
new file:   dvc/command/experiments/gc.py
new file:   dvc/command/experiments/init.py
new file:   dvc/command/experiments/ls.py
new file:   dvc/command/experiments/remove.py
new file:   dvc/command/experiments/run.py
new file:   dvc/command/experiments/show.py
new file:   dvc/command/experiments/utils.py
```

Co-authored-by:


dvc/command/experiments/utils.py
```diff
@@ -0,0 +1,9 @@
+def append_doc_link(help_message, path):
```
pmrowla, line 1: Why is this being added here? It's already defined in `dvc.command.base`

pmrowla:
Also, if we are going to split everything up in this way, we might as well just separate `push` and `pull`. Also, `apply` and `branch` should definitely be separated, `apply` does not really have anything to do with `branch`.

dvc/command/experiments/gc.py
```diff
@@ -0,0 +1,128 @@
+import argparse
+import logging
+
+from dvc.command.repro import CmdRepro, append_doc_link
```
pmrowla, line 4:
```suggestion
from dvc.command.base import CmdBase, append_doc_link
```

dvc/command/experiments/gc.py
```diff
@@ -0,0 +1,128 @@
...
+
+def _raise_error_if_all_disabled(**kwargs):
+    if not any(kwargs.values()):
+        raise InvalidArgumentError(
+            "Either of `-w|--workspace`, `-a|--all-branches`, `-T|--all-tags` "
+            "or `--all-commits` needs to be set."
+        )
+
+
+class CmdExperimentsGC(CmdRepro):
```
pmrowla, line 19:
```suggestion
class CmdExperimentsGC(CmdBase):
```

Looks like this was a pre-existing bug? It didn't affect anything, but `dvc exp gc` should not need to inherit anything specific from `dvc repro`

dvc/command/experiments/run.py
```diff
@@ -0,0 +1,132 @@
+import argparse
+import logging
+
+from dvc.command import completion
+from dvc.command.repro import CmdRepro
+from dvc.command.repro import add_arguments as add_repro_arguments
+from dvc.command.repro import append_doc_link
```
pmrowla, line 7:
```suggestion
from dvc.command.base import append_doc_link
```

We should be explicit about where we are importing this method from, otherwise it looks like we are trying to import some `repro` specific implementation of `append_doc_link`.

pmrowla (changes requested):

pmrowla (approved):

PR #7141, opened 2021-12-13
exp show: Add --drop and --keep args

### LINKED ISSUE #7079: `dvc exp show`: Unify `--include`/`--exclude` and make it work with table columns.
## Current status

There are 4 different flags for filtering metrics and params in the table: `--include-metrics`/`--exclude-metrics`, `--include-params`/`--exclude-params`.

The filtering happens [at collection time](https://github.com/iterative/dvc/blob/7d68863ba21855bb5c238af7e1597e9a3607c0fe/dvc/command/experiments.py#L99).

---

## Current Drawbacks

With the [potential addition of `data`](https://github.com/iterative/dvc/issues/6434) as a new type of column it seems that following the same pattern would result in adding 2 new flags ( `--include-data`/`--exclude-data`) to a command already overloaded with flags.

Because the filtering happens at collection time, it is tricky to implement interactions with flags that operate with table columns (i.e. `--only-changed` currently ignores `--include-X`). In addition, there are a few chunks of code specific to this filtering that can only be tested via functional tests and are not reusable in other commands.

---

## Proposal

Unify all flags into `--include` / `--exclude` and make it work with table columns. Implement the filtering logic using `TabularData` functions.

This would allow users to filter any column of the table (i.e. for parallel coordinates users would be able to filter out the `Experiment` column) and could be also used to replace column-specific flags like `--no-timestamp` (`--exclude "Created"`).

Some command-specific chunks related to filtering at collection time could be removed.

### LINKED ISSUE #7080: `dvc exp show --html`: Improve experiment-line association.
The current implementation makes the `Experiment` column distracting/useless when there are lots of experiments:

imagen

One solution would be to have `on hover` information for each line and display the associated experiment there and drop the `Experiment` column in any case. This is blocked by an external product plotly/plotly.js#3012 

As a workaround for use cases where experiment-line association is not relevant, #7079 would allow users to remove the `Experiment` column from the plot.

Additional workarounds could be implemented by tweaking the general `layout` of the plot and/or the `ticktext` of the `Experiment` column based on the number of experiments to be shown.

### LINKED ISSUE #7083: `TabularData`: Add `subset` arg to `dropna`.
Mimicking https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.dropna.html 

Would be useful in `exp show --html` to drop rows with NaNs in all metrics and params columns. Currently, `how="all"` does not drop those rows because the `Experiment` column is not NaN.

daavoo:

---

- Removed `--include-metrics` / `--include-params` and
`--exclude-metrics` / `--exclude-params`.

- Removed `--no-timestamp`  (-> `--drop Created`)

- `--drop` and `--keep` operate directly on the table columns. 
  Support regex patterns.

- `--keep` does not perform any filtering. 
  It's only used to specify columns to keep despite the other filtering. 
  Previous `--include` behavior can be replicated with negative lookahead regex.

Closes #7079
Closes #7080
Closes #7083
Pre-requisite for #6434

---

Example:

<details>
<summary>`dvc exp show`</summary>

```
| Experiment              | Created      | loss    | accuracy   | train.batch_size   | train.hidden_units   | train.dropout   | train.num_epochs   | train.lr   | train.conv_activation   |
|-------------------------|--------------|---------|------------|--------------------|----------------------|-----------------|--------------------|------------|-------------------------|
| workspace               | -            | 0.31941 | 0.8997     | 128                | 64                   | 0.3             | 10                 | 0.01       | relu                    |
| main                    | Sep 14, 2021 | 0.26484 | 0.9038     | 128                | 64                   | 0.4             | 10                 | 0.001      | relu                    |
| ├── d599d35 [exp-4f1f0] | 06:12 PM     | 0.31941 | 0.8997     | 128                | 64                   | 0.3             | 10                 | 0.01       | relu                    |
| ├── c6d3ac6 [exp-eec4c] | 06:10 PM     | 0.24435 | 0.9135     | 128                | 64                   | 0.3             | 10                 | 0.001      | relu                    |
| └── 0d5e1ae [exp-25fae] | 06:03 PM     | 0.26133 | 0.9062     | 128                | 64                   | 0.4             | 10                 | 0.001      | relu                    |
```

</details>

<details>
<summary>`dvc exp show --only-changed`</summary>

```
| Experiment              | Created      | loss    | accuracy   | train.dropout   | train.lr   |
|-------------------------|--------------|---------|------------|-----------------|------------|
| workspace               | -            | 0.31941 | 0.8997     | 0.3             | 0.01       |
| main                    | Sep 14, 2021 | 0.26484 | 0.9038     | 0.4             | 0.001      |
| ├── d599d35 [exp-4f1f0] | 06:12 PM     | 0.31941 | 0.8997     | 0.3             | 0.01       |
| ├── c6d3ac6 [exp-eec4c] | 06:10 PM     | 0.24435 | 0.9135     | 0.3             | 0.001      |
| └── 0d5e1ae [exp-25fae] | 06:03 PM     | 0.26133 | 0.9062     | 0.4             | 0.001      |
```

</details>

</details>

<details>
<summary>`dvc exp show --only-changed --keep train.batch_size`</summary>

```
| Experiment              | Created      | loss    | accuracy   | train.batch_size   | train.dropout   | train.lr   |
|-------------------------|--------------|---------|------------|--------------------|-----------------|------------|
| workspace               | -            | 0.31941 | 0.8997     | 128                | 0.3             | 0.01       |
| main                    | Sep 14, 2021 | 0.26484 | 0.9038     | 128                | 0.4             | 0.001      |
| ├── d599d35 [exp-4f1f0] | 06:12 PM     | 0.31941 | 0.8997     | 128                | 0.3             | 0.01       |
| ├── c6d3ac6 [exp-eec4c] | 06:10 PM     | 0.24435 | 0.9135     | 128                | 0.3             | 0.001      |
| └── 0d5e1ae [exp-25fae] | 06:03 PM     | 0.26133 | 0.9062     | 128                | 0.4             | 0.001      |
```

</details>

<details>
<summary>`dvc exp show --drop train.*`</summary>

```
| Experiment              | Created      | loss    | accuracy   |
|-------------------------|--------------|---------|------------|
| workspace               | -            | 0.31941 | 0.8997     |
| main                    | Sep 14, 2021 | 0.26484 | 0.9038     |
| ├── d599d35 [exp-4f1f0] | 06:12 PM     | 0.31941 | 0.8997     |
| ├── c6d3ac6 [exp-eec4c] | 06:10 PM     | 0.24435 | 0.9135     |
| └── 0d5e1ae [exp-25fae] | 06:03 PM     | 0.26133 | 0.9062     |
```

</details>

<details>
<summary>`dvc exp show --drop train.* --keep train.dropout`</summary>

```
| Experiment              | Created      | loss    | accuracy   | train.dropout   |
|-------------------------|--------------|---------|------------|-----------------|
| workspace               | -            | 0.31941 | 0.8997     | 0.3             |
| main                    | Sep 14, 2021 | 0.26484 | 0.9038     | 0.4             |
| ├── d599d35 [exp-4f1f0] | 06:12 PM     | 0.31941 | 0.8997     | 0.3             |
| ├── c6d3ac6 [exp-eec4c] | 06:10 PM     | 0.24435 | 0.9135     | 0.3             |
| └── 0d5e1ae [exp-25fae] | 06:03 PM     | 0.26133 | 0.9062     | 0.4             |
```

</details>

<details>
<summary>`dvc exp show --only-changed --drop Created`</summary>

```
| Experiment              | loss    | accuracy   | train.dropout   | train.lr   |
|-------------------------|---------|------------|-----------------|------------|
| workspace               | 0.31941 | 0.8997     | 0.3             | 0.01       |
| main                    | 0.26484 | 0.9038     | 0.4             | 0.001      |
| ├── d599d35 [exp-4f1f0] | 0.31941 | 0.8997     | 0.3             | 0.01       |
| ├── c6d3ac6 [exp-eec4c] | 0.24435 | 0.9135     | 0.3             | 0.001      |
| └── 0d5e1ae [exp-25fae] | 0.26133 | 0.9062     | 0.4             | 0.001      |
```

</details>

For #7080

<details>
<summary>`dvc exp show --only-changed --drop Experiment --drop Created`</summary>

```
| loss    | accuracy   | train.dropout   | train.lr   |               
|---------|------------|-----------------|------------|
| 0.31941 | 0.8997     | 0.3             | 0.01       |
| 0.26484 | 0.9038     | 0.4             | 0.001      |
| 0.31941 | 0.8997     | 0.3             | 0.01       |
| 0.24435 | 0.9135     | 0.3             | 0.001      |
| 0.26133 | 0.9062     | 0.4             | 0.001      |
```

### DIFF
--- a/dvc/compare.py
+++ b/dvc/compare.py
@@ -11,6 +11,7 @@ from typing import (
     List,
     Mapping,
     MutableSequence,
+    Optional,
     Sequence,
     Set,
     Tuple,
@@ -38,10 +39,20 @@ class TabularData(MutableSequence[Sequence["CellT"]]):
         self._columns: Dict[str, Column] = {name: Column() for name in columns}
         self._keys: List[str] = list(columns)
         self._fill_value = fill_value
+        self._protected: Set[str] = set()
 
     @property
     def columns(self) -> List[Column]:
         return list(map(self.column, self.keys()))
+
+    def is_protected(self, col_name) -> bool:
+        return col_name in self._protected
+
+    def protect(self, *col_names: str):
+        self._protected.update(col_names)
+
+    def unprotect(self, *col_names: str):
+        self._protected = self._protected.difference(col_names)
 
     def column(self, name: str) -> Column:
         return self._columns[name]
@@ -126,9 +137,13 @@ class TabularData(MutableSequence[Sequence["CellT"]]):
         return len(self.columns), len(self)
 
     def drop(self, *col_names: str) -> None:
-        for col_name in col_names:
-            self._keys.remove(col_name)
-            self._columns.pop(col_name)
+        to_remove = set()
+        for col in col_names:
+            if not self.is_protected(col):
+                to_remove.add(col)
+        for col in to_remove:
+            self._keys.remove(col)
+            self._columns.pop(col)
 
     def rename(self, from_col_name: str, to_col_name: str) -> None:
         self._columns[to_col_name] = self._columns.pop(from_col_name)
@@ -199,7 +214,9 @@ class TabularData(MutableSequence[Sequence["CellT"]]):
             {k: self._columns[k][i] for k in keys} for i in range(len(self))
         ]
 
-    def dropna(self, axis: str = "rows", how="any"):
+    def dropna(
+        self, axis: str = "rows", how="any", subset: Optional[List] = None
+    ):
         if axis not in ["rows", "cols"]:
             raise ValueError(
                 f"Invalid 'axis' value {axis}."
@@ -211,13 +228,15 @@ class TabularData(MutableSequence[Sequence["CellT"]]):
             )
 
         match_line: Set = set()
-        match = True
+        match_any = True
         if how == "all":
-            match = False
+            match_any = False
 
         for n_row, row in enumerate(self):
             for n_col, col in enumerate(row):
-                if (col == self._fill_value) is match:
+                if subset and self.keys()[n_col] not in subset:
+                    continue
+                if (col == self._fill_value) is match_any:
                     if axis == "rows":
                         match_line.add(n_row)
                         break
@@ -244,7 +263,12 @@ class TabularData(MutableSequence[Sequence["CellT"]]):
         else:
             self.drop(*to_drop)
 
-    def drop_duplicates(self, axis: str = "rows", ignore_empty: bool = True):
+    def drop_duplicates(
+        self,
+        axis: str = "rows",
+        subset: Optional[List] = None,
+        ignore_empty: bool = True,
+    ):
         if axis not in ["rows", "cols"]:
             raise ValueError(
                 f"Invalid 'axis' value {axis}."
@@ -254,6 +278,8 @@ class TabularData(MutableSequence[Sequence["CellT"]]):
         if axis == "cols":
             cols_to_drop: List[str] = []
             for n_col, col in enumerate(self.columns):
+                if subset and self.keys()[n_col] not in subset:
+                    continue
                 # Cast to str because Text is not hashable error
                 unique_vals = {str(x) for x in col}
                 if ignore_empty and self._fill_value in unique_vals:
@@ -266,6 +292,13 @@ class TabularData(MutableSequence[Sequence["CellT"]]):
             unique_rows = []
             rows_to_drop: List[int] = []
             for n_row, row in enumerate(self):
+                if subset:
+                    row = [
+                        col
+                        for n_col, col in enumerate(row)
+                        if self.keys()[n_col] in subset
+                    ]
+
                 tuple_row = tuple(row)
                 if tuple_row in unique_rows:
                     rows_to_drop.append(n_row)
--- a/dvc/commands/experiments/show.py
+++ b/dvc/commands/experiments/show.py
@@ -1,10 +1,10 @@
 import argparse
 import logging
 import os
+import re
 from collections import Counter, OrderedDict, defaultdict
 from datetime import date, datetime
-from fnmatch import fnmatch
-from typing import TYPE_CHECKING, Dict, Iterable, Optional
+from typing import TYPE_CHECKING
 
 from funcy import lmap
 
@@ -28,64 +28,6 @@ FILL_VALUE_ERRORED = "!"
 logger = logging.getLogger(__name__)
 
 
-def _filter_name(names, label, filter_strs):
-    ret = defaultdict(dict)
-    path_filters = defaultdict(list)
-
-    for filter_s in filter_strs:
-        path, _, name = filter_s.rpartition(":")
-        path_filters[path].append(name)
-
-    for path, filters in path_filters.items():
-        if path:
-            match_paths = [path]
-        else:
-            match_paths = names.keys()
-        for match_path in match_paths:
-            for f in filters:
-                matches = [
-                    name for name in names[match_path] if fnmatch(name, f)
-                ]
-                if not matches:
-                    raise InvalidArgumentError(
-                        f"'{f}' does not match any known {label}"
-                    )
-                ret[match_path].update({match: None for match in matches})
-
-    return ret
-
-
-def _filter_names(
-    names: Dict[str, Dict[str, None]],
-    label: str,
-    include: Optional[Iterable],
-    exclude: Optional[Iterable],
-):
-    if include and exclude:
-        intersection = set(include) & set(exclude)
-        if intersection:
-            values = ", ".join(intersection)
-            raise InvalidArgumentError(
-                f"'{values}' specified in both --include-{label} and"
-                f" --exclude-{label}"
-            )
-
-    if include:
-        ret = _filter_name(names, label, include)
-    else:
-        ret = names
-
-    if exclude:
-        to_remove = _filter_name(names, label, exclude)
-        for path in to_remove:
-            if path in ret:
-                for key in to_remove[path]:
-                    if key in ret[path]:
-                        del ret[path][key]
-
-    return ret
-
-
 def _update_names(names, items):
     for name, item in items:
         item = item.get("data", {})
@@ -103,18 +45,6 @@ def _collect_names(all_experiments, **kwargs):
             exp = exp_data.get("data", {})
             _update_names(metric_names, exp.get("metrics", {}).items())
             _update_names(param_names, exp.get("params", {}).items())
-    metric_names = _filter_names(
-        metric_names,
-        "metrics",
-        kwargs.get("include_metrics"),
-        kwargs.get("exclude_metrics"),
-    )
-    param_names = _filter_names(
-        param_names,
-        "params",
-        kwargs.get("include_params"),
-        kwargs.get("exclude_params"),
-    )
 
     return metric_names, param_names
 
@@ -308,17 +238,6 @@ def _extend_row(row, names, items, precision, fill_value=FILL_VALUE):
             row.append(ui.rich_text(str(_format_field(value, precision))))
 
 
-def _parse_filter_list(param_list):
-    ret = []
-    for param_str in param_list:
-        path, _, param_str = param_str.rpartition(":")
-        if path:
-            ret.extend(f"{path}:{param}" for param in param_str.split(","))
-        else:
-            ret.extend(param_str.split(","))
-    return ret
-
-
 def experiments_table(
     all_experiments,
     headers,
@@ -381,8 +300,9 @@ def baseline_styler(typ):
 
 def show_experiments(
     all_experiments,
+    keep=None,
+    drop=None,
     pager=True,
-    no_timestamp=False,
     csv=False,
     markdown=False,
     pcp=False,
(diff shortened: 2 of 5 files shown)


pmrowla:
How does filtering with the new options work if there is a metric and parameter with the same name?

karajan1001:
Cool, I don't like the old `_filter_name` and `_filter_names`.

dvc/compare.py
```diff
@@ -124,10 +126,17 @@ def __len__(self) -> int:
     def shape(self) -> Tuple[int, int]:
         return len(self.columns), len(self)
 
-    def drop(self, *col_names: str) -> None:
-        for col_name in col_names:
-            self._keys.remove(col_name)
-            self._columns.pop(col_name)
+    def drop(self, *col_names: str, keep: Optional[List[str]] = None) -> None:
```
karajan1001, line 129: Maybe it is better to split this into two functions? A pure drop or `__del__` and a `drop_column` it is quite weird for a `drop` function to have a `keep` option for me.

daavoo: Because I added support for fnmatch , It felt like there should be a way to have a keep.

Perhaps if instead of fnmatch we support regex we could remove keep argument

daavoo:
> How does filtering with the new options work if there is a metric and parameter with the same name?

User need to indicate the pattern matching the final column. So if there is a shared name, user would need to `--drop params.yaml:foo

dvc/compare.py
karajan1001: So, it is a `drop` supporting `wild match` 🤔 . Another way is to add a new `protect` function to protect some of the columns and can set once and effect in `drop`, `dropna` and `drop_duplicate`.

daavoo: I like the `protect` idea

dberenbaum (dismissed):
Looks great! Nice simplification of the UI.

The big question is whether removing `--include` flags will make it too hard to filter the table. Now that we have `--only-changed`, the hope is that this will cover a lot that filtering automatically (along with `--drop`). 

We haven't had much feedback recently on the table filtering, so let's try it and see what people think.

daavoo:
> Looks great! Nice simplification of the UI.
> 
> The big question is whether removing `--include` flags will make it too hard to filter the table. Now that we have `--only-changed`, the hope is that this will cover a lot that filtering automatically (along with `--drop`).
> 
> We haven't had much feedback recently on the table filtering, so let's try it and see what people think.

I was thinking that if we support `regex` matching instead of `fnmatch` there would be no need of `--keep` and users could perform similar behavior to previous `--include`.

I'm Ok with adding regex and just examples in docs. WDYT?

dberenbaum:
> I was thinking that if we support `regex` matching instead of `fnmatch` there would be no need of `--keep` and users could perform similar behavior to previous `--include`.
> 
> I'm Ok with adding regex and just examples in docs. @dberenbaum WDYT?

I knew we had this discussion before 😄 : https://github.com/iterative/dvc/pull/5720#issuecomment-810446190.

No strong preference from me. Not sure if regex is less consistent with the rest of the dvc cli, but obviously it's more powerful, so 🤷 .

dvc/compare.py
daavoo: Added `protected` property (get only) and `protect`/`unprotect` methods. Removed `keep` arg from `drop`

pmrowla:
@daavoo this needs a rebase, exp commands have been split into their own files (so `dvc/commands/experiments/show.py`) since there was getting to be way too much stuff in `commands/experiments.py`

daavoo:
> @daavoo this needs a rebase, exp commands have been split into their own files (so `dvc/commands/experiments/show.py`) since there was getting to be way too much stuff in `commands/experiments.py`

Done. I had already done that one but this required a new one because of the parallel coordinates merge

dvc/compare.py
daavoo: ping @karajan1001 , could you review?

jorgeorpinel:
Nice. On first sight it's not clear to me what the difference between include and keep are but we'll worry about that in the docs PR. Just one Q: Does the order in which filters are given matter? I.e. if I keep sometihng and then drop something what happens? Thanks

jorgeorpinel:
UPDATE: I guess it's already explained in https://github.com/iterative/dvc.org/pull/3112 so nvmd.

dvc/compare.py
```diff
@@ -37,11 +38,24 @@ def __init__(self, columns: Sequence[str], fill_value: str = ""):
...
     @property
     def columns(self) -> List[Column]:
         return list(map(self.column, self.keys()))
 
+    @property
+    def protected(self) -> Set[str]:
+        return self._protected
+
+    def protect(self, *col_names: str):
+        for col_name in col_names:
```
karajan1001, line 52: I guess the built-in methods are more efficiency.
```
self._protected.update(col_name)
```

dvc/compare.py
```diff
@@ -37,11 +38,24 @@ def __init__(self, columns: Sequence[str], fill_value: str = ""):
...
         self._keys: List[str] = list(columns)
         self._fill_value = fill_value
+        self._protected: Set[str] = set()
 
     @property
     def columns(self) -> List[Column]:
         return list(map(self.column, self.keys()))
 
+    @property
+    def protected(self) -> Set[str]:
```
karajan1001, line 48: For myself, I would use an `is_protected` function here and make the `re.match` logic inside it.  Decouple them from outside.

dvc/compare.py
```diff
@@ -222,6 +246,8 @@ def dropna(self, axis: str = "rows", how="any"):
 
         for n_row, row in enumerate(self):
             for n_col, col in enumerate(row):
+                if subset and self.keys()[n_col] not in subset:
```
karajan1001, line 249: Can we change the loop order here? Brings the `subset` judgment to the outer loop? Besides, I guess we can just iterate over the cols in the subset?

daavoo: Not sure how we could invert the loop order and keep the functionality below.

dvc/compare.py
```diff
@@ -125,9 +138,15 @@ def shape(self) -> Tuple[int, int]:
...
-            self._columns.pop(col_name)
+        to_remove = set()
+        for key in self._keys:
+            if self.is_protected(key):
+                continue
+            if any(match(d, key) for d in col_names):
+                to_remove.add(key)
+        for col in to_remove:
+            self._keys.remove(col)
+            self._columns.pop(col)
```
skshetry, line 149: Does pandas or any dataframe library support this regex filtering?
Would it be possible to do `regex.match` from outside? something as follows:

```python
cols = match_ordered(keep, td.keys())
td.project(*cols)
```

I prefer not to complicate API too much here and keep API look like that of dataframe library. All the business logics that don't make sense in a library should be better handled outside.

daavoo: Makes sense. Will move `re.match` logic to `exp show`

dvc/command/experiments/show.py
```diff
@@ -473,7 +383,22 @@ def show_experiments(
         td.drop_duplicates("cols", ignore_empty=False)
 
     if pcp:
-        td.dropna("rows", how="all")
+        if drop is None:
```
karajan1001, line 386: For myself, I would prefer first to do the regex match and then add `Created` to the drop_set.

dvc/command/experiments/show.py
```diff
@@ -473,7 +383,22 @@ def show_experiments(
...
+            drop = "Created"
+        else:
+            drop += "|Created"
+
+    if drop is not None:
+        td.drop(*[col for col in td.keys() if re.match(drop, col)])
```
karajan1001, line 392: `subset` didn't work on the `drop` here? By the way, what is the priority relationship between these args? And I would prefer to give a Set to it as the order doesn't matter here.

dvc/command/experiments/show.py
daavoo: Done

dvc/command/experiments/show.py
daavoo: > subset didn't work on the drop here? 

The regex matching was moved outside per https://github.com/iterative/dvc/pull/7141#discussion_r777537175 . 

> By the way, what is the priority relationship between these args?

If combined `--only-changed` has the least priority, `--drop` comes next, and
`--keep` has the last word (from docs)

> And I would prefer to give a Set to it as the order doesn't matter here.
Done

dberenbaum (approved):

dvc/compare.py
```diff
@@ -126,9 +137,13 @@ def shape(self) -> Tuple[int, int]:
         return len(self.columns), len(self)
 
     def drop(self, *col_names: str) -> None:
-        for col_name in col_names:
-            self._keys.remove(col_name)
-            self._columns.pop(col_name)
+        to_remove = set()
+        for col in col_names:
+            if not self.is_protected(col):
+                to_remove.add(col)
```
karajan1001, line 143: How about remove `to_remove` and directly:
```python
        for col in col_names:
            if not self.is_protected(col):
                self._keys.remove(col)
                self._columns.pop(col)

```

dvc/compare.py
```diff
@@ -211,13 +228,15 @@ def dropna(self, axis: str = "rows", how="any"):
...
-        match = True
+        match_any = True
         if how == "all":
-            match = False
+            match_any = False
 
         for n_row, row in enumerate(self):
             for n_col, col in enumerate(row):
-                if (col == self._fill_value) is match:
+                if subset and self.keys()[n_col] not in subset:
```
karajan1001, line 237: `in` operation is `O(1)` for `Set` but `O(n)` for `List`

karajan1001 (approved):
LGTM, in the most parts.

dvc/compare.py
karajan1001: It merged automatically.

daavoo: #7299

PR #7142, opened 2021-12-13
exp show: Alternate row background color

### LINKED ISSUE #7128: exp show: alternate background colors slightly to make navigation easier
It would be way easier to navigate in the table if we had different background for items nearby, e.g. VSCode:

Screen Shot 2021-12-10 at 7 12 51 PM

daavoo:

Closes #7128 

Kind of tricky to get a visually pleasant style (related https://github.com/willmcgugan/rich/discussions/363)

Here is the current state:

Captura de pantalla 2021-12-14 a las 18 14 41

Captura de pantalla 2021-12-14 a las 18 12 55

### DIFF
--- a/dvc/command/experiments/show.py
+++ b/dvc/command/experiments/show.py
@@ -373,7 +373,7 @@ def prepare_exp_id(kwargs) -> "RichText":
 
 
 def baseline_styler(typ):
-    return {"style": "bold"} if typ == "baseline" else {}
+    return {"style": "bold"} if typ == "baseline" else {"style": "default"}
 
 
 def show_experiments(
@@ -437,6 +437,9 @@ def show_experiments(
             td.drop(col)
 
     row_styles = lmap(baseline_styler, td.column("typ"))
+    for n, row_style in enumerate(row_styles):
+        if n % 2 != 0:
+            row_style["style"] += " on grey23"
 
     if not csv:
         merge_headers = ["Experiment", "rev", "typ", "parent"]
@@ -471,7 +474,7 @@ def show_experiments(
 
     td.render(
         pager=pager,
-        borders=True,
+        borders="horizontals",
         rich_table=True,
         header_styles=styles,
         row_styles=row_styles,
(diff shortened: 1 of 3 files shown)


pared (approved):

dberenbaum:
Nice work, @daavoo! Have you tried dropping the borders? Wondering if it's helpful or distracting as is.

@shcheklein what do you think?

shcheklein:
I like it, and box simple makes total sense to me. 

Also, I wonder if we need bold font (or is it your local settings, @daavoo ?)

daavoo:
> Nice work, @daavoo! Have you tried dropping the borders? Wondering if it's helpful or distracting as is.
> 
> @shcheklein what do you think?

Updated description showing `box.SIMPLE` mode (removing borders).

daavoo:
> Also, I wonder if we need bold font (or is it your local settings, @daavoo ?)

We set bold font for `baseline` experiments (no differences in the table I show which is using `--all-tags`). Here is how it looks different with no baseline experiments:

Captura de pantalla 2021-12-14 a las 18 03 50

dvc/command/experiments/show.py
```diff
@@ -373,7 +373,7 @@ def prepare_exp_id(kwargs) -> "RichText":
 
 
 def baseline_styler(typ):
-    return {"style": "bold"} if typ == "baseline" else {}
+    return {"style": "bold"} if typ == "baseline" else {"style": "default"}
```
skshetry, line 376: hmm, I don't see a "default" style in https://github.com/willmcgugan/rich/blob/2ba277ac54a153d3030b496698133ec2e4f67e1d/rich/default_styles.py. Does this even work?

daavoo: It works but it is was only added to make it easier to set `on {bcolor}`

skshetry: @daavoo, can we remove it now?

daavoo: Yes. I Forgot to do it in #7212

PR #7159, opened 2021-12-15
Add config  option `plots.auto_open` 

### LINKED ISSUE #6091: plots diff/show: configure --open flag
Enable configuration of `dvc plots diff/show --open` flag at the project level.

daavoo:

Closes #6091
Closes iterative/dvclive#201

`dvc.org` P.R: https://github.com/iterative/dvc.org/pull/3101

### DIFF
--- a/dvc/command/plots.py
+++ b/dvc/command/plots.py
@@ -96,6 +96,12 @@ class CmdPlots(CmdBase):
             ui.write(index_path.as_uri())
             auto_open = self.repo.config["plots"].get("auto_open", False)
             if self.args.open or auto_open:
+                if not auto_open:
+                    ui.write(
+                        "To enable auto opening, you can run:\n"
+                        "\n"
+                        "\tdvc config plots.auto_open true"
+                    )
                 return ui.open_browser(index_path)
 
             return 0
--- a/dvc/command/plots.py
+++ b/dvc/command/plots.py
@@ -94,8 +94,8 @@ class CmdPlots(CmdBase):
             )
 
             ui.write(index_path.as_uri())
-
-            if self.args.open:
+            auto_open = self.repo.config["plots"].get("auto_open", False)
+            if self.args.open or auto_open:
                 return ui.open_browser(index_path)
 
             return 0
--- a/dvc/repo/live.py
+++ b/dvc/repo/live.py
@@ -20,7 +20,6 @@ def webbrowser_open(url: str) -> int:
 
 
 def create_summary(out):
-
     assert out.live and out.live["html"]
 
     metrics, plots = out.repo.live.show(out.fs_path)
@@ -31,8 +30,8 @@ def create_summary(out):
     index_path = render(
         out.repo, renderers, metrics=metrics, path=html_path, refresh_seconds=5
     )
-
-    webbrowser_open(index_path)
+    if out.repo.config["plots"].get("auto_open", False):
+        webbrowser_open(index_path)
 
 
 def summary_fs_path(out: "Output") -> Optional[str]:
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -256,7 +256,10 @@ SCHEMA = {
         # enabled by default. It's of no use, kept for backward compatibility.
         Optional("parametrization", default=True): Bool,
     },
-    "plots": {"html_template": str},
+    "plots": {
+        "html_template": str,
+        Optional("auto_open", default=False): Bool,
+    },
     "exp": {
         "code": str,
         "data": str,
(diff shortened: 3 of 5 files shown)


pared (approved):

rogermparent (approved):
Just tried this out in CLI and VSCode and successfully didn't have a tab open! As long as the code is good on the DVC end, merging this will fix our issue.

dvc/command/plots.py
```diff
@@ -94,8 +96,12 @@ def run(self):
             )
 
             ui.write(index_path.as_uri())
-
-            if self.args.open:
+            auto_open = to_bool(
+                Config(validate=False)
+                .get("plots", {})
+                .get("auto_open", "false")
+            )
```
pmrowla, line 103: Why are we re-parsing configs here? This should use the `repo.config` instance:

```py
self.repo.config["plots"].get("auto_open", False)
```

dvc/repo/live.py
```diff
@@ -31,8 +34,11 @@ def create_summary(out):
     index_path = render(
         out.repo, renderers, metrics=metrics, path=html_path, refresh_seconds=5
     )
-
-    webbrowser_open(index_path)
+    auto_open = to_bool(
+        Config(validate=False).get("plots", {}).get("auto_open", "false")
+    )
```
pmrowla, line 39: Same as the other comment, we should not be re-parsing any configs here.

pmrowla (changes requested):

jorgeorpinel:
Hi. Q: where will `DVCLIVE_OPEN` be documented? Is there a PR in dvc.org? Thanks

daavoo:
> Hi. Q: where will `DVCLIVE_OPEN` be documented? Is there a PR in dvc.org? Thanks

Not sure if It should be documented at all as It is intended for VSCode internal usage

dvc/command/plots.py
daavoo: Thanks! didn't know. First place I found config usage was https://github.com/iterative/dvc/blob/30d6839a5f471fb02a7a2dcbd6a6616b71277762/dvc/command/daemon.py and just copy from there

daavoo: Is there a reason for doing it differently in `daemon`?

pmrowla: It's because it inherits from `CmdBaseNoRepo` (for commands that can be run outside of a DVC repo), so it does not have `self.repo.config`

```
class CmdDaemonBase(CmdBaseNoRepo):
```

shcheklein:
Thanks @daavoo for taking care of this quick! Should we name things in a similar way - DVC_PLOTS_AUTO_OPEN ?

dvc/repo/live.py
```diff
@@ -31,8 +32,9 @@ def create_summary(out):
     index_path = render(
         out.repo, renderers, metrics=metrics, path=html_path, refresh_seconds=5
     )
-
-    webbrowser_open(index_path)
+    auto_open = out.repo.config["plots"].get("auto_open", False)
+    if auto_open and int(os.environ.get(DVCLIVE_OPEN, "1")):
```
pmrowla, line 36: Use `utils.env2bool` for this, it covers more of the commonly used true/false-y env var cases besides `"1"`

https://github.com/iterative/dvc/blob/30d6839a5f471fb02a7a2dcbd6a6616b71277762/dvc/utils/__init__.py#L335

pmrowla (changes requested):

dvc/repo/live.py
skshetry, line 36: `plots.auto_open` vs `DVCLIVE_OPEN` are not symmetrical though. WDYT of `DVC_PLOTS_OPEN`?

skshetry: Another thought: should we be more generic and implement `core.no_browser` and `DVC_NO_BROWSER` instead?

skshetry: I guess it's a thing to discuss in #6091 then here.

daavoo: The env var, in this case, was added for VSCode specifically, not really for end-user usage (however, it's not clear yet if they will need it or not). The asymmetry was intentional in order to allow VSCode to reactive DVCLIVE HTML even if user had `plots.auto_open` enabled. 

If going for the general approach, is there a reason to have both a `dvc config` field and an env var for the same purpose?

rogermparent: We've decided on the VSCode end, at least for now, to respect the user's configuration for browser opening. Something like `DVC_NO_BROWSER` could be useful in the future for us and/or other integrations should we get user requests for a VSCode extension feature to disable browser opening outside the config.

PR #7190, opened 2021-12-22
exp show: Rename `html` to `pcp`

### LINKED ISSUE #7165: `exp show`: Rename / Split to separate command the current `--html`
> My biggest question is whether this belongs in `dvc exp show` or might be better as a separate command, like `dvc exp plot`? A `dvc exp plot` command could also potentially show other things, like `dvc plots diff` between experiments, or checkpoint trends (like dvclive plots).

_Originally posted by @dberenbaum in https://github.com/iterative/dvc/issues/6933#issuecomment-973159973_

> Why something as generic and conceptually disconnected as "HTML"? How about `--parallel-coordinates-plot`/`--pcp`? Or a separate command as @dberenbaum mentioned

_Originally posted by @jorgeorpinel in https://github.com/iterative/dvc/issues/6933#issuecomment-996059215_

> To be honest `--html` for me would mean an HTML table, not plot. I would consider renaming this to be explicit about its meaning. I think `--plot` is a bit more explicit. But can limit us later if we decide to introduce more types of plots. Probably `--pcp` is the best unfortunately.

_Originally posted by @shcheklein in https://github.com/iterative/dvc/issues/6933#issuecomment-996459692_

daavoo:
Closes #7165


tests/func/experiments/test_show.py
```diff
@@ -598,6 +598,7 @@ def test_show_only_changed(tmp_dir, dvc, scm, capsys):
     assert "bar" not in cap.out
 
 
+@pytest.mark.pcp
```
daavoo, line 601: For this feature, I have been using this custom marker locally. Does it make sense for you to include this kind of custom markers in the repo?

efiop: @daavoo Could you clarify the purpose?

daavoo: For some features like the pcp (parallel coordinates plot), the related tests are spread across different files and I use the custom marker to have a quick way of running all the tests related to the feature.

daavoo: So during development, I run `pytest -m pcp`

daavoo: The idea to include in repo is that if someone touches this feature in the future, related tests can be run without knowing where all the related files are located.

casperdcl: as an alternative if the tests are named semi-consistently then you can do `pytest -k "pcp or parallel_coordinates"` and markers aren't needed.

daavoo: `-k` is a good alternative, for sure. 

IMO, the issue with `-k` is that a feature might be affected by some function with unit tests that don't make sense to name after the feature.

`-m` is explicit, people can see it but `-k` patterns (unless documented) might be tricky if not everybody is aware of the usage.

dberenbaum (approved):

dvc/command/experiments/show.py
```diff
@@ -472,20 +472,20 @@ def show_experiments(
...
             str(x).encode("ascii", "ignore").strip().decode()
             for x in td.column("Experiment")
         ]
         out = kwargs.get("out") or "dvc_plots"
-        html_args["output_path"] = (Path.cwd() / out).resolve()
-        html_args["color_by"] = kwargs.get("sort_by") or "Experiment"
+        pcp_args["output_path"] = (Path.cwd() / out).resolve()
+        pcp_args["color_by"] = kwargs.get("sort_by") or "Experiment"
 
     td.render(
```
skshetry, line 490: This `table.render()` does not make sense for rendering anything other than a table. I'd keep these separate and just use `td.to_parallel_coordinates` helper function directly here instead of extending `render`.

dvc/command/experiments/show.py
```diff
@@ -472,20 +472,20 @@ def show_experiments(
...
         td.dropna("rows", how="all")
         td.column("Experiment")[:] = [
             # remove tree characters
             str(x).encode("ascii", "ignore").strip().decode()
             for x in td.column("Experiment")
         ]
         out = kwargs.get("out") or "dvc_plots"
-        html_args["output_path"] = (Path.cwd() / out).resolve()
-        html_args["color_by"] = kwargs.get("sort_by") or "Experiment"
+        pcp_args["output_path"] = (Path.cwd() / out).resolve()
```
skshetry, line 487:
```suggestion
        pcp_args["output_path"] = Path(out).resolve()
```

Or,
```suggestion
        pcp_args["output_path"] = os.path.abspath(out)
```

tests/func/experiments/test_show.py
casperdcl: But we're not talking about `-m` versus `-k` in general, right? Just in the case of PCP tests it seems `pytest -k "pcp or parallel_coordinate"` is future-proof & intuitive for devs working on it?

dvc/compare.py
```diff
@@ -153,7 +153,7 @@ def to_csv(self) -> str:
             writer.writerow(row)
         return buff.getvalue()
 
-    def to_parallel_coordinates(self, output_path, color_by):
+    def to_parallel_coordinates(self, output_path, color_by=None):
```
skshetry, line 156: @daavoo, please add type hints when you add any functions, no need to do it in this PR though, can be separate.

skshetry (approved):

dvc/compare.py
```diff
@@ -153,7 +154,9 @@ def to_csv(self) -> str:
             writer.writerow(row)
         return buff.getvalue()
 
-    def to_parallel_coordinates(self, output_path, color_by):
+    def to_parallel_coordinates(
+        self, output_path: "StrPath", color_by: str = None
+    ):
```
skshetry, line 159:
```suggestion
    def to_parallel_coordinates(
        self, output_path: "StrPath", color_by: str = None
    ) -> str:
```

skshetry: Again, can be done in a separate PR @daavoo.

PR #7198, opened 2021-12-24
Fix memory leak in Index.build_graph()

Suor:
`@memoize()` memory lives as long as decorated function lives, which is
permanent in this case. Since `self` is referenced as a memory key no
`Index` will ever be freed.

Changed to use `@cached_property` who's cache lives as long as the
parent instance.

**NOTE:** This is an issue for Studio since we are parsing many commits in a long running process.

### DIFF
--- a/dvc/repo/index.py
+++ b/dvc/repo/index.py
@@ -11,7 +11,7 @@ from typing import (
     Set,
 )
 
-from funcy import cached_property, memoize, nullcontext
+from funcy import cached_property, nullcontext
 
 from dvc.utils import dict_md5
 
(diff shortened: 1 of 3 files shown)


pared (approved):

Neilblaze (approved):

skshetry (approved):

skshetry:
btw @Suor, how did you figure this out? This must have been tricky.

Suor:
I used `objgraph` mostly, can't find the original article but [its docs](https://mg.pov.lt/objgraph/) are actually better. Also, used [memory profiling](https://medium.com/zendesk-engineering/hunting-for-memory-leaks-in-python-applications-6824d0518774) a bit.

PR #7202, opened 2021-12-27
config: Add ui.theme and use in exp show. 

### LINKED ISSUE #7194: `exp show`: Background color for light terminals
This looks really bad on light and solarized light terminals.
Light Terminal

Solarized light

The headers' color was already quite bad for light terminals.

_Originally posted by @skshetry in https://github.com/iterative/dvc/issues/7128#issuecomment-1000036423_

daavoo:

Closes #7194 

- Default

Captura de pantalla 2021-12-28 a las 10 42 47

Captura de pantalla 2021-12-28 a las 10 46 23

- `dvc config ui.theme light`:

Captura de pantalla 2021-12-28 a las 10 43 35

- `dvc config ui.theme dark`:

Captura de pantalla 2021-12-28 a las 10 43 17

### DIFF
--- a/dvc/command/experiments/show.py
+++ b/dvc/command/experiments/show.py
@@ -385,6 +385,7 @@ def show_experiments(
     csv=False,
     markdown=False,
     html=False,
+    theme=None,
     **kwargs,
 ):
     from funcy.seqs import flatten as flatten_list
@@ -440,9 +441,11 @@ def show_experiments(
             td.drop(col)
 
     row_styles = lmap(baseline_styler, td.column("typ"))
-    for n, row_style in enumerate(row_styles):
-        if n % 2 != 0:
-            row_style["style"] += " on grey23"
+    if theme:
+        for n, row_style in enumerate(row_styles):
+            if n % 2 != 0:
+                color = "grey85" if theme == "light" else "grey23"
+                row_style["style"] += f" on {color}"
 
     if not csv:
         merge_headers = ["Experiment", "rev", "typ", "parent"]
@@ -489,7 +492,7 @@ def show_experiments(
 
     td.render(
         pager=pager,
-        borders="horizontals",
+        borders="horizontals" if theme else True,
         rich_table=True,
         header_styles=styles,
         row_styles=row_styles,
@@ -570,6 +573,7 @@ class CmdExperimentsShow(CmdBase):
                 html=self.args.html,
                 out=self.args.out,
                 open=self.args.open,
+                theme=self.repo.config.get("ui", {}).get("theme", None),
             )
         return 0
 
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -275,4 +275,7 @@ SCHEMA = {
         "plots": str,
         "live": str,
     },
+    "ui": {
+        Optional("theme", default=None): Any("light", "dark", None),
+    },
 }
--- a/tests/func/experiments/test_show.py
+++ b/tests/func/experiments/test_show.py
@@ -671,3 +671,34 @@ def test_show_parallel_coordinates(tmp_dir, dvc, scm, mocker):
     assert main(["exp", "show", "--html"]) == 0
     html_text = (tmp_dir / "dvc_plots" / "index.html").read_text()
     assert '{"label": "foobar", "values": [2.0, null, null]}' in html_text
+
+
+def test_show_theme(tmp_dir, scm, dvc, exp_stage, mocker):
+    from dvc.command.experiments import show
+    from dvc.ui import table
+    from dvc.utils.table import Table
+
+    show_experiments = mocker.spy(show, "show_experiments")
+    rich_table = mocker.spy(table, "rich_table")
+    add_row = mocker.spy(Table, "add_row")
+
+    assert main(["exp", "show"]) == 0
+    assert not show_experiments.call_args[1]["theme"]
+    assert add_row.call_args[1]["style"] == "bold"
+    assert rich_table.call_args[1]["borders"]
+
+    with dvc.config.edit() as conf:
+        conf["ui"]["theme"] = "light"
+
+    assert main(["exp", "show"]) == 0
+    assert show_experiments.call_args[1]["theme"]
+    assert add_row.call_args[1]["style"] == "bold on grey85"
+    assert rich_table.call_args[1]["borders"] == "horizontals"
+
+    with dvc.config.edit() as conf:
+        conf["ui"]["theme"] = "dark"
+
+    assert main(["exp", "show"]) == 0
+    assert show_experiments.call_args[1]["theme"]
+    assert add_row.call_args[1]["style"] == "bold on grey23"
+    assert rich_table.call_args[1]["borders"] == "horizontals"


daavoo (commented):
Probably more changes could be done using the new `ui.light`, i.e. the background headers

dberenbaum:
How does this look on dark backgrounds? Wondering if we can make it the default.

daavoo:
> How does this look on dark backgrounds? Wondering if we can make it the default.

Not sure I follow. On dark backgrounds, enabling this option looks bad because you are not supposed to activate the option. 

If you have a dark theme in the terminal and you run `dvc config ui.light True` :

Captura de pantalla 2021-12-27 a las 20 30 54

dberenbaum:
This is helpful, but it still leaves some sizeable portion of users with bad defaults. Are the alternating rows worth it if this is the case? Maybe they could be a separate option that's off by default?

Cc @skshetry @shcheklein

pmrowla:
I don't think we should introduce a config option for light/dark backgrounds. Terminal colors are supposed to work regardless of whether you have a light or dark terminal, and terminal apps work by mapping the base 16 colors properly whether you have it set to light or dark (so default text is always `white on black` even if your terminal app displays it as dark text on a light background). This is also how solarized works (it maps the standard 16 colors to the set of either light or dark solarized colors).

---

In general, we should try to avoid using explicit colors (like `greyXX`) as background colors unless we are also explicitly specifying the foreground color as well (like we do w/the column headers to match the viewer).

For the alternating background color case, if alternating between `on black` and `on bright_black` should work in pretty much any default terminal, regardless of the user's theme. Also note that some user's themes may use the `black` equivalent as the default background color, and some may use `bright_black` (and we have no way to determine this), so we will need to explicitly alternate between the two (and set it for every row, instead of only setting it for every other row).

dvc/command/experiments/show.py
```diff
@@ -442,7 +443,8 @@ def show_experiments(
     row_styles = lmap(baseline_styler, td.column("typ"))
     for n, row_style in enumerate(row_styles):
         if n % 2 != 0:
-            row_style["style"] += " on grey23"
+            color = "grey85" if light else "grey23"
+            row_style["style"] += f" on {color}"
```
pmrowla, line 447: This can just be

```py
        color = "black" if n % 2 == 0 else "bright_black"
        row_style["style"] += f" on {color}"
```

see https://github.com/iterative/dvc/pull/7202#issuecomment-1001848870

pmrowla (changes requested):
I don't think we should introduce a config option for light/dark backgrounds. Terminal colors are supposed to work regardless of whether you have a light or dark terminal, and terminal apps work by mapping the base 16 colors properly whether you have it set to light or dark ~~(so default text is always `white on black` even if your terminal app displays it as dark text on a light background). This is also how solarized works - it maps the standard 16 colors to the set of either light or dark solarized colors.~~

So _requiring_ a config option like this would basically mean DVC is no longer a well-behaved terminal app.

---

After some testing it looks like what I suggested above still doesn't work in all terminals as a universal solution (it depends on the theme being configured in the same way as solarized). It looks like the only "safe" option for the table would be to not alternate rows at all (and leave the default foreground/background color selection up to the user).

---

One other option would be to keep the config option and only alternate colors if the user has explicitly set the option to light or dark (which would be getting into DVC having specific color themes, rather than using the user's terminal theme).

So when the option is set to "light" we would alternate between `black on white` and `black on bright_white`, and for "dark" we would alternate between `white on black` and `white on bright_black`. In the default case (where the option is unset), we would not set per-row colors at all (and would not alternate rows).

@dberenbaum

daavoo:
> One other option would be to keep the config option and only alternate colors if the user has explicitly set the option to light or dark (which would be getting into DVC having specific color themes, rather than using the user's terminal theme).
> 
> So when the option is set to "light" we would alternate between ` on white` and ` on bright_white`, and for "dark" we would alternate between ` on black` and ` on bright_black`. In the default case (where the option is unset), we would not set per-row colors at all (and would not alternate rows).

Implemented this ^

dvc/command/experiments/show.py
```diff
@@ -440,9 +441,11 @@ def show_experiments(
...
 
     row_styles = lmap(baseline_styler, td.column("typ"))
-    for n, row_style in enumerate(row_styles):
-        if n % 2 != 0:
-            row_style["style"] += " on grey23"
+    if theme:
+        for n, row_style in enumerate(row_styles):
+            if n % 2 != 0:
+                color = "grey85" if theme == "light" else "grey23"
+                row_style["style"] += f" on {color}"
```
pmrowla, line 448:
```suggestion
            if n % 2 != 0:
                color = "white" if theme == "light" else "black"
            else:
                color = "bright_white" if theme == "light" else "bright_black"
            row_style["style"] += f" on {color}"
```

We should use the standard ansi colors for white/brwhite (light grey) or black/brblack (dark grey) rather than explicit 256 color values so that users with terminal themes that override the standard colors (like solarized) get their preferred background colors.

Also we need to explicitly specify both sets of alternating backgrounds in this case, since we don't know whether the user's default background color is the bright or regular color. If we only specify every other row and picked the "wrong" bright/regular color, it would not actually alternate when the user views the table.

pmrowla (changes requested):
Assuming we actually want the theme change, this needs one more update to use the ansi colors instead of explicit 256 color values.

dvc/command/experiments/show.py
daavoo: I understand the reasoning. However, using 16 colors (i.e. `bright_white`) doesn't look really good, IMO, in the themes I tested. 

For example, solarized `white/bright_white`:

Captura de pantalla 2021-12-28 a las 11 21 39

v.s. `grey23` on every other:

Captura de pantalla 2021-12-28 a las 11 22 01

daavoo: Not sure if it's worth the complexity of adding a separate option.

Perhaps the overall alternating background is just too complicated to get it right for every scenario and we should just leave the default.

pmrowla: The difference between the two bg colors in solarized isn't supposed to be too contrasty and is expected.

But FWIW In my terminal app's (iterm2 on macos) solarized light/dark and default black/white terminals, using `black/bright_black` and `white/bright_white` I still get a more distinctive alternating pattern than your example:

Screen Shot 2021-12-28 at 7 40 47 PM
Screen Shot 2021-12-28 at 7 41 09 PM
Screen Shot 2021-12-28 at 7 41 36 PM
Screen Shot 2021-12-28 at 7 42 29 PM

but yeah, the point is that doing this as a generic solution (as opposed to having proper "theme" support where users are explicitly setting all of the color values for DVC to use everywhere) is going to be difficult, and I'd be in favor of just dropping the alternating row thing entirely.

dberenbaum: Agreed. I'm fine with either keeping optional "themes" using ansi color codes or not implementing row colors at all.

efiop: It looks like this theme complexity is not worth it at all, and we shouldn't introduce it at all right now. Even given that we will want this in the future, it should probably behave like normal CLI utils do and create something like DVC_COLORS or something like that (e.g. see GREP_COLORS/LS_COLORS/etc). The current implementation complicates local exp show code and will be really hard to maintain it.

daavoo:
Closing this for now

Issue #7152, opened 2021-12-15
add `--rev` and updated `-n=-1` behavior to exp show


karajan1001:
_Originally posted by @karajan1001 in https://github.com/iterative/dvc/issues/5489#issuecomment-992234534_

Issue #7005, opened 2021-11-19
tests: move away from use of hardcoded `master` 


skshetry:
When I do `git init`, it prints the following warning:
```console
$ git init
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change.
```

So, I don't think we should depend on the hardcoded default branch in our tests **in the future**. To do so, we should introduce `scm.default_branch` property and use that in the tests instead.

_Originally posted by @skshetry in https://github.com/iterative/dvc/pull/6857#discussion_r742097567_

skshetry:
Closing as not planned.

PR #7204, opened 2021-12-28
exp show: add `--rev` flag (#7152)

karajan1001:
fix: #7152
1. Add `--rev` flag to `dvc exp show`.
2. Add -1 support for `--num` flag`.
3. Extract revision solving to `utils`.
4. Update command unit test
5. Add a util unit test.

### DIFF
--- a/dvc/command/experiments/show.py
+++ b/dvc/command/experiments/show.py
@@ -532,8 +532,9 @@ class CmdExperimentsShow(CmdBase):
                 all_branches=self.args.all_branches,
                 all_tags=self.args.all_tags,
                 all_commits=self.args.all_commits,
-                sha_only=self.args.sha,
+                revs=self.args.rev,
                 num=self.args.num,
+                sha_only=self.args.sha,
                 param_deps=self.args.param_deps,
             )
         except DvcException:
@@ -605,13 +606,23 @@ def add_parser(experiments_subparsers, parent_parser):
         help="Show experiments derived from all Git commits.",
     )
     experiments_show_parser.add_argument(
+        "--rev",
+        type=str,
+        default=None,
+        help=(
+            "Show experiments derived from the specified revision. "
+            "Defaults to HEAD if none of `--rev`,`-a`,`-A`,`-T` is specified."
+        ),
+        metavar="<rev>",
+    )
+    experiments_show_parser.add_argument(
         "-n",
         "--num",
         type=int,
         default=1,
         dest="num",
         metavar="<num>",
-        help="Show the last `num` commits from HEAD.",
+        help="Show the last `num` commits from <rev>.",
     )
     experiments_show_parser.add_argument(
         "--no-pager",
--- a/dvc/repo/brancher.py
+++ b/dvc/repo/brancher.py
@@ -1,6 +1,4 @@
-from functools import partial
-
-from funcy import group_by
+from dvc.scm import iter_revs
 
 
 def brancher(  # noqa: E302
@@ -34,44 +32,33 @@ def brancher(  # noqa: E302
     from dvc.fs.local import LocalFileSystem
 
     saved_fs = self.fs
-    revs = revs.copy() if revs else []
 
     scm = self.scm
 
     self.fs = LocalFileSystem(url=self.root_dir)
     yield "workspace"
 
-    if revs and "workspace" in revs:
+    revs = revs.copy() if revs else []
+    if "workspace" in revs:
         revs.remove("workspace")
 
-    if all_commits:
-        revs = scm.list_all_commits()
-    else:
-        if all_branches:
-            revs.extend(scm.list_branches())
-
-        if all_tags:
-            revs.extend(scm.list_tags())
-
-    if all_experiments:
-        from dvc.repo.experiments.utils import exp_commits
-
-        revs.extend(exp_commits(scm))
+    found_revs = iter_revs(
+        scm,
+        revs,
+        all_branches=all_branches,
+        all_tags=all_tags,
+        all_commits=all_commits,
+        all_experiments=all_experiments,
+    )
 
     try:
-        if revs:
-            from dvc.fs.git import GitFileSystem
-            from dvc.scm import resolve_rev
+        from dvc.fs.git import GitFileSystem
 
-            rev_resolver = partial(resolve_rev, scm)
-            for sha, names in group_by(rev_resolver, revs).items():
-                self.fs = GitFileSystem(scm=scm, rev=sha)
-                # ignore revs that don't contain repo root
-                # (i.e. revs from before a subdir=True repo was init'ed)
-                if self.fs.exists(self.root_dir):
-                    if sha_only:
-                        yield sha
-                    else:
-                        yield ", ".join(names)
+        for sha, names in found_revs.items():
+            self.fs = GitFileSystem(scm=scm, rev=sha)
+            # ignore revs that don't contain repo root
+            # (i.e. revs from before a subdir=True repo was init'ed)
+            if self.fs.exists(self.root_dir):
+                yield sha if sha_only else ",".join(names)
     finally:
         self.fs = saved_fs
--- a/dvc/scm.py
+++ b/dvc/scm.py
@@ -1,12 +1,14 @@
 """Manages source control systems (e.g. Git)."""
 from contextlib import contextmanager
-from typing import TYPE_CHECKING, Iterator
+from functools import partial
+from typing import TYPE_CHECKING, Iterator, List, Mapping, Optional
 
+from funcy import group_by
 from scmrepo.base import Base  # noqa: F401, pylint: disable=unused-import
 from scmrepo.git import Git
 from scmrepo.noscm import NoSCM
 
-from dvc.exceptions import DvcException
+from dvc.exceptions import DvcException, InvalidArgumentError
 from dvc.progress import Tqdm
 
 if TYPE_CHECKING:
@@ -123,3 +125,56 @@ def resolve_rev(scm: "Git", rev: str) -> str:
             if len(ref_infos) > 1:
                 raise RevError(f"ambiguous Git revision '{rev}'")
         raise RevError(str(exc))
+
+
+def iter_revs(
+    scm: "Git",
+    head_revs: Optional[List[str]] = None,
+    num: int = 1,
+    all_branches: bool = False,
+    all_tags: bool = False,
+    all_commits: bool = False,
+    all_experiments: bool = False,
+) -> Mapping[str, List[str]]:
+    from dvc.repo.experiments.utils import fix_exp_head
+
+    if num < 1 and num != -1:
+        raise InvalidArgumentError(f"Invalid number of commits '{num}'")
+
+    if not any(
+        [head_revs, all_branches, all_tags, all_commits, all_experiments]
+    ):
+        return {}
+
+    head_revs = head_revs or []
+    revs = []
+    for rev in head_revs:
+        revs.append(rev)
+        n = 1
+        while True:
+            if num == n:
+                break
+            try:
+                head = fix_exp_head(scm, f"{rev}~{n}")
+                assert head
+                revs.append(resolve_rev(scm, head))
+            except RevError:
+                break
+            n += 1
+
+    if all_commits:
+        revs.extend(scm.list_all_commits())
+    else:
+        if all_branches:
+            revs.extend(scm.list_branches())
+
+        if all_tags:
+            revs.extend(scm.list_tags())
+
+    if all_experiments:
+        from dvc.repo.experiments.utils import exp_commits
+
+        revs.extend(exp_commits(scm))
+
+    rev_resolver = partial(resolve_rev, scm)
+    return group_by(rev_resolver, revs)
--- a/tests/unit/scm/test_scm.py
+++ b/tests/unit/scm/test_scm.py
@@ -1,0 +1,63 @@
+from dvc.repo.experiments import ExpRefInfo
+from dvc.scm import iter_revs
+
+
+def test_iter_revs(
+    tmp_dir,
+    scm,
+):
+    """
+    new         other
+     │            │
+    old (tag) ────┘
+     │
+    root
+    """
+    old = scm.active_branch()
+    tmp_dir.scm_gen("foo", "init", commit="init")
+    rev_root = scm.get_rev()
+    tmp_dir.scm_gen("foo", "old", commit="old")
+    rev_old = scm.get_rev()
+    scm.checkout("new", create_new=True)
+    tmp_dir.scm_gen("foo", "new", commit="new")
+    rev_new = scm.get_rev()
+    scm.checkout(old)
+    scm.tag("tag")
+    scm.checkout("other", create_new=True)
+    tmp_dir.scm_gen("foo", "other", commit="new")
+    rev_other = scm.get_rev()
+
+    ref = ExpRefInfo(rev_root, "exp1")
+    scm.set_ref(str(ref), rev_new)
+    ref = ExpRefInfo(rev_root, "exp2")
+    scm.set_ref(str(ref), rev_old)
+
+    gen = iter_revs(scm, [rev_root, "new"], 1)
+    assert gen == {rev_root: [rev_root], rev_new: ["new"]}
+    gen = iter_revs(scm, ["new"], 2)
+    assert gen == {rev_new: ["new"], rev_old: [rev_old]}
+    gen = iter_revs(scm, ["other"], -1)
+    assert gen == {
+        rev_other: ["other"],
+        rev_old: [rev_old],
+        rev_root: [rev_root],
+    }
+    gen = iter_revs(scm, ["tag"])
+    assert gen == {rev_old: ["tag"]}
+    gen = iter_revs(scm, all_branches=True)
+    assert gen == {rev_old: [old], rev_new: ["new"], rev_other: ["other"]}
+    gen = iter_revs(scm, all_tags=True)
+    assert gen == {rev_old: ["tag"]}
+    gen = iter_revs(scm, all_commits=True)
+    assert gen == {
+        rev_old: [rev_old],
+        rev_new: [rev_new],
+        rev_other: [rev_other],
+        rev_root: [rev_root],
+    }
+    gen = iter_revs(scm, all_experiments=True)
+    assert gen == {
+        rev_new: [rev_new],
+        rev_old: [rev_old],
+        rev_root: [rev_root],
+    }
--- a/dvc/repo/experiments/show.py
+++ b/dvc/repo/experiments/show.py
@@ -1,14 +1,13 @@
 import logging
 from collections import OrderedDict, defaultdict
 from datetime import datetime
-from typing import Any, Callable, Dict, Optional
+from typing import Any, Callable, Dict, List, Optional, Union
 
-from dvc.exceptions import InvalidArgumentError
-from dvc.repo import locked
+from dvc.repo import Repo, locked  # pylint: disable=unused-import
 from dvc.repo.experiments.base import ExpRefInfo
-from dvc.repo.experiments.utils import fix_exp_head
 from dvc.repo.metrics.show import _gather_metrics
 from dvc.repo.params.show import _gather_params
+from dvc.scm import iter_revs
 from dvc.utils import error_handler, onerror_collect
 
 logger = logging.getLogger(__name__)
@@ -102,10 +101,10 @@ def _collect_experiment_branch(
 
 @locked
 def show(
-    repo,
+    repo: "Repo",
     all_branches=False,
     all_tags=False,
-    revs=None,
+    revs: Union[List[str], str, None] = None,
     all_commits=False,
     sha_only=False,
     num=1,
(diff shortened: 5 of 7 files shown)


dvc/repo/experiments/__init__.py
```diff
@@ -786,6 +787,51 @@ def on_diverged(_ref: str, _checkpoint: bool):
...
+                    head = fix_exp_head(self.repo.scm, f"{rev}~{n}")
+                    assert head
+                    revs.append(resolve_rev(self.repo.scm, head))
+                except RevError:
+                    break
+                n += 1
+
+        return_revs = OrderedDict(
+            (rev, None)
+            for rev in self.repo.brancher(
```
pmrowla, line 824: Using `brancher` here is a pretty heavy dependency since it creates the git-fs and changes the current `repo` instance for each rev. It's also bit redundant since we also use `brancher` on single revisions in `_collect_experiment_commit` (where we actually do want the git-fs and modified `repo` instance).

Since at this point in the call we only want the list of revisions that the combination of `all_branches/all_tags/all_commits` would give, what we probably want is a to split the current `brancher` functionality for just getting the revisions into a helper method https://github.com/iterative/dvc/blob/ceeda6d3072fb7d862518a60e9e2f6a0ae2cf3d2/dvc/repo/brancher.py#L30-L59

(so that we can get the revs but skip the git-fs/repo modification when we don't need it)

skshetry: Something like `iter_revs`  perhaps?

```python
for rev in iter_revs(revs, all_branches=all_branches, all_tags=all_tags, all_experiments=all_experiments, all_commits=all_commits):
    pass
```

karajan1001: It is OK to rename `get_revs_from_flags` as `iter_revs` for me, but the problem preventing me merge the logic between `get_revs_from_flags` and `brancher` here is that there are some problems with the output format of `iter_revs`. 
    The output format of `brancher` without a `sha_only` flag can be `shas`, `branch names`, `tag names` or even `HEAD` or `HEAD~~~`. Besides this when one `commit` have multiple names among this, we need to the one that users provide to make them more understandable to the users. 
   While on the `get_revs_from_flags` side the output is quite simple, we just crunch all different inputs into `shas`.
   I think it is better to follow the output rule in `brancher` here for both of the two functions.

dvc/scm.py
```diff
@@ -123,3 +126,61 @@ def resolve_rev(scm: "Git", rev: str) -> str:
...
+    return_revs = revs.copy()
+    if all_commits:
+        return_revs.extend(scm.list_all_commits())
+    else:
+        if all_branches:
+            return_revs.extend(scm.list_branches())
+
+        if all_tags:
+            return_revs.extend(scm.list_tags())
+    return return_revs
```
pmrowla, line 186: This can just be `iter_revs` as described by @skshetry (or maybe something like `iter_rev_names`), and it can yield the same `sha, names` pairs that brancher currently uses. It should also support `all_experiments` the same way that brancher does. So basically this iterator should do everything that `brancher` does right now except for modifying `repo` and generating a git-fs.

So this method would look something like
```py
def iter_rev_names(
    scm: "Git",
    revs: List[str],
    all_branches: bool = False,
    ...
) Generator[...]:
    ... # get revs list the same way as brancher()
    if revs:
            rev_resolver = partial(resolve_rev, scm)
            # yield (rev, names) pairs
            yield from group_by(rev_resolver, revs).items()
```
This way code outside of brancher can use the `rev, names` items as needed without repo modification and git-fs overhead.

And now `brancher()` itself could look something like
```py
def brancher(
    self,
    *args,
    sha_only=False,
    **kwargs,
):
    saved_fs = self.fs
    try:
        for sha, names in iter_rev_names(self.scm, *args, **kwargs):
                self.fs = GitFileSystem(scm=scm, rev=sha)
                if self.fs.exists(self.root_dir):
                    if sha_only:
                        yield sha
                    else:
                        yield ", ".join(names)
```

dvc/scm.py
```diff
@@ -123,3 +126,61 @@ def resolve_rev(scm: "Git", rev: str) -> str:
...
+        while True:
+            if num == n:
+                break
+            try:
+                head = fix_exp_head(scm, f"{rev}~{n}")
+                assert head
+                revs.append(resolve_rev(scm, head))
+            except RevError:
+                break
+            n += 1
```
pmrowla, line 160: This `fix_exp_head` logic could be moved into the generalized `iter_rev_names` method (and `brancher()` by extension).

Everywhere we use `brancher` in DVC we also need the exp HEAD fixup. so we can do the `fix_exp_head` inside the refactored iterator and then remove some of the existing `fix_exp_head` calls in DVC (`exp show` + params/metrics/plots diff) so that it's always unified in one place

pmrowla (changes requested):

dvc/repo/experiments/__init__.py
pmrowla: The iterator can return both the `sha` and `names` pairs. It's really a `brancher()` design flaw that it's only designed to return either SHAs or names, but not both. see updated comment: https://github.com/iterative/dvc/pull/7204#discussion_r778556617

dvc/scm.py
```diff
@@ -123,3 +126,61 @@ def resolve_rev(scm: "Git", rev: str) -> str:
...
+
+
+def get_revs_from_args(
+    scm: "Git",
+    head_revs: Optional[List[str]] = None,
+    num: int = 1,
+    all_branches: bool = False,
+    all_tags: bool = False,
+    all_commits: bool = False,
+) -> OrderedDict:
```
skshetry, line 138: Please don't use `OrderedDict` as `dict` is guaranteed to be ordered based on insertions from 3.7+ onwards.

tests/unit/scm/test_scm.py
```diff
@@ -0,0 +1,54 @@
...
+    root
+    """
+    tmp_dir.scm_gen("foo", "init", commit="init")
+    rev_root = scm.get_rev()
+    tmp_dir.scm_gen("foo", "old", commit="old")
+    rev_old = scm.get_rev()
+    scm.checkout("new", create_new=True)
+    tmp_dir.scm_gen("foo", "new", commit="new")
+    rev_new = scm.get_rev()
+    scm.checkout("master")
```
karajan1001, line 23: @skshetry related to #7005, we don't have `scm.default_branch` yet?

dvc/scm.py
```diff
@@ -123,3 +125,60 @@ def resolve_rev(scm: "Git", rev: str) -> str:
...
+
+    if all_experiments:
+        from dvc.repo.experiments.utils import exp_commits
+
+        revs.extend(exp_commits(scm))
+
+    rev_resolver = partial(resolve_rev, scm)
+    result = {
+        sha: ",".join(name)
+        for sha, name in group_by(rev_resolver, revs).items()
```
karajan1001, line 182: After some trying it looks like the order will be preserved after group_by.

tests/unit/scm/test_scm.py
```diff
@@ -0,0 +1,54 @@
...
+    assert gen == {rev_old: "tag"}
+    gen = iter_revs(scm, all_commits=True)
+    assert gen == {
+        rev_old: rev_old,
+        rev_new: rev_new,
+        rev_other: rev_other,
+        rev_root: rev_root,
+    }
+    gen = iter_revs(scm, all_experiments=True)
+    assert gen == {rev_new: rev_new, rev_old: rev_old, rev_root: rev_root}
```
karajan1001, line 54: https://github.com/iterative/dvc/blob/79ebbc24e563c7ed3c918f9c3545b355e795794d/dvc/repo/experiments/utils.py#L127-L128

Not understand, why the baseline also been included in `all_experiments`?

pmrowla: It's because `scm.branch_revs(branch, parent)` is not inclusive. It walks backwards and includes any revs up until it reaches `parent`

dvc/scm.py
```diff
@@ -123,3 +125,60 @@ def resolve_rev(scm: "Git", rev: str) -> str:
...
+    }
+    return result
```
pmrowla, line 184: I think we should leave formatting for the names here up to the caller, and having access to the actual names themselves would make this a more useful API call in general. The `",".join(names)` is really brancher specific behavior and should happen in `brancher()` not `iter_revs`

This method could actually just return the `group_by` dict itself, and any caller that actually cares about the names can get `items()` as needed. (So the return type would be `Mapping[str, Iterable[str]]` or `Dict` instead of `Mapping`)

```suggestion
    return group_by(rev_resolver, revs)
```

This way in callers where we only care about the rev SHAs (like exp show), we can just use the SHAs like `for rev in iter_revs():`.

dvc/repo/brancher.py
```diff
@@ -34,44 +32,34 @@ def brancher(  # noqa: E302
...
+        all_commits=all_commits,
+        all_experiments=all_experiments,
+    )
 
     try:
-        if revs:
-            from dvc.fs.git import GitFileSystem
-            from dvc.scm import resolve_rev
+        if found_revs:
+            for sha, names in found_revs.items():
```
pmrowla, line 56: This `if` isn't needed

```suggestion
        for sha, names in found_revs.items():
```

dvc/repo/brancher.py
```diff
@@ -34,44 +32,34 @@ def brancher(  # noqa: E302
...
-            for sha, names in group_by(rev_resolver, revs).items():
                 self.fs = GitFileSystem(scm=scm, rev=sha)
                 # ignore revs that don't contain repo root
                 # (i.e. revs from before a subdir=True repo was init'ed)
                 if self.fs.exists(self.root_dir):
-                    if sha_only:
-                        yield sha
-                    else:
-                        yield ", ".join(names)
+                    yield sha if sha_only else names
```
pmrowla, line 63: With the other suggested change this would become
```suggestion
                    yield sha if sha_only else ",".join(names)
```

pmrowla (changes requested):

tests/unit/scm/test_scm.py
skshetry: Git does not have a concept of default branch locally. `init` uses `init.defaultBranch`/`--initial-branch` and defaults to `master`. `clone` uses currently active branch as initial branch by default.

 At least in scmrepo, it is not possible to know what branch is initial. So this feels like a tests-only requirement. Due to these issues, I have moved away from `default_branch` suggestion. I think the best way would be to keep a reference to `scm.active_branch()`, and use that instead.

skshetry: Here you may be able to avoid using hardcoded master by using `tmp_dir.branch` context manager.

https://github.com/iterative/dvc/blob/92f714e5ca7f4a9e7701a06f7d21b478049f4af7/dvc/testing/tmp_dir.py#L222

pmrowla (approved):
This will need a docs command reference update

dvc/scm.py
```diff
@@ -123,3 +125,56 @@ def resolve_rev(scm: "Git", rev: str) -> str:
...
+        raise InvalidArgumentError(f"Invalid number of commits '{num}'")
+
+    if not any(
+        [head_revs, all_branches, all_tags, all_commits, all_experiments]
+    ):
+        return {}
+
+    head_revs = head_revs or []
+    revs = []
+    for rev in head_revs:
```
skshetry, line 151: Wasn't this mutually exclusive with `num`/`n` before?

dvc/scm.py
```diff
@@ -123,3 +125,56 @@ def resolve_rev(scm: "Git", rev: str) -> str:
             if len(ref_infos) > 1:
                 raise RevError(f"ambiguous Git revision '{rev}'")
         raise RevError(str(exc))
+
+
+def iter_revs(
+    scm: "Git",
+    head_revs: Optional[List[str]] = None,
```
skshetry, line 132: Why not use `revs`?

dvc/scm.py
```diff
@@ -123,3 +125,56 @@ def resolve_rev(scm: "Git", rev: str) -> str:
             if len(ref_infos) > 1:
                 raise RevError(f"ambiguous Git revision '{rev}'")
         raise RevError(str(exc))
+
+
+def iter_revs(
```
skshetry, line 130: `iter_revs` is no longer an iterator now.

dvc/scm.py
skshetry: Can't this be done separately with `n_commits(num)` if it's exclusive, outside of `iter_revs`?

dvc/scm.py
karajan1001: We can call it `revs` + `results` compare to current `head_revs` + `revs`

dvc/scm.py
karajan1001: Any name to suggest?

dvc/scm.py
karajan1001: In the previous version `rev+num` is not mutually exclusive with `all_tags/all_branches`. And the result will be the sum of all revs. Maybe it is better to make them exclusive? @dberenbaum 
This function will be used in `exp push/pull/ls/show/remove` If I separate `n_commits(num)` here then I need to add a new function and call them five times there.

dvc/scm.py
karajan1001: We can modify these in #7245

karajan1001:
Add a new demo for this PR.

[asciicast](https://asciinema.org/a/7SGofYb1ayT4uKS3hFa6A4HVJ)

dvc/scm.py
dberenbaum: > And the result will be the sum of all revs.

How is this supposed to work? If I do `dvc exp show -a -n 10`, I get:

```
 ───────────────────────────────────────────────────────────────────────────────────────
  Experiment                     Created        avg_prec   roc_auc   prepare.split   pre
 ───────────────────────────────────────────────────────────────────────────────────────
  workspace                      -               0.60405    0.9608   0.2             201
  11-random-forest-experiments   May 29, 2021    0.60405    0.9608   0.2             201
 ───────────────────────────────────────────────────────────────────────────────────────
```

So it seems like `-n` is being ignored?

I don't feel strongly about whether we allow these to be combined. It seems strange to combine them, but if we can explain the behavior clearly, it might be fine to allow users to combine them if they want.

dberenbaum:
`dvc exp show -v -n -1` throws an error for me:

```
2022-01-25 12:25:40,695 DEBUG: Adding '/private/tmp/example-get-started/.dvc/config.local' to gitignore file.
2022-01-25 12:25:40,697 DEBUG: Adding '/private/tmp/example-get-started/.dvc/tmp' to gitignore file.
2022-01-25 12:25:40,697 DEBUG: Adding '/private/tmp/example-get-started/.dvc/cache' to gitignore file.
2022-01-25 12:25:40,711 ERROR: unexpected error - refs/remotes/origin/HEAD~12: the given reference name 'refs/remotes/origin/HEAD~12' is not valid
  File "/Users/dave/Code/dvc/dvc/cli/__init__.py", line 78, in main
    ret = cmd.do_run()
  File "/Users/dave/Code/dvc/dvc/cli/command.py", line 22, in do_run
    return self.run()
  File "/Users/dave/Code/dvc/dvc/commands/experiments/show.py", line 445, in run
    all_experiments = self.repo.experiments.show(
  File "/Users/dave/Code/dvc/dvc/repo/experiments/__init__.py", line 820, in show
  File "/Users/dave/Code/dvc/dvc/repo/__init__.py", line 48, in wrapper
  File "/Users/dave/Code/dvc/dvc/repo/experiments/show.py", line 126, in show
    iter_revs(repo.scm, revs, num, all_branches, all_tags, all_commits)
  File "/Users/dave/Code/dvc/dvc/scm.py", line 165, in iter_revs
    revs.append(resolve_rev(scm, head))
  File "/Users/dave/Code/dvc/dvc/scm.py", line 115, in resolve_rev
    return scm.resolve_rev(rev)
  File "/Users/dave/miniforge3/envs/example-get-started/lib/python3.8/site-packages/scmrepo/git/__init__.py", line 253, in _backend_func
    return func(*args, **kwargs)
  File "/Users/dave/miniforge3/envs/example-get-started/lib/python3.8/site-packages/scmrepo/git/backend/pygit2.py", line 263, in resolve_rev
    shas = {
  File "/Users/dave/miniforge3/envs/example-get-started/lib/python3.8/site-packages/scmrepo/git/backend/pygit2.py", line 264, in <setcomp>
    self.get_ref(f"refs/remotes/{remote.name}/{rev}")
  File "/Users/dave/miniforge3/envs/example-get-started/lib/python3.8/site-packages/scmrepo/git/backend/pygit2.py", line 322, in get_ref
    ref = self.repo.references.get(name)
  File "/Users/dave/miniforge3/envs/example-get-started/lib/python3.8/site-packages/pygit2/repository.py", line 1440, in get
    return self[key]
  File "/Users/dave/miniforge3/envs/example-get-started/lib/python3.8/site-packages/pygit2/repository.py", line 1436, in __getitem__
    return self._repository.lookup_reference(name)
_pygit2.InvalidSpecError: refs/remotes/origin/HEAD~12: the given reference name 'refs/remotes/origin/HEAD~12' is not valid
2022-01-25 12:25:41,566 DEBUG: Adding '/private/tmp/example-get-started/.dvc/config.local' to gitignore file.
2022-01-25 12:25:41,567 DEBUG: Adding '/private/tmp/example-get-started/.dvc/tmp' to gitignore file.
2022-01-25 12:25:41,567 DEBUG: Adding '/private/tmp/example-get-started/.dvc/cache' to gitignore file.
2022-01-25 12:25:41,568 DEBUG: Version info for developers:
DVC version: 2.9.4.dev58+g47481219
Platform: Python 3.8.5 on macOS-10.16-x86_64-i386-64bit
Supports:
        gdrive (pydrive2 = 1.7.3),
        hdfs (fsspec = 2022.1.0, pyarrow = 3.0.0),
        webhdfs (fsspec = 2022.1.0),
        http (aiohttp = 3.7.3, aiohttp-retry = 2.4.6),
        https (aiohttp = 3.7.3, aiohttp-retry = 2.4.6),
        ssh (sshfs = 2021.11.2)
Caches: local
Remotes: https
Repo: dvc, git

2022-01-25 12:25:41,569 DEBUG: Analytics is disabled.
```

karajan1001:
Looks like it is a bug from `scmrepo`. In `gitpython` backend  it raise a proper `RevError`. 

```
In [1]: from scmrepo.git import Git

In [2]: repo = Git('.')

In [3]: repo.gitpython.resolve_rev("HEAD~5")
Out[3]: '933cfcc55c98d445fddd952815881c8a5df3e797'

In [4]: repo.gitpython.resolve_rev("HEAD~6")
---------------------------------------------------------------------------
RevError                                  Traceback (most recent call last)
<ipython-input-4-2617366dae9f> in <module>
----> 1 repo.gitpython.resolve_rev("HEAD~6")

~/anaconda3/envs/dvc/lib/python3.8/site-packages/scmrepo/git/backend/gitpython.py in resolve_rev(self, rev)
    350                 return shas.pop()
    351
--> 352         raise RevError(f"unknown Git revision '{rev}'")
    353
    354     def resolve_commit(self, rev: str) -> "GitCommit":

RevError: unknown Git revision 'HEAD~6'
```

while in `pygit2` it didn't raise a proper exception.
```
In [5]: repo.pygit2.resolve_rev("HEAD~6")
---------------------------------------------------------------------------
InvalidSpecError                          Traceback (most recent call last)
<ipython-input-5-d503510da33d> in <module>
----> 1 repo.pygit2.resolve_rev("HEAD~6")

~/anaconda3/envs/dvc/lib/python3.8/site-packages/scmrepo/git/backend/pygit2.py in resolve_rev(self, rev)
    261
    262         # Look for single exact match in remote refs
--> 263         shas = {
    264             self.get_ref(f"refs/remotes/{remote.name}/{rev}")
    265             for remote in self.repo.remotes

~/anaconda3/envs/dvc/lib/python3.8/site-packages/scmrepo/git/backend/pygit2.py in <setcomp>(.0)
    262         # Look for single exact match in remote refs
    263         shas = {
--> 264             self.get_ref(f"refs/remotes/{remote.name}/{rev}")
    265             for remote in self.repo.remotes
    266         } - {None}

~/anaconda3/envs/dvc/lib/python3.8/site-packages/scmrepo/git/backend/pygit2.py in get_ref(self, name, follow)
    320         from pygit2 import GIT_REF_SYMBOLIC
    321
--> 322         ref = self.repo.references.get(name)
    323         if not ref:
    324             return None

~/anaconda3/envs/dvc/lib/python3.8/site-packages/pygit2/repository.py in get(self, key)
   1438     def get(self, key):
   1439         try:
-> 1440             return self[key]
   1441         except KeyError:
   1442             return None

~/anaconda3/envs/dvc/lib/python3.8/site-packages/pygit2/repository.py in __getitem__(self, name)
   1434
   1435     def __getitem__(self, name):
-> 1436         return self._repository.lookup_reference(name)
   1437
   1438     def get(self, key):

InvalidSpecError: refs/remotes/origin/HEAD~6: the given reference name 'refs/remotes/origin/HEAD~6' is not valid

In [6]: repo.pygit2.resolve_rev("HEAD~5")
Out[6]: '933cfcc55c98d445fddd952815881c8a5df3e797'
```

dberenbaum:
Thanks for looking into it, @karajan1001! Do you know what needs to be fixes so that it actually returns the table with the full history? Should we open a new issue or PR to fix?

karajan1001:
I think we should first fix it on the scmrepo side and then upgrade the version here.

dvc/command/experiments/show.py
```diff
@@ -604,14 +605,24 @@ def add_parser(experiments_subparsers, parent_parser):
...
+    )
     experiments_show_parser.add_argument(
         "-n",
         "--num",
         type=int,
         default=1,
         dest="num",
         metavar="<num>",
-        help="Show the last `num` commits from HEAD.",
+        help="Show the last `num` commits from <rev>.",
```
jorgeorpinel, line 625: OK so I know this is already part of the codebase, sorry. But should the metavar be <int>? "num" is not informative since the option has the same name. "int" is a type hint at least, which some time ago we agreed to use for metavars AFAIR. Either way it should be pretty obvious here but suggesting as a UX improvement. Cc @dberenbaum

jorgeorpinel: Moved to https://github.com/iterative/dvc/issues/7314 actually.

karajan1001:
> @karajan1001 In the meantime, should we roll back the `-1` support and make a separate issue to add this? This would unblock other PRs and docs as well.

This issue not only exists in `-1` conditions but on any n numbers larger than history commits. But users do not give such a big `n` normally. And if it happened, users can give a smaller `n` to get the correct output. While for `-1`, it would not work unless removing the `origin` in the remote. So I think we can roll back for now.

PR #7205, opened 2021-12-28
exp remove: refactor exp utils

### LINKED ISSUE #7196: `exp pull/push`: add multi exp name arg support
_Originally posted by @karajan1001 in https://github.com/iterative/dvc/issues/5489#issuecomment-992234534_

karajan1001:
fix #7196 

Refactor exp_ref_by_name in `dvc/repo/experiments/utils.py`.

In previous the exp_ref_by_name walk throught all of the references to
match only one exp name. This makes multi exp name seeking to be a job
of `O(mn)` complexity, Walking though the exp ref and return all of the
exp ref with greatly reduce this complexity to `O(n)`

### DIFF
--- a/dvc/repo/experiments/utils.py
+++ b/dvc/repo/experiments/utils.py
@@ -1,4 +1,14 @@
-from typing import Callable, Generator, Iterable, Optional, Set
+from collections import defaultdict
+from typing import (
+    Callable,
+    Dict,
+    Generator,
+    Iterable,
+    List,
+    Optional,
+    Set,
+    Union,
+)
 
 from scmrepo.git import Git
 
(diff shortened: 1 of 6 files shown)


dvc/repo/experiments/utils.py
```diff
@@ -13,6 +23,35 @@
...
+            ]
+        else:
+            msg = [
+                (
+                    f"Ambiguous name '{exp_name}' refers to multiple "
+                    f"experiments in '{git_remote}'. Use full refname to pull "
+                    "one of the following:"
+                ),
+                "",
+            ]
```
pmrowla, line 50: The error messages now are too specific, we don't know if we are pushing or pulling (and we could also be using `resolve_name` in something completely unrelated to push/pull). We should probably just use a more general message without `git_remote` for every case like:

`Ambiguous name '{exp_name}' refers to multiple experiments. Use one of the following full refnames instead:`

pmrowla (changes requested):

pmrowla (approved):

PR #7213, opened 2021-12-30
fs.gdrive: Fix GDRIVE_CREDENTIALS_DATA for service accounts

RCdeWit:
`_validate_config()` had a bug: it would not take into account that Google
Drive credentials can also be passed using `GDRIVE_CREDENTIALS_DATA` as
documented here [1].

This patches the validation logic to also accept this environment variable as a
means to provide service account credentials.

Before, commands like `dvc push` would exit with:

```
ERROR: failed to push data to the cloud - To use service account, set
  `gdrive_service_account_json_file_path`, and
  optionally`gdrive_service_account_user_email` in DVC config
Learn more at <https://man.dvc.org/remote/modify>
```

[1]: https://dvc.org/doc/user-guide/setup-google-drive-remote#authorization

Thanks to @duijf for helping me diagnose.

Co-authored-by: Laurens Duijvesteijn <laurens@iterative.ai>

### DIFF
--- a/dvc/fs/gdrive.py
+++ b/dvc/fs/gdrive.py
@@ -109,11 +109,14 @@ class GDriveFileSystem(FSSpecWrapper):  # pylint:disable=abstract-method
         if (
             self._use_service_account
             and not self._service_account_json_file_path
+            and "GDRIVE_CREDENTIALS_DATA" not in os.environ
         ):
             raise DvcException(
                 "To use service account, set "
                 "`gdrive_service_account_json_file_path`, and optionally"
                 "`gdrive_service_account_user_email` in DVC config\n"
+                "You can also provide these using the GDRIVE_CREDENTIALS_DATA"
+                "environment variable.\n"
                 "Learn more at {}".format(
                     format_link("https://man.dvc.org/remote/modify")
                 )
--- a/dvc/fs/gdrive.py
+++ b/dvc/fs/gdrive.py
@@ -109,14 +109,12 @@ class GDriveFileSystem(FSSpecWrapper):  # pylint:disable=abstract-method
         if (
             self._use_service_account
             and not self._service_account_json_file_path
-            and "GDRIVE_CREDENTIALS_DATA" not in os.environ
+            and GDriveFileSystem.GDRIVE_CREDENTIALS_DATA not in os.environ
         ):
             raise DvcException(
                 "To use service account, set "
                 "`gdrive_service_account_json_file_path`, and optionally"
                 "`gdrive_service_account_user_email` in DVC config\n"
-                "You can also provide these using the GDRIVE_CREDENTIALS_DATA"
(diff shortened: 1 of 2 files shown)


pmrowla:
Related: https://github.com/iterative/dvc/issues/5580

dvc/fs/gdrive.py
```diff
@@ -109,11 +109,14 @@ def _validate_config(self):
         if (
             self._use_service_account
             and not self._service_account_json_file_path
+            and "GDRIVE_CREDENTIALS_DATA" not in os.environ
```
skshetry, line 112:
```suggestion
            and self.GDRIVE_CREDENTIALS_DATA not in os.environ
```

dvc/fs/gdrive.py
```diff
@@ -109,11 +109,14 @@ def _validate_config(self):
...
         ):
             raise DvcException(
                 "To use service account, set "
                 "`gdrive_service_account_json_file_path`, and optionally"
                 "`gdrive_service_account_user_email` in DVC config\n"
+                "You can also provide these using the GDRIVE_CREDENTIALS_DATA"
+                "environment variable.\n"
```
skshetry, line 119: I think we can just leave this message out.

dvc/fs/gdrive.py
karajan1001: In the previous codes, it is `GDriveFileSystem.GDRIVE_CREDENTIALS_DATA`, better to be consistent?

dberenbaum:
@RCdeWit Thanks for the contribution! Would you be able to respond to update the PR to address the comments?

RCdeWit:
Thanks for the review, folks. Incorporated the feedback!

daavoo (commented):
Not saying it should block the P.R. but we should probably test for this (maybe in https://github.com/iterative/dvc/blob/30d6839a5f471fb02a7a2dcbd6a6616b71277762/tests/unit/remote/test_gdrive.py)

RCdeWit:
I'm working on part 2 of a blog post that covers this use case. I was hoping to release it somewhere over the next two weeks; any chance this fix will be merged by then? If not, my blog post will have to include the workaround.

```
# GDRIVE_CREDENTIALS_DATA is not read automatically when using a service account
          # Manually write it to creds.json and read that file
          echo -E '${{ secrets.GDRIVE_CREDENTIALS_DATA }}' > creds.json

          dvc remote add -d -f myremote gdrive://${{ secrets.GOOGLE_DRIVE_URI }}
          dvc remote modify myremote gdrive_use_service_account true
          dvc remote modify myremote --local gdrive_service_account_json_file_path creds.json

          dvc push

          # Either do this or add the file to your .gitignore
          # Just make sure not to push it to your repository
          rm creds.json
```

daavoo (approved):

PR #7214, opened 2021-12-31
repo: use fs instead of scm

efiop:
SCM is used for both reading and writing operations, but in the former case -
as filesystem and in the latter - as actual scm/git. So we should distinguish
those and use fs directly, which follows the overall philosophy of being able
to use Repo on any fs.

Keeping `rev=` option for `Repo` because it is a very convenient way to get
repo instance from local path but for particular branch/tag/etc.

Pre-requisite for git/dvc/repofs refactorings.

### DIFF
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -30,6 +30,7 @@ def external_repo(
     from scmrepo.git import Git
 
     from dvc.config import NoRemoteError
+    from dvc.fs.git import GitFileSystem
 
     logger.debug("Creating external repo %s@%s", url, rev)
     path = _cached_clone(url, rev, for_write=for_write)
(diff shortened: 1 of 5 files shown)


dvc/external_repo.py
```diff
@@ -86,7 +93,7 @@ def erepo_factory(url, cache_config, *args, **kwargs):
     def make_repo(path, **_kwargs):
         _config = cache_config.copy()
         if os.path.isdir(url):
-            rel = os.path.relpath(path, _kwargs["scm"].root_dir)
+            rel = os.path.relpath(path, _kwargs["fs"].fs_args["scm"].root_dir)
```
efiop, line 96: This ugliness will go away when we switch to repo paths in gitfs instead of abspaths.

pared (approved):

dvc/external_repo.py
```diff
@@ -45,12 +46,18 @@ def external_repo(
     config = _get_remote_config(url) if os.path.isdir(url) else {}
     config.update(cache_config)
 
-    root_dir = path if for_write else os.path.realpath(path)
+    if for_write:
+        root_dir = path
+        fs = None
```
skshetry, line 51: Probably we should get rid of `for_write`, it might already be broken.

efiop: Yeah, it is only used in dvcx and it has been effectively abandoned :( But write-functionality (virtual, without real checked-out repo) would be great for gitfs in the future.

PR #7221, opened 2022-01-04
output: don't raise RemoteCacheRequiredError on instantiation, make it lazy

skshetry:
As the title says. Had to adjust tests as before we were raising this issue during instantiation which meant we were not checking for `external=True` there, but now stage validation takes priority and needs explicitly `external=True` for the test to work/make sense. 

This is backported from https://github.com/iterative/dvc/pull/6868.

### DIFF
--- a/dvc/output.py
+++ b/dvc/output.py
@@ -326,9 +326,6 @@ class Output:
         self.desc = desc
 
         self.fs_path = self._parse_path(self.fs, fs_path)
-        if self.use_cache and self.odb is None:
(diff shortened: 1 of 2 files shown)


dvc/output.py
```diff
@@ -402,7 +399,10 @@ def use_scm_ignore(self):
 
     @property
```
skshetry, line 400: We could cache it too, if performance is a concern here.

pmrowla (approved):

Issue #5563, opened 2021-03-08
state: incompatible across different python versions


skshetry:
### Steps to reproduce

```bash
python3.9 -m dvc diff -v
python3.6 -m dvc diff -v
```

#### Output

```console
$ python3.9 -m dvc diff -v
2021-03-08 16:09:20,120 DEBUG: Check for update is disabled.
2021-03-08 16:09:20,202 DEBUG: state save (24563398, 1611922311010025216, 211598) 90104d9e83cfb825cf45507e90aadd27
2021-03-08 16:09:20,243 DEBUG: state save (24566815, 1611922311020025344, 12238) 94614d6650e062655f9f77507dc9c1f2
2021-03-08 16:09:20,248 DEBUG: state save (24563444, 1611922311016691968, 206122) 9106cda08aa427e73492389a0f17c72d
2021-03-08 16:09:20,417 DEBUG: Analytics is disabled.
$ python3.6 -m dvc diff -v
dvc diff -v
2021-03-08 16:09:42,589 DEBUG: Check for update is disabled.
2021-03-08 16:09:42,714 ERROR: unexpected error - unsupported pickle protocol: 5
------------------------------------------------------------
Traceback (most recent call last):
  File "/home/saugat/repos/iterative/dvc/dvc/main.py", line 50, in main
    ret = cmd.run()
  File "/home/saugat/repos/iterative/dvc/dvc/command/diff.py", line 138, in run
    self.args.a_rev, self.args.b_rev, self.args.targets
  File "/home/saugat/repos/iterative/dvc/dvc/repo/__init__.py", line 49, in wrapper
    return f(repo, *args, **kwargs)
  File "/home/saugat/repos/iterative/dvc/dvc/repo/diff.py", line 45, in diff
    results[rev] = _paths_checksums(self, repo_fs, targets_path_infos)
  File "/home/saugat/repos/iterative/dvc/dvc/repo/diff.py", line 110, in _paths_checksums
    return dict(_output_paths(repo, repo_fs, targets))
  File "/home/saugat/repos/iterative/dvc/dvc/repo/diff.py", line 145, in _output_paths
    yield _to_path(output), _to_checksum(output)
  File "/home/saugat/repos/iterative/dvc/dvc/repo/diff.py", line 133, in _to_checksum
    repo.odb.local, output.path_info, repo.odb.local.fs,
  File "/home/saugat/repos/iterative/dvc/dvc/objects/stage.py", line 152, in stage
[...]

PR #7222, opened 2022-01-04
odb: pin index pickle protocol version

### LINKED ISSUE #7220: odb: index pickle version should be explicit
Currently we don't specify a pickle version when we use `diskcache` in the ODB/remote index, so we will potentially run into the same problem as: https://github.com/iterative/dvc/issues/5563

discord context: https://discord.com/channels/485586884165107732/485596304961962003/927613942908465162

pmrowla:

Fixes #7220

### DIFF
--- a/dvc/objects/db/index.py
+++ b/dvc/objects/db/index.py
@@ -88,7 +88,7 @@ class ObjectDBIndex(ObjectDBIndexBase):
         tmp_dir: "StrPath",
         name: str,
     ):  # pylint: disable=super-init-not-called
-        from diskcache import Index
+        from diskcache import Cache, Index
 
         from dvc.fs.local import LocalFileSystem
         from dvc.utils.fs import makedirs
@@ -96,7 +96,13 @@ class ObjectDBIndex(ObjectDBIndexBase):
         self.index_dir = os.path.join(tmp_dir, self.INDEX_DIR, name)
         makedirs(self.index_dir, exist_ok=True)
         self.fs = LocalFileSystem()
-        self.index = Index(self.index_dir)
+        self.index = Index.fromcache(
+            Cache(
+                self.index_dir,
+                disk_pickle_protocol=4,
+                eviction_policy="none",
+            )
+        )
 
     def __iter__(self):
         return iter(self.index)


dvc/objects/db/index.py
```diff
@@ -96,7 +96,7 @@ def __init__(
         self.index_dir = os.path.join(tmp_dir, self.INDEX_DIR, name)
         makedirs(self.index_dir, exist_ok=True)
         self.fs = LocalFileSystem()
-        self.index = Index(self.index_dir)
+        self.index = Index(self.index_dir, disk_pickle_protocol=4)
```
skshetry, line 99: @pmrowla, will this break repo if they were using protocol 5 before?

pmrowla: Yes, users may need to remove `.dvc/tmp/index` in this case (which is the same potential issue we had when we pinned the version for state https://github.com/iterative/dvc/issues/5563#issuecomment-792728682)

skshetry: Python >=3.8 uses Protocol version 5, so we may as well just don't support compatibility between 3.7 and >=3.8.
What do you think of doing something like:
```suggestion
        self.index = Index(
            self.index_dir,
            disk_pickle_protocol=4 if sys.version_info < (3,8) else 5
        )
```

pmrowla: I'm not sure that works here since the reason users are still running into this issue is because they are mixing python versions (between 3.7 and >3.7)

pmrowla: Also, the core python [default protocol](https://docs.python.org/3/library/pickle.html#pickle.DEFAULT_PROTOCOL) version in 3.8+ is still 4, the problem for us is that diskcache explicitly uses `pickle.HIGHEST_PROTOCOL` https://github.com/grantjenks/python-diskcache/blob/d55a50ee083784afa9c85e14e41c4a2d132f3111/diskcache/core.py#L60

skshetry: I'm just worried that it may crash for users in the new release unless they remove `.dvc/tmp/index`, although I did not really see any exceptions when I migrated to change it to 4. I guess that works already?

> Also, the core python default protocol version in 3.8+ is still 4, the problem for us is that diskcache explicitly uses pickle.HIGHEST_PROTOCOL grantjenks/python-diskcache@d55a50e/diskcache/core.py#L60

yes, but the highest protocol in 3.8+ is 5.

pmrowla: It's not backwards compatible, but it won't raise any exceptions until you actually try to read/write anything from disk

karajan1001: For myself, I'd prefer to raise some messages to suggest `.dvc/tmp/index` here.

efiop (approved):

skshetry:
To be honest, I don't find it worth it to break for every python version, considering it may fail in unexpected places/ways.
As 3.7 is the minimum supported version and 3.0 is around, we could choose to leave it as-is, and just not support mixing it with other python versions.

dvc/objects/db/index.py
pmrowla: Regarding pickle protocol version, I think we should ensure compatibility across python versions that are supported by DVC, which means we should be enforcing a pinned version wherever we are using pickle (unless at some point in the future Python officially deprecates whatever ver we are using).

In the state/index use case, I don't think there's any practical benefit to using protocol ver 5 for the way we use `diskcache` (my understanding of protocol ver 5 is that it improves performance for serializing very large non-native objects, like numpy arrays and pandas dataframes), but with the way we use diskcache we are really just serializing (relatively) small native dicts.

There is an official backport library for pickle protocol v5, so if/when the state/index pickle performance becomes a real problem for us we can go that route, but I don't think it's a pressing need right now.

skshetry: @pmrowla, I agree with all of your points, I have no issues on the technical side of things.

I'm just worried about potentially breaking things here for Python 3.8+ users in the middle of 2.x version. I'm fine with this as long as we are explicit about this being a breaking change and weigh in both options:

1. By not fixing this, we break compatibility between 3.7 with 3.8+ Python versions. So, if the user used Python 3.8+ version, they cannot use dvc repo again with Python 3.7. 
2. By fixing this, we ensure compatibility between all Python versions by breaking compatibility for once for Python 3.8+ users.

pmrowla: @skshetry as a short term workaround we could just catch the protocol errors in state/index, and then add a troubleshooting link w/details suggesting users can try removing the relevant temp dirs (`md5s/links/index`) and then retry the command

pmrowla: So after some testing, it turns out that changing to protocol version 4 will not break an existing diskcache cache that was generated using protocol version 5 (as long as you are on a python version that supports v5). It will just rewrite the cache entry using protocol version 4 on the next time that entry is modified.

This test:
```py
@pytest.mark.parametrize("proto_a, proto_b", [(4, 5), (5, 4)])
def test_protocol_error(tmp_dir, proto_a, proto_b):
    with diskcache.Cache(directory=(tmp_dir / "test"), disk_pickle_protocol=proto_a) as cache:
        cache["key"] = ("value1", "value2")
        assert ("value1", "value2")  == cache["key"]
    with diskcache.Cache(directory=(tmp_dir / "test"), disk_pickle_protocol=proto_b) as cache:
        cache["key"] = ("value3", "value4")
        assert ("value3", "value4")  == cache["key"]
```
passes on 3.8, 3.9, 3.10 and fails on 3.7 (since 3.7 cannot read/write protocol version 5)

So this should not break compatibility with existing repos for anyone on python 3.8+ (but generating the troubleshooting message for py 3.7 will still be useful)

Issue #7120, opened 2021-12-09
DVC list FileNotFoundError


nikshar-symbio:
# Bug Report

## Description
When trying to list the files using `dvc list .`, I get the following error
```
2021-12-01 11:32:18,645 ERROR: unexpected error
------------------------------------------------------------
Traceback (most recent call last):
  File "/home/nikhil/.virtualenvs/init_py38/lib/python3.8/site-packages/dvc/main.py", line 55, in main
    ret = cmd.do_run()
  File "/home/nikhil/.virtualenvs/init_py38/lib/python3.8/site-packages/dvc/command/base.py", line 64, in do_run
    return self.run()
  File "/home/nikhil/.virtualenvs/init_py38/lib/python3.8/site-packages/dvc/command/ls/__init__.py", line 30, in run
    entries = Repo.ls(
  File "/home/nikhil/.virtualenvs/init_py38/lib/python3.8/site-packages/dvc/repo/ls.py", line 36, in ls
    ret = _ls(repo.repo_fs, path_info, recursive, dvc_only)
  File "/home/nikhil/.virtualenvs/init_py38/lib/python3.8/site-packages/dvc/repo/ls.py", line 69, in _ls
    metadata = fs.metadata(info)
  File "/home/nikhil/.virtualenvs/init_py38/lib/python3.8/site-packages/dvc/fs/repo.py", line 409, in metadata
    raise FileNotFoundError
FileNotFoundError
```

I've tried updating to the latest `dvc` version, and tried this using a fresh clone of the git repo, but the error still exists.

### Reproduce

The git repo is private and I can't really share the contents, but I basically do a `git clone <repo>`, and do a `dvc list . -v`.

### Expected

It should list the files tracked by `dvc`.

### Environment information

`dvc==2.8.3`

**Output of `dvc doctor`:**

```console
$ dvc doctor

DVC version: 2.8.3 (pip)
---------------------------------
Platform: Python 3.8.11 on Linux-5.10.35-rt39-symbio-x86_64-with-glibc2.27
Supports:
	webhdfs (fsspec = 2021.11.1),
	http (aiohttp = 3.7.4.post0, aiohttp-retry = 2.4.6),
	https (aiohttp = 3.7.4.post0, aiohttp-retry = 2.4.6),
	s3 (s3fs = 2021.6.1, boto3 = 1.17.49)
Cache types: hardlink, symlink
Cache directory: ext4 on /dev/nvme0n1p5
Caches: local
Remotes: s3
[...]

PR #7226, opened 2022-01-05
ls: handle broken symlink

pared:
Fixes: #7120

### DIFF
--- a/dvc/repo/ls.py
+++ b/dvc/repo/ls.py
@@ -33,7 +33,7 @@ def ls(url, path=None, rev=None, recursive=None, dvc_only=False):
         if path:
             fs_path = os.path.abspath(repo.fs.path.join(fs_path, path))
 
-        ret = _ls(repo.repo_fs, fs_path, recursive, dvc_only)
+        ret = _ls(repo, fs_path, recursive, dvc_only)
 
         if path and not ret:
             raise PathMissingError(path, repo, dvc_only=dvc_only)
@@ -46,10 +46,13 @@ def ls(url, path=None, rev=None, recursive=None, dvc_only=False):
         return ret_list
 
 
-def _ls(fs, fs_path, recursive=None, dvc_only=False):
+def _ls(repo, fs_path, recursive=None, dvc_only=False):
+    from dvc.fs._metadata import Metadata
+
     def onerror(exc):
         raise exc
 
+    fs = repo.repo_fs
     infos = []
     try:
         for root, dirs, files in fs.walk(
@@ -66,7 +69,12 @@ def _ls(fs, fs_path, recursive=None, dvc_only=False):
 
     ret = {}
     for info in infos:
-        metadata = fs.metadata(info)
+        try:
+            metadata = fs.metadata(info)
+        except FileNotFoundError:
+            # broken symlink
+            metadata = Metadata(info, repo)
+
         if metadata.output_exists or not dvc_only:
             path = (
                 fs.path.name(fs_path)
--- a/tests/func/test_ls.py
+++ b/tests/func/test_ls.py
@@ -557,3 +557,24 @@ def test_subrepo(dvc_top_level, erepo):
     assert _list_files(subrepo, ".") == common_outputs
     assert _list_files(subrepo, "scm_dir") == {"ipsum"}
     assert _list_files(subrepo, "dvc_dir") == {"lorem"}
+
+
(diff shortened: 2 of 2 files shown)


dvc/repo/ls.py
```diff
@@ -66,16 +66,28 @@ def onerror(exc):
...
+
+        except FileNotFoundError:
+            # if walk found it but there is no metadata, its broken symlink
+            if not dvc_only:
+                ret[path] = {
+                    "isout": False,
+                    "isdir": False,
+                    "isexec": False,
+                }
+
```
efiop, line 92: Note that you are ignoring FileNotFoundError if `dvc_only`, which could be a dangerous bug. Also duplicating code, so how about smth like:

```
        try:
            metadata = fs.metadata(info)
        except FileNotFoundError:
            if dvc_only:
                raise
                
            metadata = Metadata()
            
        if metadata.output_exists or not dvc_only:
            path = (
                fs.path.name(fs_path)
                if fs_path == info
                else fs.path.relpath(info, fs_path)
            )
            ret[path] = {
                "isout": metadata.is_output,
                "isdir": metadata.isdir,
                "isexec": metadata.is_exec,
            }
```

pared:
> Btw, how does CLI ls handle broken symlinks? Should we colorize it too? If so, how hard would that be?

CLI ls handles it the same way as symlinks - we get a color for symlink. 
I think that we should not consider that in that in this change: so far we haven't been differentiating between files/dirs and links. 

We could consider implementing it. The problem with that is that currently, we are differentiating the colors by determining whether the target is output/dir/executable. If none of the above, we use an extension to determine it. So far there has been no talking about links. In order to provide support, we would need to pass `islink` returned by `fsspec.info` up to the Metadata and leverage it to assign a color.

The question is whether we should do that. CLI ls is about inspecting the filesystem so it makes sense to mark the links. However the DVC is about our ML project - the most important is to differentiate between tracked and non tracked assets. I think w should not expand that. Even now, the behavior might be a bit counterintuitive. For example: if our pipeline will be producing executable, it will still be colored as output, because it is the first check we do.  I think in this case we should not follow `ls` style.

As an alternative, we could consider how to "mix" colors depending on the multiple properties of a given asset.

dvc/repo/ls.py
```diff
@@ -66,7 +69,19 @@ def onerror(exc):
 
     ret = {}
     for info in infos:
-        metadata = fs.metadata(info)
+        path = (
+            fs.path.name(fs_path)
+            if fs_path == info
+            else fs.path.relpath(info, fs_path)
+        )
```
efiop, line 76: Doesn't look used until defined below
```suggestion
```

dvc/repo/ls.py
```diff
@@ -66,7 +69,19 @@ def onerror(exc):
...
+        try:
+            metadata = fs.metadata(info)
+        except FileNotFoundError:
+            if dvc_only:
+                raise
+
+            metadata = Metadata(fs_path, repo)
```
efiop, line 83:
```suggestion
            metadata = Metadata(info, repo)
```

dvc/repo/ls.py
```diff
@@ -66,7 +69,19 @@ def onerror(exc):
...
+            fs.path.name(fs_path)
+            if fs_path == info
+            else fs.path.relpath(info, fs_path)
+        )
+        try:
+            metadata = fs.metadata(info)
+        except FileNotFoundError:
+            if dvc_only:
+                raise
+
```
efiop, line 82:
```suggestion
            # broken symlink
```

dvc/repo/ls.py
```diff
@@ -66,7 +69,15 @@ def onerror(exc):
 
     ret = {}
     for info in infos:
-        metadata = fs.metadata(info)
+        try:
+            metadata = fs.metadata(info)
+        except FileNotFoundError:
+            if dvc_only:
+                raise
```
pared, line 76: With the current approach (creating empty Metadata) I think it makes sense to remove this check.
it shouldn't matter if we use `dvc_only` or not. In my initial version lack of this was a bug, silently skipping the error, but now it seems ok to remove it. Users should not be punished for a specific check. If there is a broken symlink, other commands will make sure of erroring out

pared: Also, in the case of `dvc_only` files, we should never get here: repo_fs will be able to get the metadata even for the removed file

efiop (approved):

PR #7229, opened 2022-01-06
handle paths ending with sep in dvc.fs.path.Path.parts

### LINKED ISSUE #7228: dvc.fs.Path.parts wrong results
**EDIT**: This issue will just be for this first problem of handling a sep at the end of a path. I made the windows-style path problem a separate issue #7233 

When a path ends with the path sep, the `parts` function doesn't split. It returns a tuple with a single item:
```python
from dvc.fs.path import Path
Path('/').parts('/a/b/c/')
```
```python
('/a/b/c',)
```

A second problem occurs when using windows style paths. We get the sep between the drive and the rest of the path:
```python
Path('\\').parts('c:\\a')
```
```python
('c:', '\\', 'a')
```

The first problem could be solved by simply stripping the final sep:
```python
        drive, path = self.flavour.splitdrive(path.rstrip(self.flavour.sep))
```
but the second problem would still exist.

We should really get these results:
```python
Path('/').parts('/a/b/c/')
```
```python
('/', 'a', 'b', 'c')
```
and
```python
Path('\\').parts('c:\\a')
```
```python
('c:', 'a')
```

Note the second case is still a little different from pathlib, which would include the sep with the drive:
```python
from pathlib import PureWindowsPath
PureWindowsPath('c:\\a').parts
```
```python
('c:\\', 'a')
```
but this is probably more in-line with fsspec, which basically treats the drive letter as the first element of a relative path:
```python
fsspec.AbstractFileSystem._parent('c:/a')
``` 
```python
'c:'
``` 

version info:
```
DVC version: 2.9.4.dev28+gd90fe54d.d20220106 
Platform: Python 3.10.1 on Linux-5.15.11-arch2-1-x86_64-with-glibc2.33
Supports:
	azure (adlfs = 2021.10.0, knack = 0.9.0, azure-identity = 1.7.1),
	gdrive (pydrive2 = 1.10.0),
	gs (gcsfs = 2021.11.1),
	hdfs (fsspec = 2021.11.1, pyarrow = 6.0.1),
	webhdfs (fsspec = 2021.11.1),
	s3 (s3fs = 2021.11.1, boto3 = 1.19.8),
	ssh (sshfs = 2021.11.2),
	oss (ossfs = 2021.8.0),
	webdav (webdav4 = 0.9.3),
	webdavs (webdav4 = 0.9.3)
Caches: local
Remotes: https
Workspace directory: btrfs on /dev/mapper/nvme0n1p3_crypt
Repo: dvc, git
```

jonburdo:
Fixes #7228

### DIFF
--- a/dvc/fs/path.py
+++ b/dvc/fs/path.py
@@ -1,5 +1,6 @@
 import ntpath
 import posixpath
+import re
 
 
 class Path:
@@ -10,31 +11,15 @@ class Path:
             self.flavour = ntpath
         else:
             raise ValueError(f"unsupported separator '{sep}'")
+        self.sep_re = re.compile(f"{re.escape(sep)}+")
 
     def join(self, *parts):
         return self.flavour.join(*parts)
 
     def parts(self, path):
-        drive, path = self.flavour.splitdrive(path)
-
-        ret = []
-        while True:
-            path, part = self.flavour.split(path)
-
-            if part:
-                ret.append(part)
-                continue
-
-            if path:
-                ret.append(path)
-
-            break
-
-        ret.reverse()
-
-        if drive:
-            ret = [drive] + ret
-
+        ret = self.sep_re.split(path.strip(self.flavour.sep))
+        if path.startswith(self.flavour.sep):
+            ret = [self.flavour.sep] + ret
         return tuple(ret)
 
     def parent(self, path):


dvc/fs/path.py
```diff
@@ -10,31 +11,15 @@ def __init__(self, sep):
             self.flavour = ntpath
         else:
             raise ValueError(f"unsupported separator '{sep}'")
+        self.sep_re = re.compile(f"{re.escape(sep)}+")
```
efiop, line 14: switching to regexp is a significant change, we would need to at the very least check the performance (e.g. split into parts 1M paths or smth like that).

dvc/fs/path.py
```diff
@@ -10,31 +11,15 @@ def __init__(self, sep):
...
-            break
-
-        ret.reverse()
-
-        if drive:
-            ret = [drive] + ret
-
+        ret = self.sep_re.split(path.strip(self.flavour.sep))
+        if path.startswith(self.flavour.sep):
+            ret = [self.flavour.sep] + ret
```
efiop, line 22: and looks like it broke windows tests 🙁 Looks like the minimal change is to just rstrip the sep from the path, right?

dvc/fs/path.py
jonburdo: Here's a gist with both versions (original with rstrip as `parts` and re version as `parts_re`) https://gist.github.com/jonburdo/d45db255710152704951bcdcf9503a16

I ran a little ipython test on my machine on almost 2.5M paths:
```python
In [1]: from path_parts_comparison import Path

In [2]: import os

In [3]: from itertools import chain

In [4]: paths = list(chain(*[[d + os.sep] + [os.path.join(d, f) for f in files] for d, _, files in os.walk(os.path.expanduser('~'))]))

In [5]: ppath = Path('/')

In [6]: %time a = [ppath.parts(p) for p in paths]
CPU times: user 9.02 s, sys: 168 ms, total: 9.19 s
Wall time: 9.19 s

In [7]: %time b = [ppath.parts_re(p) for p in paths]
CPU times: user 2.8 s, sys: 222 ms, total: 3.03 s
Wall time: 3.03 s

In [8]: %time a = [ppath.parts(p) for p in paths]
CPU times: user 8.86 s, sys: 242 ms, total: 9.1 s
Wall time: 9.1 s

In [9]: %time b = [ppath.parts_re(p) for p in paths]
CPU times: user 2.97 s, sys: 190 ms, total: 3.16 s
Wall time: 3.16 s

In [10]: a == b
Out[10]: True

In [11]: a[:6]
Out[11]: 
[('/', 'home', 'jon'),
 ('/', 'home', 'jon', '.bash_history'),
 ('/', 'home', 'jon', '.bash_aliases'),
 ('/', 'home', 'jon', '.bash_logout'),
 ('/', 'home', 'jon', '.yarnrc'),
 ('/', 'home', 'jon', '.bash_profile')]

In [12]: paths[:6]
Out[12]: 
['/home/jon/',
 '/home/jon/.bash_history',
 '/home/jon/.bash_aliases',
 '/home/jon/.bash_logout',
 '/home/jon/.yarnrc',
 '/home/jon/.bash_profile']

In [13]: len(paths)
Out[13]: 2469677
```

Looks to me like regex is about 3 times faster. Obviously not a perfect test, but the regex is pretty simple and I think the repeated function calls in the current implementation add up.

This probably needs some more testing and debugging for the windows-style paths, but could be worth working off at some point.

dvc/fs/path.py
jonburdo: Yeah, let's go with the minimal fix for now. I don't want to spend a lot of time debugging the windows tests now. I made a separate PR since I still think this might be worth debugging and working off of at some point. Let me know if you don't think so, but I think the current implementation still doesn't work as expected for windows paths.

Here's a PR with the minimal fix. Looks like it passes the tests: https://github.com/iterative/dvc/pull/7234

dtrifiro:
Fixed with https://github.com/iterative/dvc/pull/7234

PR #7232, opened 2022-01-06
feat: allow user annotations in .dvc files

### LINKED ISSUE #7231: feat: allow user annotations in .dvc files
We would like to be able to associate arbitrary tags and key=value annotations with particular versions of DVC-managed files. For example, if `a.csv` and `b.csv` are managed by DVC:

```
$ dvc note set -K color -V green a.csv

$ dvc note set -K size -V medium a.csv

$ dvc note list a.csv
color: green
size: medium

$ dvc note set -K color -V blue b.csv

$ dvc note find -K color *.csv
a.csv: green
b.csv: blue

$ dvc note remove -K color a.csv

$ dvc note find -K color *.csv
b.csv: blue
```

This can be implemented by adding a `notes` section under the `meta` key in the `.dvc` file without affecting any existing functionality. While it wouldn't necessarily be part of DVC, we are excited by the possibility of having a post-commit hook in Git that constructs a search index from these tags and annotations on every commit.

gvwilson:

* [ ] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here. - **I will create the documentation PR if there is interest in adding this feature.**

- Add new `note` sub-command to set, remove, list, and find key/value notes in .dvc files.
- Extend CLI to handle operations on single and multiple files.
- Add tests for sub-command and CLI.

Closes #7231 

Questions:

1. This PR reads/writes YAML using `load_yaml` and `dump_yaml` from
   `dvc.utils.serialize` - should it use the classes from `dvc.dvcfile`
   instead, and if so, which one(s)?

2. `dvc.repo.note` uses `@locked` to ensure atomic operation across
   multiple files. Is this correct? (The aim is to make sure that
   if a key is added, deleted, or modified in multiple files, the
   operations affects all files or none in event of partial failure.)

3. The syntax is `dvc note add -K key -V value file...` An alternative
   would be `dvc note add key=value file...`, which is shorter to type
   but will cause problems if keys need to contain `=` (or any other
   marker character). Should the shorter syntax be added?

4. `note` only handles flat annotations: all keys are directly beneath
   `meta.note` in the YAML. Should nested keys be allowed so that (for
   example) users can do `dvc note add -K hi.mid.lo -V val file`?

5. `note` only handles a single value for each key. This is enough for
   our use cases, but other may want multiple values per key. Should it
   be added now, or should we wait to see whether the feature is popular
   and if so, whether other users want this?


efiop:
For the record: putting this on hold until we clear up the requirements in https://github.com/iterative/dvc/issues/7231

pared:
The original issue has been closed for now. Closing this PR for now. We will reopen if we get back to #7231

PR #7237, opened 2022-01-07
tests: clean repositories before each test runs

skshetry:
While investigating this failure: https://github.com/iterative/dvc/runs/4735592203?check_suite_focus=true#step:7:4910,
I noticed that we don't clean repos after each test runs, which meant that when some test later on called `main()` function, the function might try cleaning up multiple repositories at the end, which caused timeout for `exp init` tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -57,6 +57,14 @@ def enable_ui():
     ui.enable()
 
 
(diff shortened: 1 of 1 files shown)


skshetry:
Testing if only clearing the state is faster ...

Issue #7153, opened 2021-12-15
add the new flags (including `-A/--all-commits`, `-a/--all-branches`, `--all-tags`) to exp list


karajan1001:
_Originally posted by @karajan1001 in https://github.com/iterative/dvc/issues/5489#issuecomment-992234534_

PR #7206, opened 2021-12-28
`exp pull/push`: add multi exp name arg support 

karajan1001:
fix #7196 
New feature: now `dvc exp push/pull` can accept a list of experiment names as input.

1. add support to list input for the `exp push/pull`
2. add function tests for the multi `exp push/pull`
3. add unit test for the arg parse

* [x] ❗ I have followed the [Contributing to DVC](https://dvc.org/doc/user-guide/contributing/core) checklist.

* [ ] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here.

Thank you for the contribution - we'll try to review it as soon as possible. 🙏

wait #7205

PR #7245, opened 2022-01-10
exp list: add the new flags  (including -A/--all-commits) (#7153)

karajan1001:
fix: #7153
Unify the flags in `exp list` to `exp show` (including -A/--all-commits
)
Wait #7152 and #7206

1. Add new `-A`,`-n` flags to `exp list` command.
2. Add unit test for the flag parse.
3. unify the revs retrive logic to `exp show`.
4. Add functional test for `exp list`

### DIFF
--- a/dvc/commands/experiments/__init__.py
+++ b/dvc/commands/experiments/__init__.py
@@ -58,3 +58,43 @@ def add_parser(subparsers, parent_parser):
     for cmd in SUB_COMMANDS:
         cmd.add_parser(experiments_subparsers, parent_parser)
     fix_plumbing_subparsers(experiments_subparsers)
+
+
+def add_rev_selection_flags(
+    experiments_subcmd_parser, command: str, default: bool = True
+):
+    experiments_subcmd_parser.add_argument(
+        "-A",
+        "--all-commits",
+        action="store_true",
+        default=False,
+        help=(
+            f"{command} all experiments in the repository "
+            "(overrides `--rev` and `--num`)."
+        ),
+    )
+    default_msg = " (HEAD by default)" if default else ""
+    msg = (
+        f"{command} experiments derived from the specified `<commit>` as "
+        f"baseline{default_msg}."
+    )
+    experiments_subcmd_parser.add_argument(
+        "--rev",
+        type=str,
+        default=None,
+        help=msg,
+        metavar="<commit>",
+    )
+    experiments_subcmd_parser.add_argument(
+        "-n",
+        "--num",
+        type=int,
+        default=1,
+        dest="num",
+        metavar="<num>",
+        help=(
+            f"{command} experiments from the `--rev` baseline and from `num` "
+            "commits before it (first parents). Give a negative value to "
+            "include all first-parent commits."
+        ),
+    )
--- a/dvc/repo/experiments/ls.py
+++ b/dvc/repo/experiments/ls.py
@@ -3,49 +3,36 @@ from collections import defaultdict
 
 from dvc.repo import locked
 from dvc.repo.scm_context import scm_context
+from dvc.scm import iter_revs
+from dvc.types import Optional
 
-from .utils import (
-    exp_refs,
-    exp_refs_by_baseline,
-    remote_exp_refs,
-    remote_exp_refs_by_baseline,
-)
+from .utils import exp_refs, exp_refs_by_baseline
 
 logger = logging.getLogger(__name__)
 
 
 @locked
 @scm_context
-def ls(repo, *args, rev=None, git_remote=None, all_=False, **kwargs):
-    from scmrepo.git import Git
-
-    from dvc.scm import RevError, resolve_rev
-
-    if rev:
-        try:
-            rev = resolve_rev(repo.scm, rev)
-        except RevError:
-            if not (git_remote and Git.is_sha(rev)):
-                # This could be a remote rev that has not been fetched yet
-                raise
-    elif not all_:
-        rev = repo.scm.get_rev()
-
+def ls(
+    repo,
+    rev: Optional[str] = None,
+    all_commits: bool = False,
+    num: int = 1,
+    git_remote: Optional[str] = None,
+):
     results = defaultdict(list)
-
-    if rev:
-        if git_remote:
-            gen = remote_exp_refs_by_baseline(repo.scm, git_remote, rev)
-        else:
-            gen = exp_refs_by_baseline(repo.scm, rev)
-        for info in gen:
-            results[rev].append(info.name)
-    elif all_:
-        if git_remote:
-            gen = remote_exp_refs(repo.scm, git_remote)
-        else:
-            gen = exp_refs(repo.scm)
+    if all_commits:
+        gen = exp_refs(repo.scm, git_remote)
         for info in gen:
             results[info.baseline_sha].append(info.name)
+        return results
+
+    rev = rev or "HEAD"
+
+    revs = iter_revs(repo.scm, [rev], num)
+    rev_set = set(revs.keys())
+    ref_info_dict = exp_refs_by_baseline(repo.scm, rev_set, git_remote)
+    for rev, ref_info_list in ref_info_dict.items():
+        results[rev] = [ref_info.name for ref_info in ref_info_list]
 
     return results
--- a/dvc/scm.py
+++ b/dvc/scm.py
@@ -8,7 +8,7 @@ from scmrepo.base import Base  # noqa: F401, pylint: disable=unused-import
 from scmrepo.git import Git
 from scmrepo.noscm import NoSCM
 
-from dvc.exceptions import DvcException, InvalidArgumentError
+from dvc.exceptions import DvcException
 from dvc.progress import Tqdm
 
 if TYPE_CHECKING:
@@ -136,7 +136,7 @@ def resolve_rev(scm: "Git", rev: str) -> str:
 
 def iter_revs(
     scm: "Git",
-    head_revs: Optional[List[str]] = None,
+    revs: Optional[List[str]] = None,
     num: int = 1,
     all_branches: bool = False,
     all_tags: bool = False,
@@ -144,42 +144,39 @@ def iter_revs(
     all_experiments: bool = False,
 ) -> Mapping[str, List[str]]:
 
-    if num < 1 and num != -1:
-        raise InvalidArgumentError(f"Invalid number of commits '{num}'")
-
-    if not any(
-        [head_revs, all_branches, all_tags, all_commits, all_experiments]
-    ):
+    if not any([revs, all_branches, all_tags, all_commits, all_experiments]):
         return {}
 
-    head_revs = head_revs or []
-    revs = []
-    for rev in head_revs:
-        revs.append(rev)
+    revs = revs or []
+    results = []
+    for rev in revs:
+        if num == 0:
+            continue
+        results.append(rev)
         n = 1
         while True:
             if num == n:
                 break
             try:
                 head = f"{rev}~{n}"
-                revs.append(resolve_rev(scm, head))
+                results.append(resolve_rev(scm, head))
             except RevError:
                 break
             n += 1
 
     if all_commits:
-        revs.extend(scm.list_all_commits())
+        results.extend(scm.list_all_commits())
     else:
         if all_branches:
-            revs.extend(scm.list_branches())
+            results.extend(scm.list_branches())
 
         if all_tags:
-            revs.extend(scm.list_tags())
+            results.extend(scm.list_tags())
 
     if all_experiments:
         from dvc.repo.experiments.utils import exp_commits
 
-        revs.extend(exp_commits(scm))
+        results.extend(exp_commits(scm))
 
     rev_resolver = partial(resolve_rev, scm)
-    return group_by(rev_resolver, revs)
+    return group_by(rev_resolver, results)
--- a/tests/func/experiments/test_remote.py
+++ b/tests/func/experiments/test_remote.py
@@ -132,12 +132,13 @@ def test_list_remote(tmp_dir, scm, dvc, git_downstream, exp_stage, use_url):
     downstream_exp = git_downstream.tmp_dir.dvc.experiments
     assert downstream_exp.ls(git_remote=remote) == {}
 
+    git_downstream.tmp_dir.scm.fetch_refspecs(remote, ["master:master"])
     exp_list = downstream_exp.ls(rev=baseline_a, git_remote=remote)
     assert {key: set(val) for key, val in exp_list.items()} == {
         baseline_a: {ref_info_a.name, ref_info_b.name}
     }
 
-    exp_list = downstream_exp.ls(all_=True, git_remote=remote)
+    exp_list = downstream_exp.ls(all_commits=True, git_remote=remote)
     assert {key: set(val) for key, val in exp_list.items()} == {
         baseline_a: {ref_info_a.name, ref_info_b.name},
         baseline_c: {ref_info_c.name},
--- a/dvc/commands/experiments/ls.py
+++ b/dvc/commands/experiments/ls.py
@@ -11,9 +11,10 @@ class CmdExperimentsList(CmdBase):
     def run(self):
         names_only = self.args.names_only
         exps = self.repo.experiments.ls(
+            all_commits=self.args.all_commits,
             rev=self.args.rev,
+            num=self.args.num,
             git_remote=self.args.git_remote,
-            all_=self.args.all,
         )
         for baseline in exps:
             tag = self.repo.scm.describe(baseline)
@@ -32,6 +33,7 @@ class CmdExperimentsList(CmdBase):
 
 
 def add_parser(experiments_subparsers, parent_parser):
+    from . import add_rev_selection_flags
 
     EXPERIMENTS_LIST_HELP = "List local and remote experiments."
     experiments_list_parser = experiments_subparsers.add_parser(
@@ -41,19 +43,7 @@ def add_parser(experiments_subparsers, parent_parser):
         help=EXPERIMENTS_LIST_HELP,
         formatter_class=argparse.RawDescriptionHelpFormatter,
     )
-    experiments_list_parser.add_argument(
-        "--rev",
-        type=str,
-        default=None,
-        help=(
-            "List experiments derived from the specified revision. "
-            "Defaults to HEAD if neither `--rev` nor `--all` are specified."
-        ),
-        metavar="<rev>",
-    )
-    experiments_list_parser.add_argument(
-        "--all", action="store_true", help="List all experiments."
-    )
+    add_rev_selection_flags(experiments_list_parser, "List")
     experiments_list_parser.add_argument(
         "--names-only",
         action="store_true",
--- a/dvc/commands/experiments/show.py
+++ b/dvc/commands/experiments/show.py
@@ -515,6 +515,8 @@ class CmdExperimentsShow(CmdBase):
 
 
 def add_parser(experiments_subparsers, parent_parser):
+    from . import add_rev_selection_flags
+
     EXPERIMENTS_SHOW_HELP = "Print experiments."
     experiments_show_parser = experiments_subparsers.add_parser(
         "show",
@@ -523,6 +525,7 @@ def add_parser(experiments_subparsers, parent_parser):
         help=EXPERIMENTS_SHOW_HELP,
         formatter_class=argparse.RawDescriptionHelpFormatter,
     )
+    add_rev_selection_flags(experiments_show_parser, "Show")
     experiments_show_parser.add_argument(
         "-a",
         "--all-branches",
@@ -536,32 +539,6 @@ def add_parser(experiments_subparsers, parent_parser):
         action="store_true",
         default=False,
         help="Show experiments derived from all Git tags.",
-    )
-    experiments_show_parser.add_argument(
-        "-A",
-        "--all-commits",
-        action="store_true",
-        default=False,
-        help="Show experiments derived from all Git commits.",
-    )
-    experiments_show_parser.add_argument(
-        "--rev",
-        type=str,
-        default=None,
-        help=(
-            "Show experiments derived from the specified revision. "
-            "Defaults to HEAD if none of `--rev`,`-a`,`-A`,`-T` is specified."
-        ),
-        metavar="<rev>",
-    )
-    experiments_show_parser.add_argument(
-        "-n",
-        "--num",
-        type=int,
-        default=1,
-        dest="num",
-        metavar="<num>",
-        help="Show the last `num` commits from <rev>.",
     )
     experiments_show_parser.add_argument(
         "--no-pager",
--- a/dvc/repo/experiments/utils.py
+++ b/dvc/repo/experiments/utils.py
@@ -5,6 +5,7 @@ from typing import (
     Generator,
     Iterable,
     List,
+    Mapping,
     Optional,
     Set,
     Union,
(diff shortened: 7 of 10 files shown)


tests/func/experiments/test_remote.py
```diff
@@ -121,12 +121,13 @@ def test_list_remote(tmp_dir, scm, dvc, git_downstream, exp_stage, use_url):
     downstream_exp = git_downstream.tmp_dir.dvc.experiments
     assert downstream_exp.ls(git_remote=remote) == {}
 
-    exp_list = downstream_exp.ls(rev=baseline_a, git_remote=remote)
+    git_downstream.tmp_dir.scm.fetch_refspecs(remote, ["master:master"])
```
karajan1001, line 124: One question here, currently the rev collection is on the local side. And the git tree in local might not equal to the remote. It is better to collect the revs from the remote directly, but we don't have a related `scm` function yet?

pmrowla: I think for now we should not support using `--all-branches/--all-tags` with remote exp listing. Only supporting the existing use case (with `--rev` and `--all-commits` in place of `--all`) for remote listing should be fine for now.

karajan1001: My opinion:
1. `pull` and `push` are two sides of one coin, should stay the same here. 
2. `list` can be used for previews on other commands that should support all of the flags here.
3.  for `show`, `num` can be used to show the results improving progress,   `all_branches`, `all_tags`, `all_commits` are also useful in showing the different versions of models.
4. `pull` and `push` should support all of the flags in `show` for the experiments sharing.
5. `rev` is useful for `remove` to clean up the result after we fixed the parameters. `all_commits` which is supported now is used to clean up a repo. But I don't think that `all_tags` and `all_branches` have any actual use case here, same with `num` .

A summary:

|  cmd  |  list  |  show  |  pull/push  |  remove  |
|---|---|---|---|---|
|  --rev | yes  | yes  |  yes  |  yes  |
|  --num | yes  | yes  |  yes  |  no  |
|  --all_commits |  yes  | yes |  yes  | yes  |
|  --all_tags  |  yes  |  yes |  yes  |  no  |
|  --all_branches  |  yes  | yes  |  yes  |  no  |

@dberenbaum  What do you think of it?

dberenbaum: 😅 Thanks for the great synopsis @karajan1001!

Seems like the only questionable part is the flags for pull/push, mainly because some of them seem somewhat unlikely and potentially very long-running. I don't have a strong opinion on `--num`, `--all-commits`, `--all-tags`, `--all-branches` for push/pull. It's possible but unlikely that someone would do `dvc exp pull --all-tags; dvc exp show --all-tags`. Whatever you and @pmrowla decide on those is fine with me.

pmrowla: The issue is that `--all-tags`/`--all-branches`/`--all-commits` is ambiguous in the remote command use case. It is unclear whether the command should use the local repo's refs or the remote repo's refs in order to determine what constitutes "all tags/branches/commits", because they will not always be the same thing.

I may have a lot of local dev branches that I don't want pushed to a github repo, and the github repo may have things like release/backport/testing branches that I don't want to pull into my local copy.

So when I do `dvc exp push --all-branches`, does that mean I want to push all experiments based on all of my local branches (including ones that I have not also pushed to github)?

And on `dvc exp pull --all-branches`, does that mean I want to pull experiments from branches which only exist in the remote (and that I have not also `git pulled` locally)?

pmrowla: IMO, push/pull should be explicit, and in the branch/tag case we should require the user to use `--rev branch-name`. If they want to push/pull for all commits in that branch they can use `--rev branch-name -n -1`

karajan1001: Agreed, `tags` and `branches` might not be synchronized between remote and local, we should not support them in the remote use case, then no need in `list` also. I'll implement the following version.

|  cmd  |  list  |  show  |  pull/push  |  remove  |
|---|---|---|---|---|
|  --rev | yes  | yes  |  yes  |  yes  |
|  --num | yes  | yes  |  yes  |  yes  |
|  --all_commits |  yes  | yes |  yes  | yes  |
|  --all_tags  |  no  |  yes |  no  |  no  |
|  --all_branches  |  no  | yes  |  no  |  no  |

dberenbaum: So what happens if `push/pull --all-commits` includes branches that aren't on one of local/remote? It's useful for `list` since that only operates on the remote. `push/pull` operate on both, so maybe we should remove `push/pull --all-commits`?

Also, no strong opinion, but any reason why you decided to add support for `remove --num` in the updated table?

karajan1001: > So what happens if push/pull --all-commits includes branches that aren't on one of local/remote? It's useful for list since that only operates on the remote. push/pull operate on both, so maybe we should remove push/pull --all-commits?

The original solution follows the from repo (local in `push`, remote in `pull`), the same with the results in `list`. But for now, an experiment whose baseline doesn't exist in the repo will be limited on many commands. 

> Also, no strong opinion, but any reason why you decided to add support for remove --num in the updated table?

No very strong reasons, only because:
 1. It might be useful when cleaning a branch of experiments, and it is harmless.
 2. Forbiddening requires extra code, which is unnecessary.

pmrowla (dismissed):
This will need a docs PR.

@karajan1001 it would also be helpful to put together an asciinema demo of the new flags (it would be fine to have one demo containing all of the different commands instead of a separate demo for each of these PRs)

dvc/scm.py
```diff
@@ -132,54 +132,73 @@ def resolve_rev(scm: "Git", rev: str) -> str:
         raise RevError(str(exc))
 
 
+def last_n_commits(
```
karajan1001, line 135: Add this new function for the use of `list/push/pull/remove`. As we only have two sets of experiments retrieve flags (`rev`+`num`/`all-commits`), and the `all-commits` are well supported by our current utils. We only need this to get our experiment's collections logic unified instead of using the heavy `iter_revs` below.

dvc/repo/experiments/ls.py
```diff
@@ -3,49 +3,35 @@
...
-            gen = exp_refs_by_baseline(repo.scm, rev)
-        for info in gen:
-            results[rev].append(info.name)
-    elif all_:
-        if git_remote:
-            gen = remote_exp_refs(repo.scm, git_remote)
-        else:
-            gen = exp_refs(repo.scm)
+    if all_commits:
+        gen = exp_refs(repo.scm, git_remote)
```
iesahin, line 25: I'm reading this due to [discussion](https://github.com/iterative/dvc.org/pull/3191#discussion_r787060961) in iterative/dvc.org#3191 

`all_commits` seems to contain `refs/exps/` like experiment references, right?

dvc/command/experiments/utils.py
```diff
@@ -0,0 +1,32 @@
...
+        action="store_true",
+        default=False,
+        help=(
+            f"{command} all experiments in the repository "
+            "(overrides `--rev`)."
+        ),
+    )
+    msg = f"{command} experiments derived from the specified Git commit"
+    if default:
+        msg += "(instead of `HEAD`)."
```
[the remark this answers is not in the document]
jorgeorpinel: Maybe we should hint more, i.e. "specified Git commit, branch, tag, etc." ?

dvc/command/experiments/utils.py
```diff
@@ -0,0 +1,32 @@
...
+        metavar="<commit>",
+    )
+    experiments_subcmd_parser.add_argument(
+        "-n",
+        "--num",
+        type=int,
+        default=1,
+        dest="num",
+        metavar="<num>",
+        help=f"{command} the last `num` commits from <commit>.",
```
jorgeorpinel, line 31: Should we indicate the default?
```suggestion
        help=f"{command} the last `num` commits from <commit> "
        "(HEAD by default).",
```
Unless this only works together with `--rev`

jorgeorpinel (commented):
Minor suggestions

dvc/command/experiments/utils.py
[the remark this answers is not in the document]
jorgeorpinel: p.s. in docs (https://dvc.org/doc/command-reference/exp/show#-n) we say "from HEAD". That's incorrect right? Incomplete rather

dvc/command/experiments/utils.py
jorgeorpinel: p.s. in docs (https://dvc.org/doc/command-reference/exp/show#-n) se say "from HEAD". That's incorrect right? Incomplete rather

dvc/command/experiments/utils.py
```diff
@@ -0,0 +1,32 @@
...
+    experiments_subcmd_parser, command: str, default: bool = True
+):
+    experiments_subcmd_parser.add_argument(
+        "-A",
+        "--all-commits",
+        action="store_true",
+        default=False,
+        help=(
+            f"{command} all experiments in the repository "
+            "(overrides `--rev`)."
```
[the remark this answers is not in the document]
jorgeorpinel: Actually, -A overrides both --rev and --num right?

dvc/command/experiments/utils.py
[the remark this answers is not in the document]
jorgeorpinel: And here, actually I'm not sure we want conditional help output differences?

dvc/command/experiments/utils.py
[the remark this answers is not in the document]
karajan1001: > 👍 assuming `command` starts with capital letter i.e. "Show" or "List".

Yes, I use capital letters here.

> Actually, -A overrides both --rev and --num right?

Of course, --num only works together with `--rev`.

dvc/repo/experiments/ls.py
karajan1001: > I'm reading this due to [discussion](https://github.com/iterative/dvc.org/pull/3191#discussion_r787060961) in [iterative/dvc.org#3191](https://github.com/iterative/dvc.org/pull/3191)
> 
> `all_commits` seems to contain `refs/exps/` like experiment references, right?

`all_commits` are user commits here, while for `refs/exps/` they are generated by dvc as experiments although the storage format is `commits`.

dvc/command/experiments/utils.py
karajan1001: Oh, sorry, there is a problem, here it is `commits` instead of `experiments`.  Because here what this `-n` flag do is actually extend the baseline `rev` to its history commits. And if the `-n` is -1 then it represents the whole history of a given revision (a tag, a branch, or a commit). And the experiments based on any of the commits from the history will be listed here.

karajan1001:
add a new demo for it.

[asciicast](https://asciinema.org/a/cRD2BhdeBX8ZyQcjDNTnN5eTa)

dvc/command/experiments/utils.py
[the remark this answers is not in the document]
jorgeorpinel: Hey @karajan1001 so let's say `overrides --rev and -A` to be specific?

dvc/command/experiments/utils.py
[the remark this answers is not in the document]
jorgeorpinel: Unresolving this one since there was no change (it's missing a period when `!default`) or answer.

Still wondering about the conditional help output though... Shouldn't it always be the static? (I'm not sure what the fn-level `default` var is for here.) Cc @dberenbaum

dvc/command/experiments/utils.py
jorgeorpinel: > what this -n flag do is actually extend the baseline rev to its history commits...
if the -n is -1 then it represents the whole history of a given revision

Got it, thanks. May leave another suggestion then... ⌛

dvc/commands/experiments/utils.py
```diff
@@ -0,0 +1,35 @@
...
+    experiments_subcmd_parser.add_argument(
+        "-n",
+        "--num",
+        type=int,
+        default=1,
+        dest="num",
+        metavar="<num>",
+        help=(
+            f"{command} list experiments from the `<rev>` commit and `num` "
+            "commits before it"
```
jorgeorpinel, line 33: Per https://github.com/iterative/dvc/pull/7245#discussion_r789400084
```suggestion
            f"{command} experiments from the `--rev` baseline and from `num` "
            "commits before it. Give `-1` to include all the parent commits."
```

jorgeorpinel (commented):
More help output clarifications

dvc/scm.py
karajan1001: @pmrowla What do you think of it?
1. Now in  `exp list/push/pull/remove` we only support `--rev`, `-n` and `--all-commits`. For `all-commits` the process of it is quite different from any other args (we don't need to collect conditions), iter over all exps are just enough for it. A good example is in `remove` where the `--all-commits` is simpler than `--rev`, I don't think it is a good design to unify them ( first collect all revs, then match them for every exps).
2. Currently, `iter_revs` only used in `brancher` and `exp show`. Because in `exp list/push/pull/remove` we do not need to support `all-tags` and `all-branches`. Do we still need this? or we can just use `last_n_commits` plus `all-tags` and `all-branches` and `all-commits` in `brancher` and `exp show`, or maybe it is useful in some other commands (for example `gc` and `exp gc`)

dvc/command/experiments/utils.py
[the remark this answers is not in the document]
karajan1001: Do you mean `overrides --rev and -N`?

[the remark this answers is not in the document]
jorgeorpinel: Yes sorry @karajan1001 I meant ``(overrides `--rev` and `--num`)``

dvc/commands/experiments/__init__.py
```diff
@@ -58,3 +58,42 @@ def add_parser(subparsers, parent_parser):
...
+        default=False,
+        help=(
+            f"{command} all experiments in the repository "
+            "(overrides `--rev`and `--num`)."
+        ),
+    )
+    default_msg = "(HEAD by default)" if default else ""
+    msg = (
+        f"{command} experiments derived from the specified `<commit>` as "
+        f"baseline{default_msg}."
```
jorgeorpinel, line 79: Minor: missing a space I think 🙂 
```suggestion
    default_msg = " (HEAD by default)" if default else ""
    msg = (
        f"{command} experiments derived from the specified `<commit>` as "
        f"baseline{default_msg}."
```

dvc/command/experiments/utils.py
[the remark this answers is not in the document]
jorgeorpinel: Got it, thanks @karajan1001 !

dvc/commands/experiments/__init__.py
jorgeorpinel: But... Will it be clear that ``from the specified `<commit>` `` refers to using `--rev` ?

jorgeorpinel: UPDATE: Nvmd on that question, I see this IS the help output for --rev 😁

dvc/commands/experiments/__init__.py
```diff
@@ -58,3 +58,42 @@ def add_parser(subparsers, parent_parser):
...
+    experiments_subcmd_parser, command: str, default: bool = True
+):
+    experiments_subcmd_parser.add_argument(
+        "-A",
+        "--all-commits",
+        action="store_true",
+        default=False,
+        help=(
+            f"{command} all experiments in the repository "
+            "(overrides `--rev`and `--num`)."
```
jorgeorpinel, line 73: Missing another space
```suggestion
            "(overrides `--rev` and `--num`)."
```

dvc/repo/experiments/ls.py
iesahin: So it's the same commits as in `git log -n`?

karajan1001: > So it's the same commits as in `git log -n`?

Yes, it is.

dvc/commands/experiments/__init__.py
```diff
@@ -58,3 +58,42 @@ def add_parser(subparsers, parent_parser):
...
+    experiments_subcmd_parser.add_argument(
+        "-n",
+        "--num",
+        type=int,
+        default=1,
+        dest="num",
+        metavar="<num>",
+        help=(
+            f"{command} experiments from the `--rev` baseline and from `num` "
+            "commits before it. Give `-1` to include all the parent commits."
```
jorgeorpinel, line 97: From the long discussion in https://github.com/iterative/dvc.org/pull/3191#pullrequestreview-855848503 it seems that the behavior around `n=1` has an issue: it should show/list experiments from the given `--rev <commit>` (`HEAD` by default) AND from it's parent commit (the `1` value). Otherwise there's no way to get list/show that configuration.

The default value should be `n=0` for the current default behavior. Thanks

dvc/repo/experiments/ls.py
dberenbaum: Regarding `-n 0`, `-n -1`, etc., `git log -n 0` returns nothing, while `git log -n -1` and `git log -n -2` return all history. If we want consistency with `git log -n`, let's go with this behavior.

dvc/commands/experiments/__init__.py
karajan1001: Maybe we should follow the `git` behavior? 

https://github.com/iterative/dvc/pull/7245#discussion_r797801555

dberenbaum: Thanks, @karajan1001. @jorgeorpinel, please ping if you still have questions/concerns. Otherwise, I think we should move forward with the git behavior.

dvc/scm.py
pmrowla: I think the real issue is that `iter_revs` is still heavy in the first place, when it probably shouldn't be.

I think what we really need is another refactor of `iter_revs` and the SCM `list_branches/list_tags/etc` functions so that they actually use iterators and not lists/dicts. (But this is not high priority right now, and still goes along with the need for a `brancher` refactor in general).

I think it is fine for `exp list/push/pull/etc` to continue using `iter_revs` rather than introducing another utility here, because eventually we should really be solving this problem in `iter_revs` rather than introducing more helper functions.

@karajan1001

karajan1001: > I think it is fine for exp list/push/pull/etc to continue using iter_revs rather than introducing another utility here, because eventually we should really be solving this problem in iter_revs rather than introducing more helper functions.

Actually what I suggest is to abandon `iter_revs` and just use a lightweight `last_n_commits` function. Now argument  `all_experiments` only used in `brancher` beside this it makes the function in `dvc.scm.py` depends on something from `dvc/repo/experiments/utils/py`, while all other dependencies are in an opposite direction, I think this indicates here is a wrong place for it.

`all_branches`, `all_tags`, they only exist in `exp show` and `brancher`, and we only call them from scm
```python
        if all_branches:
            revs.extend(scm.list_branches())
            results.extend(scm.list_branches())

        if all_tags:
            revs.extend(scm.list_tags())
            results.extend(scm.list_tags())
```

The only two universal flags here are `all_commits` and `revs`. and if the command has an all_commits flag, it is better to directly collect experiments during the iteration and do not check their baselines for performance reasons. So the only logic I think should be separated here is the `rev` + `num`.

pmrowla: You are probably right that `all_experiments` doesn't belong in `iter_revs`, since it's specific to brancher. But otherwise, ideally, `iter_revs` should be just as lightweight as `last_n_commits` (and it should not work by extending a list).

If you want to keep `last_n_commits` separate for now, that's fine with me, but I do think that eventually (once we revisit brancher) that functionality will end up being folded back into a refactored `iter_revs`.

For this PR though, please change `last_n_commits` to be a generator, there is no need to add the overhead of a list here

dvc/commands/experiments/__init__.py
jorgeorpinel: > git log -n 0 returns nothing, while git log -n -1 and git log -n -2 return all history. If we want consistency with git

Agreed.

I'd also insist that the metavar name can be improved, at least to `<number>` (consistent with [git log](https://git-scm.com/docs/git-log#_commit_limiting)).

dvc/repo/experiments/ls.py
jorgeorpinel: One question on `-n` though, `git log` defaults to no value there, meaning full history. We default to `-n 0`. Is there a way to indicate full history or is that left for some other future update?

dberenbaum: Yes, like `git log -n -1` and `git log -n -2`, `dvc exp list -n -1` or `dvc exp list -n -2` should indicate full history.

dvc/commands/experiments/__init__.py
```diff
@@ -58,3 +58,43 @@ def add_parser(subparsers, parent_parser):
...
+        "-n",
+        "--num",
+        type=int,
+        default=1,
+        dest="num",
+        metavar="<num>",
+        help=(
+            f"{command} experiments from the `--rev` baseline and from `num` "
+            "commits before it. Give a negative value to include all the "
+            "parent commits."
```
jorgeorpinel, line 98: This probably needs an update similar to https://github.com/iterative/dvc.org/pull/3187#pullrequestreview-883999298 ?

jorgeorpinel: Also, I have a question: what happens when `--rev` is a merge? Which ancestors are included by the `<num>`?

pmrowla: It will include the first parent for merge commits (linear ancestors, equivalent to using tilde `~` in git). We don't support specifying a different parent (i.e carat `^` in git).

So for `--rev HEAD -n 2` You would get the commit referenced by `HEAD~2` in git (as opposed to `HEAD~2^2/HEAD~2^3...`

see: https://stackoverflow.com/a/43046393/1538451

dvc/repo/experiments/ls.py
jorgeorpinel: Not sure I follow. By "full history" I mean to include all ancestor commits, equivalent to `--all-commits` if there's a single branch in the repo.

I guess if you give a really large number like `-n -9999` that excedes the no. of ancestor commits that should do the trick.

dvc/commands/experiments/__init__.py
jorgeorpinel: Got it, thanks @pmrowla ! I wonder if we want to include this info in the description or in docs. May be relevant if people use large `-n` values (likely to include merges).

dvc/repo/experiments/ls.py
pmrowla: I think the suggestion was that `-n` with any negative number should give all ancestors of the specified revision the same way that it does in git (it doesn't matter if it's `-1` or `-99999`). `-A/--all-commits` is different since it applies to all branches in the repo.

karajan1001: > I think the suggestion was that `-n` with any negative number should give all ancestors of the specified revision the same way that it does in git (it doesn't matter if it's `-1` or `-99999`). `-A/--all-commits` is different since it applies to all branches in the repo.

The current discussion on this is that we follow the results from `git log -n`, `0` for a empty list, a negative value for the whole history, positive value for a list with `n` members.

dberenbaum:
@karajan1001 What's the status of this one? Seems like this one needs to be merged before moving forward with the related core and docs PRs.

karajan1001:
@jorgeorpinel,  Need some final check and approval. this one is the first one to be merged.

dberenbaum:
@karajan1001 I still see that `-n -1` fails, as mentioned in https://github.com/iterative/dvc/pull/7204#issuecomment-1021431886. I also thought that change was going to be rolled back, but I don't see any rollback. What's the status?

karajan1001:
> @karajan1001 I still see that `-n -1` fails, as mentioned in [#7204 (comment)](https://github.com/iterative/dvc/pull/7204#issuecomment-1021431886). I also thought that change was going to be rolled back, but I don't see any rollback. What's the status?

need to release a new version.
https://github.com/iterative/scmrepo/pull/36

dvc/commands/experiments/__init__.py
jorgeorpinel, line 98: Per https://github.com/iterative/dvc/pull/7245#pullrequestreview-884000297
```suggestion
            "commits before it (first parents). Give a negative value to "
            "include all first-parent commits."
```

jorgeorpinel (approved):
Output strings look good, thanks.

pmrowla (approved):

PR #7247, opened 2022-01-10
dvc: separate data and objects

efiop:
Initial simple separation to get the process of extracting dvc-objects
and dvc-data packages out of dvc.

dvc-objects is for:
 * defining base Object (aka HashFile) and ObjectDB
 * providing tools to work with those objects (check/transfer/etc)
 * etc

dvc-data is for:
 * defining higher level objects (e.g. Tree)
 * bulding objects from the workspace (aka staging)
 * projecting objects back into the workspace (aka checkout)
 * comparing Tree objects (aka diff)
 * defining misc object databases (e.g LocalObjectDB)
 * higher-level transfer (e.g. with Tree indexes)
 * etc

For example, the objects we use in run-cache are based on dvc-objects
and not on dvc-data, as they have absolutely nothing to do with data
management.

There are still some mixups in the code that will be addressed in the
followups.


dvc/utils/table.py
```diff
@@ -21,7 +21,7 @@ class Column(RichColumn):
 
 
 class Table(RichTable):
-    def add_column(  # pylint: disable=arguments-differ
+    def add_column(  # type: ignore[override] # pylint: disable=arguments-differ,arguments-renamed # noqa: E501
```
skshetry, line 24: The linter is raising an appropriate error message. Also this should be upstreamed, see: https://github.com/iterative/dvc/pull/5753#issuecomment-811822900.

efiop: @skshetry Indeed, should've looked deeper into that. @pmrowla any plans to upstream it?

Issue #7154, opened 2021-12-15
 make all the flags in `exp show` and `exp list`  support pull/push


karajan1001:
_Originally posted by @karajan1001 in https://github.com/iterative/dvc/issues/5489#issuecomment-992234534_

PR #7255, opened 2022-01-11
exp push/pull: make all the flags in exp show and exp list support pull/push (#7154)

karajan1001:
wait for  #7245

exp push/pull: make all the flags in exp show and exp list support pull/push (#7154)

fix: #7154
1. add all flags (A/--all-commits , -a/--all-branches, --all-tags) in `exp show` to the `exp pull/push`
2. add unit and func tests for `exp push/pull`

### DIFF
--- a/dvc/commands/experiments/pull.py
+++ b/dvc/commands/experiments/pull.py
@@ -3,17 +3,31 @@ import logging
 
 from dvc.cli.command import CmdBase
 from dvc.cli.utils import append_doc_link
+from dvc.exceptions import InvalidArgumentError
 from dvc.ui import ui
 
 logger = logging.getLogger(__name__)
 
 
 class CmdExperimentsPull(CmdBase):
-    def run(self):
+    def raise_error_if_all_disabled(self):
+        if not any(
+            [self.args.experiment, self.args.all_commits, self.args.rev]
+        ):
+            raise InvalidArgumentError(
+                "Either provide an `experiment` argument, or use the "
+                "`--rev` or `--all-commits` flag."
+            )
 
-        self.repo.experiments.pull(
+    def run(self):
+        self.raise_error_if_all_disabled()
+
+        pulled_exps = self.repo.experiments.pull(
             self.args.git_remote,
             self.args.experiment,
+            all_commits=self.args.all_commits,
+            rev=self.args.rev,
+            num=self.args.num,
             force=self.args.force,
             pull_cache=self.args.pull_cache,
             dvc_remote=self.args.dvc_remote,
@@ -21,10 +35,13 @@ class CmdExperimentsPull(CmdBase):
             run_cache=self.args.run_cache,
         )
 
-        ui.write(
-            f"Pulled experiment '{self.args.experiment}'",
-            f"from Git remote '{self.args.git_remote}'.",
-        )
+        if pulled_exps:
+            ui.write(
+                f"Pulled experiment '{pulled_exps}'",
+                f"from Git remote '{self.args.git_remote}'.",
+            )
+        else:
+            ui.write("No experiments to pull.")
         if not self.args.pull_cache:
             ui.write(
                 "To pull cached outputs for this experiment"
@@ -36,6 +53,8 @@ class CmdExperimentsPull(CmdBase):
 
 
 def add_parser(experiments_subparsers, parent_parser):
+    from . import add_rev_selection_flags
+
     EXPERIMENTS_PULL_HELP = "Pull an experiment from a Git remote."
     experiments_pull_parser = experiments_subparsers.add_parser(
         "pull",
@@ -44,6 +63,7 @@ def add_parser(experiments_subparsers, parent_parser):
         help=EXPERIMENTS_PULL_HELP,
         formatter_class=argparse.RawDescriptionHelpFormatter,
     )
+    add_rev_selection_flags(experiments_pull_parser, "Pull", False)
     experiments_pull_parser.add_argument(
         "-f",
         "--force",
@@ -89,7 +109,8 @@ def add_parser(experiments_subparsers, parent_parser):
     )
     experiments_pull_parser.add_argument(
         "experiment",
-        nargs="+",
+        nargs="*",
+        default=None,
         help="Experiments to pull.",
         metavar="<experiment>",
     )
--- a/dvc/commands/experiments/push.py
+++ b/dvc/commands/experiments/push.py
@@ -4,17 +4,32 @@ import logging
 from dvc.cli.command import CmdBase
 from dvc.cli.utils import append_doc_link
 from dvc.commands import completion
+from dvc.exceptions import InvalidArgumentError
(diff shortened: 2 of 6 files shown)


karajan1001:
Demos for the push 
[asciicast](https://asciinema.org/a/34FZguA5R0XQa2iNQbwNsupUn)
for the pull
[asciicast](https://asciinema.org/a/gNcVfpqMTgnRxzvqg1GMznp62)

dvc/commands/experiments/utils.py
```diff
@@ -0,0 +1,35 @@
+def add_rev_selection_flags(
```
skshetry, line 1: You can move these to [`dvc/commands/experiments/__init__.py`](https://github.com/iterative/dvc/blob/47d10c4a8b9bdc18dec17a9380f0593e43383c4c/dvc/commands/experiments/__init__.py) and just import it as:
```python
from . import add_rev_selection_flags
```

dberenbaum:
Nice work @karajan1001! I have a few questions:

1. Should there be an error if `dvc exp push/pull` are run without any arguments?

2. When there are no experiments to push, the output doesn't look right:
```
$ dvc exp push --no-cache myremote
Pushed experiment '[]'to Git remote 'myremote'.

To push cached outputs for this experiment to DVC remote storage,re-run this command without '--no-cache'.
```

We should recognize that no experiments have actually been pushed and make the output something like `No experiments pushed.`

3. `num` only seems to work when combined with `--rev`. Should it also be able to work from the current commit (like `dvc exp show -n 2`)?

dvc/commands/experiments/utils.py
karajan1001: Yes but looks like that would cause a circular import, so we must place this import inside `add_parser`.

karajan1001:
> 1. Should there be an error if `dvc exp push/pull` are run without any arguments?

In my opinion, For `exp push/pull/remove/gc` they are not a frequent cmd and will cause exp status changes, users must explicitly give the exps to be operated on, and should raise error if none arguments are given.

> 2. When there are no experiments to push, the output doesn't look right:
> 
> ```
> $ dvc exp push --no-cache myremote
> Pushed experiment '[]'to Git remote 'myremote'.
> 
> To push cached outputs for this experiment to DVC remote storage,re-run this command without '--no-cache'.
> ```
> 
> We should recognize that no experiments have actually been pushed and make the output something like `No experiments pushed.`

Yeah.

> 3. `num` only seems to work when combined with `--rev`. Should it also be able to work from the current commit (like `dvc exp show -n 2`)?

Yes, it works, `dvc exp show -n 2` will show the exps derive from `HEAD` and `HEAD~`

karajan1001:
> > In my opinion, For `exp push/pull/remove/gc` they are not a frequent cmd and will cause exp status changes, users must explicitly give the exps to be operated on, and should raise error if none arguments are given.
> 
> 👍
> 
> > > We should recognize that no experiments have actually been pushed and make the output something like `No experiments pushed.`
> > 
> > 
> > Yeah.
> 
> 👍
> 
> > > 3. `num` only seems to work when combined with `--rev`. Should it also be able to work from the current commit (like `dvc exp show -n 2`)?
> > 
> > 
> > Yes, it works, `dvc exp show -n 2` will show the exps derive from `HEAD` and `HEAD~`
> 
> It works for me in `dvc exp show` but not in these other commands. For example:
> 
> ```
> $ dvc exp pull -n 2 myremote
> Pulled experiment '[]' from Git remote 'myremote'.
> $ dvc exp pull --rev HEAD -n 2 myremote
> Pulled experiment '['exp-45e9f']' from Git remote 'myremote'.
> ```

Are you in the latest version? I had already removed the default `HEAD` support. And this is the latest result from my computer.

```
> dvc exp pull -n 2 myremote
ERROR: Either of `experiment`, `--all-commits` or `--rev` needs to be set.
> dvc exp pull --rev HEAD -n 2 origin
Pulled experiment '['exp0.15', 'exp0.25', 'exp0.3', 'exp0.1']' from Git remote 'origin'.
>  dvc exp pull --rev HEAD~~ origin
No experiments to pull.
```

dberenbaum:
Sorry, must have been behind. Looks good now! One last thing:

```
$ dvc exp pull
ERROR: the following arguments are required: <git_remote>, <experiment>
```

Should `<experiment>` be included as a required argument?

karajan1001:
> Sorry, must have been behind. Looks good now! One last thing:
> 
> ```
> $ dvc exp pull
> ERROR: the following arguments are required: <git_remote>, <experiment>
> ```
> 
> Should `<experiment>` be included as a required argument?

This error comes from `argparse` I had set made `experiment` to be `*` which means it can be 0 to any integer.

dberenbaum:
> add a default value for the arguments can solve this problem

Can the default be `None`?

karajan1001:
> > add a default value for the arguments can solve this problem
> 
> Can the default be `None`?

I had tried several cases on my computer. It works fine for me.

dvc/commands/experiments/__init__.py
```diff
@@ -58,3 +58,43 @@ def add_parser(subparsers, parent_parser):
...
+        "-n",
+        "--num",
+        type=int,
+        default=1,
+        dest="num",
+        metavar="<num>",
+        help=(
+            f"{command} experiments from the `--rev` baseline and from `num` "
+            "commits before it. Give a negative value to include all the "
+            "parent commits."
```
jorgeorpinel, line 98: See https://github.com/iterative/dvc/pull/7245#pullrequestreview-902522092

dvc/commands/experiments/pull.py
```diff
@@ -3,28 +3,45 @@
...
 
 
 class CmdExperimentsPull(CmdBase):
+    def raise_error_if_all_disabled(self):
+        if not any(
+            [self.args.experiment, self.args.all_commits, self.args.rev]
+        ):
+            raise InvalidArgumentError(
+                "Either of `experiment`, `--all-commits` "
+                "or `--rev` needs to be set."
```
jorgeorpinel, line 19: Same as https://github.com/iterative/dvc/pull/7275#pullrequestreview-902546358

jorgeorpinel: > Other comments in that PR also apply here.

dvc/commands/experiments/__init__.py
jorgeorpinel: > Other comments in that PR also apply here.

dvc/commands/experiments/pull.py
jorgeorpinel: > Other comments in that PR also apply here.

karajan1001: Also updated in push.py

jorgeorpinel (approved):
UI strings LGTM 👍🏼

karajan1001:
@pmrowla This one needs to be merged first before #7275.

pmrowla (approved):

PR #7256, opened 2022-01-11
plots: Fix `--json` requiring `-o`.

daavoo:
`-o` was being required even when renderers don't need output path.

### DIFF
--- a/tests/unit/command/test_plots.py
+++ b/tests/unit/command/test_plots.py
@@ -297,3 +297,48 @@ def test_plots_diff_json(dvc, mocker, capsys):
     show_json_mock.assert_called_once_with(renderers, "out")
 
     render_mock.assert_not_called()
+
+
+def test_show_json_requires_out(dvc, mocker, capsys):
+    cli_args = parse_args(
+        [
+            "plots",
+            "diff",
+            "HEAD~10",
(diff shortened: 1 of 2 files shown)


tests/unit/command/test_plots.py
```diff
@@ -297,3 +298,16 @@ def test_plots_diff_json(dvc, mocker, capsys):
     show_json_mock.assert_called_once_with(renderers, "out")
 
     render_mock.assert_not_called()
+
+
+def test_show_json_requires_out(mocker):
```
skshetry, line 303: This test may not be useful in the future and tests private function. Maybe we can just test `--json` through cli?

pared (approved):

tests/unit/command/test_plots.py
```diff
@@ -297,3 +297,48 @@ def test_plots_diff_json(dvc, mocker, capsys):
...
+            "plots",
+            "diff",
+            "HEAD~10",
+            "HEAD~1",
+            "--json",
+            "--targets",
+            "plot.csv",
+        ]
+    )
+    cmd = cli_args.func(cli_args)
```
skshetry, line 314: You can just use `main` here.

Issue #6489, opened 2021-08-25
Support for external commands?


skshetry:
It'd be amazing to have a support for external command being able to run `dvc <cmd>` by just having a `dvc-<cmd>` in the user's `PATH`, similar to `git`.

Looking at the argparse, it does not seem to be possible to add support for default subparser or a fallback subparser. So, we may be need to add all the commands from the `PATH` matching `dvc-` prefix during `ArgumentParser` build. Also may need to take into account a few things for windows
(eg: `PATHEXT`, current directory has higher precedence, multiple executable extensions, etc.).

Also see: https://github.com/iterative/cml/issues/699.

daavoo:
What about also allowing users to register custom commands via entry-points:

https://packaging.python.org/guides/creating-and-discovering-plugins/#using-package-metadata

skshetry:
Closing, so far there has not been a usecase. `live` command was deprecated/removed.

Issue #7040, opened 2021-11-24
Introduce DVC aliases


gennaro-tedesco:
`git` configuration puts at your disposal the possibility to define aliases of the like
```
$ .gitconfig

[alias]
    mycommand = ...
```
to be used as `git mycommand ...`. It could be convenient to mirror the same behaviour in DVC and allow the user to define aliases for their most common operations: for instance I find myself very often running `dvc list -R --dvc-only .` or `dvc repro --glob /pattern` and I would gladly alias them as, say,
```
$ /.dvc/config

[core]
    remote = storage
['remote "storage"']
    ...
['alias']
    files = list -R --dvc-only
    train = repro --glob train_*
```
Of course one can make the argument that all of this can be achieved by means of global shell aliases, however it is cleaner to have framework specific aliases configured within the corresponding tools only to avoid polluting the shell with a lot of unnecessary alias expansions and similar.

Issue #1500, opened 2019-01-14
dvc: introduce `dvc md5` or `dvc checksum` command


efiop:
It would compute checksum for a file/directory the same way we do it in dvc. Useful for debugging purposes.

dmpetrov:
A single command for a debugging purpose sounds like a too big API change. Is there a way to hide this command under some other command like `status` (which checks checksums) or `cache`?

A related question... We have a separate `dvc destroy` command but it looks like a revert operation to `dvc init`. Why don't we move it under `init` umbrella like `dvc init --destroy`?

efiop:
Yeah, I agree. I was even thinking about the long gone "fsck" command, but I still don't like it :) It should indeed simply be a part of status output.

Related https://github.com/iterative/dvc/issues/1202 .

About the `destroy`. We already have it and there is no better use for that name as of today, so there is little to no incentive as to move it under `dvc init`. :)

mroutis:
I'm about to work on this issue, I'm up to adding a new `dvc checksum` command, as `status` is currently for stage files, whereas `checksum` would be for files and directories.

efiop:
@mroutis let's skip this one for now.

efiop:
Closing as we now have `dvc commit` https://github.com/iterative/dvc/issues/919 and we no longer need to be able to construct checksums.

mroutis:
@efiop , it would be a nice to have, tho; kind of like a _plumbing_ command

efiop:
@mroutis Don't really see where it would be needed, and no one is asking for it after `dvc commit` was implemented.

gerardsimons:
I think this would be still be useful. Sometimes I notice my models are not the same. I probably forgot to use a fixed random seed along the way somewhere, but finding at which stage this happens is a little painful as I have to md5sum all intermediate files, and I am never 100% sure if I am doing it the same way as DVC does it.

skshetry:
> I think this would be still be useful. Sometimes I notice my models are not the same between different repro runs. I probably forgot to use a fixed random seed along the way somewhere, but finding at which stage this happens is a little painful as I have to md5sum all intermediate files, and I am never 100% sure if I am doing it the same way as DVC does it.

@gerardsimons, this is now possible with `dvc-data` cli that should be available if you installed `dvc` via pip. If not, you can install via:
```console
pip install dvc-data[cli] -U
```
There are two ways:
1. `dvc-data hash -p <file>` can only give you a md5sum for a file.
[...]

PR #7259, opened 2022-01-12
cli: support external commands available in PATH

skshetry:
This commit adds support for external commands that are available
in one of the directory set in PATH.
The executable should be in format of \'dvc-*\' format, should
have executable permissions and need to exist (duh). The suffix,
after 'dvc-' will be used as a subcommand name, and all tha arguments
after that subcommand will be passed to that external command.

Closes https://github.com/iterative/dvc/issues/6489.

### Motivation

1. This makes DVC extensible with new subcommands without having to modify DVC itself. This can help users to use
custom commands outside of DVC.

2. We may be able to maintain `dvclive` outside `dvc`. Although it's not clear if this will be enough, or if we need python-based plugins.

2. Users may be able to combine multiple dvc commands for their workflow or use it to set an alias. https://github.com/iterative/dvc/issues/7040.

3. We also lack plumbing/low-level commands that I always wanted to have, especially during debugging/development.
These commands could be modifying `dvc-objects` or manipulating `dvc.yaml`/`dvc.lock` files, or debugging pipelines workflow, etc. We have too many commands already, and we may not want to have more commands, that are niche.
This gives us the ability to share scripts among the team. See https://github.com/iterative/dvc/issues/1500.

This is not expected to be documented yet, we may try using it within the team and gather feedback first.

Other projects that support this mechanism:
- Git
- [Cargo](https://doc.rust-lang.org/cargo/reference/external-tools.html#custom-subcommands)

### DIFF
--- a/tests/unit/command/test_external.py
+++ b/tests/unit/command/test_external.py
@@ -6,7 +6,7 @@ from dataclasses import dataclass
 
 import pytest
 
-from dvc.cli import DvcParserError, parse_args, main
+from dvc.cli import DvcParserError, main, parse_args
 from dvc.commands.external import CmdExternal
 from dvc.utils.fs import path_isin
 from dvc.utils.strictyaml import make_relpath
--- a/tests/unit/command/test_external.py
+++ b/tests/unit/command/test_external.py
@@ -1,0 +1,94 @@
+import os
+import shutil
+import stat
+import sys
+from dataclasses import dataclass
+
+import pytest
+
+from dvc.cli import DvcParserError, parse_args, main
+from dvc.commands.external import CmdExternal
+from dvc.utils.fs import path_isin
+from dvc.utils.strictyaml import make_relpath
+
+
+def append_paths(*paths, env="PATH"):
+    if env in os.environ:
+        paths = [*paths, os.environ[env]]
+    return os.pathsep.join(paths)
+
+
+@dataclass
+class ScriptInfo:
+    executable: str
+    script: str
+
+
+def make_executable_path(path):
+    # On Windows, if the executable path is in cwd, shutil.which returns
+    # ".\\rel" paths, otherwise, it returns an absolute path.
+    # On other platforms, it's always an absolute path.
+    abspath = os.path.abspath(path)
+    if os.name == "nt" and path_isin(abspath, os.getcwd()):
+        return make_relpath(path)
+    return abspath
+
+
+@pytest.fixture
+def make_external_cmd(tmp_dir, monkeypatch):
+    def make(cmd: str, contents: str) -> ScriptInfo:
+        monkeypatch.setenv("PATH", append_paths(str(tmp_dir)))
+
+        contents = f"#!{sys.executable}\n{contents}\n"
+        if os.name == "nt":
+            # Here, we create a batch file that invokes python script.
+            # `.BAT` files are usually in PATHEXT, but adding them explicitly
+            # just in case.
+            # We can just run this batch file without it's extension suffix
+            # later, we can do so after adding particular extension to PATHEXT.
+            monkeypatch.setenv("PATHEXT", append_paths(".BAT", "PATHEXT"))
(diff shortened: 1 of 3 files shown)


tests/unit/command/test_external.py
```diff
@@ -0,0 +1,20 @@
...
+    ext.write_text(
+        f"#!{sys.executable}\nimport sys; print(sys.argv); sys.exit(101)"
+    )
+    ext.chmod(ext.stat().st_mode | stat.S_IEXEC)
+
+    path_value = os.pathsep.join([str(tmp_dir), os.getenv("PATH")])
+    monkeypatch.setenv("PATH", path_value)
+    assert shutil.which("dvc-ext") == str(ext)
+
+    assert main(["ext", "target", "--out", "out"]) == 101
```
casperdcl, line 20: shouldn't the output (`print(sys.argv)`) be tested explicitly too?

skshetry: Done. Thanks.

pmrowla:
> 2. We may be able to maintain `dvclive` outside `dvc`. Although it's not clear if this will be enough, or if we need python-based plugins.

Would we be better off using python entrypoints here (rather than using PATH)?

If we defined a `dvc.cli_plugin` (or similar) entry point, users could register their own plugin with attributes for `command_cls` and `add_parser`, where the user command subclasses the existing `dvc.command.base` types, and `add_parser` just conforms to the existing `dvc.command` usage.

This way everyone could expect things like `CmdBase.repo`/`CmdBase.config` to always be loaded and passed into the command object the same way (whether it's an official dvc core command or a user defined one). And we would be able to enforce that users conform to the existing parser/subparser structure in DVC, so setting up help text (and even including user commands in completion installation) would be consistent.

(The disadvantage here would be that using PATH means users aren't forced to use python, but I'm assuming that anyone writing custom DVC subcommands is probably using python and `dvc.repo.Repo` already)

skshetry:
> Would we be better off using python entrypoints here (rather than using PATH)?

I think we need both approaches, and have different use cases. Let's start with their respective pros and cons:

#### PATH based plugins:

##### Pros:
1. Low friction to get started.
2. Works with DVC binaries as well as packages.
3. Can be done on an ad-hoc basis, not too formal.

##### Cons:
1. Does not work out of the box (eg: messing up with PATH may not be easy).
2. Cannot *reliably* use DVC's Python API: DVC's API is unstable, and DVC binary invoking the external command may not be the same version as the installed DVC library.

So, this approach forces users to use the CLI interface more. This is more useful for using simple scripts the user has.

#### Plugin-based approach

### Pros:
1. Extensible, can reliably access DVC's Python API.
2. Works out of the box without any setups other than installation.
3. Works natively, like other DVC's command.

#### Cons:
1. Requires to create a library, and to be available on packaging index (eg: `pypi`)
2. Does not work with binary packages.

So, at least to me, it looks like for dvclive and similar packages that we maintain, providing an API is the best solution.
(`dvclive` could provide a `dvc-live` entrypoints as well). But for users, PATH solution may be more useful.

gvwilson:
See also #7231 (request for the ability to add notes to `.dvc` files) and #7232 (implementation of notes). A plugin mechanism based on Python entry points would have been ideal for this.

dvc/cli.py
```diff
@@ -107,13 +108,61 @@ def _find_parser(parser, cmd_cls):
...
+        # i.e `-v` and similar global flags won't be passed down.
+        # If they do support that, it needs to be specified after `cmd`
+        # explicitly, i.e. `dvc -v ext hello -v`.
+        cmd_pos = arg_strings.index(cmd)
+        namespace.args = arg_strings[cmd_pos + 1 :]
+        # emulating how we call `CmdBase` in `main`
+        namespace.func = external.CmdExternal
+        return namespace, []
+
+    def _parse_known_args(self, arg_strings, namespace):
```
dtrifiro, line 146: Do you think it's safe to rely on these argparse internals?

PR #7264, opened 2022-01-13
exp show: handle extending/renaming of metrics/params.

### LINKED ISSUE #7070: `dvc exp show`'s table: shifts params values to metrics columns after renaming metrics file
# Bug Report

`dvc exp show -A` or `dvc show -n N` shows values of params in wrong columns.

## Description

**Phase 1.** I had a simple training stage in `pipelines/example/dvc.yaml` without metrics tracking, but with params tracking.

```yaml
dummy_train:
    wdir: ../..
    cmd: python scripts/example/dummy_train.py
    deps:
    - scripts/example/dummy_train.py
    params:
    - pipelines/example/params.yaml:
      - test.value
    - pipelines/example/custom_params.yaml:
      - custom.value
```
**Phase 2.** I ran some experiments and then introduced metrics file `pipelines/example/scores.json`.

```yaml
dummy_train:
    wdir: ../..
    cmd: python scripts/example/dummy_train.py
    deps:
    - scripts/example/dummy_train.py
    params:
    - pipelines/example/params.yaml:
      - test.value
    - pipelines/example/custom_params.yaml:
      - custom.value
    metrics:
    - pipelines/example/scores.json:
        cache: false
```

**Phase 3.** I used this stage to run some more experiments and at some point I removed `pipelines/example/scores.json` and added `pipelines/example/metrics.json` instead and the ran further experiments.

**Results.** Now, when I use `dvc exp show -A` I see values from params' columns in the columns for metrics `pipelines/example/scores.json` beginning from the timepoint when I introduced metrics in experiments (phase 2). Params `test.value` and `custom.value` are shown in columns  ` pipelines/example/scores.json:mean_acc` and ` pipelines/example/scores.json:mean_loss`.

Screenshot from 2021-12-01 12-01-24

This happens only if the list of experiments is "long enough". As soon as I use, e.g `dvc exp show -n 8` everything is fine. This matches only the experiments from phase 3, where the obsolete `pipelines/example/scores.json` was already removed from the stage and Git repository.

Screenshot from 2021-12-01 12-07-09

### Reproduce

1. Create  a stage in `pipelines/example/dvc.yaml` as described in phase 2.
2. `dvc repro pipelines/example/dvc.yaml`
3. `git add pipelines/example/` and `git commit`
4. Amend `dvc.yaml` as said in phase 3 by renaming `scores.json` to `metrics.json`
5. `dvc repro pipelines/example/dvc.yaml`
6. `git rm pipelines/example/scores.json`
7. `git add pipelines/example/`
8. `dvc exp  show  -A`

### Expected

Expected to see values of params in columns for params as show in the 2nd image above.

### Environment information

**Output of `dvc doctor`:**

```console
DVC version: 2.8.3 (pip)
Platform: Python 3.8.10 on Linux-5.8.0-63-generic-x86_64-with-glibc2.32
Supports:
	hdfs (fsspec = 2021.11.0, pyarrow = 5.0.0),
	webhdfs (fsspec = 2021.11.0),
	http (aiohttp = 3.7.4.post0, aiohttp-retry = 2.4.6),
	https (aiohttp = 3.7.4.post0, aiohttp-retry = 2.4.6),
	ssh (sshfs = 2021.11.1)
Caches: local
Remotes: ssh, ssh
Repo: dvc, git
```

daavoo:
_extend_row did not consider partially empty `items`, which can occur when metris or params are renamed between commits.

Fixes #7070

### DIFF
--- a/dvc/commands/experiments/show.py
+++ b/dvc/commands/experiments/show.py
@@ -65,6 +65,9 @@ experiment_types = {
 def _collect_rows(
     base_rev,
     experiments,
+    all_headers,
+    metric_headers,
+    param_headers,
     metric_names,
     param_names,
     deps_names,
@@ -87,14 +90,18 @@ def _collect_rows(
 
     new_checkpoint = True
     for i, (rev, results) in enumerate(experiments.items()):
+        fill_value = FILL_VALUE_ERRORED if results.get("error") else fill_value
+        row_dict = {k: fill_value for k in all_headers}
+
         exp = results.get("data", {})
+
         if exp.get("running"):
             state = "Running"
         elif exp.get("queued"):
             state = "Queued"
         else:
             state = fill_value
-        executor = exp.get("executor", fill_value)
+
         is_baseline = rev == "baseline"
 
         if is_baseline:
@@ -102,9 +109,7 @@ def _collect_rows(
         else:
             name_rev = rev[:7]
 
-        exp_name = exp.get("name", "")
         tip = exp.get("checkpoint_tip")
-
         parent_rev = exp.get("checkpoint_parent", "")
         parent_exp = experiments.get(parent_rev, {}).get("data", {})
         parent_tip = parent_exp.get("checkpoint_tip")
@@ -130,26 +135,28 @@ def _collect_rows(
         if not is_baseline:
             new_checkpoint = not (tip and tip == parent_tip)
 
-        row = [
-            exp_name,
-            name_rev,
-            typ,
-            _format_time(exp.get("timestamp"), fill_value, iso),
-            parent,
-            state,
-            executor,
-        ]
-        fill_value = FILL_VALUE_ERRORED if results.get("error") else fill_value
+        row_dict["Experiment"] = exp.get("name", "")
+        row_dict["rev"] = name_rev
+        row_dict["typ"] = typ
+        row_dict["Created"] = _format_time(
+            exp.get("timestamp"), fill_value, iso
+        )
+        row_dict["parent"] = parent
+        row_dict["State"] = state
+        row_dict["Executor"] = exp.get("executor", fill_value)
+
         _extend_row(
-            row,
+            row_dict,
             metric_names,
+            metric_headers,
             exp.get("metrics", {}).items(),
             precision,
             fill_value=fill_value,
         )
         _extend_row(
-            row,
+            row_dict,
             param_names,
+            param_headers,
             exp.get("params", {}).items(),
             precision,
             fill_value=fill_value,
@@ -158,8 +165,8 @@ def _collect_rows(
             hash_info = exp.get("deps", {}).get(dep, {}).get("hash")
             if hash_info is not None:
                 hash_info = hash_info[:7]
-            row.append(hash_info or fill_value)
-        yield row
+            row_dict[dep] = hash_info
+        yield list(row_dict.values())
 
 
 def _sort_column(sort_by, metric_names, param_names):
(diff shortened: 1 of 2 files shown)


dvc/commands/experiments/show.py
```diff
@@ -65,6 +65,9 @@ def _collect_names(all_experiments, **kwargs):
 def _collect_rows(
     base_rev,
     experiments,
+    all_headers,
+    metric_headers,
+    param_headers,
```
skshetry, line 70: Would it be possible to just use `TabularData.row_from_dict()` here? The parameters are getting complicated here.

dvc/commands/experiments/show.py
```diff
@@ -130,26 +135,28 @@ def _collect_rows(
...
-        fill_value = FILL_VALUE_ERRORED if results.get("error") else fill_value
+        row_dict["Experiment"] = exp.get("name", "")
+        row_dict["rev"] = name_rev
+        row_dict["typ"] = typ
+        row_dict["Created"] = _format_time(
+            exp.get("timestamp"), fill_value, iso
+        )
+        row_dict["parent"] = parent
+        row_dict["State"] = state
+        row_dict["Executor"] = exp.get("executor", fill_value)
```
skshetry, line 146: Now this requires us to keep headers in sync in two places. (we did that before by forcing an order anyway though).

karajan1001:
Need to rebase it.

dvc/commands/experiments/show.py
daavoo: Added change using `row_from_dict` but it only saves us from passing `all_headers`. 

Unfortunately, we still need `metric_headers` and `param_headers` later in this function because of the bug being fixed. 

The overall handling of metrics and params could be simplified but feels out of the scope of this fix.

dvc/commands/experiments/show.py
daavoo: That's true but, as you mentioned, the sync was required before. By using dict keys, at least the sync is explicit (and prevents bug being fixed)

dvc/commands/experiments/show.py
skshetry: Makes sense. Thanks.

dvc/commands/experiments/show.py
skshetry, line 146: Just a nitpicky suggestion, what do you think of using `row_dict.update()` here? I'm okay with how it is as well though.

skshetry (approved):

dvc/commands/experiments/show.py
skshetry: We should simplify these parameters soon though, as it is getting complicated as `exp show` is becoming complex.
May get more confusing in the future. Anyway, thanks for looking into it. 🙂

dvc/commands/experiments/show.py
daavoo: Do you mean like the following?

```suggestion
        row_dict.update(
            Experiment=exp.get("name", ""),
            rev=name_rev,
            typ=typ,
            Created=_format_time(
                exp.get("timestamp"), fill_value, iso
            ),
            parent=parent,
            State=state,
            Executor=exp.get("executor", fill_value)
        )
```

No strong preference, is `update` more efficient or is there another reason?

skshetry: @daavoo, I meant it more like following, to use dict arg than as kwargs:
```python
row_dict.update({
    "Experiment": exp.get("name", ""),
    "rev": name_rev,
     "typ": typ,
     "Created": _format_time(exp.get("timestamp"), fill_value, iso),
     "parent": parent,
     "State": state,
     "Executor": exp.get("executor", fill_value),
     ...
})
```

It's a matter of personal preference, I guess. As I said, I am fine as-is too.

PR #7271, opened 2022-01-14
update exp init help text

dberenbaum:
> Matches https://github.com/iterative/dvc.org/pull/3071

See https://github.com/iterative/dvc.org/pull/3071#discussion_r783512996

### DIFF
--- a/dvc/commands/experiments/init.py
+++ b/dvc/commands/experiments/init.py
@@ -98,7 +98,7 @@ class CmdExperimentsInit(CmdBase):
 
 def add_parser(experiments_subparsers, parent_parser):
 
-    EXPERIMENTS_INIT_HELP = "Initialize experiments."
+    EXPERIMENTS_INIT_HELP = "Quickly setup any project to use experiments."
     experiments_init_parser = experiments_subparsers.add_parser(
         "init",
         parents=[parent_parser],


dvc/command/experiments/init.py
```diff
@@ -97,7 +97,7 @@ def run(self):
 
 def add_parser(experiments_subparsers, parent_parser):
 
-    EXPERIMENTS_INIT_HELP = "Initialize experiments."
+    EXPERIMENTS_INIT_HELP = "Quickly setup any project to use DVC Experiments."
```
skshetry, line 100: I don't think we need to mention DVC here. Also, I don't think we need to capitalize "experiments" (at least we don't do in other places).
```suggestion
    EXPERIMENTS_INIT_HELP = "Quickly setup any project to use experiments."
```

dberenbaum: @jorgeorpinel Do you have an opinion here? Does it matter if we say "experiments" here but "DVC Experiments" in the docs?

I assume the idea is to create a distinction from generic experiments, but since there's no link in the CLI help text, it might make less sense here than in the docs.

shcheklein: My 2cs - I agree with Saugat. To be honest even in docs I still have mixed feelings about "DVC Experiments" as a term- if feels a bit artificial.

jorgeorpinel: > I don't think we need to mention DVC here.
> Does it matter if we say "experiments" here but "DVC Experiments" in the docs?
> I still have mixed feelings about "DVC Experiments" as a term

I see the point, since it's DVC. FYI the reason we use the branded term "DVC Experiments" in the docs is to distinguish from experiments you may be conducting in a DVC project without using `dvc exp`. Being inside the specific context of `dvc exp` I suppose it's indeed unnecessary here 🙂 esp. since no one loves the term anyway (but it's the best we have ATM for docs at least).

jorgeorpinel (approved):

PR #7273, opened 2022-01-17
tests: several tests migrated unittest to pytest

nik123:
Part of #1819

Changes:
1. Removed `TestGit` in `basic_env.py`.
2. Several `unittest.TestCase` subclasses in `func/test_odb.py` migrated to pytest: `TestCacheLoadBadDirCache`, `TestExternalCacheDir`, `TestSharedCacheDir`, `TestCacheLinkType`, `TestCmdCacheDir`
3. Several `unittest.TestCase` subclasses in `func/test_fs.py` migrated to pytest: `TestMtimeAndSize`, `TestContainsLink`,  

I have followed the checklit. 

This PR doesn't require documentation updates.

### DIFF
--- a/tests/func/test_odb.py
+++ b/tests/func/test_odb.py
@@ -9,176 +9,168 @@ from dvc.data.db import ODBManager
 from dvc.hash_info import HashInfo
 from dvc.objects.errors import ObjectFormatError
 from dvc.utils import relpath
-from tests.basic_env import TestDir, TestDvc
 
 
-class TestCache(TestDvc):
-    def setUp(self):
-        super().setUp()
-        self.cache1_md5 = "123"
-        self.cache2_md5 = "234"
-        self.cache1 = os.path.join(
-            self.dvc.odb.local.cache_dir,
-            self.cache1_md5[0:2],
-            self.cache1_md5[2:],
-        )
-        self.cache2 = os.path.join(
-            self.dvc.odb.local.cache_dir,
-            self.cache2_md5[0:2],
-            self.cache2_md5[2:],
-        )
-        self.create(self.cache1, "1")
-        self.create(self.cache2, "2")
+def test_cache(tmp_dir, dvc):
+    cache1_md5 = "123"
+    cache2_md5 = "234"
+    cache1 = os.path.join(
+        dvc.odb.local.cache_dir,
+        cache1_md5[0:2],
+        cache1_md5[2:],
+    )
+    cache2 = os.path.join(
+        dvc.odb.local.cache_dir,
+        cache2_md5[0:2],
+        cache2_md5[2:],
+    )
+    tmp_dir.gen({cache1: "1", cache2: "2"})
 
-    def test_all(self):
-        md5_list = list(ODBManager(self.dvc).local.all())
-        self.assertEqual(len(md5_list), 2)
-        self.assertIn(self.cache1_md5, md5_list)
-        self.assertIn(self.cache2_md5, md5_list)
+    assert os.path.exists(cache1)
+    assert os.path.exists(cache2)
 
-    def test_get(self):
-        cache = ODBManager(self.dvc).local.hash_to_path(self.cache1_md5)
-        self.assertEqual(os.fspath(cache), self.cache1)
+    odb = ODBManager(dvc)
 
+    md5_list = list(odb.local.all())
+    assert len(md5_list) == 2
+    assert cache1_md5 in md5_list
+    assert cache2_md5 in md5_list
 
-class TestCacheLoadBadDirCache(TestDvc):
-    def _do_test(self, ret):
-        self.assertTrue(isinstance(ret, list))
-        self.assertEqual(len(ret), 0)
-
-    def test(self):
-        from dvc.data import load
-
-        dir_hash = "123.dir"
-        fname = os.fspath(self.dvc.odb.local.hash_to_path(dir_hash))
-        self.create(fname, "<clearly>not,json")
-        with pytest.raises(ObjectFormatError):
-            load(self.dvc.odb.local, HashInfo("md5", dir_hash))
-
-        dir_hash = "234.dir"
-        fname = os.fspath(self.dvc.odb.local.hash_to_path(dir_hash))
-        self.create(fname, '{"a": "b"}')
-        with pytest.raises(ObjectFormatError):
-            load(self.dvc.odb.local, HashInfo("md5", dir_hash))
+    odb_cache1 = odb.local.hash_to_path(cache1_md5)
+    odb_cache2 = odb.local.hash_to_path(cache2_md5)
+    assert os.fspath(odb_cache1) == cache1
+    assert os.fspath(odb_cache2) == cache2
 
 
-class TestExternalCacheDir(TestDvc):
-    def test(self):
-        cache_dir = TestDvc.mkdtemp()
+def test_cache_load_bad_dir_cache(tmp_dir, dvc):
+    from dvc.data import load
 
-        ret = main(["config", "cache.dir", cache_dir])
-        self.assertEqual(ret, 0)
+    dir_hash = "123.dir"
+    fname = os.fspath(dvc.odb.local.hash_to_path(dir_hash))
+    tmp_dir.gen({fname: "<clearly>not,json"})
+    with pytest.raises(ObjectFormatError):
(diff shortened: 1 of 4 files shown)


tests/func/test_odb.py
```diff
@@ -41,144 +41,144 @@ def test_get(self):
...
-    def test(self):
-        cache_dir = os.path.abspath(os.path.join(os.curdir, "cache"))
-        for d in ["dir1", "dir2"]:
-            os.mkdir(d)
-            os.chdir(d)
+def test_shared_cache_dir(tmp_dir, scm, dvc):
+    cache_dir = os.path.abspath(os.path.join(os.curdir, "cache"))
+    for d in ["dir1", "dir2"]:
+        os.mkdir(d)
+        os.chdir(d)
```
pmrowla, line 102: This test should really be re-written to use the `tmp_dir` fixture properly. It also does not need the `scm` or `dvc` fixtures, since the test itself is initializing DVC repos with `--no-scm`.

So the test should be something like:
```py
def test_shared_cache_dir(tmp_dir):
    cache_dir = tmp_dir / "cache"
    for d in ["dir1", "dir2"]:
        makedirs(tmp_dir / d)
        with (tmp_dir / d).chdir():
            # do the rest of the test operations
            ...
    assert not (tmp_dir / "dir1" /  ".dvc" / "cache").exists()
    assert not (tmp_dir / "dir2" /  ".dvc" / "cache").exists()
    ...
```

We can also just use `(tmp_dir / d).gen(...)` instead of the existing `with open(...): write(...)` calls as well

tests/func/test_odb.py
```diff
@@ -41,144 +41,144 @@ def test_get(self):
...
+    assert ret == 0
 
-        ret = main(["config", "cache.dir", cache_dir])
-        self.assertEqual(ret, 0)
+    assert not os.path.exists(dvc.odb.local.cache_dir)
 
-        self.assertFalse(os.path.exists(self.dvc.odb.local.cache_dir))
+    tmp_dir.gen({"foo": "foo"})
+    ret = main(["add", "foo"])
+    assert ret == 0
```
pmrowla, line 70:
```suggestion
    tmp_dir.dvc_gen({"foo": "foo"})
```

tests/func/test_odb.py
```diff
@@ -41,144 +41,144 @@ def test_get(self):
...
+    tmp_dir.gen(
+        {
+            "data_dir": {
+                "data": "data_dir/data",
+                "data_sub_dir": {"data_sub": "data_dir/data_sub_dir/data_sub"},
+            }
+        }
+    )
+    ret = main(["add", "data_dir"])
+    assert ret == 0
```
pmrowla, line 81:
```suggestion
    tmp_dir.dvc_gen(
        {
            "data_dir": {
                "data": "data_dir/data",
                "data_sub_dir": {"data_sub": "data_dir/data_sub_dir/data_sub"},
            }
        }
    )
```

tests/func/test_odb.py
```diff
@@ -41,144 +41,144 @@ def test_get(self):
...
+    assert len(os.listdir(os.path.join(cache_dir, subdirs[2]))) == 1
+
+
+def test_cache_link_type(tmp_dir, scm, dvc):
+    ret = main(["config", "cache.type", "reflink,copy"])
+    assert ret == 0
 
+    tmp_dir.gen({"foo": "foo"})
+    ret = main(["add", "foo"])
+    assert ret == 0
```
pmrowla, line 145:
```suggestion
    tmp_dir.dvc_gen({"foo": "foo"})
```

tests/func/test_odb.py
```diff
@@ -41,144 +41,144 @@ def test_get(self):
...
+    # dir path written to config should be just one level above.
+    rel = os.path.join("..", dname)
+    config = configobj.ConfigObj(dvc.config.files["repo"])
+    assert config["cache"]["dir"] == rel.replace("\\", "/")
 
-        ret = main(["add", self.FOO])
-        self.assertEqual(ret, 0)
+    tmp_dir.gen({"foo": "foo"})
+    ret = main(["add", "foo"])
+    assert ret == 0
```
pmrowla, line 176:
```suggestion
    tmp_dir.dvc_gen({"foo": "foo"})
```

pmrowla (changes requested):
Thanks for the PR @nik123 

The changes relating to moving from unittest to pytest look good, but a lot of the tests should also be updated to use the `tmp_dir`/`dvc`/`scm` fixtures correctly.  Basically, in the old tests there are a lot of places where `main()` is used, but we can now just use the pytest fixtures instead.

I noticed that you did replace the `TestDvc.create()` calls with `tmp_dir.gen()` which is a good start, but in a lot of cases there are more lines that can also be replaced.

For example
```py
tmp_dir.gen("foo", "foo")
ret = main(["add", "foo"])
assert 0 == ret
```

can be replaced with just
```
tmp_dir.dvc_gen("foo", "foo")
```

(`dvc_gen` is `tmp_dir.gen()` that automatically `dvc add`s everything as well

I marked a few cases that I noticed during the review, but please double check the tests you've migrated for similar changes that could be made.

tests/func/test_odb.py
```diff
@@ -41,144 +41,144 @@ def test_get(self):
...
-        assert main(["remote", "add", "cache", "remote://storage/tmp"]) == 0
-        assert main(["config", "cache.ssh", "cache"]) == 0
+def test_remote_references(tmp_dir, dvc):
+    ssh_url = "ssh://user@localhost:23"
+    assert main(["remote", "add", "storage", ssh_url]) == 0
+    assert main(["remote", "add", "cache", "remote://storage/tmp"]) == 0
+    assert main(["config", "cache.ssh", "cache"]) == 0
 
-        self.dvc.__init__()
+    dvc.__init__()
```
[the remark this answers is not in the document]
nik123: @pmrowla - I've removed `dvc.__init__` call and `main` calls in this test. But I had to also add `dvc.odb = ODBManager(dvc)` because `dvc.config` updates are not applied to `dvc.odb.config`. If there is a better way to update `odb`, please tell me.

nik123:
> The changes relating to moving from unittest to pytest look good, but a lot of the tests should also be updated to use the `tmp_dir`/`dvc`/`scm` fixtures correctly. Basically, in the old tests there are a lot of places where `main()` is used, but we can now just use the pytest fixtures instead.

@pmrowla I've replaced `main` calls with fixtures with two exceptions:
1. `test_shared_cache_dir`
2. Tests which explicitly declare cmd call in their name: i.e. `test_cmd_cache_dir`, `test_cmd_cache_abs_path` because their names suggest that command should be called.

pmrowla:
@nik123 thanks for making the changes.

I think it looks ok now, could you please squash your commits together? We have started merging rebased/linear history into the main DVC branch (rather than squashing all PRs into single commits).

In this case, it should be fine to squash down to a single `tests: migrate unittest to pytest` commit (instead of the several repeated `Update tests/func/test_odb.py` etc. commits).

nik123:
> I think it looks ok now, could you please squash your commits together? We have started merging rebased/linear history into the main DVC branch (rather than squashing all PRs into single commits).
> 
> In this case, it should be fine to squash down to a single `tests: migrate unittest to pytest` commit (instead of the several repeated `Update tests/func/test_odb.py` etc. commits).

@pmrowla Not sure if I understood you right but I've squashed all commits in `unittest-to-pytest` into single commit, rebased `unittest-to-pytest` on `main` and then force-pushed my branch. 

BTW could you please elaborate why you stopped using squashing?

pmrowla:
> @pmrowla Not sure if I understood you right but I've squashed all commits in `unittest-to-pytest` into single commit, rebased `unittest-to-pytest` on `main` and then force-pushed my branch.

That's fine, thanks

> BTW could you please elaborate why you stopped using squashing?

Because for larger PR's, it makes sense to merge multiple commits from the PR as separate commits instead of one giant squashed commit. This makes things like doing `git bisect` on `main` more useful since if a PR breaks something it is easier to narrow down what specific change from a PR introduced the bug.

Basically, for small/straightforward PR's, like this one, we will generally still want to squash them into a single commit, but it's now on the devs to do it selectively when it makes sense, as opposed to always squashing PR into a single commit by default.

pmrowla (approved):

pmrowla:
Thanks again @nik123 🙏

Issue #7137, opened 2021-12-13
`exp init`: clean up error messages


dberenbaum:
From #7130:

```
$ dvc -V
2.9.2
```

```
$ git init
$ dvc init
$ dvc exp init python train.py
ERROR: unexpected error - [Errno 2] No such file or directory: 'params.yaml'
```

* [x]  Why is it `unexpected` error? It should be like “ERROR: hyperparameters file ‘params.yaml’ is missing”.    ~~🤔  Why `dvc exp init` does not create it if it is required?~~

* [x]  Do we really need [Errno 2]? In what cases the code helps users? This makes the message longer and breaks human readability.

```
$ dvc exp run
ERROR: failed to reproduce 'dvc.yaml': [Errno 2] No such file or directory: '/Users/dmitry/src/test/myproj/data'
```

* [x]  full path. It is supposed to be relative. Why: 1) to make it short - the full path provides zero value in most of use cases; 2) it makes it hard to submit reports (some users are not comfortable to share their directories struture - See `'/Users/dmitry/src/test/` - bc it might contain confidential info like user and project names.

* [ ] ~~why `dvc exp init` checked `params.yaml` but didn’t check (or create) `data/` & `src/`?~~

```
$ mkdir data
$ dvc exp run
ERROR: failed to reproduce 'dvc.yaml': [Errno 2] No such file or directory: '/Users/dmitry/src/test/myproj/src'
```

* [x]  the same questions - [Errno 2], Full path.

* [ ]  ~~why it was not checked in `dvc init`? Or was not created?~~

```
$ mkdir src
$ dvc exp run
…
python: can't open file 'train.py': [Errno 2] No such file or directory
ERROR: failed to reproduce 'dvc.yaml': failed to run: python train.py, exited with 2
```

* [ ]   “exited with 2” Is it useful information? Where the exit code 2 belongs to? It might confuse users. Error code can be in debug errors not here.

* [x]  Another issue - ideally, we should have a reference to stage (a user might have multiple stages with the same or similar commands). The message might look like:    **ERROR: failed to execute stage ‘train’ from ‘dvc.yaml’**

skshetry:
>  “exited with 2” Is it useful information? Where the exit code 2 belongs to? It might confuse users. Error code can be in debug errors not here.

I disagree with this, exit codes are important, well that's how you know it's succeeded or failed (or crashed).

>  Do we really need [Errno 2]? In what cases the code helps users? This makes the message longer and breaks human readability.
[...]

PR #7282, opened 2022-01-18
repro/exp run: display stage name instead of filepath on failure

skshetry:
Before it used to say 'failed to reproduce dvc.yaml', which is now changed
to print 'failed to reproduce train' or something similar.

For stages that are in dvc.yaml in the current working directory, it will
just print the stage names. For stages in other dvc.yaml files, it'll print

in <relative_path>:<stage_name>.

```console
$ dvc exp run
ERROR: failed to reproduce 'train': [Errno 2] No such file or directory: '/private/var/folders/xh/trg29z296h70n109kwfk6g800000gn/T/tmp.fJbphlSG/data'
```
Related: https://github.com/iterative/dvc/issues/7137

### DIFF
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -157,9 +157,9 @@ class InitError(DvcException):
 
 
 class ReproductionError(DvcException):
-    def __init__(self, dvc_file_name):
-        self.path = dvc_file_name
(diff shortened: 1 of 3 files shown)


pmrowla (approved):

dberenbaum (approved):
LGTM!

PR #7283, opened 2022-01-18
find_root: unify error message

### LINKED ISSUE #7281: 'None' does not contain DVC directory
```console
cd "$(mktemp -d)"
dvc add foo  # or any other command
```

efiop:
We used to use that message for special external repo cases, but we don't
need it anymore.

Fixes #7281

Leftover from #7214

### DIFF
--- a/dvc/repo/__init__.py
+++ b/dvc/repo/__init__.py
@@ -314,8 +314,8 @@ class Repo:
     def find_root(cls, root=None, fs=None) -> str:
         from dvc.fs.local import LocalFileSystem, localfs
 
-        root_dir = os.path.realpath(root or os.curdir)
-        _fs = fs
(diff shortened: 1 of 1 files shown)


dvc/repo/__init__.py
```diff
@@ -314,8 +314,8 @@ def __repr__(self):
     def find_root(cls, root=None, fs=None) -> str:
         from dvc.fs.local import LocalFileSystem, localfs
 
-        root_dir = os.path.realpath(root or os.curdir)
-        _fs = fs
+        root = root or os.curdir
+        root_dir = os.path.realpath(root)
```
efiop, line 318: Need to get rid of this to detouch from fs. Some parts are not ready to switch away from abspaths yet, so keeping as is.

PR #7287, opened 2022-01-19
cli: refactor into a module

skshetry:
By splitting dvc.cli into a module, we'll be able to share cli-related
utilities, split parser-related stuffs into smaller parts, improve
reusability and improve API for command extensions in the future.

This PR moves main entrypoint from `dvc.main.main` to `dvc.cli.main`, but the old one is still kept for backward compatibility, but that should be considered deprecated after this PR is merged and set for removal in 3.0.


dvc/main.py
```diff
@@ -1,115 +1,2 @@
...
-        # more info.
-        close_pools()
-
-        from dvc.external_repo import clean_repos
-
-        # Remove cached repos in the end of the call, these are anonymous
-        # so won't be reused by any other subsequent run anyway.
-        clean_repos()
+# TODO: remove in 3.0, kept for backward compatibility
+from .cli import main  # noqa: F401, pylint: disable=unused-import
```
skshetry, line 2: Keeping for backward compatibility

efiop: @skshetry Could you elaborate, please? We never promised this as an API, so looks like we can get rid of it right away.

efiop:
@skshetry Should `dvc.command` be part of `dvc.cli`?

dvc/main.py
skshetry: yeah, but I have been suggesting it to the users in Discord as a way to use `dvc` as a Python API in a stable manner. Could remove though.

skshetry:
> @skshetry Should `dvc.command` be part of `dvc.cli`?

I don't think so. `dvc.cli` is about entry point and cli-parsing utilities. #7259 and #7272 has evolved my understanding about `dvc.command` as plugins.

I have a few minor refactors after this:
1. Move `dvc.command.base` to `dvc.cli.command`.
2. Rename `dvc.command` to `dvc.commands`.
3. Change `dvc.exceptions.DvcParser` to not inherit from `DvcException` as it's strictly `dvc.cli` only thing. It'll be moved as `dvc.cli.DvcParserError`.

efiop (approved):

PR #7297, opened 2022-01-20
add: Update help of `--jobs`

daavoo:

### DIFF
--- a/dvc/commands/add.py
+++ b/dvc/commands/add.py
@@ -102,7 +102,9 @@ def add_parser(subparsers, parent_parser):
         "--jobs",
         type=int,
         help=(
-            "Number of jobs to run simultaneously. "
+            "Only used along with '--to-remote'. "
+            "Number of jobs to run simultaneously "
+            "when pushing data to remote."
             "The default value is 4 * cpu_count(). "
             "For SSH remotes, the default is 4. "
         ),


dvc/commands/add.py
```diff
@@ -102,7 +102,9 @@ def add_parser(subparsers, parent_parser):
         "--jobs",
         type=int,
         help=(
-            "Number of jobs to run simultaneously. "
+            "Only used along with '--to-remote'. "
+            "Number of jobs to run simultaneously "
+            "when pushing data to remote."
             "The default value is 4 * cpu_count(). "
```
shcheklein, line 108: just to double check: is it a correct default value? (I thought it was something like `min(4, 4 * cpu_count())`? (which we should also update I think)

daavoo: It is the current default: https://github.com/iterative/dvc/blob/73fee513b955f1b44542a450318ad42eec850add/dvc/fs/base.py#L36

Do you mean that we should increase the number?

shcheklein: no, I wanted to double check if it's correct or not - that's it

I always forget which one of those two lines (https://github.com/iterative/dvc/blob/73fee513b955f1b44542a450318ad42eec850add/dvc/fs/base.py#L38) apply to what part.

I would may be check if some file systems override it. Just in case.

pared: HASH_JOBS is for calculating checksums and can be overridden with `core.checksum_jobs`. 
`add` jobs refer to `to_remote` functionality.

pared (approved):

dvc/commands/add.py
daavoo: > I would may be check if some file systems override it. Just in case.

Currently, both `_JOBS` and `HASH_JOBS` are never overwritten by any file system.

PR #7300, opened 2022-01-21
utils: add decorator to wrap diskcache pickle errors

pmrowla:

Docs PR: https://github.com/iterative/dvc.org/pull/3205

Generates a better error message w/troubleshooting link for pickle errors in state/odb-index.
Related to https://github.com/iterative/dvc/pull/7222 (see discussion in https://github.com/iterative/dvc/pull/7222#discussion_r789406572)

### DIFF
--- a/dvc/utils/decorators.py
+++ b/dvc/utils/decorators.py
@@ -1,0 +1,30 @@
+import pickle
+from typing import Callable, TypeVar
+
+from funcy import decorator
+
+from dvc.exceptions import DvcException
+
+from . import format_link
+
+_R = TypeVar("_R")
+
+
+@decorator
+def with_diskcache(call: Callable[..., _R], name: str) -> _R:
+    try:
+        return call()
+    except (pickle.PickleError, ValueError) as exc:
+        if isinstance(exc, ValueError) and not str(exc).startswith(
+            "pickle protocol"
+        ):
+            raise
+        link = format_link(
+            "https://dvc.org/doc/user-guide/troubleshooting#pickle"
+        )
+        msg = (
+            f"Could not open pickled '{name}' cache. Remove the "
+            f"'.dvc/tmp/{name}' directory and then retry this command. "
+            f"See {link} for more information."
+        )
+        raise DvcException(msg) from exc
--- a/dvc/data/db/index.py
+++ b/dvc/data/db/index.py
@@ -4,6 +4,7 @@ from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Iterable, Set
 
 from dvc.objects.errors import ObjectDBError
+from dvc.utils.decorators import with_diskcache
 
 if TYPE_CHECKING:
     from dvc.types import StrPath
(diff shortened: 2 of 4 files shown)


skshetry (approved):

dvc/utils/decorators.py
```diff
@@ -0,0 +1,30 @@
...
+            "pickle protocol"
+        ):
+            raise
+        link = format_link(
+            "https://dvc.org/doc/user-guide/troubleshooting#pickle"
+        )
+        msg = (
+            f"Could not open pickled '{name}' cache. Remove the "
+            f"'.dvc/tmp/{name}' directory and then retry this command. "
+            f"See {link} for more information."
```
jorgeorpinel, line 28: What are "pickled index/md5s/links/tests caches" ? Seems like a totally alien concept.

jorgeorpinel: p.s. I realize it refers to some of the dirs listed in https://dvc.org/doc/user-guide/project-structure/internal-files but I doubt many users will be familiar at all. Calling these locations caches on top seems extra confusing 🙃 Let me think...

jorgeorpinel: 1. So firstly, should we document somewhere that something is pickled in the .dvc/tmp/ dirs? I incline to avoid this as it seems like an unnecessary implementation detail.
2. Assuming not, let's generalize here with something like "`.dvc/tmp/{name}` seems to be corrupted. Please remove the directory and try again. See ... for more info."

> Not 100% sure on the term "corrupted" though. Rel. https://github.com/iterative/dvc.org/pull/3062

pmrowla: It's an implementation detail, but it uses a combination of sqlite db + pickled data on disk

jorgeorpinel: OK thanks for the info @pmrowla but TBH I'm still somewhat confused...

Is it possible to install a same release of DVC with different underlying Python versions?

Even after reading https://github.com/iterative/dvc/pull/7300#pullrequestreview-864458242 I'm struggling to understand how users would end up with this problem. Is there one specific pathway e.g. you installed DVC 2.x having Python 3.7 and then update only Python to 3.9 and your repo starts to fail? Or is it staying on py37 and updating DVC?

If there's a few specific ways to get here it would be helpful to list them in order to explain this (here and in https://github.com/iterative/dvc.org/pull/3205).

pmrowla: > Is it possible to install a same release of DVC with different underlying Python versions?

Yes, if you are using virtualenvs you can have more than one virtualenv with DVC installed. Or you could have both a virtualenv DVC and a binary package DVC installed on the same machine with different Python versions. Or you could have a setup where you have a repo directory that is shared across different platform installations with different Python versions - like if I have a repo dir that is exposed to VMs or containers, but the Python/DVC install inside the VM is different than my host system's Python/DVC.

Basically there is a wide range of possible ways that this could happen, especially in the case where a user has to make sure that their code works across multiple Python versions or across different platforms.

> Even after reading #7300 (review) I'm struggling to understand how users would end up with this problem. 

The issue here is that someone runs DVC in Python > 3.7, and then they run DVC in Python <= 3.7 using the same repo directory. This is not a common problem, but it happens.

Just as an example a user could have DVC installed on macos, where the default system python 3 is 3.8. They may also need to test their training code to make sure it works in linux. If the user installs a linux VM where the system python is 3.7, they would encounter this issue inside the VM, if they re-use the same repo directory (by sharing it from the host to the VM).

PR #7311, opened 2022-01-26
ignore: move debug statetement to the proper location

dtrifiro:
The "adding %s to gitignore" debug statement was always being run because of the `_ignore()` call in `dvc/repo/__init__.py:207`

thanks @pared

### DIFF
--- a/dvc/repo/scm_context.py
+++ b/dvc/repo/scm_context.py
@@ -75,13 +75,13 @@ class SCMContext:
 
         from dvc.scm import SCMError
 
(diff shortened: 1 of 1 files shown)


pared (approved):

dtrifiro:
I'm not sure why `_ignore()` is being called on every instantiation of `Repo`

https://github.com/iterative/dvc/blob/d882e21b9b013f69abff960a8bad62d822024b59/dvc/repo/__init__.py#L207

@pared mentioned that it could be because of a migration from old dvc versions

PR #7326, opened 2022-01-31
tests: fix test_exceptions

### LINKED ISSUE #7322: tests: `test_strict_yaml::test_exceptions` failing on CI
See https://github.com/iterative/dvc/runs/5000170109?check_suite_focus=true#step:7:11107

The failures are consistent, and is only happening on CI. I am not able to reproduce locally.
```
FAILED tests/func/utils/test_strict_yaml.py::test_exceptions[duplicate_keys]
FAILED tests/func/utils/test_strict_yaml.py::test_exceptions[no_hyphen_block]
FAILED tests/func/utils/test_strict_yaml.py::test_exceptions[unclosed_scalar]
FAILED tests/func/utils/test_strict_yaml.py::test_exceptions[empty_stage] - a...
FAILED tests/func/utils/test_strict_yaml.py::test_exceptions[missing_cmd] - a...
FAILED tests/func/utils/test_strict_yaml.py::test_exceptions[deps_as_dict] - ...
FAILED tests/func/utils/test_strict_yaml.py::test_exceptions[outs_as_str] - a...
FAILED tests/func/utils/test_strict_yaml.py::test_exceptions[null_value_on_outs]
FAILED tests/func/utils/test_strict_yaml.py::test_exceptions[additional_key_on_outs]
FAILED tests/func/utils/test_strict_yaml.py::test_exceptions[foreach_scalar]
FAILED tests/func/utils/test_strict_yaml.py::test_exceptions[foreach_do_do_null]
FAILED tests/func/utils/test_strict_yaml.py::test_exceptions[foreach_do_missing_cmd]
FAILED tests/func/utils/test_strict_yaml.py::test_exceptions[foreach_unknown_cmd_missing_do]
FAILED tests/func/utils/test_strict_yaml.py::test_exceptions[merge_conflicts]
```

dtrifiro:
Due to changes in `rich` `11.1.0`, all text formatted using `rich.syntax.Syntax` is now left-justified (see [0a37fcc](https://github.com/Textualize/rich/commit/0a37fcc93a025fb19821fa9a05a59aa1d3848a20)),
resulting in output padded with whitespace to the size of the tty.

This resulted in broken tests due to mismatching whitespace between actual `dvc.ui` output and test cases.

Stripping whitespace in the ui output in `test_exceptions` solves the broken tests for both `rich` `11.1.0` and `11.0.0`

fixes #7322

### DIFF
--- a/tests/func/utils/test_strict_yaml.py
+++ b/tests/func/utils/test_strict_yaml.py
@@ -20,7 +20,7 @@ While constructing a mapping, in line 3, column 5
   3 │   cmd: python train.py
 
 Found duplicate key "cmd" with value "python train.py" (original value:\
- "python \ntrain.py"), in line 4, column 5
+ "python\ntrain.py"), in line 4, column 5
   4 │   cmd: python train.py"""
 
 
@@ -161,8 +161,8 @@ stages:
 NULL_VALUE_ON_OUTS_OUTPUT = """\
 './dvc.yaml' validation failed.
 
-expected str, in stages -> stage1 -> outs -> 0 -> logs -> remote, line 6,\
- column\n9
+expected str, in stages -> stage1 -> outs -> 0 -> logs -> remote, line 6, \
+column\n9
   5 │   - logs:
   6 │   │   cache: false
   7 │   │   persist: true"""
@@ -180,7 +180,7 @@ stages:
 ADDITIONAL_KEY_ON_OUTS_OUTPUT = """\
 './dvc.yaml' validation failed.
 
-extra keys not allowed, in stages -> stage1 -> outs -> 0 -> logs -> \n\
+extra keys not allowed, in stages -> stage1 -> outs -> 0 -> logs ->
 not_existing_key, line 6, column 9
   5 │   - logs:
   6 │   │   cache: false
@@ -351,7 +351,13 @@ def test_exceptions(tmp_dir, dvc, capsys, text, expected, mocker):
     out, err = capsys.readouterr()
 
     assert not out
-    assert expected in err
+
+    # strip whitespace on the right: output is always left-justified
+    # by rich.syntax.Syntax:
+    for expected_line, err_line in zip(
+        expected.splitlines(), err.splitlines()
+    ):
+        assert expected_line == err_line.rstrip(" ")
 
 
 def test_make_relpath(tmp_dir, monkeypatch):


tests/func/utils/test_strict_yaml.py
```diff
@@ -351,7 +351,11 @@ def print_with_fixed_width(*args, **kwargs):
     out, err = capsys.readouterr()
 
     assert not out
-    assert expected in err
+
+    # strip whitespace on the right: output is always left-justified
+    # by rich.syntax.Syntax:
+    for l_expected, l_err in zip(expected.split("\n"), err.split("\n")):
+        assert l_expected.rstrip(" ") == l_err.rstrip(" ")
```
skshetry, line 358: Let's use `splitlines`, and not use `rstrip()` in the expected value.
```suggestion
    for l_expected, l_err in zip(expected.splitlines(), err.splitlines()):
        assert l_expected == l_err.rstrip(" ")
```

skshetry: Few nitpicky suggestions:

1. I prefer `_line` suffix than `l_` prefix.
2. Keep `expected` value on the right hand side.
```suggestion
    for err_line, expected_line in zip(err.splitlines(), expected.splitlines()):
        assert err_line.rstrip() == expected_line
```

But no strong opinion, just being nitpicky here.

tests/func/utils/test_strict_yaml.py
```diff
@@ -180,7 +180,7 @@
 ADDITIONAL_KEY_ON_OUTS_OUTPUT = """\
 './dvc.yaml' validation failed.
 
-extra keys not allowed, in stages -> stage1 -> outs -> 0 -> logs -> \n\
+extra keys not allowed, in stages -> stage1 -> outs -> 0 -> logs ->\n\
```
skshetry, line 183:
```suggestion
extra keys not allowed, in stages -> stage1 -> outs -> 0 -> logs ->
```

tests/func/utils/test_strict_yaml.py
```diff
@@ -20,7 +20,7 @@
   3 │   cmd: python train.py
 
 Found duplicate key "cmd" with value "python train.py" (original value:\
```
skshetry, line 22: Does this fit on a single line now?
```suggestion
Found duplicate key "cmd" with value "python train.py" (original value: "python
```

tests/func/utils/test_strict_yaml.py
```diff
@@ -20,7 +20,7 @@
...
- "python \ntrain.py"), in line 4, column 5
+ "python\ntrain.py"), in line 4, column 5
```
skshetry, line 23: Similarly, if `python` moves above, this could be written as:
```
```suggestion
train.py"), in line 4, column 5
```

tests/func/utils/test_strict_yaml.py
skshetry: Probably linters will complain though, black does not for some reason, hmm.

tests/func/utils/test_strict_yaml.py
dtrifiro: Note that using `rstrip()` instead of `rstrip(" ")` will also strip newlines

skshetry: > Note that using rstrip() instead of rstrip(" ") will also strip newlines

@dtrifiro, sorry, you are right. I was confusing it with `bytes.rstrip()`.

skshetry (approved):

skshetry:
Thank you so much @dtrifiro for the quickfix. 🙂

Issue #4112, opened 2020-06-25
run: allow tracking of all params in a file with `-p`


gcoter:
### Please provide information about your setup

**Output of `dvc version`:**

```console
DVC version: 1.0.2
Python version: 3.6.8
Platform: Linux-4.15.0-106-generic-x86_64-with-LinuxMint-18.3-sylvia
Binary: False
Package: pip
Supported remotes: http, https, ssh
Cache: reflink - not supported, hardlink - supported, symlink - supported
Repo: dvc, git
```

**Additional Information (if any):**

I tried to use `dvc run` with two `-p` arguments (because I have two parameters files). It gives me this error:
```
ERROR: dependency 'params.yaml' does not exist
```

If I remove one of the two arguments, it works. So it looks like DVC doesn't enjoy having two `-p` arguments. Is it expected? If DVC doesn't support more than one parameters file, the ERROR message should indicate that more explicitly.

skshetry:
@gcoter, does the params.yaml exist? Can you give more verbose error message (and, your command that you tried)?

gcoter:
@skshetry thanks for your quick answer! My command looks like this:
```
dvc run -d training.py -p config/params_1.json -p config/params_2.json:data.training,model -o trained_model/ -n training python training.py
```

I prefer to use JSON files so `params.yaml` does not exist in my project.

skshetry:
@gcoter, is that a typo on `-p config/params_1.json` (i.e. there's no list of parameters eg: `-p config/params_1.json:loss`).

gcoter:
@skshetry No, I thought that specifying `config/params_1.json` would select everything inside this file. Should I specify explicitly the keys I want to select?

skshetry:
@gcoter, yes, you have to specify keys that you want.

gcoter:
@skshetry Ok, indeed it works if I specify manually the keys. I thought that DVC would list all the keys by default. But I don't mind listing them myself. Thank you for your help!

skshetry:
@gcoter, but, I can see it's usefulness. I'll log this as a feature request though.

skshetry:
@gcoter, another question, why do you choose to `-p` rather than tracking whole file? for granularity tracking of parameter values, right?

gcoter:
[...]

PR #7331, opened 2022-02-01
exp init: create workspace structure by default, show workspace tree

### LINKED ISSUE #7137: `exp init`: clean up error messages

### LINKED ISSUE #7138: `exp init`: handling of nonexistent files
From #7130:

```
$ dvc -V
2.9.2
```

```
$ git init
$ dvc init
$ dvc exp init python train.py
ERROR: unexpected error - [Errno 2] No such file or directory: 'params.yaml'
```

* [ ]  ~~Why is it `unexpected` error? It should be like “ERROR: hyperparameters file ‘params.yaml’ is missing~~”.    🤔  Why `dvc exp init` does not create it if it is required?

* [ ]  ~~Do we really need [Errno 2]? In what cases the code helps users? This makes the message longer and breaks human readability.~~

```
$ dvc exp run
ERROR: failed to reproduce 'dvc.yaml': [Errno 2] No such file or directory: '/Users/dmitry/src/test/myproj/data'
```

* [ ]  ~~full path. It is supposed to be relative. Why: 1) to make it short - the full path provides zero value in most of use cases; 2) it makes it hard to submit reports (some users are not comfortable to share their directories struture - See `'/Users/dmitry/src/test/` - bc it might contain confidential info like user and project names.~~

* [ ]  why `dvc exp init` checked `params.yaml` but didn’t check (or create) `data/` & `src/`?

```
$ mkdir data
$ dvc exp run
ERROR: failed to reproduce 'dvc.yaml': [Errno 2] No such file or directory: '/Users/dmitry/src/test/myproj/src'
```

* [ ]  ~~the same questions - [Errno 2], Full path.~~

* [ ]  why it was not checked in `dvc init`? Or was not created?

```
$ mkdir src
$ dvc exp run
…
python: can't open file 'train.py': [Errno 2] No such file or directory
ERROR: failed to reproduce 'dvc.yaml': failed to run: python train.py, exited with 2
```

* [ ]   ~~“exited with 2” Is it useful information? Where the exit code 2 belongs to? It might confuse users. Error code can be in debug errors not here.~~

* [ ]  ~~Another issue - ideally, we should have a reference to stage (a user might have multiple stages with the same or similar commands). The message might look like:    **ERROR: failed to execute stage ‘train’ from ‘dvc.yaml’**~~

From a clean git repo:

```
$ dvc init
$ dvc exp init -i
…
'params.yaml' does not exist. Please retry with an existing parameters file.
```

* [ ]  If I do not have `params.yaml` I stuck here. Why don’t we offer to create an empty params file?

 * [ ]  **A kind of conclusion**: `dvc exp init` is not fully supported without interactive mode.
It looks like today the only supported flow is — interactive. The default `dvc init` has multiple assumptions about the project structure and it is not clear how a user should know about the assumptions. Why `dvc init` won’t create all the directories & files? This was a part of the initial idea - to run a single cmd for a given train.py in a live coding section while `dvc init` wraps it to an experiment.

This is how it can look like:

```
$ dvc init python train.py
Creating experiment project structure:
├── data/
├── metrics.json
├── models/
├── params.yaml
├── plots/
└── src/
Created `dvc.yaml` file with ‘train’ stage. To run, use "dvc exp run".
```

### LINKED ISSUE #7139: `exp init`: more verbose output in non-interactive mode
Related to #7130. `dvc exp init -i` has verbose output to help guide the user. Non-interactive mode should have the same. A `-q` option can be added for experienced users to hide this output.

Current `dvc exp init -i` output:

```
$ dvc exp init -i python train.py
This command will guide you to set up a train stage in dvc.yaml.
See https://s.dvc.org/g/pipeline-files.

Enter the paths for dependencies and outputs of the command.
DVC assumes the following workspace structure:
├── data
├── metrics.json
├── models
├── params.yaml
├── plots
└── src

Path to a code file/directory [src, n to omit]:
'src' does not exist in the workspace. "exp run" may fail.
Path to a data file/directory [data, n to omit]:
'data' does not exist in the workspace. "exp run" may fail.
Path to a model file/directory [models, n to omit]:
Path to a metrics file [metrics.json, n to omit]:
Path to a plots file/directory [plots, n to omit]:
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
train:
  cmd: python train.py
  deps:
  - data
  - src
  params:
  - bar
  - foo
  outs:
  - models
  metrics:
  - metrics.json:
      cache: false
  plots:
  - plots:
      cache: false

Do you want to add the above contents to dvc.yaml? [Y/n]: Y

Created train stage in dvc.yaml. To run, use "dvc exp run".
See https://s.dvc.org/g/exp/run.
```

Current `dvc exp init` output:

```
$ dvc exp init python train.py
Created train stage in dvc.yaml. To run, use "dvc exp run".
See https://s.dvc.org/g/exp/run.
```

Proposed `dvc exp init` output:

```
$ dvc exp init python train.py
This command will set up a train stage in dvc.yaml.
See https://s.dvc.org/g/pipeline-files.

DVC assumes the following workspace structure:
├── data
├── metrics.json
├── models
├── params.yaml
├── plots
└── src

'src' does not exist in the workspace. "exp run" may fail.
'data' does not exist in the workspace. "exp run" may fail.
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
train:
  cmd: python train.py
  deps:
  - data
  - src
  params:
  - bar
  - foo
  outs:
  - models
  metrics:
  - metrics.json:
      cache: false
  plots:
  - plots:
      cache: false

Created train stage in dvc.yaml. To run, use "dvc exp run".
See https://s.dvc.org/g/exp/run.
```

skshetry:
This PR creates the workspace structure by default. That means that the `code` paths and `data` paths are created if they don't exist. If the path has any extension, the empty file will be created, otherwise, a directory is created. 

Initially, this was to be created for the defaults only, but I have added the magic everywhere, so this is applied to passed arguments, defaults or config values too.

Second thing that PR adds is this output so that users are aware of the changes and their workspace structure:
```
Creating experiment project structure:
├── data
├── metrics.json
├── models
├── params.yaml
├── plots
└── src
```

Closes #7137, closes #7138 and  closes #7139.

### DIFF
--- a/dvc/commands/experiments/init.py
+++ b/dvc/commands/experiments/init.py
@@ -69,7 +69,6 @@ class CmdExperimentsInit(CmdBase):
         )
 
         text = ui.rich_text.assemble(
-            "\n" if self.args.interactive else "",
             "Created ",
             (self.args.name, "bright_blue"),
             " stage in ",
--- a/dvc/repo/experiments/init.py
+++ b/dvc/repo/experiments/init.py
@@ -8,6 +8,7 @@ from typing import (
     Callable,
     Dict,
     Iterable,
+    List,
     Optional,
     TextIO,
     Tuple,
@@ -29,6 +30,7 @@ if TYPE_CHECKING:
     from dvc.repo import Repo
     from dvc.dvcfile import DVCFile
     from rich.tree import Tree
+    from dvc.dependency import Dependency
 
 from dvc.ui import ui
 
@@ -79,16 +81,24 @@ def _disable_logging(highest_level=logging.CRITICAL):
         logging.disable(previous_level)
 
 
-def build_workspace_tree(workspace: Dict[str, str]) -> "Tree":
+def build_workspace_tree(workspace: Dict[str, str], label: str) -> "Tree":
     from rich.tree import Tree
 
-    tree = Tree(
-        "DVC assumes the following workspace structure:",
-        highlight=True,
-    )
+    tree = Tree(label, highlight=True)
     for value in sorted(workspace.values()):
         tree.add(f"[green]{value}[/green]")
     return tree
+
+
+def display_workspace_tree(
+    workspace: Dict[str, str], label: str, stderr: bool = False
+) -> None:
+    d = workspace.copy()
+    d.pop("cmd", None)
+
+    if d:
+        ui.write(build_workspace_tree(d, label), styled=True, stderr=stderr)
+    ui.write(styled=True, stderr=stderr)
 
 
 PIPELINE_FILE_LINK = "https://s.dvc.org/g/pipeline-files"
@@ -144,9 +154,9 @@ def init_interactive(
         "Enter the paths for dependencies and outputs of the command.",
         styled=True,
     )
-    if workspace:
-        ui.error_write(build_workspace_tree(workspace), styled=True)
-    ui.error_write(styled=True)
+
+    tree_label = "DVC assumes the following workspace structure:"
+    display_workspace_tree(workspace, tree_label, stderr=True)
     ret.update(
         compact(
             _prompts(prompts, defaults, validator=validator, stream=stream)
@@ -192,11 +202,40 @@ def validate_prompts(
             raise InvalidResponse(msg_format.format(value, suffices[e.errno]))
     elif key in ("code", "data"):
         if not os.path.exists(value):
-            return value, (
-                f"[yellow]'{value}' does not exist in the workspace. "
-                '"exp run" may fail.[/]'
+            typ = "file" if is_file(value) else "directory"
+            return (
+                value,
+                f"[yellow]'{value}' does not exist, "
+                f"the {typ} will be created. ",
             )
     return value
+
+
+def is_file(path: str) -> bool:
+    _, ext = os.path.splitext(path)
+    return bool(ext)
+
+
+def init_deps(deps: Iterable["Dependency"]) -> List[str]:
+    from dvc.dependency import ParamsDependency
+    from dvc.fs.local import localfs
+
+    new_deps = [
+        dep
+        for dep in deps
+        if not isinstance(dep, ParamsDependency) and not dep.exists
+    ]
+    for dep in new_deps:
+        fs_path = dep.fs_path
+        if not is_file(fs_path):
+            localfs.makedirs(fs_path)
+            continue
+
+        localfs.makedirs(localfs.path.parent(fs_path), exist_ok=True)
+        with localfs.open(fs_path, "w", encoding="utf-8"):
+            pass
+
+    return [dep.def_path for dep in new_deps]
 
 
 def init(
@@ -287,6 +326,10 @@ def init(
         with _disable_logging(), repo.scm_context(autostage=True, quiet=True):
             stage.dump(update_lock=False)
             stage.ignore_outs()
+            if not interactive:
+                label = "Creating experiment project structure:"
+                display_workspace_tree(context, label)
+            init_deps(stage.deps)
             if params:
                 repo.scm_context.track_file(params)
     else:
--- a/dvc/repo/experiments/init.py
+++ b/dvc/repo/experiments/init.py
@@ -226,7 +226,7 @@ def init_deps(stage: PipelineStage) -> None:
 
     if new_deps:
         paths = map("[green]{0}[/]".format, new_deps)
-        ui.write(f"Creating {humanize.join(paths)}", styled=True)
+        ui.write(f"Created {humanize.join(paths)}.", styled=True)
 
     # always create a file for params, detect file/folder based on extension
     # for other dependencies
--- a/dvc/repo/experiments/init.py
+++ b/dvc/repo/experiments/init.py
@@ -154,7 +154,7 @@ def init_interactive(
         styled=True,
     )
 
-    tree_label = "Experiment project structure:"
+    tree_label = "Using experiment project structure:"
     display_workspace_tree(workspace, tree_label, stderr=True)
     ret.update(
         compact(
@@ -329,7 +329,7 @@ def init(
             stage.dump(update_lock=False)
             stage.ignore_outs()
             if not interactive:
-                label = "Experiment project structure:"
+                label = "Using experiment project structure:"
                 display_workspace_tree(context, label)
             init_deps(stage, log=not interactive)
             if params:
--- a/dvc/repo/experiments/init.py
+++ b/dvc/repo/experiments/init.py
@@ -215,7 +215,7 @@ def is_file(path: str) -> bool:
     return bool(ext)
 
 
-def init_deps(stage: PipelineStage, log: bool = False) -> None:
+def init_deps(stage: PipelineStage) -> None:
     from funcy import rpartial
 
     from dvc.dependency import ParamsDependency
@@ -224,7 +224,7 @@ def init_deps(stage: PipelineStage, log: bool = False) -> None:
     new_deps = [dep for dep in stage.deps if not dep.exists]
     params, deps = lsplit(rpartial(isinstance, ParamsDependency), new_deps)
 
-    if log:
+    if new_deps:
         paths = map("[green]{0}[/]".format, new_deps)
         ui.write(f"Creating {humanize.join(paths)}", styled=True)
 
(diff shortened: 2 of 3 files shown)


skshetry:
We may want to reconsider the output for interactive mode now, especially for the one that shows workspace structure, since we create some of the files by default:
```
This command will guide you to set up a train stage in dvc.yaml.
See https://s.dvc.org/g/pipeline-files.

Enter the paths for dependencies and outputs of the command.
DVC assumes the following workspace structure:
├── data
├── metrics.json
├── models
├── params.yaml
├── plots
└── src

Path to a code file/directory [src, n to omit]:
'src' does not exist, the directory will be created.
Path to a data file/directory [data, n to omit]:
'data' does not exist, the directory will be created.
Path to a model file/directory [models, n to omit]:
Path to a metrics file [metrics.json, n to omit]:
Path to a plots file/directory [plots, n to omit]:
```

pared (approved):

dberenbaum:
Thanks, @skshetry! What about params? Will they be added separately after #4112? It might be confusing to explain what paths do and don't get created automatically. At least if params are created, it can be simplified to "inputs get created, outputs don't."

dberenbaum:
Looks good!

Two issues to consider:
1. "Creating experiment project structure" may be misleading because outputs are not created, and inputs are only created if they don't already exist.
2. I agree that changes to the interactive output would be helpful, and specifically we should try to make interactive and non-interactive mode more consistent with each other.

Maybe we can just make the output more general?

```
$ dvc exp init --data raw python src/train.py
Experiment project structure:
├── metrics.json
├── models
├── params.yaml
├── plots
├── raw
└── src

'raw' does not exist, the directory will be created.
'src' does not exist, the directory will be created.
...
```

```
$ dvc exp init -i python src/train.py
Experiment project structure:
├── data
├── metrics.json
├── models
├── params.yaml
├── plots
└── src

Path to a code file/directory [src, n to omit]:
'src' does not exist, the directory will be created.
Path to a data file/directory [data, n to omit]: raw
'raw' does not exist, the directory will be created.
Path to a model file/directory [models, n to omit]:
Path to a metrics file [metrics.json, n to omit]:
Path to a plots file/directory [plots, n to omit]:
...
```

skshetry:
> Thanks, @skshetry! What about params? Will they be added separately after #4112? It might be confusing to explain what paths do and don't get created automatically. At least if params are created, it can be simplified to "inputs get created, outputs don't."

yeah, I was thinking of just doing that after #4112 as it'll be noop. But could do that now too, as `params.yaml` file is still read by default and is still usable.

skshetry:
> 'raw' does not exist, the directory will be created.
> 'src' does not exist, the directory will be created.

For interactive mode, I'll just merge these into a single message:
```
Creating raw/, src/ and params.yaml
```

skshetry:
Also, should the message be:
```
Using/Expecting/... Experiment project structure:
```

skshetry:
I decided to not create `params.yaml` by default in this PR as it requires some changes and thoughts on interactive mode. Let's do that in a successive PR. @dberenbaum

dberenbaum:
> For interactive mode, I'll just merge these into a single message:

Do you mean for non-interactive mode, or am I misunderstanding?

Also, I would just keep them separate unless you have a strong reason to merge them. I don't see an exact guideline for this, but I prefer a single piece of information per line. It keeps the clarity of whether each path is a file or directory, it's more consistent, and it's easier to parse.

> Using/Expecting/... Experiment project structure:

I think "Using" is fine.

skshetry:
> > For interactive mode, I'll just merge these into a single message:
> 
> Do you mean for non-interactive mode, or am I misunderstanding?

Yeah sorry, I meant non-interactive mode. 

> Also, I would just keep them separate unless you have a strong reason to merge them. I don't see an exact guideline for this, but I prefer a single piece of information per line. It keeps the clarity of whether each path is a file or directory, it's more consistent, and it's easier to parse.

Right, but is something being a file or a directory important information? It may cause visual noise without adding any meaningful information. 

This is what the current implementation displays:
```console
Creating data and src
```

skshetry:
> > I have decided to not create `params.yaml` by default in this PR as it requires some changes and thoughts on interactive mode.
> 
> You could also address that in a separate PR first and come back to this if it's easier since it would make params have less special behavior that needs to be handled.

This PR handles "inputs are created", so this can already create params by empty. We just need to get rid of those checks that we have before. I don't want to make too many decisions or changes in a single PR. And I need this PR before we move to params, as it unifies input creations.


> 
> I'm seeing inconsistent behavior when using flags with/without interactive mode:

Great catch with the UI messages. The first one is an oversight on this PR. "no message" on overrides has been fixed, it'll be shown even in interactive and non-interactive mode.

The params issue will be fixed after we remove the checks for params file existence.
 
> 
> I would prefer a unified workflow like:
> 
>     1. Get template to determine expected defaults. Print default tree if in interactive mode. I would disallow `--data` and other override flags in interactive mode (I think it's confusing, overly complicated, and I can't imagine why someone would need this).

This is something that will rarely be used. I prefer it as being an override, as ignoring them may be more confusing. Although I have no strong opinion.

> 
>     2. Iterate over each dependency and output to collect overrides (from prompts in interactive mode) and create dependencies as needed (output should show creation of each as it happens).

Yeah, fixed this. Thanks.

> 
>     3. Pass context to stage create method. I think we should print the same thing in both modes here, most likely the updated tree.

Makes sense, but how should we improve message before the prompts in an interactive mode?

dberenbaum:
> > ```
> > 1. Get template to determine expected defaults. Print default tree if in interactive mode. I would disallow `--data` and other override flags in interactive mode (I think it's confusing, overly complicated, and I can't imagine why someone would need this).
> > ```
> 
> 
> This is something that will rarely be used. I prefer it as being an override, as ignoring them may be more confusing. Although I have no strong opinion.

Yeah, I agree that ignoring the arguments is a bad option. I meant that `--interactive` should be mutually exclusive with `--data` and other overrides (including `--explicit`). It signals that the overrides are intended for non-interactive mode.

It can also reduce some of the `init_interactive()` logic. It wouldn't be necessary to adjust the prompts or workspace tree. That would help ensure consistency between interactive and non-interactive modes.

> > ```
> > 3. Pass context to stage create method. I think we should print the same thing in both modes here, most likely the updated tree.
> > ```
> 
> 
> Makes sense, but how should we improve message before the prompts in an interactive mode?

IMO we can get rid of printing the tree before the prompts to simplify, especially if we are printing the updated tree at the end.

I also think we can get rid of all the `if interactive`/`if not interactive` conditionals after `repo.stage.create()` and always print:

```
Using experiment project structure:
├── metrics.json
├── models
├── mydata
├── params.yaml
├── plots
└── src

Created train stage in dvc.yaml. To run, use "dvc exp run".
See https://s.dvc.org/g/exp/run.
```

Having a confirmation prompt in interactive mode after printing the tree seems okay if you think it's helpful to keep that, but I don't know that it's necessary.

skshetry:
> Yeah, I agree that ignoring the arguments is a bad option. I meant that `--interactive` should be mutually exclusive with `--data` and other overrides (including `--explicit`). It signals that the overrides are intended for non-interactive mode.
> 
> It can also reduce some of the `init_interactive()` logic. It wouldn't be necessary to adjust the prompts or workspace tree. That would help ensure consistency between interactive and non-interactive modes.

Right, but if you are using `--data` with `--interactive`, I think it's better to honor their request and think of it as an override, considering we already have support for that and documented the behavior in CLI help as:
```
  --interactive, -i     Prompt for values that are not provided
```

> 
> IMO we can get rid of printing the tree before the prompts to simplify, especially if we are printing the updated tree at the end.

I'll do this in the next PR. 

> 
> I also think we can get rid of all the `if interactive`/`if not interactive` conditionals after `repo.stage.create()` and always print: 

Yeah, let's get rid of it then. Also in the next PR. Thanks.

skshetry:
I am merging this PR. This is a https://github.com/iterative/dvc/labels/release-blocker now until few issues that we are discussing are merged, especially the params one.

PR #7353, opened 2022-02-07
dvcfs: detach from pipeline outputs

efiop:
Introducing simple repo tree that combines all workspaces into one structure,
and allows us to detach data representation from pipelines.

Pre-requisite for dvc-data extraction and using relpaths in dvcfs/repofs.

### DIFF
--- a/dvc/data/stage.py
+++ b/dvc/data/stage.py
@@ -168,7 +168,7 @@ def _build_tree(fs_path, fs, name, **kwargs):
         assert key
         tree.add(key, meta, obj.hash_info)
 
-        tree_meta.size += meta.size
+        tree_meta.size += meta.size or 0
         tree_meta.nfiles += 1
 
     return tree_meta, tree
--- a/dvc/data/tree.py
+++ b/dvc/data/tree.py
@@ -1,7 +1,7 @@
 import json
 import logging
 import posixpath
-from typing import TYPE_CHECKING, Dict, Optional, Tuple
+from typing import TYPE_CHECKING, Dict, Iterable, Optional, Tuple
 
 from funcy import cached_property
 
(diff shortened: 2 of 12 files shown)


dvc/fs/dvc.py
```diff
@@ -209,56 +163,52 @@ def isdvc(self, path, recursive=False, strict=True):
...
+            "outs": outs,
         }
 
-        if len(outs) > 1:
+        if len(outs) > 1 and outs[0][0] != key:
+            shortest = self.repo.index.tree.shortest_prefix(key)
+            if shortest:
+                assert shortest[1][1].isdir
+                if len(shortest[0]) <= len(key):
+                    ret["isdvc"] = True
```
efiop, line 190: These odd flags and logic will be removed in a followup, when we just use fs and dvcfs separately (with subrepos), instead of passing through tons of odd flags used inappropriately in ls and diff.

dvc/data/stage.py
```diff
@@ -168,7 +168,7 @@ def _build_tree(fs_path, fs, name, **kwargs):
         assert key
         tree.add(key, meta, obj.hash_info)
 
-        tree_meta.size += meta.size
+        tree_meta.size += meta.size or 0
```
karajan1001, line 171: Excuse me, why can not just set them default 0? Like in `C` or `Java` current `None` is just like `null` point in other languages.

dvc/data/tree.py
```diff
@@ -12,12 +12,29 @@
...
 
+def _try_load(
+    odbs: Iterable["ObjectDB"],
+    hash_info: "HashInfo",
+) -> Optional["HashFile"]:
+    for odb in odbs:
+        if not odb:
+            continue
+
+        try:
```
karajan1001, line 30: Can reduce one line.
```python
with suppress(FileNotFoundError, ObjectFormatError):
```

PR #7355, opened 2022-02-07
dvcfs: use paths relative to the repo root

efiop:
Currently we use both absolute and relative (depend on cwd) paths, which is
terrible for an api, as their interpretation depends on external factors.
This PR makes dvcfs accept only paths relative to the repo root for regular
outputs and absolute paths for external outputs.

Pre-requisite for making repofs use same relative paths.

### DIFF
--- a/dvc/api.py
+++ b/dvc/api.py
@@ -27,7 +27,8 @@ def get_url(path, repo=None, rev=None, remote=None):
             raise OutputNotFoundError(path, repo)
 
         cloud = info["repo"].cloud
-        md5 = info["repo"].dvcfs.info(fs_path)["md5"]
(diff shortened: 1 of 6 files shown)


dvc/api.py
```diff
@@ -27,7 +27,8 @@ def get_url(path, repo=None, rev=None, remote=None):
             raise OutputNotFoundError(path, repo)
 
         cloud = info["repo"].cloud
-        md5 = info["repo"].dvcfs.info(fs_path)["md5"]
+        dvc_path = _repo.fs.path.relpath(fs_path, info["repo"].root_dir)
```
efiop, line 30: Will go away after the switch in repofs. Also should use dvcfs with subrepo support here, which will be done in the future (have similar requests in ls and diff functionality).

Issue #621, opened 2018-04-05
implement 'dvc destroy'


rahulguptajss:
Can you please share an example of DVC remove. I could be missing something in its command.

efiop:
`dvc remove` removes outputs of the specified dvc file. Example:
```
echo foo > file
dvc add file
dvc remove file.dvc # Equivalent to `rm file` in this particular case
```
The usefulness in more obvious when your stage has multiple of outputs.

rahulguptajss:
Thanks I was hoping that if we could have something like dvc destroy which would be opposite to dvc init. Currently if i were to clean up my project of dvc, I have to remove all files manually.

efiop:
Ah, I see. Not going to lie, didn't even think of such feature until now :smiley: Turning this issue into feature request.

PR #7360, opened 2022-02-08
remove .dvcignore on destroy

### LINKED ISSUE #7354: `dvc destroy` leaves `.dvcignore` behind

dtrifiro:
fixes #7354

### DIFF
--- a/dvc/ignore.py
+++ b/dvc/ignore.py
@@ -399,3 +399,10 @@ def init(path):
         )
 
     return dvcignore
+
+
+def destroy(path):
+    from dvc.utils.fs import remove
+
+    dvcignore = os.path.join(path, DvcIgnore.DVCIGNORE_FILE)
+    remove(dvcignore)
--- a/dvc/repo/destroy.py
+++ b/dvc/repo/destroy.py
@@ -1,3 +1,4 @@
+from dvc.ignore import destroy as destroy_dvcignore
 from dvc.utils.fs import remove
 
 from . import locked
@@ -15,4 +16,5 @@ def _destroy_stages(repo):
 def destroy(repo):
     _destroy_stages(repo)
     repo.close()
+    destroy_dvcignore(repo.root_dir)
     remove(repo.dvc_dir)
--- a/tests/func/test_repo.py
+++ b/tests/func/test_repo.py
@@ -19,6 +19,7 @@ def test_destroy(tmp_dir, dvc, run_copy):
     # Remove all the files related to DVC
     for path in [
         ".dvc",
+        ".dvcignore",
         "file.dvc",
         "file2.dvc",
         "dir.dvc",


pmrowla (approved):

pared (changes requested):
What about `.dvcignore` files that are not in the root of the project? :)

pared (approved):
My bad, only root `.dvcignore` is created automatically.

skshetry:
We may still want to delete `.dvcignore` files on other locations. We do delete every `.dvc` and `dvc.yaml`/`dvc.lock` files.

dvc/repo/destroy.py
```diff
@@ -16,3 +17,4 @@ def destroy(repo):
     _destroy_stages(repo)
     repo.close()
     remove(repo.dvc_dir)
+    remove(DvcIgnore.DVCIGNORE_FILE)
```
skshetry, line 20: This does not remove the file from root of the repo, but from user's cwd.

skshetry (changes requested):

pared:
@skshetry thats right, thought the root one is the only one created by us, so I think this one is most important to handle here.

skshetry:
> @skshetry thats right, thought the root one is the only one created by us, so I think this one is most important to handle here.

I guess it depends on what `destroy` is supposed to be. As per the original request on https://github.com/iterative/dvc/issues/621#issuecomment-379520649, it seems it was supposed to be opposite of `dvc init` and in that case it makes sense.

If it's meant to get rid of `dvc` completely (as its name `destroy` suggests), it probably should remove everything. 

But I'm okay being conservative and removing file from the root of the repo. :)

dvc/repo/destroy.py
```diff
@@ -15,4 +16,5 @@ def _destroy_stages(repo):
 def destroy(repo):
     _destroy_stages(repo)
     repo.close()
+    destroy_dvcignore(repo.root_dir)
```
skshetry, line 19: I'd just prefer keeping things simple here:
```suggestion
    remove(os.path.join(repo.root_dir, DvcIgnore.DVCIGNORE_FILE)
```

dvc/repo/destroy.py
```diff
@@ -1,3 +1,4 @@
+from dvc.ignore import destroy as destroy_dvcignore
```
skshetry, line 1: You could import it as:
```
from dvc import ignore

ignore.destroy()
```

Also, let's make this lazy.

skshetry (approved):

dvc/ignore.py
```diff
@@ -399,3 +399,10 @@ def init(path):
         )
 
     return dvcignore
+
+
+def destroy(path):
+    from dvc.utils.fs import remove
+
+    dvcignore = os.path.join(path, DvcIgnore.DVCIGNORE_FILE)
+    remove(dvcignore)
```
skshetry, line 408: It makes sense to `init` a dvcignore inside a directory as it does not exist yet, but I am not sure if it makes sense for `destroy`.

dvc/repo/destroy.py
skshetry, line 1: Nitpick: You can import it as:
```suggestion
from dvc import ignore

ignore.destroy()
```

dvc/repo/destroy.py
dtrifiro: I think it makes sense to add `destroy()` next do `init()` to make it easier to keep init/destroy in sync, although I worry it's some form of premature optimization

Issue #7341, opened 2022-02-03
dvc.api.open: Will ignore remote option if a default remote is set in config.


PietrassykFP:
# Bug Report

## Description

In some cases, when passing the remote parameter. The remote specification will get lost during the `dvc pull` execution and will be overwritten by the default remote from the dvc config.

### Reproduce

1. Create new git Repo
2. Init dvc
3. Add two different non local remotes (perferably s3)
4. Make one of the remotes the default
5. Push the same data to both remotes
6. Push git Repo
7. From somewhere outside a dvc project run:
```python
with dvc.api.open(path="myfile", remote='non-default-remote') 
```
8. Track the actually used remote with a Debugger. Or make sure both remotes require different Authentification and just provide the non-defaut-remote authentification during the api call. You will run into Permission errors.

### Expected

If the remote argument is passed this remote should allways be used, when opening files.

### Environment information

**Output of `dvc doctor`:**

```console
DVC version: 2.9.3 (pip)
---------------------------------
Platform: Python 3.8.12 on Linux-5.10.47-linuxkit-x86_64-with-glibc2.2.5
Supports:
        hdfs (fsspec = 2022.1.0, pyarrow = 1.0.1),
        webhdfs (fsspec = 2022.1.0),
        http (aiohttp = 3.8.1, aiohttp-retry = 2.4.6),
        https (aiohttp = 3.8.1, aiohttp-retry = 2.4.6),
        s3 (s3fs = 2022.1.0, boto3 = 1.20.24)
root@112f43f4f86b:/workspaces/centerline-identification-job# 
```

pared:
Initial discussion:
https://discord.com/channels/485586884165107732/485596304961962003/933674011982983219

@PietrassykFP has been setting up the credentials via environmental variables, so if the remote gets resolved properly, it should not raise issues. I created a local test aiming to verify if remote gets resolved properly  and it seem to be working properly:

```
def test_open_from_proper_remote(tmp_dir, erepo_dir, make_tmp_dir):
    default_remote = make_tmp_dir("default_storage")
    other_remote = make_tmp_dir("other_storage")

    with erepo_dir.chdir():
        with erepo_dir.dvc.config.edit() as conf:
            conf["remote"]["default"] = {"url": str(default_remote)}
            conf["core"]["remote"] = "default"
            conf["remote"]["other"] = {"url": str(other_remote)}

        erepo_dir.dvc_gen({"foo": "foo content"}, commit="create file")
[...]

PR #7363, opened 2022-02-08
tests: add test for dvc.api.open with remote

skshetry:
Covers #7341.

### DIFF
--- a/tests/func/test_api.py
+++ b/tests/func/test_api.py
@@ -205,3 +205,17 @@ def test_get_url_subrepos(tmp_dir, scm, local_cloud):
         local_cloud / "37" / "b51d194a7513e45b56f6524f2d51f2"
     )
     assert api.get_url("subrepo/bar") == expected_url
+
+
+@pytest.mark.xfail(reason="https://github.com/iterative/dvc/issues/7341")
+def test_open_from_remote(tmp_dir, erepo_dir, cloud, local_cloud):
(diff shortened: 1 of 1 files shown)


skshetry:
@efiop, this still fails on:

https://github.com/iterative/dvc/blob/b6fa1178e71353e0cd1fa4b5b6b638591fe08453/dvc/fs/dvc.py#L172

efiop:
On closer inspection, unlike before data tree introduction, this rises a somewhat unnoticed before problem of `remote` argument and metadata that the repo structure is getting built from. Plus every output might use its own remote, so `remote` becomes even more questionable. What we might be looking for here is some kind of config (mapping?) adjustment. Some use cases that come to mind:

1) change default core.remote to something else (`dvc config core.remote myremote`)
2) adjust particular remote configuration (e.g. add some keys or something) (`dvc remote modify myremote key mykey`)
3) adjust remotes for particular outputs (currently no CLI command, so can only be modified by opening dvc.yaml/data.dvc and editing by hand)

For the most part, it seems like current `remote=` argument is mostly used to emulate 1).

Overall this seems to be the same problem as providing config through cli/api, e.g. https://github.com/iterative/dvc/issues/2466

1 and 2 can be solved by supporting `config=` in API (the values will be merged with repo config and data tree will be built using it, so you won't need to provide `remote=` to open at all, which is nice), but option 3 seems to require something more advanced (data tree editing?).

PR #7367, opened 2022-02-10
plots: json: add datapoints to vega plots

pared:

Usage:

`dvc plots show/diff --json --split`

The results for `vega` will contain additional field `datapoints` and the template will not have the data filled (there will be `<DVC_METRIC_DATA>` in datapoints destination.)

As of now the `datapoints` field will exist no matter whether `--split` has been invoked or not. The only difference is that in case when not providing `--split`, the content of particular vega json will have the data filled.

cc @mattseddon

### DIFF
--- a/dvc/commands/plots.py
+++ b/dvc/commands/plots.py
@@ -14,12 +14,14 @@ from dvc.utils import format_link
 logger = logging.getLogger(__name__)
 
 
-def _show_json(renderers, path: None):
+def _show_json(renderers, path: None, split=False):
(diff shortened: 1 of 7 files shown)


pared:
EDIT:
As per our discussion with @mattseddon, the `datapoints` contains now a map `revision`: `datapoints`

pared:
After talking with @mattseddon it seems that the current state will be ok for now. This is not necessarily what we want for `--json` in the long term, but should be ok for now. Since `--json` is not officially supported, we don't have to commit to the current state in the future.

PR #7393, opened 2022-02-16
Add log statement to collect repo

Erotemic:
Minor change that adds a debugging log statement. This makes `dvc add -v` list the files it is walking, and makes it easier for a developer to find the folder that is making `dvc add` slow.

Motivating case: I was running `dvc add` on a directory, and it was taking a very long time (hours). The only info I got was that it was collecting stages, even after I ran it with a `-v`. I eventually realized there was a very large, nested folder that was not in my `.dvcignore` file, so `dvc add` was trying to walk that entire directory structure before it did anything else.

I would have realized this much quicker if `dvc add` had more info when `-v` was given. In fact, I wasn't originally sure what the issue was, I was tracking through the DVC code to see why `add` would be slow, and found this section where it walks the entire directory. I put a print statement in originally, but then thought that this might be more generally useful, so I changed it to a logging.debug statement and submitted this PR.

I don't _think_ a logging statement would add significant overhead here, but I didn't test it.

### DIFF
--- a/dvc/repo/stage.py
+++ b/dvc/repo/stage.py
@@ -473,6 +473,8 @@ class StageLoad:
         for root, dirs, files in self.repo.dvcignore.walk(
             self.fs, self.repo.root_dir
         ):
+            logger.trace(  # type: ignore[attr-defined]
+                "Collect repo walking '%s'", root)
             dvcfile_filter = partial(is_dvcfile_and_not_ignored, root)
             for file in filter(dvcfile_filter, files):
                 file_path = os.path.join(root, file)
--- a/dvc/repo/stage.py
+++ b/dvc/repo/stage.py
@@ -474,7 +474,8 @@ class StageLoad:
             self.fs, self.repo.root_dir
         ):
             logger.trace(  # type: ignore[attr-defined]
-                "Collect repo walking '%s'", root)
+                "Collect repo walking '%s'", root
+            )
             dvcfile_filter = partial(is_dvcfile_and_not_ignored, root)
             for file in filter(dvcfile_filter, files):
                 file_path = os.path.join(root, file)


dvc/repo/stage.py
```diff
@@ -473,6 +473,7 @@ def is_out_or_ignored(root, directory):
         for root, dirs, files in self.repo.dvcignore.walk(
             self.fs, self.repo.root_dir
         ):
+            logger.debug(f"Collect Repo Walking {root}")
```
pmrowla, line 476: This should probably be `trace` level instead of `debug` (so it would be available with `-vv` but not `-v`), as it will generate really verbose output anywhere we are collecting stages (which affects more than just `dvc add`)

Also, even though we use f-strings most of the time, for logging we generally try to use lazy formatting like
```py
logger.trace("Collect repo walking '%s'", root)
```

skshetry:
> Motivating case: I was running `dvc add` on a directory, and it was taking a very long time (hours). The only info I got was that it was collecting stages, even after I ran it with a `-v`. I eventually realized there was a very large, nested folder that was not in my `.dvcignore` file, so `dvc add` was trying to walk that entire directory structure before it did anything else.

We have a specific optimization to avoid that. It should never enter the tracked outputs directory unless we have a bug somewhere.
https://github.com/iterative/dvc/blob/795fe24a5f4b68b6661b3986997db403ebedf08e/dvc/repo/stage.py#L494

Erotemic:
@skshetry The issue is that I didn't know that I had a large nested directory that wasn't in a .gitignore file. It wasn't an "output" directory. I only use DVC for data storage, I don't use any of the model tracking features. It was just a large processed dataset that I was going to add to DVC but hadn't done yet. I didn't even know `dvc add` walked the entire repo when I initially noticed it was taking hours to finish.

As an aside I dug into what was doing the walk, and I was curious if an `os.scandir` solution might be faster.

I'll also note that one of the main reasons this walk was taking hours was because I was training a model at the same time which was making heavy use of IO. So that slowed down all the other ops. However, even when I killed the model it still took about a half an hour to get through the walk step.

dvc/repo/stage.py
dberenbaum: @Erotemic Would you mind addressing this comment?

dberenbaum:
@iterative/dvc What's the status here? Any remaining concerns to get this merged?

Erotemic:
I'm running into another use case where having this patch would be very useful. I know I must have a big untracked directory in my dvc repo somewhere, but there isn't currently an easy way to find it. If I had this debug trace, I could quickly see where the slowdown is coming from. 

I've rebased this on the latest main, and I've manually disabled the mypy error on `logger.trace`. I have absolutely no idea (1) why other instances of `logger.trace` don't raise this error and (2) where `logger.trace` is even defined. It does seem to be defined, but the code is not forthcoming as to where that happens.

pmrowla:
> I'm running into another use case where having this patch would be very useful. I know I must have a big untracked directory in my dvc repo somewhere, but there isn't currently an easy way to find it. If I had this debug trace, I could quickly see where the slowdown is coming from.
> 
> I've rebased this on the latest main, and I've manually disabled the mypy error on `logger.trace`. I have absolutely no idea (1) why other instances of `logger.trace` don't raise this error and (2) where `logger.trace` is even defined. It does seem to be defined, but the code is not forthcoming as to where that happens. If there is a better way to let mypy know that my usage of logger.trace is not an error, let me know.

You need to add the inline `# type: ignore[attr-defined]`, ex:
https://github.com/iterative/dvc/blob/30d6839a5f471fb02a7a2dcbd6a6616b71277762/dvc/ui/pager.py#L89

pmrowla (approved):

dberenbaum:
@skshetry Do you want to review this one?

dvc/repo/stage.py
```diff
@@ -473,6 +473,9 @@ def is_out_or_ignored(root, directory):
         for root, dirs, files in self.repo.dvcignore.walk(
             self.fs, self.repo.root_dir
         ):
+            logger.trace(  # type: ignore[attr-defined]
+                "Collect repo walking '%s'", root
+            )
```
skshetry, line 478: Hey, @Erotemic, What do you think of introducing a logger that also tracks the duration of the walk instead?

```python
walk_iter = self.repo.dvcignore.walk(self.fs, self.repo.root_dir)
if logger.isEnabledFor(logging.TRACE):
    walk_iter = log_walk(walk_iter)

for root, dirs, files in walk_iter:
    ...
```

Let me know, otherwise, we can take over. Thanks. Sorry for the late review.

skshetry: Also, we could repurpose some code from [funcy.log_iter_durations](https://github.com/Suor/funcy/blob/60910f8e601c9112312a02037a9c7d9c465e2daf/funcy/debug.py#L186), especially the `format_time`.

Erotemic: I'm not sure how relevant logging the time is, unless DVC could use that to improve the overall speed of this step (it's getting quite sluggish as my DVC repo size increases). The motivating use case here is to show a user where DVC is spending a lot of time --- its quite obvious what is taking time without explicit logging of it because you see the same root directory over and over. Also, the time for each iteration will be roughly the same, so it wont help cue a user into what directory is causing the issue.

Overall, I'm not opposed to logging the time, I would just rather see any indication of what directory is causing the issue in the new release sooner rather than later.

On the other hand, if DVC used that time information to *notice* that a directory was taking a long time and then printed out an INFO statement to inform the user without `-vv` being given, then I think that would be useful, but maybe better suited for another PR. I don't have time to implement that myself.

As an aside, the logging in funcy probably should be using `time.perf_counter` (or `time.perf_counter_ns` if available) instead of `time.time`.

skshetry:
Thanks, @Erotemic. We are discussing about introducing time logging on the trace. But I don't want to keep this PR hanging till then as it has already been quite a long time. 

Again, sorry for very late review 🙂, merging ...

Issue #6720, opened 2021-09-30
diff: ERROR: unexpected error - 'not in cache' 


josecelano:
## Description

When I tried to compare two commits I get an error.

The command is: `dvc diff f0c0269 16ba2cd`
I'm using your sample repo: https://github.com/iterative/example-get-started

### Reproduce

**Command 1** (compare commit with previous "dvc add" commit):
```
dvc diff f0c0269 16ba2cd
```

Output:
```
ERROR: unexpected error - 'not in cache'                              

Having any troubles? Hit us up at https://dvc.org/support, we are always happy to help!
```

**Command 2** (compare commit with a commit before "dvc add" commit):
```
dvc diff 3e07290 16ba2cd
```

Output:
```
Added:                                                                
    data/data.xml

files summary: 1 added
```

How to reproduce it:
```
git clone https://github.com/iterative/example-get-started
cd example-get-started/
dvc diff 16ba2cd f0c0269
```

### Expected

These are the first commits in your sample repo:
```
* 08a3b89 - (tag: 4-import-data) Import raw data (overwrite) (2021-05-20 23:56:52 -0400) <Olivaw Owlet>
* 16ba2cd - (tag: 3-config-remote) Configure default remote (2021-05-19 20:10:12 -0400) <Olivaw Owlet>
* f0c0269 - (tag: 2-track-data) Add raw data (2021-05-18 16:23:32 -0400) <Olivaw Owlet>
* 3e07290 - (tag: 1-dvc-init) Initialize DVC project (2021-05-17 12:36:52 -0400) <Olivaw Owlet>
* 90b2aea - (tag: 0-git-init) Initialize Git repository (2021-05-16 08:50:12 -0400) <Olivaw Owlet>
```

If you show changes on the third commit `f0c0269` with `git show f0c0269` you can see:
```
commit f0c026993d3dc8aa4d79426acb59b81dbb63e5a2 (tag: 2-track-data)
Author: Olivaw Owlet <64868532+iterative-olivaw@users.noreply.github.com>
Date:   Tue May 18 16:23:32 2021 -0400

    Add raw data

diff --git a/data/.gitignore b/data/.gitignore
new file mode 100644
index 0000000..fc12be5
--- /dev/null
+++ b/data/.gitignore
@@ -0,0 +1 @@
+/data.xml
diff --git a/data/data.xml.dvc b/data/data.xml.dvc
new file mode 100644
index 0000000..8fab3a9
--- /dev/null
[...]

PR #7406, opened 2022-02-18
diff: fix missing key bug (#6720)

dberenbaum:

* [N/A] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here.

`dvc diff` fails when there is no diff and `hide_missing` is true (`hide_missing` is implied when comparing two revisions) because `CmdDiff.run` tries to delete the `not in cache` info from an empty dict. It seemed easy enough to fix.

Not sure if we need a test for this. Couldn't find any relevant ones specific to the `CmdDiff` logic.

### DIFF
--- a/tests/unit/command/test_diff.py
+++ b/tests/unit/command/test_diff.py
@@ -204,20 +204,31 @@ def test_diff_show_markdown_and_hash(mocker, show_hash):
     mock_show_markdown.assert_called_once_with(diff, show_hash, False)
 
 
-def test_no_changes(mocker, capsys):
-    args = parse_args(["diff", "--json"])
+@pytest.mark.parametrize(
+    "opts",
+    (
+        [],
+        ["a_rev", "b_rev"],
+        ["--targets", "."],
+        ["--hide-missing"],
+    ),
+)
+@pytest.mark.parametrize(
+    "show, expected",
+    (
+        ([], ""),
+        (["--json"], "{}"),
+        (["--md"], "| Status   | Path   |\n|----------|--------|"),
+    ),
+)
+def test_no_changes(mocker, capsys, opts, show, expected):
+    args = parse_args(["diff", *opts, *show])
     cmd = args.func(args)
     mocker.patch("dvc.repo.Repo.diff", return_value={})
 
     assert 0 == cmd.run()
     out, _ = capsys.readouterr()
-    assert "{}" in out
-
-    args = parse_args(["diff"])
-    cmd = args.func(args)
-    assert 0 == cmd.run()
-    out, _ = capsys.readouterr()
-    assert not out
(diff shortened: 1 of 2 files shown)


dberenbaum:
Thanks @daavoo! Added a test.

tests/unit/command/test_diff.py
```diff
@@ -220,6 +220,22 @@ def test_no_changes(mocker, capsys):
     assert not out
 
 
+def test_no_changes_hide_missing(mocker, capsys):
+    args = parse_args(["diff", "--json", "--hide-missing"])
+    cmd = args.func(args)
+    mocker.patch("dvc.repo.Repo.diff", return_value={})
```
skshetry, line 226: Please use `pytest.mark.parametrize` in [test_no_changes](https://github.com/dberenbaum/dvc/blob/00d854210f8ff6446335616097a8645b826e04ae/tests/unit/command/test_diff.py#L207).

skshetry:
I'd prefer fixing this in `repo.diff()` to always return the same schema even if things didn't change.
So, instead of an empty dictionary, I'd prefer it to return the data as:

```python
>>> repo.diff()
{
    'added': [],
    'deleted': [],
    'modified': [],
    'renamed': [],
    'not in cache': []
}
```

tests/unit/command/test_diff.py
skshetry: You can change the above test `test_no_changes` to something like following:
```python
@pytest.mark.parametrize("opts", ([], ["a_rev", "b_rev"], ["--hide-missing"]))
def test_no_changes(mocker, capsys, opts):
    args = parse_args(["diff", "--json", *opts])
    cmd = args.func(args)
    mocker.patch("dvc.repo.Repo.diff", return_value={})

    assert 0 == cmd.run()
    out, _ = capsys.readouterr()
    assert "{}" in out

    args = parse_args(["diff", *opts])
    cmd = args.func(args)
    assert 0 == cmd.run()
    out, _ = capsys.readouterr()
    assert not out
```

dberenbaum: Sorry, just got a little lazy/sloppy here trying to fix this as quickly as possible.

skshetry (approved):

tests/unit/command/test_diff.py
dberenbaum: @skshetry Updated the tests with parametrized options and output formats.

dberenbaum: @skshetry Pinging on this. Could you review the updated tests when you have a chance? 🙏

Issue #6944, opened 2021-11-09
plots: extract rendering functionality to separate lib/submodule


pared:
I think we need to discuss extracting `dvc.plots.show` to a separate lib.

### Why?
1. `dvc plots` used to be only about Vega, and we made the user use a particular convention (saving plots as data files) in order to provide meaningful results in `dvc plots show/diff` commands. 
2. turns out its hard to force user's to use only one convention - hence iterative/dvc#6431, and ideas like iterative/dvc-render#7 
3. loosening the plot's requirements made us focus the `dvc` commands on providing the data rather than a visualization as a whole, and allowed us to separate the HTML rendering from the `Repo.plots` methods
4. `dvclive` could use rendering too - but we don't want to add whole DVC for `plotting` purposes only
5. Seems like after iterative/dvc#6900 we will need to do something about generated HTML and js testing, which sounds flaky, so a separate testing suite might come in handy

Related: iterative/dvclive#61, #5980 

cc @daavoo @dberenbaum

PR #7409, opened 2022-02-21
render: Extract to `dvc_render`

daavoo:

Fixes https://github.com/iterative/dvc/issues/6944

Extracted rendering logic to https://github.com/iterative/dvc-render .

Logic that remains on DVC side:

- `dvc.render.match.match_renderers`: Matching of dvc "raw plots data" with the corresponding dvc-render `Renderer`
- `dvc.render.convert.to_datapoints`: Conversion from dvc "raw plots data" (return of `dvc.repo.plots.show()`) to dvc-render `datapoints`.
- `dvc.render.convert.to_json`: Conversion from `Renderer` instances to `json` output.


dvc/render/convert.py
```diff
@@ -0,0 +1,56 @@
...
+                processed, final_props = converter.convert(
+                    revision, filename, file_data.get("data")
+                )
+                datapoints.extend(processed)
+    return datapoints, final_props
+
+
+def to_json(renderer) -> List[Dict]:
+    if renderer.TYPE == "vega":
+        # TODO: Adapt to #7367
```
daavoo, line 34: @pared whichever format is decided in https://github.com/iterative/dvc/pull/7367 I think all the information could be extracted from a renderer instance so the conversion logic can be isolated here.

dvc/commands/plots.py
```diff
@@ -72,27 +71,35 @@ def run(self):
...
             if self.args.show_vega:
                 renderer = first(
-                    filter(lambda r: isinstance(r, VegaRenderer), renderers)
+                    filter(lambda r: r.TYPE == "vega", renderers)
                 )
                 if renderer:
-                    ui.write_json(renderer.asdict())
+                    ui.write_json(json.loads(renderer.partial_html()))
                 return 0
             if self.args.json:
```
daavoo, line 81: IMO, `--json` should be returning the raw `plots_data` . 

This would be consistent with `exp show --json` and would prevent us from maintaining and updating conversion logic.

dvc/commands/plots.py
```diff
@@ -152,16 +159,16 @@ class CmdPlotsTemplates(CmdBase):
 
     def run(self):
         import os
-
+        from dvc_render.vega_templates import dump_templates
```
daavoo, line 164: I considered that the logic for dumping and loading templates made more sense in `dvc-render`. The only link between `repo` and the templates is now `repo.plots.templates_dir`

dvc/render/image_converter.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def convert(
+        self, revision, filename, data: bytes
+    ) -> Tuple[List[Dict], Dict]:
+        """
+        Convert the DVC Plots content to DVC Render datapoints.
+        Return both generated datapoints and updated properties.
+        """
+        path = self.plot_properties.get("out")
+        if path:
```
daavoo, line 45: @pared I made `out` optional here. This should still work as before when `out` is provided but now also works without output folder by encoding images with base64.

dvc/commands/plots.py
pared: I agree, thought on the other hand I would presume that users expect the same plots for DVC and vs code, even more than they would expect same plots for DVC and studio. And we already had some problems [there](https://github.com/iterative/viewer/issues/2574). @mattseddon what would you prefer more? Having datapoints created on our side and just filling the template with them or make to data -> datapoints conversion yourself?

pared (commented):
Great work @daavoo. 
I would be ok with emerging as is, though I would like for us to come to a conclusion regarding one thing:

`partial_html` name - I created this name, and after having it around for some time, it seems to me as a rather poor attempt to abstract some behavior that is not that easy to abstract. Since it will be a method of a Renderer API, it seems to me that we should reconsider it. As I am looking at the usage, usually we need to do `if render.TYPE == something` and only then call some methods. Maybe we should move `partial_html` to `_partial_html`, use it only internally to fill the DIV, and have other methods exposed, with less confusing names. For example for `Vega` it could be `get_filled_template` or `fill_template` - this at least provides some hint of what we are doing, and since we are checking the render `TYPE` everywhere, trying to have an abstract method does not make sense. What do you think?

daavoo:
> partial_html name - I created this name, and after having it around for some time, it seems to me as a rather poor attempt to abstract some behavior that is not that easy to abstract. Since it will be a method of a Renderer API, it seems to me that we should reconsider it. As I am looking at the usage, usually we need to do if render.TYPE == something and only then call some methods. Maybe we should move partial_html to _partial_html, use it only internally to fill the DIV, and have other methods exposed, with less confusing names. For example for Vega it could be get_filled_template or fill_template - this at least provides some hint of what we are doing, and since we are checking the render TYPE everywhere, trying to have an abstract method does not make sense. What do you think?

Makes sense to me. Will apply changes to `dvc-render` and update this P.R.

pared (approved):

dvc/render/vega_converter.py
```diff
@@ -76,20 +76,14 @@ def _append_index(datapoints: List[Dict]) -> List[Dict]:
...
     If some properties that are required by Template class are missing
     ('x', 'y') it will attempt to fill in the blanks.
     """
 
-    @staticmethod
-    def update(datapoints: List[Dict], update_dict: Dict):
-        for data_point in datapoints:
-            data_point.update(update_dict)
-        return datapoints
-
```
pared, line 92: Lets keep that and its usage, this is one of the functionalities created for viewer. Also, the fact that it was used outside of `convert` was intentional.

pared (changes requested):
One thing regarding `update`

dvc/render/vega_converter.py
pared: Alternative would be to consider updating `filename` and `revision` as steps that could be turned off, since we are using this mechanism.

daavoo: Updated the P.R. making `filename` and `revision` optional.

Also opened draft P.R. on studio https://github.com/iterative/studio/pull/3342 (locally I get tests to pass for studio)

dvc/commands/plots.py
pared: Let's leave this behavior as is for now. `--json` is not official, if VSCode will prefer to do the data processing by themselves, we will change it.

pared (approved):

pared:
Great job 🚀 !

PR #7412, opened 2022-02-22
[WIP] fs: use taskset ThreadPoolExecutor in download_dir

### LINKED ISSUE #7395: Does `dvc get` need to hold open all the files it tries to download?
I got dataset of 5000 files. I tried to download the dataset on the server using `dvc get` but it failed. `Too many open file`

I try to use `ulimit -n 4096` and `--jobs 1` and `_JOBS=1` (as I use SFTP) But it always fails.
https://dvc.org/doc/user-guide/troubleshooting#many-files

On the other hand `dvc import/pull` works fine. 

**So question is does `dvc get` needs to keep opened all that files?**

```
DVC version: 2.9.4 (pip)
Platform: Python 3.7.5 on Linux-4.15.0-166-generic-x86_64-with-Ubuntu-18.04-bionic
Supports:
        webhdfs (fsspec = 2022.1.0),
        ssh (sshfs = 2021.11.2)
```

pmrowla:

May fix #7395

In odb/data transfer we use an optimized version of `ThreadPoolExecutor.map` that splits the operation up into chunked tasksets based on the # of jobs, rather than the default behavior of keeping all of the possible futures open until the entire executor is closed. This reduces memory and open file descriptor usage when transferring large #'s of files. We should be using the same optimization in fs download (`dvc get`).

### DIFF
--- a/dvc/fs/base.py
+++ b/dvc/fs/base.py
@@ -1,7 +1,6 @@
 import contextlib
 import logging
 import os
-from concurrent.futures import ThreadPoolExecutor, as_completed
 from functools import partialmethod
 from multiprocessing import cpu_count
 from typing import ClassVar, Dict, Optional
@@ -14,6 +13,7 @@ from dvc.fs._callback import DEFAULT_CALLBACK, FsspecCallback
 from dvc.ui import ui
 from dvc.utils import tmp_fname
 from dvc.utils.fs import makedirs, move
+from dvc.utils.threadpool import ThreadPoolExecutor
 
 logger = logging.getLogger(__name__)
 
@@ -317,25 +317,19 @@ class FileSystem:
         download_files = FsspecCallback.wrap_fn(callback, self._download_file)
         max_workers = jobs or self.jobs
         with ThreadPoolExecutor(max_workers=max_workers) as executor:
-            futures = [
-                executor.submit(download_files, from_info, to_info)
-                for from_info, to_info in zip(from_infos, to_infos)
-            ]
-
             # NOTE: unlike pulling/fetching cache, where we need to
             # download everything we can, not raising an error here might
             # turn very ugly, as the user might think that he has
             # downloaded a complete directory, while having a partial one,
             # which might cause unexpected results in his pipeline.
-            for future in as_completed(futures):
-                # NOTE: executor won't let us raise until all futures that
-                # it has are finished, so we need to cancel them ourselves
-                # before re-raising.
-                exc = future.exception()
-                if exc:
-                    for entry in futures:
-                        entry.cancel()
-                    raise exc
+            try:
+                for _ in executor.imap_unordered(
+                    download_files, from_infos, to_infos, cancel_futures=True
+                ):
+                    pass
+            except Exception:
+                executor.shutdown(wait=False)
+                raise
 
     def _download_file(
         self,
--- a/dvc/utils/threadpool.py
+++ b/dvc/utils/threadpool.py
@@ -13,11 +13,18 @@ class ThreadPoolExecutor(futures.ThreadPoolExecutor):
         return self._max_workers
 
     def imap_unordered(
-        self, fn: Callable[..., _T], *iterables: Iterable[Any]
+        self,
+        fn: Callable[..., _T],
+        *iterables: Iterable[Any],
+        cancel_futures: bool = False,
     ) -> Iterator[_T]:
         """Lazier version of map that does not preserve ordering of results.
 
         It does not create all the futures at once to reduce memory usage.
+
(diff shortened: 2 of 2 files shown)


dvc/fs/base.py
```diff
@@ -317,25 +317,19 @@ def _download_dir(
...
-                    for entry in futures:
-                        entry.cancel()
-                    raise exc
+            try:
+                for _ in executor.imap_unordered(
+                    download_files, from_infos, to_infos
+                ):
+                    pass
+            except Exception:
+                executor.shutdown(wait=False, cancel_futures=True)
```
skshetry, line 331: [`cancel_futures`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.shutdown) is only available since Python3.9.

pmrowla: changed this to allow cancelling unfinished tasks when en exception is raised in our `imap_unordered` implementation

pmrowla:
We should probably be doing the same thing everywhere else that we are still using `ThreadPoolExecutor` as well: https://github.com/iterative/dvc/blob/cb584c6a53a4f3b5157de8715cf82c55de334c3b/dvc/data/stage.py#L139-L142

Will update this PR assuming it actually fixes the original SSH `dvc get` issue

dvc/fs/base.py
```diff
@@ -317,25 +317,19 @@ def _download_dir(
...
-                        entry.cancel()
-                    raise exc
+            try:
+                for _ in executor.imap_unordered(
+                    download_files, from_infos, to_infos, cancel_futures=True
+                ):
+                    pass
+            except Exception:
+                executor.shutdown(wait=False)
+                raise
```
skshetry, line 332: We probably don't need this, since we cancel pending futures explicitly, and `ThreadPoolExecutor.__exit__()` will be called with `wait=True` which should be a noop.

pared (approved):

daavoo:
> We should probably be doing the same thing everywhere else that we are still using `ThreadPoolExecutor` as well:
> 
> https://github.com/iterative/dvc/blob/cb584c6a53a4f3b5157de8715cf82c55de334c3b/dvc/data/stage.py#L139-L142
> 
> Will update this PR assuming it actually fixes the original SSH `dvc get` issue



> We should probably be doing the same thing everywhere else that we are still using `ThreadPoolExecutor` as well:
> 
> https://github.com/iterative/dvc/blob/cb584c6a53a4f3b5157de8715cf82c55de334c3b/dvc/data/stage.py#L139-L142
> 
> Will update this PR assuming it actually fixes the original SSH `dvc get` issue

cc @dtrifiro might be of relevance

dberenbaum:
@pmrowla What's the status of this PR?

efiop:
Looking at `dvc/fs/base.py`, considering maybe dropping our `download/upload` logic completely and living it up to particular filesystem implementation that will either use async or will download one-by-one. That might turn out to be a bit slower for some filesystems, but we can introduce threading inside of those particular filesystems. @pmrowla Do you think that would fix the original issue here?

efiop:
For the record, @skshetry will likely be getting rid of `download` from `FileSystem` as a part of https://github.com/iterative/dvc/issues/7543 , but maybe that code with the changes from this PR will find its way into particular fs implementations that need it.

pmrowla:
> @pmrowla What's the status of this PR?

I hadn't followed up on this PR yet since it does not resolve the original too many open file descriptors issue w/SSH `dvc get`.

> Looking at `dvc/fs/base.py`, considering maybe dropping our `download/upload` logic completely and living it up to particular filesystem implementation that will either use async or will download one-by-one. That might turn out to be a bit slower for some filesystems, but we can introduce threading inside of those particular filesystems. @pmrowla Do you think that would fix the original issue here?

Yeah, I think that might work.

dvc/utils/threadpool.py
```diff
@@ -29,6 +36,12 @@ def create_taskset(n: int) -> Set[futures.Future]:
...
             )
-            for fut in done:
-                yield fut.result()
+            try:
+                for fut in done:
+                    yield fut.result()
+            except Exception:
+                if cancel_futures:
+                    for fut in tasks:
+                        fut.cancel()
```
skshetry, line 45: @pmrowla, we don't need this if we backport the `cancel_futures=True` support to `Python<3.9`, right? 

We may need to extend `Executor.__exit__` to send `cancel_futures=True` as well. Let me try to propose something, hmm.

pmrowla: Right

skshetry: Proposed in #7602.

PR #7413, opened 2022-02-22
exp init: create params file by default

skshetry:
Note that this won't add support for tracking the params file,
it'll only create the params file. Some changes on error and
prompts are made.

### DIFF
--- a/dvc/dependency/param.py
+++ b/dvc/dependency/param.py
@@ -1,4 +1,3 @@
-import errno
 import logging
 import os
 import typing
@@ -133,29 +132,17 @@ class ParamsDependency(Dependency):
 
     def validate_filepath(self):
         if not self.exists:
-            raise FileNotFoundError(
-                errno.ENOENT, os.strerror(errno.ENOENT), str(self)
-            )
+            raise MissingParamsFile(f"Parameters file '{self}' does not exist")
         if self.isdir():
-            raise IsADirectoryError(
-                errno.EISDIR, os.strerror(errno.EISDIR), str(self)
+            raise ParamsIsADirectoryError(
+                f"'{self}' is a directory, expected a parameters file"
             )
 
     def read_file(self):
         _, ext = os.path.splitext(self.fs_path)
         loader = LOADERS[ext]
 
-        try:
-            self.validate_filepath()
-        except FileNotFoundError as exc:
-            raise MissingParamsFile(
-                f"Parameters file '{self}' does not exist"
-            ) from exc
-        except IsADirectoryError as exc:
-            raise ParamsIsADirectoryError(
-                f"'{self}' is a directory, expected a parameters file"
-            ) from exc
-
+        self.validate_filepath()
         try:
             return loader(self.fs_path, fs=self.repo.fs)
         except ParseError as exc:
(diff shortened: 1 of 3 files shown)


dberenbaum:
> Note that this won't add support for tracking the params file,
> it'll only create the params file.

So the purpose is to allow `dvc exp init` from an empty repo without failing?

Is it possible to get to #4112 before this? To actually use params, you would have to edit `params.yaml` and `dvc.yaml`, right? IMO that's more confusing than failing.

Note: If this PR does move forward, it creates `params.yaml` regardless of the value of `--params`.

dvc/repo/experiments/init.py
```diff
@@ -159,32 +159,24 @@ def validate_prompts(
...
+            )
     elif key in ("code", "data"):
         if not os.path.exists(value):
             typ = "file" if is_file(value) else "directory"
-            return (
-                value,
-                f"[yellow]'{value}' does not exist, "
-                f"the {typ} will be created. ",
-            )
+            return value, msg_format.format(value, typ)
```
karajan1001, line 179: Actually, I didn't like this design, sometimes this function returns a tuple and sometimes it only returns one value. It will make it hard to use.

skshetry: It won't be used directly, this is a callback function, so the "Prompt" handles this automatically. The other option would be to do `return value, None` which is also iffy.

dberenbaum (approved):
🔥 Looks great @skshetry!

dvc/dependency/param.py
```diff
@@ -37,55 +38,84 @@ class ParamsDependency(Dependency):
...
         return ret
 
     def fill_values(self, values=None):
         """Load params values dynamically."""
-        if not values:
+        if values is None:
             return
+
         info = {}
+        values = values or {}
```
karajan1001, line 61: if `values` was None here, this function had already returned in Line57?

dvc/dependency/param.py
```diff
@@ -37,55 +38,84 @@ class ParamsDependency(Dependency):
...
         ret = super().dumpd()
         if not self.hash_info:
-            ret[self.PARAM_PARAMS] = self.params
+            ret[self.PARAM_PARAMS] = self.params or {}
         return ret
 
     def fill_values(self, values=None):
         """Load params values dynamically."""
-        if not values:
+        if values is None:
```
karajan1001, line 57: Also, not quite understand here, if `values` is empty, only one thing will happen:
```python
self.hash_info = HashInfo(self.PARAM_PARAMS, info)
```

dvc/dependency/param.py
```diff
@@ -37,55 +38,84 @@ class ParamsDependency(Dependency):
...
 
-        if status.get(str(self)) == "deleted":
-            return status
+        if not self.params:
+            return config
+
+        ret = {}
+        if flatten:
+            for param in self.params:
+                try:
```
karajan1001, line 83: Just curious, `dpath.util.get` and `get_in` in `funcy`, which is better?

dvc/dependency/param.py
skshetry: Thanks, I have fixed this in #7432, which will merged first and this will be rebased on top.

dvc/dependency/param.py
skshetry: Basically, what we are trying to do is preserve the state between having a no lock entry for the params file vs empty:
```
params:
  params.yaml: {}

# vs just:
params:
```

This is reflected by `self.hash_info.value` being `None` vs `{}` (empty dict).

Issue #7434, opened 2022-03-04
exp show: throws an error with `-n -1`


karajan1001:
`dvc exp show -v -n -1` throws an error for me:

```
2022-01-25 12:25:40,695 DEBUG: Adding '/private/tmp/example-get-started/.dvc/config.local' to gitignore file.
2022-01-25 12:25:40,697 DEBUG: Adding '/private/tmp/example-get-started/.dvc/tmp' to gitignore file.
2022-01-25 12:25:40,697 DEBUG: Adding '/private/tmp/example-get-started/.dvc/cache' to gitignore file.
2022-01-25 12:25:40,711 ERROR: unexpected error - refs/remotes/origin/HEAD~12: the given reference name 'refs/remotes/origin/HEAD~12' is not valid
------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/dave/Code/dvc/dvc/cli/__init__.py", line 78, in main
    ret = cmd.do_run()
  File "/Users/dave/Code/dvc/dvc/cli/command.py", line 22, in do_run
    return self.run()
  File "/Users/dave/Code/dvc/dvc/commands/experiments/show.py", line 445, in run
    all_experiments = self.repo.experiments.show(
  File "/Users/dave/Code/dvc/dvc/repo/experiments/__init__.py", line 820, in show
    return show(self.repo, *args, **kwargs)
  File "/Users/dave/Code/dvc/dvc/repo/__init__.py", line 48, in wrapper
    return f(repo, *args, **kwargs)
  File "/Users/dave/Code/dvc/dvc/repo/experiments/show.py", line 126, in show
    iter_revs(repo.scm, revs, num, all_branches, all_tags, all_commits)
  File "/Users/dave/Code/dvc/dvc/scm.py", line 165, in iter_revs
    revs.append(resolve_rev(scm, head))
  File "/Users/dave/Code/dvc/dvc/scm.py", line 115, in resolve_rev
    return scm.resolve_rev(rev)
  File "/Users/dave/miniforge3/envs/example-get-started/lib/python3.8/site-packages/scmrepo/git/__init__.py", line 253, in _backend_func
    return func(*args, **kwargs)
  File "/Users/dave/miniforge3/envs/example-get-started/lib/python3.8/site-packages/scmrepo/git/backend/pygit2.py", line 263, in resolve_rev
    shas = {
[...]

PR #7435, opened 2022-03-04
exp show: throws an error with -n -1

karajan1001:
fix: #7434
exp show -n -1 will throw error on any repo with a remote repo setted.

1. Update new version of `scmrepo`
2. Update the unit test add a remote repo to the test case.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -75,7 +75,7 @@ install_requires =
     aiohttp-retry>=2.4.5
     diskcache>=5.2.1
     jaraco.windows>=5.7.0; python_version < '3.8' and sys_platform == 'win32'
-    scmrepo==0.0.7
+    scmrepo==0.0.9
 
 [options.extras_require]
 all =
--- a/tests/unit/scm/test_scm.py
+++ b/tests/unit/scm/test_scm.py
@@ -36,6 +36,7 @@ def test_iter_revs(
     assert gen == {rev_root: [rev_root], rev_new: ["new"]}
     gen = iter_revs(scm, ["new"], 2)
     assert gen == {rev_new: ["new"], rev_old: [rev_old]}
+    scm.gitpython.repo.create_remote("upstream", "some_remote")
     gen = iter_revs(scm, ["other"], -1)
     assert gen == {
         rev_other: ["other"],


karajan1001:
Looks like scmrepo is not compatible with current version of DVC. After PR
```
commit 29d834163ca6c5aa60be6f0f68785e3c2f6f2f87 (HEAD)
Author: Ruslan Kuprieiev <kupruser@gmail.com>
Date:   Tue Dec 28 02:49:14 2021 +0200

    fs: detach from local filesystem (#24)

    This PR makes our fs accept paths relative to the root of the repo and using / as a separator.
    Previous behaviour was a legacy from old dvc.

    Pre-requisite for making dvcfs/repofs/etc switch to the same relative repo paths and / separators.
```

Guess it arrise from the relative path and abs path change in recent `DVC` project.

skshetry:
See https://github.com/iterative/dvc/pull/7359#issuecomment-1032671631.

skshetry:
@karajan1001, I don't understand the change here. Do we need this?

karajan1001:
> @karajan1001, I don't understand the change here. Do we need this? Tests pass fine on master with `scmrepo==0.0.10`.

I used to update `scmrepo` to 0.0.9 and change the test environment in it. But for now the `scmrepo` had already been updated to a newer version. The remained line of test environment changing will ensure #7434 would be passed. I

karajan1001:
@dberenbaum so I guess for now, after updating the `scmrepo` to 0.0.10 the original #7434 had already been solved, do we still this tests or maybe we could close #7434 directly.

pmrowla:
If the underlying issue was in scmrepo, #7434 should just be closed with a link to the scmrepo fix/PR (and the appropriate test should have been added in scmrepo)

skshetry:
The fix in https://github.com/iterative/scmrepo/pull/32 was about supporting `~n` references from the title and is tested for that. The description and the change here points to some issue with git repos having remotes set, which is not tested.

karajan1001:
> The fix in [iterative/scmrepo#32](https://github.com/iterative/scmrepo/pull/32) was about supporting `~n` references from the title and is tested for that. The description and the change here points to some issue with git repos having remotes set, which is not tested.

The test here influence the line below 
```python
gen = iter_revs(scm, ["other"], -1)
```
which will test `1,2,3,...n` util receive a `ScmError`. And going back to https://github.com/iterative/dvc/pull/7204#issuecomment-1023057609,
This problem only arises in a repo with a remote repo being set.

pmrowla:
Ok, so the remote being set causes us to manually try `refs/remotes/<name>` on the DVC side. This would lead to us testing trying to resolve relative commits that may not exist in `refs/remotes`  (like `refs/remotes/origin/HEAD~6`).

The actual underlying problem was that trying to resolve relative commits that did not exist would lead to the unhandled pygit exception. This is not actually related to whether or not remotes are set at all, the same behavior would occur if we tried a non-remote ref like `refs/HEAD~999999` where that relative commit does not exist. It's just coincidental that the bug was identified by setting a remote.

So we probably do not actually need this test/change on the DVC side, since we now have an actual test for resolving non-existent commits now on the scmrepo side.

https://github.com/iterative/scmrepo/blob/330bce52c922009e608c6e3dcc16f23e5f09ff18/tests/test_git.py#L582-L583

PR #7441, opened 2022-03-07
exp run: remove unsupported (repro only) arguments

### LINKED ISSUE #6791: exp run: Remove unsupported `dvc repro` options from the help text
# Bug Report

UPDATE: Jump to https://github.com/iterative/dvc/issues/6791#issuecomment-1043363367

## Description

`dvc exp run --help` lists the following options:

```
  -s, --single-item     Reproduce only single data item without recursive dependencies check.
  -m, --metrics         Show metrics after reproduction.
  --dry                 Only print the commands that would be executed without actually executing.
  -i, --interactive     Ask for confirmation before reproducing each stage.
  -p, --pipeline        Reproduce the whole pipeline that the specified targets belong to.
  -P, --all-pipelines   Reproduce all pipelines in the repo.
  -R, --recursive       Reproduce all stages in the specified directory.
  --no-run-cache        Execute stage commands even if they have already been run with the same command/dependencies/outputs/etc
                        before.
  --force-downstream    Reproduce all descendants of a changed stage even if their direct dependencies didn't change.
  --no-commit           Don't put files/directories into cache.
  --downstream          Start from the specified stages when reproducing pipelines.
  --pull                Try automatically pulling missing cache for outputs restored from the run-cache.
  --glob                Allows targets containing shell-style wildcards.
```

Some of them are not supported. I tested a few of them and testing the rest. Could you remove those from the help text? We're going to update the command reference according to this. 

- [x] `--force`: Seems to run
- [x] `--single-item`: I didn't get what does this do in `dvc exp run` context
- [x] `--metrics` : ✅
- [x] `--interactive`: ✅
- [x] `--pipeline`: ✅
- [x] `--all-pipelines`: seems to run ✅
- [x] `--recursive`: ✅
- [x] `--no-run-cache`: seems to have no effect ❌
- [x] `--force-downstream`: 
- [x] `--no-commit`: seems to have no effect ❌
- [x] `--downstream`:
- [x] `--pull`:
- [x] `--glob`: ❌

pmrowla:

Will close #6791


daavoo (commented):
Need to also update https://github.com/iterative/dvc/blob/5249207ceb9f78755e3b22b9ca07f20f5caa56a2/dvc/commands/repro.py#L31-L46

daavoo (approved):

Issue #7444, opened 2022-03-08
exp show/list: help text for -n


dberenbaum:
From https://github.com/iterative/dvc.org/pull/3187#discussion_r821382100.

@karajan1001 @jorgeorpinel What is the suggested text now? I'm a bit lost with all the updates.

dberenbaum:
@jorgeorpinel Do you think the "first-parent" stuff in the description is too much detail/too technical for the CLI help? Also, do you think we need to reference Git somewhere in the docs when mentioning it?

No need to block any PR for these questions. We can always tweak language in future PRs.

dberenbaum:
Also should include `remove`, `push`, and `pull`.

jorgeorpinel:
> What is the suggested text now? 

In docs we went with your [last suggestion](https://github.com/iterative/dvc.org/pull/3187#pullrequestreview-903319359):

```md
- `-n <num>`, `--num <num>` - show experiments from the last `num` commits
  (first parents) starting from the `--rev` baseline. Give a negative value to include
  all first-parent commits (similar to `git log -n`).
```

> Do you think the "first-parent" stuff in the description is too much detail

I think it's important for the `-n -x` case because the tree can be long and not all parents are included. Unless you phrase it differently e.g. "negative number goes back to the git history root". Up to you.

PR #7450, opened 2022-03-10
exp show/pull/push/list: help text for -n (#7444)

karajan1001:
fix: #7444

from the discussion https://github.com/iterative/dvc.org/pull/3187#discussion_r821382100

1. Update the help message of `-n`

### DIFF
--- a/dvc/commands/experiments/__init__.py
+++ b/dvc/commands/experiments/__init__.py
@@ -93,8 +93,9 @@ def add_rev_selection_flags(
         dest="num",
         metavar="<num>",
         help=(
-            f"{command} experiments from the `--rev` baseline and from `num` "
-            "commits before it (first parents). Give a negative value to "
-            "include all first-parent commits."
+            f"{command} experiments from the last `num` commits "
+            "(first parents) starting from the `--rev` baseline. "
+            "Give a negative value to include all first-parent commits "
+            "(similar to `git log -n`)."
         ),
     )


pmrowla (approved):

dvc/commands/experiments/__init__.py
```diff
@@ -93,8 +93,9 @@ def add_rev_selection_flags(
         dest="num",
         metavar="<num>",
         help=(
-            f"{command} experiments from the `--rev` baseline and from `num` "
-            "commits before it (first parents). Give a negative value to "
-            "include all first-parent commits."
+            f"{command} experiments from the last `num` commits "
+            "(first parents) starting from the `--rev` baseline. "
+            "Give a negative value to include all first-parent commits "
+            "(similar to `git log -n`)."
```
pmrowla, line 99: I don't think we actually need to mention `git log -n` here

karajan1001: It is OK for the help text to be different between the cli and the website?

pmrowla: Since it's already been added to the other docs it's probably fine to leave it in. I'm not sure that it really needs to be mentioned in the docs either, but it's not that big of a deal either way.

dberenbaum: @karajan1001 For future reference, it is intended that there can be differences between the CLI and the website. Often, it makes sense to have more detail in the website than in the help text. That's what I was suggesting in https://github.com/iterative/dvc/issues/7444#issuecomment-1062923980, and I think the suggestion from @pmrowla makes sense as well (especially because `git log -n` with negative numbers is undocumented and there are behavioral differences between them since dvc only shows first-parent ancestors).

Not a big deal that requires another PR, but just so you know they don't have to match. cc @jorgeorpinel

PR #7460, opened 2022-03-14
fs.http: prevent hangs under some network conditions

### LINKED ISSUE #7414: pull: freezes on large datasets
When running `dvc pull` on large datasets, dvc *sometimes* hangs and never returns, here's an example of traceback (after interrupting *twice* with `KeyboardInterrupt`)

```
2022-02-22 19:56:00,887 ERROR: interrupted by the user
  File "dvc/dvc/data/transfer.py", line 176, in transfer
    _do_transfer(
  File "dvc/dvc/data/transfer.py", line 112, in _do_transfer
    total_fails += sum(processor(all_file_ids))
  File "dvc/dvc/utils/threadpool.py", line 29, in imap_unordered
    done, tasks = futures.wait(
  File "versions/3.9.10/lib/python3.9/concurrent/futures/_base.py", line 307, in wait
    waiter.event.wait(timeout)
  File "versions/3.9.10/lib/python3.9/threading.py", line 574, in wait
    signaled = self._cond.wait(timeout)
  File "versions/3.9.10/lib/python3.9/threading.py", line 312, in wait
    waiter.acquire()
KeyboardInterrupt

During handling of the above exception, another exception occurred:

  File "dvc/dvc/cli/__init__.py", line 78, in main
    ret = cmd.do_run()
  File "dvc/dvc/cli/command.py", line 22, in do_run
    return self.run()
  File "dvc/dvc/commands/data_sync.py", line 31, in run
    stats = self.repo.pull(
  File "dvc/dvc/repo/__init__.py", line 48, in wrapper
  File "dvc/dvc/repo/pull.py", line 34, in pull
    processed_files_count = self.fetch(
  File "dvc/dvc/repo/__init__.py", line 48, in wrapper
  File "dvc/dvc/repo/fetch.py", line 86, in fetch
    d, f = _fetch(
  File "dvc/dvc/repo/fetch.py", line 106, in _fetch
    downloaded += repo.cloud.pull(obj_ids, **kwargs)
  File "dvc/dvc/data_cloud.py", line 114, in pull
    return transfer(
  File "dvc/dvc/data/transfer.py", line 176, in transfer
    _do_transfer(
  File "versions/3.9.10/lib/python3.9/concurrent/futures/_base.py", line 637, in __exit__
    self.shutdown(wait=True)
  File "versions/3.9.10/lib/python3.9/concurrent/futures/thread.py", line 235, in shutdown
    t.join()
  File "versions/3.9.10/lib/python3.9/threading.py", line 1053, in join
    self._wait_for_tstate_lock()
  File "versions/3.9.10/lib/python3.9/threading.py", line 1073, in _wait_for_tstate_lock
    if lock.acquire(block, timeout):
KeyboardInterrupt
```

`$ dvc doctor`
> DVC version: 2.9.5 (pip)
> ---------------------------------
> Platform: Python 3.9.10 on macOS-12.1-arm64-arm-64bit
> Supports:
>         azure (adlfs = 2022.2.0, knack = 0.9.0, azure-identity = 1.7.1),
>         gdrive (pydrive2 = 1.10.0),
>         gs (gcsfs = 2022.1.0),
>         hdfs (fsspec = 2022.1.0, pyarrow = 7.0.0),
>         webhdfs (fsspec = 2022.1.0),
>         http (aiohttp = 3.8.1, aiohttp-retry = 2.4.6),
>         https (aiohttp = 3.8.1, aiohttp-retry = 2.4.6),
>         s3 (s3fs = 2022.1.0, boto3 = 1.20.24),
>         ssh (sshfs = 2021.11.2),
>         oss (ossfs = 2021.8.0),
>         webdav (webdav4 = 0.9.4),
>         webdavs (webdav4 = 0.9.4)
> Cache types: reflink, hardlink, symlink
> Cache directory: apfs on /dev/disk3s1s1
> Caches: local
> Remotes: s3, https
> Workspace directory: apfs on /dev/disk3s1s1
> Repo: dvc, git

To reproduce:
```bash
mkdir pull_freeze
pushd pull_freeze

git clone https://github.com/iterative/dvc-bench
cd dvc-bench
python -m venv .venv
./venv/bin/pip install -r requirements.txt dvc
dvc pull # hangs forever
```

See also https://github.com/iterative/dvc-bench/issues/319

dtrifiro:
When using an https remote (such as in the dvc-bench remote, see https://github.com/iterative/dvc-bench/issues/319), under certain situations, `dvc pull` would freeze and never return. Note that this did not happen 100% of the time, but often enough to make the bench CI always fail (note: CI was running a matrix strategy over ~12 workers, pulling a 25k file dataset on each).

Investigation pointed to network issues (as the issue could not be reproduced locally). I ended up tracking this to a connection being dropped/lost and `aiohttp` not realizing this. Due to our timeout defaults for `aiohttp.ClientSession`:
```python
aiohttp.ClientTimeout(
            total=None,
            connect=self.REQUEST_TIMEOUT,
            sock_connect=self.REQUEST_TIMEOUT,
            sock_read=None,
        )
```
(note that `sock_read=None`), `aiohttp` would keep trying to read from a dead socket thus hang.

By instantiating a `aiohttp.TCPConnector` with `enable_cleanup_closed=True`, the underlying transport is forcefully closed when the connection is lost, so that the request fails and `RetryClient` retries. 

Thanks @pared and @karajan1001 

fixes #7414

### DIFF
--- a/dvc/fs/http.py
+++ b/dvc/fs/http.py
@@ -123,7 +123,7 @@ class HTTPFileSystem(NoDirectoriesMixin, FSSpecWrapper):
             total=None,
             connect=self.REQUEST_TIMEOUT,
             sock_connect=self.REQUEST_TIMEOUT,
-            sock_read=None,
+            sock_read=self.REQUEST_TIMEOUT,
         )
 
         return RetryClient(**kwargs)
--- a/dvc/fs/http.py
+++ b/dvc/fs/http.py
@@ -81,11 +81,20 @@ class HTTPFileSystem(NoDirectoriesMixin, FSSpecWrapper):
                     f"Auth method {auth_method!r} is not supported."
                 )
 
+        # Force cleanup of closed SSL transports.
+        # https://github.com/iterative/dvc/issues/7414
+        connector_kwargs = {"enable_cleanup_closed": True}
+
         if "ssl_verify" in config:
-            with fsspec_loop():
-                client_kwargs["connector"] = aiohttp.TCPConnector(
-                    ssl=make_context(config["ssl_verify"])
-                )
+            connector_kwargs.update(ssl=make_context(config["ssl_verify"]))
+
+        with fsspec_loop():
+            client_kwargs["connector"] = aiohttp.TCPConnector(
+                **connector_kwargs
+            )
+        # The connector should not be owned by aiohttp.ClientSession since
+        # it is closed by fsspec (HTTPFileSystem.close_session)
+        client_kwargs["connector_owner"] = False
 
         # Allow reading proxy configurations from the environment.
         client_kwargs["trust_env"] = True


dvc/fs/http.py
```diff
@@ -81,11 +81,17 @@ def _prepare_credentials(self, **config):
...
                 client_kwargs["connector"] = aiohttp.TCPConnector(
-                    ssl=make_context(config["ssl_verify"])
+                    enable_cleanup_closed=True,
+                    ssl=make_context(config["ssl_verify"]),
                 )
+            else:
+                client_kwargs["connector"] = aiohttp.TCPConnector(
+                    enable_cleanup_closed=True
+                )
+            client_kwargs["connector_owner"] = False
```
dtrifiro, line 94: Note that the connector is closed by `fsspec.implementations.HTTPFilesystem.close_session`

dvc/fs/http.py
```diff
@@ -114,7 +120,7 @@ async def get_client(self, **kwargs):
             total=None,
             connect=self.REQUEST_TIMEOUT,
             sock_connect=self.REQUEST_TIMEOUT,
-            sock_read=None,
+            sock_read=self.REQUEST_TIMEOUT,
```
dtrifiro, line 123: I'm assuming that setting the `sock_read=self.REQUEST_TIMEOUT` (60s) will not create any issues: the default buffer size is 2**16,  

See `ClientSession._request`:
https://github.com/aio-libs/aiohttp/blob/f5ff95efe278c470a2ff65cabbb5f5f08ba07416/aiohttp/client.py#L511-L519

https://github.com/aio-libs/aiohttp/blob/f5ff95efe278c470a2ff65cabbb5f5f08ba07416/aiohttp/client_proto.py#L17

https://github.com/aio-libs/aiohttp/blob/f5ff95efe278c470a2ff65cabbb5f5f08ba07416/aiohttp/streams.py#L111

dtrifiro:
For the record, here's a few examples of failing CI (timeout) on `github.com/iterative/dvc-bench/actions`

- https://github.com/iterative/dvc-bench/actions/runs/1977924560
- https://github.com/iterative/dvc-bench/actions/runs/1861766336
- https://github.com/iterative/dvc-bench/actions/runs/1726203498 (one of the first builds that started failing ~2 months ago

Here's a few example of succeeding CI (running this branch):
- https://github.com/iterative/dvc-bench/actions/runs/1980656521
- https://github.com/iterative/dvc-bench/actions/runs/1980308252
- https://github.com/iterative/dvc-bench/actions/runs/1986175157

pared (approved):
Amazing research!

efiop:
> [pared](https://github.com/pared) left a comment
Amazing research!

+1, amazing work 🔥

pmrowla (approved):

dvc/fs/http.py
```diff
@@ -81,11 +81,17 @@ def _prepare_credentials(self, **config):
...
+            if "ssl_verify" in config:
                 client_kwargs["connector"] = aiohttp.TCPConnector(
-                    ssl=make_context(config["ssl_verify"])
+                    enable_cleanup_closed=True,
+                    ssl=make_context(config["ssl_verify"]),
                 )
+            else:
+                client_kwargs["connector"] = aiohttp.TCPConnector(
+                    enable_cleanup_closed=True
+                )
```
skshetry, line 93: Is there more detail on this? I'd like to see why this is not the default in `aiohttp`.

skshetry: Looks like aiohttp used to do this by default 5 years back but was changed by this commit https://github.com/aio-libs/aiohttp/commit/02b0951b1739048b8a8b80a2b406e620e5e17dce as an option. But I could not find any reasoning.

dberenbaum: From https://docs.aiohttp.org/en/stable/client_reference.html#tcpconnector:

> Some ssl servers do not properly complete SSL shutdown process, in that case asyncio leaks SSL connections. If this parameter is set to True, aiohttp additionally aborts underlining transport after 2 seconds. It is off by default.

From the description of how it works, it seems more reasonable that this isn't the default.

dtrifiro: I was too wondering why this is not the default, but digging around could not find any reasoning either. I guess this only becomes an issue under particular network conditions (broken TLS connection?), which were often encountered in dvc bench due to our large number of concurrent requests in dvc bench.

dtrifiro:
Added some context comments with the last force push

dvc/fs/http.py
dtrifiro: Somehow relevant https://github.com/aio-libs/aiohttp/pull/5406

dberenbaum: It looks like this could be the context for the change: https://github.com/aio-libs/aiohttp/issues/1767.

dberenbaum: I guess we could also ask 😄

dvc/fs/http.py
```diff
@@ -114,7 +123,7 @@ async def get_client(self, **kwargs):
             total=None,
             connect=self.REQUEST_TIMEOUT,
             sock_connect=self.REQUEST_TIMEOUT,
-            sock_read=None,
+            sock_read=self.REQUEST_TIMEOUT,
```
skshetry, line 126: I am curious to see if only this will solve the problem in `dvc-bench`, and a bit hesitant on adding `TCPConnector`
without trying out this solution first. Would you mind opening another PR with just this line change? We can merge that, and we can keep this PR open for a 1-2 days to monitor.

dtrifiro: I've tried, it solves the issues on my machine :tm: (MacOS) but it still freezes on the `dvc-bench` ci

skshetry (approved):

dtrifiro:
Gonna test out a few extra scenarios before merging this

Issue #7253, opened 2022-01-11
toml: being 1.0.0 compliant, preserving style on dump


skshetry:
We support toml as params file. There are a few issues with our current toml parsing:
1. We use `toml` library, which is not toml 1.0 standard compatible. Also, [the library is not being actively maintained]( https://github.com/uiri/toml/issues/361).
2. `toml` dumping does not preserve existing formatting.
3. https://github.com/iterative/dvc/issues/6402

We can try migrating to `tomli` (which is being proposed to be included in the standard library, see [PEP 680](https://www.python.org/dev/peps/pep-0680)), which is minimalistic and does not support `dumps`. `tomlkit` does support dumps and preserve styles and is 1.0.0 compliant. As we don't really care about performance for toml at this time, we may as well just use `tomlkit` for both.

TOML support in dvc was introduced in https://github.com/iterative/dvc/pull/4258.

PR #7468, opened 2022-03-16
Tomlkit

syntapy:

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it [here](https://github.com/iterative/dvc.org/pull/3361).

Fixes https://github.com/iterative/dvc/issues/7253


syntapy:
Ok, I think I need to change the way I'm doing the mock toml file

skshetry:
Hey @syntapy, thank you so much for contributing.

> Still maintains functionality to use legacy toml lib in case they pass a decoder to the parser

We don't need to maintain a legacy toml library. We don't directly depend on it, and only expect:
-  `parse_toml` to return us a dictionary
-  `parse_toml_for_update` to return a dictionary that when later dumped via `_dump` preserves formatting/comments.
- `_dump` that can serialize a dictionary to stream as a text
- `modify_toml` that provides a contextmanager that can update toml files and preserve formatting/comments. If `parse_toml_for_update` and `_dump` preserves formatting, this should automatically work.


Regarding tests, we probably don't need anything other than that the ordering/formatting/comments are preserved on `update`.

syntapy:
So apparently when I use Tomlkit, resulting dictionaries are encapsulated in a Tomlkit type. I had to use [this function](https://github.com/sdispater/tomlkit/issues/43#issuecomment-660415820) to convert to make a test pass.

And its not just the dict-like object that has tomlkit values, but also core values like integers are encapsulated into tomlkit types (e.g. `tomlkit.items.Integer`)

I also wrote a test for that but I'm getting trouble from the pre-commit check since it wont let me use `assert type(x) == int`

It wants me to use `isinstance(x, int)` but that doesn't do what I need. If `x` has a `tomlkit.items.Integer` type, then `isinstance(x, int)` will give true. But if we have a `dict` that contains e.g. a `tomlkit.items.Integer` type value, then we get prolems with the ruamel library

syntapy:
Hi @skshetry thank you for the feedback earlier. I have an update but I think I ran into a snag with tomlkit. Could you tell me what you think?

So apparently when I use Tomlkit, the resulting dictionaries are encapsulated in a tomlkit `Table` type, as well as values like integers and floats being encapsulated as `tomlkit.items.Integer` and `tomlkit.items.Float` types respectively.

That results in errors in a 3rd party library (ruamel to be specific)

I had to use [this function](https://github.com/sdispater/tomlkit/issues/43#issuecomment-660415820) to convert to basic python types to make tests pass

I also wrote a test for that converter, but I'm getting trouble from the pre-commit check since it wont let me use things like `assert type(x) == int`

It wants me to use `assert isinstance(x, int)` instead, but that doesn't do what I need. If `x` has a `tomlkit.items.Integer` type, then `isinstance(x, int)` will give true anyways. The purpose of the test is to make sure that the values are converted to base python types

So maybe it will be easier to use another toml library like `tomli` mentioned in the issue instead of working around this

But maybe there is a setting in `ruamel` I can find or look for otherwise

syntapy:
@skshetry Ok, I think I'm ready to have this looked at. I still want to add a test for the format preservation of `modify_toml` but if you wan to look at it now I think I've basically got it done

I dont know whats going on with the linter error though

And when I run tests locally, they pass even though theres some tracebacks being shown

As one example from running `python -m tests -s --pdb tests/func/params/test_show.py` I get

```
  File "/dvc/dvc/fs/utils.py", line 28, in _link
    func(from_path, to_path)
  File "/dvc/dvc/fs/local.py", line 145, in reflink
    System.reflink(from_info, to_info)
  File "/dvc/dvc/system.py", line 112, in reflink
    System._reflink_linux(source, link_name)
  File "/dvc/dvc/system.py", line 96, in _reflink_linux
    fcntl.ioctl(d.fileno(), FICLONE, s.fileno())
OSError: [Errno 95] Operation not supported

The above exception was the direct cause of the following exception:
```

syntapy:
Another traceback from running that local command in the above comment is

```
  File "/dvc/.venv/lib/python3.10/site-packages/funcy/flow.py", line 112, in reraise
    yield
  File "/dvc/dvc/utils/serialize/_yaml.py", line 29, in parse_yaml
    return yaml.load(text) or {}
  File "/dvc/.venv/lib/python3.10/site-packages/ruamel/yaml/main.py", line 434, in load
    return constructor.get_single_data()
  File "/dvc/.venv/lib/python3.10/site-packages/ruamel/yaml/constructor.py", line 119, in get_single_data
    node = self.composer.get_single_node()
  File "_ruamel_yaml.pyx", line 706, in _ruamel_yaml.CParser.get_single_node
  File "_ruamel_yaml.pyx", line 724, in _ruamel_yaml.CParser._compose_document
  File "_ruamel_yaml.pyx", line 775, in _ruamel_yaml.CParser._compose_node
  File "_ruamel_yaml.pyx", line 891, in _ruamel_yaml.CParser._compose_mapping_node
  File "_ruamel_yaml.pyx", line 904, in _ruamel_yaml.CParser._parse_next_event
ruamel.yaml.scanner.ScannerError: while scanning a simple key
  in "<unicode string>", line 8, column 1
could not find expected ':'
  in "<unicode string>", line 9, column 1

The above exception was the direct cause of the following exception:

  File "/dvc/dvc/utils/__init__.py", line 490, in wrapper
    vals = func(*args, **kwargs)
  File "/dvc/dvc/repo/params/show.py", line 135, in _gather_params
    param_outs, params_fs_paths = _collect_configs(repo, rev, targets=targets)
  File "/dvc/dvc/repo/params/show.py", line 31, in _collect_configs
    params, fs_paths = collect(
  File "/dvc/dvc/repo/collect.py", line 81, in collect
    outs: Outputs = _collect_outs(repo, output_filter=output_filter, deps=deps)
  File "/dvc/dvc/repo/collect.py", line 24, in _collect_outs
    index.check_graph()  # ensure graph is correct
  File "/dvc/dvc/repo/index.py", line 268, in check_graph
    self.graph  # pylint: disable=pointless-statement
  File "/dvc/.venv/lib/python3.10/site-packages/funcy/objects.py", line 28, in __get__
  File "/dvc/dvc/repo/index.py", line 155, in graph
    return build_graph(self.stages, self.outs_trie)
  File "/dvc/.venv/lib/python3.10/site-packages/funcy/objects.py", line 28, in __get__
  File "/dvc/dvc/repo/index.py", line 74, in stages
    return self.stage_collector.collect_repo(onerror=onerror)
  File "/dvc/dvc/repo/stage.py", line 497, in collect_repo
    return list(self._collect_repo(onerror))
  File "/dvc/dvc/repo/stage.py", line 480, in _collect_repo
    new_stages = self.load_file(file_path)
  File "/dvc/dvc/repo/stage.py", line 304, in load_file
    return self.load_all(path)
  File "/dvc/dvc/repo/stage.py", line 276, in load_all
    stages = dvcfile.stages  # type: ignore
  File "/dvc/dvc/dvcfile.py", line 281, in stages
    data, _ = self._load()
  File "/dvc/dvc/dvcfile.py", line 150, in _load
    return self._load_yaml(**kwargs)
  File "/dvc/dvc/dvcfile.py", line 161, in _load_yaml
    return strictyaml.load(
  File "/dvc/dvc/utils/strictyaml.py", line 282, in load
    raise YAMLSyntaxError(path, text, exc) from cause
dvc.utils.strictyaml.YAMLSyntaxError: unable to read: 'dvc.yaml', YAML file structure is corrupted
```

syntapy:
Oh ok, those are the linter error messages. I tried running pre-commit locally on main branch and now I see

.gitignore
```diff
@@ -5,6 +5,7 @@ neatlynx/__pycache__
 *.pyc
 .env/
 .env2.7/
+.venv*/
```
pmrowla, line 8: This is specific to your dev environment, not sure this change should go in this PR

dvc/utils/serialize/_toml.py
```diff
@@ -1,9 +1,46 @@
...
+logger = logging.getLogger(__name__)
+
+
+def _tomlkit_to_popo(d):
+    """Converts tomlkit objects to plain old python objects (popo)
+    We could probably just do dict(tmlkit_item.items()) but this
+    seems more robust since it handles lists
+
+    Taken from https://github.com/sdispater/tomlkit/issues/43
+    """
```
pmrowla, line 17: Feels like this is really something that belongs in an upstream PR for tomlkit and not DVC

syntapy:
@dberenbaum I think I may have finished the last aspects just now. If so then I believe this PR should be mostly done

dvc/utils/serialize/_toml.py
syntapy: I have this part done now. I am working on another aspect now

syntapy:
As per the above comment, currently waiting for https://github.com/sdispater/tomlkit/pull/201, which fixes a problem I introduced

dberenbaum:
Thanks @syntapy! Looks like this should be ready now (see https://github.com/sdispater/tomlkit/releases/tag/0.11.1)?

syntapy:
I added that release in the `setup.cfg` file just the other day. Is there anywhere else I need to do?

dberenbaum:
🤦 Sorry @syntapy, I missed your updates somehow.

pmrowla (commented):
Changes look alright to me, but all of these commits need to be squashed before we can merge

@skshetry can you take another look at the current PR

tests/func/params/test_show.py
```diff
@@ -44,6 +44,23 @@ def test_show_toml(tmp_dir, dvc):
     }
 
 
+def test_preserve_comments(tmp_dir, dvc):
```
skshetry, line 47: Can you create a new test file `tests/unit/utils/serialize/test_toml.py` and move all of these 3 tests there, please? They are not related to `test_show` or `test_params`, but lower level utilities.

Regarding the test, please try using (preferably) `modify_toml()` or `dump_toml` instead of `_dump`.

dvc/utils/serialize/_toml.py
```diff
@@ -1,9 +1,12 @@
+import logging
 from contextlib import contextmanager
 
 from funcy import reraise
 
 from ._common import ParseError, _dump_data, _load_data, _modify_data
 
+logger = logging.getLogger(__name__)
```
skshetry, line 8: This is unused, let's get rid of this? (I assume you were only using it for debugging).

setup.cfg
```diff
@@ -43,6 +43,7 @@ install_requires =
     appdirs>=1.4.3
     ruamel.yaml>=0.17.11
     toml>=0.10.1
```
skshetry, line 45: Are we using this anywhere? Let's get rid of this if we can. 
```suggestion
```

skshetry (approved):
Overall it looks great, I have a few cosmetic suggestions. 🔥 

Also thank you so much for all the upstream patches, and for helping us keep implementation simpler.  Also, apologies for the late reviews. :)

tests/func/params/test_show.py
syntapy: I'm struggling to get `modify_toml()` working

skshetry: yeah, that's a bug in `modify_data`, which returns an empty dict if the file does not exist.

https://github.com/iterative/dvc/blob/634894d3b1205f0006a8404d9914e7916a877355/dvc/utils/serialize/_common.py#L87

We can ignore this, as `modify_toml` is about updating contents (raw/dict) to existing content (special dict). It's not the other way around. You can simply test as follows:
```python
def test_preserve_comments(tmp_dir):
    from dvc.utils.serialize import _toml

    contents_fmt = """\
#A Title
[foo]
bar = {} # meaning of life
baz = [1, 2]
"""
    tmp_dir.gen("params.toml", contents_fmt.format("42"))

    with _toml.modify_toml("params.toml") as d:
        d["foo"]["bar"] //= 2
    assert (tmp_dir / "params.toml").read_text() == contents_fmt.format("21")

```

daavoo: @syntapy Not sure how you are trying to test `modify_toml`. It works as expected for me when trying in how I believe should be used:

```
$ cat foo.toml 
```

```toml
# A Title 
[foo]
bar = 42# meaning of life
baz = [1, 2]
```

```python
$ python
>>> from dvc.utils.serialize._toml import modify_toml
>>> with modify_toml("foo.toml") as data:
...     data["foo"]["bar"] = 2
... 
```

```
$ cat foo.toml
```
```toml
# A Title 
[foo]
bar = 2# meaning of life
baz = [1, 2]
```

skshetry: @daavoo, what @syntapy was saying is that it fails when you try to update the content of `foo.toml` to a new file that does not exist yet. That scenario, however is not something that we support, or intend to support with `modify_data`/`modify_toml`.

i.e

```python
from dvc.utils.serialize._toml import modify_toml

with open("foo.toml") as f:
  content = parse_toml_for_update(f.read(), "foo.toml")
with modify_toml("new_file_that_does_not_exist_yet.toml") as d:
  d.update(content)
```

The top level comment is not preserved.

daavoo: > That scenario, however is not something that we support, or intend to support with modify_data/modify_toml.

My point was that the test was not testing what we use `modify_toml` for

skshetry (approved):

PR #7469, opened 2022-03-16
scmrepo: update to 0.0.13

pmrowla:

### DIFF
--- a/dvc/external_repo.py
+++ b/dvc/external_repo.py
@@ -201,12 +201,14 @@ def _clone_default_branch(url, rev, for_write=False):
                     # given repo URL it is easier/safer for us to work with
                     # full clones in this case.
                     logger.debug("erepo: unshallowing clone for '%s'", url)
-                    _unshallow(git)
+                    git.fetch(unshallow=True)
+                    _merge_upstream(git)
                     shallow = False
                     CLONES[url] = (clone_path, shallow)
                 else:
                     logger.debug("erepo: git pull '%s'", url)
-                    git.pull()
+                    git.fetch()
+                    _merge_upstream(git)
         else:
             from dvc.scm import clone
 
@@ -217,10 +219,13 @@ def _clone_default_branch(url, rev, for_write=False):
 
                 try:
                     git = clone(url, clone_path, shallow_branch=rev)
-                    shallow = True
-                    logger.debug(
-                        "erepo: using shallow clone for branch '%s'", rev
+                    shallow = os.path.exists(
+                        os.path.join(clone_path, Git.GIT_DIR, "shallow")
                     )
+                    if shallow:
+                        logger.debug(
+                            "erepo: using shallow clone for branch '%s'", rev
+                        )
                 except CloneError:
                     pass
             if not git:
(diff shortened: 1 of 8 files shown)


setup.cfg
```diff
@@ -75,7 +75,7 @@ install_requires =
     aiohttp-retry>=2.4.5
     diskcache>=5.2.1
     jaraco.windows>=5.7.0; python_version < '3.8' and sys_platform == 'win32'
-    scmrepo==0.0.10
+    scmrepo>=0.0.11
```
skshetry, line 78: I'd prefer this to be pinned so that older DVC versions still work even if we make any incompatible changes to the `scmrepo` in the future.

pmrowla: Yeah, I was actually going to ask if you thought we should start doing something like `scmrepo<0.1.0,>=0.0.11` instead of pinning a specific release, with the assumption that the next time we add some incompatible changes (like the relpath change) in scmrepo we would go ahead and bump the minor version there instead of the patch version.

@skshetry

pmrowla:
This breaks dvc erepo because the default tracking branch isn't set for `main/master` after dulwich clone, will need another update on the scmrepo side first

tests/func/test_external_repo.py
```diff
@@ -128,60 +128,56 @@ def test_relative_remote(erepo_dir, tmp_dir):
...
 
-def test_shallow_clone_branch(erepo_dir):
+def test_shallow_clone_branch(erepo_dir, mocker):
     with erepo_dir.chdir():
         with erepo_dir.branch("branch", new=True):
             erepo_dir.dvc_gen("file", "branch", commit="create file on branch")
         erepo_dir.dvc_gen("file", "master", commit="create file on master")
 
     url = os.fspath(erepo_dir)
+    mock_clone = mocker.patch.object(Git, "clone", wraps=Git.clone)
```
skshetry, line 138: See [`mocker.spy`](https://github.com/pytest-dev/pytest-mock/#spy) instead  of `wraps=`.

dvc/external_repo.py
```diff
@@ -234,17 +239,16 @@ def _clone_default_branch(url, rev, for_write=False):
...
-        branch = git.gitpython.repo.create_head(branch_name, ref)
-        branch.set_tracking_branch(ref)
-        branch.checkout()
-    git.pull(unshallow=True)
+def _merge_upstream(git):
+    from scmrepo.exceptions import SCMError
+
+    try:
+        branch = git.active_branch()
+        upstream = f"refs/remotes/origin/{branch}"
```
skshetry, line 247: Is it okay to hardcode this? IIRC the default remote name is configurable in git.

pmrowla: dulwich clone doesn't actually check `clone.defaultRemoteName` (it only supports overriding it directly with `origin=...` equivalent to `git clone --origin <name>`), so for our purposes it will always be `origin`.

I think this is ok for now, and ocne https://github.com/iterative/scmrepo/issues/41 is addressed (and erepo is moved into scmrepo) we should add the methods needed to check git config vars, and update our dulwich clone wrapper to use `clone.defaultRemoteName` if the user actually has it set.

skshetry (approved):

Issue #6316, opened 2021-07-14
Allow multiple plots from one plot file


mattlbeck:
Consider a mock dvc.yaml such as:

```
stages:
    A:
        cmd: ...
        plots:
            - metrics.csv:
                x: x
                y: y1
            - metrics.csv:
                x: x
                y: y2
```

DVC will only show one (the last) plot. The fix is to split into multiple metrics filer with the same x but different y variables. This feels pretty inconvenient for metrics that are related in their dependent variable, and so can be stored in the same table and file.

On the surface I don't see any reason why a specified plot artifacts can only be used for one plot.

pared:
Related #5980

shcheklein:
Relevant to iterative/studio-support#23.

shcheklein:
@dberenbaum should we/can we prioritize this?

dberenbaum:
See a proposed spec for this in https://github.com/iterative/dvc/discussions/5980#discussioncomment-1026072.

@shcheklein Yup, we will be working on it soon. @pared What do you think? I think we had planned to address #6603 first to get image support looking a little better, but this is probably more important given the number of requests for it.

daavoo:
By "combining" we have been usually referring to get multiple (`y`) columns from either the same file or multiple files, which seems reasonable given how we usually recommend formatting the data in linear plots.

However, I think that the user in https://github.com/iterative/studio-support/issues/23#issuecomment-929932069 showcases a new scenario that we haven't discussed so far in https://github.com/iterative/dvc/discussions/5980#discussioncomment-1026072 . It is basically using the values of a third column (`z`) to group the values of a single column (`y`) and the template looks very clean.

IMO this use case could be easily covered on the DVC side by extending the vega template and adding a new placeholder
`DVC_METRIC_COLOR_LABEL` and we could encourage this format as a quick workaround for combining `y` values. On the DVCLive side, we could add some logic to facilitate the optional usage of this format (i.e. adding a new `stage` column for grouping based on the `metric` name).

daavoo:
I think that #6603 still needs to be addressed first because the `diff` operation for this kind of combined plots would require implementing some sort of facet grid in the HTML layout (looks like Studio would need to implement something similar as well), as @dberenbaum mentioned in https://github.com/iterative/dvc/discussions/5980#discussioncomment-1008543 .

dberenbaum:
[...]

Issue #6150, opened 2021-06-10
plots diff: unnecessary requirement for dvc.yaml to be valid


aguschin:
# Bug Report

## Description

`dvc plots diff --targets metrics/metrics_precision.tsv -- master` fails because of invalid dvc.yaml, though AFAIK no dvc.yaml is really required to plot this (the metrics_precision.tsv file is tracked with Git). This is how this plot looks like when produced with valid dvc.yaml
![Screenshot 2021-06-10 at 14 45 03](https://user-images.githubusercontent.com/6797716/121519502-7365f200-c9fa-11eb-9a7d-8c453b82c851.png)

### Reproduce

```
# metrics_precision.yaml
timestamp	step	metrics_precision
1623321612138	0	0.0
1623321617898	1	0.0
1623321622355	2	0.0
1623321628074	3	0.0563768875880167
1623321632661	4	0.33889920337730767
1623321637196	5	0.3251424198127944
1623321641976	6	0.0010819493589534024
1623321648511	7	0.35421949649508316
1623321661087	8	0.001618569316573259
1623321694687	9	0.24080049696178488

# dvc.yaml
stages:
  train:
    deps:
      - ${weights}
    cmd: python train.py
```

1. create these files in a folder
2. git init
3. git add metrics_precision.tsv
4. git commit -am 'first changes'
5. dvc plots diff -v --targets metrics_precision.tsv -- master

```
$ dvc plots diff --targets metrics/metrics_precision.tsv -- master
ERROR: failed to parse 'stages.train.deps' in 'dvc.yaml': Could not find 'weights'
```

### Expected

Expected to get plots without any error.

### Environment information

**Output of `dvc doctor`:**

```console
$ dvc doctor
DVC version: 2.3.0 (pip)
---------------------------------
Platform: Python 3.9.5 on macOS-10.16-x86_64-i386-64bit
Supports: azure, http, https
Cache types: reflink, hardlink, symlink
Cache directory: apfs on /dev/disk1s5s1
Caches: local
Remotes: azure
Workspace directory: apfs on /dev/disk1s5s1
Repo: dvc, git
```

**Additional Information (if any):**

I've encountered this error when stopped tracking params.yaml with Git. It's not like it's a common situation itself, but it seems strange that we need to look at dvc.yaml here.

```
$ dvc plots diff -v --targets metrics_precision.tsv -- master
[...]

PR #7477, opened 2022-03-18
Initial support for flexible plots

### LINKED ISSUE #7086: plots: flexible dvc.yaml spec
YAML doesn't permit dup keys also even if we change that to arrays we will have the issue of duplication and discrepancy, i.e. one template under one key and another one under the second key. So plot names should be unique string, whichever you write in target I suppose and datafile should go as key. We may still use a key if `datafile` is absent, this will simplify a very common one file case. So:
```yaml
plots:
- train_vs_val:
    x: epoch
    y: loss
    data: [train_loss.csv, val_loss.csv]


- val_f1.csv:
    x: epoch
    y: [f1_class_0, f1_class_1]
    y_label: f1
    # data is absent, using key value: val_f1.csv

# Separately plot since we have TWO plots, even though with the same data file
- scores_acc:
    x: epoch
    y: acc
    data: scores.csv
- scores_auc:
    x: epoch
    y: auc
    data: scores.csv
```

_Originally posted by @Suor in https://github.com/iterative/dvc/discussions/5980#discussioncomment-1026072_

pared:

Closes #7086


pared:
This change might end up the ordering of plots handled in #5009. Writing it down for future reference

dvc/render/vega_converter.py
```diff
@@ -161,23 +177,108 @@ def convert(
...
+
+            properties_update["y"] = "dvc_inferred_y_value"
+
+            return tmp
+
+        if isinstance(y_values, str):
+            for datapoint in datapoints:
+                tmp = datapoint.copy()
+                # TODO What about inferred Y (y_values=None)?
+                tmp[REVISION_FIELD] = f"{revision}::{filename}"
```
daavoo, line 251: Not sure how to deal with this, but having this long strings as `rev` causes the Vega legend unusable:

Captura de Pantalla 2022-05-09 a las 22 09 37

daavoo: Apart from reducing font size (not very useful), I don't see which property could be adjusted:

https://vega.github.io/vega-lite-v1/docs/legend.html

pared: Alternative approach is to make `dvc plots show` edit datapoint inside renderer (just before HTML render, to not mess up with `vscode`) and remove common prefix.

dvc/render/match.py
```diff
@@ -57,3 +64,125 @@ def match_renderers(
...
+                            filename=filename,
+                            data=plot_data,
+                            skip=skip,
+                            config_path=dvc_file,
+                        )
+                        # TODO test skip
+                        plot_datapoints.extend(dps)
+                    renderer_id = plot_id
+        if "title" not in final_props:
+            final_props["title"] = renderer_id
```
dberenbaum, line 184: Similar to stripping redundant info from the legend, it would be nice to strip `dvc.yaml::` or other path info from plots IDs in the title unless there is a conflict. Alternatively, that info could be shown in a subtitle or some other way to make it less distracting.

dberenbaum: When an explicit `title:` is set, should we also include the path somewhere, at least when they is a conflict? Right now you can have plots in two different directories with `title: Accuracy` and there is no way to distinguish them in the html.

pared: Good point, ill look into that

dberenbaum:
I'm getting an error when using `dvc plots show` from within a subdir where the `dvc.yaml` is in the parent dir and the plot file is in the subdir:

```
$ git clone git@github.com:iterative/example-get-started.git
$ cd example-get-started/evaluation
$ dvc plots show -v
2022-05-12 10:14:10,050 ERROR: unexpected error - ['workspace', 'sources', 'data', '../evaluation/importance.png']
  File "/Users/dave/Code/dvc/dvc/cli/__init__.py", line 90, in main
    ret = cmd.do_run()
  File "/Users/dave/Code/dvc/dvc/cli/command.py", line 22, in do_run
    return self.run()
  File "/Users/dave/Code/dvc/dvc/commands/plots.py", line 86, in run
    renderers = match_defs_renderers(
  File "/Users/dave/Code/dvc/dvc/render/match.py", line 147, in match_defs_renderers
    definitions_data = get_definition_data(data, plot_sources, rev)
  File "/Users/dave/Code/dvc/dvc/render/match.py", line 102, in get_definition_data
    file_result = dpath.util.get(data, [rev, "sources", "data", file])
  File "/Users/dave/miniforge3/lib/python3.9/site-packages/dpath/util.py", line 179, in get
    raise KeyError(glob)
KeyError: ['workspace', 'sources', 'data', '../evaluation/importance.png']
2022-05-12 10:14:10,715 DEBUG: Version info for developers:
DVC version: 2.10.3.dev18+gb14a972cd.d20220511
Platform: Python 3.9.2 on macOS-12.3.1-arm64-arm-64bit
Supports:
        azure (adlfs = 2022.2.0, knack = 0.9.0, azure-identity = 1.8.0),
        gdrive (pydrive2 = 1.10.1),
        gs (gcsfs = 2022.2.0),
        hdfs (fsspec = 2022.2.0, pyarrow = 7.0.0),
        webhdfs (fsspec = 2022.2.0),
        s3 (s3fs = 2022.2.0, boto3 = 1.20.24),
        ssh (sshfs = 2021.11.2),
        oss (ossfs = 2021.8.0),
        webdav (webdav4 = 0.9.5),
        webdavs (webdav4 = 0.9.5)
Caches: local
Remotes: https
Repo: dvc, git

2022-05-12 10:14:10,716 DEBUG: Analytics is disabled.
```

dvc/commands/plots.py
```diff
@@ -341,6 +349,12 @@ def _add_props_arguments(parser):
     parser.add_argument(
         "--y-label", default=None, help="Y axis label", metavar="<text>"
     )
+    parser.add_argument(
+        "--from-config",
+        help="Plot configuration file",
+        default=None,
+        metavar="<path>",
+    )
```
dberenbaum, line 357: Is there a need for this at the moment? 

If not, I'd suggest we drop it because the behavior isn't that clear to me. For example, if I do `dvc plots show plots/roc.json --from-config plots.yaml`, what if there are multiple plots that use `plots/roc.json`? If I do `dvc plots diff HEAD --from-config plots.yaml`, does HEAD get configured with `plots.yaml` from the workspace or from HEAD?

dberenbaum:
Should targets include any plot key so that it's possible to do `dvc plots show acc_train_vs_test`?

pared:
> I'm getting an error when using dvc plots show from within a subdir where the dvc.yaml is in the parent dir and the plot file is in the subdir:

Fixed

Note: rebased on lastest main, remember to update `scmrepo`

dvc/commands/plots.py
pared: There is no need, but also there has been some reluctance in introducing `plots` into `dvc.yaml` during our discussions. I would keep that and discuss this among the whole team. We can always deprecate `from-config`. Reverting changes for `dvc.yaml` will not be that easy. I was actually thinking about providing `--from-config` as initial workflow interface, like a playground, to "design" the plots. Once the plots are working as intended, the user could copy-paste them into the `dvc.yaml`

pared:
> Should targets include any plot key so that it's possible to do dvc plots show acc_train_vs_test?

As of today we can address the plots by:
`path` - should work for old type plots (plots output)
`plot_id` - will match any `{plot_id}` within any config file, eg. `plots show plot_id` will target both `dvc.yaml::plot_id` and `subdir/dvc.yaml::plot_id` (assuming they exist)
`config_path::plot_id` - should match only particular `plot_id` within `config_path`- if `config_path` is not `stage` file, `from-config` still needs to be provided

dvc/commands/plots.py
daavoo: Don't want to go in circles about the "dvc.yaml section" vs "external file" discussion but, given how tricky the error handling can get (i.e. https://github.com/iterative/dvc/issues/6150), I would prefer to be able to call `dvc plots --from-config` without getting errors from unrelated pipeline issues.

dberenbaum: > Reverting changes for `dvc.yaml` will not be that easy.

I didn't get what you mean by this. Could you explain?

> Don't want to go in circles about the "dvc.yaml section" vs "external file" discussion but, given how tricky the error handling can get (i.e. #6150), I would prefer to be able to call `dvc plots --from-config` without getting errors from unrelated pipeline issues

Agreed, but it seems out of scope and it doesn't work right now:

```
$ echo 'roc.json:\n  x: val_acc' > roc.json
$ echo '{"train": [{"val_acc": 0}]}' > roc.json
$ echo 'roc.json:\n  x: val_acc' > plots.json
$ dvc plots show --from-config plots.json
file:///Users/dave/repo/dvc_plots/index.html
$ echo invalid > dvc.yaml
$ dvc plots show --from-config plots.json
DVC failed to load some plots for following revisions: 'workspace'.
file:///Users/dave/repo/dvc_plots/index.html
```

The reason I want to exclude it (or hide it) is that it's a separate scenario that requires more testing and time, and I don't want to delay merging flexible plots to get `--from-config` right.

pared: > I didn't get what you mean by this. Could you explain?

Once we allow defining plots in `dvc.yaml`, practically there is no coming back from it before a major release (even if we deprecate it, we still need to allow for it for users that already adopted it). On the other hand I don't have a viable alternative - we don't have a way to make `dvc plots show` use `from-config` files because they are not marked in any way as files containing plots definitions.

We could also think about supporting something like:

```
# dvc.yaml
plots:
   - config_file.yaml
``` 
which would load config_file.yaml. This allows for delegating this responsibility to particular files not related to the pipeline.

dberenbaum: Based on conversation with @pared, we will make `--from-config` hidden for now to not delay implementation.

dvc/commands/plots.py
```diff
@@ -102,6 +109,7 @@ def run(self):
 
             output_file: Path = (Path.cwd() / out).resolve() / "index.html"
 
+            # TODO on no renderers still producing html
```
daavoo, line 112: `if renderers:` ?

daavoo (commented):
@pared please ping when next iteration is finished

dberenbaum:
We discussed this already @pared, but for the record here, `plots` don't work unless the `stages` key is also in `dvc.yaml`.

dberenbaum:
Looking good @pared!

How strict do you think we should be in schema validation? Right now, it seems much looser than for `stages`. You can add extra fields and they get silently ignored. You can even add extra columns to the data and they are silently ignored.

Sometimes, I like that it doesn't fail as aggressively as `stages` and DVC commands that rely on `stages`. On the other hand, if I make a typo, there's nothing to help guide me on how to correct it.

Related but not part of this PR: when `dvc.yaml` is either corrupted or fails validation, the output shows `DVC failed to load some plots for following revisions:

dberenbaum:
Plots keys with spaces aren't working. This would be helpful as a shortcut for titles.

Example:

dvc.yaml
```yaml
plots:
  "train data":
    x: epoch
    y:
      train.csv: [acc,loss]
stages:
  train:
    cmd: train
```
train.csv
```csv
epoch,acc,loss
0,0,1
1,0.2,0.8
2,0.3,0.6
3,0.4,0.4
```

pared:
> Can we support images and plots with no config in `plots` even if they have no properties? Otherwise, I have to end up marking plots in both `plots` and `stages`.

Fixed, user should be able to visualize any file by just providing its (posix-style) path in `plots` section and no config after.
eg:
```yaml
plots:
    path/to/image.jpg:
    path/to/file.csv:
    ...
```

pared:
@daavoo next iteration is done.

dvc/render/match.py
pared: Ok, so in this case, we could relatively simple include that by modyfing `dvc-render`. (It would be backward compatible). I propose to do it after the release.

pared:
> Plots keys with spaces aren't working. This would be helpful as a shortcut for titles.

Handled in: 
https://github.com/iterative/dvc-render/pull/63

dberenbaum:
@pared Is it still in draft status?

dberenbaum:
> We discussed this already @pared, but for the record here, `plots` don't work unless the `stages` key is also in `dvc.yaml`.

What about this issue? Still seems to be a problem for me.

dberenbaum:
This doesn't seem to be working:

```yaml
  acc_vs_loss:
    x: epoch
    y:
      train.csv: [acc, loss]
```

dberenbaum:
Can we clarify what issues this fixes:
* #7086
* #6316
* #7830?

pared:
> What about this issue? Still seems to be a problem for me.

Yes, this is still not working. That will require some changes in how we collect `dvc.yaml` files. I think we should create a separate issue for that, as it touches other (than `plots`) parts of dvc. The current change is humongous as it is. I would prefer to delay fixing this to a separate issue.

pared:
https://github.com/iterative/dvc/pull/7477#issuecomment-1159149773

Fixed

dberenbaum (approved):
@pared Haven't had a chance to test your latest fixes, but LGTM as long as those were addressed. Thanks!

dvc/repo/plots/__init__.py
```diff
@@ -260,6 +302,168 @@ def _collect_plots(
...
+
+    y = config.get("y", None)
+    if isinstance(y, dict):
+        sources = list(y.keys())
+    else:
+        sources = [plot_id]
+
+    # TODO bug in dvc_objects? normpath normalizes to windows path on windows
+    return _deduplicate(
+        posixpath.normpath(pathlib.join(pathlib.dirname(config_path), source))
```
daavoo, line 333: Why is this needed?

dvc/repo/plots/__init__.py
```diff
@@ -260,6 +302,168 @@ def _collect_plots(
...
+        for source in sources
+    )
```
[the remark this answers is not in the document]
pared: Set can affect ordering of datapoints. This change in plots data might affect the assignment of colors in plots. I think it is a nice feature to keep them consistent. Not required though.

dvc/repo/plots/__init__.py
pared: DvcFileSystem uses `posixpath` style paths, regardless the fs its on. For some reason, `join` and `dirname` works on Windows with the posixpaths, but the `normpath` normalizes into Windows style path, which results in problems of evaluating data sources and matching the renderers down the road. It seems like inconsistent behavior, hence TODO to investigate.

dvc/repo/plots/__init__.py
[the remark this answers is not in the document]
daavoo:
```suggestion
    y = config.get("y", None)
    if isinstance(y, dict):
        sources = list(y.keys())
    else:
        sources = [plot_id]

    # TODO bug in dvc_objects? normpath normalizes to windows path on windows
    return list(dict.from_keys((
        posixpath.normpath(pathlib.join(pathlib.dirname(config_path), source))
        for source in sources
    ))
```

dvc/commands/plots.py
```diff
@@ -74,51 +147,44 @@ def run(self):
             renderers_out = (
                 out if self.args.json else os.path.join(out, "static")
             )
-            renderers = match_renderers(
-                plots_data=plots_data,
+
+            renderers = match_defs_renderers(
+                pathlib=self.repo.fs.path,
```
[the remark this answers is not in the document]
pared: When invoking `show/diff` from different dir than `root` of the project we need to evaluate data source relative path (to the cwd) to relative path of data (which is relative  to `dvc.yaml` where plot is defined).

dvc/commands/plots.py
```diff
@@ -24,6 +24,60 @@ def _show_json(renderers, split=False):
...
+def _filter_unhandled_renderers(renderers):
+    # filtering out renderers currently unhandled by vscode extension
+    from dvc_render import VegaRenderer
+
+    def _is_json_viable(r):
+        return not (
+            isinstance(r, VegaRenderer) and _data_versions_count(r) > 1
+        )
+
+    return list(filter(_is_json_viable, renderers))
```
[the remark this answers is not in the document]
pared: I agree, we could do that. Current solution is to make sure `vscode` doesn't get `multi_source` type of plots. I hope we won't need this code in few weeks, once we make extension handle them.

dvc/commands/plots.py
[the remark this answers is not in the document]
daavoo: I guess my question is why it's required as part of the renderer matching step? 
Is it possible to resolve that at collection time?

dvc/commands/plots.py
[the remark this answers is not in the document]
daavoo: >  I hope we won't need this code in few weeks

Got it. Feel free to dismiss

dvc/commands/plots.py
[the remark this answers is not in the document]
pared: Huh, good point, we shouldn't need that here if we resolved everything during collection time. This is because I keep the plots definitions unchanged (as they are in particular dvc.yaml), I should probably reconsider that.

dvc/repo/plots/__init__.py
```diff
@@ -260,6 +302,168 @@ def _collect_plots(
...
+
+def _collect_pipeline_files(repo, targets: List[str], props):
+    from dvc.dvcfile import PipelineFile
+
+    result: Dict[str, Dict] = {}
+    dvcfiles = {stage.dvcfile for stage in repo.index.stages}
+    for dvcfile in dvcfiles:
+        if isinstance(dvcfile, PipelineFile):
+            dvcfile_defs = dvcfile.load().get("plots", {})
+            rpath = _dvcfile_relpath(repo, dvcfile)
```
daavoo, line 397: Is there a difference if we use:

```suggestion
            rpath = dvcfile.relpath
```

dvc/repo/plots/__init__.py
```diff
@@ -260,6 +302,168 @@ def _collect_plots(
...
+        fs.path.join("/", fs.from_os_path(dvcfile.path)), fs.path.getcwd()
+    )
+
+
+def _collect_output_plots(
+    repo, targets, props, onerror: Optional[Callable] = None
+):
+    fs = repo.dvcfs
+    result: Dict[str, Dict] = {}
+    for plot in repo.index.plots:
```
daavoo, line 369: Not blocker of this P.R.

I think flexible plots should be included as part of the `repo.index` somehow

dvc/render/converter/vega.py
```diff
@@ -0,0 +1,316 @@
...
+            tmp[REVISION_FIELD] = version_info["revision"]
+            return tmp
+
+        def _version_info(revision, filename=None, field=None):
+            res = {"revision": revision}
+            if filename is not None:
+                res["filename"] = filename
+            if field is not None:
+                res["field"] = field
+            return res
```
daavoo, line 235: @pared I don't fully understand the motivation for adding the `VERSION_FIELD` and the structure used. Could you clarify?

dvc/render/converter/vega.py
```diff
@@ -0,0 +1,316 @@
...
+                        for field in fields:
+                            if field in datapoint and def_filename in filename:
+                                result.append(_generate_y(datapoint, field))
+
+        self.inferred_properties = {
+            **self.inferred_properties,
+            **properties_update,
+        }
+
+        return result
```
daavoo, line 316: If I make a typo in a stage plot:

```yaml
stages:
  foo:
    cmd: python foo.py
    plots:
      - dvclive/scalars/foo.tsv:
          y: foos  # Should be foo
```

I get an explicit error `ERROR: unexpected error - Field 'foos' does not exist in provided data.`.

However, if I make a typo in the definition of the flexible plot, all the steps pass silently but `result`is an empty list and no error is raised:

```yaml
plots:
  foo:
    y: 
      dvclive/scalars/foo.tsv: foos  # Should be foo
```

daavoo (commented):
🚀 @pared . 

I think the changes look good overall. 

However, there are a lot of changes 😅. I think that, in the future, we should try to build something like this incrementally, split small P.R.s, and work on establishing some clear `must`/`should`/`nice to have` items cc @dberenbaum .

My biggest concern is that we are including quite some logic scattered around for supporting `--from-config` when it is a hidden option and we are going to initially push for the `dvc.yaml` section. 
It is already done so it is ok, but I would prefer to not include so much code that won't be really used in the short term.

---

pd. @pared it would be great if you could include some examples of all the new syntax/use cases supported in the P.R. description or somewhere else.

dvc/repo/plots/__init__.py
daavoo: https://github.com/iterative/dvc/issues/7932

dvc/render/converter/vega.py
pared: `VERSION_FIELD` is intended as future replace for `rev` - it gives more information in a structured format. I presume (still need confirmation though) it will be used especially in `vscode` extension to have better control over plots there. One more perk is that providing this information in this format could allow us to deprecate `rev` (for ex in 3.0) and make our templates better utilize data version info. For example we could modify the plots to work in this way: [example](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUMSSCGcCGgDaoZcSgB9GrIBmcAE524y8wE9z23XDQBGAAy+lUwtSOwh2WUsbNjRQaxo4QWU0EBxBNkxzX0UQOME4WSQEL3RVdSkZeSVHYhowuWSACQBRAEEAERAAXyVc-MLilTUqaXqlVPTMn38xtIzvKe8Zib8fPyWMgCYpjfXzAGYpvd2AFinj3bEpsV2ANimb3YB2KcfdgA4pt92ATinvqrgxEarTaqFQpWGFTB4wyWSUAA80L5uiYzJFbA4nK53Do9AtAmiQnUIlZrNFNDl4olkjDzPMejQ8gUiskIeVRiBqrVwsD2l0GUz+qyhuzKilZpNUH4AuLlvjZXMslK1grzFspb4dqqDhqjqrThrzqrLhrrqq7hqHqrnhrXqqPhqvqrfhr-pzAbzQeCRSNZNCJfSQIjUMiFKiLKT7I5nG4PHiNVqgu57MT0eTYlSkuhaVsBX0WSUfRVslziZ7+ZTBQXBmVfdlaUrpbt5hrFqqVsqZTntrsdX49bSDX4jbSTX4zbSLX4rbSbX47bSHX4nbSXX43dVPWC2b7-RNc0GkSiVGjI5iYzjPIcCcEU+E0zFKQks9rsr1mQMd8WATUy+hmny3SVvmn5FhyDZTF2AbyrSHbeCq3YJr2176mcFxXLc9xPC87yfD8fw-lu3q1lCqC0gcCJHmGJ4RjYUZYrGuLFMON7JqE96kumT7UtmEqnHmH7CiRHKljy-4ghW75CoWwlihBGpQXKrbNo2CESuqfhauRKGDmhxoYeaWHWjh9p4c6BHukC4ntNuYF+mRfHZMGobhui9EXnGzG+GaSZEhxUSPnEz40hKlwCdJNaQiJgLcvU1kdEBUnVl+4ESo20ztjBAaqYpmw9q+uonHpo4GZORmziZi5mauFmbvFtmyXuGRhYeIbHkmZ7RtinlGb5d5yA+FJBTxhnhcldkljFf4gABCVjaBsn1mlkHNllyw5bsGmashhWoYa6GmphlrYbauGOvhrqEfVxFRfZZWUW11EdXR57dUxJl9exA2cYFmYhRMzzzUJt2Tb+YkzRJiWMiBwOiktEzpbldJrYqqxI1tWkSn2vgDo5+36YdhnHcZp2med5mXZZREpXdEqA61Lk0W5r2MVe5OfamP1DX9vETB8QMySDP6xfI8WSdDgmC3DuyI6tynthtqoYzt-ZFfjJWE2VxMVaTVXkzVlN1RDNk3aKTXmPzDPtaeL1daz8brqxfnfQF3PBbzGS-ALkXS5ZIvllDVYLULqqy5l8uwYriGaSrONqyxBPjkd04nfOZ3LhdjtU9dNPm17VtPTbGJ25eDsqhz-lkr97sK2+EsRTToP+2Lgcw1LdYyyt4d+G2kdo5t+Xabtunq6FpUSlOvgznTlUSkuvgrhKa6+BuHo53Z5srA9jPPcXDGl15bYVy7VduyNEHe43wvTbN4tB7DHehyjyNIbXyq9+pOlY8VeNJ4nU-J18AuGe+s54WQNvBGURtZoNVuubOErVvDW1onvDy71X7HxJK7DMNdYJ13vu3b8fsb6Q0vhNTuEdoKvz7u-Aew9v6jwmEObyB0U5EzTiTDOZMs4QJVNAkEsCzYOTlA9JBhcUHuTemzVWTNnZYNPjg8+n8yGLWvuDW+rdJY+0fvJHucsY5v3gh-fcX8JjYxHL-Fh-9p4TEnsAgG1UwGU14VAtexsvS52EXlURyDmYlx6vjTBg1FEvm0iokOok4ruLvm3bRRDdG+GMXMQe2VVhJLVKYjI5j45WI1mwrWHCdZcL1jwpe-ds7uMEbuLx+wnI+F8Z1feAS-5BK5iE-6GR+LAS0VfYh6jSHdIbuQp+lClIGJoUYuhMih4JxHn-PJAD2FANTgvdOK9M58LcTA021TdI+PEX4pp6C2GtOwdxUJY9wm+0iaLaJmihmqJGXo7u21DHwSmXHAqszLETjHtrCes8HGgL5uAspnYrqVJ2aREq+zXKNLQdI5Zsj+ryK4sNC5tj8GxN6TcgOVydHLVGck6hqTaFK0ybUxhnTx4TDHIsgpSKimrO4esimkCIXbM8fdRBDTbZHMRU6U5CjzkdPMPTJKwdrlTX6YBfF8TCXPLwSS9aaSPm4zMT-JhNLmrlQBbrEBxSnGlImGuTZVlIVcpnrCpm8KpEO1ZUKtFPNTJyuimDKJGjXVyQVYk-RryJnvPJfQjVVLzDMN+bS3VmL9VAsNSC5xYL2UVM5RvGpltgxiLhXyhFZcZSOurkok1WKenDNxS3L18NYTPxbOM0lkyg3TIYd8rVms-n5L1YUg1zKSmspcRygRUK5B5zqVKXlqC7XMUDfms+GLUYVrUR6gZEqH7yoRtWlJKqyXR0+TM3wFiW3zLbfSjtjKu2L3jcaz25T+Em0tSq61u9JH2y8omQkKLgkio9pMedfTF2ysGeNR5CTMZjJ3XW7wIG8qhuxhG6l-zI0xoyJPc9GQ5y9qNWpItZKb0eNTRfVqGwx1PoPihad7Sv2BmXYQt1zc7k-uA369VqNlSQYydB7VYao06sQ+YZDKy3QlMw1e7DWyB13uSQ9QjBzbXPr0mRz95Kf1lrowByVBK13KuJY2zd8FWNbX3VkjjzCbHcc7dGuNqHQUXqE+YU1bYcNVOhcogjRGWYkcJvJ9FoqKKqZXTRkh-6qNxNShp2toGmN0lUnpzVhnW0HuPQhszSHHGxvQ9ZpGdn+23rw9-STrn-HHMWZ551ezfPUbFMpz1ZXgvetC-6qh2nmO6Y+QZylh74smfMHS+xyXgWWYTeljZ9nRPZcajUrpwYpNZvHbJ0mxXcGXOqzi6Vf65pLeGQxl5EW4IQZa7HWDnH4OmdPeZ7tXaBMYYy6q5NYmcuRry9J7NE6zLzcLclpTK3blVaC70zbSrGuRdWNF9jcW4PtsSyd3rFmxVWYyEuGzmWbujbgTUu4D3pvEeaQ6t9X1UUFtnTDj77qvtLvroBkOf2GtgZ07thte79s5M691wBKGYcDbh+U3hw3zUprGxVdHNqnuzaTa9gnlsfuls+3i9bQGfWsZrdTprtPt2tZg4z1hkPeMpf65ei2nPE2Boc4O2mIKBePrc80qdOPOZnK81+-OEvHmVdJwQmrlbJRha04rwHLG9tfNyZYpngKodndSxdwbbLDcjdw3zspZui6Y+ObT0Xoqt4y4iVL8t6ffaU-CypIHfvd0HeMxrnrWu+ts91-Dq7Imee3dj-elzj2ZvubeHm63lcnULbXUT2js0AB+8x7nk5z4Stvcs3jpI7N8dGaA3isZ1G8CLBo3itZNG8A7Fo3idZtG8MvDo3is5dG8Vedf2iD8c0OmpCDnL5f5fGSfTt31tIU3g3vAW2iD5iSW2XGnx-h0fzfhnzoXnx2iXzVlXwOg3yOm3xOj3zOkPwuhPyy0-28EvxNwkwLgxwt3QVAORVxw-Tt0U2z0fmd3PyH3ozHyRhbEAJoWAKDTwKHnAL2kgIJmgKJlgJJngLJkQIpmQKR1QPQPNgPFv2b0TzZmYJTy-R80dwz2J09C-2HzU1XSrVQH-yVVoNJXoMQkYIYWYJHlYI1nYK1k4J1m4L1l4INn4Jwwv2N3Nh81EOwIKwkKNCkNQnfxlVQO-weQpyoInynyVG0M-jUIXzn2XznzXzn03znx3zn33znyPzn1PwULQLsPGxHR3gTxwIkJ8g7xPi7zey62LR8KlXkPikUMoL-2oLnwCLQCCJMRCLAPCLUMiLUOiLUNiLUPiLUMSLUOSPKNSPEyKO3jvxzWKE4LcO5VkNKL7xBAqJIJUMlHUKoU0M3XqKg10JDX0LxkML+WML1VMJAXMKcUsLBWsOj1sKGLRywMFxbwCW4MmKtQWP808PmOmPU1UOWKUlWOY3WLY02Ni22JbV2IQ32OjUOKBWOPjVOKw3OLP0ELSP5xuPN2cIf0FTyLxxnVFXFzJ2UJeNWzeNxL81q0+OqLUNqNQD+PVABMpSBLgxBO4zBN6whP6yhI5x6KQP6PcUuLuw5xGLEOyIfzdEeKLQ8IJIoOeJJKWLJJ+J9ypLn1CLUOaIZKKLaJgM6LZL1w5L4K5IH0GN5NswyNGOe0pPb1vAIJfyIL7jFJJ3IO8JHw+OlP8Pz1NJAMVLpMOxVPX3aJZLZ01IP16LhJSKEJqTT0cNuPEIfyPgxMIJK2WhtJSPtLxKlKVFlJoPl0CPSWpIi0X1axXwO3X06y3zL131ZwP3D2E3oJsP1IbyrQfSyNRPGNfXNJt2FStOggTIGKTOJPd1TKnwVJdO+CzLCLALzKiKgKLLiLgLLKSIuirIuJrJRzfx5QFMbJHPwNbIKIJwPHeKITIK8KUJ7IoTTIHKAOHKVNHIgILJiJgJLISIQIrMNNdIEJ5NrIyXrIkUFPGNcJjMtLjI1U7O5IlN3JC0+P7MaLPJAJzIiKvInNvOnIfLnNcXhNfKXNyxXKcPv3GNyJbM73x1FS6RAoq0zyAu7PK17JqInwzLqPPI9NzNgrYMnK6IQp4MfJdHnJQsXKEVKwzWNNmwmN-Nt3-J1UAr1LIrd2PPAppOn1ougpaIYqMKYs1NLMQopg4uDMRMW14tXKwqnI3LwqxK-WuKIqbg-0JNd1+zHykuospNksvJYOvI6Pgq4JnL6KQpQNQu4oBQ-MOTGPvP0vyPwq-XFSJPIoXVtIPMqLAqosHLsoKhVPzLgo4LvO1IsLYrqOQo0qGPpm0swr8rcoCsxPIxdUlNMteOAtCokpGRPIgroLiqYLHLVMYucrMNcv4INnUoGJDKqh8pk3c3oJFKvVErmIqoso2yspisgoYLkoSvHOauSpYrSvcpfK4t2TjwwojK-JoqfwtKEu7znVKvCsTMPLCuqustiqgvsoMMct9OLMWpOPSufOrO6tSV6qF1b2bLYk3KCqfmGrtJOqqvkmkvXLoOBxVMX2LxvLYLLy31Z130fIdD+PYu5yyoNJvzQCm02rXIvMKtjP2uRj+sisOuqvlxBq0LBtVzmoMKD2Yo4LhtnLJiRu2pWpepESbzypNNYMGvfOJv3PMuxXGr-1JpxtBpAMpqaupqgJhv8rMIRoyrnJRq6s0pMXj0-OxvYO5pkMquWzKNIoBssqFon220CIprAMhqcuhpgPpoKr1iZtso8tWqcwAvZqxt0otu5sIu1sl11rEv1sFs+OFo9On1NvivNt9PX2ltSqOLlufI6sVqAtZupVVt8s5rtG5pahMqOq7L9t-wDqNsHJDqYLDpps1NhrgJjrtqHIdsTuGJdpRLdt4O5uMq9qdxIt9qiqWMDuNrqMLr0OLqlqtvLrOkrvjr1JruuMmz4tb2FMErbOEsJ15rbpGvEoNrzoAO7spN7q2P7strpqHsZpryrpZuVss2Tr6oCQGtnq3OxOKIdL3KXv+o7tTK7oLrFrNogJLsjphNZIrsPtHpGprvTQxqnovujNwsCsMvMkJv5p-18MNvXtfoYPFqZM9M-sHq4N-oVurpPsNLPveovs+rkTxsKLT0zt-QipgZKMdOfvzqAK3sBJ3qMK-utvaqcRHuwaGLDOAZ0vypuDNK+oMuKvwzIb5q2BzrgdUL4blhuApMeFn1QBuEVJuGaJuFaJuHaJuE6JuG6JuF6JuF1LmI2BrvRseg5tmxkZ2u+sgeXJEYfs-zEafrQCkfDgsbfjkboUUZ2mUbVlUYOnUaOk0ZOm0bOl0Yun0Y8qMZwcDHDPrt4cIefz2sKJ3JbrkNmPPwceJvkmcaVVcZoXcaDU8YKm8b2l8YJn8aJkCZJmCbJlCYpnCZfMiaGJEKojMfc2Kc1tvuTLKoJIybIaybJNydJXycQkKaHmKZHlKY1nKa1kqZ1mqb1lqYNnqZsMaYNIcJaddt4Z-PAaKtf0cmgd6ZSdHw02yaoUGc3WGeCNGYYXGbxkmb+Wmb1VmZAXmacUWbBWWYuNWbfImw2diZNPKfTs6aPPIYUMObGtzslFOaUnOeY0uYaOuZDVuZbXuYQ0eejWeaBVefjXeaw0+ZQu+bQvu2RIbLdsqabuBdOtEZXv9qhYGdkfkcRdi2Rbg1Re43Rd60xf62xfZNxeE3xbBfHqNJ4YBbTqvp+qRNsZ9sMZpchaVGheJVhZ93hagyZcpRZcOzZdVI5fLy5bZx5a1L5afIFfKMJa8ocX5NaYCVqe5pxIhdSbMvBYFrlacfpZdJVbYzVZ1A1YNC1ZNB1YtD1ZtANYdCNZdBNe5LNbWtNxJbVrJZnp2eIYJwdyOdILsf7yddgeOckbdbccZaUZUbUY0a0Z0b0YMfSZrvzhidJfyvye5tIdTfvulfSdlYkbpekYZY8YLZ8aLYCZLZCbLYiZrq4dMc2YBbAYEYgaEfjMXubfsdbezahfArVY7EeHPI1e9cap1f9aYoNeDdcvqYNnyZWeMZHUzTHfMfid2rnvxso0bfxIoczaocWPleXZsrXY8bkq1d9ccr1cDZSqNdDceuPa+eHbPZAfQXafFesecylbSfnfEcXdfekfffXa-a3d-d3YA4PeA8ytNZruadHRFfMe2cnd2fbOdrg8dYXeoacbffdbQ68Yw78aw6CZw4uhA4JZrvWaI6tcg5wrI6TYIspaqupcQ9o4Ufo7ccY6KeY7KdY6qfY4pk48FZwd+d44vbaatCBYOZo5fbo5Q4Y8-aY58cw4Cew5Cdw6HZwZalys0+tbFcTb-PxubvtZmOo-E-08k8M+k+M9k9M5Y-M7Y8s447w8jaFberuMg-RKc8SYJxCrc7TbnYzb09AqXZ87yZk7Gbk6mYU7maU6PbC4HyjadtPo2v+fMYTcE+c8KLtedYddeKfbvrS+Q5cdQ78+y4C-k6C8U5C+U6K8McAfA+I-czreg+ndFNnfg5S885a4M7a6M4KfQ669y56-y768K+s6GKrfqRG4CQ-dxpq4JwbcS6bem6a66YoSVZbBXaVAO5GeW5KbM4qYs5qas4aeHZMfPYq7aavasYm7rKm488cYUeFo1dXeBy1e9fNr1f9cjqNeDZjo9ZdAO5PZwZMcxp++tb1HrZE51vO9S5TKcbB43seEh-Fp1d9ZpoNcDetsPacWR7QFR9A6iZHUx5rYBdI6IaO9FWSdO4fbBcJ4otB+kdJ-J68Zh+p-h7p6R5r2Z645wcI-Z7jd4YE+57i+8zx+9oJ9m6J5F5cbF48Yp8l78el6Cdl44--orZwZ4+V5TvMe0-G72a1V091+F5XaccN4KeN58al4CZl7OkZ9QHl9U6GPU7t-Psg8c+q416-QzvveIuS4u5BayZJ-dfF6KZN7KbN6qYt+U6t-sZrts+4b47Zhtad4o-eyB8a6F6u7T7cYz7Gaz6mZz7mbz6PYL4zYi7ro58q8scEed7K6o5r7d7r9F-T6N4l999N-9-N8D7l875K6vyRMnr2-QTG9i5vdq619bqT9r5GQ94Ua95GZ95Kb94qYD7JiD5D-w5waAdQAj-wfX4nfV63+TZ34a56f39T-H4b8n8z+n7Z9Z+ufefpby24GkduD-CDmzDJ798p2g-QHK7xB6H9wed3Rvjc2b4PNW+LzdvmChv7hcomGPaAfGGOD8NX+19L9AgmH4EkDgo-EZKQLljHAKSYgeRscEVLHBmixwVoscHaLHBOixwboscF6LHBy2n+PYKe0tb2d0ETAuAeR3np3t+eifabrQJB4MDw4Mgt+CwLoRsCdoHAtWFwIOg8CjofAk6AILOhCCLoIgjyuINZ6SCse0gv7gP0r481qBFDFQZk0JRqClUGgmhFoKDQ6CCoegvaAYIJhGCiYJgkmGYLJgWCKYVgl8jYKabCtS+JAnHhX3npa1FB3TNwd-08FkkfBpKPwYhACFDwghI8EIRrDCFawIhOsKIXrBiEGw4hNhBIWsySFSC2YIQj2h-3c6vF3BfTXIYwOYGsD2BnA7gbwP4GCDhBog-vM0J+atD7B7QtXgkzf6ip4+mQrOtyV6EJ9heXgqhPkM3SFDgixQhhKULxjlC-klQvVNUJAS1CnE9QsFI0IuIzCiWIlWNvb3cwRCKW0DTYWsPoF5DBh2g4YfoNGHGDxh5gyYdYO77VsVeJpKIdzQS71duhNAnIRph2FKQ9hzGA4Q0SOEhoThLaM4QhguHRorhQKG4fGjuFYYHhKFJ4eazK5Qi3hASGIbay6FJdlByI1QqiOJToifcmIqDNiNiy4i4M+I7jISN6zEj+spI9kuSOEyUiFC1I6NnyVeGR92hVXcgRK3WquDZRbIyUByJfhciOwPItjHyMpQCjDsQo1UiKPLxii2cEorUlKKfIyjyico0rrg0VFP82Yfg+tlQK2HrCB83whERJx1Etg9RSoA0eqCNE6gTRBoM0SaAtEWgrRNoG0Q6DtEugHRGwz7nMN77vCX+Swigb9Wr5Ii6B8kPUWgCNH6jzyJoiMY1QtEximKNohMa5TiEGw-BTQiQeV0zEMjHB8A5wQoP9FnczKforNgGPAqliQx5YuSmaKjGOUrRcYlKnaKTGPVmxjwsDpF0jLFAgh3NPnr2IF6OitRSoYsagBHFoAxAY43QVWKnG1jZxDYhcQN3PxOjl+zmOzvMJIFc8cxaoyjt6NBY7jCxng4cTZWPHaDxxZ4wwReNMFXiLoi4qkdx2G7JC1xiw69rmNKwaivxqg38S6X-H+DAJ+g88cYMvHmDrxEItTtBLaEkDHem-BCYtiQkbDdxJY1CZoJPGBCgJoQkCZELAkUwIJsoovkRKfFrjo+qomDpii+HUSDxtE3wfRJKGMSKhzEmoaxKbE3ixBkI3bjBJLExcY+yw4KsyL7E9ChJ+4w8agHQlFDMJwQ7CeENwnRD8J8QmujlUUnES1xKo18fxIVGUTfR2kkSQULEnHCJJ5wqSdcJklgp2JjoobiuK2p6SyB9kgHi6KclzEBxz7ObsJMYF-j3JOIzyQSO8kkjfJWGfyWmJwaQDvu7Y9BOhM9EaTtxVE78SiNcn7DEp-I5KcKNSnij0pwmTKb6PTFBTsaAQ7ml6J+F81opzXPXqWJLEb0xAwOM0RGPNpWiYxkdO0QmJjoGiXQ6ElsejzZ7EC1xqQsiW+MwKRTbx2k4WiaP1FDTxaFoqMTTRtFxjrajYpxDNKPGd87xGBAmj32hGzYOhaQ-GpuMHGaSCxqg7aQNL2m6DRpR0iaadOmk145pS4xXotLX4LDZBQnaQkVKUH9itpjAr6doP2m-TDB-00wYDPAlXSoJeDKLu0NImqTyJLvfMdkNKnsjPpaE76YEJRmhC0ZkQjGWxKxmEScZq4ksbxLCkIDVhW42GVpNJnajyZmgymSUOpkVDaZNQ+mU2MZlDFi+UA8GSQJUl8TwprnLmVkM1G8y9x-M3wYLOOHCzzhos64eLL8mSyDSE9EvjZJLF2T4Ja0hehtLEHwz1BiM-wcjP0F-TjBAMs6BdL0lGy3yVkmWUpJCmQyeeX6Orq9OKnOS1ZJYjWQUK1k4idZBIvWSSINkZSvZzwvXMzOCkFSnphRFNp1PTbdTLu9AyOfsOjn8jY5wo+OeKMTkNTk5NIl0avz9mDSA5sfWuIJPDkHjC5GI4ucaNLnmjy51oyuU+WBmQTCBYMv2RsFCmWyHJ36YmQoVOCtyx5csDYBST2DyMNgipDYM0Q2CtENg7RDYJ0Q2DdENgvRDYFMOOCti6RSo+MIvMblqTMo0DWeSD3nnhwr5b8ZeXQlXk7R15asTeQdG3lHRd5J0feWdEPkXRj5HlU+bYNdG4zL5nYuQc9Jhkqzyi98jwRpkflKpn5NCV+UGnfkFRP5e0b+QTF-lEx-5JMQBWTGAUUxQFL5cBYkLsF5S2YuCjpnfKEmoKqE6C0lJgsQjYKh4uCkePgo1iEKtYxCnWKQr1jkKDYlCmwtQpaG0L7p7mfBZ0KYVzyySbCzdBwuCJcKGEPCvGHwr+QCK9UQikBCIqcRiKwUEii4lItmEyL6R6CQhTp2nmILmFyipeSvLXkbyt5O8veQfKPknzOJVii+cUGIWfD7F3JJBX0JQVOKXSaihohopDRaKW0OihDHoujQGKgURi+NCYqwxmKUKFilOSbNHbcSMabMieeFPhEhzuZBJUJR+PkgsKlIKi5jFEqgwxLYscSuDAku4xJLesKS-rGkvZIZLhMWSmeZZIzGyKAk5CpkYoofkRKX5Lij+W4p-keKAFXikBT4rv7DLrF9Ci2f9wQHZzlZPouYpUp+HVKplGCmZTgrmUEKFlJCpZRQpWXbc1l-itAJgsKkTLkFqhGpcSjqU+4GlbGJpZShaWHY2lqpDpeXi6Vs4elWpPpU+QGWILmpkClmQ-2zHFKEBHU3ZZ+JCXMLwKvyjsHsHPL-KdQgKg0MCpNCgqLQ4Km0JCodCUKDYmCyRWfOskFKH+MCqGbfOCUD4DlqK6pZipso4q35clAlRjUcokrClAC1ylSrBQ0rzFy4tsSMpsUrSCZVsl6TFPKUUN2VZS4Xp8pbBYqlQPKrBXysapEqMaTFMlRjVFWPUJV2SgjlxLoWXyXxSK5wRkNRV81VVSq9VVysiW4q9VX8wVUapSoUqMaZquSf3hyU1yeOuUmVfQrglbLnBntHOcl2dU9TXVC87lR6o-n6rvVf831aaoujmrBlTM6Vessvn4yFZHM+BXsvPzxr85nKpNe6t5WpqvVP8n1SKqAUBqwFvi-NQ8of5FKo189JWWqrLWf4K1KfQlBqoxrJra1OCtNQ2ozVNqyFLaqhQpI04Mqxlmc+LqWrRVsqMV1al+SmonX1qCFjakhVmopg5qYVODH2WGoLUBLNlTg+esHJdX9qg1m6p+WOt1V1q8F6aohZmubXZrA1wa+UanPbVuj4wTyldaKh2V9r11+yp9WgpfWcLPV76qdZ+pnWiK51tK7KVavDXAbEV3a29iiog1OroNrC2Deovg28KP1gir9bOp-Wtrh5LUt2tgvalrqCNc84Wv8uxXA5AV+K82qCuJWR1IV5KmOt8pdA6q0NQxIgbLICVyri13Ypjem0HWnVqlrGjensA43i1gVhKmmuCtJXW0xV8aITY8s75-rnR0TU2UuttU4akmsmuNcwqU2RLVNH87jZpr406bBNNeETZKtBlpzsati0DdDJeVhK3ltml+fZpwWOaf5zmgBa5uzWGbsZd0y9YauvmEyk6rKqDSxoXnKaQt3CsLQQoi0kKotx6mLXmrrlmzO1iWq2ZzPw1yabN6WuzW-LU3Zb+FuW4RflupWFapZI8krcutWmTze196yDeWuq1PyMtdWhzV-Kc1-yXNZ0fTagHc0WqcGeSx-lAqvVlbJ5pSvrcxoflBaMFmWzRQ1t0VNbDFLW8VW1oNI+zFt8KkDd1vCl3qE1D6+TYDWHVbb2FO22JXtsSUHbUlR2rDLNtzVDF7+524KTqu5rgb1tVWtLUNtq1YL6tY28LRNsi1Ta3NJ2t8pAIB3Y0VNK28KSd0dVg7NtNW4LSNtC0w6ctcOvLQjui00bOG4mv2ePIs0E48NoO5LpcFblkkKSNmdUIqU4EHQNGJ0bor0SmFiA6V+S61cxAx0ICexlWxnUJJZ0uk2dGNHaJzoJjc6SYvOuch5QF0QLz5QGgJaLucGKrbt-Wz-EzpB7S7DEK8+XfoK50BMedCBVXS+XV00K4VgOnXekKs3TcjdryyUCbomRm6CoCujWErp1gq61Kau2LZrqW0ljnd+NGNdjsl3M79EMun3UPD91-IA9ICIPR1RD1Fa-mwuo8ZHsKIVaGdbuqXfHtN1vzzdwQy3RU2t08FbdNhe3QaWlmZFMNxQItezOcG9b9dfNd3QFs90l7vdZe33RbsV1W7ldNu4PXboXVN74twfPPauugbd6qlhKL3XWkT0MJk9CGVPUCnT1gp+dQyvxVrrnyz6b68+4vd3AT0D6k9Q+-3SPsD1j6M9E+1ZfvvD2Ukj99uV3WZQX2HKl9felfRfrX1X6U9N+tPXfp32Z67lT++FeXF83NyUt5+T-Ryu-1n7S9WC8vWUMr0zNq9FhWvRcXr1vkR2U+jtdhpvX416dne9NvAYg26IF5Mu88uOLVi-sjoKVVyo9TNQKFcDKcr7ktLl2Hcm5NjD8V3qEmYqaDdCOg3tAYNEwmDw9TOIGvYM1yTNi6nPTNtf3EEbZ-eCg31qoNPzhDQaUQyPHENaxJDjNaQ2AYNKEcL1Ha8zcQcKIOqJdRe1uUIbeQiHdB9BvxowbgJSG2UMh0PfSsUORqrDBOaPbYY-2CHqDjhnQ84bEOuGJD7how54ZMOWLANz+1vXavnoF6yDsekHg4YDROHAhLhspm4a4IeGIEXhmzhhun1dr-DoqDvfnIEP2HQj2R8I7kciP5HojhR2I8UfiO5KyjHa+WW3vnprb0jdhzI-UfAw5GSheRqZgUbMJFHE0JR7Kt0YP0v6eDN8nqrAcN0hGtDYRxCLobxj6G9Uhh22sYYf1-aFjz+qA1du2Xv7Xi6h-XZobQXaHtjERvQ1EYMMxHDjcR44xANOOQGiDXY+QaQZqPkGNj9xrY8ER2Mto9j0aA42wyON17YVj4xQ0ysDl5jVDNxytcOrlgb1lchwnaDDwOhf0ToMdBHKsE6M1yqdJWqTX0dvZXGCSaJodSgsxMy6hpuJnxviZgKEm1kGyUk-+vkOo63alhv43ApP1zzGThiZkwVDxMEwCTJMIkzXijwoVZD-6pXlwb0nKGwkaxtQ8wtFMTJxTQ8SUxrGlM6xZTXJz42+Vt4qnkjtO4TsKYfnam60uphhPqb+SGmQExpyPNyedHh8VTFRwU-nppMUM6TCmjE+HCxMOmQ0TphDC6aBRuneEHp+8cS2K0MrejKRlzv6bYNamQzTJ7QSyZKZsmOCHJllCabhPzaOtDK69b6bn0anAzD2hk5mbFPZmJTrJqU+yZlOcn3TpplOWdpVPnH5Vk8m7YCYyMe6lQwtUMw2b1NNmDTLZo022djMdma5-27s78dgVZy0z5Ras6vUlAjmsz-gnMxMzzMmECzPaIszgcralnFDyeaA9aSrMZmlUo5nc42dzPNn8zrZws+2eLOU7FpAAXSAiYA7ASAWQBADJB2AEARgUAFAB0BQAGAggNQAtAYAIBKAjqSgBKgAAEAAamQsAByMEBhbQvIXSgcFhC1fSQt-QlASAAwIWAgAwBzAygGgL+DYB2Augn5pQJgDoB5BWQZgSgC4EKCCB0C2QAAO40BlA9AR5RlHgA0AyAWAV1jKGgsuB7AoFkAPkCgBsAaLsgMgI+GDDFVEQzFlwDgAGAABHBgP+ZYvqAWLpAbIMZbYvoBEQQEFwD9QKXaXdLyQAy0ZboBqBaLXgZi6xYGC2WlA0AHQMUFAAAAvewOSGsA6AIAcAToEBCUtpAGLkDVQJReou0Xag9F8yzpYGCyA2ACAKwDoGyB5AyA+QLMKAGgtMAEgAAGRoDZXMAIl5EEBAgC-m2AAAazgBtAyLMAOy0RdxQOWMrWVnK4IC6BAQZLcl1AMYBADIA7AjV5IMIFkBeBjwEVvIFABYtxRirSAUq-1YpCYB0ryQMIKpbYtKAZrSARwPVbQC-m9ASgKJAgDYAMAIrbAEINkEUvKWrAZAci4YEPBMWFLCAHAJteSCZWZr2QKAHkEOvJBLr11uAFdeqtRWlA41ya9mDYBWBqrSgB6ypbUsUg2AOAJAFADoC2WKQSl2QDRaWuixQA81uAItfBglWEg2QZpEBACS1XOg356iL+f-OAX6LIFkaxmEECYBhrRNhICTYJtTXVrFNqK+9aGtxXRr0Nx8JtccvoA7AUFjyyABiupX0AZAP87ZaAhI2nr6ln6lpZACS39Lhl2QMZbctmWorx4dW6pYltwB4Q1Vja1tfQDOWDbrl0y3LaIIFKgIGlhTNrd1tOX9bhtp2xWGxvFV7LOt22yAHtu+33LA1pQCLfkvi2bbUtnW5bYRsgAdA4l25HkGsBJ3lAwYHyNjf4BiA1bsgJS8jcfAK24roAL2+gEyvZWCg-VhkDXAStUWaLdFhiybe-Nt2lAuNuICjeABRWgAA)

daavoo: Thanks for the clarification! 
I guess my follow-up question is similar to my comment regarding `--from-config`, if this is not a strict requirement for getting a working version of flexible plots, should it be included in the P.R.?

dvc/repo/plots/__init__.py
pared: We should, but it seems like some kind of a bug. Need to investigate it further. Fix won't be included in this change. Added TODO.

dvc/render/converter/vega.py
pared: Yes, as its currently used in `_adjust_vega_renderers` in `dvc/commands/plots`. Also, if we would not be using this field now, we would have to have separate logic for `VegaConverter` for `vscode` and `DVC`. This field allows us to modify the datapoints we need in `commands/plots` depending on whether we produce HTML or JSON.

dvc/render/converter/vega.py
pared: Investigating, 
Note: 
```yaml
plots:
  plots.csv:
    y: foos
```
seems to be working as intended

pared: @daavoo could you provide some info on your setup? I tried with non-dvclive setup, then, thinking that's some `dvclive` issue I did also with `dvclive` but in both cases I get expected error.

pared: My script:

```bash
#!/bin/bash

set -exu
pushd $TMPDIR

wsp=test_wspace
rep=test_repo

main=$(pwd)

mkdir $rep && pushd $rep

git init 
dvc init

echo -e "step,foo
1,0.1
2,0.2
3,0.25
" > data.csv

echo -e "from dvclive import Live
l = Live('logs')
l.log('val', 1)
l.next_step()
l.log('val', 2)" > code.py


dvc add data.csv
git add -A
git commit -am "initial"

dvc run -n train --plots plots.csv -d data.csv cp data.csv plots.csv
dvc plots modify plots.csv -y foos

dvc run -n log --outs logs -d code.py python code.py

git add -A
git commit -am "second"

echo -e "plots:
  foo:
    y: 
      logs/scalars/val.tsv: foos" >> dvc.yaml

dvc plots show

```

daavoo: @pared please see my last comment. The issue is when using multiple `y`:

dvc/commands/plots.py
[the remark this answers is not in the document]
pared: Should be fixed - I am not sure that new version is better as it modifies plots definitions depending on `cwd` but on the other hand it does not require us to resolve the paths in every place that we need to use plots.

dvc/render/converter/vega.py
pared: I can confirm that this problem exists. I spend some time thinking about it and I am not sure we should make `dvc-render` raise this error. We had few changes aiming to do "as much as possible" rendering before (#5525). Now, if there is no particular column we will fail `show/diff` command even if other provided plots are ok. I think we might need to adjust renderers to log warnings and skipping rendering to keep up with this approach. 

What do you think @daavoo?

pared:
> However, there are a lot of changes 😅. I think that, in the future, we should try to build something like this incrementally, split small P.R.s, and work on establishing some clear must/should/nice to have items cc @dberenbaum .

I agree that this change is too big. Existance of `--from-config` is a result of our engineering team's discussion whether we should include plots in `dvc.yaml`. Once I created it I decided to leave it as its maintenance was not that problematic, especially because it's not an official option.

pared:
Examples, as per @daavoo [request](https://github.com/iterative/dvc/pull/7477#pullrequestreview-1016553738):

### 1. Simplest

```csv
# logs.csv
epoch,loss,accuracy
1,0.19,0.81
2,0.11,0.89
3,0.07,0.93
4,0.04,0.96
```
```yaml
# dvc.yaml
stages:
  train:
    cmd: echo "Hello world"

plots:
  logs.csv:
```
plots_show_spec_default

### 2. Customizing
```csv
# logs.csv
epoch,loss,accuracy
1,0.19,0.81
2,0.11,0.89
3,0.07,0.93
4,0.04,0.96
```

```yaml
# dvc.yaml
stages:
  train:
    cmd: echo "Hello world"

plots:
  logs.csv:
    x: epoch
    y: accuracy
    title: Displaying accuracy
    x_label: This is epoch
    y_label: This is accuracy
```
plots_show_spec_simple_custom

### 3. Multiple data series

```csv
epoch,train_loss,test_loss 
1,0.33,0.4 
2,0.3,0.28 
3,0.2,0.25
4,0.1,0.23
```
```yaml
# dvc.yaml
stages:
  train:
    cmd: echo "Hello world"

plots:
  test_vs_train_loss:
    x: epoch
    y:
      training_data.csv: [test_loss, train_loss]
    title: Compare loss training versus test
```

plots_show_spec_multiple_columns

### 4. Sourcing data from different files
```csv
# train_classes.csv
actual_class,predicted_class
dog,dog
dog,dog
dog,dog
dog,bird
cat,cat
cat,cat
cat,cat
cat,dog
bird,bird
bird,bird
bird,bird
bird,dog
```

```csv
# test_classes.csv
actual_class,predicted_class
dog,dog
dog,dog
dog,cat
bird,bird
bird,bird
bird,cat
cat,cat
cat,cat
cat,bird
```
```yaml
# dvc.yaml
stages:
  train:
    cmd: echo "Hello world"

plots:
  test_vs_train_confusion:
    x: actual_class
    y:
      train_classes.csv: predicted_class
      test_classes.csv: predicted_class
    title: Compare test vs train confusion matrix
    template: confusion
    x_label: Actual class
    y_label: Predicted class
```
plots_show_spec_conf_train_test

dvc/render/converter/vega.py
daavoo: Sounds good to me. Let's open a separate issue for it

tests/integration/plots/conftest.py
```diff
@@ -79,3 +81,109 @@ def make():
         yield image_v2, linear_v2, confusion_v2, confusion_props
 
     return make
+
+
+@pytest.fixture
+def repo_with_config_plots(tmp_dir, scm, dvc, run_copy_metrics):
```
daavoo, line 87: @pared I think we are missing tests for the case `3. Multiple data series` in https://github.com/iterative/dvc/pull/7477#issuecomment-1165429358

pared: Yes, its true. I will fix that in follow up PR, where I intend to parameterized this test to handle one use case per test. That way we won't have to create this huge `repo_with_plots`.
This specific use case (list of columns) is tested in `tests/unit/render/test_vega_converter.py::test_converter` - its a unit test but should be enough for now.

PR #7488, opened 2022-03-22
odb: set fs_path to be "" (root) for webdav 

### LINKED ISSUE #7288: url path for webdav push gets duplicated in dvc versions >=2.9.0
# Bug Report

## Description

We have identified a bug(?) introduced from version 2.8.3 to 2.9.0. When pushing dvc data to a webdav server the path of the url gets duplicated. For example: If we push to webdavs://servername.com/path/to/location it pushes to webdavs://servername.com/path/to/location/path/to/location (creating all folders of the second "path/to/location" on the server). As it works with dvc version 2.8.3 we think that this might be a bug within dvc version 2.9.0 and our setup might not be too interesting for you. However, here is a rough overview of the webdav server setup: We are running an apache2 webdav server as our remote storage for our dvc projects. This webdav server is behind a reverse proxy (nginx).

### Reproduce

Using dvc version <=2.8.3 everything works as expected. dvc version >=2.9.0 it does not work anymore.

### Expected

Version 2.9.* should work the same as 2.8.*

### Environment information

The environment information might not be too interesting, but I can add more information, if needed

**Output of `dvc doctor`:**

```console
$ dvc doctor
```
version 2.8.3 fails withtin wsl, as reported elsewhere (https://issueexplorer.com/issue/iterative/dvc/7012)

Hope the report is detailed enough. Will add information if necessary.

skshetry:

* [x] Tests pending, most likely involves change in `dvc-webdav` repo. (See https://github.com/iterative/dvc-webdav/pull/1).

Fixes #7288.

### DIFF
--- a/dvc/fs/webdav.py
+++ b/dvc/fs/webdav.py
@@ -35,9 +35,11 @@ class WebDAVFileSystem(FSSpecWrapper):  # pylint:disable=abstract-method
 
     @classmethod
     def _strip_protocol(cls, path: str) -> str:
-        from fsspec.utils import infer_storage_options
-
-        return infer_storage_options(path)["path"].lstrip("/")
+        """
+        For WebDAVFileSystem, provided url is the base path itself, so it
+        should be treated as being a root path.
+        """
+        return ""
 
     def unstrip_protocol(self, path: str) -> str:
         return self.fs_args["base_url"] + "/" + path


dvc/data/db/webdav.py
```diff
@@ -0,0 +1,27 @@
...
+
+        WebdavFS expects relative path 'path/to/file.txt' instead of
+        base_path 'dav/path/to/file.txt'. So, we are setting the
+        fs_path to be '' (root) itself instead of 'dav'.
+
+        Note that this should be considered a temporary solution, until
+        we find a better solution for filesystems that don't support full
+        path.
+        """
+        super().__init__(fs, "", **config)
```
skshetry, line 26: We use `fs_path` to keep track of prefix, but `WebdavFS` already handles that itself, so what we need is only the subpaths.

dvc/data/db/webdav.py
```diff
@@ -0,0 +1,27 @@
+from typing import TYPE_CHECKING
+
+from dvc.objects.db import ObjectDB
+
+if TYPE_CHECKING:
+    from dvc.fs.base import FileSystem
+
+
+class WebdavObjectDB(ObjectDB):
```
skshetry, line 9: We could just set `fs_path=""` in `get_odb` method, but I wanted to make webdav special. Therefore this method exists.

skshetry:
@efiop, wdyt of renaming `_strip_protocol` to `get_root_path` and `unstrip_protocol` to `get_full_url`?

skshetry:
@efiop, PTAL. I changed `_strip_protocol` to return root marker.

dvc/fs/webdav.py
```diff
@@ -35,9 +35,11 @@ def __init__(self, **config):
 
     @classmethod
     def _strip_protocol(cls, path: str) -> str:
```
skshetry, line 37: I am still confused by the whole situation with `_strip_protocol()`. The `_strip_protocol` has two use cases right now:

1. In [`fsspec.spec.AbstractFileSystem`](https://github.com/fsspec/filesystem_spec/blob/8f62e479ddd45285b3e7126e9313151f1bebd3b9/fsspec/spec.py#L160), this is used internally to strip protocols, i.e. `<scheme>://path` to just `path`.
2. In DVC, to get `fs_path` which DVC uses as a root path/fs_path. https://github.com/iterative/dvc/blob/81eee059950d6a154327e3b15912a84d4b93e178/dvc/fs/__init__.py#L117

From [the docs on `_strip_protocol`[(https://github.com/fsspec/filesystem_spec/blob/8f62e479ddd45285b3e7126e9313151f1bebd3b9/fsspec/spec.py#L161-L164):

> Turn path from fully-qualified to file-system-specific
> May require FS-specific handling, e.g., for relative paths or links.
```

It may seem like 1 and 2 are the same thing (as per docs), but they really aren't.
Regarding the above 2nd point with DVC's usecase, we can just set `fs_path`/`root_path` to be root (with `""` aka `root_marker`).

Given our plans to make it more fsspec-compatible, I don't want to ignore the first use case of `_strip_protocol()` (and, want to avoid wrongfully triggering `_strip_protocol` for other use-case). The `_strip_protocol` also being a class method does not have context about `base_url` to handle this. 

To me, given how we are using this, what we really need is `get_root_path` that defaults to `_strip_protocol` for other filesystems, and for prefixed filesystem, it returns root_marker.

Alternatively, we could add support for PrefixedFileSystem in `odb` itself as proposed before, by setting `fs_path` to a root_marker.

Comment (efiop) [reply dvc/fs/webdav.py:37]: I agree that we could use DirFileSystem, but it doesn't seem to solve the problem with the filesystem itself. It looks like we are struggling becuase of the way our config expects webdav remotes to be configured. We currently expect:

```
url = webdavs://example.com...
```

which creates a problem of distinguishing path from base url, but if we would to configure it like

```
url = webdavs://path/to/
base_url = https://example.com/...
```

that would remove this issue completely, similar to how you configure endpoint url in s3 remotes. I remember we've discussed this during our 1on1, but I don't remember particular issues with it. Please feel free to remind me.

efiop: Looking at it a bit more, it seems like that would be the correct approach, as webdavfs only cares about that base url. And it would also remove our old terrible hacks of converting webdav(s):// urls to http(s):// urls and getting confused about it in the code base 😄 But might be missing something.

skshetry: That solution to force our internal logic on the users feels wrong to me. That does the same thing as setting `fs_path` to root by default (i.e. without `url`). We can discuss changing config for sure, but that should be a separate discussion (not motivated by this bug at least).

PR #7497, opened 2022-03-24
exp init: change ui prompts structure and output

### LINKED ISSUE #6446: Epic: exp init

### LINKED ISSUE #7130: exp init experience
```
$ dvc -V
2.9.2
```

```
$ git init
$ dvc init
$ dvc exp init python train.py
ERROR: unexpected error - [Errno 2] No such file or directory: 'params.yaml'
```

- [x] Why is it `unexpected` error? It should be like “ERROR: hyperparameters file ‘params.yaml’ is missing”.  🤔  Why `dvc exp init` does not create it if it is required?
- [x] Do we really need [Errno 2]? In what cases the code helps users? This makes the message longer and breaks human readability.

```
$ dvc exp run
ERROR: failed to reproduce 'dvc.yaml': [Errno 2] No such file or directory: '/Users/dmitry/src/test/myproj/data'
```
- [x] full path. It is supposed to be relative. Why: 1) to make it short - the full path provides zero value in most of use cases; 2) it makes it hard to submit reports (some users are not comfortable to share their directories struture - See `'/Users/dmitry/src/test/` - bc it might contain confidential info like user and project names.

- [x] why `dvc exp init` checked `params.yaml` but didn’t check (or create) `data/` & `src/`?
```
$ mkdir data
$ dvc exp run
ERROR: failed to reproduce 'dvc.yaml': [Errno 2] No such file or directory: '/Users/dmitry/src/test/myproj/src'
```

- [x] the same questions - [Errno 2], Full path.

- [x] why it was not checked in `dvc init`? Or was not created?

```
$ mkdir src
$ dvc exp run
…
python: can't open file 'train.py': [Errno 2] No such file or directory
ERROR: failed to reproduce 'dvc.yaml': failed to run: python train.py, exited with 2
```

- [x]  “exited with 2” Is it useful information? Where the exit code 2 belongs to? It might confuse users. Error code can be in debug errors not here.

- [x] Another issue - ideally, we should have a reference to stage (a user might have multiple stages with the same or similar commands). The message might look like:  **ERROR: failed to execute stage ‘train’ from ‘dvc.yaml’**

From a clean git repo:
```
$ dvc init
$ dvc exp init -i
…
'params.yaml' does not exist. Please retry with an existing parameters file.
```

- [x] If I do not have `params.yaml` I stuck here. Why don’t we offer to create an empty params file?

- [ ] **A kind of conclusion**: `dvc exp init` is not fully supported without interactive mode.
It looks like today the only supported flow is — interactive. The default `dvc init` has multiple assumptions about the project structure and it is not clear how a user should know about the assumptions. Why `dvc init` won’t create all the directories & files? This was a part of the initial idea - to run a single cmd for a given train.py in a live coding section while `dvc init` wraps it to an experiment.

This is how it can look like:
```
$ dvc init python train.py
Creating experiment project structure:
├── data/
├── metrics.json
├── models/
├── params.yaml
├── plots/
└── src/
Created `dvc.yaml` file with ‘train’ stage. To run, use "dvc exp run".
```

skshetry:

Closes #7130 and closes #6446.

### Screenshots
Screen Shot 2022-03-25 at 15 04 32

---

Screen Shot 2022-03-25 at 15 05 36

### DIFF
--- a/dvc/commands/experiments/init.py
+++ b/dvc/commands/experiments/init.py
@@ -1,5 +1,6 @@
 import argparse
 import logging
+from typing import TYPE_CHECKING, List
 
 from funcy import compact
 
@@ -7,6 +8,11 @@ from dvc.cli.command import CmdBase
 from dvc.cli.utils import append_doc_link
 from dvc.exceptions import InvalidArgumentError
 from dvc.ui import ui
+
+if TYPE_CHECKING:
+    from dvc.dependency import Dependency
+    from dvc.stage import PipelineStage
+
 
 logger = logging.getLogger(__name__)
 
@@ -29,7 +35,6 @@ class CmdExperimentsInit(CmdBase):
         "plots": PLOTS,
         "live": DVCLIVE,
     }
-    EXP_LINK = "https://s.dvc.org/g/exp/run"
 
     def run(self):
         from dvc.commands.stage import parse_cmd
@@ -58,7 +63,7 @@ class CmdExperimentsInit(CmdBase):
             }
         )
 
-        initialized_stage = init(
+        initialized_stage, initialized_deps = init(
             self.repo,
             name=self.args.name,
             type=self.args.type,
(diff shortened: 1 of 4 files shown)


dvc/commands/experiments/init.py
```diff
@@ -68,25 +69,38 @@ def run(self):
             force=self.args.force,
         )
 
+        template = "[green]{0}[/]".format
+        if initialized_deps:
+            deps_paths = map(template, initialized_deps)
+            ui.write(f"Created {humanize.join(deps_paths)}.", styled=True)
+
```
skshetry, line 76: Still need to iterate on the message/formatting/coloring, etc.

pmrowla (approved):

dvc/commands/experiments/init.py
```diff
@@ -67,32 +73,63 @@ def run(self):
...
         )
+        self._post_init_display(initialized_stage, initialized_deps)
+        if self.args.run:
+            self.repo.experiments.run(targets=[initialized_stage.addressing])
+        return 0
 
-        text = ui.rich_text.assemble(
-            "Created ",
-            (self.args.name, "bright_blue"),
+    def _post_init_display(
```
skshetry, line 81: All of the outputs have been moved out of `dvc.experiments.init` now (except what we need for prompts) to `commands`.

dberenbaum:
Nice! It works really well, and I think the simplified output text is better. We can probably even drop the line with the link at the end.

I'm not sure we need so much styling specific to this command. I think we can even further simplify and make it more consistent. What do you think of this (from https://github.com/dberenbaum/dvc/tree/exp-init-ui)?

Screen Shot 2022-03-25 at 2 58 02 PM

Screen Shot 2022-03-25 at 3 00 54 PM

Light background:

Screen Shot 2022-03-25 at 3 02 09 PM

Screen Shot 2022-03-25 at 3 02 51 PM

skshetry:
> I'm not sure we need so much styling specific to this command. I think we can even further simplify and make it more consistent. What do you think of this (from [https://github.com/dberenbaum/dvc/tree/exp-init-ui](https://github.com/dberenbaum/dvc/tree/exp-init-ui?rgh-link-date=2022-03-25T19%3A08%3A21Z))?

That makes sense, I went a bit over the top. I want to keep the last message as "You can now run your experiment" as
 it's more of a tips/hints than a command. Also, first two lines being a status output, `.` (period) should not be there.

skshetry:
@dberenbaum, I’m merging this. I will iterate on top if there’s anything need to change.

PR #7517, opened 2022-03-29
ui: add revision info on dvc.yaml/dvc.lock/.dvc syntax and schema errors

skshetry:
This is done in a hacky way, so it only works for dvc.yaml/dvc.lock/.dvc for now (unless we fix brancher).

Screen Shot 2022-03-29 at 09 40 13
Screen Shot 2022-03-29 at 09 40 47

Unfortunately, `brancher` only gives us revision instead of the reference (eg: `HEAD`).

### DIFF
--- a/dvc/utils/strictyaml.py
+++ b/dvc/utils/strictyaml.py
@@ -71,13 +71,20 @@ def _prepare_code_snippets(
 
 
 class YAMLSyntaxError(PrettyDvcException, YAMLFileCorruptedError):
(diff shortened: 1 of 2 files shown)


dvc/utils/strictyaml.py
```diff
@@ -254,13 +265,14 @@ def validate(
     schema: Callable[[_T], _T],
     text: str = None,
     path: str = None,
+    rev: str = None,
```
skshetry, line 268: It feels wrong to extend exception to hold `rev`, but I can't think of better ways here. :(

PR #7523, opened 2022-03-31
setup.cfg: tests: remove unnecessary dependencies

skshetry:
Hopefully will reduce some dependabot traffic.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,12 +35,9 @@ zip_safe = False
 packages = find:
 include_package_data = True
 install_requires =
-    # See https://github.com/pyinstaller/pyinstaller/issues/1945
-    ply>=3.9
     colorama>=0.3.9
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -35,12 +35,9 @@ zip_safe = False
 packages = find:
 include_package_data = True
 install_requires =
-    # See https://github.com/pyinstaller/pyinstaller/issues/1945
-    ply>=3.9
     colorama>=0.3.9
     configobj>=5.0.6
     nanotime>=0.5.2
-    pyasn1>=0.4.1
```
skshetry, line 43: We probably don't need to specify these. `ply` is never used by any of our dependencies, `pyasn1` is used by `pydrive` and `gcsfs`. `gcsfs` already specify that range, `pydrive2` does not (but probably does not need them).

PR #7524, opened 2022-03-31
`run/repro/stage add`: regroup options/flags

jorgeorpinel:
Matching https://github.com/iterative/dvc.org/pull/3345

### DIFF
--- a/dvc/commands/repro.py
+++ b/dvc/commands/repro.py
@@ -67,33 +67,19 @@ def add_arguments(repro_parser):
         help="Reproduce even if dependencies were not changed.",
     )
     repro_parser.add_argument(
+        "-i",
+        "--interactive",
+        action="store_true",
+        default=False,
+        help="Ask for confirmation before reproducing each stage.",
+    )
+    repro_parser.add_argument(
         "-s",
         "--single-item",
         action="store_true",
         default=False,
         help="Reproduce only single data item without recursive dependencies "
         "check.",
-    )
-    repro_parser.add_argument(
-        "-m",
-        "--metrics",
-        action="store_true",
-        default=False,
-        help="Show metrics after reproduction.",
-    )
-    repro_parser.add_argument(
-        "--dry",
-        action="store_true",
-        default=False,
-        help="Only print the commands that would be executed without "
-        "actually executing.",
-    )
-    repro_parser.add_argument(
-        "-i",
-        "--interactive",
-        action="store_true",
-        default=False,
-        help="Ask for confirmation before reproducing each stage.",
     )
     repro_parser.add_argument(
         "-p",
@@ -118,17 +104,24 @@ def add_arguments(repro_parser):
         help="Reproduce all stages in the specified directory.",
     )
     repro_parser.add_argument(
-        "--force-downstream",
+        "-m",
+        "--metrics",
         action="store_true",
         default=False,
-        help="Reproduce all descendants of a changed stage even if their "
-        "direct dependencies didn't change.",
+        help="Show metrics after reproduction.",
     )
     repro_parser.add_argument(
         "--downstream",
         action="store_true",
         default=False,
         help="Start from the specified stages when reproducing pipelines.",
+    )
+    repro_parser.add_argument(
+        "--force-downstream",
+        action="store_true",
+        default=False,
+        help="Reproduce all descendants of a changed stage even if their "
+        "direct dependencies didn't change.",
     )
     repro_parser.add_argument(
         "--pull",
@@ -138,6 +131,13 @@ def add_arguments(repro_parser):
             "Try automatically pulling missing cache for outputs restored "
             "from the run-cache."
         ),
+    )
+    repro_parser.add_argument(
+        "--dry",
+        action="store_true",
+        default=False,
+        help="Only print the commands that would be executed without "
+        "actually executing.",
     )
 
 
@@ -156,13 +156,10 @@ def add_parser(subparsers, parent_parser):
     add_arguments(repro_parser)
     # repro only args
     repro_parser.add_argument(
-        "--no-run-cache",
+        "--glob",
         action="store_true",
         default=False,
-        help=(
-            "Execute stage commands even if they have already been run with "
-            "the same command/dependencies/outputs/etc before."
-        ),
+        help="Allows targets containing shell-style wildcards.",
     )
     repro_parser.add_argument(
         "--no-commit",
@@ -171,9 +168,12 @@ def add_parser(subparsers, parent_parser):
         help="Don't put files/directories into cache.",
     )
     repro_parser.add_argument(
-        "--glob",
+        "--no-run-cache",
         action="store_true",
         default=False,
-        help="Allows targets containing shell-style wildcards.",
+        help=(
+            "Execute stage commands even if they have already been run with "
+            "the same command/dependencies/outputs/etc before."
+        ),
     )
     repro_parser.set_defaults(func=CmdRepro)
--- a/dvc/commands/run.py
+++ b/dvc/commands/run.py
@@ -70,10 +70,23 @@ def add_parser(subparsers, parent_parser):
         "--file", metavar="<filename>", help=argparse.SUPPRESS
     )
     run_parser.add_argument(
+        "--single-stage",
+        action="store_true",
+        default=False,
+        help=argparse.SUPPRESS,
+    )
+    _add_common_args(run_parser)
+    run_parser.add_argument(
         "--no-exec",
         action="store_true",
         default=False,
         help="Only create dvc.yaml without actually running it.",
+    )
+    run_parser.add_argument(
+        "--no-commit",
+        action="store_true",
+        default=False,
+        help="Don't put files/directories into cache.",
     )
     run_parser.add_argument(
         "--no-run-cache",
@@ -84,17 +97,4 @@ def add_parser(subparsers, parent_parser):
             "with the same command/dependencies/outputs/etc before."
         ),
     )
-    run_parser.add_argument(
-        "--no-commit",
-        action="store_true",
-        default=False,
-        help="Don't put files/directories into cache.",
-    )
-    run_parser.add_argument(
-        "--single-stage",
-        action="store_true",
-        default=False,
-        help=argparse.SUPPRESS,
-    )
-    _add_common_args(run_parser)
     run_parser.set_defaults(func=CmdRun)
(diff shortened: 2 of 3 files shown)


dvc/commands/repro.py
```diff
@@ -156,13 +156,10 @@ def add_parser(subparsers, parent_parser):
     add_arguments(repro_parser)
     # repro only args
     repro_parser.add_argument(
```
jorgeorpinel, line 158: @dberenbaum @daavoo to really match the order in https://github.com/iterative/dvc.org/pull/3345/files#diff-6c1f3192f09e2722ba169e9fa219b3b5158bbafa470b382c2d6135db7aa1e20d we'd need to change this logic, probably, affecting `exp run` as well. Should I go ahead with that?

jorgeorpinel: The alternative is to keep this order due to necessity and change it again in docs.

.pre-commit-config.yaml
```diff
@@ -5,6 +5,7 @@ repos:
   - hooks:
       - id: black
         language_version: python3
+        additional_dependencies: ['click==8.0.4']
```
jorgeorpinel, line 8: Had to change this bc/`black` was crashing for me locally. Maybe it's an issue for other people? I'm on a totally new Mac env.

Solution from https://github.com/psf/black/issues/2964#issuecomment-1080974737

pmrowla: This is an issue with black that has been resolved in the current release (`22.3.0`), this fix belongs in a different PR to update the black pre-commit rev

jorgeorpinel: So no need for this right?
```suggestion
  - hooks:
      - id: black
        language_version: python3
```

jorgeorpinel: I can confirm that black 22.3 no longer has this issue. Reverting...

See https://github.com/iterative/dvc/pull/7526 though

dvc/commands/remote.py
```diff
@@ -113,8 +113,8 @@ def run(self):
 class CmdRemoteList(CmdRemote):
     def run(self):
         conf = self.config.read(self.args.level)
-        for name, conf in conf["remote"].items():
-            ui.write(name, conf["url"], sep="\t")
+        for name, conf_val in conf["remote"].items():
+            ui.write(name, conf_val["url"], sep="\t")
```
jorgeorpinel, line 117: Sorry also unrelated... [Tests](https://github.com/iterative/dvc/runs/5765753773?check_suite_focus=true) were complaining about this (lint check), `for loop that reassigns the iterable it is iterating with each iterable value`.

dberenbaum:
Thanks for making this PR @jorgeorpinel! Besides the comment above, it looks like there are still some other arguments that don't match https://github.com/iterative/dvc.org/pull/3345. Could you take a look? I'm also happy to take this one over if you want.

jorgeorpinel:
@dberenbaum UPDATE: I'm actually trying to focus on other things. I don't mean to abandon this: I still plan to address the feedback ASAP. But feel free to take it over as well.

karajan1001 (approved):

karajan1001 (approved):

dvc/commands/repro.py
jorgeorpinel: > to really match the order [we agreed on](https://github.com/iterative/dvc.org/pull/3345/files#diff-6c1f3192f09e2722ba169e9fa219b3b5158bbafa470b382c2d6135db7aa1e20d) we'd need to change this logic

With that I meant that the `add_arguments(repro_parser)` function call groups a bunch of options and then a few others are added after that call so without changing or flattening that fn. in the current scope we can't match the agreed order.

> we should probably reorder it so that --glob comes later ... my suggestion would be this order

I like the [current order](https://dvc.org/doc/command-reference/repro#synopsis) more but up to you. I also don't think a bug should define the help output order or docs (but `--glob` could be hidden in `exp run` for now if needed).

> > having to change exp run code for consistency 
>
> we should also have updated exp run in https://github.com/iterative/dvc.org/pull/3345 for consistency

Also, I'm not sure about this any more! The `exp run` ref [already has](https://dvc.org/doc/command-reference/exp/run#synopsis) a `{ remote options ... }` line at the moment and all the other options are unique I think... Do we mean reorder those other options? I'll open a separate discussion on reordering `exp run` not to overcomplicate this review.

jorgeorpinel:
> we should also have updated `exp run` in https://github.com/iterative/dvc.org/pull/3345 for consistency

Do we mean the [options](https://dvc.org/doc/command-reference/exp/run#synopsis) after the `{ repro options ... }` line?

image
> Is that missing `-n`/`--name` BTW?

Well, we can update it here in the core and then make a follow-up to docs (maybe necessary anyway if the `repro` output is also re-ordered again in this PR). Here's my suggestion:

```
usage: dvc exp run [-h] [-q | -v] [-f]
               { repro options ... } [-n <name>]
               [-S [<filename>:]<params_list>]
               [--queue] [--run-all] [-j <number>] [--temp]
               [-r <experiment_rev>] [--reset]
               [targets [targets ...]]
```

dvc/commands/repro.py
dberenbaum: > > to really match the order [we agreed on](https://github.com/iterative/dvc.org/pull/3345/files#diff-6c1f3192f09e2722ba169e9fa219b3b5158bbafa470b382c2d6135db7aa1e20d) we'd need to change this logic
> 
> With that I meant that the `add_arguments(repro_parser)` function call groups a bunch of options and then a few others are added after that call so without changing or flattening that fn. in the current scope we can't match the agreed order. And I'm not 100$ sure why that function exists so I'm hesitant to touch it.

Sorry, I got what you meant but didn't explain my reasoning. That function exists to be reused in `exp run`, and I don't think we should break it up. Putting the common options (between `repro` and `exp run`) first not only simplifies the code but also provides some logical order and makes the help text more consistent. To do that, `--no-commit` and `--no-run-cache` need to be at the end (and technically `--glob`, see below). Given those constraints, is there an order you prefer?

> > we should probably reorder it so that --glob comes later ... my suggestion would be this order
> 
> I like the [current order](https://dvc.org/doc/command-reference/repro#synopsis) more but up to you. I also don't think a bug should define the help output or docs (but `--glob` could be hidden in `exp run` for now if needed).

True. Dealing with `--glob` might get a bit messy then in the code, but if you there's a good reason to put it in the middle of the other options, we can work around it.

> > > having to change exp run code for consistency
> > 
> > 
> > we should also have updated exp run in [iterative/dvc.org#3345](https://github.com/iterative/dvc.org/pull/3345) for consistency
> 
> Also, I'm not sure about this any more! The `exp run` ref [already has](https://dvc.org/doc/command-reference/exp/run#synopsis) a `{ remote options ... }` line at the moment i.e. it's consistent with `repro`, and all the other options are unique I think... Do we mean reorder those other options? I'll open a separate discussion on reordering `exp run` not to overcomplicate this review.

Yeah, I missed how we handle that in the docs. I think it's fine as is.

jorgeorpinel:
> there are still some other arguments that don't match https://github.com/iterative/dvc.org/pull/3345

I can't find any missing arguments in docs other than the ones surpassed in code.

dvc/commands/run.py
```diff
@@ -70,19 +70,17 @@ def add_parser(subparsers, parent_parser):
...
     )
     run_parser.add_argument(
-        "--no-exec",
+        "--single-stage",
         action="store_true",
         default=False,
-        help="Only create dvc.yaml without actually running it.",
+        help=argparse.SUPPRESS,
     )
+    _add_common_args(run_parser)
```
jorgeorpinel, line 78: BTW shouldn't `-n|--name` be in `_add_common_args()`? @karajan1001

dvc/commands/repro.py
jorgeorpinel: > That function exists to be reused in exp run, and I don't think we should break it up. Putting the common options (between repro and exp run) first not only simplifies the code but also provides some logical order and makes the help text more consistent.

Got it, OK. And I see that `--glob` is not in it as of now so fine, let's leave the 3 repro-only ones ([mentioned](https://dvc.org/doc/command-reference/exp/run#options) in the `exp run` ref.) at the end then, but not sure about the other changes in your suggestion. Maybe

```ini
usage: dvc repro [-h] [-q | -v] [-f] [-i]
                 [-s] [-p] [-P] [-R] [-m]
                 [--downstream] [--force-downstream]
                 [--pull] [--dry]  # --glob can be moved up here later.
                 [--glob] [--no-commit] [--no-run-cache]
                 [targets [<target> ...]]
```

> Will suggest in https://github.com/iterative/dvc.org/pull/3435 too.

dvc/commands/run.py
karajan1001: I think we can

> BTW shouldn't `-n|--name` be added in `_add_common_args()`? It's a common arg both in `stage add` and `run`. Cc @karajan1001

I think we can do that.

> Also (💅🏼) should add_arguments() in repro/exp run use a similar private method name instead? _add_common_args() as well, even

I'm not quite understand here why this method to be a private method as it is used outside its file's scope.

dberenbaum: > BTW shouldn't `-n|--name` be added in `_add_common_args()`? It's a common arg both in `stage add` and `run`. Cc @karajan1001

Resolved.

dberenbaum: Never mind, rolling this change back. It breaks a bunch of tests. I think they were intentionally left separate to support some behavior that was deprecated but not actually removed (like `dvc run --single-stage`). Let's stick with the UI changes.

dberenbaum:
@jorgeorpinel @karajan1001 Could you please review the latest updates?

karajan1001 (approved):

jorgeorpinel (commented):
Well, `run` is not exactly like https://github.com/iterative/dvc.org/pull/3345/files#diff-6c1f3192f09e2722ba169e9fa219b3b5158bbafa470b382c2d6135db7aa1e20d because of the way it's coded but I think we agreed to let that go. Everything else does match https://github.com/iterative/dvc.org/pull/3345 so please merge!

dberenbaum:
> Well, `run` is not exactly like [in docs](https://github.com/iterative/dvc.org/pull/3345/files#diff-6c1f3192f09e2722ba169e9fa219b3b5158bbafa470b382c2d6135db7aa1e20d) because of the way it's coded but I think we agreed to let that go. Everything else does match [iterative/dvc.org#3345](https://github.com/iterative/dvc.org/pull/3345) so please merge!

Thanks, @jorgeorpinel! Based on the comments in this PR, we updated the docs in https://github.com/iterative/dvc.org/pull/3435, and they should all match now AFAIK.

PR #7527, opened 2022-03-31
setup.cfg: tests: try to remove more dependencies

skshetry:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=48", "wheel", "setuptools_scm[toml]>=6.3.1"]
+requires = ["setuptools>=48", "setuptools_scm[toml]>=6.3.1", "setuptools_scm_git_archive==1.1"]
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools_scm]
--- a/setup.cfg
+++ b/setup.cfg
@@ -121,7 +121,6 @@ webdhfs_kerberos = requests-kerberos==0.14.0
 terraform = tpi[ssh]>=2.1.0
 tests =
     %(terraform)s
-    wheel==0.37.1
     dvc_ssh==0.0.1a0
     # Test requirements
     pytest==7.1.1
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=48", "wheel", "setuptools_scm[toml]>=6.3.1"]
+requires = ["setuptools>=48", "setuptools_scm[toml]>=6.3.1", "setuptools_scm_git_archive==1.1"]
```
skshetry, line 2: See https://github.com/pypa/setuptools_scm/pull/689

setup.cfg
```diff
@@ -121,7 +121,6 @@ webdhfs_kerberos = requests-kerberos==0.14.0
 terraform = tpi[ssh]>=2.1.0
 tests =
     %(terraform)s
-    wheel==0.37.1
```
skshetry, line 124: `wheel` is a build-time dependency for us. Using with `pip install -e ".[dev]"` will not install wheel-packages immediately, it only takes effect afterwards.

setup.cfg
```diff
@@ -135,9 +134,6 @@ tests =
     mock==4.0.3
     pytest-timeout==2.1.0
     filelock==3.6.0
-    dvclive[image]==0.6.0
```
skshetry, line 138: `dvclive` is an optional dependency, as seen from `tests`.

setup.cfg
```diff
@@ -135,9 +134,6 @@ tests =
...
-    collective.checkdocs==0.2
-    pydocstyle==6.1.1
```
skshetry, line 140: `pydocstyle` is a transitive dependency.

setup.cfg
```diff
@@ -135,9 +134,6 @@ tests =
     mock==4.0.3
     pytest-timeout==2.1.0
     filelock==3.6.0
-    dvclive[image]==0.6.0
-    collective.checkdocs==0.2
```
skshetry, line 139: Moved to be installed during GHA.

setup.cfg
```diff
@@ -147,11 +143,12 @@ tests =
     pylint-plugin-utils==0.7
     # type-checking
     mypy==0.942
-    types-requests==2.27.15
-    types-tabulate==0.8.6
-    types-toml==0.10.4
+    types-requests>=2.27.15
+    types-tabulate>=0.8.6
+    types-toml>=0.10.4
```
skshetry, line 148: We don't really care that much about specific versions of these, will reduce a bit of noise with dependabot.

Issue #7531, opened 2022-03-31
pull: DVC installed via Ubuntu .deb fails to pull from S3 remote but DVC installed via pip works


rightparen:
# Bug Report

## Description

In Circle CI, the command `dvc pull -f assets/models` began failing between Mar 30 4:18:21 PM PDT and Mar 30 9:28:58 PM PDT when using the Ubuntu `.deb` from https://dvc.org/deb/dvc.list.  All commits in the repo fail with the same error now, and the main branch has not changed since all commits succeeded.

The same command with the same Circle CI environment and same source tree passes if we use `pip install dvc[s3]` to install DVC.

The output in the log when it fails is:
```
A       assets/models/
1 file added and 12 files fetched
Traceback (most recent call last):
  File "PyInstaller/loader/pyimod03_importers.py", line 298, in get_code
  File "PyInstaller/loader/pyimod02_archive.py", line 287, in extract
TypeError: unhashable type: 'dict'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "dvc/__main__.py", line 7, in <module>
  File "dvc/cli/__init__.py", line 121, in main
  File "dvc/analytics.py", line 37, in collect_and_send_report
  File "dvc/daemon.py", line 109, in daemon
  File "inspect.py", line 1514, in stack
  File "inspect.py", line 1491, in getouterframes
  File "inspect.py", line 1465, in getframeinfo
  File "inspect.py", line 792, in findsource
  File "inspect.py", line 745, in getmodule
  File "importlib/util.py", line 245, in __getattribute__
  File "PyInstaller/loader/pyimod03_importers.py", line 478, in exec_module
  File "PyInstaller/loader/pyimod03_importers.py", line 300, in get_code
TypeError: can only concatenate str (not "dict") to str
[2151] Failed to execute script '__main__' due to unhandled exception!
```

### Reproduce

In Circle CI, install DVC using Ubuntu `.deb` package from dvc.org:
```
    steps:
      - run:
          # DVC CLI is required to download production models from our DVC repository.
          name: Install DVC
          command: |
            sudo wget https://dvc.org/deb/dvc.list -O /etc/apt/sources.list.d/dvc.list
            # TODO: commit iterative.asc key to repo rather than downloading from dvc.org
            wget -qO - https://dvc.org/deb/iterative.asc | sudo apt-key add -
            sudo apt update
            sudo apt install dvc
            dvc version
```

[...]

PR #7545, opened 2022-04-05
build: workaround pyinstaller `inspect.getmodule` issue

pmrowla:

Will fix https://github.com/iterative/dvc/issues/7531


dvc/cli/__init__.py
```diff
@@ -48,6 +49,11 @@ def main(argv=None):  # noqa: C901
     from dvc.exceptions import DvcException, NotDvcRepoError
     from dvc.logger import FOOTER, disable_other_loggers
 
+    if sys.stderr.closed:  # pylint: disable=using-constant-test
+        logging.disable()
+    elif sys.stdout.closed:  # pylint: disable=using-constant-test
+        logging.disable(logging.INFO)
```
pmrowla, line 55: in `dvc.daemon` we double fork and close the standard file descriptors (to avoid duplicating the parent's FDs in the child daemon). On Linux, we never actually call `exec` after the double fork, we just directly call `cli.main`, which results in stdout and stderr still being closed in the daemon process. Making any logging calls in this state results in an exception due to writing to a closed file descriptor.

In normal use this leads to a silent failure when we run the DVC analytics or updater daemon in verbose/debug mode, but in pyinstaller builds this causes pyinstaller to generate an unhandled exception error message that gets printed to the original parent terminal.

skshetry: @pmrowla, can you please add this as a comment in the code itself? Thanks.

skshetry (approved):

dvc/cli/__init__.py
daavoo: Why this didn't happen before?

pmrowla: @daavoo I think it probably did and we just never noticed, because we don't run any thorough tests with the pyinstaller builds.

In pip-installed DVC, we won't ever see this exception because it just fails silently (since there is no stderr/stdout to log the exception to, the python interpreter will just exit). It gets caught and logged in the pyinstaller builds since pyinstaller's separate bootloader is still running separately from the pyinstaller package's python interpreter

pmrowla: Even if we did have more end-to-end tests, catching this would still be difficult because it requires watching the system for spawned python (or pyinstaller) processes, figuring out which one is the daemonized child process, and checking exit codes when that process terminates.

dvc/cli/__init__.py
```diff
@@ -48,6 +49,16 @@ def main(argv=None):  # noqa: C901
...
 
+    # NOTE: stderr/stdout may be closed if we are running from dvc.daemon.
+    # On Linux we directly call cli.main after double forking and closing
+    # the copied parent's standard file descriptors. If we make any logging
+    # calls in this state it will cause an exception due to writing to a closed
+    # file descriptor.
+    if sys.stderr.closed:  # pylint: disable=using-constant-test
+        logging.disable()
+    elif sys.stdout.closed:  # pylint: disable=using-constant-test
+        logging.disable(logging.INFO)
```
skshetry, line 60: What do you think of reducing the scope of this to only when we are running in the daemon mode, i.e. check if `DVC_DAEMON` is set?

pmrowla: Technically, users could run DVC in background processes in this state (without using `dvc daemon`), so I guess it depends on whether or not we think that should be an error condition.

dtrifiro (approved):

PR #7554, opened 2022-04-07
erepo: include exp refs in clones

### LINKED ISSUE #7169: get: make it work for experiments and/or improve message
`dvc get --rev exp-name`
`dvc get --rev exp-ref`

should work and I should be able to get artifacts, datasets w/o doing heavy apply.

Now the message:

```
$ dvc get example-get-started/ model.pkl --rev 73c7c57
ERROR: failed to get 'model.pkl' from 'example-get-started/' - '404 Not Found'
```

is confusing. So, even if for some reason we decide it's not needed we have to improve the message and also solve https://github.com/iterative/dvc/issues/5355 

**Use cases:**

- I want to get a few models to compare them side by side now in the workspace
- I want to get a model/checkpoint w/o affecting my workspace
- I want to get a dataset w/o doing full apply

To me it felt very natural to expect, and I though this is a bug.

**More context from Slack:**

@pmrowla 
> exp names are already supposed to be accepted anywhere in DVC that supports --rev, the issue with dvc get is probably that we do not pull exp refs when we make the erepo clone. I don’t think we have a ticket for this right now

So, this can be considered a bug?

pmrowla:

- include exp refs when cloning/updating full (non-shallow) erepo clones
- add test for `dvc get --rev=<exp>`

Will fix #7169

### DIFF
--- a/dvc/repo/experiments/utils.py
+++ b/dvc/repo/experiments/utils.py
@@ -191,3 +191,16 @@ def check_ref_format(scm: "Git", ref: ExpRefInfo):
             f"Invalid exp name {ref.name}, the exp name must follow rules in "
             "https://git-scm.com/docs/git-check-ref-format"
         )
+
+
+def fetch_all_exps(scm: "Git", url: str, progress: Optional[Callable] = None):
+    refspecs = [
+        f"{ref}:{ref}"
+        for ref in iter_remote_refs(scm, url, base=EXPS_NAMESPACE)
+        if not (ref.startswith(EXEC_NAMESPACE) or ref == EXPS_STASH)
+    ]
+    scm.fetch_refspecs(
+        url,
+        refspecs,
+        progress=progress,
+    )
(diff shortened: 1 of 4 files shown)


dvc/repo/experiments/utils.py
```diff
@@ -191,3 +191,16 @@ def check_ref_format(scm: "Git", ref: ExpRefInfo):
             f"Invalid exp name {ref.name}, the exp name must follow rules in "
             "https://git-scm.com/docs/git-check-ref-format"
         )
+
+
+def fetch_all_exps(scm: "Git", url: str, progress: Optional[Callable] = None):
```
karajan1001, line 196: Excuse me, do we need to download all exps instead of only the experiment listed after `--rev`?

pmrowla: Yes, this is intended to make clone behave the same for exp refs as it behaves for `--no-single-branch` - it will fetch all available branches and tags (so we want to fetch all available experiment refs as well).

karajan1001 (approved):

PR #7555, opened 2022-04-07
Add `getitem` to Index

PythonFZ:
# Summary

 Allow access to a stage via `getitem` in addition to iterating over the Index.

It is currently possible to iterate over `dvc.repo.Index` but you can not get a `Stage` by its stagename.
This PR introduces `getitem` supporting the name of the Node as follows:

```py
index = Index()
stage: Stage = index["stagename"]
```

This can be handy for all operations attached to `Stage` such as looking at dependecies, etc.

### DIFF
--- a/dvc/repo/index.py
+++ b/dvc/repo/index.py
@@ -93,7 +93,7 @@ class Index:
 
     def __getitem__(self, item: str) -> "Stage":
         """Get a stage by its addressing attribute."""
-        for stage in self.stages:
+        for stage in self:
             if stage.addressing == item:
                 return stage
         raise KeyError(f"{item} - available stages are {self.stages}")
--- a/dvc/repo/index.py
+++ b/dvc/repo/index.py
@@ -92,7 +92,7 @@ class Index:
         yield from self.stages
 
     def __getitem__(self, item: str) -> "Stage":
-        """Get a Stage by its addressing attribute"""
+        """Get a stage by its addressing attribute."""
         for stage in self.stages:
             if stage.addressing == item:
                 return stage
--- a/dvc/repo/index.py
+++ b/dvc/repo/index.py
@@ -91,6 +91,13 @@ class Index:
     def __iter__(self) -> Iterator["Stage"]:
         yield from self.stages
 
+    def __getitem__(self, item: str) -> "Stage":
+        """Get a Stage by its addressing attribute"""
+        for stage in self.stages:
+            if stage.addressing == item:
+                return stage
+        raise KeyError(f"{item} - available stages are {self.stages}")
+
     def filter(self, filter_fn: Callable[["Stage"], bool]) -> "Index":
         stages_it = filter(filter_fn, self)
         return Index(self.repo, self.fs, stages=list(stages_it))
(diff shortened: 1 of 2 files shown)


dvc/repo/index.py
```diff
@@ -91,6 +91,13 @@ def __contains__(self, stage: "Stage") -> bool:
     def __iter__(self) -> Iterator["Stage"]:
         yield from self.stages
 
+    def __getitem__(self, item: str) -> "Stage":
+        """Get a Stage by its addressing attribute"""
+        for stage in self.stages:
```
skshetry, line 96:
```suggestion
        for stage in self:
```

dvc/repo/index.py
```diff
@@ -91,6 +91,13 @@ def __contains__(self, stage: "Stage") -> bool:
     def __iter__(self) -> Iterator["Stage"]:
         yield from self.stages
 
+    def __getitem__(self, item: str) -> "Stage":
+        """Get a Stage by its addressing attribute"""
```
skshetry, line 95:
```suggestion
        """Get a stage by its addressing attribute."""
```

skshetry:
@PythonFZ, could you please share more details about your usecase?

dvc/repo/index.py
skshetry: We could also think of making `self.stages` a `dict` in the future, but since we don't use it in DVC, it's fine for now.

PythonFZ:
> @PythonFZ, could you please share more details about your usecase/usages?

There are two potential use cases where this would help me.
- I'm developing the [ZnTrack](https://github.com/zincware/ZnTrack) package which currently relies on subprocess calls but I would like to use the DVC python API if possible. It would help to access the `Stage` class by the name of the Node. 
- I saw on Twitter that [Aim](https://github.com/aimhubio/aim) now has a preliminary support for DVC and here this could also be helpful for easy access to sage dependencies based on the stage name.

I found myself writing the following

```py
for stage in self.stages:
     if stage.addressing == stage_name:
          return stage
```

and thought this might fit nicely to the `Index` class.

I'm not very familiar with the DVC Python API, is there a better or more intended way to achieve what I'm trying?

skshetry:
> I'm not very familiar with the DVC Python API, is there a better or more intended way to achieve what I'm trying?

There is not really a nice way to do this in DVC Python API as we rarely need to work with one single stage. For performance reasons, whenever we need to get one single stage, we just try to load from the user's workspace.

If you only need one particular stage, you can use [`repo.stage.collect()`](https://github.com/iterative/dvc/blob/03f553c54bf0e6b91626a87309d817edce61282a/dvc/repo/stage.py#L320), but it tries to load from user's workspace instead of what we have in memory:
https://github.com/iterative/dvc/blob/03f553c54bf0e6b91626a87309d817edce61282a/dvc/repo/stage.py#L320

`Repo.stage` is a lower level utility for loading and [creating stages (`repo.stage.add`/`repo.stage.create`)](https://github.com/iterative/dvc/blob/03f553c54bf0e6b91626a87309d817edce61282a/dvc/repo/stage.py#L117).

`Index` on the other hand is a relatively new concept, and we haven't been able to use it everywhere due to performance reasons as `Index` goes on to collect every `dvc.yaml` and `.dvc` files present in user's repository.
 In that case we drop to `Repo.stage` helper.

skshetry (approved):

PR #7579, opened 2022-04-16
parsing: use relative paths; force absolute paths to relative

### LINKED ISSUE #7307: import: does not work with repositories using params
Importing some data from parametrized repository does not work.

reproduction:
```bash
#!/bin/bash

set -ex

pushd $TMPDIR

wsp=test_wspace
imp_repo=import_repo
rep=test_repo

base=$(pwd)
mkdir $base/$imp_repo
mkdir $base/$rep

pushd $base/$imp_repo
git init
dvc init

echo data >> data
echo "m: 1" >> params.yaml

dvc add data
dvc run -d data -p m -o out -n train cp data out
git add -A
git commit -am "init"

pushd $base/$rep
git init
dvc init

dvc import $base/$imp_repo out -o my_out -v
```

Result:

```
dvc.parsing.ResolveError: failed to parse 'vars' in 'dvc.yaml': 'params.yaml' does not exist
```

On the first sight the culprit seems to be `dvc/repo/stage.py::_maybe_collect_from_dvc_yaml` which passes `dvc.yaml` to `StageLoad.load_all`, while it seems that it should be a path to the file.
https://github.com/iterative/dvc/blob/d882e21b9b013f69abff960a8bad62d822024b59/dvc/repo/stage.py#L66

Kudos to @oslobowl for discovering the problem.

skshetry:
`parsing.Context` now only expects the path to always be relative path. To clarify, it's relative if it's a `LocalFileSystem` and relatively-absolute (relative to the root of the repo) in GitFileSystem, as opposed to being absolutely-absolute paths.

This expectation for Context to always get a relative path is managed in `DataResolver` for now, as changing anything in higher level may not be trivial.

This change, that it uses the relative path in `LocalFileSystem` and relatively-absolute paths in `GitFileSystem` does affect `dvc.params.diff()`. Before we were using same relative/absolute paths, which worked for both filesystems, but now we will receive the data in something like following format:
```json
{
    "workspace": {
        "data": {
            "../test_params.yaml": {  # relative path from LocalFS
                "data": {
                    "vars.model1.epoch": 20,
                    "vars.model2.epoch": 35
                }
            }
        }
    },
    "HEAD": {
        "data": {
            "test_params.yaml": {  # From GitFS as it is relative from the root of the repo
                "data": {
                    "vars.model1.epoch": 15,
                    "vars.model2.epoch": 35
                }
            }
        }
    }
}
```

So, as a result of this, this had to be transformed such that they return same paths, so that we can compare them easily.

Fix #7307.

### DIFF
--- a/dvc/parsing/__init__.py
+++ b/dvc/parsing/__init__.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from collections.abc import Mapping, Sequence
 from copy import deepcopy
 from typing import (
@@ -134,9 +135,19 @@ def make_definition(
 
 class DataResolver:
     def __init__(self, repo: "Repo", wdir: str, d: dict):
+        from dvc.fs import LocalFileSystem
+
         self.fs = fs = repo.fs
+
+        if os.path.isabs(wdir):
+            start = (
+                os.curdir if isinstance(fs, LocalFileSystem) else repo.root_dir
+            )
+            wdir = relpath(wdir, start)
+            wdir = "" if wdir == os.curdir else wdir
+
         self.wdir = wdir
-        self.relpath = relpath(fs.path.join(self.wdir, "dvc.yaml"))
+        self.relpath = os.path.normpath(fs.path.join(self.wdir, "dvc.yaml"))
 
         vars_ = d.get(VARS_KWD, [])
         check_interpolations(vars_, VARS_KWD, self.relpath)
--- a/dvc/parsing/context.py
+++ b/dvc/parsing/context.py
@@ -19,7 +19,6 @@ from dvc.parsing.interpolate import (
     recurse,
     str_interpolate,
 )
-from dvc.utils import relpath
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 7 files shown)


efiop:
@skshetry Any thoughts on how this would behave itself with moving gitfs to root paths?

skshetry:
> @skshetry Any thoughts on how this would behave itself with moving gitfs to root paths? Asking in case you've looked into that 🙂

It's hard to say for all stage loading helpers, but this is how it already works with root paths of gitfs. The following code is trying to transform filesystem-absolute paths into root paths in the case of `GitFileSystem`:
https://github.com/iterative/dvc/blob/5d4d8585744bbe8163a32c9f78c9bb681b4eba56/dvc/parsing/__init__.py#L142-L147

efiop (approved):

efiop:
@skshetry Cool. With this and repofs, we can now try to use https://github.com/iterative/scmrepo/pull/56

PR #7593, opened 2022-04-19
dvcignore: use fs.path instead of os.path

efiop:
We've been trying to avoid using `os.path.abspath` before, but we actually do need it in dvcignore for path normalization. So instead, we are introducing a primitive notion of `cwd` into our filesystems, that will be useful in other applications.

Note that even though fsspec doesn't officially support `cwd`, localfs does support it but in implicit manner, which requires us to have odd checks here and there around the code, which `Path.getcwd` will reduce in the future.

### DIFF
--- a/dvc/ignore.py
+++ b/dvc/ignore.py
@@ -12,7 +12,6 @@ from dvc.fs.local import localfs
 from dvc.pathspec_math import PatternInfo, merge_patterns
 from dvc.scheme import Schemes
 from dvc.types import AnyPath, List, Optional
-from dvc.utils import relpath
 from dvc.utils.collections import PathStringTrie
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


dvc/ignore.py
```diff
@@ -343,47 +343,40 @@ def _is_ignored(
...
-        path = relpath(path, self.root_dir)
-        if path.startswith("..") or (
-            os.name == "nt"
-            and not os.path.commonprefix(
-                [os.path.abspath(path), self.root_dir]
-            )
-        ):
-            return True
-        return False
+        return not self.fs.path.isin_or_eq(path, self.root_dir)
```
efiop, line 367: The only major logical change.

karajan1001 (approved):
LGTM.

PR #7597, opened 2022-04-19
dvc.data: save and try loading raw dir objects

### LINKED ISSUE #7390: status: recalculating hashes each call
When calling `status` on modified dir, DVC shows message `Computing file/dir hashes (only done once)` for each status call.
Reproduction script:

```bash
#!/bin/bash

set -exu

pushd $TMPDIR

wsp=test_wspace
rp=test_repo

mkdir $rp && pushd $rp

git init --quiet
dvc init --quiet
dvc config core.checksum_jobs 1

git add -A
git commit -am "init"

mkdir data

for i in {1..2000}
do
	dd if=/dev/random of=data/file$i bs=1k count=10 >& /dev/null
done

dvc add data

echo modification >> data/another_file

dvc status 
dvc status 
```
It might not necessarily be an issue with DVC, as "calculation" for subsequent status calls seems to be faster than the original one. Maybe it's just a problem with a message.

dtrifiro:
When staging a directory, always save a "raw dir object" to the odb. If the corresponding ".dir" object has not been added to the odb, `stage()` calls can load the tree from the raw dir object instead of rebuilding it by walking the directory.
This can lead to significant speed improvements when calling `dvc status` for modified directories.

Fixes #7390

### DIFF
--- a/dvc/data/stage.py
+++ b/dvc/data/stage.py
@@ -1,3 +1,4 @@
+import errno
 import hashlib
 import logging
 import os
@@ -18,6 +19,9 @@ from .db.reference import ReferenceObjectDB
 if TYPE_CHECKING:
     from dvc.fs.base import AnyFSPath, FileSystem
     from dvc.objects.db import ObjectDB
+
+    from .tree import Tree
+
 
 logger = logging.getLogger(__name__)
 
@@ -233,7 +237,7 @@ def _make_staging_url(
     return url
 
 
-def _get_staging(odb: "ObjectDB") -> "ObjectDB":
+def _get_staging(odb: "ObjectDB") -> "ReferenceObjectDB":
     """Return an ODB that can be used for staging objects.
 
     Staging will be a reference ODB stored in the the global memfs.
@@ -247,7 +251,33 @@ def _get_staging(odb: "ObjectDB") -> "ObjectDB":
     return ReferenceObjectDB(fs, fs_path, state=state)
 
 
-def _load_from_state(odb, staging, fs_path, fs, name):
+def _load_raw_dir_obj(odb: "ObjectDB", hash_info: "HashInfo") -> "Tree":
+    from dvc.objects.errors import ObjectFormatError
+
+    from .tree import Tree
+
+    try:
+        tree = Tree.load(odb, hash_info.as_raw())
+        tree.check(odb)
+        tree.hash_info = hash_info
+    except ObjectFormatError as exc:
+        raise FileNotFoundError(
+            errno.ENOENT,
+            "No such object",
+            odb.hash_to_path(hash_info.as_raw().value),
+        ) from exc
+
+    return tree
+
+
+def _load_from_state(
+    odb: "ObjectDB",
+    staging: "ReferenceObjectDB",
+    fs_path: "AnyFSPath",
+    fs: "FileSystem",
+    name: str,
+    dry_run: bool,
+) -> Tuple["ObjectDB", "Meta", "HashFile"]:
     from dvc.objects.errors import ObjectFormatError
 
     from . import check, load
@@ -255,19 +285,56 @@ def _load_from_state(odb, staging, fs_path, fs, name):
 
     state = odb.state
     meta, hash_info = state.get(fs_path, fs)
-    if hash_info:
-        for odb_ in (odb, staging):
-            if odb_.exists(hash_info):
-                try:
-                    obj = load(odb_, hash_info)
-                    check(odb_, obj, check_hash=False)
-                    if isinstance(obj, Tree):
-                        meta.nfiles = len(obj)
-                    assert obj.hash_info.name == name
-                    return odb_, meta, obj
-                except (ObjectFormatError, FileNotFoundError):
-                    pass
-    raise FileNotFoundError
+    if not hash_info:
+        raise FileNotFoundError
+
+    for odb_ in (odb, staging):
+        if not odb_.exists(hash_info):
+            continue
+
+        try:
+            obj = load(odb, hash_info)
+            check(odb, obj, check_hash=False)
+        except (ObjectFormatError, FileNotFoundError):
+            continue
+
+        if isinstance(obj, Tree):
+            meta.nfiles = len(obj)
+        assert obj.hash_info.name == name
+        return odb_, meta, obj
+
+    if not hash_info.isdir:
+        raise FileNotFoundError
+
+    # Try loading the raw dir object saved by `stage`, see below and #7390
+    tree = _load_raw_dir_obj(odb, hash_info)
+    meta.nfiles = len(tree)
+    assert tree.hash_info.name == name
+
+    if not dry_run:
+        assert tree.fs
+        for key, _, oid in tree:
+            staging.add(
+                fs.path.join(fs_path, *key),
+                fs,
+                oid,
+                hardlink=False,
+                verify=False,
+            )
+
+        staging.add(
+            tree.fs_path,
+            tree.fs,
+            hash_info,
+            hardlink=False,
+        )
+
+        raw = staging.get(hash_info)
+        tree.fs = raw.fs
+        tree.fs_path = raw.fs_path
+
+    logger.debug("loaded tree '%s' from raw dir obj", tree)
+    return staging, meta, tree
 
 
 def _stage_external_tree_info(odb, tree, name):
@@ -318,7 +385,7 @@ def stage(
     staging = _get_staging(odb)
     if odb:
         try:
-            return _load_from_state(odb, staging, fs_path, fs, name)
+            return _load_from_state(odb, staging, fs_path, fs, name, dry_run)
         except FileNotFoundError:
             pass
 
@@ -336,6 +403,12 @@ def stage(
         logger.debug("staged tree '%s'", obj)
         if name != "md5":
             obj = _stage_external_tree_info(odb, obj, name)
+
+        # In order to avoid re-building the tree when it is not committed to
+        # the local odb (e.g. for a status call), we save it as a raw object.
+        # Loading this instead of building the tree can speed up `dvc status`
+        # for modified directories, see #7390
+        odb.add(obj.fs_path, obj.fs, obj.hash_info.as_raw())
     else:
         _, meta, obj = _stage_file(
             fs_path,
--- a/tests/func/test_commit.py
+++ b/tests/func/test_commit.py
@@ -177,10 +177,13 @@ def test_commit_granular_dir(tmp_dir, dvc):
 
     cache = tmp_dir / ".dvc" / "cache"
 
-    assert set(cache.glob("*/*")) == set()
+    assert set(cache.glob("*/*")) == {
+        cache / "1a" / "ca2c799df82929bbdd976557975546",
+    }
 
     dvc.commit(os.path.join("data", "foo"))
     assert set(cache.glob("*/*")) == {
+        cache / "1a" / "ca2c799df82929bbdd976557975546",
         cache / "1a" / "ca2c799df82929bbdd976557975546.dir",
         cache / "ac" / "bd18db4cc2f85cedef654fccc4a4d8",
     }
@@ -188,6 +191,8 @@ def test_commit_granular_dir(tmp_dir, dvc):
 
     dvc.commit(os.path.join("data", "subdir"))
     assert set(cache.glob("*/*")) == {
+        cache / "26" / "d6b64d96a660707412f523e8184b5f",
+        cache / "1a" / "ca2c799df82929bbdd976557975546",
         cache / "1a" / "ca2c799df82929bbdd976557975546.dir",
         cache / "ac" / "bd18db4cc2f85cedef654fccc4a4d8",
         cache / "4c" / "e8d2a2cf314a52fa7f315ca37ca445",
@@ -197,6 +202,8 @@ def test_commit_granular_dir(tmp_dir, dvc):
 
     dvc.commit(os.path.join("data"))
     assert set(cache.glob("*/*")) == {
+        cache / "26" / "d6b64d96a660707412f523e8184b5f",
+        cache / "1a" / "ca2c799df82929bbdd976557975546",
         cache / "1a" / "ca2c799df82929bbdd976557975546.dir",
         cache / "ac" / "bd18db4cc2f85cedef654fccc4a4d8",
         cache / "4c" / "e8d2a2cf314a52fa7f315ca37ca445",
--- a/tests/func/test_gc.py
+++ b/tests/func/test_gc.py
@@ -20,13 +20,16 @@ class TestGC(TestDvcGit):
         super().setUp()
 
         self.dvc.add(self.FOO)
-        self.dvc.add(self.DATA_DIR)
+        stages = self.dvc.add(self.DATA_DIR)
+        raw_dir_hash = stages[0].outs[0].hash_info.as_raw().value
+
         self.good_cache = [
             self.dvc.odb.local.hash_to_path(md5)
             for md5 in self.dvc.odb.local.all()
+            if md5 != raw_dir_hash
         ]
 
-        self.bad_cache = []
+        self.bad_cache = [self.dvc.odb.local.hash_to_path(raw_dir_hash)]
         for i in ["123", "234", "345"]:
             path = os.path.join(self.dvc.odb.local.cache_dir, i[0:2], i[2:])
             self.create(path, i)
@@ -203,7 +206,7 @@ def test_gc_no_dir_cache(tmp_dir, dvc):
     with pytest.raises(CollectCacheError):
         dvc.gc(workspace=True)
 
-    assert _count_files(dvc.odb.local.cache_dir) == 4
+    assert _count_files(dvc.odb.local.cache_dir) == 5
     dvc.gc(force=True, workspace=True)
     assert _count_files(dvc.odb.local.cache_dir) == 2
 
--- a/tests/func/test_remote.py
+++ b/tests/func/test_remote.py
@@ -154,6 +154,9 @@ def test_dir_hash_should_be_key_order_agnostic(tmp_dir, dvc):
         _, _, obj = stage(dvc.odb.local, path, dvc.odb.local.fs, "md5")
         hash1 = obj.hash_info
 
+    # remove the raw dir obj to force building the tree on the next stage call
+    dvc.odb.local.fs.remove(dvc.odb.local.hash_to_path(hash1.as_raw().value))
+
     tree = Tree.from_list(
         [{"md5": "1", "relpath": "1"}, {"md5": "2", "relpath": "2"}]
     )
--- a/dvc/data/tree.py
+++ b/dvc/data/tree.py
@@ -140,7 +140,7 @@ class Tree(HashFile):
         return tree
 
     @classmethod
-    def load(cls, odb, hash_info):
+    def load(cls, odb, hash_info) -> "Tree":
         obj = odb.get(hash_info)
 
         try:
--- a/dvc/objects/hash_info.py
+++ b/dvc/objects/hash_info.py
@@ -41,3 +41,9 @@ class HashInfo:
         if not self:
             return False
         return self.value.endswith(HASH_DIR_SUFFIX)
+
+    def as_raw(self) -> "HashInfo":
+        assert self.value
+        return HashInfo(
+            self.name, self.value.rsplit(HASH_DIR_SUFFIX)[0], self.obj_name
+        )
--- a/tests/func/test_external_repo.py
+++ b/tests/func/test_external_repo.py
@@ -219,6 +219,7 @@ def test_subrepos_are_ignored(tmp_dir, erepo_dir):
             hardlink=True,
         )
         assert set(cache_dir.glob("??/*")) == {
+            cache_dir / "e1" / "d9e8eae5374860ae025ec84cfd85c7",
             cache_dir / "e1" / "d9e8eae5374860ae025ec84cfd85c7.dir",
             cache_dir / "37" / "b51d194a7513e45b56f6524f2d51f2",
             cache_dir / "94" / "7d2b84e5aa88170e80dff467a5bfb6",
--- a/tests/func/test_odb.py
+++ b/tests/func/test_odb.py
@@ -206,6 +206,9 @@ def test_shared_cache(tmp_dir, dvc, group):
     expected = {
         os.path.join(cache_dir, "17"): dir_mode,
         os.path.join(
+            cache_dir, "17", "4eaa1dd94050255b7b98a7e1924b31"
+        ): file_mode,
+        os.path.join(
             cache_dir, "17", "4eaa1dd94050255b7b98a7e1924b31.dir"
         ): file_mode,
         os.path.join(cache_dir, "97"): dir_mode,
--- a/tests/func/test_stage.py
+++ b/tests/func/test_stage.py
@@ -13,6 +13,7 @@ from dvc.stage.run import run_stage
 from dvc.utils.serialize import dump_yaml, load_yaml
 from dvc.utils.strictyaml import YAMLValidationError
 from tests.basic_env import TestDvc
+from tests.utils import clean_staging
 
 
 def test_cmd_obj():
@@ -321,3 +322,62 @@ def test_stage_run_checkpoint(tmp_dir, dvc, mocker, checkpoint):
     mock_cmd_run.assert_called_with(
         stage, checkpoint_func=callback, dry=False, run_env=None
     )
+
+
+@pytest.mark.parametrize(
+    "dry_run, expected_staging_contents",
+    [
+        (True, set()),
+        (
+            False,
+            {
+                "37b51d194a7513e45b56f6524f2d51f2",
+                "568f3dd88592a68ef99459a5491011cd",
+                "68dde2c3c4e7953c2290f176bbdc9a54",
+                "fd4034d9514d6e875538422c8b0dbeb2.dir",
+            },
+        ),
+    ],
+)
+def test_stage_dir_optimization(
+    tmp_dir, dvc, mocker, dry_run, expected_staging_contents
+):
+    from dvc.data import stage
+    from dvc.data.tree import Tree
+
+    tmp_dir.dvc_gen(
+        {
+            "data": {
+                "foo": "bar",
+                "subdir": {"subfoo": "subbar"},
+            }
+        }
+    )
+    odb = dvc.odb.local
+
+    objs = set(odb.all())
+    clean_staging()
+
+    tmp_dir.gen({"data": {"baz": "quz"}})
+
+    stage_spy = mocker.spy(stage, "_stage_tree")
+    _, _, tree = stage.stage(odb, "data", odb.fs, odb.fs.PARAM_CHECKSUM)
+
+    assert stage_spy.called
+    assert set(odb.all()) - objs == {tree.hash_info.as_raw().value}
+    stage_spy.reset_mock()
+    clean_staging()
+
+    load_spy = mocker.spy(Tree, "load")
+    build_tree_spy = mocker.spy(stage, "_build_tree")
+
+    staging, _, tree = stage.stage(
+        odb, "data", odb.fs, odb.fs.PARAM_CHECKSUM, dry_run=dry_run
+    )
+    assert not stage_spy.called
+    assert not build_tree_spy.called
+
+    load_args, _ = load_spy.call_args
+    assert load_args[1].value == tree.hash_info.as_raw().value
+
+    assert set(staging.all()) == expected_staging_contents
--- a/tests/unit/test_hashinfo.py
+++ b/tests/unit/test_hashinfo.py
@@ -1,0 +1,17 @@
+from dvc.objects.hash_info import HashInfo
+
+
+def test_as_raw():
+    hash_info = HashInfo(
+        "md5", "a1d0c6e83f027327d8461063f4ac58a6.dir", "objname"
+    )
+
+    raw = hash_info.as_raw()
+
+    assert hash_info.name == "md5"
+    assert hash_info.value == "a1d0c6e83f027327d8461063f4ac58a6.dir"
+    assert hash_info.obj_name == "objname"
+
+    assert raw.name == "md5"
+    assert raw.value == "a1d0c6e83f027327d8461063f4ac58a6"
+    assert raw.obj_name == "objname"


dtrifiro:
Some benchmarks:
- [Large dataset](https://github.com/dtrifiro/dvc-bench/runs/6232338342?check_suite_focus=true#step:8:282)
- [Small dataset](https://github.com/dtrifiro/dvc-bench/runs/6231782423?check_suite_focus=true#step:8:279)

Download results archive here:
- [large](https://github.com/dtrifiro/dvc-bench/actions/runs/2246494794)
- [small](https://github.com/dtrifiro/dvc-bench/actions/runs/2246495433)
click "results" to download zip archive, expand it
```bash
unzip -d results results.zip
py.test-benchmark compare results/test_status/*{fix,2.10.0}*json
```
Results for `test_status` for this branch vs 2.10:
```

--------------------------------------------------------------------------- benchmark 'test_status-status': 2 tests ----------------------------------------------------------------------------
Name (time in s)                         Min               Max              Mean            StdDev            Median               IQR            Outliers     OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_status-status (0001_fix-739)     2.2167 (1.0)      2.2167 (1.0)      2.2167 (1.0)      0.0000 (1.0)      2.2167 (1.0)      0.0000 (1.0)           0;0  0.4511 (1.0)           1           1
test_status-status (0003_2.10.0)      3.5338 (1.59)     3.5338 (1.59)     3.5338 (1.59)     0.0000 (1.0)      3.5338 (1.59)     0.0000 (1.0)           0;0  0.2830 (0.63)          1           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

--------------------------------------------------------------------------- benchmark 'test_status-status-changed': 2 tests ----------------------------------------------------------------------------
Name (time in s)                                 Min               Max              Mean            StdDev            Median               IQR            Outliers     OPS            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_status-status-changed (0001_fix-739)     1.3470 (1.0)      1.3470 (1.0)      1.3470 (1.0)      0.0000 (1.0)      1.3470 (1.0)      0.0000 (1.0)           0;0  0.7424 (1.0)           1           1
test_status-status-changed (0003_2.10.0)      2.5265 (1.88)     2.5265 (1.88)     2.5265 (1.88)     0.0000 (1.0)      2.5265 (1.88)     0.0000 (1.0)           0;0  0.3958 (0.53)          1           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------------ benchmark 'test_status-status-changed-noop': 2 tests -----------------------------------------------------------------------------------
Name (time in ms)                                         Min                   Max                  Mean            StdDev                Median               IQR            Outliers     OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_status-status-changed-noop (0001_fix-739)       597.5263 (1.0)        597.5263 (1.0)        597.5263 (1.0)      0.0000 (1.0)        597.5263 (1.0)      0.0000 (1.0)           0;0  1.6736 (1.0)           1           1
test_status-status-changed-noop (0003_2.10.0)      2,545.9197 (4.26)     2,545.9197 (4.26)     2,545.9197 (4.26)     0.0000 (1.0)      2,545.9197 (4.26)     0.0000 (1.0)           0;0  0.3928 (0.23)          1           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
```

dvc/data/stage.py
```diff
@@ -336,6 +367,9 @@ def stage(
         logger.debug("staged tree '%s'", obj)
         if name != "md5":
             obj = _stage_external_tree_info(odb, obj, name)
+
+        untyped = HashInfo(obj.name, obj.hash_info.value.replace(".dir", ""))
+        odb.add(obj.fs_path, obj.fs, untyped)
```
efiop, line 372:
```suggestion
        odb.add(obj.fs_path, obj.fs, obh.hash_info.as_raw())
```

efiop: Side note: this `as_raw` is obviously not really about HashInfo, but more about ObjectID(Oid) that should be based on HashInfo. `HashInfo.isdir` is also clearly an oid thing 🙂 We'll introduce oids properly a bit later, clearly out of scope for this PR.

dvc/data/stage.py
```diff
@@ -247,6 +247,29 @@ def _get_staging(odb: "ObjectDB") -> "ObjectDB":
...
+    obj.hash_info = hash_info
+    meta.nfiles = len(obj)
+
+    staging.add(obj.fs_path, obj.fs, obj.hash_info)
+    for key, (_, entry_hash_info) in obj.iteritems():
+        staging.add(
+            odb.fs.path.join(fs_path, *key),
+            odb.fs,
+            entry_hash_info,
+            verify=verify,
```
efiop, line 268: Doesn't seem like verify is used, it is always False

efiop:
@dtrifiro Thanks for the benches! 🙏 

Btw, could you take a look at what is still taking that long in relation to this PR? I suppose it is mostly the overhead of those ref obj recreations? Just wondering if there are any low hangers that we could pick until a proper solution with persistent refobjects arrives.

dvc/data/stage.py
dtrifiro: Indeed, originally added this to to some benchmarks, `_load_untyped_dir_hash` is 6.5x faster when `verify=False`, will remove.

dtrifiro:
Here's a trace for a folder with a single tracked file (after adding another file to make it modified)
image
Here's a detail on on the for loop:
image
Here's a detail on a single `add` call in the loop: 
image

dvc/data/stage.py
```diff
@@ -247,6 +247,24 @@ def _get_staging(odb: "ObjectDB") -> "ObjectDB":
...
+
+    obj = Tree.load(odb, hash_info.as_raw())
+    obj.hash_info = hash_info
+    meta.nfiles = len(obj)
+
+    staging.add(obj.fs_path, obj.fs, obj.hash_info)
+    for key, (_, entry_hash_info) in obj.iteritems():
+        staging.add(
+            odb.fs.path.join(fs_path, *key),
+            odb.fs,
```
efiop, line 261: we are not adding files from odb.fs, we are adding them from fs. Even though they happen to match here, it is not going to be always like that.
```suggestion
            fs.path.join(fs_path, *key),
            fs,
```

efiop:
@dtrifiro Thanks for the traces 🙏 Any particular points you were able to point out?

dvc/hash_info.py
```diff
@@ -41,3 +41,9 @@ def isdir(self):
         if not self:
             return False
         return self.value.endswith(HASH_DIR_SUFFIX)
+
+    def as_raw(self) -> "HashInfo":
+        assert self.value
+        return HashInfo(
+            self.name, self.value.replace(HASH_DIR_SUFFIX, ""), self.obj_name
```
efiop, line 48: using replace it not really needed and might actually hide some bugs

```suggestion
            self.name, self.value.rsplit(HASH_DIR_SUFFIX)[0], self.obj_name
```

dvc/data/stage.py
```diff
@@ -247,6 +247,24 @@ def _get_staging(odb: "ObjectDB") -> "ObjectDB":
...
+
+    staging.add(obj.fs_path, obj.fs, obj.hash_info)
+    for key, (_, entry_hash_info) in obj.iteritems():
+        staging.add(
+            fs.path.join(fs_path, *key),
+            fs,
+            entry_hash_info,
+            verify=False,
+        )
+    return staging, meta, Tree.load(staging, hash_info)
```
efiop, line 265: Seems like this is doing much more than "load untyped dir hash". Also "load hash" is not really describing what is going on here. Maybe there is a way to clean the code up?

dvc/data/stage.py
```diff
@@ -336,6 +364,8 @@ def stage(
         logger.debug("staged tree '%s'", obj)
         if name != "md5":
             obj = _stage_external_tree_info(odb, obj, name)
+
+        odb.add(obj.fs_path, obj.fs, obj.hash_info.as_raw())
```
efiop, line 368: Let's add some note here that this is for that tricky optimization or something.

dvc/data/stage.py
```diff
@@ -247,6 +247,24 @@ def _get_staging(odb: "ObjectDB") -> "ObjectDB":
     return ReferenceObjectDB(fs, fs_path, state=state)
 
 
+def _load_untyped_dir_hash(odb, fs, fs_path, hash_info, meta, staging):
+    from .tree import Tree
+
+    obj = Tree.load(odb, hash_info.as_raw())
+    obj.hash_info = hash_info
```
efiop, line 254: This seems useless
```suggestion
```

dvc/data/stage.py
dtrifiro: How about `_stage_tree_from_untyped_hash`?

dvc/data/stage.py
```diff
@@ -247,6 +247,24 @@ def _get_staging(odb: "ObjectDB") -> "ObjectDB":
     return ReferenceObjectDB(fs, fs_path, state=state)
 
 
+def _load_untyped_dir_hash(odb, fs, fs_path, hash_info, meta, staging):
+    from .tree import Tree
+
+    obj = Tree.load(odb, hash_info.as_raw())
```
efiop, line 253: This might raise `ObjectFormatError`, which is not how `_load_from_stage` behaved and so the code outside will not be ready for this

dvc/data/stage.py
```diff
@@ -247,6 +247,24 @@ def _get_staging(odb: "ObjectDB") -> "ObjectDB":
...
+    obj.hash_info = hash_info
+    meta.nfiles = len(obj)
```
efiop, line 255: Seems like this is duplicating the older logic (which is minor), but also note that you are not validating the object anymore (see check() call above). So maybe this is a hint that you need to look into somehow unifying what you have here with what has already been in `_load_from_state`.

dvc/data/stage.py
efiop: Also note that you are loading the tree twice (which means that you are reading and parsing json twice). Do you see that taking significant time in the benchmarks? If so, is there a nice way to avoid that?

dvc/data/stage.py
```diff
@@ -233,7 +236,7 @@ def _make_staging_url(
     return url
 
 
-def _get_staging(odb: "ObjectDB") -> "ObjectDB":
+def _get_staging(odb: "ObjectDB") -> "ReferenceObjectDB":
```
efiop, line 239: Was this causing mypy complaints? Seems like this change is unrelated.

dvc/data/stage.py
```diff
@@ -247,27 +250,88 @@ def _get_staging(odb: "ObjectDB") -> "ObjectDB":
...
     from dvc.objects.errors import ObjectFormatError
 
-    from . import check, load
+    from . import check as ocheck
+    from . import load
     from .tree import Tree
 
+    try:
+        if raw_dir_hash:
+            obj = Tree.load(odb, hash_info)
```
efiop, line 287: We still need to check the object to make sure it is not corrupted

dvc/data/stage.py
```diff
@@ -247,27 +250,88 @@ def _get_staging(odb: "ObjectDB") -> "ObjectDB":
...
+
+    meta, tree = _try_load(
+        odb, hash_info.as_raw(), meta, name, raw_dir_hash=True
+    )
+    if not tree:
+        raise FileNotFoundError
+
+    from .tree import Tree
+
+    assert isinstance(tree, Tree)
```
efiop, line 332: imports in the middle of the body are a bad look unless really necessary. In this case it seems like you might as well `assert` in `_stage_raw_dir_hash`

tests/func/test_gc.py
```diff
@@ -25,6 +25,11 @@ def setUp(self):
             self.dvc.odb.local.hash_to_path(md5)
             for md5 in self.dvc.odb.local.all()
         ]
+        for hash_info in [
+            hash_info for hash_info in self.good_cache if ".dir" in hash_info
+        ]:
+            untyped_hash_info = hash_info.replace(".dir", "")
+            self.good_cache.remove(untyped_hash_info)
```
efiop, line 32: This double for looks very odd
```suggestion
        for hash_info in self.good_cache:
            if hash_info.isdir:
                self.good_cache.remove(hash_info.as_raw())
```

efiop: Also, why are we even removing it ourselves? Isn't there a nice way to adjust the test

tests/func/test_remote.py
```diff
@@ -154,6 +154,8 @@ def test_dir_hash_should_be_key_order_agnostic(tmp_dir, dvc):
         _, _, obj = stage(dvc.odb.local, path, dvc.odb.local.fs, "md5")
         hash1 = obj.hash_info
 
+    remove(dvc.odb.local.cache_dir)
+
```
efiop, line 158: Could you explain why, please?

tests/func/test_repro.py
```diff
@@ -715,6 +715,8 @@ def test(self):
         stages = self.dvc.reproduce(target)
         self.assertEqual(len(stages), 0)
 
+        self.dvc.odb.local.unprotect(self.DATA_SUB)
+
```
efiop, line 719: Btw, were you able to figure out why this got triggered?

tests/func/test_status.py
```diff
@@ -183,3 +184,42 @@ def test_params_without_targets(tmp_dir, dvc):
             }
         ]
     }
+
+
+def test_dir_optimization(tmp_dir, dvc, mocker):
```
efiop, line 189: Would it make sense to test `data.stage` instead?

tests/func/test_commit.py
```diff
@@ -146,26 +146,33 @@ def test_commit_granular_dir(tmp_dir, dvc):
...
+        cache / "1a" / "ca2c799df82929bbdd976557975546",
         cache / "1a" / "ca2c799df82929bbdd976557975546.dir",
         cache / "ac" / "bd18db4cc2f85cedef654fccc4a4d8",
         cache / "4c" / "e8d2a2cf314a52fa7f315ca37ca445",
         cache / "68" / "dde2c3c4e7953c2290f176bbdc9a54",
     }
     clean_staging()
 
-    dvc.commit(os.path.join("data"))
+    dvc.commit("data")
```
efiop, line 172: Unrelated change?

dvc/data/stage.py
```diff
@@ -247,27 +250,88 @@ def _get_staging(odb: "ObjectDB") -> "ObjectDB":
...
-                    pass
-    raise FileNotFoundError
+    if not hash_info:
+        raise FileNotFoundError
+
+    for odb_ in (odb, staging):
+        if not odb_.exists(hash_info):
+            continue
+
+        meta, obj = _try_load(odb_, hash_info, meta, name)
```
efiop, line 317: You are passing the name and the meta just to modify and assert them there, while you could just
```suggestion
        meta, obj = _try_load(odb_, hash_info)
        if isinsatnce(obj, Tree):
            meta.nfiles = len(obj)
        assert obj.hash_info.name == name
```
and avoid redundant passthrough

dvc/data/stage.py
```diff
@@ -247,27 +250,88 @@ def _get_staging(odb: "ObjectDB") -> "ObjectDB":
...
+    staging: "ReferenceObjectDB",
+    tree: "Tree",
+    fs_path: "AnyFSPath",
+    fs: "FileSystem",
+    hash_info: "HashInfo",
+) -> "Tree":
+    assert tree.fs
+    staging.add(tree.fs_path, tree.fs, hash_info)
+    for key, _, oid in tree:
+        staging.add(fs.path.join(fs_path, *key), tree.fs, oid, verify=False)
```
efiop, line 263: Btw, strictly speaking, we need to add the files first, so that at no point our odb is corrupted and so we avoid the risk of leaving it corrupted in a case of an error (not too relevant with memory odb, but still)

tests/func/test_remote.py
dtrifiro: When we call `stage()`, we also create a raw dir hash for the tree. When calling `stage` the second time we load the tree from the raw dir hash. Since `HashInfo` values are set to dummy values, `hash_to_path` (which **expects** hashes to be longer than 3 characters) returns a dir (`memory://dvc-staging/3701baaa21c459498c4c21986d6a85b80d016815f63942d17fb650599e4d1558/1/`). This results in a failure when adding the tree to staging.

The solution was to just avoid loading the raw dir hash by removing it.
To make things clearer I could just remove the raw dir hash and add a note:
```diff
-    remove(dvc.odb.local.cache_dir)
+    # remove the raw dir hash so that the next `stage` call builds the tree instead of loading it from the raw dir hash
+    odb.fs.remove(odb.hash_to_path("55be181145ff2aa155e719083a5ffb64"))

The test would also pass by replacing the hashes when creating the dummy tree:
```diff
-        [{"relpath": "1", "md5": "1"}, {"relpath": "2", "md5": "2"}]
+        [{"relpath": "1", "md5": "123"}, {"relpath": "2", "md5": "234"}]
```
so that `hash_to_path()` would work correctly, but this would essentially make this test useless as the tree would just be loaded from the raw dir hash.

We could also add an assert in hash to path in order to avoid similar issues in the future:

```diff
    def hash_to_path(self, hash_):
+       assert len(hash_) > 2
        return self.fs.path.join(self.fs_path, hash_[0:2], hash_[2:])
```

tests/func/test_gc.py
dtrifiro: What you propose does not work because `odb.all()` iterates over hash info values and `self.good_cache` contains paths

How about this? It hardcodes the raw_dir_hash value, but is pretty straight forward imho
```diff
   self.dvc.add(self.FOO)
         self.dvc.add(self.DATA_DIR)
+        raw_dir_hash = "bff5a787d16460f32e9f2e62b183b1cc"
         self.good_cache = [
             self.dvc.odb.local.hash_to_path(md5)
             for md5 in self.dvc.odb.local.all()
+            if md5 != raw_dir_hash
         ]
-        for hash_info in [
-            hash_info for hash_info in self.good_cache if ".dir" in hash_info
-        ]:
-            untyped_hash_info = hash_info.replace(".dir", "")
-            self.good_cache.remove(untyped_hash_info)
 
-        self.bad_cache = []
+        self.bad_cache = [self.dvc.odb.local.hash_to_path(raw_dir_hash)]
         for i in ["123", "234", "345"]:
             path = os.path.join(self.dvc.odb.local.cache_dir, i[0:2], i[2:])
             self.create(path, i)
```

tests/func/test_status.py
dtrifiro: Moved the test to `test_stage.py` and adapted it

tests/func/test_repro.py
dtrifiro: Indeed: I wasn't honoring `dry_run` for `stage()`. Fixed now and got rid of `unprotect` in this test.

dvc/data/stage.py
dtrifiro: Yep: it is required because since I added type hints to `_load_from_state()`

dvc/data/stage.py
```diff
@@ -247,27 +250,117 @@ def _get_staging(odb: "ObjectDB") -> "ObjectDB":
...
+    tree.fs = raw.fs
+    tree.fs_path = raw.fs_path
+    tree.hash_info = hash_info
+    return tree
+
+
+def _try_load(
+    odb: "ObjectDB",
+    hash_info: "HashInfo",
+    raw_dir_hash: bool = False,
```
pmrowla, line 307: Does `_try_load()` need this flag (and/or is this method needed at all)?

I get that this is trying to reduce duplicated load/check code between the old `_load_from_state` and the new raw tree handling, but I think this might end up making it more confusing to read.

It seems to me like this is adding a helper for loading any generic object, but in the event that the caller knows they want to load a specific hash as a specific type of object, I think the caller should just be doing that directly (rather than passing extra flags into a helper method).

So in this case, instead of using a helper method `_load_raw_dir_hash` would just start with

```py
    try:
        tree = Tree.load(odb, hash_info.as_raw())
        tree.check(odb)
    except ObjectFormatError:
        raise FileNotFoundError
```

Alternatively, if you want to keep the helper methods to reduce the exception handling boilerplate in the event we need to re-use this functionality again later, I'd prefer to see separate generic `_try_load()` and `_try_load_tree()` helper methods.

pmrowla: Basically, when we have code that looks like:

```py
tree = _try_load(odb, hash_info.as_raw(), raw_dir_hash=True)
assert isinstance(tree, Tree)
```

It leads me to think that the helper method is not actually helping, given that we are already doing 2 different things to use the "raw" hash, and we are also forced to verify that the helper is in fact returning what we expected it to return.

(Not sure if the assert statement is to clean up mypy warnings, but in this case I would agree with mypy that it is confusing to figure out what `_try_load()` is going to do if I was looking at that line in the code)

dvc/data/stage.py
```diff
@@ -247,27 +250,117 @@ def _get_staging(odb: "ObjectDB") -> "ObjectDB":
...
+
+    assert isinstance(tree, Tree)
+    assert tree.fs
+    meta.nfiles = len(tree)
+    assert tree.hash_info.name == name
+
+    if not dry_run:
+        with ui.status(f"Loading {fs_path}"):
+            for key, _, oid in tree:
+                staging.add(
```
dtrifiro, line 280: Just a quick note about this: since we [recently](https://github.com/iterative/dvc/commit/5067e1f5bc3335431c50b670282b1c85f11f3d3e) modified `ObjectDB.add` with callbacks
https://github.com/iterative/dvc/blob/b14a972cd057a5fcf6dfbee5da475e30c1f36bec/dvc/objects/db.py#L133-L146
calling `add()` with no callback arguments results in a `Tqdm` instance being created for each call, which results in a massive slow down when adding a large amount of files (at least 4x in my limited testing).

A quick solution is to modify this call with:
```python
with NoOpCallback() as cb:
    for key, _, oid in tree:
        staging.add(
            ...,
            callback=cb
        )
```
Note that this won't affect current behaviour, because the callback is never called for `ReferenceObjectDB._add_file` (it is only called for a directory, since that calls `ObjectDB._add_file`). 
Instantiating `Tqdm` prints a progressbar for the given file but this is never updated (and is too quick to be of any use anyway).

If we wish to show a progressbar for the state of staging, we'd also have to add something like:
```python
with NoOpCallback() as cb:
    with Tqdm(
        unit="files",
        desc=f"Loading {odb.fs.path.name(fs_path)}",
        total=meta.nfiles + 1,
    ) as pbar:
        for key, _, oid in tree:
            staging.add(
                ...,
                callback=cb
            )
             pbar.update()
```

dtrifiro: Note that this also makes `transfer` very slow, since it calls `dest.add` with no callbacks.

dtrifiro: Just for context, for a directory with ~10k files:

- no `callback` kwarg to `staging.add`: `dvc status  6.52s user 0.41s system 98% cpu 7.017 total`
- with `callback=noopcallback`: `dvc status  2.26s user 0.14s system 99% cpu 2.410 total`

cc @skshetry

skshetry: @dtrifiro, I think it makes sense to use `fs._callback.DEFAULT_CALLBACK` in the case of individual objects, and use Tqdm one for the tree. (My understanding is that they are just saving to configs to `refodb`/`memfs` right?)

tests/func/test_gc.py
dtrifiro: Ended up going with the following:

```diff
         self.dvc.add(self.FOO)
-        self.dvc.add(self.DATA_DIR)
+        stages = self.dvc.add(self.DATA_DIR)
+        raw_dir_hash = stages[0].outs[0].hash_info.as_raw().value
+
         self.good_cache = [
             self.dvc.odb.local.hash_to_path(md5)
             for md5 in self.dvc.odb.local.all()
+            if md5 != raw_dir_hash
         ]
 
-        self.bad_cache = []
+        self.bad_cache = [self.dvc.odb.local.hash_to_path(raw_dir_hash)]
         for i in ["123", "234", "345"]:
             path = os.path.join(self.dvc.odb.local.cache_dir, i[0:2], i[2:])
             self.create(path, i)
```

dvc/data/stage.py
```diff
@@ -247,27 +250,109 @@ def _get_staging(odb: "ObjectDB") -> "ObjectDB":
...
+
+    assert tree.fs
+    for key, _, oid in tree:
+        staging.add(
+            fs.path.join(fs_path, *key),
+            fs,
+            oid,
+            hardlink=False,
+            verify=False,
+            callback=DEFAULT_CALLBACK,
```
skshetry, line 347: I think it's better to leave this out since it is very cheap now.

dtrifiro:
Updated benchmarks can be found here:
https://github.com/dtrifiro/dvc-bench/runs/6368942357?check_suite_focus=true#step:8:13

A summary for `test_status`:
```console
----------------------------------------------------------------------------------- benchmark 'test_status-status': 3 tests ------------------------------------------------------------------------------------
Name (time in s)                                      Min                Max               Mean            StdDev             Median               IQR            Outliers     OPS            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_status-status (test_status/0001_fix-739)     25.2779 (1.0)      25.2779 (1.0)      25.2779 (1.0)      0.0000 (1.0)      25.2779 (1.0)      0.0000 (1.0)           0;0  0.0396 (1.0)           1           1
test_status-status (test_status/0002_main)        36.9391 (1.46)     36.9391 (1.46)     36.9391 (1.46)     0.0000 (1.0)      36.9391 (1.46)     0.0000 (1.0)           0;0  0.0271 (0.68)          1           1
test_status-status (test_status/0003_2.10.0)      36.9755 (1.46)     36.9755 (1.46)     36.9755 (1.46)     0.0000 (1.0)      36.9755 (1.46)     0.0000 (1.0)           0;0  0.0270 (0.68)          1           1
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

----------------------------------------------------------------------------------- benchmark 'test_status-status-changed': 3 tests ------------------------------------------------------------------------------------
Name (time in s)                                              Min                Max               Mean            StdDev             Median               IQR            Outliers     OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_status-status-changed (test_status/0001_fix-739)     14.9064 (1.0)      14.9064 (1.0)      14.9064 (1.0)      0.0000 (1.0)      14.9064 (1.0)      0.0000 (1.0)           0;0  0.0671 (1.0)           1           1
test_status-status-changed (test_status/0002_main)        26.9573 (1.81)     26.9573 (1.81)     26.9573 (1.81)     0.0000 (1.0)      26.9573 (1.81)     0.0000 (1.0)           0;0  0.0371 (0.55)          1           1
test_status-status-changed (test_status/0003_2.10.0)      27.1236 (1.82)     27.1236 (1.82)     27.1236 (1.82)     0.0000 (1.0)      27.1236 (1.82)     0.0000 (1.0)           0;0  0.0369 (0.55)          1           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------------ benchmark 'test_status-status-changed-noop': 3 tests -----------------------------------------------------------------------------------
Name (time in s)                                                   Min                Max               Mean            StdDev             Median               IQR            Outliers     OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_status-status-changed-noop (test_status/0001_fix-739)      3.9865 (1.0)       3.9865 (1.0)       3.9865 (1.0)      0.0000 (1.0)       3.9865 (1.0)      0.0000 (1.0)           0;0  0.2508 (1.0)           1           1
test_status-status-changed-noop (test_status/0002_main)        26.6765 (6.69)     26.6765 (6.69)     26.6765 (6.69)     0.0000 (1.0)      26.6765 (6.69)     0.0000 (1.0)           0;0  0.0375 (0.15)          1           1
test_status-status-changed-noop (test_status/0003_2.10.0)      27.0214 (6.78)     27.0214 (6.78)     27.0214 (6.78)     0.0000 (1.0)      27.0214 (6.78)     0.0000 (1.0)           0;0  0.0370 (0.15)          1           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

File: dvc/data/stage.py [hunk 23]
```diff
@@ -247,27 +250,106 @@ def _get_staging(odb: "ObjectDB") -> "ObjectDB":
...
+def _load_raw_dir_obj(odb: "ObjectDB", hash_info: "HashInfo") -> "Tree":
+    from dvc.objects.errors import ObjectFormatError
+
+    from .tree import Tree
+
+    try:
+        tree = Tree.load(odb, hash_info.as_raw())
+        tree.check(odb)
+    except ObjectFormatError:
+        raise FileNotFoundError
```
efiop, line 279:
```suggestion
        raise FileNotFoundError from exc
```

skshetry: While we are here, 
```suggestion
        raise FileNotFoundError(errno.ENOENT, "No such object", odb.hash_to_path(hash_info.value))
```

Or, just use `hash_info.value` for simplicity.

skshetry:
@dtrifiro, are those improvements on `test_status-status` and `test_status-status-changed` expected?

dtrifiro:
> @dtrifiro, are those improvements on `test_status-status` and `test_status-status-changed` expected?

I forgot to update the `main` branch for fork after rebasing, meaning the comparison was against a slightly outdated revision. Here's a comparison for this PR against its base (main@[`3faf4132`](https://github.com/iterative/dvc/commit/3faf4132)):

https://github.com/dtrifiro/dvc-bench/runs/6404025103?check_suite_focus=true#step:8:258


```
-------------------------------- benchmark 'test_status-status': 5 tests --------------------------------
Name (time in s)                                   Median            StdDev            Rounds  Iterations
---------------------------------------------------------------------------------------------------------
test_status-status (test_status/0001_fix-739)     22.8981 (1.10)     0.0000 (1.0)           1           1
test_status-status (test_status/0002_main)        21.4079 (1.03)     0.0000 (1.0)           1           1
test_status-status (test_status/0003_2.10.2)      20.7695 (1.0)      0.0000 (1.0)           1           1
test_status-status (test_status/0004_2.9.5)       32.8662 (1.58)     0.0000 (1.0)           1           1
test_status-status (test_status/0005_2.6.3)       36.3975 (1.75)     0.0000 (1.0)           1           1
---------------------------------------------------------------------------------------------------------

-------------------------------- benchmark 'test_status-status-changed': 5 tests --------------------------------
Name (time in s)                                           Median            StdDev            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------
test_status-status-changed (test_status/0001_fix-739)     12.6867 (1.0)      0.0000 (1.0)           1           1
test_status-status-changed (test_status/0002_main)        12.7397 (1.00)     0.0000 (1.0)           1           1
test_status-status-changed (test_status/0003_2.10.2)      12.8673 (1.01)     0.0000 (1.0)           1           1
test_status-status-changed (test_status/0004_2.9.5)       25.0225 (1.97)     0.0000 (1.0)           1           1
test_status-status-changed (test_status/0005_2.6.3)       25.2563 (1.99)     0.0000 (1.0)           1           1
-----------------------------------------------------------------------------------------------------------------

-------------------------------- benchmark 'test_status-status-changed-noop': 5 tests --------------------------------
Name (time in s)                                                Median            StdDev            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------
test_status-status-changed-noop (test_status/0001_fix-739)      3.7390 (1.0)      0.0000 (1.0)           1           1
test_status-status-changed-noop (test_status/0002_main)        12.7007 (3.40)     0.0000 (1.0)           1           1
test_status-status-changed-noop (test_status/0003_2.10.2)      11.8403 (3.17)     0.0000 (1.0)           1           1
test_status-status-changed-noop (test_status/0004_2.9.5)       24.2830 (6.49)     0.0000 (1.0)           1           1
test_status-status-changed-noop (test_status/0005_2.6.3)       25.2839 (6.76)     0.0000 (1.0)           1           1
----------------------------------------------------------------------------------------------------------------------
```

efiop (approved):
Ok, it looks like we could clean this up a bit more, but let's run with it for now and get back to it when we safe ref objects.

dvc/data/stage.py
```diff
@@ -247,27 +251,90 @@ def _get_staging(odb: "ObjectDB") -> "ObjectDB":
...
+    if not hash_info:
+        raise FileNotFoundError
+
+    for odb_ in (odb, staging):
+        if not odb_.exists(hash_info):
+            continue
+
+        try:
+            obj = load(odb, hash_info)
+            check(odb, obj, check_hash=False)
```
efiop, line 297: Btw, this PR introduced a typo: it should be `_odb` here 🙁

PR #7603, opened 2022-04-20
exp init: Replace live section with metrics and plots.

### LINKED ISSUE #7534: exp init: Revisit `type="dl"`
Once https://github.com/iterative/dvclive/issues/225 is solved, the `live` section will be deprecated in favor of using `metrics` and `plots`.

How should `exp init` behave when `type="dl"`?

daavoo:
Closes #7534

### DIFF
--- a/dvc/repo/experiments/init.py
+++ b/dvc/repo/experiments/init.py
@@ -239,18 +239,20 @@ def init(
         except MissingParamsFile:
             pass
 
-    checkpoint_out = bool(context.get("live"))
     models = context.get("models")
+    live_path = context.pop("live", None)
+    live_metrics = f"{live_path}.json" if live_path else None
+    live_plots = os.path.join(live_path, "scalars") if live_path else None
+
     stage = repo.stage.create(
         name=name,
         cmd=context["cmd"],
         deps=compact([context.get("code"), context.get("data")]),
         params=[{params: None}] if params else None,
-        metrics_no_cache=compact([context.get("metrics")]),
-        plots_no_cache=compact([context.get("plots")]),
-        live=context.get("live"),
+        metrics_no_cache=compact([context.get("metrics"), live_metrics]),
+        plots_no_cache=compact([context.get("plots"), live_plots]),
         force=force,
-        **{"checkpoints" if checkpoint_out else "outs": compact([models])},
+        **{"checkpoints" if with_live else "outs": compact([models])},
     )
 
     with _disable_logging(), repo.scm_context(autostage=True, quiet=True):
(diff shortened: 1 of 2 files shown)


tests/func/experiments/test_init.py
```diff
@@ -136,13 +136,19 @@ def test_init_interactive_when_no_path_prompts_need_to_be_asked(
...
-                # so it creates checkpoint-based output.
-                "outs": [{"models": {"checkpoint": True}}],
+                "metrics": [
+                    {"dvclive.json": {"cache": False}},
+                    {"metrics.json": {"cache": False}},
+                ],
+                # TODO: discuss:
+                # IMO, specifying `live` should not create checkpoint output
+                # unless type="dl" is passed
+                "outs": ["models"],
```
daavoo, line 146: @dberenbaum / @skshetry I might have missed the discussion about this behavior.

Enabling checkpoints makes sense with `type="dl"`, but passing `live` should not mean that `checkpoints` are also enabled. 

I have changed the previous behavior in order to decouple `live` from `checkpoints`. Checkpoints are only enabled with `type="dl"`.

dvc/repo/experiments/init.py
```diff
@@ -239,18 +253,16 @@ def init(
         except MissingParamsFile:
             pass
```
skshetry, line 255: Nitpicky: let's just specify those live_paths here:
```
    models = context.get("models")
    live = context.get("live")
    live_metrics = os.path.join(live, "metrics") if live else None
    live_plots = os.path.join(live, "scalars") if live else None
```

And, then you can specify them in `repo.stage.create` as following:
```python
 repo.stage.create(
        ...,
        metrics_no_cache=compact([context["metrics"], live_metrics]),
        plots_no_cache=compact([context["plots"], live_plots]),
)
```
 
 Makes it simpler, and avoids from saving it to the list.

tests/func/experiments/test_init.py
skshetry: It was introduced in https://github.com/iterative/dvc/pull/6716. I don't remember exactly why it was done this way, but probably it was done to unify with interactive mode (even with `--type=dl`, users can omit `live`).

It makes sense to do that anyway on `--type="dl"`.

dvc/repo/experiments/init.py
```diff
@@ -222,6 +222,7 @@ def init(
             defaults.pop("live", None)  # suppress live otherwise
 
     context: Dict[str, str] = {**defaults, **overrides}
+
```
skshetry, line 225:
```suggestion
```

skshetry (approved):

PR #7606, opened 2022-04-20
render.match: Skip errored revisions in `get_files`

daavoo:
Introduced in #7409

### DIFF
--- a/dvc/render/match.py
+++ b/dvc/render/match.py
@@ -10,8 +10,12 @@ if TYPE_CHECKING:
 
 
 def get_files(plots_data: Dict) -> List:
-    values = dpath.util.values(plots_data, ["*", "*"])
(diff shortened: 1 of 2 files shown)


dvc/render/match.py
```diff
@@ -10,8 +10,12 @@
 
 
 def get_files(plots_data: Dict) -> List:
-    values = dpath.util.values(plots_data, ["*", "*"])
-    return sorted({key for submap in values for key in submap.keys()})
```
daavoo, line 14: When a error happens during collection, `values` would be:

```
{"error": ResolveError}
```
And the removed code would fail with:

```
AttributeError: 'ResolveError' object has no attribute 'keys'
```

pared (approved):

PR #7613, opened 2022-04-21
api: Add params_show.

### LINKED ISSUE #6507: Expose `params`/ `ParamsDependency` in the Python API.
`params` are easy to include in a pipeline via [templating](https://dvc.org/doc/user-guide/project-structure/pipelines-files#templating) but using them inside a python script might not be that trivial and require several changes in the user's code (i.e. when a stage uses several params from different params files).

I think that it would be nice to expose some function in the public API that would take care of loading the `params` associated to a stage and returning them as python objects, directly usable in the python script. 

It could be something like `dvc.api.get_params()` and would take care of loading only the params defined in the `params` section of the stage.

daavoo:
Closes #6507

Uses `repo.params.show` with custom error_handler and postprocess the outputs for a more user-friendly structure.

Extend `repo.params.show` to accept `stages` argument to cover the "params of current stage" use case.

`dvc.org` P.R: https://github.com/iterative/dvc.org/pull/3459

---

## Examples

- No args.

Will use the current project as `repo` and retrieve all parameters,
for all stages, for the current revision.

```python
>>> import json
>>> import dvc.api
>>> params = dvc.api.params_show()
>>> print(json.dumps(params, indent=4))
{
    "prepare": {
        "split": 0.2,
        "seed": 20170428
    },
    "featurize": {
        "max_features": 10000,
        "ngrams": 2
    },
    "train": {
        "seed": 20170428,
        "n_est": 50,
        "min_split": 0.01
    }
}
```

- Filter with `stages`.

```python
>>> import json
>>> import dvc.api
>>> params = dvc.api.params_show(stages="prepare")
>>> print(json.dumps(params, indent=4))
{
    "prepare": {
        "split": 0.2,
        "seed": 20170428
    }
}
```

- Git URL as `repo`.

```python
>>> import json
>>> import dvc.api
>>> params = dvc.api.params_show(
...     repo="https://github.com/iterative/demo-fashion-mnist")
{
    "train": {
        "batch_size": 128,
        "hidden_units": 64,
        "dropout": 0.4,
        "num_epochs": 10,
        "lr": 0.001,
        "conv_activation": "relu"
    }
}
```

- Using `rev`.

```python
>>> import json
>>> import dvc.api
>>> params = dvc.api.get_params(
...     repo="https://github.com/iterative/demo-fashion-mnist",
...     rev="low-lr-experiment")
>>> print(json.dumps(params, indent=4))
{
      "train": {
          "batch_size": 128,
          "hidden_units": 64,
          "dropout": 0.4,
          "num_epochs": 10,
          "lr": 0.001,
          "conv_activation": "relu"
      }
}
```


dvc/api.py
```diff
@@ -91,6 +93,24 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
...
+    Args:
+        stage (str): Name of the dvc.yaml stage to get params from.
+        flatten (bool, optional): Whether to return flattened dict
+        or not. Defaults False.
+        If True, {"foo.bar": 1} would be returned, instead of
+        {"foo": {"bar": 1}}
+    """
+    repo = Repo()
+    params = {}
+    for dep in repo.index[stage].deps:
```
skshetry, line 108: Note that this may be too slow, you can use [`repo.stage.load_one(name=stage)`](https://github.com/iterative/dvc/blob/7708a8e292c2625d180b8b5569ef8f8be07e34ae/dvc/repo/stage.py#L286) which will only load that particular stage name from `dvc.yaml`.

There's also `repo.stage.get_target()` that is compatible with CLI.

If you are okay with getting multiple stages, there's also `repo.stage.load_all()` and `repo.stage.from_target()`.

dvc/api.py
```diff
@@ -91,6 +93,24 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
...
+    Args:
+        stage (str): Name of the dvc.yaml stage to get params from.
+        flatten (bool, optional): Whether to return flattened dict
+        or not. Defaults False.
+        If True, {"foo.bar": 1} would be returned, instead of
+        {"foo": {"bar": 1}}
+    """
+    repo = Repo()
+    params = {}
+    for dep in repo.stage.load_one(name=stage).deps:
```
skshetry, line 108: This won't work if the working directory is in a different location than the `dvc.yaml` file. This could happen if the script was invoked by `repro/exp-run` and the stage has `wdir` set, or the user is manually invoking them through different directory. 

Do we want to extend the API to pass the path as well?

daavoo: > Do we want to extend the API to pass the path as well?

No strong opinion. I originally wanted to keep API as simple as possible.

Do you think the different workdir would be a common scenario?

dberenbaum:
Will this work for `vars`?

daavoo:
> Btw, this makes me think of a "porcelain" idea that @skshetry was bring up before. Instead of creating `get_params` or exposing the whole `Repo()` to public, we might create a public wrapper methods to mirror the CLI. E.g. for params here, we would create `dvc params show` CLI command (just symmetry and also seems useful) and smth like `params_show("/path/to/repo", ...)` in api.py, which would just do `Repo(path).params.show(...)` and return the result.
> 
> The nice part about this "porcelain" approach is that we can expose all CLI-like functionality through API with familiar workflow, while not exposing `Repo()` and so not being limited by any promises of backward compatibility and not needing to somehow hide all non-public things that we have there (e.g. `repo.stages`, `repo.index`, etc). So it will provide an easilly maintainable python API that is symmetric to CLI workflow, while not creating a maintenance headache for us.

I get the motivation from a maintainer perspective. 

However, from a user perspective, I want Python APIs to provide me a more flexible and "pythonic" experience. 
The porcelain approach (at least for metadata collection) doesn't give me any added value and ties me to internal data structures, which tend to be not very user-friendly.

The initial motivation for this API is to cover a use case not supported by `repo.params.show`, which is to get all the params associated with a specific stage.

To set goals from a user perspective, I would use `gitpython` as aspiration. 
It is a [very popular ](https://pypistats.org/packages/gitpython) "Python API for interacting with git" . The approach is focused on Pythonic experience, not redirecting outputs of CLI git commands. See https://gitpython.readthedocs.io/en/stable/tutorial.html#meet-the-repo-type.

pmrowla:
> To set goals from a user perspective, I would use `gitpython` as an aspiration. It is a [very popular ](https://pypistats.org/packages/gitpython) "Python API for interacting with git" . The approach is focused on Pythonic experience, not redirecting/wrapping raw outputs from CLI git commands. See https://gitpython.readthedocs.io/en/stable/tutorial.html#meet-the-repo-type.

I'm not sure gitpython is the best example here, given that if you want to do anything beyond the basic/common use cases in gitpython (like we do in DVC) you have to just use the direct wrappers around the git CLI commands. The "pythonic" experience in gitpython only covers a very limited set of git functionality, and I wouldn't consider it any different than providing a porcelain API in DVC that only covers the common use cases.

efiop:
@daavoo `get_params` is not very flexible though and can fit just fine into an existing CLI API with `dvc params show`. It doesn't seem like you need to operate on our internal data structures or even know about `Repo()` class. I totally get the desire for a full-scale API to expose Repo and stuff inside, but I just wonder if users really need it now, as they usually just ask for python API alternative to a CLI command. That might be just because CLI is the only thing that is documented right now, but I have to say that working with git through gitpython/pygit2/dulwich I personally also always use porcelain, unless I'm trying to do something very involved.

For example, `get_params`(similar to open/read) is missing a few limbs right now, because you accept a stage name as an argument, but there is no programmatic way to even get a list of stage names right now. There is `dvc stage list` in CLI though and with the suggested porcelain approach, we could expose `stage_list()` from which you could get the stage names and then feed them to `params_show()`. Also `get_params` opens a repo in the current working directory, which is very limiting.

efiop:
Btw, this is a bit unrelated, but regarding the stage name, it also reminds me of a discussion we had some time ago, about potentially exposing more info to the process we are running, e.g. using env variables. Like `DVC_STAGE_NAME` or smth.

daavoo:
> I'm not sure gitpython is the best example here, given that if you want to do anything beyond the basic/common use cases in gitpython (like we do in DVC) you have to just use the direct wrappers around the git CLI commands. The "pythonic" experience in gitpython only covers a very limited set of git functionality, and I wouldn't consider it any different than providing a porcelain API in DVC that only covers the common use cases.

Not sure I follow. Would be similar if we provide the current (or a pruned version of) `Repo` API  . You access everything from a single instance. porcelain as defined above (`params_show(path)`) is a completely detached API for each retrieval of information.

Regardless, perhaps a better example would be then https://pygithub.readthedocs.io/en/latest/examples/Repository.html , in the way that the rest API returns plain JSON stuff  and the added value of pygithub is to return type classes (i.e. `Issue`) with useful methods. 

In dvc case, this would be the difference between porcelain `repo.params.show` (returning plain json) and alternative API, like accessing `repo.index.params` (provides access to `ParamsDependency` instances)

daavoo:
> Btw, this is a bit unrelated, but regarding the stage name, it also reminds me of a discussion we had some time ago, about potentially exposing more info to the process we are running, e.g. using env variables. Like `DVC_STAGE_NAME` or smth.

This is a good point. 

I was actually thinking that would be more valuable to have something like the current `updater`. Something that writes an updates a `json` somewhere (`.dvc/pipeline_status.json` or something) that can serve as source of truth for the status of the current pipeline execution (for each stage in pipeline show skipped, running, finished?).

daavoo:
> Btw, this is a bit unrelated, but regarding the stage name, it also reminds me of a discussion we had some time ago, about potentially exposing more info to the process we are running, e.g. using env variables. Like DVC_STAGE_NAME or smth.

Is there an existing discussion / issue? I opened https://github.com/iterative/dvc/issues/7782

efiop:
> In dvc case, this would be the difference between porcelain repo.params.show (returning plain json) and alternative API, like accessing repo.index.params (provides access to ParamsDependency instances)

@daavoo Note that we return python dict there and not json. We have a pretty good separation of representation in our internals right now and CLI (dvc/command/*) don't usually get anything more complex than plain python values (int/string/dict/etc), so that would be suitable for porcelain as well. Users won't need to json.load or anything like that.

daavoo:
> > In dvc case, this would be the difference between porcelain repo.params.show (returning plain json) and alternative API, like accessing repo.index.params (provides access to ParamsDependency instances)
> 
> @daavoo Note that we return python dict there and not json. We have a pretty good separation of representation in our internals right now and CLI (dvc/command/*) don't usually get anything more complex than plain python values (int/string/dict/etc), so that would be suitable for porcelain as well. Users won't need to json.load or anything like that.

Sorry for lack of clarification. I mean that the internal structure is not really oriented towards end user. 

Example of current output:

```python
params = repo.params.show()
print(result)
{'': {'data': defaultdict(<class 'dict'>, {'params.yaml': {'data': {'prepare': {'split': 0.2, 'seed': 20170428}, 'featurize': {'max_features': 3000, 'ngrams': 2}, 'train': {'seed': 20170428, 'n_est': 100, 'min_split': 64}}}})}}
```

To get what I want, I need to `params[""]["data"]["params.yaml"]`, and this is considering I am using a single parameters file. If I have more complex parameter setup (i.e. multiple file) I would need to do that for each file and then merge.

skshetry:
I am not really for making `Repo` API public, it's too much work and still won't have a better API anyway. 
If what we really need is a pythonic CLI-like wrapper, I think introducing `porcelain` is easier for us, and for users.
I think what we need to have is a high-level but still composable API like `Experiments` where we can house similar APIs (if we think that there might be other APIs  than `get_params`). Monolithic API is probably not a good idea, it'll be too complex.

Similarly, introduce `Pipeline`/`Graph` and `Datasets` to make it easier to work with pipelines and data management at a higher level.

daavoo:
>  because the gitpython Repo class does not actually provide enough "native python" functionality. IMO it ends up being more useful to just have a python API that mirrors the git CLI (like dulwich.porcelain).

Could it be that you have that opinion from a "low level" usage of gitpython (i.e. for internal DVC needs) ? 

I have a different impression from using gitpython at the level at which I would expect some (or even most) users would be using the hypothetical DVC Python.

---

side note because I might have not been clear: All my points are about "collection" of metadata (deps, metrics, outs, params, plots, stages). 

I do believe that porcelain is a good approach for "data management" and "data sync" operations

dvc/api.py
```diff
@@ -92,6 +94,175 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
         return fd.read()
 
 
+def get_params(
+    *targets: str,
+    repo: Optional[str] = None,
+    revs: Optional[Union[str, Iterable[str]]] = None,
+    deps: bool = False,
+    stages: Optional[Union[str, Iterable[str]]] = None,
+) -> Dict:
```
efiop, line 103: Hm, didn't we agree to go with porcelain-like instead? Or am I missing something?

daavoo: Not sure I understood what the API would be in porcelain-like then. What would be different?

efiop: `params_show`?

dvc/repo/params/show.py
```diff
@@ -103,14 +116,26 @@ def _collect_vars(repo, params) -> Dict:
 
 
 @locked
-def show(repo, revs=None, targets=None, deps=False, onerror: Callable = None):
+def show(
+    repo,
+    revs=None,
+    targets=None,
+    deps=False,
```
dberenbaum, line 123: Still not sure I understand this kwarg or when it's useful:

```python
>>> dvc.api.get_params()
{'prepare': {'split': 0.2, 'seed': 20170428}, 'featurize': {'max_features': 200, 'ngrams': 2}, 'train': {'seed': 20170428, 'n_est': 50, 'min_split': 0.01}, 'extra': {'foo': 1}}
>>> dvc.api.get_params(deps=True)
{'train': {'min_split': 0.01, 'n_est': 50, 'seed': 20170428}}
```

dvc/repo/params/show.py
```diff
@@ -103,14 +116,26 @@ def _collect_vars(repo, params) -> Dict:
 
 
 @locked
-def show(repo, revs=None, targets=None, deps=False, onerror: Callable = None):
+def show(
+    repo,
+    revs=None,
```
dberenbaum, line 121:
```suggestion
    rev=None,
```

dvc/api.py
```diff
@@ -92,6 +94,175 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
         return fd.read()
 
 
+def get_params(
+    *targets: str,
+    repo: Optional[str] = None,
+    revs: Optional[Union[str, Iterable[str]]] = None,
```
dberenbaum, line 100: I'm not sure if this is worth changing the structure of the returned dict. Seems like it might be better to support a single rev and have other ways to iterate over revs. WDYT?

dberenbaum: Also, revs passed without repo kwarg returns unexpected results:

```python
>>> dvc.api.get_params(revs="6-prepare-stage")
{'workspace': {'prepare': {'split': 0.2, 'seed': 20170428}, 'featurize': {'max_features': 200, 'ngrams': 2}, 'train': {'seed': 20170428, 'n_est': 50, 'min_split': 0.01}, 'extra': {'foo': 1}}, '6-prepare-stage': {'prepare': {'split': 0.2, 'seed': 20170428}, 'featurize': {'max_features': 100, 'ngrams': 1}, 'train': {'seed': 20170428, 'n_est': 50, 'min_split': 0.01}}}
```

dvc/api.py
```diff
@@ -92,6 +94,175 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
...
+    deps: bool = False,
```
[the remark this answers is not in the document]
daavoo: This kwarg was not working as expected (https://dvc.org/doc/command-reference/params/diff#--deps) when multiple stages share the same `params` file since before the P.R. (haven't bisect yet to before the P.R.).

It's suppose to be useful to prevent loading `parameters` that are present in your params file but are not connected to your pipeline (not indicated as dependencies of any stage)

dvc/api.py
```diff
@@ -92,6 +94,175 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
         return fd.read()
 
 
+def get_params(
+    *targets: str,
```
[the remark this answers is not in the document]
daavoo: No. This is also an issue with previous collection code where multiple stages share the same parameters file

dvc/api.py
daavoo: > Seems like it might be better to support a single rev and have other ways to iterate over revs. WDYT?

Sounds good

daavoo: > Also, revs passed without repo kwarg returns unexpected results:

So, you would expect `workspace` to not be present?

dvc/api.py
```diff
@@ -92,6 +94,175 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
...
+    repo: Optional[str] = None,
+    revs: Optional[Union[str, Iterable[str]]] = None,
+    deps: bool = False,
+    stages: Optional[Union[str, Iterable[str]]] = None,
```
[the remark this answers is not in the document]
daavoo: Fixed it. 

> Do you have a use case in mind to pass multiple stages?

When supporting a single `stage`, the code is the same as for multiple stages. 
I guess the use case would be similar?

dvc/api.py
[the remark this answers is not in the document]
daavoo: Fixed the behavior in separate commit https://github.com/iterative/dvc/pull/7613/commits/fbe95c26ba839d15872355da6196f9d720ab9e8f as it was not really part of the new API added

dvc/api.py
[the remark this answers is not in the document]
daavoo: Fixed

dvc/api.py
daavoo: @dberenbaum I cannot reproduce the output of `dvc.api.get_params(revs="6-prepare-stage")` from the current status of the PR. Could you check?

dvc/api.py
```diff
@@ -92,6 +94,178 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
         return fd.read()
 
 
+def get_params(
```
skshetry, line 97: Until now, `dvc.api` has been about data-management/data-registry. If we have plans to add more API, it might make more sense to add this on a separate namespace: 
```python
from dvc.api.exp import get_params
```

Or,
```python
from dvc.api import Experiments

Experiments().params()
```

dtrifiro: I think that's a good point, the syntax for the second example looks a bit weird to me though

dberenbaum: @skshetry What args/kwargs would you expect `Experiments()` and `params()` to take?

dvc/api.py
dberenbaum: Are you planning to remove revs completely or not?

dvc/api.py
```diff
@@ -92,6 +94,178 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
...
+    *targets: str,
+    repo: Optional[str] = None,
+    revs: Optional[Union[str, Iterable[str]]] = None,
+    deps: bool = False,
+    stages: Optional[Union[str, Iterable[str]]] = None,
```
dberenbaum, line 102: Not sure what's expected in this case:

```console
>>> dvc.api.get_params("src/train.py", stages="train")
{'train': {'min_split': 128, 'n_est': 50, 'seed': 20170428}}
```

I expected it to return an empty dict.

dvc/api.py
daavoo: > Until now, dvc.api has been about data-management/data-registry. If we have plans to add more API, it might make more sense to add this on a separate namespace:
> from dvc.api.exp import get_params

I am not sure it makes sense _for now_ to add a separate namespace. If we extend the API "mirroring" the CLI, all commands are at the top-level namespace (`dvc {X}`), not only data management.
 
 > from dvc.api import Experiments
 
I think this is a good discussion to have but it requires more elaborated discussion and planning, which I believe is out of the scope of this P.R.

dvc/api.py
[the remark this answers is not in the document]
daavoo: > I would expect params from all stages to be included.

I am taking a look.

> So it has no meaning when stages are provided, right?

Correct, `stages` implies `deps` but at a more granular level. I kept `deps` because it was already present but tbh not sure how valuable it is after supporting `stages`

dvc/api.py
daavoo: > I expected it to return an empty dict.

The `targets`/`stages` interaction didn't exist before the P.R. so I wasn't sure how to implement / what's the most intuitive behavior. 

I guess what you expect makes more sense than the current behavior.

dvc/api.py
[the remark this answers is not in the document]
daavoo: > I would expect params from all stages to be included.

It should work as expected now

dvc/api.py
skshetry: > If we extend the API "mirroring" the CLI, all commands are at the top-level namespace (dvc {X}), not only data management.

I thought those porcelain APIs would exist in `dvc.porcelain` namespace. 

> I think this is a good discussion to have but it requires more elaborated discussion and planning, which I believe is out of the scope of this P.R.

I think it's a good time to discuss and choose the way we want to go. Once it's merged, we'll be stuck maintaining an API, or supporting duplicate APIs which might be more confusing.

daavoo: > I thought those porcelain APIs would exist in dvc.porcelain namespace.

What is wrong / what is the difference with the `dvc.api` namespace? 
Why `dvc.api.exp` makes more sense than the parent namespace?

I get the motivation of reusing git terminology but this case seems forced to me. 
I have only seen this term used to refer to a Python API in `dulwich` and our use case looks different, as dulwich explicitly exposes a lower-level public Python API.

> Until now, dvc.api has been about data-management/data-registry

Was this by design? Why name it `dvc.api` if the scope will be limited to data management?

> I think it's a good time to discuss and choose the way we want to go. Once it's merged, we'll be stuck maintaining an API, or supporting duplicate APIs which might be more confusing.

If we name this P.R. `params_show` and save it under `dvc.porcelain`, would still be an issue? Is this the concern based on naming? 

Names aside, the scope seems completely different to me. `Experiments` API as used in the snipped will require, IMO, far more effort to implement and reach consensus.

In addition, I think for some use cases `Experiments` will just be adding an unnecessary layer of complexity (for example, the original use case that motivates the P.R. "get parameters of the current stage")

> What if I write like this:
> 
> exp = Experiments()
> exp.params()
> exp.metrics()
> 
> Is it still weird?

We already use the term `experiments` in a very specific way in DVC and I don't really understand what `Experiments` will return; all `dvc experiments` in the current workspace?

dvc/api.py
daavoo: Updated to match the expected behavior

dvc/api.py
dberenbaum: Sorry, I realized I already asked the same question above 😬 , and I don't want to block the PR. 

One last concern: 

> ```python
> >>> dvc.api.get_params(revs="6-prepare-stage")
> {'6-prepare-stage': {'prepare': {'split': 0.2, 'seed': 20170428}, 'featurize': {'max_features': 100, 'ngrams': 1}, 'train': {'seed': 20170428, 'n_est': 50, 'min_split': 0.01}}}
> ```

I would like this better if passing a str kept the return dict the same, like:

```python
>>> dvc.api.get_params(revs="6-prepare-stage")
{'prepare': {'split': 0.2, 'seed': 20170428}, 'featurize': {'max_features': 100, 'ngrams': 1}, 'train': {'seed': 20170428, 'n_est': 50, 'min_split': 0.01}}
>>> dvc.api.get_params(revs=["6-prepare-stage"])
{'6-prepare-stage': {'prepare': {'split': 0.2, 'seed': 20170428}, 'featurize': {'max_features': 100, 'ngrams': 1}, 'train': {'seed': 20170428, 'n_est': 50, 'min_split': 0.01}}}
```

daavoo: > can we instead require revs to be an iterable instead of allowing a str? 
🤔 What about `stages`? Do we follow same behavior?

Do you think it would be confusing if I implement as you suggested above:

```
>>> dvc.api.get_params(revs="6-prepare-stage")
{'prepare': {'split': 0.2, 'seed': 20170428}, 'featurize': {'max_features': 100, 'ngrams': 1}, 'train': {'seed': 20170428, 'n_est': 50, 'min_split': 0.01}}
>>> dvc.api.get_params(revs=["6-prepare-stage"])
{'6-prepare-stage': {'prepare': {'split': 0.2, 'seed': 20170428}, 'featurize': {'max_features': 100, 'ngrams': 1}, 'train': {'seed': 20170428, 'n_est': 50, 'min_split': 0.01}}}
```

dberenbaum: > 🤔 What about `stages`? Do we follow same behavior?

No, I don't think so. That's one reason why I prefer to require `revs` to be an iterable. Then, there is a difference between `stages` and `revs` args that reflects the differences in their behavior.

Changing `stages` will make it harder to use for simple use cases (since passing a single stage is the primary use case). I'm not sure we need to accept an iterable for `stages`, but it's a nice convenience. 

For `revs`, there's no obvious way to merge multiple revs. I'm also not sure there's a simple use case where I want to retrieve params from a single rev outside the current workspace.

> Do you think it would be confusing if I implement as you suggested above:

I don't mind it if you prefer it, but it seems more complicated and makes less clear why we change the return structure for `revs` but not `stages`.

dvc/api.py
```diff
@@ -92,6 +94,160 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
         return fd.read()
 
 
+def params_show(
```
skshetry, line 97: Wait, why? It’s a getter. Again, I am confused with porcelain/API discussion.

dvc/api.py
daavoo: Updated to support single `rev`

dberenbaum (approved):

dvc/api.py
efiop: @skshetry The argument during the last retro discussion was that this is at least following the CLI conventions. Could consider adding `get_params` as an alias though.

skshetry: Right, but is `dvc.api` the place to have these CLI wrappers?

daavoo: > CLI wrappers?

What do you call CLI wrappers? I think the idea is that we follow CLI *conventions* regarding names, but this doesn't mean that the API is a CLI wrapper. 
This API uses the internal Python API, raises exceptions, and returns a structure meant to be used in Python (instead of the internal structure returned by CLI `--json` options, for example).

dberenbaum: Personally, my vote would be `get_params` because the most immediate use case here is to onboard new users who aren't already familiar with the CLI.

However, I would rather get this merged than continue to discuss naming.

jorgeorpinel:
> Filter with stages.

Could DVC detect what stage you're in based on the name of the file you're in? Then the default (no args) run would auto-filter that one stage's params specifically.

And maybe when loading a single stage's params, the JSON structure shouldn't have the name as parent (so the code never needs to know the stage name). E.g.

```py
# prepare.py
...
params = dvc.api.params_show()
print(json.dumps(params))
{
  "split": 0.2,
  "seed": 20170428
}
```

dvc/api.py
```diff
@@ -214,6 +215,268 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
...
+    *targets: str,
```
jorgeorpinel, line 219: Is it very relevant to pick by param files? Maybe the natural targets should be param names (return simple dict) or even stage names and this could be a secondary optional arg.

daavoo:
> And maybe when loading a single stage's params, the JSON structure shouldn't have its name as key (so the code never needs to know it

@jorgeorpinel We don't include the stage name, this is just how the example-get-started params are set up where there are sections inside the params file matching the name of the stage:

https://github.com/iterative/example-get-started/blob/a470fae5c0ccf71641e10d99e69e97534c3d811e/params.yaml#L1-L4

dvc/api.py
daavoo: > Is it very relevant to pick by param files?

In all `dvc {x} show` / `dvc {x} diff` commands, `targets` are files, i.e. https://dvc.org/doc/command-reference/metrics/show 

> Maybe the natural targets should be param names (return simple dict)

Not sure I follow what this would look like.

daavoo:
> Could DVC detect a default stage based on the name of the file you're in? Then the no-args run would auto-filter that one stage's params specifically. Add `params_show(all=True)` to load all.

This would be quite a big assumption to make and not sure how common users have this setup. 
`stages` argument was added to support the use case of the filter by current stage, I don't think the ambiguity of trying to figure it out the current stage would be worth comparing to requiring users to "hardcode" the stage name in the call

dvc/api.py
dberenbaum: I agree that stage names are likely most useful here, but we are also trying to mimic the CLI. It's a good idea @jorgeorpinel but there has already been a lot of related discussions and I would rather get something merged than try to optimize the order of args.

jorgeorpinel: Makes sense to follow the CLI.

dvc/repo/params/show.py
```diff
@@ -103,14 +119,26 @@ def _collect_vars(repo, params) -> Dict:
...
+    targets=None,
+    deps=False,
+    onerror: Callable = None,
+    stages=None,
```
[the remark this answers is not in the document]
efiop: Does `stages` here accept stage at a custom path? It appears it doesn't, unless I'm missing something.

dvc/api.py
```diff
@@ -214,6 +215,267 @@ def read(path, repo=None, rev=None, remote=None, mode="r", encoding=None):
         return fd.read()
 
 
+def params_show(
```
efiop, line 218: @daavoo Let's please at least put it into a separate file, because it is pretty giant. E.g. `dvc/api/params.py`, keeping the imports the same.

dvc/repo/params/show.py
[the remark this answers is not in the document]
daavoo: > why do we need targets and stages?

Because they are not mutually exclusive. They have to be decoupled to support several scenarios, for example:

- A `stage` (stage_A) uses params from different `targets` (target_A, target_B).

`params_show(target_A, stage=stage_A)`

- Different parts of a single `target` (target_A) can be used for multiple `stages` (stage_A, stage_B).

`params_show(target_A, stage=stage_B)`

> But what if you have stages with same names?

I am not sure I follow. The `targets` and `stages` filters are applied separately, so it would be up to the user to provide unambiguous arguments.

> Does stages here accept stage at a custom path?

What's is a stage at a custom path?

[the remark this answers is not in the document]
efiop: > What's is a stage at a custom path?

Like `path/to/dvc.yaml:mystage`. Because `mystage` could be defined in multiple dvc.yaml at different levels.

tests/func/test_api.py
```diff
@@ -228,3 +229,128 @@ def test_open_from_remote(tmp_dir, erepo_dir, cloud, local_cloud):
         remote="other",
     ) as fd:
         assert fd.read() == "foo content"
+
+
+@pytest.fixture
```
efiop, line 234: Let's also separate the tests, since the code is now. E.g. `tests/func/api/test_params.py`

daavoo: Done

efiop (approved):
Thank you @daavoo ! Let's run with this and change on top if we'll need anything.

PR #7614, opened 2022-04-21
plots: tests: integration test for empty repo

### LINKED ISSUE #7549: plots: integration tests for vscode
The `--json` option for `dvc plots` subcommands needs integration tests that would serve as a contract between dvc and vscode.

Current use cases: 
-  plots with and without some properties - make sure proper JSON is produced
- are no plots in the repo at all
-  verify that data types for JSON structure are as expected by vscode

cc @mattseddon 
Related : iterative/vscode-dvc#1521 iterative/vscode-dvc#1520

pared:

Closes #7549


dvc/commands/plots.py
```diff
@@ -70,8 +70,13 @@ def run(self):
                     "visualization file will not be created."
                 )
 
+            out: str = self.args.out or "dvc_plots"
+
+            renderers_out = (
+                out if self.args.json else os.path.join(out, "static")
+            )
```
pared, line 77: Seems like during extracting of the renderers we dropped saving images to dir by default. I think it's more convenient regarding parsing and potentially obtaining the images from produced HTML if we keep that behavior. @daavoo Was there a particular reason for this change? If there is, we can roll this back and change the tests accordingly.

dvc/render/image_converter.py
```diff
@@ -45,9 +45,7 @@ def convert(
         if path:
             if not os.path.isdir(path):
                 os.makedirs(path, exist_ok=True)
-            src = self._write_image(
-                os.path.abspath(path), revision, filename, data
```
pared, line 49: When creating html with images under `dvc_plots/static` its better if we provide relpath. That way we can easily share the produced HTML with other users.

dvc/commands/plots.py
daavoo: No reason. I assumed that we saved to `dir` by default because there was no support for embedding `base64`. 

After I added support for `base64`, I assumed that if someone wanted to use the "save to dir" behavior, they will explicitly use the `out` argument

dvc/render/convert.py
```diff
@@ -42,8 +44,10 @@ def _group_by_rev(datapoints):
 
 
 def to_json(renderer, split: bool = False) -> List[Dict]:
+    from copy import deepcopy
+
     if renderer.TYPE == "vega":
-        grouped = _group_by_rev(renderer.datapoints)
+        grouped = _group_by_rev(deepcopy(renderer.datapoints))
```
pared, line 50: Before the change we would edit the datapoints in renderer. That would result in vega json datapoints lacking `rev` field. We did not detect that before, because vscode nowadays uses only `--split` option and handles the datapoints by themselves.

dvc/commands/plots.py
pared: Got it, my reasoning for creating `dir` structure was to make HTML easier to read and have direct access to images in `static` folder.

daavoo (commented):
@pared is this ready for review?

pared:
@daavoo yes

tests/integration/plots/test_json.py
```diff
@@ -0,0 +1,246 @@
...
+    )
+    assert (
+        dpath.util.get(data, ["content", "spec", "encoding", "y", "field"])
+        == y
+    )
+
+
+@pytest.mark.vscode
+def test_no_plots(tmp_dir, scm, dvc, capsys):
+    html_path, json_result, split_json_result = call(capsys)
```
daavoo, line 166:
```suggestion
@pytest.mark.vscode
@pytest.mark.parametrize("subcommand", ["show", "diff"])
def test_no_plots(tmp_dir, scm, dvc, capsys, subcommand):
    html_path, json_result, split_json_result = call(capsys, subcommand)
```

tests/integration/plots/test_json.py
```diff
@@ -0,0 +1,246 @@
...
+    verify_vega_props("confusion.json", json_result, **confusion_props)
+    from dvc.utils.fs import remove
+
+    # even if there is no data, call should be successful
+    remove(tmp_dir / ".dvc" / "cache")
+    remove("linear.json")
+    remove("confusion.json")
+    remove("image.png")
+    call(capsys)
+    call(capsys, subcommand="diff")
```
daavoo, line 246: Isn't this covered by `test_no_plots`? See suggestion above
```suggestion
```

daavoo (approved):

tests/integration/plots/test_json.py
pared: Not quite, in case of no plots at all, we get empty results when collecting plots definitions. 
In this case, we have definitions but we don't have the data. That will make a difference down the line when generating visualizations. This part can detect the problem with no `final_props` in `convert.py` while `no-plots` cannot because it never reaches the converter.

dvc/commands/plots.py
daavoo: One thing we overlook here is that users might want to use the embedded behavior and after these changes we don't have a way to opt-out from the `dir` creation.

daavoo: One thing we overlook here is that users might want to use the embedded behavior and after these changes we don't have a way to opt-out from the `dir` creation.

PR #7616, opened 2022-04-21
data: introduce `data.add()` method

skshetry:

### DIFF
--- a/dvc/data/add.py
+++ b/dvc/data/add.py
@@ -1,0 +1,58 @@
+from typing import TYPE_CHECKING, Tuple, Union
+
+from .stage import stage
+from .transfer import transfer
+from .tree import Tree
+
+if TYPE_CHECKING:
+    from dvc.data.meta import Meta
+    from dvc.fs.base import FileSystem
+    from dvc.ignore import DvcIgnore
+    from dvc.objects.db import ObjectDB
+    from dvc.objects.file import HashFile
+    from dvc.types import AnyPath
+
+
+def add(
+    odb: "ObjectDB",
+    fs: "FileSystem",
+    fs_path: "AnyPath",
+    name: str,
+    filter_prefix: Union["AnyPath", Tuple[str, ...]] = None,
+    hardlink: bool = False,
+    dvcignore: "DvcIgnore" = None,
+    upload: bool = False,
+    jobs: int = None,
+    no_progress_bar: bool = False,
+) -> Tuple["Meta", Union["HashFile", "Tree"]]:
+    staging, meta, obj = stage(
+        odb,
+        fs_path,
+        fs,
+        name,
+        upload=upload,
+        jobs=jobs,
+        dvcignore=dvcignore,
+        no_progress_bar=no_progress_bar,
+    )
+
+    obj_ids = {obj.hash_info}
+
+    if filter_prefix is None or isinstance(filter_prefix, tuple):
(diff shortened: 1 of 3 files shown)


dvc/data/add.py
```diff
@@ -0,0 +1,58 @@
...
+
+    if prefix and isinstance(obj, Tree):
+        obj = obj.filter(prefix)
+        obj_ids.update({oid for _, _, oid in obj})
+
+    transfer(
+        staging,
+        odb,
+        obj_ids,
+        shallow=bool(prefix),
```
skshetry, line 54: I am still trying to understand `shallow` here. Let me know if the `shallow` should be a separate option or only set on `prefix` as done here.

skshetry: My understanding is that it ignores the content of the `Tree`'s _leaf_ object. So, if we pass the content of filtered object and the full dir object, it won't try to read from the full dir object, right?

dvc/data/add.py
```diff
@@ -0,0 +1,58 @@
...
+if TYPE_CHECKING:
+    from dvc.data.meta import Meta
+    from dvc.fs.base import FileSystem
+    from dvc.ignore import DvcIgnore
+    from dvc.objects.db import ObjectDB
+    from dvc.objects.file import HashFile
+    from dvc.types import AnyPath
+
+
+def add(
```
skshetry, line 16: What should it return on `filter_prefix`? a filtered tree or a full tree?

dvc/data/add.py
```diff
@@ -0,0 +1,58 @@
...
+    odb: "ObjectDB",
+    fs: "FileSystem",
+    fs_path: "AnyPath",
+    name: str,
```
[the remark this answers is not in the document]
efiop: We shouldn't depend on fs.PARAM_CHECKSUM, because it will go away and doesn't really reflect what is going on here. E.g. hash type should be arbitrary and not depend on fs. E.g. think about sha* migration in 3.0, where it might be passed somewhere here as a name or might be a setting somewhere else, but definitely not in fs.

dvc/data/add.py
efiop: Ideally `filter` should go away completely, as we've discussed before, we need to simply pick particular part of the data tree and pass it down, instead of passing filter all the way down. But that is a bit out of context for this change, I guess.

dvc/data/add.py
efiop: CC @pmrowla

dvc/data/add.py
efiop: Btw, those changes would simplify this PR too. So might be worth waiting.

dvc/data/add.py
pmrowla: > So, if we pass the content of filtered object and the full dir object, it won't try to read from the full dir object, right?

`shallow=bool(prefix)` looks correct to me.

If `shallow=True`, `status.compare_status` will not load a full dir/tree object, and assumes that you are explicitly passing every (raw) obj ID that you want it to query.

So if `obj_ids` here contains all the filtered child obj IDs, we want `shallow=True`.

For the unfiltered case (where `prefix` is unset), we do want it to load the tree (since `obj_ids` will only contain the top level tree obj ID, and has not been updated to contain any of the child obj IDs)

skshetry:
Closing, will reopen in dvc-data repo.

PR #7626, opened 2022-04-25
Revert "dvc.objects: add callbacks for listing and estimating hashes"

skshetry:
Reverts iterative/dvc#7624


skshetry:
Reopened in #7627.

PR #7627, opened 2022-04-25
dvc.objects: add callbacks for listing and estimating hashes

skshetry:
This PR gets rid of Tqdm progress bars inside `dvc.objects`.
Those are replaced with multi-phase callback in `hashes_exist`, and
simple callable that takes `completed` value in
`odb._estimate_remote_size()`.

The `odb.list_hashes_exists()` and `odb.all()` now return iterables,
hence the caller has the responsibility of handling progress.

This is now handled in `dvc.data` by using a custom `Tqdm` progressbar.
This is ugly and is temporary, but my motivation is to remove tqdm from dvc.objects
first. We can gradually get rid of these from `dvc.data` too, but
that requires more knowledge about `dvc.data` and involves more
design work. All in time, I guess. :)

### DIFF
--- a/dvc/data/_progress.py
+++ b/dvc/data/_progress.py
@@ -1,0 +1,32 @@
+from dvc.progress import Tqdm
+
+
+class QueryingProgress(Tqdm):
+    def __init__(self, iterable=None, total=None, name=None, phase="Querying"):
+        msg_part = "cache in " + f"'{name}'" if name else "remote cache"
+        msg_fmt = "{phase} " + msg_part
+
+        self._estimating_msg = msg_fmt.format(phase="Estimating size of")
+        self._listing_msg = msg_fmt.format(phase="Querying")
+        super().__init__(
+            iterable=iterable,
+            desc=msg_fmt.format(phase=phase),
+            total=total,
+            unit="files",
+            unit_scale=False,
+            bar_format=self.BAR_FMT_DEFAULT,
+        )
+
+    def callback(self, phase, *args):
+        total = args[0] if args else self.total
+        completed = args[1] if len(args) > 1 else self.n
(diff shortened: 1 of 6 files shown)


dvc/objects/db.py
```diff
@@ -295,81 +303,48 @@ def _list_hashes_traverse(
...
-                ret = list(itertools.compress(hashes, in_remote))
-                return ret
-
-    def hashes_exist(self, hashes, jobs=None, name=None):
+        with ThreadPoolExecutor(max_workers=jobs or self.fs.jobs) as executor:
+            fs_paths = map(self.hash_to_path, hashes)
+            in_remote = executor.map(self.fs.exists, fs_paths)
+            yield from itertools.compress(hashes, in_remote)
+
+    def hashes_exist(self, hashes, jobs=None, progress=noop):
```
skshetry, line 347: I am trying not to touch this at the current stage, felt like removing the progress bar should be done first.

dvc/data/_progress.py
```diff
@@ -0,0 +1,32 @@
...
+            desc=msg_fmt.format(phase=phase),
+            total=total,
+            unit="files",
+            unit_scale=False,
+            bar_format=self.BAR_FMT_DEFAULT,
+        )
+
+    def callback(self, phase, *args):
+        total = args[0] if args else self.total
+        completed = args[1] if len(args) > 1 else self.n
```
skshetry, line 22: The progress callbacks take `completed`/`current`/`absolute` value rather than a relative/`to_increment` value. Maybe I am overengineering here, but in the future, we may want to throttle callbacks, so I felt like we should use `absolute` values in most places, instead of keeping callbacks guessing.

PR #7628, opened 2022-04-25
fs.put_file: add support for uploading fileobjects

skshetry:
Also changed the callbacks to inherit from our own so that we can
add a few more utilities to it.


dvc/fs/_callback.py
```diff
@@ -1,9 +1,47 @@
...
 class FsspecCallback(fsspec.Callback):
+    """FsspecCallback usable as a context manager, and a few helper methods."""
+
+    def wrap_attr(self, fobj: IO, method: str = "read") -> IO:
+        from tqdm.utils import CallbackIOWrapper
+
+        wrapped = CallbackIOWrapper(self.relative_update, fobj, method)
+        return cast(IO, wrapped)
+
+    def wrap_fn(self, fn: Callable["_P", "_R"]) -> Callable["_P", "_R"]:
```
skshetry, line 22:
```suggestion
    def wrap_fn(self, fn: “Callable[_P, _R]”) -> “Callable[_P, _R]”:
```

PR #7636, opened 2022-04-26
debug

skshetry:

### DIFF
--- a/.github/workflows/packages.yaml
+++ b/.github/workflows/packages.yaml
@@ -3,106 +3,11 @@ on:
   push:
     branches:
     - main
+  pull_request: {}
   release:
     types: [released, prereleased]
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


dvc/data/_progress.py
```diff
@@ -30,3 +30,6 @@ def callback(self, phase, *args):
             # let Tqdm take care of when to update
             self.set_description_str(desc, refresh=False)
         self.update_to(completed, total)
+
+    def __del__(self):
+        return super().__del__()
```
casperdcl, line 35: https://docs.python.org/3/reference/datamodel.html#object.__del__

> If a base class has a `__del__()` method, the derived class's `__del__()` method, **if any**, must explicitly call it to ensure proper deletion of the base class part of the instance.

Doesn't make sense to me how this fixes things.

PR #7643, opened 2022-04-27
`plots`: Fix repo with removed plots

daavoo:
Changes in dvc-render 0.0.5 caused an error.

In https://github.com/iterative/dvc-render/pull/24 the behavior for get_filled_template changed to return an empty string. Previously it returned a valid dictionary with empty `data` field.

Updated `to_json` logic in `dvc` to not call `get_filled_template` at all if there are no `datapoints`.

### DIFF
--- a/tests/integration/plots/test_json.py
+++ b/tests/integration/plots/test_json.py
@@ -235,12 +235,22 @@ def test_repo_with_plots(
         split_json_result,
     )
     verify_vega_props("confusion.json", json_result, **confusion_props)
+
+
+@pytest.mark.vscode
+def test_repo_with_removed_plots(tmp_dir, capsys, repo_with_plots):
     from dvc.utils.fs import remove
+
+    next(repo_with_plots())
 
     # even if there is no data, call should be successful
     remove(tmp_dir / ".dvc" / "cache")
     remove("linear.json")
     remove("confusion.json")
     remove("image.png")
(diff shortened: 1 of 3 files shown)


tests/integration/plots/test_json.py
```diff
@@ -235,12 +235,22 @@ def test_repo_with_plots(
...
 
+    next(repo_with_plots())
+
     # even if there is no data, call should be successful
     remove(tmp_dir / ".dvc" / "cache")
     remove("linear.json")
     remove("confusion.json")
     remove("image.png")
-    call(capsys)
-    call(capsys, subcommand="diff")
```
daavoo, line 246: @pared we only checked that this didn't raise an error but not the expected result. 

I updated `to_json` to return an empty list for vega plots to match the image plots behavior. 

Before https://github.com/iterative/dvc-render/pull/24 , this tests was passing but the returns for vega plots were a vega template without `data`. 

If this is not the expected return format, we need to remove:

https://github.com/iterative/dvc-render/blob/369396f9e833e9b82724b21d0c5f1b853b156d06/src/dvc_render/vega.py#L46-L47

In order to get back to previous behavior

pared (approved):
LGTM.
As discussed off this thread - I agree with @daavoo that we should not reach `to_json` if there is no data to display. Lets create issue for that?

PR #7645, opened 2022-04-27
fs.ssh: remove unused vars

skshetry:


dvc/fs/ssh.py
```diff
@@ -91,7 +87,7 @@ def _prepare_credentials(self, **config):
                 os.path.expanduser(key) for key in raw_keys
             ]
 
-        login_info["timeout"] = config.get("timeout", _SSH_TIMEOUT)
+        login_info["timeout"] = config.get("timeout", 1800)
```
skshetry, line 90: This `timeout` config is unused, as sshfs does not pass it to the connection pool. Even then, it seems `SoftConnectionPool` does not even use timeout at all, so this does not even work at all. :)

efiop (approved):

PR #7649, opened 2022-04-27
fs.utils: add atomic upload function

skshetry:


tests/func/test_remote.py
```diff
@@ -430,21 +430,3 @@ def test_push_incomplete_dir(tmp_dir, dvc, mocker, local_remote):
     assert not remote_odb.exists(out.hash_info)
     assert not remote_odb.exists(file_objs[0])
     assert remote_odb.exists(file_objs[1])
-
-
-def test_upload_exists(tmp_dir, dvc, local_remote):
```
skshetry, line 435: This is testing that if the file upload succeeds or not if that file already exists in the remote. I don't think we need to test any behaviour provided by fsspec.

dvc/stage/cache.py
```diff
@@ -210,51 +212,46 @@ def restore(self, stage, run_cache=True, pull=False):
...
 
-        odb = self.repo.cloud.get_remote_odb(remote)
-        return self._transfer(
-            _log_exceptions(odb.fs.download),
-            odb,
-            self.repo.odb.local,
-        )
+    def pull(self, remote: Optional[str], odb: Optional["ObjectDB"] = None):
+        odb = odb or self.repo.cloud.get_remote_odb(remote)
+        return self.transfer(odb, self.repo.odb.local)
```
skshetry, line 254: Unified `self.transfer()` for both `push` and `pull`.

dvc/stage/cache.py
```diff
@@ -210,51 +212,46 @@ def restore(self, stage, run_cache=True, pull=False):
...
-            )
 
+            src_name = from_fs.path.name(src)
+            parent_name = from_fs.path.name(from_fs.path.parent(src))
+            with FsspecCallback.as_tqdm_callback(
+                desc=src_name,
+                bytes=True,
+                total=-1,
+            ) as cb:
+                func(from_fs, src, to_fs, dst, callback=cb)
```
skshetry, line 244: This now creates a Tqdm progress bar, and uses `fs.utils.copy` instead of `fs.upload`/`fs.download`.

dvc/data/stage.py
```diff
@@ -27,21 +27,22 @@
...
+        stream = HashedStreamReader(fobj)
+        size = fileobj_size(fs, fobj, from_fs_path)
         with FsspecCallback.as_tqdm_callback(
             callback,
             desc=fs_path.name(from_fs_path),
             bytes=True,
             total=size,
         ) as cb:
-            upload_odb.fs.upload(stream, tmp_info, size=size, callback=cb)
+            upload_odb.fs.put_file(stream, tmp_info, size=size, callback=cb)
```
skshetry, line 45: We probably don't need to use `fs.utils.upload` here as it seems to upload to `tmp_info` already. WDYT @efiop?

efiop (approved):

dvc/fs/utils.py
```diff
@@ -1,20 +1,37 @@
...
-    from .base import FileSystem
-    from .types import AnyPath
+    from dvc.types import AnyPath
+
+    from .base import AnyFSPath, FileSystem
 
 logger = logging.getLogger(__name__)
 
 
+def upload(
```
efiop, line 19:
```suggestion
def atomic_upload(
```

?

skshetry: I added `atomic=True/False` flag to it for now, so it's no longer `atomic` by default. So I think I'll just name it `upload` for now..

efiop (approved):

PR #7650, opened 2022-04-27
dvc: use Trie instead of PathStringTrie

efiop:
PathStringTrie is error prone, as it gives you an illusion that you are working with paths, while it can't really handle important edge cases, which results in obscure bugs.

For example, when one sets `trie["/"] = "foo"`, we often expect `trie.longest_prefix("/path")` to return `"foo"`, while it is not actually like that, because `"/"` gets resolved into `("", "")` internal key, while `"/path"` is resolved into `("", "path")` and thus longest_prefix is `("",)`, which breaks our assumptions.

Instead, we should use regular `Trie` that doesn't create any illusions and forces user to be fully aware of his actions. We switched to `Trie`s long time ago in `scmrepo` and our data tree.

Additional logic change here is that our `Trie` is now based on the keys/paths relative to the repo root in both dvcignore and repofs, which saves us building additional nodes and making additional steps while traversing for long paths.

### DIFF
--- a/dvc/ignore.py
+++ b/dvc/ignore.py
@@ -6,13 +6,13 @@ from itertools import chain, groupby, takewhile
 
 from pathspec.patterns import GitWildMatchPattern
 from pathspec.util import normalize_file
+from pygtrie import Trie
 
 from dvc.fs.base import AnyFSPath, FileSystem
 from dvc.fs.local import localfs
 from dvc.pathspec_math import PatternInfo, merge_patterns
 from dvc.scheme import Schemes
 from dvc.types import List, Optional
-from dvc.utils.collections import PathStringTrie
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


dvc/ignore.py
```diff
@@ -191,8 +193,15 @@ def __init__(self, fs, root_dir):
             ignore_subrepos=True,
         )
 
-    def _update_trie(self, dirname: str, trie: PathStringTrie) -> None:
-        old_pattern = trie.longest_prefix(dirname).value
+    def _get_key(self, path):
+        parts = self.fs.path.relparts(path, self.root_dir)
+        if parts == (".",):
+            parts = ()
+        return parts
```
efiop, line 200: Obviously need to put into a common utility for repofs/dvcfs/gitfs/dvcignore/etc. Will be introduced separately.

PR #7664, opened 2022-04-29
plots: image converter return absolute paths

pared:

In #7549 we rolled back to the behavior of writing images, instead of encoding them. I made it DVC's responsibility to provide proper path to the renderer. However, without knowing where we create HTML, it is impossible to do that. We need to modify `dvc-render` to be able to evaluate relative paths of the images. Marking as WIP, as we need a fix on `dvc-render` side first and a release.

### DIFF
--- a/tests/integration/plots/test_json.py
+++ b/tests/integration/plots/test_json.py
@@ -67,13 +67,11 @@ def verify_image(tmp_dir, version, filename, content, html_path, json_result):
         tmp_dir / JSON_OUT / f"{version}_{filename}"
     ).read_bytes() == content
 
-    assert (
-        os.path.join("dvc_plots", "static", f"{version}_{filename}")
-        in html_content
-    )
+    assert os.path.join("static", f"{version}_{filename}") in html_content
 
     # there should be no absolute paths in produced HTML
-    assert str(tmp_dir) not in html_content
+    # TODO uncomment once dvc-render is adjusted
+    # assert str(tmp_dir) not in html_content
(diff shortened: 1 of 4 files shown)


tests/integration/plots/test_json.py
```diff
@@ -67,13 +67,11 @@ def verify_image(tmp_dir, version, filename, content, html_path, json_result):
...
-        in html_content
-    )
+    assert os.path.join("static", f"{version}_{filename}") in html_content
 
     # there should be no absolute paths in produced HTML
-    assert str(tmp_dir) not in html_content
+    # TODO uncomment once dvc-render is adjusted
+    # assert str(tmp_dir) not in html_content
     assert (
         tmp_dir / "dvc_plots" / "static" / f"{version}_{filename}"
```
daavoo, line 76: Does this belong to integration tests? HTML content is not really used by vscode, right?

pared:
As iterative/dvc-render#39 takes a bit of discussion I think we should merge this one. Even if its not desired behavior, it works at least, contrary to the current `main`. We will fix relpaths once `dvc-render` counterpart gets released.
cc @daavoo

tests/integration/plots/test_json.py
pared: This is true, currently I implemented html checks to make sure everything works as expected. In further work we will need to split that. I am thinking how to do it without too much test duplication.

daavoo (approved):

PR #7665, opened 2022-04-30
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/packages.yaml
+++ b/.github/workflows/packages.yaml
@@ -6,8 +6,13 @@ on:
   release:
     types: [released, prereleased]
   workflow_dispatch:
+permissions:
+  contents: read
+
 jobs:
   deb:
+    permissions:
+      contents: write  # for actions/upload-release-asset to upload release asset
     runs-on: ubuntu-18.04
     steps:
     - uses: actions/checkout@v3
@@ -39,6 +44,8 @@ jobs:
         asset_name: dvc_${{ github.event.release.tag_name }}_amd64.deb
         asset_content_type: binary/octet-stream
   rpm:
+    permissions:
+      contents: write  # for actions/upload-release-asset to upload release asset
     runs-on: ubuntu-18.04
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/packages.yaml
```diff
@@ -6,8 +6,13 @@ on:
   release:
     types: [released, prereleased]
   workflow_dispatch:
+permissions:
+  contents: read
+
 jobs:
   deb:
+    permissions:
+      contents: write  # for actions/upload-release-asset to upload release asset
```
pmrowla, line 15: Is there documentation on what the different permission scopes actually correspond to? I'm confused as to the difference between `contents` and `packages` (it seems like the relevant permission here would be packages?)

.github/workflows/tests.yaml
```diff
@@ -80,6 +80,8 @@ jobs:
         file: ./coverage.xml
         fail_ci_if_error: false
   notify:
+    permissions:
+      contents: none
```
pmrowla, line 84: Should we be setting `statuses: read` here?

also, is there a reason you aren't setting the explicit `contents: read` permission for the tests job?

pmrowla (commented):
Thanks for the PR. Could you please clarify some questions about the permissions which you are setting?

pmrowla:
I'm also unsure as to whether or not there is any risk for us without this PR? We already require manual approval before actions can be run for PR's from new contributors, and PR's require explicit manual approval from a maintainer before anything can be merged into `main`. Likewise, the packaging/asset upload actions are only run on commits to `main`.

skshetry:
From the linked article, it seems it's only an issue for `pull_request_target`, which we don't use.

skshetry:
Closing (see https://github.com/iterative/dvc/pull/7665#issuecomment-1154024563)

PR #7671, opened 2022-05-01
s3fs: Adds support for SSE client keys

### LINKED ISSUE #6141: Support provision of Client Keys for SSE-C Server Side Encryption
Hey! I'm using dvc[s3]==2.3.0 with python 3.8 to version files to a minIO s3 instance.
Server Side Encryption with Client provided keys (SSE-C, see: [Amazon Doc](https://docs.aws.amazon.com/AmazonS3/latest/userguide/specifying-s3-c-encryption.html) and [minIO Doc](https://docs.min.io/minio/baremetal/security/encryption/server-side-encryption-sse-c.html#encrypt-an-object-using-sse-c)) should be used but currently it isn't possible to specify the client key, so `dvc push` (and `dvc pull`) will fail when only setting `sse = AES256` since there is no KMS configured.

This could possibly be solved by extending the config with e.g. a key named `sse-customer-key-md5` that will be set to the S3 header field `X-Amz-Server-Side-Encryption-Customer-Key-MD5`. Example:

```
['remote "test-remote"']
    url = s3://test-remote/
    endpointurl = https://localhost:9000
    access_key_id = minio
    secret_access_key = minio123
    sse = AES256
    sse-customer-key-md5 = c2VjcmV0ZW5jcnlwdGlvbmtleWNoYW5nZW1lMTIzNBo=
```

ap-kulkarni:
Fixes #6141


ap-kulkarni:
I will raise a separate PR for documentation and link it here once the code changes are agreed by reviewers.

dtrifiro (approved):

ap-kulkarni:
Created PR at [dvc.org#3498](https://github.com/iterative/dvc.org/pull/3498) for documentations changes

dvc/fs/s3.py
```diff
@@ -103,9 +105,28 @@ def _prepare_credentials(self, **config):
...
+            sse_customer_key = base64.b64decode(config.get("sse_customer_key"))
+        sse_customer_algorithm = config.get("sse_customer_algorithm")
+        if not sse_customer_algorithm and sse_customer_key:
+            sse_customer_algorithm = "AES256"
+        sse_params = SSEParams(
+            server_side_encryption=config.get("sse"),
+            sse_customer_algorithm=sse_customer_algorithm,
+            sse_customer_key=sse_customer_key,
+            sse_kms_key_id=config.get("sse_kms_key_id"),
+        )
```
efiop, line 126: I wonder if these belong in s3fs?

ap-kulkarni: Not sure exactly what the question meant. However, I would go ahead explain the snippet tagged. This `SSEParams` is just a helper class which accepts these parameters and provides a method `to_kwargs()` which  returns a dictionary with appropriate key names required by aws libraries. Before these changes, the key names were hardcoded like below
```
        additional["ServerSideEncryption"] = config.get("sse")
        additional["SSEKMSKeyId"] = config.get("sse_kms_key_id")
```
Now, this is achieved by appropriately constructing the SSEParams object, and then updating the `additional` dictionary with the dictionary obtained from `to_kwargs()` like below

```
        additional.update(sse_params.to_kwargs())

```

dvc/config_schema.py
```diff
@@ -152,6 +152,8 @@ class RelPath(str):
                     "ssl_verify": Any(Bool, str),
                     "sse": str,
                     "sse_kms_key_id": str,
+                    "sse_customer_algorithm": str,
```
jorgeorpinel, line 155: Why not reuse `sse` for this? Instead of introducing `sse_customer_algorithm` that is. `sse` also accepts `AES256` I think, and is described as the "SSE encryption algo" to use in docs -- that could be either from KMS or customer-provided.

ap-kulkarni: The sole reason of introducing new parameter here was that AWS themselves have kept separate parameters for SSE and SSE-C. Let's say in future AWS brings a scheme where they'll offer to encrypt the data twice. Once using their key and once using users' key. In such case, we will need both parameters. May be a far fetched imagination but not impossible. Also, if the PR gets merged, dvc users can opt not to configure this parameter since the only value AWS now supports is `AES256` which we add by default if `sse_customer_key` is configured. 
It is not like I am of strong opinion to introduce a new parameter here. If majority of reviewers think that it is okay to just reuse the `sse` param, I can make that change as well.

jorgeorpinel: > AWS themselves have kept separate parameters for SSE and SSE-C

Makes sense to reflect their config schema then. Thanks for all the context!

ap-kulkarni:
Let me know if this PR requires any more work.

dberenbaum:
Sorry, @ap-kulkarni, we haven't had a chance to take a close look. We will try to do that soon and let you know if anything is needed 🙏 .

daavoo (changes requested):
Hi @ap-kulkarni ! Unfortunately, during the period of the review process, some of the files modified here (`dvc/fs/s3.py`)  were refactored out to a separate repo (https://github.com/iterative/dvc-objects/blob/main/src/dvc_objects/fs/implementations/s3.py) . You would need to move the changes there 🙏

ap-kulkarni:
Thank you @daavoo. What would be the best course of action then? Should this PR be closed and fresh sets of PRs be raised? or should I just raise PR for changes in `s3.py` in dvc-objects project and keep this PR?

daavoo:
> Should this PR be closed and fresh sets of PRs be raised? or should I just raise PR for changes in s3.py in dvc-objects project and keep this PR? 

I would just remove `s3.py` and test changes from this P.R. and open a new P.R. in `dvc-objects` project

efiop:
@ap-kulkarni Thank you for your patience!

Regarding tests, please do add them to this PR. You can run them locally to ensure that they work using `--enable-s3` flag. We are reorganizing our code base right now, so long term your tests from this PR will be migrated to https://github.com/iterative/dvc-s3

PR #7687, opened 2022-05-03
Fail on sync when there is no match for glob.

### LINKED ISSUE #7492: push/pull: globs that match no targets should fail
When using `--glob` with `push` or `pull`, a glob that matches no targets will instead act on all targets. In other words, `dvc push/pull --glob 'nomatches*'` will be the same as `dvc push/pull/repro` without any target argument.

```
$ mkdir repo
$ mkdir dvcremote
$ cd repo
$ git init
Initialized empty Git repository in /private/tmp/glob/repo/.git/
$ dvc init -q
$ dvc remote add -d default ../dvcremote
Setting 'default' as a default remote.
$ echo foo > foo
$ dvc add foo
100% Adding...|████████████████████████████████████████████████|1/1 [00:00, 52.98file/s]

    git add foo.dvc .gitignore

To enable auto staging, run:

        dvc config core.autostage true
$ dvc push --glob 'z*'
1 file pushed
$ rm foo
$ dvc pull --glob 'z*'
A       foo
1 file added
```

I would expect the commands to fail or push/pull nothing if there are no `--glob` matches.

tirkarthi:
Closes #7492 

### DIFF
--- a/dvc/exceptions.py
+++ b/dvc/exceptions.py
@@ -349,3 +349,9 @@ class CircularImportError(DvcException):
 class PrettyDvcException(DvcException):
     def __pretty_exc__(self, **kwargs):
         """Print prettier exception message."""
+
+
+class GlobDoesNotMatchError(DvcException):
+    def __init__(self, glob):
+        msg = f"Glob {glob} has no matches."
+        super().__init__(msg)
--- a/dvc/repo/pull.py
+++ b/dvc/repo/pull.py
@@ -1,6 +1,7 @@
 import logging
 from typing import TYPE_CHECKING, Optional
 
+from dvc.exceptions import GlobDoesNotMatchError
 from dvc.repo import locked
 from dvc.utils import glob_targets
 
@@ -30,6 +31,9 @@ def pull(
         targets = [targets]
 
     expanded_targets = glob_targets(targets, glob=glob)
+
+    if targets and not expanded_targets:
+        raise GlobDoesNotMatchError(targets)
 
     processed_files_count = self.fetch(
         expanded_targets,
--- a/tests/func/test_import.py
+++ b/tests/func/test_import.py
@@ -9,7 +9,11 @@ from scmrepo.git import Git
 from dvc.config import NoRemoteError
 from dvc.data.db import ODBManager
 from dvc.dvcfile import Dvcfile
-from dvc.exceptions import DownloadError, PathMissingError
+from dvc.exceptions import (
+    DownloadError,
+    GlobDoesNotMatchError,
+    PathMissingError,
+)
 from dvc.stage.exceptions import StagePathNotFoundError
 from dvc.system import System
 from dvc.utils.fs import makedirs, remove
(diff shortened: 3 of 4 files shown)


dvc/repo/pull.py
```diff
@@ -31,6 +32,9 @@ def pull(
 
     expanded_targets = glob_targets(targets, glob=glob)
```
daavoo, line 33: Could we raise the exception inside `glob_targets` instead?

Apart from pull and push, st is only being used in `dvc add` and I believe raising exception would also make sense there

tirkarthi: Thanks, done.

tests/func/test_import.py
```diff
@@ -266,6 +270,21 @@ def test_pull_wildcard_imported_directory_stage(tmp_dir, dvc, erepo_dir):
     assert (tmp_dir / "dir_imported123").read_text() == {"foo": "foo content"}
 
 
+def test_push_pull_glob_no_match(
```
daavoo, line 273: Probably this test makes more sense to be located in `tests/func/test_data_cloud.py`.

tirkarthi: Thanks, makes sense. Moved to test_data_cloud.py

tests/func/test_data_cloud.py
```diff
@@ -570,3 +571,18 @@ def test_target_remote(tmp_dir, dvc, make_remote):
...
+    tmp_dir.gen("foo", "foo")
+    erepo_dir.add_remote(config=local_remote.config)
+
+    with erepo_dir.chdir():
+        with pytest.raises(GlobDoesNotMatchError):
+            dvc.push(targets="b*", glob=True)
+
+        with pytest.raises(GlobDoesNotMatchError):
+            dvc.pull(targets="b*", glob=True)
+        assert not os.path.exists(erepo_dir.dvc.stage_cache.cache_dir)
```
daavoo, line 588: I think this is all we need for covering the changes:
 
```suggestion
def test_push_pull_glob_no_match(
    tmp_dir, dvc,
):
    with pytest.raises(GlobDoesNotMatchError):
        dvc.push(targets="b*", glob=True)

    with pytest.raises(GlobDoesNotMatchError):
        dvc.pull(targets="b*", glob=True)
```

tests/func/test_add.py
```diff
@@ -292,6 +293,11 @@ def test_add_filtered_files_in_dir(
         assert stage.outs[0].def_path in expected_def_paths
 
 
+def test_add_glob_no_result_exception(dvc):
+    with pytest.raises(GlobDoesNotMatchError):
+        dvc.add(["invalid*"], glob=True)
+
+
 class TestAddExternal(TestAdd):
     @pytest.fixture
```
efiop, line 302: Let's just add one unit test for `glob_targets` and not introduce add/pull/etc func tests, please.

dvc/exceptions.py
```diff
@@ -349,3 +349,9 @@ def __init__(self, dep, a, b):
 class PrettyDvcException(DvcException):
     def __pretty_exc__(self, **kwargs):
         """Print prettier exception message."""
+
+
+class GlobDoesNotMatchError(DvcException):
+    def __init__(self, glob):
+        msg = f"Glob {glob} has no matches."
+        super().__init__(msg)
```
efiop, line 357: Let's keep the exception in `dvc/utils/__init__.py`, closer to where it is used.

tirkarthi: Importing `DvcException` at top for custom exception raises circular import error. I have used `DvcException` itself.

efiop (approved):
Thank you, @tirkarthi ! 🙏

tirkarthi:
Thanks @efiop and @daavoo

PR #7690, opened 2022-05-03
exp: show: Include `fs_path` and `use_cache` in `--json`.

### LINKED ISSUE #7575: `exp show`: Distinguish between different dependencies
@daavoo @dberenbaum 

Quick ask - in Studio we can distinguish DVC-tracked and intermediate artifacts (to show only them by default), can we add those signals here too please?

_Originally posted by @shcheklein in https://github.com/iterative/vscode-dvc/issues/1183#issuecomment-1062436116_

daavoo:
Closes #7575


dvc/repo/experiments/show.py
```diff
@@ -45,20 +45,22 @@ def _collect_experiment_commit(
             res["params"] = params
 
         res["deps"] = {
-            dep.def_path: {
+            relpath(dep.fs_path): {
```
daavoo, line 48: Fixes https://github.com/iterative/dvc/issues/7575#issuecomment-1134242482

pmrowla: I think this should be relative to the repo root and not cwd? Otherwise the output will change depending on where you run `exp show` in the repo

daavoo: Updated

pmrowla (approved):
LGTM assuming this works for the vscode side

dvc/repo/experiments/show.py
```diff
@@ -45,20 +45,21 @@ def _collect_experiment_commit(
...
         }
-
         res["outs"] = {
-            out.def_path: {
+            relpath(out.fs_path, repo.root_dir): {
                 "hash": out.hash_info.value,
                 "size": out.meta.size,
                 "nfiles": out.meta.nfiles,
+                "use_cache": out.use_cache,
+                "is_data_source": out.stage.is_data_source,
```
mattseddon, line 62: [Q] Do I have everything I need to work this out:

> ```
> for dep in res[“deps”].values():
>     for out in res[“outs”].values():
>         if dep[“fs_path”] == out[“fs_path”]:
>             # THIS IS AN INTERMEDIATE ARTIFACT
>             dep["intermediate"] = True
>             out["intermediate"] = True
>             if out[“use_cache”]:
>                 # DEP IS DVC TRACKED
>                 dep["use_cache"] = True
>             else:
>                 # DEP IS GIT-TRACKED 
>                 dep["use_cache"] = False
> ```

The answer is yes

> Given that the rules about deciding what files should be included appear to be currently unclear and product-dependent I would prefer not to include this logic in DVC.

This does make sense to me. However, there will be much less chance of providing consistency across dependencies if we push this logic out to the consumers.

mattseddon (approved):
This will be enough for us right now as we're working on implementing:

> The first step is to get all of the deps from exp show shown in the columns tree and experiments webview.

as per https://github.com/iterative/vscode-dvc/issues/1183#issuecomment-1118051435

I think we should revisit when we get to:

> Once extra information has been added to the DVC output we will provide it to the user (in some way to be determined) so they can easily filter out what they need.

and/or when we redesign the data that the extension needs

dvc/repo/experiments/show.py
shcheklein: @mattseddon where does this snippet come from?

I see that `is_data_source` is included which is what we need I believe, @daavoo could you remind please what is the definition of the `out.stage.is_data_source`?

shcheklein:
@daavoo thanks for addressing this!

> Example code on how to use the new fields:

I'm not sure that example is 100% correct. What happens if I have only a few `.dvc` (`dvc add`) in the project? There will be no `deps` at all, right? But there will be `outs` that should be marked as "data sources".

shcheklein (changes requested):
Please review the comment regarding the `.dvc` files ... at least we should fix an example.

In general - why would I need that double cycle at all btw, why can't I just take all data source outs (I hope it includes imports) ... and for final outputs - take outs that don't have deps.

dvc/repo/experiments/show.py
mattseddon: https://github.com/iterative/dvc/issues/7575#issuecomment-1116456065

mattseddon: Correct code is in the description?

daavoo:
>  I'm not sure that example is 100% correct. What happens if I have only a few .dvc (dvc add) in the project? There will be no deps at all, right? But there will be outs that should be marked as "data sources".

The example is from the **current** DVC perspective where `deps` is what is shown in the table and artifacts not connected to the pipeline are not shown. 

VSCode has access to both `deps` and `outs` and can process and decide differently.

> Please review the comment regarding the .dvc files ... at least we should fix an example.

Reverse the loop. Iterate on `outs`:

```
for out_name, out in res["out"].items():
    if not out["is_data_source"]:
        if out_name in res["deps"]:
            # This is an intermediate artifact
        else:
            # This is a "final output"
```

> why can't I just take all data source outs (I hope it includes imports)

You can see the example above. Imports are included (test here https://github.com/iterative/dvc/blob/48fe30410494796d9cb38bc57b45216e6c7dec4d/tests/func/experiments/test_show.py#L691-L716).

>  why would I need that double cycle at all btw,

You kind of will still double looping because of `and for final outputs - take outs that don't have deps` (the `in` is still a loop).

shcheklein:
Got it, thanks @daavoo !

shcheklein (approved):

PR #7697, opened 2022-05-04
fs: add more fsspec methods

skshetry:
`put`/`from_json`/`to_json` are major APIs that remain now. `open`/`walk`/`find`/`ls`/`info`/`rm` are not yet fsspec-compliant, which I am working on next. :)

### DIFF
--- a/dvc/fs/base.py
+++ b/dvc/fs/base.py
@@ -1,3 +1,4 @@
+import datetime
 import logging
 import os
 import shutil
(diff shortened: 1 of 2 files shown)


dvc/fs/local.py
```diff
@@ -107,6 +107,9 @@ def rm_file(self, path):
     def rm(self, path, recursive=False, maxdepth=None):
         remove(path)
 
+    def cp_file(self, path1, path2, **kwargs):
```
skshetry, line 110: Right now, `cp` and `cp_file` in LocalFileSystem is same, i.e non-recursive. I'll fix that later. :)

dvc/fs/base.py
```diff
@@ -184,6 +185,15 @@ def open(
     ) -> "IO":  # pylint: disable=arguments-differ
         return self.fs.open(path, mode=mode, encoding=encoding, **kwargs)
 
+    def read_block(
```
efiop, line 188: Do we need all of these methods, if we are not using it anyway? Why not just inherit from AbstractFileSystem and get them for free?

skshetry: Maybe this one and others (`head`/`tail`/`sizes`/`du`) would work that way, but others require implementation. Since this is a wrapper, I prefer being explicit for now.

In the future, we may just use `__getattr__()` hack. :)

PR #7703, opened 2022-05-05
exp init: fixes #7534; simplifies/updates exp init --live

### LINKED ISSUE #7534: exp init: Revisit `type="dl"`

dberenbaum:
Still need to fix tests, but I thought this might be easier than explaining the changes correctly.


dvc/commands/experiments/init.py
```diff
@@ -42,6 +42,10 @@ def run(self):
         cmd = parse_cmd(self.args.command)
         if not self.args.interactive and not cmd:
             raise InvalidArgumentError("command is not specified")
+        if self.args.live and (self.args.metrics or self.args.plots):
```
daavoo, line 45: Could you clarify the reasoning for making it mutually exclusive?

dvc/repo/experiments/init.py
```diff
@@ -251,7 +248,7 @@ def init(
         metrics_no_cache=compact([context.get("metrics"), live_metrics]),
         plots_no_cache=compact([context.get("plots"), live_plots]),
         force=force,
-        **{"checkpoints" if with_live else "outs": compact([models])},
+        **{"checkpoints" if type == "dl" else "outs": compact([models])},
```
daavoo, line 251: I guess this breaks API but cleary `type "dl"` is no longer a good name and could be just `checkpoint`

daavoo: Could we alias the user facing option, without removing old version, so it's backwards compatible?

dvc/commands/experiments/init.py
dberenbaum: We are targeting simple use cases here, and I'm trying to mimic how we expected people to use `dvc stage add --live` before deprecation.. I think it makes sense to expect that users are either using dvclive for metrics and plots or tracking themselves. Obviously, there are use cases where people might want both, or they might be tracking images or something with dvclive that's not captured with `--live`, but I think this is better than assuming users want all of them. Otherwise, it becomes hard to do a basic `dvc exp init --live python src/train.py` scenario.

dvc/repo/experiments/init.py
dberenbaum: Agreed. I guess I can alias it.

dberenbaum: Actually, I don't think I can have hidden/aliased choices, so probably the best we can do is leave them both as user-facing.

dberenbaum: Or we can remove `dl`. It's new, it's not used in docs outside of the cmd ref, and it's not a command that's likely to be automated.

dvc/commands/experiments/init.py
daavoo: > Otherwise, it becomes hard to do a basic dvc exp init --live python src/train.py scenario.

But if you remove it from the prompt, there is no need to make it mutually exclusive, right? 

Unless I'm missing something, if we reach this condition would be because user has explicitly passed `live` **and** one of the others

dberenbaum:
Okay, tests are ready. Sorry, should have put in draft mode/held off on assigning you.

dberenbaum:
> We are missing test using explicit `--live` and `--metrics`/`--plots` together

Added.

daavoo (approved):

dvc/repo/experiments/init.py
```diff
@@ -200,21 +199,19 @@ def init(
...
-    with_live = type == "dl"
+    with_live = "live" in overrides
 
     if interactive:
         defaults = init_interactive(
             validator=partial(validate_prompts, repo),
             defaults=defaults,
-            live=with_live,
             provided=overrides,
+            live=with_live,
```
skshetry, line 209: Let's drop `live` kwarg, since it depends on `overrides`/`provided` values from CLI rather than the mode/type.

You can replace with something like following in https://github.com/iterative/dvc/blob/4b6f2306faaa051f595e56e216e1d5b4f1a1e3c1/dvc/repo/experiments/init.py#L86-L89
```python
output_keys = ["models"]
if "live" not in provided:
    output_keys.extend(["metrics", "plots"])
outputs_prompts = lremove(provided.keys(), output_keys)
```

dvc/repo/experiments/init.py
```diff
@@ -200,21 +199,19 @@ def init(
...
             stream=stream,
         )
     else:
         if with_live:
```
skshetry, line 213: If we go with https://github.com/iterative/dvc/pull/7703/files#r867850140, we will only have a single place where that variable is used. :)
```suggestion
        if "live" in overrides:
```

skshetry (commented):
@dberenbaum, changes look good to me. I have 2 suggestions above. I can also handle if needed, lmk. 🙂

dberenbaum:
> @dberenbaum, changes look good to me. I have 2 suggestions above. I can also handle if needed, lmk. 🙂

Thanks! Updated the PR.

dvc/commands/experiments/init.py
```diff
@@ -190,12 +188,11 @@ def add_parser(experiments_subparsers, parent_parser):
...
     experiments_init_parser.add_argument(
         "--live",
-        help="Path to log dvclive outputs for your experiments"
-        f" (default: {CmdExperimentsInit.DVCLIVE})",
+        help="Path to log dvclive outputs for your experiments",
     )
     experiments_init_parser.add_argument(
         "--type",
-        choices=["default", "dl"],
+        choices=["default", "checkpoint"],
```
jorgeorpinel, line 195: Needs doc update 🙂 (at least in `exp init --type` ref.)

PR #7705, opened 2022-05-06
objects: do not inherit from DvcException

skshetry:
These errors will now go off with a bang 💥 💣.

### DIFF
--- a/dvc/objects/errors.py
+++ b/dvc/objects/errors.py
@@ -1,15 +1,8 @@
-from dvc.exceptions import DvcException
-
-
-class ObjectError(DvcException):
+class ObjectFormatError(Exception):
     pass
 
 
-class ObjectFormatError(ObjectError):
-    pass
-
-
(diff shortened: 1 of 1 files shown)


dvc/objects/errors.py
```diff
@@ -1,15 +1,8 @@
-from dvc.exceptions import DvcException
-
-
-class ObjectError(DvcException):
-    pass
-
-
-class ObjectFormatError(ObjectError):
+class ObjectFormatError(Exception):
```
skshetry, line 1: I am trying to create a hierarchy of exception classes. I think we should try to have a hierarchy when the exceptions are related, but otherwise avoid base exceptions. This forces us to be explicit about catching related exceptions instead of some common abstract exceptions (eg: `DvcObjectsBaseError`, etc.)

See https://github.com/jelmer/dulwich/blob/master/dulwich/errors.py.

Issue #7587, opened 2022-04-19
exp: implement internal queue.kill support


pmrowla:
dvc-task/exp queue currently support `experiments.queue.remove` for removing inactive experiments. `experiments.queue.kill` API call for cancelling/terminating individual active experiments is still needed.

Job termination requires:

* Terminate `exp run` subprocess via dvc-task `proc.kill`
* Celery tasks related to the running experiment should not be [revoked](https://docs.celeryq.dev/en/stable/userguide/workers.html#commands), they should run to completion
    * Once the exp run child process is terminated, it should be treated as any other failed run (and the relevant cleanup/partial collection celery tasks should still be scheduled and run)

karajan1001:
There are still some related works
1. unit test code coverage for the `kill` method.
2. implement the `kill` flag in `shutdown`.

karajan1001:
close due to #7714 and #7799

PR #7714, opened 2022-05-09
exp: implement internal queue.kill support(#7587)

karajan1001:
fix: #7587
1. Add implement kill method for local queue class
2. Add a test to make sure the following job will be success after the
   original job was killed.

### DIFF
--- a/dvc/repo/experiments/queue/base.py
+++ b/dvc/repo/experiments/queue/base.py
@@ -171,6 +171,14 @@ class BaseStashQueue(ABC):
         """
 
     @abstractmethod
+    def kill(self, revs: str) -> None:
+        """Kill the specified running entries in the queue.
+
+        Arguments:
+            revs: Stash revs or running exp name to be killed.
+        """
+
+    @abstractmethod
     def shutdown(self, kill: bool = False):
         """Shutdown the queue worker.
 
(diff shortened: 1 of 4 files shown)


tests/unit/repo/experiments/queue/test_local.py
```diff
@@ -35,3 +35,35 @@ def test_shutdown_active_tasks(test_queue, mocker):
...
+        if result_bar.status == "RUNNING":
+            break
+        if time.time() > timeout:
+            raise AssertionError()
+
+    assert result_foo.status == "PENDING"
+    test_queue.kill("bar")
+
+    while True:
+        if result_foo.status == "SUCCESS":
```
karajan1001, line 66: Here, even if we killed the process "bar", the `result_bar.status` still returns "SUCCESS", only  `celery.control.revoke` would make it fail, and this is why the following "foo" process will continue.

dvc/repo/experiments/queue/base.py
```diff
@@ -170,6 +170,14 @@ def get_result(self, entry: QueueEntry) -> Optional[ExecutorResult]:
         This method blocks until the specified item has been collected.
         """
 
+    @abstractmethod
+    def kill(self, name: str) -> None:
```
pmrowla, line 174: Can we clarify what `name` is expected to be here? I think we probably need it to be a `revs` argument (which can be either a `-n` name or a stash revision) similar to how `BaseStashQueue.remove()` works.

In the `LocalCeleryQueue` implementation, we will need to iterate over the active `QueueEntry`'s to find the entry that matches the specified stash rev or name, and then kill the appropriate process (which is named according to `entry.stash_rev` in the `setup_exp` task).

pmrowla (approved):

pmrowla:
Looks good, just squash or cleanup your commits to get rid of the "review fix" type messages and we can merge this

PR #7715, opened 2022-05-09
TqdmCallback: make it lazy

skshetry:
Makes progress bar lazy, so this won't instantiate tqdm progress bars right away (unless `set_size`/`relative_update`/`absolute_update` are called).

As we don't use `rich` progressbar and we may lift the place where we create tqdm progress bars, I am not fixing it for now.

https://github.com/iterative/dvc/pull/7597#discussion_r867788542

### DIFF
--- a/dvc/fs/_callback.py
+++ b/dvc/fs/_callback.py
@@ -1,7 +1,9 @@
+from contextlib import ExitStack
 from functools import wraps
 from typing import IO, TYPE_CHECKING, Any, Dict, Optional, TypeVar, cast
 
 import fsspec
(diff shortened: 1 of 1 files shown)


dtrifiro:
Thanks, by doing some quick benchmarking, I'm seeing some performance improvements when adding a large (10k files) directory:
```console
$ time dvc add directory # without patch
real	0m49.129s
user	0m45.815s
sys	0m12.943s
$ time dvc add directory
real	0m27.788s
user	0m24.585s
sys	0m7.760s
```

dtrifiro (approved):

PR #7717, opened 2022-05-10
analytics: dummy analytic server, configurable host with envvar

skshetry:
This commit adds a dummy server to test analytics,
and an environment variable to allow changing to a
different host. This also makes it possible to run
undetached in the same process, similar to in updater with a simple
kwarg change.

Similar to what's done in
https://github.com/iterative/terraform-provider-iterative/pull/493.


dvc/analytics.py
```diff
@@ -2,12 +2,12 @@
...
 import os
 
-from .env import DVC_NO_ANALYTICS
+from .env import DVC_ANALYTICS_HOST, DVC_NO_ANALYTICS
 
 logger = logging.getLogger(__name__)
 
 
-def collect_and_send_report(args=None, return_code=None):
+def collect_and_send_report(args=None, return_code=None, detach: bool = True):
```
skshetry, line 10: Similar to how we have in updater:
https://github.com/iterative/dvc/blob/477f266dc561a35f433b24bc29e56de8f0dbfbee/dvc/updater.py#L86-L93

skshetry: I also propose to introduce `DVC_DAEMON_DISABLE` envvar to run `updater` and `analytics` in-process.

daavoo: `DVC_NO_DAEMON` ?

daavoo (commented):
Not sure how feasible but couldn't this env var be used "maliciously" ? Seems like we are exposing an easy way to redirect analytics traffic without users noticing

dvc/analytics.py
```diff
@@ -65,7 +70,7 @@ def send(path):
     """
     import requests
 
-    url = "https://analytics.dvc.org"
+    url = os.environ.get(DVC_ANALYTICS_HOST, "https://analytics.dvc.org")
```
daavoo, line 73: Maybe handle as an optional debug flag instead of env var? So it would need to be explicitly passed to the command

skshetry:
Closing, as there are some open questions like https://github.com/iterative/dvc/pull/7717#pullrequestreview-971946649.
We could come again to this in the future.

PR #7727, opened 2022-05-11
dvc.objects: move hash_info, state and meta into dvc.objects

skshetry:


efiop:
For the record: windows failure was unrelated and already fixed in upstream.

PR #7734, opened 2022-05-12
dvc.objects: extend diskcache to handle pickle errors

### LINKED ISSUE #7726: push: Unsupported pickle protocol: 5 on python 3.7.11
# Bug Report

## Description

When attempting to push to remote (Google Cloud Storage) using DVC version `2.10.2` and Python version `3.7.11` I encounter the following error:

```console
$ dvc push     
ERROR: unexpected error - unsupported pickle protocol: 5              

```

In my tests on python 3.8 this issue does not happen.

### Reproduce

Install DVC version `2.10.2` with Python version `3.7.11` and attempt to push to remote.

### Expected

Expected to pickle successfully and push the pickled data to remote (Google Cloud Storage).

### Environment information

```console
$ dvc doctor
DVC version: 2.10.2 (conda)
Platform: Python 3.7.11 on Darwin-21.2.0-x86_64-i386-64bit
Supports:
        gs (gcsfs = 2022.1.0),
        webhdfs (fsspec = 2022.1.0),
        https (aiohttp = 3.8.1, aiohttp-retry = 2.4.6)
Cache directory: apfs on /dev/disk1s1s1
Caches: local
Remotes: gs
Workspace directory: apfs on /dev/disk1s1s1
Repo: dvc, git
```

skshetry:
This commit decouples the UI representation with the exception by:
1. extending diskcache to raise a proper DiskError on pickling errors.
2. handling the error in `main` function where we provide a nice
   error message and hints.

Since we no longer can use DvcException in dvc.objects, this had
to be done.

To reduce duplications and error, now `dvc.objects` will provide
the caching infrastructure instead of depending on the `diskcache`.
This makes sense as `dvc.state` is already in `dvc.objects`,
and since `dvc.data` is anyway going to depend on `dvc.objects`,
`dvc.data.ODBIndex` now uses this extended version of `Cache` too.

While I was at it, the new `Cache` was also extended to default
to a constant protocol instead of `pickle.HIGHEST_PROTOCOL`.

Also added the test for UI messages.

Closes #7726.


dvc/objects/cache.py
```diff
@@ -0,0 +1,54 @@
...
+    """Extended to handle pickle errors and use a constant pickle protocol."""
+
+    def __init__(
+        self,
+        directory: str = None,
+        timeout: int = 60,
+        disk: disk = Disk,  # pylint: disable=redefined-outer-name
+        **settings: Any,
+    ) -> None:
+        settings.setdefault("disk_pickle_protocol", 4)
```
skshetry, line 51: Instead of defaulting to `pickle.HIGHEST_PROTOCOL`, this defaults to a constant value.

dvc/objects/cache.py
```diff
@@ -0,0 +1,54 @@
...
+        super().__init__(f"Could not open disk: {directory}")
+
+
+def translate_pickle_error(fn):
+    @wraps(fn)
+    def wrapped(self, *args, **kwargs):
+        try:
+            return fn(self, *args, **kwargs)
+        except (pickle.PickleError, ValueError) as exc:
+            if isinstance(exc, ValueError) and not str(exc).startswith(
```
pmrowla, line 22: based on https://github.com/iterative/dvc/issues/7726 it looks like we need to just do a substring search for `pickle`, apparently there are `ValueError`s raised in 3.7 w/messages that don't start with `pickle protocol`

pmrowla (commented):
Other than handling the exception string from https://github.com/iterative/dvc/issues/7726 LGTM

dvc/objects/cache.py
skshetry: Thanks @pmrowla. Done in a66425f. We now search for  `pickle protocol` substring.

Issue #5802, opened 2021-04-12
`stage add`: fails with dirs that will be created later by the `cmd`(s)


jorgeorpinel:
# Bug Report

## Description

If the stage being defined creates a dir and files in this dir are stage outputs, `stage add` will fail (because it can't write the corresponding .gitignore file).

### Reproduce

```console
$ dvc stage add -n hidir -o dir/hi 'mkdir dir && echo hi > dir/hi'
ERROR: unexpected error - [Errno 2] No such file or directory: '/.../dir/.gitignore'
```

### Expected

I guess the full path to the .gitignore in question should be created if needed. Or maybe create it at `repro`? Not sure how, but I would expect the stage definition not to fail for this.

### Environment information

**Output of `dvc doctor`:**

```console
$ dvc doctor
DVC version: 2.0.17 (pip)
---------------------------------
Platform: Python 3.6.9 on Linux-5.4.72-microsoft-standard-WSL2-x86_64-with-Ubuntu-18.04-bionic
Supports: gdrive, hdfs, http, https, s3, ssh, oss
Cache types: <https://error.dvc.org/no-dvc-cache>
Caches: local
Remotes: None
Workspace directory: ext4 on /dev/sdb
Repo: dvc, git
```

dberenbaum:
@skshetry This also can cause `dvc exp init` to fail:

```
$ dvc exp init --model models/model.h5 python src/train.py
ERROR: unexpected error - [Errno 2] No such file or directory: '/Users/dave/repo/models/.gitignore'

Having any troubles? Hit us up at https://dvc.org/support, we are always happy to help!
```

dberenbaum:
@skshetry I don't want to slow down other work for this, but this seems like a fairly serious bug for `exp init`. It actually breaks the example in https://dvc.org/doc/command-reference/exp/init#example-interactive-mode. What do you think about priority and solution?

skshetry:
@dberenbaum, for `exp init`, we can consider creating the parent directory of those outputs, where the `.gitignore` file will be added.

To solve the issue on `stage add` too, we need to fix how we create `.gitignore` files. We need to change the _gitignore-generator_ to backtrack up to the dvc's root directory, likewise on `dvc remove`.

dberenbaum:
> However, we can also choose not to generate `.gitignore` entries on `exp init`/`stage add` in these edge cases, as successive `dvc repro` and `dvc exp run` will try to generate them as well.

[...]

PR #7740, opened 2022-05-12
exp init: don't add outs to .gitignore

dberenbaum:
Addresses https://github.com/iterative/dvc/issues/5802#issuecomment-1115709992 by not adding to `.gitignore` during `exp init` since it breaks for subdirectories. Adding to `.gitignore` is still done on `exp run` / `repro`.

### DIFF
--- a/dvc/repo/experiments/init.py
+++ b/dvc/repo/experiments/init.py
@@ -253,7 +253,6 @@ def init(
 
     with _disable_logging(), repo.scm_context(autostage=True, quiet=True):
         stage.dump(update_lock=False)
-        stage.ignore_outs()
         initialized_deps = init_deps(stage)
         if params:
             repo.scm_context.track_file(params)
--- a/tests/func/experiments/test_init.py
+++ b/tests/func/experiments/test_init.py
@@ -175,8 +175,6 @@ def test_when_params_is_omitted_in_interactive_mode(tmp_dir, scm, dvc):
     scm._reset()
     assert scm.is_tracked("dvc.yaml")
     assert not scm.is_tracked("params.yaml")
-    assert scm.is_tracked(".gitignore")
-    assert scm.is_ignored("models")
 
 
 def test_init_interactive_params_validation(tmp_dir, dvc, capsys):
@@ -286,8 +284,6 @@ def test_init_default(tmp_dir, scm, dvc, interactive, overrides, inp, capsys):
     scm._reset()
     assert scm.is_tracked("dvc.yaml")
     assert scm.is_tracked("params.yaml")
-    assert scm.is_tracked(".gitignore")
-    assert scm.is_ignored("models")
     out, err = capsys.readouterr()
 
     assert not out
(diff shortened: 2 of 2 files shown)


daavoo (approved):
Out of the scope of the P.R. but was wondering why exp init doesn't use `stage.add`?

daavoo (changes requested):
Need to update tests:

```
assert scm.is_tracked(".gitignore")
```

dberenbaum:
> Need to update tests:

Sorry, forgot to remove those lines before before pushing. It should pass now.

> Out of the scope of the P.R. but was wondering why exp init doesn't use `stage.add`?

@skshetry can answer, but I'm guessing from https://github.com/iterative/dvc/blob/37a20e922f065dc7c2814e56d9d8ed1c5ed97b6f/dvc/repo/experiments/init.py#L254 that it's because we need to modify some things like disabling output.

skshetry:
> Out of the scope of the P.R. but was wondering why exp init doesn't use `stage.add`?

Previously, we used to ask for confirmation in an interactive mode before creating the stage. `exp init` also auto-stages `dvc.yaml` and `params.yaml` file which `stage add` does not do by default (unless `core.autostage=true`).

Now, this PR is another reason for not using `stage add`. `stage add` also auto-commits the file, whereas `dvc exp init` may require more processing before committing the file (if not now, in the future anyway), so I don't think `stage add` is a good fit.

We could extend `stage add` for sure, but dumping stages are complex and can be done in a lot of ways, I don't want to overcomplicate it with all the complexities that may be introduced by `exp init`.

dberenbaum:
I'm fine with keeping the scope of this PR as small as possible for now. Could one of you please review now that the tests have been fixed?

skshetry:
@dberenbaum, I am a bit hesitant to change the behaviour here, although no strong opinion. I'm okay with:

1. either git-ignoring what we can and leave the rest to `repro`,
2. git-ignoring by traversing upwards (it's going to require some changes in scmrepo, so definitely not a straightforward fix),
3. or, generate intermediate directories.

DVC expects the path to the output to exist in a lot of places (more in relation to stage's working directory, but usually that's where `.dvc` files are created by default). WDYT of 1 and 3, or a combination of both options?

dberenbaum:
@skshetry Can you explain why you are hesitant (and why dvc should generate `.gitignore` files at all when [creating stages](https://github.com/iterative/dvc/issues/5802#issuecomment-1116565862))?

I opened a PR for 3 in https://github.com/iterative/dvc/pull/7751. That would solve the immediate issue, although the rationale for adding to `.gitignore` during stage creation is still unclear to me.

skshetry:
> @skshetry Can you explain why you are hesitant (and why dvc should generate `.gitignore` files at all when [creating stages](https://github.com/iterative/dvc/issues/5802#issuecomment-1116565862))?
> 
> I opened a PR for 3 in #7751. That would solve the immediate issue, although the rationale for adding to `.gitignore` during stage creation is still unclear to me.

The only concern is that the users may `git add` accidentally in between, the `.gitignore` prevents that.

dberenbaum:
Good point @skshetry. I'll close this one for now since the issue was addressed in #7752. We can come back to it if needed.

PR #7742, opened 2022-05-13
dvc.objects.hash: return a string of the hash instead of an integer

wictory:
I got the following error, so I fixed in order to proceed with I was doing. 

```
...
...
  File "/stuff/src/dvc/dvc/output.py", line 548, in save                                                                                                                                                            
    _, self.meta, obj = ostage(                                                                                                                                                                                     
  File "/stuff/src/dvc/dvc/data/stage.py", line 364, in stage                                                                                                                                                       
    _, meta, obj = _stage_file(                                                                                                                                                                                     
  File "/stuff/src/dvc/dvc/data/stage.py", line 58, in _stage_file                                                                                                                                                  
    meta, hash_info = hash_file(fs_path, fs, name, state=state)                                                                                                                                                     
  File "/stuff/src/dvc/dvc/objects/hash.py", line 133, in hash_file                                                                                                                                                 
    assert ".dir" not in hash_info.value                                                                  
TypeError: argument of type 'int' is not iterable                                                   

```

### DIFF
--- a/dvc/objects/hash.py
+++ b/dvc/objects/hash.py
@@ -80,7 +80,7 @@ def _hash_file(
 
     if hasattr(fs, name):
         func = getattr(fs, name)
-        return func(fs_path), info
+        return f"{func(fs_path):032x}", info
 
     if name == "md5":
         return file_md5(fs_path, fs, callback=callback), info


pmrowla:
`hash_info.value` should never be getting set to an integer, can you elaborate on your setup? What dvc command is leading to this condition in the first place?

skshetry:
@pmrowla, this is in `_hash_file()`, which should return a string. `file_md5` and other ETag/Content-MD5 are already a string, and it makes sense to be consistent with value returned from `fs.checksum()` (which returns an integer).

pmrowla:
> @pmrowla, this is in `_hash_file()`, which should return a string. `file_md5` and other ETag/Content-MD5 are already a string, and it makes sense to be consistent with value returned from `fs.checksum()` (which returns an integer).

In our base `FileSystem`, `checksum()` is typed to return a string
https://github.com/iterative/dvc/blob/dc1d3e880aa2f0779e8ad18de6afc5543f5a3947/dvc/fs/base.py#L275-L276

If that's supposed to be an integer, then I think this change should be explicit that if `name` is `checksum`, then we do the string conversion. It is not obvious at all what is going on when we essentially just have `str(getattr(...))`

Also it seems like we probably need to clarify what exactly the expected conversion from an fsspec integer checksum to a DVC `HashInfo` should be.

wictory:
When quickly check, I get this exception when I run `dvc repro` and `dvc status`, there could be other situations too.

In the situation I have: `func=<bound method FileSystem.checksum of <dvc.fs.http.HTTPFileSystem object at 0x7f04fcaf3700>>`, and `fs_path='http://hostname:port/file.zip'`.

When doing some backtracking I got to the following [line](https://github.com/fsspec/filesystem_spec/blob/a8829696d341e62ca420fcde166434bf10dc68d4/fsspec/spec.py#L612) (in a fsspec repo). Judging from the docstring, the checksum isn't really "well defined".

I will make another "creative" proposal to this.

dvc/objects/hash.py
```diff
@@ -80,7 +80,8 @@ def _hash_file(
 
     if hasattr(fs, name):
```
karajan1001, line 81: Typing can check all these errors.
```
hasattr(fs, name) -> (str, str):
```

dtrifiro:
Hi @wictory,
could you please add a full traceback and command with verbose flag (`-v`) and the output of `dvc doctor`?

skshetry:
The corresponding code has moved to [`iterative/dvc-objects`](https://github.com/iterative/dvc-objects) repository. Opened https://github.com/iterative/dvc-objects/issues/56 for further investigation.

Closing, will need further investigation in https://github.com/iterative/dvc-objects/issues/56. Please feel free to join the discussion.

PR #7750, opened 2022-05-16
Catch correct exception class in params.show()

Suor:


dvc/repo/params/show.py
```diff
@@ -117,8 +119,8 @@ def show(repo, revs=None, targets=None, deps=False, onerror: Callable = None):
     # Hide workspace params if they are the same as in the active branch
     try:
         active_branch = repo.scm.active_branch()
-    except (TypeError, NoSCMError):
-        # TypeError - detached head
+    except (SCMError, NoSCMError):
```
skshetry, line 122: I think this may still get raised on a `gitpython` backend, which probably should be handled upstream.

Suor: You have this construction in `metrics.show()` too, and it doesn't have `TypeError` there.

Suor: BTW, maybe moving this to commands code will improve code - this feels like a presentation logic.

pmrowla (approved):

PR #7757, opened 2022-05-17
dvc.objects: move HashedStreamReader to objects, use it for calculating fobj_md5

skshetry:
Now HashedStreamReader will be used to calculate md5 in fobj_md5/hash_file.

Also:
* adds in `bytes_md5` that may be useful for testing.
* adds a hasher-registry with hashlib
* adds a `text=` kwarg to be able to specify to use dos2unix or not
  (useful for testing)
* makes `localfs` the default filesystem when `fs` arg is not passed
  (makes it easier to test in the REPL)


dvc/objects/hash.py
```diff
@@ -16,52 +20,101 @@
...
             break
-        chunk = data if binary else dos2unix(data)
-        hash_md5.update(chunk)
+    return stream.hash_value
 
 
 def file_md5(
     fname: "AnyFSPath",
-    fs: "FileSystem",
+    fs: "FileSystem" = localfs,
```
skshetry, line 96: Added `localfs` as a default fs, useful for testing in repl.

dvc/objects/hash.py
```diff
@@ -16,52 +20,101 @@
...
     callback: "FsspecCallback" = DEFAULT_CALLBACK,
+    text: Optional[bool] = None,
```
skshetry, line 98: Added `text` to change dos2unix behaviour if we want, useful for testing.

dvc/objects/hash.py
```diff
@@ -16,52 +20,101 @@
...
     return data.replace(b"\r\n", b"\n")
 
 
-def _fobj_md5(
-    fobj: BinaryIO,
-    hash_md5: "hashlib._Hash",
-    binary: bool,
-    chunk_size: int = 2**20,
-) -> None:
+def get_hasher(name: str) -> "hashlib._Hash":
```
skshetry, line 27: Hash registry of sorts.

dvc/objects/hash.py
```diff
@@ -16,52 +20,101 @@
...
+    try:
+        return getattr(hashlib, name)()
+    except AttributeError:
+        return hashlib.new(name)
+
+
+class HashedStreamReader(io.IOBase):
```
skshetry, line 34: I'd name this `HashFile` but that's already taken. If you have any better names, please do suggest.

dvc/objects/hash.py
```diff
@@ -16,52 +20,101 @@
...
+    # for which we may need to buffer till DEFAULT_CHUNK_SIZE.
+    assert chunk_size >= DEFAULT_CHUNK_SIZE
+    stream = HashedStreamReader(fobj, text=text)
     while True:
-        data = fobj.read(chunk_size)
+        data = stream.read(chunk_size)
         if not data:
             break
-        chunk = data if binary else dos2unix(data)
-        hash_md5.update(chunk)
```
skshetry, line 48: If we were using >=3.7, we could:
```python
while data := stream.read(chunk_size):
    pass
return stream.hash_value
```

dvc/objects/hash.py
```diff
@@ -16,52 +20,101 @@
...
+        super().__init__()
+
+    def readable(self) -> bool:
+        return True
+
+    def tell(self) -> int:
+        return self.fobj.tell()
+
+    @cached_property
+    def _reader(self):
```
skshetry, line 54: I think we should just use buffered interface, i.e fobj.read.

skshetry: Added in 5749e73.

dvc/objects/hash.py
skshetry: Okay, renamed into `HashStreamFile`.

efiop (approved):

Issue #7590, opened 2022-04-19
exp queue: monitor queue status


pmrowla:
Support monitoring queue status and viewing output of running experiments

* [x] `dvc queue status`
* [x] `dvc queue attach`
* [x] `dvc queue logs`

Optional/under discussion:

Remove queued exps from `exp show` by default and implement `exp show --queue` option to include queued exps in the table

dberenbaum:
Does this include how to deal with failed experiments? We need the ability to see which experiments failed instead of having them "disappear."

Sorry for changing requirements here, but can `dvc queue attach` show all output up to the current point instead of streaming only new output? This would also be really useful for completed and failed experiments. Otherwise, I think it's hard to see what's happening in the experiments, and we may need a separate command like `dvc queue log`.

PR #7758, opened 2022-05-17
exp queue: monitor queue status

karajan1001:
fix: #7590
1. Add two new command `queue status` and `queue attach`
2. Add unit test for the CLI

Status is too complex for a CLI PR. Would implement this part later.

### DIFF
--- a/dvc/commands/queue/attach.py
+++ b/dvc/commands/queue/attach.py
@@ -1,0 +1,44 @@
+import argparse
+import logging
+
+from dvc.cli.command import CmdBase
+from dvc.cli.utils import append_doc_link
+
+logger = logging.getLogger(__name__)
+
+
+class CmdQueueAttach(CmdBase):
+    """Attach outputs of a exp task in queue."""
+
+    def run(self):
+        self.repo.experiments.celery_queue.attach(
+            rev=self.args.experiment,
+            encoding=self.args.encoding,
+        )
+
+        return 0
+
+
+def add_parser(queue_subparsers, parent_parser):
+    QUEUE_ATTACH_HELP = "Attach outputs of a experiment task in queue."
+    queue_attach_parser = queue_subparsers.add_parser(
+        "attach",
+        parents=[parent_parser],
+        description=append_doc_link(QUEUE_ATTACH_HELP, "queue/attach"),
+        help=QUEUE_ATTACH_HELP,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+    queue_attach_parser.add_argument(
+        "-e",
+        "--encoding",
+        help=(
+            "Text encoding for redirected output. Defaults to"
+            "`locale.getpreferredencoding()`."
+        ),
+    )
+    queue_attach_parser.add_argument(
+        "experiment",
+        help="Experiments in queue to attach.",
+        metavar="<experiment>",
+    )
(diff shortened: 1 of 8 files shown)


dvc/commands/queue/attach.py
```diff
@@ -0,0 +1,52 @@
...
+            "`locale.getpreferredencoding()`."
+        ),
+    )
+    queue_attach_parser.add_argument(
+        "-i",
+        "--sleep_interval",
+        type=int,
+        metavar="<sleep_interval>",
+        help="Sleep interval for follow iterations (when waiting for output).",
+    )
```
pmrowla, line 46: I don't think we need to expose this in the UI

dvc/repo/experiments/queue/base.py
```diff
@@ -147,6 +147,38 @@ def clear(self) -> List[str]:
...
+            commit = self.scm.resolve_commit(rev)
+            return datetime.fromtimestamp(commit.commit_time)
+
+        for queue_entry in self.iter_active():
+            result.append(
+                {
+                    "Rev": queue_entry.stash_rev,
+                    "Name": queue_entry.name,
+                    "timestamp": _get_timestamp(queue_entry.stash_rev),
+                    "Status": "Running",
```
pmrowla, line 166: For internal APIs (and not UI) these keys should all be lowercased for consistency

dvc/commands/queue/status.py
```diff
@@ -0,0 +1,42 @@
...
+            Mapping[str, Optional[str]]
+        ] = self.repo.experiments.celery_queue.status()
+        all_headers = ["Rev", "Name", "Created", "Status"]
+        td = TabularData(all_headers)
+        for exp in result:
+            created = format_time(exp.get("timestamp"))
+            td.append(
+                [exp["Rev"], exp.get("Name", ""), created, exp["Status"]]
+            )
+        td.render()
```
pmrowla, line 28: I think we might want to sort these by timestamp, but that doesn't have to go in this PR (since we'll probably up with different sorting/filtering flags later)

dvc/commands/queue/status.py
```diff
@@ -0,0 +1,42 @@
+import argparse
+import logging
+from typing import List, Mapping, Optional
+
+from dvc.cli.command import CmdBase
+from dvc.cli.utils import append_doc_link
+from dvc.compare import TabularData
+
+from ..experiments.show import format_time
```
skshetry, line 9:
```suggestion
from dvc.experiments.show import format_time
```

pmrowla: I think the original here was correct, the helper is from `commands.experiments.show` (and not `dvc.repo.experiments`).

If anything though, this kind of formatting method should maybe go into `dvc.cli.utils` instead?

dvc/repo/experiments/queue/local.py
```diff
@@ -260,6 +260,21 @@ def shutdown(self, kill: bool = False):
...
             self.celery.control.shutdown()
 
+    def attach(
+        self,
+        rev: str,
+        encoding: Optional[str] = None,
+    ):
+        from dvc.ui import ui
+
+        queue_entry: Optional[QueueEntry] = self.get_queue_entry_by_names(
```
pmrowla, line 270: we probably need a way to filter out active only or queued only when looking up exps by name (we should only be able to attach to active exps)

pmrowla (approved):

PR #7764, opened 2022-05-17
objects.fs: use split in Path.name

dtrifiro:
calling `parts()` to get name requires a lot of unnecessary splitting, which quickly adds up when dealing with large datasets.

### DIFF
--- a/dvc/objects/fs/path.py
+++ b/dvc/objects/fs/path.py
@@ -80,7 +80,7 @@ class Path:
         )
 
     def name(self, path):
-        return self.parts(path)[-1]
+        return self.flavour.basename(path)
 
     def suffix(self, path):
         name = self.name(path)


dtrifiro:
Waiting for `dvc-bench` run: https://github.com/dtrifiro/dvc-bench/actions/runs/2338354293

efiop:
Btw, alternative might be to make `parts` a generator, but that would require a bit more changes around the code base, but probably a long term correct solution.

dvc/objects/fs/path.py
```diff
@@ -80,7 +80,7 @@ def parents(self, path):
         )
 
     def name(self, path):
-        return self.parts(path)[-1]
+        return path.rsplit(self.flavour.sep, maxsplit=1)[-1]
```
efiop, line 83: This doesn't handle `foo/` unlike the old implementation 🙁

dtrifiro: Updated the implementation, also added some tests, currently waiting for benchmarks

dvc/objects/fs/path.py
```diff
@@ -80,7 +80,18 @@ def parents(self, path):
...
+
+        parent, name = self.flavour.split(path)
+
+        if name:
+            return name
+
+        if parent == path:
+            return ""
+
+        return self.name(parent)
```
efiop, line 94: why not
```suggestion
        return self.flavour.basename(path)
```
then?

efiop: But it breaks the tests that are testing arguably incorrect behaviour 😄 I just pointed out that old implementation was handling it differently. Matching `os.path.basename` behaviour is the correct thing to do, so we can just use it directly and not even have to test it specifically.

PR #7778, opened 2022-05-18
fs/data: move repo related utils to dvc

skshetry:
This moves get_fs_config/get_fs_cloud and ODBManager to DVC as they require DVC related utilities, and we don't need them strictly in dvc-objects/dvc-data.
This also adds support to ReferenceHashFile to load from any filesystem, using it's module name and filename.
Now we can get rid of DVCFileSystem dependency in ReferenceHashFile. This is a temporary solution, we really need a registry.

(I am not quite sure if this works in pyinstaller mode, but I think it should, it's how fsspec does it if I understand correctly. Anyway, I am willing to take risk here.)

### DIFF
--- a/dvc/data/reference.py
+++ b/dvc/data/reference.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING, Optional
 
 from dvc.objects.errors import ObjectFormatError
 from dvc.objects.file import HashFile
+from dvc.objects.fs import FS_MAP, LocalFileSystem
 
 if TYPE_CHECKING:
     from dvc.objects.db import ObjectDB
@@ -20,6 +21,7 @@ class ReferenceHashFile(HashFile):
     PARAM_HASH = "hash"
     PARAM_CHECKSUM = "checksum"
     PARAM_FS_CONFIG = "fs_config"
+    PARAM_FS_CLS = "fs_name"
 
     def __init__(
         self,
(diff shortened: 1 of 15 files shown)


dvc/data/reference.py
```diff
@@ -57,11 +59,16 @@ def to_bytes(self):
         # ReferenceHashFiles should currently only be serialized in
         # memory and not to disk
         fs_path = self.fs_path
+        fs_cls = type(self.fs)
+        mod = None
+        if fs_cls not in FS_MAP:
+            mod = ".".join((fs_cls.__module__, fs_cls.__name__))
```
skshetry, line 65: Will only use this if it is not in FS_MAP, i.e. not implemented in dvc.objects.fs.

dvc/data/reference.py
```diff
@@ -57,11 +59,16 @@ def to_bytes(self):
...
+        fs_cls = type(self.fs)
+        mod = None
+        if fs_cls not in FS_MAP.values() and fs_cls != LocalFileSystem:
+            mod = ".".join((fs_cls.__module__, fs_cls.__name__))
         dict_ = {
             self.PARAM_PATH: fs_path,
             self.PARAM_HASH: self.hash_info,
             self.PARAM_CHECKSUM: self.checksum,
             self.PARAM_FS_CONFIG: self.config_tuple(self.fs),
+            self.PARAM_FS_CLS: mod,
```
efiop, line 71: This feels weird to save the cls itself. How about we just save the protocol (e.g. `dvc` for `dvcfs`)?

efiop: Discussed with @skshetry, turns out that's how fsspec's to_json/from_json behaves too. So keeping it.

PR #7785, opened 2022-05-19
dvc: remove dvc-data, fix imports and tests

skshetry:
* Removes `dvc-data`
* Fixes imports and tests
* Adds a dependency to dvc-data.


setup.cfg
```diff
@@ -75,7 +75,7 @@ install_requires =
     scmrepo==0.0.22
     dvc-render==0.0.5
     dvclive>=0.7.3
-    dvc-objects==0.0.1
+    dvc-data==0.0.1
```
skshetry, line 78: I removed direct dependency on `dvc-objects` for now, so that it'd be easier to release them.

PR #7786, opened 2022-05-19
exp: speed up repro execution with untracked directories in workspace

dtrifiro:
When large untracked directories are present in the workspace, a lot of time is spent collecting untracked files in `scm.status` and `scm.is_dirty` (which also calls `status`). `scmrepo==0.0.23` adds `untracked_files=no` flag which avoids collecting untracked files.

### DIFF
--- a/dvc/repo/experiments/__init__.py
+++ b/dvc/repo/experiments/__init__.py
@@ -196,7 +196,7 @@ class Experiments:
                             branch_name = ExpRefInfo.from_ref(branch).name
                         else:
                             branch_name = f"{resume_rev[:7]}"
-                        if self.scm.is_dirty():
+                        if self.scm.is_dirty(untracked_files=False):
                             logger.info(
                                 "Modified checkpoint experiment based on "
                                 "'%s' will be created",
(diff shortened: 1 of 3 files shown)


dtrifiro:
A note about tests:

https://github.com/iterative/scmrepo/pull/68 removed the workaround we used to get `file://` urls working on windows, since support was added in `dulwich==0.20.38` (https://github.com/jelmer/dulwich/pull/965).
All tests using windows file urls were relying on this workaround and were using a broken file URL scheme on windows. To use a correctly formatted [RFC 8089](https://datatracker.ietf.org/doc/html/rfc8089) file URL scheme (on windows), it is sufficient to replace `\` with `/` on windows (or use `as_posix`).

pmrowla (approved):

dvc/repo/experiments/__init__.py
```diff
@@ -398,7 +398,7 @@ def reproduce_one(
             self.reset_checkpoints()
 
         if not (queue or tmp_dir or machine):
-            staged, _, _ = self.scm.status()
+            staged, _, _ = self.scm.status(untracked_files="no")
             if staged:
```
skshetry, line 402: While having some discussions related to using pygit2 for `status` for performance reasons with @dtrifiro, I noticed that this is equivalent to `git diff --staged --quiet`. So, this is equivalent to following:
```suggestion
            if git.pygit2.repo.diff(cached=True):
```

PR #7799, opened 2022-05-23
exp: implement `celery_queue.shutdown(kill=...)`

karajan1001:
1. Add some more unit test for `celery_queue.kill`
2. Implement `kill` argument for `celery_queue.shutdown`

### DIFF
--- a/dvc/repo/experiments/queue/base.py
+++ b/dvc/repo/experiments/queue/base.py
@@ -43,7 +43,7 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-@dataclass
+@dataclass(frozen=True)
 class QueueEntry:
     dvc_root: str
     scm_root: str
--- a/tests/unit/repo/experiments/queue/test_local.py
+++ b/tests/unit/repo/experiments/queue/test_local.py
@@ -1,4 +1,9 @@
+import time
+
+import pytest
 from celery import shared_task
+
+from dvc.repo.experiments.exceptions import UnresolvedExpNamesError
 
 
 def test_shutdown_no_tasks(test_queue, mocker):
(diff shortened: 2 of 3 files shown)


dvc/repo/experiments/queue/base.py
```diff
@@ -53,22 +53,22 @@ class QueueEntry:
...
 
+    def _key(self):
+        return (self.dvc_root, self.scm_root, self.stash_ref, self.stash_rev)
+
+    def __hash__(self):
+        return hash(self._key())
+
+    def __eq__(self, other):
+        return isinstance(other, QueueEntry) and self._key() == other._key()
+
```
pmrowla, line 71: This should be implemented using the automatically generated hash and comparison features in `dataclass`: https://docs.python.org/3/library/dataclasses.html

So for optional fields we would just set them like

```py
from dataclasses import dataclass, field

@dataclass
class QueueEntry:
    ...
    branch: Optional[str] = field(compare=False)
````

File: tests/unit/repo/experiments/queue/test_local.py [hunk 2]
```diff
@@ -1,5 +1,13 @@
+import time
+from collections import namedtuple
+
+import pytest
 from celery import shared_task
 
+from dvc.repo.experiments.exceptions import UnresolvedExpNamesError
+
+Entry = namedtuple("Entry", ["stash_rev"])
```
pmrowla, line 9: Instead of defining this test only type you can use the `Mock()` type and set the specific attributes you need. So in these tests you could do

```py
mock_queue_entry = mocker.Mock(stash_rev="...")
```

This makes it a bit more obvious that the object is a mocked/testing only object.

Alternatively, if you are going to define test only types it's better to name them with the `_` prefix (like `_Entry`) so that it's obvious it's an internal testing type specific to this module (and not something we are importing from DVC code)

pmrowla: Current PR LGTM otherwise

pmrowla (approved):

PR #7811, opened 2022-05-25
plots: grouping: stop using dpath.util.search

pared:

Seems like `dpath.util.search` is really slow. Tried modifying the code to stop using `search` and use `dpath.util.get` instead. There was a significant improvement (4x) but still seemed unreasonably long. Ended up implementing this method myself. 

We are still using `dpath.util.new` but its influence is orders of magnitude smaller than "reading" methods.

test repository:

```bash
#!/bin/bash

set -exu
pushd $TMPDIR

wsp=test_wspace
rep=test_repo

main=$(pwd)

mkdir $rep && pushd $rep

git init
dvc init

echo "numpy" >> requirements.txt
pip install -r requirements.txt

git add -A
git commit -am "init"


echo -e "p: 1" > params.yaml

echo -e "from random import random
import numpy as np
import sys

num_files = int(sys.argv[1])
linspace_size = int(sys.argv[2])

def load_params():
    import yaml
    with open('params.yaml') as fd:
            return yaml.safe_load(fd)

params = load_params()
p = params['p']

def calc_metric(index):
    metric = np.power(np.linspace(0,1,linspace_size),p)

    if index % 2 == 0:
       metric = 1 - metric

    metric += np.random.random(linspace_size)/(index+1)
    result=[]
    for elem in list(metric):
        result.append({f'val_{index}': elem})
    return result

import json
if __name__ == '__main__':
    for i in range(num_files):
        metric = calc_metric(i)
        with open(f'metric_{i}.json', 'w') as fd:
            json.dump(metric, fd)
" > train.py

echo -e "
from subprocess import run
outputs = []
for i in range(40):
    outputs.append('--plots')
    outputs.append(f'metric_{i}.json')

run(['dvc', 'run', '-d', 'train.py', '--params', 'p', '-n', 'train'] + outputs + ['python train.py 40 1000'])
" > call_dvc.py

git add -A
git commit -am "initial"

python call_dvc.py

git add -A
git commit -am "initial"

echo -e "p: 3" > params.yaml
dvc repro
```

command: `time dvc plots diff`

Result for main:
`14.88s user 0.22s system 99% cpu 15.182 total`

Result for this change:
`1.33s user 0.22s system 95% cpu 1.627 total`

### DIFF
--- a/dvc/render/match.py
+++ b/dvc/render/match.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, Dict, List, Optional
+from typing import TYPE_CHECKING, Dict, Optional
 
 import dpath.util
 
@@ -10,20 +10,18 @@ if TYPE_CHECKING:
     from dvc.types import StrPath
 
 
-def get_files(plots_data: Dict) -> List:
-    files = set()
-    for rev in plots_data.keys():
-        for file in plots_data[rev].get("data", {}).keys():
-            files.add(file)
-    sorted_files = sorted(files)
-    return sorted_files
-
-
 def group_by_filename(plots_data: Dict) -> Dict:
-    files = get_files(plots_data)
-    grouped = {
-        file: dpath.util.search(plots_data, ["*", "*", file]) for file in files
-    }
+    grouped: Dict[str, Dict] = {}
+
+    for revision in plots_data.keys():
+        data = plots_data[revision].get("data", {})
+        for file in data.keys():
+            content = data.get(file)
+            if content:
+                dpath.util.new(
+                    grouped, [file, revision, "data", file], content
+                )
+
     return grouped
 
 
--- a/tests/unit/test_plots.py
+++ b/tests/unit/test_plots.py
@@ -1,8 +1,6 @@
 import json
 import os
 
-from dvc.render.match import get_files
-
 
 def test_plots_order(tmp_dir, dvc):
     tmp_dir.gen(
@@ -28,9 +26,9 @@ def test_plots_order(tmp_dir, dvc):
             name="stage2",
         )
 
-    assert get_files(dvc.plots.show()) == [
+    assert list(dvc.plots.show()["workspace"]["data"]) == [
         "p.json",
+        os.path.join("sub", "p4.json"),
         "p1.json",
         os.path.join("sub", "p3.json"),
-        os.path.join("sub", "p4.json"),
     ]
--- a/tests/func/test_live.py
+++ b/tests/func/test_live.py
@@ -6,7 +6,6 @@ import pytest
 from funcy import first
 
 from dvc import stage as stage_module
-from dvc.render.match import get_files
 
 pytest.importorskip("dvclive", reason="no dvclive")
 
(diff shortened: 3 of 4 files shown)


pared:
I would love it if someone could confirm the results.
cc @mattseddon @dberenbaum @shcheklein

mattseddon:
> I would love it if someone could confirm the results. cc @mattseddon @dberenbaum @shcheklein

I can confirm. 

Last release: 

`dvc plots diff  14.96s user 0.38s system 94% cpu 16.241 total`

`main`:

`dvc plots diff  15.37s user 0.36s system 102% cpu 15.339 total`

This:

`dvc plots diff  2.06s user 0.46s system 107% cpu 2.353 total`

mattseddon:
image

Performance issues will be back with us again now:

https://user-images.githubusercontent.com/37993418/170389802-6edbdac2-739e-4f68-88de-64b639953638.mov

cc @sroy3 

Relates to https://github.com/iterative/vscode-dvc/issues/1689 & https://github.com/iterative/vscode-dvc/issues/1643

dberenbaum:
@karajan1001 @iterative/dvc Could someone review please? It's an important performance improvement for the VS Code release.

efiop:
@dberenbaum Mostly waiting for https://github.com/iterative/dvc-bench/pull/352 to be able to confirm and keep an eye on.

dvc/render/match.py
```diff
@@ -19,12 +20,34 @@ def get_files(plots_data: Dict) -> List:
...
+    revisions = list(plots_data)
+
+    grouped: Dict[str, Dict] = defaultdict(dict)
+
+    for revision in revisions:
+        for file in files:
+            path = [revision, "data", file]
+            content = _get(plots_data, path)
+            if content:
+                dpath.util.new(grouped[file], path, content)
```
efiop, line 49: Asuming the code above is correct, why not just
```suggestion
    for revision in revisions:
        for file in files:
            data = plots_data[revision].get("data", {})
            content = data.get(file)
            if content:
                dpath.util.new(grouped[file], path, content)
```
and reduce the complexity and not even need `_get`(not used anywhere else anyway)?

dvc/render/match.py
```diff
@@ -19,12 +20,34 @@ def get_files(plots_data: Dict) -> List:
...
+    return dict(grouped)
```
efiop, line 50: Assuming that the code above works, why not just

```suggestion
    grouped = {}

    for revision in plots_data.keys():
        data = plots_data[revision].get("data", {})
        for file in data.keys():
            content = data.get(file)
            if content:
                dpath.util.new(grouped, [file, revision, "data", file], content)

    return grouped
```

and not need `get_files`(redundant walk and sort. Btw, not used anywhere, why wasn't it `_get_files`? same question about other functions here),  not need `_get`, and not need revisions (were creating a list for no reason) and reduce complexity?

pared: I think it was used somewhere else. Dropping it now.

tests/unit/test_plots.py
```diff
@@ -28,9 +26,9 @@ def test_plots_order(tmp_dir, dvc):
             name="stage2",
         )
 
-    assert get_files(dvc.plots.show()) == [
```
pared, line 31: We should have not have used that here, as it was supposed to test the order of `show` results.

efiop:
Thank you!

PR #7823, opened 2022-05-27
api: open: Raise ValueError if rev is used in wrong mode. 

### LINKED ISSUE #7405: api.open: Write mode doesn't work with rev argument
# Bug Report

## Description

Error when using `dvc.api.open("foo", rev="HEAD", mode="w")`. See below for full error.

It's easy enough to work around this, but we should either fix it or not support write mode.

### Reproduce

In a fresh git+dvc repo, I added a `foo` file and committed:

```bash
$ echo foo > foo
$ git add foo
$ git commit -m "adds foo"
```

Then I tried to write a copy with `dvc.api.open()`:

```python
>>> with dvc.api.open("foo", mode="w") as f:
...     f.write("foo_copy")
...
8
>>> with dvc.api.open("foo", rev="HEAD", mode="w") as f:
...     f.write("foo_copy")
...
  File "<stdin>", line 1, in <module>
  File "/Users/dave/miniforge3/envs/example-get-started/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/Users/dave/Code/dvc/dvc/api.py", line 76, in _open
    with _repo.open_by_relpath(
  File "/Users/dave/miniforge3/envs/example-get-started/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/Users/dave/Code/dvc/dvc/repo/__init__.py", line 499, in open_by_relpath
    with fs.open(
  File "/Users/dave/Code/dvc/dvc/fs/repo.py", line 228, in open
    return fs.open(path, mode=mode, encoding=encoding)
  File "/Users/dave/Code/dvc/dvc/fs/fsspec_wrapper.py", line 80, in open
    return self.fs.open(path, mode=mode)
  File "/Users/dave/miniforge3/envs/example-get-started/lib/python3.8/site-packages/fsspec/spec.py", line 997, in open
    self.open(
  File "/Users/dave/miniforge3/envs/example-get-started/lib/python3.8/site-packages/fsspec/spec.py", line 1009, in open
    f = self._open(
  File "/Users/dave/miniforge3/envs/example-get-started/lib/python3.8/site-packages/scmrepo/fs.py", line 84, in _open
    raise NotImplementedError
NotImplementedError
```

### Expected

Being able to write a file from a specified revision.

### Environment information

**Output of `dvc doctor`:**

```console
$ dvc doctor
DVC version: 2.9.4.dev58+g47481219
Platform: Python 3.8.5 on macOS-10.16-x86_64-i386-64bit
Supports:
        gdrive (pydrive2 = 1.7.3),
        hdfs (fsspec = 2022.1.0, pyarrow = 3.0.0),
        webhdfs (fsspec = 2022.1.0),
        http (aiohttp = 3.7.3, aiohttp-retry = 2.4.6),
        https (aiohttp = 3.7.3, aiohttp-retry = 2.4.6),
        ssh (sshfs = 2021.11.2)
Caches: local
Remotes: local
Repo: dvc, git
```

daavoo:
Closes #7405
 

dvc/api.py
```diff
@@ -33,21 +34,154 @@ def get_url(path, repo=None, rev=None, remote=None):
...
+    rev: Optional[str] = None,
+    remote: Optional[str] = None,
+    mode: str = "r",
+    encoding: Optional[str] = None,
 ):
     """
-    Open file in the supplied path tracked in a repo (both DVC projects and
-    plain Git repos are supported). For Git repos, HEAD is used unless a rev
-    argument is supplied. The default remote is tried unless a remote argument
-    is supplied. It may only be used as a context manager:
```
dberenbaum, line 43: Is the PR title accurate here or even the primary purpose of the PR  😄? I like what you are doing, but I think the existing first paragraph is a nice summary. Not sure if we need to amend the docs, but feels like we are losing some useful info at the top (supporting DVC or Git repos, use HEAD and default remote by default).

dvc/api.py
```diff
@@ -33,21 +34,154 @@ def get_url(path, repo=None, rev=None, remote=None):
...
+        https://docs.python.org/3/library/functions.html#open
+
+    .. _Codec:
+        https://docs.python.org/3/library/codecs.html#standard-encodings
+
+    .. _file object:
+        https://docs.python.org/3/glossary.html#term-file-object
+
     """
+    if rev is not None and "r" not in mode:
```
pmrowla, line 182:
```suggestion
    if "r" not in mode:
```

I don't think we should support write-mode at all, regardless of whether or not `rev` is in use. Even in the workspace-only scenario, write behavior for `api.open` is not well defined.

I think this is really a use-case that dvcx is meant to solve - where there is a clear definition of what a write/modify operation is supposed to do (i.e. write + dvc-add + commit and so on)

pmrowla (changes requested):

dvc/api.py
pmrowla: Basically, the `api.open` context is really only intended to provide the same behavior as `api.read`, but in situations where it's more useful to have the file-like object w/`read()` attribute, rather than directly calling `api.read()` itself.

dvc/api.py
pmrowla: I don't have a problem with keeping the docstring changes in this PR as long as the general doc changes are kept in a separate commit from the mode=write behavior related changes

daavoo: @dberenbaum Do you think that the info about argument defaults should be present in the "short description" section of the docstring?

pmrowla (approved):
code changes LGTM, this can be merged as soon as the docstring discussions are settled

skshetry:
@daavoo, please split the PR, the docs change are unrelated to the bugfix.

dvc/api.py
daavoo: > but what about (both DVC projects and plain Git repos are supported)?

This should be documented in the `repo` argument. 

I extracted the content from (https://dvc.org/doc/api-reference/open):

```
        repo (str, optional): location of the DVC project.
            Defaults to the current project (found by walking up from the
            current working directory tree).
            It can be a URL or a file system path.
            Both HTTP and SSH protocols are supported for online Git repos
            (e.g. [user@]server:project.git).
```

Do you think something is missing?

dberenbaum: I don't think it's that clear that it can be used for either DVC repos or regular Git repos, but I'm not sure anyone is using it for that anyway and definitely don't want to keep holding up getting this merged, so let's merge it.

dberenbaum (approved):

dvc/api.py
daavoo: > I don't think it's that clear that it can be used for either DVC repos or regular Git repos

Updated the description

PR #7827, opened 2022-05-30
testing: use shutil.copytree with dirs_exist_ok

skshetry:
See https://docs.python.org/3/library/shutil.html#shutil.copytree. There is `dirs_exist_ok` kwarg in Python >3.8.


efiop:
Thank you!

efiop:
Turns out, when I tried it myself I totally blinded the `for` loop 🤦

PR #7828, opened 2022-05-31
exp: Replace status Running with Pending if no worker runs.

karajan1001:
Currently our `queue status` shows Running even if no worker running at
    present, replace it with "Pending"

    1. Replace "running" with "Pending" if no active worker
    2. Add unit test for `queue status`

### DIFF
--- a/dvc/repo/experiments/queue/base.py
+++ b/dvc/repo/experiments/queue/base.py
@@ -160,7 +160,6 @@ class BaseStashQueue(ABC):
             return datetime.fromtimestamp(commit.commit_time)
 
         for queue_entry in self.iter_active():
-            print(queue_entry, queue_entry.name)
             result.append(
                 {
                     "rev": queue_entry.stash_rev,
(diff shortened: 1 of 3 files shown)


dvc/repo/experiments/queue/base.py
```diff
@@ -149,17 +149,18 @@ def clear(self) -> List[str]:
...
 
-        result: List[Mapping[str, Optional[str]]] = []
+        result: List[Dict[str, Optional[str]]] = []
 
         def _get_timestamp(rev):
             commit = self.scm.resolve_commit(rev)
             return datetime.fromtimestamp(commit.commit_time)
 
         for queue_entry in self.iter_active():
+            print(queue_entry, queue_entry.name)
```
pmrowla, line 163:
```suggestion
```

pmrowla:
Right, I think the problem is that after the chain changes we are using the wrong task ID in our shutdown handler. We need to be checking the cleanup task IDs.

https://github.com/iterative/dvc/blob/db520d7fc62f027e8a49a07af46bbd20ee8b8cb0/dvc/repo/experiments/queue/local.py#L300-L303

It may just be easier for us to get the level of atomicity we need by generating a single task instead of separating it into chained subtasks. We can discuss this during this weeks 1 on 1.

pmrowla:
closing in favor of https://github.com/iterative/dvc/pull/7835

PR #7835, opened 2022-06-01
exp: use single atomic task for celery runs

pmrowla:

### DIFF
--- a/dvc/repo/experiments/queue/local.py
+++ b/dvc/repo/experiments/queue/local.py
@@ -245,25 +245,14 @@ class LocalCeleryQueue(BaseStashQueue):
         for queue_entry in to_kill:
             self.proc.kill(queue_entry.stash_rev)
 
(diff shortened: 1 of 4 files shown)


dvc/repo/experiments/queue/local.py
```diff
@@ -287,25 +245,14 @@ def kill(self, revs: Collection[str]) -> None:
...
-                for _, task_entry in tasks:
+        self.celery.control.shutdown()
+        if kill:
+            for _, task_entry in self._iter_active_tasks():
+                try:
                     self.proc.kill(task_entry.stash_rev)
-        else:
-            self.celery.control.shutdown()
+                except ProcessLookupError:
+                    continue
```
pmrowla, line 255: Since exp task is atomic, we can rely on the default celery `shutdown()` behavior to shut down worker after the currently running task has completed.

Issue #7840, opened 2022-06-03
queue remove: add remove support for short revs


karajan1001:
Now cmd `queue remove` can not accept short rev as a name.

```bash
$ dvc queue status
Rev                                       Name    Created    Status
1f3680589b996903a040687c168000093157816b  I40     11:19 AM   Queued
386a8290005cc33093bb8a698b369a980c5f2e32  I30     11:19 AM   Queued
$ dvc queue remove 1f36
Removed experiments in queue:
```

pmrowla:
closed in #7886

PR #7841, opened 2022-06-03
queue remove: add remove support for short revs(#7840)

karajan1001:
fix: #7840

Now queue remove can not accept short rev as name.
1. We add support for short rev as name in `queue remove`.
2. Better output if no queue task found for name.
3. short rev on kill can only be work for rev with lens longer than 4

### DIFF
--- a/dvc/commands/queue/status.py
+++ b/dvc/commands/queue/status.py
@@ -23,7 +23,7 @@ class CmdQueueStatus(CmdBase):
         for exp in result:
             created = format_time(exp.get("timestamp"))
             td.append(
-                [exp["rev"], exp.get("name", ""), created, exp["status"]]
+                [exp["rev"][:7], exp.get("name", ""), created, exp["status"]]
             )
         td.render()
 
--- a/tests/unit/command/test_queue.py
+++ b/tests/unit/command/test_queue.py
@@ -99,7 +99,9 @@ def test_experiments_stop(dvc, scm, mocker):
     m.assert_called_once_with(kill=True)
 
 
-def test_experiments_status(dvc, scm, mocker):
+def test_experiments_status(dvc, scm, mocker, capsys, caplog):
+    from datetime import datetime
+
     cli_args = parse_args(
         [
             "queue",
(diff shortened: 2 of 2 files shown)


dvc/repo/experiments/queue/base.py
```diff
@@ -569,11 +574,12 @@ def match_queue_entry_by_name(
...
-                if rev.startswith(exp_name):
-                    result[exp_name] = entry
-                    break
-            else:
+            if len(exp_name) >= 4:
+                for rev, entry in entry_rev_list:
+                    if rev.startswith(exp_name):
+                        result[exp_name] = entry
+                        break
+            if exp_name not in result:
```
pmrowla, line 582: This string length comparison isn't really safe for us to do. We should really be using `resolve_rev` here instead of using the min length + `startswith` check.

dvc/repo/experiments/queue/base.py
```diff
@@ -129,12 +129,17 @@ def remove(self, revs: Collection[str]) -> List[str]:
...
         for stash_rev, stash_entry in self.stash.stash_revs.items():
-            if stash_rev in revs:
-                to_remove[stash_rev] = stash_entry
-                removed.append(stash_rev)
-            elif stash_entry.name in revs:
-                to_remove[stash_rev] = stash_entry
-                removed.append(stash_entry.name)
+            for rev in revs:
+                if len(rev) < 4:
+                    continue
```
pmrowla, line 134: I think we should be using `match_queue_entry_by_name` here. Once we have the matching `QueueEntry`'s, we can easily get the corresponding `StashEntry`'s to pass into `_remove_revs` with something like:
```py
queue_entries = match_queue_entry_by_name(revs, ...)
to_remove = {
    entry.stash_rev: self.stash.stash_revs[entry.stash_rev]
    for entry in queue_entries.values()
}
self._remove_revs(to_remove)
```

pmrowla (changes requested):

dvc/repo/experiments/queue/base.py
pmrowla: One other alternative here would be use scmrepo's `Git.is_sha` (and keep the `startswith` check):

```suggestion
            if self.scm.is_sha(exp_name):
                for rev, entry in entry_rev_list:
                    if rev.startswith(exp_name):
                        result[exp_name] = entry
                        break
            if exp_name not in result:
```

pmrowla:
@karajan1001 latest changes look good, I'd still like to get rid of the string length check though (see latest review comments)

PR #7849, opened 2022-06-06
setup: set upper bound on networkx

efiop:
#7848


efiop:
Ok, let's limit it for now to unblock development. Still need to see how to live with it long-term

PR #7859, opened 2022-06-07
readme: mention vs code extension

dberenbaum:
Adds a couple of brief references to the VS Code extension.

Questions/issues to consider:
* Where (else) in the readme should we mention the extension?
* Should we be giving any kind of summary or additional info about the extension in the DVC readme?
* The quickstart could be moved above "How DVC works", which would move up the reference to the extension. We should rewrite/reorganize the rest of the readme, but I didn't want to introduce unrelated changes in this PR.
* Studio might also be mentioned more and earlier (for example, when discussing sharing and comparing experiments), but this can be a separate discussion.


README.rst
```diff
@@ -35,8 +35,9 @@ learning projects:
 #. **Iterate** fast with straightforward YAML pipelines. When you make changes,
    only run the steps impacted by those changes.
 
-#. **Track** experiments in your local Git repo (no servers needed) and choose
-   which to keep and share.
+#. **Track** experiments in your local Git repo (no servers needed) and
+   [IDE](https://marketplace.visualstudio.com/items?itemName=Iterative.dvc).
```
shcheklein, line 39: IDE -> VS Code

README.rst
```diff
@@ -95,6 +96,9 @@ Please read `Get Started <https://dvc.org/doc/get-started>`_ guide for a full ve
 |                                   | | ``$ dvc push``                                                           |
 +-----------------------------------+----------------------------------------------------------------------------+
 
+Prefer to work in your IDE? Check out the [DVC Extension for Visual Studio
```
shcheklein, line 99: doesn't look very natural tbh

shcheklein (changes requested):
A few ideas:
- add badge (instead of donate - we don't need it and it's not very useful anyway)
- add link (section before badges)
- add a link in the line 26
- even more extreme add "in CLI (or VS code)" to the line 17
- mention in the installation section

README.rst
shcheklein: sorry, I meant that the whole paragraph feels a bit too much here - looks like an ad a bit. If we have enough organic mentions, I would say we can completely skip it, wdyt?

README.rst
```diff
@@ -7,8 +7,9 @@
 • `Chat (Community & Support) <https://dvc.org/chat>`_
 • `Tutorial <https://dvc.org/doc/get-started>`_
 • `Mailing List <https://sweedom.us10.list-manage.com/subscribe/post?u=a08bf93caae4063c4e6a351f6&id=24c0ecc49a>`_
+• `VS Code Extension <https://marketplace.visualstudio.com/items?itemName=Iterative.dvc>`_
 
-|CI| |Maintainability| |Coverage| |Donate| |DOI|
+|CI| |Maintainability| |Coverage| |Donate| |VSCode| |DOI|
```
shcheklein, line 12: VS Code

shcheklein:
@dberenbaum let's merge it Monday end of the day? (there is a merge conflict, and could you please resolve the open conversations - just to make sure that you saw the feedback :) ).

README.rst
dberenbaum: This doesn't accept whitespace. It's not visible anyway on the page, so I don't think it's an issue. Let me know if you think it's a problem.

dberenbaum:
Well I forgot this is rst and not md 🤦 . Please hold.

README.rst
```diff
@@ -7,8 +7,9 @@
 • `Chat (Community & Support) <https://dvc.org/chat>`_
 • `Tutorial <https://dvc.org/doc/get-started>`_
 • `Mailing List <https://sweedom.us10.list-manage.com/subscribe/post?u=a08bf93caae4063c4e6a351f6&id=24c0ecc49a>`_
+• `VS Code Extension <https://marketplace.visualstudio.com/items?itemName=Iterative.dvc>`_
 
-|CI| |Maintainability| |Coverage| |DOI|
+|CI| |Maintainability| |Coverage| |VSCode| |DOI|
```
shcheklein, line 12: VS Code

shcheklein: is this text visible?

shcheklein: okay, probably not :)

dberenbaum: @shcheklein It's not visible and won't work with whitespace. See https://github.com/iterative/dvc/pull/7859#discussion_r893955863

dberenbaum: Source from https://docutils.sourceforge.io/docs/user/rst/quickref.html#citations:

> Citation labels contain alphanumerics, underlines, hyphens and fullstop

dberenbaum: Never mind, it does work per https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#substitution-definitions, so I must have made some mistake when trying it before. Updated now.

dberenbaum:
@shcheklein tldr changed to `VS Code` (I mistakenly thought rst couldn't handle whitespace there).

shcheklein (approved):

PR #7862, opened 2022-06-08
readme: remove donate requests

dberenbaum:
See https://github.com/iterative/dvc/pull/7859#pullrequestreview-998911416. Donate badges/buttons seem unnecessary and might not be well received since DVC is developed and maintained by a for-profit org.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,3 +1,0 @@
-# These are supported funding model platforms
-
-patreon: DVCorg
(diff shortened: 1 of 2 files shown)


skshetry:
@dberenbaum @shcheklein, should we also remove [`.github/FUNDING.yml`](https://github.com/iterative/dvc/blob/c9a3cb803ca8f5b0fd70a02fb6ed415035bc24c5/.github/FUNDING.yml)?

https://github.com/iterative/dvc/blob/c9a3cb803ca8f5b0fd70a02fb6ed415035bc24c5/.github/FUNDING.yml

PR #7873, opened 2022-06-10
run-cache: fix push from Windows to remote filesystems

skshetry:
We were passing paths with '\\' in the filename instead of posix paths to
remote filesystems which were causing failures to upload.
Although in some filesystems, this was failing silently, so upload would seem
to have passed, but the download later would fail.

### DIFF
--- a/dvc/stage/cache.py
+++ b/dvc/stage/cache.py
@@ -213,6 +213,7 @@ class StageCache:
         cached_stage.checkout()
 
     def transfer(self, from_odb, to_odb):
+        from dvc.fs import LocalFileSystem
         from dvc.fs.callbacks import Callback
 
         from_fs = from_odb.fs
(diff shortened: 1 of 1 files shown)


dvc/stage/cache.py
```diff
@@ -226,6 +227,9 @@ def transfer(self, from_odb, to_odb):
 
         for src in from_fs.find(runs):
             rel = from_fs.path.relpath(src, from_odb.fs_path)
+            if not isinstance(to_fs, LocalFileSystem):
+                rel = from_fs.path.as_posix(rel)
```
skshetry, line 231: @efiop, this shows that we really need to use posix paths everywhere.

efiop: Hopefully this should be mitigated with run-cache object introduction.

efiop (approved):
Nice catch!

PR #7895, opened 2022-06-14
deps: bump dvc-data to 0.0.6; fix imports

skshetry:
See inline comments.

### DIFF
--- a/dvc/data_cloud.py
+++ b/dvc/data_cloud.py
@@ -6,9 +6,9 @@ from typing import TYPE_CHECKING, Iterable, Optional
 from dvc_data.db import get_index
 
 if TYPE_CHECKING:
+    from dvc_data.hashfile.db import HashFileDB
+    from dvc_data.hashfile.hash_info import HashInfo
     from dvc_data.status import CompareStatusResult
-    from dvc_objects.db import ObjectDB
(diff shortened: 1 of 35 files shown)


dvc/data_cloud.py
```diff
@@ -6,9 +6,9 @@
 from dvc_data.db import get_index
 
 if TYPE_CHECKING:
+    from dvc_data.hashfile.db import HashFileDB
```
skshetry, line 9: Changed `ObjectDB` type to `HashFileDB`. I didn't change it elsewhere as it seems we'll have few more iterations on this, and it's better to use stricter version (i.e. `ObjectDB` where possible).

dvc/repo/__init__.py
```diff
@@ -411,16 +410,7 @@ def used_objs(
...
 
-        def _add_suffix(objs: Set["HashFile"], suffix: str) -> None:
-            from itertools import chain
-
-            from dvc_data import iterobjs
-
-            for obj in chain.from_iterable(map(iterobjs, objs)):
-                if obj.name is not None:
-                    obj.name += suffix
-
```
skshetry, line 422: Gets rid of this functionality to set name. :)

efiop (approved):

PR #7907, opened 2022-06-16
parsing: Support dict unpacking in `cmd`.

### LINKED ISSUE #6107: More flexible dvc.yaml parameterisation
## Context

One of DVC usage scenarios we should consider is the following:
1. User has a big and complex real-life repo and workflows around it (one example is CI/CD)
1. User wants to apply DVC without modifications to his current scripts, because that will trigger code changes in subsequent workflows, which the user don’t want to do right now, but rather later, when DVC will be integrated and will prove it is useful for the user.

One example for this is the [YOLOv5 repo](https://github.com/ultralytics/yolov5). The github repo has only [testing in CI](https://github.com/ultralytics/yolov5/blob/develop/.github/workflows), but for sure we can have a private fork and multiple CI/CD workflows which train, test, export and deploy models.

## Problem

In this case user could have scripts [which run like this](https://github.com/ultralytics/yolov5/blob/develop/.github/workflows/ci-testing.yml#L67):
```
python train.py --img 128 --batch 16 --weights weights/${{ matrix.model }}.pt --cfg models/${{ matrix.model }}.yaml --epochs 1 --device $di
```

The scripts could have a lot of args (this `train.py` has about 20 of them, all handled by Python argparse module). If I don’t want to modify my scripts yet, I’d need to duplicate all arguments in two places: in params.yaml and in dvc.yaml, thus having three slightly different copies of the arguments.

Indeed, this is a bad practice that creates a technical debt of maintaining three copies of the same and shouldn’t be used for too long, but as a user I would prefer to migrate to DVC step by step, not breaking anything in current workflows in the process. And if I’m not sure I want to migrate and just want to give DVC a try I would want to keep all my current scripts unchanged to remove DVC later if I won’t be planning on using it.

## Suggestion
Expand args dictionary and lists from params in dvc.yaml to allow user to avoid duplicating every param in dvc.yaml:
```
# params.yaml
train-params:
  param1: 1
  param2: 2
  param3: [value1, value2]


# dvc.yaml
stages:
  train:
    cmd: python train.py ${train-params}
```

which gets translated to
```
python train.py --param1 1 --param2 2 --param3 value1 value2
```

## Few questions we should also consider:

<details>
<summary><b>1. Is there a use case when a user wants to expand a nested dict?</b></summary>

Example:
```
# params.yaml
train-params:
  param1:
    nested-param: value

# dvc.yaml
stages:
  train:
    cmd: python train.py ${train-params}
```
</details>

<details>
<summary><b>2. There are optional args like `--resume` which could be supplied or not. We could support this like following:</b></summary>

```
# dvc.yaml
stages:
  train:
    deps:
      - train.py
    cmd: python train.py --resume ${resume}
```

In case we want to run just “python train.py”
```
$ dvc exp run -S resume= 
> python train.py

# params.yaml
resume:
```
In case we want to run “python train.py --resume path-to-weights.pt”:
```
$ dvc exp run -S resume=path-to-weights.pt
> python train.py --resume path-to-weights.pt

# params.yaml
resume: path-to-weights.pt
```
Right now the first use case is not supported at all:
```
$ dvc exp run -S param=  
ERROR: Must provide a value for parameter 'param'
```
Though params.yaml could be like
```
# params.yaml
param:
```
And experiment could be run, but “None” value will be used
```
$ dvc exp run
> python train.py None
```
That said, the last modification could break backwards compatibility if anyone uses the later approach with None substitution, but it looks more like a bug, then a solution, and we can replace this behaviour in the future.
</details>

<details>
<summary><b>3. Are there other examples of CLI parameters that scripts can take, but dvc.yaml / `dvc exp run` doesn't support now?</b></summary>

Would be great to gather them is well.
</details>

My perspective here is limited to python and bash scripts, so I can miss something important. I'm going to gather more examples and possible issues and add more information about them. Meanwhile, it would be great to hear what others think.

daavoo:
Allow to use dictionaries as values for template interpolation but only inside the `cmd` key.

Given the following params.yaml and dvc.yaml:

```yaml
# params.yaml
dict:
  foo: foo
  bar: bar
```

```yaml
# dvc.yaml
stages:
  stage1:
    cmd: python script.py ${dict}
```

The dictionary will be unpacked with the following syntax:

```yaml
# dvc.yaml
stages:
  stage1:
    cmd: python script.py --foo foo --bar bar
```

Closes #6107

### DIFF
--- a/dvc/parsing/interpolate.py
+++ b/dvc/parsing/interpolate.py
@@ -1,11 +1,12 @@
 import re
 import typing
-from collections.abc import Mapping
+from collections.abc import Iterable, Mapping
 from functools import singledispatch
 
 from funcy import memoize, rpartial
 
 from dvc.exceptions import DvcException
+from dvc.utils.flatten import flatten
 
 if typing.TYPE_CHECKING:
     from typing import List, Match
@@ -80,6 +81,45 @@ def _(obj: bool):
     return "true" if obj else "false"
 
 
+@to_str.register(dict)
+def _(obj: dict):
+    from dvc.config import Config
+
+    config = Config().get("parsing", {})
+
+    result = ""
+    for k, v in flatten(obj).items():
+
+        if isinstance(v, bool):
+            if v:
+                result += f"--{k} "
+            else:
+                if config.get("bool", "store_true") == "boolean_optional":
+                    result += f"--no-{k} "
+
+        elif isinstance(v, str):
+            result += f"--{k} '{v}' "
+
+        elif isinstance(v, Iterable):
+            for n, i in enumerate(v):
+                if isinstance(i, str):
+                    i = f"'{i}'"
+                elif isinstance(i, Iterable):
+                    raise ParseError(
+                        f"Cannot interpolate nested iterable in '{k}'"
+                    )
+
+                if config.get("list", "nargs") == "append":
+                    result += f"--{k} {i} "
+                else:
+                    result += f"{i} " if n > 0 else f"--{k} {i} "
+
+        else:
+            result += f"--{k} {v} "
+
+    return result.rstrip()
+
+
 def _format_exc_msg(exc: "ParseException"):
     from pyparsing import ParseException
 
@@ -148,23 +188,33 @@ def get_expression(match: "Match", skip_checks: bool = False):
     return inner if skip_checks else parse_expr(inner)
 
 
+def validate_value(value, key):
+    from .context import PRIMITIVES
+
+    not_primitive = value is not None and not isinstance(value, PRIMITIVES)
+    not_foreach = key is not None and "foreach" not in key
+    if not_primitive and not_foreach:
+        if isinstance(value, dict):
+            if key == "cmd":
+                return True
+        raise ParseError(
+            f"Cannot interpolate data of type '{type(value).__name__}'"
+        )
+
+
 def str_interpolate(
     template: str,
     matches: "List[Match]",
     context: "Context",
     skip_checks: bool = False,
+    key=None,
 ):
-    from .context import PRIMITIVES
-
     index, buf = 0, ""
     for match in matches:
         start, end = match.span(0)
         expr = get_expression(match, skip_checks=skip_checks)
         value = context.select(expr, unwrap=True)
-        if value is not None and not isinstance(value, PRIMITIVES):
-            raise ParseError(
-                f"Cannot interpolate data of type '{type(value).__name__}'"
-            )
+        validate_value(value, key)
         buf += template[index:start] + to_str(value)
         index = end
     buf += template[index:]
--- a/tests/func/parsing/test_interpolated_entry.py
+++ b/tests/func/parsing/test_interpolated_entry.py
@@ -259,3 +259,59 @@ def test_vars_load_partial(tmp_dir, dvc, local, vars_):
         d["vars"] = vars_
     resolver = DataResolver(dvc, tmp_dir.fs_path, d)
     resolver.resolve()
+
+
+@pytest.mark.parametrize(
+    "bool_config, list_config",
+    [(None, None), ("store_true", "nargs"), ("boolean_optional", "append")],
+)
+def test_cmd_dict(tmp_dir, dvc, bool_config, list_config):
+    with dvc.config.edit() as conf:
+        if bool_config:
+            conf["parsing"]["bool"] = bool_config
+        if list_config:
+            conf["parsing"]["list"] = list_config
+
+    data = {
+        "dict": {
+            "foo": "foo",
+            "bar": 2,
+            "string": "spaced string",
+            "bool": True,
+            "bool-false": False,
+            "list": [1, 2, "foo"],
+            "nested": {"foo": "foo"},
+        }
+    }
+    (tmp_dir / DEFAULT_PARAMS_FILE).dump(data)
+    resolver = DataResolver(
+        dvc,
+        tmp_dir.fs_path,
+        {"stages": {"stage1": {"cmd": "python script.py ${dict}"}}},
+    )
+
+    if bool_config is None or bool_config == "store_true":
+        bool_resolved = " --bool"
+    else:
+        bool_resolved = " --bool --no-bool-false"
+
+    if list_config is None or list_config == "nargs":
+        list_resolved = " --list 1 2 'foo'"
+    else:
+        list_resolved = " --list 1 --list 2 --list 'foo'"
+
+    assert_stage_equal(
+        resolver.resolve(),
+        {
+            "stages": {
+                "stage1": {
+                    "cmd": "python script.py"
+                    " --foo 'foo' --bar 2"
+                    " --string 'spaced string'"
+                    f"{bool_resolved}"
+                    f"{list_resolved}"
+                    " --nested.foo 'foo'"
+                }
+            }
+        },
+    )
--- a/dvc/config_schema.py
+++ b/dvc/config_schema.py
@@ -280,4 +280,8 @@ SCHEMA = {
         "plots": str,
         "live": str,
     },
+    "parsing": {
+        "bool": All(Lower, Choices("store_true", "boolean_optional")),
+        "list": All(Lower, Choices("nargs", "append")),
+    },
 }
--- a/dvc/parsing/__init__.py
+++ b/dvc/parsing/__init__.py
@@ -303,7 +303,7 @@ class EntryDefinition:
     ) -> DictStr:
         try:
             return context.resolve(
-                value, skip_interpolation_checks=skip_checks
+                value, skip_interpolation_checks=skip_checks, key=key
             )
         except (ParseError, KeyNotInContext) as exc:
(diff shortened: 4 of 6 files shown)


skshetry:
> Allow to use dictionaries as values for template interpolation but only inside the `cmd` key.
> 
> Given the following params.yaml and dvc.yaml:
> 
> ```yaml
> # params.yaml
> dict:
>   foo: foo
>   bar: bar
> ```
> 
>  ```yaml
> # dvc.yaml
> stages:
>   stage1:
>     cmd: python script.py ${dict}
> ```
> 
> The dictionary will be unpacked with the following syntax:
> 
> ```yaml
> # dvc.yaml
> stages:
>   stage1:
>     cmd: python script.py --foo foo --bar bar
> ```
> 
> Closes #6107
> 
> * [x]  ❗ I have followed the [Contributing to DVC](https://dvc.org/doc/user-guide/contributing/core) checklist.
> * [ ]  📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here.

Any recent motivations for this? 🙂

daavoo:
> Any recent motivations for this? 🙂

When exploring how to better integrate `dvc params` with ML frameworks, this showed as low-hanging fruit for some frameworks like [Yolov5](https://github.com/iterative/yolov5) (linked issue), [HuggingFace](https://huggingface.co/docs/transformers/v4.20.0/en/internal/trainer_utils#transformers.HfArgumentParser) , [PyTorch Lightning](https://pytorch-lightning.readthedocs.io/en/stable/common/lightning_cli.html#use-of-command-line-arguments), etc. 
In addition, the default syntax is good enough for custom training scripts using simple `argparse` and having more flexibility of configuring `cmd` calls of CLI tools without having to edit `dvc.yaml` for each iteration (adding/removing argument)

tests/func/parsing/test_interpolated_entry.py
```diff
@@ -259,3 +261,13 @@ def test_vars_load_partial(tmp_dir, dvc, local, vars_):
         d["vars"] = vars_
     resolver = DataResolver(dvc, tmp_dir.fs_path, d)
     resolver.resolve()
+
+
+def test_cmd_dict(tmp_dir, dvc):
+    (tmp_dir / DEFAULT_PARAMS_FILE).dump(CONTEXT_DATA)
```
dberenbaum, line 267: What about non-default params files? Doesn't seem to work with those right now.

dberenbaum:
Some data types seem like they get tricky 😓 :

* Strings: We probably need to quote these? Otherwise, they can easily turn into a mess like "--model random forest"
* Booleans: I think these should be expanded to just `--foo` with no value if true and ignored if false. Or they could be set to `--no-foo` if false, but this seems like too much magic to me. 
* Iterables: They don't work right now. I'd suggest `--foo 1 --foo 2` to be explicit and consistent with argparse.

tests/func/parsing/test_interpolated_entry.py
daavoo: > What about non-default params files? Doesn't seem to work with those right now.

Not sure I follow, the changes in this P.R. should work the same way as the current templating behavior

dberenbaum: My misunderstanding

tests/func/parsing/test_interpolated_entry.py
```diff
@@ -259,3 +259,38 @@ def test_vars_load_partial(tmp_dir, dvc, local, vars_):
         d["vars"] = vars_
     resolver = DataResolver(dvc, tmp_dir.fs_path, d)
     resolver.resolve()
+
+
+def test_cmd_dict(tmp_dir, dvc):
+    data = {
```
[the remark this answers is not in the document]
dberenbaum: > I ignored it for now (nothing gets added to the interpolated string) .
> Maybe `--no-{x}` makes sense (i.e. is also the [default used in click](https://click.palletsprojects.com/en/8.1.x/options/#boolean-flags)) , the worst thing that can happen is that we will add a flag that it's ignored?

No strong preference here.

> Not sure if we should raise an explicit exception if we detect a nested list or just render as if it was a string, which will probably raise an exception on the user code when parsing the args.

If easy enough, I would prefer the former, but doesn't need to block the PR.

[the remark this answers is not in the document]
daavoo: > No strong preference here.

Added the `--no-` prefix for negative booleans instead of ignoring them.

> If easy enough, I would prefer the former, but doesn't need to block the PR.

Added explicit exception

dvc/parsing/interpolate.py
```diff
@@ -80,6 +81,31 @@ def _(obj: bool):
     return "true" if obj else "false"
 
 
+@to_str.register(dict)
+def _(obj: dict):
+    result = ""
+    for k, v in flatten(obj).items():
+        if isinstance(v, bool):
```
karajan1001, line 88: This only works for `argparse.BooleanOptionalAction` not for `store_true/false`. Maybe need some better way to handle this?

dvc/parsing/interpolate.py
```diff
@@ -80,6 +81,31 @@ def _(obj: bool):
...
+            result += f"--{k} '{v}' "
+        elif isinstance(v, Iterable):
+            for i in v:
+                if isinstance(i, str):
+                    i = f"'{i}'"
+                elif isinstance(i, Iterable):
+                    raise ParseError(
+                        f"Cannot interpolate nested iterable in '{k}'"
+                    )
+                result += f"--{k} {i} "
```
karajan1001, line 103:
```python
> import argparse
> parser = argparse.ArgumentParser()
> parser.add_argument('--foo', nargs='*')
> parser.parse_args('--foo x y'.split())
Namespace(foo=['x', 'y'])

> parser.parse_args('--foo x --foo y'.split())
Namespace(foo=['y'])
```

Looks like setting arguments for multi times will overwrite the previous one.

dvc/parsing/interpolate.py
daavoo: The same issue about different options, as commented below https://github.com/iterative/dvc/pull/7907#discussion_r903429914 

Which option do you think we should consider as the most appropriate default? 

Note that this can be used for things beyond `argparse`, for example you can use the interpolation to pass flags to arbitrary executables that you call inside `cmd`

dvc/parsing/interpolate.py
dberenbaum: Let's pick one behavior for now. Is it possible to see which one is more widely used in existing ML CLIs? I would guess the `nargs` `--foo 1 2` behavior is more common, even though I find the explicit `append` behavior easier to read.

daavoo: Added `config.parsing.list` with `nargs` and `append` options. It doesn't look that bad to have the options.

dvc/parsing/interpolate.py
karajan1001: I also have no idea about it.

dberenbaum: I think that's more than enough to support for now.

tests/func/parsing/test_interpolated_entry.py
```diff
@@ -259,3 +259,59 @@ def test_vars_load_partial(tmp_dir, dvc, local, vars_):
         d["vars"] = vars_
     resolver = DataResolver(dvc, tmp_dir.fs_path, d)
     resolver.resolve()
+
+
+@pytest.mark.parametrize(
+    "bool_config", [None, "store_true", "boolean_optional"]
```
karajan1001, line 265: 1. Maybe we only need to test two different conditions? The `None` can be tested through some other method (Test if the default value is set corretly)?
2. Because these two configurations is independent, maybe we can test them in an (I'm not sure how to describe it) instead of using a product.  ["a1b1", "a2b2"] instead of ["a1", "a2"] x ["b1", "b2"]. 
The two suggestion above can reduce the tests amount from 9 to 2.

karajan1001:
LGTM for the most part, another concern is that the two new configurations still can't solve all of the problems as two different types of` arg` might exist in one command.

tests/func/parsing/test_interpolated_entry.py
daavoo: I updated it to 3 test cases (`["a1b1", "a2b2"]`). I still included the default in the same test, not sure if it's worth it to separate it as the body of the test would be the same

daavoo:
> another concern is that the two new configurations still can't solve all of the problems as two different types of arg might exist in one command.

Indeed. There are many unsolved issues but the idea for this initial P.R. is to cover basic scenarios to don't overcomplicate the logic until users request for specific issues

karajan1001 (approved):

dvc/parsing/interpolate.py
skshetry: Just noticed this, it's a bad idea to make `dvc.yaml` parsing conditional, it should be self-contained.

PR #7908, opened 2022-06-16
config: remove core.jobs dead code

### LINKED ISSUE #1626: config: option for `--jobs` value
Several DVC commands have a `--jobs` option that has default values (e.g, https://dvc.org/doc/commands-reference/gc).

Afaik there is no way to change the default value. Having the option to change it through `dvc config` would be useful.

Can you consider adding it?

Thanks

dberenbaum:
Fixes #1626. There was a `core.jobs` option that's not supported by the schema nor documented. I guess it was intended to apply across multiple remotes, but nobody has asked about it, so I'm proposing to remove it and close the issue.

### DIFF
--- a/dvc/fs/__init__.py
+++ b/dvc/fs/__init__.py
@@ -104,11 +104,6 @@ def get_cloud_fs(repo, **kwargs):
     except Invalid as exc:
         raise RepoConfigError(str(exc)) from None
 
-    if "jobs" not in remote_conf:
-        jobs = core_config.get("jobs")
-        if jobs:
-            remote_conf["jobs"] = jobs
-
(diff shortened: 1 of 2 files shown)


skshetry:
> Fixes #1626. There was a `core.jobs` option that's not supported by the schema nor documented.

It is specified in the schema:
https://github.com/iterative/dvc/blob/830c116b4e1e01eaeff024e3e64c442d21b09cdb/dvc/config_schema.py#L80

dberenbaum:
@skshetry it's specified under remote but not under core, which is what this removes. Unless I'm missing something?

dtrifiro (approved):

Issue #7592, opened 2022-04-19
Epic: exp queue CLI


pmrowla:
Meta-issue for initial planned `exp queue` issues. (This does not cover "nice to haves" as outlined in the [notion proposal](https://www.notion.so/iterative/Queueing-Managing-Experiment-Execution-bb07bf856cd242bd98a2c87cfc6e75d7))

dvc-task/exp backend tasks:

* [x] https://github.com/iterative/dvc/issues/7587

`exp queue` CLI tasks:

* [x] https://github.com/iterative/dvc/issues/7589
* [x] https://github.com/iterative/dvc/issues/7590
* [x] https://github.com/iterative/dvc/issues/7591

daavoo:
Should we close https://github.com/iterative/dvc/issues/5615 in favor of this one?

pmrowla:
I think we can leave it open for now, this is more just tasking for the current planned work, but the original discussion contains some more features that are currently unplanned/out of scope

pmrowla:
@dberenbaum `queue start/stop/remove/kill` are all available in the dev branch for testing now. The status commands and `stop --kill` will be added in follow-up PRs.

dberenbaum:
This is looking good now! I wanted to summarize some of the questions/issues I still have:

1. What to do with existing commands and flags? Specifically, `dvc exp run --run-all` and `dvc exp remove --queue`?
2. How can we check if the queue is active/running/processing?
3. Should we show both queued and failed experiments in `dvc exp show` and have flags to show/hide them?
4. How should queued experiments be named and should they match the final experiment name?

pmrowla:
> 1. What to do with existing commands and flags? Specifically, `dvc exp run --run-all` and `dvc exp remove --queue`?

I think the existing commands should essentially be aliases to the new behavior, and formally deprecated, so when the user runs them we would output something like

> `dvc exp run --run-all` is deprecated and will be removed in a future DVC release. Please use `dvc queue start` instead.

For the aliased behavior:

- `exp run --temp` should just be `exp run --queue` + `queue start` + `queue logs <exp>`
- `exp run --run-all` should just be `queue start`. I don't think we should try to duplicate the existing log behavior for `--run-all` since it wasn't particularly useful anyways (instead we can just output a message saying use `queue logs` to view logs for a specific experiment)
- `exp remove --queue` should be something like `queue remove --all`
[...]

PR #7924, opened 2022-06-21
queue remove: can remove successed,failed tasks msgs

karajan1001:
Followed from [#7592](https://github.com/iterative/dvc/issues/7592#issuecomment-1157176923) 

1. Seperate `remove` method to a new file.
2. Add `queued`, `failed`, `processed` flags to remove.
3. Add new unit tests for `queue status --queue/fail/success`
4. Implement methods to remove done tasks.
5. add two new unit test for `celery_queue.remove`
6. bump into dvc-task version 0.0.13

A demo for this PR.
[asciicast](https://asciinema.org/a/PCejdbblayS3hkBmiRZmRTAwB)

### DIFF
--- a/dvc/repo/experiments/queue/base.py
+++ b/dvc/repo/experiments/queue/base.py
@@ -131,15 +131,27 @@ class BaseStashQueue(ABC):
     def get(self) -> QueueGetResult:
         """Pop and return the first item in the queue."""
 
-    def remove(self, revs: Collection[str]) -> List[str]:
+    def remove(
+        self,
+        revs: Collection[str],
+        _all: bool = False,
+        queued: bool = False,
+        **kwargs,
+    ) -> List[str]:
         """Remove the specified entries from the queue.
 
         Arguments:
             revs: Stash revisions or queued exp names to be removed.
+            queued: Remove all queued tasks.
+            all: Remove all tasks.
 
         Returns:
             Revisions (or names) which were removed.
         """
+
+        if _all or queued:
+            return self.clear()
+
         removed: List[str] = []
         to_remove: Dict[str, ExpStashEntry] = {}
         queue_entries = self.match_queue_entry_by_name(
@@ -152,18 +164,15 @@ class BaseStashQueue(ABC):
                 ]
                 removed.append(name)
 
-        self._remove_revs(to_remove)
+        self._remove_revs(to_remove, self.stash)
         return removed
 
-    def clear(self) -> List[str]:
-        """Remove all entries from the queue.
-
-        Returns:
-            Revisions which were removed.
-        """
+    def clear(self, **kwargs) -> List[str]:
+        """Remove all entries from the queue."""
         stash_revs = self.stash.stash_revs
         removed = list(stash_revs)
-        self._remove_revs(stash_revs)
+        self._remove_revs(stash_revs, self.stash)
+
         return removed
 
     def status(self) -> List[Dict[str, Any]]:
@@ -176,27 +185,13 @@ class BaseStashQueue(ABC):
             commit = self.scm.resolve_commit(rev)
             return datetime.fromtimestamp(commit.commit_time)
 
-        failed_revs: Dict[str, ExpStashEntry] = (
-            dict(self.failed_stash.stash_revs)
-            if self.failed_stash is not None
-            else {}
-        )
-
         def _format_entry(
             entry: QueueEntry,
             status: str = "Unknown",
-            result: Optional[ExecutorResult] = None,
         ) -> Dict[str, Any]:
             name = entry.name
             # NOTE: We fallback to Unknown status for experiments
             # generated in prior (incompatible) DVC versions
-            if result is None and entry.stash_rev in failed_revs:
-                status = "Failed"
-            elif result is not None:
-                if result.exp_hash:
-                    if result.ref_info:
-                        status = "Success"
-                        name = result.ref_info.name
             return {
                 "rev": entry.stash_rev,
                 "name": name,
@@ -213,12 +208,17 @@ class BaseStashQueue(ABC):
             for queue_entry in self.iter_queued()
         )
         result.extend(
-            _format_entry(queue_entry, result=exp_result)
-            for queue_entry, exp_result in self.iter_done()
+            _format_entry(queue_entry, status="Failed")
+            for queue_entry, _ in self.iter_failed()
+        )
+        result.extend(
+            _format_entry(queue_entry, status="Success")
+            for queue_entry, _ in self.iter_success()
         )
         return result
 
-    def _remove_revs(self, stash_revs: Mapping[str, ExpStashEntry]):
+    @staticmethod
+    def _remove_revs(stash_revs: Mapping[str, ExpStashEntry], stash: ExpStash):
         """Remove the specified entries from the queue by stash revision."""
         for index in sorted(
             (
@@ -228,7 +228,7 @@ class BaseStashQueue(ABC):
             ),
             reverse=True,
         ):
-            self.stash.drop(index)
+            stash.drop(index)
 
     @abstractmethod
     def iter_queued(self) -> Generator[QueueEntry, None, None]:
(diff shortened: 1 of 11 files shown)


karajan1001:
@pmrowla 
The current problem is that the queued experiment I prepared will be executed wrongly by some celery worker in the background, making the tests really unstable. Any ideas about it?

dvc/repo/experiments/queue/base.py
```diff
@@ -131,15 +131,27 @@ def put(self, *args, **kwargs) -> QueueEntry:
     def get(self) -> QueueGetResult:
         """Pop and return the first item in the queue."""
 
-    def remove(self, revs: Collection[str]) -> List[str]:
+    def remove(
+        self,
+        revs: Collection[str],
+        _all: bool = False,
```
pmrowla, line 137:
```suggestion
        all_: bool = False,
```

When trying to avoid collisions with reserved keywords either use a trailing underscore or change the name to something else like `remove_all`.

Try not to use a leading underscore for this (like `_all`), since the leading underscore is usually used to indicate either that a var is supposed to be private or that it is intentionally unused.

tests/func/experiments/test_queue.py
```diff
@@ -0,0 +1,149 @@
...
+        dvc.experiments.run(
+            failed_exp_stage.addressing,
+            params=[f"foo={i+queue_length}"],
+            queue=True,
+            name=name,
+        )
+    dvc.experiments.run(run_all=True)
+    output = capsys.readouterr()
+    assert "Failed to reproduce experiment" in output.err
+    return name_list
```
pmrowla, line 60: Fixtures should not contain tests themselves, they should only provide any necessary setup and/or cleanup needing to run tests that use the fixture.

It's general it would be ok to have a test fixture that sets up some set of failed experiments, but `test_failed_exp_celery` should be separate from that. It does not make sense to have a repeated UI test that runs every time we re-use that fixture.

However, I think that you probably won't be able to use fixtures that run celery experiments in this way. We will end up hitting race conditions due to how the pytest-celery worker executes tasks.

Basically it will be safer to not use fixtures that run experiments, and setup everything in individual tests.

tests/func/experiments/test_queue.py
```diff
@@ -0,0 +1,149 @@
...
+    assert len(status) == 9
+    for task in failed_tasks:
+        assert status[task] == "Failed"
+    for task in success_tasks:
+        assert status[task] == "Success"
+    for task in queued_tasks:
+        assert status[task] == "Queued"
+
+
+def test_queue_remove(dvc, failed_tasks, success_tasks, queued_tasks):
```
pmrowla, line 98: We don't really need a func test for this since we already have a unit test for queue remove. The only additional thing the func test is doing is `status()` checks, but `status()` should be tested separately anyways.

tests/func/experiments/test_queue.py
```diff
@@ -0,0 +1,149 @@
...
+    queue = dvc.experiments.celery_queue
+    done_result = first(queue.iter_done())
+    name = done_result.entry.stash_rev
+    captured = capsys.readouterr()
+    queue.logs(name, follow=follow)
+    captured = capsys.readouterr()
+    assert "failed to reproduce 'failed-copy-file'" in captured.out
+
+
+def test_queue_status(dvc, failed_tasks, success_tasks, queued_tasks):
```
pmrowla, line 85: Rather than executing experiments, it would be better to test `status()` unit test `status()` with mocked results of `iter_queued/iter_active/iter_done`

pmrowla (changes requested):

pmrowla:
> @pmrowla The only problem remained is that the queued experiment I prepared will be executed wrongly by some celery worker in the background, making the tests really unstable. Any ideas about it?

@karajan1001 it's likely due to using the celery pytest worker inside multiple fixtures at once.

In general, we should try to move away from adding complex func tests in favor of unit testing whenever possible. In this case, for celery we do need a couple of func tests to make sure that execution and the UI work properly (in both failed and successful cases), but for testing specific backend or API calls (like remove/status) it's better to use unit tests + mocked celery behavior

pmrowla:
One other thing to note here regarding commits, ideally we want to keep changes in separate commits where possible. In this PR it would be nice to have the separation with:

- 1 commit for the dependency version bump. If there are specific changes required to make the new dep version work with DVC then those changes can also go into the version bump commit (like when a dependency has a backwards incompatible API change)
- 1 commit for the `queue remove` related changes
- 1 commit for the `queue status` related changes

Basically for small "review fix" type changes it makes sense to squash them into the larger commit, but for larger independent changes that affect different parts of the code base or different parts of the CLI/API, having separate commits makes it easier to keep track of what changed and when. So follow-up changes/fixes to `remove` or `status` can get squashed into each relevant commit before merging the PR, but ideally the final PR would not contain commits that mix changes to both `remove` and `status`.

---

This isn't really a requirement for this PR, but it's something to be aware of in the future

karajan1001:
> One other thing to note here regarding commits, ideally we want to keep changes in separate commits where possible. In this PR it would be nice to have the separation with:
> 
> * 1 commit for the dependency version bump. If there are specific changes required to make the new dep version work with DVC then those changes can also go into the version bump commit (like when a dependency has a backwards incompatible API change)
> * 1 commit for the `queue remove` related changes
> * 1 commit for the `queue status` related changes

After removing the functional test of `queue status`. It only contains `queue remove` related changes and the dependency bump. I can separate them later.

dvc/repo/experiments/queue/base.py
pmrowla, line 137:
```suggestion
        all_: bool = False,
```

looks like this one was missed

dvc/repo/experiments/queue/remove.py
```diff
@@ -0,0 +1,185 @@
...
+        if success:
+            done_tasks.extend(self.iter_success())
+        done_entries = [result.entry for result in done_tasks]
+        _remove_done_tasks(self, done_entries)
+        removed.extend(_get_names(done_tasks))
+
+    return removed
+
+
+def remove(
```
pmrowla, line 128:
```suggestion
def celery_remove(
```

Since this is in `queue/remove.py` and is specific to the Celery version of the queue, make sure to name it appropriately so it's obvious this is not used in the base queue or other local queue types

dvc/repo/experiments/queue/remove.py
```diff
@@ -0,0 +1,185 @@
...
+    all_: bool = False,
+) -> List[str]:
+    """Remove the specified entries from the queue.
+
+    Arguments:
+        revs: Stash revisions or queued exp names to be removed.
+        queued: Remove all queued tasks.
+        failed: Remove all failed tasks.
+        success: Remove all success tasks.
+        all: Remove all tasks.
```
pmrowla, line 143:
```suggestion
        all_: Remove all tasks.
```

dvc/repo/experiments/queue/remove.py
```diff
@@ -0,0 +1,185 @@
...
+                names.append(entry.result.ref_info.name)
+                continue
+            entry = entry.entry
+        name = entry.name
+        name = name or entry.stash_rev[:7]
+        names.append(name)
+    return names
+
+
+def clear(self: "LocalCeleryQueue", **kwargs) -> List[str]:
```
pmrowla, line 95:
```suggestion
def celery_clear(self: "LocalCeleryQueue", **kwargs) -> List[str]:
```

dvc/repo/experiments/queue/base.py
```diff
@@ -213,12 +208,17 @@ def _format_entry(
...
+        )
+        result.extend(
+            _format_entry(queue_entry, status="Success")
+            for queue_entry, _ in self.iter_success()
         )
         return result
 
-    def _remove_revs(self, stash_revs: Mapping[str, ExpStashEntry]):
+    @staticmethod
+    def _remove_revs(stash_revs: Mapping[str, ExpStashEntry], stash: ExpStash):
```
pmrowla, line 221: Since this is no longer really a private method it should be moved out of the queue classes and into `experiments/stash.py` as `ExpStash.remove_revs(self, stash_revs)`

tests/func/experiments/test_queue.py
```diff
@@ -0,0 +1,104 @@
...
+    queue = dvc.experiments.celery_queue
+    done_result = first(queue.iter_done())
+    name = done_result.entry.stash_rev
+    captured = capsys.readouterr()
+    queue.logs(name, follow=follow)
+    captured = capsys.readouterr()
+    assert "failed to reproduce 'failed-copy-file'" in captured.out
+
+
+def test_queue_remove_done(dvc, failed_tasks, success_tasks):
```
pmrowla, line 67: Do we need still need this test? With unit tests for `remove()` we should not need any additional func tests here

pmrowla (changes requested):

tests/func/experiments/test_queue.py
karajan1001: I think we can simplify them. We tested the remove method case by case as well as the cli test. But we don't have a place to test the whole process. This is why I added a functional test here, But I think we can simplify it, We don't need to test all of the cases, but only one of them.

pmrowla (approved):

PR #7943, opened 2022-06-28
introduce new data:status command

skshetry:
See [Consolidate Repo Status](https://www.notion.so/iterative/Consolidate-repo-status-ed3cd60f706f4fcaba1d3f3cac1498e9).

### DIFF
--- a/dvc/commands/data.py
+++ b/dvc/commands/data.py
@@ -1,0 +1,174 @@
+import argparse
+import logging
+from typing import TYPE_CHECKING
+
+from funcy import compact, log_durations
+
+from dvc.cli.command import CmdBase
+from dvc.cli.utils import append_doc_link, fix_subparsers
+from dvc.ui import ui
+
+if TYPE_CHECKING:
+    from dvc.repo.data import Status as DataStatus
+
+
+logger = logging.getLogger(__name__)
+
+
+class CmdDataStatus(CmdBase):
+    COLORS = {
+        "not_in_cache": "red",
+        "committed": "green",
+        "uncommitted": "yellow",
+        "untracked": "cyan",
+    }
+    LABELS = {
+        "not_in_cache": "Not in cache",
+        "committed": "DVC committed changes",
+        "uncommitted": "DVC uncommitted changes",
+        "untracked": "Untracked files",
+        "unchanged": "DVC unchanged files",
+    }
+    HINTS = {
+        "not_in_cache": 'use "dvc pull <file>..." '
+        "to update your local storage",
+        "committed": "git commit the corresponding dvc files "
+        "to update the repo",
+        "uncommitted": 'use "dvc commit <file>..." to track changes',
+        "untracked": 'use "git add <file> ..." or '
+        'dvc add <file>..." to commit to git or to dvc',
+        "git_dirty": "there are {}changes not tracked by dvc, "
+        'use "git status" to see',
+    }
+
+    def _process_status(self, status: "DataStatus"):
+        """Flatten stage status, and filter empty stage status contents."""
+        for stage, stage_status in status.items():
+            items = stage_status
+            if isinstance(stage_status, dict):
+                items = {
+                    file: state
+                    for state, files in stage_status.items()
+                    for file in files
+                }
+            if not items:
+                continue
+            yield stage, items
+
+    def _show_status(self, status: "DataStatus") -> int:
+        git_info = status.pop("git")  # type: ignore[misc]
+        result = dict(self._process_status(status))
+        if not result:
+            no_changes = "No changes"
+            if git_info.get("is_empty", False):
+                no_changes += " in an empty git repo"
+            ui.write(f"{no_changes}.")
+
+        for idx, (stage, stage_status) in enumerate(result.items()):
+            if idx:
+                ui.write()
+
+            label = self.LABELS.get(stage, stage.capitalize() + " files")
+            header = f"{label}:"
+            color = self.COLORS.get(stage, "normal")
+
+            ui.write(header)
+            if hint := self.HINTS.get(stage):
+                ui.write(f"  ({hint})")
+
+            if isinstance(stage_status, dict):
+                items = [
+                    ": ".join([state, file])
+                    for file, state in stage_status.items()
+                ]
+            else:
+                items = stage_status
+
+            for item in items:
+                ui.write(f"\t[{color}]{item}[/]".expandtabs(8), styled=True)
+
+        if (hint := self.HINTS.get("git_dirty")) and git_info.get("is_dirty"):
+            message = hint.format("other " if result else "")
+            ui.write(f"[blue]({message})[/]", styled=True)
+        return 0
+
+    def run(self) -> int:
+        with log_durations(logger.trace, "in data_status"):  # type: ignore
+            status = self.repo.data_status(
+                granular=self.args.granular,
+                untracked_files=self.args.untracked_files,
+                with_dirs=self.args.with_dirs,
+            )
+
+        if not self.args.unchanged:
+            status.pop("unchanged")  # type: ignore[misc]
+        if self.args.json:
+            status.pop("git")  # type: ignore[misc]
+            ui.write_json(compact(status))
+            return 0
+        return self._show_status(status)
+
+
+def add_parser(subparsers, parent_parser):
+    data_parser = subparsers.add_parser(
+        "data",
+        parents=[parent_parser],
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+    data_subparsers = data_parser.add_subparsers(
+        dest="cmd",
+        help="Use `dvc data CMD --help` to display command-specific help.",
+    )
+    fix_subparsers(data_subparsers)
+
+    DATA_STATUS_HELP = (
+        "Show changes between the last git commit, "
+        "the dvcfiles and the workspace."
+    )
+    data_status_parser = data_subparsers.add_parser(
+        "status",
+        parents=[parent_parser],
+        description=append_doc_link(DATA_STATUS_HELP, "data/status"),
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        help=DATA_STATUS_HELP,
+    )
+    data_status_parser.add_argument(
+        "--json",
+        action="store_true",
+        default=False,
+        help="Show output in JSON format.",
+    )
+    data_status_parser.add_argument(
+        "--show-json",
+        action="store_true",
+        default=False,
+        dest="json",
+        help=argparse.SUPPRESS,
+    )
+    data_status_parser.add_argument(
+        "--granular",
+        action="store_true",
+        default=False,
+        help="Show granular file-level info for DVC-tracked directories.",
+    )
+    data_status_parser.add_argument(
+        "--unchanged",
+        action="store_true",
+        default=False,
+        help="Show unmodified DVC-tracked files.",
+    )
+    data_status_parser.add_argument(
+        "--untracked-files",
+        choices=["no", "all"],
+        default="no",
+        const="all",
+        nargs="?",
+        help="Show untracked files.",
+    )
+    data_status_parser.add_argument(
+        "--with-dirs",
+        action="store_true",
+        default=False,
+        help=argparse.SUPPRESS,
+    )
+    data_status_parser.set_defaults(func=CmdDataStatus)
--- a/dvc/commands/data.py
+++ b/dvc/commands/data.py
@@ -41,7 +41,8 @@ class CmdDataStatus(CmdBase):
         'use "git status" to see',
     }
 
-    def _process_status(self, status: "DataStatus"):
+    @staticmethod
+    def _process_status(status: "DataStatus"):
         """Flatten stage status, and filter empty stage status contents."""
         for stage, stage_status in status.items():
             items = stage_status
@@ -55,9 +56,10 @@ class CmdDataStatus(CmdBase):
                 continue
             yield stage, items
 
-    def _show_status(self, status: "DataStatus") -> int:
+    @classmethod
+    def _show_status(cls, status: "DataStatus") -> int:
         git_info = status.pop("git")  # type: ignore[misc]
-        result = dict(self._process_status(status))
+        result = dict(cls._process_status(status))
         if not result:
             no_changes = "No changes"
             if git_info.get("is_empty", False):
@@ -68,12 +70,12 @@ class CmdDataStatus(CmdBase):
             if idx:
                 ui.write()
 
-            label = self.LABELS.get(stage, stage.capitalize() + " files")
+            label = cls.LABELS.get(stage, stage.capitalize() + " files")
             header = f"{label}:"
-            color = self.COLORS.get(stage, "normal")
+            color = cls.COLORS.get(stage, "normal")
 
             ui.write(header)
-            if hint := self.HINTS.get(stage):
+            if hint := cls.HINTS.get(stage):
                 ui.write(f"  ({hint})")
 
             if isinstance(stage_status, dict):
@@ -87,7 +89,7 @@ class CmdDataStatus(CmdBase):
             for item in items:
                 ui.write(f"\t[{color}]{item}[/]".expandtabs(8), styled=True)
 
-        if (hint := self.HINTS.get("git_dirty")) and git_info.get("is_dirty"):
+        if (hint := cls.HINTS.get("git_dirty")) and git_info.get("is_dirty"):
             message = hint.format("other " if result else "")
             ui.write(f"[blue]({message})[/]", styled=True)
         return 0
@@ -102,6 +104,8 @@ class CmdDataStatus(CmdBase):
 
         if not self.args.unchanged:
             status.pop("unchanged")  # type: ignore[misc]
+        if self.args.untracked_files == "no":
+            status.pop("untracked")
         if self.args.json:
             status.pop("git")  # type: ignore[misc]
             ui.write_json(compact(status))
--- a/dvc/repo/data.py
+++ b/dvc/repo/data.py
@@ -215,6 +215,17 @@ class Status(TypedDict):
     git: GitInfo
 
 
+def _transform_git_paths_to_dvc(repo: "Repo", files: List[str]):
+    """Transform files rel. to Git root to DVC root, and drop outside files."""
+    rel = repo.fs.path.relpath(repo.root_dir, repo.scm.root_dir).rstrip("/")
+    if rel in (os.curdir, ""):
+        return files
+
+    prefix = rel + os.sep
+    length = len(prefix)
+    return [file[length:] for file in files if file.startswith(prefix)]
+
+
 def status(repo: "Repo", untracked_files: str = "no", **kwargs: Any) -> Status:
     from scmrepo.exceptions import SCMError
 
@@ -235,12 +246,14 @@ def status(repo: "Repo", untracked_files: str = "no", **kwargs: Any) -> Status:
         unchanged &= set(committed_diff.pop("unchanged", []))
 
     git_info = _git_info(repo.scm, untracked_files=untracked_files)
+    untracked = git_info.get("untracked", [])
+    untracked = _transform_git_paths_to_dvc(repo, untracked)
     # order matters here
     return Status(
         not_in_cache=not_in_cache,
         committed=committed_diff,
         uncommitted=uncommitted_diff,
-        untracked=git_info.get("untracked", []),
+        untracked=untracked,
         unchanged=list(unchanged),
         git=git_info,
     )
--- a/dvc/repo/data.py
+++ b/dvc/repo/data.py
@@ -235,6 +235,7 @@ def status(repo: "Repo", untracked_files: str = "no", **kwargs: Any) -> Status:
         unchanged &= set(committed_diff.pop("unchanged", []))
 
     git_info = _git_info(repo.scm, untracked_files=untracked_files)
+    # order matters here
     return Status(
         not_in_cache=not_in_cache,
         committed=committed_diff,
--- a/dvc/repo/data.py
+++ b/dvc/repo/data.py
@@ -1,0 +1,245 @@
+import os
+from collections import defaultdict
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, TypedDict, cast
+
+if TYPE_CHECKING:
+    from scmrepo.base import Base
+
+    from dvc.output import Output
+    from dvc.repo import Repo
+    from dvc_data.hashfile.db import HashFileDB
+    from dvc_data.hashfile.obj import HashFile
+
+
+def _in_cache(obj: Optional["HashFile"], cache: "HashFileDB") -> bool:
+    from dvc_objects.errors import ObjectFormatError
+
+    if not obj:
+        return False
+    if not obj.hash_info.value:
+        return False
+
+    try:
+        cache.check(obj.hash_info.value)
+        return True
+    except (FileNotFoundError, ObjectFormatError):
+        return False
+
+
+def _shallow_diff(
+    root: str,
+    old_obj: Optional["HashFile"],
+    new_obj: Optional["HashFile"],
+    cache: "HashFileDB",
+) -> Dict[str, List[str]]:
+    # TODO: add support for shallow diff in dvc-data
+    # TODO: we may want to recursively do in_cache check
+    d = {}
+
+    from dvc_data.objects.tree import Tree
+
+    if isinstance(new_obj, Tree):
+        root = os.path.sep.join([root, ""])
+
+    if not _in_cache(old_obj, cache):
+        d["not_in_cache"] = [root]
+
+    if old_obj is None and new_obj is None:
+        return d
+    if old_obj is None:
+        return {"added": [root], **d}
+    if new_obj is None:
+        return {"deleted": [root], **d}
+    if old_obj.hash_info != new_obj.hash_info:
+        return {"modified": [root], **d}
+    return {"unchanged": [root], **d}
+
+
+def _granular_diff(
+    root: str,
+    old_obj: Optional["HashFile"],
+    new_obj: Optional["HashFile"],
+    cache: "HashFileDB",
+    with_dirs: bool = False,
+) -> Dict[str, List[str]]:
+    from dvc_data.diff import ROOT
+    from dvc_data.diff import diff as odiff
+    from dvc_data.objects.tree import Tree
+
+    def path_join(root: str, *paths: str) -> str:
+        if not isinstance(new_obj, Tree):
+            return root
+        return os.path.sep.join([root, *paths])
+
+    diff_data = odiff(old_obj, new_obj, cache)
+    drop_root = not with_dirs and isinstance(new_obj, Tree)
+
+    output: Dict[str, List[str]] = defaultdict(list)
+    for state in ("added", "deleted", "modified", "unchanged"):
+        items = getattr(diff_data, state)
+        output[state].extend(
+            path_join(root, *item.new.key)
+            for item in items
+            if not (drop_root and item.new.key == ROOT)
+        )
+        # TODO: PERF: diff is checking not_in_cache for each even if we only
+        # need it for the index.
+        # BUG: not_in_cache file also shows up as modified in staged and
+        # unstaged. We currently don't know if it is really modified.
+        output["not_in_cache"].extend(
+            path_join(root, *item.new.key)
+            for item in items
+            if not item.old.in_cache
+            and not (drop_root and item.new.key == ROOT)
+            and state != "added"
+        )
+    return output
+
+
+def _diff(
+    root: str,
+    old_obj: Optional["HashFile"],
+    new_obj: Optional["HashFile"],
+    cache: "HashFileDB",
+    granular: bool = False,
+    with_dirs: bool = False,
+) -> Dict[str, List[str]]:
+    if granular:
+        return _granular_diff(
+            root, old_obj, new_obj, cache, with_dirs=with_dirs
+        )
+    return _shallow_diff(root, old_obj, new_obj, cache)
+
+
+class GitInfo(TypedDict, total=False):
+    staged: Dict[str, List[str]]
+    unstaged: Dict[str, List[str]]
+    untracked: List[str]
+    is_empty: bool
+    is_dirty: bool
+
+
+def _git_info(scm: "Base", untracked_files: str = "all") -> GitInfo:
+    from scmrepo.exceptions import SCMError
+
+    from dvc.scm import NoSCM
+
+    if isinstance(scm, NoSCM):
+        return {}
+
+    try:
+        scm.get_rev()
+    except SCMError:
+        empty_repo = True
+    else:
+        empty_repo = False
+
+    staged, unstaged, untracked = scm.status(untracked_files=untracked_files)
+    # NOTE: order is important here.
+    return GitInfo(
+        staged=staged,
+        unstaged=unstaged,
+        untracked=untracked,
+        is_empty=empty_repo,
+        is_dirty=any([staged, unstaged, untracked]),
+    )
+
+
+def _diff_index_to_wtree(repo: "Repo", **kwargs: Any) -> Dict[str, List[str]]:
+    from dvc_data.build import build
+
+    unstaged_diff = defaultdict(list)
+    for out in repo.index.outs:
+        out = cast("Output", out)
+        if not out.use_cache:
+            continue
+
+        try:
+            _, _, new = build(
+                out.odb,
+                out.fs_path,
+                out.fs,
+                out.fs.PARAM_CHECKSUM,
+                ignore=out.dvcignore,
+                dry_run=True,
+            )
+        except FileNotFoundError:
+            new = None
+
+        cache = repo.odb.local
+        root = str(out)
+        old = out.get_obj()
+        d = _diff(root, old, new, cache, **kwargs)
+        for state, items in d.items():
+            if not items:
+                continue
+            unstaged_diff[state].extend(items)
+    return unstaged_diff
+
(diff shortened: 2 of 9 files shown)


dberenbaum:
Very minor suggestion: Can we make the "uncommitted changes" red instead of green?

skshetry:
> Very minor suggestion: Can we make the "uncommitted changes" red instead of green?

You mean the header or the complete section?

skshetry:
I don't think `uncommitted changes` has to be red, it should be yellow at most. It's more useful for not-in-cache. :)

dvc/commands/data.py
```diff
@@ -0,0 +1,339 @@
...
+    return shallow_diff(out, old_obj, new_obj, cache)
+
+
+class CmdDataStatus(CmdBase):
+    COLORS = {
+        "not_in_cache": "red",
+        "unstaged": "green",
+        "staged": "yellow",
+        "untracked": "cyan",
+    }
```
skshetry, line 115:
```suggestion
    COLORS = {
        "not_in_cache": "red",
        "unstaged": "yellow",
        "staged": "green",
        "untracked": "cyan",
    }
```

dberenbaum:
Need to think about how to handle `--untracked-files`:

* Should we show them by default? Seems a bit different from Git because:
  * Without `normal` option, it will get really busy.
  * Users don't expect to track the entire repo with DVC.
  * If we suggest doing `dvc data status` and `git status` as a pair, they become redundant for untracked files.
* Since it's different from Git, should we try not to copy the same syntax? What about just a boolean flag? What about a slightly different name (`--untracked`)?

skshetry:
I'm finding the `untracked` section quite useful. I think in parallel, we should see if we can implement `untracked-files=normal` easily.
If it's too slow, we can roll back to not showing untracked files by default. If it's difficult to implement, then we can drop `normal` and don't show by default.

dberenbaum:
If I have a tracked directory and am missing the cache, it will show as added even if the `.dvc` file hasn't changed:

```bash
$ mkdir data
$ dvc add data
$ git add .
$ git commit -m "add data"
$ rm -rf .dvc/cache
$ dvc data status
Not in cache:
  (use "dvc pull <file>..." to update your local storage)
        data

DVC uncommitted changes:
  (use "dvc commit <file>..." to track changes)
        added: data/
```

dberenbaum:
Can we ignore pipelines outputs where `cache: false`?

For example, if I run an experiment in https://github.com/iterative/example-get-started, the output looks like:

```bash
$ dvc data status
DVC committed changes:
  (git commit the corresponding dvc files to update the repo)
        modified: data/prepared/
        modified: data/features/
        modified: model.pkl
        modified: scores.json
        modified: prc.json
        modified: roc.json
(there are other changes not tracked by dvc, use "git status" to see)
$ git status
On branch master
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   dvc.lock
        modified:   params.yaml
        modified:   prc.json
        modified:   roc.json
        modified:   scores.json

no changes added to commit (use "git add" and/or "git commit -a")
```

Ideally, `dvc data status` would not show scores.json, prc.json, or roc.json.

dvc/commands/data.py
dberenbaum: I don't think they all need unique colors. In Git, both untracked and unstaged are red. Can we do the same: make everything red except for staged/committed? WDYT?

dberenbaum: Missed your comment above. I'm fine with whatever you decide.

skshetry:
> Can we ignore pipelines outputs where `cache: false`?

It should be ignored, I tried fixing it in [231805a](https://github.com/iterative/dvc/pull/7943/commits/231805a386adf479caca98e10beb08db64d3a044). I don't see in example-get-started locally. Have you fetched the latest commit from this branch?

https://github.com/skshetry/dvc/blob/de58c695516908b49cd62fe22ed7b17c64339850/dvc/commands/data.py#L259

dberenbaum:
Looks amazing @skshetry!

If we can decide on the default behavior for untracked files and address https://github.com/iterative/dvc/pull/7943#issuecomment-1172623234 and https://github.com/iterative/dvc/pull/7943#issuecomment-1172624914, I think it's ready to merge.

For untracked files, I still think we should exclude by default. Even if we can support normal mode (and we can't for now), the other reasons still apply. I agree it's useful when debugging, but in real usage I don't think it is. Most people treat DVC-tracked outputs as special for large data and wouldn't expect DVC to show the state of the whole repo, especially since they still need to use git status alongside it.

skshetry:
> If I have a tracked directory and am missing the cache, it will show as added even if the `.dvc` file hasn't changed:

I think this is https://github.com/iterative/dvc/issues/7661. I have also noticed that. Will require a fix upstream.

https://github.com/skshetry/dvc/blob/de58c695516908b49cd62fe22ed7b17c64339850/dvc/commands/data.py#L85-L86

dberenbaum:
> > Can we ignore pipelines outputs where `cache: false`?
> 
> It should be ignored, I tried fixing it in [231805a](https://github.com/iterative/dvc/pull/7943/commits/231805a386adf479caca98e10beb08db64d3a044). I don't see in example-get-started locally. Have you fetched the latest commit from this branch?

Yeah, unfortunately it looks like it's still an issue.

```
$ dvc -V
2.11.1.dev63+g4e22e4178
```

skshetry:
> Can we ignore pipelines outputs where `cache: false`?

Sorry, this one is different. Even though it's a pipeline outputs, it's a data being tracked by DVC, right?

dberenbaum:
> Sorry, this one is different. Even though it's a pipeline output, it's data being tracked by DVC, right?

Right, I just noticed that was about not-in-cache.

🤔  I see what you mean, but I think we should ignore `cache: false` outputs. They are being tracked for pipeline purposes. Since we are trying to differentiate between pipeline and data management functionality here, I don't think it belongs. DVC isn't managing those file versions.

For example:

```bash
$ dvc data status
DVC committed changes:
  (git commit the corresponding dvc files to update the repo)
        modified: data/prepared/
        modified: data/features/
        modified: model.pkl
        modified: scores.json
        modified: prc.json
        modified: roc.json
(there are other changes not tracked by dvc, use "git status" to see)

$ git restore dvc.lock

$ dvc checkout
M       data/prepared/
M       model.pkl
M       data/features/

$ dvc data status
DVC uncommitted changes:
  (use "dvc commit <file>..." to track changes)
        modified: scores.json
        modified: prc.json
        modified: roc.json
(there are other changes not tracked by dvc, use "git status" to see)
```

skshetry:
> thinking I see what you mean, but I think we should ignore `cache: false` outputs. They are being tracked for pipeline purposes. Since we are trying to differentiate between pipeline and data management functionality here, I don't think it belongs. DVC isn't managing those file versions.

Fixed in 29c2fa1.

skshetry:
> For untracked files, I still think we should exclude by default. Even if we can support normal mode (and we can't for now), the other reasons still apply. I agree it's useful when debugging, but in real usage I don't think it is. Most people treat DVC-tracked outputs as special for large data and wouldn't expect DVC to show the state of the whole repo, especially since they still need to use git status alongside it.

With b3dd1e020, `--untracked-files` are disabled by default.

dberenbaum (approved):

skshetry:
@mattseddon, could you please take a look at this PR? 

What you need is implemented as `dvc data status —untracked-files —unmodified` (no other flags needed).

This currently does not detect renames, and does not check for external remotes changes (but I dont think VSCode needs this for now). Let me know how it works. 🙂

mattseddon:
> @mattseddon, could you please take a look at this PR?
> 
> What you need is implemented as `dvc data status --json --granular --unmodified --untracked-files` (no other flags needed).
> 
> This currently does not detect renames, and does not check for external remotes changes (but I dont think VSCode needs this for now). Let me know how it works. 🙂

@skshetry having some IRL technical difficulties today should be able to confirm tomorrow 👍🏻. Thank you.

mattseddon:
Hey @skshetry, 

Thanks for giving me the chance to take a look. The output looks good.

One thing that I would ask for is that the `--granular` output also contains the non-granular output. E.g for the `example-get-started` repo with changes in the workspace we have:

```bash
❯ dvc data status --json --unmodified --untracked-files
{                                                                                                                                                   
  "dvc-uncommitted": {
    "modified": [
      "data/prepared/",
      "data/features/",
      "model.pkl"
    ]
  },
  "dvc-unmodified": [
    "data/data.xml"
  ],
  "untracked": [
    "some/untracked/file.txt"
  ]
}
```

and with granular we get:

```bash
❯ dvc data status --json --granular --unmodified --untracked-files
{                                                                                                                                                   
  "dvc-uncommitted": {
    "modified": [
      "data/prepared/train.tsv",
      "data/prepared/test.tsv",
      "data/features/test.pkl",
      "model.pkl"
    ],
    "deleted": [
      "data/features/train.pkl"
    ]
  },
  "dvc-unmodified": [
    "data/data.xml"
  ],
  "untracked": [
    "some/untracked/file.txt"
  ]
}
```

The desired output would be:

```bash
❯ dvc data status --json --granular --unmodified --untracked-files
{                                                                                                                                                   
  "dvc-uncommitted": {
    "modified": [
      "data/prepared/",
      "data/prepared/train.tsv",
      "data/prepared/test.tsv",
      "data/features/",
      "data/features/test.pkl",
      "model.pkl"
    ],
    "deleted": [
      "data/features/train.pkl"
    ]
  },
  "dvc-unmodified": [
    "data/data.xml"
  ],
  "untracked": [
    "some/untracked/file.txt"
  ]
}
```

The reason for this request is that we need to know that `data/prepared` & `data/features` are `deps`/`outs` (as shown in the `dvc.yaml`) so that we can perform actions on them directly if the user clicks on them in the UI (e.g. `push`/`pull`). We are currently relying on the `exp show` data for this information and it would be really good to not have to do that. It also means won't have to duplicate the logic to decorate the parent directory. It would be good to get this information for all sections excluding untracked.

Also came across this bug:
```
        # BUG: not_in_cache file also shows up as modified in staged and
        # unstaged. We currently don't know if it is really modified.
```

We currently work around a few problems with not in cache in the extension. It looks like we exclude anything that is not in cache from the modified/deleted statuses. I can see that you're still working on this.

LMK if there is anything else that you need from me now.

### Only partially related:

@dberenbaum do you think we should change the group names in the SCM view to Committed/Uncommitted Changes? What we currently have are two groups. 

One called "Changes" (which would match to uncommitted):

image

And another called "Ready For Git Commit" (matches to committed):

image

"Changes" does exist in the Git SCM view for unstaged changes. When these resources are `git add`(ed) the group changes to "Staged Changes". We can discuss separately/offline, LMK.

dberenbaum:
> @dberenbaum do you think we should change the group names in the SCM view to Committed/Uncommitted Changes?

I think [DVC] Committed/[DVC] Uncommitted Changes make sense. Comparing to the CLI, we also started by trying to mirror git status. After we had some discussion, it seemed more straightforward to use committed/uncommitted since those terms are already somewhat established for DVC. WDYT?

jorgeorpinel:
Hiy. I see there's a long convo but the issue description doesn't have any context. Is there any one place to start from? An internal doc or a quick summary (motivation)?

> OR, put it in a docs PR for https://github.com/iterative/dvc.org/issues/3732 🙂

mattseddon:
> Hiy. I see there's a long convo but the issue description doesn't have any context. Is there any one place to start from? An internal doc or a quick summary (motivation)? I'm guessing VS Code Extension integration-related.
> 
> > And/OR, put it in a docs PR for [iterative/dvc.org#3732](https://github.com/iterative/dvc.org/issues/3732) 🙂

[Consolidate Repo Status](https://www.notion.so/iterative/Consolidate-repo-status-ed3cd60f706f4fcaba1d3f3cac1498e9)

dvc/commands/data.py
```diff
@@ -0,0 +1,364 @@
...
+    data_status_parser.add_argument(
+        "--json", action="store_true", default=False
+    )
+    data_status_parser.add_argument(
+        "--granular",
+        action="store_true",
+        default=False,
+    )
+    data_status_parser.add_argument(
+        "--unmodified",
```
dtrifiro, line 353: I find `unmodified` to be a bit strange, how about using `unchanged` instead?

RCdeWit:
Quick thought as mentioned during the DVC demo:

If possible, it might be nice to display a slightly different warning/info when it's only `dvc.lock` that hasn't been updated. To me, it seems that users would get in that scenario if they have used `dvc commit` but not `git commit` yet.

So instead of

```there are {}changes not tracked by dvc, "
        'use "git status" to see'```
        
Something like

```
All changes have been committed to DVC, but 
the changes to "dvc.lock" have not been committed
to Git yet. Use "git add . && git commit" 
```

Then again, perhaps this is overcomplicating it. The current warning already nudges users in the right direction. 😄

File: dvc/commands/data.py (hunk 2 above)
Comment (skshetry) [reply dvc/commands/data.py:353]: `unchanged` makes sense to me as well, as in `change -> replace` and `modify -> update`. We have `modified` to mean that something was updated, but `unmodified` is not quite the opposite of `modified`, it's the opposite of `modified + added + deleted + ...` etc, as in not replaced/changed.

(Note that it's how it works internally, I have been remapping on the CLI.) 

cc @dberenbaum

**Comment by skshetry** [comment]
> Quick thought as mentioned during the DVC demo:
> 
> If possible, it might be nice to display a slightly different warning/info when it's only `dvc.lock` that hasn't been updated. To me, it seems that users would get in that scenario if they have used `dvc commit` but not `git commit` yet.
> 
> So instead of
> 
> ```
>         'use "git status" to see'
> ```
> 
> Something like
> 
> ```
> All changes have been committed to DVC, but 
> the changes to "dvc.lock" have not been committed
> to Git yet. Use "git add . && git commit" 
> ```
> 
> Then again, perhaps this is overcomplicating it. The current warning already nudges users in the right direction. smile

`dvc.yaml`/`dvc.lock` files are also about pipelines, which `data status` does not care. So I am hesitant on specializing the hint. There are also too many files that we care about: `dvc.lock`/`.dvc`/`.gitignore`/`.dvcignore`/`dvc.yaml`, etc.

And this requires just 1 more file in the workspace to break the hint, so I am not sure if this will be very useful given the very narrow window where this hint will be useful.

**Comment by skshetry** [comment]
> The reason for this request is that we need to know that `data/prepared` & `data/features` are `deps`/`outs` (as shown in the `dvc.yaml`) so that we can perform actions on them directly if the user clicks on them in the UI (e.g. `push`/`pull`). We are currently relying on the `exp show` data for this information and it would be really good to not have to do that. It also means won't have to duplicate the logic to decorate the parent directory. It would be good to get this information for all sections excluding untracked.

@mattseddon, I think we can add `--withdirs` flag (cc @dberenbaum). In the long run, for these kinds of information, I think VSCode needs to use `dvc index` or something similar to gather these information.

File: dvc/commands/data.py (hunk 2 above)
Comment (dberenbaum) [reply dvc/commands/data.py:353]: No strong opinion on the name here. I'm fine with `unchanged` if you both prefer.

**Comment by dberenbaum** [comment]
> @mattseddon, I think we can add `--withdirs` flag (cc @dberenbaum). In the long run, for these kinds of information, I think VSCode needs to use `dvc index` or something similar to gather these information.

Yeah, I don't have too strong an opinion. I guess it's more "natural" and similar to git to show either dirs or full paths rather than both, but I think it's fine to either add `--withdirs` or show it that way by default.

**Comment by skshetry** [comment]
@mattseddon @dberenbaum, I have added `--with-dirs`. It might be confusing on CLI to see duplicated entries even on granular.

File: dvc/commands/data.py [hunk 3]
```diff
@@ -0,0 +1,364 @@
...
+    }
+    LABELS = {
+        "not_in_cache": "Not in cache",
+        "staged": "DVC committed changes",
+        "unstaged": "DVC uncommitted changes",
+        "untracked": "Untracked files",
+        "unmodified": "DVC unmodified files",
+    }
+    JSON_LABELS = {
+        "staged": "dvc-committed",
```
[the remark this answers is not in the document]
dberenbaum: I slightly prefer `dvc-committed`/`untracked` to `committed`/`untracked` because the first one at least hints that `untracked` is not `dvc-untracked`. However, it's not a strong opinion, especially for JSON since users of it probably already have a good idea of what each label means. I'm fine with whatever you decide.

skshetry:
> @dberenbaum @skshetry are we going to make this command externally visible, do we need docs for it?

yes, it will be visible and we need docs. I'll start working on it this week, see https://github.com/iterative/dvc.org/issues/3732.

dvc/commands/data.py
```diff
@@ -0,0 +1,387 @@
...
+        "unstaged": "yellow",
+        "staged": "green",
+        "untracked": "cyan",
+    }
+    LABELS = {
+        "not_in_cache": "Not in cache",
+        "staged": "DVC committed changes",
+        "unstaged": "DVC uncommitted changes",
+        "untracked": "Untracked files",
+        "unchanged": "DVC unchanged files",
```
jorgeorpinel, line 134: Could be more descriptive. Ideas:
```suggestion
        "not_in_cache": "Not in DVC cache",
        "staged": "Committed changes to files tracked by DVC",
        "unstaged": "Uncommitted changes to files tracked by DVC",
        "untracked": "Files not tracked by DVC (or Git)",
        "unchanged": "Existing files tracked by DVC (no changes)",
```

jorgeorpinel:
Just a quick thought on

> @mattseddon @dberenbaum, I have added --with-dirs

Shouldn't be hard to merge the JSON output of both calls (with and without `--granular`) as an alternative. Not sure about the performance though

dvc/commands/data.py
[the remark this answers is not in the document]
skshetry: `dvc` is redundant, will remove from JSON output. I am open for reconsideration.

dvc/repo/data.py
```diff
@@ -0,0 +1,246 @@
...
+    return GitInfo(
+        staged=staged,
+        unstaged=unstaged,
+        untracked=untracked,
+        is_empty=empty_repo,
+        is_dirty=any([staged, unstaged, untracked]),
+    )
+
+
+def diff_index_to_wtree(repo: "Repo", **kwargs: Any) -> Dict[str, List[str]]:
```
efiop, line 149: Let's make these hidden
```suggestion
def _diff_index_to_wtree(repo: "Repo", **kwargs: Any) -> Dict[str, List[str]]:
```

efiop: Same with others in this file.

dvc/commands/data.py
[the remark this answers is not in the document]
jorgeorpinel: As for stdout, it's better to specify "DVC-(un)tracked)" since there's also Git-(un)tracked. Clearly DVC status commands shouldn't report about what Git is doing but there's a lot of similar functionality/terminology here, and users can get confused.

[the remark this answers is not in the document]
skshetry: Technically, `DVC-untracked == Git-untracked`. For others it makes sense, and we have `DVC` suffix added to them.

dvc/commands/data.py
skshetry: It's a bit verbose, to be honest. I think we can simplify it to:
```suggestion
        "not_in_cache": "Not in cache",
        "staged": "Changes tracked by DVC",
        "unstaged": "Changes not tracked by DVC",
```

I am not sure about `untracked` and `unchanged`. `not_in_cache` could stay simpler (I don't think `dvc` is adding any information, `cache` implies `dvc cache`).

skshetry:
@efiop, this should be ready for review/merge. :)

setup.cfg
```diff
@@ -125,6 +125,7 @@ tests =
     pytest-xdist==2.5.0
     pytest-mock==3.8.2
     pytest-lazy-fixture==0.6.3
+    pytest-test-utils==0.0.8
```
skshetry, line 128: Started using matchers from `pytest-test-utils`.

mattseddon:
Started on moving to the new command. Found an issue with untracked files. In the following output `untracked` is relative to the git root and not the dvc repository root (dvc root is in the demo subdirectory):

```
{                                                                                                                                                                                                    
  "uncommitted": {
    "modified": [
      "model.pt",
      "training_metrics/scalars/loss.tsv",
      "training_metrics/scalars/acc.tsv",
      "training_metrics/",
      "misclassified.jpg",
      "predictions.json"
    ]
  },
  "untracked": [
    "demo/test.pl"
  ]
}
```

tests/func/test_data_status.py
```diff
@@ -0,0 +1,148 @@
...
+
+    assert dvc.data_status(granular=True, untracked_files="all") == {
+        "committed": {
+            "added": M.unordered(
+                join("dir", "bar"),
+                join("dir", "foobar"),
+            )
+        },
+        "uncommitted": {
+            "added": [join("dir", "baz")],
```
mattseddon, line 57: [Q] (probably a dumb one or could be unimportant) Is this possible? Won't using `dvc add` send a resource straight to `committed`? Is there another way to track/add an untracked resource that I'm not aware of?

skshetry: > Is this possible?

yes, for example, when you modify the dataset, but you haven't done `dvc commit` them yet.

> Won't using dvc add send a resource straight to committed?

yes, this will make them `committed`. 

> Is there another way to track/add an untracked resource that I'm not aware of?

If they are part of the dataset, eg: `dataset/file` in a `dataset`, they can be simply added as `dvc commit dataset/file`.
These will show up in `uncommitted`.

Otherwise, if they are standalone and not a part of any dvc-tracked outputs, they'll need to be `dvc add` first. They will show up as `untracked`.

mattseddon: The reason I ask is I am trying to type the JSON output. Is the following correct:

```
{
  uncommitted: {
    deleted: string[]
    modified: string[]
  }
  committed: {
    added: string[]
    deleted: string[]
    modified: string[]
    renamed: string[]
  }
  not_in_cache: string[]
  unchanged: string[]
  untracked: string[]
}
```

Did I miss anything?

skshetry: That's correct, except `renamed` is not implemented yet.

mattseddon: thank you 🙏🏻

skshetry:
> Started on moving to the new command. Found an issue with untracked files. In the following output `untracked` is relative to the git root and not the dvc repository root (dvc root is in the demo subdirectory):
>
> LMK if this was deliberate. Thanks.

Great catch. That's an oversight on my part. Will fix it. Thanks.

skshetry:
@mattseddon, I have tried fixing the sub dir untracked files issue. Let me know if it fixes for you.

mattseddon:
> @mattseddon, I have tried fixing the sub dir untracked files issue. Let me know if it fixes for you.

LGTM

efiop (approved):

dberenbaum:
Great work on this @skshetry!

dvc/commands/data.py
```diff
@@ -0,0 +1,178 @@
...
+    }
+    LABELS = {
+        "not_in_cache": "Not in cache",
+        "committed": "DVC committed changes",
+        "uncommitted": "DVC uncommitted changes",
+        "untracked": "Untracked files",
+        "unchanged": "DVC unchanged files",
+    }
+    HINTS = {
+        "not_in_cache": 'use "dvc pull <file>..." '
```
jorgeorpinel, line 33: I thought we used back ticks in these texts? E.g.
```suggestion
        "not_in_cache": 'Use `dvc pull <file>` '
```

dvc/commands/data.py
```diff
@@ -0,0 +1,178 @@
...
+        "to update your local storage",
```
jorgeorpinel, line 34: - storage -> cache?
- Missing sentence period `.`

dvc/commands/data.py
```diff
@@ -0,0 +1,178 @@
...
+        "committed": "git commit the corresponding dvc files "
+        "to update the repo",
```
jorgeorpinel, line 36:
```suggestion
        "committed": "`git commit` the corresponding DVC metafiles "
        "to update the repo.",
```

dvc/commands/data.py
```diff
@@ -0,0 +1,178 @@
...
+        "uncommitted": 'use "dvc commit <file>..." to track changes',
```
jorgeorpinel, line 37:
```suggestion
        "uncommitted": 'Use `dvc commit <file>` to track changes.',
```

dvc/commands/data.py
```diff
@@ -0,0 +1,178 @@
...
+        "untracked": 'use "git add <file> ..." or '
+        'dvc add <file>..." to commit to git or to dvc',
```
jorgeorpinel, line 39:
```suggestion
        "untracked": 'Use `git add <file>` or '
        '`dvc add <file>` to track with Git or DVC.',
```

dvc/commands/data.py
```diff
@@ -0,0 +1,178 @@
...
+        "git_dirty": "there are {}changes not tracked by dvc, "
+        'use "git status" to see',
```
jorgeorpinel, line 41:
```suggestion
        "git_dirty": "There are {}changes not tracked by DVC. "
        'Use `git status` to see them.',
```

dvc/commands/data.py
```diff
@@ -0,0 +1,178 @@
...
+            yield stage, items
+
+    @classmethod
+    def _show_status(cls, status: "DataStatus") -> int:
+        git_info = status.pop("git")  # type: ignore[misc]
+        result = dict(cls._process_status(status))
+        if not result:
+            no_changes = "No changes"
+            if git_info.get("is_empty", False):
+                no_changes += " in an empty git repo"
```
jorgeorpinel, line 66:
```suggestion
                no_changes += " in an empty Git repo"
```

jorgeorpinel (commented):
Some string suggestions after checking https://github.com/iterative/dvc.org/pull/3812.

dvc/commands/data.py
```diff
@@ -0,0 +1,178 @@
...
+    )
+    data_subparsers = data_parser.add_subparsers(
+        dest="cmd",
+        help="Use `dvc data CMD --help` to display command-specific help.",
+    )
+    fix_subparsers(data_subparsers)
+
+    DATA_STATUS_HELP = (
+        "Show changes between the last git commit, "
+        "the dvcfiles and the workspace."
```
jorgeorpinel, line 130: Oh and migrating discussion from https://github.com/iterative/dvc.org/pull/3812#pullrequestreview-1059748858 cc @skshetry @shcheklein:

```suggestion
    DATA_STATUS_HELP = (
        "Show changes between the last Git commit, "
        "DVC metafiles, and the workspace."
```

jorgeorpinel: > metafile is only docs-concept
this term has not been introduced in DVC (maybe for good reasons).

I think the only reason is that we don't have a system to ensure consistency in terminology between docs and help output but we probably should. It would improve the UX.

> dvcfiles is also a weird term
we should... not introduce internal terms into help / docs.

➕

And it sounds like `.dvc` files specifically. What about `dvc.lock`?

jorgeorpinel: This could be further simplified though, to just:

`Show changes in DVC-tracked data between the last Git commit and the workspace.`

dvc/commands/data.py
```diff
@@ -0,0 +1,178 @@
...
+        "committed": "green",
+        "uncommitted": "yellow",
+        "untracked": "cyan",
+    }
+    LABELS = {
+        "not_in_cache": "Not in cache",
+        "committed": "DVC committed changes",
+        "uncommitted": "DVC uncommitted changes",
+        "untracked": "Untracked files",
+        "unchanged": "DVC unchanged files",
```
jorgeorpinel, line 30: One more.

```suggestion
    LABELS = {
        "not_in_cache": "Not in cache",
        "committed": "Committed changes",
        "uncommitted": "Uncommitted changes",
        "untracked": "Untracked files",
        "unchanged": "Unchanged files",
```

Reason: "DVC" seems unnecessary here (the command is `dvc ...` so it's assumed). And the wording can result confusing, e.g. do we mean "`dvc commit`-ed changes"? or "`git commit`-ed changes to DVC metafiles"?

dvc/repo/data.py
```diff
@@ -0,0 +1,259 @@
...
+    if isinstance(new_obj, Tree):
+        root = os.path.sep.join([root, ""])
+
+    if not _in_cache(old_obj, cache):
+        d["not_in_cache"] = [root]
+
+    if old_obj is None and new_obj is None:
+        return d
+    if old_obj is None:
+        return {"added": [root], **d}
```
jorgeorpinel, line 50: I thought of one more thing: This should be called `new` I think, because you can `dvc add` an updated file and it will be listed as `modified`.

dvc/commands/data.py
```diff
@@ -0,0 +1,178 @@
...
+        help="Show granular file-level info for DVC-tracked directories.",
+    )
+    data_status_parser.add_argument(
+        "--unchanged",
+        action="store_true",
+        default=False,
+        help="Show unmodified DVC-tracked files.",
+    )
+    data_status_parser.add_argument(
+        "--untracked-files",
```
jorgeorpinel, line 165: One more...

Should the 2nd option be just `--untracked`? Unclear why `-files` is only in that name (plus it can be directories too).

dvc/commands/data.py
shcheklein: I think DVC was introduced to be extra explicit about Git vs DVC ... otherwise committed might confuse people

dvc/repo/data.py
shcheklein: hmm, it's interesting Git does "new file" ... VS Code does "added"

dvc/commands/data.py
shcheklein: yep, let's introduce this alternative now, and deprecate the previous option, drop it later

dvc/commands/data.py
jorgeorpinel: I think it's clear what we're in a DVC context when using the `dvc data status` command. No strong opinion on keeping "DVC" but the wording could be clearer e.g. "Committed to DVC" and "Not committed to DVC".

shcheklein: might be ... but this command takes both - Git and DVC concepts (compares workspace to HEAD). Again, it's fine to keep it super explicit here, let's not optimize here for now

dvc/commands/data.py
skshetry: Metafile is a confusing term. Tbh I don't even know what it is, I can guess that it's a file having metadata about something. Also, it's a logical concept, not a physical one like `.dvc` and `dvc.yaml`, so I find it to be an unnecessary redirection. Unlike docs, we can afford repetition which is not that many.

Here, I think we can just avoid mentioning the files at all.
> Show changes in the data tracked by DVC in the workspace.

dvc/commands/data.py
skshetry: We have made changes to this in successive PRs. Now it says:
```
```

dvc/commands/data.py
skshetry: I prefer quotes, it's easier on eyes. :)

dvc/commands/data.py
skshetry: It's a hint, no need to upper case it.

dvc/commands/data.py
shcheklein: > Show changes in DVC-tracked data between the last Git commit and the workspace.

yep, we can avoid metafiles, dvcfiles.

-----

my 2cs - I find both terms suboptimal, but would probably prefer metafiles - since, yes they are technically metafiles - they contain a spec for data, yes metadata about data. At least it sounds reasonable to me. Plus we already use it in docs. May be we can avoid using this low lever terminology in help messages at all btw- that would the best option for end users.

dvc/commands/data.py
shcheklein: I guess it's a convention already to code-inline this kind of things?

skshetry: We do use backticks on help messages, but I am not sure about the reasoning. The only other place we print commands is in `git add` hints where we just indent those.

dvc/commands/data.py
skshetry: Git uses `--untracked-files`, so that's a major reason why this is this way. Also, `--untracked-files` at the moment is recursive, so the `files` is more correct. There are some questions regarding it's behaviour: https://github.com/iterative/dvc/issues/8061, so I'd defer it until then.

dvc/repo/data.py
skshetry: I am not sure about this one tbh.

dvc/commands/data.py
jorgeorpinel: > Here, I think we can just avoid mentioning the files at all.
we can avoid using this low lever terminology in help messages

➕➕

dvc/repo/data.py
jorgeorpinel: > VS Code does "added"

I see `A` (added) in both Explorer and Source Control views:

image

image

dvc/commands/data.py
jorgeorpinel: OK but remember that Git's UI is not very good. We should not aim to copy it just because it's Git. "Untracked" can refer to plural files so both are equally correct, I think.

dvc/commands/data.py
jorgeorpinel: My only rec. here is to aim for consistency if possible. Probably not major (up to you) but it's a product quality question.

dvc/commands/data.py
jorgeorpinel: I don't understand why hints don't need capitalization or punctuation. Are they part of a more complete sentence?

Issue #7949, opened 2022-06-29
gdrive: raises unexpected error - name: drive version: v2 (again)


BrettRyland:
# Bug Report

<!--
## Issue name

Issue names must follow the pattern `command: description` where the command is the dvc command that you are trying to run. The description should describe the consequence of the bug. 

Example: `repro: doesn't detect input changes`
-->

## Description

As reported [here](https://github.com/iterative/dvc/issues/5618#issuecomment-1155019515), the latest version (2.11.0) of dvc on Ubuntu 22.04 LTS still has the bug reported in issue [#5618](https://github.com/iterative/dvc/issues/5618).
Downgrading dvc to 2.10.2 allows the push to work properly.

### Reproduce

On an existing repository using gdrive, add some files then try to push them.
```console
brett@br-workhorse:~/repos/arctic-ice$ dvc status
models.dvc:                                                                                                                                                                                                                                                                                           
        changed outs:
                modified:           models

brett@br-workhorse:~/repos/arctic-ice$ dvc add models
100% Adding...|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████|1/1 [00:01,  1.52s/file]

brett@br-workhorse:~/repos/arctic-ice$ dvc push                                                                                                                                                                                                                                                       
ERROR: unexpected error - name: drive  version: v2                                                                                                                                                                                                                                                    

Having any troubles? Hit us up at https://dvc.org/support, we are always happy to help!
```

With extra verbosity:
```console
brett@br-workhorse:~/repos/arctic-ice$ dvc push -vv
2022-06-29 19:59:18,216 TRACE: Namespace(all_branches=False, all_commits=False, all_tags=False, cd='.', cmd='push', cprofile=False, cprofile_dump=None, func=<class 'dvc.commands.data_sync.CmdDataPush'>, glob=False, instrument=False, instrument_open=False, jobs=None, parser=DvcParser(prog='dvc', usage=None, description='Data Version Control', formatter_class=<class 'argparse.RawTextHelpFormatter'>, conflict_handler='error', add_help=False), pdb=False, quiet=0, recursive=False, remote=None, run_cache=False, targets=[], verbose=2, version=None, viztracer=False, viztracer_depth=None, with_deps=False, yappi=False)
2022-06-29 19:59:18,399 TRACE:    16.49 ms in collecting stages from /home/brett/repos/arctic-ice
[...]

PR #7954, opened 2022-06-30
build: Downgrade pyinstaller

daavoo:
Per https://github.com/iterative/dvc/issues/7949#issuecomment-1170927010


efiop:
Added a note to that file so we remember why it is frozen.

PR #7961, opened 2022-07-02
logger: use lazy formatting

### LINKED ISSUE #1843: logger: use logging format interpolation

alexmojaki:
Fixes #1843

### DIFF
--- a/dvc/analytics.py
+++ b/dvc/analytics.py
@@ -187,4 +187,4 @@ def _find_or_create_user_id():
             return user_id
 
     except LockError:
-        logger.debug("Failed to acquire %r", lockfile)
+        logger.debug("Failed to acquire '%s'", lockfile)
--- a/dvc/analytics.py
+++ b/dvc/analytics.py
@@ -50,7 +50,7 @@ def is_enabled():
             Config(validate=False).get("core", {}).get("analytics", "true")
         )
 
-    logger.debug("Analytics is {}abled.".format("en" if enabled else "dis"))
+    logger.debug("Analytics is %sabled.", "en" if enabled else "dis")
 
     return enabled
 
@@ -187,4 +187,4 @@ def _find_or_create_user_id():
             return user_id
 
     except LockError:
-        logger.debug(f"Failed to acquire '{lockfile}'")
+        logger.debug("Failed to acquire %r", lockfile)
--- a/dvc/data_cloud.py
+++ b/dvc/data_cloud.py
@@ -64,7 +64,7 @@ class DataCloud:
         return get_odb(cls(**config), fs_path, **config)
 
     def _log_missing(self, status: "CompareStatusResult"):
-        if status.missing:
+        if status.missing and logger.isEnabledFor(logging.WARNING):
             missing_desc = "\n".join(
                 f"name: {hash_info.obj_name}, {hash_info}"
                 for hash_info in status.missing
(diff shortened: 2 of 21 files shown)


dvc/analytics.py
```diff
@@ -187,4 +187,4 @@ def _find_or_create_user_id():
             return user_id
 
     except LockError:
-        logger.debug(f"Failed to acquire '{lockfile}'")
+        logger.debug("Failed to acquire %r", lockfile)
```
alexmojaki, line 190: Here I have changed `'{lockfile}'` to `%r` (i.e. use the `repr`, as opposed to `%s`) which takes care of quotes and looks better if `lockfile` itself happens to contain single quotes. I've done this several times in this PR wherever the value appears to be a string.

dvc/repo/experiments/push.py
```diff
@@ -85,7 +85,7 @@ def _push(
     from ...scm import GitAuthError
 
     refspec_list = [f"{exp_ref}:{exp_ref}" for exp_ref in refs]
-    logger.debug(f"git push experiment '{refs}' -> '{git_remote}'")
+    logger.debug("git push experiment %r -> '%s'", refspec_list, git_remote)
```
alexmojaki, line 88: Here I've changed `refs` to `refspec_list` to match `pull.py` just above.

dvc/analytics.py
alexmojaki: Oops, this caused tests to fail on Windows with good reason, as backslashes got escaped too, turning `'C:\\Users\\runneradmin\\AppData...'` into `'C:\\\\Users\\\\runneradmin\\\\AppData...`. So I guess I'll revert all of these to just `'%s'`. Any objections?

dvc/data_cloud.py
```diff
@@ -64,7 +64,7 @@ def _init_odb(self, name):
         return get_odb(cls(**config), fs_path, **config)
 
     def _log_missing(self, status: "CompareStatusResult"):
-        if status.missing:
+        if status.missing and logger.isEnabledFor(logging.WARNING):
```
efiop, line 67: hm, could you explain the reasoning behind this change, please? Seems unrelated to the rest.

efiop (approved):
Thank you!

dvc/data_cloud.py
efiop: @alexmojaki That seems unnecessary though. I'll adjust it back for now

PR #7963, opened 2022-07-02
checkout: use dvcignore

### LINKED ISSUE #7374: checkout/pull: dvcignored contents are removed on checkout
```bash
cd "$(mktemp -d)"
dvc init --no-scm
dvc remote add -d local "$(mktemp -d)"
mkdir -p data/processed
echo "foo" > data/foo
echo "bar" > data/processed/bar
echo "data/processed/" >> .dvcignore
dvc add data
dvc push
rm -rf .dvc/cache data/foo
dvc pull | tee
```

This is a regression introduced in b349b39d6266a28705b03b1b06e5c0cf98a9080d (see #5812), which was released in [2.2.0](https://github.com/iterative/dvc/releases/tag/2.2.0) and is present since then.

alexmojaki:
Fixes #7374

Note that there are many places the `ignore` argument could be added instead. This is the deepest call, right before delegating to `dvc_data`. Not sure if that's the right choice.

Will add tests soon.

### DIFF
--- a/dvc/output.py
+++ b/dvc/output.py
@@ -583,6 +583,7 @@ class Output:
         from dvc_data.checkout import CheckoutError as _CheckoutError
         from dvc_data.checkout import LinkError, PromptError
 
+        kwargs.setdefault("ignore", self.dvcignore)
         try:
             return checkout(*args, **kwargs)
         except PromptError as exc:
@@ -627,7 +628,6 @@ class Output:
                 obj,
                 self.odb,
                 relink=True,
-                ignore=self.dvcignore,
                 state=self.repo.state,
                 prompt=prompt.confirm,
             )
--- a/tests/func/test_ignore.py
+++ b/tests/func/test_ignore.py
@@ -418,3 +418,27 @@ def test_run_dvcignored_dep(tmp_dir, dvc, run_copy):
     tmp_dir.gen({".dvcignore": "dir\n", "dir": {"foo": "foo"}})
     run_copy(os.path.join("dir", "foo"), "bar", name="copy-foo-to-bar")
     assert (tmp_dir / "bar").read_text() == "foo"
+
+
+def test_pull_ignore(tmp_dir, dvc, local_cloud):
+    from dvc.utils.fs import remove
+
+    tmp_dir.dvc_gen(
+        {
+            ".dvcignore": "data/processed/",
+            "data": {"foo": "foo", "processed": {"bar": "bar"}},
+        }
+    )
+    tmp_dir.add_remote(config=local_cloud.config)
+    dvc.add("data")
+    dvc.push()
+
+    foo_path = tmp_dir / "data" / "foo"
+    foo_path.unlink()
+    assert not foo_path.exists()
+
+    remove(tmp_dir / ".dvc/cache")
+    dvc.pull()
+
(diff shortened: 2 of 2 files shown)


tests/func/test_ignore.py
```diff
@@ -418,3 +418,27 @@ def test_run_dvcignored_dep(tmp_dir, dvc, run_copy):
...
+    tmp_dir.add_remote(config=local_cloud.config)
+    dvc.add("data")
+    dvc.push()
+
+    foo_path = tmp_dir / "data" / "foo"
+    foo_path.unlink()
+    assert not foo_path.exists()
+
+    remove(tmp_dir / ".dvc/cache")
+    dvc.pull()
```
alexmojaki, line 441: If I change this to `dvc.checkout()` then it fails again with the same error about removing files without confirmation. I tried to fix this but got stuck, I'm not familiar enough with this codebase. I can share what I found while debugging. I'm actually not even sure if the error is expected or not. The issue gave me the impression that both `pull` and `checkout` should succeed, is that correct?

dvc/output.py
```diff
@@ -583,6 +583,7 @@ def _checkout(self, *args, **kwargs):
         from dvc_data.checkout import CheckoutError as _CheckoutError
         from dvc_data.checkout import LinkError, PromptError
 
+        kwargs.setdefault("ignore", self.dvcignore)
```
karajan1001, line 586: Only two places are calling this `_checkout` method. One in line `583` had already had the `ignore` argument, and another one is in line `785` in this file. I think we should either remove both of them and add a default value to `ignore` or remove the default one but pass them explicitly in the calls.

karajan1001:
Thanks for your PR @alexmojaki.

tests/func/test_ignore.py
alexmojaki: Right, so basically it seems that `checkout` breaks down when the cache is deleted, which seems fair and maybe not worth trying to change? `fetch` updates the cache, I'm guessing `status -c` does too. And `pull` does a `fetch` before `checkout`.

Do you know if there are other ways to trigger this bug of `checkout` previously not ignoring files properly that don't involve deleting the cache?

karajan1001: `status -c` doesn't pull down the cache but it will pull down the `.dir` index.  And it is used in generating a data tree object. 
https://github.com/iterative/dvc/blob/954de0cae44e31dfcf45630a2cc7e7f5acc7f00c/dvc/output.py#L772
I think the current problem is that,  in `dvc-data` the `diff` function didn't take `dvcignore` into account when the tree object is missing.

https://github.com/iterative/dvc-data/blob/010cdfb0527209eebbcec4a33c476cf4f532f963/src/dvc_data/checkout.py#L252

karajan1001: To solve this completely solve this problem, some modifications to `dvc-data` is required.

karajan1001: Would you like to solve it yourself or wait for someone else to do this? But I think it is different problem. We can merge this one as it already solved one part of the question.

alexmojaki: I don't know how to solve it myself, someone else should do this.

karajan1001 (approved):

PR #7970, opened 2022-07-04
Add `--time` to `dvc gc` and `dvc exp gc`

### LINKED ISSUE #678: `dvc gc --time <datetime>`
#669 

@dmpetrov : 

> the data time could be used for the garbage collector - very natural way to reduce size. dvc gc --time month - clean everything not from the current branch and older than a month.

karajan1001:
fix #678 

1. Move old unittest gc tests to pytest
2. Add a new flag `--time` to `dvc gc` and `dvc exp gc`
3. Fix an argument checking error and update the error msg.
4. Add argument `commit_time` to `gc`/`brancher` and `iter_revs`
5. Add unit test for iter_revs.

Docs are still required to get this feature finished.


dvc/repo/brancher.py
```diff
@@ -72,7 +78,9 @@ def brancher(  # noqa: E302
             self.root_dir = self.fs.path.join("/", *repo_root_parts)
 
             if cwd_parts:
-                cwd = self.fs.path.join("/", *cwd_parts)
+                cwd = self.fs.path.join(  # type: ignore[unreachable]
```
karajan1001, line 81: It looks like it is an issue of `mypy`. I tried some tests and found that these parts of the code can be reached.

dvc/commands/gc.py
```diff
@@ -104,6 +107,17 @@ def add_parser(subparsers, parent_parser):
...
     )
+    gc_parser.add_argument(
+        "--time",
+        type=str,
+        dest="commit_time",
+        metavar="<commit_time>",
+        default=None,
+        help=(
+            "Keep experiments from the commits after(include) a certain time."
+            "The time format is '%Y-%m-%d'"
```
dberenbaum, line 118:
```suggestion
            "The time format is '%%Y-%%m-%%d'"
```

Currently getting an error:

```bash
$ dvc gc --help
ERROR: unexpected error - unsupported format character 'Y' (0x59) at index 85
```

dberenbaum:
Some questions:

1. What if all my commits are older than the date provided?
2. How should it interact with the other flags (`-w`, `-a`, etc.)?

dberenbaum:
> 2\. How should it interact with the other flags (`-w`, `-a`, etc.)?

I would suggest we keep each option mutually exclusive to simplify usage. WDYT?

Also, we need to update the warning message:

```bash
$ dvc gc --time 2022-07-04
WARNING: This will remove all cache except items used in the workspace of the current repo.
```

dberenbaum:
> > What if all my commits are older than the date provided?
> 
> It will remove all of their caches.

It looks like it will preserve the cache of the current workspace, right? Does it imply `-w`?
 
> > How should it interact with the other flags (-w, -a, etc.)?
> 
> The current behavior is that it will preserve all of the caches from commits satisfying any of these flags, or we can say it is a `or` relation in logic.
> 
> > I would suggest we keep each option mutually exclusive to simplify usage. WDYT?
> 
> This will make it impossible to preserve a union between them. For example all tags plus all branches.

Sounds good. This stuff always confuses me 😄 .

karajan1001:
> It looks like it will preserve the cache of the current workspace, right? Does it imply `-w`?

Currently no, I think we should make the workspace automatically satisfy the time limitation.

dvc/repo/brancher.py
```diff
@@ -17,6 +20,8 @@ def brancher(  # noqa: E302
         all_branches (bool): iterate over all available branches.
         all_commits (bool): iterate over all commits.
         all_tags (bool): iterate over all available tags.
+        commit_date (str): Keep experiments from the commits after(include)
+                            a certain time.  The time format is '%Y-%m-%d'
```
pmrowla, line 24:
```suggestion
                            a certain date. Date must match the extended ISO 8601 format (YYYY-MM-DD).
```

dberenbaum:
> > It looks like it will preserve the cache of the current workspace, right? Does it imply `-w`?
> 
> Currently no, I think we should make the workspace automatically satisfy the time limitation.

It preserves something for me, not sure if it's workspace or HEAD. See this bash script:

```bash
git clone git@github.com:iterative/example-get-started.git
cd example-get-started
dvc pull
dvc gc -f --time 2023-01-01
```

I get `No unused 'local' cache to remove.`

karajan1001:
> It preserves something for me, not sure if it's workspace or HEAD. See this bash script:

Sorry, I checked the code yesterday and found that both dvc gc and `exp gc` actually preserve `workspace` (all of modes are doing this ), And there are some problems with the warning messages for them, which had been updated by this PR.

dberenbaum (approved):

pmrowla:
LGTM, @karajan1001 please squash the review fix commits

dvc/commands/experiments/gc.py
```diff
@@ -110,6 +120,17 @@ def add_parser(experiments_subparsers, parent_parser):
         default=False,
         help="Keep experiments derived from all Git commits.",
     )
+    experiments_gc_parser.add_argument(
+        "--date",
+        type=str,
+        dest="commit_date",
+        metavar="<commit_date>",
```
jorgeorpinel, line 127: How about `<yyyymmdd>` or something that gives an idea of the type/format expected? `--data <date>` is not very informative.

pmrowla (approved):

Issue #7695, opened 2022-05-04
gc files older than specific rev


Danial-Alh:
I want to remove all DVC files being at least ten commits behind HEAD from my remote repo. Is there any way to do that with the `gc` command?

Thanks

karajan1001:
Looks like we should also introduce our `-n` flag to `dvc gc` and `dvc exp gc`

dberenbaum:
@karajan1001 How much time would it take to implement this?

karajan1001:
> @karajan1001 How much time would it take to implement this?

As we had already refactored the branched I think it'll cost about 1-2 days.

Issue #5066, opened 2020-12-09
allow skipping certain git revisions/commits in DVC


skshetry:
If the `dvc.yaml` files are corrupted in certain revisions, `dvc` just fails and there's no way to skip that.
@courentin raised #3885 for this and again on #5037 to specifically address this.

For most parts of DVC, it might be okay and we can also teach params/metrics/plots to cope with failures in certain revisions.
But, for other commands such as push/pull/gc, dvc might just fail, even if we skip those failing commits, it'd be better not to skip and ask users to make their intentions clearer to provide stronger "data versioning" requirements.

And, for `gc`, it's even dangerous, and for `push`/`pull/`gc`, it might just be confusing.

So, coming to my proposal, it'd be better if we could provide a way for users to tell DVC to skip certain commits.
One of the ways would be to provide `--ignore-revs` or similarly named flags in `push/pull/gc`, etc.

But, it might not be always easier to use those flags and might be cumbersome especially in CI.
To solve this, we can provide a config (eg: `ignore_revsFile`) where you could provide a file with list of revisions to ignore:

```
# .dvc/config
[core]
ignore_revsFile = .dvc-ignore-revs

# .dvc-ignore-revs
# list of revisions for the DVC to ignore
a12ef
12345
```

The file will always be read from the latest checked out commit.

If the file is cumbersome, we could instead provide a way of adding a list of revs in `.dvc/config` as they are always read from the workspace.

efiop:
Could consider creating a `gc`-specific section for it and for future gc-related config options.

efiop:
Viewer team might find this interesting too CC @Suor

Suor:
We simply skip broken commits, not sure this will affect viewer in any way.

courentin:
I feel having a `.dvc-ignore-revs` is a good idea indeed!

Suor:
Might be used to silent errors in viewer too, I guess. And skipping is faster than parsing of cause.

dberenbaum:
Some related work was done in #5984 for certain commands to skip bad revisions. Specifying revisions to ignore would still be a nice feature, but I'm lowering the priority for now since it's not in our current roadmap.

PR #7971, opened 2022-07-04
gc: Add support for `--rev` and `--num`.

daavoo:
Closes https://github.com/iterative/dvc/issues/7695


karajan1001:
Related to #7970 , I also used to want to implement `-n` but later switch to `--time`. The difference comes from when you are working on a tree-like `Git` repository. If you use `gc -n` on a branch then it might clean some other branches. The `-n` is designed for picking up a set of experiments to show or share. I think it's not so friendly on `gc` especially for those big and complex projects.

But I admit that didn't take the combination of `--rev` into account. This can solve the previous `multi branch` problem.

pared (changes requested):
What is the intended usage here? 
If I run:
`dvc gc --rev HEAD --num 2 `
I get
`dvc.exceptions.InvalidArgumentError: Either of `-w|--workspace`, `-a|--all-branches`, `-T|--all-tags` `--all-experiments` or `--all-commits` needs to be set.`

EDIT:
seems like `rev` is not passed in `commands/gc.py`, also there is another call of `_raise_error_if_all_disabled` there.

pared:
For the following example:
```
#!/bin/bash

set -xu
pushd $TMPDIR

wsp=test_wspace
rep=test_repo

main=$(pwd)

mkdir $rep && pushd $rep

git init
dvc init

echo 1 > data.txt

dvc add data.txt
git add -A
git commit -am "first on master"

git checkout -b some_branch
echo 9 > data.txt
dvc commit -f
git commit -am "first on branch"

git checkout master

echo 2 > data.txt
dvc commit -f
git commit -am "second on master"

echo 3 > data.txt
dvc commit -f
git commit -am "third on master"

echo 4 > data.txt
dvc commit -f
git commit -am "fourth on master"

git merge some_branch 
git checkout --theirs data.txt.dvc
git commit -am "merged with some_branch"

LATEST=$(git rev-parse HEAD)
dvc gc --rev $LATEST --num 3 -f
```

We have following revision log:

```
*   59dd199 (HEAD -> master) merged with some_branch
|\  
| * b7aea0a (some_branch) first on branch
* | 44b2999 fourth on master
* | 203bd88 third on master
* | 36e87f3 second on master
|/  
* 6c4d660 first on master
```
What is the expected behavior here?
I specified `--num 3`. Should data from `first on master` be preserved?

dvc/commands/gc.py
```diff
@@ -18,6 +18,7 @@ def run(self):
             all_tags=self.args.all_tags,
             all_commits=self.args.all_commits,
             workspace=self.args.workspace,
+            rev=self.args.rev
```
daavoo: Not sure why we also call `_raise_error_if_all_disabled` given that we also call it inside `self.repo.gc`

dvc/commands/gc.py
```diff
@@ -11,15 +11,6 @@
 
 class CmdGC(CmdBase):
     def run(self):
-        from dvc.repo.gc import _raise_error_if_all_disabled
```
daavoo, line 14: I's already being called inside `repo.gc`

dberenbaum:
Need to coordinate with https://github.com/iterative/dvc/pull/7970.

Similar to that PR, it's unclear how `--rev` and `-n` are expected to interact with the other flags.

tests/func/test_gc.py
```diff
@@ -401,3 +401,22 @@ def test_gc_all_experiments(tmp_dir, scm, dvc):
     assert (
         tmp_dir / ".dvc" / "cache" / baz_hash[:2] / baz_hash[2:]
     ).read_text() == "baz"
+
+
+def test_gc_rev_num(tmp_dir, scm, dvc):
```
karajan1001, line 406: As we had already had a unit test for `iter_revs` with `--num` and `--rev`, maybe we only need a mocked test?

daavoo: > maybe we only need a mocked test?

Maybe it would be enough, although given that the added one is not expensive I prefer to have a somehow redundant but functional test

daavoo:
> Need to coordinate with #7970.
> 
> Similar to that PR, it's unclear how `--rev` and `-n` are expected to interact with the other flags.

As answered in https://github.com/iterative/dvc/pull/7970#issuecomment-1175798432 , flags are complementary. 

I think the message prompted makes this kind of clear:

```console
$ dvc gc -w
This will remove all cache except items used in 
the workspace
```
```console
$ dvc gc --all-tags
This will remove all cache except items used in 
the workspace 
and all git tags
```
```console
$ dvc gc --rev HEAD --num 2 --all-tags
This will remove all cache except items used in 
the workspace 
and all git tags 
and last 2 commits from HEAD
```

karajan1001 (approved):

daavoo:
> I was confused when trying examples like this:
> 
> ```shell
> $ dvc gc -n 2 # Why doesn't this work with default HEAD rev?
> WARNING: This will remove all cache except items used in the workspace of the current repo.
> Are you sure you want to proceed? [y/n]: y
> ERROR: Either of `-w|--workspace`, `-a|--all-branches`, `-T|--all-tags` `--all-experiments`, `--all-commits` or `--rev` needs to be set.
> $ dvc gc -a -n 2 # Why does this work?
> WARNING: This will remove all cache except items used in the workspace and all git branches of the current repo.
> Are you sure you want to proceed? [y/n]: y
> No unused 'local' cache to remove.
> ```
> 
> It's unclear if `dvc gc -a -n 2` will:
> 
>     1. Remove all cache except items used in all branches and the last 2 commits from HEAD
> 
>     2. Remove all cache except items used in all branches and the last 2 commits from each branch.
> 
> 
> It sounds like the first is correct, but the usage doesn't make this clear to me.

None of the 2 are correct. `-n` does nothing unless `--rev` is also passed.
`dvc gc -a -n 2` does the same as `dvc gc -a`. We should raise an exception if `-n` is provided but not `--rev`.

> It might be simpler to leave out `--rev`. I'm having a hard time imagining when it would be useful, and it would be easier to explain `-n` as being strictly in relation to HEAD.

I don't see a reason to restrict it to only `HEAD`.

Would the behavior still be confusing if we raise an error if `-n` is provided but not `--rev`?

daavoo:
> > I don't see a reason to restrict it to only `HEAD`.
> 
> I was thinking we could at least default `--rev` to `HEAD`, which would simplify it and conforms to some other commands like `dvc exp list/show`, but I guess the point with these dangerous commands is to make them kind of a pain and very explicit to avoid accidents, right?

Yes, I think that for the specific case of `gc` is better to require explicit args.

> > Would the behavior still be confusing if we raise an error if `-n` is provided but not `--rev`?
> 
> 👍
> 

Passing `--num` without `--rev` now raises 
```
ERROR: `--num` can only be used alongside `--rev`
```

> What if I wanted to do something like `dvc gc --rev HEAD --rev baseline --rev some_other_tag`? That would depart from how `--rev` is used in `exp` commands and is a separate request, but I'm wondering if we are making that future possibility harder by introducing a different `--rev` behavior (related to #5066, which asks for _skipping_ specific revs). Not a blocker, just a random thought.

Opened #8010

dberenbaum (approved):

PR #7973, opened 2022-07-04
metrics: recursively show directories specified as metrics in pipeline

### LINKED ISSUE #2973: Adding multiple metrics files at once
Hello,

my current ML project runs a DVC pipeline which trains a model for several different segments of the data, and produces a metrics file for each one.
I want to be able to track how these metrics evolve over time; however, the segments of data that are evaluated in each run depend on some parameters on the first step of the pipeline, and so the number of output metrics files changes every time I run the pipeline.

I have tried several approaches:
* adding `-m metrics` to my DVC run command, where metrics/ is a folder where I store all my metrics files. This worked, but when calling `dvc metrics show -a -R` it simply showed the DVC file for that folder, instead of parsing the individual metrics.
* Using a wildcard -m metrics/* - this simply doesn't work.
* Setting the metrics directory as an output with `-o`, and then adding each file as a metric using a script loop: I get the following error `ERROR: failed to add metric file 'metrics/metric-1.json' - unable to find DVC-file with output 'metrics/metric-1.json`

I am out of ideas! Is this functionality supported, or do I need to have a fixed number of metrics files for each pipeline?

alexmojaki:
Fixes #2973

### DIFF
--- a/dvc/output.py
+++ b/dvc/output.py
@@ -709,21 +709,15 @@ class Output:
                 f"verify metric is not supported for {self.protocol}"
             )
 
-        if not (self.metric or self.plot):
+        if not self.metric or self.plot:
             return
 
         if not os.path.exists(self.fs_path):
             return
 
-        if os.path.isdir(self.fs_path):
-            if self.plot:
-                logger.debug("directory '%s' cannot be used as plot.", self)
-            # Metrics allow directories, avoid the istextfile check
-            return
-
+        name = "metrics" if self.metric else "plot"
         if not istextfile(self.fs_path, self.fs):
             msg = "binary file '{}' cannot be used as {}."
-            name = "metrics" if self.metric else "plot"
             raise DvcException(msg.format(self.fs_path, name))
 
     def download(self, to, jobs=None):
--- a/dvc/output.py
+++ b/dvc/output.py
@@ -712,7 +712,7 @@ class Output:
         if not self.metric or self.plot:
             return
 
-        if not os.path.exists(self.fs_path):
+        if not os.path.isfile(self.fs_path):
             return
 
         name = "metrics" if self.metric else "plot"
--- a/dvc/output.py
+++ b/dvc/output.py
@@ -715,14 +715,15 @@ class Output:
         if not os.path.exists(self.fs_path):
             return
 
-        name = "metrics" if self.metric else "plot"
         if os.path.isdir(self.fs_path):
-            msg = "directory '%s' cannot be used as %s."
-            logger.debug(msg, str(self), name)
+            if self.plot:
+                logger.debug("directory '%s' cannot be used as plot.", self)
+            # Metrics allow directories, avoid the istextfile check
             return
 
         if not istextfile(self.fs_path, self.fs):
             msg = "binary file '{}' cannot be used as {}."
+            name = "metrics" if self.metric else "plot"
             raise DvcException(msg.format(self.fs_path, name))
 
     def download(self, to, jobs=None):
--- a/dvc/output.py
+++ b/dvc/output.py
@@ -709,7 +709,7 @@ class Output:
                 f"verify metric is not supported for {self.protocol}"
             )
 
-        if not self.metric or self.plot:
+        if not (self.metric or self.plot):
             return
 
         if not os.path.exists(self.fs_path):
--- a/dvc/commands/update.py
+++ b/dvc/commands/update.py
@@ -28,7 +28,10 @@ class CmdUpdate(CmdBase):
 
 
 def add_parser(subparsers, parent_parser):
-    UPDATE_HELP = "Update data artifacts imported from other DVC repositories."
+    UPDATE_HELP = (
+        "Update data artifact imported (via dvc import or dvc import-url) "
+        "from an external DVC repository or URL."
+    )
     update_parser = subparsers.add_parser(
         "update",
         parents=[parent_parser],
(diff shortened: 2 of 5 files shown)


dvc/output.py
```diff
@@ -709,20 +709,21 @@ def verify_metric(self):
                 f"verify metric is not supported for {self.protocol}"
             )
 
-        if not self.metric or self.plot:
+        if not (self.metric or self.plot):
```
alexmojaki, line 712: This change is not related to the rest of the PR. The original line looked like a mistake (I make this kind of mistake all the time) since the code below (especially `name = "metrics" if self.metric else "plot"`) implied that plots still needed to be checked. But now the test `test_ignore_parsing_error` fails when calling `run_copy_metrics`. It's hard to tell, but it looks like the test is meant to be testing that errors are ignored in `show`, not `run`. So should I update the test to not fail on `run_copy_metrics`? Or should I remove checks for plots in this method?

alexmojaki:
Should I make a docs PR for this?

dvc/output.py
pmrowla: This was not a bug. The line setting name to `plot` was left over from a point when plots did need to be checked, but that is no longer the case, plots are now permitted to be non-text files.

pared: Yep, this check blocks later potential exceptions that might occur, for:
1. non text files - as Peter mentioned we do support binaries for some time now
2. dirs - plots can be dirs

"plots" mentions in strings seems to be obsolete logic that haven't been caught when introducing support for images/dirs. We should probably fix that, but it should be a separate PR. @alexmojaki can I ask you to create issue for that?

alexmojaki: Is it possible for something to be both a metric and a plot?

pmrowla:
@alexmojaki thanks for the PR, but I'm not sure this is actually behavior we want in DVC right now. Basically in the original issue (https://github.com/iterative/dvc/issues/2973) there are two separate problems:

1. `dvc metrics show -a -R` does not actually recursively find all metrics files in the repo
2. DVC does not allow metrics to be a directory

I think the `help wanted` tag on the original issue was applied with regard to the first problem, but it is actually already resolved in the latest version of DVC. `metrics show` now uses the unified repo collection methods which properly walk the entire tree when using the recursive flags. (This is our mistake in not updating the tags/discussion in the original issue)

This PR actually implements the second feature, which is a pretty significant change in DVC behavior and needs more discussion as to whether or not this should be supported (https://github.com/iterative/dvc/issues/3930, https://github.com/iterative/dvc/issues/2973#issuecomment-639185271)

dvc/output.py
```diff
@@ -712,15 +712,10 @@ def verify_metric(self):
         if not self.metric or self.plot:
             return
 
-        if not os.path.exists(self.fs_path):
+        if not os.path.isfile(self.fs_path):
             return
```
pmrowla, line 716: If we are going to allow a metrics output to be a directory, we need to do the validation checks on the files inside the directory (and walk the directory tree). We cannot skip the checks entirely

alexmojaki:
I'm with @dberenbaum. One comment links to another comment which links to discord, but I couldn't see any reasoning behind the decisions.

The way I see it, it's good to be able to specify metrics paths in stage/pipeline files so that they don't have to be specified as targets of the `metrics show` command. Currently that applies to files but not directories. `metrics show` supports directories for explicit targets with the `-R` flag, so I think it should do the same for directories specified in stage/pipeline files.

Since `metrics show` ignores directory targets unless `-R` is specified, one might ask if there should be a similarly explicit flag for metrics directories in stage/pipeline files. That really depends on why `-R` is even required in the first place. After all, specifying a directory path in the CLI already seems pretty explicit. The only reason I can think of is that a directory may be unintentionally targeted by a command with a wildcard such as `dvc metrics show *`. Since I don't think such wildcards are supported in pipeline files, I don't think that reason applies here. Is there any other reason that `-R` isn't just assumed?

pmrowla:
Supporting metrics directories is something that can be done in DVC, which is why the original issue (#2973) remains open.

The problem here is that "supporting metrics directories" goes beyond addressing `dvc metrics show -R` behavior. It will require updating DVC internals behavior (for outputs, stage add/save, repo collection, and the entire metrics + exp UI) and these changes all have to be tested.

After discussion with the rest of core DVC team (and @dberenbaum), due to the scope of the work required, this is something that we would prefer be implemented by the core team when we have the bandwidth to do so. And as I noted earlier, the original `help wanted` tag was regarding the incorrect `metrics show -R` UI behavior from 2 years ago, which has already been taken care of.

@alexmojaki sorry for any confusion here, and we do appreciate the PR! But for now we are going to close the PR and keep the original issue open.

If you have a specific need for metrics directories support, please feel free to discuss your use case/requirements in the original issue, as it will help us determine how much the core team needs to prioritize the feature request.

dvc/output.py
pared: In the usual use case, no. Though we have (now deprecated, but still supported till 3.0) `live` output which is using logic that was a mix of both. Why?

alexmojaki: I was wondering what the cleaned version of this code would look like. Keeping `if not self.metric or self.plot:` as is instead of just `if not self.metric:` seems to imply that `self.metric and self.plot` might be true, in which case checks would be skipped.

pared: Huh, this is a very valid point. I don't think there is a use case where `self.metric==True and self.plots==True`. I guess we could simplify this check.

PR #7976, opened 2022-07-05
dvc queue: merge dev-task dev branch

pmrowla:

Ongoing docs PR: https://github.com/iterative/dvc.org/pull/3715

TODO:
- [ ] https://github.com/iterative/dvc-task/issues/72


dberenbaum:
Good point. 

Do we know how much it tends to slow down install (obviously it depends, but curious if we have any estimates or benchmarks)? Do you foresee installation causing any problems beyond size/time?

I see 2 ways we could handle it:
1. Add `dvc[dvc-task/queue/whatever]`. If we do this, can we ensure that any commands dependent on queuing fail quickly with a useful hint to install the missing dependencies?
2. Add something like `dvc[core]` to install only the minimal deps for DVC, but default to installing everything (see https://docs.dask.org/en/stable/install.html). `dvc[all]` is unfortunately kind of useless outside of testing since all other optional deps are about different storage backends that are never all wanted.

I'd be fine with either, but I think how we handle the missing dependencies in option 1 is important if we go that way.

Stupid question, but I assume the binaries are equivalent to `dvc[all]`?

skshetry:
> 2. Add something like `dvc[core]` to install only the minimal deps for DVC, but default to installing everything

This is not possible, extras cannot be a subset of `install_requires`. The only way this can work is by having a separate package `dvc-core`, like it’s done in the doc that you have shared.

shcheklein:
This also can have a direct impact on the VS Code extension since it exposes queue quite directly in its UI  - we would need to detect the missing parts of DVC and ask users to install or ideally install it for them. I also curious first to know more about the `Do we know how much it tends to slow down install (obviously it depends, but curious if we have any estimates or benchmarks)? Do you foresee installation causing any problems beyond size/time?`.

shcheklein:
Or even, does it make sense to release it as-is and iterate if we see some significant problems?

pmrowla:
> Stupid question, but I assume the binaries are equivalent to dvc[all]?

Yes, this is correct

> Do we know how much it tends to slow down install (obviously it depends, but curious if we have any estimates or benchmarks)? Do you foresee installation causing any problems beyond size/time?

I don't think there is any noticeable impact on install time. With regard to size, on python3 and macos celery+deps adds about 10MB of disk space (currently DVC with no remotes installed is ~115MB)

It's probably fine for us to leave it as-is for now.

dberenbaum:
@pmrowla What's the status?

pmrowla:
We're still waiting on the dvc-task conda package to be approved but otherwise this is ready to be merged

skshetry:
> We're still waiting on the dvc-task conda package to be approved but otherwise this is ready to be merged

We can wait for that downstream on dvc-feedstock.

skshetry:
Studio uses `celery`, and a very old version of that (recent is `5.3`, it uses `4.4.2`), so they are definitely going to be in conflict via indirect `kombu` dependency. We won't be able to update studio to the dvc's latest without either upgrading celery to the latest version in Studio (DVC depends on an alpha version of kombu, so that'll still not solve a issue right now).

efiop:
@skshetry Studio will need to update on their end. Shouldn't be a problem.

PR #7998, opened 2022-07-11
tests: limit printed durations to 50

daavoo:
Having 300+ lines reporting durations of `0.X` seconds doesn't feel helpful in the CI context and makes the navigation of the logs cumbersome. 
The flag can still be overridden locally.


efiop:
Thanks!

